PR #18, opened 2018-03-18
chore: Better error message in new command.

playpauseandstop:
Update error message for existed directory in `new` command: add a space after directory path,

## Before

```
octave:path playpauseandstop$ poetry new .

                                                                                 
  [RuntimeError]                                                                 
  Destination /Users/playpauseandstop/Projects/pathexists and is not empty  
                                                                                 

new [--name] [--] <path>
```

## After

```
octave:path playpauseandstop$ poetry new .

                                                                                 
  [RuntimeError]                                                                 
  Destination /Users/playpauseandstop/Projects/path exists and is not empty  
                                                                                 

new [--name] [--] <path>
```

### DIFF
--- a/poetry/console/commands/new.py
+++ b/poetry/console/commands/new.py
@@ -26,7 +26,7 @@ class NewCommand(Command):
             if list(path.glob('*')):
                 # Directory is not empty. Aborting.
                 raise RuntimeError(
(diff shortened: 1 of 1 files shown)


sdispater:
Thanks!

PR #32, opened 2018-03-24
Allow the use of the period character in dependency names.

### LINKED ISSUE #31: packages with a . in their name fail to install
```
$ poetry add Qt.py
Using version ^1.1 for Qt.py

  [InvalidProjectFile]
  [dependencies] 'Qt.py' does not match any of the regexes: '^[a-zA-Z-_0-9]+$'
```
The `pyproject.toml` file looks correct but Poetry isn't happy with the dot it seems?

monkeez:
Update to the schema file to allow packages to have a period in their name, to fix #31.

I'm not too knowledgeable with regex so please let me know if this is incorrect or if this will cause other knock-on issues in Poetry.

Thanks!

### DIFF
--- a/poetry/json/schemas/poetry-schema.json
+++ b/poetry/json/schemas/poetry-schema.json
@@ -61,7 +61,7 @@
                 }
             },
             "patternProperties": {
-                "^[a-zA-Z-_0-9]+$": {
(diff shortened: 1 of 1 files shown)


sdispater:
Thanks!

Issue #77, opened 2018-04-30
poetry install bad python version


jgirardet:
If poetry is installed with python 3.6, when you specifiy python2.7 in [tool.poetry.dependencies] it will create a virtualenv under python 3.6.
I know it's the intended behaviour  (it's in the doc) but I think It's a real issue.

So poetry should check while creating virtualenv that the python version are not compatible (or  not the expected one).
Good idea are every where and the pipenv way (specify --python 3.6 or - python 2.7) is maybe a good way to do it.

thanks for poetry

jgirardet:
maybe something like this :
```poetry install
Warning : THe current python version is not supported in pyproject.toml, do you want to continue ?(y/n).
```
and/ or
```
poetry run something
warning: you're running under a python version not supported in pyproject.toml
```

sdispater:
@jgirardet @rainyday Thanks both of you for your input. I agree that `poetry` should give some kind of feedback that there is an incompatibility. I will try to add it to the next feature release (`0.9.0')

And @rainyday the purpose of the `python` dependency is to declare for which Python versions your project is compatible with. This is particularly useful if you publish it.

jgirardet:
I can propose some help if you want

cs01:
I am running into this too.

PR #82, opened 2018-05-01
WIP : Change python version at install

jgirardet:
This PR is  related to #77 

you can choose your python version:
`poetry install --python 3.6`

It will install the asked python version if present on the system.

Doing this I have to change the virtualenv name to make poetry know where to search.
The new name is project name + a md5 calculated from the path of the project
`/path/to/virtualenv/myproject-fzef84z38ef4ze8f4z3e`
So the venv can always be found if the project's path doesn't change.

Tests are not yet fully ok.

this doesn't solve the problemeof consistency between the created venv and  the python version needed in the pyproject.toml

### DIFF
--- a/poetry/console/commands/install.py
+++ b/poetry/console/commands/install.py
@@ -10,6 +10,7 @@ class InstallCommand(VenvCommand):
         { --dry-run : Outputs the operations but will not execute anything
                       (implicitly enables --verbose). }
         { --E|extras=* : Extra sets of dependencies to install. }
+        { --python=* : Python version used to create virtualenv. ex :  2.7, 3.5 }
     """
 
     help = """The <info>install</info> command reads the <comment>pyproject.toml</> file from
--- a/poetry/console/commands/install.py
+++ b/poetry/console/commands/install.py
@@ -10,9 +10,8 @@ class InstallCommand(VenvCommand):
         { --dry-run : Outputs the operations but will not execute anything
                       (implicitly enables --verbose). }
         { --E|extras=* : Extra sets of dependencies to install. }
-        { --python=* : Python version used to create virtualenv. ex :  2.7, 3.5 }
+        { --python=* : Python version used to create virtualenv. ex 2.7, 3.5 }
     """
-
     help = """The <info>install</info> command reads the <comment>pyproject.toml</> file from
 the current directory, processes it, and downloads and installs all the
 libraries and dependencies outlined in that file. If the file does not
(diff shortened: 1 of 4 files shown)


sdispater:
Thanks for taking the time to put this PR together.

Unfortunately, this is not something I want to implement in `poetry`. The idea is that if you want to use Python 2.7 for your project, you install poetry for the Python 2.7 of your system (ideally installed with `pyenv`).

jgirardet:
Hi,
I don't really understand where the problem is. If you don't specify anything it will install with the current system python version.
Maybe I'm wrong but I think that the 2 behaviors can exist together.

I think that this PR simplifies much the usage of poetry and makes it more accessible for new users.
I did this PR because I was lost at start using poetry about dealing with different python version.
Why would I use pyenv + poetry where everything can be done with poetry ?

PS: considering poetry is a  new (great) thing for new project I'm close to open a Issue asking to drop python2 and let you more time  for other things ;-)

thanks for this project anyway

jgirardet:
ok I see where you wanna go.
so in a certain way you have to use pyenv.
To be really sure that I understand, you install poetry in all pyenv versions or only in your system ?

jgirardet:
If you run different virtualenv with different python version, how the hashes un pyproject.lock could be ok for all different versions of python ?

sdispater:
When poetry is locking the dependencies it is agnostic and, thus, will lock any dependency regardless of its requirements.

However, when installing dependencies it will skip the ones that are not needed.

This is the same thing for hashes, every hashes are put in the lock file to make it it environment independant.

laike9m:
> Thanks for taking the time to put this PR together.
> 
> Unfortunately, this is not something I want to implement in `poetry`. The idea is that if you want to use Python 2.7 for your project, you install poetry for the Python 2.7 of your system (ideally installed with `pyenv`).

Say I'm using Poetry with pyenv-virtualenv, if I understand correctly, the recommended way is to install Poetry for every virtual environment, is that correct?

sdispater:
@laike9m Not anymore. If you use the recommended installer Poetry will pickup the currently activated Python version to create virtualenvs, so you only need to install it once.

Also if you are inside an activated virtualenv Poetry will automatically detect it and use it.

laike9m:
@sdispater That's really cool. Thanks!

paranoidi:
@sdispater That is not a good solution, it will create a virtualenv which is different (and quite possibly incompatible) from what is declared in pyproject.toml

LucianU:
I hit this issue as well. Using the recommended install approach on Mac OS X, I got `poetry` for Python 2. I discovered that when I tried to add `ipython` as a dependency of my project.

What happens if I want `poetry` to handle both projects for Python 2 and 3? Do I need to use `pyenv`?

PR #97, opened 2018-05-08
evaluate "path" dependencies as relative to the pyproject.toml, not to the CWD

radix:
poetry recently added support for searching upwards in the directory hierarchy for `pyproject.toml`. However, this wasn't playing well with the `path` dependency type when the dependency had a relative directory. This PR makes it evaluate `path` dependencies relative to the pyproject.toml instead of the CWD.

### DIFF
--- a/poetry/packages/package.py
+++ b/poetry/packages/package.py
@@ -110,6 +110,7 @@ class Package(object):
         self._platform_constraint = EmptyConstraint()
 
         self.cwd = None
+        self.root_dir = None
 
     @property
     def name(self):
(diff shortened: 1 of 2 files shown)


poetry/packages/package.py
```diff
@@ -110,6 +110,7 @@ def __init__(self, name, version, pretty_version=None):
         self._platform_constraint = EmptyConstraint()
 
         self.cwd = None
+        self.root_dir = None
```
sdispater: I think we can just drop cwd here in favor of root_dir to make it clearer.

sdispater:
Thanks!

PR #114, opened 2018-05-18
update README

cs01:
* update some spelling and grammer
* also add 

> In other words, poetry uses `pyproject.toml` to replace `setup.py`, `requirements.txt`, `setup.cfg`, `MANIFEST.in` and the newly added `Pipfile`. 

because I think that fact is a big selling point and should be stated up front

### DIFF
--- a/README.md
+++ b/README.md
@@ -96,8 +96,10 @@ fpath+=~/.zfunc
 
 ## Introduction
 
-`poetry` is a tool to handle dependencies installation, building and packaging of Python packages.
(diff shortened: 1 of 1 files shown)


jgirardet:
> In other words, poetry uses pyproject.toml to replace setup.py, requirements.txt, setup.cfg, MANIFEST.in and the newly added Pipfile.

+1 for this clearer explanation

sdispater:
Thanks!

PR #123, opened 2018-05-20
Add src to sys.path if module is in src

### LINKED ISSUE #122: poetry script: unable to find entry point module, if package resides in src
`poetry script` is unable to find the entry point module in `src`

```
[NameError]
name 'Module' is not defined
script <script-name> [<args>]...
```

diefans:
fixes #122


sdispater:
Thanks! Looks good to me 👍

PR #135, opened 2018-05-24
Fix self:update command on Windows.

moigagoo:


poetry/console/commands/self/update.py
```diff
@@ -85,20 +87,23 @@ def update(self, release):
         prefix = sys.prefix
         base_prefix = getattr(sys, 'base_prefix', None)
         real_prefix = getattr(sys, 'real_prefix', None)
+        bin_dir = 'Scripts' if system() == 'Windows' else 'bin'
+        poetry_exe = 'poetry.exe' if system() == 'Windows' else 'poetry'
+        pip_exe = 'pip.exe' if system() == 'Windows' else 'pip'
```
jacebrowning: This might be easier to maintain and test as a small `Venv` wrapper class with these values as properties.

moigagoo: Sorry, I don't quite follow. Are you having something like this in mind:
```
class Venv(object):
    @property
    def bin_dir(self):
        return 'Scripts' if system() == 'Windows' else 'bin'

…

        prefix_poetry = Path(prefix) / Venv().bin_dir / poetry_exe

```
What's the gain of it exactly?

sdispater:
Thanks for taking the time to make this PR. However, this has been fixed (along with other things) in commit 043b9ec99481706f47c0ef47768eea0ed55bc688 on the `master` branch.

So I am closing this but thanks again!

PR #138, opened 2018-05-24
Add Dockerfile

etienne-napoleone:
Hello, 
I thought it would be nice to add a base Dockerfile to poetry. I initially made https://hub.docker.com/r/etiennenapoleone/docker-python-poetry/ but it might be better to include it in the project directly. Let me know if you would do it another way.
Thanks for this great tool!

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,13 +1,14 @@
 FROM python:3.6
 
-# install poetry 
+# install poetry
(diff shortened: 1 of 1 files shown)


etienne-napoleone:
Seems like it's not needed. Also still use deprecated ON BUILD. Discarding this PR

Issue #59, opened 2018-04-19
Poetry fails to install on macOS High Sierra using Homebrew Python 3


lgw4:
Poetry fails to install on macOS High Sierra (10.13.4) using Homebrew Python 3 (3.6.5).
```
$ curl -sSL https://raw.githubusercontent.com/sdispater/poetry/master/get-poetry.py | python3
Retrieving metadata

Installing version: 0.8.5
  - Getting dependencies
An error has occured: Command '('python', '-m', 'pip', 'install', 'poetry==0.8.5', '--target', '/var/folders/zk/1bgll_755dddcqh6snjdsmt00000gn/T/poetry-installer-ra11xs2z/dist')' returned non-zero exit status 1.
b'/usr/bin/python: No module named pip\n'
```
The reason this happens is the `Installer.call()` method has `python` hardcoded as the Python executable name. In my opinion, this is a soft violation of [PEP 394](https://www.python.org/dev/peps/pep-0394/). To solve this problem, call `sys.executable` to obtain the filesystem location of the version of Python running the `get-poetry.py` script.
Pull request [#60](https://github.com/sdispater/poetry/pull/60) has a suggested fix.

lgw4:
Upon further testing, it appears that Homebrew's Python 3 install may be broken in such a way that installing Poetry will never be easy. I still believe the suggested fix is correct for most POSIX systems (that is, it will make sure that Poetry is installed into the `site-packages` of the Python version running `get-poetry.py`.

lgw4:
@bradj I'm not sure what a good solution to this would be (or even if there is one). As far as I can tell, there is some reason the Homebrew Python packagers set both the `home` and `prefix/exec-prefix` values. As you pointed out, the _fix_ on SO really isn't one. If anyone has suggestions, I'd love to hear them.

nchammas:
Chiming in to say that I'm having the same issue as @bradj on macOS with a Homebrewed Python:

```
distutils.errors.DistutilsOptionError: must supply either home or prefix/exec-prefix -- not both
```

I think this particular problem is a dup of #36.

lgw4:
[...]

PR #146, opened 2018-05-27
Fix installation permissions

Victor-Savu:
Avoid having to use superuser (administrator) privileges for platforms where the global pip installation directory is not writable by the regular user.

Se `pip help install | grep -A 4 -e '--user'`:
```
  --user                      Install to the Python user install directory for
                              your platform. Typically ~/.local/, or
                              %APPDATA%\Python on Windows. (See the Python
                              documentation for site.USER_BASE for full
                              details.)
```
Docs for [site.USER_BASE](https://docs.python.org/3.7/library/site.html#site.USER_BASE).

Fixes:
- #59 

Open questions:

- [ ] Should this be optional? If so, should it be opt-in, or opt-out?
- [ ] Should we prompt the user that the user install directory (`site.USER_BASE`) needs to be in the `$PATH` / `%PATH%` variable in order for `poetry` to be recognized as a command?

### DIFF
--- a/get-poetry.py
+++ b/get-poetry.py
@@ -287,6 +287,7 @@ class Installer:
                 "install",
                 "--upgrade",
                 "--no-deps",
+                '--user',
                 os.path.join(dir, "poetry-{}-{}.whl".format(version, tag)),
             )
 
--- a/get-poetry.py
+++ b/get-poetry.py
@@ -287,7 +287,7 @@ class Installer:
                 "install",
                 "--upgrade",
                 "--no-deps",
-                '--user',
+                "--user",
                 os.path.join(dir, "poetry-{}-{}.whl".format(version, tag)),
             )
 


cauebs:
It appears to have worked.
```
~ $  curl -sSL https://raw.githubusercontent.com/Victor-Savu/poetry/patch-1/get-poetry.py | python
Retrieving metadata

Installing version: 0.10.2
  - Getting dependencies
  - Vendorizing dependencies
  - Installing poetry

poetry (0.10.2) successfully installed!
~ $ poetry --version
Poetry 0.10.2
```

jgirardet:
I proposed sudo in https://github.com/sdispater/poetry/pull/163
but I agree --user is better

Victor-Savu:
@jgirardet I am a bit worried by the idea that different packages live in different places, but I have compassion for people who don't have the privilege of `sudo`.

Victor-Savu:
@sdispater Do you have any suggestions or preference regarding the two open questions in this PR?

ranisalt:
The README instructions are broken because of this missing flag, yet it does not seem to be priority.

sdispater:
Closing since with with the new installer permissions error should no longer exist.

PR #148, opened 2018-05-28
Fix setting explicit version number

rainyday:
Fixes 
```
[AttributeError]
'str' object has no attribute 'text'
```
when trying to set version explicitly.


rainyday:
@sdispater sorry to be a nag, but do you have an eta on what version you can include this in? Trying to migrate away from pipenv for a project but setting the version during our builds is important for my use case.

Issue #149, opened 2018-05-28
Pipsi like deployment option


brycearden:
Hey there!

First off, I am a huge fan of `poetry` and can't wait to see how this project changes Python packaging. I would like to start taking advantage of the `pyproject.tml` style of packaging, but I need to be able to deploy my packages on a remote mount, for which I have been using [pipsi](https://github.com/mitsuhiko/pipsi) for that.

It would be awesome for `poetry install` to support installing a package in a local virtual environment and providing a `bin` that has symlinks to the installed entry point script.

cs01:
Maybe `poetry install <pkg> --global`?

cauebs:
Maybe the functionality exposed by the current `install` command could be left to `develop` and `install` could work like Cargo's `install` and `pipsi`

PR #163, opened 2018-05-30
use python3 by default to install poetry

jgirardet:
Python3 should be default.

nobody on linux may execute this whithout sudo due to permission.

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,7 +16,7 @@ from the rest of your system by vendorizing its dependencies. This is the
 recommended way of installing `poetry`.
 
 ```bash
-curl -sSL https://raw.githubusercontent.com/sdispater/poetry/master/get-poetry.py | sudo -H python3
+curl -sSL https://raw.githubusercontent.com/sdispater/poetry/master/get-poetry.py | python3
 ```
 
 Alternatively, you can download the `get-poetry.py` file and execute it separately.


bertjwregeer:
There is a PEP specific to this issue: https://www.python.org/dev/peps/pep-0394/ and it recommends that `python` continue to link to `python2`, whereas `python3` should link to the latest version of Python 3 installed on a system.

This change will hurt users, whereas using `python3` it will do the right thing most of the time.

jgirardet:
Hello, 
maybe @bertjwregeer i'm not sure to all understand you said.
My mind is : everyone is leaving python2 so poetry should recommand python3. maybe my PR is not the good way to do it.
Maybe I'm wrong but I believe that on every system where python3 is system standard, there are both python and python3.

bertjwregeer:
Ah, I misread the PR. Sorry!

README.md
```diff
@@ -16,7 +16,7 @@ from the rest of your system by vendorizing its dependencies. This is the
 recommended way of installing `poetry`.
 
 ```bash
-curl -sSL https://raw.githubusercontent.com/sdispater/poetry/master/get-poetry.py | python
+curl -sSL https://raw.githubusercontent.com/sdispater/poetry/master/get-poetry.py | sudo -H python3
```
osteele: This introduces two changes: `python` -> `python3`; and the introduction of `sudo -H`.

On my machine (macOS with homebrew-installed python@2 and python3), using this with just the `python3` change had the desired affect: it created a python3 virtual environment such that both 
`poetry run which python` and `poetry run which python3` refer to a Python in the appropriatevirtualenv,  and both `poetry run python --version` and `poetry run python3 --version` display `Python 3.6.5`.

So yay for that! (I'd been trying to figure out how to use Poetry for Python 3 development, and was surprised that putting `python = ">=3.0"` in `pyproject.toml` didn't work. This PR saved me.)

I suspect, based on prior experience with `sudo` and homebrew, that if I'd run it with the `sudo`, I would have mixed up some ownerships in  `/usr/local/` in a way that could cause trouble downstream.

Maybe the `sudo -H` part of this change can either simply be removed, or moved to a different PR since if it's necessary at all, its has different compatibility issues and may be solving a different issue.

osteele:
The change from `python` -> `python3` work for me. I'm skeptical of addition of `sudo -H`. I left a longer comment on the line change. (Also, thanks for doing this. Finding this PR got me unblocked and trying out Poetry.)

sdispater:
While the change for `python3` seems good to me, I am not comfortable recommending sudo. Generally, I don't think it's a good idea to install tools like poetry for the root user.

jgirardet:
ok but there still is a problem with the recommanded way to install poetry :
`curl -sSL https://raw.githubusercontent.com/sdispater/poetry/master/get-poetry.py | python3`
will alway fail on  a standard linux distro due to permission :
```
jimmy@jimmyMint ~ $ curl -sSL https://raw.githubusercontent.com/sdispater/poetry/master/get-poetry.py | python3
Retrieving metadata

Installing version: 0.10.3
  - Getting dependencies
  - Vendorizing dependencies
  - Installing poetry
An error has occured: Command '('/usr/bin/python3', '-m', 'pip', 'install', '--upgrade', '--no-deps', '/tmp/poetry-installer-rr8vh6e7/poetry-0.10.3-py2.py3-none-any.whl')' returned non-zero exit status 1
Processing /tmp/poetry-installer-rr8vh6e7/poetry-0.10.3-py2.py3-none-any.whl
Installing collected packages: poetry
Could not install packages due to an EnvironmentError: [Errno 13] Permission non accordée: '/usr/local/lib/python3.5/dist-packages/__pycache__/toml.cpython-35.pyc'
Consider using the `--user` option or check the permissions.
```

cauebs:
I agree with @sdispater and personally prefer the solution proposed by #146

cauebs:
Is there a strong enough reason to recommend Python 3 explicitly? Don't get me wrong, I'm 100% pro-Python3, but Poetry supports 2.7 just as well.

jgirardet:
actually I don't know if this PR is a good thing. it stays complicated to install even with the --user of #146 

@sdispater why not just use pipsi. it works very good with poetry and install is only `pipsi install poetry` and thats it.

cauebs:
What is the problem in installing with `--user`?

For me, if poetry is installed through pipsi, it doesn't work properly. Also, I personally think poetry should provide pipsi's functionality, as suggested in #149.

jgirardet:
I just think that installing an app via a script is something painfull.
What doesn't work by you with pipsi ?

cauebs:
Let's assume that I installed poetry through pipsi. If I create a new project and let poetry create the virtualenv, pip is not linked in `.venv/bin` and every install will fail with "permission denied" because system-wide pip will be used.

jgirardet:
you are right. I did not see it because I only use poetry with pyenv.
But not sure is it pipsi's fault. maybe poetry which should use the pip of the created virtualenv.

cauebs:
That's the thing: the created virtualenv has no pip.

sdispater:
Closing since the new installer is now more flexible.

PR #193, opened 2018-06-08
Document the `--src` flag for `poetry new`

ojii:
`poetry new` supports `--src`, but that is not documented.


sdispater:
Thanks!

Issue #222, opened 2018-06-16
tox support


sobolevn:
Hi, thanks for this awesome package. It look really promising.

Here's my problem: I am using `tox` and `pyproject.toml` for my library.
When trying to use them together that's what happens:

```
» tox
ERROR: No setup.py file found. The expected location is:
  /Users/sobolev/Documents/github/mimesis-factory/setup.py
You can
  1. Create one:
     https://packaging.python.org/tutorials/distributing-packages/#setup-py
  2. Configure tox to avoid running sdist:
     http://tox.readthedocs.io/en/latest/example/general.html#avoiding-expensive-sdist
```

What is the recommended solution for this workflow? 
The only thing that I do not want to do is returning to the old `setup.py`.

rcisterna:
Hi, I'm in the same situation here. It seems like this is a [tox issue](https://github.com/tox-dev/tox/issues/573), which depends on a [pip issue](https://github.com/pypa/pip/issues/5407).

I really hope that all this awesome tools implements **PEP-517** and **PEP-518** soon.

sdispater:
For now, you can use `poetry` with `tox` by using something similar to what I did in `pendulum`: https://github.com/sdispater/pendulum/blob/master/tox.ini

vberlier:
Thanks a heap I was just wondering how `tox` would work without any `setup.py`. Perfect timing!

valignatev:
If we'd think about `setup.py` as some kind of a lock file which you never generate by hand, maybe an option for `poetry lock` which generates `setup.py` might be helpful. Of course, it doesn't work for those who want to avoid storing `setup.py` in their repo. Just brainstorming here.

cs01:
@valignatev this was suggested in https://github.com/sdispater/poetry/issues/191

tadeoos:
I think it would be a good idea to add the pendulum example to FAQ (or start a wiki?)

tadeoos:
I think we can close this issue for now

sobolevn:
Yeap, #227 solves the problem. Thanks everyone!

mlissner:
For those arriving here, the solution mentioned in #227 is this link: https://python-poetry.org/docs/faq/#is-tox-supported

PR #227, opened 2018-06-17
Add info about tox to FAQ section

tadeoos:
This PR adds info about using tox with Poetry as discussed in #222.

There is no contributing guide so I am not sure if this should be merged to `master` or `develop` branch, but since other PRs are merging into master, I followed their lead.

### DIFF
--- a/docs/docs/faq.md
+++ b/docs/docs/faq.md
@@ -32,3 +32,22 @@ with the new major version of your dependency.
 
 For example instead of using `>=3.4` you should use `~3.4` which allows all versions `<4.0`.
 The `^` operator works very well with libraries following [semantic versioning](https://semver.org).
+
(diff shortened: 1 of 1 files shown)


tadeoos:
@sdispater I want to make sure that the info in FAQ is actually enough: I wonder if tox will be able to execute the `poetry` commands in venvs since it is not being installed anywhere explicitly...

sdispater:
As long as you have `poetry` installed on your system, you should be fine. That's why we have to specify `whitelist_externals = poetry` to tell `tox` to execute the `poetry` executable located outside the tox virtualenv.

tadeoos:
Great, thanks!

sdispater:
Looks good to me. Thanks!

PR #268, opened 2018-06-28
WIP: implement namespace packages

uSpike:

I added a `module` key to `pyproject.toml` which allows developers to specify the module path to which the projects code lives.  This should enable namespace packages such as the following:

```
[tool.poetry]
name = "abc-foo"
module = "abc.foo"
```
corresponds with a layout like:
```
pyproject.toml
abc/
    foo/
        __init__.py
```

It wouldn't support a namespace package within the `src` directory, which would be nice.  I don't think that'd be too hard to implement, so I'm leaving this as WIP.


uSpike:
Looks like a bunch of commits were just merged to master that accomplishes what I was trying to do.

PR #275, opened 2018-06-30
Fix broken json schema preventing DirectoryDependency to be added

casio:
Trying to add deps with both `--path` and `--dev` was raising:

`jsonschema.exceptions.ValidationError: {'path': '../some'} is not valid under any of the given schemas`

Seems the schema definitions for dependencies & dev-dependencies have diverged, so this little PR fixes the above error.
However, even with this PR, there's eg the `python` property missing in dev-deps - should both properties rather point an abstract definition, DRY style?

### DIFF
--- a/poetry/json/schemas/poetry-schema.json
+++ b/poetry/json/schemas/poetry-schema.json
@@ -140,6 +140,9 @@
                         },
                         {
                             "$ref": "#/definitions/file-dependency"
+                        },
+                        {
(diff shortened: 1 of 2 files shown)


sdispater:
Tests seem to be broken

casio:
Was a bit busy these last days and you seem to have fixed the issue already in ceb82d6d060a7aa9b63f9f0f8a54125874e250b9, right?

So I'm gonna close this one.

Issue #270, opened 2018-06-29
Does poetry have an equivalent of ` pipenv --rm`?


alok:
Basically a way to uninstall the corresponding virtualenv easily.

cauebs:
Not yet, no. Just playing devil's advocate, how would that be useful?
I've also seen the suggestion of being able to print the virtual environment's path. Maybe a `poetry env` command could be created to handle these?

alok:
It's useful for easily cleaning out deps when you no longer need a copy of them. I often clone a project to modify some code and test out the changes, but afterwards I don't need the virtualenv since I'm finished patching it. I've also cloned multiple copies of a project so I can work on one copy while a long running job is running on the other, and I don't want deps taking up disk space once I'm done.

cauebs:
I implemented it just for fun: https://github.com/cauebs/poetry/commit/1dc35a2d1bedadebcc55c5871278fe8cf01ef6a5

But for this we'll really need @sdispater's input, because I fear we're adding too many commands and duplicating features.

calvertvl:
The other simple solution is an alias or a script; since I pull those down to all systems I use anyway.

I've been using the following:
```
rm -rf "$(dirname "$(dirname "$(poetry run which python)")")"
```
Note that the failure mode (for running outside of a poetry environment) is to run `rm -rf .` but `rm` will refuse to remove the current directory, so it should be relatively safe.

sdispater:
This is supported in the `1.0.0` prereleases. See #731

tombohub:
does it exist now or do we need to write the whole environment name?

Issue #198, opened 2018-06-09
Expose Poetry's virtualenv


ambv:
If the user doesn't have an active virtualenv, Poetry is already creating and using a default per-project virtualenv behind the scenes. There is no easy way to get to it after it's created.

If you'd be willing to accept a PR to this effect, I'd be happy to contribute the following:
* add a `poetry shell` command equivalent to `pipenv shell`
* make `poetry add/remove/install` output the path to the virtualenv that it's using if there is no active virtualenv

I know that your current available alternative is `poetry run`. My problem with it is that it gets verbose fast. `poetry run python -m X` instead of just `X` in an activated virtualenv.

moigagoo:
It sounds reasonable, but I have concerns about Windows compatibility. Pipenv seems to discriminate Windows users and AFAIK `pipenv shell` doesn't work with pwsh or cmd. If you're going to implement `poetry shell`, please don't forget about Windows.

moigagoo:
I'll be happy to contribute then :-)

sdispater:
The output of the virtualenv path is already implemented but only when you use the `add/remove/install` commands in verbose or higher mode (`-v` option).

Regarding the `poetry shell`, if someone to tackle this, I'd gladly review a PR :-)

ethanhs:
FWIW, I'd be happy to test/fix/implement things on Windows. I really like poetry and am keen for this functionality.

jgirardet:
@cauebs look at run.py in commands to starts.

cauebs:
Thanks for the tip, @jgirardet. I gave it a go, it seems to work and I opened #201.

jaroel:
Here's a workaround to get it:
$ poetry run which python
/Users/roel/Library/Caches/pypoetry/virtualenvs/koei-py3.6/bin/python

sobolevn:
I will also gently remind about a possibility to have a local `.venv` folder for dependencies.
`pipenv` has `PIPENV_VENV_IN_PROJECT=true` option which allows that.

sdispater:
@sobolevn This is already supported, see #213

cauebs:
@moigagoo @ethanhs I think #201 is close to being merged, but it only features provisional support for Windows (which hasn't even been tested yet). Would you take a look and let me know how I can improve it? It would probably be important to know how to properly which program to run, if not `cmd`, and which flags are necessary.

moigagoo:
[...]

PR #278, opened 2018-07-05
Add env command

cauebs:
To satisfy the needs expressed in #270 (and, to some extent, #198), I added an `env` command.
```
~/foo $ pt env
Virtualenv path: /home/cauebs/foo/.venv
Python version: 3.6.5
Implementation: CPython

~/foo $ pt env --python
/home/cauebs/foo/.venv/bin/python

~/foo $ pt env --pip
/home/cauebs/foo/.venv/bin/pip

~/foo $ pt env --path
/home/cauebs/foo/.venv

~/foo $ pt env --rm
Removing virtual environment at /home/cauebs/foo/.venv
```

If this gets a greenlight, I'll add tests and docs.

### DIFF
--- a/poetry/console/application.py
+++ b/poetry/console/application.py
@@ -19,6 +19,7 @@ from .commands import BuildCommand
 from .commands import CheckCommand
 from .commands import ConfigCommand
 from .commands import DevelopCommand
+from .commands import EnvCommand
 from .commands import InitCommand
 from .commands import InstallCommand
 from .commands import LockCommand
@@ -111,6 +112,7 @@ class Application(BaseApplication):
             CheckCommand(),
             ConfigCommand(),
             DevelopCommand(),
+            EnvCommand(),
             InitCommand(),
             InstallCommand(),
             LockCommand(),
--- a/poetry/console/commands/__init__.py
+++ b/poetry/console/commands/__init__.py
@@ -4,6 +4,7 @@ from .build import BuildCommand
 from .check import CheckCommand
 from .config import ConfigCommand
 from .develop import DevelopCommand
+from .env import EnvCommand
 from .init import InitCommand
 from .install import InstallCommand
 from .lock import LockCommand
--- a/poetry/console/commands/env.py
+++ b/poetry/console/commands/env.py
@@ -10,7 +10,7 @@ class EnvCommand(VenvCommand):
         { --p|path : Show the path to virtual environment. }
         { --python : Show the path to the python executable. }
         { --pip : Show the path to the pip executable. }
-        { --r|remove : Remove the virtual environment. }
+        { --rm : Remove the virtual environment. }
     """
 
     help = """The env command displays information about the virtual environment.
@@ -28,7 +28,7 @@ If it doesn't exist when the command is ran, it will be created."""
         elif self.option("pip"):
             self.info(self.venv.pip)
 
-        elif self.option("remove"):
+        elif self.option("rm"):
             self.line("Removing virtual environment at <info>{}</>".format(path))
             rmtree(path, ignore_errors=True)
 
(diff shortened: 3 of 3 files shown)


sdispater:
Thanks for taking the time to put this PR together.

Some information are already present in the "debug:info" command. So I'm not sure about having duplicate commands.

Also, I am not a big fan of options executing actions. I know this is the `pipenv` way but I find it quite counter-intuitive. I prefer having command namespace (like the `debug` one). That way we could have a `env:remove` command.

Finally, I am still on the fence here about exposing the virtualenv since, to me, this is an implementation detail and, ideally, I would like Poetry to not even use them and use a more clever approach. I am not saying it will come anytime soon but introducing a `env` command will tie up Poetry to the virtualenv way.

Anyway, I am leaving this PR open for now since I still need to wrap my head around it and to continue the discussion.

cauebs:
That's reasonable enough. Don't feel bad about closing this PR, it was just an experiment. But I do think #270 requires some action, and that feature can be implemented in another command just as easily.

What alternative to virtual environments do you have in mind? I don't know any other approaches.
I just feel the need to remind you that the Python community is extremely diverse, and in trying to be inventive we can end up creating even more problems in the long run.

golyalpha:
Could poetry maybe output just the path to the venv on `poetry env`? It's what `pipenv` does if you give it the `--venv` switch.

yunti:
@golyalpha Perhaps the command `poetry config settings.virtualenvs.in-project true` could be broadened to take an input so the venv location could be specified if needed. This would help cover a lot of scenarios and aid adoption as we wouldn't have to wait for official support for IDEs etc...

I agree with @sdispater though, it would be ideal if we generally didn't have to think about a venv. I think poetry has already done the hard parts by being able to detect them and setup if required. The main other requirements are to view the location and remove them.  

@golyalpha definitely has a point in that `debug:info` isn't the most intuitive command for the venv location, but I would be hesitant to add a full `env` command if we are trying to abstract that away.
I'm not sure of the best place/command for it though. (show?)

As for removing a venv I would have it has a flag to `remove` so it can be used to remove packages, and the virtualenv too. Something like `remove --all` would be clear without having to think in terms of a venv.

funkyfuture:
beside the fact that'd i'd prefer `poetry env rm`, i'd find that an useful interface in automated deployment tasks.

golyalpha:
@yunti It's great that Poetry is trying to abstract the env away from the developer, but that is not exactly something that IDEs like. IDEs like knowing where the interpreter itself is, not how they can get to it through the usage of and external tool (in poetry's case: `poetry run python`), they like being able to quickly run the interpreter directly, which is also admittedly faster, than going through poetry, while producing the same results.
Of course, letting the IDE add project dependencies through anything other than Poetry is a big no-no, but we already know that that isn't really an issue with VSCode and Pipenv.  

TL;DR:
What I'm trying to say is, that IDEs should be able to easily access the interpreter directly, since there is no reason to prevent them from doing so and it's faster, and while it is hard (read: near impossible) to prevent an IDE from doing `python -m pip install some-module`, I think that IDEs and their plugin developers understand that if a project is managed by some dependency management tool, it's a good idea to *not* bypass it.

Note:
Having an env command could allow project owners and it's contributors to define path to python as: `$(poetry env)\Scripts\python` (if bash, but it's quite similar on powershell), entirely bypassing the need for in-project venvs (where their usage is developer, not project, dependent) or having a plugin that supports Poetry when it comes to just executing the code, and providing autocomplete (of course, it's still on the dev to use poetry to install and manage deps, until IDEs/Python support plugins finally come with support for Poetry).

sdispater:
Thanks everyone for the insight and @cauebs for taking the time to make this PR!

I am closing this PR since it is superseded by #731. If you have further feedback to make please do it on #731 instead.

cauebs:
@sdispater It was more of a proof of concept. I'm quite happy with the design you ended up implementing. Great job, as usual ;)

PR #296, opened 2018-07-09
point to the run command in the script command's help

glehmann:
don't let the user wonder what command to use - just point it to him


sdispater:
Thanks!

PR #298, opened 2018-07-10
Fix typo in docs

radek-sprta:
Seems like there is typo in one of the examples. After eliminating the extra colon, I was able to build and install the package correctly.


sdispater:
Thanks!

Issue #177, opened 2018-06-05
Add package from git not working on windows 10


shawegit:
Adding a package from a git repository on windows 10 does not work. The package is downloaded from the git repository as required without any issue. However, at some point some files in pypoetry-git-package_nameSOME_VALUE\\.git\\objects\\pack\\ are tried to be deletethat are write protected. Below is a complete example stack trace showing the errors when trying to execute the command 
poetry -v add pendulum --git https://github.com/sdispater/pendulum.git

A solution to this might be
```python
import stat
def on_rm_error(func, path, exc_info):
    os.chmod(path, stat.S_IWRITE)
    os.unlink(path)

shutil.rmtree(tmp_dir.as_posix(), onerror=on_rm_error)
```

[PermissionError]
[WinError 5] Access is denied: 'C:/Users/Public/Temp/pypoetry-git-pendulumaj5p2d8l\\.git\\objects\\pack\\pack-37344d1961bffa13deda99e13c97ecda4c0c0013.idx'

Exception trace:
 c:\program files (x86)\python36-32\lib\site-packages\cleo\application.py in run() at line 94
   status_code = self.do_run(input_, output_)
 c:\program files (x86)\python36-32\lib\site-packages\poetry\console\application.py in do_run() at line 87
   return super(Application, self).do_run(i, o)
 c:\program files (x86)\python36-32\lib\site-packages\cleo\application.py in do_run() at line 197
   status_code = command.run(input_, output_)
 c:\program files (x86)\python36-32\lib\site-packages\poetry\console\commands\command.py in run() at line 72
   return super(BaseCommand, self).run(i, o)
 c:\program files (x86)\python36-32\lib\site-packages\cleo\commands\base_command.py in run() at line 146
   status_code = self.execute(input_, output_)
 c:\program files (x86)\python36-32\lib\site-packages\cleo\commands\command.py in execute() at line 107
   return self.handle()
 c:\program files (x86)\python36-32\lib\site-packages\poetry\console\commands\add.py in handle() at line 134
   status = installer.run()
 c:\program files (x86)\python36-32\lib\site-packages\poetry\installation\installer.py in run() at line 75
   self._do_install(local_repo)
 c:\program files (x86)\python36-32\lib\site-packages\poetry\installation\installer.py in _do_install() at line 158
   ops = solver.solve(use_latest=self._whitelist)
 c:\program files (x86)\python36-32\lib\site-packages\poetry\puzzle\solver.py in solve() at line 35
[...]

PR #300, opened 2018-07-11
Fix bug that prevents installing package from git on windows

shawegit:
This PR fixes an issues that prevents windows user to install packages from git repositories using poetry, Currently git on windows downloads something into a folder that is write protected. When poetry delets the downloaded stuff for cleanup, remove thig specific folder raises an exception see issue #177.


ptink:
Would be extremely helpful to have this merged in as it's still an issue on 0.11.3

poetry/puzzle/provider.py
```diff
@@ -10,6 +10,10 @@
 from tempfile import mkdtemp
 from typing import List
 
+from typing import Union
```
ptink: unused import

shawegit:
@sdispater Thanks for fixing it in 0.12.8

Issue #233, opened 2018-06-19
Can't use private repository in `poetry add`


ojii:
Poetry version: 0.10.3
Python Version: 3.6.5

(Note: I redacted URL of my private repository)

pyproject.toml:

```toml
[tool.poetry]
name = "project"
version = "0.1.0"
description = ""
authors = []

[[tool.poetry.source]]
name = "private"
url = "<url>"

[tool.poetry.dependencies]
python = "^3.6"

[tool.poetry.dev-dependencies]
pytest = "^3.5"
```

I ran the following two commands:

* `poetry config repositories.private <url>`
* `poetry config http-basic.private <username>`

Checking the config files on disk, the values are correct.

When I run `poetry search <private package>` or `poetry add <private package>` poetry cannot find the package. Here's the output of `poetry -vvv add`:

```
poetry -vvv add <private package>
                                  
[ValueError]       
Package [<private package>] not found.  
                                  
Exception trace:
 /usr/local/lib/python3.6/site-packages/poetry/_vendor/cleo/application.py in run() at line 94
   status_code = self.do_run(input_, output_)
 /usr/local/lib/python3.6/site-packages/poetry/console/application.py in do_run() at line 87
   return super(Application, self).do_run(i, o)
 /usr/local/lib/python3.6/site-packages/poetry/_vendor/cleo/application.py in do_run() at line 197
   status_code = command.run(input_, output_)
 /usr/local/lib/python3.6/site-packages/poetry/console/commands/command.py in run() at line 72
   return super(BaseCommand, self).run(i, o)
 /usr/local/lib/python3.6/site-packages/poetry/_vendor/cleo/commands/base_command.py in run() at line 146
   status_code = self.execute(input_, output_)
 /usr/local/lib/python3.6/site-packages/poetry/_vendor/cleo/commands/command.py in execute() at line 107
   return self.handle()
 /usr/local/lib/python3.6/site-packages/poetry/console/commands/add.py in handle() at line 65
   packages, allow_prereleases=self.option("allow-prereleases")
 /usr/local/lib/python3.6/site-packages/poetry/console/commands/init.py in _determine_requirements() at line 215
   requirement["name"], allow_prereleases=allow_prereleases
 /usr/local/lib/python3.6/site-packages/poetry/console/commands/init.py in _find_best_version_for_package() at line 245
   name, required_version, allow_prereleases=allow_prereleases
[...]

PR #306, opened 2018-07-14
Adding support for HTTP Basic Auth when installing and searching from legacy repo (#233)

MarcDufresne:
# Pull Request Check List

Hi! I really like `poetry` and would love to be able to use it at work, sadly it doesn't support HTTP Basic Auth when pulling from private repositories so that has kept us from using it. See #233 

That is why I have come up with this PR. It basically adds support for pulling from private repos using the auth values stored in `http-basic.<repo-name>` by inserting them in the repo URL dynamically. I've also made sure that the username and password are never output to the shell.

While doing this I tried to have the most minimal impact on existing user. For example, anyone that already defines username and password directly in the URL of the `pyproject.toml` file won't suddenly have issues since I check for existing credentials and skip my code if anything is found (But I suggest those who do this now use the `http-basic` configuration). Additionally, I make sure that if no credentials are found for this specific repo the URL won't be altered at all.

~~I did not write tests yet because I wanted my design validated first and because I am not sure how to approach tests for this feature.~~
I wrote a unit test that mocks reading `http-basic` from the config and uses the returned values to create a new URL and then validates that the URL was properly constructed.
I have also tested the functionality locally using a devpi-server behind a HTTP Basic Auth protected web server, with the config below

`pyproject.toml`:
```toml
[[tool.poetry.source]]
url = "http://localhost:3142/root/pypi/+simple/"
name = "devpi"
```

```bash
poetry config http-basic devpi user pass
```

### DIFF
--- a/poetry/installation/pip_installer.py
+++ b/poetry/installation/pip_installer.py
@@ -25,9 +25,14 @@ class PipInstaller(BaseInstaller):
         if package.source_type == "legacy" and package.source_url:
             parsed = urlparse.urlparse(package.source_url)
             if parsed.scheme == "http":
+                if not parsed.username:
+                    from_host = parsed.netloc
+                else:
+                    from_host = parsed.netloc[parsed.netloc.index("@") + 1 :]
+
                 self._io.write_error(
                     "    <warning>Installing from unsecure host: {}</warning>".format(
-                        parsed.netloc
+                        from_host
                     )
                 )
                 args += ["--trusted-host", parsed.netloc]
--- a/poetry/installation/pip_installer.py
+++ b/poetry/installation/pip_installer.py
@@ -3,6 +3,9 @@ import tempfile
 
 from subprocess import CalledProcessError
 
+from poetry.utils.helpers import get_http_basic_auth
+
+
 try:
     import urllib.parse as urlparse
 except ImportError:
@@ -25,19 +28,27 @@ class PipInstaller(BaseInstaller):
         if package.source_type == "legacy" and package.source_url:
             parsed = urlparse.urlparse(package.source_url)
             if parsed.scheme == "http":
-                if not parsed.username:
-                    from_host = parsed.netloc
-                else:
-                    from_host = parsed.netloc[parsed.netloc.index("@") + 1 :]
-
                 self._io.write_error(
                     "    <warning>Installing from unsecure host: {}</warning>".format(
-                        from_host
+                        parsed.netloc
                     )
                 )
                 args += ["--trusted-host", parsed.netloc]
 
-            args += ["--index-url", package.source_url]
+            # This here won't work, need some pointers
+            auth = get_http_basic_auth(package.source_url)
+            if auth:
+                index_url = "{scheme}://{username}:{password}@{netloc}{path}".format(
+                    scheme=parsed.scheme,
+                    username=auth[0],
+                    password=auth[1],
+                    netloc=parsed.netloc,
+                    path=parsed.path,
+                )
+            else:
+                index_url = package.source_url
+
+            args += ["--index-url", index_url]
 
         if update:
             args.append("-U")
--- a/poetry/installation/pip_installer.py
+++ b/poetry/installation/pip_installer.py
@@ -35,8 +35,7 @@ class PipInstaller(BaseInstaller):
                 )
                 args += ["--trusted-host", parsed.netloc]
 
-            # This here won't work, need some pointers
-            auth = get_http_basic_auth(package.source_url)
+            auth = get_http_basic_auth(package.source_reference)
             if auth:
                 index_url = "{scheme}://{username}:{password}@{netloc}{path}".format(
                     scheme=parsed.scheme,
--- a/poetry/repositories/legacy_repository.py
+++ b/poetry/repositories/legacy_repository.py
@@ -28,6 +28,7 @@ from cachy import CacheManager
 
 import poetry.packages
 
+from poetry.config import Config
 from poetry.locations import CACHE_DIR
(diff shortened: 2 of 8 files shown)


MarcDufresne:
@sdispater @cauebs This is ready for review anytime you have time! Thanks!

poetry/repositories/legacy_repository.py
```diff
@@ -159,6 +160,15 @@ def __init__(self, name, url, disable_cache=False):
             requests.session(), cache=FileCache(str(self._cache_dir / "_http"))
         )
 
+        url_parts = urlparse.urlparse(self._url)
+        if not url_parts.username:
+            username, password = get_http_basic_auth(self.name)
+            if username and password:
+                netloc = "{}:{}@{}".format(username, password, url_parts.netloc)
```
sdispater: Instead of recreating the url by hand we can just pass the auth information to the session created in `__init__()`. That way the `_url` attribute is left untouched while still providing the auth information to each request made to the remote url.

```python
self._session.auth = get_http_basic_auth(self.name)
```
For this to work we have to make a small change to `get_http_basic_auth()` so that it returns `None` when no authentication information could be found. This makes the code much simpler since we don't have to chech each time if username/password exist. 
```

File: poetry/utils/helpers.py [hunk 2]
```diff
@@ -77,3 +78,11 @@ def parse_requires(requires):  # type: (str) -> Union[list, None]
...
     if requires_dist:
         return requires_dist
+
+
+def get_http_basic_auth(repository_name):  # type: (str) -> (str, str)
+    config = Config.create("auth.toml")
+    repo_auth = config.setting("http-basic.{}".format(repository_name))
+    if repo_auth:
+        return repo_auth["username"], repo_auth["password"]
+    return "", ""
```
sdispater: We should return `None` for the reasons specified above.

pyproject.toml
```diff
@@ -55,6 +55,7 @@ pygments-github-lexers = "^0.0.5"
 black = { version = "^18.3-alpha.0", python = "^3.6" }
 pre-commit = "^1.10"
 tox = "^3.0"
+mock = "^2.0"
```
sdispater: This is not necessary since we already have `pytest-mock` which provides a more convenient way to mock objects and functions.

tests/repositories/test_legacy_repository.py
```diff
@@ -43,3 +44,11 @@ def test_page_absolute_links_path_are_correct():
     for link in page.links:
         assert link.netloc == "files.pythonhosted.org"
         assert link.path.startswith("/packages/")
+
+
+@mock.patch("poetry.repositories.legacy_repository.get_http_basic_auth")
```
sdispater: We shoud use the `mocker` fixture provided by `pytest-mock`

tests/repositories/test_legacy_repository.py
```diff
@@ -43,3 +44,11 @@ def test_page_absolute_links_path_are_correct():
...
+def test_http_basic_auth_repo(mock_get_auth):
```
sdispater: Testing this will be more complex with the changes I requested since we will need to check the url that will be called in `_get()`.

poetry/installation/pip_installer.py
```diff
@@ -25,9 +25,14 @@ def install(self, package, update=False):
         if package.source_type == "legacy" and package.source_url:
             parsed = urlparse.urlparse(package.source_url)
             if parsed.scheme == "http":
+                if not parsed.username:
```
sdispater: With the changes requested below. The authentication part of the url will not appear here, so this check is not necessary.

However, this will be an issue since `pip` won't be aware of the credentials. So this is where we should modify the url to include the credentials, I think.

MarcDufresne:
Thanks for the feedback, it's your project so we'll do it your way :P

For the session, setting the `auth` was actually my first implementation but wouldn't work with the `_download` method in `PyPiRepository` since it doesn't use the session (and I tried using the session in it but downloads would timeout for some reason), this is the cleanest way I've found that worked without changing too much of the code/design. Let me know what you think.

I've actually been using my fork since last week and it's working great at least.

sdispater:
@MarcDufresne It's weird that the downloads timeout by using the session. Could this be caused by the fact that we encapsulate the sessions in a `CacheControl` object?

sdispater:
I just checked and I can't reproduce the timeouts. Everything works on my end by using the session in the `LegacyRepository` class.

MarcDufresne:
@sdispater Thanks for trying it on your end, can you share a snippet? IIRC I had to make `PyPIRepository._download` use `self.session.get`, that's the part that somehow didn't work for me.

Oh and I should be fixing the comments shortly!

sdispater:
@MarcDufresne I just overrode `_download()` in `LegacyRepository` to use `self._session.get()`, nothing more and it worked.

MarcDufresne:
Ok cool, I'll do it like that then, I didn't want to change too much code at first, but if you're okay with it, I'll do it like that.

MarcDufresne:
@sdispater Can you take a look at `pip_installer.py`, I'm looking for a way to get the name of the repo (to get the auth) but I'm not sure how.

sdispater:
@MarcDufresne At the moment there is no easy way to retrieve the name of the repository from the lock file.

I think we can store it in the `source_reference` attribute of a `Package` since it's currently not used for `LegacyRepository` objects.

See here: https://github.com/sdispater/poetry/blob/master/poetry/repositories/legacy_repository.py#L239

sdispater:
Thanks a lot for this!

MarcDufresne:
Well thanks for your feedback and accepting this! Glad to see poetry becoming even more useful :D

PR #319, opened 2018-07-19
Add pipsi to installation methods

### LINKED ISSUE #318: Add a note discussing installing poetry with pipsi
As a Python dev, installing tools like `poetry` with `pipsi` makes for a very convenient experience. However, the interaction between the currently available `pipsi` on PyPI and how `poetry` makes broken virtualenvs. This has been noted in a couple of `poetry` GitHub issues already.

A note in the `poetry` documentation, that describes installing via `pipsi`, and how to handle the resulting broken envs, would be helpful for users wanting to install that way. While partially off topic (since the problem really isn't with `poetry`) it adds real value to the documentation.

See also:

* https://github.com/sdispater/poetry/issues/59
* https://github.com/sdispater/poetry/issues/169

cauebs:
Closes #318.

### DIFF
--- a/README.md
+++ b/README.md
@@ -42,6 +42,16 @@ pip install --user poetry
 Be aware, however, that it will also install poetry's dependencies
 which might cause conflicts.
 
+Using [pipsi](https://github.com/mitsuhiko/pipsi) to install `poetry` is also possible.
+
+```bash
+pipsi install poetry
+```
+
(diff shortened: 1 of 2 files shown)


funkyfuture:
This seems solved in e29810ef472a07ae465f0ebc1f0760ad009724c5 to me.

Issue #305, opened 2018-07-14
Request for documention on how to use poetry with multiple python versions


SethMMorton:
- [X] I have searched the [issues](https://github.com/sdispater/poetry/issues) of this repo and believe that this is not a duplicate.

## Issue

In order to frame my request, let me describe how I anticipate some users come to use poetry (this at least was the case for me):

  1. A developer realizes they need a more integrated dependency management solution than just setuptools, and discovers pipenv because it is the top hit on google.
 2. They do a bit of reading, and decide that pipenv must be good because it is For Humans&trade; and it is (or at least was) The Officially Recommended Python Packaging Tool from Python.org. They try it out.
 3. The developer inevitably attempts to do something outside the scope of what pipenv is trying to solve and then discover a) that pipenv is *really* only intended for application developers, and if one is a library developer you still need to use `setup.py` (making them feel like a second-class citizen), and/or b) the responses to issues filed for pipenv are usually either rude or dismissive.
 4. They look for an alternative, and find poetry. There is no rudeness in the issues! The tool actually tries to help *all* developers! Huzzah!

For people like myself that came to poetry by way of pipenv, the approach to managing virtual environments taken by poetry tripped me up for quite some time because it differs significantly from pipenv.

 - In pipenv, one specifies the python version to use for virtual environment creation, and it searches the PATH and creates it for you if a match is found.
 - In poetry, it creates the virtual environment from the python against which poetry was installed, or uses a pre-made `.venv`.

This is a fundamental difference in philosophy between the tools which leads to a fundamental difference in usage on this specific aspect of project/dependency management. Because I had been first exposed to the pipenv way, I (wrongly) assumed that poetry also worked this way and struggled to get poetry to use a python 3.5 virtual environment because I had installed poetry against python 3.6; this was especially confusing because I had set the python version in the `pyproject.toml` to 3.5. I did notice the [Poetry and virtualenvs](https://poetry.eustace.io/docs/basic-usage/#poetry-and-virtualenvs) section, but for someone who has not yet grokked the design philosophy of poetry there is not enough information for someone used to what pipenv does to adapt to poetry's approach. 

### REQUEST (finally)

I think the documentation for poetry would be improved if one (or both) of the following were added to documentation:

[...]

PR #332, opened 2018-07-24
Update documentation on migrating and using different python versions (#305)

festinuz:
This pull request addresses #305.

Note that its something of a draft and could use some polishing. Still, it feels like its better than nothing :)

### DIFF
--- a/docs/docs/basic-usage.md
+++ b/docs/docs/basic-usage.md
@@ -163,3 +163,20 @@ or create a brand new one for you to always work isolated from your global Pytho
     What this means is if you project is Python 2.7 only you should
     install `poetry` for your global Python 2.7 executable and use
     it to manage your project.
+
+
+## Using a specific python version
+
+A recommended approach to set a specific python version for your project is to use [pyenv](https://github.com/pyenv/pyenv).
+
+Here are the steps you would take to achieve that:
+1. [Install pyenv](https://github.com/pyenv/pyenv#installation) to your system
+2. Install a specific python version:
+```
+pyenv install 3.7.0
+```
+3. Install poetry for this python version
+```
+pyenv global 3.7.0 && curl -sSL https://raw.githubusercontent.com/sdispater/poetry/master/get-poetry.py | python
+```
+After that you can follow the steps described in [basic usage documentation](/basic-usage/).
--- a/docs/docs/migrating.md
+++ b/docs/docs/migrating.md
@@ -2,13 +2,13 @@
 
 This section covers how to migrate from other common worflows to using poetry.
 
-#### `virtualenv/pip`
+## virtualenv/pip
 If you used plain `virtualenv` in your project before, that probably means you have a `.venv` folder at the root of your project. In that case, all you need to do is to install requirements from your `requirements.txt` file with poetry. Poetry will use the `.venv` folder of your project.
 
-#### `virtualenvwrapper/pip`
+## virtualenvwrapper/pip
 If you used `virtualenvwrapper` then the path is the same as with `virtualenv` with an extra step of locating the folder in which your virtual environment is stored. An alternate route would be to (if needed) set a specific python version with pyenv and create a new virtual environment [with poetry](/basic-usage/).
 
-#### `pipenv`
+## pipenv
 If you're coming from `pipenv`, then `poetry` will strike you as something similar but different.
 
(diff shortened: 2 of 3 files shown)


SethMMorton:
I think that a short blurb on what `pyenv` is and why it is recommended might be useful. I also think it would be very useful to demonstrate using `poetry` and `pyenv` to develop against multiple python versions, and switch between the two.  I have seen it mentioned in multiple issues that one *can* do this, but I have not yet seen any steps on how to do this.

jgirardet:
to give an idea:
I create a virtualenv "myvenv" my pyenv to let me choose the python version.
For example `pyenv virtualenv  3.7.0 myenv`.
I put in `.python-version` :  `myvenv 3.5.5 3.6.5 3.7.0`.

tox is configured like for poetry's repo  with `envlist py35, py36,py37`

So when I come inside the directory, 'myvenv" is auto activated which is cool to launch some commands easily.
I run `tox` to test under 3.5 3.6 and 3.7

If you do not want to use tox, you can create custom venv with pyenv with preferred python version and switch between them with `pyenv local myvenv`

festinuz:
Thanks for suggestions, i'll try to come up with an update tomorrow. 

By the way if any of you feel like it, i would really love some contributions, as writing an easy-to-read documentation is not my strong suit.

jgirardet:
I formalized it : https://github.com/jgirardet/poetry/blob/master/docs/docs/basic-usage.md
But with 'my way' you do not need to install poetry for each python version. It's confusing for people if it's not the same I think (I personnaly only use the pipsi method)

festinuz:
Oh well, it seems like we have different approaches in mind.

I myself just put the version @sdispater showed in #305. In his version you leave management of virtualenv to poetry, while in your you create virtualenvs manually.

Both approaches are valid, but i think that only one should be in the docs to avoid confusion.

 .                                               | sdispater | jgirardet 
-----------------------------------|-------------|-------------
allows to use muptiple versions | yes | yes
multiple poetry installs | yes | no
create virtualenv manually | no | yes
needs `.python-version` file | yes* | yes

> \* WIth sdispater approach, what happens if you come from project A (python 2.7) to project B folder (python 3.6) and start to use poetry when:
> A) Virtualenv is already created by poetry for this project (python 3.6)
> B) Virtualenv is not created for this project
> Will it use python 2.7 in both? Will it use 3.6 in one, and 2.7 in other?

Thoughts?

jgirardet:
> Since you create your virtualenvs manually, if you share your project across multiple PCs you will run into a following problem: there will be no match for your python version set in .python-version file on any other system - That virtualenv you originally created could have any python versio

### about creating virtuelenv
In every project  you fork on github, the python version used by maintainer is never "included" with. You usually use your system python or what you (installed via deb or tgz file...). The maintainer only say to you : "my package works with python 3.5 and3.6 only". It's up to you to have it on your system.
So when someone clone/fork your repo he will use poetry with poetry's python version or a custom venv he created by anyway he wants : virtualenv, python -m venv, pyenv...

### about testing
Where I agree with you is concerning the testing part. since the user wants to run tests with tox but can't have every supported python version, tox.ini should have the `skip_missing_interpreters=True` so only available interpreters will be used.

### about .python-version:
I think this should not  be put in vcs. It's up to the user to choose how he deals with the different python version. As an example : If use use pyenv, i'll create locally a .python-version my my desired pyenv version and use `tox`. but If I already have many python-version on my system not installed via pyenv, I  do not need to add a .`python-version`

If you want to use pyenv without virtualenv with multiple poetry install  you still have to use a the command `pyenv local 3.7.0` and it will create a .python-version with version in it and then you'll be able to use to use poetry's specific python version.

### about multiple poetry install
Since you have to update each python version, for each new poetry update, I think i'ts really cumbersome.

### about choosing a default way to install or to do things
After reflexion, if documentation is clear : "you have this way with those benefits or counterpart and this way with this benefits an counter parts" I think both could live together.

end of the flood :wink:

festinuz:
Good point on upgrading poetry, that really is a "killer" in my eyes. That leaves me wondering if there is another way to manage it, as both suggested ways look sub-optimal.

sdispater:
I am closing this PR as a lot has happened since then. The installer no longer requires to be installed for each python version and we will have #731 when version `1.0.0` lands.

Issue #337, opened 2018-07-26
Support for .env files


ptink:
- [x] I have searched the [issues](https://github.com/sdispater/poetry/issues) of this repo and believe that this is not a duplicate.

## Issue

First off I really like poetry- I'd actually only recently migrated over to pipenv but I already see the advantages of poetry and would ideally like to use it full time.

I primarily work with Django and other web frameworks that can require a large amount of configuration which- when following 12 factor app practices- involves lots of environment variables that necessitates the use of a `.env` file.

Is there a possibility of adding in `.env` file support, specifically for poetry's `run` command so that the resulting command runs both in the virtualenv and with the environment as configured in the `.env` file?

sdispater:
I don't think this is in the scope of Poetry.

Since Poetry is not configured via environment variables there are no reasons for it to load a `.env` file.

If the libraries you use need it, it's their responsability to load it not Poetry's.

jacebrowning:
I've always liked using [direnv](https://direnv.net/) for this. It supports loading `.env` files [automatically](https://github.com/direnv/direnv/issues/284).

ptink:
@jacebrowning direnv isn't cross platform unfortunately. When the environments I worked in were solely on Linux I used direnv in conjunction with [vex](https://pypi.org/project/vex/) to achieve functionality similar to `pipenv run <command>` for example. This workflow doesn't work on Windows which was a major reason for my discovery and subsequent switch to pipenv when I realised it could do the same thing.

I agree that poetry doesn't need .env files for it's own configuration (although it's not a bad idea to use them to allow for directory-specific configuration that may differ from the globals defined in the config file) but I don't think it's out of scope for poetry to load `.env` files for `run` and `shell` when you consider that they are both attempts to isolate the given command in an environment that is specific to the current project. 

I understand the concerns about scope creep, especially when the PR adds a new dependency. If you're exclusively writing libraries then this particular use case may not occur very often but for 12 factor apps it's pretty vital to have tooling to support it- and it always seemed like a neat solution to me to have the virtualenvironment be in control of it.

AlJohri:
I've used https://github.com/jezdez/envdir/ for something similar.

it uses an `envs` folder instead of a `.env` file.

https://github.com/toddbluhm/env-cmd is also good if you want to stick to the `.env` file

it should be simple to use an alias that chains the two commands together

ptink:
[...]

PR #339, opened 2018-07-26
Adding automatic loading of .env file

MarcDufresne:
# Pull Request Check List

-----

This is in support of #337, it doesn't include tests but I didn't find any for the `run` and `shell` command. I'll defer to your decision on if I should add tests, what kind of test, or if it can wait.

~~I feel like eventually a config value to change the name of the file to load instead of hardcoding `.env` could be nice but may not be needed now.~~


poetry/utils/helpers.py
```diff
@@ -77,3 +81,10 @@ def parse_requires(requires):  # type: (str) -> Union[list, None]
 
     if requires_dist:
         return requires_dist
+
+
+def load_dotenv_if_exists():
+    workdir_path = Path(os.getcwd())
+    env_file_path = workdir_path / ".env"
+    if env_file_path.is_file():
+        load_dotenv(env_file_path)
```
ptink: Might be overthinking it but if we did something like

```
read_dotenv = dotenv.get_key(env_file_path, 'POETRY_READ_DOTENV')
if read_dotenv and read_dotenv.lower() not in ('false', 'n', 'no', 'off', '0'):
    load_dotenv(env_file_path)
```

This gives users a way to turn off reading the .env file on a per-directory basis even if we have a global setting that turns it on?

poetry/utils/helpers.py
```diff
@@ -77,3 +81,10 @@ def parse_requires(requires):  # type: (str) -> Union[list, None]
 
     if requires_dist:
         return requires_dist
+
+
+def load_dotenv_if_exists():
+    workdir_path = Path(os.getcwd())
```
ptink: I think it's worth adding a `settings.commands.dotenv` (or better named) config setting here- as you suggested it would let people set it to an alternative filename (i.e. `poetry.env`). setting it to empty / false could turn it off globally?

ptink:
lgtm 👍

pkulev:
Maybe rebase is better then merging develop? It makes history much cleaner.

```shell
$ git remote update
$ git rebase -i upstream/develop
```

lorinkoz:
Hey, what's the status of this?

sdispater:
Closing this since I have mentioned in #337 that this is beyond the scope of Poetry

PR #361, opened 2018-07-31
Add option to override project directory.

rainyday:
# Pull Request Check List

Being able to set the project directory at the command line is extremely useful for using automated tools and running in repositories that contain multiple projects (which are an unfortunate reality in a lot of environments).

Adding this also required making `run` compatible with pre-command options.

This implementation is far from ideal but a clean way of doing this will necessitate changes to `cleo` from what I can tell. Mainly looking for input/feedback.

### DIFF
--- a/poetry/console/application.py
+++ b/poetry/console/application.py
@@ -8,6 +8,7 @@ from cleo.formatters import Formatter
 from cleo.inputs import ArgvInput
 from cleo.outputs import ConsoleOutput
 from cleo.outputs import Output
+from cleo.inputs import InputOption
 
 from poetry import __version__
 
@@ -47,8 +48,14 @@ class Application(BaseApplication):
 
         self._poetry = None
         self._skip_io_configuration = False
+        self._project_dir = None
         self._formatter = Formatter(True)
         self._formatter.add_style("error", "red", options=["bold"])
+        self._definition.add_option(
+            InputOption(
+                "--project", "", InputOption.VALUE_REQUIRED, "Path to project directory"
+            )
+        )
 
     @property
     def poetry(self):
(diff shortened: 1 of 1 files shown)


mgeisler:
Yay, this option would be useful for me!

I use a little wrapper script to make it easy for people in my team to run Python tools packaged with Poetry. The scripts looks like this:
```
#!/usr/bin/env bash

set -e

# Helper for starting a Python script with dependencies managed by
# Poetry. Use it in the shebang line of your Python script:
#
#   #!/usr/bin/env runpoetry
#
# When invoked, runpoetry will make sure to call 'poetry install' to
# install dependencies from your pyproject.lock file.

# The realpath command doesn't exist on OS X, so here is a portable
# replacement.
realpath() {
    local path="$1"
    if [[ -h "$path" ]]; then
        local target=$(readlink "$path")
        if [[ $target[0] =~ '^/' ]]; then
            echo "$target"
        else
            echo "$(dirname "$path")/$target"
        fi
    else
        echo "$path"
    fi
}

# Poetry looks for the pyproject.toml file the current directory, so
# we need to change to the directory of the script being invoked
# (passed in $1). Using realpath allows $1 to be a symlink to the real
# Python script.
OLD_PWD=$PWD
cd "$(dirname "$(realpath "$1")")"

# We can now install the dependencies in the virtualenv Poetry
# maintains.
poetry install --quiet --no-interaction --no-dev

# We now invoke the script and take care to change back to the
# original working directory first. We pass the target directory and
# any positional arguments as positional arguments to the inner Bash
# process. That way it can take care of passing them through unchanged
# to the Python invocation.
exec poetry run bash -c 'cd "$0" && python "$@"' "$OLD_PWD" "$@"
```

As mentioned in the comments, the `runpoetry` wrapper script can be used as the shebang line and will invoke `poetry install` before invoking the original script with `poetry run`.

However, the path changing code is a bit scary and it would be awesome to be able to use a `--project` flag instead.

sdispater:
Closing this since this is not a direction I don't want Poetry to go in at the moment.

PR #367, opened 2018-08-03
Better support for private repositories

ojii:
While 0.11.3 added rudimentary support for installation from private repositories (unrelated to this PR: search still seems to not work at all), it does not support some more advanced set ups, such as hosting the private packages on S3 using something like pypicloud. In that case, the Simple index has links to S3 with the authentication in the query string. Sending basic auth headers to S3 will fail, because S3 only allows one kind of authentication (which is already included in the query string). To solve this, I propose to change how authentication for repositories is done in poetry by only sending the username/password to the host for which they are registered in the config.

Right now the PR simply does it in legacy repository. This works for me to add private packages in my setup. Before I try to add tests etc, I would like to get feedback if this approach is acceptable or not.

# Pull Request Check List

- [ ] Make use of the new auth system everywhere it is appropriate

### DIFF
--- a/poetry/repositories/legacy_repository.py
+++ b/poetry/repositories/legacy_repository.py
@@ -1,6 +1,8 @@
 import cgi
 import re
 
+from requests.auth import HTTPBasicAuth, AuthBase
+
 try:
     import urllib.parse as urlparse
 except ImportError:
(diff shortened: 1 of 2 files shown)


ojii:
poetry 0.12 still doesn't work with private repositories (at least pypicloud which is what I use). This PR or something similar could solve this. @sdispater if I resolve the conflicts, would you consider merging this?

ojii:
@sdispater what needs to be done to merge this?

ojii:
a simple `poetry add <private>` seems to work, so closing this.

PR #389, opened 2018-08-12
Add a --lock option to poetry update

greysteil:
First up, thanks for Poetry! Seriously impressive bit of kit, particularly all the work on the resolver.

I run [Dependabot](https://dependabot.com), and a couple of customers have asked me to support Poetry (issue [here](https://github.com/dependabot/feedback/issues/57)). It would make it *way* easier for me to do so if you'd be happy to add a `--lock` option to the `poetry update` command, as implemented in this PR. (Alternatively, allowing specific packages to be updated when calling `poetry lock` would work for me, too, but the UI on on that doesn't feel intuitive.)

What do you reckon? `npm` have this pattern with `--package-lock-only`, so there's precedent.

I haven't added a test because there weren't any previously, and the code is pretty trivial.

### DIFF
--- a/README.md
+++ b/README.md
@@ -364,6 +364,8 @@ poetry update requests toml
 #### Options
 
 * `--dry-run` : Outputs the operations but will not execute anything (implicitly enables --verbose).
+* `--no-dev` : Do not install dev dependencies.
+* `--lock` : Do not perform install (only update the lockfile).
 
 ### add
 
--- a/docs/docs/cli.md
+++ b/docs/docs/cli.md
@@ -149,6 +149,8 @@ poetry update requests toml
 ### Options
 
 * `--dry-run` : Outputs the operations but will not execute anything (implicitly enables --verbose).
+* `--no-dev` : Do not install dev dependencies.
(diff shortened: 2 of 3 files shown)


greysteil:
@sdispater - any thoughts? Would love to have this merged. :octocat:

sdispater:
First of all thanks for taking the time to put this PR together!

However, unless I am missing something `update --lockfile-only` is equivalent to using the `lock` command.

greysteil:
> However, unless I am missing something update --lockfile-only is equivalent to using the lock command.

Without a whitelist of package passed, yes. The intention of this option, however, is to be able to say `poetry update requests --lockfile-only`.

sdispater:
@greysteil I see what you mean. Sorry for missing that.

Looks good to me overall. I think, however, that we can rename the option to a much simpler `--lock`.

greysteil:
Renamed!

greysteil:
And tests are passing, so hopefully this is now good to go. Thanks for your advice on it @sdispater.

sdispater:
Thanks!

PR #391, opened 2018-08-14
Avoid crashing on missing link

FinchPowers:
I am not sure if this is a desirable behaviour, but if the remote package is gone, there is an issue with `default_link = links[0]` right after my edits. The list is empty and it crashed with IndexError.


sdispater:
Commit 69f2f6dbad560708e3dab1ce7388389baffb1ad6 takes care of this (and other related issues)

Issue #281, opened 2018-07-05
How can I ignore the 'VIRTUAL_ENV' environment variable?


jacebrowning:
- [x] I have searched the [issues](https://github.com/sdispater/poetry/issues) of this repo and believe that this is not a duplicate.

## Issue

In my projects, I create the virtual environment in the project root (`$ poetry config settings.virtualenvs.in-project true`) so that I can explicitly manage dependencies and cache them between builds.

Travis CI runs builds inside of its own virtual environment (thus sets `VIRTUAL_ENV`).

The current logic consults `VIRTUAL_ENV` before any other settings: https://github.com/sdispater/poetry/blob/ae1773d68f341c34d9124353f7a952d5fd3cfefb/poetry/utils/venv.py#L47-L50

How can I tell `poetry` to ignore `VIRTUAL_ENV` and always create a virtual environment?

`pipenv` gets around this by adding another environment variable (`PIPENV_IGNORE_VIRTUALENVS`). but I think `poetry` can do better.

---

## Proposal A

The default value of `settings.virtualenvs.create` is inferred from `VIRTUAL_ENV` (`false` if set, `true` if unset).

If the user runs `$ poetry config settings.virtualenvs.create true` then `poetry` will always create a virtual environment irrespective of `VIRTUAL_ENV`.

## Proposal B

Add `settings.virtualenvs.ignore-active` to ignore `VIRTUAL_ENV` when creating a virtual environment.

## Proposal  C

If `settings.virtualenvs.in-project` is set and `./.venv/` does not exist then `VIRTUAL_ENV` is ignored.

bertjwregeer:
@jacebrowning, run poetry while unsetting `VIRTUAL_ENV`:

```env -u VIRTUAL_ENV poetry```

Now even though you are "Inside" an virtual environment, poetry won't know about it.

**Comment by jacebrowning** [comment]
@bertjwregeer Thanks for the tip. This actually results in the same behavior, but for a different reason.

Given this part of the Travis CI configuration:

```
cache:
  directories:
    - .venv
```

1. Travis CI creates `.venv` from the (yet to exist) cached virtual environment. 
2. Poetry sees this empty `.venv` directory and incorreclty assumes the virtual environment has already been created.
3. It them proceds to install dependencies globaly (also noted on https://github.com/sdispater/poetry/issues/108), which is actually into Travis CI's existing virtual environment.
[...]

PR #404, opened 2018-08-23
Improve virtualenv detection and management

sdispater:
This PR improves the way Poetry detects virtualenvs and the context it is running in.

The main change here is that it no longer relies only on the `VIRTUAL_ENV` environment variable to get the virtualenv but also checks the `real_prefix/base_prefix` `sys` properties.


jacebrowning:
How does this impact https://github.com/sdispater/poetry/issues/281?

sdispater:
@jacebrowning It does not impact #281.

It's not possible to ignore the `VIRTUAL_ENV` environment variable without breaking other things, especially `tox` if you want to use the global `poetry` executable.

The reason for that is if Poetry prefers the `virtualenvs.in-project` setting over `VIRTUAL_ENV`, it will always use the `.venv` virtualenv regardless of the current virtualenv `tox` has created.

PR #415, opened 2018-08-31
Added --no-dev option to show command

rodcloutier:
This PR intends to allow the `show` command to filter the displayed dependencies based on their category by adding a `--no-dev` option.

# Pull Request Check List

### DIFF
--- a/poetry/console/commands/show.py
+++ b/poetry/console/commands/show.py
@@ -10,6 +10,7 @@ class ShowCommand(EnvCommand):
 
     show
         { package? : Package to inspect. }
+        { --no-dev : Do not list the dev dependencies. }
         { --t|tree : List the dependencies as a tree. }
         { --l|latest : Show the latest version. }
         { --o|outdated : Show the latest version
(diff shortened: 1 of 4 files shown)


poetry/console/commands/show.py
```diff
@@ -36,11 +37,11 @@ def handle(self):
         if self.option("outdated"):
             self.input.set_option("latest", True)
 
-        locked_repo = self.poetry.locker.locked_repository(True)
+        locked_repo = self.poetry.locker.locked_repository(not self.option("no-dev"))
 
         # Show tree view if requested
         if self.option("tree") and not package:
-            requires = self.poetry.package.requires + self.poetry.package.dev_requires
```
sdispater: I don't think this should be completely removed. We still need to add the dev packages if the `--no-dev` option is not set.

rodcloutier: You are right. I have corrected the file.

sdispater:
Thanks a lot!

PR #418, opened 2018-09-02
Require at least TOML Kit 0.4.4 for performance

jacebrowning:
This will incorporate the fix for https://github.com/sdispater/tomlkit/issues/10 -- a significant performance bug on Python 2.


sdispater:
Thanks!

PR #427, opened 2018-09-12
Fix 'poetry.utils.env.Env.run(shell=True)' failing when the path to the venv contains a space (resolves #199)

### LINKED ISSUE #199: Poetry add fails if path to venv contains space
If I run the CLI command `poetry add <package-name>` whilst in a virtual environment to which the path contains a space I get the following error:
```
[VenvCommandError]  
Command C:\Users\Oisin Smith\PycharmProjects\WxEpos\venv\Scripts\python.exe -c "import sys; print('.'.join([str(s) for s in sys.version_info[:3]]))" errored with the following output:
'C:\Users\Firstname' is not recognized as an internal or external command operable program or batch file
```

The problem seems to be that the venv is somewhere inside the `C:\Users\Firstname Lastname` directory that contains a space. This seems to be the Line causing the error https://github.com/sdispater/poetry/blob/master/poetry/utils/venv.py#L175

One possible fix would be to change [this](https://github.com/sdispater/poetry/blob/master/poetry/utils/venv.py#L241) line to add quotes around the `bin` (ie. `bin = f"'{bin}'"`) string.` I'm not sure where it would be most appropriate to change, so if you let me know, I could create a PR.

Thank you for your great work on this package!

cosmicspacedragon:
One issue that I discovered while debugging this issue, is that `subprocess.Popen(shell=True)` decides which shell to use (on Windows) based on the `ComSpec` environment variable. 
Usually, this should point to `C:\Windows\System32\system32.exe` and that is what my fix assumes.
But if someone has set that to something like `C:\Windows\System32\WindowsPowerShell\v1.0\powershell.exe`, then that person will have some issues since PowerShell wants a `.` in front of a direct path to an executable for that executable to be executed.
I have chosen not to devise a solution for this issue since I suspect that the number of people that have `ComSpec` pointed at `powershell.exe` is minimal at best.

---
## Pull Request Check List

~~No documentation changes are needed since this is just a bug-fix.~~
EDIT: Might need to add some documentation to add some documentation for the `ComSpec` as `powershell.exe` breaking things issue.


poetry/utils/venv.py
```diff
@@ -243,6 +243,8 @@ def run(self, bin, *args, **kwargs):
         call = kwargs.pop("call", False)
 
         if shell:
+            if self._windows:
+                cmd[0] = '"{}"'.format(cmd[0])
```
rooterkyberian: two things:
* we should escape path on all platforms (`"` should work fine - `'` while safer on UNIX-like systems may indeed not work on windows, but I didn't check)
* we should most likely escape all arguments this way when transform list of args to a string for shell to interpret

tests/utils/test_venv.py
```diff
@@ -0,0 +1,19 @@
+
+import sys
+
+from poetry.utils import venv
+
+
+def test_venv_run_on_windows_with_paths_with_spaces(tmpdir):
```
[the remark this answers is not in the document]
rooterkyberian: even if the issue was Windows only (I don't believe it is) we still want to make sure we support spaces in paths on all platforms

cosmicspacedragon:
The latest changes escape arguments. I'll add some tests tomorrow or something.

poetry/utils/venv.py
cosmicspacedragon: Regarding escaping paths on Windows:
* PowerShell seems to want either an ampersand or a dot in front of the escaped path for the executable to be executed `[&|.] '<path to exe>'`. If there's no ampersand/dot, it just echoes the path back.
* PowerShell is fine with both single and double quotes.
* CMD only wants the escaped path.
* CMD only wants double quotes.
Do note that this is based on me just trying stuff for myself. The actual details on escaping executable paths for PowersShell might be a little different.

cosmicspacedragon:
The code currently seems like it's flipping `/` to `\` and sometimes `\\`.
Is this a real problem (because of escape sequences or something)?

It also fails when escaping `python` to `'python'` on AppVeyor (which I assume is running some form of Linux).
CMD has no such problem and PowerShell doesn't care beyond the quirks I've already mentioned.

cosmicspacedragon:
`git clone https://github.com/sdispater/poetry.git`
`cd poetry`
`git checkout 58dbca4`
`python -m poetry install`

---

Worked fine with `ComSpec` as `cmd.exe`, but `ComSpec` as `powershell.exe` is still borked.
Why anyone would mess with `ComSpec` is beyond me, but I tested it just for completeness' sake.

PR #430, opened 2018-09-12
[Documentation] Added instructions for Oh-My-Zsh completion

declared-lambda:
# Pull Request Check List

N/A
N/A

### DIFF
--- a/docs/docs/index.md
+++ b/docs/docs/index.md
@@ -86,6 +86,11 @@ poetry completions fish > ~/.config/fish/completions/poetry.fish
 
 # Zsh
 poetry completions zsh > ~/.zfunc/_poetry
+
+# Oh-My-Zsh
+mkdir $ZSH/plugins/poetry
+poetry completions zsh > $ZSH/plugins/poetry/_poetry
+
 ```
 
 !!! note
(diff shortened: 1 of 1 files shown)


sdispater:

However, unless I am missing something, won't the instructions for `zsh` also work on `oh-my-zsh`?

declared-lambda:
They do work but it's not clear as neither `compinit` nor fpath are in `~/.zshrc`.  


For `oh-my-zsh` to load _poetry, `fpath+=~/.zfunc` needs to be put before `oh-my-zsh`'s 
`plugins(...)` call.

The other solution to this, is the one I proposed in the PR, where completions are exported in `oh-my-zsh`'s plugins folder and is loaded by adding poetry in `plugins( ... )` call.


I initially opened a PR in `oh-my-zsh` that added `poetry`'s completions in the plugins folder but was rejected and the proposed solution in the reply is not working at all.  
https://github.com/robbyrussell/oh-my-zsh/pull/7132  

I can revert it back and add a note for `oh-my-zsh` to insert `fpath+=~/.zfunc` before `plugins(...)` if that'd be better.

levi-rs:
I found out about Poetry from [this HN post](http://andrewsforge.com/article/python-new-package-landscape/), and wanted to give it a try. I couldn't get the zsh completion instructions to work, as @declared-lambda outlined, as I use oh-my-zsh. I followed the PR changes and they worked perfectly.

webknjaz:
This PR is good. I'd only suggest adding an omz plugin to the upstream repo as it's a true way to introducing such integration.

sdispater:
Thanks!

MisLink:
How about outputting to `$ZSH/custom/plugins`? `$ZSH/plugins` is under git and means accepted by oh-my-zsh maintainers.

PR #440, opened 2018-09-17
[DOC] Relative links broken on https://poetry.eustace.io/

nayyarv:
If you check the relative links in https://poetry.eustace.io/docs/basic-usage/ for example, you get a 404 error. I was able to work around it, but thought it was a good idea to have it fixed upstream.

I'm not super sure how to test it, but I did manually edit the URLs and it worked fine. I did a regex search to find the URLs, I think it's exhaustive, but can't be sure

Additionally, I noticed the Introduction can't be clicked in the navigation, probably because it's the index.html page and doesn't have a nav_item.url.

I think this in nav.html
```html
{%- if nav_item.url %}
<a class="{% if nav_item.active%}current{%endif%}" href="{{ base_url }}/{{ nav_item.url }}">{{ nav_item.title }}</a>
{%- else %}
<span class="caption-text">{{ nav_item.title }}</span>
{%- endif %}
```
needs to be edited, but I don't have a good sense of what's the best here, but given that the pages are all URLs I don't see the need for this if statement. I would just have the link created unconditionally.

### DIFF
--- a/docs/docs/basic-usage.md
+++ b/docs/docs/basic-usage.md
@@ -1,7 +1,7 @@
 # Basic usage
 
 For the basic usage introduction we will be installing `pendulum`, a datetime library.
-If you have not yet installed Poetry, refer to the [Introduction](/) chapter.
+If you have not yet installed Poetry, refer to the [Introduction](/docs/) chapter.
 
 ## Project setup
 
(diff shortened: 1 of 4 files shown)


sdispater:
Thanks!

PR #442, opened 2018-09-18
fix: publishing will not fail if wrong credentials

jondot:
Fixing a bug in publishing functionality:

- When poetry is given wrong credentials, it will act as if publishing succeeded.

My analysis shows that while trying to "catch" the edge case for non-registered package, the case for bad auth credentials is missed.

Simplified the branching logic and made catching the non-registered package case more explicit leads to resolution.


brycedrennan:
Thanks for noticing this. It looks like it was fixed later in https://github.com/sdispater/poetry/commit/903514060748496f9c87513ba7fb33ed29158325

PR #484, opened 2018-10-09
Fix several DeprecationWarning: invalid escape sequence

BoboTiG:
Hello,

A little patch to fix several invalid escape sequences.


sdispater:
Thanks!

PR #493, opened 2018-10-16
Package Version + python version handling is not correct, tests to prove

### LINKED ISSUE #439: Poetry installing dev/beta releases
I think this may be more to do with the way certain packages release on pip, in my case pytorch and tensorflow - I've not noticed this behaviour in other packages to date.
Combined with neural nets being used to generate poetry, it's tricky to google search for this so apologies if this is a duplicate.

Ubuntu 18.04
Poetry 0.11.5

Tensorflow toml
```toml
[tool.poetry.dependencies]
python = "^3.6"
numpy = "^1.15.0"
scipy = "^1.1.0"
scikit-learn = "^0.19.2"
tensorflow = "^1.10.1"
tensorflow-gpu = "^1.10.1"
```
```
  - Installing scikit-learn (0.19.2)
  - Installing tensorflow (1.11.0rc0)
  - Installing tensorflow-gpu (1.11.0rc0)
  - Installing ujson (1.35)
```
In this case this happens because I've specified numpy 1.15, only the 1.11 rc is able to work correctly. I'm not sure this is optimal behaviour - I shouldn't get an alpha release in this case. Once I bump numpy to 1.14 I get 1.10.1 again.

In pytorch, it's a bit different and I'm not sure what happens

```toml
[tool.poetry.dependencies]
python = "^3.6"
numpy = "^1.15.0"
scipy = "^1.1.0"
scikit-learn = "^0.19.2"
torch ="^0.4.1"
torchvision = "^0.2.1"
```
```
➜ poetry update
Updating dependencies
Resolving dependencies... (48.5s)

Package operations: 7 installs, 0 updates, 0 removals

Writing lock file
...
  - Installing pillow (5.2.0)
  - Installing torch (0.4.1.post2)
                                                                                                                                         
[VenvCommandError]                                                                                                        
Command ['pip', 'install', '--no-deps', 'torch==0.4.1.post2'] errored with the following output:                                       
Collecting torch==0.4.1.post2
```
Again doing `pip3 install --user torch torchvision` will install `0.4.1` while poetry tries to install the post2 release. I'm not sure what's going on in this version tbh. I've worked around by setting equal for the time being.

nayyarv:
Hi, this is in relation to #439 where I found that poetry kept installing pytorch 0.4.1post2 despite being for 3.7 only and my venv/pyproject file being 3.6. This will focus on `requests` and the tests I developed while trying to fix #439 that might be helpful.

As of requests 2.19, the `requires_python` field is filled in in the `info` and in `[releases][{version}]` in the pypi metadata. I've included the stanza for 2.19 in the requests.json as an illustration, but have not updated the top level `info` section. Release 2.18.0 has a blank `requires_python` section for comparison

As a result, I'd expect the `python_constraint` of the package for 2.18.0 to be `*` i.e. any while for `2.19.0` it should be `">=2.6, !=3.0.*, !=3.1.*, !=3.2.*, !=3.3.*"`. However, the 2.19 test fails, expecting a python constraint of * too.

It appears poetry only looks at the package top level `[info][requires_python]` (this can be verified by changing the json and seeing the results of the test), not the release version info which tells us which package versions are compatible with which python versions. If requests drops support for python2.7 in 2.2, I won't be able install 2.19.1, even if it supports 2.7. 

The function `_get_release_info` in `pypi_repository.py` seems to be the one to blame
```python
json_data = self._get("pypi/{}/{}/json".format(name, version))
info = json_data["info"]
data = {
    "name": info["name"],
    "version": info["version"],
    "summary": info["summary"],
    "platform": info["platform"],
    "requires_dist": info["requires_dist"],
    "requires_python": info["requires_python"],
    "digests": [],
    "_cache_version": str(self.CACHE_VERSION),
}

try:
    version_info = json_data["releases"][version]
except KeyError:
    version_info = []

for file_info in version_info:
    data["digests"].append(file_info["digests"]["sha256"])
....
return data
# there is more code below, but version_info is unused after above for loop
```

This seems incorrect and explains why version handling is not being done correctly for packages that don't gracefully support a large number of python versions.

As for `torch`, each version has a separate release targeting different python versions. And it uses `python_version` instead to specify which python version it's targeting. This is also not handled in poetry.

This seems to be a somewhat fundamental bug in poetry's handling of pypi metadata and packaging. Poetry needs to track the union of the various versions each release handles and then install the correct one as required once dependency resolution is complete.

### DIFF
--- a/tests/repositories/fixtures/pypi.org/json/requests.json
+++ b/tests/repositories/fixtures/pypi.org/json/requests.json
@@ -2528,6 +2528,42 @@
         "url": "https://files.pythonhosted.org/packages/b0/e1/eab4fc3752e3d240468a8c0b284607899d2fbfb236a56b7377a329aa8d09/requests-2.18.4.tar.gz"
       }
     ],
+    "2.19.0": [
+      {
+        "comment_text": "",
+        "digests": {
+          "md5": "1ae2b89f8ea3e4aea8b199987fb2aae9",
+          "sha256": "421cfc8d9dde7d6aff68196420afd86b88c65d77d8da9cf83f4ecad785d7b9d6"
+        },
+        "downloads": -1,
+        "filename": "requests-2.19.0-py2.py3-none-any.whl",
+        "has_sig": false,
+        "md5_digest": "1ae2b89f8ea3e4aea8b199987fb2aae9",
+        "packagetype": "bdist_wheel",
+        "python_version": "py2.py3",
+        "requires_python": ">=2.6, !=3.0.*, !=3.1.*, !=3.2.*, !=3.3.*",
+        "size": 91865,
+        "upload_time": "2018-06-12T14:46:15",
+        "url": "https://files.pythonhosted.org/packages/cc/15/e1c318dbc20032ffbe5628837ca0de2d5b116ffd1b849c699634010f6a5d/requests-2.19.0-py2.py3-none-any.whl"
+      },
+      {
+        "comment_text": "",
+        "digests": {
+          "md5": "8a7844c58d496e9e92481de459830229",
+          "sha256": "cc408268d0e21589bcc2b2c248e42932b8c4d112f499c12c92e99e2178a6134c"
+        },
+        "downloads": -1,
+        "filename": "requests-2.19.0.tar.gz",
+        "has_sig": false,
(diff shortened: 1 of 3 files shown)


sdispater:
Closing since this is not a mergeable PR.

PR #515, opened 2018-10-18
fix repository links (poet ➜ poetry)

rooterkyberian:
Seems there still were some `poet` repository links lingering.
This causes https://pypi.org/project/poetry/ to link `poet` repository and its statistics instead of `poetry`.


sdispater:
Thanks!

PR #545, opened 2018-10-23
POC of conda.

tshauck:
> If you have *any* questions to *any* of the points above, just **submit and ask**!  This checklist is 
here to *help* you, not to deter you from contributing!

In light of this, I was hoping to show how conda support might work, and ask a couple questions about next steps.

Overall, the code itself is pretty straightforward. Miniconda has a `CONDA_PREFIX` env var in the same way that virtualenv has `VIRTUAL_ENV`.

The tricky part comes later in

https://github.com/sdispater/poetry/blob/475c6572ac3eb04d5a4136a4b01890020ae9129e/poetry/utils/env.py#L430-L438

Where the `VIRTUAL_ENV` env var gets set temporarily, but this wouldn't work for conda as it would expect `CONDA_PREFIX` to be set.  My thought there would be to tell `class VirtualEnv` if it is inside a conda or virtualenv, and if so, set the appropriate env var, but maybe there's a better way?

The other question was w.r.t. testing. Do you have some sense for what would be an appropriate test setup for this PR? As far as I can tell testing `VIRTUAL_ENV` doesn't exist.

As an aside, it's worth noting, as a workaround for installing with conda 

```
$ VIRTUAL_ENV=/Users/thauck/src/miniconda3/envs/poetry poetry install
```

seems to work.


tshauck:
@sdispater, gonna go head and close this, but if you think this is worth doing (and you could answer my questions), I'll reopen and finish it up.

PR #546, opened 2018-10-23
Support symlinking the poetry script

jhrmnn:
It is common to symlink executables installed at different locations to a single directory, so that `PATH` doesn't have to contain dozens of entries. The `poetry` script created by the `get-poetry.py` installer doesn't support being symlinked to from somewhere else. This patch fixes that.

# Pull Request Check List

### DIFF
--- a/get-poetry.py
+++ b/get-poetry.py
@@ -195,7 +195,7 @@ import glob
 import sys
 import os
 
-lib = os.path.realpath(os.path.join(os.path.dirname(__file__), "..", "lib"))
+lib = os.path.normpath(os.path.join(os.path.realpath(__file__), "../..", "lib"))
 
 sys.path.insert(0, lib)
 


zyv:
Oh, sorry, my PR #685 is a duplicate of this one, but I find the approach I suggested better in that I see no reason to replace original `realpath` with `normpath`, because resolved path might in turn be a symlink.

azag0:
> because resolved path might in turn be a symlink.

I don't think so. `os.path.realpath()` [removes all](https://docs.python.org/3.7/library/os.path.html#os.path.realpath) symbolic links along the path, not just the endpoint. So after the first call to `realpath()`, one can be sure to not encounter any more symbolic links.

The reason to use `realpath()` only once is that it has to look at the file system, whereas `normpath()` works only on the string, so it's much cheaper. Which doesn't really matter in this case, but that was the reason for why I did it the way I did it.

zyv:
> So after the first call to realpath(), one can be sure to not encounter any more symbolic links.

Ah, I think I understand what do you mean - normpath(`/real/path/without/symlinks/../../lib`) -> `/real/path/lib`, but the fact that it might be a symlink actually should not pose a problem anymore, unlike one level above. The performance argument is moot though - the symlink will have to be resolved upon import at latest anyways.

I'll be happy if either of the requests get merged...

sdispater:
Thanks!

PR #562, opened 2018-10-25
Use configured repositories as sources by default

abn:
This change adds globally configured extra repositories from config.toml as dependency sources. These are added after project specific sources and before PyPI.

The motivation behind this change was to enable a cleaner configuration of private repositories/mirrors under automated pipeline build environments without hard coding repository configurations within the project sources. Additionally, this provides functionality similar to pip's extra_index config allowing global configuration of repositories.

# Pull Request Check List

Resolves: #625

### DIFF
--- a/docs/docs/repositories.md
+++ b/docs/docs/repositories.md
@@ -54,7 +54,11 @@ with the `--username` and `--password` options.
 Now that you can publish to your private repository, you need to be able to
 install dependencies from it.
 
-For that, you have to edit your `pyproject.toml` file, like so
+By default, Poetry uses all repositories configured as shown above for package
+installation.
+
+If you would like to use a repository that was not configured like above, you have to
+edit your `pyproject.toml` file, like so
 
 ```toml
 [[tool.poetry.source]]
--- a/poetry/poetry.py
+++ b/poetry/poetry.py
@@ -46,6 +46,17 @@ class Poetry:
         for source in self._local_config.get("source", []):
             self._pool.add_repository(self.create_legacy_repository(source))
 
+        # Add any additional repositories configured globally as source
+        # This is done after local config sources and before PyPI
+        extra_repositories = self._config.setting("repositories", default={})
+        for repository_name, repository_config in extra_repositories.items():
+            if "url" in repository_config:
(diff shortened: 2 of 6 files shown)


abn:
@sdispater would be good get your thoughts on this one. Happy to make any changes as required.

sdispater:
This PR is superseded by #908 which introduces a way to specify a source as the default

abn:
@sdispater the feature being worked on in #908 is going to be definitely helpful. However, if I am understanding the change there, it is different to what was intended here.

In #908, only repositories configured in the project's `pyproject.toml` is accounted for. The intention here was to use repositories configured in poetry's global `config.toml` without the need to add a source for those in every project.

Is this behaviour intended to be added as part of #908?

drunkwcodes:
What is your use case?

I think it is convenient for changing dependency without changing `pyproject.toml`.

But it also tends to bypass `pyproject.toml` and make it obsolete.
I can't find a use case by myself to see this as a requirement.

abn:
@drunkwcodes for what it is worth, example scenarios are mentioned in the description. 

Expanding on that, having a global configuration has merits in the context of production use cases and build pipelines final application builds might use specific repositories that might only be accessible from build environments. In addition, this has the potential to assist developer workflows, for example in scenarios where a devpi servers are used.

The implementation proposed falls back to global repositories, ie. packages are searched first in repositories configured by the project configuration and if they are not found, we fallback to global repositories.

drunkwcodes:
@abn But the implementation sucks more than your lengthy comment. According to your test, 
the line `assert repositories.index(source_global) < repositories.index(source_pypi)` does not mean "global config as the fallback."

abn:
@drunkwcodes but `assert repositories.index(source_local) < repositories.index(source_global)` does.

drunkwcodes:
@abn You didn't answer my question.

What's more, I remembered poetry's interactive package picker did not print index url out.
And the user's global config is volatile and often overlooked.

So you changed the precedence of index, and not refining the picker, hence change the default behavior indirectly.

Reaching PyPI with best effort and disabling this explicitly are the acceptable ones.

PR #573, opened 2018-10-28
Correct a couple typos in get-poetry.py

altendky:


altendky:
Should I move this to `develop`?

PR #579, opened 2018-10-29
Add import command

dmerejkowsky:
Right now only Pipfile is supported. --package-name and --package-version
are required because there's no way to guess them when importing from
Pipfile

### DIFF
--- a/poetry/console/commands/import_.py
+++ b/poetry/console/commands/import_.py
@@ -34,7 +34,7 @@ class ImportCommand(Command):
         return self.import_pipfile(from_file)
 
     def import_pipfile(self, pipfile):
-        # When using Pipfile, there's now way to get name and version
+        # When using Pipfile, there's no way to get name and version
         # so ask user to set them on the command line
         name = self.get_required_name()
         if not name:
--- a/poetry/console/commands/import_.py
+++ b/poetry/console/commands/import_.py
@@ -1,0 +1,145 @@
+# -*- coding: utf-8 -*-
+from __future__ import unicode_literals
+
+from poetry.layouts import layout
+from .command import Command
+from ...utils._compat import Path
+from ...utils.toml_file import TomlFile
+
+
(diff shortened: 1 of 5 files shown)


dmerejkowsky:
@sdispater We're currently in the process of migrating a bunch of Python projects from pipenv to poetry, hence this patch.

If you agree to integrate this feature in poetry, then I'll start writing documentation.

The code is good enough for us anyway :)

Thanks for the awesome project and clean code, the patch was a joy to write :+1:

poetry/console/commands/import_.py
```diff
@@ -0,0 +1,145 @@
...
+
+        if from_file != "Pipfile":
+            self.line("<error>Unsupported config file: {}".format(from_file))
+            self.line("Feel free to open a bug report!")
+            return 1
+
+        return self.import_pipfile(from_file)
+
+    def import_pipfile(self, pipfile):
+        # When using Pipfile, there's now way to get name and version
```
itsthejoker: there's now way?

poetry/console/commands/import_.py
```diff
@@ -0,0 +1,145 @@
...
+
+    """
+
+    def __init__(self):
+        super(ImportCommand, self).__init__()
+
+    def handle(self):
+        from_file = self.get_from_file()
+        if not from_file:
+            return 1
```
itsthejoker: Would it make more sense to return `True / False` rather than 0s and 1s?

dmerejkowsky: I think the return value of handle() is the return code of the script, but I may be wrong

sdispater:
Thanks for taking the time to put this PR together!

However, this is not a direction I want the project to head in. This introduces complexity and maintenance on a feature that is only temporary if Poetry ever gets traction.

dmerejkowsky:
Fair enough :)

PR #610, opened 2018-11-08
401 - Add venv command

zymergen-luke:
This makes it easier to programmatically access paths relative to the
virtualenv.  A common use case is to activate the virtualenv

    source ${poetry venv}/bin/activate

This programmatic access can be used to automate virtualenvs upon cd-ing
into a project or as an alias, for example

https://github.com/sdispater/poetry/issues/401

# Pull Request Check List

I'm basing this change on the `master` branch, as the `develop` branch appears to be behind master. For example, the MockEnv class does not exist yet on develop, despite existing in master.

I'm just throwing this up here to get this ticket started, but I'm up to alternative ideas about implementation.

A few other thoughts:
1. Maybe we should have an `info` command which is like `debug:info` but returns a single field. e.g. `poetry info:venv` would return the venv while `poetry info:python_version` would return the python version.
2. Or perhaps we merge this with debug:info so that it's something like `debug:info:venv` (although given that it would become api, I wouldn't like the debug prefix), while `debug:info` gives you all the info
3. I could also imagine a command to directly get the activate script. Either returning the path to the script:
`source ${poetry activate}`
or the contents of the activate script:
`eval ${poetry activate}`

### DIFF
--- a/README.md
+++ b/README.md
@@ -423,6 +423,22 @@ dependencies:
  - pytzdata >=2017.2.2
 ```
 
+
+### venv
+
+To find the path to the virtualenv directory, you can use the `venv` command.
+
+```bash
+poetry venv
+```
+
+This may be helpful if you need to find paths relative to the virtualenv. For
+example if you'd like to activate the virtualenv manually.
+
+```bash
+source ${poetry venv}/bin/activate
+```
+
 #### Options
 
 * `--no-dev`: Do not list the dev dependencies.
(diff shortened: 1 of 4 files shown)


zymergen-luke:
@sdispater - Can I get your thoughts on this?

sdispater:
Thanks for taking the time to make this PR and sorry for not getting back to you sooner.

However, I will close this PR since it is superseded by #731.

zymergen-luke:
Thanks for getting that feature in!

PR #660, opened 2018-11-21
bugfix search only names

jgirardet:
poetry was never searching only in names.

That said, maybe name search  should be the default behavior instead of `summary` is definitively to noisy ?


jgirardet:
do you think @sdispater  that it could be in the next release ?

sdispater:
Thanks!

And yes it will land in the next bugfix release.

PR #676, opened 2018-11-27
Fix CONTRIBUTING to use 'install' and not 'develop'

eblume:
Super minor fix to the documentation. My understanding is that `develop` has been deprecated in favor of `install`.


sdispater:
Thanks!

Issue #571, opened 2018-10-28
poetry shell does not activate virtualenv 


timonbimon:
- [x] I am on the [latest](https://github.com/sdispater/poetry/releases/latest) Poetry version.
- [x] I have searched the [issues](https://github.com/sdispater/poetry/issues) of this repo and believe that this is not a duplicate.

- **OS version and name**: Mac OS X, High Sierra
- **Poetry version**: 0.12.5

## Issue
Similar to ```pipenv shell```, I would have expected that when running ```poetry shell``` the virtualenv gets activated, but apparently this is not the case...

```console
➜ which python
/Users/timon/.pyenv/shims/python
➜ poetry shell
Spawning shell within /Users/timon/Library/Caches/pypoetry/virtualenvs/YOLO-SAR-py3.7
➜ which python
/Users/timon/.pyenv/shims/python
➜ source /Users/timon/Library/Caches/pypoetry/virtualenvs/yolo-sar-py3.7/bin/activate
➜ which python
/Users/timon/Library/Caches/pypoetry/virtualenvs/yolo-sar-py3.7/bin/python
```

for comparison
```console
➜ poetry run which python
/Users/timon/Library/Caches/pypoetry/virtualenvs/yolo-sar-py3.7/bin/python
```

Am I misunderstanding something and this is expected behaviour or is it a bug? 

Thanks a lot already for your time :)

itsthejoker:
Came to file the same issue. My shell shows that a virtualenv is activated, but it's empty; if I manually activate the .venv (installing locally to the project directory) then it works, but I would expect `poetry shell` to do that.

seansfkelley:
You might be experiencing #497.

itsthejoker:
@seansfkelley Confirmed, am experiencing #497. Good catch!

sobolevn:
I am also experiencing this.
I have all my setup opensourced, so one can follow all the step of my `zsh`/`pyenv`/`poetry shell` madness. Here it is: https://github.com/sobolevn/dotfiles

Related parts:
- https://github.com/sobolevn/dotfiles/blob/master/shell/.exports#L22
- https://github.com/sobolevn/dotfiles/blob/master/shell/.exports#L62-L63

Things I have tried:

```sh
if [[ -z "$VIRTUAL_ENV" ]]; then
  eval "$(pyenv init -)"
fi
```

[...]

Issue #497, opened 2018-10-17
improve reliability of `poetry shell` configuring $PATH properly


seansfkelley:
- [x] I am on the [latest](https://github.com/sdispater/poetry/releases/latest) Poetry version.
- [x] I have searched the [issues](https://github.com/sdispater/poetry/issues) of this repo and believe that this is not a duplicate.
- [x] If an exception occurs when executing a command, I executed it again in debug mode (`-vvv` option).

- **OS version and name**: macOS 10.14
- **Poetry version**: 0.11.5

## Issue

From https://github.com/sdispater/poetry/issues/198#issuecomment-430742299.

Preface: I'm not sure if this can be realistically solved by Poetry or if you're too much at the mercy of each person's shell configuration. But here goes.

Certain shell configurations interact poorly with `poetry shell`; see https://github.com/sdispater/poetry/issues/198#issuecomment-424016403 and https://github.com/sdispater/poetry/issues/172 for examples. Resolutions to the issue tend to be ad-hoc workarounds that involve spelunking in your shell configuration, see https://github.com/sdispater/poetry/issues/198#issuecomment-430742299 and https://github.com/sdispater/poetry/issues/172#issuecomment-401554154 respectively.

The root of the issue seems to be that Poetry attaches the virtualenv to the beginning of `$PATH`, then spawns the shell, which may override that configuration by attaching more things to the beginning of `$PATH`. If possible, modifying `$PATH` after the shell initializes ought to give a better result that will work by default on more systems. I don't recall seeing this issue with `pipenv shell`, but admittedly I didn't dig as deep while I was using it as my primary tool.

timonbimon:
also ran into this problem (that pyenv prepends its path to $PATH after poetry does). The fix suggested by @seansfkelley (modifying $PATH after the shell initializes) would probably help :)

itsthejoker:
I wrote my own solution to this. It's a hack and a mildly kludgy one, but it does work and I'm happy with it.

Requirements to implement as-is:
 - fish shell
- a pyenv installation that's brutally murdering your PATH
- a folder to drop scripts into that you can hardcode the path to

In `~/.config/fish/config.fish`, add this to the bottom so that it runs after pyenv:

```sh
function fix_shell
    eval (python ~/fish_scripts/fix_poetry_path.py)
end

fix_shell
```
Adjust the above script path to whatever you want. The python script:

```python
import os
import sys

def fix_path():
[...]

Issue #172, opened 2018-06-01
poetry and pyenv don't work well together


MrGreenTea:
I am using pyenv to manage my python versions and virtualenvs. This seems to cause some problems with poetry. 

```bash
>> which python
~/.pyenv/shims/python
>> which pip
~/.pyenv/shims/pip
>> poetry run which python
~/.cache/pypoetry/virtualenvs/XXXXX-py3.6/bin/python
>> poetry run which pip
~/.pyenv/shims/pip
```

This seems to cause the issue, that packages are installed in to the python version set by pyenv, but the python interpreter is used from the poetry virtualenv. This is no problem, when I activate the virtualenv first and then run `poetry install` 
But if I want to rely on poetry's automatic virtualenv management, this doesn't work anymore.

If you need more info I'll gladly provide it.

Thanks for your great work and tool, it's a real pleasure to work with :)

tkossak:
Would like to use it with pyenv too. For me it doesn't install in pyenv python:

```
> mkdir /tmp/2                                                                                                                                    
> cd /tmp/2                                                                                                                                       
> pyenv shell 3.6.5                                                                                                                     
> pyenv version                                                                                                                         
3.6.5 (set by PYENV_VERSION environment variable)

> pyenv which python                                                                                                                    
/home/kossak/.pyenv/versions/3.6.5/bin/python

> python -VV                                                                                                                            
Python 3.6.5 (default, Apr 25 2018, 13:20:33) 
[GCC 5.4.0 20160609]

> pip -V                                                                                                                                
pip 9.0.3 from /home/kossak/.pyenv/versions/3.6.5/lib/python3.6/site-packages (python 3.6)

> wget https://raw.githubusercontent.com/sdispater/poetry/master/get-poetry.py                                                          
--2018-06-07 14:47:40--  https://raw.githubusercontent.com/sdispater/poetry/master/get-poetry.py
Resolving raw.githubusercontent.com (raw.githubusercontent.com)... 151.101.112.133
Connecting to raw.githubusercontent.com (raw.githubusercontent.com)|151.101.112.133|:443... connected.
HTTP request sent, awaiting response... 200 OK
Length: 9670 (9,4K) [text/plain]
Saving to: ‘get-poetry.py’

get-poetry.py                           100%[=============================================================================>]   9,44K  --.-KB/s    in 0s      

2018-06-07 14:47:41 (37,0 MB/s) - ‘get-poetry.py’ saved [9670/9670]


[...]

PR #678, opened 2018-11-27
Include a fix for pyenv with poetry shell in documentation

eblume:
Please see #497, #571, #172, and possibly #198 for related issues.

In the case described by @itsthejoker here: https://github.com/sdispater/poetry/issues/497#issuecomment-437160361

a proposed solution is to have `fish` execute a script to fix $PATH after pyenv has initialized. I believe a better solution is to guard pyenv for $POETRY_ACTIVE. This PR includes documentation for this fix, as I suspect many people are hitting this roadblock.

I am not able to vouch that this fix works (or is needed) on any shells other than `fish`, but I would assume before merging you might want to include documentation for those shells.

Hopefully this is helpful!

### DIFF
--- a/README.md
+++ b/README.md
@@ -94,6 +94,23 @@ fpath+=~/.zfunc
 ```
 
 
+## Using `poetry shell` or `poetry run` alongside `pyenv`
+
+The commonly used python version management tool [pyenv](https://github.com/pyenv/pyenv) includes, by design,
+a layer of 'shims' that allow the same commands to route to different executables
+based on configuration variables. To avoid interference between the two systems,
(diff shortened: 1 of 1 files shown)


itsthejoker:
Oh. Thaaaat's a lot easier than my solution 😂 

An important side note is that in Fish, this solution will only work if you're manually declaring the inits for pyenv inside your config instead of using the omf `pyenv` plugin (which helpfully mangles the path for us: https://github.com/oh-my-fish/plugin-pyenv/blob/master/init.fish). Might need an additional line in your update.

eblume:
Please see https://github.com/sdispater/poetry/issues/497#issuecomment-443619684 for a real fix.

PR #719, opened 2018-12-07
Fix sonnet handling of single specified pythons

eblume:
I noticed that sonnet was erroring when given a command like::

    sonnet -P 3.7

It's possible I'm misunderstanding the expected syntax of the command, but I think this commit fixes the issue.


sdispater:
Thanks!

Can you rebase your PR on the `master` branch as it's a bug fix?

Issue #718, opened 2018-12-07
Don't modify ~/.profile if `.poetry/bin` already in $PATH


garyo:
I'd like to make it easier to install poetry on all my dev machines; using the curl method always modifies `~/.profile` which is annoying since I carefully manage all my dotfiles, so I have to go around undoing that mod after installing poetry. (I add it to the path myself.)
It would be nice if `get-poetry.py` would check if `~/.poetry/bin` is already in $PATH before modifying `.profile`.
I know I can get the installer and then run it locally and answer 'n', but there's also no cmd line option for `--no-set-path` for when I want to do it silently. (I can pipe `yes n` to it on Mac & Linux, that's a workaround.)

eblume:
@garyo something like this? ^

garyo:
Sweet!

PR #720, opened 2018-12-08
Add --no-set-path option for get-poetry.py

eblume:
This PR is a proposed resolution for #718 - the naming of `Installer._no_set_path` is unfortunate in that it has different semantics than the existing `Installer._modify_path` but sounds the same. I believe I've struck a happy medium here, but I am totally open to re-naming any of these variables.

### DIFF
--- a/get-poetry.py
+++ b/get-poetry.py
@@ -301,13 +301,15 @@ class Installer:
         version=None,
         preview=False,
         force=False,
+        nosetpath=False,
         accept_all=False,
         base_url=BASE_URL,
     ):
         self._version = version
         self._preview = preview
         self._force = force
-        self._modify_path = True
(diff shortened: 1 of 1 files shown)


eblume:
Hmmm, no idea why windows w/python 2.7 failed... it looks like poetry didn't get installed even though the script finished. I don't have a windows box to test.

gimbo:
This is great and I want it - however, I note that when you run this script, it still reports that:

```

$HOME/.poetry/bin

This path will then be added to your `PATH` environment variable by
modifying the profile files located at:

$HOME/.profile
$HOME/.zprofile
```

Just a thought...

gimbo:
(Also the `--uninstall` message mentions these paths; maybe that should be modified too.)

PR #725, opened 2018-12-10
[724] handle ast.Name values in extras_require

### LINKED ISSUE #724: Install command fails when extras_require.values() contains variables

- **OS version and name**: Windows 7
- **Poetry version**: 0.12.10
- **Link of a [Gist](https://gist.github.com/) with the contents of your pyproject.toml file**: n/a
## Issue

When installing a dependency which is specifying extra dependencies via a variable, like
`extras_require={"test": tests_require, ...` (see [graphene](https://github.com/graphql-python/graphene/blob/master/setup.py)), you get an exception `[AttributeError], 'Name' object has no attribute 'elts'` from `poetry/utils/setup_reader.py:285`. The code makes the implicit assumption that values in `extras_require` are lists.

asodeur:
# Pull Request Check List

Fixes #724.

There are more possibilities of confusing the parser, like calls as values for, ie `extras_require={'key': func(). ...}`. IMHO there needs to be a way to give-up and raise an expressive error message.

### DIFF
--- a/poetry/utils/setup_reader.py
+++ b/poetry/utils/setup_reader.py
@@ -282,7 +282,11 @@ class SetupReader(object):
 
         if isinstance(value, ast.Dict):
             for key, val in zip(value.keys, value.values):
-                extras_require[key.s] = [e.s for e in val.elts]
+                if isinstance(val, ast.Name):
+                    val = self._find_variable_in_body(body, val.id)
+
+                if isinstance(val, ast.List):
+                    extras_require[key.s] = [e.s for e in val.elts]
         elif isinstance(value, ast.Name):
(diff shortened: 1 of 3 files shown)


sdispater:
Thanks for this!

In my opinion, this is a bug fix so it should be based on `master`.

That being said, this PR looks good to me 👍

asodeur:
Well, we have to thank you for this awesome tool.

I cherry-picked the two commits into a new branch [branch](https://github.com/asodeur/poetry/tree/issue_724_on_master) based on `master`. AFAIK there is no way to change the source branch of a PR. Shall I close this and open a new one based on `master`?

asodeur:
Closing in favour of #883

Issue #655, opened 2018-11-20
Poetry using the wrong Python version (not related to pyenv)


zmitchell:
- [x] I have searched the [issues](https://github.com/sdispater/poetry/issues) of this repo and believe that this is not a duplicate.
- [x] I have searched the [documentation](https://poetry.eustace.io/docs/) and believe that my question is not covered.

## Question

I'm having an issue adding a package to a project because `poetry` is picking up on the wrong Python version. I'm pretty confident that something has gone awry in my system, but I'm doing my best to salvage the situation and not have to go through the whole process of installing Raspbian from scratch on this Raspberry Pi. I installed `poetry` using the recommended method (`curl ...`). Can you point me in the right direction? As mentioned in the title, I'm not using `pyenv`. All of the relevant issues I found were related to `pyenv`, so I'm not really sure how to get `poetry` to use the correct Python.

Here are some terminal outputs to help you diagnose the issue:
```
pi@widget:temp_logger$ python --version
Python 3.5.3
pi@widget:temp_logger!$ sudo python --version
Python 2.7.13
pi@widget:temp_logger$ poetry --version
Poetry 0.12.9
pi@widget:temp_logger$ poetry debug:info

Poetry
======

 * Version: 0.12.9
 * Python:  2.7.13


Virtualenv
==========

 * Python:         2.7.13
 * Implementation: CPython
 * Path:           NA


System
======

 * Platform: linux2
 * OS:       posix
 * Python:   /usr


pi@widget:temp_logger$ poetry add adafruit-blinka

[RuntimeError]
The current Python version (2.7.13) is not supported by the project (^3.5)
Please activate a compatible Python version.

add [-D|--dev] [--git GIT] [--path PATH] [-E|--extras EXTRAS] [--optional] [--python PYTHON] [--platform PLATFORM] [--allow-prereleases] [--dry-run] [--] <name> (<name>)...

```

sdispater:
As you can see in `debug:info`, Poetry picks up Python 2.7.13 and not Python 3.5.3. What is the output of `which python`? Does it point to your 2.7.13 Python or 3.5.3?

zmitchell:
Ah, here we go. It turns out `python` was aliased to `python3`, and doing `which python` uncovered that. Now `python --version` returns 2.7.13, and `which python` returns `/usr/bin/python`. So how do I get `poetry` to use Python 3 instead?

sdispater:
@buriy Commit 94304e9 should fix your issue.

[...]

PR #727, opened 2018-12-11
Fix `utils.env.SystemEnv.run("python", ...` on Windows

asodeur:
# Pull Request Check List
Fixes the issue that `poetry` does always use the primary system interpreter on Windows when called like
`<some path>\python.exe -m poetry` and `<some path>` is not inside a virtual environment, see #655.

Maybe superseeded by PR #731.

- [ ] Added **tests** for changed code. Not sure how to test this beyond trivial unit tests as pretty specific setup involving at least two interpreters is required.


asodeur:
Closing in favour of #882

Issue #621, opened 2018-11-13
Add a way to set a fixed Python version for a project


sdispater:
- [x] I have searched the [issues](https://github.com/sdispater/poetry/issues) of this repo and believe that this is not a duplicate.

## Issue

Since this is an issue that has arisen before (see #522) and there is some confusion around which Python version Poetry uses, introducing a more explicit way to tell which Python version to use seems reasonable.

Before we go into more details of what it would look like, it's important to note that the default behavior (using the currently activated Python) will remain. This new feature will only make it more convenient for people to choose explicitly what they want to use.

So, with that being said, here is what a possible implementation could look like.

This requires the introduction of a new command which would accept a python executable as an argument. The argument must be a full path to a Python executable or a `python` command that is in the path:

```bash
poetry project use --python python
poetry project use --python python3
poetry project use --python /full/path/to/python
```

What this command would do is create a virtualenv like the ones currently created in the form `{project_name}-py3.6`, and write the association of the virtualenv and the project to a file so that Poetry knows what to use.

If the virtualenv already exists and it is not currently associated with the project, it will be associated. This makes switching between virtualenvs easier.

This will likely be implemented for the release of the first stable `1.0.0` release.

sdispater:
> Since you propose to create the virtualenv with the project use command why not use the install command which already exists ?

No, because this is not the purpose of `install` which is only to install the dependencies.

Having a separate command makes the association of the current project with a specific Python version explicit.

> About the name of the new command

Regarding the name of the command, it is not set yet and is just here to show the intended use case.

> Should we handle the micro version ?

No, I don't think so. To me the minor version is the important one, not the patch version.

> Should we take care of project with same name ? maybe add a hash (path of project, author of project) in the name ?

I wondered too but the cases where you have two Python projects with the same name are rare and do not warrant changing the current implementation.

> About the multiplication of virtualenv in cache with this new option

This can be handled by the future `env` command. And now that I think about it, I think we can make the `env` command handle this feature (like a `env use` command).

jgirardet:
So if I understand your mind, poetry workflow will have to entry points:
`poetry install` : install using current python version
`poetry env use whatever-version`: install using custom version

I deduce :
`poetry env use /path/to/python3.6` : check if venv exists, create if needed, flag this venv in a config file for this project, then install poetry.
then
`poetry env use`  python3.5 : does same thing for python3.5 if python3.5 is in path or fail
 then
[...]

PR #731, opened 2018-12-11
New env command and sub commands

sdispater:
## Pull Request Check List

This PR introduces a new `env` command with related sub commands, to make it easier to explicitly associate a Python version with the current project. This addresses #621.

Here are the changes:

### New `env use` command

The `env use` command tells Poetry which Python version
to use for the current project.

```bash
poetry env use /full/path/to/python
```

If the python executable is in the `PATH`, it's possible to use it:

```bash
poetry env use python3.7
```

It's also possible in this case to only specify the minor Python version:

```bash
poetry env use 3.7
```

To disable the explicitly activated virtualenv, it's possible to use the
special `system` Python version to retrieve the default behavior:

```bash
poetry env use system
```

### New `env info` command

The `env info` command displays basic information about the currently activated virtualenv:

```bash
poetry env info
```

will output something similar to this:

```text
Virtualenv
Python:         3.7.1
Implementation: CPython
Path:           /path/to/poetry/cache/virtualenvs/test-O3eWbxRl-py3.7
Valid:          True

System
Platform: darwin
OS:       posix
Python:   /path/to/main/python
```

It's possible to only display the path of the virtualenv by passing the `--path` option
to `env info`:

```bash
poetry env info --path
```

### New `env list` command

The `env list` command lists all the virtualenvs associated with the current virtualenv.

```bash
poetry env list
```

will output something like the following:

```text
test-O3eWbxRl-py2.7
test-O3eWbxRl-py3.6
test-O3eWbxRl-py3.7 (Activated)
```

### New `env remove` command

The `env remove` command deletes virtualenvs associated with the current project:

```bash
poetry env remove /full/path/to/python
```

Similarly to `env use`, it's possible to pass either `python3.7`, `3.7` or the name of
the virtualenv (as returned by `env list`):

```bash
poetry env remove python3.7
poetry env remove 3.7
poetry env remove test-O3eWbxRl-py3.7
```

### Virtualenvs names now depend on the path of the project

This makes having different projects with the same name possible.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
 
 - Added an `export` command to export the lock file to other formats (only `requirements.txt` is currently supported).
 - Added a `env:info` command to get basic information about the current environment.
+- Added a `env:use` to control the Python version used by the project.
 
 ### Changed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@
 - The `debug:info` command has been renamed to `debug info`.
 - The `debug:resolve` command has been renamed to `debug resolve`.
 - The `self:update` command has been renamed to `self update`.
+- Changed the way virtualenvs are stored (names now dependent of the project's path).
 
 ### Fixed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,8 +5,10 @@
 ### Added
 
 - Added an `export` command to export the lock file to other formats (only `requirements.txt` is currently supported).
-- Added a `env:info` command to get basic information about the current environment.
-- Added a `env:use` to control the Python version used by the project.
+- Added a `env info` command to get basic information about the current environment.
+- Added a `env use` command to control the Python version used by the project.
+- Added a `env list` command to list the virtualenvs associated with the current project.
+- Added a `env remove` command to delete virtualenvs associated with the current project.
 
 ### Changed
 
@@ -15,7 +17,7 @@
 - The `debug:info` command has been renamed to `debug info`.
 - The `debug:resolve` command has been renamed to `debug resolve`.
 - The `self:update` command has been renamed to `self update`.
-- Changed the way virtualenvs are stored (names now dependent of the project's path).
+- Changed the way virtualenvs are stored (names now depend on the project's path).
 
 ### Fixed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 ### Added
 
 - Added an `export` command to export the lock file to other formats (only `requirements.txt` is currently supported).
+- Added a `env:info` command to get basic information about the current environment.
 
 ### Changed
 
--- a/docs/docs/basic-usage.md
+++ b/docs/docs/basic-usage.md
@@ -155,19 +155,44 @@ When you execute the `install` command (or any other "install" commands like `ad
 Poetry will check if it's currently inside a virtualenv and, if not, will use an existing one
 or create a brand new one for you to always work isolated from your global Python installation.
 
-!!!note
+By default, Poetry will use the currently activated Python version
+to create the virtualenv for the current project.
+    
+To easily switch between Python versions, it is recommended to
+use [pyenv](https://github.com/pyenv/pyenv) or similar tools.
+    
+For instance, if your project is Python 2.7 only, a standard workflow
+would be:
+    
+```bash
+pyenv install 2.7.15
+pyenv local 2.7.15  # Activate Python 2.7 for the current project
+poetry install
+```
 
-    To create the virtualenv for the current project, Poetry will use
-    the currently activated Python version.
+However, this might not be feasible for your system, especially Windows where `pyenv`,
+is not available. To circumvent that you can use the `env:use` command to tell
+Poetry which Python version to use for the current project.
 
-    To easily switch between Python versions, it is recommended to
-    use [pyenv](https://github.com/pyenv/pyenv) or similar tools.
+```bash
+poetry env:use /full/path/to/python
+```
 
-    For instance, if your project is Python 2.7 only, a standard workflow
-    would be:
+If you have the python executable in your `PATH` you can use it:
 
-    ```bash
-    pyenv install 2.7.15
-    pyenv local 2.7.15  # Activate Python 2.7 for the current project
-    poetry install
-    ```
+```bash
+poetry env:use python3.7
+```
+
+You can even just use the minor Python version in this case:
+
+```bash
+poetry env:use 3.7
+```
+
+If you want to disable the explicitly activated virtualenv, you can use the
+special `system` Python version to retrieve the default behavior:
+
+```bash
+poetry env:use system
+```
--- a/docs/docs/basic-usage.md
+++ b/docs/docs/basic-usage.md
@@ -171,28 +171,83 @@ poetry install
 ```
 
 However, this might not be feasible for your system, especially Windows where `pyenv`,
-is not available. To circumvent that you can use the `env:use` command to tell
+is not available. To circumvent that you can use the `env use` command to tell
 Poetry which Python version to use for the current project.
 
 ```bash
-poetry env:use /full/path/to/python
+poetry env use /full/path/to/python
 ```
 
 If you have the python executable in your `PATH` you can use it:
 
 ```bash
-poetry env:use python3.7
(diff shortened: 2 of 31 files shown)


sdispater:
@sprt This is a per-system feature, so the information is only persisted on the system the commands have been executed on. This is so that anyone can manage it's python versions and virtualenvs separately.

jgirardet:
That was a long way, but here it is thanks

sdispater:
@garyo 

> which virtualenv env vars are used?

What do you mean? If you mean the `VIRTUAL_ENV` env var, Poetry will use it to determine the current virtualenv if it is set but Poetry does not set it itself, unless you use `poerty shell`.

>  where virtualenvs go when created/used by poetry (and how to use ~/.virtualenvs if that's not the default)

Poetry creates the virtualenvs in its own cache directory, see: https://poetry.eustace.io/docs/configuration/#settingsvirtualenvspath-string

However, if you want them to be stored elsewhere, you can use the config command:

```bash
poetry config settings.virtualenvs.path ~/.virtualenvs
```

> Does it use the .project file in the virtualenv?

What do you mean?

> Also how to activate a poetry venv (I think it's just . $(poetry env info -p)/bin/activate or Scripts/ on Windows)

You have two choices here: `. $(poetry env info --path)/bin/activate` or `poetry shell`. Note that `poetry shell` will be improved before release `1.0.0`.

Also, note that this is not a virtualenvwrapper replacement. The created virtualenvs are mainly intended to be used by Poetry and are mostly an implementation detail. The goal of Poetry is to not make the end users, especially newcomers, worry about virtualenvs.

As to how to use it right now, I would recommend waiting for the next alpha release of version `1.0.0` which I will make once this PR is merged. It would then be just a matter of executing `poetry self:update --preview`.

garyo:
Yes, primarily `VIRTUAL_ENV`. Also `virtualenvwrapper` and maybe others respect `WORKON_HOME` as the root of the virtualenv tree. 
The `.project` file in the virtualenv can be used to cd to the project dir when activating, given only the virtualenv name. (I'm not sure if the standard activate script uses this or not.) (I don't think this is vital, I was just asking.)
As for activating, I tried subshells when I was using `pipenv` and hated them -- shell history, dir stack, shell vars, having to remember to not deactivate but exit instead... I definitely prefer activating in the current shell. Just my opinion.
Why do you say it's not a virtualenvwrapper replacement? It looks like it'll do everything I need -- which admittedly is pretty simple: just create envs in a central location, install stuff, activate in shells, and deactivate. Since virtualenvwrapper seems to be no longer maintained (last commit: Aug or Sept 2017 and already broken in python3.7) I think there is a need for something here and I'm hoping (at least for my simple use case) poetry can be it.

OK, I'll wait for the next alpha!

golyalpha:
It would also be nice to have something like `poetry env info --python` to display path to the Python interpreter executable, just like IDEs like VSCode want it (as outlined in #278).

By the way, since we're still a long way off from having Poetry supported by the VSCode's Python extension, I've been scouring through the it's docs, and I've found a way to at least get venvs from Poetry to show up in the interpreter selection menu:
```
"python.venvPath": "[path to poetry venvs]"
```
putting this in the user settings.json file (not the project one) results in this appearing in the interpreter selection menu:
image
You'll still have to fire up shell whenever you want to install project dependencies though.

bersace:
@sdispater how do you tell your editor to use poetry venv ? I'm using spacemacs, and got it to properly validate syntax and navigate in code by telling venv in pyenv's `.python-version` file. That would be nice to be able to bridge poetry env management and editor.

Anyway, that's a great step, and i'm willing to test it as soon as alpha released :-)

garyo:
@bersace I use emacs `virtualenvwrapper.el` (available from ELPA as `virtualenvwrapper` although not related to the normal `virtualenvwrapper` python package -- it's basically an all-emacs implementation of the same idea). 
It respects `WORKON_HOME` and falls back to `~/.virtualenvs`.

bersace:
@garyo how do you tell emacs which venv to activate when you open file / switch buffer ? `.venv` file ? I had issue with emacs not finding the right flake8 until I fall back to `.python-version`.

garyo:
My venvs are in `~/.virtualenvs` so `venv-workon` from that package TAB-completes the list. It's not ideally automated but it works enough for me.
It would be cool if the emacs code would parse the `.project` files in the virtualenvs and present any matching ones first, or just auto-select if only one. But that's a different project so let's take it up over there :-)

paranoidi:
This is nice welcomed addition! However please ensure that `poetry install` will automatically use correct version from pyproject.toml to create the virtualenv. The command is so simple and elegant that it would shame to complicate it by having to specify python version or worse create a wrong incompatible virtualenv. Which is unfortunately what is happening currently.

maggyero:
@sdispater What do you think of @paranoidi suggestion?

maggyero:
@sdispater, I think that @paranoidi's suggestion would improve the user experience dramatically. The default virtual environments created by the `poetry add`, `poetry build`, `poetry install`, `poetry lock`, `poetry remove`, `poetry run`, `poetry shell`, `poetry show` and `poetry update` commands should not be determined by the Python interpreter called by the `#!/usr/bin/env python` shebang line of the `poetry` script in the first place, but by the `tool.poetry.dependencies.python` property of the `pyproject.toml` file.

For instance, _in the absence of an activated virtual environment_ and with the following `pyproject.toml` file:

```
[tool.poetry.dependencies]
python = "^3.7"
```

the `poetry add`, `poetry build`, `poetry install`, `poetry lock`, `poetry remove`, `poetry run`, `poetry shell`, `poetry show` and `poetry update` commands would create a Poetry built-in virtual environment with the Python 3.7 interpreter (like if the user had run the `poetry env use 3.7` command), instead of one with the Python 2.7 interpreter as currently (like if the user had run the `poetry env use system` command).

jabbalaci:
Is it available? "`poetry env`" says this command is not defined.

tatianafrank:
> Is it available? "`poetry env`" says this command is not defined. I have version 0.12.17.

I think this feature is part 1.0 release.

tony:
I'm having some difficulties getting the virtualenv path regardless of shell state (important for me due to keeping my session configuration scripts simple as possible):

It can't reliably give the `pipenv` equivalent of `source `pipenv --venv`/bin/activate` which gives the activate regardless of being already in an env or not.  

- outside the virtualenv, `poetry env info` isn't accessible since apparently multiple virtualenvs can exist. so `poetry env list` has to be used: ` source $(poetry env list --full-path | head -n 1 2> /dev/null)/bin/activate`
- inside the virtualenv env, `poetry env info` is used

If anyone has the same problem and considers this a feature request or a bug, feel free to make an issue. This seems like an observation / bikeshed at this point since only I'm complaining

cglacet:
Is it supposed to work now (1.1.4)?

If it is, I get the following error:

```bash
❯ poetry env use python3.6.10
/bin/sh: python3.6.10: command not found

  EnvCommandError

  Command python3.6.10 -c "import sys; print('.'.join([str(s) for s in sys.version_info[:3]]))" errored with the following return code 127, and output:


  at ~/.pyenv/versions/3.8.0/lib/python3.8/site-packages/poetry/utils/env.py:345 in activate
       341│                     shell=True,
       342│                 )
       343│             )
       344│         except CalledProcessError as e:
    →  345│             raise EnvCommandError(e)
       346│
       347│         python_version = Version.parse(python_version.strip())
       348│         minor = "{}.{}".format(python_version.major, python_version.minor)
       349│         patch = python_version.text
```

I do have this version installed: 

```bash
❯ pyenv versions
  system
  3.6.1
  3.6.10
```

I tried to run using only `3.6`, but it doesn't work either: 

```bash
❯ poetry env use python3.6
pyenv: python3.6: command not found

The `python3.6' command exists in these Python versions:
  3.6.1
  3.6.10
  3.6.10/envs/raquel-3.6.10
  raquel-3.6.10

Note: See 'pyenv help global' for tips on allowing both
      python2 and python3 to be found.
```

If I set the local interpreter to a 3.6 version it seems to work but I guess that's not the expected behaviour:

```bash
❯ pyenv local 3.6.10
❯ poetry env use python3.6
Recreating virtualenv .venv
Using virtualenv: .venv
```

But I still have only a single environnement visible by poetry: 

```bash
❯ poetry env list
.venv (Activated)
```

And trying to switch back to the previous (3.8.0) will also fail.

I usually use `pyenv virtualenv` to handle my local environments.

PR #733, opened 2018-12-12
Doc: Clarify that add installs packages

alshain:
I was migrating from pipenv and already installed packages, so I cared about whether something would only touch my pyproject.toml file or actually install some packages.

I hope this helps others.


sdispater:
Thanks!

Issue #732, opened 2018-12-11
Which pip version does poetry use?


timonbimon:
- [x] I have searched the [issues](https://github.com/sdispater/poetry/issues) of this repo and believe that this is not a duplicate.
- [x] I have searched the [documentation](https://poetry.eustace.io/docs/) and believe that my question is not covered.

## Question
I was wondering how poetry chooses which pip version it uses when installing new packages? And how can we make it choose a different one?

I am asking this question because of the following problem I encountered:

While trying to install "cryptography" in a Docker container with poetry I got the following error message:

```
Command ['/root/.cache/pypoetry/virtualenvs/hybrid-py3.7/bin/pip', 'install'  
, '--no-deps', 'cryptography==2.4.2'] errored with the following output:      
Collecting cryptography==2.4.2                                                
  Downloading https://files.pythonhosted.org/packages/f3/39/d3904df7c56f8654  
691c4ae1bdb270c1c9220d6da79bd3b1fbad91afd0e1/cryptography-2.4.2.tar.gz (468k  
B)                                                                            
  Could not find a version that satisfies the requirement cffi!=1.11.3,>=1.7  
 (from versions: )                                                            
No matching distribution found for cffi!=1.11.3,>=1.7                         
You are using pip version 10.0.1, however version 18.1 is available.    
```

I found the error a little bit weird in the first place, because it does not occur when I run `poetry install` locally, but after a little bit of googling it seemed that the cffi problem was related to the pip version and should be fixable, by upgrading pip => which leads to my question: My dockerfile is based on python:3.7-alpine and has pip 18.1 installed, so why does poetry still try to use pip version 10.0.1?

Thanks a lot to anyone taking the time to read through this :)

Edit: To fix my issue, I just had to do `poetry run pip install --upgrade pip` before `poetry install`, but I guess my question where poetry gets the idea to use 10.0.1 from still holds? :)

eblume:
This is most likely related: https://github.com/sdispater/poetry/issues/712

asodeur:
I just ran into this as well and think it is a separate issue.

`venv` creates virtual environments using `ensurepip` which contain `pip==10.0.1` and `setuptools==39.0.1`. This explains where `pip==10.0.1` is comming from on Python >=3.3. 

[...]

PR #740, opened 2018-12-14
Upgrade pip/setuptools in venv to versions from base environment

asodeur:
# Pull Request Check List
Pedestrian fix for #732.

This PR upgrades pip/setuptools to the versions found in the base environment during `venv` creation. Hence, if pip/setuptools versions in your base environment can install a package, `poetry` can install it into the created virtual environment.

No attempt is made to resolve any build requirements declared in either `setup.py/setup_requires` or `pyproject.toml/[build-system]`.


Congee:
Any reason it's not merged? @sdispater

zyv:
We are having this problem right now, and are not sure what we can do at least as a stopgap fix. For some weird reason Poetry decided that from now on, it installs everything with `pip==19.0.3`, which doesn't work, whereas latest is 19.1.1 and the last known working is 18.1. We have pinned our system images to 19.1.1, but Poetry ignores that ATM.

@mm-matthias

stevegore:
Also having this issue. Can this please be merged? @sdispater ?

username-dot-com:
I could also use this fix. Is there any issue causing this to not be merged @sdispater ?

stevegore:
Current helper script to resolve virtualenv issues:

```bash
#!/bin/bash

echo "--- Checking appropriate prerequisites are installed."

# Check correct version of Python
python -V | grep 3.
if [ $? -ne 0 ]
then
    echo "Please ensure you have Python 3.6 activated. Pyenv is recommended."
    exit 1
fi

# Check Poetry installed
poetry --version | grep 0.12
if [ $? -ne 0 ]
then
    echo "--- Poetry not detected, installing..."
    curl -sSL https://raw.githubusercontent.com/sdispater/poetry/master/get-poetry.py | python
fi

# Do dry run to create virtual environment
echo "--- Using Poetry to create virtual environment"
poetry install --dry-run -q

# Activate created virtual environment
echo "--- Activating virtual environment"
VIRTUAL_ENV=$(poetry run python  -c "import os; print(os.environ['VIRTUAL_ENV'])")
echo "Virtual environment path is $VIRTUAL_ENV"
source $VIRTUAL_ENV/bin/activate
if [ $? -ne 0 ]
then
    echo "Could not activate the virtual environment!"
    exit 2
fi

echo "--- Updating pip and setuptools"
pip install -q --upgrade pip setuptools
deactivate

# Do the final install
echo "--- Installing dependencies"
poetry install
```

tests/fixtures/simple_project/pyproject.toml
```diff
@@ -1,25 +1,25 @@
...
-
-classifiers = [
-    "Topic :: Software Development :: Build Tools",
-    "Topic :: Software Development :: Libraries :: Python Modules"
-]
-
-# Requirements
-[tool.poetry.dependencies]
-python = "~2.7 || ^3.6"
+[tool.poetry]
```
brycedrennan: Whats changing in this file?

poetry/utils/env.py
```diff
@@ -318,6 +318,23 @@ def build_venv(cls, path):
...
+        # make sure pip and setuptools are on the same version as the base environment
+        from pip import __version__ as pip_version
+        from setuptools import __version__ as setuptools_version
+
+        the_venv = VirtualEnv(Path(path))
+        the_venv.run(
+            "python", "-m", "pip", "install", "--upgrade", "pip=={}".format(pip_version)
+        )
+        the_venv.run(
+            "python",
```
brycedrennan: This would mean that the creation of new environments would require internet access.

tests/fixtures/simple_project/pyproject.toml
asodeur: There should not be any changes, looks like somehow Windows line endings slipped in. I'll revert this.

poetry/utils/env.py
asodeur: Good point. I'll check what can be done. Since we are trying to update to the version available in the base environment there should be some way to get along w/o internet connection.

brycedrennan: I agree that it seems like there should be a way but the discussion on the python bug tracker seems to think its trickier than it sounds. I hope you find a way though!  I wonder if you should be making a pull request into `ensurepip` instead though.

https://bugs.python.org/issue30628

stevegore:
I believe this is still an issue for some dependencies. I have the use the following in my builds:

```
            if grep -q -E "(googleapis-common-protos|tensorboard|tensorflow)" poetry.lock
            then
              poetry run pip install -q --upgrade pip setuptools
            fi
```

Tensorflow, for example, has minimum pip requirements: https://www.tensorflow.org/install/pip

Understand however if this is beyond the scope of what poetry is attempting to resolve.

finswimmer:
In the meantime poetry has switched to virtualvenv for creating the venv's. This will make sure to get the latest pip.

Thanks a lot for your contribution anyway :+1: 

fin swimmer

abn:
Closed by #2666

Issue #746, opened 2018-12-17
Basic authentication fails if API redirects include username/password


connorbrinton:
- [ x ] I am on the [latest](https://github.com/sdispater/poetry/releases/latest) Poetry version.
- [ x ] I have searched the [issues](https://github.com/sdispater/poetry/issues) of this repo and believe that this is not a duplicate.
- [ x ] If an exception occurs when executing a command, I executed it again in debug mode (`-vvv` option).

- **OS version and name**: Mac OS Mojave
- **Poetry version**: 0.12.10

## Issue
I use a private PyPI server which supports basic HTTP authentication over HTTPS. I configure this server in Poetry with the following commands:
```
poetry config repositories.example https://pypi.example.com/reponame
poetry config http-basic.example "${USERNAME}" "${PASSWORD}"
```

In order to list package details on the server, Poetry performs a GET request with the `Authorization` header properly set. The server returns a JSON response which includes URLs pointing to additional package details. However, the URLs returned by the server are encoded with the username and password "baked-in" to the URL, as `https://<username>:<password>@pypi.example.com/reponame`. Ironically, this actually results in Poetry not sending any authentication when it attempts to follow the URLs. This results in an authentication failure, an empty sdist file and an extraction error from `TarFile.open`.

As a point of comparison, `pip` is able to handle this situation and successfully install packages from this repository.

### Cause
Poetry does not send authentication when following requests like those shown above because of [line 14 of `poetry/repositories/auth.py`](https://github.com/sdispater/poetry/blob/master/poetry/repositories/auth.py#L14), which checks to make sure that the `netloc` (which includes username, password, hostname and port) matches the repository's `netloc` before authenticating a request.

### Proposed Solutions
- Match based on `hostname` (provided by `urlparse/urllib.parse`) instead of `netloc`
- Strip alternative authentication from JSON API URLs
- Use alternative authentication from JSON API URLs if available before trying to use repository authentication

I would be happy to contribute a pull request if there is interest in any of the above solutions.

Thanks for working on Poetry! 😁

paymog:
My company has this problem too.

connorbrinton:
I went ahead and submitted a pull request that fixes this issue by matching based on `hostname` rather than `netloc`: #751

connorbrinton:
By the way, it looks like this is a regression caused by commit 6c5be4b806271822fe121f1c0a5e5303bc0c1629.

PR #751, opened 2018-12-18
Fix request authentication when credentials are included in URLs

connorbrinton:
Some package repositories which require basic HTTP authentication include the user's credentials in URLs returned as part of JSON responses. Because Poetry's request authentication system checks for a matching `netloc`, this ends up breaking basic HTTP authentication for such repositories. This pull request compares by `hostname` instead of `netloc`. Because `hostname` doesn't include any extra URL authentication information, this fixes access to these repositories.

This fixes issue #746.

## Pull Request Check List

- [ x ] Updated **documentation** for changed code.


sdispater:
Thanks!

Issue #761, opened 2018-12-24
Support generation of poetry manged setup.py file


abn:
- [x] I have searched the [issues](https://github.com/sdispater/poetry/issues) of this repo and believe that this is not a duplicate.

## Issue

As things stand today, poetry packages cannot be installed using `pip` or any other package managers from a VCS. This is useful when working with unreleased/unpublished versions of dependencies, where dependencies are managed by 'pip' or another tool. This was briefly touched on in #34 for a similar use case.

With this in mind, it could be great if poetry provided a solution, in a limited capacity, similar to what is implemented by [poetry-setup](https://pypi.org/project/poetry-setup/). In order to achieve this, poetry could provide a command, eg: `poetry setuptools:setupfile`, that generates a usable `setup.py` file using the mechanism used today by `SdistBuilder`.

In theory, this could be migrated to be a plugin once #693 is implemented.

As an example. this, if implemented, could be used by projects as a pre-commit hook to generate `setup.py` file on commit. This should allow projects using `pip` etc. to install unpublished projects developed using poetry (if the maintainer wants it).

In addition to the above use case, this also allows for tools like PyCharm etc., to auto detect requirements and metadata.

Can follow this up with a proposed implementation PR.

Relates-to: #34

webknjaz:
@abn other way to do it is to actually publish each commit to Test PyPI :)

abn:
@webknjaz that would definitely work for a subset of use cases. The generation of `setup.py` and treating it similar to a lock file allows for a broader range of compatibility.

cjw296:
Won't pip 19.0 remove the need for this? (as it understands pyproject.toml) - you'll need to add some manual stuff to it though, until https://github.com/sdispater/poetry/issues/744 is resolved.

abn:
@cjw296 I will see if I can validate that, if pip understands poetry projects, that will most likely cover a large chunk of the use cases. However, without this change, the other tools need to either support PEP-518 or poetry projects explicitly. With that in mind, I do think this is still useful even if pip support is completed.

radek-sprta:
[Snaps](https://snapcraft.io/) require [setuptools and setup.py](https://docs.snapcraft.io/python-apps/6741), at least for now, so having way to generate setup.py from poetry would help with that.

sdll:
Any updates on that?

stevegore:
[dephell](https://github.com/dephell/dephell) will do this for you:

[...]

PR #762, opened 2018-12-25
Add support for setup.py as an export format

abn:
This change introduces `setup.py` as a valid export format.

Resolves: #761

### DIFF
--- a/docs/docs/cli.md
+++ b/docs/docs/cli.md
@@ -433,7 +433,7 @@ poetry export -f requirements.txt > requirements.txt
 
 !!!note
 
-    Only the `requirements.txt` format is currently supported.
+    Only the `requirements.txt` and `setup.py` format is currently supported.
 
 ### Options
 
--- a/poetry/console/commands/export.py
+++ b/poetry/console/commands/export.py
@@ -13,16 +13,22 @@ class ExportCommand(Command):
     options = [
         option("format", "f", "Format to export to.", flag=False),
         option("output", "o", "The name of the output file.", flag=False),
-        option("without-hashes", None, "Exclude hashes from the exported file."),
-        option("dev", None, "Include development dependencies."),
+        option(
+            "without-hashes", None, "Exclude hashes from the exported file if used."
+        ),
+        option("dev", None, "Include development dependencies if configurable."),
         option(
             "extras",
             "E",
-            "Extra sets of dependencies to include.",
+            "Extra sets of dependencies to include if configurable.",
             flag=False,
             multiple=True,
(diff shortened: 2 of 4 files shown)


stefansedich:
Any reason this one has not had any movement on it? would be useful to me right now.

abn:
@stefansedich i do think that the functionality still has value, but I am no longer sure if this is the right approach for this. I would like to attempt rewriting it so that it is a format for the export command.

jwiede:
If this is going to sit for a while, could we at least get it as a "preview" version for those of us who need it (PyCharm users and/or those not willing to deal with dephell equivalent)?

abn:
@jwiede thanks for bumping this I'll see if i can get this cleaned up and ready for review again. Not sure where the project team sits on this one though.

vegarsti:
+1 on something like this being useful!

jacques-:
Would be incredibly useful to me as well, +1

finswimmer:
> Would be incredibly useful to me as well, +1

Why? :smile:

jacques-:
So it's a bit of a long story but here goes. We were using poetry for our packages and apps, but because poetry install takes > 20mins to update deps now we want to use poetry only for modules, and use pipenv to manage apps (see https://github.com/python-poetry/poetry/issues/2094#issuecomment-646323089  - we don't publish apps as modules an  pipenv installs from lock much faster it saves debugging time). As our local debugging flow builds app in docker, this install time from lock really matters to us as well. However, ran into another problem yesterday that pipenv doesn't support adding editable deps that don't have setup.py so breaks our cludge temporary fix.

Ultimately the pipenv thing was intended as a temporary measure that would save some time, hoping that performance of poetry would improve in future. We really like poetry and it is awesome everywhere else, it's just that in some projects with lots of deps it is making our dev flow very slow.

sdispater:
@jacques- Regarding the speed of installation, you might want to take a look at #2595 which improves performances significantly and will be available in the next feature release.

jacques-:
Awesome, waiting with great anticipation for next release!

Do you think https://github.com/python-poetry/poetry/pull/2149 will make it in as well @sdispater ?

teto:
this looks useful to me to enter editable mode. Why closing it ?

abn:
@teto you can create a setup shim with the following content for that.

```py
import setuptools; setuptools.setup()
```

Additionally, poetry is in the process of removing the internal supprot for generating `setup.py` files. If such an export is desirable, it can be implemented by the community via the upcoming pluggin system as a plugin to the export command.

Issue #764, opened 2018-12-28
exclude option fails on paths with capital letters


TOffergeld:
- [x] I am on the [latest](https://github.com/sdispater/poetry/releases/latest) Poetry version.
- [x] I have searched the [issues](https://github.com/sdispater/poetry/issues) of this repo and believe that this is not a duplicate.
- [ ] If an exception occurs when executing a command, I executed it again in debug mode (`-vvv` option).

- **OS version and name**: Win10 64-bit, Py3.7
- **Poetry version**: 0.12.10
- **Link of a [Gist](https://gist.github.com/) with the contents of your pyproject.toml file**: https://gist.github.com/TOffergeld/89d5fa517edbd51ccc8eac518f5e2035
## Issue
There seems to be an issue with the exclude section in the pyproject.toml files with paths containing capital letters.
The issue can be reproduced with the example .toml file and a directory setup as follows:

```
poetry_example
-- __init__.py
-- Foo
------ Bar.py
------ lowercasebar.py
-- bar
------ foo.py
------ Capitalfoo.py
```

`poetry build -vv` yields:
```
Building poetry_example (0.1.0)
 - Building sdist
 - Adding: poetry_example\__init__.py.py
 - Adding: poetry_example\bar\Capitalfoo.py
 - Adding: poetry_example\Foo\Bar.py
 - Adding: poetry_example\Foo\lowercasebar.py
 - Adding: pyproject.toml
 - Built poetry_example-0.1.0.tar.gz

 - Building wheel
 - Adding: [...]\poetry_example-0.1.0\poetry_example\__init__.py.py
 - Adding: [...]\poetry_example-0.1.0\poetry_example\bar\Capitalfoo.py
 - Adding: [...]\poetry_example-0.1.0\poetry_example\Foo\Bar.py
 - Adding: [...]\poetry_example-0.1.0\poetry_example\Foo\lowercasebar.py
 - Built poetry_example-0.1.0-py3-none-any.whl
```
in which only one (bar/foo.py) of the four files specified to be excluded actually gets excluded.

The same issues also seems to affect exclusions through the vcs.

TOffergeld:
So the issue obviously is that Path.glob() returns case-insensitive files on a case-sensitive fs (such as Windows).
Hence we need to fix the paths back to their original case (and make sure this doesn't happen again anywhere else in the build process).

sdispater:
This is fixed in the latest `0.12.11` release (commit 27d6fcf).

PR #783, opened 2019-01-07
Fix exclude with capital letters (#764)

TOffergeld:
# Pull Request Check List

I have added test coverage for #764 along with a proposed solution for the issue.
However, there are a few edge cases where the solution will fail to correctly determine the files to be excluded. Namely, if one or more wildcards are contained in the string and a subsection (determined by ```str.split("*")``` is contained in the string with different capitalization (which for the time being raises a NotImplementedError, however I'm unsure whether this is appropriate).

I'm happy to make adjustments to this pr if required.


TOffergeld:
Marking this WIP until I get the CI to pass.

TOffergeld:
ToDo:

- [ ] Remove pathlib dependency for python<3.4
- [ ] Handle ```test_builder_find_invalid_case_sensitive_excluded_files``` for case-insensitive filesystems (see Travis jobs)

TOffergeld:
@sdispater I think I'm at a point where this can be reviewed - the test failures seem to be unrelated (same as in c640aaffd6db5ab65247cd9c009997559f41bf48 as far as I can tell).

Currently a path exclusion like ```/bar/*/Bar/*.py``` will cause a NotImplementedError as it seems somewhat tricky to determine that the first ```bar``` should **not** be replaced by ```Bar``` but the second does. I'm happy to try to come up with something if you think this functionality is worth adding.

PR #801, opened 2019-01-16
Adds support for adding package from remote urls

### LINKED ISSUE #695: Install of a package from URL

## Issue

Sometimes packages easily avaliable only by url, for example [binaries for Windows](https://www.lfd.uci.edu/~gohlke/pythonlibs/). It would be nice to be able to do something like:

`poetry add pycurl --url https://download.lfd.uci.edu/pythonlibs/h2ufg7oq/pycurl-7.43.1-cp37-cp37m-win_amd64.whl`

For example, pipenv [supports it](https://github.com/pypa/pipenv/issues/480#issuecomment-330605268) :)

canassa:

This PR adds support for installing packages from remote URLs, by passing a `--url` parameter. e.g.:

`poetry add minibar --url https://github.com/canassa/minibar/archive/master.zip`

I really needed this feature in order to keep using poetry in my internal company projects, therefore I hacked this PR together yesterday. This is really a proof-of-concept and not something production grade, I am opening this PR because I would like to get some feedback about this before I invest more time in it. Some question that I have are:

- Is this a welcomed feature?
- Do you agree with the new `--url` parameter, or do you prefer something else?
- Did I took the correct approach by subclassing `Dependency` and implementing the `search_for_url` that download the file and calls the `search_for_file` method?

Fixes #695 

Thanks!

### DIFF
--- a/poetry/puzzle/provider.py
+++ b/poetry/puzzle/provider.py
@@ -207,7 +207,7 @@ class Provider:
         return [package]
 
     def search_for_url(self, dependency):  # type: (UrlDependency) -> List[Package]
-        tmp_dir = Path(mkdtemp(prefix="pypoetry-git-{}".format(dependency.name)))
+        tmp_dir = Path(mkdtemp(prefix="pypoetry-url-{}".format(dependency.name)))
 
         # TODO: The file name might be in the request header itself
         file_name = urlparse(dependency.url).path.split("/")[-1]
--- a/poetry/puzzle/provider.py
+++ b/poetry/puzzle/provider.py
@@ -218,13 +218,21 @@ class Provider:
         with tmp_file.open("wb") as f:
             shutil.copyfileobj(response.raw, f)
 
-        file_dep = FileDependency(
+        file_dependency = FileDependency(
             dependency.name,
             tmp_file,
             category=dependency.category,
             optional=dependency.is_optional(),
         )
-        return self.search_for_file(file_dep)
+        for extra in dependency.extras:
+            file_dependency.extras.append(extra)
+
+        package = self.search_for_file(file_dependency)[0]
+
+        package.source_type = "url"
+        package.source_url = dependency.url
+
+        return [package]
 
     def search_for_file(self, dependency):  # type: (FileDependency) -> List[Package]
         if dependency.path.suffix == ".whl":
@@ -485,6 +493,7 @@ class Provider:
             "directory",
             "file",
             "git",
+            "url",
         }:
             package = DependencyPackage(
                 package.dependency,
--- a/poetry/puzzle/provider.py
+++ b/poetry/puzzle/provider.py
@@ -1,10 +1,15 @@
 import glob
 import logging
 import os
+import shutil
+
 import pkginfo
 import re
 import time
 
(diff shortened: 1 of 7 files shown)


jaysonsantos:
I think you would have to implement your own dependency because if you use file's it would still reference your local file and if you distribute the lock you would have problems reproducing the install on a remote system for example.

Git dependency will pass all informations needed to reproduce on the remote system.
Here is an example of lock file using minibar with --git:

```
[[package]]
category = "main"
description = "Simple text progress bar library"
name = "minibar"
optional = false
python-versions = "*"
version = "0.4.0"

[package.source]
reference = "25484a383790ba09df69b4e1a6d65e018743326c"
type = "git"
url = "https://github.com/canassa/minibar"

```

jaysonsantos:
Ah, git won't propagate hashes to metadata but IMO url should do it as file does

canassa:
@jaysonsantos Thanks! I fixed the issue

canassa:
I noticed that my implementation doesn't work for all kinds of files. For example, it works if I point the url to an archive that was generated with `poetry build` but it doesn't work if I try to point to a tarball with only the source code. For example, this works:

`poetry add minibar --url https://files.pythonhosted.org/packages/12/34/f1d8396d246f9649e164208d9c28e3c2bdb91077c59f86368b10fba36324/minibar-0.5.0.tar.gz`

But this doesn't work:

`poetry add minibar --url https://github.com/canassa/minibar/archive/0.4.0.tar.gz`

canassa:
A friendly bump 😉 

I really would like to see this feature merge, please let me know if there is anything else I could do to help move this forward.

poetry/puzzle/provider.py
```diff
@@ -198,6 +206,34 @@ def search_for_vcs(self, dependency):  # type: (VCSDependency) -> List[Package]
 
         return [package]
 
+    def search_for_url(self, dependency):  # type: (UrlDependency) -> List[Package]
+        tmp_dir = Path(mkdtemp(prefix="pypoetry-git-{}".format(dependency.name)))
```
romaia: maybe pypoetry-url-{}?

canassa: Thanks! fixed

canassa:
I have been using this branch for a while now and so far so good!

The only problem that I had is that while this is working fine if I put the dependency in the `[tool.poetry.dependencies]` section, it doesn't seem to work if I put it in the `[tool.poetry.dev-dependencies]` section. I still didn't have the chance to debug that

sklarsa:
What is the roadmap for getting this onto master?

canassa:
@sklarsa  I think that it would at least require that the issue that I mentioned in my previous comment to be fixed. Other than that I think the PR works as I used my fork for quite a while without issues. Unfornutely, installing a poetry fork is a very cumbersome process and I don't want to force my coworkers into having to install a custom fork just to run my project.

Given that I haven't received any feedback from a maintainer in 2 months, I just assumed that this project is currently dead and I moved back to pip, therefore I won't be updating this PR any longer.

drunkwcodes:
This feature overlaps with `git clone` and `wget`.

Issue #817, opened 2019-01-20
Missing version names of a package cause poetry to crash


dvf:
- [X] I am on the [latest](https://github.com/sdispater/poetry/releases/latest) Poetry version.
- [X] I have searched the [issues](https://github.com/sdispater/poetry/issues) of this repo and believe that this is not a duplicate.
- [X] If an exception occurs when executing a command, I executed it again in debug mode (`-vvv` option).

- **OS version and name**: macOS 10.14.2
- **Poetry version**: 0.12.11
- **Link of a [Gist](https://gist.github.com/) with the contents of your pyproject.toml file**: https://gist.github.com/dvf/6c749cbad7d362e716361a0ca17f8d4e

## Issue
Poetry crashes when trying to install `some-private-dep` pinned at a parsable version (1.1.5). 

**There are missing versions of this private package, causing Poetry to be calling `Version.parse` with `text = None`**

Here's the Traceback:

```python-traceback
~/P/brxp ❯❯❯ poetry add some-private-dep==1.1.5 -vvv
Using virtualenv: /Users/dvf/Library/Caches/pypoetry/virtualenvs/someproject-py3.6
                                        
[TypeError]              
expected string or bytes-like object  
                                        
Exception trace:
 /Users/dvf/.poetry/lib/poetry/_vendor/py3.6/cleo/application.py in run() at line 94
   status_code = self.do_run(input_, output_)
 /Users/dvf/.poetry/lib/poetry/console/application.py in do_run() at line 88
   return super(Application, self).do_run(i, o)
 /Users/dvf/.poetry/lib/poetry/_vendor/py3.6/cleo/application.py in do_run() at line 197
   status_code = command.run(input_, output_)
 /Users/dvf/.poetry/lib/poetry/console/commands/command.py in run() at line 77
   return super(BaseCommand, self).run(i, o)
 /Users/dvf/.poetry/lib/poetry/_vendor/py3.6/cleo/commands/base_command.py in run() at line 146
   status_code = self.execute(input_, output_)
 /Users/dvf/.poetry/lib/poetry/_vendor/py3.6/cleo/commands/command.py in execute() at line 107
   return self.handle()
 /Users/dvf/.poetry/lib/poetry/console/commands/add.py in handle() at line 69
   packages, allow_prereleases=self.option("allow-prereleases")
 /Users/dvf/.poetry/lib/poetry/console/commands/init.py in _determine_requirements() at line 244
   allow_prereleases=allow_prereleases,
[...]

PR #818, opened 2019-01-20
Check for version of 'None'

dvf:
- [ ] Added **tests** for changed code. **I wasn't sure how to write tests for this, any help would be most appreciated!**
---

I haven't added tests because I wasn't sure how to test for something like this, any help would be appreciated 🙏. Linked to #817

### DIFF
--- a/poetry/semver/version.py
+++ b/poetry/semver/version.py
@@ -196,11 +196,12 @@ class Version(VersionRange):
 
     @classmethod
     def parse(cls, text):  # type: (str) -> Version
+        match = COMPLETE_VERSION.match(text)
         try:
             match = COMPLETE_VERSION.match(text)
         except TypeError:
             raise ParseVersionError('Unable to parse "{}".'.format(text))
-        
+
         if match is None:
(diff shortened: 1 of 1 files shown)


poetry/semver/version.py
```diff
@@ -197,6 +197,11 @@ def include_max(self):
     @classmethod
     def parse(cls, text):  # type: (str) -> Version
         match = COMPLETE_VERSION.match(text)
```
DanCardin: You'll need to delete this line first, it seems

dvf: Wait why? It's only failing in Python 2.7

DanCardin: What I mean to say is that, line 199 `match = COMPLETE_VERSION.match(text)` is the same as line 201 inside the `try`/`catch`.

So when `.match` raises a `TypeError` it never makes it to the `try`/`catch`ed version.

I deleted line 199, and it appeared to fix https://github.com/sdispater/poetry/issues/837

DanCardin:
this one was superceded by https://github.com/sdispater/poetry/pull/917, so can be closed

Issue #820, opened 2019-01-21
`poetry config settings.virtualenvs.path` returns quoted string


naringas:
- [x] I am on the [latest](https://github.com/sdispater/poetry/releases/latest) Poetry version.
- [x] I have searched the [issues](https://github.com/sdispater/poetry/issues) of this repo and believe that this is not a duplicate.
- No exception

- **OS version and name**: irrelevant, but is MacOS High Sierra 10.13.6
- **Poetry version**: 0.12.11
- not `pyproject.toml` related

## Issue

the command: `poetry config settings.virtualenvs.path` returns the path but with quotes.
Therefore it is not possible to do:
```
cd $(poetry config settings.virtualenvs.path)
```
because bash can't handle the quotes

naringas:
see PR

PR #821, opened 2019-01-21
removes quotes from config command when returning single setting value

naringas:
# Pull Request Check List

based on develop, see issue #820


poetry/console/commands/config.py
```diff
@@ -130,7 +130,11 @@ def handle(self):
                 )
 
                 for value in values:
-                    self.line(value[1])
+                    if isinstance(value[1], str) and value[1][0] == value[1][-1] == '"':
```
drunkwcodes: Would you like to add ws check in here?

naringas: sorry, what do you mean by "ws"?

drunkwcodes: White space characters.

naringas: I cannot get any command such as `cd $(poetry config settings.virtualenvs.path)` to work no matter what I do.

if `poetry config settings.virtualenvs.path` returns the strings with whitepsace escaped, e.g. a setting with the vaule `/some path/dir with space` returns `/some\ path/dir\ with\ space` usring the command inside like `ls $(...)` still doesn't work.

So I don't know how to proceed. I guess I will drop the whole thing.

naringas:
no good

PR #716, opened 2018-12-07
Use 'python -m pip' to run pip

uSpike:
Should fix #712

I'm not sure how to test this... or if we need to?

Issue #712, opened 2018-12-06
poetry appears to be trying to install with the wrong version of pip


nottrobin:
<!--
  Hi there! Thank you for discovering and submitting an issue.

  Before you submit this; let's make sure of a few things.
  Please make sure the following boxes are ticked if they are correct.
  If not, please try and fulfill these first.
-->

<!-- Checked checkbox should look like this: [x] -->
- [x] I am on the [latest](https://github.com/sdispater/poetry/releases/latest) Poetry version.
- [x] I have searched the [issues](https://github.com/sdispater/poetry/issues) of this repo and believe that this is not a duplicate.
- [x] If an exception occurs when executing a command, I executed it again in debug mode (`-vvv` option).

<!--
  Once those are done, if you're able to fill in the following list with your information,
  it'd be very helpful to whoever handles the issue.
-->

- **OS version and name**: Docker image: ubuntu:xenial (as root)
- **Poetry version**: 0.12.10
- **Link of a [Gist](https://gist.github.com/) with the contents of your pyproject.toml file**: https://gist.github.com/nottrobin/e104b9ec140cae4f403f381ed71b3307

## Issue
<!-- Now feel free to write your issue, but please be descriptive! Thanks again 🙌 ❤️ -->

In my system, I have both `python` and `pip`, and `python3` and `pip3` commands. (Installed with `apt install python-pip python3-pip`).

I installed poetry with `pip3 install poetry`, and you can see that it's using `python3` as its binary:

``` bash
$ head -n 1 `which poetry`
#!/usr/bin/python3
```

I've also configured it not to create virtualenvs, and I'm running this as root inside a `ubuntu:xenial` docker container.

So now my `poetry install` errors, because it claims not to have python3.5:

``` bash
# poetry install -vvv
Skipping virtualenv creation, as specified in config file.
Installing dependencies from lock file


Package operations: 1 install, 0 updates, 0 removals, 1 skipped

  - Skipping pytz (2018.7) Already installed
  - Installing django (2.1.4)

[EnvCommandError]
Command ['/usr/bin/pip', 'install', '--no-deps', 'django==2.1.4'] errored with the following output:                                           
Collecting django==2.1.4                                                                                                                       
  Using cached https://files.pythonhosted.org/packages/83/f7/4939b60c4127d5f49ccb570e34f4c59ecc222949220234a88e4f363f1456/Django-2.1.4.tar.gz  
    Complete output from command python setup.py egg_info:                                                                                     
[...]

PR #831, opened 2019-01-24
Commonize pip running logic and use system pip in system env

K900:

This is a more complete fix for #712 and #716 - there's now a common function to run the appropriate pip for an environment with given arguments, and all (AFAICT) locations where pip is invoked inside Poetry are changed to use it. It's technically a behavior change for cases where Poetry is running outside a virtual environment _and_ with a non-default Python, but I'm not sure if those are common enough to consider this a serious breaking change. The behavior inside virtualenvs is unaffected.


K900:
For a bit more context, this allows using Poetry to install stuff in a system environment for non-system Pythons, e.g. `pip3 install poetry && poetry install` will do the right thing in a CentOS container where `pip` is always a symlink to `pip2`.

PR #861, opened 2019-02-04
Commonize pip running logic and use system pip in system env

K900:

This is a more complete fix for #712 and #716 - there's now a common function to run the appropriate pip for an environment with given arguments, and all (AFAICT) locations where pip is invoked inside Poetry are changed to use it. It's technically a behavior change for cases where Poetry is running outside a virtual environment _and_ with a non-default Python, but I'm not sure if those are common enough to consider this a serious breaking change. The behavior inside virtualenvs is unaffected.

For a bit more context, this allows using Poetry to install stuff in a system environment for non-system Pythons, e.g. `pip3 install poetry && poetry install` will do the right thing in a CentOS container where `pip` is always a symlink to `pip2`.

This PR is identical to #831, except now on its own branch

### DIFF
--- a/poetry/installation/pip_installer.py
+++ b/poetry/installation/pip_installer.py
@@ -115,7 +115,7 @@ class PipInstaller(BaseInstaller):
             raise
 
     def run(self, *args, **kwargs):  # type: (...) -> str
-        return self._env.run("python", "-m", "pip", *args, **kwargs)
+        return self._env.run_pip(*args, **kwargs)
 
     def requirement(self, package, formatted=False):
         if formatted and not package.source_type:
--- a/poetry/masonry/builders/editable.py
+++ b/poetry/masonry/builders/editable.py
@@ -31,16 +31,14 @@ class EditableBuilder(Builder):
 
         try:
             if self._env.pip_version < Version(19, 0):
-                self._env.run("python", "-m", "pip", "install", "-e", str(self._path))
+                self._env.run_pip("install", "-e", str(self._path))
             else:
                 # Temporarily rename pyproject.toml
                 shutil.move(
                     str(self._poetry.file), str(self._poetry.file.with_suffix(".tmp"))
                 )
                 try:
-                    self._env.run(
-                        "python", "-m", "pip", "install", "-e", str(self._path)
-                    )
+                    self._env.run_pip("install", "-e", str(self._path))
                 finally:
                     shutil.move(
                         str(self._poetry.file.with_suffix(".tmp")),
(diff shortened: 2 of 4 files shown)


K900:
I'm not sure why the build is failing here, but I'm pretty sure it's not my fault

K900:
@sdispater sorry for pinging you directly, but any chance you could take a look at this and #862? These are two pretty small fixes that make things a lot easier for people who want to use Poetry in a CI environment

K900:
Rebased on the latest master

K900:
Rebased again, and also adjusted the 0.12.14 `EditableBuilder` changes to match the behavior everywhere else. This is also technically a breaking behavior change, but the previous behavior also made no sense.

brycedrennan:
@K900 thanks for contributing! sorry for the delay.  

It looks like the underlying issues you referenced may now be resolved.  Do you think this PR is still needed?

K900:
I am not 100% sure, to be honest. Do you mean the 1.0 branch or 0.12? I can re-test on both tomorrow.

K900:
All right, so I've built a small reproducible example of the behavior [here](https://github.com/K900/poetry-861-example). The master branch is for Poetry 0.12.17, the `poetry-1.0` branch is for 1.0.0b1, both exhibit the same behavior.

The idea is that we're in a Docker environment, building a container that will include our dependencies (managed by Poetry, of course) installed directly into the system, without a virtualenv (because we don't need one).

We install Poetry itself into the "right" Python environment, and expect it to behave similarly to pip - that is, to manage the environment that it itself is installed in.

However, what Poetry actually does is call `python -m pip install` etc., which on our mixed-Python system refers to Python 2, and not 3, and we end up with the dependencies installed on the wrong Python version (if installed at all).

K900:
Rebased the commits on the latest master again. No changes this time.

brycedrennan:
Thanks I'll take a look!

K900:
Any updates on this? I'd love to get this in before 1.0, so there's no issues with breaking behavior...

K900:
Rebasing to latest master, again

K900:
Rebased to the latest master (which is now even with develop I think?), again. I've got some failing tests locally, but those seem to fail on unchanged master as well...

K900:
Rebased on the latest master. This actually reduces the amount of changes a bit, thanks to importlib-metadata implicitly doing the right thing in places where we had to parse `pip freeze` before.

sdispater:
Thanks a lot!

PR #862, opened 2019-02-04
Fix returning error code 0 on HTTP errors in poetry publish

K900:
Cleo checks `errno`, which is set on Requests errors (since they're a subclass of `IOError`). So, catch the error here instead and fail directly with exit code 1. This fixes running `poetry publish` in a CI environment. Another solution might be to handle the `e.errno is None` case explicitly in Cleo.

### DIFF
--- a/poetry/console/commands/publish.py
+++ b/poetry/console/commands/publish.py
@@ -1,3 +1,5 @@
+from requests import HTTPError
+
 from .command import Command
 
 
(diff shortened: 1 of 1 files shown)


mockersf:
I have the same issue with running `poetry publish` in my CI, and this fix works for me

mockersf:
workaround: in your CI, use
```
poetry publish | tee >(cat) | grep succeeded
```
`| tee >(cat)` will output the logs to stdout, and `| grep succeeded` will return an exit code different that 0 if succeeded is not found

K900:
Rebased onto latest master. @sdispater is this a better fit for develop perhaps, considering it changes behavior?

K900:
This was actually fixed in 903514, which I completely missed (sorry!), so I'll just close this.

PR #870, opened 2019-02-06
Make sdist determinisitic by setting gzip mtime to 0

achow101:
A GZip file contains a timestamp with it's last modification time. This timestamp by default is the current time which will make `poetry build` produce non-deterministic sdist archives. To make sdist archives deterministic, this timestamp must be set to a fixed time. So this PR sets that time to 0.

- (N/A) Updated **documentation** for changed code.

### DIFF
--- a/poetry/masonry/builders/sdist.py
+++ b/poetry/masonry/builders/sdist.py
@@ -58,7 +58,7 @@ class SdistBuilder(Builder):
         target = target_dir / "{}-{}.tar.gz".format(
             self._package.pretty_name, self._meta.version
         )
-        gz = GzipFile(target.as_posix(), mode="wb")
+        gz = GzipFile(target.as_posix(), mode="wb", mtime=0)
         tar = tarfile.TarFile(
             target.as_posix(), mode="w", fileobj=gz, format=tarfile.PAX_FORMAT
         )
--- a/tests/masonry/builders/test_sdist.py
+++ b/tests/masonry/builders/test_sdist.py
@@ -1,5 +1,6 @@
 # -*- coding: utf-8 -*-
 import ast
+import gzip
 import shutil
 import tarfile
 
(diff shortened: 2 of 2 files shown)


brycedrennan:
Interesting.  Can you fix the formatting?

achow101:
> Interesting. Can you fix the formatting?

Done.

poetry/masonry/builders/sdist.py
```diff
@@ -64,7 +64,7 @@ def build(self, target_dir=None):  # type: (Path) -> Path
         target = target_dir / "{}-{}.tar.gz".format(
             self._package.pretty_name, self._meta.version
         )
-        gz = GzipFile(target.as_posix(), mode="wb")
+        gz = GzipFile(target.as_posix(), mode="wb", mtime=0)
```
finswimmer: Hello @achow101,

thanks a lot for your contribution. Fixing this looks necessary to me. Are there any reasons for settings the mtime to 0 and not to the current time via `time.time()`?

fin swimmer

achow101: It's literally the whole point of this PR. A fixed time is needed otherwise the resulting archive is non-deterministic. Setting to `time.time()` means that the timestamp will continue to be variable which is what I'm trying to fix.

finswimmer: Ah , sorry it was obviously to late yesterday. The current timestamp is already used if the parameter is omitted and I thought it is something similar to the already merged #1541.

So let's change my question :) Why do you want the file to be deterministic?

achow101: I subscribe to the idea that the releases that you publish should be built in such a way that other people can repeat the build process and produce the exact same files, down to the bit. This makes it much easier to audit the published binaries/archives. https://reproducible-builds.org/ basically explains why.

There's also a bunch of things already in poetry which make the build results more reproducible, so reproducibility is clearly something that is intended in this project.

sdispater:
Thanks for your contribution!

This looks reasonable to me. Could you rebase your changes onto `master`?

achow101:
Rebased

abn:
Apoligies for dropping the ball on this one. Feel free to ping on discord if something falls through the crack again. We appricate your contributions, issues reports, PRs and doc fixes alike.

@achow101 this needs to move `poetry-core` now.

abn:
@achow101 closing this here as this needs to go to poetry-core now anyway.

achow101:
Moved to poetry-core: https://github.com/python-poetry/poetry-core/pull/105

PR #906, opened 2019-02-20
Set the virtualenv directory name via configuration

simoncoulton:
This is to solve a use case I came across when working within a Docker container while mounting a volume from my local machine. Local machine is running OSX, container is running Alpine. My local has the .venv folder in the application directory so that VS Code can use it for autocompletion, however if that folder exists Poetry will fail to even `in-project` and `create` are both set to `false` (a virtualenv is not required within the isolated container).

An alternative solution would be to move the `create_venv` section above the check for a local virtualenv, however I felt this was more flexible.

### DIFF
--- a/docs/docs/configuration.md
+++ b/docs/docs/configuration.md
@@ -22,6 +22,11 @@ Defaults to `true`.
 Create the virtualenv inside the project's root directory.
 Defaults to `false`.
 
+### `settings.virtualenvs.name`: string
+
+The name of the virtualenv folder that will be created.
+Defaults to `.venv`.
+
 ### `settings.virtualenvs.path`: string
 
 Directory where virtualenvs will be created.
--- a/poetry/utils/env.py
+++ b/poetry/utils/env.py
@@ -161,13 +161,15 @@ class Env(object):
         in_venv = os.environ.get("VIRTUAL_ENV") is not None
 
         if not in_venv:
+            config = Config.create("config.toml")
+
             # Checking if a local virtualenv exists
-            if (cwd / ".venv").exists():
-                venv = cwd / ".venv"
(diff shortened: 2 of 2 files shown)


simoncoulton:
@sdispater any word if something like this sort of functionality is going to be available? It's just adding the ability to specify a name for the .venv folder via a configuration setting.

brycedrennan:
I'm not familiar with VSCode. Does it really require the virtual env be inside the project folder?

simoncoulton:
@brycedrennan doesn't require the venv to be within the folder, the issue is that I'm mounting the source folder in Docker with the binary being created for Alpine, which is different than the development environment. The autocompletion won't work as the binaries are for different operating systems. In order to get around that, I do a `poetry install` outside of the container in my local environment, however because I can't change the folder name of virtualenv that is created by Poetry, I have to rename the original folder (that was created during the containers CMD call), do the poetry install, rename that to .env, rename the original back to .venv, and set the Python path to the .env.

Perhaps this is just a particular use case that I've come up against during development, but it probably doesn't hurt having that folder name be configurable anyway.

abn:
@simoncoulton I am walking through the older PRs filtering out relevant ones. Can I get some clarity around what te need here really is?

In typical cases you really do not want your local (host machine) virtualenv to be mounted in unless the container and host environment are similar (ie. same ABI, same OS etc.). In most cases I have seen, `.venv` is ignored via the `.dockerignore` file anyway. An alternative is to use a build 

That said, having the in-project virtual environment directory configurable does not sound unreasonable. I am just unsure how common the use case is. And poetry should recreate the environment if it cannot use the `python` binary anyway.

neersighted:
I'm going to close this in favor of #2601 for now -- please comment there if the proposed design would work for your use case.

PR #908, opened 2019-02-21
Improve support for alternative package repositories

sdispater:
This PR aims at improving support for alternative repositories.

It introduces a way to specify which repository to use when looking for a dependency by providing a `source` like the following:

```toml
[tool.poetry.dependencies]
pendulum = {version = "^2.0", source = "custom-pypi"}

[[tool.poetry.source]]
name = "custom-pypi"
url = "https//example.com/simple"
```

## TODO

- [x] Add support for specifying a source for a specific dependency
- [x] Improve installation from alternative repositories

## Pull Request Check List


HolgerPeters:
Since this is a draft, maybe its not too late to add some comment.

First of all, I like that you can configure from which repository a package should be pulled.

I feel like what I want in my daily life is probably a solution that is similar to what maven does for the Java world.  Maven allows you to

1. configure several repositories and give them a name with an identifier.
2. configure mirrors for these repositories.

The config is a bit verbose and spread over pom.xml (for the repositories) and `settings.xml` (https://maven.apache.org/guides/mini/guide-mirror-settings.html), but I think one should be able to map it to nice toml syntax in principle.

# Examples

## Company Setup

Company uses packages from pypi and a local artifact store that contains their local builds, `local-index`. They run a pypi-cache on a local artifactory server, but if that service is unavailable they are perfectly happy to pull packages from pypi. So they would define two remote pypi repositories, one for pypi, one for `local-index`. For pypi, they define a `local-pypi-mirror`.

## Self-built packages

A developer wants to have all packages built from source on their machines (so that they can profit from optimizations). They run a local [devpi](https://devpi.net/) service, and upload packages to an index on that devpi using `devpi-builder`. They don't want a fallback to remote pypi, because that would mean pulling in artifacts that did not go through their build process.

PR #929, opened 2019-03-02
Support double dash in run command arguments

### LINKED ISSUE #915: poetry run: unclear precedence of global vs script arguments

- **Arch Linux**
- **Poetry version 1.0.0a2**

## Issue

When running `poetry run <script> <script args>`, at least the global argument `-h` takes precidence over the script arguments. I have a project using argparse with valid arguments `-h` and `--help`. Neither of these arguments work with `poetry run ./example.py <argument>`. What does work is using `--h` (notice the double dash instead of a single one). The same problem occurs when the entry point is set in `[tool.poetry.scripts]`.

Minimal code example `test.py` that just outputs argv
```test.py
#!/usr/bin/env python
import sys
print(sys.argv)
```

Output of `poetry run ./test.py -v` is as expected: `['./test.py', '-v']`
Output of `poetry run ./test.py -V` is as expected: `['./test.py', '-V']`
But  `poetry run ./test.py -h` outputs the same as running `poetry run -h`.

PhilipTrauner:

This PR introduces support for the [double-dash](http://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap12.html#tag_12_02) (`--`) argument in `run` argument lists.

`poetry run python3 -- -h` would have previously resulted in `clikit.config.default_application_config.DefaultApplicationConfig.resolve_help_command` being called, which would consequently stop command propagation and print the help text pertaining to the issued command. An event listener with a higher priority (1 instead of 0) than `resolve_help_command` is added to the Poetry application config to prevent this from happening.
Commands that include global options which precede the double dash (e.g. `poetry run python3 -h -- -h`) are left untouched by the listener, and will still have the same outcome as before.

Closes #915

### DIFF
--- a/tests/console/commands/test_run.py
+++ b/tests/console/commands/test_run.py
@@ -14,3 +14,15 @@ def test_run_passes_all_args(app, mocker):
     tester.execute("python -V")
 
     assert [["python", "-V"]] == env.executed
+
+
+def test_double_dash(app, mocker):
+    env = MockEnv(path=Path("/prefix"), base=Path("/base/prefix"), is_venv=True)
+    mocker.patch("poetry.utils.env.EnvManager.get", return_value=env)
+
+    command = app.find("run")
+    tester = CommandTester(command)
+
+    tester.execute("python -h -- -v")
+
+    assert [["python", "-v"]] == env.executed
--- a/docs/docs/cli.md
+++ b/docs/docs/cli.md
@@ -313,6 +313,8 @@ poetry run my-script
 
 Note that this command has no option.
 
+All arguments following a double dash (`--`) are passed on to the given command. 
+
 ## shell
 
 The `shell` command spawns a shell,
--- a/poetry/console/args/run_args_parser.py
+++ b/poetry/console/args/run_args_parser.py
@@ -37,7 +37,7 @@ class RunArgsParser(DefaultArgsParser):
             # Do not include options that preceed the double dash
             double_dash_idx = tokens.index("--")
             for idx, token in enumerate(tokens):
-                if not (idx < double_dash_idx and token.startswith("-")):
+                if not (idx <= double_dash_idx and token.startswith("-")):
                     self._arguments[last_arg.name].append(token)
         else:
             self._arguments[last_arg.name] = tokens[:]
--- a/poetry/console/args/run_args_parser.py
+++ b/poetry/console/args/run_args_parser.py
@@ -30,12 +30,14 @@ class RunArgsParser(DefaultArgsParser):
         tokens = raw_args.tokens[:]
 
         last_arg = list(fmt.get_arguments().values())[-1]
+
         self._arguments[last_arg.name] = []
 
-        while True:
-            try:
-                token = tokens.pop(0)
-            except IndexError:
-                break
-
-            self._arguments[last_arg.name].append(token)
+        if "--" in tokens:
+            # Do not include options that preceed the double dash
+            double_dash_idx = tokens.index("--")
+            for idx, token in enumerate(tokens):
+                if not (idx < double_dash_idx and token.startswith("-")):
+                    self._arguments[last_arg.name].append(token)
+        else:
+            self._arguments[last_arg.name] = tokens[:]
--- a/poetry/console/config/application_config.py
+++ b/poetry/console/config/application_config.py
@@ -26,6 +26,29 @@ class ApplicationConfig(BaseApplicationConfig):
             ConsoleEvents.PRE_HANDLE.value, self.register_command_loggers
         )
         self.add_event_listener(ConsoleEvents.PRE_HANDLE.value, self.set_env)
+        self.add_event_listener(
+            ConsoleEvents.PRE_RESOLVE.value, self.suppress_global_options, priority=1
+        )
+
+    def suppress_global_options(
+        self, event, event_name, _
+    ):  # type: (PreHandleEvent, str, ...) -> None
+        args = event.raw_args
+        application = event.application
+
+        if len(args.tokens) > 0 and args.tokens[0] == "run" and "--" in args.tokens:
+            # Check if a global option token preceeds the first double dash
+            double_dash_idx = args.tokens.index("--")
+            for idx in range(double_dash_idx):
+                if args.tokens[idx].startswith("-"):
+                    return
+
+            command = application.get_command("run")
+
+            parsed_args = command.parse(args)
+
+            event.set_resolved_command(ResolvedCommand(command, parsed_args))
+            event.stop_propagation()
 
     def register_command_loggers(
         self, event, event_name, _


PhilipTrauner:
Sorry for being impatient, but I'm running into this limitation quite often when developing command line applications. I am aware of the fact that `poetry shell` can be used to work around the issue, but I figure that supporting the double-dash would be beneficial regardless, as it is part of the UNIX utility syntax guidelines. (@sdispater)

drunkwcodes:
Maybe placing these changes in there will be more useful?

https://github.com/sdispater/clikit/blob/master/src/clikit/args/default_args_parser.py

PhilipTrauner:
> Maybe placing these changes in there will be more useful?
> 
> [sdispater/clikit:src/clikit/args/default_args_parser.py@`master`](https://github.com/sdispater/clikit/blob/master/src/clikit/args/default_args_parser.py)

It appears as if the double-dash [*is* supported](https://github.com/sdispater/clikit/blob/323c87822f6e89ad5af8548d2e3c18410a5ae648/src/clikit/args/default_args_parser.py#L98) by `clikit` for parsed arguments, but the `help` subcommand tests for `-h` and `--help` in the raw arguments (which includes **all** tokens, not only those after the double-dash) [instead](https://github.com/sdispater/clikit/blob/323c87822f6e89ad5af8548d2e3c18410a5ae648/src/clikit/config/default_application_config.py#L141).

It might make sense to make the change to the help command resolver in `clikit`, but that could affect the behaviour of existing `cleo` applications.

PhilipTrauner:
Again, sorry for being impatient, but is there any update on this / should I try to modify the help command resolver in `clikit` instead? (@sdispater, @drunkwcodes)

tests/console/commands/test_run.py
```diff
@@ -14,3 +14,15 @@ def test_run_passes_all_args(app, mocker):
...
+def test_double_dash(app, mocker):
+    env = MockEnv(path=Path("/prefix"), base=Path("/base/prefix"), is_venv=True)
+    mocker.patch("poetry.utils.env.EnvManager.get", return_value=env)
+
+    command = app.find("run")
+    tester = CommandTester(command)
+
+    tester.execute("python -h -- -v")
+
+    assert [["python", "-v"]] == env.executed
```
drunkwcodes: Since not using MockEnv, may I ask a simple question?

Do you mean `[["python", "-h"]]`?

drunkwcodes:
Since not using `MockEnv`, may I ask a simple question?
Is there a typo in the test?

Do you mean `[['python', '-h']]`?

And I really want to hear the voice of @sdispater about the structure.

PhilipTrauner:
I basically replicated the behaviour of `cargo` for this PR:
```bash
cargo run --bin <binary> -v -- --param-1 1 --param-2 2
```
<p align="center"><b>↓</b></p>

```bash
<binary> --param-1 1 --param-2 2
```

The resulting argument vector does not contain `-v`, as it is a valid `cargo` parameter that precedes the double-dash (`cargo` is consequently invoked in *verbose* mode).

If `poetry` was to follow the same logic (which it should, in my opinion), the resulting output vector should be constructed in the same way:
```bash
poetry run <binary> -h -- -v
```
<p align="center"><b>↓</b></p>

```bash
<binary> -v
```

The assertion in the included test is therefor correct if I'm not mistaken.

drunkwcodes:
https://unix.stackexchange.com/questions/11376/what-does-double-dash-mean-also-known-as-bare-double-dash

PhilipTrauner:
```bash
grep -- -v file
```

`-v` occurs after the double-dash, and is therefor considered an operand, not an option. That is the exact behaviour that the test ensures.

Citing the UNIX utility syntax guidelines:
> The first -- argument that is not an option-argument should be accepted as a delimiter indicating the end of options. Any following arguments should be treated as operands, even if they begin with the '-' character.

Applying this logic to the test case:
```bash
poetry run python -h -- -v
```
<p align="center"><b>↓</b></p>

```bash
python -v
```
The `-h` flag is still consumed by `poetry`, as it should be.

drunkwcodes:
Difference of the output betweewn with `--` and without `--`?

PhilipTrauner:
> Difference of the output betweewn with `--` and without `--`?

```
app = <tests.console.conftest.Application object at 0x108774dd0>
mocker = <pytest_mock.MockFixture object at 0x108774250>

    def test_double_dash(app, mocker):
        env = MockEnv(path=Path("/prefix"), base=Path("/base/prefix"), is_venv=True)
        mocker.patch("poetry.utils.env.EnvManager.get", return_value=env)

        command = app.find("run")
        tester = CommandTester(command)

        tester.execute("python -h -- -v")

>       assert [["python", "-v"]] == env.executed
E       AssertionError: assert [['python', '-v']] == [['python', '-h', '--', '-v']]
E         At index 0 diff: ['python', '-v'] != ['python', '-h', '--', '-v']
E         Full diff:
E         - [['python', '-v']]
E         + [['python', '-h', '--', '-v']]
E         ?             ++++++++++++

tests/console/commands/test_run.py:27: AssertionError
```

The argument vector is not constructed correctly without the changes to the `run` subcommand. Neither `-h` nor the `--` should be part of it.

drunkwcodes:
Please make it pass and then explain the usage of your `--`.

PhilipTrauner:
> Please make it pass and then explain the usage of your `--`.

It *is* passing with this PR applied! I just checked out f2e2ed470ca1560111226b5f6f8f9dc4a28d319a and copy-pasted the newly introduced test into `tests/console/commands/test_run.py` to demonstrate the current behaviour and address your comment.

Here's a quick example why double-dash support is useful:

Suppose you are developing a command line app and using `poetry run` to execute it. Your app will most likely have a `-h` option to provide some kind of usage information. 
Now you want to inspect that usage information yourself, for whatever reason: 
```
poetry run <cli-app> -h
```
<p align="center"><b>↓</b></p>

```
USAGE
  poetry run <args1> ... [<argsN>]

ARGUMENTS
  <args>                 The command and arguments/options to run.

GLOBAL OPTIONS
  -h (--help)            Display this help message
  -q (--quiet)           Do not output any message
  -v (--verbose)         Increase the verbosity of messages: "-v" for normal
                         output, "-vv" for more verbose output and "-vvv" for
                         debug
  -V (--version)         Display this application version
  --ansi                 Force ANSI output
  --no-ansi              Disable ANSI output
  -n (--no-interaction)  Do not ask any interactive question
```

Well, that was unexpected, but *not* incorrect. `poetry` displays its own help text as soon as it spots a `-h` flag in the argument vector, regardless of the flags position.

Now that that's established: How would one go about invoking the `-h` option of the command line app that is being invoked by `poetry run` instead? That's right, with a double dash! (as per UNIX utility syntax guidelines)

```
poetry run <cli-app> -- -h
```
<p align="center"><b>↓</b></p>

```
USAGE
  poetry run <args1> ... [<argsN>]

ARGUMENTS
  <args>                 The command and arguments/options to run.

GLOBAL OPTIONS
  -h (--help)            Display this help message
  -q (--quiet)           Do not output any message
  -v (--verbose)         Increase the verbosity of messages: "-v" for normal
                         output, "-vv" for more verbose output and "-vvv" for
                         debug
  -V (--version)         Display this application version
  --ansi                 Force ANSI output
  --no-ansi              Disable ANSI output
  -n (--no-interaction)  Do not ask any interactive question
```

Still no luck. It turns out that `poetry` still displays its own help text, therefor making it impossible to invoke options of run targets that are also global options of `poetry`.

Let's back up and pretend that `poetry` *did* behave correctly in this instance, and suppress its own help text. The argument vector that is constructed for the run target would *still* be incorrect, because it is a verbatim copy of all options succeeding `run` (in this case `<cli-app> -- -h`, when it should actually be `<cli-app> -h`).

I really hope this clears things up. If you still struggle to understand what this is good for, please clone my branch and try it out.

drunkwcodes:
That is clear. Please rewrite you test with the examples in your comment.

absassi:
Just my two cents: even though there is the double-dash convention, it is also typical that when a command forward flags to another command, all options passed after the name of the command to spawn are blindly passed along (and if there is a double-dash, it would also be passed to the spawned command as well). Two examples that come to my mind that follow this approach are `docker run` and `yarn run`.

In my opinion, this is better from an user experience point-of-view, since the typical usage of `poetry run` is to pass flags to the command being run, not to `poetry run` itself. At least, if I would have to guess how to use the `run` command without reading the help, I would first try, e.g. `poetry run --ansi my-command --some-flag` before trying `poetry run --ansi -- my-command --some-flag`, before finally trying `poetry run --ansi my-command -- --some-flag`.

PhilipTrauner:
Since there has been some activity here recently, let me emphasize that the [`clikit` fix](https://github.com/sdispater/clikit/pull/2) that @kevinastone proposed is definitely the better solution, which sadly doesn't change the fact that it is unlikely to get merged in because it has also been ignored for close to a month now. This PR could serve as a temporary workaround until the underlying `clikit` shortcoming is addressed.

Issue #932, opened 2019-03-04
Add an option for the install command to not create a virtual environment


funkyfuture:
- [x] I have searched the [issues](https://github.com/sdispater/poetry/issues) of this repo and believe that this is not a duplicate.
- [x] I have searched the [documentation](https://poetry.eustace.io/docs/) and believe that my question is not covered.

## Feature Request

one of the reasons that sometimes speaks against using `poetry` for a project is its mostly hard-wired assumption that you want to use it in the context of a virtual environment.

use-cases, that deviate from that assumption,  would be the installation in a virtual machine of a CI solution where the extra overhead isn't needed (and thus a waste of resources) and the installation in a (Docker) container where a virtual environment isn't wanted.

the way to currently achieve that is

```shell
$ poetry config settings.virtualenvs.create false
$ poetry install
$ rm ~/.config/pypoetry
```

something like `poetry install --no-venv` would be more straigthforward imo.

i looked around the code, and my impression it that it isn't simple to achieve as the (not) virtual environment has already been selected at the point when `install`'s options are evaluated. maybe it should be a global option anyway, but i didn't look into that approach yet.

sdispater:
Your use case is already mentionned here: https://poetry.eustace.io/docs/faq/#i-dont-want-poetry-to-manage-my-virtualenvs-can-i-disable-it

This is the recommended and only way to achieve what you want. Adding an option is neither desirable nor necessary, especially once #618 is implemented.

funkyfuture:
well, i should have explicitly mentioned that i desire to use no configuration file at all in such situations (reflected in the 3rd step to deploy an application as referenced above). that makes that option both desirable and necessary.

funkyfuture:
a sidenote regarding the context of using `poetry` to deploy into production environments: given [this overhead](https://github.com/sdispater/poetry/pull/933/files#r265990600) that is needed, a standalone script that can install a project would be superb.

PR #933, opened 2019-03-04
Added basic Docker example

robertpro:
Simple example to use Poetry inside a Docker container

### DIFF
--- a/README.md
+++ b/README.md
@@ -95,6 +95,27 @@ fpath+=~/.zfunc
 ```
 
 
+## Docker example
+
+```dockerfile
+FROM python:3.7-alpine
+
+LABEL maintainer="name <email>"
+
+# copy poetry files
+COPY poetry.lock pyproject.toml ./
+
+# install poetry, dependencies, then remove poetry
+RUN pip --no-cache-dir install poetry poetry-setup \
+    && poetry config settings.virtualenvs.create false \
+    && poetry install \
+    && pip uninstall poetry -y \
+    && rm -rf ~/.config/pypoetry
+
+WORKDIR /app
+```
+
+
 ## Introduction
 
 `poetry` is a tool to handle dependency installation as well as building and packaging of Python packages.


README.md
```diff
@@ -95,6 +95,29 @@ fpath+=~/.zfunc
...
+MAINTAINER name <email>
+
+# copy pyproject file
+COPY pyproject.toml .
+
+# install poetry
+RUN pip --no-cache-dir install poetry poetry-setup
+
+# We don't need a virtualenv since is Docker...
+RUN poetry config settings.virtualenvs.create false
```
funkyfuture: I'd say the artifact that this command creates shoudl be removed after installation. This implies that these three layers that are created by `RUN` directives should be merged. `poetry` itself should also be uninstalled after it did its job, imo. #932 is related.

robertpro: I am sorry, but I don't understand what are you saying... So I uninstall `poetry` and what else?

funkyfuture:
```Dockerfile
RUN pip --no-cache-dir install poetry poetry-setup \
 && poetry config settings.virtualenvs.create false \
 && poetry install \
 && pip uninstall poetry \  # and all of its dependencies ... :-/
 && rm ~/.config/pypoetry
```

that would be it.

funkyfuture: `poetry-setup` isn't used at all, is it?

robertpro: you are right, the only use case is that you are testing a new dependency so you install with poetry

README.md
```diff
@@ -95,6 +95,29 @@ fpath+=~/.zfunc
 ```
 
 
+## Docker example
+
+```dockerfile
+FROM python:3.7-alpine
+
+MAINTAINER name <email>
```
pawamoy: [MAINTAINER is deprecated](https://docs.docker.com/engine/reference/builder/#maintainer-deprecated).

You should use `LABEL maintainer="name <email>"` instead, and also put it at the end to avoid cache invalidation when you only change the label :slightly_smiling_face:

robertpro: nice to know, will update!

robertpro:
Updated!

hangtwenty:
A Docker example could be good but I don't think this is sufficient. A production Dockerfile would include `poetry.lock`. See: 

* ["Integrating Python Poetry with Docker"](https://stackoverflow.com/questions/53835198/integrating-python-poetry-with-docker/54830677#comment97058172_54763270) 
* https://github.com/wemake-services/wemake-django-template/issues/697

funkyfuture:
i think the various answers to the SO question show that different approaches are more or less valid depending on the context. i dislike the currently top answer for it use of environment variables that are eventually not used. there's a link in its comments that discusses the use of a multi-stage build. while my answer is quick when building after sole code changes w/o changing dependencies.

robertpro:
I don't think this should cover all possible environments, its just an example, the first time I tried making this work, it took me like a day (first time with Docker and Poetry...) and as you can it has some missing work, what I want to say is that if you need an example, this is it.

robertlagrant:
> I don't think this should cover all possible environments, its just an example, the first time I tried making this work, it took me like a day

I know the feeling. The alternate way to install Poetry - from the get_poetry script - I couldn't get to work at all in Docker, as `source`ing the env file never seemed to work.

README.md
robertlagrant: @funkyfuture doesn't this still have the problem of potentially conflicting versions between Poetry's dependencies and app dependencies?

funkyfuture: indeed. but would these confilcts matter? at the point where poetry is executed it has all its dependencies loaded. if the project installation breaks these, be it so, poetry won't be executed another time. nonetheless i think a standalone script that just does what the `install` command achieves would be the best solution to deploy projects.

hangtwenty:
> I don't think this should cover all possible environments, its just an example, the first time I tried making this work, it took me like a day (first time with Docker and Poetry...) and as you can it has some missing work, what I want to say is that if you need an example, this is it.

It’s great to share that but unless it is general use it may be  suited for a blogpost or gist.github.com entry until it’s more general-use. It’s good that you added the lock file, but there are still other details that (I think) should be covered if included in Poetry official docs, like the no-interaction and no-ansi flags (see the links I added in previous comment)

(I’m not a maintained though and obviously defer to the maintainers)

PR #953, opened 2019-03-11
Requirements export fix

SfinxCZ:
This pull request modifies the export of dependencies to `requirements.txt` format. 
The currently released implementation has problem with custom repositories. It exports custom repositories next to the given dependency which is not correct (see https://pip.readthedocs.io/en/1.1/requirements.html#indexes-find-links or https://github.com/pypa/pip/blob/master/src/pip/_internal/req/req_file.py#L131).

This PR changes current behavior in multiple ways. 
First, it puts all links to external dependencies to beginning of the output `requirements.txt` file. 
Next, it changes `--index-url` to `--extra-index-url` as `--extra-index-url` adds to the list of queried repositories instead of replacing the list. 
And finally, it adds credentials for respective repositories to the repository URL.

### DIFF
--- a/poetry/console/commands/export.py
+++ b/poetry/console/commands/export.py
@@ -42,7 +42,7 @@ class ExportCommand(Command):
                 "</warning>"
             )
 
-        exporter = Exporter(self.poetry.locker)
+        exporter = Exporter(self.poetry.locker, self.poetry.pool)
         exporter.export(
             fmt,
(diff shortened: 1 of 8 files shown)


sdispater:
This is handled (among other things) in #1277, so I am closing this PR.

Nonetheless, thank you for submitting this!

Issue #807, opened 2019-01-18
Poetry downloads `Numpy` as Mac OSX Wheel on Windows Laptop & Throws [WinError 145]


KyleKing:
- [x] I am on the [latest](https://github.com/sdispater/poetry/releases/latest) Poetry version.
- [x] I have searched the [issues](https://github.com/sdispater/poetry/issues) of this repo and believe that this is not a duplicate. (Searched for WinError 145, numpy, etc.)
- [x] If an exception occurs when executing a command, I executed it again in debug mode (`-vvv` option).

- **OS version and name**: Windows 7
- **Poetry version**: `0.12.11`
- **Linked pyproject.toml file (renamed as `.toml.txt` to upload)**: [pyproject.toml.txt](https://github.com/sdispater/poetry/files/2771199/pyproject.toml.txt)

## Issue

Can be reproduced on my laptop with the below script. When attempting to add Numpy, it returns a directory not empty error and can't get past that.

```sh
poetry new test_numpy
cd test_numpy
poetry install
poetry add numpy -vvv
```

In the debug message below, Poetry downloads `numpy-1.16.0-cp27-cp27m-macosx_10_6_intel.macosx_10_9_intel.macosx_10_9_x86_64.macosx_10_10_intel.macosx_10_10_x86_64.wh` but I'm on a Windows system

```
Microsoft Windows [Version 6.1.7601]
Copyright (c) 2009 Microsoft Corporation.  All rights reserved.

(py37) C:\Users\king.kyle\Developer\_test_numpy>poetry install
Updating dependencies
Resolving dependencies...

Nothing to install or update

Writing lock file

  - Installing test_numpy (0.1.0)

(py37) C:\Users\king.kyle\Developer\_test_numpy>poetry add numpy -vvv
Using virtualenv: C:\Users\king.kyle\AppData\Local\pypoetry\Cache\virtualenvs\mm-upgrade-utility-py3.7
PyPI: No release information found for numpy-0.9.6, skipping
PyPI: No release information found for numpy-0.9.8, skipping
PyPI: No release information found for numpy-1.0.3, skipping
PyPI: No release information found for numpy-1.0.4, skipping
PyPI: No release information found for numpy-1.0b1, skipping
PyPI: No release information found for numpy-1.0b4, skipping
PyPI: No release information found for numpy-1.0b5, skipping
PyPI: No release information found for numpy-1.0rc1, skipping
PyPI: No release information found for numpy-1.0rc2, skipping
[...]

PR #955, opened 2019-03-12
Use Python Version and OS to Select the Appropriate Wheel (#944)  {Against "master" Branch}

### LINKED ISSUE #944: Poetry Downloads 1st Wheel File from PyPi Regardless of Python version or OS. Unusable on Windows for projects with Pandas/Numpy

## SysInfo

- **OS version and name**: Windows 7 [Version 6.1.7601]
- **Poetry version**: 1.0.0a2
- **pyproject.toml file (labeled as .txt to upload - nothing special, just default)**: [pyproject.toml.txt](https://github.com/sdispater/poetry/files/2946237/pyproject.toml.txt)

## Issue

I think there is a problem in the wheels resolver in Poetry. I found a couple of test cases that build wheels targeting multiple OS and Python versions. In all cases, it looks like Poetry is downloading the first Wheel regardless of Python/OS constraints, then crashes with a `[WinError 145] The directory is not empty: 'C:\\Users\\KING~1.KYL\\AppData\\Local\\Temp\\tmpopks2va_'`.

Originally I posted issue #807, but I think this might be a larger issue with Poetry and not just Numpy, so I think a new ticket is more appropriate to track this issue.

Below is the abbreviated output for attempting an install of each of numpy and pywin32. The full log for all 6 packages is attached: [FullLogs.txt](https://github.com/sdispater/poetry/files/2946242/FullLogs.txt)

I think the relevant line is `PyPI: Downloading wheel: pywin32-224-cp27-cp27m-win32.whl` (cp27) where the correct version should have been: `pywin32-224-cp37-cp37m-win32.whl` (cp37)

- [numpy](https://pypi.org/project/numpy/#files)
- [pywin32](https://pypi.org/project/pywin32/#files)
- [wxPython](https://pypi.org/project/wxPython/#files)
- [Pillow](https://pypi.org/project/Pillow/#files)
- [Pygame](https://pypi.org/project/Pygame/#files)
- [tensorflow](https://pypi.org/project/tensorflow/#files)

```sh
>> poetry --version
Poetry version 1.0.0a2

>> poetry new TestWheelResolver
>> cd TestWheelResolver
>> poetry add pywin32 -vvv
Using virtualenv: C:\Users\king.kyle\AppData\Local\pypoetry\Cache\virtualenvs\mmuu-_zx5-I-Q-py3.7
PyPI: No release information found for pywin32-210, skipping
PyPI: No release information found for pywin32-214, skipping
PyPI: 3 packages found for pywin32 *
Using version ^224.0 for pywin32

Updating dependencies
Resolving dependencies...
   1: fact: testwheelresolver is 0.1.0
   1: derived: testwheelresolver
   1: fact: testwheelresolver depends on pywin32 (^224.0)
   1: fact: testwheelresolver depends on pytest (^3.0)
   1: selecting testwheelresolver (0.1.0)
   1: derived: pytest (^3.0)
   1: derived: pywin32 (^224.0)
PyPI: 44 packages found for pytest >=3.0,<4.0
PyPI: No release information found for pywin32-210, skipping
PyPI: No release information found for pywin32-214, skipping
PyPI: 1 packages found for pywin32 >=224.0,<225.0
PyPI: Getting info for pywin32 (224) from PyPI
PyPI: No dependencies found, downloading archives
PyPI: Downloading wheel: pywin32-224-cp27-cp27m-win32.whl
   1: Version solving took 16.753 seconds.
   1: Tried 1 solutions.

>> poetry add numpy -vvv
Using virtualenv: C:\Users\king.kyle\AppData\Local\pypoetry\Cache\virtualenvs\mmuu-_zx5-I-Q-py3.7
PyPI: No release information found for numpy-0.9.6, skipping
PyPI: No release information found for numpy-0.9.8, skipping
PyPI: No release information found for numpy-1.0.3, skipping
PyPI: No release information found for numpy-1.0.4, skipping
PyPI: No release information found for numpy-1.0b1, skipping
PyPI: No release information found for numpy-1.0b4, skipping
PyPI: No release information found for numpy-1.0b5, skipping
PyPI: No release information found for numpy-1.0rc1, skipping
PyPI: No release information found for numpy-1.0rc2, skipping
PyPI: No release information found for numpy-1.0rc3, skipping
PyPI: No release information found for numpy-1.1.1, skipping
PyPI: No release information found for numpy-1.2.0, skipping
PyPI: No release information found for numpy-1.2.1, skipping
PyPI: No release information found for numpy-1.4.0, skipping
PyPI: 47 packages found for numpy *
Using version ^1.16 for numpy

Updating dependencies
Resolving dependencies...
   1: fact: testwheelresolver is 0.1.0
   1: derived: testwheelresolver
   1: fact: testwheelresolver depends on numpy (^1.16)
   1: fact: testwheelresolver depends on pytest (^3.0)
   1: selecting testwheelresolver (0.1.0)
   1: derived: pytest (^3.0)
   1: derived: numpy (^1.16)
PyPI: 44 packages found for pytest >=3.0,<4.0
PyPI: No release information found for numpy-0.9.6, skipping
PyPI: No release information found for numpy-0.9.8, skipping
PyPI: No release information found for numpy-1.0.3, skipping
PyPI: No release information found for numpy-1.0.4, skipping
PyPI: No release information found for numpy-1.0b1, skipping
PyPI: No release information found for numpy-1.0b4, skipping
PyPI: No release information found for numpy-1.0b5, skipping
PyPI: No release information found for numpy-1.0rc1, skipping
PyPI: No release information found for numpy-1.0rc2, skipping
PyPI: No release information found for numpy-1.0rc3, skipping
PyPI: No release information found for numpy-1.1.1, skipping
PyPI: No release information found for numpy-1.2.0, skipping
PyPI: No release information found for numpy-1.2.1, skipping
PyPI: No release information found for numpy-1.4.0, skipping
PyPI: 3 packages found for numpy >=1.16,<2.0
PyPI: Getting info for numpy (1.16.2) from PyPI
PyPI: No dependencies found, downloading archives
PyPI: Downloading sdist: numpy-1.16.2.zip
PyPI: Downloading wheel: numpy-1.16.2-cp27-cp27m-macosx_10_6_intel.macosx_10_9_intel.macosx_10_9_x86_64.macosx_10_10_intel.macosx_10_10_x86_64.whl
   1: Version solving took 26.382 seconds.
   1: Tried 1 solutions.
```

I will have some time this weekend to try to help figure out the root cause and if so, submit a PR. I'll also compare to Mac and see if this is Windows-only. Need to consider the wide list of possible platforms: https://github.com/easybuilders/easybuild/wiki/OS_flavor_name_version

This might be what needs to be updated: https://github.com/sdispater/poetry/blob/271e995844db11a40c7c4a0b320bf9bd600db444/poetry/repositories/pypi_repository.py#L423-L425

**Update**: so the OSError is *Windows-only*. On Mac, the wrong cp27 wheel is downloaded for cp37 (`[tool.poetry.dependencies]\n python = "^3.7"`) for both Pillow and Pygame, but install completes without issue.

I started a potential fix in this branch: https://github.com/KyleKing/poetry/tree/fix/944-select_wheel

KyleKing:
Resolves #944 & #807 

**Issue**: If a package only has `platform_specific_wheels`, Poetry just downloads the first one, which is usually cp27 and MacOSX (i.e.: [numpy](https://pypi.org/project/numpy/#files))

**Changes**: This PR adds logic to determine the best wheel file to download for [Poetry supported platforms](https://poetry.eustace.io/docs/#system-requirements). If the best wheel can't be identified, the method falls back on the original behavior to return the first platform wheel. This fixes the *[OS Error]* on Windows and improves matching for Linux and Mac.

Also, tested on a couple of different computers as a sanity check (Win7-64, Win10-64, and MacOSX Mojave)

# Pull Request Check List

- [X] Updated **documentation** for changed code. {N/A - no changes necessary}

### DIFF
--- a/poetry/repositories/pypi_repository.py
+++ b/poetry/repositories/pypi_repository.py
@@ -445,9 +445,7 @@ class PyPiRepository(Repository):
         # Format the expected platform name as used by package authors
         os_map = {"windows": "win", "darwin": "macosx"}
         os_name = (
-            os_map[sys_info["plat"]]
-            if sys_info["plat"] in os_map
-            else sys_info["plat"]
+            os_map[sys_info["plat"]] if sys_info["plat"] in os_map else sys_info["plat"]
         )
         machine = sys_info["machine"]
         if os_name == "win" and machine == "x86":
--- a/poetry/repositories/pypi_repository.py
+++ b/poetry/repositories/pypi_repository.py
@@ -451,7 +451,8 @@ class PyPiRepository(Repository):
         bit_label = "32" if sys_info["is32bit"] else "64"
         py_label = "cp{}".format("".join(sys_info["pyver"][:2]))
         self._log(
-            "Attempting to determine best match for: {}".format(sys_info), level="debug"
+            "Attempting to determine best wheel file for: {}".format(sys_info),
+            level="debug",
         )
 
         platform_matches = []
@@ -461,7 +462,12 @@ class PyPiRepository(Repository):
             if os_name in plat:
                 match_py = m.group("pyver") == py_label
                 if match_py and (bit_label in plat or "x86_64" in plat):
-                    self._log("Found best wheel match: {}".format(url), level="debug")
+                    self._log(
+                        "Selected best platform, bit, and Python version match: {}".format(
+                            url
+                        ),
+                        level="debug",
+                    )
                     return url
                 elif match_py:
                     platform_matches.insert(0, url)
--- a/poetry/repositories/pypi_repository.py
+++ b/poetry/repositories/pypi_repository.py
@@ -1,5 +1,6 @@
 import logging
 import os
+import platform
 import tarfile
 import zipfile
 
@@ -421,10 +422,64 @@ class PyPiRepository(Repository):
                 return info
 
             if platform_specific_wheels and "sdist" not in urls:
-                # Pick the first wheel available and hope for the best
-                return self._get_info_from_wheel(platform_specific_wheels[0])
+                # Attempt to select the best platform-specific wheel
+                best_wheel = self._pick_platform_specific_wheel(
+                    platform_specific_wheels
+                )
+                return self._get_info_from_wheel(best_wheel)
 
         return self._get_info_from_sdist(urls["sdist"][0])
+
+    def get_sys_info(self):  # type: () -> Dict[str, str]
+        # Return system information. Can be overridden for testing
+        return {
+            "plat": platform.system().lower(),
+            "machine": platform.machine().lower(),
+            "pyver": platform.python_version_tuple(),
+        }
+
+    def _pick_platform_specific_wheel(
+        self, platform_specific_wheels
+    ):  # type: (list) -> str
+        sys_info = self.get_sys_info()
+        # Format the expected platform name as used by package authors
+        os_map = {"windows": "win", "darwin": "macosx"}
+        os_name = (
+            os_map[sys_info["plat"]]
+            if sys_info["plat"] in os_map
+            else sys_info["plat"]
+        )
+        machine = sys_info["machine"]
+        if os_name == "win" and machine == "x86":
+            machine = "32"  # Fix search string for Windows 32bit systems
+        py_label = "cp{}".format("".join(sys_info["pyver"][:2]))
+        self._log(
+            "Attempting to determine best match for: {}".format(sys_info), level="debug"
+        )
+
+        platform_matches = []
+        for url in platform_specific_wheels:
+            m = wheel_file_re.match(Link(url).filename)
+            plat = m.group("plat")
+            if os_name in plat:
+                match_py = m.group("pyver") == py_label
+                if match_py and (machine in plat or "x86_64" in plat):
+                    self._log("Found best wheel match: {}".format(url), level="debug")
+                    return url
+                elif match_py:
+                    platform_matches.insert(0, url)
+        if len(platform_matches) > 0:
+            # Return first platform match as more specificity couldn't be determined
+            self._log(
+                "Selecting wheel file: {}".format(platform_matches[0]), level="debug"
+            )
+            return platform_matches[0]
+        # Could not pick the best wheel, return the first available and hope for the best
+        self._log(
+            "Matching failed, selecting wheel file: {}".format(platform_matches[0]),
+            level="debug",
+        )
+        return platform_specific_wheels[0]
 
     def _get_info_from_wheel(
         self, url
--- a/poetry/repositories/pypi_repository.py
+++ b/poetry/repositories/pypi_repository.py
@@ -90,6 +90,21 @@ class PyPiRepository(Repository):
 
         super(PyPiRepository, self).__init__()
 
+    @property
+    def sys_info(self):  # type: () -> Dict[str, str]
+        """
+        Return dictionary describing the OS and Python configuration.
+        """
+        if self._sys_info:
+            return self._sys_info
+        self._sys_info = {
+            "plat": platform.system().lower(),
+            "is32bit": sys.maxsize <= 2 ** 32,
+            "imp_name": sys.implementation.name,
+            "pyver": platform.python_version_tuple(),
+        }
+        return self._sys_info
+
     def find_packages(
         self,
         name,  # type: str
@@ -439,28 +454,18 @@ class PyPiRepository(Repository):
 
         return self._get_info_from_sdist(urls["sdist"][0])
 
-    def get_sys_info(self):  # type: () -> Dict[str, str]
-        # Return system information. Can be overridden for testing
-        return {
-            "plat": platform.system().lower(),
-            "is32bit": sys.maxsize <= 2 ** 32,
-            "imp_name": sys.implementation.name,
-            "pyver": platform.python_version_tuple(),
-        }
-
     def _pick_platform_specific_wheel(
         self, platform_specific_wheels
     ):  # type: (list) -> str
-        sys_info = self.get_sys_info()
         # Format information for checking the PEP425 "Platform Tag"
         os_map = {"windows": "win", "darwin": "macosx"}
-        os_name = os_map.get(sys_info["plat"], sys_info["plat"])
-        bit_label = "32" if sys_info["is32bit"] else "64"
+        os_name = os_map.get(self.sys_info["plat"], self.sys_info["plat"])
+        bit_label = "32" if self.sys_info["is32bit"] else "64"
         # Format information for checking the PEP425 "Python Tag"
-        imp_abbr = self.IMP_NAME_LOOKUP.get(sys_info["imp_name"].lower(), "py")
-        py_abbr = "".join(sys_info["pyver"][:2])
+        imp_abbr = self.IMP_NAME_LOOKUP.get(self.sys_info["imp_name"].lower(), "py")
+        py_abbr = "".join(self.sys_info["pyver"][:2])
         self._log(
-            "Attempting to determine best wheel file for: {}".format(sys_info),
+            "Attempting to determine best wheel file for: {}".format(self.sys_info),
             level="debug",
         )
 
--- a/poetry/repositories/pypi_repository.py
+++ b/poetry/repositories/pypi_repository.py
@@ -95,14 +95,13 @@ class PyPiRepository(Repository):
         """
         Return dictionary describing the OS and Python configuration.
         """
-        if self._sys_info:
-            return self._sys_info
-        self._sys_info = {
-            "plat": platform.system().lower(),
-            "is32bit": sys.maxsize <= 2 ** 32,
-            "imp_name": sys.implementation.name,
-            "pyver": platform.python_version_tuple(),
-        }
+        if not hasattr(self, "_sys_info"):
+            self._sys_info = {
+                "plat": platform.system().lower(),
+                "is32bit": sys.maxsize <= 2 ** 32,
+                "imp_name": sys.implementation.name,
+                "pyver": platform.python_version_tuple(),
+            }
         return self._sys_info
 
     def find_packages(
--- a/poetry/repositories/pypi_repository.py
+++ b/poetry/repositories/pypi_repository.py
@@ -1,4 +1,5 @@
 import logging
+import sys
 import os
 import platform
 import tarfile
@@ -434,7 +435,7 @@ class PyPiRepository(Repository):
         # Return system information. Can be overridden for testing
         return {
             "plat": platform.system().lower(),
-            "machine": platform.machine().lower(),
+            "is32bit": sys.maxsize <= 2 ** 32,
             "pyver": platform.python_version_tuple(),
         }
 
@@ -447,9 +448,7 @@ class PyPiRepository(Repository):
         os_name = (
             os_map[sys_info["plat"]] if sys_info["plat"] in os_map else sys_info["plat"]
         )
-        machine = sys_info["machine"]
-        if os_name == "win" and machine == "x86":
-            machine = "32"  # Fix search string for Windows 32bit systems
+        bit_label = "32" if sys_info["is32bit"] else "64"
         py_label = "cp{}".format("".join(sys_info["pyver"][:2]))
         self._log(
             "Attempting to determine best match for: {}".format(sys_info), level="debug"
@@ -461,7 +460,7 @@ class PyPiRepository(Repository):
             plat = m.group("plat")
             if os_name in plat:
                 match_py = m.group("pyver") == py_label
-                if match_py and (machine in plat or "x86_64" in plat):
+                if match_py and (bit_label in plat or "x86_64" in plat):
                     self._log("Found best wheel match: {}".format(url), level="debug")
                     return url
                 elif match_py:
@@ -469,12 +468,17 @@ class PyPiRepository(Repository):
         if len(platform_matches) > 0:
             # Return first platform match as more specificity couldn't be determined
             self._log(
-                "Selecting wheel file: {}".format(platform_matches[0]), level="debug"
+                "Selecting first wheel file for platform {}: {}".format(
+                    os_name, platform_matches[0]
+                ),
+                level="debug",
             )
             return platform_matches[0]
         # Could not pick the best wheel, return the first available and hope for the best
         self._log(
-            "Matching failed, selecting wheel file: {}".format(platform_matches[0]),
+            "Matching was unsuccessful, selecting first wheel file: {}".format(
+                platform_specific_wheels[0]
+            ),
             level="debug",
         )
         return platform_specific_wheels[0]
--- a/poetry/repositories/pypi_repository.py
+++ b/poetry/repositories/pypi_repository.py
@@ -58,6 +58,14 @@ class PyPiRepository(Repository):
 
     CACHE_VERSION = parse_constraint("0.12.0")
 
+    # Implementation name lookup specified by PEP425
+    IMP_NAME_LOOKUP = {
+        "cpython": "cp",
+        "ironpython": "ip",
+        "pypy": "pp",
+        "jython": "jy",
+    }
+
     def __init__(self, url="https://pypi.org/", disable_cache=False, fallback=True):
         self._name = "PyPI"
         self._url = url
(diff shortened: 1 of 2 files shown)


Pluckerpluck:
I'll try to test this out tomorrow (UK). This is a much needed fix as, with `numpy` being broken, Poetry is unusable in my field.

This should also prioritize platform specific wheels over sdist if I'm looking at this correctly? I remember having a problem where pip could install a package just fine but poetry (at least when run via tox) had to build it, which failed as I didn't have a C compiler. So this could potentially fix that as well.

Pluckerpluck:
Appears to work perfectly for me. I can now install Numpy and haven't yet noticed any side effects.

Pluckerpluck:
New problem:

Python 32bit can be run on a 64bit PC. In this situation this downloads the 64bit version, which again fails to run.

At least, that's what I assume the problem is. I ran this with 32bit python on a separate PC and it failed (and downloaded the amd64 version of PyWin32). Tomorrow I can run it against a 64 bit version of python to confirm this is the problem, but I don't have time today.

-------

This doesn't change the fact that this pull request is still valid and is much more effective than the "randomly pick the first one" option that previously existed, but it is still a problem.

KyleKing:
@Puckerpluck, I think that case is fixable and I’ll look tonight. I remember planning to use the bit-version of Python instead of the OS, but I’ll have to check

KyleKing:
@Pluckerpluck do you have an example 64-bit package that failed to install on 32-bit Python? I can make it one of the test cases

KyleKing:
@sdispater this branch should be ready for review. Let me know if you have any questions or would like to see any changes

---

@Pluckerpluck, the latest changes should address the case where Python is 32-bit and OS is 64. I tested on a Win7-64 with Python 32-bit:

```bat
(py372) C:\Users\king.kyle\Desktop\TEST>poetry add pywin32 -vvv
Using virtualenv: C:\Users\king.kyle\AppData\Local\pypoetry\Cache\virtualenvs\TEST-py3.7
PyPI: No release information found for pywin32-210, skipping
PyPI: No release information found for pywin32-214, skipping
PyPI: 3 packages found for pywin32 *
Using version ^224.0 for pywin32

Updating dependencies
Resolving dependencies...
   1: fact: test is 0.1.0
   1: derived: test
   1: fact: test depends on pywin32 (^224.0)
   1: selecting test (0.1.0)
   1: derived: pywin32 (^224.0)
PyPI: No release information found for pywin32-210, skipping
PyPI: No release information found for pywin32-214, skipping
PyPI: 1 packages found for pywin32 >=224.0,<225.0
PyPI: Getting info for pywin32 (224) from PyPI
PyPI: No dependencies found, downloading archives
PyPI: Attempting to determine best match for: {'plat': 'windows', 'not32bit': False, 'pyver': ('3', '7', '2')}
PyPI: Found best wheel match: https://files.pythonhosted.org/packages/8a/37/917c4020e93e0e854d4cbff1cbdf14ec45a6d1cedf52f8cafdea5b22451a/pywin32-224-cp37-cp37m-win32.whl
PyPI: Downloading wheel: pywin32-224-cp37-cp37m-win32.whl
   1: selecting pywin32 (224)
   1: Version solving took 4.029 seconds.
   1: Tried 1 solutions.


Writing lock file

  - Installing pywin32 (224)

(py372) C:\Users\king.kyle\Desktop\TEST>python
Python 3.7.2 (default, Jan  2 2019, 17:27:13) [MSC v.1912 32 bit (Intel)] :: Anaconda, Inc. on win32
>>> import sys
>>> import platform
>>> platform.machine()
'AMD64'
>>> sys.maxsize > 2**32
False
>>>
```

Pluckerpluck:
Turns out my error was unrelated. While I can now confirm that 32 bit python downloads 32 bit versions, is it actually a requirement to work? I honestly have no idea and had just assumed it from the identical error.

------------

I might actually be suffering from a bug within `shutil.rmtree` on windows which can fail sometimes. There are multiple instances of this occurring, with issues in both [pip](https://github.com/pypa/pip/issues/4910) and [ansible](https://github.com/ansible/ansible/issues/34335).

I'll report this as its own issue once I've investigated further (if it's definitely a bug).

KyleKing:
@sdispater is there anything I can do to help merge this feature?

Would it be better to submit this change to the development branch? Are there any additional tests or other changes you would like to see before merging?

This PR fixes a showstopper issue for us and has been in use for the better part of two months!

drunkwcodes:
The comparison with `PackageFinder` in `setuptools`?

KyleKing:
@drunkwcodes, I think you're asking what makes this PR different

I checked the source for `setuptools.PackageFinder`: [https://github.com/pypa/setuptools/setuptools/__init__.py#L47](https://github.com/pypa/setuptools/blob/b31997dbcd935ded39868b42d765830ad6e29808/setuptools/__init__.py#L47)

The `PackageFinder` class appears to identify all Python packages within a local filesystem directory

This PR parses the list of wheel filenames for the optimal OS and Python-specific one

drunkwcodes:
@KyleKing I see.

PEP 425 compatibility?
And I think having a room for fuzzy search as a fallback method is good.

poetry/repositories/pypi_repository.py
```diff
@@ -421,11 +423,72 @@ def _get_info_from_urls(
...
+        bit_label = "32" if sys_info["is32bit"] else "64"
+        py_label = "cp{}".format("".join(sys_info["pyver"][:2]))
+        self._log(
+            "Attempting to determine best wheel file for: {}".format(sys_info),
+            level="debug",
+        )
+
+        platform_matches = []
+        for url in platform_specific_wheels:
+            m = wheel_file_re.match(Link(url).filename)
```
KyleKing: Uses the `wheel_file_re` regex to parse the wheel filename for the PEP425 compatibility tags

poetry/repositories/pypi_repository.py
```diff
@@ -421,11 +423,72 @@ def _get_info_from_urls(
...
+            plat = m.group("plat")
+            if os_name in plat:
+                match_py = m.group("pyver") == py_label
```
KyleKing: Checks if the major python version from self.sys_info (i.e. `cp37`) matches the PEP425 `python tag`

poetry/repositories/pypi_repository.py
```diff
@@ -421,11 +423,72 @@ def _get_info_from_urls(
...
+                if match_py and (bit_label in plat or "x86_64" in plat):
```
KyleKing: Checks the `platform tag`

KyleKing:
The logic steps use the PEP425-specified tags to find the best match. I commented a few of the relevant lines

(Sorry for the extra emails, I meant to put those comments in one review)

kung-foo:
I had a similar issue with a `llvmlite` dependency. It would fetch a 2.7 macosx wheel and then add a requires for `enum34` which breaks when running on python > 3.4. This PR fixed it!

poetry/repositories/pypi_repository.py
```diff
@@ -421,11 +431,74 @@ def _get_info_from_urls(
...
-                return self._get_info_from_wheel(platform_specific_wheels[0])
+                # Attempt to select the best platform-specific wheel
+                best_wheel = self._pick_platform_specific_wheel(
+                    platform_specific_wheels
+                )
+                return self._get_info_from_wheel(best_wheel)
 
         return self._get_info_from_sdist(urls["sdist"][0])
 
+    def get_sys_info(self):  # type: () -> Dict[str, str]
```
kbakk: could sys_info be a property on class `PyPiRepository`, or `Repository`? Given that it may be used elsewhere, this could prevent a few calls (not that they are very expensive).

KyleKing: @KeltonKarboviak, thanks for the input. I reviewed the code base and I think that setting `sys_info` as a class property will actually be less performant. As far as I can tell, `PyPiRepository` is always initialized when used (the class and properties aren't reused), which means that if `sys_info` is a class property, then those functions will be called for each class instance/repository. However, the local hardware information is only needed for a subset of python repositories that otherwise can't be identified. When the hardware information is needed, it is only called once before looping through the potential wheel file names

Let me know if there is a something I overlooked. I already staged code changes, so this would be easy to implement

KyleKing:
Pushed new PR to the develop (poetry 1.x.x) branch #1144. I wanted the 1.X.X features locally and maybe that PR will have a better chance of being merged

poetry/repositories/pypi_repository.py
kbakk: @KyleKing I was thinking in the lines of this, which would only be called when cls().sys_info is called.

```python
@property
def sys_info(self):  # type: () -> Dict[str, str]
    # Return system information. Can be overridden for testing
    if self._sys_info:
        return self._sys_info
    self._sys_info = {
        "plat": platform.system().lower(),
        "is32bit": sys.maxsize <= 2 ** 32,
        "imp_name": sys.implementation.name,
        "pyver": platform.python_version_tuple(),
    }
    return self._sys_info
```

You would need to update 

https://github.com/sdispater/poetry/blob/41b321f990bd66e3053f76585f75188e629ec890/tests/repositories/test_pypi_repository.py#L53

https://github.com/sdispater/poetry/blob/41b321f990bd66e3053f76585f75188e629ec890/poetry/repositories/pypi_repository.py#L454

KyleKing: Oh you're right, thanks!

sdispater:
I am on the fence regarding this.

While that prevents installation errors it defeats the "platform-agnostic lock file" purpose. I wonder if it would be preferable to build markers based on the wheel files if no sdist is present, that way we are making sure that the package won't be installed on systems that are not compatible.

@KyleKing What do you think?

KyleKing:
I've researched/rewritten this response a number of times, so hopefully this final iteration makes sense! Let me know what you think

This PR shouldn't impact the lock file. The changes fix the resolver for selecting the wheel file in the case that an sdist or generic distribution isn't available.

I ran a test initializing a project on a Mac, then cloning and installing the project on Windows (used the code from this PR in both cases). With the downloaded Mac `poetry.lock` file, Poetry picked the correct Windows wheel file for [`numpy`](https://pypi.org/project/numpy/#files) and the lock file was unchanged. I also tried deleting the lock file and confirming that Poetry created the exact same file. This makes sense as the poetry.lock file stores every wheel file hash and doesn't explicitly indicate the wheel file used.

See my demo at: [KyleKing/Poetry-EdgeCaseTesting/PR955-20190623](https://github.com/KyleKing/Poetry-EdgeCaseTesting/tree/PR955-20190623) / [commits](https://github.com/KyleKing/Poetry-EdgeCaseTesting/commits/PR955-20190623)

---

Separately, I think it would be great to have environment markers in Poetry. I would be happy to submit or help review a PR

The markers would fix the case where a developer adds a Windows-only dependency (i.e. `pywin32`) to the `pyproject.toml` file. Then, when attempting to run `poetry install` on a Mac/Linux, Poetry crashes installing `pywin32` (which would have been safely skipped if a marker was available)

Environment markers looks to be [specified in PEP-508](https://www.python.org/dev/peps/pep-0508/#environment-markers) and this is an [example from pipenv: `"markers": "sys_platform == 'linux'"`](https://github.com/pypa/pipenv/blob/master/Pipfile.lock#L598))

---

(More for my reference) Here are some external links:

- [Discussion on pipenv for OS-specific locking](https://github.com/pypa/pipenv/issues/598#issue-258749520)
    - [Follow up - looks like it was removed](https://github.com/pypa/pipenv/issues/753#issuecomment-350175362)
- [PipEnv Discussion on hash verification breaking when new wheel files uploaded](https://github.com/pypa/pipenv/issues/2618#issuecomment-408243400)
- [PipEnv issue with markers for ref](https://github.com/pypa/pipenv/issues/2378)
- [StackOverflow question on yarn.lock](https://stackoverflow.com/questions/44862135/maintaining-yarn-lock-cross-platform)
- [NPM package-lock documentation](https://docs.npmjs.com/files/package-lock.json)
    - "All optional dependencies should be included even if they’re uninstallable on the current platform.""
- [NPM ignores the lock file and just uses the package.json](https://github.com/agconti/how-package-lock-works) - not sure if this is still the case

KyleKing:
Oops, I misunderstood your question and didn't realize Poetry supports environment markers (I've edited my last comment)

I think the desired behavior is that poetry will crash when attempting to install a dependency not available for the specific system. This error is clear and indicates that the project hasn't been tested/designed for the platform the user is attempting to install on and will require code changes. If Poetry were to *auto-assign* markers (python version/environment), the error would be buried at install and instead would later appear as a `no module named ___` ImportError.

Sometimes authors will add [additional wheel files after a release](https://github.com/pypa/pipenv/issues/2618#issue-343087945), although rare, could cause Poetry to add markers that are no longer relevant a few days later. Poetry would also likely need to be able to disable auto-marker assignment by dependency and the setting would need to be preserved when the project is cloned.

I think the existing support for user-specified markers is the right solution. I'm always open to editing this PR, so let me know what you think

KyleKing:
@sdispater, do you think you're still on the fence?

- This change implements PEP425 support for packages that do not produce sdist files nor generic wheels
- This PR does not change the lockfile
- I think it should be up to the user's to specify proper environment markers. Having Poetry auto-assign markers would only be burying errors (PEP20 - "Errors should never pass silently")

sdispater:
@KyleKing The thing is if a package only has Windows wheels and no sdist, we cannot declare it as compatible with any system, we have to declare it as being a Windows only package by setting the right markers. If not, we will have an error later on when pip tries to install the package on MacOs, for instance, and can't find an appropriate file.

KyleKing:
Yeah, but I think that’s the right behavior to know that your platform isn’t supported when you run ‘poetry install’ rather than later when getting a possibly unintuitive import error

If a package author added a Windows-only package, then they are likely using it without a fallback. If they intend to support multiple platforms, they need to both make code changes to handle that importerror and indicate that the file is Windows only with a marker. Some examples are mypy and pytype that are Linux only. I tried installing them with Poetry and saw right away they didn’t support Windows when the install failed.

If you think the auto-marker-adding feature is the right forward, I’ll start a PR; however, I don’t think that new feature should hold up this PR (I have some follow up questions on the edge cases). This PR is a critical fix for non-Mac/Linux users, so it would be really beneficial to merge. There are identical changes submitted against the development branch, if you think that would be better than merging to master while the auto-marker feature is under development

(Typed on mobile, so hopefully reads well)

sdispater:
@KyleKing It all comes down to the fact that the lock file is system agnostic and, as such, can't have packages locked that are not compatible with any system. That defeats the purpose of the lock file.

What this means is we cannot rely on the current system in any way to generate the lock file. The only source of truth to generate a lock file is the `pyproject.toml` file.

Pluckerpluck:
> It all comes down to the fact that the lock file is system agnostic and, as such, can't have packages locked that are not compatible with any system.

I disagree with this statement. Specifically, I argue that it is still system agnostic even if components within it are not. The lockfile can implicitly state that a build is actually impossible on a given platform when libraries do not exist. In fact, I would go so far as to say that it's a requirement. 

Using a marker implies that the system will just not use that package if you're on the wrong platform. But that will result in the program crashing, as that was a **required** package. So all you've done is move the error elsewhere.

I believe that in the case of wheels not existing (and no sdist) poetry should error cleanly with a descriptive error message. That would make the most sense to me.

sdispater:
@Pluckerpluck 

> The lockfile can implicitly state that a build is actually impossible on a given platform when libraries do not exist

Then, by definition, it is no longer system agnostic.

> Using a marker implies that the system will just not use that package if you're on the wrong platform. But that will result in the program crashing, as that was a required package. So all you've done is move the error elsewhere.

But if you lock a package that was not meant to be locked for a specific system, it will crash at installation time because the dependency will not be found due to the fact that no suitable sdist or wheels exist.

Pluckerpluck:
> Then, by definition, it is no longer system agnostic.

The structure and contents are still system agnostic. In the same way that Java is platform agnostic yet you can use the win32 library for direct access. It will crash when you try to run it, but that's OK.

Basically, it only has to be system agnostic when the program it's building can be system agnostic.

> But if you lock a package that was not meant to be locked for a specific system, it will crash at installation time because the dependency will not be found due to the fact that no suitable sdist or wheels exist.

Which, in my opinion, is the correct time to crash. At this point the entire application is platform specific. Well, it should "exit" cleanly with a sensible error. 

Alternatively you have to embed somewhere that the **entire** application in Windows only.

KyleKing:
@sdispater what would like me to do to continue?

KyleKing:
This PR fixes the bug reported in #944 

Since the scope for this PR has been expanded to be a new feature and not just a bug fix, I'll start implementing the auto-marker assignment against the Develop branch in #1144

PR #979, opened 2019-03-22
Fix inverted --no-root cmd option

### LINKED ISSUE #973: --no-root behavior is inverted on latest develop
[This](https://github.com/sdispater/poetry/commit/37ec1447b3508ee0bbdb41f8e5773ed5bfae0654#diff-427299ba040b8502b4d29846e595c2d0R59) should probably be `if self.option("no-root")`, to _not_ install the root package when `--no-root` is provided.

seleem1337:
Fixes #973


K900:
Can confirm this fixes the behavior to be sane again.

Issue #370, opened 2018-08-04
AttributeError caused by author field, breaking `poetry build`


nackjicholson:
- [x] I am on the [latest](https://github.com/sdispater/poetry/releases/latest) Poetry version.
- [x] I have searched the [issues](https://github.com/sdispater/poetry/issues) of this repo and believe that this is not a duplicate.
- [x] If an exception occurs when executing a command, I executed it again in debug mode (`-vvv` option).

## Issue

```
poetry build -f sdist
```

With author field in pyproject.toml

```toml
authors = ["R&D <researchanddevelopment@example.com>"]
```

Output:

```
[AttributeError]               
'NoneType' object has no attribute 'group'  
```

Perhaps this is for a good reason, but it confused me for a long while because I didn't know this meant something was wrong in my pyproject.toml file. It would be useful to have indications of toml validation problems by field.

Changed to:

```
authors = ["Research Development <researchanddevelopment@example.com>"]
```

Works, and that is fine. Just didn't know there were validation rules for these strings.

kam1sh:
Faced this problem today, on version 0.12.11 (Python 3.7.2). Had this field in pyproject.toml:  
`authors = ["support@example.com]`

Changed to `authors = ["Example support <support@example.com>"]`, works fine.

I'll try to fix this in this or next weekend, because it looks like obvious regex mismatch.

kam1sh:
@nackjicholson What do you think, maybe replacing `'NoneType' object has no attribute 'group'  ` to `Invalid project's author` would be enough?

kam1sh:
Well... my PR got softly rejected, but you or someone else could find some useful solutions among my commits =/

drunkwcodes:
Thanks,  @kam1sh.

The problem is '&' mismatched in `AUTHOR_REGEX`.

Here are some common punctuation marks in names.
I would like to make a PR to add them into regex with test cases if the list has no issue.

```
,
.
’
()
-

&
@ 
# 

™
®
㊣
↗
↙

```

drunkwcodes:
And I thought this is a good place to discuss introducing `email` lib.
It handles arbitrary marks so well in the case of formal email form.

Here are example results with inputs listed above:


```
In [1]: from email.utils import parseaddr

In [2]: parseaddr('R&D <r&d@example.com>')
Out[2]: ('R&D', 'r&d@example.com')

In [3]: parseaddr('R#D <r#d@example.com>')
[...]

PR #981, opened 2019-03-22
Fix AttributeError on accessing package author (#370)

kam1sh:
# Pull Request Check List

There is fix for #370, when `poetry build` crashed on accessing incorrect author fields.

Also, I could change `AUTHOR_REGEX` to support for email-only authors, but I don't think there is need to handle those situations. At least it's not crashing right now. 

Or maybe better throw warning on authors like that?

### DIFF
--- a/tests/packages/test_package.py
+++ b/tests/packages/test_package.py
@@ -15,6 +15,7 @@ def test_package_authors():
     assert package.author_name == "John Doe"
     assert package.author_email is None
 
+
 def test_author_email_only():
     """
     Checks Package for correct handling of unparsable authors
--- a/tests/packages/test_package.py
+++ b/tests/packages/test_package.py
@@ -5,15 +5,25 @@ import pytest
 
 from poetry.packages import Package
 
+AUTHORS = (
+    "Sébastien Eustace <sebastien@eustace.io>",
+    "Sébastien Eustace<sebastien@eustace.io>",
+    "Sébastien Eustace sebastien@eustace.io",
+)
 
-def test_package_authors():
+
+@pytest.mark.parametrize("author", AUTHORS)
+def test_package_authors(author):
     package = Package("foo", "0.1.0")
 
-    package.authors.append("Sébastien Eustace <sebastien@eustace.io>")
+    package.authors.append(author)
     assert package.author_name == "Sébastien Eustace"
     assert package.author_email == "sebastien@eustace.io"
 
-    package.authors[0] = "John Doe"
+
+def test_author_name():
+    package = Package("foo", "0.1.0")
+    package.authors.append("John Doe")
     assert package.author_name == "John Doe"
     assert package.author_email is None
 
--- a/tests/packages/test_package.py
+++ b/tests/packages/test_package.py
@@ -1,26 +1,41 @@
 # -*- coding: utf-8 -*-
 from __future__ import unicode_literals
 
+import pytest
+
 from poetry.packages import Package
 
 
-def test_package_authors():
+AUTHORS = (
+    "Sébastien Eustace <sebastien@eustace.io>",
+    "Sébastien Eustace<sebastien@eustace.io>",
+    "Sébastien Eustace sebastien@eustace.io",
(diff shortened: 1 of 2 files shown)


tests/packages/test_package.py
```diff
@@ -14,3 +14,13 @@ def test_package_authors():
...
+
+
+def test_author_email_only():
+    """
+    Checks Package for correct handling of unparsable authors
+    (email only for example)
+    """
+    package = Package("foo", "0.1.0")
+    package.authors.append("<support@example.com>")
+    assert not package.author_name
```
drunkwcodes: May you add `assert package.author_email` accordingly?

kam1sh: Thanks for fast reviewing. I added `assert not package.author_email`, because parsing only email would need changing `AUTHOR_REGEX`. 
Of course, I could add something like `author, email = match.author, match.email if match else None, raw_author`, but then there could be situations when someone specified only name instead of email, so idk =/

drunkwcodes: Well... it worth changing the regex for this fix.

kam1sh: Done! I had do add additional regex pattern, because it's hard to parse `support@example.com` and `<support@example.com>` in one. But that new pattern is easy to understand.

kam1sh:
Sure, tomorrow I add that in tests and try to throw away `re`s, but idk about seventh. I think if there is comma in author, then it's user fault.
Didn't knew about mail.utils, thank you!

drunkwcodes:
It's getting worse.

**"Throwing a custom exception"** makes me feel pleasant comparing to this buggy commit. The second half of the function messed up with the wrong cases to handle.

Not to mention the coding style.

kam1sh:
Not arguing that. So I have to bring back old regex and throw `ValueError` (or something) on pattern mismatch?

I'm striving ¯\_(ツ)_/¯

drunkwcodes:
@kam1sh If I were you, I would make more discussion on the #370.

kam1sh:
After short talk I decided to keep old regex and throw exception with text `{} is not a valid author's name` when it could not be parsed. Also added test for that. 

What's next?

drunkwcodes:
@kam1sh This thread would be a real thing if and only if you fix the regex to let the special characters match as I mentioned at the first review.

drunkwcodes:
@kam1sh This is far from elegant. You should stop now or after next failure.

The regex of author name should keep being a whitelist.
The usage of `email` lib is unnecessary.

And the tests are worse thant not existed.

The PR can't be accepted at all and you kept striving to prove it.

kam1sh:
I don't know what to say.

Fine, I'm closing the PR. Sorry for wasting your time.

uSpike:
> @kam1sh This is far from elegant. You should stop now or after next failure.
> And the tests are worse than not existing.
> The PR can't be accepted at all and you kept striving to prove it.

@drunkwcodes I don't think this is a positive way to interact with someone who is willing to put in the effort of contributing.  I can't speak to the technical aspect of this thread, but I t hink this kind of messaging will turn people off to contributing.  That said, I appreciate that you're taking the time to do code reviews.

drunkwcodes:
@uSpike But it saves time. For subscribers. Please. Fight me.

uSpike:
@drunkwcodes not saying anything is better than 

@kam1sh just FYI you don't have to close this.
> @uSpike Fight me.

No.

@kam1sh feel free to reopen this PR.  The only opinion that matters for this project is @sdispater

sdispater:
@drunkwcodes Please refrain from commenting or reviewing the work of others if you can't be nice or helpful while doing so. This is not the first time I've seen you being rude here so I will only ask this once: please be considerate in your comments or do not bother writing them at all.

@kam1sh You can reopen your PR. I will review it when I have time. I know I am not the most reactive at the moment but I will eventually do it. I have the final say when it comes to accepting or rejecting PRs, so as long as you haven't heard from me you can leave it open.

PR #1008, opened 2019-04-02
Added settings.prepend-index.url configuration

mellon85:
# Pull Request Check List

Added settings.prepend-index.url to add a repository url can be added to the set of indexes used by setting this configuration. It will be prepended and so preferred over other sources of packages.

### DIFF
--- a/docs/docs/cli.md
+++ b/docs/docs/cli.md
@@ -14,7 +14,6 @@ poetry up
 
 calls `poetry update`.
 
-
 ## Global options
 
 * `--verbose (-v|vv|vvv)`: Increase the verbosity of messages: "-v" for normal output, "-vv" for more verbose output and "-vvv" for debug.
--- a/docs/docs/configuration.md
+++ b/docs/docs/configuration.md
@@ -31,6 +31,12 @@ Defaults to one of the following directories:
 - Windows: `C:\Users\<username>\AppData\Local\pypoetry\Cache/virtualenvs`
 - Unix:    `~/.cache/pypoetry/virtualenvs`
 
+### settings.prepend-index.url : string
+
+A repository url can be added to the set of indexes used by setting this
+configuration. It will be prepended and so preferred over other sources of
+packages.
+
 ### `repositories.<name>`: string
 
 Set a new alternative repository. See [Repositories](/docs/repositories/) for more information.
--- a/poetry/console/commands/config.py
+++ b/poetry/console/commands/config.py
@@ -67,6 +67,7 @@ To remove a repository (repo is a short alias for repositories):
                 lambda val: str(Path(val)),
                 str(Path(CACHE_DIR) / "virtualenvs"),
             ),
+            "settings.prepend-index.url": (str, str, ""),
         }
 
         return unique_config_values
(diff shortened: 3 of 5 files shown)


mellon85:
@sdispater  Would like a suggestion on where I could add tests for verifying this option

drunkwcodes:
The repository setting is not a temporary setting, it had better appear in the config file and not in env.

mellon85:
@drunkwcodes This setting is not meant to replace the setting in the toml file, it's meant to be used to be able to switch to use a local cache if needed.

If the needed index is not reachable at all times (VPN, flaky/slow network) devpi can be used as a local cache (i.e. devpi) to let it use the local data when the upstream server is unreachable.
It's akin to pipenv PIP_INDEX_URL/PIP_EXTRA_INDEX_URL/--pypi-mirror

I should have clarified it in the pull request.

drunkwcodes:
I know. But it does effect the behavior of the program in a implicit and unreasonable manner.

Reference:
https://stackoverflow.com/questions/32272911/precedence-of-configuration-options-environment-registry-configuration-file-a

mellon85:
The stack overflow post doesn't really bring up that point. The issue at
hand is working on many projects and not wanting to change the repository
configuration that would be committed to use a local cache.

I agree it might be too implicit, but then I can also log it to inform the
user that the environment variable was found and is being used.

Il gio 4 apr 2019, 00:29 drunkwcodes <notifications@github.com> ha scritto:

mellon85:
@drunkwcodes Thought about it and it made sense to have it in the configuration file. I've updated my pull request. Still would need some guidance on how to test this change.

drunkwcodes:
@mellon85 What about https://poetry.eustace.io/docs/repositories/#using-a-private-repository ?

mellon85:
@drunkwcodes that's not the use case, as explained before this is to be used for pypi caches like devpi

drunkwcodes:
Yes it is.

mellon85:
@drunkwcodes I don't think so.

If I use a local cache I am not going to change project data inside my repository.
Adding a source to a project makes sense if it's going to be available to anybody cloning the project, as for instance a private index.
What this feature is intended for is for local caches to avoid slow downloads or unavailable sources (as they might accessible only via VPN, or when traveling and having no network).

How would your suggestion actually solve this situation?

drunkwcodes:
@mellon85 I didn't figure out the difference of effects between this PR and existing function.

Would you explain it again?

Pluckerpluck:
Imagine a situation where you can only access the internet 20% of the time. As a result, you use devpi to locally cache PyPi **for your use only**. This allows you to achieve a better effective uptime than you otherwise would have.

As this is only for your own use, it should **not** be added into the pyproject.toml file. It instead should be a global setting.

---------

Whether this should be included as a feature is a different question, but the concept of local caching is a reasonable one.

I also have an alternate use case where I want to temporarily override a version of a dependency. Being able to insert my own index would make life easier in that regard.

marco-silva0000:
I've opened #1105 because of a similar issue with devpi. commenting just to reference it.

Secrus:
This has drifted to far from current codebase and features. Closing.

Issue #1031, opened 2019-04-12
poetry add/update/install can fail on Windows during temp directory cleanup


Pluckerpluck:
If something goes wrong during an installation, a file is often held open for just a little too long for `TemporaryDirectory` causing it to error out with errors such as:

```
[OSError]
[WinError 145] The directory is not empty: 'C:\\Users\\KING~1.KYL\\AppData\\Local\\Temp\\tmp774m6ge1'

...

 C:\Users\king.kyle\AppData\Local\Continuum\anaconda2\envs\py37\lib\contextlib.py in __exit__() at line 119
   next(self.gen)
 C:\Users\king.kyle\.poetry\lib\poetry\utils\helpers.py in temporary_directory() at line 35
   yield name
 C:\Users\king.kyle\AppData\Local\Continuum\anaconda2\envs\py37\lib\tempfile.py in __exit__() at line 805
   self.cleanup()
 C:\Users\king.kyle\AppData\Local\Continuum\anaconda2\envs\py37\lib\tempfile.py in cleanup() at line 809
   _shutil.rmtree(self.name)
 C:\Users\king.kyle\AppData\Local\Continuum\anaconda2\envs\py37\lib\shutil.py in rmtree() at line 507
   return _rmtree_unsafe(path, onerror)
 C:\Users\king.kyle\AppData\Local\Continuum\anaconda2\envs\py37\lib\shutil.py in _rmtree_unsafe() at line 395
   onerror(os.rmdir, path, sys.exc_info())
 C:\Users\king.kyle\AppData\Local\Continuum\anaconda2\envs\py37\lib\shutil.py in _rmtree_unsafe() at line 393
   os.rmdir(path)

add [-D|--dev] [--git GIT] [--path PATH] [-E|--extras EXTRAS] [--optional] [--python PYTHON] [--platform PLATFORM] [--allow-prereleases] [--dry-run] [--] <name> (<name>)...
```

This is an **incredibly** unhelpful error and masks all sorts of other problems.

It is causes by a bug in `shutil.rmtree` which can fail on Windows from time to time. It can fail for a variety of reasons, but very commonly it's simply a process holding the file active for just a little too long resulting in `os.rmdir` failing. This is despite the directory actually being empty when you check (i.e. it's just slow).

This is a known error in Python, of which there is no plan to fix: https://bugs.python.org/issue29982

[...]

PR #1032, opened 2019-04-12
Fix rmtree failures on Windows (#1031)

Pluckerpluck:
Fixes https://github.com/sdispater/poetry/issues/1031

-------

- [x] Added **tests** for changed code. (N/A)
- [x] Updated **documentation** for changed code. (N/A)

Existing tests cover the code change + no need for documentation.

If you **really** want a test that attempts to simulate the flaky behavior for windows I can try to write one, but I can't see any sensible way to do that. So potentially change `robust_rmtree` to a private function to indicate no need for direct testing.

---------

Avoid the use of `TemporaryDirectory`, instead manually dealing with temporary files to allow custom cleanup.

Retrying `shutil.rmtree` multiple times avoids the removal race conditions found in Windows causing:

1. Irrelevant errors when other failures cause removal delays
2. Failures during arbitrary installs caused by antivirus software

### DIFF
--- a/poetry/utils/helpers.py
+++ b/poetry/utils/helpers.py
@@ -3,6 +3,7 @@ import re
 import shutil
 import stat
 import tempfile
+import time
 
 from contextlib import contextmanager
 from typing import List
@@ -29,17 +30,29 @@ def normalize_version(version):  # type: (str) -> str
 
 @contextmanager
 def temporary_directory(*args, **kwargs):
-    try:
-        from tempfile import TemporaryDirectory
+    name = tempfile.mkdtemp(*args, **kwargs)
 
-        with TemporaryDirectory(*args, **kwargs) as name:
-            yield name
-    except ImportError:
-        name = tempfile.mkdtemp(*args, **kwargs)
+    yield name
 
-        yield name
+    robust_rmtree(name)
 
-        shutil.rmtree(name)
+
+def robust_rmtree(path, max_retries=3):
+    """Robustly tries to delete paths.
+
+    Retries several times if an OSError occurs.
+    If the final attempt fails, the Exception is propagated
+    to the caller.
+    """
+    for i in range(max_retries - 1):
+        try:
+            shutil.rmtree(path)
+            return  # Only hits this on success
+        except OSError:
+            time.sleep(2)
+
+    # Final attempt, pass any Exceptions up to caller.
+    shutil.rmtree(path)
 
 
 def parse_requires(requires):  # type: (str) -> List[str]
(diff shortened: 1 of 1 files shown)


absassi:
@sdispater, can we have this fix in the next version? I'm having to manually apply this patch on Poetry, because otherwise I cannot use Poetry in a large project I work on (since there are many dependencies, this error almost always happens in some random package during dependency resolution).

poetry/utils/helpers.py
```diff
@@ -29,17 +30,32 @@ def normalize_version(version):  # type: (str) -> str
...
+
+def robust_rmtree(path):
+    """Robustly tries to delete paths.
+
+    Retries several times if an OSError occurs.
+    If the final attempt fails, the Exception is propagated
+    to the caller.
+    """
+    timeout = 0.001
+    while timeout < 2:
```
brycedrennan: why 2 seconds instead 1 like the python version? Without a strong justification otherwise, I think we should use `1.0`

poetry/utils/helpers.py
```diff
@@ -29,17 +30,32 @@ def normalize_version(version):  # type: (str) -> str
...
-        name = tempfile.mkdtemp(*args, **kwargs)
+    yield name
 
-        yield name
+    robust_rmtree(name)
 
-        shutil.rmtree(name)
+
+def robust_rmtree(path):
+    """Robustly tries to delete paths.
```
brycedrennan: to match the style of the codebase, please put this on a newline

poetry/utils/helpers.py
Pluckerpluck: I was running 2 seconds simply becuase I felt that it wasn't an unreasonable wait that little extra longer to increase robustness when this is simply gating a crash situation (i.e. the program will error at the end of the timeout if it fails).

CPython is using it in their tests, so obviously they are much more concious of the tradeoff between robustness and speed.

That being said, 1s is almost certainly enough so I'm fine switching it to that. And we can increase it in the future if people still face the issue.

Pluckerpluck:
I'll make the changes when I get time. Likely this weekend.

sbethur:
I think #1140 is attempting to fix the same issue.

finswimmer:
Hello,

with the changes introduced in https://github.com/python-poetry/poetry/commit/d22afa9d933529c01b74a6127fa41ce49bf0182b there is at least a second try if something went wrong. Maybe that's enough for the problem that this PR is trying to solve?

Closing this for now. @Pluckerpluck please rebase, adopt your changes and leave a comment if you think that this is still neccessary.

Thanks a lot for your contribution! :+1: 

fin swimmer

eblis:
This still occurs with Poetry 1.1.12 on Windows :(
Can we get this merged please ?

antaradas94:
Actually, the error also continues with Poetry 1.3.0 on Windows. 😞

Issue #1015, opened 2019-04-03
[Request] Add a output file option for the export command


NyanKiyoshi:
- [x] I have searched the [issues](https://github.com/sdispater/poetry/issues) of this repo and believe that this is not a duplicate.
- [x] I have searched the [documentation](https://poetry.eustace.io/docs/) and believe that my question is not covered.

## Feature Request
Hello, I think it would be very handy to have a `-o` parameter in poetry's `export` command so we could be able to tell poetry to which file we want to export the requirements, independently from the format (which as of now, the only format supported is `requirements.txt`).

So we could prevent doing such things:
```
poetry export --dev -f requirements.txt && mv requirements.txt requirements_dev.txt \
    && poetry export -f requirements.txt
```

Instead, we would have:
```
poetry export --dev -f requirements.txt -o requirements_dev.txt \
    && poetry export -f requirements.txt
```

Also, it would be completely cross platform.

bersace:
Funny. You posted this ticket just 45 minutes after #1014 on the same subject.

NyanKiyoshi:
Closed by #1035, thanks @richin13!

PR #1035, opened 2019-04-12
Add `-o` flag for `export` command

yggi49:
The export command now prints to standard output by default.  To store the
output in a file, the new `-o <filename>` argument can be used.

# Pull Request Check List


poetry/utils/exporter.py
```diff
@@ -59,5 +60,13 @@ def _export_requirements_txt(
             line += "\n"
             content += line
 
-        with filepath.open("w", encoding="utf-8") as f:
-            f.write(decode(content))
+        self._output(content, cwd, output)
+
+    def _output(self, content, cwd, output=None):  # type: (str, Path, str) -> None
+        decoded = decode(content)
+        if output is not None:
```
drunkwcodes: Please write `if output is None` condition first because it's default value.
```suggestion
        if output is None:
            sys.stdout.write(decoded)
```

poetry/utils/exporter.py
```diff
@@ -14,19 +16,18 @@ def __init__(self, lock):  # type: (Locker) -> None
         self._lock = lock
 
     def export(
-        self, fmt, cwd, with_hashes=True, dev=False
-    ):  # type: (str, Path, bool, bool) -> None
+        self, fmt, cwd, output=None, with_hashes=True, dev=False
+    ):  # type: (str, Path, str, bool, bool) -> None
```
richin13: It's actually an `Optional[str]`

richin13:
This addresses #1015, please reference existing issues so that others are aware of the work being done :sweat_smile:

yggi49:
> This addresses #1015, please reference existing issues so that others are aware of the work being done

Thank you for pointing me; I obviously missed this.

yggi49:
> LGTM with a minor suggestion in regard of the usage: `poetry export > requirements.txt`.

Thank you for the input; fixed with commit https://github.com/sdispater/poetry/pull/1035/commits/e687a22c0a2e149b965a20e0361385881ecb65ee.

poetry/utils/exporter.py
yggi49: Good point; fixed with commit https://github.com/sdispater/poetry/pull/1035/commits/36cb4300894f9fb6114971df1b546db31fb46f4e.

poetry/utils/exporter.py
yggi49: Learned something new again; thank you! Fixed with commit https://github.com/sdispater/poetry/pull/1035/commits/36cb4300894f9fb6114971df1b546db31fb46f4e.

poetry/utils/exporter.py
```diff
@@ -17,7 +17,7 @@ def __init__(self, lock):  # type: (Locker) -> None
 
     def export(
         self, fmt, cwd, output=None, with_hashes=True, dev=False
-    ):  # type: (str, Path, str, bool, bool) -> None
+    ):  # type: (str, Path, Optional[str], bool, bool) -> None
```
richin13: Just make sure you import `Optional` from the typing module

yggi49:
Note: if this gets accepted, I would be inclined to submit an additional MR afterwards that makes the `-f` flag optional, with `requirements.txt` assumed to be the default.  Would that be of interest?

richin13:
Pinging @sdispater!

tests/console/commands/test_export.py
```diff
@@ -131,3 +131,25 @@ def test_export_fails_on_invalid_format(app, repo):
...
+    command = app.find("export")
+    tester = CommandTester(command)
+
+    tester.execute("--format requirements.txt")
+
+    expected = """\
+foo==1.0.0
+"""
+
+    out, err = capsys.readouterr()
```
sdispater: You don't need to use `capsys` here. You can use `tester.io.fetch_output()`.

poetry/console/commands/export.py
```diff
@@ -46,6 +49,7 @@ def handle(self):
         exporter.export(
```
sdispater: I think we can pass the command's io to the exporter. This will help testing the write to stdout without relying to the actual stdout.

Issue #798, opened 2019-01-14
Special characters not supported in the authors section of pyproject.toml


taljaards:
- [x] I am on the [latest](https://github.com/sdispater/poetry/releases/latest) Poetry version.
- [x] I have searched the [issues](https://github.com/sdispater/poetry/issues) of this repo and believe that this is not a duplicate.
- [x] If an exception occurs when executing a command, I executed it again in debug mode (`-vvv` option).

- **OS version and name**: Windows 10 17763.253
- **Poetry version**: 0.12.11
- **pyproject.toml file**: https://gist.github.com/taljaards/b06c1dda1ece0b8d443eb9512213e47d

## Issue
Poetry does not support special characters in the "authors" section of the `pyproject.toml` file.
I created a new folder, did `poetry init` and answered no to most of the yes/no questions.

Afterwards, when doing `poetry add`, it fails:
```
C:\Users\Stephan\Source\jupyter-lab>poetry add jupyterlab -vvv

[UnicodeDecodeError]
'utf-8' codec can't decode byte 0xe9 in position 88: invalid continuation byte

Exception trace:
 C:\Users\Stephan\.poetry\lib\poetry\_vendor\py3.7\cleo\application.py in run() at line 94
   status_code = self.do_run(input_, output_)
 C:\Users\Stephan\.poetry\lib\poetry\console\application.py in do_run() at line 88
   return super(Application, self).do_run(i, o)
 C:\Users\Stephan\.poetry\lib\poetry\_vendor\py3.7\cleo\application.py in do_run() at line 197
   status_code = command.run(input_, output_)
 C:\Users\Stephan\.poetry\lib\poetry\console\commands\command.py in run() at line 77
   return super(BaseCommand, self).run(i, o)
 C:\Users\Stephan\.poetry\lib\poetry\_vendor\py3.7\cleo\commands\base_command.py in run() at line 136
   self.initialize(input_, output_)
 C:\Users\Stephan\.poetry\lib\poetry\console\commands\env_command.py in initialize() at line 18
   current_env = Env.get(self.poetry.file.parent)
 C:\Users\Stephan\.poetry\lib\poetry\console\commands\command.py in poetry() at line 62
   return self.get_application().poetry
 C:\Users\Stephan\.poetry\lib\poetry\console\application.py in poetry() at line 60
   self._poetry = Poetry.create(os.getcwd())
 C:\Users\Stephan\.poetry\lib\poetry\poetry.py in create() at line 98
[...]

PR #1040, opened 2019-04-16
Fix parsing of author information

yggi49:
Instead of relying on regular expressions, this patch leverages Python’s
builtin `email.utils.parseaddr()` functionality to parse an RFC-822-compliant
email address string into its name and address parts.

This should also resolve issues with special characters in the name part; see
issues #370 and #798.

# Pull Request Check List


poetry/utils/helpers.py
```diff
@@ -100,3 +101,21 @@ def _on_rm_error(func, path, exc_info):
...
+    .. note::
+
+       If the input string does not contain an `@` character, it is
+       assumed that it represents only a name without an email address.
+
+    :param address: the email address string to parse.
+    :return: a 2-tuple with the parsed name and email address.  If a
+             part is missing, ``None`` will be returned in its place.
+    """
+    if "@" not in address:
```
drunkwcodes: Considering comma-separated inputs( In [15]: ), please put this below `parseaddr()` if not raising exception.
https://github.com/sdispater/poetry/pull/981#pullrequestreview-218088417

```suggestion
    name, email = parseaddr(address)
    if not name and "@" not in address:
        return (email, None)
```

tests/utils/test_helpers.py
```diff
@@ -66,3 +70,51 @@ def test_get_http_basic_auth_without_password(config):
 
 def test_get_http_basic_auth_missing(config):
     assert get_http_basic_auth(config, "foo") is None
+
+
+def test_parse_author_simple_name_and_email():
```
drunkwcodes: Please write tests in a function named `test_parse_author` and separate cases with comments.

It's good to organize code and tests with the same structure.

poetry/utils/helpers.py
yggi49: If I’m doing this, then just providing a full name without an email address does not work any more, as evidenced by the failure of the existing `test_package_authors()` test as well as one of the test cases that are part of this PR:

```
    def test_package_authors():
        package = Package("foo", "0.1.0")

        package.authors.append("Sébastien Eustace <sebastien@eustace.io>")
        assert package.author_name == "Sébastien Eustace"
        assert package.author_email == "sebastien@eustace.io"

        package.authors.insert(0, "John Doe")
>       assert package.author_name == "John Doe"
E       AssertionError: assert 'John' == 'John Doe'
E         - John
E         + John Doe

tests/packages/test_package.py:15: AssertionError

[...]

    def test_parse_author_simple_name_only():
        name, email = parse_author("John Doe")
>       assert name == "John Doe"
E       AssertionError: assert 'John' == 'John Doe'
E         - John
E         + John Doe

tests/utils/test_helpers.py:83: AssertionError
```

This is due to the fact that Python’s `parseaddr()` does the following for `John Doe`:

```Python console
>>> parseaddr('John Doe')
('', 'John')
```

Hence I went the pragmatic route and simply checked for the presence of an “@” character in the provided input string, as I would assume that otherwise the user’s intention was to specify a name only.  (Apart from that, I believe—at least for Poetry’s purposes—that any author email address should always contain an “@” character.)

Having a comma character in the input string might happen of course.  I’d consider this a rather quirky edge case, though, so I think the benefits of this PR outweigh the drawbacks.

If this really is a blocker for this PR to get merged, it would be great if you could specify what `parse_author()` is supposed to return for different kinds of input strings that contain a comma.

tests/utils/test_helpers.py
yggi49: I do not think this is a good idea.  Each of the eight `test_parse_author_*` functions that I added represents a semantically self-contained test.

Imagine the implementation of `parse_author()` was changed in a way that would cause four of these tests to fail.  If all the assertions were packed together in a single test function, running the tests would reveal only the first one of these failures.

Therefore, if it is merely for the purpose of test organization, I would rather group the related tests into a separate class, but still keep the distinct test functions:

```python
class TestParseAuthor(object):
    """Tests for the :func:`parse_author` function."""

    def test_simple_name_and_email(self):
        [...]

    def test_simple_name_only(self):
        [...]
```

Of course I am not the one to make the calls here. If you prefer having everything in a single `test_parse_author()` function, so be it—just let me know. =)

drunkwcodes: I still think those cases had better be in a function. We can sort out those cases by precedence and treat them as all-or-none cases.

The cases which are not formal as documented should be placed at last.

poetry/utils/helpers.py
drunkwcodes: I see. Thanks for illustration. Now it only needs check `'@'` twice in email part, before and after `parsesddr()` to cover this two cases.

And `str(None) == 'None'` is also a concern for who uses this as the input to database.
So I followed the empty string style same as `parseaddr()`

tests/utils/test_helpers.py
yggi49: Then, so be it—commit 619593c5ef645285285267844f1ca52445df934f refactors all the tests for `parse_author()` into a single function.

poetry/utils/helpers.py
yggi49: > And `str(None) == 'None'` is also a concern for who uses this as the input to database.
> So I followed the empty string style same as `parseaddr()`

I am well aware that `str(None)` is not always useful.  Initially, I actually had `parse_author()` return empty strings for missing fields.  However, the already existing `test.packages.test_package.test_package_authors()` test contains the following (note the final assertion):

```python
def test_package_authors():
    package = Package("foo", "0.1.0")

    [...]

    package.authors.insert(0, "John Doe")
    assert package.author_name == "John Doe"
    assert package.author_email is None
```

Therefore, I assumed that this was the expected way to go for Poetry, and I changed my implementation to returning `None`.

If this basic behavior should be changed to using empty strings instead of `None`, it would probably require a thorough evaluation first on how far reaching consequences this would entail.  However, I believe that this is out of scope for this PR and should be handled separately if so desired.  After all, I only stumbled upon this issue because of having a fancy unicode character in my company’s name that Poetry was unable to deal with. ;-)

drunkwcodes: It can be another issue.

yggi49: > I see. Thanks for illustration. Now it only needs check `'@'` twice in email part, before and after `parsesddr()` to cover this two cases.

Okay—just to be on the same page—so this is how you envision `parse_author()` to handle the funky use cases?

```python console
>>> parse_author("asf,dfu@t.b")
(None, 'asf')
>>> parse_author("asf,<dfu@t.b>")
(None, 'asf')
>>> parse_author("asf, dfu@t.b")
(None, 'asf')
```

drunkwcodes: After a quick search in this project, what matters the most is this condition statement: https://github.com/sdispater/poetry/blob/2f2cec03b6b14e882522f7f6042ecdafebda2ca5/poetry/layouts/layout.py#L71


Though the author name is usually initialized as `None`.

I will open another issue and PR after a deeper investigation

drunkwcodes: > > I see. Thanks for illustration. Now it only needs check `'@'` twice in email part, before and after `parsesddr()` to cover this two cases.
> 
> Okay—just to be on the same page—so this is how you envision `parse_author()` to handle the funky use cases?
> 

What I expected is:

 ```python
>>> parse_author("asf,dfu@t.b")
('asf'. '')
>>> parse_author("asf,<dfu@t.b>")
('asf', '')
>>> parse_author("asf, dfu@t.b")
('asf', '')
```

`parse_author()` will place those obviously non-email information into author name part.
Though it loses the real email.

But sometimes it doesn't really need full information.

And considering the impact including in breaking tests and further discussion, I thought this PR is better to be like this to solve the issue immediately.

Comment (yggi49) [reply poetry/utils/helpers.py]: Thank you for elaborating.  Please see commit edba8fd that implements this extended behavior!

File: poetry/utils/helpers.py [hunk 3]
```diff
@@ -100,3 +101,26 @@ def _on_rm_error(func, path, exc_info):
...
+    .. note::
+
+       If the input string does not contain an ``@`` character, it is
+       assumed that it represents only a name without an email address.
+
+    :param address: the email address string to parse.
+    :return: a 2-tuple with the parsed name and email address.  If a
+             part is missing, ``None`` will be returned in its place.
+    """
+    if "@" not in address:
```
drunkwcodes:
```suggestion
    if '@' not in address:
    return (address, None)
    tokens = re.split(r"[ ,<]+", address)
    author = ' '.join(tokens[:-1])
    email = tokens[-1]
    return (author or None, parseaddr(email)[1] or None)
```

tests/utils/test_helpers.py
```diff
@@ -66,3 +70,60 @@ def test_get_http_basic_auth_without_password(config):
...
+
+    # Email address only
+    name, email = parse_author("john.doe@example.com")
+    assert name is None
+    assert email == "john.doe@example.com"
+
+    # Non-RFC-conform cases with unquoted commas
+    name, email = parse_author("asf,dfu@t.b")
+    assert name == "asf"
+    assert email is None
```
drunkwcodes:
```suggestion
    assert email == "dfu@t.b"
```

tests/utils/test_helpers.py
```diff
@@ -66,3 +70,60 @@ def test_get_http_basic_auth_without_password(config):
...
+
+    name, email = parse_author("asf,<dfu@t.b>")
+    assert name == "asf"
+    assert email is None
```
drunkwcodes:
```suggestion
    assert email == "dfu@t.b"
```

tests/utils/test_helpers.py
```diff
@@ -66,3 +70,60 @@ def test_get_http_basic_auth_without_password(config):
...
+
+    name, email = parse_author("asf, dfu@t.b")
+    assert name == "asf"
+    assert email is None
```
drunkwcodes:
```suggestion
    assert email == "dfu@t.b"
```

poetry/packages/package.py
```diff
@@ -1,6 +1,5 @@
 # -*- coding: utf-8 -*-
 import copy
-import re
```
drunkwcodes: It needs `re`.

tests/utils/test_helpers.py
drunkwcodes: Now `@ID` is supported. IPython somewhat needs '@' completion.

```
In [5]: parse_author("@asf")
Out[5]: ("@asf", None)

In [6]: parse_author("@asf <asf@t.b>")
Out[6]: ("@asf", "asf@t.b")

In [7]: parse_author("@asf,<asf@t.b>")
Out[7]: ("@asf", "asf@t.b")

In [8]: parse_author("@asf,asf@t.b")
Out[8]: ("@asf", "asf@t.b")

In [9]: parse_author("@asf, asf@t.b")
Out[9]: ("@asf", "asf@t.b")

In [10]: parse_author("@asf,")
Out[10]: ("@asf", None)

In [11]: parse_author("@asf <>")
Out[11]: ("@asf", None)
```

drunkwcodes:
Here is the latest revision in my gist with runnable tests including in `@ID`, multi-commas, white-spaces cases and empty email cases:

https://gist.github.com/drunkwcodes/d2551bc41f54a1120434ad8f24dc49af

yggi49:
Is it really necessary to attempt to make *something* out of even the most obscure construed use cases?  For example, I don’t quite get why

```python
parse_author("@asf,asf@t.b")
```

should result in

```python
("@asf", "asf@t.b")
```

To me, that seems like a totally arbitrary, non-obvious choice.

—First and foremost, this PR attempts to resolve (in a simple way) the two issues that are linked at the top, which it does.  Everything beyond it might be something for a separate feature request.

drunkwcodes:
@yggi49 The cases needs to be polished of course. 

The problem was the unclear exception. But I want to eliminate it.
And I think it is obvious if there isn't any comma (or certain ascii code punctuations) in the email address.

After all, the change will enable arbitary input in `pyproject.toml` author field without any failure and the information will be reserved even when the function failed to parse.

yggi49:
> After all, the change will enable arbitrary input in `pyproject.toml` author field without any failure and the information will be reserved even when the function failed to parse.

I beg to disagree.  Just as licenses are validated, I think it makes totally sense to validate the “author” field’s input as well, and make users aware and actively inform/educate them if their input does not correspond to a given, _very_ common and widespread format.

That said, this PR is still lacking a second reviewer—maybe there is a third opinion here?

drunkwcodes:
@yggi49 The field may be referred to `Author-email`. https://www.python.org/dev/peps/pep-0345/#author-optional

The author field doesn't have the constraint by itself. And both fields are optional.

Maybe we can put the spell checking in `poetry init`, letting those `pyproject.toml` crafted by hands pass if the `Author` field has minimun information at least?

Secrus:
Is there still interest in bringing this to Poetry? PR seems stale.

yggi49:
I am definitely interested.

My main motivation was to support special characters in author names, e.g. `my·fancy·company <development@my-fancy-company.example.com>`; the initially suggested solution did just that.

However, as [this](https://github.com/python-poetry/poetry/pull/1040#issuecomment-500136457) [conversation](https://github.com/python-poetry/poetry/pull/1040#issuecomment-500170584) [evolved](https://github.com/python-poetry/poetry/pull/1040#issuecomment-502293359), it seems that Poetry shall also support various other ways of data input for the author/email field.  If someone could provide a concrete specification of which use cases shall be covered in which ways, I will gladly continue working on the implementation.

Secrus:
@python-poetry/core can I please get some eyes on that? seems consensus is needed before this could progress.

yggi49:
@neersighted, why was this PR closed without further feedback?  As stated, I am willing to contribute on improving author information parsing; however, I will need direction as my initial solution was not deemed sufficient.

neersighted:
> @neersighted, why was this PR closed without further feedback? As stated, I am willing to contribute on improving author information parsing; however, I will need direction as my initial solution was not deemed sufficient.

Hi, I have been doing some branch cleanup and it looks like this was an unintended casualty as it was opened against an ancient, stale branch. It looks like Github does not allow me to retarget this PR against master, so a new PR should be opened with a link to this one for the previous discussion.

yggi49:
For the new PR: Does it still make sense to have the functionality work with Poetry 1.1, with Poetry 1.2 looming on the horizon?  Or should I just target the 1.2 line?

Secrus:
> For the new PR: Does it still make sense to have the functionality work with Poetry 1.1, with Poetry 1.2 looming on the horizon? Or should I just target the 1.2 line?

Target `master` branch, we will handle potential backports to 1.2 if needed.

PR #1072, opened 2019-05-02
Return a clear error when HTTP calls to legacy repo fail, instead of PackageNotFound

MarcDufresne:
This PR aims to make Poetry a bit more clear about issues when communicating with a legacy repo. 

For example, it happened a few times that we would get a `PackageNotFound` error when trying to install a package from a private repo, when in fact the actual issue was bad credentials. The simple fix is simply to return the HTTP error message instead of a generic `PackageNotFound`.

With this change you will now see:

```
[RepositoryError]                                                              
401 Client Error: Unauthorized for url: https://my-repo.com/repo/my-package/
```

Instead of:

```
[PackageNotFound]        
Package [my-package] not found. 
```

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -34,11 +34,6 @@ optional = false
 python-versions = ">=2.7, !=3.0.*, !=3.1.*, !=3.2.*, !=3.3.*"
 version = "19.1.0"
 
-[package.extras]
-dev = ["coverage", "hypothesis", "pympler", "pytest", "six", "zope.interface", "sphinx", "pre-commit"]
-docs = ["sphinx", "zope.interface"]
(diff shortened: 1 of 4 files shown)


MarcDufresne:
Done! I was hesitant to add tests for `_get` and `_download` since they were not already tested and didn't want to take design decisions regarding testing. 
However since you asked for them, I took the liberty of adding `requests-mock` and test both happy and sad path for both methods, let me know what you think.

dan-hipschman-od:
Any objection to reopening and merging this? I was about to make a PR for the same improvement, but this one seems perfectly good. Thanks!

finswimmer:
Closing this in favor of https://github.com/python-poetry/poetry/pull/2459

PR #1075, opened 2019-05-04
Use subprocess.run() to properly capture return codes.

### LINKED ISSUE #1074: Env.run doesn't check return code if input is passed in

- **OS version and name**:  macOS 10.14.3
- **Poetry version**: 0.12.15

## Issue

Env.run() is used internally for running a number of scripts to introspect the managed virtualenv and contained python. These scripts are passed as input to the function and then to subprocess.communicate(). However, in the case of an error from the spawned python process, the return code is not checked and any error text is passed back to the caller - which will assume its valid output. This makes for confusing error messages from poetry - making it harder to diagnose issues like #654 and #1071

Imaclean74:
Fixes #1074

# Pull Request Check List

- [ ] Updated **documentation** for changed code. NA - no functionality changed.

Adding a fix for #1074. 

A couple of notes on the implementation :
-  subprocess.communicate does not raise a CalledProcessException if the subprocess returns a non-zero return code. However subprocess.run() does - if you pass check=True.
- subprocess.run()  is not available on Python 3.4 or earlier - however, we can use the subprocess32 backport. in fact, the code to add the equivalent check around the existing communicate() call wasn't extensive - but this way uses the preferred mechanism ( subprocess.run() ) and we don't need to effectively re-implement subprocess.run() in env.py.
- I leveraged the existing approach in _compat.py to pull in the backport - ie similar to pathlib2.


Imaclean74:
removing WIP status - this should be good to go. One enhancement I'd like would be to have a test case demonstrating the invalid venv case - however, I'm not able to repro it on my mac machine :(

Imaclean74:
@sdispater - any chance of getting this one looked at ? The issue is masking at least a couple of other issues that are manifesting as JsonDecode errors.

sdispater:
Thanks!

PR #1082, opened 2019-05-06
Use consistent package naming in example

thejohnfreeman:
The description of the last example doesn't match the example, and the name `mypackage` doesn't match the style of `extra_package`, so I just made it `my_package` everywhere.


sdispater:
Thanks!

Issue #1087, opened 2019-05-07
Review of (explicit) encoding for files being created in text mode


vlcinsky:
- **OS version and name**: Windows and other systems not using UTF-8 default encoding
- **Poetry version**: 0.12.15

Resolving issue #1027 by PR #1085 we have found a pattern, which might cause `poetry` to fail on systems, which are not using default utf-8 encoding.

The pattern is opening files for writing in text mode without explicit encoding.

# Possibly correct default encoding
Some files are written and used on the same system, thus shall be fine using their own default encoding for writing and reading the content.

Here are the candidates:
```
get-poetry.py|564 col 18| with open(os.path.join(POETRY_BIN, "poetry.bat"), "w") as f:
get-poetry.py|573 col 14| with open(os.path.join(POETRY_BIN, "poetry"), "w") as f:
get-poetry.py|585 col 14| with open(os.path.join(POETRY_HOME, "env"), "w") as f:
get-poetry.py|606 col 18| with open(profile, "r") as f:
get-poetry.py|610 col 22| with open(profile, "a") as f:
get-poetry.py|703 col 18| with open(profile, "r") as f:
get-poetry.py|719 col 18| with open(profile, "w") as f:
```
# Files which shall use explicit encoding
Some files shall have specific encoding by default. Using default encoding is probably an error.

## Candidates in `get-poetry`
```
get-poetry.py|390 col 18| with open(os.path.join(POETRY_LIB, "poetry", "__version__.py")) as f:
```

## Candidates in core code:
```
poetry/spdx/__init__.py|29 col 10| with open(licenses_file) as f:
poetry/spdx/updater.py|23 col 14| with open(file, "w") as f:
poetry/spdx/updater.py|30 col 17| with urlopen(url) as r:
poetry/installation/pip_installer.py|196 col 18| with open(setup, "w") as f:
poetry/layouts/src.py|17 col 27| with package_init.open("w") as f:
poetry/layouts/layout.py|143 col 28| with tests_default.open("w") as f:
poetry/layouts/layout.py|155 col 21| with poetry.open("w") as f:
poetry/layouts/standard.py|17 col 27| with package_init.open("w") as f:
poetry/json/__init__.py|22 col 10| with open(schema) as f:
poetry/puzzle/provider.py|349 col 39| with requires.open() as f:
poetry/repositories/pypi_repository.py|536 col 35| with requires.open() as f:
poetry/repositories/pypi_repository.py|548 col 35| with requires.open() as f:
[...]

PR #1088, opened 2019-05-07
Review of (explicit) encoding for files being created in text mode

vlcinsky:
see issue #1087 

Includes changes for PR #1085

### DIFF
--- a/get-poetry.py
+++ b/get-poetry.py
@@ -206,7 +206,7 @@ if __name__ == "__main__":
     main()
 """
 
-BAT = '@echo off\r\npython "{poetry_bin}" %*\r\n'
+BAT = u'@echo off\r\npython "{poetry_bin}" %*\r\n'
 
 
 PRE_MESSAGE = """# Welcome to {poetry}!
--- a/get-poetry.py
+++ b/get-poetry.py
@@ -191,6 +191,7 @@ POETRY_LIB_BACKUP = os.path.join(POETRY_HOME, "lib-backup")
 
 
 BIN = """#!/usr/bin/env python
+# -*- coding: utf-8 -*-
 import glob
 import sys
 import os
@@ -387,7 +388,10 @@ class Installer:
 
         current_version = None
         if os.path.exists(POETRY_LIB):
-            with open(os.path.join(POETRY_LIB, "poetry", "__version__.py")) as f:
+            # explicit utf-8 for python source file
+            with open(
+                os.path.join(POETRY_LIB, "poetry", "__version__.py"), encoding="utf-8"
+            ) as f:
                 version_content = f.read()
 
             current_version_re = re.match(
@@ -561,6 +565,7 @@ class Installer:
             os.mkdir(POETRY_BIN, 0o755)
 
         if WINDOWS:
+            # implicit system encoding for poetry.bat
             with open(os.path.join(POETRY_BIN, "poetry.bat"), "w") as f:
                 f.write(
                     BAT.format(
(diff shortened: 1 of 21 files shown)


vlcinsky:
Note, that this PR includes modifications from PR #1085

vlcinsky:
Note, that my colleague @rhorenov have found, there is probably no need to port this PR into 1.0.0.x as this parts seems to be reimplemented there and is working well.

get-poetry.py
```diff
@@ -58,6 +58,10 @@
 except ImportError:
     winreg = None
 
+try:
+    u = unicode
```
sdispater: Do we really need this alias?

get-poetry.py
```diff
@@ -561,17 +569,21 @@ def make_bin(self):
             os.mkdir(POETRY_BIN, 0o755)
 
         if WINDOWS:
+            # implicit system encoding for poetry.bat
```
sdispater: These comments (and those in the other files) are not really necessary and add too much noise to the codebase

get-poetry.py
vlcinsky: Please, keep it there. Original code was failing on AppVeyor tests, this was the solution to make it working on python 2.7 and 3.x.

get-poetry.py
vlcinsky: I will remove those comments.

sdispater: There are still extraneous comments that are not required

tests/utils/fixtures/setups/ansible/setup.py
```diff
@@ -6,11 +6,12 @@
 import re
 import sys
 import warnings
-
 from collections import defaultdict
 from distutils.command.build_scripts import build_scripts as BuildScripts
 from distutils.command.sdist import sdist as SDist
 
+from ansible.release import __author__, __version__
```
sdispater: Why are these test files modified?

sdispater: They never prevented the tests to pass so I am not sure why they would need to be modified

vlcinsky: Good catch.

It is apparently result of applying isort and black on the code and in this case it is possibly wrong (as the import shall probably happen after path manipulation).

Simply, this change shall be rejected.

vlcinsky:
all comments shall be resolved now (I feel a bit lost in the GitHub review procedure but hope to accept all the change requests).

tests/utils/fixtures/setups/ansible/setup.py
sdispater: Well, sorry if I was not clear but what I meant is that the fixtures `setup.py` files do not need to be modified as far as I know.

vlcinsky: The only modifications of `setup.py` currently are explicitly opening the file (often JSON one) using utf-8. If this is removed and given code runs (for whatever reason) on a system with other than utf-8 default encoding, there is a chance it would fail.

In `.../ansible/setup.py` are three such cases:
- dump JSON file
- load JSON file
- read requirements file

Regarding JSON files the rules are simple: open it in text mode and specify encoding utf-8. Dealing with JSON files differently is error prone to local default encoding.

Regarding reading requirements.txt - it seems to me safer to be explicit and do not allow any space for encoding related issues.

Do you want me to remove any of these three modifications?

sdispater: The fixtures files are taken directly from existing projects to help testing them as-is and, as such, should not be modified.

vlcinsky: Good. All changes to `...../ansible/setup.py` reverted.

sdispater: Thanks! However I also meant the other fixture files that you modified and not only the `ansible` one.

vlcinsky: Thanks for your endless patience.

setup.py for fixtures (3 cases) reverted to their original content.

PR #1106, opened 2019-05-16
Add commands index and doc link to commands section.

cedarkuo:
Hi,

I just started to use poetry recently, and I couldn't find `run` and `shell` commands in READMD.md file. I searched more poetry commands information in https://poetry.eustace.io/docs/cli/, and there are many commands still not sync between README.md and poetry docs website. So I think maybe can add commands index to README.md file for searching more easily.

I would be grateful if there any suggestion about this idea. Thanks.

### DIFF
--- a/README.md
+++ b/README.md
@@ -261,6 +261,23 @@ So, in the end, you will still need to manage a few configuration files to prope
 
 ## Commands
 
+- [Global options](https://poetry.eustace.io/docs/cli/#global-options)
+- [new](https://poetry.eustace.io/docs/cli/#new)
+- [init](https://poetry.eustace.io/docs/cli/#init)
+- [install](https://poetry.eustace.io/docs/cli/#install)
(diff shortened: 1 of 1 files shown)


drunkwcodes:
The command links in the first half of `README.md` is unnecessary.
`README.md` only needs to provide common use to start.

Or the documentation is not readable enough?

Is there any use case in `README.md` need to be renewed?

cedarkuo:
When I use dependency management tool I think it can do these things.
1. create separate local env (project env) from global env
2. install packages to project env
3. run and test python file with project env

I think this version `README.md` can't guide me to step 3 yet, or am I missing somethings in `README.md`?

I thought about to add some example or add `run` section for this purpose, but not sure is this a better change or not. What do you think about those idea?

Thank you for your reply. I am grateful to have opportunity to rethink this change.

PR #1111, opened 2019-05-19
Limit dev dependencies to the environment python version, if specified.

bob-schumaker:
# Pull Request Check List

I have package for both Python 2 and 3, i.e.:
```
[tool.poetry.dependencies]
python = "~2.7 || ^3.6"
```
Since I want to use pylint inside my pyenv environment, I have this in pyproject.toml:
```
pylint = [
    { version = "^1.9.4", python = "~2.7" },
    { version = "^2.3", python = "^3.6" }
]
```
With the latest change to poetry.py where dev-dependencies now have their category set, this caused `poetry update` to fail to resolve a version.

### DIFF
--- a/poetry/poetry.py
+++ b/poetry/poetry.py
@@ -17,7 +17,10 @@ from .repositories import Pool
 from .repositories.auth import Auth
 from .repositories.legacy_repository import LegacyRepository
 from .repositories.pypi_repository import PyPiRepository
+from .semver import parse_single_constraint
+from .semver.version import Version
 from .spdx import license_by_id
+from .utils.env import Env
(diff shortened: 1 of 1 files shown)


sdispater:
There is definitely a bug regarding multiple constraint dev dependencies but this is not the right way to fix this because if you do that your lock file is no longer deterministic and depends on the current Python version and will no longer work for other Python versions.

The reason this bug happens is actually in the `Provider` class which does not retrieve dev dependencies properly.

bob-schumaker:
Ah, that makes sense. Didn't think that through deeply enough (still getting my head around all of the details :-)

PR #1113, opened 2019-05-20
catch and hold ConnectionError until all Repositories are tried

### LINKED ISSUE #1105: catch and hold ConnectionError untill all Repositories are tested
- [x ] I have searched the [issues](https://github.com/sdispater/poetry/issues) of this repo and believe that this is not a duplicate.
- [ x] I have searched the [documentation](https://poetry.eustace.io/docs/) and believe that my question is not covered.

## Feature Request

It would be useful if instead of raising ConnectionError when the repository is offline, poetry would try to get the package from other available reppos in the Pool.

Currently If any of the repositories is attempted and fails with a connection error it will stop the process with a failure. My proposal is to catch the exception(logging a warning in the process) until all of the repo pool is tried, and raise the connection error if it can't find the package anywhere.

As a workaround I believe I can use a different pyproject.toml file for different environments, but this has other problems. 

My specific use case is to use devpy as a pypi proxy on a local network to speedup installs.
witch is useful while developing with docker as you often need to recreate the virtualenv and reinstall everything. Its also a good way to expand on options so that you can integrate with the CI env more easily.

I looked into the code and I think I can work on this issue.
This may relate to #595, but I don't think its the same problem.

marco-silva0000:
# Pull Request Check List

fixes #1105

I've based this on https://github.com/sdispater/poetry/tree/improve-alternative-repositories-support as I believe this fits in well

Needs documentation to clarify the new options available.


marco-silva0000:
Will do this week, need to verify the validity of this change again, and re-test a bit.

abn:
@marco-silva0000 appreciate your contriubtion, if this something you stll want to work on, please rebase on master and re-open.

PR #1119, opened 2019-05-23
Fixed path based dev-dependencies break `install --no-dev` when the d…

jjuris:
Fixed path based dev-dependencies break `install --no-dev` when the directory path does not exist.


poetry/poetry.py
```diff
@@ -147,7 +148,7 @@ def create(cls, cwd):  # type: (Path) -> Poetry
 
                 package.add_dependency(name, constraint)
 
-        if "dev-dependencies" in local_config:
+        if "dev-dependencies" in local_config and not "--no-dev" in sys.argv:
```
jjuris: Hm this change may affect all commands ( did not test it yet )

iScrE4m: Indeed, this doesn't look like the correct place to manage the behavior of the `install` command. What the `--no-dev` option does is it sets `Installer.dev_mode` attribute and you can probably work off of that.

jjuris: Hello,
yes you are right, this is a bad place, its more like patch / workaround. Unfortunately if you look closely into the code, you would find that class Poetry defined in this file is initialised in every single command, even the one which does not to anything effective, therefore it is required to change this control mechanism somehow
JJ.

jjuris:
> You shouldn't modify version in `CHANGELOG.md`, `poetry/__version__.py`, `pyproject.toml`, leave that to the repo owner, who knows which version will be released.

Yes, its habit, this is not going to be merged, ignore it

brycedrennan:
Closing since we were told to ignore.

PR #1127, opened 2019-05-27
Autocomplete instructions for Zsh with Homebrew

jmfederico:
When using Zsh with Homebrew, one can use Homebrew's functions directory instead of creating a new one inside the user's home directory.


sdispater:
Thanks!

Issue #1114, opened 2019-05-20
Malformed METADATA when using optional git dependencies


tyrion:
- [x] I am on the [latest](https://github.com/sdispater/poetry/releases/latest) Poetry version.
- [x] I have searched the [issues](https://github.com/sdispater/poetry/issues) of this repo and believe that this is not a duplicate.
- [x] If an exception occurs when executing a command, I executed it again in debug mode (`-vvv` option).

- **Poetry version**: 0.12.16
- **Link of a [Gist](https://gist.github.com/) with the contents of your pyproject.toml file**: https://gist.github.com/tyrion/fb7b70d437951548f5ee1b5946b9386e

## Issue

Poetry generates a malformed `METADATA` file when there is an optional git dependency specified in `pyproject.toml` (see attached gist).

To reproduce you just need to place the files contained in the attached gist in a directory and run:
```
$ pip install .
[..]
pip._vendor.pkg_resources.RequirementParseError: Parse error at "'extra =='": Expected stringEnd
```

The responsible for this behaviour is the `poetry.masonry.api.prepare_metadata_for_build_wheel` function, which is called by pip.

The generated `METADATA` file looks something like this:
```
Metadata-Version: 2.1
[..]
Requires-Python: >=3.6
Classifier: Programming Language :: Python :: 3
Classifier: Programming Language :: Python :: 3.6
Classifier: Programming Language :: Python :: 3.7
Provides-Extra: stubs
Requires-Dist: numpy-stubs @ git+https://github.com/numpy/numpy-stubs.git@master; extra == "stubs"
```

You can verify that the last line is incorrect using `pkg_resources`:
```
>>> pkg_resources.Requirement.parse('numpy-stubs @ git+https://github.com/numpy/numpy-stubs.git@master; extra == "stubs"')
pkg_resources.RequirementParseError: Invalid requirement, parse error at "'extra =='"
```
Or in the exact same way pip does:
```
>>> from pip._vendor.packaging import requirements as req
>>> req.REQUIREMENT.parseString('numpy-stubs @ git+https://github.com/numpy/numpy-stubs.git@master; extra == "stubs"')
pip._vendor.pyparsing.ParseException: Expected stringEnd (at char 67), (line:1, col:68)
```

To fix the issue it is necessary to add a space before the ';':
```
>>> req.REQUIREMENT.parseString('numpy-stubs @ git+https://github.com/numpy/numpy-stubs.git@master ; extra == "stubs"')
[...]

PR #1129, opened 2019-05-27
Fix METADATA generation for git dependencies with extras

fsal:
Fix issue #1114 
Insert trailing space at the end of the github URL before the semicolon when `extra` field id specified. This trick prevents `pip` parser to include the semicolon (that is a avalid URL character) in the URL. In this way `pip` can install the wheel properly. 
Tested with `poetry 0.12.16`,`poetry 1.0.0a3` and `pip 19.1.1`.

--------------------------------------------
# Pull Request Check List

- [x] Updated **documentation** for changed code: small bug fix, no need for updated docs.


iwoloschin:
Any chance we could get this merged soon?

poetry/packages/vcs_dependency.py
```diff
@@ -71,6 +71,11 @@ def base_pep_508_name(self):  # type: () -> str
 
         requirement += " @ {}+{}@{}".format(self._vcs, self._source, self.reference)
 
+        # add trailing space if 'extra' field is specified between URL
```
sdispater: The `base_pep_508_name` property should return the base name without any extra space since this is the pre-marker part of a dependency.

This code should instead by but inside the `to_pep_508()` method.

fsal:
@sdispater I moved the code to the `to_pep_508()` method in vcs_dependency.py. I also made the changes python2 compliant, but the appveyor job on python2.7 fails because the `typing` module is not present inside the environment.

poetry/packages/vcs_dependency.py
```diff
@@ -78,3 +78,12 @@ def is_vcs(self):  # type: () -> bool
 
     def accepts_prereleases(self):
         return True
+
+    def to_pep_508(self, with_extras=True):  # type: (bool) -> str
+        requirement = super(VCSDependency, self).to_pep_508(with_extras)
+        # add trailing space if 'extra' field is specified between URL
+        # and ';'
+        if self.in_extras:
+            requirement = requirement.replace("; extra", " ; extra")
```
sdispater: That won't work if there are other markers (like `python_version`)

iwoloschin:
Any update on this?  Thanks!

brycedrennan:
@fsal Here is an idea that feels more explicit than finding the first semicolon.   

in `dependency.py` Line 202
```python
        if markers:
            if self.is_vcs():
                # url based requirements should have a space before semicolon
                # https://www.python.org/dev/peps/pep-0508/#grammar
                requirement += " "

            if len(markers) > 1:
                markers = ["({})".format(m) for m in markers]
                requirement += "; {}".format(" and ".join(markers))
            else:
                requirement += "; {}".format(markers[0])

        return requirement
```

Could you implement this and also If add a test case with multiple markers?  If so I think we can get this merged in.

Issue #1109, opened 2019-05-18
Include tests is poetry's sdist


jayvdb:
I am starting the process of packaging poetry for openSUSE  [here](https://build.opensuse.org/package/show/home:jayvdb:coala:test-rig/python-poetry), and notice that the sdist doesnt contain tests, and the GitHub archive doesnt contain `setup.py` .  As a result, there is no single versioned tarball which can be used to build *and* test poetry.  I need to unpack both to have a usable poetry for building a testable rpm.

The easiest way to solve this is to include the tests in the poetry sdist.

eli-schwartz:
I have just run into the same issue on Arch Linux, and this makes it difficult to package any software released using poetry.

eli-schwartz:
@jayvdb By the way, are you aware of the similar packaging issue in #866?

jayvdb:
I wasn't aware of it.  Thanks for the heads-up.  Anyone looking to workaround some of these packaging problems with pyproject.toml projects , I found @dephell to work quite well, and the maintainer is very responsive .

eli-schwartz:
Odd coincidence -- I'm actually attempting to package dephell as of yesterday. :p
Once that is done I am going to check out how feasible it is to add build-depends on python-dephell for poetry projects, and use dephell convert on the github sources (with testsuite) to generate a setup.py.

sdispater:
I plan on making the [`packages`](https://poetry.eustace.io/docs/pyproject/#packages) section of the `pyproject.toml` file support specifying the format for which the package should be included (no format meaning all of them).

```toml
[tool.poetry]
# ...
packages = [
    { include = "my_package" },
    { include = "tests", format = "sdist" },
]
```

PR #1133, opened 2019-05-28
Per format packages support

### LINKED ISSUE #1067: Allow building sdist and wheel with different files
Poetry has an `include` setting in pyproject.toml that allows specifying additional files to be included in the distribution (e.g. tests, docs, config files). This works, but it seems what you specify here will be included in *both* the source distribution and the wheel.

A common practice is to put stuff like tests into the source distribution only, but not into the wheel. You can achieve this with distutils by putting this additional stuff into MANIFEST.in which will be used only for the source distribution. I currently see no way to achieve the same with poetry. Can this be added as a feature? Or am I overlooking something?

sdispater:
## Pull Request Check List

This PR adds support for specifying packages on a per-format basis. This is especially useful for adding a `tests` packages to the `sdist` which is a common case:

```toml
[tool.poetry]
# ...
packages = [
    { include = "my_package" },
    { include = "tests", format = "sdist" },
]
```

This should fix #1067 and #1109

### DIFF
--- a/docs/docs/pyproject.md
+++ b/docs/docs/pyproject.md
@@ -113,6 +113,20 @@ packages = [
 ]
 ```
 
+If you want to restrict a package to a specific [build](#build) format you can specify
+it by using `format`:
+
+```toml
+[tool.poetry]
+# ...
+packages = [
+    { include = "my_package" },
+    { include = "tests", format = "sdist" },
+]
+```
+
+From now on, only the `sdist` build archive will include the `tests` package.
+
 !!!note
 
     Using `packages` disables the package auto-detection feature meaning you have to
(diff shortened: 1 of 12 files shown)


docs/docs/pyproject.md
```diff
@@ -113,6 +113,20 @@ packages = [
...
+```toml
+[tool.poetry]
+# ...
+packages = [
+    { include = "my_package" },
+    { include = "tests", format = "sdist" },
+]
+```
+
+From now on, only the the `sdist` build archive will include the `tests` package.
```
iScrE4m:
```suggestion
From now on, only the `sdist` build archive will include the `tests` package.
```

drunkwcodes:
Can it specify files exclude in certain format?

docs/docs/pyproject.md
sdispater: Thanks!

PR #1156, opened 2019-06-08
Typing syntax fixes

kam1sh:
* Fix some function signatures that were messing with ellipsis
  for example, `fun(a, ...)` is not valid, but `fun(...)` is

* Dict value types - is that legal? ([puzzle/provider.py](https://github.com/sdispater/poetry/pull/1156/files#diff-d6fa69916de34b4fe357e0d6315bc805L221))

* Ignore assigning instance to classes
  (singletons like Infinity in [version/utils.py](https://github.com/sdispater/poetry/pull/1156/files#diff-1b900c3d89e0b0da23cc1bb84f446567L30))

# Pull Request Check List

- [ ] Added **tests** for changed code -- changed only type comments and introduced variable in provider.py; no new tests needed
- [ ] Updated **documentation** for changed code -- well, I can add some notes about mypy in contributing docs if all poetry team uses mypy every day as you use `black`.

I see that Poetry uses a lot of types. Maybe after fixing all typing errors (there is a lot of them right now) we could add mypy in the list of the pre-commit hooks?

P.S. Poetry has two almost identical CONTRIBUTING.md and docs/docs/contributing.md?

### DIFF
--- a/mypy.ini
+++ b/mypy.ini
@@ -1,0 +1,8 @@
+[mypy]
+check_untyped_defs = True
+ignore_errors = False
+ignore_missing_imports = True
+strict_optional = True
+warn_unused_ignores = True
+warn_redundant_casts = True
+warn_unused_configs = True
--- a/poetry/config.py
+++ b/poetry/config.py
@@ -33,7 +33,11 @@ class Config:
     def content(self):
         return self._content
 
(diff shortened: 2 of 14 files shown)


sdispater:
Thanks!

PR #1166, opened 2019-06-13
Add long_description_content_type to sdist setup kwargs

steinnes:
This PR attempts to add the content type of long_description to setup.py kwargs when building sdists.

I noticed that the content type is included in the wheel metadata, but missing from the setup() call in sdist packages.

I did not update any documentation as part of this pull request, because I do not believe this is a change/feature requiring documentation, rather a minor bugfix.

### DIFF
--- a/poetry/masonry/builders/sdist.py
+++ b/poetry/masonry/builders/sdist.py
@@ -176,6 +176,13 @@ class SdistBuilder(Builder):
 
             extra.append("'python_requires': {!r},".format(python_requires))
 
+        if self._meta.description_content_type:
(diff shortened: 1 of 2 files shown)


brycedrennan:
Thanks for contributing!  Would you be able to rebase this into one commit?

steinnes:
Yes certainly :-)

mbargull:
We just encountered this behavior too.
@brycedrennan, is there anything else to be done here before this can be merged?
(cc @tedil)

neersighted:
Superseded by #1994

PR #1180, opened 2019-06-20
Run command initialize() in order to ensure that directory structures are created

hozn:
- [x] (N/A - bug) Updated **documentation** for changed code.

With the update the Cleo 0.7.x (https://github.com/sdispater/cleo/blob/master/CHANGELOG.md) the `initialize()` method of `ConfigCommand` was no longer being called by the superclass.  This PR just calls it explicitly from `handle()` in order to ensure that the directories are created.   Also adds a test to ensure that this functionality works correctly.

### DIFF
--- a/poetry/console/commands/config.py
+++ b/poetry/console/commands/config.py
@@ -78,7 +78,7 @@ To remove a repository (repo is a short alias for repositories):
 
         return unique_config_values
 
-    def initialize(self, i, o):
+    def initialize(self):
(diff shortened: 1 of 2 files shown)


hozn:
I am away from technology for awhile, but can check back on this in a
couple weeks.

sdispater:
The `config` has been refactored (see #1349) and this should no longer be needed.

PR #1183, opened 2019-06-21
Add rich file metadata to poetry.lock

adisbladis:
This is so files can be distinguished based on their meta attributes.
My proposed solution to https://github.com/sdispater/poetry/issues/1172.

`metadata.hashes` is still around though I'd like to start seeing that as deprecated and only for backwards compatibility.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,7 @@
 - The `debug:resolve` command has been renamed to `debug resolve`.
 - The `self:update` command has been renamed to `self update`.
 - Changed the way virtualenvs are stored (names now depend on the project's path).
+- Added files metadata including hash. This deprecates metadata.hashes in `poetry.lock`.
 
 ### Fixed
 
--- a/poetry.lock
+++ b/poetry.lock
@@ -7,6 +7,16 @@ optional = false
 python-versions = "*"
 version = "1.4.3"
 
+[[package.files]]
+hash = "d8b24664561d0d34ddfaec54636d502d7cea6e29c3eaf68f3df6180863e2166e"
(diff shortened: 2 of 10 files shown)


adisbladis:
Here is a demo of how this metadata can be used: https://github.com/adisbladis/poetry2nix/tree/patched-poetry

adisbladis:
@sdispater Could I get a comment on if this is something that you'd like to incorporate? Then I can fix the conflicts.

Caligatio:
Not a maintainer but I think your approach is entirely reasonable. Quickly glanced at your diff and noticed (maybe) some bugs.

adisbladis:
> Not a maintainer but I think your approach is entirely reasonable. Quickly glanced at your diff and noticed (maybe) some bugs.

What bugs did you notice?

Caligatio:
There is a slight logic bug (I think) with your new `files` property for a package.  In [poetry/installation/pip_installer.py](https://github.com/adisbladis/poetry/blob/file-metadata/poetry/installation/pip_installer.py#L115), you assume there is a `hash` key if there is a `files` property.  However, there is an edge case in [poetry/repositories/legacy_repository.py](https://github.com/adisbladis/poetry/blob/file-metadata/poetry/repositories/legacy_repository.py#L383) in which a hash is not written into the `file` dict.  I believe this would cause a KeyError to be raised on install.

adisbladis:
@sdispater Is this something that could possibly be merged before 1.0? Since I'm making changes to the lock format itself this _needs_ to go in before 1.0.

adisbladis:
Since https://github.com/sdispater/poetry/pull/1460 this can be closed.

PR #1185, opened 2019-06-22
Add poetry-version attribute to poetry.lock

adisbladis:
I want to know what version of poetry created a lock file. 

This also changes some of the testing behaviour to make field data easier to work with. 

### DIFF
--- a/tests/helpers.py
+++ b/tests/helpers.py
@@ -3,6 +3,9 @@ from poetry.packages import Package
 
 from poetry.utils._compat import Path
 
+from functools import wraps
+import tomlkit
+
 
 FIXTURE_PATH = Path(__file__).parent / "fixtures"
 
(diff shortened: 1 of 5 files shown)


tests/helpers.py
```diff
@@ -28,3 +31,63 @@ def fixture(path=None):
...
+        for idx, vals in enumerate(zip(a, b)):
+            p = make_path(idx)
+            if ".".join(p) not in ignore_paths:
+                assert_deepequals(*vals, _path=p, ignore_paths=ignore_paths)
+
+    elif isinstance(a, dict):
+        if not isinstance(b, dict):
+            raise err
+
+        for key in set(list(a.keys()) + list(b.keys())):
```
agoose77: `keys()` views are already set-like, so you can reduce this to
```
for key in a.keys() + b.keys():
```
:)

adisbladis: No, you cannot:
``` python
>>> d = {}
>>> d.keys() + d.keys()
  File "<stdin>", line 1, in <module>
TypeError: unsupported operand type(s) for +: 'dict_keys' and 'dict_keys'
>>>
```

agoose77:
Apologies, I meant to use the set union | instead of addition

adisbladis:
A few months later and I can't even remember why I wanted this.. The testing changes may be worthwhile to incorporate anyway, but I'm not gonna spend time pushing this PR through.

bibz:
I have a use case for this: knowing what flavour of the data structure to expect. The idea is that poetry can warn/error directly if it encounters an unknown flavour. (I.e. poetry should be updated.)

For instance, poetry 1.0.0b2 changed the `poetry.lock` flavour by replacing `[metadata.hashes]` with `[metadata.files]`. I did trip on that due to having different version of poetry available on my machine. I then updated the `poetry.lock` files but not everybody was aware they had to update their toolchain.

I understand this is an operational issue we have to fix in our team. But I believe the cost of including a guard in poetry is worth it. The current implementation is backwards compatible and handle the older schema properly. But it cannot detect when a new schema is in place, and thus will fail when facing a too-recent `poetry.lock`.

Issue #1163, opened 2019-06-12
'Pool' object has no attribute 'default'


m1nde:
- [x] I am on the [latest](https://github.com/sdispater/poetry/releases/latest) Poetry version: 1.0.0a4
- [x] I have searched the [issues](https://github.com/sdispater/poetry/issues) of this repo and believe that this is not a duplicate.
- [x] If an exception occurs when executing a command, I executed it again in debug mode (`-vvv` option).

- **OS version and name**: macOS 10.14
- **Poetry version**: 1.0.0a4

Using poetry 1.0.0a3 and 1.0.0a4, adding a package from a private repository gives the following error:
```
  - Installing my-private-package (0.0.1)

[AttributeError]
'Pool' object has no attribute 'default'
````

Full stack trace:
```
Traceback (most recent call last):
  File "/Users/0x72/.poetry/lib/poetry/_vendor/py3.7/clikit/console_application.py", line 131, in run
    status_code = command.handle(parsed_args, io)
  File "/Users/0x72/.poetry/lib/poetry/_vendor/py3.7/clikit/api/command/command.py", line 112, in handle
    status_code = self._do_handle(args, io)
  File "/Users/0x72/.poetry/lib/poetry/_vendor/py3.7/clikit/api/command/command.py", line 160, in _do_handle
    return getattr(handler, handler_method)(args, io, self)
  File "/Users/0x72/.poetry/lib/poetry/_vendor/py3.7/cleo/commands/command.py", line 92, in wrap_handle
    return self.handle()
  File "/Users/0x72/.poetry/lib/poetry/console/commands/add.py", line 158, in handle
    status = installer.run()
  File "/Users/0x72/.poetry/lib/poetry/installation/installer.py", line 75, in run
    self._do_install(local_repo)
  File "/Users/0x72/.poetry/lib/poetry/installation/installer.py", line 292, in _do_install
    self._execute(op)
  File "/Users/0x72/.poetry/lib/poetry/installation/installer.py", line 308, in _execute
    getattr(self, '_execute_{}'.format(method))(operation)
  File "/Users/0x72/.poetry/lib/poetry/installation/installer.py", line 333, in _execute_install
    self._installer.install(operation.package)
  File "/Users/0x72/.poetry/lib/poetry/installation/pip_installer.py", line 58, in install
    if repository.name != self._pool.default.name:
```

[...]

PR #1193, opened 2019-06-25
Add `default` property for Pool

pbzweihander:
# Pull Request Check List

---

I have some issues with installing packages from a private repository. (#1163)
In `PipInstaller`, the installer checks `self._pool.default` repository to determine it is default or not. But there is no `default` attribute in `Pool`.
I think it was a mistake. So I added a property `default` to solve this problem.

### DIFF
--- a/poetry/repositories/pool.py
+++ b/poetry/repositories/pool.py
@@ -27,6 +27,12 @@ class Pool(BaseRepository):
         super(Pool, self).__init__()
 
     @property
+    def default(self):  # type: () -> Repository
+        if not self.has_default():
+            raise ValueError("Default repository does not exist.")
+        return self._repositories[0]
+
+    @property
     def repositories(self):  # type: () -> List[Repository]
         return self._repositories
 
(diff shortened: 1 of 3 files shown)


Caligatio:
As commented in #1163 by @droserasprout, I believe you mis-implemented the getter "default"  It's supposed to return a simple boolean whereas your implementation is returning a Repository object.

pbzweihander:
@Caligatio 
But `PipInstaller` already using a unimplemented attribute named `default`. And clearly `PipInstaller` expects it as a `Repository` object.

https://github.com/sdispater/poetry/blob/434c50e5c6fa2472db6d40223053bcbb981ca29e/poetry/installation/pip_installer.py#L58

I know the `_default` field is a simple boolean value and `PipInstaller` should use `_repositories` or something to get the default repository. But I think this bug is very critial (I can't use repositories at all!) so I made this PR as some kind of quick-patch that modifying the code smallest as possible.

I hope the better idea come out and fix this simple bug, so I can use poetry normally.

Caligatio:
Yep, you're definitely right!  I suppose it also doesn't make a lot of sense to have a getter for a property when there is already a method to get the same thing.

@stephsamson Any chance this can be prioritized?  This is also a blocker for a friend of mine (he's having to abandon using Poetry).

sdispater:
Commit 6f4aa21 fixes this. I added you as a co-author.

Thanks for for making this PR

PR #1195, opened 2019-06-26
Set path in shell

maw:
`poetry shell` never worked properly for me. Other people reported this in https://github.com/sdispater/poetry/issues/497 and https://github.com/sdispater/poetry/issues/571 but the solutions and workarounds mentioned never worked for me.

The reason is that early on in my shell's config, I reset `PATH` to a known state and go on to build it up from there. So, even though poetry sets it properly, my config undoes it.

My solution: test `POETRY_ACTIVE` and then append `POETRY_PYTHON_PATH` to `PATH`.

Notes:
- I don't know if `POETRY_PYTHON_PATH` is ideal, since `PYTHONPATH` already exists and is used differently. Maybe `POETRY_PYTHON_BIN` or `POETRY_PYTHON_BIN_PATH` or maybe something else would be better.
- I don't have any tests for this, sorry.

Checklist

### DIFF
--- a/docs/docs/cli.md
+++ b/docs/docs/cli.md
@@ -348,6 +348,9 @@ The `shell` command spawns a shell,
 according to the `$SHELL` environment variable,
 within the virtual environment.
 If one doesn't exist yet, it will be created.
+The `$POETRY_ACTIVE` environment variable is set to `1`
+and the `$POETRY_PYTHON_PATH` is set to the
+`bin/` directory in the virtual environment.
 
 ```bash
 poetry shell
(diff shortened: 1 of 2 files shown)


sdispater:
@maw I think a lot of problems with the `shell` command have been fixed in #1395. Could you test if your specific issue is fixed with the `1.0.0b2` release?

maw:
@sdispater I'll try to give it a try during the week or, more likely, this weekend. I'll let you know.

If it doesn't do the trick for me, I'll have a go at writing some tests for @stephsamson

maw:
@sdispater @stephsamson I've been using recent builds (I'm on 1.0.0b5 and was using 1.0.0b4 previously) for the past two weeks and the problem as I experienced it doesn't occur anymore.  Thanks.

As an aside, I'd like to mention that Poetry has really come a long way in the past year. Python desperately needed a tool like this, managed by people with the with the intention of seeing the project through to a state where it could be used for serious work. I think you're just about there. Good work!

PR #1201, opened 2019-06-29
adds support for equivalent of `pip -t` for `poetry install`

### LINKED ISSUE #1174: pip install -t equivalent
When working with serverless, (but surely other contexts too), we run a script essentially like so:

```
#!/usr/bin/env bash
set -e

export PIP_EXTRA_INDEX_URL=https://pypi.fury.io/${GEMFURY_TOKEN}/foobar
poetry export --format requirements.txt --without-hashes
poetry run pip install -r requirements.txt . --target package
rm requirements.txt

cd package
find . -name "*.pyc" -delete
zip -9 -r ../package.zip .
cd ..
rm -fr package

npx sls deploy
```

It would be great if poetry had the ability to output packages itself so that we could e.g.:

```
poetry install --target package
zip -9 -r package.zip package
npx sls deploy
rm package.zip
```

I couldn't find a pre-existing issue for this but also finding it surprising this issue wouldn't have come up yet. So maybe my title phrasing is off or this is already possible?

brno32:
Resolves #1174 

### DIFF
--- a/docs/docs/cli.md
+++ b/docs/docs/cli.md
@@ -124,6 +124,12 @@ poetry install --extras "mysql pgsql"
 poetry install -E mysql -E pgsql
 ```
 
+It is possible to mimic the `-t` option supported by pip by passing the `--target` option.
+
+```bash
+poetry install --target .
+```
+
 By default `poetry` will install your project's package everytime you run `install`:
 
 ```bash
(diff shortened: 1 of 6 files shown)


peterdeme:
Can this merged?

sdispater:
Thanks for this PR.

However, I don't want to change the behavior of the `install` command, even with an option.

That being said, this will likely be implemented in another form in a future version of Poetry, so stay tuned :-)

lrepolho:
Any reason this wasn't merged or an alternative provided by the project so far? This would be really useful for those working with AWS Lambda environment.

Issue #1189, opened 2019-06-23
OS Specific dependencies


deepio:
- [x] I have searched the [issues](https://github.com/sdispater/poetry/issues) of this repo and believe that this is not a duplicate.
- [x] I have searched the [documentation](https://poetry.eustace.io/docs/) and believe that my question is not covered.

## Feature Request
Sometimes a projects dependencies are different depending on the OS you are installing them on. I would like to be able to define different dependencies based on the OS. 

My suggestion would be to keep `[tool.poetry.dependencies]` for dependencies that are the same for all OS's, and add `[tool.poetry.dependencies.macOS]`, `[tool.poetry.dependencies.win]` or `[tool.poetry.dependencies.linux]` for example when there are some that are OS specific.

Currently I can get away with installing for mac using extra, because the regular install luckily  works with windows. But to support linux, I need specific dependencies that unfortunately only work on linux and would make poetry install fail on other OS's.

deepio:
I decided to cherry pick extras for the os.

Issue #738, opened 2018-12-13
No documentation about markers (such as platform)


thernstig:
- [x ] I have searched the [issues](https://github.com/sdispater/poetry/issues) of this repo and believe that this is not a duplicate.

## Issue
There seems to be no mention in the documentation of the usage and supported markers, but they do seem to be supported according to https://github.com/sdispater/poetry/issues/21

jimbo1qaz:
What `platform` do I use for Windows 10 x64?

I'm using Poetry 0.12.10.

```sh
poetry remove pywin32-ctypes
pip uninstall -y pywin32-ctypes
pyinstaller
poetry add --platform=windows pywin32-ctypes
pyinstaller
```

And I"ve tried windows, Windows, nt, and pywin32-ctypes never gets installed.

carlosperate:
I was looking into this recently and cound't find the documentation either, are platform markers officially supported?

deepio:
Could this be documented please?
https://github.com/sdispater/poetry/issues/1189

thernstig:
/ unstale 📦

eterna2:
can I check if 
```
extras_require = \
{':sys_platform == "windows"': ['pywin32-ctypes>=0.2.0,<0.3.0']}
```
is a valid setup.py format? Because I get format error (extras_require must be a dict of xyz) when I tried to do `python setup.py install`

zillionare:
where can I find the documentation for specifying sys_platform. According to the thread, it's added to the document, however I can't find it in the document.

sinoroc:
@zillionare It's in the "_[Dependency specification -- Using environment markers](https://python-poetry.org/docs/dependency-specification/#using-environment-markers)_" section.

dclong:
Is it possible to restrict the whole package to be on, say, Linux only? Do I have to use marker for every package or is there a simple way of specifying OS restrictions?

sinoroc:
> Is it possible to restrict the whole package to be on, say, Linux only? Do I have to use marker for every package or is there a simple way of specifying OS restrictions?

@dclong Use trove classifiers. They are an informal (non-enforced) restriction:

```
[tool.poetry]
# ...
classifiers = [
    "Operating System :: POSIX :: Linux",
]
```

* https://github.com/python-poetry/poetry/issues/3356#issuecomment-726163405
* https://pypi.org/classifiers/
* https://python-poetry.org/docs/pyproject/#classifiers

dclong:
[...]

PR #1207, opened 2019-07-02
Include platform option in documentation: #1189 #738

deepio:
# Pull Request Check List

~~- [ ] Added **tests** for changed code.~~

### If it's a bug fix or only a documentation update, it should be based on the `master` branch.

#1189 #738 
It is unclear that dependencies specific to a platform can be handled by poetry.

### DIFF
--- a/README.md
+++ b/README.md
@@ -381,9 +381,15 @@ poetry add requests pendulum
 
 #### Options
 
-* `--D|dev`: Add package as development dependency.
+* `--dev (-D)`: Add package as development dependency.
+* `--git`: The url of the Git repository.
+* `--path`: The path to a dependency.
+* `--extras (-E)`: Extras to activate for the dependency.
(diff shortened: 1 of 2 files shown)


jimbo1qaz:
This is wrong.
- it must match `sys.platform`. You should mention this in the documentation.
- sys.platform is "win32" not "windows".

deepio:
> This is wrong, unless poetry has changed since the last time i checked, or unless i was wrong in #738.
> 
> * it must match `sys.platform`. You should mention this in the documentation.
> * sys.platform is "win32" not "windows".

You're right, fixed.

deepio:
/ unstale

finswimmer:
Hello,

these changes are now obsolete, because content is removed from README and is only available in the docs.

Thanks a lot for your suggestions!

fin swimmer

PR #1213, opened 2019-07-09
fix python 2 open file encoding in pip installer

duilio:
Fix Python 2 compatibility for pip installer by using `io.open` instead of `open`.

### DIFF
--- a/poetry/installation/pip_installer.py
+++ b/poetry/installation/pip_installer.py
@@ -1,3 +1,4 @@
+import io
 import os
 import shutil
 import tempfile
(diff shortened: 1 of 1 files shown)


poetry/installation/pip_installer.py
```diff
@@ -193,7 +194,7 @@ def install_directory(self, package):
             # We also need it for non-PEP-517 packages
             builder = SdistBuilder(Poetry.create(pyproject.parent), NullEnv(), NullIO())
 
-            with open(setup, "w", encoding="utf-8") as f:
```
duilio: `encoding` is not a valid param in python 2 `open`. Both Python 2 and Python 3 have an `open` function in the `io` module that is actually the same as the builtin Python 3 `open`.

sdispater:
Thanks. However, now that `develop` has been merged into `master`, this PR is no longer relevant.

PR #1225, opened 2019-07-14
Accepts licenses by their long name, e.g. Apache License 2.0

### LINKED ISSUE #310: `poetry init` should accept the long license name in addition to the abbreviation

## Issue

A minor quality-of-life improvement to `poetry init` would be to accept the long name for a license, e.g. "Apache License 2.0", in addition to the abbreviation, "Apache-2.0". It looks like Poetry [already tracks both variations](https://github.com/sdispater/poetry/blob/af8f86cda703144e9261b700d3165433aa2d3603/poetry/spdx/data/licenses.json#L147-L148).

```
$ poetry init

This command will guide you through creating your poetry.toml config.

Package name [flintrock]:
<snipped>
License []: Apache License 2.0
 Invalid license id: Apache License 2.0
License []: Apache-2.0
<success>
```

The reason I tried the longer version of the license name first, and the reason I think it would make sense for Poetry to accept that variation, is because that's what I had in my call to `setuptools.setup(license=...)` in my `setup.py`. I imagine other users might try the same and try to work through `poetry init` by copying things out of their `setup.py`.

sgarcialaguna:
Fixes #310 

### DIFF
--- a/poetry/spdx/__init__.py
+++ b/poetry/spdx/__init__.py
@@ -6,25 +6,27 @@ from io import open
 from .license import License
 from .updater import Updater
 
-_licenses = None
+_licensesByShortIdentifier = None
+_licensesByLongName = None
 
 
 def license_by_id(identifier):
-    if _licenses is None:
+    if _licensesByShortIdentifier is None or _licensesByLongName is None:
         load_licenses()
 
     id = identifier.lower()
 
-    if id not in _licenses:
+    if id not in _licensesByShortIdentifier and id not in _licensesByLongName:
         raise ValueError("Invalid license id: {}".format(identifier))
 
-    return _licenses[id]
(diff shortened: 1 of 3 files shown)


poetry/spdx/__init__.py
```diff
@@ -6,33 +6,40 @@
 from .license import License
 from .updater import Updater
 
-_licenses = None
+_licensesByShortIdentifier = None
+_licensesByLongName = None
```
flaviut: Why separate them into two variables? why not store both the long and the short license names in the same dict?

sgarcialaguna: Because it's easier to look them up that way.

brycedrennan:
Thanks for taking time to make a contribution.  This ended up getting addressed in a [different commit](https://github.com/sdispater/poetry/commit/53c22155e59d9efa4979311aa66f40cf39270ced).

PR #1230, opened 2019-07-15
Fix empty marker issues

### LINKED ISSUE #978: Poetry fails on keyboard add/install requirements
- [ x] I am on the [latest](https://github.com/sdispater/poetry/releases/latest) Poetry version.
- [ x] I have searched the [issues](https://github.com/sdispater/poetry/issues) of this repo and believe that this is not a duplicate.
- [x ] If an exception occurs when executing a command, I executed it again in debug mode (`-vvv` option).

- **OS version and name**: Windows 10, Python 3.7.2
- **Poetry version**: 

## Issue

Python 3.7.2
PIP 19.0.3
Setuptools 40.8.0
Poetry 1.0.0a2
######################################
Using version ^0.13.3 for keyboard

Updating dependencies
Resolving dependencies...

[InvalidRequirement]
Invalid requirement, parse error at "'; <empty'"

Nitori-:
This PR addresses a family of related problems we were seeing when depending on particular packages.

Based on what I saw in the codebase, there is a concept of an "empty" marker. However there were a few places in the code that didn't handle these empty markers properly.

Here's an example `pyproject.toml` where we were hitting the issue:

```toml
[tool.poetry]
name = "example_project"
version = "1.0.0"
description = ""
authors = [""]

[tool.poetry.dependencies]
python = ">=2.7, != 3.0.*, != 3.1.*, != 3.2.*, != 3.3.*"

pytest = "^3.0.6"
pylama = "^7.6"
pylint = { version = "^2.2", python = ">= 3.4" }
pytest-cov = "^2.6"
black = { version = "18.9b0", python = ">= 3.6" }
sphinx = { version = "^2.0", python = ">= 3.5" }

[build-system]
requires = ["poetry>=0.12"]
```

Doing `poetry install` once to generate the `poetry.lock` and then running `poetry install` again will reproduce the issue:

```
$ poetry install

[InvalidRequirement]
Invalid requirement, parse error at "'; <empty'"

install [--no-dev] [--dry-run] [-E|--extras EXTRAS] [--develop DEVELOP]
```
Also I believe this fixes #978 which seems to be hitting a similar problem.

# Pull Request Check List

### DIFF
--- a/poetry/semver/__init__.py
+++ b/poetry/semver/__init__.py
@@ -161,4 +161,7 @@ def parse_single_constraint(constraint):  # type: (str) -> VersionConstraint
         else:
             return version
 
+    if constraint == "<empty>":
+        return VersionRange()
+
     raise ValueError("Could not parse version constraint: {}".format(constraint))
--- a/poetry/packages/__init__.py
+++ b/poetry/packages/__init__.py
@@ -102,7 +102,10 @@ def dependency_from_pep_508(name):
             for op, version in or_:
                 # Expand python version
                 if op == "==":
-                    version = "~" + version
+                    if "*" in version:
+                        version = "==" + version
+                    else:
+                        version = "~" + version
                     op = ""
                 elif op == "!=":
                     version += ".*"
--- a/poetry/packages/package.py
+++ b/poetry/packages/package.py
@@ -340,7 +340,7 @@ class Package(object):
 
         name = "{} (=={})".format(self._name, self._version)
 
-        if not self.marker.is_any():
+        if not self.marker.is_any() and not self.marker.is_empty():
             name += " ; {}".format(str(self.marker))
 
         return dependency_from_pep_508(name)
--- a/tests/packages/test_main.py
+++ b/tests/packages/test_main.py
@@ -169,3 +169,10 @@ def test_dependency_from_pep_508_with_not_in_op_marker():
     assert (
         str(dep.marker) == 'python_version not in "3.0,3.1,3.2" and extra == "export"'
     )
+
+
+def test_dependency_from_pep_508_with_python_version_wildcard():
+    name = 'requests; python_version == "2.7.*"'
+    dep = dependency_from_pep_508(name)
+    assert dep.name == "requests"
+    assert str(dep.python_versions) == "==2.7.*"
--- a/tests/packages/test_package.py
+++ b/tests/packages/test_package.py
@@ -2,6 +2,7 @@
 from __future__ import unicode_literals
 
 from poetry.packages import Package
+from poetry.version.markers import EmptyMarker
 
 
 def test_package_authors():
@@ -14,3 +15,11 @@ def test_package_authors():
     package.authors.insert(0, "John Doe")
     assert package.author_name == "John Doe"
     assert package.author_email is None
+
+
+def test_package_empty_marker():
+    package = Package("foo", "0.1.0")
+    package.marker = EmptyMarker()
+
+    dep = package.to_dependency()
+    assert dep.python_versions == "*"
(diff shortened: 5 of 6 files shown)


poetry/semver/__init__.py
```diff
@@ -161,4 +161,7 @@ def parse_single_constraint(constraint):  # type: (str) -> VersionConstraint
         else:
             return version
 
+    if constraint == "<empty>":
+        return VersionRange()
```
sdispater: An empty constraint is not the same thing as a wildcard (`*`) constraint. An empty constraint is an unsolvable constraint which always evaluates to false. This can happen when intersecting to incompatible constraints.

This is not the proper way to fix this issue

Nitori-: Fair enough, but then I'm not sure how to really fix the problem then.

kevinkjt2000: A comment by @bjoernpollex https://github.com/sdispater/poetry/issues/804#issuecomment-512263672 asks if having `<empty>` be inserted is unintentional.  If `<empty>` being in the lock file is indeed unintentional, it would follow that `<empty>` does not need to be handled later (as it would not be there).

bjoernpollex: The `<empty>` is not even in the lock file. It's being generated by poetry internally, and then when it tries to parse it again, it fails. I'll try to come up with a minimal reproduction. In my case it happens, as @sdispater described, when intersecting different constraints. However, the constraints being intersected are `extras`, and I'm not sure what the proper semantics for those are.

bjoernpollex:
@Nitori- I've been looking into this some more, and from what I understand, what happens here is the following. When `poetry` resolves the dependencies [here][1], it will find that `pylint` depends on `isort`, which depends on `backports.functools-lru-cache; python<3.2`. After resolution, `poetry` [builds a dependency graph][2] (I think it's actually building a tree here, but might be wrong). While building this graph, the marker for each package in each branch it occurs is derived by [intersecting the declared markers with the markers of its parent in the dependency graph][3] (this way, marker get propagated down to transitive dependencies). This is where the `<empty>` marker can get generated. In this case, the marker for `pylint` is `python>=3.4` (as declared in `pyproject.toml`), but the marker for `backports.functools-lru-cache` is `python<3.2` - the intersection of the two is empty, because the `backports.functools-lru-cache` is not needed (at least in that branch of the dependency graph)!

In the next step derives the appropriate markers and installation order for each package by [flattening that graph][4]. If a package occurs in multiple branches of the graph, the markers get unioned, and thus a previously `<empty>` marker can turn into a valid one. In your example, this doesn't happen though, and so the `<empty>` marker ends up in `poetry.lock`.

@sdispater I'm wondering, if the above explanation is correct, should packages with an `<empty>` marker just be filtered out at this point? Or should this issue be fixed upstream, so that the solver doesn't even include the package in the solution?

[1]: https://github.com/sdispater/poetry/blob/4dbcb3dba0e987694e30644f2539ad3da7fde17c/poetry/puzzle/solver.py#L170-L172
[2]: https://github.com/sdispater/poetry/blob/4dbcb3dba0e987694e30644f2539ad3da7fde17c/poetry/puzzle/solver.py#L182
[3]: https://github.com/sdispater/poetry/blob/4dbcb3dba0e987694e30644f2539ad3da7fde17c/poetry/puzzle/solver.py#L211-L214
[4]: https://github.com/sdispater/poetry/blob/4dbcb3dba0e987694e30644f2539ad3da7fde17c/poetry/puzzle/solver.py#L185-L197

bjoernpollex:
@sdispater @stephsamson @brycedrennan It would be great to get some feedback on my comment above - I'm happy to work on that, but would need some guidance to avoid wasting time.

Nitori-:
OK so I realized that this fix might have been simpler than I understood it to be. If I just parse `<empty>` as an `EmptyConstraint()` the bug seems to go away.

PR #1236, opened 2019-07-19
Use `os.execvp` for `poetry run`

### LINKED ISSUE #993: `poetry run` should use execve syscall

- **OS version and name**: any
- **Poetry version**: 11.5

## Issue

When running a program through `poetry run`, [`subprocess.call`](https://github.com/sdispater/poetry/blob/0.11.5/poetry/utils/venv.py#L251) is used to execute the command. This has the negative side effect of causing poetry to swallow all signals, making it difficult to wrap commands that depend on signals to be handled properly.

I think the behavior that `poetry run` should have is that the command provided is `exec`'d using the [`exceve`](https://docs.python.org/2/library/os.html#os.execve) syscall, which will replace the process and allow signals to be handled properly. This is similar to the behavior of similar utilities, like [`bundle exec`](https://github.com/bundler/bundler/blob/d368aa9e6257a27fab31efe0396069c8977cdc99/lib/bundler/cli/exec.rb#L53) in ruby.

Poetry should play nice with other unix utilities.

## Steps to reproduce

```console
$ cd $(mktemp -d)
$ poetry init
$ go get -u github.com/ejholmes/sigp
$ sigp sleep 60 &
$ kill -TERM $!
signal 15 received (terminated)
signal 20 received (child exited)
$ poetry run sigp sleep 60 &
$ kill -TERM $! # notice that sigp doesn't print anything, because it didn't get the TERM signal
$ ps -A | grep sigp # zombie process now
18862 ttys013    0:00.01 sigp sleep 60
```

pbzweihander:
------

This resolves #993.

`poetry run` uses `subprocess.call` to run the other program, making poetry spawn a new child process.
The problem is when `SIGTERM` is handed over; poetry does not pass it to the child process, instead terminates itself and makes zombie processes.
So I changed `poetry run` to use `os.execvp`, so poetry will replace itself with a new process and properly handle signals.

## Test

I made this small script in `poetry/tester.py`:

```python
import signal
import time


def handler(signum, frame):
    print(f"signal {signum} received")


def main():
    print("start")
    signal.signal(signal.SIGTERM, handler)
    time.sleep(10)
    print("end")


if __name__ == "__main__":
    main()
```

Then ran `poetry run python poetry/tester.py` and tried to kill it.

### Before

```bash
$ poetry run python poetry/tester.py &
[1] 23987
start
$ ps -ax | grep tester
24270 pts/11   S      0:00 python /home/.../poetry run python poetry/tester.py
24350 pts/11   S      0:00 /home/.../python poetry/tester.py
24359 pts/11   S+     0:00 grep --color=auto tester
$ kill -TERM $!
[1]+  Terminated              poetry run python poetry/tester.py
$ ps -ax | grep tester
24350 pts/11   S      0:00 /home/.../python poetry/tester.py
24446 pts/11   S+     0:00 grep --color=auto tester
$ 
end
```

### Fixed

```bash
$ python -m poetry run python poetry/tester.py &
[1] 25282
start
$ ps -ax | grep tester
25282 pts/11   S      0:00 /home/.../python poetry/tester.py
25364 pts/11   S+     0:00 grep --color=auto tester
$ kill -TERM $!
signal 15 received
$ 
end
[1]+  Done                    python -m poetry run python poetry/tester.py
```

## Test - scripts feature

I added the following lines to `pyprojects.toml`.

```toml
...
[tools.poetry.scripts]
...
tester = "poetry.tester:main"
```

Then ran `poetry run tester` and tried to kill it.

### Before

```bash
$ poetry run tester &
[1] 27077
start
$ ps -ax | grep tester
27077 pts/11   S      0:00 python /home/.../poetry run tester
27173 pts/11   S      0:00 /home/.../python -c import sys; from importlib import import_module; sys.argv = ['tester']; import_module('poetry.tester').main()
27189 pts/11   S+     0:00 grep --color=auto tester
$ kill -TERM $!
[1]+  Terminated              poetry run tester
$ ps -ax | grep tester
27173 pts/11   S      0:00 /home/.../python -c import sys; from importlib import import_module; sys.argv = ['tester']; import_module('poetry.tester').main()
27266 pts/11   S+     0:00 grep --color=auto tester
$ 
end
```

### Fixed

```bash
$ python -m poetry run tester &
[1] 27996
start
$ ps -ax | grep tester
27996 pts/11   S      0:00 /home/.../python -c import sys; from importlib import import_module; sys.argv = ['tester']; import_module('poetry.tester').main()
28082 pts/11   S+     0:00 grep --color=auto tester
$ kill -TERM $!
signal 15 received
$ 
end
[1]+  Done                    python -m poetry run tester
```

### DIFF
--- a/poetry/console/commands/run.py
+++ b/poetry/console/commands/run.py
@@ -40,15 +40,13 @@ class RunCommand(EnvCommand):
         cmd = ["python", "-c"]
 
         cmd += [
-            '"import sys; '
+            "import sys; "
             "from importlib import import_module; "
             "sys.argv = {!r}; {}"
-            "import_module('{}').{}()\"".format(
-                args, src_in_sys_path, module, callable_
-            )
+            "import_module('{}').{}()".format(args, src_in_sys_path, module, callable_)
         ]
 
-        return self.env.run(*cmd, shell=True, call=True)
+        return self.env.execute(*cmd)
 
     @property
     def _module(self):
--- a/poetry/utils/env.py
+++ b/poetry/utils/env.py
@@ -712,7 +712,12 @@ class Env(object):
     def execute(self, bin, *args, **kwargs):
         bin = self._bin(bin)
 
-        return os.execvp(bin, [bin] + list(args), **kwargs)
+        if not self._is_windows:
+            return os.execvp(bin, [bin] + list(args))
+        else:
+            exe = subprocess.Popen([bin] + list(args), **kwargs)
+            exe.communicate()
+            return exe.returncode
 
     def is_venv(self):  # type: () -> bool
         raise NotImplementedError()
--- a/poetry/utils/env.py
+++ b/poetry/utils/env.py
@@ -712,7 +712,7 @@ class Env(object):
     def execute(self, bin, *args, **kwargs):
         bin = self._bin(bin)
 
-        return subprocess.call([bin] + list(args), **kwargs)
+        return os.execvp(bin, [bin] + list(args), **kwargs)
 
     def is_venv(self):  # type: () -> bool
         raise NotImplementedError()
--- a/poetry/utils/env.py
+++ b/poetry/utils/env.py
@@ -713,7 +713,11 @@ class Env(object):
         bin = self._bin(bin)
 
         if not self._is_windows:
-            return os.execvp(bin, [bin] + list(args))
+            args = [bin] + list(args)
+            if "env" in kwargs:
+                return os.execvpe(bin, args, kwargs["env"])
+            else:
+                return os.execvp(bin, args)
         else:
             exe = subprocess.Popen([bin] + list(args), **kwargs)
             exe.communicate()


bertjwregeer:
:+1: on this however using `os.execvp` does not work on Windows. Please take a look at and steal (my permission is hereby given to relicense it under whatever Poetry is licensed as) my code from here: https://github.com/bertjwregeer/vrun/tree/master/src/vrun which supports both Unix and Windows.

bertjwregeer:
specifically you want this file: https://github.com/bertjwregeer/vrun/blob/master/src/vrun/oscompat.py

simnalamburt:
The document says that `os.execvp` is available in Windows. May I ask what exception is thrown when we try to use `os.execvp` in Windows?

> 스크린샷 2019-07-24 오후 2 57 00

https://docs.python.org/3/library/os.html#os.execvp

pbzweihander:
I found in Windows, `exec` does not actually replace current process and instead creates a new one and exits the current one.

- Reference:
  - https://stackoverflow.com/questions/7004687/os-exec-on-windows
  - https://bugs.python.org/issue9148
  - https://docs.microsoft.com/en-us/previous-versions/visualstudio/visual-studio-2010/431x4c1w(v=vs.100)

So I think poetry needs to create a new process and wait until the new one exits. (https://github.com/bertjwregeer/vrun/blob/a726be8106a00621361f60079ac403e5bc16864a/src/vrun/oscompat.py#L34)

bertjwregeer:
If you try to use the exec calls on Windows you get into a really weird state where the process you are running can't be controlled from the cmd.exe window it is spawned from.

You need to do the whole popen dance so that you can send key strokes from the foreground to the process that was spawned on Windows.

bertjwregeer:
I see that @pbzweihander got there before me! :-D

Please steal my code, I didn't come to that conclusion the easy way!

pbzweihander:
I have updated the code but didn't test yet. (I don't have Windows machine) And I also tried to handle `shell` and `env` keyword arguments, but I'm not sure it works properly... :thinking:

simnalamburt:
Now I think this PR is ready to be merged since I see no more blockers.

brycedrennan:
Thanks for your contribution.  I'm not sure when someone will have a chance to dig into this.  

I imagine its pretty hard to add an automated test for this?

bertjwregeer:
What kind of tests are you looking for?

maksbotan:
@sdispater so, are there plans to merge this one? :)

sdispater:
Thanks a lot for this PR!

rpdelaney:
Does this have any effect on performance?

simnalamburt:
It'll boost performance since `os.execvp` is faster than `subprocess.call`

akdor1154:
I think this broke calling scripts on Windows - if I do `poetry run script.py` it used to get called as `subprocess.call(script.py, shell=True)` but now gets called as `Popen([script.py])` This doesn't work, see https://stackoverflow.com/questions/912830/using-subprocess-to-run-python-script-on-windows.

simnalamburt:
It’s actually calling `Popen([‘python3.exe’, blabla])` which seems OK according to your link https://stackoverflow.com/a/912847

benjirewis:
I've noticed killing poetry processes with SIGTERM creates a status code of 143. While that error code does indicate graceful termination after receiving SIGTERM, why not just return 0?

PR #1237, opened 2019-07-19
WIP - Plugin system

sdispater:
## Pull Request Check List

This is a WIP implementation of the previously mentionned plugin system. It is not fully functional at the moment (and does not work at all with the recommended installer). The implementation will likely change as I make progress on it. Nonetheless, the general idea of what it will look like is here.

Note that it will likely not make it to the `1.0.0` release.

### DIFF
--- a/docs/docs/plugins.md
+++ b/docs/docs/plugins.md
@@ -1,0 +1,147 @@
+# Plugins
+
+You may wish to alter or expand Poetry's functionality with your own.
+For example if your environment poses special requirements
+on the behaviour of Poetry which do not apply to the majority of its users
+or if you wish to accomplish something with Poetry in a way that is not desired by most users.
+
+In these cases you could consider creating a plugin to handle your specific logic.
+
+
+## Creating a plugin
+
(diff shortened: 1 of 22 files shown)


kapilt:
one comment, its a little unclear if this is present already, but I was wondering about ensuring that builtin commands can't be overridden to avoid inconsistencies in user experience.

keattang:
Eagerly awaiting this! Any ideas when it may get released?

danieleades:
This seems like a really powerful and flexible plugin system. I'm looking to seeing how people extend poetry's functionality using this framework to make it a powerful test runner and CI tool.

I also wonder if this is not quite heavy weight for the many people who are looking for a simple hook. Consider `Yarn`'s hook system where scripts definesdin the project manifest with 'special' names are invoked for specific methods. For example if you have a script defined called 'prepublish', that script will be called before publishing.

How can I use this plugin to create a simple hook at a specific point? My use-case, for example, is to run code-generation on some proto files. Creating an entire package to run a one line script seems excessive.

sdispater:
Closing in favor of #3733

PR #1240, opened 2019-07-19
Force an update when changing a dependency to point to a local directory

pdavies:
This fixes the following bug:

* Your pyproject.toml references a dependency
```
[tool.poetry.dependencies]
my-dependency = "^1.0.0"
```
* You realise you need to make parallel changes to the dependency, so you change the reference to point to a local checkout of the dependency:
```
[tool.poetry.dependencies]
my-dependency = { path = "../my-dependency/" }
```
* You run `poetry update` - but nothing happens.  It insists that the dependency is already installed, and we continue to use the installed copy in `site-packages`. Worse, `poetry show` now believes that you are using the local directory, even though you're not (as `pip show` confirms).

### DIFF
--- a/poetry/puzzle/solver.py
+++ b/poetry/puzzle/solver.py
@@ -84,6 +84,13 @@ class Solver:
                     elif package.version != pkg.version:
                         # Checking version
                         operations.append(Update(pkg, package))
+                    elif (
+                        package.source_type == "directory"
+                        and pkg.source_type != "directory"
+                    ):
+                        # If we are changing to installing from a local directory, force
+                        # the install to take place.
+                        operations.append(Update(pkg, package))
                     else:
                         operations.append(Install(package).skip("Already installed"))
 


poetry/puzzle/solver.py
```diff
@@ -84,6 +84,13 @@ def solve(self, use_latest=None):  # type: (...) -> List[Operation]
                     elif package.version != pkg.version:
                         # Checking version
                         operations.append(Update(pkg, package))
+                    elif (
+                        package.source_type == "directory"
+                        and pkg.source_type != "directory"
```
sdispater: I wonder if we can generalize this for any type of change in the dependency source type (`file` -> `directory` or `directory` -> `git` for instance)

dimbleby: That seems like it ought to be the right answer, but it looks as though the detection of source type for installed packages is not currently reliable:

- so far as I can see, it is [only ever set to "git"](https://github.com/sdispater/poetry/blob/64e8af3b4f19d3991a983745cfd313d37c92e802/poetry/repositories/installed_repository.py#L38) (or nothing at all)
- and actually, even packages that are installed as "directory" are reported as "git", if it happens that the code is in a git repository (which is fairly likely to be the case).
  - I suppose by the time you get to doing `pip freeze`, the two cases are indistinguishable

Presumably in an ideal world the source_type of installed packages would be made reliable - but it's not clear to me how to achieve that.

jacebrowning:
@pdavies Are you still working on this? If not, I can try to work on a fix for this bug.

dimbleby:
This should be fixed in 1.0, by bf2d515 - [here](https://github.com/sdispater/poetry/blob/2fa4678d9c7fc383ca3393273cc5b03a397482d7/poetry/puzzle/solver.py#L95)

pdavies:
Whoops, I haven't been on GItHub in ages. Happy to see that this has been fixed independently!

PR #1268, opened 2019-07-31
Add semver utils, using semver library

dazza-codes:
- boolean check for a sem-ver string
- filter and sort a sequence of sem-ver strings
  - these utils are used to filter and sort a list of git tags in a subsequent PR

The focus of this PR is only on sem-ver support and not full PEP-0440 support, see:
- https://legacy.python.org/dev/peps/pep-0440/#semantic-versioning

> Semantic versions containing a hyphen (pre-releases - clause 10) or a plus sign (builds - clause 11) are not compatible with this PEP and are not permitted in the public version field.

This implies that using `parse_single_contraint` and the poetry `Version` implementations may not be compatible with the intention of supporting sem-ver (assuming poetry aims only to support PEP-0440, although the poetry philosophy seems to be not to impose a specific versioning strategy on the user).  This PR is focused on sem-ver support (given that full sem-ver support is not 100% compatible with PEP-0440).  The responsibility for sem-ver parsing belongs with the `semver` dependency, to avoid recreating that wheel.

## checklist
- [x] Added tests for changed code.
- [x] Updated documentation for changed code
  - no API docs required for this, AFAIK

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -14,7 +14,7 @@ marker = "python_version >= \"2.7\" and python_version < \"2.8\" and (sys_platfo
 name = "asn1crypto"
 optional = false
 python-versions = "*"
-version = "0.24.0"
+version = "1.0.1"
 
 [[package]]
 category = "dev"
@@ -41,12 +41,7 @@ description = "Classes Without Boilerplate"
 name = "attrs"
 optional = false
 python-versions = ">=2.7, !=3.0.*, !=3.1.*, !=3.2.*, !=3.3.*"
-version = "19.1.0"
-
-[package.extras]
-dev = ["coverage", "hypothesis", "pympler", "pytest", "six", "zope.interface", "sphinx", "pre-commit"]
-docs = ["sphinx", "zope.interface"]
-tests = ["coverage", "hypothesis", "pympler", "pytest", "six", "zope.interface"]
+version = "19.2.0"
 
 [[package]]
 category = "dev"
@@ -63,9 +58,6 @@ attrs = ">=18.1.0"
 click = ">=6.5"
 toml = ">=0.9.4"
 
-[package.extras]
-d = ["aiohttp (>=3.3.2)", "aiohttp-cors"]
-
 [[package]]
 category = "main"
 description = "httplib2 caching for requests"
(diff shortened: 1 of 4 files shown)


tests/semver/test_sem_ver_utils.py
```diff
@@ -0,0 +1,46 @@
...
+
+@pytest.mark.parametrize(
+    "constraint,result",
+    [
+        ("*", True),
+        ("*.*", True),
+        ("v*.*", True),
+        ("*.x.*", True),
+        ("x.X.x.*", True),
+        # ('!=1.0.0', Constraint('!=', '1.0.0.0')),
```
brycedrennan: whats going on here?

tests/semver/test_sem_ver_utils.py
```diff
@@ -0,0 +1,46 @@
+import pytest
+
+from poetry.semver import is_sem_ver_constraint
+from poetry.semver import sem_ver_sorted
+
+
+@pytest.mark.parametrize(
+    "constraint,result",
```
brycedrennan: should these also be handled? 

https://legacy.python.org/dev/peps/pep-0440/#local-version-identifiers

tests/semver/test_sem_ver_utils.py
darrenleeweber: - adapted from `tests/semver/test_main.py`
- enabled and passing in amended commit (to follow)

tests/semver/test_sem_ver_utils.py
darrenleeweber: - punt to backlog/requirement for parse_constraint
- there are more details on those specifics in tests/semver/test_main.py
- a simpler test on this utility is just to mock/spy `parse_single_contraint` to check that it is called with given arguments and leave all the details to the tests on that function; would that be better?

darrenleeweber: - added doc strings
- added mock-spy on parser
- done with amended commits (for now)
- 🎾 > @brycedrennan

darrenleeweber:
All review comments addressed
- leaving one conversion for @brycedrennan to consider/resolve

darrenleeweber:
I'm kinda giving up on contributing to poetry
- the review and merge cycle takes too long
- the project doesn't appear to be open to easy modifications or even simple refactoring

Issue #618, opened 2018-11-12
Add support for per-project settings


sdispater:
- [x] I have searched the [issues](https://github.com/sdispater/poetry/issues) of this repo and believe that this is not a duplicate.

## Issue

Sometimes you might want some settings to be set on a per-project basis. For instance, you might want your virtualenv to be created inside the project's directory only for the current project.

To make this possible, we could add a `--local` option to the `config` command.

The example use case above would then look like this:

```bash
poetry config settings.virtualenvs.in-project true --local
```

And alternative way would be to make namespace command: `config:local`:

```bash
poetry config:local settings.virtualenvs.in-project true
```

jacebrowning:
Has any consideration been given to storing some of these local settings in `pyproject.toml`? I would like every user of a given project to have a virtual environment at `./venv/` so that other tools/scripts can depend on it.

GuillaumeDesforges:
Would be a very useful feature. Or should I say that it is a must-have?

jacebrowning:
To further expand on my proposal, perhaps Poetry could build settings in the following order:
1. Use the value from `pyproject.toml` (possibly a new `[tool.poetry.meta]` section)
2. Use the globally set value from `$ poetry config`
3. Use the default value

jobec:
That matches with how `git config` does it, except for one more level: `system`
https://git-scm.com/docs/git-config#Documentation/git-config.txt---global

Pip also does this in 3 levels:
https://pip.pypa.io/en/stable/user_guide/#config-file

The system wide setting is something I use quite often. It allows all users on a system to have the same settings, irrespective of who is setting up an environment. That might be useful when different users update an application on a server, for example.

drunkwcodes:
@jobec There is a better place: #964 even if you are talking about `git config --system`.

P.S I'm thinking about having a gitter channel for this at the same time.

raxod502:
Can I put in a request for passing config options on the command line, like `git -c`? I am writing a tool that automatically runs Poetry. The tool needs to set these options for certain commands, but I don't want it to overwrite the user's configuration file. Updating and then automatically reverting the configuration file is super disgusting, but I can't think of any better way to do it with the current release of Poetry.

kylebebak:
Just want to second @jacebrowning 's proposal -- project-specific settings would be most useful to devs if they lived in `pyproject.toml`.

[...]

PR #1272, opened 2019-07-31
Improve the way settings are managed

sdispater:
## Pull Request Check List

This PR is a refactoring and improvement of the configuration part of Poetry which introduces the following changes.

### Local, per-project settings

The `config` command now supports a `--local` option which will store the given settings to a local `poetry.toml` file.

```bash
poetry config --local virtualenvs.in-project true
```

### Configuration via environment variables

Poetry can now be configured using environment variables. This is particularly useful in CI environments and avoids having to create configuration files.

The environment variables must be prefixed by `POETRY_` and are comprised of the uppercase
name of the setting with dots and dashes replaced by underscore, here is an example:

```bash
export POETRY_VIRTUALENVS_PATH=/path/to/virtualenvs/directory
```

This also works for secret settings, like credentials:

```bash
export POETRY_HTTP_BASIC_MY_REPOSITORY_PASSWORD=secret
```

### The `settings.` prefix is no longer necessary

This one is a breaking change. The settings are now stored differently and no longer needs to be prefixed by `settings.`. If you have already configured settings you will need to configure them again.


poetry/config/config.py
```diff
@@ -0,0 +1,115 @@
...
+from typing import Optional
+
+from poetry.locations import CACHE_DIR
+from poetry.utils._compat import Path
+from poetry.utils._compat import basestring
+
+_NOT_SET = object()
+
+boolean_validator = lambda val: val in {"true", "false", "1", "0"}
+boolean_normalizer = lambda val: True if val in ["true", "1"] else False
```
yedpodtrzitko:
```suggestion
boolean_normalizer = lambda val: val in {"true", "1"}
```

sdispater: Fixed. Thanks!

levrik:
I'm quite surprised that the local configuration is written to its own file instead of using `pyproject.toml`. What's the reason for that?

notpushkin:
Sorry for bumping this but I'm interested in this decision, too. I understand it allows storing per-project settings while git-ignoring the `poetry.toml` file, but for some projects, storing .venv relatively to the repository is a must (e. g. some scripts depend on this, as noted in https://github.com/python-poetry/poetry/issues/618). Is there a way to support reading from both `pyproject.toml` and `poetry.toml`?

austinmw:
I've read through the various related issues, but is it still not possible to specify a local venv from the pyproject.toml file? For example:

```toml
[settings.virtualenvs]
in-project = "true"
```

PR #1275, opened 2019-08-02
Add support for PyPI API tokens

sdispater:
## Pull Request Check List

PyPI recently added support for token-based authentication to upload packages: https://pypi.org/help/#apitoken

This PR adds support for that use case in Poetry.

Users can now configure tokens via the `config` command:

```bash
poetry config pypi-token.pypi my-token
```

It's also possible to provide an environment variable holding the value of the token:

```bash
export POETRY_PYPI_TOKEN_PYPI=my-token
```

### DIFF
--- a/docs/docs/repositories.md
+++ b/docs/docs/repositories.md
@@ -39,8 +39,18 @@ If you do not specify the password you will be prompted to write it.
 
 !!!note
 
-    To publish to PyPI, you can set your credentials for the repository
-    named `pypi`:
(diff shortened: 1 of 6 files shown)


danieleades:
just for some clarification, is this intended to support token-based authentication on alternative pypi registries, or only the public registry?

I can confirm that i can add packages from my local (artifactory) pypi registry using username/password auth, but not with an API key

Issue #1303, opened 2019-08-15
Package Name Case Sensitivity


iwoloschin:
- [x] I am on the [latest](https://github.com/sdispater/poetry/releases/latest) Poetry version.
- [x] I have searched the [issues](https://github.com/sdispater/poetry/issues) of this repo and believe that this is not a duplicate.
- [x] If an exception occurs when executing a command, I executed it again in debug mode (`-vvv` option).

- **OS version and name**: macOS 14.6
- **Poetry version**: 1.0.0b1

## Issue
I'm trying to specify a development version of Logbook:
```toml
logbook = {rev = "185b9ed959f2a41b605bbd93bdbd83ed9efbc5bd",git = "https://github.com/getlogbook/logbook.git"}
```
Poetry is complaining:
```
$ poetry update
Updating dependencies
Resolving dependencies... (1.9s)

[RuntimeError]
The dependency name for logbook does not match the actual package's name: Logbook

Traceback (most recent call last):
  File "/Users/iwolosch/.poetry/lib/poetry/_vendor/py3.7/clikit/console_application.py", line 132, in run
    status_code = command.handle(parsed_args, io)
  File "/Users/iwolosch/.poetry/lib/poetry/_vendor/py3.7/clikit/api/command/command.py", line 119, in handle
    status_code = self._do_handle(args, io)
  File "/Users/iwolosch/.poetry/lib/poetry/_vendor/py3.7/clikit/api/command/command.py", line 167, in _do_handle
    return getattr(handler, handler_method)(args, io, self)
  File "/Users/iwolosch/.poetry/lib/poetry/_vendor/py3.7/cleo/commands/command.py", line 92, in wrap_handle
    return self.handle()
  File "/Users/iwolosch/.poetry/lib/poetry/console/commands/update.py", line 49, in handle
    return installer.run()
  File "/Users/iwolosch/.poetry/lib/poetry/installation/installer.py", line 74, in run
    self._do_install(local_repo)
  File "/Users/iwolosch/.poetry/lib/poetry/installation/installer.py", line 161, in _do_install
    ops = solver.solve(use_latest=self._whitelist)
  File "/Users/iwolosch/.poetry/lib/poetry/puzzle/solver.py", line 38, in solve
    packages, depths = self._solve(use_latest=use_latest)
  File "/Users/iwolosch/.poetry/lib/poetry/puzzle/solver.py", line 173, in _solve
    self._package, self._provider, locked=locked, use_latest=use_latest
[...]

PR #1305, opened 2019-08-18
Use canonical name when validating package from directory

abn:
This change handles cases where a project uses a non-canonical name in its setup configuration (eg: [Logbook](https://github.com/getlogbook/logbook/blob/develop/setup.py#L188).

Incorrect behavior was introduced with c6a2d6e17772baf31fae4ece3f8b2449e2a36d97 where the comparison used the raw name value instead of the canonical name.

Closes: #1303 

### Pull Request Check List


brycedrennan:
@abn thanks for contributing!

Does this bug exist in master as well? if so we should make the pull request into master

abn:
@bradsbrown this bug was introduced with https://github.com/sdispater/poetry/commit/c6a2d6e17772baf31fae4ece3f8b2449e2a36d97 which was only merged to develop. The previous logic did the right thing and hence the bug should not exist on master.

Issue #1297, opened 2019-08-13
[EnvCommandError] when installing package from URL


dpraul:
- [x] I am on the [latest](https://github.com/sdispater/poetry/releases/latest) Poetry version.
- [x] I have searched the [issues](https://github.com/sdispater/poetry/issues) of this repo and believe that this is not a duplicate.
- [x] If an exception occurs when executing a command, I executed it again in debug mode (`-vvv` option).

- **OS version and name**: Windows 10
- **Poetry version**: 1.0.0b1

<details><summary>pyproject.toml</summary>
<p>

```toml
[tool.poetry]
name = "fact-extract"
version = "0.0.0"
description = "description"
authors = ["Dylan <email@example.com>"]

[tool.poetry.dependencies]
python = "^3.7"
click = "^7.0"
scrapy = "^1.7"
textacy = "^0.8.0"

[tool.poetry.dev-dependencies]
pytest = "^3.0"
black = {version = "^19.3-beta.0", allows-prereleases = true}

[build-system]
requires = ["poetry>=0.12"]
build-backend = "poetry.masonry.api"

```

</p>  
</details>

## Issue

Using the `poetry add` command doesn't appear to work with URLs all the time. In particular, installing a spaCy (nlp library) model from a GitHub release (this is their [recommended way to do it for production usage](https://spacy.io/usage/models#download-pip)). Totally guessing but the issue may be because the package doesn't exist on PyPI, or maybe doesn't report its version correctly. 

Here is the command and output:

```
C:\projects\fts\fact-extract>poetry add https://github.com/explosion/spacy-models/releases/download/en_core_web_sm-2.1.0/en_core_web_sm-2.1.0.tar.gz#egg=en_core_web_sm

Updating dependencies
Resolving dependencies...


Package operations: 4 installs, 0 updates, 0 removals

  - Installing en_core_web_sm (2.1.0 https://github.com/explosion/spacy-models/releases/download/en_core_web_sm-2.1.0/en_core_web_sm-2.1.0.tar.gz#egg=en_core_web_sm)

[EnvCommandError]
Command ['C:\\Users\\Dylan\\AppData\\Local\\pypoetry\\Cache\\virtualenvs\\fact-extract-LQupvckr-py3.7\\Scripts\\python.exe', '-m', 'pip', 'install', '--no-deps', 'en-core-web-sm==2.1.0'] errored with the following return code 1, and output:
Collecting en-core-web-sm==2.1.0
[...]

PR #1260, opened 2019-07-28
Add support for url dependencies

sdispater:
## Pull Request Check List

- [x] Added **tests** for changed code.
- [x] Updated **documentation** for changed code.

This PR adds support for URL dependencies.

Adding a URL dependency to the current project can be done in two ways: via the `add` command or by modifying the `pyproject.toml` file directly.

```bash
poetry add https://example.com/packages/my-package-1.0.0.tar.gz
```

```toml
[tool.poetry.dependencies]
my-package = {url = "https://example.com/packages/my-package-1.0.0.tar.gz"}
```

PR #1306, opened 2019-08-18
Fix support for url sourced packages in pip installer

abn:
This change fixes the installation of packages using a url source.
Previously, the installer attempted to install the package using the
name ignoring the source url.

Closes: #1297 

## Pull Request Check List

### DIFF
--- a/poetry/installation/pip_installer.py
+++ b/poetry/installation/pip_installer.py
@@ -138,6 +138,9 @@ class PipInstaller(BaseInstaller):
                 package.source_url, package.source_reference, package.name
             )
 
+        if package.source_type == "url":
+            return "{}#egg={}".format(package.source_url, package.name)
+
         return "{}=={}".format(package.name, package.version)
 
     def create_temporary_requirement(self, package):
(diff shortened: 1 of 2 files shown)


brycedrennan:
You're on a roll with these fixes! thanks!

brycedrennan:
I'm trying to figure out how this interacts with
 - https://pip.pypa.io/en/stable/news/#id137
 - https://github.com/pypa/pip/issues/4187
...especially if there are security implications. I'll need time to get up to speed. Any insights you have welcome.

abn:
@brycedrennan my 2 cents here is that using a url to install a dependency, while not the best of solutions, has valid scenarios (#1297 being an example). This should be allowed at a project maintainer's discretion - hopefully with an understanding of the side-effects. Short answer to your question regarding arbitrary packages is, yes.

The changes in this PR are aimed at only fixing the current expected behavior that was introduced in #1260 .

I do think that we should keep track of the pypa community discussions around dependency links as we are relying on pip under the hood for installation. Additionally, I'd consider the security discussion to be of broader scope and warranting a separate issue discussing any actions, even if it is adding further documentation, that needs to be taken. With web url's, we could probably rely on content hash to detect changes but that would most likely be the extent of low-effort technical solutions.

brycedrennan:
well said. I agree that this is part of a broader discussion.  I'll take a look at this.

sdispater:
Thanks!

PR #1310, opened 2019-08-20
Fix a silent upload error

### LINKED ISSUE #858: Upload to PyPI silently fail

- **OS version and name**: MacOS Mojave
- **Poetry version**: 0.12.10
- [pyproject.toml](https://gist.github.com/jpscaletti/030ba432b413d32565d47e065b9badb7)

## Issue
`poetry publish` fails silently if the Pypi server returns an error:

1. If I use invalid credentials (it even shows the "Uploading foobar.tar.gz 100%" mesage)
2. If the Pypi server returns an error.

I was trying to upload a project with an invalid name ("pipes", invalid because Python standard library has a module named the same) and nothing indicated me that the publishing was failing.

If I try with twine this is the output instead:
```
$ twine upload dist/*
Uploading distributions to https://upload.pypi.org/legacy/
Uploading Pipes-0.1.0-py3-none-any.whl
100%|████████████████████| 4.26k/4.26k [00:00<00:00, 7.88kB/s]
NOTE: Try --verbose to see response content.
HTTPError: 403 Client Error: The user 'jpscaletti' isn't allowed to upload to
project 'pipes'. See https://pypi.org/help/#project-name for more information.
for url: https://upload.pypi.org/legacy/
```

I think #742 is a specific case of this issue.

sztomi:
Fixes #858 by first issuing a HEAD request and following a redirect.
This does not cover further redirects though, so it's possible to make this more robust. It does cover the most common use cases (PyPI and TestPyPI will work correctly)

See my comment for more details: https://github.com/sdispater/poetry/issues/858#issuecomment-520239647

# Pull Request Check List

- [ ] Updated **documentation** for changed code. (n/a: this is a bugfix)

### DIFF
--- a/poetry/masonry/publishing/uploader.py
+++ b/poetry/masonry/publishing/uploader.py
@@ -6,6 +6,7 @@ import re
 from typing import List
 
 import requests
+import requests.status_codes as status_codes
 
 from requests import adapters
 from requests.exceptions import HTTPError
@@ -228,6 +229,11 @@ class Uploader:
                 encoder, lambda monitor: bar.set_progress(monitor.bytes_read)
             )
 
+            # Check for a redirect first
+            resp = session.head(url)
+            if resp.status_code == status_codes.codes.moved:
+                url = resp.headers["Location"]
+
             bar.start()
 
             resp = session.post(
@@ -237,7 +243,7 @@ class Uploader:
                 headers={"Content-Type": monitor.content_type},
             )
 
-            if resp.ok:
+            if resp.status_code == status_codes.codes.ok:
                 bar.finish()
 
                 self._io.writeln("")
--- a/tests/console/commands/test_publish.py
+++ b/tests/console/commands/test_publish.py
@@ -2,6 +2,7 @@ def test_publish_returns_non_zero_code_for_upload_errors(app, app_tester, http):
     http.register_uri(
         http.POST, "https://upload.pypi.org/legacy/", status=400, body="Bad Request"
     )
+    http.register_uri(http.HEAD, "https://upload.pypi.org/legacy/")
 
     exit_code = app_tester.run(
         [("command", "publish"), ("--username", "foo"), ("--password", "bar")]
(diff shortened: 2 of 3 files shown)


sztomi:
Unfortunately I do have a failing unit test and I don't know how to fix it. 

This test fails:

```
    expected = """
Publishing simple-project (1.2.3) to PyPI
 - Uploading simple-project-1.2.3.tar.gz 0%
 - Uploading simple-project-1.2.3.tar.gz 100%
 - Uploading simple-project-1.2.3.tar.gz 100%

[UploadError]
HTTP Error 400: Bad Request

"""

    assert app_tester.get_display(True).startswith(expected)
```

The output gets unexpected whitespace, but I have no idea why. Any help is appreciated.

image

tests/console/commands/test_publish.py
```diff
@@ -14,10 +15,10 @@ def test_publish_returns_non_zero_code_for_upload_errors(app, app_tester, http):
  - Uploading simple-project-1.2.3.tar.gz 0%
  - Uploading simple-project-1.2.3.tar.gz 100%
  - Uploading simple-project-1.2.3.tar.gz 100%
-
```
brycedrennan: Looks like you removed whitespace here - was that intentional? Might be related to the test failure.

poetry/masonry/publishing/uploader.py
```diff
@@ -228,6 +229,11 @@ def _upload_file(self, session, url, file):
                 encoder, lambda monitor: bar.set_progress(monitor.bytes_read)
             )
 
+            # Check for a redirect first
+            resp = session.head(url)
+            if resp.status_code == status_codes.codes.moved:
```
brycedrennan: This approach means we'd have to make two requests every time, even when there is no redirect. Wouldn't we be able to just try the endpoint and then follow the redirect if necessary?

sztomi: That's what I did originally, but that means POSTing the package data twice, so anyone who configured the wrong URL for their repo will have a big perf hit. On the other hand, a HEAD request is very cheap and isn't noticeable. Happy to do some benchmarks if you are not convinced.

tests/console/commands/test_publish.py
sztomi: oh, I totally did. Don't know why I didn't notice that. But it explains my test failure. I'll correct that.

sztomi:
Actually, I realized that 301 means _permanently_ moved. So I'm going to open a different PR to follow the redirect at the time it is saved in the config.

sztomi:
Superseded by the linked PR above ^

poetry/masonry/publishing/uploader.py
brycedrennan: That's a good point that it would be expensive to POST twice.

Issue #1257, opened 2019-07-26
Problem vendoring subprocess32 in Ubuntu


sklarsa:
- [x] I am on the [latest](https://github.com/sdispater/poetry/releases/latest) Poetry version.
- [x] I have searched the [issues](https://github.com/sdispater/poetry/issues) of this repo and believe that this is not a duplicate.
- [x] If an exception occurs when executing a command, I executed it again in debug mode (`-vvv` option).

- **Ubuntu 16.04/18.04**: 

## Issue
I'm building a Dockerfile and am seeing a RuntimeWarning after running any poetry command that works with the venv.  Since I'm using  poetry to run a web server (gunicorn), I'm worried that this could have a negative impact on app performance.  I haven't seen this before, so it appears to have been introduced recently?

The warning:
```
/root/.poetry/lib/poetry/_vendor/py2.7/subprocess32.py:149: RuntimeWarning: The _posixsubprocess module is not being used. Child process reliability may suffer if your program uses threads.
  "program uses threads.", RuntimeWarning)
```

I installed poetry using the `get-poetry.py` script, and have even updated to the latest version.  Here is the Dockerfile that I'm using, and the relevant output:
```
FROM ubuntu:16.04
RUN apt-get update && apt-get install -y \
    software-properties-common \
    curl \
    htop \
    python \
    python-dev \
    wget
RUN curl -sSL https://raw.githubusercontent.com/sdispater/poetry/master/get-poetry.py | python
ENV PATH=$PATH:/root/.poetry/bin
RUN poetry self:update --preview
RUN poetry version
```

And the output for the final command...
```
$: poetry version
/root/.poetry/lib/poetry/_vendor/py2.7/subprocess32.py:149: RuntimeWarning: The _posixsubprocess module is not being used. Child process reliability may suffer if your program uses threads.
  "program uses threads.", RuntimeWarning)

[ImportError]
No module named pkg_resources
```

marcelomendes:
I'm having same issue, Debian current, Python 3.4

autoferrit:
I am getting this under latest ubuntu 18 under windows WSL. When installing, I used a python 3.7 version to run it, which came from pyenv.

michielboekhoff:
Hey there. Upon encountering the same issue, I implemented a quick fix for people using Python 3. The installer previously did not respect the installing Python version, which I propose to fix in #1312.

michielboekhoff:
[...]

PR #1312, opened 2019-08-21
Respect Python version in bin script

michielboekhoff:
- [ N/A ] Added **tests** for changed code.
- [ ✔️ ] Updated **documentation** for changed code.
---

This PR makes the Poetry installer `get-poetry.py` respect the installing version of Python in installing the `~/.poetry/bin/poetry` script. 

Simply, if `python3` is used to install Poetry, then use that as the Python executable to run Poetry as. If it is `python` (e.g. Python 2), then use `python`.

This should fix the issues some people are having in #1257.

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,7 +14,7 @@ It supports Python 2.7 and 3.4+.
 
 Poetry provides a custom installer that will install `poetry` isolated
 from the rest of your system by vendorizing its dependencies. This is the
-recommended way of installing `poetry`.
+recommended way of installing `poetry`. The installer installs a script to run Poetry, which will use the version of Python used to install Poetry.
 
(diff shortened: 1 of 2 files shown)


brycedrennan:
Thanks for contributing to poetry! This duplicates open pull request #1042.  

I have [concerns](https://github.com/sdispater/poetry/pull/1042#pullrequestreview-278187471) about the (nearly identical) approach that was taken in both pull requests.  I'd welcome your thoughts on what the best way forward is.

michielboekhoff:
Thanks @brycedrennan! Apologies for the duplication, I didn't see that - I'll put my thoughts on the PR and on the issue.

PR #1320, opened 2019-08-23
remove NoReturn, use None instead

toppk:

I don't think the use of NoReturn is correct.  These functions return, just with None

pinging original author @abn of the original feature.

https://github.com/sdispater/poetry/issues/210
https://github.com/sdispater/poetry/pull/774

For reference, this breaks pypy compatibility, which has an older version of typing.

### DIFF
--- a/poetry/utils/helpers.py
+++ b/poetry/utils/helpers.py
@@ -7,7 +7,6 @@ import tempfile
 
 from contextlib import contextmanager
 from typing import List
-from typing import NoReturn
 from typing import Optional
 
 from keyring import delete_password, set_password, get_password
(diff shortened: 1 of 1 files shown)


brycedrennan:
@toppk According to [this explanation](https://stackoverflow.com/questions/48038149/noreturn-vs-none-in-void-functions-type-annotations-in-python-3-6) you are correct.  I'm not super familiar with python typing: aren't these just comments? how does this break compatibility?

dmontagu:
The `NoReturn` is being imported, which will cause an error if not present in the implementation’s typing module. 

I think the idea that they are comments might stem from the fact that if you include `from __future__ import annotations`, the annotations are all strings are not evaluated. But even then, if you include typing-related imports, they will be executed just like any other python import.

Given the function isn’t actually a NoReturn (which should only be used if the function is guaranteed to throw an exception under all circumstances) it seems to make sense to change this, especially given that it is causing compatibility issues.

abn:
@toppk good catch; over looked that one. :+1:

abn:
@brycedrennan can we get this merged please?

Issue #1318, opened 2019-08-23
Poetry 1.0.0b1 Fails to Install Wheel Files (`[TypeError] 'NoneType' object is not iterable'`)


KyleKing:
- [X] I am on the [latest](https://github.com/sdispater/poetry/releases/latest) Poetry version.
- [X] I have searched the [issues](https://github.com/sdispater/poetry/issues) of this repo and believe that this is not a duplicate.
- [X] If an exception occurs when executing a command, I executed it again in debug mode (`-vvv` option).

- **OS version and name**: Windows 7 (will test on Mac tonight)
- **Poetry version**: 1.0.0b1
- **Link of a [Gist](https://gist.github.com/) with the contents of your pyproject.toml file**: N/A - toml file is irrelevant. Reproduce with `poetry new ...`. Script below

## Issue

Installing wheel files fails with the latest version of Poetry. I've tried multiple wheel files and seems to be a systemic bug.

This was probably related to the changes to remove `--git` and allowing the URL to be specified directly. I'll take a look tonight, but I'll be away this weekend and wanted to share a bug report

```sh
poetry add ./itsdangerous-1.1.0-py2.py3-none-any.whl -vvv
Using virtualenv: C:\Users\king.kyle\AppData\Local\pypoetry\Cache\virtualenvs\testnoniterablenonetype-NTCsUG5T-py3.7

[TypeError]
'NoneType' object is not iterable

Traceback (most recent call last):
  File "c:\users\king.kyle\appdata\local\continuum\anaconda2\envs\py372\lib\site-packages\clikit\console_application.py", line 132, in run
    status_code = command.handle(parsed_args, io)
  File "c:\users\king.kyle\appdata\local\continuum\anaconda2\envs\py372\lib\site-packages\clikit\api\command\command.py", line 119, in handle
    status_code = self._do_handle(args, io)
  File "c:\users\king.kyle\appdata\local\continuum\anaconda2\envs\py372\lib\site-packages\clikit\api\command\command.py", line 167, in _do_handle
    return getattr(handler, handler_method)(args, io, self)
  File "c:\users\king.kyle\appdata\local\continuum\anaconda2\envs\py372\lib\site-packages\cleo\commands\command.py", line 92, in wrap_handle
    return self.handle()
  File "c:\users\king.kyle\appdata\local\continuum\anaconda2\envs\py372\lib\site-packages\poetry\console\commands\add.py", line 89, in handle
[...]

PR #1322, opened 2019-08-24
Fix file installation for wheels with no Requires-Dist

abn:
This issue does not affect master as the relevant code only exists on develop.

Closes: #1318 

# Pull Request Check List


poetry/utils/inspector.py
```diff
@@ -49,7 +49,7 @@ def inspect_wheel(
             "version": "",
             "summary": "",
             "requires_python": None,
-            "requires_dist": None,
+            "requires_dist": [],
```
KyleKing: Thanks for the quick PR last weekend! I've been using a version of this fix for the last week without issue

In the `inspect_sdist` method, the `requires_dist` key is type cast as a list ([line 104](https://github.com/sdispater/poetry/blob/2464c19196a89a109cbe2c810835f0e0bc896722/poetry/utils/inspector.py#L104)). For consistency, [line 73](https://github.com/sdispater/poetry/blob/2464c19196a89a109cbe2c810835f0e0bc896722/poetry/utils/inspector.py#L73) could be `info["requires_dist"] = list(meta.requires_dist)` instead of `info["requires_dist"] = meta.requires_dist`

KyleKing: FYI: updated my prior comment. Also to add, in the case that `pkginfo.Wheel(str(file_path))` throws a ValueError, `meta` is undefined
https://github.com/sdispater/poetry/blob/2464c19196a89a109cbe2c810835f0e0bc896722/poetry/utils/inspector.py#L55-L60

Which would cause an error in a subsequent line

https://github.com/sdispater/poetry/blob/2464c19196a89a109cbe2c810835f0e0bc896722/poetry/utils/inspector.py#L70-L72

Which should probably be:

```py
if meta.requires_python:
    info["requires_python"] = meta.requires_python
```

sdispater:
Thanks!

PR #1326, opened 2019-08-25
Fix pytest fixtures on Win10

Caligatio:
# Pull Request Check List

---
Two conftest.py files are currently attempting to perform symlinking as part of a fixture which requires either [admin privileges or "Developer Mode" to be enabled](https://blogs.windows.com/windowsdeveloper/2016/12/02/symlinks-windows-10/) on Win10.  This PR just wraps the symlink call in a try/except and falls back to copying if it fails.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -87,8 +87,14 @@ def mock_download(self, url, dest):
     if dest.exists():
         os.unlink(str(dest))
 
-    if PY2 and WINDOWS:
-        shutil.copyfile(str(fixture), str(dest))
+    if WINDOWS:
+        if PY2:
+            shutil.copyfile(str(fixture), str(dest))
+        else:
+            try:
+                os.symlink(str(fixture), str(dest))
+            except OSError:
+                shutil.copyfile(str(fixture), str(dest))
     else:
         os.symlink(str(fixture), str(dest))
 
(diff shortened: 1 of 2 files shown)


tests/conftest.py
```diff
@@ -87,8 +87,14 @@ def mock_download(self, url, dest):
     if dest.exists():
         os.unlink(str(dest))
 
-    if PY2 and WINDOWS:
-        shutil.copyfile(str(fixture), str(dest))
+    if WINDOWS:
```
brycedrennan: I'm tempted to replace this whole block with:
```python
	    try:
                os.symlink(str(fixture), str(dest))
            except OSError:
                # symlink doesn't work on windows in python 2 or without admin permissions
                shutil.copyfile(str(fixture), str(dest))

Caligatio: So os.symlink isn't available on Windows in Python 2 which would then maybe throw an `AttributeError` instead? I was thinking along the same lines as your try/except and also was unsure given the previous use of flags :)

brycedrennan: Makes sense. Perhaps an inline comment is in order regarding the reasoning (admin privileges etc)?

Caligatio: Added and pushed

Issue #1249, opened 2019-07-24
pip 19.2. breaks poetry (private repository)


eliasmistler:
- [X] I am on the [latest](https://github.com/sdispater/poetry/releases/latest) Poetry version.
- [X] I have searched the [issues](https://github.com/sdispater/poetry/issues) of this repo and believe that this is not a duplicate.
- [x] If an exception occurs when executing a command, I executed it again in debug mode (`-vvv` option).

- **OS version and name**: OSX Mojave 10.14.5
- **Poetry version**: 0.12.17 (latest)
- **Pip version**: 19.2.1. (latest)

*Note*: This issue is closely related to [pip issue #6775](https://github.com/pypa/pip/issues/6775)

## Issue

I am using a private repository for my packages, which is working with the latest poetry and pip 19.1.1. With pip 19.2.1, I am getting the following error:

```
ERROR: Could not install packages due to an EnvironmentError: Failed to parse: https://me@mycompa.ny:mypassword@my/pypi/simple/mylib/ 
```

This happens with essentially every install command, i.e. `update`, `add`, `install`.

As per [pip issue #6775](https://github.com/pypa/pip/issues/6775), this is actually the correct behaviour, and I should be encoding my variables. I can do that and it works directly in pip 19.2.1:
```bash
alias urlencode='python3 -c "import sys; from urllib.parse import quote; print(quote(sys.argv[1]))"'

pip install --index-url="https://$(urlencode $MY_USER):$(urlencode $MY_PASSWORD)@my/pypi/simple" mylib
```

My expectation was that `poetry` would take care of the url encoding where necessary. However, I gave it a try by using the already encoded username and password:

```bash
poetry config http-basic.previse $(urlencode $MY_USER) $(urlencode $MY_PASSWORD)
```

That however yields the following exception, suggesting that my private repo was actually not being used:
```
Package [mylib] not found.
```

As a side note, maybe logging the repository being queried would be useful - I did not get any information about it when using `-vvv`.

Caligatio:
Can you check to see if #1341 solves your problem?

PR #1341, opened 2019-08-30
Add support for special characters in username/password for pip

Caligatio:
# Pull Request Check List

----

Per #1249 and [pip #6776](pip/issues#6775), newer versions of pip require require the username and password to be properly encoded/quoted.  This PR simply `quote`s the URL when calling pip.

As referenced in #1249, we do not want to pre-quote the strings as that fouls up the metadata fetch which does not use pip.

### DIFF
--- a/poetry/repositories/legacy_repository.py
+++ b/poetry/repositories/legacy_repository.py
@@ -21,6 +21,11 @@ from typing import Generator
 from typing import Optional
 from typing import Union
 
+try:
+    from urllib.parse import quote
+except ImportError:
+    from urllib import quote
+
 import requests
 
 from cachecontrol import CacheControl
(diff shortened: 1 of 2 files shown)


stephsamson:
Does this bug also exist on master? If so, then this PR should be merged into master and not into develop.

oleastre:
See my comment on #1402 , the fix should be applied to `master` branch as well, but the code changed between the two versions and this patch can not be applied as is to `master`

Caligatio:
As @oleastre mentioned, there is a significant code divergence between `master` and `develop` so the patch cannot be applied to `master`. My recommendation is to not let perfect be the enemy of good and just merge both #1402 and this PR.

Honestly, this `master` vs `develop` situation is a giant pain and, as highlighted by this situation, is impeding work. It's only getting worse since there was a major refactor in `develop` that changes many core functions.

sdispater:
Thanks!

PR #1342, opened 2019-08-30
Follow repo perm redirect

sztomi:
This PR supersedes https://github.com/sdispater/poetry/pull/1310 and fixes https://github.com/sdispater/poetry/issues/858

In the documentation, I took the liberty to change the "foo.bar" domain to that of TestPyPI, because it's a common URL that people might want to add.

# Pull Request Check List

### DIFF
--- a/docs/docs/repositories.md
+++ b/docs/docs/repositories.md
@@ -22,10 +22,13 @@ repository.
 Adding a new repository is easy with the `config` command.
 
 ```bash
-poetry config repositories.foo https://foo.bar/simple/
+poetry config repositories.testpypi https://test.pypi.org/legacy/
 ```
 
-This will set the url for repository `foo` to `https://foo.bar/simple/`.
+This will set the url for repository `testpypi` to `https://test.pypi.org/legacy/`.
+
+If the URL provided returns a permanent redirect, the redirect location will
+be saved instead. Most commonly, PyPI returns a redirect from `/legacy` to `/legacy/`.
 
 ### Configuring credentials
 
--- a/poetry/console/commands/config.py
+++ b/poetry/console/commands/config.py
@@ -7,6 +7,7 @@ from cleo import option
 from poetry.utils.helpers import (
     keyring_repository_password_del,
     keyring_repository_password_set,
+    normalize_url,
 )
 from .command import Command
 
@@ -163,7 +164,7 @@ To remove a repository (repo is a short alias for repositories):
                 return 0
 
             if len(values) == 1:
-                url = values[0]
+                url = normalize_url(values[0])
 
                 config_source.add_property(
                     "repositories.{}.url".format(m.group(1)), url
--- a/poetry/utils/helpers.py
+++ b/poetry/utils/helpers.py
@@ -152,3 +152,24 @@ def merge_dicts(d1, d2):
             merge_dicts(d1[k], d2[k])
         else:
             d1[k] = d2[k]
+
+
+def normalize_url(url):  # type: (str) -> str
+    """ Checks a URL for a permanent redirect and returns the redirect location
+        if there is one. If there is no redirect, it returns the same.
+    """
+    import requests
+
+    status_codes = requests.status_codes.codes
+    response = requests.head(url)
+    # The two status codes below are similar, but moved_permanently allows the
+    # agent to change a POST request to a GET (that's what requests does BTW
+    # when it is told to follow redirects).
+    # For poetry, as an agent, the only thing that matters is that both of these
+    # are permanent and cacheable.
+    if response.status_code in (
+        status_codes.moved_permanently,
+        status_codes.permanent_redirect,
+    ):
+        return response.headers["Location"]
+    return url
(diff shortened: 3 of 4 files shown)


Caligatio:
Wouldn't it make more sense to somehow hook URL usage versus config?  Many people don't use the `config` command and this also means you would potentially catch a redirect once during the lifetime of a project.

sztomi:
@Caligatio What do they use instead? My [original PR](https://github.com/sdispater/poetry/pull/1310) did follow the redirect at the time it was returned from the server, but there no really good options there. Either:

  * POST the data to the endpoint, and if there is a redirect, follow it manually and POST again (meaning double upload time)
  * Issue a HEAD request before the POST, and follow any redirects (that's what I implemented in my original PR) - this results in an extra request for everyone, for each upload.

sztomi:
(clicked the wrong button)

Caligatio:
I'm a bit torn whether the below ideas are over-engineering the situation or not.  However, I would *definitely* have a `follow_redirects=True` on the POST.  The current code depends on, essentially, one redirect-change happening in the lifetime of the repository otherwise we'll have another silent error.  Now, in terms of possible over-engineering ideas:

- Check the redirect on repository instantiation (happens per poetry run)?
- Check the redirect on every POST (i.e. do a HEAD everytime)?
- Check the redirect on the first usage of a repository?

To be clear here, I'm not a maintainer so my ideas are just ideas, not requests.

sztomi:
>  However, I would definitely have a follow_redirects=True on the POST

Unfortunately, that doesn't help. It sends a 301 redirect, which allows changing a POST to a GET, and that is exactly what requests does. The end result is that the silent upload failure still happens (because a GET request to the same endpoint succeeds).


> Check the redirect on repository instantiation (happens per poetry run)?

I don't know enough about the poetry internals to judge this, but if you can point out where it happens, I'm happy to investigate.

> Check the redirect on every POST (i.e. do a HEAD everytime)?

That's basically my original proposal, but @brycedrennan was concerned about the performance impact.

> Check the redirect on the first usage of a repository?

This is not substantially different from the current PR (unless I'm misunderstanding something).

> To be clear here, I'm not a maintainer so my ideas are just ideas, not requests

Ideas and discussion are very much appreciated, maintainer or not :)

Caligatio:
I confused source and upload repositories in my mind when I gave above options; sorry about that.  There are actually fewer places that are possibilities:

- When someone configs an upload repository (like you have)
  - Pros: It can write the updated value back to the config
  - Cons: The update only happens at the time of the original `config` call
- In the [Publisher class](/sdispater/poetry/blob/develop/poetry/masonry/publishing/publisher.py#L57)
  - Pros: The updated value gets updated per publish, no matter how many files
  - Cons:
    - I don't believe you can write the value back to the config
    - This involves making a new requests session (see previous comment about certs)
- In the [Uploader class](/sdispater/poetry/blob/develop/poetry/masonry/publishing/uploader.py#L99)
  - Pros: The updated value gets updated per publish, no matter how many files
  - Cons: I don't believe you can write the value back to the config

I don't think there is a perfect solution here.  In the `Uploader` seems best in terms of always working but then you don't get to perma-write the value back to the config.  Maybe throw a warning in Uploader if an updated URL is detected?  The idea was brought up in #1310.  I could maybe make the argument that a warning and forcing the user to update their own config is clearer than auto updating a config :)

sztomi:
> I don't believe you can write the value back to the config

Why not though? 

I think the *ideal* solution would be all of the following: 

  * Perform a normal POST
  * Check for a redirect
    * If there is a redirect:
      * Redo POST to the redirect location
      * Save redirect location as URL if the redirect is permanent
  * Check for a permanent redirect at config-save time

The last point is essentially this PR. I think both changes would improve the behavior independently and it is worth doing both.

Caligatio:
It appears that the way configs were handled changed drastically between master and develop.  It would have been entirely possible to write the values back in master as a handle was kept to each of the configs and they closely resembled files.  In develop, the concept of the config was abstracted away and the individual config files are merged into the abstracted config; there is no longer a handle to the underlying files.

brycedrennan:
@sztomi Thanks for your work on this and the first iteration of the PR.  

I'm a big fan of just providing a meaningful error message that will direct the user to fix their config.  This avoids the need for extra requests or automatic configuration updates. 

What do you think? Perhaps your original PR can be re-purposed to that?

sztomi:
I think this current PR would improve the situation independently, so I would keep this in addition to what you suggest. I suppose a clear error message is good though.

sdispater:

However, I don't the `config` command should have any logic in it, it's just a way to edit the configuration file, and especially not make a POST request to the configured url. So I am closing this.

The work should be to improve the `publish` command and catch upload errors and display them properly.

sztomi:
thanks @sdispater - do you think that `publish` should fix the configuration automatically when a permanent redirect is detected?

brycedrennan:
For my part I think it’s safer to just provide a meaningful error. The user can then fix the configuration themselves.

PR #1346, opened 2019-09-01
Fix ValueError when finding packages with invalid metadata

### LINKED ISSUE #1345: Searching for `pygame` results in ValueError

- **OS version and name**: Ubuntu 19.04
- **Poetry version**: 0.12.17
- **Link of a [Gist](https://gist.github.com/) with the contents of your pyproject.toml file**: n/a (the error is not related to pyproject.toml contents and happens even without one)

### Steps

1. Run `poetry search pygame`

The following exception raised:

```
[ParseVersionError]  
Unable to parse ".9".             
                                    
Exception trace:
 /home/tamas/.poetry/lib/poetry/_vendor/py3.6/cleo/application.py in run() at line 94
   status_code = self.do_run(input_, output_)
 /home/tamas/.poetry/lib/poetry/console/application.py in do_run() at line 88
   return super(Application, self).do_run(i, o)
 /home/tamas/.poetry/lib/poetry/_vendor/py3.6/cleo/application.py in do_run() at line 197
   status_code = command.run(input_, output_)
 /home/tamas/.poetry/lib/poetry/console/commands/command.py in run() at line 77
   return super(BaseCommand, self).run(i, o)
 /home/tamas/.poetry/lib/poetry/_vendor/py3.6/cleo/commands/base_command.py in run() at line 146
   status_code = self.execute(input_, output_)
 /home/tamas/.poetry/lib/poetry/_vendor/py3.6/cleo/commands/command.py in execute() at line 107
   return self.handle()
 /home/tamas/.poetry/lib/poetry/console/commands/search.py in handle() at line 20
   results = PyPiRepository().search(self.argument("tokens"), flags)
 /home/tamas/.poetry/lib/poetry/repositories/pypi_repository.py in search() at line 224
   result = Package(hit["name"], hit["version"], hit["version"])
 /home/tamas/.poetry/lib/poetry/packages/package.py in __init__() at line 40
   self._version = Version.parse(version)
 /home/tamas/.poetry/lib/poetry/semver/version.py in parse() at line 205
   raise ParseVersionError('Unable to parse "{}".'.format(text))
```

This issue was reported before, but closed by the author later: https://github.com/sdispater/poetry/issues/432

sztomi:
Some packages (like `pygame-music-grid`) have unparseable metadata. This resulted in a ValueError. pip ignores packages like these, so it's reasonable to do the same in poetry itself and only display the well-formed ones.

Fixes #1345.

# Pull Request Check List

  - n/a: this is a bugfix

### DIFF
--- a/poetry/repositories/pypi_repository.py
+++ b/poetry/repositories/pypi_repository.py
@@ -221,9 +221,17 @@ class PyPiRepository(Repository):
         hits = client.search(search, "or")
 
         for hit in hits:
-            result = Package(hit["name"], hit["version"], hit["version"])
-            result.description = to_str(hit["summary"])
-            results.append(result)
+            try:
+                result = Package(hit["name"], hit["version"], hit["version"])
+                result.description = to_str(hit["summary"])
+                results.append(result)
+            except ParseVersionError:
+                self._log(
+                    'Unable to parse version "{}" for the {} package, skipping'.format(
+                        hit["version"], hit["name"]
+                    ),
+                    level="debug",
+                )
 
         return results
 
--- a/tests/repositories/fixtures/pypi.org/json/pygame-music-grid.json
+++ b/tests/repositories/fixtures/pypi.org/json/pygame-music-grid.json
@@ -1,0 +1,81 @@
+{
+    "info":{
+        "author":"eric dexter",
+        "author_email":"irc.dexter@gmail.com",
+        "bugtrack_url":null,
+        "classifiers":[
+            "Topic :: Multimedia :: Sound/Audio :: Editors"
+        ],
+        "description":"a clickable grid for drum machines, piano rolls that is customizble from an init \r\nfile)or will be) that will include the script to be ran when a definable button is \r\nhit  written in pygame and tested with python 2.5",
+        "description_content_type":null,
+        "docs_url":null,
+        "download_url":"http://www.ziddu.com/download/5498230/pygamepianorollbeta.90.zip.html",
+        "downloads":{
+            "last_day":-1,
+            "last_month":-1,
+            "last_week":-1
+        },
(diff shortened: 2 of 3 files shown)


sztomi:
Hi @brycedrennan, I updated my branch with:

  - A more strict exception (`ParseVersionError`) that is actually raised instead of `ValueError`
  - A unit test for the issue along with a malformed json package metadata that would cause an issue without the fix. I actually discovered that a very similar logic is implemented in the `find_packages` method so I added that to the same test.

poetry/repositories/pypi_repository.py
```diff
@@ -221,9 +221,17 @@ def search(self, query, mode=0):
...
+            try:
+                result = Package(hit["name"], hit["version"], hit["version"])
+                result.description = to_str(hit["summary"])
+                results.append(result)
+            except ParseVersionError:
+                self._log(
+                    'Unable to parse version "{}" for the {} package, skipping'.format(
+                        hit["version"], hit["name"]
+                    ),
+                    level="debug",
```
bjoernpollex: I wonder if this should be a warning that's printed to the shell? This way it much more likely that someone will raise the issue with the upstream.

sztomi: I think there is a fairly large number of malformed packages on pypi that are easy to run into when you search for a fragment of a package name. This is almost the same logging that is done in `find_packages` in the same file.

sdispater:
Thanks!

poetry/repositories/pypi_repository.py
sztomi: Yep

randallpittman:
N.B. This does not fix having already-installed packages with un-parseable versions. It seems to only fix searches that result in unparseable versions. See #2167.

Issue #144, opened 2018-05-26
peotry version doesn't bump the value in __version__


ksamuel:
$ poetry version 0.1.1
    Bumping version from 0.1.0 to 0.1.1
    $ grep version pyproject.toml 
    version = "0.1.1"
    $ grep version */__init__.py
    src/__init__.py:__version__ = '0.1.0'
    $ poetry --version
    Poetry 0.9.0

I don't know if it's intended or not. A way to do that safely is to parse the root `__init__.py`, detect `__version__`, back it up, extract the ast, bump the version string and replace it, then extract the new ast and compare the result. If both ast are the same, except for the version, the file semantics have been preserved. Otherwise, rollback the change and display an error message stating we can't bump the version safely.

sdispater:
At the moment this is not possible no.

`poetry version` will only bump the `version` in `pyproject.toml`. This is not intended and that's definitely something I plan on adding.

Victor-Savu:
I really wish that the `version` field in the `.toml` ended up superseding the `__version__` string in `__init__.py`. This would reduce the burden of having to keep multiple places in the code in sync.

ksamuel:
Setuptoos has solved this problem for 2 years:

https://setuptools.readthedocs.io/en/latest/setuptools.html#configuring-setup-using-setup-cfg-files

You can put all metadata in setup.cfg instead of setup.py. For the version, you can specify:

    version = attr: src.__version__

Which will load the version from the package directly, this way you have it in one place only. It works as long as you has setuptools >= 30.3 (current version is 38.5.1) so basically any package with a setup.py can already benefit from that without using any additional tool.


 
Le 27/05/2018 à 07:28, Victor-Nicolae Savu a écrit :

Victor-Savu:
@ksamuel Yup, I use that as well :) It feels a bit backwards that it needs to be referenced in two places, but I suppose that is just my ignorance with respect to the benefits of having programmatic access to the version of a package. I can kind of imagine a scenario, but I have never seen a real-world example of where that was used (again, it's just my inexperience, I am not criticizing such a practice :) )

jgirardet:
you may use that 

```
import toml
from pathlib import Path

def get_version():
   path = Path(__file__).resolve().parents[1] / 'pyproject.toml'
   pyproject = toml.loads(open(str(path)).read())
   return pyproject['tool']['poetry']['version']

__version__ = get_version()
```

ksamuel:
[...]

PR #1354, opened 2019-09-05
Update version strings in project files. (#144)

JnyJny:
#144 

Added a dry-run option to version command and logic to
replace current version strings with next version in:
- pyproject.toml
- python source ( assignments & comparisons )
- markdown | restructured text | plain text

# Pull Request Check List

**Note**: If your Pull Request introduces a new feature or changes the current behavior, it should be based on the `develop` branch. If it's a bug fix or only a documentation update, it should be based on the `master` branch.

### DIFF
--- a/README.md
+++ b/README.md
@@ -264,7 +264,6 @@ So, in the end, you will still need to manage a few configuration files to prope
 
 ## Commands
 
-
 ### new
 
 This command will help you kickstart your new Python project by creating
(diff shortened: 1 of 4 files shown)


JnyJny:
I took a stab at writing tests for this new feature, but ran into trouble due to lack of familiarity with how to use/extend the fixtures already provided. I'll keep working the problem.

tarkatronic:
This is an interesting approach, but I could see a lot of potential for false positives, in places like a changelog, or even list of requirements. What if you require a specific version of a package, which matches the version of your local package?

I think a better approach might be explicit configuration of the locations of the version, such as how `bumpversion` works. The alternative would be expecting the user to perform a dry run before *every* version bump, which as a user, is simply not going to happen.

JnyJny:
> This is an interesting approach, but I could see a lot of potential for false positives, in places like a changelog, or even list of requirements. What if you require a specific version of a package, which matches the version of your local package?

Excellent point that I had not considered. 

> I think a better approach might be explicit configuration of the locations of the version, such as how `bumpversion` works. The alternative would be expecting the user to perform a dry run before _every_ version bump, which as a user, is simply not going to happen.

Excellent feedback, thank you. I'm not sure how opinionated poetry wants to be about how a package is structured so my goal was to avoid specifying where version data should go.

funkyfuture:
i think this functionality is out of `poetry`'s scope. but it might be fitting a coming plugin ecosystem.

JnyJny:
@funkyfuture I disagree, poetry already changes the project version number in pyproject.toml, it's a small step forward to take the burden off the developer and change it in the shipping (code/docs) part of the project.

sdispater:
Thanks for you contribution!

However, this is beyond the intended purpose of `version` which is to update the version in the `pyproject.toml` which is the only file under the control of Poetry.

I am closing this but that's a feature that can be implemented as a plugin when they are supported.

Issue #547, opened 2018-10-23
Interactive - Define dependencies invalid for well known packages


shepherdjay:
- [x] I am on the [latest](https://github.com/sdispater/poetry/releases/latest) Poetry version.
- [x] I have searched the [issues](https://github.com/sdispater/poetry/issues) of this repo and believe that this is not a duplicate.
- [x] If an exception occurs when executing a command, I executed it again in debug mode (`-vvv` option).

- Windows 10:
- 0.12.4:

## Issue
When running poetry init on a project and I get to the step asking me to declare my dependencies I attempted to enter "Flask" and it provided a list of 100 items, none of which was just "Flask" so I entered it again, and again, and it eventually threw:

```
"Value "flask" is invalid"
```

I verified the pypi package name. Other flask packages like Flask-WTF work fine.

PR #1364, opened 2019-09-07
Fix interactive dependencies behavior.

### LINKED ISSUE #528: Selecting packages interactively is no fun at all

## Issue

While executing `poetry init` i get asked to interactively define my dependencies. This feature is very welcome at this point, BUT the way it works is a little suboptimal. While `poetry add` never made the wrong decision for me so far,  in this interactive mode after wanting to install Django i get a list of 100 search results not even containing the actual Django package.
Poetry never tries to just guess and install the right packages like it does on `poetry add`. 

Why?
Adding Django (and any other package) just works, but interactively selecting is no fun!

Please make it fun again and just install the best guess!
Or to be more save at least display the best guess (as it would be resolved on add) and just let the user confirm and maybe select an alternative.

aklajnert:
This PR is meant to fix annoying behavior of the interactive dependencies filling during ``poetry init``.  
In current behavior, if you will try to add ``flake8``, ``flask`` or ``SQLAlchemy``, you will get a list of 100 packages, but none of them will be the one you're looking for. 

It seems it is a bug within the Pypi RPC API. Poetry is looking for the package by looking for any package with the given name in the package name **or** summary. The PyPi API bug seems to treat **or** as **and**. The API is already deprecated, and this functionality will still require changes.  

My change, however, has slightly worked around the faulty API. If the repository will have a package with exactly the same name as user wants, it will add it without listing packages with similar names. This will be more consistent, as the ``add`` method works the same. If the package with exactly the same name won't be found, it will fall back to the previous behavior.

This should resolve #528 and #547.

### DIFF
--- a/poetry/console/commands/init.py
+++ b/poetry/console/commands/init.py
@@ -3,7 +3,6 @@ from __future__ import unicode_literals
 
 import os
 import re
-
 from typing import Dict
 from typing import List
 from typing import Tuple
@@ -12,17 +11,14 @@ from typing import Union
 from cleo import option
 from tomlkit import inline_table
 
-from poetry.utils._compat import Path
 from poetry.utils._compat import OrderedDict
+from poetry.utils._compat import Path
 from poetry.utils._compat import urlparse
-from poetry.utils.helpers import temporary_directory
-
 from .command import Command
 from .env_command import EnvCommand
 
 
 class InitCommand(Command):
-
     name = "init"
     description = (
         "Creates a basic <comment>pyproject.toml</> file in the current directory."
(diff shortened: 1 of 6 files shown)


edisongustavo:
I've just run into this issue Today and I'm glad there is already a PR to fix it.

aklajnert:
Thanks @stephsamson, the merge is pretty nasty so I need some time to do it. I'll try to finish that this week, but I can't promise.

aklajnert:
@stephsamson - the conflicts are now resolved

aklajnert:
OK, I guess that this PR can be closed because of #1530

PR #950, opened 2019-03-11
Allow local version identifier in wheel filenames #756

kbakk:
Fixes cases where the version part of the wheel file name would become

`0.1b1_gitbranch.buildno.1`

instead of

`0.1b1+gitbranch.buildno.1`

fixes #756 

- [x] Added **tests** for changed code.
- ~[] Updated **documentation** for changed code.~ N/A

PR #1365, opened 2019-09-07
Allow uploading of wheels with local version identifier

hoefling:
This is a follow-up of #950 that allows publishing of wheels with local version identifier. Many thanks in advance for reviewing!

- [x] ~~Added **tests** for changed code.~~ Extended existing ones instead of adding dupes.
- [ ] ~~Updated **documentation** for changed code.~~ Same reason as in #950.

### DIFF
--- a/poetry/masonry/builders/wheel.py
+++ b/poetry/masonry/builders/wheel.py
@@ -19,7 +19,7 @@ from poetry.__version__ import __version__
 from poetry.semver import parse_constraint
 from poetry.utils._compat import decode
 
-from ..utils.helpers import normalize_file_permissions
+from ..utils.helpers import normalize_file_permissions, escape_name, escape_version
 from ..utils.package_include import PackageInclude
 from ..utils.tags import get_abbr_impl
 from ..utils.tags import get_abi_tag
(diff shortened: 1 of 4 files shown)


kbakk:
Good catch. I wasn't using the upload functionality, so I missed it. Seeing how I missed it, it could happen again in the future. Perhaps create a static function on WheelBuilder, which handles filename? And use this in both places (poetry/masonry/builders/wheel.py and poetry/masonry/publishing/uploader.py).

hoefling:
@kbakk I guess this went unnoticed because both `re.sub`s in `WheelBuilder` and `Uploader` are code dupes, so refactoring them is indeed a good solution, but I'm unsure where to place the code. Are you suggesting a method in `WheelBuilder` and import it in the `Uploader`? Or maybe add another pair of functions like `escape_name`/`escape_version` to `poetry.utils.helpers`? Or maybe `escaped_name`/`escaped_version` properties to `poetry.packages.package.Package`?

```python
class Package(object):
    ...
    @property
    def escaped_name(self):
        return re.sub(r"[^\w\d.]+", "_", self.pretty_name, flags=re.UNICODE)
```

I'm not sure where to place both `re.sub`s, so any decision is welcome from my side.

kbakk:
All seems to be viable options. In addition, there's `poetry.masonry.utils.helpers`.

hoefling:
@kbakk great suggestion, I like that! Added the necessary changes.

hoefling:
@sdispater sure thing, done!

sdispater:
Thanks!

PR #1367, opened 2019-09-10
#1350 - Improvements to handling of License file(s)

### LINKED ISSUE #866: Generated setup.py should use setuptools.setup instead of distutils.core.setup

- **OS version and name**: Gentoo Linux
- **Poetry version**: 0.12.11

The command `poetry build` generates a `setup.py` file based on the contents of `pyproject.toml`. The setup file imports `distutils.core.setup` as the setup function, but passes build arguments that are only available in `setuptools.setup`, such as `entry_points`. When the package is installed via pip, this is not an issue, because pip still uses `setuptools.setup`, regardless of what is imported (see the [distutils docs](https://docs.python.org/3/library/distutils.html)). However, other package managers may run into issues, because they use distutils. My distribution's package manager (Portage) generates warnings:
```
usr/lib64/python3.6/distutils/dist.py:261: UserWarning: Unknown distribution option: 'install_requires'
  warnings.warn(msg)
/usr/lib64/python3.6/distutils/dist.py:261: UserWarning: Unknown distribution option: 'entry_points'
  warnings.warn(msg)
/usr/lib64/python3.6/distutils/dist.py:261: UserWarning: Unknown distribution option: 'python_requires'
  warnings.warn(msg)
```
In this case, no scripts will be installed, because the `entry_points` argument is not applicable for distutils.

Since `setup.py` uses features that are explicitly part of setuptools, but not distutils, I suggest that `setuptools.setup` should be imported. Would you accept a PR for this?

### LINKED ISSUE #1350: License Files listing in Metadata

## Feature Request

Poetry does not currently specify which files are included as license files in the appropraite Metadata field for license file names. The appropriate section of code:

https://github.com/sdispater/poetry/blob/2ed53be4f7d9c86087e4a0b82984b9a718211a9b/poetry/masonry/builders/builder.py#L158-L204 

Just needs to be extended so that it includes the gathered license files in the package metadata with the appropriate formatting. I suspect it will also need https://github.com/sdispater/poetry/blob/master/poetry/masonry/builders/builder.py#L128-L136 to be modified so that it emits a list of license specific file names that can be used when generating the license metadata. As otherwise it will not be possible to tell which of the added files are actually license files. It may be worth including a change to allow manually specifying the license files by a package maintainer while making these changes since it will probably require changes to the license file gathering code anyway.

techdragon:
- Fixes #1350
- Added support for specifying a set of license files as part of the project config.
- Moved the license file config into the package configuration. (`Factory.create_poetry()`) this matches how similar configurations such as “readme” are configured.
- Updated the builders to use the project level license file list instead of searching for their own (previously different) lists of license files.
- Moved the existing auto-discovery behaviours to become the default case during `create_poetry()` if nothing specific was configured.
- Added tests for the new features
- Documented the new configuration field.
- Added the “license-files” field to the poetry-schema.json
- Updated pre-commit config to use the correct black source repo, and set the default language for better consistency with any other pre-commit checks that will be added in the future.
- Fix #866 by modifing the base template used by `SdistBuilder`. It now uses `setuptools`, since as documented by @seifertm in issue #866 `pip` uses `setuptools` anyway and this clears the way to immediately supporting https://github.com/pypa/setuptools/pull/1767 once this is merged and released from `setuptools`. Until `setuptools` updates to support multiple license files, this will result in a single log line in the debug output from `setuptools` so this is ok to merge before the change to `setuptools` is merged.

# Pull Request Check List


poetry/masonry/builders/builder.py
```diff
@@ -146,8 +146,9 @@ def find_files_to_add(self, exclude_build=True):  # type: (bool) -> list
         )
         to_add.append(Path("pyproject.toml"))
 
-        # If a license file exists, add it
-        for license_file in self._path.glob("LICENSE*"):
+        # If any license files exist, add them.
+        for license_filename in list(self._package.license_files):
```
bjoernpollex: Is it necessary to wrap this in a `list` call here?

techdragon: Just checked and no it looks like I don't. I guess I forgot that sets are itterable when I was writing that bit. Fix incoming.

poetry/masonry/builders/sdist.py
```diff
@@ -137,6 +137,14 @@ def build_setup(self):  # type: () -> bytes
             else:
                 pass
 
+        license_file_list = list(self._package.license_files)
+        if len(license_file_list) == 1:
```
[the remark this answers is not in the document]
techdragon: I'm kind of hedging my bets here. The current situation re: distutils/setuptools is that it expects a single license_file, however there is the open PR on setuptools to bring it into line with the currently emerging standard of allowing multiple license files like the wheel package metadata does.

I'd prefer to drop the special case for 1 too, but I wasn't sure how far I could push that kind of behavioural change.

poetry/poetry.py
```diff
@@ -129,6 +129,16 @@ def create(cls, cwd):  # type: (Path) -> Poetry
         if "readme" in local_config:
             package.readme = Path(poetry_file.parent) / local_config["readme"]
 
+        # Add any manually defined or automatically discovered license files.
+        if "license-files" in local_config:
+            for file in local_config["license-files"]:
```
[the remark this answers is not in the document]
techdragon: It's so easy to forget these things sometimes. Thanks, I'll fix that.

techdragon:
The rebase to keep on top of #1349 and the other changes was a little tricky, but I feel like the changes still fit pretty well and it was worth it since new tests found an easily fixed mistake with how I was using Path objects in `sdist.py` :smile: 

This PR is once again ready for a merge! 🎉

techdragon:
T.L.D.R. Version = Yes `poetry` can use `include` for license files, but it doesn't know which files it included in the package are the license files for the package. This change makes it so `poetry` knows which included files are license files so the package builder classes can use that info.

@sdispater You are right about being able to achieve some of this just by using `include` but the point of the change is to better handle the metadata that is represented by the new field. 

At the moment you can include a license file using `include`, but this just includes them the same as any other file and does not associate them as special in any way. 

There is currently an evolving discussion about the next set of changes to the package metadata format. https://discuss.python.org/t/improving-license-clarity-with-better-package-metadata/2154 and while the discussion isn't finalised yet the idea of adding a specific field to the package metadata to specify which included files are license related seems to have no one arguing against it so it is likely to be part of the next revision. 

The changes in this PR, put `poetry` in a position to easily adapt to this potential future revision by keeping track of which included files were license related so that this can be passed on in the package metadata.

Case in point, I mentioned the open PR on `setuptools`, which when merged will make it possible to pass a list into the `license_files` keyword argument of the `setup()` function. I've added this functionality into the sdist builder class since setuptools is pretty tolerant of keyword arguments it doesn't know how to handle, as outlined in the linked PR.

docs/docs/pyproject.md
```diff
@@ -40,6 +40,12 @@ The recommended notation for the most common licenses is (alphabetical):
 Optional, but it is highly recommended to supply this.
 More identifiers are listed at the [SPDX Open Source License Registry](https://www.spdx.org/licenses/).
 
+## license-files
+
+The license file(s) belonging to this package. **optional**
+
+Poetry will automatically include files matching `LICENSE*` and `COPYING*` but if you have licenses files that do not fit these patterns, you can configure `poetry` to include one or more files as the license files. These paths should be relative to the project root directory, where the `pyproject.toml` lives.
```
funkyfuture: a comma is missing before `but` and the line is quiet long to be read in a terminal.

poetry/factory.py
```diff
@@ -81,6 +81,16 @@ def create_poetry(
         if "readme" in local_config:
             package.readme = Path(poetry_file.parent) / local_config["readme"]
 
+        # Add any manually defined or automatically discovered license files.
+        if "license-files" in local_config:
+            for license_file in local_config["license-files"]:
+                package.license_files.add(Path(poetry_file.parent) / license_file)
+        else:
+            for license_file in poetry_file.parent.glob("LICENSE*"):
```
funkyfuture: i'm not sure whether the stdlib's globbing implementation supports this, but `{COPYING*,LICENSE*}` should be enough.

poetry/masonry/builders/sdist.py
```diff
@@ -137,6 +137,11 @@ def build_setup(self):  # type: () -> bytes
             else:
                 pass
 
+        license_files = list(map(str, self._package.license_files))
```
funkyfuture: why the `map` when the schema defines an item to be a string?

poetry/masonry/builders/wheel.py
```diff
@@ -181,9 +181,11 @@ def _write_metadata(self, wheel):
             with self._write_to_zip(wheel, self.dist_info + "/entry_points.txt") as f:
                 self._write_entry_points(f)
 
-        for base in ("COPYING", "LICENSE"):
-            for path in sorted(self._path.glob(base + "*")):
-                self._add_file(wheel, path, "%s/%s" % (self.dist_info, path.name))
+        for license_file_name in self._package.license_files:
+            license_file = self._path / license_file_name
+            self._add_file(
+                wheel, license_file, "%s/%s" % (self.dist_info, license_file.name)
```
funkyfuture: `license_file.name` is `license_file_name` and if the latter was used here, the `license_file` symbol wouldn't have to be allocated.

poetry/factory.py
techdragon: The standard library globbing doesnt support multiple patterns unfortunately. https://docs.python.org/3/library/glob.html#glob.glob

I've just tried a few variations but none of them are particularly nice looking, and `black` has a tendency to reformat them over more lines than the code they would replace.

```python
            for glob_pattern in ("LICENSE*", "COPYING*"):
                for license_file in poetry_file.parent.glob(glob_pattern):
                    package.license_files.add(
                        license_file.relative_to(poetry_file.parent)
                    )
```
The code above was the nicest result and I'm happy to change it but it is at the end of the day... really just a fancier way of doing the same thing as the original code below.
```python
            for license_file in poetry_file.parent.glob("LICENSE*"):
                package.license_files.add(license_file.relative_to(poetry_file.parent))
            for license_file in poetry_file.parent.glob("COPYING*"):
                package.license_files.add(license_file.relative_to(poetry_file.parent))
```

poetry/masonry/builders/sdist.py
techdragon: The map is to transform them into strings. Since `self._package.license_files` is a list of `Path` objects they were getting written out to the final `setup.py` file as `repr()` style objects `<Path ... >` which was't very useful for setuptools. :-)

docs/docs/pyproject.md
techdragon: I've tidied it up a little and fixed the missing comma. Next push will include the changes.

poetry/masonry/builders/wheel.py
techdragon: While I'll admit the names would imply that, the step is needed. Otherwise `_add_file()` will raise `FileNotFoundError`.  So I've renamed them in the next push to be more clear what their purpose is.

poetry/masonry/builders/builder.py
```diff
@@ -146,8 +146,9 @@ def find_files_to_add(self, exclude_build=True):  # type: (bool) -> list
         )
         to_add.append(Path("pyproject.toml"))
 
-        # If a license file exists, add it
-        for license_file in self._path.glob("LICENSE*"):
```
sdispater: I think we should keep this line if there are no `license_files` specified in the `pyproject.toml` file, so that people relying on the current behavior are not surprised and published packages are not suddenly missing license files.

sdispater:
As an optional feature I think this is a good idea. It will likely not make it to the `1.0.0` version since it's now pretty much feature-frozen but it will be integrated in a future minor version.

poetry/masonry/builders/builder.py
techdragon: @sdispater - I agree this should remain the "default path" and I made sure to retain this behaviour as part of my changes. This functionality was consolidated into the package setup stuff (which is now in `factory.py`) and is currently being discussed above: https://github.com/sdispater/poetry/pull/1367#discussion_r328299477 & https://github.com/sdispater/poetry/pull/1367#discussion_r328491317

The specific lines for automatically adding `LICENSE*` and `COPYING*` is in `factory.py` here:
- https://github.com/techdragon/poetry/blob/issue-1350-license-metadata-improvements/poetry/factory.py#L84-L92

And I also added a test case for this existing/default behaviour when no `license_files` have been specified. 😃https://github.com/techdragon/poetry/blob/issue-1350-license-metadata-improvements/tests/masonry/builders/test_complete.py#L394-L416

poetry/factory.py
techdragon: @funkyfuture Any more thoughts on this?

poetry/masonry/builders/builder.py
techdragon: @sdispater Since the functionality is retained as desired, but powered by new code in a different place, if you don't mind, id like to mark this as resolved. 😄

techdragon:
Back for a new year and I see that https://github.com/pypa/setuptools/pull/1767 has been merged and poetry 1.0.0 has shipped! So I'm going to try and find a little time this week to update this PR.

Secrus:
@techdragon are you still interested in bringing this to Poetry? If so, please merge master and change the target branch to master. This will allow for review against the current codebase.

PR #1400, opened 2019-09-23
If target package is lower version, call it out as a downgrade

### LINKED ISSUE #1394: Downgrading says it's updating

When I downgrade version Poetry says I'm "updating". Maybe "downdating" or even "downgrading" would be more appropriate.

```
  - Updating django (2.2.5 -> 2.1.12)
```

tarkatronic:
Fixes #1394 

As per the example in the description of that issue: 
```sh
$ poetry add "Django==2.1.12"

Updating dependencies
Resolving dependencies... (8.0s)

Writing lock file


Package operations: 0 installs, 1 update, 1 removal

  - Downgrading django (2.2.5 -> 2.1.12)
  - Removing sqlparse (0.3.0)
$
```

This is a very simplistic implementation of the fix, and very superficial. Things to note:
* There are no tests; I couldn't find any tests encompassing the strings which are output
* The `Package operations` still list an update
* I did not modify the `Update` operation, so its str and repr methods still say `Update`/`Updating`.

I considered taking a bit more of a holistic approach to this, but as my brain went further down that path, I realized it would likely result in a new operation type and possibly even a new command.

### DIFF
--- a/poetry/installation/installer.py
+++ b/poetry/installation/installer.py
@@ -344,7 +344,8 @@ class Installer:
 
         if self._execute_operations or self.is_dry_run():
             self._io.write_line(
-                "  - Updating <info>{}</> (<comment>{}</> -> <comment>{}</>)".format(
+                "  - {} <info>{}</> (<comment>{}</> -> <comment>{}</>)".format(
+                    "Updating" if source.version < target.version else "Downgrading",
                     target.pretty_name,
                     source.full_pretty_version,
                     target.full_pretty_version,
--- a/tests/console/commands/test_add.py
+++ b/tests/console/commands/test_add.py
@@ -716,6 +716,36 @@ Package operations: 1 install, 0 updates, 0 removals
     assert content["dependencies"]["foo"] == "^1.1.2"
 
 
+def test_add_existing_package_with_lower_exact_version_displays_as_downgrade(
+    app, repo, installer, installed
+):
+    content = app.poetry.file.read()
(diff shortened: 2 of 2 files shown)


tarkatronic:
@stephsamson Thanks for the pointer! I hadn't thought of looking at those tests. I've added one fairly specific test to verify this output change.

Do you have any thoughts on the other concerns raised in my PR description above?

stephsamson:
@tarkatronic sorry for the late response and thanks for being patient with us!

To address your other concerns, I think it would make sense to have an abstract base class `Operation` -- this could for e.g. be refactoring the existing `Update` and from there making an `Upgrade` and `Downgrade` `Operation`.

neersighted:
Hi @tarkatronic! It's been a while. Any interest in continuing this? I glanced at the current implementation and it's been cleaned up quite a bit. Feel free to close this if you're no longer interested.

tarkatronic:
> Hi @tarkatronic! It's been a while. Any interest in continuing this? I glanced at the current implementation and it's been cleaned up quite a bit. Feel free to close this if you're no longer interested.

Wow it has been a while; I totally forgot about this PR! Let me get a look at what's changed around and I'll see if I can't modernize this change.

Secrus:
@tarkatronic it's been a while again. Are you still interested in bringing this to Poetry? If so, please merge the master branch, change the target branch to master, and resolve potential issues.

PR #1408, opened 2019-09-25
Fixes error messages that point to list items

funkyfuture:
poetry printed a misleading error message when the error stemmed from an item in a field. i didn't add a test because i wasn't able to spot an error messaging related that would exist.

### DIFF
--- a/poetry/json/__init__.py
+++ b/poetry/json/__init__.py
@@ -31,7 +31,9 @@ def validate_object(obj, schema_name):  # type: (dict, str) -> List[str]
     for error in validation_errors:
         message = error.message
         if error.path:
-            message = "[{}] {}".format(".".join(error.path), message)
+            message = "[{}] {}".format(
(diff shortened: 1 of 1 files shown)


poetry/json/__init__.py
```diff
@@ -31,7 +31,9 @@ def validate_object(obj, schema_name):  # type: (dict, str) -> List[str]
     for error in validation_errors:
         message = error.message
         if error.path:
-            message = "[{}] {}".format(".".join(error.path), message)
+            message = "[{}] {}".format(
+                ".".join(str(x) for x in error.absolute_path), message
```
sdispater: Is there any reason why you use `absolute_path` instead of `path`?

funkyfuture: as far as i remember because both are an alias and the name i use is clearer what it represents.

sdispater:
Thanks!

Issue #1417, opened 2019-09-28
Various issues when package name differs from directory name.


KelSolaar:
- [x] I am on the [latest](https://github.com/sdispater/poetry/releases/latest) Poetry version.
- [x] I have searched the [issues](https://github.com/sdispater/poetry/issues) of this repo and believe that this is not a duplicate.
- [ ] If an exception occurs when executing a command, I executed it again in debug mode (`-vvv` option).

- **OS version and name**: maOS 10.14.6
- **Poetry version**: 1.0.0b1
- **Link of a [Gist](https://gist.github.com/) with the contents of your pyproject.toml file**: https://gist.github.com/KelSolaar/c2c853ca329899534494e2ddf8a1aa96

## Issue
This relates to https://github.com/sdispater/poetry/issues/705 but I'm not able to solve all my issues without massaging a few things. The PyPi package name of my project differs from the actual python package name:

- **colour-science** for the [PyPi package](https://pypi.org/project/colour-science/).
- **colour** for the internal python package and same for the [Github repository](https://github.com/colour-science/colour).

My `pyproject.toml` file was originally working fine when using **colour** for the `name` attribute:

```toml
[tool.poetry]
name = "colour"
```

When trying to build with `poetry build`, I noticed that it was generating a **colour.0.3.13.tar.gz** archive that would not work for PyPi, thus I changed the TOML file as follows:

```toml
[tool.poetry]
name = "colour-science"
packages = [
    { include="colour", from="." },
]
```

I was able to build, but now, after cloning the repo into a default **colour** directory, issuing `poetry install` creates a virtual environment with the **colour-science** prefix however the `poetry env info -p` command does not find the virtual environment because I'm assuming it is looking for an environment with the **colour** prefix instead.

So there is a consistency issue between how the name of the candidate virtual environment is generated with the `poetry install` command and the `poetry env` one.

It also seems like it would be useful to have an attribute to handle the PyPi distribution name explicitly.

Cheers,

Thomas

KelSolaar:
So looking at the 1.0.0b1 code for getting the environment: 
https://github.com/sdispater/poetry/blob/1caf20a291d3a494ab815a9b2746d2f152abb893/poetry/console/commands/env/info.py#L17 It looks like it is indeed relying only on the directory name to find the environment and not the *name* attribute value from the `pyproject.toml` file.

peter-schmidbauer:
[...]

PR #1418, opened 2019-09-28
Implement support for distribution name.

KelSolaar:
This PR implements support for distribution name.

As highlighted in #1417, it might be required to adopt a different distribution name on PyPi than that of the package name or git repository. In our case, our package and Github repository are named **colour**,  but because **colour** was taken on PyPi, we picked **colour-science** for the distribution.

While it is possible to use the `packages` property to some degree:

```toml
[tool.poetry]
name = "colour-science"
packages = [
    { include="colour", from="." },
]
```

this creates a lot of unneeded extra work to manage the virtual environments and it also highlighted some inconsistency between the various commands.

This PR updates the schema and adds a new `distribution` property which gives granular control over how the **sdist** and **wheel** are generated.

The property name, i.e. _distribution_, was chosen in accordance with [Setup Tools documentation](https://packaging.python.org/tutorials/packaging-projects/):

> setup() takes several arguments. This example package uses a relatively minimal set:
> 
> name is the distribution name of your package. This can be any name as long as only contains letters, numbers, _ , and -. It also must not already taken on pypi.org. Be sure to update this with your username, as this ensures you won’t try to upload a package with the same name as one which already exists when you upload the package.
> 

# Pull Request Check List


poetry/masonry/builders/sdist.py
```diff
@@ -55,16 +55,19 @@ def build(self, target_dir=None):  # type: (Path) -> Path
         if not target_dir.exists():
             target_dir.mkdir(parents=True)
 
-        target = target_dir / "{}-{}.tar.gz".format(
-            self._package.pretty_name, self._meta.version
+        basename = (
+            self._package.pretty_name
+            if self._package.distribution is None
+            else self._package.distribution
         )
```
funkyfuture: you can shorten this expression to: `basename = self._package.distribution or self._package.pretty_name`. also apllies for the `wheels` module in this package.

sdispater:
Thanks for your contribution!

However, like you said, the `packages` property was designed in part to cover this use case so I don't think adding a new `property` is necessary, especially since I think its best to keep the Poetry section of the `pyproject.toml` file as simple as possible.

KelSolaar:
All good! I still think there is important value to it because it solves elegantly the problem of the mismatching distribution vs package name, and especially the issues mentioned in https://github.com/sdispater/poetry/issues/1417.

What we are doing which is **lame** but possible because we use `invoke` for a lot of things is that we are doctoring the `pyproject.toml` file before building, effectively changing the header from

```toml
[tool.poetry]
name = "colour"
```

to

```toml
[tool.poetry]
name = "colour-science"
packages = [
    { include="colour", from="." },
]
```

building and then reverting the file: https://github.com/colour-science/colour/blob/c0b6522ae347cbdd58080161093f447634d68d87/tasks.py#L349

Once issues in https://github.com/sdispater/poetry/issues/1417 are fixed maybe this PR is not as much needed.

PR #1428, opened 2019-10-01
Improve run command

maksbotan:
Currently "poetry run" command uses subprocess.call for starting
programs and scripts. This approach has a problem --- signals like
Ctrl-C are caught by poetry process itself and not by the process the
user has run.

This can be observed this way: run "poetry run python" and press Ctrl-C.
Normally Python interpreter will catch the signal and print
"KeyboardInterrupt" exception, continuing execution normally. But when
run by poetry, it exits immediately. This behavior is rather annoying,
as one has to restart interpreter after accidentially doing Ctrl-C.
Moreover, after python exits, it leaves terminal in a broken state,
requiring the user to use "reset" or similar command.

This commit fixes this behavior by using "os.execvp" instead of
subprocess. This replaces poetry process directly with target process,
eliminating an intermediate process.

This approach is used in other products like Pipenv and Haskell's build
system stack. See these links for examples and more info:

- https://github.com/pypa/pipenv/blob/master/pipenv/core.py#L2462
- https://github.com/commercialhaskell/stack/issues/527
- https://hackage.haskell.org/package/rio-0.1.12.0/docs/RIO-Process.html#v:exec

# Pull Request Check List

# Notes

- I'm unsure if it makes sense to add tests for this functionality. Right now there are no tests for `run` command in poetry.
- Windows runs the old way, this issue seems not to be present there.
- Also, it seems that nothing has to be added to the docs, as this a simple fix.

### DIFF
--- a/poetry/console/commands/run.py
+++ b/poetry/console/commands/run.py
@@ -27,18 +27,16 @@ class RunCommand(EnvCommand):
 
         src_in_sys_path = "sys.path.append('src'); " if self._module.is_in_src() else ""
 
-        cmd = ["python", "-c"]
+        cmd = ["-c"]
 
         cmd += [
-            '"import sys; '
+            "import sys; "
             "from importlib import import_module; "
             "sys.argv = {!r}; {}"
-            "import_module('{}').{}()\"".format(
-                args, src_in_sys_path, module, callable_
-            )
+            "import_module('{}').{}()".format(args, src_in_sys_path, module, callable_)
         ]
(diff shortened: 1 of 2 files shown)


sdispater:

However, this is already covered by #1236 so I am closing this one to avoid duplicates.

maksbotan:
Oh, great!
Sorry for that. I'm sure I tried searching github before making my pr, but did not find that one for some reason.

Issue #1401, opened 2019-09-24
poetry show package throws error


stinovlas:
- [x] I have searched the [issues](https://github.com/sdispater/poetry/issues) of this repo and believe that this is not a duplicate.
- [x] If an exception occurs when executing a command, I executed it again in debug mode (`-vvv` option).

- **OS version and name**: Ubuntu 19.04 Disco Dingo
- **Poetry version**: `1.0.0b1`

## Issue
I added `django` as a dependency to my project. When I call `poetry show django -vvv`, I get following error:

```
Using virtualenv: /home/jmusilek/.envs/test

[TypeError]
render() missing 1 required positional argument: 'io'

Traceback (most recent call last):
  File "/home/jmusilek/.poetry/lib/poetry/_vendor/py3.7/clikit/console_application.py", line 132, in run
    status_code = command.handle(parsed_args, io)
  File "/home/jmusilek/.poetry/lib/poetry/_vendor/py3.7/clikit/api/command/command.py", line 119, in handle
    status_code = self._do_handle(args, io)
  File "/home/jmusilek/.poetry/lib/poetry/_vendor/py3.7/clikit/api/command/command.py", line 167, in _do_handle
    return getattr(handler, handler_method)(args, io, self)
  File "/home/jmusilek/.poetry/lib/poetry/_vendor/py3.7/cleo/commands/command.py", line 92, in wrap_handle
    return self.handle()
  File "/home/jmusilek/.poetry/lib/poetry/console/commands/show.py", line 89, in handle
    table.render()
```

The same command is working fine with `poetry@0.12.17`, I'm getting the error only in `poetry@1.0.0b1`.

PR #1414, opened 2019-09-27
Fix for "render takes 2 arguments" error.

JustinTArthur:
Fixes output like:
```
poetry show -l pytest

[TypeError]
render() takes at least 2 arguments (1 given)
```

Original line may have been for a future version of cleo.

PR #1430, opened 2019-10-01
Fix console output render for show command

abn:
Simple change to support clickit interface change.

Resolves: #1401


dimbleby:
duplicates #1414

PR #1439, opened 2019-10-06
Fix adding requests with the security extra

Jamim:
Hello @sdispater,

Using `poetry` from `master`/`develop` branches or any recent alpha/beta release, you can't add the `requests` package with the `security` extra.
The `requests` package requires the `idna` twice (as a main dependency and as an optional one) and this case wasn't handled properly in `Locker._dump_package` function.
These changes make the `_dump_package` ensures that all elements of the constraints list have the same type in order to make them renderable as TOML.

Thank you for the great tool!

### Pull Request Check List


Jamim:
Hello Sébastien,

I've just added a corresponding test for the locker.
However, I believe that it's worth keeping an existing test since it checks the full flow that was broken and doesn't slow down CI to much.
Please let me know if you still want me to remove it.

Thanks!

tests/packages/test_locker.py
```diff
@@ -177,5 +177,47 @@ def test_lock_file_should_not_have_mixed_types(locker, root):
     with locker.lock.open(encoding="utf-8") as f:
         content = f.read()
 
-    print(content)
+    assert expected == content
+
+
+def test_lock_extra_duplicates_main(locker, root):
```
sdispater: I think this test duplicates the one above, doesn't it?

Jamim: Now we have three related tests:
 - `test_lock_file_should_not_have_mixed_types`
 - `test_lock_extra_duplicates_main`
 - `test_add_extra_duplicates_main`

All the tests are similar in general, so please let me know which ones should be removed/modified.
Thanks!

sdispater: I think for the sake of simplicity we can keep `test_lock_file_should_not_have_mixed_types` only.

Jamim: I've removed two other tests.

Issue #1384, opened 2019-09-18
Files listed in exclude are included when building wheel only


finswimmer:
- [x] I am on the [latest](https://github.com/sdispater/poetry/releases/latest) Poetry version.
- [x] I have searched the [issues](https://github.com/sdispater/poetry/issues) of this repo and believe that this is not a duplicate.
- [x] If an exception occurs when executing a command, I executed it again in debug mode (`-vvv` option).

- **OS version and name**:Ubuntu 18.04
- **Poetry version**: Poetry 0.12.17
- **Link of a [Gist](https://gist.github.com/finswimmer/d5e9cd7d6a0668313d4ec757baf60fea) with the contents of your pyproject.toml file**: 

## Issue

I have defined a file to be excluded from the build in my `pyproject.toml`. When running `poetry build` this file is excluded in the sdist and wheel. When running `poetry build -f wheel` the file is included, running `poetry build -f sdist` the file is excluded.

Way to reproduce:

```
$ poetry new issue
$ cd issue
$ poetry install
$ touch issue/exclude_me.py
```

Add `issue/exclude_me.py` to the `pyproject.toml`

```
[tool.poetry]
name = "issue"
version = "0.1.0"
description = ""
authors = ["Fin Swimmer"]
exclude = ["issue/exclude_me.py"]
```

Run `poetry build -vvv`:

```
Using virtualenv: /home/finswimmer/.cache/pypoetry/virtualenvs/issue-py3.6
Building issue (0.1.0)
 - Building sdist
 - Adding: issue/__init__.py
 - Adding: pyproject.toml
 - Built issue-0.1.0.tar.gz

 - Building wheel
 - Adding: /tmp/tmpvo4lwtdk/issue-0.1.0/issue/__init__.py
 - Built issue-0.1.0-py3-none-any.whl
```

Run `poetry build -vvv -f wheel`:

```
Using virtualenv: /home/finswimmer/.cache/pypoetry/virtualenvs/issue-py3.6
Building issue (0.1.0)
 - Building wheel
 - Adding: /home/finswimmer/issue/issue/__init__.py
 - Adding: /home/finswimmer/issue/issue/exclude_me.py
 - Built issue-0.1.0-py3-none-any.whl
```

Run `poetry build -vvv -f sdist`:

```
Using virtualenv: /home/finswimer/.cache/pypoetry/virtualenvs/issue-py3.6
Building issue (0.1.0)
 - Building sdist
 - Adding: issue/__init__.py
[...]

PR #1459, opened 2019-10-11
* check if relative filename is in excluded file list

finswimmer:
This is a fix for #1384 .

The method `find_excluded_files` in `wheel.py` unnecessarily overwrite the original method inherited from `Builder` by returning just an empty set. I removed this method.

Comparison, if the file is contained in  `excluded`, was done with the absolute filename while the filenames in `excluded` are relative filepath name. Also one have to take care about the types. While `rel_file` is of type `Path` the names in the `excluded` set are `str`.

### DIFF
--- a/poetry/masonry/builders/wheel.py
+++ b/poetry/masonry/builders/wheel.py
@@ -153,7 +153,7 @@ class WheelBuilder(Builder):
                 else:
                     rel_file = file.relative_to(self._path)
 
-                if file in excluded:
+                if str(rel_file) in excluded:
                     continue
 
                 if file.suffix == ".pyc":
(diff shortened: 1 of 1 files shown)


finswimmer:
Works on Linux and Mac, but not on Windows? Ok, needs more investigation. Or do you have any guess about it?

poetry/masonry/builders/wheel.py
```diff
@@ -153,7 +153,7 @@ def _copy_module(self, wheel):
                 else:
                     rel_file = file.relative_to(self._path)
 
-                if file in excluded:
+                if str(rel_file) in excluded:
```
sdispater:
```suggestion
                if rel_file.as_posix() in excluded:
```

finswimmer:
Argh, of course the windows file path delimiters make trouble. Thanks for pointing to the correct direction.

Everything fine now?

sdispater:
Thanks!

finswimmer:
You're welcome. Thanks for this wonderful project!

I closed the corresponding issue #1384 .

finswimmer:
Hello again @sdispater ,

should we/can we make this change in the current stable version as well?

fin swimmer

PR #1461, opened 2019-10-12
provider:search_for_directory - set package.root_dir

khfayzullaev:
-- `package` from provider.py:search_for_directory
   is missing root_dir value. This fixes the problem.
-- fix https://github.com/sdispater/poetry/issues/1455

I hope I am not missing anything. The `root_dir` is eventually used here https://github.com/sdispater/poetry/blob/21ed84ed9b422a049c3e2d6b639d9de53ab60dc6/poetry/installation/pip_installer.py#L128 to build the path to the local package.

### DIFF
--- a/poetry/puzzle/provider.py
+++ b/poetry/puzzle/provider.py
@@ -277,6 +277,9 @@ class Provider:
 
         package.source_url = dependency.path.as_posix()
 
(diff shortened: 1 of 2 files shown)


stephsamson:
Thanks for your contribution! 🍡 💫

PR #1463, opened 2019-10-12
Fix `extras` hyperlink

m1so:
# Pull Request Check List

### DIFF
--- a/docs/docs/cli.md
+++ b/docs/docs/cli.md
@@ -117,7 +117,7 @@ poetry install --no-dev
 ```
 
 You can also specify the extras you want installed
-by passing the `--E|--extras` option (See [Extras](#extras) for more info)
+by passing the `--E|--extras` option (See [Extras](/docs/pyproject/#extras) for more info)
 
 ```bash
 poetry install --extras "mysql pgsql"


m1so:
Hey @stephsamson, not sure what you mean exactly. 

It seems like other pages use the same url structure when linking across sections. For example: https://github.com/sdispater/poetry/blob/master/docs/docs/configuration.md#L3

Additionally, https://poetry.eustace.io/docs/pyproject/#extras seems to work just fine in the browser.

Please let me know if I missed something

stephsamson:
Hey @m1so thanks for the quick response! You're right that it's the structure in the other sections but it looks to be breaking on the GitHub docs page too when you click on the link from that page [see here](https://github.com/sdispater/poetry/blob/master/docs/cli/#config).

Yes, that link works fine in the browser but the problem is the linking in the GitHub docs page. Ideally, neither would be breaking :(

m1so:
Oh I see... could maybe the docs folder be moved one level down, placing the `mkdocs.yml` to the root of the repo and moving the theme folder to possibly `/assets`?

Not really sure how the aforementioned placement would affect building the docs as I couldn't find any related commands in the Makefile. 

Another solution might be changing https://www.mkdocs.org/user-guide/configuration/#docs_dir to `.` instead of the default `docs` and subsequently removing the `/docs/docs` directory.

mvoitko (approved):

finswimmer:
It looks like this is fixed in the meantime.

Thanks a lot for your contribution anyway :+1:

Issue #1425, opened 2019-09-30
`poetry env use` and `poetry env remove` are based on the directory name instead of `tool.poetry.name`


maggyero:
The `poetry env use {python}` and `poetry env remove {python}` commands incorrectly use the directory name of a project instead of the `tool.poetry.name` property of the `pyproject.toml` file to create/delete a virtual environment.

This clashes with the `poetry add`, `poetry build`, `poetry install`, `poetry lock`, `poetry remove`, `poetry run`, `poetry shell`, `poetry show` and `poetry update` commands which correctly use the `tool.poetry.name` property of the `pyproject.toml` file to implicitly create a virtual environment.

## Example

For instance if a virtual environment was implicitly created with `poetry install` (so based on `tool.poetry.name`):

```
$ poetry install
Creating virtualenv rule_detection-Yz5oMgHf-py3.7 in C:\Users\maggyero\AppData\Local\pypoetry\Cache\virtualenvs
Using virtualenv: C:\Users\maggyero\AppData\Local\pypoetry\Cache\virtualenvs\{tool.poetry.name}-Yz5oMgHf-py3.7
```

and then removed with `poetry env remove python` (so based on the directory name of the project), a `ValueError` is raised if `{directory_name}` ≠ `{tool.poetry.name}`:

```
$ poetry env remove python
[ValueError]
Environment "{directory_name}-Yz5oMgHf-py3.7" does not exist.
```

FYI @stephsamson @sdispater

KelSolaar:
Hi @maggyero,

I have referenced it here: #1417

sdispater:
This has been fixed in #1477

maggyero:
Thanks a lot @sdispater!

PR #1469, opened 2019-10-15
Fix #1417 - poetry env commands are not using correct package name

### LINKED ISSUE #1417: Various issues when package name differs from directory name.

peter-schmidbauer:
As outlined in issues #1417 and #1425, the `poetry env` commands are behaving unexpectedly when the package name specified in pyproject.toml differs from the directory name of the package. 

This PR tries to fix this behavior.

I changed the current tests to use the package name instead of the directory name, I assume that's good enough for this issue. If more testing is needed, let me know.

### DIFF
--- a/poetry/console/commands/env/info.py
+++ b/poetry/console/commands/env/info.py
@@ -14,7 +14,9 @@ class EnvInfoCommand(Command):
         from poetry.utils.env import EnvManager
 
         poetry = self.poetry
-        env = EnvManager(poetry.config).get(cwd=poetry.file.parent)
+        env = EnvManager(poetry.config).get(
+            cwd=poetry.file.parent, name=poetry.package.name
+        )
 
         if self.option("path"):
             if not env.is_venv():
--- a/poetry/console/commands/env/list.py
+++ b/poetry/console/commands/env/list.py
@@ -15,9 +15,11 @@ class EnvListCommand(Command):
 
         poetry = self.poetry
         manager = EnvManager(poetry.config)
-        current_env = manager.get(self.poetry.file.parent)
+        current_env = manager.get(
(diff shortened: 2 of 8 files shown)


sdispater:
Thanks for your contribution!

However, this is superseded by #1477 so I am closing your PR.

Issue #1497, opened 2019-10-25
cleo and clikit upgrade break poetry


itajaja:
cleo was released an hour ago, and it was a minor change, but includes a breaking change from clikit. since clikit is used **directly** in poetry, clikit should not be a transitive dependency, but an actual depenendecy in this poetry's pyproject.toml. The issue for now is to install the previous version of cleo before installing poetry.

cc @c0state

erj1:
+1

sdispater:
I just published a `1.0.0b3` release that should fix the issue.

Note that only the `1.0` prereleases were affected since previous versions are using a version of `cleo` that does not depend on `clikit`.

itajaja:
thanks @sdispater !

hvdklauw:
This does not fix it for us.

We still can't upgrade past 1.0.0a5 because of #1313 which we need because of #338 

Using poetry to install the packages really isn't an option because it makes our CI times way longer then they need to be.

I love the way poetry works, but I think we need to start looking at alternatives (again).

PR #1498, opened 2019-10-25
define clikit as a direct dependency

itajaja:
this is not a fix for https://github.com/sdispater/poetry/issues/1497 but it's the right thing to do, and will avoid similar problems in the future. in general, the rule of thumb is that if you are directly importing something, then it needs to be specified as a direct dependency, and not rely on transitivity to be there (because it also might introduce silent breaking changes, see #1497)


sdispater:
Thanks!

This has already been fixed on `master`, so I am closing this.

itajaja:
awesome 💪

Issue #1480, opened 2019-10-18
Conflicting dependences where different Python versions are specified.   Because usignals depends on both ipython (^5) and ipython (^7), version solving failed.


kierun:
- [X] I am on the [latest](https://github.com/sdispater/poetry/releases/latest) Poetry version.
- [X] I have searched the [issues](https://github.com/sdispater/poetry/issues) of this repo and believe that this is not a duplicate.
- [X] If an exception occurs when executing a command, I executed it again in debug mode (`-vvv` option).

- **OS version and name**: CentOS 7, Python 3.7.4
- **Poetry version**: 0.12.17
- **Link of a [Gist](https://gist.github.com/) with the contents of your pyproject.toml file**: 

## Issue

I have this so that my module works on either 2.7 or 3.7 and in `pyproject.toml`, I specified:

```
[tool.poetry.dependencies]
python = "^2.7 || ^3.7"  # ← Run on either 2.7 or 3.7???
[…]
[tool.poetry.dev-dependencies]
ipython = [
    {version = "^5", python = "~2.7"},  # ← Install version 5 under py2.7???
    {version = "^7", python = ">=3.5"}  # ← Install version 7.8 under py37???
]
```

Which I assume means that I want ipython version 5 for 2.7 and ipython version 7 for any python above 3.5. However, what I get is:

```
poetry -vvv debug:resolve
Resolving dependencies...
[…]
   1: fact: usignals depends on ipython (^5)
   1: fact: usignals depends on ipython (^7)
[…]
   1: derived: ipython (^7.8)
   1: conflict: usignals depends on ipython (^5)
   1: ! not ipython (^5) is satisfied by ipython (^7)
   1: ! which is caused by "usignals depends on ipython (^7)"
   1: ! thus: version solving failed
   1: Version solving took 0.016 seconds.
   1: Tried 1 solutions.
```

_Is this a bug or did I misunderstand something?_

kierun:
WOOT! Thank you.

Is there a way I can test this on my setup? If not, when is the next release out

agilgur5:
@kierun looks like this was released in [`1.0.0b3`](https://github.com/sdispater/poetry/releases/tag/1.0.0b3)

Issue #666, opened 2018-11-23
SolverProblemError with multiple constraints that do not intersect


mattyclarkson:
- [x] I am on the [latest](https://github.com/sdispater/poetry/releases/latest) Poetry version.
- [x] I have searched the [issues](https://github.com/sdispater/poetry/issues) of this repo and believe that this is not a duplicate.
- [x] If an exception occurs when executing a command, I executed it again in debug mode (`-vvv` option).

- **OS version and name**: MacOSX
- **Poetry version**: 0.12.8
- **Link of a [Gist](https://gist.github.com/) with the contents of your pyproject.toml file**: [pyproject.toml][pyproject]

## Issue

I would like to list `pylint` as a dependency of my tool. `pylint^1` supports Python 2 and `pylint^2.0` supports Python 3. From the documentation, I assumed the following would work:

```toml
pylint = [
  { version = "^2", python = "^3" },
  { version = "^1", python = "^2" }
]
```

But I get a `SovlerProblemError` that says *Because phidias depends on both pylint (^2) and pylint (^1), version solving failed.*

I have the full [output][stdout] in the Gist. It has the following in it:

```
   1: fact: phidias depends on pylint (^2)
   1: fact: phidias depends on pylint (^1)
```

That shouldn't be true as I can't be running a Python interpreter that validates **both** of the Python semantic versions `^2`/`^3`.

[pyproject]: https://gist.github.com/mattyclarkson/64ed8bf8444961ed5c5c083d166a5a93#file-pyproject-toml
[stdout]: https://gist.github.com/mattyclarkson/64ed8bf8444961ed5c5c083d166a5a93#file-stdout

mattyclarkson:
@sdispater I'm happy to fix this up if you can point me in the correct direction. I'm just trying to find where the dependencies are loaded and when the `python` check occurs.

sdispater:
The bug comes from the method here: https://github.com/sdispater/poetry/blob/master/poetry/puzzle/provider.py#L452

Basically, for the root package we take the dependencies  as is and do not perform the proper checks like for any other package.

Fixing this is actually simple: remove the return early condition.

```python
if not package.is_root() and package.source_type not in {"directory", "file", "git"}:
    package = DependencyPackage(
        package.dependency,
        self._pool.package(
            package.name, package.version.text, extras=package.requires_extras
        ),
    )
```

Feel free to make a PR with a test covering this and I'd gladly review it. If you don't have time, let me know and I will do it.

[...]

PR #1509, opened 2019-10-27
add failing test case: multiple constraints on root devDeps

agilgur5:
- SolverProblemError occurs with multiple constraints that do not
  intersect in dev-dependencies
  - this was fixed for main dependencies in 0.12.11 and 1.0.0a2, but
    still occurs for devDeps
    - moving devDeps to main makes the error go away

See my comment in https://github.com/sdispater/poetry/issues/666#issuecomment-531100022 as well as another confirming this occurs https://github.com/sdispater/poetry/issues/666#issuecomment-541338055 .
The failing test case puts this bug into a reproducible case and solidifies the existence of the bug.

I tried to fix the bug on my own, but couldn't figure out where it occurs / what's causing it. There's very little code that deviates between devDeps and main deps -- as far as I can tell, only where `category = "dev"` or `dev_requires` is used.

# Pull Request Check List

### DIFF
--- a/tests/puzzle/test_solver.py
+++ b/tests/puzzle/test_solver.py
@@ -1570,3 +1570,21 @@ def test_multiple_constraints_on_root(package, solver, repo):
         ops,
         [{"job": "install", "package": foo15}, {"job": "install", "package": foo25}],
     )
+
+
+def test_multiple_constraints_on_root_dev(package, solver, repo):
+    package.add_dependency("foo", {"version": "^1.0", "python": "^2.7"}, category="dev")
+    package.add_dependency("foo", {"version": "^2.0", "python": "^3.7"}, category="dev")
+
+    foo15 = get_package("foo", "1.5.0")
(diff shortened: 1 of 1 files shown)


agilgur5:
Oh, looks like this might have been fixed last week in https://github.com/sdispater/poetry/pull/1482 . That would make #1480 a duplicate of #666 (but for devDeps). My comments in #666 have gotten a number of upvotes and there wasn't a response there, so I'm guessing folks haven't seen #1482 and #1480. 

Per https://github.com/sdispater/poetry/issues/1480#issuecomment-543829067, not sure if it's been released yet to try.

agilgur5:
Ok, once I updated Poetry to a prerelease version in order to install Poetry's deps (see https://github.com/sdispater/poetry/issues/1510#issuecomment-546731837), I was able to run the latest tests on `master` and threw this in and it passed. So seems like this duplicates the test in #1482 , will close this out.

PR #1494, opened 2019-10-25
Fix installing Poetry for Python3

hongquan:
Previously, running _get-poetry.py_ with Python 3 didn't give you ability to invoke Poetry with Python 3. This PR is to fix that.

The PR also update installation documentation, giving more examples.

### Pull Request Check List

- [ ] Added **tests** for changed code.
- [x] Updated **documentation** for changed code.

PR #1518, opened 2019-10-29
Detect python version

michielboekhoff:
This PR allows users to manually specify the Python interpreter to use when installing Poetry. If one is not passed, it will try to find a `python3` executable on the PATH, and otherwise default to the `python` executable. This is an implementation as discussed with @brycedrennan in #1042.

### DIFF
--- a/get-poetry.py
+++ b/get-poetry.py
@@ -17,6 +17,7 @@ What this means is that one Poetry installation can serve for multiple
 Python versions.
 """
 import argparse
+import distutils.spawn
 import hashlib
 import json
 import os
@@ -196,7 +197,7 @@ POETRY_LIB = os.path.join(POETRY_HOME, "lib")
 POETRY_LIB_BACKUP = os.path.join(POETRY_HOME, "lib-backup")
 
 
-BIN = """#!/usr/bin/env python
+BIN = """#!{python_binary}
 # -*- coding: utf-8 -*-
 import glob
 import sys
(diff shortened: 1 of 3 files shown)


michielboekhoff:
Not sure what's going on with the Windows builds there.

The Linux 2.7 build appears to be failing as it was using the `python3` executable. I'm fixing this in the CI now.

Jamim:
Similar PR: #1494.

michielboekhoff:
@Jamim - thanks for that!

This is as per the discussion in #1042, going a few steps further than what was done there and in #1494.

nottrobin:
Can't wait for this! When I saw https://github.com/python-poetry/poetry/releases/tag/1.0.0 had been released, I really thought it would have included this fix :(

michielboekhoff:
@nottrobin - cheers for the enthusiasm about this! I've just updated this project against the latest `develop` branch.

Apologies about the force-pushing; I did this to clean up some commits and to keep the history clean.

If anyone would like to take a look at this, it'd be much appreciated. I'd still love to see this included in Poetry.

nottrobin:
A note: When running poetry v1.0.0 I'm still getting a `pip not found` error. I thought this had been fixed in previous work, but it must have crept back in.

Should this PR use a similar mechanism for ensuring that the `pip` binary used is appropriate? Either `{python} -m pip {command}` or `{pip} {command}`?

get-poetry.py
```diff
@@ -312,13 +313,15 @@ def __init__(
         force=False,
         accept_all=False,
         base_url=BASE_URL,
+        binary=None,
```
mvoitko, line 316: This probably should not be None

PR #1519, opened 2019-10-30
init: change dev dependency prompt

and800:
A very simple fix, continuation for #1221. It seems that words `require` and `require-dev` have come from another package manager, and these words don't make much sense in Poetry. The prompt for main deps has been already patched with #1221, I've just done the same for dev deps.


and800:
The CI had errored, so I resubmitted, it seems ok now.

PR #1520, opened 2019-10-30
Flake8 config

cc-a:
Quite speculative but this PR provides the flake8 configuration file I used whilst working on poetry. It's compatible with black so should be a useful addition for people who use flake8 in their IDE's as standard.

### DIFF
--- a/.flake8
+++ b/.flake8
@@ -1,0 +1,3 @@
+[flake8]
+max-line-length = 88
+ignore = E501, E203, W503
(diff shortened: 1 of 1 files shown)


finswimmer:
Hello @cc-a ,

thanks a lot for your contribution! In the meantime @sdispater has included a `.flake8` which includes your suggestions and something more. This is why I close the PR.

fin swimmer

PR #1511, opened 2019-10-28
Equivalent behaviour to pip install --find-links

cc-a:
# Pull Request Check List

- [x] Added **tests** for changed code.
- [x] Updated **documentation** for changed code.

# The What

This PR partially satisfies #1391. This PR implements a new repository class that reproduces the behaviour of the [--find-links](https://pip.pypa.io/en/stable/reference/pip_wheel/#find-links) flag to `pip install`. This allows installation of packages from online locations not compliant with Pep 503. 

# The How

The new class FindLinksRepository is implemented as a subclass of LegacyRepository as the desired behaviours are very similar with only slight changes in expected directory structure. Minor changes were made to LegacyRepository to provide functions that could be overwritten in the subclass. Combined with implementing FilteredPage as a subclass of Page this provides a fairly minimal approach to implementing this new functionality. A few minor associated changes in factory.py and pip_installer.py were required to accommodate the new repository class. 

# The Testing

The majority of the business logic of FindLinksRepository is derived from LegacyRepository and hence testing of this has not been repeated. Instead testing focuses on making sure that the behaviour of the major public functions of LegacyRepository is recreated with an appropriately structured test set.

A simple high level test can be performed using pytorch as an example. Create the below entry in a pyproject.toml:

```toml
[tool.poetry.dependencies]
python = "^3.7"
torch = "1.2.0+cpu"

[[tool.poetry.source]]
name = "pytorch"
url = "https://download.pytorch.org/whl/torch_stable.html"
find_links = true
```
Then `poetry install`.

# Not done yet

The behaviour of --find-links to install from local paths or file:// urls has not yet been replicated. I will follow up with another PR for this functionality if this one is successful.

PR #1527, opened 2019-10-31
Tidy up

cc-a:
This PR takes a pass at some untidyness noticed when working on #1511. Mostly sorted out unnecessary or missing imports and redefinitions.

### DIFF
--- a/poetry/config/config.py
+++ b/poetry/config/config.py
@@ -4,6 +4,7 @@ import os
 import re
 from copy import deepcopy
 from typing import Any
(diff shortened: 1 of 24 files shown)


finswimmer:
Hello @cc-a ,

thanks for contributing to poetry!

It looks like you don't work anymore on this PR. Furthermore I couldn't find any unnecessary imports in the current code. This is why I will close the PR. If you disagree please leave a comment.

fin swimmer

PR #1536, opened 2019-11-03
Support installing and uninstalling on Fish shell

### LINKED ISSUE #544: Installation does not support the 'fish' shell

- **OS version and name**: Linux Mint (as Ubuntu 16.04)
- **Poetry version**: 0.12.4
- **Link of a [Gist](https://gist.github.com/) with the contents of your pyproject.toml file**: N/A

## Issue
Currently, the install script does not correctly detect and configure the fish shell. On my distribution, the correct thing to do would be to add the following line to `~/.config/fish/conf.d/poetry.fish` (which is a fairly common idiom):
```
set PATH ~/.poetry/bin $PATH
```

Note that there is no `=` or `:`, and `~` is considered idiomatic (although $HOME works as well, I believe... this could need testing.)

Detection can be done via inspection of $SHELL which should generally be set to `/usr/bin/fish` but certainly should have `basename $SHELL` as `fish`.

Happy to submit a PR if you'd like one, a bit too busy to do that as part of the report but I can do it later if you're interested.

Thanks!

justinmayer:
Previously, the install script mentioned $PATH modifications that do not
apply to Fish. Users may have thought everything was all set up, only to
find invoking the `poetry` executable would fail in a Fish shell
environment because Poetry's bin directory could not be found on $PATH.

Now, if Poetry's bin directory is not found in Fish's $PATH environment
variable, that directory is added to the `fish_user_paths` universal
variable, making the `poetry` executable immediately available to
current and future Fish shell sessions.

Fixes #544


justinmayer:
@stephsamson: Thanks for reviewing. As far as I can tell, that appears to be a glitch with GitHub Actions. If you look at the output, the cause shows:

    ##[error]fatal: couldn't find remote ref refs/pull/1535/merge

The [actual linting step](https://github.com/sdispater/poetry/pull/1536/checks?check_run_id=286805283) appears to pass without problems. Would you mind taking another look?

justinmayer:
@stephsamson: I triggered a new build, and I suspected, this was most likely a GitHub CI glitch. All tests are now passing — I think this is ready to merge.

stephsamson:
@justinmayer woo! Thanks @justinmayer 🚀 one more thing, it makes more sense to merge this against `master` instead of `develop` so that other `fish` users can have this ASAP. I'll merge it on that branch :)

PR #1542, opened 2019-11-05
allow to read package version from specified variable

### LINKED ISSUE #144: peotry version doesn't bump the value in __version__

vanyakosmos:

There is a problem when you need to specify version of the package twice: in `pyproject.toml` and in `package.__version__`. We can't read `pyproject.toml` to calculate `__version__` because config file is not available to the final user of package.

I propose to add alternative definition of version where you need to specify path to the variable which will hold the value of package version.

```toml
[tool.poetry]
version = { attr = "package.sub_module:__version__" }
```
for project with structure:
```
my-package
        ├── pyproject.toml
        ├── README.rst
        └── package
            ├── __init__.py
            └── sub_module.py  # has __version__ variable

```

this pull request should close #144


vanyakosmos:
isort failed on "setup.py" files despite `[tool.isort] skip_glob = ["*/setup.py"]` in `pyproject.toml` 🤔

finswimmer:
I've contacted @sdispater as I has the same problem with the `isort` test. It works now.

ryanhiebert:
Hi @vanyakosmos , I was watching this pull request, and it looks like you closed it without resolution. Is there some other issue that conflicts or otherwise relates to this one?

vanyakosmos:
"Double source of package version" problem is controversial (so controversial that some doesn't even see it as problem). Related issue has been opened for over a year and resolution is not yet in sight. 
As for now this pull request just wastes space in PRs lists. So I decided to close PR and wait for some conclusions and thoughts from Sebastien regarding this issue.

PR #1592, opened 2019-11-17
Handle nested wildcards in package includes correctly. Fixes: #1379.

### LINKED ISSUE #1379: tool.poetry.packages handles wildcards * and ** incorrectly in pyproject.toml

## Issue

The [documentation gives the following example](https://poetry.eustace.io/docs/pyproject/#packages) for using the `tool.poetry.packages` property of the `pyproject.toml` file:

```
[tool.poetry]
# ...
packages = [
    { include = "mypackage" },
    { include = "extra_package/**/*.py" },
]
```

However the command `poetry build` raises the following error when the` "extra_package/**/*.py"` path pattern matches multiple files/directories in at least one the of its path segments:

```
[ValueError]
/Users/maggyero/Desktop/project/extra_package/bar/baz.py is not a package.
```

with the following directory layout:

```
project
├── .git
├── mypackage
│   └── foo.py
├── extra_package
│   └── bar
│       ├── baz.py
│       └── qux.py
└── pyproject.toml
```

Here the last `*.py` path segment of the `"extra_package/**/*.py"` path pattern matches `baz.py` and `qux.py`, which triggers the `ValueError`. Deleting the `project/extra_package/bar/qux.py` file removes the `ValueError`. And adding a `project/extra_package/quux` directory raises the same `ValueError`, since this time the second `**` path segment of the `"extra_package/**/*.py"` path pattern matches `bar` and `qux`. So it seems that matching multiple files/directories in the path segments of a path pattern is not allowed. It looks like a _bug_, because each item of the `tool.poetry.packages` list is supposed to include a single top-level module or package and optionally specific submodules (like `"extra_package/**/*.py"` which includes the top-level `extra_package` package and submodules that have a `.py` extension).

Also, since the `tool.poetry.packages` property of the `pyproject.toml` file can include both directories and files, that is to say both packages and non-package modules, I think we should rename to `tool.poetry.modules`.

stephsamson:
# Pull Request Check List

This PR fixes #1379. I've also added some unit tests for `PackageInclude`.

### DIFF
--- a/poetry/masonry/utils/package_include.py
+++ b/poetry/masonry/utils/package_include.py
@@ -44,10 +44,10 @@ class PackageInclude(Include):
             # Probably glob
             self._is_package = True
 
-            # The __init__.py file should be first
+            # Packages no longer need an __init__.py in python3
             root = self._elements[0]
-            if root.name != "__init__.py":
-                raise ValueError("{} is not a package.".format(root))
+            if root.name == "__init__.py":
+                pass
 
             self._package = root.parent.name
         else:
--- a/poetry/masonry/utils/package_include.py
+++ b/poetry/masonry/utils/package_include.py
@@ -12,7 +12,6 @@ class PackageInclude(Include):
             base = base / source
 
         super(PackageInclude, self).__init__(base, include, formats=formats)
-
         self.check_elements()
 
     @property
@@ -35,6 +34,8 @@ class PackageInclude(Include):
         return self.check_elements()
 
     def check_elements(self):  # type: () -> PackageInclude
+        root = self._elements[0]
+
         if not self._elements:
             raise ValueError(
                 "{} does not contain any element".format(self._base / self._include)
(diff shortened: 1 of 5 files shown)


poetry/masonry/utils/package_include.py
```diff
@@ -44,10 +44,10 @@ def check_elements(self):  # type: () -> PackageInclude
             # Probably glob
             self._is_package = True
 
-            # The __init__.py file should be first
+            # Packages no longer need an __init__.py in python3
             root = self._elements[0]
-            if root.name != "__init__.py":
```
sdispater: I think we can change this line to check if there is at least one Python file, regardless of whether it's a `__init__.py` file or not. If there is no Python file, it's not a package and we should raise the exception.

stephsamson: Thanks, I've also added it now in commit 1d0e99d.

PR #1595, opened 2019-11-17
Automate release process

sdispater:
## Pull Request Check List

This PR adds the automation of Poetry releases so that any maintainer can make a release without my intervention.

This workflow is triggered whenever a release tag is pushed. In detail, it does the following:

- Build the custom releases, those needed for the official Poetry installer, for each OS
- Create a new Github release with the tag name
- Upload the custom releases as assets of the Github release
- Upload both the source distribution file and the wheel file to PyPI

Note that it **should** work but will be properly tested for the next release since it's not possible to test it directly from the PR.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,0 +1,237 @@
+name: Release
+
+on:
+  push:
+    tags:
+      - '*.*.*'
+
+jobs:
+
+  Linux:
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/checkout@v1
+    - name: Building release
+      run: |
+        make linux_release
+    - name: Upload release file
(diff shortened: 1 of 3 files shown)


justinmayer:
I love the idea of automating releases, particularly when connecting that to merging pull requests. So much so that I wrote [AutoPub](https://github.com/autopub/autopub) to accomplish exactly that. You read more about the motivation here: https://justinmayer.com/projects/autopub/

Given that this project already uses GitHub Actions for CI functionality, it would be trivial to add the same auto-publication feature to this project as well. Here's an example of GitHub CI configuration that does this already: https://github.com/pelican-plugins/simple-footnotes/blob/master/.github/workflows/main.yml

Just thought I'd mention this in case it's useful. Cheers!

PR #1603, opened 2019-11-18
add sync option to version command

### LINKED ISSUE #144: peotry version doesn't bump the value in __version__

vanyakosmos:
Sync option will allow to update `__version__` variable alongside with version from `pyproject.toml`.

### Usage example

Search for `__version__` variable in `PACKAGE_NAME.py` or `PACKAGE_NAME/__*.py` and update it based of version in `pyproject.toml`:
```
poetry version --sync
```

Update version in both `pyproject.toml` and version file:
```
poetry version patch --sync
```

Sync `__version__` from `path/to/file` with version in `pyproject.toml`:
```
poetry version --sync path/to/file
poetry version major --sync path/to/file
```

### Info

This PR is less radical (and also less usable, imo) version of #1542. This PR implements changes proposed in #144, but instead of changing AST and recreating the file I used regular expression, because recreation of file from new tree will require additional dependency and regex works just fine.

----- 


finswimmer:
Hello  @vanyakosmos,

thanks a lot for your contribution. :+1:  

However, having the version in multiple places is very error-prone. There should always be only one source of truth.

fin swimmer

Issue #1597, opened 2019-11-17
Exclude is not respected when building package.


mhozza:
- [x] I am on the [latest](https://github.com/sdispater/poetry/releases/latest) Poetry version.
- [x] I have searched the [issues](https://github.com/sdispater/poetry/issues) of this repo and believe that this is not a duplicate.
- [x] If an exception occurs when executing a command, I executed it again in debug mode (`-vvv` option).

- **OS version and name**: Linux 5.3.11-arch1-1
- **Poetry version**: 1.0.0b5
- **Link of a [Gist](https://gist.github.com/) with the contents of your pyproject.toml file**: https://gist.github.com/mhozza/345f427c76880295fe5c8b3d630925d6
## Issue

Exclude is not respected when building package.

I have `**/node_modules/**` in my `.gitignore` and I also included `exclude = ["**/node_modules"]` in my pyproject.toml file, but none of them causes the _node_modules_ to be ignored.

When I do `poetry build -vvv` I see a lot of lines similar to:
```
 - Adding: /home/mio/source/trojstenweb/trojsten/special/plugin_prask_2_4_1/static/plugin_prask_2_4_1/js/node_modules/babel-preset-es2015/node_modules/babel-plugin-transform-es2015-block-scoping/node_modules/babel-runtime/node_modules/core-js/fn/log.js
```

finswimmer:
Hello,

this is already fixed by https://github.com/sdispater/poetry/pull/1464 but obviously it is not included in the current version

Can you please check the output of `git ls-files --others -i --exclude-standard`? Because I would thought that the `.gitgnore` should work. If the files you want to excluded that you have already added them with `git add`. You will have to remove them from the git tracking. 

Nevertheless with the fix already in dev the `pyproject.toml` should work.

@sdispater Is this correct that the fix isn't in the current release?

mhozza:
Thanks for the response.

`git ls-files --others -i --exclude-standard` contains these files. I'm quite sure they are not commited to the repository or added using `git add`.
`exclude = ["**/node_modules/**/*"]` makes no difference, so I'm not sure if the fix will actually help. I'm wondering if recursive glob should be used.

Not sure if relevant, but I also have

```toml
packages = [
  { include = "trojsten"}
]
```
in my pyproject.toml file.

finswimmer:
I had the time to take a closer look at this:

* In any case have the `.gitignore` version works for me
* `exclude = ["**/node_modules/**/*"]` works
[...]

PR #1606, opened 2019-11-19
FIX: Regression of not excluding files from build

finswimmer:
This PR changes the way, how it is checked if a file should be excluded from build.

Fixes: #1597 

# Pull Request Check List

### DIFF
--- a/poetry/masonry/builders/builder.py
+++ b/poetry/masonry/builders/builder.py
@@ -2,9 +2,11 @@
 import re
 import shutil
 import tempfile
+
 from collections import defaultdict
 from contextlib import contextmanager
-from typing import Set, Union
+from typing import Set
+from typing import Union
 
 from clikit.api.io.flags import VERY_VERBOSE
 
@@ -13,9 +15,11 @@ from poetry.utils._compat import glob
 from poetry.utils._compat import lru_cache
 from poetry.utils._compat import to_str
 from poetry.vcs import get_vcs
+
 from ..metadata import Metadata
 from ..utils.module import Module
 from ..utils.package_include import PackageInclude
+
 
 AUTHOR_REGEX = re.compile(r"(?u)^(?P<name>[- .,\w\d'’\"()]+) <(?P<email>.+?)>$")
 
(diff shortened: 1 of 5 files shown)


sdispater:
@finswimmer Shouldn't this be based on `master`?

finswimmer:
Maybe? :) Because my PR https://github.com/sdispater/poetry/pull/1464 was based on develop I did here the same way.

Any easy way to switch?

sdispater:
@finswimmer Oh sorry, I thought it was a bug fix but it was actually a regression fix for your previous PR so `develop` is good :-)

sdispater:
@finswimmer I changed the base branch to `master` since your previous fix is already on `master`. You will need to rebase your branch onto `master`.

finswimmer:
@sdispater: That was a hard work doing my first rebase. But now I got it (I think) :rocket:

sdispater:
@finswimmer I don't think your rebase went well since this PR includes commits that are not relevant to your changes. You only need to rebase your 4 commits onto `master`. You might want to use `git rebase --onto` in order to change the base branch.

finswimmer:
TIL: Never ever (never!) use the wrong branch!

Better now?

poetry/masonry/builders/builder.py
```diff
@@ -112,10 +102,18 @@ def find_excluded_files(self):  # type: () -> Set[str]
         return result
 
     def is_excluded(self, filepath):  # type: (Union[str, Path]) -> bool
-        if not isinstance(filepath, basestring):
-            filepath = filepath.as_posix()
+        exclude_path = Path(filepath)
+
+        while True:
```
sdispater: Why not `while len(exclude_path.parts) > 1`?

I try to avoid `while True` statements to have a more explicit condition.

finswimmer: Because - at least in theory - the root folder could be excluded as well. (Which doesn't make sense, if you want  to build a package, but people have sometimes strange ideas).

Of course I could check this if I leave the loop. If you would like this more, I could implement it.

sdispater: You know what? We can leave it like this for now. I don't think this is too much of an issue.

mhozza:
Can confirm the fix works. Thanks!

PR #1630, opened 2019-11-25
fixes issue where exporting to requirements.txt always prepends -e to dependencies

petergaultney:
fixes issue where exporting to requirements.txt always prepends -e to dependencies even if they're not directory dependencies that are installed as editable.

I am not fully educated on the difference between a "Package" and a "Dependency" in `poetry`-parlance, but looking at the way that "Package" _always_ defines `develop = True` by default, and the way the source code in `packages/package.py` provides the `develop` constraint only to `DirectoryDependency`, it seems likely to me that the intent within the requirements.txt `exporter.py` was to use the `dependency.develop` check rather than the `package.develop` check.

In any case, this fixes the issue where my dependencies were being exported as `-e ` (editable) despite being file dependencies rather than editable directory dependencies. It should _also_ fix https://github.com/sdispater/poetry/issues/1451, because now that `develop=false` constraint, which is getting properly added to the Dependency object (but is never added to the Package object) will be used at export time.

### DIFF
--- a/poetry/packages/dependency.py
+++ b/poetry/packages/dependency.py
@@ -63,6 +63,7 @@ class Dependency(object):
 
         self.is_root = False
         self.marker = AnyMarker()
+        self._develop = False
 
     @property
     def name(self):
@@ -154,6 +155,10 @@ class Dependency(object):
             requirement += " ({})".format(str(self.constraint).replace(" ", ""))
 
         return requirement
+
+    @property
+    def develop(self):  # type: () -> bool
+        return self._develop
 
     def allows_prereleases(self):
         return self._allows_prereleases
--- a/poetry/utils/exporter.py
+++ b/poetry/utils/exporter.py
@@ -94,7 +94,7 @@ class Exporter(object):
                 dependency.marker = package.marker
 
                 line = "{}".format(package.source_url)
-                if package.develop:
+                if dependency.develop:
                     line = "-e " + line
             else:
                 dependency = package.to_dependency()


finswimmer:
Hello @petergaultney ,

thank you very much for your contribution!

As you stated out, the intended default behavior is `develop = True`. Your code would change it to `False`. This is why some test fail.

May I ask you to:
- change the default to behavior
- write a test that covers your change

Thanks a lot! :+1:

petergaultney:
@finswimmer thanks for taking a look. I don't think I changed the default behavior for anything - `Dependency` did not have a default `develop` property before my change, as far as I can see. `Package` does, but I did not change `Package` - just the fact that the `package` object is no longer used to determine whether the dependency should be installed as editable.

I would love to understand better what Package and Dependency represent internally to poetry - perhaps there is a better solution than the one I have proposed.

finswimmer:
Hello @petergaultney ,

I took a closer look into this and the problem seems to be bigger. :( In my understanding, the information if `develop=false` is set, should be seen in the `poetry.lock`. But this isn't the case.  Overall, setting `develop=false` doesn't do anything for me. :(

I already have some ideas what have to be done. So if you agree, I open a new issue for this problem to keep track of the status and will close your PR for now as I haven't the feeling that your changes are enough. Feel free to disagree and just leave a comment.

Thanks a lot for pointing to this issue!

fin swimmer

petergaultney:
Thanks @finswimmer , that sounds good. I look forward to seeing the solution, as I think it will help me understand Poetry better, and I'd love to be able to contribute to the codebase at some point.

finswimmer:
Hey @petergaultney ,

the issue I've open for this is #1654, the corresponding PR is #1656 .

I started to contribute to poetry just like you. I had a problem that needs to be solved and started to read through the code. What really helps is looking at the tests. Try to identify one, that seems to cover your issue. Use the debugger of your IDE to walk through the code. This is how I get used to the code base.

Keep on contributing. We really appreciate it! :+1: 

fin swimmer

PR #1650, opened 2019-11-28
Bugfix: Empty and Any markers saved to lockfile

JBKahn:
When locking some private packages of ours (that have extras) it used to copy those extras over in `1.0.0b3` when we changed that behaviour it's now the `Any` marker and when it stringifies that when generating the lockfile it creates a format the lockfile doesn't support.

**bugfix**: Originally, the `MarkerUnion` didn't correctly respond to `is_any` or `is_empty` because it didn't evaluate the collection of markers which led to it trying to stringify those markers.

This on conjunction with the other fix means that the `[Any, Any]` case is simply excluded from the lockfile just like a single Any case.

**bugfix**
Before:
`' or '.join([Any, Any])` -> `' or '`
`' or '.join([Any, REAL MARKER])` -> `' or REAL MARKER'`
`' or '.join([Empty, Empty])` -> `'<empty> or <empty>'`

Now
`' or '.join([Any, Any])` -> `''`
`' or '.join([Any, REAL MARKER])` -> `'REAL MARKER'`
`' or '.join([Empty, Empty])` -> `''`

This means if we have a mix of markers with any/empty that won't be propagated to the lockfile.

### DIFF
--- a/poetry/version/markers.py
+++ b/poetry/version/markers.py
@@ -633,7 +633,7 @@ class MarkerUnion(BaseMarker):
         )
 
     def is_any(self):
-        return all(m.is_any() for m in self._markers)
+        return any(m.is_any() for m in self._markers)
 
     def is_empty(self):
         return all(m.is_empty() for m in self._markers)
--- a/poetry/version/markers.py
+++ b/poetry/version/markers.py
@@ -628,7 +628,15 @@ class MarkerUnion(BaseMarker):
         return h
 
     def __str__(self):
-        return " or ".join(str(m) for m in self._markers)
+        return " or ".join(
+            str(m) for m in self._markers if not m.is_any() and not m.is_empty()
+        )
+
+    def is_any(self):
+        return all(m.is_any() for m in self._markers)
+
+    def is_empty(self):
+        return all(m.is_empty() for m in self._markers)
 
 
 def parse_marker(marker):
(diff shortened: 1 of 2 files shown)


JBKahn:
@sdispater thoughts?

JBKahn:
@finswimmer I see you're also able to commit on this repo, any thoughts?

poetry/version/markers.py
```diff
@@ -628,7 +628,15 @@ def __hash__(self):
         return h
 
     def __str__(self):
-        return " or ".join(str(m) for m in self._markers)
+        return " or ".join(
+            str(m) for m in self._markers if not m.is_any() and not m.is_empty()
+        )
+
+    def is_any(self):
+        return all(m.is_any() for m in self._markers)
```
sdispater: Since it's a union only one of the markers needs to return `True` for `is_any()`.
```suggestion
        return any(m.is_any() for m in self._markers)
```

JBKahn:
Same for empty?

sdispater:
@JBKahn No, it's fine for `is_empty()` since all markers must be empty for the union to be empty as well.

JBKahn:
K, then I think with that test change, this should be good for the next release?

JBKahn:
@sdispater on a side note, I'm not super familiar with the code yet but if you're looking for help maintaining poetry, I'm happy to give some time to the project.

Issue #1654, opened 2019-11-30
exporting to requirements.txt always prepends -e to dependencies


finswimmer:
- [x] I am on the [latest](https://github.com/sdispater/poetry/releases/latest) Poetry version.
- [x] I have searched the [issues](https://github.com/sdispater/poetry/issues) of this repo and believe that this is not a duplicate.
- [x] If an exception occurs when executing a command, I executed it again in debug mode (`-vvv` option).

- **Poetry version**:  1.0.0b8

## Issue
As reported by @petergaultney in https://github.com/sdispater/poetry/pull/1630 `-e` get always prepended for folder dependencies regardless `develop = false` is set in `pyproject.toml` or not.

Further investigation has shown, that this setting isn't transported `poetry.lock`. Overall `develop = false` doesn't have any effect as expected for now.

NoahTheDuke:
This also fails for urls:
```
en-core-web-sm = {url = "https://github.com/explosion/spacy-models/releases/download/en_core_web_sm-2.2.0/en_core_web_sm-2.2.0.tar.gz"}
```
becomes
```
-e https://github.com/explosion/spacy-models/releases/download/en_core_web_sm-2.2.0/en_core_web_sm-2.2.0.tar.gz
```
instead of
```
https://github.com/explosion/spacy-models/releases/download/en_core_web_sm-2.2.0/en_core_web_sm-2.2.0.tar.gz \
    --hash=sha256:65cea113d7509d5ab515583bbcbc9a31b4157bebbb73133d4110d2945476eaa4
```

finswimmer:
Hello @NoahTheDuke ,

thanks a lot for reporting this! 

The correct use of `-e` for folder dependencies is now fixed by #1656 .

The fix for url dependencies need some more work. I already found out, what have to be done and I will provide another PR the next days.

fin swimmer

PR #1656, opened 2019-12-01
fix (#1654): exporting to requirements.txt always prepends -e to dependencies

finswimmer:
This PR fixes (at least partially) #1654.

It now respects the `develop` value set in `pyproject.toml` when exporting to `requirements.txt` and also installing in non-development mode works again.

# Pull Request Check List

### DIFF
--- a/poetry/packages/locker.py
+++ b/poetry/packages/locker.py
@@ -146,7 +146,6 @@ class Locker(object):
                 package.source_url = info["source"]["url"]
                 package.source_reference = info["source"]["reference"]
 
-
             packages.add_package(package)
 
         return packages
--- a/poetry/packages/locker.py
+++ b/poetry/packages/locker.py
@@ -143,6 +143,9 @@ class Locker(object):
                 package.source_url = info["source"]["url"]
                 package.source_reference = info["source"]["reference"]
 
+                if "develop" in info["source"]:
+                    package.develop = info["source"]["develop"]
+
             packages.add_package(package)
 
         return packages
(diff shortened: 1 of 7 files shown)


finswimmer:
I'm not fully satisfied with this PR now. But I already pushed it to get some feedback.

- We are still missing tests for `locker` and `install` command and I'm not sure how to include them.
- My path dependencies are still get install in development mode, regardless I have set `develop=false`. I will take a look at this in the next step.
- There are more places that looks suspicious to me, that the `develop` value is missed, e.g.

https://github.com/sdispater/poetry/blob/c1f9838b516a7af02b3efb867cddcdba4773d5a6/poetry/packages/package.py#L406-L415

finswimmer:
Hm, any hints for me to get the tests pass @sdispater , @stephsamson ? Why the pathes doesn't seem to be correct? Locally it works.

finswimmer:
Ok, tests seem to work now. Not sure why it works here. The paths definitely doesn't work local.

Time for feedback please :)

poetry/packages/locker.py
```diff
@@ -301,5 +304,7 @@ def _dump_package(self, package):  # type: (poetry.packages.Package) -> dict
             }
             if package.source_type:
                 data["source"]["type"] = package.source_type
+            if package.source_type == "directory":
+                data["source"]["develop"] = package.develop
```
sdispater: Is this the right place to put this? It's an attribute of the package rather than the source, so I think it's better to put it in the `[[package]]` section.

finswimmer: Good point! Changed it.

Issue #1440, opened 2019-10-06
Document how to `add` a dependency with an exact version


homeworkprod:
Hi, Poetry newbie here, using Poetry v0.12.17.

Trying to get my dependencies specified in `requirements.txt` format into Poetry took me longer than expected. Here's what I've found.

- [x] I have searched the [issues](https://github.com/sdispater/poetry/issues) of this repo and believe that this is not a duplicate.

The short documentation section on [Exact [version] requirements](https://poetry.eustace.io/docs/versions/#exact-requirements), as opposed to others, includes no examples of the specific version format string and, this feels especially important for this case, a full Poetry command to add a specific version of a dependency.

I wanted to install a specific version of the `attrs` package (as an example), so I tried this:

- `poetry add attrs 19.1.0` results in
  ```
  Using version ^19.2 for attrs
  
  [PackageNotFound]  
  Package [19.1.0] not found.
  ```
  Nothing added to `pyproject.toml`.
  It seems that Poetry expects the version number to be a package name, which fits the usage string of the `add` command (`add [options here …] <name> (<name>)...`). That is fine. 

- `poetry add attrs==19.1.0`
  This is the format I expect to work as I'm used to it from `requirements.txt` and probably `setup.py` and other common, but older, approaches to specify requirements.

  The line `attrs = "=19.1.0"` gets added to `pyproject.toml`. This doesn't look wrong per se, but my understanding is that the equality sign is at least redundant/superfluous here and, thus, should not have been added automatically.

  And I'm not sure if it will actually work later on. Maybe Poetry assumes the second `=` is the start of the version number? That sounds like trouble.

  Maybe Poetry should normalize `==` to `=` as a special case?

- `poetry add attrs=19.1.0`
  This, to me, uses an unexpected format, coming from `requirements.txt` (where `==` seems to be/is common, though I'm not sure if a single `=` would work, too).

  However, the line `attrs = "19.1.0"` gets added to `pyproject.toml` and it looks correct to me.

- `poetry add 'attrs 19.1.0'`
  Okay, let's try to group package name and version number into arguments using quotes.

  Again, the line `attrs = "19.1.0"` gets added to `pyproject.toml`.

  This is fine, and not having to type `=` is nice (though the quotes then become necessary with this format), but should be documented.

- `poetry add 'attrs == 19.1.0'`
[...]

PR #1657, opened 2019-12-01
Add examples for 'poetry add' with a version constraint (#1440)

bitswamp:
This addresses the documentation aspect of #1440.

It's a documentation-only change which affects the main readme and the output of `poetry help add`. 

These are the places I looked first for this info, but if there are other suitable spots to change, or if a different constraint format is preferred, let me know.

Thanks for your time!

### DIFF
--- a/docs/docs/configuration.md
+++ b/docs/docs/configuration.md
@@ -100,7 +100,7 @@ Defaults to one of the following directories:
 
 - macOS:   `~/Library/Caches/pypoetry`
 - Windows: `C:\Users\<username>\AppData\Local\pypoetry\Cache`
-- Unix:    `~/.cache/pypoetry/virtualenvs`
+- Unix:    `~/.cache/pypoetry`
 
 ### `virtualenvs.create`: boolean
 


viq:
It would be appreciated for this to get in, as apparently people repeatedly stumble on this issue.

finswimmer:
Hello @bitswamp ,

thanks a lot for helping to improve the documentation!

At least in my tests, it is not necessary to begin the version constraint with a `=`. Just use the sign you need.

Do you agree with this?

fin swimmer

viq:
I disagree with this, 'package<5.0' doesn't work, you need to use
'package=<5.0' to get that constraint.

finswimmer:
Hello again,

can you address what "doesn't work"?

My testing:

```
[tool.poetry]
name = "my_project"
version = "0.1.0"
description = ""

[tool.poetry.dependencies]
python = "^3.7"

[build-system]
requires = ["poetry>=0.12"]
```

```
$ poetry install
Creating virtualenv my-project in /home/finswimmer/1440/.venv
Updating dependencies

Writing lock file

$ poetry add requests\<2.19.0

Updating dependencies

Writing lock file


Package operations: 5 installs, 0 updates, 0 removals

  - Installing certifi (2019.11.28)
  - Installing chardet (3.0.4)
  - Installing idna (2.6)
  - Installing urllib3 (1.22)
  - Installing requests (2.18.4)

$ poetry add "requests<2.18.0"

Updating dependencies

Writing lock file


Package operations: 0 installs, 3 updates, 0 removals

  - Updating idna (2.6 -> 2.5)
  - Updating urllib3 (1.22 -> 1.21.1)
  - Updating requests (2.18.4 -> 2.17.3)
```

Works without the `=`. The important thing when using `<` is to escape this sign by `\` or put the whole string within `"`. Because within a shell the `<` have a special meaning. The same for the `>`. I think it would be a good idea to give a note about this in the docs, because this might not be obvious to all people.

bitswamp:
I see in the 1.0.0 announcement another format was presented using @ instead of =. It's in the docs now at https://python-poetry.org/docs/cli/#add. 

I'll update this PR to remove the readme change and add the same text from the docs to the output of `poetry help add`.

Issue #1214, opened 2019-07-09
Install globally with --user


raxod502:
- [x] I have searched the [issues](https://github.com/sdispater/poetry/issues) of this repo and believe that this is not a duplicate.
- [x] I have searched the [documentation](https://poetry.eustace.io/docs/) and believe that my question is not covered.

## Feature Request

I am using Poetry to install project dependencies into a Docker container where I do not have root access. Since using a virtualenv would be unnecessary (there is only ever one project in this container) and a waste of time (this is done as part of a web service aiming at speed), I configure `settings.virtualenvs.create` to `false`. However, this causes Poetry to try to install the packages system-globally, which fails because I don't have root access in the container.

I would like a way to tell Poetry to use `pip install --user` instead of just `pip install` to install packages when virtualenvs are disabled.

richin13:
I picked up this issue and started working on an implementation

Can somebody re-open the issue please? cc: @sdispater

finswimmer:
Hey @richin13 ,

thanks a lot for your contribution! I reopened this issue and assigned you for seeing that this is in progress.

fin swimmer

richin13:
Awesome, thank you @finswimmer

abn:
Shouldn't the following just work?
```sh
pip install --user /path/to/poetry-project
```
This should result in your project being built in an isolated environment without development requirements and `pip` installing the built package respecting the `--user` flag.

harrybiddle:
@richin13 did your submission go anywhere or was it ignored? Currently my workaround is:

```bash
poetry config virtualenvs.create false
pip install --user --requirement <(poetry export --format requirements.txt)
```

However, this doesn't give me the project itself installed into the user.

abn:
@harrybiddle why not let pip install the project itself as suggested above?

@raxod502 @richin13 I beleive this behaviour is dependent on the underlying pip version (ie. the pip provided in your container). This often times  (if ensured via `python -m ensure pip`) tends to be older versions.I beleive since `20.0`, `pip` defaults to installing to user site when not root. Trying a `pip install --upgrade pip` prior to doing `poetry install` might do the trick here.

harrybiddle:
Hey @abn, thanks for the help, I really appreciate it.

I just gave this a go and it worked, except for two things: (1) my dev dependencies didn't get installed, and (2) I think your command installs the dependencies in the `pyproject.toml` rather than the pinned dependencies in `poetry.lock`, is that correct?

[...]

PR #1679, opened 2019-12-05
[WIP] Allow --user installs

richin13:
# Pull Request Check List

# Description

This PR introduces a new feature of user installs (pip `--user` flag) to support use cases like the one described in #1214
It enables this behaviour by introducing a new config item: `settings.user-installs`

The motivation to add this through a config flag was to avoid changes to poetry's CLI. The `--user` is sent to the `pip install` command. I noticed `--user` is also used in `pip list` & `pip config`. Does poetry uses any of these under the hood?

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -100,7 +100,7 @@ description = "Python package for providing Mozilla's CA Bundle."
 name = "certifi"
 optional = false
 python-versions = "*"
-version = "2019.9.11"
+version = "2019.11.28"
 
 [[package]]
 category = "main"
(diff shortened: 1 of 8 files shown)


paretje:
I didn't try the PR yet, but I did experiment with user local installs using poetry in general. At the moment this can be done using poetry by setting `virtualenvs.create` to `false` and the environment variable `PIP_USER` to `yes`. The only issue for me at the moment is the fact that `SystemEnv` is initialised with `sys.path`. `Env.path` is used to determine the location to clone git dependencies to, i.e. poetry tries to clone git repositories to subfolders of `/usr/src`. So another possible path to enable user local installs would be to be able to configure that path.

PR #1690, opened 2019-12-07
Work around for --no-deps failure in pip

### LINKED ISSUE #1689: Installing directory poetry package with dependencies in secondary source fails

- **MacOS 10.14**: 
- **1.0.0b8**: 

## Issue
Due to https://github.com/pypa/pip/issues/7444 installing a directory which is managed by poetry or has a pyproject.toml file present will cause the `--no-deps` argument to be ignored. 

This can go unnoticed as long as you are only working with pypi dependencies but when your package depends on a private pypi repository this causes installs to fail.

wakemaster39:
I am not sure about the checklist, I went looking at the unit tests and there is no coverage for this area of the pip_installer so not sure where I should be targeting to start with that.

I don't think this needs to be documented as it should all be invisible in the background but maybe we want something visible  so users know it is messing around like this?

I also wasn't sure as this changes behaviour but it is also a bug fix. I can close and resubmit if it is preferred against the develop branch. I am impacted by this issue now so I treated it as a bug.

This should resolve #1689 

# Pull Request Check List

### DIFF
--- a/poetry/installation/pip_installer.py
+++ b/poetry/installation/pip_installer.py
@@ -176,12 +176,13 @@ class PipInstaller(BaseInstaller):
         return name
 
     def install_directory(self, package):
+        from uuid import uuid4
+
         from poetry.masonry.builder import SdistBuilder
         from poetry.factory import Factory
         from poetry.utils._compat import decode
         from poetry.utils.env import NullEnv
         from poetry.utils.toml_file import TomlFile
-        from uuid import uuid4
 
         if package.root_dir:
             req = os.path.join(package.root_dir, package.source_url)
--- a/poetry/installation/pip_installer.py
+++ b/poetry/installation/pip_installer.py
@@ -225,7 +225,7 @@ class PipInstaller(BaseInstaller):
         args.append(req)
 
         try:
-            random_path = os.path.join(req, "pyproject.toml", str(uuid4()))
+            random_path = os.path.join(req, "pyproject.toml" + str(uuid4()))
             if pyproject.exists():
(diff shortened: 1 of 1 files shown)


finswimmer:
Whoooah, thanks a lot! :rocket: :guitar: 

This one was making me headache the last days. I have already found out, that it has something to do with `pip` but not what goes wrong exactly.

I'm not sure if it is a good idea to rename the `pyproject.toml` to something cryptic. There might be cases where the reverting to `pyproject.toml` doesn't work. So I would prefer something like `pyproject.toml.orig` or similar.

fin swimmer

wakemaster39:
happy to do an append or whatever makes the most sense here, I typically don't see guids in my code base so I would know what I am looking for. It is also that way to avoid the collision, but appending to the file name would also solve this.

wakemaster39:
This is a hack, I don't think this should be merged in but I will leave it here if I can't find a better alternative. I will link the information I have found in the issue and try and see if I can come up with a proper fix.

finswimmer:
Hello @wakemaster39 ,

yes this is a workaround and I would like to see a proper solution instead just like you. Thanks for your investigations and informations in the issue! I hope @sdispater has also some time to look at this.

What works for me in the meantime, is to make sure that within the venv the pip version is <19.

gvskalyan:
When would this be merged into master?

finswimmer:
Hey @gvskalyan,

it is unclear if this ever get merged, because this is "just" a hack and not a clean solution for the underlying problem. See also #1689.

@sdispater has to decide.

wakemaster39:
superseded by #1875

PR #1694, opened 2019-12-09
Show example of specifying version constraints

viq:
Show example of how to specify version constraints, since *nowhere* is it mentioned that `=` is required.


viq:
https://github.com/sdispater/poetry/pull/1657/ does it better.

Issue #1720, opened 2019-12-13
Install script details old API


SimonBiggs:
- [x] I am on the [latest](https://github.com/python-poetry/poetry/releases/latest) Poetry version.
- [x] I have searched the [issues](https://github.com/python-poetry/poetry/issues) of this repo and believe that this is not a duplicate.
- [x] If an exception occurs when executing a command, I executed it again in debug mode (`-vvv` option).

- **OS version and name**: Windows 10
- **Poetry version**: 1.0.0
- **Link of a [Gist](https://gist.github.com/) with the contents of your pyproject.toml file**: 

## Issue

The download and install script still states "You can uninstall at any time with `poetry self:uninstall`"

![image](https://user-images.githubusercontent.com/6559099/70766221-31ca9b80-1db1-11ea-869c-ee7f234c88b8.png)

finswimmer:
Hello @SimonBiggs ,

thanks a lot for the hint. Will be fixed with #1722 .

fin swimmer

SimonBiggs:
My pleasure :) thanks for all the hard work!!

PR #1722, opened 2019-12-13
fix: change `self:uninstall` to `self uninstall` in `get-poetry.py`

finswimmer:
Fixes: #1720 

# Pull Request Check List


sdispater:
Now that I think about it, the `self uninstall` command was never implemented (due to issues on Windows), so I wonder if we should just remove mentions to it until we actually go with the implementation.

finswimmer:
If it isn't implemented it should be removed :wastebasket: :grinning:

PR #1737, opened 2019-12-13
fix: removed hint for uninstall poetry, after installation.

finswimmer:
As discussed in https://github.com/python-poetry/poetry/pull/1722#issuecomment-565346594 this PR remove the hint about uninstalling poetry, because this is not fully implemented yet.

### DIFF
--- a/get-poetry.py
+++ b/get-poetry.py
@@ -226,6 +226,9 @@ It will add the `poetry` command to {poetry}'s bin directory, located at:
 {poetry_home_bin}
 
 {platform_msg}
+
+You can uninstall at any time with by executing this script with the --uninstall option,
+and these changes will be reverted.
 """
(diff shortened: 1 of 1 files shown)


get-poetry.py
```diff
@@ -226,10 +226,6 @@ def expanduser(path):
 {poetry_home_bin}
 
 {platform_msg}
-
-You can uninstall at any time with `poetry self uninstall`,
-or by executing this script with the --uninstall option,
```
sdispater: The `--uninstall` option of the `get-poetry.py` script is implemented so we can leave this line

finswimmer: You mean `poetry self uninstall` isn't implemented but `get-poetry.py --uninstall` is?

PR #1741, opened 2019-12-14
Docs update: Update basic-usage.md

Birdi7:
Documentation update: `pyproject.toml` which was generated using poetry v.1.0.0 is different from the one specified in the docs, so I updated it with the file from 1.0.0


neersighted:
Thanks for the PR, and sorry review was so long overdue. I've updated this for the latest master/`poetry init` output in #4737.

Issue #1336, opened 2019-08-28
Include exclude globs with whitelist pattern


dpgaspar:
Hi,

Making a POC for using poetry with https://github.com/apache/incubator-superset/pull/8079 but I'm having some issues

First superset has a complex directory structure, I'm struggling with the `include`, `exclude` parameters in glob, using `MANIFEST.in` we could whitelist for example:
```
recursive-exclude superset/static *
recursive-include superset/static/assets/branding *
recursive-include superset/static/assets/dist *
```

But using poetry toml we can't, how can I implement the same?

```
# excluding all assets possibilities ( dotted directories and files )
exclude = [
    "superset/assets/*",
    "superset/assets/.*",
    "superset/assets/**/*",
    "superset/assets/**/.*",
    "superset/assets/**/.*/*",
    "superset/assets/**/.*/**/*",
    "superset/assets/**/.*/**/.*",
    "tests/**"
]

include = [
    "superset/assets/branding/**",
    "superset/assets/dist/**",
]
```
With this pattern all superset/assets get excluded. Any advice?

rschmied:
This is also an issue for me. Like the @dpgaspar I want to exclude the superset of a directory and including **specific** files from that directory using
```
exclude = [
    "assets/sample/**/*"
]
include = [
    "assets/sample/subdir/one.yaml",
    "assets/sample/subdir/two.yaml"
]
```

This did work until recently and I suspect https://github.com/python-poetry/poetry/commit/12f981c53dcd50dac237337bdc6e90c32fe58236 has something to do with it. basically, since i upgraded from a previous 1.0.0 pre-release build, this was working as exected.

Not anymore with 1.0.0. Will downgrade for now.

finswimmer:
The reason for this is, that there is no check whether a file is explicit included or not. poetry builds a list of all files it can found and remove those from this list, that are excluded by `.gitignore` or via `exclude` in `pyproject.toml`.

rschmied:
Thanks. So, apparently this worked in previous builds and according to @dpgaspar this is also something that the MANIFEST supports.

How would this be done now with Poetry? I have lots of files in that sample directory but only want to include a minimal subset.

finswimmer:
Could you please tell me in which version this works?

finswimmer:
Sorry, I was wrong. poetry checks for explicit included files and doesn't exclude them. What might have changed is, that you have to start with the package name or a wildcard when giving the paths:

```
exclude = [
[...]

PR #1750, opened 2019-12-16
fix: don't exclude files that are explicit included in wheel (#1336)

finswimmer:
Explicit included files wasn't included in wheels if in excluding rule was there. This PR fixes it.

Fixes: #1336

# Pull Request Check List


sdispater:
At some point, I think we will need to go over the `masonry` module and improve it because I think we can make the code more clear especially regarding exclusions and inclusions.

In the end, explicit includes should always take precedence over excludes.

Issue #1479, opened 2019-10-18
Poetry breaks when there are spaces in path to Python


absassi:
- [x] I am on the [latest](https://github.com/sdispater/poetry/releases/latest) Poetry version.
- [x] I have searched the [issues](https://github.com/sdispater/poetry/issues) of this repo and believe that this is not a duplicate.
- [x] If an exception occurs when executing a command, I executed it again in debug mode (`-vvv` option).

- **OS version and name**: Windows 10
- **Poetry version**: 1.0.0b1
- **Link of a [Gist](https://gist.github.com/) with the contents of your pyproject.toml file**: 

## Issue

Cannot use Python interpreter at "C:\Program Files\Python38\python.exe", because Poetry makes a subprocess call using shell without properly quoting arguments. Note that, while I found this issue in Windows, any operating system is affected. It's just that spaces in paths are not so common in other systems. The above path is the default installation folder of Python for Windows when installed for all users.

```
>poetry env use -vvv "C:\Program Files\Python38\python.exe"
'C:\Program' is not recognized as an internal or external command,
operable program or batch file.

[EnvCommandError]
Command C:\Program Files\Python38\python.exe -c "import sys; print('.'.join([str(s) for s in sys.version_info[:3]]))" errored with the following return code 1, and output:


Traceback (most recent call last):
  File "C:\Users\asassi\.poetry\lib\poetry\_vendor\py3.6\clikit\console_application.py", line 132, in run
    status_code = command.handle(parsed_args, io)
  File "C:\Users\asassi\.poetry\lib\poetry\_vendor\py3.6\clikit\api\command\command.py", line 119, in handle
    status_code = self._do_handle(args, io)
  File "C:\Users\asassi\.poetry\lib\poetry\_vendor\py3.6\clikit\api\command\command.py", line 167, in _do_handle
    return getattr(handler, handler_method)(args, io, self)
  File "C:\Users\asassi\.poetry\lib\poetry\_vendor\py3.6\cleo\commands\command.py", line 92, in wrap_handle
    return self.handle()
  File "C:\Users\asassi\.poetry\lib\poetry\console\commands\env\use.py", line 24, in handle
    env = manager.activate(self.argument('python'), poetry.file.parent, self._io)
  File "C:\Users\asassi\.poetry\lib\poetry\utils\env.py", line 177, in activate
    raise EnvCommandError(e)
```

[...]

PR #1774, opened 2019-12-20
Correctly handle spaces in command executable

### LINKED ISSUE #1747: `poetry env use` on Windows needs to deal with spaces in path or take a command line list
- [ ] ~~If an exception occurs when executing a command, I executed it again in debug mode (`-vvv` option).~~

- **OS version and name**: Windows 10 1909 x64
- **Poetry version**: 1.0.0
- **Link of a [Gist](https://gist.github.com/) with the contents of your pyproject.toml file**: N/A

## Issue

On Windows, Python is typically installed into `C:\Program Files`, i.e. `C:\Program Files\Python38\python.exe`. Attempting to make poetry use it rather than the Python it is installed for results in:

```
'C:\Program' is not recognized as an internal or external command,
operable program or batch file.

[EnvCommandError]
Command C:\Program Files\Python38\python.exe -c "import sys; print('.'.join([str(s) for s in sys.version_info[:3]]))" errored with the following return 
code 1, and output:
```

In this case, the path must be treated as `["C:\\Program Files\\Python38\\python.exe"]`.

Python for Windows also comes with the `py` program for selecting which installed Python gets called. You can call it like `py` and it will start the latest Python, or like `py -2.7-32` to start that other thing. In that case, the string must be interpreted as `["py", "-2.7-32"]`

frostming:
# Pull Request Check List

Fix #1747 

use `list_to_shell_command` to wrap executable.

### DIFF
--- a/poetry/utils/env.py
+++ b/poetry/utils/env.py
@@ -190,7 +190,7 @@ class EnvManager(object):
         try:
             python_version = decode(
                 subprocess.check_output(
-                    " ".join(
+                    list_to_shell_command(
                         [
                             python,
                             "-c",
@@ -427,7 +427,7 @@ class EnvManager(object):
         try:
             python_version = decode(
                 subprocess.check_output(
-                    " ".join(
+                    list_to_shell_command(
                         [
                             python,
                             "-c",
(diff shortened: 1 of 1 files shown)


absassi:
This also fixes #1479.

PR #1786, opened 2019-12-23
Improve detection of installed packages

### LINKED ISSUE #1612: All dependencies are always updated
- [x] If an exception occurs when executing a command, I executed it again in debug mode (`-vvv` option). [N/A]

- **OS version and name**: macOS Catalina (10.15.1)
- **Poetry version**: 1.0b4, 1.0b5, 1.0b6
- **Link of a [Gist](https://gist.github.com/) with the contents of your pyproject.toml file**: https://gist.github.com/wylee/53e74a3413df25e1b594803e193d40d4

## Issue

Since 1.0b4, when running `poetry update`, all of my dependencies are always updated. I've tried clearing the cache and deleting `poetry.lock`.

#1582 is similar but seems to affect only project that use a private package index.

### LINKED ISSUE #1711: poetry re-installs up-to-date packages when virtualenvs.create is false
- [x] I am on the [latest](https://github.com/python-poetry/poetry/releases/latest) Poetry version. (`1.0.0b9`)

- **OS version and name**: Debian buster
- **Poetry version**: `1.0.0b9`

## Issue
Thanks for the great work! With current releases we experience that already up-to-date packages (with the correct versions) are reinstalled by poetry, as it tries to update them. This seems to happen because we don't use poetry's virtualenv feature (using `poetry config virtualenvs.create false`).

Here's a part of the `poetry install` output, when running it the second time:
```
  - Updating humanfriendly (4.18 /usr/local/lib/python3.6/dist-packages -> 4.18)
  - Updating imageio (2.6.1 /usr/local/lib/python3.6/dist-packages -> 2.6.1)
  - Updating importlab (0.5.1 /usr/local/lib/python3.6/dist-packages -> 0.5.1)
```

It seems that dist-packages can never match the version, so poetry tries to upgrade them, even when the version actually matches.

Output of `poetry debug info`:
```
Poetry
Version: 1.0.0b9
Python:  3.6.8

Virtualenv
Python:         3.6.8
Implementation: CPython
Path:           NA

System
Platform: linux
OS:       posix
Python:   /usr
```

sdispater:
## Pull Request Check List

In version 1.0.0, we changed the way the installed packages are detected to improve the detection of the type of packages (directory vs git dependencies for instance). However, this introduced issues where Poetry always reinstalled the packages because it was detecting the wrong type of package or because it was not retrieving it from the correct directory. This PR should fix most of these issues.

Fixes #1711
Fixes #1612

### DIFF
--- a/poetry/utils/env.py
+++ b/poetry/utils/env.py
@@ -703,6 +703,7 @@ class Env(object):
 
         self._marker_env = None
         self._pip_version = None
+        self._site_packages = None
 
     @property
     def path(self):  # type: () -> Path
@@ -758,20 +759,25 @@ class Env(object):
 
     @property
     def site_packages(self):  # type: () -> Path
-        # It seems that PyPy3 virtual environments
-        # have their site-packages directory at the root
-        if self._path.joinpath("site-packages").exists():
-            return self._path.joinpath("site-packages")
+        if self._site_packages is None:
+            site_packages = []
+            dist_packages = []
+            for entry in self.sys_path:
+                entry = Path(entry)
+                if entry.name == "site-packages":
+                    site_packages.append(entry)
+                elif entry.name == "dist-packages":
+                    dist_packages.append(entry)
 
-        if self._is_windows:
-            return self._path / "Lib" / "site-packages"
+            if not site_packages and not dist_packages:
+                raise RuntimeError("Unable to find the site-packages directory")
 
-        return (
-            self._path
-            / "lib"
-            / "python{}.{}".format(*self.version_info[:2])
-            / "site-packages"
-        )
+            if site_packages:
+                self._site_packages = site_packages[0]
+            else:
+                self._site_packages = dist_packages[0]
+
+        return self._site_packages
 
     @property
     def sys_path(self):  # type: () -> List[str]
@@ -1114,6 +1120,7 @@ class MockEnv(NullEnv):
         os_name="posix",
         is_venv=False,
         pip_version="19.1",
+        sys_path=None,
         **kwargs
     ):
         super(MockEnv, self).__init__(**kwargs)
@@ -1124,6 +1131,7 @@ class MockEnv(NullEnv):
         self._os_name = os_name
         self._is_venv = is_venv
         self._pip_version = Version.parse(pip_version)
+        self._sys_path = sys_path
 
     @property
     def version_info(self):  # type: () -> Tuple[int]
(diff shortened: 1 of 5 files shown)


poetry/utils/env.py
```diff
@@ -758,20 +759,25 @@ def pip_version(self):
...
+                elif entry.name == "dist-packages":
+                    dist_packages.append(entry)
+
+            if not site_packages and not dist_packages:
+                raise RuntimeError("Unable to find the site-packages directory")
+
+            if site_packages:
+                self._site_packages = site_packages[0]
+            else:
+                self._site_packages = dist_packages[0]
```
finswimmer: I don't know where the `site_packages` property is used, so it isn't obvious to me, why only the first site_package is stored and why `dist-packages` are never stored when site_packages where found.

sdispater: The `site_packages` property is used in the `InstalledRepository` class (see https://github.com/python-poetry/poetry/blob/master/poetry/repositories/installed_repository.py#L32).

We only need the "site-packages" directory where `pip` will install the dependencies which is, as far as I know, the `site-packages` in virtual environments and most OS and `dist-packages` for Debian-based distributions.

finswimmer: Ah, ok I got it. I wonder if something with [sysconfig](https://docs.python.org/3/library/sysconfig.html) would be more reliable? But as far as I can see, we must be able to detect if we are inside a virtual environment and of we are on a posix or nt system, to get the correct path (`purelib` seems to be the one we are interested in), are we?

sdispater: Thanks for pointing me to `sysconfig`! Do you know if there are some systems where `purelib` and `platlib` are different? If there is none we could indeed use `purelib` and avoid the heuristics we are currently doing.

finswimmer: I'm not sure. I've just discovered that module as well. There is something in [PEP427](https://www.python.org/dev/peps/pep-0427/#what-s-the-deal-with-purelib-vs-platlib) about the difference.

sdispater: In the end, it seems that all installed packages will reside in `purelib` and/or `platlib`, so to get populate the `InstalledRepository` we can just check both, I think. I'll test it out and see where it goes.

sdispater: After testing it out it seems that using `sysconfig` won't help us. While it works in some cases, for Debian-based systems, for instance, it won't return the directory we need. We could use `site.getsitepackages()` for non-virtual environments but we would need to find a way to replicate it for virtual environments which would lead to using the same heuristic this PR already uses.

I think we can leave the PR as-is for now since I tested it on Ubuntu, MacOS and Windows without any issue and it fixes the issues I mentioned above.

Issue #1697, opened 2019-12-09
Github SSH String and SdistBuilder Incompatible


wakemaster39:
- [x] I am on the [latest](https://github.com/sdispater/poetry/releases/latest) Poetry version.
- [x] I have searched the [issues](https://github.com/sdispater/poetry/issues) of this repo and believe that this is not a duplicate.
- [x] If an exception occurs when executing a command, I executed it again in debug mode (`-vvv` option).

- **MaxOS 10.14**: 
- **1.0.0b8**: 

## Issue
When working with git dependency such as `git@github.com:sdispater/poetry.git`. Working with poetry everything is totally fine things install and function properly and it makes it seem like you have configured poetry correctly.

This falls apart when you try and install that package in another repo. Basically the `SdistBuilder` outputs a string like `git+git@github.com:sdispater/poetry.git@master`. What it actually needs to be is `git+ssh://git@github.com/sdispater/poetry.git`.

This can easily be fixed in the package by changing the git url to be `ssh://git@github.com/sdispater/poetry.git`. The issue is there is no way to know that this is a problem until you are trying to consume your work and the terrible output messages from pip when you try and install this will leave you in debugging hell especially since poetry deletes the `setup.py` file it generates so you can't view it to rerun it.

There are two different ways to handle this problem as I see it, the first would be to make it so that poetry does not accept git urls that are not compliant with [PEP-508](https://www.python.org/dev/peps/pep-0508/). 

The second option is to upgrade `VCSDependency` so that it detects urls in this format and converts them to a compliant format when `dependency.to_pep_508()` is called. 

I personally prefer the second because it is easier to copy and paste out of GitHub and would hopefully make the friction less when new users are adopting poetry.

I am happy to implement either, I just need to know the preferred direction.

finswimmer:
Hey @wakemaster39 ,

I'm not sure if I understand your problem. Could you please provide a step-by-step manual to see where your problem arises?

Thanks a lot!

fin swimmer

wakemaster39:
1. Create a project with a git dependency, add it using the ssh URL copied from Github such as `git@github.com:pallets/flask.git`
2. Build your new package `poetry build`
3. Open the outputted tar.gz and edit the setup.py file
4. view the `install_requires` section and see the illegal line similar to `flask @ git+git@github.com:pallets/flask.git@master`

The legal format would be `flask @ git+ssh://git@github.com/pallets/flask.git@master`

[...]

PR #1799, opened 2019-12-28
create valid PEP508 string for git dependencies with ssh (#1697)

finswimmer:
Fixes: #1697 

# Pull Request Check List

### DIFF
--- a/poetry/packages/vcs_dependency.py
+++ b/poetry/packages/vcs_dependency.py
@@ -1,3 +1,5 @@
+from poetry.vcs import git
+
 from .dependency import Dependency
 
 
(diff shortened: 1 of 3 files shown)


poetry/packages/vcs_dependency.py
```diff
@@ -1,3 +1,5 @@
+from poetry.vcs import git
```
matemax:
```suggestion
from ...vcs import git
```

finswimmer: Hello @matemax,

I'm not a big fan of relative imports. "Explicit is better than implicit." and is better to read. Absolute imports are also used (most if the time) in the poetry code.

matemax: Hello @finswimmer ,
It is ony suggestion. I try use these code for me and noticed relative and not relative import in one file.

sdispater: The rule I have personally is if the import comes from the same module/package, it should be a relative import otherwise it should be an absolute import. An example of this is https://github.com/python-poetry/poetry/blob/master/poetry/masonry/builders/builder.py

PR #1810, opened 2019-12-31
Fix export git dependencies with `-e ` only if develop mode is on

shenek:
# Pull Request Check List

Currently `pip_installer` behaves differently than `export`.
If `develop` is `False` it doesn't add `-e ` to requirement.
But `export` always adds `-e ` to requirement.

see https://github.com/python-poetry/poetry/blob/develop/tests/installation/test_pip_installer.py#L65

This patch updates the `export` behavior so it matches the behavior of `pip_installer`.

closes: #1816

### DIFF
--- a/docs/docs/dependency-specification.md
+++ b/docs/docs/dependency-specification.md
@@ -103,6 +103,15 @@ flask = { git = "https://github.com/pallets/flask.git", rev = "38eb5d3b" }
 numpy = { git = "https://github.com/numpy/numpy.git", tag = "v0.13.2" }
 ```
 
+In some cases you may also want to install the dependency in non-editable mode.
+This can be achieved by setting `develop` key to `false`.
+
+```toml
+[tool.poetry.dependencies]
+requests = { git = "https://github.com/kennethreitz/requests.git", tag = "v2.22.0", develop = false }
+```
+
+
 ## `path` dependencies
 
 To depend on a library located in a local directory or file,
--- a/poetry/json/schemas/poetry-schema.json
+++ b/poetry/json/schemas/poetry-schema.json
@@ -326,6 +326,10 @@
                     "items": {
                         "type": "string"
                     }
+                },
+                "develop": {
+                    "type": "boolean",
+                    "description": "Whether to install the dependency in development mode."
                 }
             }
         },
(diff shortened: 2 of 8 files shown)


finswimmer:
Hello @shenek,

thanks a lot for your contribution. We really appreciate it :+1: 

Unfortunately I believe more work needs to be done, to implement what you suggest.

At the moment git dependencies are always installed as editable. A "develop=False" in the `pyproject.toml` isn't implemented. This must be added to the [schema](poetry/json/schemas/poetry-schema.json).

The information must be transported to the `poetry.lock` file. See this PR (https://github.com/python-poetry/poetry/pull/1656) to get an idea of how it works.

If it's not an editable package, you must calculate the hash and add these value to the `poetry.lock`.

I don't know yet if there are more changes needed, that the git dependency get installed as a "normal" package.

fin swimmer

PS: While I'm writing this I noticed you've opened an Feature Request for this (https://github.com/python-poetry/poetry/issues/1816) :smile:

shenek:
Hello @finswimmer ,

thanks for the hints. I'll try to expand this pull request. The original intention of this PR was only to expand exports. However later I realized that there are some other parts missing so I created an issue.

Cheers.

shenek:
Hi @finswimmer ,

I added a couple of changes. From my point of view it seems to be working. I can install `{"git" = "...", "develop" = false}` dependencies and the dependencies seems to be installed properly. However I'm still missing a few of points.

Firstly I'm not sure how to deal with the hashes. Should `content-hash` be placed into `metadata` if `develop=false`? Or meant something else?

And secondly I don't know how to properly test the added feature (especially the lock file generation part).

Cheers.

kalemas:
Is there something i missed trying to install from that branch?
https://pastebin.com/EFZM8LZQ

shenek:
> Is there something i missed trying to install from that branch?
> https://pastebin.com/EFZM8LZQ

This PR is based on devel branch of poetry which might have removed support for python2. So to make it work you should try to use it with python3.

Secrus:
This is no longer valid PR. Export was moved to [`poetry-plugin-export`](https://github.com/python-poetry/poetry-plugin-export).

Issue #755, opened 2018-12-20
Dependency rooted in a subfolder of a GIT repository


sandrinr:
- [x] I have searched the [issues](https://github.com/sdispater/poetry/issues) of this repo and believe that this is not a duplicate.

## Feature request

Currently, as far as I can figure, it is impossible to use a GIT repository as a dependency source where the `setup.py` or `pyproject.toml` file is not in the root directory but in some sub-folder of the repository.

E.g. GIT repo structure:
* GIT root
  * pythonmodule
    * setup.py / pyproject.toml
    * module1.py
    * module2.py
  * javamodule
    * pom.xml
  * cmodule
    * Makefile

In this example I would like to be able to point poetry to the `pythonmodule` directory inside this repo, for example using `poetry add mydep --git https://github.com/foo/bar.git --path path/to/mydep`. I don't think this can be accomplished via a separate repository definition, however, the documentation is a bit scarce on that topic.

bwdeng:
Same here, i'm able to install a git dependency using pip with its [VCS support](https://pip.readthedocs.io/en/stable/reference/pip_install/#vcs-support)

```
pip install "git+https://github.com/nmslib/hnswlib.git@bb2a1d1#egg=hnswlib&subdirectory=python_bindings"
```

It would be nice if we can do it using `poetry add`.

zbarnett:
I would love to switch to Poetry and it looks like this is the only thing preventing me. Very excited to see this feature is in progress!

finswimmer:
Ready for testing :)

velovix:
@finswimmer Thank you for your work on this! Is there any way I can help with getting your PRs ready for review?

TheButlah:
Was this functionality implemented? My team relies on poetry and we just started breaking some of our code up into a separate git repo - we would need this to continue to use poetry

sinoroc:
> Was this functionality implemented? My team relies on poetry and we just started breaking some of our code up into a separate git repo - we would need this to continue to use poetry

As far as I can tell, it has been implemented, but not merged nor released. This still-in-progress implementation is here: https://github.com/python-poetry/poetry/pull/1822.

finswimmer:
It isn't fully implemented until now. A lot has changed since I wrote #1822. This PR is now splitted in https://github.com/python-poetry/poetry-core/pull/9 and https://github.com/python-poetry/poetry/pull/2242 But there is still a lot to do. I cannot tell when this will be ready.

[...]

PR #1822, opened 2020-01-03
support git subfolder (#755)

finswimmer:
This PR implements the possibility to add git dependencies, where the package is located within a subdirectory  of the repository.

The path to the subdirectory can be specified as follows:

```
poetry add "git+https://github.com/demo/project_in_subdirectory.git?subdirectory=pyproject-demo"
```

Where `pyproject-demo` is a subfolder relative to the root after checkout the repository. This can be combined with the revision tag:

```
poetry add "git+https://github.com/demo/project_in_subdirectory.git#dev?subdirectory=pyproject-demo"
```

To specify the subdir in the `pyproject.toml` a new argument `subdirectory` was introduced:

```
mypackage = {git = "https://github.com/demo/project_in_subdirectory.git", rev = "dev", subdirectory = "pyproject-demo"}
```

To implement this feature is was necessary  to checkout the repository first to a temporary folder and copy the desired folder to it's destination. For this a new `DownloadTmpDir` class was implemented, which avoids downloading from the same git source multiple times, during run time.

I would like to get some feed back from people, who can test this on a real world use case. Especially if this all works as expected.

Implements: #755 

# Pull Request Check List

### DIFF
--- a/poetry/puzzle/provider.py
+++ b/poetry/puzzle/provider.py
@@ -33,7 +33,6 @@ from poetry.utils._compat import PY35
 from poetry.utils._compat import OrderedDict
 from poetry.utils._compat import Path
 from poetry.utils._compat import urlparse
-from poetry.utils.cache import DownloadCache
 from poetry.utils.env import EnvCommandError
 from poetry.utils.env import EnvManager
 from poetry.utils.env import VirtualEnv
@@ -41,6 +40,7 @@ from poetry.utils.helpers import parse_requires
 from poetry.utils.helpers import temporary_directory
 from poetry.utils.inspector import Inspector
 from poetry.utils.setup_reader import SetupReader
+from poetry.utils.temp import DownloadTmpDir
 from poetry.utils.toml_file import TomlFile
 from poetry.vcs.git import Git
 from poetry.version.markers import MarkerUnion
@@ -186,7 +186,7 @@ class Provider:
         if vcs != "git":
             raise ValueError("Unsupported VCS dependency {}".format(vcs))
 
-        tmp_dir = Path(DownloadCache.mkcache(url, prefix="pypoetry-git"))
+        tmp_dir = Path(DownloadTmpDir.mkcache(url, prefix="pypoetry-git"))
 
         git = Git()
 
--- a/poetry/puzzle/provider.py
+++ b/poetry/puzzle/provider.py
@@ -169,6 +169,7 @@ class Provider:
             dependency.source,
             dependency.reference,
             name=dependency.name,
+            subdirectory=dependency.subdirectory,
         )
 
         for extra in dependency.extras:
@@ -182,8 +183,8 @@ class Provider:
 
     @classmethod
     def get_package_from_vcs(
-        cls, vcs, url, reference=None, name=None
-    ):  # type: (str, str, Optional[str], Optional[str]) -> Package
+        cls, vcs, url, reference=None, name=None, subdirectory=None
+    ):  # type: (str, str, Optional[str], Optional[str], Optional[str]) -> Package
         if vcs != "git":
             raise ValueError("Unsupported VCS dependency {}".format(vcs))
 
@@ -201,7 +202,13 @@ class Provider:
 
             revision = git.rev_parse(reference, tmp_dir).strip()
 
-            package = cls.get_package_from_directory(tmp_dir, name=name)
+            if subdirectory is None:
+                package = cls.get_package_from_directory(tmp_dir, name=name)
+            else:
+                package = cls.get_package_from_directory(
+                    tmp_dir / subdirectory, name=name
+                )
+                package.source_subdirectory = Path(subdirectory).as_posix()
 
             package.source_type = "git"
             package.source_url = url
--- a/poetry/puzzle/provider.py
+++ b/poetry/puzzle/provider.py
@@ -34,6 +34,7 @@ from poetry.utils._compat import PY35
 from poetry.utils._compat import OrderedDict
 from poetry.utils._compat import Path
 from poetry.utils._compat import urlparse
+from poetry.utils.cache import DownloadCache
 from poetry.utils.env import EnvCommandError
 from poetry.utils.env import EnvManager
 from poetry.utils.env import VirtualEnv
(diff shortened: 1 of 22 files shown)


Red-Eyed:
I'm very interesting to see this changes in release.

approximator:
Hi @finswimmer,
Thank you for this feature!
I tested it on my project with private deps using a command like:

```
poetry add "git+ssh://git@reposerver.com:12345/prj_group/common_libs.git?subdirectory=packages/some_package"
```

And it works just fine.

poetry/utils/exporter.py
```diff
@@ -81,6 +81,9 @@ def _export_requirements_txt(
                 line = "-e git+{}@{}#egg={}".format(
                     package.source_url, package.source_reference, package.name
                 )
+                if package.source_subdirectory:
+                    line += "?subdirectory={}".format(package.source_subdirectory)
```
abn:
```suggestion
                    line += "&subdirectory={}".format(package.source_subdirectory)
```

See https://pip.readthedocs.io/en/stable/reference/pip_install/#vcs-support

poetry/puzzle/provider.py
```diff
@@ -182,34 +181,34 @@ def search_for_vcs(self, dependency):  # type: (VCSDependency) -> List[Package]
...
-    ):  # type: (str, str, Optional[str], Optional[str]) -> Package
+        cls, vcs, url, reference=None, name=None, subdirectory=None
+    ):  # type: (str, str, Optional[str], Optional[str], Optional[str]) -> Package
         if vcs != "git":
             raise ValueError("Unsupported VCS dependency {}".format(vcs))
 
-        tmp_dir = Path(
-            mkdtemp(prefix="pypoetry-git-{}".format(url.split("/")[-1].rstrip(".git")))
-        )
+        tmp_dir = Path(DownloadTmpDir.mkcache(url, prefix="pypoetry-git"))
```
abn: This this is not longer required, since this has changed:

https://github.com/python-poetry/poetry/blob/063d19aa7d5e9b8601a2658237dfc70995410c7e/poetry/puzzle/provider.py#L190-L192

poetry/vcs/git.py
```diff
@@ -14,7 +14,8 @@
     "port": r"\d+",
     "path": r"[\w~.\-/\\]+",
     "name": r"[\w~.\-]+",
-    "rev": r"[^@#]+",
+    "rev": r"[^@#?]+",
+    "subdir": r"[\w\-/\\]+",
```
abn: A bit unsure about this one. As per the current poetry docuemtnation the argument for `poetry add` when providing a url needs to be a valid git url, which the python vcs dependency url us not. In the curent version we support `<url>#<rev>` (valid git) which is not adherant to pip specification (this seems to be standardised by PEP 610). It would seem that this is not fully compliant to either specification. The use of query string syntax might also vendor dependant (github vs gitlab).

Do you think this could be confusing? And if so, should we pick one and stick with it?

poetry/utils/exporter.py
finswimmer: Hey @abn,
thanks a lot for your comments. You're right here :+1:

poetry/puzzle/provider.py
finswimmer: Hm, don't know if you link the correct lines?

The reason for the way I create the temp folder for cloning the repository is, that at the moment poetry clone the repository twice during install. The first time to resolve the package version and second time when installing. For the case, let's say I like to add 3 packages from subdirectory of the same repo, poetry would clone this repo 6 times. With the newly added class `DownloadTmpDir` it will clone the repository only once during runtime and reuse it for every package that is clone from the same repository.

poetry/vcs/git.py
finswimmer: When starting with this PR I thought about a new command line argument. I refuse it because it will lead to more changes in the code compared with passing it along with the URL. Second, I was inspired by the way pip is doing this. I guess most new users would find this intuitive.

Other opinions are welcome. I will have to work on this PR the next days, because of the changes introduced by poetry_core now.

abn: I like the idea for sure, I am all for consistency with PEP 610. The concern right now is that this change would mean we will neither conform to that nor git's url spec. We should probably make a decision how how to handle this.

More changes might not be a bad thing if it makes things clearer. Additionally, we can definitely tackle exposing this via a cli option in a later PR. Right now, I guess we just need to decide which url spec to follow, we can accept both too - but this would mean some assumptions in error handling during parsing.

There are definitely pros and cons for following either format. From a cli perspective having a standard git url and options that map to the toml spec will reduce the cognitive load on end users as it is explcit (more intuitive?) for someone who is familiar with poetry's config I feel (eg: `poetry add <git url> --subdirectory=x --revision=x` will map cleanly to dependency specification in toml). On the other hand, having it be the same will mean, atleast for new users as you said, be more intuitive.

poetry/puzzle/provider.py
abn: Yeah, you can ignore the snippet. I understand the intention here now.

Do you know if there was a reason why we need a temporary clone here at all? Can we for example, make this not runtime specific? Clone a repo to a deterministic source directory (maybe a name containing `hash(source_url)`) and switch revisions instead of a new clone? This should avoid cloing across runtimes - might be advantageous in non subdirectory use case as well. If so, I would say it might even be a separate PR independant from the subdirectory change as it has broader impact.

poetry/vcs/git.py
abn: Thinking more about it; we should probably consider enabling both by doing the following.

1. For users who use `poetry add vcs+protocol://repo_url[@rev]/#egg=pkg&subdirectory=pkg_dir` we simply convert this to the appropriate inline table in the toml `pkg = { vcs = "protocol://repo_url", rev = "rev", subdirectory = "pkg_dir"}`. Any cli options `--subdirectory | --revision` overrides values parsed from vcs url. 
2. For users who use `poetry add vcs+protocol://repo_url[#rev]` we maintain current behavior with the possibility of using new cli options.
3. Error out with recommendation if the url is explicitly added to the toml file and is not a valid remote url.

(1) would be really be important for cases where the pip VCS dependency url is copied over.

LongBeachHXC:
how we looking on this one? This would be extremely useful for my use case.

remram44:
Is this done or wontfix? Related PRs have been closed or are still open.

ashnair1:
This commit (python-poetry/poetry-core#192) was recently merged into poetry-core's master so this issue might be solved in 1.2

PR #1834, opened 2020-01-05
WIP: Add no-install param to poetry add

adisbladis:
# Pull Request Check List

### DIFF
--- a/poetry/console/commands/add.py
+++ b/poetry/console/commands/add.py
@@ -39,6 +39,11 @@ class AddCommand(EnvCommand, InitCommand):
             None,
             "Output the operations but do not execute anything (implicitly enables --verbose).",
         ),
+        option(
+            "no-install",
+            None,
+            "Add dependency without installing it in the virtual env.",
+        ),
     ]
 
     help = """The add command adds required packages to your <comment>pyproject.toml</> and installs them.
@@ -142,6 +147,7 @@ If you do not specify a version constraint, poetry will choose a suitable one ba
         )
 
         installer.dry_run(self.option("dry-run"))
+        installer.no_install(self.option("no-install"))
         installer.update(True)
         installer.whitelist([r["name"] for r in requirements])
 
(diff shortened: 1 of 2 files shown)


sdispater:
Since there is no feature request issue related to this: which use case does this try to solve?

I don't see the value of this feature and seems more like and edge but I might be wrong.

adisbladis:
It is a bit of a "strange" use-case. I want to let poetry deal with dependency solving _only_ and let https://github.com/nix-community/poetry2nix create/load the environments.

juhoautio:
I have a slightly different kind of use case for this feature. I have a project generator that wraps poetry to create a structure that extends poetry's defaults. For this I'm using `poetry new`, followed by `poetry add` to add some additional default dependencies. However, `poetry add` also creates the virtual environment and installs the dependencies, which takes some time.

On the other hand, the interactive `poetry init` allows specifying additional dependencies without installing them. But it's not ideal to wrap the interactive command when further input is not wanted by the user.

To explain the use case a bit more, workflow is that a user could run `mygenerator new` quickly, then edit the `pyproject.toml` manually to for example remove unwanted default dependencies, and run `poetry install` once done with the manual editing. Also, as `poetry init` & `poetry new` don't create the virtual environment, I would like my generator to be able to skip it.

My use case could be cleanly implemented with any of these changes in poetry:
- Add a `--no-install` option to `poetry add`
- Add options `--dependencies` and `--dev-dependencies` to `poetry new` (this is my favorite, because why not, when `init` already supports these flags)
- Add option `--no-interactive` to `poetry init`
- Add a function in poetry lib that allows easily modifying `pyproject.toml` so that it preserves the original layout of `poetry new`

tbenst:
I found a workaround...if you ctrl-c at the right time, the appropriate lines are added to `pyproject.toml`, and you can subsequently do `poetry lock`! LOL no-install is better...

adisbladis:
Poetry now has a `--lock` argument which seems to achieve the exact same thing I was aiming for.

Issue #897, opened 2019-02-16
export requirements.txt - local dependency file exported as editable


visciang:
- **OS version and name**: Linux
- **Poetry version**: 1.0.0a2

## Issue
`poetry export --without-hashes --format=requirements.txt` ouputs local file dependencies as "editable":

```
-e local/path/package.whl
```
This is reported as an error when parsed by `pip install`:

```
local/path/package.whl should either be a path to a local project
or a VCS url beginning with svn+, git+, hg+, or bzr+
```

The correct format should not include the `-e` part (ref. https://pip.pypa.io/en/stable/reference/pip_install/#requirements-file-format).

NOTE:
This is covered in `test_exports.py` https://github.com/sdispater/poetry/blob/develop/tests/utils/test_exporter.py#L353

gizmo385:
Bumping this issue. We're noticing a similar problem and it's causing problems for our serverless deployments. Pip is complaining whenever we try to install from the `requirements.txt` file that `poetry export` generates. Is there a fix for this issue?

severinsimmler:
+1

riclima:
I've come across something similar. My export came with the editable flag for a dependency using a Git url. Had to remove the flag before I could deploy it using Pip in production. Using Poetry 1.0.0b2.

iwoloschin:
I think this line here needs to be changed to remove the `-e`?

https://github.com/sdispater/poetry/blob/master/poetry/utils/exporter.py#L84

I see a few other `-e`'s scattered through the file though, so I'm not sure if there's something else that needs to be replaced as well?

In any case, a quick workaround is to use `sed` to strip out the offending `-e`s.
```
poetry export -f requirements.txt | sed 's/-e //'
```

filipelbc:
I'm having the same issue here.

finswimmer:
Thanks for reporting!

#1840 should fix it.

fin swimmer

JulianFerry:
Unless I'm mistaken this still adds the -e tag? I'm calling `poetry export --without-hashes -f requirements.txt > requirements.txt` with `Poetry version 1.0.9`

finswimmer:
Hello @JulianFerry ,

can you show an example of your output where you think it's still wrong?

fin swimmer

JulianFerry:
As a full reproducible example, I did:
```
poetry new dependency
poetry new proj
cd proj
```
then added this to my pyproject.toml:
```
dependency = { path = "../dependency" }
```
then:
```
[...]

PR #1840, opened 2020-01-07
fix: don't export local file dependency as editable (#897)

finswimmer:
At the moment local file dependencies are export as editable. This is not valid.

Fixes: #897

# Pull Request Check List

### DIFF
--- a/poetry/utils/exporter.py
+++ b/poetry/utils/exporter.py
@@ -94,7 +94,7 @@ class Exporter(object):
                 dependency.marker = package.marker
 
                 line = "{}".format(package.source_url)
-                if package.develop:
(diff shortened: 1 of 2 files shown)


poetry/utils/exporter.py
```diff
@@ -94,7 +94,7 @@ def _export_requirements_txt(
                 dependency.marker = package.marker
 
                 line = "{}".format(package.source_url)
-                if package.develop:
+                if package.develop and package.source_type in ["directory", "url"]:
```
sdispater: We only want directory packages here, I think. As far as I know url packages cannot be editable.

finswimmer: Good catch :+1: Fixed it.

Issue #1668, opened 2019-12-04
Add Python shebang to get-poetry.py script


n8sty:
- [x] I have searched the [issues](https://github.com/sdispater/poetry/issues) of this repo and believe that this is not a duplicate.
- [x] I have searched the [documentation](https://poetry.eustace.io/docs/) and believe that my question is not covered.

## Feature Request
I use Poetry as part of a few different CI flows. In some of these I put (and cache) `get-poetry.py` into `/usr/local/bin/` ie:

```
sudo curl -sSL --output /usr/local/bin/get-poetry https://raw.githubusercontent.com/sdispater/poetry/master/get-poetry.py
```

Since `get-poetry` doesn't have a shebang, it's not executable. This isn't a big deal, it's not too difficult to call `python /usr/local/bin/get-poetry` or add an appropriate shebang at the top of the file, but why not just add `#!/usr/bin/env python` as a shebang to the file to support this use case?

finswimmer:
Hello @n8sty,


the `get-poetry` script is now deprecated in favor of the new `install-poetry` script. This is why I close your ticket.

fin swimmer

PR #1878, opened 2020-01-13
get-poetry.py fallback to standard executables

### LINKED ISSUE #1877: poetry bootstrap is unable to fallback on `python3` executable
The `get-poetry.py` script relies on `python` being available in the shell.

It should be a little more flexible, such as allowing `python3` and supporting [PEP 397](https://www.python.org/dev/peps/pep-0397/).

jonapich:

I wasn't sure how to spin new tests here; I believe the get-poetry.py script isn't part of the test suite, is that possible?

This makes the installation more flexible by allowing `python3` on linux (when `python` cannot be found; typical in a lot of 2 -> 3 migration scenarios).

On windows, we add some basic support for [PEP 397](https://www.python.org/dev/peps/pep-0397/) by leveraging the `py.exe` launcher. Since Windows uses the .bat file to launch the .py file, I removed the shebang from the windows launcher so that we don't end up writing `/usr/bin/env py.exe -3` :sweat_smile: 

fixes #1877

### DIFF
--- a/get-poetry.py
+++ b/get-poetry.py
@@ -591,7 +591,7 @@ class Installer:
 
         for executable in allowed_executables:
             try:
-                subprocess.check_call(executable + "--version", shell=True)
+                subprocess.check_call(executable + " --version", shell=True)
             except subprocess.CalledProcessError:
                 continue
 
--- a/get-poetry.py
+++ b/get-poetry.py
@@ -585,22 +585,35 @@ class Installer:
                 gz.close()
 
     def _which_python(self):
+        """Decides which python executable we'll embed in the launcher script."""
         allowed_executables = ["python", "python3"]
         if WINDOWS:
             allowed_executables += ["py.exe -3", "py.exe -2"]
 
+        version_matcher = re.compile(r"^Python (?P<major>\d+)\.(?P<minor>\d+)\..+$")
+        fallback = None
         for executable in allowed_executables:
             try:
-                subprocess.check_call(executable + " --version", shell=True)
+                raw_version = subprocess.check_output(
+                    executable + " --version", stderr=subprocess.STDOUT, shell=True
+                ).decode("utf-8")
             except subprocess.CalledProcessError:
                 continue
 
-            return executable
+            match = version_matcher.match(raw_version.strip())
+            if match and int(match.groupdict()["major"]) >= 3:
+                return executable
+            if fallback is None:
+                # keep this one as the fallback; it was the first valid exec we found.
+                fallback = executable
 
-        raise RuntimeError(
-            "No python executable found in shell environment. Tried: "
-            + str(allowed_executables)
-        )
+        if fallback is None:
+            raise RuntimeError(
+                "No python executable found in shell environment. Tried: "
+                + str(allowed_executables)
+            )
+
+        return fallback
 
     def make_bin(self):
         if not os.path.exists(POETRY_BIN):
--- a/get-poetry.py
+++ b/get-poetry.py
@@ -590,6 +590,7 @@ class Installer:
         if WINDOWS:
             allowed_executables += ["py.exe -3", "py.exe -2"]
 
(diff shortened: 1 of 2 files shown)


jonapich:
adjusts #1668

jonapich:
I'd like to change the behavior further if you'll let me. I think the new `_which_python()` method should return `python3` instead of `python` if `python` points to 2.x and python3 is available?

This python/python3 setup seems to be standard in Ubuntu16 and Ubuntu18 at the least.

get-poetry.py
```diff
@@ -585,22 +585,35 @@ def _make_lib(self, version):
...
-                subprocess.check_call(executable + " --version", shell=True)
+                raw_version = subprocess.check_output(
+                    executable + " --version", stderr=subprocess.STDOUT, shell=True
+                ).decode("utf-8")
             except subprocess.CalledProcessError:
                 continue
 
-            return executable
+            match = version_matcher.match(raw_version.strip())
+            if match and int(match.groupdict()["major"]) >= 3:
```
frankier: Python 3.4 will be deprecated before Python 2.7. I think the preference should therefore be: Python 3.5+ > Python 2.7+ > Python 3.4.x > Any other Python version.

frankier: Another point is that the versions of the python and python3 executables can change. Probably the simplest way to deal with this is to check at start up https://github.com/python-poetry/poetry/pull/1922 - but obviously that could slow things down.

jonapich: I changed the code so that it will favor the first 3.5+ version it finds, else it will fallback on the legacy behavior (i.e. first valid one).

The PR behind #1922 is interesting (bilingual script!! awesome!!) but the goal of this PR is to make it work on machines that don't have a `python` executable installed; the version check is a necessary mean to favor a `python` executable that's already a py3 one but if I could avoid it, I would.

Personally I would simply use `sys.executable` here, which would point to the executable used to run `get-poetry.py`. It would be a drastic change in behavior, and would potentially break if the user was in a virtual environment that he destroys later on... so I skipped that approach.

frankier: It would also break if the user later installed pyenv for example. Indeed pyenv can switch python versions around a lot.

Perhaps the best would be to keep this PR to just changing the shebang to say python3 if only that is available then do a runtime check in Python instead (faster) which would check the version the script is currently running under. If it's <3.5 and it can find >=3.5 in the PATH e.g. first python3 in the PATH then it could use exec to try and upgrade itself. The advantage is this would have almost zero start up overhead for people who get >= 3.5 from the shebang.

jonapich: The more I think about it and the more I think we should not actually care about the python version being used:

- Using a deprecated python3 will not make poetry unstable or vulnerable.
- Using python 2 is supported, and I believe it will remain supported after EOL.
- There was no attempt at protecting ourselves from py<=2.6 either (which would most likely make it unstable/vulnerable!)

The 3.5 is correct in today's context, but eventually it won't be anymore. When that happens, no one will come back here to keep this in-check.

The use case of this fix is to support systems that don't have a `python` installed, because it's such a common scenario. 

BUT I will keep a simple >=3.0 check because removing python2 and losing the `python` executable is a more common scenario than adding it. Therefore, it's less risky for the user to create a new `python3` link to `python` in order to make poetry work than having to create a `python` link to `python3` when you're trying to get rid of your python2 environment, 'coz then you may run into weird issues of py3 running py2 code...

But frankly I think this should be a configuration element somewhere. But chickens and eggs, right?

frankier: Note when I said deprecated I meant deprecated by Poetry specifically. Poetry won't work will old versions of Python 3 because the early versions had much support for running code on 2 and 3 than the newer version in the Python 3 series. The bad scenario is using an unsupported version of Python in preference to an installed supported one -- using an unsupported version when nothing else is available is unfortunate, but not as annoying.

PR #1892, opened 2020-01-15
Lazy Keyring intialization for PasswordManager

k4nar:
When one specify a source in the `pyproject.toml`, a PasswordManager instance is initialized in the factory function, in order to extract the Basic Auth credentials from the config. While this is fine, this had the undesirable effect to instantiate the Keyring, even when it's not needed.

I had two concerns with this:
- I don't have a keyring on my system, so every Poetry command was printing a warning. I don't really want this warning unless I'm doing stuff related to credentials, like `poetry publish`
- Initializing the keyring adds a small delay to Poetry startup. A very rough estimation on my computer tells me it's about 50 to 100ms, it might be more on slower systems.

Another solution would be to make the parsing of parts of the configuration lazy, so that we don't lead the sources if we don't need to. But that's way more intrusive & bug prone.

### DIFF
--- a/poetry/utils/password_manager.py
+++ b/poetry/utils/password_manager.py
@@ -117,35 +117,40 @@ class KeyRing:
 class PasswordManager:
     def __init__(self, config):
         self._config = config
-        self._keyring = KeyRing("poetry-repository")
-        if not self._keyring.is_available():
-            logger.warning("Using a plaintext file to store and retrieve credentials")
+        self._keyring = None
 
     @property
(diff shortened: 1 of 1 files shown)


sdispater:
@k4nar Thanks for your contribution and for fixing this. In my opinion, this is a bug fix so this should be based on the `master` branch.

Could you rebase your branch on master?

k4nar:
Rebased :+1: .

PR #1910, opened 2020-01-17
Fix errors when using the --help option

### LINKED ISSUE #1625: poetry complains about missing argument when using `--help`

## Issue

I don't know whether this is a poetry issue or cleo and if this problem arises in earlier versions.

When I type `poetry add --help` I receive the error message

```
Not enough arguments (missing: "name").
```

Similar for `poetry remove --help`

```
Not enough arguments (missing: "packages").
```

If I append any name I get the help page.

The expected behavior would be, that whenever I use `--help`, the help page should be displayed and mandatory arguments for sub command shouldn't be checked.

Saw this with version 1.0.0b6 and 1.0.0b7

sdispater:
# Pull Request Check List

This PR fixes the behavior of the `--help` option for some commands where it would cause an error. This was due to the fact that we were trying to resolve an incomplete command. Now if the command cannot be resolved we will fall back on the default CliKit behavior.

Fixes #1625

### DIFF
--- a/poetry/console/config/application_config.py
+++ b/poetry/console/config/application_config.py
@@ -9,6 +9,7 @@ from clikit.api.event import PRE_HANDLE
 from clikit.api.event import PreHandleEvent
 from clikit.api.event import PreResolveEvent
 from clikit.api.event.event_dispatcher import EventDispatcher
+from clikit.api.exceptions import CliKitException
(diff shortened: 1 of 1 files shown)


cjw296:
I feel like I'm missing something: where are the tests run as part of CI? What python versions are they run against?

PR #1911, opened 2020-01-17
Fix handling of forward slashes and url encoding in credentials

lubert:
# Pull Request Check List

## Problem
- Credentials containing `/` do not urlencode it, since `quote()` passes the `safe='/'` by default 

## Solution
- Pass `safe=''` so that `/` is encoded properly

### DIFF
--- a/poetry/repositories/legacy_repository.py
+++ b/poetry/repositories/legacy_repository.py
@@ -51,10 +51,10 @@ except ImportError:
 
 
 try:
-    from urllib.parse import quote, unquote
+    from urllib.parse import quote
 
 except ImportError:
-    from urllib import quote, unquote
+    from urllib import quote
 
 
 with warnings.catch_warnings():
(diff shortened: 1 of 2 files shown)


lubert:
Hi @k4nar, I had the issue with Python 3.7.1.

The documentation for 2 and 3 list `safe='/'` as a default argument, which I believe is why `/` isn't encoded.
https://docs.python.org/3/library/urllib.parse.html#urllib.parse.quote
https://docs.python.org/2/library/urllib.html#urllib.quote
Though I do see `quote_via` as a param for `urlencode`

That's a fair point about genuine creds containing a urlencoded string, I was trying to solve for an already encoded password passed via env, which is what we have at work in our build pipeline and is nice for `pip` since it doesn't do any encoding/unencoding.
I'll update the PR, since we can always decode before passing it to `poetry config`.

k4nar:
Ah yes, sorry, I was reading the doc of `urlencode` instead of `quote` :) .

lubert:
@sdispater, any chance of getting this merged into the next release?

sdispater:
Thanks!

PR #1912, opened 2020-01-17
Add --source option to "poetry add"

### LINKED ISSUE #1886: Allow to specify the source when adding a package

## Feature Request
Since Poetry 1.0 we can specify the source of package in the `pyproject.toml`, in order to use another index than PyPI. This is very convenient, but there is no way to specify this when using the `poetry add` command.

This is especially inconvenient when you're trying to update an existing private dependency with `poetry add`, as it will erase its source.

My suggestion would be to add a `--source` option to `poetry add`, taking the name of a source defined in the `pyproject.toml`. I don't think adding a package from a source not defined previously should be accepted.

k4nar:
Closes #1886

Tested with a project of mine, I think it does the job :) .

### DIFF
--- a/poetry/console/commands/add.py
+++ b/poetry/console/commands/add.py
@@ -33,6 +33,12 @@ class AddCommand(EnvCommand, InitCommand):
             "Platforms for which the dependency must be installed.",
             flag=False,
         ),
+        option(
+            "source",
+            None,
+            "Name of the source to use to install the package.",
+            flag=False,
+        ),
         option("allow-prereleases", None, "Accept prereleases."),
         option(
             "dry-run",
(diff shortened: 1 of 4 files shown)


poetry/console/commands/add.py
```diff
@@ -123,6 +131,9 @@ def handle(self):
             if self.option("platform"):
                 constraint["platform"] = self.option("platform")
 
+            if self.option("source"):
+                constraint["source"] = self.option("source")
```
sdispater: I think we should check if the source is defined in the `pyproject.toml` file before adding it to the constraint.

k4nar: It will fail later with an explicit message (there is a test for it). I find it better than to silently ignore the constraint, WDYT?

sdispater: Ah yes, I didn't see it! You are right.

k4nar:
Rebased :rocket: .

Issue #1902, opened 2020-01-16
Regular expression catastrophic backtracking in Git URL parsing


mschwager:
- [x] I am on the [latest](https://github.com/python-poetry/poetry/releases/latest) Poetry version.
- [x] I have searched the [issues](https://github.com/python-poetry/poetry/issues) of this repo and believe that this is not a duplicate.
- [x] If an exception occurs when executing a command, I executed it again in debug mode (`-vvv` option).

- **OS version and name**: `Linux 4.15.0-74-generic x86_64 GNU/Linux`
- **Poetry version**: Github `master`
- **Link of a [Gist](https://gist.github.com/) with the contents of your pyproject.toml file**: N/A

## Issue

Hi there! I've been working on a new Python static analysis tool called [Dlint](https://github.com/dlint-py/dlint). Most recently I've been working on a rule that searches for regular expression denial-of-service: [DUO138](https://github.com/dlint-py/dlint/blob/master/docs/linters/DUO138.md). When running this rule against your codebase I found a few violations:

```
$ python -m flake8 --select=DUO138 poetry
poetry/vcs/git.py:22:5: DUO138 catastrophic "re" usage - denial-of-service possible
poetry/vcs/git.py:33:5: DUO138 catastrophic "re" usage - denial-of-service possible
poetry/version/version.py:14:19: DUO138 catastrophic "re" usage - denial-of-service possible
```

*Note that `DUO138` hasn't been released to PyPI yet, so if you want to run the rule yourself you'll have to install from Github: `python -m pip install https://github.com/dlint-py/dlint/tarball/master`.*

After further investigation, it appears the violations in `poetry/vcs/git.py` are true positives, and the violation in `poetry/version/version.py` is a false positive.

If we dig into the Git parsing violations:

```python
re.compile(
    r"^(git\+)?"
    r"(?P<protocol>https?|git|ssh|rsync|file)://"
    r"(?:(?P<user>.+)@)*"
    r"(?P<resource>[a-z0-9_.-]*)"
    r"(:?P<port>[\d]+)?"
    r"(?P<pathname>[:/]((?P<owner>[\w\-]+)/(?P<projects>([\w\-/]+)/)?)?"
    r"((?P<name>[\w\-.]+?)(\.git|/)?)?)"
    r"([@#](?P<rev>[^@#]+))?"
    r"$"
),
re.compile(
    r"^(?:(?P<user>.+)@)*"
[...]

PR #1913, opened 2020-01-17
Avoid nested quantifiers with overlapping character space on git url parsing (#1902

finswimmer:
This PR started as a fix for #1902 to avoid nested quantifiers with overlapping character space during git url parsing.

It' result is a rework of the pattern matching and adding a test for `parse_url` to make this part more reliable.

Fixes: #1902

### DIFF
--- a/poetry/vcs/git.py
+++ b/poetry/vcs/git.py
@@ -22,7 +22,7 @@ PATTERNS = [
     re.compile(
         r"^(git\+)?"
         r"(?P<protocol>https?|git|ssh|rsync|file)://"
-        r"(?:(?P<user>.+)@)*"
+        r"(?:(?P<user>\w+)@)*"
         r"(?P<resource>[a-z0-9_.-]*)"
         r"(:?P<port>[\d]+)?"
         r"(?P<pathname>[:/]((?P<owner>[\w\-]+)/(?P<projects>([\w\-/]+)/)?)?"
@@ -31,7 +31,7 @@ PATTERNS = [
         r"$"
     ),
     re.compile(
-        r"^(?:(?P<user>.+)@)*"
+        r"^(?:(?P<user>\w+)@)*"
         r"(?P<resource>[a-z0-9_.-]*)[:]*"
         r"(?P<port>[\d]+)?"
         r"(?P<pathname>/?(?P<owner>.+)/(?P<projects>([\w\-/]+)/)?(?P<name>.+).git)"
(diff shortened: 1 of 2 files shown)


k4nar:
Is there a coverage in unit test for those regexp? It looks like there should be :smile: .

sdispater:
@k4nar Do you mean adding the cases mentioned in the original issue #1902? If so, I concur that we should add tests that cover these.

poetry/vcs/git.py
```diff
@@ -22,7 +22,7 @@
     re.compile(
         r"^(git\+)?"
         r"(?P<protocol>https?|git|ssh|rsync|file)://"
-        r"(?:(?P<user>.+)@)*"
+        r"(?:(?P<user>\w+)@)*"
```
sdispater:
```suggestion
        r"(?:(?P<user>[a-zA-Z0-9_.-])@)*"
```

I think we only want standard characters here with the addition of `_`, `.`  and `-`. If you think `\w` is still needed, however I don't think so, adding those three characters is still needed.

finswimmer: What ever works is fine for me :) 

I thinks we should use the same pattern for the same groups. For example `resource` is one time `[a-z0-9_.-]*` and one time `[\w.\-]+`. But I'm not totally sure, what is allowed in each group. Any suggestions?

sdispater: I agree we should have consistency between the different regexps. I think for `resource`  `[a-z0-9_.-]+` should be enough.

finswimmer: I did some rework to be consistent within the pattern. I also added tests for `parse_url` which uncovered some problems with regexes. Now only the first of the two pattern for matching "protocol://" and for matching "user@" are used for our test data. Do you think we can remove the two other? The less regex we need, the better :smiley:

sdispater:
@finswimmer Unless I'm mistaken, the affected strings mentioned (like `"https://" + "@" * 64 + "!"`) in #1902 are not present in the tests. I think we should add them to avoid regressions.

finswimmer:
@sdispater I've added a test for this. But something went wrong with the checks on MacOS?

sdispater:
@finswimmer Yes. I don't know what happened since nothing has changed, but it seems that the cached venv for MacOS is missing pip somehow. And since there is not easy way to clear the Github Actions cache at the moment I will need to make a PR to invalidate it.

sdispater:
@finswimmer Cache issues should be fixed on master. You will have to rebase your branch to pick up the fixes

PR #1922, opened 2020-01-21
Make Poetry executable use python3 from path if >= 3.5

frankier:
Currently SetupReader only works with Python versions >=3.5 -- but on some versions of Ubuntu and Debian, Poetry may run under Python 2, even when Python 3 is available.

On some distributions /usr/bin/python is still Python 2 by default and /usr/bin/python3 is Python 3. pyenv follows a similar convention, shimming both the python and python3 executables.

This patch changes the poetry executable to first use the python executable, if it is >= 3.5, then use the python3 executable, if it is >= 3.5, and otherwise fall back to whichever version is available under python.

### DIFF
--- a/get-poetry.py
+++ b/get-poetry.py
@@ -197,8 +197,23 @@ POETRY_LIB = os.path.join(POETRY_HOME, "lib")
 POETRY_LIB_BACKUP = os.path.join(POETRY_HOME, "lib-backup")
 
 
-BIN = """#!/usr/bin/env python
-# -*- coding: utf-8 -*-
+BIN = """#!/bin/sh
+# This script is bilingual.
+# The first part is POSIX shell, and the second part Python.
+""\":"
+check_version() {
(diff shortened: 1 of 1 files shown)


sdispater:
I think this is a duplicate of #1878 or at least close in spirit, so I am closing this to centralize discussions and reviews.

jonapich:
@frankier can you expand a bit on your issue? Technically, poetry can prepare 3.8 environments from a 2.7 runtime, so I'm not sure why SetupReader would suffer from poetry's interpreter version :confused:

frankier:
The code running SetupReader is run within the main Poetry process, not within the context of the environment. Whichever Python version ends up running $PATH/.poetry/bin/poetry is used. On some systems this will be Python 2.7.

jonapich:
Thanks for the additional info!

The general case of favoring python3 over python2 will fix it for you, but I think your claim points to a project-specific problem. 

I think you should be able to write this in your pyproject.toml file (maybe it works, I dunno):

    [build-system]
    requires = ["poetry>=0.12", "python>=3.5"]

PR #1923, opened 2020-01-21
Use install_egg_info as backup to create egg for old distutils packages

frankier:
For old distutils based packages setup.py egg_info will not create an egg_info, and so we must fall back to SetupReader. This doesn't work on Python < 3.5 so it would be nice to get an egg_info if possible. It appears that some versions of distutils allow us to get an egg_info using install_egg_info. This patch just adds that as an extra backup before resorting to SetupReader.


sdispater:
Thanks a lot for your contribution!

However, Poetry is about moving forward, not backward, and we can't keep supporting old packages.

This is especially true in this case since `egg_info` is only executed for `git` and `directory` dependencies which are expected to be somewhat up-to-date. Even executing `egg_info` is not ideal since it will generate metadata that are only valid for the current system. It will eventually be removed for a more modern solution.

Issue #1931, opened 2020-01-22
Parsing of Wheel version from PEP508 url does not completely extract version number


edwardgeorge:
- [x] I am on the [latest](https://github.com/python-poetry/poetry/releases/latest) Poetry version.
- [x] I have searched the [issues](https://github.com/python-poetry/poetry/issues) of this repo and believe that this is not a duplicate.
- [x] If an exception occurs when executing a command, I executed it again in debug mode (`-vvv` option).

- **OS version and name**: MacOS 10.14.6 Mojave
- **Poetry version**: 1.0.2
- **Link of a [Gist](https://gist.github.com/) with the contents of your pyproject.toml file**: n/a

## Issue
When a dependency has a dependency with a PEP508 '@' URI reference pointing to a wheel, and particularly when the version of the requirement is >=10 then the version it parses is incorrect. Specifically it only parses the first digit of the version number.

In our case, our dependency (here `ourdep`) has the following metadata:
```
Requires-Dist: log-setup @ https://.../log-setup/12/log_setup-14.0.0-py2.py3-none-any.whl
```
_(private url censored)_

Then we see in the debug logs:
```
   1: fact: ourdep (479) depends on log_setup (1)
```

And therefore `pippi update -vvv` fails with a `SolverProblemError`.

```
   1: conflict: ourdep (479) depends on log_setup (1)
   1: ! ourdep (479) is partially satisfied by not ourdep (>=457,<479 || >479)
   1: ! which is caused by "ourdep (>=457,<479 || >479) requires log-setup (1)"
   1: ! thus: ourdep (>=457) requires log-setup (1)
   1: ! not log-setup (1) is satisfied by log-setup (>=12)
   1: ! which is caused by "ourproject depends on log-setup (>=12)"
   1: ! thus: ourdep is forbidden
   1: ! ourdep (>=457) is satisfied by ourdep (>=457)
   1: ! which is caused by "ourproject depends on ourdep (>=457)"
   1: ! thus: version solving failed
```

This is caused by the regexp for wheels in `poetry/packages/__init__.py` in `dependency_from_pep_508()`:
```
if link.is_wheel:
    m = re.match(r"^(?P<namever>(?P<name>.+?)-(?P<ver>\d.*?))", link.filename)
```
which is too _non-greedy_ and only matches the first digit:

```
(Pdb) link.filename
'log_setup-14.0.0-py2.py3-none-any.whl'
[...]

PR #1932, opened 2020-01-22
fix: correct parsing of wheel version with regex.

edwardgeorge:
The previous regexp was only taking the first integer of the version number,
this presented problems when the major version number reached double digits.

Poetry would determine that the version of the dependency is '1', rather than,
ie: '14'. This caused failures to solve versions.

Detailed description of issue in bug report that I filed here: https://github.com/python-poetry/poetry/issues/1931

This fixes the issue for me and allows `poetry update` to correctly solve versions when a wheel file is referenced with a version >=10.

Unrelated to the specifics of this bug, but I noticed that when parsing a url to a wheel it does not return a `URLDependency` like non-wheel based urls but a `Dependency` instance. Therefore `dep.is_url()` returns `False` and the dependency only contains the name and parsed version as the constraint, the url itself is discarded. I'm not sure if this is intended or not?

# Pull Request Check List

- [ ] Updated **documentation** for changed code. _not appropriate_

### DIFF
--- a/poetry/packages/__init__.py
+++ b/poetry/packages/__init__.py
@@ -2,6 +2,7 @@ import os
 import re
 
 from poetry.semver import Version
+from poetry.utils.patterns import wheel_file_re
 from poetry.version.requirements import Requirement
 
 from .dependency import Dependency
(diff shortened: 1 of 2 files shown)


poetry/packages/__init__.py
```diff
@@ -70,7 +70,7 @@ def dependency_from_pep_508(name):
             link = Link(path_to_url(os.path.normpath(os.path.abspath(link.path))))
         # wheel file
         if link.is_wheel:
-            m = re.match(r"^(?P<namever>(?P<name>.+?)-(?P<ver>\d.*?))", link.filename)
+            m = re.match(r"^(?P<namever>(?P<name>.+?)-(?P<ver>\d[^-]*))", link.filename)
```
sdispater: I think we can reuse the `wheel_file_re` from `poetry.utils.patterns`. That way we keep the consistency in the way we read wheel file names

edwardgeorge: aha, thanks. I've updated to use `wheel_file_re`.

Issue #1946, opened 2020-01-25
Exception on poetry build, incorrect gitignore handling logic.


kawing-chiu:
- [x] I am on the [latest](https://github.com/python-poetry/poetry/releases/latest) Poetry version.
- [x] I have searched the [issues](https://github.com/python-poetry/poetry/issues) of this repo and believe that this is not a duplicate.
- [x] If an exception occurs when executing a command, I executed it again in debug mode (`-vvv` option).

- **OS version and name**: Arch Linux
- **Poetry version**: 1.0.2
- **Link of a [Gist](https://gist.github.com/) with the contents of your pyproject.toml file**: [pyproject.toml](https://gist.github.com/kawing-chiu/d887d0c23923a3389acdae5eac87949f)

## Issue
I got an exception when running `poetry build -vvv`, on the wheel building step:
![image](https://user-images.githubusercontent.com/5235260/73121697-af401900-3fb7-11ea-8b45-e844381b6bbf.png)

The issue here is that, poetry seems to be trying to use the `/.git` directory under my root directory to get ignored files, which does not make sense. I dig into the source and find that the [`get_vcs()`](https://github.com/python-poetry/poetry/blob/1.0.2/poetry/vcs/__init__.py#L12) logic here
```
def get_vcs(directory):  # type: (Path) -> Git
    directory = directory.resolve()
    for p in [directory] + list(directory.parents):
        ...
```
is incorrect. Here is the content of `[directory] + list(directory.parents)`, the list of directories that `get_vcs()` searches for `.git`:
```
[PosixPath('/tmp/tmp51qstjuk/mypackage-0.3.0.dev3'), PosixPath('/tmp/tmp51qstjuk'), PosixPath('/tmp'), PosixPath('/')]
```
When running `poetry build -vvv`, the `CompleteBuilder` will [unpack the sdist file to a tmpdir](https://github.com/python-poetry/poetry/blob/1.0.2/poetry/masonry/builders/complete.py#L53):
```
with self.unpacked_tarball(sdist_file) as tmpdir:
    WheelBuilder.make_in(
        Factory().create_poetry(tmpdir),
        ...
```
[...]

PR #1947, opened 2020-01-25
get_vcs starts searching git folder from tmp dir instead of project (#1946)

finswimmer:
At the moment, when building sdist and wheel the wheel is created from the unpacked sdist in a tmp dir. When trying to find the excluded file by git it starts searching the `.git` folder from this tmp dir instead of the original path.

This PR fixes this.

Furthermore `git rev-parse --show-toplevel` is used to get the git's root folder instead of finding a folder called `.git`. This should be more reliable.

I have no idea how to test this. So any help would be useful :)

Fixes: #1946 

# Pull Request Check List

### DIFF
--- a/poetry/masonry/builders/builder.py
+++ b/poetry/masonry/builders/builder.py
@@ -75,7 +75,9 @@ class Builder(object):
     @lru_cache(maxsize=None)
     def find_excluded_files(self):  # type: () -> Set[str]
         # Checking VCS
-        vcs = get_vcs(self._path)
+        vcs = get_vcs(
+            self._original_path if hasattr(self, "_original_path") else self._path
+        )
         if not vcs:
             vcs_ignored_files = set()
         else:
--- a/poetry/masonry/builders/builder.py
+++ b/poetry/masonry/builders/builder.py
@@ -44,6 +44,7 @@ class Builder(object):
         self._io = io
         self._package = poetry.package
         self._path = poetry.file.parent
+        self._original_path = self._path
 
         packages = []
         for p in self._package.packages:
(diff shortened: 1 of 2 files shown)


poetry/masonry/builders/builder.py
```diff
@@ -75,7 +75,9 @@ def build(self):
     @lru_cache(maxsize=None)
     def find_excluded_files(self):  # type: () -> Set[str]
         # Checking VCS
-        vcs = get_vcs(self._path)
+        vcs = get_vcs(
+            self._original_path if hasattr(self, "_original_path") else self._path
```
sdispater: What we could do is add a `_original_path` to the base `Builder` class which would default to the `value` of `self._path`. That way we avoid this check and always use `self._original_path`.

poetry/vcs/__init__.py
```diff
@@ -1,22 +1,28 @@
...
-                return Git(p)
-            except (subprocess.CalledProcessError, OSError):
-                # Either git could not be found or does not exist
-                warnings.warn(
-                    "git executable could not be found", category=RuntimeWarning
-                )
-
-                return
+    working_dir = Path.cwd()
+    os.chdir(str(directory))
```
sdispater: Shouldn't we keep the `.resolve()` call?

poetry/masonry/builders/builder.py
finswimmer: Good idea :+1:

poetry/vcs/__init__.py
finswimmer: Not sure if the is absolutely necessary. But at least it doesn't hurt :smiley:

PR #1948, opened 2020-01-26
fix(git): get commit sha of git commit from annotated tags

### LINKED ISSUE #1916: poetry tries to reinstall from git when using annotated tags as references

- **OS version and name**: Debian Buster
- **Poetry version**: 1.0.2
- **Link of a [Gist](https://gist.github.com/) with the contents of your pyproject.toml file**: https://gist.github.com/jrmlhermitte/bbfda76d59b820560bf2ff4a1e57d42d

## Issue
This came from a discussion on issue #691 

If we re-run the gist included above, poetry will try re-installing the referred library, even though the git tag has not changed.

This was discovered on the issue mentioned above. Here is the reply from @bibz 
> Thanks for the MRE.

> I can reproduce your example now.
> You found a different bug than we had, due to how git handled signed annotated tags.

> To summarise:

>    The (annotated) tag v5.1.3 itself is a git object with the short sha 73f60e6
>    The tree pointed to by the tag has the short sha 4056bbb

> The lock file refers to the tag object, because it is what v5.1.3 resolves to directly.
 Since the checked-out tree has a different sha, poetry thinks it is not up-to-date (the references are      different) and then re-install it.

> I think we need a new issue for this.

Would it be possible to investigate this, and possibly change the behaviour to not re-install the package?

Thanks for the great work on this wonderful library, keep it up!

jrmlhermitte:
# Description

This is an attempt to close #1916. I think all we need to do is make sure that the revision always points to the SHA of the commit in question and not the SHA of the tag object.
There is a question on stackoverflow [here](https://stackoverflow.com/questions/1862423/how-to-tell-which-commit-a-tag-points-to-in-git) about this, where a few solutions are listed.
I chose the command `git rev-list TAG --max-count=1`

# Testing and Documentation
I have tested this on a regular tag and annotated tag at the command line and they both pointed to the right corresponding commit hash. However, I have not tested this by running poetry itself.

For the tests, I am not sure how to test this other than generating a fake git repo and creating commits. Let me know if you would prefer that and I'll try to find out how to do it.
For the documentation, I left a comment by the call to explain the purpose of this command. It's not ideal, but I can't think of where else to put documentation.

### DIFF
--- a/poetry/vcs/git.py
+++ b/poetry/vcs/git.py
@@ -229,9 +229,9 @@ class Git:
                 folder.as_posix(),
             ]
 
-        # git rev-list TAG --max-count=1 works with annotated and
-        # non-annotated tags.
-        args += ["rev-list", rev, "--max-count=1"]
+        # We need "^{commit}" to ensure that the commit SHA of the commit the
+        # tag points to is returned, even in the case of annotated tags.
+        args += ["rev-parse", rev + "^{commit}"]
 
         return self.run(*args)
 
--- a/poetry/vcs/git.py
+++ b/poetry/vcs/git.py
@@ -231,7 +231,7 @@ class Git:
 
         # git rev-list TAG --max-count=1 works with annotated and
         # non-annotated tags.
-        args += ["rev-list", rev, '--max-count=1']
+        args += ["rev-list", rev, "--max-count=1"]
 
         return self.run(*args)
 
--- a/poetry/vcs/git.py
+++ b/poetry/vcs/git.py
@@ -229,7 +229,9 @@ class Git:
                 folder.as_posix(),
             ]
 
-        args += ["rev-parse", rev]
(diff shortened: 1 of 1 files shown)


jrmlhermitte:
@bibz could you also maybe weigh in? thanks!

bibz:
@jrmlhermitte What do you think of using `git rev-parse TAG^{commit}` instead?
The command does not change but we specifically ask git to provide us the revision to a _commit_, not any _object_.

As to how to test, I do not have any other idea to offer.

jrmlhermitte:
@bibz good idea. My git knowledge is not great. This is a much better change, thanks!

jrmlhermitte:
Is it okay to keep this as is or should I add more tests? I right now didn't intend on adding any but I can (will take some extra time). Thanks!

jrmlhermitte:
Great, and thank you for this great library! Closest one to proper version management! :-)

Issue #1871, opened 2020-01-11
Processing configuration for http auth (environment variables)


fralik:
<!-- Checked checkbox should look like this: [x] -->
- [x] I am on the [latest](https://github.com/python-poetry/poetry/releases/latest) Poetry version.
- [x] I have searched the [issues](https://github.com/python-poetry/poetry/issues) of this repo and believe that this is not a duplicate.
- [x] If an exception occurs when executing a command, I executed it again in debug mode (`-vvv` option).

- **OS version and name**: Windows 10, 64-bit.
- **Poetry version**: 1.0.2

## Issue
Documentation [says](https://python-poetry.org/docs/configuration/#using-environment-variables) that any config value can be passed through an environment variable `POETRY_<config>` as well as secrets for private repos. However, I suspect that the code doesn't process secrets correctly.

**Setup:**
1. There is no `auth.toml` file.
2. Url for the private repo is added in `config.toml`.
3. `pyproject.toml` contains url and name for the same private repo.
4. `pyproject.toml` references one package from the private repo.
5. Variables `POETRY_HTTP_BASIC_REPO_PASSWORD` and `POETRY_HTTP_BASIC_REPO_USERNAME` are set.
6. Variable `POETRY_HTTP_BASIC_REPO` is not set.
7. Some debug output (`print`) is added to poetry source code.

**Outcome:**
```
>poetry install --dry-run --no-ansi -vvv                                                                        
Getting auth for repo
Need to get setting http-basic.repo
Checking variable POETRY_HTTP_BASIC_REPO, value <object object at 0x00000239FFCD5900>, not set <object object at 0x00000239FFCD5900>
Value is not set, so continue
Checking key: http-basic
Will return default: None
Failed to get auth                                                                                                                                            
Using virtualenv: c:\home\anaconda3\envs\env-poetry                                                                                                           
Updating dependencies                                                                                                                                         
Resolving dependencies...                                                                                                                                     
   1: fact: demo is 0.1.0                                                                                                                           
   1: derived: demo                                                                                                                                 
   1: fact: demo depends on treeview (~0.1.0)                                                                                                       
   1: selecting demo (0.1.0)                                                                                                                        
   1: derived: treeview (~0.1.0)                                                                                                                              
   1: fact: no versions of treeview match >=0.1.0,<0.2.0                                                                                                      
   1: conflict: no versions of treeview match >=0.1.0,<0.2.0                                                                                                  
[...]

PR #1961, opened 2020-01-30
Support using environment variables to configure username and passwords

orf:
# Pull Request Check List

As per #1871 you cannot configure HTTP basic auth using environment variables. This is a problem for a lot of use cases:

1. Docker - using build time secrets means you can inject the HTTP basic auth configuration as an environment and download packages from a private repository without including this information in the resulting container. That's harder to do if you execute `poetry config ...`.

2. User onboarding - many companies inject credentials into laptops as environment variables, and tools should be configured to work with those with a minimum of configuration. Simply having `POETRY_HTTP_BASIC_MY_INTERNAL_PYPI_USERNAME` and  `POETRY_HTTP_BASIC_MY_INTERNAL_PYPI_PASSWORD` should be sufficient.

### DIFF
--- a/poetry/utils/password_manager.py
+++ b/poetry/utils/password_manager.py
@@ -148,11 +148,19 @@ class PasswordManager:
         self._keyring.delete_password(name, "__token__")
 
     def get_http_auth(self, name):
+        username = self._config.get("http-basic.{}.username".format(name))
+        password = self._config.get("http-basic.{}.password".format(name))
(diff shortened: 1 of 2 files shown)


sdispater:
Thanks for you contribution.

However, this issue has already been handled in https://github.com/python-poetry/poetry/pull/1909 so I am closing this since it's a duplicate.

PR #1971, opened 2020-02-03
Update pip to defined versions after creating a new venv [feature]

### LINKED ISSUE #732: Which pip version does poetry use?

### LINKED ISSUE #1651: poetry should upgrade pip and setuptools in virtualenvs it creates

- **OS version and name**:  RHEL7.5
- **Poetry version**: Poetry version 1.0.3

## Issue

Poetry does not appear to upgrade the version of pip that appears in virtualenvs it creates and manages. This means that, in my case, I get an ancient pip 10.x that doesn't know to use binary wheels, causing installs of packages to fail as I don't have python-devel on the machines I deploy to.

When poetry creates a virtualenv, it should, as part of that process, upgrade pip to the latest version available, and continue to upgrade every time it's about to do anything that installs or removes packages from a virtualenv.

The workaround for me is:

```
python3.6 `which poetry` shell
pip install -U pip
```

### LINKED ISSUE #1661: poetry install uses an old pip version

- **OS version and name**: ubuntu 19.04
- **Poetry version**: 0.12.17
`poetry init --name foo --dependency tensorflow:==2.1.0rc0 --no-interaction && poetry install -vvv`

## Issue
It seems that `poetry install` creates an venv using python venv module. This module creates an environment with a pretty old pip version. The old pip version have some issues -- for example, it does not recognize `manylinux2010` packages at all.

This can lead to problems: a user can try to run a `poetry install` using a toml/lock files that was created in another computer. If, on the other computer, a user manually upgraded pip in the poetry venv, it will not be documented anywhere, leading to an inconsistent behavior (success at one case, fail on another), and hard to debug errors.
See that command above for an example.

Furthermore, no manual override exists (i.e. `poetry install --upgrade-pip-first`)

some valid solutions might be:

* always upgrade pip to most recent version, before any install/upgrade (but it means that future pip versions that breaks som API used by poetry will be problematic)
* always upgrade pip to a specific version, before any install/upgrade (but it means that users cannot use different pip versions, if they are needed for some reason)
* log the venv's pip version to the `poetry.lock` file, and ensure that this version is installed before any install/upgrade

### LINKED ISSUE #1962: Couldn't install tensotflow==2.1.0
- [*] I am on the [latest](https://github.com/python-poetry/poetry/releases/latest) Poetry version.
- [*] I have searched the [issues](https://github.com/python-poetry/poetry/issues) of this repo and believe that this is not a duplicate.
- [*] If an exception occurs when executing a command, I executed it again in debug mode (`-vvv` option).

- **OS version and name**: Docker image python:3.6.9
- **Poetry version**: 1.0.2

## Issue

I have not way to install latest tensorflow with poetry==1.0.2. If try to install I have this trace:

```
[EnvCommandError]
Command ['/home/zenv/.cache/pypoetry/virtualenvs/1-ZUkQNI8o-py3.6/bin/pip', 'install', '--no-deps', 'tensorflow==2.1.0'] errored with the following return code 1, and output: 
Collecting tensorflow==2.1.0
  Could not find a version that satisfies the requirement tensorflow==2.1.0 (from versions: 0.12.1, 1.0.0, 1.0.1, 1.1.0rc0, 1.1.0rc1, 1.1.0rc2, 1.1.0, 1.2.0rc0, 1.2.0rc1, 1.2.0rc2, 1.2.0, 1.2.1, 1.3.0rc0, 1.3.0rc1, 1.3.0rc2, 1.3.0, 1.4.0rc0, 1.4.0rc1, 1.4.0, 1.4.1, 1.5.0rc0, 1.5.0rc1, 1.5.0, 1.5.1, 1.6.0rc0, 1.6.0rc1, 1.6.0, 1.7.0rc0, 1.7.0rc1, 1.7.0, 1.7.1, 1.8.0rc0, 1.8.0rc1, 1.8.0, 1.9.0rc0, 1.9.0rc1, 1.9.0rc2, 1.9.0, 1.10.0rc0, 1.10.0rc1, 1.10.0, 1.10.1, 1.11.0rc0, 1.11.0rc1, 1.11.0rc2, 1.11.0, 1.12.0rc0, 1.12.0rc1, 1.12.0rc2, 1.12.0, 1.12.2, 1.12.3, 1.13.0rc0, 1.13.0rc1, 1.13.0rc2, 1.13.1, 1.13.2, 1.14.0rc0, 1.14.0rc1, 1.14.0, 2.0.0a0, 2.0.0b0, 2.0.0b1)
No matching distribution found for tensorflow==2.1.0
You are using pip version 18.1, however version 20.0.2 is available.
You should consider upgrading via the 'pip install --upgrade pip' command.

```

tsvikas:
Until now, a new venv created with poetry would have an undefined pip version (for example, in my setup it's pip 18.1). This could be problematic, as older pip versions are not compatible with some packages (manylinux2010/2014 packages - for example, tensorflow 2), and might have security risks. On the other hand, future pip versions might introduce API changes that will prevent poetry from working.

This code forces new venvs to update their pip to a version from a defined range of versions. Updating pip happens just after the venv creation (and not each time before package installation). This will allow users to manually change it if the need arises.

Code is loosely based on https://github.com/python-poetry/poetry/pull/740

I know that the need for internet connectivity in venv creation might be an issue, but I was not sure how to tackle it, and I wanted to get feedback on this code first.

### Closes:
closes #732 (partially - this PR does not install setuptools)
closes #1661
closes #1962
closes #1651

### Pull Request Check List
Tests were passing in my setup, but might need tweaking to pass with different setups (i.e. python 3.4)
No relevant documentation found, but I did my best to add relevant comments. 
Should I update the changelog?

### DIFF
--- a/poetry/utils/env.py
+++ b/poetry/utils/env.py
@@ -34,6 +34,20 @@ from poetry.utils.toml_file import TomlFile
 from poetry.version.markers import BaseMarker
 
 
+# pip version requirements when creating venvs are specified here.
+# The maximal version is chosen to be the last version in the 19.3 branch, as
+# it seems to be relatively bug free, and to work well with poetry (poetry
+# uses pip internally, and pip API might change in ways that can break stuff).
+# Note that pip 19.3 is required to find and install manylinux2014 packages.
+# This version should be updated from time to time.
+PIP_VERSION_MAX = "19.3.1"
+# We use a different minimal version to allow fallback in case of older python
+# versions. This version was chosen to be 19.1.1, as it is the most recent
+# version that supports python 3.4.
+# When poetry will drop support for EOL python versions, it might be possible
+# to update this version, or to use a pinned version instead.
+PIP_VERSION_MIN = "19.1.1"
+
 GET_ENVIRONMENT_INFO = """\
 import json
 import os
@@ -656,26 +670,36 @@ class EnvManager(object):
             except CalledProcessError as e:
                 raise EnvCommandError(e)
 
-            return
+        else:
+            try:
+                from venv import EnvBuilder
 
-        try:
-            from venv import EnvBuilder
+                # use the same defaults as python -m venv
+                if os.name == "nt":
+                    use_symlinks = False
+                else:
+                    use_symlinks = True
 
-            # use the same defaults as python -m venv
-            if os.name == "nt":
-                use_symlinks = False
-            else:
-                use_symlinks = True
+                builder = EnvBuilder(with_pip=True, symlinks=use_symlinks)
+                build = builder.create
+            except ImportError:
+                # We fallback on virtualenv for Python 2.7
+                from virtualenv import create_environment
 
-            builder = EnvBuilder(with_pip=True, symlinks=use_symlinks)
-            build = builder.create
-        except ImportError:
-            # We fallback on virtualenv for Python 2.7
-            from virtualenv import create_environment
+                build = create_environment
 
-            build = create_environment
+            build(path)
 
-        build(path)
+        # update pip version
+        the_venv = VirtualEnv(Path(path))
+        the_venv.run(
+            "python",
+            "-m",
+            "pip",
+            "install",
+            "--upgrade",
+            "pip>={},<={}".format(PIP_VERSION_MIN, PIP_VERSION_MAX),
+        )
 
     def remove_venv(self, path):  # type: (str) -> None
         shutil.rmtree(path)
--- a/poetry/utils/env.py
+++ b/poetry/utils/env.py
@@ -698,7 +698,8 @@ class EnvManager(object):
             "pip",
             "install",
             "--upgrade",
-            "pip>={},<={}".format(PIP_VERSION_MIN, PIP_VERSION_MAX),
+            '"pip>={},<={}"'.format(PIP_VERSION_MIN, PIP_VERSION_MAX),
+            shell=True,
         )
 
     def remove_venv(self, path):  # type: (str) -> None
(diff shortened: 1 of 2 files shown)


maksbotan:
Hi,

I've also run into this issue, when `pip` in the created venv was at 9.0 and failed to do some commands the expected way.

It would be nice if poetry would update `pip` in the venv to some sane version before installing packages.

finswimmer:
Thanks a lot for your contribution @tsvikas!

In the meantime this PR is superseded by #2666, this is why I will close yours.

Issue #1883, opened 2020-01-14
poetry install: Conflict with poetry vendored dependencies [Docker/no venv]


dmfigol:
- [X] I am on the [latest](https://github.com/python-poetry/poetry/releases/latest) Poetry version.
- [X] I have searched the [issues](https://github.com/python-poetry/poetry/issues) of this repo and believe that this is not a duplicate.
- [ ] If an exception occurs when executing a command, I executed it again in debug mode (`-vvv` option).

- **OS version and name**: Docker / python:3.6-slim-stretch
- **Poetry version**: 1.0.2
- **Link of a [Gist](https://gist.github.com/) with the contents of your pyproject.toml file**: https://github.com/nornir-automation/nornir/blob/e416bee609b8ad0d2ef1f8467dc3a33b9854f11e/pyproject.toml

## Issue
This seems to be related to #1882 but I wouldn't call it a duplicate
Dockerfile: https://github.com/nornir-automation/nornir/blob/e416bee609b8ad0d2ef1f8467dc3a33b9854f11e/Dockerfile
We are using `poetry config virtualenvs.create false` so dependencies are installed to the `/usr/local/lib/python3.6/site-packages`, however we can see in the logs that there is some conflict between poetry vendored dependencies and our dependencies
```
...
Step 8/11 : RUN poetry install --no-interaction
 ---> Running in b626817c6265
Skipping virtualenv creation, as specified in config file.
Installing dependencies from lock file
Warning: The lock file is not up to date with the latest changes in pyproject.toml. You may be getting outdated dependencies. Run update to update them.


Package operations: 95 installs, 19 updates, 0 removals

  - Installing decorator (4.4.0)
  - Installing ipython-genutils (0.2.0)
  - Updating six (1.13.0 /root/.poetry/lib/poetry/_vendor/py3.6 -> 1.12.0)
  - Updating attrs (19.3.0 /root/.poetry/lib/poetry/_vendor/py3.6 -> 19.1.0)
  - Installing parso (0.5.1)
  - Updating ptyprocess (0.6.0 /root/.poetry/lib/poetry/_vendor/py3.6 -> 0.6.0)
  - Updating pycparser (2.19 /root/.poetry/lib/poetry/_vendor/py3.6 -> 2.19)
  - Updating pyrsistent (0.14.11 /root/.poetry/lib/poetry/_vendor/py3.6 -> 0.15.4)
  - Installing traitlets (4.3.2)
  - Installing wcwidth (0.1.7)
[...]

PR #1981, opened 2020-02-05
Fix vendor package as installed package (#1883)

Trim21:

- [X] <del>Updated **documentation** for changed code.</del>

Fixes: #1883


poetry/repositories/installed_repository.py
```diff
@@ -1,3 +1,5 @@
+import poetry
```
sdispater:
```suggestion
from poetry import _CURRENT_VENDOR
```

PR #1994, opened 2020-02-07
sdist builder - include long description content type

kapilt:
minor change to include long description content type for sdist builder masonry along with test.

I was using a variant of the script here https://github.com/python-poetry/poetry/issues/761#issuecomment-568229920 to generate setup.py for compatibility with existing users while we transition to poetry.


finswimmer (approved):
Thanks a lot!

LGTM :+1: 

Similar: #1166

neersighted:
Hi @kapilt -- would you mind submitting this as a PR against poetry-core? The location of the files has changed. Thanks!

Issue #1893, opened 2020-01-15
Insufficient error message when passing invalid argument to -f option of export command


515hikaru:
- [x] I have searched the [issues](https://github.com/python-poetry/poetry/issues) of this repo and believe that this is not a duplicate.

## Issue

Though this ticket is similar to #1640, different a bit.I think the `Exporter` error message is not enough.

When I used `export` command for the first time, I didn't know how to specify `-f` option. So I tried:

```
$ poetry export -f requirements -o requirements.txt 
```

The result of the above command is below.

```
[ValueError]
Invalid export format: requirements
```

poetry said I was wrong, but it did not tell me how should I do. The error message should tell us how to specify `-f` option. For example, if it show list of accepted formats, it helping for me.

PR #1995, opened 2020-02-07
cli export - show valid formats on err

kapilt:
if a user passes in a bad value, tell them what the possible good values are. the valid ones aren't in --help for the command, rather than making them find the website docs or read the source, just tell them.

Resolves: #1893

### DIFF
--- a/poetry/console/commands/export.py
+++ b/poetry/console/commands/export.py
@@ -29,7 +29,11 @@ class ExportCommand(Command):
         fmt = self.option("format")
 
         if fmt not in Exporter.ACCEPTED_FORMATS:
-            raise ValueError("Invalid export format: {}".format(fmt))
+            raise ValueError(
+                "Invalid export format: {} valid formats: {}".format(
+                    fmt, ", ".join(Exporter.ACCEPTED_FORMATS)
+                )
+            )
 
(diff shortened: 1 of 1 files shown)


poetry/console/commands/export.py
```diff
@@ -29,7 +29,11 @@ def handle(self):
         fmt = self.option("format")
 
         if fmt not in Exporter.ACCEPTED_FORMATS:
-            raise ValueError("Invalid export format: {}".format(fmt))
+            raise ValueError(
+                "Invalid export format: {} valid formats: {}".format(
```
stephsamson: Thanks for your PR! 🙌

Adding a newline in between would be better for readability.

cbgrey:
Since I got bit by this today I'd like to see this PR get accepted. Is there anything I can do to help? Not sure how it works to "unstick" a PR.

kapilt:
i'll take a look at updating it later today.

finswimmer:
Hello @kapilt,

do you have the time to rebase onto master and include @stephsamson suggestion?

fin swimmer

kapilt:
this is bitrotted a bit, and the current ux on export conveys enough info that  its extraneous.

Issue #475, opened 2018-10-06
Support creating stub-only packages


rominf:
- [x] I have searched the [issues](https://github.com/sdispater/poetry/issues) of this repo and believe that this is not a duplicate.

## Issue

[MyPy documentation](https://mypy.readthedocs.io/en/latest/installed_packages.html#making-pep-561-compatible-packages) explains how to create stub-only packages. It would be nice if Poetry supported stub-only packages. Current behaviour:
```
$ fd
ordered_set-stubs
pyproject.toml
LICENSE
README.md
ordered_set-stubs/__init__.pyi
$ poetry build
Skipping virtualenv creation, as specified in config file.
Building ordered_set-stubs (0.1.0)
                                                      
[ValueError]                           
No file/folder found for package ordered-set-stubs  
                                                      
build [-f|--format FORMAT]
```

sdispater:
It's already supported. Just add the following to the `[tool.poetry]` section.

```toml
[tool.poetry]
# Basic package information

packages = [
    {include = "ordered_set-stubs"}
]
```

Then, if you run `poetry build` you should have stub-only distributions.

sdispater:
You can have more information in the documentation: https://poetry.eustace.io/docs/pyproject/#packages

rominf:
Thank you, @sdispater, it worked. Sorry didn't figured out myself.

fsenart:
@sdispater right after version 1.0.0, in version 1.0.1, stub-only packages are no longer supported. One gets an `... is not a package."` error. This issue should be reopened.

PR #2000, opened 2020-02-08
Fix building distribution of PEP 561 stub-only packages

hoefling:
# Pull Request Check List

- [ ] ~~Updated **documentation** for changed code~~ no documentation updates needed, this patch only fixes what has worked before.

This is a fix for a regression in #475, first introduced in c4018503944e0b25dbab3d397ff9e4d7d750c31c. Added explicit check for PEP 561 stub-only packages, put duplicate code in a separate method. To avoid the regression reintroduced, added tests for:
* stub-only wheels,
* stub-only source dists,
* stub-only package name checking.

Thank you in advance for a review!

### DIFF
--- a/poetry/masonry/utils/package_include.py
+++ b/poetry/masonry/utils/package_include.py
@@ -33,6 +33,20 @@ class PackageInclude(Include):
 
         return self.check_elements()
 
+    @property
+    def stub_only(self):  # type: () -> bool
+        # returns `True` if this a PEP 561 stub-only package,
+        # see [PEP 561](https://www.python.org/dev/peps/pep-0561/#stub-only-packages)
+        return self.package.endswith("-stubs") and all(
+            el.suffix == ".pyi" for el in self.elements if el.is_file()
+        )
+
+    @property
+    def has_modules(self):  # type: () -> bool
+        # Packages no longer need an __init__.py in python3, but there must
+        # at least be one .py file for it to be considered a package
+        return any(element.suffix == ".py" for element in self.elements)
+
     def check_elements(self):  # type: () -> PackageInclude
         if not self._elements:
             raise ValueError(
@@ -43,20 +57,18 @@ class PackageInclude(Include):
         if len(self._elements) > 1:
             # Probably glob
             self._is_package = True
+            self._package = root.parent.name
 
-            # Packages no longer need an __init__.py in python3, but there must
-            # at least be one .py file for it to be considered a package
-            if not any([element.suffix == ".py" for element in self._elements]):
+            if not self.stub_only and not self.has_modules:
                 raise ValueError("{} is not a package.".format(root.name))
 
-            self._package = root.parent.name
         else:
             if root.is_dir():
                 # If it's a directory, we include everything inside it
                 self._package = root.name
                 self._elements = sorted(list(root.glob("**/*")))
 
-                if not any([element.suffix == ".py" for element in self._elements]):
+                if not self.stub_only and not self.has_modules:
                     raise ValueError("{} is not a package.".format(root.name))
 
                 self._is_package = True
--- a/poetry/masonry/utils/package_include.py
+++ b/poetry/masonry/utils/package_include.py
@@ -38,7 +38,10 @@ class PackageInclude(Include):
         # returns `True` if this a PEP 561 stub-only package,
         # see [PEP 561](https://www.python.org/dev/peps/pep-0561/#stub-only-packages)
         return self.package.endswith("-stubs") and all(
-            el.suffix == ".pyi" for el in self.elements if el.is_file()
+            el.suffix == ".pyi"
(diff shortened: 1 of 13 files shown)


poetry/masonry/utils/package_include.py
```diff
@@ -33,6 +33,20 @@ def refresh(self):  # type: () -> PackageInclude
 
         return self.check_elements()
 
+    @property
+    def stub_only(self):  # type: () -> bool
+        # returns `True` if this a PEP 561 stub-only package,
+        # see [PEP 561](https://www.python.org/dev/peps/pep-0561/#stub-only-packages)
+        return self.package.endswith("-stubs") and all(
+            el.suffix == ".pyi" for el in self.elements if el.is_file()
```
henribru: This doesn't work if there is a `py.typed` file, does it? While the PEP states "Note that for stub-only packages adding a py.typed marker is not needed since the name *-stubs is enough to indicate it is a source of typing information.", it also says "If a stub package is partial it MUST include partial\n in a top level py.typed file." So a partial stub-only package will have `py.typed` in addition to .pyi files.

hoefling: @henribru stub packages and stub-only packages are not the same. The quote you refering to applies to stub packages (`pyi` stubs included alongside the `py` modules), while stub-only packages don't include a `py.typed` file, no matter whether complete or not. Real-world examples: AFAIK neither [`numpy-stubs`](https://github.com/numpy/numpy-stubs) nor [`django-stubs`](https://github.com/typeddjango/django-stubs) are complete yet they don't include a `py.typed` marker.

hoefling: @henribru I may be wrong though - do you have an example of stub-only packages using the partial marker?

henribru: The PEP states "Stub packages - these packages SHOULD supersede any installed inline package. They can be found at foopkg-stubs for package foopkg.", so I'm not sure that's right. The Partial Stub Packages section also describes merging stub packages and runtime packages, implying they're referring to separate packages, not ones containing both stubs and regular modules. 

Unfortunately the only example I can give are my own stub-only packages, so that doesn't count for much. I don't know if those packages don't include it because they have misunderstood the spec or I have. It's definitely not worded as clear as it should be.

hoefling: @henribru I have played with `mypy` a bit - your understanding of PEP 561 is the correct one, there is definitely a difference in `mypy` behaviour: without the partial marker, `mypy` doesn't see modules that are not covered (`error: Cannot find implementation or library stub for module`) while having the marker, `mypy` does find the module. I agree with you now - the current impl is not conform with PEP 561, will update. Thank you for a thorough and helpful review!

henribru: Great to get that cleared up, thanks for verifying it! And thank you for making this PR, looking forward to Poetry working for my stub packages again.

hoefling: @henribru should be fixed now - I didn't add an explicit check for the marker file content since it seems that its presence doesn't affect `mypy` behaviour if the stubs are complete, it's simply ignored.

hoefling:
@finswimmer I have accidentally removed the review request you have added - can you add it again?

hoefling:
@abn sure thing, done!

poetry/masonry/utils/package_include.py
```diff
@@ -33,6 +33,23 @@ def refresh(self):  # type: () -> PackageInclude
 
         return self.check_elements()
 
+    @property
+    def stub_only(self):  # type: () -> bool
```
abn, line 37:
```suggestion
    def is_stub_only(self):  # type: () -> bool
```

poetry/masonry/utils/package_include.py
```diff
@@ -43,20 +60,18 @@ def check_elements(self):  # type: () -> PackageInclude
...
 
-            self._package = root.parent.name
         else:
             if root.is_dir():
                 # If it's a directory, we include everything inside it
                 self._package = root.name
                 self._elements = sorted(list(root.glob("**/*")))
 
-                if not any([element.suffix == ".py" for element in self._elements]):
+                if not self.stub_only and not self.has_modules:
```
abn, line 74:
```suggestion
                if not self.is_stub_only and not self.has_modules:
```

hoefling:
@abn no worries, better apply changes now than deal with bad design later. I guess I was too distracted by `package` and `source` properties, but should look at `is_package`/`is_module` methods instead.

abn:
@hoefling thank you for the fix; feel free to provide a port of the fix to python-poetry/core repo too if you can. Otherwise, I can do a port sometime later this week.

hoefling:
@abn gladly, was actually easier than I thought. python-poetry/core#28

Issue #2009, opened 2020-02-10
export to requirements.txt without freezing version


gsemet:
It would be really helpful to be able to export to requirements.txt without freezing the version, and keeping the markers.
This is an important feature in order to support "libraries", where you want to inject the content of `requirements.txt` into `setup.py` 's `install_requires`. This is what PBR does, and with pipenv I was used to keep `requirements.txt` in sync with `Pipfile`. I have written a little tool for that: [pipenv-to-requirements](https://github.com/gsemet/pipenv-to-requirements).

Now I am trying to migrate to poetry, and facing the same issue: I want my prod packages to be added into `install_requires` automatically, or keep a `requirements.txt` into my packages, so that when my packages will be installed by pip, it will install its dependencies automatically.

So, what I need when I am building a library is to read `tool.poetry.dependencies` from `pyproject.toml` and generate a `requirements.txt` without freezeing and with the markers.

Can't acheive that with `poetry export` (it freeze the version and loose the markers), neither with `dephell`. I have no pbl submitting this feature in poetry if you accept it :)

PR #2013, opened 2020-02-10
Add --no-freeze option to poetry export

gsemet:

Hi. This is my fist little submission.
This PR provides a solution for #2009. This allows to generate a `requirements.txt` without frozen version in it, so that user can keep `requirements.txt` when building a setuptools package using for example PBR.

### DIFF
--- a/poetry/utils/exporter.py
+++ b/poetry/utils/exporter.py
@@ -32,7 +32,8 @@ class Exporter(object):
         dev=False,
         extras=None,
         with_credentials=False,
-    ):  # type: (str, Path, Union[IO, str], bool, bool, bool) -> None
+        no_freeze=False,
+    ):  # type: (str, Path, Union[IO, str], bool, bool, bool, bool) -> None
         if fmt not in self.ACCEPTED_FORMATS:
             raise ValueError("Invalid export format: {}".format(fmt))
 
@@ -43,6 +44,7 @@ class Exporter(object):
             dev=dev,
             extras=extras,
             with_credentials=with_credentials,
+            no_freeze=no_freeze,
         )
 
     def _export_requirements_txt(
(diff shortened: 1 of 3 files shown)


tests/console/commands/test_export.py
```diff
@@ -90,9 +92,11 @@ def test_export_exports_requirements_txt_file_locks_if_no_lock_file(app, repo):
 
     assert app.poetry.locker.lock.exists()
 
-    expected = """\
-foo==1.0.0
-"""
+    expected = dedent(
```
gsemet: sorry for this side fix, but I really hate when code alignment is broken, especially in unit test where perf isn't that key...

kapilt:
so rather than just using a completely unpinned name, i think this should include the pyproject.toml dependency specification if its a direct dep, albeit translated to some minimal pip compatible semantic (https://www.python.org/dev/peps/pep-0440/#version-specifiers)

gsemet:
Yes I mean what I give as restriction in pyproject.toml

poetry/utils/exporter.py
```diff
@@ -96,6 +99,9 @@ def _export_requirements_txt(
                 line = "{}".format(package.source_url)
                 if package.develop and package.source_type == "directory":
                     line = "-e " + line
+            elif no_freeze:
+                dependency = package.to_dependency()
+                line = package.name
```
kapilt: afaics, this strips the dependency to just being the naked package name, sans the pyproject.toml specification.

gmelikov: +1, it would be great to move soft deps like `1.0.*` as is too.

Secrus:
Export was moved to [`poetry-plugin-export`](https://github.com/python-poetry/poetry-plugin-export) package. Please refer to it with further issues and PRs.

Issue #2042, opened 2020-02-15
pip check fails because of the importlib-metadata version


sobolevn:
- [x] I am on the [latest](https://github.com/python-poetry/poetry/releases/latest) Poetry version.
- [x] I have searched the [issues](https://github.com/python-poetry/poetry/issues) of this repo and believe that this is not a duplicate.
- [x] If an exception occurs when executing a command, I executed it again in debug mode (`-vvv` option).

- **OS version and name**: 
- **Poetry version**: 1.0.3
- **Link of a [Gist](https://gist.github.com/) with the contents of your pyproject.toml file**: https://github.com/wemake-services/wemake-django-template/blob/b4102849daf420bc3a6af673ff5cfd8af9a70835/%7B%7Bcookiecutter.project_name%7D%7D/pyproject.toml

When installing `poetry` with `pip install` inside the Docker container, `pip check` command fails.
Dockerfile: https://github.com/wemake-services/wemake-django-template/blob/master/%7B%7Bcookiecutter.project_name%7D%7D/docker/django/Dockerfile

Current solution is to limit the `importlib-metadata` version to `<1.2.0,>=1.1.3` in `pyproject.toml`

But, I guess the full solution is to remove version restriction from `poetry`'s side.

abn:
@sobolevn does this still happen with 1.0.5 in your environment?

sobolevn:
Yes, this still happens:

```
poetry 1.0.5 has requirement importlib-metadata<1.2.0,>=1.1.3; python_version < "3.8", but you have importlib-metadata 1.6.0.
```

gregelin:
@abn Bumping this version because I am having the same issues as @sobolevn.

finswimmer:
At the moment the version for `importlib-metadata` is fixed to `~1.1.3`. I guess we can set it to `^1.1.3`, can we @sdispater ?

sobolevn:
Maybe even `>=1.1.3`?

finswimmer:
Hello,

in the current preview release (1.1.0b3) the importlib-metadata dependency is defined as `importlib-metadata = {version = "^1.6.0", python = "<3.8"}`.

So this issue should be fixed.

fin swimmer

PR #2044, opened 2020-02-15
Adds `pip check` command to CI

### LINKED ISSUE #2043: Run `pip check` during the CI
This little command might help to find different issues with the packages installed.
Possibly related to https://github.com/python-poetry/poetry/issues/2042

sobolevn:
Refs #2042
Closes #2043

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -52,6 +52,7 @@ jobs:
       run: |
         source $HOME/.poetry/env
         poetry run pytest -q tests
+        poetry run pip check
 
   MacOS:
     needs: Linting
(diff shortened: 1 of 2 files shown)


sobolevn:
Here are the first problems identified: https://github.com/python-poetry/poetry/runs/447546650?check_suite_focus=true

abn:
@sobolevn thank you for reporting this. However, I do not think there is any clear advantage of this as this is simply testing the current environment that is already installed and managed by poetry in the CI environment. The test should be in a clean virtual environment with poetry installed from a locally built wheel perhapse.

That said the new `pip` dependency resolver should error out on installation if something is wrong anyway. So `pip check` might be redundant here. 

I am going to close this for now, but happy to hear back if this is still useful.

Issue #2046, opened 2020-02-15
poetry build error with path develop dependency


kapilt:
- [x] I am on the [latest](https://github.com/python-poetry/poetry/releases/latest) Poetry version.
- [x] I have searched the [issues](https://github.com/python-poetry/poetry/issues) of this repo and believe that this is not a duplicate.
- [x] If an exception occurs when executing a command, I executed it again in debug mode (`-vvv` option).

- **OS version and name**: OS X Catalina (10.15.2)
- **Poetry version**: 1.0.3
- **Link of a [Gist](https://gist.github.com/) with the contents of your pyproject.toml file**:  https://gist.github.com/kapilt/5b3808ad3e54f5ae83147d259b1f0cef

## Issue

Running poetry build with a directory develop dependency results in multiple errors. the initial traceback (with -vvv) is in the gist, and results from a relative path usage when using a directory dependency, simply adding a self._full_path = (self._base / self._path).resolve() to line 33 of directory_dependency works past that. with that fix the second error occurs when trying to build the wheel (post successful sdist build),

```
 [ValueError]
Directory /private/var/folders/m0/9m4c62yx3sd12290vsfcc14m0000gn/T does not seem to be a Python package

Traceback (most recent call last):
  File "/Users/kapilt/.poetry/lib/poetry/_vendor/py3.8/clikit/console_application.py", line 131, in run
    status_code = command.handle(parsed_args, io)
  File "/Users/kapilt/.poetry/lib/poetry/_vendor/py3.8/clikit/api/command/command.py", line 120, in handle
    status_code = self._do_handle(args, io)
  File "/Users/kapilt/.poetry/lib/poetry/_vendor/py3.8/clikit/api/command/command.py", line 171, in _do_handle
    return getattr(handler, handler_method)(args, io, self)
  File "/Users/kapilt/.poetry/lib/poetry/_vendor/py3.8/cleo/commands/command.py", line 92, in wrap_handle
    return self.handle()
  File "/Users/kapilt/.poetry/lib/poetry/console/commands/build.py", line 30, in handle
    builder.build(fmt)
  File "/Users/kapilt/.poetry/lib/poetry/masonry/builder.py", line 21, in build
    return builder.build()
  File "/Users/kapilt/.poetry/lib/poetry/masonry/builders/complete.py", line 55, in build
    Factory().create_poetry(tmpdir),
  File "/Users/kapilt/.poetry/lib/poetry/factory.py", line 99, in create_poetry
[...]

PR #2047, opened 2020-02-15
directory dependency - resolve relative paths

kapilt:
addresses #2046  by resolving relative references in directory paths, with this build --format=wheel and build --format=sdist both work, albeit with an unconstrained dep name in generated PKG-INFO.

# Pull Request Check List


kapilt:
looks like the exception (filenotfound on 3.5) needs to be casted for the existing test that wants a valueerror here.

PR #2064, opened 2020-02-20
Keep empty in-project venv dir when recreating venv to allow for docker volumes

TheButlah:
In this PR, I have made it so that should a venv folder be rebuilt due to it being considered broken,  the folder will be emptied but not deleted before repopulating it.

I've been running into issues when attempting to install poetry in a docker container with a mounted volume for the venv. My workflow is as follows:
1. Build the image
2. Run the development container with a volume mounted at `~/myproject/.venv` and `virtualenvs.in-project` set to `true`
3. Attempt to `poetry install`

This results in the following error:
```
The virtual environment found in /home/myuser/myproject/.venv seems to be broken.
Recreating virtualenv robotics in /home/myuser/myproject/.venv

[OSError]
[Errno 16] Device or resource busy: '/home/myuser/myproject/.venv'
```

This is because when a docker volume is mounted to a folder, you cannot delete that folder. With this fix, no attempt to delete the folder will be made, instead only the contents will be deleted. This allows poetry to install the venv to a docker volume.

### DIFF
--- a/poetry/utils/env.py
+++ b/poetry/utils/env.py
@@ -15,6 +15,7 @@ from typing import Dict
 from typing import List
 from typing import Optional
 from typing import Tuple
+from typing import Union
 
 import tomlkit
 
@@ -403,7 +404,7 @@ class EnvManager(object):
                 if venv.path.name == python:
                     # Exact virtualenv name
                     if not envs_file.exists():
-                        self.remove_venv(str(venv.path))
+                        self.remove_venv(venv.path)
 
                         return venv
 
@@ -413,7 +414,7 @@ class EnvManager(object):
 
                     current_env = envs.get(base_env_name)
                     if not current_env:
-                        self.remove_venv(str(venv.path))
+                        self.remove_venv(venv.path)
 
                         return venv
 
@@ -421,7 +422,7 @@ class EnvManager(object):
                         del envs[base_env_name]
                         envs_file.write(envs)
 
-                    self.remove_venv(str(venv.path))
+                    self.remove_venv(venv.path)
 
                     return venv
 
@@ -475,7 +476,7 @@ class EnvManager(object):
                     del envs[base_env_name]
                     envs_file.write(envs)
 
-        self.remove_venv(str(venv))
+        self.remove_venv(venv)
 
         return VirtualEnv(venv)
 
@@ -621,7 +622,7 @@ class EnvManager(object):
                 "Creating virtualenv <c1>{}</> in {}".format(name, str(venv_path))
             )
 
-            self.build_venv(str(venv), executable=executable)
+            self.build_venv(venv, executable=executable)
         else:
             if force:
                 if not env.is_sane():
@@ -633,8 +634,8 @@ class EnvManager(object):
                 io.write_line(
                     "Recreating virtualenv <c1>{}</> in {}".format(name, str(venv))
                 )
-                self.remove_venv(str(venv))
-                self.build_venv(str(venv), executable=executable)
+                self.remove_venv(venv)
+                self.build_venv(venv, executable=executable)
             elif io.is_very_verbose():
                 io.write_line("Virtualenv <c1>{}</> already exists.".format(name))
 
@@ -657,7 +658,9 @@ class EnvManager(object):
         return VirtualEnv(venv)
 
     @classmethod
-    def build_venv(cls, path, executable=None):
+    def build_venv(
+        cls, path, executable=None
+    ):  # type: (Union[Path,str], Optional[str]) -> ()
         if executable is not None:
             # Create virtualenv by using an external executable
             try:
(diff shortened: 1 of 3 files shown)


TheButlah:
Looks like I'm failing a test. I've fixed it now by updating the mocked out functions to reflect the changes I made in `EnvManager`.

TheButlah:
Looks like my fix worked. However, the MacOS tests are failing - it appears to be a github actions issue unrelated to my changes.

poetry/utils/env.py
```diff
@@ -677,8 +682,20 @@ def build_venv(cls, path, executable=None):
...
+            return
+
+        # Otherwise, delete all files and folders but the toplevel one
+        for filename in os.listdir(path):
+            file_path = os.path.join(path, filename)
+
+            if os.path.isfile(file_path) or os.path.islink(file_path):
+                os.remove(file_path)
+            elif os.path.isdir(file_path):
+                shutil.rmtree(file_path)
```
finswimmer: Hello @TheButlah,

thanks a lot for contribution! :+1: 

I think we can simplify this, by always deleting the content of the venv folder instead of removing the folder. We should leave a comment why we don't delete the whole folder, otherwise there is a good chance the code get reverted at some time.

To be consistent with the rest of the code, could you please use `Path` instead of `os.path`?

fin swimmer

TheButlah: Hey @finswimmer!

Sure, I can get the code to use `Path` instead of `os.path`!

When I added the option to either delete or keep the virtualenv folder, I wasn't sure under which any situations the folder iteself should be deleted. For example, I noticed that poetry seems to randomly (?) generate a string for the venv -would there be a situation where the venv gets renamed due to this or any other reason? Because if so ostensibly poetry should delete the old env along with its folder right?

TheButlah:
Looks like my type annotations broke the python 2.7 version. Should I remove them? I thought poetry used features that are only present on versions more recent than 2.7?

TheButlah:
Removed the type annotations I added

poetry/utils/env.py
TheButlah: @finswimmer just following on my question. Is it actually safe to always leave the folder? I would expect that it should only be left when `in-project` is set to `True`

finswimmer: Hello @TheButlah, the name for the venv folder isn't random. It's hash based on the python version used, the package name and the path. Recreating is only triggered when poetry detects, that the venv folder it wants to use, is broken. So I think deleting the content of the folder and not the folder directly is always possible.

TheButlah: @finswimmer OK, sounds good. I made `remove_venv` simply empty the folder contents in all scenarios. Requesting a review.

TheButlah:
@finswimmer I resolved the comment so we can continue the discussion, if needed, here. I believe this PR is ready for final review.

poetry/utils/env.py
```diff
@@ -682,21 +683,33 @@ def build_venv(cls, path, executable=None):
...
+    def remove_venv(cls, path):  # type: (Path) -> None
+        assert path.is_dir()
+
+        # Delete all files and folders but the toplevel one. This is because sometimes
+        # the venv folder is mounted by the OS, such as in a docker volume. In such
+        # cases, an attempt to delete the folder itself will result in an `OSError`.
+        # See https://github.com/python-poetry/poetry/pull/2064
+        for file_path in path.iterdir():
+
+            if file_path.isfile() or file_path.is_symlink():
```
finswimmer: Small mistake: Using pathlib's Path it must be `is_file()`.

TheButlah:
Sounds good. Shall I write all the tests under the assumption that the directory will be successfully deleted, as before? And then just provide a new, separate test case that checks to ensure the directory remains but is empty in the event that an OSError is thrown? Or should I somehow test out both OSError and no error in all the existing tests (I'm not sure how I would go about that)?

finswimmer:
> Shall I write all the tests under the assumption that the directory will be successfully deleted, as before? And then just provide a new, separate test case that checks to ensure the directory remains but is empty in the event that an OSError is thrown?

This is what I would do :smiley:

TheButlah:
OK @finswimmer sorry for the delay, but I've updated the code so that we check for the OSError and handle that case. Let me know if there's anything else do be done!

TheButlah:
@finswimmer any updates on this? It should be ready for a merge. The mac tests fail but I think thats a CI issue on your end - I ran the tests on my mac locally and it worked just fine.

TheButlah:
@sdispater @finswimmer any updates on this?

ATGardner:
I suspect this is what currently blocks me from mounting the .venv folder as a separate mount point from the rest of my python project, inside a docker container.
Will it be merged anytime soon?

TheButlah:
I gave up constantly rebasing and pinging because no one responded :(

ATGardner:
Yeah, I don't blame you. I'm just trying to raise awareness to this PR again, to hopefully get it in there. Seems like a simple and clean solution for that problem. Thank you for making it in the first place.

TheButlah:
Great! Thanks :)

PR #2065, opened 2020-02-20
fix: use correct badge on README

MichaelAquilina:
Noticed that the README status badge was incorrectly showing tests as failing even though the latest commits to master were passing.

I've updated the badge so that its filtered to only commits in the master branch and also added a link to the corresponding workflows page.

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,7 +7,7 @@ ensuring you have the right stack everywhere.
 
 It supports Python 2.7 and 3.4+.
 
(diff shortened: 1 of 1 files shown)


MichaelAquilina:
Not sure why MacOS tests would fail from a test like this :thinking:

sdispater:
Thanks for your contribution!

The test issues have been fixed on `master`. Could you rebase your branch onto `master`?

MichaelAquilina:
Hi @sdispater, I've just rebased :) :crossed_fingers:

sdispater:
Thanks!

Issue #2072, opened 2020-02-21
pip-installing this codebase fails


bryant1410:
- [x] I am on the [latest](https://github.com/python-poetry/poetry/releases/latest) Poetry version.
- [x] I have searched the [issues](https://github.com/python-poetry/poetry/issues) of this repo and believe that this is not a duplicate.
- [x] If an exception occurs when executing a command, I executed it again in debug mode (`-vvv` option).

- **OS version and name**: Ubuntu 18.04
- **Poetry version**: from master (96191ca)
- **Link of a [Gist](https://gist.github.com/) with the contents of your pyproject.toml file**: N/A

## Issue

The following command fails in a new Python 3.6.8 env, with the latest pip, setuptools, and wheel (pip==20.0.2, setuptools==45.2.0, wheel==0.34.2).

```bash
pip install git+https://github.com/python-poetry/poetry
```

<details>
  <summary>Command output</summary>

  ```
  Collecting git+https://github.com/python-poetry/poetry
    Cloning https://github.com/python-poetry/poetry to /tmp/pip-req-build-6cd8p6hc
    Running command git clone -q https://github.com/python-poetry/poetry /tmp/pip-req-build-6cd8p6hc
    Installing build dependencies ... done
    Getting requirements to build wheel ... error
    ERROR: Command errored out with exit status 1:
     command: /[...]/.pyenv/versions/3.6.8/envs/test/bin/python3.6 /[...]/.pyenv/versions/3.6.8/envs/test/lib/python3.6/site-packages/pip/_vendor/pep517/_in_process.py get_requires_for_build_wheel /tmp/tmp61iyu9vu
         cwd: /tmp/pip-req-build-6cd8p6hc
    Complete output (36 lines):
    Traceback (most recent call last):
      File "[...]/.pyenv/versions/3.6.8/envs/test/lib/python3.6/site-packages/pip/_vendor/pep517/_in_process.py", line 257, in <module>
        main()
      File "[...]/.pyenv/versions/3.6.8/envs/test/lib/python3.6/site-packages/pip/_vendor/pep517/_in_process.py", line 240, in main
        json_out['return_val'] = hook(**hook_input['kwargs'])
      File "[...]/.pyenv/versions/3.6.8/envs/test/lib/python3.6/site-packages/pip/_vendor/pep517/_in_process.py", line 91, in get_requires_for_build_wheel
        return hook(config_settings)
[...]

PR #2090, opened 2020-02-26
Add poetry requirements to build-system requirements

thejcannon:
This is a "fix" for #2072 , however I don't really consider it a long-term fix. I have verified it does allow `pip install`-ing the package however, by doing just that t this fork 😄.

# Pull Request Check List

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -77,7 +77,7 @@ poetry = "poetry.console:main"
 
 
 [build-system]
-requires = ["intreehooks"]
+requires = ["intreehooks", "clikit"]
 build-backend = "intreehooks:loader"
 
 [tool.intreehooks]
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -88,6 +88,7 @@ requires = [
     "tomlkit >=0.5.8, <1.0.0",
     "pkginfo >=1.4, <2.0.0",
     "cachy >=0.3.0, <1.0.0",
+    "html5lib >=1.0, <2.0.0",
     "cachecontrol [filecache] >=0.12.4, <1.0.0",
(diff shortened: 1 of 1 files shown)


pyproject.toml
```diff
@@ -77,7 +77,21 @@ poetry = "poetry.console:main"
 
 
 [build-system]
-requires = ["intreehooks"]
+requires = [
+    "intreehooks",
+    # According to PEP 517 we shouldn't rely on external dependencies.
```
thejcannon: Should a comment be made above where we list the requirements to also update this list?

thejcannon:
If we _really_ wanted to, a test could be written that uses [`modulefinder`](https://docs.python.org/3.6/library/modulefinder.html) to find what packages `poetry/masonry/api.py` uses, and ensure this list matches the result. Additionally, the test(s) could correlate the `dependencies` caret versions with the PEP 440 versions here.

It would programmatically ensure this package could be `pip` installed from Git. But is also hardcore.

adisbladis:
This "fix" will make it almost impossible to properly solve environments where the _application_ is depending on any of these libraries.

My two cents: The vendoring approach used by `get-poetry.py` (and the Nix expressions I maintain to package Poetry) seems to be the only sane way to separate _application dependencies_ from _package manager dependencies_.

thejcannon:
If I understand correctly, that would only be true if the user has installed `poetry` in the same environment as their application, right? In that case, they already have these dependencies.

Isn't the advice to have `poetry` installed outside of the application environment?

finswimmer:
Hello @thejcannon,

in the meantime the issue you've linked is fixed. So I'm closing yours.

fin swimmer

thejcannon:
Thanks a lot for the tool 👍

Issue #2093, opened 2020-02-27
cannot import name create_environment with virtualenv >= 20


cjw296:
- [X] I am on the [latest](https://github.com/python-poetry/poetry/releases/latest) Poetry version.
- [ ] I have searched the [issues](https://github.com/python-poetry/poetry/issues) of this repo and believe that this is not a duplicate.
- [ ] If an exception occurs when executing a command, I executed it again in debug mode (`-vvv` option).

- **Poetry version**: 1.0.3
- **pyproject.toml file**: https://github.com/cjw296/carly/blob/master/pyproject.toml

## Issue

virtualenv 20 was released earlier this month, and looks like CircleCI's images were just rebuilt to use it. So, now, on Python 2, `poetry install` fails as follows:

```
circleci@32ce45b8e19b:~/project$ poetry install -v
Creating virtualenv carly-3aSsmiER-py2.7 in /home/circleci/.cache/pypoetry/virtualenvs

[ImportError]
cannot import name create_environment

Traceback (most recent call last):
  File "/usr/local/lib/python2.7/site-packages/clikit/console_application.py", line 131, in run
    status_code = command.handle(parsed_args, io)
  File "/usr/local/lib/python2.7/site-packages/clikit/api/command/command.py", line 120, in handle
    status_code = self._do_handle(args, io)
  File "/usr/local/lib/python2.7/site-packages/clikit/api/command/command.py", line 163, in _do_handle
    self._dispatcher.dispatch(PRE_HANDLE, event)
  File "/usr/local/lib/python2.7/site-packages/clikit/api/event/event_dispatcher.py", line 22, in dispatch
    self._do_dispatch(listeners, event_name, event)
  File "/usr/local/lib/python2.7/site-packages/clikit/api/event/event_dispatcher.py", line 89, in _do_dispatch
    listener(event, event_name, self)
  File "/usr/local/lib/python2.7/site-packages/poetry/console/config/application_config.py", line 89, in set_env
    env = env_manager.create_venv(io)
  File "/usr/local/lib/python2.7/site-packages/poetry/utils/env.py", line 610, in create_venv
    self.build_venv(str(venv), executable=executable)
  File "/usr/local/lib/python2.7/site-packages/poetry/utils/env.py", line 674, in build_venv
    from virtualenv import create_environment
```

The most pragmatic solution, at this point, is to have poetry pin the virtualenv it uses to `<20`, or just vendor it, since realistically, it's only going to be used on Python 2, right?

[...]

PR #2096, opened 2020-02-27
fix(utils.env): import cli_run from virtualenv

finswimmer:
In virtualenv v20 the method `create_environment` is replaced by `cli_run`.

This PR fixes this.

Fixes: #2093


sdispater:
Thanks!

PR #2121, opened 2020-03-02
remove 'm' and 'd' ABI tags for Python 3.8 wheels

koehlma:
Hey everyone,

as of Python 3.8 the `m` and `d` ABI flags became obsolete (see [1] and [2]). In fact, if present, they do prevent the installation of wheels. However, poetry seems to add those flags anyway. As a result, I cannot install wheels build with `poetry build` on the same system they were built on.

This pull request fixes a `RuntimeWarning` if the flags are not present in the environment and prevents the flags from being added to the ABI tag if the Python version is `>= 3.8`.

Cheers,
Maximilian

[1]: https://docs.python.org/3/whatsnew/3.8.html#build-and-c-api-changes
[2]: https://docs.python.org/3/whatsnew/3.8.html#debug-build-uses-the-same-abi-as-release-build

Resolves: https://github.com/sdispater/pendulum/issues/456

### DIFF
--- a/poetry/masonry/utils/tags.py
+++ b/poetry/masonry/utils/tags.py
@@ -75,11 +75,16 @@ def get_abi_tag(env):
         if get_flag(
             env,
             "Py_DEBUG",
-            lambda: hasattr(sys, "gettotalrefcount"),
-            warn=(impl == "cp"),
+            lambda: (hasattr(sys, "gettotalrefcount") and env.version_info < (3, 8)),
+            warn=(impl == "cp" and env.version_info < (3, 8)),
         ):
             d = "d"
-        if get_flag(env, "WITH_PYMALLOC", lambda: impl == "cp", warn=(impl == "cp")):
+        if get_flag(
+            env,
+            "WITH_PYMALLOC",
+            lambda: (impl == "cp" and env.version_info < (3, 8)),
+            warn=(impl == "cp" and env.version_info < (3, 8)),
+        ):
             m = "m"
         if get_flag(
             env,
--- a/poetry/masonry/utils/tags.py
+++ b/poetry/masonry/utils/tags.py
@@ -75,25 +75,23 @@ def get_abi_tag(env):
         if get_flag(
             env,
             "Py_DEBUG",
-            lambda: (hasattr(sys, "gettotalrefcount") and env.version_info < (3, 8)),
+            lambda: hasattr(sys, "gettotalrefcount"),
             warn=(impl == "cp" and env.version_info < (3, 8)),
         ):
             d = "d"
-        if get_flag(
-            env,
-            "WITH_PYMALLOC",
-            lambda: (impl == "cp" and env.version_info < (3, 8)),
-            warn=(impl == "cp" and env.version_info < (3, 8)),
-        ):
-            m = "m"
-        if get_flag(
-            env,
-            "Py_UNICODE_SIZE",
-            lambda: sys.maxunicode == 0x10FFFF,
-            expected=4,
-            warn=(impl == "cp" and env.version_info < (3, 3)),
-        ) and env.version_info < (3, 3):
-            u = "u"
+        if env.version_info < (3, 8):
+            if get_flag(
+                env, "WITH_PYMALLOC", lambda: impl == "cp", warn=(impl == "cp")
+            ):
+                m = "m"
+            if get_flag(
+                env,
+                "Py_UNICODE_SIZE",
+                lambda: sys.maxunicode == 0x10FFFF,
+                expected=4,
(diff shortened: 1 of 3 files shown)


nirvana-msu:
I can confirm I faced the exact same issue. Any binary wheel built with Poetry 1.0.5 for Python 3.8 under Windows will have `m` appended to the tag - which in turn makes the wheel impossible to install as per [pypa/pip/issues/6885](https://github.com/pypa/pip/issues/6885) & [pypa/pip/pull/6874](https://github.com/pypa/pip/pull/6874)

@sdispater @stephsamson this seems like critical issue, with a very simple fix. Would be great if you find some time to review/merge this PR. Btw even `pendulum` wheel is broken due to the same issue: [sdispater/pendulum/issues/456](https://github.com/sdispater/pendulum/issues/456)

koehlma:
> It would be great if you could write a test for this -- one doesn't exist yet but you can put it under `tests/masonry/utils`.

In principle I could do that, however, I find it a bit hard to get started as I do not know where the `env` object is coming from and what its general structure is. If someone could write a skeleton and construct a mock `env`, I will provide a test ASAP. Unfortunately, I currently lack the time to dig deeper into this.

poetry/masonry/utils/tags.py
```diff
@@ -75,11 +75,16 @@ def get_abi_tag(env):
         if get_flag(
             env,
             "Py_DEBUG",
-            lambda: hasattr(sys, "gettotalrefcount"),
-            warn=(impl == "cp"),
+            lambda: (hasattr(sys, "gettotalrefcount") and env.version_info < (3, 8)),
+            warn=(impl == "cp" and env.version_info < (3, 8)),
```
frostming, line 79: debug flag "u" doesn't require python < 3.8

poetry/masonry/utils/tags.py
```diff
@@ -75,11 +75,16 @@ def get_abi_tag(env):
...
         ):
             d = "d"
-        if get_flag(env, "WITH_PYMALLOC", lambda: impl == "cp", warn=(impl == "cp")):
+        if get_flag(
+            env,
+            "WITH_PYMALLOC",
+            lambda: (impl == "cp" and env.version_info < (3, 8)),
```
frostming, line 85: Python < 3.8 should be mandatory check instead of fallback value

This line should be changed to:
```python
if env.version_info < (3, 8) and get_flag(...):
```
See reference implementation: https://github.com/pypa/pip/blob/e40d2670facb6cfc815f948c05ca73abb3200f50/src/pip/_vendor/packaging/tags.py#L173-L184

frostming:
@koehlma I added some review comment.

For testing, you can construct a mocked env with
```python
from poetry.utils.env import MockEnv

env = MockEnv(version_info=(3, 8, 0))
```

poetry/masonry/utils/tags.py
nirvana-msu: That makes sense, although  I don't think `WITH_PYMALLOC` can be overridden by env var as you say. I checked the code of `sysconfig.get_config_vars` (as I was looking for a workaround, while waiting on this fix) and there seems to be no way to actually change the behavior. Which makes this issue even more critical as there really seems to be no workaround (?)

frostming: I made my previous comment wrong, it can't be changed by env var, but the fallback only takes effect when there is no value returned by `sysconfig.get_config_vars()`, which is not the case for `WITH_PYMALLOC` under `python 3.8`. So with the previous implementation, the value is NOT changed at all.

frostming: After all, follow the reference implementation used by `pip` would make things right.

poetry/masonry/utils/tags.py
koehlma: I get the following runtime warning with Python 3.8:
```
RuntimeWarning: Config variable 'Py_DEBUG' is unset, Python ABI tag may be incorrect
```
So, at least there should be no warning if the flag is not present.

frostming: It makes sense

koehlma:
Here you go. I wrote a test and adapted the `MockEnv` such that I can overwrite config variables.

tests/masonry/utils/test_tags.py
```diff
@@ -0,0 +1,23 @@
+from poetry.masonry.utils.tags import get_abi_tag
+from poetry.utils.env import MockEnv
+
+
+def test_tags_cpython38():
```
abn, line 5: Would be great to add a `< 3.8` case as well here.

koehlma: There now is a test for CPython 3.7 as well.

PR #2127, opened 2020-03-04
Add support to use repository url when unspecified

greghaynes:
Currently, even if a url is defined for a user level configured
repository we require it also be specified in the project's
pyproject.toml. With this change, if no url is specified in a
pyproject.toml a user level configured repository url is used as a
fallback. This also improves the error message when no URL is found for
a repository.

# Pull Request Check List

### DIFF
--- a/docs/docs/repositories.md
+++ b/docs/docs/repositories.md
@@ -121,6 +121,11 @@ a custom certificate authority or client certificates, similarly refer to the ex
 looking for packages.
 
 
+If no url is specified for a repository in your `pyproject.toml`, poetry will fall
+back to the url configured for a corresponding repository.foo given a source with
+name = foo.
+
+
 ### Disabling the PyPI repository
 
 If you want your packages to be exclusively looked up from a private
--- a/poetry/factory.py
+++ b/poetry/factory.py
@@ -236,12 +236,18 @@ class Factory:
         from .utils.helpers import get_client_cert, get_cert
         from .utils.password_manager import PasswordManager
 
+        if "name" not in source:
+            raise RuntimeError("Missing [name] in source.")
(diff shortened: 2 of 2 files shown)


greghaynes:
I havent added a test for this - looking through our fixtures we dont seem to have a case for specifying user-level repositories? I am not sure how much effort it would be to add that.

finswimmer:
Hello @greghaynes ,

thanks a lot for your contribution!

However, I believe it is not a good idea to implement it. In my opinion one should always be able to see in the `pyproject.toml` what resources are needed to install/build the package.

Maybe @sdispater has a different opinion about this?

fin swimmer

greghaynes:
@finswimmer Thanks for the reply!

That's an interesting idea, and I can understand the motivation. It is a significant departure from how folks often use existing tools (e.g. pip.conf) though and so at a minimum I think we need to be explicit about this. I'd be happy to throw up some added docs if we think it through and agree it makes sense.

One concern I have with this design is, it is very common IME for CI setups to run a pypi mirror for the sake of stability. In this setup, it's not desired to include the mirror in the final product / source as you don't want external users consuming your mirror resources (if they even have access to it). The solution is generally to install a user or system wide index url in pip.conf. Although this PR doesn't allow us to support this use case fully, I don't understand how we can reconcile your proposal to always specify repositories in the repo's pyproject.toml with this use case?

greghaynes:
I think there's some additional motivation to support this use case given this recent pypa-dev ML: https://groups.google.com/forum/#!topic/pypa-dev/Pdnoi8UeFZ8

Secrus:
Thank you for contributing. This has been decided to be rejected as it will break features around the lockfile.

PR #2130, opened 2020-03-05
EditableBuilder: use pip with `--no-deps`

### LINKED ISSUE #2126: Why is the EditableBuilder running pip without --no-deps ?
We've got a  problem involving [this repo](https://github.com/TankerHQ/sdk-python) at work and we are still trying to figure out exactly what the problem is - when we do, we'll update this issue :)

But during our investigation, we came across these lines in poet's `EditableBuilder` sources:

```python
class EditableBuilder(Builder):
    def build(self):
        return self._setup_build()

    def _setup_build(self):
        builder = SdistBuilder(self._poetry, self._env, self._io)
        setup = self._path / "setup.py"
        ...
        try:
           ...
           self._env.run_pip("install", "-e", str(self._path))
        finaly:
          ...
```

My question is: why is pip run _without_ the `--no-deps` flags? Is this an oversight, or is something else going on here?

dmerejkowsky:
Notes:

Fix #2126 

* Technically this is a change in behavior, albeit a subtle one, but I think we should update the Change Log anyway
* This fixes the issue I talked about in #2126 - go there to learn about the gory details

### DIFF
--- a/poetry/masonry/builders/editable.py
+++ b/poetry/masonry/builders/editable.py
@@ -31,14 +31,14 @@ class EditableBuilder(Builder):
 
         try:
             if self._env.pip_version < Version(19, 0):
-                self._env.run_pip("install", "-e", str(self._path))
+                self._env.run_pip("install", "-e", str(self._path), "--no-deps")
             else:
                 # Temporarily rename pyproject.toml
                 shutil.move(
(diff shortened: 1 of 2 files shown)


dmerejkowsky:
Looks like this was superseeed by #2360 , closing

Issue #2115, opened 2020-03-02
"Unable to detect the current shell." when running Vagrant command


dms-cat:
- [x] I am on the [latest](https://github.com/python-poetry/poetry/releases/latest) Poetry version.
- [x] I have searched the [issues](https://github.com/python-poetry/poetry/issues) of this repo and believe that this is not a duplicate.
- [x] If an exception occurs when executing a command, I executed it again in debug mode (`-vvv` option).

- **OS version and name**: Ubuntu 18.04
- **Poetry version**: 1.0.5
- **Link of a [Gist](https://gist.github.com/) with the contents of your pyproject.toml file**: https://gist.github.com/dms-cat/651c4fe2e55a83984b6cdb436eb102a1

## Issue

`vagrant ssh default --command='poetry shell'` fails with

> Unable to detect the current shell.

If I instead run `vagrant ssh default` and then `poetry shell` it works fine.

I've also tried variants like `vagrant ssh default --command='bash -ic "poetry shell"'` with the same result.

sambonner:
It looks the maintainer of Shellington [recommends providing a default shell as a fallback](https://github.com/sarugaku/shellingham#notes-for-application-developers). Is there a reason why this was not implemented for Poetry?

finswimmer:
Hello @sambonner,

thanks a lot for reporting and linking to a possible solution.

Fix is on the way: #2147 

fin swimmer

johannesgiani:
@dms-cat can you provide an example vagrant file of how you provision poetry? I am stuck here.

PR #2147, opened 2020-03-08
try to find shell via os.environ if detect_shell fail (#2115)

finswimmer:
If poetry is not run within a shell, like explained in #2115, shell detection fails.

[shellingham](https://github.com/sarugaku/shellingham#notes-for-application-developers) itself explain how one can try to detect a default shell via environment variables. This PR implements the suggestion.

Fixes: #2115

### DIFF
--- a/poetry/utils/shell.py
+++ b/poetry/utils/shell.py
@@ -42,17 +42,18 @@ class Shell:
 
         try:
             name, path = detect_shell(os.getpid())
-        except ShellDetectionFailure:
+        except (RuntimeError, ShellDetectionFailure):
+            shell = None
+
             if os.name == "posix":
-                shell = os.environ["SHELL"]
+                shell = os.environ.get("SHELL")
             elif os.name == "nt":
-                shell = os.environ["COMSPEC"]
-            else:
+                shell = os.environ.get("COMSPEC")
+
+            if not shell:
                 raise RuntimeError("Unable to detect the current shell.")
 
             name, path = Path(shell).stem, shell
-        except RuntimeError:
-            raise RuntimeError("Unable to detect the current shell.")
 
         cls._shell = cls(name, path)
 
--- a/poetry/utils/shell.py
+++ b/poetry/utils/shell.py
@@ -9,6 +9,7 @@ from shellingham import ShellDetectionFailure
 from shellingham import detect_shell
 
 from ._compat import WINDOWS
+from ._compat import Path
 from .env import VirtualEnv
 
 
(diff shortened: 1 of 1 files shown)


poetry/utils/shell.py
```diff
@@ -9,6 +9,7 @@
 from shellingham import detect_shell
 
 from ._compat import WINDOWS
+from ._compat import Path
```
dms-cat: Should that be merged with the previous line?

poetry/utils/shell.py
```diff
@@ -41,7 +42,16 @@ def get(cls):  # type: () -> Shell
...
+        except ShellDetectionFailure:
+            if os.name == "posix":
+                shell = os.environ["SHELL"]
+            elif os.name == "nt":
+                shell = os.environ["COMSPEC"]
+            else:
+                raise RuntimeError
+
+            name, path = Path(shell).stem, shell
+        except RuntimeError:
```
dms-cat: Did you intend to catch the inner `RuntimeError` here to add the message to it? Because it won't. Try the following test.py:

```python
class FirstException(Exception):
    pass


class SecondException(Exception):
    pass


try:
    raise FirstException()
except FirstException:
    raise SecondException()
except SecondException:
    print("Caught")
```

Result:

```text
$ python3 test.py 
  File "test.py", line 10, in <module>
    raise FirstException()
__main__.FirstException

During handling of the above exception, another exception occurred:

  File "test.py", line 12, in <module>
    raise SecondException()
__main__.SecondException
```

poetry/utils/shell.py
```diff
@@ -41,7 +42,16 @@ def get(cls):  # type: () -> Shell
...
         try:
             name, path = detect_shell(os.getpid())
-        except (RuntimeError, ShellDetectionFailure):
+        except ShellDetectionFailure:
+            if os.name == "posix":
+                shell = os.environ["SHELL"]
+            elif os.name == "nt":
+                shell = os.environ["COMSPEC"]
+            else:
+                raise RuntimeError
```
dms-cat: Should this maybe be a custom `UnknownShellError` or something like it?

poetry/utils/shell.py
finswimmer: poetry's isort configuration is preferring keeping this separated.

poetry/utils/shell.py
finswimmer: I stopped counting how often I made this mistake :smile: You are absolutely right.

poetry/utils/shell.py
finswimmer: I think it's more simply to just repeat the error message here. A custom `ErrorClass` looks a bit to much to me here.

Thanks a lot for your valuable input :+1:

poetry/utils/shell.py
```diff
@@ -41,7 +42,16 @@ def get(cls):  # type: () -> Shell
...
+        except ShellDetectionFailure:
+            if os.name == "posix":
+                shell = os.environ["SHELL"]
+            elif os.name == "nt":
+                shell = os.environ["COMSPEC"]
+            else:
+                raise RuntimeError("Unable to detect the current shell.")
+
+            name, path = Path(shell).stem, shell
+        except RuntimeError:
```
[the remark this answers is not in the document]
finswimmer: According to https://github.com/sarugaku/shellingham#notes-for-application-developers the env vars should always exists. But yes, testing it doesn't hurt.

PR #2150, opened 2020-03-08
Use dark_gray color for resolving dependencies msg instead of black+bold

petobens:
In terminals where black+bold != gray the elapsed time is shown in black
and hence it's invisible to the user.

Details: https://github.com/hishamhm/htop/issues/35#issuecomment-266465736

### DIFF
--- a/poetry/puzzle/provider.py
+++ b/poetry/puzzle/provider.py
@@ -807,7 +807,7 @@ class Provider:
             yield
         else:
             indicator = Indicator(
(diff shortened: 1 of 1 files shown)


petobens:
In the following image you can see the mentioned problem (and the fix) on kitty terminal:
Screenshot_2020-03-08_10:27:33

petobens:
Deprecated by https://github.com/python-poetry/poetry/pull/2230

PR #2172, opened 2020-03-11
Add a --remove-untracked option to the install command.

### LINKED ISSUE #648: How do I uninstall packages that have been removed from pyproject.toml?

## Question

pipenv has `pipenv clean` which removes all packages that are no longer in Pipfile.lock from the virtualenv. What is the equivalent with poetry? The docs say there is no need for `pipenv sync` and `pipenv clean` due to the `remove` command https://github.com/sdispater/poetry#remove-command

But how does this work? I would need to run `poetry remove XXX` against all my virtualenvs - manually keeping track of all packages that have been removed and remembering to remove them myself?

Is there not any way to sync my poetry.lock to my virtualenv to ensure the virtualenv only has the packages specified in poetry.lock installed?

I have a virtualenv on each box I deploy to, and keep pyproject.toml and poetry.lock under source code control. I then want a way to ensure each of those virtualenvs only has the packages in poetry.lock installed.

PetterS:
Fixes #648. 

As discussed in the linked issue. @finswimmer seemed to approve this flag, @sdispater. 

This is a feature that we are missing since moving away from pipenv. Removing old packages is part of reproducibility.


PetterS:
Could anyone have a look at this? I am happy to finish this up with tests and documentation.

Mattwmaster58:
I'm not a poetry team member, but I took a quick look at this anyway, and I have to say you did a thorough job as far as I can tell 👍

PetterS:
Thanks Matt!

@stephsamson, do you think you could take a look?

ulgens:
How does this handle packages like pip and setuptools which are not in lock file but should be kept in venv?

PetterS:
> How does this handle packages like pip and setuptools which are not in lock file but should be kept in venv?

This is a great question and the answer is that, unfortunatley, currently it does not. This needs to be addressed.

PetterS:
@sdispater did you have a chance to look at this? This is functionality that we really would like to have in poetry (we used `pipenv clean` before).

Thanks,
Petter

PetterS:
Is there something I could do here, besides waiting for a review? We'd love to have this in Poetry!

ayroblu:
Bumping this cause we're waiting on this

PetterS:
> Bumping this cause we're super interested in this

Yes, the lack of this flag has been a pain point in my team since we switched from pipenv to poetry. People are `rm -rf` their venvs and installing from scratch.

How can we proceed here? @stephsamson @sdispater @finswimmer

PetterS:
@stephsamson Done. Moved the logic to the solver.

poetry/puzzle/solver.py
```diff
@@ -21,14 +21,15 @@
 
 
 class Solver:
-    def __init__(self, package, pool, installed, locked, io):
+    def __init__(self, package, pool, installed, locked, io, remove_untracked=False):
```
stephsamson, line 24: A very minor change needed as we make a concentrated effort to clean up the Poetry codebase by adding type annotations whenever we touch a piece of code. Thank you! 🙌 

```suggestion
    def __init__(
        self,
        package,  # type: ProjectPackage
        pool,  # type: Pool
        installed,  # type: Repository
        locked,  # type: Repository
        io,  # type: ConsoleIO
        remove_untracked=False  # type: bool
    ):
```

PetterS: Thanks. I really like type annotations.

PetterS:
OK, those type annotations did not work. I'll try to fix them.

poetry/puzzle/solver.py
snejus: Is there a reason behind adding them in the comments rather than directly?

Issue #2020, opened 2020-02-11
Add support for named Other/Proprietary Licenses


cfarrend:
- [x] I have searched the [issues](https://github.com/python-poetry/poetry/issues) of this repo and believe that this is not a duplicate.
- [x] I have searched the [documentation](https://python-poetry.org/docs/) and believe that my question is not covered.

### Currently
When you want to use the Other/Proprietary License license type you specify in your `pyproject.toml` file:
```toml
license = "Proprietary"
```

When uploaded to PyPi, the license shows as the following in the metadata section:
```
License: Other/Proprietary License (Proprietary)
```
### Desired
We wish to specify our own License name for the Other/Proprietary License instead of the generic "Proprietary" such that we can put in links to our own licenses, or provide the name of the license if not natively supported by PyPi. Some examples are here:
* https://pypi.org/project/aws-kinesis-agg/
* https://pypi.org/project/cdat-lite/
* https://pypi.org/project/pytket-projectq/

### Code
Having a dig around the code seems to be around here
https://github.com/python-poetry/poetry/blob/ed4434253762f943255af20e98098b19a65ca1a6/poetry/spdx/__init__.py#L48

https://github.com/python-poetry/poetry/blob/master/poetry/spdx/license.py

Issue #2198, opened 2020-03-18
Proprietary liecense option not documented


mahmoudhossam:
- [x] I have searched the [issues](https://github.com/python-poetry/poetry/issues) of this repo and believe that this is not a duplicate.

## Issue
Proprietary licenses have been supported as of #867, but the [documentation](https://python-poetry.org/docs/pyproject/#license) now only refers to open source ones without mentioning that it's possible to mark a project as having a proprietary license.

I think this should also be mentioned in the documentation.

PR #2202, opened 2020-03-19
Relax licence restrictions to support custom licences

abn:
This change adds support of use of custom licenses for projects. If the
value specified for `licence` is not in the SPDX License List or 
"Proprietary", we treat it as a custom licence withe value (as is) as 
the name.

As a side-effect, the validation of licence is no longer performed when
executing `poetry check`.

Resolves: #2020 #2198

# Pull Request Check List

### DIFF
--- a/docs/docs/pyproject.md
+++ b/docs/docs/pyproject.md
@@ -40,6 +40,10 @@ The recommended notation for the most common licenses is (alphabetical):
 Optional, but it is highly recommended to supply this.
 More identifiers are listed at the [SPDX Open Source License Registry](https://www.spdx.org/licenses/).
 
+!!!note
+
+    If your project is proprietary and does not use a specific licence, you can set this value as `Proprietary`.
+
 ## authors
 
 The authors of the package. **Required**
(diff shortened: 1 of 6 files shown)


abn:
This is probably the simplest change to support custom licences. The main reservation I had was that now, if someone makes a typo (eg: `license = "MIIT"`) it will no longer be reported as an error. Which in the grand scheme of things, might not be an issue. But figured I'd mention it here, in case an alternative behavior is preferred.

stephsamson:
> This is probably the simplest change to support custom licences. The main reservation I had was that now, if someone makes a typo (eg: `license = "MIIT"`) it will no longer be reported as an error.

Maybe a simple distance measure could be use to catch such typos.

abn:
I feel it might be adding unnecessarily complexity without much guaretee in return. And, personally, I do not think the impact of a typo in the licence warrants such a safegueard. Besides, this can be added in later if we really need it.

abn:
This might need to move to https://github.com/python-poetry/core I think.

abn:
This has now been split into two PRs.

1. Doc fix for use of proprietary licence (https://github.com/python-poetry/poetry/pull/2231)
2. Custom license support for a future version of poetry-core (https://github.com/python-poetry/core/pull/5)

The documentation fix for (2) will be created if/when the version of poetry-core is merged.

PR #2212, opened 2020-03-20
Use a lightweight core library

### LINKED ISSUE #1993: Lightweight core library
Currently, to build a Poetry project via the `PEP-517` build mechanism, you need this section:

```toml
[build-system]
requires = ["poetry>=1.0.3"]
```

This works but this has the undesirable side-effect of pulling the full Poetry stack and its dependencies. Most of them are not needed to build the project and expose the metadata.

To prevent that, we need to extract parts of the Poetry stack into a more lightweight library with the sole purpose of building projects.

Ideally, this core library should not have any dependency to avoid errors to to circular build dependencies when using `pip install --no-binary :all:`.

sdispater:
## Pull Request Check List

This PR extracts the core utilities of Poetry (like the building of distributions) into a separate package ([poetry-core](https://github.com/python-poetry/core)) that is lightweight and without any declared dependency, the dependencies are vendored instead.

This will help address the issues mentioned in #1993.

Fixes #1993


docs/docs/faq.md
```diff
@@ -42,8 +42,8 @@ So, in your `pyproject.toml` file, add this section if it does not already exist
 
 ```toml
 [build-system]
-requires = ["poetry>=0.12"]
-build-backend = "poetry.masonry.api"
+requires = ["poetry-core>=1.0.0"]
```
stephsamson: Is it possible to have `core` somehow namespaced under `poetry` so that we will instead have `requires = ["poetry.core>=1.0.0"]`? Using a dash in the package name can be quite confusing, for example, when do we use the dash and the underscore (in `docs/pyproject.md`,  it requires `poetry_core`).

pradyunsg: Both of these are equivalent to PyPI - so “_-” won't matter to PyPI. You can use poetry_core instead of poetry-core and they'll refer to the same thing in the dependency specification.

That's the interoperability standard (PEP 508/440) and I expect poetry follows it. :)

stephsamson: Ah okay, good to know, thanks!

sdispater: I would have preferred to use a namespace package but, as far as I know, it's not possible since poetry is already its own package.

It would require to no longer have the `poetry` package but multiple packages (like `poetry-core` and `poetry-cli`) so I don't think it's worth it for now.

However, if someone tells me there is a way to do it without big changes I'd be happy to change it.

sdispater: In the end, I managed to make the `poetry` namespace work, so `poetry-core` will provide the `poetry.core` package.

PR #2230, opened 2020-03-26
Improve UI

sdispater:
## Pull Request Check List

- [ ] Added **tests** for changed code.
- [ ] Updated **documentation** for changed code.

This PR changes the user interface a bit to make it more intuitive. It changes colors in some places to be more readable and provide a consistent interface on dark and light terminal themes.

It also upgrades Cleo and CliKit to provide beautiful and more intuitive errors with various levels of detail depending on the current verbosity. The rendered stack traces will also collapse duplicated frames (useful for recursion errors).

Here are some examples:

**Verbose mode**

Screenshot 2020-03-25 at 19 49 43

**Debug mode**

Screenshot 2020-03-25 at 19 51 08

**Duplicated frames**

Screenshot 2020-03-25 at 19 58 08

PR #2235, opened 2020-03-27
Release 1.1.0a1

sdispater:
This release **must** be downloaded via the `get-poetry.py` script and not via the `self update` command.

### Changed

- The core features of Poetry have been extracted in to a separate library: `poetry-core` ([#2212](https://github.com/python-poetry/poetry/pull/2212)).
- The build backend is no longer `poetry.masonry.api` but `poetry.core.masonry.api` which requires `poetry-core>=1.0.0a5` ([#2212](https://github.com/python-poetry/poetry/pull/2212)).
- The exceptions are now beautifully displayed in the terminal with various level of details depending on the verbosity ([2230](https://github.com/python-poetry/poetry/pull/2230)).


CHANGELOG.md
```diff
@@ -1,5 +1,16 @@
 # Change Log
 
+## [1.1.0a1] - 2020-03-27
+
+This release **must** be downloaded via the `get-poetry.py` script and not via the `self update` command.
```
stephsamson: I think this should also be documented in the README.md -- I'm afraid many people may not catch it if this information is only in the Changelog.

abn:
Does this break PEP 518 support for older `pyproject.toml` files when `requires=["poetry>=0.12"]` and `build-backend = "poetry.masonry.api"`?

CHANGELOG.md
sdispater: I don't think it's worth it. By the time the `1.1.0` version is out, we will have fixed it in the `self update` command.

And since it's a prerelease version that will not be officially announced on the website, the risk for people to accidentally update to this version is low.

sdispater:
@abn No, it will still work since there is a compatibility layer here https://github.com/python-poetry/poetry/blob/develop/poetry/masonry/api.py

PR #2240, opened 2020-03-28
feat(installer): add only dev mode option

max-wittig:
This is a draft PR. I'm very much open to changes and suggestions!
I'm not sure, if I maybe should name it `only-main` instead and if `no-root` should automatically apply or not.

This commandline parameter adds the ability to install only dev
packages. That is often needed, when poetry is run in a CI/CD system
like GitLab CI or GitHub actions, where often not the whole test
suite for every PR is used, but only linters or static
type checkers are required. These often reside in dev dependencies.

I think this option is needed to avoid having to use another tool like `tox` or having to export files like `requirements-black.txt` to manage things like
linting before a PR.

Example use-case:
* Pull Request only pipeline, which only executes `black`, `flake8`, `pylint`, `mypy` and other linters or static type checkers. To execute this on a branch, no main dependencies are needed.

### DIFF
--- a/docs/docs/cli.md
+++ b/docs/docs/cli.md
@@ -116,6 +116,13 @@ the `--no-dev` option.
 poetry install --no-dev
 ```
 
+In contrast, if you only need the dev dependencies to e.g. run linters, you can use the
+`only-dev` option.
+
+```bash
+poetry install --only-dev
+```
+
 You can also specify the extras you want installed
 by passing the `--E|--extras` option (See [Extras](#extras) for more info)
 
--- a/poetry/console/commands/install.py
+++ b/poetry/console/commands/install.py
@@ -10,6 +10,7 @@ class InstallCommand(EnvCommand):
 
     options = [
         option("no-dev", None, "Do not install the development dependencies."),
+        option("only-dev", None, "Only install the development dependencies."),
         option(
             "no-root", None, "Do not install the root package (the current project)."
         ),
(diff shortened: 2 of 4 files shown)


max-wittig:
@sdispater Would you mind to take a look?

max-wittig:
Any chance that anybody could take a look?

max-wittig:
@abn Would you mind to take a look?

abn:
@max-wittig I am not sure if this will behave as exected in all cases. I like having the ability to install only a subset of dependencies. However, in this implementation, if a dependency in the dev section requires something that was explicitly added in the dependencies section, it will not be installed.

Here is an example project.

```toml
[tool.poetry]
name = "dev-dep-project"
version = "0.1.0"
description = ""
authors = ["Bender Rodriguez <bender@planetexpress.com>"]

[tool.poetry.dependencies]
python = "^3.7"
attrs = "^19.3.0"
toml = "^0.10.0"

black = "^19.10b0"

[build-system]
requires = ["poetry-core>=1.0.0a5"]
```

Using the `--only-dev` option will not install `attrs` nor `toml` as required by `black`. 
```console
$ poetry install --only-dev
Creating virtualenv dev-dep-project in /tmp/dev-dep-project/.venv


Package operations: 6 installs, 0 updates, 0 removals

  - Installing appdirs (1.4.3)
  - Installing click (7.1.1)
  - Installing pathspec (0.8.0)
  - Installing regex (2020.4.4)
  - Installing typed-ast (1.4.1)
  - Installing black (19.10b0)
```

The common solutions for this include;

1. using extras for required actions
2. using tox to manage linting etc as an integration

It might be worthwhile to rethink the requirement here.

max-wittig:
@abn Thanks for your review. The problem with all these other tools is that then my test requirements are not hashed. Poetry provided a nice way to have hashed dependencies everywhere. Otherwise you have to use tools like hashin and updating these things is a pain.

max-wittig:
Was implemented in https://github.com/python-poetry/poetry/commit/d2485b8f849f0bbfd8f6933b625ea07f11f80615

Issue #458, opened 2018-09-29
Poetry does not execute in the same directory when it finds pyproject.toml


slurps-mad-rips:
- [x] I am on the [latest](https://github.com/sdispater/poetry/releases/latest) Poetry version.
- [x] I have searched the [issues](https://github.com/sdispater/poetry/issues) of this repo and believe that this is not a duplicate.
- [x] If an exception occurs when executing a command, I executed it again in debug mode (`-vvv` option).

- **OS version and name**: macOS High Sierra 10.13.6
- **Poetry version**: 0.11.5
- ~~Link of a [Gist](https://gist.github.com/) with the contents of your pyproject.toml file~~

## Issue

Hello! I've not attached a gist, as this is more a behavior problem. I *really* like that I can run poetry from any directory within my project and it will look up the path. However, when I'm running `poetry add`, or `poetry install`, it will execute from the directory I'm currently in, rather than where the `pyproject.toml` file is located. I ran into a bug that had me ripping my hair out. I realized that the package I was trying to install (`importlib_resources`) had a dependency that conflicted with the name of a module I had in the directory I was in. If it had executed from the location of the `pyproject.toml` this would not have happened. I did figure this out in the end, but it would be nice to not have to think or know about this when calling `poetry add` and friends.

sdispater:
Could you test again with the latest stable release, and also with the latest `1.0.0` prerelease to see if the issue still occurs. Closing this for now but feel free to reopen if the issue still occurs.

Issue #454, opened 2018-09-24
ValueError/TypeError with path dependency


apnewberry:
- [x] I am on the [latest](https://github.com/sdispater/poetry/releases/latest) Poetry version.
- [x] I have searched the [issues](https://github.com/sdispater/poetry/issues) of this repo and believe that this is not a duplicate.
- [x] If an exception occurs when executing a command, I executed it again in debug mode (`-vvv` option).

- **OS version and name**: Ubuntu 18.04
- **Poetry version**: 0.11.5
- **toml** below

## Issue

**Note:** the dependency uses setup.py and Pipfile/Pipfile.lock.

```toml
[tool.poetry]

name="my_demo"
version = "0.0.1"
description = ""
authors = ["me"]

[tool.poetry.dependencies]
python = "^3.6"

epype = {path = "/home/user/dotfiles/dotfiles/.emacs.d/python/epype"}

```

```python
$ poetry lock -vvv
Using virtualenv: /home/user/.cache/pypoetry/virtualenvs/epype-py3.6
Updating dependencies
Resolving dependencies...
   1: fact: my-demo is 0.0.1
   1: derived: my-demo
   1: fact: my-demo depends on epype (0.0.1)
   1: selecting my-demo (0.0.1)
   1: derived: epype (0.0.1)
   1: selecting epype (0.0.1)
   1: Version solving took 0.004 seconds.
   1: Tried 1 solutions.

                                   
[ValueError]        
Invalid type <class 'NoneType'>  
                                   
Exception trace:
 /home/user/.envs/dotfiles-_bp8OBTv/lib/python3.6/site-packages/cleo/application.py in run() at line 94
   status_code = self.do_run(input_, output_)
 /home/user/.envs/dotfiles-_bp8OBTv/lib/python3.6/site-packages/poetry/console/application.py in do_run() at line 88
   return super(Application, self).do_run(i, o)
 /home/user/.envs/dotfiles-_bp8OBTv/lib/python3.6/site-packages/cleo/application.py in do_run() at line 197
   status_code = command.run(input_, output_)
 /home/user/.envs/dotfiles-_bp8OBTv/lib/python3.6/site-packages/poetry/console/commands/command.py in run() at line 77
   return super(BaseCommand, self).run(i, o)
 /home/user/.envs/dotfiles-_bp8OBTv/lib/python3.6/site-packages/cleo/commands/base_command.py in run() at line 146
   status_code = self.execute(input_, output_)
[...]

PR #2262, opened 2020-04-03
[WIP] Add test for pip installing pendulum failing

altendky:
sdispater/pendulum#458

This PR is most likely going to be deleted but I wanted to provide a reproducible scenario for ... well, the `--no-binary` part is what I did in #458 but the error is what is reported in #454.

I will explicitly point out that this would \_not\_ use the `master` source of poetry but rather whatever pendulum claims to depend on.  Again, this is a publicly running demonstration, not a proposed test or solution.  My guess at the solution is to add `setuptools` as a poetry installation dependency so it will have setuptools when 517 results in poetry installation.  Or maybe the `build-system` `requires` for pendulum should just get a setuptools.  I dunno...

### DIFF
--- a/tests/installation/test_pip_installer.py
+++ b/tests/installation/test_pip_installer.py
@@ -9,7 +9,9 @@ from poetry.packages.package import Package
 from poetry.repositories.legacy_repository import LegacyRepository
 from poetry.repositories.pool import Pool
 from poetry.utils._compat import Path
-from poetry.utils.env import EnvManager, NullEnv, VirtualEnv
+from poetry.utils.env import EnvManager
+from poetry.utils.env import NullEnv
+from poetry.utils.env import VirtualEnv
 
 
 @pytest.fixture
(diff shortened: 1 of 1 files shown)


tests/installation/test_pip_installer.py
```diff
@@ -62,6 +67,39 @@ def test_requirement_source_type_url():
...
+    venv = VirtualEnv(venv_path)
+    yield venv
+
+    shutil.rmtree(str(venv.path))
+
+
+def test_can_install_pendulum_sdist(tmp_venv):
+    installer = PipInstaller(tmp_venv, NullIO(), Pool())
+
+    installer.run("install", "--no-binary", ":all:", "pendulum")
```
abn, line 100: This will fail today because poetry-core `1.0.0a5` cannot be installed from source. When pendulum is attempted to be built using PEP-517.

Using `pip install --no-binary pendulum` should progress further. But will fail when we hit pendulum. Which again attempts to install using setuptools.

This should be resolved once python-poetry/core#24 is merged.

abn:
I suspect this is no longer an issue. Closing this for now. Feel free to re-open if this still needs to be added into the codebase.

PR #2265, opened 2020-04-03
[Bugfix] Moved access to first element to after the length check

Qu4tro:

Hey :wave:
Quick fix for a problem I was having. My `pyproject.toml` `tool.poetry.packages` was wrong, but I was getting `[IndexError] list index out of range` which is not very helpful. 

Clearing this, helped me get an error message, that was actually helpful:
`/home/user/repo/pkg/pkg does not contain any element` 

It needs some tests, but still unsure of where to start.

### DIFF
--- a/poetry/masonry/utils/package_include.py
+++ b/poetry/masonry/utils/package_include.py
@@ -34,13 +34,12 @@ class PackageInclude(Include):
         return self.check_elements()
 
     def check_elements(self):  # type: () -> PackageInclude
-        root = self._elements[0]
-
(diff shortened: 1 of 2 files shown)


abn:
@Qu4tro thank for the fix. Can you also apply the same changes to https://github.com/python-poetry/core/blob/master/poetry/core/masonry/utils/package_include.py please?

Also, might be great to have a test case so that we do not regress later.

Qu4tro:
Done and done.

Issue #2015, opened 2020-02-10
Better documentation for how files are included in built artifacts


AstraLuma:
- [x] I have searched the [issues](https://github.com/python-poetry/poetry/issues) of this repo and believe that this is not a duplicate.

## Issue

While helping with https://github.com/licensezero/cli/issues/3, I was looking for "how to do package data with poetry" but found the docs were unclear or at least extremely implicit about it.

finswimmer:
Hello @astronouth7303,

could you please add more details about what you want to add to a package and/or what information you are missing?

fin swimmer

astronouth7303:
What setuptools calls package data: Non-module files.

The specific discussion was about `licensezero.json`, but this varies from static web assets to templates to UI definitions.

thejohnfreeman:
From what I understand, every file in your package directory will be included in the package, _unless it is ignored by your `.gitignore`_. If you want to include a file that is ignored by `.gitignore`, e.g. generated source code in my case, then add it to [`[tool.poetry].include`](https://python-poetry.org/docs/pyproject/#include-and-exclude).

meshy:
@thejohnfreeman that's not working for me -- if I explicitly include a file that's in `.gitignore`, I'm seeing this when running `poetry build`:

```
[IndexError]
list index out of range
```

Specifically, the sdist step seems to succeed, but the wheel step fails.

thejohnfreeman:
Do you have a link to the failing example, maybe in a branch on GitHub?

meshy:
@thejohnfreeman yes! Thanks for the quick reply.

Here's the branch: https://github.com/meshy/django-schema-graph/tree/poetry
And here's the related PR: https://github.com/meshy/django-schema-graph/pull/12

meshy:
Here's the stacktrace -- hopefully it'll help:

```pytb
Traceback (most recent call last):
  File "/.../pip/_vendor/pep517/_in_process.py", line 257, in <module>
    main()
  File "/.../pip/_vendor/pep517/_in_process.py", line 240, in main
    json_out['return_val'] = hook(**hook_input['kwargs'])
  File "/.../pip/_vendor/pep517/_in_process.py", line 110, in prepare_metadata_for_build_wheel
    return hook(metadata_directory, config_settings)
  File "/.../poetry/masonry/api.py", line 39, in prepare_metadata_for_build_wheel
    builder = WheelBuilder(poetry, SystemEnv(Path(sys.prefix)), NullIO())
[...]

PR #2268, opened 2020-04-05
Add documentation on inline tables in include. Resolves: Core #8.

stephsamson:
Resolves: #2353
Relates to: python-poetry/poetry-core#6
Resolves: #3823

- [x] Added **tests** for changed code. (see https://github.com/python-poetry/core/pull/6)

This PR is an accompaniment to the pull request on core which add inline tables to `include` in pyproject.toml.

### DIFF
--- a/docs/docs/pyproject.md
+++ b/docs/docs/pyproject.md
@@ -173,14 +173,14 @@ If a VCS is being used for a package, the exclude field will be seeded with the 
 include = ["CHANGELOG.md"]
 ```
 
-You can also use inline tables for `include`:
+You can also specify the formats for which these patterns have to be included, as shown here:
 
 ```toml
 [tool.poetry]
 # ...
 include = [
     { path = "tests", format = "sdist" },
-    { path = "for_wheel.txt", format = "wheel" }
+    { path = "for_wheel.txt", format = ["sdist", "wheel"] }
 ]
 ```
 
--- a/docs/docs/pyproject.md
+++ b/docs/docs/pyproject.md
@@ -184,6 +184,8 @@ include = [
 ]
 ```
 
+If no format is specified, it will default to include to a `wheel` only.
+
 ```toml
 exclude = ["my_package/excluded.py"]
 ```
--- a/docs/docs/pyproject.md
+++ b/docs/docs/pyproject.md
@@ -184,7 +184,7 @@ include = [
 ]
 ```
 
-If no format is specified, it will default to include to a `wheel` only.
+If no format is specified, it will default to include both `sdist` and `wheel`.
 
 ```toml
 exclude = ["my_package/excluded.py"]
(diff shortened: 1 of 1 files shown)


abn:
This cannot be merged until next core release with python-poetry/core#8 is used in poetry.

abn:
@stephsamson is the documentation on this still up-to-date? Also, we should consider rebasing this onto `develop` as `master` (1.0.x) releases will not have this feature.

docs/docs/pyproject.md
```diff
@@ -173,10 +173,25 @@ If a VCS is being used for a package, the exclude field will be seeded with the
...
+```
+
 ```toml
 exclude = ["my_package/excluded.py"]
 ```
 
+!!!note
+
+    You can either use strings or inline tables for `include` but not both.
+
```
sdispater, line 194: This can be removed since the TOML 1.0.0rc1 specification now allows mixed-types arrays.

sdispater (changes requested):

sbraz:
Hi, can this doc be fixed and merged? It looks like the key was changed from `include` to `path`, making this the correct syntax that works with poetry-core 1.0.0:
```toml
include = [
    { path = "tests", format = "sdist" },
    { path = "for_wheel.txt", format = "wheel" }
]
```

docs/docs/pyproject.md
```diff
@@ -173,6 +173,17 @@ If a VCS is being used for a package, the exclude field will be seeded with the
 include = ["CHANGELOG.md"]
 ```
 
+You can also use inline tables for `include`:
```
abn, line 176:
```suggestion
You can also specify the formats for which these patterns have to be included, as shown here.
```

Should we add a multi-format entry to the example?

abn (changes requested):

docs/docs/pyproject.md
stephsamson: yes, good idea

sinoroc:
Related: https://github.com/python-poetry/poetry/issues/2015 https://github.com/python-poetry/poetry/issues/3380

docs/docs/pyproject.md
```diff
@@ -173,6 +173,17 @@ If a VCS is being used for a package, the exclude field will be seeded with the
 include = ["CHANGELOG.md"]
 ```
 
+You can also specify the formats for which these patterns have to be included, as shown here:
+
+```toml
+[tool.poetry]
+# ...
+include = [
```
abn, line 181: Might also want to note what the default behaviour here is, ie. if `format` is left out.

sinoroc:
What is the difference between [_`packages`_](https://python-poetry.org/docs/pyproject/#packages) and [_`include`_](https://python-poetry.org/docs/pyproject/#include-and-exclude)?

See for example: https://github.com/python-poetry/poetry/issues/3380#issuecomment-734915524

abn (approved):

PR #2279, opened 2020-04-07
doc: explain why `update` sometimes causes a no-op

dmerejkowsky:
Just something my co-workers struggled with - without having authored `pyproject.toml` themselves, they went straight to the `cli.md` page - so I thought I'd duplicate a bit of information from the `basic-usage.md` there.

Maybe there's a better way to fix this, though, like explaining why `update` did not update anything or something like that.

Or maybe it should go in the FAQ.

Tell me what you think!

### DIFF
--- a/docs/docs/cli.md
+++ b/docs/docs/cli.md
@@ -165,6 +165,11 @@ If you just want to update a few packages and not all, you can list them as such
 poetry update requests toml
 ```
 
+Note that this will not update versions for dependencies outside their version constraints specified
+in the `pyproject.toml` file. In other terms, `poetry update foo` will be a no-op if the version constraint
+specified for `foo` is `~2.3` or `2.3` and `2.4` is available. In order for `foo` to be updated, you must
+update the constraint, for example `^2.3`. You can do this using the `add` command.
+
 ### Options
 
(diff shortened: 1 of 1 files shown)


docs/docs/cli.md
```diff
@@ -165,6 +165,11 @@ If you just want to update a few packages and not all, you can list them as such
 poetry update requests toml
 ```
 
+Note that for this to work, you must not be using an exact version specification in your
+`pyproject.toml` file. In other terms, `poetry update foo` will be a no-op if the `pyproject.toml`
+contains `foo = "2.3"`. In order for foo to be updated to "2.4", you should first edit the
+configuration file to have `foo =  "^2.3"`.
+
```
abn, line 172:
```suggestion
Note that this will not update versions for dependencies outside their version constraints specified
in the `pyproject.toml` file. In other terms, `poetry update foo` will be a no-op if the version constraint 
specified for `toml` is `~2.3` or `2.3` and `2.4` is available. In order for foo to be updated, you must 
update the constraint, for example `^2.3`. You can do this using the `add` command. 

```
A few minor suggestions.

dmerejkowsky:
Thank y'all

PR #2296, opened 2020-04-12
use prepare_metadata_for_build_wheel instead of parsing setup.py

finswimmer:
This PR is a complete rework, how poetry gather metadata about non-poetry project, that are necessary for dependency resolution, in case no `sdist` or `wheel` is available.

In the original implementation this was done by trying to parse the `setup.py` or `setup.cfg`. This has the strong limitation to situation where version and dependencies are not determine during runtime. And poetry has to run with python3, because the `ast` module for python2 works in a different way.

Build tools like `setuptools` that implements PEP 517 should provide a method `prepare_metadata_for_build_wheel` for getting access to the metadata one need for dependency resolution.

To gain access to this API hook following steps are needed:

* create a separate environment in which the build tool is installed
* run `prepare_metadata_for_build_wheel` for the build backend on the folder containing the package
* parse the output with `importlib.metadata`

Lot of this work is already done in the [pep517 package](https://pypi.org/project/pep517/).

This implementation has the advantage:

* metadata for complex `setup.py` can be obtained
* the implementation in not necessary limited to `setuptools`. Virtually any PEP517 compliant build system can be integrated easily
* the code is less complex

The only downside is, that the whole process needs more time than parsing. In my opinion this is negligible over the advantages.
 
# Pull Request Check List

- [*] Added **tests** for changed code.
<strike>- [ ] Updated **documentation** for changed code.</strike>


pradyunsg:
This seems to be reimplementing a lot of the functionality from `pep517`. Consider directly using pep517 for the implementation here?

finswimmer:
@pradyunsg Your comment was fast :)

First I was a bit irritated by it, what you mean by `pep517`. After reading [PEP517](https://www.python.org/dev/peps/pep-0517/) once again I've found this part:

> A Python library will be provided which frontends can use to easily call hooks this way.

And then I found this: https://pypi.org/project/pep517/ I guess this is what you meant?

Looks promising. Will have a look at this.

Thanks a lot! :+1:

pradyunsg:
We're all kinds of bad at naming things. :)

I was on mobile, so linking to the PyPI page was... possible but annoyingly difficult. Sorry!

tests/console/commands/test_init.py
```diff
@@ -144,7 +144,7 @@ def test_interactive_with_git_dependencies(app, repo, mocker, poetry):
     command = app.find("init")
     command._pool = poetry.pool
 
-    mocker.patch("poetry.utils._compat.Path.open")
+    # mocker.patch("poetry.utils._compat.Path.open")
```
abn, line 147: If you have not considered it yet, you could consider adding a test fixture to do this since this is reused. Something like this might be useful.

```py
@pytest.fixture
def mocked_pyproject_toml(mocker):
    def side_effect(self, *args, **kwargs):
        if self.name == "pyproject.toml":
            return MagicMock()
        return self.open(*args, **kwargs)

    patched_open = mocker.patch("poetry.utils._compat.Path.open")
    patched_open.side_effect = side_effect
    yield patched_open
```

tests/puzzle/test_provider.py
```diff
@@ -116,16 +116,16 @@ def test_search_for_vcs_read_setup_with_extras(provider, mocker):
     }
 
 
-def test_search_for_vcs_read_setup_raises_error_if_no_version(provider, mocker):
+def test_search_for_vcs_read_setup_with_dynamic_version(provider, mocker):
```
abn, line 119: Should we consider retaining the old case as  well since it is an expected failure?

tests/utils/fixtures/setups/pendulum/setup.py
```diff
@@ -1,38 +1,36 @@
...
+    # "pendulum.tz.zoneinfo",
+    # "pendulum.utils",
 ]
 
 package_data = {"": ["*"]}
 
 install_requires = ["python-dateutil>=2.6,<3.0", "pytzdata>=2018.3"]
 
-extras_require = {':python_version < "3.5"': ["typing>=3.6,<4.0"]}
+extras_require = {'typing:python_version < "3.5"': ["typing>=3.6,<4.0"]}
```
abn, line 33: :+1:

tests/utils/test_setup_reader.py
```diff
@@ -134,28 +131,10 @@ def test_setup_reader_read_setup_kwargs(setup):
     assert expected_python_requires == result["python_requires"]
 
 
-@pytest.mark.skipif(not PY35, reason="AST parsing does not work for Python <3.4")
-def test_setup_reader_read_setup_call_in_main(setup):
```
abn, line 138: Does this case still work as expected with pep517?

poetry/utils/setup_reader.py
```diff
@@ -1,379 +1,96 @@
...
 from ._compat import Path
-from ._compat import basestring
 
 
 try:
-    from configparser import ConfigParser
+    from importlib.metadata import PathDistribution
 except ImportError:
-    from ConfigParser import ConfigParser
+    from importlib_metadata import PathDistribution
```
abn, line 15: This might be redundant?

poetry/utils/setup_reader.py
```diff
@@ -1,379 +1,96 @@
...
-
-        return result
-
-    @classmethod
-    def _is_empty_result(cls, result):  # type: (Dict[str, Any]) -> bool
-        return (
-            not result["install_requires"]
-            and not result["extras_require"]
-            and not result["python_requires"]
+    def read_from_pep517_hook(cls, directory):
```
abn, line 27: Fwiw, I think it might be great to retain the name for now. Since we are reading form the givent directory.

poetry/utils/setup_reader.py
```diff
@@ -1,379 +1,96 @@
...
-        # Inspecting keyword arguments
-        result["name"] = self._find_single_string(setup_call, body, "name")
-        result["version"] = self._find_single_string(setup_call, body, "version")
-        result["install_requires"] = self._find_install_requires(setup_call, body)
-        result["extras_require"] = self._find_extras_require(setup_call, body)
-        result["python_requires"] = self._find_single_string(
-            setup_call, body, "python_requires"
-        )
+        with pep517.envbuild.BuildEnvironment() as env, temporary_directory() as tmp_dir:
+            env.pip_install([cls.build_requires])
```
abn, line 33: Does pep517 do something like `python -m ensurepip` when using pip commands?

poetry/utils/setup_reader.py
```diff
@@ -1,379 +1,96 @@
...
+                "version": distribution.version,
+                "summary": distribution.metadata.get("Summary"),
+                "install_requires": [],
+                "python_requires": distribution.metadata.get("Requires-Python"),
+                "extras_require": {},
+            }
+
+            if distribution.requires:
+                for record in distribution.requires:
+                    requirements = record.split(";", 1)
```
abn, line 51: can we simply use `poetry.core.packages.dependency_from_pep_508(record)` here?

poetry/utils/setup_reader.py
```diff
@@ -1,379 +1,96 @@
...
+                            result["install_requires"].append(
+                                "{project_name};{marker}".format(
+                                    project_name=project_name, marker=marker
+                                )
+                            )
+
+                    except IndexError:
+                        result["install_requires"].append(project_name)
+
+        egg_info = Path(directory).glob("*.egg-info")
```
abn, line 92: We should not need this anymore correct? Since the metadata will be generated in the isolated environment?

tests/utils/fixtures/setups/pendulum/setup.py
```diff
@@ -1,38 +1,36 @@
...
-    "pendulum.locales.pt_br",
-    "pendulum.locales.zh",
-    "pendulum.mixins",
-    "pendulum.parsing",
-    "pendulum.parsing.exceptions",
-    "pendulum.tz",
-    "pendulum.tz.data",
-    "pendulum.tz.zoneinfo",
-    "pendulum.utils",
+    # "pendulum._extensions",
```
abn, line 6: Is this because we use a truncated source in the fixture?

sdispater:
Thanks for the thorough and comprehensive work on this PR!

However, I think the purpose of the `SetupReader` is lost in this PR. The idea was to not execute the `setup.py` file to get the package metadata for security and consistency reasons (since a `setup.py` file is not necessarily deterministic).

But for projects using `setuptools` as a build backend getting the metadata will execute `setup.py` (see https://github.com/pypa/setuptools/blob/master/setuptools/build_meta.py#L158).

By going this route, we might end up losing performances (using the `BuildEnvironment` might be costly) and security for a marginal gain which is getting more accurate metadata for non deterministic `setup.py` files.

I am not too keen on executing anything just for the sake of getting metadata (even using the `prepare_metadata_for_build_wheel` does not make sense in a context of dependency resolution because it means downloading and using an external tool for something that should be static).

I know it's a tricky subject and not everyone agrees but, in my opinion, the build backend should only be used to build a wheel for installation and nothing else.

That being said I appreciate the work you put into it, don't get me wrong, and if someone can guarantee me that nothing will ever be executed by going this PR route, I am willing to do it. But for now I would prefer not to go this way, even if it follows the current packaging guidelines.

And, again, this is great work and I appreciate the time put into it.

And I am open for discussion :-)

abn:
@sdispater the ast parsing solution can still be retained. In our current implementation, as I understand it, we do execute the `setup.py` file before the AST parsing anyway using `python setup.py egg_info`. From what I can tell, using pep517, we get a more rounded solution here that could potentially handle more cases than we already do.

sdispater:
@abn This is true only for git and directory dependencies (since in this case we assume people know what they are doing). So if we remove the change from the `Inspector` class which is used by the `LegacyRepository` and `PyPIRepository` classes then I think it would be more acceptable.

The other thing that bothers me (even though I must admit it's already the case) is that it will build a new environment each time. A better solution would be to create a unique environment lazily that will be reused during the dependency resolution and cleaned up after.

I know the AST parser is flaky (I wrote most of it 😅) but it works reasonably well so we should try to rely on it as much as possible.

abn:
Ha, I agree that the AST parser is a decent fallback - we can obviously improve it as we go along. Considering we can cache this information for packages and that this would be the exception cases, I do not see the perf-hit to be considerable or unmanagable spinning up an isolated environment (we can always improve this later). I guess there are different views on that one. 

Fwiw, if we can get better metadata, I feel that this or a similar solution might be worth it. It will also mean we can support non poetry PEP-517 source packages too (the ones without the metadata). Also, this discussion is relevant for #2301 as well.

finswimmer:
Hello @sdispater ,

I can understand your reservation about executing something to get the necessary and I would also prefer something static. That's why I started the discussion [here](https://discuss.python.org/t/standardized-way-for-receiving-dependencies/3821)

The current result of the discussion is, that using `prepare_metadata_for_build_wheel` is the intended way to get the data we need for a dependency resolution. I would be happy if we can go on in the discussion and can convince more people, that this is not the optimal way.

The biggest plus for my implementation I see in the fact that we can support other packaging tools aside poetry and `setuptools`.

Maybe we can use `prepare_metadata_for_build_wheel` as a fallback for now, if ast parsing isn't enough?

@abn: Thanks a lot for your code comments. I will go through them later.

pradyunsg:
I agree that using the AST parser, when the backend is setuptools, is a good idea. Maybe the setuptools backend should probably be doing these AST shenanigans instead, but that's not something under poetry's control, so it's definitely a good idea to use the AST parser for situations where it's pretty unambigous what the results would be.

For non-setuptools backends though, I do think it'd be good to have PEP 517 support in poetry. :)

tests/console/commands/test_init.py
abn: This was already added for another change at https://github.com/python-poetry/poetry/blob/95e3490a76c876f2e5487499fa592cd434ddc901/tests/conftest.py#L122-L134

Could reuse this.

sdispater:
So, I agree with @pradyunsg: If the backend is `setuptools` we use the AST parser (unless all the information we need is in the `setup.cfg` file). Note that I am pretty sure we can further improve the parser.

For other backends, we use the appropriate PEP-517 hooks. I don't think we can use the `pep517` library however. It uses in a few places `sys.executable` directly meaning we can't use it with the environment system we have in place in Poetry.

Issue #22, opened 2018-03-19
Does this replace poet?


ulope:
While trying to remember the exact name of this project and googling I first found your poet package which was a bit confusing.
Could you clarify what the (if any) relationship between the two is?

Also: Thanks a ton for bringing some fresh air into the python packaging world!

sdispater:
Thanks for your interest in Poetry and your kind words!

Yes, Poetry replaces poet. This is due to the fact that I wanted to start over with a cleaner codebase.

I will add a message and the poet's README to acknowledge that it is now obsolete.

thebigmunch:
You could also use GitHub's repository deprecation feature on the poet repository and change its description to point to poetry.

sdispater:
The poet project has been archived and I added a line saying that it's been deprecated in favor of poetry.

PR #2301, opened 2020-04-13
Refactor: unify package inspection logic

abn:
This change brings together logic repeated in various places to discover package information. As a side-effect of this various edge cases have been improved and we shouldhave one place to change the code for package inspection related logic.

In addition to the above, this alos cleans up some test code and also unify call of urlparse and download.

Note this requires python-poetry/core#22. In order for the builds to work, I have created a branch for core that works around a few non issues with the current poetry releases used in ci. Additionally, the FreeBSD tests are failing because of a symbolic link issue (unrelated),

# Pull Request Check List
- ~~Updated **documentation** for changed code.~~


abn:
Builds are failing due to https://github.com/python-poetry/poetry/pull/2302

pyproject.toml
```diff
@@ -23,7 +23,7 @@ classifiers = [
 # Requirements
 [tool.poetry.dependencies]
 python = "~2.7 || ^3.5"
-poetry-core = "^1.0.0a5"
+poetry-core = { git = "https://github.com/abn/core.git", branch = "file-uri-scheme-sdist" }
```
stephsamson, line 26: I assume this will be changed once [#22](https://github.com/python-poetry/core/pull/22) is resolved?

abn: Yes.

abn:
@sdispater i have ensured that we only execute `python setup.py egg_info` for git/path dependencies as it was before.

poetry/inspection/info.py
```diff
@@ -0,0 +1,499 @@
...
+from poetry.utils.env import VirtualEnv
+from poetry.utils.setup_reader import SetupReader
+from poetry.utils.toml_file import TomlFile
+
+
+logger = logging.getLogger(__name__)
+
+
+class PackageInfoError(ValueError):
+    def __init__(self, path):
```
stephsamson, line 35: What about type hinting for `path` here for consistency?

PR #2302, opened 2020-04-13
Fix incorrect sdist builder call

abn:


poetry/installation/pip_installer.py
```diff
@@ -210,9 +208,7 @@ def install_directory(self, package):
             # file since pip, as of this comment, does not support
             # build-system for editable packages
             # We also need it for non-PEP-517 packages
-            builder = SdistBuilder(
-                Factory().create_poetry(pyproject.parent), NullEnv(), NullIO()
-            )
+            builder = SdistBuilder(Factory().create_poetry(pyproject.parent),)
```
stephsamson, line 211: What about explicitly specifying the `ignore_packages_formats` parameter too?

abn: Don't think we need that here. We are okay with the defaults here I think.

stephsamson: :+1: what about removing the `ignore_packages_formats` parameter? It is [only used once](https://github.com/python-poetry/core/search?q=ignore_packages_formats&unscoped_q=ignore_packages_formats), in the [`CompleteBuilder`](https://github.com/python-poetry/core/blob/8044be832da14080b8717a2ce15c513be144ea5f/poetry/core/masonry/builders/complete.py#L35).

abn: Builder is in core. I think that will have to be a separate change.

PR #2303, opened 2020-04-13
Use `requirements.txt` as default export format

yggi49:
Prior to this change, users were always forced to specify the `--format`
option when invoking `poetry export`, although `requirements.txt` is
currently the only accepted value.

This patch modifies `poetry export` to make the `--format` option
nonobligatory.  If omitted, `requirements.txt` will be assumed as the
default value.

*(This is a follow-up to [my comment for PR #1035](https://github.com/python-poetry/poetry/pull/1035#issuecomment-483346513).)*

# Pull Request Check List

### DIFF
--- a/poetry/console/commands/export.py
+++ b/poetry/console/commands/export.py
@@ -11,7 +11,13 @@ class ExportCommand(Command):
     description = "Exports the lock file to alternative formats."
 
     options = [
-        option("format", "f", "Format to export to.", flag=False),
+        option(
+            "format",
+            "f",
+            "Format to export to.",
+            flag=False,
+            default=Exporter.ACCEPTED_FORMATS[0],
+        ),
         option("output", "o", "The name of the output file.", flag=False),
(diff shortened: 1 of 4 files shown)


poetry/console/commands/export.py
```diff
@@ -11,7 +11,13 @@ class ExportCommand(Command):
     description = "Exports the lock file to alternative formats."
 
     options = [
-        option("format", "f", "Format to export to.", flag=False),
+        option(
+            "format",
+            "f",
+            "Format to export to.",
+            flag=False,
+            default=Exporter.ACCEPTED_FORMATS[0],
```
abn, line 19: It would be good if we are less dynamic about this since we are documenting `requirements.txt` explicitly as the default. Might be worthwhile to split the format into constants for now.

yggi49: I was actually torn whether this was worth adding a constant or not.  Thanks for the second opinion; I just updated the MR accordingly.

PR #2308, opened 2020-04-14
Sort extras constraints of dependencies in lockfile

osantana:
The extras constraints ordering is not guaranteed on metadata and it causes changes in poetry.lock file during the update process even when the dependency did not change.

# Pull Request Check List


osantana:
Thanks!

Issue #2309, opened 2020-04-14
Automate freebsd release artifact from cirrus-ci


abn:
With #2297 FreeBSD bdists will be built. As mentioned in the PR, we should add configuration for [automated upload of release artifacts](https://cirrus-ci.org/examples/#release-assets),

Relates-to: #1647 #2159

PR #2311, opened 2020-04-14
ci: artifact release automation for freebsd

abn:
Resolves: #2309

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -6,10 +6,7 @@ test_task:
   env:
     matrix:
       - PYTHON: python2.7
-      - PYTHON: python3.5
-      - PYTHON: python3.6
       - PYTHON: python3.7
-      - PYTHON: python3.8
   python_script:
     - PYPACKAGE=$(printf '%s' $PYTHON | tr -d '.')
     - SQLPACKAGE=$(printf '%s-sqlite3' $PYPACKAGE | sed 's/thon//')
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -29,11 +29,13 @@ test_task:
 release_task:
   name: "Release / FreeBSD"
   only_if: $CIRRUS_TAG != ''
+  env:
+    GITHUB_TOKEN: ENCRYPTED[e637a1887c028ea4e44867715a8a4b7c5e27568559a15c45a3e33739b0d97fc050394c728a44e9bfe7c246179810cad7]
   freebsd_instance:
     matrix:
       - image_family: freebsd-12-1-snap
       - image_family: freebsd-11-3-snap
-  python_script: pkg install -y python3 python27 python35 python36 python37 python38
+  python_script: pkg install -y curl python3 python27 python35 python36 python37 python38
   pip_script:
     - python2.7 -m ensurepip
     - python3.5 -m ensurepip
@@ -41,5 +43,17 @@ release_task:
     - python3.7 -m ensurepip
     - python3.8 -m ensurepip
   build_script: ./make-nix-release.sh
+  upload_script: |
+    for fpath in releases/*
+    do
+      echo "Uploading $fpath..."
+      name=$(basename "$fpath")
+      url_to_upload="https://uploads.github.com/repos/$CIRRUS_REPO_FULL_NAME/releases/$CIRRUS_TAG/assets?name=$name"
+      curl -X POST \
+        --data-binary @$fpath \
+        --header "Authorization: token $GITHUB_TOKEN" \
+        --header "Content-Type: application/octet-stream" \
+        $url_to_upload
+    done
   archive_artifacts:
     path: "releases/*"
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -4,6 +4,7 @@ on:
   push:
     paths-ignore:
       - 'docs/**'
+      - '.cirrus.yml'
     branches:
       - master
       - develop


abn:
@sdispater can you please add an encrupted token in the file? Just replace `qwerty` with the ecrypted token as described at https://cirrus-ci.org/examples/#release-assets.

.cirrus.yml
```diff
@@ -42,5 +48,32 @@ release_task:
     - python3.7 -m ensurepip
     - python3.8 -m ensurepip
   build_script: ./make-nix-release.sh
+  script: |
+    #!/usr/bin/env bash
+
+    if [[ "$CIRRUS_RELEASE" == "" ]]; then
```
neersighted, line 54: I've discovered that jobs won't fire with `$CIRRUS_RELEASE` unless the release is tagged from the web UI. I'd suggest just eliding this check since we check for `$CIRRUS_TAG` above.

.cirrus.yml
```diff
@@ -42,5 +48,32 @@ release_task:
     - python3.7 -m ensurepip
     - python3.8 -m ensurepip
   build_script: ./make-nix-release.sh
+  script: |
+    #!/usr/bin/env bash
```
neersighted, line 52: The shebang should not be necessary. Also, I'm pretty sure a prefix is required here, along the lines of  `upload_script:`

.cirrus.yml
```diff
@@ -42,5 +48,32 @@ release_task:
...
+
+    file_content_type="application/octet-stream"
+    files_to_upload=(releases/*)
+
+    for fpath in $files_to_upload
+    do
+      echo "Uploading $fpath..."
+      name=$(basename "$fpath")
+      url_to_upload="https://uploads.github.com/repos/$CIRRUS_REPO_FULL_NAME/releases/$CIRRUS_RELEASE/assets?name=$name"
+      curl -X POST \
```
neersighted, line 72: You may need to `pkg install curl` -- not sure if Cirrus/GCP include curl by default in their images.

.cirrus.yml
```diff
@@ -42,5 +48,32 @@ release_task:
...
+      echo "Not a release. No need to deploy!"
+      exit 0
+    fi
+
+    if [[ "$GITHUB_TOKEN" == "" ]]; then
+      echo "Please provide GitHub access token via GITHUB_TOKEN environment variable!"
+      exit 1
+    fi
+
+    file_content_type="application/octet-stream"
```
neersighted, line 64: I'd probably just hard-code this in the `curl` invocation.

.cirrus.yml
```diff
@@ -42,5 +48,32 @@ release_task:
...
+    files_to_upload=(releases/*)
```
neersighted, line 65: This won't work on FreeBSD -- unless they're shipping `bash` by default in the image, which would be odd. Assuming only POSIX `sh` is available, I would just change line 67 to `for fpath in releases*`.

.cirrus.yml
```diff
@@ -42,5 +48,32 @@ release_task:
...
+
+    for fpath in $files_to_upload
+    do
+      echo "Uploading $fpath..."
+      name=$(basename "$fpath")
+      url_to_upload="https://uploads.github.com/repos/$CIRRUS_REPO_FULL_NAME/releases/$CIRRUS_RELEASE/assets?name=$name"
```
neersighted, line 71: I think this will have to be `$CIRRUS_TAG` instead of `$CIRRUS_RELEASE`, if my comment above holds.

.cirrus.yml
```diff
@@ -19,7 +16,14 @@ test_task:
...
     - $PYTHON -m pip install poetry -U --pre
     - $PYTHON -m poetry config virtualenvs.in-project true
-  deps_script: $PYTHON -m poetry install
+    - $PYTHON -m poetry config --list
+  # todo: cache handling need to be fixed
+  venv_cache:
+    folder: ".venv"
+    fingerprint_script: echo freebsd-$($PYTHON -c "import sys; print('-'.join(str(v) for v in sys.version_info))")-$(sha512 poetry.lock)
+  deps_script: |
+    $PYTHON -m poetry run pip --version >/dev/null 2>&1 || rm -vrf .venv
```
neersighted, line 25: I experimented with doing this exactly, and found that `pip` would run but not the test suite. Here be dragons, if you want to go through with this.

abn: I am just using it for now to try figure out whats going on. Did you find much else out?

neersighted: Not really -- `pip --version` passed but the test suite broke pretty spectacularly. On Python 3, Python can't find `pytest` in the venv -- and on Python 2.7, it can but something blows up when it tests symlink support? Maybe your deeper knowledge of Poetry/the test suite will help here.

https://cirrus-ci.com/task/5953586152603648
https://cirrus-ci.com/task/6462722447507456
https://cirrus-ci.com/task/4773872587243520

neersighted:
Relevant to the release issue: cirruslabs/cirrus-ci-docs#452

.cirrus.yml
abn: Really appreciate your freedback here @neersighted. I am not quite familiar with FreeBSD. 

From what I can tell, the issue probably has to do with how the virtual environment creation is done on FreeBSD systems either due to the system python package itself or how cirrus-ci sets up the build environment. 

The primary difference, outside of platform specifics, in py3.7 and py2.7 virutal environment creation is the use of the built-in `venv` module and external `virtualenv` package respectively. I am aware that the newer versions of behave differently to the built-in module in that on systems like FreeBSD, the lib64 directory is not a symlink to the lib directory. 

In the case of py3.7, a cache hit as no effect (ie. packages are downloaded again) and in the case of py2.7 a cache hit causes the test suite to fail as mentioned above.

Might consider caching the pip and poetry caches instead of the virtual environment for now so that atleast downloads are improved until the venv can be made portable.

One theory, if it is indeed an issue with symlink, is that the cirrus might be using an archiving option that does not like symlinks.

neersighted: I see -- thanks for the insight. I agree that we could keep the pip package cash -- it should speed things up with no downside. On the other hand, most of the time seems to be in disk I/O as these GCP VMs don't seem to exactly be IOPS monsters... The savings may be as little as 20 seconds so I'm not sure it's worth the extra complexity.

abn: Created #2315 to handle caching. Thanks for your help @neersighted.

.cirrus.yml
```diff
@@ -19,7 +16,9 @@ test_task:
     - $PYTHON -m pip install pip -U
     - $PYTHON -m pip install poetry -U --pre
     - $PYTHON -m poetry config virtualenvs.in-project true
-  deps_script: $PYTHON -m poetry install
+    - $PYTHON -m poetry config --list
+  deps_script: |
```
neersighted, line 20: Just nitpicking here, but might as well change this back.

neersighted:
@sdispater With 1.10 coming up, I'd really like to see this merged. Anything else it needs?

abn:
This should be now ready for merge.

Issue #1884, opened 2020-01-14
CLI command to remove all old environments


zachvalenta:
- [x] I have searched the [issues](https://github.com/python-poetry/poetry/issues) of this repo and believe that this is not a duplicate. -> https://github.com/python-poetry/poetry/issues?utf8=%E2%9C%93&q=is%3Aissue+is%3Aopen+removing+old+environments
- [x] I have searched the [documentation](https://python-poetry.org/docs/) and believe that my question is not covered. --> https://python-poetry.org/docs/managing-environments/#deleting-the-environments

## Feature Request

__tldr__: Is there a way to way to clean up old environments en masse?

__context__: I often spin up a quick project to explore an idea and then tear it down once I'm past the exploration phase. Using `pip`, the environment could be co-located to the project i.e. `rm <my_project>` would also remove the environment. With Poetry, environments are located outside the project, and thus it's possible to delete the project without deleting the environment. As a result, you could have environments at `path/to/pypoetry/virtualenvs/` whose projects have been deleted. Currently, Poetry only offers the ability to remove an environment using the CLI from within the project itself; if you've already deleted the project, you need to navigate to `path/to/pypoetry/virtualenvs/` and manually remove its hanging environment. It'd be nice if Poetry had a CLI command in which you could remove all environments that no longer correspond to source code directories on the operating system.

finswimmer:
Hello,

you can [config](https://python-poetry.org/docs/configuration/#virtualenvsin-project-boolean) poetry to put the venv within the project using `virtualenvs.in-project 
 true`.

I don't think it is possible to remove venv where the project does not exists anymore via CLI, because there is no information about where the location of the project was.

Having a command to cleanup all venv for an existing project should be possible and seems quite useful to me.

fin swimmer

jtrakk:
There could be an option


```
poetry env remove --all
```
or it could be

```
poetry env remove one python3.6
poetry env remove all
```

abn:
Resolved-by: #3212

am1006:
> Resolved-by: #3212

Not sure if this was answered before. After trial and error, I don't quite get how this merge could fix the problem that OP tried to fix. From #3212 I can see that the `-all` argument can remove all environments that are attached to the current project. Could you advise me how this will remove all hanging environments (i.e., the problem that OP raised)?

DominikRoB:
Found this here, because I have the same issue as OP.
[...]

PR #2341, opened 2020-04-24
added --all option to remove

rschwarzkopf:
# Pull Request Check List

Resolves: #1884 (not really, but it [implements a proposal](https://github.com/python-poetry/poetry/issues/1884#issuecomment-574506572) made by @finswimmer)

 

I tried to base this PR on `develop`, but the tests do not run at all in that branch, so I based it on `master` although it introduces a new feature.

 
I added an `--all` option to `env remove` to remove all environments at once. The idea to use an option is [also taken from the issue](https://github.com/python-poetry/poetry/issues/1884#issuecomment-575932990). Of the two variants, I prefer the option because it does not break the existing cli.

On the other hand, this approach requires a little workaround, because the `python` argument to `env remove` needs to be optional for `--all` to work, but without the `--all` option it is a required argument. I'm manually throwing an exception if neither `--all` nor an argument are given - in fact I throw the exception CliKit would have thrown in that case. I added an additional test just for this case.

Maybe it would be best to add a complete new subcommand for this. I just had no good idea how to name it. Maybe `remove-all`, but I wasn't sure if you like dashes in commands. So if anyone has a good idea I happily adopt my PR.


neersighted:
#3212 was landed which implements this. Thanks for your PR!

Issue #2329, opened 2020-04-20
RecursionError when installing MoviePy


roniemartinez:
- [x] I am on the [latest](https://github.com/python-poetry/poetry/releases/latest) Poetry version.
- [x] I have searched the [issues](https://github.com/python-poetry/poetry/issues) of this repo and believe that this is not a duplicate.
- [x] If an exception occurs when executing a command, I executed it again in debug mode (`-vvv` option).

- **OS version and name**: Mac OS X 10.15.4
- **Poetry version**: 1.0.5

## Issue
```console
% poetry add moviepy -E optional    
Using version ^1.0.2 for moviepy

Updating dependencies
Resolving dependencies... (17.9s)

[RecursionError]
maximum recursion depth exceeded
```

isuckatprogram:
Is there a fix to this?

abn:
Fixed in `1.1.0b2`.

Issue #2133, opened 2020-03-05
[RecursionError] maximum recursion depth exceeded while calling a Python object with Spyder-kernels


danieltomasz:
- [x] I am on the [latest](https://github.com/python-poetry/poetry/releases/latest) Poetry version.
- [x] I have searched the [issues](https://github.com/python-poetry/poetry/issues) of this repo and believe that this is not a duplicate.
- [x] If an exception occurs when executing a command, I executed it again in debug mode (`` option).

- **OS version and name**: Mac Catalina darwin
- **Poetry version**: 1.0.5 
- **Link of a [Gist](https://gist.github.com/) with the contents of your pyproject.toml file**: 

## Issue
```
poetry add -D spyder-kernels
Using version ^1.8.1 for spyder-kernels

Updating dependencies
Resolving dependencies... (41.1s)

[RecursionError]
maximum recursion depth exceeded while calling a Python object
```

I was able to pinpoint problem to this
```
Duplicate dependencies for ipykernel
   0: Different requirements found for ipykernel (*) and ipykernel (>=5.1.3).
   1: Version solving took 0.029 seconds.
   1: Tried 1 solutions.
   0: Retrying dependency resolution for Python (>=3.8,<4.0).
```

which was caused by `spyder-kernels`requirements
```

REQUIREMENTS = [
  ...
    'ipykernel',
    'ipykernel>=5.1.3; python_version>"2"',
  ...
]
```

after  removing  first `ipykernel` I was able to install spyder-kernels from my fork

abn:
Fixed in `1.1.0b2`.

Issue #2008, opened 2020-02-10
RecursionError


dulm:
- [x ] I am on the [latest](https://github.com/python-poetry/poetry/releases/latest) Poetry version.
- [x ] I have searched the [issues](https://github.com/python-poetry/poetry/issues) of this repo and believe that this is not a duplicate.
https://github.com/python-poetry/poetry/issues/1238
https://github.com/python-poetry/poetry/issues/1885
- [ x] If an exception occurs when executing a command, I executed it again in debug mode (`-vvv` option).

- **OS version and name**: mint 19.3
- **Poetry version**:  1.0.3
- **Link of a [Gist](https://gist.github.com/) with the contents of your pyproject.toml file**: 
```
[tool.poetry]
name = "test1"
version = "0.1.0"
description = ""
authors = ["lqk"]

[tool.poetry.dependencies]
python = "^3.7"



[tool.poetry.dev-dependencies]


#[[tool.poetry.source]]
#name = "ali"
#url = "https://mirrors.aliyun.com/pypi/simple"
#default=true



[build-system]
requires = ["poetry>=0.12"]
build-backend = "poetry.masonry.api"
```

## Issue
1. poetry add pyautogui
1. del the   pyautogui    in  pyproject.toml
1. poetry add pyautogui

throws
```
(.venv) (base) lqk@shenzhou-mint:~/projects/PycharmProjects/Test1$ poetry add pyautogui -vvv

[RuntimeError]
[tool.poetry] section not found in pyproject.toml

Traceback (most recent call last):
  File "/home/lqk/opt/pipx/venvs/poetry/lib/python3.7/site-packages/clikit/console_application.py", line 131, in run
    status_code = command.handle(parsed_args, io)
  File "/home/lqk/opt/pipx/venvs/poetry/lib/python3.7/site-packages/clikit/api/command/command.py", line 120, in handle
    status_code = self._do_handle(args, io)
  File "/home/lqk/opt/pipx/venvs/poetry/lib/python3.7/site-packages/clikit/api/command/command.py", line 163, in _do_handle
    self._dispatcher.dispatch(PRE_HANDLE, event)
  File "/home/lqk/opt/pipx/venvs/poetry/lib/python3.7/site-packages/clikit/api/event/event_dispatcher.py", line 22, in dispatch
    self._do_dispatch(listeners, event_name, event)
[...]

Issue #1885, opened 2020-01-14
Conditional dependencies results in exceeeding maximum recursion depth (for manimlib with pycairo)


JulianStier:
- [x] I am on the [latest](https://github.com/python-poetry/poetry/releases/latest) Poetry version.
- [x] I have searched the [issues](https://github.com/python-poetry/poetry/issues) of this repo and believe that this is not a duplicate. It might be related to
  - https://github.com/python-poetry/poetry/issues/1222
  - https://github.com/python-poetry/poetry/issues/1238
- [x] If an exception occurs when executing a command, I executed it again in debug mode (`-vvv` option).

- **OS version and name**: Ubuntu 18.04.3 LTS // Linux 4.15.0-74-generic #84-Ubuntu SMP Thu Dec 19 08:06:28 UTC 2019 x86_64 x86_64 x86_64 GNU/Linux
- **Poetry version**: 1.0.2
- **Link of a [Gist](https://gist.github.com/) with the contents of your pyproject.toml file**: https://gist.github.com/JulianStier/e8b041396b26bdc3b542530404e28277

## Issue
Trying to install manimlib:
``poetry add manimlib``

Error:
> [RecursionError]
maximum recursion depth exceeded while calling a Python object

Executed with *-vv*:
>    0: Duplicate dependencies for pycairo
   0: Different requirements found for pycairo (1.17.1) and pycairo (>=1.18.0).

A look into *https://github.com/3b1b/manim/blob/master/requirements.txt*:
> pycairo==1.17.1; sys_platform == 'linux'
pycairo>=1.18.1; sys_platform == 'win32'

Apparently poetry does not recognize conditionally defined requirements based on "sys_platform".

lvkswrks:
Probably related / same as #992

abn:
Fixed in `1.1.0b2`. Requires regeneration of lock file. Note that new lock file is not compatible with older versions of `poetry`.

PR #2342, opened 2020-04-24
Fix handling of markers when resolving

sdispater:
# Pull Request Check List

Resolves: #1238 
Resolves: #1885
Resolves: #2008 
Resolves: #2133
Resolves: #2329

This one was long overdue. The way we handled environment markers when resolving was sub optimal to say the least.

Here is what this PR actually does:

If "duplicated" dependencies are found with different environment markers, Poetry will relaunch the dependency resolution with specific dependency overrides to build separate dependency graphs. For instance: `slash==1.10.0` has the following dependencies:

```
ipython
ipython (1.2.4) ; implementation_name == "pypy"
```

So, this will trigger two separate dependency resolutions:

- One with the following overrides: `{<Package slash-1.10.0>: {'ipython': <Dependency IPython (*)>}}`
- One with the following override: `{<Package slash-1.10.0>: {'ipython': <Dependency IPython (1.2.1)>}}`

A side effect of that is we will now have a more accurate lock file with less multiple version of a same package.

The only drawback to this change is speed: if there are multiple cases like this in the dependency graph, Poetry will need to start anew with new overrides each time which can drastically impact performances. That's why the use of multiple constraints dependencies should be used as a last resort because it makes the dependency resolution that much harder.

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -1,6 +1,7 @@
 [[package]]
 category = "main"
 description = "A small Python module for determining appropriate platform-specific dirs, e.g. a \"user data dir\"."
+marker = "python_version >= \"2.7\" and python_version < \"2.8\" or python_version >= \"3.6\" and python_version < \"4.0\""
 name = "appdirs"
 optional = false
 python-versions = "*"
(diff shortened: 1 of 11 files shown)


lvkswrks:
Probably resolves: #992 as well

Issue #2326, opened 2020-04-20
[InvalidRequirement] Invalid requirement, parse error at "'extra =='"


simonepri:
- [x] I am on the [latest](https://github.com/python-poetry/poetry/releases/latest) Poetry version.
- [x] I have searched the [issues](https://github.com/python-poetry/poetry/issues) of this repo and believe that this is not a duplicate.
- [x] If an exception occurs when executing a command, I executed it again in debug mode (`-vvv` option).

- **OS version and name**: maxOS 10.15
- **Poetry version**: 1.0.5

## Issue

When I run:

```bash
poetry add git+https://github.com/huggingface/transformers.git#a21d4fa410dc3b4c62f93aa0e6bbe4b75a101ee9 -vvv
```

The following exception is thrown:

```bash
[InvalidRequirement]
Invalid requirement, parse error at "'extra =='"

Traceback (most recent call last):
  File "/usr/local/Cellar/poetry/1.0.5_1/libexec/vendor/lib/python3.7/site-packages/clikit/console_application.py", line 131, in run
    status_code = command.handle(parsed_args, io)
  File "/usr/local/Cellar/poetry/1.0.5_1/libexec/vendor/lib/python3.7/site-packages/clikit/api/command/command.py", line 120, in handle
    status_code = self._do_handle(args, io)
  File "/usr/local/Cellar/poetry/1.0.5_1/libexec/vendor/lib/python3.7/site-packages/clikit/api/command/command.py", line 171, in _do_handle
    return getattr(handler, handler_method)(args, io, self)
  File "/usr/local/Cellar/poetry/1.0.5_1/libexec/vendor/lib/python3.7/site-packages/cleo/commands/command.py", line 92, in wrap_handle
    return self.handle()
  File "/usr/local/Cellar/poetry/1.0.5_1/libexec/lib/python3.7/site-packages/poetry/console/commands/add.py", line 89, in handle
    packages, allow_prereleases=self.option('allow-prereleases')
  File "/usr/local/Cellar/poetry/1.0.5_1/libexec/lib/python3.7/site-packages/poetry/console/commands/init.py", line 293, in _determine_requirements
    requires = self._parse_requirements(requires)
  File "/usr/local/Cellar/poetry/1.0.5_1/libexec/lib/python3.7/site-packages/poetry/console/commands/init.py", line 381, in _parse_requirements
    "git", url.url, reference=pair.get("rev")
[...]

PR #2344, opened 2020-04-24
Fix requirement string parsing for URLs

simonepri:
Resolves: #2326

If we want support url containing the `;` character, we can use the following regex instead:

`((?!; )(?!;^)(?! ).)+`

### DIFF
--- a/poetry/version/requirements.py
+++ b/poetry/version/requirements.py
@@ -161,7 +161,7 @@ IDENTIFIER = Combine(ALPHANUM + ZeroOrMore(IDENTIFIER_END))
 NAME = IDENTIFIER("name")
 EXTRA = IDENTIFIER
 
-URI = Regex(r"[^ ]+")("url")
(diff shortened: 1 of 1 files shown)


abn:
@simonepri looks like this also will affect `packaging`.

https://github.com/pypa/packaging/blob/61672bf9f507f38e84ce2786a1c42f55fa0a3153/packaging/requirements.py#L45-L46

Also, would be great to get this also submitted to python-poetry/core, since this will need to be fixed for `develop` too.

simonepri:
@abn Which regex shall we use as a fix?

abn:
@simonepri see https://github.com/python-poetry/poetry/issues/2326#issuecomment-619382136 I think we can fix this in the utils.

simonepri:
Closing in favour of https://github.com/python-poetry/core/pull/30 and https://github.com/python-poetry/poetry/pull/2349

Issue #2228, opened 2020-03-26
Timer appears dark on dark terminal


jtrakk:
With a dark terminal background, the timer is hard to see.

![image](https://user-images.githubusercontent.com/43392409/77696726-1011b080-6f6b-11ea-8185-6992b68704ac.png)

bhaveshpraveen:
I'd like to work on this issue if no one's picked it up yet.

abn:
@bhaveshpraveen feel free to, but @sdispater might have some thoughts here. I'd suggest creating a draft PR if you have a solution to propose.

sdispater:
This issue was caused by the fact that we were using a black color in bold which could be too dark in some terminals and this has already been fixed in this PR: #2230.

This was already reported in https://github.com/python-poetry/poetry/pull/2150.

So I am closing this since it's already been handled.

PR #2350, opened 2020-04-26
Fix Timer appears dark on dark terminal 

bhaveshpraveen:
Resolves: #2228

Original Issue: https://github.com/python-poetry/poetry/issues/2228

This removes the hard-coded black color. 

**Before**
image

**After** 
image

Since the hardcoded color is removed the timer just takes the color depending on the terminal theme.

### DIFF
--- a/poetry/puzzle/provider.py
+++ b/poetry/puzzle/provider.py
@@ -803,7 +803,7 @@ class Provider:
             self._io.write_line("Resolving dependencies...")
             yield
         else:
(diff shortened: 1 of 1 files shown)


sdispater:
Thanks for your contribution!

However, the timer was always supposed to be dimmed compared to the rest of the information.

Using a black color in bold was not the right choice because it could be too dark in some terminals and this has already been fixed in this PR: #2230.

Thanks again for your contribution but closing this since it's now considered fixed.

Issue #1906, opened 2020-01-17
Display why package was installed


shenek:
- [X] I have searched the [issues](https://github.com/python-poetry/poetry/issues) of this repo and believe that this is not a duplicate.
- [X] I have searched the [documentation](https://python-poetry.org/docs/) and believe that my question is not covered.

## Feature Request
As a poetry user I would like to find out why is this dependency used (by which packages it is required).

JS's dependency manager yarn contains a `why` subcommand which does basically what I mentioned above.
see https://yarnpkg.com/lang/en/docs/cli/why/

This feature would be quite helpful in poetry as well. Because it could be quite uncomfortable to figure out why the package was installed using `poetry show --tree`.

Or perhaps adding a filter to `poetry show` would be useful.
```
$ poetry show --tree
black 19.10b0 The uncompromising code formatter.
├── appdirs *
├── attrs >=18.1.0
├── click >=6.5
├── pathspec >=0.6,<1
├── regex *
├── toml >=0.9.4
└── typed-ast >=1.4.0
cachecontrol 0.12.6 httplib2 caching for requests
├── lockfile >=0.9
├── msgpack >=0.5.2
└── requests *
    ├── certifi >=2017.4.17 
    ├── chardet >=3.0.2,<3.1.0 
    ├── idna >=2.5,<2.9 
    └── urllib3 >=1.21.1,<1.25 
cachy 0.3.0 Cachy provides a simple yet effective caching library.
cleo 0.7.6 Cleo allows you to create beautiful and testable command-line interfaces.
└── clikit >=0.4.0,<0.5.0
    ├── enum34 >=1.1,<2.0 
    ├── pastel >=0.1.0,<0.2.0 
    ├── pylev >=1.3,<2.0 
    └── typing >=3.6,<4.0 
clikit 0.4.1 CliKit is a group of utilities to build beautiful and testable command line interfaces.
├── enum34 >=1.1,<2.0
├── pastel >=0.1.0,<0.2.0
├── pylev >=1.3,<2.0
└── typing >=3.6,<4.0
functools32 3.2.3-2 Backport of the functools module from Python 3.2.3 for use on 2.7 and PyPy.
glob2 0.6 Version of the glob module that can capture patterns and supports recursive wildcards
html5lib 1.0.1 HTML parser based on the WHATWG HTML specification
├── six >=1.9
└── webencodings *
httpretty 0.9.7 HTTP client mock for Python
└── six *
[...]

PR #2086, opened 2020-02-25
Show why package is required

shenek:
# Pull Request Check List

- [X] Added **tests** for changed code.
- [X] Updated **documentation** for changed code.

Add a flag `--why` to `poetry show`. User should see a filtered output why the package is required.
Example (run on poetry repo):
```
$ poetry show --tree --why pluggy
pluggy 0.13.1 plugin and hook calling mechanisms for python
pytest 4.6.9 pytest: simple powerful testing with Python
└── pluggy >=0.12,<1.0
tox 3.12.1 tox is a generic virtualenv management and test command line tool
└── pluggy >=0.3.0,<1
tox 3.14.3 tox is a generic virtualenv management and test command line tool
└── pluggy >=0.12.0,<1
pytest-cov 2.8.1 Pytest plugin for measuring coverage.
└── pytest >=3.6
    └── pluggy >=0.12,<1.0 
pytest-mock 1.13.0 Thin-wrapper around the mock package for easier use with py.test
└── pytest >=2.7
    └── pluggy >=0.12,<1.0 
pytest-sugar 0.9.2 pytest-sugar is a plugin for pytest that changes the default look and feel of pytest (e.g. progressbar, show tests that fail instantly).
└── pytest >=2.9
    └── pluggy >=0.12,<1.0 
```
or
```
$ poetry show --why pluggy
pluggy       0.13.1 plugin and hook calling mechanisms for python
pytest       4.6.9  pytest: simple powerful testing with Python
tox          3.12.1 tox is a generic virtualenv management and test command line tool
tox          3.14.3 tox is a generic virtualenv management and test command line tool
pytest-cov   2.8.1  Pytest plugin for measuring coverage.
pytest-mock  1.13.0 Thin-wrapper around the mock package for easier use with py.test
pytest-sugar 0.9.2  pytest-sugar is a plugin for pytest that changes the default look and feel of pytest (e.g. progressbar, show tests that fail ...
```

closes: #1906

PR #2351, opened 2020-04-26
poetry show <package> now also shows which packages depend on it

snejus:
# Pull Request Check List

Partly resolves: #1906 

#1906 Requested an ability to know **why** packages are required. Even though the suggestion outlined a reverse-tree based approach, I believe that this can also be solved using a more simple approach: by providing the `required by` information in the `poetry show <package>` output - done with just 15 lines of code. 

I had this patch applied locally since a couple of weeks ago and I found it extremely useful - there's no need to do `poetry show --tree | ag <package> -C 20` anymore just to find out where some 3rd party dependencies come from.

Summary of what's been done:
* 15 lines of business logic in `commands/show.py`
* A test for `poetry show <package>` command - realised a test for this command didn't exist, so it tests both the default output and my changes
* Documentation update and a little typo fix, where `dependencies` isn't supposed to have a colon at the end. 

Please see the screenshot below for a little teaser:
image

### DIFF
--- a/docs/cli.md
+++ b/docs/cli.md
@@ -407,10 +407,13 @@ name        : pendulum
 version     : 1.4.2
 description : Python datetimes made easy
 
-dependencies:
+dependencies
  - python-dateutil >=2.6.1
  - tzlocal >=1.4
  - pytzdata >=2017.2.2
+
+required by
+ - calendar >=1.4.0
 ```
 
 ### Options
--- a/poetry/console/commands/show.py
+++ b/poetry/console/commands/show.py
@@ -180,6 +180,13 @@ lists all packages available."""
 
                 return 0
 
+            required_by = {}
+            for parent in locked_packages:
+                children = {d.pretty_name: d.pretty_constraint for d in parent.requires}
+
+                if pkg.pretty_name in children:
+                    required_by[parent.pretty_name] = children[pkg.pretty_name]
+
             rows = [
                 ["<info>name</>", " : <c1>{}</>".format(pkg.pretty_name)],
                 ["<info>version</>", " : <b>{}</b>".format(pkg.pretty_version)],
@@ -197,6 +204,14 @@ lists all packages available."""
                         " - <c1>{}</c1> <b>{}</b>".format(
                             dependency.pretty_name, dependency.pretty_constraint
                         )
+                    )
+
+            if required_by:
+                self.line("")
+                self.line("<info>required by</info>")
+                for parent, requires_version in required_by.items():
+                    self.line(
+                        " - <c1>{}</c1> <b>{}</b>".format(parent, requires_version)
                     )
 
             return 0
--- a/poetry/console/commands/show.py
+++ b/poetry/console/commands/show.py
@@ -180,12 +180,12 @@ lists all packages available."""
 
                 return 0
 
-            required_by = {}
-            for parent in locked_packages:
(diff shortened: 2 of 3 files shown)


snejus:
image

Can include this as an extra if anyone's keen ;)

seansfkelley:
Looks neat! Does this traverse the ancestry all the way to the root pyproject.toml, or just one level up from the specified package?

snejus:
@seansfkelley - it's the latter, it only specifies direct dependencies. 

I guess the reverse-tree approach would be more suitable for the traversal, which personally I find a bit too noisy, at least for the workflow I'm used to (maybe I just haven't tried a good solution though) - there's actually a PR open that aims to do it: #2086, though it has been stale for a while, so it'd be great if people could have a look at it since it looks quite promising. 

The _required by_ in poetry show removes the noise and simply tells the dev _hey, that's the list of libraries that use it and they specify these constraint - and yeah, now you see which one of those constrains you from bringing that new feature in_. I find it very useful whenever I need to integrate a minor/major bump coming from 3rd party dependency - I see whether it's supported or blocked in a simple way. I found that I'm usually aware of my project dependencies enough to _kind of know_ where they're coming from considering higher levels - it's the versions that cause pain :) There's always an extra `pip show <parent-package>` command available.

The reverse-tree, I guess, has a different use case. One I could think of would be analysing bigger open source projects that devs are new to. `poetry install -vvv` -> sees a dodgy package -> ah where did this come from? -> runs `poetry show --tree --reverse <dodgy-package>` (or `--why`, as in the PR above) -> sees a nice overview of what's going on. 

Actually, `pipdeptree` has implemented it in a way, see a screenshot below for `six` usage within `poetry`:
image

And then `poetry show six` - obviously it's an unfair comparison considering the looks :sweat_smile: 
image

snejus:
image
Forgot to mention - `pip` implements it too - though since _good looking_ or _dependency management_ aren't exactly the keywords usually associated with pip, it's only useful as a backup.

seansfkelley:
Great, thank you for the explanation! `pipdeptree` looks quite handy.

My usual use-case for needing to answer this question is that I've traced an issue through debugging to a particular package, and I want to know which of the dependencies I've declared is bringing it in so I can take (indirect) action to resolve the problem, usually by picking a better version for my dependency or seeing if I can just remove it. All that is to say -- it doesn't come up during version upgrades, so I don't have anything else to narrow down which packages are suspicious.

In any case, this change would already make my life easier, even if it's not exactly what I'm looking for.

snejus:
@sdispater could this by any chance go in, considering that it's very small and has virtually no impact on the rest of the codebase? It's been a couple of months now that me and most of the people around who use Poetry, patching the releases with this - I even have a little bash function that does it. It's extremely useful for bigger projects.

K900:
Just another bump/upvote for this. Very interested in this.

finswimmer:
Hello @snejus,

I really like this one. Would you find some time to rebase your PR onto master?

fin swimmer

snejus:
@finswimmer sorry, was a bit in a rush and forgot to switch the base branch. Is there anything else that's required from my side?

finswimmer:
I'm very sorry to ping you again @snejus. For some reasons the SonarCloud Code Analysis failed. We removed this check for now. So could you rebase once more?

snejus:
of course! :)

finswimmer (approved):
Great :+1:

Issue #2225, opened 2020-03-26
poetry sets incorrect package source type when multiple site-package directories are present


abn:
- [x] I am on the [latest](https://github.com/python-poetry/poetry/releases/latest) Poetry version.
- [x] I have searched the [issues](https://github.com/python-poetry/poetry/issues) of this repo and believe that this is not a duplicate.
- [x] If an exception occurs when executing a command, I executed it again in debug mode (`-vvv` option).

## Issue
TLDR; installed package detection breaks when multiple site-specific directories exist.

In certain scenarios, there exists multiple site-specific package directories. The most
common of this is due to the distinction between `purelib` and `platlib` paths
(see [sysconfig#Installation Paths](https://docs.python.org/3/library/sysconfig.html#installation-paths)).

Another example is the `python:2.7` image.
```console
docker run --rm -it python:2.7 python -c 'import site; print(site.getsitepackages())'
['/usr/local/lib/python2.7/site-packages', '/usr/local/lib/site-python']
```

Different distros have their own view of site-specific file directories. And I suspect, different python installers do too.

The issue can be reproduced using the snippet provided at the end. Here, in an environment with `virtualenv >= 20.0.0`, tox creates the environment such that `lib64` directory is not linked to `lib`.

Note that;
- we use a Red Hat distro (Fedora) as this is one where the `platlib` directory differs from the `purelib` directory.
- **if** the virtual environment was created using the `venv` module or `virtualenv < 20.0.0`, `lib64` will be a symlink to `lib`.
- the change in behaviour for `virtualenv` is by design (https://github.com/pypa/virtualenv/issues/1751)

Once executed you will notice that the first tox run executes as expected.
```console
<snipped>
Package operations: 13 installs, 0 updates, 0 removals, 1 skipped
<snipped>
```

However, the second run will update all packages. This is not expected.
```console
<snipped>
Package operations: 0 installs, 13 updates, 0 removals, 1 skipped
<snipped>
```

The cause for this is the differing source types for the installed package and required package.

https://github.com/python-poetry/poetry/blob/4715f1decabc196baae6b5b3d86cfa40aed81679/poetry/puzzle/solver.py#L95-L96

The root cause for this is because the package is being detected as a non-standard package since
only the first seen site-package directory is being checked.

https://github.com/python-poetry/poetry/blob/4715f1decabc196baae6b5b3d86cfa40aed81679/poetry/repositories/installed_repository.py#L45-L52

[...]

Issue #2079, opened 2020-02-24
Poetry always updates packages when installed to system python


artslob:
- [x] I am on the [latest](https://github.com/python-poetry/poetry/releases/latest) Poetry version.
- [x] I have searched the [issues](https://github.com/python-poetry/poetry/issues) of this repo and believe that this is not a duplicate.
- [x] If an exception occurs when executing a command, I executed it again in debug mode (`-vvv` option).

- **OS version and name**: Linux Debian (in docker)
- **Poetry version**: 1.0.3
- **Link of a [Gist](https://gist.github.com/) with the contents of your pyproject.toml file**: https://gist.github.com/artslob/1e3d70f8e93f85c2a4ab989332edf170

## Issue

In gist link I attached Dockerfile, pyproject.toml and poetry.lock files. Issue is when building docker image I used `poetry config virtualenvs.create false` to install dependencies to current system python interpreter. It works, but executing `poetry install` several times always updates some packages, which already installed:
```
Step 6/7 : RUN poetry install
 ---> Running in 482dd5a533eb
Skipping virtualenv creation, as specified in config file.
Installing dependencies from lock file


Package operations: 0 installs, 13 updates, 0 removals

  - Updating six (1.14.0 /root/.poetry/lib/poetry/_vendor/py3.7 -> 1.14.0)
  - Updating pycparser (2.19 /root/.poetry/lib/poetry/_vendor/py3.7 -> 2.19)
  - Updating webencodings (0.5.1 /root/.poetry/lib/poetry/_vendor/py3.7 -> 0.5.1)
  - Updating zipp (1.1.0 /root/.poetry/lib/poetry/_vendor/py3.7 -> 2.2.0)
  - Updating certifi (2019.11.28 /root/.poetry/lib/poetry/_vendor/py3.7 -> 2019.11.28)
  - Updating cffi (1.13.2 /root/.poetry/lib/poetry/_vendor/py3.7 -> 1.14.0)
  - Updating chardet (3.0.4 /root/.poetry/lib/poetry/_vendor/py3.7 -> 3.0.4)
  - Updating idna (2.8 /root/.poetry/lib/poetry/_vendor/py3.7 -> 2.9)
  - Updating importlib-metadata (1.1.3 /root/.poetry/lib/poetry/_vendor/py3.7 -> 1.5.0)
  - Updating pyparsing (2.4.6 /root/.poetry/lib/poetry/_vendor/py3.7 -> 2.4.6)
[...]

Issue #730, opened 2018-12-11
Poetry install takes too long and installs the package again unnecessarily


Alexspayne:
- [X] I have searched the [issues](https://github.com/sdispater/poetry/issues) of this repo and believe that this is not a duplicate.

## Issue

I'm wondering why it takes 13-30 seconds to install my package (on two different machines). 
This occurs when the package and it's dependencies are already installed.
`$ time poetry install -vvv
Using virtualenv: /Users/apayne/Envs/smithers
Installing dependencies from lock file

Nothing to install or update

  - Skipping markupsafe (1.1.0) Already installed
  - Skipping click (7.0) Already installed
  - Skipping itsdangerous (1.1.0) Already installed
  - Skipping jinja2 (2.10) Already installed
  - Skipping werkzeug (0.14.1) Already installed
  - Skipping aniso8601 (4.0.1) Already installed
  - Skipping flask (1.0.2) Already installed
  - Skipping jsonschema (2.6.0) Already installed
  - Skipping pytz (2018.7) Already installed
  - Skipping six (1.11.0) Already installed
  - Skipping flask-restplus (0.12.1) Already installed
  - Skipping passlib (1.7.1) Already installed
  - Skipping peewee (3.7.1) Already installed
  - Installing smithers (0.1.0)

real	0m29.542s
user	0m5.344s
sys 0m2.370s`

Smithers is the name of the package I'm installing with poetry.  The installing smithers step takes the majority of the time.  Is poetry doing anything but creating a symlink?  I don't understand why this should take so long.  If the package is already installed in editable mode, why does poetry try installing it again?

Much thanks!

sdispater:
The `install` command installs the current project in editable mode by internally creating a temporary `setup.py` file and executing `pip install -e .`. It does it each time `install` is executed to ensure that any new elements like entrypoints are properly installed.

I plan on adding an option to `install` to disable the installation of the current project.

bertjwregeer:
@sdispater is there any reason for the use of a `setup.py` instead of creating the appropriate directory structure/updating the virtual env directly?

If there is a better ticket to mention this in, please let me know.

sdispater:
@bertjwregeer For pure python projects, we could do without the `setup.py` file. This would require creating the proper files and directories which are, as far as I know, an `easy_install.pth` in the `site-packages` directory of the virtualenv and the `{project}.egg-info` directory in the current directory project.

[...]

PR #2360, opened 2020-04-27
Improve editable installs

sdispater:
## Pull Request Check List

Resolves: #730
Resolves: #2079 
Resolves: #2126
Resolves: #2225 

This PR improves and makes proper use of the `EditableBuilder`.

This change affects the `install` command in two ways:

- The root project will now be installed much faster by no longer relying on generating a `setup.py` file.
- The Poetry packages that needs to be installed in editable mode, via Poetry, will also skip the `setup.py` step.

For the details of how it's done:

- It creates a `.pth` file in the `site-packages` (`purelib`) directory of the current environment which points to the absolute path of the current project.
- It creates a `.dist-info` directory with the `INSTALLER`, `METADATA`, `RECORD` and `entry_points.txt` files.
- It creates the appropriate console scripts (entry points).

I took this opportunity to improve the installed packages detection by improving the detection of the `purelib` directory which was not so trivial after all.


PetterS:
@sdispater @abn 

The latest Poetry  on develop now fails on my code base for my editable install with the following message:
```
  AttributeError

  'Package' object has no attribute 'build_script'

  at /usr/local/lib/python3.6/dist-packages/poetry/packages/dependency_package.py:18 in __getattr__
      14|     def clone(self):  # type: () -> DependencyPackage
      15|         return self.__class__(self._dependency, self._package.clone())
      16|
      17|     def __getattr__(self, name):
    > 18|         return getattr(self._package, name)
      19|
      20|     def __setattr__(self, key, value):
      21|         if key in {"_dependency", "_package"}:
      22|             return super(DependencyPackage, self).__setattr__(key, value)
```

Perhaps related to this PR?

sdispater:
@PetterS Thanks for pointing this out.

I stumbled upon this issue this week-end myself and I already have a fix for it that I need to test before making a pull request.

I'll let you know when it's ready.

PetterS:
Cheers! 👍

Issue #2187, opened 2020-03-14
Strange skipping of dependencies when python requirement is specified ("Not needed for the current environment")


adithyabsk:
- [x] I am on the [latest](https://github.com/python-poetry/poetry/releases/latest) Poetry version.
- [x] I have searched the [issues](https://github.com/python-poetry/poetry/issues) of this repo and believe that this is not a duplicate.
- [x] If an exception occurs when executing a command, I executed it again in debug mode (`-vvv` option).

### Poetry Debug Info

```
Poetry
Version: 1.0.5
Python:  3.6.8

Virtualenv
Python:         3.6.8
Implementation: CPython
Path:           /Users/adithyabalaji/.pyenv/versions/3.6.8/envs/test368
Valid:          True

System
Platform: darwin
OS:       posix
Python:   /Users/adithyabalaji/.pyenv/versions/3.6.8
```

### MVP Erroring pyproject.toml
```
[tool.poetry]
name = "test"
version = "0.1.0"
description = ""
authors = ["Adithya"]

[tool.poetry.dependencies]
python = "^3.6.1"

[tool.poetry.dev-dependencies]
seed-isort-config = { version = "^2.1.0", python = "^3.6.1"}

[build-system]
requires = ["poetry>=0.12"]
build-backend = "poetry.masonry.api"
```

## Issue
When I run poetry install I expect all packages to be installed when running python 3.6.8, that doesn't seem to happen and I get the following output. It does work if I remove the python requirement on the dependency.

While this configuration doesn't make much sense in this case, consider the case where multiple python versions need to be supported (ie in my case I'm making the last release for python 2.7 for a package and thus, need to support multiple python versions at least for one more release.

I also know that this bug isn't isolated to seed-isort-config, I tried a few other packages just to be safe and found it produces the same result for flake8, etc... So it seems to be a poetry issue.

```
$ poetry install -vvv
Using virtualenv: /Users/adithyabalaji/.pyenv/versions/3.6.8/envs/test368
Installing dependencies from lock file

No dependencies to install or update

  - Skipping cached-property (1.5.1) Not needed for the current environment
  - Skipping aspy.refactor-imports (2.1.0) Not needed for the current environment
  - Skipping seed-isort-config (2.1.0) Not needed for the current environment
```

adithyabsk:
Might be related to #2177

adithyabsk:
[...]

Issue #2138, opened 2020-03-06
Picking wrong platform installing a dependency with multiple constraints


leon19:
- [x] I am on the [1.0.5](https://github.com/python-poetry/poetry/releases/1.0.5) Poetry version.
- [x] I have searched the [issues](https://github.com/python-poetry/poetry/issues) of this repo and believe that this is not a duplicate.
- [x] If an exception occurs when executing a command, I executed it again in debug mode (`-vvv` option).

- **OS version and name**: Ubuntu 18.04
- **Poetry version**: 1.0.5
- **Link of a [Gist](https://gist.github.com/) with the contents of your pyproject.toml file**: https://gist.github.com/leon19/42bf22ddc791e0500c614d461c32a486

## Issue
Markers are not working properly with multiple constraints dependencies as stated in [the docs](https://python-poetry.org/docs/versions/#multiple-constraints-dependencies)

I am trying to install a dependency from an URL based on the platform, different URLs for every platform (`markers = "sys_platform == 'linux'"`) and poetry is picking the URL from the platform `darwin` when I'm under a `linux` platform

I have not tested yet what happens on a Darwin platform. Will update the issue when I do

## Update:

If I reverse the order of the dependencies, the first line the darwin url, it works on Linux. The same behaviour applies on Darwin

abn:
@leon19 try chaging it to the following (this is only a partial workaround - see below).
```toml
guildai = [
  {url = "https://files.pythonhosted.org/packages/10/ed/ffc1a004da1a272131e776f4b1d24d980b1dbb25b3bbaea4aff5ff7ef538/guildai-0.7.0rc7-cp37-cp37m-manylinux1_x86_64.whl", platform = "linux"},
  {url = "https://files.pythonhosted.org/packages/17/bc/2bc49e33d9343e689723c0b67aa45653d79fc4b5342c12bd7836aca58e7b/guildai-0.7.0rc7-cp37-cp37m-macosx_10_14_x86_64.whl", platform = "darwin"}
]
```

This will work if you are on a linux environment. I did notice that the generated lockfile will only contain the first specified url (in this case command will succeed since I was working on a linux machine and linux platform was specified first). If I were to change the order, it will fail again. However, when building the wheel generated the correct metadata.
```
[...]

Issue #1772, opened 2019-12-19
Invalid requirement, parse error at "'and ()'"


chunyang:
- [x] I am on the [latest](https://github.com/python-poetry/poetry/releases/latest) Poetry version.
- [x] I have searched the [issues](https://github.com/python-poetry/poetry/issues) of this repo and believe that this is not a duplicate.
- [x] If an exception occurs when executing a command, I executed it again in debug mode (`-vvv` option).

- **OS version and name**: Ubuntu 18.04
- **Poetry version**: 1.0.0
- **Link of a [Gist](https://gist.github.com/) with the contents of your pyproject.toml file**: https://gist.github.com/chunyang/a084ce013da1bafb7988163b1dff57ec

## Issue

I am unable to add `apache-beam[gcp]` as a dependency via
```
poetry add "apache-beam==2.16.0" --extras gcp -vvv
```
Attempting to do so gives the following error:
```
...

[InvalidRequirement]
Invalid requirement, parse error at "'and ()'"

Traceback (most recent call last):
  File "/home/chuck.yang/.poetry/lib/poetry/_vendor/py2.7/clikit/console_application.py", line 131, in run
    status_code = command.handle(parsed_args, io)
  File "/home/chuck.yang/.poetry/lib/poetry/_vendor/py2.7/clikit/api/command/command.py", line 120, in handle
    status_code = self._do_handle(args, io)
  File "/home/chuck.yang/.poetry/lib/poetry/_vendor/py2.7/clikit/api/command/command.py", line 171, in _do_handle
    return getattr(handler, handler_method)(args, io, self)
  File "/home/chuck.yang/.poetry/lib/poetry/_vendor/py2.7/cleo/commands/command.py", line 92, in wrap_handle
    return self.handle()
  File "/home/chuck.yang/.poetry/lib/poetry/console/commands/add.py", line 149, in handle
    status = installer.run()
  File "/home/chuck.yang/.poetry/lib/poetry/installation/installer.py", line 74, in run
    self._do_install(local_repo)
  File "/home/chuck.yang/.poetry/lib/poetry/installation/installer.py", line 225, in _do_install
    ops = solver.solve(use_latest=whitelist)
  File "/home/chuck.yang/.poetry/lib/poetry/puzzle/solver.py", line 36, in solve
    packages, depths = self._solve(use_latest=use_latest)
  File "/home/chuck.yang/.poetry/lib/poetry/puzzle/solver.py", line 181, in _solve
[...]

Issue #1728, opened 2019-12-13
Some markers are serialized into the lockfile incorrectly, resulting in crash


madig:
- [x] I am on the [latest](https://github.com/python-poetry/poetry/releases/latest) Poetry version.
- [x] I have searched the [issues](https://github.com/python-poetry/poetry/issues) of this repo and believe that this is not a duplicate.
- [x] If an exception occurs when executing a command, I executed it again in debug mode (`-vvv` option).

- **OS version and name**: Windows 10 x64 1909
- **Poetry version**: 1.0.0
- **Link of a [Gist](https://gist.github.com/) with the contents of your pyproject.toml file**: N/A

## Issue

Package dependency markers that worked in 0.12.17 now crash 1.0.0. Reproducer:

1. `poetry new aaa`
2. `cd aaa`
3. `poetry add -vvv fonttools[unicode]`

This culprit is the following lock entry:
```text
[[package]]
category = "main"
description = "Unicodedata backport for python 2/3 updated to the latest unicode version."
marker = "python_version < \"3.8\" and platform_python_implementation != \"PyPy\" and ()"
name = "unicodedata2"
optional = false
python-versions = "*"
version = "12.1.0"
```

Note the `and ()` in there. Original marker at https://github.com/fonttools/fonttools/blob/fd606ae813f1ed3f8ea42f487968fd036eb06d8a/setup.py#L51-L58.

Backtrace:
```text
Using virtualenv: C:\...\AppData\Local\pypoetry\Cache\virtualenvs\aaa-zvnyiRwK-py3.7
PyPI: No release information found for fonttools-1.0, skipping
PyPI: No release information found for fonttools-2.0b1, skipping
PyPI: 89 packages found for fonttools *
Using version ^4.2.2 for fonttools

Updating dependencies
Resolving dependencies...
   1: fact: aaa is 0.1.0
   1: derived: aaa
   1: fact: aaa depends on fonttools (^4.2.2)
   1: fact: aaa depends on pytest (^5.2)
   1: fact: aaa depends on pytest (^5.2)
   1: selecting aaa (0.1.0)
   1: derived: pytest (^5.2)
   1: derived: fonttools (^4.2.2)
PyPI: 7 packages found for pytest >=5.2,<6.0
PyPI: No release information found for fonttools-1.0, skipping
PyPI: No release information found for fonttools-2.0b1, skipping
PyPI: 1 packages found for fonttools >=4.2.2,<5.0.0
[...]

PR #2361, opened 2020-04-27
Improve dependency resolution and markers handling

sdispater:
# Pull Request Check List

Resolves: #2187 
Resolves: #2138 
Resolves: #1728
Resolves: #1457
Resolves: #907
Resolves: #1772
And possibly more

This PR improves the overall dependency resolution logic and the way we handle environment markers during resolution.

Since managing environment markers was proving to be a headache, the best course of action was to just no longer try to dump and reconcile them in the lock file at the package level.

The environment markers of each dependencies (and transitive dependencies) are now dumped in the lock file. That way we can do a full resolution using the lock file for the current environment before installing.

This was made possible by the previous work in #2342 

This introduces changes to the lock file which are backwards compatible (new versions of Poetry won't break with the older lock files) but not forward compatible (older versions of Poetry might break with this new lock file format).

Unfortunately, there is no way around it and this is the best way to fix a lot of issues with the dependency resolution aspect of Poetry.

Hopefully, the 1.1 release will be good enough for people not to want to downgrade to a previous version 😅

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -67,4 +67,4 @@ jobs:
 
       - name: Run pytest
         shell: bash
-        run: poetry run pytest -q tests
+        run: poetry run pytest -v tests
--- a/poetry/console/commands/debug/resolve.py
+++ b/poetry/console/commands/debug/resolve.py
@@ -28,9 +28,8 @@ class DebugResolveCommand(InitCommand):
     loggers = ["poetry.repositories.pypi_repository"]
 
     def handle(self):
+        from poetry.core.packages.project_package import ProjectPackage
         from poetry.io.null_io import NullIO
-        from poetry.core.packages import ProjectPackage
-        from poetry.installation.installer import Installer
         from poetry.puzzle import Solver
         from poetry.repositories.pool import Pool
         from poetry.repositories.repository import Repository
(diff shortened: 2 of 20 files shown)


flokno:
I just want to let you know that your work is much appreciated and I'm looking forward to poetry 1.1

snejus:
I love this change, however, this is wrong

> This introduces changes to the lock file which are backwards compatible (new versions of Poetry won't break with the older lock files) but not forward compatible (older versions of Poetry might break with this new lock file format).

**It is** a backwards-incompatible change unfortunately, at least from the user's perspective - which is the context it usually refers to. Looks like this will have to go out as `2.x` and is also a bit unfortunate since people can't easily integrate the rest of improvements that are compatible. Do you reckon there's a different way which isn't based on checking out the repo and building it manually?

This probably could have been released as an extra (unless it's very much related to the rest of the changes).

sdispater:
@snejus This is the definition of backwards compatibility, though: new versions of Poetry will understand the old lock files and it won't break any current project.

Also, the lock file, as it is, is an implementation detail and is not part of the public API so the versioning of Poetry is not impacted by the changes in the lock file format, even though we try to make these changes backwards compatible which is the case here.

snejus:
Thanks for clarifying this! I agree with this. So, if I have `1.1.0a2` but my collaborator has `1.1.0a1`, can we continue working on the same project without them requiring to upgrade? When I update the lock file, does Poetry keep the old format in place or change it to the new one?

I assumed the latter and therefore raised the question - if that's not the case please excuse me :)

PR #2371, opened 2020-05-02
Fix some code quality and bug risk issues

pnijhara:
# Pull Request Check List

Resolves: #issue-number-here

---

- Remove unnecessary comprehension
- Remove assert statement from non test files

---

Find the other issues found here - [https://deepsource.io/gh/pnijhara/poetry/issues/?category=recommended](https://deepsource.io/gh/pnijhara/poetry/issues/?category=recommended)

This PR also adds `.deepsource.toml` configuration file to run DeepSource analysis on the repo with. Upon enabling DeepSource, the analysis will run on every PR and commit to detect 560+ types of issues in the changes — including bug risks, anti-patterns, security vulnerabilities, etc.

To enable DeepSource analysis after merging this PR, please follow these steps:

1. [Signup](https://deepsource.io/signup/) on DeepSource with your GitHub account and grant access to this repo.
2. Activate analysis on this repo [here](http://deepsource.io/gh/python-poetry/poetry).

You can also look at the [docs](https://deepsource.io/docs/guides/quickstart.html) for more details. Do let me know if I can be of any help!

### DIFF
--- a/poetry/mixology/incompatibility.py
+++ b/poetry/mixology/incompatibility.py
@@ -57,7 +57,8 @@ class Incompatibility:
                     # irrelevant, since we already know that mutually exclusive version
                     # ranges are incompatible. We should never derive an irrelevant
                     # incompatibility.
-                    assert by_ref[ref] is not None
+                    if by_ref[ref] is None:
+                        raise AssertionError
                 else:
                     by_ref[ref] = term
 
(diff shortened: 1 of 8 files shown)


poetry/mixology/incompatibility.py
```diff
@@ -57,7 +57,8 @@ def __init__(
                     # irrelevant, since we already know that mutually exclusive version
                     # ranges are incompatible. We should never derive an irrelevant
                     # incompatibility.
-                    assert by_ref[ref] is not None
+                    if by_ref[ref] is None:
+                        raise AssertionError
```
finswimmer, line 61: Hello @pnijhara,

thanks a lot for your contribution and trying to make the code better. However, I don't agree with the suggested changes here.

Using `assert` outside a test is in my opinion absolutely valid. They can be used as an internal self-check. Usually they should never raise, but if so you know there is something wrong in your code. The change you suggest here is just another way of writing the same thing.

poetry/semver/version_union.py
```diff
@@ -38,7 +38,7 @@ def of(cls, *ranges):
         if not flattened:
             return EmptyConstraint()
 
-        if any([constraint.is_any() for constraint in flattened]):
+        if any(constraint.is_any() for constraint in flattened):
```
finswimmer, line 41: Using generator expression is a good idea :+1:

poetry/mixology/incompatibility.py
funkyfuture: i don't now whether this is the intention, but the changes would also raise errors when the interpreter would be called with the [-O option](https://docs.python.org/3/using/cmdline.html#cmdoption-o).

pnijhara: Well, that's the intention. As far as I know, while running python with -O (optimize) flag, it will remove the assert statements. So, it is a good practice to not to use assert statements outside of tests IMHO.

mvoitko: @pnijhara AFAIK asserts are also slow. Maybe you could run a small performance test as an additional argument?

pnijhara: @mvoitko I am not sure about the speed. But yes, I can test it with `timeit`.

Secrus:
Closing. This is now 2 years old, with no updates. Codebase changed significantly in that time.

PR #2375, opened 2020-05-02
Simplify printing package operations

PetterS:
The created strings aren't used.


PetterS:
@stephsamson should be relatively uncontroversial.

PR #2378, opened 2020-05-03
Fix mispelled content-type in release action

radusuciu:
Assuming this wasn't poetic expression: `text/pain` -> `text/plain`


sdispater:
Thanks!

Issue #1632, opened 2019-11-26
Allow user to override PyPI URL without modifying pyproject.toml


JacobHenner:
- [X] I have searched the [issues](https://github.com/sdispater/poetry/issues) of this repo and believe that this is not a duplicate.
- [X] I have searched the [documentation](https://poetry.eustace.io/docs/) and believe that my question is not covered.

## Feature Request
Similar to one of the proposals in https://github.com/sdispater/poetry/issues/1070 (which was recently marked stale), Poetry should allow the user to override the default repository URL (PyPI). The user should be able to do this without modifying pyproject.toml.

In certain environments (e.g. corporate networks) PyPI is unavailable, but a mirror exists. These users should be able to specify the address of the mirror without modifying project files, as the mirror settings are irrelevant to other contributors in different environments. Similarly, if a mirror user adds a dependency, the generated lock file should not list the user's mirror as the source. The source should remain the default (which in most cases would refer to standard PyPI). 

This feature exists in pipenv, see https://github.com/pypa/pipenv/issues/2075 (where the need for this functionality is described in greater detail) and https://github.com/pypa/pipenv/pull/2281.

countergram:
This is essential for many business uses, not simply when PyPI is unavailable but also in any case where the organization has its own libraries (not uncommon). Note that since some private repo tools (e.g. Nexus) use basic auth URLs, putting the repo URL into a project config file is absolutely inappropriate and a global config or environment variable (e.g. pip.conf, PIP_INDEX_URL) is necessary.

cjw296:
https://github.com/sdispater/poetry/issues/625 also seems related.

Something I tried, which might be nice to make work:
```
poetry config repositories.pypi https://.../+simple/
```

vlcinsky:
@sdispater, I wonder, if #1070 elaboration of requested feature is usable as is or it needs some update. If so, I volunteer to join effort with one or few others, have a telco and try to move this request on as this is one of two showstoppers for our usage of poetry (the other is managing versions of the resulting package - but this I definitely do not want to discuss here).

lovepocky:
a simple patch for ci/cd:

```Dockerfile
# install dependencies from lock file
COPY pyproject.toml poetry.lock /opt/app/

RUN sed -i "s/${origin_pypi_url}/${private_pypi_cache_url}/g" poetry.lock
RUN sed -i "s/${origin_pypi_url}/${private_pypi_cache_url}/g" pyproject.toml

[...]

Issue #664, opened 2018-11-23
[ConnectionError] when I use Chinese pypi mirror


zbjdonald:
- [x] I am on the latest Poetry version.
- [x] I have searched the issues of this repo and believe that this is not a duplicate.
- [x] If an exception occurs when executing a command, I executed it again in debug mode (-vvv option)

- OS: Ubuntu 16,04
- poetry version: 0.12.10

## Issue

This is my related config:
```
[[tool.poetry.source]]
name = "ustc"
url = "http://mirrors.ustc.edu.cn/pypi/web/simple/"
```
When I use Chinese mirror to speed up installing procedure, updating dependencies takes a long time and reach following error:
```
[ConnectionError]                                                                 
('Connection aborted.', RemoteDisconnected('Remote end closed connection without response',))
```

Exception trace:
```
..............
ustc: The cache for pytest-flake8 1.0.2 is outdated. Refreshing.
ustc: Downloading wheel: pytest_flake8-1.0.2-py2.py3-none-any.whl
   1: Version solving took 390.841 seconds.
   1: Tried 1 solutions.
                                                                                                 
[ConnectionError]                                                                 
('Connection aborted.', RemoteDisconnected('Remote end closed connection without response',))  
                                                                                                 
Exception trace:
 /home/zbjdonald/.poetry/lib/poetry/_vendor/py3.6/cleo/application.py in run() at line 94
   status_code = self.do_run(input_, output_)
 /home/zbjdonald/.poetry/lib/poetry/console/application.py in do_run() at line 88
   return super(Application, self).do_run(i, o)
 /home/zbjdonald/.poetry/lib/poetry/_vendor/py3.6/cleo/application.py in do_run() at line 197
   status_code = command.run(input_, output_)
 /home/zbjdonald/.poetry/lib/poetry/console/commands/command.py in run() at line 77
   return super(BaseCommand, self).run(i, o)
 /home/zbjdonald/.poetry/lib/poetry/_vendor/py3.6/cleo/commands/base_command.py in run() at line 146
   status_code = self.execute(input_, output_)
 /home/zbjdonald/.poetry/lib/poetry/_vendor/py3.6/cleo/commands/command.py in execute() at line 107
   return self.handle()
 /home/zbjdonald/.poetry/lib/poetry/console/commands/install.py in handle() at line 57
   return_code = installer.run()
 /home/zbjdonald/.poetry/lib/poetry/installation/installer.py in run() at line 76
   self._do_install(local_repo)
[...]

Issue #559, opened 2018-10-25
Force Poetry to not use public PyPI


zehauser:
- [x] I have searched the [issues](https://github.com/sdispater/poetry/issues) of this repo and believe that this is not a duplicate.
- [x] I have searched the [documentation](https://poetry.eustace.io/docs/) and believe that my question is not covered.

My sincere apologies is this is answered elsewhere. I tried my best to find a clear answer in the docs and Github issues but could not.

I would like to be able to ensure that all dependencies are installed from a private repository, and that will not use public PyPI is not used for any dependency resolution. In other words, I want behavior like `pip install --index-url`. 

Basically, I want to ensure that Poetry fails with an error if I've accidentally failed to add a dependency to my private PyPI mirror.

Is this possible with Poetry currently? If not, would a feature request or PR be accepted?

madig:
I think it's enough if you add
```
[[tool.poetry.source]]
name = "private-pypi"
url = "https://pypi.yourdomain.com/simple/"
```
to your `pyproject.toml`. Poetry seems to exclusively use `--index-url`, not `--extra-index-url` in this case.

nirvana-msu:
I have a related, but more general request. I have to work within a firewalled environment where `pypi.org` is not available. Attempting to make HTTP request to it just produces a DNS errors. Instead we've got an internal PyPI mirror. While `poetry install` works well with `[tool.poetry.source]` config, many other commands always fallback to and/or exclusively use `pypi.org`, e.g. `poetry self:update` (yes, I'd like to be able to update poetry from an internal PyPI mirror), `poetry search` etc making them useless.

I would like to have an option to completely ignore the public `pypi.org` across all commands, either by respecting the global `[repositories]` config setting, or some other setting, or maybe by passing explicitly via command options (less convenient). Either I'm missing something, or there currently seems to be no way at all to use these commands in a firewalled environment.

To be honest, when I first came across the `[repositories]` config setting, I thought it would do exactly that (only use the provided repositories across all commands) but unfortunately it's only being used for publishing.

5long:
I'd like to chime in here:

For developers living in China, downloading packages from the canonical PyPI (strictly speaking, `files.pythonhosted.org`) might be pretty slow. Right now it's just 200 KB/s in my office.

Being able to download from a local mirror of PyPI could be life-changing.

craynic:
It seems that Poetry will respect option `index-url` in `pip.conf`.
[...]

PR #2395, opened 2020-05-08
Allow to specify source url in global pip config

LittleboyHarry:
# Pull Request Check List

Resolves: #1632 #559 #664

Let poetry resolve pip config `global.index-url` as default pypi repository if exists.

### DIFF
--- a/poetry/factory.py
+++ b/poetry/factory.py
@@ -78,7 +78,35 @@ class Factory(BaseFactory):
         # Always put PyPI last to prefer private repositories
         # but only if we have no other default source
         if not poetry.pool.has_default():
-            poetry.pool.add_repository(PyPiRepository(), True)
+            # try to read customized pip source url from global pip config if exists
+            from .utils._compat import Path
+            from .utils.env import SystemEnv,EnvCommandError
+            try:
+                import sys
+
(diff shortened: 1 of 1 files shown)


LittleboyHarry:
Auto merge review check bot fails because some old code in this file don't fit `black` format style.

sdispater:
Thanks for your contribution!

However, I do not feel comfortable using information coming from other tools automatically. This should be done explicitly by the user.

LittleboyHarry:
Hope there will be one way to specify **custom default pip repository in poetry global config** in the new version.

DanielW10001:
@sdispater How about promoting the user to ask whether to use the `global.index-url` from pip config when it presents?

swist:
This feature would be very useful for scenarios where jwt for authenticating with the registry is prepended to the beginning of the repo url, AWS codebuild for example builds repo urls like so:
```
https://aws:<JWT>@<domain>-<aws-account>.d.codeartifact.eu-west-1.amazonaws.com/pypi/python/simple/
```
Current setup that requires poetry users to define this as a static url inside of `pyproject.tml` makes it impossible to use (because these are sessioned to ~12hours, JWT gets re-rolled afterwards)

Issue #2386, opened 2020-05-05
Dataclasses solver failure with conditional python marker


mands:
- [x] I am on the [latest](https://github.com/python-poetry/poetry/releases/latest) Poetry version.
- [x] I have searched the [issues](https://github.com/python-poetry/poetry/issues) of this repo and believe that this is not a duplicate.
- [x] If an exception occurs when executing a command, I executed it again in debug mode (`-vvv` option).

- Linux (Fedora 32):
- **Poetry version**:
```bash
$ poetry debug info

Poetry
Version: 1.0.5
Python:  3.8.2

Virtualenv
Python:         3.8.2
Implementation: CPython
Path:           /home/projects/poetry-dataclasses/poetry-demo/.venv
Valid:          True

System
Platform: linux
OS:       posix
Python:   /usr
```
### pyproject.yaml

```yaml
[tool.poetry]
name = "poetry-demo"
version = "0.1.0"
description = ""
authors = []

[tool.poetry.dependencies]
python = "^3.6.1"
dataclasses = {version = "^0.7", python = "~3.6.1"}
dacite = "^1.5.0"

[build-system]
requires = ["poetry>=0.12"]
build-backend = "poetry.masonry.api"
```

## Issue
Following on from #1413, the solution there seems to suggest user python version markers to specify when to install a conditional dependencty - in my case the dataclasses-backport on python 3.6.*. However this seems to result in a `SolverProblemError` resolving in certain cases even with the python specifier as indicated in my `pyproject.yaml` above .

Things seem to work fine until I also install the addional library `dacite` (https://github.com/konradhalas/dacite) - which also has a dependency on dataclasses when using python 3.6. The `setup.py` for it is as follows,

```
    install_requires=['dataclasses;python_version<"3.7"'],
```

Once this library is also installed, adding any other packages to my project results in the aforementioned `SolverProblemError`, e.g.

```bash
$ poetry add arrow
Using version ^0.15.6 for arrow

Updating dependencies
Resolving dependencies... (0.0s)

[SolverProblemError]
The current project's Python requirement (^3.6.1) is not compatible with some of the required packages Python requirement:
  - dataclasses requires Python >=3.6, <3.7

Because dataclasses (0.7) requires Python >=3.6, <3.7
 and no versions of dataclasses match >0.7,<0.8, dataclasses is forbidden.
So, because poetry-demo depends on dataclasses (^0.7), version solving failed.
```

[...]

Issue #2076, opened 2020-02-22
Dependency (pytype) rejected for project supporting Python 3.6+ even though compatible


cjolowicz:
Poetry's solver rejects installation of pytype in a project supporting Python 3.6+, even though pytype's Python constraint is compatible with it.

Repro:

1. Add the following **pyproject.toml**

```toml
[tool.poetry]
name = "foobar"
version = "0.1.0"
description = ""
authors = ["User <user@example.com>"]

[tool.poetry.dependencies]
python = "^3.6"

[tool.poetry.dev-dependencies]

[build-system]
requires = ["poetry>=0.12"]
build-backend = "poetry.masonry.api"
```

2. Run `poetry add --dev --python="<3.8" pytype`

This results in the following error message:

```sh
Creating virtualenv poetry-pytype-VZF4hhnk-py3.7 in ~/Library/Caches/pypoetry/virtualenvs
Using version ^2020.2.20 for pytype

Updating dependencies
Resolving dependencies... (0.0s)

[SolverProblemError]
The current project's Python requirement (^3.6) is not compatible with some of the required packages Python requirement:
  - pytype requires Python !=3.0.*,!=3.1.*,!=3.2.*,!=3.3.*,!=3.4.*,<3.8,>=2.7

Because no versions of pytype match >2020.2.20,<2021.0.0
 and pytype (2020.2.20) requires Python !=3.0.*,!=3.1.*,!=3.2.*,!=3.3.*,!=3.4.*,<3.8,>=2.7, pytype is forbidden.
So, because poetry-pytype depends on pytype (^2020.2.20), version solving failed.
```

The same error happens with the command `poetry add --dev pytype`. (The `--python="<3.8"` was added because pytype currently does not support Python 3.8.)

cjolowicz:
pytype 2020.2.20 added the following python_requires declaration:

```
python_requires = >=2.7, !=3.0.*, !=3.1.*, !=3.2.*, !=3.3.*, !=3.4.*, <3.8
```

IIUC, Poetry refuses to install pytype because the root project declares compatibility with Python ^3.6. This includes 3.8 and up, which would not be supported by pytype. Poetry does not care that we only install pytype in environments with Python < 3.8, using the `--python` option to `poetry add`.

It seems, when determining compatibility of Python version constraints, Poetry does not take into account which Python versions a dependency is installed for.

cjolowicz:
Adding an explicit `>=3.6` works:

```
[...]

Issue #2054, opened 2020-02-18
Unhelpful/wrong error message


nicolas-leydet:
On Poetry version 1.0.3

Steps to reproduce:
```
mktmpenv -p python3.7
pip install poetry
poetry init
poetry add fluent-logger
```

Error message: 
```
Using version ^0.9.4 for fluent-logger

Updating dependencies
Resolving dependencies... (0.0s)

[SolverProblemError]
The current project's Python requirement (^3.7) is not compatible with some of the required packages Python requirement:
  - fluent-logger requires Python >=2.7,!=3.0,!=3.1,!=3.2,!=3.3,<3.9

Because no versions of fluent-logger match >0.9.4,<0.10.0
 and fluent-logger (0.9.4) requires Python >=2.7,!=3.0,!=3.1,!=3.2,!=3.3,<3.9, fluent-logger is forbidden.
So, because tmp-ef8119aac07c59ce depends on fluent-logger (^0.9.4), version solving failed.
```

The same error happens with python version 3.6 but not with 2.7
A classic `pip install fluent-logger` is installing the package without error

There's maybe an issue with fluent-logger but the message doesn't seem to describe the issue correctly.

absassi:
I have a similar issue with PySide2, which also limits itself to `<3.9`.

I had to set my Python dependency to match exactly the PySide2's Python dependency (which is also the intersection of all Python version constraints in my set of dependencies) to be able to resolve:

    python = ">= 3.7, != 3.8.0, < 3.9"

One can argue that Poetry is correct, since `^3.7` includes Python versions `>3.9,<4.0`, for which there is no `fluent-logger` (or `PySide2`) available, so it can't really resolve them (for all allowed Python versions and platforms).

I agree the message could be more clear, stating the range of versions where the requirement fails, as I spend quite some time trying to figure out what was happening. Maybe Poetry should be clear in the error message and in the docs that dependency resolution must satisfy all allowed Python versions and platforms, and not just the currently selected environment.

jules-ch:
@absassi

> One can argue that Poetry is correct, since ^3.7 includes Python versions >3.9,<4.0
 There is an error, the python constraint here, python=^3.7 translates to >=3.7,<4.0 during the python constaint resolution process in poetry 1.0.5.

So poetry should be able to resolve <3.9 constraint.

Issue #1930, opened 2020-01-22
Adding PySide2 fails to resolve version


madig:
- [x] I am on the [latest](https://github.com/python-poetry/poetry/releases/latest) Poetry version.
- [x] I have searched the [issues](https://github.com/python-poetry/poetry/issues) of this repo and believe that this is not a duplicate.
- [ ] ~~If an exception occurs when executing a command, I executed it again in debug mode (`-vvv` option).~~

- **OS version and name**: Windows 10, Python 3.8.1
- **Poetry version**: 1.0.2
- **Link of a [Gist](https://gist.github.com/) with the contents of your pyproject.toml file**: See below

## Issue

Adding PySide2 as a dependency seems to trigger a resolver bug. See below. `python = "^3.8"` seems to clash with PySide2's limiter `< 3.9`. The package can be installed without problems in a 3.8.1 venv (pyside2 5.14.0 only works on Python 3.8.1+ on Windows).

```
~\AppData\Local\Temp
> poetry new ppp
Created package ppp in ppp
```
```
~\AppData\Local\Temp
> cd .\ppp\
```
```
~\AppData\Local\Temp\ppp
> cat .\pyproject.toml
[tool.poetry]
name = "ppp"
version = "0.1.0"
description = ""
authors = ["..."]

[tool.poetry.dependencies]
python = "^3.8"

[tool.poetry.dev-dependencies]
pytest = "^5.2"

[build-system]
requires = ["poetry>=0.12"]
build-backend = "poetry.masonry.api"
```
```
~\AppData\Local\Temp\ppp
> poetry add pyside2=="^5.14"
Creating virtualenv ppp in C:\...\Temp\ppp\.venv

Updating dependencies
Resolving dependencies...

[SolverProblemError]
The current project's Python requirement (^3.8) is not compatible with some of the required packages Python requirement:
  - pyside2 requires Python >=2.7, !=3.0.*, !=3.1.*, !=3.2.*, !=3.3.*, !=3.4.*, <3.9

Because no versions of pyside2 match >5.14,<6.0
 and pyside2 (5.14.0) requires Python >=2.7, !=3.0.*, !=3.1.*, !=3.2.*, !=3.3.*, !=3.4.*, <3.9, pyside2 is forbidden.   So, because ppp depends on pyside2 (^5.14), version solving failed.
```

madig:
Switching to PyQt5 works around the issue.

JrooTJunior:
[...]

PR #2396, opened 2020-05-08
Add solutions for common errors

sdispater:
# Pull Request Check List

Resolves: #2386 
Resolves: #2076
Resolves: #2054
Resolves: #1930

This PR adds a way to provide solutions for common errors that users may encounter while using Poetry.

For now, the only solution implemented is for the resolution process failures due to Python requirements incompatibilities, which is the most common source of confusion regarding the way the resolver works.

This will serve as the base implementation for future solutions that we might want to add to provide an overall better user experience.

Here is an example:

Screenshot 2020-05-08 at 11 23 41

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -215,7 +215,7 @@ version = "0.4.3"
 [[package]]
 category = "main"
 description = "Updated configparser from Python 3.7 for Python 2.6+."
-marker = "python_version >= \"2.7\" and python_version < \"2.8\" or python_version == \"2.7\" and python_version == \"2.7\" or python_version < \"3\""
+marker = "python_version < \"3\" or python_version >= \"2.7\" and python_version < \"2.8\""
 name = "configparser"
 optional = false
 python-versions = ">=2.6"
@@ -228,7 +228,7 @@ testing = ["pytest (>=3.5,<3.7.3 || >3.7.3)", "pytest-checkdocs (>=1.2)", "pytes
 [[package]]
 category = "main"
 description = "Backports and enhancements for the contextlib module"
-marker = "python_version >= \"2.7\" and python_version < \"2.8\" or python_version < \"3.4\""
+marker = "python_version < \"3.4\" or python_version >= \"2.7\" and python_version < \"2.8\""
 name = "contextlib2"
 optional = false
 python-versions = ">=2.7, !=3.0.*, !=3.1.*, !=3.2.*, !=3.3.*"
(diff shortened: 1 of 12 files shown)


poetry/mixology/version_solver.py
```diff
@@ -40,7 +39,7 @@ class VersionSolver:
     def __init__(
         self,
         root,  # type: ProjectPackage
-        provider,  # type: Provider
+        provider,
```
abn, line 42: Is there a reason why we are dropping the type hint here?

poetry.lock
```diff
@@ -223,6 +223,15 @@ version = "5.1"
 [package.extras]
 toml = ["toml"]
 
+[[package]]
+category = "main"
+description = "Manage Python errors with ease"
+marker = "python_version >= \"3.6\" and python_version < \"4.0\""
+name = "crashtest"
```
abn, line 230: I can see crashtest on pypi but not on github, is this expected?

poetry/solution_providers/__init__.py
```diff
@@ -0,0 +1,3 @@
+from .python_requirement_incompatibility_solution_provider import (
```
abn, line 1: These modules are a mouth full :) Would be great to put them into a structure, eg `poetry.solutions.providers.incompatibility.python_requreiment`. 

Would `solutions` be better placed in `poetry.mixology` consdiering we only really provide solutions for version solving issues or is the intent here to provide a generic mechanism?

poetry/mixology/version_solver.py
sdispater: I don't remember exactly. I thin I got a cyclic import error at some point due to this import. I can check again to see if we really need to remove it.

poetry/solution_providers/__init__.py
sdispater: I wanted them to be as explicit as possible :-) But yeah I can see if we can reorganize a bit.

The idea is to extend the solutions beyond solver issues, for instance for configuration and validation errors.

poetry.lock
sdispater: It's not open source yet, I need to write a proper README for the project first. It should be ready tomorrow.

poetry/mixology/version_solver.py
sdispater: So, I just checked and we need to remove it to avoid a cyclic import error

poetry/solution_providers/__init__.py
sdispater: So, I reorganized a bit and now each module (in this case `mixology`) can provide its own solutions and providers.

This allows for a more flexible implementation.

poetry.lock
sdispater: It's now open source: https://github.com/sdispater/crashtest

PR #2400, opened 2020-05-09
Fix editable installations for Poetry packages

sdispater:
# Pull Request Check List

Followup to #2360 

Basically, editable installs for transitive packages failed due to the `build_script` attribute not being available on standard `Package` objects.

Also, the generated scripts weren't correct if the callable was a module attribute.

Finally, the loading of editable installed packages was not correct and they weren't marked as being directory dependencies which would lead to them being always updated when installing.

Note that we will rely on [PEP 610](https://www.python.org/dev/peps/pep-0610/) in the future for detection of such packages.

### DIFF
--- a/poetry/installation/pip_installer.py
+++ b/poetry/installation/pip_installer.py
@@ -1,7 +1,6 @@
 import os
 import tempfile
 
-from io import open
 from subprocess import CalledProcessError
 
 from clikit.api.io import IO
@@ -177,7 +176,6 @@ class PipInstaller(BaseInstaller):
     def install_directory(self, package):
         from poetry.factory import Factory
         from poetry.io.null_io import NullIO
-        from poetry.utils._compat import decode
         from poetry.masonry.builders.editable import EditableBuilder
         from poetry.utils.toml_file import TomlFile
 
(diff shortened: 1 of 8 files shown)


poetry/installation/pip_installer.py
```diff
@@ -210,27 +210,27 @@ def install_directory(self, package):
...
-        if has_poetry and package.develop and not package.build_script:
-            # This is a Poetry package in editable mode
-            # we can use the EditableBuilder without going through pip
-            # to install it, unless it has a build script.
-            builder = EditableBuilder(
-                Factory().create_poetry(pyproject.parent), self._env, NullIO()
-            )
-            builder.build()
+        if has_poetry:
+            package_poetry = Factory().create_poetry(pyproject.parent)
```
abn, line 214: We should probably only instantiate if it is requred.

poetry/installation/pip_installer.py
```diff
@@ -210,27 +210,27 @@ def install_directory(self, package):
...
-            builder = SdistBuilder(Factory().create_poetry(pyproject.parent))
+                # We need to rely on creating a temporary setup.py
+                # file since the version of pip does not support
+                # build-systems
+                # We also need it for non-PEP-517 packages
+                builder = SdistBuilder(package_poetry)
 
-            with open(setup, "w", encoding="utf-8") as f:
-                f.write(decode(builder.build_setup()))
+                with open(setup, "w", encoding="utf-8") as f:
```
abn, line 232: You can consider reworking this so that we can use the context manager from `SdistBuilder`.

```py
with SdistBuilder(package_poetry).setup_py() as setup:
    pass
```

poetry/installation/pip_installer.py
sdispater: We need it just below for the `build_script` check. Or do you have something else in mind?

poetry/installation/pip_installer.py
sdispater: I will change it. Thanks for pointing this out.

sdispater: I changed it.

poetry/installation/pip_installer.py
abn: I was more thinking of using an `isinstance` or `hasattr` check. But this is not strictly required.

PetterS:
@sdispater Anything left to do here? I'd love to give this a spin.

PR #2402, opened 2020-05-10
Fix #2089: Make Python implementation type as part of the venv name.

### LINKED ISSUE #2089: `poetry use pypy3` switches to a Python 3.6 venv if it already exists

- **OS version and name**: Ubuntu 18.04
- **Poetry version**: 1.0.3
- **Link of a [Gist](https://gist.github.com/) with the contents of your pyproject.toml file**: https://gist.github.com/thedrow/5b7d581db41f2d8d40ca9e4c9c98d283

## Issue

When I try to switch my environment to PyPy3 using ``poetry use pypy3`` but have previously typed `poetry use 3.6` poetry switches to the *CPython* 3.6 virtualenv instead of creating a new virtualenv for pypy3.
You can reproduce this by doing the following:
```
$ poetry env use 3.6
Using virtualenv: /home/thedrow/.cache/pypoetry/virtualenvs/jumpstarter-H8PwDp0q-py3.6
$ poetry env use pypy3
Using virtualenv: /home/thedrow/.cache/pypoetry/virtualenvs/jumpstarter-H8PwDp0q-py3.6
$ poetry run which pypy3
/home/thedrow/.pyenv/shims/pypy3 # should be the virtualenv's pypy
```

PetterS:
Resolves: #2089. Now "pypy" is part of the venv name so that it can be used together with poetry. Virtualenvs using Cpython will not change names to ensure backwards compatibility.

### DIFF
--- a/poetry/utils/env.py
+++ b/poetry/utils/env.py
@@ -259,7 +259,11 @@ class EnvManager(object):
             self.create_venv(io, executable=python, force=create)
 
         # Activate
-        envs[base_env_name] = {"minor": minor, "patch": patch}
+        envs[base_env_name] = {
+            "minor": minor,
+            "patch": patch,
+            "implementation": python_implementation,
+        }
         envs_file.write(envs)
 
         return self.get(reload=True)
@@ -279,9 +283,11 @@ class EnvManager(object):
             envs = envs_file.read()
             env = envs.get(name)
             if env is not None:
+                python_implementation = env.get("implementation", "py")
                 io.write_line(
                     "Deactivating virtualenv: <comment>{}</comment>".format(
-                        venv_path / (name + "-py{}".format(env["minor"]))
+                        venv_path
+                        / (name + "-{}{}".format(python_implementation, env["minor"]))
                     )
                 )
                 del envs[name]
@@ -293,6 +299,9 @@ class EnvManager(object):
             return self._env
 
         python_minor = ".".join([str(v) for v in sys.version_info[:2]])
+        python_implementation = self.normalize_python_implementation(
+            platform.python_implementation()
+        )
 
         venv_path = self._poetry.config.get("virtualenvs.path")
         if venv_path is None:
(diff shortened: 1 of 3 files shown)


PetterS:
Does not yet work since the "-py" prefix is still hard-coded in a few more places in `env.py`. I don't know enough about how Poetry fetches its config to fix it right now. 

When the following command is performed:
```
poetry env use /path/to/pypy3
```
Poetry prints:
```
Creating virtualenv poetry-0nHquv0u-pypy3.6 in /home/petter/.cache/pypoetry/virtualenvs                                                                                                                          Using virtualenv: /home/petter/.cache/pypoetry/virtualenvs/poetry-0nHquv0u-py3.6
```
So it creates the correct environment, but then still decides to use the old one.

PetterS:
It now works for me locally with pypy, so this is something that can be discussed.

thedrow:
I'll give it a shot :)

PetterS:
@abn would be nice if we could discuss this.

abn:
In principle the change looks good. How would this handle reuse of an existing environment for the project?

PetterS:
Since this PR only affects the name of the directory for Python implementations other than CPython, I think it should be fine?

Secrus:
@PetterS are you still interested in bringing it to Poetry? If so, please merge changes from master branch and resolve potential conflicts. Whats more, please include some documentation on that matter.

PetterS:
Yes, I think it would be useful.

PetterS:
Well this patch does not apply anymore. It has been 2 years! 
But there is some more support for pypy now it seems: [interpreter_name](https://github.com/python-poetry/poetry/blob/979d945863ca3b9915b23653285af82efdbcf73f/src/poetry/utils/env.py#L93)

perhaps it works now? Otherwise, it should be easier to set the environment name since the info is already fetched.

thedrow:
It behaves exactly the same according to the code but yes, I should be much easier to implement this behavior now.

Issue #1941, opened 2020-01-24
[FAQ]: The tox answer should recommend `poetry install --no-root` instead of `poetry install`


thejohnfreeman:
- [x] I have searched the [issues](https://github.com/python-poetry/poetry/issues) of this repo and believe that this is not a duplicate.

In the FAQ, there is a [recommendation](https://python-poetry.org/docs/faq/#is-tox-supported) for how to use tox with Poetry. It suggests this `tox.ini`:

```ini
[tox]
isolated_build = true
envlist = py27, py36

[testenv]
whitelist_externals = poetry
commands =
    poetry install -v
    poetry run pytest tests/
```

With this configuration, when a user runs tox:

1. tox packages the project with Poetry.
2. tox creates a virtual environment.
3. tox installs the package it built, but not its dependencies (`inst-nodeps`).
4. tox calls `poetry install`, which installs the project dependencies, and then **installs the project, as editable**, overwriting what tox installed in step 3.
5. tox calls `poetry run pytest tests` which **tests against the package as it exists in the working directory, not as it was packaged by Poetry**.

To fix this, the recommended `commands` should be:

```ini
commands =
    poetry install --no-root -v
    poetry run pytest tests/
```

sinoroc:
I believe _poetry_ should be left out of `tox.ini` entirely. This is the pattern I would recommend:

`tox.ini`
```
[tox]
envlist = py{36,37,38}-django{22,30}
isolated_build = True

[testenv]
# ...
deps =
    django22: Django==2.2
    django30: Django==3.0
extras =
    test
commands =
    pytest
```

`pyproject.toml`
```
[tool.poetry]
# ...

[tool.poetry.dependencies]
python = "^3.6"
django = "^2.2 || ^3.0"
#
pytest = { version = "^5.2", optional = true }

[tool.poetry.extras]
test = ["pytest"]

[build-system]
# ...
```

Alexspayne:
> I believe _poetry_ should be left out of `tox.ini` entirely. This is the pattern I would recommend:
> 
> `tox.ini`
> 
> ```
> [tox]
> envlist = py{36,37,38}-django{22,30}
> isolated_build = True
> 
> [testenv]
> # ...
> deps =
>     django22: Django==2.2
>     django30: Django==3.0
> extras =
>     test
> commands =
>     pytest
> ```
> 
> `pyproject.toml`
> 
> ```
> [tool.poetry]
> # ...
> 
[...]

PR #2416, opened 2020-05-13
Suggest Tox isolated_build w/o poetry install

bittner:
Adapts the suggestion about how to integration Poetry with Tox (roughly) following @sinoroc's suggestion. The new way avoids a common anti-pattern (or better: outdated approach) in CI, the imperative installation of prerequisites.

Resolves: #1941

### DIFF
--- a/docs/docs/faq.md
+++ b/docs/docs/faq.md
@@ -38,26 +38,24 @@ The `^` operator works very well with libraries following [semantic versioning](
 Yes. By using the [isolated builds](https://tox.readthedocs.io/en/latest/config.html#conf-isolated_build) `tox` provides,
 you can use it in combination with the PEP 517 compliant build system provided by Poetry.
 
-So, in your `pyproject.toml` file, add this section if it does not already exist:
+In your `pyproject.toml` file, ensure that the following section is present:
 
 ```toml
 [build-system]
-requires = ["poetry>=0.12"]
+requires = ["poetry>=1.0"]
 build-backend = "poetry.masonry.api"
 ```
 
-And use a `tox.ini` configuration file similar to this:
+Use a `tox.ini` configuration file similar to this:
 
-```INI
+```ini
 [tox]
+envlist = py36, py37
 isolated_build = true
-envlist = py27, py36
 
 [testenv]
-whitelist_externals = poetry
-commands =
-    poetry install -v
-    poetry run pytest tests/
+deps = pytest
+commands = pytest {posargs}
 ```
 
 ## I don't want Poetry to manage my virtual environments. Can I disable it?
@@ -68,6 +66,6 @@ and is an overhead, like when working with containers.
 
 In this case, you can disable this feature by setting the `virtualenvs.create` setting to `false`:
 
-```bash
+```console
 poetry config virtualenvs.create false
 ```


igor47:
as [commented](https://github.com/python-poetry/poetry/issues/1941#issuecomment-631562961) on #1941 , i'm not sure this is the right approach. the currently-documented suggestion allows me to test my application with identical dependencies between local dev, CI, and production. i'm not sure how to do this otherwise

bittner:
@igor47 Sure, this might work for you this way. But there should be _a better way_ to achieve just that. For you and for everyone else.

In a Tox configuration you should specify the dependencies in a _declarative_ way (that's what `deps =` is for). Then, any deployment strategy Tox supports can install just those dependencies. In other words, _Tox_ can choose which application or procedure it applies to ensure that those dependencies are installed.

By suggesting a `foobar install` command we're giving up on that clean separation of concerns, and on the flexibility we gain with that. That's bad. And we should make sure people apply the better strategy. Hence this change.

Secrus:
Hi, @bittner are you still interested in contributing this to Poetry? If so, please describe both options as according to the team, both have their use.

bittner:
> Hi, @bittner are you still interested in contributing this to Poetry? If so, please describe both options as according to the team, both have their use.

I'm not sure I understand. Do you intend, keep the current description and suggest my change only as an alternative way in the docs?

Secrus:
> I'm not sure I understand. Do you intend, keep the current description and suggest my change only as an alternative way in the docs?

Yes, that would be the plan.

finswimmer:
`tox` can be configured in multiple ways. It depends on what should be the code under test and which dependencies should be installed. The docs should include the several use cases.

1.
```
[tox]
isolated_build = true

[testenv]
deps =
    pytest
commands =
    pytest tests/ --import-mode importlib
```

`tox` will create an `sdist` package of the project and uses `pip` to install it in a fresh environment. Thus dependencies are resolved by `pip`.

2.
```
[tox]
isolated_build = true

[testenv]
whitelist_externals = poetry
commands_pre =
    poetry install --no-root --sync
commands =
    poetry run pytest tests/ --import-mode importlib
```

`tox` will create an `sdist` package of the project and uses `pip` to install it in a fresh environment. Thus dependencies are resolved by `pip` in the first place. But afterwards we run Poetry, which will install the locked dependencies into the environment.

3.
```
[tox]
isolated_build = true

[testenv]
skip_install = true
whitelist_externals = poetry
commands_pre =
    poetry install
commands =
    poetry run pytest tests/ --import-mode importlib
```

`tox` will not do any install. Poetry installs all the dependencies and the current package an editable mode. Thus tests are running against the local files and not the builded and installed package.

bittner:
That's all good.

Just to avoid misunderstandings, my point is not what you _can_ do, but what you _should_ do.

Usually, the tool dictates how you should do things. I believe, you should tell Tox what to do and it will do it for you (i.e. no need to write down the installation command with your dependencies; only compile a list of dependencies). If you go against the design of the tool you interfere with some of its features and make changes of the setup more difficult.

As an alternative, if you want to run just sequences of actual commands then write a Makefile or a shell script.

As Poetry is a very popular tool, we have the responsibility to tell developers how to do things the right way. Hence this PR.

finswimmer:
Hey @bittner,

the problem with "telling developers how to do things the right way" here, is that there is not a single way. It depends on the goal.

Your suggested changes describes the  way how to use tox in the same way, a lot of people are doing it, when developing a package with setuptools.

The current description, describes a way how to use locked dependencies. That's something a lot of poetry user wanted to do.

That's why I think it would be a great benefit for the docs to explain the different options and when to use what.

fin swimmer

finswimmer:
Superseded by https://github.com/python-poetry/poetry/pull/6026

PR #2421, opened 2020-05-14
normalize path if it is a case insensitive file system

### LINKED ISSUE #2419: Virtualenv not always found in case insensitive filesystem

- **OS version and name**: Windows 10
- **Poetry version**: 1.0.5
- **Link of a [Gist](https://gist.github.com/) with the contents of your pyproject.toml file**: _irrelevant_

## Issue

Poetry does not consider case sensitivity of the filesystem when computing the hash of the virtual environment name. This makes it to fail to find the environment in case insensitive filesystems depending on the case of letters in the path.

For example, in Windows, having a project named "test" in `C:\test` and accessing it from `c:\test` or `C:\Test` or any other case variation. This cannot be done from command prompt, as it always canonicalize the path, but it can be easily done from scripts (e.g. Jenkinsfile).

The hash is computed in `EnvManager.generate_env_name` in `poetry.utils.env`:

```
>>> from poetry.utils.env import EnvManager
>>> EnvManager.generate_env_name('test', 'C:\\test')
'test-bFaF-DbV'
>>> EnvManager.generate_env_name('test', 'c:\\test')
'test-7_Oi5G5Y'
```

I expected the hash to be computed always the same, when the filesystem is case-insensitive (or at least use `os.path.normcase` or similar, which is does the job in Windows).

Since I cannot control how the path case is in my build environment, the workaround I'm currently using is to set `virtualenvs.in-project = true`.

0xF4D3C0D3:
closes #2419(Virtualenv not always found in case insensitive filesystem)
the util func for checking whether it is a case sentive
or not is from [Steve Cohen](https://stackoverflow.com/a/36580834/7899226)

# Pull Request Check List
Resolves: #2419

I have no idea where to edit for documentation that I made the default behavior of `EnvManager.generate_env_name` normalize case of cwd. If needed, please let me know where to edit thank you.

### DIFF
--- a/tests/utils/test_env.py
+++ b/tests/utils/test_env.py
@@ -1,6 +1,7 @@
 import os
 import shutil
 import sys
+import tempfile
 
 import pytest
 import tomlkit
@@ -831,3 +832,23 @@ def test_env_site_packages_should_raise_an_error_if_no_site_packages(tmp_dir):
 
     with pytest.raises(RuntimeError):
         env.site_packages
+
+
+def test_case_should_be_ignored_if_fs_is_case_sensitive(tmp_dir):
+    with tempfile.NamedTemporaryFile(prefix="TmP") as tmp_file:
+        if not os.path.exists(tmp_file.name.lower()):
+            path_with_lowercase = Path(tmp_dir) / "lowerpath"
+            path_with_uppercase = Path(tmp_dir) / "UPPERPATH"
+            path_with_bothcase = Path(tmp_dir) / "BoThCaSe"
+
+            venv_with_lowercase = EnvManager.generate_env_name(
+                "simple-project", str(path_with_lowercase)
+            )
+            venv_with_uppercase = EnvManager.generate_env_name(
+                "simple-project", str(path_with_uppercase)
+            )
+            venv_with_bothcase = EnvManager.generate_env_name(
+                "simple-project", str(path_with_bothcase)
+            )
+
+            assert venv_with_lowercase != venv_with_uppercase != venv_with_bothcase
--- a/poetry/utils/env.py
+++ b/poetry/utils/env.py
@@ -711,7 +711,8 @@ class EnvManager(object):
     def generate_env_name(cls, name, cwd):  # type: (str, str) -> str
         name = name.lower()
         sanitized_name = re.sub(r'[ $`!*@"\\\r\n\t]', "_", name)[:42]
-        h = hashlib.sha256(encode(cwd)).digest()
+        normalized_cwd = os.path.normcase(cwd)
+        h = hashlib.sha256(encode(normalized_cwd)).digest()
         h = base64.urlsafe_b64encode(h).decode()[:8]
 
         return "{}-{}".format(sanitized_name, h)


0xF4D3C0D3:
I think this is unnecessary. I'm waiting for the original issuer's feedback.

tests/utils/test_env.py
```diff
@@ -831,3 +832,23 @@ def test_env_site_packages_should_raise_an_error_if_no_site_packages(tmp_dir):
...
+                "simple-project", str(path_with_lowercase)
+            )
+            venv_with_uppercase = EnvManager.generate_env_name(
+                "simple-project", str(path_with_uppercase)
+            )
+            venv_with_bothcase = EnvManager.generate_env_name(
+                "simple-project", str(path_with_bothcase)
+            )
+
+            assert venv_with_lowercase != venv_with_uppercase != venv_with_bothcase
```
radoering, line 854: I think, in its current state, the test does not do what it is supposed to do (even if the approach is fine). The test only checks if different pathes result in different venv names. (The pathes are different even if case is ignored!) The test should at least check two pathes that are equal for case insensitive file systems but different for case sensitive file systems.

Further, there are two minor quirks:
1. The assert will be true even if venv_with_lowercase and venv_with_bothcase are equal, e.g.
   ```
   >>> 'a' != 'b' != 'a'
   True
   ```
   Probably, there should be three asserts each comparing a pair of venv names.

2. The name of the test seems a little bit ambiguous to me (the meaning is "test: case should ...", but reads "'test case' should ...") and probably the wrong way around (case should be ignored for case insensitive file systems and not for case sensitive file systems).

All in all, my proposal for improvement:

```suggestion
def test_generate_env_name_ignores_case_for_case_insensitive_fs(tmp_dir):
    path_with_lowercase = Path(tmp_dir) / "cwd"
    path_with_uppercase = Path(tmp_dir) / "CWD"
    path_with_bothcase = Path(tmp_dir) / "CwD"

    venv_with_lowercase = EnvManager.generate_env_name(
        "simple-project", str(path_with_lowercase)
    )
    venv_with_uppercase = EnvManager.generate_env_name(
        "simple-project", str(path_with_uppercase)
    )
    venv_with_bothcase = EnvManager.generate_env_name(
        "simple-project", str(path_with_bothcase)
    )

    with tempfile.NamedTemporaryFile(prefix="TmP") as tmp_file:
        is_case_insensitive_fs = os.path.exists(tmp_file.name.lower())

    if is_case_insensitive_fs:
        assert venv_with_lowercase == venv_with_uppercase
        assert venv_with_lowercase == venv_with_bothcase
        assert venv_with_uppercase == venv_with_bothcase
    else:
        assert venv_with_lowercase != venv_with_uppercase
        assert venv_with_lowercase != venv_with_bothcase
        assert venv_with_uppercase != venv_with_bothcase
```

radoering (changes requested):
Imho, the change is reasonable. I couldn't find any documentation on the naming either. Maybe, there should be a changelog entry that says that the naming of virtualenv folders for Windows users will change. Thus, an existing virtualenv is not used anymore after updating poetry but a new one is created.

At least, the test needs some approvement. See comment on concerning lines.

radoering:
@0xF4D3C0D3: May you consider to incorporate my requested changes and rebase your branch? Otherwise, if you don't mind I would like to create a new PR to make another try at getting this fix merged.

0xF4D3C0D3:
@radoering thanks for your contribution and feedback. However I think I couldn't do what you want soon. it seems to be better to close this issue and open new PR. Good luck!

Issue #1639, opened 2019-11-27
'poetry init' should use existing pyproject.toml file


dekoza:
- [x] I have searched the [issues](https://github.com/sdispater/poetry/issues) of this repo and believe that this is not a duplicate.
- [x] I have searched the [documentation](https://poetry.eustace.io/docs/) and believe that my question is not covered.

## Feature Request

I think that `poetry init` should detect if existing pyproject.toml already contains poetry entries and if not, it should add itself to it instead of exiting with error. I have basic (almost brute-force) method prepared but I want to make it in more elegant way. Anyway, expect PR soon :smile:

finswimmer:
Hello @dekoza ,

sounds like a good idea to me :+1: 

I would really like to see your PR :)

fin swimmer

ttimbers:
I would like to see this feature too! It would be very helpful when paired with Cookiecutter templates.

dekoza:
I'm sorry, my commercial project really eats up a lot of my time lately. If anyone wants to take over, don't hesitate :slightly_smiling_face:

PR #2448, opened 2020-05-21
poetry init uses existing pyproject.toml if possible

finswimmer:
At the moment a `poetry init` will fail, if a `pyproject.toml` already exists.

As this file is becoming more and more widely used to store configuration for other tools like `black` and `isort`, this PR changes the behavior to only skip creating the content for `pyproject.toml` if the file already contains a `tool.poetry` section or a `build-system` section.

closes: #1639

### DIFF
--- a/poetry/console/commands/init.py
+++ b/poetry/console/commands/init.py
@@ -23,9 +23,6 @@ from .command import Command
 from .env_command import EnvCommand
 
 
-# import tomlkit
-
-
 class InitCommand(Command):
     name = "init"
     description = (
--- a/poetry/console/commands/init.py
+++ b/poetry/console/commands/init.py
@@ -74,14 +74,11 @@ The <c1>init</c1> command creates a basic <comment>pyproject.toml</> file in the
             with (Path.cwd() / "pyproject.toml").open() as toml_file:
                 original_toml = tomlkit.loads(toml_file.read())
 
-                try:
-                    if original_toml["tool"]["poetry"]:
-                        self.line(
-                            "<error>A pyproject.toml file with a poetry section already exists.</error>"
-                        )
-                        return 1
-                except KeyError:
-                    pass
+                if original_toml.get("tool", {}).get("poetry"):
+                    self.line(
+                        "<error>A pyproject.toml file with a poetry section already exists.</error>"
+                    )
+                    return 1
 
                 if original_toml.get("build-system"):
                     self.line(
@@ -216,7 +213,7 @@ The <c1>init</c1> command creates a basic <comment>pyproject.toml</> file in the
             dev_dependencies=dev_requirements,
         )
 
-        content = layout_.generate_poetry_content()
+        content = layout_.generate_poetry_content(original_toml)
         if self.io.is_interactive():
             self.line("<info>Generated file</info>")
             self.line("")
@@ -230,10 +227,6 @@ The <c1>init</c1> command creates a basic <comment>pyproject.toml</> file in the
 
         with (Path.cwd() / "pyproject.toml").open("w", encoding="utf-8") as f:
             f.write(content)
-
-            if original_toml:
-                f.write("\n")
-                f.write(tomlkit.dumps(original_toml))
 
     def _determine_requirements(
         self, requires, allow_prereleases=False, source=None
--- a/poetry/console/commands/init.py
+++ b/poetry/console/commands/init.py
@@ -67,16 +67,16 @@ The <c1>init</c1> command creates a basic <comment>pyproject.toml</> file in the
         from poetry.utils._compat import Path
         from poetry.utils.env import SystemEnv
 
-        original_toml = PyProjectTOML(Path.cwd() / "pyproject.toml")
+        pyproject = PyProjectTOML(Path.cwd() / "pyproject.toml")
 
-        if original_toml.file.exists():
-            if original_toml.is_poetry_project():
+        if pyproject.file.exists():
+            if pyproject.is_poetry_project():
                 self.line(
                     "<error>A pyproject.toml file with a poetry section already exists.</error>"
                 )
                 return 1
 
-            if original_toml.data.get("build-system"):
+            if pyproject.data.get("build-system"):
                 self.line(
                     "<error>A pyproject.toml file with a defined build-system already exists.</error>"
                 )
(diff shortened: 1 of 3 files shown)


poetry/console/commands/init.py
```diff
@@ -66,9 +67,26 @@ def handle(self):
...
+                        )
+                        return 1
+                except KeyError:
+                    pass
+
+                if original_toml.get("build-system"):
+                    self.line(
+                        "<error>A pyproject.toml file with a defined build-system already exists.</error>"
+                    )
+                    return 1
```
abn, line 89:
```suggestion
                if original_toml.get("tool", {}).get("poetry) or original_toml.get("build-system"):
                    self.line(
                        "<error>A pyproject.toml file with a poetry section already exists.</error>"
                    )
                    return 1
```

Should we also handle cases where `tool.poetry` exists and `build-system` does not?

poetry/console/commands/init.py
```diff
@@ -202,6 +220,10 @@ def handle(self):
         with (Path.cwd() / "pyproject.toml").open("w", encoding="utf-8") as f:
             f.write(content)
 
+            if original_toml:
```
abn, line 223: Might be a good idea to update `Layout.generate_poetry_content()` to accept base content and pass `original_toml` into it. Also, we should consider appending to the file.

poetry/console/commands/init.py
finswimmer: Nice idea to set the default return value for `get` to an empty dict!

I would prefer to keep the checks for  `tool.poetry` and `build-system` separated to give a proper error message.

> Should we also handle cases where tool.poetry exists and build-system does not?

Shouldn't the code above already handle this? Whenever one of these both is in the `pyproject.toml` it should quit.

poetry/console/commands/init.py
finswimmer: Yes updating `Layout.generate_poetry_content()` sounds like a good idea.

I also thought about appending to the file but decided to prepend because in my opinion the poetry stuff contains the metadata stuff for the project. And this is something I personally would expect at the beginning of the file.

tarkatronic:
Instead of simply raising an error when a different `build-system` section already exists, would it be possible to prompt the user to convert it to use Poetry?

I recently ran into this when [converting a project from setuptools to Poetry](https://github.com/godaddy/tartufo/pull/52). In this case, I already had a `pyproject.toml` set up for a few other tools, and had a `build-system` defined. IIRC there can be cases where having a `pyproject.toml` without a `build-system` produces errors in pip, so I think this may not be an uncommon scenario.

poetry/console/commands/init.py
abn: The issue with that is that we might end up pushing file comments down. In certain cases projects could have copyright/licence headers. imho, it is safer to append. This is somethign we can also tackle later.

abn: @finswimmer let me know your thoughts on this and I can go ahead approve this one.

finswimmer: The point about the comments is a very good one. I guess I will change it later this week.

abn:
@finswimmer can we rebase this please?

poetry/console/commands/init.py
```diff
@@ -67,9 +68,23 @@ def handle(self):
         from poetry.utils._compat import Path
         from poetry.utils.env import SystemEnv
 
+        original_toml = None
+
         if (Path.cwd() / "pyproject.toml").exists():
-            self.line("<error>A pyproject.toml file already exists.</error>")
-            return 1
+            with (Path.cwd() / "pyproject.toml").open() as toml_file:
```
abn, line 74: We can now move this to use `PyProjectTOML` now that core has been updated. We can also deal with that in a later change if required. Let me know what is preferred.

https://github.com/python-poetry/poetry-core/blob/a056c464ebaf0f1638b6b0be97e4be20dc4a678d/poetry/core/pyproject/toml.py#L13

abn:
@finswimmer here is an example for the init command test cases: https://github.com/abn/poetry/commit/e378a88fe51049f2220946b685c4f98bfeb1be5c

abn:
@finswimmer rebase please.

finswimmer:
I'v expected this @abn :) Is done.

abn:
> Instead of simply raising an error when a different `build-system` section already exists, would it be possible to prompt the user to convert it to use Poetry?

@tarkatronic I thinnk that should be a new feature request as the scope of that is a bit larger than what is changed here.

tarkatronic:
> > Instead of simply raising an error when a different `build-system` section already exists, would it be possible to prompt the user to convert it to use Poetry?
> 
> @tarkatronic I thinnk that should be a new feature request as the scope of that is a bit larger than what is changed here.

That is totally reasonable. 😄

poetry/console/commands/init.py
```diff
@@ -67,9 +67,20 @@ def handle(self):
         from poetry.utils._compat import Path
         from poetry.utils.env import SystemEnv
 
-        if (Path.cwd() / "pyproject.toml").exists():
-            self.line("<error>A pyproject.toml file already exists.</error>")
-            return 1
+        original_toml = PyProjectTOML(Path.cwd() / "pyproject.toml")
+
+        if original_toml.file.exists():
+            if original_toml.is_poetry_project():
```
abn, line 73: You can combine this, is project should check if the file exists.

poetry/console/commands/init.py
```diff
@@ -67,9 +67,20 @@ def handle(self):
         from poetry.utils._compat import Path
         from poetry.utils.env import SystemEnv
 
-        if (Path.cwd() / "pyproject.toml").exists():
-            self.line("<error>A pyproject.toml file already exists.</error>")
-            return 1
+        original_toml = PyProjectTOML(Path.cwd() / "pyproject.toml")
```
abn, line 70:
```suggestion
        pyproject = PyProjectTOML(Path.cwd() / "pyproject.toml")
```

abn (changes requested):
Looking good, minor changes.

poetry/console/commands/init.py
finswimmer: Yes, but for the check if a build-system is defined, I have to check if the file exists first anyway. So I found it cleaner to read if I first check if the file exists and then include any test I like in this block.

poetry/console/commands/init.py
finswimmer: As you like.

abn (approved):

PR #2459, opened 2020-05-24
Display better messages on legacy repo HTTP errors

dan-hipschman-od:
# Changes

This change gives a clear message to users when there's an error connecting to a private repo.

Current message:
```
$ export POETRY_HTTP_BASIC_PYPI_LOCAL_PASSWORD="incorrect-password"
$ poetry add private-package

[ValueError]
Could not find a matching version of package private-package
```
New message
```
  RepositoryError

  401 Client Error: Unauthorized for url: https://xyz.jfrog.io/xyz/api/pypi/pip/simple/private-package/
```

Notes:
* This PR will supersede: https://github.com/python-poetry/poetry/pull/1072. This PR is similar but doesn't introduce a new dependency.
* This will cause poetry to error out if it can't connect to a private repo, whereas currently it will ignore it. This seems better. However, there may be people who have an unused private repo configured, to which poetry can't connect. They'll start getting an error message. It should be an easy fix for them though.
* There might also possibly be some private repos that return (non-404) errors for certain endpoints, and it's conceivable that some user actually depends on this. E.g., maybe their repo returns 400 for /pytest, and they depend on that so poetry continues on to try PyPI. This seems like it would be an issue with the repo, but I'm just thinking through all the possible cases.

# Pull Request Check List

- [ ] ~Updated **documentation** for changed code.~ (seems unnecessary)

### DIFF
--- a/poetry/repositories/legacy_repository.py
+++ b/poetry/repositories/legacy_repository.py
@@ -27,6 +27,7 @@ from poetry.utils.patterns import wheel_file_re
 from ..inspection.info import PackageInfo
 from .auth import Auth
 from .exceptions import PackageNotFound
+from .exceptions import RepositoryError
 from .pypi_repository import PyPiRepository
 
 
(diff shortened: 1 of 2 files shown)


dan-hipschman-od:
Does anyone have a few minutes to take a look at this? It's a pretty small change, and fixes an issue that has come up at my company a few times. Thanks so much! :)

PR #2485, opened 2020-06-02
Added a --no flag to accompany the --yes flag

eacoeytaux:
Hello,

I wanted to add in a `--no`/`-n` flag, that works exactly the opposite of the `--yes`/`-y` flag.  My reason for doing this is that I am trying to automate installation of poetry on multiple machines that have special permissions on `.profile`, which means running the installation with the `--yes` flag or just by default returns an error code when trying to modify the PATH variable, even though it does technically install poetry.  As far as I can tell, the only way to specify _not_ to modify the PATH variable is to run get-poetry.py with user input, since piping it in any form makes it automatically try to modify the PATH variable, but this is not an option for me as I need the installation to be fully automated.

Ideally for situations like these, someone like me could run `python get-python.py --no` and not have to worry about user input while still ensuring get-poetry does not try to  modify the PATH variable.

I hope you find this contribution worthwhile.  Thank you!


reeves17:
^bump This would be great to have.

Arzte:
I'd love to see this, I'm trying to automate the install and management of poetry with my zsh plugin manager, [zdharma/zinit](https://github.com/zdharma/zinit) I don't have a use to modify my path at any point (My path is managed by zinit with `atload'PATH+=:$HOME/.poetry/bin'`) so this would allow me to avoid unneeded prompts when poetry is updated.

agore1:
Would love to see this implemented, too!

neersighted:
`get-poetry.py` is now in maintenance mode. Thanks for the PR -- and please submit PRs against `install.python-poetry.org` if the new script doesn't meet your needs.

Issue #2106, opened 2020-02-28
.poetry/lib/poetry/_vendor/py2.7/subprocess32.py:149: RuntimeWarning:


JulianChia:
After installing Poetry with:

`$ curl -sSL https://raw.githubusercontent.com/python-poetry/poetry/master/get-poetry.py | python`

or

`$ curl -sSL https://raw.githubusercontent.com/python-poetry/poetry/master/get-poetry.py | python3`

I see this warning:

```
$ poetry --version
~/.poetry/lib/poetry/_vendor/py2.7/subprocess32.py:149: RuntimeWarning: The _posixsubprocess module is not being used. Child process reliability may suffer if your program uses threads.
  "program uses threads.", RuntimeWarning)
Poetry version 1.0.3
```

```
$ uname -srvmo
Linux 5.3.0-40-generic #32~18.04.1-Ubuntu SMP Mon Feb 3 14:05:59 UTC 2020 x86_64 GNU/Linux
```
Not sure why there is this warning???

babuloseo:
Try using Python3
curl -sSL https://raw.githubusercontent.com/python-poetry/poetry/master/get-poetry.py | python3

JulianChia:
@babuloseo As mentioned above, it gave the warning as mentioned above, except that it is now poetry version 1.0.5:

```
$ poetry --version
~/.poetry/lib/poetry/_vendor/py2.7/subprocess32.py:149: RuntimeWarning: The _posixsubprocess module is not being used. Child process reliability may suffer if your program uses threads.
  "program uses threads.", RuntimeWarning)
Poetry version 1.0.5
```

It is strange that py2.7 is used for a python3.

finswimmer:
Hello,

both things are described in several issues here. But I took this to investigate the `subprocess32` thing a bit more. 

There are a few hits about this, when searching the web. It looks like the `_posixsubprocess` module needs to be compiled during install `subprocess32`. When vendoring the resulting package, and therefor transport it to a different system than it was compiled, seems to be a problem.

Nevertheless poetry seems to work fine despite of this message. So one could ignore it.

If you really want to get rid of it, I've found a workaround: You have to create your own vendor on the target system:

```bash
$ python -m pip install subprocess32 --no-deps --target ~/.poetry/lib/poetry/_vendor/py2.7 --upgrade
```

wohali:
One small edit to the `get-poetry.py` script fixes this:

[...]

PR #2547, opened 2020-06-11
Make get-poetry.py fallback to standard executables

sdispater:
This is the proper version of #2426 after trying to fix it in #2535.

The changes made in #2426 (and #1878) can break systems that want to use Python 2 (via python) with a python3 executable available. This is the setup we have in our own CI for Python 2.7.

This PR fixes this by selecting the first executable available. Note that it will still look for python3 is python does not exist.

Also, the original implementation was no longer writing the shebang to the `poetry` script on Windows which broke the bash shell on Windows

# Pull Request Check List

### DIFF
--- a/get-poetry.py
+++ b/get-poetry.py
@@ -197,8 +197,7 @@ POETRY_LIB = os.path.join(POETRY_HOME, "lib")
 POETRY_LIB_BACKUP = os.path.join(POETRY_HOME, "lib-backup")
 
 
-BIN = """#!/usr/bin/env python
-# -*- coding: utf-8 -*-
+BIN = """# -*- coding: utf-8 -*-
 import glob
 import sys
 import os
@@ -217,7 +216,7 @@ if __name__ == "__main__":
     main()
 """
 
-BAT = u('@echo off\r\npython "{poetry_bin}" %*\r\n')
+BAT = u('@echo off\r\n{python_executable} "{poetry_bin}" %*\r\n')
 
 
 PRE_MESSAGE = """# Welcome to {poetry}!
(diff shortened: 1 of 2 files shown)


ddaspit:
Unfortunately, this doesn't fix issue #2106. For systems that have both Python 2 (/usr/bin/python) and Python 3 (/usr/bin/python3) installed (many Linux distributions), it will still use the Python 2 executable by default.

get-poetry.py
```diff
@@ -589,23 +588,63 @@ def _make_lib(self, version):
...
+                + str(allowed_executables)
+            )
+
+        return fallback
+
     def make_bin(self):
         if not os.path.exists(POETRY_BIN):
             os.mkdir(POETRY_BIN, 0o755)
 
+        python_executable = self._which_python()
```
bersace, line 628: I still don't understand the rationnal behind searching ourself for python executable while we are running one. In which situation will `_which_python` returns something else that `sys.executable` ?

get-poetry.py
```diff
@@ -589,23 +588,63 @@ def _make_lib(self, version):
...
+    def _which_python(self):
+        """Decides which python executable we'll embed in the launcher script."""
+        allowed_executables = ["python", "python3"]
+        if WINDOWS:
+            allowed_executables += ["py.exe -3", "py.exe -2"]
+
+        # \d in regex ensures we can convert to int later
+        version_matcher = re.compile(r"^Python (?P<major>\d+)\.(?P<minor>\d+)\..+$")
+        fallback = None
+        for executable in allowed_executables:
```
cjw296, line 600: @sdispater - the fact that this *prefers* python 2 is really unfortunate :-(
Any chance it could look for the newest python first?

finswimmer: `python` should link to your preferred python version. So it's absolutely understandable that poetry will pick up this one. If your `python` doesn't link to your preferred python version, please look up how you can change it in your distribution.

bersace: I don't buy it. `/usr/bin/python` is python2 for compatibility. Changing this on released distro is wrong.

What's the point of having `python3 get-poetry.py` installing to `python2` ?

bersace: I'm wondering how your comment comply with https://www.python.org/dev/peps/pep-0394 :)

finswimmer: Where it violates PEP394?
> If the python command is installed, it is expected to invoke either the same version of Python as the python3 command or as the python2 command.

cjw296: @finswimmer - `python` is very often the system python, and no, you shouldn't be messing around with that given that doing so can cripple package managers like yum. A common pattern is to, instead, have `python3` be the more recent interpreter and so the choice of ordering here is really unfortunate. I don't see how the experience would be degraded for your described use case were the ordering to be reversed.

This, unfortunately, appears to be part of a continuing trend in poetry to enforce everything working with a thing called `python` that has to already match the exact version of python you need for you project and having to torture your whole operating system and environment to make that happen, rather than the more common, convenient and flexible approach of allowing the python binary to be used by a tool like poetry to be specified explicitly where it can't be sensibly inferred.

sdispater: @cjw296 We already tried defaulting to `python3` if it was available but it was breaking some systems so we reverted the changes.

There is nothing wrong with defaulting to `python` and if it's not the version that you want to use you have the option to use the `env use` command which purpose is exactly that or you create and activate a virtual environment and Poetry will automatically pick it up.

bersace: @sdispater we're talking about the python running poetry itself, not the python running managed project. does `env use` select the interpreter running poetry itself ?

bersace: @sdispater actually, the point is not fallbacking to `python` but simply install poetry for the interpreter running get-poetry. Just like `pip` installs a package for the interpreter running pip.

bersace: So does `get-pip.py` by default.

sdispater: @bersace The point of the official installer is to pick up the currently activate Python executable so you can easily switch between Python versions without having to install Poetry for each version. If we were to to hardcode the executable used to install Poetry this would defeat this purpose.

bersace: By currently activated Python, do you mean pyenv or poetry-managed project python interpreter in a virtualenv ?

cjw296: @sdispater - what kind of breaking were you seeing when looking for `python3` first? 
How would use use the `env use` command when running `get-poetry.py`?

@bersace - I'm afraid @sdispater may well means `pyenv`. It's such an unfortunate choice to recommend as it's a terrible tool in professionally managed environments where `pyenv` either fails completely because development headers aren't available, or it builds broken python binaries because development headers for "optional" modules like `zlib` and `ncurses` aren't present, or perhaps worse still builds its own python binaries rather than using system-provided builds that have been optimised for the current environment by a sysadmin or build team.

Issue #2576, opened 2020-06-17
Private Repositories Fail Silently on Auth Errors


TimOrme:
- [x] I am on the [latest](https://github.com/python-poetry/poetry/releases/latest) Poetry version.
- [x] I have searched the [issues](https://github.com/python-poetry/poetry/issues) of this repo and believe that this is not a duplicate.
- [x] If an exception occurs when executing a command, I executed it again in debug mode (`-vvv` option).

- **OS version and name**: MacOS 10.15.5
- **Poetry version**: 1.0.9
- **Link of a [Gist](https://gist.github.com/) with the contents of your pyproject.toml file**: https://gist.github.com/TimOrme/d69dc1f3ee6ce31cf93d154bd62d0595

## Issue
If using a private repository to install external dependencies, it currently fails silently if that private repository rejects the request due to authentication issues.  It would be good to instead at least log a warning to inform the user that the server didn't accept the request.  Otherwise, it seems to indicate a problem with the packages actually being present.

```bash
 % poetry add mypackage

[ValueError]
Could not find a matching version of package mypackage
```

If `mypackage` is only present in the private repo, but they don't have credentials set properly, it should indicate to the user that there was auth error on that repo.

diogo-gaivota:
How did you know that it was authentication problem? I'm having a similar issue here, but even with -vvv, i hadn't any point on it.

TimOrme:
I actually attached a debugger to discover it.  I also added the PR here https://github.com/python-poetry/poetry/pull/2577 which adds logging for that case, so I was able to verify in that branch that the return status was 403.

diogo-gaivota:
The problem is the same to me. I got an 401 that failed silently.

PR #2577, opened 2020-06-17
Add log warning if auth fails when accessing repository

TimOrme:
Resolves: #2576 

This just adds a more informative error message in the event that a repository fails due to an authorization error.  Right now, poetry fails silently on that case.

I tried to add tests here, but because this is part of the core `_get` method, I couldn't really do so using the existing mock.  I could perform mocking at the `requests` level, but wasn't sure if that was the usual way in the poetry project.  Let me know the best way to test this.

Thanks!
Tim

### DIFF
--- a/poetry/repositories/legacy_repository.py
+++ b/poetry/repositories/legacy_repository.py
@@ -426,4 +426,8 @@ class LegacyRepository(PyPiRepository):
         if response.status_code == 404:
             return
 
+        if response.status_code == 403:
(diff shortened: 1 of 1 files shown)


austinbutler:
Would be nice to also have a 401 check. In my case it was a 401 with `Artifactory configured to accept only encrypted passwords but received a clear text password, getting the encrypted password can be done via the WebUI.`.

TimOrme:
@austinbutler added this as well.  Thanks!

abn (approved):

PR #2582, opened 2020-06-19
Add missing as_posix() calls

msakai:
As PR #2398 [changed package.root_dir from str to Path](https://github.com/python-poetry/poetry/pull/2398/files#diff-d1ad1ffd5c14b8d5a698ea2f12b5c0e7L281) , we need to call `as_posix()` to convert it to `str` before passing to `os.path.join`. Otherwise, it raises `'PosixPath' object has no attribute 'endswith'` error on Python <3.6.

# Pull Request Check List

Resolves: #issue-number-here

### DIFF
--- a/poetry/installation/pip_installer.py
+++ b/poetry/installation/pip_installer.py
@@ -143,7 +143,7 @@ class PipInstaller(BaseInstaller):
 
         if package.source_type in ["file", "directory"]:
             if package.root_dir:
-                req = os.path.join(package.root_dir.as_posix(), package.source_url)
+                req = (package.root_dir / package.source_url).as_posix()
             else:
                 req = os.path.realpath(package.source_url)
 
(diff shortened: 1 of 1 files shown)


poetry/installation/pip_installer.py
```diff
@@ -143,7 +143,7 @@ def requirement(self, package, formatted=False):
 
         if package.source_type in ["file", "directory"]:
             if package.root_dir:
-                req = os.path.join(package.root_dir, package.source_url)
+                req = os.path.join(package.root_dir.as_posix(), package.source_url)
```
abn, line 146:
```suggestion
                req = package.root_dir / package.source_url
```

msakai: Thanks for the suggestion and +1.
I added a commit to avoid using `os.path.join` but still convert the result to `str` as `req = (package.root_dir / package.source_url).as_posix()` because I think `req` is expected to be `str`.

PR #2583, opened 2020-06-19
Update shell.py to support tcsh

JnyJny:
These small changes allow `poetry shell` to properly activate a python virtual environment when the user's shell is `tcsh`.

# Pull Request Check List

Resolves: #2584 


JnyJny:
Also, I am apparently the last tcsh user left on the planet. :so alone:

Issue #2588, opened 2020-06-21
Poetry failed to detect conflicts between package "scikit-learn 0.23.1" and "python2.7"


zbw0046:
- [x] I am on the [latest](https://github.com/python-poetry/poetry/releases/latest) Poetry version.
- [x] I have searched the [issues](https://github.com/python-poetry/poetry/issues) of this repo and believe that this is not a duplicate.
- [x] If an exception occurs when executing a command, I executed it again in debug mode (`-vvv` option).

- **OS version and name**: Debian 9.0
- **Poetry version**: 1.0.9
- **Link of a [Gist](https://gist.github.com/) with the contents of your pyproject.toml file**: https://gist.github.com/zbw0046/d0ee62349a162c2e173329636e61fa13

### Description

I'm using python2.7, and write a `pyproject.toml` file with the requirement `scikit-learn == 0.23.1`. Poetry is expected to raise an error because `scikit-learn` does not support python2 at version `0.23.1`. However, in my poetry it passed conflict checking.

### Step to reproduce
1. Using this toml file: https://gist.github.com/zbw0046/d0ee62349a162c2e173329636e61fa13
2. Run `poetry debug resolve -vvv`
3. You will get: 
```
Resolution results:

future                 0.18.2   python_version ~= "2.7" and sys_platform == "linux2"
numpy                  1.16.5   python_version ~= "2.7" and sys_platform == "linux2"
crcmod                 1.7      python_version ~= "2.7" and sys_platform == "linux2"
ffmpeg-python          0.1.17   python_version ~= "2.7" and sys_platform == "linux2"
opencv-python-headless 3.4.2.17 python_version ~= "2.7" and sys_platform == "linux2"
scikit-learn           0.23.1   python_version == "2.7" and sys_platform == "linux2"
```
### Expected behavior
Poetry should raise an error that `scikit-learn 0.23.1` conflicts with `python 2.7`.

finswimmer:
Hello @zbw0046 ,

you are using a custom repository. In the repository there is only a `sdist` for `scikit-learn`. This `sdist` doesn't provide any information which python version it needs within the `PKG-INFO`, which is what `poetry` relies on. So this is something on `scikit-learn` side. Could you please report it there?

[...]

PR #2589, opened 2020-06-21
[Repositories] consider "data-requires-python" anchor in the link when finding packages

zbw0046:
## Pull Request Check List

Resolves: #2588

## Description
According to [PEP 503](https://www.python.org/dev/peps/pep-0503/), a repository MAY include a `data-requires-python` attribute on a file link. This exposes the `Requires-Python` metadata field, specified in `PEP 345`, for the corresponding release. Where this is present, installer tools SHOULD ignore the download when installing to a Python version that doesn't satisfy the requirement. 

However, in current logic `data-requires-python` in link is not considered in the `_get_release_info` method of `legacyRepository` class, which is the root cause of the #2588 .

This PR adds the checking of `data-requires-python` attribute. If it exists, we set `requires_python` in `release_info` as it says.

### DIFF
--- a/poetry/repositories/legacy_repository.py
+++ b/poetry/repositories/legacy_repository.py
@@ -389,6 +389,7 @@ class LegacyRepository(PyPiRepository):
                 )
             )
         urls = defaultdict(list)
+        url_requires_python_dict = {}
         files = []
         for link in links:
             if link.is_wheel:
@@ -398,6 +399,7 @@ class LegacyRepository(PyPiRepository):
             ):
                 urls["sdist"].append(link.url)
 
+            url_requires_python_dict[link.url] = link.requires_python
             h = link.hash
             if h:
                 h = link.hash_name + ":" + link.hash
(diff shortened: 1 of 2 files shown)


zbw0046:
An example of the link: <a href="/tos/pkg/pypi/scikit-learn/scikit-learn-0.23.1.tar.gz#sha256=e3fec1c8831f8f93ad85581ca29ca1bb88e2da377fb097cf8322aa89c21bc9b8" data-requires-python=">=3.6">scikit-learn-0.23.1.tar.gz</a>

dimbleby:
Per https://github.com/python-poetry/poetry/issues/2588#issuecomment-1172769456, poetry seems to get the right answer today - though I'm not sure that it's through the mechanism that this MR proposes.

If you are still interested in contributing this: the code has moved on quite a bit and so it'll need a rebase.  Also tests would be good...

neersighted:
Thanks for the PR! This is now implemented in Poetry 1.2.

neersighted:
Ah, shoot, I realized that I misread the PR on mobile -- I thought this was for yanked releases. This could potentially let Poetry do less work if we could rule out dists before downloading them. As such, I'm going to re-open it to discuss, even if it would need a lot of cleanup for merge into current Poetry.

PR #2595, opened 2020-06-26
New and faster installer implementation

sdispater:
## Pull Request Check List

This PR introduces a new installation experience with a brand new installer that supports parallel operations. This was made possible by splitting the different parts of the installation process into several new classes:

- The `Executor` class is now the main point of entry and will be used by the existing `Installer` class if opted in via the `use_executor` method (note that the new installer will be opt-out by default since I want to gather as much feedback as possible). The `Executor` is responsible for executing a batch of operations (install, uninstall , update), in parallel by default. Note that not all operations will be executed in parallel but only those with the same depth in the dependency tree.
- The `Chooser` class is responsible for picking the appropriate distributions (sdist or wheel) for a given package and environment.
- The `Chef` class does not do much at the moment, it only manages the distributions cache, but will eventually be responsible for building wheels before package installation, this will be implemented in the `1.2` release.
- The `Authenticator` class is reponsible for handling credentials when downloading distributions from custom indices. Most of the ideas have been taken from the cache system of `pip`.

This is the first step towards the removal of calls to `pip` for installation of packages. We now use it exclusively to install distributions available locally.

The installer can be opted out via the `experimental.new-installer` setting:

```
poetry config experimental.new-installer false
```

Here is an example of the new installer output:

Screen Recording 2020-05-29 at 17 39 38


poetry/installation/authenticator.py
```diff
@@ -0,0 +1,133 @@
...
+                    # using the password attribute of the return value)
+                    credentials = auth.split(":", 1)
+                else:
+                    credentials = auth, None
+
+                credentials = tuple(
+                    None if x is None else urlparse.unquote(x) for x in credentials
+                )
+
+        if credentials[0] is not None or credentials[0] is not None:
```
MichaelAquilina, line 101: seems like this is just the same thing twice? I suspect the second one is meant to be `credentials[1]`?

sdispater: Nice catch!

I just fixed it and added tests accordingly.

poetry/console/commands/install.py
```diff
@@ -85,15 +85,32 @@ def handle(self):
             # If this is a true error it will be picked up later by build anyway.
             return 0
 
-        self.line(
-            "  - Installing <c1>{}</c1> (<c2>{}</c2>)".format(
-                self.poetry.package.pretty_name, self.poetry.package.pretty_version
+        self.line("")
+        if not self._io.supports_ansi() or self.io.is_debug():
+            self.line(
+                "<b>Installing</> the current project: <c1>{}</c1> (<c2>{}</c2>)".format(
```
MichaelAquilina, line 91: might be a syntax I am not aware of, but shouldnt this be closed with `</b>`? (similar issue below)

poetry/console/config/application_config.py
```diff
@@ -104,6 +115,35 @@ def set_env(self, event, event_name, _):  # type: (PreHandleEvent, str, Any) ->
...
+        command = event.command.config.handler  # type: WithInstaller
+        if not isinstance(command, WithInstaller):
+            return
+
+        if command.installer is not None:
+            return
+
+        if not isinstance(command, EnvCommand):
+            raise RuntimeError(
+                "A command that needs an installer must also inherit from EnvCommand"
```
abn, line 130: Since this is the case, wouldn't it be better if we have a new class `InstallerCommand(EnvCommand, WithInstaller)`? Additionally, can we consider removing the `WithInstaller` mixin since it might be better to have a child class that has an `installer` attribute. The mixin simply is adding an unnecessary abstraction here at the moment.

Furthermore, we could also consider doing away with this method in favour of `installer` being initialised within `InstallerCommand.__init__` as we should have all the values required at that point.

poetry/console/commands/mixins/with_installer.py
```diff
@@ -0,0 +1,17 @@
...
+class WithInstaller(object):
+
+    _installer = None
+
+    @property
+    def installer(self):  # type: () -> Installer
+        return self._installer
+
+    def set_installer(self, installer):  # type: (Installer) -> None
+        self._installer = installer
```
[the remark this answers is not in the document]
abn: See comment in `application_config.py`, we might be better off with a new `InstallerCommand(EnvCommand)` class instead of a mixin.

poetry/installation/authenticator.py
```diff
@@ -0,0 +1,133 @@
...
+        from requests.auth import HTTPBasicAuth
+
+        request = Request(method, url)
+
+        username, password = self._get_credentials_for_url(url)
+
+        if username is not None and password is not None:
+            request = HTTPBasicAuth(username, password)(request)
+
+        session = self.session
```
abn, line 48:
```suggestion
```

poetry/installation/authenticator.py
```diff
@@ -0,0 +1,133 @@
...
+        prepared_request = session.prepare_request(request)
```
abn, line 49:
```suggestion
        prepared_request = self.session.prepare_request(request)
```

poetry/installation/authenticator.py
```diff
@@ -0,0 +1,133 @@
...
+
+        proxies = kwargs.get("proxies", {})
+        stream = kwargs.get("stream")
+        verify = kwargs.get("verify")
+        cert = kwargs.get("cert")
+
+        settings = session.merge_environment_settings(
```
abn, line 56:
```suggestion
        settings = self.session.merge_environment_settings(
```

poetry/installation/authenticator.py
```diff
@@ -0,0 +1,133 @@
...
+            prepared_request.url, proxies, stream, verify, cert
+        )
+
+        # Send the request.
+        send_kwargs = {
+            "timeout": kwargs.get("timeout"),
+            "allow_redirects": kwargs.get("allow_redirects", True),
+        }
+        send_kwargs.update(settings)
+        resp = session.send(prepared_request, **send_kwargs)
```
abn, line 66:
```suggestion
        resp = self.session.send(prepared_request, **send_kwargs)
```

poetry/installation/authenticator.py
```diff
@@ -0,0 +1,133 @@
...
+        session = self.session
+        prepared_request = session.prepare_request(request)
+
+        proxies = kwargs.get("proxies", {})
+        stream = kwargs.get("stream")
+        verify = kwargs.get("verify")
+        cert = kwargs.get("cert")
+
+        settings = session.merge_environment_settings(
+            prepared_request.url, proxies, stream, verify, cert
```
abn, line 57:
```suggestion
            url=prepared_request.url, 
            proxies=kwargs.get("proxies", {}), 
            stream=kwargs.get("stream"),
            verify=kwargs.get("verify"),
            cert=kwargs.get("cert"),
```

Nit-picky, but i feel something like this is more readable

poetry/installation/authenticator.py
```diff
@@ -0,0 +1,133 @@
...
+            request = HTTPBasicAuth(username, password)(request)
+
+        session = self.session
+        prepared_request = session.prepare_request(request)
+
+        proxies = kwargs.get("proxies", {})
+        stream = kwargs.get("stream")
+        verify = kwargs.get("verify")
+        cert = kwargs.get("cert")
+
```
abn, line 55:
```suggestion
```
See below.

poetry/installation/authenticator.py
```diff
@@ -0,0 +1,133 @@
...
+            "allow_redirects": kwargs.get("allow_redirects", True),
+        }
+        send_kwargs.update(settings)
+        resp = session.send(prepared_request, **send_kwargs)
+
+        resp.raise_for_status()
+
+        return resp
+
+    def _get_credentials_for_url(
```
abn, line 72: Should the following methods move to a more general use location like utils? Might not be required if we intend to only use creds here. We can also defer this decision.

poetry/installation/authenticator.py
```diff
@@ -0,0 +1,133 @@
...
+if TYPE_CHECKING:
+    from typing import Any
+    from typing import Optional
+    from typing import Tuple
+
+    from clikit.api.io import IO
+    from requests import Request  # noqa
+    from requests import Response  # noqa
+    from requests import Session  # noqa
+
```
abn, line 16: Any reason why we defer the runtime imports for requests?

poetry/installation/authenticator.py
```diff
@@ -0,0 +1,133 @@
...
+    def __init__(self, config, io):  # type: (Config, IO) -> None
+        self._config = config
+        self._io = io
+        self._session = None
+        self._credentials = {}
+        self._password_manager = PasswordManager(self._config)
+
+    @property
+    def session(self):  # type: () -> Session
+        from requests import Session  # noqa
```
abn, line 30: Would be great to avoid these if possible - both lazy initialisatio and deferred imports. Unless this provides a significant improvement.

poetry/repositories/installed_repository.py
```diff
@@ -59,8 +61,15 @@ def load(cls, env):  # type: (Env) -> InstalledRepository
...
-                            package.source_url = directory.as_posix()
+
+                            is_directory = (
+                                os.path.exists(str(directory))
+                                or directory.is_absolute()
+                            )
+
+                            if is_directory:
+                                package.source_type = "directory"
+                                package.source_url = directory.as_posix()
```
abn, line 72:
```suggestion
                            if directory.is_dir():
                                package.source_type = "directory"
                                package.source_url = directory.as_posix()
```

poetry/utils/_compat.py
```diff
@@ -23,6 +23,11 @@
 except ImportError:
     import urlparse
 
+try:
+    from os import cpu_count
+except ImportError:
```
abn, line 28:
```suggestion
except ImportError:  # Python 2
```
Jus so that we can clean this up later

.github/workflows/main.yml
```diff
@@ -49,7 +49,7 @@ jobs:
         shell: bash
         run: |
           curl -fsS -o get-poetry.py https://raw.githubusercontent.com/python-poetry/poetry/master/get-poetry.py
-          python get-poetry.py -y
+          python get-poetry.py -y --preview
```
abn, line 52: We had reverted this earlier right?

poetry/installation/chooser.py
```diff
@@ -0,0 +1,180 @@
...
+from packaging.tags import Tag
+from poetry.core.packages.package import Package
+from poetry.core.packages.utils.link import Link
+from poetry.repositories.pool import Pool
+from poetry.utils.env import Env
+from poetry.utils.patterns import wheel_file_re
+
+
+class InvalidWheelName(Exception):
+
```
abn, line 15:
```suggestion
```

poetry/installation/chooser.py
```diff
@@ -0,0 +1,180 @@
...
+    pass
+
+
+class Wheel(object):
+    def __init__(self, filename):  # type: (str) -> None
+        wheel_info = wheel_file_re.match(filename)
+        if not wheel_info:
+            raise InvalidWheelName("{} is not a valid wheel filename.".format(filename))
```
abn, line 23: Do we expect this to be valid for local wheels as well? I am guessing right now we only expect this code path to be traversed only if we are installing from an index?

poetry/installation/executor.py
```diff
@@ -0,0 +1,587 @@
...
+        self._lock.acquire()
+        if executed:
+            self._executed_operations += 1
+            self._executed[operation.job_type] += 1
+        else:
+            self._skipped[operation.job_type] += 1
+
+        self._lock.release()
+
+    def run(self, *args, **kwargs):  # type: (...) -> str
```
abn, line 224:
```suggestion
    def pip(self, *args, **kwargs):  # type: (...) -> str
```

poetry/installation/chooser.py
```diff
@@ -0,0 +1,155 @@
...
+
+from poetry.core.packages.package import Package
+from poetry.core.packages.utils.link import Link
+from poetry.repositories.legacy_repository import LegacyRepository
+from poetry.repositories.pypi_repository import PyPiRepository
+from poetry.utils.env import Env
+from poetry.utils.patterns import wheel_file_re
+
+
+class Wheel(object):
```
pradyunsg, line 19: @agronholm Yet another place a `Wheel` class has been invented. :)

poetry/installation/executor.py
```diff
@@ -0,0 +1,416 @@
...
+        has_build_system = False
+        if pyproject.exists():
+            pyproject_content = pyproject.read()
+            has_poetry = (
+                "tool" in pyproject_content and "poetry" in pyproject_content["tool"]
+            )
+            # Even if there is a build system specified
+            # pip as of right now does not support it fully
+            # TODO: Check for pip version when proper PEP-517 support lands
+            # has_build_system = ("build-system" in pyproject_content)
```
pradyunsg: I'm not sure I follow -- could you elaborate on what this means?

poetry/installation/chooser.py
agronholm: This one looks like it still supports Python 2, so the new wheel API will not help here.

poetry/console/commands/install.py
sdispater: It's not a syntax error, no, and it's supported by cleo/clikit: an empty closing tag will close any open tag.

poetry/console/config/application_config.py
sdispater: Having an `InstallerCommand` class is actually a good idea.

When you say

> we could also consider doing away with this method

which method are you referring to.

poetry/installation/authenticator.py
sdispater: I don't foresee this to be used elsewhere at the moment. If that were to be the case we can always refactor later.

poetry/installation/authenticator.py
sdispater: Actually, I plan on making more imports lazy.

The reasoning behind this is that imports are costly for CLI tools and we should defer them as much as possible.

.github/workflows/main.yml
sdispater: Yes, but with the change of lock file format in #2361 we need to use the preview version for now. This will be reverted back once the stable version of `1.1.0` is released.

poetry/installation/chooser.py
sdispater: The wheel files naming is standardized so it should be valid for local wheels too.

poetry/installation/executor.py
sdispater: This is an old comment that is no longer true and the appropriate changes have already be done on Poetry's side

davidszotten:
hi. excited to try this out! 

when i tried it i get a `KeyError: interpreter_name` in `poetry/installation/chef.py:97`  (`key_parts["interpreter_name"] = self._env.marker_env["interpreter_name"]`)

i wonder if it's because we have `virtualenvs.create false` and `self._env == SystemEnv("/usr/local")`

poetry/console/config/application_config.py
abn: The `set_installer` methods. Right now we seem to use this in order to defer the installer creation until the required variables are populated. With the introduction of an `InstallerCommand` class we can make this more seamless.

poetry/installation/authenticator.py
sdispater: I'd prefer to leave it as-is since it avoids the cost, albeit small, of the property access.

poetry/console/config/application_config.py
sdispater: We can't instantiate the installer in `__init__` since at this point (the console application creation) we don't have all the information.

poetry/console/commands/mixins/with_installer.py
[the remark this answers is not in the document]
sdispater: I added the `InstallerCommand` :-)

poetry/installation/executor.py
sdispater: Renamed to `run_pip()`

poetry/installation/chooser.py
sdispater: Done :-)

poetry/utils/_compat.py
sdispater: Done :-)

sdispater:
@davidszotten Thanks for reporting this!

There were actually missing elements for `SystemEnv`. It should be fixed now.

sdispater:
@abn Your various points have been addressed and/or fixed.

davidszotten:
thanks, this seems to have fixed the `KeyError`

next q: from the output it _looks like_ it grabs 5 packages at a time, and downloads and installs them in parallel. But it looks like waits until all 5 are installed before grabbing another set of 5.  Am i misreading the output? if not, is this intentional? (as opposed to as soon as the first one finishes, grabbing another, to always have 5 on the go)

davidszotten:
also:
all devs in our team aren't on the latest poetry version yet. i've noticed that newer versions are adding `type = "legacy"` to the lock file (we use a custom devpi to serve an internal package index)

latest poetry still happily installs even without `type=legacy` entries, but the new installer seems to ignore our `tool.poetry.source` and fail to find any links when this marker is missing.  this may be intentional but i thought worth flagging

davidszotten:
finally got it running end-to-end. some numbers might be of interest (slow laptop + ~220 deps)

before:
`poetry install`: 6m45s
`poetry export + pip install -r`:  5s + 3m59s

after
`poetry install`: 4m1s

poetry/installation/executor.py
```diff
@@ -0,0 +1,587 @@
...
+
+        if operations and (self._enabled or self._dry_run):
+            self._display_summary(operations)
+
+        # We group operations by priority
+        groups = itertools.groupby(operations, key=lambda o: -o.priority)
+        i = 0
+        self._sections = OrderedDict()
+        for _, group in groups:
+            for chunk in chunked(group, self._max_workers):
```
davidszotten, line 115: i may be missing something, but if we move the initial output from here into the task itself, we could just create tasks for all operations at once and let the executor's max_workers limit the number of concurrent tasks

that way a single slow package install wan't stall the whole pipeline but let the other threads continue installing, while still keeping a limit on the total number of concurrent ops

something like

```
        for _, group in groups:
            tasks = [
                # _execute_operation includes writing the `pending` messages
                self._executor.submit(self._execute_operation, operation)
                for operation in group
            ]
            concurrent.futures.wait(tasks)
````

**Comment by sdispater** [comment]
@davidszotten The order of installation always was by depth in the dependency tree/graph. So this new installer installs a batch of same-depth packages before moving to another batch.

The reason for that is that a limited number of packages requires their runtime dependencies to be installed before they can be built (they are setuptools based packages).

This is not set in stone though, so at some point we may lift this restriction.

**Comment by davidszotten** [comment]
again, apologies if i'm misunderstanding, but i thought the dependency graph depts was handled by the grouping (`groups = itertools.groupby(operations, key=lambda o: -o.priority)`) . i'm not suggesting running groups in parallel, i'm just suggesting removing the chunking that we also have within a single group (`for chunk in chunked(group, self._max_workers):`)

File: poetry/console/commands/install.py (hunk 2 above)
Comment (davidszotten) [reply poetry/console/commands/install.py:90]: how come you only use it for `<b>` and not e.g. the following `<c1>`?

**Comment by MichaelAquilina** [comment]
@sdispater I keep getting the following error trying to setup poetry locally on this branch (latest commit id `285901629f434a056e9bf0a2bcb0d151df61876c`):

```
$ git clone git@github.com:python-poetry/poetry.git
$ git checkout new-installer
$ cd poetry
$ poetry install

[RecursionError]
maximum recursion depth exceeded
```

Everything works fine on the master branch so I suspect this is a regression.

**Comment by sdispater** [comment]
@MichaelAquilina There was a change in the lock file format with the `1.1.0a2` release which this PR uses so you can't use it with a previous version of Poetry.

**Comment by MichaelAquilina** [comment]
> @MichaelAquilina There was a change in the lock file format with the 1.1.0a2 release which this PR uses so you can't use it with a previous version of Poetry.

Right makes sense!

I've got it working as follows for anyone else who is stuck:

```
$ git checkout 1.1.0a2
$ poetry install
$ git checkout new-installer
$ poetry shell
$ poetry install
```

Let me know if any part of that is wrong @sdispater so that I can update my comment

**Comment by davidszotten** [comment]
i had a similar issue but got stuck trying to understand why `-vvv` wasn't working correctly for `poetry install`. is it possible that this is also broken by this pr?

**Comment by sdispater** [comment]
@davidszotten What is different about the debug mode (`-vvv`)?

**Comment by jessekrubin** [comment]
Just curious, does this use some sort o async-eventloop?

**Comment by davidszotten** [comment]
> @davidszotten What is different about the debug mode (`-vvv`)?

hm, now that i'm using @MichaelAquilina 's installation method i can't reproduce it, so maybe it was something else going wrong. apologies for the noise

**Comment by sdispater** [comment]
I added a commit to gracefully handle errors and stop further operations to be executed.

It looks like this:

Screenshot 2020-07-09 at 15 06 10

File: poetry/installation/executor.py [hunk 23]
```diff
@@ -109,49 +101,54 @@ def execute(self, operations):
 
         # We group operations by priority
         groups = itertools.groupby(operations, key=lambda o: -o.priority)
-        i = 0
         self._sections = OrderedDict()
         for _, group in groups:
-            for chunk in chunked(group, self._max_workers):
```
davidszotten: yay!

poetry/installation/executor.py
```diff
@@ -109,49 +101,54 @@ def execute(self, operations):
...
-                                    "  <fg=default;options=bold,dark>•</> {message}: "
-                                    "<fg=default;options=bold,dark>Skipped</> "
-                                    "<fg=default;options=dark>for the following reason:</> "
-                                    "<fg=default;options=bold,dark>{reason}</>".format(
-                                        message=self.get_operation_message(operation),
-                                        reason=operation.skip_reason,
-                                    )
+                            self._lock.acquire()
+                            self._sections[id(operation)] = self._io.section()
+                            self._sections[id(operation)].write_line(
```
davidszotten, line 116: i still think the `Pending` messages need to move into the task executor since (unless i'm misreading) otherwise _every_ package gets marked as pending (which while true fills the screen pushing the stuff actually being installed off-screen)

maybe we don't need the `Pending` markers at all?

sdispater: It's changed so that the pending message is only displayed in the `_do_execute_operation()` method.

sdispater:
@davidszotten I fixed the Ctrl+C behavior if you want to give it a shot.

davidszotten:
> @davidszotten I fixed the Ctrl+C behavior if you want to give it a shot.

thanks! no stacktraces, though it seems to only cancel the running tasks and continues (starts) installing. see inline comment for a suggestion

poetry/installation/executor.py
```diff
@@ -0,0 +1,653 @@
...
+                tasks.append(self._executor.submit(self._execute_operation, operation))
+
+            try:
+                for task in as_completed(tasks):
+                    task.result()
+            except KeyboardInterrupt:
+                self._shutdown = True
+
+            if self._shutdown:
+                self._executor.shutdown(wait=True)
```
davidszotten, line 123: maybe we need something like `[task.cancel() for task in tasks]` before calling shutdown?

poetry/installation/executor.py
```diff
@@ -0,0 +1,653 @@
...
+        for _, group in groups:
+            tasks = []
+            for operation in group:
+                if self._shutdown:
+                    break
+
+                tasks.append(self._executor.submit(self._execute_operation, operation))
+
+            try:
+                for task in as_completed(tasks):
```
davidszotten, line 117: how come you prefer a loop over [`concurrent.futures.wait()`](https://docs.python.org/3/library/concurrent.futures.html#concurrent.futures.wait)

poetry/installation/executor.py
```diff
@@ -0,0 +1,653 @@
...
+        if operations and (self._enabled or self._dry_run):
+            self._display_summary(operations)
+
+        # We group operations by priority
+        groups = itertools.groupby(operations, key=lambda o: -o.priority)
+        self._sections = OrderedDict()
+        for _, group in groups:
+            tasks = []
+            for operation in group:
+                if self._shutdown:
```
davidszotten, line 111: should this check not be outside the `for op in group` loop?

poetry/installation/executor.py
```diff
@@ -0,0 +1,653 @@
...
+                    break
+
+                tasks.append(self._executor.submit(self._execute_operation, operation))
+
+            try:
+                for task in as_completed(tasks):
+                    task.result()
+            except KeyboardInterrupt:
+                self._shutdown = True
```
davidszotten, line 120: could we not just (cancel tasks and) shutdown the executor and just return here? if i'm reading correctly nothing else happens if `shutdown` becomes `True` anyway

davidszotten:
hm, something else has gone wrong in 35225a7. before, the status message (Pending, Downloading, Installing) would disappear after each package finished installing. Now it remains next to each package (so i can no longer tell what is still installing from what is done)

poetry/installation/executor.py
sdispater: I changed it to `wait()` per your suggestion :-)

poetry/installation/executor.py
sdispater: Done :-)

poetry/installation/executor.py
sdispater: It's just to have the same code path whether we have a KeyboardInterrupt or an actual error, since an error will also set `_shutdown`.

sdispater:
@davidszotten Yes, I noticed that as well but it should be fixed now.

poetry/repositories/legacy_repository.py
```diff
@@ -304,6 +304,13 @@ def package(self, name, version, extras=None):  # type: (...) -> Package
 
             return package
 
+    def find_links_for_package(self, package):
```
MichaelAquilina, line 307: minor, but you missed the type annotation here :) have you considered turning on `disallow_untyped_defs`?

poetry/installation/executor.py
sdispater: It needs to be there otherwise we might submit tasks even though the executor is about to shutdown.

poetry/installation/executor.py
davidszotten: ah, missed that the tasks can also set `_shutdown`. thanks

poetry/installation/executor.py
davidszotten: ah, missed that the tasks can also set `_shutdown`. thanks

sdispater:
Thanks to everyone who reviewed this and especially @davidszotten for testing it thoroughly.

I am pretty confident that this is in a stable enough state to be released in the next alpha release of the `1.1.0` version so I'll go ahead and merge it.

If further bugs and issues are discovered we will address them in the next alpha (which should be the last) and beta releases.

davidszotten:
🎉 
thanks so much for your work on poetry!

poetry/installation/executor.py
davidszotten: done?

sdispater: Not done actually 😅 I think I reverted the change inadvertently. I'll make a new PR to add it.

danihodovic:
@sdispater is it possible to track what makes an installation slower?

I'd like to debug the below install time to see if any package is causing problems.
```
$ poetry add watchdog@latest -D
Using version ^0.10.3 for watchdog
Updating dependencies
Resolving dependencies... (959.6s)
```

abn:
@danihodovic I am unable to reproduce that in a new project.

```console
$ poetry@1.1.0rc1 add watchdog 
Using version ^0.10.3 for watchdog

Updating dependencies

Writing lock file

  • Installing pathtools (0.1.2)
  • Installing watchdog (0.10.3)
```

Just to make sure we are not missing anything, what version of poetry are you using? If you are using `1.0.10` and the lock file was generated by a `1.1` release, resolving will not complete. Alternatively, it could be a reslult of some other dependencies in your `pyproject.toml` file. If the issue persists, please raise an issue, this PR is closed.

danihodovic:
I'm not interested in opening a new bug on the slow dependency resolution as there are plenty open already. I want to know how to debug it when the dependency resolution takes an unreasonable amount of time.

abn:
Use `poetry add -D -vvv watchdog@latest`.

Vozf:
I've got some problems in gitlab ci using new poetry parallel installation.
the `poetry install` fails with
```
Retrying HTTP request in 0.5 seconds.
Retrying HTTP request in 0.5 seconds.

  ConnectionError

  ('Connection aborted.', ConnectionResetError(104, 'Connection reset by peer'))

  at ~/.poetry/lib/poetry/_vendor/py3.8/requests/adapters.py:498 in send
      494│                     low_conn.close()
      495│                     raise
      496│ 
      497│         except (ProtocolError, socket.error) as err:
    → 498│             raise ConnectionError(err, request=request)
      499│ 
      500│         except MaxRetryError as e:
      501│             if isinstance(e.reason, ConnectTimeoutError):
      502│                 # TODO: Remove this in 3.0.0: see #2811


  ConnectionError

  ('Connection aborted.', ConnectionResetError(104, 'Connection reset by peer'))

  at ~/.poetry/lib/poetry/_vendor/py3.8/requests/adapters.py:498 in send
      494│                     low_conn.close()
      495│                     raise
      496│ 
      497│         except (ProtocolError, socket.error) as err:
    → 498│             raise ConnectionError(err, request=request)
      499│ 
      500│         except MaxRetryError as e:
      501│             if isinstance(e.reason, ConnectTimeoutError):
      502│                 # TODO: Remove this in 3.0.0: see #2811
```
I don't even know which package causes this. but it can be fixed with `poetry config experimental.new-installer false`

earshinov:
We also face occasional errors when using new version of Poetry:

```
Skipping virtualenv creation, as specified in config file.
Package operations: 119 installs, 13 updates, 0 removals
  • Installing colorama (0.4.1)
  • Installing markupsafe (1.1.1)
  • Updating six (1.11.0 /usr/lib/python3/dist-packages -> 1.11.0)
  EnvCommandError
  Command ['/usr/bin/python3.8', '-m', 'pip', 'install', '--no-deps', '/root/.cache/pypoetry/artifacts/cf/46/78/4a92e28748cc243c738fb10b6dcaae48ea8ddc2684dfff839e0ce9f6c4/MarkupSafe-1.1.1-cp38-cp38-manylinux1_x86_64.whl'] errored with the following return code 1, and output: 
  Processing /root/.cache/pypoetry/artifacts/cf/46/78/4a92e28748cc243c738fb10b6dcaae48ea8ddc2684dfff839e0ce9f6c4/MarkupSafe-1.1.1-cp38-cp38-manylinux1_x86_64.whl
  Installing collected packages: MarkupSafe
  ERROR: Could not install packages due to an EnvironmentError: [Errno 2] No such file or directory: '/usr/local/lib/python3.8/dist-packages/~ix-1.15.0.dist-info'
  
  
  at /usr/local/lib/python3.8/dist-packages/poetry/utils/env.py:948 in _run
       944│                 output = subprocess.check_output(
       945│                     cmd, stderr=subprocess.STDOUT, **kwargs
       946│                 )
       947│         except CalledProcessError as e:
    →  948│             raise EnvCommandError(e, input=input_)
       949│ 
       950│         return decode(output)
       951│ 
       952│     def execute(self, bin, *args, **kwargs):
```

For now we decided to run `poetry install` in CI in a loop like this:

```
RUN poetry_succeeded=;                 \
    for i in $(seq 3);                 \
    do                                 \
        if poetry install --no-root;   \
        then                           \
            poetry_succeeded=true;     \
            break;                     \
        fi;                            \
    done;                              \
    if [ -z "$poetry_succeeded" ];     \
    then                               \
        exit 1;                        \
    fi;                                \
```

abn:
@earshinov you can disable the new installer using `poetry config experimental.new-installer false` for now.

abn:
If there are issues with the installer, please first search [existing issues](https://github.com/python-poetry/poetry/issues). If an issue does not exist, please create one with details on how to reproduce it, `-vvv` logs and if possible a `pyproject.toml` file to use.

For issues related to the new installer, if a fix has not already been made for your issue, you may want to disable the new installer like so until the issue has been resolved.

```sh
poetry config experimental.new-installer false
```

Issue #918, opened 2019-02-26
tox - InvocationError - ModuleNotFoundError: No module named 'poetry'


darrenleeweber:
This issue is different from #881 

## versions
```
$ cat /etc/lsb-release 
DISTRIB_ID=Ubuntu
DISTRIB_RELEASE=18.04
DISTRIB_CODENAME=bionic
DISTRIB_DESCRIPTION="Ubuntu 18.04.2 LTS"

$ conda --version
conda 4.6.7

$ which python
/home/joe/.cache/pypoetry/virtualenvs/org-style-py3.7/bin/python

$ poetry --version
Poetry 0.12.11

$ pip --version
pip 18.1 from /home/joe/.cache/pypoetry/virtualenvs/org-style-py3.7/lib/python3.7/site-packages/pip (python 3.7)

$ tox --version
3.7.0 imported from /home/joe/.cache/pypoetry/virtualenvs/org-style-py3.7/lib/python3.7/site-packages/tox/__init__.py

```

## pyproject.toml

```
[tool.poetry]
name = "org_style"
version = "0.1.1"
description = "Awesome Style Project"
authors = ["Joe <joe@example.com>"]

repository = "https://gitlab.com/org/org_style.git"
homepage = "https://gitlab.com/org/org_style.git"

readme = 'README.md'  # Markdown files are supported

include = ["org/**/*"]
exclude = ["tests/**/*"]

packages = [
    {include = "org"}
]

[tool.poetry.dependencies]
python = "^3.6"
matplotlib = "*"
numpy = "*"
pillow = "*"

[tool.poetry.dev-dependencies]
autopep8 = "*"
codecov = "*"
flake8 = "*"
flake8-type-annotations = "*"
ipdb = "*"
ipython = "*"
invoke-release = "*"
moto = "*"
mypy = "*"
pylint = "*"
pytest = "*"
pytest-benchmark = "*"
pytest-cov = "*"
pytest-datadir = "*"
pytest-mock = "*"
pytest-runner = "*"
pytest-voluptuous = "*"
pytest-xdist = "*"
readme-renderer = { version = "*", extras = ["md"] }
setuptools = "*"
sphinx = "*"
tox = "*"
twine = "*"
wheel = "^0.31.1"  # pinned for invoke-release
yapf = "*"

[build-system]
requires = ["poetry>=0.12"]
build-backend = "poetry.masonry.api"
```

## tox.ini

From https://poetry.eustace.io/docs/faq/#is-tox-supported

[...]

PR #2598, opened 2020-06-27
Whitelisting externals for an isolated tox build hardly makes any sense. 

akarsh1995:
# Pull Request Check List

Resolves: #918

Minor update

Problem: 
- While running tox in azure-pipelines I bumped into this Invocation error.

Solution:
- Specifying the absolute isolated poetry path in the `build-system` tox environment namely ".package".
- Refer: [Default isolated build env dir @ tox](https://tox.readthedocs.io/en/latest/config.html#conf-isolated_build_env)

### DIFF
--- a/docs/docs/faq.md
+++ b/docs/docs/faq.md
@@ -54,10 +54,9 @@ isolated_build = true
 envlist = py27, py36
 
 [testenv]
-whitelist_externals = poetry
 commands =
-    poetry install -v
-    poetry run pytest tests/
+    {toxworkdir}/.package/bin/poetry install -v
+    {toxworkdir}/.package/bin/poetry run pytest tests/
 ```
 
 ## I don't want Poetry to manage my virtual environments. Can I disable it?


abn:
@akarsh1995 this is not compatible with upcoming changes. The reason why your configuration changes work is because today is that we specify the PEP 517 `build-system` to required `poetry`. A side-effect is that within the `.package` environment you have the `poetry` command available. With newer projects; we will be using `poetry-core`, which will not include the cli. The `build-system` section of `poetry` managed projects will change as follows.

```diff
 [build-system]
-requires = ["poetry>=1.0"]
-build-backend = "poetry.masonry.api"
+requires = ["poetry-core>=1.0.0a3"]
+build-backend = "poetry.core.masonry.api"
 ```

You can work around this by using a confgiuration like this:
```ini
[tox]
minversion = 3.3.0
isolated_build = true

[testenv]
deps = poetry
...
```

In build environments for projects already using tox/poetry etc. common practice is to install this before executing `tox`. Adding the above, will simply increase the environment setup time with no real advantage.

PR #2601, opened 2020-06-29
Add a new setting "virtualenvs.in-project-dir" 

moriyoshi:
# What this patch accomplishes

This patch adds a new setting `virtualenvs.in-project-dir`, with which one can specify the location where an in-project virtualenv is created when `virtualenvs.in-project` is set to true.

# Pull Request Check List


Secrus:
@moriyoshi are you still interested in bringing it to Poetry? If so, please update your branch with code from `master` branch and solve potential conflicts.

neersighted:
To avoid adding another knob, I would actually want to make the `virtualenvs.in-project` option a Union of str and bool -- `True` would be interpreted as the default (`.venv`), and a string would specify a custom name.

Issue #2608, opened 2020-06-30
Poetry crashing on empty dist-info folder


bharel:
- [x] I am on the [latest](https://github.com/python-poetry/poetry/releases/latest) Poetry version.
- [x] I have searched the [issues](https://github.com/python-poetry/poetry/issues) of this repo and believe that this is not a duplicate.
- [x] If an exception occurs when executing a command, I executed it again in debug mode (`-vvv` option).

- **OS version and name**: macOSX 15
- **Poetry version**: 1.1.0a2

## Issue
If you try to run `poetry update` when you have an empty .dist-info folder in your current working directory, or inside any other directory on the `sys.path`, poetry will crash with a random non-indicative exception, and you'll stand there crying, not knowing what you did wrong.

![image](https://user-images.githubusercontent.com/742359/86182741-2e261880-bb39-11ea-9f79-65bf72f8d77c.png)

After crying for 2 hours, I've figured out the issue and created a PR.

Now I'm happy.

Secrus:
The problem seems to no longer exist. Closing

PR #2609, opened 2020-06-30
Skip empty dist-info folder

bharel:
# Pull Request Check List

Resolves: #2608

Hurray


abn:
@bharel I am not entirely sure this is a good solution because the empty `dist-info` means the environment is probably broken or the package could also have bad metadata. Ignoring the issue might cause unintended side-effects. I'd be more interested to know if poetry caused this in the first place.

Secrus:
The problem no longer exists, tested on the latest master. Closing

Issue #1900, opened 2020-01-16
`'BinOp' object has no attribute 's'` when trying to add a git dependency with poetry


herochao:
- [x] I am on the [latest](https://github.com/python-poetry/poetry/releases/latest) Poetry version.
- [x] I have searched the [issues](https://github.com/python-poetry/poetry/issues) of this repo and believe that this is not a duplicate.
- [x] If an exception occurs when executing a command, I executed it again in debug mode (`-vvv` option).

- **OS version and name**: Archlinux with Kernel Linux 5.4.11-arch1-1 & CentOS7 with Kernel Linux 3.10.0-1062.9.1.el7.x86_64
- **Poetry version**: 1.0.2
- **Link of a [Gist](https://gist.github.com/) with the contents of your pyproject.toml file**: [pyproject.toml Gist](https://gist.github.com/herochao/dd49324f06582f60be349c3428bbc66c)

## Issue

When trying to add statsmodels in dev version to the project on either of the two systems using
```
poetry add git+https://github.com/statsmodels/statsmodels.git -vvv
```
(and without -vvv), I get `'BinOp' object has no attribute 's'` [(full output with -vvv)](https://gist.github.com/herochao/dd49324f06582f60be349c3428bbc66c#gistcomment-3140322)
`pip install git+https://github.com/statsmodels/statsmodels.git` works fine though.

abn:
The reason why the add is failing is because the fallback setup.py reader does not like complex setup files very much. Same with the original issue - ast evaluation craps out.

The `extras_requires` extracted from this line is assumed to be a `List[str]` in the fallback reader.

https://github.com/statsmodels/statsmodels/blob/9bf1d0734dbb45abcfc6db7fab4564db851babeb/setup.py#L348-L348

Unfortunately for us, it is a binary operation in the AST and hence the failure.

https://github.com/statsmodels/statsmodels/blob/9bf1d0734dbb45abcfc6db7fab4564db851babeb/setup.py#L73-L77

The fallback itself is triggered when the `python setup.y egg_info` command fails when determining package information of the fresh clone. There are various reasons why this might happen. Most common in these scenarios is that the build envrionment (a clean venv as used by poetry) does not contain build time dependencies (which may include platform deps too).

Workaround for local development, would be to build a wheel and add it to the dependencies section.

```toml
statsmodels = { file = "/path/to/wheel" }
```

jayvdb:
I have a similar case, where there is a named variable in extra_requires.

[...]

PR #2632, opened 2020-07-04
inspection: use pep517 metadata build

abn:
This change replaces `setup.py` execution in favour of pep517 metadata builds. The proposed approach executes `pep517.meta.build` instead of `python setup.py egg_info`.

This does not change the order of preference in the inspection flow for execution vs AST parsing. A build is only perfromed in cases where existing metadata inspection, `SetupReader` parsing etc returns an incomplete/unusable `PackageInfo` instance or raises `PackageInfoError`.

The original proposal was made in #2296 by @finswimmer.

Closes: #2296 
Resolves: #1900

# Pull Request Check List


poetry/inspection/info.py
```diff
@@ -392,10 +401,44 @@ def _get_poetry_package(path):  # type: (Path) -> Optional[ProjectPackage]
...
+        try:
+            info = cls.from_setup_files(path)
+            if info.requires_dist is not None:
+                return info
+        except PackageInfoError:
+            pass
+
+        try:
+            with temporary_directory() as tmp_dir:
+                pep517_metadata_build(
```
sdispater, line 421: The issue I have with this is that `pep517` uses `sys.executable` (i.e. the Python interpreter used to run Poetry) and, thus, is not compatible with Poetry's environment system.

That's why I mentioned in https://github.com/python-poetry/poetry/pull/2296#issuecomment-636839053 that we would most likely not be able to use `pep517`.

abn: We can workaround that by executing `python -m pep517.meta.build` under the project's environment.

sdispater: The python version of Poetry's environment is not necessarily the same as the one from the project's environment so if a build backend relies on it, and this is particularly true for setuptools, we might end up with the wrong information.

abn: @sdispater I have added a commit wherein we retail the old behaviour wrt venv before building metadata.

abn:
@sdispater bumping this once again with the proposed changes addressing your concerns

Issue #2617, opened 2020-07-02
Exported requirements.txt does not include --extra-index-url


fredrikaverpil:
- [x ] I am on the [latest](https://github.com/python-poetry/poetry/releases/latest) Poetry version.
- [x ] I have searched the [issues](https://github.com/python-poetry/poetry/issues) of this repo and believe that this is not a duplicate.
- [ ] If an exception occurs when executing a command, I executed it again in debug mode (`-vvv` option).

- **OS version and name**: RHEL Linux 7.6, Python 3.7.4
- **Poetry version**: 1.0.9
- **Link of a [Gist](https://gist.github.com/) with the contents of your pyproject.toml file**: n/a

## Issue

I _think_ I found a bug...
I have two internal repos (the second one requries credentials which I have stored in my pyproject.toml):
```
[[tool.poetry.source]]
name = "internal-mirror"
url = "https://server/artifactory/api/pypi/pypi-remote/simple"
default = true

[[tool.poetry.source]]
name = "myproject"
url = "https://server/artifactory/api/pypi/myproject-pypi-local/simple"
secondary = true
```

Poetry successfully installs using both repos, nice!
But when I export to requirements.txt, I only get the --index-url and not the --extra-index-url definitions, which makes this requirements file incomplete and pip is unable to find my secondary repo:
```
$ poetry export --format requirements.txt --without-hashes --output requirements.txt
```
```
--index-url https://server/artifactory/api/pypi/pypi-remote/simple

myproject==3.1.7
```

To me, this looks like a bug. I would expect that the --extra-index-url would also appear in the exported requirements.txt file, right?

klintan:
I have a similar issue: running `poetry export --without-hashes -f requirements.txt -o requirements.txt --with-credentials` manually, and then using `usePoetry: false` it works, and my private secondary repo is defined like so in `pyproject.toml` 

```
[[tool.poetry.source]]
name = "team-***"
url = "https://***.io/***/api/pypi/team-***/simple"
secondary = true
```

however it seems it doesn't export it correctly in the python-serverless poetry (or at least if using poetry directly it fails: 

```
  Error: STDOUT: Looking in indexes: https://pypi.org/simple, https://***.jfrog.io/***/api/pypi/team-***/simple
  User for ***.jfrog.io:

  STDERR: ERROR: Exception:
  Traceback (most recent call last):
[...]

PR #2640, opened 2020-07-07
Include --extra-index-url

sublee:
# Pull Request Check List

Resolves: #2617

---

# Bug

In some cases, an exported `requirements.txt` does not include `--extra-index-url` even it is required. The reason is that `--index-url` line may overwrite the buffered `--extra-index-url` lines. This bug depends on how the URLs are sorted.

# Fix

To fix this bug, I made `--index-url` line do not overwrite the buffered `--extra-index-url` lines. Also, to test the fix, I had to change `Pool.remove_repository()` on a default repository to reset `Pool.has_default()`.

### DIFF
--- a/poetry/repositories/pool.py
+++ b/poetry/repositories/pool.py
@@ -87,6 +87,9 @@ class Pool(BaseRepository):
         if idx is not None:
             del self._repositories[idx]
 
+        if idx == 0 and self.has_default():
+            self._default = False
+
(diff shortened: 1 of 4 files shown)


Secrus:
`export` is now handled by [poetry-plugin-export](https://github.com/python-poetry/poetry-plugin-export). Closing as no longer valid.

PR #2602, opened 2020-06-29
Fix #2597 by checking that the directory exists

PetterS:
Resolves: #2597

- [x] Added **tests** for changed code.

PR #2648, opened 2020-07-10
Release 1.1.0a3

sdispater:
### Added

- New installer which provides a faster and better experience ([#2595](https://github.com/python-poetry/poetry/pull/2595)).

### Fixed

- Fixed resolution error when handling duplicate dependencies with environment markers ([#2622](https://github.com/python-poetry/poetry/pull/2622)).
- Fixed erroneous resolution errors when resolving packages to install ([#2625](https://github.com/python-poetry/poetry/pull/2625)).
- Fixed errors when detecting installed editable packages ([#2602](https://github.com/python-poetry/poetry/pull/2602)).


aviramha:
Do we need to do anything to enable #2595 or is it enabled by default?
It doesn't seem faster :(

sdispater:
@aviramha It's enabled by default, yes. However, it's possible that on some CI systems the parallelization is not activated.

aviramha:
> @aviramha It's enabled by default, yes. However, it's possible that on some CI systems the parallelization is not activated.

What do you mean by that? What kind of parallelization needs to be enabled?

sdispater:
@aviramha For now, there is nothing you can do to force the parallelization on some CI systems.

Note that I assumed that you were not seeing speed improvements in your CI because that's the only case where the parallelization might not be enabled, however if it's not the case I would like to know to track down the issue.

As a side note, the relevant code snippet that explains why it's not activated on some CI systems is here: https://github.com/python-poetry/poetry/blob/develop/poetry/installation/executor.py#L42

However, it was here more as a safety net in the first iterations on the feature and I think it can be removed now that it has been thoroughly tested.

aviramha:
You're correct the improvement isn't seen on bitbucket pipelines.

aviramha:
@sdispater I took the liberty of removing the check. I really want to speed up my CI :)

PR #2650, opened 2020-07-10
Ensure we have the latest pip version in CI before installing dependencies

sdispater:
Following the changes made in #2595 the chosen wheels for installation can not be always be installed with old versions of pip if they are too specific or with recent tags.

This caused our CI to fail. This PR fixes it by ensuring we use the latest version of pip.

## Pull Request Check List


PetterS:
Isn't this a sign that the version of pip used when creating the virtualenv should be much newer? Or specified in the pyproject file?

Issue #496, opened 2018-10-16
auto-resolve all/most merge conflicts


seansfkelley:
- [x] I have searched the [issues](https://github.com/sdispater/poetry/issues) of this repo and believe that this is not a duplicate.

## Issue

When two devs install dependencies on separate branches, it is very easy to end up merge-conflicted, in particular, the `metadata.content-hash` key often changes. It is very unclear how to resolve this manually, so I often delete the lockfile (or perhaps just that key) and rebuild it and, basically, hope that it comes out the same.

It seems like in some scenarios that merge conflicts could be resolved automatically based on pyproject. https://github.com/yarnpkg/yarn/ does this, for instance.

simonpercivall:
+1 Resolving the `content-hash` and trivial conflicts in the individual hashes section would be very welcome.

jimbo1qaz:
Workaround:
- remove content-hash line
- `poetry add pathlib2`
- `poetry remove pathlib2`.

Will this issue be resolved soon?

Does a missing content-hash hurt security or not?

seansfkelley:
Yarn's implementation of this is way less insane than I was expecting:

1. Extract the two versions of the conflicted file.
    https://github.com/yarnpkg/yarn/blob/a7334da31bf783af7a3efab451589fe7ac34c748/src/lockfile/parse.js#L397

2. Blindly try to parse the files, and if that's successful, shallowly merge them.
    https://github.com/yarnpkg/yarn/blob/a7334da31bf783af7a3efab451589fe7ac34c748/src/lockfile/parse.js#L399

If the merge conflict resulted in a syntax error, it fails. Yarn's lockfile structure is designed to make this easier: it's flat, it's sorted alphabetically and node is okay with duplicated dependencies in the tree.

itajaja:
is `content-hash` mandatory though? the main problem I am encountering is that dependabot will have to ALWAYS rebase all the PRs, which could be avoided if `content-hash` was not in poetry.lock. in theory shouldn't be necessary, as it can always be computed by the other hashes, right>

chdsbd:
I think this is still a valid issue and should not be closed.

cjolowicz:
I have written a stand-alone script as a stopgap. It only goes half the way, resolving the merge conflict for `metadata.content-hash`. The remaining conflicts, if any, should be quite trivial to merge manually. Use at your own risk.

https://github.com/cjolowicz/scripts/blob/master/python/poetry-merge-lock.py

[...]

PR #2654, opened 2020-07-11
feat: add --no-hash option to update, preventing metadata.content-hash

donbowman:
Currently when creating a lockfile, a metadata.content-hash field is
generated, containing the hash of the sorted contents of pyproject.toml.

However (see #496) this causes merge conflicts, particularly
when using a tool such as dependabot that makes multiple parallel
branches with dependency changes.

Add a --no-hash option to poetry update. This causes metadata.content-hash
to be omitted. In turn poetry will later complain that the lock file
may not be fresh, (although it is correct), and the user can run poetry lock
to update it, after all the merges are complete.

dependabot currently runs `poetry update <DEPENDENCY> --lock`, I would
propose to incrporate this PR to poetry and then modify dependabot
to have teh --no-hash option.

# Pull Request Check List

Resolves: #issue-number-here

### DIFF
--- a/docs/docs/cli.md
+++ b/docs/docs/cli.md
@@ -174,6 +174,7 @@ update the constraint, for example `^2.3`. You can do this using the `add` comma
 * `--dry-run` : Outputs the operations but will not execute anything (implicitly enables --verbose).
 * `--no-dev` : Do not install dev dependencies.
 * `--lock` : Do not perform install (only update the lockfile).
+* `--no-hash` : Do not write a content-hash to the lockfile. Useful with merge conflics.
 
 ## add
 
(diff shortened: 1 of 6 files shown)


Secrus:
Hi, @donbowman are you still interested in this topic?

donbowman:
> Hi, @donbowman are you still interested in this topic?

yes, dependabot still is inneficient for us.

Secrus:
The Poetry team has decided to reject this idea. This doesn't solve the underlying issue, which is that there is no good way of having a more mergeable lock file with our current data model.

neersighted:
See #496 for more discussion of this.

PR #2663, opened 2020-07-13
Enable parallelization in CI

aviramha:
As discussed in https://github.com/python-poetry/poetry/pull/2648 this check is probably no longer needed.

### DIFF
--- a/poetry/installation/executor.py
+++ b/poetry/installation/executor.py
@@ -29,7 +29,7 @@ from .operations.update import Update
 
 
 class Executor(object):
-    def __init__(self, env, pool, config, io, parallel=None):
+    def __init__(self, env, pool, config, io, parallel=True):
         self._env = env
         self._io = io
         self._dry_run = False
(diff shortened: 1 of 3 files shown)


aviramha:
I'm not sure why the tests failed - is develop broken or something?

abn:
```
>       assert expected == tester.io.fetch_output()
E       AssertionError: assert '\nUpdating de...1.2 9cf87a2)\n' == '\nUpdating dep... tag found.\n\n'
E           
```
Might want to strip the values before comparision.

aviramha:
Thanks! I'll commit it soon. I wonder how I broke that test with my change tho. The parallel feature prints more new lines?

aviramha:
I think it's OK now - Cirrus CI seems to fail randomly (before even starting to test). I don't think it's related..

abn:
> Thanks! I'll commit it soon. I wonder how I broke that test with my change tho. The parallel feature prints more new lines?

Output testing across python versions and platforms can get wierd. The wierd bit is that the new line occured only for one env.

I have retriggered the failing Cirrus CI job.

aviramha:
@abn the cirrus bug seems to be known and was declared as fixed,  I sent the failing task in the relevant issue. Hopefully they'll reopen it.
https://github.com/cirruslabs/cirrus-ci-docs/issues/592

aviramha:
Looks good now, any comments?

PR #2666, opened 2020-07-13
Standardise on using virtualenv package

abn:
Until now poetry used the built-in venv module available in Python 3.
This has presented a few concerns. Chief of which has been
inconsistent environment setup. This includes, but is not limited to,
pip version used by the `ensurepip` module which is in turn used by the
built-in `venv.EnvBuilder`. Additionally, the `virtualenv` package
allows for faster environment creation. This can also allow us to, going
forward, drop ad-hoc code required to inspect and manage environments.

See also: #2650
Closes: #732 

- [x] Added **tests** for changed code.]

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -1150,7 +1150,7 @@ docs = ["sphinx", "jaraco.packaging (>=3.2)", "rst.linker (>=1.9)"]
 testing = ["pathlib2", "unittest2", "jaraco.itertools", "func-timeout"]
 
 [metadata]
-content-hash = "363cbb8480f9bbda72b318c6edfcaa1de130ceecba7c7d0e774005a3bca91df5"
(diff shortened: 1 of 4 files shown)


ericriff:
I'm on version `1.1.0a2`, which is the latest, but new environments still pick up pip 8 (From python 3.5) instead of 20 which I have installed on my system.
`$ python3 -m poetry shell
Creating virtualenv poetry-project in /home/eric/project/poetry/.venv
Spawning shell within /home/eric/project/poetry/.venv

$ pip --version
pip 8.1.1 from /home/eric/project/poetry/.venv/lib/python3.5/site-packages (python 3.5)
`
Shouldn't the usage of virtualenv instead of venv change this behavior?

If I use raw virtualenv it works.
`$ python3 -m virtualenv .venv
$ source .venv/bin/activate
$ pip --version
pip 20.2.2 from /home/eric/project/virtualenv/.venv/lib/python3.5/site-packages/pip (python 3.5)
`

Issue #2667, opened 2020-07-13
Cannot resolve git branch names using Windows MSYS installations of git


donkopotamus:
- [x] I am on the [latest](https://github.com/python-poetry/poetry/releases/latest) Poetry version.
- [x] I have searched the [issues](https://github.com/python-poetry/poetry/issues) of this repo and believe that this is not a duplicate.
- [x] If an exception occurs when executing a command, I executed it again in debug mode (`-vvv` option).

- **OS version and name**: Windows 7 + MSYS
- **Poetry version**: >=1.0.5

## Issue

This issue is specific to Windows installations that are using an `msys` or `cygwin` version of `git`.  Suppose we wish to install a specific branch of a repository:

    something = { git = "ssh://git@.../something.git", branch = "some/branch"}

Then an error will be raised
```
CalledProcessError
Command '['git', '--git-dir', ..., 'rev-parse', u'some/branch^{commit}']' returned non-zero exit status 128
```
This error is raised in `poetry >= 1.0.5`, but not in `poetry==1.0.3`.  It is due to a change in how revisions are resolved to commit hashs.  In particular, `poetry>=1.0.05` effectively executes the following in a `subprocess` to resolve the revision

    git ... rev-parse 'some/branch^{commit}'

(while in `poetry==1.0.03` it simply uses `git ... rev-parse 'some/branch'` avoiding the problem about to detailed)

Unfortunately, due to some intermediate `msys` argument parsing machinery, the braces are dropped unless they are escaped as `\{commit\}`, meaning `git` actually receives a meaningless `git ... rev-parse 'some/branch^commit'`.  We can observe this in a python interpreter:
```
>>> subprocess.check_output(['git'] + [..., 'rev-parse', 'some/branch^{commit}'])
fatal: ambiguous argument 'some/branch^commit': unknown revision or path not in the working tree.
                                       ^^^^^^
```

This means:
* on nearly all platforms we wish to be executing `git ... rev-parse 'some/branch^{commit}'`
* on `msys` and `cygwin` based platforms we wish to be executing `git ... rev-parse 'some/branch^\{commit\}'`

We can avoid having to make decisions about the syntax however if instead we execute the equivalent form

    git ... rev-parse 'some/branch^0'

PR #2668, opened 2020-07-13
Alter rev-parse syntax to support msys git

donkopotamus:
On windows, `msys` and `cygwin` versions of `git` need braces to be escaped when they are invoked via subprocess.  (This appears to be an unfortunate side effect of the machinery that converts windows paths to posix-like paths for the executable, and is usually hidden when executing from an msys based shell).

If they are not escaped, then `rev^{commit]` becomes a meaningless `rev^commit` to an `msys` `git`, resulting in an error.

In contrast, we do **not** want to escape the braces if passing `rev^{commit}` to any other form of `git`.  We can avoid having to decide whether escaping is needed by using the alternative notation `^0`.

As noted in the [`rev-parse` documentation](https://git-scm.com/docs/git-rev-parse#Documentation/git-rev-parse.txt-emltrevgtlttypegtemegemv0998commitem)

> A suffix `^` followed by an object type name enclosed in brace pair means dereference the object at `<rev>` recursively ... 
> `<rev>^0` is a short-hand for `<rev>^{commit}`.

# Pull Request Check List

Resolves: #2667

- [ ] Added **tests** for changed code. *No additional tests should be necessary*
- [ ] Updated **documentation** for changed code. *No additional documentation necessary*

### DIFF
--- a/poetry/vcs/git.py
+++ b/poetry/vcs/git.py
@@ -229,9 +229,15 @@ class Git:
                 folder.as_posix(),
             ]
 
-        # We need "^{commit}" to ensure that the commit SHA of the commit the
-        # tag points to is returned, even in the case of annotated tags.
-        args += ["rev-parse", rev + "^{commit}"]
+        # We need "^0" (an alternative to "^{commit}") to ensure that the
+        # commit SHA of the commit the tag points to is returned, even in
+        # the case of annotated tags.
+        #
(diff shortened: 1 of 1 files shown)


donkopotamus:
@finswimmer Rebasing is no help here, all the relevant code has moved to the `poetry-core` repo!  Have reopened this as python-poetry/poetry-core#69

Issue #2564, opened 2020-06-16
poetry does not prefer custom repository but docs say it does


MrGreenTea:
- [x] I am on the [latest](https://github.com/python-poetry/poetry/releases/latest) Poetry version.
- [x] I have searched the [issues](https://github.com/python-poetry/poetry/issues) of this repo and believe that this is not a duplicate.
- [x] If an exception occurs when executing a command, I executed it again in debug mode (`-vvv` option).

- **OS version and name**: elementary OS 5.1.5 Hera x86_64
- **Poetry version**: 1.1.0a1 / 1.0.9
- **Link of a [Gist](https://gist.github.com/MrGreenTea/ccabb109385da851718bb1e0dd2b10f0) with the contents of your pyproject.toml file**: https://gist.github.com/MrGreenTea/ccabb109385da851718bb1e0dd2b10f0

## Issue

When adding a custom repository that contains a package that also exists on pypi.org, it will be downloaded from pypi.org.
My use case is adding a wheel for typed-ast to my custom repo that works on windows and python 3.9, as such a wheel is not provided on pypi.org yet.
Sadly this does not work and installation on windows fails.
In the [docs](https://python-poetry.org/docs/repositories/#install-dependencies-from-a-private-repository) it says
> Any custom repository will have precedence over PyPI.

The wheel also does not appear in `poetry.lock`:
```toml
typed-ast = [
    {file = "typed_ast-1.4.1-cp35-cp35m-manylinux1_i686.whl", hash = "sha256:73d785a950fc82dd2a25897d525d003f6378d1cb23ab305578394694202a58c3"},
    {file = "typed_ast-1.4.1-cp35-cp35m-manylinux1_x86_64.whl", hash = "sha256:aaee9905aee35ba5905cfb3c62f3e83b3bec7b39413f0a7f19be4e547ea01ebb"},
    {file = "typed_ast-1.4.1-cp35-cp35m-win32.whl", hash = "sha256:0c2c07682d61a629b68433afb159376e24e5b2fd4641d35424e462169c0a7919"},
    {file = "typed_ast-1.4.1-cp35-cp35m-win_amd64.whl", hash = "sha256:4083861b0aa07990b619bd7ddc365eb7fa4b817e99cf5f8d9cf21a42780f6e01"},
[...]

Issue #1677, opened 2019-12-05
Private repos don't actually take precedence


m1nde:
- [x] I have searched the [issues](https://github.com/sdispater/poetry/issues) of this repo and believe that this is not a duplicate.

## Issue

I am using poetry 1.0.0b8. According to the latest documentation (https://github.com/sdispater/poetry/blob/master/docs/docs/repositories.md), if I have a project with a custom source in my pyproject.toml, it will always take precedence when installing packages.

This is not the case: We have a custom python package in our company JFrog that we named pyvault. By chance, there is a package in PyPI with the same name. In any project where we have our internal JFrog as a source, `poetry add pyvault` still adds the pyvault from PyPI. We can change that by specifying `pyvault = { version = "^1.0", source = "jfrog" }` in the pyroject.toml, so there is a work around, but the documentation is wrong.

pappasam:
I'm also curious to see whether, and how, this treatment of private package indexes has been intentionally changed. Maybe there's a more-elegant treatment of alternative sources that just needs better documentation :crossed_fingers:

adawalli:
I also noticed this behavior when using artifactory. My work around is to use artifactory virtual repository, and then add 
`default=true`
to the TOML entry for the repo, but its not ideal

Imaclean74:
I was also surprised by this new behavior. Previously raised this issue : https://github.com/python-poetry/poetry/issues/1658
setting default=true does the trick but it's not obvious this is required

levrik:
A comment here states that the PyPi repo is always added as last so private repos are preferred.
https://github.com/python-poetry/poetry/blob/master/poetry/factory.py#L185-L188
But it's added as the default repo which seems to always be put in the first place here:
https://github.com/python-poetry/poetry/blob/master/poetry/repositories/pool.py#L53

PR #2678, opened 2020-07-16
Fixes custom repositories with wrong precedence

linw1995:
# Pull Request Check List

- Resolves: #2564 
- Resolves: #1677

1. If add any custom repository, its precedence will over PyPI.
2. If not, there is only one repository PyPI. 

So there is no point to make PyPI repository default. 

The fixtures directory `tests/fixtures/with_custom_source` is just copied from `tests/fixtures/with_default_source`, and remove the `default=true` configuration from `pyproject.toml` file.


neersighted:
This is obsolete after the refactoring of the Repository code. If you think there is still value in this, please re-open the PR and rebase it/explain why the current behavior is incorrect. Thanks!

Issue #2676, opened 2020-07-16
`SolverProblemError` happens with unclear version specification error


himkt:
- [x] I am on the [latest](https://github.com/python-poetry/poetry/releases/latest) Poetry version.
- [x] I have searched the [issues](https://github.com/python-poetry/poetry/issues) of this repo and believe that this is not a duplicate.
- [x] If an exception occurs when executing a command, I executed it again in debug mode (`-vvv` option).

- **OS version and name**: Ubuntu 20.04
- **Poetry version**: 1.0.9
- **Link of a [Gist](https://gist.github.com/) with the contents of your pyproject.toml file**: https://gist.github.com/himkt/b92e6391b150fa320fa7813df4738943
- `setup.py`: https://gist.github.com/himkt/832038a04fd0f81f66d67553b5a847c8
- PyPI: https://pypi.org/project/konoha
- Raw log of `poetry add -vvv konoha -E all_with_integrations`: https://gist.github.com/himkt/0cbdb631d1555cb96aafea347c5da1bb

## Issue

I encountered the error in installing a third-party library, named [Konoha](https://github.com/himkt/konoha) which I'm developing myself.
When I try to install konoha with extras by running the following command: `poetry add konoha -E all_with_integrations`, I got the following error message.

```bash
[SolverProblemError]
Because no versions of konoha match >4.4.0,<5.0.0
 and konoha (4.4.0) depends on sudachidict_core (>=20200330,<20200331), konoha (>=4.4.0,<5.0.0) requires sudachidict_core (>=20200330,<20200331).
So, because no versions of sudachidict-core match >=20200330,<20200331
 and sample depends on konoha (^4.4.0), version solving failed.
The command '/bin/sh -c poetry add konoha -E all_with_integrations' returned a non-zero code: 1
```

> konoha (4.4.0) depends on sudachidict_core (>=20200330,<20200331), konoha (>=4.4.0,<5.0.0) requires sudachidict_core (>=20200330,<20200331)

It doesn't make sense for me because the depending and the requiring are identical.
(`sudachidict_core (>=20200330,<20200331)`)

You can reproduce the results by executing the following step.

```
git clone git@github.com:himkt/poetry-failure.git && \
  cd poetry-failure && docker build -t himkt/poetry-failure .
```

I'm not sure if it is the problem of Poetry or the problem of my `pyproject.toml`.
So I'd really appreciate if someone gives me any kind of information or pointers. 🙇‍♂️ 

[...]

PR #2680, opened 2020-07-16
fix: fix handling of packages referenced by the multiple 'extra'

orisano:
# Pull Request Check List

Resolves: #2676 

For a package referenced by a single extra, a SingleMarker is used, where the result of without_extra is AnyMarker.
In the case of a package referenced by a multiple extra, a MarkerUnion is used, where the result of without_extra is an empty MarkerUnion.
There was a problem because the package was not installed if the package's Marker was an empty MarkerUnion.

### DIFF
--- a/poetry/version/markers.py
+++ b/poetry/version/markers.py
@@ -658,7 +658,10 @@ class MarkerUnion(BaseMarker):
         return False
 
     def without_extras(self):
-        return self.exclude("extra")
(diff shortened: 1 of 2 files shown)


abn:
@orisano is this still required? see: https://github.com/python-poetry/poetry/issues/2676#issuecomment-676040942

Issue #1992, opened 2020-02-07
Bundle command


sdispater:
As suggested in https://github.com/python-poetry/poetry/issues/537, the addition of a `bundle` command could prove useful.

There are two cases that we've seen pop up here: the ability to bundle the project and its dependencies in an arbitrary virtual environment and the ability to download the distributions to a specific directory. The former should be implemented directly into Poetry since generating a complete virtual environment makes sense to deploy a Poetry-managed project and the latter could be implemented in a plugin since it's a less common case.

### Implementation

Basically the `bundle` command will be a top-level command (similar to `env`) which will expose various sub commands.

Initially, only the `venv` sub command will be available:

```bash
poetry bundle venv my-venv/
```

If the virtual environment directory already exists it will be removed and regenerated.

Specifying a Python version to use to generate the virtual environment will also be possible:

```bash
poetry bundle venv my-venv/ --python 3.8
poetry bundle venv my-venv/ --python python3.8
poetry bundle venv my-venv/ --python /usr/local/bin/python3
```

### Extensibility

The `bundle` command should be extensible so that new sub commands can be added with the use of plugins. This means that this feature depends on the implementation of the plugin system (which is in progress, see #1237)

kapilt:
something that came up in the context of #537  and something i've been interested in is the ability to publish packages to pypi that are fully frozen (ie incorporating lockfile deps into the upload), but keep the dev dependency set a bit more loose (per pyproject.toml). the issue we've found is semver isn't universal adhered to, and i've realized that many of our project's older pypi releases from an even a year ago aren't even installable by users due to dep graph version drift even with semver pins. 

unfortunately its not clear if thats really the intent with bundle, i'm a little unclear on the use case w/ the current description, ie. how is bundle differentiated to install command. afaict download / extract use case is a bit more common, because its a useful cache artifact for offline install or particular serverless environments, but virtualenvs aren't easily relocatable/portable per se.

zdog234:
Is there a PR in progress for this? I'm going to have a forced staycation next week, so if I have any time left after installing ethernet, I'd love to spend some time getting the ball rolling on this

sdispater:
@zdog234 No PR yet but a WIP branch : https://github.com/python-poetry/poetry/tree/bundle-command

Only tests are missing but the feature is almost ready.

jtc42:
Any updates on this? The WIP branch seems to have gone a bit quiet. This would be super useful for a bunch of projects I’m involved in.

[...]

Issue #537, opened 2018-10-19
Feature Request: poetry deploy


bertjwregeer:
- [x] I have searched the [issues](https://github.com/sdispater/poetry/issues) of this repo and believe that this is not a duplicate.

## poetry deploy

What I am looking for is a way to generate a `poetry.lock` and my project wheel and then deploy them without needing to have any of the source/`pyproject.tom` or anything else available. Preferably into a `venv` of my choosing, or a default created one.

Running `poetry build` with a `--with-lock` that copies the existing `poetry.lock`, renames it to be versioned and adds the hash for the newly built wheel, and then places that in the `dist` folder.

Then when a user wants to deploy, they specifically do:

`poetry deploy myapp-0.1.0-poetry.lock`

And `poetry` would create a new virtualenv (or update it) and installs the application with the locked dependencies, verifying the wheel file (which has to exist in the same folder as the lock file, or maybe add a `--wheel-path` to allow an alternate location to find wheels, default to next to .lock fall back to PyPI).

Either way, the end result should be a virtualenv that has all of the dependencies + package installed as set in the lock file without needing to have the source code available.

This would be great for doing deployments inside of Docker containers, or Heroku build-packs so that you don't need to ship the full source, just the lock file and a wheel.

mmerickel:
I think `poetry build --lock` is a really interesting way to do it. This would just snapshot the pinned dependency set into the wheel instead of the loosely pinned constraints in `pyproject.toml`. Then I use that wheel with any other tool (probably `pip`) in deployment to install my app.

```
$ poetry build --lock
$ python3 -m venv env
$ env/bin/pip install --find-links=dist myapp
```

The alternative is to just use `poetry build` to get a loosely constrained wheel, and some other endpoint like `poetry build --export-requirements-file requirements.txt` that dumps out a file of locked deps. This would be something along the lines of https://github.com/sdispater/poetry/issues/100.

```
$ poetry build --export-requirements-file requirements.txt
$ python3 -m venv env
$ env/bin/pip install --find-links=dist -r requirements.txt myapp
```

hynek:
The lack of something like this is pretty much the only reason I cannot put poetry into production.

As of now, if you're in an activated virtualenv and call `poetry install`, it does exactly what you want to.

However it's impossible to pass it as a parameter or env variable which is a problem on build servers.

A cheap win would be something like `TARGET_VENV=/app poetry install` but I understand that that's rather clunky, especially in the context of the whole design.

[...]

PR #2682, opened 2020-07-17
Bundle command implementation

sdispater:
This PR adds a new `bundle venv` command as discussed initially in #537 and formalized in #1992.

Basically, this command creates a virtual environment in the given directory and install the dependencies of the project and the project itself inside it.

## Pull Request Check List

Resolves: #537
Resolves: #1992 


abn:
I feel like we are re-implementing a fair amount of logic here wrt building, env management, and installation. I am not a 100% sure if it is out of necessity (haven't looked at the code changes in detail yet). With the creation of the `virtualenv`, we should be careful in that the environment created should not use symlinks etc for platlib (not sure how this is handled out of the box). Additionally, can't we skip installation of setuptools/pip/wheel etc during environment creation? Since we can rely on the development environment's `pip` to handle installation to specified site right?

Overall I am a bit weary of this functionality, while useful, introduces a fair amount of fragility.

sdispater:
> I feel like we are re-implementing a fair amount of logic here wrt building, env management, and installation.

I don't understand. This command does not re-implement anything but uses existing modules (env management, building and installation) in a specific way, i.e. to bundle the project and its dependencies in a "portable" way.

The purpose of the `bundle` sub commands is to do exactly that. The idea is to make it extensible so that it can be used to bundle projects in various, ready-to-deploy formats.

Imaclean74:
I haven't looked thru the implementation in detail - but is the goal to achieve something like what [shiv](https://github.com/linkedin/shiv) is doing ?  I've been using poetry build + shiv to produce a deployable artifact. Great if this can all be done thru poetry. I guess the key difference is that Shiv is also packing things into an executable .zip ( .pyz ), but otherwise the packaging of dependencies looks pretty similar.

docs/docs/cli.md
```diff
@@ -482,6 +482,48 @@ associated with a specific project.
...
+poetry bundle venv /path/to/environment -p python3.8
+poetry bundle venv /path/to/environment -p 3.8
+```
+
+!!!note
+
+    If the virtual environment already exists, two things can happen:
+
+    - **The python version of the virtual environment is the same as the main one**: the dependencies will be synced (updated or removed).
+    - **The python version of the virtual environment is different**: the virtual environment will be recreated from scratched.
```
Imaclean74, line 519: scratch ?

bratao:
+1, just to contribute to the conversation. This feature is a must for FaaS scenarios such as AWS lambda, where we need to deploy a filly contained app, with all the dependencies @abn @sdispater

docs/docs/cli.md
kkozmic, line 519: scratch
```suggestion
    - **The python version of the virtual environment is different**: the virtual environment will be recreated from scratch.
```

abn:
@sdispater with all the things going on around venv/zipapp/wheelhouse etc; I think I need to revisit my stance on this. I think I am amenable to this top-level command being introduced to `poetry`. Moved this to draft for now. Once this is doen I want to do a [PEP 441](https://www.python.org/dev/peps/pep-0441/) bundler as well.

Moving this to draft for now in order verify things are still up-to-date. Will review once rebased and marked as ready again.

Imaclean74:
> @sdispater with all the things going on around venv/zipapp/wheelhouse etc; I think I need to revisit my stance on this. I think I am amenable to this top-level command being introduced to `poetry`. Moved this to draft for now. Once this is doen I want to do a [PEP 441](https://www.python.org/dev/peps/pep-0441/) bundler as well.
> 
> Moving this to draft for now in order verify things are still up-to-date. Will review once rebased and marked as ready again.

+1 on the zipapp bundler. Done right - it would remove the need to use a separate tool like shiv or pex

garthk:
It's possible to clone the workings of the `poetry/bundle/venv_bundler.py` in this pull request, _eg._ to end up with this [`bundler.py`](https://gist.github.com/garthk/543413981d615e9b69d2a84f4232882f#file-bundler-py) you can copy into any repository with a `tool.poetry.dev-dependencies` entry for `poetry` itself. Yes, a proper integrated bundle command would be better.

docs/docs/cli.md
retr0h: > I haven't looked thru the implementation in detail - but is the goal to achieve something like what shiv is doing ? I've been using poetry build + shiv to produce a deployable artifact. Great if this can all be done thru poetry. I guess the key difference is that Shiv is also packing things into an executable .zip ( .pyz ), but otherwise the packaging of dependencies looks pretty similar.

@Imaclean74 any chance you can provide more details on this?

sdispater:
Closing in favor of an official plugin (https://github.com/python-poetry/poetry-bundle-plugin) which will be made extensible so that users can add their own bundlers to the command.

PR #2683, opened 2020-07-17
Python 2.7 and 3.5 deprecation

sdispater:
This PR makes the necessary changes to notify users of the deprecation of Python 2.7 and 3.5 in Poetry:

- The documentation has been updated
- The CLI will now display a message when using Python 2.7 and 3.5.

Screenshot 2020-07-17 at 16 44 09

## Pull Request Check List

### DIFF
--- a/README.md
+++ b/README.md
@@ -5,7 +5,10 @@ ensuring you have the right stack everywhere.
 
 ![Poetry Install](https://raw.githubusercontent.com/python-poetry/poetry/master/assets/install.gif)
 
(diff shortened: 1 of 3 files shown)


abn:
@sdispater we should probably clarify that we are talking about poetry's runtime and not the managed project's python version.

sdispater:
@abn I added a note that should clarify that Poetry will still be able to manage any Python environment.

Screenshot 2020-07-19 at 00 45 47

PR #2714, opened 2020-07-24
use run_pip of env in executor (#2701)

sdispater:
Port #2701 to `master` since it was inadvertently merged to `develop`.


abn:
@sdispater lets drop the develop branch. We should then be required to edit the PR to change target branch.

PR #2327, opened 2020-04-20
Git dependencies respect poetry.lock on install

wakemaster39:
# Pull Request Check List

Resolves: #2325

- [x] Added **tests** for changed code.
- [ ] Updated **documentation** for changed code.

There is no documentation change required (I think)

Issue #1614, opened 2019-11-21
Update the lock file without upgrading dependencies


cjolowicz:
- [x] I have searched the [issues](https://github.com/sdispater/poetry/issues) of this repo and believe that this is not a duplicate.
- [x] I have searched the [documentation](https://poetry.eustace.io/docs/) and believe that my question is not covered.

## Feature Request

Please provide a way to update the lock file without upgrading dependencies.

After adding a `[tool.poetry.extras]` section to pyproject.toml, Poetry displays the following warning, for example on install:

> Warning: The lock file is not up to date with the latest changes in pyproject.toml. You may be getting outdated dependencies. Run update to update them.

That's fine, but if I run `poetry update` it upgrades my dependencies, which is not what I want at this time. If I run `poetry lock` instead, it still upgrades dependencies. Am I missing something?

Here are the relevant files and the commit:

- [pyproject.toml](https://github.com/cjolowicz/hypermodern-python/blob/b778b99b4e0b81fdbc503580117a8d16a239f07c/pyproject.toml)
- [poetry.lock](https://github.com/cjolowicz/hypermodern-python/blob/b778b99b4e0b81fdbc503580117a8d16a239f07c/poetry.lock)
- [commit](https://github.com/cjolowicz/hypermodern-python/commit/b778b99b4e0b81fdbc503580117a8d16a239f07c)

Sorry for not providing a smaller reproducible example, it's quite tricky to generate a poetry.lock file with outdated dependencies.

finswimmer:
Hello @cjolowicz ,

have you tried running `poetry install` after `poetry lock`?

fin swimmer

cjolowicz:
Hi @finswimmer, thanks for your reply!

I just reproduced this with Poetry 1.0.0b6. Originally I was on 1.0.0b4 when filing this issue.

To reproduce, clone [my repository](https://github.com/cjolowicz/hypermodern-python) and check out b778b99. Invoke `poetry lock`, then `git diff`. The diff shows that Poetry added an `[extras]` section to the lock file (as it should have), but also upgraded several packages (pyparsing 2.4.4 -> 2.4.5, pytest 5.2.2 -> 5.3.0, pytest-mock 1.11.2 -> 1.12.1, xdoctest 0.10.2 -> 0.10.3).

I don't understand how running `poetry install` after `poetry lock` has any bearing on this. My issue is that there is no method to synchronize the lock file with pyproject.toml without also upgrading dependencies.

[...]

Issue #2325, opened 2020-04-20
Git Dependencies do not respect lock file


wakemaster39:
- [x] I am on the [latest](https://github.com/python-poetry/poetry/releases/latest) Poetry version.
- [x] I have searched the [issues](https://github.com/python-poetry/poetry/issues) of this repo and believe that this is not a duplicate.
- [x] If an exception occurs when executing a command, I executed it again in debug mode (`-vvv` option).

- **OS version and name**: macOS, RedHat 7.6, Fedora 31
- **Poetry version**: 1.0.5
```
[tool.poetry]
name = "test2"
version = "0.1.0"
description = ""
authors = ["Cameron Hurst <cahurst@cisco.com>"]

[tool.poetry.dependencies]
python = "^3.8"
test-abc = {git="git@xxxx:xxxxxxxx/test-abc.git"}

[tool.poetry.dev-dependencies]

[build-system]
requires = ["poetry>=0.12"]
build-backend = "poetry.masonry.api"
```

## Issue
I have created two test projects so I could verify this, simple contain no code just a git repo where one of them lives. The expectation I had, is what when I do a `poetry install` for the first time on my `test2` repo that is shown above, a poetry lock file would be generated which contains the revision that the git should be locked against. This currently happens

```
[package.source]
reference = "6795f40dbe1f9a28dee1c3c5b2fae09763afa0be"
type = "git"
url = "git@xxxxx:xxxxx/test-abc.git"
```

I then go and push a change to `test-abc` and run `poetry install` again in test 2. My expectation here is that nothing changes and life is good. This is where I believe to be a Critical bug occurs:

```
Package operations: 0 installs, 1 update, 0 removals

  - Updating test-abc (0.1.0 6795f40 -> 0.1.0 a10f113)
```

The contents of the poetry.lock file remain unchanged, but I now have an updated dependency.

Now rev locking to that specific revision works, it is just the default does not respect the poetry.lock file defeating the purpose of the lock file here in my opinion.

There are a couple other issues I found on here which reference resolving tags and branches to specific hashes in the lock file, so it seems like the intent here is for git to install the specific hash and not latest by default.

wakemaster39:
I have done more digging and made a PR, but I have run into one problem
https://github.com/python-poetry/poetry/blob/288d99739fba4425def7c567d71537ce5bd2a0f9/tests/puzzle/test_solver.py#L1276-L1309

[...]

PR #2720, opened 2020-07-24
Reuse locked package info for vcs/file/url deps

abn:
Prior to this change, when add/remove/lock commands were executed,
poetry would re-inspect all locked vcs/file/url dependencies. This is
sub-optimal and not required. This change enables reuse of locked
package info. 

Might Resolves: #1614
Closes: #2327 #2325

# Pull Request Check List

### DIFF
--- a/poetry/puzzle/provider.py
+++ b/poetry/puzzle/provider.py
@@ -31,6 +31,7 @@ from poetry.packages import DependencyPackage
 from poetry.packages.package_collection import PackageCollection
 from poetry.puzzle.exceptions import OverrideNeeded
 from poetry.repositories import Pool
+from poetry.repositories import Repository
 from poetry.utils._compat import OrderedDict
 from poetry.utils._compat import Path
 from poetry.utils._compat import urlparse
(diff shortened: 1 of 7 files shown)


wakemaster39:
I gave this a shot to see how it compares to the solution in #2327 as I am all game to kill it. Couple issues I ran into while testing.

`poetry update` does not update git dependencies, that is a feature that would be beneficial. Work around was deleting the poetry lock file.

After using the work around with the upgraded package, doing a poetry install did not downgrade the installed git dependency to the locked version. 

The second issue is pretty critical in my eyes at least as this would then not respect the lock file in a different way than it does currently of always updating to latest.

poetry/puzzle/provider.py
```diff
@@ -67,6 +68,11 @@ def __init__(
         self._in_progress = False
         self._overrides = {}
         self._deferred_cache = {}
+        self._locked_packages = {}
```
sdispater, line 71: I am not sure this is the right place to put this.

We already have a locked system in the `VersionSolver` class, so we might be able to fix and improve it there.

abn: I'll have a whack at that; I was going for the lowest possible place to put this logic otherwise we will end up re-implementing the logic at multiple places. It might also be a good idea to re-work the structure a bit for `2.x`.

wakemaster39:
@abn is there anything I can help with either this PR or #2327 to keep things moving? This is a pretty annoying problem with the git lock not being respected I have to keep working around.

abn:
Closing in favour of #3875.

Issue #2697, opened 2020-07-21
"--author" option for "new"


tmywada:
- [x] I have searched the [issues](https://github.com/python-poetry/poetry/issues) of this repo and believe that this is not a duplicate.
- [x] I have searched the [documentation](https://python-poetry.org/docs/) and believe that my question is not covered.

## Feature Request
Scenario:
Generate new folder with specific author name(s)
Command:
poetry new --src "XXX" (--author "xxx yy@gmail.com")
Expected outcome:
new directory, and a toml file with defined authors

Current version returns an error message ('' The "--author" option does not exist.") We can get the expected output by:
1) generate folder with "new" and swap toml file
2) generate toml file with "initi" option
It would be nice to add "--author" option for "new" command. 

The following is the error message when I use "--author" option for "new" command:
```
  Stack trace:

  11  ~/.poetry/lib/poetry/_vendor/py3.7/clikit/console_application.py:124 in run
      self, args, input_stream, output_stream, error_stream

  10  ~/.poetry/lib/poetry/console/config/application_config.py:213 in create_io
      resolved_command = application.resolve_command(args)

   9  ~/.poetry/lib/poetry/_vendor/py3.7/clikit/console_application.py:110 in resolve_command
      return self._config.command_resolver.resolve(args, self)

   8  ~/.poetry/lib/poetry/_vendor/py3.7/clikit/resolver/default_resolver.py:34 in resolve
      return self.create_resolved_command(result)

   7  ~/.poetry/lib/poetry/_vendor/py3.7/clikit/resolver/default_resolver.py:166 in create_resolved_command
      if not result.is_parsable():

   6  ~/.poetry/lib/poetry/_vendor/py3.7/clikit/resolver/resolve_result.py:43 in is_parsable
      self._parse()

   5  ~/.poetry/lib/poetry/_vendor/py3.7/clikit/resolver/resolve_result.py:49 in _parse
      self._parsed_args = self._command.parse(self._raw_args)

   4  ~/.poetry/lib/poetry/_vendor/py3.7/clikit/api/command/command.py:113 in parse
      return self._config.args_parser.parse(args, self._args_format, lenient)

   3  ~/.poetry/lib/poetry/_vendor/py3.7/clikit/args/default_args_parser.py:53 in parse
      self._parse(args, _fmt, lenient)

   2  ~/.poetry/lib/poetry/_vendor/py3.7/clikit/args/default_args_parser.py:101 in _parse
[...]

PR #2745, opened 2020-07-29
Feature/new author option

spenserblack:
Resolves: #2697 

Adds an `--author` option to the `new` command. This PR basically just ports the `--author` option from `init` to `new`.

  - I didn't see any tests for the `new` command, so I didn't add any
  - Added **Options** header under the ***new*** header

### DIFF
--- a/docs/docs/cli.md
+++ b/docs/docs/cli.md
@@ -65,6 +65,12 @@ my-package
     └── test_my_package.py
 ```
 
+### Options
+
+* `--name`: Set the resulting package name.
+* `--author`: Author name of the package.
+* `--src`: Use the src layout for the project.
+
 ## init
 
 This command will help you create a `pyproject.toml` file interactively
--- a/poetry/console/commands/new.py
+++ b/poetry/console/commands/new.py
@@ -16,6 +16,7 @@ class NewCommand(Command):
     arguments = [argument("path", "The path to create the project at.")]
     options = [
         option("name", None, "Set the resulting package name.", flag=False),
+        option("author", None, "Author name of the package.", flag=False),
         option("src", None, "Use the src layout for the project."),
     ]
 
@@ -47,8 +48,8 @@ class NewCommand(Command):
         readme_format = "rst"
 
         config = GitConfig()
-        author = None
-        if config.get("user.name"):
+        author = self.option("author")
+        if (not author) and config.get("user.name"):
             author = config["user.name"]
             author_email = config.get("user.email")
             if author_email:
--- a/poetry/console/commands/new.py
+++ b/poetry/console/commands/new.py
@@ -49,7 +49,15 @@ class NewCommand(Command):
 
         config = GitConfig()
         author = self.option("author")
-        if (not author) and config.get("user.name"):
+        if author:
+            from poetry.packages.package import AUTHOR_REGEX
+
+            if not AUTHOR_REGEX.match(author):
+                raise ValueError(
+                    "Invalid author string. Must be in the format: "
+                    "John Smith <john@example.com>"
+                )
+        elif config.get("user.name"):
             author = config["user.name"]
             author_email = config.get("user.email")
             if author_email:


poetry/console/commands/new.py
```diff
@@ -16,6 +16,7 @@ class NewCommand(Command):
     arguments = [argument("path", "The path to create the project at.")]
     options = [
         option("name", None, "Set the resulting package name.", flag=False),
+        option("author", None, "Author name of the package.", flag=False),
```
abn, line 19: I am wondering if we should use `--author-name/--author-email`? Would also be great to add the default value in as well.

abn (changes requested):

poetry/console/commands/new.py
spenserblack: >I am wondering if we should use `--author-name/--author-email`?

Would that be *instead* of `--author`, or *with* `--author`? Should these options then be implemented on the `init` command as well?

>Would also be great to add the default value in as well.

Would the default value be git config user fields if available, `None` if not available? This is already the behavior on line 60, though not explicitly defined in the `option` definition.

abn: I would think instead of. 

As for the defaults, yes, same behaviour but initialised outside of the command so that it can be shown in the help message.

poetry/console/commands/new.py
```diff
@@ -4,6 +4,8 @@
 from cleo import option
 
 from poetry.utils.helpers import module_name
+from poetry.vcs.git import GitConfig
```
abn, line 7:
```suggestion
from poetry.core.vcs.git import GitConfig
```

poetry/console/commands/new.py
```diff
@@ -46,13 +52,21 @@ def handle(self):
...
+                "`--author-name` with no default value not used when option `--author-email` "
+                "is defined or has a default value. "
+                "A value for `--author-name` is required when `--author-email` is defined."
+            )
+        if author_name:
             if author_email:
-                author += " <{}>".format(author_email)
+                author = "{name} <{email}>".format(name=author_name, email=author_email)
+            else:
+                author = author_name
```
abn, line 69: This will fail validation I think. We need both to construct a valid author right?

docs/docs/cli.md
```diff
@@ -65,6 +65,12 @@ my-package
     └── test_my_package.py
 ```
 
+### Options
+
+* `--name`: Set the resulting package name.
+* `--author`: Author name of the package.
```
abn, line 71: Guessing this needs update?

abn (changes requested):

poetry/console/commands/new.py
abn: I just saw your comment https://github.com/python-poetry/poetry/pull/2745#discussion_r485630817. Should be fine. I am wondering if PyPI likes it. Should probably check there too.

docs/docs/cli.md
spenserblack: Good catch! Yes, will update.

poetry/console/commands/new.py
spenserblack: This results in an `ImportError` for me

poetry/console/commands/new.py
spenserblack: I've looked at some documentation at various sites, but based on [packaging.python.org](https://packaging.python.org/specifications/core-metadata/#author), the author is
>A string containing the author’s name at a minimum; additional contact information may be provided.

Which makes me *think* PyPI would be OK with it.

But [Poetry's *own* documentation](https://python-poetry.org/docs/pyproject/#authors) states

>Authors must be in the form `name <email>`.

So I believe that I should make the author *always* be either `None` or `"Author Name <author@email.com>"` after all. Will push up a new commit.

poetry/console/commands/new.py
abn: https://www.github.com/python-poetry/poetry-core/tree/master/poetry%2Fcore%2Fvcs%2Fgit.py

spenserblack: Yes, sorry about that. This branch was pretty outdated, I needed to merge from `upstream/master`. Thanks!

spenserblack:
Thanks for your help, @abn! I think these new commits address everything.

spenserblack:
:thinking: Not sure why tests passed on 6f64dd0b4d45f63b635502050c4675b0700801cc but fail on 87f256ac921925c342273b1cd0e52fa6db0b460e. Changes to docs shouldn't affect that failing test, should they?

PR #2747, opened 2020-07-29
use run_pip of env in executor (#2701)

jstriebel:
This ensures to use the correct python executable.
Using `run` simply uses `sys.prefix` with a bin suffix, which is not necessarily the currently used python executable.

E.g. when linking a specific python version to a custom directory, using `sys.prefix` may point to the wrong python version:
* /usr/bin/python2.7
* /usr/bin/python3.6
* /usr/bin/python is a symlink to 2.7
* /usr/bin/local/python (or any other location) is a symlink to 3.6 which shadows the former symlink
`sys.prefix` is `/usr` in this case, which is correct, but the `python` executable there points to the wrong version.

`run_pip` in the env fixes this already by using `sys.executable`, see https://github.com/python-poetry/poetry/blob/develop/poetry/utils/env.py#L964
This should be used in the executor as well.

---------------------------------

This change landed on the `develop` branch by mistake in #2701

### DIFF
--- a/poetry/installation/executor.py
+++ b/poetry/installation/executor.py
@@ -280,7 +280,7 @@ class Executor(object):
 
     def run_pip(self, *args, **kwargs):  # type: (...) -> int
         try:
-            self._env.run("python", "-m", "pip", *args, **kwargs)
+            self._env.run_pip(*args, **kwargs)
         except EnvCommandError as e:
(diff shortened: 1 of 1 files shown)


jstriebel:
Duplicate of #2714

Issue #2744, opened 2020-07-29
1.1.0b2 causes EnvCommandError (Non-posix paths with file:// protocol on Windows)


fredrikaverpil:
- [x] I am on the [latest](https://github.com/python-poetry/poetry/releases/latest) Poetry version.
- [x] I have searched the [issues](https://github.com/python-poetry/poetry/issues) of this repo and believe that this is not a duplicate.
- [x] If an exception occurs when executing a command, I executed it again in debug mode (`-vvv` option).

- **OS version and name**: Windows 10
- **Poetry version**: 1.1.0b2
- **Link of a [Gist](https://gist.github.com/) with the contents of your pyproject.toml file**: 

## Issue

When on Poetry 1.0.10 there are no issues. But after having updated to poetry 1.1.0b2, I constantly get an EnvCommandError caused by a FileNotFoundError (a temporary setup.py file is not found):

```bash
$ poetry --version
Poetry version 1.1.0b2

$ poetry lock
Updating dependencies
Resolving dependencies...

Writing lock file

$ poetry install -vvv                                         
Using virtualenv: C:\Users\fredrik\code\repos\mylittleci\venv
Installing dependencies from lock file

Finding the necessary packages for the current system

Package operations: 2 installs, 0 updates, 0 removals

  • Installing zipp (3.1.0)

  Stack trace:

  7  c:\users\fredrik\.local\pipx\venvs\poetry\lib\site-packages\poetry\installation\executor.py:183 in _execute_operation
      181│
      182│             try:
    → 183│                 result = self._do_execute_operation(operation)
      184│             except EnvCommandError as e:
      185│                 if e.e.returncode == -2:

  6  c:\users\fredrik\.local\pipx\venvs\poetry\lib\site-packages\poetry\installation\executor.py:257 in _do_execute_operation
      255│             return 0
      256│
    → 257│         result = getattr(self, "_execute_{}".format(method))(operation)
      258│
      259│         if result != 0:

  5  c:\users\fredrik\.local\pipx\venvs\poetry\lib\site-packages\poetry\installation\executor.py:394 in _execute_install
      392│
      393│     def _execute_install(self, operation):  # type: (Install) -> None
    → 394│         return self._install(operation)
      395│
      396│     def _execute_update(self, operation):  # type: (Update) -> None

  4  c:\users\fredrik\.local\pipx\venvs\poetry\lib\site-packages\poetry\installation\executor.py:432 in _install
      430│             args.insert(2, "-U")
[...]

PR #2750, opened 2020-07-30
Fix non-posix paths with file:// protocol on Windows

fredrikaverpil:
# Pull Request Check List

Resolves: #2744 

### DIFF
--- a/poetry/installation/chef.py
+++ b/poetry/installation/chef.py
@@ -81,7 +81,7 @@ class Chef:
         links = []
         for archive_type in archive_types:
             for archive in cache_dir.glob("*.{}".format(archive_type)):
-                links.append(Link("file://{}".format(str(archive))))
+                links.append(Link("file://{}".format(str(archive.as_posix()))))
 
         return links
 
--- a/poetry/installation/chef.py
+++ b/poetry/installation/chef.py
@@ -81,7 +81,7 @@ class Chef:
         links = []
         for archive_type in archive_types:
             for archive in cache_dir.glob("*.{}".format(archive_type)):
-                links.append(Link("file://{}".format(str(archive.as_posix()))))
+                links.append(Link(archive.as_uri()))
 
         return links
 
--- a/tests/installation/test_chef.py
+++ b/tests/installation/test_chef.py
@@ -44,17 +44,19 @@ def test_get_cached_archives_for_link(config, mocker):
         ),
     )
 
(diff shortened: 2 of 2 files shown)


poetry/installation/chef.py
```diff
@@ -81,7 +81,7 @@ def get_cached_archives_for_link(self, link):  # type: (Link) -> List[Link]
         links = []
         for archive_type in archive_types:
             for archive in cache_dir.glob("*.{}".format(archive_type)):
-                links.append(Link("file://{}".format(str(archive))))
+                links.append(Link("file://{}".format(str(archive.as_posix()))))
```
abn, line 84:
```suggestion
                links.append(Link(archive.as_uri()))
```

fredrikaverpil:
@abn hm, the errors I am getting... I don't think they are related to my PR, right?

tests/installation/test_chef.py
```diff
@@ -73,3 +73,34 @@ def test_get_cache_directory_for_link(config):
     )
 
     assert expected == directory
+
+
+def test_no_backslashes_with_file_protocol(config, mocker):
```
abn, line 78: You should instead probably just updated the existing test case to verify the issue. Like so;

```py
def test_get_cached_archives_for_link(config, mocker):
    chef = Chef(
        config,
        MockEnv(
            marker_env={"interpreter_name": "cpython", "interpreter_version": "3.8.3"}
        ),
    )

    distributions = Path(__file__).parent.parent.joinpath("fixtures/distributions")
    mocker.patch.object(
        chef,
        "get_cache_directory_for_link",
        return_value=distributions,
    )

    archives = chef.get_cached_archives_for_link(
        Link("https://files.python-poetry.org/demo-0.1.0.tar.gz")
    )

    assert archives
    assert set(archives) == {
        Link(path.as_uri())
        for path in distributions.glob("demo-0.1.0*")
    }
```

abn:
> I don't think they are related to my PR, right?
Correct. #2751

poetry/installation/chef.py
fredrikaverpil: Oh, I was googling around for urlencoding this somehow like this, but I failed to find that this was possible with pathlib. I'm starting to like pathlib more and more 😄 👍

tests/installation/test_chef.py
fredrikaverpil: Done.

Issue #2752, opened 2020-07-30
Poetry does not work with tomlkit 0.6.0


tlc28:
Hi Poetry!

Poetry 1.0.10 is unfortunately broken on Guix (https://guix.gnu.org) since I upgraded tomlkit to version 0.6.0!
The version required by Poetry is "^0.5.11", which translates to ">=0.5.11,<0.6.0". We only have one version of tomlkit packaged, and it's the latest one, 0.6.0.

From what I understand reading SemVer (https://semver.org), minor versions should only "add functionality in a backwards compatible manner", and so the API should be stable. So 0.6.0 should work with Poetry. And, as a matter of fact, when I force it (by modifying the requirements), Poetry seems to be working as expected.

I have 3 options:
- modify the requirements at build time (quick fix);
- package tomlkit version "0.5" alongside the "latest" version (long term solution);
- report the problem upstream to see if there's a cleaner solution.

For the time being, I'll implement solution 1. I'll implement solution 2 if you don't see it as a problem that should be fixed upstream.

WDYT?

tlc28:
Thanks @abn!
Actually, we have the same problem with cleo and clikit… but the versions have been updated in dev. So we'll have to live with a patch until 1.1.0 is out and can be packaged.

abn:
@tlc28 just so you know, 1.1.0 will require `tomlkit-0.7.0` see comments on #2753. I suspect so will cleo and clikit.

tlc28:
Thanks @abn

PR #2753, opened 2020-07-30
Relax tomlkit version requirement

abn:
Resolves: #2752

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -35,7 +35,7 @@ cachecontrol = { version = "^0.12.4", extras = ["filecache"] }
 pkginfo = "^1.4"
 html5lib = "^1.0"
 shellingham = "^1.1"
-tomlkit = "^0.5.11"
+tomlkit = ">=0.7.0,<1.0.0"
 pexpect = "^4.7.0"
 packaging = "^20.4"
 virtualenv = { version = "^20.0.26" }
--- a/poetry.lock
+++ b/poetry.lock
@@ -1028,8 +1028,8 @@ category = "main"
 description = "Style preserving TOML library"
 name = "tomlkit"
 optional = false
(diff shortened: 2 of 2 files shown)


pyproject.toml
```diff
@@ -35,7 +35,7 @@ cachecontrol = { version = "^0.12.4", extras = ["filecache"] }
 pkginfo = "^1.4"
 html5lib = "^1.0"
 shellingham = "^1.1"
-tomlkit = "^0.5.11"
+tomlkit = ">=0.5.0,<1.0.0"
```
sdispater, line 38:
```suggestion
tomlkit = "^0.7.0"
```

I want to stay on the conservative side of things but also ensure we have the latest version available at the time which is the 0.7.0 release, published today: https://github.com/sdispater/tomlkit/pull/104

abn: @sdispater with dependencies like `tomlkit` we might end up having issues similar to #2752 where downstream poetry will be harder to package. If we do not rely on specific features or  are not affected by any bugs in question, we should consider leaving it at the lowest compatible version.

sdispater: @abn Unfortunately, we don't have a lot of choices if we want to support TOML `1.0.0rc1` fully. The `0.6.0` release introduced the heterogeneous arrays support and the `0.7.0` release fixed a lot of issues with `1.0.0rc1` support which were not necessarily backwards compatible if people were relying on the previous, buggy behavior.

abn: I reckon that is solid rationale to require 0.7.0.

abn: @sdispater are you okay with `>=0.7.0,<1.0.0`? I think `>=0.7.0,<0.8.0` might be a too restrictive.

PR #1515, opened 2019-10-29
Default README format to Markdown in "new" (#280)

finswimmer:
This PR implements #280.

**NEW**

* A new parameter called `--readme` for `poetry new` is introduced, which used for defining the README format. This can be set either to `md` or `rst`
* The readme is included in the `pyproject.toml`

**CHANGE**

* The default README format is now `.md`

**FIX**

* Test for `poetry new` is added

# Pull Request Check List

This is just a reminder about the most common mistakes. Please make sure that you tick all *appropriate* boxes.  But please read our [contribution guide](https://poetry.eustace.io/docs/contributing/) at least once, it will save you unnecessary review cycles!

- [x] Added **tests** for changed code.
- [x] Updated **documentation** for changed code.

**Note**: If your Pull Request introduces a new feature or changes the current behavior, it should be based
on the `develop` branch. If it's a bug fix or only a documentation update, it should be based on the `master` branch.

If you have *any* questions to *any* of the points above, just **submit and ask**!  This checklist is here to *help* you, not to deter you from contributing!

Issue #280, opened 2018-07-05
Default README format to Markdown in "new"?


taion:
- [x] I have searched the [issues](https://github.com/sdispater/poetry/issues) of this repo and believe that this is not a duplicate.

## Issue
When Markdown support was initially added in https://github.com/sdispater/poetry/issues/15, Warehouse had somewhat questionable Markdown rendering. At this point, however, it appears that those issues have been resolved, and things like syntax highlighting now work.

As such, given the relative popularity of Markdown even for Python projects, it might make more sense for `new` to default to creating a `README.md` rather than a `README.rst`.

nanthony007:
This is a small issue I understand but has there been any resolution or progress on this? I would be open to contributing if needed.

chemsaf3:
any progress on this? seems like most use md due to github so may not know rst much, would make sense to change to md as default

Frank5000:
Looks like there are a few open PRs that seem to work

PR #2768, opened 2020-08-04
Support namespace packages with new command and misc. improvements

abn:
**Improve supported layout creation**
- remove unnecessary template duplication
- stop generating default test case (this is out of scope for poetry)
- stop generating content for package `__init__.py`
- simplify inheritance to only require base directory overrides
- remove "StandardLayout" (replaced with `Layout`)
- support handling namespace package names given in the form "a.b.c"
- specify include packages by default (be explicit)

**Support namespace package creation (via --name)**

With this change, names containing "." are treated as namespace
packages. The following behaviour is now expected.
- "a.b.c" creates package name "a-b-c" with directory structure  "a/b/c"
- "a-b-c" creates package name "a-b-c" with directory structure  "a_b_c"
- "a.b_c" creates package name "a-b-c" with directory structure  "a/b_c"
- "a_b_c" creates package name "a-b-c" with directory structure "a_b_c"

**Add --readme option with default fromat markdown**

Previously, the new command created a `.rst` readme files as default. This is now changed to use markdown. Additionally,
a new option `--readme` has been added. (changes ported from #1515 by @finswimmer).

Closes: ##1515 #280

**Misc. changes**
- do not add `pytest` dev dependency as this should be left upto the end-user and can be achieved using the cli
- fix handling of input path when it is not relative
- add test coverage for command


poetry/layouts/layout.py
```diff
@@ -83,17 +94,21 @@ def create(self, path, with_tests=True):
...
         poetry_content["version"] = self._version
         poetry_content["description"] = self._description
         poetry_content["authors"].append(self._author)
+
         if self._license:
             poetry_content["license"] = self._license
+        else:
+            poetry_content.remove("license")
+
+        poetry_content["packages"].append(self.get_package_include())
```
finswimmer, line 111: Hm, according to your PR info, it is intended to include `packages` in any case. I like the current behavior more, to only use it, if the structure or name differs from the default. The reason for this is, that it looks to me like a redundancy to say "I have a package called x, please include x". But as I said: Just a personal opinion :)

poetry/layouts/layout.py
```diff
@@ -105,9 +120,14 @@ def generate_poetry_content(self):
...
-        build_system_version = ">=" + BUILD_SYSTEM_MIN_VERSION
+        build_system_version = ""
+
+        if BUILD_SYSTEM_MIN_VERSION is not None:
+            build_system_version = ">=" + BUILD_SYSTEM_MIN_VERSION
         if BUILD_SYSTEM_MAX_VERSION is not None:
-            build_system_version += ",<" + BUILD_SYSTEM_MAX_VERSION
+            if build_system_version:
+                build_system_version += ","
+            build_system_version += "<" + BUILD_SYSTEM_MAX_VERSION
```
finswimmer, line 130: Is there currently a way to set `BUILD_SYSTEM_MIN_VERSION` and `BUILD_SYSTEM_MAX_VERSION`?

sdispater (dismissed):
> "a-b-c" creates package name "a-b-c" with directory structure "a/b/c"

I am not sure we want this, and it's a change to the existing behavior. Putting a dash in the package name is common but does not necessarily mean we want a namespace package.

poetry/console/commands/new.py
```diff
@@ -59,25 +62,24 @@ def handle(self):
...
             name,
             "0.1.0",
             author=author,
             readme_format=readme_format,
             python=default_python,
-            dev_dependencies=dev_dependencies,
         )
         layout_.create(path)
 
+        path.resolve()
```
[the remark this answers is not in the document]
abn: The absolute value is required to determine relative path later on.

poetry/layouts/layout.py
```diff
@@ -50,13 +41,16 @@ def __init__(
         description="",
         readme_format="md",
         author=None,
-        license=None,
+        license=None,  # noqa
```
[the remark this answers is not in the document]
abn: `license` shadows a built-in function.

poetry/layouts/layout.py
abn: I thought about that, however I believe that being explicit is a better practice here. Quite often we endup with bad usages because the development path included auto-discovered packages. There are cases where the name and package name are different (`a-b-c` vs `a.b.c` vs `a_b_c`). Additionally, it allows for easier editing if required as opposed to having to look up the option etc. 

We can discard it if the project name and package name are the same maybe?

poetry/layouts/layout.py
abn: We do not. The expectation at the moment is that this is dependent on the code (ie. hardcoded if required).

poetry/console/commands/new.py
```diff
@@ -59,25 +62,24 @@ def handle(self):
...
-        if parse_constraint("<3.5").allows_any(python_constraint):
-            dev_dependencies["pytest"] = "^4.6"
-        if parse_constraint(">=3.5").allows_all(python_constraint):
-            dev_dependencies["pytest"] = "^5.2"
-
         layout_ = layout_(
             name,
             "0.1.0",
             author=author,
             readme_format=readme_format,
```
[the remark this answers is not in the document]
abn: Can do.

abn:
> I am not sure we want this, and it's a change to the existing behavior. Putting a dash in the package name is common but does not necessarily mean we want a namespace package.

Personally, I reckon this is a better default. Especially since we can do `poetry new --name a_b_c a-b-c` we can also add a confirmation step if required.

poetry/console/commands/new.py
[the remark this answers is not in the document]
finswimmer: Shouldn't it be `path = path.resolve()`?

poetry/layouts/layout.py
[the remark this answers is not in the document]
finswimmer: Good to know :)

poetry/layouts/layout.py
finswimmer: In my opinion, than we should just remove it, if we don't need it yet.

poetry/layouts/layout.py
finswimmer: > We can discard it if the project name and package name are the same maybe?

That would be something, I can live with :+1:

poetry/layouts/layout.py
abn: We will need it in the future if we end up breaking backwards compatibility. We might also be able to extend new/init to support this. Hence why I was hesitant to remove this altogether.

poetry/console/commands/new.py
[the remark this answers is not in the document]
abn: _doh_ yes ... egg on my face :)

abn:
@sdispater I have left the current beahviour as it is for now; but supporting `--name a.b.c` as namespace packages.
@finswimmer all your recoomendations have been incorporated (except the min vesrsion one).

abn:
@python-poetry/core this is rebased and ready for review again

poetry/layouts/layout.py
```diff
@@ -1,16 +1,19 @@
...
 
+from poetry.utils._compat import Path  # noqa
+from poetry.utils.helpers import canonicalize_name
 from poetry.utils.helpers import module_name
 
 
-TESTS_DEFAULT = u"""from {package_name} import __version__
-
+if TYPE_CHECKING:
+    from typing import Optional
```
finswimmer, line 14: There is no need to import it only for type_checking, is it?

poetry/layouts/layout.py
```diff
@@ -70,6 +76,30 @@ def __init__(
...
 
+    @property
+    def basedir(self):  # type: () -> Path
+        return Path()
+
+    @property
+    def package_path(self):  # type: () -> Path
+        return self.basedir / self._package_path_relative
+
+    def get_package_include(self):  # type: () -> Optional[InlineTable]
```
finswimmer, line 87: `InlineTable` should be in quotes, because it's imported only for type checking.

finswimmer (changes requested):
I have some minor points about type hinting :)

abn:
Good ol' type hinting. Fixed :)

abn:
@finswimmer can you have a look at this one again please?

sdispater (approved):
Looks good to me 👍 

However, it will have to wait the `1.1` release before being merged (I put it into the `1.2` project).

abn:
Sigh :( I hoped to get this in this release. 1.2 it is!

abn:
@sdispater this is ready again; CI might have failures that should be fixed once #3900 is merged.

stephsamson (approved):

Issue #2756, opened 2020-07-31
Poetry says that "environment seems to be broken" all the time


keddad:
- [x] I am on the [latest](https://github.com/python-poetry/poetry/releases/latest) Poetry version.
- [x] I have searched the [issues](https://github.com/python-poetry/poetry/issues) of this repo and believe that this is not a duplicate.
- [x] If an exception occurs when executing a command, I executed it again in debug mode (`-vvv` option).

- **OS version and name**: Fedora 32
- **Poetry version**: 1.0.10
- **Link of a [Gist](https://gist.github.com/) with the contents of your pyproject.toml file**: [Link](https://gist.github.com/keddad/3d34e7a03ee1e87b12518599962f4651)
- I've also done `poetry config virtualenvs.path .venv`

## Issue

Apparently, poetry invalidates virtual environment even if literally nothing happened to it. Here is an example where I install a package and and try to list packages, but it says that environment is broken and installs it again: [Gist](https://gist.github.com/keddad/d46026eeeb8142687b4ebfd22ca70c0d). My only idea is that PyCharm, which is configured to use the venv, messes something up but it shouldn't write anything in venv, I guess.

finswimmer:
@keddad I encountered this problem, when the venv is activated and poetry detect it as broken and than try to recreate it. So make sure your venv is deactivated and run `poetry install` once again. Afterwards everything should work normal.

keddad:
Sadly, that didn't really work in my case. I'm clearly not running any venvs:
```
[keddad@localhost socialcomp]$ which python
/usr/bin/python
```

but `poetry install && poetry show` still results in "broken environment" issues.

keddad:
Ok, that's clearly an issue with how `VirtualEnv.is_sane()` works. It simply checks if there are python and pip present:
```
def is_sane(self): # 
        # A virtualenv is considered sane if both "python" and "pip" exist.
        return os.path.exists(self._bin("python")) and os.path.exists(self._bin("pip"))
```
[
github.com/python-poetry/poetry/blob/master/poetry/utils/env.py#L1168](
github.com/python-poetry/poetry/blob/master/poetry/utils/env.py#L1168)

Since after `poetry install` my venv contains there files, it is probably looking for them in the wrong spot.

keddad:
[...]

Issue #1770, opened 2019-12-19
About environment mangement and .venv folder


azdkj532:
- [x] I have searched the [issues](https://github.com/python-poetry/poetry/issues) of this repo and believe that this is not a duplicate.

## Issue

I found that poetry prefer using `.venv` folder inside as the virtual environment of project. But this behavior is not described in doc. I think it will be good to write down this.

finswimmer:
Hello @azdkj532,

this is definitly not the default behaviour. I guess you have set in in your config, either global or local, sometime before :)

fin swimmer

etijskens:
I am surprised by the answer of fin swimmer
this is my config.
> poetry config --list
cache-dir = "/Users/etijskens/Library/Caches/pypoetry"
virtualenvs.create = true
virtualenvs.in-project = true
virtualenvs.path = "{cache-dir}/virtualenvs"  # /Users/etijskens/Library/Caches/pypoetry/virtualenvs
I see no reference that '.venv' is the default name...

azdkj532:
@finswimmer Could you have a look on this ? May it's a bug.
```
~/Workspace/plurkcore(master ✗) poetry env info

Virtualenv
Python:         2.7.17
Implementation: CPython
Path:           /home/larry/.cache/pypoetry/virtualenvs/plurkcore-xdkz89K8-py2.7
Valid:          True

System
Platform: linux
OS:       posix
Python:   /home/larry/.pyenv/versions/2.7.17
~/Workspace/plurkcore(master ✗) virtualenv .venv
New python executable in /home/larry/Workspace/plurkcore/.venv/bin/python2.7
Also creating executable in /home/larry/Workspace/plurkcore/.venv/bin/python
Installing setuptools, pip, wheel...
done.
~/Workspace/plurkcore(master ✗) poetry env info

Virtualenv
Python:         2.7.17
Implementation: CPython
Path:           /home/larry/Workspace/plurkcore/.venv
Valid:          True

System
Platform: linux
OS:       posix
Python:   /home/larry/.pyenv/versions/2.7.17
~/Workspace/plurkcore(master ✗)

```

azdkj532:
It seems that it doesn't respect the configuration.

finswimmer:
@etijskens: You have `virtualenvs.in-project = true` in your config. `poetry` then creates the venv within the project and calls the folder `.venv`. The name is hardcoded.

[...]

PR #2771, opened 2020-08-05
Respect virtualenvs.in-project setting, if .venv is found in project folder

finswimmer:
At the moment, when there is a folder called `.venv` within the project folder, `poetry` will use it as the path to the venv regardless whether `virtualenvs.in-project` is set to `true` or not. This leads to confusions as described in #1770 and #2756.

With this PR poetry will check if  `virtualenvs.in-project` is set to `true` if it finds a `.venv` folder.

Closes: #1770
Closes: #2756

### DIFF
--- a/docs/docs/configuration.md
+++ b/docs/docs/configuration.md
@@ -114,7 +114,7 @@ Defaults to `false`.
 
 If set to `true`, the virtualenv wil be created and expected in a folder named `.venv` within the root directory of the project.
 
-If set to `None`, `poetry` will use the virtualenv from the `.venv` directory only if available. 
+If set to `None`, `poetry` will use the virtualenv from the `.venv` directory only if available.
 
 ### `virtualenvs.path`: string
 
--- a/docs/docs/configuration.md
+++ b/docs/docs/configuration.md
@@ -112,6 +112,10 @@ Defaults to `true`.
 Create the virtualenv inside the project's root directory.
 Defaults to `false`.
 
+If set to `true`, the virtualenv wil be created and expected in a folder named `.venv` within the root directory of the project.
+
+If set to `None`, `poetry` will use the virtualenv from the `.venv` directory only if available. 
+
 ### `virtualenvs.path`: string
 
 Directory where virtual environments will be created.
--- a/docs/docs/configuration.md
+++ b/docs/docs/configuration.md
@@ -114,7 +114,7 @@ Defaults to `false`.
 
 If set to `true`, the virtualenv wil be created and expected in a folder named `.venv` within the root directory of the project.
 
-If set to `None`, `poetry` will use the virtualenv from the `.venv` directory only if available.
+If not set explicitly, `poetry` will use the virtualenv from the `.venv` directory only if available.
 
 ### `virtualenvs.path`: string
 
--- a/docs/docs/configuration.md
+++ b/docs/docs/configuration.md
@@ -33,7 +33,7 @@ which will give you something similar to this:
 ```toml
 cache-dir = "/path/to/cache/directory"
 virtualenvs.create = true
-virtualenvs.in-project = false
+virtualenvs.in-project = null
 virtualenvs.path = "{cache-dir}/virtualenvs"  # /path/to/cache/directory/virtualenvs
 ```
 
@@ -110,11 +110,12 @@ Defaults to `true`.
 ### `virtualenvs.in-project`: boolean
 
 Create the virtualenv inside the project's root directory.
-Defaults to `false`.
+Defaults to `None`.
 
 If set to `true`, the virtualenv wil be created and expected in a folder named `.venv` within the root directory of the project.
 
-If not set explicitly, `poetry` will use the virtualenv from the `.venv` directory only if available.
+If not set explicitly (default), `poetry` will use the virtualenv from the `.venv` directory when one is available. If set to `false`, `poetry` will ignore any existing `.venv` directory.
+
 
 ### `virtualenvs.path`: string
 
--- a/poetry/utils/env.py
+++ b/poetry/utils/env.py
@@ -17,17 +17,16 @@ from typing import Optional
 from typing import Tuple
 from typing import Union
 
-import tomlkit
-
-from clikit.api.io import IO
-
 import packaging.tags
+import tomlkit
 import virtualenv
 
+from clikit.api.io import IO
 from packaging.tags import Tag
 from packaging.tags import interpreter_name
 from packaging.tags import interpreter_version
 from packaging.tags import sys_tags
+
 from poetry.core.semver import parse_constraint
 from poetry.core.semver.version import Version
 from poetry.core.version.markers import BaseMarker
@@ -1018,8 +1017,8 @@ class SystemEnv(Env):
         # on some distributions it does not return the proper paths
         # (those used by pip for instance). We go through distutils
         # to get the proper ones.
-        from distutils.core import Distribution
         from distutils.command.install import SCHEME_KEYS  # noqa
+        from distutils.core import Distribution
 
         d = Distribution()
         d.parse_config_files()
--- a/poetry/utils/env.py
+++ b/poetry/utils/env.py
@@ -347,9 +347,10 @@ class EnvManager(object):
         if not in_venv or env is not None:
             # Checking if a local virtualenv exists
             if (cwd / ".venv").exists() and (cwd / ".venv").is_dir():
-                venv = cwd / ".venv"
+                if self._poetry.config.get("virtualenvs.in-project") is not False:
+                    venv = cwd / ".venv"
 
-                return VirtualEnv(venv)
+                    return VirtualEnv(venv)
 
             create_venv = self._poetry.config.get("virtualenvs.create", True)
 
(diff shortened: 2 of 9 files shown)


poetry/utils/env.py
```diff
@@ -348,9 +347,10 @@ def get(self, reload=False):  # type: (bool) -> Env
         if not in_venv or env is not None:
             # Checking if a local virtualenv exists
             if (cwd / ".venv").exists() and (cwd / ".venv").is_dir():
-                venv = cwd / ".venv"
+                if self._poetry.config.get("virtualenvs.in-project"):
```
abn, line 350: Should we handle `None` as "use detected environment"?

tests/installation/test_chef.py
```diff
@@ -1,4 +1,5 @@
 from packaging.tags import Tag
+
```
abn, line 2: Is this a related change? Same as the ones below.

poetry/utils/env.py
finswimmer: Didn't know that we can set it to `None` :) As @sdispater suggested in https://github.com/python-poetry/poetry/issues/1770#issuecomment-589642728 I could change it to detect `.venv` in case `in-project` is set to `true` or `None`. But than I guess I will update the docs as well, to the behavior clear.

tests/installation/test_chef.py
finswimmer: I didn't introduce these changes intentional. This is what `isort` has done. Don't know why.

poetry/utils/env.py
abn: It's the case when it was not explicitly set right? Unless we ensure that not set means true by default.

tests/installation/test_chef.py
abn: Weird. I guess isort ran outside of pre-commit.

finswimmer: hm, yes. I'm used to run `isort` manually. Now I've run it via pre-commit and at least the tests passed here. But I have the strong feeling, that the pre-commit version doesn't respect the configuration in the `pyproject.toml`. Can you confirm this?

abn: Hmm, it might be that we never ran `isort` on the entire code base after the config was changed. It might be a good idea to do a PR with `isort` run on the whole codebase.

tests/utils/test_env.py
```diff
@@ -74,17 +74,35 @@ def test_virtualenvs_with_spaces_in_their_path_work_as_expected(tmp_dir, manager
...
+
+    venv_folder = Path(poetry.file.parent) / ".venv"
+    venv_folder.mkdir()
+    poetry.config.config["virtualenvs"]["in-project"] = False
+
+    venv = manager.get()
+
+    assert venv.path != venv_folder
+
+    shutil.rmtree(str(venv_folder))
```
abn, line 105:
```suggestion
    shutil.rmtree(venv_folder.as_posix())
```

tests/utils/test_env.py
```diff
@@ -74,17 +74,35 @@ def test_virtualenvs_with_spaces_in_their_path_work_as_expected(tmp_dir, manager
...
+
+
+def test_env_get_venv_with_venv_folder_present(manager, poetry):
+    if "VIRTUAL_ENV" in os.environ:
+        del os.environ["VIRTUAL_ENV"]
+
+    venv_folder = Path(poetry.file.parent) / ".venv"
+    venv_folder.mkdir()
+    poetry.config.config["virtualenvs"]["in-project"] = False
```
abn, line 99: Can we not merge these two together with an additional parameter?

docs/docs/configuration.md
```diff
@@ -112,6 +112,10 @@ Defaults to `true`.
 Create the virtualenv inside the project's root directory.
 Defaults to `false`.
 
+If set to `true`, the virtualenv wil be created and expected in a folder named `.venv` within the root directory of the project.
+
+If set to `None`, `poetry` will use the virtualenv from the `.venv` directory only if available.
```
abn, line 117:
```suggestion
If not set explicitly, `poetry` will use the virtualenv from the `.venv` directory only if available.
```

tests/utils/test_env.py
finswimmer: Hm, would this work on windows?

tests/utils/test_env.py
finswimmer: I thought about this too, but decided to keep them separated to show the different behavior more explicit.

docs/docs/configuration.md
finswimmer: This might be confusing, because the default is `false`. One could interpret "not set explicitly" as "default not changed".

tests/utils/test_env.py
abn: Lets leave this for now; but in theory yes it should since windows is POSIX compliant. And the use of `str()` if i recall can have some issues in various environments. Although, recently I have discovered using `pathlib2`'s `as_posix` method in `<=3.5` has other issues too. Cannot win!

tests/utils/test_env.py
abn: Would be great if we were making the test suites easier to maintain by removing code duplication. But not a show stopper at the moment.

docs/docs/configuration.md
abn: The ens user should never set `None`. By default it is `false` because when we attempt to get the config we get `None` because it is not set. This distinction is something we should handle internally hence the suggestion.

tests/utils/test_env.py
abn, line 105:
```suggestion
@pytest.fixture
def in_project_venv_dir(poetry):
    os.environ.pop("VIRTUAL_ENV", None)
    venv_dir = poetry.file.parent.joinpath(".venv")
    venv_dir.mkdir()

    try:
        yield venv_dir
    finally:
        venv_dir.rmdir()


@pytest.mark.parametrize("in_project", [True, False, None])
def test_env_get_venv_with_venv_folder_present(manager, poetry, in_project_venv_dir, in_project):
    poetry.config.config["virtualenvs"]["in-project"] = in_project
    venv = manager.get()

    if in_project is False:
        assert venv.path != in_project_venv_dir
    else:
        assert venv.path == in_project_venv_dir
```

Something like this might be cleaner.

abn (changes requested):
Minor suggestions. Otherwise looks good.

poetry/utils/env.py
```diff
@@ -347,9 +347,10 @@ def get(self, reload=False):  # type: (bool) -> Env
         if not in_venv or env is not None:
             # Checking if a local virtualenv exists
             if (cwd / ".venv").exists() and (cwd / ".venv").is_dir():
-                venv = cwd / ".venv"
+                if self._poetry.config.get("virtualenvs.in-project") is not False:
```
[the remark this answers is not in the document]
finswimmer: Make sense.

tests/utils/test_env.py
finswimmer: I like it :+1:

docs/docs/configuration.md
```diff
@@ -112,6 +112,10 @@ Defaults to `true`.
 Create the virtualenv inside the project's root directory.
 Defaults to `false`.
```
abn, line 113: Should have caught this earlier, we default to `unset` correct?

docs/docs/configuration.md
```diff
@@ -112,6 +112,10 @@ Defaults to `true`.
...
 
+If set to `true`, the virtualenv wil be created and expected in a folder named `.venv` within the root directory of the project.
+
+If not set explicitly, `poetry` will use the virtualenv from the `.venv` directory only if available.
```
abn, line 117:
```suggestion
If not set explicitly, `poetry` will use the virtualenv from the `.venv` directory when one is available. If set to `false`, `poetry` will ignore any existing `.venv` directory.
```

abn (commented):
Missed the documentation in the last pass.

docs/docs/configuration.md
finswimmer: No, the default is still `false`. (At least I haven't changed it). And in my opinion we shouldn't change this behavior.

abn: Well the issue is that with this change we break existing behaviour. Currently if default (`false`), and a `.venv` directory is found `poetry` uses it. This behaviour can be replicated by configuring default to to be `None`  with the new changes. And hence why I reckon we should update the documentation and default value to reflect this. Otherwise, we end up creating new venvs where this "misfeature" was relied upon.

finswimmer: While it is correct that this PR changes the current behavior and it could be replicate by setting to `None`, I don't think we should change the default value. Before this PR the behavior was surprising to many people and now it behave like the most people would expect. So it's more like a fix and not like a new feature.

But it's definitely something that should be stated out clearly in the release notes.

abn: The problem with that is that the current users who a) have not explicitly ever set it to `true` or `false`, and b) relies on existing `.venv` directories in their projects will suddently end up creating new virtual environments after upgrade. I do not think this is acceptable. Changing the default however, ensures that this case is covered for. And there is no difference for the other scenarios.

abn (approved):

Issue #2034, opened 2020-02-13
Add @generated to poetry.lock for Phabricator support


lovesegfault:
- [x] I have searched the [issues](https://github.com/python-poetry/poetry/issues) of this repo and believe that this is not a duplicate.
- [x] I have searched the [documentation](https://python-poetry.org/docs/) and believe that my question is not covered.

## Feature Request

The company I work at uses Phabricator internally, which in turn uses the `@generated` string to detected whether a file shouldn't be diffed due to being generated by a tool. Would it be possible to add a header with that text to `poetry.lock`?

I'll note that there is some precedent, at least in Rust's Cargo, to do this: https://github.com/rust-lang/cargo/issues/6180

lovesegfault:
cc. @finswimmer @sdispater

vishalkuo:
I'd be happy to tackle this if there's no traction

PR #2773, opened 2020-08-05
adds @generated to header of lockfile

vishalkuo:
# Pull Request Check List

Resolves: #2034 

It's useful to both developer tools and for users to recognize the lockfile is generated. This header should achieve both

- [ ] Updated **documentation** for changed code. (Not sure if applicable)


Secrus:
Hi, are you still interested in contributing this to Poetry? If so, please merge the master branch and fix conflicts. This would allow for a review to be done.

vishalkuo:
Rebased, please take a look

src/poetry/packages/locker.py
```diff
@@ -397,6 +398,7 @@ def set_lock_data(self, root: Package, packages: list[Package]) -> bool:
             del package["files"]
 
         lock = document()
+        lock.add(comment("@" + "generated"))
```
neersighted, line 401: Let's make this a longer form comment -- something like:

```toml
# This file is automatically @generated by Poetry and should not be changed by hand.
```

neersighted (changes requested):

src/poetry/packages/locker.py
neersighted: Also, let's store that string in a constant, then import it in the tests instead of duplicating it.

src/poetry/packages/locker.py
```diff
@@ -49,7 +49,11 @@
 
 logger = logging.getLogger(__name__)
 _GENERATED_IDENTIFIER = "@" + "generated"
-GENERATED_COMMENT = f"This file is automatically {_GENERATED_IDENTIFIER} by Poetry and should not be changed by hand."
+GENERATED_COMMENT = (
+    f"This file is automatically {_GENERATED_IDENTIFIER} by Poetry and should not be"
+    " changed by hand."
+)
```
danieleades, line 55: this might be overkill. You can probably hard code this (in one place) as a single string. Certainly it's not necessary to construct the identifier from two strings

```python
GENERATED_COMMENT = (
    f"This file is automatically @generated by Poetry and should not be"
    " changed by hand."
)
```

vishalkuo: In my experience, you don't want the `@generated` string in your source since it causes tooling to not show changes to the source file itself

danieleades: ah gotcha

src/poetry/packages/locker.py
neersighted: Yeah, I think I'm going to ask that you please pull in the string from the constant instead of repeating it -- both because of the need to not have a literal `@generated` as well as in the name of DRY I think it's better. Other tests do include hardcoded values, but in some cases that's the only way, and in others it just needs refactoring.

tests/packages/test_locker.py
```diff
@@ -193,7 +199,7 @@ def test_locker_properly_loads_nested_extras(locker: Locker):
 python-versions = "*"
 
 [package.dependencies]
-c = {version = "^1.0", optional = true}
+c = {{version = "^1.0", optional = true}}
```
neersighted, line 202: What's going on here with these extra braces?

vishalkuo: This is how you escape `f` string formatting:
```
In [1]: my_str = "{hello world} {bar}"

In [2]: bar = "baz"

In [3]: my_str = f"{hello world} {bar}"
  Input In [3]
    my_str = f"{hello world} {bar}"
           ^
SyntaxError: f-string: invalid syntax


In [4]: my_str = f"{{hello world}} {bar}"

In [5]: print(my_str)
{hello world} baz
```

tests/packages/test_locker.py
```diff
@@ -81,7 +81,9 @@ def test_lock_file_data_is_ordered(locker: Locker, root: ProjectPackage):
     with locker.lock.open(encoding="utf-8") as f:
         content = f.read()
 
-    expected = """\
+    expected = f"""\
+# {GENERATED_COMMENT}
```
neersighted, line 85: Looks like flake8 isn't happy with the explicit `#` here -- not sure off the top of my head how to resolve it, but we'll need to solve that to merge this.

vishalkuo: fixed

vishalkuo:
hi, any updates on this?

neersighted (approved):

marcorichetta:
Hi! I had to do some digging to find out which version of poetry a lockfile was generated with.

Is there a possibility to add the Poetry version in lockfile metadata?

Related: https://github.com/python-poetry/poetry/discussions/6763

PR #2780, opened 2020-08-08
docs: add mention of editable installation

fredrikaverpil:
This adds a mention about editable installs for path installations, as discussed in https://github.com/python-poetry/poetry/issues/34

### DIFF
--- a/docs/docs/dependency-specification.md
+++ b/docs/docs/dependency-specification.md
@@ -117,6 +117,12 @@ my-package = { path = "../my-package/" }
 my-package = { path = "../my-package/dist/my-package-0.1.0.tar.gz" }
 ```
 
+The package is installed as an "editable installation" by default. This can be specified more explicitly:
+
+```toml
+my-package = { path = "../my-package/", develop = true }
+```
+
 
 ## `url` dependencies
 


docs/docs/dependency-specification.md
```diff
@@ -117,6 +117,12 @@ my-package = { path = "../my-package/" }
 my-package = { path = "../my-package/dist/my-package-0.1.0.tar.gz" }
 ```
 
+The package is installed as an "editable installation" by default. This can be specified more explicitly:
```
abn, line 120: The conext of tis might be confusing since we have an example of a file dependency directly above, which is not installed in a "development mode" but rather instaled as an sdist.  

I am also wondering if, within poetry (outside current pr scope), we should use "development mode" consistently as opposed to "editable install".

fredrikaverpil: Hm, I agree it can be confusing with that sdist. I realized now that by adding another code block, it makes the docs less "lightweight".

How about this:

When specifying a directory, the package is installed in editable/development mode by default. This can be specified explicitly:

```toml
my-package = { path = "../my-package/", develop = true }
```

fredrikaverpil: About editable/development; doesn't pip call it --editable ?
It is my understanding that there actually is something completely else which is referred to "development mode": https://docs.python.org/dev/library/devmode.html

vjpr: Is this accurate? When I install a path package in 1.1.0b3, it does not automatically add it as an editable install. The only way is to add `develop = true`.

fredrikaverpil: With Poetry 1.0.10, I have added the entries manually into `pyproject.toml` and then I don't have to add `develop = true` for poetry to always use the latest commit on master. Maybe this behavior changed?

vjpr: Maybe manual entries work, but `poetry add ../foo` does not.

fredrikaverpil: Would be good to know if this behavior was changed with Poetry 1.1. I might need to test this out actually. I feel like the default behavior should not change.

finswimmer: The change of the default behavior was introduced by https://github.com/python-poetry/poetry-core/commit/72c803cbe5086ab7cdfa644319690d2f5a9fb7bb. After talking to @sdispater it  turned out, that this change was intended. So we must adopt the docs!

finswimmer:
Hello,

this PR is superseded by #2987.

Thanks a lot for your contributions :+1: 

fin swimmer

Issue #2789, opened 2020-08-10
Poetry build does not install build dependencies (affects both install and build)


mdgoldberg:
- [x] I am on the [latest](https://github.com/python-poetry/poetry/releases/latest) Poetry version.
- [x] I have searched the [issues](https://github.com/python-poetry/poetry/issues) of this repo and believe that this is not a duplicate.
- [x] If an exception occurs when executing a command, I executed it again in debug mode (`-vvv` option).

- **OS version and name**: OS X 10.15.5 (also tested in docker on ubuntu 16.04)
- **Poetry version**: 1.1.0b2
- **Link of a [Gist](https://gist.github.com/) with the contents of your pyproject.toml file**: https://gist.github.com/mdgoldberg/96a9955c8c823f2127dbc471a954b898

## Issue

I created a repo with a minimal reproduction: https://github.com/mdgoldberg/poetry-cython-example

The problem here is that Cython isn't installed prior to building/installing the root package, even though Cython is specified in `build-system.requires` in the pyproject.toml. Here's how I am reproducing the issues, for both `poetry install` and `poetry build`:
```bash
$ git clone git@github.com:mdgoldberg/poetry-cython-example.git
$ cd poetry-cython-example
$ poetry install -vvv
Creating virtualenv poetry-cython-example in /Users/mattgoldberg/poetry-cython-example/.venv
Using virtualenv: /Users/mattgoldberg/poetry-cython-example/.venv
Installing dependencies from lock file

Finding the necessary packages for the current system

Installing the current project: poetry-cython-example (0.1.0)
  - Building package poetry-cython-example in editable mode
  - Executing build script: build.py
Traceback (most recent call last):
  File "/Users/mattgoldberg/poetry-cython-example/build.py", line 6, in <module>
    from Cython.Build import cythonize
ModuleNotFoundError: No module named 'Cython'
  - Adding poetry_cython_example.pth to /Users/mattgoldberg/poetry-cython-example/.venv/lib/python3.6/site-packages for /Users/mattgoldberg/poetry-cython-example
  - Adding the poetry_cython_example-0.1.0.dist-info directory to /Users/mattgoldberg/poetry-cython-example/.venv/lib/python3.6/site-packages
$
$ poetry build -f wheel -vvv
Building poetry-cython-example (0.1.0)
  - Building wheel
Traceback (most recent call last):
  File "build.py", line 6, in <module>
    from Cython.Build import cythonize
ModuleNotFoundError: No module named 'Cython'

[...]

PR #2794, opened 2020-08-10
Support build system requirements

abn:
This change adds support of build requirements. This is implemented such that build requirements are always installed on `poetry install` and `poetry install --no-dev` since the package installation will require `build-system.requires` packages. When `poetry build` is executed, existing behaviour is retained for cases where additional build requirements are not required. If additional, dependencies are required (eg: `cython`) these and only these are installed before build and environment executable is configured for use when build script is executed. 

Resolves: #2789

Requires: #2732 python-poetry/poetry-core#58


mdgoldberg:
Any updates here @abn ? Is this still being developed?

abn:
@mdgoldberg @schneiderfelipe I am guessing there is still interest in this? I have had some push backs from core about this. But happy to push it forward.

schneiderfelipe:
> @mdgoldberg @schneiderfelipe I am guessing there is still interest in this? I have had some push backs from core about this. But happy to push it forward.

Very much interested. Out of curiosity, what's holding this back?

espdev:
@abn,
Support of build dependencies is very much needed. Please see my last comment:
https://github.com/python-poetry/poetry/issues/2789#issuecomment-955717907

GuillaumeDesforges:
@abn are you still on this? Would be very helpful :slightly\_smiling\_face:

abn:
@GuillaumeDesforges guess now I am :D unfortunately https://github.com/python-poetry/poetry-core/pull/58 got a bit stuck. Let me have a rethink with the new groups changes. The appraoch I took there will no longer really work. Since "dev" is not necesarily a group that is always going to be present.

That said, in the interim I would recommend you redefine the dependency like so.

```sh
poetry add --group build 'cpython@*'
```

Or use the TOML.

```toml
[tool.poetry.group.build.dependencies]
cpython = "*"
```

Personally, I do not like that I'd have to re-specify the dependency, but that is where it is at present.

abn:
Closing this in favour of #5401 for now as the core issue was stuck without concensus.

Issue #2125, opened 2020-03-03
poetry config not working for HTTP auth so unable to pull private package


viveksair:
- [x] I am on the [latest](https://github.com/python-poetry/poetry/releases/latest) Poetry version.
- [x] I have searched the [issues](https://github.com/python-poetry/poetry/issues) of this repo and believe that this is not a duplicate.
- [ ] If an exception occurs when executing a command, I executed it again in debug mode (`-vvv` option).

- **OS version and name**: 
``` 
debian:10 
root@91ca4e71776b:/# uname -a
Linux 91ca4e71776b 4.9.93-linuxkit-aufs #1 SMP Wed Jun 6 16:55:56 UTC 2018 x86_64 GNU/Linux
```
- **Poetry version**:
```
poetry --version
Poetry version 1.0.3
```
- **Link of a [Gist](https://gist.github.com/) with the contents of your pyproject.toml file**: 
```
  
[tool.poetry]
name = "foo_bar"
version = "0.1.0"
description = ""
authors = ["Your Name <you@example.com>"]

[tool.poetry.dependencies]
python = "^3.7"
private_package = "^1.0.13"


[tool.poetry.dev-dependencies]



[tool.poetry.scripts]


[[tool.poetry.source]]
name = "foo"
url = "https://foo.bar/simple"
default = true		

[build-system]
requires = ["poetry>=1.0.3"]
build-backend = "poetry.masonry.api"
```

## Issue

```
$ poetry config repositories. foo https://foo.bar/simple
$ poetry config $USER_ID $USER_PASSWORD
$ poetry add private_package

[ValueError]
Could not find a matching version of package private_package

```
and I tried with env as well ```POETRY_HTTP_BASIC_FOO_USERNAME POETRY_HTTP_BASIC_FOO_PASSWORD``` but not working.

but `poetry add private_package` working after mounting config & password at ``` ~/.config/pypoetry/auth.toml & ~/.config/pypoetry/config.toml ```

problem with mounting config is throwing user and password when error occurred while `poetry install` and I have to add `default = true` to `[[tool.poetry.source]]` to my  pyproject.toml

4F2E4A2E:
Thank you poetry for being awesome and thanks @viveksair for creating this issue.

I also can confirm this bug.

**Python version: 3.6.8
Poetry Version: 1.0.5**
**Install state**: fresh install, everything from scratch
[...]

PR #2796, opened 2020-08-11
Log message on non 200 responses from repository

bneijt:
# Pull Request Check List

Helps with: #2125

- [x] Updated **documentation** for changed code. (no update needed)

I have not taken the time to add testing to the configuration. I would like to get feedback before putting in more effort as the change seems so obvious that there must be a reason why it has not been done earlier and that reason might devalue the whole PR.

### DIFF
--- a/poetry/repositories/legacy_repository.py
+++ b/poetry/repositories/legacy_repository.py
@@ -1,4 +1,5 @@
 import cgi
+import logging
 import re
 import warnings
 
@@ -29,6 +30,8 @@ from .auth import Auth
 from .exceptions import PackageNotFound
 from .pypi_repository import PyPiRepository
 
+
+logger = logging.getLogger(__name__)
 
 try:
     import urllib.parse as urlparse
@@ -359,10 +362,22 @@ class LegacyRepository(PyPiRepository):
 
         return data.asdict()
 
-    def _get(self, endpoint):  # type: (str) -> Union[Page, None]
+    def _get(self, endpoint):  # type: (str) -> Optional[Page]
         url = self._url + endpoint
         response = self.session.get(url)
         if response.status_code == 404:
             return
+        if 400 <= response.status_code < 500:
+            logger.warning(
+                "Request failed with status {} for url {}.".format(
+                    response.status_code, url
+                )
+            )
+        elif response.status_code != 200:
+            logger.debug(
+                "Request failed with non 200 status code ({}) for url {}.".format(
+                    response.status_code, url
+                )
+            )
 
         return Page(url, response.content, response.headers)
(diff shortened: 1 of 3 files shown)


poetry/repositories/legacy_repository.py
```diff
@@ -359,10 +359,13 @@ def _get_release_info(self, name, version):  # type: (str, str) -> dict
...
         return data.asdict()
 
-    def _get(self, endpoint):  # type: (str) -> Union[Page, None]
+    def _get(self, endpoint):  # type: (str) -> Optional[Page]
         url = self._url + endpoint
         response = self.session.get(url)
         if response.status_code == 404:
             return
-
+        if response.status_code != 200:
```
abn, line 367: This could cause some unintended side-effects. I would suggest we log a debug message for now as a start.

bneijt: 1) Adding debug seems a to low level, maybe warning on 401 and debug otherwise?
2) Does adding debug logging require a test?
3) How could we get a clear picture on the possible unintended side-effects, maybe we could find and fix those?

abn: 1. If the code is `>=400 and <500`, warning/info should be okay I guess, for everything else a debug should be fine. However, the intention here would be to assist in debugging why something does not work, this would mean executing the command with `-v` if it fails.
2. If it is trivial, I would say just post the cli output of your manual verification in the PR description could do the trick. A test would be great though, because you can use pytest to intercept and veriffy logging along with mocking a response. This allows us to catch any regression later on.
3. That might be difficult difficult; since legacy repository implementations are very wild west unfortunately. The side-effect might be at run-time for end-users using a non-standard repository. For example, we have seen different response codes for success from artifactory/azure private indices etc. Hence why I am reluctant to raise an error on `!=200`. We do not want to break any working scenarios.

bneijt: 1) Will update my PR to reflect 1 (warning for 400 to 500 range) and debug otherwise
2) Ok, will add a mock test to cover the log line
3) Ah, shame. I can understand it's a wild west, maybe we should introduce configuration to pick the backend implementation of an source at some point or maybe we could directly hook into some pip code for repository handling at a later date.

I will work on 1 and 2 and resolve this discussion at the time those changes are in the PR. Thank you for responding.

abn:
This might need a rebase.

bneijt:
There are enough issues with the private repositories being fixed that this will probably be fixed in another pull request or bug fix, so I'm closing this PR because it's probably to little to late to actually help anymore.

abn:
Apoligies for dropping the ball on this one. Feel free to ping on discord if something falls through the crack again. We appricate your contributions, issues reports, PRs and doc fixes alike.

Issue #2803, opened 2020-08-13
search has no --only-name (-N) option


bphunter1972:
- [x] I have searched the [issues](https://github.com/python-poetry/poetry/issues) of this repo and believe that this is not a duplicate.

## Issue

https://python-poetry.org/docs/cli/#search says that the search command has an option, --only-name, but it does not.

```
▶ poetry help search
USAGE
  poetry search <tokens1> ... [<tokensN>]

ARGUMENTS
  <tokens>               The tokens to search for.

GLOBAL OPTIONS
  -h (--help)            Display this help message
  -q (--quiet)           Do not output any message
  -v (--verbose)         Increase the verbosity of messages: "-v" for normal output, "-vv" for more verbose output and "-vvv" for debug
  -V (--version)         Display this application version
  --ansi                 Force ANSI output
  --no-ansi              Disable ANSI output
  -n (--no-interaction)  Do not ask any interactive question
```

```
▶ poetry search --only-name

[NoSuchOptionException]
The "--only-name" option does not exist.

```
<img width="836" alt="image" src="https://user-images.githubusercontent.com/5446342/90097847-c0e6d500-dceb-11ea-802c-df19db48a878.png">

finswimmer:
Good catch @bphunter1972 :+1:

PR #2804, opened 2020-08-13
remove outdated `--only-name` search option from docs

finswimmer:
This PR removes outdated `--only-name` search option from docs.

# Pull Request Check List

Resolves: https://github.com/python-poetry/poetry/issues/2803

- [x] Updated **documentation** <strike>for changed code.</strike>

### DIFF
--- a/docs/docs/cli.md
+++ b/docs/docs/cli.md
@@ -413,10 +413,6 @@ This command searches for packages on a remote index.
 poetry search requests pendulum
 ```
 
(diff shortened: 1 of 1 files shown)


abn:
Hmm looks like ignoring builds and "expected" checks are incompatible. :(

abn:
@finswimmer created a workaround that still makes things quicker but works around the above issue. See #2805.

Issue #2807, opened 2020-08-14
Poetry cannot resolve dependency from Qiskit Aer 0.6.0


ikkoham:
- [x] I am on the [latest](https://github.com/python-poetry/poetry/releases/latest) Poetry version.
- [x] I have searched the [issues](https://github.com/python-poetry/poetry/issues) of this repo and believe that this is not a duplicate.
- [x] If an exception occurs when executing a command, I executed it again in debug mode (`-vvv` option).

- **OS version and name**: macOS 10.15.6 
- **Poetry version**: 1.0.10 and prerelease

## How to reproduce?

`poetry add qiskit-aer==0.6.0` to the blank env.

## Issue

`poetry add qiskit-aer==0.5.2` can resolve the dependency and install required packages.
However, from qiskit-aer==0.6.0, the dependency cannot be resolved, so poetry installs only qiskit-aer.

Related issue: https://github.com/Qiskit/qiskit-aer/issues/880

ikkoham:
Caused by https://github.com/Qiskit/qiskit-aer/pull/775/files

abn:
When poetry inspects this package the `sdist` tarball gets downloaded. This is unpacked and inspected. The `PKG-INFO` contains the following.

```
Metadata-Version: 2.1
Name: qiskit-aer
Version: 0.6.1
Summary: Qiskit Aer - High performance simulators for Qiskit
Home-page: https://github.com/Qiskit/qiskit-aer
Author: AER Development Team
Author-email: qiskit@us.ibm.com
License: Apache 2.0
Description: # Qiskit Aer
<-- SNIPPED -->
Keywords: qiskit aer simulator quantum addon backend
Platform: UNKNOWN
Classifier: Environment :: Console
Classifier: License :: OSI Approved :: Apache Software License
Classifier: Intended Audience :: Developers
Classifier: Intended Audience :: Science/Research
Classifier: Operating System :: Microsoft :: Windows
Classifier: Operating System :: MacOS
Classifier: Operating System :: POSIX :: Linux
Classifier: Programming Language :: C++
Classifier: Programming Language :: Python :: 3 :: Only
Classifier: Programming Language :: Python :: 3.5
Classifier: Programming Language :: Python :: 3.6
Classifier: Programming Language :: Python :: 3.7
Classifier: Programming Language :: Python :: 3.8
Classifier: Topic :: Scientific/Engineering
Description-Content-Type: text/markdown
```

[...]

PR #2810, opened 2020-08-14
info: enable PEP 517 fallback builds by default

abn:
This change ensures that PEP 517 fallback builds are enabled by default
for when metadata inspection and setup.{py,cfg} parsing fails. This is
required in order to ensure that package requirements are correctly
identified for packages beyond first level dependencies.

Resolves: #2807
Resolves: https://github.com/python-poetry/poetry/issues/2850

### DIFF
--- a/poetry/inspection/info.py
+++ b/poetry/inspection/info.py
@@ -488,15 +488,16 @@ class PackageInfo:
 
     @classmethod
     def from_directory(
-        cls, path, allow_build=False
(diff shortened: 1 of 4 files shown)


finswimmer (approved):
LGTM :+1: 

Because this changes default behavior I let it to @sdispater to merge.

abn:
Merging this for now. We can revert it later if this causes issues.

Issue #1165, opened 2019-06-13
pip should be listed as a system requirement


lochsh:
- [x] I have searched the [issues](https://github.com/sdispater/poetry/issues) of this repo and believe that this is not a duplicate.

## Issue

My understanding is that pip is a system requirement for poetry.  See my comment here of poetry failing to work without pip: https://github.com/sdispater/poetry/issues/612#issuecomment-501652698

The [current system requirements ](https://poetry.eustace.io/docs/#system-requirements) only list Python as a system requirement.  Although the need for pip might seem obvious given poetry installs Python packages, I believe it should be explicitly listed as a requirement.

Although many standard Python installations come with Pip, installing Python 2 from source does not (I think this has changed with Python 3).  I don't think it should be assumed that a Python installation being present on a machine means a pip installation will be too. At least not until Python 2 is officially unsupported (I know it's not long! Using Python 2 was not my choice.).

Do let me know if I've misunderstood anything.

abn:
The only scenario where `pip` is externally managed is when Poetry is configured to use the system envrionment and not create a project virtual environment.

In `1.0.z` releases under Python 2.7 this `pip` is ensured by `virtualenv` and under Python 3 `pip` installation is ensured via `ensurepip` that is called by the `venv` built-in module. With `1.1.0` this becomes unified with the use of `virutalenv` and its embeded `pip` versions (#2666).

The custom installer can be used without issues in an environment that does not have pip. See below.

```console
$ docker run --rm -it fedora:latest bash -c '\
>     type pip || :; \
>     type pip3 || :; \
>     python3 -m pip || :;\
>     curl -sLO https://raw.githubusercontent.com/python-poetry/poetry/master/get-poetry.py;\
>     python3 get-poetry.py --preview --yes; \
>     source $HOME/.poetry/env; \
>     poetry --version; \
>     poetry new foobar; \
>     pushd foobar; \
>     poetry install'
bash: line 1: type: pip: not found
bash: line 2: type: pip3: not found
/usr/bin/python3: No module named pip
Retrieving Poetry metadata

# Welcome to Poetry!

This will download and install the latest version of Poetry,
a dependency and package manager for Python.

It will add the `poetry` command to Poetry's bin directory, located at:

$HOME/.poetry/bin

This path will then be added to your `PATH` environment variable by
modifying the profile files located at:

$HOME/.profile
$HOME/.bash_profile

You can uninstall at any time by executing this script with the --uninstall option,
[...]

PR #2818, opened 2020-08-16
added pip requirment for config virtualenv.create=False

KGB33:
Resolves: #1165

- [x] Added **tests** for changed code. -- NA - no code changed
- [x] Updated **documentation** for changed code. -- NA - no code changed

### DIFF
--- a/docs/docs/configuration.md
+++ b/docs/docs/configuration.md
@@ -107,6 +107,10 @@ Defaults to one of the following directories:
 Create a new virtual environment if one doesn't already exist.
 Defaults to `true`.
 
+!!!note:
+	While `pip` is not a system requirment for poetry when installed via the provided installer,
+	`pip` is required if this is set to `false`.
+
 ### `virtualenvs.in-project`: boolean
 
 Create the virtualenv inside the project's root directory.
--- a/docs/docs/configuration.md
+++ b/docs/docs/configuration.md
@@ -108,8 +108,8 @@ Create a new virtual environment if one doesn't already exist.
 Defaults to `true`.
 
 !!!note:
-	While `pip` is not a system requirment for poetry when installed via the provided installer,
-	`pip` is required if this is set to `false`.
+        When setting this configuration to `false`, the Python environment used must have `pip` 
+        installed and available.
 
 ### `virtualenvs.in-project`: boolean
 
--- a/docs/docs/index.md
+++ b/docs/docs/index.md
@@ -6,7 +6,7 @@ It allows you to declare the libraries your project depends on and it will manag
 
 ## System requirements
 
-Poetry requires Python 2.7 or 3.5+ and pip. It is multi-platform and the goal is to make it work equally well
+Poetry requires Python 2.7 or 3.5+. It is multi-platform and the goal is to make it work equally well
 on Windows, Linux and OSX.
 
 !!! note


abn:
This should no longer be required since #2666 right? The embedded pip from `virtualenv` is used.

KGB33:
For installing project dependencies that makes sense but how does poetry install it's `_vendor` packages? 

The `_vendor` folder in the (latest) release on Github is empty.

abn:
From `poetry-1.1.0b2-linux.tar.gz` the `poetry/_vendor` direcotry is already populated with the vendored dependencies.

```console
$ ls poetry/_vendor
py2.7  py3.5  py3.6  py3.7  py3.8  py3.9
```

At the moment, the `get-poetry.py` script does not rely on `pip` or any external dependnecies other than a valid python version. Note that the vendor directory is only valid for when installed using the custom installer (`get-poetry.py`). For installation using `pip` or `pipx` dependencies are managed by the package management frontend.

KGB33:
Gotcha. I checked the source code zip and just incorrectly assumed the `_vender` folder was un-populated  for all of them.

abn:
@KGB33 this still has value; but maybe make it clear that this is only required for cases where the project is using the system environment (`poetry config virtualenvs.create false`). And maybe this is better off in the "Configuration" section of the docs.

KGB33:
@abn Sorry for the delay, I've added a simple note to the configuration docs. I was thinking adding an error message might be helpful as well, however, I haven't gotten a docker image to work for me quite yet.

KGB33:
Got the docker image to work for me and the error message is very clear. 

```
Skipping virtualenv creation, as specified in config file.

Package operations: 71 installs, 0 updates, 0 removals

  • Installing pycparser (2.20)

  EnvCommandError

  Command ['/usr/local/bin/python3', '-m', 'pip', 'install', '--no-deps', '/root/.cache/pypoetry/artifacts/37/8e/5a/0ea4f84bc7f11e0e3468110efa2c7783241ea7eaa63a92a751de06f78f/pycparser-2.20-py2.py3-none-any.whl'] errored with the following return code 1, and output:
  /usr/local/bin/python3: No module named pip


  at /poetry_install/libutils/env.py:915 in _run
       911│                 output = subprocess.check_output(
       912│                     cmd, stderr=subprocess.STDOUT, **kwargs
       913│                 )
       914│         except CalledProcessError as e:
    →  915│             raise EnvCommandError(e, input=input_)
       916│
       917│         return decode(output)
       918│
       919│     def execute(self, bin, *args, **kwargs):
```

docs/docs/configuration.md
```diff
@@ -107,6 +107,10 @@ Defaults to one of the following directories:
 Create a new virtual environment if one doesn't already exist.
 Defaults to `true`.
 
+!!!note:
+	While `pip` is not a system requirment for poetry when installed via the provided installer,
+	`pip` is required if this is set to `false`.
```
abn, line 112:
```suggestion
        When setting this configuration to `false`, the Python environment used must have `pip` 
        installed and available.
```

Rephrasing it a bit to make it clear(er) - I think. :)

KGB33:
What about adding why. Something like:

```
!!!note:
        When setting this configuration to `false`, the Python environment used must have `pip` 
        installed and available because Poetry cannot use the vendored `virtualenv`. 

````

abn (approved):

Issue #2819, opened 2020-08-17
SolverProblemError with papermill


AssassinTee:
## Information

- [x] I am on the [latest](https://github.com/python-poetry/poetry/releases/latest) Poetry version.
- [x] I have searched the [issues](https://github.com/python-poetry/poetry/issues) of this repo and believe that this is not a duplicate.
- [ ] If an exception occurs when executing a command, I executed it again in debug mode (`-vvv` option).

- **OS version and name**: Ubuntu 18 LTS
- **Poetry version**: 1.0.10 (and 1.0.5, too)
- **Link of a [Gist](https://gist.github.com/) with the contents of your pyproject.toml file**: Not required

## Issue
When installing [papermill](https://github.com/nteract/papermill) with poetry i got the following error message:

```
[SolverProblemError]
Because no versions of papermill match >2.1.2,<3.0.0
 and papermill (2.1.2) depends on black (*), papermill (>=2.1.2,<3.0.0) requires black (*).
So, because no versions of black match *
 and **myframework** on papermill (^2.1.2), version solving failed.
```

It looks like it failed to find any version of [black](https://pypi.org/project/black/), which is weird.

## workaround

I just installed black by hand and papermill after that with:

```
poetry add black
poetry add papermill
```

This resolved my issue but I leave it here for the poetry devs and other users running into this issue

abn:
@AssassinTee can you try the latest release (`1.0.10`) or pre-release (`1.1.0b2`) please? This seems to be resolved.
```console
$ poetry --version
Poetry version 1.0.10
$ poetry add black
Using version ^19.10b0 for black

Updating dependencies
Resolving dependencies... (0.1s)

Writing lock file


Package operations: 8 installs, 0 updates, 0 removals

  - Installing appdirs (1.4.4)
  - Installing attrs (19.3.0)
  - Installing click (7.1.2)
  - Installing pathspec (0.8.0)
  - Installing regex (2020.7.14)
  - Installing toml (0.10.1)
  - Installing typed-ast (1.4.1)
  - Installing black (19.10b0)
$ poetry add papermill
Using version ^2.1.2 for papermill

Updating dependencies
Resolving dependencies... (0.4s)

Writing lock file


Package operations: 27 installs, 0 updates, 0 removals

  - Installing decorator (4.4.2)
  - Installing ipython-genutils (0.2.0)
  - Installing six (1.15.0)
  - Installing pyrsistent (0.16.0)
[...]

PR #2821, opened 2020-08-17
Handle (*) constraint for pre-release only packages

abn:
This change ensures that packages that only have a pre-release versions
are pragmatically discovered via `find_packages` when a constraint (*)
is specified.

Resolves: #2819

# Pull Request Check List


abn:
Merging this for now. We can revert it later if this causes issues.

Issue #1584, opened 2019-11-15
Poetry export does not include setuptools even when it is a direct dependency.


apollo13:
- [x] I am on the [latest](https://github.com/sdispater/poetry/releases/latest) Poetry version.
- [x] I have searched the [issues](https://github.com/sdispater/poetry/issues) of this repo and believe that this is not a duplicate.
- [x] If an exception occurs when executing a command, I executed it again in debug mode (`-vvv` option).

- **OS version and name**: Fedora 31
- **Poetry version**: 1.0.0b4 (at the time of writing b5 is not on PyPi, will retry later on)
- **Link of a [Gist](https://gist.github.com/) with the contents of your pyproject.toml file**: https://gist.github.com/apollo13/1d0c72978531d3a5688f1e3847fa6fe6

## Issue
Using the `pyproject.toml` file from the Gist and executing `poetry lock` I get the following lockfile:
```
[[package]]
category = "main"
description = "WSGI HTTP Server for UNIX"
name = "gunicorn"
optional = false
python-versions = ">=3.4"
version = "20.0.0"

[package.dependencies]
setuptools = ">=3.0"

[package.extras]
eventlet = ["eventlet (>=0.9.7)"]
gevent = ["gevent (>=0.13)"]
setproctitle = ["setproctitle"]
tornado = ["tornado (>=0.2)"]

[metadata]
content-hash = "20c8a1cc7d149b3ad58ca5063dfd37654f03ba65ee8b484d2faefe7ce28b7daf"
python-versions = "^3.7"

[metadata.files]
gunicorn = [
    {file = "gunicorn-20.0.0-py2.py3-none-any.whl", hash = "sha256:0806b5e8a2eb8ba9ac1be65d7b743ec896fc25f5d6cb16c5e051540157b315bb"},
    {file = "gunicorn-20.0.0.tar.gz", hash = "sha256:ef69dea4814df95e64e3f40b47b7ffedc6911c5009233be9d01cfd0d14aa3f50"},
]
```
setuptools is correctly discovered as dependency of gunicorn (see https://github.com/benoitc/gunicorn/blob/94ab2091173c6037b504f94e56f4e88816d540bf/setup.py#L71-L77 -- it does indeed require it), but the lockfile *does not contain* any hashes for setuptools.
If I now export to requirements.txt the file looks like this:
```
gunicorn==20.0.0 \
[...]

PR #2826, opened 2020-08-17
Drop poetry's dependency on pip and setuptools in project virtual environment

abn:
An initial attempt at using the embedded pip wheel from virtualenv package directly. This avoids the need for any critical packages in the project's virtual environment.

Resolves: #1584

## Testing this pull request
Since this is a fundamental shift in how we setup virtual environments, testing a broad range of use cases would be great.

### Using pipx
```sh
pipx install --suffix=@2826 'poetry @ git+https://github.com/python-poetry/poetry.git@refs/pull/2826/head'
```

Once this is done, you can use this version of poetry as shown here.
```sh
poetry@2826 new foobar
pushd foobar
poetry@2826 add pycowsay
poetry@2826 run pycowsay
```
If you want to do a `pip list`, there exists a convinience wrapper that will use the embedded `pip` if the environment does not contain one.

```console
$ poetry@2826 run pip list
Package        Version
-------------- -------
attrs          20.3.0
more-itertools 8.7.0
packaging      20.9
pluggy         0.13.1
py             1.10.0
pycowsay       0.0.0.1
pyparsing      2.4.7
pytest         5.4.3
wcwidth        0.2.5
```

You will notice here that none of `wheel`, `setuptools` nor `pip` exists in the virtual environment.

```sh
set -xe
python -m pip install -q git+https://github.com/python-poetry/poetry.git@refs/pull/2826/head
poetry new foobar
pushd foobar
poetry add pycowsay
poetry run pip list
EOF
```

### DIFF
--- a/poetry/installation/pip_installer.py
+++ b/poetry/installation/pip_installer.py
@@ -2,6 +2,7 @@ import os
 import tempfile
 import urllib.parse
 
+from pathlib import Path
 from subprocess import CalledProcessError
 from typing import TYPE_CHECKING
 from typing import Any
@@ -9,14 +10,15 @@ from typing import Union
 
 from cleo.io.io import IO
 
+from poetry.installation.base_installer import BaseInstaller
 from poetry.core.pyproject.toml import PyProjectTOML
 from poetry.repositories.pool import Pool
 from poetry.utils._compat import encode
 from poetry.utils.env import Env
+from poetry.utils.env import EnvManager
+from poetry.utils.env import VirtualEnv
 from poetry.utils.helpers import safe_rmtree
-
-from .base_installer import BaseInstaller
-
+from poetry.utils.helpers import temporary_directory
 
 if TYPE_CHECKING:
     from poetry.core.packages.package import Package
@@ -193,7 +195,7 @@ class PipInstaller(BaseInstaller):
         else:
             req = os.path.realpath(package.source_url)
 
-        args = ["install", "--no-deps", "-U"]
+        args = ["install", "--no-deps", "-U", "--root", str(self._env.path)]
 
         pyproject = PyProjectTOML(os.path.join(req, "pyproject.toml"))
 
@@ -239,7 +241,11 @@ class PipInstaller(BaseInstaller):
 
         args.append(req)
 
-        return self.run(*args)
+        with temporary_directory() as tmp_dir:
+            venv_dir = Path(tmp_dir) / ".venv"
+            EnvManager.build_venv(venv_dir.as_posix(), with_pip=True)
+            venv = VirtualEnv(venv_dir, venv_dir)
+            return venv.run("pip", *args)
 
     def install_git(self, package: "Package") -> None:
         from poetry.core.packages.package import Package
--- a/poetry/installation/pip_installer.py
+++ b/poetry/installation/pip_installer.py
@@ -10,15 +10,15 @@ from typing import Union
 
 from cleo.io.io import IO
 
-from poetry.installation.base_installer import BaseInstaller
 from poetry.core.pyproject.toml import PyProjectTOML
+from poetry.installation.base_installer import BaseInstaller
 from poetry.repositories.pool import Pool
 from poetry.utils._compat import encode
 from poetry.utils.env import Env
-from poetry.utils.env import EnvManager
-from poetry.utils.env import VirtualEnv
 from poetry.utils.helpers import safe_rmtree
-from poetry.utils.helpers import temporary_directory
+from poetry.utils.pip import pip_editable_install
+from poetry.utils.pip import pip_install
+
 
 if TYPE_CHECKING:
     from poetry.core.packages.package import Package
@@ -190,12 +190,12 @@ class PipInstaller(BaseInstaller):
 
         from poetry.factory import Factory
 
+        req: Path
+
         if package.root_dir:
             req = (package.root_dir / package.source_url).as_posix()
         else:
-            req = os.path.realpath(package.source_url)
-
-        args = ["install", "--no-deps", "-U", "--root", str(self._env.path)]
+            req = Path(package.source_url).resolve(strict=False)
 
         pyproject = PyProjectTOML(os.path.join(req, "pyproject.toml"))
 
@@ -230,22 +230,16 @@ class PipInstaller(BaseInstaller):
 
                 with builder.setup_py():
                     if package.develop:
-                        args.append("-e")
-
-                    args.append(req)
-
-                    return self.run(*args)
+                        return pip_editable_install(
+                            directory=req, environment=self._env
+                        )
+                    return pip_install(
+                        path=req, environment=self._env, deps=False, upgrade=True
+                    )
 
         if package.develop:
-            args.append("-e")
-
-        args.append(req)
-
-        with temporary_directory() as tmp_dir:
-            venv_dir = Path(tmp_dir) / ".venv"
-            EnvManager.build_venv(venv_dir.as_posix(), with_pip=True)
-            venv = VirtualEnv(venv_dir, venv_dir)
-            return venv.run("pip", *args)
+            return pip_editable_install(directory=req, environment=self._env)
+        return pip_install(path=req, environment=self._env, deps=False, upgrade=True)
 
     def install_git(self, package: "Package") -> None:
         from poetry.core.packages.package import Package
--- a/poetry/utils/env.py
+++ b/poetry/utils/env.py
@@ -1087,6 +1087,9 @@ class Env(object):
         return True
 
     def run(self, bin: str, *args: str, **kwargs: Any) -> Union[str, int]:
+        if bin == "pip":
+            return self.run_pip(*args, **kwargs)
+
         bin = self._bin(bin)
         cmd = [bin] + list(args)
         return self._run(cmd, **kwargs)
@@ -1131,6 +1134,9 @@ class Env(object):
         return decode(output)
 
     def execute(self, bin: str, *args: str, **kwargs: Any) -> Optional[int]:
+        if bin == "pip":
+            return self.run_pip(*args, **kwargs)
+
         bin = self._bin(bin)
 
         if not self._is_windows:
--- a/poetry/utils/env.py
+++ b/poetry/utils/env.py
@@ -30,6 +30,7 @@ from packaging.tags import Tag
 from packaging.tags import interpreter_name
 from packaging.tags import interpreter_version
 from packaging.tags import sys_tags
+from virtualenv.seed.wheels.embed import get_embed_wheel
 
 from poetry.core.semver.helpers import parse_constraint
 from poetry.core.semver.version import Version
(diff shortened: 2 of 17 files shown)


mikenerone:
_Some_ sort of approach to let me lock `pip`, `setuptools`, and `wheel` for my own project is pretty important to me. Either removing Poetry's dependence on them as this branch does, or a config syntax that lets me bless overrides of Poetry's norms, or _something_. Right now, my team's dev environments are broken by default because the installed deps don't match the requirements in the package metadata. We workaround it for the moment with `poetry run pip install ...` (and the versions we need seem to work fine for Poetry), but it would be much, much better if these package deps could be treated per normal.

gsemet:
We are dealing with exactly the same issue using pipenv, it always wants the latest version even if it breaks something. Please ensure at least to let the ability of user to freeze the version somehow. Ideally yes, it would use its own, independent setuptools/wheel, but for some complex projets that needs certain version of setuptools, especially with the version 50.x breaking much stuff with pyinstaller packaging, I am looking to a better alternative than `p[ipenv|etry] run pip install 'setuptools<50'`

abn:
@gsemet can you provide a bit more context please? This change only means that poetry's use is isolated to specific versions of pip/setuptools and decoupled. If your project depends on it, you can specify it as either a `dev-dependency` or `build-system.requires`. PEP 517 builds still take place in isolated environments, this means the `build-system.requires` of the project will be respected.

gsemet:
Yes, this change is actually what i would to see. I am using poetry on my newer projects and does not face *yet* this issue I have on older projects still using pipenv. Pipenv always updates to the latest setuptools available on pypi within the project's virtualenv (also `pip` and `wheels` packages), and it occurs several times that the new version broke something. For reproductibility, there should not be these "moving" parts.

So yes, I am pretty much very interesting is seeing the right behavior in poetry: poetry either vendors its own setuptools/wheel package so that they change with poetry updates, but if my package wants some specific version of setuptools for any reason, it should not cause any incompatibility with poetry. From out point of view, `setuptools`, `pip` and `wheels` are 3 packages like any others.

abn:
@gsemet ah, I think I misunderstood your first message. Glad that this is something that is useful. However, do note that `poetry` today treats `setuptools`, `pip` and `wheel` as "unsafe" packages in the environment. This means that, while, we do not explicitly update these versions, they cannot be managed by project. 

This is what I (putting on my poetry user hat), want to change in this PR.

Poetry (putting on my maintainer hat again), has been gradually reducing reliance on `pip` and `setuptools`. Right now, we only use `pip` for wheel installations (wheels poetry downloads) and editable installs `pip install -e /path/to/something`. The reliance on `setuptools` is a side-effect of supporting editable dependency installations.

gsemet:
So, I hope this PR lands in the next version of poetry :)

apollo13:
I just saw this PR. This would be a huge improvement for us if it ends up fixing #1584 -- Thank you for your work on this @abn

poetry/installation/pip_installer.py
```diff
@@ -189,7 +193,7 @@ def install_directory(self, package):
         else:
             req = os.path.realpath(package.source_url)
 
-        args = ["install", "--no-deps", "-U"]
+        args = ["install", "--no-deps", "-U", "--root", str(self._env.path)]
```
apollo13, line 196: I tried to trigger this condition because I think that `--root` might be wrong (the rest in this change correctly uses `--prefix`) but I could not figure out how to trigger `install_directory`. Even when I have:
```
django = {path = "django.git", develop = false}
```
it does not do a directory install. Do you have any tips for me here?

poetry/installation/pip_installer.py
```diff
@@ -232,18 +236,21 @@ def install_directory(self, package):
...
                     return self.run(*args)
 
         if package.develop:
-            args.append("-e")
+            return pip_editable_install(req, self._env)
 
         args.append(req)
 
-        return self.run(*args)
+        with temporary_directory() as tmp_dir:
```
apollo13, line 249: I think this could use `pip_install` from the helpers, no?

poetry/utils/env.py
```diff
@@ -679,19 +681,38 @@ def create_venv(
 
     @classmethod
     def build_venv(
-        cls, path, executable=None
-    ):  # type: (Union[Path,str], Optional[Union[str, Path]]) -> virtualenv.run.session.Session
+        cls,
+        path,
+        executable=None,
+        with_pip=False,
```
apollo13, line 687: I think `create_venv` should now explicitly call `build_venv` with `with_wheel=False, with_setuptools=False` since those are no longer needed in the venv and would be uninstalled?

poetry/utils/env.py
```diff
@@ -679,19 +681,38 @@ def create_venv(
...
+            "--no-download",
+            "--no-periodic-update",
+            "--python",
+            executable or sys.executable,
+        ]
+
+        if not with_pip:
+            opts.append("--no-pip")
+        else:
+            if with_wheel is None:
```
apollo13, line 704: What is the rationale for this?

apollo13 (commented):
@abn Looking mostly good so far. I have left a few question, would be glad if you could get back on those.

poetry/installation/pip_installer.py
abn: @apollo13 can you try with latest changes please?

poetry/installation/pip_installer.py
abn: Correct. Fixed.

poetry/utils/env.py
abn: Added a comment for this as we need to handle cases where pip is enabled and wheel has not been explicitly disabled.

poetry/utils/env.py
abn: Probably a good thing; but doing so will break so many of the tests as they are currently fragile in that we test called with. We can do this in a later change I think.

apollo13 (approved):
I have tested against my original issue description, as well as the questions I have asked in this PR and so far it looks good. Great work!

stephsamson (approved):

sdispater (approved):
Looks good to me 👍 

That being said, I really want us to move away from subprocesses call to `pip` to install packages. I've started work on it but it's no small task.

apollo13:
> That being said, I really want us to move away from subprocesses call to pip to install packages. I've started work on it but it's no small task.

Can you tell me how you imagine installation without pip? Using pip imo ensures that the installation is always as intended and one does not have to patch poetry if pip's installation behavior changes etc… I wonder if a custom installation is worth it.

sdispater:
@apollo13 Just so you know, there are already discussions among the PyPA maintainers to provide standalone libraries to install packages without relying on `pip`.

Basically, today, installing a Python package is as "simple" as:

- If you have a wheel, install it following existing guidelines (there is a PEP mentioning it)
- If you have an sdist, build a wheel via a PEP-517 backend (or default to `setuptools`) and install the wheel following the existing guidelines

apollo13:
@sdispater Thanks, I know that installing wheels at least is relatively easy. But as with many things, the devil is probalby in the details. If there is a standalone library to install stuff (split out of pip), then by all means, using it makes sense. I am just a little bit afraid that without such a library the implementation might run into subtle bugs etc…

abn:
> * If you have an sdist, build a wheel via a PEP-517 backend

This is a solved problem and we should really use https://github.com/pypa/pep517

Related: #2983

> That being said, I really want us to move away from subprocesses call to `pip` to install packages. I've started work on it but it's no small task.

This change is in-line with that move, since we abstracting away the pip install calls, I suspect the next step is get implementation for wheel install done, combine with `pep517` library to do the build (if sdist).

sdispater:
@abn 

> This is a solved problem and we should really use https://github.com/pypa/pep517

Well, not really, since `pep517` makes subprocess calls to `pip` (https://github.com/pypa/pep517/blob/master/pep517/envbuild.py#L98) which we want to avoid. It also doesn't have reusable environments, i.e. you need to create a brand new environment for each sdist which is costly.

That's why I said it's no small task to make it `pip` agnostic.

apollo13:
It seems as if https://github.com/pypa/build would be a `pep517` compatible installer that doesn't utilize pip at all -- at least that is what a quick look through the source suggested.

abn:
At present it does use `pip`, see https://github.com/pypa/build/blob/2c1da830f214580e8e4b94e4c88345963fa32d2e/src/build/env.py#L157-L170

That said, if we do have support for wheel installations, and access to `setuptools` wheels (bundled via `virtualenv`), I suspect doing so will only require poetry to override the `build` lib implementation to use poetry's own wheel installation. But I can see what you mean by it not being a simple task. Although, I would definitely have to ask if that is worthwhile for poetry to implement and manage. Additionally, the "new environment" for each `sdist` is probably a good thing to be honest as it ensures build isolation. I doubt that will be an issue if you have local caching available.

TBBle:
You don't actually _override_ pypa/build's env-setup, since the _caller_ is the one calling it, i.e. pip is only sub-called by `IsolatedEnvBuilder.install` from [`_build_in_isolated_env` in `__main__.py`](https://github.com/pypa/build/blob/2c1da830f214580e8e4b94e4c88345963fa32d2e/src/build/__main__.py#L49-L59) which is part of the CLI for the pypa/build package:
```
def _build_in_isolated_env(builder, outdir, distributions, config_settings):
    # type: (ProjectBuilder, str, List[str], ConfigSettings) -> None
    for distribution in distributions:
        with IsolatedEnvBuilder() as env:
            builder.python_executable = env.executable
            builder.scripts_dir = env.scripts_dir
            # first install the build dependencies
            env.install(builder.build_dependencies)
            # then get the extra required dependencies from the backend (which was installed in the call above :P)
            env.install(builder.get_dependencies(distribution))
            builder.build(distribution, outdir, config_settings)
```

So you just want to create a venv with the necessary dependencies yourself, and do this (`_build_in_current_env`from [`__main__.py`](https://github.com/pypa/build/blob/2c1da830f214580e8e4b94e4c88345963fa32d2e/src/build/__main__.py#L62-L73), this is the `--no-isolation` path for their CLI tool)
```
def _build_in_current_env(builder, outdir, distributions, config_settings, skip_dependencies=False):
    # type: (ProjectBuilder, str, List[str], ConfigSettings, bool) -> None
    for dist in distributions:
        if not skip_dependencies:
            missing = builder.check_dependencies(dist)
            if missing:
                _error(
                    'Missing dependencies:'
                    + ''.join('\n\t' + dep for deps in missing for dep in (deps[0], _format_dep_chain(deps[1:])) if dep)
                )

        builder.build(dist, outdir, config_settings)
```

So for Poetry, I expect that we just populate a venv as we do now, and use pypa/build's [`ProjectBuilder(srcdir)`](https://github.com/pypa/build/blob/2c1da830f214580e8e4b94e4c88345963fa32d2e/src/build/__init__.py#L137) (`builder` above, this wraps/exposes the PEP-517 hooks) to build the wheel we needed using that venv.

This also puts Poetry directly in control of whether it uses a single scratch venv to build all necessary wheels for the "real" project venv, or uses an isolated venv for each wheel.

bmw:
I'm eagerly awaiting this change as a resolution to https://github.com/python-poetry/poetry/issues/1584. When do people expect this PR to be included in a release?

abn:
@bmw the next planned release is `1.2.0a0`, which should be landing in the next month or so. In the meantime, you can use the version from master.

```sh
pipx install --force --suffix=@master 'poetry @ git+https://github.com/python-poetry/poetry.git@master'
```

Issue #2179, opened 2020-03-13
Ability to specify the directory to run poetry in instead of always using current directory


svermeulen:
- [x ] I have searched the [issues](https://github.com/python-poetry/poetry/issues) of this repo and believe that this is not a duplicate.
- [ x] I have searched the [documentation](https://python-poetry.org/docs/) and believe that my question is not covered.

## Feature Request

I'd like to be able to run `poetry run X` while also setting the current directory to whatever I want.  Right now, this appears to be impossible, since I have to set the current directory to wherever my `pyproject.toml` file is located instead.  Can there be a command line option to override this default?

svermeulen:
I did find the following workaround which works
```
#!/bin/sh
STARTDIR=`pwd`
cd TOML_DIR
poetry run bash -c "cd $STARTDIR && python -m foo $*"
```
Where TOML_DIR is the path to the directory with your `pyproject.toml`

bphunter1972:
This works well if your module has an `if __name__ == '__main__'`, which isn't precisely what will get created with a [tool.poetry.scripts] specification.

noamraph:
Just in case it helps someone, here's the script I used. It finds TOML_DIR from the location of the executable, so it doesn't need to be hardcoded. It also uses a function, like [tool.poetry.scripts].

```
#!/bin/sh
STARTDIR=$(pwd)
TOML_DIR=$(dirname "$0")
cd "$TOML_DIR" || exit
poetry run bash -c "cd $STARTDIR && python -c 'from foo.foo import main; main()' $*"
```

abn:
I am a bit unclear on the issue here. Poetry, similar to git, will detect the `pyproject.toml` as long as the current directory is a child directory.

Here is an example where `poetry run` is executed within the `src` directory in the project.

```console
$ tree .
.
├── poetry.lock
├── pyproject.toml
├── README.rst
├── src
│   ├── poetry
│   │   └── run
│   │       └── context
│   │           └── __init__.py
│   └── poetry_run_context.egg-info
│       ├── dependency_links.txt
│       ├── PKG-INFO
│       ├── SOURCES.txt
│       └── top_level.txt
└── tests
    └── __init__.py

6 directories, 9 files

$ type pytest
bash: type: pytest: not found

$ cd src/

$ poetry run type pytest
pytest is /tmp/poetry-run-context/.venv/bin/pytest
```

abn:
@noamraph something like this will also work.

```toml
[...]

PR #2829, opened 2020-08-19
Update pyproject.md

noamraph:
Explain that `poetry install` installs the scripts in the virtualenv. See #2179 for the use case.

Having a script inside the virtualenv allows it to be run from a different working directory. Both @bphunter1972 and I didn't think about running `poetry install` in order to create the script. So I think that saying this explicitly may help others as well.

### DIFF
--- a/docs/docs/pyproject.md
+++ b/docs/docs/pyproject.md
@@ -216,6 +216,8 @@ poetry = 'poetry.console:run'
 
 Here, we will have the `poetry` script installed which will execute `console.run` in the `poetry` package.
 
+When developing, run `poetry install` after editing the `scripts` section, to install the scripts in the virtualenv.
+
 ## `extras`
 
 Poetry supports extras to allow expression of:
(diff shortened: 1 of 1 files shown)


docs/docs/pyproject.md
```diff
@@ -216,6 +216,8 @@ poetry = 'poetry.console:run'
 
 Here, we will have the `poetry` script installed which will execute `console.run` in the `poetry` package.
 
+When developing, run `poetry install` after editing the `scripts` section, to install the scripts in the virtualenv.
```
abn, line 219:
```suggestion
!!!note

    When a script is added or updated, run `poetry install` to make them available in the project's virtualenv.
```

abn (commented):
Maybe consider making this a note? I think it is important enough to be marked as such.

noamraph:
> Maybe consider making this a note? I think it is important enough to be marked as such.

@abn Sure!

bphunter1972:
Doesn't `poetry install` install the scripts NOT into the virtual environment, but into the user's .local/bin (LINUX)?

More generally, isn't this correct: 

* `poetry shell` creates a shell in the virtual environment which has your libraries and scripts.
* `poetry install` installs the the libraries and scripts into your system, which for me was my .local directory.

That's what worked for me.

finswimmer (approved):

noamraph:
@bphunter1972 at least for me, `poetry install` added scripts to the virtualenv's `bin` directory.

Issue #98, opened 2018-05-08
BUG: When poetry add failed, package is listed as added but is not installed. Also add proper handling for multiple packages install in "poetry add": when one package is already added, others are not getting added.


buriy:
If "poetry add" fails with an error on a package installation,
then after fixing the error it says "already added".
I need to remember to use "poetry install" then instead. 

The same happens if you want to add more than one entry, and the first entry was already installed: it writes that the entry was already added, not doing anything.

And the third issue is that if one issue breaks, "poetry add" doesn't add anything and tries to run an install first!

Fourth, after a "poetry add" fail, I have to add the entry *manually* and then run "poetry install" which writes that "you probably updated your file, run poetry update for that".
That means that while "poetry add" *can* check for the project file updates, "poetry install" can't. That behavior feels wrong.

Suggestions: 
1) remove error "Package A is already present", and when trying to add an existing package, instead, show what is the current version of that package in the project file, like "Package A of version 2.3.1 is selected". Or, maybe, even better it is to replace the already added version instead of displaying current version.
2) on "poetry add", first add *everything* to the project file, only then run install
3) probably, poetry install should be consistent with poetry add, or maybe "poetry add" without arguments should work like "poetry update".

stephsamson:
@buriy Can you provide an example so that I can reproduce these issues on my machine?

buriy:
@stephsamson 
There are several different use cases here. The easiest one is here:
```
poetry add jupyter
poetry add jupyter pydub ```
->
```
[ValueError]           
Package jupyter is already present  
```

buriy:
Here is the recent activity

finswimmer:
I have to investigate it a little bit more. But I guess this it what happens:

When you add  a package and something went wrong during installation and the exception isn't catched properly you and up having the package listed as an dependency in the `pyproject.toml` but it is not listed in the `poetry.lock`.

The simple workaround in this case is to remove the entry from `pyproject.toml` and use `poetry add` again.

buriy:
@finswimmer 
I heard poetry is a tool that helps users, rather than asking doing everything manually ;)
This needs to be fixed.

guiemi:
Same here. Visual Studio Code keeps asking me to install Pylama even if it's already added by Poetry.

jacebrowning:
It's my understanding that the `Package jupyter is already present` error when attempting to `poetry add jupyter` a second time is the expected behavior.

[...]

PR #2886, opened 2020-09-06
Do not raise error if package is already in pyproject.toml on poetry add

finswimmer:
At the moment `poetry` raises a `ValueError` when trying to add a package that is already in the `pyproject.toml`. This PR change the behaviour as follows:

* Do not raise ValueError if a dependency is added via `poetry add` and this dependency is already in `pyproject.toml`.
* Instead just print a warning and skip this package

# Pull Request Check List

Resolves: https://github.com/python-poetry/poetry/issues/98

### DIFF
--- a/poetry/console/commands/add.py
+++ b/poetry/console/commands/add.py
@@ -1,3 +1,4 @@
+# -*- coding: utf-8 -*-
 from typing import Dict
 from typing import List
 
--- a/poetry/console/commands/add.py
+++ b/poetry/console/commands/add.py
@@ -89,13 +89,18 @@ class AddCommand(InstallerCommand, InitCommand):
         if section not in poetry_content:
             poetry_content[section] = {}
 
-        packages = self.remove_existing_packages_from_input(
+        existing_packages = self.get_existing_packages_from_input(
             packages, poetry_content, section
         )
 
+        if existing_packages:
+            self.notify_about_existing_packages(existing_packages)
+
+        packages = [name for name in packages if name not in existing_packages]
+
         if not packages:
             self.poetry.file.write(content)
-            self.line("Nothing to add.", style="info")
+            self.line("Nothing to add.")
             return 0
 
         requirements = self._determine_requirements(
@@ -184,7 +189,7 @@ class AddCommand(InstallerCommand, InitCommand):
 
         return status
 
-    def remove_existing_packages_from_input(
+    def get_existing_packages_from_input(
         self, packages, poetry_content, target_section
     ):  # type: (List[str], Dict, str) -> List[str]
         existing_packages = []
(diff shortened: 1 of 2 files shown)


poetry/console/commands/add.py
```diff
@@ -184,3 +183,21 @@ def handle(self):
...
+                if key.lower() == name.lower():
+                    existing_packages.append(name)
+                    self.line(
+                        "{name} is already in pyproject.toml. Skipping.".format(
+                            name=name
+                        )
+                    )
+        packages = [name for name in packages if name not in existing_packages]
+
+        return packages
```
abn, line 203: Would be great if we can make this a filter to split out existing and new packages so that we can, at the end of the command execution (if not quiet) notify the use that these packages already exists and at version x, and hence were skipped.

The final message should also probably indicate that if they want to update they should use either update or `poetry add package@version syntax`.

sdispater:
I think from a purely UX standpoint, a good approach would be to display some guidance for user, for instance:

```
"package" is already present in the pyproject.toml file.

If you want to update it to the latest compatible version you can use `poetry update package`.
If you prefer to upgrade it to the latest available version you use `poetry add package@latest`.
```

poetry/console/commands/add.py
```diff
@@ -89,13 +89,18 @@ def handle(self):
         if section not in poetry_content:
             poetry_content[section] = {}
 
-        packages = self.remove_existing_packages_from_input(
+        existing_packages = self.get_existing_packages_from_input(
             packages, poetry_content, section
         )
 
+        if existing_packages:
```
abn, line 97: Looking good. Do you reckon it might be better to move the message to the end so that it is more .. visible to the user rather than lost somewhere at the top? Granted people might not add a lot of packages, but if they added one with a lot dependencies the install log could cover up the message.

finswimmer: I prefer having this message at the beginning,  because it saying something like "Ok, I will do it for you, but ..." The paragraph is large enough to be an eye-catcher. So when someone start the process they should recognize that it looks different than normal.

Putting it at the end, I would find myself always asking: "Hm, did it install everything fine".

But at the end it - as always - depends on how people are used to read/scan messages :)

abn: Typically terminals scroll messages up, so I'd expect this to be more in the summary of results from the operation. However, since this is not a major issue, I am not going to block on this.

poetry/console/commands/add.py
```diff
@@ -86,18 +90,19 @@ def handle(self):
...
-                        continue
+        existing_packages = self.get_existing_packages_from_input(
+            packages, poetry_content, section
+        )
 
-                    raise ValueError("Package {} is already present".format(name))
+        if existing_packages:
+            self.notify_about_existing_packages(existing_packages)
+
+        packages = [name for name in packages if name not in existing_packages]
```
abn, line 100: How would this handle scenarios like the following?

```sh
# this works today
poetry add aiohttp^2.0
poetry add aiohttp^3.0
poetry add aiohttp@latest
```

This along with cases where git/url were used the existing packages were simply ignored. The new behaviour might be regression as the requested spec is not being considered.

abn (changes requested):

poetry/console/commands/add.py
finswimmer: As the input is simply compared to the package name in the `pyproject.toml`, this will work. Because there will be no package called `aiohttp^2.0`. There it at least for the `@latest` case a test here: https://github.com/python-poetry/poetry/blob/120cf431905c4e650d8f221d96f6ac7ffa484e52/tests/console/commands/test_add.py#L732-L734

abn (approved):

PR #2899, opened 2020-09-10
handle poetry init non-interactive dependencies

rotu:
1. in non-interactive mode, suppress "This command will guide you through..."
2. in interactive mode, respect command line --dependency if the user chooses NO for "Would you like to define your main dependencies interactively?"


poetry/console/commands/init.py
```diff
@@ -158,26 +159,29 @@ def handle(self):
...
-            help_displayed = True
+            if self.io.is_interactive():
+                self.line(help_message)
+                help_displayed = True
             requirements = self._format_requirements(
                 self._determine_requirements(self.option("dependency"))
             )
-            self.line("")
+            if self.io.is_interactive():
+                self.line("")
```
finswimmer, line 169: I'm not sure if we really need this `self.line("")`. I guess at least it can be placed directly after the last output above and we can avoid this second if-statement.

Same thing applies to line 183-184.

finswimmer (changes requested):

Just a minor change request. 

And can you please also create a test for the non-interactive `poetry init` please?

fin swimmer

poetry/console/commands/init.py
rotu: It serves to separate the interactive output of `_determine_requirements`, so that would change the output. Still want the change?

rotu:
@finswimmer Any more actions needed from me? It looks like the check failed for mysterious reasons unrelated to my development.

finswimmer:
@rotu: Could you please rebase onto master and see what happens to the checks?

rotu:
@finswimmer, backmerged master

finswimmer:
@rotu: Thanks a lot. The tests passes now, but now we have a conflict. :( Can you please resolve it?

Thanks a lot for your contribution!

finswimmer (approved):

PR #2900, opened 2020-09-10
tests: clean up fixtures and console tests

abn:
This change cleans up a few of the fixtures used in the test suite to be more consistent and also renames ambiguous helper classes.

There are a lot of lines/files affected, however there are not functional changes. All of it is refactoring the test suite to rely more on fixtures and unify some redundant code.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -1,5 +1,7 @@
 import os
+import re
 import shutil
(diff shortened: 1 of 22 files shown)


sdispater (approved):
Looks good to me 👍

Issue #2369, opened 2020-04-30
poetry run does not relay exit code


pawamoy:
- [x] I am on the [latest](https://github.com/python-poetry/poetry/releases/latest) Poetry version.
- [x] I have searched the [issues](https://github.com/python-poetry/poetry/issues) of this repo and believe that this is not a duplicate.
---
- **OS version and name**: ArchLinux (ArchLabs) 5.6.7-arch1-1
- **Poetry version**: 1.0.5
- **Link of a [Gist](https://gist.github.com/) with the contents of your pyproject.toml file**: https://github.com/pawamoy/poetry-issue-2369/blob/master/pyproject.toml

## Issue

Poetry does not relay the exit code of a script (as in from Poetry's `scripts` section) when ran with `poetry run`.

To reproduce, run the following snippet or check the repo directly: https://github.com/pawamoy/poetry-issue-2369

```console
$ git clone https://github.com/pawamoy/poetry-issue-2369
$ cd poetry-issue-2369
$ poetry install -v
$ poetry shell
$ poetry-issue-2369
$ echo $?
1
$ exit
$ poetry run poetry-issue-2369
$ echo $?
0
```

Additional information:

```console
$ poetry run which poetry-issue-2369
/home/pawamoy/.cache/pypoetry/virtualenvs/poetry-issue-2369-Q6VvSue2-py3.8/bin/poetry-issue-2369

$ cat /home/pawamoy/.cache/pypoetry/virtualenvs/poetry-issue-2369-Q6VvSue2-py3.8/bin/poetry-issue-2369
#!/home/pawamoy/.cache/pypoetry/virtualenvs/poetry-issue-2369-Q6VvSue2-py3.8/bin/python
#EASY-INSTALL-ENTRY-SCRIPT: 'poetry-issue-2369','console_scripts','poetry-issue-2369'
__requires__ = 'poetry-issue-2369'
import re
import sys
from pkg_resources import load_entry_point

if __name__ == '__main__':
    sys.argv[0] = re.sub(r'(-script\.pyw?|\.exe)?$', '', sys.argv[0])
    sys.exit(
        load_entry_point('poetry-issue-2369', 'console_scripts', 'poetry-issue-2369')()
    )
```

Note how the installed script wraps the entry-point call into `sys.exit(...)`.

It means that, as a best-practice, the entry-point itself should *not* use `sys.exit(...)`.

My main function returns an integer, and it's the `__main__` module that wraps it in `sys.exit(...)`.

[...]

PR #2904, opened 2020-09-11
Relay exit code in poetry run

pawamoy:
# Pull Request Check List

Resolves: #2369 

I'm a bit stuck. While testing the run command, it seems the `tester.io` buffer stays empty. Why is that? The command prints things on stdout, the buffer should get them.

Also, it seems the fix has no effect? If I navigated through the code correctly, the script is ultimately run like this:

```python
return os.execvp("python", [
    "python", "-c",
        "import sys; from importlib import import_module; "
        "sys.argv = ('relay', '2'); sys.exit(import_module('project.main').relay())"
])
```

`os.execvp` [does not return](https://docs.python.org/3/library/os.html#os.execvp), and rather replaces the current process, so anyway the exit code should be 2, and `tester.status_code` stays at 0.

Any idea? :confused:


pawamoy:
Well OK, the fix *does* fix the issue, it's just hard to test it. I noticed the the env has `_execute = False`, so the script is not actually ran, so that explains why the status code is not 2. I tried setting `env._execute = True` and patching `os.execvp` with `subprocess.call`, but the import machinery cannot find the `project` package. Maybe I'd need to install or fake the installation of the package?

I'm a bit lost :sweat_smile: 

But the fix should actually work :slightly_smiling_face:

pawamoy:
I'm using PDM now, not interested anymore, feel free to reopen, reuse the code in another PR, or simply ping me!

Issue #2642, opened 2020-07-07
Strange downgrades when adding new deps.


dekoza:
- [x] I am on the [latest](https://github.com/python-poetry/poetry/releases/latest) Poetry version.
- [x] I have searched the [issues](https://github.com/python-poetry/poetry/issues) of this repo and believe that this is not a duplicate.
- [x] If an exception occurs when executing a command, I executed it again in debug mode (`-vvv` option).

- **OS version and name**: Manjaro Linux, latest
- **Poetry version**: 1.0.9
- **Link of a [Gist](https://gist.github.com/) with the contents of your pyproject.toml file**: https://gist.github.com/dekoza/cb68a98e677c984c1188fc2a634ada41

## Issue
I observed strange downgrade when adding new deps to my project. Aside the fact that whole process takes almost 15 minutes (8 minutes resolving, 7 minutes lock writing; I blame it on `aws-cdk` galaxy), some packages get downgraded. Then, after the install, I run `poetry update` (another 15 minutes) - each downgraded package is upgraded back to latest version.

Both my pyproject.toml and example process is attached in gist link above.

tomzx:
It appears that for the dependencies you've specified, it takes about 10s to do version solving, but that it takes a large amount of time to go through graph building (`Solver._build_graph()`) using `poetry` 1.1.10.

Through a process of adding dependencies, I've identified that `aws-cdk.aws-rds` and `ws-cdk.aws-ecs` (together), `aws-cdk.aws-ecs-patterns` (by itself) seems to be causing some issue. There's over 700k calls  (I had to kill it) to `Solver._build_graph()` which doesn't appear to be normal.

Graph building will need to have better logging facilities to debug this issue :)

abn:
Is this still an issue? With the `pyprojec.toml` provided:

```console
   1: Version solving took 74.620 seconds.
   1: Tried 1 solutions.

Writing lock file
```

Unable to reproduce the downgrades with the version on master. If this is still an issue please comment.

PR #2149, opened 2020-03-08
Changed indirect access to more direct access methods to improve performance

glitchedcat:
When resolving libraries with large dependancy chains such as the "aws_cdk" the `solve` method becomes more and more expensive due to the recursive dependancy resolution. Most of this cost comes from indirect variable access (`getattr()` and `@property`). See the breakdown image below which shows `getattr()` accounting for 30% of the runtime. 
Screenshot 2020-03-08 at 12 22 42
The change to `poetry/packages/dependency_package.py` results in a 30% performance improvement and the changes to `poetry/packages/package.py` result in a further 10% improvement (workings below). 

### Method

1. Install initial 31 dependancies (see list below)
2. Time the addition of a further dependancy `time poetry add aws_cdk.aws_codestar` (see original timings)
3. Time the removal of the dependancy `time poetry remove aws_cdk.aws_codestar` (see original timings)
4. Add `@property` accessors in `poetry/packages/dependency_package.py`
5. Repeat add and remove timing (see V1 timings)
6. Add direct access for `name` variable in `poetry/packages/package.py`
7. Repeat add and remove timing (see V2 timings)

### Timings
*Timings in seconds

Run # | Original add | Original remove | V1 Property Add | V1 Property Remove | V2 Direct Name Add | V2 Direct Name Remove
-- | -- | -- | -- | -- | -- | --
1 | 225.36 | 220.3 | 151.27 | 146.52 | 134.32 | 132.09
2 | 223.19 | 219.73 | 149.62 | 149.45 | 133.40 | 132.49
3 | 223.23 | 220.61 | 148.55 | 147.52 | 134.08 | 131.68
4 | 224.63 | 222.46 | 150.95 | 150.28 | 133.96 | 131.23
5 | 224.72 | 223.38 | 150.42 | 150.00 | 132.97 | 130.42
Average | 224.23 | 221.30 | 150.16 | 148.75 | 133.75 | 131.58

### Initial installed dependancies 
```[tool.poetry.dependencies]
python = "^3.7"
"aws_cdk.aws-accessanalyzer" = "^1.27.0"
"aws_cdk.aws-acmpca" = "^1.27.0"
"aws_cdk.aws-amazonmq" = "^1.27.0"
"aws_cdk.aws-amplify" = "^1.27.0"
"aws_cdk.aws-apigateway" = "^1.27.0"
"aws_cdk.aws-apigatewayv2" = "^1.27.0"
"aws_cdk.aws-appconfig" = "^1.27.0"
[...]

PR #2906, opened 2020-09-11
Improve performance of package ordering/tagging after version solving

zweger:
This PR greatly improves the performance of package ordering/tagging which occurs during package solving, after the version solving itself has been completed. In certain cases, the current algorithm has exponential time complexity and starts to really break down when the package dependency graph is dense. I've added a test case for one such scenario, where the dependencies form a transitive tournament. On `master` this test takes several minutes to run with my hardware, but only takes around 100ms with this patch.

I've noticed on personal projects that it's very easy to run into this issue when using `aws_cdk`.

At the core of the new algorithm is a depth-first search which is used to compute a topological sort of the packages. Using the topological ordering, the longest path / depth of each package can be computed.

Along the way I noticed that the current algorithm does not correctly check for dependency cycles and in certain cases will trigger a `RecursionError`. I've added a test case for this scenario as well.

# Pull Request Check List

This may be related to #2642 and #2149, which both mention using `aws_cdk`.

### DIFF
--- a/poetry/puzzle/solver.py
+++ b/poetry/puzzle/solver.py
@@ -1,4 +1,3 @@
-import abc
 import enum
 import time
 
@@ -170,12 +169,7 @@ class Solver:
                     operations.append(Uninstall(installed))
 
         return sorted(
-            operations,
-            key=lambda o: (
-                -o.priority,
-                o.package.name,
-                o.package.version,
-            ),
+            operations, key=lambda o: (-o.priority, o.package.name, o.package.version,),
         )
 
     def solve_in_compatibility_mode(self, overrides, use_latest=None):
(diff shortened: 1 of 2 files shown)


abn:
@sdispater this might need your review.

dekoza:
Can I do anything to help this patch reach production asap? It's among the top fixes on my "urgently needed" list.

dekoza:
This works like a charm. Dependencies that took 40 minutes in total (20 resolving + 20 locking) are now resolved in 18 seconds. Thank you @zweger !

sdispater (approved):
Great work @zweger!

This part of the code always felt clunky – even though it was doing the job – but I never had the time to improve it so I appreciate that you took the time to do it.

Thanks a lot!

PR #2907, opened 2020-09-11
Update Unix cache directory in the docs

mvoitko:
The default Unix cache directory is currently ~/.poetry


abn:
@mvoitko I don't think this is accurate. 

https://www.github.com/python-poetry/poetry/tree/master/poetry%2Flocations.py

Issue #2572, opened 2020-06-17
Add option to install only dev dependencies


jdinunzio:
- [x] I have searched the [issues](https://github.com/python-poetry/poetry/issues) of this repo and believe that this is not a duplicate.
- [x] I have searched the [documentation](https://python-poetry.org/docs/) and believe that my question is not covered.

## Feature Request

Add an option to `poetry install` to install only dev dependencies, e.g. `poetry install --dev-only`.

Use Case:

Being able to quickly install only dev dependencies in a Continuous Integration environment to run basic quality checks as fast as possible.

As a developer, I run a series of quick checks on my code, like pylint, flake, black, and mypy, both locally and in a CI environment. Locally, those tools can be installed by running `poetry install`  together with the rest of the package's dependencies. 

In a CI environment I would like to run those tests as quickly as possible to minimize the reponse time of the CI pipeline in the detection of problem in the code. Making a full install of all dependencies could be time consuming for certain projects.

Adding an option to install dev dependencies only will simplify the scripts that run those first stages in the CI pipeline.

irfanahm3d:
Likewise. Facing a similar scenario and I feel it would be a great feature to have! 🙂

dannywillems:
Same here

nniehoff:
Even though PR #2917 has been merged the option still does not exist in 1.1.6.

finswimmer:
Hey @nniehoff,

everything that is in master will be available in poetry 1.2. poetry 1.1 will not receive any new feature.

fin swimmer

jose-lpa:
I'm not sure this has been implemented as people expected it.

I understood from all related issues and PRs that people wanted the same as I actually want: be able to install the development dependencies, **but without altering the main dependencies status**.

What this actually does, as far as I can see, is to make sure that **only the development dependencies are installed**. So it actually **removes any other dependency not specified as `dev`**.

Quick check screenshot:

![Screenshot from 2021-06-29 16-16-18](https://user-images.githubusercontent.com/2226098/123823737-5fc0f380-d8f5-11eb-902d-d9ab2a7fecc9.png)

So as you can see the command `poetry install --dev-only` actually left the project unusable, because it removed all core packages just by not being specified as `dev` dependencies.

I'm not sure what people wanted, but I seriously doubt anyone would expect this behavior when issuing a command to "install development dependencies". 

[...]

PR #2917, opened 2020-09-14
Add option to install only dev dependencies (#2572)

opelr:
# Pull Request Check List

Resolves: #2572

### DIFF
--- a/docs/docs/cli.md
+++ b/docs/docs/cli.md
@@ -109,6 +109,13 @@ the `--no-dev` option.
 poetry install --no-dev
 ```
 
+Conversely, you can specify to the command that you only want to install the development dependencies
+by passing the `--dev-only` option. Note that `--no-dev` takes priority if both options are passed.
+
+```bash
+poetry install --dev-only
+```
+
 If you want to remove old dependencies no longer present in the lock file, use the
 `--remove-untracked` option.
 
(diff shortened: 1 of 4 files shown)


TylerLubeck:
Hi! Is there anything I can do to help get this merged? It would be hugely helpful for me

finswimmer (changes requested):
Thanks a lot for your contribution @opelr. :+1: 

This one looks good to me. Can you please rebase onto master again, to force the tests to run again?

fin swimmer

finswimmer (changes requested):
Thanks a lot for the quick rebase.

Just one more thing. At the moment the project itself get's installed at the end. I think we should disable this, just like one can do with the `--no-root` flag.

opelr:
@finswimmer Sounds good! I updated my commit to reflect your request, let me know if that's what you had in mind.

finswimmer (approved):
Thanks a lot :rocket:

TylerLubeck:
Thanks!! y'all are heros

noamba:
Hi @opelr @finswimmer, this is a very useful feature for me, many thanks for getting it done. 
I don't think this feature is in release 1.1.5, I assume this will be released sometime in the future? Do you have an idea when? Thanks, Noam

colbyhub:
@finswimmer it would be very helpful for our team to use this feature! Even if it was just in a preview release, our team could start taking advantage of this. Thank you for your work on this project! :muscle:

jose-lpa:
I'm not sure this has been implemented as people expected it.

I understood from all related issues and PRs that people wanted the same as I actually want: be able to install the development dependencies, **but without altering the main dependencies status**.

What this actually does, as far as I can see, is to make sure that **only the development dependencies are installed**. So it actually **removes any other dependency not specified as `dev`**.

Quick check screenshot:

Screenshot from 2021-06-29 16-16-18

So as you can see the command `poetry install --dev-only` actually left the project unusable, because it removed all core packages just by not being specified as `dev` dependencies.

I'm not sure what people wanted, but I seriously doubt anyone would expect this behavior when issuing a command to "install development dependencies". 

In my case, what I wanted is to use `poetry install --no-dev` in a Dockerfile to build a container image that can be used in both development, CI and Production. Then use `poetry install --dev-only` in the CI scripts as a previous step before running tests.

Issue #2908, opened 2020-09-11
poetry env remove does not work with in-project = true


ericriff:
Hi Guys, 

I'm using a poetry virtual env with `in-project = true` to have the environment created inside the project.
Anyways, if I run the command `poetry env remove python3` instead of removing `.venv` it errors out saying that `
Environment "poetry-project-RoyKdWrW-py3.5" does not exist.`

This command does work as expected if we store the environments in the regular cache.

It is a very minor bug since I can always do `rm -rf .venv` though.

Thanks for the amazing work.

- [X] I am on the [latest](https://github.com/python-poetry/poetry/releases/latest) Poetry version.
- [X] I have searched the [issues](https://github.com/python-poetry/poetry/issues) of this repo and believe that this is not a duplicate.
- [X] If an exception occurs when executing a command, I executed it again in debug mode (`-vvv` option).

- **OS version and name**: Ubuntu 16.04
- **Poetry version**: 1.1.0a2

8F3E:
Think this is because, unlike `poetry env list`, removal doesn't check for a `.venv` file if `in-project = true`. Happy to try and patch.

ericriff:
Thanks for looking into this!
I have another question regardless using `in-project = true`. This is a very useful feature since it allows us to easily integrate Poetry to our builds with docker/devcontainers since we don't have to mount external folders if we keep the environment in the project root.
The way Poetry implements this (which is the same way Pipenv does it AFAIK) is creating a `.venv` folder in the root of the project if the `in-project` flag is set.  This has a big limitation compared to using the global cache approach: you can only have one environment "ready to go" at the time, breaking the `poetry env use xxx` functionality. Usually this is not a problem since a project is locked to a given environment, but there are cases when it would be really cool to have multiple environments ready and switch between them with a simple command. They way it works now means that we have to delete `.venv` and re-build another environment.
Would it be possible to have something like `in-project` that makes Poetry generate different environments inside the root of the project?
Like .venv-python3.5, `.venv-python3.6`.

Thanks!

8F3E:
Sounds reasonable. Perhaps open a new issue as it would (I assume) require a change in the way poetry treats `in-project = true` globally, as opposed to implementing missing functionality?

max-sixty:
Is this a duplicate of https://github.com/python-poetry/poetry/issues/2124?

ericriff:
In deed it it. I'll close it.

Issue #2924, opened 2020-09-16
Support multi-version support for in-project virtual environment management


abn:
The current scenario when `virtualenvs.in-project = true` is specified, is that we default to creating `.venv` as the virtual environment. This needs to be improved to support multiple versions of python.

Ideally we will end up having a structure similar to the one below.

```
.
└── .venv
    ├── 3.7
    ├── 3.8
    └── 3.9
```

### Concensus Required
- [ ] Directory structure and naming.
- [ ] Should we default to this behaviour or make it opt-in only?
- [ ] If we want to default to this behaviour, should we start by opt-in first, and then opt-out?
- [ ] How should existing environments be handled? We can detect this and set project level config.

### Tasks
- [ ] Drive concensus, approved by members of @python-poetry/core 
- [ ] Implement new structure
- [ ] Implement test cases that validate new environment creation(s), gracefull handling of existing environments, compatibility with existing `env` sub-commands.

Related issues: #2908

finswimmer:
I would like to keep the current behavior for `virtualenvs.in-project = true`. It is common to have the the venv in a folder called `.venv`. Also some IDE can then autodetect the venv.

But I like the idea to have multiple venvs in the projects folder itself. We can come already quite close to it by setting `virtualenvs.path = ".virtualenvs"`. But this only works if you than run all your `poetry` commands at the same folder level as this folder. So if we could set e.g. `virtualenvs.path = "{project-dir}/virtualenvs"` and `project-dir` get's interpreted as the folder containing the `pyproject.toml`, we are already done.

Of course we could also hardcode the folder within the project's root where the different venv gets installed.

artorias111:
Hi! I can take this up if no one's working on it

finswimmer:
For sure @artorias111. We always appreciate contribution.

But we need to first come to a consensus on how it should look like, as @abn described in his post.

fin swimmer

KGB33:
What if poetry were to use symlinks to change the python version while keeping the `.venv` folder?

```
./
├── .py_versions/
│  ├── .py3.8/
│  └── .py3.10/
├── src/
├── test/
├── .venv -> .py_versions/.py3.8
└── pyproject.toml
```

Where `.py3.8/` and `.py3.10/` are virtualenvs for their respective python versions. 

One downside of using a symlink is that `rm -r .venv` won't remove the virtualenv.

abn:
[...]

PR #2748, opened 2020-07-30
add support for `poetry env remove` when `virtualenvs.in-project = true`

ITProKyle:
# Pull Request Check List

Resolves: #2124

- [x] Added **tests** for changed code.
- [x] Updated **documentation** for changed code.

## What Changed

- `poetry env remove` argument now defaults to `.venv` for ease of use with the static value
- when `.venv` is passed as the `<python>` argument and `virtualenvs.in-project=true`, it will remove the `.env` directory in the root of the project

Issue #2124, opened 2020-03-03
poetry env remove with virtualenvs.in-project = true


fgervais:
- [x] I am on the [latest](https://github.com/python-poetry/poetry/releases/latest) Poetry version.
- [x] I have searched the [issues](https://github.com/python-poetry/poetry/issues) of this repo and believe that this is not a duplicate.
- [x] If an exception occurs when executing a command, I executed it again in debug mode (`-vvv` option).

- **OS version and name**: 18.04.3 LTS (Bionic Beaver)
- **Poetry version**: Poetry version 1.0.5
- **Link of a [Gist](https://gist.github.com/) with the contents of your pyproject.toml file**: https://gist.github.com/fgervais/5a33d9df4cca064f02ce7fa6136c7e5e

## Issue
I think `poetry env remove` fails when `virtualenvs.in-project = true`

```
$ poetry env info

Virtualenv
Python:         3.6.9
Implementation: CPython
Path:           /home/fgervais/personal/test/.venv
Valid:          True

System
Platform: linux
OS:       posix
Python:   /usr

$ poetry config --list
cache-dir = "/home/fgervais/.cache/pypoetry"
virtualenvs.create = true
virtualenvs.in-project = true
virtualenvs.path = "{cache-dir}/virtualenvs"  # /home/fgervais/.cache/pypoetry/virtualenvs
```

```
$ poetry env remove 3.6.9

[ValueError]
Environment "test-dT4kcP9T-py3.6" does not exist.
```

esciara:
Found the same bug on macos:
```
$ sw_vers
ProductName:	Mac OS X
ProductVersion:	10.15.3
BuildVersion:	19D76

$ uname -a
Darwin Ems-MacBook-Pro 19.3.0 Darwin Kernel Version 19.3.0: Thu Jan  9 20:58:23 PST 2020; root:xnu-6153.81.5~1/RELEASE_X86_64 x86_64
```

ketozhang:
Same issue with a more relative command

```sh
$ poetry env remove .venv/bin/python
```

---

In addition, requesting more simple relative remove command

```sh
$ poetry env remove .
```

jidicula:
Also encountered the same bug on macOS 11:

```
❯ sw_vers
ProductName:	macOS
ProductVersion:	11.4
BuildVersion:	20F71

❯ uname -a
[...]

PR #2925, opened 2020-09-16
[WIP] fix poetry env remove does not work with in-project = true

8F3E:
#2908 

Fixing issue where a `poetry env remove .venv` fails in a project with `virtualenvs.in-project = true` in `poetry.toml` (so the virtualenv is stored in `.venv`).

# Pull Request Check List

Resolves: #issue-number-here

### DIFF
--- a/poetry/utils/env.py
+++ b/poetry/utils/env.py
@@ -417,7 +417,7 @@ class EnvManager(object):
         envs_file = TomlFile(venv_path / self.ENVS_FILE)
         base_env_name = self.generate_env_name(self._poetry.package.name, str(cwd))
 
-        if python.startswith(base_env_name):
+        if python.startswith(base_env_name) or python == ".venv":
             venvs = self.list()
             for venv in venvs:
                 if venv.path.name == python:


poetry/utils/env.py
```diff
@@ -417,7 +417,7 @@ def remove(self, python):  # type: (str) -> Env
         envs_file = TomlFile(venv_path / self.ENVS_FILE)
         base_env_name = self.generate_env_name(self._poetry.package.name, str(cwd))
 
-        if python.startswith(base_env_name):
+        if python.startswith(base_env_name) or python == ".venv":
```
abn, line 420: We shouldn't really do this if the environment was not created by us. There needs to be some flag somewhere (maybe even a marker file) in the env directory to determine if we created it.

8F3E: How about checking `self._poetry.config.get("virtualenvs.in-project")` is true, similar to what's done when listing the envs?

abn: The issue is that poetry supports existing environments. So, in cases where, lets say, someone needs some wierd system lib available etc., they might create `.venv` manually and also have `virtualenvs.in-project` set to `true`. In this scenario, there is no guarentee that we created it. Obviously other simpler scenarios exists too. Just picked an obtuse one.

8F3E: Does a marker file / flag like you say exist somewhere already or will that need to be added when Poetry creates a `.venv`?

abn: At the moment no. If we need to support this cleanly we should add one and add a good message if one is not found when user tries to delete this. 

Virtualenv already creates a `pyvenv.cfg`, either relying on that or placing a `poetry.toml` file in the environment root maybe sufficient. Not sure what is the better approach here at the moment.

abn: The latter affords expansion later.

8F3E: So for a workaround (other than `rm` as noted in the initial issue) until #2924 (presumably?) changes the approach for a `.venv` would the following be sufficient?

1. Check `poetry.toml` exists in the project root.
2. Check `poetry.toml` contains `in-project = true` in the appropriate place.

Would a situation exist where a user has set locally `in-project = true` but has manually created `.venv` _and_ doesn't want to remove it when running `poetry env remove .venv`?

8F3E:
Original issue duplicate of #2124, which has earlier PR #2748, so closing.

Issue #2942, opened 2020-09-20
Poetry 1.1.0b3 fails to self update


albireox:
- [X] I am on the [latest](https://github.com/python-poetry/poetry/releases/latest) Poetry version.
- [X] I have searched the [issues](https://github.com/python-poetry/poetry/issues) of this repo and believe that this is not a duplicate.
- [x] If an exception occurs when executing a command, I executed it again in debug mode (`-vvv` option).

- **OS version and name**: macOS 11.0
- **Poetry version**: 1.1.0b3
- **Link of a [Gist](https://gist.github.com/) with the contents of your pyproject.toml file**: N/A

## Issue

After upgrading to 1.1.0b3 it seems the self upgrade command fails

```console
albireoproˇalbireo ~ » poetry self update --preview

  TypeError

  find_packages() got an unexpected keyword argument 'allow_prereleases'

  at .poetry/lib/poetry/console/commands/self/update.py:91 in handle
       87│         if not version:
       88│             version = ">=" + __version__
       89│
       90│         repo = PyPiRepository(fallback=False)
    →  91│         packages = repo.find_packages(
       92│             "poetry", version, allow_prereleases=self.option("preview")
       93│         )
       94│         if not packages:
       95│             self.line("No release found for the specified version")
```

Using `-vvv` weirdly gives a completely different message

```console
albireoproˇalbireo ~ » poetry -vvv self update --preview

  Stack trace:

  11  .poetry/lib/poetry/_vendor/py3.8/clikit/console_application.py:123 in run
      io = io_factory(

  10  .poetry/lib/poetry/console/config/application_config.py:221 in create_io
      resolved_command = application.resolve_command(args)

   9  .poetry/lib/poetry/_vendor/py3.8/clikit/console_application.py:110 in resolve_command
      return self._config.command_resolver.resolve(args, self)

   8  .poetry/lib/poetry/_vendor/py3.8/clikit/resolver/default_resolver.py:43 in resolve
      result = self.process_default_commands(args, application.default_commands)

   7  .poetry/lib/poetry/_vendor/py3.8/clikit/resolver/default_resolver.py:104 in process_default_commands
      if resolved_command.is_parsable():

   6  .poetry/lib/poetry/_vendor/py3.8/clikit/resolver/resolve_result.py:43 in is_parsable
      self._parse()

   5  .poetry/lib/poetry/_vendor/py3.8/clikit/resolver/resolve_result.py:49 in _parse
[...]

PR #2944, opened 2020-09-21
Fix: self update failes with unexpected keyword argument error

finswimmer:
The `find_packages` method expects a `Dependency` as a parameter. This PR fixes this.

Closes: #2942


abn:
@finswimmer can you try rebasing on master please?

abn (approved):

Issue #898, opened 2019-02-16
Feature Request:  Please don't modify my dotfiles


markfaine:
- [x ] I have searched the [issues](https://github.com/sdispater/poetry/issues) of this repo and believe that this is not a duplicate.
- [x ] I have searched the [documentation](https://poetry.eustace.io/docs/) and believe that my question is not covered.

## Feature Request

Please don't modify my dotfiles when installing.  This is just bad form.  The default should be to expect the user to do this. Providing an option to do it in the installer is fine, it should just not be the default.

jacebrowning:
Isn't that the sole purpose of the installer -- to get `poetry` added to PATH?

markfaine:
No, just to install it. Which would include downloading it and any dependencies, basically, ensuring it will work if called directly.  Though, of course it can *optionally* add it to the path.

pmav99:
@jacebrowning Poetry needs to be installed in a different python distribution (think virtualenv) than your main project, because poetry and your own project might have conflicting dependencies.

markfaine:
Ok, I get that but I don't see what that has to do with modifying my environment files.  It's just adding the path to poetry to the PATH environment variable.  I'd rather always manage the path myself and not have every tool I install modifying PATH.  It's a very slippery slope and before you know it you have a terribly cluttered path, you have no idea what is in your dotfiles, and you've lost all control over the environment.  Like most users of *nix based operating systems, I have custom dotfiles that take care of managing PATH (along with everything else) and I want to maintain absolute control over it.  This is not an uncommon opinion, in fact, I'd say it's probably the prevailing opinion.

neighthan:
I just found `poetry` today and was going to post this issue if it wasn't already here. I think _asking_ to add it to my path is great; some users might not mess with dotfiles much, and that could be a good convenience. I keep my dotfiles in a git repo, though, and share them across machines, and I don't like them having random things appended.

markfaine:
What I see most tools do is output a message at the end, something like:  Please add the following to your .bashrc...  Prompting to add it is okay too I guess, so long as there is still a non-interactive means of installing.

neighthan:
Is this fixed already? I just installed `poetry` again (on Windows this time), and this is the first thing that comes up:

```
>python get-poetry.py
Retrieving Poetry metadata

Before we start, please answer the following questions.
You may simply press the Enter key to leave unchanged.
Modify PATH variable? ([y]/n)
```

Calinou:
[...]

PR #2945, opened 2020-09-21
Inform user about option to skip changing PATH

bodograumann:
I also added the `-` argument to `python`, because without it you cannot simply append flags for the script.
Hope this is just the same for powershell, as for bash.

Cf. #898

This is only a documentation change.

Btw. your contributing guidelines should say which branch to work from (I assume it’s master) and maybe use less bold highlight to increase readability.

### DIFF
--- a/docs/docs/index.md
+++ b/docs/docs/index.md
@@ -23,11 +23,11 @@ recommended way of installing `poetry`.
 
 ### osx / linux / bashonwindows install instructions
 ```bash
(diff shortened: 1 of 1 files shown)


finswimmer (approved):
Thanks a lot! Looks good to me :+1:

dlukes:
> I also added the - argument to python, because without it you cannot simply append flags for the script.
> Hope this is just the same for powershell, as for bash.

It should be, the command line argument is interpreted by the `python` executable, not the shell, so it should work the same across different shells :)

PR #2948, opened 2020-09-22
Bump version to 1.1.0b4

abn:
### Changed

- When running under Python 2.7 on Windows, install command will be limited to one worker to mitigate threading issue ([#2941](https://github.com/python-poetry/poetry/pull/2941)).


abn:
Manually uploaded FreeBSD artifacts because the cirrus ci documentation seems to be incorrect regarding `CIRRUS_RELEASE`.

```
#!/usr/bin/env bash
if [[ "$CIRRUS_RELEASE" == "" ]]; then
  echo "Not a release. No need to deploy!"
  exit 0
fi
Not a release. No need to deploy!
```

PR #2953, opened 2020-09-23
cirrus: ensure we use github api to fetch release id and skip tests on tag

abn:
This change ensures
- we fetch release id from github api if one is not available via `CIRRUS_RELEASE` environment variable
- tests are skipped for release builds


abn:
@sdispater please make sure we merge this prior to tagging the next beta for 1.1.0.

sdispater (approved):

Issue #2761, opened 2020-08-01
Unable to retrieve the package version for git on preview 1.1.0b2


serjflint:
- [ ] I am on the [latest](https://github.com/python-poetry/poetry/releases/latest) Poetry version.
- [x] I have searched the [issues](https://github.com/python-poetry/poetry/issues) of this repo and believe that this is not a duplicate.
- [x] If an exception occurs when executing a command, I executed it again in debug mode (`-vvv` option).

- **OS version and name**: Linux 5.7.11-1-MANJARO
- **Poetry version**: 1.1.0b2
- **Link of a [Gist](https://gist.github.com/) with the contents of your pyproject.toml file**: https://gist.github.com/serjflint/dfbb4bc8293ebce28b256aa53c22c8ac

## Issue
I need the preview build for solving another bug. But now git packages are not working.
Works as intended on the latest release 1.0.10.

```
  Stack trace:

  9  ~/.poetry/lib/poetry/_vendor/py3.8/clikit/console_application.py:131 in run
      129│             parsed_args = resolved_command.args
      130│ 
    → 131│             status_code = command.handle(parsed_args, io)
      132│         except KeyboardInterrupt:
      133│             status_code = 1

  8  ~/.poetry/lib/poetry/_vendor/py3.8/clikit/api/command/command.py:120 in handle
      118│     def handle(self, args, io):  # type: (Args, IO) -> int
      119│         try:
    → 120│             status_code = self._do_handle(args, io)
      121│         except KeyboardInterrupt:
      122│             if io.is_debug():

  7  ~/.poetry/lib/poetry/_vendor/py3.8/clikit/api/command/command.py:171 in _do_handle
      169│         handler_method = self._config.handler_method
      170│ 
    → 171│         return getattr(handler, handler_method)(args, io, self)
      172│ 
      173│     def __repr__(self):  # type: () -> str

  6  ~/.poetry/lib/poetry/_vendor/py3.8/cleo/commands/command.py:92 in wrap_handle
       90│         self._command = command
       91│ 
    →  92│         return self.handle()
       93│ 
       94│     def handle(self):  # type: () -> Optional[int]

  5  ~/.poetry/lib/poetry/console/commands/add.py:101 in handle
       99│                     raise ValueError("Package {} is already present".format(name))
      100│ 
    → 101│         requirements = self._determine_requirements(
      102│             packages,
      103│             allow_prereleases=self.option("allow-prereleases"),

[...]

PR #2957, opened 2020-09-24
Fix cases where poetry should fall back to pep517 metadata parsing

finswimmer:
https://github.com/python-poetry/poetry/issues/2761 discovered cases where `poetry` should fall back to pep517 to read metadata.

According to the [docs of setup.cfg](https://setuptools.readthedocs.io/en/latest/userguide/declarative_config.html#specifying-values) the version can be specified by linking to an attribute in a module by using `attr: somemodule:__version__`. At the moment `poetry` will extract this literally and will try to parse it later, when it cannot fall back to pep517. This PR changes it, by trying to parse the version at the moment of extracting from the `setup.cfg`. If it failes a `ParseVersionError` is thrown and poetry will fallback to pep517 parsing.

The second case appears if requirements and name or version (but not both) can be extracted from either `setup.py` or `setup.cfg`. name and version are mandatory information. If poetry cannot find any requirements, it cannot differ whether there are no or it was unable to parse it. So poetry should fall back to pep517 whenever any of these values are missing.

# Pull Request Check List

Resolves: https://github.com/python-poetry/poetry/issues/2761

### DIFF
--- a/poetry/inspection/info.py
+++ b/poetry/inspection/info.py
@@ -326,7 +326,7 @@ class PackageInfo:
             requires_python=python_requires,
         )
 
-        if not info.name or not info.version:
+        if not (info.name and info.version) and not info.requires_dist:
             # there is nothing useful here
             raise PackageInfoError(path)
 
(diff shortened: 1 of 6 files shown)


abn:
@finswimmer can we also add a case in the info test suite to ensure that pep517 build was called?

finswimmer:
@abn Done.

abn (approved):

Issue #2965, opened 2020-09-25
Can not load package data resources when running in a zipapp


sinoroc:
- **OS version and name**: Ubuntu Linux 18.04
- **Poetry version**:   1.0.10 (in a [_`zipapp`_](https://docs.python.org/3.6/library/zipapp.html))
- **Python**: CPython 3.6.9

## Issue

```
$ poetry config -vvv --list

[ValueError]
Schema poetry-schema does not exist.

Traceback (most recent call last):
  File "/home/sinoroc/.local/bin/poetry/clikit/console_application.py", line 131, in run
    status_code = command.handle(parsed_args, io)
  File "/home/sinoroc/.local/bin/poetry/clikit/api/command/command.py", line 120, in handle
    status_code = self._do_handle(args, io)
  File "/home/sinoroc/.local/bin/poetry/clikit/api/command/command.py", line 171, in _do_handle
    return getattr(handler, handler_method)(args, io, self)
  File "/home/sinoroc/.local/bin/poetry/cleo/commands/command.py", line 92, in wrap_handle
    return self.handle()
  File "/home/sinoroc/.local/bin/poetry/poetry/console/commands/config.py", line 75, in handle
    local_config_file = TomlFile(self.poetry.file.parent / 'poetry.toml')
  File "/home/sinoroc/.local/bin/poetry/poetry/console/commands/command.py", line 10, in poetry
    return self.application.poetry
  File "/home/sinoroc/.local/bin/poetry/poetry/console/application.py", line 49, in poetry
    self._poetry = Factory().create_poetry(Path.cwd())
  File "/home/sinoroc/.local/bin/poetry/poetry/factory.py", line 48, in create_poetry
    check_result = self.validate(local_config)
  File "/home/sinoroc/.local/bin/poetry/poetry/factory.py", line 272, in validate
    validation_errors = validate_object(config, 'poetry-schema')
  File "/home/sinoroc/.local/bin/poetry/poetry/json/__init__.py", line 22, in validate_object
    raise ValueError('Schema {} does not exist.'.format(schema_name))
```

Culprit is probably [this block of code](https://github.com/python-poetry/poetry/blob/1.0.10/poetry/json/__init__.py#L10-L19):
```
# [...]
SCHEMA_DIR = os.path.join(os.path.dirname(__file__), "schemas")

class ValidationError(ValueError):
    pass

def validate_object(obj, schema_name):  # type: (dict, str) -> List[str]
[...]

PR #2974, opened 2020-09-25
Use pkgutil.get_data for zip-safe resource loading

sinoroc:
When installed in a Zip file (`zipapp`, `*.pyz`) loading resources
from package data can not be done by directly reading files relative
to `__file__`, since there is no valid path on the file system for
zipped files.

Using 'pkgutil.get_data' ensures that loading files from a Zip archive
is handled accordingly.

Resolves: #2965

### DIFF
--- a/poetry/json/__init__.py
+++ b/poetry/json/__init__.py
@@ -1,13 +1,11 @@
 import json
 import os
+import pkgutil
 
 from io import open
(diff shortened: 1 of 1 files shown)


sinoroc:
Bug fixes go only on `master` right now.

PR #2975, opened 2020-09-25
Use pkgutil.get_data for zip-safe resource loading

sinoroc:
When running from a Zip file (`zipapp`, `*.pyz`), loading resources
from package data can not be done by directly reading files relative
to `__file__`, since there is no valid path on the file system for
zipped files.

Using `pkgutil.get_data` ensures that loading files from a Zip archive
is handled accordingly.

Resolves: #2965

### DIFF
--- a/poetry/json/__init__.py
+++ b/poetry/json/__init__.py
@@ -1,13 +1,11 @@
 import json
 import os
+import pkgutil
 
 from io import open
 from typing import List
 
 import jsonschema
-
-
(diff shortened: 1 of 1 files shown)


abn:
@sinoroc appreciate the change; we might wait till we release `1.1` before reviewing this one. I am all for this or a similar change. For poetry, I think we are good to use `importlib.resources`.

neersighted:
No longer relevant as we've removed the schema from Poetry itself

Issue #2980, opened 2020-09-25
parsing issue when we have extras with poetry add


ycd:
I mentioned this problem in #2782 but it still occurs in version **1.0.10**

@abn approved it's a bug in parsing, see [here](https://github.com/python-poetry/poetry/issues/2782#issuecomment-671458256)

### Reproducing the problem

When we have space after comma it can't parses the rest of extras. 

```python
poetry add databases[postgresql, sqlite]

[ValueError]
Could not find a matching version of package databases[postgresql,
```

It works perfectly fine without a space.

```python
poetry add databases[postgresql,sqlite]
Using version ^0.3.2 for databases
```

ycd:
Thanks for such a quick fix @finswimmer , @abn

abn:
@ycd thank you for testing and reporting. :)

PR #2985, opened 2020-09-27
Fix: allow whitespace in list of extras for a package

finswimmer:
This fix allows having whitespaces in the list of extras during `poetry add`, e.g. `poetry add databases[postgresql, sqlite]`.

# Pull Request Check List

Resolves: https://github.com/python-poetry/poetry/issues/2980

### DIFF
--- a/poetry/console/commands/init.py
+++ b/poetry/console/commands/init.py
@@ -373,7 +373,7 @@ The <c1>init</c1> command creates a basic <comment>pyproject.toml</> file in the
         for requirement in requirements:
             requirement = requirement.strip()
             extras = []
-            extras_m = re.search(r"\[([\w\d,-_]+)\]$", requirement)
+            extras_m = re.search(r"\[([\w\d,-_ ]+)\]$", requirement)
             if extras_m:
                 extras = [e.strip() for e in extras_m.group(1).split(",")]
                 requirement, _ = requirement.split("[")
--- a/tests/console/commands/test_init.py
+++ b/tests/console/commands/test_init.py
@@ -558,8 +558,9 @@ python = "~2.7 || ^3.6"
     assert 'pytest = "^3.6.0"' in output
 
 
-def test_add_package_with_extras_and_whitespace(init_command):
-    result = init_command._parse_requirements(["databases[postgresql, sqlite]"])
+def test_add_package_with_extras_and_whitespace(app):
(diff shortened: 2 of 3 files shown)


tests/console/commands/test_init.py
```diff
@@ -556,3 +556,13 @@ def test_predefined_and_interactive_dev_dependencies(tester, repo):
     assert expected in output
     assert 'pytest-requests = "^0.2.0"' in output
     assert 'pytest = "^3.6.0"' in output
+
+
+def test_add_package_with_extras_and_whitespace(app):
+    command = app.find("init")
+    result = command._parse_requirements(["databases[postgresql, sqlite]"])
```
abn, line 563:
```suggestion
def test_add_package_with_extras_and_whitespace(tester):
    result = tester._parse_requirements(["databases[postgresql, sqlite]"])
```

poetry/console/commands/init.py
```diff
@@ -373,7 +373,7 @@ def _parse_requirements(
         for requirement in requirements:
             requirement = requirement.strip()
             extras = []
-            extras_m = re.search(r"\[([\w\d,-_]+)\]$", requirement)
+            extras_m = re.search(r"\[([\w\d,-_ ]+)\]$", requirement)
```
abn, line 376: Shoud we just use `\s` here?

abn (changes requested):

poetry/console/commands/init.py
finswimmer: The range of `\s` is to wide. I don't think supporting e.g. tab characters here.

tests/console/commands/test_init.py
finswimmer: It look like it must be 'test._command._parse_requirements` ...

poetry/console/commands/init.py
abn: Hmm, I guess for this case it is fine (ish). For PEP 508, whitespace typically refers to "non-line breaking whitespace". That is narrowier than `\s` and but wider than ` `. I doubt anyone is going to type in `foo[a, \t b]` :smirk:

abn (approved):

abn (changes requested):
Issue talks about `add` command but the fix is in `init` command?

abn (approved):

PR #2998, opened 2020-09-29
preserve line endings in toml files

### LINKED ISSUE #2993: keep old line breaking in toml files

## Feature Request

I keep pyproject.toml and lock file in my code repo with \n as eol. but poetry will always make it crlf on Windows, it's pretty annoying.

create a pr in https://github.com/sdispater/tomlkit/pull/110

Trim21:
I'm not sure if poetry should take care of line breaking or `tomlkit` should do it... 🤔

Resolves: #2993

- <del> Updated **documentation** for changed code.</del>


abn:
@Trim21 I honestly do not like how we need to do it here. Since `tomlkit` is supposed to be style preserving and read/write is implemented there, it stands to reason that it must implicitly handle detection, preservation and modification of line endings.

Also note that the logic for toml handling is moving to `poetry.core.toml`.

Trim21:
@abn You are right, I fotget that tomlkit do style preserving.

Issue #2996, opened 2020-09-29
Editable install by user fails with permission error


agoose77:
- [x] I am on the [latest](https://github.com/python-poetry/poetry/releases/latest) Poetry version.
- [x] I have searched the [issues](https://github.com/python-poetry/poetry/issues) of this repo and believe that this is not a duplicate.
- [x] If an exception occurs when executing a command, I executed it again in debug mode (`-vvv` option).

- **Docker python:3.8-alpine**: 
- **1.1.0rc1**: 
- **[pyproject.toml](https://gist.github.com/agoose77/5b08db26a49c8119a9a943583ef724b3)**
- **[Dockerfile](https://gist.github.com/agoose77/647769f65494f6a82ffc3f625eca0457)**

## Issue
I am using docker to build a containerised application. Inside the container, I install poetry as root, whilst I perform the installation of my package as a non-root user. This fails after trying to write to the system site-packages directory. I have disabled virtualenvs for my use case.

This error was introduced between 1.0.10 and the RC release, and I think it likely that the recent change to editable installs was the culprit.
Traceback: https://gist.github.com/agoose77/7c117214ddab309e8ba2069ecc36701c

agoose77:
For posterity, you can force the old behaviour with
```toml
[tool.poetry.build]
generate-setup-file = true
script = "math import *;build = lambda _: _;from math"
```
Although, as cool as Python injection is, I would prefer to avoid doing this!

abn:
@agoose77 in most cases I would recommend that you use a poetry manage virtual environment. Using system sites are not recommended for applications. 

Also, `poetry config virtualenvs.in-project true` can also be useful.

agoose77:
> @agoose77 in most cases I would recommend that you use a poetry manage virtual environment. Using system sites are not recommended for applications. 
> 
> Also, `poetry config virtualenvs.in-project true` can also be useful.

Thanks. Normally I use a venv for those reasons, and it's great. I need to avoid using one here because it greatly simplifies my docker image (which has several ways of being used). Pip seems to default to using the XDG local environment, could poetry respect that too?

abn:
@agoose77 using a poetry managed environment should not make a difference here tbh. As in just change the entry point to `poetry shell` and you should always have a ready environment etc. But yeah, I understand use cases might be different.

See if the fix i proposed in #3000 works for now. Its a bit tricky sometimes to decide between what site to use.

agoose77:
[...]

PR #3000, opened 2020-09-29
env: enable usersite for system environments

abn:
Resolves: #2996


sdispater:
It seems that pip only falls back on the user path only if the default `site-packages` is not writable:

```
Defaulting to user installation because normal site-packages is not writeable
```

So, we could keep the same logic.

abn:
Yeah I thought about that. But this would mean explicit check. I guess  we can localise the check just before we write the pth file.

sdispater (approved):

PR #2887, opened 2020-09-06
Fix and improve handling of extras while resolving dependencies

sdispater:
## Pull Request Check List

Resolves: python-poetry/poetry#2545
Resolves: python-poetry/poetry#2494
Resolves: python-poetry/poetry#2300
Resolves: python-poetry/poetry#2080
Resolves: python-poetry/poetry#1609

This one was bothering for a while and was quite the struggle to fix.

Basically, extras could trip up Poetry if the package already existed in the dependency graph leading to dependencies being ignore or uninstalled when updating dependencies.

To make it work, a rather big refactoring was required (see https://github.com/python-poetry/poetry-core/pull/78) to make dependencies and packages comparison more accurate.

The main trick here is that a package with extras will now have its non-extra version as a dependency to help the resolver figure out what needs to be done.

Overall, it simplifies the code by always adding all dependencies of packages (even optional ones) and it's the `Provider` that will now decide which ones should be exposed to the resolver based on the extras selected.

One last thing of note is that source urls of file and directory dependencies will now be the fully resolved path, i.e. absolute, to actually be able to make comparisons.

- [x] Added **tests** for changed code.
- [ ] Updated **documentation** for changed code.

PR #3007, opened 2020-09-30
Add note on editable installation

fredrikaverpil:
This adds a note in the docs about editable installation, following the change in #2887, as part of release https://github.com/python-poetry/poetry/releases/tag/1.1.0b3

### DIFF
--- a/docs/docs/dependency-specification.md
+++ b/docs/docs/dependency-specification.md
@@ -117,7 +117,10 @@ my-package = { path = "../my-package/", develop = false }
 my-package = { path = "../my-package/dist/my-package-0.1.0.tar.gz" }
 ```
 
-For directories, set `develop = true` for editable installation.
+!!!note
+
+    Before poetry 1.1 directory path dependencies were installed in editable mode by default. You should set the `develop` attribute explicitly,
+    to make sure the behavior is the same for all poetry versions.
 
 ## `url` dependencies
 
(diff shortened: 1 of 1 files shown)


abn:
@fredrikaverpil hasn't this already been done? https://github.com/python-poetry/poetry/commit/d4eb422e453b0b153e6bacd21dbc5ae372e59f98#diff-e3e2a9bfd88566b05001b02a3f51d286

fredrikaverpil:
Are there two places where path dependency syntax is specified then?

docs/docs/dependency-specification.md
```diff
@@ -111,12 +111,13 @@ you can use the `path` property:
...
 [tool.poetry.dependencies]
 # directory
-my-package = { path = "../my-package/" }
+my-package = { path = "../my-package/", develop = false }
 
 # file
 my-package = { path = "../my-package/dist/my-package-0.1.0.tar.gz" }
 ```
 
+For directories, set `develop = true` for editable installation.
```
abn, line 120:
```suggestion
!!!note

    Before poetry 1.1 directory path dependencies were installed in editable mode by default. You should always set the `develop` attribute explicit,
    to make sure the behavior is the same for all poetry versions.
```

abn (changes requested):

abn (approved):

Issue #1894, opened 2020-01-15
poetry export: missing --trusted-host argument for non-HTTPS repos


1somorphism:
- [x] I am on the [latest](https://github.com/python-poetry/poetry/releases/latest) Poetry version.
- [x] I have searched the [issues](https://github.com/python-poetry/poetry/issues) of this repo and believe that this is not a duplicate.
- [ ] If an exception occurs when executing a command, I executed it again in debug mode (`-vvv` option).

- **OS version and name**: Ubuntu 16.04.2 TLS and Windows 10
- **Poetry version**: 1.0.2
- **Link of a [Gist](https://gist.github.com/) with the contents of your pyproject.toml file**: https://gist.github.com/1somorphism/f3a1ced4e6507383b5f16daffcb7a140

## Issue
Ideally line 1 of the output of ```poetry export -f requirements.txt``` for the file above should be

```--extra-index-url http://10.0.0.0:8080/repository/pypi-hosted/simple --trusted-host 10.0.0.0```

**Comment by 0xADD1E** [comment]
I've just worked around this by just writing out the trusted host line first 
```
echo --trusted-host 10.0.0.0 > requirements.txt
poetry export -f requirements.txt >> requirements.txt
```

But I also agree that this is somewhat unexpected behaviour.

My intuition for how this would work was that there would be an option in `tool.poetry.source` that would allow for this (and possibly also turn off the somewhat annoying "insecure host" notice when installing packages)

gusbemacbe:
@1somorphism, @0xADD1E, @edwardgeorge, @bmispelon, @blueyed, @estyxx, @abn, @finswimmer 

Please check my discussion topic #4869.

japita-se:
this very useful issue has been merged?

PR #3014, opened 2020-10-01
Add --trusted-host for non HTTPS repo (#1894)

estyxx:
# Pull Request Check List

Resolves: #1894

This will add `--trusted-host {url}` to the`requirement.txt` export at the beginning of the file if the repo is not HTTPS.


estyxx:
P. S. I saw tests are broken... I wasn't expecting them, think the release effected somehow... I'm looking at them
(I will set it to WIP if was possible)

Apparently, with HTTP legacy repository, I got dependencies that are not in the current repository, so it couldn't find them...

estyxx:
Fyi this issue is related to Hacktoberfest :) 
And for my point of view could be merge tnks :)

poetry/utils/exporter.py
```diff
@@ -154,6 +154,8 @@ def _export_requirements_txt(
                 url = (
                     repository.authenticated_url if with_credentials else repository.url
                 )
+                if repository.url.startswith("http://"):
```
abn, line 157: Might be better to use url parsing and checking for parsed `scheme` here. You can see authenticator code to see an example. Besure to use the compat implementation.

tests/utils/test_exporter.py
```diff
@@ -779,6 +779,61 @@ def test_exporter_exports_requirements_txt_with_legacy_packages(tmp_dir, poetry)
     assert expected == content
 
 
+def test_exporter_exports_requirements_txt_with_legacy_packages_trusted_host(
+    tmp_dir, poetry
+):
+    poetry.pool.add_repository(LegacyRepository("custom", "http://example.com/simple",))
+    poetry.locker.mock_lock_data(
```
abn, line 786: You can drop most of the mock; effectively a no package data is fine since we are only testing the the indices. No point in duplicating data here.

poetry/utils/exporter.py
```diff
@@ -154,6 +154,8 @@ def _export_requirements_txt(
...
+                    indexes_header += "--trusted-host {}\n".format(url)
```
abn, line 158: Shouldn't this be hostname and not the url?

abn (changes requested):

poetry/utils/exporter.py
estyxx: yup!

tests/utils/test_exporter.py
estyxx: I remove a package from the mock data, think I can't remove other stuff or the test will fail...

poetry/utils/exporter.py
estyxx: sure!

jaklan:
Very useful PR, waiting for that!

abn (approved):

jaklan:
@abn I see it was merged to `master` branch, but not `1.1`, but doesn't it actually classify also as a bug fix, so it can be merged there and provided with the next patch release?

abn:
@jaklan happy accept this ported to `1.1`. 

Not all bugfixes get backported, and considering this is not a regression we didn't port it.

PR #3027, opened 2020-10-01
fix #1969

### LINKED ISSUE #1969: Clarify what happens when virtualenvs.create is false.

## Issue

This issue is about docs on configuration option [`virtualenvs.create`](https://python-poetry.org/docs/configuration/#virtualenvscreate-boolean). Currently it says about what happens when option is true. I think, its better to state explicitly, that when option is `false`, then `poetry` will install dependencies to current python interpreter/virtualenv.

NicklasTegner:
# Pull Request Check List

Resolves: #1969

Due note, that only documentation was updated in this pr, so no code changed.

### DIFF
--- a/docs/docs/configuration.md
+++ b/docs/docs/configuration.md
@@ -106,6 +106,7 @@ Defaults to one of the following directories:
 
 Create a new virtual environment if one doesn't already exist.
 Defaults to `true`.
+If set to `false`, poetry will install dependencies into the current python environment.
 
 !!!note:
         When setting this configuration to `false`, the Python environment used must have `pip`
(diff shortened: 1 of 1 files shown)


docs/docs/configuration.md
```diff
@@ -106,6 +106,7 @@ Defaults to one of the following directories:
 
 Create a new virtual environment if one doesn't already exist.
 Defaults to `true`.
+If set to `false`, poetry will install dependencies into the current python environment.
```
finswimmer, line 109: Can we have a new line before? :)

finswimmer (changes requested):
Thanks a lot!

Just one minor suggestion.

finswimmer (approved):
Thanks :+1:

Issue #3026, opened 2020-10-01
RuntimeError


remram44:
- [x] I am on the [latest](https://github.com/python-poetry/poetry/releases/latest) Poetry version.
- [x] I have searched the [issues](https://github.com/python-poetry/poetry/issues) of this repo and believe that this is not a duplicate.
- [x] If an exception occurs when executing a command, I executed it again in debug mode (`-vvv` option).

- **OS version and name**: Ubuntu Linux 20.04.1
- **Poetry version**: 1.1.0
- **Link of a [Gist](https://gist.github.com/) with the contents of your pyproject.toml file**: https://gitlab.com/ViDA-NYU/datamart/datamart/-/blob/1f15e3db009837cb51ee22274ce9bf3b96e205d7/pyproject.toml

## Issue

Since the update, running `poetry install` crashes with a `RuntimeError`.

<details>
<summary><code>poetry install -vvv</code></summary>

```
Virtualenv datamart-TD2l9kg0-py3.8 already exists.
Using virtualenv: /home/remram/.cache/pypoetry/virtualenvs/datamart-TD2l9kg0-py3.8
Installing dependencies from lock file

Finding the necessary packages for the current system

  Stack trace:

  17  ~/.poetry/lib/poetry/_vendor/py3.8/clikit/console_application.py:131 in run
       129│             parsed_args = resolved_command.args
       130│ 
     → 131│             status_code = command.handle(parsed_args, io)
       132│         except KeyboardInterrupt:
       133│             status_code = 1

  16  ~/.poetry/lib/poetry/_vendor/py3.8/clikit/api/command/command.py:120 in handle
       118│     def handle(self, args, io):  # type: (Args, IO) -> int
       119│         try:
     → 120│             status_code = self._do_handle(args, io)
       121│         except KeyboardInterrupt:
       122│             if io.is_debug():

  15  ~/.poetry/lib/poetry/_vendor/py3.8/clikit/api/command/command.py:171 in _do_handle
       169│         handler_method = self._config.handler_method
       170│ 
     → 171│         return getattr(handler, handler_method)(args, io, self)
       172│ 
       173│     def __repr__(self):  # type: () -> str

  14  ~/.poetry/lib/poetry/_vendor/py3.8/cleo/commands/command.py:92 in wrap_handle
        90│         self._command = command
        91│ 
     →  92│         return self.handle()
        93│ 
        94│     def handle(self):  # type: () -> Optional[int]

[...]

Issue #1837, opened 2020-01-06
Unable to install using python 3 on clean Ubuntu 18.04 image


jonapich:
I am trying to install poetry as early as possible in a clean `docker run -it ubuntu:18.04` interactive session:

    apt update
    apt install python3 curl
    curl -sSL https://raw.githubusercontent.com/python-poetry/poetry/master/get-poetry.py | python3

at this point, poetry installed but won't run because there's no "python" on the system; only python3. so I made a symlink and tried again:

    poetry --version

    Traceback (most recent call last):
      File "/root/.poetry/lib/poetry/_vendor/py3.6/keyring/backend.py", line 203, in _load_plugins
        init_func = ep.load()
      File "/root/.poetry/lib/poetry/_vendor/py3.6/importlib_metadata/__init__.py", line 92, in load
        module = import_module(match.group('module'))
      File "/usr/lib/python3.6/importlib/__init__.py", line 126, in import_module
        return _bootstrap._gcd_import(name[level:], package, level)
      File "<frozen importlib._bootstrap>", line 994, in _gcd_import
      File "<frozen importlib._bootstrap>", line 971, in _find_and_load
      File "<frozen importlib._bootstrap>", line 955, in _find_and_load_unlocked
      File "<frozen importlib._bootstrap>", line 665, in _load_unlocked
      File "<frozen importlib._bootstrap_external>", line 678, in exec_module
      File "<frozen importlib._bootstrap>", line 219, in _call_with_frames_removed
      File "/usr/lib/python3/dist-packages/keyrings/alt/Windows.py", line 9, in <module>
        from . import file_base
      File "/usr/lib/python3/dist-packages/keyrings/alt/file_base.py", line 13, in <module>
        from keyring.util.escape import escape as escape_for_ini
    ModuleNotFoundError: No module named 'keyring.util.escape'

I retried the steps using `apt install python3-pip` followed by `python3 -m pip install poetry` instead of the bootstrap script, but I get the same error when trying to run poetry.

What am I missing exactly?

jonapich:
I tried the same steps using python2 instead of python3. I get a different outcome:

    root@af0c0fe8f81c:~/.poetry/bin# ./poetry --version
    /root/.poetry/lib/poetry/_vendor/py2.7/subprocess32.py:149: RuntimeWarning: The _posixsubprocess module is not being used. Child process reliability may suffer if your program uses threads.
      "program uses threads.", RuntimeWarning)
    Poetry version 1.0.0
[...]

Issue #721, opened 2018-12-09
Include installation instructions for python3


nottrobin:
- [x] I have searched the [issues](https://github.com/sdispater/poetry/issues) of this repo and believe that this is not a duplicate.

## Issue

The README states as the ideal way to install Poetry:

``` bash
curl -sSL https://raw.githubusercontent.com/sdispater/poetry/master/get-poetry.py | python
```

But, on an Ubuntu Xenial system with only Python 3 installed, this leads to the error `bash: python: command not found`. If I simply pipe the curl command through `python3` instead then it succeds, but then I get:

``` bash
$ poetry install
/usr/bin/env: 'python': No such file or directory
```

This is because the top of the installed file at `$HOME/.poetry/bin/poetry` contains `#! /usr/bin/env python`.

If, instead, I install poetry with `pip3 install poetry`, then `head -n 1 $(which poetry)` shows `#!/usr/bin/python3` at the top of the file, which works perfectly.

Ubuntu, from 18.04 Xenial onwards, includes only Python 3 by default. And the minimal version of Xenial doesn't come with Python out of the box at all.

If you install Python 3 on Ubuntu, the binary that's installed is `python3`. `python` doesn't exist unless you've also installed Python 2. This [deliberately](https://wiki.ubuntu.com/Python/3) follows [PEP 394](https://legacy.python.org/dev/peps/pep-0394/):

> - python2 will refer to some version of Python 2.x.
> - python3 will refer to some version of Python 3.x.
> - for the time being, all distributions should ensure that python refers to the same target as python2.
> - however, end users should be aware that python refers to python3 on at least Arch Linux (that change is what prompted the creation of this PEP), so python should be used in the shebang line only for scripts that are source compatible with both Python 2 and 3.
> - in preparation for an eventual change in the default version of Python, Python 2 only scripts should either be updated to be source compatible with Python 3 or else to use python2 in the shebang line.

Although the PEP states "python should be used in the shebang line only for scripts that are source compatible with both Python 2 and 3" I'm not sure that this is compatible with the statement "all distributions should ensure that python refers to the same target as python2", because this means that if a system doesn't have Python 2 then the `python` binary won't exist. I think scripts have to do their best to use `python3` if they know it's supported, and most I've encountered do.

I think maybe it would be good if the suggested install command was:

``` bash
curl -sSL https://raw.githubusercontent.com/sdispater/poetry/master/get-poetry.py | ( python3 || python )
```

[...]

PR #3031, opened 2020-10-01
Don't swallow ImportError in temporary_directory()

remram44:
If the context wrapped by the `temporary_directory()` context manager raised `ImportError` (for example because `distutils.util` cannot be imported, #721 #1837), it would previously keep going, causing a `RuntimeError` from contextlib:

```
RuntimeError: generator didn't stop after throw()
```

# Pull Request Check List

Resolves: #3026


abn (approved):
Thank you for the fix @remram44 - good find!

Issue #3028, opened 2020-10-01
Poetry 1.1.0 Unable to Resolve Private Packages from pre-1.1.0 poetry.lock


xM8WVqaG:
- [x] I am on the [latest](https://github.com/python-poetry/poetry/releases/latest) Poetry version.
- [x] I have searched the [issues](https://github.com/python-poetry/poetry/issues) of this repo and believe that this is not a duplicate.
- [x] If an exception occurs when executing a command, I executed it again in debug mode (`-vvv` option).

- **OS version and name**: `Ubuntu 18.04 bionic`
- **Poetry version**: `Poetry version 1.0.10`
- **Link of a [Gist](https://gist.github.com/) with the contents of your pyproject.toml file**: `N/A`

## Issue

Starting with 1.1.0 of Poetry `poetry install` is failing for some of our private packages:

```
$ poetry install --no-interaction --no-ansi
Creating virtualenv project-name-nxCwz_m9-py3.8 in /builds/REDACTED/.cache/poetry/virtualenvs
Installing dependencies from lock file
Warning: The lock file is not up to date with the latest changes in pyproject.toml. You may be getting outdated dependencies. Run update to update them.
Package operations: 85 installs, 0 updates, 0 removals
  • Installing six (1.15.0)
<SNIP>
  • Installing private-package-1 (0.2.1)
  RuntimeError
  Unable to find installation candidates for private-package-1 (0.2.1)
  at /usr/local/lib/python3.8/site-packages/poetry/installation/chooser.py:72 in choose_for
       68│ 
       69│             links.append(link)
       70│ 
       71│         if not links:
    →  72│             raise RuntimeError(
       73│                 "Unable to find installation candidates for {}".format(package)
       74│             )
       75│ 
       76│         # Get the best link
  • Installing flake8-print (3.1.4)
  • Installing hypothesis (5.26.0)
  • Installing isort (4.3.21)
  • Installing jsonschema (3.2.0)
  • Installing mock (4.0.2)
  • Installing mypy (0.780)
  • Installing networkx (2.4)
  • Installing pytest-cov (2.10.1)
  • Installing private-package-2 (0.1.6)
  RuntimeError
  Unable to find installation candidates for private-package-2 (0.1.6)
  at /usr/local/lib/python3.8/site-packages/poetry/installation/chooser.py:72 in choose_for
       68│ 
       69│             links.append(link)
       70│ 
       71│         if not links:
    →  72│             raise RuntimeError(
[...]

PR #3034, opened 2020-10-01
lock: allow for no-update refresh of lock files

abn:
With the changes to the lock file it is important that users are able to refresh their lock files without updating compatible versions.

Relates-to: #3028


sdispater (changes requested):
Looks good to me 👍 

Nonetheless, we should add some tests to ensure everything is in order.

abn:
@sdispater added a test case for this as well

sdispater (approved):

Issue #3022, opened 2020-10-01
Dependency resolution failure due to misunderstanding extras with incompatible variants


michamos:
- [x] I am on the [latest](https://github.com/python-poetry/poetry/releases/latest) Poetry version.
- [x] I have searched the [issues](https://github.com/python-poetry/poetry/issues) of this repo and believe that this is not a duplicate.
- [x] If an exception occurs when executing a command, I executed it again in debug mode (`-vvv` option).

- **OS version and name**: Ubuntu 18.04.5 LTS
- **Poetry version**: 1.1.0
- **Link of a [Gist](https://gist.github.com/) with the contents of your pyproject.toml file**: https://gist.github.com/michamos/6d86f62a18e54ea98754c0084c6e6d47

## Issue

When trying to install `invenio-search` with `extras = ["elasticsearch7"]`, dependency resolution fails. From the log of [`poetry install -vvv`](https://gist.github.com/michamos/6d86f62a18e54ea98754c0084c6e6d47#file-poetry-install-log-txt-L13-L20), it looks like `poetry` thinks that all `elaststicsearch*` extras should be installed simultaneously, whereas they are mutually exclusive in the `invenio-search` [`setup.py`](https://github.com/inveniosoftware/invenio-search/blob/f3d6768846ce0cda59dbfa1c90524fc8ea77d723/setup.py#L31-L62).

PR #3046, opened 2020-10-02
provider: ensure ony activated extras are used

abn:
Resolves: #3022

### DIFF
--- a/poetry/puzzle/provider.py
+++ b/poetry/puzzle/provider.py
@@ -432,7 +432,7 @@ class Provider:
                 self._pool.package(
                     package.name,
                     package.version.text,
-                    extras=package.dependency.extras,
+                    extras=list(package.dependency.extras),
                     repository=package.dependency.source_name,
                 ),
             )
@@ -478,12 +478,12 @@ class Provider:
             if self._env and not dep.marker.validate(self._env.marker_env):
                 continue
 
-            if (
-                dep.is_optional()
-                and dep.name not in optional_dependencies
-                and not package.is_root()
-            ):
-                continue
+            if not package.is_root():
+                if (dep.is_optional() and dep.name not in optional_dependencies) or (
+                    dep.in_extras
+                    and not set(dep.in_extras).intersection(package.dependency.extras)
+                ):
+                    continue
 
             _dependencies.append(dep)
 
--- a/tests/puzzle/test_solver.py
+++ b/tests/puzzle/test_solver.py
@@ -449,6 +449,163 @@ def test_solver_returns_extras_if_requested(solver, repo, package):
     assert ops[0].package.marker.is_any()
 
 
+@pytest.mark.parametrize(("enabled_extra",), [("one",), ("two",), (None,)])
+def test_solver_returns_extras_only_requested(solver, repo, package, enabled_extra):
+    extras = [enabled_extra] if enabled_extra is not None else []
+
+    package.add_dependency(Factory.create_dependency("A", "*"))
+    package.add_dependency(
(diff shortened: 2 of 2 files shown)


poetry/puzzle/provider.py
```diff
@@ -478,12 +478,11 @@ def complete_package(
...
-            if (
-                dep.is_optional()
-                and dep.name not in optional_dependencies
-                and not package.is_root()
-            ):
-                continue
+            if not package.is_root():
+                if (
+                    dep.is_optional() and dep.name not in optional_dependencies
+                ) or not set(dep.in_extras).issubset(package.dependency.extras):
```
michamos, line 484: Hi, thanks for working on the issue I reported! I don't know much about the inner workings of `poetry`, but i was curious about the fix and had a look. If I'm interpreting this correctly, you now also skip a dependency if the set of extras in which it's declared isn't a subset of the extras flagged for installation.

If my interpretation is correct, I believe this should rather be 
```suggestion
                ) or not set(dep.in_extras).intersection(package.dependency.extras):
```
Case in point: package `C` declared in both extras `one` and `two` of package `B`, but only `B[one]` declared as a dependency for `A`. This should cause `C` to be selected.

If I completely misunderstood, feel free to dismiss my comment.

abn: @michamos I do not fully follow the scenario, just to make sure; does this case cover it?

https://github.com/python-poetry/poetry/pull/3046/commits/205a1796243f3e00d0b1ff042a93b91d87231b12

michamos: I don't think it does. I'm not sure how to express it in the tests directly, so I'll write it down using `pyproject.toml` syntax. What I mean is the following for `B`:
```toml
[tool.poetry]
name = "B"

[tool.poetry.dependencies]
C = { version = "1.0", optional = true }

[tool.poetry.extras]
one = ["C"]
two = ["C"]
```
then have the `pyproject.toml` for `A` contain
```toml
B = { version = "1.0", extras = ["one"] }
```

I believe your code won't select C as a dependency as `bool({"one", "two"}.issubset({"one"})) is False`.

abn: So, this case? https://github.com/python-poetry/poetry/pull/3046/commits/50dc9a09da7d7ac75223d29602cb8c1af458b2e0#diff-4e7cde4641210ec491c396893ff273d0R509

tests/puzzle/test_solver.py
```diff
@@ -449,6 +449,168 @@ def test_solver_returns_extras_if_requested(solver, repo, package):
...
+    package_b = get_package("B", "1.0")
+    package_c = get_package("C", "1.0")
+
+    dep10 = get_dependency("C", "*", optional=True)
+    dep10._in_extras.append("one")
+    dep10.marker = parse_marker("extra == 'one'")
+
+    dep20 = get_dependency("C", "*", optional=True)
+    dep20._in_extras.append("two")
+    dep20.marker = parse_marker("extra == 'two'")
```
michamos, line 524: If this is really equivalent to my example in https://github.com/python-poetry/poetry/pull/3046/files#r498772216 it would work. Naïvely I would think it rather corresponds to having only one dependency object but with `dep._in_extras == ["one", "two"]` (otherwise there's no point in it being a `list`), but I don't know the code well-enough (or at all for that matter) to be confident about it.

abn: Ah, I follow. Good catch mistake on my part. :)

finswimmer (approved):

Issue #3048, opened 2020-10-02
poetry lock has an issue with removing the marker caveat on certain packages


xM8WVqaG:
- [x] I am on the [latest](https://github.com/python-poetry/poetry/releases/latest) Poetry version.
- [x] I have searched the [issues](https://github.com/python-poetry/poetry/issues) of this repo and believe that this is not a duplicate.
- [x] If an exception occurs when executing a command, I executed it again in debug mode (`-vvv` option).

- **OS version and name**: `Ubuntu 18.04 bionic`
- **Poetry version**: `master-04833bc` (1.1.0)
- **Link of a [Gist](https://gist.github.com/) with the contents of your pyproject.toml file**: N/A

## Issue

Following up to #3028, I grabbed master to run `poetry lock --no-update` on every project we have to fix the lockfile.

However, after running `poetry lock --no-update` on a package that installed `docker` it removed the `win32` caveat breaking the install.

Source of `pywin32`:

```
$ poetry show --tree
|-- docker >=4.2.1,<5.0.0
|   |-- pypiwin32 223 
|   |   `-- pywin32 >=223 
|   |-- requests >=2.14.2,<2.18.0 || >2.18.0 
|   |   |-- certifi >=2017.4.17 
|   |   |-- chardet >=3.0.2,<4 
|   |   |-- idna >=2.5,<3 
|   |   `-- urllib3 >=1.21.1,<1.25.0 || >1.25.0,<1.25.1 || >1.25.1,<1.26 
|   |-- six >=1.4.0 
|   `-- websocket-client >=0.32.0 
|       `-- six * (circular dependency aborted here)
```

```
$ poetry install
Installing dependencies from lock file

Package operations: 65 installs, 0 updates, 0 removals

<snip>
  • Installing pywin32 (228): Failed

  RuntimeError

  Unable to find installation candidates for pywin32 (228)

  at /REDACTED/venv/lib/python3.8/site-packages/poetry/installation/chooser.py:72 in choose_for
       68│ 
       69│             links.append(link)
       70│ 
       71│         if not links:
    →  72│             raise RuntimeError(
       73│                 "Unable to find installation candidates for {}".format(package)
       74│             )
       75│ 
       76│         # Get the best link

  • Installing rsa (4.6)
  • Installing smmap (3.0.4)
```

Here is the relevant diff on the `poetry.lock`:

```diff
 [[package]]
-category = "dev"
-description = "Python for Window Extensions"
[...]

PR #3051, opened 2020-10-02
lock: resolve dependencies on no-update

abn:
This change ensures that we resolve dependencies to correctly propagate
markers when using `lock --no-update`.

Resolves: #3048


xM8WVqaG:
I can confirm that this branch fixes #3048

finswimmer (approved):

Issue #3069, opened 2020-10-03
publish: raise an error on redirects due to trailing slashes


abn:
This issue was originally brought up by @sztomi in #858, see referenced comments.

When encountering a `30X` error, requests library treats it as an "ok" response. And hence Poetry will not raise an error, but instead fail silently. We should report this error similar to any other `HTTPError` raised (#2285) and upload should fail.

Note that we do not want to explicitly allow redirects on POST requests or try to handle this scenario in any "smart" way. we simply want to error out gracefully.

### References
- https://github.com/python-poetry/poetry/issues/858#issuecomment-703170627
- https://github.com/python-poetry/poetry/issues/858#issuecomment-520239647

PR #3083, opened 2020-10-05
publish: raise an error on redirects

ObserverOfTime:
Resolves: #3069

---

I have a few questions regarding my changes:

* Should I add a helpful note in the documentation?
* Does the uploader need to also handle 302 redirects?
* Should the error message be consistent with 4xx errors?

### DIFF
--- a/poetry/publishing/uploader.py
+++ b/poetry/publishing/uploader.py
@@ -265,13 +265,24 @@ class Uploader:
                         allow_redirects=False,
                         headers={"Content-Type": monitor.content_type},
                     )
-                if dry_run or resp.ok:
+                if dry_run or 200 <= resp.status_code < 300:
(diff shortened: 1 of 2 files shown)


poetry/publishing/uploader.py
```diff
@@ -265,13 +265,22 @@ def _upload_file(
                         allow_redirects=False,
                         headers={"Content-Type": monitor.content_type},
                     )
-                if dry_run or resp.ok:
+                if dry_run or resp.status_code == 200:
```
abn, line 268: As was mentioned on the issue; not all indices return 200 on suceess. And hence why it was recommended to use ok and not 30x. Alternatively accept the 200 range.

ObserverOfTime:
Any idea why the CI is failing?

Issue #3087, opened 2020-10-05
key for sequential installation


Ver1Sus:
The new version of Poetry 1.1.1 installs dependency parallel - but this feature broke my programm

I use `websocket` and `websocket-client`, where `websocket-client` extends `websocket` by replacin  `__init__.py` and redefines WebSocket class, adding ssl support. 

In version 1.0.0 I installed `websocket` -> `websocket-client`, but now, in CI during rpm creation, the libraries were installed as `websocket-client` -> `websocket`, and after that, rpm has the old `WebSocket` class without ssl support

abn:
@Ver1Sus shouln't websocker-client depend on websocket? Or does it still order incorrectly when this is already the case?

Ver1Sus:
> shouln't websocker-client depend on websocket?

Should, but it doesn't) I fixed this, by using `poetry config experimental.new-installer false` before `poetry install` - and installation become consistent.

So, this issue may be closed, but maybe need add some feature in `pyproject.toml` to set prioritet on the installed packages

PR #3088, opened 2020-10-05
config: add support for installer.parallel

abn:
This change allows users to disable parallel execution while using the
new installer.

Resolves: #3087


sdispater (approved):
Looks good to me 👍

PR #3091, opened 2020-10-05
Remove executable bit from files without shebangs

hroncok:
Not sure if this goes to develop or master, happy to rebase.


abn:
oh wow; thanks @hroncok - I did not even notice these - woops.

Master is good thanks.

abn (approved):

PR #3095, opened 2020-10-06
Fix #2658: Run all uninstalls serially

PetterS:
Resolves: #2658 

- [x] Added **tests** for changed code.

This is one way of fixing the issue. Most of the changes are to make this code testable.

@abn

Issue #2658, opened 2020-07-12
Race condition in new installer (1.1.0rc1) with --remove-untracked


PetterS:
To reproduce:
- Add `ipython = "*"` as a dev dependency to a repo.
- `poetry lock`
- `poetry install`
- Remove it again.
- `poetry lock`
- `poerty install --remove-untracked`

I get this:
```
Installing dependencies from lock file

Package operations: 0 installs, 0 updates, 15 removals

  • Removing backcall (0.2.0)
  • Removing decorator (4.4.2)
  • Removing ipython (7.16.1): Failed

  EnvCommandError

  Command ['/home/petter/.cache/pypoetry/virtualenvs/openai-test-cYZNiyR8-py3.8/bin/python', '-m', 'pip', 'uninstall', 'ipython', '-y'] errored with the following return code 2, and output:
  ERROR: Exception:
  Traceback (most recent call last):
    File "/home/petter/.cache/pypoetry/virtualenvs/openai-test-cYZNiyR8-py3.8/lib/python3.8/site-packages/pip/_internal/cli/base_command.py", line 186, in _main
      status = self.run(options, args)
    File "/home/petter/.cache/pypoetry/virtualenvs/openai-test-cYZNiyR8-py3.8/lib/python3.8/site-packages/pip/_internal/commands/uninstall.py", line 51, in run
      session = self.get_default_session(options)
    File "/home/petter/.cache/pypoetry/virtualenvs/openai-test-cYZNiyR8-py3.8/lib/python3.8/site-packages/pip/_internal/cli/req_command.py", line 74, in get_default_session
      self._session = self.enter_context(self._build_session(options))
    File "/home/petter/.cache/pypoetry/virtualenvs/openai-test-cYZNiyR8-py3.8/lib/python3.8/site-packages/pip/_internal/cli/req_command.py", line 84, in _build_session
      session = PipSession(
    File "/home/petter/.cache/pypoetry/virtualenvs/openai-test-cYZNiyR8-py3.8/lib/python3.8/site-packages/pip/_internal/network/session.py", line 241, in __init__
      self.headers["User-Agent"] = user_agent()
    File "/home/petter/.cache/pypoetry/virtualenvs/openai-test-cYZNiyR8-py3.8/lib/python3.8/site-packages/pip/_internal/network/session.py", line 159, in user_agent
      setuptools_version = get_installed_version("setuptools")
[...]

Issue #3086, opened 2020-10-05
Editable installs sometimes cannot be imported (poetry 1.1.0)


ponas:
- [x] I am on the [latest](https://github.com/python-poetry/poetry/releases/latest) Poetry version.
- [x] I have searched the [issues](https://github.com/python-poetry/poetry/issues) of this repo and believe that this is not a duplicate.
- [x] If an exception occurs when executing a command, I executed it again in debug mode (`-vvv` option).

- **OS version and name**: Fedora 32 (also reproduced with Docker base image python:3-slim)
- **Poetry version**: 1.1.1
- **Link of a [Gist](https://gist.github.com/) with the contents of your pyproject.toml file**: https://gist.github.com/ponas/f02c8228a4e366bc255f4f4d1dd32cec

## Issue
This issue first appears with Poetry 1.1.0. I cannot reproduce this issue on 1.0.9.

When Poetry installs the two packages from git, no exceptions occur, but _sometimes_ one of them cannot be imported in the resulting environment:
```
$ poetry install
[snip]
  • Installing cerebrum-client (1.9.2 95a1839)
  • Installing tofh (2.0.6 8f9dac0)
$ poetry run python -c 'import tofh'
$ poetry run python -c 'import cerebrum_client'
Traceback (most recent call last):
  File "<string>", line 1, in <module>
ModuleNotFoundError: No module named 'cerebrum_client'
```

I was unable to reproduce this by installing packages such as Django, requests and flask via git, making me think this may be related to something in our `setup.py`s. Might not be a bug in poetry, but this behavior is new :-)

## Workaround
Current workaround is to disable the new installer in favour of the old one. (https://github.com/python-poetry/poetry/issues/3086#issuecomment-703692203)

mchesterkadwell:
I am experiencing the same issue since the release of 1.1.1 (couldn't reproduce myself on 1.1.0). 

My Docker image build has 5 git repos pulled in by SSH from Bitbucket. Poetry reports they are installing, but after running `pip list` in a container built from the image it shows one or more of them will not actually be installed, seemingly at random. I have run the build multiple times and each time get a different combination of correctly installed packages.

I get the same `ModuleNotFoundError: No module named 'foo'` when trying to run the application.

ponas:
A temporary workaround is to disable the new installer with:
`POETRY_EXPERIMENTAL_NEW_INSTALLER="" poetry install`
or 
`poetry config experimental.new-installer false`

[...]

PR #3099, opened 2020-10-06
executor: execute parallel unsafe ops serially

abn:
Resolves: #2658 #3086
Closes: #3095

### DIFF
--- a/poetry/installation/executor.py
+++ b/poetry/installation/executor.py
@@ -107,14 +107,33 @@ class Executor(object):
         self._sections = OrderedDict()
         for _, group in groups:
             tasks = []
+            serial_operations = []
             for operation in group:
                 if self._shutdown:
                     break
+
+                # Some operations are unsafe, we mus execute them serially in a group
+                # https://github.com/python-poetry/poetry/issues/3086
+                # https://github.com/python-poetry/poetry/issues/2658
+                #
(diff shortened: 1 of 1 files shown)


poetry/installation/executor.py
```diff
@@ -107,14 +107,30 @@ def execute(self, operations):  # type: (Operation) -> int
...
+                    serial_operations.append(operation)
+                    continue
+
                 tasks.append(self._executor.submit(self._execute_operation, operation))
 
             try:
                 wait(tasks)
+
+                for operation in serial_operations:
+                    wait([self._executor.submit(self._execute_operation, operation)])
```
PetterS, line 132: Don't have to submit and wait here. Can just call `_execute_operation`.

PetterS (commented):
Thanks a lot for this! 👍 

I think that in order to test this, some refactoring of the code is needed.

poetry/installation/executor.py
abn: Yup, but I wanted keep calls consistent via `self._executor` for now.

abn:
> I think that in order to test this, some refactoring of the code is needed.

Race conditions are inherently hard to test. I would be happy with no regressions and manual verification for now.

finswimmer (approved):
LGTM :+1:

Issue #3079, opened 2020-10-05
PermissionError installing root package in 1.1.0


agoose77:
- [x] I am on the [latest](https://github.com/python-poetry/poetry/releases/latest) Poetry version.
- [x] I have searched the [issues](https://github.com/python-poetry/poetry/issues) of this repo and believe that this is not a duplicate.
- [x] If an exception occurs when executing a command, I executed it again in debug mode (`-vvv` option).

- **OS version and name**: Docker (Ubuntu 20.04)
- **Poetry version**: 1.1.0
- **Link of a [Gist](https://gist.github.com/) with the contents of your pyproject.toml file**: https://gist.github.com/agoose77/5b08db26a49c8119a9a943583ef724b3
- **Link to a DockerFile**: https://gist.github.com/agoose77/647769f65494f6a82ffc3f625eca0457

## Issue
This issue follows from #2996 

The line that install the `dist-info` directory uses a single env directory: 

https://github.com/python-poetry/poetry/blob/ae6f574bcb32c42675e34238be6498d585efee4f/poetry/masonry/builders/editable.py#L190

 It seems to me like the fix would be similar to #3000

Thank you for all of the hard work Poetry team.

abn:
Verified against the 1.1 branch.

```console
podman run --rm -i --entrypoint bash python:3.7 <<EOF
pip install --quiet git+https://github.com/python-poetry/poetry.git@1.1
adduser --disabled-password --gecos "User" --uid 1000 user
su - user
poetry new foobar
cd foobar
export PYTHONDONTWRITEBYTECODE=1
export POETRY_VIRTUALENVS_CREATE=0
poetry install -vvv
EOF
```

abn:
@agoose77 can you try #3107 please?

agoose77:
@abn this passes my existing build-step test. Thanks for the quick turn around!

PR #3107, opened 2020-10-06
Improved support for system environment

abn:
**utils/env: better support system site packages dir**
This change improves handling of site-packages under system env, by
gracefully handling fallbacks to user site when required and possible.

Resolves: #3079

This was verified using the following Dockerfile.

```Dockerfile
FROM python:3.8

RUN useradd -m user
RUN pip install git+https://github.com/abn/poetry.git@issue/3079

USER user
WORKDIR /home/user

RUN poetry config virtualenvs.create false
RUN poetry new foobar

WORKDIR /home/user/foobar

RUN poetry install
RUN poetry run pip show foobar
RUN ls -all /home/user/.local/lib/python3.8/site-packages
```

**editable: use writable script dir for system env**

This change ensures that, When using system environment, poetry falls
back to `userbase` if default location is not writable.

**utils/env: ensure we create user directories**

### DIFF
--- a/poetry/installation/pip_installer.py
+++ b/poetry/installation/pip_installer.py
@@ -113,7 +113,9 @@ class PipInstaller(BaseInstaller):
             raise
 
         # This is a workaround for https://github.com/pypa/pip/issues/4176
-        nspkg_pth_file = self._env.site_packages / "{}-nspkg.pth".format(package.name)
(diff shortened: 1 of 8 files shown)


sdispater (approved):
Looks good to me 👍

Issue #3116, opened 2020-10-07
poetry show fails to resolve private package with extras


Agalin:
- [x] I am on the [latest](https://github.com/python-poetry/poetry/releases/latest) Poetry version.
- [x] I have searched the [issues](https://github.com/python-poetry/poetry/issues) of this repo and believe that this is not a duplicate.
- [x] If an exception occurs when executing a command, I executed it again in debug mode (`-vvv` option).

- **OS version and name**: Windows 10, RHEL 7.8
- **Poetry version**: 1.1.0-1.1.2
- **Simple pyproject.toml with a private repo**:
```
[[tool.poetry.source]]
name = "private-repo"
url = "<repo url>"

[tool.poetry.dependencies]
python = "^3.6,<3.9"
private-dependency = {extras = ["extras"], version = "^1.20.35723", source = "private-repo"}
```

## Issue

When running `poetry show` with a private package that has extras installed command fails with SolverProblemError.

For this package there are multiple extras, some with just public packages, some with mix of public and private. It doesn't matter which one is required, command fails even if only public packages are required by extras marker but succeeds if there are no extras specified to install.

Specifying extras for public packages doesn't break the command. Generating lock file from scratch (`poetry lock`) doesn't change anything.

Install, add, lock are working correctly for this dependency. Only show fails.

Version 1.0.10 works correctly.

<details>
<summary><b>Output:</b></summary>
<p>

```

Using virtualenv: C:\Users\user\AppData\Local\pypoetry\Cache\virtualenvs\private-package-UtOW8WEE-py3.6

  Stack trace:

  6  c:\users\user\appdata\local\programs\python\python36\lib\site-packages\clikit\console_application.py:131 in run
      129│             parsed_args = resolved_command.args
      130│
    → 131│             status_code = command.handle(parsed_args, io)
      132│         except KeyboardInterrupt:
      133│             status_code = 1

  5  c:\users\user\appdata\local\programs\python\python36\lib\site-packages\clikit\api\command\command.py:120 in handle
      118│     def handle(self, args, io):  # type: (Args, IO) -> int
      119│         try:
    → 120│             status_code = self._do_handle(args, io)
      121│         except KeyboardInterrupt:
      122│             if io.is_debug():

  4  c:\users\user\appdata\local\programs\python\python36\lib\site-packages\clikit\api\command\command.py:171 in _do_handle
      169│         handler_method = self._config.handler_method
      170│
[...]

PR #3117, opened 2020-10-07
show: ignore dependency source when finding package

abn:
Resolves: #3116


finswimmer (changes requested):
Any chance to add a useful test for it?

abn:
TBH; testing this will be convoluted. And far more effort than worth it I feel.

finswimmer (approved):

PR #3118, opened 2020-10-07
export: restore -e for develop exports

mrcljx:
# Pull Request Check List

The `line` variable containing `-e` was consistently overwritten.

### DIFF
--- a/tests/utils/test_exporter.py
+++ b/tests/utils/test_exporter.py
@@ -107,8 +107,8 @@ def test_exporter_can_export_requirements_txt_with_standard_packages(
         content = f.read()
 
     expected = """\
-bar==4.5.6
-foo==1.2.3
+-e bar==4.5.6
+-e foo==1.2.3
 """
 
     assert expected == content
@@ -162,9 +162,9 @@ def test_exporter_can_export_requirements_txt_with_standard_packages_and_markers
         content = f.read()
 
     expected = """\
-bar==4.5.6
-baz==7.8.9; sys_platform == "win32"
-foo==1.2.3; python_version < "3.7"
+-e bar==4.5.6
+-e baz==7.8.9; sys_platform == "win32"
+-e foo==1.2.3; python_version < "3.7"
 """
 
     assert expected == content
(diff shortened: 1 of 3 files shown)


abn (changes requested):
Couple of issues here, 

1. I am not sure if `-e` works well for PEP 517 directory dependencies
2. I do not think the checks are correct for this, only directory dependencies and vcs dependencies support editable installs.

sirlantis:
> 1. I am not sure if `-e` works well for PEP 517 directory dependencies

Shouldn't the user then have removed `develop = true`?

> 2. I do not think the checks are correct for this, only directory dependencies and vcs dependencies support editable installs.

Will make the change.

abn:
> Shouldn't the user then have removed `develop = true`?

I guess this was existing behaviour. The the difference is poetry can handle PEP 517 packages, so a user might have `develop = true` set for these. Either way, since this is not changing `1.0.10` behaviour it should be okay.

tests/utils/test_exporter.py
```diff
@@ -497,7 +497,7 @@ def test_exporter_can_export_requirements_txt_with_git_packages(tmp_dir, poetry)
         content = f.read()
 
     expected = """\
-foo @ git+https://github.com/foo/foo.git@123456
+-e foo @ git+https://github.com/foo/foo.git@123456
```
abn, line 500: Hmm, these are not marked as `develop` in the lock data above. This should not, be `-e` prefixed. I think there is a bug some place else. I wonder if https://github.com/python-poetry/poetry-core/pull/98 is related?

sirlantis: Ah, I felt like I saw a PR for this. I can wait for that to land and then rebase my change here.

abn: Would be great if you can test with that branch locally atleast to confirm this is the case. Else we will have to dig into it seperately.

sirlantis: Verified that against that branch the tests don't need updating.

sirlantis:
Converting to Draft until https://github.com/python-poetry/poetry-core/pull/98 is landed.

PR #3120, opened 2020-10-07
export: fix directory dependencies

sirlantis:
# Pull Request Check List

`pip` doesn't support `foo @ some/path/foo` directory dependencies.

Depends on #3118 (doesn't really depend on it, but will be in conflict), which is why this PR is in draft mode for now (and also includes the same changes).

### DIFF
--- a/poetry/utils/exporter.py
+++ b/poetry/utils/exporter.py
@@ -94,14 +94,16 @@ class Exporter(object):
                 or dependency.is_directory()
             )
 
-            if is_direct_reference:
-                line += requirement
-            else:
+            if not is_direct_reference:
                 line += "{}=={}".format(package.name, package.version)
                 if ";" in requirement:
                     markers = requirement.split(";", 1)[1].strip()
                     if markers:
                         line += "; {}".format(markers)
+            elif dependency.is_directory():  # pip doesn't support ` @ ` for directories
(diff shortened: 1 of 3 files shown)


poetry/utils/exporter.py
```diff
@@ -94,14 +94,16 @@ def _export_requirements_txt(
...
-                line = requirement
-            else:
-                line = "{}=={}".format(package.name, package.version)
+            if not is_direct_reference:
+                line += "{}=={}".format(package.name, package.version)
                 if ";" in requirement:
                     markers = requirement.split(";", 1)[1].strip()
                     if markers:
                         line += "; {}".format(markers)
+            elif dependency.is_directory():  # pip doesn't support ` @ ` for directories
```
abn, line 103: I do not think this should be done, because directory dependencies can be PEP 517 dependencies, which means `-e` will fail.

sirlantis: I don't follow. Having a line `foo @ some/path` in `requirements.txt` will break `pip install` with `ERROR: Invalid requirement: 'foo @ some/path'`? Should the exporter add a new PEP508 format besides `requirements.txt`?

abn: I think what is being discussed in #3121 is a more complete solution. But yes, the comments in #3118 are more current.

Should this be closed in favour of that?

PR #3124, opened 2020-10-07
Doc: poetry lock --no-update

bsitruk:
# Pull Request Check List

Resolves: #issue-number-here

### DIFF
--- a/docs/docs/cli.md
+++ b/docs/docs/cli.md
@@ -423,6 +423,12 @@ This command locks (without installing) the dependencies specified in `pyproject
 poetry lock
 ```
 
+To avoid updating pinned dependencies, you can use the `--no-update` option.
+
+```bash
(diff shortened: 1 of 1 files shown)


docs/docs/cli.md
```diff
@@ -423,6 +423,12 @@ This command locks (without installing) the dependencies specified in `pyproject
 poetry lock
 ```
 
+To avoid updating pinned dependencies, you can use the `--no-update` option.
```
paavanb, line 426: I think it would be better to match the pattern used in all the other commands, and add an "Options" section with a line for `--no-update`.

abn (changes requested):
As indicated by @paavanb the options seciton might be the way to go.

bsitruk:
Not relevant anymore.

Issue #3144, opened 2020-10-08
Poetry fails with KeyError if the PATH environment variable is not present


exhuma:
- [x] I am on the [latest](https://github.com/python-poetry/poetry/releases/latest) Poetry version.
- [x] I have searched the [issues](https://github.com/python-poetry/poetry/issues) of this repo and believe that this is not a duplicate.
- [x] If an exception occurs when executing a command, I executed it again in debug mode (`-vvv` option).

- **OS version and name**: Ubuntu 18.04
- **Poetry version**: 1.0.10

## Issue

When running in CI using a docker container the `PATH` environment variable is not set and causes an issue with poetry. Unfortunately I don't see any traceback. Here's a snipped showing the issue:

```
root@5d1e49d5433c:~/src# unset PATH
root@5d1e49d5433c:~/src# /usr/local/bin/poetry run -vvv pip install pip

[KeyError]
'PATH'
```

exhuma:
With the debugger I was able to get the traceback:

```
(Pdb) p traceback.print_tb(e.__traceback__)
  File "/usr/local/lib/python3.6/dist-packages/clikit/console_application.py", line 131, in run
    status_code = command.handle(parsed_args, io)
  File "/usr/local/lib/python3.6/dist-packages/clikit/api/command/command.py", line 120, in handle
    status_code = self._do_handle(args, io)
  File "/usr/local/lib/python3.6/dist-packages/clikit/api/command/command.py", line 163, in _do_handle
    self._dispatcher.dispatch(PRE_HANDLE, event)
  File "/usr/local/lib/python3.6/dist-packages/clikit/api/event/event_dispatcher.py", line 22, in dispatch
    self._do_dispatch(listeners, event_name, event)
  File "/usr/local/lib/python3.6/dist-packages/clikit/api/event/event_dispatcher.py", line 89, in _do_dispatch
    listener(event, event_name, self)
  File "/usr/local/lib/python3.6/dist-packages/poetry/console/config/application_config.py", line 92, in set_env
    env = env_manager.create_venv(io)
  File "/usr/local/lib/python3.6/dist-packages/poetry/utils/env.py", line 490, in create_venv
    env = self.get(reload=True)
  File "/usr/local/lib/python3.6/dist-packages/poetry/utils/env.py", line 333, in get
    return VirtualEnv(venv)
  File "/usr/local/lib/python3.6/dist-packages/poetry/utils/env.py", line 1045, in __init__
[...]

PR #3146, opened 2020-10-08
Fix issue when PATH is missing in os.environ

exhuma:
# Pull Request Check List

Resolves: #3144 

I'm not really sure any of the two check-boxes apply for this change. Is the effort of preparing `os.environ` for the unit-test really worth the effort? Considering that this might interfere with other tests, so it needs to be mocked or patched. I feel that all this effort for this small change is a bit over-the-top. The call itself (and thus a runtime syntax-error) is already covered by existing tests.

If you disagree, let me know and I will add the required test.

As for documentation, as this is an internal change not externally visible I don't think it warrants docs. Again, let me know if you disagree.

And finally, I decided to target `master` instead of `develop` for this PR because it is a bugfix on the current released version. But that's anyway something you can modify yourself 😉 

### DIFF
--- a/poetry/utils/env.py
+++ b/poetry/utils/env.py
@@ -1212,7 +1212,7 @@ class VirtualEnv(Env):
             del os.environ[key]
 
     def _updated_path(self):
-        return os.pathsep.join([str(self._bin_dir), os.environ["PATH"]])
+        return os.pathsep.join([str(self._bin_dir), os.environ.get("PATH", "")])
 
 
 class NullEnv(SystemEnv):


exhuma:
Not sure why the checks fail. I ran the pre-commit hooks and they all passed.

abn (approved):

abn:
@exhuma jfyi `develop` is defunct at the moment. If you want this in a bug fix release, please also create PR against `1.1` once this is merged.

exhuma:
Done in #3159

Issue #3134, opened 2020-10-07
Poetry links to python3 and not python3.x


ericriff:
- [X] I am on the [latest](https://github.com/python-poetry/poetry/releases/latest) Poetry version.
- [X] I have searched the [issues](https://github.com/python-poetry/poetry/issues) of this repo and believe that this is not a duplicate.
- [X] If an exception occurs when executing a command, I executed it again in debug mode (`-vvv` option).

- **OS version and name**: Ubuntu 16
- **Poetry version**: 1.1.1

## Issue
I have multiple versions of `python3` on my system and I use `update-alternatives` to choose where `python3` points to.

I specify the desired python version for each project in the `pyproject` file as follows:
```
[tool.poetry.dependencies]
python = "==3.5.2"
```
So in this case I enforce version 3.5.2. The problem is that poetry creates a symlink to `python3` and not `python3.5`, which works fine until I change the default python version:

```bash
$ tree .venv/bin
.venv/bin
├── activate
├── activate.csh
├── activate.fish
.
.
├── python -> /usr/bin/python3
├── python3 -> python
├── python3.5 -> python
.
.
```
If I specifically tell poetry to use `python3.5`, it creates the correct symlink:
**EDIT:** This is only true if the default version of `python` is not the one specified in the `pyprojet`. That means that, in this case, if the default version of `python3` is `3.5`, the following commands produce the same output as the previous ones.
```
$ poetry env use python3.5
$ poetry install
$ tree .venv/bin
.venv/bin
├── activate
├── activate.csh
├── activate.fish
.
.
├── python -> /usr/bin/python3.5
├── python3 -> python
├── python3.5 -> python
.
.
```
This is a huge problem for me since the python version should be enforced very strictly. I have taken a look at how `pipenv` handles this and it creates the symlink to the specific python version and not just plain `python3`.

I also don't see any reference to the python version in the `lockfile`. Is this normal?

Thanks!

ericriff:
I did some more digging and this behavior was introduced after dropping `venv` in favor of `virtualenv`, since they work in different ways: `venv` copies over the python executables to the environment whereas `virtualenv` creates symlinks. 

```
$ m -rf .venv
$ ~/.poetry/bin/poetry shell
The currently activated Python version 2.7.12 is not supported by the project (3.5.2).
Trying to find and use a compatible version. 
Using python3 (3.5.2)
Creating virtualenv ...
...
$ tree .venv/bin
.venv/bin
├── activate
[...]

PR #3157, opened 2020-10-09
add config option to force `--always-copy` parameter for virtualenv

finswimmer:
`virtualenv` uses symlink to the interpreter when creating a venv. While this reduces space and runtime this might lead to unexpected results, when the interpreter to which the symlink points get upgraded.

This PR introduce a config option `virtualenvs.options.always-copy`. If setting to `true` the `--always-copy` parameter is added to virtualenv command on creating the venv.

To keep the current behavior as default, this option is set to `false` by default.

# Pull Request Check List

Resolves: https://github.com/python-poetry/poetry/issues/3134

### DIFF
--- a/docs/docs/configuration.md
+++ b/docs/docs/configuration.md
@@ -34,6 +34,7 @@ which will give you something similar to this:
 cache-dir = "/path/to/cache/directory"
 virtualenvs.create = true
 virtualenvs.in-project = null
+virtualenvs.options.always-copy = true
 virtualenvs.path = "{cache-dir}/virtualenvs"  # /path/to/cache/directory/virtualenvs
 ```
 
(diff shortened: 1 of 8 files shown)


abn:
@finswimmer LGTM. Althought, I am wondering if we should names psace this into something like `virtualenvs.options.always-copy` or smilar. I feel we will have more options to be exposed. Alternatively we can also consider supporting `virtualenvs.options.*`, boolean beign flags that simply get transfored to `--<flag>`. This will also enable us to support options like `system-site` etc.

finswimmer:
> Althought, I am wondering if we should names psace this into something like virtualenvs.options.always-copy or smilar. I feel we will have more options to be exposed.

Yes, this sounds like a good idea.

> Alternatively we can also consider supporting virtualenvs.options.*, boolean beign flags that simply get transfored to --<flag>.

IMO that's a bit risky. I'm quite sure there will be people, who are able to create a config that will break poetry ;) So, we should carefully choose which flag we want to include.

abn (approved):

Issue #3097, opened 2020-10-06
Duplicate entries when running poetry update --dry-run


MarcoGlauser:
- [x] I am on the [latest](https://github.com/python-poetry/poetry/releases/latest) Poetry version.
- [x] I have searched the [issues](https://github.com/python-poetry/poetry/issues) of this repo and believe that this is not a duplicate.
- [ ] If an exception occurs when executing a command, I executed it again in debug mode (`-vvv` option).

- **OS version and name**: Ubuntu 20.04
- **Poetry version**: 1.1.1
- **Link of a [Gist](https://gist.github.com/) with the contents of your pyproject.toml file**: https://gist.github.com/MarcoGlauser/804cbf48001f7a7d1d94d6da9bc88071

## Issue
The command `poetry update --dry-run` duplicates the output of packages as seen below.
![image](https://user-images.githubusercontent.com/924855/95180307-173f3400-07c2-11eb-917d-230097698155.png)

I first noticed it in poetry 1.1.0. I could reproduce it on different machines with the attached gist.

ahlaw:
Hi, can I take this issue? I think I found the problem, when the `_dry_run` attribute is set, the executor writes after passing the `_should_write_operation` check [here](https://github.com/python-poetry/poetry/blob/6a96dd22f4cf22f6f6d5bdb36c44c02097d57bee/poetry/installation/executor.py#L181) and then again inside `_do_execute_operation` [here](https://github.com/python-poetry/poetry/blob/6a96dd22f4cf22f6f6d5bdb36c44c02097d57bee/poetry/installation/executor.py#L264). `_dry_run` should be removed from one of those checks.

abn:
@ahlaw most definitey! Feel free to create a draft PR for this if you want to discuss a solution.

PR #3168, opened 2020-10-10
Executor: Remove duplicate entry on dry-run

ahlaw:
Removes additional check caused dry runs to output operations a second time.

Resolves: [#3097](https://github.com/python-poetry/poetry/issues/3097)

### DIFF
--- a/poetry/installation/executor.py
+++ b/poetry/installation/executor.py
@@ -261,13 +261,16 @@ class Executor(object):
 
             return 0
 
-        if not self._enabled or self._dry_run:
+        if not self._enabled:
             self._io.write_line(
                 "  <fg=blue;options=bold>•</> {message}".format(
                     message=operation_message,
                 )
             )
 
+            return 0
+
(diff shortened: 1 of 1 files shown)


poetry/installation/executor.py
```diff
@@ -261,7 +261,7 @@ def _do_execute_operation(self, operation):
 
             return 0
 
-        if not self._enabled or self._dry_run:
+        if not self._enabled:
```
abn, line 264: I do not think this is the correct fix. Now, if the operation is a dry run we end up performing the operation since the `return 0` in this block is never hit.

abn (changes requested):

poetry/installation/executor.py
ahlaw: Yes, the check should still be there but not write if it is a dry run, sorry for missing that!

neersighted:
Closing this in favor of #4660 which is tested and takes a more idiomatic approach.

Issue #2620, opened 2020-07-03
Self install to PATH of other shell profiles


sceniclife:
Using a vanilla macOS Catalina v10.15.5. The default shell profile Terminal uses is zsh. (~/.zshrc)
Poetry does not automatically install to PATH for this type of set up (possibly others).
Could it be set up to handle the different types of shell profiles?

Example:
I installed poetry via curl:

```
curl -sSL https://raw.githubusercontent.com/python-poetry/poetry/master/get-poetry.py | python
```

And I receive the message:
```
Retrieving Poetry metadata

# Welcome to Poetry!

This will download and install the latest version of Poetry,
a dependency and package manager for Python.

It will add the `poetry` command to Poetry's bin directory, located at:

$HOME/.poetry/bin

This path will then be added to your `PATH` environment variable by
modifying the profile files located at:

$HOME/.profile
$HOME/.zprofile

You can uninstall at any time by executing this script with the --uninstall option,
and these changes will be reverted.

Installing version: 1.0.9
  - Downloading poetry-1.0.9-darwin.tar.gz (10.46MB)

Poetry (1.0.9) is installed now. Great!

To get started you need Poetry's bin directory ($HOME/.poetry/bin) in your `PATH`
environment variable. Next time you log in this will be done
automatically.

To configure your current shell run `source $HOME/.poetry/env`
```

I do not have either `$HOME/.profile` or `$HOME/.zprofile`, after install I run `poetry --version` and receive:
```
zsh: command not found: poetry
```

The ticket #507 was opened before but was closed since the user found that you can manually modify `~/.zshrc` and add `export PATH="$HOME/.poetry/bin:$PATH"` to get it to work. (verified it does work)

This ticket is to request the feature of having poetry automatically detect the shell profile and add to the PATH automatically.

abitrolly:
`zsh` doesn't create `.zprofile` automatically - https://www.reddit.com/r/zsh/comments/fjcd35/theres_no_zprofile_file_in_my_home_directory_in/ It also doesn't use `.profile` because it may be not compatible. `.zprofile` is also not loaded by graphical terminals - https://askubuntu.com/questions/708924/zprofile-is-not-read-on-graphical-login-when-user-s-default-shell-is-zsh

I see the solution to add to `.zshrc `, because it is run for both login and non-login shells https://www.reddit.com/r/zsh/comments/e882c4/what_is_the_difference_between_zshrc_and_zprofile/

PR #3169, opened 2020-10-10
Use `.zshrc` instead of `.zprofile`

abitrolly:
Because `.zprofile` is only run for login shells, it is not used
by graphical terminals. `.zprofile` is also not created by default.

Resolves: #2620

  No tests for `get-poetry.py`

  Not documented as far as I can see.

### DIFF
--- a/get-poetry.py
+++ b/get-poetry.py
@@ -903,7 +903,7 @@ class Installer:
 
         if "zsh" in SHELL:
             zdotdir = os.getenv("ZDOTDIR", HOME)
-            profiles.append(os.path.join(zdotdir, ".zprofile"))
+            profiles.append(os.path.join(zdotdir, ".zshrc"))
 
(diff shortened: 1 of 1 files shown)


abitrolly:
There raises a concern that if somebody had a `.zprofile` file before and installed `poetry`, then uninstall won't clean it up. However, the chance for `.zprofile` to be present is low, because it is only created by advanced Linux users for login shell. And the probability for those users uninstalling poetry using installer is even lower.

I still could add the code to cleanup Poetry lines from all profiles that were once used. Though I would  factor out the PATH management into a separate class if anybody would be willing to review that.

mvoitko (approved):

abitrolly:
This is also a `hacktoberfest` related issue.

@mvoitko how'd think are there any chances to get this merged until the end of Hacktoberfest?

stephsamson (approved):

Issue #3154, opened 2020-10-09
Cirrus CI: Wait for release to be ready before attempting to push an artifact to github releases


abn:
Presently, the cirrus ci release logic checks if release is ready, if not it simply skips the upload. This needs to be changed to ensure that we wait for a reasonable amount before skipping.

https://github.com/python-poetry/poetry/blob/6a96dd22f4cf22f6f6d5bdb36c44c02097d57bee/.cirrus.yml#L53-L59

saifalsowaidi:
relesea assistant 12 oct 2020

Secrus:
The code that the issue references no longer exists. Closing

PR #3181, opened 2020-10-12
Support retries when checking associated release

cereblanco:
# Pull Request Check List

Resolves: https://github.com/python-poetry/poetry/issues/3154

Currently, the cirrus ci release logic checks if release is ready, if not it simply skips the upload.
This PR will try to check if release is ready for at most 20 times, and `sleep 30` seconds within those intervals

Risks: maybe it's an overkill solution? 😅 

[ ] Added **tests** for changed code.
[ ] Updated **documentation** for changed code.

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -50,12 +50,19 @@ release_task:
   upload_script: |
     #!/usr/bin/env bash
 
-    if [[ "$CIRRUS_RELEASE" == "" ]]; then
+    NUM_TRIES=0
+    until [ $NUM_TRIES -eq 5 ]
+    do
+      echo "Retrying to find a release associated with this tag"
       CIRRUS_RELEASE=$(curl -sL https://api.github.com/repos/$CIRRUS_REPO_FULL_NAME/releases/tags/$CIRRUS_TAG | jq -r '.id')
-      if [[ "$CIRRUS_RELEASE" == "null" ]]; then
+      [[ "$CIRRUS_RELEASE" != "null" ]] && break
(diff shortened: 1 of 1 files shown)


.cirrus.yml
```diff
@@ -50,12 +50,19 @@ release_task:
   upload_script: |
     #!/usr/bin/env bash
 
-    if [[ "$CIRRUS_RELEASE" == "" ]]; then
+    NUM_TRIES=0
+    until [ $NUM_TRIES -eq 5 ]
```
abn, line 54: Considering in the best case scenarios cirrus takes around 4 minutes to get here and the release workflow in github actions takes on average around 8 minutes to get to "Create release" step, I would say we should probably wait at the very least for 4 minutes. That said, I am more inclined to suggest a 10 minute hard limit at 30 second intervals. So 20 retries. The current retries might be insufficient for what we need.

abn (changes requested):

.cirrus.yml
cereblanco: @abn thank you for the feedback. i updated the PR here: https://github.com/python-poetry/poetry/pull/3181/commits/a0847594097b2a3be6aa3babfeaa049be15a179f. wdyt?

abn: One way to test it is to create a cirrus job on your own fork and link the test runs. However, I am okay to waive the requirement for this one and we will just test it at next release. :)

abn: Just let me know if you would want to try make a test out of interest.

Secrus:
The code this PR refers to no longer exists. Closing

Issue #2822, opened 2020-08-17
Add Docker Installation/Upgrading To Docs


9mido:
Stackoverflow and some other sites have methods for Poetry to be installed with Docker. But they are not all noob friendly and might not be the best ways to do things. 

Please add to the docs the recommended way to install poetry with docker and how to upgrade poetry with docker.

tetienne:
To understand clearly, you want to run poetry through a docker container, or build an image with Poetry inside?

9mido:
@tetienne Good point. I now see #1879 which has best practices and just realized this may be a duplicate of that issue. 

There are quite a few use cases to account for like you mentioned. That includes regular container run commands, Dockerfile and YML configurations. Accounting for the most common use cases in the Poetry documentation would be ideal.

max-pfeiffer:
Have a look at these two Docker projects, they provide base images for dockerizing your web application with Poetry:

- https://github.com/max-pfeiffer/uvicorn-poetry
- https://github.com/max-pfeiffer/uvicorn-gunicorn-poetry

Issue #1879, opened 2020-01-13
Document docker poetry best practices


mxab:
- [x ] I have searched the [issues](https://github.com/python-poetry/poetry/issues) of this repo and believe that this is not a duplicate.

## Issue

There are several discussions on how to build your final application for a docker image

- https://github.com/python-poetry/poetry/issues/1178
- https://stackoverflow.com/questions/53835198/integrating-python-poetry-with-docker

Would be great if the poetry team could provide a how to poetry with docker documentation

bufke:
I would add #1227 to this as well.

fralik:
I recently described usage of Poetry with [private repos](https://github.com/fralik/poetry-with-private-repos).

michael0liver:
I have been playing around with multi-stage builds and came up with this example:

```dockerfile
# `python-base` sets up all our shared environment variables
FROM python:3.8.1-slim as python-base

    # python
ENV PYTHONUNBUFFERED=1 \
    # prevents python creating .pyc files
    PYTHONDONTWRITEBYTECODE=1 \
    \
    # pip
    PIP_NO_CACHE_DIR=off \
    PIP_DISABLE_PIP_VERSION_CHECK=on \
    PIP_DEFAULT_TIMEOUT=100 \
    \
    # poetry
    # https://python-poetry.org/docs/configuration/#using-environment-variables
    POETRY_VERSION=1.0.3 \
    # make poetry install to this location
    POETRY_HOME="/opt/poetry" \
    # make poetry create the virtual environment in the project's root
    # it gets named `.venv`
    POETRY_VIRTUALENVS_IN_PROJECT=true \
    # do not ask any interactive question
    POETRY_NO_INTERACTION=1 \
    \
    # paths
    # this is where our requirements + virtual environment will live
    PYSETUP_PATH="/opt/pysetup" \
    VENV_PATH="/opt/pysetup/.venv"


# prepend poetry and venv to path
ENV PATH="$POETRY_HOME/bin:$VENV_PATH/bin:$PATH"


# `builder-base` stage is used to build deps + create our virtual environment
FROM python-base as builder-base
RUN apt-get update \
    && apt-get install --no-install-recommends -y \
        # deps for installing poetry
        curl \
        # deps for building python deps
        build-essential

# install poetry - respects $POETRY_VERSION & $POETRY_HOME
RUN curl -sSL https://raw.githubusercontent.com/sdispater/poetry/master/get-poetry.py | python

[...]

Issue #856, opened 2019-02-01
Dockfile example using poetry?


eisque:
I want to dockerize my app but I am getting issues related to paths and stuff.

pawamoy:
Did you find examples? If yes, could you share them? Others could find this useful.

I myself used Poetry in a Dockerfile like so:

```dockerfile
FROM python:3.6
RUN mkdir /app
WORKDIR /app

RUN pip install poetry
RUN poetry config settings.virtualenvs.create false
COPY poetry.lock pyproject.toml README.md /app/

# to prevent poetry from installing my actual app,
# and use docker caching ability:
RUN mkdir -p /app/src/app
RUN touch /app/src/app/__init__.py

RUN poetry install -n

COPY src /opt/srv/taskhub/src

EXPOSE 8000
CMD ["gunicorn", "-c", "docker/services/gunicorn/conf.py", "--bind", ":8000", "--chdir", "src/app", "core.wsgi:application"]
```

eisque:
```
FROM python:3.7.2
COPY . /app
WORKDIR /app
RUN curl -sSL https://raw.githubusercontent.com/sdispater/poetry/master/get-poetry.py | python
RUN /root/.poetry/bin/poetry install
CMD ["/root/.poetry/bin/poetry", "run", "python", "./start.py"]
````
my docker file

max-pfeiffer:
Have a look at these two projects, they provide base images for dockerizing your web application:

- https://github.com/max-pfeiffer/uvicorn-poetry
- https://github.com/max-pfeiffer/uvicorn-gunicorn-poetry

PR #3209, opened 2020-10-14
#2822 Add dockerfile install instructions

Saphyel:
# Pull Request Check List

Resolves: #2822

Ideally if the poetry team can provide an official **poetry.Dockerfile** that would be amazing, I'd be happy to help/maintining/looking after so people can only need to implement the **project.Dockerfile** file.

it's also possible to do it in one file combining the commands if you do not care too much about the best practices in docker.

I'm happy to push or accept any changes on my commit/s

### DIFF
--- a/docs/docs/index.md
+++ b/docs/docs/index.md
@@ -29,6 +29,24 @@ curl -sSL https://raw.githubusercontent.com/python-poetry/poetry/master/get-poet
 ```powershell
 (Invoke-WebRequest -Uri https://raw.githubusercontent.com/python-poetry/poetry/master/get-poetry.py -UseBasicParsing).Content | python -
 ```
+### dockerfile install instructions
+poetry.Dockerfile
+```Dockerfile
+FROM python:3.8-slim
+
+ENV POETRY_VERSION=1.1.3
+
+RUN apt-get update;apt-get install -y curl;rm -rf /var/lib/apt/lists/* \
+  ;curl -sSL https://raw.githubusercontent.com/python-poetry/poetry/master/get-poetry.py | python -
+```
+project.Dockerfile
+```Dockerfile
+FROM python:3.8-slim
+
+COPY --from=poetry:1 /root/.poetry /root/.poetry
+ENV PATH="/root/.poetry/bin:${PATH}"
+RUN poetry config virtualenvs.create false
+```
 
 !!! note
 


docs/docs/index.md
```diff
@@ -29,6 +29,24 @@ curl -sSL https://raw.githubusercontent.com/python-poetry/poetry/master/get-poet
 ```powershell
 (Invoke-WebRequest -Uri https://raw.githubusercontent.com/python-poetry/poetry/master/get-poetry.py -UseBasicParsing).Content | python -
 ```
+### dockerfile install instructions
+poetry.Dockerfile
```
abn, line 33: Why is all this required? Shouldn't this just be `pip install poetry`?

docs/docs/index.md
```diff
@@ -29,6 +29,24 @@ curl -sSL https://raw.githubusercontent.com/python-poetry/poetry/master/get-poet
...
+RUN apt-get update;apt-get install -y curl;rm -rf /var/lib/apt/lists/* \
+  ;curl -sSL https://raw.githubusercontent.com/python-poetry/poetry/master/get-poetry.py | python -
+```
+project.Dockerfile
+```Dockerfile
+FROM python:3.8-slim
+
+COPY --from=poetry:1 /root/.poetry /root/.poetry
+ENV PATH="/root/.poetry/bin:${PATH}"
+RUN poetry config virtualenvs.create false
```
abn, line 48: This is not something we recommend at all.

abn (changes requested):

abn:
One of the reasons why we have yet to officially provide a "poetry contianer" is that poetry can be installed in container environments using `pip install poetry` and this is what should be done. This is because the official installer is not required here as within a container multiple python versions are not expected. And the poetry install step is usually cached anyway. So maintaining an official container does not really make much sense since, in order to be useful you would have to provide versions for every python base containers. Additionally, even in a container environment, you are better off using a virtual environment unless it is explicitly not required - eg: using system packages etc.

What we might do later, is a provide a repository with container usage patterns for various use cases and build scenarios.

docs/docs/index.md
Saphyel: I hate virtualenvs

Saphyel:
The problem I see using `pip install poetry` is you will install some dependencies in your container that you may not need for your project or can be a conflict with your project.
In this way you are importing a folder.
I'm not sure why require to use virtual environment when that's the whole point of docker.

docs/docs/index.md
Saphyel: What if I change the whole thing to:
```suggestion
This is not officially supported yet. Use your container/image as a virtual machine and install it using any way mentioned here and we recommend that you still use virtualenvs.
```

br3ndonland:
https://github.com/python-poetry/poetry/pull/3209#issuecomment-708739500

> maintaining an official container does not really make much sense

@abn I agree. There's no need for the Poetry team to maintain a Docker container image. The official Python image is adequate.

> What we might do later, is a provide a repository with container usage patterns for various use cases and build scenarios.

This would be nice to have. I work with Poetry in Docker frequently, and I'm happy to help with this in the future. I would suggest closing this PR, and providing info on Docker use cases outside of the core docs.

https://github.com/python-poetry/poetry/pull/3209#issuecomment-708760335

> The problem I see using pip install poetry is you will install some dependencies in your container that you may not need for your project or can be a conflict with your project.

I've never seen dependency conflicts arise in this situation. I would guess it's quite rare, but I would welcome a reproducible example of Docker dependency conflicts.

> I'm not sure why require to use virtual environment when that's the whole point of having containers.

@Saphyel if you're looking for a Docker configuration that might suit your preferences, I would suggest the following:

```dockerfile
FROM python:3.8
ENV POETRY_HOME=/opt/poetry POETRY_VIRTUALENVS_CREATE=false PYTHONPATH=/app
COPY poetry.lock pyproject.toml /app/
WORKDIR /app/
RUN curl -fsS -o get-poetry.py https://raw.githubusercontent.com/python-poetry/poetry/master/get-poetry.py && \
  python get-poetry.py -y && . $POETRY_HOME/env && \
  poetry install --no-dev --no-interaction --no-root
COPY mypackage /app/mypackage
CMD python /app/mypackage/main.py

```

abn:
Regarding dependnecy conflicts; there are a few things to note here.

1. Usage patterns cannot assume that the base container is `docker.io/python:*` as a lot of use cases use distro base images and construct "supported" python instances. There is no guarentee that the python site is going to be clean. A particularly annoying example is that of the ubuntu base image (this might be resolved in newer base images) using custom `distutil` and `virtualenv` hacks.
2. While the python base images are typically up-to-date, and site is empty except for `setuptools`, `pip` and `wheel` (expected virutalenv packages), this might not always be guareteed.
3. At the moment, poetry is only essentially tested on the versions of `pip`, `setuptools` and `wheel` bundled with whatever version requirement poetry specifies for `virtualenv`. We have had issues with pip versions in system installed pythons. For example; `python -m ensurepip` will not install the latest version of `pip` but rather the version that is paired with that `python` version.

Putting all that together, recommending not using a virtual environment is not necesarily the best idea. And unfortunately, there has been bugs in the issue history if you dig that have been caused due to this. All that said, I am not necesarily saying it is "wrong", but that it just shouldn't be the recommended pattern.

9mido:
I am really confused here. 

https://python-poetry.org/docs/#installing-with-pip

The poetry docs state that "Be aware that it will also install Poetry's dependencies which might cause conflicts with other packages." is on par with what @Saphyel said. 

But the opposite is being recommended in this discussion. 

Then what is the recommended way @abn ? Can we see an example? 

Should we use the method @br3ndonland recommended or something else like in issues #856 or #1879 recommends?

Saphyel:
@9mido I think they are recommending using virtualenv inside of docker, I know is sort of defeating the whole purpose of using containers and having an even bigger image.

So if you have a cloud based project the recommendation is use pip, poetry is not ready yet.

br3ndonland:
> I am really confused here.

@9mido you're not alone in your confusion. There's a GitHub Discussion topic on this now (https://github.com/python-poetry/poetry/discussions/1879), so I posted a [comment](https://github.com/python-poetry/poetry/discussions/1879#discussioncomment-346113) there. Hope it helps.

robertlagrant:
> I think they are recommending using virtualenv inside of docker, I know is sort of defeating the whole purpose of using containers

I don't think this defeats the whole purpose of using containers at all. Containers provide sealed, standardised deployment units. Using virtual environments in them will still do that.

neersighted:
> 
> 
> > I think they are recommending using virtualenv inside of docker, I know is sort of defeating the whole purpose of using containers
> 
> I don't think this defeats the whole purpose of using containers at all. Containers provide sealed, standardised deployment units. Using virtual environments in them will still do that.

I have to agree with @robertlagrant here -- that is what is being said. Containers do not eliminate the desire to keep your app's dependencies separate from Poetry. While you could subvert Poetry by installing it into a virtualenv in your `Dockerfile` and then having Poetry install app dependencies into the global site-packages, inverting the standard paradigm like that serves no real functional purpose. You would merely be venturing into unexplored territory out of aesthetic reasons.

Anyway, this suggestion is the real takeaway here:

> What we might do later, is a provide a repository with container usage patterns for various use cases and build scenarios.

Any work to provide such a resource would be appreciated. It could be officially promoted/supported by the Poetry team, or remain a standalone resource with 'community' status in the docs.

I'm going to close this PR for now, but please do use it as a base for the above repo/resource.

Issue #3077, opened 2020-10-05
Poetry Never Updates pywin32


adam-grant-hendry:
- [x ] I am on the [latest](https://github.com/python-poetry/poetry/releases/latest) Poetry version.
- [x] I have searched the [issues](https://github.com/python-poetry/poetry/issues) of this repo and believe that this is not a duplicate.
- [x] If an exception occurs when executing a command, I executed it again in debug mode (`-vvv` option).

- **OS version and name**: Windows 10, version 1909
- **Poetry version**: 1.1.0
- **Link of a [Gist](https://gist.github.com/) with the contents of your pyproject.toml file**: 

## Issue

`poetry update` always attempts to update `pywin32`, but never does. Every new install lists `pywin32` as requiring update

```
>>> poetry update
Updating dependencies
Resolving dependencies...

Package operations: 0 installs, 1 update, 0 removals

  • Updating pywin32 (228 C:/Code/project/.venv/Lib/site-packages/pythonwin -> 228)
```

The `pyproject.toml` file is as follows:

```
[tool.poetry]
name = "project"
version = "0.1.0"
description = "test"
license = "Apache-2.0"

[tool.poetry.dependencies]
python = "^3.8"
numpy = "^1.19.2"
matplotlib = "^3.3.2"
pywin32 = "^228"

[tool.poetry.dev-dependencies]
black = "^20.8b1"
flake8 = "^3.8.4"
pytest = "^6.1.1"
pylint = "^2.6.0"
jupyter = "^1.0.0"
notebook = "^6.1.4"

[build-system]
requires = ["poetry-core>=1.0.0"]
build-backend = "poetry.core.masonry.api"
```

I am running Python 3.8.6.

adam-grant-hendry:
Reverting `poetry` to version `1.0.10` resolves this issue, but it is still uncertain to me why this occurs with version `1.1.0`.

abn:
Have you tried re-creating the virtual environment?

adam-grant-hendry:
Recreating the venv, I still have the error. Not working now with 1.0.10 either.

adam-grant-hendry:
Occurs in version 1.1.1 as well.

rajuah:
I see this as well starting with version `1.1.0`
Updating to `1.1.1` didnt resolve the issue

jgentil:
I have this same problem.  Following for solution...

jgentil:
[...]

PR #3210, opened 2020-10-14
Handle non-editable packages with pth files

abn:
When detecting installed packages, this change ensures that packages
with .pth files are not incorrectly marked as editable. A package is
considered editable only if at least one of the paths detected is
not in the environment site.

Resolves: #3077


absassi:
@abn, this works for me. Thanks.

abn:
@python-poetry/triage can I get this one reviewed please?

sdispater (approved):

PR #3214, opened 2020-10-15
[bug] script entry point import path

akeeman:
```toml
[tool.poetry.scripts]
baz = "bar:baz.boom.bim"
```

currently translates to

```python
from bar import baz.boom

if __name__ == '__main__':
    baz.boom.bim()
```

but should be

```python
from bar import baz  # <- notice the difference

if __name__ == '__main__':
    baz.boom.bim()
```

This is the case when, for instance, this is an entry point

```python
# file __init__.py in module bar
class baz:
    class boom:
        @classmethod
        def bim():
            #  ...
```

Also extended the tests to test sub-module import

### DIFF
--- a/poetry/masonry/builders/editable.py
+++ b/poetry/masonry/builders/editable.py
@@ -139,7 +139,7 @@ class EditableBuilder(Builder):
         for script in scripts:
             name, script = script.split(" = ")
             module, callable_ = script.split(":")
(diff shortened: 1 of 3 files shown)


akeeman:
I don't think those test failures are due to the changes I made...

akeeman:
@stephsamson @finswimmer

abn (approved):

Issue #3224, opened 2020-10-16
Extra dependencies for package on private repo uninstalled when running `poetry install`.


iksteen:
- [x] I am on the [latest](https://github.com/python-poetry/poetry/releases/latest) Poetry version.
- [x] I have searched the [issues](https://github.com/python-poetry/poetry/issues) of this repo and believe that this is not a duplicate.
- [x] If an exception occurs when executing a command, I executed it again in debug mode (`-vvv` option).

- **OS version and name**: Arch Linux
- **Poetry version**: 1.1.3
- **Link of a [Gist](https://gist.github.com/) with the contents of your pyproject.toml file**: https://gist.github.com/iksteen/bd2e570b8986f348e136038439d6be19

## Issue
After installing a package with extras from a private repository and running `poetry install` again, the extra dependencies are uninstalled. For easy reproducibility I've created a private repository that's publicly accessible and contains a single package (iksteen-aaa).

- Start out with a fairly empty pyproject.toml: https://gist.github.com/iksteen/83a3d5b55a8274bfbb434d72676fea70
- Now add the iksteen-aaa package with the example extra:
```
$ poetry add iksteen-aaa -E example
Using version ^0.1.0 for iksteen-aaa

Updating dependencies
Resolving dependencies... (1.6s)

Writing lock file

Package operations: 9 installs, 0 updates, 0 removals

  • Installing pycparser (2.20)
  • Installing cffi (1.14.3)
  • Installing pyasn1 (0.4.8)
  • Installing six (1.15.0)
  • Installing cryptography (3.1.1)
  • Installing ecdsa (0.14.1)
  • Installing rsa (4.6)
  • Installing python-jose (3.2.0)
  • Installing iksteen-aaa (0.1.0)
```
- Run `poetry install` again and see the extra dependencies getting uninstalled:
```
$ poetry install
Installing dependencies from lock file

Package operations: 0 installs, 0 updates, 8 removals

  • Removing cffi (1.14.3)
  • Removing cryptography (3.1.1)
  • Removing ecdsa (0.14.1)
  • Removing pyasn1 (0.4.8)
  • Removing pycparser (2.20)
  • Removing python-jose (3.2.0)
  • Removing rsa (4.6)
  • Removing six (1.15.0)
```

[...]

PR #3229, opened 2020-10-16
locker: handle nested extras requirement

abn:
Previously, when using locked repository, incorrect dependency instance
was created when a dependency's extra requirement activated a
nested extra. This change ensures that these are correctly
loaded.

As part of this change new lock files write PEP 508 serialised form of
extra dependencies in order to reuse core logic to parse specification
of extra requirement.

Resolves: #3224


sdispater (approved):
Looks good to me 👍

Issue #3219, opened 2020-10-15
Frequent intermittent connection failures


maroux:
- [x] I am on the [latest](https://github.com/python-poetry/poetry/releases/latest) Poetry version.
- [x] I have searched the [issues](https://github.com/python-poetry/poetry/issues) of this repo and believe that this is not a duplicate.
- [x] If an exception occurs when executing a command, I executed it again in debug mode (`-vvv` option).

- **OS version and name**: Debian GNU/Linux 10 (buster)
- **Poetry version**: 1.1.3
- **Link of a Gist with the contents of your pyproject.toml file**: https://gist.github.com/maroux/5bfc38375acd0e1c1a710fc0e62b8e93

## Issue
**Environment:** Docker debian slim buster image running in Google Cloud Platform Cloud Build
**Installation:** `poetry install -vvv --no-root --no-dev`
**Outcome:** Fails with networking error intermittently, but frequently enough that upgrade to poetry v1.1 is blocked on solving this issue.
**Notes:** Older version (v1.0.9) works fine but is slow because of old installer.
**Verbose logs:**
```
16:05:17 Step #2 - "build": Skipping virtualenv creation, as specified in config file.
16:05:17 Step #2 - "build": Installing dependencies from lock file
16:05:17 Step #2 - "build": 
16:05:17 Step #2 - "build": Finding the necessary packages for the current system
16:05:17 Step #2 - "build": 
16:05:17 Step #2 - "build": Package operations: 17 installs, 1 update, 0 removals, 11 skipped
16:05:17 Step #2 - "build": 
16:05:17 Step #2 - "build":   ? Installing certifi (2020.6.20): Skipped for the following reason: Already installed
16:05:17 Step #2 - "build":   ? Installing chardet (3.0.4): Skipped for the following reason: Already installed
16:05:17 Step #2 - "build":   ? Installing idna (2.10): Skipped for the following reason: Already installed
16:05:17 Step #2 - "build":   ? Installing markupsafe (1.1.1)
16:05:17 Step #2 - "build":   ? Installing pyparsing (2.4.7): Skipped for the following reason: Already installed
16:05:17 Step #2 - "build":   ? Installing pytz (2020.1): Skipped for the following reason: Already installed
16:05:17 Step #2 - "build":   ? Installing six (1.15.0): Skipped for the following reason: Already installed
16:05:17 Step #2 - "build":   ? Installing urllib3 (1.25.10): Skipped for the following reason: Already installed
[...]

PR #3230, opened 2020-10-17
Do not reuse requests sessions

abn:
When running under certain environment, reuse of request sessions maybe
causing network connectivity issues. This change ensures that request
sessions are not reused across requests.

Relates-to: #3219

### DIFF
--- a/poetry/installation/authenticator.py
+++ b/poetry/installation/authenticator.py
@@ -29,7 +29,6 @@ class Authenticator(object):
     def __init__(self, config, io=None):  # type: (Config, Optional[IO]) -> None
         self._config = config
         self._io = io
-        self._session = None
         self._credentials = {}
         self._password_manager = PasswordManager(self._config)
 
@@ -45,10 +44,7 @@ class Authenticator(object):
 
     @property
     def session(self):  # type: () -> requests.Session
-        if self._session is None:
-            self._session = requests.Session()
-
-        return self._session
+        return requests.Session()
 
     def request(
         self, method, url, **kwargs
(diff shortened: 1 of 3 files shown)


abn:
This seems to fix the issue; verified by the community on the issue.

sloev:
please get this released asap, we have had **broken Travis builds** for many days now because of this.

abn:
@sloev I can appreciate your urgency. But please consider that this is a community project with a very small team of volunteers. Would appreciate it if you refrained from demands, I am going to assume this was not the intention. 

As for your build issues. You can consider temporarily disabling the experimental installer as described [here](https://python-poetry.org/blog/announcing-poetry-1-1-0.html#brand-new-installer) untill `1.1.4` is released.

sloev:
@abn it wasn't my intention, sorry if it came across as such. (personal lesson learned: dont communicate on gh issues when under pressure) 
I will join the [discord](https://discordapp.com/invite/awxPgve) and see if i can help out instead of these "bursts of frustration"
Have a great day everybody and sorry for the noise 🤗

sdispater (approved):
Looks good to me 👍 

However, do you know whether the overhead of having to create new sessions for each download is negligible or not?

abn:
> However, do you know whether the overhead of having to create new sessions for each download is negligible or not?

There will be a performance hit. Last I checked, for around 50 packages in a clean environment it was around 2-3 seconds. That said, it will vary according to groups etc. But I think it is reasonable enough for now until we move to a more ... parallel safe http client. Right now since this is a major issue in container environments, I think we should go ahead with it.

Issue #3155, opened 2020-10-09
🧹 Test Suite Cleanup: Permanent Issue


abn:
This is a great issue to work on if you are new to the Poetry code base and want to start contributing. Over the years the poetry test suite has grown significantly, this means that there is significant amount of code duplication in-efficient 

If you are submitting a pull request associated with this, please keep the pull requests small and modular. This means, do not do too much in one PR and try to do only closely related changes in on PR. Additionally, make sure you add this to your PR description.

```
Relates-to: #3155
```

If this is your first contribution here, please refer to #3061 before continuing.

#### Convert reused code/logic into fixtures
There is a significant amount of code and data duplication in the test suite that can be converted to [pytest fixtures](https://docs.pytest.org/en/stable/fixture.html).

####  Adding coverage
There are certain code paths within the codebase that may not necessarily be covered within the test suite. You can.

a) Find them, and raise an issue to cover them.
b) (a) and also write some tests.

Known missing:
- [ ] `poetry remove` command.
- [ ] `poetry build` command.

#### Improve Parallel Safetly
Presently, because a lot of tests generate and cleanup package metadata, or share resources the tests suite is not parallel execution safe. Any improvements towards this goal is always welcome. Most of this will be covered by proper use of fixtures, but identifying shared file resources being used and working on on-write copies would go a long way in this goal.

### Low hanging targets
- [ ] Make [test_installer.py](https://github.com/python-poetry/poetry/blob/master/tests/installation/test_installer.py) test suite pass when run from any directory. Currently, tests expect that they be executed from the project root. There are file path comparisions that will fail since they use hard coded paths, eg: `tests/fixture/dist/some.whl`.
- [ ] Make [test_lock.py::test_lock_no_update](https://github.com/python-poetry/poetry/blob/b0b4a39f3e4558fb9975ec50662165ace6acf7d4/tests/console/commands/test_lock.py#L32) work on windows CI environemnts.

PR #3231, opened 2020-10-17
Make test suites  pass when run from any directory

cereblanco:
# Pull Request Check List

Relates-to: #3155 

Currently,  `test_installer.py` and `test_installer_old.py` expect that they be executed from the project root. 
Example running test on tests/installation directory produces:
Screen Shot 2020-10-17 at 2 46 01 PM

This PR patches current working directory `Path.cwd` to be always on project root. 

### DIFF
--- a/tests/installation/test_installer.py
+++ b/tests/installation/test_installer.py
@@ -76,15 +76,10 @@ class CustomInstalledRepository(InstalledRepository):
         return cls()
 
 
-@pytest.fixture
-def working_directory():
-    return Path(__file__).parent.parent.parent
-
-
 @pytest.fixture(autouse=True)
-def mock_path_cwd(mocker, working_directory):
+def mock_path_cwd(mocker, poetry_root_dir):
     yield mocker.patch(
-        "poetry.core.utils._compat.Path.cwd", return_value=working_directory
+        "poetry.core.utils._compat.Path.cwd", return_value=poetry_root_dir
     )
 
 
--- a/tests/installation/test_installer.py
+++ b/tests/installation/test_installer.py
@@ -76,6 +76,13 @@ class CustomInstalledRepository(InstalledRepository):
         return cls()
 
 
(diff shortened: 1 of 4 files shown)


tests/installation/test_installer.py
```diff
@@ -76,6 +76,18 @@ def load(cls, env):
         return cls()
 
 
+@pytest.fixture
+def working_directory():
```
abn, line 80: Can we move this into `tests/conftest.py` and rename it into something referring poetry source root?

cereblanco: @abn thanks for the feedback!  i moved it here: https://github.com/python-poetry/poetry/pull/3231/commits/87556f8d3f842b7872b1812fad4da19975df432a

I also used it for test_exporter https://github.com/python-poetry/poetry/pull/3231/commits/584a9e648d7ac6e1e077b04272e095947d377b7a

cereblanco: I also moved the `mock_path_cwd` to conftest. wdyt? thank you

abn: I suspect that should be fine; althought would like to double check if other test cases were mocking cwd. I believe there are test cases that update cwd to tbe a temporary project directory - just need to make sure that this fixture does not override those in the future. It might just be working right now due to the order the fixtures are loaded. Would be great if you can cross check that too.

cereblanco: @abn you're right, there are cases in other files that patch the cwd too. It will work with the current setup because it will be loaded first and then overwritten later. But I reverted it nevertheless, because it will be required by a just few and not all test suites anyway. thank you

abn: Thank you for checking up on that.

abn:
Restarted failing bsd run.

neersighted:
This should be fixed in master -- feel free to re-open/rebase if you can still observe the breakage.

Issue #2933, opened 2020-09-18
Ctrl-C during poetry add doesn't revert changes


Vozf:
If I interrupt poetry add with ctrl-C it still adds dependency to pyproject.yaml. This should be either reverted or be done in the end of installation

finswimmer:
Hello @Vozf,

this one is similar to yours: https://github.com/python-poetry/poetry/issues/381

fin swimmer

PR #3233, opened 2020-10-17
Add command catch KeyboardInterrupt and restore original_content (2933, 318)

estyxx:
# Pull Request Check List

Resolves: #2933 #318 

I wrapped KeyboardInterrupt where we are changing the content and write it... 
At line 103 I also write `original_content` instead of content, because if there isn't any package to add I will expect that nothing change... let me know if I'm wrong... 

FYI this issue is related to Hacktoberfest

### DIFF
--- a/poetry/console/commands/add.py
+++ b/poetry/console/commands/add.py
@@ -100,7 +100,6 @@ class AddCommand(InstallerCommand, InitCommand):
         packages = [name for name in packages if name not in existing_packages]
 
         if not packages:
-            self.poetry.file.write(original_content)
             self.line("Nothing to add.")
             return 0
 
--- a/poetry/console/commands/add.py
+++ b/poetry/console/commands/add.py
@@ -155,7 +155,6 @@ class AddCommand(InstallerCommand, InitCommand):
         try:
             # Write new content
             self.poetry.file.write(content)
-            raise KeyboardInterrupt("ops")
 
             # Cosmetic new line
             self.line("")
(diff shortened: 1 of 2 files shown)


poetry/console/commands/add.py
```diff
@@ -100,7 +100,7 @@ def handle(self):
         packages = [name for name in packages if name not in existing_packages]
 
         if not packages:
-            self.poetry.file.write(content)
+            self.poetry.file.write(original_content)
```
abn, line 103: Should we just remove this as we do not modify anything before this?

poetry/console/commands/add.py
```diff
@@ -152,29 +152,28 @@ def handle(self):
...
+            if self.option("lock"):
+                self._installer.lock()
 
-        self._installer.whitelist([r["name"] for r in requirements])
+            self._installer.whitelist([r["name"] for r in requirements])
 
-        try:
             status = self._installer.run()
-        except Exception:
+        except BaseException:
```
abn, line 175: Maybe add a comment here so we retain why we made this change.

abn (changes requested):
Couple of minor comments.

poetry/console/commands/add.py
estyxx: yes, think so, tests seem to work anyway...

poetry/console/commands/add.py
estyxx: added!

poetry/console/commands/add.py
```diff
@@ -173,6 +172,7 @@ def handle(self):
 
             status = self._installer.run()
         except BaseException:
+            # Catch all exception including KeyboardInterrupt in order to restore the original content
```
abn, line 175:
```suggestion
            # Using BaseException here as some exceptions, eg: KeyboardInterrupt, do not inherit from Exception
```

Sorry, should have been more specific about what I had in mind.

abn (approved):

abn:
@estyxx thank you for your contributions!

PR #3234, opened 2020-10-17
Fixes #3127: Do not log in install if quiet is set

### LINKED ISSUE #3127: Poetry install -q (and update -q) produce messages

- **OS version and name**: Xubuntu 20.04
- **Poetry version**: 1.1.2
- **Link of a [Gist](https://gist.github.com/) with the contents of your pyproject.toml file**: https://gist.github.com/berislavlopac/949972163f24f734ea84c27fbb27b2f4

## Issue
Running `poetry update -q` and `poetry install -q` produces output, although somewhat more limited than normally.

```
~/D/test ❯❯❯ (test) poetry install
Updating dependencies
Resolving dependencies... (0.3s)

Writing lock file

Package operations: 6 installs, 0 updates, 0 removals

  • Installing certifi (2020.6.20)
  • Installing chardet (3.0.4)
  • Installing idna (2.10)
  • Installing urllib3 (1.25.10)
  • Installing requests (2.24.0)
  • Installing starlette (0.13.8)
```

```
~/D/test ❯❯❯ (test) poetry install -q
  • Installing certifi (2020.6.20)
  • Installing chardet (3.0.4)
  • Installing idna (2.10)
  • Installing urllib3 (1.25.10)
  • Installing requests (2.24.0)
  • Installing starlette (0.13.8)
```

Using multiple `q`s (`poetry install -qq`) has the same result as none at all.

I was expecting no messages at all, as was the case in earlier versions.

nicoschmdt:
# Pull Request Check List

Resolves: #3127

I think the updated documentation part doesn't apply, if necessary I'd like some instructions where to update properly.

I would like to add I am currently involved in a hacktoberfest event.

### DIFF
--- a/tests/installation/test_installer.py
+++ b/tests/installation/test_installer.py
@@ -5,6 +5,7 @@ import sys
 
 import pytest
 
+from clikit.io import BufferedIO
 from clikit.io import NullIO
 
 from poetry.core.packages import ProjectPackage
@@ -1831,3 +1832,28 @@ def test_installer_can_handle_old_lock_files(
 
     # colorama will be added
     assert 8 == installer.executor.installations_count
+
+
+def test_quiet(package, pool, locker, env, installed, config, repo):
+    io = BufferedIO()
+    io.set_quiet(True)
+
+    installer = Installer(
+        io,
+        env,
+        package,
+        locker,
+        pool,
+        config,
+        installed=installed,
+        executor=Executor(env, pool, config, io),
+    )
+    installer.use_executor(True)
+
+    package_a = get_package("A", "1.0")
+    repo.add_package(package_a)
+    package.add_dependency(Factory.create_dependency("A", "~1.0"))
+
+    installer.run()
+
+    assert io.fetch_output() == ""
--- a/tests/installation/test_installer.py
+++ b/tests/installation/test_installer.py
@@ -1834,9 +1834,28 @@ def test_installer_can_handle_old_lock_files(
     assert 8 == installer.executor.installations_count
 
 
-def test_quiet(package, pool, locker, env, installed, config, repo):
+@pytest.mark.parametrize(
+    "quiet,lines",
+    [
+        (True, []),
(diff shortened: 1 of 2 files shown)


tests/installation/test_installer.py
```diff
@@ -1831,3 +1832,28 @@ def test_installer_can_handle_old_lock_files(
 
     # colorama will be added
     assert 8 == installer.executor.installations_count
+
+
+def test_quiet(package, pool, locker, env, installed, config, repo):
```
abn, line 1837:
```suggestion
@pytest.mark.parameterize("quiet,lines", [(True, []), (False, [".."])]
def test_quiet(package, pool, locker, env, installed, config, repo, quiet, lines):
```

Would be good to test both the cases here even thought this is indirectly verified elsewhere.

abn (changes requested):
Looks good, minor suggestion for the tests.

tests/installation/test_installer.py
nicoschmdt: I updated this PR with the suggested changes. In the `quiet=False` assertion, I tested with the whole command output but I'm not sure if this is the proper way. I don't know a better way to compare the output, do you have any suggestions?

tests/installation/test_installer.py
```diff
@@ -1831,3 +1832,47 @@ def test_installer_can_handle_old_lock_files(
...
+    )
+    installer.use_executor(True)
+
+    package_a = get_package("A", "1.0")
+    repo.add_package(package_a)
+    package.add_dependency(Factory.create_dependency("A", "~1.0"))
+
+    installer.run()
+
+    assert io.fetch_output().splitlines() == lines
```
abn, line 1878:
```suggestion
    output = io.fetch_output().strip()
    assert (not output and quiet) or (output and not queit)
```

Testing output can be painful. We could do this instead since we do not really have to validate the content for these tests as they are verified elsewhere.

Issue #3228, opened 2020-10-16
get-poetry.py installation with --file Does not rename lib


bphunter1972:
- [x] I am on the [latest](https://github.com/python-poetry/poetry/releases/latest) Poetry version.
- [x] I have searched the [issues](https://github.com/python-poetry/poetry/issues) of this repo and believe that this is not a duplicate.
- [x] If an exception occurs when executing a command, I executed it again in debug mode (`-vvv` option).

- **OS version and name**: CentOS7
- **Poetry version**: 1.1.3
- **Link of a [Gist](https://gist.github.com/) with the contents of your pyproject.toml file**: N/A

## Issue
I have a need to install poetry from a file because the server I'm running on is not currently allowing downloads outside of the local domain. This is fine since we have our own PyPI mirror.

Having downloaded the get-poetry.py script and the poetry-1.1.3.tar.gz from the mirror, I ran: 
./get-poetry.py --file poetry-1.1.3.tar.gz

This appears to work fine, but on closer inspection the .poetry/lib dir contains a directory named poetry-1.1.3. Simply renaming this to just poetry won't work because it really needs to poetry sub-directory to be extracted. I believe the problem is located in Installer::extract_lib.

DustinMoriarty:
I don't see get-poetry.py in the coverage. I would be happy to take a shot at creating some unit tests for the get-poetry script to see if we can reproduce this issue with tests. However, creating tests for get-poetry.py may be a larger task, so I could create a separate issue for it.

DustinMoriarty:
I was able to reproduce this with a unit test. I will look into the cause.

DustinMoriarty:
@bphunter1972 : Where are you getting the tar.gz file from? It looks like there may be a difference between an sdist off of pypi and the version that the installer downloads from https://github.com/python-poetry/poetry/releases.

DustinMoriarty:
@bphunter1972 : You should be able to install just fine from the github release location. However, see PR #3239 for support for using the sdist from your pypi mirror. I can see how in some organizations the pypi mirror could be preferred over github releases. Let me know if this works for you.

bphunter1972:
Got it. Confirmed that this works.
Yeah, adding more help on where the --file must come from would be great!

abn:
@bphunter1972 just to be clear you don't need the `get-poetry` script to install from PyPI or a mirror.

You can simply use `pip install` or `pipx install` (probably preferred) for those.

bphunter1972:
[...]

PR #3239, opened 2020-10-18
Add support for get-poetry.py --file using sdist

DustinMoriarty:
# Pull Request Check List

Resolves: #3228

* This pull request adds support for install from an sdist file for the get-poetry script. Currently, the file structure of sdist vs the github releases is different due to the poetry package being nested one level deeper in the sdist file structure. Strictly speaking, this is a feature. However, the previous behavior may have resulted in a user experience that was difficult to distinguish from a bug.
* Check the library location and raise an error if the expected poetry library is not created from the file provided by the user.
* Added some unit tests for the new functionality in the get-poetry.py file. The test fixtures used and the import structure can be used for more unit tests of get-poetry.py in the future.

### DIFF
--- a/tests/scripts/test_get_poetry.py
+++ b/tests/scripts/test_get_poetry.py
@@ -1,0 +1,69 @@
+import os
+import shutil
+
+
+try:
+    # Python 3.5+
+    import importlib.util
+except ImportError:
+    # Python 2.7
+    import imp
+
+from pathlib import Path
+
+import pytest
+
+from poetry.core.masonry.builder import Builder
+from poetry.factory import Factory
+
+
+# We need to set the poetry home environment variable prior to importing the script since it gets set once at import
+# time.
+SOURCE_HOME = Path(__file__).parent.parent.parent
+GET_POETRY_PATH = SOURCE_HOME / "get-poetry.py"
+TMP_DIR = Path(__file__).parent / "tmp"
+POETRY_HOME = TMP_DIR / ".poetry"
+os.environ["POETRY_HOME"] = str(POETRY_HOME)
+try:
+    # Python 3.5+
+    spec = importlib.util.spec_from_file_location("getpoetry", GET_POETRY_PATH)
+    getpoetry = importlib.util.module_from_spec(spec)
+    spec.loader.exec_module(getpoetry)
+except ImportError:
+    # Python 2.7
+
+    getpoetry = imp.load_source("getpoetry", str(GET_POETRY_PATH))
+
+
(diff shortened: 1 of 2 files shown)


tests/scripts/test_get_poetry.py
```diff
@@ -0,0 +1,72 @@
+import os
+import shutil
+
+
+try:
+    # Python 3.5+
```
abn, line 6: You can drop compatibility changes as we no longer test under py2 or py35.

tests/scripts/test_get_poetry.py
```diff
@@ -0,0 +1,72 @@
...
+
+import pytest
+
+from poetry.core.masonry.builder import Builder
+from poetry.factory import Factory
+
+
+# We need to set the poetry home environment variable prior to importing the script since it gets set once at import
+# time.
+SOURCE_HOME = Path(__file__).parent.parent.parent
```
abn, line 26: Use the fixture. And convert the following to fixtures as well.

tests/scripts/test_get_poetry.py
```diff
@@ -0,0 +1,72 @@
...
+GET_POETRY_PATH = SOURCE_HOME / "get-poetry.py"
+TMP_DIR = Path(__file__).parent / "tmp"
```
abn, line 28: Use the existing`tmp_dir` fixture

abn (commented):
I'm not sure why this is required because`sdists` are not supported in this install mechanism. And really, shouldn't be supported via `get-poetry.py`.

DustinMoriarty:
I was wondering if that would be the case. Perhaps improved docs and error handling is the way to go. Thoughts?

DustinMoriarty:
Also, just curious what the story is with all the OS specific builds. From what I can tell they all just have the poetry source in them. However, I am guessing there is an important use case that I am missing.

abn:
The github releases contain vendored packages under `poetry/_vendored`. These could be unfortunately `(interpreter, os)` specific.

Improving the help doc for `--file`to indicate that only github releases are supported. As for error handling I guess you could validate the archive in-memory before extracting.

tests/scripts/test_get_poetry.py
DustinMoriarty: Well that simplifies things!

DustinMoriarty:
I am going to close this pull request because sdist files will not be supported. See PR #3259 for an alternative resolution to issue #3228 by improving documentation and source file validation.

PR #3241, opened 2020-10-18
Use global fixture_dir instead of defining one on each test file

cereblanco:
# Pull Request Check List

Relates-to: https://github.com/python-poetry/poetry/issues/3155

There exists `fixture_dir` in conftest visible to all test suites. So we might as well use it in our tests.
There are actually more other files that we can update, I will continue updating them in other PRs.

Risk: maybe i'm oversimplifying? 😅 

### DIFF
--- a/tests/inspection/test_info.py
+++ b/tests/inspection/test_info.py
@@ -12,23 +12,19 @@ from poetry.utils.env import EnvCommandError
 from poetry.utils.env import VirtualEnv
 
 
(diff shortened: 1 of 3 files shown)


neersighted (changes requested):
This looks like a good change overall! While yes, maybe it is a little naive, I think it's good incremental progress than can always be fine-tuned if the need presents itself later.

`master` has moved a lot in the last year -- can you rebase/apply your changes to the current state of the repo (including new/changed tests)?

Issue #3049, opened 2020-10-02
Relative `cache-dir` breaks the new installer


matejcik:
- [x] I am on the [latest](https://github.com/python-poetry/poetry/releases/latest) Poetry version.
- [x] I have searched the [issues](https://github.com/python-poetry/poetry/issues) of this repo and believe that this is not a duplicate.
- [x] If an exception occurs when executing a command, I executed it again in debug mode (`-vvv` option).

- **OS version and name**: Linux Mint 20, Python 3.8.2
- **Poetry version**: 1.1.0, master
- **Link of a [Gist](https://gist.github.com/) with the contents of your pyproject.toml file**: https://gist.github.com/matejcik/f7b0d6ca9bea9e6613f7a99433d7ad65

## Issue
When `cache-dir` is set to a relative path (in poetry.toml), bad things happen.

To reproduce:

1. place contents of above gist into a directory
2. `poetry install`
3. `rm -r .venv`
4. `poetry install` again

**Expected result:** this should work fine

**Actual result:** a spectacularly unhelpful stack trace with the following error: `ValueError: relative path can't be expressed as a file URI`

after modifying something or other, the actual stack trace shows up:
```
ValueError: relative path can't be expressed as a file URI
Traceback (most recent call last):
  File "/home/matejcik/.local/lib/python3.8/site-packages/poetry/installation/executor.py", line 180, in _execute_operation
    result = self._do_execute_operation(operation)
  File "/home/matejcik/.local/lib/python3.8/site-packages/poetry/installation/executor.py", line 255, in _do_execute_operation
    result = getattr(self, "_execute_{}".format(method))(operation)
  File "/home/matejcik/.local/lib/python3.8/site-packages/poetry/installation/executor.py", line 390, in _execute_install
    return self._install(operation)
  File "/home/matejcik/.local/lib/python3.8/site-packages/poetry/installation/executor.py", line 416, in _install
    archive = self._download(operation)
  File "/home/matejcik/.local/lib/python3.8/site-packages/poetry/installation/executor.py", line 559, in _download
    return self._download_link(operation, link)
  File "/home/matejcik/.local/lib/python3.8/site-packages/poetry/installation/executor.py", line 564, in _download_link
    archive = self._chef.get_cached_archive_for_link(link)
  File "/home/matejcik/.local/lib/python3.8/site-packages/poetry/installation/chef.py", line 49, in get_cached_archive_for_link
    archives = self.get_cached_archives_for_link(link)
[...]

PR #3245, opened 2020-10-19
Fix relative paths for cache-dir

ericriff:
Resolves: #3049

This PR addresses an issue with having relative paths on the cache-dir, AFAIK it only happens on python3.7 and up. I'm not sure of the internals of Poetry so the maintainers should take a look at this to make sure it doesn't introduce undesired behaviors.

All the credit goes to @matejcik since he came up with the fix. I'm just opening the PR since I really need this fixed and the issue where he reported this has been quiet for about 2 weeks.

### DIFF
--- a/poetry/installation/chef.py
+++ b/poetry/installation/chef.py
@@ -81,7 +81,7 @@ class Chef:
         links = []
         for archive_type in archive_types:
             for archive in cache_dir.glob("*.{}".format(archive_type)):
-                links.append(Link(archive.as_uri()))
+                links.append(Link(archive.resolve().as_uri()))
 
         return links
 


ericriff:
Can someone please take a look at this?

dylrich:
We would very much appreciate someone reviewing this - this bug has become a major problem for us!

ericriff:
Since it was a problem for us as well we ended up using a custom version with this patch in it. I would love to have this merged upstream so we don't have to maintain our own fork just for this.

ericriff:
Closing this since I've found some undesired behavior/bug.
In my `poetry.toml` I set `cache-dir = ".poetry"` and what I expected is poetry to create the cache folder in a `.poetry` folder located next to the `pyproject.toml` file, in the root of my project. 
The actual behavior is the `.poetry` folder being created in the current working directory at the moment of invoking a poetry command. This means that you end up with a `.poetry` cache folder being created in each folder you invoke a poetry command.
I didn't catch this before since I interect with my project from its root.

I don't know poetry well enough to fix this, so I'm closing the PR.
I would love if any of the main devs get involdes in this, this is some very usefull functionallity if you ask me.

Rjdrenth:
Thanks for the update. I was using this fix to resolve some issues as well. If you end up finding another solution, I'd appreciate greatly appreciate it if you could write it here.

ericriff:
I didn't come up with this fix. It was proposed here https://github.com/python-poetry/poetry/issues/3049 but since no one was opening a PR for it I did it.
Anyways, if I find a better approach I'll post it here.

jamesanto:
I am facing the same issue, is it possible to reopen the PR?

Issue #3220, opened 2020-10-16
Add ability to togglle dependency 'develop'


Seanny123:
- [x] I have searched the [issues](https://github.com/python-poetry/poetry/issues) of this repo and believe that this is not a duplicate.
- [x] I have searched the [documentation](https://python-poetry.org/docs/) and believe that my question is not covered.

## Feature Request

Path-dependencies in Poetry can be [optionally marked as `'develop' = True`](https://python-poetry.org/docs/dependency-specification/#path-dependencies). However, there's no way to mark it as such via the `poetry add` command.

It would be nice to set the `'develop' = True` when adding a path. For example:
```
poetry add ./my-package/ --develop
```
Or
```
poetry add ./my-package/ --editable
```

I think I can implement this myself if the core devs consider this feature to be beneficial.

finswimmer:
Hello  @Seanny123,

yes this is something really useful :+1: 

@abn listed this also in his suggestion in https://github.com/python-poetry/poetry/issues/3004.

fin swimmer

abn:
I think this can happen independent of #3004. If this happens first, we just update the key it operates on when #3004 lands.

0scarB:
@python-poetry/triage Hi, really like the project. Have used it for a couple of personal projects. Mind if I take on the issue

finswimmer:
Hello @0scarB,

I don't know if @Seanny123 already started to work on it? Otherwise we really appreciate any contribution :+1: 

fin swimmer

Seanny123:
Have no started yet. Go for it @0scarB

0scarB:
@Seanny123 thx, I will :)

0scarB:
Only [directory](https://github.com/python-poetry/poetry-core/blob/fa522603bc2aff99127adaf585046f0ad50db973/poetry/core/packages/directory_dependency.py#L11-L21) and [vcs dependencies](https://github.com/python-poetry/poetry-core/blob/fa522603bc2aff99127adaf585046f0ad50db973/poetry/core/packages/vcs_dependency.py#L10-L28)support the `develop` attribute. 

If we add `develop = true` without checking the `source_type` first then the follow unintuitive error message is raised for dependency that do not support it
```
RuntimeError

  The Poetry configuration is invalid:
    - [dependencies.<dependency>] {'version': '<version>', 'develop': True} is not valid under any of the given schemas
```
It would be better to raise an explicit error such as `The develop flag is not supported for dependencies of type <source_type>`
[...]

PR #3250, opened 2020-10-19
Issue 3220 add develop flag for add command

0scarB:
# Pull Request Check List

Resolves: #3220

Simply adding the `develop` attribute causes the following [issue](https://github.com/python-poetry/poetry/issues/3220#issuecomment-712486145) however - only [directory](https://github.com/python-poetry/poetry-core/blob/fa522603bc2aff99127adaf585046f0ad50db973/poetry/core/packages/directory_dependency.py#L11-L21) and [vcs dependencies](https://github.com/python-poetry/poetry-core/blob/fa522603bc2aff99127adaf585046f0ad50db973/poetry/core/packages/vcs_dependency.py#L10-L28) support the `develop` attribute. 

If we add `develop = true` without checking the `source_type` first then the follow unintuitive error message is raised for dependencies that do not support it
```
RuntimeError

    - [dependencies.<dependency>] {'version': '<version>', 'develop': True} is not valid under any of the given schemas
```
It would be better to raise an explicit error such as `The develop flag is not supported for dependencies of type <source_type>`
Unfortunately I don't see an elegant way of doing this without changing how `poetry-core` dependencies handle the `develop` parameter.

### DIFF
--- a/poetry/console/commands/add.py
+++ b/poetry/console/commands/add.py
@@ -50,7 +50,11 @@ class AddCommand(InstallerCommand, InitCommand):
             "Output the operations but do not execute anything (implicitly enables --verbose).",
         ),
         option("lock", None, "Do not perform operations (only update the lockfile)."),
-        option("editable", None, "Add as dependency in editable mode."),
+        option(
+            "editable",
+            None,
+            "Add as dependency in editable mode. Will cause command to fail for non path or VCS dependencies!",
+        ),
     ]
     help = (
         "The add command adds required packages to your <comment>pyproject.toml</> and installs them.\n\n"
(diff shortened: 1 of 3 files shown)


poetry/console/commands/add.py
```diff
@@ -50,6 +50,7 @@ class AddCommand(InstallerCommand, InitCommand):
             "Output the operations but do not execute anything (implicitly enables --verbose).",
         ),
         option("lock", None, "Do not perform operations (only update the lockfile)."),
+        option("develop", None, "Add as dependency in editable mode."),
```
abn, line 53: Please use `--editable`

abn: The documentation should also mention that this is only supported for `VCS` and `path` dependnecies.

0scarB: @abn My bad, thought documentation for cli options were automatically generated

finswimmer:
Thanks a lot for your contribution @0scarB.

In the meantime this PR is superseded by https://github.com/python-poetry/poetry/pull/3940.

fin swimmer

Issue #3238, opened 2020-10-18
Custom repository doesn't have precedence over PyPI


jaklan:
- [x] I am on the [latest](https://github.com/python-poetry/poetry/releases/latest) Poetry version.
- [x] I have searched the [issues](https://github.com/python-poetry/poetry/issues) of this repo and believe that this is not a duplicate.
- [x] If an exception occurs when executing a command, I executed it again in debug mode (`-vvv` option).

- **OS version and name**: macOS Mojave 10.14.6
- **Poetry version**: 1.1.3
- **Link of a [Gist](https://gist.github.com/) with the contents of your pyproject.toml file**:
```toml
[[tool.poetry.source]]
name = 'foobar'
url = 'http://repository.intranet.foobar.com/artifactory/api/pypi/pypi-virtual/simple'

[build-system]
requires = ["poetry_core>=1.0.0"]
build-backend = "poetry.core.masonry.api"
```

## Issue
In docs we can read:
> Any custom repository will have precedence over PyPI.

In the above example, `foobar` is a proxy repo incl. both internal packages and the public ones from PyPI. Running `poetry install` should cause a creation of `poetry.lock`, where all the packages have `foobar` in their `source` field. However, it's completely opposite - none of the packages has a `source` field, which means all of them are downloaded directly from PyPI.

When I add `default = true` to the above `source` element, which means PyPI is disabled, everything works properly - generated `poetry.lock` includes `foobar` as a `source` for all packages and installation doesn't break, so it's not a problem with repository itself.

It's especially frustrating when you want to use `poetry export` (for any reason), because then the final `requirements.txt` doesn't include neither `--index-url` nor `--extra-index-url` (it seems it depends on the actually used sources, not the `pyproject.toml` content).

jaklan:
I've just found the very similar issues:
https://github.com/python-poetry/poetry/issues/1677
https://github.com/python-poetry/poetry/issues/2564
but the problem seems not to be resolved for almost a year, so I will keep this issue open to bump the topic.

abn:
Can you please try the fix at #3251.

### Using pipx
```sh
pipx install --suffix=@3251 'poetry @ git+https://github.com/python-poetry/poetry.git@refs/pull/3251/head'
```

### Using a container (podman | docker)
```sh
podman run --rm -i --entrypoint bash python:3.8 <<EOF
set -xe
python -m pip install -q git+https://github.com/python-poetry/poetry.git@refs/pull/3251/head
install -d foobar
[...]

PR #3251, opened 2020-10-19
pool: ensure sources are prioritised over PyPI

abn:
When a project specifies non default sources, PyPI gets added as the
default source. This will prioritise packages available in PyPI when
the package exists in both index. This change ensures that PyPI is
only used as a default when no other sources are provided.

Resolves: #1677 #2564 #3238

### DIFF
--- a/poetry/factory.py
+++ b/poetry/factory.py
@@ -70,7 +70,8 @@ class Factory(BaseFactory):
         )
 
         # Configuring sources
-        for source in poetry.local_config.get("source", []):
+        sources = poetry.local_config.get("source", [])
+        for source in sources:
             repository = self.create_legacy_repository(source, config)
             is_default = source.get("default", False)
             is_secondary = source.get("secondary", False)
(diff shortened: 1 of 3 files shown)


sdispater (approved):

intgr:
Note that this caused a regression in our use case, we had configured a custom source in `[[tool.poetry.source]]`, but the certificate was not trusted. Since 1.1.3 did not use the configured source, it worked, but 1.1.4 started failing by surprise.

It's not a big deal, but on the other hand, we would prefer not to have ugly surprises in patch level updates.

Maybe such behavior-changing should not go into patch-level releases?

sinoroc:
@intgr 
@tomzx contributed a possible fix here: https://github.com/python-poetry/poetry/pull/3406
Would you mind testing it if you get a chance?

abn:
> Note that this caused a regression in our use case, we had configured a custom source in `[[tool.poetry.source]]`, but the certificate was not trusted. Since 1.1.3 did not use the configured source, it worked, but 1.1.4 started failing by surprise.
> 
> It's not a big deal, but on the other hand, we would prefer not to have ugly surprises in patch level updates.
> 
> Maybe such behavior-changing fixes should not go into patch-level releases?

@intgr One thing to note here is that this was a bug fix to what became expected behaviour with 1.1.0 - ie. how various repository sources configured were handled. This was expected to keep priorities as described in the documentation, however this was not the case. 

That said, I can also appreciate that how this is viewed can be subjective, Definitely, not intentional to cause surprises in patch releases and not something we want to make a habbit of either.

Issue #3199, opened 2020-10-13
poetry install fails on KeyError even for defined environment variables


pepastach:
- [x] I am on the [latest](https://github.com/python-poetry/poetry/releases/latest) Poetry version.
- [x] I have searched the [issues](https://github.com/python-poetry/poetry/issues) of this repo and believe that this is not a duplicate.
- [x] If an exception occurs when executing a command, I executed it again in debug mode (`-vvv` option).

- **OS version and name**: Debian Linux 10 Buster
- **Poetry version**: 1.1.2
- **Link of a [Gist](https://gist.github.com/) with the contents of your pyproject.toml file**: https://gist.github.com/pepastach/85b133f035fbb805cacd4d0fd75562b1

## Issue

We run `poetry install --remove-untracked` and the process randomly fails on various "missing" environment variables. This has been reported in #3144 for the `PATH` variable (which happens quite a lot, more than other variables), but I have seen it for many other variables (often GitLab or Docker related). In all cases they are defined.

It fails really randomly. Different dependencies being installed, different environment variables. Sometimes it does not fail (but most often it does). 
Sometimes it fails on multiple environment variables. Like `KeyError 'PATH'`, `KeyError 'CI_JOB_STAGE'` or `KeyError 'DOCKER_ENV_CI_DISPOSABLE_ENVIRONMENT'` in the same run. It seems to me like some kind of race condition in the parallel install. I can't prove this hypothesis until we can disable parallel installation. I think it's scheduled for an upcoming release.

Here is an example output where it fails on the `PATH` variable.
```
Creating virtualenv bp-xxx-3RqKXHa2-py3.7 in /builds/xxx/.cache/poetry/pypoetry/virtualenvs
Installing dependencies from lock file
Package operations: 106 installs, 0 updates, 0 removals
  • Installing attrs (20.2.0)
  • Installing six (1.15.0)
  • Installing cattrs (1.0.0)
...
Stack trace:
  8  /usr/local/lib/python3.7/site-packages/poetry/installation/executor.py:199 in _execute_operation
      197│ 
      198│             try:
    → 199│                 result = self._do_execute_operation(operation)
      200│             except EnvCommandError as e:
      201│                 if e.e.returncode == -2:
  7  /usr/local/lib/python3.7/site-packages/poetry/installation/executor.py:273 in _do_execute_operation
      271│             return 0
      272│ 
    → 273│         result = getattr(self, "_execute_{}".format(method))(operation)
      274│ 
[...]

PR #3253, opened 2020-10-20
env: do not modify os.environ

abn:
Replace updates of os.environ with explcit passing of `env` to
subprocess calls.

Relates-to: #3199

### DIFF
--- a/poetry/utils/env.py
+++ b/poetry/utils/env.py
@@ -10,6 +10,7 @@ import sysconfig
 import textwrap
 
 from contextlib import contextmanager
+from copy import deepcopy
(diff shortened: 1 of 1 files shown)


samvasko:
Could you please describe how is this fix related to the bug you've mentioned? Is the assumption that the failures occur because the subprocess does not have access to the parent env variables?

abn:
It does not resolve the linked issue but only relates to it.

The assumption here is that part of the issue (not whole of it) is due to how env vars are being passed in the thread pool.

stephsamson (approved):

Issue #1645, opened 2019-11-28
Some tests fail because of wrong rights on /foo


mcepl:
- [x] I am on the [latest](https://github.com/sdispater/poetry/releases/latest) Poetry version.
- [x] I have searched the [issues](https://github.com/sdispater/poetry/issues) of this repo and believe that this is not a duplicate.
- [x] If an exception occurs when executing a command, I executed it again in debug mode (`-vvv` option).

- **OS version and name**: openSUSE/Tumbleweed (Linux)
- **Poetry version**: 1.0.0b8

## Issue
Plenty of tests fail because of the traceback similar to this one (from ):
```
[   88s] ____________________________ test_add_no_constraint ____________________________
[   88s] 
[   88s] app = <tests.console.conftest.Application object at 0x7fd532dff4a8>
[   88s] repo = <tests.console.conftest.Repository object at 0x7fd532b587b8>
[   88s] installer = <poetry.installation.noop_installer.NoopInstaller object at 0x7fd532bc1470>
[   88s] 
[   88s]     def test_add_no_constraint(app, repo, installer):
[   88s]         command = app.find("add")
[   88s]         tester = CommandTester(command)
[   88s]     
[   88s]         repo.add_package(get_package("cachy", "0.1.0"))
[   88s]         repo.add_package(get_package("cachy", "0.2.0"))
[   88s]     
[   88s] >       tester.execute("cachy")
[   88s] 
[   88s] tests/console/commands/test_add.py:19: 
[   88s] _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
[   88s] /usr/lib/python3.7/site-packages/cleo/testers/command_tester.py:61: in execute
[   88s]     self._status_code = command.run(args, self._io)
[   88s] /usr/lib/python3.7/site-packages/clikit/api/command/command.py:116: in run
[   88s]     return self.handle(self.parse(args), io)
[   88s] /usr/lib/python3.7/site-packages/clikit/api/command/command.py:120: in handle
[   88s]     status_code = self._do_handle(args, io)
[   88s] /usr/lib/python3.7/site-packages/clikit/api/command/command.py:163: in _do_handle
[   88s]     self._dispatcher.dispatch(PRE_HANDLE, event)
[...]

PR #3255, opened 2020-10-20
tests: cleanup cache and http usage

abn:
- ensure tests rely on temporary cache directory
- remove external http call requirement for lock --no-update

Relates-to: #1645

### DIFF
--- a/tests/config/test_config.py
+++ b/tests/config/test_config.py
@@ -5,8 +5,8 @@ def test_config_get_default_value(config):
     assert config.get("virtualenvs.create") is True
 
 
-def test_config_get_processes_depended_on_values(config):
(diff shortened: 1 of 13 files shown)


bnavigator:
The path fixes are not necessary, if you prescribe the running pytest from a virtualenv python interpreter (#1645). Could move the  the `lock --no-update` fix to a separate patch for 1.1.x?

abn:
@bnavigator as its not a functional change I do not think this needs to be ported to 1.1. 

Here is a standalone patch for you: https://gist.githubusercontent.com/abn/13a38b1c4f61002555eb5d39e494627c/raw/6165ef1610ef907a36aea272207c8de1faf2499a/poetry-1645-1.1.patch

stephsamson (approved):

Issue #2252, opened 2020-04-01
Standard project structure not specified or not easily findable


sandrinr:
- [x] I have searched the [issues](https://github.com/python-poetry/poetry/issues) of this repo and believe that this is not a duplicate.

## Issue

When Poetry is unable to determine the packages of the current project (folder). It prints an error such as:
```
[ModuleOrPackageNotFound]
No file/folder found for package <package>
```

When a user then tries to find something about this error on the web he will find some closed issues talking about the `packages` portion of the `pyproject.toml`. There and in some other locations (e.g. GitHub) you can find "If your project structure differs from the standard one supported by poetry...". However, it is not defined (or will not pop up in search results) defined, as far as I can see, what exactly this standard structure might be.

By looking at https://python-poetry.org/docs/basic-usage/ one might guess what it could be.

GabrielSimonetto:
Could you tell us what an example of this error would be? By looking on the website it seems to me it's somehow defined for most cases

```
poetry-demo
├── pyproject.toml
├── README.rst
├── poetry_demo
│   └── __init__.py
└── tests
    ├── __init__.py
    └── test_poetry_demo.py
```

By searching [issues with this error](https://github.com/python-poetry/poetry/issues?q=is%3Aissue+ModuleOrPackageNotFound), it seems to me there is no single answer as to how to define this structure, since most issues have been solved by some ad hoc knowledge

GabrielSimonetto:
@python-poetry/triage

abn:
The standard structure being referred to is the structure where the project's root package is in the same directory as the `pyproject.toml` file (project root).

By default, if [`tool.poetry.packages`](https://python-poetry.org/docs/pyproject/#packages) is not defined, poetry will attempt to look for [`module_name(tool.poetry.name)`](https://python-poetry.org/docs/pyproject/#name). If this is not found the error is raised. 

This typically happens when `poetry init` was used to create the `pyproject.toml` (or it was created manually) and the package name does not match expected derived package name from project name; or the existing project uses `src` layout without specifying `packages`.

finswimmer:
I guess @GabrielSimonetto's main point is, that it is not stated clearly in the docs  how poetry detects the package and this should be improved. PR is always welcome :)

HallerPatrick:
Made a pull request, which should provide sufficient information in the docs

woolinsilver-kainos:
This stuff is non-obvious and very confusing IMHO.
[...]

PR #3257, opened 2020-10-20
Updates in Basic Usage Documentation for further specification on project and module names 

HallerPatrick:
Following documentation changes should suffice to at least know that `tool.poetry.name` can be resolved as
module name and for other cases the `/doc/pyproject#packages` page is referenced.

Resolves: #2252


docs/docs/basic-usage.md
```diff
@@ -41,6 +41,11 @@ python = "*"
 pytest = "^3.4"
 ```
 
+Make sure that `tool.peotry.name` matches the module name or otherwise specified within
```
stephsamson, line 44:
```suggestion
Make sure that `tool.poetry.name` matches the module name or otherwise specified within
```

docs/docs/basic-usage.md
```diff
@@ -41,6 +41,11 @@ python = "*"
...
+`tool.peotry.packages`.
```
stephsamson, line 45:
```suggestion
`tool.poetry.packages`.
```

stephsamson (changes requested):
Just a small typo :)

TBBle:
I just bounced off exactly this problem, and eventually found this pending PR (via #2252) documenting exactly what I was doing wrong (`tool.poetry.name` wasn't the package name, and there was no prompt that this resulted in no packages being installed, just the listed dependencies).

That said, this text is maybe a little unclear, as it suggests that 
> `tool.poetry.name` [is] specified within `tool.poetry.packages`

which is not the problem.

So perhaps
> Poetry assumes your package contains a packge with the same name as `tool.poetry.name`. If this is not the case, populate `tool.poetry.packages` to specify your package or packages.

Per [Modules](https://docs.python.org/3/tutorial/modules.html), what we're talking about here are Packages (directories with _\_\_init\_\_.py_, "a collection of modules"), not Modules ("a file containing Python definitions and statements").

PierreGuilmin:
I really think this would be a great adding to the documentation, as I just lost quite some time finding this information. I think @TBBle phrasing is the best, I can quickly write the new PR if this one keeps staling.

neersighted (changes requested):
> 
> 
> I just bounced off exactly this problem, and eventually found this pending PR (via #2252) documenting exactly what I was doing wrong (`tool.poetry.name` wasn't the package name, and there was no prompt that this resulted in no packages being installed, just the listed dependencies).
> 
> That said, this text is maybe a little unclear, as it suggests that
> 
> > `tool.poetry.name` [is] specified within `tool.poetry.packages`
> 
> which is not the problem.
> 
> So perhaps
> 
> > Poetry assumes your package contains a packge with the same name as `tool.poetry.name`. If this is not the case, populate `tool.poetry.packages` to specify your package or packages.
> 
> Per [Modules](https://docs.python.org/3/tutorial/modules.html), what we're talking about here are Packages (directories with ___init__.py_, "a collection of modules"), not Modules ("a file containing Python definitions and statements").

I have to agree with the above -- @HallerPatrick, would you be willing to rephrase a bit along those lines?

HallerPatrick:
@neersighted updated the docs after the wording of @TBBle. Could you also maybe squash my upstream merge commits of my fork?

TBBle (approved):

neersighted (approved):

Issue #3263, opened 2020-10-21
 Poetry attempts to install all VCS deps as editable, even when `develop = false`.


triptec:
- [x] I am on the [latest](https://github.com/python-poetry/poetry/releases/latest) Poetry version.
- [x] I have searched the [issues](https://github.com/python-poetry/poetry/issues) of this repo and believe that this is not a duplicate.
- [x] If an exception occurs when executing a command, I executed it again in debug mode (`-vvv` option).

- **OS version and name**: Ubuntu 20.04
- **Poetry version**: 1.1.3

## Issue

Hello, I'm having a hard time adding a git dependency and even though i set develop = false, the command to pip install has a "-e" and then fails with```
ERROR: File "setup.py" not found. Directory cannot be installed in editable mode: /usr/local/src/fastapi
  (A "pyproject.toml" file was found, but editable mode currently requires a setup.py based build.)```

abn:
Can you please try the fix at #3264.

### Using pipx
```sh
pipx install --suffix=@3264 'poetry @ git+https://github.com/python-poetry/poetry.git@refs/pull/3264/head'
```

### Using a container (podman | docker)
```sh
podman run --rm -i --entrypoint bash python:3.8 <<EOF
set -xe
python -m pip install -q git+https://github.com/python-poetry/poetry.git@refs/pull/3264/head
install -d foobar
pushd foobar
cat > pyproject.toml <<TOML
[tool.poetry]
name = "foobar"
version = "0.1.0"
description = ""
authors = ["Your Name <your.name@example.com>"]

[tool.poetry.dependencies]
python = "^3.8"
fastapi = {git = "https://github.com/tiangolo/fastapi.git", develop = false}

[tool.poetry.dev-dependencies]

[build-system]
requires = ["poetry-core>=1.0.0"]
build-backend = "poetry.core.masonry.api"
TOML
poetry install
EOF
```

triptec:
I’ll give it a try later today

PR #3264, opened 2020-10-21
Ensure vcs dependencies preserve editable flag

abn:
Resolves: #3263


finswimmer (approved):
It somehow looks familiar to me. :thinking: Is it a regression?

abn:
> It somehow looks familiar to me.  Is it a regression?

I think we had a similar issue a while back. I have added a test case to ensure we do not regress.

PR #3274, opened 2020-10-23
Update basic-usage.md

jli:
fix whitespace for one-liner venv activate command


neersighted:
Looks like this is solved on current mater. Thanks for the PR!

PR #3283, opened 2020-10-25
env: pass full path of python executable to virtualenv creation

### LINKED ISSUE #3284: Poetry creates virtualenv with wrong version of Python despite finding the right one

- **OS version and name**: macOS 10.15.7
- **Poetry version**: 1.1.4
- **Link of a [Gist](https://gist.github.com/) with the contents of your pyproject.toml file**: no need, full reproducer listed below

## Issue

I'm running into an issue in which when creating a virtualenv with poetry, it is able to find a matching version, but still creates the virtualenv with the wrong version. Reproducer from scratch:

```
％ mkdir /tmp/sample-proj
％ cd /tmp/sample-proj
％ poetry init -n --python '~3.8'

This command will guide you through creating your pyproject.toml config.


You can specify a package in the following forms:
  - A single name (requests)
  - A name and a constraint (requests@^2.23.0)
  - A git url (git+https://github.com/python-poetry/poetry.git)
  - A git url with a revision (git+https://github.com/python-poetry/poetry.git#develop)
  - A file path (../my-package/my-package.whl)
  - A directory (../my-package/)
  - A url (https://example.com/packages/my-package-0.1.0.tar.gz)

％ poetry install
The currently activated Python version 3.9.0 is not supported by the project (~3.8).
Using python3 (3.8.6)
Creating virtualenv sample-proj-WINP6DfI-py3.8 in /Users/fsouza/Library/Caches/pypoetry/virtualenvs
Updating dependencies

Writing lock file
％ /Users/fsouza/Library/Caches/pypoetry/virtualenvs/sample-proj-WINP6DfI-py3.8/bin/python --version
Python 3.9.0
```

The interesting bits are:

- poetry correctly complains about my current version of Python being not supported
- poetry correctly finds a supported version
- poetry incorrectly creates the virtualenv with the wrong version

For now, as a workaround, I've been manually managing my virtualenvs.

Moreover, I am using homebrew to install poetry, so this may be a side effect of how homebrew installs poetry, and not an issue with poetry itself, but I figured using shutil.which to pass the full path of the found executable when creating the virtualenv is a harmless change. Let me know if this makes sense.

fsouza:
# Pull Request Check List

Closes #3284.

Let me know if this feels off.

- [ ] Updated **documentation** for changed code

Skipped docs since this is a bugfix? Also, didn't actually add tests, but rather updated the tests for create_env.

### DIFF
--- a/poetry/utils/env.py
+++ b/poetry/utils/env.py
@@ -730,7 +730,7 @@ class EnvManager(object):
 
                 if supported_python.allows(Version.parse(python_patch)):
                     io.write_line("Using <c1>{}</c1> ({})".format(python, python_patch))
-                    executable = python
+                    executable = shutil.which(python)
                     python_minor = ".".join(python_patch.split(".")[:2])
                     break
 
--- a/tests/utils/test_env.py
+++ b/tests/utils/test_env.py
@@ -652,6 +652,7 @@ def test_create_venv_tries_to_find_a_compatible_python_executable_using_generic_
         "subprocess.check_output",
         side_effect=check_output_wrapper(Version.parse("3.7.5")),
     )
+    mocker.patch("shutil.which", return_value="/opt/bin/python3")
     m = mocker.patch(
         "poetry.utils.env.EnvManager.build_venv", side_effect=lambda *args, **kwargs: ""
     )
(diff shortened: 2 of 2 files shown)


matangover:
Thanks, I also have the same issue and I think that this is the correct fix, since `virtualenv` has the following code to process the `python` argument:
```python
@classmethod
    def from_string_spec(cls, string_spec):
        # ...
        if os.path.isabs(string_spec):
            path = string_spec
        else:
            # ...generate fuzzy version spec based on name alone (poetry passes "python3" so virtualenv will use any Python 3.X)
```

fsouza:
Any thoughts on this?

Issue #3269, opened 2020-10-22
`Poetry export` should use per-package --index-url instead of single --extra-index-url


jaklan:
- [x] I am on the [latest](https://github.com/python-poetry/poetry/releases/latest) Poetry version.
- [x] I have searched the [issues](https://github.com/python-poetry/poetry/issues) of this repo and believe that this is not a duplicate.
- [x] If an exception occurs when executing a command, I executed it again in debug mode (`-vvv` option).

- **OS version and name**: macOS Mojave 10.14.6
- **Poetry version**: 1.1.3 with #3251 fix

## Issue

This issue is a follow-up to the problem identified in #3238:

> But... there's one more thing. I've also run:
> `poetry@3251 export -f requirements.txt -o requirements.txt --without-hashes --with-credentials`
> and the first line of the output file is:
> `--extra-index-url http://repository.intranet.foobar.com/artifactory/api/pypi/pypi-virtual/simple`
> 
> When using `default = true` we get then:
> `--index-url http://repository.intranet.foobar.com/artifactory/api/pypi/pypi-virtual/simple`
> 
> At first sight it seems okay, but we have to remember about the issue with pip and `--extra-index-url`:
> https://pydist.com/blog/extra-index-url
> https://github.com/pypa/pip/issues/5045
> I assume it's already handled by Poetry itself, but to make it error-prone when exporting the `requirements.txt` file imho we should also keep the approach of specifying individual index per each package, e.g.:
> `--index-url http://repository.intranet.foobar.com/artifactory/api/pypi/pypi-virtual/simple anyconfig==0.9.11`
> 
> PS Of course there should be also `--trusted-host` flag in any of these scenarios due to `http`, but there was another PR for that, afair - already merged, just not released yet.

@abn has already confirmed the validity of the issue, but also provided one concern:

> @jaklan appreciate the input on that. I'd suggest we raise that as a new issue. I agree that a per package index is better for the extra-index case. One worry, however, is that people seem to be attached to the current export format because some do text processing of the output it for various reasons.

So we also have to consider how to introduce such a change not to break any existing workflows.

sinoroc:
> So we also have to consider how to introduce such a change not to break any existing workflows.

I guess a new flag `--without-indexes` could help, maybe?

estyxx:
I can work on that! 
[...]

PR #3292, opened 2020-10-26
Use per-package --index-url in export command (3269)

estyxx:
# Pull Request Check List

Resolves: #3269 

Add `--with-indexes` option to the export command in order to add `--index-url` per each package.

FYI this PR is related to Hacktoberfest :)

### DIFF
--- a/poetry/console/commands/export.py
+++ b/poetry/console/commands/export.py
@@ -29,6 +29,7 @@ class ExportCommand(Command):
             multiple=True,
         ),
         option("with-credentials", None, "Include credentials for extra indices."),
+        option("with-indexes", None, "Include --index-url per each package."),
     ]
 
     def handle(self):
@@ -71,4 +72,5 @@ class ExportCommand(Command):
             dev=self.option("dev"),
             extras=self.option("extras"),
             with_credentials=self.option("with-credentials"),
+            with_indexes=self.option("with-indexes"),
         )
--- a/poetry/utils/exporter.py
+++ b/poetry/utils/exporter.py
@@ -32,6 +32,7 @@ class Exporter(object):
         dev=False,
         extras=None,
         with_credentials=False,
+        with_indexes=False,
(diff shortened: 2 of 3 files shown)


abn:
@estyxx we will have to discuss the solution for this. The issue itself will not be resolved with this, I suspect there might be value in being able to toggle the indices (but might require another issue). Note on this however, we will have to make it opt-out instead of of-in. Otherwise this will become a breaking change that might cause issues for existing workflows. :(

estyxx:
@abn I read your comments and I'm confused about the opt-out: as now the default requitement.txt is *without* per package indexing, so thought we want to make optional to include them... 
With opt-out, we are changing what is now the "default" and we will always include per-package-flag by default... 
But thinking now our goal is to ensure requirements.txt are exported in order to not choose the wrong (homonym) package during the installation, so including a per-package-flag by default can be the solution to this problem...

abn:
> I read your comments and I'm confused about the opt-out: as now the default requitement.txt is _without_ per package indexing, so thought we want to make optional to include them...
> With opt-out, we are changing what is now the "default" and we will always include per-package-flag by default...

My thought here was that we could allow for "export requirements without any index" (opt-out of using indexes) as a new option. However, thinking about this now, there really is no "demand" for such a feature.

> But thinking now our goal is to ensure requirements.txt are exported in order to not choose the wrong (homonym) package during the installation, so including a per-package-flag by default can be the solution to this problem...

There are a few concerns here. 

The first being that it is not possible to have a per-package index. Because pip does not like that. If `pip` cannot consume it, the `requirements.txt` is pretty much useless. 

The second, which is a non-technical concern, is that we will be encouraging bad bahviour here by allowing "homonym" pacakges.

neersighted:
Closing this as the exporter has been moved to https://github.com/python-poetry/poetry-plugin-export -- feel free to open an issue or PR there to discuss if this feature is desirable or not.

PR #3298, opened 2020-10-27
Deprecate to_str

stephsamson:
Relates-to: #3318

### DIFF
--- a/poetry/utils/_compat.py
+++ b/poetry/utils/_compat.py
@@ -263,26 +263,6 @@ def encode(string, encodings=None):
     return string.encode(encodings[0], errors="ignore")
 
 
-def to_str(string):
-    if isinstance(string, str) or not isinstance(string, (unicode, bytes)):
-        return string
-
-    if PY2:
-        method = "encode"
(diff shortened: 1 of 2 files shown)


poetry/utils/_compat.py
```diff
@@ -263,26 +263,6 @@ def encode(string, encodings=None):
...
-def to_str(string):
-    if isinstance(string, str) or not isinstance(string, (unicode, bytes)):
-        return string
-
-    if PY2:
-        method = "encode"
-    else:
-        method = "decode"
-
-    encodings = ["utf-8", "latin1", "ascii"]
```
abn, line 275: Should we handle this via `str(x, encoding=encoding)`? We should add a test case for PyPI repositories that evalueatets this.

abn (changes requested):

stephsamson:
Closing for #3405

abn:
@stephsamson #3405 did not remove the method though right? I think we still need  to remove encode/decode/to_str in a separate change.

PR #3301, opened 2020-10-28
hash check when installing packages from cache

ombschervister:
When installing packages, poetry uses cache to avoid redownloading files. 
If cached package file is corrupted (due to failed download, etc), poetry will not work until you delete this file from cache directory manually.

This PR adds hash check when installing packages from cache: if hash is incorrect, the file will be redownloaded.

# Pull Request Check List

### DIFF
--- a/src/poetry/installation/chef.py
+++ b/src/poetry/installation/chef.py
@@ -2,6 +2,7 @@ from __future__ import annotations
 
 import hashlib
 import json
+import logging
 
 from pathlib import Path
 from typing import TYPE_CHECKING
@@ -10,12 +11,15 @@ from poetry.core.packages.utils.link import Link
 
 from poetry.installation.chooser import InvalidWheelName
 from poetry.installation.chooser import Wheel
+from poetry.utils.helpers import get_file_hash
 
 
 if TYPE_CHECKING:
 
     from poetry.config.config import Config
     from poetry.utils.env import Env
+
+_logger = logging.getLogger(__name__)
 
 
 class Chef:
@@ -42,10 +46,6 @@ class Chef:
         return archive.suffix == ".whl"
 
     def get_cached_archive_for_link(self, link: Link) -> Link:
-        # If the archive is already a wheel, there is no need to cache it.
-        if link.is_wheel:
-            return link
-
         archives = self.get_cached_archives_for_link(link)
 
         if not archives:
@@ -81,7 +81,19 @@ class Chef:
         links = []
         for archive_type in archive_types:
             for archive in cache_dir.glob(f"*.{archive_type}"):
-                links.append(Link(archive.as_uri()))
+                accept = True
+                if link.hash is not None:
+                    real_hash = get_file_hash(
+                        archive, link.hash_name  # type: ignore[arg-type]
+                    )
+                    if real_hash != link.hash:
+                        _logger.warning(
+                            f"Cache of {link.filename} is corrupted. The file will be"
+                            " reloaded."
+                        )
+                        accept = False
+                if accept:
+                    links.append(Link(archive.as_uri()))
 
         return links
 
--- a/src/poetry/utils/helpers.py
+++ b/src/poetry/utils/helpers.py
@@ -1,5 +1,7 @@
 from __future__ import annotations
 
+import hashlib
+import io
 import os
 import re
 import shutil
@@ -156,6 +158,18 @@ def is_dir_writable(path: Path, create: bool = False) -> bool:
         return False
     else:
         return True
+
+
+def get_file_hash(filepath: Path, hash_name: str) -> str:
+    block_size = io.DEFAULT_BUFFER_SIZE
+    with open(filepath, "rb") as f:
+        res_hash = hashlib.new(hash_name)
+        while True:
+            buffer = f.read(block_size)
+            if not buffer:
+                break
+            res_hash.update(buffer)
+        return res_hash.hexdigest()
 
 
 def pluralize(count: int, word: str = "") -> str:
--- a/tests/utils/test_helpers.py
+++ b/tests/utils/test_helpers.py
@@ -1,5 +1,9 @@
 from __future__ import annotations
 
+import hashlib
+import io
+import tempfile
+
 from pathlib import Path
 from typing import TYPE_CHECKING
 
@@ -10,6 +14,7 @@ from poetry.core.utils.helpers import parse_requires
 from poetry.utils.helpers import canonicalize_name
 from poetry.utils.helpers import get_cert
 from poetry.utils.helpers import get_client_cert
+from poetry.utils.helpers import get_file_hash
 
 
 if TYPE_CHECKING:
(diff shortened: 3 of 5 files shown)


ombschervister:
I don't understand why the test fails on FreeBSD/python3.8. Can someone suggest?

poetry/installation/chef.py
```diff
@@ -75,13 +77,25 @@ def get_cached_archive_for_link(self, link):  # type: (Link) -> Optional[Link]
...
     def get_cached_archives_for_link(self, link):  # type: (Link) -> List[Link]
+
         cache_dir = self.get_cache_directory_for_link(link)
 
         archive_types = ["whl", "tar.gz", "tar.bz2", "bz2", "zip"]
         links = []
         for archive_type in archive_types:
             for archive in cache_dir.glob("*.{}".format(archive_type)):
-                links.append(Link(archive.as_uri()))
+
```
earonesty, line 87: adding a test to https://github.com/python-poetry/poetry/blob/master/tests/installation/test_chef.py would be nice to cement the contract that a bad link hash is not accepted

earonesty:
fixes: https://github.com/python-poetry/poetry/issues/3326

joooeey:
I think it would be better to have Poetry remove the broken file from the cache in place of the user. The cache is the software's responsiblity, not the user's.

ombschervister:
@joooeey 
> I think it would be better to have Poetry remove the broken file from the cache in place of the user. The cache is the software's responsiblity, not the user's.

In this request, Poetry itself removes the broken files. When loading the library, Poettry warns about the corruption of the file in the cache, and then overwrites the file itself.

image

poetry/installation/chef.py
ombschervister: done

earonesty (approved):
lgtm

earonesty:
why is this not more important?  poetry needs to check hashes, or why bother using it?

earonesty (approved):
some documentation should be added, but otherwise, looks good.

ombschervister:
> some documentation should be added, but otherwise, looks good.

Is this something I should do?
Do you mean that I should add comments to my code in the request?
Also I see the documentation in the `docs` folder but I don't see a place for such detailed implementation details

Secrus:
> > some documentation should be added, but otherwise, looks good.
> 
> Is this something I should do? Do you mean that I should add comments to my code in the request? Also I see the documentation in the `docs` folder but I don't see a place for such detailed implementation details

Hi. When it comes to documentation, it's expected to add necessary changes to docs if needed. `.md` files in the `docs` directory are named the same as sections in the documentation on the Poetry website. However, in this case, I don't think there is a need for change. We don't have information in the documentation about how the cache is working.

src/poetry/installation/chef.py
```diff
@@ -81,7 +83,18 @@ def get_cached_archives_for_link(self, link: Link) -> List[Link]:
...
         for archive_type in archive_types:
             for archive in cache_dir.glob(f"*.{archive_type}"):
-                links.append(Link(archive.as_uri()))
+
+                accept = True
+
+                if link.hash:
+                    real_hash = get_file_hash(archive, link.hash_name)
+                    if real_hash != link.hash:
+                        logger = logging.getLogger(__name__)
```
Secrus, line 92: loggers are usually declared on top of files as globals. Also, you need to add the logger to `loggers` in `src.poetry.console.application` in line 234.

src/poetry/installation/chef.py
```diff
@@ -81,7 +83,18 @@ def get_cached_archives_for_link(self, link: Link) -> List[Link]:
...
+                        logger.warning(f"cache of {link.filename} is corrupted")
```
Secrus, line 93:
```suggestion
                        logger.warning(f"Cache of {link.filename} is corrupted. It won't be used.")
```
Leaving only a warning that the cache is corrupted, might confuse user.

src/poetry/utils/helpers.py
```diff
@@ -135,6 +136,24 @@ def is_dir_writable(path: Path, create: bool = False) -> bool:
...
+        res_hash = hashlib.new(hash_name)
+
+        while True:
+            buffer = f.read(block_size)
+            if not buffer:
+                break
+
+            res_hash.update(buffer)
+
+        return res_hash.hexdigest()
```
Secrus, line 154: Is there a need for that many empty lines?

tests/utils/test_helpers.py
```diff
@@ -76,3 +80,18 @@ def test_get_client_cert(config: "Config"):
...
+
+        test_data = [b"", b"12345", b"123" * 8000]
+
+        for i, data in enumerate(test_data):
+
+            tmp_file = Path(tmp_dir) / f"file{i}"
+            tmp_file.write_bytes(data)
+
+            real_hash = hashlib.md5(data).hexdigest()
+            assert get_file_hash(tmp_file, "md5") == real_hash
```
Secrus, line 97: The same as above, I don't think you need that many empty lines.

Secrus (changes requested):

src/poetry/installation/chef.py
ombschervister: I did, but after adding the logger to `src.poetry.console.application`, the messages began to be duplicated
image

tests/utils/test_helpers.py
ombschervister: fixed

src/poetry/installation/chef.py
```diff
@@ -42,10 +46,6 @@ def is_wheel(self, archive: Path) -> bool:
         return archive.suffix == ".whl"
 
     def get_cached_archive_for_link(self, link: Link) -> Link:
-        # If the archive is already a wheel, there is no need to cache it.
-        if link.is_wheel:
-            return link
```
ombschervister, line 47: Also whl was downloaded every poetry install, but at the same time saved to the cache folders.
So the cache exists but was useless.
So I removed this lines.

ombschervister:
@Secrus should i do something else?

earonesty:
@0xDEC0DE and @sdispater worked on a similar issue, maybe they can approve.

src/poetry/installation/chef.py
```diff
@@ -81,7 +81,19 @@ def get_cached_archives_for_link(self, link: Link) -> list[Link]:
...
-                links.append(Link(archive.as_uri()))
+                accept = True
+                if link.hash is not None:
+                    real_hash = get_file_hash(
+                        archive, link.hash_name  # type: ignore[arg-type]
+                    )
+                    if real_hash != link.hash:
+                        _logger.warning(
+                            f"Cache of {link.filename} is corrupted. The file will be"
+                            " reloaded."
```
0xDEC0DE, line 92: Perhaps including the hashes in the error messages might help people spot weird problems?

i.e., if a bunch of cached files were truncated to 0 bytes, they'd all log with the same hash, and users would know something wacky had happened.

Or it might be pointless noise.  Who knows?

0xDEC0DE (approved):
Makes sense.  One small suggestion, inline.

wagnerluis1982:
I think this PR is not relevant anymore (and must be closed), since, as far as I understand, current Poetry version already check hashes for every file, cached or not.

But, perhaps the author can salvage something. Two things come to my mind: (1) I didn't find on current code a dedicated test to verify cached hash mismatch, and (2) maybe it can be introduced the nice log messages when there is a cache mismatch.

Issue #3309, opened 2020-10-30
AttributeError on poetry export


finswimmer:
I stumbled on it, while checking type hints for poetry. In the case `poetry export` is called with `-v` and no lock file exists, poetry raises an AttributeError:

```
$ poetry export -f requirements.txt -vvv 
The lock file does not exist. Locking.

  Stack trace:

  8  ~/.poetry/lib/poetry/_vendor/py3.8/clikit/console_application.py:131 in run
      129│             parsed_args = resolved_command.args
      130│ 
    → 131│             status_code = command.handle(parsed_args, io)
      132│         except KeyboardInterrupt:
      133│             status_code = 1

  7  ~/.poetry/lib/poetry/_vendor/py3.8/clikit/api/command/command.py:120 in handle
      118│     def handle(self, args, io):  # type: (Args, IO) -> int
      119│         try:
    → 120│             status_code = self._do_handle(args, io)
      121│         except KeyboardInterrupt:
      122│             if io.is_debug():

  6  ~/.poetry/lib/poetry/_vendor/py3.8/clikit/api/command/command.py:171 in _do_handle
      169│         handler_method = self._config.handler_method
      170│ 
    → 171│         return getattr(handler, handler_method)(args, io, self)
      172│ 
      173│     def __repr__(self):  # type: () -> str

  5  ~/.poetry/lib/poetry/_vendor/py3.8/cleo/commands/command.py:92 in wrap_handle
       90│         self._command = command
       91│ 
    →  92│         return self.handle()
       93│ 
       94│     def handle(self):  # type: () -> Optional[int]

  4  ~/.poetry/lib/poetry/console/commands/export.py:53 in handle
      51│                 options.append(("-v", None))
      52│ 
    → 53│             self.call("lock", options)
      54│ 
      55│         if not locker.is_fresh():

  3  ~/.poetry/lib/poetry/_vendor/py3.8/cleo/commands/command.py:107 in call
      105│             args = ""
      106│ 
    → 107│         args = StringArgs(args)
      108│         command = self.application.get_command(name)
      109│ 

  2  ~/.poetry/lib/poetry/_vendor/py3.8/clikit/args/string_args.py:20 in __init__
      18│         parser = TokenParser()
      19│ 
    → 20│         self._tokens = parser.parse(string)
      21│         self._option_tokens = list(
      22│             itertools.takewhile(lambda arg: arg != "--", self.tokens)

[...]

PR #3310, opened 2020-10-30
Fix AttributeError on export

finswimmer:
# Pull Request Check List

Resolves: https://github.com/python-poetry/poetry/issues/3309

BTW: The type hint says `call` expects a string and not a list of string. Both things seems to work :thinking:


abn (approved):
Looks like the `call` is doing something wierd might need to look at cleo/clikit.

PR #3314, opened 2020-11-01
fix and add several type hints

finswimmer:
This PR adds type hints to all method arguments and return values (except tests).

For proper work https://github.com/python-poetry/poetry-core/pull/104 must be merged first.

### DIFF
--- a/poetry/config/config.py
+++ b/poetry/config/config.py
@@ -19,11 +19,11 @@ from .dict_config_source import DictConfigSource
 _NOT_SET = object()
 
 
-def boolean_validator(val):
+def boolean_validator(val):  # type: (str) -> bool
     return val in {"true", "false", "1", "0"}
 
 
-def boolean_normalizer(val):
+def boolean_normalizer(val):  # type: (str) -> bool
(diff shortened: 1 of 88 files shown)


poetry/config/config.py
```diff
@@ -20,11 +20,11 @@
 _NOT_SET = object()
 
 
-def boolean_validator(val):
+def boolean_validator(val):  # type: (str) -> bool
```
mvoitko, line 23: why don't use type annptations?

mvoitko (changes requested):

poetry/config/config.py
finswimmer: You mean `def boolean_validator(val: str) -> bool:` instead? Because this is not supported by python2. And we haven't dropped python2 support until now. Yes the master branch is for the next feature release, which will drop python2 support. But I plan to port these changes to the 1.1 branch as well and switch to the type annotation syntax after the next feature release. Otherwise the code base between the to branches becomes to different and it becomes more difficult to port fixes between them.

sdispater (approved):

Issue #742, opened 2018-12-15
Upload to TestPyPI silently failed.


kk6:
- [x ] I am on the [latest](https://github.com/sdispater/poetry/releases/latest) Poetry version.
- [x ] I have searched the [issues](https://github.com/sdispater/poetry/issues) of this repo and believe that this is not a duplicate.
- [x ] If an exception occurs when executing a command, I executed it again in debug mode (`-vvv` option).

- **OS version and name**: Mac OSX 10.14.2
- **Poetry version**: 0.12.10
- **Link of a [Gist](https://gist.github.com/) with the contents of your pyproject.toml file**: 
https://gist.github.com/kk6/47980a543bc3f249f2bbe1d91660610d

## Issue

I tried uploading the test package to TestPyPI using Poetry.

```sh
~/s/kk6-demo $ poetry config repositories.testpypi https://test.pypi.org/simple
~/s/kk6-demo $ poetry publish -r testpypi

Publishing kk6-demo (0.1.0) to testpypi
Username: kk6



Password:

 - Uploading kk6-demo-0.1.0.tar.gz 100%
 - Uploading kk6_demo-0.1.0-py3-none-any.whl 100%
```

As far as looking at the output, the upload seems to be successful. However, when I look at TestPyPI's own project list, I can not find the package I uploaded earlier.

So, trying uploading in the same way using `twine` as a test, `HTTPError: 400 Client Error` returned.

```sh
~/s/kk6-demo $ poetry run twine upload --repository-url https://test.pypi.org/legacy/ dist/*
Enter your username: kk6
Enter your password:
Uploading distributions to https://test.pypi.org/legacy/
Uploading kk6_demo-0.1.0-py3-none-any.whl
100%|██████████████████████████████████████████████████████████████████████████████| 3.77k/3.77k [00:01<00:00, 2.36kB/s]
Uploading kk6_demo-0.1.1-py3-none-any.whl
100%|██████████████████████████████████████████████████████████████████████████████| 3.77k/3.77k [00:01<00:00, 3.13kB/s]
Uploading kk6-demo-0.1.0.tar.gz
100%|██████████████████████████████████████████████████████████████████████████████| 3.49k/3.49k [00:00<00:00, 9.05kB/s]
NOTE: Try --verbose to see response content.
HTTPError: 400 Client Error: 'None' is an invalid value for Home-Page. Error: Invalid URI See https://packaging.python.org/specifications/core-metadata for url: https://test.pypi.org/legacy/
```

[...]

PR #3315, opened 2020-11-01
Raise UploadError when HTTP 308 occurs when publishing

timotk:
# Pull Request Check List

Resolves: https://github.com/python-poetry/poetry/issues/742
Related PR: https://github.com/python-poetry/poetry/pull/3083

I was experiencing HTTP redirect errors on poetry v1.14.0, even though 301 redirects are being handled properly. It appears I was getting a 308. I have added the 308 so that it will be caught and added a new test.

@ObserverOfTime You recently worked on this as well. Could you have a look and share your thoughts?

### DIFF
--- a/poetry/publishing/uploader.py
+++ b/poetry/publishing/uploader.py
@@ -272,7 +272,7 @@ class Uploader:
                         )
                     )
                     bar.finish()
-                elif resp.status_code == 301:
+                elif resp.status_code in (301, 308):
                     if self._io.output.supports_ansi():
                         self._io.overwrite(
                             " - Uploading <c1>{0}</c1> <error>{1}</>".format(
--- a/tests/publishing/test_uploader.py
+++ b/tests/publishing/test_uploader.py
@@ -46,6 +46,18 @@ def test_uploader_properly_handles_301_redirects(http):
     )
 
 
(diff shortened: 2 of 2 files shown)


ObserverOfTime (approved):
Might as well add 302 & 307 but LGTM.

tests/publishing/test_uploader.py
```diff
@@ -46,6 +46,18 @@ def test_uploader_properly_handles_301_redirects(http):
     )
 
 
+def test_uploader_properly_handles_308_redirects(http):
```
neersighted, line 49: To avoid copy-pasting, let's parameterize this test instead.

poetry/publishing/uploader.py
```diff
@@ -272,7 +272,7 @@ def _upload_file(
                         )
                     )
                     bar.finish()
-                elif resp.status_code == 301:
+                elif resp.status_code in (301, 308):
```
neersighted, line 275: Since we're dropping Python 2 support, let's take advantage of the enums in https://docs.python.org/3/library/http.html#http.HTTPStatus instead of using the raw status codes.

neersighted (changes requested):

Secrus:
Changes made in this PR were added to the code base. Closing.

Issue #3303, opened 2020-10-28
Regression: 403 Auth Failure for S3-hosted (Cloudfront) Private Repos


dvf:
- [x] I am on the [latest](https://github.com/python-poetry/poetry/releases/latest) Poetry version.
- [x] I have searched the [issues](https://github.com/python-poetry/poetry/issues) of this repo and believe that this is not a duplicate. (There are similar tickets, notably #3291 but they aren't specifically targeting a repo hosted on S3/Cloudfront)
- [X] If an exception occurs when executing a command, I executed it again in debug mode (`-vvv` option).

- **OS version and name**: `python:3.7-alpine` (Docker)
- **Poetry version**: 1.1.4
- **Link of a [Gist](https://gist.github.com/) with the contents of your pyproject.toml file**: https://gist.github.com/dvf/a117ebdb0358cb388c5145cfdd39b46e

# Issue
- We noticed a regression from Poetry 1.0.3 to 1.1.4. 
- We're using a private PyPI repository hosted on S3 protected by Cloudfront using Basic Auth (we use [s3pypi](https://github.com/novemberfiveco/s3pypi))

When we tried to install a new private dependency, or remove a non-private dependency (with common sub dependencies with a private dependency) we saw this error:

```
➜  project git:(master) poetry remove django-structlog
Updating dependencies
Resolving dependencies... (0.2s)

  RepositoryError

  403 Client Error: Forbidden for url: https://XXXXXXXXXXXXXXXX/pyjwt/

  at ~/.poetry/lib/poetry/repositories/legacy_repository.py:393 in _get
      389│             if response.status_code == 404:
      390│                 return
      391│             response.raise_for_status()
      392│         except requests.HTTPError as e:
    → 393│             raise RepositoryError(e)
      394│ 
      395│         if response.status_code in (401, 403):
      396│             self._log(
      397│                 "Authorization error accessing {url}".format(url=url), level="warn"
```
# Hypothesis of why this is happening

**By default, Cloudfront returns a `403` for non-existing files.** This is a good security practice as it prevents leakage of information to unauthenticated parties. However, (from the above code) Poetry *only falls back to public PyPI if it encounters a `404` when requesting a package*.

# Quick fix
The quick fix for this is to have your private repo (in our case Cloudfront) return a default error response code of 404. 
Here are docs on how to do that: https://docs.aws.amazon.com/AmazonCloudFront/latest/DeveloperGuide/custom-error-pages-response-code.html

# Recommendations
[...]

Issue #3041, opened 2020-10-02
Cannot download packages from private PyPi repository using HTTP basic auth with Poetry 1.1.0 


MasterNayru:
- [x] I am on the [latest](https://github.com/python-poetry/poetry/releases/latest) Poetry version.
- [x] I have searched the [issues](https://github.com/python-poetry/poetry/issues) of this repo and believe that this is not a duplicate.
- [x] If an exception occurs when executing a command, I executed it again in debug mode (`-vvv` option).

- **OS version and name**: Fedora 32
- **Poetry version**: Poetry 1.1.0
- **Link of a [Gist](https://gist.github.com/) with the contents of your pyproject.toml file**: https://gist.github.com/MasterNayru/0014ec89ae05d8f7651d18d8a107ec90

## Issue
We have been using Poetry to pull down packages from a private PyPi repository and everything has been working fine until Poetry 1.1.0. We are configuring poetry to talk to our private PyPi installation by HTTP basic auth, and that auth works perfectly fine to resolve which versions of a package to install. The problem seems to be that that same auth is then used in the requests to download wheels from PyPi, which causes the following error to occur:

```
$ poetry config http-basic.myprivaterepo <username> <password>
$ poetry update -vvv

<snip>

   2  ~/.pyenv/versions/3.8.5/lib/python3.8/site-packages/poetry/repositories/pypi_repository.py:454 in _download
       452│ 
       453│     def _download(self, url, dest):  # type: (str, str) -> None
     → 454│         return download_file(url, dest, session=self.session)
       455│ 
       456│     def _log(self, msg, level="info"):

   1  ~/.pyenv/versions/3.8.5/lib/python3.8/site-packages/poetry/utils/helpers.py:98 in download_file
        96│ 
        97│     with get(url, stream=True) as response:
     →  98│         response.raise_for_status()
        99│ 
       100│         with open(dest, "wb") as f:

  HTTPError

  400 Client Error: Bad Request for url: https://deckard-pip.s3.amazonaws.com/1234/my_broken_dependency/my_broken_dependency-0.1.3-py3-none-any.whl?AWSAccessKeyId=<key>&Signature=kz30gf304b%2F%2F93pQeUSPrto5MiE%3D&x-amz-security-token=<token>&Expires=1601690152

  at ~/.pyenv/versions/3.8.5/lib/python3.8/site-packages/requests/models.py:941 in raise_for_status
      937│         elif 500 <= self.status_code < 600:
[...]

PR #3337, opened 2020-11-07
Fix handling errors 401, 403 when looking for repositories

cereblanco:
# Pull Request Check List

fixes error handling related to: #3303

### DIFF
--- a/tests/repositories/test_legacy_repository.py
+++ b/tests/repositories/test_legacy_repository.py
@@ -310,14 +310,8 @@ def test_get_200_returns_page(http):
     assert repo._get("/foo")
 
 
-def test_get_404_returns_none(http):
-    repo = MockHttpRepository({"/foo": 404}, http)
-
-    assert repo._get("/foo") is None
-
-
-def test_get_4xx_and_5xx_raises(http):
-    endpoints = {"/{}".format(code): code for code in {401, 403, 500}}
+def test_get_40x_and_returns_none(http):
+    endpoints = {"/{}".format(code): code for code in {401, 403, 404}}
     repo = MockHttpRepository(endpoints, http)
 
     for endpoint in endpoints:
--- a/tests/repositories/test_legacy_repository.py
+++ b/tests/repositories/test_legacy_repository.py
@@ -310,13 +310,18 @@ def test_get_200_returns_page(http):
     assert repo._get("/foo")
 
 
-def test_get_40x_and_returns_none(http):
-    endpoints = {"/{}".format(code): code for code in {401, 403, 404}}
-    repo = MockHttpRepository(endpoints, http)
+@pytest.mark.parametrize("status_code", [401, 403, 404])
+def test_get_40x_and_returns_none(http, status_code):
+    repo = MockHttpRepository({"/foo": status_code}, http)
 
-    for endpoint in endpoints:
-        with pytest.raises(RepositoryError):
-            repo._get(endpoint)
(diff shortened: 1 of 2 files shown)


MeanderingCode:
My experience of our pypicloud returning an `HTTP 401` is that `requests` raises an `HTTPError`, meaning that I believe your response status code should go in the `except HTTPError` block.

But that aside, this does not fix the authentication problems some (myself included) are finding with, for example, a pypicloud serving private packages directly with HTTP Basic auth or via authenticated AWS S3 URLs.

See https://github.com/python-poetry/poetry/issues/3041

Perhaps proper error handling, as this addresses, is separate, but I wanted to ensure you had more context.

MeanderingCode:
Looks like the exception is raised manually, below your conditional, so I believe I spoke in error.

jensgustafsson:
What's the status on this PR?

tests/repositories/test_legacy_repository.py
```diff
@@ -309,16 +309,16 @@ def test_get_200_returns_page(http):
     assert repo._get("/foo")
 
 
-def test_get_404_returns_none(http):
-    repo = MockHttpRepository({"/foo": 404}, http)
+def test_get_401_403_404_and_returns_none(http):
```
abn, line 312:
```suggestion
def test_get_40x_and_returns_none(http):
```

Additioanlly would be great to parameterise it instead of using a single test case.

abn (changes requested):
Thank you for your contribution @cereblanco. This might require a rebase as well.

tests/repositories/test_legacy_repository.py
```diff
@@ -309,16 +309,16 @@ def test_get_200_returns_page(http):
     assert repo._get("/foo")
 
 
-def test_get_404_returns_none(http):
-    repo = MockHttpRepository({"/foo": 404}, http)
+def test_get_40x_and_returns_none(http):
```
abn, line 312: You should also maybe consider using [pytest.parameterize](https://docs.pytest.org/en/stable/parametrize.html) to pass in the different status codes, this means that each case would be an independent case rather than a bulk test failure if things regress.

abn (changes requested):

tests/repositories/test_legacy_repository.py
cereblanco: Hello @abn, done rebasing/fixing conflicts, and updating tests with parameterize. 😸

stephsamson:
Hi @cereblanco this PR still needs to be rebased because of the breaking CI build. Thank you in advance!

tests/repositories/test_legacy_repository.py
abn: Oh hey! sorry i have been terrible at coming back to this. See @stephsamson comment :) Again, apologies.

abn (approved):

erdemuysal:
Hi 🤚 

How can I get information about when the new version with this change will be released? 
We need this in our scenario.

ogunoz:
@erdemuysal I faced the same issue. They approved the MR for poetry 1.2.0 release

Currently it is available only as an alpha release. It solved my problem.
You can update your poetry with either:
```poetry self update --preview```
or:
``` pip install https://github.com/python-poetry/poetry/releases/download/1.2.0a2/poetry-1.2.0a2-py3-none-any.whl```

judahrand:
Is there a reason not to back port this to 1.11? @stephsamson

Issue #3320, opened 2020-11-03
wheel_file_re does not match version correctly


radek-sprta:
- [x] I am on the [latest](https://github.com/python-poetry/poetry/releases/latest) Poetry version.
- [x] I have searched the [issues](https://github.com/python-poetry/poetry/issues) of this repo and believe that this is not a duplicate.
- [x] If an exception occurs when executing a command, I executed it again in debug mode (`-vvv` option).

- **OS version and name**: Debian 10.6
- **Poetry version**: 1.1.4
- **Link of a [Gist](https://gist.github.com/) with the contents of your pyproject.toml file**: https://gitlab.com/radek-sprta/blog/-/blob/master/pyproject.toml

## Issue
wheel_file_re fails to match version for packages with single digit versions. Example is `markdown-captions` package (filename `markdown_captions-2-py3-none-any.whl`). Results in the following error:

```
  • Installing markdown-captions (2): Failed

  AttributeError

  'NoneType' object has no attribute 'replace'

  at ~/.poetry/lib/poetry/installation/chooser.py:27 in __init__
       23│             raise InvalidWheelName("{} is not a valid wheel filename.".format(filename))
       24│ 
       25│         self.filename = filename
       26│         self.name = wheel_info.group("name").replace("_", "-")
    →  27│         self.version = wheel_info.group("ver").replace("_", "-")
       28│         self.build_tag = wheel_info.group("build")
       29│         self.pyversions = wheel_info.group("pyver").split(".")
       30│         self.abis = wheel_info.group("abi").split(".")
       31│         self.plats = wheel_info.group("plat").split(".")
```

sinoroc:
I could recreate the issue with `poetry add markdown-captions`.

@python-poetry/triage
Seems like a severe enough bug.

abn:
@python-poetry/triage

matinalsun:
I could recreate the issue with ```poetry run bootstrap```

finswimmer:
Fix is ready for review #3338

dhiltonp:
#3411 adds a test for this issue with `Wheel`

muppetjones:
Any update on when this fix will be merged and released?

jma:
> Any update on when this fix will be merged and released?

+1

stephsamson:
#3338 merged.

PR #3338, opened 2020-11-07
fix (utils.patterns): recognize one digit version in filename

finswimmer:
The current regex for wheel filename doesn't recognize one digit versions in the filename. This PR fixes this.

Furthermore the current regex makes the version part in the wheel name optional. According to PEP 491 this part is mandatory. This is fixed by this PR as well.

Feel free to provide more filenames for the tests :)

# Pull Request Check List

Resolves: https://github.com/python-poetry/poetry/issues/3320

### DIFF
--- a/tests/utils/test_patterns.py
+++ b/tests/utils/test_patterns.py
@@ -1,0 +1,39 @@
+import pytest
+
+from poetry.utils import patterns
+
+
+@pytest.mark.parametrize(
+    ["filename", "expected"],
+    [
+        (
+            "markdown_captions-2-py3-none-any.whl",
+            {
+                "namever": "markdown_captions-2",
+                "name": "markdown_captions",
+                "ver": "2",
+                "build": None,
+                "pyver": "py3",
+                "abi": "none",
+                "plat": "any",
+            },
+        ),
+        (
+            "SQLAlchemy-1.3.20-cp27-cp27mu-manylinux2010_x86_64.whl",
+            {
+                "namever": "SQLAlchemy-1.3.20",
+                "name": "SQLAlchemy",
+                "ver": "1.3.20",
+                "build": None,
+                "pyver": "cp27",
+                "abi": "cp27mu",
+                "plat": "manylinux2010_x86_64",
+            },
+        ),
+    ],
+)
+def test_wheel_file_re(filename, expected):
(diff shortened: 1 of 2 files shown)


jedie (approved):

tests/utils/test_patterns.py
```diff
@@ -0,0 +1,45 @@
...
+                "ver": "1.3.20",
+                "build": None,
+                "pyver": "cp27",
+                "abi": "cp27mu",
+                "plat": "manylinux2010_x86_64",
+            },
+        ),
+    ],
+)
+def test_wheel_file_re(filename, expected):
```
abn, line 35:
```suggestion
def test_wheel_file_re(filename, namever, name, ver, build, pyver, abi, plat):
```

abn (changes requested):
Minor recommendation to make things a bit more explicit.

dhiltonp:
This regex change will also break the case where there is no version: see #3411.

tests/utils/test_patterns.py
dhiltonp: updated in #3460

finswimmer: Hey @abn, sorry for the late response. These are a lot of parameter. So in my opinion, the definition of them is harder to read.

abn: I am nto sure if I agree with that per-say. However, my concern is more towards maintanability and the suggestion was based onthe fact that each key were explicitly tested. The suggestion is merely a recommendation happy for this to go without it if you feel strongly about it.

finswimmer: Changed as discussed at discord.

dhiltonp: Can we get this merged? Poetry completely breaks when importing these single-digit packages.

ScriptAutomate: Awesome to see this solution here. Since it looks like this has been resolved potentially by @finswimmer, are there any further unresolved items that you can find, @abn? This is a primary blocker of adoption in some of my workflows, otherwise I'm stuck monkey patching `poetry` to be able to support this with some packages in my dependency tree.

stephsamson (approved):

Issue #3265, opened 2020-10-22
Poetry can't run scripts from dependencies (which themselves use poetry) installed as editable on Windows


kevincon:
- [x] I am on the [latest](https://github.com/python-poetry/poetry/releases/latest) Poetry version.
- [x] I have searched the [issues](https://github.com/python-poetry/poetry/issues) of this repo and believe that this is not a duplicate.
- [x] If an exception occurs when executing a command, I executed it again in debug mode (`-vvv` option).

- **OS version and name**: Windows 10 Enterprise version 1909 build 18363.1139
- **Poetry version**: 1.1.3
- **Link of a [Gist](https://gist.github.com/) with the contents of your pyproject.toml file**: https://github.com/kevincon/poetry_editable_dep_script_bug_repro

## Issue

My team is experiencing that `poetry run` is unable to run scripts registered with poetry by editable-installed dependencies of our main poetry project on Windows.

I've created this repo to provide a minimal project to demonstrate this issue: https://github.com/kevincon/poetry_editable_dep_script_bug_repro

Also I'm using Python 3.8.3 32-bit.

In that repo, the main project, `parent` , has a local, editable path dependency of `child` which is itself a Poetry project.

Both the `parent` project and the `child` project register a script using the `tool.poetry.scripts` section in their respective `pyproject.toml` files. All these scripts do is print out `Hello from parent CLI!` or `Hello from child CLI!`, respectively.

Installing the `parent` project is successful:

```
C:\Users\kevincon\repos\poetry_editable_dep_script_bug_repro 
λ poetry install
Creating virtualenv parent-MPYm-h1y-py3.8 in C:\Users\kevincon\AppData\Local\pypoetry\Cache\virtualenvs
Installing dependencies from lock file

Package operations: 1 install, 0 updates, 0 removals

  • Installing child (0.1.0 C:/Users/kevincon/repos/poetry_editable_dep_script_bug_repro/child)

Installing the current project: parent (0.1.0)
```

As is running the `parent` project's script:

```
C:\Users\kevincon\repos\poetry_editable_dep_script_bug_repro 
λ poetry run parent-cli
Hello from parent CLI!
```

But trying to run the `child` project's script fails with this error (with 100% reproducibility):

```
C:\Users\kevincon\repos\poetry_editable_dep_script_bug_repro 
λ poetry run child-cli

  FileNotFoundError

  [WinError 2] The system cannot find the file specified

[...]

Issue #2481, opened 2020-06-02
poetry run causes [FileNotFoundError] with some scripts


huwper:
- [x] I am on the [latest](https://github.com/python-poetry/poetry/releases/latest) Poetry version.
- [x] I have searched the [issues](https://github.com/python-poetry/poetry/issues) of this repo and believe that this is not a duplicate.
- [x] If an exception occurs when executing a command, I executed it again in debug mode (`-vvv` option).

- **OS version and name**: Windows 10 Enterprise
- **Poetry version**: Poetry version 1.0.5
- **Link of a [Gist](https://gist.github.com/) with the contents of your pyproject.toml file**: https://gist.github.com/huwper/efb95344f986939bf4fd717fbeb64a91

## Issue

Using poetry run to run scripts fails on some scripts. I think this is probably in cases where the script has no file extension. The linked pyproject.toml installs scons into the virtual environment. Scons  comes with a script called `scons` which is installed in the `Scripts` directory. The following error occurs when running the script:

```
c:\Projects\personal\poetry-scons-issue>poetry run scons

[FileNotFoundError]
[WinError 2] The system cannot find the file specified
```

Running `poetry venv info` to get path/to/venv and running scons with the explicit path works fine

```
> path\to\venv\Scripts\scons

scons: *** No SConstruct file found.
......
```

scons also comes with a script called `scons.bat` which does the same thing. `poetry run scons.bat` does work

jonapich:
Same problem here, slightly different setup:

- Project B defines a script called `genall`
- Project A has a `path` dependency to Project B
- I can run `poetry run genall` in Project B
- I cannot in project A (FileNotFoundError)
- In Project A's virtual env, in the Scripts folder, there are 2 files: `genall` and a `genall.cmd`

jonapich:
I tried a workaround, but it didn't work. The result is weird so I'm sharing it:

- Added a `poetry.scripts` section to Project A, a copy of the one in Project B
- Recreated the virtual environment from scratch
- Tried to launch `poetry run genall` from Project A again:

```
PS C:\...> poetry run genall

  FileNotFoundError

  [WinError 2] The system cannot find the file specified

  at ~\AppData\Local\Programs\Python\Python38\lib\subprocess.py:1307 in _execute_child
      1303│             sys.audit("subprocess.Popen", executable, args, cwd, env)
      1304│
      1305│             # Start the process
      1306│             try:
[...]

PR #3339, opened 2020-11-07
Fix running console scripts of editable dependencies on Windows

kevincon:
# Pull Request Check List

Resolves: #3265
Resolves: #2481

- [x] Updated **documentation** for changed code. (I'm not aware of any documentation that needs updating for this change)

This PR changes `poetry run` to use `shell=True` when executing the user-provided command on Windows.

This ensures that the [CMD wrapper script file that Poetry generates for console scripts of editable dependencies](https://github.com/python-poetry/poetry/blob/d646d45d8776a88bff6ad8126f52b92cd944f8b9/poetry/masonry/builders/editable.py#L172-L184) can be called via `poetry run` by only specifying the name of the console script (i.e. without the `.cmd` extension of the wrapper script file) because `shell=True` causes the command to be executed via `cmd.exe` on Windows which has support for calling `.cmd` files without specifying the `.cmd` extension. 

I don't think that adding `shell=True` here only for Windows should cause any issues. I saw that the `run()` method on the `Env` class (which calls the `_run()` private method) already uses `shell=True` when executing any command on Windows: https://github.com/python-poetry/poetry/blob/825cf6602319f8e31426e6d557a122a230229e83/poetry/utils/env.py#L1067-L1068

I also think I understand from #1236 that the motivation for having a separate `execute()` method from the `run()` method on the `Env` class was so that `poetry run` could use `execute()` to ensure that signals sent to `poetry` are forwarded correctly to what is executed by `poetry run`, but I don't think that adding `shell=True` should interfere with that goal on Windows.


kevincon:
@finswimmer can you help me find a reviewer for this PR?

arzamas500:
Any progress on that? Seems like an easy-fix?

kevincon:
@arzamas500 I have resolved the merge conflicts, but we will need a maintainer to approve the check workflows since this is my first time contributing.

aarmn:
What happened in the end? Should we use his version meanwhile it's being reviewed? is it merged and I am too novice to understand? or is this critical issue that make poetry unusable on win is open since Apr 2021? 
Is there any contribution we can do to speed up the process? I'm pretty noob in this GitHub ci stuff

kevincon:
@finswimmer it's been a few years since I opened this PR and I believe the targeted issues are still present in the latest Poetry release(s). Is there anything I need to do before a maintainer would approve running the remaining PR checks (since I am a first-time contributor) and/or review this PR?

abn:
@kevincon I have approved the run. Will review it in the coming week unless someone else gets to it before me.

kevincon:
Thanks @abn! I saw a test that was recently introduced in #5243 failed on Windows due to the changes in this PR, so I just pushed a commit to fix it. Can you approve running the checks once more?

abn:
@kevincon how does this change impact non command prompt shells on windows?

kevincon:
> how does this change impact non command prompt shells on windows?

@abn I'm not familiar with any other shells used on Windows besides Command Prompt (but understand [they are technically supported by Python via the `COMSPEC` environment variable](https://docs.python.org/3/library/subprocess.html#:~:text=On%20Windows%20with%20shell%3DTrue%2C%20the%20COMSPEC%20environment%20variable%20specifies%20the%20default%20shell.)). A quick skim of https://en.wikipedia.org/wiki/Comparison_of_command_shells suggests there might only be https://jpsoft.com/products/tcc-le.html and https://www.4dos.info/, both created by JP Software. The latter appears to have been abandoned (last release in 2009), and the former claims to be a superset of Command Prompt and aims to be compatible with it.

Secrus:
@kevincon how about Powershell? is it fully compatible with CMD?

kevincon:
@Secrus I don't think Powershell is compatible with `COMSPEC`, see https://social.technet.microsoft.com/Forums/lync/en-US/584fff37-af8b-4560-bcb0-324c84275a09/using-powershell-as-comspec?forum=ITCG and https://superuser.com/a/289760.

I don't think that Powershell not being compatible with `COMSPEC` should prevent us from making this change, just that Powershell in general can't be used with `COMSPEC`.

neersighted (approved):

dimbleby:
Per https://github.com/python-poetry/poetry/issues/5773#issuecomment-1146876056 I suspect this has broken commands with more than one word.

Issue #2778, opened 2020-08-07
Build packages with pinned dependencies from .lock file


artslob:
- [x] I have searched the [issues](https://github.com/python-poetry/poetry/issues) of this repo and believe that this is not a duplicate.
- [x] I have searched the [documentation](https://python-poetry.org/docs/) and believe that my question is not covered.

## Feature Request

As title says, this feature is about to `poetry build` build wheels and packages with pinned versions specified in `poetry.lock` file. Command can be something like `poetry build --locked`.

Its duplicate of closed #1307. Reason to close this issue was argument about libraries and applications (in short):
> Libraries use wheels for distribution. Their dependencies should not be very tight so libraries can coexist with each other. Thats why `poetry build` dont use versions from lock file. There are also applications. Applications need pinned dependencies from lock file, so all environments (dev and prod) use same dependencies. Applications should use docker for distribution.

So I think its not actually true that applications should not use wheels and pypi for distribution, sometimes its reasonable to have package with pinned dependencies from lock file. Example of such use case can be some cli tools like https://github.com/ytdl-org/youtube-dl, so development team and end-user use same dependencies. Another example is some backend application that deployed on server by wheel package.

tomzx:
I'd think it is reasonable for a tool such as `poetry` to generate `Requires-Dist` constraints in the wheel's `METADATA`, knowing that it is the user's responsibility to ensure that whatever they've built can be installed in the other projects that need them.

I assume that what you want as behavior is that each package listed in your `pyproject.toml` have the exact version defined in your `poetry.lock`. All the packages those packages depend on would not be part of the list of packages defined in the built wheel.

Is that an accurate assessment?

artslob:
Well, no actually. What I want is that when package is builded it will require exact versions of all dependencies from lock file. These versions can be extracted using poetry export -f requirements.

Right now when package is builded it requires versions specified in pyproject.toml file.

Tobotimus:
I'm certainly a fan of this idea, and made a similar [comment](https://github.com/python-poetry/poetry/issues/1307#issuecomment-674336917) regarding that "Applications are not distributed via wheels" argument on the original issue.

sinoroc:
@artslob @Tobotimus @tomzx A PR has been filed: https://github.com/python-poetry/poetry/pull/3341. Would you mind giving it a try, giving feedback on it, reviewing the code? Whatever helps to get things moving

artslob:
I will try to find some time to check it

[...]

Issue #3334, opened 2020-11-06
Distinguish between libraries and applications


MartinThoma:
- [x] I have searched the [issues](https://github.com/python-poetry/poetry/issues) of this repo and believe that this is not a duplicate.
- [x] I have searched the [documentation](https://python-poetry.org/docs/) and believe that my question is not covered.

## Feature Request
When you have a pure application, e.g. a package which is intended not to be imported by other packages, but only to be executed directly (e.g. CLI application, a web service, some GUI program), it would be nice to have reliable installations at the client.

For reliable installations, it is important that the dependencies (including transitive ones) are pinned. They are pinned in the poetry.lock file, but that file is not uploaded / used when a package built with poetry, uploaded to PyPI.org is installed via pip.

My suggestion is:

* Within `[tool.poetry]`, add `is-pure-application: true / false` or similar (never-included? no-library?)
* When this attribute is present and true, then poetry adds the transitive dependencies to the shipped `setup.py` file
* Then, when installed via `pipx`, one can be absolutely certain about what is installed and have way more control about the environment.

I'm uncertain, however, how to deal with the "normal" pip which does not create a virtual environment for such a package. That would be problematic.

sinoroc:
I get the general idea, but the suggested method does not make sense to me.

From my point of view, what would make more sense would be to instruct `poetry build` to write the pinned dependencies as `Requires-Dist` in the distributions.

This is typically the kind of things I would like to see implemented as a plugin, not necessarily part of _poetry_ proper. Because this seems like one nice way to shoot yourself in the foot, if you pin everything in the distributions without knowing what you are doing.

> I'm uncertain, however, how to deal with the "normal" pip which does not create a virtual environment for such a package. That would be problematic.

This does not seem feasible. I do not see what role _pip_ would play in that. I'd rather recommend another plugin that would output a _zipapp_ containing pinned dependencies.

MartinThoma:
> this seems like one nice way to shoot yourself in the foot

Yes, for sure 😁 👍

MartinThoma:
Could you please point me to documentation about how to create plugins for poetry?

sinoroc:
The plugin system is under development, scheduled for [version 1.2](https://github.com/python-poetry/poetry/projects/6):

* [_WIP - Plugin system (#1237)_](https://github.com/python-poetry/poetry/pull/1237)

finswimmer:
[...]

PR #3341, opened 2020-11-08
Build packages with pinned dependencies from .lock file

mrijken:
# Pull Request Check List

Resolves: #2778

### DIFF
--- a/docs/docs/cli.md
+++ b/docs/docs/cli.md
@@ -320,6 +320,7 @@ Note that, at the moment, only pure python wheels are supported.
 ### Options
 
 * `--format (-f)`: Limit the format to either `wheel` or `sdist`.
+* `--lock (-l)`: Add all pinned dependencies from `poetry.lock` to `pyproject.toml` of source and wheels archive
 
 ## publish
 
--- a/poetry/console/commands/build.py
+++ b/poetry/console/commands/build.py
@@ -9,7 +9,8 @@ class BuildCommand(EnvCommand):
     description = "Builds a package, as a tarball and a wheel by default."
 
     options = [
-        option("format", "f", "Limit the format to either sdist or wheel.", flag=False)
+        option("format", "f", "Limit the format to either sdist or wheel.", flag=False),
+        option("lock", "l", "Add all dependencies as locked dependencies in the distributed pyproject.toml", flag=True)
     ]
 
     loggers = [
@@ -25,6 +26,8 @@ class BuildCommand(EnvCommand):
         if self.option("format"):
             fmt = self.option("format")
 
+        lock = self.option("lock")
+
         package = self.poetry.package
         self.line(
             "Building <c1>{}</c1> (<c2>{}</c2>)".format(
@@ -32,5 +35,24 @@ class BuildCommand(EnvCommand):
             )
         )
 
+        original_content = self.poetry.file.read()
+        if lock:
+            section = "dependencies"
+
+            content = self.poetry.file.read()
+            poetry_content = content["tool"]["poetry"]
+            if section not in poetry_content:
+                poetry_content[section] = {}
+
+            for dependency_package in self._poetry.locker.lock_data["package"]:
+                name = dependency_package["name"]
+                version = dependency_package["version"]
+                poetry_content[section][name] = version
+
+            self.poetry.file.write(content)
+
         builder = Builder(self.poetry)
         builder.build(fmt, executable=self.env.python)
+
+        if lock:
+            self.poetry.file.write(original_content)
--- a/poetry/console/commands/build.py
+++ b/poetry/console/commands/build.py
@@ -10,7 +10,12 @@ class BuildCommand(EnvCommand):
 
     options = [
         option("format", "f", "Limit the format to either sdist or wheel.", flag=False),
-        option("lock", "l", "Add all dependencies as locked dependencies in the distributed pyproject.toml", flag=True)
+        option(
+            "lock",
+            "l",
+            "Add all dependencies as locked dependencies in the distributed pyproject.toml",
+            flag=True,
+        ),
     ]
 
     loggers = [
--- a/poetry/console/commands/build.py
+++ b/poetry/console/commands/build.py
@@ -25,7 +25,7 @@ class BuildCommand(EnvCommand):
     ]
 
     def handle(self):
-        from poetry.core.masonry import Builder
+        from poetry.core.masonry import builder
 
         fmt = "all"
         if self.option("format"):
@@ -49,15 +49,18 @@ class BuildCommand(EnvCommand):
             if section not in poetry_content:
                 poetry_content[section] = {}
 
-            for dependency_package in self._poetry.locker.lock_data["package"]:
+            for dependency_package in self.poetry.locker.lock_data["package"]:
                 name = dependency_package["name"]
                 version = dependency_package["version"]
                 poetry_content[section][name] = version
 
-            self.poetry.file.write(content)
+            self._write_pyproject_toml(content)
 
-        builder = Builder(self.poetry)
+        builder = builder.Builder(self.poetry)
         builder.build(fmt, executable=self.env.python)
 
         if lock:
-            self.poetry.file.write(original_content)
+            self._write_pyproject_toml(original_content)
+
+    def _write_pyproject_toml(self, content):
+        self.poetry.file.write(content)
--- a/tests/console/commands/test_build.py
+++ b/tests/console/commands/test_build.py
@@ -1,0 +1,83 @@
+import pytest
+
+from poetry.console.commands.build import BuildCommand
+from poetry.core.masonry import builder
+from poetry.utils._compat import Path
+
+
+@pytest.fixture
+def source_dir(tmp_path):  # type: (Path) -> Path
+    yield Path(tmp_path.as_posix())
+
+
+@pytest.fixture
+def tester(command_tester_factory):
+    return command_tester_factory("build")
+
+
+class MockedBuilder:
+    def __init__(self, *args, **kw):
+        pass
+
+    def build(self, *args, **kw):
+        pass
+
+
+@pytest.fixture
+def poetry_with_lock_file(project_factory, fixture_dir):
+    source = fixture_dir("lock")
+    pyproject_content = (source / "pyproject.toml").read_text(encoding="utf-8")
+    poetry_lock_content = (source / "poetry.lock").read_text(encoding="utf-8")
+    return project_factory(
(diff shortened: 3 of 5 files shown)


sinoroc:
@mrijken 
Is that the same feature request as has been discussed in https://github.com/python-poetry/poetry/issues/3334 ?

mrijken:
Yes, the use case is the same as in #3334.

mrijken:
What is needed to get this PR reviewed? Or is it just time ;-)

sinoroc:
> What is needed to get this PR reviewed? Or is it just time ;-)

Maintainers need to find time, I guess. I asked some people who seem to want a similar feature to look at your PR, test it if possible.

Also note: https://github.com/python-poetry/poetry-core/pull/71

MartinThoma:
Thank you for taking the time and doing it! The `--lock` option makes sense to me, although it for sure means that people should install such "locked" applications in their own (isolated) environment. As @sinoroc phrased it in another context: "this seems like one nice way to shoot yourself in the foot". Without more tooling support and more documentation, I fear that this might be used in a wrong way: Either by people locking dependencies in libraries or users installing applications in the global environment.

Also, how could you now upgrade the dependencies? Would this locking basically lose all information about version restrictions in the pyproject.toml?

mrijken:
For sure the people who uses this, needs to know the consequences. In our company we want to be 100% sure that the app and the dependencies build and checked by Jenkins is the same as the one installed. We check regular whether updates of dependencies are present and can rebuild the app with the locked dependencies again. When we install that new version of the app, we know for sure that it passes all tests with the upgraded dependencies. Are there other methods / best practices to achieve that without a lock?

sinoroc:
Maybe an alternative solution is to export the lockfile of the application (`App`) to a `requirements.txt` and the install in staging/production as `pip install App -r requirements.txt`.
_[Off the top of my head, I have not tested this.]_

abn:
@mrijken sorry it took a while for me to get to this one. I do not think that this is the right level at which we should support this feature. The main reason why I say this is because this introduces 2 different kinds of wheels when (a) built via `poetry` as the front end using the `--lock` option and (b) when built via another PEP-517 compatible front-end. This is, imho, is a dangerous inconsistency. This is why I proposed the original change in `poetry-core` as this allows for a consistent behaviour.

In the current situation, the common solution for relying on pinned version in an environment is to use the `poetry install --no-dev` command.

artslob:
@abn can you please clarify about what inconsistency are you talking? 

>common solution for relying on pinned version in an environment is to use the poetry install --no-dev command

But `--no-dev` flag just allows to **install** package without dependencies. Also for this flag to work you need download source code.

Flag that was requested in #2778 is about `building` packages with dependecnies as specified in `poetry.lock` file. If maintainer decided to build package with `--lock` flag it means package should be installed with same dependencies as in lock file. I assume this flag would be primary used for private package management, not for public libraries.

sinoroc:
One thing that came to my mind, while seeing this topic pop up again:

This has to be done carefully. If there is a way to build 2 wheels (or 2 sdist) of the same project, 1 with pinned dependencies and 1 without pinned dependencies, then they definitely should have different file names. Because most of the packaging tools are built around the idea that if 2 distributions have the same file name then they are the exact same thing. So those 2 distributions (pinned and unpinned) should be differentiable by comparing the file names.

I do not know what is the best way to go about this, maybe by adding something to the version string (see PEP 440).

If only the pinned wheel of the project is distributed ever, then I guess it is all fine. But if it is intended to publish the 2 variants (even privately), then I think there might be issues.

_[I did not check if it has been said already or not, apologies if it has]_

artslob:
From PEP 440
>Specific build information may also be included in local version labels

Maybe add `+lock` local label to builded version?

But
>local version identifiers MUST NOT be permitted in version specifiers

Also
>"local version label" concepts to allow system integrators ... as well as to allow the incorporation of build tags into the versioning of binary distributions

artslob:
>If there is a way to build 2 wheels (or 2 sdist) of the same project, 1 with pinned dependencies and 1 without pinned dependencies, then they definitely should have different file names

Is it really need 2 different names? From version convention (PEP 440) there is no standartizied way to differentiate locked and ususal versions, so maybe it should be responsibility of package maintainer to choose to build package as locked or unlocked versions?

sinoroc:
@artslob 

> > If there is a way to build 2 wheels (or 2 sdist) of the same project, 1 with pinned dependencies and 1 without pinned dependencies, then they definitely should have different file names
> 
> Is it really need 2 different names? From version convention (PEP 440) there is no standartizied way to differentiate locked and ususal versions, so maybe it should be responsibility of package maintainer to choose to build package as locked or unlocked versions?

If the project maintainers only ever publish the distribution variant with pinned dependencies, then fair enough, no problem. Same if they only ever publish the distribution with unpinned dependencies. BUT, if project maintainers publish both a distribution with pinned dependencies and another with unpinned, then they should have different file names. So, maybe poetry should also offer a way to do this name differentiation (if possible according to PEP440 of course).

Maybe this is a case where there should be an option in `pyproject.toml` to say: "by default all distributions of this project should be created with pinned dependencies", reason most likely is that the project is an application and not a library. The point would be to reduce the risk of having both pinned and unpinned being published. But this is very risky from my point of view, what gets pinned today, might not be the same thing that will get pinned tomorrow. The results of the dependency resolution (the lock file) depends on what is available on the index (PyPI), and it changes constantly.

To be honest, my personal point of view right now, is that I don't think it makes sense to have pinned dependencies in a distribution. There are other ways to solve what we are trying to solve here. On of the possible solutions is to look towards things like zipapp, pex, shiv, etc. those are much better suited for applications with pinned dependencies. Note that there is also a general intention to formalize/standardise a "_requirements_" format (the next `requirements.txt`) but from what I understood, it is very challenging task to get this right.

artslob:
>BUT, if project maintainers publish both a distribution with pinned dependencies and another with unpinned, then they should have different file names

Why?

---

Also, if we always build packages with poetry using requirement constraints from pyproject.toml (usually in form of `pkg >= 1.0 < 2.0`) what point of having `.lock`. file with pinned deps? I think `.lock` file has sense for application type projects.

I think idea of specifing type of python project (library or application) is good, so it will have consistency with builder and explains existance of `.lock` file.

abn:
> @abn can you please clarify about what inconsistency are you talking?

If we follow this proposal, you have 2 different behaviours here for building project wheels.

1. Using `poetry build --lock`.
2. Using `poetry build` or using a PEP-517 frontend (eg: `pip wheel`).

Since poetry is not always involved in the (2) case there is no way to reproduce the "locked" builds consistently. It is important to note that there are a lot of cases where, even when a wheel exists on pypi, due to system compatibilities or end user choices (eg: `pip install --no-binary :all:`), you end up having to rebuild the package wheel from source relying on PEP-517 builds.

> But `--no-dev` flag just allows to **install** package without dependencies. Also for this flag to work you need download source code.

Correct.
 
> Flag that was requested in #2778 is about **building** packages with dependecnies as specified in `poetry.lock` file. If maintainer decided to build package with `--lock` flag it means package should be installed with same dependencies as in lock file. I assume this flag would be primary used for private package management, not for public libraries.

Sure. But we cannot really introduce a new option with this assumption unfortunately. Addtiionally, as noted above, we cannot guarantee that a wheel will always be installed either. This is particularly why we need to ensure both cases mentioend above behave consistently.

Another point of note here is that wheels may not necessarily be the right solution for a "locked" solution. It might be that a [zipapp](https://docs.python.org/3/library/zipapp.html#module-zipapp) might serve the use case better especially in the "private" package use cases.

tharradine:
I have an alternative idea for this problem, which still allows the building of wheels with locked dependencies. I would really love to continue allowing the distribution of applications via wheels - whilst zipapp or docker might currently provide more conducive features for distributing applications, wheels are still just something "installable" - there's nothing about them that restricts them to be used as a library, it's just that, IMO, current build backends (such as poetry-core and setuptools) don't lend themselves to creating wheels for applications.

I would say this feature belongs in the build backend (poetry-core), and should be configurable from the `pyproject.toml` file. E.g. if we add an optional key in the `[build-system]` table such as `locked_dependencies = true|false`, then poetry-core would build the wheel with locked dependencies. I believe this would remove that inconsistency which @abn is (rightfully) concerned about, as it becomes frontend-agnostic. Whilst this design might not be as "flexible" for the person building the wheel, I think it captures the real use case of #3334 correctly - this key would be set to true for applications, false otherwise.

bneijt:
I'm using a separate script to achieve this functionality and I'm interested in getting this to work as a poetry plugin in the future. For now, just using `poetry add --dev poetry-lock-package` works for all my build requirements.

I've done a write-up here: https://bneijt.nl/pr/poetry-lock-package/

tom-bowles:
> Anyway, I believe this whole thing of having locked dependencies in the wheel is a bad idea. With this you are kind of making it hard to install security fixes of the dependencies and things like that.
> 
> If I am not mistaken, if you were to try to update a dependency in such a case, pip would complain. Maybe it would not prevent the update, but it would at least display some message.

My use case is for an internal application that will *always* be deloyed using pipx. Pipx creates a separate virtualenv just for the application and installs only that application in it. Adopting a fixed version of a dependency would be up to the application owners, who would update the lock file, test and publish a new version of the application distribution.

Applications are inherently different to libraries. An application is the root node of the dependency graph, which means it *can* and arguably *should* enforce specific versions across that entire graph, thus ensuring that the dependency versions used are consistent throughout development, testing and production deployment. This is the primary purpose of lock files as I see it - they're less critical for libraries, though still very useful.

sinoroc:
@tom-bowles Applications and libraries have to be handled differently, yes. I still believe pinning all dependencies in the application's wheel metadata is counter-productive. If a dependency gets a security fix released, I don't want to wait until the maintainers of the application update the pinned dependencies and release a new wheel of the application, I want to update the dependency right now.

richardwhitefoot:
I think it would be good for this to became a core feature and give users the choice whether to use it or not.  The use-case for building internal applications is good, and when deploying the same application to multiple environments (ie. internet & secure extranet) you really want to be sure dependencies are consistent and without bypassing a lot of the good features of Poetry you can't do that at present.

bneijt:
I think this PR will not make it, meanwhile there has been the initiation of the export plugin for the newer poetry releases and if this PR is going to be accepted it probably needs to land there.

For the closely related "build a wheel file with pinned dependencies" I have already created an issue there: https://github.com/python-poetry/poetry-export-plugin/issues/1

neersighted:
Per all the discussion above, I'm going to close this for now as this implementation is a likely dead end. I'd suggest discussion occur in #2778, and we can migrate to a Github discussion if desired.

Issue #3110, opened 2020-10-06
poetry 1.1.x causes SSLCertVerificationError when installing packages from private repo


zyxue:
I have a poetry project, configured with a private pypi repo, that used to work well, but after upgrading poetry to 1.1.x (tried 1.1.0 to 1.1.2), it causes error like

```
#16 26.50 Retrying HTTP request in 0.5 seconds.
#16 27.13 Retrying HTTP request in 1.0 seconds.
#16 28.25 Retrying HTTP request in 1.5 seconds.
#16 29.87 Retrying HTTP request in 2.0 seconds.
#16 32.00 Retrying HTTP request in 2.5 seconds.
#16 34.71 
#16 34.71   SSLError
#16 34.71 
#16 34.71   HTTPSConnectionPool(host='<PRIVATE-REPO-HOST>)', port=443): Max retries exceeded with url: <URL-to-myapp-0.1.3-py3-none-any.whl (Caused by SSLError(SSLCertVerificationError(1, '[SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed: unable to get local issuer certificate (_ssl.c:1124)')))
#16 34.71 
#16 34.71   at /usr/local/lib/python3.8/site-packages/requests/adapters.py:514 in send
#16 34.76       510│                 raise ProxyError(e, request=request)
#16 34.76       511│ 
#16 34.76       512│             if isinstance(e.reason, _SSLError):
#16 34.76       513│                 # This branch is for urllib3 v1.22 and later.
#16 34.76     → 514│                 raise SSLError(e, request=request)
#16 34.76       515│ 
#16 34.76       516│             raise ConnectionError(e, request=request)
#16 34.76       517│ 
#16 34.76       518│         except ClosedPoolError as e:
#16 34.76 
#16 ERROR: executor failed running [/bin/sh -c poetry install --no-dev]: runc did not terminate sucessfully
```

when downloading packages that are only available in the private repo.

When I downgrade poetry to 1.0.10, it then works fine.

abn:
@zyxue can you try setting [`REQUESTS_CA_BUNDLE`](https://requests.readthedocs.io/en/master/user/advanced/)?

neilpanchal:
Is it possible to ignore SSL certs in poetry? We have an internal private pypi repository hosted without SSL certs.

zyxue:
[...]

Issue #2593, opened 2020-06-25
poetry config certificates.my-private-pypi.cert doesn't work


z0h:
- [x] I am on the [latest](https://github.com/python-poetry/poetry/releases/latest) Poetry version.
- [x] I have searched the [issues](https://github.com/python-poetry/poetry/issues) of this repo and believe that this is not a duplicate.
- [x] If an exception occurs when executing a command, I executed it again in debug mode (`-vvv` option).

- **OS version and name**: MacOS 10.15.5
- **Poetry version**: 1.09
- **Link of a [Gist](https://gist.github.com/) with the contents of your pyproject.toml file**: https://gist.github.com/nbka/5cca8fdacc04700a1abbbfd4b31e8d4a
(irrelevant bits omitted)

## Issue
Configuring a certificate for a private repo via the below doesn't work:

`poetry config certificates.my-private-pypi.cert /Users/xxx/yyy_ca.pem`

However pointing to the pem file via "export REQUESTS_CA_BUNDLE" will make it work. If it matters the below configs are also set:

```
virtualenvs.create = true
virtualenvs.in-project = true
```

jamiemoore:
I can't seem to make export REQUESTS_CA_BUNDLE work either.  When set, Pip works fine.  Poetry not so much.  Configured as per the documentation. https://python-poetry.org/docs/repositories/ using poetry version 1.1.4

pierresouchay:
same issue here

intgr:
I can confirm this is fixed in Poetry 1.2.0b2

Issue #1012, opened 2019-04-03
Allow root CA bundle to be configured


jobec:
- [ ] I have searched the [issues](https://github.com/sdispater/poetry/issues) of this repo and believe that this is not a duplicate.
**This is a duplicate of closed issue #790**
- [x] I have searched the [documentation](https://poetry.eustace.io/docs/) and believe that my question is not covered.

## Feature Request

Allow and alternate root CA bundle to be specified from CLI and in the global/project settings file. Similar to how pip does it. This is crucial in environments where direct access to pypi is blocked and when using internal pypi mirrors with certificates signed by an internal root CA. Or with SSL traffic inspecting firewalls.

https://pip.pypa.io/en/stable/reference/pip/?highlight=proxy#cmdoption-cert

Having to set an environment variable like `REQUESTS_CA_BUNDLE` for an underlying package is not user-friendly and expects the user to know what poetry does under the hood.

MichalMazurek:
Hi,
It would be also nice to disable certificate checking, I'm having a lot of issues with a self-signed certificate for private PyPI repo. I can fix that, but first, it would be great to know if this will be welcomed into the project. This can be done together with adding support for the request above.

hangtwenty:
can you confirm/deny whether `REQUESTS_CA_BUNDLE` actually works?

>Having to set an environment variable like REQUESTS_CA_BUNDLE for an underlying package is not user-friendly and expects the user to know what poetry does under the hood.

Arguably that's just a documentation problem.

While encapsulation can be good, `requests` is _extremely common,_ so much so that most people treat it as if it's basically part of the Python stdlib. `REQUESTS_CA_BUNDLE` is barely more 'under the hood' of a parameter, than `PYTHONUNBUFFERED` or other python-official features. So really as long as it is documented and guaranteed that `requests` is used, that could be enough for immediate future.

----

The bigger question, I think, is whether/how-much to expose Pip directly through Poetry. 

i.e. In pytest there is the env var, `PYTEST_ADDOPTS` and it appends flags to any pytest invocation. there could be `POETRY_PIP_ADDOPTS` etc. _maybe._ Or instead of an env var it could be something in config; with pytest as example again, it has `pytest_addopts = ...` supported from its config file (`.ini` in that case, similar to `.toml`). https://docs.pytest.org/en/latest/customize.html#adding-default-options Of course it won't be exactly like pytest, it's just a worthwhile example to consider.

[...]

PR #3349, opened 2020-11-10
Use configured CA certificate when downloading packages

intgr:
# Pull Request Check List

Resolves: possibly #3110 #2593 #1012 (I haven't looked into the nuances)

It seems that #1325 tried to add support to supply custom CA certificates for custom 3rd party repositories (via `poetry config certificates.$repo.cert`).

But the implementation was incomplete: `poetry lock` properly uses the correct CA certs for custom repository access, but `poetry install` still fails with a cert error because the package download code path does not use the configured CA certificates.

This issue surfaced for us when upgrading Poetry from 1.1.3 to 1.1.4, the minor release had change #3251 that gave our custom repository priority (it was ignored previously); the root cause was challenging to track down.

This PR is a hack to make artifact downloads use configured CA file too. I wanted to submit this first to get feedback whether I'm on the right track.

Error that occurs without this patch:
```
% poetry install

  • Installing django (3.1.3): Pending...
Retrying HTTP request in 0.5 seconds.
Retrying HTTP request in 1.0 seconds.
Retrying HTTP request in 1.5 seconds.
Retrying HTTP request in 2.0 seconds.
  • Installing django (3.1.3): Failed

  SSLError

  HTTPSConnectionPool(host='***', port=443): Max retries exceeded with url: /api/pypi/pypi/packages/packages/7f/17/16267e782a30ea2ce08a9a452c1db285afb0ff226cfe3753f484d3d65662/Django-3.1.3-py3-none-any.whl (Caused by SSLError(SSLCertVerificationError(1, '[SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed: self signed certificate in certificate chain (_ssl.c:1122)')))

  at /usr/local/Cellar/poetry/1.1.4/libexec/vendor/lib/python3.9/site-packages/requests/adapters.py:514 in send
      510│                 raise ProxyError(e, request=request)
      511│ 
      512│             if isinstance(e.reason, _SSLError):
      513│                 # This branch is for urllib3 v1.22 and later.
    → 514│                 raise SSLError(e, request=request)
      515│ 
      516│             raise ConnectionError(e, request=request)
      517│ 
      518│         except ClosedPoolError as e:
```

### DIFF
--- a/poetry/installation/chooser.py
+++ b/poetry/installation/chooser.py
@@ -82,7 +82,7 @@ class Chooser:
 
         return chosen
 
-    def _get_links(self, package):  # type: (Package) -> List[Link]
+    def _get_repository(self, package):
         if not package.source_type:
             if not self._pool.has_repository("pypi"):
                 repository = self._pool.repositories[0]
@@ -90,6 +90,10 @@ class Chooser:
                 repository = self._pool.repository("pypi")
         else:
             repository = self._pool.repository(package.source_reference)
+        return repository
+
+    def _get_links(self, package):  # type: (Package) -> List[Link]
+        repository = self._get_repository(package)
 
         links = repository.find_links_for_package(package)
 
--- a/poetry/installation/executor.py
+++ b/poetry/installation/executor.py
@@ -584,9 +584,10 @@ class Executor(object):
 
         archive = self._chef.get_cached_archive_for_link(link)
         if archive is link:
+            repository = self._chooser._get_repository(operation.package)
             # No cached distributions was found, so we download and prepare it
             try:
-                archive = self._download_archive(operation, link)
+                archive = self._download_archive(operation, link, repository)
             except BaseException:
                 cache_directory = self._chef.get_cache_directory_for_link(link)
                 cached_file = cache_directory.joinpath(link.filename)
@@ -611,9 +612,13 @@ class Executor(object):
 
         return archive
 
-    def _download_archive(self, operation, link):  # type: (Operation, Link) -> Path
+    def _download_archive(self, operation, link, repository):  # type: (Operation, Link, LegacyRepository) -> Path
         response = self._authenticator.request(
-            "get", link.url, stream=True, io=self._sections.get(id(operation), self._io)
+            "get",
+            link.url,
+            stream=True,
+            io=self._sections.get(id(operation), self._io),
+            verify=getattr(repository, 'cert'),
         )
         wheel_size = response.headers.get("content-length")
         operation_message = self.get_operation_message(operation)


poetry/installation/chooser.py
```diff
@@ -82,14 +82,18 @@ def choose_for(self, package):  # type: (Package) -> Link
...
             else:
                 repository = self._pool.repository("pypi")
         else:
             repository = self._pool.repository(package.source_reference)
+        return repository
+
+    def _get_links(self, package):  # type: (Package) -> List[Link]
+        repository = self._get_repository(package)
 
         links = repository.find_links_for_package(package)
```
intgr, line 98: What's the most appropriate way to solve this? Should I add a `repository` attribute to the `Link` class? 🤔

poetry/installation/executor.py
```diff
@@ -611,9 +612,13 @@ def _download_link(self, operation, link):
...
 
-    def _download_archive(self, operation, link):  # type: (Operation, Link) -> Path
+    def _download_archive(self, operation, link, repository):  # type: (Operation, Link, LegacyRepository) -> Path
         response = self._authenticator.request(
-            "get", link.url, stream=True, io=self._sections.get(id(operation), self._io)
+            "get",
+            link.url,
+            stream=True,
+            io=self._sections.get(id(operation), self._io),
+            verify=getattr(repository, 'cert'),
```
intgr, line 621: Only `LegacyRepository` has the `cert` attribute, should I have an `isinstance()` check or is `getattr()` good enough?

intgr:
Ping @finswimmer @stephsamson any feedback on whether I'm on the right track with this?

Spectre5:
I get this same error, it would be great to get this fixed!  That said, here is my work-around that is working, but it just seems unnecessary since this is the same `pem` file that I already provided to `poetry` via `poetry config certificates.$repo.cert` command.  A simple fix might be to just put this cert value into `os.environ` at the start of the `poetry`?

`CURL_CA_BUNDLE="path/to/cerl/ca-bundle.pem" poetry install –vvv`

neersighted:
This is obsolete as of the merge of #5719.

gazpachoking:
I don't believe this is obsolete. I'm still having the same problem in 1.2b3. If I specify a custom certificate for a private repository (or turn off certificate verification completely for that repository) everything goes fine util it tries to download the wheel. At that point I get a certificate verify failed error. It seems that the certificate.REPO.cert configuration setting applies when communicating with the api for a given repo, but not when downloading the wheels from that repo.

neersighted:
This PR still conflicts/won't resolve that -- please open an issue with so we can take a look.

PR #3350, opened 2020-11-10
Fixed missing space in env activation 'one-liner'

adnan-st:
`source`poetry env info --path`/bin/activate`
replaced with:
`source `poetry env info --path`/bin/activate`

Was resulting in error like:
```
`source`poetry env info --path`/bin/activate
zsh: no such file or directory: source/Users/Voltron/Library/Caches/pypoetry/virtualenvs/some-project-RqqT3xfp-py3.8/bin/activate`
```


finswimmer (approved):
Thanks :+1:

PR #3375, opened 2020-11-18
Account for symlinked .venv when removing

vibe:
# Pull Request Check List

Resolves: Sometimes the `.venv` is a symlink, such as in caching strategies for CI/CD environments (codebuild, etc)  
Related: https://github.com/python-poetry/poetry/pull/2064

We are already asserting the file path is a directory, so if OS Error 20 is triggered, which is "Not a directory" we simply delete everything but the top level folder.

- [ X] Added **tests** for changed code.
- [ X] Updated **documentation** for changed code.

### DIFF
--- a/poetry/utils/env.py
+++ b/poetry/utils/env.py
@@ -834,13 +834,16 @@ class EnvManager(object):
             return
         except OSError as e:
             # Continue only if e.errno == 16
-            if e.errno != 16:  # ERRNO 16: Device or resource busy
+            if (
+                e.errno != 16 and e.errno != 20
+            ):  # ERRNO 16: Device or resource busy, # ERRNO 20: Not a directory
                 raise e
 
         # Delete all files and folders but the toplevel one. This is because sometimes
-        # the venv folder is mounted by the OS, such as in a docker volume. In such
+        # the venv folder is mounted or symlinked by the OS, such as in a docker volume, codebuild, etc. In such
         # cases, an attempt to delete the folder itself will result in an `OSError`.
         # See https://github.com/python-poetry/poetry/pull/2064
+        # See https://bugs.python.org/issue1669
         for file_path in path.iterdir():
             if file_path.is_file() or file_path.is_symlink():
                 file_path.unlink()
--- a/tests/utils/test_env.py
+++ b/tests/utils/test_env.py
@@ -572,7 +572,9 @@ def test_remove_also_deactivates(tmp_dir, manager, poetry, config, mocker):
     assert venv_name not in envs
 
 
-def test_remove_keeps_dir_if_not_deleteable(tmp_dir, manager, poetry, config, mocker):
+def test_remove_keeps_dir_if_not_deleteable_os_err_16(
+    tmp_dir, manager, poetry, config, mocker
+):
     # Ensure we empty rather than delete folder if its is an active mount point.
     # See https://github.com/python-poetry/poetry/pull/2064
     config.merge({"virtualenvs": {"path": str(tmp_dir)}})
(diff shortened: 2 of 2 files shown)


vibe:
Hmm, I see a test failed for 3.6 where the error code number is 2 and not 0.


Gross.

Maybe it's worth it to make the symlink check explicit instead of checking the code number...

tests/utils/test_env.py
```diff
@@ -619,6 +621,55 @@ def err_on_rm_venv_only(path, *args, **kwargs):
...
+    mocker.patch(
+        "poetry.utils._compat.subprocess.check_output",
+        side_effect=check_output_wrapper(Version.parse("3.6.6")),
+    )
+
+    original_rmtree = shutil.rmtree
+
+    def err_on_rm_venv_only(path, *args, **kwargs):
+        if path == str(venv_path):
+            raise OSError(20, "Test error")  # ERRNO 16: Device or resource busy
```
neersighted, line 653: Looks like this comment is incorrect.

poetry/utils/env.py
```diff
@@ -834,13 +834,16 @@ def remove_venv(cls, path):  # type: (Union[Path,str]) -> None
...
             # Continue only if e.errno == 16
-            if e.errno != 16:  # ERRNO 16: Device or resource busy
+            if (
+                e.errno != 16 and e.errno != 20
+            ):  # ERRNO 16: Device or resource busy, # ERRNO 20: Not a directory
                 raise e
 
         # Delete all files and folders but the toplevel one. This is because sometimes
-        # the venv folder is mounted by the OS, such as in a docker volume. In such
+        # the venv folder is mounted or symlinked by the OS, such as in a docker volume, codebuild, etc. In such
```
neersighted, line 843: Please reformat this comment block to have a uniform width.

poetry/utils/env.py
```diff
@@ -834,13 +834,16 @@ def remove_venv(cls, path):  # type: (Union[Path,str]) -> None
             return
         except OSError as e:
             # Continue only if e.errno == 16
-            if e.errno != 16:  # ERRNO 16: Device or resource busy
+            if (
```
neersighted, line 837: I'd like to see the `errno` module's named constants used here. A construct like `e.errno in (errno.ENOTDIR, errno.EBUSY)` would be more obvious and maintainable.

poetry/utils/env.py
```diff
@@ -834,13 +834,16 @@ def remove_venv(cls, path):  # type: (Union[Path,str]) -> None
             return
         except OSError as e:
             # Continue only if e.errno == 16
```
neersighted, line 836: Out of date comment

poetry/utils/env.py
```diff
@@ -834,13 +834,16 @@ def remove_venv(cls, path):  # type: (Union[Path,str]) -> None
...
+                e.errno != 16 and e.errno != 20
+            ):  # ERRNO 16: Device or resource busy, # ERRNO 20: Not a directory
                 raise e
 
         # Delete all files and folders but the toplevel one. This is because sometimes
-        # the venv folder is mounted by the OS, such as in a docker volume. In such
+        # the venv folder is mounted or symlinked by the OS, such as in a docker volume, codebuild, etc. In such
         # cases, an attempt to delete the folder itself will result in an `OSError`.
         # See https://github.com/python-poetry/poetry/pull/2064
+        # See https://bugs.python.org/issue1669
```
neersighted, line 846: I don't think this issue is relevant -- it describes a bug that is long-resolved, and not the behavior of throwing an exception when called on a symlink.

tests/utils/test_env.py
```diff
@@ -619,6 +621,55 @@ def err_on_rm_venv_only(path, *args, **kwargs):
     m.side_effect = original_rmtree  # Avoid teardown using `err_on_rm_venv_only`
 
 
+def test_remove_keeps_dir_if_not_deleteable_os_err_20(
```
neersighted, line 624: Let's parameterize this whole test for different `errno` constants instead of doing this copy-and-paste.

neersighted (changes requested):

abn:
Stale. Closing as this has not had any activity for a while, and the error the change attempts to fix is no longer reproducible.

Issue #3362, opened 2020-11-13
'NoneType' object has no attribute '_source_type'


lazka:
Poetry 1.1.4 on Debian sid and Ubuntu 20.10:

* `poetry init`
* `poetry add pyyaml==5.3.1 -vvv`

```python
  AttributeError

  'NoneType' object has no attribute '_source_type'

  at ~/.local/lib/python3.8/site-packages/poetry/inspection/info.py:548 in from_directory
      544│                         raise
      545│ 
      546│                     # we discovered PkgInfo but no requirements were listed
      547│ 
    → 548│         info._source_type = "directory"
      549│         info._source_url = path.as_posix()
      550│ 
      551│         return info
      552│ 
```

lazka:
Downgrading to 1.0.10 makes things work again

finswimmer:
Hello @lazka,

I'm unable to reproduce it. So let's start with the default questions: How have you installed poetry? Can you show the output of `poetry config --list`?

fin swimmer

lazka:
```shell
> poetry config --list
cache-dir = "/home/lazka/.cache/pypoetry"
experimental.new-installer = true
installer.parallel = true
virtualenvs.create = true
virtualenvs.in-project = null
virtualenvs.path = "{cache-dir}/virtualenvs"  # /home/lazka/.cache/pypoetry/virtualenvs
```

lazka:
@finswimmer try clearing your poetry cache. If pyyaml got installed with an older poetry before then it also works with the newer one because it doesn't download the sdist.

lazka:
It seems to fail because requires_dist is None, but the package has no dependencies, so that seems.. expected?

lazka:
I've created #3381 as a start

lazka:
ok, turns out updating all dependencies of poetry fixes it somehow

agatti:
For what it's worth, I can reproduce this on macOS and Python 3.8 with [aiohttp](https://github.com/aio-libs/aiohttp) as a failng dependency - starting from a newly created project.  Using Poetry version 1.1.4, with just `poetry init` and `poetry add "aiohttp[speedups]"` I get the very same error. 

Just in case, my configuration looks like this:

```bash
❯ poetry config --list
cache-dir = "/Users/agatti/Library/Caches/pypoetry"
experimental.new-installer = true
installer.parallel = true
virtualenvs.create = true
virtualenvs.in-project = null
[...]

PR #3381, opened 2020-11-19
Fix installation of pyyaml

lazka:
"poetry add pyyaml==5.3.1" was failing with:
   AttributeError: 'NoneType' object has no attribute '_source_type'

pyyaml doesn't have any dependencies and various code in PackageInfo() assumes
no dependencies means looking for other sources. from_directory() didn't handle
the case where none of the sources return something and would fail like noted above.

Instead of failing use the last valid result and return it.

Resolves: #3362


lazka:
found a different solution, see #3362

lazka:
Closing in favor of #3614

PR #3384, opened 2020-11-19
console: refactor export command

abn:
This change refactors export command to use format specific sub-command instead of the previously used `--format` option.

Due to limitations of clikit/cleo, we manipulate `sys.argv` to detect deprecated usage and route command to replicate current behaviour.

We should, in the future (once cle tester supports it), move common options like `--output` to the top level command as global options.

Rationale: Options relating to export is typically format specific. Handling this using conditional logic becomes a bit painful. This allows for cleaner seperation and testing.

### DIFF
--- a/poetry/console/application.py
+++ b/poetry/console/application.py
@@ -109,6 +109,40 @@ class Application(BaseApplication):
 
         return commands
 
+    def handle_deprecated_commands(self):
+        import sys
+
+        if sys.argv and sys.argv[1] == "export":
+            # TODO: Remove in 1.3 release
+            if "--help" not in sys.argv and "-h" not in sys.argv:
+                for export_sub_command in ExportCommand.commands:
+                    if export_sub_command.name in sys.argv:
+                        break
+                else:
+                    sys.argv.insert(2, "requirements.txt")
+                    message = (
+                        "\n"
+                        "The top level <info>export</info> command and the <info>--format</info> option has been "
+                        "deprecated. It is recommended that the use which be replaced with format specific "
+                        "sub-commands.\n\n"
+                        ""
+                        "Deprecated usage:\n"
+                        "<fg=red>poetry export --format requirements.txt --without-hashes</>\n"
+                        "<fg=red>poetry export --without-hashes</>\n"
+                        "\n"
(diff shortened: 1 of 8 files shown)


abn:
@sdispater Not sure if there is better way to do this with cleo; couple of issues I had to workaround were

1. Lack of a mechanism to specify a default fallback sub-command passing through the current options without raising a `NoSuchOption`. See workaround in `Application` implementation.
2. Lack of support for global options for subcommands when used with command tester. The workaround here was to move the options from global to sub-commands themselves (eg: `--output`).

poetry/console/application.py
```diff
@@ -109,6 +109,40 @@ def get_default_commands(self):  # type: () -> list
...
+                        "Deprecated usage:\n"
+                        "<fg=red>poetry export --format requirements.txt --without-hashes</>\n"
+                        "<fg=red>poetry export --without-hashes</>\n"
+                        "\n"
+                        "Updated usage:\n"
+                        "<info>poetry export requirements.txt --without-hashes</info>"
+                    )
+                    self._preliminary_io.error_line(
+                        "<fg=yellow>{}</>\n".format(message)
+                    )
```
Leonardofreua, line 138:
```suggestion
                    self._preliminary_io.error_line(f"<fg=yellow>{message}</>\n")
```

tests/exporter/test_exporter_requirements_txt.py
```diff
@@ -105,9 +105,9 @@ def test_exporter_can_export_requirements_txt_with_standard_packages(
     )
     set_package_requires(poetry)
 
-    exporter = Exporter(poetry)
+    exporter = RequirementsTxtExporter(poetry)
 
-    exporter.export("requirements.txt", Path(tmp_dir), "requirements.txt")
+    exporter.export(Path(tmp_dir), "requirements.txt")
```
Leonardofreua, line 110: Consider putting the `requirements.txt` string into a constant.

Leonardofreua (changes requested):

abn:
This is no longer relevant here as the export command is being moved into the `poetry-plugin-export` package.

PR #3389, opened 2020-11-19
document ssh connection for git dependencies

torstenrudolf:
# Pull Request Check List

it was not documented how to use ssh connection for git dependencies. I found various issues on this and it took me some time to find the correct syntax. this PR adds it to the doc.

### DIFF
--- a/docs/docs/dependency-specification.md
+++ b/docs/docs/dependency-specification.md
@@ -98,6 +98,12 @@ flask = { git = "https://github.com/pallets/flask.git", rev = "38eb5d3b" }
 numpy = { git = "https://github.com/numpy/numpy.git", tag = "v0.13.2" }
 ```
 
+To use ssh connection (for example for private reopsitories) simply follow the standard git syntax:
+```toml
+[tool.poetry.dependencies]
+requests = { git = "git@github.com:requests/requests.git" }
+```
+
 ## `path` dependencies
 
 To depend on a library located in a local directory or file,
--- a/docs/docs/dependency-specification.md
+++ b/docs/docs/dependency-specification.md
@@ -98,7 +98,7 @@ flask = { git = "https://github.com/pallets/flask.git", rev = "38eb5d3b" }
 numpy = { git = "https://github.com/numpy/numpy.git", tag = "v0.13.2" }
 ```
 
-To use ssh connection (for example for private reopsitories) simply follow the standard git syntax:
+To use an SSH connection,  for example in the case of  private repositories, simply follow the standard git syntax:
 ```toml
 [tool.poetry.dependencies]
 requests = { git = "git@github.com:requests/requests.git" }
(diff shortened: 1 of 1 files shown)


docs/docs/dependency-specification.md
```diff
@@ -98,6 +98,12 @@ flask = { git = "https://github.com/pallets/flask.git", rev = "38eb5d3b" }
 numpy = { git = "https://github.com/numpy/numpy.git", tag = "v0.13.2" }
 ```
 
+To use ssh connection (for example for private reopsitories) simply follow the standard git syntax:
```
stephsamson, line 101:
```suggestion
To use an SSH connection, for e.g. in the case of private repositories, simply follow the standard git syntax:
```

torstenrudolf: thank you. good improvement, only will slightly change `for e.g.`, since `e.g.` translates to `for example`.

```suggestion
To use an SSH connection,  for example in the case of  private repositories, simply follow the standard git syntax:
```

docs/docs/dependency-specification.md
```diff
@@ -98,6 +98,12 @@ flask = { git = "https://github.com/pallets/flask.git", rev = "38eb5d3b" }
 numpy = { git = "https://github.com/numpy/numpy.git", tag = "v0.13.2" }
 ```
 
+To use an SSH connection,  for example in the case of  private repositories, simply follow the standard git syntax:
```
torstenrudolf, line 101:
```suggestion
To use an SSH connection, for example in the case of private repositories, simply follow the standard git syntax:
```

justinmayer:
Isn't the protocol shown in this PR the Git protocol? That is, _not_ SSH? https://git-scm.com/book/en/v2/Git-on-the-Server-The-Protocols

stephsamson:
@justinmayer thanks for the link! Based on that, it seems like it is still an SSH protocol "using the shorter scp-like syntax for the SSH protocol":
```
$ git clone [user@]server:project.git
```

Let me know if I misunderstood, thanks!

docs/docs/dependency-specification.md
```diff
@@ -98,6 +98,12 @@ flask = { git = "https://github.com/pallets/flask.git", rev = "38eb5d3b" }
 numpy = { git = "https://github.com/numpy/numpy.git", tag = "v0.13.2" }
 ```
 
+To use an SSH connection, for example in the case of private repositories, simply follow the standard git syntax:
```
justinmayer, line 101: May I suggest a slightly-different phrasing here?
> To use an SSH connection, for example in the case of private repositories, use the following example syntax:

justinmayer:
@stephsamson: It would seem that you are correct. I had assumed that leaving off the protocol identifier would default to the Git protocol, when in fact it seems to default to using the shorter SCP-like syntax for the SSH protocol.

I made a minor suggested change to the accompanying text. What do you think?

docs/docs/dependency-specification.md
stephsamson: This more general phrase makes sense! Thank you 💫

stephsamson (approved):

PR #3396, opened 2020-11-21
Support Python 3.9 on FreeBSD

neersighted:
Just a quick change since Python 3.9 has been in the repos for 5 weeks now.


neersighted:
Updated this to be a patch series with a closely related set of fixes/improvements. They are separate commits so we can easily reject one or more if preferred. Still, I think making FreeBSD use the same release pattern that Github Actions has migrated to makes the most sense.

neersighted:
It looks like this is no longer necessary in light of the move to the `install-poetry.py` script. Awesome!

PR #3411, opened 2020-11-24
Allow Wheel to work when no version is specified in filename

dhiltonp:
I hit a problem where poetry wasn't handling a package correctly.

It turns out that the wheel_file_re pattern correctly treats `version` as optional, but `Wheel` wasn't.

I've updated Wheel to properly handle this case and added a test using the exact package name that tripped me up.


dhiltonp:
@finswimmer's fix in #3338 is the better fix, I made a mistake when examining the regex and thought the current regex was working correctly.

Issue #2422, opened 2020-05-14
poetry.lock hashes not checked when running poetry install


slai:
- [x] I am on the [latest](https://github.com/python-poetry/poetry/releases/latest) Poetry version.
- [x] I have searched the [issues](https://github.com/python-poetry/poetry/issues) of this repo and believe that this is not a duplicate.
- [ ] If an exception occurs when executing a command, I executed it again in debug mode (`-vvv` option).

- **OS version and name**: Ubuntu Bionic (within Docker)
- **Poetry version**: 1.0.5
- **Link of a [Gist](https://gist.github.com/) with the contents of your pyproject.toml file**: https://gist.github.com/slai/9d0d442fe7e4f4ea04e8f658b675192a#file-poetry_no_hash_check-dockerfile-L22-L33

## Issue

It appears that Poetry does not check the hashes of the packages in poetry.lock when installing packages using `poetry install`. 

The Dockerfile in this gist (https://gist.github.com/slai/9d0d442fe7e4f4ea04e8f658b675192a) demonstrates the issue. It runs `poetry lock`, mangles the hashes in poetry.lock, then runs `poetry install`. This succeeds, with no errors.

I would expect `poetry install` to fail in this case, or at least print a warning for packages where the hash does not match.

signed8bit:
It's a bit concerning that this issue has been open for nearly a year now. Is this something the core team is looking to address?

FlorianLudwig:
It doesn't seem so. I opened a new MR here: https://github.com/python-poetry/poetry-core/pull/113

Since the code moved into a different repo.

signed8bit:
Thanks for the pointer @FlorianLudwig!

pietrodn:
List of PRs related to this issue:

- https://github.com/python-poetry/poetry-core/pull/113 (needs a rebase)
- https://github.com/python-poetry/poetry-core/pull/159
- https://github.com/python-poetry/poetry/pull/3885
- https://github.com/python-poetry/poetry/pull/3423
- https://github.com/python-poetry/poetry/pull/3301 (needs a rebase)
- https://github.com/python-poetry/poetry/pull/2611 (needs a rebase)

Can we merge one of the valid ones? Not checking the hashes in `poetry.lock` poses a significant security problem.

pietrodn:
In particular python-poetry/poetry-core#159 + #3885 are meant to be fixed together from what I understand. The other PRs can be closed.

remram44:
[...]

PR #3423, opened 2020-11-27
Cleanup hash checking

FlorianLudwig:
# Pull Request Check List

Depends on https://github.com/python-poetry/poetry-core/pull/113
Further improves: #2422 

# What is this about

There are two points where checksum are (supposed to be*) validated in poetry:

1. When choosing a download link the hash sum is queried from the repo and checked against a list of known hashes
2. After downloading the downloaded package is checked to have one of the known hashes

This MR changes a few things:

1. Only check against the hash for the corresponding file name instead of all hashes - this improves security and makes it possible to improve output to the user.
2. Improve output: I the current version* if the hash mismatches this could be for two reasons, the hash is for a different file (e.g. the lock file was created for the .tar.gz but it was deleted and a wheel was uploaded instead) or because the hash of the file does not match. These are two different scenarios which the user cannot differentiate. This MR makes it transparent for the user what is actually happening.
3. Code Clean-up: When selecting which link to use for download, if a matching hash exists for the link is considered. But the code filters out all links with mismatching hashes before the sorting is even reached. So that part was removed.

```
* The current version does not validate hashes at all. I am referring to how the code is intended to do it.
```

# Open TODOs
1. This really needs proper tests - and I am willing to contribute them but I had some difficulties, [see other MR](https://github.com/python-poetry/poetry-core/pull/113)
2. I am not sure about how to properly handle the error/warning output when a mismatching hash occurs. I python logging - which I guess is okay but not as nice to look at as the usual poetry output. On the other hand, mismatching hashes means something ugly is going on... :man_shrugging: 
3. Docs need to be added - but I would welcome input on code and CLI output before updating the docs.

### DIFF
--- a/poetry/installation/chooser.py
+++ b/poetry/installation/chooser.py
@@ -95,24 +95,7 @@ class Chooser:
             repository = self._pool.repository(package.source_reference)
 
         links = repository.find_links_for_package(package)
-
-        hashes = [f["hash"] for f in package.files]
-        if not hashes:
-            return links
-
-        selected_links = []
-        for link in links:
-            if not link.hash:
-                selected_links.append(link)
-                continue
-
-            h = link.hash_name + ":" + link.hash
-            if h not in hashes:
-                continue
-
-            selected_links.append(link)
-
-        return selected_links
+        return [link for link in links if self._is_link_hash_allowed_for_package(link, package)]
 
     def _sort_key(self, package, link):  # type: (Package, Link) -> Tuple
         """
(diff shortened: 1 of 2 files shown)


JonZeolla:
@FlorianLudwig can you elaborate on the current hash validation gap?  Wondering if there's a separate issue for that

FlorianLudwig:
@JonZeolla the current gap is not a gap, in the current version hashes are not validated at all.

JonZeolla:
Thank you for the details

FlorianLudwig:
@JonZeolla also see https://github.com/python-poetry/poetry/issues/2422

neersighted:
Feel free to correct me if I'm wrong, but I'm reasonably sure that #4740 obsoletes this PR for the most part. Pointing out follow-up work that is not included in that PR but is in this one would be appreciated.

I'm currently in the process of working out some more test coverage for that PR -- it will be ready soon.

PR #1181, opened 2019-06-20
Enhancement: add "extras" example to scripts docs

bradsbrown:
After digging around in the poetry code to figure out
how to include an `extras` key in a cli script,
I thought it might be a useful thing to add to the docs
for future users.

# Pull Request Check List

This is just a reminder about the most common mistakes. Please make sure that you tick all *appropriate* boxes.  But please read our [contribution guide](https://poetry.eustace.io/docs/contributing/) at least once, it will save you unnecessary review cycles!

- [n/a] Added **tests** for changed code.
- [x] Updated **documentation** for changed code.

**Note**: If your Pull Request introduces a new feature or changes the current behavior, it should be based
on the `develop` branch. If it's a bug fix or only a documentation update, it should be based on the `master` branch.

If you have *any* questions to *any* of the points above, just **submit and ask**!  This checklist is here to *help* you, not to deter you from contributing!

PR #3431, opened 2020-11-30
Fix script entrypoints with extras

akeeman:

---

- [x] Resolve invalid script entry point file when script has extras
- [x] Don't add script file when not all extras are selected

---
Consider this `pyproject.toml` file, which contains a script that has an extra (as implemented, but not yet documented -> https://github.com/python-poetry/poetry/pull/1181)
```toml
# pyproject.toml
# ...

[tool.poetry.extras]
my_extra = []

[tool.poetry.scripts]
foo = {callable = "my_package:foo", extras = ["my_extra"]}
```

Installing the above with `poetry install` givens you a script file including the 

```python
#!/path/to/python
from my_package import foo

if __name__ == '__main__':
    foo[my_extra]()
```

The last line is obviously incorrect. It should have been
```python
    foo()
```

In my opinion the file shouldn't have been installed at all in the first place, as those extras often go along with installing some optional packages to let them work. So I chose to skip the installation of the entrypoint as a whole if not all given extras are demanded. Note that I say "in my opinion", as this is different from setuptools. Setuptools will always install all script entrypoints.

### DIFF
--- a/poetry/console/commands/install.py
+++ b/poetry/console/commands/install.py
@@ -102,6 +102,8 @@ dependencies and not including the current project, run the command with the
             self.line("")
             return 0
 
+        builder.extras(extras)
+
         builder.build()
 
         if self._io.supports_ansi() and not self.io.is_debug():
--- a/poetry/masonry/builders/editable.py
+++ b/poetry/masonry/builders/editable.py
@@ -36,6 +36,8 @@ class EditableBuilder(Builder):
         self._env = env
         self._io = io
 
+        self._extras = []
+
     def build(self):
         self._debug(
             "  - Building package <c1>{}</c1> in <info>editable</info> mode".format(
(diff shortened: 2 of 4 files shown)


abn:
This will need to move to https://github.com/python-poetry/poetry-core as the relevant logic is now there.

akeeman:
So, no fixes for v1.1?

abn:
@akeeman apologies. I groked the issue incorrectly. Reopening.

akeeman:
That explains a lot! Thanks.

akeeman:
Hi, van this pr be handled some time? It should be an easy fix for a simple but annoying problem. Yet it's open for some time...

finswimmer:
Looks like this is fixed in the current preview release (1.2.0b2). Can you confirm it?

akeeman:
1.2.0b2 indeed does not have the issue. Is there any news on when it'll come out of beta?

Food for an other issue: 1.2.0b2 does install the executable even when the extra was not asked for.

What's new for me is that `poetry run my-command` works fine in 1.1.13, because it doesn't seem to call the generated executable. It can act as a workaround.

neersighted:
> 1.2.0b2 indeed does not have the issue. Is there any news on when it'll come out of beta? 1.2.0b2 does install the executable even when the extra was not asked for.

https://github.com/python-poetry/poetry/issues/5586

Is it worth rebasing this PR to not install entrypoints when the required extra is not installed? I understand we're not doing anything /wrong/ currently because that's the pip behavior, but I'm open to being convinced that not installing entrypoints is beneficial to the user.

akeeman:
I guess it would be better to do that in a separate pr pr, so let's not do that here.

PR #3457, opened 2020-12-06
tests: Fix incompatibility with pytest-mock>=3.3.0

seppeljordan:
Version 3.3.0 of pytest mock renamed MockFixture into MockerFixture. There is
as alias available for backwards compatibility. The import path of this alias
is not pytest_mock.pluging though but just pytest_mock instead.

For reference: Check out this [diff between version 3.2.0 and 3.3.0 of pytest-mock](https://github.com/pytest-dev/pytest-mock/compare/v3.2.0...v3.3.0).  Unfortunately you have to expand the diff of `src/pytest_mock/plugin.py` to see the change. This makes it impossible to link to the appropriate line (which is line 43 in the newer version).

# Pull Request Check List

Resolves: #issue-number-here
I did not find any issues related to this as of creating this PR.

I think non of the above check boxes apply.

### DIFF
--- a/tests/repositories/test_installed_repository.py
+++ b/tests/repositories/test_installed_repository.py
@@ -3,7 +3,7 @@ from typing import Optional
 
 import pytest
 
-from pytest_mock.plugin import MockFixture
+from pytest_mock import MockFixture
 
 from poetry.core.packages import Package
 from poetry.repositories.installed_repository import InstalledRepository


seppeljordan:
Hi @finswimmer @abn @stephsamson @sdispater 

I would like to close this PR since it is already open since 1 month and i got no reaction. That is totally okay though. I would just like to know if this fix is not wanted or if there might be another reason, why nobody looked into this (e.g. no capacity).

Cheers

sinoroc:
I would venture: no capacity. They look at issues and PRs when time permits, which these days amounts to a couple of issues/PRs per week. Patience...
_[Could take months. I guess at some point, there might be a sprint of some sort and they'll review everything all at once.]_

seppeljordan:
Okay, I am sick of this issue sitting in my inbox. Who ever wants to re submit this as their own, feel free to use my code.

PR #3460, opened 2020-12-07
update for #3338 - one digit version in filename

dhiltonp:
This resolves the comment left in #3338: https://github.com/python-poetry/poetry/pull/3338#discussion_r526512536


Secrus:
This code is already in the project, so PR is no longer valid. Closing.

Issue #2643, opened 2020-07-07
Poetry 1.1.0a2 Breaks Support for File Dependencies on Networked Drives


KyleKing:
- [X] I am on the [latest](https://github.com/python-poetry/poetry/releases/latest) {*pre-release} Poetry version.
- [X] I have searched the [issues](https://github.com/python-poetry/poetry/issues) of this repo and believe that this is not a duplicate.
- [X] If an exception occurs when executing a command, I executed it again in debug mode (`-vvv` option).

- **OS version and name**: `Win10 1909`
- **Poetry version**:  `1.1.0a2`
- **Link of a [Gist](https://gist.github.com/) with the contents of your pyproject.toml file**: `See example toml snippet at bottom`

## Issue

`Poetry 1.1.0a2` attempts to determine the relative path to a file dependency, but this fails (`os.path.relpath` raises a ValueError) if the dependency is not on the same drive. In my use case, we deploy wheel/tar files to a folder on a networked drive (`\\company.com\dfsroot\etc...`) so that the files are accessible to all internal users.

`Poetry 1.1.0a1` does not have this issue, so I have downgraded to 1.1.0a1 for now

https://github.com/python-poetry/poetry/blob/feef53be6043c604ae0ea96f3b02ca9748091b8a/poetry/puzzle/provider.py#L680-L690

### Logs

`poetry update -vvv`

```
Using virtualenv: C:\Users\me\project_b\.venv
Updating dependencies
Resolving dependencies...
   1: fact: project_b is 0.0.0
   1: derived: project_b
   1: fact: project_b depends on project_a (0.1.1 \\company.com\dfsroot\project_a-0.1.1-py3-none-any.whl)
   1: selecting project_b (0.0.0)
   1: derived: project_a (0.1.1 \\company.com\dfsroot\project_a-0.1.1-py3-none-any.whl)

   1: Version solving took 0.540 seconds.
   1: Tried 1 solutions.

  Stack trace:

  13  ~\.poetry\lib\poetry\_vendor\py3.7\clikit\console_application.py:131 in run
       129│             parsed_args = resolved_command.args
       130│ 
     → 131│             status_code = command.handle(parsed_args, io)
       132│         except KeyboardInterrupt:
       133│             status_code = 1

  12  ~\.poetry\lib\poetry\_vendor\py3.7\clikit\api\command\command.py:120 in handle
       118│     def handle(self, args, io):  # type: (Args, IO) -> int
       119│         try:
[...]

PR #3466, opened 2020-12-09
Fix regression for file dependencies on networked drives

KyleKing:
## Summary

Starting in `Poetry 1.1.0a2`, the step that formats the relative path for the lock file doesn't properly handle networked or other external drives which should lock the user-specified absolute path. This PR fixes that regression and adds tests for the various combinations of file or directory dependency paths

I would be happy to make any improvements or changes or answer any questions to help get this merged, thanks for reviewing!

## Pull Request Check List

Resolves: #2643 

- [X] Updated **documentation** for changed code. {N/A}

### DIFF
--- a/poetry/packages/locker.py
+++ b/poetry/packages/locker.py
@@ -561,12 +561,11 @@ class Locker(object):
         if package.source_url:
             url = package.source_url
             if package.source_type in ["file", "directory"]:
-                # The lock file should only store paths relative to the root project
-                url = Path(
-                    os.path.relpath(
-                        Path(url).as_posix(), self._lock.path.parent.as_posix()
-                    )
-                ).as_posix()
+                # For all local paths, the lock file should store the relative path from the root folder
+                url = Path(url).as_posix()
+                package_root = self._lock.path.parent
+                if url.startswith(".") or Path(url).anchor == package_root.anchor:
+                    url = Path(os.path.relpath(url, package_root.as_posix())).as_posix()
 
             data["source"] = dict()
 
--- a/tests/packages/test_locker.py
+++ b/tests/packages/test_locker.py
@@ -529,3 +529,34 @@ content-hash = "c3d07fca33fba542ef2b2a4d75bf5b48d892d21a830e2ad9c952ba5123a52f77
     _ = locker.lock_data
 
     assert 0 == len(caplog.records)
+
+
+@pytest.mark.parametrize(
+    "source_type, source_url, expected_url",
+    [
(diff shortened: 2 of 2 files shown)


KyleKing:
Just following up. Is there someone who could review this? If so, I think this should be pretty straightforward. We have been using this fork internally and I added tests accordingly

KyleKing:
Following up again. This is a pretty straightforward change and I would really appreciate feedback!

I would be happy to make any requested changes

poetry/packages/locker.py
```diff
@@ -561,12 +561,11 @@ def _dump_package(self, package):  # type: (Package) -> dict
...
-                # The lock file should only store paths relative to the root project
-                url = Path(
-                    os.path.relpath(
-                        Path(url).as_posix(), self._lock.path.parent.as_posix()
-                    )
-                ).as_posix()
+                # For all local paths, the lock file should store the relative path from the root folder
+                url = Path(url).as_posix()
+                package_root = self._lock.path.parent
+                if url.startswith(".") or Path(url).anchor == package_root.anchor:
```
abn, line 567: Should this be using an "is absolute" check instead? Additionally, I am not sure if this is the right fix here. Am I correct in assuming that the intention here is to not modify the path if is a network drive?

KyleKing: Hi @abn, thanks for reviewing

My intention was to preserve the original logic of storing the relative path when possible. This would apply only to paths that were relative already or had the same root. This passes absolute paths not on the local file system unmodified

Using `.is_absolute()` would be true for both a `C:\Path` and a `Z:\Path` but the latter cannot be a relative path because it is on a different drive. In the latter case, the path should properly be configured by the user and passed through

Let me know if there are other test cases to add or you would like to see a different approach

I think there could be an argument to treat the toml file as the source of truth for file paths where the locker should use the user's path without modification. Even if the shortest relative path is calculated for the lock file, the toml would still have a specific path that should have been relative

Alternatively, this could be implemented with a try/except rather than a "look before you leap" approach

iamgp:
Great work KyleKing. I've just run into a situation which this PR would fix. Would love this to have some more attention.

poetry/packages/locker.py
KyleKing: @abn / @finswimmer anything I can do to help move this change along?

abn: Apologies for not getting to this. Will get to it later today.

KyleKing:
It was easier to self-host a repo then implement this minor fix to poetry 😞

PR #3470, opened 2020-12-09
Fix #3434: poetry init interactive dependency search is case sensitive

### LINKED ISSUE #3434: poetry init - doesn't find or install a given dependency, even with the exact name

- **OS version and name**: Mac OS 10.13
- **Poetry version**: 1.1.4
- **Link of a [Gist](https://gist.github.com/) with the contents of your pyproject.toml file**: irrelevant

## Issue

Trying to create a new project with `poetry init`, I ask to add dependencies interactively and I get:

```
Search for package to add (or leave blank to continue): PyGithub
Found 20 packages matching PyGithub

Enter package # to add, or the complete package name if it is not listed:
 [0] pygithub33
 [1] pygithub3
 [2] pygithub3-intellisense
 [3] PyGithub-ReadOnly
 [4] pygithub-redux
 [5] PyGithub-requests
 [6] pygit-annek
 [7] gitshoes
 [8] pygithubutils
 [9] poebot
 > PyGithub
Value "PyGithub" is invalid
```

There is a `PyGithub` on PyPI: <https://pypi.org/project/PyGithub/> so I guess the behaviour is wrong.

1. PyGithub should be the first match (as it's the best match).
2. It should be possible to add PyGithub in anycase by forcing the package name.

glencairn:
# Pull Request Check List

Resolves: #3434

---

## Problem(s)

This PR fixes an issue with the `poetry init` command, more specifically the interactive dependency specification feature. When comparing a given package name against the matches in the pool, the names are considered to be an exact match if only if the casing matches. This behavior deviates from [PEP-426](https://www.python.org/dev/peps/pep-0426/#name) which states that comparisons of distribution names **must** be case insensitive.

This PR also removes some weird behavior where a case-insensitive match was purposely being _excluded_ from the list of package choices presented to the user. I assume that was to limit an exact match from appearing in the list twice. 🤷 

## Solution

I modified the exact match check to be case-insensitive.

## Testing

Added a pretty straight-forward test, `test_interactive_with_dependencies_case_sensitivity`, to cover the original issue. You can run the test against the first commit to see the failure and re-run the test against the head of this branch to see it pass. All existing tests continue to pass.

## Other Issue

In order to properly test this, I needed to modify the `search` method of the `Repository` class to also be case-insensitive. This is my first code contribution to `poetry` so I'm not sure of what the side-effects of that might be, if there are any. Any insight would be helpful!

### DIFF
--- a/poetry/console/commands/init.py
+++ b/poetry/console/commands/init.py
@@ -255,18 +255,15 @@ The <c1>init</c1> command creates a basic <comment>pyproject.toml</> file in the
                 else:
                     choices = []
                     matches_names = [p.name for p in matches]
-                    exact_match = constraint["name"] in matches_names
+                    exact_match = constraint["name"].lower() in matches_names
                     if exact_match:
                         choices.append(
-                            matches[matches_names.index(constraint["name"])].pretty_name
+                            matches[matches_names.index(constraint["name"].lower())].pretty_name
                         )
 
                     for found_package in matches:
                         if len(choices) >= 10:
                             break
-
-                        if found_package.name.lower() == constraint["name"].lower():
-                            continue
 
                         choices.append(found_package.pretty_name)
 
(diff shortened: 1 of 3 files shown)


glencairn:
For whatever reason the FreeBSD tests weren't run...

image

Secrus:
The issue was fixed in the meantime. Closing

Issue #3463, opened 2020-12-08
Wrong python version in new installation


ricardo-reis-1970:
- [x] I am on the [latest](https://github.com/python-poetry/poetry/releases/latest) Poetry version.
- [x] I have searched the [issues](https://github.com/python-poetry/poetry/issues) of this repo and believe that this is not a duplicate.
- [x] If an exception occurs when executing a command, I executed it again in debug mode (`-vvv` option).

- **OS version and name**: 
- **Poetry version**: 
- **Link of a [Gist](https://gist.github.com/) with the contents of your pyproject.toml file**: 

## Issue

I have two version of Python:
- 3.9.0
- 3.6.6

I'm currently developing a project that is somewhat locked to 3.6.6, but also running experiments on latest Python, so I need both. I installed 3.6.6 first, so when I run `where python`, I get
```
C:\Program Files\Python39\python.exe
C:\Program Files\Python36\python.exe
```
This means that simply running `python` will run 3.9.0, as it is ahead in the `PATH`.

Having said all this, 3.6.6 is still available and here's the very top of my `pyproject.toml`:
```toml
[tool.poetry.dependencies]
python = "^3.6.6"
```
With this, I was expecting to have `poetry` to create the virtual environment correctly, but it does not. It simply goes to the first `python.exe` it can find &mdash; 3.9.0 &mdash; and then nothing will run.

Workarounds:
- uninstall Python 3.9.0 whenever I need to recreate the `.venv`: highly inconvenient;
- create the virtual environment by hand: I am using poetry to do that.

**How can I enforce the Python version? Am I doing something wrong?**

This is not a current duplicate, but I came across complaints online from 2 years ago along the same lines, and these were supposed to have been fixed long ago.

sinoroc:
As far as I can tell `^3.6.6` is equivalent to `>=3.6.6 <4.0.0`, so `3.90` is a valid Python version:

* https://python-poetry.org/docs/dependency-specification/#caret-requirements

So this could be normal expected behaviour.

ricardo-reis-1970:
Dear @sinoroc,

Thank you for your reply. Indeed that is the case and you _would_ be on the spot. Unfortunately &mdash; my bad, I forgot to mention &mdash; I obviously tried removing the caret and "freezing" that specific version, to no avail. That was indeed the very reason I decided to open this ticket.

[...]

PR #3475, opened 2020-12-10
Fix fallback to other interpreters when specifying a patchlevel.

jaharkes:
When a project specifies a narrow range for the supported python
versions, i.e. "python ~= 3.6.1" and the current interpreter falls
outside of that range, the fallback code fails to test an available
python3.6 interpreter because of a too strict a test.

# Pull Request Check List

Resolves: #3463

### DIFF
--- a/src/poetry/utils/env.py
+++ b/src/poetry/utils/env.py
@@ -847,7 +847,7 @@ class EnvManager:
                         supported_python
                     ):
                         continue
-                elif not supported_python.allows_all(
+                elif not supported_python.allows_any(
                     parse_constraint(python_to_try + ".*")
                 ):
                     continue
--- a/tests/utils/test_env.py
+++ b/tests/utils/test_env.py
@@ -1093,3 +1093,35 @@ def test_env_finds_fallback_executables_for_generic_env(tmp_dir, manager):
 
     assert Path(venv.python).name == expected_executable
     assert Path(venv.pip).name == expected_pip_executable
+
+
+def test_create_venv_accepts_fallback_version_w_nonzero_patchlevel(
+    manager, poetry, config, mocker, config_virtualenvs_path
+):
+    if "VIRTUAL_ENV" in os.environ:
+        del os.environ["VIRTUAL_ENV"]
+
+    poetry.package.python_versions = "~3.5.1"
+    venv_name = manager.generate_env_name("simple-project", str(poetry.file.parent))
+
+    check_output = mocker.patch(
+        "subprocess.check_output",
+        side_effect=lambda cmd, *args, **kwargs: str(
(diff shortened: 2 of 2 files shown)


jaharkes:
I guess I should elaborate, I hit this when I upgraded my Ubuntu system from 18.04 to 20.04.

The upgrade switched the default python interpreter to 3.8 and the project I was working on had a dependency that was only available for python<3.8. And to add to the fun, another dependency claimed to only support python>=3.6.1.

So I installed python-3.6.12 from the deadsnakes ppa and switched the dependency from `python = "^3.6.1"` to `python = "~3.6.1"`. But poetry complained it didn't find a compatible version. However when I specify `"~3.6"` it does find the installed python3.6 binary, but then complains later on that the "current project's Python requirement (>=3.6,<3.7) is not compatible with some of the required packages Python requirements".

Some debugging later, I found that when it is looking for the version of installed python interpreters there is an optimization that skips running interpreters that do not fall in the requested range of supported python versions. And, as is, the test runs `if not Version(">=3.6.1,<3.7").allows_all("3.6.*")` which evaluated to false and so it never bothers to even check the installed python-3.6.12.

jaharkes:
I'm just rebasing this PR periodically and make sure all the tests still work locally with tox.

Right now I'm locally getting a type error from the python3.7 tests, which is also present in the upstream version without my patch. Also for some reason sonarcloud seems to be complaining about code duplication in test_env.py, but in a part of the file unrelated to the modifications from this patch.

The PR has been waiting for review since December, is there anything I can do to help it make progress?

jaharkes:
ping @python-poetry/triage

In my environment, I have multiple python version installed locally, The default python3 interpreter is 3.8.10 but I also have python-3.6.15. The following pyproject.toml is unable to find the 3.6.15 interpreter without this patch.
 
```toml
[tool.poetry]
name = "python_version"
version = "0.1.0"
description = ""
authors = ["test <test@example.org>"]

[tool.poetry.dependencies]
python = ">=3.6.5,<3.7"

[build-system]
```

Without patch
```sh
$ poetry shell
The currently activated Python version 3.8.10 is not supported by the project (>=3.6.5,<3.7).

  NoCompatiblePythonVersionFound

  Poetry was unable to find a compatible version. If you have one, you can explicitly use it via the "env use" command.
```

With patch
```sh
$ poetry shell
The currently activated Python version 3.8.10 is not supported by the project (>=3.6.5,<3.7).
Using python3.6 (3.6.15)
Creating virtualenv python-version-avkAwIfu-py3.6 in /home/.../.cache/pypoetry/virtualenvs
Spawning shell within /home/.../.cache/pypoetry/virtualenvs/python-version-avkAwIfu-py3.6
```

neersighted (approved):

Issue #3476, opened 2020-12-11
repository credential resolution from config is ambiguous


torque:
- [x] I am on the [latest](https://github.com/python-poetry/poetry/releases/latest) Poetry version.
- [x] I have searched the [issues](https://github.com/python-poetry/poetry/issues) of this repo and believe that this is not a duplicate.
- [x] If an exception occurs when executing a command, I executed it again in debug mode (`-vvv` option).

- **Poetry 1.1.4**

## Issue
As a practical example, gitlab provides public/private package registries on projects which follow the general url form:
`https://gitlab.com/api/v4/projects/<project id>/packages/pypi/simple`.

Let's say a user is part of two separate (private) projects on gitlab, **foo** and **bar**, and they configure poetry on their system as follows:

```sh
poetry config repositories.foo-pypi "https://gitlab.com/api/v4/projects/foo/packages/pypi/simple"
poetry config http-basic.foo-pypi "foo-username" "foo-password"
poetry config repositories.bar-pypi "https://gitlab.com/api/v4/projects/bar/packages/pypi/simple"
poetry config http-basic.bar-pypi "bar-username" "bar-password"
```

In project **baz** they have the following dependency

```toml
[tool.poetry.dependencies]
bar-library = {version = "^1.0.0", source = "bar-pypi"}
```

When poetry goes to look up `bar-library`, it will try to use the `foo-username` and `foo-password` credentials, which will fail as these are the wrong credentials (this is somewhat heinous in gitlab's case, as it just replies with a 404 on bad credentials, I believe to avoid leaking information about private resources. The 404 just looks like a missing package to poetry rather than a credential error, so it ends up erroring in the solver).

The reason the wrong credentials are used is in `poetry/installation/authenticator.py`, specifically in `Authenticator._get_credentials_for_netloc_from_config`. The problem is that poetry does a credential lookup based on the `netloc` component of the parsed repository URL (in this case, `gitlab.com`) and it compares it to the `netloc` component of each repository URL in the config until it finds a match. Because both of the configured repositories have the same `netloc` in their URL, the credentials for the first one stored in the configuration file will always be used.

I suspect an appropriate fix would involve passing the repository name through to the authenticator (when available) rather than performing the potentially lossy process of name -> url -> name that is currently used.

sinoroc:
```
[tool.poetry.dependencies]
bar-library = {version = "^1.0.0", source = "bar-pypi"}
```

[...]

PR #3400, opened 2020-11-22
Update document, dependency-specification.md

squirrel532:
Add description of `source` property. Enhance doc of #908 .

I search around and find this feature in issue, it will be good if we write it down to documents.

# Pull Request Check List

Resolves:  -

- [ ] Added **tests** for changed code.
- [x] Updated **documentation** for changed code.

PR #3477, opened 2020-12-11
Docs: adding a note about being able to specify a package's source repository

glencairn:
# Pull Request Check List

Resolves: #3476

- [ ] Added **tests** for changed code. (N/A)

---

Adding a little note to the documentation about the ability to specify the source repository for a dependency.

As mentioned in the linked issue, poetry's dependency resolution was failing because the first private repository poetry looked in for a specified dependency returned a 404. Whether or not that's the right behavior is a different discussion, but the problem can immediately be solved using the `source` property.

### DIFF
--- a/docs/cli.md
+++ b/docs/cli.md
@@ -352,7 +352,7 @@ about dependency groups.
 
 ### Options
 
-* `--group (-D)`: The group to add the dependency to.
+* `--group (-G)`: The group to add the dependency to.
 * `--dev (-D)`: Add package as development dependency. (**Deprecated**)
 * `--editable (-e)`: Add vcs/path dependencies as editable.
 * `--extras (-E)`: Extras to activate for the dependency. (multiple values allowed)
@@ -360,8 +360,8 @@ about dependency groups.
 * `--python`: Python version for which the dependency must be installed.
 * `--platform`: Platforms for which the dependency must be installed.
 * `--source`: Name of the source to use to install the package.
-* `---allow-prereleases`: Accept prereleases.
-* `--dry-run`: Outputs the operations but will not execute anything (implicitly enables --verbose).
+* `--allow-prereleases`: Accept prereleases.
+* `--dry-run`: Output the operations but do not execute anything (implicitly enables --verbose).
 * `--lock`: Do not perform install (only update the lockfile).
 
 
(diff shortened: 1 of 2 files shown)


sinoroc:
I link https://github.com/python-poetry/poetry/issues/3269 as somewhat related.

glencairn:
@sinoroc, thanks for the link. I also updated the list of options in the documentation for the `poetry add` command since it was definitely missing several of the available options.

lcbm:
Sorry in advance, if this isn't the right place to ask... also, if there's documentation on what I'm about to say, please LMK! I wasn't able to find anything on `source` _usage_ and it's _intended behavior_ :sob: 

It **is** clear to me, from the documentation in this MR, that one purpose of `source` is to specify, from a bunch of external sources I may have, which one I want to use for each specific package.

Before elaborating what **is not** clear to me, please consider the `pyproject.toml` below :point_down: 

```text
[tool.poetry.dependencies]
private-dependency = {version = "^1.0.0", source = "private-source"}

# and 'private-library' depends on a 'public-library' which is not in "private-source" (my private registry)

[[tool.poetry.source]]
name = "private-source"
default = true
```

It **is not** clear to me if:

1. `source` will enforce that `private-library` and all of its requirements are installed from `private-source` (as `pip` does with `--index-url`) OR if `source` will enforce that only `private-library` is installed from `private-source` and its requirements will be installed according to poetry's default behavior (**precedence** to `source` but will still look for the `private-dependency`'s dependencies in `pypi`);

2. let's say I have multiple `sources` in my `pyproject.toml` as well as multiple dependencies and each dependency is referencing their own source. If I want to ensure that each of my dependencies are looked up **exclusively** in their respective sources, can I all my sources with `default = true` under `[[tool.poetry.source]]`?

Lastly, I looked for an example `pyproject.toml` with multiple repositories [here](https://python-poetry.org/docs/repositories/) and couldn't find one... I'll drop a suggestion here to add a piece in this MR for that as well, considering that it's in that context that `source` should be used :smile:

ggabriel96:
There's another PR doing this, https://github.com/python-poetry/poetry/pull/3400. It's just missing the CLI docs.

neersighted (dismissed):

neersighted (approved):
Rebased onto current master. Just waiting on the CI to be unbroken to merge this (see #4753).

Issue #3355, opened 2020-11-12
How to handle several repositories with same content?


gsemet:
Hello,

I have a use case I do not see supported by poetry. Tell me if I am wrong:
- I have an artifactory A that can mirror pypi.org (this is caching, which is good)
- This actifactory also stores private package
- A is on private network on a dedicated baremetal server running in-premise
- A is only accessible to some user (private network)

so far so good.

I also have another artifactory server B that I need to use a different authentication scheme, and behave like this:
- packages on A is duplicated on B
- using B costs a bit more (it's on a cloud server), and require a stronger authentication
- so when possible, A is to be used preferably
- if A is not available, then B can be used.

if I do not use the lockfile, it can work. I would be happy if I could generate 2 lockfiles (with the same version of each package of course), and user could "choose" if he wants to fetch from A or from B. Any idea on how to do it properly?

sinoroc:
Have you already configured custom repositories? How do your current `pyproject.toml` and/or `poetry.toml` files look like?

gsemet:
My idea (maybe wrong, tell me !) is that I would like to have something like:
- Pyproject.toml:
```
[[tool.poetry.source]]
name = "artifactory-A"
url = "http://artifactory.myserverA/artifactory/api/pypi/pypi/simple"
default = true

[[tool.poetry.source]]
name = "artifactory-B"
url = "http://artifactory.myserverB/artifactory/api/pypi/pypi/simple"
default = false
```

It would generate 2 lock files: `poetry-A.lock` and `poetry-B.lock`. If A is available, `poetry install` would use `poetry-A.lock`, if not, it would use `poetry-B.lock`.

Content of both file would need to be synchronised (the same packages can be found on both servers). I did that by
- `rm poetry.lock`
- change the `default` key in `tool.poetry.source`
- rename `poetry.lock` to the wanted lock filename
- now by renaming the lock file, I can install from the server I want.

I know that having 2 lock files would look like overkill, but this is the only way to reuse the existing behavior without having to rework everything. From what I see on !447, it looks like if the `poetry.lock` file could be configurable in both commands `poetry lock --no-update` and `poetry install` it would do the job.

sinoroc:
I see... Are lock files critical in your use case?

I had in my mind that you wanted to be able to override the main source repository on a case-by-case basis (per user maybe). For example, say you have the following in your `pyproject.toml`:

```
[[tool.poetry.source]]
name = "myindex"
url = "https://a.dev/simple/"
[...]

Issue #2940, opened 2020-09-19
Need a way to set a global PyPI repo URL for all project manually.


dev-techmoe:
- [x] I have searched the [issues](https://github.com/python-poetry/poetry/issues) of this repo and believe that this is not a duplicate.
- [x] I have searched the [documentation](https://python-poetry.org/docs/) and believe that my question is not covered.

## Feature Request

Hello everyone, I'm a python developer live in Mainland China. As well known the network connection is so bad in my country that we cannot connect to the official PyPI repo directly sometimes. So most of us usually some PyPI mirrors in our region (like https://mirrors.aliyun.com/pypi/simple/, it's deployed by AlibabaCloud) to get a fast and stabilizing network connection. PIP gives us a way to set the index globally using `pip config set global.index-url `. But I cannot find related config items in Poetry.
Okay, I know what you want to say, I know I can set repo info in my `pyproject.toml` like this.
```toml
[[tool.poetry.source]]
name = "aliyun"
url = "https://mirrors.aliyun.com/pypi/simple/"
default = true
```
But it will cause some tricky problems. If I have some workmates in other countries, they don't need repo mirrors to speed up service, even it will make them cannot use PyPI repo (thanks to stupid GFW), or we need to use Github Action as our CI service, they all don't need this config item. **So I think the PyPI repo URL should not as a project config in my project repo**, we need a way to set in our dev environment and use it for all projects. Thank you.

gaoming714:
I see what you mean, there should really be a global control, or I hope the project leader can use even pip configuration.
Because I pip to download when it is convenient to use pqi to control repo.

Is there some one can answer it, make poetry use pip repo.

Maybe you need someway to connect google, you know it.

abn:
@dev-techmoe I can definitely appreciate your situation and I do agree that there are various cases where you would want to use a mirror instead of public pypi index.

Poetry has some hard-coded assumptions about PyPI. This means taht simply setting a url for PyPI will not work. The assumptions include (but not limited to), the use of the JSON endpoint to get package metadata, There is no guarentee that this exists for all mirrors. We need to,

1. Allow for poetry to be confired with first-class knowledge of PyPI mirrors.
2. Improve legacy repository support to ensure that we detect "true" legacy repos (ie, without json endpoints).
3. With these changes, we should be able to safely configure mirror urls and identify if they are warehouse compatible and be smart about what endpoints we use.

[...]

PR #3497, opened 2020-12-17
WIP - Allow empty repository url if it's already configured

orlevii:
# Pull Request Check List

Resolves: #issue-number-here

------

I've tested it manually and it works as expected.

How do I write a test for this change?

### DIFF
--- a/docs/docs/repositories.md
+++ b/docs/docs/repositories.md
@@ -97,6 +97,13 @@ name = "foo"
 url = "https://foo.bar/simple/"
 ```
 
+Alternatively, if you've already configured your repository with `poetry config repositories.foo`, you can remove the `url` key from your `pyproject.toml`.
+
+```toml
+[[tool.poetry.source]]
+name = "foo"
+```
+
 From now on, Poetry will also look for packages in your private repository.
 
 !!!note
--- a/poetry/factory.py
+++ b/poetry/factory.py
@@ -142,15 +142,15 @@ class Factory(BaseFactory):
         from .utils.helpers import get_cert
         from .utils.helpers import get_client_cert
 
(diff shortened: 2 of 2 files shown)


sinoroc:
Isn't there something like the repos from `pyproject.toml` are only for fetch (download) and the repos from `poetry config` (`poetry.toml`, etc.) are only for publish (upload)?

orlevii:
@sinoroc What do you mean?

I don't want to specify the repository url in each and every `pyporject.toml`. 

This change allows to configure the repository globally (on the machine) using `poetry config`, and under the `pyporject.toml` specify only the repository name to use.

Is there already a way to do that?

sinoroc:
I agree with you. I believe this is how it should be. I disagree with poetry's design decision to have index URLs in the `pyproject.toml` altogether, as I stated in https://github.com/python-poetry/poetry/issues/3355#issuecomment-726683158. There are some more tickets going in that same direction.

I am just pointing what the current intended behaviour seems to be, as far as I understood it from the following section of the doc for example (and the whole page actually):
* https://python-poetry.org/docs/repositories/#install-dependencies-from-a-private-repository

I could be mistaken of course.

orlevii:
@sinoroc 
Yeah, got your point.
Looks like repositories added with `poetry config` can be only used for publishing purposes.

(just like shown here: https://github.com/python-poetry/poetry/issues/2940#issuecomment-710682131)

It causes some inconsistencies.. Setting credentials with `poetry config http-basic.repo USERNAME PASSWORD` is effecting  both installation and publishing commands.

Anyway, I hope this issue will be addressed soon.

Secrus:
No activity for a long time. Closing

Issue #3465, opened 2020-12-08
Request control of number of parallel processes


jmuncaster:
On a fast computer with a fast internet connection and a library with lots of dependencies, we get errors running the default `poetry install` because too many parallel processes are hitting the pypi servers at once.

Disabling the new installer fixes this problem.

If we could control the size of the parallel process pool we could achieve some of the speed gains of the new installer without the risks of being blocked by the pypi server.

Reference PR below

_Originally posted by @abn in https://github.com/python-poetry/poetry/issues/2595#issuecomment-704875923_

danielloader:
This is quite the pain when you have a gigabit home connection, pypicloud is exhausting mysql connections when I try to resolve a large dependency tree.

PR #3516, opened 2020-12-27
Add maximum of workers to config

maxispeicher:
Resolves: #3465 

Added a new `installer.max-workers` property to the configuration. If `installer.parallel` is set to true, the executor tries to get the new configuration property and will limit the number of processes to the number set in the configuration. If the value is not set, the same default value as before will be used.


poetry/installation/executor.py
```diff
@@ -42,14 +43,12 @@ def __init__(self, env, pool, config, io, parallel=None):
...
         if parallel:
-            # This should be directly handled by ThreadPoolExecutor
-            # however, on some systems the number of CPUs cannot be determined
-            # (it raises a NotImplementedError), so, in this case, we assume
-            # that the system only has one CPU.
-            try:
-                self._max_workers = os.cpu_count() + 4
-            except NotImplementedError:
-                self._max_workers = 5
+            self._max_workers = get_default_max_workers()
```
neersighted, line 46: This logic seems quite jumbled up.

I'd suggest something like this instead (more elegant constructs are of course welcomed):

```python
            self._max_workers = config.get("installer.max-workers", get_default_max_workers())
            self._max_workers = min(self._max_workers, max_workers)
```

neersighted (changes requested):

poetry/installation/executor.py
```diff
@@ -54,6 +56,7 @@ def __init__(
         config: "Config",
         io: "IO",
         parallel: bool = None,
+        max_workers: Optional[int] = None,
```
neersighted, line 59: I don't see a need to make this configurable here -- unless you'd like to move the config of this value outside the class. I'd argue against any additional churn, however.

neersighted, line 59: I don't see a need to make this configurable here -- unless you'd like to move the config of this value outside the class. I'd argue against any additional churn, however.

poetry/utils/helpers.py
```diff
@@ -141,3 +141,18 @@ def is_dir_writable(path: Path, create: bool = False) -> bool:
         return False
     else:
         return True
+
+
+def get_max_workers(desired_max_workers: Optional[int] = None):
```
neersighted, line 146: Given that this is only used in the one location and not unit-tested separately, I think I'd prefer to see this as a `@staticmethod` `def _get_max_workers` below the function it's used in.

poetry/installation/executor.py
```diff
@@ -68,14 +71,10 @@ def __init__(
...
         if parallel:
-            # This should be directly handled by ThreadPoolExecutor
-            # however, on some systems the number of CPUs cannot be determined
-            # (it raises a NotImplementedError), so, in this case, we assume
-            # that the system only has one CPU.
-            try:
-                self._max_workers = os.cpu_count() + 4
-            except NotImplementedError:
-                self._max_workers = 5
+            self._max_workers = config.get(
```
neersighted, line 74: I think this should be inverted, actually, e.g.:

```python
self._max_workers = self._get_max_workers(desired_max_workers=config.get('installer-max-workers'))
```

poetry/console/commands/config.py
```diff
@@ -87,6 +88,11 @@ def unique_config_values(self) -> Dict[str, Tuple[Any, Any, Any]]:
                 boolean_normalizer,
                 True,
             ),
+            "installer.max-workers": (
+                lambda val: int(val) > 0,
+                lambda val: int(val),
+                get_max_workers(),
```
neersighted, line 94: Since we're only going to check this again, let's just make the default None.

neersighted (changes requested):
Another round of feedback -- I still think the logic is really jumbled, but can be simplified by calling `self._get_max_workers()` in only one place.

I also think we need slightly better tests and docs -- first off, if we're going to cap the number of max workers with `min()`, we should document that.

Secondly, let's be sure to test that behavior using mocking of both a working (but different from the testing machine) `os.cpu_count()` and a version that always throws `NotImplementedError`.

poetry/installation/executor.py
maxispeicher: At first I thought to add it in the same as `parallel`, which is explicitly added but also unused right now. But for now I've removed it.

tests/config/test_config.py
```diff
@@ -11,22 +12,36 @@
...
+from poetry.config.config import boolean_normalizer
+from poetry.config.config import int_normalizer
 
 
 if TYPE_CHECKING:
     from pathlib import Path
 
 
-def get_boolean_options(config: Optional[Dict[str, Any]] = None) -> str:
+def flatten(config: Optional[Dict[str, Any]], parent_key: str = "", sep="."):
```
neersighted, line 23: The type signature here is wrong (config cannot be `None`), and I'm not really a fan of this non-generator implementation. I think we should just add a dev dependency on [FlatDict](https://flatdict.readthedocs.io/en/stable/index.html) and use it in place of this custom function.

tests/config/test_config.py
```diff
@@ -11,22 +12,36 @@
...
+def get_options_based_on_normalizer(
+    normalizer: Callable, config: Optional[Config] = None
+) -> str:
     if config is None:
-        config = Config.default_config
+        config = Config()
 
-    for k, v in config.items():
-        if isinstance(v, bool) or v is None:
+    flattened_config = flatten(config.default_config)
```
neersighted, line 40: Doesn't this make passing in a `Config` meaningless as we'll always use the static default? Feels like the original was more correct.

tests/config/test_config.py
```diff
@@ -43,10 +58,14 @@ def test_config_get_processes_depended_on_values(
...
-        for name in get_boolean_options():
-            env_var = "POETRY_{}".format(re.sub("[.-]+", "_", name).upper())
-            yield name, env_var, env_value, value
+    for normalizer, values in [
+        (boolean_normalizer, [("true", True), ("false", False)]),
+        (int_normalizer, [("4", 4), ("2", 2)]),
+    ]:
+        for env_value, value in values:
+            for name in get_options_based_on_normalizer(normalizer=normalizer):
+                env_var = "POETRY_{}".format(re.sub("[.-]+", "_", name).upper())
```
neersighted, line 67: Let's use `string + concat` here instead of `.format()`.

neersighted (changes requested):
Overall looks good, and the executor tests are well-done. I just have some general feedback on the changes to the config tests.

tests/config/test_config.py
maxispeicher: I've added FlatDict, however, I'm also not a big fan of including a package just for this single test case.

tests/config/test_config.py
maxispeicher: Yes you're right passing a `Config` would be meaningless, but also in the original implementation it was confusing in my eyes. The problem why it needed to change is that the check for `None` does not indicate a `bool` config variable anymore.
So either the config names need to be hard-coded again or the normalizer can be used.

tests/config/test_config.py
maxispeicher: `f-Strings` should also work fine right? Or do you prefer `+`?

neersighted: I generally try to avoid a f-string where we're adding a single character and have some inline string transformation -- `+ concat` is just more readable with less syntax.

tests/config/test_config.py
neersighted: Yeah, I could go either way on that as well. However, we already do so with Zipp, Deepdiff, etc, and it's a common pattern in general and likely useful to have around to prevent additional bespoke implementations.

tests/config/test_config.py
neersighted: I think it's fine to hardcode to always use the default config (`Config.default_config`) and we can modify this if another test case needs to pass one in later.

tests/config/test_config.py
```diff
@@ -20,24 +19,9 @@
...
-
-def get_options_based_on_normalizer(
-    normalizer: Callable, config: Optional[Config] = None
-) -> str:
-    if config is None:
-        config = Config()
-
-    flattened_config = flatten(config.default_config)
+def get_options_based_on_normalizer(normalizer: Callable) -> str:
+    config = Config()
```
neersighted, line 23: This line does nothing -- let's skip it...

tests/config/test_config.py
```diff
@@ -20,24 +19,9 @@
...
+    flattened_config = FlatDict(config.default_config, delimiter=".")
```
neersighted, line 24: ...in favor of this.

```suggestion
    flattened_config = FlatDict(Config.default_config, delimiter=".")
```

neersighted (changes requested):

neersighted (approved):

PR #3521, opened 2020-12-31
* added additional description for poetry add covering git+ssh URL us…

m0rcq:
Adds additional description covering the use of git+ssh URLs with poetry add command which is a common use case when using privare repos.

# Pull Request Check List

Resolves: n/a


sinoroc:
I think there is at least one similar pull request already. Not sure if there is overlap or not.

* https://github.com/python-poetry/poetry/pull/3389

stephsamson (approved):
No overlap and it's a good to be more verbose, especially in cases like these. Thanks both!

PR #3522, opened 2020-12-31
Remove non-existing --path option, add missing options

fredrikaverpil:
I can't see this option in `poetry add --help` and I've also tried to actually run the command `poetry add --path $PATH` and I get this:

```python
clikit.api.args.exceptions.NoSuchOptionException: The "--path" option does not exist.
```

I verified the above with both version 1.1.4 as well as latest master commit https://github.com/python-poetry/poetry/commit/2aab9bcd495e11e5f4491aa72a9510773cc4a90e.

### DIFF
--- a/docs/docs/cli.md
+++ b/docs/docs/cli.md
@@ -257,9 +257,14 @@ poetry add "git+https://github.com/pallets/flask.git@1.1.1[dotenv,dev]"
 ### Options
 
 * `--dev (-D)`: Add package as development dependency.
-* `--optional` : Add as an optional dependency.
-* `--dry-run` : Outputs the operations but will not execute anything (implicitly enables --verbose).
-* `--lock` : Do not perform install (only update the lockfile).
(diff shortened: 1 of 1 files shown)


finswimmer (changes requested):
Thanks a lot for this change  @fredrikaverpil .

I've noticed that this section is missing some of the options described in `poetry add --help`. Could you please add those?

fin swimmer

fredrikaverpil:
> Thanks a lot for this change @fredrikaverpil .
> 
> I've noticed that this section is missing some of the options described in `poetry add --help`. Could you please add those?

@finswimmer sure, have a look now. Maybe the title of this PR should be updated...

finswimmer (changes requested):

fredrikaverpil:
@finswimmer 👍

finswimmer (approved):
Thanks a lot @fredrikaverpil :+1:

Issue #3397, opened 2020-11-21
publish command doesn't publish to PyPI despite showing no errors


adder46:
- [x] I am on the [latest](https://github.com/python-poetry/poetry/releases/latest) Poetry version.
- [x] I have searched the [issues](https://github.com/python-poetry/poetry/issues) of this repo and believe that this is not a duplicate.
- [x] If an exception occurs when executing a command, I executed it again in debug mode (`-vvv` option).

- **Linux 5.4.0-53-generic XUbuntu 20.04 x86_64 GNU/Linux**: 
- **Poetry version 1.1.4**: 
- **Link of a [Gist](https://gist.github.com/adder46/a75befcd640e03e9322a7ea28500bc09) with the contents of your pyproject.toml file**: 

## Issue

I have an issue with poetry not publishing the package to PyPI. The project is called `dpaster`, however the PyPI repo name is called `thepaster`. I've configured it to point to the remote repository:

The output of config command:
```
$ poetry config --list
cache-dir = "/home/alex/.cache/pypoetry"
experimental.new-installer = true
installer.parallel = true
repositories.dpaster.url = "https://pypi.org/project/thepaster"
virtualenvs.create = true
virtualenvs.in-project = null
virtualenvs.path = "{cache-dir}/virtualenvs"  # /home/alex/.cache/pypoetry/virtualenvs
```

The output of publish command:

```
$ poetry publish -r dpaster -vvv

Username: adder46
Password: 
Publishing dpaster (3.2.2) to dpaster
 - Uploading dpaster-3.2.2-py3-none-any.whl 100%
 - Uploading dpaster-3.2.2.tar.gz 100%
```
As you can see, it doesn't report any error, however, the project ends up not getting published.

dusktreader:
I can confirm the same behavior when attempting to publish to a gitlab pypi clone.

Trying the same publishing sequence with twine showed that the upload of the files was successful, but registering the new versions was getting rejected by a 403 error due to permissions.

Publishing with twine:
```bash
$ python -m twine upload --repository gitlab dist/* --verbose                                                                                                                                                                                   
Uploading distributions to https:/---.com/api/v4/projects/1330/packages/pypi
  dist/fetch_config-0.1.4-py3-none-any.whl (3.8 KB)
  dist/fetch_config-0.1.5-py3-none-any.whl (3.8 KB)
  dist/fetch_config-0.4.4-py3-none-any.whl (4.2 KB)
[...]

PR #3549, opened 2021-01-07
publish: print on failure, simplify logic

hauntsaninja:
Should help with cases like #3397

The logic coalesces a couple things, so that we hit a single except clause
that prints "Uploading <package> failed"

The existing tests in https://github.com/python-poetry/poetry/blob/master/tests/publishing/test_uploader.py
should verify that this doesn't cause regressions.

### DIFF
--- a/poetry/publishing/uploader.py
+++ b/poetry/publishing/uploader.py
@@ -208,30 +208,10 @@ class Uploader:
     def _upload(
         self, session: requests.Session, url: str, dry_run: Optional[bool] = False
     ) -> None:
-        try:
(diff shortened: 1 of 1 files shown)


neersighted:
This is pretty stale now. Would you be willing to rebase onto current `master`?

hauntsaninja:
Merged in master!

neersighted:
This is blocked by `master`'s build being broken. I'd suggest rebasing onto #4753 -- this PR can be merged once that PR makes it into master.

neersighted (approved):

Issue #3542, opened 2021-01-06
why get-poetry.py still prompting for PATH modification even after providing --no-modify-path?


mohan43u:
- [x] I am on the [latest](https://github.com/python-poetry/poetry/releases/latest) Poetry version.
- [x] I have searched the [issues](https://github.com/python-poetry/poetry/issues) of this repo and believe that this is not a duplicate.
- [x] If an exception occurs when executing a command, I executed it again in debug mode (`-vvv` option).

- **OS version and name**: Linux
## Issue

I downloaded get-poetry.py as per the Documentation (https://python-poetry.org/docs/#osx-linux-bashonwindows-install-instructions) and disabled modifying PATH using (--no-modify-path). But still get-poetry.py prompting to modify shell rc files or not, why? isn't explicitly setting modify_path (through --no-modify-path) not enough for get-poetry.py to understand the intention of the user?

This prompting becomes irritating when automating poetry installation, I have to do `"echo 'n' | python3 get-poetry.py --no-modify-path"` instead of simply saying `"python3 get-poetry.py --no-modify-path"`.

finswimmer:
The `get-poetry.py` is now deprecated and will be replaced by `install-poetry.py` which will not change any env variables at all.

See https://python-poetry.org/blog/announcing-poetry-1.2.0a1/#deprecation-of-the-get-poetrypy-script

Issue #2445, opened 2020-05-20
POETRY_CACHE_DIR or cache-dir config are not used at all.


mdgilene:
- [x] I am on the [latest](https://github.com/python-poetry/poetry/releases/latest) Poetry version.
- [x] I have searched the [issues](https://github.com/python-poetry/poetry/issues) of this repo and believe that this is not a duplicate.
- [x] If an exception occurs when executing a command, I executed it again in debug mode (`-vvv` option).

- **OS version and name**: Windows 10
- **Poetry version**: 1.0.5
- **Link of a [Gist](https://gist.github.com/) with the contents of your pyproject.toml file**: https://gist.github.com/mdgilene/0da1568f8ac90615e89cd1331115d80a

## Issue

Setting the `cache-dir` location using either the environment variable `POETRY_CACHE_DIR` or with `poetry config cache-dir /path/to/new/cache` does not work. Poetry still uses the default location `C:\Users\<user>\AppData\Local\pypoetry\Cache`.

This will not work for me as there are size constraints on my user profile directory. Beacuse of this I need to relocate this directory to another drive.

I believe this to be cause by the fact that the environment/config settings are never read when determining the cache location.

poetry/utils/appdirs.py
```python
def user_cache_dir(appname):
    r"""
    Return full path to the user-specific cache dir for this application.
        "appname" is the name of application.
    Typical user cache directories are:
        macOS:      ~/Library/Caches/<AppName>
        Unix:       ~/.cache/<AppName> (XDG default)
        Windows:    C:\Users\<username>\AppData\Local\<AppName>\Cache
    On Windows the only suggestion in the MSDN docs is that local settings go
    in the `CSIDL_LOCAL_APPDATA` directory. This is identical to the
    non-roaming app data dir (the default returned by `user_data_dir`). Apps
    typically put cache data somewhere *under* the given dir here. Some
    examples:
        ...\Mozilla\Firefox\Profiles\<ProfileName>\Cache
        ...\Acme\SuperApp\Cache\1.0
    OPINION: This function appends "Cache" to the `CSIDL_LOCAL_APPDATA` value.
    """
    if WINDOWS:
        # Get the base path
        path = os.path.normpath(_get_win_folder("CSIDL_LOCAL_APPDATA"))

        # Add our app name and Cache directory to it
        path = os.path.join(path, appname, "Cache")
    elif sys.platform == "darwin":
        # Get the base path
        path = expanduser("~/Library/Caches")

        # Add our app name to it
[...]

PR #3550, opened 2021-01-07
improved get-poetry.py and added POETRY_HOME support to poetry command

mohan43u:
* fixed prompting even after no-modify-path explicitly given in commandline
* use POETRY_HOME if available instead of XDG_*_HOME variables in nix systems

# Pull Request Check List

Resolves: #3542
Resolves: #2445

### DIFF
--- a/docs/docs/index.md
+++ b/docs/docs/index.md
@@ -60,11 +60,12 @@ python install-poetry.py --uninstall
 POETRY_UNINSTALL=1 python install-poetry.py
 ```
 
-By default, Poetry is installed into the user's platform-specific home directory.
-If you wish to change this, you may define the `POETRY_HOME` environment variable:
+By default, Poetry is installed into the user's platform-specific home directory. If you wish to change this, you may define the `POETRY_HOME` environment variable. In Linux, this variable can also be used with `poetry` command to substitute `XDG_CACHE_HOME`, `XDG_CONFIG_HOME` and `XDG_DATA_HOME` variables.
 
 ```bash
 POETRY_HOME=/etc/poetry python install-poetry.py
+POETRY_HOME=/tmp/poetry poetry config --list
+POETRY_HOME=/tmp/poetry poetry env info
 ```
 
 If you want to install prerelease versions, you can do so by passing `--preview` option to `install-poetry.py`
--- a/get-poetry.py
+++ b/get-poetry.py
@@ -476,7 +476,7 @@ class Installer:
         return version, current_version
 
     def customize_install(self):
-        if not self._accept_all:
+        if not self._accept_all and self._modify_path:
             print("Before we start, please answer the following questions.")
             print("You may simply press the Enter key to leave unchanged.")
 
--- a/poetry/utils/appdirs.py
+++ b/poetry/utils/appdirs.py
@@ -37,7 +37,9 @@ def user_cache_dir(appname: str) -> str:
 
     Typical user cache directories are:
         macOS:      ~/Library/Caches/<AppName>
-        Unix:       ~/.cache/<AppName> (XDG default)
+        Unix:       ${POETRY_HOME}/.cache/<AppName> or
+                    ${XDG_CACHE_HOME}/<AppName> or
+                    ~/.cache/<AppName>
         Windows:    C:\Users\<username>\AppData\Local\<AppName>\Cache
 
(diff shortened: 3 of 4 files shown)


mohan43u:
Can someone review this code?

The changes I made helps keep everything related to 'poetry' inside one place under POETRY_HOME

TBBle:
I don't see how this resolves #2445, unless the intention is to not support the `cache-dir` config option, or at least document that it *only* applies to the [virtual-env default path](https://github.com/python-poetry/poetry/blob/bf30ca696aa47bcc42937910d620c95c95c0741e/poetry/config/config.py#L37) and [the 'artifacts' directory](https://github.com/python-poetry/poetry/blob/bf30ca696aa47bcc42937910d620c95c95c0741e/poetry/installation/chef.py#L25-L27), but not other forms of caching done by Poetry.

mohan43u:
> I don't see how this resolves #2445,

given [reply](https://github.com/python-poetry/poetry/issues/2445#issuecomment-758372662)

> unless the intention is to not support the `cache-dir` config option, or at least document that it _only_ applies to the [virtual-env default path](https://github.com/python-poetry/poetry/blob/bf30ca696aa47bcc42937910d620c95c95c0741e/poetry/config/config.py#L37) and [the 'artifacts' directory](https://github.com/python-poetry/poetry/blob/bf30ca696aa47bcc42937910d620c95c95c0741e/poetry/installation/chef.py#L25-L27), but not other forms of caching done by Poetry.

I'm not able to understand `but not other forms of caching done by Poetry` part. could you please elaborate?

TBBle:
Right now, the `cache-dir` config _is_ used for 'virtualenv' (by default) and 'artifacts' (hard-coded), but not 'repositories' which ignores the setting and uses `locations.CACHE_DIR`. That's the problem seen in #2445, that sometimes it ignores the cache-dir config.

This PR doesn't change that behaviour, it instead adds `POETRY_HOME` as a preference over `XDG_CACHE_HOME`, which I separately think is the wrong approach, but certainly doesn't fix that the cache-dir config option _should_ override this, and does not.

mohan43u:
> This PR doesn't change that behaviour, it instead adds `POETRY_HOME` as a preference over `XDG_CACHE_HOME`, which I separately think is the wrong approach, but certainly doesn't fix that the cache-dir config option _should_ override this, and does not.

This PR doesn't address #2445 directly, but it fixes the underlying problem which lead to #2445 issue. As I already [mentioned](https://github.com/python-poetry/poetry/issues/2445#issuecomment-758566243), `XDG_*_HOME` dirs are suppose to be fallback dirs when user not explicitly specify the location to store files. But poetry is using these dirs as they are the one user has to use and expects user to change `XDG_*_HOME` variables rather than providing a way to explicitly specify where to store the files.

TBBle:
That's not the underlying problem in #2445. The underlying problem there _is_ the surface problem:
> `poetry config cache-dir /tmp/poetrycache` does not prevent Poetry trying to write to to `$XDG_CACHE_HOME`.

With this PR, the the same problem exists, except you add "or `$POETRY_HOME/.cache`" to the end of the problem description. It doesn't change the problem.

`POETRY_HOME` is _not_ a way to "explicitly specify the location to store (cache) files". It's a way to explicitly specify the location to install Poetry's binaries when using get-poetry.py, which has no relationship with #2445.

mohan43u:
There is no solution in current poetry to address the problem this PR is addressing which is providing user a way to explicitly specify the location to use rather than asking user to change `XDG_*_HOME` dirs.

TBBle:
If by "location to use", you only mean "Poetry's cache files", then there is a [documented solution](https://python-poetry.org/docs/configuration/#cache-dir-string), it just doesn't work: #2445.

If by "location to use" you include everything that Poetry touches (i.e. you want a "standalone" or "portable" install) then no, there isn't such a thing, but supporting that through `POETRY_HOME` _doesn't_ fix #2445, which is a problem with redirecting Poetry's caches specifically, not producing a "standalone" install.

neersighted:
Closing this for a couple reasons: `get-poetry.py` is deprecated, and `$POETRY_HOME` has always been meant to control where poetry's isolated environment is created, but it was not intended to silo all of the files poetry generates.

I do think it's worth addressing the inconsistent cache dir in poetry -- but this PR is only tangentially related and not the way to go about solving that problem.

PR #3583, opened 2021-01-16
version: take --short into account when updating

### LINKED ISSUE #3577: --short has no effect when `poetry version` is passed a new version

- **OS version and name**: Ubuntu 16.04
- **Poetry version**: 1.1.4
- **Link of a [Gist](https://gist.github.com/) with the contents of your pyproject.toml file**: n/a

## Issue

Hi there!

Assuming the current version of a project is 0.1.0, this is what happens when trying to update to the next patch version with the `--short` option:

```console
$ poetry version patch --short
```

Instead, I would like this to be output (only the new version number):

```console
$ poetry version patch --short
0.1.1
```

My use case is scripting: if the user only provides a bump rule to my script, I'd like to be able to easily parse the new version computed by poetry so that I can e.g. create a new git tag easily, and so on.

Thanks!

sprt:
Fixes #3577.

This makes `poetry version --short <version>` output only the updated
version number, similarly to `poetry version --short`.

# Pull Request Check List

Resolves: #issue-number-here

- [x] Updated **documentation** for changed code. `N/A IMHO`

### DIFF
--- a/poetry/console/commands/version.py
+++ b/poetry/console/commands/version.py
@@ -48,11 +48,14 @@ patch, minor, major, prepatch, preminor, premajor, prerelease.
                 self.poetry.package.pretty_version, version
             )
 
-            self.line(
-                "Bumping version from <b>{}</> to <fg=green>{}</>".format(
-                    self.poetry.package.pretty_version, version
+            if self.option("short"):
+                self.line("{}".format(version))
+            else:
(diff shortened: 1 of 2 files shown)


stephsamson (approved):

stephsamson:
Thanks!

PR #3601, opened 2021-01-22
Add json export

bneijt:
Adds json export feature

Resolves: to be issued

# Draft pull request

Resolves: #issue-number-here


bneijt:
Having an intermediate format is stupid, I think accessing the lock toml is just as good. Abandoning this idea

Issue #2783, opened 2020-08-09
Fall back to public PyPI if 'ConnectionError' occurs


jaklan:
- [x] I have searched the [issues](https://github.com/python-poetry/poetry/issues) of this repo and believe that this is not a duplicate.
- [x] I have searched the [documentation](https://python-poetry.org/docs/) and believe that my question is not covered.

## Feature Request

I have a private PyPI repository defined in `pyproject.toml`. The repository is located within my corporate network. When I don't use VPN and try to install any package, I get such an error:
```
  ConnectionError

  HTTPConnectionPool(host=SOME_HOST, port=80): Max retries exceeded with url: SOME_URL (Caused by NewConnectionError('<urllib3.connection.HTTPConnection object at 0x109f17f10>: Failed to establish a new connection: [Errno 8] nodename nor servname provided, or not known'))
```
which is quite obvious. But what I would like to see is an option to ignore the error and try to install packages from the next expected repository, in that case - the public PyPI. Currently, the fallback mechanism works only for cases, when a repository is reachable, but some package is missing.

stormsw:
Any progress on merging the PR?

abn:
@stormsw does #3337 resolve this?

stormsw:
@abn I think this one is bit different.
The lines are here https://github.com/python-poetry/poetry/blob/master/poetry/repositories/legacy_repository.py#L435
```python
        try:
            response = self.session.get(url)
            if response.status_code in (401, 403):
                self._log(
                    "Authorization error accessing {url}".format(url=url),
                    level="warning",
                )
                return
            if response.status_code == 404:
                return
            response.raise_for_status()
        except requests.HTTPError as e:
            raise RepositoryError(e) # <<< here we have the error
```

ralbertazzi:
Honestly I wouldn't want to see this ever getting implemented in Poetry. It goes against the whole determinism / reproducibility idea of the lockfile that is one of the keypoints of the projects. If the lockfile says that a package must be downloaded from repo xyz then that same package must always be downloaded from repo xyz.

Note how there can be several reasons why you want to make sure packages are always downloaded from the same source, other than mere determinism:
- The package on the private repo might not be exactly the same as the one on pypi (eg a wheel compiled in a different way) 
- You want to avoid [dependency confusion attacks](https://fossa.com/blog/dependency-confusion-understanding-preventing-attacks/)

[...]

PR #3490, opened 2020-12-14
Issue #3483 Add cert retrieval for requests

dusktreader:
The authenticator.py code already retrieves credentials from the config
for every request based on url matching. This change makes the
authenticator also retrieve certs from the config for each request based
on url matching.

Also includes unit tests.

# Pull Request Check List

Resolves: #3483 

- [x] Added **tests** for changed code.
- [x] Updated **documentation** for changed code.

PR #3608, opened 2021-01-27
Handle 403 response from secondary repos

GooseYArd:
# Pull Request Check List

Resolves: #2783

Hi all- this is my first PR for the project and thus I'm still somewhat clueless about the architecture as well as the teams practices, so begging your pardon in advance. Although this fixes the issue I've been encountering, I'm not confident that it's the ideal fix. I'm consulting with our Artifactory admins now to see if its possible to modify the behavior of the repo. Also I suspect that a few other open bugs related to failures when using secondary repositories may be the result of the same issue.

### DIFF
--- a/docs/docs/repositories.md
+++ b/docs/docs/repositories.md
@@ -120,6 +120,10 @@ a custom certificate authority or client certificates, similarly refer to the ex
 `certificates` section. Poetry will use these values to authenticate to your private repository when downloading or
 looking for packages.
 
+401 or 403 responses from secondary repositories will generate a warning
+message but are not considered as fatal, as some types of private repository
+servers may return a 403 if a request is made for a package not present on the
+server.
 
 ### Disabling the PyPI repository
 
--- a/poetry/repositories/legacy_repository.py
+++ b/poetry/repositories/legacy_repository.py
@@ -395,6 +395,7 @@ class LegacyRepository(PyPiRepository):
             response = self.session.get(url)
             if response.status_code == 404:
                 return
+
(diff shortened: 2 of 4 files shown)


GooseYArd:
This appears to be a dupe of #3337

GooseYArd:
@abn whoops I meant to close this PR after I noticed #3337, I'll close it now and may send it up again with just the docs changes. By the way- I made this change after tracking down a pair of issues that prevented us from being able to use our private pypi repository (which requires client cert auth) with poetry. As it turns out, both of them already had fixes in pull requests, but both pull requests have been un-merged for a very long time. I'm wondering if you might mind to look at the other as well? It's #3490 . We'd love to use poetry at my company, but these two small issues make it impossible, and I'm worried that it might be difficult for us to get bug fixes merged in the future :(

abn:
@GooseYArd happy to get that reviewed and merged for 1.2.

Regarding your "getting bug fixes merged" concern, I do intend to focus a bit more on the more enterprise-y use cases and hardening going forward fwiw. But that said, as with most FOSS, we are volunteer driven with all core maintainers holding down fairly involved day jobs. We do want to add more maintainers who are willing to spend more time reviewing community contributions, but this is a challenge on its own. :)

As the project matures, I suspect the need for this kind of bug fixes or improvements would be less critical I feel, since hopefully then it would be gnarly edge cases.

GooseYArd:
@abn I understand completely, and delighted to help out however I can :) Thanks and good luck with the next release!!

PR #3618, opened 2021-01-29
Migrate to the new version of Cleo

sdispater:
This PR changes the code base to migrate to the new, work in progress, version of Cleo which will help improve things in order to better implement some features planned for the 1.2 release.

The major changes are the following:

- A new command loading system that make it possible to only load and execute the requested command without having to import any other command. This will help with performance issues when executing Poetry (note that a lot of changes must be done internally to improve performances further).
- The sub command system has changed and now works by using namespaced commands to achieve the same thing.


finswimmer (approved):
Looks good to me. :+1: 

I recognize that you've started to use type hints instead of type comments. I've avoided this in my PRs regarding adding missing type hints, to make cherry-picking to the 1.1 easier. But I guess once this PR is merged, we should switch the type comments to type hints (with com2ann this seems to be quiet easy) for the sake of consistency.

I let it to you to merge this PR.

sdispater:
@finswimmer The `master` branch is already Python 3.6+ only so there is no reason not to use type hints, that's why I started introducing them.

And, yes, I plan on converting everything to type hints.

GooseYArd:
This morning I noticed that "poetry env info -p" and "poetry env info --path" are failing with a stack trace; a bisect narrowed it down to this PR. I haven't had a chance to dig into it yet.

Issue #3570, opened 2021-01-13
Poetry install dependencies/dev-dependencies from lock file


nathanrpage97:
- [x] I have searched the [issues](https://github.com/python-poetry/poetry/issues) of this repo and believe that this is not a duplicate.
- [x] I have searched the [documentation](https://python-poetry.org/docs/) and believe that my question is not covered.

## Feature Request

Add the ability to run `poetry install {list-of-dependencies}` /  `poetry install --dev {list-of-dev-dependencies}`.

This will install these dependencies and their sub-dependencies from the lock file instead of installing all dependencies. This is useful in CI where not all dev-dependencies are needed for a job.

sinoroc:
Maybe look at the planned feature [_dependency groups_](https://github.com/python-poetry/poetry/issues/1644).

nathanrpage97:
I've seen that, but I still think that this is also useful to give the most fine grained control on dependency installation.

finswimmer:
Duplicate of #3033

PR #3623, opened 2021-01-30
install only given dev dependency

finswimmer:
This PR introduce a new option `--dev` for the `poetry install` command. Instead of installing all development dependencies, one can use this option, to limit the installation to the given dev dependencies. `--dev` can be used together with `--dev-only` to decide, whether the dev dependencies should be installed together with the package or without.

Closes: https://github.com/python-poetry/poetry/issues/3570

# Pull Request Check List

### DIFF
--- a/docs/docs/cli.md
+++ b/docs/docs/cli.md
@@ -116,6 +116,16 @@ by passing the `--dev-only` option. Note that `--no-dev` takes priority if both 
 poetry install --dev-only
 ```
 
+If not all development dependencies should be installed, but only a specific list, you can do this py passing
+ the `--dev` option. This option can be used multiple times. If only development dependencies should be used, you
+can use this option together with the `--dev-only` option.
+
+```bash
+poetry install --dev black
+poetry install --dev black --dev isort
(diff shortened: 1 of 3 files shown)


sdispater:
I think it would be better to wait for the implementation of the dependency groups before we go this route.

Poetry was never about installing single dependencies separately but always about managing groups of dependencies consistently.

Also, it could lead to option bloat and confusion which defeats the simplicity purpose of Poetry.

jcmdln:
I think rather than providing yet another option, it would be better for `poetry install` to accept arguments much like how `poetry add` works, but only refers to packages defined in `poetry.lock` if it exists. Alternatively, `poetry add` could add some invariant option to `--lock` where it only adds a package based on a lockfile.

Dependency groups look nice but are overkill for parallelized jobs where you want to run one utility, like mypy, without having to install a bunch of other dependencies. The solution to that with dependency groups would be to assign something to multiple groups, one with the package name and another in a broader `ci` group or something if desired, but I'd rather have one less explicit location to declare since my tests are wrapped with `tox` anyway.

Issue #3344, opened 2020-11-09
Unable to install packages from Azure DevOps private repo, 401 unauthorized


FilBot3:
I am trying to install from an Azure DevOps Artifacts Python feed. I can publish to the feed if I use the following:

```bash
poetry publish \
  --repository 'azure' \
  --username 'orgname' \
  --password "$(cat ~/azdo_pat.txt)" \
  --build
```

Which I set up by doing:

```bash
poetry config \
  --local \
  repostiory.azure \
  https://pkgs.dev.azure.com/${ORG_NAME}/${PROJECT_NAME}/_packaging/python-azure-artifacts/pypi/simple/
```

So then my `poetry.toml` has:

```toml
[repositories]
[repositories.azure]
url = "https://pkgs.dev.azure.com/${ORG_NAME}/${PROJECT_NAME}/_packaging/python-azure-artifacts/pypi/simple/"
```

My `pyproject.toml` has this section added to it.

```toml
[[tool.poetry.source]]
name = "azure"
url = "https://pkgs.dev.azure.com/${ORG_NAME}/${ROFJECT_NAME}/_packaging/python-azure-artifacts/pypi/simple/"
secondary = true
```

Now, when ever I attempt to do anything with a package, I get 401 unauthorized errors. [I had issues before with publishing](https://stackoverflow.com/q/63909006/2009612) to the feed when configuring the `http-basic.azure` before. 

```bash
poetry add azdo                                                                                           2 ↵

RepositoryError

401 Client Error:  for url: https://pkgs.dev.azure.com/${ORG_NAME}/${PROJECT_NAME}/_packaging/python-azure-artifacts/pypi/simple/azdo/

at ~/.poetry/lib/poetry/repositories/legacy_repository.py:393 in _get
389│             if response.status_code == 404:
390│                 return
391│             response.raise_for_status()
392│         except requests.HTTPError as e:
→ 393│             raise RepositoryError(e)
394│
395│         if response.status_code in (401, 403):
396│             self._log(
397│                 "Authorization error accessing {url}".format(url=url), level="warn"
```

What do I need to do to make the `add` or `install` portion work now?

My `pyproject.toml`

```toml
[tool.poetry]
name = "developer.dudley.terraform_stuff"
version = "0.1.0"
description = "Python Tasks for the project"
authors = ["Phillip Dudley <Predatorian3@gmail.com>"]
license = "MIT"

[[tool.poetry.source]]
name = "azure"
[...]

PR #3647, opened 2021-02-05
Add azure keyring support

bneijt:
# Pull Request Check List

Resolves: #3344 

I have added support for keyring lookup based on repository url via a `get_credentials` call, which is what `artifacts-keyring` seems to do. Every `get_credentials` call is at Microsoft speed, so I had to also add a cache on the keyring values.

### DIFF
--- a/poetry/installation/authenticator.py
+++ b/poetry/installation/authenticator.py
@@ -152,9 +152,8 @@ class Authenticator(object):
                 continue
 
             parsed_url = urllib.parse.urlsplit(url)
-
             if netloc == parsed_url.netloc:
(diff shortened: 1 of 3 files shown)


btjones-me:
Hi @bneijt  thanks very much for this PR - I'm having this problem and would love for this functionality to be merged. Do we have any idea on when this might be released ?

bneijt:
Closing this pr.

The caching required is quite ugly and we have been happily working around this issue by just doing `poetry config` with the right keys in our CI pipeline. A scripted version of this approach is available in https://github.com/python-poetry/poetry/issues/3344#issuecomment-774066065

No reason to keep this open.

PR #3660, opened 2021-02-08
Fixed propagation of markers to dependencies. Fixes #3254

### LINKED ISSUE #3254: 1.1.3 regression. poetry exports wrong constraints

- **OS version and name**: Fedora 33
- **Poetry version**: 1.1.3
- **Link of a [Gist](https://gist.github.com/) with the contents of your pyproject.toml file**: https://gist.github.com/apollo13/cda3ea5e3a135d67eeed4be099bcb7e9

## Issue

The linked pyproject.toml generates a broken requirements.txt (it wrongly applies `sys_platform=darwin`). But first off lets see the graph:
```
pyinstaller 4.0 PyInstaller bundles a Python application and all its dependencies into a single package.
├── altgraph *
├── macholib >=1.8
│   └── altgraph >=0.15 
├── pefile >=2017.8.1
│   └── future * 
├── pyinstaller-hooks-contrib >=2020.6
└── pywin32-ctypes >=0.2.0
```

as we can see the toplevel dependency `pyinstaller` directly requires `althgraph` and then `macholib` also requires `altgraph`.

Looking at `poetry.lock` we see:
```
[[package]]
name = "pyinstaller"
version = "4.0"
description = "PyInstaller bundles a Python application and all its dependencies into a single package."
category = "main"
optional = false
python-versions = "*"

[package.dependencies]
altgraph = "*"
macholib = {version = ">=1.8", markers = "sys_platform == \"darwin\""}
pefile = {version = ">=2017.8.1", markers = "sys_platform == \"win32\""}
pyinstaller-hooks-contrib = ">=2020.6"
pywin32-ctypes = {version = ">=0.2.0", markers = "sys_platform == \"win32\""}

[[package]]
name = "macholib"
version = "1.14"
description = "Mach-O header analysis and editing"
category = "main"
optional = false
python-versions = "*"

[package.dependencies]
altgraph = ">=0.15"
```

This tells us that `pyinstaller` depends on `macholib` on darwin systems. The export looks like this:
```
altgraph==0.17; sys_platform == "darwin" \
    --hash=sha256:c623e5f3408ca61d4016f23a681b9adb100802ca3e3da5e718915a9e4052cebe \
    --hash=sha256:1f05a47122542f97028caf78775a095fbe6a2699b5089de8477eb583167d69aa
future==0.18.2; python_version >= "2.6" and python_full_version < "3.0.0" and sys_platform == "win32" or python_full_version >= "3.3.0" and sys_platform == "win32" \
    --hash=sha256:b1bead90b70cf6ec3f0710ae53a525360fa360d306a86583adc6bf83a4db537d
macholib==1.14; sys_platform == "darwin" \
    --hash=sha256:c500f02867515e6c60a27875b408920d18332ddf96b4035ef03beddd782d4281 \
    --hash=sha256:0c436bc847e7b1d9bda0560351bf76d7caf930fb585a828d13608839ef42c432
pefile==2019.4.18; sys_platform == "win32" \
    --hash=sha256:a5d6e8305c6b210849b47a6174ddf9c452b2888340b8177874b862ba6c207645
pyinstaller-hooks-contrib==2020.9 \
    --hash=sha256:a5fd45a920012802e3f2089e1d3501ef2f49265dfea8fc46c3310f18e3326c91 \
    --hash=sha256:c382f3ac1a42b45cfecd581475c36db77da90e479b2f5bcb6d840d21fa545114
pyinstaller==4.0 \
    --hash=sha256:970beb07115761d5e4ec317c1351b712fd90ae7f23994db914c633281f99bab0
pywin32-ctypes==0.2.0; sys_platform == "win32" \
    --hash=sha256:24ffc3b341d457d48e8922352130cf2644024a4ff09762a2261fd34c36ee5942 \
    --hash=sha256:9dc2d991b3479cc2df15930958b674a48a227d5361d413827a4cfd0b5876fc98
```

Here `altgraph` is incorrectly labeled as `darwin`, propbably due to the fact that `macholib` is only required on `darwin` and as such the constraints trickles down. This makes sense but should not happen if there is another dependency on the same package.

apollo13:
Resolves: #3254

I ticked the documentation checkbox simply because I think there is no part in the documentation that I could adjust for this.

@abn This issue got introduced in 1188b31287df42de615d25b1642ec894c2bb0c12 which you authored; do you think you could give this PR a review?

The main logic change here is that I changed marker merging from `intersect` to `union`. The rationale for this can be followed easier if you look at the following (commented) dependency graph from the ticket:

```
pyinstaller 4.0 PyInstaller bundles a Python application and all its dependencies into a single package.
├── altgraph *  # PyInstaller always requires altgraph
├── macholib >=1.8 # This has a marker to only install on darwin
│   └── altgraph >=0.15  # This is a dependency also of macholib
```

Now the situation is as follows: Since `altgraph` is a non-marked dependency by `PyInstaller` it should *always* be installed, there is no way around this. An intersection of `*` and `sys_platform='darwin'` yields `sys_platform='darwin'` which would result `altgraph` to be only installed on `darwin`. An union ensure that it is installed always.

### DIFF
--- a/tests/utils/test_exporter.py
+++ b/tests/utils/test_exporter.py
@@ -175,6 +175,200 @@ foo==1.2.3; python_version < "3.7"
     assert expected == content
 
 
+def test_exporter_can_export_requirements_txt_poetry(tmp_dir, poetry):
+    """Regression test for #3254"""
+
+    poetry.locker.mock_lock_data(
+        {
+            "package": [
+                {
+                    "name": "poetry",
+                    "version": "1.1.4",
+                    "category": "main",
+                    "optional": False,
+                    "python-versions": "*",
+                    "dependencies": {"keyring": "*"},
+                },
+                {
+                    "name": "junit-xml",
+                    "version": "1.9",
+                    "category": "main",
+                    "optional": False,
+                    "python-versions": "*",
+                    "dependencies": {"six": "*"},
+                },
+                {
+                    "name": "keyring",
+                    "version": "21.8.0",
+                    "category": "main",
+                    "optional": False,
+                    "python-versions": "*",
+                    "dependencies": {
+                        "SecretStorage": {
+                            "version": "*",
+                            "markers": "sys_platform == 'linux'",
+                        }
+                    },
+                },
+                {
+                    "name": "secretstorage",
+                    "version": "3.3.0",
+                    "category": "main",
+                    "optional": False,
+                    "python-versions": "*",
+                    "dependencies": {"cryptography": "*"},
+                },
+                {
+                    "name": "cryptography",
+                    "version": "3.2",
+                    "category": "main",
+                    "optional": False,
+                    "python-versions": "*",
+                    "dependencies": {"six": "*"},
+                },
+                {
+                    "name": "six",
+                    "version": "1.15.0",
+                    "category": "main",
+                    "optional": False,
+                    "python-versions": "*",
+                },
+            ],
+            "metadata": {
+                "python-versions": "*",
+                "content-hash": "123456789",
+                "hashes": {
+                    "poetry": [],
+                    "keyring": [],
+                    "secretstorage": [],
+                    "cryptography": [],
+                    "six": [],
+                    "junit-xml": [],
+                },
+            },
+        }
+    )
+    set_package_requires(
+        poetry, skip={"keyring", "secretstorage", "cryptography", "six"}
+    )
+
+    exporter = Exporter(poetry)
+
+    exporter.export("requirements.txt", Path(tmp_dir), "requirements.txt")
+
+    with (Path(tmp_dir) / "requirements.txt").open(encoding="utf-8") as f:
+        content = f.read()
+
+    # The dependency graph:
+    # junit-xml 1.9 Creates JUnit XML test result documents that can be read by tools such as Jenkins
+    # └── six *
+    # poetry 1.1.4 Python dependency management and packaging made easy.
+    # ├── keyring >=21.2.0,<22.0.0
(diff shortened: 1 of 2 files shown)


tests/utils/test_exporter.py
```diff
@@ -241,7 +435,7 @@ def test_exporter_can_export_requirements_txt_with_nested_packages_and_markers(
             "c==7.8.9; sys_platform == 'win32' and python_version < '3.7'"
         ),
         "d": dependency_from_pep_508(
-            "d==0.0.1; python_version < '3.7' and platform_system == 'Windows' and sys_platform == 'win32'"
+            "d==0.0.1; platform_system == 'Windows' and python_version < '3.7' or sys_platform == 'win32' and python_version < '3.7'"
```
apollo13, line 438: @abn This change changes tested behavior but I'd argue that the previous behavior was in fact wrong:

  * `a` has `python_version < 3.7`
  * `a` depends on `b` & `c` which have the markers `platform_system == 'Windows'` and `sys_platform == 'win32'` respectively
  * `b` and `c` both depends on `d` and as such their markers should trickle down, but `d` should get installed whenever *one* of the parent markers is true, not only when *both* are true.

On the other hand this `python_version<3.7` is required for all packages, so the constraint could be rewritten as:
```
(platform_system == 'Windows' or sys_platform == 'win32' ) and python_version < '3.7'"
```
but that would probably be more work (and the current form is technically correct). For large dependency trees it might be worth to get the python version out of the nested constraints as far as possible though. But I guess this might require plenty of specialcasing.

stephsamson (approved):

Issue #3663, opened 2021-02-08
poetry-core 1.0.2 breaks dependency resolution for packages without the specified dependency version (*) in debug mode


kleschenko:
- [x] I am on the [latest](https://github.com/python-poetry/poetry/releases/latest) Poetry version.
- [x] I have searched the [issues](https://github.com/python-poetry/poetry/issues) of this repo and believe that this is not a duplicate.
- [x] If an exception occurs when executing a command, I executed it again in debug mode (`-vvv` option).

- **OS version and name**: 
- **Poetry version**: 1.1.4
- **Link of a [Gist](https://gist.github.com/) with the contents of your pyproject.toml file**: https://gist.github.com/kleschenko/77df4d568dff3722b61caae402b00b5a

## Issue

Poetry started to fail to lock dependencies containing `*` in their sub-dependency versions after the upgrade to `poetry-core==1.0.2`  when running in debug mode (`-vv`).

```
$ poetry lock -vv
Skipping virtualenv creation, as specified in config file.
Updating dependencies
Resolving dependencies...
   1: fact: poetry-test is 0.1.0
   1: derived: poetry-test
   1: fact: poetry-test depends on flake8-black (>=0.2.1,<0.3.0)
   1: selecting poetry-test (0.1.0)
   1: derived: flake8-black (>=0.2.1,<0.3.0)
   1: fact: flake8-black (==0.2.1) depends on flake8 (>=3.0.0)
   1: Version solving took 0.521 seconds.
   1: Tried 1 solutions.

  Stack trace:

  14  ~/.local/pipx/venvs/poetry/lib/python3.6/site-packages/clikit/console_application.py:131 in run
      status_code = command.handle(parsed_args, io)

  13  ~/.local/pipx/venvs/poetry/lib/python3.6/site-packages/clikit/api/command/command.py:120 in handle
      status_code = self._do_handle(args, io)

  12  ~/.local/pipx/venvs/poetry/lib/python3.6/site-packages/clikit/api/command/command.py:171 in _do_handle
      return getattr(handler, handler_method)(args, io, self)

  11  ~/.local/pipx/venvs/poetry/lib/python3.6/site-packages/cleo/commands/command.py:92 in wrap_handle
      return self.handle()

  10  ~/.local/pipx/venvs/poetry/lib/python3.6/site-packages/poetry/console/commands/lock.py:34 in handle
      return self._installer.run()

[...]

PR #3667, opened 2021-02-09
fix: debug message may break due to no dep version

honnix:
When colorizing the debug message, a depended package without version
constraint may fail regex match. This PR makes the version info part
optional.

# Pull Request Check List

Resolves: #3663

### DIFF
--- a/poetry/puzzle/provider.py
+++ b/poetry/puzzle/provider.py
@@ -710,7 +710,7 @@ class Provider:
 
         if message.startswith("fact:"):
             if "depends on" in message:
-                m = re.match(r"fact: (.+?) depends on (.+?) \((.+?)\)", message)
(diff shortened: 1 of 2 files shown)


honnix:
> ld-elf.so.1: /usr/local/lib/libpython3.8.so.1.0: Undefined symbol "close_range@FBSD_1.6"

Looks like FreeBSD CI is broken?

honnix:
Not sure this is the right fix. Maybe I should look into why the original message doesn't contain `(*)`.

honnix:
@kleschenko Could you please take a look at this? Thanks.

kleschenko:
@honnix looks like this issue has been already resolved in the latest `poetry`

honnix:
@kleschenko ah ok that's great! Thanks for letting me know. Closing.

abn:
Thank you for the contribution nonetheless @honnix

PR #3701, opened 2021-02-18
Prefer 'python3' executables over 'python' during install

orf:
Currently the setup script uses `python` first, and falls back to `python3`. This means python2 is often picked up on systems that have Python3 installed.

This MR swiches the check order to use `python3` first, then `python`.

In the future when Poetry drops python2 support, without this change the install script will be broken on any system that has Python2 installed as `python`, regardless of `python3` being present.

### DIFF
--- a/get-poetry.py
+++ b/get-poetry.py
@@ -638,7 +638,7 @@ class Installer:
 
     def _which_python(self):
         """Decides which python executable we'll embed in the launcher script."""
-        allowed_executables = ["python", "python3"]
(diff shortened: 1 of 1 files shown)


dchevell:
I hope this gets merged - I was nearly put off using poetry when the recommended install method defaulted to python2 on macOS. This seems like a no brainer

abn:
The `get-poetry.py` script will be deprecated in favour of #3706 once that is merged.

abn (approved):

Issue #3534, opened 2021-01-03
Create a .exe file using Default Installer


naveen521kk:
- [x] I have searched the [issues](https://github.com/python-poetry/poetry/issues) of this repo and believe that this is not a duplicate.
- [x] I have searched the [documentation](https://python-poetry.org/docs/) and believe that my question is not covered.

## Feature Request

First, using the default installer as specified in the documentation, https://python-poetry.org/docs/#windows-powershell-install-instructions, it creates a `.bat` file in `%USERPROFILE%\.poetry\bin` folder. I want it to create an `exe` file instead of a batch file.

It can be done using the way `setuptools` does it. Essentially, it would be to find `cli.exe` or `cli-64.exe` and rename it to `poetry.exe` and create a file called `poetry-script.py` instead of the currently present, `poetry`(without any extension) file.
Usually, `setuptools` is installed by default. If it not, the `.tar.gz` for windows, can include a copy of `cli.exe` and `cli-64.exe` which can be easily compiled using, https://github.com/pypa/setuptools/blob/main/launcher.c

Note: `poetry-script.py` needs to have proper shebang with full paths.

TBBle:
The [`distlib.scripts package`](https://distlib.readthedocs.io/en/stable/reference.html#the-distlib-scripts-package) is what Pip uses to generate executables for this use-case. Rather than bundling a stock executable that runs a script by the same name, it generates an executable with the script text appeneded, taking advantage of a Python 3 feature (I think).

naveen521kk:
Well, sorry for this late response. I think it would be better to use distlib because it provides some API. but still unsure how to...

abn:
@naveen521kk happy to look at clearer proposal on how we might do this. If using `pip` or `pipx` to install `poetry` does this anyway, I would be inclined to rather make a not int he documentation and move on. 

Since the bootstrap script is intended to work with no or limited dependencies, I would be hesitent to even require `distlib`. As this is not necessarily always available and there are known issues inconsistencies with distro vendored installations of the package (eg: ubuntu).

However, if we indeed proceed to making changes, they will need to be on top of #3706 as that will be the new bootstrap mechanism. It might also be a good idea to verify if the problem goes away by virtue of the new changes, specifically how `poetry` installed into the virtual environment. Test this out and let us know.

Note: I am not recommending either of the following options, but simply identifying a way forward in the discussion.

#### Option 1
[...]

Issue #3187, opened 2020-10-13
Py3.10


dimaqq:
Python-3.10.0a1 is available, it would be awesome to be able to use it.

Today I can work around like so:
```command
pushd ~/.poetry/lib/poetry/_vendor/
ln -s py3.9 py3.10
```

IIRC from 3.9, following needs to be done:
- [ ] ~change in code to add vendored deps for 3.10~
- [ ] change in the test matrix to include 3.10 (#3842)
- [ ] ~change in the release script to build against 3.10 (some platforms?~

abn:
@dimaqq you can [install poetry via `pipx`](https://python-poetry.org/docs/#installing-with-pipx) or plain old [`pip`](https://python-poetry.org/docs/#installing-with-pip) to use under python 3.10. However, note that the current "supported" python versions for discovery does not include `3.10`. However, you can manually activate an environment for your project.

dimaqq:
Oh yes, I did use `poetry env use 3.10` for an existing project.
Simple projects (few deps) work just fine with the work-around.
Complex projects are likely to not work due to some dep stumbling on BPO-40747

HacKanCuBa:
Note that `pip install poetry` under Python 3.10rc (alpine) fails compiling `msgpack` and `cyptography`, but I guess it's an issue for those packages. I just let this here in case someone else arrives here like me.

Issue #2030, opened 2020-02-12
On windows poetry shell always start a cmd


ggtools:
- [x] I have searched the [issues](https://github.com/python-poetry/poetry/issues) of this repo and believe that this is not a duplicate.
- [x] I have searched the [documentation](https://python-poetry.org/docs/) and believe that my question is not covered.

## Feature Request

On Windows it looks like `poetry shell` always spawn a cmd even when run from a Powershell window. As expected the `$SHELL` environment variable is not working. This could be worked around by using `poetry run "C:\Program Files\PowerShell\7-preview\pwsh.exe"`

TBBle:
I'm not sure why `shellingham.detect_shell()` is not working, but always starting `cmd.exe` is because on Windows, Poetry [ignores `$SHELL` and uses `$COMPSEC` instead](https://github.com/python-poetry/poetry/blob/258ba2992309544213cbf9c9410de9b35349d8c1/poetry/utils/shell.py#L50-L51).

A quick test suggested that `shellingham.detect_shell()` should work, returning `('pwsh', 'C:\...\pwsh.exe')` as I'd expect when I just install `shellingham` from PyPI.

I tried setting `$ENV:COMSPEC` before running `poetry shell` and it broke `poetry shell`, and also `poetry run ...` and in fact any other `poetry` command, because the installation of Poetry has put a `poetry.bat` in my path, which tries to use `$ENV:COMSPEC` if it's set, to run.

So if I in-line the batch file, and set `$ENV:COMPSEC`, it works...
```
$ENV:COMSPEC="C:\Program Files\PowerShell\7\pwsh.exe"
py -3 "${ENV:USERPROFILE}\.poetry\bin\poetry" shell
```

Of course, that's not a workaround, but suggests that if Poetry honoured `$SHELL` if present, and only if absent fell back to `$COMPSEC`, it'd be doing that the help says it does.

Issam-seghir:
> 
py -3 ${ENV:USERPROFILE}\.poetry\bin\poetry shell
this command work for me 💯

PR #3706, opened 2021-02-19
Implement a new bootstrapping method

sdispater:
This method introduces a new script `install-poetry.py` which implements a new way to bootstrap Poetry. The advantages of this new method are:

- It no longer relies on the prebuilt release packages of Poetry (see https://github.com/python-poetry/poetry/releases/tag/1.1.4 for an example) which had the issues of having frozen dependencies which made upgrading dependencies much harder without releasing a new bug fix release of Poetry. This will eventually make our release process easier.
- It relies on standards for the install locations of both Poetry and the `poetry` script.
- It supports installing from a git repository or a file/directory.
- It no longer tries to modify the PATH but instead will just provide guidance on how to do it if necessary.
- It will use the correct Python interpreter, i.e. the one used to invoke the script, when calling the `poetry` script.

As to what the script actually does:

  - Downloads the virtualenv package to a temporary directory and add it to sys.path.
  - Creates a virtual environment in the correct OS data dir which will be
      - `%APPDATA%\\pypoetry` on Windows
      -  `~/Library/Application Support/pypoetry` on MacOS
      - `${XDG_DATA_HOME}/pypoetry` (or `~/.local/share/pypoetry` if it's not set) on UNIX systems
      - In `${POETRY_HOME}` if it's set.
  - Installs the latest or given version of Poetry inside this virtual environment.
  - Installs a `poetry` script in the Python user directory (or `${POETRY_HOME/bin}` if `POETRY_HOME` is set).

Resolves: #3534
Resolves: #2030
Resolves: #3819
Resolves: #3828
Resolves: #3187

## Pull Request Check List

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -39,14 +39,21 @@ jobs:
 
       - name: Bootstrap poetry
         shell: bash
-        run: |
-          python -m ensurepip
-          python -m pip install --upgrade pip
-          python -m pip install .
+        run: python install-poetry.py -y
+
+      - name: Update PATH
+        if: ${{ matrix.os != 'Windows' }}
+        shell: bash
+        run: echo "$HOME/.local/bin" >> $GITHUB_PATH
+
+      - name: Update Path for Windows
+        if: ${{ matrix.os == 'Windows' }}
+        shell: bash
+        run: echo "$APPDATA\Python\Scripts" >> $GITHUB_PATH
 
       - name: Configure poetry
         shell: bash
-        run: python -m poetry config virtualenvs.in-project true
+        run: poetry config virtualenvs.in-project true
 
       - name: Set up cache
         uses: actions/cache@v2
@@ -58,12 +65,12 @@ jobs:
       - name: Ensure cache is healthy
         if: steps.cache.outputs.cache-hit == 'true'
         shell: bash
-        run: timeout 10s python -m poetry run pip --version || rm -rf .venv
+        run: timeout 10s poetry run pip --version || rm -rf .venv
 
       - name: Install dependencies
         shell: bash
-        run: python -m poetry install
+        run: poetry install
 
       - name: Run pytest
         shell: bash
-        run: python -m poetry run python -m pytest -p no:sugar -q tests/
+        run: poetry run python -m pytest -p no:sugar -q tests/
--- a/README.md
+++ b/README.md
@@ -17,42 +17,76 @@ The [complete documentation](https://python-poetry.org/docs/) is available on th
 ## Installation
 
 Poetry provides a custom installer that will install `poetry` isolated
-from the rest of your system by vendorizing its dependencies. This is the
-recommended way of installing `poetry`.
+from the rest of your system.
 
+### osx / linux / bashonwindows install instructions
 ```bash
-curl -sSL https://raw.githubusercontent.com/python-poetry/poetry/master/get-poetry.py | python
+curl -sSL https://raw.githubusercontent.com/python-poetry/poetry/master/install-poetry.py | python -
+```
+### windows powershell install instructions
+```powershell
+(Invoke-WebRequest -Uri https://raw.githubusercontent.com/python-poetry/poetry/master/install-poetry.py -UseBasicParsing).Content | python -
 ```
 
-Alternatively, you can download the `get-poetry.py` file and execute it separately.
+**Warning**: The previous `get-poetry.py` installer is now deprecated, if you are currently using it
+you should migrate to the new, supported, `install-poetry.py` installer.
 
-The setup script must be able to find one of following executables in your shell's path environment:
+The installer installs the `poetry` tool to Poetry's `bin` directory. This location depends on you system:
 
-- `python` (which can be a py3 or py2 interpreter)
-- `python3`
-- `py.exe -3` (Windows)
-- `py.exe -2` (Windows)
+- `$HOME/.local/bin` for Unix
+- `%APPDATA%\Python\Scripts` on Windows
 
-If you want to install prerelease versions, you can do so by passing `--preview` to `get-poetry.py`:
+If this directory is not on you `PATH`, you will need to add it manually
+if you want to invoke Poetry with simply `poetry`.
 
-```bash
-python get-poetry.py --preview
-```
+Alternatively, you can use the full path to `poetry` to use it.
 
-Similarly, if you want to install a specific version, you can use `--version`:
+Once Poetry is installed you can execute the following:
 
 ```bash
-python get-poetry.py --version 0.7.0
+poetry --version
 ```
 
-Using `pip` to install `poetry` is also possible.
+If you see something like `Poetry (version 1.2.0)` then you are ready to use Poetry.
+If you decide Poetry isn't your thing, you can completely remove it from your system
+by running the installer again with the `--uninstall` option or by setting
+the `POETRY_UNINSTALL` environment variable before executing the installer.
 
 ```bash
-pip install --user poetry
+python install-poetry.py --uninstall
+POETRY_UNINSTALL=1 python install-poetry.py
 ```
 
-Be aware, however, that it will also install poetry's dependencies
-which might cause conflicts.
+By default, Poetry is installed into the user's platform-specific home directory.
+If you wish to change this, you may define the `POETRY_HOME` environment variable:
+
+```bash
+POETRY_HOME=/etc/poetry python install-poetry.py
+```
+
+If you want to install prerelease versions, you can do so by passing `--preview` option to `install-poetry.py`
+or by using the `POETRY_PREVIEW` environment variable:
+
+```bash
+python install-poetry.py --preview
+POETRY_PREVIEW=1 python install-poetry.py
+```
+
+Similarly, if you want to install a specific version, you can use `--version` option or the `POETRY_VERSION`
+environment variable:
+
+```bash
+python install-poetry.py --version 1.2.0
+POETRY_VERSION=1.2.0 python install-poetry.py
+```
+
+You can also install Poetry for a `git` repository by using the `--git` option:
+
+```bash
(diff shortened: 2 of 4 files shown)


install-poetry.py
```diff
@@ -0,0 +1,810 @@
...
+    parser.add_argument("--version", help="install named version", dest="version")
+    parser.add_argument(
+        "-f",
+        "--force",
+        help="install on top of existing version",
+        dest="force",
+        action="store_true",
+        default=False,
+    )
+    parser.add_argument(
```
abn, line 749: This script does not modify path right?

install-poetry.py
```diff
@@ -0,0 +1,810 @@
...
+        "--git",
+        dest="git",
+        action="store",
+        help=(
+            "Install from a git repository instead of fetching the latest version "
+            "of Poetry available online."
+        ),
+    )
+
```
abn, line 789: We should probably add an option to recreate the virtual environment.

install-poetry.py
```diff
@@ -0,0 +1,810 @@
...
+                [sys.executable, "-m", "pip", "install", "virtualenv", "-t", tmp_dir],
+                stdout=subprocess.PIPE,
+                stderr=subprocess.STDOUT,
+            )
+
+            sys.path.insert(0, tmp_dir)
+
+            import virtualenv
+
+            virtualenv.cli_run([str(env_path), "--clear"])
```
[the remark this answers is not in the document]
abn: Additionally we should consider the use of `--copies` and `--prompt`.

install-poetry.py
```diff
@@ -0,0 +1,810 @@
...
+    )
+    parser.add_argument(
+        "--uninstall",
+        help="uninstall poetry",
+        dest="uninstall",
+        action="store_true",
+        default=False,
+    )
+    parser.add_argument(
+        "--file",
```
abn, line 772: Are we expecting wheel here? Or are we anticipating a fully offline install?

.github/workflows/main.yml
```diff
@@ -38,14 +38,21 @@ jobs:
 
       - name: Bootstrap poetry
         shell: bash
-        run: |
-          python -m ensurepip
-          python -m pip install --upgrade pip
-          python -m pip install .
+        run: python install-poetry.py -y
```
abn, line 41: Should we use the script here? I would think the better approach here would be to handle test cases outside of this job. We have gone down the path of using this script before and often times it is has ended up in more issues than solutions.

docs/docs/index.md
```diff
@@ -18,109 +18,86 @@ on Windows, Linux and OSX.
 ## Installation
 
 Poetry provides a custom installer that will install `poetry` isolated
-from the rest of your system by vendorizing its dependencies. This is the
-recommended way of installing `poetry`.
+from the rest of your system.
+
+This is the recommended way of installing `poetry`.
```
abn, line 23: I would like us to consider not recommending one method but rather treat script, pip and pipx as equals.

install-poetry.py
```diff
@@ -0,0 +1,810 @@
+"""
```
abn, line 1: The script itself works from a basic test and I do think this is the right direction, ie. move away from platform bundles.

abn (changes requested):

docs/docs/index.md
```diff
@@ -18,109 +18,86 @@ on Windows, Linux and OSX.
...
-!!!warning
 
-    Be aware that it will also install Poetry's dependencies
-    which might cause conflicts with other packages.
+### Alternative installation methods
 
 #### Installing with `pipx`
 
-Using [`pipx`](https://github.com/cs01/pipx) to install Poetry is also possible. [pipx] is used to install Python CLI applications globally while still isolating them in virtual environments. This allows for clean upgrades and uninstalls. pipx supports Python 3.6 and later. If using an earlier version of Python, consider [pipsi](https://github.com/mitsuhiko/pipsi).
+Using [`pipx`](https://github.com/cs01/pipx) to install Poetry is also possible.
```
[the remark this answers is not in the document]
sdispater: Fixed :-)

install-poetry.py
```diff
@@ -0,0 +1,810 @@
...
+
+        if self._git:
+            specification = "git+" + version
+        elif self._file:
+            specification = version
+        else:
+            specification = f"poetry=={version}"
+
+        subprocess.call(
+            [str(python), "-m", "pip", "install", specification],
```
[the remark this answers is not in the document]
sdispater: It should, yes. I don't see why it wouldn't but I know Windows can be weird.

install-poetry.py
sdispater: I removed the option

install-poetry.py
sdispater: What do you mean? The virtual environment is always recreated. Or do you mean an option to not recreate the virtual environment?

install-poetry.py
sdispater: I changed the option to `--path` which can take either a distribution (wheel or sdist) or a directory. This means that completely offline installations are no longer possible.

docs/docs/index.md
sdispater: I removed the recommendation.

install-poetry.py
sdispater: It will also help make plugins straightfoward, i.e. entrypoints, which is not easily doable with the current solution.

.github/workflows/main.yml
sdispater: I think we should, yes. It would be weird in my opinion to not use our own documented method of installing Poetry.

abn: There are cople of things here.

1. I do not think we should recommend one option over the other. Atleast not when it comes to bootstrap, pip or pipx.
2. I would agree that this helps with our test coverage, however, this has historically been problematic causing us to flap between `pip install` and our bootstrap for various reasons. I would be more agreeable with the solution if the bootstrap mechanism was not a moving target.

All that said, I do not see this as a blocker for this change.

README.md
```diff
@@ -17,42 +17,76 @@ The [complete documentation](https://python-poetry.org/docs/) is available on th
 ## Installation
 
 Poetry provides a custom installer that will install `poetry` isolated
-from the rest of your system by vendorizing its dependencies. This is the
-recommended way of installing `poetry`.
+from the rest of your system.
```
abn, line 20: Not a blocker, but should we link to our documentation here instead of maintaining duplication information here? Might be good for the brevity of the README as well.

install-poetry.py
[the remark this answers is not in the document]
abn: Note `copies` is probably more robust to avoid system python changes, and also in CI environments where the install might be cached.

install-poetry.py
abn: I guess `force` take cares of not recreating and `virtualenv --clear` takes care of recreating the existing environment, I might have overlooked this.

abn (changes requested):
Over all looks great. Couple of suggestions.

One comment on the virtual environment creation options to ensure some resilience for the installation. Otherwise good to go.

naveen521kk:
I tried this script and it failed
```pwsh
C:\Users\User> (Invoke-WebRequest -Uri https://raw.githubusercontent.com/python-poetry/poetry/8a506eceb90e7b7d1bc1cd837d303c32e8b03bec/install-poetry.py
-UseBasicParsing).Content | python -
Retrieving Poetry metadata

# Welcome to Poetry!

This will download and install the latest version of Poetry,
a dependency and package manager for Python.

C:\Users\User\AppData\Roaming\Python\Scripts

You can uninstall at any time by executing this script with the --uninstall opti
on,
and these changes will be reverted.

Installing Poetry (1.1.5)
Installing Poetry (1.1.5): Creating environment
  File "<stdin>", line 808, in <module>
  File "<stdin>", line 804, in main
  File "<stdin>", line 417, in run
  File "<stdin>", line 439, in install
  File "<stdin>", line 503, in make_env
ModuleNotFoundError: No module named 'virtualenv'
```

install-poetry.py
```diff
@@ -0,0 +1,808 @@
...
+                colorize("b", version),
+                colorize("comment", "Creating environment"),
+            )
+        )
+
+        env_path = self._data_dir.joinpath("venv")
+
+        with temporary_directory() as tmp_dir:
+            subprocess.call(
+                [sys.executable, "-m", "pip", "install", "virtualenv", "-t", tmp_dir],
```
abn, line 496: We should do a check for pip, as pip is not always present. If it's not present we have one of two options.

1. Automatically execute `python -m ensurepip`. 
2. Error out, instructing user to install pip and hint at 1, since they may want to use distro specific installation. 

Alternatively, `virtualenv` wheel should be usable without installation I think. That would make a cleaner approach.

abn (changes requested):

.github/workflows/main.yml
abn: Related: https://github.com/python-poetry/poetry-core/pull/144/commits/81fc0d42d806583096392a822202483e85e1b4e8

install-poetry.py
abn: https://github.com/python-poetry/poetry/issues/3843

install-poetry.py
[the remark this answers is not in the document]
abn: This will become less of an issue once https://github.com/python-poetry/poetry/issues/3843 is handled.

.github/workflows/main.yml
abn: This can move outside this pull request as it should not block progress on this.

abn (dismissed):
Spun out the improvement regarding virtualenv bootstrapping to https://github.com/python-poetry/poetry/issues/3843

Otherwise looks good to me.

abn:
@sdispater this should be good to go after a rebase and conflict resolution.

abn (approved):

PR #3710, opened 2021-02-20
Typo

vpoulailleau:
Naming after https://github.com/python-poetry/poetry-core

BTW, thanks so much for `poetry`, I :heart: it!

### DIFF
--- a/docs/docs/pyproject.md
+++ b/docs/docs/pyproject.md
@@ -300,4 +300,4 @@ build-backend = "poetry.core.masonry.api"
 !!!note
 
     If your `pyproject.toml` file still references `poetry` directly as a build backend,
-    you should update it to reference `poetry_core` instead.
+    you should update it to reference `poetry-core` instead.


purificant:
`poetry init` generates the following block in version 1.1.4

```
[build-system]
```

Perhaps the docs should also be updated to reflect this along the lines of this PR.

docs/docs/pyproject.md
```diff
@@ -300,4 +300,4 @@ build-backend = "poetry.core.masonry.api"
 !!!note
 
     If your `pyproject.toml` file still references `poetry` directly as a build backend,
-    you should update it to reference `poetry_core` instead.
+    you should update it to reference `poetry-core` instead.
```
purificant, line 303: Should this change also be made on line 291?

vpoulailleau: According to PyPI, yes: https://pypi.org/project/poetry-core/ and https://pypi.org/project/poetry_core/

But, I assume this has to be confirmed by the team of Poetry.

abn:
@vpoulailleau thank you for your contribution here. However, apologies as I came across #3758 before this one and merged that.

Duplicate-of: #3758

FTR: both names are functionally equivalent.

Issue #2867, opened 2020-08-29
Poetry always detects venvs created by mingw-w64-x86_64-python (MSYS2) as broken


daniel-a-melo:
- [X] I am on the [latest](https://github.com/python-poetry/poetry/releases/latest) Poetry version.
- [X] I have searched the [issues](https://github.com/python-poetry/poetry/issues) of this repo and believe that this is not a duplicate.
- [X] If an exception occurs when executing a command, I executed it again in debug mode (`-vvv` option).

- **OS version and name**: Windows 10
- **Poetry version**: 1.0.10
- **Link of a [Gist](https://gist.github.com/) with the contents of your pyproject.toml file**: 

```toml
[tool.poetry]
name = "poetry-demo2"
version = "0.1.0"
description = ""
authors = ["Your Name <you@example.com>"]

[tool.poetry.dependencies]
python = "^3.8"
pendulum = "^2.1.2"

[tool.poetry.dev-dependencies]
pytest = "^5.2"

[build-system]
requires = ["poetry>=0.12"]
build-backend = "poetry.masonry.api"
```

## Issue

When using MSYS2, package `mingw-w64-x86_64-python`  (https://packages.msys2.org/package/mingw-w64-x86_64-python) venvs are created with POSIX-style layout :

```
 daniel@GALILEO ~/A/L/p/C/v/poetry-demo2-raCoCXDb-py3.8> pwd
/c/Users/Daniel/AppData/Local/pypoetry/Cache/virtualenvs/poetry-demo2-raCoCXDb-py3.8
daniel@GALILEO ~/A/L/p/C/v/poetry-demo2-raCoCXDb-py3.8> ls -al
total 5
drwxr-xr-x 1 daniel daniel   0 Aug 29 23:05 ./
drwxr-xr-x 1 daniel daniel   0 Aug 29 23:05 ../
drwxr-xr-x 1 daniel daniel   0 Aug 29 23:06 bin/
drwxr-xr-x 1 daniel daniel   0 Aug 29 23:05 include/
drwxr-xr-x 1 daniel daniel   0 Aug 29 23:05 lib/
-rw-r--r-- 1 daniel daniel 106 Aug 29 23:05 pyvenv.cfg
```

Every subsequent command will display the message "The virtual environment found in <path> seems to be broken" and the venv will be recreated. I

**Steps to reproduce**

1 - Install MSYS2
2 - From MSYS shell, install python mingw64 build : `pacman -S mingw-w64-x86_64-python`
[...]

PR #3713, opened 2021-02-21
utils/envs: fix envs in MSYS2 always broken

danyeaw:
When running poetry in MSYS2 it gives a warning that the virtual environment found seems to be broken messages. This change detects when in MSYS2 in order to use the bin directory for the virtualenv instead of Scripts which is normally used in Windows.

# Pull Request Check List

Resolves: #2867

### DIFF
--- a/poetry/utils/env.py
+++ b/poetry/utils/env.py
@@ -879,9 +879,13 @@ class Env(object):
 
     def __init__(self, path: Path, base: Optional[Path] = None) -> None:
         self._is_windows = sys.platform == "win32"
+        self._is_mingw = sysconfig.get_platform() == "mingw"
 
+        if not self._is_windows or self._is_mingw:
+            bin_dir = "bin"
+        else:
+            bin_dir = "Scripts"
         self._path = path
-        bin_dir = "bin" if not self._is_windows else "Scripts"
         self._bin_dir = self._path / bin_dir
 
         self._base = base or path


danyeaw:
Hi all, could I please get a review on this PR? This has a huge impact on being able to deploy cross platform OSS apps for Windows. Thanks!

poetry/utils/env.py
```diff
@@ -879,9 +879,13 @@ class Env(object):
 
     def __init__(self, path: Path, base: Optional[Path] = None) -> None:
         self._is_windows = sys.platform == "win32"
+        self._is_mingw = sysconfig.get_platform() == "mingw"
```
abn, line 882:
```suggestion
        self._is_mingw = sys.platform == "msys"
```
This should work here, right? Would be good to keep the usage consistent. I am not too familiar with using mingw, would be good to get a confirmation.

poetry/utils/env.py
```diff
@@ -879,9 +879,13 @@ class Env(object):
...
 
+        if not self._is_windows or self._is_mingw:
+            bin_dir = "bin"
+        else:
+            bin_dir = "Scripts"
```
abn, line 887:
```suggestion
        bin = {"win32": "Scripts"}.get(sys.platform, "bin")
```

Would this be slightly better maintainable considering we only care about exceptions?

abn (changes requested):

poetry/utils/env.py
danyeaw: @abn Thanks for the review, I really appreciate it!

No that won't work:
```
$ python
Python 3.8.8 (default, Feb 20 2021, 07:16:03)  [GCC 10.2.0 64 bit (AMD64)] on win32
>>> import sys
>>> print(sys.platform)
win32
```

poetry/utils/env.py
danyeaw: This doesn't work:
```
>>> bin = {"win32": "Scripts"}.get(sys.platform, "bin")
>>> print(bin)
Scripts
```

abn: Given msys seems to only patch `get_platform()` , guess that is expected.

abn (approved):

danyeaw:
Thanks @abn!

PR #3724, opened 2021-02-24
Mitigate exception due to non-matching regex

maschinensehen:
# Pull Request Check List

Resolves: exception when running poetry with -vv or -vvv verbosity

Running poetry update -vv results in the following exception:

  AttributeError

  'NoneType' object has no attribute 'group'

  at /usr/local/lib/python3.7/site-packages/poetry/puzzle/provider.py:699 in debug
      695│
      696│         if message.startswith("fact:"):
      697│             if "depends on" in message:
      698│                 m = re.match(r"fact: (.+?) depends on (.+?) \((.+?)\)", message)
  → 699│                 m2 = re.match(r"(.+?) \((.+?)\)", m.group(1))
      700│                 if m2:
      701│                     name = m2.group(1)
      702│                     version = " ({})".format(m2.group(2))
      703│                 else:

### DIFF
--- a/poetry/puzzle/provider.py
+++ b/poetry/puzzle/provider.py
@@ -711,20 +711,21 @@ class Provider:
         if message.startswith("fact:"):
             if "depends on" in message:
                 m = re.match(r"fact: (.+?) depends on (.+?) \((.+?)\)", message)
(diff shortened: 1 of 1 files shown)


Secrus:
This code is already present in current `master` branch. Closing.

Issue #3729, opened 2021-02-25
Poetry "depends on" debug regex fails to match if dependancy has missing version info


joshuakraitberg:
Happens with `poetry update -vv`, workaround is to just do `poetry update`.

https://github.com/python-poetry/poetry/blob/8312e3f2dbfa126cd311c666fea30656941e1bd3/poetry/puzzle/provider.py#L692-L699

If `message == "fact: xxx (==1.7.0) depends on cryptography"` you'll end up with `m == None` and an `AttributeError` on the next line.

dimbleby:
I don't know in what circumstances the version info would be missing. 

But that this has had no further interest in the last 18 months or so hints that it is either already fixed, or so rare that no-one is hitting it.

Suggest closing this and associated MR.

Secrus:
Closing for reasons mentioned above. Also, no other reports on this were filed. If you stumble over this bug, feel free to open a new issue.

PR #3731, opened 2021-02-25
Updated puzzle debug regex to handle missing version info

joshuakraitberg:
# Pull Request Check List

Resolves: #3729
Related: #3724, #3739

### DIFF
--- a/poetry/puzzle/provider.py
+++ b/poetry/puzzle/provider.py
@@ -710,7 +710,7 @@ class Provider:
 
         if message.startswith("fact:"):
             if "depends on" in message:
-                m = re.match(r"fact: (.+?) depends on (.+?) \((.+?)\)", message)
+                m = re.match(r"fact: (.+?) depends on ((.+?)($| \((.+?)\)))", message)
                 m2 = re.match(r"(.+?) \((.+?)\)", m.group(1))
                 if m2:
                     name = m2.group(1)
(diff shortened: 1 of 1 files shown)


poetry/puzzle/provider.py
```diff
@@ -710,7 +710,7 @@ def debug(self, message: str, depth: int = 0) -> None:
 
         if message.startswith("fact:"):
             if "depends on" in message:
-                m = re.match(r"fact: (.+?) depends on (.+?) \((.+?)\)", message)
+                m = re.match(r"fact: (.+?) depends on ((.+?)($| \((.+?)\)))", message)
```
sdispater, line 713: Since we reference `m.group(2)` and `m.group(3)`, this require more changes.

```suggestion
                m = re.match(r"fact: (.+?) depends on (?:(.+?)($| \((.+?)\)))", message)
```

And we need to change the way we format the message in case `m.group(3)` is empty.

sdispater (changes requested):

poetry/puzzle/provider.py
joshuakraitberg: With the regex as I made it m.group(2) == m.group(3).

So `message == "fact: xxx (==1.7.0) depends on cryptography"`, group 2 and 3 would be `cryptography`.

I have changed to match your request, if group 3 is `None` I put a `*`.  Not sure what you'd like to see here.

I think having group 2 and 3 be equal in this case might be the most logical solution for what the debug log prints.

Secrus:
See the linked issue for closing reasons.

Issue #693, opened 2018-12-01
[Discussion] Poetry to provide a decent hook-based plugin system


webknjaz:
It's a follow-up for https://github.com/sdispater/poetry/pull/672#issuecomment-443363390.

- [x] I have searched the [issues](https://github.com/sdispater/poetry/issues) of this repo and believe that this is not a duplicate.

## Issue

So setuptools have this nice hook points system where they allow replacing certain utils by some third-party implementations.
It can catch up with what's available by iterating through a list of entrypoints with names corresponding to hooks.
This is a fairly popular technique among setuptools-based dists. For example, pytest and tox support having third-party dists being automatically activated and used when installed given they "advertise" their entry points as required by the plugin system. These two tools use https://github.com/pytest-dev/pluggy to achieve this.

Back to Poetry, I think lots of people agree that providing ability to extend such a great tool is a widely wanted feature.
This would allow adding custom dist/package format generators, custom metadata generators etc. They could be used to retrieve long description from some esoteric formats or places converting that into supported markups. This would also make possible having a custom version constructor like popular plugins to setuptools facilitating version sourcing from Git/Hg (#140/#672). Or, another example, exporting a pinned env and direct dependencies into pip-friendly formatted requirements. And who knows, maybe someone will even try out having another resolver replacement?

Any thoughts on how this might look like?

jgirardet:
Hi,
At the time I was searching a way to specify python version at install, I thought a bit about it.

What does we want ?
- add some custom command:
the easiest I think. It just needs to look for new commands in the setup tools entry point and add it in console/application.py

- add some flags/option to existing command:
already more tricky. should it be added for all commands ? I see something like it : a do_plugin_task() a start of `handle` which would apply some change before the builtin code. it isn't something hard except that poetry runs via cleo which define the option in the class docstring. so if you want to add an option to an existing command, it has to be changed at instantiation time. It means parsing the docstring change it and instantiate the "new" command. You need to have a clear logic in it since many plugins could change the same command.

What is different in poetry for doing this ?
Poetry is not installed via pip so using the standard way of entry point maybe not be ok (I'm not sure about it). It should come from a poetry command. someting like : ` poetry plugin install some_plugin` and the plugin could be in pypi.

Should we try to implement this ?
I would be ok, maybe others too but my feeling is that @sdispater since the beginning of the project,  likes to implement himself the new core features of poetry. It's me feeling, I'm not sure about sebastien's mind about it.

[...]

Issue #2270, opened 2020-04-05
Support subprojects in a poetry project


abn:
### Background & Rationale
This request is inspired by [RPM Package Manger](https://rpm.org/)’s capability to build [subpackages](http://ftp.rpm.org/max-rpm/s1-rpm-subpack-spec-file-changes.html) from the same [Spec File](http://ftp.rpm.org/max-rpm/s1-rpm-build-creating-spec-file.html).

Here, I want to propose and discuss replication a version of this capability can be replicated within poetry to allow for simplified user experience for a python project maintainer, especially when either maintaining namespace packages and/or multi-project source trees. While strict project separation is a good thing in most cases, it might not always be the more pragmatic scenario for package maintainers.

For our purposes here, we can refer to each of theses packages as a subproject. And all subprojects are managed under a single poetry project. This means that there is only a single `pyproject.toml` file and a shared project root directory with either a shared source tree or independent source trees (subdirectory) for each subproject.

### Description
Let us consider the scenario of multiple namespace packages being maintained in a single repository with the following structure.

```console
    namespace-project/
    └── src
        └── namespace
            └── package
                ├── one
                │   └── __init__.py
                ├── three
                │   └── __init__.py
                └── two
                    └── __init__.py
```

Note that this will still apply even if different source directories exists within the root directory for each subproject.

Here the intention could be that we want to distribute 3 packages, namely,  `namespace-package-one` , `namespace-package-two` and `namespace-package-three`.

For the purpose of this example, let us assume that `namespace-package-three` depends on `namespace-package-one`. The `pyproject.toml` file could look something like this.

New sections are annotated with comments detailing them and expected behaviour.

```toml
[build-system]
requires = ["poetry-core>=1.0"]
build-backend = "poetry.core.masonry.api"

[tool.poetry]
name = "namespace-package"
version = "1.0.0-alpha.0"
description = ""
authors = [
    "Bender Rodriguez <bender@planetexpress.com>"
]
license = "MIT"
readme = "README.md"
repository = "https://git.planetexpress.com/bender/python-namespace-package"
keywords = []
classifiers = [
    "Intended Audience :: Developers",
    "Operating System :: OS Independent",
    "License :: OSI Approved :: MIT License",
    "Programming Language :: Python :: 3 :: Only",
    "Programming Language :: Python :: 3.8",
]

# this section remains as is, but now specifies shared dependencies
[tool.poetry.dependencies]
python = "^3.8"

[...]

PR #3733, opened 2021-02-26
Implement a plugin system

sdispater:
This PR introduces a plugin system to add commands and features to Poetry more easily.

It uses the standard entrypoints system that exists today and will make it possible to build two kinds of plugins:

- Generic plugins that will make it possible to interact with the `Poetry` instance created by Poetry. This will be useful to, for instance, implement a plugin that will dynamically set the version of the project based on VCS tags and commits.
- Application plugins that will make it possible to add commands to the Poetry CLI.

This PR also provides three new commands to manage plugins:

- `plugin add` to install plugins
- `plugin remove` to uninstall plugins
- `plugin show` to list currently installed plugins.

### Todo

- [x] Tests for the `plugin add` command
- [x] The `plugin remove` command
- [x] The `plugin show` command

## Pull Request Check List

Resolves: #693 

### DIFF
--- a/docs/docs/cli.md
+++ b/docs/docs/cli.md
@@ -566,12 +566,12 @@ poetry plugin add poetry-plugin --dry-run
 
 * `--dry-run`: Outputs the operations but will not execute anything (implicitly enables --verbose).
 
-### `plugin list`
+### `plugin show`
 
-The `plugin list` command lists all the currently installed plugins.
+The `plugin show` command lists all the currently installed plugins.
 
 ```bash
-poetry plugin list
+poetry plugin show
 ```
 
 ### `plugin remove`
--- a/docs/docs/cli.md
+++ b/docs/docs/cli.md
@@ -538,3 +538,46 @@ To only remove a specific package from a cache, you have to specify the cache en
 ```bash
 poetry cache clear pypi:requests:2.24.0
 ```
+
+## plugin
+
+The `plugin` namespace regroups sub commands to manage Poetry plugins.
+
+### `plugin add`
+
+The `plugin add` command installs Poetry plugins and make them available at runtime.
+
+For example, to install the `poetry-plugin` plugin, you can run:
+
+```bash
+poetry plugin add poetry-plugin
+```
+
+The package specification formats supported by the `plugin add` command are the same as the ones supported
+by the [`add` command](#add).
+
+If you just want to check what would happen by installing a plugin, you can use the `--dry-run` option
+
+```bash
+poetry plugin add poetry-plugin --dry-run
+```
+
+#### Options
+
+* `--dry-run`: Outputs the operations but will not execute anything (implicitly enables --verbose).
+
+### `plugin list`
+
+The `plugin list` command lists all the currently installed plugins.
+
+```bash
+poetry plugin list
+```
+
+### `plugin remove`
+
+The `plugin remove` command removes installed plugins.
+
+```bash
+poetry plugin remove poetry-plugin
+```
--- a/docs/docs/plugins.md
+++ b/docs/docs/plugins.md
@@ -1,0 +1,172 @@
+# Plugins
+
+Poetry supports using and building plugins if you wish to
+alter or expand Poetry's functionality with your own.
+
+For example if your environment poses special requirements
+on the behaviour of Poetry which do not apply to the majority of its users
+or if you wish to accomplish something with Poetry in a way that is not desired by most users.
+
+In these cases you could consider creating a plugin to handle your specific logic.
+
+
+## Creating a plugin
+
+A plugin is a regular Python package which ships its code as part of the package
+and may also depend on further packages.
+
+### Plugin package
+
+The plugin package must depend on Poetry
+and declare a proper [plugin](/docs/pyproject/#plugins) in the `pyproject.toml` file.
+
+```toml
+[tool.poetry]
+name = "my-poetry-plugin"
+version = "1.0.0"
+
+# ...
+[tool.poetry.dependency]
+python = "~2.7 || ^3.7"
+poetry = "^1.0"
+
+[tool.poetry.plugins."poetry.plugin"]
+demo = "poetry_demo_plugin.plugin:MyPlugin"
+```
+
+### Generic plugins
+
+Every plugin has to supply a class which implements the `poetry.plugins.Plugin` interface.
+
+The `activate()` method of the plugin is called after the plugin is loaded
+and receives an instance of `Poetry` as well as an instance of `cleo.io.IO`.
+
+Using these two objects all configuration can be read
+and all public internal objects and state can be manipulated as desired.
+
+Example:
+
+```python
+from cleo.io.io import IO
+
+from poetry.plugins.plugin import Plugin
+from poetry.poetry import Poetry
+
+
+class MyPlugin(Plugin):
+
+    def activate(self, poetry: Poetry, io: IO):
+        version = self.get_custom_version()
+        io.write_line(f"Setting package version to <b>{version}</b>")
+        poetry.package.set_version(version)
+
+    def get_custom_version(self) -> str:
+        ...
+```
+
+### Application plugins
+
+If you want to add commands or options to the `poetry` script you need
+to create an application plugin which implements the `poetry.plugins.ApplicationPlugin` interface.
+
+The `activate()` method of the application plugin is called after the plugin is loaded
+and receives an instance of `console.Application`.
+
+```python
+from cleo.commands.command import Command
+from poetry.plugins.application_plugin import ApplicationPlugin
+
+
+class CustomCommand(Command):
+    
+    name = "my-command"
+    
+    def handle(self) -> int:
+        self.line("My command")
+        
+        return 0
+
+        
+def factory():
+    return CustomCommand()
+
+
+class MyApplicationPlugin(ApplicationPlugin):
+    def activate(self, application):
+        application.command_loader.register_factory("my-command", factory)
+```
+
+!!!note
+
+    It's possible to do the following to register the command:
+
+    ```python
+    application.add(MyCommand())
+    ```
+
+    However, it is **strongly** recommended to register a new factory
+    in the command loader to defer the loading of the command when it's actually
+    called.
+
+    This will help keep the performances of Poetry good.
+
+The plugin also must be declared in the `pyproject.toml` file of the plugin package
+as an `application.plugin` plugin:
+
+```toml
+[tool.poetry.plugins."poetry.application.plugin"]
+foo-command = "poetry_demo_plugin.plugin:MyApplicationPlugin"
+```
+
+!!!warning
+
+    A plugin **must not** remove or modify in any way the core commands of Poetry.
+
+
+### Event handler
+
+Plugins can also listen to specific events and act on them if necessary.
+
+There are two types of events: application events and generic events.
+
+These events are fired by [Cleo](https://github.com/sdispater/cleo)
+and are accessible from the `cleo.events.console_events` module.
+
+- `COMMAND`: this event allows attaching listeners before any command is executed.
+- `SIGNAL`: this event allows some actions to be performed after the command execution is interrupted.
+- `TERMINATE`: this event allows listeners to be attached after the command.
+- `ERROR`: this event occurs when an uncaught exception is raised.
+
+Let's see how to implement an application event handler. For this example
+we will see how to load environment variables from a `.env` file before executing
+a command.
+
+
+```python
+from cleo.events.console_events import COMMAND
+from cleo.events.console_command_event import ConsoleCommandEvent
+from cleo.events.event_dispatcher import EventDispatcher
+from dotenv import load_dotenv
+from poetry.console.application import Application
+from poetry.console.commands.env_command import EnvCommand
+from poetry.plugins.application_plugin import ApplicationPlugin
+
+
+class MyApplicationPlugin(ApplicationPlugin):
+    def activate(self, application: Application):
+        application.event_dispatcher.add_listener(COMMAND, self.load_dotenv)
+
+    def load_dotenv(
+        self, event: ConsoleCommandEvent, event_name: str, dispatcher: EventDispatcher
+    ) -> None:
+        command = event.io
+        if not isinstance(command, EnvCommand):
+            return
+        
+        io = event.io
+
+        if io.is_debug():
+            io.write_line("<debug>Loading environment variables.</debug>")
+
+        load_dotenv()
+```
--- a/docs/docs/plugins.md
+++ b/docs/docs/plugins.md
@@ -199,7 +199,7 @@ poetry plugin remove poetry-plugin
 You can also list all currently installed plugins by running:
 
 ```shell
-poetry plugin list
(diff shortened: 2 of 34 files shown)


docs/docs/cli.md
```diff
@@ -534,3 +534,46 @@ To only remove a specific package from a cache, you have to specify the cache en
...
+
+The `plugin list` command lists all the currently installed plugins.
+
+```bash
+poetry plugin list
+```
+
+### `plugin remove`
+
+The `plugin remove` command lists all the currently installed plugins.
```
s0undt3ch, line 575: I know it's in early stages, but;

Removes a plugin.... ?

docs/docs/plugins.md
```diff
@@ -0,0 +1,233 @@
...
+```shell
+poetry plugin list
+```
+
+### With `pipx inject`
+
+If you used `pipx` to install Poetry you can add the plugin packages via the `pipx inject` command.
+
+```shell
+pips inject poetry poetry-plugin
```
s0undt3ch, line 212:
```suggestion
pipx.inject poetry poetry-plugin
```

cognifloyd (commented):
Ooh. This is exciting [WIP] development. Looking forward to it!

Are plugins prohibited with `poetry new`? Not sure why people would need them in that context, but 🤷.

poetry/console/application.py
```diff
@@ -272,6 +298,45 @@ def set_installer(
...
+        name = self._get_command_name(io)
+        command_name = ""
+        if name:
+            try:
+                command_name = self.find(name).name
+            except CommandNotFoundException:
+                pass
+
+        self._disable_plugins = (
+            io.input.has_parameter_option("--no-plugins") or command_name == "new"
```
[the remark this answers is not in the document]
sdispater: I think it's a leftover restriction from a previous shot at implementing the plugin system.

The current implementation should not need such restriction.

I'll check and remove it if needed.

docs/docs/cli.md
```diff
@@ -534,3 +534,46 @@ To only remove a specific package from a cache, you have to specify the cache en
...
+```bash
+poetry plugin list
+```
+
+### `plugin remove`
+
+The `plugin remove` command removes installed plugins.
+
+```bash
+poetry plugin remove poetry-plugin
```
1ace, line 578: Ditto, `s/poetry-plugin/my-plugin/`

1ace: And several other occurrences actually, I'm not going to post a note each time ^^

1ace:
I interpret your docs to mean "yes", but can you confirm whether it's possible with this system to have a repo-specific plugin that lives in the repo itself? or does it need to be installed in one of the sites for poetry to see it?

That would be a really useful feature I think :)

sdispater:
@1ace Plugins will be global and not on a per-project basis.

docs/docs/plugins.md
```diff
@@ -0,0 +1,233 @@
...
+as an `application.plugin` plugin:
+
+```toml
+[tool.poetry.plugins."poetry.application.plugin"]
+foo-command = "poetry_demo_plugin.plugin:MyApplicationPlugin"
+```
+
+!!!warning
+
+    A plugin **must not** remove or modify in any way the core commands of Poetry.
```
cognifloyd, line 123: What do you mean by "modify"?
Could a plugin, for example, implement something like #2270 by adjusting `build`/`publish` to create/push one wheel per package? I'm guessing "no".

Are you thinking that a plugin that wants to implement #2270 would actually add new commands like `poetry subpackages build` and `poetry subpackages publish`? And then those commands would be able to reuse the same poetry machinery used by `poetry build` and `poetry publish`?

cognifloyd (changes requested):
So excited to use this!

Could you add a note to the docs to say that plugins do not affect `poetry new`?

cognifloyd (approved):
:tada:

docs/docs/plugins.md
```diff
@@ -0,0 +1,233 @@
...
+
+
+### Event handler
+
+Plugins can also listen to specific events and act on them if necessary.
+
+There are two types of events: application events and generic events.
```
stephsamson, line 130: I think more information should be supplied on the difference(s) between application events and generic events.

docs/docs/cli.md
```diff
@@ -538,3 +538,46 @@ To only remove a specific package from a cache, you have to specify the cache en
...
+
+* `--dry-run`: Outputs the operations but will not execute anything (implicitly enables --verbose).
+
+### `plugin list`
+
+The `plugin list` command lists all the currently installed plugins.
+
+```bash
+poetry plugin list
+```
```
stephsamson, line 575: Should this not be `plugin show`? And if not, then other places in the PR must be updated to use the command `list` and not `show`.

poetry/packages/locker.py
```diff
@@ -595,3 +595,19 @@ def _dump_package(self, package: Package) -> dict:
...
+
+class NullLocker(Locker):
+    def __init__(self, locked=False):  # type: (bool) -> None
+        self._locked = locked
+
+    def is_locked(self):  # type: () -> bool
+        return self._locked
+
+    def set_lock_data(self, root, packages):  # type: (...) -> bool
+        return True
```
stephsamson, line 608: What is the value of this method if it always `return True` and has no other side effects?

docs/docs/plugins.md
sdispater: This was a leftover from the previous attempt at implementing the plugin system. I removed it.

docs/docs/cli.md
sdispater: It's actually `show` :-) I fixed it.

poetry/packages/locker.py
sdispater: The `NullLocker` was a leftover from the previous attempt at implementing the plugin system. I removed it.

stephsamson (dismissed):
🚢

docs/docs/cli.md
```diff
@@ -538,3 +538,46 @@ To only remove a specific package from a cache, you have to specify the cache en
...
+#### Options
+
+* `--dry-run`: Outputs the operations but will not execute anything (implicitly enables --verbose).
+
+### `plugin show`
+
+The `plugin show` command lists all the currently installed plugins.
+
+```bash
+poetry plugin list
```
alejandro-angulo, line 574: Should this be `poetry plugin show`? If not, looks like there's a header missing for `plugin list`

stephsamson: Yeah it should be `poetry plugin show`.

cognifloyd (approved):

poetry/console/commands/plugin/add.py
```diff
@@ -0,0 +1,263 @@
...
+        # We create a `pyproject.toml` file based on all the information
+        # we have about the current environment.
+        self.create_pyproject_from_package(root_package, env_dir)
+
+        # We add the plugins to the dependencies section of the previously
+        # created `pyproject.toml` file
+        pyproject = tomlkit.loads(
+            env_dir.joinpath("pyproject.toml").read_text(encoding="utf8")
+        )
+        poetry_dependency_section = pyproject["tool"]["poetry"]["dependencies"]
```
abn, line 141: We should really use the pyproject classes here and avoid reimplementing certain things. Would be great to see that be the pattern for cases where we interact with the pyproject file.

poetry/console/commands/plugin/add.py
```diff
@@ -0,0 +1,263 @@
...
+
+        return update_command.run(
+            IO(
+                StringInput(" ".join(argv)),
+                self._io.output,
+                self._io.error_output,
+            )
+        )
+
+    def create_pyproject_from_package(
```
abn, line 190: Shoud this go into factory?

poetry/console/commands/plugin/add.py
```diff
@@ -0,0 +1,263 @@
...
+            if len(constraint) == 1 and "version" in constraint:
+                constraint = constraint["version"]
+
+            dependency_section[dep.name] = constraint
+
+        path.joinpath("pyproject.toml").write_text(
+            pyproject.as_string(), encoding="utf-8"
+        )
+
+    def get_existing_packages_from_input(
```
abn, line 238: This might be better served as a reusable logic I feel

poetry/console/commands/plugin/add.py
```diff
@@ -0,0 +1,263 @@
...
+        # If not, it might lead to side effects since other installed packages
+        # might not be required by Poetry but still taken into account when resolving dependencies.
+        installed_repository = InstalledRepository.load(
+            system_env, with_dependencies=True
+        )
+        repository = Repository()
+
+        root_package = None
+        for package in installed_repository.packages:
+            if package.name in Provider.UNSAFE_PACKAGES:
```
abn, line 117: Note that we no longer consider pip or setuptools unsafe.

poetry/packages/project_package.py
```diff
@@ -0,0 +1,24 @@
...
+    from typing import Optional
+    from typing import Union
+
+    from poetry.core.semver.version import Version  # noqa
+
+
+class ProjectPackage(_ProjectPackage):
+    def set_version(
+        self, version, pretty_version=None
+    ):  # type: (Union[str, Version], Optional[str]) -> ProjectPackage
```
abn, line 16: This can move to new type hinting syntax.

.coveragerc
```diff
@@ -0,0 +1,4 @@
+[report]
```
abn, line 1: This should really move into pyproject. https://github.com/abn/cookiecutter-python-project/blob/732bf8e890da9847dca407d2f30f254fd963bdca/%7B%7B%20cookiecutter.directory_name%20%7D%7D/pyproject.toml#L78-L125

poetry/console/commands/command.py
```diff
@@ -11,9 +12,17 @@
...
+
     @property
     def poetry(self) -> "Poetry":
-        return self.get_application().poetry
+        if self._poetry is None:
+            return self.get_application().poetry
+
+        return self._poetry
+
+    def set_poetry(self, poetry: "Poetry") -> None:
```
abn, line 24: Should this be a property setter?

poetry/console/commands/init.py
```diff
@@ -430,20 +430,32 @@ def _parse_requirements(self, requirements: List[str]) -> List[Dict[str, str]]:
 
                     result.append(pair)
                     continue
-            elif (os.path.sep in requirement or "/" in requirement) and cwd.joinpath(
-                requirement
-            ).exists():
-                path = cwd.joinpath(requirement)
+            elif (os.path.sep in requirement or "/" in requirement) and (
```
abn, line 433: We should take this opportunity to move this bit to pathlib.

abn (changes requested):
Added some minor review comments. I think this looks really exciting.

As for application plugins, how do you see us adding things like export formats? It seems we have to add new commands for this and not extend existing commands. This might be that `export -f requirements.txt` will become `export requirements.txt --option`, which a-okay - just was not sure from the doc how sub comands can be added.

poetry/console/commands/plugin/add.py
sdispater: Changed :-)

poetry/console/commands/plugin/add.py
sdispater: I moved it to the factory

poetry/console/commands/plugin/add.py
sdispater: Honestly, I don't think it warrants separating it from the command. It's a small function so I don't mind leaving it there.

poetry/console/commands/plugin/add.py
sdispater: Removed :-)

poetry/packages/project_package.py
sdispater: Done!

.coveragerc
sdispater: Done!

poetry/console/commands/init.py
sdispater: Most of the code here is using pathlib so I am not sure what you mean

poetry/console/commands/command.py
sdispater: I prefer to have explicit setters for the public API. It feels clearer this way.

sdispater:
@abn Regarding the `export` command, there are two ways to go:

- Make use of subcommands rather than a `--format` option.
- Make the `export` command pluggable so that the `--format` option can be hooked into.

Once we make it an official plugin we'll see how it goes. For now, I don't have a strong preference even though I feel like keeping the option makes more sense.

As to how sub commands can be added, it's simple since subcommands are nothing more than namespaced commands. So basically, it woould look like this:

```python
from cleo.commands.command import Command
from poetry.plugins.application_plugin import ApplicationPlugin


class ExportRequirementsCommand(Command):

    name = "export requirements.txt"

    def handle(self) -> int:
        # Do the export

        return 0


def export_command_factory():
    return ExportRequirementsCommand()


class ExportApplicationPlugin(ApplicationPlugin):
    def activate(self, application):
        application.command_loader.register_factory(
            "export requirements.txt", export_command_factory
        )
```

docs/docs/cli.md
sdispater: Fixed!

poetry/console/commands/init.py
abn: Oh I was referring to removing the `(os.path.sep in requirement or "/" in requirement)` check entirely as `Path.exists()` should handle this. 

Also out of interest, pathlib does not seem to handle checks correct for file URIs. We might need to handle that separately (outside of this PR).

```console
>>> pathlib.Path("file://tmp").exists()
False
>>> pathlib.Path("file:/tmp").exists()
False
```

sdispater: It's done like so to check early if the requirement looks like a path before trying to check if it exists

abn: Not blocking on this as it's an existing logic. Can deal with this outside the pr.

abn (approved):

henryiii:
Just curious, would it be possible to add an extension building system via a plugin? So poetry handles the Python files and CMake (for example) handles the compiling? I assume a little extra work would be needed to ensure wheels has the right names and such, but how close might this be to working?

abn:
@henryiii If I am understanding you correct, I suspect you can achieve that today using the build script without a plugin. https://github.com/sdispater/pendulum/blob/411d0aa41a5f39c5f4a2f43a3c369c2dd24787db/pyproject.toml#L51-L53

You can add the cmake call into the `build.py`.  If this is not the case, I would suggest you open a discussion.

henryiii:
Thanks @abn, that's good to know about. (Though, @sdispater, two notes; just to keep in mind, distutils is slated for removal in Python 3.12, and you shouldn't name a top-level file `build.py`, because then you can't build the project with `python -m build`, I've had to debug this before).

I'll probably open a discussion, because I'm interested in developing a scikit-build plugin that would integrate CMake with Poetry.

mpeteuil:
I was wondering if there is a recommended naming pattern for plugins? Like `poetry_plugin_X` or something.

PR #3736, opened 2021-02-26
Export setup py

JasonAtallah:
# Pull Request Check List

Resolves: #761

### DIFF
--- a/docs/docs/cli.md
+++ b/docs/docs/cli.md
@@ -485,14 +485,10 @@ This command exports the lock file to other formats.
 poetry export -f requirements.txt --output requirements.txt
 ```
 
-!!!note
-
-    Only the `requirements.txt` format is currently supported.
-
 ### Options
 
 * `--format (-f)`: The format to export to (default: `requirements.txt`).
-  Currently, only `requirements.txt` is supported.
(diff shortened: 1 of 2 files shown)


abn:
Thank you for your contribution @JasonAtallah. Unfortunately, this is not a feature that will be supported in poetry. The linked pull request should provide more context. I will update #761 as well.

Duplicate-of: #762.

JasonAtallah:
> Thank you for your contribution @JasonAtallah. Unfortunately, this is not a feature that will be supported in poetry. The linked pull request should provide more context. I will update #761 as well.
> 
> Duplicate-of: #762.

@abn Why will this not be supported? There is no new functionality, it's just reusing existing methods. I know personally this would be a great addition as not everyone in my organization uses Poetry.

abn:
I can understand the need in your scenario, however since the existing code is being dropped in a future release (as indicated in my last comment on #761), it does not make sense to keep it just to support this use case. Additionally, the generated setup file is not necessarily complete anymore as Poetry's supported features have outgrown it.

If the community has an interest in this feature, it could be implemented as an export format plug-in. See #3733 for progress on that front.

PR #3758, opened 2021-03-06
Fix reference to `poetry-core` package

jodal:
The distribution is named `poetry-core`, not `poetry_core`.


abn (approved):
This should not cause an issue since `poetry_core` is the canonical form of `poetry-core`. However, having this consistent is good.

MinchinWeb:
When can this documentation update be pushed live?

PR #3811, opened 2021-03-20
POETRY_EXPERIMENTAL_NEW_INSTALLER is interpreted as a boolean

AWhetter:
# Pull Request Check List

The `POETRY_EXPERIMENTAL_NEW_INSTALLER` environment variable is not currently interpreted as a boolean value. This means that you have to set the value to an empty string to disable it.
This change adds the option to the list of options that are interpreted as booleans so that you can also disable it by setting it as anything other than "1" or "true".

### DIFF
--- a/tests/config/test_config.py
+++ b/tests/config/test_config.py
@@ -1,6 +1,21 @@
 import os
+import re
 
 import pytest
+
+from poetry.config.config import Config
+
+
+def get_boolean_options(config=None):
+    if config is None:
+        config = Config.default_config
+
+    for k, v in config.items():
+        if isinstance(v, bool) or v is None:
+            yield k
+        if isinstance(v, dict):
+            for suboption in get_boolean_options(v):
+                yield "{}.{}".format(k, suboption)
 
 
 @pytest.mark.parametrize(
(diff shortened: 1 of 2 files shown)


poetry/config/config.py
```diff
@@ -140,6 +140,7 @@ def _get_normalizer(self, name: str) -> Callable:
             "virtualenvs.create",
             "virtualenvs.in-project",
             "virtualenvs.options.always-copy",
+            "experimental.new-installer",
             "installer.parallel",
         }:
```
AWhetter, line 145: I decided to make sure that this list is kept correct in the future by having the tests generate a complete list of options that should be interpreted as booleans and validating each one. However it would be just as valid to generate the list here (or do it at the class level so that it happens once) and then hardcoding a few values to check in the tests.
Doing it this way would mean that we could hand pick test cases that we know cover a lot of conditions and check that the environment variable names are correct . For example

```python
@pytest.mark.parametrize(
    ("name", "env_var", "env_value", "value"),
    [
        ("installer.parallel", "POETRY_INSTALLER_PARALLEL", "true", True),
        ("installer.parallel", "POETRY_INSTALLER_PARALLEL", "false", False),
        ("virtualenvs.in-project", "POETRY_VIRTUALENVS_IN_PROJECT", "true", True),  # Using this poetry option makes sure that we replace both dots and dashes
        ("virtualenvs.in-project", "POETRY_VIRTUALENVS_IN_PROJECT", "false", False),
    ],
)
def test_config_get_from_environment_variable(...):
    ...
``` 

Let me know if you would prefer me to switch to using this approach instead.

tests/config/test_config.py
```diff
@@ -17,13 +32,12 @@ def test_config_get_processes_depended_on_values(config, config_cache_dir):
...
-        ("virtualenvs.create", "true", True),
-        ("virtualenvs.create", "false", False),
+        (name, env_value, value)
+        for name in get_boolean_options()
+        for env_value, value in [("true", True), ("false", False)]
     ],
 )
 def test_config_get_from_environment_variable(config, environ, name, env_value, value):
-    env_var = "POETRY_{}".format("_".join(k.upper() for k in name.split(".")))
+    env_var = "POETRY_{}".format(re.sub("[.-]+", "_", name).upper())
```
abn, line 41: I would move this into the generation of the tupes as this would mean we can see what we are testing in the test case name.

abn (dismissed):
Minor recommendation for tests.

AWhetter:
Thanks for the fast review! I've made the suggested changes.

neersighted (changes requested):
Both the fix and the test changes look good. I think that ultimately your approach of using a generator is the better, even if it adds a bit to the mental overhead when maintaining the test.

Could you please break up your changes into two comments with good commit messages? LGTM once that's done.

neersighted (approved):
Opted to merge as-is.

Issue #2543, opened 2020-06-11
PEP 440 version matching of local version identifier clause (e.g. 1.2.3+local) not implemented correctly in solver


Korijn:
- [x] I am on the [latest](https://github.com/python-poetry/poetry/releases/latest) Poetry version.
- [x] I have searched the [issues](https://github.com/python-poetry/poetry/issues) of this repo and believe that this is not a duplicate.
- [x] If an exception occurs when executing a command, I executed it again in debug mode (`-vvv` option).

- **OS version and name**: Windows 10, but also in a Linux docker container running Ubuntu 18.04
- **Poetry version**: 1.0.9
- **Link of a [Gist](https://gist.github.com/) with the contents of your pyproject.toml file**: https://gist.github.com/Korijn/b88b1605221574f51ee72b75efc03dbe

## Issue

In [PEP 440](https://www.python.org/dev/peps/pep-0440/#version-matching), it says the following in the section "Version Matching":

> If the specified version identifier is a public version identifier (no local version label), then the local version label of any candidate versions MUST be ignored when matching versions.

I believe this clause is being interpreted incorrectly by Poetry. When running `poetry lock` on a `pyproject.toml` file (see the gist linked above) with the folowing lines, poetry incorrectly reports a conflict:

```
torch = "1.4.0+cpu"
torchvision = "0.5.0+cpu"
```

```
   1: conflict: torchvision (0.5.0+cpu) depends on torch (1.4.0)
   1: ! not torch (1.4.0) is satisfied by torch (1.4.0+cpu)
   1: ! which is caused by "bla depends on torch (1.4.0+cpu)"
   1: ! thus: torchvision is forbidden
   1: ! torchvision (0.5.0+cpu) is satisfied by torchvision (0.5.0+cpu)
   1: ! which is caused by "bla depends on torchvision (0.5.0+cpu)"
   1: ! thus: version solving failed
```

As said in the first line of the debug output, the package torchvision (0.5.0+cpu) depends on torch (1.4.0). However, I've specified that I need torch (1.4.0+cpu). The PEP says that the local identifier MUST be ignored when performing the version match for the constraint coming from the torchvision (0.5.0+cpu) package. So I would say that this conflict is actually an acceptable solution, given the constraints.

Note that Pip has no problem installing this combination of packages.

This is an issue with Poetry's solver.

stadlerb:
Same problem here - is there any way to disable URL filename version checking?

Jefffish09:
[...]

Issue #1151, opened 2019-06-05
Poetry incorrectly normalizes dev builds as alpha builds


epage:
- [x] I am on the [latest](https://github.com/sdispater/poetry/releases/latest) Poetry version.
- [x] I have searched the [issues](https://github.com/sdispater/poetry/issues) of this repo and believe that this is not a duplicate.
- [ ] If an exception occurs when executing a command, I executed it again in debug mode (`-vvv` option).

- **OS version and name**: Windows 10
- **Poetry version**: 0.12.16
- **Link of a [Gist](https://gist.github.com/) with the contents of your pyproject.toml file**: 

## Issue

While investigating #1150, I noticed that when [normalizing pre-release](https://github.com/sdispater/poetry/blob/master/poetry/semver/version.py#L295), `dev` is translated to `alpha` when [PEP 440](https://www.python.org/dev/peps/pep-0440/) qualifies these as distinct stages.  Doing this will cause problems when installing packages that use both stages of development (which we do internally).

epage:
As a user, this is a bit frustrating for issues to be automatically closed when the problem is unresponsiveness from the maintainers.  I would have gone in and fixed this except I already saw that my other PRs were sitting around, not making any progress.

saisasidhar:
@epage could you please add an appropriate label to this issue so the maintainers could have a look at it ?

epage:
I can't apply labels.

Secrus:
This appears to be fixed now.

Issue #892, opened 2019-02-13
Wrong version installed when using PEP 440 version epochs


jacebrowning:
- [x] I am on the [latest](https://github.com/sdispater/poetry/releases/latest) Poetry version.
- [x] I have searched the [issues](https://github.com/sdispater/poetry/issues) of this repo and believe that this is not a duplicate.
- [x] If an exception occurs when executing a command, I executed it again in debug mode (`-vvv` option).

- **OS version and name**: macOS 10.14.3
- **Poetry version**: 0.12.11
- **Link of a [Gist](https://gist.github.com/) with the contents of your pyproject.toml file**: N/A

## Issue

`1!2019.2.12` is requested but `1!2015.2.1` is installed:
```
$ poetry new demo
$ cd demo
$ poetry add "foursquare==1\!2019.2.12"
Creating virtualenv demo-py3.7 in /private/tmp/demo/.venv

Updating dependencies
Resolving dependencies... (0.2s)
Package operations: 13 installs, 0 updates, 0 removals

Writing lock file

  - Installing certifi (2018.11.29)
  ...
  - Installing foursquare (1!2015.2.1)
  ...
```

I see code that seems to support this, but no tests: https://github.com/sdispater/poetry/search?q=epoch&unscoped_q=epoch

jacebrowning:
I believe this issue is still valid and should be treated as a bug.

JBKahn:
The semver Version doesn't support it, but the version Version parser does. It seems a bit strange to have two different parsers, I think if the semver Parser only parsed the constraint out but not the versions that could work? @sdispater how do you think this should be structured?

PR #3831, opened 2021-03-25
poetry-core update with lark pep440 implementation

abn:
This change consumes the upstream changes to semver `Version` class implementation.

Relates-to: python-poetry/poetry-core#140
Relates-to: python-poetry/poetry#2543
Relates-to: python-poetry/poetry#1151
Relates-to: python-poetry/poetry#892

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -134,7 +134,7 @@ python-versions = ">=3.6,<4.0"
 
 [[package]]
 name = "cryptography"
-version = "3.4.6"
+version = "3.4.7"
 description = "cryptography is a package which provides cryptographic recipes and primitives to Python developers."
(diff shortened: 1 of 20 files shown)


stephsamson (approved):

pinarkavak:
Why is this PR not put in the release 1.1.6?

abn:
@pinarkavak this is available in `1.2.0a1` and was not back ported to the `1.1.z` stream as these only support bug fixes.

Also, do note that the lark usage was reverted in (https://github.com/python-poetry/poetry-core/pull/152) due to performance concerns. However, the improvements to pep440 are still retained.

pinarkavak:
Thank you.

PR #3835, opened 2021-03-25
pip installer: fix installed script executable

abn:
When pip installs into an alternate prefix, it uses the current executable
when writing package scripts during install (eg: black, pytest etc.).

With this change, we:
1. Use the active environment to trigger pip embedded virtual environment.
2. Enforcing the use of `--pep-517` when installing non-wheel packages.
3. Handle zip-unsafe bug in pip's vendored pep517 under python3.6.


apollo13 (approved):
Tested locally and will ship it as part of our CI poetry to provide more testing.

stephsamson (approved):

PR #3838, opened 2021-03-26
Minor clean up of code for consistency and performance

abn:
1. Make `ephemeral_environment` arguments aligned with `build_env`.
2. Disable pip version check when when using `pip_install`.


poetry/utils/pip.py
```diff
@@ -20,7 +20,7 @@ def pip_install(
     path = Path(path) if isinstance(path, str) else path
     is_wheel = path.suffix == ".whl"
 
-    args = ["install", "--prefix", str(environment.path)]
+    args = ["install", "--disable-pip-version-check", "--prefix", str(environment.path)]
```
stephsamson, line 23: Can we add a short comment about why are disable pip version check here?

stephsamson (approved):

PR #3846, opened 2021-03-27
env: do not modify os.environ

abn:
Replace updates of os.environ with explcit passing of `env` to
subprocess calls.

Resolves: #3199
Forwardport-of: #3253

### DIFF
--- a/poetry/utils/env.py
+++ b/poetry/utils/env.py
@@ -11,6 +11,7 @@ import sysconfig
 import textwrap
 
 from contextlib import contextmanager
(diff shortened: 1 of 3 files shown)


stephsamson (approved):

max-wittig:
Thanks for this awesome fix! Then we can finally use the new installer and could finally get rid of this line:

> poetry config experimental.new-installer false

Is a bugfix release planned in the next weeks?

Thanks for the work, you're doing!

Issue #3842, opened 2021-03-27
ci: add support for python 3.10 experimental builds


abn:
We should implement support for 3.10 CI builds in our pipelines for testing. This would be easily done once the new bootstrap ( is completed (#3706).

This was orginally raised in #3828.

> Btw, maybe it'll be a great idea to enable Python 3.10 in your pipeline too, you just need to allow failure in that step while the official version isn't released yet! For more information [check this out](https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#jobsjob_idcontinue-on-error)

See similar work done in `poetry-core`: https://github.com/python-poetry/poetry-core/pull/144.

PR #3859, opened 2021-03-30
ci: add support for python 3.10 experimental builds

cereblanco:
# Pull Request Check List

Resolves: https://github.com/python-poetry/poetry/issues/3842

- Updated CIs for tests and skip.yaml
- I used `pip to install poetry` because Windows envs encounter error using `python install-poetry.py -y`

~~[ ] Added **tests** for changed code.~~
~~[ ] Updated **documentation** for changed code.~~

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -8,6 +8,7 @@ on:
     branches:
       - master
       - develop
+      - add-python-3-10-to-ci
   pull_request:
     paths-ignore:
       - 'docs/**'
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -18,11 +18,17 @@ on:
 jobs:
   tests:
     name: ${{ matrix.os }} / ${{ matrix.python-version }}
-    runs-on: ${{ matrix.os }}-latest
+    runs-on: ubuntu-latest
+    continue-on-error: ${{ matrix.experimental }}
     strategy:
       matrix:
         os: [Ubuntu, MacOS, Windows]
         python-version: [3.6, 3.7, 3.8, 3.9]
+        experimental: [false]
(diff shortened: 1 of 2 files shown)


.github/workflows/main.yml
```diff
@@ -18,11 +18,17 @@ on:
 jobs:
   tests:
     name: ${{ matrix.os }} / ${{ matrix.python-version }}
-    runs-on: ${{ matrix.os }}-latest
+    runs-on: ubuntu-latest
```
abn, line 21: This should use `matrix.os` else we end up running all tests on linux.

.github/workflows/main.yml
```diff
@@ -39,7 +45,7 @@ jobs:
 
       - name: Bootstrap poetry
         shell: bash
-        run: python install-poetry.py -y
+        run: pip install poetry
```
abn, line 48: This can remain using the install script.

.github/workflows/skip.yml
```diff
@@ -29,9 +29,15 @@ jobs:
     name: ${{ matrix.os }} / ${{ matrix.python-version }}
```
abn, line 29: We shouldn't need changes here as we use this to trick the mandatory action checks. And this won't be mandatory until 3.10 is stable.

abn (changes requested):
Thank you @cereblanco , added some minor comments.

cereblanco:
> What was the error on using the install script? We should probably fix that.

@abn i reverted the unnecessary changes.. 
the error was caused by `runs-on: ubuntu-latest` instead of `matrix.os`  🤦🏼 😅 which i reverted as well

thank you

abn (approved):

PR #3867, opened 2021-03-31
Properly detect if packages are a git installation over directory installation

wakemaster39:
# Pull Request Check List

Resolves: #2921

This PR supersedes my original work with  #2327 to be aligned with the updated code base

There are no tests around the changed code for me to base anything off of for me to get started. 

I found no documentation around this, but this is a change in behaviour so not sure where it needs to be documented.

### DIFF
--- a/poetry/repositories/installed_repository.py
+++ b/poetry/repositories/installed_repository.py
@@ -69,6 +69,11 @@ class InstalledRepository(Repository):
                                 # this is required to handle pathlib oddity on win32 python==3.5
                                 path = lib.joinpath(path)
                         paths.add(path)
+
+        src_path = env.path / "src" / name
+        if not paths and src_path.exists():
+            paths.add(src_path)
+
         return paths
 
     @classmethod


sdispater:

However, except the checkout of resolved references, I don't think this is the proper fix to this issue. The underlying cause of the problem comes from the fact that locked packages are not used when resolving with the lock file information.

See #3875 for a simpler fix that don't require changes to the `Provider` and the `InstalledRepository` classes.

wakemaster39:
@sdispater I have reworked this fix based on the changes in #3875. That PR did not cover all the issues, the last remaining on is on `poetry update` because the packages are being installed via an editable install it doesn't find the src folder in its paths so it looks like a directory install. This causes poetry to not update the package.

amanda-wee:
Could we get this included in the next version of Poetry (if it does work)? It is clear that issue #2921, despite being closed, has not been fixed.

galamit86:
> Could we get this included in the next version of Poetry (if it does work)? It is clear that issue #2921, despite being closed, has not been fixed.

+1

nirhal:
This PR solved for me! :)
Can you please merge it to the next release?

DataInformer:
Is there anything we can do to help merge this or some fix for the git dependencies not updating?

sdispater (approved):

devunt:
@sdispater When will this fix get released? I saw several releases have been made after this PR being merged, but none of that included this change. Our CI/CD pipeline keeps failing due to this bug so I would like to know whether/when will this fix get included in the upcoming 1.x release. Thank you for your effort.

Issue #2463, opened 2020-05-26
Only use hashes in requirements.txt if there are no VCS repositories


sumanthratna:
- [x] I am on the [latest](https://github.com/python-poetry/poetry/releases/latest) Poetry version.
- [x] I have searched the [issues](https://github.com/python-poetry/poetry/issues) of this repo and believe that this is not a duplicate.
- [x] If an exception occurs when executing a command, I executed it again in debug mode (`-vvv` option). (n/a)

- **OS version and name**: macOS 10.5.5 Beta
- **Poetry version**: 1.0.5
- **Contents of pyproject.toml file**: 
```toml
[tool.poetry]
name = "test-poetry-proj"
version = "0.1.0"
description = "desc"
authors = ["Sumanth Ratna <sumanthratna@gmail.com>"]

[tool.poetry.dependencies]
python = "^3.6.1"
numpy = "^1.18.4"
hyperopt = {git = "https://github.com/hyperopt/hyperopt.git"}


[tool.poetry.dev-dependencies]
pytest = "^5.4.2"

[tool.poetry.scripts]

[build-system]
requires = ["poetry>=0.12"]
build-backend = "poetry.masonry.api"
```

## Issue
When I run `poetry export -f requirements.txt > requirements.txt`, the resulting `requirements.txt` file has a bunch of hashes. Unfortunately, that means I can't run `pip install -r requirements.txt` on another machine, because it returns this error:
```
ERROR: Can't verify hashes for these requirements because we don't have a way to hash version control repositories:
    hyperopt from git+https://github.com/hyperopt/hyperopt.git#egg=hyperopt (from -r requirements.txt (line 192))
```

I'm labeling this as a bug report instead of a feature request because this seems like a fault in the way Poetry behaves when exporting.

nyuszika7h:
This is still an issue. Sounds simple enough to fix.

ryancausey:
Just ran into this while trying to export a requirements.txt for a two-stage docker build. Is there something that needs to be done to help the open PR along?

hterik:
There is a `--without-hashes` option for `poetry export` one can use if you know beforehand there are git requirements in your pyproject.toml.

PR #3874, opened 2021-04-02
Disable hashes for requirements.txt if a VCS dependency is encountered

nyuszika7h:
# Pull Request Check List

Resolves: #2463

### DIFF
--- a/docs/cli.md
+++ b/docs/cli.md
@@ -606,6 +606,8 @@ Only the `requirements.txt` format is currently supported.
 * `--dev`: Include development dependencies.
 * `--extras (-E)`: Extra sets of dependencies to include.
 * `--without-hashes`: Exclude hashes from the exported file.
+  This is automatically enabled if any VCS dependencies are present
+  because of pypa/pip#4995.
 * `--without-urls`: Exclude source repository urls from the exported file.
 * `--with-credentials`: Include credentials for extra indices.
 
--- a/src/poetry/console/commands/export.py
+++ b/src/poetry/console/commands/export.py
@@ -1,6 +1,9 @@
+import warnings
+
 from cleo.helpers import option
 
 from poetry.utils.exporter import Exporter
+from poetry.utils.exporter import VCSHashesDisabledWarning
 
 from .command import Command
 
@@ -68,13 +71,24 @@ class ExportCommand(Command):
             )
 
         exporter = Exporter(self.poetry)
-        exporter.export(
-            fmt,
-            self.poetry.file.parent,
-            output or self.io,
-            with_hashes=not self.option("without-hashes"),
-            dev=self.option("dev"),
-            extras=self.option("extras"),
-            with_credentials=self.option("with-credentials"),
-            with_urls=not self.option("without-urls"),
-        )
+        with warnings.catch_warnings():
+            warnings.filterwarnings("error", category=VCSHashesDisabledWarning)
+            try:
+                exporter.export(
+                    fmt,
+                    self.poetry.file.parent,
+                    output or self.io,
+                    with_hashes=not self.option("without-hashes"),
+                    dev=self.option("dev"),
+                    extras=self.option("extras"),
+                    with_credentials=self.option("with-credentials"),
+                    with_urls=not self.option("without-urls"),
+                )
+            except VCSHashesDisabledWarning:
+                self.line(
+                    "<warning>"
+                    "Warning: Hashes have been disabled because "
+                    "a VCS dependency was encountered which "
+                    "does not support hashes."
+                    "</warning>"
+                )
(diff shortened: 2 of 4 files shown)


nyuszika7h:
Something is wrong, it's not actually outputting the requirements.txt file.

zyv:
This would improve the situation w.r.t. moneymeets/python-poetry-buildpack#19, but is disabling hashes completely actually a good idea? Why to omit hashes only for VCS dependencies?

nyuszika7h:
Because pip is dumb. You either have hashes for all packages or none of them, you can't mix it.

nyuszika7h:
I forgot to unmark this as a draft. I've rebased the PR to fix the merge conflict, should be ready for review/merging now.

nyuszika7h:
Yes, it is still relevant. Sorry for the late reply.

nyuszika7h:
I've rebased the branch to be compatible with the latest changes.

AlexChalk:
@nyuszika7h It's probably worth `@`ing a maintainer to ask for a review, do you know who the right person would be?

zyv:
AFAIK, maintainers are @finswimmer and @abn , so maybe the could voice their opinion?

I’m personally -1 on automatically enabling `—without-hashes`, because it has security implications and in my opinion should be an explicit and conscious decision. Too bad that hashes apparently cannot be omitted for VCS dependencies only. This would habe been a good idea in my view.

There have been some discussions in the related pip bug however in the last few weeks:

https://github.com/pypa/pip/issues/6469

It seems that the maintainers would be happy to accept a per-requirement opt out as the first step towards solving this problem. Maybe if you have time @nyuszika7h you could try to solve the problem at the pip side first?

nyuszika7h:
Trust me, it's not going to get fixed in pip anytime soon. pypa/pip#4995 has been open since 2018 and pypa/pip#6469 since 2019. Their stance basically seems to be "poetry or whoever needs this should fix it themselves".

I specifically coded this PR to explicitly warn about hashes being disabled in such cases, so it should not reduce security in most cases, unless it's run as part of an automated script and log output is not reviewed. I'd be slightly less happier but also OK with the alternative of simply refusing to generate a requirements.txt in such cases unless `--without-hashes` is explicitly passed. It's still better than the status quo, where Poetry happily generates an invalid requirements.txt file that pip will refuse to install from.

zyv:
@nyuszika7h maybe I’m being too optimistic since I didn’t ever have to work with pip maintainers, but they explicitly said on multiple occasions the PRs addressing the problem will be reviewed and accepted, and no such PRs seemed to happen… so yeah, the issues have been open for a very long time, but I didn’t get an impression that they are blocking people, rather nobody seems to be interested enough to try to fix it.

nyuszika7h:
The problem is that there isn't a clear-cut solution to it. Should pip support hashing VCS dependencies somehow, or ignore the lack of hashes unless `--require-hashes` is used (which follows the principle of least surprise, but some may argue it's a downgrade in security from the current behavior)? I could try, but I don't expect it to be resolved quickly.

It seems to me like this behavior isn't necessarily considered a bug by pip maintainers, even though they'd be potentially open to changing it. I would still argue that until it is resolved, Poetry should stop producing invalid requirements.txt files one way or another (whether that is the current proposed patch or aborting if VCS dependencies are encountered and `--without-hashes` is not used), as this has been the behavior of pip for years and a conservative approach in Poetry would not break if pip decided to be more permissive in the future, it would just have to be updated to allow it on pip >= X.Y, and meanwhile people could still use `--without-hashes`.

zyv:
> The problem is that there isn't a clear-cut solution to it. Should pip support hashing VCS dependencies somehow, or ignore the lack of hashes unless `--require-hashes` is used (which follows the principle of least surprise, but some may argue it's a downgrade in security from the current behavior)? I could try, but I don't expect it to be resolved quickly.

Yes, the 100% solution is to teach pip how to hash VCS requirements, but this is complicated and I think there is no consensus how to proceed with that.

However, in pypa/pip#6469 it was proposed to first add `--hash=skip` so that hashes can be used for non-VCS requirements and disabled for VCS requirements. If this is done, then poetry could add `--hash=skip` and warn the used instead of hard-fail. This I personally would consider "safe enough", because pip can't do better, only worse.

> I would still argue that until it is resolved, Poetry should stop producing invalid requirements.txt files one way or another (whether that is the current proposed patch or aborting if VCS dependencies are encountered and `--without-hashes` is not used). Pushing the responsibility back and forth isn't going to achieve anything.

I agree with you that producing broken `requirements.txt` is simply wrong. I would prefer poetry to fail and require `--without-hashes` explicitly instead of silently enabling it - but I'm not the one who makes decisions here - this is just my opinion :-)

neersighted:
Closing this as the code has moved into https://github.com/python-poetry/poetry-plugin-export -- please discuss/PR this over there.

PR #3875, opened 2021-04-02
Fix locked VCS dependencies always being updated

sdispater:
The issue was caused by multiple factors:

- Every package was marked as "use the latest version" which led to the packages not being retrieved from the locked repository. While this did not matter for "normal" packages this had the unintended side effect of pulling the latest revision of VCS dependencies regardless of the locked version. This was the main problem.
- Even when the locked version was used, Poetry would still clone VCS dependencies even though it was no longer needed. This is solved by using the `Provider.load_deferred(False)` since we already have all the information when installing from the lock file.

## Pull Request Check List

Resolves: #2921 

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -390,7 +390,7 @@ dev = ["pre-commit", "tox"]
 
 [[package]]
 name = "poetry-core"
-version = "1.1.0a0"
+version = "1.1.0a1"
 description = "Poetry PEP 517 Build Backend"
 category = "main"
 optional = false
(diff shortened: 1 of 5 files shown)


abn:
This seems to be a dupe of #3867.

sdispater:
> This seems to be partially a dupe of #3867.

It's not really a duplicate (except for the checkout of resolved references). This PR actually fixes the issue by using an existing mechanism available in the `Provider` and by making sure that locked packages are actually used when resolving dependencies against the lock file which was not actually the case up until now.

abn:
Cirrus tests are failing as they use uncached dependencies. Seems the GHA cached the "latest" master from `poetry-core` with the `3.10` fix, which was fixed on `poetry` master via https://github.com/python-poetry/poetry/pull/3846/commits/26a260ab8d9a8580aff67fb23a169c3d8e3c148d. Since this change is fixing the issue, that caused us using an unlocked version of `poetry-core` git source, we might need to regenerate the lock file with the latest master perhaps?

abn (changes requested):
Looks good, and looks like the cirrus failures is actually due to the fix taking effect. See comment above.

sdispater:
@abn I updated the lock file with the latest revision of `poetry-core`

abn:
@sdispater looks like the PEP-610 changes caused a conflict. Otherwise LGTM

sdispater:
@abn Conflicts are fixed

abn (approved):

Issue #2452, opened 2020-05-22
Implement PEP-610


sdispater:
- [x] I have searched the [issues](https://github.com/python-poetry/poetry/issues) of this repo and believe that this is not a duplicate.
- [x] I have searched the [documentation](https://python-poetry.org/docs/) and believe that my question is not covered.

## Feature Request

`pip` recently added support for [PEP-610](https://www.python.org/dev/peps/pep-0610/) which gives the possibility to track the origin of an installed distribution via a `direct_url.json` file in the `.dist-info` directory.

This feature request is two-fold:

* Properly read and handle the `direct_url.json` file. This will avoid some heuristics we do in the `InstalledRepository` class.
* Generate the `direct_url.json` file when installing packages with Poetry. This is especially necessary since we will eventually move away from `pip` for the package installation.

Ideally, this should be implemented as soon as possible. However, since there are still quite a few features left to do for the `1.1` release, this might have to wait for the `1.2` release.

mvoitko:
Can I help with this?

abn:
@mvoitko sure thing, feel free to submit a draft PR. If you want guidance around the code base, ping me on our discord server. I will post some more details on here later.

PR #3876, opened 2021-04-02
Add PEP 610 support

sdispater:
This PR adds support for [PEP-610](https://www.python.org/dev/peps/pep-0610/) which introduced a standard way of recording the origin of direct URL packages (i.e. file, directory, url and VCS packages).

Two things needed to be done for Poetry to be fully compliant:

- Write the `direct_url.json` file to record the origin of the package
- Read the `direct_url.json` file in `InstalledRepository.load()` to retrieve the origin of the information. This also avoids some of the heuristics we are currently doing to guess the origin of the installed packages.

One thing of note is that existing environments will see updates for all their "standard" packages due to the fact that `pip` already supports PEP-610 and since we are installing wheels from our own cache, `pip` consider them as file packages. That's also why we need to remove the `direct_url.json` file generated by `pip` when installing packages. This will be a one-time thing to set the current environments right. New environments will not be affected.

## Pull Request Check List

Resolves: #2452 

### DIFF
--- a/poetry/repositories/installed_repository.py
+++ b/poetry/repositories/installed_repository.py
@@ -1,10 +1,15 @@
 import itertools
+import json
 
 from pathlib import Path
+from typing import TYPE_CHECKING
 from typing import Set
+from typing import Tuple
 from typing import Union
 
 from poetry.core.packages.package import Package
+from poetry.core.packages.utils.utils import url_to_path
+from poetry.core.utils.helpers import canonicalize_name
 from poetry.core.utils.helpers import module_name
 from poetry.utils._compat import metadata
 from poetry.utils.env import Env
@@ -13,6 +18,9 @@ from .repository import Repository
 
 
 _VENDORS = Path(__file__).parent.parent.joinpath("_vendor")
+
+if TYPE_CHECKING:
+    from importlib.metadata import Distribution
 
 
 try:
@@ -68,21 +76,14 @@ class InstalledRepository(Repository):
         return paths
 
     @classmethod
-    def set_package_vcs_properties_from_path(cls, src: Path, package: Package) -> None:
+    def get_package_vcs_properties_from_path(cls, src: Path) -> Tuple[str, str, str]:
         from poetry.core.vcs.git import Git
 
         git = Git()
         revision = git.rev_parse("HEAD", src).strip()
         url = git.remote_url(src)
 
-        package._source_type = "git"
-        package._source_url = url
-        package._source_reference = revision
-
-    @classmethod
-    def set_package_vcs_properties(cls, package: Package, env: Env) -> None:
-        src = env.path / "src" / package.name
-        cls.set_package_vcs_properties_from_path(src, package)
+        return "git", url, revision
 
     @classmethod
     def is_vcs_package(cls, package: Union[Path, Package], env: Env) -> bool:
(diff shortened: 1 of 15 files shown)


poetry/repositories/installed_repository.py
```diff
@@ -99,6 +101,127 @@ def is_vcs_package(cls, package: Union[Path, Package], env: Env) -> bool:
...
+                    unquote(urlparse(url_reference["url"]).path)
+                ).as_posix()
+            else:
+                # URL distribution
+                source_type = "url"
+                source_url = url_reference["url"]
+        elif "dir_info" in url_reference:
+            # Directory distribution
+            source_type = "directory"
+            source_url = Path(url_reference["url"]).as_posix()
```
abn, line 202: We should use `poetry.core.utils.url_to_path` falling back to treating as `path`.

poetry/repositories/installed_repository.py
```diff
@@ -99,6 +101,127 @@ def is_vcs_package(cls, package: Union[Path, Package], env: Env) -> bool:
...
+        url_reference = json.loads(
+            path.joinpath("direct_url.json").read_text(encoding="utf-8")
+        )
+        if "archive_info" in url_reference:
+            # File or URL distribution
+            if url_reference["url"].startswith("file:"):
+                # File distribution
+                source_type = "file"
+                source_url = Path(
+                    unquote(urlparse(url_reference["url"]).path)
```
abn, line 193: We should reuse `poetry.core.utils.url_to_path` here perhaps?

poetry/repositories/installed_repository.py
```diff
@@ -99,6 +101,127 @@ def is_vcs_package(cls, package: Union[Path, Package], env: Env) -> bool:
...
+            source_url=source_url,
+            source_reference=source_reference,
+            source_resolved_reference=source_resolved_reference,
+        )
+        package.description = distribution.metadata.get("summary", "")
+
+        return package
+
+    @classmethod
+    def create_package_from_pep610(cls, distribution: "Distribution") -> Package:
```
abn, line 176: On a general note (not a blocker for this PR), I feel that we have various ways to "create a package" spread across the code base, we should at some point bring them all together so it becomes slightly easier to maintain.

abn (changes requested):
Some minor feedback.

tests/installation/test_executor.py
```diff
@@ -19,17 +20,31 @@
...
 @pytest.fixture
 def env(tmp_dir):
     path = Path(tmp_dir) / ".venv"
     path.mkdir(parents=True)
+
     return MockEnv(path=path, is_venv=True)
 
 
+@pytest.fixture
+def venv(tmp_dir):
```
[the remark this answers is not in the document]
sdispater: I don't want to create a full virtual environment here since it's not necessary and slower.

poetry/repositories/installed_repository.py
sdispater: For now, I don't anticipate that we will use it elsewhere. If we ever do we will see where we can put it, I just don't want us to make early refactoring if we don't need to.

sdispater:
@abn Most of your requested changes have been addressed.

tests/installation/test_executor.py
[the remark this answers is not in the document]
abn: The existing `tmp_venv` fixture does exactly what you are doing here, right? https://github.com/python-poetry/poetry/blob/b753aaf4c3c08ef0e54941a6616fe318fdf4f6e4/tests/conftest.py#L194-L202

Additionally the `ephemeral_environment` also does the same thing, hence why i suggested replacing `tmp_venv` fixture logic with it. 

https://github.com/python-poetry/poetry/blob/b753aaf4c3c08ef0e54941a6616fe318fdf4f6e4/poetry/utils/env.py#L1461-L1480

abn (changes requested):

tests/installation/test_executor.py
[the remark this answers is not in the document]
sdispater: @abn Yes, sorry I think I got confused with something else. I changed the tests to use `tmp_venv` instead.

abn (approved):

Issue #2922, opened 2020-09-16
poetry install: missing linebreaks with --no-ansi


Eugeny:
## Issue
As per above, `poetry install` doesn't output linebreaks between "Installing X(Y)..." lines and no linebreak at the very end either when `--no-ansi` is set.

Alabate:
I've found a weird workaround: Piping the output of the command to anything make line break "appear".

So this command show linebreaks
```
poetry install --no-ansi | cat
```

Eugeny:
Nice trick, thanks!

PR #3881, opened 2021-04-06
Tests for ansi and no-ansi installs + bugfix for ansi install

vgautam:
# Pull Request Check List

Resolves: #2922 (sort of, more context below)

#2922 is mostly fixed on the `master` branch (but not in v1.1.5) because of the cleo upgrade, but there's been a regression on the `--ansi` part of things because of some other changes with how you're doing pip installs. Pip installation should emit an integer "return code" and not the output to stdout of the pip install command. Fixing that allows [this line in executor.py](https://github.com/python-poetry/poetry/blob/master/poetry/installation/executor.py#L298-L299) to work properly and resolves the new ansi bug.

I have confirmed that the tests I wrote fail on `master` but pass on my branch, and I've linted my code.

### DIFF
--- a/poetry/installation/executor.py
+++ b/poetry/installation/executor.py
@@ -117,6 +117,26 @@ class Executor:
 
         return self
 
+    def pip_install(
+        self, req: Union[Path, str], upgrade: bool = False, editable: bool = False
+    ) -> int:
+        func = pip_install
+        if editable:
+            func = pip_editable_install
+
+        try:
+            func(req, self._env, upgrade=upgrade)
+        except EnvCommandError as e:
+            output = decode(e.e.output)
+            if (
+                "KeyboardInterrupt" in output
+                or "ERROR: Operation cancelled by user" in output
+            ):
+                return -2
+            raise
+
+        return 0
+
     def execute(self, operations: List["OperationTypes"]) -> int:
         self._total_operations = len(operations)
         for job_type in self._executed:
@@ -483,9 +503,7 @@ class Executor:
             )
         )
         self._write(operation, message)
-        return pip_install(
-            str(archive), self._env, upgrade=operation.job_type == "update"
-        )
+        return self.pip_install(str(archive), upgrade=operation.job_type == "update")
 
     def _update(self, operation: Union[Install, Update]) -> int:
         return self._install(operation)
@@ -577,13 +595,13 @@ class Executor:
 
                 with builder.setup_py():
                     if package.develop:
-                        return pip_editable_install(req, self._env)
-                    return pip_install(req, self._env, upgrade=True)
+                        return self.pip_install(req, editable=True)
+                    return self.pip_install(req, upgrade=True)
 
         if package.develop:
-            return pip_editable_install(req, self._env)
+            return self.pip_install(req, editable=True)
 
-        return pip_install(req, self._env, upgrade=True)
+        return self.pip_install(req, upgrade=True)
 
     def _install_git(self, operation: Union[Install, Update]) -> int:
         from poetry.core.vcs import Git
--- a/tests/installation/test_executor.py
+++ b/tests/installation/test_executor.py
@@ -44,6 +44,22 @@ def io():
 
 
 @pytest.fixture()
+def io_decorated():
+    io = BufferedIO(decorated=True)
+    io.output.formatter.set_style("c1", Style("cyan"))
+    io.output.formatter.set_style("success", Style("green"))
+
+    return io
+
+
+@pytest.fixture()
+def io_not_decorated():
+    io = BufferedIO(decorated=False)
+
+    return io
+
+
+@pytest.fixture()
 def pool():
     pool = Pool()
     pool.add_repository(MockRepository())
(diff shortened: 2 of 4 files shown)


poetry/installation/executor.py
```diff
@@ -295,9 +295,6 @@ def _do_execute_operation(self, operation: "OperationTypes") -> int:
 
         result = getattr(self, f"_execute_{method}")(operation)
 
-        if result != 0:
-            return result
-
```
sdispater, line 300: We need to keep this check in place.

The underlying error is that the `pip_install()` function returns a string and not an integer.

We need to encapsulate the `pip_install()` (and `pip_editable_install()`) calls so that the check make in `Executor.run_pip()` (see https://github.com/python-poetry/poetry/blob/master/poetry/installation/executor.py#L320) is also used.

sdispater (changes requested):

poetry/installation/executor.py
vgautam: Alright, will do - thanks!

poetry/utils/pip.py
```diff
@@ -44,20 +45,30 @@ def pip_install(
...
+        environment.run_pip(*args)
     except EnvCommandError as e:
         if sys.version_info < (3, 7) and not is_wheel:
             # Under certain Python3.6 installs vendored pip wheel does not contain zip-safe
             # pep517 lib. In this cases we create an isolated ephemeral virtual environment.
             with ephemeral_environment(
                 executable=environment.python, with_pip=True, with_setuptools=True
             ) as env:
-                return environment.run(
+                environment.run(
```
abn, line 56: This might need to be wrapped as well to cover all cases.

poetry/utils/pip.py
```diff
@@ -44,20 +45,30 @@ def pip_install(
...
+
+        output = decode(e.e.output)
+        if (
+            "KeyboardInterrupt" in output
+            or "ERROR: Operation cancelled by user" in output
+        ):
+            return -2
+
         raise PoetryException(f"Failed to install {path.as_posix()}") from e
+    return 0
```
abn, line 71: Does this mean we do not propagate the command return code? Additionally, this also means that we loose debug information when pip fails to install things. 

Also, the fix for returning "int" code will be to pass `call=True` to `run_pip` as this will use `subrprocess.call()` insteald of `subprocess.run()`.

abn (changes requested):

poetry/utils/pip.py
vgautam: I was going off of what @sdispater pointed me to in executor.py so it sounds like both need to change then? https://github.com/python-poetry/poetry/blob/f1b1cf9345837b916e15837d0258f1cc6a0b3f91/poetry/installation/executor.py#L324-L337 

I suppose one part of the answer is to just change both these places to say `call=True` - I assume that this will catch keyboard interrupts too

poetry/utils/pip.py
```diff
@@ -44,20 +45,30 @@ def pip_install(
...
             ) as env:
-                return environment.run(
+                environment.run(
                     env._bin("pip"),
                     *args,
                     env={**os.environ, "PYTHONPATH": str(env.purelib)},
                 )
+            return 0
+
+        output = decode(e.e.output)
```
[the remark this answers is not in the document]
vgautam: I think so, yes. Looks like the [compat code here](https://github.com/python-poetry/poetry/blob/master/poetry/utils/_compat.py#L13-L25) tries a bunch of different encodings and sees what works to decode the string. I suspect this is useful for different systems.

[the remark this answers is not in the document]
abn: I keep forgetting subprocess calls returns encoded strings.

poetry/utils/pip.py
abn: Might need to check this. Might also be a reasonably to move logic to `run_pip` or `run` in `poetry.util.env`.

vgautam: While we're at it, I think we aren't supposed to be using `subprocess.call` anyway because we're [on python3.6+](https://github.com/python-poetry/poetry/blob/master/pyproject.toml#L28) and because you [_can_ get the returncode](https://docs.python.org/3/library/subprocess.html#subprocess.CompletedProcess.returncode) from a `subprocess.CompletedProcess` object (which is returned by `subprocess.run()`)
From Python docs:
> The run() function was added in Python 3.5; if you need to retain compatibility with older versions, see the Older high-level API section.

sdispater: @abn No, we don't propagate the return code of the subprocess. The executor only needs to be aware of whether the command succeeded (return code 0) or was interrupted (return code -1 or if the output contains `KeyboardInterrupt` or `ERROR: Operation cancelled by user`). Any other error will be raised and displayed to the user so that does not prevent debugging.

abn: Hmm. Won't with this change, the output of the pip install command in the subprocess run be lost? Maybe that's what's happening today anyway and it's a non-issue. I haven't checked if all failure modes (eg: successful completion but build failure) raises a command error. If this is the case, we should not handle this in run_pip anyway and the method should always return an int.

sdispater: > Won't with this change, the output of the pip install command in the subprocess run be lost?

It won't if there is an error since an EnvCommandError will be raised which contains the return code and output. In case of success, it does not matter since we don't need to display it to the end user.

vgautam:
Not quite done with this. I've added a bunch of tests though and I think I have a better idea of what needs to be done. Will work on it tomorrow / later and re-request review when I'm ready.

poetry/utils/pip.py
vgautam: I've added tests for this behaviour - a CalledProcessError from subprocess raises an EnvCommandError in poetry (env.run), but a KeyboardInterrupt is just raised as-is and not caught. We'll have to catch it somewhere downstream like in the example from @sdispater. But I think the better way to do it is to look for the return code -2 rather than the string `KeyboardInterrupt` / `Operation cancelled by user`.

This code will also need to be changed, I think, because EnvCommandErrors are a disjoint set from KeyboardInterrupts, so if my understanding is correct, line 226 will never fire because it's impossible for the returncode of an EnvCommandError to be -2.
https://github.com/python-poetry/poetry/blob/master/poetry/installation/executor.py#L222-L228

sdispater: > But I think the better way to do it is to look for the return code -2 rather than the string KeyboardInterrupt / Operation cancelled by user.

That won't work since `pip` might return a return code of `1` in case of `KeyboardInterrupt`. That's why I added this test to ensure we have a `-2` return code.

vgautam: ah! that's annoying :P Alright, I'll keep that check and also add a test for it

sdispater:
@vgautam Are you still working on this?

vgautam:
Ah sorry I've been busy with other stuff but I was planning to finish this work this weekend

sdispater:
@vgautam Thanks! For some guidance I think you are overthinking this and an encapsulation method for `pip_install/pip_editable_install` should be enough to do the trick:

```python
    def pip_install(
        self, req: Union[Path, str], upgrade: bool = False, editable: bool = False
    ) -> int:
        func = pip_install
        if editable:
            func = pip_editable_install

        try:
            func(req, self._env, upgrade=upgrade)
        except EnvCommandError as e:
            output = decode(e.e.output)
            if (
                "KeyboardInterrupt" in output
                or "ERROR: Operation cancelled by user" in output
            ):
                return -2

            raise

        return 0
```

tests/installation/test_executor.py
```diff
@@ -187,6 +203,64 @@ def test_execute_should_show_errors(config, mocker, io, env):
...
 
+def test_execute_works_with_ansi_output(
+    mocker, config, pool, io_decorated, tmp_dir, mock_file_downloads, env
+):
+    config = Config()
+    config.merge({"cache-dir": tmp_dir})
+
+    executor = Executor(env, pool, config, io_decorated)
+
+    mocker.patch.object(env, "_run", return_value=0)
```
sdispater, line 214: Since the issue comes from the fact that it currently returns the standard output, this patch should also return a string an not an integer, otherwise this test does not test the fix we need (as-is it works even on the currently broken code).

sdispater (changes requested):

tests/installation/test_executor.py
vgautam: fixed!

sdispater (approved):
Thanks!

PR #3884, opened 2021-04-06
Add a special error message for missing commands

### LINKED ISSUE #3883: poetry run gives a cryptic FileNotFoundError if the command does not exist

## Feature Request

Consider the following run:

```
$ poetry run does-not-exist

  FileNotFoundError

  [Errno 2] No such file or directory

  at /usr/lib/python3.9/os.py:607 in _execvpe
       603│         path_list = map(fsencode, path_list)
       604│     for dir in path_list:
       605│         fullname = path.join(dir, file)
       606│         try:
    →  607│             exec_func(fullname, *argrest)
       608│         except (FileNotFoundError, NotADirectoryError) as e:
       609│             last_exc = e
       610│         except OSError as e:
       611│             last_exc = e
```

This is confusing to look at. It would make sense to provide an actual error message if a package is missing. Something like the following would be much better:

```
$ poetry run does-not-exist

Error: Command does-not-exist doesn't exist. Please check that it is installed, and that the command name is spelled correctly.
```

SeriousBug:
If a given command is missing, `poetry run` would result in an exception being printed out.

This pull request adds an error message describing what the issue is.

closes #3883

# Pull Request Check List

Resolves: #issue-number-here

### DIFF
--- a/poetry/console/commands/run.py
+++ b/poetry/console/commands/run.py
@@ -28,7 +28,15 @@ class RunCommand(EnvCommand):
         if scripts and script in scripts:
             return self.run_script(scripts[script], args)
 
-        return self.env.execute(*args)
+        try:
+            return self.env.execute(*args)
+        except FileNotFoundError:
+            self.line_error("")
+            self.line_error(
+                "<error>Error: Command {} is not found.</error>".format(script)
+            )
+            self.line_error("")
+            return 1
 
     @property
     def _module(self) -> "Module":
--- a/poetry/console/commands/run.py
+++ b/poetry/console/commands/run.py
@@ -31,11 +31,7 @@ class RunCommand(EnvCommand):
         try:
             return self.env.execute(*args)
         except FileNotFoundError:
-            self.line_error("")
-            self.line_error(
-                "<error>Error: Command {} is not found.</error>".format(script)
-            )
-            self.line_error("")
+            self.line_error(f "<error><c1>{script}</c1>: command not found.</error>")
             return 1
 
     @property


poetry/console/commands/run.py
```diff
@@ -28,7 +28,15 @@ def handle(self) -> Any:
...
 
-        return self.env.execute(*args)
+        try:
+            return self.env.execute(*args)
+        except FileNotFoundError:
+            self.line_error("")
+            self.line_error(
+                "<error>Error: Command {} is not found.</error>".format(script)
+            )
+            self.line_error("")
```
abn, line 38:
```suggestion
            self.line_error(f "<error><c1>{script}</c1>: command not found.</error>")
```

Maybe something more concise is better?

abn (changes requested):
Minor suggestion.

abn:
Would be great if we can add a test case for this.

## Testing PR
### Using pipx
```sh
pipx install --suffix=@3884 'poetry @ git+https://github.com/python-poetry/poetry.git@refs/pull/3884/head'
poetry@3884 new foobar
cd foobar
poetru@3884 run does-not-exist
```

```sh
set -xe
python -m pip install -q git+https://github.com/python-poetry/poetry.git@refs/pull/3884/head
poetry new foobar
pushd foobar
poetry run does-not-exist
EOF
```

dimaqq:
@SeriousBug would you consider pushing some kind of update, so that the builder runs again?
Sadly, the job result has expired, and none can see why it failed.

poetry/console/commands/run.py
```diff
@@ -28,7 +28,11 @@ def handle(self) -> Any:
         if scripts and script in scripts:
             return self.run_script(scripts[script], args)
 
-        return self.env.execute(*args)
+        try:
+            return self.env.execute(*args)
+        except FileNotFoundError:
+            self.line_error(f "<error><c1>{script}</c1>: command not found.</error>")
```
Leonardofreua, line 34: Remove whitespace before starting the string:
```suggestion
            self.line_error(f"<error><c1>{script}</c1>: command not found.</error>")
```

Leonardofreua (changes requested):

SeriousBug:
I'm very busy right now and don't have the time to complete this pull request. If someone would like to see this change in poetry, please fork my repository and submit a new pull request.

PR #3885, opened 2021-04-07
Throw a RuntimeError on hash mismatch in Chooser._get_links

0xDEC0DE:
Throw a specific exception in the case of finding a matching name+version, but none of the digests for a link matching the `poetry.lock` metadata.

Resolves: #2422

- [✅] Added **tests** for changed code.
- [✅] Updated **documentation** for changed code.

### DIFF
--- a/poetry/installation/chooser.py
+++ b/poetry/installation/chooser.py
@@ -106,6 +106,11 @@ class Chooser:
 
             selected_links.append(link)
 
+        if links and not selected_links:
+            raise RuntimeError(
+                f"Retrieved digest for link {link.filename}({h}) not in poetry.lock metadata {hashes}"
+            )
+
         return selected_links
 
     def _sort_key(self, package: Package, link: Link) -> Tuple:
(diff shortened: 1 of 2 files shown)


0xDEC0DE:
This will work best with https://github.com/python-poetry/poetry-core/pull/159

poetry/installation/chooser.py
```diff
@@ -70,7 +70,10 @@ def choose_for(self, package: Package) -> Link:
             links.append(link)
 
         if not links:
-            raise RuntimeError(f"Unable to find installation candidates for {package}")
+            raise RuntimeError(
+                f"Unable to find installation candidates for {package} "
+                f"matching digests: {[p['hash'] for p in package.files]}"
+            )
```
sdispater, line 76: Wouldn't that be misleading since there might be other reasons why no links could be found, right ?

0xDEC0DE: I'm not sure I follow; there are a bunch of criteria for a failing match, this is now simply showing more of them so users can work out what's gone wrong.  But I'm open to suggestions for a better error message if this one is no good.

0xDEC0DE: I force-pushed a new iteration of this change that throws a separate exception for this; that should reduce the ambiguity.

sdispater (approved):
Looks good to me 👍

Issue #2918, opened 2020-09-15
[1.1.0b2] Bumping version leaves orphaned dist-info in site-packages


MrGreenTea:
- [x] I am on the [latest](https://github.com/python-poetry/poetry/releases/latest) Poetry version.
- [x] I have searched the [issues](https://github.com/python-poetry/poetry/issues) of this repo and believe that this is not a duplicate.
- [x] If an exception occurs when executing a command, I executed it again in debug mode (`-vvv` option).

- **OS version and name**: elementary OS 5.1 Hera
- **Poetry version**: 1.1.0b2
- **Link of a [Gist](https://gist.github.com/) with the contents of your pyproject.toml file**: https://gist.github.com/MrGreenTea/992a1795471c00d727c02920d4288f16

## Issue
When updating my packages version with `poetry version` old dist-info directories are left in `site-packages`.
This trips up `pip`, `importlib` and `pkg_resources` for example when trying to find the packages information.
To reproduce:
(run with `poetry config virtualenvs.in-project true`)
1. `❯ poetry new example` 
2. `❯ poetry install`
3. `❯ ls .venv/lib/python3.8/site-packages/example*.dist-info`
```
.venv/lib/python3.8/site-packages/example-0.1.0.dist-info:
INSTALLER  METADATA  RECORD
```
4. `❯ poetry version minor && poetry install`
5. `❯ ls .venv/lib/python3.8/site-packages/example*.dist-info`
```
.venv/lib/python3.8/site-packages/example-0.1.0.dist-info:
INSTALLER  METADATA  RECORD

.venv/lib/python3.8/site-packages/example-0.2.0.dist-info:
INSTALLER  METADATA  RECORD
```
6. `❯ poetry run pip show example`
```
Name: example
Version: 0.2.0
Summary: 
Home-page: None
Author: Jonas Bulik
Author-email: jonas@bulik.dev
License: None
Location: /tmp/example/.venv/lib/python3.8/site-packages
Requires: 
Required-by: 
```
7. `❯ poetry version 0.1.0`
7. `❯ poetry run pip show example`
```
Name: example
Version: 0.2.0
Summary: 
Home-page: None
Author: Jonas Bulik
Author-email: jonas@bulik.dev
License: None
Location: /tmp/example/.venv/lib/python3.8/site-packages
Requires: 
Required-by: 
```

finswimmer:
Hello @MrGreenTea,

[...]

PR #3900, opened 2021-04-08
site: use importlib instead of adhoc file searches

abn:
This change replaces various cases in which installed distribution and
file look-ups used path searches with importlib.metadata backed 
look-ups.

This fixes issues with `dist-info` lookup for PEP610 implementation as 
well as `.pth` file look-up and `dist-info` removal.

Resolves: #2918

### DIFF
--- a/poetry/installation/executor.py
+++ b/poetry/installation/executor.py
@@ -714,19 +714,6 @@ class Executor:
     def _should_write_operation(self, operation: Operation) -> bool:
         return not operation.skipped or self._dry_run or self._verbose
 
-    @staticmethod
-    def _package_dist_info_path(package: "Package") -> Path:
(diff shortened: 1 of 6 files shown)


sdispater:
It seems that the CI errors on Cirrus are due to these changes. Could you take a look?

abn:
@sdispater this should be good now. Instead of trying to make the previous fix work on freebsd, I ended up fixing the underlying issue. This should be good to review now.

stephsamson (approved):

lkev:
Hi, is this fix included in the latest version of poetry (1.1.13)?

I'm still running into the issue in #2918.

simonprovost:
Same here, running into the issue! @lkev found something?

PR #3910, opened 2021-04-09
ci: clean up release task

abn:
This change is in preparation for the removal of the old bootstrap
method in favour of the new `install-poetry` script. Releases no
longer require platform specific build artifacts.


abn:
#3900 should fix the failures.

stephsamson (approved):

PR #3937, opened 2021-04-13
Update CLI docs to include information about poetry lock --no-update

paavanb:
Adds an "options" section for the `lock` command, including documentation about the `--no-update` flag. Also adds a note regarding the default behavior of `poetry lock` to update dependencies.

# Pull Request Check List

Resolves: #1614

### DIFF
--- a/docs/docs/cli.md
+++ b/docs/docs/cli.md
@@ -419,7 +419,9 @@ poetry search requests pendulum
 
 This command locks (without installing) the dependencies specified in `pyproject.toml`.
 
-By default, it will update all dependencies within their version constraints.
+!!!note
+
+     By default, this will lock all dependencies to the latest available compatible versions. To only refresh the lock file, use the `--no-update` option.
 
 ```bash
 poetry lock
--- a/docs/docs/cli.md
+++ b/docs/docs/cli.md
@@ -419,9 +419,15 @@ poetry search requests pendulum
 
 This command locks (without installing) the dependencies specified in `pyproject.toml`.
 
+By default, it will update all dependencies within their version constraints.
+
 ```bash
 poetry lock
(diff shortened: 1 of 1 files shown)


docs/docs/cli.md
```diff
@@ -419,10 +419,16 @@ poetry search requests pendulum
...
 
+By default, it will update all dependencies within their version constraints.
+
 ```bash
 poetry lock
 ```
 
+### Options
+
+* `--no-update`: Avoid updating dependencies that already satisfy their version constraints.
```
abn, line 430:
```suggestion
* `--no-update`: Do not update locked versions, only refresh lock file.
```

Let's re-use the help doc.

docs/docs/cli.md
```diff
@@ -419,10 +419,16 @@ poetry search requests pendulum
 
 This command locks (without installing) the dependencies specified in `pyproject.toml`.
 
+By default, it will update all dependencies within their version constraints.
```
abn, line 422:
```suggestion
!!!note

     By default, this will lock all dependencies to the latest available compatible versions. To only refresh the lock file, use the `--no-update` option.
```

abn (changes requested):

abn (approved):

paavanb:
Thanks! When can I expect the changes to be reflected on the main site?

abn:
That should happen with the `1.2.0` release. The documentation site only reflects latest stable release.

paavanb:
Isn't `--no-update` available on the stable 1.1 release? I had set this PR to be merged into the `1.1` branch because of that.

abn:
If `1.1.6` goes out, it should be updated yes.

Issue #3964, opened 2021-04-17
The following sequence leaves folders behind in site-packages (`master`)


abn:
On my machine, the following sequence leaves folders behind in site-packages:

Install `b` from path:

```bash
$ poetry@3940 add ../b/

Updating dependencies
Resolving dependencies... (0.1s)

Writing lock file

Package operations: 1 install, 0 updates, 0 removals

  • Installing b (0.0.0 /home/fredrik/code/repos/b): Building...
```

Then I wanted to see what happens if I try to add it again but with the `--editable` arg:

```bash
$ poetry@3940 add --editable ../b/

Updating dependencies
Resolving dependencies... (0.1s)

Writing lock file

Package operations: 0 installs, 1 update, 0 removals

  • Updating b (2.1.3.dev0+g77bcc5e.d20210417114942 /home/fredrik/code/repos/b -> 0.0.0 /home/fredrik/code/repos/b)
```

Finally uninstalling `b`:

```bash
$ poetry@3940 remove b
Updating dependencies
Resolving dependencies... (0.1s)

Writing lock file

Package operations: 0 installs, 0 updates, 1 removal

  • Removing b (0.0.0 /home/fredrik/code/repos/b)
```

But unfortunately, `b` was not completely removed, likely due to the fact that I was able to run `poetry add` twice:

![image](https://user-images.githubusercontent.com/994357/115108952-7650e300-9f73-11eb-9efe-7d460854e4c8.png)

And now it seems the .venv is b0rked, as there is suddenly a `None` package:

```bash
$ poetry@3940 show

  TypeError

  expected string or bytes-like object

  at ~/.local/pipx/venvs/poetry@3940/lib/python3.9/site-packages/poetry/core/utils/helpers.py:27 in canonicalize_name
       23│ _canonicalize_regex = re.compile(r"[-_]+")
       24│
       25│
       26│ def canonicalize_name(name: str) -> str:
    →  27│     return _canonicalize_regex.sub("-", name).lower()
       28│
       29│
       30│ def module_name(name: str) -> str:
       31│     return canonicalize_name(name).replace(".", "_").replace("-", "_")
```

_Originally posted by @fredrikaverpil in https://github.com/python-poetry/poetry/pull/3940#issuecomment-821798593_

abn:
This is likely related to #3900.

fredrikaverpil:
[...]

PR #3967, opened 2021-04-17
Add direct_url.json to RECORD upon creation

fredrikaverpil:
# Pull Request Check List

Resolves: #3964

___

Seems the dist-info/RECORD was introduced by PEP-376 back in 2009:
http://packaging.python.org/specifications/recording-installed-packages.html

> "If the RECORD file is missing, tools that rely on .dist-info must not atempt to uninstall or upgrade the package. (This does not apply to tools that rely on other sources of information, such as system package managers in Linux distros.)"

Also, [this](https://discuss.python.org/t/pep-610-recording-the-origin-of-distributions-installed-from-direct-url-references/1535/73) answered my question on whether direct_url.json should go into RECORD or not (it should).

### DIFF
--- a/poetry/installation/executor.py
+++ b/poetry/installation/executor.py
@@ -777,8 +777,9 @@ class Executor:
                     encoding="utf-8",
                 )
 
-                with dist._path.joinpath("RECORD").open(mode="a") as record:
-                    record.write(str(dist._path.joinpath("direct_url.json")))
+                if dist._path.joinpath("RECORD").exists():
+                    with dist._path.joinpath("RECORD").open(mode="a") as record:
+                        record.write(str(dist._path.joinpath("direct_url.json")))
 
     def _create_git_url_reference(
         self, package: "Package"
--- a/poetry/installation/executor.py
+++ b/poetry/installation/executor.py
@@ -783,7 +783,15 @@ class Executor:
                     with record.open(mode="a") as f:
                         writer = csv.writer(f)
                         writer.writerow(
-                            [str(dist._path.joinpath("direct_url.json").relative_to(record.parent.parent)), "", ""]
+                            [
+                                str(
+                                    dist._path.joinpath("direct_url.json").relative_to(
+                                        record.parent.parent
+                                    )
+                                ),
+                                "",
+                                "",
+                            ]
                         )
 
     def _create_git_url_reference(
--- a/poetry/installation/executor.py
+++ b/poetry/installation/executor.py
@@ -1,3 +1,4 @@
+import csv
 import itertools
 import json
 import os
--- a/poetry/installation/executor.py
+++ b/poetry/installation/executor.py
@@ -777,6 +777,9 @@ class Executor:
                     encoding="utf-8",
                 )
 
+                with dist._path.joinpath("RECORD").open(mode="a") as record:
+                    record.write(str(dist._path.joinpath("direct_url.json")))
+
     def _create_git_url_reference(
         self, package: "Package"
     ) -> Dict[str, Union[str, Dict[str, str]]]:
--- a/poetry/installation/executor.py
+++ b/poetry/installation/executor.py
@@ -780,7 +780,7 @@ class Executor:
 
                 record = dist._path.joinpath("RECORD")
                 if record.exists():
-                    with record.open(mode="a") as f:
+                    with record.open(mode="a", encoding='utf-8') as f:
                         writer = csv.writer(f)
                         writer.writerow(
                             [
--- a/poetry/installation/executor.py
+++ b/poetry/installation/executor.py
@@ -777,9 +777,13 @@ class Executor:
                     encoding="utf-8",
                 )
 
-                if dist._path.joinpath("RECORD").exists():
-                    with dist._path.joinpath("RECORD").open(mode="a") as record:
-                        record.write(str(dist._path.joinpath("direct_url.json")))
+                record = dist._path.joinpath("RECORD")
+                if record.exists():
+                    with record.open(mode="a") as f:
+                        writer = csv.writer(f)
+                        writer.writerow(
+                            [str(dist._path.joinpath("direct_url.json").relative_to(record.parent.parent)), "", ""]
+                        )
 
     def _create_git_url_reference(
         self, package: "Package"
--- a/poetry/installation/executor.py
+++ b/poetry/installation/executor.py
@@ -780,7 +780,7 @@ class Executor:
 
                 record = dist._path.joinpath("RECORD")
                 if record.exists():
-                    with record.open(mode="a", encoding='utf-8') as f:
+                    with record.open(mode="a", encoding="utf-8") as f:
                         writer = csv.writer(f)
                         writer.writerow(
                             [
--- a/tests/installation/test_executor.py
+++ b/tests/installation/test_executor.py
@@ -361,7 +361,8 @@ def verify_installed_distribution(venv, package, url_reference=None):
         direct_url_entry = direct_url_file.relative_to(record_file.parent.parent)
         assert direct_url_file.exists()
         assert str(direct_url_entry) in {
-            row.split(",")[0] for row in record_file.read_text(encoding="utf-8").splitlines()
+            row.split(",")[0]
+            for row in record_file.read_text(encoding="utf-8").splitlines()
         }
         assert json.loads(direct_url_file.read_text(encoding="utf-8")) == url_reference
     else:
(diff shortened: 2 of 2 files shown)


poetry/installation/executor.py
```diff
@@ -759,6 +759,10 @@ def _save_url_reference(self, operation: "OperationTypes") -> None:
                     encoding="utf-8",
                 )
 
+                if dist._path.joinpath("RECORD").exists():
+                    with dist._path.joinpath("RECORD").open(mode="a") as record:
+                        record.write(str(dist._path.joinpath("direct_url.json")))
```
sdispater, line 764: I don't think it will work since the `RECORD` file is a CSV file so the added line should be a valid CSV line.

Also, it should not be an absolute path to the file but relative to the site-packages directory.

```suggestion
                record = dist._path.joinpath("RECORD")
                if record.exists():
                    with record.open(mode="a") as f:
                        writer = csv.writer(f)
                        writer.writerow(
                            [str(dist._path.joinpath("direct_url.json").relative_to(record.parent.parent)), "", ""]
                        )
```

Finally, could you add a test covering this change?

sdispater (changes requested):

poetry/installation/executor.py
fredrikaverpil: @sdispater it did work, I tried it myself. But you are right that it should be treated as CSV.

I'll look into how I can add a test. I guess I should write one which installs a package from path, then removes it - and assert that there's no lingering `direct_url.json` file in the package's dist-info folder?

sdispater: @fredrikaverpil For the test, simply ensuring that the RECORD file has the `direct_url.json` reference should be enough.

fredrikaverpil: @sdispater I've now applied your code suggestion and added an additional assert to the existing `verify_installed_distribution` test.

I'm noticing that in your code suggestion you changed the absolute path into a relative one. According to [the docs](https://packaging.python.org/specifications/recording-installed-packages/#the-record-file), it's okay to use _either_ absolute or relative. So relative seems nicer 👍 But, I just want you to know that the `poetry add ../b --editable` command resulted in the following `RECORD` on my machine (Ubuntu 20.04), where the `direct_url.json` path was the _only_ one relative:

```csv
/home/fredrik/code/repos/poetry/.venv/lib/python3.8/site-packages/b.pth,sha256=PB1gW2CKFckiIksXEn9ySceS8mhVpDf9NNMba5UTaAE,31
/home/fredrik/code/repos/poetry/.venv/bin/bclixq,sha256=oUuzygRt7DgCtrAZlWpxIDam0n76RCpGe6FU5P-zx3s,115
/home/fredrik/code/repos/poetry/.venv/lib/python3.8/site-packages/b-0.0.0.dist-info/METADATA,sha256=zaDEAXqbNAXJXxcfQbscYIbhAxFGCpoYI37JYOJNyo4,564
/home/fredrik/code/repos/poetry/.venv/lib/python3.8/site-packages/b-0.0.0.dist-info/INSTALLER,sha256=gcMg69IJ3d31kOPdejOXCeJ_7u2OVE9WaZ5A1idCOi0,6
/home/fredrik/code/repos/poetry/.venv/lib/python3.8/site-packages/b-0.0.0.dist-info/entry_points.txt,sha256=qroUuuKxidhz-9dsg5bVbzH_PX6YmETz_ZmpKlajQ08,39
/home/fredrik/code/repos/poetry/.venv/lib/python3.8/site-packages/b-0.0.0.dist-info/RECORD,,
b-0.0.0.dist-info/direct_url.json,,
```

poetry/installation/executor.py
```diff
@@ -777,6 +778,22 @@ def _save_url_reference(self, operation: "OperationTypes") -> None:
                     encoding="utf-8",
                 )
 
+                record = dist._path.joinpath("RECORD")
+                if record.exists():
+                    with record.open(mode="a") as f:
```
abn, line 783:
```suggestion
                    with record.open(mode="a", encoding='utf-8') as f:
```

tests/installation/test_executor.py
```diff
@@ -356,8 +356,14 @@ def verify_installed_distribution(venv, package, url_reference=None):
 
     direct_url_file = distribution._path.joinpath("direct_url.json")
 
+    record_file = distribution._path.joinpath("RECORD")
+    with open(record_file, "r") as f:
+        record_file_contents = f.read().splitlines()
+    direct_url_entry = direct_url_file.relative_to(record_file.parent.parent)
```
abn, line 362: This might be better off as a csv reader populating a set from the first column. Should make the assertion as simple as:

```py
assert direct_url_entry in records
```

abn: Alternatively maybe use `Path.read_text` and/or `Path.open`.

abn (changes requested):

tests/installation/test_executor.py
fredrikaverpil: Hm, not entirely sure what you mean here.

The records are returned as a list of strings. These strings are comma separated. So I can't just `assert direct_url_entry in record_file_contents`.

I believe you mean that my assert was overly complicated with that list comprehension expression, and I agree.

I interpret your comment such as that there should be a nicer/cleaner/simpler way of asserting. I believe we need to assert that we should have at least one row in the `RECORD` file which begins with the direct_url.json file. So I am not sure if `Path.read_text` will cut it.

I'll make a couple of suggestions.

fredrikaverpil: Creating another list we can assert with:

```python
    with open(record_file, "r") as f:
        records_first_column = [row.split(",")[0] for row in f]
    assert str(direct_url_entry) in records_first_column
```

Using `Path.read_text`:

```python
assert str(direct_url_entry) in record_file.read_text()
```

@abn do you think the `Path.read_text` approach is sufficient or should I go with the one which checks the first column of the csv RECORD file?

tests/installation/test_executor.py
```diff
@@ -356,8 +356,14 @@ def verify_installed_distribution(venv, package, url_reference=None):
...
+
     if url_reference is not None:
         assert direct_url_file.exists()
+        assert [r for r in record_file_contents if r.startswith(str(direct_url_entry))]
         assert json.loads(direct_url_file.read_text(encoding="utf-8")) == url_reference
     else:
         assert not direct_url_file.exists()
```
fredrikaverpil, line 369:
```suggestion
    record_file = distribution._path.joinpath("RECORD")
    direct_url_entry = direct_url_file.relative_to(record_file.parent.parent)
    with open(record_file, "r") as f:
        records_first_column = [row.split(",")[0] for row in f]


    if url_reference is not None:
        assert direct_url_file.exists()
        assert str(direct_url_entry) in records_first_column
        assert json.loads(direct_url_file.read_text(encoding="utf-8")) == url_reference
    else:
        assert not direct_url_file.exists()
```

tests/installation/test_executor.py
fredrikaverpil: I committed the first example to the test, as I think it's better to check only the first column's values.

tests/installation/test_executor.py
```diff
@@ -356,8 +356,14 @@ def verify_installed_distribution(venv, package, url_reference=None):
...
     direct_url_file = distribution._path.joinpath("direct_url.json")
 
+    record_file = distribution._path.joinpath("RECORD")
+    direct_url_entry = direct_url_file.relative_to(record_file.parent.parent)
+    with open(record_file, "r") as f:
+        records_first_column = [row.split(",")[0] for row in f]
+
     if url_reference is not None:
         assert direct_url_file.exists()
+        assert str(direct_url_entry) in records_first_column
```
abn, line 366:
```suggestion
    if url_reference is not None:
        record_file = distribution._path.joinpath("RECORD")
        direct_url_entry = direct_url_file.relative_to(record_file.parent.parent)
        assert direct_url_file.exists()
        assert str(direct_url_entry) in {
            row.split(",")[0] for row in record.read_text(encoding="utf-8").splitlines()
        }
```

abn: Minor suggestion. The comprehension is not required, but might make sense to move the read to within the if block.

abn (changes requested):

tests/installation/test_executor.py
fredrikaverpil: @abn looks good, applied your suggestion!

abn (approved):

ganeshgrandhi:
May I know in which version this is released?

neersighted:
This is present in the `1.2.x` series (see https://python-poetry.org/blog/announcing-poetry-1.2.0/#support-for-direct-origin-url-records-pep-610 as well as the changelog).

Issue #3971, opened 2021-04-18
Add --lock option to "poetry remove"


agronholm:
- [X] I have searched the [issues](https://github.com/python-poetry/poetry/issues) of this repo and believe that this is not a duplicate.
- [X] I have searched the [documentation](https://python-poetry.org/docs/) and believe that my question is not covered.

## Feature Request

Poetry already has `add --lock` and `update --lock`. In order to have a workflow that does not involve virtualenvs, this is a missing piece.

v-goncharenko:
Yes, this is very handy functionality which `poetry` currently lacks

PR #3972, opened 2021-04-18
Added --lock option for "poetry remove"

agronholm:
# Pull Request Check List

Resolves: #3971

Poetry already supports "add --lock" and "update --lock" but for some reason the corresponding "remove --lock" feature was missing.

### DIFF
--- a/poetry/console/commands/remove.py
+++ b/poetry/console/commands/remove.py
@@ -19,6 +19,7 @@ class RemoveCommand(InstallerCommand):
             "Output the operations but do not execute anything "
             "(implicitly enables --verbose).",
         ),
+        option("lock", None, "Do not perform operations (only update the lockfile)."),
     ]
 
     help = """The <info>remove</info> command removes a package from the current
@@ -73,6 +74,8 @@ list of installed packages
         self._installer.verbose(self._io.is_verbose())
         self._installer.update(True)
         self._installer.whitelist(requirements)
+        if self.option("lock"):
+            self._installer.lock()
 
         status = self._installer.run()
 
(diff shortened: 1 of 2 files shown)


agronholm:
I need some help writing a test for this.

agronholm:
Also next I might want to address the fact that these commands still create a useless virtualenv.

poetry/console/commands/remove.py
```diff
@@ -73,6 +74,8 @@ def handle(self) -> int:
         self._installer.verbose(self._io.is_verbose())
         self._installer.update(True)
         self._installer.whitelist(requirements)
+        if self.option("lock"):
+            self._installer.lock()
 
         status = self._installer.run()
```
abn, line 80: Intaller should not run if its a lock only op.

agronholm:
Why does the `add` command then do the same? I wouldn't know what to do in this situation which is why I asked for help on Discord.

poetry/console/commands/remove.py
agronholm: Why does the add command then do the same? I wouldn't know what to do in this situation which is why I asked for help on Discord.

neersighted:
I'm going to close this as it seems to be stalled out and very naive in its current form. I would be happy to review a future version with robust tests.

agronholm:
It's stalled because I never got any help with the implementation, particularly with tests. I would be happy to continue work on this if somebody gave me the information I'm missing.

neersighted:
I'd take a look at similar tests around locking -- particularly `test_lock.py`. It's a pretty good example of the level of mocking needed to set up a command-line test, as well as how to invoke commands and check Poetry's results after.

Issue #3567, opened 2021-01-12
user and password available in poetry.lock


sblondon:
- [x] I am on the [latest](https://github.com/python-poetry/poetry/releases/latest) Poetry version.
- [x] I have searched the [issues](https://github.com/python-poetry/poetry/issues) of this repo and believe that this is not a duplicate.

- **OS version and name**: Debian 10 (buster)
- **Poetry version**: 1.1.4

## Issue
I'm unsure this should be considered as a bug.

In `poetry.lock`, the `url` field fills packages location. If HTTP authentication is used, the user and password are provided in the URL so it's readable in the file. For example:

```
[package.source]
type = "legacy"
url = "https://user:password@pypi.domain.tld/site/prod/+simple"
reference = "whatever"
```

It could be replaced by `https://**user**:**password**@...`. However, I wonder if it's the good strategy because the `poetry.lock` file can't be used directly to rebuild a virtualenv without doing search/replace those strings. What do you think about it?

I could provide a PR but I prefer to get some feedback before.

1ace:
My understanding is that `poetry.lock` is meant to contain all the information necessary to accurately re-create an environment. As such, you can't remove information from it and rely on finding that information somewhere else afterward.

robertlagrant:
Would vote that this is not a bug, but would suggest that in the presence of http auth that a warning is printed to the console.

sblondon:
I like the warning idea. I will see to implement it.

timos-flwls:
I've just noticed the same re github tokens. The warning would be a good first step, but I really think this needs to be revisited. How many people are unwittingly committing credentials to repos because of this surprising behaviour?

IMO credentials are per user, and lockfiles are designed shared across multiple environments and users. Therefore philosophically credentials should really not be in lockfiles (or anywhere in plain text in a repo), especially without explicit opt-in (!).

Surely everyone would rather have some (potential) friction with other users installing their dependencies, rather than potentially full read access to their private pypi/company github org being available in plain text in every package.

The insane thing is, if I find and replace the token out of my lockfile, the install works fine! i.e. replacing `myorg @ git+https://MYTOKEN@github.com/myorg/repo.git@main` with `myorg @ git+https://www.github.com/myorg/repo.git@main`

PR #3995, opened 2021-04-24
Add a warning if credentials are written in lock file

sblondon:
# Pull Request Check List

Resolves: #3567

This PR adds a `logging.warning()` if some credentials are written in the lock file. It's based on [this comment](https://github.com/python-poetry/poetry/issues/3567#issuecomment-824632468).

I wonder if changing the attribute (`Locker.contains_credential`) to a method would be better (or not). The message could probably be improved but I did not find the better sentence.


Secrus:
Hi, @sblondon are you still interested in bringing this to Poetry? If so, please sync changes from the master branch and resolve conflicts. That would allow for review.

sblondon:
@Secrus I still think it would be better to add the warning. I rebased the patch (and resolve the conflicts).

All environments fail in CI failed because there is an error when `poetry run mypy`mypy' in src/poetry/utils/env.py. I don't think it's linked to the patch. I will search if it exists on `master` too.

src/poetry/packages/locker.py
```diff
@@ -420,6 +421,10 @@ def set_lock_data(self, root: Package, packages: list[Package]) -> bool:
         return False
 
     def _write_lock_data(self, data: TOMLDocument) -> None:
+        if self.contains_credential:
+            logger.warning(
+                "Credential was provided in packages list. It's included in the lock file."
```
neersighted, line 426:
```suggestion
                "A package URL contains credentials, which will be included in poetry.lock."
```

neersighted (changes requested):

src/poetry/packages/locker.py
```diff
@@ -61,6 +61,7 @@ def __init__(self, lock: str | Path, local_config: dict[str, Any]) -> None:
         self._local_config = local_config
         self._lock_data: TOMLDocument | None = None
         self._content_hash = self._get_content_hash()
+        self.contains_credential = False
```
neersighted, line 64:
```suggestion
        self._contains_credential = False
```

src/poetry/packages/locker.py
```diff
@@ -420,6 +421,10 @@ def set_lock_data(self, root: Package, packages: list[Package]) -> bool:
         return False
 
     def _write_lock_data(self, data: TOMLDocument) -> None:
+        if self.contains_credential:
+            logger.warning(
+                "A package URL contains credentials, which will be included in poetry.lock."
```
neersighted, line 426:
```suggestion
                "A package URL contains credentials and will be written to poetry.lock."
```

neersighted (changes requested):
Two minor changes -- sorry for forgetting to ask for the property to be renamed earlier.

Additionally, we need to get CI passing so we can merge this.

src/poetry/packages/locker.py
sblondon: I changed every occurences of `_contains_credential` attribute.

sblondon:
> Additionally, we need to get CI passing so we can merge this.

Yes, I have to understand why it fails since the rebase.

src/poetry/packages/locker.py
```diff
@@ -420,6 +421,10 @@ def set_lock_data(self, root: Package, packages: list[Package]) -> bool:
         return False
 
     def _write_lock_data(self, data: TOMLDocument) -> None:
+        if self._contains_credential:
+            logger.warning(
+                "A package URL contains credentials and will be written to poetry.lock."
```
dimbleby, line 426: This looks like a frustrating error message - I might plausibly have hundreds of packages in `poetry.lock`, being told that _one_ of them has credentials is going to leave me with work to do...

Can this warning say which package is the problematic one?

sblondon: The warning is not triggered by a package. It is displayed if credentials are in the URL of an added source in `package.source` section.

dimbleby: Those credentials are associated with a particular package, no?  eg in your testcase, it's package `A` that triggers the warning.

If I have a lockfile with hundreds of packages in it, then a more helpful warning would point me at `A` rather than only telling me that there's a problem, somewhere.

sblondon: When I got this issue, the URL coming from the `package.source` was added to every packages in the lock file. So it's not useful. However, adding the URL in the message could help the user:

```
Package URL contain credentials (https://alice:a_passwd@pypi.domain.tld/site/prod/+simple) and will be written to poetry.lock.
```

From file:
```
[package.source]
type = "legacy"
url = "https://alice:a_passwd@pypi.domain.tld/site/prod/+simple"
reference = "whatever"
```

sblondon:
@neersighted The error in CI is not the tests of poetry ('Run pytest' and 'Run test (integration suite)' are green) but the tests in 
poetry-plugin-export step fail ('Run pytest (poetry-plugin-export)'). 

`tests.helpers.TestLocker` class exists in python-poetry and poetry-plugin-export repositories. In poetry-plugin-export repository, `tests.helpers.TestLocker` class does not mock [`_contains_credential` attribute](https://github.com/python-poetry/poetry-plugin-export/blob/main/tests/helpers.py#L38) . `_contains_credential` attribute is added in this PR so it has never been necessary in `TestLocker` mock in poetry-plugin-export. So the tests raise `AttributeError: 'TestLocker' object has no attribute '_contains_credential'`.

To fix the error, I can do a PR to add the attribute in the `tests.helpers.TestLocker.__init__` method in poetry-plugin-export repository.

What do you think about it?

neersighted:
> @neersighted The error in CI is not the tests of poetry ('Run pytest' and 'Run test (integration suite)' are green) but the tests in poetry-plugin-export step fail ('Run pytest (poetry-plugin-export)').
> 
> `tests.helpers.TestLocker` class exists in python-poetry and poetry-plugin-export repositories. In poetry-plugin-export repository, `tests.helpers.TestLocker` class does not mock [`_contains_credential` attribute](https://github.com/python-poetry/poetry-plugin-export/blob/main/tests/helpers.py#L38) . `_contains_credential` attribute is added in this PR so it has never been necessary in `TestLocker` mock in poetry-plugin-export. So the tests raise `AttributeError: 'TestLocker' object has no attribute '_contains_credential'`.
> 
> To fix the error, I can do a PR to add the attribute in the `tests.helpers.TestLocker.__init__` method in poetry-plugin-export repository.
> 
> What do you think about it?

Sorry I never saw this/got back to you, @sblondon... Feel free to make a PR against that repo and link to it here. However, I suspect that we'll simply try and decouple the testing of the two repos in the near future ala #5980.

sblondon:
I created a PR in https://github.com/python-poetry/poetry-plugin-export/pull/93 to add the `TestLocker._contains_credential` attribute in the library.

I agree it would be good to fix such dependency between the two repositories. The changes  in `TestLocker.__init__` classes from `poetry` and `poetry-plugin-export` are identical so the refactoring will not be harder.

sblondon:
The patch in poetry-plugin-extra is merged.
When a new release will be available, I will update the dependency in pyproject.toml in order to fix the tests.

src/poetry/packages/locker.py
neersighted: This LGTM, except I think we should resolve this thread. Can we include the URL?

sblondon: @neersighted I can modify `_contains_credential` to be the URL instead of a boolean status. In such case, the name fits less well. If it's still ok fo you, I'll be happy to modify it. An other way would be to rename it but it will also require to rename it in the poetry-plugin-export  repository. The third way is to not add the URL. My favorite way is the first one (modifying the content of `_contains_credential`). What do you prefer?

Secrus:
Stalled

Issue #312, opened 2018-07-17
sorting packages


tkossak:
- [x] I have searched the [issues](https://github.com/sdispater/poetry/issues) of this repo and believe that this is not a duplicate.

## Issue
When I use 'poetry init' and get to the step for searching packages, it looks like huge mess: https://imgur.com/a/WA9CcnJ
Maybe we can get this list sorted alphabetically? Or/except maybe display packages that are most likely what I look for (because the name is exactly what i typed) on the bottom, so I don't have to scroll the list in search for its number.

nchammas:
I found this to be a minor issue with `poetry init`, too. One easy improvement that could be made would be to always rank any exact match at the top.

Further improvements are certainly possible (perhaps Poetry could somehow leverage the search result rankings returned by https://pypi.org/), but just placing exact matches at the top would be a nice improvement by itself.

tkossak:
Not at the top - at the bottom (like https://github.com/Jguer/yay is doing) so we don't have to scroll for best matches but see them instantly where the cursor is (on the bottom that is).

sdispater:
I agree that it is a usability issue. The idea of putting the more relevant packages at the bottom seems a good improvement to start with.

rafalkrupinski:
I'd sort dependencies alphabetically even in pyproject.toml, I find that  helps with merging branches if they both have new entries added

n8sty:
I took a crack at trying to fix this to learn a bit more about Poetry's guts and contribute to a tool that's made my work a bit easier. I think I may have discovered another issue that relates to specifying packages interactively, and is potentially solved by sorting.

When I go through a `poetry init` flow in an empty directory and specify either `Django` or `django` I can't actually get either one picked up by https://github.com/sdispater/poetry/blob/f4803e91bca440e6f4af6ad54e396d01c024a76f/poetry/console/commands/init.py#L189-L193. I believe this happens because the `choice` method of `cleo.Command` only allows what's returned in the array of choices to be selected from. And, the array of responses when using either of those search terms doesn't include `[dD]jango`. Digging more, I tried using `xmlrpc.client.ServerProxy` to query PyPi by `xmlrpc.client.ServerProxy("https://pypi.python.org/pypi").search({'name': 'django'}, "or")[0]` which returned (what I'm looking for)

```
{'name': 'Django',
 'summary': 'A high-level Python Web framework that encourages rapid development and clean, pragmatic design.',
 'version': '2.2.4',
[...]

PR #3996, opened 2021-04-24
Introduce dependency sorting

br3ndonland:
Resolves: #312
Resolves: #4383

## Description

Thank you for Poetry. It's a huge help to the Python community, and I enjoy working with it.

When installing a package with `poetry add`, the package is appended to the appropriate section of the _pyproject.toml_. Some developers have their dependencies sorted alphabetically, so they have to manually sort the new dependency into the appropriate place in _pyproject.toml_. It would be great to have the dependencies automatically sorted.

This PR will introduce the ability to sort dependencies alphabetically when adding. Dependency sorting can be configured with `poetry config dependencies.sort true` or `export POETRY_DEPENDENCIES_SORT=true`. If `true`, dependencies will be sorted alphabetically when adding. For the default dependency group, `python` will be at the top of the list.

## Related

https://github.com/python-poetry/poetry/issues/312#issuecomment-466782002
https://github.com/python-poetry/poetry/issues/312#issuecomment-724623426
python-poetry/poetry#2308
python-poetry/poetry#4260

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -171,6 +171,14 @@ If you encounter any problems with it, set to `true` to use the system git backe
 
 Defaults to `false`.
 
+### `dependencies.sort`
+
+**Type**: boolean
+
+Sort dependencies alphabetically when adding.
+
+Defaults to `false`.
+
 ### `installer.parallel`
 
 **Type**: boolean
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -175,7 +175,7 @@ Defaults to `false`.
 
 **Type**: boolean
 
-Sort dependencies alphabetically when adding.
+Sort dependencies in `pyproject.toml` alphabetically when adding.
 
 Defaults to `false`.
 
--- a/src/poetry/console/commands/add.py
+++ b/src/poetry/console/commands/add.py
@@ -235,7 +235,7 @@ The add command adds required packages to your <comment>pyproject.toml</> and in
             sorted_dependencies = OrderedDict(sorted(section.items()))
             if sorted_dependencies.get("python"):
                 sorted_dependencies.move_to_end("python", last=False)
-            if group == "default":
+            if group == MAIN_GROUP:
                 poetry_content["dependencies"] = sorted_dependencies
             else:
                 poetry_content["group"][group]["dependencies"] = sorted_dependencies
--- a/src/poetry/console/commands/add.py
+++ b/src/poetry/console/commands/add.py
@@ -2,6 +2,7 @@ from __future__ import annotations
 
 import contextlib
 
+from collections import OrderedDict
 from typing import Any
 
 from cleo.helpers import argument
@@ -229,6 +230,15 @@ The add command adds required packages to your <comment>pyproject.toml</> and in
                     root_dir=self.poetry.file.parent,
                 )
             )
+
+        if self.poetry.config.get("dependencies.sort") is True:
+            sorted_dependencies = OrderedDict(sorted(section.items()))
+            if sorted_dependencies.get("python"):
+                sorted_dependencies.move_to_end("python", last=False)
+            if group == "default":
+                poetry_content["dependencies"] = sorted_dependencies
+            else:
+                poetry_content["group"][group]["dependencies"] = sorted_dependencies
 
         # Refresh the locker
         self.poetry.set_locker(
--- a/src/poetry/console/commands/init.py
+++ b/src/poetry/console/commands/init.py
@@ -12,7 +12,6 @@ from typing import Union
 
 from cleo.helpers import option
 from packaging.utils import canonicalize_name
-from tomlkit import document
 from tomlkit import inline_table
 
 from poetry.console.commands.command import Command
@@ -56,13 +55,6 @@ class InitCommand(Command):
             flag=False,
             multiple=True,
         ),
-        option(
-            "sort-dependencies",
-            None,
-            "Sort dependencies in pyproject.toml alphabetically.",
-            flag=True,
-            value_required=False,
-        ),
         option("license", "l", "License of the package.", flag=False),
     ]
 
@@ -80,9 +72,9 @@ The <c1>init</c1> command creates a basic <comment>pyproject.toml</> file in the
         from pathlib import Path
 
         from poetry.core.pyproject.toml import PyProjectTOML
-        from poetry.core.toml.file import TOMLFile
         from poetry.core.vcs.git import GitConfig
 
+        from poetry.config.config import Config
         from poetry.layouts import layout
         from poetry.utils.env import SystemEnv
 
@@ -103,6 +95,7 @@ The <c1>init</c1> command creates a basic <comment>pyproject.toml</> file in the
                 )
                 return 1
 
+        config = Config()
         vcs_config = GitConfig()
 
         if self.io.is_interactive():
@@ -218,19 +211,9 @@ You can specify a package in the following forms:
             if self.io.is_interactive():
                 self.line("")
 
-        sort_dependencies = self.option("sort-dependencies")
-        question = "Would you like to sort your dependencies alphabetically?"
-        if self.confirm(question, True):
-            sort_dependencies = True
-            if self.io.is_interactive():
-                self.line("")
-        if sort_dependencies is True:
+        if config.get("dependencies.sort") is True:
             requirements = OrderedDict(sorted(requirements.items()))
             dev_requirements = OrderedDict(sorted(dev_requirements.items()))
-            local_config_file = TOMLFile(Path.cwd() / "poetry.toml")
-            local_config_file_contents = document()
-            local_config_file_contents.update({"dependencies": {"sort": True}})
-            local_config_file.write(local_config_file_contents)
 
         layout_ = layout("standard")(
             name,
--- a/src/poetry/console/commands/init.py
+++ b/src/poetry/console/commands/init.py
@@ -2,6 +2,7 @@ from __future__ import annotations
 
 import sys
 
+from collections import OrderedDict
 from pathlib import Path
 from typing import TYPE_CHECKING
 from typing import Any
@@ -11,6 +12,7 @@ from typing import Union
 
 from cleo.helpers import option
 from packaging.utils import canonicalize_name
+from tomlkit import document
 from tomlkit import inline_table
 
 from poetry.console.commands.command import Command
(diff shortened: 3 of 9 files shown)


dlech:
It would be great if this was enabled by default. Or if that is not acceptable, add a setting to `poetry.toml` so that it can be enforced on a per-project basis rather than having to train everyone to always type in `--sorted`.

br3ndonland:
> It would be great if this was enabled by default. Or if that is not acceptable, add a setting to `poetry.toml` so that it can be enforced on a per-project basis rather than having to train everyone to always type in `--sorted`.

Thanks for the feedback! I agree, I would like to see sorting become the default behavior.

The `poetry.toml` config setting is a great idea - maybe something like `poetry config dependencies.sort true` or `export POETRY_DEPENDENCIES_SORT=true`.

br3ndonland:
The latest commit will remove the `--sorted (-S)` option. Instead, dependency sorting can be configured with `poetry config dependencies.sort true` or `export POETRY_DEPENDENCIES_SORT=true`. If `true`, dependencies will be sorted alphabetically when adding, keeping `python` at the top of the list.

The `dependencies.sort` option will default to `false` to avoid changing the previous default behavior, but as mentioned in the comments, dependency sorting could potentially be enabled by default in the future.

I'll look forward to some input from the maintainers.

donegjookim (dismissed):

donegjookim:
@br3ndonland Good work!

br3ndonland:
Dependency sorting has now been updated for Poetry dependency groups (added in python-poetry/poetry#4260 and [1.2.0a2](https://github.com/python-poetry/poetry/releases/tag/1.2.0a2)).

tests/console/commands/test_add.py
```diff
@@ -693,6 +696,63 @@ def test_add_constraint_with_platform(
...
+        expected_content.move_to_end("python", last=False)
+        poetry_content = OrderedDict(
+            app.poetry.file.read()["tool"]["poetry"]["dependencies"]
+        )
+    else:
+        poetry_content = OrderedDict(
+            app.poetry.file.read()["tool"]["poetry"]["group"][group]["dependencies"]
+        )
+
+    assert expected_content == poetry_content
```
radoering, line 753:
```suggestion
    assert poetry_content == expected_content
```

We prefer actual-expected order. (See #5129 for details.)

docs/configuration.md
```diff
@@ -116,6 +116,13 @@ Defaults to one of the following directories:
 - Windows: `C:\Users\<username>\AppData\Local\pypoetry\Cache`
 - Unix:    `~/.cache/pypoetry`
 
+### `dependencies.sort`
+
+**Type**: boolean
+
+Sort dependencies alphabetically when adding.
```
radoering, line 123:
```suggestion
Sort dependencies in `pyproject.toml` alphabetically when adding.
```

radoering (changes requested):
In general, looks good. 👍

Just one thing (besides the nitpicks below): I think the sort option should be considered in `poetry init`, too. Can you add this, please? 🙏

tests/console/commands/test_add.py
br3ndonland: Fixed by 5ea016d. Thank you for helping to standardize the test assertions.

docs/configuration.md
br3ndonland: Fixed by 23fe532.

br3ndonland:
> In general, looks good. 👍
> 
> Just one thing (besides the nitpicks below): I think the sort option should be considered in `poetry init`, too. Can you add this, please? 🙏

Thank you for your review!

I'm happy to update `poetry init` to support dependency sorting. I've taken a first pass at it with the latest commit. Feel free to make further suggestions if you would like to see it implemented differently.

src/poetry/console/commands/init.py
```diff
@@ -213,6 +223,20 @@ def handle(self) -> int:
...
+            sort_dependencies = True
+            if self.io.is_interactive():
+                self.line("")
+        if sort_dependencies is True:
+            requirements = OrderedDict(sorted(requirements.items()))
+            dev_requirements = OrderedDict(sorted(dev_requirements.items()))
+            local_config_file = TOMLFile(Path.cwd() / "poetry.toml")
+            local_config_file_contents = document()
+            local_config_file_contents.update({"dependencies": {"sort": True}})
+            local_config_file.write(local_config_file_contents)
```
br3ndonland, line 238: The dependency sorting in `poetry init` currently works like this:

- For noninteractive use, add a `--sort-dependencies` flag.
- For interactive use, after the user has added all their dependencies, ask them if they want to sort the dependencies alphabetically.
- If dependency sorting is selected (either interactive or noninteractive), Poetry will write a `poetry.toml` file to the new project with `dependencies.sort` set to `true`.

Happy to consider suggestions on the implementation.

finswimmer: Hey,

in my opinion we shouldn't introduce another flag or questions here and just respect whatever is set in the config.

br3ndonland: Sure, that's fine. I reverted the new flag.

By "the config," you mean the user's global Poetry config?

I guess the approach would be to read the global config, and if there's a `dependencies.sort` setting there, use that.

src/poetry/console/commands/init.py
```diff
@@ -91,6 +93,7 @@ def handle(self) -> int:
                 )
                 return 1
 
+        config = Config()
```
br3ndonland, line 96: Updated to read the global config, instead of adding a new flag to `init` (https://github.com/python-poetry/poetry/pull/3996#discussion_r851708290)

abn:
One issue I have with this is that this is a client level configuration for a project level preference. 

For example, in poetry, we specify poetry-core and default plugins at top right after python constraint.

Personally, I like an always sorted model, but I don't think this is desirable across the board.

radoering:
> One issue I have with this is that this is a client level configuration for a project level preference.

Can't we just create a [local project specific](https://python-poetry.org/docs/master/configuration/#local-configuration) `config.toml` and explicitly enable/disable sorting if it's relevant for the project. Or should the project specific `config.toml` be considered a user specific file, too?

abn:
Good question. I'm not sure if I have a definitive answer here. While that will work, I don't think we've considered project `config.toml` something we expect to be checked in. Additionally, I'd be reluctant to suggest folks who don't want this to disable it in yet another project file. And yes, historically this has been something for client side preferences and not project level.

br3ndonland:
Dependency sorting has now been updated for the renamed dependency group (`default` -> `main`), introduced in python-poetry/poetry#5465 and [1.2.0b2](https://github.com/python-poetry/poetry/releases/tag/1.2.0b2).

Kavan72:
@abn @br3ndonland any update on this?

radoering:
> One issue I have with this is that this is a client level configuration for a project level preference.

Unfortunately, I think we need consensus on this. Is it ok to have a client level configuration for this or should it be a project specific option in `pyproject.toml`. Especially @abn

br3ndonland:
Sure, happy to discuss further.

> One issue I have with this is that this is a client level configuration for a project level preference.

I don't understand the issue. Isn't this just a commentary on Poetry's config system in general? The feature proposed by this PR works in the same way as any other Poetry config setting.

If you want it to be a "project level setting," users would follow the paradigm suggested in the [current docs on local configuration](https://python-poetry.org/docs/master/configuration/#local-configuration):

> Your local configuration of Poetry application is stored in the `poetry.toml` file, which is separate from `pyproject.toml`.

So users would run `poetry config dependencies.sort true --local`, and check in the `poetry.toml`.

src/poetry/console/commands/add.py
```diff
@@ -230,6 +231,15 @@ def handle(self) -> int:
                 )
             )
 
+        if self.poetry.config.get("dependencies.sort") is True:
+            sorted_dependencies = OrderedDict(sorted(section.items()))
```
andrei-shabanski, line 235: One more issue here: casting `section` to dictionary causes loss of comments in project file. That might be critical for some projects. Please, try to fix the issue

br3ndonland: > One more issue here: casting `section` to dictionary causes loss of comments in project file. That might be critical for some projects. Please, try to fix the issue

If comments were preserved, where would they go? We could probably preserve inline comments, but single line comments would be more difficult.

The most common use case for comments that I have seen is breaking up dependencies into subsections, like this:

```toml
# tests
pytest = "*"
# docs
mkdocs = "*"
```

If the dependencies were sorted, where would the comments go? It would probably end up like this:

```toml
# tests
# docs
mkdocs = "*"
pytest = "*"
```

A more effective approach might be to remove the comments and separate these commented sections into [Poetry 1.2 dependency groups](https://python-poetry.org/docs/managing-dependencies/#dependency-groups).

```toml
[tool.poetry.group.test.dependencies]
pytest = "*"

[tool.poetry.group.docs.dependencies]
mkdocs = "*"
```

I'm happy to try preserving comments if there is a suggested data structure from TOML Kit.

andrei-shabanski: Also comments might be used to temporarily disable a dependency or explain a dependency directly below the comment.
In these cases, it's quite important to keep those comments. I think it would be good to make the comments stick to the next dependency line:

```
[tool.poetry.dependencies]
simplejson = "^3.17.5"  # used in requests to serialize decimals
# Consider to replace requests with httpx 
# to improve performance.
# httpx = "^0.23.0"
requests = "^2.26.0"
fastapi = "^1"
```

should turn into

```
[tool.poetry.dependencies]
fastapi = "^1"
# Consider to replace requests with httpx 
# to improve performance.
# httpx = "^0.23.0"
requests = "^2.26.0"
simplejson = "^3.17.5"  # used in requests to serialize decimals
```

Your example is about a case that can be resolved using poetry functional, as you mentioned. Users should be responsible for proper use of Poetry. So I would keep comments in this case as well

andrei-shabanski: I don't know how to use tomlkit properly. But I can just share my thoughts.

It seems comment lines are stored in [`section.value.body`](https://github.com/sdispater/tomlkit/blob/master/tomlkit/container.py#L46) list with `None` as a key and `Comment(..)` as a value. You can't get an access to the comment lines via `section.items()` method. I didn't find any other way to access to the comments but `section.value.body`. And I don't see methods to remove comment lines from the section (neither `section.remove()` nor `section.clear()` don't remove the comments). So, you should re-create an entire section and append sorted lines into it, or try to sort `section.value.body` in-place.

br3ndonland:
Closing this PR because it has been open for almost two years, the maintainers have still not reached consensus, I am not willing to endlessly maintain this branch, I am tired of dealing with the flaky Cirrus CI jobs, and I am no longer using Poetry (br3ndonland/inboard#56).

PR #3999, opened 2021-04-27
Replace install with add in outdated doc

cauebs:


cauebs:
Actually, I think I'm mistaken here. I just noticed `poetry install` doesn't install all extras for the current package, and it _does_ have an `--extras` option.

PR #4003, opened 2021-04-28
Change keyring dependency to >=21.2.0

NeilGirdhar:
keyring appears to be increasing the major version for minor changes every few months: https://github.com/jaraco/keyring/compare/v22.4.0...v23.0.0

Perhaps it would be better to use a >= dependency here?

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -45,7 +45,7 @@ tomlkit = ">=0.7.0,<1.0.0"
 pexpect = "^4.7.0"
 packaging = "^20.4"
 virtualenv = "^20.4.3"
-keyring = "^21.2.0"
+keyring = ">=21.2.0"
 entrypoints = "^0.3"
 importlib-metadata = {version = "^1.6.0", python = "<3.8"}
 dataclasses = {version = "^0.8", python = "~3.6"}
--- a/poetry.lock
+++ b/poetry.lock
@@ -161,7 +161,7 @@ python-versions = ">=3.6, <3.7"
 
 [[package]]
 name = "deepdiff"
-version = "5.2.3"
+version = "5.5.0"
 description = "Deep Difference and Search of any Python object/data."
 category = "dev"
 optional = false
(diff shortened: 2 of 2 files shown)


pyproject.toml
```diff
@@ -41,7 +41,7 @@ tomlkit = ">=0.7.0,<1.0.0"
 pexpect = "^4.7.0"
 packaging = "^20.4"
 virtualenv = "^20.4.3"
-keyring = "^21.2.0"
+keyring = ">=21.2.0"
```
abn, line 44:
```suggestion
keyring = ">=21.2.0,<24.0.0"
```
Might be better to keep an upper limit. Additionally, we should probably update the lock file as well to get 23.x tested.

abn (changes requested):

pyproject.toml
NeilGirdhar: The keyring project is not using semantic versioning so there's no reason to believe that version 24 is a "major" step from version 23.  At their "major" release cadence of less than one month, pip will be reporting errors again next month with this upper limit.

In this case, I think it would be better to leave it open-ended and clamp the range in case there's a breakage.

It may also be worth asking the keyring maintainers to start using semantic versioning.

abn: I am honestly on the fence on this. But assuming keyring is not following semver, this makes sense. I had a brief look on their changelog and it seems breaking changes are not highlighted anyway.

Poetry being an application and "expected" installations being in isolated environments, pip warnings should not really impact poetry installs, outside of distro installs. 

TLDR: I agree and we can try the min version see how much pain it actually brings :)

abn (approved):

ryandesign:
Please backport this fix to the 1.1 branch. See https://trac.macports.org/ticket/64429.

NeilGirdhar:
@ryandesign https://github.com/python-poetry/poetry/pull/5041 Done as requested.  Would you please monitor the pull request and take charge of interacting with the Poetry team with respect to any concerns or questions they may have?

Issue #3225, opened 2020-10-16
Support for pip --constraint


robsonpeixoto:
- [x] I have searched the [issues](https://github.com/python-poetry/poetry/issues) of this repo and believe that this is not a duplicate.
- [x] I have searched the [documentation](https://python-poetry.org/docs/) and believe that my question is not covered.

## Feature Request

Tools like airflow has some constraints that need be respected, [for exemplo ](https://raw.githubusercontent.com/apache/airflow/constraints-1.10.12/constraints-3.7.txt), when are adding dependencies in a docker image.

This project could add `constraints` that accept a list of FQDN with all constraints. For example:

```toml
contrains = [
  "https://raw.githubusercontent.com/apache/airflow/constraints-1.10.12/constraints-3.7.txt",
  "file://{CWD}/constraints.txt"
]
```

abn:
@robsonpeixoto it is not clear what you mean here. I doubt this is within the scope of Poetry. How do you imagine thise being used when generating the lock file? Feels like you are essentially removing the resolver from the equation.

If airflow needs these pinned dependencies, shouldn't this be done so within the package's core metadata? It looks like it already is.

See snipped output below.
```console
$ curl -sL https://pypi.org/pypi/apache-airflow/json | jq '.info.requires_dist'
[
  "google-auth-httplib2 (>=0.0.1) ; extra == 'gcp'",
...
  "google-cloud-vision (<2.0.0,>=0.35.2) ; extra == 'gcp'",
  "grpcio-gcp (>=0.2.2) ; extra == 'gcp'",
  "pandas-gbq ; extra == 'gcp'",
  "PyOpenSSL ; extra == 'gcp_api'",
  "google-api-python-client (<2.0.0,>=1.6.0) ; extra == 'gcp_api'",
...
  "grpcio-gcp (>=0.2.2) ; extra == 'gcp_api'",
  "pandas-gbq ; extra == 'gcp_api'",
  "Flask-OAuthlib (>=0.9.1) ; extra == 'github_enterprise'",
  "oauthlib (!=2.0.3,!=2.0.4,!=2.0.5,<3.0.0,>=1.1.2) ; extra == 'github_enterprise'",
  "requests-oauthlib (==1.1.0) ; extra == 'github_enterprise'",
  "Flask-OAuthlib (>=0.9.1) ; extra == 'google_auth'",
[...]

PR #4005, opened 2021-04-29
Support constraints file when resolving versions

honnix:
# Pull Request Check List

Resolves: #3225

This depends on https://github.com/python-poetry/poetry-core/pull/172

This PR tries to address #3225 by pulling in all constraint dependencies as root dependencies when resolving version, and removing all unreachable ones afterwards.

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -16,11 +16,11 @@ python-versions = ">=2.7, !=3.0.*, !=3.1.*, !=3.2.*, !=3.3.*"
 
 [[package]]
 name = "attrs"
-version = "21.1.0"
+version = "21.2.0"
 description = "Classes Without Boilerplate"
 category = "dev"
 optional = false
-python-versions = ">=2.7, !=3.0.*, !=3.1.*, !=3.2.*, !=3.3.*"
+python-versions = ">=2.7, !=3.0.*, !=3.1.*, !=3.2.*, !=3.3.*, !=3.4.*"
 
 [package.extras]
 dev = ["coverage[toml] (>=5.0.2)", "hypothesis", "pympler", "pytest (>=4.3.0)", "six", "mypy", "pytest-mypy-plugins", "zope.interface", "furo", "sphinx", "sphinx-notfound-page", "pre-commit"]
@@ -403,9 +403,9 @@ importlib-metadata = {version = ">=1.7.0", markers = "python_version < \"3.8\""}
 
 [package.source]
 type = "git"
-url = "https://github.com/python-poetry/poetry-core.git"
-reference = "master"
-resolved_reference = "8b1cc5ea8d4557a4d215c873ebf2acfe282ad64d"
+url = "https://github.com/honnix/poetry-core.git"
+reference = "constraints"
+resolved_reference = "7807177a3237c4fb924b8458e0aa6777673d0691"
 
 [[package]]
 name = "pre-commit"
@@ -705,7 +705,7 @@ testing = ["pytest (>=4.6)", "pytest-checkdocs (>=1.2.3)", "pytest-flake8", "pyt
 [metadata]
 lock-version = "1.1"
 python-versions = "^3.6"
-content-hash = "e38d34da68dcd63ef9f9d999f60b8e19d8da7f988031c3a6e6865d51c1fadeda"
+content-hash = "bcc09861037d0f44d8e9a446952470dc3e10470b07a30ca4fb1cb3d42f08d092"
 
 [metadata.files]
 appdirs = [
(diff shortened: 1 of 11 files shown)


honnix:
@abn @sdispater What do you think of this? Not sure why all the CI actions are not moving but I unit test it locally. Thanks and  please let me know whether this change makes sense.

abn:
Hi @honnix the tests are not running due to https://docs.github.com/actions/managing-workflow-runs/approving-workflow-runs-from-public-forks, I have approved the runs.

As for the change itself, I'll be honest, I am not really enthused about this. I cannot see this solving a real problem. This fundamentally breaks the determinsm poetry brings to your environment. 

Further, the constraints file can be used with a `pip` install of the poetry projects in environments that requirement them anyway. If poetry is installing the package with the proposed constraints, this means you are triggerring (or will have to trigger) a dependency resolution leading to effectively a new locked versions.

I wonder if being able to do something like this is what is really required.

```ini
[testenv]
whitelist_externals = poetry
install_command =
    pip install -U -c{env:UPPER_CONSTRAINTS_FILE:https://git.openstack.org/cgit/openstack/requirements/plain/upper-constraints.txt} {opts} {packages}
commands =
    poetry install -vv --dev-only --no-root
    poetry run pytest {posargs} tests/
```

Note that this right now will not work since `--dev-only` removes all packages that are not marked dev.

honnix:
@abn Thanks for looking at this PR. Here are some thoughts I have in mind and please take a look whether they make sense.

> This fundamentally breaks the determinsm poetry brings to your environment

The constraints files don't result installing any additional packages that are not declared in `pyproject.toml`, and the purpose of it is to enforce package version constraints globally. If I understood it correctly, running `poetry install/lock` etc. with the same content of `pyproject.toml` and constraints files, and the same set of packages available in repositories, will result the same `poetry.lock`. So it is still deterministic.

> If poetry is installing the package with the proposed constraints, this means you are triggerring (or will have to trigger) a dependency resolution leading to effectively a new locked versions.

If I did it correctly, when a `poetry.lock` exists, `poetry install` should just install things declared in `poetry.lock` faithfully; constraints files will not be read and used to automatically regenerate `poetry.lock` and it's up to the user to re-lock. If we worry about constraints file content changes that users are not aware of, we could version a constraints file, so for any given constraints file once it's generated it becomes immutable.

> I cannot see this solving a real problem

For a CLI package to be distributed to a public repository, this is true because very often a CLI package will pin all dependencies and users don't need to care what to be installed, so there is no reason for CLI developers to use a constraints file.

When coming to a CLI built and distributed to an organization using private repository, it is possible due to security reasons that a certain version range of a package or the package completely is banned globally in the organization, and anything developed there should respect the global constraints. Developers of the CLI can simply pull in the constraints files and they will be informed when a solution cannot be found so they know there are some versions/packages they cannot use.

For a library to be distributed to a public repository, this is true most of the time because downstream consumers of the lib will likely not be aware of constraints files used by the library, and they only rely on the metadata of the package. So the developers of the library don't have incentive or reason to use a constraints file.

However it could also be the case that the library is a part of a collection of libraries that are supposed to work together in a single runtime environment (where everything being constrained by a set of constraints files), then it will be useful for the library developers to use the same set of constraints file so they are confident that the library will never violate dependency constraints that may result runtime failure. For example if the library developers update `pyproject.toml` for dependency `A` to be `A = ^3.0.0` while there exists a constraints file saying `A = ^2.0.0`, they will not be able to find a solution and they will be informed that their library cannot use a later version of `A` because it may fail at runtime and may also cause transitive dependency troubles for their downstream consumers. This scenario is much more likely to happen in an organization using private repositories, where libraries are maintained by different teams but are supposed to work together in the same backend service or data pipeline.

For a backend service or data pipeline, this is actually extremely useful because figuring out a set of dependencies that for sure can work at runtime is not trivial. With a set of constraints files that are manual crafted by a group of engineers in an organization, the backend service or pipeline developers will not need to care about dependency versions, so for many cases they could just declare `A = *` in `pyproject.toml` and let the constraints files decide what a version of A should be. Of course the constraints files cannot cover all possible dependencies, but they will gradually cover the majority of use cases in the organization, leaving only a few things for a backend service or pipeline developer to figure out. This idea is nothing new in literature and has been proven to be a good success. For example [Maven](https://maven.apache.org/guides/introduction/introduction-to-dependency-mechanism.html#bill-of-materials-bom-poms) and [Gradle](https://docs.gradle.org/current/userguide/java_platform_plugin.html) in Java land, and `pip`. _`pip` set off to build this feature for exactly the same reason (I have a teammate who was part of the team built this feature in `pip`)._

>  wonder if being able to do something like this is what is really required.

Yeah it might be possible to hack it to work with a pip constraints file and use pip directly, but to me that is sub-optimal and kinda defeats the purpose of using poetry as a single tool.

lastthyme (approved):

honnix:
@abn What do you think of those thoughts? Should I keep my hope that this PR might get accepted? With the change throughput of this project, this PR could quickly go outdated, so if it is very unlikely to get it accepted, I will not spend additional effort resolving all the conflicts. Thanks.

honnix:
@abn Any update on this? I'm totally fine withdrawing this proposal. Please kindly let me know. Thanks.

abn:
@honnix I haven't really looked into it yet. Unfortunately haven't had the time to focus on poetry the last couple of weeks. Will look at your response as soon as I have some time.

honnix:
I think it is time for me to close this. Thanks.

Issue #3952, opened 2021-04-15
1.0.3: tests/ directory missing in idst pypy tar ball


kloczek:
Looks like something is wrong with https://pypi.io/packages/source/p/poetry-core/poetry-core-1.0.3.tar.gz because in that archive is not included tests/ directory.

finswimmer:
Why it should?

kloczek:
Obvioely to have test suite included in dist tar ball which wou;d be possible to fire on for example packaging poetry to confirm that some basis functionalieties of the poetry module or some other build env components used by poetry test suite are ok :)

eli-schwartz:
As implemented in #1133 it is possible for poetry-based projects to specify files which are added to sdist but not wheel. https://python-poetry.org/docs/pyproject/#packages

The intention is for people to be able to include test data and the like inside sdist archives, which is exceedingly useful for third-party ecosystems e.g. running integrated unittests when shipping packages... rather than needing to do painful contortions with the github repo and post processing.

It is wonderful that other projects using poetry can now ship tests in their source code releases. It would be even nicer if the sdist releases for "poetry" and "poetry-core" themselves could make use of this feature!

kwshi:
More reasons to include tests in the sdist tarball:
- Including tests in package tarballs is common, and arguably _standard_, practice among Python packages (examples: [requests](http://ix.io/3lln), [black](http://ix.io/3llp), [pandas](http://ix.io/3llq)).
- Doing so makes it possible for third-party package repository maintainers (Arch community repos, Void, SUSE) to run tests as part of the package release flow (as @eli-schwartz mentioned above).  There is also a related discussion in #1109.

eli-schwartz:
Please note that the entire rationale for adding a new feature to poetry was, specifically, that it is in fact useful for project developers to utilize the feature. Poetry should be brave enough to dogfood this UX.

The PyPI releases contain both sdist and wheel. The point of a wheel is to be very quick to install prebuilt code that is lean and small to download. The point of an sdist is to be ***source code***. Source code should include everything necessary to build and test the project, because if not, then why on earth would anyone ever use it instead of the wheel???

Why is it so necessary for the process of trying to help more people get a copy of the poetry software, to be pain and suffering? All I want to do is help people get poetry. Part of doing that is to do end to end tests and verify that poetry *works* in the ISV ecosystem, just like we do for other python projects, and just like we do for software written in C/C++/go/rust and built using cmake or autotools or what have you.

[...]

PR #4007, opened 2021-04-30
dist: add tests to sdists

abn:
Resolves: #3952

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -9,6 +9,10 @@ license = "MIT"
 
 readme = "README.md"
 
+include = [
+    { path = "tests", format = "sdist" }
+]
+
 homepage = "https://python-poetry.org/"
 repository = "https://github.com/python-poetry/poetry"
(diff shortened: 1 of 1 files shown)


eli-schwartz:
Thank you! Would you be able to do the same for poetry-core too?

kwshi:
I have opened python-poetry/poetry-core#173 implementing this.  (By the way, is the `format = "sdist"` in the include clause needed?  Based on what I can tell, even if I omitted it the `tests` module doesn't end up in the resulting wheel anyway.)

sdispater (approved):

abn:
> I have opened [python-poetry/poetry-core#173](https://github.com/python-poetry/poetry-core/pull/173) implementing this. (By the way, is the `format = "sdist"` in the include clause needed? Based on what I can tell, even if I omitted it the `tests` module doesn't end up in the resulting wheel anyway.)

I would prefer that we are explicit about the intention. Hence why we specify the format.

PR #4008, opened 2021-04-30
use src layout

abn:
Does what it says :smile: No functional changes, `poetry` has been moved to `src/poetry`.


neersighted:
For those of us who are more ignorant: what are the benefits/drawbacks of the `src` layout?

abn:
I think Hynek did a better job than I could to present the trade offs. 

https://hynek.me/articles/testing-packaging/

neersighted:
> 
> 
> I think Hynek did a better job than I could to present the trade offs.
> 
> https://hynek.me/articles/testing-packaging/

Let's do the same in core?

neersighted (approved):

PR #4011, opened 2021-04-30
env: default to enabling pip/wheels/setuptools

abn:
For project virtual environments, default to enabling pip, setuptools
and wheel packages to retain existing stable behaviour to prevent 
unexpected breakages caused by development environments making 
assumptions of base package availability in virtual environments.

Poetry itself does not require the use of these packages and will 
execute correctly within environments that do not have these packages.

This change retains the ability to manage these packages as direct 
project dependency as introduced in #2826. All poetry internal 
execution of pip is retaining the use of the wheel embedded within 
the virtualenv package used by poetry.

In cases where a one of these reserved packages are being managed as a
project dependency, the will be treated as any other project 
dependency. Executing `poetry install --remove-untracked` will not 
remove any of these reserved packages. However, `poetry add pip` and 
`poetry remove pip` will trigger the update and removal of `pip` 
respectively.

Relates-to: #2826
Relates-to: #3916

Closes: #3920

### DIFF
--- a/poetry/puzzle/solver.py
+++ b/poetry/puzzle/solver.py
@@ -63,9 +63,30 @@ class Solver:
         self._overrides = []
         self._remove_untracked = remove_untracked
 
+        self._preserved_package_names = None
+
     @property
     def provider(self) -> Provider:
         return self._provider
+
+    @property
+    def preserved_package_names(self):
+        if self._preserved_package_names is None:
+            self._preserved_package_names = {
+                self._package.name,
+                *Provider.UNSAFE_PACKAGES,
+            }
+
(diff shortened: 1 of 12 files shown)


poetry/puzzle/solver.py
```diff
@@ -63,10 +65,32 @@ def __init__(
         self._overrides = []
         self._remove_untracked = remove_untracked
 
+        self._poetry_config = config
```
sdispater, line 68: Is this still needed? I don't see where it's used in the solver .

lastthyme (approved):

sdispater (approved):
Looks good to me 👍

PR #4024, opened 2021-05-02
fix a bunch of typing errors

danieleades:
this pull request takes the total number of mypy errors from 656 to 617

it's a start...

### DIFF
--- a/poetry/console/commands/env_command.py
+++ b/poetry/console/commands/env_command.py
@@ -1,4 +1,5 @@
 from typing import TYPE_CHECKING
+from typing import Optional
 
 from .command import Command
 
@@ -14,8 +15,8 @@ class EnvCommand(Command):
         super(EnvCommand, self).__init__()
 
     @property
-    def env(self) -> "Env":
+    def env(self) -> Optional[Env]:
         return self._env
 
-    def set_env(self, env: "Env") -> None:
+    def set_env(self, env: Env) -> None:
         self._env = env
--- a/poetry/console/commands/env_command.py
+++ b/poetry/console/commands/env_command.py
@@ -1,11 +1,8 @@
-from typing import TYPE_CHECKING
 from typing import Optional
 
+from poetry.utils.env import Env
+
 from .command import Command
-
-
-if TYPE_CHECKING:
-    from poetry.utils.env import Env
 
 
 class EnvCommand(Command):
--- a/poetry/installation/installer.py
+++ b/poetry/installation/installer.py
@@ -2,6 +2,7 @@ from typing import TYPE_CHECKING
 from typing import Iterable
 from typing import List
 from typing import Optional
+from typing import Sequence
 from typing import Union
 
 from cleo.io.io import IO
@@ -499,7 +500,7 @@ class Installer:
         self._installer.remove(operation.package)
 
     def _populate_local_repo(
-        self, local_repo: Repository, ops: List[Operation]
+        self, local_repo: Repository, ops: Sequence[Operation]
     ) -> None:
         for op in ops:
             if isinstance(op, Uninstall):
@@ -514,7 +515,7 @@ class Installer:
 
     def _get_operations_from_lock(
         self, locked_repository: Repository
-    ) -> List[Operation]:
+    ) -> Sequence[Operation]:
         installed_repo = self._installed_repository
         ops = []
 
@@ -543,7 +544,7 @@ class Installer:
 
         return ops
 
-    def _filter_operations(self, ops: List[Operation], repo: Repository) -> None:
+    def _filter_operations(self, ops: Sequence[Operation], repo: Repository) -> None:
         extra_packages = self._get_extra_packages(repo)
         for op in ops:
             if isinstance(op, Update):
--- a/poetry/mixology/incompatibility.py
+++ b/poetry/mixology/incompatibility.py
@@ -1,3 +1,4 @@
+from typing import Callable
 from typing import Dict
 from typing import Iterator
 from typing import List
@@ -459,7 +460,7 @@ class Incompatibility:
             term.dependency.pretty_name, term.dependency.pretty_constraint
         )
 
-    def _single_term_where(self, callable: callable) -> Optional[Term]:
+    def _single_term_where(self, callable: Callable[[Term], bool]) -> Optional[Term]:
         found = None
         for term in self._terms:
             if not callable(term):
--- a/poetry/mixology/incompatibility.py
+++ b/poetry/mixology/incompatibility.py
@@ -272,18 +272,18 @@ class Incompatibility:
         other_line: Optional[int],
     ) -> Optional[str]:
         if len(self._terms) == 1 or len(other.terms) == 1:
-            return
+            return None
 
         this_positive = self._single_term_where(lambda term: term.is_positive())
         if this_positive is None:
-            return
+            return None
 
         other_positive = other._single_term_where(lambda term: term.is_positive())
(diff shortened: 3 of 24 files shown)


abn:
I do not think most of these are required. The only one that I can see as being the right change is at https://github.com/python-poetry/poetry/pull/4024/commits/943fb358ddb6effe4b8da23fe7969765b7610405#diff-89c37d9db9d559ea6d241ff3fcbee2053643e688e99a177327fd0ad15580ba26R239.

danieleades:
> I do not think most of these are required. The only one that I can see as being the right change is at [943fb35#diff-89c37d9db9d559ea6d241ff3fcbee2053643e688e99a177327fd0ad15580ba26R239](https://github.com/python-poetry/poetry/commit/943fb358ddb6effe4b8da23fe7969765b7610405#diff-89c37d9db9d559ea6d241ff3fcbee2053643e688e99a177327fd0ad15580ba26R239).

These are *all* required in the sense that the original code is an error with respect to the current mypy configuration.

The `return` errors *could* be suppressed with a config change.

There's a number of cases where I've swapped variant types for invariant types. This is required for correctness. See https://mypy.readthedocs.io/en/latest/generics.html#variance-of-generic-types

poetry/installation/installer.py
```diff
@@ -477,7 +478,7 @@ def _execute_uninstall(self, operation: Uninstall) -> None:
         self._installer.remove(operation.package)
 
     def _populate_local_repo(
-        self, local_repo: Repository, ops: List[Operation]
+        self, local_repo: Repository, ops: Sequence[Operation]
```
danieleades, line 481: @abn
here's an example of the use of a 'covariant' type (required for *correctness*)

the `_populate_local_repo` method accepts `ops: List[Operation]`, however `Operation` is a base class. In practice we expect that this method should accept a list of some derived class (such as `Uninstall`, for example).

**the problem is that `List[Subclass]` is *not* equivalent to `List[Baseclass]`** (see these docs - http://mypy.readthedocs.io/en/latest/generics.html#variance-of-generic-types)

however `Sequence[Subclass]` *is* equivalent to `Sequence[Baseclass]`. Similarly, `Iterable[Subclass]` *is* equivalent to `Iterable[Baseclass]`.

poetry/utils/extras.py
```diff
@@ -14,7 +14,7 @@ def get_extra_package_names(
     packages: Sequence["Package"],
     extras: Mapping[str, List[str]],
     extra_names: Sequence[str],
-) -> Iterator[str]:
+) -> Iterable[str]:
```
danieleades, line 17: @abn 
this too is a *correctness* bug. the returned value of this function is *not* an iterator, but it *is* a type than can be converted *into* an iterator. That's the distinction between `Iterator` and `Iterable`

poetry/utils/env.py
```diff
@@ -1117,11 +1117,13 @@ def site_packages(self) -> SitePackages:
     def usersite(self) -> Optional[Path]:
         if "usersite" in self.paths:
             return Path(self.paths["usersite"])
+        return None
```
danieleades, line 1120: @abn 
explicitly returning `None` is required to conform with PEP8 and to satisfy mypy

from PEP8-

> Be consistent in return statements. Either all return statements in a function should return an expression, or none of them should. If any return statement returns an expression, any return statements where no value is returned should explicitly state this as return None, and an explicit return statement should be present at the end of the function (if reachable)

it's not without controversy, see this discussion - https://github.com/python/mypy/issues/7511

it depends what the goal is. The goal of this pull request is not strictly correctness, it is to reduce the amount of noise that mypy makes, in the hope that this is a small step in the direction of actually being able to effectively use Mypy to lint this codebase (and *also* correctness!).

In a perfect world, the number of type errors would be 0, and then mypy could be added as a pre-commit hook or CI job to ensure type errors are not reintroduced. Baby steps.

poetry/console/commands/env_command.py
```diff
@@ -1,10 +1,8 @@
-from typing import TYPE_CHECKING
-
-from .command import Command
+from typing import Optional
 
+from poetry.utils.env import Env
 
-if TYPE_CHECKING:
-    from poetry.utils.env import Env
```
danieleades, line 7: this pattern is used in the case that an unconditional import would cause a circular import. This isn't required here.

poetry/mixology/incompatibility.py
```diff
@@ -459,14 +460,14 @@ def _terse(self, term: Term, allow_every: bool = False) -> str:
             term.dependency.pretty_name, term.dependency.pretty_constraint
         )
 
-    def _single_term_where(self, callable: callable) -> Optional[Term]:
+    def _single_term_where(self, callable: Callable[[Term], bool]) -> Optional[Term]:
```
danieleades, line 463: the built-in `callable` can't be used for type annotations

poetry/packages/project_package.py
```diff
@@ -12,7 +12,7 @@
 class ProjectPackage(_ProjectPackage):
     def set_version(
         self, version: Union[str, "Version"], pretty_version: Optional[str] = None
-    ) -> "ProjectPackage":
+    ) -> None:
```
danieleades, line 15: this method doesn't return...

poetry/utils/env.py
```diff
@@ -1327,7 +1329,7 @@ def _bin(self, bin: str) -> str:
 
         return str(bin_path)
 
-    def __eq__(self, other: "Env") -> bool:
+    def __eq__(self, other: object) -> bool:
```
danieleades, line 1332: this is required to satisfy Liskov subsitution principle- namely that the derived class's `__eq__` should have the same, or a less restrictive function signature for any overridden baseclass methods.

poetry/utils/env.py
```diff
@@ -1578,17 +1580,19 @@ def __init__(
     def get_pip_command(self, embedded: bool = False) -> List[str]:
         return [self._bin("python"), self.pip_embedded if embedded else self.pip]
 
-    def _run(self, cmd: List[str], **kwargs: Any) -> int:
+    def _run(self, cmd: List[str], **kwargs: Any) -> Optional[int]:
```
danieleades, line 1583: this change, while correct, actually increases the total number of errors, since the `None` case is not handled in the calling code.

poetry/utils/shell.py
```diff
@@ -61,7 +62,7 @@ def get(cls) -> "Shell":
 
         return cls._shell
 
-    def activate(self, env: VirtualEnv) -> None:
+    def activate(self, env: VirtualEnv) -> Optional[int]:
         if WINDOWS:
             return env.execute(self.path)
```
danieleades, line 67: this method return `Optional[int]`

danieleades:
Ping @abn

danieleades:
@abn do you mind approving the workflows?

danieleades:
rebased. @abn, it's that time again...

skovhus (commented):
This looks very solid and I cannot see any controversial changes here. I hope this gets merged soon.

danieleades:
> This looks very solid and I cannot see any controversial changes here. I hope this gets merged soon.

i'm not holding my breath...

danieleades:
@abn  there's over 700 typing errors now in this repo, even after these changes. Is there any appetite for reducing/resolving these?

neersighted (approved):
Thanks for this work! I'm very interesting in getting the typing issues reduced and getting mypy into the Code Quality checks. Feel free to reach out offline (e.g. Discord) to coordinate on this.

Issue #3412, opened 2020-11-25
poetry install exits with exit code 1 when run in a docker container


achow101:
- [x] I am on the [latest](https://github.com/python-poetry/poetry/releases/latest) Poetry version.
- [x] I have searched the [issues](https://github.com/python-poetry/poetry/issues) of this repo and believe that this is not a duplicate.
- [x] If an exception occurs when executing a command, I executed it again in debug mode (`-vvv` option).

- **OS version and name**: Debian stretch docker container, Arch host
- **Poetry version**: 1.1.4
- **Link of a [Gist](https://gist.github.com/) with the contents of your pyproject.toml file**: https://github.com/bitcoin-core/HWI/blob/def565b5ac5d98b5fcc453328fc9a20207c6e00a/pyproject.toml

## Issue

When trying to install the dependencies with poetry for my project in a docker container, poetry just exits with no errors or warning, just an exit code of 1.

The Dockerfile is https://github.com/bitcoin-core/HWI/blob/def565b5ac5d98b5fcc453328fc9a20207c6e00a/contrib/build.Dockerfile and the script being run is https://github.com/bitcoin-core/HWI/blob/def565b5ac5d98b5fcc453328fc9a20207c6e00a/contrib/build_bin.sh. It specifically errors on the line `poetry install -E qt`. Changing this command to include `-vvv` does not effect the output.

The output is
```
+ poetry install -E qt
Creating virtualenv hwi-G-6EQZNy-py3.6 in /root/.cache/pypoetry/virtualenvs
Installing dependencies from lock file

Package operations: 28 installs, 0 updates, 0 removals
```

This previously worked when poetry 1.0.10 was used and it seems like 1.1.0 is the version where this stopped working.

If I run this on my host, also with 1.1.4, there are no issues. It seems to only be in this container. The script doesn't seem to make a difference; I can have a shell in the container, do `poetry install` manually, and it still fails in the same way.

brandon-leapyear:
I got the same thing, and bisecting my changes, I noticed that it happens when `LANG=en_US.UTF-8` is set, but the Docker image I'm using doesn't have that locale configured. I think this is a bug, `poetry` should either be independent of a misconfigured `LANG` setting, or show better logs

achow101:
I was able to work around this by adding

```
ENV LC_ALL=C.UTF-8
ENV LANG=C.UTF-8
ENV LANGUAGE=C.UTF-8
``` 

to my Dockerfile.

[...]

PR #4038, opened 2021-05-04
Fix for 3.6 with locale not set

brandonchinn178:
# Pull Request Check List

Resolves: https://github.com/python-poetry/poetry/issues/3412

I manually tested this by first installing the version of `poetry` on `master`, then running:
```bash
LANG= /path/to/poetry install
```
This should fail. Then making the change I did here and rerunning the above works.

Not sure what tests or docs should be updated, if any.

### DIFF
--- a/poetry/console/application.py
+++ b/poetry/console/application.py
@@ -21,6 +21,7 @@ from cleo.io.io import IO
 from cleo.io.outputs.output import Output
 
 from poetry.__version__ import __version__
+from poetry.core.utils._compat import PY37
 
 from .command_loader import CommandLoader
 from .commands.command import Command
@@ -142,6 +143,20 @@ class Application(BaseApplication):
         error_output: Optional[Output] = None,
     ) -> IO:
         io = super().create_io(input, output, error_output)
+
+        # Remove when support for Python 3.6 is removed
+        # https://github.com/python-poetry/poetry/issues/3412
+        if (
+            not PY37
+            and hasattr(io.output, "_stream")
+            and hasattr(io.output._stream, "buffer")
+            and io.output._stream.encoding != "utf-8"
+        ):
+            import io as io_
+
+            io.output._stream = io_.TextIOWrapper(
+                io.output._stream.buffer, encoding="utf-8"
+            )
 
         # Set our own CLI styles
         formatter = io.output.formatter


abn:
@brandonchinn178 thank for your contribution. However, I do not think this fixes the issue. I was also unable to reproduce the issue with 1.1.6.

brandonchinn178:
You were unable to reproduce running the above command using poetry installed with python 3.6?

abn:
Never mind, was using the wrong base container. I can reproduce the error as shown below.

```sh
podman run --rm -i --entrypoint bash python:3.6 <<EOF
set -xe
python -m pip install -q poetry
unset LANG
poetry --version
poetry new foobar
pushd foobar
poetry add pycowsay
EOF
```

```console
Package operations: 12 installs, 0 updates, 0 removals


  UnicodeEncodeError

  'ascii' codec can't encode character '\u2022' in position 2: ordinal not in range(128)

  at /usr/local/lib/python3.6/site-packages/clikit/io/output_stream/stream_output_stream.py:24 in write

  UnicodeEncodeError

  'ascii' codec can't encode character '\u2022' in position 2: ordinal not in range(128)

  at /usr/local/lib/python3.6/site-packages/clikit/io/output_stream/stream_output_stream.py:24 in write

Failed to add packages, reverting the pyproject.toml file to its original content.
```

abn:
I can also verify that the error does not occur with this PR. However the tests seem to not like it.

```sh
podman run --rm -i --entrypoint bash python:3.6 <<EOF
set -xe
python -m pip install -q git+https://github.com/python-poetry/poetry.git@refs/pull/4038/head
unset LANG
poetry --version
poetry new foobar
pushd foobar
poetry add pycowsay
EOF
```

poetry/console/application.py
```diff
@@ -137,6 +137,21 @@ def create_io(
...
+        if (
+            sys.version_info[:2] <= (3, 6) and
+            hasattr(io.output, "_stream") and
+            hasattr(io.output._stream, "buffer")
+        ):
+            import io as io_
+            io.output._stream = io_.TextIOWrapper(
+                io.output._stream.buffer, encoding='utf-8'
+            )
+        # fmt: on
```
abn, line 153: I would rather let black do its thing than introduce an exception. Adding 2 lines to remove 2 does not seem like a good gain.

Also note there is a `PY36` constant in `poetry.utils._compat`.

abn (dismissed):

poetry/console/application.py
brandonchinn178: The point wasn't to reduce line count, the point was to avoid making it seem like `import sys` was unrelated to the `if` block. But sure, I can do that.

The `PY36` constant checks if it's at least 3.6, so that won't work. But I can do `not PY37`. Also, it's `poetry.core.utils._compat`

brandonchinn178:
@abn any progress on this?

neersighted:
Looks like tests are still failing -- any interest in continuing work on this and getting it merged?

brandonchinn178:
The tests for 3.7+ were broken here: https://github.com/python-poetry/poetry/commit/b6e24401cff4463a37ee3de0924ac8d45fe11903. Do I need to fix those?

I'll look at the 3.6 failure

neersighted:
> 
> 
> The tests for 3.7+ were broken here: [b6e2440](https://github.com/python-poetry/poetry/commit/b6e24401cff4463a37ee3de0924ac8d45fe11903). Do I need to fix those?
> 
> I'll look at the 3.6 failure

`master` is fixed -- go ahead and rebase and it will solve the extraneous test failures.

brandonchinn178:
Done

neersighted (approved):

PR #4039, opened 2021-05-05
get-poetry.py: Make sure poetry is executable by all users.

parched:
Previously only the owner had executable permissions on unix, now everyone does which is useful when installing as root.

I realised after I made this change that get-poetry.py is now deprecated in favour of install-poetry.py so feel free to reject this if you don't think it's needed.

### DIFF
--- a/get-poetry.py
+++ b/get-poetry.py
@@ -688,17 +688,13 @@ class Installer:
                     )
                 )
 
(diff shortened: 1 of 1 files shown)


neersighted:
Closing this as `get-poetry.py` is deprecated and frozen. Please test `install-poetry.py`/`install.python-poetry.org` and create issues for any deficiencies you find.

PR #4047, opened 2021-05-07
install: rename --dev-only to --only-dev

abn:
This aligns option to existing options like --no-dev and --no-root.

Relates-to: #2917


abn:
This is now superceded by the deprecation of `--dev-only` due to the dependency group changes.

PR #4063, opened 2021-05-13
install poetry wrapper function in bashrc.d

meeuw:
# Pull Request Check List

~Resolves:~

- ~Added **tests** for changed code.~
- ~Updated **documentation** for changed code.~

I noticed that `install-poetry.py` doesn't install poetry in `$PATH` anymore (as `get-poetry.py` did). I'm not sure why this has been changed and I hope it doesn't block this PR.

I don't like that `poetry shell` enters commands in my bash history (`pexpect.sendline`) and that it starts a new shell (lacking the latest shell history). This (sourced) wrapper function fixes that as well.

`~/.bashrc.d/` is available on Fedora 34, I'm not sure since what release this has been added and if it's available on other (Linux) distributions.

I haven't checked for compatibility with other shells (zsh/fish etc.), if you like this change I'm willing to do check/fix it.

### DIFF
--- a/install-poetry.py
+++ b/install-poetry.py
@@ -22,6 +22,7 @@ import site
 import subprocess
 import sys
 import tempfile
+import textwrap
 
 from contextlib import closing
 from contextlib import contextmanager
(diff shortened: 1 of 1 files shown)


neersighted:
This is pretty much out-of-scope for `install-poetry.py` -- one of the big changes from `get-poetry.py` is that it does not modify the user's environment.

If you'd like to implement a wrapper like this, I'd suggest you create a plugin for your shell (e.g. oh-my-zsh) or add it directly to your shell config.

A distant third option would be shell integration ala `rbenv init` -- but I don't know that there is a lot of will/want to implement something like that.

Sorry about that -- and feel free to reach out if you want any advice on integrating Poetry further into your shell environment.

meeuw:
That makes sence, thanks @neersighted 

I was already afraid of such an answer so I created this:

https://github.com/meeuw/poetry-wrapper

PR #4088, opened 2021-05-22
Fedora installation method added

Kadermiyanyedi:
# Pull Request Check List

Resolves: #issue-number-here

### DIFF
--- a/docs/docs/index.md
+++ b/docs/docs/index.md
@@ -24,6 +24,14 @@ from the rest of your system.
 ```bash
 curl -sSL https://raw.githubusercontent.com/python-poetry/poetry/master/install-poetry.py | python -
 ```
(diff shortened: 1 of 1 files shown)


Secrus:
DNF installer is not maintained by Poetry team and can not by suggested in official docs as it would imply that the team is responsible for changes and fixes of that installer. If you are willing to, you can change it to a suggestion note for users to look for installers in their package repositories, those however, are not maintained by the team (and will not be for foreseeable future).

neersighted:
Closing this as out of scope as we don't officially support Distro packages -- like @Secrus said you're welcome to PR a mention of them, but it needs to be made clear that they are unofficial and may contain untested changes like different dependency versions from upstream.

Issue #2888, opened 2020-09-06
Activated-virtualenv-detection (including in tox) fails if there's also an active, poetry-managed, out-of-project env


mikenerone:
- [x] I am on the [latest](https://github.com/python-poetry/poetry/releases/latest) Poetry version.
- [x] I have searched the [issues](https://github.com/python-poetry/poetry/issues) of this repo and believe that this is not a duplicate.

- **OS version and name**: MacOS 10.15.6
- **Poetry version**: 1.1.0b2
- **Link of a [Gist](https://gist.github.com/) with the contents of your pyproject.toml file**: **n/a**

## Issue
Detection of an active virtualenv doesn't work if there's a poetry-managed env "active" (i.e. listed as "Activated" in `poetry env list). I've added a few blank lines for readability below.
```
-----------> poetry env list
-----------> poetry env info -p
-----------> source ~/.pyenv/versions/external_venv/bin/activate    # Activate an external venv

(external_venv) -----------> poetry env info -p
/Users/mikenerone/.pyenv/versions/3.8.5/envs/external_venv  # So far, so good

(external_venv) -----------> poetry env use 3.8    # Create a poetry-managed venv
Creating virtualenv superapp-HVMyRI9d-py3.8 in /Users/mikenerone/Library/Caches/pypoetry/virtualenvs
Using virtualenv: /Users/mikenerone/Library/Caches/pypoetry/virtualenvs/superapp-HVMyRI9d-py3.8  # First sign of trouble

(external_venv) -----------> poetry env info -p
/Users/mikenerone/Library/Caches/pypoetry/virtualenvs/superapp-HVMyRI9d-py3.8  # BAD: My activated venv is no longer detected

(external_venv) -----------> poetry env remove 3.8    # Remove the poetry-managed venv
Deleted virtualenv: /Users/mikenerone/Library/Caches/pypoetry/virtualenvs/superapp-HVMyRI9d-py3.8

(external_venv) -----------> poetry env info -p
/Users/mikenerone/.pyenv/versions/3.8.5/envs/external_venv  # Removing the poetry-managed venv restored the active-venv detection
```
[...]

PR #4098, opened 2021-05-24
Fix and clarify EnvManager.get()

mikenerone:
An activated virtualenv should always be honored, and an existing project virtualenv should always be selected based on the virtualenvs.in-project config setting (or even if not-existing if create=True).

Resolves: #2888

Note: No test/doc updates yet


Secrus:
@mikenerone are you still interested in continuing work on this?

Secrus:
Closing. No updates for over a year

Issue #4093, opened 2021-05-23
install-poetry in docker 


peter-dimov:
- [x] I am on the [latest](https://github.com/python-poetry/poetry/releases/latest) Poetry version.
- [x] I have searched the [issues](https://github.com/python-poetry/poetry/issues) of this repo and believe that this is not a duplicate.
- [ x] If an exception occurs when executing a command, I executed it again in debug mode (`-vvv` option).

- **OS version and name**:  Debian GNU/Linux 10 (buster)
- **Poetry version**:  1.1.6

## Issue

Installing poetry in docker using the recommended method does not work as a layer in the dockerfile but works fine when -it'ing into the container and running it.

`RUN curl -sSL https://raw.githubusercontent.com/python-poetry/poetry/master/install-poetry.py | python -`

produces the expected output:

`(...)
Poetry (1.1.6) is installed now. Great!

You can test that everything is set up by executing:

'poetry --version'`

but when jumping into the directory (home/.local/share/pypoetry/venv/bin), these are the available executables:

- activate  
- activate.csh  
- activate.fish  
- activate.ps1  
- activate.xsh  
- activate_this.py  
- pip  
- pip-3.7  
- pip3  
- pip3.7  
- python  
- python3  
- python3.7  
- wheel  
- wheel-3.7  
- wheel3  
- wheel3.7

In contrast, when omitting the line in the dockerfile, running the container interactively and curl'ing install-poetry, the installation works fine.

In addition, it seems that pip install virtualenv needs to be done manually as well or the layer fails to build with:

`ModuleNotFoundError: No module named 'virtualenv'`

Here is a barebones replication:

Dockerfile:
`FROM apache/airflow:latest

USER root

RUN apt-get update && apt-get install -y make build-essential libssl-dev zlib1g-dev \
  libbz2-dev libreadline-dev libsqlite3-dev wget curl llvm libncurses5-dev \
  libncursesw5-dev xz-utils tk-dev libffi-dev liblzma-dev python-openssl git python

RUN pip install virtualenv

RUN curl -sSL https://raw.githubusercontent.com/python-poetry/poetry/master/install-poetry.py | python -
RUN poetry -V`

abn:
The issue here is caused by the following line in the base image.

https://github.com/apache/airflow/blob/74821c8d999fad129b905a8698df7c897941e069/Dockerfile#L375

As a workaround you can update your Dockerfile like so.
```Dockerfile
FROM apache/airflow:latest

USER root

[...]

Issue #4089, opened 2021-05-23
install-poetry fails with PIP_REQUIRE_VIRTUALENV


neersighted:
Because `install-poetry.py` does not check the exit status of `pip` when bootstrapping the temporary `virtualenv` module, users are confronted with an odd error when they have `$PIP_REQUIRE_VIRTUALENV` set in their environment.

Output looks something like the following:

```
Installing Poetry...
Retrieving Poetry metadata

# Welcome to Poetry!

This will download and install the latest version of Poetry,
a dependency and package manager for Python.

It will add the `poetry` command to Poetry's bin directory, located at:

/home/neersighted/.local/share/pypoetry/bin

You can uninstall at any time by executing this script with the --uninstall option,
and these changes will be reverted.

Installing Poetry (1.1.6): Creating environment
Traceback (most recent call last):
  File "<stdin>", line 808, in <module>
  File "<stdin>", line 804, in main
  File "<stdin>", line 417, in run
  File "<stdin>", line 439, in install
  File "<stdin>", line 505, in make_env
AttributeError: module 'virtualenv' has no attribute 'cli_run'
```

This can easily be corrected by invoking the script again with `PIP_REQUIRE_VIRTUALENV=` to 'reset' the environmental variable.

For some background:
Pip provides this opt-in so that users can ensure they don't accidentally clutter or break their system installation of Python. Normally when this feature blocks execution of a command, it is called out to the user like this:

```
ERROR: Could not find an activated virtualenv (required).
```

Because pip is being invoked by `install-poetry.py` without checking exit status or output, this information is lost.

I would suggest explicitly resetting `PIP_REQUIRE_VIRTUALENV` in the installer script unconditionally to ensure that this interaction doesn't cause the failure above.

neersighted:
The installer changes proposed in #4099 greatly simplify the virtualenv-related code and solve half this issue (unexplained errors/discarded information). Now we just have to determine if the installer breaking due to this setting is a bug, or if the user is responsible for unsetting it before running the installer.

Thoughts, @sdispater?

neersighted:
I was wrong about this -- #4099 with actually entirely solve this issue as Poetry will no longer invoke `pip` outside a virtualenv! Nice!

PR #4099, opened 2021-05-24
installer: use venv and improve error handling

abn:
This change drops the dependency on `virtualenv` for the installer,
replacing environment creation with the built-in `venv` module available
in Python >= 3.2.

In addition, this change also improves error handling for subprocess
commands.

The following behavioural changes are also introduced.

- An error log is written to a secure temporary file with stdout and
  tracebacks on error.
- If an existing virtual environment exists prior to installation, this
  is saved, and used for recovery if installation fails.
  
Resolves: #4093
Resolves: #4089

### DIFF
--- a/install-poetry.py
+++ b/install-poetry.py
@@ -10,7 +10,8 @@ It does, in order:
       - In `${POETRY_HOME}` if it's set.
   - Installs the latest or given version of Poetry inside this virtual environment.
   - Installs a `poetry` script in the Python user directory (or `${POETRY_HOME/bin}` if `POETRY_HOME` is set).
-  - On failure, the error log is written to poetry-installer-error-*.log.
+  - On failure, the error log is written to poetry-installer-error-*.log and any previously existing environment
+    is restored.
 """
 
 import argparse
@@ -520,21 +521,13 @@ class Installer:
             )
         )
 
-        env = self.make_env(version)
-        self.install_poetry(version, env)
-        self.make_bin(version, env)
+        with self.make_env(version) as env:
+            self.install_poetry(version, env)
+            self.make_bin(version, env)
+            self._data_dir.joinpath("VERSION").write_text(version)
+            self._install_comment(version, "Done")
 
-        self._overwrite(
-            "Installing {} ({}): {}".format(
-                colorize("info", "Poetry"),
-                colorize("b", version),
-                colorize("success", "Done"),
-            )
-        )
-
-        self._data_dir.joinpath("VERSION").write_text(version)
-
-        return 0
+            return 0
 
     def uninstall(self) -> int:
         if not self._data_dir.exists():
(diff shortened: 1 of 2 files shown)


abn:
cc: @sdispater

sdispater:
@abn Won't that be an issue on Debian-based system which, as far as I recall, do not have the `python-venv` package installed?

abn:
I thought this was now a built-in on there as well. Will check. We can do a fall back. But think now there is also install-virtualenv.

neersighted:
> 
> 
> I thought this was now a built-in on there as well. Will check. We can do a fall back. But think now there is also install-virtualenv.

Pretty sure this is a non-issue. The `python3.7-venv` package just contains the `pyvenv-3.7` wrapper -- it looks like the `venv` module belongs to the `libpython3.7-stdlib` package: https://packages.debian.org/buster/amd64/libpython3.7-stdlib/filelist

install-poetry.py
```diff
@@ -277,6 +262,64 @@ def temporary_directory(*args, **kwargs):
...
+        builder = venv.EnvBuilder(clear=True, with_pip=True, symlinks=False)
+        builder.ensure_directories(target)
+        builder.create(target)
+
+        env = cls(target)
+
+        # we do this here to ensure that outdated system default pip does not trigger older bugs
+        env.pip("install", "--upgrade", "pip")
+
+        return cls(target)
```
neersighted, line 296: Is there a reason to not return `env`? Seems like the class is just a simple holder for the `target`/`path` so there's no reason not to re-use it.

install-poetry.py
```diff
@@ -277,6 +262,64 @@ def temporary_directory(*args, **kwargs):
...
+        self._python = self._path.joinpath(
+            "Scripts/python.exe" if WINDOWS else "bin/python"
+        )
+
+    @property
+    def path(self):
+        return self._path
+
+    @classmethod
+    def make(cls, target: Path) -> "VirtualEnvironment":
```
neersighted, line 284: Maybe naming this `path` (like the getter/attribute) would make this slightly more clear.

neersighted (dismissed):
Looks good to me! I have some minor (very minor) nitpicks, but this nicely solves the issue I was having and eliminates the complexity of shelling out to `virtualenv`!

neersighted (dismissed):
I've rebased this/fixed the merge conflicts and implemented the suggestions I made before (discussed offline with @abn) -- this is ready to merge assuming the checks pass.

Since #4666 was landed in master this will be subject to full integration test coverage.

neersighted (approved):
I've rebased this onto master, and squashed my contributions. We're now testing green across the board, with only one weird hack/workaround (can't call Subprocess with a `Path` on Windows).

For posterity, we were struggling with macOS + Python 3.10 -- it ended up being an issue with blanking the environment. Omitting that eliminates a whole class of weirdness -- we can always revisit if needed.

neersighted:
> 
> 
> @abn Won't that be an issue on Debian-based system which, as far as I recall, do not have the `python-venv` package installed?

It turns out I was wrong and we do end up with an error -- however, since Ubuntu/Debian monkey-patch Python, it's quite readable and is not a blocker:

```
Retrieving Poetry metadata

# Welcome to Poetry!

This will download and install the latest version of Poetry,
a dependency and package manager for Python.

/home/ubuntu/.local/bin

You can uninstall at any time by executing this script with the --uninstall option,
and these changes will be reverted.

Installing Poetry (1.1.11): Creating environment
The virtual environment was not created successfully because ensurepip is not
available.  On Debian/Ubuntu systems, you need to install the python3-venv
package using the following command.

    apt install python3.8-venv

You may need to use sudo with that command.  After installing the python3-venv
package, recreate your virtual environment.

Failing command: ['/home/ubuntu/.local/share/pypoetry/venv/bin/python3', '-Im', 'ensurepip', '--upgrade', '--default-pip']
```

abn:
I have a wip fix for that coming as soon as I fix merge conflicts from master. 

We'll be falling back to a virtualenv zipapp.

neersighted (changes requested):
Looks good -- the best fusion of both our changes. I'd like to fix that redundant `return cls(target)` and merge this!

neersighted (approved):

PR #4086, opened 2021-05-22
Added pip style keyring password lookup as a fallback.

Darsstar:
This way Poetry gives keyring backends like Microsoft's `artifacts-keyring` or Google's `keyrings.google-artifactregistry-auth` a chance to retrieve the credentials. Since Microsoft's Azure DevOps Personal Access Tokens will expire after some time the `artifacts-keyring` implementation will generate a new one for you to counteract that.

By "pip style" I mean  keyring requests based on the repository url instead of poetry config paths.

Tested by modifing installation/authenticator.py on my Poetry 1.2.0a1 installation.

# Pull Request Check List

Resolves: python-poetry/poetry#3344

- [x] Added **tests** for changed code.
- [x] Updated **documentation** for changed code.

PR #4120, opened 2021-05-31
Adding OAuth-style authentication using get_credential

caseydialpad:
# Pull Request Check List

This PR is to extend support for keyrings by using the `get_credential` functionality of `keyring` to allow OAuth-style authentication. If `username` AND `password` is absent, the keyring will be queried for OAuth-style credentials.

This change has been tested with Google Cloud's Artifact Registry. This change should also work for other private PyPI providers who use OAuth and a custom keyring backend to authenticate users.

Resolves: No existing ticket number.


caseydialpad:
It looks like there is a similar PR at https://github.com/python-poetry/poetry/pull/4086. I'll leave this open for now - it may be valuable to compare two different techniques although the #4086 seems cover a wider array of authentication techniques using keyrings.

caseydialpad:
Closing due to existence of superior solution at PR mentioned above.

PR #4122, opened 2021-05-31
Use `safe_extra` to standardize extra dependencies

deepyaman:
# Pull Request Check List

Resolves: #4123

### DIFF
--- a/src/poetry/utils/helpers.py
+++ b/src/poetry/utils/helpers.py
@@ -160,7 +160,7 @@ def pluralize(count: int, word: str = "") -> str:
 
 
 def safe_extra(extra: str) -> str:
-    """Convert an arbitrary string to a standard 'extra' name
+    """Convert an arbitrary string to a standard 'extra' name.
 
     Any runs of non-alphanumeric characters are replaced with a single '_',
     and the result is always lowercased.
--- a/src/poetry/utils/helpers.py
+++ b/src/poetry/utils/helpers.py
@@ -157,3 +157,15 @@ def pluralize(count: int, word: str = "") -> str:
     if count == 1:
         return word
     return word + "s"
+
+
+def safe_extra(extra: str) -> str:
+    """Convert an arbitrary string to a standard 'extra' name
+
+    Any runs of non-alphanumeric characters are replaced with a single '_',
+    and the result is always lowercased.
+
+    See
+    https://github.com/pypa/setuptools/blob/452e13c/pkg_resources/__init__.py#L1423-L1431.
+    """
+    return re.sub("[^A-Za-z0-9.-]+", "_", extra).lower()
--- a/src/poetry/puzzle/provider.py
+++ b/src/poetry/puzzle/provider.py
@@ -36,6 +36,7 @@ from poetry.packages.package_collection import PackageCollection
 from poetry.puzzle.exceptions import OverrideNeeded
 from poetry.repositories.exceptions import PackageNotFound
 from poetry.utils.helpers import download_file
+from poetry.utils.helpers import safe_extra
 from poetry.vcs.git import Git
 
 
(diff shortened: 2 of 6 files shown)


deepyaman:
I can't quite figure out why the two running jobs are failing; the tests pass on my machine (aside from `venv`-related ones that didn't pass even when I first cloned the project), and the timeouts seem to happen before tests run.

finswimmer (changes requested):
Thanks a lot for your contribution @deepyaman :+1: 

I've talked about it with @sdispater and he's fine with being not case sensitiv here. But instead of using `pkg_resources.safe_extra` it would be better to implement our own helper method. Thus we can avoid a dependency that isn't absolutely necessary. The implementation seems to be trivial: https://github.com/pypa/pkg_resources/blob/6f81a44010d1266494025647dd1e1f0befa5b26b/pkg_resources/__init__.py#L1435

fin swimmer

deepyaman:
> But instead of using `pkg_resources.safe_extra` it would be better to implement our own helper method.

@finswimmer Sorry for the delay, but this is done!

neersighted (commented):
This looks good to me -- can we rebase this so it can be merged?

deepyaman:
> Looks like we still need to get CI working -- also there's a lot of junk commits. Could we squash this PR down to 1-3 (logical/meaningful) commits?

Sorry, I had to step out, but I was in the middle of setting up my local env to resolve the issues. It's been a year, so a lot has changed, including my laptop. :)

I'll ping you once I've got them resolved, and I can also squash down to a change or two (I wrongly assumed or forgot that Poetry doesn't just squash-merge everything).

deepyaman:
@neersighted Think we're good now! I'll just update with rebase since there's been more pushes, but I don't imagine there will be any issues.

neersighted:
> > Looks like we still need to get CI working -- also there's a lot of junk commits. Could we squash this PR down to 1-3 (logical/meaningful) commits?
> 
> As an unrelated note on this, looking at the (recent) commit history, there's a mix of Conventional Commit and old-school sentence-case commits. Maybe it makes sense to standardize on one?

We're not picky yet but eventually I'll get around to gating commits on conventional commit compliance :smile:

And we do squash merge when needed, but during a more involved change (with a developer that knows how to rebase) like this I prefer to have logical commits to make bisecting easier -- e.g. in this case I'd prefer add tests then add code to make the change.

deepyaman:
> I'd prefer add tests then add code to make the change.

@neersighted Do you want me to do this, or is this good to go as is? I've never broken commits down like that (because generally try to ensure that tests pass on each commit on the main branch on projects I've worked on), but I can update it if that's the standard for Poetry and what we need to close this out.

neersighted:
> > I'd prefer add tests then add code to make the change.
> 
> @neersighted Do you want me to do this, or is this good to go as is? I've never broken commits down like that (because generally try to ensure that tests pass on each commit on the main branch on projects I've worked on), but I can update it if that's the standard for Poetry and what we need to close this out.

It's nothing enforced -- merely a preference for review, and given we're already reviewed this change one commit is fine :smile:

src/poetry/utils/helpers.py
```diff
@@ -157,3 +157,15 @@ def pluralize(count: int, word: str = "") -> str:
     if count == 1:
         return word
     return word + "s"
+
+
+def safe_extra(extra: str) -> str:
+    """Convert an arbitrary string to a standard 'extra' name
```
neersighted, line 163:
```suggestion
    """Convert an arbitrary string to a standard 'extra' name.
```

neersighted (approved):

deepyaman:
@neersighted Thank you so much for your help in closing this out!

PR #4179, opened 2021-06-14
add flag `--use-env` to poetry commands

finswimmer:
The PR adds a new global parameter `--use-env`. It adds the ability to tell poetry which python executable it should use in the same way `poetry env use` can do it.

The benefits of the parameter are:
 * running commands in several environments without switching the default python executable first
 * if in-project venvs are used, there is no need to add an entry to the `envs.toml`

I haven't add any tests until now. Recommendations are welcome :)

# Pull Request Check List

Resolves: #issue-number-here

### DIFF
--- a/docs/cli.md
+++ b/docs/cli.md
@@ -27,6 +27,8 @@ then `--help` combined with any of those can give you more information.
 * `--no-ansi`: Disable ANSI output.
 * `--version (-V)`: Display this application version.
 * `--no-interaction (-n)`: Do not ask any interactive question.
+* `--no-plugins`       Disables plugins.
+* `--use-env=USE-ENV`  The python executable to use.
 
 
 ## new
--- a/docs/managing-environments.md
+++ b/docs/managing-environments.md
@@ -18,8 +18,8 @@ To achieve this, it will first check if it's currently running inside a virtual 
 If it is, it will use it directly without creating a new one. But if it's not, it will use
 one that it has already created or create a brand new one for you.
 
-By default, Poetry will try to use the currently activated Python version
-to create the virtual environment for the current project.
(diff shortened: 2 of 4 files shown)


felipou:
I was about to create a PR with a similar idea that I started coding yesterday, but found this PR when looking for more references now.

My idea was to create a new config `virtualenvs.default-executable`, which could be a path for a python executable, but with two special values: "system" (the default, keeping the same behavior as today, that is, using `sys.executable`) and `shell` which would use `shutil.which("python")` to get the currently active Python version in the shell. I have no idea if that would be portable, I only use poetry on Linux.

I even wrote some simple code for that, but as I was testing and reviewing, I noticed that the version check code was only for the `sys.executable` python, so this would require modifying the existing code a lot more (which seemed very risk to me, since it was the first time I was looking at this project's code).

But anyway, I think this would be a lot better (for me at least), since I would only need to change the configuration globally once, and poetry would work as needed for all my project when using pyenv.

finswimmer:
@felipou: I have another proof-of-concept here: https://github.com/python-poetry/poetry/pull/4852 Would be happy if you could test it.

felipou:
> @felipou: I have another proof-of-concept here: #4852 Would be happy if you could test it.

Interesting, very similar idea. Just by looking at the code, it seems to be working. I'll see if I can test it sometime soon.

I liked how you implemented it without modifying the existing code much, and also how you picked the shell python executable by executing a command instead of using `shutil.which` as I had suggested. Seems to me like your approach would probably be more portable, but it's  just a guess.

finswimmer:
Closing this in favor of https://github.com/python-poetry/poetry/pull/4852.

PR #4191, opened 2021-06-18
Update basic-usage.md

martinlarsalbert:
I'm completely new to poetry and didn't understand where the virtual environments are installed. But I found the following command very helpfull:

```bash
$ poetry show -v
```

So here is a suggestion to add this as a hint in this documentation.

# Pull Request Check List

Resolves: #issue-number-here

### DIFF
--- a/docs/basic-usage.md
+++ b/docs/basic-usage.md
@@ -88,6 +88,12 @@ It will automatically find a suitable version constraint **and install** the pac
 ## Using your virtual environment
 
 By default, poetry creates a virtual environment in `{cache-dir}/virtualenvs` (`{cache-dir}\virtualenvs` on Windows).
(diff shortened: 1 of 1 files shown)


finswimmer:
Hello @martinlarsalbert,

thanks a lot for your contribution and the suggestion to improve the docs. :+1: 

However, your suggestion is not correct. `poetry show` is intended to show information about an installed package. `poetry show -v` increases the verbosity and shows some debug messages like the venv used.

To see where the venv is installed you should use `poetry env info`. This is already described [here](https://python-poetry.org/docs/managing-environments/#displaying-the-environment-information).

fin swimmer

martinlarsalbert:
Thanks a lot!

Issue #4085, opened 2021-05-22
poetry 1.2.0a1: AttributeError when installing package from secondary repo


maksbotan:
- [ ] I am on the [latest](https://github.com/python-poetry/poetry/releases/latest) Poetry version.
- [x] I have searched the [issues](https://github.com/python-poetry/poetry/issues) of this repo and believe that this is not a duplicate.
- [x] If an exception occurs when executing a command, I executed it again in debug mode (`-vvv` option).

- **OS version and name**: macOS 10.15
- **Poetry version**: 1.2.0a1
- **Link of a [Gist](https://gist.github.com/) with the contents of your pyproject.toml file**: 

## Issue

Hi! I'm testing new 1.2.0a1 version as an early-adopter to find potential issues.

My `pyproject.toml` defines a secondary repo:

```toml
[[tool.poetry.source]]
name = "XXX"
url = "https://pypi.XXX.YYY"
secondary = true
```

And I have one dependency which comes from this private pypi repo. Poetry 1.1.3 installed it without errors, but on 1.2.0a1 I get this stacktrace:

```
  • Installing FOOBAR (1.7.2): Pending...
  • Installing FOOBAR (1.7.2): Failed

  Stack trace:

  5  ~/Library/Application Support/pypoetry/venv/lib/python3.8/site-packages/poetry/installation/executor.py:244 in _execute_operation
      242│ 
      243│             try:
    → 244│                 result = self._do_execute_operation(operation)
      245│             except EnvCommandError as e:
      246│                 if e.e.returncode == -2:

  4  ~/Library/Application Support/pypoetry/venv/lib/python3.8/site-packages/poetry/installation/executor.py:321 in _do_execute_operation
      319│             return 0
      320│ 
    → 321│         result = getattr(self, f"_execute_{method}")(operation)
      322│ 
      323│         if result != 0:

  3  ~/Library/Application Support/pypoetry/venv/lib/python3.8/site-packages/poetry/installation/executor.py:462 in _execute_install
      460│ 
      461│     def _execute_install(self, operation: Union[Install, Update]) -> int:
    → 462│         status_code = self._install(operation)
      463│ 
      464│         self._save_url_reference(operation)

  2  ~/Library/Application Support/pypoetry/venv/lib/python3.8/site-packages/poetry/installation/executor.py:498 in _install
      496│             archive = self._download_link(operation, Link(package.source_url))
      497│         else:
    → 498│             archive = self._download(operation)
      499│ 
[...]

PR #4212, opened 2021-06-24
Fix AttributeError on hash mismatch

paulmelnikow:
This replaces the AttributeError reported in #4085 with the proper exception.

# Pull Request Check List

Resolves: N/A (related to #4085, however the underlying error is not fixed)

### DIFF
--- a/poetry/installation/executor.py
+++ b/poetry/installation/executor.py
@@ -679,8 +679,9 @@ class Executor:
                 ).hash()
             )
             if archive_hash not in {f["hash"] for f in package.files}:
+                filename = archive.filename if isinstance(archive, Link) else archive
                 raise RuntimeError(
-                    f"Invalid hash for {package} using archive {archive.name}"
+                    f"Invalid hash for {package} using archive {filename}"
                 )
 
             self._hashes[package.name] = archive_hash


paulmelnikow:
I'm seeing these test failures locally too, though they do not seem to be related to the changes in this PR.

paulmelnikow:
@finswimmer Could I ask you to take a look at this?

If you think a test is needed to merge it, I'd appreciate some guidance on that. If there is some existing coverage of these code paths I'm happy to add some tests of the error case. Though if not I'm probably not quite ready to create tests in an area of the code where there's no test coverage now. In any case if you feel that's needed, would appreciate pointers on where to start! Thanks!

sdispater:
@paulmelnikow Thanks for the fix!

Could you rebase your PR on the the latest version of the `master` branch?

paulmelnikow:
@sdispater Thanks for reviewing! Done!

paulmelnikow:
I thought I had reduced complexity for Sonar but it looks like not.

I imagine the way to fix this is to break out a helper method from this one. Perhaps lines 655-671 could become a new method? If you agree, could you suggest what that method should be called?

paulmelnikow:
> I imagine the way to fix this is to break out a helper method from this one. Perhaps lines 655-671 could become a new method? If you agree, could you suggest what that method should be called?

@sdispater Could you offer any guidance on this? I took another look at how to refactor this to satisfy Sonar and it seems more like what's needed is not a second helper function, but to push some of this code complexity into a lower-level abstraction. But I'm happy to execute whatever refactor you suggest.

paulmelnikow:
@sdispater gentle nudge

paulmelnikow:
@sdispater I'd love to get this merged!

neersighted:
> @sdispater I'd love to get this merged!

Sorry about the delays here -- does this still reproduce on master?

neersighted:
Actually, it looks like #4531 fixed this and I already validated it does not reproduce. Please comment/re-open if that's not the case and we still need to get a fix in.

PR #4213, opened 2021-06-25
Wheel installer and builder

sdispater:
This PR is the first step towards getting rid of subprocess calls to `pip` for package installations and removals.

The wheel building is handled by https://github.com/pypa/build and the wheel installation is handled by https://github.com/pradyunsg/installer (with modifications to be more lenient).

Note that the wheel building is **not** parallelized due to race conditions that I am yet to track down (it seems like the issue comes from `build` when calling the backend hooks). This is not a blocker though.

Overall this speeds up quite a bit the installation times, especially once wheels are cached.

## Pull Request Check List


pradyunsg:
> Note that the wheel building is **not** parallelized due to race conditions that I am yet to track down

These come from setuptools reusing the same build directory for performing builds and not having any concurency protections. These race conditions exist in newer versions of pip as well, when building in parallel, since pip has switched to in-tree builds as well.

neersighted:
Closing since this is superseded.

Issue #3520, opened 2020-12-31
Natively support `py` launcher on Windows


TBBle:
- [x] I have searched the [issues](https://github.com/python-poetry/poetry/issues) of this repo and believe that this is not a duplicate.
- [x] I have searched the [documentation](https://python-poetry.org/docs/) and believe that my question is not covered.

## Feature Request

`poetry env use 3.9` is really nice, but assumes a binary named `python3.9` is in the path.

On Windows, the Python installer does not install such version-named binaries, and so to avoid having multiple `python.exe` on the path, the `py` launcher is included, which is used by `py -<version>` exactly as Poetry is trying to do.

It would be nice if Poetry could correctly use this. I couldn't find a way to do this easily, as
```
poetry env use 'py -3.9'
```
tries to execute a binary named "py -3.9" (I suspect this is an intended regression due to fixing #1747), and without the quotes, it tries to take `-3.9` as an option for itself, and fails.

As a workaround, I am, under PowerShell, doing this:
```
poetry env use $(py -3.8 -c 'import sys; print(sys.executable)')
```
which works fine.

Note that this isn't a problem for simply `poetry install` workflows, as it deafults to the Python with which Poetry is associated, and Poetry already knows the location of this binary. However, I hit this with a package that does not support Python 3.9, while I have Poetry installed under Python 3.9.

This problem also appears when trying to clean up venvs, as `poetry env remove` does not support "default Python" the way `poetry install` does.

`poetry destroy` was rejected in #926, so you can only clean up Poetry env by naming either the env's full name or the path to the Python binary; the latter requires the same workaround I'm using above.

radoering:
I wonder if it might be better to use a lib that implements [PEP-514](https://peps.python.org/pep-0514/) (e.g. [pythonfinder](https://pypi.org/project/pythonfinder/)) to find executables instead of using an application like `py` launcher?

TBBle:
Yeah, that would make sense too, since [`py.exe` has been rewritten in Python 3.11 as a PEP-514-consuming tool](https://github.com/python/cpython/issues/90724), so the results should be consistent.

eyadmba:
Any workaround for now?
Is there a way to switch the default python version for a session for example?

absassi:
@eyadmba the workaround is to specify the full path to the `python.exe`. It can be either the path to the one in the Python installation folder or the one in the Poetry environment, it doesn't matter. You can either use the following command, as indicated by @TBBle:

[...]

PR #4217, opened 2021-06-26
Support Python Launcher on Windows when using `poetry env use` and `poetry env remove` (#3520)

sitd0813:
# Pull Request Check List

Resolves: #issue-number-here

Fix <https://github.com/python-poetry/poetry/issues/3520>

~~I did not add any test because I'm not really familiar with PyTest and thus had no idea how to add Windows specific test. However, the code was built and tested on Windows 10 device and it proved to resolve <https://github.com/python-poetry/poetry/issues/3520> without apparent errors.~~

EDIT

Added test to verify if `EnvManager._execute_python_script` works as expected.


sitd0813:
Both tests currently failed due to `ConnectionError`. I have no idea why it occurred. It seems like there is an issue with the testing platform.

poetry/utils/env.py
```diff
@@ -448,22 +461,14 @@ def activate(self, python: str, io: IO) -> "Env":
...
-                            "-c",
-                            "\"import sys; print('.'.join([str(s) for s in sys.version_info[:3]]))\"",
-                        ]
-                    ),
-                    shell=True,
-                )
-            )
+            python_version = execute_python_script(
+                python,
+                "\"import sys; print('.'.join([str(s) for s in sys.version_info[:3]]))\"",
```
Leonardofreua, line 466: What do you think about creating a constant for the script `"\"import sys; print('.'.join([str(s) for s in sys.version_info[:3]]))\""`?

sitd0813: Thank you for reviewing my code. I added the constant for the script. Also, I made the `execute_python_script` a member function of `EnvManager` as it is only used inside it. It is now `EnvManager._execute_python_script`.

Leonardofreua: Great. Good catch.

sitd0813: Is there anything else I need to do? It seems that learning more test require approval.

Leonardofreua: I think it needs more tests and one more reviewer.

dalito:
As a fan of the launcher I have tested your PR with poetry-1.2.0a2 on Win10. It mostly works. 👍  

Only when removing an env an error occurs. The env is removed nevertheless. The error is probably due to a "feature" in Win10. When you don't have python on path but enter "python", Win10 suggests to install Python from the Microsoft store. So testing for "python" gives an unexpected output. Maybe it is better to test for the presence of "pip"?

```
λ poetry env use python3.8
Creating virtualenv mared-NfQ5kqas-py3.8 in C:\Users\zzz\AppData\Local\pypoetry\Cache\virtualenvs
Using virtualenv: C:\Users\zzz\AppData\Local\pypoetry\Cache\virtualenvs\mared-NfQ5kqas-py3.8

λ poetry env remove python3.8

  CalledProcessError

  Command 'python -W ignore -' returned non-zero exit status 9009.

  at C:\dev\Python39\lib\subprocess.py:528 in run
       524│             # We don't call process.wait() as .__exit__ does that for us.
       525│             raise
       526│         retcode = process.poll()
       527│         if check and retcode:
    →  528│             raise CalledProcessError(retcode, process.args,
       529│                                      output=stdout, stderr=stderr)
       530│     return CompletedProcess(process.args, retcode, stdout, stderr)
       531│
       532│

  EnvCommandError

  Command python -W ignore - errored with the following return code 9009, and output:
  Python was not found but can be installed from the Microsoft Store: https://go.microsoft.com/fwlink?linkID=2082640input was : import sys

  if hasattr(sys, "real_prefix"):
      print(sys.real_prefix)
  elif hasattr(sys, "base_prefix"):
      print(sys.base_prefix)
  else:
      print(sys.prefix)


  at ~\_poetry\venv\lib\site-packages\poetry\utils\env.py:1285 in _run
      1281│                 output = subprocess.check_output(
      1282│                     cmd, stderr=subprocess.STDOUT, env=env, **kwargs
      1283│                 )
      1284│         except CalledProcessError as e:
    → 1285│             raise EnvCommandError(e, input=input_)
      1286│
      1287│         return decode(output)
      1288│
      1289│     def execute(self, bin: str, *args: str, **kwargs: Any) -> Optional[int]:
```

neersighted:
The current consensus is that we should use `pythonfinder` instead of trying to write out this logic ourselves. Feel free to re-open this PR if you'd like to switch to that approach -- otherwise, @bmarroquin has expressed interest in PRing `pythonfinder` support.

PR #3406, opened 2020-11-24
Add PyPI registry correctly to pool depending on other sources/Fix source.secondary = true being ignored

tomzx:
In the event where we defined sources that were set as
secondary = True, we would end up with PyPI being after this source
when it should have acted as default in that case.

The main issue stems from the fact that it's not because you have
sources configured that PyPI should not be a default. Instead, PyPI
should be default if there are no sources with secondary = False and
not default if there are sources with secondary = True.

Resolves #2339, resolves #3306, resolves #3238

Issue #3010, opened 2020-09-30
Potential thread safety issue under certain environments: Incorrectly nested style tag found


abn:
In certain environments, specifically CI environments, this issue occurs randomly. It would seem that when using the new executor for installation the `io` being used is not thread safe and `pastel` fails attempting to pop the wrong stack.

```console
  • Installing glob2 (0.6)

ValueError

Incorrectly nested style tag found.

  • Installing httpretty (0.9.7)
```

PetterS:
There are also other issues with the parallel installation: see #2658.

machielg:
I also have this problem when doing poetry (1.1.2) install in gitlab. The workaround is disable the new installer: `poetry config experimental.new-installer false`

PetterS:
Yeah, this should be relatively easy to fix. Just wrap a lot of calls to `self._io` with a lock in the executor. Alteratively, make `IO` from clikit safe to call from multiple threads.

vikigenius:
Yeah, I am also facing this issue on aws, when using the user-data script to install poetry packages

PetterS:
For automation or when reliability is needed, the new installer should definitely be disabled. 

We have a config file in the repo disabling it by default. If anyone wants to use it (e.g. when setting up a new env), they can do so manually.

Michal-Szelag-Sonos:
I hit this couple times now, also in CI.  If I keep running into this, I may disable new installer which is a bummer since it significantly speeds up the install in my project with 100+ dependencies.

sdispater:
@Michal-Szelag-Sonos Which Python version are you using when the error occurs?

Michal-Szelag-Sonos:
> @Michal-Szelag-Sonos Which Python version are you using when the error occurs?

I am seeing this on 2.7

PetterS:
Should happen on any Python version since `IO` is not thread-safe.

sdispater:
> > @Michal-Szelag-Sonos Which Python version are you using when the error occurs?
> 
> I am seeing this on 2.7

Yes, from experience I only have seen this on Python 2.7 and not on other Python versions, I am not sure why. I don't know if it's a coincidence or not though.

It may also be caused by the `futures` package used on Python 2.7.

Note that every write in the `Executor` class occurs after having acquired a lock so even though the `IO` is not thread safe, everything should work as expected.

Michal-Szelag-Sonos:
[...]

PR #4241, opened 2021-07-01
cherry-pick PR #3406 for v1.1

### LINKED ISSUE #2339: Poetry not respecting order of evaluation for repository sources
- [X] I am on the [latest](https://github.com/python-poetry/poetry/releases/latest) Poetry version.

- **OS version and name**: Mac OS 10.15.3
- **Poetry version**: 1.0.5

## Issue
So, I am trying to configure a pyproject.toml file that supports pulling files from pypi.org whenever possible (ie: because on my site the performance is better) but will support pulling packages that aren't found on pypi.org from a secondary location (ie: a private pypi repo). So to start with I added a section like what follows to a sample toml file for testing:

```
[[tool.poetry.source]]
name = 'default'
url = 'https://pypi.python.org/simple'
default = true

[[tool.poetry.source]]
name = "private"
url = "https://private/server/url
secondary = true

[tool.poetry.dependencies]
python = "^3.6"
sphinx = "*"
```

Based on the docs, I was under the impression that by putting `default=true` on the first repo config, and `secondary=true` on the other one, that my goal would be achieved: if a package exists on pypi.org it'd pull it from there and if not it'd pull it from my secondary repo. However, that does not seem to be the case.

For the sake of this discussion I am using average performance metrics from running `poetry lock` on this toml file and comparing the length of time it takes to resolve the dependencies of the one Python package mentioned (ie: `sphinx` which is a standard Python package available on pypi.org). 

So, running `poetry lock` on this file using the configuration I posted above takes about 20-25 seconds to complete. For comparison purposes I simply removed the second source definition from the toml file giving me:

```
[[tool.poetry.source]]
name = 'default'
url = 'https://pypi.python.org/simple'
default = true

[tool.poetry.dependencies]
python = "^3.6"
sphinx = "*"
```

Re-running the same common against this toml file takes between 2 and 3 seconds to complete - a full order of magnitude difference in performance. Now, for my 3rd and final test I removed the first repo source leaving just the second one, giving me the following configuration block:

```
[[tool.poetry.source]]
name = "private"
url = "https://private/server/url
secondary = true

[tool.poetry.dependencies]
python = "^3.6"
sphinx = "*"
```

Re-running the same command again takes about the same 20-25 seconds to complete (NOTE: our private pypi repository is also a mirror of the public pypi.org repo so I can compare the same package versions and resolution times in all cases).

So, based on these results, the only explanation I can see is that when the second repository definition is in place the public pypi.org repo is being ignored or something and the secondary / private repo is still being accessed for some reason.

My expectation here is that poetry would iterate over the various repos defined in the toml file in order of declaration. For each package listed in the toml file it should then try to access / index each package against each repository stopping when it finds the first match. If this were true then I would have expected my first test case above to perform identically to the second one because the package I'm testing (sphinx) should exist on pypi.org, and all of it's transitive dependencies will be available there as well. However this is obviously not the case.

So my question is - is this a bug or is there some additional configuration options I need to specify to get this to work as I was expecting?

### LINKED ISSUE #3238: Custom repository doesn't have precedence over PyPI

### LINKED ISSUE #3306: poetry 1.1.4 locks all packages to be pulled from secondary repo

- **OS version and name**: macOS 10.14
- **Poetry version**: 1.1.4

## Issue

Hi,

I'm experiencing this behavior on 1.1.4. In `pyproject.toml` I have configured a secondary repo:

```toml
[[tool.poetry.source]]
name = "xxx"
url = "https://PRIVATE"
secondary = true
```

As such, I expect that `poetry` will fetch most packages from the main PyPI. And 1.1.3 indeed does that. However, 1.1.4 adds this configuration to every package (non-private as well) in `poetry.lock`:

```toml
[[package]]
name = "alabaster"
version = "0.7.12"
description = "A configurable sidebar-enabled Sphinx theme"
category = "main"
optional = true
python-versions = "*"

[package.source]
type = "legacy"
url = "https://PRIVATE"
reference = "xxx"
```

IMO this is unexpected behavior. I did `git bisect` and found out that this bug was introduced in 74fc8eb1cafe2006c22a969d5c0cd53968e46f2a, in an attempt to fix other pypi-related issues.

jstriebel:
This cherry-picks #3406 for v1.1 for a new patch-release. From the original PR:

> Add PyPI registry correctly to pool depending on other sources (#3406)
> 
> In the event where we defined sources that were set as
> secondary = True, we would end up with PyPI being after this source
> when it should have acted as default in that case.
> 
> The main issue stems from the fact that it's not because you have
> sources configured that PyPI should not be a default. Instead, PyPI
> should be default if there are no sources with secondary = False and
> not default if there are sources with secondary = True.

# Pull Request Check List

Related to PR #3406, resolves #2339, resolves #3306, resolves #3238 for v1.1

### DIFF
--- a/poetry/factory.py
+++ b/poetry/factory.py
@@ -88,14 +88,15 @@ class Factory(BaseFactory):
 
             poetry.pool.add_repository(repository, is_default, secondary=is_secondary)
 
-        # Always put PyPI last to prefer private repositories
-        # but only if we have no other default source
-        if not poetry.pool.has_default():
-            has_sources = bool(sources)
-            poetry.pool.add_repository(PyPiRepository(), not has_sources, has_sources)
-        else:
+        # Put PyPI last to prefer private repositories
+        # unless we have no default source AND no primary sources
+        # (default = false, secondary = false)
+        if poetry.pool.has_default():
             if io.is_debug():
                 io.write_line("Deactivating the PyPI repository")
+        else:
+            default = not poetry.pool.has_primary_repositories()
+            poetry.pool.add_repository(PyPiRepository(), default, not default)
 
         return poetry
 
--- a/poetry/repositories/pool.py
+++ b/poetry/repositories/pool.py
@@ -22,6 +22,7 @@ class Pool(BaseRepository):
         self._lookup = {}  # type: Dict[str, int]
         self._repositories = []  # type: List[Repository]
         self._default = False
+        self._has_primary_repositories = False
         self._secondary_start_idx = None
 
         for repository in repositories:
@@ -37,6 +38,9 @@ class Pool(BaseRepository):
 
     def has_default(self):  # type: () -> bool
         return self._default
+
+    def has_primary_repositories(self):  # type: () -> bool
+        return self._has_primary_repositories
 
     def has_repository(self, name):  # type: (str) -> bool
         name = name.lower() if name is not None else None
@@ -81,6 +85,7 @@ class Pool(BaseRepository):
             self._repositories.append(repository)
             self._lookup[repository_name] = len(self._repositories) - 1
         else:
+            self._has_primary_repositories = True
             if self._secondary_start_idx is None:
                 self._repositories.append(repository)
                 self._lookup[repository_name] = len(self._repositories) - 1
--- a/tests/fixtures/with_non_default_multiple_secondary_sources/pyproject.toml
+++ b/tests/fixtures/with_non_default_multiple_secondary_sources/pyproject.toml
@@ -1,0 +1,24 @@
+[tool.poetry]
+name = "my-package"
+version = "1.2.3"
+description = "Some description."
+authors = [
+    "Your Name <you@example.com>"
+]
+license = "MIT"
+
+# Requirements
+[tool.poetry.dependencies]
+python = "~2.7 || ^3.6"
+
+[tool.poetry.dev-dependencies]
+
+[[tool.poetry.source]]
+name = "foo"
+url = "https://foo.bar/simple/"
+secondary = true
+
+[[tool.poetry.source]]
+name = "bar"
+url = "https://bar.baz/simple/"
+secondary = true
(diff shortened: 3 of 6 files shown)


jstriebel:
It's not clear to me why the CI fails with
```
ValueError

Incorrectly nested style tag found.
```
Seems like the random error from #3010. The tests seem to work fine locally.

jstriebel:
@sdispater Since you merged the original PR, could you have a look at this cherry-pick for v1.1? Thanks a lot!

abn (approved):

abn:
@jstriebel approving and merging this one. However, please note that we do not at present have a plan for another `1.1` release.

caniko:
> @jstriebel approving and merging this one. However, please note that we do not at present have a plan for another `1.1` release.

But it is a broken feature, shouldn't broken features be patched? Especially for those who won't upgrade to the next major/minor?

jstriebel:
> However, please note that we do not at present have a plan for another 1.1 release.

@abn Understandable, however it would be very useful, especially considering that `1.2` introduces bugs that might keep someone from upgrading to that version (in my case #4254). I'm therefore using the commit of this PR to install poetry atm.

Issue #4245, opened 2021-07-02
Directory does not exist errors for path dependencies


md384:
- [x] I am on the [latest](https://github.com/python-poetry/poetry/releases/latest) Poetry version.
- [x] I have searched the [issues](https://github.com/python-poetry/poetry/issues) of this repo and believe that this is not a duplicate.
- [x] If an exception occurs when executing a command, I executed it again in debug mode (`-vvv` option).

- **OS version and name**: MacOS Big Sur
- **Poetry version**:  1.1.7
- **Link of a [Gist](https://gist.github.com/) with the contents of your pyproject.toml file**: NA

## Issue

`v1.1.7` can cause path dependencies to not be found when packages are not all at the same directory depth. This appears to happen because the dependencies are searched for relative to the lockfile rather than relative to the lockfile and package.

As an example, for the directory structure

```
/
  pyproject.toml
  poetry.lock
  /lib
    /libA
      pyproject.toml
      poetry.lock
    /libB
      pyproject.toml
      poetry.lock
```

and relevant part of the `poetry.lock` file for the root

```
[[package]]
name = "lib-a"
version = "0.1.0"
description = ""
category = "main"
optional = false
python-versions = "^2.7.9"
develop = true

[package.dependencies]
lib-b = {path = "../libB", develop = true}

[package.source]
type = "directory"
url = "lib/libA"

[metadata]
lock-version = "1.1"
python-versions = "*"
content-hash = "115cf985d932e9bf5f540555bbdd75decbb62cac81e399375fc19f6277f8c1d8"

[metadata.files]
lib-a = []
lib-b = []
```

In `1.1.7`, we get a `Directory ../lib-b does not exist` error because it tries to find the libB dependency at `/../lib-b` rather than `/lib/libA/../libB`.

This is likely a problem with previous versions of poetry but `1.1.7` changed the lockfiles for path dependencies to be paths rather than versions so becomes apparent when installing a package.

bstadlbauer:
This seems to be a duplicate/related to https://github.com/python-poetry/poetry/issues/3368

skovhus:
I'm experiencing the same issue. Downgrading to 1.1.6 is a temporary workaround.

Dylan-Sharp:
> I'm experiencing the same issue. Downgrading to 1.1.6 is a temporary workaround.

I downgraded poetry to v1.1.6 and still experience this same issue.

skovhus:
[...]

PR #4246, opened 2021-07-02
fix: create path dependencies relative to package rather than lockfile (#4245)

md384:
Resolves path dependencies relative to the package and lockfile rather than just the lockfile.

Resolves: #4245

### DIFF
--- a/poetry/packages/locker.py
+++ b/poetry/packages/locker.py
@@ -180,11 +180,7 @@ class Locker:
                 root_dir = self._lock.path.parent
                 if package.source_type == "directory":
                     # root dir should be the source of the package relative to the lock path
-                    root_dir = Path(
-                        os.path.relpath(
-                            Path(package.source_url), self._lock.path.parent
-                        )
-                    ).resolve()
+                    root_dir = Path(package.source_url)
 
                 if isinstance(constraint, list):
                     for c in constraint:
--- a/poetry/packages/locker.py
+++ b/poetry/packages/locker.py
@@ -176,20 +176,26 @@ class Locker:
                         package.marker = parse_marker(split_dep[1].strip())
 
             for dep_name, constraint in info.get("dependencies", {}).items():
+
+                root_dir = self._lock.path.parent
+                if package.source_type == "directory":
+                    # root dir should be the source of the package relative to the lock path
+                    root_dir = Path(
+                        os.path.relpath(
+                            Path(package.source_url), self._lock.path.parent
+                        )
+                    ).resolve()
+
                 if isinstance(constraint, list):
                     for c in constraint:
                         package.add_dependency(
-                            Factory.create_dependency(
-                                dep_name, c, root_dir=self._lock.path.parent
-                            )
+                            Factory.create_dependency(dep_name, c, root_dir=root_dir)
                         )
 
                     continue
 
                 package.add_dependency(
-                    Factory.create_dependency(
-                        dep_name, constraint, root_dir=self._lock.path.parent
-                    )
+                    Factory.create_dependency(dep_name, constraint, root_dir=root_dir)
                 )
 
             if "develop" in info:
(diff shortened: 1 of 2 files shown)


poetry/packages/locker.py
```diff
@@ -176,20 +176,26 @@ def locked_repository(
                         package.marker = parse_marker(split_dep[1].strip())
 
             for dep_name, constraint in info.get("dependencies", {}).items():
+
+                root_dir = self._lock.path.parent
+                if package.source_type == "directory":
```
md384, line 181: Not sure if we need to handle `file` source types too

poetry/packages/locker.py
```diff
@@ -176,20 +176,26 @@ def locked_repository(
...
+                    # root dir should be the source of the package relative to the lock path
+                    root_dir = Path(
+                        os.path.relpath(
```
abn, line 184: Couple of points;
1. Move to using pathlib if possible.
2. We cannot always be sure a "directory" package is relative as it could also be an absolute path.

abn (changes requested):
In addition to the comments in the reivew, I wonder if this is going to cause issues if `libB` from your issue is also a dependency of `libC`, I suspect not if `libC` is at the same location as `libA`, however could get a bit hairy if it is not and/or `libC`is nested in `libA`.

md384:
@abn to test this, I created a new project with the following structure

```
my-package/
  pyproject.toml
  lib/
    lib-a/
      lib-a/
      pyproject.toml
      lib-c/
        lib-c/
        pyproject.toml
    lib-b/
      lib-b/
      pyproject.toml
      lib-d/
        lib-d/
        pyproject.toml
  tools/
    my_tool/
      my_tool/
      pyproject.toml
```

I created nested libraries across directories like you suggested. `my-package` depends on `lib-a` which depends on `lib-b` which depends on `lib-c` which depends on `lib-d` which depends on `my_tool`. I added the relevant part of the poetry lock file for `my-project` (the root package) below.

```
[[package]]
name = "lib-a"
version = "0.1.0"
description = ""
category = "main"
optional = false
python-versions = "^3.9"
develop = true

[package.dependencies]
lib-b = {path = "../lib_b", develop = true}

[package.source]
type = "directory"
url = "lib/lib_a"

[[package]]
name = "lib-b"
version = "0.1.0"
description = ""
category = "main"
optional = false
python-versions = "^3.9"
develop = true

[package.dependencies]
lib-c = {path = "../lib_a/lib_c", develop = true}

[package.source]
type = "directory"
url = "lib/lib_b"

[[package]]
name = "lib-c"
version = "0.1.0"
description = ""
category = "main"
optional = false
python-versions = "^3.9"
develop = true

[package.dependencies]
lib-d = {path = "../../lib_b/lib_d", develop = true}

[package.source]
type = "directory"
url = "lib/lib_a/lib_c"

[[package]]
name = "lib-d"
version = "0.1.0"
description = ""
category = "main"
optional = false
python-versions = "^3.9"
develop = true

[package.dependencies]
my-tool = {path = "../../../tools/my_tool", develop = true}

[package.source]
type = "directory"
url = "lib/lib_b/lib_d"

[[package]]
name = "my-tool"
version = "0.1.0"
description = ""
category = "main"
optional = false
python-versions = "^3.9"
develop = true

[package.source]
type = "directory"
url = "tools/my_tool"
```

I was able to install and import all the libs from `my-package`.

```
my-package % poetry run python
Python 3.9.5 (default, May  4 2021, 03:36:27) 
[Clang 12.0.0 (clang-1200.0.32.29)] on darwin
>>> import lib_a
>>> import lib_b
>>> import lib_c
>>> import lib_d
>>> import my_tool
```

I hope no-one would do such a thing as this though 😉

md384:
@abn it turns out that we already calculate the relative path https://github.com/python-poetry/poetry/blob/97e63f629b9e669740603754f39a67ff1fa166dc/poetry/packages/locker.py#L119 so I am just using that now.

You commented before that a directory source url could be an absolute path. Does that line need to be changed?

sdispater (approved):

sdispater:
Thanks!

Issue #1644, opened 2019-11-28
Support dependencies groups


pawamoy:
- [x] I have searched the [issues](https://github.com/sdispater/poetry/issues) of this repo and believe that this is not a duplicate.
- [x] I have searched the [documentation](https://poetry.eustace.io/docs/) and believe that my question is not covered.

## Feature Request

As @mozartilize stated in [this comment](https://github.com/sdispater/poetry/issues/1007#issuecomment-496779710), [groups](https://bundler.io/guides/groups.html) from Ruby's gemfile could be a great addition and a solution to our complex workflows and different environments (local, production, ci, docs, etc.).

Instead of supporting more `tool.poetry.*-dependencies` sections, we could add support for a `groups` property on dependencies. It would be more flexible than rigid sections.

I also understand there is the `extras` functionality, but using extras changes the behavior of `poetry install`, as one needs to now specify the extras to install all the dependencies. Also, extras are opt-in only, and often times we need opt-out.

```toml
[tool.poetry.dependencies]
python = "^3.6"
requests = "*"

[tool.poetry.dev-dependencies]
bandit = { version = "^1.5", groups = ["ci"] }
black = { version = "*", allows-prereleases = true, groups = ["ci"] }
coverage = { version = "=5.0a8", allows-prereleases = true, groups = ["ci"] }
flake8 =  { version = "^3.6", groups = ["ci"] }
ipython =  { version = "^7.2", groups = ["local"] }
isort = { version = "^4.3", extras = ["pyproject"], groups = ["ci"] }
jinja2-cli =  { version = "^0.7.0", groups = ["local"] }
pylint = { git = "https://github.com/PyCQA/pylint", groups = ["ci"] }
pytest =  { version = "^4.3", groups = ["ci"] }
pytest-cov =  { version = "^2.8", groups = ["ci"] }
pytest-sugar =  { version = "^0.9.2", groups = ["ci"] }
pytest-xdist =  { version = "^1.26", groups = ["ci"] }
recommonmark =  { version = "^0.6.0", groups = ["docs"] }
safety =  { version = "^1.8", groups = ["ci"] }
sphinx =  { version = "^1.8", groups = ["docs"] }
sphinx-rtd-theme =  { version = "^0.4.2", groups = ["docs"] }
toml =  { version = "^0.10.0", groups = ["docs"] }
```

Also:

- `default` group would be attributed to dependencies without groups (`requests` in the above example)
[...]

PR #4260, opened 2021-07-07
Dependency groups

sdispater:
This PR adds support for dependency groups to have a more granular organization of dependencies.

For a complete list of changes and improvements you can refer to the updated documentation in this PR.

## Pull Request Check List

Resolves: #1644

### DIFF
--- a/docs/managing-dependencies.md
+++ b/docs/managing-dependencies.md
@@ -1,0 +1,148 @@
+---
+draft: false
+layout: single
+menu:
+  docs:
+    weight: 11
+title: Managing dependencies
+type: docs
+---
+
+
+# Managing dependencies
+
+## Dependency groups
+
+Poetry provides a way to **organize** your dependencies by **groups**. For instance, you might have
+dependencies that are only needed to test your project or to build the documentation.
+
+To declare a new dependency group, use a `tool.poetry.group.<group>` section
+where `<group>` is the name of your dependency group (for instance, `test`):
+
+```toml
+[tool.poetry.group.test]  # This part can be left out
+
+[tool.poetry.group.test.dependencies]
+pytest = "^6.0.0"
+pytest-mock = "*"
+```
+
+{{% note %}}
+All dependencies **must be compatible with each other** across groups since they will
+be resolved regardless of whether they are required for installation or not (see [Installing group dependencies]({{< relref "#installing-group-dependencies" >}})).
+
+Think of dependency groups as **labels** associated with your dependencies: they don't have any bearings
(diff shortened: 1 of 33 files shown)


docs/managing-dependencies.md
```diff
@@ -0,0 +1,141 @@
...
+type: docs
+---
+
+
+# Managing dependencies
+
+## Dependency groups
+
+Poetry provides a way to **organize** your dependencies by **groups**. For instance, you might have
+dependencies that are only needed to test you project or to build the documentation.
```
sbast, line 17:
```suggestion
dependencies that are only needed to test your project or to build the documentation.
```

tests/console/commands/test_remove.py
```diff
@@ -1,13 +1,161 @@
...
     return command_tester_factory("remove")
 
 
+def test_remove_without_specific_group_removes_from_all_groups(
+    tester, app, repo, command_tester_factory, installed
+):
+    """
+    Removing without specifying a group removes packages from all groups.
+    """
+    installed.add_package(Package("foo", "2.0.0"))
```
Leonardofreua, line 19: What do you think about maybe creating constants for the strings `baz`, `foo`, `1.0.0`, and `2.0.0`, since they are repeated so often?

If you agree, you will need to change the other test files that use these same values.

tests/installation/test_installer.py
```diff
@@ -334,43 +337,63 @@ def _configure_run_install_dev(locker, repo, package, installed):
...
+    installer, locker, repo, package, installed
+):
+    _configure_run_install_dev(
+        locker, repo, package, installed, with_optional_group=True
+    )
+
+    installer.with_groups(["dev"])
+    installer.run()
+
+    assert 0 == installer.executor.installations_count
```
Leonardofreua, line 392: Maybe create a constant for the value `0`?

tests/packages/test_locker.py
```diff
@@ -86,7 +86,7 @@ def test_lock_file_data_is_ordered(locker, root):
 [metadata]
 lock-version = "1.1"
 python-versions = "*"
-content-hash = "115cf985d932e9bf5f540555bbdd75decbb62cac81e399375fc19f6277f8c1d8"
+content-hash = "178f2cd01dc40e96be23a4a0ae1094816626346346618335e5ff4f0b2c0c5831"
```
Leonardofreua, line 89: It would be interesting to create a constant for `content-hash`. Here we see an example that if it had a constant, it would not be necessary to make 7 changes to the same value.

Leonardofreua (changes requested):

PR #4267, opened 2021-07-08
Fix version constraint operator in the docs FAQ

sisp:
The FAQ in the docs used the wrong version constraint operator for discouraging the use of `>=` by suggesting one that constrains the Python version to `>=3.4,<4`. The more concise operator `~3.4` was used, but it should be `^3.4`.


finswimmer (approved):
Thanks :+1: :)

Issue #3860, opened 2021-03-30
🧹 Remove compatibility code: Python < 3.6


abn:
This is a great issue to work on if you are new to the Poetry code base and want to start contributing. Over the years the poetry code base has added various fixes that catered to now unsupported versions of python. With the release of `1.2`, we will drop support for all Python versions prior to `3.6`. With this we want to clean up the code base of any such workarounds we did to support backwards compatibility and/or quirks.

We would like to invite first time contributors to work on helping us remove these bits of code that we no longer require. Small changes are welcome.

If you are submitting a pull request associated with this, please keep the pull requests small and modular. This means, do not do too much in one PR and try to do only closely related changes in on PR. Additionally, make sure you add this to your PR description.

```
Relates-to: #3860
```

If this is your first contribution here, please refer to #3061 before continuing.

abn:
This should now be done thanks largely to the pyugrade plugin.

PR #3988, opened 2021-04-22
Remove < Python 3.6 compatibility code

kamyar:
Hey, I saw the issue linked below and immediately thought of pyupgrade and as poetry already uses pre-commit added the config for it with target set to python 3.6.

I know the issue says "please keep the pull requests small and modular." but as the changes are autogenerated and rather simple(in terms of mental focus needed to review) I thought might be ok to submit them together.

Please let me know if multiple small PRs is still preferred, I will open related syntax updates together (either by file/directory or by type of change)

I already went through all changes twice and did not notice anything that would change the behaviour.
Also please let me know if there is something missing.

p.s. thank you for such a great tool, it has been a true delight to use it both at work and personal project. :)

p.p.s. below are some suggested extra changes that convert `f"{foo}"` to `foo`, if they look good I will create a single commit with the fixes rather than creating one commit per suggestion.

# Pull Request Check List
Resolves: #3860

- [x] Added **tests** for changed code.   **(Not Needed?)**
- [x] Updated **documentation** for changed code. **(Not Needed?)**

PR #4298, opened 2021-07-20
Replace some .format() in show.py by f-string (#3860).

Leonardofreua:
Has replaced some excerpts that were using `.format()` in the `show.py` (show command) file.

These points have been changed taking into account what has already been changed in PR #3988

# Pull Request Check List

Resolves:  #3860

- [ ] ~~Added **tests** for changed code.~~


neersighted:
We're probably going to implement this en-mass via `pyupgrade`. Thanks for the PR!

Issue #3368, opened 2020-11-16
Poetry install fails for nested local dependencies and develop = false


TrevorPace:
- [x] I am on the [latest](https://github.com/python-poetry/poetry/releases/latest) Poetry version.
- [x] I have searched the [issues](https://github.com/python-poetry/poetry/issues) of this repo and believe that this is not a duplicate.
- [x] If an exception occurs when executing a command, I executed it again in debug mode (`-vvv` option).

- Standard Python:3.8 docker image
- 1.1.4

## Issue

I am attempting to install a poetry based application with a monorepo-like hierarchy. For simplicity sake the structure looks something like:

```
/lib
    /pkg1
        pyproject.toml
    /pkg2
        pyproject.toml
/apps
    /app1
        pyproject.toml
```

The apps/app1/pyproject.toml is completely empty other than a reference to one of the libraries: `pkg1 = { path = "../../lib/pkg1", develop = false }`

pkg1 has then a similar local dependency on pkg2: `pkg2 = { path = "../pkg2", develop = false }`

An example repo of this is here: https://github.com/TrevorPace/poetry-localdep-issue

When attempting to run `poetry install` in `apps/app1` to create the initial poetry.lock file I get the following issues:

```
Updating dependencies
Resolving dependencies... (0.1s)

Writing lock file

Package operations: 2 installs, 0 updates, 0 removals

  • Installing pkg2 (0.1.0 /home/trevor/git/trevor.pace/poetry-localdep-issue/libs/pkg2)
  • Installing pkg1 (0.1.0 /home/trevor/git/trevor.pace/poetry-localdep-issue/libs/pkg1): Failed

  EnvCommandError

  Command ['/home/trevor/.cache/pypoetry/virtualenvs/app1-Jm016c7i-py3.8/bin/pip', 'install', '--no-deps', '-U', '/home/trevor/git/trevor.pace/poetry-localdep-issue/libs/pkg1'] errored with the following return code 1, and output:
  Processing /home/trevor/git/trevor.pace/poetry-localdep-issue/libs/pkg1
    Installing build dependencies: started
    Installing build dependencies: finished with status 'done'
    Getting requirements to build wheel: started
    Getting requirements to build wheel: finished with status 'done'
      Preparing wheel metadata: started
      Preparing wheel metadata: finished with status 'error'
      ERROR: Command errored out with exit status 1:
[...]

PR #4320, opened 2021-07-26
Fix nested local dependencies (#3368)

lucasvieirasilva:
This PR resolves the issue when the project has more than 2 levels of path dependencies.

```
/lib
    /pkg1
        pyproject.toml
    /pkg2
        pyproject.toml
/apps
    /app1
        pyproject.toml
```  

### Solution

Check when the source package is `directory` to use the property `source_url` as `root_dir` instead of using the `self._lock.path.parent` if the source package is not a directory type the installation process continues normally.

That makes [poetry-core](https://github.com/python-poetry/poetry-core/blob/master/poetry/core/packages/directory_dependency.py#L30) resolve correctly the full path property.

# Pull Request Check List

Resolves: #3368 

- [ ] <s>Updated **documentation** for changed code.</s>

### DIFF
--- a/poetry/packages/locker.py
+++ b/poetry/packages/locker.py
@@ -187,10 +187,13 @@ class Locker:
 
                     continue
 
+                root_dir = self._lock.path.parent
+
+                if package.source_type == "directory":
+                    root_dir = Path(package.source_url)
+
                 package.add_dependency(
-                    Factory.create_dependency(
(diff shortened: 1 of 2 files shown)


lucasvieirasilva:
Duplicated fix https://github.com/python-poetry/poetry/pull/4246

Issue #3957, opened 2021-04-16
Poetry removes it's own dependencies with `install --no-dev` when it manages the environment it is installed in


gbdlin:
- [x] I am on the [latest](https://github.com/python-poetry/poetry/releases/latest) Poetry version.
- [x] I have searched the [issues](https://github.com/python-poetry/poetry/issues) of this repo and believe that this is not a duplicate.  
  Well... This issue is mentioned by @sontek in https://github.com/python-poetry/poetry/issues/3139#issuecomment-776355443_ but the response below advises to create a new issue, so here I am.
- [x] If an exception occurs when executing a command, I executed it again in debug mode (`-vvv` option).

- **OS version and name**: Docker image based on `python:3.9.1`
- **Poetry version**: 1.1.6
- **Link of a [Gist](https://gist.github.com/) with the contents of your pyproject.toml file**: https://gist.github.com/gbdlin/311dda3cd34672370cd97bf07f306a61

## Issue
When you specify a package in the `dev-dependencies` that is either a dependency of the poetry itself or it depends on one of the poetry dependencies, and you've installed poetry inside an environment that poetry is supposed to manage, poetry will uninstall this mentioned dependency after invoking `poetry install --no-dev`, making itself unusable in the future.

I'm aware that it is advised to install poetry outside of the environment it should manage, but this is very often seen inside docker images (such example provided in the gist linked above) to save on the resources.

Example output presenting mentioned issue (without `-vvv` ands):

```
$ pip install poetry
Requirement already satisfied: pip in /usr/local/lib/python3.9/site-packages (21.0.1)
Collecting poetry
  Downloading poetry-1.1.6-py2.py3-none-any.whl (172 kB)
[ ... 70 irrelevant output lines truncated ... ]
Collecting appdirs<2,>=1.4.3
  Downloading appdirs-1.4.4-py2.py3-none-any.whl (9.6 kB)
Installing collected packages: [ ... ] appdirs, [ ... ] poetry
Successfully installed [ ... ] appdirs-1.4.4 [ ... ] poetry-1.1.6 [ ... ]

$ poetry config virtualenvs.create false && poetry install --no-dev

Skipping virtualenv creation, as specified in config file.
Installing dependencies from lock file

Package operations: 0 installs, 0 updates, 1 removal

  • Removing appdirs (1.4.4)
Removing intermediate container af920367a74d
```

[...]

PR #4336, opened 2021-07-31
Improve package operations management

sdispater:
This PR aims at improving the user experience when installing packages and at minimizing the install/uninstall operations when ignoring groups or opting-in groups.

An example of this is the following sequence:

```bash
$ poetry install
$ poetry install --without dev  # formerly --no-dev
```

It would remove the existing development packages which was counter-intuitive. Now, the existing development packages will be kept in the environment.

Any locked packages will be kept in the environment regardless of what is asked via the `install` command unless the `--sync` option is specified.

```bash
$ poetry install --without dev --sync
$ poetry install --only dev --sync
```

Overall, this should be less surprising for end users makes working with dependency groups easier.

## Pull Request Check List

### DIFF
--- a/docs/cli.md
+++ b/docs/cli.md
@@ -163,11 +163,19 @@ The `--dev-only` option is now deprecated. You should use the `--only dev` notat
 See [Dependency groups]({{< relref "managing-dependencies#dependency-groups" >}}) for more information
 about dependency groups.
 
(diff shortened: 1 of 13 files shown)


a5r0n:
👍🏻 
#3957 and #3139 related

PR #4339, opened 2021-08-01
Run tests involving file fixtures in a temporary directory

swiatekm:
This change allows the test suite to be run in parallel and installer tests to be run from any directory. The runtime goes from ~60 seconds single-threaded to ~17 seconds with [pytest-xdist](https://github.com/pytest-dev/pytest-xdist) on my machine with maximum parallelism.

A major barrier towards running tests in parallel is that some of them modify files in project fixtures. As a simple fix, copy the whole project fixture directory tree to a temporary folder and run tests there. Some of the fixtures for installer tests contain paths to the project fixtures. To make those work, now that the paths are temporary, make installer tests use the project fixture directory as their working directory.

This is far from a comprehensive cleanup, as the test and fixture dependencies are still difficult to understand. I have some ideas on how they could be made more explicit and consistent (and possibly more performant), but I wanted to put this change up for review before diving in too deep.

On a separate note, fixtures and helpers for the new and old installer tests are mostly the same and can be factored out and shared. This PR just makes the same change in both of those modules to limit the scope and make review easier, but I do have those changes ready, and can add them here if needed.

# Pull Request Check List

Relates-to: #3155

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -109,6 +109,11 @@ def with_chained_keyring(mocker):
     keyring.set_keyring(ChainerBackend())
 
 
+# The fixtures directory. We copy files from here to a test-specific execution
+# directory, as some tests mutate the files within.
+FIXTURE_DIR = Path(__file__).parent / "fixtures"
+
+
 @pytest.fixture
 def config_cache_dir(tmp_dir):
     path = Path(tmp_dir) / ".cache" / "pypoetry"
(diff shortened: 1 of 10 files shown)


Secrus:
Hi, @swiatekm are you still interested in contributing this to Poetry?

branchvincent:
Superseded by #5745

Issue #4035, opened 2021-05-04
Add ability to specify package sources that are not scanned/used by default


jgentil:
- [x] I have searched the [issues](https://github.com/python-poetry/poetry/issues) of this repo and believe that this is not a duplicate.
- [x] I have searched the [documentation](https://python-poetry.org/docs/) and believe that my question is not covered.

## Feature Request

When specifying additional sources in pyproject.toml as "secondary", the current behavior of Poetry is to include this source in all of it's scans for dependency resolution for adding or installing packages.  This is almost entirely unnecessary in many cases, as a secondary source is normally only needed to install very specific packages and would not normally host any other packages from pypi (and if they do, it's likely not "secondary" then but "default".) 

Since this behavior is already defined, I'd like to propose a third option, `targeted` which specifies that a source should be used only if a dependency definition explicitly specifies it in it's `source` parameter.  That package's dependencies would be scanned from that repository, but it would be otherwise ignored.

One specific reason that this is desperately needed is that in several of my corporate projects, we use Azure DevOps artifacts system private PyPI.  This system is notoriously slow and a small project's dependency resolution scan take upwards of 10 minutes with Poetry scanning it for every single project dependency.  To mitigate this, we currently force `source="pypi"` in all dependency specifications to target it specifically.

zevisert:
+1 for this! 

Adding fuel to the fire, I can view the access log for my private package repository -- and poetry _always_ scans `[{tool.poetry.source = { default=false, secondary=true }]` sources for all transitive dependencies, even if we use your `source="pypi"` mitigation. 

I tested with a project that has no private dependencies, and by adding a new source, but without changing dependencies at all the `poetry update` time went from `0.32` to over `200` seconds!

zevisert:
With transitive dependencies in mind, the only way I can think to address this is with your suggestion for a `targeted` or alike key in the `tool.poetry.source` table.

zevisert:
I'd really love to move this feature forward! Every time I need to update packages at work I have to sit there watching poetry `1.1.11` take 300+ seconds in most of our projects to resolve dependencies is annoying when I think I have a PR I have can do it in less than 15.. 

I just heard about @python-poetry/triage Hi! 👋

IWillPull:
I'm also experiencing it and having builds take 30+ minutes longer when using one `secondary` dependency sounds more like a bug, not a feature request.

dimbleby:
this should be absorbed into #6713

PR #4353, opened 2021-08-04
Use targeted repositories that are only scanned explicitly

zevisert:
This PR adds functionality for "targeted" repositories, which are not scanned during solving unless a package explicitly marks it as a source.

Example: 

```toml
[[tool.poetry.source]]
name = "slow"
targeted = true

[tool.poetry.dependencies]
python = "^3"
your_package = { version = "^1", source="slow" }
```

# Pull Request Check List

Resolves: #4035 

### DIFF
--- a/docs/repositories.md
+++ b/docs/repositories.md
@@ -148,6 +148,30 @@ a custom certificate authority or client certificates, similarly refer to the ex
 `certificates` section. Poetry will use these values to authenticate to your private repository when downloading or
 looking for packages.
 
+#### Using targeted repositories
+
+Some repositories are slow and can negatively affect the speed of dependency resolution. You can add
+slow repositories using `targeted` mode, which forces poetry to only communicate with the repository
+for packages that explicitly use the repository as their source.
+
+```toml
+[[tool.poetry.source]]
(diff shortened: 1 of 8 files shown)


zevisert:
I haven't yet added a test to confirm that solving _doesn't_ communicate with a targeted source unless explicitly requested. The tests started to look pretty complicated with all of the `MockRespository` functionality, so I thought it best to first check if there's interest in merging this PR at all.

neersighted:
I'm not sure this is the approach we are going to take -- it's generally agreed that the current design is poor, but we would like to make a breaking change and improve it instead of tacking on more variant settings that increase the difficulty in reasoning. I would suggest joining the Discord server to discuss this with the core team, or creating an issue to talk about a new design for repositories.

I'm going to close this for now as this won't get merged in its current form -- feel free to re-open if suggestions for a new design occur and they crystalize into code, however.

zevisert:
No problem, the code changes here were mostly just hacks to get the idea somewhat demonstrable. I agree with that sentiment tacking on more hacks is not the way to go! 

🚀

Issue #4354, opened 2021-08-04
Poetry detects Python 3.10 as Python 3.1


9999years:
```python
from pathlib import Path

import poetry.utils.env

print(
    poetry.utils.env.VirtualEnv(path=Path("./.tox/py310")).marker_env["python_version"]
)
# Prints "3.1"
```

- **OS version and name**: Fedora 34
- **Poetry version**: 1.1.7

9999years:
Looks like this originated in https://github.com/python-poetry/poetry/commit/7a12c390d2e9e04566d36eb774db8aa64822a3be#diff-48120c42d1d0a5860b317966e7d1e0a65e48e4fee6f6a0795fad5848dcf43ef0

apollo13:
Dupe of the older https://github.com/python-poetry/poetry/pull/4340 which just got merged

PR #4340, opened 2021-08-01
Fixed python_version for Python >= 3.10

apollo13:
This seems to be a rather trivial fix. I am not sure where to put tests exactly (if needed at all, pointers welcome)

PR #4355, opened 2021-08-04
Fix Python 3.10 version being detected as 3.1

9999years:
Resolves: #4354 

- [ ] Added **tests** for changed code. Testing requires mocking `import sys`, which is difficult to say the least. We could use something like `forbiddenfruit` but it seems like overkill. What do you think?
- [ ] Updated **documentation** for changed code. N/A


9999years:
Fixed in #4340

PR #4392, opened 2021-08-17
remove import of __future__.print_function

venthur:
This PR removes all remaining imports of `__future__.print_function`.
This is not needed anymore since we support Python versions >= 3.6

Relates-to: #3860

# Pull Request Check List

Resolves: #issue-number-here


neersighted:
Obsolete as this is already fixed in master. Thanks for the PR!

PR #4393, opened 2021-08-17
Remove all remaining imports of __future__.absolute_import

venthur:
This removes all remaining imports of `__future__.absolute_import` as
this is not needed anymore in Python >= 3.6

Relates-to: #3860

# Pull Request Check List

Resolves: #issue-number-here


neersighted:
It looks like this already was implemented in `master`. Thanks for the PR, and please do submit PRs for any Python compatibility/legacy code you find!

PR #4394, opened 2021-08-17
removed all remaining "coding: utf-8" headers

venthur:
we don't need them anymore since we only support python >= 3.6

Relates-to: #3860

# Pull Request Check List

Resolves: #issue-number-here


neersighted:
Closed due to staleness -- all the `coding:` headers lest the one in `get-poetry.py` are removed, and as `get-poetry.py` is in maintenance mode I am inclined to let it be.

Issue #4399, opened 2021-08-18
Poetry 1.2.0a2 - Experimental installer: pip install fails for some packages


orlevii:
- [x] I am on the [latest](https://github.com/python-poetry/poetry/releases/latest) Poetry version.
- [x] I have searched the [issues](https://github.com/python-poetry/poetry/issues) of this repo and believe that this is not a duplicate.
- [x] If an exception occurs when executing a command, I executed it again in debug mode (`-vvv` option).

- **Poetry version**: 1.2.0a2

## Issue

Reproducible easily with this `pyproject.toml`:
```toml
[tool.poetry]
name = "installer-bug"
version = "0.1.0"
description = ""
authors = ["Or Levi <orlevi128@gmail.com>"]

[tool.poetry.dependencies]
python = "~3.7.1"
pystan = "2.19.1.1"
fbprophet = "^0.7.1"


[build-system]
requires = ["poetry-core"]
build-backend = "poetry.core.masonry.api"
```

The experimental installer will fail to install `fbprophet`
The error message is:
```    ERROR: Failed building wheel for fbprophet
  Failed to build fbprophet
  ERROR: Could not build wheels for fbprophet which use PEP 517 and cannot be installed directly
```

After some debugging I've found out the new installer is generating a different pip command then the old one.
**New installer 1.2.0a2:** (fails)
```
/Users/or.levi/projects/poetry/_bug/venv/bin/python \
    /Users/or.levi/projects/poetry/venv/lib/python3.7/site-packages/virtualenv/seed/wheels/embed/pip-21.0.1-py3-none-any.whl/pip install \
    --use-pep517 \
    --disable-pip-version-check \
    --prefix /Users/or.levi/projects/poetry/_bug/venv \
    --no-deps \
    file:/Users/or.levi/Library/Caches/pypoetry/artifacts/9f/c1 \
    /4f/7442d1f86f6dafe5693a8f472a4047ced6db8760c461ddef1645c4f34c/fbprophet-0.7.1.tar.gz
```

Looks like the `--use-pep517` makes some problems here.
Also, the pip commands that works are completely different for some reason.

-----------

**Old Installer 1.2.0a2:** (works)
```
/Users/or.levi/projects/poetry/_bug/venv/bin/python \
    /Users/or.levi/projects/poetry/venv/lib/python3.7/site-packages/virtualenv/seed/wheels/embed/pip-21.0.1-py3-none-any.whl/pip install \
[...]

PR #4400, opened 2021-08-18
fix: remove --use-pep517 flag from pip command

orlevii:
# Pull Request Check List

Resolves: https://github.com/python-poetry/poetry/issues/4399

### DIFF
--- a/poetry/utils/pip.py
+++ b/poetry/utils/pip.py
@@ -25,9 +25,6 @@ def pip_install(
     # network call that adds a lot of wait time when installing a lot of packages.
     args = ["install", "--disable-pip-version-check", "--prefix", str(environment.path)]
 
-    if not is_wheel:
-        args.insert(1, "--use-pep517")
-
     if upgrade:
         args.append("--upgrade")
 


neersighted:
I think this PR is misguided at best. What's happening here is the `pystan` [release archive](https://files.pythonhosted.org/packages/10/9f/dcecf31ac08842a675ec337a14439b37d75f6da42f034b1fd5eb90b10b84/pystan-2.19.1.1.tar.gz
) is broken. It cannot be built under PEP 517, which specifies an isolated environment for the build tools.

Try reproducing with the following:

```sh
docker run --rm -i --entrypoint bash python:3.9 <<EOF
set -xe
curl -LO https://files.pythonhosted.org/packages/10/9f/dcecf31ac08842a675ec337a14439b37d75f6da42f034b1fd5eb90b10b84/pystan-2.19.1.1.tar.gz
tar xvf pystan-2.19.1.1.tar.gz
cd pystan-2.19.1.1
python -m pip install build
python -m build
EOF
```

While setuptools is compatible with PEP 517, checking that your dependencies are installed by importing them is non-standard and not supported.
A wheel (for your platform) would partially solve this, but is not uploaded.

I would suggest migrating to a newer pystan which solves these build woes, or if that is not possible, vendoring the code and patching the checks out of `setup.py`. While building with classical setuptools (in a non-isolated environment) does technically not trigger this, Poetry is moving to PEP-517 installs only (as is pip [in the future](https://github.com/pypa/pip/issues/9175)), and as such, I don't think this is an appropriate fix.

PR #4417, opened 2021-08-20
Adds File existence check to Uploader

1nF0rmed:
This handles the TODO for checking the existence of the files being
uploaded by the Uploader class.
https://github.com/python-poetry/poetry/blob/c967a4a5abc6a0edd29c57eca307894f6e1c4f16/poetry/publishing/uploader.py#L229
In the case that a file does not exist, it raises a UploadError.

The raised error message is as follows:
```Wheel or Tar files associated with the Project Package do not exist```

# Pull Request Check List

Resolves: NA

- Updated **documentation** for changed code. [NA]

Questions:

- At the moment, the error message being shown is pretty generic. Should I be specific with the file that could not be found?

### DIFF
--- a/src/poetry/publishing/uploader.py
+++ b/src/poetry/publishing/uploader.py
@@ -212,8 +212,6 @@ class Uploader:
         skip_existing: bool = False,
     ) -> None:
         for file in self.files:
-            # TODO: Check existence
-
             self._upload_file(session, url, file, dry_run, skip_existing)
 
     def _upload_file(
@@ -225,6 +223,9 @@ class Uploader:
         skip_existing: bool = False,
     ) -> None:
         from cleo.ui.progress_bar import ProgressBar
+
+        if not file.is_file():
+            raise UploadError(f"Archive ({file}) does not exist")
 
         data = self.post_data(file)
         data.update(
--- a/tests/publishing/test_uploader.py
+++ b/tests/publishing/test_uploader.py
@@ -119,3 +119,14 @@ def test_uploader_skip_existing_bubbles_unskippable_errors(
 
     with pytest.raises(UploadError):
         uploader.upload("https://foo.com", skip_existing=True)
+
+
+def test_uploader_properly_handles_file_not_existing(
+    mocker: MockerFixture, http: type[httpretty.httpretty], uploader: Uploader
+):
+    mocker.patch("pathlib.Path.is_file", return_value=False)
+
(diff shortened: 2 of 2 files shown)

### CODE REVIEW

**Comment by 1nF0rmed** [comment]
@finswimmer Please can you review my PR?

**Review by abn — DISMISSED** [review]

**Comment by neersighted** [comment]
Could you rebase onto `master` and implement @abn's feedback? Thanks!

**Comment by 1nF0rmed** [comment]
Sure - I'll rebase this and update to match the suggestions :)

**Comment by neersighted** [comment]
Looks like a bad rebase. I'd suggest reverting using your reflog, fetching `upstream` and rebasing onto `FETCH_HEAD`.

File: src/poetry/publishing/uploader.py [hunk 1]
```diff
@@ -225,6 +224,10 @@ def _upload_file(
     ) -> None:
         from cleo.ui.progress_bar import ProgressBar
 
+        if not file.is_file():
+            self._io.write_error(f"Archive ({file}) does not exist", True)
```
neersighted, line 228: I don't know that this `write_error` is needed -- the error message should be communicated to the user just through `raise`, right?

1nF0rmed: It should error out to the user but in the previous comment, it was mentioned to write it via `self._io`.

neersighted: The comment above meant the two options to be mutually exclusive -- that is to say, are we better off dropping everything on the floor (`raise`), or writing a message (`self._io`) and explaining we ran into error at the end (non-zero status code). I'm inclined to lean towards the first without any strong feelings from anyone else.

neersighted (dismissed):

neersighted:
pre-commit.ci autofix

tests/publishing/test_uploader.py
```diff
@@ -93,6 +93,15 @@ def test_uploader_registers_for_appropriate_400_errors(
     assert register.call_count == 1
 
 
+def test_uploader_properly_handles_file_not_existing(mocker, http, uploader):
```
neersighted, line 96:
```suggestion
def test_uploader_properly_handles_file_not_existing(
    mocker: MockerFixture, http: type[httpretty.httpretty], uploader: Uploader
):
```

neersighted (approved):

neersighted (approved):

PR #4420, opened 2021-08-21
Backport security fix #3885 for package hashes to 1.1

pietrodn:
Backport the security fix of #3885 on Poetry 1.1.
This commit has already been merged in `master` but it didn't make it to the 1.1 minor.

This fix is one of the two changes to make Poetry check the hashes of the downloaded files after `poetry.lock`.

Unfortunately this PR isn't sufficient to correct the bug, as [this other PR](https://github.com/python-poetry/poetry-core/pull/159) needs to be backported to `poetry-core` 1.0.
Still, it is necessary.

### DIFF
--- a/poetry/installation/chooser.py
+++ b/poetry/installation/chooser.py
@@ -109,6 +109,11 @@ class Chooser:
 
             selected_links.append(link)
 
+        if links and not selected_links:
+            raise RuntimeError(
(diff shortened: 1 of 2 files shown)


jowparks:
FYI this backport broke our build since there is still a bug somehow with external pypi repos and `md5` hashes

I downgraded our poetry to `poetry==1.0.10` to fix for now.

error is documented here:

Issue #4016, opened 2021-04-30
Further SSL problems with Poetry and a private Gitlab package registry


marshalc:
- [x] I am on the [latest](https://github.com/python-poetry/poetry/releases/latest) Poetry version.
- [x] I have searched the [issues](https://github.com/python-poetry/poetry/issues) of this repo and believe that this is not a duplicate.
- [x] If an exception occurs when executing a command, I executed it again in debug mode (`-vvv` option).

- **OS version and name**: macOS 11.3
- **Poetry version**: 1.1.6

## Issue

Closely related to #745 and #1012, I am experiencing problems working with poetry against a private (Gitlab) based package registry, on an internal project. The locked nature of this make this very hard for me to do a public replication of the problem, so apologies whilst I have to be descriptive (if a little vague for some sensitive info).

I have two poetry based python+django projects. The first (`atlas-models`) I have managed to package as a reusable app, and build and publish (with just poetry) to our internal Gitlab Package Registry. Presently there are two versions of the package in the registry (but I've tried with between 1 & 3 versions there, that doesn't appear to be a factor).

In my second project, when I attempt to add the first package via `poetry add atlas-models -vvv`, I get the following (abbreviated) output:

```
➜  ouh-mpages-development-repo git:(master) ✗ poetry add atlas-models -vvv

Using virtualenv: /Users/carl/Projects/pages.oxnet/documentation/ouh-mpages-development/ouh-mpages-development-repo/.venv
pages-oxnet-atlas: 2 packages found for atlas-models *
PyPI: No packages found for atlas-models *
Using version ^0.3.1 for atlas-models

Updating dependencies
Resolving dependencies...
   1: fact: mpages mkdocs build is 0.1.0
   1: derived: mpages mkdocs build
   1: fact: mpages mkdocs build depends on Django (^3.2)
   1: fact: mpages mkdocs build depends on atlas-models (^0.3.1)
   1: selecting mpages mkdocs build (0.1.0)
   1: derived: atlas-models (>=0.3.1,<0.4.0)
   1: derived: Django (>=3.2,<4.0)
pages-oxnet-atlas: 1 packages found for atlas-models >=0.3.1,<0.4.0
PyPI: No packages found for atlas-models >=0.3.1,<0.4.0
   1: fact: atlas-models (0.3.1) depends on Django (>=3.2,<4.0)
   1: fact: atlas-models (0.3.1) depends on cx-oracle (>=8,<9)
[...]

PR #4446, opened 2021-08-27
Python poetry#4016

dabla:
Resolves: #4016

I'm experiencing the same issue on poetry 1.1.8 but I have created this patch to solve this issue. The problem is that by default, the requests library doesn't retry when a connection fails, and CacheControlAdapter also doesn't pass any retry parameters.

In our case this issue only arises on Windows behind an enterprise proxy, not on Linux, but as most developers work on Windows machines, we needed to fix this issue. We didn't experience this issue with poetry 1.0.10 though.

### DIFF
--- a/poetry/cache/cache_control.py
+++ b/poetry/cache/cache_control.py
@@ -1,0 +1,32 @@
+from cachecontrol.adapter import CacheControlAdapter
+from cachecontrol.cache import DictCache
+from urllib3.util.retry import Retry
+
+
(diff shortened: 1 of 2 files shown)


Secrus:
Hi @dabla  are you still interested in contributing this to Poetry? If so, please change the target branch to master, merge changes and resolve conflicts.

Secrus:
Closing. No updates for a long time.

PR #4466, opened 2021-09-02
Fix printing install errors redirected to stdout

nikolaik:
After #4443 stderr was redirected to stdout, so stderr would be None and throw an error.

https://github.com/python-poetry/poetry/blob/a1a0a1d3f5e870dbf0afb6bedb5dbff1be8fca19/install-poetry.py#L588


finswimmer (approved):
Thanks a lot :+1:

Wikigoagec:
[	["grc"] = "polytonic",](url)

PR #4467, opened 2021-09-02
Fix installing under MinGW on Windows

s0600204:
`python[.exe]`, when installed on Windows via (msys2) MingW, can be found under a path structure similar to that you might find on *nix systems.

The `install-poetry.py` script was failing to take this into consideration, causing installation issues in the above case.

A similar fault in poetry itself was resolved in #3713.

----

### Pull Request Check List

Related: #2867, #3713

    - *There don't appear to be any tests for `install-poetry.py` (except the CI run).*

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,7 +19,7 @@ The [complete documentation](https://python-poetry.org/docs/) is available on th
 Poetry provides a custom installer that will install `poetry` isolated
 from the rest of your system.
 
-### osx / linux / bashonwindows install instructions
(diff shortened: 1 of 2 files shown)


neersighted:
`install-poetry.py` has been migrated to https://github.com/python-poetry/install.python-poetry.org. Please open a PR over there with your changes.

I am not sure when we will migrate them back to this repo, but given the canonical source has moved, that will be the best place to send your PR.

s0600204:
Ok, noted. Thanks for the information.

Issue #4474, opened 2021-09-04
Install error not shown on stdout 


sernst:
- [x] I am on the [latest](https://github.com/python-poetry/poetry/releases/latest) Poetry version.
- [x] I have searched the [issues](https://github.com/python-poetry/poetry/issues) of this repo and believe that this is not a duplicate.
- [x] If an exception occurs when executing a command, I executed it again in debug mode (`-vvv` option).

- **OS version and name**: alpine linux 3.14.2
- **Poetry version**: N/A
- **Link of a [Gist](https://gist.github.com/) with the contents of your pyproject.toml file**: N/A

## Issue

When trying to install poetry on alpine linux, I encountered the following:

```
Retrieving Poetry metadata

# Welcome to Poetry!

This will download and install the latest version of Poetry,
a dependency and package manager for Python.

It will add the `poetry` command to Poetry's bin directory, located at:

/root/.local/bin

You can uninstall at any time by executing this script with the --uninstall option,
and these changes will be reverted.

Installing Poetry (1.1.8): Installing Poetry
Traceback (most recent call last):
  File "/root/.install-poetry.py", line 440, in run
    self.install(version)
  File "/root/.install-poetry.py", line 464, in install
    self.install_poetry(version, env_path)
  File "/root/.install-poetry.py", line 585, in install_poetry
    subprocess.run(
  File "/usr/lib/python3.9/subprocess.py", line 528, in run
    raise CalledProcessError(retcode, process.args,
subprocess.CalledProcessError: Command '['/root/.local/share/pypoetry/venv/bin/python', '-m', 'pip', 'install', 'poetry==1.1.8']' returned non-zero exit status 1.

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/root/.install-poetry.py", line 833, in <module>
    sys.exit(main())
  File "/root/.install-poetry.py", line 829, in main
    return installer.run()
  File "/root/.install-poetry.py", line 443, in run
    colorize("error", f"\nAn error has occurred: {e}\n{e.stderr.decode()}")
AttributeError: 'NoneType' object has no attribute 'decode'
```

In this case it was because there was a dependency not available, e.g. `gcc`. But instead of seeing that output, I got the error within the install script itself. It turns out that this is because the error output for these missing packages appears in stdout and not stderr.

To reproduce this:

```
docker run --rm -it --entrypoint=/bin/sh alpine
```

and then in the container:

```
apk update 
apk add python3
[...]

PR #4475, opened 2021-09-04
Fix Install Error Not Shown (#4474)

sernst:
An AttributeError in the install script happens when an install operation failure doesn't result in stderr output. In these cases, it
appears the data is written to stdout instead. To fix this, the error output will use stdout when no stderr output exists. If neither of these exist a "No output available." message will be shown instead to prevent the AttributeError from obscuring what is actually going on in the flow of the install script.

# Pull Request Check List

Resolves: #4474

### DIFF
--- a/install-poetry.py
+++ b/install-poetry.py
@@ -439,9 +439,8 @@ class Installer:
         try:
             self.install(version)
         except subprocess.CalledProcessError as e:
-            print(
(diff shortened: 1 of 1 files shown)


finswimmer:
Thanks a lot your contribution @sernst :+1: 

This one is superseded by #4466

PR #4425, opened 2021-08-22
Check for returncode when pip installing Poetry in `install-poetry.py`

mkniewallner:
# Pull Request Check List

Resolves: #4424

- [ ] ~Added **tests** for changed code.~ Not applicable
- [ ] ~Updated **documentation** for changed code.~ Not applicable

Issue #4424, opened 2021-08-22
`install-poetry.py` doesn't exit with proper status code when installation fails


mkniewallner:
- [ ] ~I am on the [latest](https://github.com/python-poetry/poetry/releases/latest) Poetry version.~ Not applicable, as it's on the installer
- [x] I have searched the [issues](https://github.com/python-poetry/poetry/issues) of this repo and believe that this is not a duplicate.
- [ ] ~If an exception occurs when executing a command, I executed it again in debug mode (`-vvv` option).~ Not applicable, as it's on the installer

- **OS version and name**: ArchLinux and Debian 11

## Issue
`install-poetry.py` doesn't exit with the proper exit code if the installation of Poetry using `pip` failed.
This is due to `subprocess.run` command in `install_poetry` not using `check=True` to make sure that an exception is raised if the output returned with an exit code other than 0: https://github.com/python-poetry/poetry/blob/c967a4a5abc6a0edd29c57eca307894f6e1c4f16/install-poetry.py#L584-L588
See documentation for `subprocess.run` https://docs.python.org/3/library/subprocess.html#subprocess.run
```
If check is true, and the process exits with a non-zero exit code, a CalledProcessError exception will be raised.
```

### Reproducibility
To showcase the issue (and also show what the expected behaviour should be with a fix), I've created a dedicated repository with 2 Dockerfile that GitHub Actions ran on: https://github.com/mkniewallner/showcase-poetry-installer-issue

In [the workflow](https://github.com/mkniewallner/showcase-poetry-installer-issue/actions/runs/1156455315), two jobs have been defined:
- `poetry-master`: Tries to install poetry using Python 3.10.0rc1 via `install-poetry.py` from `python-poetry/poetry` repository
- `poetry-with-fix`: Tries to install poetry using Python 3.10.0rc1 via `install-poetry.py` from `mkniewallner/poetry` repository on [a commit](https://github.com/mkniewallner/poetry/commit/84745973f308eb80013a529279cfdbcb284de013) containg a potential fix

On both jobs, it wasn't possible to install Poetry because of missing system dependencies for `cffi` module, but:
- on `poetry-master`, the step that tries to install Poetry doesn't return with a non-zero exit code, as it run `poetry --version` just after (see [logs](https://github.com/mkniewallner/showcase-poetry-installer-issue/runs/3394866749?check_suite_focus=true))
[...]

PR #4476, opened 2021-09-05
Use local `install-poetry.py` on CI for tests and test against Python 3.10.0-rc.2 on all platforms

mkniewallner:
# Pull Request Check List

- [x] ~Added **tests** for changed code.~ Not applicable
- [x] ~Updated **documentation** for changed code.~ Not applicable

## Context
When working on https://github.com/python-poetry/poetry/issues/4424 and its associated PR https://github.com/python-poetry/poetry/pull/4425, I unfortunately broke installer for Python 3.6, because `capture_output` is not available in Python 3.6.
First of all, sorry about that, I should have been more careful.

The fact that there's no tests for `install-poetry.py` script couldn't prevent the mistake from happening.

## Proposal
This pull request updates the CI workflow that runs tests so that we install poetry using the installer from the current branch, rather than the one on `master` branch.

While this would not replace full functional tests, this could be a quick win to ensure that poetry is still installable on all supported Python versions and OSes on pull requests that update `install-poetry.py`.
To showcase this, I've reintroduced the change that made the installer fail on Python 3.6 [here](https://github.com/mkniewallner/poetry/commit/c8578319d4c088d54bb8ada59d27e9aeed62a0f1), which correctly makes the workflow fail for Python 3.6 jobs [on the CI](https://github.com/mkniewallner/poetry/runs/3517983970?check_suite_focus=true).

Alternatively, if we have other reasons to install poetry from `master` branch, we could add a dedicated job that specifically tests the installation of poetry through `install-poetry.py`, and keep installing poetry from `master` branch when running unit tests.

This PR also updates the matrix so that we run the tests on Python 3.10.0-rc.2, and make sure that it makes the workflow fail if the tests don't pass, as the RC2 is IMO stable enough to be worth having green tests.
If this part should stay as is, I'd be happy to revert the commit, or make necessary changes.

Note that quotes have been added for versions because it avoid YAML to interpret them as floats (see https://github.com/actions/setup-python/pull/175 and https://github.com/actions/setup-python/issues/198).

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -41,9 +41,7 @@ jobs:
         shell: bash
         env:
           BOOTSTRAP_ARGS: ${{ startsWith(matrix.python-version, '3.10') && '--git https://github.com/python-poetry/poetry.git' || '' }}
-        run: |
-          curl -sL https://raw.githubusercontent.com/python-poetry/poetry/master/install-poetry.py \
-            | python - -y ${{ env.BOOTSTRAP_ARGS }}
+        run: python install-poetry.py -y ${{ env.BOOTSTRAP_ARGS }}
 
       - name: Update PATH
         if: ${{ matrix.os != 'Windows' }}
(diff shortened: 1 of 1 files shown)


neersighted:
I'm closing this for now as `install-poetry.py` has been broken out into its own repo. We will likely try to factor it out of workflows in the future -- if you'd like to submit more granular versions of this PR, you are welcome to!

PR #4488, opened 2021-09-08
Prefer authenticator from repository when available

kaos:
Signed-off-by: Andreas Stenius <andreas.stenius@svenskaspel.se>

Resolves: #1012 #2593 #3110 #4016 
Supersedes: #3349
Introduced: #2595

Feature toggle to side step this issue until resolved:
```
poetry config experimental.new-installer false
```

This will ensure that the configured certificates and auth are used when downloading files from a legacy repository.
There are quite a few issues relating to this being broken, so the list of resolves above may be incomplete.

### DIFF
--- a/poetry/repositories/legacy_repository.py
+++ b/poetry/repositories/legacy_repository.py
@@ -28,7 +28,6 @@ from poetry.core.semver.version_constraint import VersionConstraint
 from poetry.core.semver.version_range import VersionRange
 from poetry.locations import REPOSITORY_CACHE_DIR
 from poetry.utils.helpers import canonicalize_name
-from poetry.utils.helpers import download_file
 from poetry.utils.helpers import temporary_directory
 from poetry.utils.patterns import wheel_file_re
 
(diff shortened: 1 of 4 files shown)


poetry/repositories/legacy_repository.py
```diff
@@ -444,6 +443,3 @@ def _get(self, endpoint: str) -> Optional[Page]:
             )
 
         return Page(response.url, response.content, response.headers)
-
-    def _download(self, url, dest):  # type: (str, str) -> None
-        return download_file(url, dest, session=self.session)
```
kaos, line 449: Removed this, as it is an exact copy of the one from the base class, `PypiRepository`.

kaos:
I can't fathom that of all the issues that has been reported with custom pypi registries and SSL issues, that no one picked up that the new installer can be turned off:
https://github.com/python-poetry/poetry/pull/2595#issuecomment-704875923

From what I understand, this gives you the same installer that is used in 1.0.10, that several users reports works as it should. Even downgrading to as a work-around for the issues.

kaos:
Ping. Any feedback on this?

neersighted:
Closed and re-opened to re-trigger checks being run.

Minhtikeaung (approved):

neersighted:
This should have been made obsolete by #5518 -- please test on master and comment/re-open if you think some of these changes are still valid.

PR #4514, opened 2021-09-17
tests: Make test_installer.py test suite pass when run from any directory

KochankovID:
# Pull Request Check List

Resolves: #3155

I found out that the problem was wrong directory for the poerty.lock
It created in the working directory, but tests expected that poetry.lock would be in the root directory.
So I made it relatively.

### DIFF
--- a/install-poetry.py
+++ b/install-poetry.py
@@ -528,6 +528,10 @@ class Installer:
 
             virtualenv.cli_run([str(env_path), "--clear"])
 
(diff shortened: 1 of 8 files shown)


neersighted:
Hi @KochankovID, and thanks for opening a PR! The PR #4585 has resolved this with a better design (use of conftest/fixtures). That being said, I hope you continue to contribute additional PRs in the future!

PR #4522, opened 2021-09-19
Poetry Install with `pip` unavailable shouldn't throw Exception (Fixes #4521)

### LINKED ISSUE #4521: 'install-poetry.py' throws 'ModuleNotFoundError' if 'pip' is not available.
- [x] I tried to install [latest](https://github.com/python-poetry/poetry/releases/latest) Poetry version.
- [x] There is no debug mode (`-vvv`) in `install-poetry.py`

- **OS version and name**: Debian 11 (Bullseye) Kernel: v5+
- **Poetry version**: v1.1.9
- **Link of a [Gist](https://gist.github.com/) with the contents of your pyproject.toml file**: _Not Required_

## Issue

The `install-poetry.py` fails after throwing Exception: `ModuleNotFoundError: No module named 'virtualenv'` when installing on Linux System with `pip3` present but `pip` not present.

This is not a desired behavior as this exception should be caught as `CalledProcessError` in the `Installer.install()` method.

I checked the code in `install_poetry.py`, the issue seems to be the use of `subprocess.call()` method (in `Installer.make_env()` method) instead of `subprocess.check_call()` or better instead `subprocess.run()` method (with `check=True` param).

The command invoked through `subprocess.call()` fails and the return code is non-zero too, however, it doesn't throw `CalledProcessError` exception, hence the `import virtualenv` fails later and raises `ModuleNotFoundError` exception.

sci-c0:
Fixes #4521

When installing Poetry through `install-poetry.py`, if `pip` is not
available, Exception should not be raised, rather installation should
be aborted with user-friendly error msg.

# Pull Request Check List

Resolves: #issue-number-here

### DIFF
--- a/install-poetry.py
+++ b/install-poetry.py
@@ -520,7 +520,7 @@ class Installer:
                 [sys.executable, "-m", "pip", "install", "virtualenv", "-t", tmp_dir],
                 stdout=subprocess.PIPE,
                 stderr=subprocess.STDOUT,
-                check=True
+                check=True,
             )
 
             sys.path.insert(0, tmp_dir)
--- a/install-poetry.py
+++ b/install-poetry.py
@@ -516,10 +516,11 @@ class Installer:
         env_path = self._data_dir.joinpath("venv")
 
         with temporary_directory() as tmp_dir:
-            subprocess.call(
+            subprocess.run(
                 [sys.executable, "-m", "pip", "install", "virtualenv", "-t", tmp_dir],
                 stdout=subprocess.PIPE,
                 stderr=subprocess.STDOUT,
+                check=True
             )
 
             sys.path.insert(0, tmp_dir)


finswimmer (approved):
LGTM :+1:

sci-c0:
@finswimmer I don't exactly understood the issue here as why would Linter fail...

As shown in the Linter diff, should I add an extra `,` at the end of args list in the function call?

Can you guide me here?

finswimmer:
> Can you guide me here?

Poetry provides `pre-commit` hooks for running the linter. Just run `pre-commit install` once followed by `pre-commit run --all-files` after changing code.

frjonsen:
Since this MR only seems to be hanging on a linter check and I just spent 20 minutes scratching my head as to why poetry wouldn't install, I'm writing this comment to hopefully get the authors attention. If you have a moment to fix this, it'd probably be helpful to more people.

neersighted (approved):
I've fixed the formatting issues -- this will get merged once the checks complete.

Issue #4523, opened 2021-09-20
Poetry refuses to install package with correct hash


MartinWallgren:
- [x] I am on the [latest](https://github.com/python-poetry/poetry/releases/latest) Poetry version.
- [x] I have searched the [issues](https://github.com/python-poetry/poetry/issues) of this repo and believe that this is not a duplicate.
- [x] If an exception occurs when executing a command, I executed it again in debug mode (`-vvv` option).

- **OS version and name**: Debian Buster
- **Poetry version**: 1.1.9
- **Link of a [Gist](https://gist.github.com/) with the contents of your pyproject.toml file**: https://gist.github.com/MartinWallgren/65ffceae6b597698602afdcdae927d7f

## Issue

Poetry refuses to install a package even though the checksum is correct. Looking at the output it seems as if the cheksum stored in the lock file is md5 and the checksum used during installation is sha256. 

Both sha256:3ae5020d5eddabcb57db9211e3f1a46ebafa28cb31cdeb4a497189041757bb7b and md5:75dbe554e7838a35e3a5836887cf9efc are valid checksums for this package according to our index (artifactory).

```
❯ poetry install
Installing dependencies from lock file

Package operations: 1 install, 0 updates, 0 removals

  • Installing axis-json-log-formatter (0.1.0): Failed

  RuntimeError

  Retrieved digest for link axis_json_log_formatter-0.1.0.tar.gz(sha256:3ae5020d5eddabcb57db9211e3f1a46ebafa28cb31cdeb4a497189041757bb7b) not in poetry.lock metadata ['md5:75dbe554e7838a35e3a5836887cf9efc']

  at ~/.poetry/lib/poetry/installation/chooser.py:115 in _get_links
      111│
      112│         if links and not selected_links:
      113│             raise RuntimeError(
      114│                 "Retrieved digest for link {}({}) not in poetry.lock metadata {}".format(
    → 115│                     link.filename, h, hashes
      116│                 )
      117│             )
      118│
      119│         return selected_links
```

MartinWallgren:
I have tried to re-create the lock file without any difference in behavior.

danygielow:
This seems to be a bug in poetry-core 1.0.5. With poetry-core 1.0.4 packages with md5 hashes are installed normally.

orShap:
looks like when trying to install poetry 1.1.8 in Dockerfile (with poetry-core 1.0.4 locked in poetry.lock file) 
[...]

PR #4525, opened 2021-09-20
feat(installer): check all hash digest of package

hydra-zim:
Resolves: https://github.com/python-poetry/poetry/issues/4523

### DIFF
--- a/poetry/installation/executor.py
+++ b/poetry/installation/executor.py
@@ -671,19 +671,19 @@ class Executor:
                 archive = self._chef.prepare(archive)
 
         if package.files:
-            file_dependency = FileDependency(
+            file_dep = FileDependency(
                 package.name,
                 Path(archive.path) if isinstance(archive, Link) else archive,
             )
 
             for file in package.files:
-                hash_name, expected_digest = file['hash'].split(":")
-                if file_dependency.hash(hash_name) != expected_digest:
+                hash_name, expected_digest = file["hash"].split(":")
+                if file_dep.hash(hash_name) != expected_digest:
                     raise RuntimeError(
                         f"Invalid hash for {package} using archive {archive.name}"
                     )
 
-            self._hashes[package.name] = file_dependency.hash()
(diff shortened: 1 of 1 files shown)


poetry/installation/executor.py
```diff
@@ -671,19 +671,19 @@ def _download_link(self, operation: Union[Install, Update], link: Link) -> Link:
                 archive = self._chef.prepare(archive)
 
         if package.files:
-            archive_hash = (
-                "sha256:"
-                + FileDependency(
-                    package.name,
-                    Path(archive.path) if isinstance(archive, Link) else archive,
-                ).hash()
+            file_dependency = FileDependency(
```
awilkins, line 674: Maybe rename this to `file_dep` or even `dep` ; there is a module called `file_dependency` and I accidentally auto-imported it in my IDE when typing a similar name.

poetry/installation/executor.py
```diff
@@ -671,19 +671,19 @@ def _download_link(self, operation: Union[Install, Update], link: Link) -> Link:
...
-                raise RuntimeError(
-                    f"Invalid hash for {package} using archive {archive.name}"
-                )
 
-            self._hashes[package.name] = archive_hash
+            for file in package.files:
+                hash_name, expected_digest = file['hash'].split(":")
+                if file_dependency.hash(hash_name) != expected_digest:
+                    raise RuntimeError(
+                        f"Invalid hash for {package} using archive {archive.name}"
```
awilkins, line 683: Maybe mention which hash type is invalid here.

Also, the linter is moaning about the use of single-quotes.

awilkins (changes requested):
General comments :

- I'm not a maintainer (sadly), but I do have an interest in seeing this fixed quickly
- To that end, I'd prefer this PR to target the 1.1 release branch and not `master`
- I feel like some tests would expedite this PR ; I'll look at that as soon as I have the time, but looks like what needs to happen is some mods to `MockRepository` to override `find_links_for_package`.

PR #4529, opened 2021-09-20
[1.1] Ensure all hash types are checked

sdispater:
## Pull Request Check List

Resolves: #4523

- [x] Added **tests** for changed code.
- [ ] Updated **documentation** for changed code.

PR #4534, opened 2021-09-21
Validate hash regardless of type

awilkins:
# Pull Request Check List

Resolves: #4523

Also see : #4529 which does the same thing.

**NB : this requires a higher version of `poetry-core`, which I presume will be `1.0.6`**

Lockfile has not been updated - because this has not been released yet. Have linked to git for time being.

---

- Check hash regardless of which type has been used


awilkins:
Superceded in release 1.1.10

PR #4573, opened 2021-09-29
Improves hash mismatch exception language

colindean:
# Pull Request Check List

I can't get tests to run locally off of master for some reason, so I didn't run them here. There don't seem to be any tests for the particular method from which I extracted this new method. I'm down for writing some tests for the new method if I could get `make test` working, or even `poetry install` for that matter. I might need to just blow away my poetry working directory and recreate it.

I think this area could be further improved by representing the hashes as a `dataclass` or something containing the hash type and the string digest. The types of hashes would be more robust then instead of hardcoding sha256 here and in `FileDependency.hash()`. It would also enable enumeration of hash types of known hashes so that `FileDependency.hash()` could return a list of hashes based on requested types. Doing this might alleviate the md5/sha256 problems arising from changes in poetry-core 1.0.6+ that affect Poetry 1.1.7+ users using Artifactory.

### DIFF
--- a/poetry/installation/executor.py
+++ b/poetry/installation/executor.py
@@ -678,9 +678,10 @@ class Executor:
                     Path(archive.path) if isinstance(archive, Link) else archive,
                 ).hash()
             )
-            if archive_hash not in {f["hash"] for f in package.files}:
+            known_hashes = {f["hash"] for f in package.files}
+            if archive_hash not in known_hashes:
                 raise RuntimeError(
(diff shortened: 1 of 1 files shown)


abn (dismissed):

neersighted (approved):
Looks like @abn's concerns are addressed. While it would be good to add tests here, this is just a straight refactor of some jumbled code, and I'd like to quickly land a follow-up PR with the ability to accept `md5` hashes. As such, we'll focus on adding tests to the improved version later.

PR #4574, opened 2021-09-30
Rename whitelist_externals in FAQ

wwuck:
`whitelist_externals` is now deprecated in tox and has been renamed to `allowlist_externals`

https://tox.readthedocs.io/en/latest/config.html#conf-allowlist_externals

### DIFF
--- a/docs/faq.md
+++ b/docs/faq.md
@@ -65,7 +65,7 @@ isolated_build = true
 envlist = py27, py36
 
 [testenv]
-whitelist_externals = poetry
+allowlist_externals = poetry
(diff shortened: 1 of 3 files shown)


Mariatta:
I think the minimum `tox` version should be bumped to >=3.18 in https://github.com/python-poetry/poetry/blob/82459bd6f38785882a8eff0fbce2a8b0119b205c/pyproject.toml#L59

neersighted (approved):
LGTM! Thanks for bumping the version and fixing our config.

PR #4599, opened 2021-10-06
Exclude .venv from Time Machine backups

probablykasper:

By default `.venv` folders are in `~/Library/Caches`, but they can also be stored in project folders using the `virtualenvs.in-project` config option. When this happens, cache is being stored in a non-standard location, so this PR makes sure backup solutions like Time Machine know not to back it up.

Adds an `com.apple.metadata:com_apple_backup_excludeItem` xattr for `.venv` folders on macOS. This means `.venv` folders won't be included in Time Machine backups (and other backup solutions respecting this attribute).

This is the same as what Rust's package manager does for their `target` folder.


tests/utils/test_env.py
```diff
@@ -77,6 +78,20 @@ def test_virtualenvs_with_spaces_in_their_path_work_as_expected(tmp_dir, manager
     assert venv.run("python", "-V", shell=True).startswith("Python")
 
 
+def test_venv_backup_exclusion(tmp_dir, manager):
```
Jc2k, line 81: What about adding:

```python
@pytest.mark.skipif(sys.platform == "darwin", reason="requires darwin")
```

Then you can get rid of the if on line 86.

poetry/utils/env.py
```diff
@@ -1002,7 +1003,20 @@ def build_venv(
 
         args.append(str(path))
 
-        return virtualenv.cli_run(args)
+        cli_result = virtualenv.cli_run(args)
+
+        print("path", str(path))
```
Jc2k, line 1008: Did you mean to leave these prints in? I'm guessing they are leftovers from debugging?

poetry/utils/env.py
```diff
@@ -1002,7 +1003,20 @@ def build_venv(
...
+        print("args", args)
+
+        # Exclude the venv folder from from macOS Time Machine backups
+        if sys.platform == "darwin":
+            xattr.setxattr(
+                str(path),
+                "com.apple.metadata:com_apple_backup_excludeItem",
+                b"bplist00_\x10\x11com.apple.backupd\x08\x00\x00\x00\x00\x00\x00\x01\x01\x00\x00\x00\x00\x00\x00\x00\x01\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x1c",
```
Jc2k, line 1016: Would it be less intimidating to use plistlib here? E.g.

```
            xattr.setxattr(
                str(path),
                "com.apple.metadata:com_apple_backup_excludeItem",
                plistlib.dumps('com.apple.backupd', fmt=plistlib.FMT_BINARY),
            )
```

I don't have a strong opinion on this, apart from it's clearer to anyone working on the code that it is just a weird encoding of the string `com.apple.backupd`. Otherwise given the length of the hex you might think it is setting a bunch of undocumented backup related settings.

(But leave the test alone to verify it's doing the right thing).

Jc2k (commented):
I'm just another user, but would like this feature too so thought i'd give it a quick review.

Do you have any references for where `com.apple.metadata:com_apple_backup_excludeItem` comes from and how stable it is likely to be? Looking at cargo [here](https://github.com/rust-lang/cargo/pull/4386/files) it uses `CFURLSetResourcePropertyForKey` rather than xattr directly. I agree xattr is the way to go, especially as the objc libs for python are huge, but a few links to other projects relying on setting or checking the xattr directly might help other reviewers to verify that the change is correct.

tests/utils/test_env.py
probablykasper: Good call

poetry/utils/env.py
probablykasper: I don't think it's worth adding another dependency for that. I went and added a comment explaining it instead, hopefully that'll do.

probablykasper: Aaah, in that case I'll go ahead and change it

probablykasper:
@finswimmer I disagree with that. I strongly disagree with that. By default Poetry stores `.venv` in `~/Library/Caches`, which is indeed ignored by any sensible backup tool. But Poetry also lets you store `.venv` in project folders (Forgot this isn't default, should've explained when I opened the PR).

When tools put cache in a non-standard location, they should make sure it's treated correctly, like Rust does. I don't think it's reasonable to put the burden on backup solutions to figure out how every package manager out there works.

neersighted:
> 
> 
> In addition to cargo (Rust), it looks like npm (Node.js) plans to implement this for `node_modules` in npm 9.
> 
> @finswimmer Would love to know what you think, and what others think if you're still skeptical about it. Thanks!

I'm generally with @finswimmer on this one -- it seems inelegant and a bit out of scope. I'd love to see links to implementations in other language's package managers to get an idea of what the wider world thinks of this.

probablykasper:
@neersighted Could you please elaborate on how it's inelegant and out of scope?

When you store cache outside of a cache folder, I see it as a responsibility to make sure it's properly treated as cache. In this case, there's luckily a conventional way to go about doing this.

The only alternative I see is to let all the backup services out there learn how every single package manager/tool/etc stores cache, and risk potentially ignoring folders they shouldn't. That, I definitely don't think is elegant.

neersighted:
My rationale is that this is historically out of scope, and that backup tools generally do not hoover up dotfiles. Obviously times change and my sensibilities are probably rooted a bit in Unix neckbeard-ism.

This seems more involved/elaborate than I would like, but your mention of other projects is a point in favor of the change. I'm not necessarily opposed -- but I'd like to see the consensus from similar projects to help inform my judgement.

probablykasper:
> backup tools generally do not hoover up dotfiles

Didn't think about that, but it depends on the purpose of the backup tool. If you have a full or bootable backup, you definitely want to include dotfiles since lots of tools use dotfiles for configs (git, npm, docker etc).

> This seems more involved/elaborate than I would like

I don't really think so, unless I'm misunderstanding you. It's a simple folder xattr, with a basic plist-encoded string value.

You can see Rust's implementation [here](https://github.com/rust-lang/cargo/blob/e165bc881e47c8a687e8cd08e55324f4e9cc47d7/crates/cargo-util/src/paths.rs#L616) (that function is called [here](https://github.com/rust-lang/cargo/blob/216f915c46b8ada2323423d049314ba18247ef95/src/cargo/core/compiler/layout.rs#L156)). They're directly calling the macOS Obj-C API which adds this.
Original issue requesting this: https://github.com/rust-lang/cargo/issues/3884
2019 issue requesting it to be fixed due to regression: https://github.com/rust-lang/cargo/issues/7189

poetry/utils/env.py
```diff
@@ -1002,7 +1003,19 @@ def build_venv(
 
         args.append(str(path))
 
-        return virtualenv.cli_run(args)
+        cli_result = virtualenv.cli_run(args)
+
+        # Exclude the venv folder from from macOS Time Machine backups
+        if sys.platform == "darwin":
```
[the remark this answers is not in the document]
probablykasper: I don't think I'll tackle that in this PR, especially when there's already resistance to merging it. To be fair, even though other platforms are more common, Time Machine might still be the most common full+incremental backup solution

neersighted (commented):
Can we add a TODO comment to add some backup-ignore marker for other platforms, and then rebase this?

probablykasper:
Sure thing, good now @neersighted?

src/poetry/utils/env.py
```diff
@@ -1087,7 +1088,20 @@ def build_venv(
 
         args.append(str(path))
 
-        return virtualenv.cli_run(args)
+        cli_result = virtualenv.cli_run(args)
+
+        # Exclude the venv folder from from macOS Time Machine backups
+        # TODO: Add backup-ignore markers for other platforms too
+        if sys.platform == "darwin":
+            import xattr  # type: ignore[import]
```
neersighted, line 1096: Instead of adding a `# type: ignore` comment let's add this to the Mypy settings in pyproject.toml

neersighted (changes requested):

pyproject.toml
```diff
@@ -66,6 +66,7 @@ shellingham = "^1.1"
 tomlkit = ">=0.7.0,<1.0.0"
 # exclude 20.4.5 - 20.4.6 due to https://github.com/pypa/pip/issues/9953
 virtualenv = "(>=20.4.3,<20.4.5 || >=20.4.7)"
+xattr = {version = "^0.9.7", markers = "sys_platform == 'darwin'"}
```
neersighted, line 69:
```suggestion
xattr = { version = "^0.9.7", markers = "sys_platform == 'darwin'" }
```

src/poetry/utils/env.py
probablykasper: Cool, added it there instead :)

neersighted (approved):

Issue #4612, opened 2021-10-07
`poetry lock --no-update` upgrades `coverage` version


Parnassius:
- [X] I am on the [latest](https://github.com/python-poetry/poetry/releases/latest) Poetry version.
- [X] I have searched the [issues](https://github.com/python-poetry/poetry/issues) of this repo and believe that this is not a duplicate.
- [X] If an exception occurs when executing a command, I executed it again in debug mode (`-vvv` option).

- **OS version and name**: Debian 11
- **Poetry version**: 1.11.1, 1.2.0a2, 34ca1657cbd2834d98e3bb99ed76fd2df2e6b3b2
- **Link of a [Gist](https://gist.github.com/) with the contents of your pyproject.toml file**: https://gist.github.com/Parnassius/1dc50ce25562b9a1bf779724d820b53d (contains both `pyproject.toml` and `poetry.lock`)

## Issue

Running `poetry lock --no-update` with the above `pyproject.toml` and `poetry.lock` upgrades `coverage` from 6.0 to 6.0.1. Below is the output of the command run with the `-vvv` option (with poetry 1.1.11):
<details>

```Resolving dependencies...
   1: fact: test-poetry-update is 0.1.0
   1: derived: test-poetry-update
   1: fact: test-poetry-update depends on pytest-cov (*)
   1: selecting test-poetry-update (0.1.0)
   1: derived: pytest-cov
   1: fact: pytest-cov (3.0.0) depends on pytest (>=4.6)
   1: fact: pytest-cov (3.0.0) depends on coverage (>=5.2.1)
   1: selecting pytest-cov (3.0.0)
   1: derived: coverage[toml] (>=5.2.1)
   1: derived: pytest (>=4.6)
PyPI: No release information found for coverage-2.5, skipping
PyPI: No release information found for coverage-2.6, skipping
PyPI: No release information found for coverage-2.75, skipping
PyPI: No release information found for coverage-2.76, skipping
PyPI: No release information found for coverage-2.77, skipping
PyPI: No release information found for coverage-2.78, skipping
PyPI: No release information found for coverage-2.8, skipping
PyPI: No release information found for coverage-2.80, skipping
PyPI: No release information found for coverage-2.85, skipping
PyPI: 7 packages found for coverage >=5.2.1
   1: fact: pytest (6.2.5) depends on attrs (>=19.2.0)
   1: fact: pytest (6.2.5) depends on iniconfig (*)
[...]

Issue #3128, opened 2020-10-07
`poetry update <single-pkg>` updates a different package with extras specification


immerrr:
- [x] I am on the [latest](https://github.com/python-poetry/poetry/releases/latest) Poetry version.
- [x] I have searched the [issues](https://github.com/python-poetry/poetry/issues) of this repo and believe that this is not a duplicate.
- [x] If an exception occurs when executing a command, I executed it again in debug mode (`-vvv` option).

- **OS version and name**: Ubuntu 16.04
- **Poetry version**: 1.1.2
- **Link of a [Gist](https://gist.github.com/) with the contents of your pyproject.toml file**: https://gist.github.com/immerrr/0d55d8958683d26a024153cade39b266

## Issue

It looks like packages with extras circumvent the whitelisting restriction when updating a single package with
```
poetry update <single-pkg>
```

Here's a repro:
- in the pyproject.toml linked above, change watchdog version to be strictly `0.10.2`, and do `poetry update`
- now, revert the watchdog version requirement to be `^0.10.2`, but this time do `poetry update decorator`
- expected behaviour: poetry does not change anything, as there is no new version of `decorator` package
- actual behaviour: 

<details>

<summary>
watchdog is updated to `0.10.3`
</summary>

```
$ poetry -V
Poetry version 1.1.2
$ poetry show --tree
decorator 4.4.2 Decorators for Humans
watchdog 0.10.2 Filesystem events monitoring
├── argh >=0.24.1
├── pathtools >=0.1.1
└── pyyaml >=3.10
$ poetry update decorator -vvv
Using virtualenv: /mnt/extraspace/virtualenvs/poetry-test-TzTtOsUh-py3.7
Updating dependencies
Resolving dependencies...
   1: fact: poetry-test is 0.1.0
   1: derived: poetry-test
   1: fact: poetry-test depends on watchdog[watchmedo] (^0.10.2)
   1: fact: poetry-test depends on decorator (^4.4.2)
   1: selecting poetry-test (0.1.0)
   1: derived: decorator (^4.4.2)
   1: derived: watchdog[watchmedo] (^0.10.2)
PyPI: No release information found for watchdog-0.3.4, skipping
PyPI: No release information found for watchdog-0.3.6, skipping
PyPI: No release information found for watchdog-0.6.1, skipping
PyPI: 2 packages found for watchdog >=0.10.2,<0.11.0
PyPI: No release information found for decorator-3.4.1, skipping
[...]

PR #4618, opened 2021-10-09
Fix for "poetry lock --no-update" so that dependencies with extras are not updated unnecessarily

radoering:
# Pull Request Check List

Resolves: #4612
Resolves: #3128
Resolves: #5185

When calling `poetry lock` with `--no-update` dependencies with extras should not be updated unnecessarily.

### DIFF
--- a/src/poetry/mixology/version_solver.py
+++ b/src/poetry/mixology/version_solver.py
@@ -378,8 +378,9 @@ class VersionSolver:
                             if version.version != locked.version:
                                 version = None
                             break
-            with suppress(IndexError):
-                version = version or packages[0]
+            if version is None:
+                with suppress(IndexError):
+                    version = packages[0]
 
             if version is None:
                 # If there are no versions that satisfy the constraint,
--- a/src/poetry/mixology/version_solver.py
+++ b/src/poetry/mixology/version_solver.py
@@ -366,21 +366,21 @@ class VersionSolver:
                 )
                 return dependency.complete_name
 
-            version = None
+            package = None
             if dependency.name not in self._use_latest:
                 # prefer locked version of compatible (not exact same) dependency;
                 # required in order to not unnecessarily update dependencies with
                 # extras, e.g. "coverage" vs. "coverage[toml]"
                 locked = self._locked.get(dependency.name, None)
             if locked is not None:
-                version = next(
+                package = next(
                     (p for p in packages if p.version == locked.version), None
                 )
-            if version is None:
+            if package is None:
                 with suppress(IndexError):
-                    version = packages[0]
+                    package = packages[0]
 
-            if version is None:
+            if package is None:
                 # If there are no versions that satisfy the constraint,
                 # add an incompatibility that indicates that.
                 self._add_incompatibility(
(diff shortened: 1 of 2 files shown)


radoering:
@neersighted: Shall I rebase the branch so that the FreeBSD test can succeed or should I wait for your review?

neersighted:
> 
> 
> @neersighted: Shall I rebase the branch so that the FreeBSD test can succeed or should I wait for your review?

Please go ahead and rebase! There are enough changes in master that significantly stale PRs will need to be rebased to have confidence in testing anyway. Thanks!

src/poetry/mixology/version_solver.py
```diff
@@ -374,10 +375,20 @@ def _get_min(dependency: Dependency) -> Tuple[bool, int]:
...
+                # extras, e.g. "coverage" vs. "coverage[toml]"
+                locked = self._locked.get(dependency.name, None)
+                if locked is not None:
+                    for version in packages:
+                        if version.version <= locked.version:
+                            if version.version != locked.version:
+                                version = None
+                            break
+            with suppress(IndexError):
+                version = version or packages[0]
```
neersighted, line 391: I feel like this line is a bit clunky -- do you think this is cleaner than a `None` check?

src/poetry/mixology/version_solver.py
```diff
@@ -374,10 +375,20 @@ def _get_min(dependency: Dependency) -> Tuple[bool, int]:
...
-                version = packages[0]
-            except IndexError:
-                version = None
+            version = None
+            if dependency.name not in self._use_latest:
+                # prefer locked version of compatible (not exact same) dependency;
+                # required in order to not unnecessarily update dependencies with
+                # extras, e.g. "coverage" vs. "coverage[toml]"
+                locked = self._locked.get(dependency.name, None)
+                if locked is not None:
```
neersighted, line 384: There's not a good way to unroll the nested logic here, is there? This is a bit dense, but I'm not seeing an obvious simplification myself.

neersighted (changes requested):

src/poetry/mixology/version_solver.py
radoering: Probably, a None check is easier to understand. I will change it.

src/poetry/mixology/version_solver.py
radoering: If you want to refrain from taking advantage of the fact that the list is sorted and thus from the early return if the version is not found, you can simplify the loop as follows:

```
if locked is not None:
    for version in packages:
        if version.version == locked.version:
            break
```

neersighted: In that case I'm wondering if `next()` wouldn't be clearer -- e.g.

```python
    version = next((p.version for p in packages if p.version == locked.version), None)
```

Thoughts?

radoering: I like it, but it has to be:

`version = next((p for p in packages if p.version == locked.version), None)`

Maybe a bit confusing that the variable `version` is a `package`. Should we rename it to `package`?

radoering: @neersighted Sorry for bothering, but I had the impression that the review of this PR has already been well advanced - maybe, the naming of the variable being the only point to be clarified before merging. (Correct me if I'm wrong.) I'd appreciate if you find some minutes so that we may finalize it.

neersighted (approved):

PR #4625, opened 2021-10-12
install-poetry: prompt before deleting `data_dir`

br3ndonland:
The _install-poetry.py_ script supports an `--uninstall` option. The uninstall performs `shutil.rmtree(str(self._data_dir))`.

https://github.com/python-poetry/poetry/blob/82459bd6f38785882a8eff0fbce2a8b0119b205c/install-poetry.py#L479-L500

If `$POETRY_HOME` is set, `self._data_dir` is set to `$POETRY_HOME`.

https://github.com/python-poetry/poetry/blob/82459bd6f38785882a8eff0fbce2a8b0119b205c/install-poetry.py#L395

https://github.com/python-poetry/poetry/blob/82459bd6f38785882a8eff0fbce2a8b0119b205c/install-poetry.py#L134-L136

Uninstallation will therefore delete the entire `$POETRY_HOME` directory.

This PR will update _install-poetry.py_ to prompt the user before performing this destructive operation.

### DIFF
--- a/install-poetry.py
+++ b/install-poetry.py
@@ -484,6 +484,18 @@ class Installer:
 
             return 1
 
+        if not self._accept_all:
+            self._write(
+                "{}: Uninstalling Poetry will {} {}.".format(
+                    colorize("warning", "WARNING"),
+                    colorize("warning", "DELETE"),
+                    self._data_dir,
+                )
+            )
+            continue_uninstall = input("Do you want to continue? ([y]/n) ") or "y"
+            if continue_uninstall.lower() not in {"y", "yes"}:
+                return 1
+
         version = None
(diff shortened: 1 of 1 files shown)


neersighted:
I'm thinking this is out of scope for two reasons. The first is the issue of setting `$POETRY_HOME=~/.local` was never intended, and the `POETRY_HOME` directory was meant to be managed entirely by Poetry. The docs could likely be improved on this (e.g. if you put anything else in that directory, it may be destroyed), but this is not out of line with most developer tools.

Secondly, the installer script was recently migrated to https://github.com/python-poetry/install.python-poetry.org. I would suggest opening an issue/PR there if you still feel this change is important.

PR #4634, opened 2021-10-13
Remove the `doc` env from tox.ini

Mariatta:
In https://github.com/python-poetry/poetry/commit/bce4ac5d2c7e35c98d196e89f2325fdc11c5f6e9, the `doc` env was removed from tox.ini, and the doc was restructured differently. Since the `doc` env is not used anymore, I removed it from `tox.ini` file.

Relates-to: #4142

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -1,7 +1,7 @@
 [tox]
 minversion = 3.3.0
 isolated_build = True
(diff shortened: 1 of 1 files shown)


Mariatta:
This is for hacktoberfest. Thank you.

Issue #2153, opened 2020-03-08
ResourceWarning: unclosed socket


kbaikov:
- [x] I have searched the [issues](https://github.com/python-poetry/poetry/issues) of this repo and believe that this is not a duplicate.
- [x] I have searched the [documentation](https://python-poetry.org/docs/) and believe that my question is not covered.

## Issue

If i run python with warnings enabled (set the env var `PYTHONWARNINGS="default"` ), and try to do `poetry update` or any remote connection related operation i get the following message:
```bash
C:\Users\<username>\.poetry\lib\poetry\console\application.py:54: ResourceWarning: unclosed <ssl.SSLSocket fd=868, family=AddressFamily.AF_INET, type=SocketKind.SOCK_STREAM, proto=0, laddr=('157.163.199.78', 57798), raddr=('151.101.192.223', 443)>
  self._poetry = None
ResourceWarning: Enable tracemalloc to get the object allocation traceback
```

Poetry version 1.0.5
Python 3.8.2 x64
Windows 10

Update: added the fuller message with the location of the warning in code: `C:\Users\<username>\.poetry\lib\poetry\console\application.py:54:`

kbaikov:
Ping @stephsamson 
Any activity on this bug?

ShahriyarR:
The same warning I got while doing:

```
$ poetry install
Updating dependencies
Resolving dependencies... (0.4s)/home/shako/.poetry/lib/poetry/repositories/pypi_repository.py:207: ResourceWarning: unclosed <ssl.SSLSocket fd=3, family=AddressFamily.AF_INET, type=SocketKind.SOCK_STREAM, proto=6, laddr=('192.168.1.57', 50624), raddr=('151.101.64.223', 443)>
  data = self._get("pypi/{}/json".format(name))
ResourceWarning: Enable tracemalloc to get the object allocation traceback
```

kbaikov:
Works for me with Poetry (version 1.2.0a2) and Python 3.10

PR #4643, opened 2021-10-15
fix: "ResourceWarning: unclosed SSLSocket"

kkirsche:
Fixes: #2153

When running `poetry update`, during the download phase when `PYTHONWARNINGS="default"` environment variable is set, multiple different sessions trigger resource warnings. These are caused by leaving a session (aka not explicitly closing it) that uses keep alive behavior after a request has been issued. 

```
/Users/username/Library/Application Support/pypoetry/venv/lib/python3.10/site-packages/poetry/installation/executor.py:594: ResourceWarning: unclosed <ssl.SSLSocket fd=5, family=AddressFamily.AF_INET, type=SocketKind.SOCK_STREAM, proto=0, laddr=('REDACTED', 59025), raddr=('140.108.3.33', 80)>
```

This line references:
```
 593   │             try:
 594   │                 archive = self._download_archive(operation, link)
 595   │             except BaseException:
 596   │                 cache_directory = self._chef.get_cache_directory_for_link(link)
```

This also fixes:

```
/Users/username/Library/Application Support/pypoetry/venv/lib/python3.10/site-packages/poetry/repositories/pypi_repository.py:241: ResourceWarning: unclosed <ssl.SSLSocket fd=13, family=AddressFamily.AF_INET, type=SocketKind.SOCK_STREAM, proto=0, laddr=('REDACTED', 59721), raddr=('148.131.15.33', 80)>
  json_data = self._get("pypi/{}/{}/json".format(package.name, package.version))
ResourceWarning: Enable tracemalloc to get the object allocation traceback
```

The issue is being caused because the session is being left open when the class is later cleaned up. By adding a destructor method, we can close the session if it exists, preventing this error.

# Pull Request Check List

I have not added tests or documentation as this should be covered more broadly already. It's unclear to me what specifically should be tested in this instance as it occurs specifically at the class destruction time. I don't think calling `__del__` directly is a proper test of what is being fixed, as it's related to the warnings generated at the time the class instance destruction occurs rather than simply verifying that the session is closable.

### DIFF
--- a/src/poetry/repositories/pypi_repository.py
+++ b/src/poetry/repositories/pypi_repository.py
@@ -83,6 +83,9 @@ class PyPiRepository(RemoteRepository):
     def session(self) -> CacheControl:
         return self._session
 
+    def __del__(self) -> None:
+        self._session.close()
+
     def find_packages(self, dependency: Dependency) -> List[Package]:
         """
         Find packages on the remote server.
--- a/src/poetry/utils/authenticator.py
+++ b/src/poetry/utils/authenticator.py
@@ -50,6 +50,10 @@ class Authenticator:
 
         return self._session
 
+    def __del__(self) -> None:
+        if self._session is not None:
+            self._session.close()
+
     def request(self, method: str, url: str, **kwargs: Any) -> requests.Response:
         request = requests.Request(method, url)
         username, password = self.get_credentials_for_url(url)


neersighted (dismissed):

neersighted:
> 
> 
> Fix validated by issue creator in the issue.

You've got some merge conflicts -- LGTM once they're resolved.

neersighted (approved):

neersighted (approved):

PR #4647, opened 2021-10-17
removed a pitfall where the built-in name dir was being shadowed and …

NaelsonDouglas:
**The problem**
There was a built-in name shadowing pitfall on install-poetry.py where the name dir, which is a Python built-in method, was being used as a variable.;

If for any reason the method gets called within that context, it would cause a had to detect bug because dir(arg) would not be a method being called, but a variable being called as a method.

On top of this, the name itself was not being used for any action

Solution
Removed the name using underscore notation

### DIFF
--- a/install-poetry.py
+++ b/install-poetry.py
@@ -178,7 +178,7 @@ def _get_win_folder_from_registry(csidl_name):
         _winreg.HKEY_CURRENT_USER,
         r"Software\Microsoft\Windows\CurrentVersion\Explorer\Shell Folders",
     )
-    dir, type = _winreg.QueryValueEx(key, shell_folder_name)
+    dir, _ = _winreg.QueryValueEx(key, shell_folder_name)
 
     return dir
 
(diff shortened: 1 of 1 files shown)


install-poetry.py
```diff
@@ -178,7 +178,7 @@ def _get_win_folder_from_registry(csidl_name):
         _winreg.HKEY_CURRENT_USER,
         r"Software\Microsoft\Windows\CurrentVersion\Explorer\Shell Folders",
     )
-    dir, type = _winreg.QueryValueEx(key, shell_folder_name)
+    dir, _ = _winreg.QueryValueEx(key, shell_folder_name)
```
chomechome, line 181: PR description and the change are contradictory. You removed the shadowed `type` built-on function, but `dir` is still there.
```suggestion
    directory, _ = _winreg.QueryValueEx(key, shell_folder_name)
    return directory
```

neersighted:
Please submit a PR to https://github.com/python-poetry/install.python-poetry.org. I'm not sure when/if we'll migrate changes back to this repo, but that is now the canonical location of the installer script.

PR #4666, opened 2021-10-21
ci: add install-poetry script workflow

abn:


neersighted (changes requested):
I think the concern I have with this approach is the sheer number of extra checks this adds. What if we only ran the full Matrix on Ubuntu (all Python versions and arguments), and instead just did a basic install test on macOS and Windows?

neersighted:
Okay, I think that's fair. I guess one of my questions is if the Github UI will be cluttered up by a million skipped checks -- if not, I think we should merge this as a first step and then think about how to go about breaking the install script out into its own repo without breaking existing automation that relies on the current location.

neersighted (approved):
Looks good in light of the filtering based on paths!

Issue #4615, opened 2021-10-08
`poetry env` doesn't work on Windows after install-poetry.py


tom-bowles:
- [x] I am on the [latest](https://github.com/python-poetry/poetry/releases/latest) Poetry version.
- [x] I have searched the [issues](https://github.com/python-poetry/poetry/issues) of this repo and believe that this is not a duplicate.
- [x] If an exception occurs when executing a command, I executed it again in debug mode (`-vvv` option).

- **OS version and name**: Windows 10
- **Poetry version**: 1.1.11
- **Link of a [Gist](https://gist.github.com/) with the contents of your pyproject.toml file**: N/A

## Issue

After installing poetry, creating a fresh project and running `poetry env list` the following error occurs:

    EnvCommandError

    Command python -W ignore - errored with the following return code 1, and output:
    'python' is not recognized as an internal or external command,
    operable program or batch file.
    input was : import sys

    if hasattr(sys, "real_prefix"):
        print(sys.real_prefix)
    elif hasattr(sys, "base_prefix"):
        print(sys.base_prefix)
    else:
        print(sys.prefix)

    at c:\...\poetry\venv\lib\site-packages\poetry\utils\env.py:1183 in _run
        1179│                 output = subprocess.check_output(
        1180│                     cmd, stderr=subprocess.STDOUT, **kwargs
        1181│                 )
        1182│         except CalledProcessError as e:
      → 1183│             raise EnvCommandError(e, input=input_)
        1184│
        1185│         return decode(output)
        1186│
        1187│     def execute(self, bin, *args, **kwargs):

This appears to be due to the following at lines 1233-1236 in https://github.com/python-poetry/poetry/blob/1.1/poetry/utils/env.py:

                if not bin.endswith(".exe"):
                    bin_path = self._bin_dir / (bin + ".exe")
                else:
                    bin_path = self._path / bin

I believe in the first case self._path should be used instead of self._bin_dir.

This problem doesn't appear to happen (the function is never called) when poetry is installed with `get-poetry.py` instead of `install-poetry.py`. I don't know why that is, but it's a good workaround. I think the above code is wrong nonetheless.

Luttik:
@tom-bowles great research. Poetry 1.1.8 works for me too, the latest version still seems to be broken.

zndr27:
Any timeline on when this will be resolved?

clintonroy:
Have you tried installing poetry with pipx at all, does that get around this issue?

[...]

PR #4682, opened 2021-10-28
fix: `poetry env` broken on Windows (#4615)

tom-bowles:
Resolves: #4615

Looks like this if statement was copied from further up in the function and only one of the branches updated to use self._path.

### DIFF
--- a/tests/utils/test_env.py
+++ b/tests/utils/test_env.py
@@ -1001,6 +1001,27 @@ def test_env_finds_the_correct_executables(tmp_dir, manager):
     assert Path(venv.python).name == expected_executable
     assert Path(venv.pip).name.startswith(expected_pip_executable.split(".")[0])
 
+@pytest.mark.skipif(os.name != "nt", reason="Testing functionality that is only relevant on Windows")
+def test_env_finds_python_executable_in_base_path_on_windows(tmp_dir, manager):
+    venv_path = Path(tmp_dir) / "Virtual Env"
+    manager.build_venv(str(venv_path), with_pip=True)
+    venv = VirtualEnv(venv_path)
+
+    executable = "python.exe"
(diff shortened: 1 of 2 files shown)


abn:
@tom-bowles can you rebase this so the cirrus tests can have a rerun please?

tom-bowles:
@abn - Thanks for taking a look! I've rebased now.

abn:
@tom-bowles should have mentioned this earlier, any cance we can add in a test case for this so we do not regress?

tom-bowles:
@abn - no worries, I'll have a crack at that today

tom-bowles:
@abn - Added test.

neersighted (changes requested):
LGTM -- we just need the formatting fixed (run `pre-commit run --all-files`) by black to merge.

tests/utils/test_env.py
```diff
@@ -1001,6 +1001,27 @@ def test_env_finds_the_correct_executables(tmp_dir, manager):
...
+
+    if bin_dir_path.exists():
+        if base_dir_path.exists():
+            bin_dir_path.unlink()
+        else:
+            bin_dir_path.rename(base_dir_path)
+
+    venv = VirtualEnv(venv_path)
+
+    assert Path(venv.python) == base_dir_path
```
neersighted, line 1023: Looks like this test fails on Python 3.6 on Windows, so I spoke a little too soon. Can we take a look at that failure and get it corrected?

dimbleby:
Looks like this was fixed incidentally at #5007.

Apparently that one was allowed through without testcases so I suppose it could _just about_ be worth keeping this open if anyone is motivated, more than a year later, to add a unit test.

But I expect this should now be closed.

PR #3512, opened 2020-12-24
Fix dropped markers on dep walk

johnmacnamararseg:
# Pull Request Check List

Resolves: #3511

- [x] Added **tests** for changed code.

It looks like the line `if key not in nested_dependencies:` was added to solve the issue of circular dependencies but this introduced the issue at hand because the `key` in this routine is not a unique key when a package has a dependency with multiple markers. Under conditions outlined in issue #3511 , this can result in markers not correctly getting assigned to lower depth packages.

I'd be happy to describe this issue and implementation choices in more detail if needed

PR #4686, opened 2021-10-28
Fix dropped markers in dependency walk

dimbleby:
# Pull Request Check List

Resolves: #3511

I would very much like a fix for #3511, and tonight for the first time I tried to understand the proposed fix at #3512.  I must admit I found it all a bit confusing!  Here is a version that makes a _much_ more localised change - my hope is that this is more palatable for merging.

Again the aim is to make sure that when we encounter an already-encountered dependency, we do not drop the associated markers.  But I've done this just by moving where we make the `if key not in nested_dependencies` check.

It's definitely possible that I've missed some good reason why #3512 is more complicated than this; but I have borrowed the unit test case added at #3512 and that still passes (and fails without the fix).  And I've also tested the fix in a couple of real life projects where I have run into this bug, and it does the job for those too.

@johnmacnamararseg @abn

### DIFF
--- a/poetry/packages/locker.py
+++ b/poetry/packages/locker.py
@@ -248,16 +248,15 @@ class Locker:
                         locked_package.to_dependency().constraint
                     )
 
-                if key not in nested_dependencies:
-                    for require in locked_package.requires:
-                        if require.marker.is_empty():
-                            require.marker = requirement.marker
-                        else:
-                            require.marker = require.marker.intersect(
-                                requirement.marker
-                            )
+                for require in locked_package.requires:
+                    if require.marker.is_empty():
+                        require.marker = requirement.marker
+                    else:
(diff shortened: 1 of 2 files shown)


neersighted (approved):
Thanks for the excellent work here -- this is a very minimal and easy to understand changeset. Thanks @johnmacnamararseg for the excellent test case as well.

neersighted:
Shoot, I forgot to add a Co-authored-by line to that commit, as the test was from #3512. Sorry @johnmacnamararseg.

polys:
Any updates on a release that will include this fix?

johnmacnamararseg:
@dimbleby thanks for the improvement and helping get this across the finish line. much easier to follow than what I originally had. very happy to see this merged in!

Issue #4670, opened 2021-10-23
Dependency resolution takes hours for multiple python version markers


rootcss:
- [x] I am on the [latest](https://github.com/python-poetry/poetry/releases/latest) Poetry version.
- [x] I have searched the [issues](https://github.com/python-poetry/poetry/issues) of this repo and believe that this is not a duplicate.
- [x] If an exception occurs when executing a command, I executed it again in debug mode (`-vvv` option).

- **OS version and name**: macOS Big Sur 11.4
- **Poetry version**: 1.1.11

## Issue
My project requires to use multiple python versions. Some of the packages are not compatible for both of those python versions, for which I'm using [multiple-constraints-dependencies](https://python-poetry.org/docs/dependency-specification/#multiple-constraints-dependencies) as shown below. Without this, the `poetry lock --no-update -vvv` command took just 10 seconds. However, after adding this it takes forever to run the dependency resolution. Below is the `pyproject.toml` file.

```
[tool.poetry]
name = "some-service"
version = "0.1.0"

[build-system]
requires = ["poetry-core>=1.0.0"]
build-backend = "poetry.core.masonry.api"

[tool.poetry.dependencies]
python = "3.6.10 || 3.9.5"
numpy = "1.19.3"
dropbox = [
    {version = "7.1.0", python = "3.6.10"},
    {version = "9.5.0", python = "3.9.5"}
]
gevent = [
    {version = "1.2.2", python = "3.6.10"},
    {version = "21.1.2", python = "3.9.5"}
]
guppy3 = [
    {version = "3.0.9", python = "3.6.10"},
    {version = "3.1.1", python = "3.9.5"}
]
Markdown = [
    {version = "2.6.6", python = "3.6.10"},
    {version = "3.3.4", python = "3.9.5"}
]
pexpect = [
    {version = "4.1.0", python = "3.6.10"},
    {version = "4.8.0", python = "3.9.5"}
]
scikit-learn = [
    {version = "0.20.4", python = "3.6.10"},
    {version = "0.24.2", python = "3.9.5"}
]
scipy = [
    {version = "1.5.4", python = "3.6.10"},
    {version = "1.7.1", python = "3.9.5"}
]
psycopg2 = [
    {version = "2.7.3.2", python = "3.6.10"},
    {version = "2.9.1", python = "3.9.5"}
]

[tool.poetry.dev-dependencies]
coverage = "4.4.1"
```

[...]

Issue #3367, opened 2020-11-15
Resolution fails for boto3/botocore in multiple python versions


mcouthon:
- [x] I am on the [latest](https://github.com/python-poetry/poetry/releases/latest) Poetry version.
- [x] I have searched the [issues](https://github.com/python-poetry/poetry/issues) of this repo and believe that this is not a duplicate.
- [x] If an exception occurs when executing a command, I executed it again in debug mode (`-vvv` option).

- **OS version and name**: macOS Catalina 10.15.7
- **Poetry version**: Poetry version 1.1.4
- **Link of a [Gist](https://gist.github.com/) with the contents of your pyproject.toml file**: [link](https://gist.github.com/mcouthon/2a64cb6ae101cccb21fd9e80f3336ad8)

## Issue

We need very specific `boto3/botocore` versions for compatibility with legacy code. In our non-poetry projects we set the following in the `requirements.txt`:

```
botocore==1.12.192;python_version<"3.0"
botocore==1.17.63;python_version>"3.0"
boto3==1.9.192;python_version<"3.0"
boto3==1.14.63;python_version>"3.0"
```

And installation works great in both python 2.7 and python 3.x.

I've attempted to create a comparable `pyproject.toml`:

```toml
[tool.poetry.dependencies]
python = "^2.7 || ^3.6"

botocore = [
    {version = "1.12.192", python = "^2"},
    {version = "1.17.63", python = "^3"}
]
boto3 = [
    {version = "1.9.192", python = "^2"},
    {version = "1.14.63", python = "^3"}
]
```

But `poetry lock` results in an error (on both python versions). Here's the error in a python 3.8 env:

```
   1: conflict: boto3 (1.14.63) depends on botocore (>=1.17.63,<1.18.0)
   1: ! not botocore (>=1.17.63,<1.18.0) is satisfied by botocore (1.12.192)
   1: ! which is caused by "test depends on botocore (1.12.192)"
   1: ! thus: boto3 is forbidden
   1: ! boto3 (1.14.63) is satisfied by boto3 (1.14.63)
   1: ! which is caused by "test depends on boto3 (1.14.63)"
   1: ! thus: version solving failed
```

And while the resolution starts with:

```
   0: Retrying dependency resolution with the following overrides ({Package('test', '0.1.0'): {'botocore': <Dependency botocore (1.12.192)>, 'boto3': <Dependency boto3 (1.9.192)>}}).
```

It at some point switches to:

```
[...]

PR #4695, opened 2021-11-01
Ignore combinations of overrides whose intersection of markers is empty

radoering:
# Pull Request Check List

Resolves: #3367
Resolves: #4670
Resolves: #4870
Resolves: #4947

Currently, poetry tries all combinations for multiple-constraints-dependencies, even if markers exclude each other, which means that the concerning combination is not relevant. As a consequence, dependency resolution may fail even if there is a solution as in #3367 or dependency resolution will take unnecessarily long as in #4670.

Instead of considering all combinations, only combinations whose intersection of markers is not empty should be considered in the overrides.

In the following example, it is clear that only two of the four combinations are relevant (one with both dependencies for 3.6 and one with both dependencies for 3.7). Mixed combinations are not sensible.

```
lib1= [
    {version = "1.0", python = "~3.6"},
    {version = "2.0", python = "~3.7"}
]
lib2 = [
    {version = "1.0", python = "~3.6"},
    {version = "2.0", python = "~3.7"},
]
```
However, there is one special case that has to be considered. Let's extend the previous example as follows:

```
lib1= [
    {version = "1.0", python = "~3.6"},
    {version = "2.0", python = "~3.7"}
]
lib2 = [
    {version = "1.0", python = "~3.6"},
    {version = "2.0", python = "~3.7"},
    {version = "3.0", python = "~3.8"}
]
```

Now, only considering the combinations whose intersection of markers is not empty is not enough because the constraint for lib2 for python 3.8 would be ignored. Thus, when determining the overrides, a dummy override with the inversion of the union of all markers of a dependency has to be added if there is no any marker dependency. In the example above, an empty constraint for lib1 with markers `not (python = "~3.6" or python = "~3.7")` (inversion of union of all markers of lib1) has to be added to the overrides. Thus, a solution for lib2 for python 3.8 can be found, not considering lib1 at all.

I am aware that this solution heavily relies on poetry's ability to recognize if the intersection of markers is empty and that poetry probably will fail in this task for more complex markers. However, I think that even in that case the drawback of this solution are only some additional (unnecessary but harmless) dependency resolutions (for the dummy overrides).

PR python-poetry/poetry-core#226 improves poetry's ability to recognize if the intersection of markers is empty.

### DIFF
--- a/src/poetry/puzzle/provider.py
+++ b/src/poetry/puzzle/provider.py
@@ -663,8 +663,9 @@ class Provider:
                         dep_other.set_constraint(
                             dep_other.constraint.intersect(dep_any.constraint)
                         )
-            else:
-                # if there is no any marker dependency,
+            elif not inverted_marker.is_empty():
+                # if there is no any marker dependency
+                # and the inverted marker is not empty,
                 # a dependency with the inverted union of all markers is required
                 # in order to not miss other dependencies later, for instance:
                 #   - foo (1.0) ; python == 3.7
@@ -680,10 +681,11 @@ class Provider:
 
             overrides = []
             overrides_marker_intersection = AnyMarker()
-            for _dep in self._overrides.get(package, {}).values():
-                overrides_marker_intersection = overrides_marker_intersection.intersect(
-                    _dep.marker
-                )
+            for dep_overrides in self._overrides.values():
+                for _dep in dep_overrides.values():
+                    overrides_marker_intersection = (
+                        overrides_marker_intersection.intersect(_dep.marker)
+                    )
             for _dep in _deps:
                 if not overrides_marker_intersection.intersect(_dep.marker).is_empty():
                     current_overrides = self._overrides.copy()
--- a/src/poetry/puzzle/provider.py
+++ b/src/poetry/puzzle/provider.py
@@ -16,8 +16,10 @@ from typing import Iterator
 
 from cleo.ui.progress_indicator import ProgressIndicator
 from poetry.core.packages.utils.utils import get_python_constraint_from_marker
+from poetry.core.semver.empty_constraint import EmptyConstraint
 from poetry.core.semver.version import Version
 from poetry.core.vcs.git import Git
+from poetry.core.version.markers import AnyMarker
 from poetry.core.version.markers import MarkerUnion
 
 from poetry.inspection.info import PackageInfo
@@ -70,7 +72,7 @@ class Provider:
         self._search_for: dict[Dependency, list[Package]] = {}
         self._is_debugging = self._io.is_debug() or self._io.is_very_verbose()
         self._in_progress = False
-        self._overrides: dict = {}
+        self._overrides: dict[DependencyPackage, dict[str, Dependency]] = {}
         self._deferred_cache: dict[Dependency, Package] = {}
         self._load_deferred = True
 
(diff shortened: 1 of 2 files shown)


maciejskorski:
I am watching this hopping poetry gets improved on markers!

However, let me share my concern regarding this specific PR.

On the file (real use-case)
```
[tool.poetry]
name = "test"
version = "0.1.0"
description = "demonstrates that poetry triggers overrides to easily (can be early pruned?)"
authors = ["Maciej Skorski <maciej.skorski@gmail.com>"]

[tool.poetry.dependencies]
python = "~3.8"
azureml-dataset-runtime = "1.*"
```
and
```
poetry install --dry-run -vv
```
I observe much more overrides and longer install time than in the release 1.2.0a2

radoering:
@maciejskorski 

In order to be able to ignore irrelevant combinations of overrides, additional overrides for implicit "package not explicitly required" dependencies have to be added first. Otherwise, dependency resolution will be erroneous in some cases (see my second example in https://github.com/python-poetry/poetry/pull/4695#issue-1041347543). Thus, the number of overrides can swing in both directions (less overrides if many of them can be ignored or more overrides if little of them can be ignored).
In some cases the number of overrides is reduced (e.g. in #4670 **from 510 overrides to 24**).
In other cases the number of overrides is increased (in your example **from 6 to 12**).

This PR heavily relies on the recognition of empty markers. With further improvement of marker intersection/union/inversion, the number of overrides could often be reduced. python-poetry/poetry-core#226 is just a first step in that direction. Nevertheless, I think this PR makes dependency resolution certainly more correct, probably more robust (see linked issues with failing resolution) and hopefully more often faster than slower.

I have analyzed your example. Its overrides originate in the following two multiple constraint dependencies:
- [azureml-dataset-runtime](https://pypi.org/pypi/azureml-dataset-runtime/1.37.0/json)
  `numpy (!=1.19.3); sys_platform == "linux"`
  `numpy (!=1.19.4); sys_platform == "win32"`
- [msal-extensions](https://pypi.org/pypi/msal-extensions/0.3.1/json)
  `portalocker (<2,>=1.0) ; python_version == "2.7" and platform_system != "Windows"`
  `portalocker (<2,>=1.6) ; python_version == "2.7" and platform_system == "Windows"`
  `portalocker (<3,>=1.0) ; python_version >= "3.5" and platform_system != "Windows"`
  `portalocker (<3,>=1.6) ; python_version >= "3.5" and platform_system == "Windows"`

**Without PR**

Poetry is already able to ignore two of the portalocker dependencies because their python versions don't fit the project's requirement. Thus, there are two groups of multiple constraint dependencies each containing two dependencies.
Without the PR, this results in `2 + 2 x 2 == 6` overrides.

**With PR**

The PR does not help in this case because there is no empty intersection of markers (assuming that the connection between `win32` and `Windows` is not considered by poetry at the moment).

In the first group, there is an implicit dependency that has to be considered: numpy is not explicitly required if `sys_platform != "linux" and sys_platform != "win32"`. Without the PR this implicit dependency is not considered and strictly speaking that is just wrong. (Neglecting that numpy may always be required and the requirements specification of azureml-dataset-runtime might just not consider that there is something else than `win32` and `linux`.)

In the second group, the implicit dependency is: portalocker is not explicitly required if `python < "3.5"` (Actually, it is `(python < "3.5" or platform_system == "Windows") and (python < "3.5" or platform_system != "Windows")` because that's the inversion of the union of all markers of the considered dependencies and poetry currently is not able to simplify this.) Thus, instead of two groups with two dependencies, we have two groups with three dependencies resulting in `3 + 3 x 3 == 12` overrides.

If poetry would be able to simplify the marker of the implicit portalocker dependency and build the intersection with the project's python restriction (`python = "~3.8"`), the marker of the implicit dependency would be empty so that the second group would still have only two dependencies resulting in `3 + 3 x 2 == 9` overrides. Further, if poetry would even be able to recognize that `sys_platform == "linux" and platform_system == "Windows"` as well as `sys_platform == "win32" and platform_system != "Windows"` as well as `sys_platform != "linux" and sys_platform != "win32" and platform_system == "Windows"` are empty markers,
3 combinations of overrides could be ignored resulting in only 6 overrides to consider.

maciejskorski:
@radoering 

thanks for that detailed clarification. I understand your patch may not help in this case, still increasing the number of costly operations is what worries me. Ideally I don't want the patch to be more expensive, or at least to have that communicated transparently. Per my understanding, you need to extend the logic to handle correctly some boundary cases. Do we really need that? Would they appear at al if the sub-dependencies were specified correctly? 

Following up on your thoughts on environmental markets, I have prepared a draft proposal (tests included) for early filtering packages. Instead of exponentially many combinations arising from splitting logic, we could have match less by early matching to the system/platform markers. Would be great to hear your opinion on this.

abn (approved):

Issue #4311, opened 2021-07-23
got a expected string or bytes-like object when running install


ElricleNecro:
- [X] I am on the [latest](https://github.com/python-poetry/poetry/releases/latest) Poetry version.
- [X] I have searched the [issues](https://github.com/python-poetry/poetry/issues) of this repo and believe that this is not a duplicate.
- [X] If an exception occurs when executing a command, I executed it again in debug mode (`-vvv` option).

- **OS version and name**: Arch Linux
- **Poetry version**: 1.1.7
- ```toml
  [tool.poetry]
  name = "testy"
  version = "0.1.0"
  description = ""
  authors = ["Test user"]

  [tool.poetry.dependencies]
  python = ">=3.7,<4.0"
  dgl = [
  	{ git = "https://github.com/dmlc/dgl.git", tag = "v0.7.0", markers = "sys_platform == 'linux'"},
  	{ version="^0.7", markers = "sys_platform == 'darwin'"}
  ]

  [build-system]
  requires = ["poetry-core>=1.0.0"]
  build-backend = "poetry.core.masonry.api"
  ```
## Issue
When using this `pyproject.toml`, I get the following exception:
```
  TypeError

  expected string or bytes-like object

  at /usr/lib/python3.9/site-packages/poetry/core/utils/helpers.py:27 in canonicalize_name
       23│ _canonicalize_regex = re.compile(r"[-_]+")
       24│ 
       25│ 
       26│ def canonicalize_name(name):  # type: (str) -> str
    →  27│     return _canonicalize_regex.sub("-", name).lower()
       28│ 
       29│ 
       30│ def module_name(name):  # type: (str) -> str
       31│     return canonicalize_name(name).replace(".", "_").replace("-", "_")
```

aalok-sathe:
Indeed, the bug seems to be stemming from [this line](https://github.com/python-poetry/poetry/blob/10d555984485088cf5d55979c1b235286de8e456/poetry/repositories/installed_repository.py#L115).

In what circumstance is the name returned by distribution metadata not a string? When it is `None` for some reason?

aalok-sathe:
But yes, I am encountering this bug quite often, and arbitrarily, so it's pretty annoying.
Sometimes poetry works fine, but then suddenly it will start throwing this error out of nowhere.

abn:
Closing as a duplicate of #3628 as it has more information.

Issue #3628, opened 2021-02-01
"expected string or bytes-like object" on poetry install


wichert:
- [X] I am on the [latest](https://github.com/python-poetry/poetry/releases/latest) Poetry version.
- [X] I have searched the [issues](https://github.com/python-poetry/poetry/issues) of this repo and believe that this is not a duplicate.
- [X] If an exception occurs when executing a command, I executed it again in debug mode (`-vvv` option).

- **OS version and name**: based on official docker image `python:3.8-slim-buster` (see below for Dockerfile)
- **Poetry version**: 1.1.4
- **virtualenv version**: 20.4.2
- **Link of a [Gist](https://gist.github.com/) with the contents of your pyproject.toml file**: https://gist.github.com/wichert/ba05b0505c24ee7c71c5d7deb2ac2cfe

## Issue
When I run `poetry install` in a private GitHub runner on an GHES installation I started getting an error (see further below for full output):

```
  TypeError

  expected string or bytes-like object

  at /usr/local/lib/python3.8/site-packages/poetry/core/utils/helpers.py:24 in canonicalize_name
```

## Debug output
```
Creating virtualenv amp-portal in /__w/backend/backend/amp-portal/.venv
Using virtualenv: /__w/backend/backend/amp-portal/.venv

  Stack trace:

  11  /usr/local/lib/python3.8/site-packages/clikit/console_application.py:131 in run
       129│             parsed_args = resolved_command.args
       130│ 
     → 131│             status_code = command.handle(parsed_args, io)
       132│         except KeyboardInterrupt:
       133│             status_code = 1

  10  /usr/local/lib/python3.8/site-packages/clikit/api/command/command.py:120 in handle
       118│     def handle(self, args, io):  # type: (Args, IO) -> int
       119│         try:
     → 120│             status_code = self._do_handle(args, io)
       121│         except KeyboardInterrupt:
       122│             if io.is_debug():

   9  /usr/local/lib/python3.8/site-packages/clikit/api/command/command.py:163 in _do_handle
       161│         if self._dispatcher and self._dispatcher.has_listeners(PRE_HANDLE):
       162│             event = PreHandleEvent(args, io, self)
     → 163│             self._dispatcher.dispatch(PRE_HANDLE, event)
       164│ 
       165│             if event.is_handled():

   8  /usr/local/lib/python3.8/site-packages/clikit/api/event/event_dispatcher.py:22 in dispatch
        20│ 
        21│         if listeners:
[...]

PR #4702, opened 2021-11-02
Handles name == None case (re: issue #4311, #3628)

aalok-sathe:
Addresses issue #4311 and others, related to `canonicalize_name` throwing TypeError because a str or bytes like object was expected.

# Pull Request Check List

Resolves: #4311, #3628

### DIFF
--- a/poetry/repositories/installed_repository.py
+++ b/poetry/repositories/installed_repository.py
@@ -231,8 +231,13 @@ class InstalledRepository(Repository):
                 metadata.distributions(path=[entry]),
                 key=lambda d: str(d._path),
             ):
-                name = canonicalize_name(distribution.metadata["name"])
-
+                try:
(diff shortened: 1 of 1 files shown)


neersighted (commented):
I'm not sure this is the best way to solve this problem -- one one hand, poetry should probably be more resilient to unexpected problems like empty egg/dist-info. On the other hand, if this is a poetry-managed environment, it's likely that poetry is creating these problems for itself.

I'll have to think about this one personally, after doing some digging and reproduction work.

At the very least, to be merge-able, we need to create a reproduction and regression test for this change.

aalok-sathe:
@neersighted You're probably right; I was looking for a quick fix at the time to remove a blockade in a project. Unfortunately I am not able to spend enough time on this at the moment to come up with a better solution.

Issue #4265, opened 2021-07-08
pip_editable_install() got an unexpected keyword argument 'upgrade'


jnoortheen:
- [ ] I am on the [latest](https://github.com/python-poetry/poetry/releases/latest) Poetry version.
- [x] I have searched the [issues](https://github.com/python-poetry/poetry/issues) of this repo and believe that this is not a duplicate.
- [ ] If an exception occurs when executing a command, I executed it again in debug mode (`-vvv` option).

- **OS version and name**: Manjaro Linux
- **Poetry version**: 1.2.0a1 (on master)
- **Link of a [Gist](https://gist.github.com/) with the contents of your pyproject.toml file**: 

## Issue

when installing packages without pyproject.toml in edit mode, it raises following exception

```

  at ~/.local/pipx/venvs/poetry@master/lib/python3.9/site-packages/poetry/installation/executor.py:129 in pip_install
      125│         if editable:
      126│             func = pip_editable_install
      127│ 
      128│         try:
    → 129│             func(req, self._env, upgrade=upgrade)
      130│         except EnvCommandError as e:
      131│             output = decode(e.e.output)
      132│             if (
      133│                 "KeyboardInterrupt" in output
```

jnoortheen:
the particular line that is causing to install the dependency is 
```
xonsh = { path = "xsh", extras = ["full"], develop = true }
```

desophos:
I'm on Poetry version 1.2.0a2 and I've encountered this issue as well.

It appears to be caused by the mismatched signatures of [`pip_install`](https://github.com/python-poetry/poetry/blob/61301a195685608cc3ea95a9df2aaec049a27dce/poetry/utils/pip.py#L18) and [`pip_editable_install`](https://github.com/python-poetry/poetry/blob/61301a195685608cc3ea95a9df2aaec049a27dce/poetry/utils/pip.py#L63). The `upgrade` kwarg is passed to both functions [in executor.py](https://github.com/python-poetry/poetry/blob/61301a195685608cc3ea95a9df2aaec049a27dce/poetry/installation/executor.py#L129), but `pip_editable_install` doesn't take such an argument.

Adding an `upgrade` kwarg to `pip_editable_install` fixed the problem:
```python
def pip_editable_install(directory: Path, environment: Env, upgrade=True) -> Union[int, str]:
    return pip_install(
        path=directory, environment=environment, editable=True, deps=False, upgrade=upgrade
    )
```

adamcharnock:
[...]

PR #4720, opened 2021-11-08
Fixed issue #4265

akharlamov:
pip_editable_install() got an unexpected keyword argument 'upgrade'

# Pull Request Check List

Resolves: #issue-number-here

### DIFF
--- a/get-poetry.py
+++ b/get-poetry.py
@@ -36,7 +36,6 @@ from gzip import GzipFile
 from io import UnsupportedOperation
 from io import open
 
-
 try:
     from urllib.error import HTTPError
     from urllib.request import Request
@@ -50,7 +49,6 @@ try:
     input = raw_input
 except NameError:
     pass
-
 
 try:
     try:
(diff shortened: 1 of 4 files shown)


neersighted:
Hi @akharlamov! Thanks for the contribution! Unfortunately, I don't think this is the best way to solve the linked issue -- @abn's suggestion of dropping the wrapper altogether is going to be the better fix. You're welcome to open another PR taking that approach if you'd like.

I'd like to note two other issues with this PR to keep in mind for another go-round -- first, there's a lot of whitespace churn that is not in agreement with [black](https://github.com/psf/black) -- so make sure to format your PRs with black in order to pass lint.

Second, we won't drop support for Python < 3.8, so bumping the Python version in `pyproject.toml` is going to be a non-starter (notice how some of the checks failed).

Keep that in mind if you want to take another shot we'll get it merged!

akharlamov:
Thanks for the critique. I'll take a look at the proposed design. May take some time though. 

I missed out on the `black` and updated python version by mistake. Will keep in mind. 

Thanks again

Issue #4721, opened 2021-11-08
required extras cause dependency not to be installed


wakemaster39:
- [x] I am on the [latest](https://github.com/python-poetry/poetry/releases/latest) Poetry version.
- [x] I have searched the [issues](https://github.com/python-poetry/poetry/issues) of this repo and believe that this is not a duplicate.
- [x] If an exception occurs when executing a command, I executed it again in debug mode (`-vvv` option).

- **OS version and name**:  macOS 11
- **Poetry version**:  1.2.0a2

## Issue

If all instances of a dependency requires extras, then the dependency is not being installed.

https://github.com/python-poetry/poetry/blob/5dcf24da04275425631c817419ad5cb17cbd0177/poetry/puzzle/solver.py#L156-L179

The continue on line 173 guarantees that this package never hits the add lines and won't be added.

neersighted:
Hi @wakemaster39! Thanks for opening an issue! It's pretty unclear what the expected behavior and the actual behavior is -- can you please clearly describe them both, as well as provide a `pyproject.toml` that reproduces the issue in a Gist?

zdog234:
I just ran into the same behavior.

It's showing up [with python3.10](https://gist.github.com/zdog234/7abfd318a762ab3a2dfeeb103fc55e6a) but isn't showing up [with python3.9](https://gist.github.com/zdog234/38255034458f18a569605352ebdb244f).

The expected behavior is that, when I run `poetry run pytest`, I should get an output saying "collected 0 items" and "no tests ran...". Instead, i get a `ModuleNotFoundError`:

```
...
File "/Users/zdufour/Github/poetry-issue/.venv/lib/python3.10/site-packages/pytest_cov/plugin.py", line 6, in <module>
    import coverage
ModuleNotFoundError: No module named 'coverage'
```

k-sriram:
I have had the same problem. It can even be recreated in a blank poetry project.

I am using `python 3.10` and `poetry 1.2.0a2`, on ubuntu-20.04 on WSL2

To recreate:

```sh
$ python -m venv deptest-venv
$ source deptest-venv/bin/activate
$ poetry new deptest
Created package deptest in deptest
$ cd deptest
$ poetry add "coverage[toml]"
Using version ^6.2 for coverage

Updating dependencies
Resolving dependencies... (0.8s)

Writing lock file

Package operations: 1 install, 0 updates, 0 removals

  • Installing tomli (1.2.2)
$ pip freeze
[...]

PR #4722, opened 2021-11-08
fix: required extras packages are installed

wakemaster39:
# Pull Request Check List

Resolves: #4721

I do not understand depths properly so this likely needs fixing. In my local example I have depths of (2,4,3) and this just always chooses the first instance.

### DIFF
--- a/poetry/puzzle/solver.py
+++ b/poetry/puzzle/solver.py
@@ -163,14 +163,18 @@ class Solver:
                         and not _package.is_same_package_as(package)
                         and _package.version == package.version
                     ):
+                        missing_deps = []
                         for dep in package.requires:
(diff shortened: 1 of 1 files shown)


neersighted:
Hi @wakemaster39! I think I'm going to opt to close this, for a couple reasons. Most importantly the code is broken and causes a deadlock. The change in logic is also non-obvious and there is no test coverage. I would suggest coming up with a minimal reproduction on your linked issue (#4721) and then writing a test case to reproduce it; and then adjusting Poetry's internal logic.

If you're unsure where to start, I will take a look at your issue later -- others may chime in as well, so make sure to provide a good/small step-by-step to reproduce your issue.

PR #4735, opened 2021-11-11
Clean up .cirrus.yml

neersighted:
Minor whitespace cleanup, push the $PATH out of the Matrix (which is
more aligned with Github Actions), and make the list of tasks more
obvious.

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -6,24 +6,25 @@ test_task:
   only_if: $CIRRUS_TAG == ''
   skip: "!changesInclude('.cirrus.yml', 'poetry.lock', 'pyproject.toml', '**.json','**.py')"
   env:
-    PATH: /.local/bin:${PATH}
     matrix:
       - PYTHON: python3.8
       - PYTHON: python3.9
-  python_script:
+  pkg_script:
     - PYPACKAGE=$(printf '%s' $PYTHON | tr -d '.')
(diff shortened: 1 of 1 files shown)


.cirrus.yml
```diff
@@ -6,24 +6,25 @@ test_task:
...
       - PYTHON: python3.9
-  python_script:
+  pkg_script:
     - PYPACKAGE=$(printf '%s' $PYTHON | tr -d '.')
     - SQLPACKAGE=$(printf '%s-sqlite3' $PYPACKAGE | sed 's/thon//')
     - pkg install -y git-lite curl $PYPACKAGE $SQLPACKAGE
   pip_script:
     - $PYTHON -m ensurepip
     - $PYTHON -m pip --disable-pip-version-check install -U pip
+  env_script:
```
abn, line 19: Ooi, what is the diffrence between the previous way of setting the path and this?

abn (commented):
LGTM, single question.

.cirrus.yml
neersighted: Mostly it doesn't clutter up the Cirrus UI/API with more matrix data since it's a static value.

Also, it makes it easier for me to compare the Github Workflow and the Cirrus one as they are more similar to each other.

abn (approved):

PR #4486, opened 2021-09-08
respect hash algorithm when verifying downloaded archives.

kaos:
Signed-off-by: Andreas Stenius <andreas.stenius@svenskaspel.se>

Resolves: #4085

- [x] Added **tests** for changed code.

Depends on python-poetry/poetry-core#196, so marked as draft until there is a new poetry-core version we can use here.

PR #4740, opened 2021-11-12
feat: Check all hash types when validating archives

### LINKED ISSUE #4578: Enable multiple file hashing methods for backwards and forwards compatibility

## Issue

My PR #4573 improves where the hash of an archive is checked against the hashes in the lockfile. While working on that PR, I spotted a potential improvement but want to collect feedback before I implement it.

I think this area could be further improved by representing the hashes as a `dataclass` or something containing the hash type and the string digest. The types of hashes would be more robust then instead of hardcoding sha256 in `poetry/installation/executor.py` and in `FileDependency.hash()`. It would also enable enumeration of hash types of known hashes so that `FileDependency.hash()` could return a list of hashes based on requested types. Doing this might alleviate the md5/sha256 problems arising from changes in poetry-core 1.0.6+ that affect Poetry 1.1.7+ users using Artifactory in at least https://github.com/python-poetry/poetry/issues/4523.

I'm willing to undertake this implementation if no maintainers object.

neersighted:
This allows for much-improved compatibility when interfacing with legacy
PyPI-compatible repositories.

This is a successor PR to #4486, and closes #4578 and #4085.
This PR is also a forward-port of #4529.

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -22,7 +22,7 @@ tests_no_zope = ["coverage[toml] (>=5.0.2)", "hypothesis", "pympler", "pytest (>
 
 [[package]]
 name = "backports.entry-points-selectable"
-version = "1.1.0"
+version = "1.1.1"
 description = "Compatibility shim providing selectable entry points for older implementations"
 category = "main"
 optional = false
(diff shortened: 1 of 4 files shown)


sdispater:
I think we can implement it similarly to what is present in 1.1 (see https://github.com/python-poetry/poetry/pull/4529).

To be honest, I forgot to port it to the master branch when fixing it in the 1.1 branch.

neersighted:
> 
> 
> I think we can implement it similarly to what is present in 1.1 (see #4529).
> 
> To be honest, I forgot to port it to the master branch when fixing it in the 1.1 branch.

Didn't even see that! I'll port that version to master and push that to this PR in a bit.

poetry/installation/executor.py
```diff
@@ -672,30 +673,40 @@ def _download_link(self, operation: Union[Install, Update], link: Link) -> Link:
                 archive = self._chef.prepare(archive)
 
         if package.files:
-            archive_hash = self._validate_archive_hash(archive, package)
+            archive_hashes = self._validate_archive_hash(archive, package)
 
-            self._hashes[package.name] = archive_hash
+            self._hashes[package.name] = archive_hashes[0]
```
neersighted, line 678: This is the bit I'm not sure about -- is it worth adding some code here to prefer a particular hash, or are we satisfied with just picking the first member of the set?

haf:
What's the status of this?

Poetry 1.1.11 fails with

> 'Link' object has no attribute 'name'

1.1.12 fails with:

```
Package operations: 34 installs, 0 updates, 0 removals

  • Installing fonttools (4.29.1): Failed

  RuntimeError

  Invalid hashes (sha256:9f14340a5d9b83844a568fd5df480b4a9c915d98803e750094b810b16653a4a7) for fonttools (4.29.1) using archive fonttools-4.29.1-py3-none-any.whl. Expected one of sha256:1933415e0fbdf068815cb1baaa1f159e17830215f7e8624e5731122761627557, sha256:2b18a172120e32128a80efee04cff487d5d140fe7d817deb648b2eee023a40e4.
```

neersighted:
Closing this as not a complete design/unable to be merged in its present form.

Issue #873, opened 2019-02-09
Can readme section conains several files?


proofit404:
Hi, thanks for an awesome tool.

To publish a package to the PyPI I usually combine project long description from the GitHub README and CHANGELOG from the docs site like this:

```python
readme = open("README.rst").read() + open("CHANGELOG.rst").read()

setup(
    ...
    long_description=readme,
    ...
)
```

Is there something similar to this technique in poetry?

Best regards,
Artem.

wagnerluis1982:
If suitable to the maintainer, @sdispater, I would like to implement the ability to combine more than one readme file.

yatmanov:
I second that.

zillionare:
Please have this implemented, thanks!

DudeNr33:
Any news on this? The linked PR is open since last year. This would be a really helpful addition!

wagnerluis1982:
Given python-poetry/poetry-core#118 was reverted, can someone reopen this issue?

PR #4743, opened 2021-11-13
Add docs and schema for multiple readme files support

wagnerluis1982:
- Add documentation for code at python-poetry/poetry-core#248
- Update `poetry-schema.json` to match core's version

# Pull Request Check List

Related to #873

### DIFF
--- a/docs/pyproject.md
+++ b/docs/pyproject.md
@@ -73,6 +73,29 @@ The readme file of the package. **Optional**
 
 The file can be either `README.rst` or `README.md`.
 
+```toml
+[tool.poetry]
+# ...
+readme = "README.rst"
+```
+
+If you want to use multiple readme files, you can by wrapping the files in a list.
+
+```toml
+[tool.poetry]
+# ...
+readme = [
+    "README.rst",
+    "CHANGELOG.rst"
+]
+```
+
+{{% note %}}
+Note that a line break (`\n`) will be added between each file in the generated readme.
+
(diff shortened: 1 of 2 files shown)


neersighted (changes requested):
This doc change is relatively meaningless without the core changes to go with it -- by that, I mean that even after your changes are merged into core, they will not be reflected.

Between alpha releases, poetry `master` normally tracks poetry-core. I would suggest waiting until your PR lands in core, and then updating `pyproject.toml` to reflect the Git dependency so that your changes can be used.

wagnerluis1982:
@neersighted, please remember to also merge this, given core changes at python-poetry/poetry-core#248 is merged 😊

neersighted:
I'm aware! I'm not sure of merging this yet because of two things -- the schema in this repo is actually useless, so I'm thinking we merge a PR to drop it and drop that change from this PR.

Additionally, the docs in `master` will be out of sync with the poetry-core dep -- still debating whether we add a git dep or just wait to merge this until after a poetry-core release. I'm not an expert there so I'll have to pick some other brains.

Secrus:
@neersighted did anything change in the mean time or is it still undecided?

neersighted:
> @neersighted did anything change in the meantime or is it still undecided?

We've had a core release since then, so we should be good to just drop the schema in this PR, and update the docs to reflect the current reality.

neersighted:
This has been superseded by #5678 and #5158. Thank you for your contribution!

PR #4750, opened 2021-11-13
feat(ci): switch to pre-commit.ci for code quality

neersighted:
pre-commit.ci is faster, and has less moving parts than our current Github Actions workflow. It integrates with Github checks, it can auto-correct PRs if desired (turned off in this PR), and it will automatically generate PRs to keep our `pre-commit` hooks up to date.

Note that this PR won't actually show pre-commit.ci working until an organization admin enables it (additionally, the protected branch config will have to be updated).


.pre-commit-config.yaml
```diff
@@ -1,41 +1,50 @@
...
+      - id: check-case-conflict
+      - id: check-ast
+      - id: debug-statements
+      - id: check-toml
+      - id: check-json
+
+  - repo: https://github.com/pre-commit/pygrep-hooks
+    rev: v1.9.0
+    hooks:
+      # - id: python-check-blanket-noqa # XXX: disallow blanket # noqa
```
neersighted, line 18: I've added commented out hooks that I think could benefit the project, but that currently would cause a lot of churn out-of-scope of this initial PR.

.pre-commit-config.yaml
```diff
@@ -1,41 +1,50 @@
...
+
+  - repo: https://github.com/pycqa/isort
+    rev: 5.10.1
+    hooks:
+      - id: isort
+        types: [python] # FIXME: pyi support that doesn't conflcit with black
+
   - repo: https://github.com/psf/black
-    rev: 20.8b1
+    rev: 21.10b0
```
neersighted, line 30: All the pre-commit hooks are updated in this PR; pre-commit.ci will keep them up-to-date in the future. The non-config changes in this PR are a result of our out-of-date hooks.

.pre-commit-config.yaml
```diff
@@ -1,41 +1,50 @@
...
+      # - id: python-check-blanket-noqa # XXX: disallow blanket # noqa
+      - id: python-check-blanket-type-ignore
+      - id: python-check-mock-methods
+      # - id: python-use-type-annotations # XXX: disallow type comments
+
+  - repo: https://github.com/pycqa/isort
+    rev: 5.10.1
+    hooks:
+      - id: isort
+        types: [python] # FIXME: pyi support that doesn't conflcit with black
```
neersighted, line 27: Currently getting an odd bug where `isort` tries to add two blank lines after imports in `.pyi` files, contrary to both typeshed and black conventions. If I can (later) bisect this properly I'll open an issue on isort.

.pre-commit-config.yaml
```diff
@@ -1,41 +1,50 @@
...
       - id: black
 
   - repo: https://gitlab.com/pycqa/flake8
-    rev: 3.8.4
+    rev: 4.0.1
     hooks:
       - id: flake8
-        additional_dependencies: [flake8-bugbear]
+        additional_dependencies:
+          # - flake8-docstring # XXX: lint docstrings
```
neersighted, line 39: These are a harder sell, but still are useful tools. I think the mypy project should be completed before either. Some of the non-PEP8 compliant names may be part of the exposed external API and therefore require at least a minor version change to be corrected.

pyproject.toml
```diff
@@ -72,38 +72,61 @@ build-backend = "poetry.core.masonry.api"
 
 
 [tool.isort]
```
neersighted, line 74: Many of the options specified here were cargo-culted and did not do anything, or were done poorly. This is cleaned up to be as simple as I can manage given my experience with the tooling.

pyproject.toml
```diff
@@ -72,38 +72,61 @@ build-backend = "poetry.core.masonry.api"
...
-skip_glob = ["*/setup.py"]
-filter_files = true
-known_first_party = "poetry"
+extend_skip = ["setup.py"]
 
 
 [tool.black]
-line-length = 88
+target-version = ['py36']
 include = '\.pyi?$'
```
neersighted, line 87: Exclude actually wasn't doing anything here -- modern black has a much better approach that ignores files even when pre-commit passes them directly.

pyproject.toml
```diff
@@ -72,38 +72,61 @@ build-backend = "poetry.core.masonry.api"
...
-lines_after_imports = 2
+combine_as_imports = true
 lines_between_types = 1
-use_parentheses = true
+lines_after_imports = 2
 src_paths = ["poetry", "tests"]
-skip_glob = ["*/setup.py"]
-filter_files = true
-known_first_party = "poetry"
+extend_skip = ["setup.py"]
```
neersighted, line 82: This avoids clobbering the stock skip-list.

pyproject.toml
```diff
@@ -72,38 +72,61 @@ build-backend = "poetry.core.masonry.api"
...
+warn_unused_ignores = true
+files = "poetry/"
+
+# The following whitelist is used to allow for incremental adoption
+# of Mypy. Modules should be removed from this whitelist as and when
+# their respective type errors have been addressed. No new modules
+# should be added to this whitelist.
+# see https://github.com/python-poetry/poetry/pull/4510.
+
+[[tool.mypy.overrides]]
```
neersighted, line 110: This will cause merge conflicts with some of the mypi work, but I think this is a much better place to keep the config (pyproject.toml is meant to be the one file to rule them all, after all), and makes it much more concise.

pyproject.toml
```diff
@@ -72,38 +72,61 @@ build-backend = "poetry.core.masonry.api"
...
-  | dist
-  | tests/.*/setup.py
-)/
+force-exclude = '''
+.*/setup\.py$
 '''
 
+
+[tool.mypy]
+python_version = "3.6"
```
neersighted, line 94: I've added supported Python versions to all the code quality tools -- it's some duplication, but it's all in pyproject.toml and prevents environment-specific differences.

neersighted:
Incidentally, it looks like pre-commit.ci is going to run against all future PR pulls. To avoid unwanted auto-fix commits as much as possible, I can create a reduced subset of this that just disables the auto-fix commits until the larger PR can be merged.

image

pyproject.toml
```diff
@@ -72,38 +72,58 @@ build-backend = "poetry.core.masonry.api"
...
-skip_glob = ["*/setup.py"]
-filter_files = true
-known_first_party = "poetry"
+extend_skip = ["setup.py"]
 
 
 [tool.black]
-line-length = 88
+target-version = ['py36']
 include = '\.pyi?$'
```
branchvincent, line 87: this was the default (now extended to [include `.ipynb`](https://github.com/psf/black/blob/5e191c29d4492d14f9090a924a77b3dc055443a6/src/black/const.py#L3)), maybe just remove?

neersighted: Good catch -- it's in the black docs so somehow I assumed it was needed.

.pre-commit-config.yaml
```diff
@@ -1,41 +1,50 @@
...
-    rev: 5.7.0
-    hooks:
-      - id: isort
-        additional_dependencies: [toml]
-        exclude: ^.*/?setup\.py$
 
-  - repo: https://github.com/pre-commit/pre-commit-hooks
-    rev: v3.4.0
-    hooks:
-      - id: trailing-whitespace
```
neersighted, line 33: These are gone because `black` provides them for Python files. If they're desired for other file types they can easily be added back.

neersighted:
Fixed the failing checks by making the old `code-quality` workflow a no-op.

.pre-commit-config.yaml
abn: I believe this is an isort regression, there is a historic version where this issue was resolved to align with the black profile.

See: https://github.com/PyCQA/isort/issues/1284

.pre-commit-config.yaml
abn: I'd honestly remove unused hooks for now, and add them as and when we enable them.

abn (changes requested):
Changes look good to me. I'd request that we remove the commented out checks for now and enable them as and when we need them.

.pre-commit-config.yaml
```diff
@@ -1,41 +1,50 @@
...
+      - id: isort
+        types: [python] # FIXME: pyi support that doesn't conflcit with black
+
   - repo: https://github.com/psf/black
-    rev: 20.8b1
+    rev: 21.10b0
     hooks:
       - id: black
 
   - repo: https://gitlab.com/pycqa/flake8
```
branchvincent, line 34: fyi, `flake8` has moved to GitHub https://github.com/PyCQA/flake8 (gitlab is now just a mirror)

neersighted: Good catch!

neersighted:
> 
> 
> Changes look good to me. I'd request that we remove the commented out checks for now and enable them as and when we need them.

Done!

tests/repositories/fixtures/pypi.org/json/clikit.json
```diff
@@ -1,6 +1,6 @@
 {
   "info": {
-    "author": "Sébastien Eustace",
+    "author": "S\u00e9bastien Eustace",
```
abn, line 3: Is this one expected?

neersighted: This is expected given https://simplejson.readthedocs.io/en/latest/#simplejson.dumps. There is no provision to disable this upstream -- I could add a argument if it's desired.

On the other hand, the spec allows for either encoded or non-encoded UTF-8 characters. This file is going to decode the same. If it's a bit too far (vs just re-indenting), I can roll it back until we get the ability to disable these escapes.

.pre-commit-config.yaml
neersighted: Yeah, came across that as well. Pinged the maintainer -- for now I'm just going to leave the FIXME.

tests/repositories/fixtures/pypi.org/json/clikit.json
neersighted: It turns out I was looking at an old version of the hook and upstream already has support for turning this off.

https://github.com/pre-commit/pre-commit-hooks/blob/v4.0.1/pre_commit_hooks/pretty_format_json.py#L77

abn (approved):

PR #4755, opened 2021-11-14
more typechecking fixes

dimbleby:
# Pull Request Check List

Resolves: #issue-number-here

Encouraged by seeing #4751 merged - and so fast, thank you! - I took on a few more of the relatively easy mypy warnings.

I have tried to prefer updating the annotations where possible, but this time I have found reasons to make some modest rearrangements to the code:

- [`Config.name`](https://github.com/python-poetry/poetry/blob/1ae27b04f2a29fa32b5814b542580ae3699fe8dc/poetry/config/config.py#L51-L53) seems to be just broken as there is no `_file` attribute.  I find no callers, so have removed it.

- refactored authentication in uploader.py so that mypy can see that the username and password are not `None`

- AIUI poetry is dropping support for python2, so I just went with python3 imports of `html.unescape` and `urllib.parse.quote`

- renamed `LegacyRepositoy._get()` as `LegacyRepository._get_page()`: so far as I can see it has little in common with its parent class's `PyPiRepository._get()`, and mypy complains about them not being the same

One place where I have just updated the annotation to match the code but which smells funny to me is in pool.py.  Is it really intended that the `repository_name` indexing `self._lookup` might be `None`?  I tried asserting that it was not, but lots of testcases failed...

### DIFF
--- a/poetry/config/config.py
+++ b/poetry/config/config.py
@@ -14,9 +14,6 @@ from .config_source import ConfigSource
 from .dict_config_source import DictConfigSource
 
 
-_NOT_SET = object()
-
-
 def boolean_validator(val: str) -> bool:
     return val in {"true", "false", "1", "0"}
 
@@ -27,7 +24,7 @@ def boolean_normalizer(val: str) -> bool:
 
 class Config:
 
-    default_config = {
+    default_config: Dict[str, Any] = {
         "cache-dir": str(CACHE_DIR),
         "virtualenvs": {
             "create": True,
@@ -45,12 +42,8 @@ class Config:
         self._config = deepcopy(self.default_config)
         self._use_environment = use_environment
         self._base_dir = base_dir
-        self._config_source = DictConfigSource()
-        self._auth_config_source = DictConfigSource()
-
-    @property
-    def name(self) -> str:
-        return str(self._file.path)
+        self._config_source: ConfigSource = DictConfigSource()
+        self._auth_config_source: ConfigSource = DictConfigSource()
 
     @property
     def config(self) -> Dict:
@@ -114,9 +107,9 @@ class Config:
             env = "POETRY_{}".format(
                 "_".join(k.upper().replace("-", "_") for k in keys)
             )
-            value = os.getenv(env, _NOT_SET)
-            if value is not _NOT_SET:
-                return self.process(self._get_normalizer(setting_name)(value))
+            env_value = os.getenv(env)
+            if env_value is not None:
+                return self.process(self._get_normalizer(setting_name)(env_value))
 
         value = self._config
         for key in keys:
--- a/poetry/publishing/uploader.py
+++ b/poetry/publishing/uploader.py
@@ -57,8 +57,8 @@ class Uploader:
         self._poetry = poetry
         self._package = poetry.package
         self._io = io
-        self._username = None
-        self._password = None
+        self._username: Optional[str] = None
+        self._password: Optional[str] = None
 
     @property
     def user_agent(self) -> str:
@@ -89,14 +89,15 @@ class Uploader:
 
         return sorted(wheels + tars)
 
-    def auth(self, username: str, password: str) -> None:
+    def auth(self, username: Optional[str], password: Optional[str]) -> None:
         self._username = username
         self._password = password
 
     def make_session(self) -> requests.Session:
         session = requests.session()
-        if self.is_authenticated():
-            session.auth = (self._username, self._password)
+        auth = self.get_auth()
+        if auth is not None:
+            session.auth = auth
 
         session.headers["User-Agent"] = self.user_agent
         for scheme in ("http://", "https://"):
@@ -104,8 +105,11 @@ class Uploader:
 
         return session
 
-    def is_authenticated(self) -> bool:
-        return self._username is not None and self._password is not None
+    def get_auth(self) -> Optional[Tuple[str, str]]:
+        if self._username is None or self._password is None:
+            return None
+
+        return (self._username, self._password)
 
     def upload(
         self,
@@ -147,16 +151,15 @@ class Uploader:
 
         md5_digest = md5_hash.hexdigest()
         sha2_digest = sha256_hash.hexdigest()
+        blake2_256_digest: Optional[str] = None
         if _has_blake2:
             blake2_256_digest = blake2_256_hash.hexdigest()
-        else:
-            blake2_256_digest = None
 
+        py_version: Optional[str] = None
         if file_type == "bdist_wheel":
             wheel_info = wheel_file_re.match(file.name)
-            py_version = wheel_info.group("pyver")
-        else:
-            py_version = None
+            if wheel_info is not None:
+                py_version = wheel_info.group("pyver")
 
         data = {
             # identify release
@@ -231,7 +234,7 @@ class Uploader:
             }
         )
 
-        data_to_send = self._prepare_data(data)
+        data_to_send: List[Tuple[str, Any]] = self._prepare_data(data)
 
         with file.open("rb") as fp:
             data_to_send.append(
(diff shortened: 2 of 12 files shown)


poetry/config/config.py
```diff
@@ -45,12 +42,8 @@ def __init__(
         self._config = deepcopy(self.default_config)
         self._use_environment = use_environment
         self._base_dir = base_dir
-        self._config_source = DictConfigSource()
-        self._auth_config_source = DictConfigSource()
-
-    @property
```
neersighted, line 51: Funny this wasn't use -- was probably added defensively in case it was needed when the config code was written. Given it's dead, no harm in removing it.

poetry/config/config.py
```diff
@@ -114,9 +107,9 @@ def get(self, setting_name: str, default: Any = None) -> Any:
             env = "POETRY_{}".format(
                 "_".join(k.upper().replace("-", "_") for k in keys)
             )
-            value = os.getenv(env, _NOT_SET)
-            if value is not _NOT_SET:
-                return self.process(self._get_normalizer(setting_name)(value))
+            env_value = os.getenv(env)
```
neersighted, line 110: Wow, the old construct was gross. This is a lot better.

poetry/publishing/uploader.py
```diff
@@ -89,23 +89,27 @@ def files(self) -> List[Path]:
...
 
         session.headers["User-Agent"] = self.user_agent
         for scheme in ("http://", "https://"):
             session.mount(scheme, self.adapter)
 
         return session
 
-    def is_authenticated(self) -> bool:
-        return self._username is not None and self._password is not None
+    def get_auth(self) -> Optional[Tuple[str, str]]:
```
neersighted, line 108: This is a much more elegant construct, and `is_authenticated` was an ambiguous name. Nice.

poetry/publishing/uploader.py
```diff
@@ -217,8 +220,8 @@ def _upload(
             ):
                 try:
                     self._register(session, url)
-                except HTTPError as e:
-                    raise UploadError(e)
+                except HTTPError as e2:
```
neersighted, line 223: You're going to get conflicts here most likely as this was just refactored (and it fixed the poor structure here).

poetry/publishing/uploader.py
```diff
@@ -230,7 +233,7 @@ def _do_upload(
 
             resp = self._upload_file(session, url, file, dry_run)
 
-            if not dry_run:
+            if resp is not None:
```
neersighted, line 236: See above.

poetry/repositories/legacy_repository.py
```diff
@@ -419,7 +404,7 @@ def _get_release_info(self, name: str, version: str) -> dict:
 
         return data.asdict()
 
-    def _get(self, endpoint: str) -> Optional[Page]:
```
neersighted, line 422: What was wrong with _get?

poetry/repositories/pool.py
```diff
@@ -22,11 +22,11 @@ def __init__(
         if repositories is None:
             repositories = []
 
-        self._lookup: Dict[str, int] = {}
+        self._lookup: Dict[Optional[str], int] = {}
```
neersighted, line 25: Not sure why there is an `Optional` here -- None will all hash to the same type in the dict, and I don't think it should ever end up here. I suspect the error is found elsewhere.

poetry/repositories/repository.py
```diff
@@ -24,7 +24,7 @@ def __init__(self, packages: List["Package"] = None, name: str = None) -> None:
             self.add_package(package)
 
     @property
-    def name(self) -> str:
+    def name(self) -> Optional[str]:
```
neersighted, line 27: Likewise I'm not sure that this Optional is correct. I think the source stems somewhere else.

neersighted (changes requested):
Looks mostly good with some questions and some corrections. Needs to be rebased back onto current master which refactored some of the same files and fixed some of the issues addressed here in a better way.

poetry/repositories/legacy_repository.py
dimbleby: > renamed LegacyRepositoy._get() as LegacyRepository._get_page(): so far as I can see it has little in common with its parent class's PyPiRepository._get(), and mypy complains about them not being the same

poetry/repositories/pool.py
dimbleby: > One place where I have just updated the annotation to match the code but which smells funny to me is in pool.py. Is it really intended that the repository_name indexing self._lookup might be None? I tried asserting that it was not, but lots of testcases failed...

poetry/repositories/repository.py
dimbleby: ditto

dimbleby:
Rebased.  I have mostly answered your questions by quoting myself, do say if you'd already seen those comments and still are confused!

poetry/repositories/legacy_repository.py
neersighted: I think the correct solution here is going to be refactoring LegacyRepository to not inherit from PypiRepository. I'd suggest doing that in a follow-up PR, and leaving some `# type:` and `# FIXME` annotations here in the meantime.

poetry/repositories/pool.py
neersighted: I'm pretty sure this is unintentional behavior (even if it doesn't cause a bug). For now I'd just whitelist it with an annotation, and circle back to address this on a follow-up PR.

poetry/repositories/repository.py
neersighted: Above

neersighted (changes requested):

poetry/repositories/legacy_repository.py
dimbleby: Even if that's the right long-term fix, `get_page()` seems to me a sensible name for this method: which gets a `Page`.  Indeed separating this class from the `PyPiRepository` would leave _less_ reason for their methods to have the same names.

So I'd prefer to leave this change in place: so that we can leave this file mypy-clean - rather than having to whitelist it and allow other errors to slip back in.

(I do not expect to follow up with the change that you propose.  Not that I disagree with it; but it's more work than I intend to be taking on here.)

poetry/repositories/pool.py
dimbleby: I'm in two minds about this.  
- The advantage of changing the annotation as I have it is that we can then remove this file from the whitelist, and be confident that further typing errors will not happen.  
- The advantage of your suggestion is that it reminds us that there is a problem here.

Well, why not both?  Suggest that we leave the annotation in place for now - it does, after all, reflect what the code is actually doing - but also add a `# FIXME` saying "surely there's something wrong here".

(Again, I do not expect to follow-up with any fix for that - this looks like quite a pervasive change).

poetry/repositories/repository.py
dimbleby: I think the proposed FIXME comment in pool.py should be sufficient: it'll be impossible to get rid of the Option there without considering this.

(For much the same reasons as elsewhere, I think it's valuable to have annotations that match the reality of the code, even if we do think that the code is suboptimal)

dimbleby:
I've added a single FIXME - reasoning, as above, that 
- `_get_page()` is a good name regardless of future refactoring
- accurate annotations are valuable even for inaccurate code

You'll say if you're not persuaded that this is good enough!

poetry/repositories/legacy_repository.py
neersighted: I think that's sounds logic -- let's go ahead and add a `# TODO:` annotation above the class as a reminder that this needs to be revisited.

poetry/repositories/pool.py
neersighted: I think the compromise here is good. Fair enough!

neersighted (changes requested):
I'd like to leave one more annotation and then we can land this.

The real purpose of mypy checking (in my mind, and many of the other maintainers I suspect) is to force us to consider the correctness of our code and soundness of design. Any PRs to get a file 'mypy clean' will be touching on correctness as well -- like we saw with the (small but excellent) credentials refactor.

I think that leaving some annotations where we've identified soundness issues thanks to mypy will suffice for now -- and hopefully before too long we can revisit those files and refactor them to be more robust.

dimbleby:
Sounds good, thanks for the thoughtful review

neersighted (approved):

PR #4761, opened 2021-11-15
ci: test all platforms on py3.10

branchvincent:
# Pull Request Check List

Update tests to use the released version of Python 3.10. The required checks may need to be updated to include the new version.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -17,20 +17,27 @@ on:
 
 jobs:
   tests:
-    name: ${{ matrix.os }} / ${{ matrix.python-version }}
+    name: ${{ matrix.os }} / ${{ matrix.python-version }} ${{ matrix.suffix }}
     runs-on: ${{ matrix.os }}-latest
     continue-on-error: ${{ matrix.experimental }}
     strategy:
       matrix:
         os: [Ubuntu, MacOS, Windows]
-        python-version: [3.6, 3.7, 3.8, 3.9]
+        python-version: ["3.6", "3.7", "3.8", "3.9", "3.10"]
         experimental: [false]
         bootstrap-args: [""]
+        suffix: [""]
         include:
           - os: Ubuntu
-            python-version: "3.10.0-alpha - 3.10.0"
+            python-version: "3.10"
             experimental: true
             bootstrap-args: "--git https://github.com/python-poetry/poetry.git"
+            suffix: (HEAD)
+          - os: Ubuntu
+            python-version: "3.11-dev"
+            experimental: true
+            bootstrap-args: "--git https://github.com/python-poetry/poetry.git"
+            suffix: (HEAD)
       fail-fast: false
     steps:
       - uses: actions/checkout@v2
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -9,6 +9,7 @@ test_task:
     matrix:
       - PYTHON: python3.8
       - PYTHON: python3.9
+      - PYTHON: python3.10
   pkg_script:
     - PYPACKAGE=$(printf '%s' $PYTHON | tr -d '.')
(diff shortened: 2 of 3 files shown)


neersighted (changes requested):
A couple thoughts here. `skip.yml` will need to be updated with the same matrix (it serves to mark our required checks as passing when they are skipped).

Also, I'd like to keep a single test with Poetry from `master` installing itself. What I would suggest doing is keeping both the `python-version` and `include` changes, but add a `(HEAD)` marker to the name of the `include`d job.

neersighted:
Also, Python 3.11 alpha (in addition to master/HEAD on 3.10) would be good to add as experimental as well. I think a suffix variable should be used to separate out the jobs.

branchvincent:
sounds good! thoughts on adding 3.10 to `.cirrus.yaml`?

https://github.com/python-poetry/poetry/blob/aaf8806227ac1dc63eac2bb0991b0c1525350986/.cirrus.yml#L10-L11

neersighted:
> 
> 
> sounds good! thoughts on adding 3.10 to `.cirrus.yaml`?
> 
> https://github.com/python-poetry/poetry/blob/aaf8806227ac1dc63eac2bb0991b0c1525350986/.cirrus.yml#L10-L11

Go ahead! The `python310` package is in the repos now.

.github/workflows/main.yml
```diff
@@ -17,20 +17,27 @@ on:
...
-            python-version: "3.10.0-alpha - 3.10.0"
+            python-version: "3.10"
             experimental: true
             bootstrap-args: "--git https://github.com/python-poetry/poetry.git"
+            suffix: (HEAD)
+          - os: Ubuntu
+            python-version: "3.11-dev"
+            experimental: true
+            bootstrap-args: "--git https://github.com/python-poetry/poetry.git"
+            suffix: (Python Dev)
```
neersighted, line 40:
```suggestion
            suffix: (HEAD)
```

`HEAD` should be sufficient here as the version is already `3.11-dev`

.github/workflows/main.yml
```diff
@@ -17,20 +17,27 @@ on:
 
 jobs:
   tests:
-    name: ${{ matrix.os }} / ${{ matrix.python-version }}
+    name: ${{ matrix.os }} / ${{ matrix.python-version }} ${{ matrix.suffix }}
```
neersighted, line 20: I wonder if this will add a space to the job name/break the required checks.

neersighted (changes requested):

.github/workflows/main.yml
neersighted: It looks like this doesn't cause any issues. Cool!

branchvincent: i wondered that as well, but the jobs still show the `Required` annotation, and highlighting the space between e.g. `3.6 (pull_request)` shows only a single space which is the same as before this PR. maybe github actions trims trailing spaces?

neersighted (approved):

branchvincent:
hmm, i'm not a fan of the overall actions status showing a failure ❌  when it's only a failure from an experimental test...

neersighted:
> 
> 
> hmm, i'm not a fan of the overall actions status showing a failure ❌ when it's only a failure from an experimental test...

Yeah, I've spent the last little while looking for a workaround, but there is none. Let's disable it for now until we can get Python 3.11 passing for a decent amount of the time.

branchvincent:
[from this comment](https://github.com/actions/toolkit/issues/399#issuecomment-695039432), seems adding `continue-on-error` to the individual step will give the run a ✅ even when it fails. the downside would be manually inspecting the logs to determine the actual pass/fail status, but does give us an overall ✅

neersighted:
I went ahead and merged this as I found a (test suite level) workaround for the broken Python 3.11 check. I'll create a PR that includes that and a couple other workflow enhancements.

PR #4762, opened 2021-11-15
ci: test improvements

neersighted:
* Drop testing of 3.11 / install using HEAD for now to increase test suite stability. This can be revisited at a later time.
* DRY workflows and remove the install-poetry.py tests as this repo is no longer canonical.
* Migrate to macOS 11 and Windows Server 2022 (Windows 11) images to expose poetry to more recent OSes. 
* Fully fill out the FreeBSD matrix without taking any additional time, by scaling down VMs to fit in the 8-core concurrency limit.
* Make Github Actions faster on pull requests by killing jobs when there is a new push via concurrency limits.
* Run isort against our three `.pyi` files again with guidance from upstream on how to resolve the style conflict with `black`.
* Add a couple incidental pre-commit hooks.
* Clean up `.flake8` and try to reduce the amount of cargo culting and unknown config options.


neersighted:
Dropped an attempt to test against `windows-2022` and `macos-11` due to an inability to make the matrix present a pretty name but require the specific images.

.github/workflows/main.yml
```diff
@@ -15,6 +15,10 @@ on:
     branches:
       - '**'
 
+concurrency:
+  group: ${{ github.head_ref }}
+  cancel-in-progress: true
```
branchvincent, line 20: do you want to cancel in-progress tests on the e.g. `master` branch (this will result in ❌ when skipped)? or just PR's? if the latter 
```suggestion
  cancel-in-progress: ${{ github.event_name == 'pull_request' }}
```

neersighted: I was reasonably sure from the docs that `${{ github.head_ref }}` would only be present during PRs, and an empty concurrent group would prevent this from taking effect. Did you find something to the contrary?

branchvincent: ah, nope! just missed the `head_ref` only being present on PR's

.github/workflows/main.yml
```diff
@@ -15,10 +15,19 @@ on:
...
+
 jobs:
   tests:
     name: ${{ matrix.os }} / ${{ matrix.python-version }} ${{ matrix.suffix }}
-    runs-on: ${{ matrix.os }}-latest
+    runs-on: >-
```
neersighted, line 25: A little gross but I got it to work.

.pre-commit-config.yaml
```diff
@@ -26,7 +28,12 @@ repos:
     rev: 5.10.1
     hooks:
       - id: isort
-        types: [python] # FIXME: pyi support that doesn't conflict with black
+        name: "isort (python)"
+        types: [python]
+      - id: isort
+        name: "isort (pyi)"
+        types: [pyi]
+        args: [--lines-after-imports, "-1"]
```
neersighted, line 36: Also mildly gross, but it does work.

.cirrus.yml
```diff
@@ -1,12 +1,16 @@
...
+  cpu: 1
+  memory: 4G
 
 test_task:
   name: "Tests / FreeBSD / "
   only_if: $CIRRUS_TAG == ''
   skip: "!changesInclude('.cirrus.yml', 'poetry.lock', 'pyproject.toml', '**.json','**.py')"
   env:
     matrix:
+      - PYTHON: python3.6
```
neersighted, line 12: Note that with the scaled down VMs, all FreeBSD tests run in about the same time as the slowest Github Actions tests (4 minutes) since they're fully parallel.

.flake8
```diff
@@ -1,24 +1,20 @@
 [flake8]
-max-line-length = 88
-ignore = E501, E203, W503
+ignore =
```
neersighted, line 2: Cleaned up as best I can through trial and error.

.github/workflows/main.yml
```diff
@@ -15,30 +15,41 @@ on:
...
       matrix:
-        os: [Ubuntu, MacOS, Windows]
+        os: [Ubuntu, macOS, Windows]
         python-version: ["3.6", "3.7", "3.8", "3.9", "3.10"]
-        experimental: [false]
         bootstrap-args: [""]
         suffix: [""]
         include:
           - os: Ubuntu
+            image: ubuntu-latest
```
neersighted, line 34: Figured out a way to elegantly include updated OS images (not just blanket `-latest` images).

.github/workflows/main.yml
```diff
@@ -15,30 +15,41 @@ on:
...
             bootstrap-args: "--git https://github.com/python-poetry/poetry.git"
             suffix: (HEAD)
           - os: Ubuntu
+            image: ubuntu-latest
             python-version: "3.11-dev"
-            experimental: true
             bootstrap-args: "--git https://github.com/python-poetry/poetry.git"
             suffix: (HEAD)
       fail-fast: false
+    defaults:
```
neersighted, line 50: This significantly cleans up our workflows.

.github/workflows/skip.yml
```diff
@@ -3,18 +3,18 @@ name: Skip All Jobs
 
 on:
   push:
-    paths:
-      - '!**'
-      - 'docs/**'
-      - '.cirrus.yml'
+    paths-ignore:
```
neersighted, line 6: Allows for cleaner copy-and-paste from the main workflow.

.github/workflows/release.yml
```diff
@@ -6,17 +6,13 @@ on:
...
-  Release:
+  release:
+    name: Release
     runs-on: ubuntu-latest
-
     steps:
       - name: Checkout code
         uses: actions/checkout@v2
 
-      - name: Get tag
```
neersighted, line 16: Seems totally unused.

.pre-commit-config.yaml
```diff
@@ -9,12 +9,14 @@ repos:
       - id: end-of-file-fixer
       - id: check-merge-conflict
       - id: check-case-conflict
-      - id: check-toml
       - id: check-json
+      - id: check-toml
+      - id: check-yaml
```
neersighted, line 14: Why not -- we have Cirrus/Github Actions YAML files.

pyproject.toml
```diff
@@ -55,6 +55,8 @@ importlib-metadata = {version = "^1.6.0", python = "<3.8"}
 dataclasses = {version = "^0.8", python = "~3.6"}
 
 [tool.poetry.dev-dependencies]
+# required for python 3.11
+py = ">=1.11.0"
```
neersighted, line 59: Temporary, but necessary for Python 3.11.

.flake8
abn: We should at a later time leverage in-line excludes where we really need them and eventually remove these.

.github/workflows/main.yml
abn: This adds some toil for us right? Wonder if dependency bot supports bumping these.

neersighted: I think you might have commented too low -- were you talking about the images?

If so, I don't think so. These change rarely, and there is an announcement on the Github Blog.

.flake8
neersighted: I think we're always going to keep E501 for what it's worth, as black is not strict about its line wrapping and flake8 is (which is why we don't tell flake8 about our line length).

abn (approved):

.github/workflows/main.yml
abn: Correct. Ah will check the post. Thanks.

.flake8
```diff
@@ -1,24 +1,21 @@
 [flake8]
-max-line-length = 88
-ignore = E501, E203, W503
+min_python_version = 3.6.0
+ignore =
+    # E501: Line too long
+    E501,
+    # W503: Line break occurred before a binary operator
+    W503
```
neersighted, line 7: This one is also likely to stay @abn -- the accepted practice is opposite what flake8 suggests.

See:
https://www.flake8rules.com/rules/W503.html
https://github.com/psf/black/issues/52
https://github.com/psf/black/pull/36

Although, oddly current flake8 is supposed to have this off by default?
https://github.com/psf/black/blob/6559bdbd9dcf71ba0017d66bb133c744bd83b0e3/docs/guides/using_black_with_other_tools.md
That doesn't appear to be happening here, which is odd.

hukkin: @neersighted FYI the check is off because `ignore` is used and it overrides all on-by-default ignores. There is `extend-ignore` if you want to extend, not override.

neersighted: Yeah, I figured that out. It's actually resolved in master now. Thanks for looking into this!

PR #4764, opened 2021-11-15
chore(puzzle): remove type errors

branchvincent:
# Pull Request Check List

Work towards typechecking the entire codebase by unignoring `poetry.puzzle`

### DIFF
--- a/src/poetry/installation/operations/operation.py
+++ b/src/poetry/installation/operations/operation.py
@@ -1,9 +1,12 @@
 from typing import TYPE_CHECKING
 from typing import Optional
+from typing import TypeVar
 
 
 if TYPE_CHECKING:
     from poetry.core.packages.package import Package
+
+T = TypeVar("T", bound="Operation")
 
 
 class Operation(object):
@@ -41,13 +44,13 @@ class Operation(object):
     def format_version(self, package: "Package") -> str:
         return package.full_pretty_version
 
-    def skip(self, reason: str) -> "Operation":
+    def skip(self: T, reason: str) -> T:
         self._skipped = True
         self._skip_reason = reason
 
         return self
 
-    def unskip(self) -> "Operation":
+    def unskip(self: T) -> T:
         self._skipped = False
         self._skip_reason = None
 
--- a/src/poetry/puzzle/provider.py
+++ b/src/poetry/puzzle/provider.py
@@ -1,6 +1,7 @@
 import logging
 import os
 import re
+import tempfile
 import time
 import urllib.parse
 
@@ -12,6 +13,7 @@ from typing import Dict
 from typing import Iterator
 from typing import List
 from typing import Optional
+from typing import Set
 from typing import Union
 
 from cleo.ui.progress_indicator import ProgressIndicator
@@ -39,7 +41,6 @@ from poetry.repositories import Pool
 from poetry.utils.env import Env
 from poetry.utils.helpers import download_file
 from poetry.utils.helpers import safe_rmtree
-from poetry.utils.helpers import temporary_directory
 
 
 logger = logging.getLogger(__name__)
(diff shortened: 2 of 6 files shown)


src/poetry/puzzle/provider.py
```diff
@@ -725,6 +726,7 @@ def debug(self, message: str, depth: int = 0) -> None:
         if message.startswith("fact:"):
             if "depends on" in message:
                 m = re.match(r"fact: (.+?) depends on (.+?) \((.+?)\)", message)
+                assert m is not None
```
branchvincent, line 729: this may not be preferable

src/poetry/puzzle/solver.py
```diff
@@ -294,9 +295,10 @@ def __init__(
 
         if not previous:
             self.category = "dev"
-            self.groups = frozenset()
+            self.groups: FrozenSet[str] = frozenset()
             self.optional = True
         else:
+            assert dep is not None
```
branchvincent, line 301: another questionable `assert`

src/poetry/installation/operations/operation.py
```diff
@@ -1,10 +1,13 @@
 from typing import TYPE_CHECKING
 from typing import Optional
+from typing import TypeVar
 
 
 if TYPE_CHECKING:
     from poetry.core.packages.package import Package
 
+T = TypeVar("T", bound="Operation")
```
neersighted, line 9: For someone less experienced with Python typing -- what purpose does the change to `TypeVar` serve here?

src/poetry/puzzle/provider.py
```diff
@@ -352,11 +353,11 @@ def search_for_url(self, dependency: URLDependency) -> List[Package]:
     @classmethod
     def get_package_from_url(cls, url: str) -> Package:
         with temporary_directory() as temp_dir:
-            temp_dir = Path(temp_dir)
+            _temp_dir = Path(temp_dir)
```
neersighted, line 356: Slightly unrelated to the original goal here, but I wouldn't mind seeing this switched to stdlib `TemporaryDirectory` in this PR, which should also alleviate the need for the `_temp_dir` shuffle.

src/poetry/puzzle/provider.py
neersighted: I think we'd rather throw an exception here -- while it's a good insight from mypy that there is no error handling here and we implicitly make this assertion, I would like to avoid use of the `assert` keyword.

src/poetry/puzzle/solver.py
neersighted: Ditto

neersighted (changes requested):

src/poetry/puzzle/provider.py
branchvincent: makes sense, chose to `raise ValueError`. lmk if you want to see something different

src/poetry/puzzle/provider.py
branchvincent: done. note `TemporaryDirectory` still returns a `str`, but refactored a bit to remove the unnecessary intermediate var

src/poetry/installation/operations/operation.py
branchvincent: Great question! Prior to this change:

```
$ pre-commit run -a mypy
src/poetry/puzzle/transaction.py:58: error: Argument 1 to "append" of "list" has incompatible type "Operation"; expected "Union[Install, Uninstall, Update]"
Found 1 error in 1 file (checked 153 source files)
```

The function is typed as literally returning an instance of `Operation`, but in reality it returns `self`. The implications are that `mypy` gets confused for subclasses of `Operation`. See https://mypy.readthedocs.io/en/stable/kinds_of_types.html#the-type-of-class-objects and much prior https://github.com/python/mypy/issues/1212

neersighted (approved):

src/poetry/installation/operations/operation.py
neersighted: I see! The more you know!

Issue #4688, opened 2021-10-29
"Unable to find installation candidates for" packages cached from a LegacyRepository (`/simple`)


awilkins:
- [X] I am on the [latest](https://github.com/python-poetry/poetry/releases/latest) Poetry version.
- [X] I have searched the [issues](https://github.com/python-poetry/poetry/issues) of this repo and believe that this is not a duplicate.
   - Caveat : I have insufficient detail to decide if #4679 is a dupe of this issue, but it is possible since it has similar symptoms
- [X] If an exception occurs when executing a command, I executed it again in debug mode (`-vvv` option).

- CodeBuild image `aws/codebuild/standard:5.0` (Ubuntu 20.04)
- Poetry 1.1.11 : as installed by `pip`
  - Yes, this is not the recommended way to install Poetry
  - However, the recommended way ought to have the same issue (installs same version of `cachecontrol`, at present)
  - Confirmed that using `install-poetry.py` on a fresh install reproduces this issue

## Issue

Have been debugging this inside a container based on `aws/codebuild/standard:5.0` (since this is the target build container for our CI pipelines).

Pipelines are installing poetry via `pip` and failing. Build works fine on my workstation, even after a `poetry self update`, which doesn't seem to update dependencies if they already meet the version constraints (like `pip`).

After a fresh install of Poetry 1.1.11 via `install-poetry.py`, now also having similar issues (with different symptoms but I presume the same cause) locally.

Issues are resolved by activating Poetry's venv and running

```
pip install cachecontrol==0.12.6
```

This only applies to packages installed (from a Nexus repo) via the `/simple` interface, which uses the LegacyRepository class.

## Suspected issue

Working on a deeper diagnosis, but thought I'd get the bug report opened.

`cachecontrol` released `0.12.7` as of 12 hours ago, clearly there is a bug or change in behaviour (aka - bug, for a patch version bump) that breaks Poetry for this use case.

The issue seems to be caching of the HTML metadata from the `/simple` type repo ; while the cache file on disk is successfully updated each time it is fetched, the cache-wrapped response object for the metadata returns an empty byte buffer to the program. This results in the `Page` object not finding any links and `get_links_for_package` failing.

In a tree with a lockfile and in an environment with an empty cache, this means you will be able to install that package once and once only - the first time caches the metadata, which then subsequently triggers this bug and prevents further access to the metadata. Sadly all my pipelines build multiple packages with common dependencies.

Now going to look and see what the problem in `cachecontrol` is.

[...]

PR #4766, opened 2021-11-15
chore: update cachecontrol to ^0.12.9

neersighted:
Update CacheControl to ^0.12.9 in light of bug fixes and #4688.

Also, re-organized the list of dev-dependencies and alphabetize non-`poetry-core` deps.


abn (approved):
Approving this for now, but we should next time split out version updates and reordering etc.

PR #4784, opened 2021-11-17
chore: clean up unused imports

neersighted:


src/poetry/repositories/legacy_repository.py
```diff
@@ -15,7 +15,6 @@
 from typing import Optional
 from urllib.parse import quote
 
-import requests
```
abn, line 18: Hmm there is a 

```
except requests.HTTPError as e:
```

so technically its not unused, I suspect we are simply not seeing an error because `requests.HTTPError` gets imported as a side effect of the `auth` and `exception` import.

abn (approved):

PR #4785, opened 2021-11-17
ci: coverage and SonarCloud

neersighted:
Add coverage using SonarCloud (for now, we can switch to Codecov if needed). Additionally, unify the invocation of the test suite and add pytest-clarity for local use.

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -28,7 +28,7 @@ test_task:
     - poetry config virtualenvs.in-project true
   test_script:
     - poetry install
-    - poetry run pytest -p no:sugar --junitxml=junit.xml
+    - poetry run pytest -p no:clarity -p no:sugar --junitxml=junit.xml
   on_failure:
     annotate_failure_artifacts:
       path: junit.xml
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -24,10 +24,10 @@ test_task:
   env_script:
     - echo "PATH=/.local/bin:${PATH}" >> $CIRRUS_ENV
   poetry_script:
-    - curl -sL https://install.python-poetry.org | $PYTHON - -y
+    - curl -sL https://install.python-poetry.org | $PYTHON - -y --git https://github.com/python-poetry/poetry
     - poetry config virtualenvs.in-project true
   test_script:
-    - poetry install
+    - poetry install --with=dev,ci
     - poetry run pytest -p no:clarity -p no:sugar --junitxml=junit.xml
   on_failure:
     annotate_failure_artifacts:
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -28,7 +28,7 @@ test_task:
     - poetry config virtualenvs.in-project true
   test_script:
     - poetry install
-    - poetry run pytest -q --junitxml=junit.xml tests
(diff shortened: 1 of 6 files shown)


.cirrus.yml
```diff
@@ -28,7 +28,7 @@ test_task:
     - poetry config virtualenvs.in-project true
   test_script:
     - poetry install
-    - poetry run pytest -q --junitxml=junit.xml tests
+    - poetry run python -m pytest -vv -p no:clarity -p no:sugar --junitxml=junit.xml tests/
```
abn, line 31: Is there a reason why we do this? ie. are we expection pytest  from outside the env to be used somehow?

.github/workflows/main.yml
```diff
@@ -79,8 +78,8 @@ jobs:
       - name: Install dependencies
         run: poetry install
 
-      - name: Install pytest plugin
-        run: poetry run pip install pytest-github-actions-annotate-failures
-
       - name: Run pytest
-        run: poetry run python -m pytest -p no:sugar -q tests/
+        run: poetry run python -m pytest -vv --color=yes -p no:clarity -p no:sugar --cov=src/poetry --cov-report xml tests/
```
abn, line 82: we can add coverage run config and pytest args into `pyproject.toml` right?

pyproject.toml
```diff
@@ -54,16 +54,20 @@ entrypoints = "^0.3"
...
 [tool.poetry.dev-dependencies]
+tox = "^3.18"
 pytest = "^6.2"
 pytest-cov = "^2.8"
 pytest-mock = "^3.5"
-pre-commit = { version = "^2.6", python = "^3.6.1" }
-tox = "^3.18"
 pytest-sugar = "^0.9"
+pytest-clarity = "^1.0.1"
+pytest-github-actions-annotate-failures = "^0.1.5"
```
abn, line 65: should we move this to an extra? and make it optional?

abn (changes requested):

.cirrus.yml
neersighted: No -- it's done for consistency with the Github action.

.github/workflows/main.yml
neersighted: Yes, but it significantly increases the execution time of the test suite for local testing, so I feel that it's better done here. Also, we only want to run coverage on Github Actions as Codecov does not currently have an uploader that runs on FreeBSD.

(it's more complicated than that -- technically we can install node.js and their typescript uploader on FreeBSD -- there's just no ready-to-run bundle)

pyproject.toml
neersighted: We could certainly. The code is a no-op unless `$GITHUB_RUNNER` is in the environment, but it will pull in a few more dependencies. I'm open to going either way.

abn: I'd either take that out of pyproject entirely and install in the action run, however that might mean we might have incorrect dependencies. Hence I am leaning more towards a new group (now that we have groups) rather than extras come to think of it :)

Secrus:
@neersighted what is the status of this?

neersighted:
> @neersighted what is the status of this?

Needs a rebase, I'll try to dust it off in a day or two. Also needs to me to do some stuff on the Cirrus CI web UI to validate it works.

PR #4786, opened 2021-11-17
ci: randomize testsuite using pytest-randomly

neersighted:
Randomize the test suite to prevent bugs being masked by execution order -- this already caught one bug in Cleo. Additionally, bump and use importlib_metadata so we have the same set of features across all supported Python versions (i.e. 3.7-3.10).


branchvincent:
fyi with #5766, this is now unblocked

abn (approved):

Issue #4776, opened 2021-11-16
Code Quality: Additional lints in this repository (help wanted!)


neersighted:
We're doing a bunch of code quality cleanup, and as part of that effort I'd like to implement some additional lints/coding conventions. There will be a lot of manual changes here, so I think these will be some items to have multiple people tackle. 

In no particular order:

- ~Add `ban-relative-imports = true` to `.flake8` and remove relative imports from the repo.~
- ~Add `flake8-type-checking==1.0.4` to `.pre-commit-config.yaml` with `enable-extensions = TC, TC2` in `.flake8`, and rework our type annotations for zero runtime cost using string constants. When we drop Python 3.6 support, pyupgrade will be used to automatically switch to `from __future__ import annotations` and unquote the annotations.~
- ~Add `flake8-use-fstring==1.3` to `.pre-commit-config.yaml` and fix all the use of `.format()` on string constants in the repo. Start at level 0, and try level 1 if possible. Level 2 is likely useless in this repo (I think we will always use `.format()` on string variables), but you could prove me wrong.~
- ~Remove `E501` from `.flake8` -- currently, black leaves our long comments, docstrings, and string constants alone -- let's hand-format them to comply with the line length limit.~

Please call dibs in this issue to prevent duplicated effort.

(suggestions for additional code quality improvements/tools to enforce them are also welcomed here)

neersighted:
(suggestions for additional code quality improvements/tools to enforce them are also welcomed here)

neersighted:
Here are some other tools I have glanced at, but not yet fully investigated for usefulness (of the tool itself, or the change):

* https://github.com/best-doctor/flake8-class-attributes-order
* https://github.com/wemake-services/flake8-eradicate
* https://github.com/MartinThoma/flake8-simplify
* https://github.com/m-burst/flake8-pytest-style
* https://github.com/PyCQA/pep8-naming
* https://github.com/pycqa/flake8-docstrings
* https://github.com/gforcada/flake8-builtins

neersighted:
This comment tracks the implementation status of new lints:

* https://github.com/sco1/flake8-annotations @ 4775

branchvincent:
> * Add `ban-relative-imports = true` to `.flake8` and remove relative imports from the repo.

happy to take this one!

danieleades:
i'll take `pep8-naming`

wagnerluis1982:
[...]

PR #4798, opened 2021-11-18
Use flake8 simplify

radoering:
# Pull Request Check List

Relates-to: #4776

I plan to have one commit per violation code, so you can easily check if you like all of them.

Before I tackle [SIM300](https://github.com/MartinThoma/flake8-simplify/issues/16) Yoda conditions (which is by far the most frequent violation): You also think that this should be fixed, don't you? (Just want to make sure because I know some people who like Yoda conditions). Hopefully, most of these conditions can be replaced with some sort of regex, I still have to come up with. 😀

### DIFF
--- a/.flake8
+++ b/.flake8
@@ -10,6 +10,8 @@ extend-ignore =
     E501,
     # E203: Whitespace before ':' (pycqa/pycodestyle#373)
     E203,
+    # SIM106: Handle error-cases first
+    SIM106,
 per-file-ignores =
     # F401: Module imported by unused (non-implicit modules)
     # TC002: Move third-party import '...' into a type-checking block
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -36,6 +36,7 @@ repos:
           - flake8-comprehensions==3.7.0
           - flake8-no-pep420==1.2.0
           - flake8-quotes==3.3.1
+          - flake8-simplify==0.14.2
           - flake8-tidy-imports==4.5.0
           - flake8-type-checking==1.1.0
           - flake8-typing-imports==1.11.0
--- a/get-poetry.py
+++ b/get-poetry.py
@@ -125,8 +125,8 @@ def is_decorated():
     if platform.system().lower() == "windows":
         return (
             os.getenv("ANSICON") is not None
-            or "ON" == os.getenv("ConEmuANSI")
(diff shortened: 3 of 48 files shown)


finswimmer:
> Before I tackle [SIM300](https://github.com/MartinThoma/flake8-simplify/issues/16) Yoda conditions (which is by far the most frequent violation): You also think that this should be fixed, don't you? 

Yes, should be fixed :)

neersighted:
The initial work looks good, and the list of lints at the attached repo also look very helpful for ensuring readable code. As long as the lint isn't too aggressive and the error messages are helpful (don't want to overwhelm new contributors), implementing this looks good to me.

radoering:
I fixed all violations except for `SIM106: Handle error-cases first` because I think in many cases the code would be more difficult to understand resp. maintain after fixing `SIM106` than before fixing it.

If the violation is in a simple `if`-`else` it might be ok. One could debate whether an additional `not` or an `!=` instead of `==` in the condition is worth the exception being thrown first, but if the violation is in an `if`-`elif`-`...`-`else` chain, it becomes really ugly (in my opinion). Let's take the violation at [src/poetry/puzzle/solver.py:298](https://github.com/python-poetry/poetry/blob/cbbd92ceb5938a43a1f4666cdaf9599c74650442/src/poetry/puzzle/solver.py#L294-L303) for example:

```
if not previous:
    # do something
elif dep:
    # do another thing
else:
    raise ValueError
```

In order to handle the error first, I have to add:

```
if not (not previous or dep):
    raise ValueError
```

which could be simplified to

```
if previous and not dep:
    raise ValueError
```

But what happens if an additional case has to be added (another `elif`)? The condition of this case has to be merged into the error handling condition! Thus, you must not forget to do that and you have to be careful not to make a mistake doing it. This seems like some sort of complicated code duplication to me...

In my opinion, `SIM106` should be added to the ignore list.

@finswimmer, @neersighted: What do you think?

neersighted:
> @neersighted: What do you think?

I do think that the general structure proposed by SIM106 is more readable, but I agree that long-term maintainability may be compromised if the pattern is not easily internalized.

I'm fine with disabling the check for now, and revisiting later, personally. All the other changes on this PR look like pretty unambiguous wins for readability and maintainability.

neersighted (approved):

neersighted (approved):

PR #4799, opened 2021-11-18
Re-order pre-commit hooks

sondrelg:
Hi! 

There is no issue for this, so feel free to discard the PR if not wanted.

I just took a look at the pre-commit config and noticed that you're currently running some static checkers (flake8 and yesqa) before code formatters like black and isort. From my own experience I think you can actually eliminate a few false positives if you run all your formatters first - ahead of static checkers 👍 For me, black in particular tends to fix a lot of flake8 errors related to code formatting.

Let me know if you prefer another order.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -18,29 +18,6 @@ repos:
       - id: debug-statements
       - id: check-docstring-first
 
-  - repo: https://github.com/pre-commit/pygrep-hooks
-    rev: v1.9.0
-    hooks:
-      - id: python-check-mock-methods
-      - id: python-use-type-annotations
-      - id: python-check-blanket-type-ignore
-      - id: python-check-blanket-noqa
-
-  - repo: https://github.com/asottile/yesqa
-    rev: v1.3.0
-    hooks:
-      - id: yesqa
-        additional_dependencies: &flake8_deps
-          - flake8-broken-line==0.4.0
-          - flake8-bugbear==21.9.2
-          - flake8-comprehensions==3.7.0
-          - flake8-no-pep420==1.2.0
-          - flake8-quotes==3.3.1
-          - flake8-tidy-imports==4.5.0
-          - flake8-type-checking==1.1.0
-          - flake8-typing-imports==1.11.0
-          - pep8-naming==0.12.1
-
   - repo: https://github.com/asottile/pyupgrade
     rev: v2.29.0
     hooks:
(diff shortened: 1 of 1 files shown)


.pre-commit-config.yaml
```diff
@@ -64,6 +41,29 @@ repos:
     hooks:
       - id: black
 
+  - repo: https://github.com/pre-commit/pygrep-hooks
```
neersighted, line 44: I don't know there's any reason to move `pygrep` as it's just a static set of regexes. Feel free to convince me otherwise (if there's a strong reason behind this). I'm fine with moving yesqa lower, however.

neersighted (changes requested):

.pre-commit-config.yaml
neersighted: The flipside of this is I think running `yesqa` after `flake8` might cause less surprising behaviors. I'd be interested in that change on that PR.

sondrelg: Ah yes, I see that the diff seems to indicate that I've moved `pygrep` and `flake8` down, but really what I've actually done is move `isort` and `black` up, since they are not static.

The ordering was:

- pre-commit-hooks (changes code)
- pygrep (static)
- yesqa (static)
- pyupgrade (changes code)
- isort (changes code)
- black (changes code)
- flake8 (static)

and I changed it to

- pre-commit-hooks (changes code)
- pyupgrade (changes code)
- isort (changes code)
- black  (changes code)
- pygrep (static)
- yesqa (static)
- flake8 (static)

The order of the code-changing hooks relative to themselves is the same, and the same goes for static checks.

Does that make sense? 🙂 The github diff viewer doesn't really do this justice so my bad for not highlighting it earlier

sondrelg: Do you want me to change it to this? (flake8 running before yesqa)

pre-commit-hooks (changes code)
pyupgrade (changes code)
isort (changes code)
black (changes code)
pygrep (static)
flake8 (static)
yesqa (static)

neersighted: yesqa is actually a code changer. It removes any noqa that does not change the output of flake8. So by your logic here it should probably go before isort but after pyupgrade?

sondrelg: Oh, I see! Out of all of the hooks, yesqa is the one I don't use myself. So maybe:

```
1. pre-commit-hooks
2. pyupgrade
3. yesqa 
4. isort 
5. black 
```

Would that make sense, or would another order make more sense to you?

Do you still agree with the premise that putting code altering hooks before static hooks (flake8 in particular) is a good thing, to eliminate potential false positives?

neersighted: I have to admit I'm not entirely convinced that this matters in practice, but I'm happy to defer to your greater experience. I think the order described above should be just fine and I'd be happy to accept the change.

sondrelg: That's fair enough. I'll close this then for now 🙂

PR #4805, opened 2021-11-20
Add missing type hints & Force type hints - Part 3

finswimmer:
This PR is the final part of a series which add missing type hints to both package code and test code. Furthermore flake-annotations is added as a flake8 plugin to enforce type hints in the future.

### DIFF
--- a/.flake8
+++ b/.flake8
@@ -3,6 +3,7 @@ min_python_version = 3.6.0
 max-line-length = 88
 ban-relative-imports = true
 inline-quotes = double
+mypy-init-return = True
 enable-extensions = TC, TC2
 type-checking-exempt-modules = typing, typing-extensions
 extend-ignore =
@@ -12,10 +13,14 @@ extend-ignore =
     E203,
     # SIM106: Handle error-cases first
     SIM106,
+    ANN101,
+    ANN102,
 per-file-ignores =
     # F401: Module imported by unused (non-implicit modules)
     # TC002: Move third-party import '...' into a type-checking block
     __init__.py:F401,TC002,
+    tests/test_*:ANN201
+    tests/**/test_*:ANN201
 extend-exclude =
     # Frozen and not subject to change in this repo:
     get-poetry.py,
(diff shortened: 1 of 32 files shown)


.flake8
```diff
@@ -12,10 +13,14 @@ extend-ignore =
     E203,
     # SIM106: Handle error-cases first
     SIM106,
+    ANN101,
```
neersighted, line 16: Let's add comments here explaining what each error is.

.flake8
```diff
@@ -12,10 +13,14 @@ extend-ignore =
...
+    ANN102,
 per-file-ignores =
     # F401: Module imported by unused (non-implicit modules)
     # TC002: Move third-party import '...' into a type-checking block
     __init__.py:F401,TC002,
+    tests/test_*:ANN201
```
neersighted, line 22: Above/like the above comments.

.flake8
```diff
@@ -3,6 +3,7 @@ min_python_version = 3.6.0
 max-line-length = 88
 ban-relative-imports = true
 inline-quotes = double
+mypy-init-return = True
```
neersighted, line 6: Let's add a comment here explaining which plugin this corresponds to / maybe a doc link? Also let's go for a lower case `true` since that's what we've done in the rest of the file.

.pre-commit-config.yaml
```diff
@@ -41,6 +41,7 @@ repos:
           - flake8-type-checking==1.1.0
           - flake8-typing-imports==1.11.0
           - pep8-naming==0.12.1
+          - flake8-annotations
```
neersighted, line 44: Let's lock this to a version and alphabetize it.

neersighted (changes requested):
Minor nits over the lint config -- overall LGTM. Today I learned about the Pytest parameterized kwargs splitting a string on "," -- I've only ever used the list syntax myself.

.flake8
finswimmer: > Minor nits over the lint config 
Added the comments as suggested. Thanks for reviewing :+1:

neersighted (approved):

PR #4806, opened 2021-11-20
Use flake8-eradicate

radoering:
Relates-to: #4776

Very low hanging fruit, but perhaps too little benefit or maybe just an investment in the future.

There seems to be no commented out code in poetry (or `eradicate` can't find it).

`eradicate` only finds 5 false positives in [this comment](https://github.com/python-poetry/poetry/blob/5c328a3637b0efd93d6581539d2f3a3d062fc03d/src/poetry/puzzle/provider.py#L534-L551):

```
src/poetry/puzzle/provider.py:540:1: E800 Found commented out code
src/poetry/puzzle/provider.py:541:1: E800 Found commented out code
src/poetry/puzzle/provider.py:544:1: E800 Found commented out code
src/poetry/puzzle/provider.py:550:1: E800 Found commented out code
src/poetry/puzzle/provider.py:551:1: E800 Found commented out code
```

When extending the whitelist, you have to ask yourself how specific or general you want the regex to be.

### DIFF
--- a/.flake8
+++ b/.flake8
@@ -5,6 +5,7 @@ ban-relative-imports = true
 inline-quotes = double
 enable-extensions = TC, TC2
 type-checking-exempt-modules = typing, typing-extensions
+eradicate-whitelist-extend = ^-.*;
 extend-ignore =
     # E501: Line too long (FIXME: long string constants)
     E501,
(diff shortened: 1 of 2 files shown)


neersighted:
It seems like it could be useful to aid code review and/or help prevent new contributors from having more things to fix on a PR. That being said, I haven't seen a lot of commented out code during review. Thoughts, @finswimmer and @abn?

branchvincent:
it may be worth exploring other options for finding dead code, such as [`vulture`](https://github.com/jendrikseipp/vulture) or [`dead`](https://github.com/asottile/dead) (both of which seem to find unused code here but also false positives)

neersighted (approved):

Issue #4821, opened 2021-11-24
Random locking failure due to CacheControl returning an empty bytes response


j-martin:
- [x] I am on the [latest](https://github.com/python-poetry/poetry/releases/latest) Poetry version.
- [x] I have searched the [issues](https://github.com/python-poetry/poetry/issues) of this repo and believe that this is not a duplicate.
- [ ] If an exception occurs when executing a command, I executed it again in debug mode (`-vvv` option).

- **OS version and name**: macOS Monterey
- **Poetry version**: 1.1.11

## Issue

When adding new packages or locking we observed random failures where the JSON response is empty (as in `b""`).

Through https://github.com/python-poetry/poetry/pull/4717 and https://github.com/pypa/warehouse/issues/10387 we (myself and @kapilt) found that CacheControl is returning empty data even though there should be serialized JSON object.

Removing session CacheControl seems to solve the issue.

kapilt:
fwiw, i haven't actually hit this in a while using latest poetry releases.. i think the actual issue was solved on the pypi cdn side.

PR #4717, opened 2021-11-05
pypi, retry when receiving a 200 with empty body

j-martin:
For some reason, pypi is returning a 200 and an empty body from time to time. The next time you query the same endpoint it returns a proper JSON object. This is not fixing a bug in poetry, but working around the issue.

This was resulting in issue where anything operation modifying the lock file would randomly fail due to `JSONDecodeError`. This is unrelated to https://github.com/python-poetry/poetry/pull/4592

# Pull Request Check List

Resolves: Issue not created.

- [ ] Added **tests** for changed code. <=== N/A
- [ ] Updated **documentation** for changed code. <=== N/A

PR #4823, opened 2021-11-24
workaround pypi empty responses by disabling cache control

kapilt:
# Pull Request Check List

Resolves: #4821 
Alternative to: #4822 and #4717 

This is a work-around fix for https://github.com/pypa/warehouse/issues/10387 . Thanks @j-martin for also looking at this.

This disables cache control for package metadata due to issues observed that pypi can return empty responses. The package artifact is still maintained. The diff is a little larger due to need to keeping type checker happy and removing tests assuming cache control usage.

- [ x ] Updated **documentation** for changed code.

### DIFF
--- a/src/poetry/repositories/legacy_repository.py
+++ b/src/poetry/repositories/legacy_repository.py
@@ -10,6 +10,7 @@ from pathlib import Path
 from typing import TYPE_CHECKING
 from typing import Iterator
 from typing import List
+from typing import Mapping
 from typing import Optional
 from urllib.parse import quote
 
@@ -40,8 +41,6 @@ from poetry.utils.patterns import wheel_file_re
 
 
 if TYPE_CHECKING:
-    import requests.structures
-
     from poetry.core.packages.dependency import Dependency
 
 with warnings.catch_warnings():
@@ -62,9 +61,7 @@ class Page:
         ".tar",
     ]
 
-    def __init__(
-        self, url: str, content: bytes, headers: requests.structures.CaseInsensitiveDict
-    ) -> None:
+    def __init__(self, url: str, content: bytes, headers: Mapping) -> None:
         if not url.endswith("/"):
             url += "/"
 
--- a/src/poetry/repositories/legacy_repository.py
+++ b/src/poetry/repositories/legacy_repository.py
@@ -8,8 +8,6 @@ from collections import defaultdict
 from html import unescape
 from pathlib import Path
 from typing import TYPE_CHECKING
-from typing import Any
-from typing import Dict
 from typing import Iterator
 from typing import List
 from typing import Optional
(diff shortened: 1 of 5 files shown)


src/poetry/repositories/legacy_repository.py
```diff
@@ -62,7 +62,9 @@ class Page:
         ".tar",
     ]
 
-    def __init__(self, url: str, content: str, headers: Dict[str, Any]) -> None:
+    def __init__(
+        self, url: str, content: bytes, headers: requests.structures.CaseInsensitiveDict
```
abn, line 66: I would prefer that we stuck to `Dict` or `Mappting`. We do want to drop requests at some point.

src/poetry/repositories/pypi_repository.py
```diff
@@ -72,18 +71,11 @@ def __init__(
         )
 
         self._cache_control_cache = FileCache(str(release_cache_dir / "_http"))
-        self._session = CacheControl(
```
abn, line 75: I am not a 100% convinced this is the right solution for this. I'd much rather have e-tag used in the cache key.

Skipping cache for empty responses or invalid cache values, should be handled by gracefully handling failure - ie. clear the invalid cache and retry or fallback.

j-martin: I guess that's the naive approach [here](https://github.com/python-poetry/poetry/pull/4717).

tests/fixtures/git/github.com/demo/demo/demo.egg-info/PKG-INFO
```diff
@@ -6,7 +6,8 @@ Home-page: https://github.com/demo/demo
 Author: Sébastien Eustace
 Author-email: sebastien@eustace.io
 License: MIT
-Description: UNKNOWN
 Platform: UNKNOWN
-Provides-Extra: bar
 Provides-Extra: foo
+Provides-Extra: bar
+
```
neersighted, line 12: This is an invalid change that came along for the ride. Still need to find what test is causing this instability.

neersighted (changes requested):

src/poetry/repositories/pypi_repository.py
neersighted: That approach is at the request level and not really at the cache level. I think an intermediate approach (e.g. layered on top of CacheControl and factoring in ETAGS) is the way to go. The caching is an important optimization, and adding a retry system at the high-level HTTP (e.g. `.get()`) API seems more a hack than a fix.

tests/fixtures/git/github.com/demo/demo/demo.egg-info/PKG-INFO
kapilt: ack, yeah not sure where that line came from, will remove

src/poetry/repositories/legacy_repository.py
kapilt: switched to mapping, mypy was complaining about the dict.

kapilt:
i haven't actually hit this in a while using latest poetry release, closing.

PR #4824, opened 2021-11-24
Replace the export command by the export plugin

sdispater:
This PR replaces the existing `export` command by the `poetry-export-plugin`.

It also makes changes in the way plugins are handled via the respective `plugin` commands that is simpler and more reliable. Note that there is currently an issue in the installer/solver that removes installed packages if they are not required by Poetry or any plugin. I am still working on a fix that will be done in a subsequent PR.

## Pull Request Check List


src/poetry/installation/installer.py
```diff
@@ -355,7 +355,7 @@ def _do_install(self, local_repo: Repository) -> int:
         return self._execute(ops)
 
     def _write_lock_file(self, repo: Repository, force: bool = True) -> None:
-        if force or (self._update and self._write_lock):
+        if (force or (self._update and self._write_lock)) and not self._dry_run:
```
neersighted, line 358: Looks like a bug fix that should get a separate commit/PR?

src/poetry/packages/locker.py
```diff
@@ -11,6 +11,7 @@
 from typing import Iterable
```
neersighted, line 11: Changes in this file look good, but I think we should break them out into a separate commit.

src/poetry/packages/locker.py
```diff
@@ -463,15 +449,7 @@ def _get_content_hash(self) -> str:
...
-    def _get_lock_data(self) -> "TOMLDocument":
-        if not self._lock.exists():
-            raise RuntimeError("No lockfile found. Unable to read locked packages")
-
-        try:
-            lock_data = self._lock.read()
-        except TOMLKitError as e:
-            raise RuntimeError(f"Unable to read the lock file ({e}).")
-
+    def _verify_lock_data(self, lock_data: dict) -> NoReturn:
```
neersighted, line 452: `NoReturn` is not correct -- it signifies a method that will never return (i.e. does not terminate, always throws an exception, etc). If there is no `return` statement the correct type is still `None` (as Python will add an implicit return at the end of the function).

src/poetry/packages/locker.py
```diff
@@ -591,9 +567,7 @@ def _dump_package(self, package: Package) -> dict:
             if package.source_type in ["file", "directory"]:
                 # The lock file should only store paths relative to the root project
                 url = Path(
-                    os.path.relpath(
-                        Path(url).as_posix(), self._lock.path.parent.as_posix()
-                    )
+                    os.path.relpath(Path(url).as_posix(), self._root.as_posix())
```
neersighted, line 570: This immediately looks way cleaner, yay!

src/poetry/puzzle/provider.py
```diff
@@ -188,6 +188,7 @@ def search_for_vcs(self, dependency: "VCSDependency") -> List["Package"]:
 
         dependency._constraint = package.version
         dependency._pretty_constraint = package.version.text
+        dependency._source_resolved_reference = package.source_resolved_reference
```
neersighted, line 191: Unless I'm missing something, this line got duplicated. Oops?

tests/fixtures/git/github.com/demo/demo/demo.egg-info/PKG-INFO
```diff
@@ -6,7 +6,9 @@ Home-page: https://github.com/demo/demo
 Author: Sébastien Eustace
 Author-email: sebastien@eustace.io
 License: MIT
-Description: UNKNOWN
 Platform: UNKNOWN
-Provides-Extra: bar
 Provides-Extra: foo
+Provides-Extra: bar
+
+UNKNOWN
```
neersighted, line 13: Some test suite instability is causing changes to fixtures that came along for the ride. We need to back these changes out.

tests/fixtures/git/github.com/demo/demo/demo.egg-info/SOURCES.txt
```diff
@@ -4,4 +4,4 @@ demo.egg-info/PKG-INFO
 demo.egg-info/SOURCES.txt
 demo.egg-info/dependency_links.txt
 demo.egg-info/requires.txt
-demo.egg-info/top_level.txt
+demo.egg-info/top_level.txt
```
neersighted, line 7: Above.

tests/fixtures/git/github.com/demo/demo/demo.egg-info/dependency_links.txt
```diff
@@ -0,0 +1 @@
+
```
neersighted, line 1: Above.

src/poetry/factory.py
```diff
@@ -126,10 +134,15 @@ def create_config(cls, io: Optional["IO"] = None) -> Config:
         return config
 
     @classmethod
-    def configure_sources(
-        cls, poetry: Poetry, sources: List[Dict[str, str]], config: Config, io: "IO"
-    ) -> None:
-        for source in sources:
+    def create_pool(cls, config: "Config", io: Optional["IO"] = None) -> "Pool":
```
neersighted, line 137: These changes look good. I'd love to see a more detailed commit message, and maybe the addition of `PluginCommandMixin` broken out into a separate commit.

src/poetry/factory.py
```diff
@@ -74,10 +74,18 @@ def create_poetry(
...
+        config.merge(
+            {
+                "sources": {
+                    source["name"]: source
+                    for source in poetry.local_config.get("source", [])
+                }
+            }
         )
 
+        # Configuring sources
```
neersighted, line 86: These comments are duplicates, and the `merge()` call is a bit overnested. I think we should break the generator expression out into a separate line (e.g. L98 in `add.py`).

tests/console/commands/plugin/conftest.py
```diff
@@ -32,26 +25,13 @@ def installed() -> InstalledRepository:
...
-    ) -> None:
-        pool = Pool()
-        pool.add_repository(repo)
-        poetry.set_pool(pool)
-
-    return _configure_sources
-
-
 @pytest.fixture(autouse=True)
 def setup_mocks(
```
neersighted, line 29: Nice to see this monster get simplified a bit.

src/poetry/console/commands/plugin/plugin_command_mixin.py
```diff
@@ -0,0 +1,125 @@
...
+
+            version = entry_point.distro.version
+            if canonicalize_name(entry_point.distro.name) in existing_plugins:
+                version = existing_plugins[canonicalize_name(entry_point.distro.name)]
+
+            root_package.add_dependency(
+                Factory.create_dependency(entry_point.distro.name, version)
+            )
+
+        root_package.python_versions = ".".join(str(v) for v in env.version_info[:3])
```
neersighted, line 74: This pattern is so common that I think we should have a getter (could even be cached by `functools`) in `Env`.

src/poetry/console/commands/plugin/plugin_command_mixin.py
```diff
@@ -0,0 +1,125 @@
...
+        )
+
+    def update(
+        self,
+        env: "Env",
+        root_package: "ProjectPackage",
+        io: "IO",
+        whitelist: List[str],
+    ) -> int:
+        import os
```
neersighted, line 93: This particular import is redundant. See other comments about scoped imports as a whole.

src/poetry/console/commands/plugin/plugin_command_mixin.py
```diff
@@ -0,0 +1,125 @@
...
+
+    from poetry.core.packages.project_package import ProjectPackage
+    from poetry.utils.env import Env
+
+
+class PluginCommandMixin:
+    def create_env_package(
+        self, env: "Env", existing_plugins: dict
+    ) -> "ProjectPackage":
+        from poetry.core.packages.project_package import ProjectPackage
```
neersighted, line 24: Not sure if the scoped imports here are necessary, or wise. It feels like a premature optimization that adds significant mental overhead and complexity. See comments elsewhere in the review.

src/poetry/console/commands/plugin/remove.py
```diff
@@ -44,24 +47,69 @@ def handle(self) -> int:
...
+
+            if plugin in existing_plugins:
+                del existing_plugins[plugin]
+
+            removed_plugins.append(plugin)
+
+        if not removed_plugins:
+            return 1
+
+        _root_package = root_package
```
neersighted, line 95: This shuffle is a bit gross -- can we clean up the variable names to be better here?

src/poetry/factory.py
```diff
@@ -126,10 +134,15 @@ def create_config(cls, io: Optional["IO"] = None) -> Config:
...
 
     @classmethod
-    def configure_sources(
-        cls, poetry: Poetry, sources: List[Dict[str, str]], config: Config, io: "IO"
-    ) -> None:
-        for source in sources:
+    def create_pool(cls, config: "Config", io: Optional["IO"] = None) -> "Pool":
+        from poetry.repositories.pool import Pool
+
+        if io is None:
```
neersighted, line 140: This feels a little unnecessary -- is it a big burden to require passing in `IO`? The caller could always provide a `NullIO`.

src/poetry/factory.py
```diff
@@ -126,10 +134,15 @@ def create_config(cls, io: Optional["IO"] = None) -> Config:
         return config
 
     @classmethod
-    def configure_sources(
-        cls, poetry: Poetry, sources: List[Dict[str, str]], config: Config, io: "IO"
-    ) -> None:
-        for source in sources:
+    def create_pool(cls, config: "Config", io: Optional["IO"] = None) -> "Pool":
+        from poetry.repositories.pool import Pool
```
neersighted, line 138: Really not sure we should be adding scoped imports at all.

tests/installation/test_installer.py
```diff
@@ -306,6 +310,76 @@ def test_run_update_after_removing_dependencies(
...
+
+    package.add_dependency(Factory.create_dependency("A", "~1.0"))
+    package.add_dependency(Factory.create_dependency("B", "~1.1"))
+    package.add_dependency(Factory.create_dependency("C", "~1.2"))
+
+    installer.update(True)
+    installer.whitelist(["c"])
+    installer.run()
+
+    assert 0 == installer.executor.installations_count
```
neersighted, line 378: These are Yoda conditions -- let's swap operands so they read naturally.

tests/installation/test_installer.py
```diff
@@ -27,7 +27,7 @@
 from poetry.installation import Installer as BaseInstaller
 from poetry.installation.executor import Executor as BaseExecutor
 from poetry.installation.noop_installer import NoopInstaller
-from poetry.packages import Locker as BaseLocker
+from poetry.packages.locker import Locker as BaseLocker
```
neersighted, line 30: I feel like this import is a bit confusing since there is now a `BaseLocker` in that file. Can we clean it up at all?

tests/console/commands/test_add.py
```diff
@@ -357,7 +357,7 @@ def test_add_directory_constraint(
     mocker: "MockerFixture",
 ):
     p = mocker.patch("pathlib.Path.cwd")
-    p.return_value = Path(__file__).parent
+    p.return_value = Path(__file__).parent.parent.parent / "fixtures" / "git"
```
neersighted, line 360: This parent calls are a bit messy, but also I think we should address them as part of a 'fixtures-to-get-fixtures' refactor to the test suite.

src/poetry/factory.py
neersighted, line 138: This delayed import really seems useless -- the import will run almost right away anything. I think this pattern tends to be a premature optimization that is hard to maintain and that we should address import overhead with a higher-level/more unified approach if it's an issue.

neersighted (changes requested):
Looks good overall, but it's a very large monolithic change and I think we need to break it up for ease of review, testing, and rollback. Moderate feedback on design patterns, idioms, and code smells. There's some very easy to fix code quality issues that `pre-commit` should break down easily enough.

My main overall concern is continuing a pattern of delayed/scoped imports. I am generally of the opinion that they are adding significant mental overhead and potential for breakage and are likely to become 'stale' as code is refactored. I would much rather add explicit at-the-start-of-the-file imports and consider a project-wide solution to startup time if it's getting out of control.

Solutions might be PyOxidizer (through the official installer), a Mercurial-style lazy-load-with-exceptions-list, or something else.

Check out these articles for more ideas (and be warned that my knowledge of this field is nascent):
https://gregoryszorc.com/blog/2019/01/10/what-i%27ve-learned-about-optimizing-python/
https://snarky.ca/lazy-importing-in-python-3-7/

src/poetry/console/commands/plugin/plugin_command_mixin.py
```diff
@@ -0,0 +1,125 @@
...
+        import os
+
+        from pathlib import Path
+
+        from poetry.factory import Factory
+        from poetry.installation.installer import Installer
+        from poetry.repositories.installed_repository import InstalledRepository
+
+        env_dir = Path(
+            os.getenv("POETRY_HOME") if os.getenv("POETRY_HOME") else env.path
```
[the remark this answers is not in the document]
dimbleby: `os.getenv()` accepts an optional second parameter as a default, so all that you need is `os.getenv("POETRY_HOME", env.path)`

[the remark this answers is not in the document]
neersighted: Ah, good point. Still I think `POETRY_HOME` should be encapsulated and not scattered around as a string constant.

neersighted:
Another question -- did the fixes from #4686 get ported to the exporter plugin? @dimbleby, if not, porting your tests/fixes as well as any insight into the marker-negation regression would be appreciated!

dimbleby:
#4686 has a testcase, so long as this MR doesn't break that then I guess it's fine.

As for the "marker-negation regression": I believe it's #3660 that causes `bar==7.8.9` to be split into `bar==7.8.9; platform_system != "Windows" or platform_system == "Windows"` in that testcase.

So far as I know #3660 is a good fix, and it just exposes this oddity.  

I imagine this could be fixed, if desired, by recognising the special case of taking a union between a `SingleMarker` and its own inverse, [here](https://github.com/python-poetry/poetry-core/blob/af08f1ce720da467c9bf3d43eed3d9ebaf4ad7fb/src/poetry/core/version/markers.py#L282).

neersighted:
> #4686 has a testcase, so long as this MR doesn't break that then I guess it's fine.
> 
> As for the "marker-negation regression": I believe it's #3660 that causes `bar==7.8.9` to be split into `bar==7.8.9; platform_system != "Windows" or platform_system == "Windows"` in that testcase.
> 
> So far as I know #3660 is a good fix, and it just exposes this oddity.
> 
> I imagine this could be fixed, if desired, by recognising the special case of taking a union between a `SingleMarker` and its own inverse, [here](https://github.com/python-poetry/poetry-core/blob/af08f1ce720da467c9bf3d43eed3d9ebaf4ad7fb/src/poetry/core/version/markers.py#L282).

`test_exporter.py` is completely eliminated in this PR, in favor of a copy that was factored out into the plugin version. I'm still not sure if your fix made the leap (haven't looked yet).

Thanks for the PR -- interesting to know the breakage was in core (and it explains why I had trouble finding it) --- I'll take a look in a bit.

tests/console/commands/test_add.py
branchvincent: note that you can also use [`Path.parents`](https://docs.python.org/3/library/pathlib.html#pathlib.PurePath.parents):

```
    p.return_value = Path(__file__).parents[2] / "fixtures" / "git"
```

src/poetry/console/commands/plugin/add.py
```diff
@@ -74,57 +71,53 @@ def handle(self) -> int:
...
 
-        root_package = None
-        for package in installed_repository.packages:
-            if package.name == "poetry":
-                root_package = ProjectPackage(package.name, package.version)
-                for dependency in package.requires:
-                    root_package.add_dependency(dependency)
+            self.line(
+                "\nIf you want to upgrade it to the latest compatible version, "
+                "you can use `poetry plugin add plugin@latest.\n"
```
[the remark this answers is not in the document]
sdispater: I want to distinguish the `latest` special notation from true versions, hence the `@` notation to treat it differently.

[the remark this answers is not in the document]
neersighted: Hmm, fair enough. It's mildly surprising, but not totally out of left field since node uses this syntax. I think we should call it out as poetry-specific loudly in the docs then.

PR #4838, opened 2021-11-27
Fix dropped markers in dependency walk

finswimmer:
# Pull Request Check List

Resolves: #issue-number-here

### DIFF
--- a/poetry/packages/locker.py
+++ b/poetry/packages/locker.py
@@ -245,16 +245,15 @@ class Locker(object):
                         locked_package.to_dependency().constraint
                     )
 
-                if key not in nested_dependencies:
-                    for require in locked_package.requires:
-                        if require.marker.is_empty():
-                            require.marker = requirement.marker
-                        else:
-                            require.marker = require.marker.intersect(
-                                requirement.marker
-                            )
+                for require in locked_package.requires:
+                    if require.marker.is_empty():
+                        require.marker = requirement.marker
+                    else:
+                        require.marker = require.marker.intersect(requirement.marker)
 
-                        require.marker = require.marker.intersect(locked_package.marker)
+                    require.marker = require.marker.intersect(locked_package.marker)
+
+                    if key not in nested_dependencies:
                         next_level_dependencies.append(require)
 
             if requirement.name in project_level_dependencies and level == 0:
--- a/tests/utils/test_exporter.py
+++ b/tests/utils/test_exporter.py
@@ -752,6 +752,79 @@ foo==1.2.3
     assert expected == content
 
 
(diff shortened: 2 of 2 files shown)


dimbleby:
Looks like you are considering back-porting this to 1.1 - which would be much appreciated, thank you.

re the failing test, it seems as though the difference is that the 1.1 branch is able to collapse
```
bar==7.8.9; platform_system != "Windows" or platform_system == "Windows"
```
into just
```
bar==7.8.9
```
which is nice (I wonder when this was lost).  Hopefully not a reason to avoid taking the fix!

finswimmer:
> Looks like you are considering back-porting this to 1.1 - which would be much appreciated, thank you.

Yeah, at least I've tried. Because I've absolutely no clue about this topic I've passed this one over to @neersighted :smile:

dimbleby:
I think you can simply update the test to match what happens:
```diff
diff --git a/tests/utils/test_exporter.py b/tests/utils/test_exporter.py
index 816673a6..6cae9928 100644
--- a/tests/utils/test_exporter.py
+++ b/tests/utils/test_exporter.py
@@ -817,7 +817,7 @@ def test_exporter_can_export_requirements_txt_with_nested_packages_and_multiple_
         content = f.read()

     expected = """\
-bar==7.8.9; platform_system != "Windows" or platform_system == "Windows"
+bar==7.8.9
 baz==10.11.13; platform_system == "Windows"
 foo==1.2.3
 """
```
If there is anything to worry about here it would be: why doesn't master do this?  But it doesn't seem like anything much to get excited about.

neersighted:
> I think you can simply update the test to match what happens:
> 
> ```diff
> diff --git a/tests/utils/test_exporter.py b/tests/utils/test_exporter.py
> index 816673a6..6cae9928 100644
> --- a/tests/utils/test_exporter.py
> +++ b/tests/utils/test_exporter.py
> @@ -817,7 +817,7 @@ def test_exporter_can_export_requirements_txt_with_nested_packages_and_multiple_
>          content = f.read()
> 
>      expected = """\
> -bar==7.8.9; platform_system != "Windows" or platform_system == "Windows"
> +bar==7.8.9
>  baz==10.11.13; platform_system == "Windows"
>  foo==1.2.3
>  """
> ```
> 
> If there is anything to worry about here it would be: why doesn't master do this? But it doesn't seem like anything much to get excited about.

I've spent some time noodling with this, and concluded that I can't find it and will probably need to bisect. @dimbleby, since you've done other work in this area, a PR fixing this regression on master (with a separate test, of course) would be greatly appreciated. In the mean time I'll just change the test on 1.1 so we can cut the release since the regression is in master.

neersighted (approved):

PR #4839, opened 2021-11-28
Update documentation commands to use new installer

mkniewallner:
# Pull Request Check List

Since the new installer seems to be the supported way of installing/updating/removing Poetry, and the documentation was updated for the installation process, I believe that other commands for installing/updating/removing Poetry should also use the new installer.

### DIFF
--- a/README.md
+++ b/README.md
@@ -88,7 +88,7 @@ You can also install Poetry for a `git` repository by using the `--git` option:
 curl -sSL https://install.python-poetry.org | python - --git https://github.com/python-poetry/poetry.git@master
 ````
 
-**Note**: Note that the installer does not support Python < 3.6.
+_Note that the installer does not support Python < 3.6._
 
 ## Updating `poetry`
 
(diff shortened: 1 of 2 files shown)


neersighted (changes requested):
Thanks for this change! One more detail to take care of -- would you mind changing the `python` invocations to `python3`? Debian and friends don't even have a `python` command without a metapackage that is non-default. Thanks!

mkniewallner:
> Thanks for this change! One more detail to take care of -- would you mind changing the `python` invocations to `python3`? Debian and friends don't even have a `python` command without a metapackage that is non-default. Thanks!

Done in https://github.com/python-poetry/poetry/pull/4839/commits/d32eee02974ffa58fc8344f2598727d4133000d2
I kept `python` for Windows though because I don't know if it works the same way.
Happy to also update it if it makes sense though.

neersighted (approved):

Issue #4844, opened 2021-11-29
Documentation fix - poetry add with square brackets requires quotes


ttamg:
A small documentation fix required to make it more intuitive how to install packages with options. An example would be `uvicorn` with dependencies such as `uvicorn[standard]'

When using poetry to add this in the default way we get an error 

```
poetry add uvicorn[standard]
```

The error such as ... `zsh: no matches found: uvicorn[standard]`.  This is on a Mac and using zsh.

Calling the `poetry add` but using quotes works just fine.  

```
poetry add 'uvicorn[standard]'
```

It took me 10 mins to find this little hack as this doesn't seem to be a `pip` requirement.  So that others don't have the same hassle, could we make this clear in the documentation please?

The poetry documentation here https://python-poetry.org/docs/cli/#add has the following which I think is misleading

```python
poetry add requests[security,socks]
poetry add "requests[security,socks]~=2.22.0"
poetry add "git+https://github.com/pallets/flask.git@1.1.1[dotenv,dev]"
```

My suggestion is
1.  Remove the first line as it doesn't work (on Mac anyway)
2. Add in a note saying that when there are options in `[]` then make sure you use quotes in the name of the package.

I'd be happy to do the PR on this but first let me know if you think this is a good idea.  I see there are 148 PRs open so not sure if they are getting airtime.

Thanks
Matt

neersighted:
This is not related to Poetry, but your shell. ZSH is treating the braces as a special expansion syntax.

That being said, a PR adding double quotes would be welcome -- it's safe in all shells, even if the user is using a shell that does implicit square brace expansion.

ttamg:
@neersighted OK here's a simple docs PR that adds in comment for these shells that have issues.
#4847

PR #4847, opened 2021-11-29
Documentation improvement for issue #4844

ttamg:
Resolves: #4844

Small docs fix that adds a warning box to suggest using quotes on some shells such as Mac Zsh.  

- [ ] Added **tests** for changed code.  NO CHANGED CODE - ONLY DOCS IN PR

### DIFF
--- a/docs/cli.md
+++ b/docs/cli.md
@@ -341,6 +341,10 @@ poetry add "requests[security,socks]~=2.22.0"
 poetry add "git+https://github.com/pallets/flask.git@1.1.1[dotenv,dev]"
 ```
 
+{{% warning %}}
+Be aware that some shells (in particular Mac Zsh) may not recognise the package name without quotes - for example `poetry add requests[security,socks]`. In this case try instead with quotes `poetry add "requests[security,socks]"`
+{{% /warning %}}
+
 If you want to add a package to a specific group of dependencies, you can use the `--group (-G)` option:
 
 ```bash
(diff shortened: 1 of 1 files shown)


docs/cli.md
```diff
@@ -341,6 +341,10 @@ poetry add "requests[security,socks]~=2.22.0"
 poetry add "git+https://github.com/pallets/flask.git@1.1.1[dotenv,dev]"
 ```
 
+{{% warning %}}
+Be aware that some shells (in particular Mac Zsh) may not recognise the package name without quotes - for example `poetry add requests[security,socks]`. In this case try instead with quotes `poetry add "requests[security,socks]"`
+{{% /warning %}}
+
```
neersighted, line 347: I don't think this warning is very good -- ZSH will behave the same on all platforms, so qualifying it with macOS is useless at best and FUD at worst. I would suggest instead changing line 339 to `"requests[security,socks]"` and changing this warning to be more generalized:

```suggestion
{{% warning %}}
Some shells may treat square braces (`[` and `]`) as special characters. It is suggested to always quote arguments containing these characters to prevent unexpected shell expansion.
{{% /warning %}}

```

neersighted (changes requested):

docs/cli.md
ttamg: Good idea.  Much clearer.  Done.  Thanks for the suggestion

neersighted (approved):

neersighted:
This PR was based off of a stale master and doesn't have the correct workflows. Can you please rebase on top of current master to get this merged?

ttamg:
@neersighted I believe I have done that now.  Next time would do on a separate branch as you suggest.  Hope this is now rebased correctly.  If not then please let me know what I missed!  Thanks

Issue #4595, opened 2021-10-05
Governance - or, "what do we do with all these pull requests?"


danieleades:
It's a testament to the quality of this project that it attracts such a high volume of issues and pull requests. Poetry is a pleasure to use and solves a fairly universal problem in Python development. For that reason, it's fast becoming a ubiquitous choice for package management.

By virtue of being so popular, and also because it's so exposed to the host system (by it's nature), Poetry users find no end of possible improvements, bugs, or unsupported edge cases. At the time of writing there are 1.1k open issues and 214 open pull requests. Because (presumably) of the high volume, many of the issues are ignored, and many of the pull requests stay unreviewed until they stagnate. at the time of writing there are 364 open issues or pull requests that have never received a single comment (`is:open comments:<1`)

Admittedly there's probably a lot of garbage in there, but it's not the most welcoming environment for would-be first time contributors.

I don't have a solution here, but i think it's pretty clear some structural/governance changes are required to sort through and make sense of all of this in a constructive way.

danieleades:
I don't know how to solve this, but I suspect "more maintainers" is probably the actual best way forward.

I've not spent any time on the Discord server, but it looks pretty active, and i'm guessing that's pretty useful. Has that been the general experience?

some possible ideas-

## More Maintainers

throwing warm bodies at the problem is a tried and true method, but managing to do that while maintaining code quality and project vision will require changes to the eponymous Governance. Involving more people in a project involves a democratisation of trust, vision, and quality. Doing *that* *could* require more stringent code quality standards (see below), and will almost certainly involve a much clearer statement of...

## Project Vision

Goals, Non-Goals. Detailed roadmaps (#1856 is a pretty good start). Types of pull requests that will be considered, and types of pull requests that will be summarily closed. PRs in the first category should be reviewed quickly. PRs in the second category should be closed quickly. Probably stale PRs should also be closed automagically. You'll notice there's no "eternal purgatory" category, that's intentional.

## Code Quality

Code quality in this project is pretty good. Continuing to ratchet up those standards provides greater confidence when merging and reviewing pull requests.

a couple of potential improvements here:

- track and report code coverage as an informational tool when reviewing pull requests. [codecov](https://about.codecov.io/) is nice for this.
- incrementally introduce *typing* to the codebase

just spitballing here. Not trying to start a flame war on python's optional typing.

## monetisation/incentives

[...]

PR #4896, opened 2021-12-15
add a workflow to close stale issues

danieleades:
adds a workflow for automatically closing out stale issues and PRs

this workflow *may* need additional permissions to run, but i'm not sure how to test this.

see https://github.com/marketplace/actions/close-stale-issues#recommended-permissions

### DIFF
--- a/.github/workflows/stale.yml
+++ b/.github/workflows/stale.yml
@@ -1,0 +1,16 @@
+name: "Stale issue handler"
+on:
+  workflow_dispatch:
+    schedule:
+      - cron: "0 0 * * *"
+
+jobs:
+  stale:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/stale@v4.1.0
+        with:
+          days-before-stale: 30
+          days-before-close: 10
+          stale-issue-message: "This issue is stale because it has been open 30 days with no activity. Remove `stale` label or comment or it will be closed in 10 days"
+          stale-pr-message: "This pull request is stale because it has been open 30 days with no activity. Remove `stale` label or comment or it will be closed in 10 days"


dimbleby:
I understand that this project has a large pile of unaddressed issues so: but I do not like stalebot as a solution.  (Actually I'm not even sure that this is a problem that needs a solution).

As someone who occasionally raises issues against projects with such bots: it's just horrible.  Problems very rarely solve themselves just because time has passed; and being badgered to confirm this obvious truth is infuriating.

Perhaps it would be an interesting exercise to look at the most recent dozen or so reports that this workflow would try to close out - and judge whether they are actually resolved already.  I bet most of them are not.

If there's desire among maintainers to keep the issue list small, I suggest that this will only usefully be solved by more maintainer time (whether that be through more commitment from those who are already active, or by bringing more people on board).  

Alternatively there's a perfectly reasonable view that says that a large issue count is no big deal: it's all just data, and issues provide a space for users to share workarounds and plan fixes etc.

danieleades:
@dimbleby that's a fair perspective, and I see where you're coming from. I don't fully agree though- to me those sound like arguments for a *relaxed* stalebot config, not for not having one at all.

I'm not sure I'd characterise the volume of open PRs and issues as 'data', but rather as 'noise'. Especially when contributors can just as easily search closed issues as open ones.

To be clear, the problem I'm aiming to solve is not the volume of open tickets, but rather the volume of issues/PRs that are stale, unmaintained, unmergable, irrelevant, etc. I think stale PRs that are simply long-running or blocked are very very much in the minority. There's a huge number that have never been interacted with since they were opened *at all*, and the codebase has since moved on.

See #4595

Would you be happier if the expiry times for stale objects were longer? If so, what numbers do you think are reasonable?

dimbleby:
No, I don't think that tweaking the parameters helps from my point of view.  My clear preference would be for a human (or humans) to do a sweep and close down things that they judged to be stale.

I could just about see a case for a one-time lets-close-everything-older-than-X, if that signaled maintainer intention to keep the issue list down in future.  

But I feel very strongly that having a bot that closes down issues simply because there has been no recent update is just a horrible experience for contributors.

danieleades:
> My clear preference would be for a human (or humans) to do a sweep and close down things that they judged to be stale

we don't disagree on this. If a human maintainer were able to perform this sweep (and continue to do so going forward), then that would absolutely be preferred.

> But I feel very strongly that having a bot that closes down issues simply because there has been no recent update is just a horrible experience for contributors.

For the sake of argument, let's say we configured the bot to close PRs that have not been updated in any way in the last year-

would a bot giving a contributor a warning that their PR will be closed after 12 months of inactivity unless they comment on it be significantly worse than the experience of being completely ignored for 12 months?

dimbleby:
On the whole I think I've said my piece here and should leave it to project maintainers to decide whether this is something they want.

> would a bot giving a contributor a warning that their PR will be closed after 12 months of inactivity unless they comment on it be significantly worse than the experience of being completely ignored for 12 months?

It depends.  If commenting brings some sort of commitment that the issue will not continue to be ignored, then it may be worthwhile.  If - which I'm afraid is usually the case with a stalebot - contributors are expected to comment _and_ to be ignored by humans: then yes, it's meaningfully worse.

Kaltsoon:
Poetry needs more active maintainers, not a bot that closes issues.

danieleades:
> Poetry needs more active maintainers, not a bot that closes issues.

i don't think there's a github action for that

once again, i don't disagree that's a better solution.

finswimmer:
Thanks a lot for preparing this PR :+1:

I've already asked @sdispater if he is willing to run a stale bot for a while, but he haven't had the time for a decision.

We don't want to run a stale bot on a regular bases, because we know that some people don't like it, but it would help to run it for some weeks to clean up the issue tracker. Doing this we and new contributors can faster recognize which topics are still valid.

muhammedseydali (approved):

muhammedseydali:
.

neersighted:
Closing this for now as we haven't reached a consensus and we've been slowly trimming away the backlog. Would love to have more volunteers for triage -- please reach out if you're interested.

Issue #4678, opened 2021-10-27
git repo with hypthens fails with error


michael-c-hoffman:
- [ x] I am on the [latest](https://github.com/python-poetry/poetry/releases/latest) Poetry version.
- [x ] I have searched the [issues](https://github.com/python-poetry/poetry/issues) of this repo and believe that this is not a duplicate.
- [ x] If an exception occurs when executing a command, I executed it again in debug mode (`-vvv` option).

- **OS version and name**: Fedora 34
- **Poetry version**: 1.1.5
- **Link of a [Gist](https://gist.github.com/michael-c-hoffman/34ded6f312419fda97254437c052380b)

## Issue
When adding a git repo that includes a hyphen in the org/path poetry fails to add with an error:
```
TypeError

  expected string or bytes-like object

  at ~/.local/lib/python3.9/site-packages/poetry/core/utils/helpers.py:27 in canonicalize_name
       23│ _canonicalize_regex = re.compile(r"[-_]+")
       24│ 
       25│ 
       26│ def canonicalize_name(name):  # type: (str) -> str
    →  27│     return _canonicalize_regex.sub("-", name).lower()
       28│ 
       29│ 
       30│ def module_name(name):  # type: (str) -> str
       31│     return canonicalize_name(name).replace(".", "_").replace("-", "_")
```

gabrielbdsantos:
I'm having the same issue on poetry 1.1.11.

`poetry add --help` gives `poetry add git+https://github.com/python-poetry/poetry.git` as an example, which is indeed a valid one. I have just tried it. However, when I try `poetry add git+https://github.com/tpaviot/pythonocc-core.git`, poetry throws that issue.

So, it seems that the problem is not with hyphens in general, but only in the package names, right?

michael-c-hoffman:
Correct

michael-c-hoffman:
yes I believe that is what is needed

michael-c-hoffman:
can this get merged into the next release?

neersighted:
> 
> 
> can this get merged into the next release?

Hi @michael-c-hoffman -- I'm not aware of any outstanding PRs with this fix. A PR with the fix and a regression test would be welcomed, and I'd love to be pointed to an already-extant one if I missed it.

nathan-kairon:
I got the same issue

abn:
I suspect the reported issue has been resolved with #5428 as the vay we create the directory was changed. This might also be a duplicate of #3392.

[...]

PR #4922, opened 2021-12-21
fix: git subdirectory dependencies fail to resolve with a TypeError on name

steved:
# Pull Request Check List

Resolves: #4678

This is an attempt to create a fix and test for #4678 where the addition of subdirectory support for Git repositories (https://github.com/python-poetry/poetry-core/pull/192) does seem to require some added code for package name discovery. The error manifests as the TypeError in the associated issue when creating a Git dependency with a subdirectory (hyphen actually ends up not mattering).

I'm having trouble creating a clean unit test, however, as the bug is actually partially in the [`_pep517_metadata`](https://github.com/python-poetry/poetry/blob/6a35cc41e950d4ff0f1efad649c1d15d689a9ce2/src/poetry/inspection/info.py#L442) method that is entirely [mocked out](https://github.com/python-poetry/poetry/blob/6a35cc41e950d4ff0f1efad649c1d15d689a9ce2/tests/conftest.py#L203). There may be a way to craft an `UNKNOWN.egg-info` directory instead, but some direction for a better test would be helpful.

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -94,7 +94,7 @@ python-versions = ">=3.6.1"
 
 [[package]]
 name = "charset-normalizer"
-version = "2.0.7"
+version = "2.0.9"
 description = "The Real First Universal Charset Detector. Open, modern and actively maintained alternative to Chardet."
 category = "main"
 optional = false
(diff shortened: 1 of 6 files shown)


steved:
cc @sdispater

Issue #4919, opened 2021-12-21
`poetry-core` 'downstream' tests are failing


danieleades:
the downstream tests CI job is failing on master.

see https://github.com/python-poetry/poetry-core/pull/259

dimbleby:
introduced at https://github.com/python-poetry/poetry-core/commit/af08f1ce720da467c9bf3d43eed3d9ebaf4ad7fb I think, per https://github.com/python-poetry/poetry-core/pull/255#issuecomment-981086491

radoering:
It seems that the bug fixed by python-poetry/poetry-core@af08f1c has been exploited downstream. 😉 I tried to fix that in the attached PR.

PR #4932, opened 2021-12-23
Fix misuse of pretty_constraint, fix downstream test (poetry-core)

radoering:
# Pull Request Check List

Resolves: #4919

The pretty_constraint fix in python-poetry/poetry-core@af08f1c broke a downstream test (test_exporter_can_export_requirements_txt_with_nested_packages_and_markers_any), exposing a misuse of the fact that pretty_constraint was not updated when setting the constraint. This PR fixes the downstream test by fixing the misuse.

### DIFF
--- a/src/poetry/utils/exporter.py
+++ b/src/poetry/utils/exporter.py
@@ -1,3 +1,4 @@
+import itertools
 import urllib.parse
 
 from typing import TYPE_CHECKING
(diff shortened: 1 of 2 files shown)


src/poetry/utils/exporter.py
```diff
@@ -69,13 +70,21 @@ def _export_requirements_txt(
         content = ""
         dependency_lines = set()
 
-        for dependency_package in self._poetry.locker.get_project_dependency_packages(
-            project_requires=self._poetry.package.all_requires, dev=dev, extras=extras
+        for package, groups in itertools.groupby(
```
neersighted, line 73: Can you walk me through this set of changes and why they are needed? I can follow the line-by-line, but I'm not sure I'm groking the bigger picture.

Also, this file is going to be removed in favor of the exporter plugin. Can you please make a PR against that repo as well?

radoering: If you are wondering: I accidentally deleted my branch while porting the fix to the exporter plugin. Sorry about that. 😅

The changes in `exporter.py` are required because `get_project_dependency_packages` returns objects of type `DependencyPackage`, which consist of a `Dependency` and a `Package`. There can be multiple dependency packages containing the same package for different dependencies, e.g. package `a(1.1)` for dependency `a(>=1.0)` with markers `sys_platform == "win32"` and package `a(1.1)` for dependency `a(>=1.1)` with markers `sys_platform == "linux"`. The changes are required in order to receive only one entry for `a==1.1` in requirements.txt instead of two entries with different markers.

This behaviour can be observed in the following test cases:

`test_exporter_can_export_requirements_txt_with_nested_packages_and_multiple_markers`

expected:
`bar==7.8.9 ; platform_system != "Windows" or platform_system == "Windows"`

without changes:
```
bar==7.8.9 ; platform_system != "Windows"
bar==7.8.9 ; platform_system == "Windows"
```

`test_exporter_can_export_requirements_txt_with_nested_packages_and_markers_any`

expected:
`a==1.2.3`

without changes:
```
a==1.2.3
a==1.2.3 ; python_version < "3.8"
```

`test_exporter_can_export_requirements_txt_pyinstaller`

expected:
```
altgraph==0.17
```

without changes:
```
altgraph==0.17
altgraph==0.17 ; sys_platform == "darwin"
```

neersighted (approved):

Issue #4874, opened 2021-12-09
Only access keyring credentials on demand


nbcomplete:
- [x] I have searched the [issues](https://github.com/python-poetry/poetry/issues) of this repo and believe that this is not a duplicate.
- [x] I have searched the [documentation](https://python-poetry.org/docs/) and believe that my question is not covered.

## Feature Request

It seems that right now poetry eagerly loads credentials from the system keyring. When no credentials are configured everything works fine until a command such as `poetry install` hits a private repository, but when credentials are configured, poetry loads them for every command invocation, even if the command does not need access to any repositories (such as `poetry run`).

This doesn't seem like a big deal, but on macOS this means that you have to type your password (the dialog doesn't even support TouchID unfortunately) every single time you run a poetry command with private repository credentials configured. The alternative would be to click the "Always allow" button in the password prompt, but since the python binary is generally not cryptographically signed, this would open an attack vector for a tampered python binary to read the credentials without user interaction.

Granted, the attack vector is small and unlikely to be exploited, but I don't feel comfortable doing this on a company machine where saving passwords in plaintext on disk is generally prohibited, considering this comes very close to a plaintext password on disk.

Hence it would be cool if poetry only accessed the keyring on demand when it actually needs the password for accessing a repository.

axxelG:
This has an impact on users of the Python extension for Visual Studio Code, too. The extension uses `poetry env list --full-path` to discover virtualenvs but the command runs in the background so the password prompt is blocking the command which makes poetry and VS Code unusable.
[https://github.com/microsoft/vscode-python/issues/18204](https://github.com/microsoft/vscode-python/issues/18204)

abn:
I suspect the spirit of this change has been implemented as a side-effect of #5518. Please comment if I am missing something.

PR #4937, opened 2021-12-25
Lazy load credentials for legacy repositories

axxelG:
Move loading credentials from LegacyRepository.__init__() separate function and call it from _get_page() and authenticated_url()
This avoids loading credentials for commands that don't use them (e. g. poetry env list). This helps to avoid annoying users that have to unlock their keyring and if such commands are called as background tasks (CI pipelines, VS-Code environment detection).

Questions and remarks from my side:
- Because of the current implementation I assumed that legacy repositories always have basic auth. If authentication is optional, I would need help to figure out possible cases and to adapt (parameterize) the tests. My pytest knowledge is very limited.
- Are you appreciating the more verbose docstring/comment on `authenticated_url()`? I'm fine with removing it.
- I was not sure if `_set_authentication_creds()` should be a private or public function. Tell me if you prefer it to become `set_authentication_creds()`

Resolves: #4874

- [ x ] ~~Added~~ changed **tests** for changed code.

### DIFF
--- a/src/poetry/repositories/legacy_repository.py
+++ b/src/poetry/repositories/legacy_repository.py
@@ -188,12 +188,6 @@ class LegacyRepository(PyPiRepository):
             self._authenticator.session, cache=FileCache(str(self._cache_dir / "_http"))
         )
 
-        username, password = self._authenticator.get_credentials_for_url(self._url)
-        if username is not None and password is not None:
-            self._authenticator.session.auth = requests.auth.HTTPBasicAuth(
-                username, password
-            )
-
         if self._cert:
             self._authenticator.session.verify = str(self._cert)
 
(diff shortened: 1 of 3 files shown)


axxelG:
pre-commit.ci autofix

src/poetry/repositories/legacy_repository.py
```diff
@@ -212,9 +206,19 @@ def client_cert(self) -> Path | None:
 
     @property
     def authenticated_url(self) -> str:
-        if not self._session.auth:
-            return self.url
+        """
+        Set authentication credentials and return a URL with basic auth
```
abn, line 210: This code path is only used for uploads not for requests etc. see Authenticator for details.

abn:
I suspect the spirit of this change has been implemented as a side-effect of #5518. Please comment if I am missing something.

Issue #4952, opened 2021-12-29
early prune based on markers


maciejskorski:
- [x] I have searched the [issues](https://github.com/python-poetry/poetry/issues) of this repo and believe that this is not a duplicate.
- [x] I have searched the [documentation](https://python-poetry.org/docs/) and believe that my question is not covered.

## Feature Request

As of now poetry spends a lot of time on overrides which doesn't match the environmental version or platform requirements. 
They are filtered in the very end of the process. It would be great to have early pruning of such overrides:

Consider this example run within a virtual environment with Python 3.8
```
[tool.poetry.dependencies]
python = "~3.8"
pandas = [
    {version="1.3.*",markers="python_version=='3.9.*'"},
    {version="1.2.*",markers="python_version=='3.7.*'"},
    {version="1.1.*",markers="python_version=='3.10.*'"},
]
....
```
Petry correctly concludes the empty set of matching packages, however it splits the graph in at least 3 branches first - and this escalates the more markers we have. It would help to compare the markers early and ignore some of the branches.

maciejskorski:
Here are my further thoughts once I got familiar with the code:

Install configs, written to the .lock file are platform agnostic, e.g. cover all possible scenarios. This exponentially impacts the running time in some scenarios (combine pandas, black, and azureml in its modern releases). 

Adding pre-filtering to match the target markers seems easy, but it's going to change the way log file looks like so more work on the tests side.

radoering:
The example is only a special case of the issue, isn't it? Actually, in the example all of the pandas constraints should be ignored because they don't match the project's python constraint. (I would almost consider it a bug that this currently is not the case.) If the python constraint would be `"^3.7"` instead of `"~3.8"`, it would be correct to consider all pandas overrides when locking (no matter in which environment).

I think, in the example you can even force the desired behaviour by `python` instead of `markers`:

```
[tool.poetry.dependencies]
python = "~3.8"
pandas = [
    {version="1.3.*",python="3.9.*"},
    {version="1.2.*",python="3.7.*"},
    {version="1.1.*",python="3.10.*"},
]
```

When using this syntax, the pandas constraints are ignored...

In my opinion, creating one lock file for all platforms and python versions on an arbitrary platform with one of the supported python versions (together with multiple constraint dependencies in the pyproject.toml) is one of the most important features of poetry. Further, once the lock file has been created, installation does not require locking again.

maciejskorski:
> The example is only a special case of the issue, isn't it? 
[...]

PR #4958, opened 2021-12-31
Ignore dependencies not matching the project's python constraint...

radoering:
...no matter if the python version is specified by "python" keyword or by "markers"

# Pull Request Check List

Relates-to: #4952

Currently, the following equivalent dependency specifications result in different lock files:

```
[tool.poetry.dependencies]
python = "~3.8"
pandas = {version="1.3.*", python="~3.9"}
```

```
[tool.poetry.dependencies]
python = "~3.8"
pandas = {version="1.3.*", markers="python_version=='3.9.*'"}
```

When using `python` the pandas dependency not matching the project's python constraint is ignored. When using `markers` the dependency is not ignored.

This PR harmonizes the behaviour between both variants.

### DIFF
--- a/src/poetry/mixology/version_solver.py
+++ b/src/poetry/mixology/version_solver.py
@@ -70,6 +70,7 @@ class VersionSolver:
         """
         start = time.time()
         root_dependency = Dependency(self._root.name, self._root.version)
+        root_dependency.python_versions = self._root.python_versions
         root_dependency.is_root = True
 
         self._add_incompatibility(
--- a/tests/puzzle/test_solver.py
+++ b/tests/puzzle/test_solver.py
@@ -420,6 +420,30 @@ def test_solver_fails_if_mismatch_root_python_versions(
         solver.solve()
 
 
+def test_solver_ignores_python_restricted_if_mismatch_root_package_python_versions(
+    solver: Solver, repo: Repository, package: ProjectPackage
+):
+    solver.provider.set_package_python_versions("~3.8")
+    package.add_dependency(
+        Factory.create_dependency("A", {"version": "1.0", "python": "<3.8"})
+    )
+    package.add_dependency(
+        Factory.create_dependency(
+            "B", {"version": "1.0", "markers": "python_version < '3.8'"}
+        )
+    )
+
+    package_a = get_package("A", "1.0")
+    package_b = get_package("B", "1.0")
+
+    repo.add_package(package_a)
+    repo.add_package(package_b)
+
+    transaction = solver.solve()
+
+    check_solver_result(transaction, [])
+
+
 def test_solver_solves_optional_and_compatible_packages(
     solver: Solver, repo: Repository, package: ProjectPackage
 ):


maciejskorski:
I believe this is related to #4959

radoering:
@maciejskorski 

1. This PR only fixes inconsistent behavior and doesn't change the fact that the lock file is independent from the current environment (platform and Python version). The Python version of the used environment does not matter here. Only the Python constraint in the pyproject.toml is relevant. It may be similar but it is not the same. A filter on `python_version` in your PR filters dependencies by the current Python version (which is exactly one version). Here the dependencies are filtered by all Python versions matching the Python constraint of the pyproject.toml.
2. Probably, you are right. But that has to be solved in another PR. ;)

maciejskorski:
@radoering 

I see you proposed to inherit the missing constraint from the parent. 

However, as documented in #4965, the real issue is that the keyword-type and marker-type python constraints exclude each other, as per the current core logic. While your approach works as a temporary solution, fixing the markers for good would be a more stable and beneficial solution (especially that the docs say, markers and python versions should work in tandem).

radoering:
> 2\. I see you proposed to inherit the missing constraint from the parent, is that right?

The constraint is propagated from the root package to the root dependency. This makes the filter logic you reference take effect.

> 2\. the real issue is that the keyword-type and marker-type python constraints exclude each other

The issue addressed by this PR has nothing to do with keyword-type and marker-type python constraints excluding each other. If you look at the examples at the top, you will see there is either a keyword-type or a marker-type constraint (not both together).



> 2\. By fixing the parser (to [collect versions as expected and promised in docs](https://python-poetry.org/docs/dependency-specification/#expanded-dependency-specification-syntax)) we would fix the problem at the root for good, which to me sounds better than this post-hoc patch?

Fixing the root cause sounds good, but when I revert the fix in version_solver.py from this PR and apply the changes to factory.py from your PR, the test added in this PR fails. Thus, I suppose this PR is independent of your PR.

maciejskorski:
@radoering 

I run your example, but would encourage you to consider the code I provided in reciprocity:-)

So, if you set the debugger at the guilty lines I indicated above (poetry/src/poetry/puzzle/provider.py)
```
[d._python_constraint for d in requires]
```
you will see that 

- the filtering you propose is already in place 
- the filtering in your use-case fails as python version is not populated at all -wrongly. (reason: #4965)

fix the root-cause rather than patching post-hoc, and we will be all fine. 

As of now, yes, both our PRs are drafts and not fully tested.

radoering:
@maciejskorski: I'm afraid I don't quite follow you.

I see that there are several places where filtering takes place. At first, the lines you mention (= early filtering). And later the following lines (= late filtering), which also were in place before this PR:

https://github.com/python-poetry/poetry/blob/1aa4d5b8b0114baa9e878e5692addbebb62c3e72/src/poetry/puzzle/provider.py#L692-L713

The fix in this PR triggers one of the `continue` statements in these lines by setting `python_versions` of the root dependency.

> I run your example, and that's why I encourage you to consider the code I provided in reciprocity:-)

I understand that filtering earlier is better. However, I am not sure which code you're referring to. If I don't mess something up, your first commit https://github.com/python-poetry/poetry/pull/4970/commits/6d5409be3b2abbb410ac2ad33e8d516ef1059fc8 does neither trigger early nor late filtering and fails the test added in this PR. Your later commit https://github.com/python-poetry/poetry/pull/4970/commits/ec42fa5a0ae0235d1c447bde70b978bd928cd93a triggers early filtering and passes the test. However, other tests in `test_solver.py` fail when I apply https://github.com/python-poetry/poetry/pull/4970/commits/ec42fa5a0ae0235d1c447bde70b978bd928cd93a. In contrast, the fix in this PR does not break any existing test.

Nevertheless, even if there is a solution that triggers early filtering for this use case and passes all tests, I don't see a reason why `python_versions` of the root dependency should not be set.

maciejskorski:
@radoering 

Sorry, I think we are both active on few issues and are about to mix topics so let me try to clarify better.

First, nothing logically wrong with picking the lost constraint from the parent/root.
However it seems the matching mechanism you propose to add, already exists. For that referenced above `provider .py lines 494-496`.

I am asking if you know why is this code not filtering out wrong python versions? It should.

radoering:
@maciejskorski: OK, so I've wrongly assumed that you are referring to the other PR. Sorry about that.

If I set a breakpoint at the mentioned lines and execute the new test, I see that package A is discarded because a python_constraint is set and package B is not discarded because no python constraint is set. Thus, ensuring that a python_constraint is set in package B somehow would be an alternative solution.

> However it seems the matching mechanism you propose to add, already exists.

I don't propose to add any matching mechanism. The filtering already exists. It is just not triggered because `python_versions` of the dependency is not set. As you can see in https://github.com/python-poetry/poetry/pull/4958/files, I only added one line of code in `version_solver.py` and a test.

maciejskorski:
@radoering 

Yes! We are closer and closer to agree on this. Like I detailed in related issues, the reason for non-setting it lies deeply in `poetry-core`. The true reason is that python constraint was not deducted from the marker, and I want the fix there.

To cut this discussion short, I checked out that  **most recent poetry-core improves handling markers** and to me it solves your issue?

Please make sure you install `poetry-core@master` (not any precached version from pip). Then on my machine, the problem is resolved as python constraints and markers are aligned...





Note (I wasn't aware of this myself) that `poetry-core` has improved handling markers.

radoering:
You are right. It seems the issue has already been fixed upstream. Thanks for pointing this out.

I still wonder if there are some cases where early filtering is not possible so that setting the root dependency's python constraint is required for late filtering. Maybe, I'll dig a bit deeper if I find some time.

radoering:
Closing PR because issue has already been fixed upstream. Created separate PR # #5307 to add only the test case.

Issue #4963, opened 2022-01-01
Using "run" with no arguments gives unclear error message


noa-weiss:
- [x] I am on the [latest](https://github.com/python-poetry/poetry/releases/latest) Poetry version.
- [x] I have searched the [issues](https://github.com/python-poetry/poetry/issues) of this repo and believe that this is not a duplicate.
- [x] If an exception occurs when executing a command, I executed it again in debug mode (`-vvv` option).

- WSL-2 on Windows 11
- Poetry version 1.1.12

## Issue
Using "poetry run" without arguments gives the following error message:

```
FileNotFoundError

  [Errno 2] No such file or directory: b'/snap/bin/run'

  at /usr/lib/python3.8/os.py:601 in _execvpe
       597│         path_list = map(fsencode, path_list)
       598│     for dir in path_list:
       599│         fullname = path.join(dir, file)
       600│         try:
    →  601│             exec_func(fullname, *argrest)
       602│         except (FileNotFoundError, NotADirectoryError) as e:
       603│             last_exc = e
       604│         except OSError as e:
       605│             last_exc = e
```

It would be better, and much clearer, to have an error message that informs the user the command is failing due to missing arguments.

PR #4964, opened 2022-01-01
Add sane error when "run" is used without arguments

hibukki:
When someone writes:
`poetry run` (with nothing else)

Before the fix, they'd get this long thing:

```
  FileNotFoundError

  [Errno 2] No such file or directory: b'/Users/my_user/.rvm/bin/run'

  at /Library/Frameworks/Python.framework/Versions/3.7/lib/python3.7/os.py:604 in _execvpe
       600│         path_list = map(fsencode, path_list)
       601│     for dir in path_list:
       602│         fullname = path.join(dir, file)
       603│         try:
    →  604│             exec_func(fullname, *argrest)
       605│         except (FileNotFoundError, NotADirectoryError) as e:
       606│             last_exc = e
       607│         except OSError as e:
       608│             last_exc = e
```

After the fix, they get:
```
Missing arguments; try `poetry run python your_script.py`
```

Added a test, but didn't change the documentation.

(This is my first contribution, feel free to say/fix if I did something wrong)

Thanks to @noa-weiss for helping

Resolves: #4963

### DIFF
--- a/src/poetry/console/application.py
+++ b/src/poetry/console/application.py
@@ -201,6 +201,10 @@ class Application(BaseApplication):
 
             input = cast(ArgvInput, io.input)
             run_input = RunArgvInput([self._name or ""] + input._tokens)
+            
+            if 1 == len(input._tokens):
+                raise ValueError("Missing arguments; try `poetry run python your_script.py`")
+            
             # For the run command reset the definition
             # with only the set options (i.e. the options given before the command)
             for option_name, value in input.options.items():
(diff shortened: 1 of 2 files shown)


finswimmer:
Hello @hibukki,

thanks a lot for your contribution. We really appreciate this :+1: 

For me it looks like this is already fixed in the current `master`, maybe upstream by `cleo`:

```
$  poetry run       

Not enough arguments (missing: "args")
```

Can you confirm this?

fin swimmer

branchvincent:
Thanks for the PR! As noted, this should already be fixed and will be released with `1.2`. I'm going to close but do re-open if you notice anything different

Issue #4966, opened 2022-01-01
Missing return for link.is_wheel==True


Anselmoo:
- [x] I am on the [latest](https://github.com/python-poetry/poetry/releases/latest) Poetry version.
- [x] I have searched the [issues](https://github.com/python-poetry/poetry/issues) of this repo and believe that this is not a duplicate.

- **OS version and name**: macOS:latest
- **Poetry version**: **1.1.12**
- **Link of a [Gist](https://gist.github.com/) with the contents of your pyproject.toml file**: 

## Issue

According to PR #4962, there is the question left, if `pass` should not be replaced by a `return`. This would be also consistent with the typing definition `Optional[Link]`.

https://github.com/python-poetry/poetry/blob/1aa4d5b8b0114baa9e878e5692addbebb62c3e72/src/poetry/installation/chef.py#L44-L47

At the moment 
```python 
archives = self.get_cached_archives_for_link(link)
``` 
will be always excetuded.

Anselmoo:
@finswimmer issue is assigend

finswimmer:
Thanks @Anselmoo :+1: I've asked the other maintainers to have a look onto the PR, because I cannot estimate the impact of this change (Even if I guess it's correct).

PR #4967, opened 2022-01-01
Fixed Bug #4966: Replace pass by return

Anselmoo:
# Pull Request Check List

Resolves: #4966


neersighted (approved):
This is a pretty obvious oversight. Thanks!

Issue #3444, opened 2020-12-03
poetry fails to resolve constraints of both python and markers


bentheiii:
- [x] I am on the [latest](https://github.com/python-poetry/poetry/releases/latest) Poetry version.
- [x] I have searched the [issues](https://github.com/python-poetry/poetry/issues) of this repo and believe that this is not a duplicate.
- [x] If an exception occurs when executing a command, I executed it again in debug mode (`-vvv` option).

- **OS version and name**: windows/linux (reproduced both in windows and WSL)
- **Poetry version**: 1.1.4
- **Link of a [Gist](https://gist.github.com/) with the contents of your pyproject.toml file**: https://gist.github.com/bentheiii/4a54c1d59fe40eb87dbe31711fdf8f8f

## Issue

The toml file provided fails to lock, due to a version conflict on pytype and python. The message is:

```
  The current project's Python requirement (>=3.7,<4.0) is not compatible with some of the required packages Python requirement:
    - pytype requires Python <3.9,>=3.5, so it will not be satisfied for Python >=3.9,<4.0
    - pytype requires Python <3.9,>=3.5, so it will not be satisfied for Python >=3.9,<4.0
    - pytype requires Python <3.9,>=3.5, so it will not be satisfied for Python >=3.9,<4.0
    - pytype requires Python <3.9,>=3.5, so it will not be satisfied for Python >=3.9,<4.0
    - pytype requires Python <3.9,>=3.5, so it will not be satisfied for Python >=3.9,<4.0
    - pytype requires Python <3.9,>=3.5, so it will not be satisfied for Python >=3.9,<4.0
    - pytype requires Python <3.9,>=3.5, so it will not be satisfied for Python >=3.9,<4.0
    - pytype requires Python <3.9,>=3.5, so it will not be satisfied for Python >=3.9,<4.0
    - pytype requires Python <3.9,>=3.5, so it will not be satisfied for Python >=3.9,<4.0
    - pytype requires Python <3.9,>=3.5, so it will not be satisfied for Python >=3.9,<4.0
    - pytype requires Python <3.9,>=3.5, so it will not be satisfied for Python >=3.9,<4.0
    - pytype requires Python <3.9,>=3.5, so it will not be satisfied for Python >=3.9,<4.0
    - pytype requires Python <3.9,>=3.5, so it will not be satisfied for Python >=3.9,<4.0
    - pytype requires Python <3.9,>=3.5, so it will not be satisfied for Python >=3.9,<4.0
[...]

Issue #4965, opened 2022-01-01
poetry doesn't parse python constraints when markers are present


maciejskorski:
- [x] I am on the [latest](https://github.com/python-poetry/poetry/releases/latest) Poetry version.
- [x] If an exception occurs when executing a command, I executed it again in debug mode (`-vvv` option).

Clarifies in-depth the common root-cause for several reported issues
#3444 
#4959
#4958

- **OS version and name**: Linux
- **Poetry version**: 1.2.0a2
- **Python** 3.8

## Issue

`poetry` fails to parse python constraints when markers appear. 

This should be supported as per the [official documentation](https://python-poetry.org/docs/dependency-specification/#expanded-dependency-specification-syntax).

This also has a side effect on the performance and the lock file, as this way packages bypass early compatibility checks on versions; also wrong versions may get installed.

**Reproduce**
The following test confirms that `poetry` does not populate python constraints due to markers
```
[tool.poetry.dependencies]
python = "3.8"
pandas = [
    {version="1.3.2",python="3.9",markers="sys_platform=='linux'"},
    {version="1.3.1",python="3.8"}
]
```
then setting a breakpoint inside `complete_package` [method](https://github.com/python-poetry/poetry/blob/f6022eade7485a3b017ef0c8060dffed12e3cdb2/src/poetry/puzzle/provider.py#L493) gives 
```
# [(d.name+str(d._constraint),d._python_constraint) for d in requires]
[('pandas1.3.2', <VersionRange (*)>), ('pandas1.3.1', <Version 3.8>)]
```
Furthermore, due to the broken constraint poetry would pick up the wrong (1.3.2) version to install.

maciejskorski:
The issue is probably inside the function `poetry.core.factory.configure_package`, which translates `poetry.local_config` (json-like) into `package` (python object)

https://github.com/python-poetry/poetry-core/blob/6215dd18e05b17ea94271fb6bedb127a90a820c7/src/poetry/core/factory.py#L110)

Issue #4959, opened 2021-12-31
poetry selects wrong python version when markers present


maciejskorski:
- [x] I am on the [latest](https://github.com/python-poetry/poetry/releases/latest) Poetry version.
- [x] I have searched the [issues](https://github.com/python-poetry/poetry/issues) of this repo and believe that this is not a duplicate.
- [x] If an exception occurs when executing a command, I executed it again in debug mode (`-vvv` option).

- **OS version and name**: linux + Python 3.8.10
- **Poetry version**: 1.2.0a2 (recent master)

## Issue
Multiple constraints on python version fail when combined with valid markers.

This does work fine
```
# selecting 1.3.1 matching the system
[tool.poetry.dependencies]
python = "~3.8"
pandas = [
    {version="1.3.2",python='~3.9'},
    {version="1.3.1",python='~3.8'},
]
```
but not this
```
# selecting incompatible 1.3.2 and ignoring 1.3.1 - error
[tool.poetry.dependencies]
python = "~3.8"
pandas = [
    {version="1.3.2",python='~3.9',markers="sys_platform=='linux'"},
    {version="1.3.1",python='~3.8'},
]
```
when I move everything to markers it works fine again

```
# selecting 1.3.1 - fine!
[tool.poetry.dependencies]
python = "~3.8"
pandas = [
    {version="1.3.2",markers="python_version=='3.9' and sys_platform=='linux'"},
    {version="1.3.1",markers="python_version=='3.8'"},
]
```

radoering:
Probably duplicate of #3444

maciejskorski:
Related to #3444 but highlights bit different issues.  The logs show that `poetry` this these are different packages.
```
  • Installing pandas (1.3.1)
  • Installing pandas (1.3.2)
 ```

PR #4970, opened 2022-01-01
fix to python constraints ignored when markers present #4965

maciejskorski:
# Pull Request Check List

Resolves: 

#3444
#4959
#4958
#4965 

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -425,8 +425,8 @@ importlib-metadata = {version = ">=1.7.0", markers = "python_version < \"3.8\""}
 [package.source]
 type = "git"
 url = "https://github.com/python-poetry/poetry-core.git"
-reference = "1.1.0a6"
-resolved_reference = "52b7cdcba04549f40aac7bc430ec410b6d2c7e6a"
+reference = "af08f1c"
+resolved_reference = "af08f1ce720da467c9bf3d43eed3d9ebaf4ad7fb"
 
 [[package]]
 name = "pre-commit"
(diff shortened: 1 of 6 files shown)


dimbleby:
Should this be a fix to the classmethod on poetry-core's `Factory`?

maciejskorski:
> Should this be a fix to the classmethod on poetry-core's `Factory`?

@dimbleby I think so! For the guilty lines please have a look to #4965. Either we propose to change `create_dependencies` in `poetry.core`, or apply a patch to the inherited class in `poetry`?

dimbleby:
If the fix belongs in poetry-core, surely just make an MR against that project?  I'm not seeing why this is "more complicated"!

maciejskorski:
@dimbleby @radoering appreciate your feedback.

> If the fix belongs in poetry-core, surely just make an MR against that project? I'm not seeing why this is "more complicated"!

Not clear where the fix belongs to. Note that `poetry.core.factory` exposes a base/template class, which is then rewritten in `poetry.factory`. Isn't perfectly legit to fix the problem in the inherited and complete class, rather than in a template?

Furthermore, think of side effects when we change the template class from `poetry.core.factory`, this will escalate to other descendent, tests included. 

Please advise, but don't be too quick. I don't know the code well enough to see there will be no side effects :-)

dimbleby:
I don't see any upside in having two slightly different implementations for `create_dependency()`, one of which has a bug and one of which doesn't.

radoering:
Without having the difference between poetry's `Factory` class and poetry-core's base `Factory` class investigated further, what I don't like is the amount of code duplication if the change is made in the sub class. If the change can be made in the base class without importing anything from poetry and has the nature of a fix (instead of a new feature), it probably belongs to poetry-core in my opinion.

As for side effects, I hope they will be revealed by reviews and existing tests.

neersighted (changes requested):
Can you please rebase this onto the latest poetry-core release which includes the fixes to make these tests pass?

miigotu:
I just hit this bug with this constraint:
`cryptography = {version = "<35.0.0", markers = "sys_platform == 'linux' and platform_machine in 's390x, ppc64le,'", python = "<3.8"}`

neersighted:
Hi @miigotu -- this PR is stale and the issues it set out to address have been solved in a successor PR against poetry-core. If you are running into an issue you believe is related, please open a new bug report with a pyproject.toml, and detailed comparisons of what Poetry is doing vs. what you expect it to be doing.

Thanks!

PR #4977, opened 2022-01-04
utils/appdirs.py: raise exception if fail to get appdata folder

ombschervister:
When Poetri tries to get the user's folder, it does not check the result of the `SHGetFolderPathW` function.


Secrus:
This is now handled by `platformdirs` package. Closing

PR #4985, opened 2022-01-05
Wrap long lines to comply with flake8 E501

stinodego:
Resolves: #4776

Hi! This is my first PR for `poetry`; please let me know how I can improve my work.

Because of the nature of the issue being fixed (linting rules), this PR touches a lot of files. There are no functional changes, however, so I figured it would be OK for a single PR. Let me know if you would like me to break it up into smaller PRs (an option would be `tests` first, then `src`, for example).

# Changes
**_(updated 2020-01-08)_**

* Updated `.flake8` to remove the `E501` error (line too long) from the list of ignored warnings.
* ~Added `poetry` to `isort` namespace packages - see discussion below for the reason why.~
* Explicitly ignored the `E501` error for some long test function names and the occasional URL/path/hash.
* Apply `black --experimental-string-processing` to handle long strings. Added this functionality to `pre-commit`.
* Manually wrap some strings that `black` did not format (notably strings without spaces).
* Manually wrap multi-line strings.
* Manually wrap some docstrings and in-line comments.
* Fixed a typo in `src/poetry/console/commands/source/add.py`: "already exits" -> "already exists"
* Fixed a few more rare typos in docs/comments.

# String wrapping approach
**_(updated 2020-01-07)_**

I try to be consistent with the formatting guidelines of `black`, and the style of the rest of the code base.

For long regular strings that are not handled by `black`, I wrap them using implicit string concatenation, with **leading** space and if it's an f-string, an 'f' only at the lines where it's required:

```python
# Original
long_string = f"This string is way too long for a code base with a maximum character limit of {88}."

# Wrapped
long_string = (
    "This string is way too long for a code base"
    f" with a maximum character limit of {88}."
)
```

For multi-line strings, I use backslash breaks with **leading** space:
```python
# Original
long_string = """\
This string is way too long for a code base with a maximum character limit of 88.
"""

# Wrapped
long_string = """\
This string is way too long for a code base\
 with a maximum character limit of 88.
"""
```

### DIFF
--- a/src/poetry/__init__.py
+++ b/src/poetry/__init__.py
@@ -1,4 +1,1 @@
-__path__ = __import__("pkgutil").extend_path(
-    __path__,  # type: ignore[has-type]
-    __name__,
-)
+__path__ = __import__("pkgutil").extend_path(__path__, __name__)  # type: ignore[has-type]  # noqa: E501
--- a/src/poetry/__init__.py
+++ b/src/poetry/__init__.py
@@ -1,1 +1,4 @@
-__path__ = __import__("pkgutil").extend_path(__path__, __name__)  # type: ignore[has-type]
+__path__ = __import__("pkgutil").extend_path(
+    __path__,  # type: ignore[has-type]
+    __name__,
+)
--- a/src/poetry/utils/env.py
+++ b/src/poetry/utils/env.py
@@ -975,7 +975,8 @@ class EnvManager:
             else flags.pop("no-setuptools", True)
         )
 
-        # we want wheels to be enabled when pip is required and it has not been explicitly disabled
+        # we want wheels to be enabled when pip is required and it has not been
+        # explicitly disabled
         flags["no-wheel"] = (
             not with_wheel
             if with_wheel is not None
@@ -1184,7 +1185,8 @@ class Env:
         """
         Path to current pip executable
         """
-        # we do not use as_posix() here due to issues with windows pathlib2 implementation
+        # we do not use as_posix() here due to issues with windows pathlib2
+        # implementation
         path = self._bin(self._pip_executable)
         if not Path(path).exists():
             return str(self.pip_embedded)
--- a/src/poetry/utils/env.py
+++ b/src/poetry/utils/env.py
@@ -262,7 +262,8 @@ class SitePackages:
 
         if not results and strict:
             raise RuntimeError(
-                f'Unable to find a suitable destination for "{path}" in {paths_csv(self._candidates)}'
+                f'Unable to find a suitable destination for "{path}" in'
+                f" {paths_csv(self._candidates)}"
             )
 
         return results
@@ -416,7 +417,10 @@ class EnvCommandError(EnvError):
     def __init__(self, e: CalledProcessError, input: Optional[str] = None) -> None:
         self.e = e
 
-        message = f"Command {e.cmd} errored with the following return code {e.returncode}, and output: \n{decode(e.output)}"
+        message = (
+            f"Command {e.cmd} errored with the following return code {e.returncode},"
+            f" and output: \n{decode(e.output)}"
+        )
         if input:
             message += f"input was : {input}"
         super().__init__(message)
@@ -481,7 +485,8 @@ class EnvManager:
                         [
                             python,
                             "-c",
-                            "\"import sys; print('.'.join([str(s) for s in sys.version_info[:3]]))\"",
+                            "\"import sys; print('.'.join([str(s) for s in"
+                            ' sys.version_info[:3]]))"',
                         ]
                     ),
                     shell=True,
@@ -727,7 +732,8 @@ class EnvManager:
                         [
                             python,
                             "-c",
-                            "\"import sys; print('.'.join([str(s) for s in sys.version_info[:3]]))\"",
+                            "\"import sys; print('.'.join([str(s) for s in"
+                            ' sys.version_info[:3]]))"',
                         ]
                     ),
                     shell=True,
@@ -802,7 +808,8 @@ class EnvManager:
                         [
                             executable,
                             "-c",
-                            "\"import sys; print('.'.join([str(s) for s in sys.version_info[:3]]))\"",
+                            "\"import sys; print('.'.join([str(s) for s in"
+                            ' sys.version_info[:3]]))"',
                         ]
                     ),
                     shell=True,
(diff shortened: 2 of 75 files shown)


finswimmer:
Hello @stinodego,

thanks a lot for the work you've already done :+1: 

It took a while but I've found out what's going on with `isort`. It's due to the change in `src/poetry/__init__.py`. See https://github.com/PyCQA/isort/issues/1879

To fix it we can put this in the isort config section in the `pyproject.toml`:

```toml
namespace_packages = ["poetry"]
```

Regarding  the string wrappings: 

`black` has an experimental flag `--experimental-string-processing` (See [Release Notes](https://black.readthedocs.io/en/stable/change_log.html#id39)) This is already used by larger projects like django, pandas and sqlalchemy and will become default in a future version (see https://github.com/psf/black/issues/2188). I use this in my project as well. I think we can/should add this to Poetry as well.

Because `black` has some different opinion about the wrapping then how you have done it, I would ask you to revert your string wrapping changes first and run black with the above flag. :pray: 

fin swimmer

stinodego:
Hi @finswimmer ! Thanks for your feedback, I will implement the changes you requested.

And kudos for finding out what caused that `isort` behaviour; definitely an obscure corner case!

Will redo this PR this weekend.

stinodego:
PR has been re-done! Looking forward to your feedback.

PS: I see one CI check failing; I don't believe this is something I can fix?

src/poetry/__init__.py
```diff
@@ -1 +1,4 @@
-__path__ = __import__("pkgutil").extend_path(__path__, __name__)  # type: ignore[has-type]
+__path__ = __import__("pkgutil").extend_path(
+    __path__,  # type: ignore[has-type]
+    __name__,
+)
```
finswimmer, line 4: I'm afraid this change still makes some trouble. For example PyCharm isn't able to follow `poetry.core` imports.

I vague remember that this code line is quite critical for pkgutils namespace packages. But I don't understand why exactly :man_shrugging: So I'm afraid we have to make sure that this stays a one-liner and is the first line of this file (and can than remove the `namespace_packages` setting for isort again). :(

finswimmer (changes requested):
Looks quite good to me. Thanks a lot :+1: 

Can you please add the `--experimental-string-processing` flag to the list of args for `black` in the `.pre-commit-config.yaml`?

src/poetry/__init__.py
stinodego: Funny how much trouble a small, seemingly innocuous change can cause :D I have added a new commit reverting the change, and just slapped on a `  # noqa: E501` here.

stinodego:
> Can you please add the `--experimental-string-processing` flag to the list of args for `black` in the `.pre-commit-config.yaml`?

Done! Had to exclude the `get-poetry.py` and `install-poetry.py` scripts, because those do not comply. I borrowed the `exclude` regex from the pyupgrade hook config.

finswimmer:
> Actually, would it not be better to include this setting in the pyproject.toml? 

This is a very good idea :+1: 

>  Had to exclude the get-poetry.py and install-poetry.py scripts, because those do not comply.

Hm, I don't see any reasons why we should exclude them. What problems have you faced?

stinodego:
> > Had to exclude the get-poetry.py and install-poetry.py scripts, because those do not comply.
> 
> Hm, I don't see any reasons why we should exclude them. What problems have you faced?

These scripts do not follow the string formatting rules of `black`. So if we do not exclude them, pre-commit will fail. So we should exclude them either in `pyproject.toml` (which I have done now) or in the `pre-commit` config.

finswimmer:
> I don't mind formatting these scripts too, but I understood that these scripts were 'frozen' and not allowed to be changed.

It's fine to change these scripts if needed. Here we have just changes of the format. We should not exclude these scripts from `black`.

We have fixed the failing check  in the meantime (Well, actually we drop it). You will need to rebase your branch onto master to pass all checks.

stinodego:
All right, done!

src/poetry/utils/env.py
```diff
@@ -481,7 +487,8 @@ def activate(self, python: str, io: "IO") -> "Env":
                         [
                             python,
                             "-c",
-                            "\"import sys; print('.'.join([str(s) for s in sys.version_info[:3]]))\"",
+                            "\"import sys; print('.'.join([str(s) for s in"
+                            ' sys.version_info[:3]]))"',
```
finswimmer, line 491: I wonder if we should keep the original format here for better readability. What do you think? (same for the lines 730 and 805, 859)

This can be achieved by `#fmt: off/on` (https://stackoverflow.com/a/58584557/9750706)

tests/installation/test_executor.py
```diff
@@ -252,11 +252,23 @@ def test_execute_works_with_ansi_output(
...
+        " \x1b[39m\x1b[36mpytest\x1b[39m\x1b[39m"
+        " (\x1b[39m\x1b[39;1m3.5.2\x1b[39;22m\x1b[39m)\x1b[39m:"
+        " \x1b[34mDownloading...\x1b[39m",
+        "\x1b[34;1m•\x1b[39;22m \x1b[39mInstalling"
+        " \x1b[39m\x1b[36mpytest\x1b[39m\x1b[39m"
+        " (\x1b[39m\x1b[39;1m3.5.2\x1b[39;22m\x1b[39m)\x1b[39m:"
+        " \x1b[34mInstalling...\x1b[39m",
+        "\x1b[32;1m•\x1b[39;22m \x1b[39mInstalling"
+        " \x1b[39m\x1b[36mpytest\x1b[39m\x1b[39m"
+        " (\x1b[39m\x1b[32m3.5.2\x1b[39m\x1b[39m)\x1b[39m",  # finished
```
finswimmer, line 271: IMO the original formatting is better readable. Maybe we can agree as a rule of thumb to turn off formatting (like shown in https://stackoverflow.com/a/58584557/9750706) for expected strings in test in general?

finswimmer (changes requested):

src/poetry/utils/env.py
stinodego: In this case, because the exact same string is repeated 4 times in this class, I'd suggest defining a string constant in the class (next to `ENVS_FILE`) called something like `VERSION_SCRIPT`.

Then we won't have to wrap it, and won't have to fiddle with black formatting.

tests/installation/test_executor.py
stinodego: I agree that the result here is not very readable here. I can fix this one and the one in `tests/utils/test_helpers.py`; the autoformatting is fine in all other cases as far as I can see.

Most expected strings are multi-line strings, which `black` doesn't touch. Do you wish to keep the original formatting there as well? I can revert my manual changes. Ignoring errors for these is actually quite easy by putting a `# noqa: E501` after the final `"""`.

src/poetry/utils/env.py
finswimmer: Good idea  :+1: I saw there is already a `GET_PYTHON_VERSION` at the top of the file. I think we can use this one.

tests/installation/test_executor.py
finswimmer: LGTM :+1:

src/poetry/utils/env.py
stinodego: Even better! Adjusted the last commit to use the existing constant.

finswimmer (approved):
Beautiful :+1: :1st_place_medal: :rocket:

Issue #4862, opened 2021-12-02
Virtual environment name mapping - make more portable, add ability to override, store as project ID?


zEdS15B3GCwq:
- [x] I have searched the [issues](https://github.com/python-poetry/poetry/issues) of this repo and believe that this is not a duplicate.
- [x] I have searched the [documentation](https://python-poetry.org/docs/) and believe that my question is not covered.

## Feature Request
Poetry, if allowed to automatically manage them, creates virtual environments with names roughly like <project name>-<hash>-<python version>. This design choice makes life simple in some cases, but adds a _hidden state_ to the project that complicates things in some situations. Mainly, the hash depends on information outside the scope of the project (e.g. project folder path), which is a questionable design choice in my opinion. I think that adding a little flexibility in this regard would make Poetry a better tool overall.

Just to mention a few cases where the current name mapping causes problems: #3459 and #4654. Briefly, the hash in the virtual env name seems to depend on the path where the project is located and perhaps other things. Rename the project folder - poetry will create a new virtual env and you'll need to reinstall all the packages from zero; plus the old env will linger forgotten. If you share workspace settings on multiple computers (stored in project folders, e.g. for VSCode), the env path will need to be updated each time you switch to another computer; similar in multi-user situations.

@sinoroc asks in #3459, how a design change would help Poetry? (S)He also mentions that the current design makes things simpler by not having to remember the virtual env's name, like you need to if you use `venv` or `Conda`. I think Poetry only makes things simpler if you use it the way it's supposed to be used, but doesn't have the flexibility to move beyond that. New users, who are not familiar with virtual environment management, might find the automated environment creation useful ... until they rename the project folder ...  and then get a big surprise when their environment is gone without warning. Moreover, while managing workspace settings for VSCode (example above) has nothing to do with Poetry, easier integration with other tools is better for Poetry, isn't it?

In my case, I use `pyenv` (`pyenv-win` on Windows) to manage Python versions; manage virtual environments with `venv`, and let `poetry` manage packages inside virtual envs. However, the only thing I need `venv` for, is to get the virtual environment name I want - if only poetry would let me change the name mapping, I could drop `venv` from the pipeline.

[...]

PR #5001, opened 2022-01-07
New config option to allow no hash in venv name

zEdS15B3GCwq:
# Pull Request Check List

Resolves: feature request #4862

### Summary

Adds config option `virtualenvs.path-independent_naming` that omits the project path-dependent hash from newly created venv names.

The rationale behind this is that the hash was inconvenient in some use cases, e.g. project folders could not be moved without "losing" the existing venv; and sharing project folders (cloud storage or network share) was inconvenient as some local IDE (VSCode) settings would conflict between computers depending on the project path. This config option eliminates these issues.

### Usage example:
```
mkdir projectname
cd projectname
poetry config --local virtualenvs.path-independent_naming true
poetry init # use default name etc.
poetry show # implicitly creates venv "projectname-pyX.Y", without path hash
```

Now the project folder can be moved, and poetry will keep using the same venv.

This setting is intended to be for **personal** use, i.e. NOT shared on git, same as other poetry settings.

### Potential problems

Can confuse poetry if there are multiple projects with the same package name. However, this setting is not enabled by default, and I am assuming that users who enable it know what they're doing. And there's a warning in the documentation.

Otherwise, same potential issues as with `in-project`, i.e. changing the config when venvs already exist for the project causes new venvs to be created and the previous one is "lost" to poetry, unless the setting is reverted. This is not new.

### Implementation

The referred issue (#4862) broadly talks about several possible solutions to the problems I mentioned here, and while I do believe that _poetry_'s handling of venvs needs some rethinking from the ground up, here I aimed to minimize changes. The default behaviour is unaffected, so users won't notice the difference unless they need this option.

The code for `path-independent_naming` is a copy of `in-project`, at least as much as possible.

1) The biggest change is that `generate_venv_name` used to be a _class method_ of `EnvManager`, but in order for it to be able to access the configuration, it was necessary to change it to a normal method. While this change is substantial in theory, I believe keeping it a _class method_ and adding parameters would've been more disruptive. 

2) Naturally, `generate_venv_name` accesses the configuration and if this setting is `true`, it skips adding the hash to the venv name.

3) Configuration documentation updated (config --list part and added new entry for option)

4) New option added as a copy of `in-project` in both configuration code and tests.

5) Tests changed from using class method to using EnvManager instance methods. Other code wasn't affected, I believe.

### Naming

I tried to give it a name that describes what it can be used for, so that users who don't know how the venv names are generated can still understand the purpose. However, I feel that `path-independent_naming` is somewhat complicated and perhaps not descriptive enough. Suggestions are welcome.


zEdS15B3GCwq:
Is the "Tests / FreeBSD / PYTHON:python3.6 Failing after 4s" error something that I'm supposed to fix? It seems to fail before even running the tests.

zEdS15B3GCwq:
I'm aware that this PR is behind master and has conflicts due to other changes to config options. I'll handle that when/if it gets any attention.

neersighted:
Sorry to get to this after so much time -- as you can see, this repo is busy and larger changes can be hard to review. We're clearing the 1.2 release hangover so hopefully the backlog will get attacked more aggressively in the coming weeks.

Anyway, I am not sure there is a compelling reason to merge/introduce this. Poetry's virtual environment management is intentionally very minimal and simple. This is because Poetry has always respected an existing virtual environment and the bail-out for users wanting more than Poetry offers is to just use whatever pre/non-Poetry workflow and tooling they prefer.

Does this really offer anything over activating an environment and managing it however you like? Poetry's two settings of "treat the environment as a hidden implementation detail and abstract it for me" and "put it in the project" seem to be more than sufficient to me, and indeed, most use cases I have discussed or encountered.

zEdS15B3GCwq:
Hi, thanks for taking time to review this.

> Does this really offer anything over activating an environment and managing it however you like? Poetry's two settings of "treat the environment as a hidden implementation detail and abstract it for me" and "put it in the project" seem to be more than sufficient to me, and indeed, most use cases I have discussed or encountered.

I'd love to let Poetry manage the virtualenvs for me, but the current "take it or leave it" approach doesn't let me do that. Yes, I can manage my own environments, or I can just change Poetry to suit my own ways - and I do, but I'd rather I didn't need to. In my opinion, being configurable doesn't mean that Poetry can't be a tool that just takes care of things for you - it can enable Poetry to work well for more users.

To be honest, I think this option shouldn't be needed, because Poetry virtualenvs shouldn't depend on the project path in the first place - but then I'd be talking about design decisions, which is a more complex topic.

Ultimately, I have no idea how many other users would benefit from this setting. If you think not many, then there's no point in including it, indeed.

Secrus:
@zEdS15B3GCwq thank you for your effort. Since there wasn't much progress on that feature (and it is not a direction we want to go with) I am going to close this.

Issue #4566, opened 2021-09-28
Poetry after 1.1.9 fails to recognize conda environments as valid


thurse93:
- [x] I am on the [latest](https://github.com/python-poetry/poetry/releases/latest) Poetry version.
- [x] I have searched the [issues](https://github.com/python-poetry/poetry/issues) of this repo and believe that this is not a duplicate.
- [x] If an exception occurs when executing a command, I executed it again in debug mode (`-vvv` option).

- **OS version and name**: Windows 10 20H2
- **Poetry version**: 1.1.9

## Issue
Since poetry version 1.1.9 poetry does not recognize valid virtual environments created by conda.

We use conda (via tox-conda) to test our libraries with different versions of python (since we're in a corporate network disconnected from the internet we have to rely on repositories for the python interpreters). Minimal tox example to reproduce the problem:

tox.ini
```
[tox]
requires = tox-conda
envlist = debug

[testenv]

[testenv:debug]
deps =
    poetry == 1.1.8
commands =
    python -m poetry env info
```
Produces something like:
```
debug run-test: commands[0] | python -m poetry env info

Virtualenv
Python:         3.7.11
Implementation: CPython
Path:           C:\path\to\project\.tox\debug
Valid:          True
```
Whereas
```
[tox]
requires = tox-conda
envlist = debug

[testenv]

[testenv:debug]
deps =
    poetry == 1.1.9
commands =
    python -m poetry env info
```
Leads to a broken environment (in the eyes of poetry):
```
debug run-test: commands[0] | python -m poetry env info

Virtualenv
Python:         3.7.11
Implementation: CPython
Path:           C:\path\to\project\.tox\debug
Valid:          False
```

We think this is because `conda create` does not put the python executables inside the Scripts directory like virtualenv does. Instead, they are placed in the root directory of the virtualenv folder.

If we understood https://github.com/python-poetry/poetry/pull/4507 right, poetry 1.1.9 expects all binaries in the Scripts (or bin) Directory of the virtualenv directory.

zEdS15B3GCwq:
I'm also having the same issue. Tested poetry 1.1.9 through 1.1.11 (latest).

My setup is the following:

- Windows 10
- conda 4.10.3 (latest)
- system Python 3.9.9
- poetry installed 1.1.9 ... 1.1.11
- conda environment created with `conda create -n test python=3.9 pip=21.3`

pyproject.toml:

    [tool.poetry]
    name = "test"
    version = "0.1.0"
[...]

PR #5007, opened 2022-01-08
Fix: conda environment is detected as broken (#4566)

finswimmer:
`conda` puts the `python` and `pip` executable on windows in different location. So poetry is unable to detect a valid venv. This PR fixes this.

There is also a port for 1.1: https://github.com/python-poetry/poetry/pull/5008

Resolves: https://github.com/python-poetry/poetry/issues/4566

### DIFF
--- a/src/poetry/utils/env.py
+++ b/src/poetry/utils/env.py
@@ -1073,6 +1073,7 @@ class Env:
     def __init__(self, path: Path, base: Optional[Path] = None) -> None:
         self._is_windows = sys.platform == "win32"
         self._is_mingw = sysconfig.get_platform().startswith("mingw")
+        self._is_conda = bool(os.environ.get("CONDA_DEFAULT_ENV"))
 
         if not self._is_windows or self._is_mingw:
             bin_dir = "bin"
@@ -1133,10 +1134,15 @@ class Env:
     def parent_env(self) -> "GenericEnv":
         return GenericEnv(self.base, child_env=self)
 
-    def find_executables(self) -> None:
+    def _find_python_executable(self) -> None:
+        bin_dir = self._bin_dir
+
+        if self._is_windows and self._is_conda:
+            bin_dir = self._path
+
         python_executables = sorted(
             p.name
-            for p in self._bin_dir.glob("python*")
+            for p in bin_dir.glob("python*")
             if re.match(r"python(?:\d+(?:\.\d+)?)?(?:\.exe)?$", p.name)
         )
         if python_executables:
@@ -1146,6 +1152,7 @@ class Env:
 
             self._executable = executable
 
+    def _find_pip_executable(self) -> None:
         pip_executables = sorted(
             p.name
             for p in self._bin_dir.glob("pip*")
@@ -1157,6 +1164,10 @@ class Env:
                 pip_executable = pip_executable[:-4]
 
             self._pip_executable = pip_executable
+
+    def find_executables(self) -> None:
+        self._find_python_executable()
+        self._find_pip_executable()
 
     def get_embedded_wheel(self, distribution: str) -> Path:
         return get_embed_wheel(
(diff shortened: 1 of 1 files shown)


zEdS15B3GCwq:
The changes in `_init_` and `find_executables` seem straightforward, but I would like to understand the `_bin` part. Just some ideas off the top of my head, pardon me if my ideas are way off.

```
                if bin_path.exists():
                    return str(bin_path)
                elif Path(self._bin_dir.parent / (bin + ".exe")).exists():
                    return str(Path(self._bin_dir.parent / (bin + ".exe")))
```

Wouldn't it make sense to make use of `self._is_conda` here as well to make the _elif_ branch more specific?

Also, though this is not strictly about your PR, the `self._is_windows and not bin.endswith(".exe")` case seems to be unnecessarily replicated inside the `if not bin_path.exists():` case, which is also only Windows-specific ... it appears that this needs simplification. 

I'm talking about this section in `env.py`:

```
        if self._is_windows and not bin.endswith(".exe"):
            bin_path = self._bin_dir / (bin + ".exe")
        else:
            bin_path = self._bin_dir / bin

        if not bin_path.exists():
            # On Windows, some executables can be in the base path
            # This is especially true when installing Python with
            # the official installer, where python.exe will be at
            # the root of the env path.
            if self._is_windows:
                if not bin.endswith(".exe"):
                    bin_path = self._bin_dir / (bin + ".exe")
                else:
                    bin_path = self._path / bin

                if bin_path.exists():
                    return str(bin_path)

            return bin
```

Or is it an oversight and the inside one should look like `self._path / (bin + ".exe")`?

```
        if not bin_path.exists():
            # On Windows, some executables can be in the base path
            # This is especially true when installing Python with
            # the official installer, where python.exe will be at
            # the root of the env path.
            if self._is_windows:
                if not bin.endswith(".exe"):
                    bin_path = self._path / (bin + ".exe")
                else:
                    bin_path = self._path / bin
```

In this case, it might be clearer to do this:

```
        if self._is_windows and not bin.endswith(".exe"):
            bin_path = self._bin_dir / (bin + ".exe")
        else:
            bin_path = self._bin_dir / bin

        if bin_path.exists():
            return str(bin_path)

        # On Windows, some executables can be in the base path
        # This is especially true when installing Python with
        # the official installer, where python.exe will be at
        # the root of the env path.
        if self._is_windows:
            if not bin.endswith(".exe"):
                bin_path = self._path / (bin + ".exe")
            else:
                bin_path = self._path / bin

            if bin_path.exists():
                return str(bin_path)

        # Windows, Conda blah-blah explanation
        if self._is_windows and self_is_conda:
            if not bin.endswith(".exe"):
                bin_path = self._path.parent / (bin + ".exe")
            else:
                bin_path = self._path.parent / bin

            if bin_path.exists():
                return str(bin_path)

        return bin
```

I think this is a clearer structure. Just an illustration, not tested code.

Of course, all `not bin.endswith(".exe")` could be reduced to a single check and modification at the start:

```
        if self._is_windows and not bin.endswith(".exe"):
            bin = bin + ".exe"
```

zEdS15B3GCwq:
On the same note, in `find_executables`, perhaps

```
    def find_executables(self) -> None:
        if self._is_windows and self._is_conda:
            python_bin_dir = self._bin_dir.parent
        else:
            python_bin_dir = self._bin_dir
```

could be

```
    def find_executables(self) -> None:
        if self._is_windows and self._is_conda:
            python_bin_dir = self._path  # <<< this?
        else:
            python_bin_dir = self._bin_dir
```

finswimmer:
Thanks a lot for your feedback @zEdS15B3GCwq. :+1: 

I'm aware that there are some ugly parts that should be improved. That's why its a draft pr and I called it a proof-of-concept. Once I get some feedback in the corresponding issue whether it works as expected or not I will start improving the code.

fin swimmer

zEdS15B3GCwq (commented):
sorry, ignore this

finswimmer:
> sorry, ignore this

No, I will not. You've given some good points :smiley:

zEdS15B3GCwq:
> > sorry, ignore this
> 
> No, I will not. You've given some good points 😃

That's nice of you, but I meant that particular message (or did you known already).

neersighted (approved):

PR #5013, opened 2022-01-09
Update test fixtures egg-info formatting

stinodego:
Relates-to: #3155

I did not make a separate issue for this, but it could fall under 'improve test suite'. I briefly mentioned this on Discord and figured I'd just research and write a fix.

The specific issue here is that when running tests, three files are always being modified in the folder `tests/fixtures/git/github.com/demo/demo/demo.egg-info`:
* `dependency_links.txt`
* `PKG-INFO`
* `SOURCES.txt`

The offending test is `test_executor_should_write_pep610_url_references_for_git`.

# Cause

After some research, I found out that the resulting changes are not due to some bug, but pip actually creates the `egg-info` in this format.

For `dependency_links.txt` and `SOURCES.txt` this was easy to see; the pre-commit end-of-file fixer messed up the original formatting of pip. `dependency_links.txt` should really be an empty file with two newlines, while `SOURCES.txt` should not have a trailing newline at all.

For `PKG-INFO` the new file format looks somewhat weird. It turns out that the `Description` in the original file is the value for the `long_description` parameter for the `setup` function (while `description` maps to `Summary`). It makes sense to put the long description at the bottom, as this usually contains a lot of newlines and messes up the formatting.

I'm not sure in which version of pip/setuptools these changes were introduced, but I believe the `egg-info` in the test suite should mirror the 'real' egg info as closely as possible. So the fix here would be accept this new formatting, and to disable the pre-commit end-of-file fixer for any `egg-info` fixtures.

# Changes

* Disabled pre-commit end-of-file fixer for `*.egg-info/` folder contents _(please double-check my regex!)_
* Updated `*.egg-info/` folder contents to adhere to the formatting resulting from using the latest pip version.

This takes care of the issue.

I am left wondering why that specific test even writes an `egg-info` in the first place, since it is not being installed as editable. Would appreciate if anyone can elucidate this for me!


jinzaizhichi (approved):

stinodego:
Hey @neersighted , have you managed to take a look at this? It seemed to me like quite a small/simple fix. Or is there something here that makes this difficult to judge?

abn (approved):
LGTM. I would really love for tests of this nature to be run with its own copy of things, isolated.

PR #5041, opened 2022-01-14
Change keyring dependency to >=21.2.0

NeilGirdhar:


NeilGirdhar:
This is a backport of https://github.com/python-poetry/poetry/pull/4003

neersighted (approved):

davegaeddert:
Just FYI, I think I hit a potential issue related to this in the 1.1.3 release: https://github.com/python-poetry/poetry/issues/5189#issuecomment-1036414909

NeilGirdhar:
@davegaeddert Can you help me understand why this works with old version, but not the new one?

davegaeddert:
@NeilGirdhar sure thing — I learned a little more about it over the weekend. Basically in [keyring v23.2.0 they added a backend for "libsecret"](https://github.com/jaraco/keyring/blob/main/CHANGES.rst#v2320) and I think the short answer is that, in a specific environment, it makes it seem like an available backend but then it throws an error when you actually use it. I think it's for them to fix ([I opened an issue](https://github.com/jaraco/keyring/issues/561) and [a PR to add a failing test](https://github.com/jaraco/keyring/pull/564), but not sure what the fix is yet). The environment where you hit the error is essentially a headless linux, where Poetry is installed alongside the the "gi" package (ex. `import gi`). In GitHub Actions, if you do a basic setup like this then you hit the error:
```yaml
on: push

jobs:
  test:
    runs-on: ubuntu-latest
    steps:
    - uses: actions/checkout@v2
    - run: pip3 install -U pip poetry
```

Let me know what you think.

NeilGirdhar:
@davegaeddert Awesome!  Thanks for getting the ball rolling with such a well-described issue, and explanation here.

PR #5075, opened 2022-01-20
Use master of poetry-core in master of poetry

radoering:
# Pull Request Check List

Due to the fact that a new poetry release comes together with a new poetry-core release and to avoid issues like #4958 (fixing a bug that has already been fixed upstream) and ease fixes for issues like #4919 (poetry-core downstream tests failing which require a fix in poetry), poetry master should depend on poetry-core master.

When creating a new release branch the dependency of poetry-core should be changed to a version dependency.

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -415,11 +415,18 @@ version = "1.1.0a6"
 description = "Poetry PEP 517 Build Backend"
 category = "main"
 optional = false
-python-versions = ">=3.6,<4.0"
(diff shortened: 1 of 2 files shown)


abn:
I don't think that is generally a sound plan. Largely because both are intended to be independent projects even if their development is parallel at present. Also would lead to breakages due to changes introduced in core, and that is unpleasant all around.

PR #5076, opened 2022-01-20
Added canonicalize_name() to console/commands/init.py

evanrittenhouse:
# Pull Request Check List

Resolves: #5035
Resolves: #5033

- [ ] Added **tests** for changed code - didn't add tests, but new code passes all existing tests.

### DIFF
--- a/src/poetry/console/commands/init.py
+++ b/src/poetry/console/commands/init.py
@@ -18,6 +18,7 @@ from tomlkit import inline_table
 
 from poetry.console.commands.command import Command
 from poetry.console.commands.env_command import EnvCommand
+from poetry.utils.helpers import canonicalize_name
 
 
 if TYPE_CHECKING:
@@ -275,10 +276,14 @@ You can specify a package in the following forms:
                 else:
                     choices = []
                     matches_names = [p.name for p in matches]
-                    exact_match = constraint["name"] in matches_names
+                    exact_match = canonicalize_name(constraint["name"]) in matches_names
                     if exact_match:
                         choices.append(
-                            matches[matches_names.index(constraint["name"])].pretty_name
+                            matches[
+                                matches_names.index(
+                                    canonicalize_name(constraint["name"])
+                                )
+                            ].pretty_name
                         )
 
                     for found_package in matches:
--- a/src/poetry/console/commands/init.py
+++ b/src/poetry/console/commands/init.py
@@ -245,20 +245,16 @@ You can specify a package in the following forms:
         return 0
 
     def _generate_choice_list(
-        self, matches: List["Package"], canonicalized_name: str, raw_name: str
-    ) -> Tuple[List[str], str]:
+        self, matches: List["Package"], canonicalized_name: str
+    ) -> List[str]:
         choices = []
         matches_names = [p.name for p in matches]
         exact_match = canonicalized_name in matches_names
-        info_string: str = (
-            f"Found <info>{len(matches)}</info> packages matching <c1>{raw_name}</c1>"
-        )
         if exact_match:
             choices.append(matches[matches_names.index(canonicalized_name)].pretty_name)
 
         for found_package in matches:
             if len(choices) >= 10:
-                info_string += "\nNOTE: Output has been truncated for readability"
                 break
 
             if found_package.name == canonicalized_name:
@@ -266,7 +262,7 @@ You can specify a package in the following forms:
 
             choices.append(found_package.pretty_name)
 
-        return choices, info_string
+        return choices
 
     def _determine_requirements(
         self,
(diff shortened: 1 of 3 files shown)


finswimmer (changes requested):
Thanks a lot. Looks good to me :+1: 

As discussed here my suggestion for a test:

You can take the [test_interactive_with_dependencies](https://github.com/python-poetry/poetry/blob/335f209c2dd96cb33e75279c4a52cb25b1966bc3/tests/console/commands/test_init.py#L122) as an example and play around with adding packages to the repo, give different spelling (upper/lower cases) of the package you search and always expect to find it in the first position.

src/poetry/console/commands/init.py
```diff
@@ -270,10 +271,14 @@ def _determine_requirements(
                 else:
                     choices = []
                     matches_names = [p.name for p in matches]
-                    exact_match = constraint["name"] in matches_names
+                    exact_match = canonicalize_name(constraint["name"]) in matches_names
```
neersighted, line 274: It would be good to use a local variable such as `canonicalized_name` instead of calling the helper twice.

neersighted (changes requested):

src/poetry/console/commands/init.py
```diff
@@ -237,6 +240,26 @@ def handle(self) -> int:
...
+        exact_match = canonicalized_name in matches_names
+        if exact_match:
+            choices.append(matches[matches_names.index(canonicalized_name)].pretty_name)
+
+        for found_package in matches:
+            if len(choices) >= 10:
+                break
+
+            if found_package.name.lower() == canonicalized_name:
+                continue
```
finswimmer, line 257: I think we must remove the `lower()` here now.

tests/console/commands/test_init.py
```diff
@@ -242,6 +248,24 @@ def test_interactive_with_git_dependencies(
     assert expected in tester.io.fetch_output()
 
 
+@pytest.mark.parametrize("package_name", ["flask", "Flask", "flAsK"])
+def test_generate_choice_list(tester: CommandTester, package_name: str):
```
finswimmer, line 252: Can you add another test, that shows that if there are more than 10 packages, the list is truncated and the package we are searching is still the first result?

tests/utils/test_helpers.py
```diff
@@ -79,3 +80,17 @@ def test_get_client_cert(config: "Config"):
     config.merge({"certificates": {"foo": {"client-cert": client_cert}}})
 
     assert get_client_cert(config, "foo") == Path(client_cert)
+
+
+def test_canonicalize_name():
+    names = ["flask", "Flask", "FLASK", "FlAsK", "fLaSk57", "flask-57"]
+    canonicalized_names = [canonicalize_name(name) for name in names]
```
finswimmer, line 87: I would prefer to have it as parameterize test. IMO  that would be better to read.

finswimmer (changes requested):

tests/console/commands/test_init.py
evanrittenhouse: @finswimmer I've added the truncation condition to the test you highlighted. Felt that extending the package list to > 10 items, then checking for both conditions which result from the `_generate_choice_list()` method would be cleaner. If you'd like me to break them out for separation of concerns, I can also do that - just wanted to keep it DRY.

src/poetry/console/commands/init.py
```diff
@@ -262,28 +285,13 @@ def _determine_requirements(
...
-                        if len(choices) >= 10:
-                            break
-
-                        if found_package.name.lower() == constraint["name"].lower():
-                            continue
-
-                        choices.append(found_package.pretty_name)
+                    choices = self._generate_choice_list(matches, canonicalized_name)
 
                     self.line(
```
finswimmer, line 296: While we are here. Can you extend this by an info in the case the list is truncated? Then we can close https://github.com/python-poetry/poetry/issues/5033 as well.

src/poetry/console/commands/init.py
```diff
@@ -237,6 +240,30 @@ def handle(self) -> int:
...
+            if len(choices) >= 10:
+                info_string += "\nNOTE: Output has been truncated for readability"
+                break
+
+            if found_package.name == canonicalized_name:
+                continue
+
+            choices.append(found_package.pretty_name)
+
+        return choices, info_string
```
finswimmer, line 265: Hm, we shouldn't create and return this info_string within the method. It would mean that this method is doing more than "one thing". Instead we should keep it there where it was before and just extend the message by a "Showing the first 10 results" or similar in the case of `len(matches) >= 10`.

src/poetry/console/commands/init.py
```diff
@@ -262,34 +285,24 @@ def _determine_requirements(
...
-                            break
-
-                        if found_package.name.lower() == constraint["name"].lower():
-                            continue
+                    choices = self._generate_choice_list(matches, canonicalized_name)
 
-                        choices.append(found_package.pretty_name)
-
-                    self.line(
+                    info_string: str = (
```
finswimmer, line 296: Maybe I'm a bit nitpicking here, but the type hint is unnecessary.

finswimmer (approved):
Thanks a lot for your contribution. Looks good to me :+1:

Issue #5038, opened 2022-01-14
`poetry lock --check` has no output


DetachHead:
- [X] I am on the [latest](https://github.com/python-poetry/poetry/releases/latest) Poetry version.
- [X] I have searched the [issues](https://github.com/python-poetry/poetry/issues) of this repo and believe that this is not a duplicate.
- [X] If an exception occurs when executing a command, I executed it again in debug mode (`-vvv` option).

- **OS version and name**: windows 10
- **Poetry version**: 1.2.0a2

## Issue
it would be nice if there was some sort of output other than just the exit code
```py
> poetry lock --check

> echo %errorlevel%
1
```

perhaps an error message like
```
Error: poetry.lock is out of date. run `poetry update` to fix it
```

radoering:
Sounds good. I am just wondering if `poetry update` is the most suitable recommendation or if `poetry lock [--no-update]` is better suited? It seemed natural to me to call `lock` if `lock --check` failed. That way you can decide if you only want to bring `pyproject.toml` and `poetry.lock` in line (without updating already locked dependencies) or if you want to update these, too. As far as I understand, `poetry lock` is the same as `poetry update --lock`. Probably, the advantage of `poetry update` over `poetry lock` is that you don't have to call `poetry install` separately.

artemrys:
@radoering I can take this one, it seems to be an easy change. Do we just need to add some a log here https://github.com/python-poetry/poetry/blob/335f209c2dd96cb33e75279c4a52cb25b1966bc3/src/poetry/console/commands/lock.py#L39-L44 and update / create tests?

radoering:
@artemrys: Yes, that seems to be the correct part of the code. Tests are in [test_lock.py](https://github.com/python-poetry/poetry/blob/335f209c2dd96cb33e75279c4a52cb25b1966bc3/tests/console/commands/test_lock.py). There are also tests for other commands that already check output. You can use these tests as a guide.

Further, I think in case of success (if `poetry.lock` is consistent with `pyproject.toml`, exit code 0) there should be a message, too.

PR #5081, opened 2022-01-21
Add output for poetry lock --check

artemrys:
# Pull Request Check List

Resolves: #5038

### DIFF
--- a/src/poetry/console/commands/export.py
+++ b/src/poetry/console/commands/export.py
@@ -57,13 +57,12 @@ class ExportCommand(Command):
             self.call("lock", " ".join(options))
 
         if not locker.is_fresh():
-            self.line_error(
-                "<warning>"
-                "Warning: The lock file is not up to date with "
-                "the latest changes in pyproject.toml. "
-                "You may be getting outdated dependencies. "
-                "Run update to update them."
-                "</warning>"
+            self._io.write_line(
+                "<error>"
+                "Error: poetry.lock is not consistent with pyproject.toml. "
+                "You may be getting improper dependencies. "
+                "Run `poetry update` to fix it."
+                "</error>"
             )
 
         exporter = Exporter(self.poetry)
--- a/src/poetry/console/commands/export.py
+++ b/src/poetry/console/commands/export.py
@@ -57,9 +57,9 @@ class ExportCommand(Command):
             self.call("lock", " ".join(options))
 
         if not locker.is_fresh():
-            self._io.write_line(
+            self.line_error(
                 "<warning>"
-                "Error: poetry.lock is not consistent with pyproject.toml. "
+                "Warning: poetry.lock is not consistent with pyproject.toml. "
                 "You may be getting improper dependencies. "
                 "Run `poetry update` to fix it."
                 "</warning>"
--- a/src/poetry/console/commands/export.py
+++ b/src/poetry/console/commands/export.py
@@ -61,7 +61,7 @@ class ExportCommand(Command):
                 "<warning>"
                 "Warning: poetry.lock is not consistent with pyproject.toml. "
                 "You may be getting improper dependencies. "
-                "Run `poetry update` to fix it."
+                "Run `poetry lock [--no-update]` to fix it."
                 "</warning>"
             )
 
--- a/src/poetry/console/commands/export.py
+++ b/src/poetry/console/commands/export.py
@@ -58,11 +58,11 @@ class ExportCommand(Command):
 
         if not locker.is_fresh():
             self._io.write_line(
-                "<error>"
+                "<warning>"
                 "Error: poetry.lock is not consistent with pyproject.toml. "
                 "You may be getting improper dependencies. "
                 "Run `poetry update` to fix it."
-                "</error>"
+                "</warning>"
             )
 
         exporter = Exporter(self.poetry)
--- a/src/poetry/console/commands/lock.py
+++ b/src/poetry/console/commands/lock.py
@@ -38,7 +38,7 @@ file.
 
         if self.option("check"):
             if self.poetry.locker.is_locked() and self.poetry.locker.is_fresh():
-                self.line("poetry.lock is consistent with pyproject.toml")
+                self.line("poetry.lock is consistent with pyproject.toml.")
                 return 0
             self.line(
                 "<error>"
--- a/src/poetry/console/commands/lock.py
+++ b/src/poetry/console/commands/lock.py
@@ -38,10 +38,13 @@ file.
 
         if self.option("check"):
             if self.poetry.locker.is_locked() and self.poetry.locker.is_fresh():
-                self.line("poetry.lock is up to date")
+                self.line("poetry.lock is consistent with pyproject.toml")
                 return 0
             self.line(
-                "Error: poetry.lock is out of date. Run `poetry update` to fix it"
+                "<error>"
+                "Error: poetry.lock is not consistent with pyproject.toml. "
+                "Run `poetry lock [--no-update]` to fix it."
+                "</error>"
             )
             return 1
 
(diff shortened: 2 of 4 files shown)


radoering:
I don't like the wording _"up to date"_ resp. _"out of date"_ because it is kind of ambiguous. The command does not check if some dependencies in the lock file are not up to date (as long as they comply with the constraints from `pyproject.toml`). The command only checks if `poetry.lock` is consistent with `pyproject.toml` (see [documentation](https://python-poetry.org/docs/master/cli/#options-9)). Maybe, we should stick with the wording from the documentation?

Further, I would say the minimum fix is running `poetry lock --no-update` (only making `poetry.lock` consistent with `pyproject.toml` not updating any dependencies already locked). Thus, I would propose to run `poetry lock [--no-update]` (square brackets for not required but sensible option).

I am aware that the warning when running `poetry install` says: `Warning: The lock file is not up to date with the latest changes in pyproject.toml. You may be getting outdated dependencies. Run update to update them.` I think this message is ok, because it defines _up to date_ more precisely by adding _with the latest changes in pyproject.toml_. However, I think `poetry lock [--no-update]` would be a better recommendation in this case, too.

All in all, I think the wording should be harmonized. Either **consistent with `pyproject.toml`** or **up to date with the latest changes in `pyproject.toml`**. Further, I would propose to recommend running `poetry lock [--no-update]` instead of `poetry update`.

But that is only my opinion. Any strong opinions from @python-poetry/triage ?

src/poetry/console/commands/lock.py
```diff
@@ -37,11 +37,13 @@ def handle(self) -> int:
...
         if self.option("check"):
-            return (
-                0
-                if self.poetry.locker.is_locked() and self.poetry.locker.is_fresh()
-                else 1
+            if self.poetry.locker.is_locked() and self.poetry.locker.is_fresh():
+                self.line("poetry.lock is up to date")
+                return 0
+            self.line(
+                "Error: poetry.lock is out of date. Run `poetry update` to fix it"
```
radoering, line 44: You should use `<error>...</error>` as in https://github.com/python-poetry/poetry/blob/28d05d002475f0cc55e7cc46ae01f9a3ce4e03a8/src/poetry/console/commands/check.py#L26
for coloring.

radoering (changes requested):

src/poetry/console/commands/lock.py
artemrys: Should we wait on an opinion from traiate team on this one?

radoering: I think adding `<error>...</error>` is safe.

Considering the wording: Since there are no replies yet, either nobody had time to think about it or has no strong opinion or is just fine with the proposal. 😉 If you want to speed up the process (and are not afraid from some more iterations of changes) we can push this PR until I will approve it. Before merging, a core member will look at the changes anyway and may (or may not) request further changes.

Thus, if you want to proceed, I propose to stick with the wording from the documentation (`poetry.lock` is (not) consistent with `pyproject.toml`) and recommend running `poetry lock [--no-update]`. Otherwise, you can also just wait some more days for additional opinions.

finswimmer: > I propose to stick with the wording from the documentation (poetry.lock is (not) consistent with pyproject.toml) and recommend running poetry lock [--no-update]


I like this one :+1:

artemrys: Updated messages :)

radoering:
Can you adapt the warnings in https://github.com/python-poetry/poetry/blob/08ed8ed14f8b4d079c5925156fbaa362b48e8cd6/src/poetry/installation/installer.py#L261-L264 and https://github.com/python-poetry/poetry/blob/08ed8ed14f8b4d079c5925156fbaa362b48e8cd6/src/poetry/console/commands/export.py#L62-L65 to the new wording, please?

I'd say we replace the first and the last sentence with the sentences from the error message and replace "outdated" in the second sentence by "improper" (or similar).

src/poetry/console/commands/lock.py
```diff
@@ -37,11 +37,16 @@ def handle(self) -> int:
         )
 
         if self.option("check"):
-            return (
-                0
-                if self.poetry.locker.is_locked() and self.poetry.locker.is_fresh()
-                else 1
+            if self.poetry.locker.is_locked() and self.poetry.locker.is_fresh():
+                self.line("poetry.lock is consistent with pyproject.toml")
```
radoering, line 41: I would add a punctuation mark at the end of the message.

artemrys: Done

artemrys:
I've updated `installer`'s and `export`'s messages to use new wording. I can add tests for that code in the next PR.

radoering:
Sorry, for nitpicking again, but hopefully the last request for change from me. 😉 I didn't meant to change the warnings in `installer` and `export` to errors but only to change the wording. Due to the fact that the commands are not aborted, a warning seems to be more suitable than an error for these commands.

Just to be clear: The error for `poetry check --lock` is suitable.

src/poetry/console/commands/export.py
```diff
@@ -57,12 +57,11 @@ def handle(self) -> None:
             self.call("lock", " ".join(options))
 
         if not locker.is_fresh():
-            self.line_error(
+            self._io.write_line(
```
radoering, line 60: Probably, `line_error()` is correct here, because warnings should go to stderr instead of stdout. I know, in `installer.py` the warning is written to stdout... Maybe, something that can be harmonized for all warnings in a separate PR...

src/poetry/console/commands/export.py
```diff
@@ -57,12 +57,11 @@ def handle(self) -> None:
...
                 "<warning>"
-                "Warning: The lock file is not up to date with "
-                "the latest changes in pyproject.toml. "
-                "You may be getting outdated dependencies. "
-                "Run update to update them."
+                "Error: poetry.lock is not consistent with pyproject.toml. "
```
radoering, line 62: This should start with `"Warning: "`

src/poetry/installation/installer.py
```diff
@@ -258,10 +258,9 @@ def _do_install(self, local_repo: Repository) -> int:
             if not self._locker.is_fresh():
                 self._io.write_line(
                     "<warning>"
-                    "Warning: The lock file is not up to date with "
-                    "the latest changes in pyproject.toml. "
-                    "You may be getting outdated dependencies. "
-                    "Run update to update them."
+                    "Error: poetry.lock is not consistent with pyproject.toml. "
```
radoering, line 261: This should also start with `"Warning: "`

src/poetry/console/commands/export.py
artemrys: Updated

src/poetry/console/commands/export.py
artemrys: Updated

src/poetry/installation/installer.py
artemrys: Updated

radoering (commented):
Sorry, I missed this before.

src/poetry/console/commands/export.py
```diff
@@ -59,10 +59,9 @@ def handle(self) -> None:
         if not locker.is_fresh():
             self.line_error(
                 "<warning>"
-                "Warning: The lock file is not up to date with "
-                "the latest changes in pyproject.toml. "
-                "You may be getting outdated dependencies. "
-                "Run update to update them."
+                "Warning: poetry.lock is not consistent with pyproject.toml. "
+                "You may be getting improper dependencies. "
+                "Run `poetry update` to fix it."
```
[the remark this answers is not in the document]
radoering: Same in `installer.py`.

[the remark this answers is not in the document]
artemrys: Done

radoering (approved):

radoering:
Finally looks good to me.

@finswimmer : Can you approve running the workflows and consider merging if the tests pass?

finswimmer (approved):

PR #5055, opened 2022-01-18
pyproject: Drop python 3.6

tony:
Resolves: https://github.com/python-poetry/poetry/issues/4997

# Pull Request Check List

- [ ] Requires https://github.com/python-poetry/poetry-core/pull/263
- [ ] See also: https://github.com/python-poetry/install.python-poetry.org/pull/11

Resolves: #4997

- [ ] Added **tests** for changed code.
- [x] Updated **documentation** for changed code.

Issue #4997, opened 2022-01-06
Deprecate Python 3.6


tony:
Python 3.6 [has reached end of life](https://devguide.python.org/devcycle/#end-of-life-branches) last month on December 23rd, 2021.

- [x] I have searched the [issues](https://github.com/python-poetry/poetry/issues) of this repo and believe that this is not a duplicate.
- [x] I have searched the [documentation](https://python-poetry.org/docs/) and believe that my question is not covered.

## Issue

Since Python 3.6 support has officially ended, 3.6 compatibility can be removed throughout the codebase (including CI, docs)

finswimmer:
Hello @tony,

we've discussed this already within the Poetry team and everyone likes to drop python3.6. According to [pypistats](https://pypistats.org/packages/poetry) there seems to be ~5% that are still using python3.6. This is why we haven't made a final decision until now.

fin swimmer

tony:
@finswimmer Okay if I make a PR for this?

tony:
@finswimmer Should we keep this issue open for https://github.com/python-poetry/poetry-core/pull/263 and https://github.com/python-poetry/install.python-poetry.org/pull/11?

poetry-core's issue tracker redirects back to poetry's: https://github.com/python-poetry/poetry-core/issues/new/choose

markus-oberhumer:
Chiming in late, but are there any *technical* reasons for not supporting Python 3.6 any longer?

Red Hat Enterprise Linux 8 (RHEL8) and all it's clones (like Alma Linux 8, Rocky Linux 8) use Python 3.6 and will be supported until May 2029 (and therefore will get fixes / backports from Red Hat), and as this is a really widespread Linux distribution this change does affect millions of servers.

tony:
@markus-oberhumer I think the next step would be to raise this concern in a new issue (referencing we had here)

finswimmer:
Hello @markus-oberhumer,

we are simply not enough people to support dead python versions. It's time consuming to work around things that doesn't work in older python versions. Furthermore it holds us back to improve poetry if we cannot use new python features.

We didn't write it anywhere until now, but between the maintainers there is an agreement, that we will promise to support a python version until it's end of life. After that we can drop the support at any time.

fin swimmer

(Personal opinion: The date when the end of life of a software is reached, doesn't mean "start transition", it means "end of the transition period")

markus-oberhumer:
[...]

PR #5088, opened 2022-01-22
chore: enable pep 563/585

branchvincent:
This enables [PEP 563](https://www.python.org/dev/peps/pep-0563/) and [PEP 585](https://www.python.org/dev/peps/pep-0585/) :fire:

Blocked by #5055
Relates-To: #4997

### DIFF
--- a/.flake8
+++ b/.flake8
@@ -8,7 +8,7 @@ inline-quotes = double
 # Allow omission of a return type hint for __init__ if at least one argument is annotated
 # used by flake8-annotations
 mypy-init-return = true
-enable-extensions = TC, TC2
+enable-extensions = TC, TC1
 type-checking-exempt-modules = typing, typing-extensions
 eradicate-whitelist-extend = ^-.*;
 extend-ignore =
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -50,7 +50,7 @@ repos:
     hooks:
       - id: pyupgrade
         args: [--py37-plus]
-        exclude: '^(install|get)-poetry.py$'
+        exclude: ^(install|get)-poetry.py$
 
   - repo: https://github.com/pycqa/isort
     rev: 5.10.1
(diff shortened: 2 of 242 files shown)


neersighted:
I think this might be a bit too aggressive -- is the new union syntax really available? I see all the tests pass, but according to [PEP 604](https://www.python.org/dev/peps/pep-0604/) this wasn't added til 3.10. My memory is a bit rusty however.

branchvincent:
Yea, you're right technically `x: int | str` is a syntax error for <3.10. However, with `from __future__ import annotations` this becomes equivalent to `x: "int | str"`. So by default it's not evaluated at run-time, but [mypy knows how to interpret it](https://mypy.readthedocs.io/en/stable/runtime_troubles.html#using-x-y-syntax-for-unions).

I should have noted that `pyupgrade` [handled this migration](https://github.com/asottile/pyupgrade#pep-585-typing-rewrites), the only other change here was removing unused imports (for which I used [pycln](https://github.com/hadialqattan/pycln)).

neersighted:
A second PR adding `pycln` to the pre-commit config would be welcome, I think.

Consensus seems to be that the new Union syntax is harmless and nicer to write. Let's go ahead with this.

neersighted (approved):
Approved, but I'd like to wait for at least one other maintainer to eyeball this before we merge.

danieleades:
I don't hate this. I've got a *feeling* that the `pyupgrade` hook can also add the `from __future__ import annotations` line, but perhaps only on later targets (maybe 3.9?). Not sure if anyone can confirm or deny that.

branchvincent:
> `pyupgrade` hook can also add the `from __future__ import annotations` line

I checked, it can definitely [remove them](https://github.com/asottile/pyupgrade#__future__-import-removal), but I can't find any indication of it adding them. See https://github.com/asottile/pyupgrade/issues/255#issuecomment-753509147 for an explanation of why it doesn't support that

branchvincent:
rebased due to conflicts

neersighted (approved):

Issue #5107, opened 2022-01-26
poetry stores repo credentials by host, it should include path


kapilt:
if you have multiple pypi repositories off the same domain but with different paths, poetry will fail to authenticate properly because it will store the credentials for the repository based on the domain, but not inclusive of path, meaning it will use the wrong credentials for a given domain.

- [ x ] I am on the [latest](https://github.com/python-poetry/poetry/releases/latest) Poetry version.
- [ x ] I have searched the [issues](https://github.com/python-poetry/poetry/issues) of this repo and believe that this is not a duplicate.
- [ x ] If an exception occurs when executing a command, I executed it again in debug mode (`-vvv` option).

```
$ poetry config --list | grep repo

repositories.assetdb.url = "https://corp-1122334455.d.codeartifact.us-east-1.amazonaws.com/pypi/assetdb/"
repositories.c7n-next.url = "https://corp-1122334455.d.codeartifact.us-east-1.amazonaws.com/pypi/c7n/"
```

traceback

```
  • Installing c7n (0.1.2): Pending...
  • Installing c7n (0.1.2): Failed

  RepositoryError                                                                                         
                                                     
  401 Client Error: Unauthorized for url: https://corp-1122334455.d.codeartifact.us-east-1.amazonaws.com/pypi/stacklet.c7n/simple/c7n/
                                                     
  at ~/.local/share/pypoetry/venv/lib/python3.9/site-packages/poetry/repositories/legacy_repository.py:393 in _get
      389│             if response.status_code == 404:
      390│                 return                                                                         
      391│             response.raise_for_status()                                                                                                                                                                  
      392│         except requests.HTTPError as e:                                                        
    → 393│             raise RepositoryError(e)                                                           
      394│                                           
      395│         if response.status_code in (401, 403):
      396│             self._log(
      397│                 "Authorization error accessing {url}".format(url=response.url),
```

Versions

-  Linux/Ubuntu 20.04
-  Poetry 1.11

The underlying issue appears to be how credentials are retrieved, it simply grabs the credentials for the first repository that has a domain match
https://github.com/python-poetry/poetry/blob/master/src/poetry/utils/authenticator.py#L151

kapilt:
instead the url comparison should be made with netloc + path

kapilt:
actually its a slightly more complex, as the path has some variance due to the `simple` suffix
```
(Pdb) parsed_url.netloc + parsed_url.path
[...]

PR #5108, opened 2022-01-26
fix: support multiple repositories with the same domain and different creds

kapilt:
# Pull Request Check List

Resolves: #5107

Existing tests that got touched were technically broken per repo spec in
https://www.python.org/dev/peps/pep-0503/


tgolsson:
@kapilt Drive-by comment, not in a position to say yay or nay but in what way were the tests broken? 

Note this part of PEP503:
> There are no constraints on where the files must be hosted relative to the repository.

agni-sairent:
> Also not in any position to resolve this issue, but I agree that there was no need to change the repository URL in tests, according to linked PEP.

abn:
Can folks impacted by this issue please test out #5518?

PR #5113, opened 2022-01-27
chore(commands): remove most type errors

branchvincent:
# Pull Request Check List

Resolves: #issue-number-here

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -97,6 +97,7 @@ force-exclude = '''
 [tool.mypy]
 check_untyped_defs = true
 ignore_missing_imports = true
+show_error_codes = true
 warn_redundant_casts = true
 warn_unused_configs = true
 warn_unused_ignores = true
@@ -113,29 +114,7 @@ module = [
   'poetry.config.file_config_source',
   'poetry.console.application',
   'poetry.console.logging.formatters.builder_formatter',
-  'poetry.console.commands.add',
-  'poetry.console.commands.build',
-  'poetry.console.commands.cache.clear',
-  'poetry.console.commands.command',
-  'poetry.console.commands.config',
-  'poetry.console.commands.debug.resolve',
-  'poetry.console.commands.end',
-  'poetry.console.commands.env_command',
-  'poetry.console.commands.export',
   'poetry.console.commands.init',
-  'poetry.console.commands.installer_command',
-  'poetry.console.commands.install',
-  'poetry.console.commands.lock',
-  'poetry.console.commands.new',
-  'poetry.console.commands.plugin.add',
-  'poetry.console.commands.remove',
-  'poetry.console.commands.run',
-  'poetry.console.commands.self.update',
-  'poetry.console.commands.shell',
-  'poetry.console.commands.show',
-  'poetry.console.commands.source.add',
-  'poetry.console.commands.update',
-  'poetry.console.commands.version',
   'poetry.inspection.info',
   'poetry.installation.chef',
   'poetry.installation.chooser',
--- a/src/poetry/console/commands/add.py
+++ b/src/poetry/console/commands/add.py
@@ -2,6 +2,7 @@ import contextlib
 
 from typing import Dict
 from typing import List
+from typing import cast
 
 from cleo.helpers import argument
 from cleo.helpers import option
@@ -236,7 +237,7 @@ You can specify a package in the following forms:
         if self.option("lock"):
             self._installer.lock()
 
-        self._installer.whitelist([r["name"] for r in requirements])
+        self._installer.whitelist([cast(str, r["name"]) for r in requirements])
 
         status = self._installer.run()
 
--- a/src/poetry/console/commands/config.py
+++ b/src/poetry/console/commands/config.py
@@ -7,6 +7,7 @@ from typing import Dict
 from typing import List
 from typing import Optional
 from typing import Tuple
+from typing import Union
 from typing import cast
 
 from cleo.helpers import argument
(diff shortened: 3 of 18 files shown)


pyproject.toml
```diff
@@ -97,6 +97,7 @@ force-exclude = '''
 [tool.mypy]
 check_untyped_defs = true
 ignore_missing_imports = true
+show_error_codes = true
```
branchvincent, line 100: this [shows error codes](https://mypy.readthedocs.io/en/stable/error_codes.html#displaying-error-codes), which was useful for our flake8 lint that requires `type: ignore[error-code]`

src/poetry/console/commands/export.py
```diff
@@ -48,11 +48,11 @@ def handle(self) -> None:
             self.line_error("<comment>The lock file does not exist. Locking.</comment>")
             options = []
             if self.io.is_debug():
-                options.append(("-vvv", None))
+                options.append("-vvv")
```
branchvincent, line 51: Found a bug here 😄

```console
$ rm poetry.lock && poetry export -vv
...
  TypeError

  sequence item 0: expected str instance, tuple found
  at src/poetry/console/commands/export.py:57 in handle
       53│                 options.append(("-vv", None))
       54│             elif self.io.is_verbose():
       55│                 options.append(("-v", None))
       56│ 
    →  57│             self.call("lock", " ".join(options))
       58│ 
       59│         if not locker.is_fresh():
       60│             self.line_error(
       61│                 ""
```

src/poetry/console/commands/source/add.py
```diff
@@ -86,7 +86,7 @@ def handle(self) -> Optional[int]:
                 )
                 return 1
 
-            if source.name == name:
+            if new_source and source.name == name:
```
branchvincent, line 89: Without this check, then the below assignment `source = new_source` maybe have produced `source = None`

src/poetry/console/commands/config.py
```diff
@@ -158,20 +159,19 @@ def handle(self) -> Optional[int]:
...
+                if setting_key not in self.unique_config_values:
                     raise ValueError(f"There is no {setting_key} setting.")
 
                 value = config.get(setting_key)
 
                 if not isinstance(value, str):
-                    value = json.dumps(value)
-
-                self.line(value)
+                    self.line(json.dumps(value))
```
neersighted, line 168: This logic is a bit jumbled, and I feel this change makes it moderately worse. Can we express this idiom better?

src/poetry/console/commands/add.py
```diff
@@ -236,7 +237,7 @@ def handle(self) -> int:
         if self.option("lock"):
             self._installer.lock()
 
-        self._installer.whitelist([r["name"] for r in requirements])
+        self._installer.whitelist([cast(str, r["name"]) for r in requirements])
```
branchvincent, line 240: reached for a `cast` here as `requirements` is too loosely typed as `Dict[str, Union[str, List[str]]]`. planning to tighten that up after removing all ignored files

src/poetry/console/commands/plugin/add.py
```diff
@@ -112,9 +112,11 @@ def handle(self) -> int:
                 root_package = ProjectPackage(package.name, package.version)
                 for dependency in package.requires:
                     root_package.add_dependency(dependency)
-
                 break
 
+        if not root_package:
```
neersighted, line 117: This change is good, but seems to be unrelated. Can we break it out into a separate PR?

src/poetry/console/commands/remove.py
```diff
@@ -83,8 +83,8 @@ def handle(self) -> int:
         if "group" in poetry_content and not poetry_content["group"]:
             del poetry_content["group"]
 
-        removed = set(removed)
-        not_found = set(packages).difference(removed)
+        removed_ = set(removed)
```
neersighted, line 86: Not sure I like the underscore suffix here. Why not `removed_set` or something similar?

src/poetry/console/commands/shell.py
```diff
@@ -32,6 +33,9 @@ def handle(self) -> None:
 
         self.line(f"Spawning shell within <info>{self.env.path}</>")
 
+        if not isinstance(self.env, VirtualEnv):
```
neersighted, line 36: This should also be factored out into its own change.

src/poetry/console/commands/version.py
```diff
@@ -80,27 +80,27 @@ def increment_version(self, version: str, rule: str) -> "Version":
         from poetry.core.semver.version import Version
 
         try:
-            version = Version.parse(version)
+            version_ = Version.parse(version)
```
neersighted, line 83: Same as above, I'd rather call this `parsed` or similar.

neersighted (changes requested):

src/poetry/console/commands/add.py
neersighted: That's kind of what I expected, sounds good.

src/poetry/console/commands/version.py
branchvincent: almost did that originally

src/poetry/console/commands/plugin/add.py
branchvincent: yup, just added `type: ignore` comments instead

src/poetry/console/commands/config.py
branchvincent: reverted

neersighted (approved):

PR #5129, opened 2022-01-31
Prefer actual-expected to expected-actual in assertions

radoering:
Relates-to: python-poetry/poetry#4776

As far as I know, there is no general convention about the order (actual-expected vs. expected-actual) in pytest assertions. However, I think a consistent order is desirable in order to quickly grasp what is going wrong in a failing test.

In my opinion, there is at least one strong argument for actual-expected:
`SIM300` of `flake8-simplify` is about avoiding yoda conditions. Thus, `age == 42` is preferred to `42 == age`. **Avoiding yoda conditions often results in actual-expected order.**

Further, most examples in pytest documentation are in actual-expected order, e.g. https://docs.pytest.org/en/6.2.x/index.html. (The same applies to unittest: https://docs.python.org/3/library/unittest.html) Probably, actual-expected is the more natural or at least more common order.

I am not aware if there is a linter that can help to keep actual-expected order because sometimes the context is required in order to decide which is actual and which is expected. At least, if most of the tests use a consistent order it will be more likely that this order is used in new tests.

### DIFF
--- a/tests/console/commands/debug/test_resolve.py
+++ b/tests/console/commands/debug/test_resolve.py
@@ -43,7 +43,7 @@ msgpack-python 0.5.3
 cachy          0.2.0
 """
 
-    assert expected == tester.io.fetch_output()
+    assert tester.io.fetch_output() == expected
 
 
 def test_debug_resolve_tree_option_gives_the_dependency_tree(tester: "CommandTester"):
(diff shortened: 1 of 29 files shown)


danieleades (approved):

finswimmer (approved):
LGTM :+1:

PR #5135, opened 2022-02-01
fixes mypy issue on windows

bmarroquin:
# Pull Request Check List

Resolves: #issue-number-here

mypy on windows was complaining about non-existent attributes. 
image

It looks like mypy doesn't honor the Windows boolean that was being previously used. By performing a sys.platform check in the if statement, mypy properly handles the early exit on windows.


radoering:
I can reproduce the issue and confirm that the fix is working. Maybe, we should add a comment why the constant is not used?

radoering (approved):

finswimmer (approved):
Thanks a lot :pray:

Issue #5141, opened 2022-02-02
Incorrect `poetry export` output with dependency conditional on Python version


glennmatthews:
- [x] I am on the [latest](https://github.com/python-poetry/poetry/releases/latest) Poetry version.
- [x] I have searched the [issues](https://github.com/python-poetry/poetry/issues) of this repo and believe that this is not a duplicate.
- [x] If an exception occurs when executing a command, I executed it again in debug mode (`-vvv` option).

- **OS version and name**: MacOS 12.2
- **Poetry version**: 1.1.12

## Issue

Running `poetry export --format requirements.txt --without-hashes` with the below `pyproject.toml` produces an invalid requirements file, specifically:

```
click==7.1.2
click==8.0.3; python_version >= "3.7"
```

which will fail pip for any python version of 3.7 or later. Presumably the `click==7.1.2` dependency should include a `python_version < "3.7"`, but it does not.

Minimal pyproject.toml to reproduce this issue:

```toml
[tool.poetry]
name = "poetry-test"
version = "0.1.0"
description = ""
authors = ["Glenn Matthews <glenn.matthews@networktocode.com>"]

[tool.poetry.dependencies]
python = "^3.6"
celery = [
    {version = "~5.1.0", python = "<3.7"},
    {version = "~5.2.0", python = ">=3.7"},
]

[build-system]
requires = ["poetry-core>=1.0.0"]
build-backend = "poetry.core.masonry.api"
```

```
% poetry export --format requirements.txt --without-hashes
amqp==5.0.9
billiard==3.6.4.0
cached-property==1.5.2
celery==5.1.2; python_version < "3.7"
celery==5.2.3; python_version >= "3.7"
click-didyoumean==0.0.3
click-plugins==1.1.1
click-repl==0.2.0
click==7.1.2    <------------------<< seems incorrect
click==8.0.3; python_version >= "3.7"
colorama==0.4.4; python_version >= "3.7" and python_full_version < "3.0.0" and platform_system == "Windows" or platform_system == "Windows" and python_version >= "3.7" and python_full_version >= "3.5.0"
importlib-metadata==4.10.1
kombu==5.1.0; python_version >= "3.6" and python_version < "3.7"
kombu==5.2.3; python_version >= "3.7"
prompt-toolkit==3.0.3
pytz==2021.3
six==1.16.0
typing-extensions==4.0.1; python_version < "3.8" and python_version >= "3.7"
vine==5.0.0
wcwidth==0.2.5
[...]

PR #5154, opened 2022-02-05
group packages correctly when exporting

dimbleby:
# Pull Request Check List

Resolves: #issue-number-here

The attempt to `groupby` during exports can be foiled by cases where a dependency is present with more than one version: eg if we see `[foo 1, foo 2, foo 1]` then we end up with three groups rather than two.

This improves #5141, in particular the lines for click that we now get for the example given there are
```
click==7.1.2 ; python_version >= "3.6" and python_full_version < "3.0.0" and python_version < "3.7" or python_version >= "3.6" and python_version < "3.7" and python_full_version >= "3.5.0"
click==8.0.3 ; python_version >= "3.7"
```
which looks acceptable.

However, in the effort to simplify this to a testcase I ran into at least two bugs that I _don't_ know how to fix, and these are exhibited by the testcase...

---

## Bug 1

Here's what we get for `click`:

```
click==7.1.2 ; python_version >= "2.7" and python_full_version < "3.0.0" and python_version < "3.7" or python_full_version >= "3.5.0" and python_version < "3.7" or python_full_version >= "3.6.2" and python_full_version < "4.0.0" or python_version >= "2.7" and python_full_version < "3.0.0" or python_full_version >= "3.5.0"
click==8.0.3 ; python_version >= "3.6"
```
two of the 7.1.2 possibilities overlap the 8.0.3 marker, which is not OK.

The cause seems to be that when the export code comes to process the "top-level" entries for click-didyoumean and click-plugins it sees that these should be installed for a broad range of python versions, and decides that therefore so should their dependency click.

As best I can see the fundamental problem is that the lock file does not record which packages are actually top-level dependencies, and which are not.  I don't see how to fix this without recording the actual top-level requirements as taken from `pyproject.toml`, which would allow the export code to reach click-didyoumean and click-plugins only in a path where it is taking account of parent markers.

---

## Bug 2

The output markers for click-plugins and click-didyoumean (sort of) are _also_ faulty in the test.  
```
 click-didyoumean==0.0.3 ; python_version < "3.7"
 click-didyoumean==0.3.0 ; python_full_version >= "3.6.2" and python_full_version < "4.0.0"
 click-plugins==1.1.1 ; python_version < "3.7"
```
- click-plugins should be installed unconditionally 
- click-didyoumean 0.0.3 is right, but only accidentally
- click-didyoumean 0.3.0 is wrong, it should be >= "3.7" (per bug 1, it is getting the versions from the "top-level" which is wrong)

The new problem here is that the marker setter at https://github.com/python-poetry/poetry-core/blob/0113a5a172e57424dbf5b2031d8c77097c725b5e/src/poetry/core/packages/dependency.py#L161 does not update the python versions correctly:
- start with a dependency whose marker is "< 3.7"
- update to an Any marker
- that setter does not see `python_versions` in the Any marker, and therefore does not clear "python < 3.7" from the `_python_versions`

### DIFF
--- a/src/poetry/packages/locker.py
+++ b/src/poetry/packages/locker.py
@@ -342,7 +342,7 @@ class Locker:
                     requirement.marker
                 )
 
-        return sorted(nested_dependencies.values(), key=lambda x: x.name.lower())
+        return nested_dependencies.values()
 
     def get_project_dependency_packages(
         self,
--- a/src/poetry/utils/exporter.py
+++ b/src/poetry/utils/exporter.py
@@ -1,7 +1,9 @@
-import itertools
 import urllib.parse
 
+from collections import defaultdict
 from typing import TYPE_CHECKING
+from typing import Dict
+from typing import List
 from typing import Optional
 from typing import Sequence
 from typing import Union
@@ -15,6 +17,8 @@ if TYPE_CHECKING:
     from pathlib import Path
 
     from cleo.io.io import IO
+    from poetry.core.packages.dependency_package import DependencyPackage
+    from poetry.core.packages.package import Package
 
     from poetry.poetry import Poetry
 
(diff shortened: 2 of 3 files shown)


dimbleby:
The CI pipeline points out that bug 2 is new since poetry-core 1.1.0a6. I had been testing locally with master

dimbleby:
The natural way to approach bug 1, I think, would be to introduce an entry in poetry.lock for the top-level package, recording its dependencies.  (This is how Cargo.lock works in rust-world).

I don't have a sense of how breaking that would be...

dimbleby:
superseded by #5156

PR #5156, opened 2022-02-06
Refactor export

dimbleby:
# Pull Request Check List

Refactoring `poetry export` to deal with bugs encountered in the testcase written for #5141, in particular "bug 1" of #5154.

This has turned out to be a rather larger set of changes than I had intended or than I am completely happy about submitting in a single MR.  But the fact is that I kept encountering bugs, and there is no real improvement until they all are fixed.  

Outline of the changes made:

- when selecting a locked package in the walk, make sure that not only does its version satisfy the constraint on the requirement, but also that its markers are compatible with the markers on the requirement
- when selecting a locked package in the walk, avoid selecting two different versions of the package with overlapping markers
- remove oddities from `__walk_dependency_level()`
  - (specifically: an undesirable special case for the top-level dependencies, and [this](https://github.com/python-poetry/poetry/blob/31657e7e5c47da758c14645035e59eb7813ac2f8/src/poetry/packages/locker.py#L252) which I can't make sense of except as some sort of workaround for the other bugs)
  - in fact I have reworked this as a very plain iterative breadth-first search
- removed some unused parameters `pinned_versions` and `with_nested`

We also need a fix in poetry-core to include markers in the hash of a `Dependency`, I'll do that separately.  I expect that'll prevent the pipeline from passing on this MR.  Edit: I rearranged this MR so that it doesn't rely on further poetry-core changes, but it seems that the tests do still require a recent version of that project.

The proof of all this pudding is in the testcase at the end of it all `test_exporter_doesnt_confuse_repeated_packages`, where we finally manage to export a set of requirements that is not impossible, doesn't have overlaps, and generally looks good.

Again, apologies for the size of the MR.  I actually think that overall this is a simplification, but there was rather a lot of code to simplify.  Do ask if you have any questions.

### DIFF
--- a/src/poetry/packages/locker.py
+++ b/src/poetry/packages/locker.py
@@ -32,6 +32,7 @@ from poetry.utils.extras import get_extra_package_names
 
 
 if TYPE_CHECKING:
+    from poetry.core.semver.version_constraint import VersionConstraint
     from poetry.core.version.markers import BaseMarker
     from tomlkit.items import InlineTable
     from tomlkit.toml_document import TOMLDocument
@@ -315,9 +316,19 @@ class Locker:
     def get_project_dependency_packages(
         self,
         project_requires: list[Dependency],
+        project_python_marker: VersionConstraint | None = None,
         dev: bool = False,
         extras: bool | Sequence[str] | None = None,
     ) -> Iterator[DependencyPackage]:
+        # Apply the project python marker to all requirements.
+        if project_python_marker is not None:
+            marked_requires: list[Dependency] = []
+            for require in project_requires:
+                require = deepcopy(require)
+                require.marker = require.marker.intersect(project_python_marker)
+                marked_requires.append(require)
+            project_requires = marked_requires
+
         repository = self.locked_repository(with_dev_reqs=dev)
 
         # Build a set of all packages required by our selected extras
--- a/src/poetry/packages/locker.py
+++ b/src/poetry/packages/locker.py
@@ -32,6 +32,7 @@ from poetry.utils.extras import get_extra_package_names
 
 
 if TYPE_CHECKING:
+    from poetry.core.version.markers import BaseMarker
     from tomlkit.items import InlineTable
     from tomlkit.toml_document import TOMLDocument
 
@@ -203,69 +204,80 @@ class Locker:
 
     @staticmethod
     def __get_locked_package(
-        _dependency: Dependency, packages_by_name: dict[str, list[Package]]
+        dependency: Dependency,
+        packages_by_name: dict[str, list[Package]],
+        decided: dict[Package, Dependency] | None = None,
     ) -> Package | None:
         """
         Internal helper to identify corresponding locked package using dependency
         version constraints.
         """
-        for _package in packages_by_name.get(_dependency.name, []):
-            if _dependency.constraint.allows(_package.version):
-                return _package
-        return None
+        decided = decided or {}
+
+        # Get the packages that are consistent with this dependency.
+        packages = [
+            package
+            for package in packages_by_name.get(dependency.name, [])
+            if package.python_constraint.allows_all(dependency.python_constraint)
+            and dependency.constraint.allows(package.version)
+        ]
+
+        # If we've previously made a choice that is compatible with the current
+        # requirement, stick with it.
+        for package in packages:
+            old_decision = decided.get(package)
+            if (
+                old_decision is not None
+                and not old_decision.marker.intersect(dependency.marker).is_empty()
+            ):
+                return package
+
+        return next(iter(packages), None)
 
     @classmethod
-    def __walk_dependency_level(
+    def __walk_dependencies(
         cls,
         dependencies: list[Dependency],
-        level: int,
-        pinned_versions: bool,
         packages_by_name: dict[str, list[Package]],
-        project_level_dependencies: set[str],
-        nested_dependencies: dict[tuple[str, str], Dependency],
-    ) -> dict[tuple[str, str], Dependency]:
-        if not dependencies:
-            return nested_dependencies
+    ) -> dict[Package, Dependency]:
+        nested_dependencies: dict[Package, Dependency] = {}
 
-        next_level_dependencies = []
-
-        for requirement in dependencies:
-            key = (requirement.name, requirement.pretty_constraint)
-            locked_package = cls.__get_locked_package(requirement, packages_by_name)
-
-            if locked_package:
-                # create dependency from locked package to retain dependency metadata
-                # if this is not done, we can end-up with incorrect nested dependencies
-                constraint = requirement.constraint
-                pretty_constraint = requirement.pretty_constraint
-                marker = requirement.marker
-                requirement = locked_package.to_dependency()
-                requirement.marker = requirement.marker.intersect(marker)
-
-                key = (requirement.name, pretty_constraint)
-
-                if not pinned_versions:
-                    requirement.set_constraint(constraint)
-
-                for require in locked_package.requires:
-                    if require.marker.is_empty():
-                        require.marker = requirement.marker
-                    else:
-                        require.marker = require.marker.intersect(requirement.marker)
-
-                    require.marker = require.marker.intersect(locked_package.marker)
-
-                    if key not in nested_dependencies:
-                        next_level_dependencies.append(require)
-
-            if requirement.name in project_level_dependencies and level == 0:
-                # project level dependencies take precedence
+        visited: set[tuple[Dependency, BaseMarker]] = set()
+        while dependencies:
+            requirement = dependencies.pop(0)
+            if (requirement, requirement.marker) in visited:
                 continue
+            visited.add((requirement, requirement.marker))
+
+            locked_package = cls.__get_locked_package(
+                requirement, packages_by_name, nested_dependencies
+            )
 
             if not locked_package:
-                # we make a copy to avoid any side-effects
-                requirement = deepcopy(requirement)
+                raise RuntimeError(f"Dependency walk failed at {requirement}")
 
+            # create dependency from locked package to retain dependency metadata
+            # if this is not done, we can end-up with incorrect nested dependencies
+            constraint = requirement.constraint
+            marker = requirement.marker
+            extras = requirement.extras
+            requirement = locked_package.to_dependency()
+            requirement.marker = requirement.marker.intersect(marker)
+
+            requirement.set_constraint(constraint)
+
+            for require in locked_package.requires:
+                if require.in_extras and extras.isdisjoint(require.in_extras):
+                    continue
+
+                require = deepcopy(require)
+                require.marker = require.marker.intersect(
+                    requirement.marker.without_extras()
+                )
+                if not require.marker.is_empty():
+                    dependencies.append(require)
+
+            key = locked_package
             if key not in nested_dependencies:
                 nested_dependencies[key] = requirement
             else:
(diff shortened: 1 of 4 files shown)


dimbleby:
Ooh, poetry already has a root package (just not derived from the lock file).  I should be able to undo most of the changes here, just keeping the core algorithm improvements.

finswimmer:
Hey @dimbleby,

please note that the export command will be replaced by the export-plugin. See https://github.com/python-poetry/poetry/pull/4824/. The export-plugin is located at https://github.com/python-poetry/poetry-export-plugin.

fin swimmer

dimbleby:
I've pushed a version that uses the existing root package, which removes large amounts of uninteresting changes in the unit tests.

The remaining impact on existing unit tests is that `poetry export` now includes the project-level python version in the markers - which I am happy is sensible, and correct.

Am not overly concerned about the plugin, most of the changes here are in the locker anyway.

Still seems to need recent poetry-core, would it be good to tag and release a new version of that project soon?

src/poetry/packages/locker.py
```diff
@@ -204,102 +205,104 @@ def locked_repository(self, with_dev_reqs: bool = False) -> "Repository":
...
-                        require.marker = requirement.marker
-                    else:
-                        require.marker = require.marker.intersect(requirement.marker)
+            locked_package = cls.__get_locked_package(
+                requirement, packages_by_name, nested_dependencies
+            )
 
-                    require.marker = require.marker.intersect(locked_package.marker)
+            if not locked_package:
+                # Should normally be able to satisfy all requirements, but this case is
```
dimbleby, line 258: I think this really ought not to happen, or if it does it indicates that something has gone wrong.  

The reason that I have left this check here rather than raising some sort of exception is the non-dev version of [this test](https://github.com/python-poetry/poetry/blob/31657e7e5c47da758c14645035e59eb7813ac2f8/tests/utils/test_exporter.py#L1527) in which:

- we find `bar`, which has the `baz` extra set
- but when we encounter `baz` we find that it is dev-only

I wonder whether it is even possible for a lockfile to be constructed this way; or anyway whether it doesn't show that something has gone wrong already.  Seems like `baz` should be a non-dev dependency, if there's a non-dev path to wanting it.

So perhaps we should remove that testcase, and be more assertive here that we expect to find a `locked_package` meeting whatever requirement we have reached.

dimbleby: tweaked the testcase so that `bar` is a dev dependency, which I think is likely what was intended

dimbleby:
rebased for:
- merge conflicts with typing changes
- to pick up poetry-core 1.1.0a7, which allows the new testcase to pass

src/poetry/packages/locker.py
dimbleby: changed my mind about this after encountering a real life example, reinstated the original testcase and made a fix to the code to pay more attention to what extras we want

src/poetry/packages/locker.py
```diff
@@ -203,145 +204,111 @@ def locked_repository(self, with_dev_reqs: bool = False) -> Repository:
...
-        )
+        return nested_dependencies
 
     @classmethod
     def get_project_dependencies(
         cls,
         project_requires: list[Dependency],
         locked_packages: list[Package],
-        pinned_versions: bool = False,
-        with_nested: bool = False,
```
abn, line 291: Is there a reason why we are droping this functionality? Export should not, in all cases provide locked versions nor nested dependencies. With this change, only the `pip freeze` use case can be addressed.

src/poetry/utils/exporter.py
```diff
@@ -70,21 +70,23 @@ def _export_requirements_txt(
...
         dependency_lines = set()
 
-        for package, groups in itertools.groupby(
-            self._poetry.locker.get_project_dependency_packages(
-                project_requires=self._poetry.package.all_requires,
-                dev=dev,
-                extras=extras,
-            ),
-            lambda dependency_package: dependency_package.package,
+        # Get project dependencies, and add the project-wide marker to them.
```
abn, line 73: Why should the project level marker be required here? Isin't this causing more noise without any realy functional benifit as the python compatibility refers to the project and not it's dependencies.

abn (changes requested):

src/poetry/utils/exporter.py
dimbleby: this is necessary to get the right answers!  we are now more careful in the main walk about selecting packages that are compatible with python markers - https://github.com/dimbleby/poetry/blob/2d61c3475d8746d6ff1d6f3e1cdfb0070815b468/src/poetry/packages/locker.py#L221 - so we need to pass in a suitably restricted initial marker or we might not find any solution at all.

src/poetry/packages/locker.py
dimbleby: it was unused function, no?  I removed it because I found no callers using it.

dimbleby:
I added another testcase showcasing a fix made in this MR relating to extras.  On master, `test_exporter_handles_extras_next_to_non_extras` results in:

```
another-thing==1.0.0
localstack-ext==1.0.0
localstack==1.0.0
something-else==1.0.0
something==1.0.0
```
but `another-thing` is wrong, that should only be installed if `localstack-ext[baz]` is installed, which it isn't.

src/poetry/packages/locker.py
abn: There is an intention to move Locker logic into core to support use cases where you would need pinned versions in wheels. This can be either for direct dependencies or nested. That was the rationale why this was introduced. I would suggest we keep this ability in so we can move it over easier later.

src/poetry/utils/exporter.py
abn: Since the exporter should rely on locked versions only, I do not see why we won't find a solution as the locker is aware of the project constraints. The reason why I am not a big fan of this is because the exports now look like this.

```
foo==1.0.0 ; python_version >= "2.7" and python_version < "2.8" or python_version >= "3.4" and python_version < "4.0"
```

While I agree that this is more exact, this is not really ideal. I can see the community not being happe about it as we already have had issues pile up when markers became a bit verbose in earlier iterations.

src/poetry/packages/locker.py
dimbleby: This function is currently both unused and untested.  And to be honest I'm not sure exactly what those parameters even mean.

I'd much prefer to get the case that we have working; we can always add function - with testcases! - later if and when it is wanted or needed.

src/poetry/utils/exporter.py
dimbleby: I understand.  But the line of code that I pointed at really is an essential correctness fix: and once you are committed to that you have to make sure that you start with a sufficiently narrow range of python versions, else that line will reject packages with python requirements that are too broad.

I'm open to other approaches if they can be found, I just don't currently see it.  I encourage you to experiment if you have other ideas in mind - the new `test_exporter_doesnt_confuse_repeated_packages` testcase work turned out to uncover lots of interesting bugs...

I suspect that most of us who export `requirements.txt` don't much care about the cosmetics any more than we care about the contents of `poetry.lock` - and anyway we're willing to put up with a little extra noise if that's the trade for getting the right answer.

src/poetry/packages/locker.py
abn: In general, irrespective of test coverage, I would be typically reluctant to change the signature of a public class method. Especially when refactoring, unlesss there is a pressing need. However, this is probably a special scenario as this is only used by `poetry` and `poetry-export-plugin` at present.

For some context this, I beleive, the parameters were introduced as as step change while working on the following:

- https://github.com/python-poetry/poetry-core/pull/83
- https://github.com/python-poetry/poetry-core/pull/71

See the example metadata in 71 to see the originally intended use cases.

For one reason or the other, progresson that halted but I do intend to pick it up again.

Additionally, export plugin will also eventualy need 2 kinds of exports, locked and constrainted exports.

In retrospect, you are correct in that more test coverage should be present. However, I would prefer coverage be added rather than functionality removed. That said, in this scenario I do think either is fine as long as the new changes do not somehow preclude the re-introduction of the functionality.

I'll go over the changes with a fresh pair of eyes today to see if applying this breaks something or makes things harder to change once the exoirt plugin gets decoupled.

Fwiw, apprecate the changes you have put into this, this is really useful for improving the maintainability, :pray:

src/poetry/utils/exporter.py
abn: Understandable. The way I might approach this is to pass `python_constraint=root_package.python_constraint` into `Locker.get_project_dependency_packages()` and then into the other methods. And do something like `package.python_constraint.allows_all(dependency.python_constraint.intersect(python_constraint))` in `Locker.__get_locked_package()`. 

This allows the flexibility to fetch dependencies given a python constraint and lets us keep the current behaviour.

We can discuss the change in behaviour in a separate PR and focus on the refactor and fixes here.

> I suspect that most of us who export requirements.txt don't much care about the cosmetics

Personally, I would agree. However, in certain cases readability has been indicated as a preference in some issues. I just want to decouple that discussion from the functional changes here.

dimbleby: That's do-able.  However it results in requirements.txt files that are ugly in a different way - and that are also inaccurate.

Consider a project whose python constraint is `^3.8`, and which happens to use some package A whose marker includes a broader constraint than this: say `^2.7 or ^3.6`.

Then under your proposal A and its children carry this maker through the search, and so that's what shows up in the exported `requirements.txt`.  

That's both ugly _and_ wrong!

In the MR as it stands, this broader marker gets merged with the project constraint, and so what is exported is just `>= 3.8 and <4`.

I acknowledge that the new output is noisier than it used to be: but I really would argue that it is now correct when it previously wasn't.  It seems completely reasonable to me that the exported requirements.txt should have the same constraints as the project being exported.

dimbleby: I'd add that as a result of recent work to improve simplifications of marker union and intersection, we are in general getting much more compact markers out than we used to.  

That doesn't change the question of whether including the project marker is in general desirable or not; but if the main worry is only cosmetic then it does greatly soften the blow.

dimbleby: actually for some combination of amusement and clarification... `poetry export` on poetry itself is kind of entertaining.

[this](https://gist.github.com/dimbleby/b5d2ae5a3d269a865a08bd9730ccc601) is what poetry 1.1.3 does.  And [this](https://gist.github.com/dimbleby/a0c465efe2d8e3c38dd779a459346918) is what this MR gets us.

After is both neater and more correct.  It's hard to understand what 1.1.3's:
```
zipp==3.7.0; python_version >= "3.7" and python_version < "3.8" and (python_version >= "3.6" and python_full_version < "3.0.0" and python_version < "3.8" or python_full_version >= "3.5.0" and python_version < "3.8" and python_version >= "3.6") or python_version < "3.8"
```
really means, but it certainly insists on `<"3.8"`, which is wrong.  (`importlib-metadata` is installed for all versions, so `zipp` must follow).

This is mostly just for fun, of course the master branch benefits from marker simplifications regardless of how we resolve this particular sub-question.

dimbleby: actually, for completeness, here's what we get following your idea of hiding the project python constraint during the search: https://gist.github.com/dimbleby/d0954dbce65411af1904f98957242060

it demonstrates very clearly the problem I noted - I don't think it's an improvement

abn: > That's both ugly and wrong!

I would argue that the changes proposed in this PR only makes the edge case look better, and makes the majority cases look unnecessarily noisy. However, it is not wrong. As per the lock file, the applied constraint is `^2.7 or ^3.6` for `A`. And hence it is dutifully dumping the marker as is, or atleast that is what I would expect.

> That doesn't change the question of whether including the project marker is in general desirable or not; but if the main worry is only cosmetic then it does greatly soften the blow.

At present, for me, there are the following open questions:
1. Cosmetic - this is not so important. Although I personally prefer the less noisy version as it is (a) easier to reason about than having all lines with markers view inspecting, (b) maps to real-world use cases of `requirements.txt`, (c) it is redundant for almost all use cases of the export.
2. Extensibility -  In the near future two changes are expected that are related to this. The first is de-coupling of the export command into its own plugin. And the second, is the use of the `Locker` for various cases like that of inspecting the lockfile (build, third-party tools etc).

Let's assume that we say we do not care about (1), I get the feeling you feel strongly about it :wink:. Let's keep the project constraint in as per your current changes. We can always deal with it later if there is push back from the community.

Hwoever, for (2), I would like to have the logic for "intersecting" root project python constraint be moved into `get_project_dependency_packages` (however that is done, passing the project or passing in the required python constraint). This is because otherwise we end up having to re-implement this logic for any use case that inspects dependencies from the lockfile in order to not fail.

dimbleby: > I would argue that the changes proposed in this PR only makes the edge case look better, and makes the majority cases look unnecessarily noisy. 

Per the example of poetry itself in https://github.com/python-poetry/poetry/pull/5156#discussion_r833719416: your "edge case" may actually be the majority case!

Anyway, I've added a commit which I think is what you're asking for, let me know if I've misunderstood.

abn: No, that looks about right. Thank you.

src/poetry/utils/exporter.py
```diff
@@ -70,21 +69,19 @@ def _export_requirements_txt(
...
-        for package, groups in itertools.groupby(
-            self._poetry.locker.get_project_dependency_packages(
-                project_requires=self._poetry.package.all_requires,
-                dev=dev,
-                extras=extras,
-            ),
-            lambda dependency_package: dependency_package.package,
+        # Get project dependencies.
+        groups = ["dev"] if dev else []
+        root_package = self._poetry.package.with_dependency_groups(groups)
```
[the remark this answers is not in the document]
dimbleby: sounds good

lovesegfault:
I've been hitting an annoying bug that is fixed by this, thanks for the fix @dimbleby :)

dimbleby:
ugh, merge conflicts.

@abn I don't think this is waiting on anything, can we merge it?

abn:
Ugh. My bad. I needed to do a once over and we would be good. I'll get to it today.

abn (approved):
Approving this for now. The logic is improved, current behaviour is largely maintained outside of project level python version marker propagation. This drops unused features but they can be reintroduced later as required.

Agalin:
Any chance for another beta release? It seems to finally fix our dependency issues mentioned in #3511 while beta1 is not compatible with the export plugin.

PR #5163, opened 2022-02-07
update pipelines

finswimmer:
* add python3.10 to release pipeline

* update actions/setup-python to @v2
* add python3.9 and python3.10 to checks
* update pytest for python >=3.6 otherwise it will fail on python3.10

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -22,7 +22,7 @@ jobs:
     strategy:
       matrix:
         os: [Ubuntu, MacOS, Windows]
-        python-version: [2.7, 3.6, 3.7, 3.8, 3.9, "3.10"]
+        python-version: [2.7, 3.5, 3.6, 3.7, 3.8, 3.9, "3.10"]
     steps:
       - uses: actions/checkout@v2
 
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -22,12 +22,12 @@ jobs:
     strategy:
       matrix:
         os: [Ubuntu, MacOS, Windows]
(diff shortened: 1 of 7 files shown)


make-nix-release.sh
```diff
@@ -14,6 +14,7 @@ if [ "$OSTYPE" == "linux-gnu" ]; then
   $PYTHON get-poetry.py -y
   POETRY="$PYTHON $HOME/.poetry/bin/poetry"
   RUNTIMES[5]="${PYTHON39:+-P "3.9:$PYTHON39"}"
+  RUNTIMES[5]="${PYTHON310:+-P "3.10:$PYTHON310"}"
```
abn, line 17:
```suggestion
  RUNTIMES[6]="${PYTHON310:+-P "3.10:$PYTHON310"}"
```

.github/workflows/main.yml
```diff
@@ -22,12 +22,12 @@ jobs:
     strategy:
       matrix:
         os: [Ubuntu, MacOS, Windows]
-        python-version: [2.7, 3.5, 3.6, 3.7, 3.8]
+        python-version: [2.7, 3.5, 3.6, 3.7, 3.8, 3.9, "3.10"]
```
abn, line 25: Just for consistency, I would recommend we quote all values now.

abn (changes requested):

make-nix-release.sh
finswimmer: :+1:

.github/workflows/main.yml
finswimmer: done

abn (approved):

Issue #1382, opened 2019-09-17
Build and install the root package as a wheel in one invocation


maggyero:
@sdispater, in https://github.com/sdispater/poetry/issues/730 you said:

> The `install` command installs the current project in editable mode by internally creating a temporary `setup.py` file and executing `pip install -e .`. It does it each time `install` is executed to ensure that any new elements like entrypoints are properly installed.

Could you allow [non-editable installations](https://pip.pypa.io/en/stable/reference/pip_install/#editable-installs) by adding a `--no-editable` _command-line parameter_ for these three commands:

- `poetry install --no-editable`
- `poetry add --no-editable`
- `poetry update --no-editable`

The [documentation](https://poetry.eustace.io/docs/cli/#add) states that it is already possible for local dependencies, but with a `develop` _file parameter_ in the `pyproject.toml` file:

> If you don't want the dependency to be installed in editable mode you can specify it in the pyproject.toml file:
>
> [tool.poetry.dependencies]
> my-package = {path = "../my/path", develop = false}

Here I think that a command-line parameter would be more appropriate than a file parameter, as it is something that you want to configure at the command-line level rather than at the project level (like with `pip` where sometimes you want to run `pip install .` and sometimes `pip install -e .`). So if you could also drop the `develop` file parameter that would be awesome.

Being able to install projects in non-editable mode is so important, especially for deployment on servers. Currently we cannot use the `poetry install` command exclusively for such installations, we still have to resort to the `pip install` command.

maggyero:
Up.

cjolowicz:
This is an important feature for multi-stage Docker builds. 

Multi-stage Docker builds for Python typically install the project into a virtualenv during the build stage, and then copy the virtualenv over into a slim final image. An editable install creates a .egg-link file that links to the source code, and this link would only be valid for the duration of the build stage.

Currently my build stage installs projects by exporting from Poetry to pip:

```Dockerfile
COPY pyproject.toml poetry.lock ./
RUN /root/.poetry/bin/poetry export -f requirements.txt | /venv/bin/pip install -r /dev/stdin 
COPY . ./
RUN /root/.poetry/bin/poetry build && /venv/bin/pip install dist/*.whl
```

The requested feature would allow us to use Poetry directly:

```Dockerfile
COPY pyproject.toml poetry.lock ./
RUN /root/.poetry/bin/poetry install --no-root
COPY . ./
RUN /root/.poetry/bin/poetry install --no-editable
```

[...]

PR #5174, opened 2022-02-09
Allow root package non-editable installation

rszamszur:
# Pull Request Check List

Partially Resolves: #1382 (Implemented `--no-editable` for install command)

If `--no-editable` flag is present, use `SdistBuilder` as a builder, build `setup.py`, then finally run `pip_install` utility method without deps and without editable mode.

I did some manual tests by trying to install back and forth with and without editable mode, which at first glance seems to work. However, after installing in editable mode again, the package directory (`.venv/lib/python3.x/site-packages/package_name`) seems still to exists but it's empty:
```shell
$ $ poetry run poetry install --no-editable
...
$ tree ../../.venv/lib/python3.9/site-packages/demo/
../../.venv/lib/python3.9/site-packages/demo/
├── cli
│   ├── cli.py
│   ├── commands
│   │   ├── __init__.py
│   │   └── __pycache__
│   │       └── __init__.cpython-39.pyc
│   ├── conftest.py
│   ├── __init__.py
│   └── __pycache__
│       ├── cli.cpython-39.pyc
│       ├── conftest.cpython-39.pyc
│       └── __init__.cpython-39.pyc
├── __init__.py
├── __main__.py
├── __pycache__
│   ├── __init__.cpython-39.pyc
│   ├── __main__.cpython-39.pyc
│   └── version.cpython-39.pyc
└── version.py
$ $ poetry run poetry install
...
$ tree ../../.venv/lib/python3.9/site-packages/demo/
../../.venv/lib/python3.9/site-packages/demo/
├── cli
│   ├── commands
│   │   └── __pycache__
│   └── __pycache__
└── __pycache__
```
It doesn't break the installation, but I just wanted to point that out. Lastly, the `*.pth` file is cleaned correctly.

### DIFF
--- a/docs/cli.md
+++ b/docs/cli.md
@@ -213,6 +213,12 @@ poetry install --no-root
 Installation of your project's package is also skipped when the `--only`
 option is used.
 
+If you do not want to install the current project in [editable](https://pip.pypa.io/en/stable/cli/pip_install/#install-editable) mode, run the `install` command with the `--no-editable` flag:
+
+```bash
+poetry install --no-editable
+```
+
 ### Options
 
 * `--without`: The dependency groups to ignore for installation.
@@ -223,6 +229,7 @@ option is used.
 * `--no-root`: Do not install the root package (your project).
 * `--dry-run`: Output the operations but do not execute anything (implicitly enables --verbose).
 * `--extras (-E)`: Features to install (multiple values allowed).
+* `--no-editable`: Do not install the root package in editable mode.
 * `--no-dev`: Do not install dev dependencies. (**Deprecated**)
 * `--dev-only`: Only install dev dependencies. (**Deprecated**)
 * `--remove-untracked`: Remove dependencies not presented in the lock file. (**Deprecated**)
--- a/src/poetry/console/commands/install.py
+++ b/src/poetry/console/commands/install.py
@@ -95,10 +95,10 @@ dependencies and not including the current project, run the command with the
     _loggers = ["poetry.repositories.pypi_repository", "poetry.inspection.info"]
 
     def handle(self) -> int:
+        from poetry.core.masonry.builders.sdist import SdistBuilder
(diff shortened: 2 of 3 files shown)


rszamszur:
Regarding the `poetry update` command. It'll probably need some logic to update the project if it was installed with `--no-editable`.
The first idea that comes to my mind is to detect that automatically. If `PACKAGE_NAME.pth` file does not exists under `VENV_PATH/lib/python3.x/site-packages/` directory then reinstall, else do nothing. What do you think?

Have a good one!

rszamszur:
Regarding the `poetry update` command. It'll probably need some logic to update the project if it was installed with `--no-editable`.
The first idea that comes to my mind is to detect that automatically. If `PACKAGE_NAME.pth` file does not exists under `VENV_PATH/lib/python3.x/site-packages/` directory then reinstall, else do nothing. What do you think?

Have a good one!

abn:
The generation of `setup.py` is a deprecated feature and is something that will be dropped in future releases. As mentioned above there will be a few different paths which will allow similar use cases. Bundle plugin being one of those options, outside of that there is also a possibility that core will be lock file aware (eg: python-poetry/poetry-core#83) allowing install of locked poetry projects from source via `pip` (or any PEP-517 front end) in a target environment (particularly useful in container builds).

Alternatively, I could see `poetry install [--distribution | --no-editable]` or similar enabled via a wheel build followed by a call to the installer instead of the editable install it does today.

Note that any such feature will need to handle the scenario where the current project becomes editable by virtue of the source being  in the current directory as this will get added to `sys.path`. This means, even if a project is installed without being editable, it may still behave as such.

neersighted:
Closing this for now because it has a design that is not compatible with the current direction of the rest of the project -- I would suggest checking out the resources linked above and opening an issue to discuss any shortcomings they may have if they don't suffice for your use-case.

PR #5167, opened 2022-02-08
Release 1.1.13

finswimmer:
## Fixed

- Fixed an issue where envs in MSYS2 always reported as broken ([#4942](9https://github.com/python-poetry/poetry/pull/4942))
- Fixed an issue where conda envs in windows are always reported as broken([#5008](https://github.com/python-poetry/poetry/pull/5008))
- Fixed an issue where Poetry doesn't find its own venv on `poetry self update` ([#5048](https://github.com/python-poetry/poetry/pull/5048))

PR #5176, opened 2022-02-10
ci: preview release builds on relevant changes

abn:
This change enables a preview build of release artefacts when relevant sources change. This is only triggerred on PRs as we do not want the artefacts for every branch.

This also includes a few dirty workarounds to add support for 3.10. See also #5163 and #5167. Additionally, broken windows builds are also fixed.


abn:
Please keep the commits separate and do not squash.

Issue #5177, opened 2022-02-10
Hide python switching warning message


Dragas:
- [x] I am on the [latest](https://github.com/python-poetry/poetry/releases/latest) Poetry version.
- [x] I have searched the [issues](https://github.com/python-poetry/poetry/issues) of this repo and believe that this is not a duplicate.
- [x] If an exception occurs when executing a command, I executed it again in debug mode (`-vvv` option).

- **OS version and name**: Ubuntu 20.04 LTS server
- **Poetry version**: 1.1.12 (Jesus christ figuring this out shouldn't be roundabout process)
- **Link of a [Gist](https://gist.github.com/) with the contents of your pyproject.toml file**: (https://gist.github.com/Dragas/10cc26ed3b38b60bdae8c5e995f30db5)

## Issue

I am going to tell you about weirdest process yet.

For reasons I must write development code on windows machine, but I can only run python (and in turn, poetry) on my development virtual machine via SSH. Intellij offers such functionality via "SSH Interpreter" where you project would be synchronized to remote machine and then invoked via SSH. This works fine (both debugging, and running) without poetry. 

In the project that we are developing we must use python 3.9 (as per the minimal reproducable sample), but ubuntu 20.04 server comes with 3.8 by default, and it seems that upgrading to 3.9 breaks some parts of the operating system. This is circumvented by installing 3.9 via apt and setting default version to 3.8 via `update-alternatives` tool. When you use non-required version to run poetry, poetry checks the operating system for installed python versions and switches to it, but emits a warning message. This breaks Intellij's debugging mode via SSH, because intellij runs the project twice (first with some debugging arguments to determine address and port, second time the actual project), and uses stdout from first invocation to pass address and port for remote debugging. Since poetry outputs a warning message that python was switched, the invoked commandline from 

```
ssh://python@192.168.80.129:22/home/python/.poetry/bin/poetry_wrapped.sh -u /home/python/.pycharm_helpers/pydev/pydevd.py --multiproc --qt-support=auto --client "0.0.0.0" --port 38929 --file /home/python/main.py
```
becomes
```
ssh://python@192.168.80.129:22/home/python/.poetry/bin/poetry_wrapped.sh -u /home/python/.pycharm_helpers/pydev/pydevd.py --multiproc --qt-support=auto --client "he currently activated Python version 3.8.10 is not supported by the project (^3.9).
Trying to find and use a compatible version. 
Using python3.9 (3.9.5)
[...]

PR #5179, opened 2022-02-10
Write errors and warnings to stderr (instead of stdout)

radoering:
# Pull Request Check List

Resolves: #5177

Currently, warnings are often (not always) written to stdout. Sometimes even errors are written to stdout.

Probably, errors and warnings should be written to stderr. See [this Stack Overflow post](https://stackoverflow.com/questions/1430956/should-i-output-warnings-to-stderr-or-stdout) for example.


finswimmer (approved):
LGTM :+1:

PR #5191, opened 2022-02-11
Require urllib3 ^1.26.0

davegaeddert:
I'm not actually sure how to make a test for this, but in debugging some other issues I found that [urllib3 needs to be at least version 1.26.0](
https://github.com/urllib3/urllib3/compare/1.25.11...1.26.0#diff-53a96dd57c993a39101318e9df47f8aab6610e8c2c1064f8109f66943dcbbd78R233) to use the `allowed_methods` parameter on the Retry class:

https://github.com/python-poetry/poetry/blob/7cc684981983963dc202e1a249a4b66667b468bd/src/poetry/publishing/uploader.py#L73

Or, it needs to continue using `method_whitelist` which is deprecated but not going to be removed until urllib3 v2.0... That would probably be safer in terms of restricting the downstream dependencies? Curious what you think. But maybe the restriction is ok considering people aren't screaming about `poetry publish` failing...

[This is where I first saw it.](https://github.com/dropseed/workhorse/runs/5159936470?check_suite_focus=true)
```sh
urllib3                1.25.8
...
   1  ~/.local/lib/python3.8/site-packages/poetry/publishing/uploader.py:106 in make_session
       104│         session.headers["User-Agent"] = self.user_agent
       105│         for scheme in ("http://", "https://"):
     → 106│             session.mount(scheme, self.adapter)
       107│ 
       108│         return session

  TypeError

  __init__() got an unexpected keyword argument 'allowed_methods'

  at ~/.local/lib/python3.8/site-packages/poetry/publishing/uploader.py:70 in adapter
       66│         return user_agent("poetry", __version__)
       67│ 
       68│     @property
       69│     def adapter(self) -> adapters.HTTPAdapter:
    →  70│         retry = util.Retry(
       71│             connect=5,
       72│             total=10,
       73│             allowed_methods=["GET"],
       74│             status_forcelist=[500, 501, 502, 503],
```

Probably introduced in #4580.

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -79,7 +79,7 @@ python-versions = ">=3.6.1"
 
 [[package]]
 name = "charset-normalizer"
-version = "2.0.10"
+version = "2.0.12"
 description = "The Real First Universal Charset Detector. Open, modern and actively maintained alternative to Chardet."
 category = "main"
 optional = false
@@ -110,11 +110,11 @@ python-versions = ">=2.7, !=3.0.*, !=3.1.*, !=3.2.*, !=3.3.*, !=3.4.*"
 
 [[package]]
 name = "coverage"
-version = "6.2"
+version = "6.3.2"
 description = "Code coverage measurement for Python"
 category = "dev"
 optional = false
-python-versions = ">=3.6"
+python-versions = ">=3.7"
 
 [package.extras]
 toml = ["tomli"]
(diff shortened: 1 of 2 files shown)


neersighted (approved):
Could you please rebase this to fix conflicts?

abn:
@davegaeddert isin't the `urllib3` instance in question vendored by `requests` and not explicitly a top level dependency of poetry?

abn (approved):

PR #5224, opened 2022-02-22
Backport PR #3337 - Fix handling errors 401, 403 when looking for repositories

### LINKED ISSUE #3303: Regression: 403 Auth Failure for S3-hosted (Cloudfront) Private Repos

1ace:
Backport of #3337 which fixes #3303 to the 1.1 branch, so that it can be part of the next 1.1.x release :slightly_smiling_face:

/cc @finswimmer

### DIFF
--- a/poetry/repositories/legacy_repository.py
+++ b/poetry/repositories/legacy_repository.py
@@ -386,18 +386,17 @@ class LegacyRepository(PyPiRepository):
         url = self._url + endpoint
         try:
             response = self.session.get(url)
+            if response.status_code in (401, 403):
+                self._log(
+                    "Authorization error accessing {url}".format(url=url),
+                    level="warning",
+                )
+                return
             if response.status_code == 404:
                 return
             response.raise_for_status()
         except requests.HTTPError as e:
             raise RepositoryError(e)
-
-        if response.status_code in (401, 403):
-            self._log(
-                "Authorization error accessing {url}".format(url=response.url),
-                level="warn",
-            )
-            return
 
         if response.url != url:
             self._log(
--- a/tests/repositories/test_legacy_repository.py
+++ b/tests/repositories/test_legacy_repository.py
@@ -320,19 +320,18 @@ def test_get_200_returns_page(http):
     assert repo._get("/foo")
 
 
-def test_get_404_returns_none(http):
-    repo = MockHttpRepository({"/foo": 404}, http)
+@pytest.mark.parametrize("status_code", [401, 403, 404])
+def test_get_40x_and_returns_none(http, status_code):
+    repo = MockHttpRepository({"/foo": status_code}, http)
 
     assert repo._get("/foo") is None
 
 
-def test_get_4xx_and_5xx_raises(http):
-    endpoints = {"/{}".format(code): code for code in {401, 403, 500}}
-    repo = MockHttpRepository(endpoints, http)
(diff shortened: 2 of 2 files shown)


Maciej-Zwolinski (approved):
Jesus, why didn't anybody fix this before or has approved this merge earlier?

As is, the code handling 401 and 403 exceptions is inaccessible, hence should either be removed or moved before exception is raised to be actually functional. I lean heavily towards handling 40x exceptions instead of raising them and causing entire build process go to hell because somebody either forgot to log in or (more likely) you are trying to find non-existent packages in a private repository that returns 403 instead of 404 (looking at you pytorch...).

Secrus:
Poetry team won't do another 1.1.x release, so this PR is no longer valid. Closing.

Maciej-Zwolinski:
Ok, thank you for clarification. I was not up to date with release schedule of poetry, since the 1.1.x is still the default version when looking for docs (https://python-poetry.org/docs/). Perhaps it'd be a good idea to point towards 1.2 as a recommended release (even if it's only in preview).

Issue #5240, opened 2022-02-27
poetry show gives clickit "NoSuchOptionException(name)" error


swills1:
- [x ] I am on the [latest](https://github.com/python-poetry/poetry/releases/latest) Poetry version.
- [x ] I have searched the [issues](https://github.com/python-poetry/poetry/issues) of this repo and believe that this is not a duplicate.
- [ x] If an exception occurs when executing a command, I executed it again in debug mode (`-vvv` option).

- **OS version and name**: Fedora 35
- **Poetry version**: 3.10.1
- **Link of a [Gist](https://gist.github.com/) with the contents of your pyproject.toml file**: N/A
## Issue
Hello.

I am getting this clickit error;

```
NoSuchOptionException

  The "--show" option does not exist.

  at ~/.local/lib/python3.10/site-packages/clikit/args/default_args_parser.py:300 in _add_long_option
      296│     def _add_long_option(
      297│         self, name, value, tokens, fmt, lenient
      298│     ):  # type: (str, Optional[str], List[str], ArgsFormat, bool) -> None
      299│         if not fmt.has_option(name):
    → 300│             raise NoSuchOptionException(name)
      301│ 
      302│         option = fmt.get_option(name);
      303│ 
      304│         if value is False:
```

These are the steps to reproduce

 ```
poetry new packagename
cd packagename
poetry shell
poetry --show
```

This isn't preventing work, I just wanted to report it. I did search issues, but I may have missed one talking about this.

**EDIT**

I was turned around on what I was doing - true issue is - https://github.com/python-poetry/poetry/issues/5240#issuecomment-1052926766

branchvincent:
I believe you're looking for the [`poetry show`](https://python-poetry.org/docs/cli/#show) command

swills1:
I apologize, I wasn't thinking when I posted.. I do get an error with poetry show, I completely typed up the wrong thing. So

- poetry new package
- cd package
- poetry shell
- poetry show

SolverProblemError

  Because test depends on pytest (^5.2) which doesn't match any versions, version solving failed.

  at ~/.local/lib/python3.10/site-packages/poetry/puzzle/solver.py:241 in _solve
      237│             packages = result.packages
      238│         except OverrideNeeded as e:
      239│             return self.solve_in_compatibility_mode(e.overrides, use_latest=use_latest)
      240│         except SolveFailure as e:
    → 241│             raise SolverProblemError(e)
      242│ 
      243│         results = dict(
[...]

PR #5242, opened 2022-02-27
fix(show): handle missing lockfile

branchvincent:
`poetry show` relies on the existence of `poetry.lock` to display dependencies, but it currently throws a `SolverProblemError` if this file doesn't exist yet. This checks if the lockfile is missing and errors, if needed.

# Pull Request Check List

Resolves: #5240

### DIFF
--- a/src/poetry/console/commands/show.py
+++ b/src/poetry/console/commands/show.py
@@ -130,6 +130,13 @@ lists all packages available."""
         if self.option("default"):
             only_groups.append("default")
 
+        if not self.poetry.locker.is_locked():
+            self.line_error(
+                "<error>Error: poetry.lock not found. Run `poetry lock` to create"
+                " it.</error>"
+            )
+            return 1
+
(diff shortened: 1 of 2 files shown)


src/poetry/console/commands/show.py
```diff
@@ -130,6 +130,10 @@ def handle(self) -> Optional[int]:
         if self.option("default"):
             only_groups.append("default")
 
+        if not self.poetry.locker.is_locked():
+            self.line_error("<comment>The lock file does not exist. Locking.</comment>")
+            self.call("lock")
```
finswimmer, line 135: IMO `poetry show` shouldn't have the side effect of locking. A message about the missing lock file and maybe an instruction on what to do now should be sufficient.

branchvincent: Yea I went back and forth on that as well, I'm fine either way 👍

finswimmer (approved):
LGTM :+1:

PR #5246, opened 2022-02-28
readme: do not repeat installation instructions

abn:


README.md
```diff
@@ -18,138 +18,9 @@ The [complete documentation](https://python-poetry.org/docs/) is available on th
...
-For `zsh`, you must then add the following line in your `~/.zshrc` before
-`compinit` (not for homebrew setup):
-
-```zsh
-fpath+=~/.zfunc
-```
-
+Instructions on how to install `poetry` can be found [here](https://python-poetry.org/docs/master/#installation).
+You can also refer [here](https://python-poetry.org/docs/master/#enable-tab-completion-for-bash-fish-or-zsh) for
+information on how to enable tab completion in your environment.
```
finswimmer, line 23: A lot of people seems to find it confusing, that the instructions in the README of the master branch doesn't reflect the latest stable version. Maybe we can make it more clear by linking to both (master, latest stable) documentations?

finswimmer (changes requested):

README.md
abn: Partially the reason why this PR was created. I was deliberitely avoiding having both listed for now as we are transitioning installation methods. Specifically, we would want avoid folks going to stable one by default. With this change the README will no longer explicitly contain this instruction anyway. And further, going forward we really want to avoid having version specific installation methods. With that in mind, having both listed here is less likely to help the issue.

branchvincent (approved):

PR #5248, opened 2022-02-28
remove `__future__` import from src/poetry/__init__.py

finswimmer:
Setting `__path__` in `src/poetry/__init__.py` must be the first line, otherwise IDEs like PyCharm cannot follow imports to `poetry.core`.


branchvincent (dismissed):
wonder if it's worth adding a comment in the file about it needing to be the first line

branchvincent (approved):

neersighted (approved):

Issue #5250, opened 2022-02-28
When keyring is disabled, poetry tries to use it anyways. 


lakinwecker:
- [x] I am on the [latest](https://github.com/python-poetry/poetry/releases/latest) Poetry version.
- [x] I have searched the [issues](https://github.com/python-poetry/poetry/issues) of this repo and believe that this is not a duplicate.
- [x] If an exception occurs when executing a command, I executed it again in debug mode (`-vvv` option).

- **Arch linux up to date**: 
- **1.1.13**: 
- **Link of a [Gist](https://gist.github.com/) with the contents of your pyproject.toml file**: [pyproject.toml](https://gist.github.com/lakinwecker/f45a753fea8c5650fe6367c254e6af15)

## Issue
If you disable keyring via it's recommended route: https://github.com/jaraco/keyring#disabling-keyring you end up with a "null" keyring being used. This is a useless keyring, but poetry continues to try and use it regardless.   poetry should treat it the same as if keyring was unavailable.

lakinwecker:
To be clear, when it fails this way, it fails to do things like it doesn't try and use the environment variable. It just silently ignores it.

tyler-8:
> To be clear, when it fails this way, it fails to do things like it doesn't try and use the environment variable. It just silently ignores it.

This is a key symptom. I've been very confused as to why my correctly set environment variables for username/password of a private repo are not working. It wasn't until I disabled the keyring (as apparently I'd configured poetry to use it in the past) that I realized that truly the envars were being bypassed. Instead I get prompted for the password when doing `poetry publish -r myrepo`.

```
# envars
POETRY_HTTP_BASIC_MYREPO_USERNAME=someuser
POETRY_HTTP_BASIC_MYREPO_PASSWORD=password

# poetry config --list

repositories.myrepo.url = "https://internalpypi.domain.com/pypi"
```

kadler:
I ran in to this same issue today. Looks like a similar check like this needs to be made for "null":

https://github.com/python-poetry/poetry/blob/master/src/poetry/utils/password_manager.py#L100-L102

kadler:
As a workaround, the "fail" backend can be used: `export PYTHON_KEYRING_BACKEND=keyring.backends.fail.Keyring`

lakinwecker:
> I ran in to this same issue today. Looks like a similar check like this needs to be made for "null":
> 
[...]

PR #5251, opened 2022-02-28
when "disabled", keyring uses a "null" keyring and should be unavaila…

lakinwecker:
…ble.

# Pull Request Check List

Resolves: #5250

I didn't see any tests or docs for this part of the code, happy to update them as required.

### DIFF
--- a/src/poetry/utils/password_manager.py
+++ b/src/poetry/utils/password_manager.py
@@ -94,7 +94,7 @@ class KeyRing:
 
         backend = keyring.get_keyring()
         name = backend.name.split(" ")[0]
-        if name == "fail":
+        if name in ("fail", "null"):
             logger.debug("No suitable keyring backend found")
             self._is_available = False
         elif "plaintext" in backend.name.lower():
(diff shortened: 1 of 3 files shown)


abn:
@lakinwecker can you add some regression test coverage for this?

lakinwecker:
@abn I looked for tests related to this stuff and failed to find any.  Are there any other unit tests that do things similarily that I can follow to make tests? If so, I can try and add them, yes.

radoering:
@lakinwecker Tests are in [test_password_manager.py](https://github.com/python-poetry/poetry/blob/master/tests/utils/test_password_manager.py). Probably, you should add two tests similar to https://github.com/python-poetry/poetry/blob/15f6ec1758d783d4cb3070a8027725bc611c0a90/tests/utils/test_password_manager.py#L190-L195

one with a simple null and one with a chained null. (And maybe you should rename the fixture `with_chained_keyring` to `with_chained_fail_keyring` so you can add `with_chained_null_keyring`). The fixtures are here: https://github.com/python-poetry/poetry/blob/15f6ec1758d783d4cb3070a8027725bc611c0a90/tests/conftest.py#L109-L127

lakinwecker:
`test_keyring_with_chainer_backend_and_not_compatible_only_should_be_unavailable` doesn't fail when you mess with the "fail" code that I modified.  So I'm not convinced this test properly exercises it.

lakinwecker:
Alright, the mocks were wrong for those chainer tests. I fixed them and now the tests fail if I change the name of the 'bad' keyrings in the `password_manager.py` checks.  So, I think this adds to tests to properly exercise it.

Is there documentation I should also update?

PR #5255, opened 2022-03-01
chore(repositories): remove type errors

branchvincent:
[`Distribution`](https://github.com/python/importlib_metadata/blob/14cce75299645467adcd17352cb07caada32c444/importlib_metadata/__init__.py#L513) seems to be mostly an abstract class, with a single concrete [`PathDistribution`](https://github.com/python/importlib_metadata/blob/14cce75299645467adcd17352cb07caada32c444/importlib_metadata/__init__.py#L910). We assume we're working with a`PathDistribution`, which has a `_path` but [not publicly typed](https://github.com/python/typeshed/blob/cf3ea5b6e651cea36c3539f66fdb33b585fc85d8/stdlib/importlib/metadata/__init__.pyi#L187-L190), so not sure what else to do besides `type: ignore`

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -120,7 +120,6 @@ module = [
   'poetry.installation.executor',
   'poetry.installation.installer',
   'poetry.installation.pip_installer',
-  'poetry.repositories.installed_repository',
   'poetry.utils.env',
 ]
 ignore_errors = true
--- a/src/poetry/repositories/installed_repository.py
+++ b/src/poetry/repositories/installed_repository.py
@@ -106,7 +106,7 @@ class InstalledRepository(Repository):
     ) -> Package:
         # We first check for a direct_url.json file to determine
         # the type of package.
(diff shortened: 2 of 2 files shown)


dimbleby:
I think that you are running the typechecking only through precommit?  That is in an environment that does not install most of the poetry dependencies, and as a result it misses most of the errors.  So I believe that this MR - or if not specifically this MR then the recent sequence of typechecking MRs - are in a false paradise.

https://gist.github.com/dimbleby/b9133cbafee6e127418a538bd7986f75 is the set of errors that I currently get when running mypy from within the poetry environment.

dimbleby:
#5279

branchvincent:
> I think that you are running the typechecking only through precommit?

good catch! you're definitely right, thanks for calling that out

dimbleby:
Looks good to me - I also don't see a better way.

neersighted (approved):

PR #5262, opened 2022-03-02
chore(utils): remove type errors

eyllanesc:


src/poetry/utils/exporter.py
```diff
@@ -174,9 +177,11 @@ def _export_requirements_txt(
...
-            output.write(decoded)
-        except AttributeError:
+        if isinstance(output, str):
             filepath = cwd / output
             with filepath.open("w", encoding="utf-8") as f:
                 f.write(decoded)
+            return
+        with contextlib.suppress(AttributeError):
+            # FIXME
+            output.write(decoded)
```
branchvincent, line 187: I imagine we don't need to worry about `AttributeError` and can just rely on the instance check:

```suggestion
        output.write(decoded)
```

dimbleby:
#5279

dimbleby:
I seem to have largely duplicated this in #5464, sorry about that.  It's a shame both because of the unnecessary extra work but also because there are things here that I probably like more than what I did, in particular the extra types you added look interesting.

Secrus:
This PR has already been addressed by other contributions and is no longer valid.

PR #4520, opened 2021-09-18
Check current python executable version

yokomotod:
Hi.

As #3426 said, current poetry only checks the python constraint in pyproject.toml when creating a virtualenv.

This PR added a python version check for created virtualenv as well, at each command execution.

### Demo

(Supporse python 3.8 and 3.9 are installed)

First, `poetry install` with `pyproject.tom` which requires python3.8:

```toml
[tool.poetry]
name = "ptry"
version = "0.1.0"
description = ""
authors = ["Your Name <you@example.com>"]

[tool.poetry.dependencies]
python = "~3.8"

[tool.poetry.dev-dependencies]

[build-system]
requires = ["poetry-core>=1.0.0"]
build-backend = "poetry.core.masonry.api"

```

It creates virtualenv with python 3.8.x executable.

```
$ poetry run python -V
Python 3.8.10
```

Then change `pyproject.tom` to `python = "^3.9"`.

With this PR, now `poetry install` aborts with error message:

```

  InvalidCurrentPythonVersionError

  Current Python version (3.8.10) is not allowed by the project (^3.9).
  Please change python executable via the "env use" command.

  at ~/.poetry/lib/poetry/utils/env.py:651 in create_venv
       647│             current_python = Version.parse(
       648│                 ".".join(str(c) for c in env.version_info[:3])
       649│             )
       650│             if not self._poetry.package.python_constraint.allows(current_python):
    →  651│                 raise InvalidCurrentPythonVersionError(
       652│                     self._poetry.package.python_versions, current_python
       653│                 )
       654│             return env
       655│ 
```

As above message suggests, by running `poetry env use python3.9` , `poetry install` will work again.

I expect this PR covers all commands which should check python version, like `uninstall`, `run`, `shell` etc.

# Pull Request Check List

Resolves: #3426

- [x] Added **tests** for changed code.
- ~~[ ] Updated **documentation** for changed code.~~

PR #5274, opened 2022-03-04
PEP610: handle optional vcs_info.requested_version

abn:
As per PEP610, presence of `vcs_info.requested_version` is optional in
direct_url.json for vcs packages. Prior to this change, Poetry would
crash when loading an environment with a vcs package not installed by
Poetry.

This PR also contains a linting fix for issue introduced in #4520.

**Reference:**
https://www.python.org/dev/peps/pep-0610/#specification
> A requested_revision key (type string) MAY be present naming a branch/tag/ref/commit/revision/etc (in a format compatible with the VCS) to install.

### DIFF
--- a/src/poetry/repositories/installed_repository.py
+++ b/src/poetry/repositories/installed_repository.py
@@ -201,8 +201,10 @@ class InstalledRepository(Repository):
             # VCS distribution
             source_type = url_reference["vcs_info"]["vcs"]
             source_url = url_reference["url"]
-            source_reference = url_reference["vcs_info"]["requested_revision"]
             source_resolved_reference = url_reference["vcs_info"]["commit_id"]
+            source_reference = url_reference["vcs_info"].get(
+                "requested_revision", source_resolved_reference
+            )
 
         package = Package(
             distribution.metadata["name"],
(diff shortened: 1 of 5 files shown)


src/poetry/repositories/installed_repository.py
```diff
@@ -201,8 +201,10 @@ def create_package_from_pep610(cls, distribution: metadata.Distribution) -> Pack
             # VCS distribution
             source_type = url_reference["vcs_info"]["vcs"]
             source_url = url_reference["url"]
-            source_reference = url_reference["vcs_info"]["requested_revision"]
             source_resolved_reference = url_reference["vcs_info"]["commit_id"]
+            source_reference = url_reference["vcs_info"].get(
+                "requested_revision", source_resolved_reference
+            )
```
finswimmer, line 207: What's the reason for not setting `source_reference` to `None` if `requested_revision` is not available?

abn: The rationale was that we do not want unexpected runtime failures when something references this. But it also seems is the least worst behavior.

finswimmer (approved):

PR #5275, opened 2022-03-04
fixes pre-commit

eyllanesc:


abn:
This should be a dupe of commit included in #5274.

Issue #5017, opened 2022-01-10
Code Quality: Incremental adoption of mypy type checks


finswimmer:
`poetry` and `poetry-core` are full type annotated and checking types by mypy is prepared. However these checks are deactivated for most of the modules.

In this ticket I like to invite everybody to pick one or more modules in `poetry` and `poetry-core` and fix errors reported by mypy. Please link any PR you are starting to this issue. 

In process:

* [x] https://github.com/python-poetry/poetry-core/pull/253
* [ ] https://github.com/python-poetry/poetry/pull/5262
* [ ] https://github.com/python-poetry/poetry/pull/5260
* [x] https://github.com/python-poetry/poetry/pull/5258
* [x] https://github.com/python-poetry/poetry/pull/5255
* [x] https://github.com/python-poetry/poetry/pull/5399
* [x] https://github.com/python-poetry/poetry/pull/5402

dimbleby:
@finswimmer can I suggest that you'll likely get better engagement with this sort of thing if maintainers are able to merge the PRs reasonably promptly?  I know that I would much more likely have felt encouraged to continue the work of https://github.com/python-poetry/poetry-core/pull/253 if I saw it being committed.

finswimmer:
Hello @dimbleby,

I'm one of the maintainers (just if you haven't know it). I absolutely agree with your sentiment that reacting on a PR in short time would get better engagement. But unfortunately this is not always possible. I think you already know it. We are all doing this in our free time. And sometimes one has to decide whether to review a PR a comment on some of the 30-50 new issues/questions/feature request per week. Some reviews require also discussion between the maintainers, which takes time.

If you feel that something takes to long, please ping us in the isssue/PR or on the discord server.

fin swimmer

dimbleby:
I intend no upset, I'm here because I think poetry is a good thing, and I'm nothing but grateful to those who've got us this far.  Please be assured that I am not trying to knock or discourage anyone, my apologies if it seemed otherwise.

But yeah, my experience is that contributing to poetry generally is a slow business.  Probably #4595 would have been a better place to talk about it though.

KrakenBite:
Hi! I would like to work on this issue as I am interested in adoption of mypy. Can I just start?

abn:
[...]

PR #5279, opened 2022-03-05
run mypy in proper environment

dimbleby:
Experimenting with running mypy in the proper environment via the github workflow, rather than precommit.  

(Or, perhaps, in addition to precommit... for discussion)

As this should show, if everything works as expected, the fact that precommit runs in an environment that does not include most of poetry's dependencies is causing mypy to fail to report lots of errors.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -85,6 +85,9 @@ jobs:
       - name: Run pytest
         run: poetry run python -m pytest -p no:sugar -q tests/
 
+      - name: Run mypy
+        run: poetry run mypy
+
       - name: Run pytest (integration suite)
         run: poetry run python -m pytest -p no:sugar -q --integration tests/integration
 
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -81,14 +81,6 @@ repos:
       - id: flake8
         additional_dependencies: *flake8_deps
 
-  - repo: https://github.com/pre-commit/mirrors-mypy
-    rev: v0.950
-    hooks:
-      - id: mypy
-        pass_filenames: false
-        additional_dependencies:
-          - types-requests
-
   - repo: https://github.com/pre-commit/pre-commit
     rev: v2.18.1
     hooks:
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -122,6 +122,13 @@ $ poetry install
 $ poetry run pytest tests/
 ```
 
+Poetry uses [mypy](https://github.com/python/mypy) for typechecking, and the CI
+will fail if it finds any errors.  To run mypy locally:
+
+```bash
+$ poetry run mypy
+```
+
 Poetry uses the [black](https://github.com/psf/black) coding style and you must ensure that your
 code follows it. If not, the CI will fail and your Pull Request will not be merged.
 
--- a/docs/contributing.md
+++ b/docs/contributing.md
@@ -139,6 +139,13 @@ $ poetry install
 $ poetry run pytest tests/
 ```
 
+Poetry uses [mypy](https://github.com/python/mypy) for typechecking, and the CI
+will fail if it finds any errors.  To run mypy locally:
+
+```bash
+$ poetry run mypy
+```
+
 Poetry uses the [black](https://github.com/psf/black) coding style and you must ensure that your
 code follows it. If not, the CI will fail and your Pull Request will not be merged.
 
--- a/poetry.lock
+++ b/poetry.lock
@@ -331,6 +331,33 @@ optional = false
 python-versions = "*"
 
 [[package]]
+name = "mypy"
+version = "0.950"
+description = "Optional static typing for Python"
+category = "dev"
+optional = false
+python-versions = ">=3.6"
+
+[package.dependencies]
+mypy-extensions = ">=0.4.3"
+tomli = {version = ">=1.1.0", markers = "python_version < \"3.11\""}
+typed-ast = {version = ">=1.4.0,<2", markers = "python_version < \"3.8\""}
+typing-extensions = ">=3.10"
+
+[package.extras]
+dmypy = ["psutil (>=4.0)"]
+python2 = ["typed-ast (>=1.4.0,<2)"]
+reports = ["lxml"]
+
+[[package]]
+name = "mypy-extensions"
+version = "0.4.3"
+description = "Experimental type system extensions for programs checked with the mypy typechecker."
+category = "dev"
+optional = false
+python-versions = "*"
+
+[[package]]
 name = "nodeenv"
 version = "1.6.0"
 description = "Node.js virtual environment builder"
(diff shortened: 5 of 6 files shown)


dimbleby:
Well indeed that makes the point.  The typing of this project is nowhere near as clean as precommit suggests.

I invite discussion on what the right approach is.  

In the mean time, I will push a second commit to this MR that ignores enough of these errors to go green again.  At least we're then explicit about the typing being faulty.

eyllanesc:
If the feature is going to be implemented then I would recommend that it should be done when poetry fully supports mypy.

dimbleby:
> If the feature is going to be implemented then I would recommend that it should be done when poetry fully supports mypy.

Could you explain what your thinking is?

The intention at #4510 was to make it possible incrementally to improve the typing of this project: by enabling typechecking, then gradually fixing the errors and un-ignoring them.  That seems very sensible.

Do you mean that you think we should not turn on typechecking before the project is fully clean?  Perhaps I have misunderstood.

eyllanesc:
I put myself on the side of the contributor. Let's say I implement some feature in a new file using an existing class that doesn't have typing enabled or has typing errors so my new class then I have 2 options: 1. Add to override list or 2. Fix the problem of the existing class for the github workflows to pass. Although thinking about it, I don't see anything wrong with it.

Since I changed my mind I recommend using the pre-commit and running not only mypy but the other tasks.

.github/workflows/main.yml
```diff
@@ -84,3 +84,6 @@ jobs:
 
       - name: Run pytest
         run: poetry run python -m pytest -p no:sugar -q tests/
+
+      - name: Run mypy
+        run: poetry run mypy src
```
abn, line 89: Do we need this run for every platform/python combination?

eyllanesc: No, I don't think so, it is enough with the highest version of python and in ubuntu. mypy does not change in other environments.

radoering: > Do we need this run for every platform/python combination?

I think, we should run it on every platform. #5135 proved that mypy can fail on Windows even if it succeeds on Linux.

branchvincent (dismissed):

.github/workflows/main.yml
branchvincent: maybe include `tests` as well?

finswimmer: Running against all platform/python combination seems to me a bit overkill. I would prefer that `mypy` gets its own job and is running against all platform but only one for one specified python. (But maybe spinning up these jobs have more overhead, than running within the "tests" job? :thinking: )

Furthermore we should drop `mypy` from the `pre-commit` checks.

finswimmer (changes requested):

.github/workflows/main.yml
dimbleby: I put it all in the same job because indeed the cost of running mypy is surely dwarfed by the cost of installing poetry and running poetry install.  I kind of agree that it would be cleaner to have it separate; but also I don't care very much.

Similarly on multiple platform/python combinations - different platforms and python versions can give different results so it's not crazy to run against all of them.  On the other hand I don't suppose we'd lose very much by just picking one; I don't really care about that detail either.

I mostly intend this specific MR as a demonstration that the pre-commit approach is missing stuff, and that we should do better.  Feel free either to merge this as it is, or a rearranged version as any of you prefer.

radoering:
Once again we had a platform specific mypy failure in master (see #5524) that would have been revealed by this PR. I suppose the more mypy checking we are doing the more often such issues will occur.

Although running against all platform/python combinations may seem a bit overkill, it's the most accurate way and since combinations run in parallel it shouldn't affect the overall time to wait for CI too much. I don't know if we need to budget CI time, do we? If we don't, I'd just say go for it.

As for the pre-commit hook, of course it's nice to get early feedback from a pre-commit hook but pre-commit passing does not mean CI will pass. If you don't run tests (strictly speaking on all platforms with all python versions) locally you only get the feedback later in CI. With mypy it's just the same. I'd say better to have thorough mypy checks in CI than a mediocre mypy pre-commit hook. So let's get rid of it. (When removing the pre-commit hook, we should add a statement about running mypy in the [contribution guide](https://github.com/python-poetry/poetry/blob/master/CONTRIBUTING.md#local-development)).

In my opinion, we could have already benefited from this PR. I may miss something but I don't see any significant drawbacks.

@abn @finswimmer Any strong feelings against this PR?

dimbleby:
I have quite a few typechecking MRs open (I know I'm not the only one), I'd love to see them getting merged.

There are two competing / cooperating things going on: MRs to make the typechecking stricter, and MRs to fix the errors

On the making things stricter front:
- this one tightens things up in poetry, we should do the equivalent in poetry-core (though I can report that this will not currently reveal anything new)
- Adding py.typed in both projects is desirable, IMO the typing has gotten far enough that this is now worthwhile
- Not yet started, but it would be good to turn up the mypy strictness flags in this project (I am satisfied that they are already strict in poetry-core)

On the fixing things front: well, there are several MRs listed in #5017.  All are as far as I know ready to go, the most relevant to right now is #5485 which fixes many of the errors revealed here.

dimbleby:
rebase post #5485 - note that we no longer need to add additional mypy exceptions to pyproject.toml

radoering:
@dimbleby It seems we might find consensus when running mypy via [pytest-mypy](https://pypi.org/project/pytest-mypy/). At first glance, it just looks as good to me as running mypy directly. Do you feel like trying this? I'd appreciate your opinion.

dimbleby:
Looks as though pytest-mypy doesn't Just Work as a drop-in replacement, though perhaps configuration settings could be found to make it do so.  Eg it is hooking into pytest's autodiscovery of files rather than respecting the configuration in pyproject.toml.

Also any error messages get mixed up with the test script runs which I at least find confusing.

Why is this better than running mypy directly?

radoering:
In my personal opinion it is not. There are some concerns that it's an additional step contributors have to perform before committing (before: pytest + pre-commit, now: pytest + mypy + pre-commit).

Nevertheless, I have the go from @abn that we can proceed with this PR as is and solve that later. Just one question and two requests:

- You've configured mypy to only run on `src`, not on `test`. I suppose there are still issues in `test` that have to be resolved? (Can be done later in a separate PR.)
- Can you please adapt the [contribution guide](https://github.com/python-poetry/poetry/blob/master/CONTRIBUTING.md#local-development) so it says to run mypy in addition to pytest and pre-commit?
- Probably, we should remove the mypy pre-commit hook, shouldn't we?

dimbleby:
> I suppose there are still issues in test that have to be resolved?

Yes.  I consider the typechecking of test scripts to be very low priority; but if others care to come along and make it happen then I've no objection.

Pushed updates to the contribution guide and to remove the pre-commit hook.

radoering (approved):

PR #5288, opened 2022-03-08
show poetry and poetry-core version in output of `poetry about`

finswimmer:
Adds info about the version of poetry and poetry-core to the output of `poetry about`.

### DIFF
--- a/src/poetry/console/commands/about.py
+++ b/src/poetry/console/commands/about.py
@@ -10,9 +10,14 @@ class AboutCommand(Command):
     description = "Shows information about Poetry."
 
     def handle(self) -> None:
+        from poetry.utils._compat import metadata
+
         self.line(
-            """\
-<info>Poetry - Package Management for Python</info>
+            f"""\
+<info>Poetry - Package Management for Python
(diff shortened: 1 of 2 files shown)


src/poetry/console/commands/about.py
```diff
@@ -10,9 +10,16 @@ class AboutCommand(Command):
...
+        from poetry.__version__ import __version__
+
         self.line(
-            """\
-<info>Poetry - Package Management for Python</info>
+            f"""\
+<info>Poetry - Package Management for Python
+
+Version: {__version__}
+Poetry-Core Version: {poetry.core.__version__}</info>
```
abn, line 22: We should not rely on `__version__` here, but should rather use importlib. We should consider dropping it in the future.

finswimmer: Using `importlib` was my first try, but for whatever reason I changed it. Happy to use it again.

finswimmer: Done.

abn (approved):

PR #5301, opened 2022-03-12
support for POETRY_CONFIG_DIR, POETRY_DATA_DIR and POETRY_CACHE_DIR environment variables

mohan43u:
* use PROJECT_HOME if available instead of XDG variables in unix systems

# Pull Request Check List

I have created this new PR as I'm not at all satisfied modifying XDG variables in order to make poetry properly set its cache dir under my preferred directory. For some reason, my previous PR https://github.com/python-poetry/poetry/pull/3550 was rejected because I tried to modify depreciated get-poetry.py (which is silly because when that PR created, get-poetry.py was the primary way to install poetry) and I tried to use the already available POETRY_HOME variable to achieve my purpose. 

Anyway, I rectified both  issues and creating this PR in hope that poetry devs understand XDG variables are common variables used by many application and should be used as fallback rather then expecting users to modify XDG in order to make poetry store its cache under their preferred directory.

May fix https://github.com/python-poetry/poetry/issues/2445

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -115,6 +115,34 @@ This also works for secret settings, like credentials:
 export POETRY_HTTP_BASIC_MY_REPOSITORY_PASSWORD=secret
 ```
 
+## Default Directories
+
+Poetry uses the following default directories:
+
+### Config Directory
+
+- Linux: `$XDG_CONFIG_HOME/pypoetry` or `~/.config/pypoetry`
+- Windows: `%APPDATA%\pypoetry`
+- MacOS: `~/Library/Preferences/pypoetry`
+
+You can override the Config directory by setting the `POETRY_CONFIG_DIR` environment variable.
+
+### Data Directory
+
+- Linux: `$XDG_DATA_HOME/pypoetry` or `~/.local/share/pypoetry`
+- Windows: `%APPDATA%\pypoetry`
+- MacOS: `~/Library/Application Support/pypoetry`
+
+You can override the Data directory by setting the `POETRY_DATA_DIR` or `POETRY_HOME` environment variables. If `POETRY_HOME` is set, it will be given higher priority.
+
+### Cache Directory
+
+- Linux: `$XDG_CACHE_HOME/pypoetry` or `~/.cache/pypoetry`
+- Windows: `%APPDATA%\pypoetry\Cache`
+- MacOS: `~/Library/Caches/pypoetry`
+
+You can override the Cache directory by setting the `POETRY_CACHE_DIR` environment variable.
+
 ## Available settings
 
 ### `cache-dir`


Secrus:
This seems to be answering a very specific needs and is not really generic. `PROJECT_HOME` might be already used by someone in their setup and this change would break it.

mohan43u:
> This seems to be answering a very specific needs and is not really generic.

the problem here is, there is no way to make poetry use a specific directory to keep all its stuff, if you look at pip, it can keep all its stuff under PYTHONUSERBASE and PIP_CACHE_DIR, pip dont use XDG_* variables (like poetry). This whole thing is all about providing way for poetry to keep its stuff under one place and not to depend on XDG_* variables.

> `PROJECT_HOME` might be already used by someone in their setup and this change would break it.

yeah, PROJECT_HOME is wrong, POETRY_HOME is already wrong, If I change it to BLAHBLAHBLAHBLAH, it is also going to be used by somebody so it is also wrong. funny thing is, somebody said I should not use POETRY_HOME for anything, even poetry dont use POETRY_HOME apart from the installer. it looks like POETRY_HOME will be the datadir if it was set according to this line https://github.com/python-poetry/poetry/blob/a80f9850fc5e8c6aaefb19693d286aa6204a9f3a/src/poetry/locations.py#L43

Apologies for my rant. We can change this to something unique like POETRY_PROJECT_HOME, this is not going to be an issue.

All I'm asking is provide a way to keep poetry's files in one place just like how I can keep pip's files in one place by setting PYTHONUSERBASE and PIP_CACHE_DIR pointing to single directory.

Secrus:
Thank you for your input on the matter, I understand it better now. Would custom `POETRY_CONFIG_DIR` and `POETRY_CACHE_DIR` satisfy your needs? If you are willing to make a PR, I think those extra variables with priority: custom > XDG* > defaults, and of course documentation for that, would be a good addition.

neersighted:
I think this can be converted into a simple docs dir -- the config mechanism supports environmental variables for all settings, so we have always supported `POETRY_CACHE_DIR`, `POETRY_CONFIG_DIR`, and `POETRY_DATA_DIR` -- try these out on your current install :)

mohan43u:
> I think this can be converted into a simple docs change -- the config mechanism supports environmental variables for all settings (and this is documented, though we don't call it out for this use case), so we have always supported `POETRY_CACHE_DIR`, `POETRY_CONFIG_DIR`, and `POETRY_DATA_DIR` -- try these out on your current install smile_cat

neersighted:
@mohan43u are you interested in rolling this PR over into documentation of the default paths/how to override them with env vars?

mohan43u:
> the config mechanism supports environmental variables for all settings (and this is documented, though we don't call it out for this use case), so we have always supported `POETRY_CACHE_DIR`, `POETRY_CONFIG_DIR`, and `POETRY_DATA_DIR`

https://python-poetry.org/docs/configuration/#using-environment-variables

I think the issue is outside of config module, the reason why https://github.com/python-poetry/poetry/issues/2445 happening is because of cache/repositories directory created using CACHE_DIR

So these environment variables are required outside of config.

mohan43u:
> Wouldn't it be ok to use?

`REPOSITORIES_CACHE_DIR`is outside of config module. only config module can handles this setting https://python-poetry.org/docs/configuration/#cache-dir

mohan43u:
> @mohan43u are you interested in rolling this PR over into documentation of the default paths/how to override them with env vars?

@neersighted we have to address `REPOSITORY_CACHE_DIR` issue. I think the code change which is available in this PR fixes this issue.

neersighted:
> > @mohan43u are you interested in rolling this PR over into documentation of the default paths/how to override them with env vars?
> 
> @neersighted we have to address `REPOSITORY_CACHE_DIR` issue. I think the code change which is available in this PR fixes this issue.

I'm not sure I understand the difference between `REPOSITORY_CACHE_DIR` and `POETRY_CACHE_DIR`. I can say that `POETRY_CACHE_DIR` is definitely respected -- try setting `POETRY_CACHE_DIR=./cache` and then doing `poetry install` on Poetry itself.

Can you please help me understand the specific expected behavior vs result, and how this change (which duplicates the existing env var logic in the config module) resolves it?

mohan43u:
> > > @mohan43u are you interested in rolling this PR over into documentation of the default paths/how to override them with env vars?
> > 
> > 
> > @neersighted we have to address `REPOSITORY_CACHE_DIR` issue. I think the code change which is available in this PR fixes this issue.
> 
> I'm not sure I understand the difference between `REPOSITORY_CACHE_DIR` and `POETRY_CACHE_DIR`. I can say that `POETRY_CACHE_DIR` is definitely respected -- try setting `POETRY_CACHE_DIR=./cache` and then doing `poetry install` on Poetry itself.
> 
> Can you please help me understand the specific expected behavior vs result, and how this change (which duplicates the existing env var logic in the config module) resolves it?

mohan43u:
> I'm not sure I understand the difference between `REPOSITORY_CACHE_DIR` and `POETRY_CACHE_DIR`. I can say that `POETRY_CACHE_DIR` is definitely respected -- try setting `POETRY_CACHE_DIR=./cache` and then doing `poetry install` on Poetry itself.

I set `POETRY_CACHE_DIR=~/testpoetry/cache`,  I still see `~/.cache/pypoetry/cache/repositories` created by poetry which it should not suppose to do, this is the reason for https://github.com/python-poetry/poetry/issues/2445. This PR is basically modifying the behavior of poetry and make it more flexible in configuring default directories.

This is not at all relevant to the config environment variables which you are confusing with.

neersighted:
Ah, I see, we're not resolving variables in order right now -- looks like the change that introduced REPOSITORY_CACHE_DIR forgot to think about config resolution order.

@abn came up with a fix for it offline, that fixes the issue by solving the underlying cause and has tests at #5672. I've merged that PR -- could you please test on the latest master and make sure it meets your use case?

mohan43u:
thanks for fixing this. I can now happily use POETRY_CACHE_DIR, POETRY_CONFIG_DIR (and most probably use POETRY_HOME) to achieve what I want.

mohan43u:
> We could still use a PR that documents the default paths and how to override them with these variables explicitly if you're interested grinning

Sure, I'll update the documentation. But I have doubt whether `POETRY_DATA_DIR` will work or not? as I see data_dir() definition in locations.py uses POETRY_HOME environment variable if available, otherwise it is not using `POETRY_DATA_DIR` environment variable. It directly uses XDG location.

POETRY_CONFIG_DIR environment variable was properly used before falling back to XDG location.

mohan43u:
@neersighted I already updated this PR with the docs change. Let me know if any other changes we need to this PR.

docs/configuration.md
```diff
@@ -9,6 +9,35 @@ menu:
...
 ---
 
+# Default Directories
+
+Poetry uses these following default directories
+
+## Config Directory
+
+- Linux: `$XDG_CONFIG_HOME/$appname` or `~/.config/$appname`
+- Windows:`%USERPROFILE%\AppData\Local\$appauthor\$appname` (not roaming) or`%USERPROFILE%\AppData\Roaming\$appauthor\$appname` (roaming)
```
Secrus, line 19:
```suggestion
- Windows: `%USERPROFILE%\AppData\Local\$appauthor\$appname` (not roaming) or `%USERPROFILE%\AppData\Roaming\$appauthor\$appname` (roaming)
```
What are `roaming` and `not roaming`? and `$appauthor` and `$appname`? Also, you forgot some spaces.

docs/configuration.md
```diff
@@ -9,6 +9,35 @@ menu:
...
+- MacOS: `~/Library/Preferences/$appname/$version`
+
+You can override this Config directory by setting `POETRY_CONFIG_DIR` environment variable
+
+
+## Data Directory
+
+- Linux: `$XDG_DATA_HOME/$appname` or `~/.local/share/$appname`
+- Windows:`%USERPROFILE%\AppData\Local\$appauthor\$appname` (not roaming) or`%USERPROFILE%\AppData\Roaming\$appauthor\$appname` (roaming)
+- MacOS: `~/Library/Application Support/$appname/$version`
```
Secrus, line 29: the same as above. "magic" variables and spaces.

docs/configuration.md
```diff
@@ -9,6 +9,35 @@ menu:
...
+
+You can override this Data Directory by setting `POETRY_DATA_DIR` environment variable. If `POETRY_HOME` environment variable is available, then `POETRY_HOME` will be given higher priority.
+
+## Cache Directory
+
+- Linux: `$XDG_CACHE_HOME/$appname` or `~/.cache/$appname`
+- Windows:`%USERPROFILE%\AppData\Local\$appauthor\$appname\Cache\$version`
```
Secrus, line 36:
```suggestion
- Windows: `%USERPROFILE%\AppData\Local\$appauthor\$appname\Cache\$version`
```

Secrus (changes requested):

docs/configuration.md
neersighted: I would actually rewrite these at `%APPDATA%` and `%LOCALAPPDATA%`.

neersighted (changes requested):
Gonna take a look in a bit, but I think we're definitely going to need a SPAG and formatting pass.

mohan43u:
@Secrus @neersighted pushed again with recommended changes.

mohan43u:
@neersighted @Secrus could you please check if anything required?

docs/configuration.md
```diff
@@ -9,6 +9,35 @@ menu:
...
+
+## Cache Directory
+
+- Linux: `$XDG_CACHE_HOME/pypoetry` or `~/.cache/pypoetry`
+- Windows: `%APPDATA%\pypoetry\Cache`
+- MacOS: `~/Library/Caches/pypoetry`
+
+You can override this Cache directory by setting `POETRY_CACHE_DIR` environment variable
+
 # Configuration
```
Secrus, line 41: This is now not rendered (as this is second h1 on one page). What's more, text after this header should be reworked as this doubles the information about config directories.

Secrus (changes requested):

docs/configuration.md
mohan43u: @Secrus I reworked and put "Default Directories" section under "Configuration" header. Hope this will be proper now.

Secrus (approved):

docs/configuration.md
```diff
@@ -115,6 +115,35 @@ This also works for secret settings, like credentials:
 export POETRY_HTTP_BASIC_MY_REPOSITORY_PASSWORD=secret
 ```
 
+## Default Directories
+
+Poetry uses these following default directories
```
neersighted, line 120:
```suggestion
Poetry uses the following default directories:
```

docs/configuration.md
```diff
@@ -115,6 +115,35 @@ This also works for secret settings, like credentials:
...
+
+### Config Directory
+
+- Linux: `$XDG_CONFIG_HOME/pypoetry` or `~/.config/pypoetry`
+- Windows: `%APPDATA%\pypoetry`
+- MacOS: `~/Library/Preferences/pypoetry`
+
+You can override this Config directory by setting `POETRY_CONFIG_DIR` environment variable
```
neersighted, line 128:
```suggestion
You can override the Config directory by setting the `POETRY_CONFIG_DIR` environment variable.
```

docs/configuration.md
```diff
@@ -115,6 +115,35 @@ This also works for secret settings, like credentials:
...
+
+
+### Data Directory
+
+- Linux: `$XDG_DATA_HOME/pypoetry` or `~/.local/share/pypoetry`
+- Windows: `%APPDATA%\pypoetry`
+- MacOS: `~/Library/Application Support/pypoetry`
+
+You can override this Data Directory by setting `POETRY_DATA_DIR` environment variable. If `POETRY_HOME` environment variable is available, then `POETRY_HOME` will be given higher priority.
```
neersighted, line 137:
```suggestion
You can override the Data directory by setting the `POETRY_DATA_DIR` or `POETRY_HOME` environment variables. If `POETRY_HOME` is set, it will be given higher priority.
```

docs/configuration.md
```diff
@@ -115,6 +115,35 @@ This also works for secret settings, like credentials:
...
+
+### Cache Directory
+
+- Linux: `$XDG_CACHE_HOME/pypoetry` or `~/.cache/pypoetry`
+- Windows: `%APPDATA%\pypoetry\Cache`
+- MacOS: `~/Library/Caches/pypoetry`
+
+You can override this Cache directory by setting `POETRY_CACHE_DIR` environment variable
```
neersighted, line 145:
```suggestion
You can override the Cache directory by setting the `POETRY_CACHE_DIR` environment variable.
```

neersighted (changes requested):
Needs a little refinement, and then all the commits should be squashed down to one. Looks good to merge after that.

neersighted (approved):

Issue #5019, opened 2022-01-10
KeyError: Package('pycurl', '7.44.1', source_type='legacy' ...) on celery update


HoneyryderChuck:
- [x] I am on the [latest](https://github.com/python-poetry/poetry/releases/latest) Poetry version.
- [x] I have searched the [issues](https://github.com/python-poetry/poetry/issues) of this repo and believe that this is not a duplicate.
- [x] If an exception occurs when executing a command, I executed it again in debug mode (`-vvv` option).

- **OS version and name**: docker image (`python:3.9-buster`)
- **Poetry version**: 1.1.12

```toml
# relevant contents of the toml
boto3 = "*"
celery = {extras = ["sqs"], version = "*"}
# this is what I have to have in order for things to work
# pycurl = "*"
```

## Issue

I want to update `celery` to the latest version (5.2.3). Nothing in the `pyproject.toml` stops me, i.e. no transitive dependency mismatches. However, when I `poetry update`, it's unable to upgrade `pycurl` to latest deployed version. So then I add `pycurl` to the toml, although I don't use it directly. update works. Now, I have to keep `pycurl` around in the toml for everything to work. If I remove it, I get the following error on `poetry install`:

```
KeyError

  Package('pycurl', '7.44.1', source_type='legacy', source_url='https://our-pypi-proxy', source_reference='company')

  at ~/.local/lib/python3.9/site-packages/poetry/puzzle/solver.py:270 in _solve
      266│
      267│                 continue
      268│
      269│             final_packages.append(package)
    → 270│             depths.append(results[package])
      271│
      272│         # Return the packages in their original order with associated depths
      273│         return final_packages, depths
      274│
```

I think I should have to carry a transitive dependency in the toml. Can someone see what's wrong?

HoneyryderChuck:
Indeed, that's the workaround.  But it shouldn't be required, that's poetry job.

henrysingleton:
Also happening for me, identical behaviour for poetry versions 1.1.4, 1.1.11 and 1.1.12. 
Using Python 3.9.7. 

Example dependencies that will reproduce the issue: 

```
[tool.poetry.dependencies]
python = "^3.9"
celery = {extras = ["sqs"], version = "5.2.3"}
```

Either manually installing pycurl first or adding `pycurl = "*"` to requirements also worked for me.

dimbleby:
duplicate #4683

philipGaudreau:
[...]

Issue #4683, opened 2021-10-28
Poetry cannot install dvc[ssh] package


lambda-science:
- [x] I am on the [latest](https://github.com/python-poetry/poetry/releases/latest) Poetry version.
- [x] I have searched the [issues](https://github.com/python-poetry/poetry/issues) of this repo and believe that this is not a duplicate.
- [x] If an exception occurs when executing a command, I executed it again in debug mode (`-vvv` option).

- **OS version and name**: Windows 10 21H1 with WSL2 (Ubuntu 20.04)
- **Poetry version**: 1.1.11
- **Link of a [Gist](https://gist.github.com/) with the contents of your pyproject.toml file**: https://pastebin.com/avWt6UNk

## Issue

This issue has been cross-posted in the dvc git repo. Here: https://github.com/iterative/dvc/issues/6886
dvc[ssh] can't be added to a project using Poetry env manager. 
When I create my package environnement using Poetry adding dvc with [ssh] extra create a weird unrelated error.
Here is my full error log from poetry https://pastebin.com/MBveHXbW

### Reproductibility
Minimal failing example:
1. poetry init
2. poetry add dvc[ssh]
3. poetry install

Full reproductibilty: get the pyproject.toml at the bottom, get an python installation with poetry installed and simply run `poetry install` with the file in the working directory.

I have the weird feeling that this issue is due to the fact that you need to first install dvc package and THEN dvc[ssh] (bcrypt looks missing) as described here: https://dvc.org/doc/install/linux#install-with-pip 
But that's not possible with Poetry, I can't have two line called "dvc" (one vanilla, one with the extra) and I can't have a line starting with dvc[ssh]

```
KeyError

  Package('bcrypt', '3.2.0')

  at ~/miniconda3/lib/python3.9/site-packages/poetry/puzzle/solver.py:270 in _solve
```
By removing the dvc with extras line from .toml file and doing `poetry add dvc` it works, but I can't use ssh plugin. Doing `poetry add dvc[ssh]` fails with the erros in the previous pastebin log.

EDIT: By using a fresh .toml file from poetry with only dvc[ssh] added, the same error arise.

sjawhar:
To help in debugging this, the bcrypt package doesn't get visited in the DFS here:

https://github.com/python-poetry/poetry/blob/1da3e7292f59fc539fd2ddc2c057fcbd98a9bd60/poetry/puzzle/solver.py#L206-L213

[...]

PR #5305, opened 2022-03-13
Solver fixes

dimbleby:
# Pull Request Check List

Resolves: #4683
Resolves: #5019

A series of commits tidying up the depth-first search in the solver, and a fix for #4683 (and its duplicate #5019)

- `seen` on the `PackageNode` should clearly be a set rather than a list, for efficient lookup
  -  (actually I suspect that this is redundant because of `visited` in the search, and should be removed altogether... maybe another time)

- `visited` in the search was strangely and unnecessarily complicated, I made it normal

- `dfs_visit()` always returned `True`; it has no need of a return value at all

- A testcase and fix for #4683: in the language of the added testcase, the fix is to make sure that we visit both `B` and `B[bar]` in the search

### DIFF
--- a/src/poetry/puzzle/solver.py
+++ b/src/poetry/puzzle/solver.py
@@ -339,7 +339,7 @@ class PackageNode(DFSNode):
                         and dependency.constraint.allows(pkg.version.stable)
                     )
                     and not any(
-                        child.package.name == pkg.name
+                        child.package.complete_name == pkg.complete_name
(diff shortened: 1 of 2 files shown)


dimbleby:
(I have rearranged my branches so that this MR contains only the fix, and the refactoring is in #5309)

radoering (approved):
LGTM 👍

PR #5309, opened 2022-03-13
Removing unused and unnecessary code

dimbleby:
# Pull Request Check List

A follow-up to my remark in #5305 that `seen` looked redundant on the node in the solver's depth first search, because the surrounding search already had a `visited`.

The doubt in my mind was that there was a second caller of `reachable()`, [here](https://github.com/python-poetry/poetry/blob/6485bc23d6497c7731e0f1a635f960b33f2ae99e/src/poetry/puzzle/solver.py#L217), and I hadn't taken the time to figure out the effect on that call site.

On reflection it is clear that the `seen` cache will be populated by the time we get to that second caller - which happens after the depth-first search has completed.  As a result, that second `reachable()` never returns anything other than an empty list.

(I also confirmed this by asserting as much in the code, and verifying that the test suite still passed)

Since nothing seems to be broken by that list being empty, I've assumed that we don't need that second `reachable()` call after all; and then it is clear that `seen` is indeed redundant.  So I've made two commits:

- remove the code which was made useless by `seen`
- remove `seen`

(I probably have given myself a small merge conflict with #5305, happy to sort that out if and when the first of these is merged).

### DIFF
--- a/src/poetry/puzzle/solver.py
+++ b/src/poetry/puzzle/solver.py
@@ -1,6 +1,5 @@
 from __future__ import annotations
 
-import enum
 import time
 
 from collections import defaultdict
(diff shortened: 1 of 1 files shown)


dimbleby:
(I have rearranged this and #5305 so that this one contains the refactoring and the other contains only the fix)

radoering (approved):
Good catch and well prepared. At some points, I had to think twice about it, but in the end I am quite sure you are absolutely right. 👍

PR #5319, opened 2022-03-15
Require --without-hashes for requirements.txt with VCS dependencies

nyuszika7h:
# Pull Request Check List

Resolves: #2463

This is a safer alternative to #3874. It will prevent Poetry from generating broken requirements.txt files, while also alleviating concerns on the previous pull request about potential security implications of hashes being automatically disabled without an explicit opt-in.

### DIFF
--- a/docs/cli.md
+++ b/docs/cli.md
@@ -619,6 +619,8 @@ This command is also available as a pre-commit hook. See [pre-commit hooks](/doc
 * `--dev`: Include development dependencies.
 * `--extras (-E)`: Extra sets of dependencies to include.
 * `--without-hashes`: Exclude hashes from the exported file.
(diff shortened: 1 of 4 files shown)


huonw:
This sounds like a reasonable idea to me. FWIW, we hit a similar issue with directory dependencies (`foo = { path = "foo/" }`), which also cannot be hashed: https://github.com/pypa/pip/blob/b73b128632a2fe228e611675b03e2c642f6fd55e/src/pip/_internal/exceptions.py#L509-L517

Potentially this could be generalised to cover that case too?

neersighted:
Closing this as the code has moved into https://github.com/python-poetry/poetry-plugin-export -- please discuss/PR this over there.

PR #2958, opened 2020-09-25
repository/legacy: calculate sha256 if unavailable

abn:
In some cases, legacy repositories might not provide a checksum as a
url fragment or use a deprecated algorithm. In these scenarios, this
change ensures that poetry downloads and calculates the sha256 checksum
for the file.

Resolves: #1631 #1553

PR #5326, opened 2022-03-16
add handling for non-sha hash in lock file and/or repository metadata when choosing a link

### LINKED ISSUE #4085: poetry 1.2.0a1: AttributeError when installing package from secondary repo

### LINKED ISSUE #4523: Poetry refuses to install package with correct hash

### LINKED ISSUE #4578: Enable multiple file hashing methods for backwards and forwards compatibility

ITProKyle:
This PR moves existing logic into a new method for reuse, pulls the logic from #4529 forward into 1.2, and adds a condition to handle legacy repositories (Sonatype Nexus) returning md5 hashes in its metadata instead of sha256 like poetry expects.

resolves #4523 
resolves #4578
resolves #4085 
supersedes #4740 

# Pull Request Check List

### DIFF
--- a/src/poetry/installation/chooser.py
+++ b/src/poetry/installation/chooser.py
@@ -6,6 +6,7 @@ from typing import TYPE_CHECKING
 
 from packaging.tags import Tag
 
+from poetry.repositories.pypi_repository import PyPiRepository
 from poetry.utils.patterns import wheel_file_re
 
 
@@ -100,6 +101,15 @@ class Chooser:
         selected_links = []
         for link in links:
             if not link.hash:
+                selected_links.append(link)
+                continue
+
+            if (
+                link.hash_name
+                and not link.hash_name.startswith("sha")
+                and isinstance(repository, PyPiRepository)
+                and repository.get_sha_hash_from_link(link) in hashes
+            ):
                 selected_links.append(link)
                 continue
 
--- a/src/poetry/repositories/pypi_repository.py
+++ b/src/poetry/repositories/pypi_repository.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
+import hashlib
 import logging
 import os
 import urllib.parse
@@ -453,3 +454,40 @@ class PyPiRepository(RemoteRepository):
 
     def _log(self, msg: str, level: str = "info") -> None:
         getattr(logger, level)(f"<debug>{self._name}:</debug> {msg}")
+
+    def get_sha_hash_from_link(self, link: Link) -> str | None:
+        """Get sha256|384|512 hash for a file from the provided link.
+
+        If the hash type included in the link is not sha256|384|512,
+        convert it to sha256.
+
+        """
+        file_hash = f"{link.hash_name}:{link.hash}" if link.hash else None
+
+        if not link.hash or (
+            link.hash_name not in ("sha256", "sha384", "sha512")
+            and hasattr(hashlib, link.hash_name)
+        ):
+            with temporary_directory() as temp_dir:
+                filepath = Path(temp_dir) / link.filename
+                self._download(link.url, str(filepath))
+
+                known_hash = (
+                    getattr(hashlib, link.hash_name)() if link.hash_name else None
+                )
+                required_hash = hashlib.sha256()
+
+                chunksize = 4096
+                with filepath.open("rb") as f:
+                    while True:
+                        chunk = f.read(chunksize)
+                        if not chunk:
+                            break
+                        if known_hash:
+                            known_hash.update(chunk)
+                        required_hash.update(chunk)
+
+                if not known_hash or known_hash.hexdigest() == link.hash:
+                    file_hash = f"{required_hash.name}:{required_hash.hexdigest()}"
+
+        return file_hash
--- a/src/poetry/installation/executor.py
+++ b/src/poetry/installation/executor.py
@@ -655,16 +655,28 @@ class Executor:
             package.name,
             archive_path,
         )
-        archive_hash = "sha256:" + file_dep.hash()
         known_hashes = {f["hash"] for f in package.files}
+        known_types = (h.split(":")[0] for h in known_hashes)
 
-        if archive_hash not in known_hashes:
+        archive_hashes = set()
+        for hash_type in known_types:
+            archive_hashes.add(f"{hash_type}:{file_dep.hash(hash_type)}")
+
+        if archive_hashes.isdisjoint(known_hashes):
             raise RuntimeError(
-                f"Hash for {package} from archive {archive_path.name} not found in"
-                f" known hashes (was: {archive_hash})"
+                f"Invalid hashes ({', '.join(sorted(archive_hashes))}) for"
+                f" {package} using archive {archive.name}. "
+                f"Expected one of {', '.join(sorted(known_hashes))}."
             )
 
-        return archive_hash
+        try:
+            _, sha256 = next(
+                filter(lambda h: h.startswith("sha256:"), archive_hashes)
+            ).split(":")
+        except StopIteration:
+            sha256 = file_dep.hash("sha256")
+
+        return f"sha256={sha256}"
 
     def _download_archive(self, operation: Install | Update, link: Link) -> Path:
         response = self._authenticator.request(
--- a/src/poetry/repositories/legacy_repository.py
+++ b/src/poetry/repositories/legacy_repository.py
@@ -1,14 +1,12 @@
 from __future__ import annotations
 
 import cgi
-import hashlib
 import re
 import urllib.parse
 import warnings
 
 from collections import defaultdict
 from html import unescape
-from pathlib import Path
 from typing import TYPE_CHECKING
 from typing import Any
 from typing import Iterator
@@ -35,12 +33,12 @@ from poetry.repositories.exceptions import RepositoryError
 from poetry.repositories.pypi_repository import PyPiRepository
 from poetry.utils.authenticator import Authenticator
 from poetry.utils.helpers import canonicalize_name
-from poetry.utils.helpers import download_file
-from poetry.utils.helpers import temporary_directory
 from poetry.utils.patterns import wheel_file_re
 
 
 if TYPE_CHECKING:
+    from pathlib import Path
+
     from poetry.core.packages.dependency import Dependency
 
 with warnings.catch_warnings():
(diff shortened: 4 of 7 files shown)


src/poetry/repositories/pypi_repository.py
```diff
@@ -453,3 +454,40 @@ def _download(self, url: str, dest: str) -> None:
...
+        """
+        file_hash = f"{link.hash_name}:{link.hash}" if link.hash else None
+
+        if not link.hash or (
+            link.hash_name not in ("sha256", "sha384", "sha512")
+            and hasattr(hashlib, link.hash_name)
+        ):
+            with temporary_directory() as temp_dir:
+                filepath = Path(temp_dir) / link.filename
+                self._download(link.url, str(filepath))
```
abn, line 473: This might get a bit problematic as it could end up folks downloading all available builds for all dependencies, well n+1. In the interim I reckon best to download to cache and use the downloaded file as is without moving it around. Atleast that will save redownload when relocking or re-solving.

src/poetry/installation/chooser.py
```diff
@@ -103,6 +104,15 @@ def _get_links(self, package: Package) -> list[Link]:
                 selected_links.append(link)
                 continue
 
+            if (
+                link.hash_name
+                and not link.hash_name.startswith("sha")
+                and isinstance(repository, PyPiRepository)
```
abn, line 110: Should we consider makign this apply broadly?

abn (changes requested):
While this is in the right direction, I wonder if we should first consider expanding the lockfile to include and/or allow non sha checksums if only md5 is available from non pypi sources. This can be done in interations as well.

src/poetry/installation/chooser.py
ITProKyle: Broadly as in for any repo type? The `get_sha_method_from_link` method would need to be added to `Repository` in that case since this check only exists to ensure the existence of the required method. I'm not entirely sure what the logic of the method would need to be for all the other types.

ITProKyle:
>  I wonder if we should first consider expanding the lockfile to include and/or allow non sha checksums if only md5 is available from non pypi sources

@abn I did a bit of research on this - specifically targeted at the logic affecting my use case. PR #2958 (resolving 2 issues) implemented the logic to calculate sha256 that I moved into the `get_sha_method_from_link` method. Allowing non-sha checksums would break exporting to requirements.txt for use with pip (unless using `--without-hashes`) as [pip no longer supports using md5, sha1, and sha224 checksums](https://pip.pypa.io/en/stable/cli/pip_install/#hash-checking-mode).

Poetry could calculate sha256 on export but given that pip no longer supports these weaker checksums due to security concerns, I would lean towards not wanting poetry to rely on them in a lock file either.

src/poetry/repositories/pypi_repository.py
ITProKyle: Reviewing the codebase, it looks like I'm going to have to pass `poetry.installation.chef.Chef` into the repository object or create it within the object to have access to the caching logic. To do so, I would need to know the `Env` when registering the repos (at a minimum when registering the legacy repos). I'm having trouble locating where I could pull `Env` from at that point. Any advice?

The other option I can think of would to be pass `Chef` as an argument to `Repository.get_release_info`/`Repository.package` & `Chooser.choose_for` since all of them would need to have the ability to download and calculate sha256 for a file. But, that seems excessive when it's only needed in cerin circumstances.

neersighted:
(also for those reading along at home, note that I have/had a PR at #4740 that stalled out for the same reasons -- I'm still rather interested in fixing this and will likely implement this once we figure out what the design should be)

ljnsn:
I understand that this is not trivial and you have to be careful about how you implement it. I assume it's still going to take a while, unfortunately that means we cannot install packages hosted either on Nexus or Artifactory with poetry, as they both return md5 hashes.

For what it's worth, my current workaround is

`poetry run pip install --index-url https://private-repo .`

Since pip supports installing from pyproject.tml this works, it only installs the main dependencies though, no dev dependencies. I install them with `poetry install --only dev`, but that of course only works if they're not hosted on the private repo.

dimbleby:
> we cannot install packages hosted either on Nexus or Artifactory with poetry, as they both return md5 hashes.

I don't know about Nexus, but I do know that recent artifactory returns sha256 hashes - per https://www.jfrog.com/jira/browse/RTFACT-18495

(I know this because that upgrade caused its own confusion: artifactory suddenly returning sha256 hashes meant that `poetry.lock` files needed rebuilding, and local poetry caches needed clearing.)

Of course you may or may not be able to upgrade your artifactory.

lacer93:
Older Gemfury repositories also return md5 hashes (I suppose the emphasis is on the older as I created a new one to create a reproducible environment and it now returns sha256 hashes while the old one still returns md5.. but I can't find a setting to change it). 

I understand that md5 is not as secure as sha256 and whatnot but it is still used in a lot of places and it is far from being insecure. At this point it would be wise to consider the fact that a lot of people are blocked by this (all of our pipelines are hardcoded to some ancient poetry & poetry-core version..) and are unable to use new poetry features. 

Would it be possible to simply add a config value in the .toml file to the private repo's section which would signal to poetry what hashing algorithm to use? That way, PyPI is untouched and the user has the ability (and the responsibility) to control the behaviour.

lacer93:
@neersighted @abn is there any progress here?

radoering:
I'm closing this PR because it seems to be stale and at least partially superseded by other PRs. If you think it still contains something worth reviving, feel free to resolve the merge conflicts and reopen it or just create a new PR.

Issue #5217, opened 2022-02-18
Poetry install localstack['runtime'] fails


orientalperil:
- [x] I am on the [latest](https://github.com/python-poetry/poetry/releases/latest) Poetry version.
- [x] I have searched the [issues](https://github.com/python-poetry/poetry/issues) of this repo and believe that this is not a duplicate.
- [x] If an exception occurs when executing a command, I executed it again in debug mode (`-vvv` option).

- **OS version and name**: Mac 10.15.6
- **Poetry version**: 1.1.13
- **Link of a [Gist](https://gist.github.com/) with the contents of your pyproject.toml file**: https://gist.github.com/orientalperil/0bf3229601e34a4c3e9484f748fd5278

## Issue

Poetry runs dependency resolver forever when I run `poetry add localstack['runtime']`

dimbleby:
It looks as though poetry's first attempt at a solution is in some way incompatible with awscli - and that it sits there working through the many hundreds of versions of awscli, failing to find one that is compatible.  This seems to cause it to build up such a huge number of constraints that it kinda grinds to a halt.

I was able to get to a solution in reasonable time by first adding an explicit dependency on a recent awscli, that seems to be enough to send poetry down a path that's not a dead end.  However that also ran into #4683...

abn:
Resolved by #5335

dimbleby:
This is not resolved by #5335 - that MR improves things, but still `localstack['runtime']` is uninstallable without hints about picking a recent awscli

abn:
I think the fix here is to constrain `awscli` together with the PR to get a more reasonable solving time. The add did succeed for me after an hour or so. :(

Doubt these cases will ever be improved with the metadata currently available.

dimbleby:
it's not really about the metadata (though it does help when that is available).  As best I recall from when I looked at this at #5335-time, we end up building Term objects with massively elaborate constraints, and then operations to compare those constraints become expensive.  It's 100% CPU-bound computation that is the bottleneck, not downloading and inspecting wheels.

I don't have great ideas about how to reorganise the data structures to fix this and you may be right that it will never get improved.  I might or might not have another think about it some day...

Personally I'd have left this open so that we acknowledge that it's still a problem, and maybe someone else will see it and feel like doing something about it.  But no big deal I guess.

abn:
[...]

PR #5335, opened 2022-03-20
Solver performance

dimbleby:
Some performance improvements derived from investigating #5217.

At time of writing, `poetry install localstack['runtime']` is unusably slow.
What happens is something like this:

- that package has a dependency on `pyyaml>=5.1`, which poetry resolves early on
  by picking pyyaml 6.0
- that package also has a dependency on `awscli>=1.14.18`
- there are a huge number of versions of `awscli` satisfying that constraint,
  every single one of them depending on a version of pyyaml that is older than
  6.0.

Poetry spends absolutely forever trying to find a compatible version of awscli,
and failing.
Eventually it will decide that they all fail, backtrack, and get to a solution -
but not fast!

(It could be more complicated than that: I think that there may be more than one
package playing the pyyaml role in this story.)

The first commit in this MR just tidies up a testcase that I encountered going
wrong while working out a fix.
The testcase says that cachy at `^0.1.0` was requested and that `0.1.0.dev1` was
installed.
But that would never happen, because poetry treats `0.1.0.dev1` as being earlier
than `0.1.0`.

The second commit borrows from <https://github.com/pubgrub-rs/pubgrub/pull/88>
to considerably speed up propagation of incompatibilities.

The third commit improves our repeated searches for suitable packages.
During the search, once a package has been rejected it can only become a valid
candidate again if we backtrack.
Since we build up constraints linear in size with the number of rejected
packages - like `>=1.14.18,<1.14.19 || >1.14.19,<1.14.20 || >1.14.20,...` - the
repeated work to reject non-matching packages becomes expensive.
I've avoided that by introducing a cache that discards packages as we go.

The fourth commit puts a cache on some `Term` methods that were showing up as
expensive.
(In spite of this, `Term.satisfies()` remains prominent in the profiling.
Probably these two caches should be seen as a sticking plaster trying - but not
fully succeeding - to compensate for something expensive in the algorithm.)

I have mostly been testing with a pyproject.toml that contains the following two
constraints:

```
awscli = ">=1.19.0"
localstack = { version = "0.14.1.1", extras = ["runtime"] }
```

where the first of these is to keep a cap on the number of `awscli` packages
that we must check.

The fixes in this MR take solution of the above from about 11 minutes to about 1
minute for me.
(Not counting download time, I'm testing with a populated cache.)

That's still a long time: and removing the awscli line so that we have only the
localstack requirement, we still take longer than I was willing to wait to get
to a solution...

These fixes all should be clear improvements; but I don't have a good sense for
whether the things that they address are relatively common, or rare.

### DIFF
--- a/src/poetry/mixology/version_solver.py
+++ b/src/poetry/mixology/version_solver.py
@@ -55,6 +55,7 @@ class VersionSolver:
         self._use_latest = use_latest
 
         self._incompatibilities: dict[str, list[Incompatibility]] = {}
+        self._contradicted_incompatibilities: set[Incompatibility] = set()
         self._solution = PartialSolution()
 
     @property
@@ -103,6 +104,9 @@ class VersionSolver:
             # we can derive stronger assignments sooner and more eagerly find
             # conflicts.
             for incompatibility in reversed(self._incompatibilities[package]):
+                if incompatibility in self._contradicted_incompatibilities:
+                    continue
+
                 result = self._propagate_incompatibility(incompatibility)
 
                 if result is _conflict:
@@ -149,6 +153,7 @@ class VersionSolver:
                 # If term is already contradicted by _solution, then
                 # incompatibility is contradicted as well and there's nothing new we
                 # can deduce from it.
+                self._contradicted_incompatibilities.add(incompatibility)
                 return None
             elif relation == SetRelation.OVERLAPPING:
                 # If more than one term is inconclusive, we can't deduce anything about
@@ -165,6 +170,8 @@ class VersionSolver:
         # incompatibility is satisfied and we have a conflict.
         if unsatisfied is None:
             return _conflict
+
+        self._contradicted_incompatibilities.add(incompatibility)
 
         adverb = "not " if unsatisfied.is_positive() else ""
         self._log(f"derived: {adverb}{unsatisfied.dependency}")
@@ -255,6 +262,7 @@ class VersionSolver:
                 or most_recent_satisfier.cause is None
             ):
                 self._solution.backtrack(previous_satisfier_level)
+                self._contradicted_incompatibilities.clear()
                 if new_incompatibility:
                     self._add_incompatibility(incompatibility)
 
--- a/src/poetry/mixology/version_solver.py
+++ b/src/poetry/mixology/version_solver.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
+import functools
 import time
 
 from contextlib import suppress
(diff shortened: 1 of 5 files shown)


src/poetry/mixology/version_solver.py
```diff
@@ -23,12 +24,43 @@
...
     from poetry.core.packages.project_package import ProjectPackage
 
+    from poetry.packages import DependencyPackage
     from poetry.puzzle.provider import Provider
 
 
 _conflict = object()
 
 
+class DependencyCache:
```
abn, line 34: Can/Should we move the logic into the provider itself so that the use of the cache is implicit?

src/poetry/mixology/version_solver.py
```diff
@@ -23,12 +24,43 @@
...
+    The key observation here is that during the search - except at backtracking
+    - once we have decided that a dependency is invalid, we never need check it
+    again.
+    """
+
+    def __init__(self, provider: Provider):
+        self.provider = provider
+        self.cache: dict[str, list[Package]] = {}
+
+    @functools.lru_cache(maxsize=128)
```
abn, line 47: Is there a real advantage/disadvantage of limiting the size here? ie. when do we start seeing issues?

abn (commented):
I like the general idea. Left some high level comments for now.

src/poetry/mixology/version_solver.py
dimbleby: I think that this cache is specialized to the solver: really its point is to do the slightly weird thing where it filters itself on access.  I wondered about calling it a `WinnowingCache` or something like that but it's an unusual word... better ideas welcome!

So I don't think that this cache should be moved to a place where it might be mistaken for a regular cache.

src/poetry/mixology/version_solver.py
dimbleby: I don't think it's a big deal either way.  

In the kind of edge case that this MR targets, we can get a large number of not-quite-the-same dependencies that differ only because they have different constraints, and I don't think that it's useful to keep the earlier versions of these dependencies cached.  But I don't suppose it does much harm either.

src/poetry/mixology/version_solver.py
abn: This is fine for now.

abn (approved):

mm-matthias:
I used 1.2.0b2 to check if this PR solves the cases mentioned in #5896 and the duplicated/similar tickets. While everything seems to run a bit faster poetry still tries to evaluate every single one of the awscli versions. Usually this means downloading every single package from pypi first which makes a single update run take hours, because there are thousands of awscli versions.
To make these cases significantly faster poetry has to avoid evaluating thousands of individual package versions in the first place. After all if I give a manual hint to the package version as outlined in #5896 there is actually a valid solution that is reached in a few seconds. It's just that the current solver does not seem smart enough to avoid evaluating solutions that involve a few thousand package versions.

dimbleby:
It is only hindsight that could possibly allow poetry to know that all of those thousands of awscli versions are not compatible with the current state of the search.  During the search, it has no choice but to check.

Regardless of any potential future optimisations, specifying a recent lower bound for awscli is always likely to be the best way to guide poetry in the right direction.  If you encounter packages that don't do this (and therefore trigger the painful case), do consider asking them if they wouldn't mind making the change eg as localstack did in https://github.com/localstack/localstack/pull/5912

Issue #4577, opened 2021-09-30
`poetry install --without` installs optional dependency groups


pmav99:
- [x] I am on the [latest](https://github.com/python-poetry/poetry/releases/latest) Poetry version.
- [x] I have searched the [issues](https://github.com/python-poetry/poetry/issues) of this repo and believe that this is not a duplicate.
- [x] If an exception occurs when executing a command, I executed it again in debug mode (`-vvv` option).

```
Poetry
Version: 1.2.0a2
Python:  3.9.7

Virtualenv
Python:         3.8.12
Implementation: CPython
Path:           /tmp/aaa/.venv
Valid:          True

System
Platform: linux
OS:       posix
Python:   /usr
```

This is somewhat similar to https://github.com/python-poetry/poetry/issues/4401 but that was about poetry 1.1 so I opened a different ticket.

## Issue

Let's say we have two optional groups, `test` and `foo`. `foo` does not contain any packages, but this is not important here. I am just keep it empty to speed up dependency resolution etc.

``` toml
[tool.poetry]
name = "aaa"
version = "0.1.0"
description = ""
authors = ["pmav99 <pmav99@mail.com>"]
readme = "README.md"

[tool.poetry.dependencies]
python = "^3.9"

[tool.poetry.group.test]
optional = true
[tool.poetry.group.foo]
optional = true

[tool.poetry.group.test.dependencies]
pytest= "*"

[tool.poetry.group.foo.dependencies]

[build-system]
requires = ["poetry-core"]
build-backend = "poetry.core.masonry.api"
```

Running `poetry update` followed by `poetry install` works great, i.e. `pytest` will not be installed since nobody asked to install the group `test`.:
```
poetry update --quiet
poetry install --quiet
python -mpytest --help || echo 'pytest not installed'
```

If we try to run `poetry install --with foo` then still everything works as expected. I.e. pytest remains non-installed. 

```
poetry install --with foo
python -mpytest --help || echo 'pytest not installed'
```

Nevertheless, when we try to run `poetry install --without foo`, then the dependencies of group `test` get installed too!

```
$ poetry install --without foo
Installing dependencies from lock file

Package operations: 8 installs, 0 updates, 0 removals

  • Installing pyparsing (2.4.7)
  • Installing attrs (21.2.0)
  • Installing iniconfig (1.1.1)
  • Installing packaging (21.0)
  • Installing pluggy (1.0.0)
  • Installing py (1.10.0)
  • Installing toml (0.10.2)
[...]

PR #5338, opened 2022-03-21
misc. improvements to group filtering logic and related options

abn:
- Fix `install --only` logic to no longer implicitly disable root package installation. (added test coverage)
- Dedupe logic for group aware command and improve "active" group determination.
- Reduce cognitive load for installer by removing `without` and `with` group specifications, replacing with `only` group logic.
- Improve test cases to more reliably verify installer logic and increase coverage.
- Increase coverage for `install` command with group filtering options.
- Improve documenration relating to dependency group to further highlight distinction between `default` group and `default dependencies`.
- Get documentation in sync with cli `--help` output and warning around combining `--only` with `--with | --without`.
- Deprecate the use of `--default` option to reduce CLI cognitive load on users.
- Mark `install --remove-untracked` as deprecated
- Improve test coverage for show command with groups.
- Translate legacy options correctly. `--no-dev` => `--only default`, `--only-dev` => `--without default` etc.  

Closes: #4361
Resolves: #4577

Discussion: https://github.com/orgs/python-poetry/teams/core/discussions/8

PS: Commits are separated by functional changes.

### DIFF
--- a/docs/cli.md
+++ b/docs/cli.md
@@ -215,10 +215,10 @@ option is used.
 
 ### Options
 
-* `--without`: The dependency groups to ignore for installation.
-* `--with`: The optional dependency groups to include for installation.
-* `--only`: The only dependency groups to install.
-* `--default`: Only install the default dependencies.
+* `--without`: The dependency groups to ignore.
+* `--with`: The optional dependency groups to include.
+* `--only`: The only dependency groups to include.
+* `--default`: Only include the default dependencies. (**Deprecated**)
 * `--sync`: Synchronize the environment with the locked packages and the specified groups.
 * `--no-root`: Do not install the root package (your project).
 * `--dry-run`: Output the operations but do not execute anything (implicitly enables --verbose).
@@ -227,6 +227,9 @@ option is used.
 * `--dev-only`: Only install dev dependencies. (**Deprecated**)
 * `--remove-untracked`: Remove dependencies not presented in the lock file. (**Deprecated**)
 
+{{% note %}}
+When `--only` is specified, `--with` and `--without` options are ignored.
+{{% /note %}}
 
 ## update
 
@@ -252,9 +255,17 @@ update the constraint, for example `^2.3`. You can do this using the `add` comma
 
 ### Options
 
+* `--without`: The dependency groups to ignore.
+* `--with`: The optional dependency groups to include.
+* `--only`: The only dependency groups to include.
+* `--default`: Only include the default dependencies. (**Deprecated**)
 * `--dry-run` : Outputs the operations but will not execute anything (implicitly enables --verbose).
-* `--no-dev` : Do not install dev dependencies.
+* `--no-dev` : Do not update the development dependencies. (**Deprecated**)
 * `--lock` : Do not perform install (only update the lockfile).
+
+{{% note %}}
+When `--only` is specified, `--with` and `--without` options are ignored.
+{{% /note %}}
 
 ## add
 
@@ -425,15 +436,18 @@ required by
 
 ### Options
 
-* `--without`: Do not show the information of the specified groups' dependencies.
-* `--with`: Show the information of the specified optional groups' dependencies as well.
-* `--only`: Only show the information of dependencies belonging to the specified groups.
-* `--default`: Only show the information of the default dependencies.
-* `--no-dev`: Do not list the dev dependencies.
+* `--without`: The dependency groups to ignore.
+* `--with`: The optional dependency groups to include.
+* `--only`: The only dependency groups to include.
+* `--default`: Only include the default dependencies. (**Deprecated**)
+* `--no-dev`: Do not list the dev dependencies. (**Deprecated**)
 * `--tree`: List the dependencies as a tree.
 * `--latest (-l)`: Show the latest version.
 * `--outdated (-o)`: Show the latest version but only for packages that are outdated.
 
+{{% note %}}
+When `--only` is specified, `--with` and `--without` options are ignored.
+{{% /note %}}
 
 ## build
 
--- a/docs/managing-dependencies.md
+++ b/docs/managing-dependencies.md
@@ -36,6 +36,7 @@ on whether their dependencies will be resolved and installed **by default**, the
 the dependencies logically.
 {{% /note %}}
 
+{{% note %}}
 The dependencies declared in `tool.poetry.dependencies` are part of an implicit `default` group.
 
 ```toml
@@ -47,6 +48,7 @@ pendulum = "*"
 pytest = "^6.0.0"
 pytest-mock = "*"
 ```
+{{% /note %}}
 
 {{% note %}}
 **A note about the `dev-dependencies` section**
@@ -109,7 +111,14 @@ If the group does not already exist, it will be created automatically.
 
 ### Installing group dependencies
 
-**By default**, dependencies across **all groups** will be installed when executing `poetry install`.
+**By default**, dependencies across **all non-optional groups** will be installed when executing
+`poetry install`.
+
+{{% note %}}
+The default set of dependencies for a project includes the implicit `default` group defined in
+`tool.poetry.dependencies` as well as all groups that are not explicitly marked as an
+[optional group]({{< relref "#optional-groups" >}}).
+{{% /note %}}
 
 You can **exclude** one or more groups with the `--without` option:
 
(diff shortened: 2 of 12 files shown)


abn:
This is a draft for now as it still requires some changes to documentation to clarify expected behaviour and add more test cases to ensure we do not regress.

src/poetry/installation/installer.py
```diff
@@ -281,18 +269,8 @@ def _do_install(self, local_repo: Repository) -> int:
...
-            elif self._without_groups:
-                # Default dependencies without selected groups
-                root = self._package.without_dependency_groups(self._without_groups)
-            else:
-                # Only selected groups
-                root = self._package.with_dependency_groups(
-                    self._only_groups, only=True
-                )
+        if self._groups is not None:
+            root = self._package.with_dependency_groups(self._groups, only=True)
```
radoering, line 273:
```suggestion
            root = self._package.with_dependency_groups(list(self._groups), only=True)
```

src/poetry/console/commands/group_command.py
```diff
@@ -0,0 +1,114 @@
...
+                    "<warning>The `<fg=yellow;options=bold>--dev-only</>` option is"
+                    " deprecated, use the `<fg=yellow;options=bold>--only dev</>`"
+                    " notation instead.</warning>"
+                )
+                groups["only"].add("dev")
+
+        if groups["only"] and (groups["with"] or groups["without"]):
+            self.line_error(
+                "<warning>The `<fg=yellow;options=bold>--with</>` and "
+                "`<fg=yellow;options=bold>--without</>` options are ignore when used"
```
radoering, line 102:
```suggestion
                "`<fg=yellow;options=bold>--without</>` options are ignored when used"
```

tests/console/commands/test_install.py
```diff
@@ -4,32 +4,112 @@
...
+        case + (with_root,)
+        for case in [
+            ("", {"default", "foo", "bar", "baz", "bim"}),
+            ("--only default", {"default"}),
+            ("--only foo", {"foo"}),
+            ("--only foo,bar", {"foo", "bar"}),
+            ("--only bam", {"bam"}),
+            ("--with bam", {"default", "foo", "bar", "baz", "bim", "bam"}),
+            ("--without foo,bar", {"default", "baz", "bim"}),
+            ("--with foo,bar --without baz --without bim --only bam", {"bam"}),
```
radoering, line 77: I'd like even more test cases:
`--default` (even if it is deprecated and will probably be removed before release)
`--with non-optional` (useless with)
`--with optional --without non-optional`
`--without default` (just in case even if it is no separate path in code)
`--without non-optional` (useless without)
`--with non-optional --without non-optional` (precedence, no effect)
`--with optional --without optional` (precedence, useless with)

Since you can combine `--with` and `--without`, it should be documented which one takes precedence. It's not yet docurmented anywhere, is it?

radoering (changes requested):

src/poetry/installation/installer.py
abn: Ack, I was hoping to make the type hint change in core to be `Iterable` than `list`. But we can change that later.

radoering: Actually, it is not just the type hint but the implementation. The method requires at least a `Container`. You can't do `group_name not in groups` if `groups` is only an `Iterable`.

tests/console/commands/test_install.py
```diff
@@ -4,32 +4,122 @@
...
+    tester: CommandTester,
+    mocker: MockerFixture,
 ):
     """
     Group options are passed properly to the installer.
     """
-    mocker.patch.object(tester.command.installer, "run", return_value=1)
+    mocker.patch.object(tester.command.installer, "run", return_value=0)
+    editable_builder_mock = mocker.patch(
+        "poetry.masonry.builders.EditableBuilder", side_effect=ModuleOrPackageNotFound()
```
radoering, line 104: `side_effect` does not seem to be required here. The test still succeeds without `side_effect`. Can it safely be removed?

docs/managing-dependencies.md
```diff
@@ -36,6 +36,7 @@ on whether their dependencies will be resolved and installed **by default**, the
 the dependencies logically.
 {{% /note %}}
 
+{{% note %}}
 The dependencies declared in `tool.poetry.dependencies` are part of an implicit `default` group.
```
radoering, line 40:
```suggestion
The dependencies declared in `tool.poetry.dependencies` are part of an implicit `default` group.
In other words, `tool.poetry.dependencies` is equivalent to `tool.poetry.group.default.dependencies`.
```

In the other issue (in poetry-core) we have come to the conclusion that it is sensible to mention the explicit default group.

radoering (changes requested):

tests/console/commands/test_show.py
```diff
@@ -101,6 +101,174 @@ def test_show_basic_with_installed_packages(
     assert tester.io.fetch_output() == expected
 
 
+def _configure_project_with_groups(poetry: Poetry, installed: Repository) -> None:
+    poetry.package.add_dependency(Factory.create_dependency("cachy", "^0.1.0"))
+
+    poetry.package.add_dependency_group(DependencyGroup("time", True))
```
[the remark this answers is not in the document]
abn: I agree with you, and I really think we should adopt a more concious practice of avoiding positional args where possible. I will update this call. But food for thought.

tests/console/commands/test_show.py
```diff
@@ -101,6 +101,174 @@ def test_show_basic_with_installed_packages(
...
+
+@pytest.mark.parametrize(
+    ("options", "expected"),
+    [
+        (
+            "",
+            """\
+cachy  0.1.0 Cachy package
+pytest 3.7.3 Pytest package
+""",
```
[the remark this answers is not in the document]
abn: No strong feelings. Will execute.

docs/managing-dependencies.md
radoering: Oh yes, I didn't consider that. As long as it can be misused, it's probably better not to mention it explicitly.

tests/console/commands/test_show.py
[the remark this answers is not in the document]
radoering: No strong feelings. I tend towards using it, but I see at least in that case the benefit in readability is marginal.

docs/cli.md
```diff
@@ -227,6 +227,9 @@ option is used.
 * `--dev-only`: Only install dev dependencies. (**Deprecated**)
```
finswimmer, line 227: AFAIR the `--dev-only` option was never contained within a release. So I think we can remove it now already or at least with final 1.2 release.

tests/console/commands/test_install.py
```diff
@@ -4,32 +4,122 @@
...
+            ("--with bam --without bam", {"default", "foo", "bar", "baz", "bim"}),
+            ("--with foo --without foo", {"default", "bar", "baz", "bim"}),
+            # deprecated options
+            ("--default", {"default"}),
+            ("--no-dev", {"default"}),
+            ("--dev-only", {"foo", "bar", "baz", "bim"}),
+        ]
+        for with_root in {True, False}
+    ],
+)
```
finswimmer, line 91:
```suggestion
@pytest.mark.parametrize(
    ["options", "groups"],
    [
        ("", {"default", "foo", "bar", "baz", "bim"}),
        ("--only default", {"default"}),
        ("--only foo", {"foo"}),
        ("--only foo,bar", {"foo", "bar"}),
        ("--only bam", {"bam"}),
        ("--with bam", {"default", "foo", "bar", "baz", "bim", "bam"}),
        ("--without foo,bar", {"default", "baz", "bim"}),
        ("--without default", {"foo", "bar", "baz", "bim"}),
        ("--with foo,bar --without baz --without bim --only bam", {"bam"}),
        # net result zero options
        ("--with foo", {"default", "foo", "bar", "baz", "bim"}),
        ("--without bam", {"default", "foo", "bar", "baz", "bim"}),
        ("--with bam --without bam", {"default", "foo", "bar", "baz", "bim"}),
        ("--with foo --without foo", {"default", "bar", "baz", "bim"}),
        # deprecated options
        ("--default", {"default"}),
        ("--no-dev", {"default"}),
        ("--dev-only", {"foo", "bar", "baz", "bim"}),
    ],
)
@pytest.mark.parametrize("with_root", [True, False])
```

tests/console/commands/test_install.py
```diff
@@ -4,32 +4,122 @@
...
+    if not with_root:
+        options = f"--no-root {options}"
+
+    tester.execute(options)
+
+    package_groups = set(tester.command.poetry.package._dependency_groups.keys())
+    installer_groups = set(tester.command.installer._groups)
 
-    tester.execute("--with foo,bar --without baz --without bim --only bam")
+    assert installer_groups <= package_groups
```
finswimmer, line 115: Shouldn't it be `len(installer_groups) <= len(package_groups)`?

finswimmer (changes requested):

tests/console/commands/test_install.py
abn: Why? If you do len you won't be verifying the values, just the count. Current assertion is that the installed groups are a subset of the project groups.

docs/cli.md
abn: I think you are correct, however I'd prefer to do the removal in a separate PR, ideally after the next pre-release prior to 1.2 along with default.

tests/console/commands/test_install.py
finswimmer: I simply wasn't aware of this syntax. Maybe use `issubset()` instead?

abn: I don't feel that is required per say, let me know if you feel strongly about it.

radoering (approved):

finswimmer (approved):

PR #5258, opened 2022-03-02
chore(console): remove type errors

eyllanesc:

PR #5368, opened 2022-03-28
solved mypy errors for poetry.console.application 

tarun-jethwani:
# Pull Request Check List

Resolves: #5017 

- introduced type hinting for self._poetry ( line number 96 ) and loggers ( line number 224) in https://github.com/tarun-jethwani/poetry/blob/fix_mypy_issues_for_poetry_console_applications/src/poetry/console/application.py
- code change does not require adding tests and documentation update

### DIFF
--- a/src/poetry/console/application.py
+++ b/src/poetry/console/application.py
@@ -93,7 +93,7 @@ class Application(BaseApplication):
     def __init__(self) -> None:
         super().__init__("poetry", __version__)
 
-        self._poetry = None
+        self._poetry: Poetry | None = None
         self._io: IO | None = None
         self._disable_plugins = False
         self._plugins_loaded = False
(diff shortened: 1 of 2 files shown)


src/poetry/console/application.py
```diff
@@ -221,7 +221,7 @@ def register_command_loggers(
 
         io = event.io
 
-        loggers = [
+        loggers: list[Any] = [
```
branchvincent, line 224: Although this silences `mypy`, `loggers` is really only a `list[str]`. The actual issue stems from re-using the same variable name for different types. So a more direct solution would be to avoid it on line 235:

```python
for logger_name in loggers:
    logger = logging.getLogger(logger_name)
```

branchvincent (changes requested):
thanks a lot for your contribution 👍 , I just have one comment

tarun-jethwani:
@branchvincent thank you so much for your feedback, I understood your changes and reason behind that but I recently found out that someone is already working on this module for mypy errors there is an open PR for that 

Meanwhile I have taken up modules inside  poetry.mixology will solve mypy errors for that and sumit a New PR

Please stay updated on this issue  to review my Next PR

branchvincent:
sounds good @tarun-jethwani, I'll close this pr in favor of #5258

Issue #4439, opened 2021-08-25
Dependency resolution differences (wrong) when using custom (i.e. not pypi) repository


Korijn:
- [x] I am on the [latest](https://github.com/python-poetry/poetry/releases/latest) Poetry version.
- [x] I have searched the [issues](https://github.com/python-poetry/poetry/issues) of this repo and believe that this is not a duplicate.
- [x] If an exception occurs when executing a command, I executed it again in debug mode (`-vvv` option).

- **OS version and name**: Windows 10 and Ubuntu 20.04
- **Poetry version**: 1.1.8
- **Link of a [Gist](https://gist.github.com/) with the contents of your pyproject.toml file**: not needed to explain

## Issue

Any source that is defined in pyproject.toml that is not pypi, is [always handled internally as a LegacyRepository](https://github.com/python-poetry/poetry/blob/a3c5cf7c9559dbb0b0372aa2e30b5ce3f2a849af/poetry/factory.py#L140).

That means metadata is not collected from API calls, but always by downloading and parsing packages, usually sdists.

I probably don't have to explain how this is bad for performance in terms of speed, but you can see people notice it because it is quite significant! See for example #4113 

Sometimes however, in cases where package metadata _would have been available on an API endpoint_, but poetry can't figure out what the metadata is by parsing the sdist, this leads to problems in dependency resolution.

For example, scikit-image 0.17.2 sdist imports numpy in its setup.py, but it doesn't specify any build requirements in pyproject.toml, so running setup.py fails. Poetry then just silently concludes scikit-image doesn't have any dependencies, which is clearly wrong.

This is exactly what happens in #3464 and is also how I first encountered this bug.

If you install this package from pypi however, everything goes smoothly because the metadata is collected from the API endpoint instead.

So in short, for the exact same dependencies, depending on what source repository you use: pypi or something else, you may not get the same dependency resolution. Even if the alternative source is a direct reverse proxy to pypi.

### Option 1 - Fully automated

This is the ideal option. Poetry becomes clever enough to figure out for any source if it can provide metadata via an API just like pypi can. A mechanism needs to be built that tests this per configured source.

You could look at hostnames to try and optimize this guessing game a little bit.

### Option 2 - User configurable

Allow users to configure the capabilities a source has available in pyproject.toml. This would basically put the responsibility with the user to tell poetry what APIs can be consumed.

```toml
[[tool.poetry.source]]
name = "foo"
url = "https://foo.bar/simple/"
[...]

Issue #3464, opened 2020-12-08
Poetry cannot resolve scikit-image dependencies from a private PyPi server


orf:
- [x] I am on the [latest](https://github.com/python-poetry/poetry/releases/latest) Poetry version.
- [x] I have searched the [issues](https://github.com/python-poetry/poetry/issues) of this repo and believe that this is not a duplicate.
- [x] If an exception occurs when executing a command, I executed it again in debug mode (`-vvv` option).

- **OS version and name**: MacOS 11
- **Poetry version**: 1.1.4

## Issue

Given a project that depends on `scikit-image`, and using an external (private) repository that does not implement PyPi's JSON api, running `poetry debug resolve scikit-image` resolves `scikit-image` to have _no_ dependencies (output at the end). However this is incorrect, and the console output clearly shows there is an exception while trying to invoke the PEP517 module.

Unfortunately this manifests itself as a `poetry lock` working just fine, but incorrectly specifying that there are no dependencies and thus `scikit-image` fails to import.

I've tried various steps to debug this:
1. Creating different venvs with different python versions
2. Explicitly setting the `scikit-image` dependency with `source="pypi"`
3. Trying to download the source package and invoking `pep517` manually to see what happens
4. Upgrading `pip`, `wheel` and `setuptools` everywhere I can. Side note: I cannot see where `WARNING: You are using pip version 20.2.3; however, version 20.3.1 is available.` is coming from 😱 

I'm willing to help create a fix for this, but I need some pointers as I'm in the dark about what could be going wrong here. 

The output:

```
❯ poetry debug resolve scikit-image -vvv
Resolving dependencies...
   1: fact: document-model-gateway is 1.0.0
   1: derived: document-model-gateway
   1: fact: document-model-gateway depends on scikit-image (^0.17.2)
   1: selecting document-model-gateway (1.0.0)
   1: derived: scikit-image (^0.17.2)
onfido: 1 packages found for scikit-image >=0.17.2,<0.18.0
PyPI: No release information found for scikit-image-0.12.dev0, skipping
PyPI: No release information found for scikit-image-0.9.2, skipping
PyPI: 1 packages found for scikit-image >=0.17.2,<0.18.0
onfido: Downloading sdist: scikit-image-0.17.2.tar.gz
[...]

PR #5385, opened 2022-03-30
Prefer compatible wheels over source distribution for legacy repository dependency resolution

gpfv:
# Pull Request Check List

Resolves: #4439 and #3464

- [x] Updated **documentation** for changed code. (not needed)

PyPi repositories use the pypi.org JSON api for fetching dependencies, where for an external repo (legacy repo) the dependencies are determined by going through the files available in the repo and inspecting the metadata (actually PyPi repositories also use this method as fallback when requires_dist is None). Universal wheels are prefered, then source distribution and only after that a desperate attempt (# Pick the first wheel available and hope for the best) is made to extract the dependencies from a platform specific wheel.

It is cumbersome to retrieve this from source, the file needs to be extracted, which takes a long time, and even after it is not always certain the correct ones can be determined, due to  lack of standardization and arbitrary code execution in setup.py. So either this step has to be improved, implying all past and future build systems need to be supported, or another option is to use a well defined package distribution format, which already exist. The wheel does not need to be reinvented, it just needs to be used.

As dependencies could potentially differ (although could not find an example of this easily) depending on the platform and python version, only the appropriate wheels are considered by matching with compatibility tags. 

-> Using compatible wheels over sdist will result in a more correct dependency resolution at a greater speed.

for example PyQt5 (5.15.4) requires_dist
 * pypi  ['PyQt5-sip (>=12.8, <13)', 'PyQt5-Qt5 (>=5.15)'] 
 * bdist ['PyQt5-sip (>=12.8, <13)', 'PyQt5-Qt5 (>=5.15)'] 
 * sdist None

wheel : ~3s not cached, ~0.2s cached
tar.gz: ~10s both not cached and cached (needs to be extracted everytime)

### DIFF
--- a/src/poetry/repositories/http.py
+++ b/src/poetry/repositories/http.py
@@ -7,6 +7,7 @@ import urllib
 
 from abc import ABC
 from collections import defaultdict
+from packaging.tags import sys_tags, Tag
 from pathlib import Path
 from typing import TYPE_CHECKING
 from urllib.parse import quote
@@ -143,7 +144,7 @@ class HTTPRepository(CachedRepository, ABC):
             universal_wheel = None
             universal_python2_wheel = None
             universal_python3_wheel = None
-            platform_specific_wheels = []
+            platform_specific_wheels = {}
             for wheel in wheels:
                 link = Link(wheel)
                 m = wheel_file_re.match(link.filename)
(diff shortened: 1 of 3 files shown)


abn:
@gpfv this might need some TLC. :)

gpfv:
Testing Linux Capabilities added
will now only mock on Windows the two platforms win32 and win-amd64

neersighted (commented):
LGTM after a rebase.

neersighted:
Superseded by #6547

PR #5392, opened 2022-03-31
Fix format string for url cleaning of other sources

JanLuca:
In 5814e335221047bbd5ca4d7def6994acc0a9c1ca a regression has been introduced that new code using f-strings omitted the call of `ord()` to clean the url of other sources included.

At the same time I observed that the format string option to add an leading zero for numbers < 10 is missing. It may be unlikely to happen in the real world but without the flag the encoding of the ASCII symbols 0-9 would have a space in the output (e,g,"\0" -> "% 0" and not "%00" as expected).


radoering:
Good catch. 👍 Can you please add a small unit test for `clean_link()`?

JanLuca:
I added a small test which should catch such regressions in the future :)

radoering (approved):

PR #5401, opened 2022-04-02
build.script: use build environment for execution

abn:
With this change, Poetry now creates an ephemeral build environment
with all requirements specified under `build-system.requires` when a
build script is specified. Otherwise, project environment is reused.

Resolves: #2789
Supercedes: python-poetry/poetry-core#58

Decided to solve this at the poetry level as we do not yet have consensus on if the build-system requirements should be considered as a project's development dependendency.

**Alternative approaches**
- `build-system.requires` is loaded into a new implicit dependency group callded `build-system`, this will mean that these will affect dependency resolution. However, this will also allow for the easier development of build scripts.

**Future Enhancements**
- The build environment is cached within the project's virtual environment. This will require Poetry to handle entire life-cycle if the dependencies change etc. This might also cause issues with system environments. However, this will also provide speedups.
- Reuse Poetry's internal wheel fetch logic instead of relying on `pip`.


yajo:
This doesn't solve the problem of `buidl-system.requires` not being locked in `poetry.lock`, does it?

IMHO they should be locked there too, or we can't really have reproducible builds.

abn:
@yajo unfortunately this PR is not intended to solve that.

Additionally, there is also the added issue that the build time dependency is managed by the PEP 517 frontends. The short version is, even if the lockfile contained your build system depdnencies your builds will not be reproducible across all cases. The only case it covers if you used `poetry build`. The only way to achieve reproducible builds is to lock the version in your `pyproject.toml` `build-system.requires`. Obviously, this will have to be manually managed.

src/poetry/utils/env.py
```diff
@@ -50,7 +52,8 @@
     from cleo.io.io import IO
     from poetry.core.version.markers import BaseMarker
 
-    from poetry.poetry import Poetry
+
+P = TypeVar("P", bound=Poetry)
```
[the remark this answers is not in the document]
abn: Poetry instance can come from core or from Poetry itself, hence type var.

src/poetry/console/commands/build.py
```diff
@@ -23,11 +30,13 @@ class BuildCommand(EnvCommand):
     def handle(self) -> None:
         from poetry.core.masonry.builder import Builder
 
-        fmt = self.option("format") or "all"
-        package = self.poetry.package
-        self.line(
-            f"Building <c1>{package.pretty_name}</c1> (<c2>{package.version}</c2>)"
-        )
+        env: Env
```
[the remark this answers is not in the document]
abn: It was complaining when I tried it last, I should do that again to see if anything has changed since the last time with all the other changes going on.

src/poetry/utils/env.py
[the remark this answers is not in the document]
neersighted: Ah, that makes sense

src/poetry/utils/env.py
```diff
@@ -1839,6 +1842,46 @@ def ephemeral_environment(
...
+) -> Iterator[Env]:
+    """
+    If a build script is specified for the project, there could be additional build
+    time dependencies, eg: cython, setuptools etc. In these cases, we create an
+    ephemeral build environment with all requirements specified under
+    `build-system.requires` and return this. Otherwise, the given default project
+    environment is returned.
+    """
+    if not env or poetry.package.build_script:
+        with ephemeral_environment(executable=env.python if env else None) as venv:
```
neersighted, line 1857: Pretty sure this `executable=` will do nothing as you just checked `not env`.

src/poetry/utils/env.py
```diff
@@ -1839,6 +1842,46 @@ def ephemeral_environment(
...
+            overwrite = io and io.output.is_decorated() and not io.is_debug()
+            if io:
+                requires = map(
```
neersighted, line 1860: Let's use a list comprehension here -- it's more readable and more performant

src/poetry/utils/env.py
```diff
@@ -1839,6 +1842,46 @@ def ephemeral_environment(
...
+                    lambda r: f"<c1>{r}</c1>", poetry.pyproject.build_system.requires
+                )
+                if not overwrite:
+                    io.write_line("")
+
+                io.overwrite(
+                    "<b>Preparing</b> build environment with build-system requirements"
```
neersighted, line 1867:
```suggestion
                    "<b>Preparing</b> build environment with build-system requirements: "
```

neersighted (changes requested):

src/poetry/utils/env.py
abn: There can be a case where env is none and a build script exists.

src/poetry/utils/env.py
abn: THere is a whitespace in the next line and I prefer it to be  "build-system requirements a, b" instead of "build-system requirements: a, b" as it is more a sentence.

src/poetry/utils/env.py
neersighted: Ah, yeah, I read the `or` as a `and` -- oops.

neersighted (approved):

mattpeople:
note for future searchers: poetry 1.2.0 broke our previously successful builds (where we use a custom `build = 'script.py'` setting) until we added `setuptools` manually as:

```ini
[build-system]
requires = ["poetry-core", "setuptools"]
```

PR #5402, opened 2022-04-03
solved mypy errors for poetry.mixology

tarun-jethwani:
# Pull Request Check List

Resolves: #5017

This PR fixes mypy errors for entire **poetry.mixology** package and remove those corresponding fixed modules from _[[tool.mypy.overrides]]_ section in _pyproject.toml_ file

### DIFF
--- a/src/poetry/mixology/failure.py
+++ b/src/poetry/mixology/failure.py
@@ -115,7 +115,9 @@ class _Writer:
         incompatibility_string = str(incompatibility)
 
         cause = incompatibility.cause
-        details_for_cause = {}
+        # using assert to suppress mypy [union-attr]
+        assert isinstance(cause, ConflictCause)
+        details_for_cause: dict = {}
         if isinstance(cause.conflict.cause, ConflictCause) and isinstance(
             cause.other.cause, ConflictCause
         ):
@@ -135,7 +137,7 @@ class _Writer:
                 if conflict_line is not None:
                     with_line = cause.conflict
                     without_line = cause.other
-                    line = conflict_line
+                    line: int | None = conflict_line
                 else:
                     with_line = cause.other
                     without_line = cause.conflict
@@ -200,7 +202,9 @@ class _Writer:
                     numbered=numbered,
                 )
             elif self._is_collapsible(derived):
-                derived_cause: ConflictCause = derived.cause
+                derived_cause = derived.cause
+                # using assert to suppress mypy [union-attr]
+                assert isinstance(derived_cause, ConflictCause)
                 if isinstance(derived_cause.conflict.cause, ConflictCause):
                     collapsed_derived = derived_cause.conflict
                     collapsed_ext = derived_cause.other
@@ -239,7 +243,9 @@ class _Writer:
         if self._derivations[incompatibility] > 1:
             return False
 
-        cause: ConflictCause = incompatibility.cause
+        cause = incompatibility.cause
+        # using assert to suppress mypy [union-attr]
+        assert isinstance(cause, ConflictCause)
         if isinstance(cause.conflict.cause, ConflictCause) and isinstance(
             cause.other.cause, ConflictCause
         ):
--- a/src/poetry/mixology/incompatibility.py
+++ b/src/poetry/mixology/incompatibility.py
@@ -50,15 +50,15 @@ class Incompatibility:
                 ref = term.dependency.complete_name
 
                 if ref in by_ref:
-                    by_ref[ref] = by_ref[ref].intersect(term)
-
+                    intersect = by_ref[ref].intersect(term)
                     # If we have two terms that refer to the same package but have a
                     # null intersection, they're mutually exclusive, making this
                     # incompatibility irrelevant, since we already know that mutually
                     # exclusive version ranges are incompatible. We should never derive
                     # an irrelevant incompatibility.
                     err_msg = f"Package '{ref}' is listed as a dependency of itself."
-                    assert by_ref[ref] is not None, err_msg
+                    assert intersect is not None, err_msg
+                    by_ref[ref] = intersect
                 else:
                     by_ref[ref] = term
 
@@ -83,23 +83,13 @@ class Incompatibility:
         return self._terms
 
     @property
-    def cause(
-        self,
-    ) -> (
-        RootCause
-        | NoVersionsCause
-        | DependencyCause
-        | ConflictCause
-        | PythonCause
-        | PlatformCause
-        | PackageNotFoundCause
-    ):
+    def cause(self) -> IncompatibilityCause:
         return self._cause
 
     @property
     def external_incompatibilities(
         self,
-    ) -> Iterator[ConflictCause | Incompatibility]:
+    ) -> Iterator[Incompatibility]:
         """
         Returns all external incompatibilities in this incompatibility's
         derivation graph.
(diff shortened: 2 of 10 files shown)


src/poetry/mixology/failure.py
```diff
@@ -135,7 +137,7 @@ def _visit(
                 if conflict_line is not None:
                     with_line = cause.conflict
                     without_line = cause.other
-                    line = conflict_line
+                    line: int | None = conflict_line
```
dimbleby, line 140: `line` is never `None` in either branch

src/poetry/mixology/incompatibility.py
```diff
@@ -83,23 +83,13 @@ def terms(self) -> list[Term]:
...
-    ) -> (
-        RootCause
-        | NoVersionsCause
-        | DependencyCause
-        | ConflictCause
-        | PythonCause
-        | PlatformCause
-        | PackageNotFoundCause
-    ):
+    def cause(self) -> IncompatibilityCause:
```
dimbleby, line 86: nice

src/poetry/mixology/partial_solution.py
```diff
@@ -36,15 +36,15 @@ def __init__(self) -> None:
         # negative Assignments that refer to that package.
         #
         # This is derived from self._assignments.
-        self._positive: dict[str, Term] = {}
+        self._positive: dict[str, Term | None] = {}
```
dimbleby, line 39: this - and the same on `_negative` - look wrong, I really don't think these should be `None`.

cf the first commit in unmerged #5335: https://github.com/python-poetry/poetry/pull/5335/commits/72fed082d0c1952d302c2b11b6b9315f47241e09

suggest that the negative case should be similar, something like:
```python
term = assignment if old_negative is None else assignment.intersect(old_negative)
assert term is not None
```

tarun-jethwani: Hi @dimbleby , thanks your review and feedback,  the above changes worked for old_negative, made similar changes for 

>         if old_positive is not None:
>             term = old_positive.intersect(assignment)
>             assert term is not None
>             self._positive[name] = term
>             return

but unit test is failing at assertion error _**tests\console\commands/test_show.py:579 test_show_outdated_has_prerelease_and_allowed**_   

the reason I made self._positive: dict[str, Term | None] = {} because later at some point return value of old_positive.intersect(assignment) can be None, please checkout the definition of intersect , which was returning None, in last else branch , 

**that is the reason to avoid  Incompatible types in assignment (expression has type "Optional[Term]", target has type "Term")**  

I made the **self._positive: dict[str, Term | None] = {}**

It is working fine in case of self._negative: dict[str, dict[str, Term]] = {} with above mentioned fix, and No unit test is failing but I think for the sake of symmetry both should be Term|None,

dimbleby: I showed how to fix the unit test in the commit that I linked (justification is in the MR comment)

tarun-jethwani: yes, thank you, I understood , thanks for your suggestion

tarun-jethwani: Hi @dimbleby , I have completed the changes as expected

tarun-jethwani:
resolved merge conflicts

src/poetry/mixology/failure.py
```diff
@@ -200,7 +204,9 @@ def _visit(
                     numbered=numbered,
                 )
             elif self._is_collapsible(derived):
-                derived_cause: ConflictCause = derived.cause
+                derived_cause = derived.cause
+                # using assert to suppress mypy [union-attr]
+                assert isinstance(derived_cause, ConflictCause)
```
abn, line 209:
```suggestion
                derived_cause: ConflictCause = cast(derived.cause)
```

abn: Similar in other places.

src/poetry/mixology/incompatibility.py
```diff
@@ -50,15 +50,16 @@ def __init__(self, terms: list[Term], cause: IncompatibilityCause) -> None:
                 ref = term.dependency.complete_name
 
                 if ref in by_ref:
-                    by_ref[ref] = by_ref[ref].intersect(term)
+                    value = by_ref[ref].intersect(term)
```
abn, line 53: Is this required?

src/poetry/mixology/incompatibility.py
```diff
@@ -134,18 +125,18 @@ def __str__(self) -> str:
             assert len(self._terms) == 1
             assert self._terms[0].is_positive()
 
-            cause: PythonCause = self._cause
+            python_cause = self._cause
```
abn, line 128:
```suggestion
            python_cause: PythonCause = cast(PythonCause, self._cause)
```

Similar further down.

src/poetry/mixology/incompatibility.py
```diff
@@ -385,7 +380,11 @@ def _try_requires_through(
         return "".join(buffer)
 
     def _try_requires_forbidden(
-        self, other: Incompatibility, details: dict, this_line: int, other_line: int
+        self,
+        other: Incompatibility,
+        details: dict,
```
abn, line 385: Can we get more specific here?

src/poetry/mixology/incompatibility.py
```diff
@@ -134,18 +125,18 @@ def __str__(self) -> str:
             assert len(self._terms) == 1
             assert self._terms[0].is_positive()
 
-            cause: PythonCause = self._cause
+            python_cause = self._cause
             text = f"{self._terse(self._terms[0], allow_every=True)} requires "
-            text += f"Python {cause.python_version}"
+            text += f"Python {python_cause.python_version}"
```
abn, line 130: Alternatively.

```suggestion
            text += f"Python {cast(PythonCause, self._cause).python_version}"
```

abn (changes requested):
Good work @tarun-jethwani few minor suggestions/questions.

src/poetry/mixology/incompatibility.py
tarun-jethwani: hi @abn , looked at in details, specific in terms of types is not possible here, the arguments take up the type of the arguments passed by the calling functions, tried to match the types from there 
src/poetry/mixology/incompatibility.py <line number 216>
   def and_to_string(
        self,
        other: Incompatibility,
        details: dict,
        this_line: int | None,
        other_line: int | None,
    ) -> str:

and details:dict is not even used its redundant

src/poetry/mixology/incompatibility.py
tarun-jethwani: Hi @abn , this was giving->  **mypy error: Redundant cast to "PythonCause"  [redundant-cast]:** 

but this worked instead,  actually the starting of the code block itself starts with 
        **elif isinstance(self._cause, PythonCause):**
            assert len(self._terms) == 1
            assert self._terms[0].is_positive()

            text = f"{self._terse(self._terms[0], allow_every=True)} requires "
            **text += f"Python {self._cause.python_version}"**

so no need for assigning it to a different variable or casting it explicitly

src/poetry/mixology/incompatibility.py
tarun-jethwani: yes, actually **_by_ref_ has type  _dict[str, Term]_**, but **_by_ref[ref].intersect(term)_** can optionally return **"None"**  has return value, thats the reason to assign optional "None" value inside another variable(**value**) asserting it for not "None" instead of directly storing it in by_ref otherwise give _mypy error expression is type optional[Term] i.e can be **Term | None** and variable has type **Term**_

src/poetry/mixology/failure.py
tarun-jethwani: done

src/poetry/mixology/incompatibility.py
tarun-jethwani: directly using self._cause worked, no need to assign and create a new variable "python_cause"

tarun-jethwani: details given below

tarun-jethwani:
Hi @abn , thanks for your comments and review, I have resolved and taken care of your suggestions and questions, 
please verify the requested changes

src/poetry/mixology/incompatibility.py
branchvincent: since `dict` was already here, I'm fine with any improvements as a follow up

branchvincent (approved):
thanks @tarun-jethwani, this looks great to me. I'll wait a bit before merging to see if @abn has any other comments

abn:
Thanks for handling this @branchvincent. @tarun-jethwani thank you for your patience, I think this looks good - will merge it shortly.

abn (approved):

Issue #5378, opened 2022-03-29
Solver fails when dependency has multiple (constrained) dependency definitions for same package


NicolasT:
<!-- Checked checkbox should look like this: [x] -->
- [x] I am on the [latest](https://github.com/python-poetry/poetry/releases/latest) Poetry version.
- [x] I have searched the [issues](https://github.com/python-poetry/poetry/issues) of this repo and believe that this is not a duplicate.
- [x] If an exception occurs when executing a command, I executed it again in debug mode (`-vvv` option).

<!--
  Once those are done, if you're able to fill in the following list with your information,
  it'd be very helpful to whoever handles the issue.
-->

- **OS version and name**: Fedora 35
- **Poetry version**: 1.1.8 (on host, installed from Fedora repo/RPM) and 1.1.13 (in container to test with latest version, installed via `pip`)
- **Link of a [Gist](https://gist.github.com/) with the contents of your pyproject.toml file**: <!-- Gist Link Here -->

<details>
<summary>Just including the file (which is the output of a simple `poetry init` accepting all defaults):</summary>

```toml
[tool.poetry]
name = "test"
version = "0.1.0"
description = ""
authors = ["Your Name <you@example.com>"]

[tool.poetry.dependencies]
python = "^3.10"

[tool.poetry.dev-dependencies]

[build-system]
requires = ["poetry-core>=1.0.0"]
build-backend = "poetry.core.masonry.api"
```
</details>

## Issue
<!-- Now feel free to write your issue, but please be descriptive! Thanks again 🙌 ❤️ -->
When trying to `poetry add psychopy`, which currently (rightfully) selects version 2022.1.2, the solver bails out:
```
Using version ^2022.1.2 for PsychoPy                                                                                                                                                                                                          
                                                                                                                                                                                                                                              
Updating dependencies                                                                                                                                                                                                                         
Resolving dependencies... (2.4s)                                                                                                                                                                                                              
                                                                                                                                                                                                                                              
  SolverProblemError                                                                                                                                                                                                                          
                                                                                                                                                                                                                                              
  Because psychopy (2022.1.2) depends on both python-vlc (>=3.0.12118) and python-vlc (3.0.11115), psychopy is forbidden.                                                                                                                     
  So, because no versions of psychopy match >2022.1.2,<2023.0.0                                                                                                                                                                               
   and test depends on PsychoPy (^2022.1.2), version solving failed.                                                                                                                                                                          
                                                                                                                                                                                                                                              
[...]

PR #5403, opened 2022-04-03
provider: merge duplicate dependencies by marker (avoid unnecessary solver failures)

radoering:
# Pull Request Check List

Resolves: #5378

### DIFF
--- a/tests/puzzle/test_solver.py
+++ b/tests/puzzle/test_solver.py
@@ -1300,6 +1300,43 @@ So, because no versions of a match !=1.0
     assert str(e.value) == expected
 
 
+def test_solver_duplicate_dependencies_different_constraints_merge_by_marker(
+    solver: Solver, repo: Repository, package: Package
+):
+    package.add_dependency(Factory.create_dependency("A", "*"))
+
+    package_a = get_package("A", "1.0")
+    package_a.add_dependency(
+        Factory.create_dependency("B", {"version": "^1.0", "python": "<3.4"})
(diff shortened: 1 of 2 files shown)


tests/puzzle/test_solver.py
```diff
@@ -1300,6 +1300,43 @@ def test_solver_duplicate_dependencies_different_constraints_same_requirements(
...
+    repo.add_package(package_b11)
+    repo.add_package(package_b20)
+
+    transaction = solver.solve()
+
+    check_solver_result(
+        transaction,
+        [
+            {"job": "install", "package": package_b10},
+            {"job": "install", "package": package_b20},
```
abn, line 1334: I am guesing here one is skipped? We should verify this correcty.

radoering: I don't think so. `b10` is for `python < "3.4"` and `b20` is for `python >= "3.4"`. I thought `skip` is only relevant if something has already been installed?

radoering: I checked: `skip` is `False` for both. Actually, it seems that is specified implicitly: https://github.com/python-poetry/poetry/blob/fb13b3a676f476177f7937ffa480ee5cff9a90a5/tests/puzzle/test_solver.py#L99-L100

The test is inspired by `test_solver_duplicate_dependencies_different_constraints` where both packages are not skipped, too.

abn (approved):

PR #5412, opened 2022-04-05
plugins: cleanup and rename type to group

abn:
Minor improvements to plugin implementation to make downstream implementations a bit easier for application plugins. The main change here is the renaming of `type` property to `group`. This reflects reality more as this is referring to the [entrypoint group](https://packaging.python.org/en/latest/specifications/entry-points/#data-model) the plugin is made available under and avoids shadowing built-in `type`.

### DIFF
--- a/src/poetry/plugins/application_plugin.py
+++ b/src/poetry/plugins/application_plugin.py
@@ -7,14 +7,20 @@ from poetry.plugins.base_plugin import BasePlugin
 
 if TYPE_CHECKING:
     from poetry.console.application import Application
+    from poetry.console.commands.command import Command
 
 
 class ApplicationPlugin(BasePlugin):
     """
-    Base class for plugins.
+    Base class for application plugins.
     """
 
-    type = "application.plugin"
+    group = "poetry.application.plugin"
+
+    @property
+    def commands(self) -> list[type[Command]]:
+        return []
 
     def activate(self, application: Application) -> None:
-        raise NotImplementedError()
+        for command in self.commands:
+            application.command_loader.register_factory(command.name, lambda: command())
--- a/src/poetry/console/application.py
+++ b/src/poetry/console/application.py
@@ -313,9 +313,10 @@ class Application(BaseApplication):
         self._disable_plugins = io.input.has_parameter_option("--no-plugins")
 
         if not self._disable_plugins:
+            from poetry.plugins.application_plugin import ApplicationPlugin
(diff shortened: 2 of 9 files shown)


src/poetry/plugins/application_plugin.py
```diff
@@ -7,14 +7,20 @@
...
     """
-    Base class for plugins.
+    Base class for application plugins.
     """
 
-    type = "application.plugin"
+    group = "poetry.application.plugin"
+
+    @property
+    def commands(self) -> list[type[Command]]:
```
radoering, line 21: Should the `AddCommandPlugin` in `test_application` return something in `commands()`?

radoering (commented):
`test_plugin_manager.py` still uses `"plugin"` and `type` instead of `group`. Maybe, this should be changed, too.

radoering (approved):

lovesegfault:
I'm not sure whether this PR is to blame, but if I do `poetry plugin show` I get this:

```console
$ poetry plugin show   

  KeyError

  'export'

  at ~/.local/share/pypoetry/venv/lib/python3.8/site-packages/poetry/console/commands/plugin/show.py:55 in handle
       51│             category = "plugins"
       52│             if issubclass(plugin, ApplicationPlugin):
       53│                 category = "application_plugins"
       54│ 
    →  55│             package = packages_by_name[canonicalize_name(entry_point.name)]
       56│             plugins[package.pretty_name]["package"] = package
       57│             plugins[package.pretty_name][category].append(entry_point)
       58│ 
       59│         for name, info in plugins.items():
```

This is _after_ installing the export plugin.

abn:
@lovesegfault it could be. But I'm not sure. I'd recommend that you consider waiting till #5413 is merged. Also note that the export plugin is now also renamed. The old one will only work with poetry `<=1.2.0b1`.

I do expect another PR after that to look at the commands. If it remains an issue with the above PR, please create a new issue.

abn:
@lovesegfault had a quick look, and no that issue is not because of this PR. The root cause here is that poetry currently expects the entrypoint name to be the same as the package name. Not entirely sure if this was intentional. Since the export plugin is not named `export` the current versions of the plugin wont work with the show command.

https://github.com/python-poetry/poetry-plugin-export/blob/ddf3dde33cf3c5d1a0e4c53723d677014b38809e/pyproject.toml#L27-L28

Issue #5311, opened 2022-03-14
Unable to `poetry install` twisted trunk and treq


DMRobertson:
- [ ] I am on the [latest](https://github.com/python-poetry/poetry/releases/latest) Poetry version.
- [x] I have searched the [issues](https://github.com/python-poetry/poetry/issues) of this repo and believe that this is not a duplicate.
- [x] If an exception occurs when executing a command, I executed it again in debug mode (`-vvv` option).

- **OS version and name**: Fedora 35
- **Poetry version**: 1.1.12
- Pyproject.toml:
  ```toml
  [tool.poetry]
  name = "test"
  version = "0.1.0"
  description = ""
  authors = ["David Robertson <david.m.robertson1@gmail.com>"]
  
  [tool.poetry.dependencies]
  python = "^3.10"
  Twisted = {git = "https://github.com/twisted/twisted.git", branch = "trunk"}
  treq = "^22.2.0"
  
  [tool.poetry.dev-dependencies]
  
  [build-system]
  requires = ["poetry-core>=1.0.0"]
  build-backend = "poetry.core.masonry.api"
  ```

## Issue
I tried to install twisted trunk and treq as follows:

- `poetry init`
- `poetry add git+https://github.com/twisted/twisted.git#trunk`
- `poetry add treq`

Step three tried to install `treq ^22.2.0` which is the newest version. This failed: see [add treq.txt](https://github.com/python-poetry/poetry/files/8246059/add.treq.txt).

I find the output hard to read (not helped by Twisted's date-based versioning scheme), but I think what's happening is:

- treq 22.2.0 [requires](https://github.com/twisted/treq/blob/d67fb48c0ff12177454ce3fa65e6f8cc92a718e1/setup.py#L35) ` "Twisted[tls] >= 18.7.0"`
- poetry deduces the set of possible twisted versions: `Thus, treq (>=22.2.0,<23.0.0) requires twisted (18.7.0 || 18.9.0 || 19.2.0 || 19.2.1 || 19.7.0 || 19.10.0 || 20.3.0 || 21.2.0 || 21.7.0 || 22.1.0 || 22.2.0).`
- This doesn't include include twisted `22.2.0.post0`, the trunk version; I'm guessing this isn't being considered and that's why resolution fails?

I've also tried marking both with `allow_prereleases=true` but that didn't seem to change any behaviour---I'm guessing this refers to prereleases on PyPI rather than from-git prereleases.

Am I doing something wrong here?

[...]

PR #5413, opened 2022-04-05
export: replace in-tree implementation with poetry-plugin-export

abn:
Once the export plugin code releases 1.x (python-poetry/poetry-export-plugin#46), this PR should add changes to replace the export command.

Closes: #4824
~~Depends-on: #5412~~

**Notes:** 
- Maintain commits when merging.
- Reviewing per commit might be easier.


src/poetry/console/application.py
```diff
@@ -306,10 +306,13 @@ def _configure_installer(self, command: InstallerCommand, io: IO) -> None:
...
         command.set_installer(installer)
 
-    def _load_plugins(self, io: IO) -> None:
+    def _load_plugins(self, io: IO = None) -> None:
         if self._plugins_loaded:
             return
 
+        if io is None:
+            io = NullIO()
+
```
radoering, line 315: This seems to be only required for the test. Not sure, if it's better to just pass NullIO() in the test and don't introduce a default.

radoering (changes requested):
Before merging this PR, there are some issues in `poetry-plugin-export` that should be fixed. (Its tests ran against the old in-tree implementation!)

See python-poetry/poetry-plugin-export#50 for proposed changes.

However, the `poetry-export-plugin` PR is blocked by #5435 which should be merged into master first. Then, this branch can be rebased and python-poetry/poetry-plugin-export#50 can be updated to pass the tests.

Further, there is a change in behavior that I am not sure is intentional: The in-tree export command did not export dev dependencies by default. There was even a test for that default behavior. The `poetry-plugin-export` exports all non-optional dependencies (including dev-depdendencies) by default.

abn:
Thanks @radoering. Ugg; good catches. As for the beaviour change, I commented on the python-poetry/poetry-plugin-export#50. I need to take a closer look at #5435, first glance reminds of https://github.com/python-poetry/poetry/issues/5311#issuecomment-1082474562.

src/poetry/console/application.py
abn: Was on the fence on this one, I am okay either way, but figured we will eventually need this anyway and since the method strictly didn't require a non NullIO, figure it reasonable.

But yeah, will cover it when I do the next review sweep for this one.

abn: Looked through the other cases where this is done, and since the factory allready has a similar pattern employed, I am inclined to keep it as is for now unless there is a strong objection on this.

radoering (changes requested):
I think there two open points that build on each other:

1. decision on default behavior (including `--with/--without` handling), see latest changes in https://github.com/python-poetry/poetry-plugin-export/pull/50 for the variant "no breaking change but different behavior of export command compared to install command"
2. update documentation (should be done after decision on default behavior was made) btw, the documentation has to remain in this repo?

abn:
`poetry-plugin-export==1.0.1` should fix the issues. I have also added an additional test run to verify everything works.

radoering:
> 2\. update documentation (should be done after decision on default behavior was made) btw, the documentation has to remain in this repo?

Still needs to be done.

abn:
Ack, will update.

Until we have a better strategy for handling this, documentation should remain in this repo as the current website publishing only accounts for this repository. But, in the mid-term (outside of this PR), figure out how to get the website to also support any documentation we put in the plugin repository.

The `export` plugin is a snow flake at the moment, so handling it as is for now is okay I reckon. But if we start adding more "bundled" plugins, we should come up with a better solution sooner.

radoering (approved):

Issue #4629, opened 2021-10-12
Poetry not pulling git submodules when installing scipy from source


JPFrancoia:
- [x] I am on the [latest](https://github.com/python-poetry/poetry/releases/latest) Poetry version.
- [x] I have searched the [issues](https://github.com/python-poetry/poetry/issues) of this repo and believe that this is not a duplicate.
- [x] If an exception occurs when executing a command, I executed it again in debug mode (`-vvv` option).

- **OS version and name**: Docker
- **Poetry version**: 1.1.11

## Issue

**This isn't about installing scipy**, I have no problem installing this dependency when it's the latest version from Pypi (i.e: when the wheel is already built).

When installing scipy from source, I'm running into a problem that seems to be because poetry didn't do a `git submodule update --init` when cloning the source.

I will provide minimal Docker files to reproduce the issue.

This is the baseline, where I install a branch of scipy from source. This works:

```docker
FROM python:3.8

RUN apt-get update \
    && apt-get install -y make gcc g++ curl libpq-dev python3-dev libpng-dev libxft-dev gfortran python3-dev libopenblas-dev liblapack-dev

COPY . /app

WORKDIR app/

RUN pip install poetry

RUN pip install git+https://github.com/mckib2/scipy@boost-ibeta-derivative-overflow
```

However, when I try this instead:

```docker
FROM python:3.8

RUN apt-get update \
    && apt-get install -y make gcc g++ curl libpq-dev python3-dev libpng-dev libxft-dev gfortran python3-dev libopenblas-dev liblapack-dev

COPY . /app

WORKDIR app/

RUN pip install poetry

RUN poetry config virtualenvs.create false \
  && pip install wheel \
  && poetry install -vv -n --no-dev
```

I end up with the following error:

```
Traceback (most recent call last):
#9 785.8         File "/usr/local/lib/python3.8/site-packages/pip/_vendor/pep517/in_process/_in_process.py", line 349, in <module>
#9 785.8           main()
#9 785.8         File "/usr/local/lib/python3.8/site-packages/pip/_vendor/pep517/in_process/_in_process.py", line 331, in main
#9 785.8           json_out['return_val'] = hook(**hook_input['kwargs'])
#9 785.8         File "/usr/local/lib/python3.8/site-packages/pip/_vendor/pep517/in_process/_in_process.py", line 151, in prepare_metadata_for_build_wheel
#9 785.8           return hook(metadata_directory, config_settings)
[...]

Issue #5188, opened 2022-02-11
Cloning git repo 4 times when adding a git dependency


yajo:
- [x] I am on the [latest](https://github.com/python-poetry/poetry/releases/latest) Poetry version.
- [x] I have searched the [issues](https://github.com/python-poetry/poetry/issues) of this repo and believe that this is not a duplicate.
- [x] If an exception occurs when executing a command, I executed it again in debug mode (`-vvv` option).

- **OS version and name**: Fedora Silverblue 35
- **Poetry version**: 1.1.13
- **Link of a [Gist](https://gist.github.com/) with the contents of your pyproject.toml file**: Not needed. 

## Issue

To reproduce it:

1. Add a basic `pyproject.toml` with `poetry init`
2. Run `poetry add git+https://github.com/odoo/odoo.git#15.0`

It will take about forever x4.

I included locally a cherry-pick of https://github.com/python-poetry/poetry-core/pull/290 with:

```
pipx install poetry
pipx inject poetry git+https://github.com/moduon/poetry-core.git@stable-git-clone-blobless
```

Then, repeat those steps, and it will take about 5 minutes. Still too much.

I have executed this command:

    time py-spy record --format speedscope --idle --threads --subprocesses --output ~/Downloads/poetry.speedscope.json.txt poetry add git+https://github.com/odoo/odoo.git#15.0

It produced this tracing file, that you can upload to https://www.speedscope.app/ to browse the performance [poetry.speedscope.json.txt](https://github.com/python-poetry/poetry/files/8048313/poetry.speedscope.json.txt)

Once you're browsing that, use the top dropdown thread selector and choose these threads:

- Process 75070 Thread 75070 "" (1/124)
- Process 75070 Thread 75272 "Thread-5 (_install)" (3/124)

Search for "clone" (Use Ctrl+F to open search). You'll see 4 clones being highlighted. I put screenshots here to make it easier in case you're not familiarized with speedscope:

![image](https://user-images.githubusercontent.com/973709/153586470-2fadad2f-5afb-40b3-93f7-a1e2b667771f.png)
![image](https://user-images.githubusercontent.com/973709/153586571-eb9d2fcf-5c12-4311-a98f-ee0e7a09cf58.png)

You can see that each one of those sections takes about 1:00 to 1:20 minutes. Sum the normal poetry operations for solving dependencies and you have the about 5 minutes it takes.

[...]

Issue #5105, opened 2022-01-26
git branch depencies and force-push


jeremad:
- [x] I am on the [latest](https://github.com/python-poetry/poetry/releases/latest) Poetry version.
- [x] I have searched the [issues](https://github.com/python-poetry/poetry/issues) of this repo and believe that this is not a duplicate.
- [x] If an exception occurs when executing a command, I executed it again in debug mode (`-vvv` option).

- **OS version and name**: All -- Non relevant
- **Poetry version**: 1.1.12
- **Link of a [Gist](https://gist.github.com/) with the contents of your pyproject.toml file**: https://gist.github.com/jeremad/9d1cb02b4fa3858640ae93528db5c268

## Issue

I use poetry with a git dependency on a particular branch:
```toml
foo = { git = "https://github.com/bar/foo.git", branch = "feat/branch" }
```
I run `poetry update` and I have my lock file, with the *commit sha* of this particular branch at this particular time.

Now I force push my branch `feat/branch` and I try running `poetry install`, I have the following error:

```
CalledProcessError

  Command '['git', '--git-dir', '.../.cache/pypoetry/virtualenvs/poetry-bug-MXzYaQCP-py3.8/src/foo/.git', '--work-tree', '.../.cache/pypoetry/virtualenvs/poetry-bug-MXzYaQCP-py3.8/src/foo', 'checkout', '99eeaa1910a597818b134319db5641cd61d27978']' returned non-zero exit status 128.

  at ~/.poetry/lib/poetry/utils/_compat.py:217 in run
      213│                 process.wait()
      214│                 raise
      215│             retcode = process.poll()
      216│             if check and retcode:
    → 217│                 raise CalledProcessError(
      218│                     retcode, process.args, output=stdout, stderr=stderr
      219│                 )
      220│         finally:
      221│             # None because our context manager __exit__ does not use them.
```

This is because the *commit sha* of the `poetry.lock` file does not exist anymore (or is not in a branch at least).

I would like this use-case to be valid, without having to run `poetry update` every time my branch changes. This is perhaps more of a feature request than a bug report. I understand that the current behavior is also valid. Could it be an option in the configuration not to write the commit sha in the lock file?

This is related to this issue: https://github.com/python-poetry/poetry/issues/5002 AFAICT

Thank you all for the awesome work,
regards

abn:
@jeremad can you verify if #5428 fixes this issue?

jeremad:
Sure thing!

Issue #5002, opened 2022-01-07
fails to install from git repo when commit doesn't belong to a branch


DetachHead:
- [X] I am on the [latest](https://github.com/python-poetry/poetry/releases/latest) Poetry version.
- [X] I have searched the [issues](https://github.com/python-poetry/poetry/issues) of this repo and believe that this is not a duplicate.
- [X] If an exception occurs when executing a command, I executed it again in debug mode (`-vvv` option).

- **OS version and name**: windows 10
- **Poetry version**: 1.1.12
- **Link of a [Gist](https://gist.github.com/) with the contents of your pyproject.toml file**: https://github.com/KotlinIsland/basedtyping/blob/master/pyproject.toml

## Issue
https://github.com/KotlinIsland/basedtyping/runs/4741756977?check_suite_focus=true
```
  Stack trace:

  8  ~\.local\pipx\venvs\poetry\lib\site-packages\poetry\installation\executor.py:203 in _execute_operation
      201│ 
      202│             try:
    → 203│                 result = self._do_execute_operation(operation)
      204│             except EnvCommandError as e:
      205│                 if e.e.returncode == -2:

  7  ~\.local\pipx\venvs\poetry\lib\site-packages\poetry\installation\executor.py:277 in _do_execute_operation
      275│             return 0
      276│ 
    → 277│         result = getattr(self, "_execute_{}".format(method))(operation)
      278│
      279│         if result != 0:

  6  ~\.local\pipx\venvs\poetry\lib\site-packages\poetry\installation\executor.py:412 in _execute_install
      410│ 
      411│     def _execute_install(self, operation):  # type: (Install) -> None
    → 412│         return self._install(operation)
      413│
      414│     def _execute_update(self, operation):  # type: (Update) -> None

  5  ~\.local\pipx\venvs\poetry\lib\site-packages\poetry\installation\executor.py:431 in _install
      429│ 
      430│         if package.source_type == "git":
    → 431│             return self._install_git(operation)
      432│
      433│         if package.source_type == "file":

  4  ~\.local\pipx\venvs\poetry\lib\site-packages\poetry\installation\executor.py:576 in _install_git
      574│             reference = package.source_reference
      575│ 
    → 576│         git.checkout(reference, src_dir)
      577│
      578│         # Now we just need to install from the source directory

  3  ~\.local\pipx\venvs\poetry\lib\site-packages\poetry\core\vcs\git.py:281 in checkout
      279│         args += ["checkout", rev]
      280│
[...]

PR #5428, opened 2022-04-08
replace git command use with dulwich

abn:
This change introduces dulwich as the git backend, instead of system git executable. Together with an LRU cache when inspecting git package, this considerable improves performance for dependency solver and reuse of source when project has git dependencies.

In cases where dulwich fails with an HTTPUnauthorized error, Poetry falls back to system provided git client as a temporary measure. This will be replaced in the future once dulwich supports git credentials.

As an effect of this, Poetry no longer relies on the subprocess git client implemented in `poetry-core`. This can be removed as soon as core decouples its auto-excluded files detection from git CLI's ignored file listing.

In this change, we also introduce the `experimental.system-git-client` configuration for users needing a escape hatch.

```sh
poetry config experimental.system-git-client true
```

Resolves: #4629 #5002 #3392 #2475 #5188 #5105 #5493

## Example Improvement
When updating a project with git dependency things are much faster.

```console
$ cat pyproject.toml | grep git
poetry = {git = "https://github.com/python-poetry/poetry.git"
$ time poetry@master update --quiet

real	0m38.240s
user	1m57.819s
sys	0m6.819s
$ time poetry@local update --quiet

real	0m7.095s
user	0m3.590s
sys	0m0.334s
```
### Using pipx
```sh
pipx install --suffix=@5428 'poetry @ git+https://github.com/python-poetry/poetry.git@refs/pull/5428/head'
```

```sh
podman run --rm -i --entrypoint bash python:3.10 <<EOF
set -xe
python -m pip install -q git+https://github.com/python-poetry/poetry.git@refs/pull/5428/head
poetry new foobar
poetry add git+https://github.com/python-poetry/poetry.git#master
poetry update
EOF
```

## To Do
- [x] support repositories with submodules
- [x] if revision and local head matches, short-circuit clone

## Future Improvements
- (optional) ~currently source clones when solving uses `{cache-dir}/src` instead of `{virtual-env}/src` as provider does not have the environment available at that time~
- (future) allow clearing of source cache via cache clear command

### DIFF
--- a/src/poetry/vcs/git/backend.py
+++ b/src/poetry/vcs/git/backend.py
@@ -1,0 +1,378 @@
+from __future__ import annotations
+
+import dataclasses
+import logging
+import re
+
+from pathlib import Path
+from subprocess import CalledProcessError
+from typing import TYPE_CHECKING
+
+from dulwich import porcelain
+from dulwich.client import HTTPUnauthorized
+from dulwich.client import get_transport_and_path
+from dulwich.config import ConfigFile
+from dulwich.config import parse_submodules
+from dulwich.errors import NotGitRepository
+from dulwich.refs import ANNOTATED_TAG_SUFFIX
+from dulwich.repo import Repo
+
+from poetry.console.exceptions import PoetrySimpleConsoleException
+from poetry.utils.helpers import safe_rmtree
+
+
+if TYPE_CHECKING:
+    from dataclasses import InitVar
+
+    from dulwich.client import FetchPackResult
+    from dulwich.client import GitClient
+
+
+logger = logging.getLogger(__name__)
+
+
+def is_revision_sha(revision: str | None) -> bool:
+    return re.match(r"^\b[0-9a-f]{5,40}\b$", revision or "") is not None
+
+
+def annotated_tag(ref: str | bytes) -> bytes:
+    if isinstance(ref, str):
+        ref = ref.encode("utf-8")
+    return ref + ANNOTATED_TAG_SUFFIX
+
+
+@dataclasses.dataclass
+class GitRefSpec:
+    branch: str | None = None
+    revision: str | None = None
+    tag: str | None = None
+    ref: bytes = dataclasses.field(default_factory=lambda: b"HEAD")
+
+    def resolve(self, remote_refs: FetchPackResult) -> None:
+        """
+        Resolve the ref using the provided remote refs.
+        """
+        self._normalise(remote_refs=remote_refs)
+        self._set_head(remote_refs=remote_refs)
+
+    def _normalise(self, remote_refs: FetchPackResult) -> None:
+        """
+        Internal helper method to determine if given revision is
+            1. a branch or tag; if so, set corresponding properties.
+            2. a short sha; if so, resolve full sha and set as revision
+        """
+        if self.revision:
+            ref = f"refs/tags/{self.revision}".encode()
+            if ref in remote_refs.refs or annotated_tag(ref) in remote_refs.refs:
+                # this is a tag, incorrectly specified as a revision, tags take priority
+                self.tag = self.revision
+                self.revision = None
+            elif (
+                self.revision.encode("utf-8") in remote_refs.refs
+                or f"refs/heads/{self.revision}".encode() in remote_refs.refs
+            ):
+                # this is most likely a ref spec or a branch incorrectly specified
+                self.branch = self.revision
+                self.revision = None
+        elif (
+            self.branch
+            and f"refs/heads/{self.branch}".encode() not in remote_refs.refs
+            and (
+                f"refs/tags/{self.branch}".encode() in remote_refs.refs
+                or annotated_tag(f"refs/tags/{self.branch}") in remote_refs.refs
+            )
+        ):
+            # this is a tag incorrectly specified as a branch
+            self.tag = self.branch
+            self.branch = None
+
+        if self.revision and self.is_sha_short:
+            # revision is a short sha, resolve to full sha
+            short_sha = self.revision.encode("utf-8")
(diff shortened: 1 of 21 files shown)


abn:
Testing this pull request.

### Using pipx
```sh
pipx install --suffix=@5428 'poetry @ git+https://github.com/python-poetry/poetry.git@refs/pull/5428/head'
```

```sh
podman run --rm -i --entrypoint bash python:3.10 <<EOF
set -xe
python -m pip install -q git+https://github.com/python-poetry/poetry.git@refs/pull/5428/head
poetry new foobar
poetry add git+https://github.com/python-poetry/poetry.git
EOF
```

src/poetry/utils/vcs/git.py
```diff
@@ -0,0 +1,193 @@
...
+from dulwich.repo import Repo
+from poetry.core.utils.helpers import canonicalize_name
+
+
+if TYPE_CHECKING:
+    from dulwich.client import FetchPackResult
+    from dulwich.client import GitClient
+
+
+class Git:
```
finswimmer, line 24: This class doesn't handle any states. So it would be more pythonic to not create a class here and just have the methods within the `git` module.

src/poetry/utils/vcs/git.py
```diff
@@ -0,0 +1,193 @@
...
+        source_root: Path,
+        name: str | None = None,
+        branch: str | None = None,
+        tag: str | None = None,
+        revision: str | None = None,
+        clean: bool = False,
+    ) -> Repo:
+        source_root.mkdir(parents=True, exist_ok=True)
+
+        name = name or canonicalize_name(url.rsplit("/", 1)[-1].rstrip(".git"))
```
finswimmer, line 95: `rstrip()` strips any character given and not only the word  ".git"

src/poetry/utils/vcs/git.py
```diff
@@ -0,0 +1,193 @@
...
+    def clone(
+        cls,
+        url: str,
+        source_root: Path,
+        name: str | None = None,
+        branch: str | None = None,
+        tag: str | None = None,
+        revision: str | None = None,
+        clean: bool = False,
+    ) -> Repo:
```
finswimmer, line 92: This method is quite large. It would be easier to read and maintain if we extract some stuff into its own methods.

finswimmer (dismissed):

src/poetry/utils/vcs/git.py
abn: Not necessarily. This pattern allows for a more cleaner usage. Don't think this is required even if no state is maintained.

src/poetry/utils/vcs/git.py
abn: Good catch. It was a copy paste from existing code elsewhere.

src/poetry/utils/vcs/git.py
abn: Hmm, I'll see if it makes sense to break it up.

alxgruU:
hi, any news for this issue ? we really need this fix

Darsstar:
[Since Dulwich does not support credential helpers](https://github.com/jelmer/dulwich/issues/873) that use case gets broken by this PR.

I don't expect this PR getting merged to be insurmountable for us since we could just publish it as a package on our Azure DevOps Artifact Feed after some minor refactoring.  But feedback is feedback.

abn:
Thanks @Darsstar that is something I was debating. I'm planning one of the following as a follow up PR:

1. Fallback option to system git as a temporary stop gap.
2. Add credential handling via similar mechanisms to `http-basic.repo` config. 

I think 2 will be desirable.

abn:
> [Since Dulwich does not support credential helpers](https://github.com/jelmer/dulwich/issues/873) that use case gets broken by this PR.
> 
> I don't expect this PR getting merged to be insurmountable for us since we could just publish it as a package on our Azure DevOps Artifact Feed after some minor refactoring. But feedback is feedback.

@Darsstar  I have added a fallback for cases where http auth fails retaining old behaviour until dulwich has a fix upstream. Can you test it works as expected in your case?

Darsstar:
I can confirm the fallback works.

src/poetry/vcs/git/backend.py
```diff
@@ -0,0 +1,396 @@
...
+        url: str,
+        name: str | None = None,
+        branch: str | None = None,
+        tag: str | None = None,
+        revision: str | None = None,
+        source_root: Path | None = None,
+        clean: bool = False,
+    ) -> Repo:
+        if not source_root:
+            source_root = cls.get_default_source_root()
```
finswimmer, line 346: Just a personal taste:

```suggestion
        source_root = source_root or cls.get_default_source_root()
```

Some line below you are doing this as well with `name`.

tests/integration/test_utils_vcs_git.py
```diff
@@ -0,0 +1,259 @@
...
+if TYPE_CHECKING:
+    from _pytest.tmpdir import TempdirFactory
+    from dulwich.client import FetchPackResult
+    from dulwich.client import GitClient
+    from pytest_mock import MockerFixture
+
+    from tests.conftest import Config
+
+
+# these tests are integration as they rely on an external repository
```
finswimmer, line 31: We should extend this comment by saying on which external repository these tests rely. (I guess https://github.com/python-poetry/test-fixture-vcs-repository.git ?) Also updating the README in this repo, with information what this repo is for, would be useful.

finswimmer (changes requested):

src/poetry/vcs/git/backend.py
abn: Sure, I suspect additional logic got stripped away.

finswimmer (approved):

Issue #4667, opened 2021-10-22
Poetry init OSError on windows 10


NujanZh:
- [x] I am on the [latest](https://github.com/python-poetry/poetry/releases/latest) Poetry version.
- [x] I have searched the [issues](https://github.com/python-poetry/poetry/issues) of this repo and believe that this is not a duplicate.
- [x] If an exception occurs when executing a command, I executed it again in debug mode (`-vvv` option).

- **Windows 10**: 
- **1.1.11**: 
- **I can't init so i don't have pyprojecttoml**: 

## Issue

I type `poetry init` and it give me this error:
`
[WinError 1920] Доступ к этому файлу из системы отсутствует (Translate: this file cannot be accessed from the system.): 'C:\\Users\\*****\\AppData\\Local\\Microsoft\\WindowsApps\\PythonSoftwareFoundation.Python.3.8_qbz5n2kfra8p0\\python.exe\\Scripts'

  at C:\Program Files\WindowsApps\PythonSoftwareFoundation.Python.3.8_3.8.2800.0_x64__qbz5n2kfra8p0\lib\pathlib.py:1198 in stat
      1194│
      1195│         Return the result of the stat() system call on this path, like
      1196│         os.stat() does.
      1197│         """
    → 1198│         return self._accessor.stat(self)
      1199│
      1200│     def owner(self):
      1201│
      1202│         Return the login name of the file owner.

`

roslovets:
The same error in Windows 11 for Python 3.9 installed via Windows Store.

noi4eg:
same error win10 for Python 3.9 and 3.10 installed via WS

Fethi1:
same error

new-divos:
same error

spavett:
same error:

Win 11 
python 3.9 from windows store
poetry 1.1.12 installed via powershell

I was trying to run the init through pycharm which gives some traceback:

  [WinError 1920] The file cannot be accessed by the system: 'C:\\Users\\steve\\AppData\\Local\\Microsoft\\WindowsApps\\PythonSoftwareFoundation.Python.3.9_qbz5n2kfra8p0\\python.exe\\Scripts'

  at C:\Program Files\WindowsApps\PythonSoftwareFoundation.Python.3.9_3.9.2544.0_x64__qbz5n2kfra8p0\lib\pathlib.py:1232 in stat

Traceback (most recent call last):
[...]

PR #5433, opened 2022-04-09
Issue #4667: msft has a different bin directory for python

Charlieung:
# Pull Request Check List

Resolves: #4667 

Just adds another condition to check if it's in the default MSFT directory, which doesn't have the default /bin path for the python executable.

### DIFF
--- a/src/poetry/utils/env.py
+++ b/src/poetry/utils/env.py
@@ -1122,6 +1122,7 @@ class Env:
 
     def __init__(self, path: Path, base: Path | None = None) -> None:
         self._is_windows = sys.platform == "win32"
+        self._is_windows_from_microsoft_store = Path.home() / 'AppData' / 'Local' / 'Microsoft' in Path(sys.executable).parents
         self._is_mingw = sysconfig.get_platform().startswith("mingw")
         self._is_conda = bool(os.environ.get("CONDA_DEFAULT_ENV"))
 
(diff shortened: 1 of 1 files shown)


TechnoTalksDev (approved):
Seems great to help with Microsoft Store installs

src/poetry/utils/env.py
```diff
@@ -1130,7 +1131,7 @@ def __init__(self, path: Path, base: Path | None = None) -> None:
         else:
             bin_dir = "Scripts"
         self._path = path
-        self._bin_dir = self._path / bin_dir
+        self._bin_dir = self._path if self._is_windows_from_microsoft_store else self._path / bin_dir
```
neersighted, line 1134:
```suggestion
        self._bin_dir = self._path
        if not self._is_windows_from_microsoft_store:
            self._path /= bin_dir
```

neersighted (changes requested):

chumbert:
I'd be interested in seeing this merged.

If @Charlieung isn't active here anymore I can open another PR with the changes suggested by @neersighted.

finswimmer:
@chumbert: Please test it with 1.2.0rc1. Might be that this is solved already by another PR.

abn:
The MS store version does have [other issues](https://docs.python.org/3/using/windows.html#redirection-of-local-data-registry-and-temporary-paths) as well. The python.org version is probably preferred for development environments. Additionally, I suspect the original issue this PR set out to fix is no longer present.

Issue #5031, opened 2022-01-13
(📚) The documentation doesn't mention the `self` command at all


KotlinIsland:
- [x] I have searched the [issues](https://github.com/python-poetry/poetry/issues) of this repo and believe that this is not a duplicate.

## Issue

`poetry list` shows the `self` subcommand and `poetry help "self update"` details it, but the [docs](https://python-poetry.org/docs/master/cli) don't mention it at all.

PR #5450, opened 2022-04-14
replace `poetry plugin` commands with more generic `poetry self` commands

abn:
### Conceptual Changes

1. The poetry instance's runtime additional packages are now maintained in `<user-config-dir>/pypoetry/pyproject.toml` instead of the current (`1.2.0b1`) localtion at the base of the virtual envrionment. This was changes in order to preserve plugin/additional package configuration even when the virtual environment (Poetry's runtime environment) was recreated. You can simply do `poetry self install` to get everything back in order. (_Might need to implement `self lock` command to faciititate some use cases_).
2. The `self` command now simply switches context to `<user-config-dir>/pypoetry` and treats it as a Poetry managed project.
3. Additional packages are stored in the `tool.poetry.group.additional.dependencies` section of the `<user-config-dir>/pypoetry/pyproject.toml` file and can be listed using `poetry self show --addons`.
4. Additional packages can now be any package, not just Poetry plugins. This allows for use cases where you want to override a single dependency in the environment and/or add plugin for another package (eg: keyring, requests etc.).

### Commands Added
<details>
<summary><b>self add</b></summary>
<p>

```console
$ poetry self add --help

Description:
  Add additional packages to Poetry's runtime environment.

Usage:
  self add [options] [--] <name>...

Arguments:
  name                     The packages to add.

Options:
  -e, --editable           Add vcs/path dependencies as editable.
  -E, --extras=EXTRAS      Extras to activate for the dependency. (multiple values allowed)
      --source=SOURCE      Name of the source to use to install the package.
      --allow-prereleases  Accept prereleases.
      --dry-run            Output the operations but do not execute anything (implicitly enables --verbose).
<--snipped-->

Help:
  The self add command installs additional package's to Poetry's runtime environment.
  
  This is managed in the /home/abn/.config/pypoetry/pyproject.toml file.
  
  If you do not specify a version constraint, poetry will choose a suitable one based on the available package versions.
  
  You can specify a package in the following forms:
<--snipped-->
```

</p>
</details>

<details>
<summary><b>self remove</b></summary>
<p>

```console
$ poetry self remove --help

Description:
  Remove additional packages from Poetry's runtime environment.

Usage:
  self remove [options] [--] <packages>...

Arguments:
  packages              The packages to remove.

Options:
      --dry-run         Output the operations but do not execute anything (implicitly enables --verbose).
<--snipped-->

Help:
  The self remove command removes additional package's to Poetry's runtime environment.
  
  This is managed in the /home/abn/.config/pypoetry/pyproject.toml file.
```

</p>
</details>

<details>
<summary><b>self install</b></summary>
<p>

```console
$ poetry self install --help

Description:
  Add additional packages to Poetry's runtime environment.

Usage:
  self install [options]

Options:
      --sync            Synchronize the environment with the locked packages and the specified groups.
      --dry-run         Output the operations but do not execute anything (implicitly enables --verbose).
<--snipped-->

Help:
  The self install command all additional packages specified are installed in the current runtime environment.
  
  This is managed in the /home/abn/.config/pypoetry/pyproject.toml file.
  
  You can add more packages using the self add command and remove them using the self remove command.
```

</p>
</details>

<details>
<summary><b>self show</b></summary>
<p>

```console
$ poetry self show --help

Description:
  Show packages from Poetry's runtime environment.

Usage:
  self show [options] [--] [<package>]

Arguments:
  package               The package to inspect

Options:
      --addons          List only add-on packages installed.
  -t, --tree            List the dependencies as a tree.
  -l, --latest          Show the latest version.
  -o, --outdated        Show the latest version but only for packages that are outdated.
<--snipped-->

Help:
  The self show command behaves similar to the show command, but
  working within Poetry's runtime environment. This lists all packages installed within
  the Poetry install environment.
  
  To show only additional packages that have been added via self add and their
  dependencies use self show --addons.
  
  This is managed in the /home/abn/.config/pypoetry/pyproject.toml file.
```

</p>
</details>

<details>
<summary><b>self show plugins</b></summary>
<p>

```console
$ poetry self show plugins --help

Description:
  Shows information about the currently installed plugins.

Usage:
  self show plugins [options]

Options:
<--snipped-->

Help:
  The self show plugins command lists all installed Poetry plugins.
  
  Plugins can be added and removed using the self add and self remove commands respectively.
  
  This command does not list packages that do not provide a Poetry plugin.
```

</p>
</details>

### Commands Changed
1. `self update` now wraps around `update` instead for special handling.
2. `plugin add` is now deprecated and re-routes to `self add`
5. `plugin remove` is now deprecated and re-routes to `self remove`
6. `plugin show` is now deprecated and re-routes to `self show plugins`

### Example Usages
```sh
poetry self add artifacts-keyring
poetry self add git+https://github.com/python-poetry/poetry-core@master
poetry self add poetry-plugin-export
```

### Using pipx
```sh
pipx install --suffix=@5450 'poetry @ git+https://github.com/python-poetry/poetry.git@refs/pull/5450/head'
poetry@5450 self add poetry-export-plugin
poetry@5450 self remove poetry-export-plugin
poetry@5450 self install --sync
poetry@5450 self update --preview
poetry@5450 self add artifacts-keyring
poetry@5450 self add git+https://github.com/python-poetry/poetry-core@master
cat ~/.config/pypoetry/pyproject.toml
```

```sh
podman run --rm -i --entrypoint bash docker.io/python:3.10 <<EOF
set -xe
python -m pip install -q git+https://github.com/python-poetry/poetry.git@refs/pull/5450/head
poetry self add poetry-export-plugin
poetry self remove poetry-export-plugin
poetry self install --sync
poetry self update --preview
poetry self add artifacts-keyring
poetry self add git+https://github.com/python-poetry/poetry-core@master
cat ~/.config/pypoetry/pyproject.toml
EOF
```


abn:
Testing this changeset.

### Using pipx
```sh
pipx install --suffix=@5450 'poetry @ git+https://github.com/python-poetry/poetry.git@refs/pull/5450/head'
poetry@5450 self add poetry-export-plugin
poetry@5450 self remove poetry-export-plugin
poetry@5450 self install --sync
poetry@5450 self update --preview
cat ~/.config/pypoetry/pyproject.toml
```

```sh
podman run --rm -i --entrypoint bash docker.io/python:3.10 <<EOF
set -xe
python -m pip install -q git+https://github.com/python-poetry/poetry.git@refs/pull/5450/head
poetry self add poetry-export-plugin
poetry self remove poetry-export-plugin
poetry self install --sync
poetry self update --preview
cat ~/.config/pypoetry/pyproject.toml
EOF
```

abn:
Functional changes are ready for review. Will need to add some documentation changes.

src/poetry/console/commands/self/self_command.py
```diff
@@ -0,0 +1,119 @@
...
+
+    def reset(self) -> None:
+        self.reset_env()
+        self.reset_poetry()
+
+    def handle(self) -> int:
+        self.reset()
+
+        with self.with_system_project():
+            return self._system_project_handle()
```
finswimmer, line 119: Can you please explain when and how this method is used?

The other `self` commands inherit from `SelfCommand` and doesn't overwrite the `handle` method. In case of classes that inherit from multiple base classes `SelfCommand` is always the most left one, so it's implementation would take precedence over implementation in the class(es), wouldn't it?

abn: If I am understanding your concern correctly, the short answer is [C3 linearlization](https://en.wikipedia.org/wiki/C3_linearization) ensures the order of execution is defined by the [MRO](https://www.python.org/download/releases/2.3/mro/) ([simplified](https://www.educative.io/edpresso/what-is-mro-in-python)) in the child class not parent/siblings. The example below should show a cut down version of what is happening.

<details>
<summary><b>example.py</b></summary>
<p>

```py


def mro(clz: type) -> None:
    print(f"MRO({clz.__name__}): ", " -> ".join(map(lambda m: m.__name__, clz.mro())))


class InstallerCommand:
    def handle(self) -> None:
        print("InstallerCommand.handle() from ", self.__class__.__name__)
        mro(InstallerCommand)
        raise NotImplementedError()


class InstallCommand(InstallerCommand):
    def handle(self) -> None:
        print("InstallCommand.handle() from ", self.__class__.__name__)
        mro(InstallCommand)


class SelfCommand(InstallerCommand):
    def _system_project_handle(self) -> None:
        print("SelfCommand._system_project_handle() from ", self.__class__.__name__)
        super().handle()

    def handle(self) -> None:
        print("SelfCommand.handle() from ", self.__class__.__name__)
        mro(SelfCommand)
        self._system_project_handle()


class SelfInstallCommand(SelfCommand, InstallCommand):
    # this is a passthrough command, ie. it modifies options and description only
    pass


class SelfCustomCommand(SelfCommand):
    # this command is a new command that is not wrapping an existing command, or needs
    # pre-processing prior to calling another command

    def _system_project_handle(self) -> None:
        print(
            "SelfCustomCommand._system_project_handle() from ", self.__class__.__name__
        )
        mro(SelfCustomCommand)


if __name__ == "__main__":
    print("##########################")
    mro(SelfInstallCommand)
    SelfInstallCommand().handle()
    print("##########################")
    mro(SelfCustomCommand)
    SelfCustomCommand().handle()
    print("##########################")

```

</p>
</details>

The above code will output something like this.
```console
##########################
MRO(SelfInstallCommand):  SelfInstallCommand -> SelfCommand -> InstallCommand -> InstallerCommand -> object
SelfCommand.handle() from  SelfInstallCommand
MRO(SelfCommand):  SelfCommand -> InstallerCommand -> object
SelfCommand._system_project_handle() from  SelfInstallCommand
InstallCommand.handle() from  SelfInstallCommand
MRO(InstallCommand):  InstallCommand -> InstallerCommand -> object
##########################
MRO(SelfCustomCommand):  SelfCustomCommand -> SelfCommand -> InstallerCommand -> object
SelfCommand.handle() from  SelfCustomCommand
MRO(SelfCommand):  SelfCommand -> InstallerCommand -> object
SelfCustomCommand._system_project_handle() from  SelfCustomCommand
MRO(SelfCustomCommand):  SelfCustomCommand -> SelfCommand -> InstallerCommand -> object
##########################
```

Regarding when it is used, self commands are currently designed for 2 use cases.

1. A wrapper to poetry top level commands like (add, remove, show) that only differs in that it manages the "poetry system" project; and may have different options/arguments/descriptions etc.
2. A command that either has its own handle logic or requiring pre-processing prior to executing the system project.

The `SelfCommand` base class handles the boilerplate of switching context and resetting the instance. The `_system_project_handle` method acts as an in-context overwritable hook for child classes.

I will add some additional in-code documentation later.

finswimmer: I was right with the interpretation of what `handle` method is called in case of multiple inheritance. I just overlooked that `_system_project_handle` actually calls the `handle` method of the "non-self command". Got it now :+1:

src/poetry/console/commands/self/self_command.py
```diff
@@ -0,0 +1,127 @@
...
+            self.generate_system_pyproject()
+            self._poetry = Factory().create_poetry(
+                self.system_pyproject.parent, io=self._io, disable_plugins=True
+            )
+
+    @property
+    def poetry(self) -> Poetry:
+        if self._poetry is None:
+            self.reset_poetry()
+        assert self._poetry
```
finswimmer, line 92: Unnecessary, right?

src/poetry/console/commands/self/update.py
```diff
@@ -36,192 +30,30 @@ class SelfUpdateCommand(Command):
...
+The <c1>self update</c1> command updates Poetry version in its current runtime \
+environment.
+"""
+
+    def _system_project_handle(self) -> int:
+        self.write("<info>Updating Poetry version ...</info>\n\n")
+        application = cast(Application, self.application)
+        add_command: AddCommand = cast(AddCommand, application.find("add"))
+        add_command.set_env(self.env)
+        application._configure_installer(add_command, self._io)
```
finswimmer, line 43: We should think about renaming `_configure_installer` to `configure_installer` if there is a need for public access to this method.

src/poetry/console/commands/self/self_command.py
abn: That was to satisfy mypy. I will replace that with `return cast(Poetry, self._poetry)` instead.

src/poetry/console/commands/self/update.py
abn: I agree. Fixed this.

src/poetry/console/commands/self/self_command.py
```diff
@@ -0,0 +1,127 @@
...
+
+        if self.system_pyproject.exists():
+            content = PyProjectTOML(self.system_pyproject).poetry_config
+
+            for key in {"group", "source"}:
+                if key in content:
+                    preserved[key] = content[key]
+
+        package = ProjectPackage(name="poetry-instance", version=__version__)
+        package.add_dependency(Dependency(name="poetry", constraint=f"^{__version__}"))
```
finswimmer, line 71: I think we shouldn't have a version range for poetry here and pin it instead. We don't know what range - if any - is given by plugin authors and we might end up in situation where a plugin is only available for a small range.

finswimmer (changes requested):

src/poetry/console/commands/self/self_command.py
abn: I agree. My initial implementation did the same. However, bug fixed in python-poetry/poetry-core#327 causes problems. We can pin it once a new core release with the fix is used.

finswimmer: Have you marked this accidentally as resolved? There's still a version range instead of a pinned version. :thinking:

src/poetry/console/commands/self/install.py
```diff
@@ -0,0 +1,32 @@
+from __future__ import annotations
+
+
+try:
+    from poetry.core.packages.dependency_group import MAIN_GROUP
+except ImportError:
+    MAIN_GROUP = "default"
```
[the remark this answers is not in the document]
abn: It's a compatibility fix for the upcoming core release. There are a few places in the code where this is used as the DEFAULT_GROUP import doesn't exist in the current core version.

src/poetry/console/commands/self/self_command.py
abn: Not accidental. Will move to pinned version once a new core version is used in poetry.

finswimmer: What's your plan to handle this? Merging now and change it in another PR once a new core release is out? (Hopefully we don't forget it until then :crossed_fingers:)? Or leave this PR open until core is released?

abn: I'd rather get the fix done later as the impact of this is not really huge.

finswimmer (dismissed):

finswimmer (dismissed):
At least the docs about how to add a plugin should be updated as well. At best this is done for all `self` commands :)

Secrus:
Docs are missing (you could also add mention of `self update` there, it would close #5031).

abn:
One scenario that has not been handled is a case where multiple `poetry` instances exist. Right now everything shares the config.

abn:
@finswimmer gentle nudge

finswimmer (approved):
Let it fly :airplane:

Issue #999, opened 2019-03-30
Display progress bar when downloading larger wheels


xelibrion:
- [x] I have searched the [issues](https://github.com/sdispater/poetry/issues) of this repo and believe that this is not a duplicate.
- [x] I have searched the [documentation](https://poetry.eustace.io/docs/) and believe that my question is not covered.

## Feature Request

Hi! 

Just tried to install Pytorch in my very first Poetry project and some of the experience left me guessing what was actually happening on my PC.

When installing a larger wheel (I believe for Pytorch it is in 500-700MB range) in a brand new environment, developers are presented with incredibly long dependency resolution times. 
```
Resolving dependencies... (115.8s)
```
After that follows a package installation phase, that takes a bit of time as well (it seems there's some internet activity is happening during this phase as well), but specifying `-vvv` does not show anything other than:
```
  - Installing torch (1.0.1.post2)
```

Would be great to add a bit more transparency in what's actually happening when installing those larger packages.

hangtwenty:
This sort-of relates to [how `filelock` can get into a deadlock'd state depending on the cache](https://github.com/sdispater/poetry/issues/832#issuecomment-480708729); if you don't know about the progress for the big dependency, it may be ambiguous if you are deadlocked vs just waiting for a big download.

PR #5451, opened 2022-04-14
add progress to `download_file` if in Indicator context

tgolsson:
# Pull Request Check List

No meaningful documentation to add and hard to test a progress-bar.

---

Resolves: #999

This code adds a context field for the Indicator that can be used for download_file.
image

This has to be integrated with the `Indicator` to synchronize the output, otherwise it leads to corruption/racey output since `ProgressIndicator.auto` writes from a second thread. 

Based on that requirement I opted to use a class field to store this context, as otherwise almost every function involved in dependency resolution got an `indicator` argument to feed it to `download_file`. This felt extremely noisy, and also hit some circular imports.


src/poetry/puzzle/provider.py
```diff
@@ -49,6 +49,18 @@
...
+
+    @staticmethod
+    def set_context(context: str | None) -> None:
+        Indicator.CONTEXT = context
+
+    def _formatter_context(self) -> str:
+        if Indicator.CONTEXT is None:
+            return " "
+        else:
+            return f": {Indicator.CONTEXT} "
```
abn, line 62:
```suggestion
            return f" <c1>{Indicator.CONTEXT}</> "
```

src/poetry/utils/helpers.py
```diff
@@ -94,16 +94,41 @@ def download_file(
...
                 f.write(chunk)
 
+                if set_indicator:
+                    fetched_size += len(chunk)
+                    percent = (fetched_size * 100) // total_size
+                    if percent > last_percent:
+                        last_percent = percent
+                        Indicator.set_context(f"Downloading {url} {percent:3}%")
+
+    Indicator.set_context(None)
```
abn, line 130: We should probably put this into a context manager to ensure it gets always reset. How does this behave when parallel downloads are in progress?

src/poetry/puzzle/provider.py
tgolsson: @abn Shouldn't that be `</c1>` at the end?

src/poetry/utils/helpers.py
tgolsson: I'll see if I can do a context manager in a nice pattern. 

I wasn't aware that parallel DLs could happen; that complicates things a lot. Can you give an example for when that happens and I'll investigate?

src/poetry/puzzle/provider.py
abn: Both work from testing. But more specific is better I guess.

src/poetry/utils/helpers.py
radoering: Just an educated guess (not verified):

I don't think that parallel downloads happen during dependency resolution (afaik dependency resolution is sequential). I could imagine that parallel downloads can happen during installation (if the packages are not yet cached) because multiple packages can be installed in parallel. Even though the `download_file` function is used outside of dependency resolution, the `Indicator` seems to be only used during dependency resolution.

tgolsson: Yeah; I couldn't see any paths where that would happen. I'm back at my computer now after easter so will take another check tonight if I can find time.

neersighted (dismissed):

neersighted (approved):

Issue #3413, opened 2020-11-25
Flag to install all extras


l0b0:
- [x] I have searched the [issues](https://github.com/python-poetry/poetry/issues) of this repo and believe that this is not a duplicate.
- [x] I have searched the [documentation](https://python-poetry.org/docs/) and believe that my question is not covered.

## Feature Request

It would be handy to have a flag like `--all-extras` to install every package listed in any of the extras.

The specific use case is that I want the IDE to index the packages for every subsystem (AWS Lambda jobs in this case) for easier development.

Workarounds:

1. Creating a separate "extra" section with all the packages from all the other extras ([example](https://github.com/biocatchltd/yellowbox/pull/38)). This means duplication, which means it would be really easy for the sections to get out of sync.
2. Listing each extra specifically in the developer instructions.
3. [This ugly code](https://stackoverflow.com/a/68216833/96588), used in anger [here](https://github.com/linz/geostore/blob/67ac5a6f248aedb5da11f7e37d11b04a28292e8a/.github/workflows/test.yml#L39-L40): `poetry install --extras="$(sed --quiet '/\[tool\.poetry\.extras\]/,/^\[/{s/^\(.*\) = \[/\1/p}' pyproject.toml | sed --null-data 's/\n/ /g;s/ $//')" --remove-untracked`

sinoroc:
Related: https://github.com/python-poetry/poetry/issues/3369

thclark:
Seconded for this one - I want to provide my team a vscode devcontainer that installs everything out of the box, and don't want to have to remember to update the `.devcontainer` file if we add extras to the library we're developing... so sounds like the same use case as the OP.

abn:
Would supporting `--extras .` or `--extras "*" ` as an alias for enabling all extras work here?

l0b0:
`--all-extras` would be more explicit, and would avoid a) a special value and b) supporting any kind of regex, which would be another can of worms.

kroeschl:
I don't have an opinion either way, but I'd like to use this feature to make dev testing somewhat easier. I posted a PR for `--all-extras`: #5452

rjurney:
Any idea when this will be released?

ecs-jnguyen:
It looks like the PR only got merged to master which is on version 1.2.0 beta. Could we get the `--all-extras` flag merged into 1.1 which is installed by default?

rjurney:
@ecs-jnguyen @kroeschl @neersighted that would be awesome!

[...]

PR #5452, opened 2022-04-14
All extras

kroeschl:
# Add `--all-extras` flag to `install` command

Resolves: #3413

Add an `--all-extras` flag to the `install` command for projects which use multiple sets of extras. Update documentation for the new flag.


src/poetry/console/commands/install.py
```diff
@@ -79,12 +80,20 @@ def handle(self) -> int:
...
-        extras = []
-        for extra in self.option("extras"):
-            if " " in extra:
-                extras += [e.strip() for e in extra.split(" ")]
-            else:
-                extras.append(extra)
+        if self.option("extras") and self.option("all-extras"):
+            raise ValueError(
+                "You cannot specify explicit extras while installing all extras"
+            )
```
abn, line 86: Since this is a user level error, let's replace this with either a `self.line_error` with `<error>` formatting; or a `PoetrySimpleConsoleError`. The `ValueError` iirc raises a stack trace when encountered.

abn (dismissed):
Thank you @kroeschl. Looks good, single comment.

tests/console/commands/test_install.py
```diff
@@ -130,3 +136,42 @@ def test_sync_option_is_passed_to_the_installer(
...
+    The --extras doesn't make sense with --all-extras.
+    """
+    mocker.patch.object(tester.command.installer, "run", return_value=0)
+
+    tester.execute("--extras foo --all-extras")
+
+    assert tester.status_code == 1
+    assert (
+        tester.io.fetch_error()
+        == "You cannot specify explicit `--extras` while installing `--all-extras`.\n"
```
l0b0, line 176: I'm not sure about using `--all-extras` as words rather than treating it as a single token in this error message. I'd expect something like "You can't specify both `--extras` and `--all-extras`.".

src/poetry/console/commands/install.py
```diff
@@ -79,12 +80,23 @@ def handle(self) -> int:
...
-                extras += [e.strip() for e in extra.split(" ")]
-            else:
-                extras.append(extra)
+        if self.option("extras") and self.option("all-extras"):
+            self.line_error(
+                "<error>You cannot specify explicit"
+                " `<fg=yellow;options=bold>--extras</>` while installing"
+                " `<fg=yellow;options=bold>--all-extras</>`.</error>"
+            )
+            return 1
```
l0b0, line 89: What do you think of `--all-extras` instead overriding `--extras`, maybe emitting a warning rather than an error when both are specified?

pkulev:
Will `--extras all` a better choice?

l0b0:
> Will `--extras all` a better choice?

No, that makes for a "magic" value which could break existing setups.

kroeschl:
> > Will `--extras all` a better choice?
> 
> No, that makes for a "magic" value which could break existing setups.

If we really wanted to go that route, `--extras "*"` or similar would work better since Poetry (at least 1.1.13) currently chokes on an extras group named `*`, so no one could be depending on it.

l0b0:
> > > Will `--extras all` a better choice?
> > 
> > 
> > No, that makes for a "magic" value which could break existing setups.
> 
> If we really wanted to go that route, `--extras "*"` or similar would work better since Poetry (at least 1.1.13) currently chokes on an extras group named `*`, so no one could be depending on it.

I'd avoid that too:

- It's another magic value, unfamiliar to anyone who isn't using globs.
- People familiar with globs will complain that it can't be used like an actual glob, for example in `--extras='backend-*'`.

src/poetry/console/commands/install.py
neersighted: I'd rather throw the error to prevent unexpected behavior for inattentive users.

src/poetry/console/commands/install.py
```diff
@@ -79,12 +80,23 @@ def handle(self) -> int:
...
-        for extra in self.option("extras"):
-            if " " in extra:
-                extras += [e.strip() for e in extra.split(" ")]
-            else:
-                extras.append(extra)
+        if self.option("extras") and self.option("all-extras"):
+            self.line_error(
+                "<error>You cannot specify explicit"
+                " `<fg=yellow;options=bold>--extras</>` while installing"
+                " `<fg=yellow;options=bold>--all-extras</>`.</error>"
```
neersighted, line 87:
```suggestion
                " `using <fg=yellow;options=bold>--all-extras</>`.</error>"
```

tests/console/commands/test_install.py
```diff
@@ -132,3 +138,42 @@ def test_sync_option_is_passed_to_the_installer(
...
```
neersighted, line 178:
```suggestion
        == "You cannot specify explicit `--extras` while installing using `--all-extras`.\n"
```

src/poetry/console/commands/install.py
```diff
@@ -79,12 +80,23 @@ def handle(self) -> int:
...
-        for extra in self.option("extras", []):
-            if " " in extra:
-                extras += [e.strip() for e in extra.split(" ")]
-            else:
-                extras.append(extra)
+        if self.option("extras") and self.option("all-extras"):
+            self.line_error(
+                "<error>You cannot specify explicit"
+                " `<fg=yellow;options=bold>--extras</>` while installing"
+                " `using <fg=yellow;options=bold>--all-extras</>`.</error>"
```
neersighted, line 87:
```suggestion
                " using `<fg=yellow;options=bold>--all-extras</>`.</error>"
```

neersighted:
pre-commit.ci autofix

neersighted (approved):

kroeschl:
Thanks for cleaning this up @neersighted. 👍

rjurney:
@neersighted @kroeschl any idea when this will be released? It isn't in 1.1.14 I see.

kroeschl:
> @neersighted @kroeschl any idea when this will be released? It isn't in 1.1.14 I see.

This is in [1.2.0b2](https://github.com/python-poetry/poetry/releases/tag/1.2.0b2) and later, but it was never ported it back to 1.1.x.

AIGeneratedUsername:
It looks like you forgot about adding `--all-extras` to the `export` command. With this merge request you made Poetry functionality incomplete. 

@kroeschl Would you add `--all-extras` to the `export` command? Sorry, I have problems with my availability to submit a PR by myself.

neersighted:
> With this merge request you made Poetry functionality incomplete.

The addition of new features in Poetry proper does not relate to the implementation of new features in plugins. I don't think it's helpful or fair to say an accepted change that doesn't scratch your particular itch or do everything you would like to see is deficient in this situation.

In any case, there is an open request on python-poetry/poetry-plugin-export#153 to implement the same feature in the plugin; please add your :+1: there.

PR #5458, opened 2022-04-16
handle poetry 1.1 lock content hash

abn:
This change ensures that content-hash generated by poetry < 1.2 does not trigger unnecessary warnings and operations. We do this by using a backwards compatible hashing logic.

### DIFF
--- a/src/poetry/packages/locker.py
+++ b/src/poetry/packages/locker.py
@@ -46,7 +46,8 @@ class Locker:
 
     _VERSION = "1.1"
 
-    _relevant_keys = ["dependencies", "group", "source", "extras"]
+    _legacy_keys = ["dependencies", "source", "extras", "dev-dependencies"]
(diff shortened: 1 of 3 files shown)


src/poetry/packages/locker.py
```diff
@@ -82,6 +83,7 @@ def is_fresh(self) -> bool:
         metadata = lock.get("metadata", {})
 
         if "content-hash" in metadata:
+            print(self._content_hash, lock["metadata"]["content-hash"])
```
finswimmer, line 86: Was this just for testing and can be cleaned up?

finswimmer (changes requested):

src/poetry/packages/locker.py
abn: ugg; yeah. fixed.

finswimmer (approved):
LGTM :+1:

PR #5461, opened 2022-04-16
feat(console/commands): add --all-groups

lovesegfault:
# Pull Request Check List

Resolves: N/A

This adds a utility flag `--all-groups` to `install` which enables all available
dependency groups.

It's handy if you have lots of groups which you almost always want.

### DIFF
--- a/docs/cli.md
+++ b/docs/cli.md
@@ -163,6 +163,13 @@ It's also possible to only install specific dependency groups by using the `only
 poetry install --only test,docs
 ```
 
+You may also enable all available dependency groups with the `--all-groups`
+option:
+
+```bash
+poetry install --all-groups
+```
+
 {{% note %}}
 The `--dev-only` option is now deprecated. You should use the `--only dev` notation instead.
 {{% /note %}}
--- a/src/poetry/console/commands/group_command.py
+++ b/src/poetry/console/commands/group_command.py
@@ -18,6 +18,11 @@ class GroupCommand(EnvCommand):
     def _group_dependency_options() -> list[Option]:
         return [
             option(
+                "all-groups",
+                None,
+                "Include all dependency groups.",
+            ),
+            option(
                 "without",
                 None,
                 "The dependency groups to ignore.",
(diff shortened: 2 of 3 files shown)


lovesegfault:
One alternative to this would be to treat `all` a special value to `--with`, so you could say `--with=all` and it would understand that.

I felt like a separate CLI flag was less magic though.

lovesegfault:
Hm, I guess one cool thing about allowing `--with=all` is you could do things like:

```bash
poetry install --with=all --without='foo,bar'
```

@abn Thoughts?

abn:
I would like to understand how common this use case is.

Another change I was thinking was to allow for a group to extend from another. If we had something like that I wonder if this is still something that makes sense. 

```toml
[tool.poetry.group.foo.dependencies]
pytest = "*" 

[tool.poetry.group.bar]
extends = ["foo"]

[tool.poetry.group.bar.dependencies]
aiohttp = "*"
```

lovesegfault:
> I would like to understand how common this use case is.
> 
> Another change I was thinking was to allow for a group to extend from another. If we had something like that I wonder if this is still something that makes sense.
> 
> ```toml
> [tool.poetry.group.foo.dependencies]
> pytest = "*" 
> 
> [tool.poetry.group.bar]
> extends = ["foo"]
> 
> [tool.poetry.group.bar.dependencies]
> aiohttp = "*"
> ```

I like that a lot, I'm going to give it a shot :)

courtland:
I'm guessing this is now stale, but FWIW I would like this functionality as well. I have 4-5 groups, and in some cases I always want to install all of them. This would be easier than having to update CI scripts or what not if a group is ever added.

jgangemi:
is this issue dead? has it been succeeded by anything? i could also use this functionality.

tomzx:
I think it's fair to be able to install all groups without having to specify them all. How about extending `--with` to support `--with="*"` which will match all groups?

jgangemi:
that's workable for me. what wound be the stain getting this implemented?

sobolevn:
I also have a use-case for it, sad that this is not implemented.

MischaPanch:
Same for me, any chance to reopen this?

vancromy:
Just to add on here that a classic use case (at least within our org) is to ensure that versions are up to date for all groups. This is done both in a CI/CD pipeline to ensure our lock files are updated with `poetry update --no-interaction --lock --with=all_groups` and also done during development. It's a hassle having to hardcode groups in our CI/CD config files and remembering to keep these up to date. It would be really nice having a handy shortcut for installing/updating all dependency groups.

Issue #5122, opened 2022-01-28
repository.secondary=true + dependency.source broken?


pierresouchay:
- [x] I am on the [latest](https://github.com/python-poetry/poetry/releases/latest) Poetry version.
- [x] I have searched the [issues](https://github.com/python-poetry/poetry/issues) of this repo and believe that this is not a duplicate.
- [x] If an exception occurs when executing a command, I executed it again in debug mode (`-vvv` option).

- **OS version and name**: MAC OS
- **Poetry version**: 1.1.12
- **Link of a [Gist](https://gist.github.com/pierresouchay/60f3e8b9e20ee0d23f0e4604c7a73019) with the contents of your pyproject.toml file**: https://gist.github.com/pierresouchay/60f3e8b9e20ee0d23f0e4604c7a73019

## Issue

Following a discussion on #4920

When creating a source called "my-source" with `secondary=true`, I was expecting the repository to be hit under those conditions:

* if a given package is not present on pypi
* source=my-source is set on a dependency

This is actually not the case, the repository my-source is actually hit each time for every package, for instance when doing a poetry update.

This is causing a few issues if "my-source" is very slow for instance, as time increase on my local setup from 5s to 120s (a project with lots of dependencies).

Is there is a way to use "my-source" ONLY for a few packages?

Here is the output of `poetry update -vvv` with the pyproject.toml from https://gist.github.com/pierresouchay/60f3e8b9e20ee0d23f0e4604c7a73019 👍 

```
poetry update -vvv
Using virtualenv: <REDACTED>
Updating dependencies
Resolving dependencies...
   1: fact: multi-sources is 0.1.0
   1: derived: multi-sources
   1: fact: multi-sources depends on requests (^2.27.1)
   1: fact: multi-sources depends on changelog (^0.5.7)
   1: selecting multi-sources (0.1.0)
   1: derived: changelog (>=0.5.7,<0.6.0)
   1: derived: requests (>=2.27.1,<3.0.0)
my-source: 1 packages found for changelog >=0.5.7,<0.6.0
   1: selecting changelog (0.5.7)
my-source: 1 packages found for requests >=2.27.1,<3.0.0
   1: fact: requests (2.27.1) depends on urllib3 (>=1.21.1,<1.27)
   1: fact: requests (2.27.1) depends on certifi (>=2017.4.17)
   1: fact: requests (2.27.1) depends on charset-normalizer (>=2.0.0,<2.1.0)
[...]

Issue #4704, opened 2021-11-03
Adding a source even with secondary true take precedence


Ierezell:
- [X] I am on the [latest](https://github.com/python-poetry/poetry/releases/latest) Poetry version.
- [X] I have searched the [issues](https://github.com/python-poetry/poetry/issues) of this repo and believe that this is not a duplicate.
- [X] If an exception occurs when executing a command, I executed it again in debug mode (`-vvv` option).

- **OS version and name**: Linux 5.14.12 & 5.16.0
- **Poetry version**: 1.1.11 & 1.2.0a2

## Issue
Adding a source like
```
[[tool.poetry.source]]                                                          
    name = "pytorch"                                                                
    url = "https://download.pytorch.org/whl/cu111"                                    
    secondary = true   
```
Works perfectly well when installing torch with 
```
torch = { version = "=1.10.0+cpu", source = "pytorch"}    
```

but when trying to add `mypy` with poetry add `poetry add mypy` then I got a : 

```
403 Client Error: Forbidden for url: https://download.pytorch.org/whl/cpu/mypy
```

Which obviously doesn't work as mypy is not in the pytorch package repository....
I don't get why pypi is not the default and "Pytorch" just as a fallback. 

Thanks in advance for any help,
Have a great day.

mxab:
I think I'm experiencing the same issue

gregorybchris:
I can confirm that all versions `1.1.0` through `1.1.11` have this same issue. Downgrading to `1.0.10` is a temporary workaround.

mosauter:
Can Confirm this issue is still present in `1.1.12`

Ierezell:
Hello there :) 

Sorry to bump but is there any news about that ? 

It's really breaking all poetry for my workflow (I need pytorch cu11 which needs to be downloaded from their source). 
I'm willing to contribute if needed. 

Have a great day !

Ierezell:
For those who are still struggling..... 

Without sources, without any of the fancy stuff, just plan dependency declaration.
You can declare manually the wheels : 

```shell
torch = [
    { url="https://download.pytorch.org/whl/cu113/torch-1.10.1%2Bcu113-cp37-cp37m-linux_x86_64.whl", python=">=3.7,<3.8", markers="sys_platform == 'linux'"},
[...]

PR #5442, opened 2022-04-12
add prefetching of index in PEP503 repositories

tgolsson:
# Pull Request Check List

Resolves: #4885, partially

- [x] Added **tests** for changed code.
- [x] Updated **documentation** for changed code.

# Description

This PR adds a new keyword `indexed` when defining a LegacyRepository (e.g. PEP503) source. The `indexed` keyword enables the use of a prefetched and cached index, which will limit the amount of unnecessary calls. 

Currently, if one configures a secondary repository for Poetry it'll get queried for all dependencies, no matter whether it's default, primary, or secondary. For projects with lots of dependencies (transitive or direct) this leads to a lot of unnecessary calls to a host which can't serve the requested package. One such case is using GPU-based packages from https://download.pytorch.org/whl/, where most other packages should be served from Pypi. This leads to user confusion due to error messages; and takes a *lot* of time.

While update-time on a cold cache is dominated by downloading every possible GPU package; this PR changes *noop* `poetry update` time from 30-40 seconds to <5 seconds. In total, this repository has 89 dependencies (reported by `poetry show`). 

Also, it removes all errors from querying subpages that don't exist. 

This depends on a PR to `poetry-core`, and thus a release there: https://github.com/python-poetry/poetry-core/pull/323

### Poetry==1.2.0b1

```
===> multitime results
1: poetry update
            Mean        Std.Dev.    Min         Median      Max
real        36.818      2.927       32.470      36.176      44.032
user        3.478       0.040       3.408       3.480       3.568
sys         0.126       0.030       0.075       0.123       0.181
```

### This PR

```
===> multitime results
1: poetry update
            Mean        Std.Dev.    Min         Median      Max
real        4.617       0.211       4.410       4.588       5.165
user        3.362       0.162       3.235       3.301       3.813
sys         0.097       0.027       0.059       0.099       0.151
```

PR #5472, opened 2022-04-21
repository: do not call secondary repositories if package is find in primary ones

ashangit:
# Pull Request Check List

Resolves: #5122 and #4704

### DIFF
--- a/src/poetry/factory.py
+++ b/src/poetry/factory.py
@@ -126,9 +126,9 @@ class Factory(BaseFactory):
         cls, poetry: Poetry, sources: list[dict[str, str]], config: Config, io: IO
     ) -> None:
         for source in sources:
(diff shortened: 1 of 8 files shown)


tgolsson:
I investigated this a while back and suggested #5442 as an alternative fix for at least some of these cases; keeping backwards compatibility which this fix doesn't. #4704 is semi-fixed in 1.20b1 at least and doesn't crash; and my fix is better for that (for PEP503 repos) since it solves the underlying problem of just throwing random requests at webservers.

Also, your version is more complicated than it needs to be, you can look at the secondary index inside the pool to figure out when to break -- this makes the PR much smaller and uses info that Poetry already has 🎉

abn:
This is expected behavior. See https://python-poetry.org/docs/master/repositories/#secondary-package-sources.

PR #5475, opened 2022-04-21
Canonicalized names

dimbleby:
Companion to https://github.com/python-poetry/poetry-core/pull/328


dimbleby:
pushed a commit that should pass the unit tests with either old or new behaviour, allowing https://github.com/python-poetry/poetry-core/pull/328 to be merged.

Then when that eventually gets published we can disallow the old behaviour again.

abn (approved):

PR #5479, opened 2022-04-22
Fix: git subdirectory dependencies fail to resolve with a TypeError on name

skairunner:
# Pull Request Check List

Resolves: #4678

This is a resurrection of https://github.com/python-poetry/poetry/pull/4922, fixing an issue with package name discovery on git repos with subdirectories. All I've done is merge master and clean up some dust. 

It appears that the test written by steved works correctly, as disabling the fix causes it to fail. I also am unsure whether  [_pep517_metadata](https://github.com/python-poetry/poetry/blob/6a35cc41e950d4ff0f1efad649c1d15d689a9ce2/src/poetry/inspection/info.py#L442) is actually related to the bug, I took a cursory look and it seemed unrelated. If the current unit test is unacceptable please tell me.


skairunner:
master branch has resolved this issue

abn:
@skairunner happy for a test case to be added. :)

PR #5480, opened 2022-04-22
Automatically detect local certificates

bollard:
Specifying the `REQUESTS_CA_BUNDLE` is a useful way to include self signed certificates for non-Windows platforms, however on Windows things are a bit trickier. This change proposes to use [`create_default_context()`](https://docs.python.org/3/library/ssl.html#ssl.create_default_context) (which in turn calls, [`load_default_certs()`](https://docs.python.org/3/library/ssl.html#ssl.SSLContext.load_default_certs) from the Python SSL library to include the required certificates for us.

For background, I'm on a Windows machine in a corporate environment trying (and otherwise failing) to `poetry publish` to an internal Artifactory repo. All credit to [this](https://stackoverflow.com/a/50215614/895029) Stack Overflow answer for the code.

Thanks

First time contributing - will have a look at tests and docs over the weekend to get myself more familiar. Hope that's alright.

# Pull Request Check List

Resolves: #issue-number-here

### DIFF
--- a/src/poetry/publishing/uploader.py
+++ b/src/poetry/publishing/uploader.py
@@ -2,6 +2,7 @@ from __future__ import annotations
 
 import hashlib
 import io
+import ssl
 
 from typing import TYPE_CHECKING
 from typing import Any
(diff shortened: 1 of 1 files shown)


Darsstar:
Something that should probably be mentioned is whether you tried [python-certifi-win32](https://pypi.org/project/python-certifi-win32/) or not.

Secrus:
Stalled

Issue #5476, opened 2022-04-21
Private repositories broken after bdfdc62


lovesegfault:
- [x] I am on the [latest](https://github.com/python-poetry/poetry/releases/latest) Poetry version.
- [x] I have searched the [issues](https://github.com/python-poetry/poetry/issues) of this repo and believe that this is not a duplicate.
- [x] If an exception occurs when executing a command, I executed it again in debug mode (`-vvv` option).

- **OS version and name**: NixOS
- **Poetry version**: dca0c562c42cb79a494eff982fc8a52c632100f4

## Issue

I was trying to bump our internal Poetry from 5900f3761cd0811c3e717b356b1b877b971c4ff0 to dca0c562c42cb79a494eff982fc8a52c632100f4 when I noticed our private repositories stopped working.

Upon attempting `poetry lock --no-update`, I got the following:
```text
  ValueError

  Package('redacted', '21.11.4+ad7d6ed') is not in list

  at ~/.local/share/pypoetry/venv/lib/python3.8/site-packages/poetry/repositories/legacy_repository.py:111 in package
      107│         Note that this will be cached so the subsequent operations
      108│         should be much faster.
      109│         """
      110│         try:
    → 111│             index = self._packages.index(Package(name, version, version))
      112│ 
      113│             return self._packages[index]
      114│         except ValueError:
      115│             package = super().package(name, version, extras)

The following error occurred when trying to handle this error:


  Stack trace:

  29  ~/.local/share/pypoetry/venv/lib/python3.8/site-packages/cleo/application.py:330 in run
       328│ 
       329│             try:
     → 330│                 exit_code = self._run(io)
       331│             except Exception as e:
       332│                 if not self._catch_exceptions:

  28  ~/.local/share/pypoetry/venv/lib/python3.8/site-packages/poetry/console/application.py:174 in _run
       172│         self._load_plugins(io)
       173│ 
     → 174│         return super()._run(io)
       175│ 
       176│     def _configure_io(self, io: IO) -> None:

  27  ~/.local/share/pypoetry/venv/lib/python3.8/site-packages/cleo/application.py:425 in _run
       423│                 io.set_input(ArgvInput(argv))
       424│ 
     → 425│         exit_code = self._run_command(command, io)
       426│         self._running_command = None
       427│ 

  26  ~/.local/share/pypoetry/venv/lib/python3.8/site-packages/cleo/application.py:467 in _run_command
       465│ 
       466│         if error is not None:
     → 467│             raise error
[...]

PR #5484, opened 2022-04-22
fix(link_sources): don't bail on an invalid version

lovesegfault:
# Pull Request Check List

Resolves: #5476

Registries can have packages with broken versions, which we need to just ignore.
As-is the code will hard-fail when it encounters anything with a version it
cannot parse.

### DIFF
--- a/src/poetry/repositories/link_sources/base.py
+++ b/src/poetry/repositories/link_sources/base.py
@@ -1,6 +1,6 @@
 from __future__ import annotations
 
-import contextlib
+import logging
 import re
 
 from abc import abstractmethod
@@ -17,6 +17,9 @@ from poetry.utils.patterns import wheel_file_re
 
 if TYPE_CHECKING:
     from poetry.core.packages.utils.link import Link
+
+
+logger = logging.getLogger(__name__)
 
 
 class LinkSource:
@@ -46,7 +49,7 @@ class LinkSource:
         for link in self.links:
             pkg = self.link_package_data(link)
 
-            if pkg.name == name and pkg.version and pkg.version not in seen:
+            if pkg and pkg.name == name and pkg.version and pkg.version not in seen:
                 seen.add(pkg.version)
                 yield pkg.version
 
@@ -55,7 +58,7 @@ class LinkSource:
         for link in self.links:
             pkg = self.link_package_data(link)
 
-            if pkg.name and pkg.version:
+            if pkg and pkg.name and pkg.version:
                 yield pkg
 
     @property
@@ -63,7 +66,7 @@ class LinkSource:
     def links(self) -> Iterator[Link]:
         raise NotImplementedError()
 
-    def link_package_data(self, link: Link) -> Package:
+    def link_package_data(self, link: Link) -> Package | None:
         name, version = None, None
         m = wheel_file_re.match(link.filename) or sdist_file_re.match(link.filename)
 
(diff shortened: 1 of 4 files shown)


lovesegfault:
cc. @abn and @radoering :)

src/poetry/repositories/link_sources/base.py
```diff
@@ -79,13 +83,18 @@ def link_package_data(self, link: Link) -> Package:
         with contextlib.suppress(ValueError):
             version = Version.parse(version)
 
-        return Package(name, version, source_url=link.url)
+        try:
+            return Package(name, version, source_url=link.url)
+        except Exception:
```
radoering, line 88:
```suggestion
        except InvalidVersion:
```

radoering (changes requested):
See https://github.com/python-poetry/poetry/issues/5476#issuecomment-1107523221

src/poetry/repositories/link_sources/base.py
lovesegfault: Shouldn't we keep going no matter what exception we get? I think currently, only `InvalidVersion` can be raised, but the semantics of "if _anything_ goes wrong, ignore and continue" seem right to me.

WDYT?

radoering: I think it's debatable.

On the one hand catching `Exception` seems more robust, but on the other hand it's an anti pattern that masks unexpected errors and makes debugging difficult (see [SO answer](https://stackoverflow.com/a/21950), [Real Python article](https://realpython.com/the-most-diabolical-python-antipattern/) or [Pylint Warning](https://pylint.pycqa.org/en/latest/messages/warning/broad-except.html)).

Since the only exception we are expecting is `InvalidVersion` and another exception may disguise something, I tend to being specific.

radoering: By the way, we don't have to pass an invalid version to `Package` at all since we've already tried to parse some lines above. Instead of with `with contextlib.suppress(...)` we can do something like:

```
try:
    version = Version.parse(version)
except InvalidVersion:
    return None
return Package(name, version, source_url=link.url)
```

radoering:
@lovesegfault Just wanted to ask if you think you'll find the time to write the tests in the next one/two weeks? I'd really like to include this fix in the next pre-release to avoid a regression? 😀

lovesegfault:
> @lovesegfault Just wanted to ask if you think you'll find the time to write the tests in the next one/two weeks? I'd really like to include this fix in the next pre-release to avoid a regression? grinning

Hey, I'm currently changing jobs and moving cities, I don't think I'll have time to hack on this for at least another couple weeks.

I can try to apply your suggestion quickly today, but I can't spend more time chasing after the root cause here.

radoering:
Don't bother. If you don't mind I can add some tests myself. I appreciate your discovery and analysis of the issue.

lovesegfault:
> Don't bother. If you don't mind I can add some tests myself. I appreciate your discovery and analysis of the issue.

Please don't hesitate to push to this branch, I don't mind at all :)

abn:
@radoering I was fixing something related; pushed up a test case for this with some minor changes to original.

radoering:
@abn Just added the fix for the `else` path, added some necessary checks, removed some unnecessary checks and added some more narrow unit tests.

abn (approved):

PR #5485, opened 2022-04-23
more typechecking

dimbleby:
Experimenting with the effect of putting a `py.typed` in poetry-core, so that types from poetry-core count for something over here.

That generates lots and lots of new errors, I've picked off a fairly random selection of them here until I got bored.

### DIFF
--- a/src/poetry/repositories/pool.py
+++ b/src/poetry/repositories/pool.py
@@ -1,6 +1,5 @@
 from __future__ import annotations
 
-from contextlib import suppress
 from typing import TYPE_CHECKING
 
 from poetry.repositories.exceptions import PackageNotFound
@@ -136,8 +135,9 @@ class Pool(Repository):
             raise ValueError(f'Repository "{repository}" does not exist.')
 
         if repository is not None and not self._ignore_repository_names:
-            with suppress(PackageNotFound):
-                return self.repository(repository).package(name, version, extras=extras)
+            package = self.repository(repository).package(name, version, extras=extras)
+            if package is not None:
+                return package
         else:
             for repo in self._repositories:
                 try:
@@ -146,8 +146,6 @@ class Pool(Repository):
                     continue
 
                 if package:
-                    self._packages.append(package)
-
                     return package
 
         raise PackageNotFound(f"Package {name} ({version}) not found.")
--- a/src/poetry/repositories/pool.py
+++ b/src/poetry/repositories/pool.py
@@ -135,18 +135,15 @@ class Pool(Repository):
             raise ValueError(f'Repository "{repository}" does not exist.')
 
         if repository is not None and not self._ignore_repository_names:
-            package = self.repository(repository).package(name, version, extras=extras)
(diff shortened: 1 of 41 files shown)


dimbleby:
(errors reported in #5476 would have been revealed by the typechecking, if poetry-core had a py.typed)

dimbleby:
This MR's project of cleaning up errors revealed by having a `py.typed` in `poetry-core` is complete as of the latest commit, with a couple of exceptions:

- obviously there are still quite a few files excluded from mypy checking altogether, I haven't made a dent in those
- I still have these two errors:

```
src/poetry/repositories/link_sources/base.py:82: error: Argument 1 to "Package" has incompatible type "Optional[str]"; expected "str"  [arg-type]
src/poetry/repositories/link_sources/base.py:82: error: Argument 2 to "Package" has incompatible type "Optional[Version]"; expected "Union[str, Version]"  [arg-type]
```

which is #5476, more or less.  cc @abn.

(recent commits have mostly been finding a way around the typechecking when dealing with tomlkit, whose types seem to be... hard to use)

dimbleby:
Although I now see that because pre-commit runs in an incomplete environment - #5279 - it doesn't see all the errors that it should and therefore insists that I a making a redundant cast.  Hmm.

src/poetry/repositories/pool.py
```diff
@@ -136,8 +135,9 @@ def package(
             raise ValueError(f'Repository "{repository}" does not exist.')
 
         if repository is not None and not self._ignore_repository_names:
-            with suppress(PackageNotFound):
-                return self.repository(repository).package(name, version, extras=extras)
```
radoering, line 140: Are you sure that `PackageNotFound` can't be raised? For example `PyPiRepository.package()` calls `get_release_info()`, which can raise `PackageNotFound`.

dimbleby: Ah, good spot.  I had found the method on the base `Repository` class, where it does not raise the exception, but does return `None`.

That's inconsistent with all of the implementations on inherited classes, which do raise the exception and do not return None...

I'll change the base to raise the exception too, that seems the least invasive.

Having said which: in this particular bit of code there really is no point in suppressing the exception - because the effect will only be that it gets raised anyway ten lines later...  I'll tidy that too.

src/poetry/repositories/pool.py
```diff
@@ -135,18 +135,15 @@ def package(
...
-
-                if package:
-                    return package
+            return self.repository(repository).package(name, version, extras=extras)
+
+        for repo in self._repositories:
+            try:
+                package = repo.package(name, version, extras=extras)
+            except PackageNotFound:
+                continue
```
radoering, line 144: Why this change in logic removing the else?

dimbleby: this doesn't change the logic: if we go into the if branch, then we either return or we raise an exception.  The 'else' is now redundant.

radoering (approved):

PR #5486, opened 2022-04-23
remove poetry.packages.project_package

dimbleby:
cf https://github.com/python-poetry/poetry-core/pull/330

so far as I can see, `poetry.packages.project_package.ProjectPackage` exists only to provide `set_version()`, but:
- the only use of that is as a not-very-convincing example of something that a plugin might do
- it breaks the hashing of the ProjectPackage
- whether `set_version()` is a good or a bad thing, it surely belongs in poetry-core's ProjectPackage

So this MR removes the poetry object and replaces the plugin tests and example with a different not-very-convincing action (setting the readme).

If the version-setting function is desired it should be resolved over in poetry-core, I think that this MR is good to merge either way.


neersighted (approved):

david-caro:
This removal is unfortunate! I'm working on generating dynamically the version of a package and this removal makes it impossible.
I'll try to give a look at the code, see if I can get something working if nobody gets it before.

dimbleby:
Ah, sorry about that.  But the old code really was broken in the ways that this MR identifies...

If you need similar function I reckon you'll want a `with_version()` method over in poetry-core, which should take a clone of the underlying package and set the version.  That both puts the method in the right repository, and doesn't break hashing.

PR #5501, opened 2022-04-26
factory: cleanup creation from package

abn:
This change ensures that generated toml file from packages contain all relevant metadata and handles groups correct.

Split from #5450

### DIFF
--- a/src/poetry/factory.py
+++ b/src/poetry/factory.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
+import contextlib
 import logging
 
 from typing import TYPE_CHECKING
@@ -22,10 +23,17 @@ from poetry.plugins.plugin_manager import PluginManager
 from poetry.poetry import Poetry
 
 
+try:
+    from poetry.core.packages.dependency_group import MAIN_GROUP
+except ImportError:
+    MAIN_GROUP = "default"
+
+
 if TYPE_CHECKING:
     from pathlib import Path
 
     from cleo.io.io import IO
+    from poetry.core.packages.package import Package
 
     from poetry.repositories.legacy_repository import LegacyRepository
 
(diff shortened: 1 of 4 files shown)


src/poetry/factory.py
```diff
@@ -215,8 +222,28 @@ def create_pyproject_from_package(cls, package: ProjectPackage, path: Path) -> N
             if len(constraint) == 1 and "version" in constraint:
                 constraint = constraint["version"]
 
-            dependency_section[dep.name] = constraint
+            for group in dep.groups:
+                if group == "default":
```
radoering, line 226:
```suggestion
                if group == MAIN_GROUP:
```

src/poetry/factory.py
```diff
@@ -178,23 +180,28 @@ def create_legacy_repository(
...
+    ) -> TOMLDocument:
         import tomlkit
 
-        from poetry.layouts.layout import POETRY_DEFAULT
-
-        pyproject = tomlkit.loads(POETRY_DEFAULT)
-        content = pyproject["tool"]["poetry"]
+        pyproject = tomlkit.document()
+        pyproject["tool"] = tomlkit.table()
+        pyproject["tool"]._is_super_table = True
```
radoering, line 190:
```suggestion
        pyproject["tool"] = tomlkit.table(is_super_table=True)
```

Analoguous for other super tables.

src/poetry/factory.py
```diff
@@ -178,23 +180,28 @@ def create_legacy_repository(
...
+        content = pyproject["tool"]["poetry"] = tomlkit.table()
 
         content["name"] = package.name
         content["version"] = package.version.text
         content["description"] = package.description
         content["authors"] = package.authors
+        content["license"] = package.license.id if package.license else ""
+        content["classifiers"] = package.classifiers
+        content["readme"] = [*package.readmes]
```
radoering, line 199: What about `maintainers`, `homepage` and so on? Should we consider everything from the [docs](https://python-poetry.org/docs/master/pyproject/)?

radoering (dismissed):
Maybe, there should be a unit test for `create_pyproject_from_package`?

src/poetry/factory.py
abn: Won't work because of a bug in the import order. The poetry-core vendored dependency takes precedence and it's an older version at present. Will have to wait until next core version for this change.

src/poetry/factory.py
abn: Probably should eh? Will work it in.

src/poetry/factory.py
```diff
@@ -205,23 +212,50 @@ def create_legacy_repository(
...
+    def create_pyproject_from_package(
+        cls, package: Package, path: Path | None = None
+    ) -> TOMLDocument:
         import tomlkit
 
-        from poetry.layouts.layout import POETRY_DEFAULT
-
-        pyproject: dict[str, Any] = tomlkit.loads(POETRY_DEFAULT)
-        content = pyproject["tool"]["poetry"]
+        pyproject = tomlkit.document()
```
dimbleby, line 220: This is losing typing fixes.  Please consider running mypy at least locally in the poetry virtual environment - and also merging or otherwise resolving #5279.

```
src/poetry/factory.py:222: error: Item "Item" of "Union[Item, Container]" has no attribute "_is_super_table"  [union-attr]
src/poetry/factory.py:222: error: Item "Container" of "Union[Item, Container]" has no attribute "_is_super_table"  [union-attr]
src/poetry/factory.py:223: error: Unsupported target for indexed assignment ("Union[Item, Container]")  [index]
src/poetry/factory.py:253: error: "Item" has no attribute "append"  [attr-defined]
src/poetry/factory.py:286: error: "Item" has no attribute "_is_super_table"  [attr-defined]
src/poetry/factory.py:288: error: Unsupported right operand type for in ("Item")  [operator]
src/poetry/factory.py:289: error: Unsupported target for indexed assignment ("Item")  [index]
src/poetry/factory.py:290: error: Value of type "Item" is not indexable  [index]
src/poetry/factory.py:292: error: Value of type "Item" is not indexable  [index]
src/poetry/factory.py:293: error: Value of type "Item" is not indexable  [index]
src/poetry/factory.py:295: error: Value of type "Item" is not indexable  [index]
```

src/poetry/factory.py
```diff
@@ -205,23 +212,54 @@ def create_legacy_repository(
...
+        content["readme"] = []
+
+        for readme in package.readmes:
+            try:
+                # since packages resolve readme paths, un-resolve it
+                readme_posix_path = readme.relative_to(package.root_dir).as_posix()
+            except ValueError:
+                # we use a try/except since `is_relative_to` is not available in
+                # Python < 3.9
+                readme_posix_path = readme.as_posix()
```
radoering, line 255: That's true, but `is_relative_to` is not used here, only `relative_to`, which is available. It's even used in a test that did not fail. 😉

However, at least according to the type hints `package.root_dir` can be None. Thus, that should be handled somehow.

radoering (changes requested):
Is [extras](https://python-poetry.org/docs/master/pyproject/#extras) still missing?

src/poetry/factory.py
abn: Obviously :) it's there because `if path.is_relative_to()` seemed better and I had to replace that in a previous commit. And don't want people repeating my eagerness ;)

abn:
> Is [extras](https://python-poetry.org/docs/master/pyproject/#extras) still missing?

Even added a test now :D

radoering (approved):

PR #5503, opened 2022-04-26
application: set global log handlers

abn:
This change sets log handlers to use the same ones used by the application.

For example, logging for repository 401/403 looks like this today.

```console
Resolving dependencies... (0.6s)<debug>torch:</debug> Authorization error accessing https://download.pytorch.org/whl/cu113/httpx/
```

With this fix, it looks like this.

```console
Resolving dependencies... (0.6s)torch: Authorization error accessing https://download.pytorch.org/whl/cu113/httpx/
```

### DIFF
--- a/src/poetry/console/application.py
+++ b/src/poetry/console/application.py
@@ -232,23 +232,30 @@ class Application(BaseApplication):
         handler = IOHandler(io)
         handler.setFormatter(IOFormatter())
 
+        level = logging.WARNING
+
+        if io.is_debug():
+            level = logging.DEBUG
+        elif io.is_very_verbose() or io.is_verbose():
+            level = logging.INFO
+
+        logging.basicConfig(level=level, handlers=[handler])
+
         for name in loggers:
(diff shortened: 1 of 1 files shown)


src/poetry/console/application.py
```diff
@@ -232,23 +232,27 @@ def register_command_loggers(
...
             # The builders loggers are special and we can actually
             # start at the INFO level.
             if logger.name.startswith("poetry.core.masonry.builders"):
-                level = logging.INFO
-
-            if io.is_debug():
-                level = logging.DEBUG
-            elif io.is_very_verbose() or io.is_verbose():
-                level = logging.INFO
+                _level = logging.INFO
```
radoering, line 253: The log level for `"poetry.core.masonry.builders"` is always `INFO` now. Before it could also be `DEBUG`. Is this intended?

abn: Good catch fixing.

radoering (approved):

Issue #4231, opened 2021-06-28
pytorch and poetry


yonimedhub:
- [X ] I am on the [latest](https://github.com/python-poetry/poetry/releases/latest) Poetry version.
- [X ] I have searched the [issues](https://github.com/python-poetry/poetry/issues) of this repo and believe that this is not a duplicate.
- [ X] If an exception occurs when executing a command, I executed it again in debug mode (`-vvv` option).

- **OS version and name**: docker on linux with cuda
- **Poetry version**: 1.1.7
- **Link of a [Gist](https://gist.github.com/) with the contents of your pyproject.toml file**: <below

## Issue

pytorch is now pep503 compliant (https://github.com/pytorch/pytorch/issues/25639#issuecomment-861707149) but I still can't add torch and torchvision.
I've tried using this:
```
torch = {url = "https://download.pytorch.org/whl/cu111/torch-1.9.0%2Bcu111-cp38-cp38-linux_x86_64.whl"}
torchvision = {url = "https://download.pytorch.org/whl/cu111/torchvision-0.10.0%2Bcu111-cp38-cp38-linux_x86_64.whl"}
```
but got this:
```
Updating dependencies
Resolving dependencies... (326.0s)

  SolverProblemError

  Because torchvision (0.10.0+cu111) depends on torch (1.9.0)
   and  depends on torch (1.9.0+cu111), torchvision is forbidden.
  So, because  depends on torchvision (0.10.0+cu111), version solving failed.
```

and if I try using 
```
 [[tool.poetry.source]]
 name = "torch_rep"
 url = "https://download.pytorch.org/whl/cu111/"
```
I end up disabling the pypi and can't download anything else...

Piroro-hs:
```toml
[tool.poetry.dependencies]
torch = { version = "=1.9.0+cu111", source = "pytorch" }
torchvision = { version = "=0.10.0+cu111", source = "pytorch" }

[[tool.poetry.source]]
name = "pytorch"
url = "https://download.pytorch.org/whl/cu111/"
secondary = true
```

This works for me (`poetry install` is so slow due to #4035, though).

yonimedhub:
this is my `pyproject.toml`:
```
[tool.poetry]
name = ""
version = "0.1.0"
description = ""
authors = ["Your Name <you@example.com>"]

[tool.poetry.dependencies]
python = "3.8.5"
numpy = "^1.21.0"
matplotlib = "^3.4.2"
[...]

PR #5509, opened 2022-04-27
repositories: add support for PEP 658

abn:
This change allows Poetry to make use of PEP 503 "simple" API repositories that implement PEP 658 for core metadata.

Requires: https://github.com/python-poetry/poetry-core/pull/333
Relates-to: https://github.com/python-poetry/poetry/issues/4231#issuecomment-1111400536


.github/workflows/main.yml
```diff
@@ -96,7 +96,7 @@ jobs:
         with:
           path: poetry-plugin-export
           repository: python-poetry/poetry-plugin-export
-          ref: refs/tags/${{ steps.poetry-plugin-export-version.outputs.version }}
+          ref: main
```
abn, line 99: This is to avoid integration test failures due to poetry-core from the future.

pyproject.toml
```diff
@@ -34,7 +34,7 @@ generate-setup-file = false
 [tool.poetry.dependencies]
 python = "^3.7"
 
-poetry-core = "^1.1.0a7"
+poetry-core = {git = "https://github.com/abn/poetry-core", branch = "util-link-pep-658"}
```
abn, line 37: Testing poetry-core changes.

Secrus:
Updated and rebased to current master

src/poetry/repositories/http_repository.py
```diff
@@ -98,10 +99,29 @@ def _get_info_from_sdist(self, url: str) -> PackageInfo:
...
+    def _get_info_from_urls(
+        self,
+        urls: dict[str, list[str]],
+        metadata: dict[str, pkginfo.Distribution] | None = None,
+    ) -> PackageInfo:
+        metadata = metadata or {}
         # Prefer to read data from wheels: this is faster and more reliable
-        wheels = urls.get("bdist_wheel")
-        if wheels:
+        if wheels := urls.get("bdist_wheel"):
```
ralbertazzi, line 124: As far as I got it, we risk falling back to package inspection when:
- A high priority package (wheel) does not have PEP 658 metadata
- A low priority package (sdist) does have PEP 658 metadata

In this case we would get the sdist metadata from the repository but then we wouldn't use it.

I would prefer a more aggressive strategy that favours PEP 658 over package inspection. Something like (in this order):
1. PEP 658 metadata of wheel
2. PEP 658 metadata of sdist
3. wheel inspection
4. sdist inspection

While currently it feels more like:
1. PEP 658 metadata of wheel
2. wheel inspection
3. PEP 658 of sdist
4. sdist inspection

One way you could achieve this is by reverting changes to `_get_info_from_urls` (which would still act on packages only) and inside `_links_to_data` something like:
```python
if metadata:
    # We got PEP 658 metadata from at least one package.
    # Either take one randomly or use some fancier logic (wheels over sdist)
    info = _get_info_from_metadata(metadata)
else:
    info = _get_info_from_urls(urls)
```

src/poetry/repositories/http_repository.py
```diff
@@ -211,11 +245,39 @@ def _links_to_data(self, links: list[Link], data: PackageInfo) -> dict[str, Any]
                 f' "{data.version}"'
             )
         urls = defaultdict(list)
+        metadata = {}
```
ralbertazzi, line 248: Could you add explicit typing? It makes the code easier to follow
```suggestion
        metadata: dict[str, pkginfo.Distribution] = {}
```

src/poetry/repositories/http_repository.py
ralbertazzi: Possible network optimization in case we go for the "random" approach: as soon as we got PEP 658 metadata from one url, stop querying others as we don't need anything else

src/poetry/repositories/http_repository.py
```diff
@@ -211,11 +245,39 @@ def _links_to_data(self, links: list[Link], data: PackageInfo) -> dict[str, Any]
...
+                            (
+                                f"Metadata file hash ({metadata_hash}) does not match"
+                                f" expected hash ({link.metadata_hash})."
+                            ),
+                            level="warning",
+                        )
+
+                    distribution.parse(response.content)
+                    metadata[link.url] = distribution
+                except requests.HTTPError:
```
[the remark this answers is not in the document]
radoering: This is not the fallback, if the repository provides no metadata, it's the fallback if the repository says it provides metadata but there's a bad link or something like that? Not sure if we should fail, but I'd probably print at least a warning.

src/poetry/repositories/http_repository.py
```diff
@@ -211,11 +245,39 @@ def _links_to_data(self, links: list[Link], data: PackageInfo) -> dict[str, Any]
...
+                    assert link.metadata_url is not None
+                    response = self.session.get(link.metadata_url)
+                    distribution = pkginfo.Distribution()
+                    assert link.metadata_hash_name is not None
+                    metadata_hash = getattr(hashlib, link.metadata_hash_name)(
+                        response.text.encode()
+                    ).hexdigest()
+
+                    if metadata_hash != link.metadata_hash:
+                        self._log(
```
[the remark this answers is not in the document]
radoering: I absolutely agree that we should not use metadata with a wrong hash. One thing to consider: The hash is not mandatory. PEP 658 ("should" not "must"):

> The repository SHOULD provide the hash of the Core Metadata file

Thus, we probably should handle repositories that do not provide metadata.

src/poetry/repositories/http_repository.py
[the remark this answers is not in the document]
ralbertazzi: >  it's the fallback if the repository says it provides metadata but there's a bad link or something like that

Yes, I got that. I would still prefer failing loudly and investigating exactly what's happening.

1. If it's a network issue, it's likely to present again when fetching the package
2. If it's a repository issue then it shouldn't be Poetry responsibility to fix its behaviour.
3. It feels consistent with all other places where we make HTTP requests, in which there's no explicit catch of HTTP exceptions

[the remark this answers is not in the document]
radoering: > Yes, I got that.

That was more like a question to make sure, I'm right. (I have not looked at it in detail.) 😀

> 2\. If it's a repository issue then it shouldn't be Poetry responsibility to fix its behaviour.

Strictly speaking, yes. However, sometimes it's nicer to be robust.

On the other side: As long as we don't know about any widely used broken repositories, your proposal should be fine.

dimbleby:
I see that pypi recently started supporting offering up metadata separately from wheels - and I wouldn't mind betting that it's the only major repository that does so.  Kind of funny if pypi is the one repository for which poetry does _not_ try using that metadata.

Of course support for pypi can be added later so this is not anything against this MR; it just amuses me.

(Using the metadata from pypi could be a fairly big win, I think, for those packages where the JSON API does not give the metadata that poetry wants.  The JSON API has the problem that when it reports an empty list of dependencies, it could mean _either_ "this package has no dependencies" or "I don't know the answer"; so poetry has to download and inspect a potentially large distribution file.  Being able instead to download a small metadata file should make a meaningful different to resolution time in some cases)

Secrus:
> I see that pypi recently started supporting offering up metadata separately from wheels - and I wouldn't mind betting that it's the only major repository that does so. Kind of funny if pypi is the one repository for which poetry does _not_ try using that metadata.
> 
> Of course support for pypi can be added later so this is not anything against this MR; it just amuses me.
> 
> (Using the metadata from pypi could be a fairly big win, I think, for those packages where the JSON API does not give the metadata that poetry wants. The JSON API has the problem that when it reports an empty list of dependencies, it could mean _either_ "this package has no dependencies" or "I don't know the answer"; so poetry has to download and inspect a potentially large distribution file. Being able instead to download a small metadata file should make a meaningful different to resolution time in some cases)

Yeah, it was already raised on Discord, I am working on adding that.

src/poetry/repositories/http_repository.py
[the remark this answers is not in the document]
radoering: changed from `debug` to `warning`

src/poetry/repositories/http_repository.py
```diff
@@ -211,11 +245,39 @@ def _links_to_data(self, links: list[Link], data: PackageInfo) -> dict[str, Any]
...
         for link in links:
             if link.yanked and not data.yanked:
                 # drop yanked files unless the entire release is yanked
                 continue
+            if link.has_metadata:
+                try:
+                    assert link.metadata_url is not None
+                    response = self.session.get(link.metadata_url)
+                    distribution = pkginfo.Distribution()
+                    assert link.metadata_hash_name is not None
```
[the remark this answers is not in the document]
radoering: not relevant anymore (because hashes are optional)

src/poetry/repositories/http_repository.py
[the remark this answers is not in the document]
radoering: Hashes are optional now, but if there is a hash and it does not match, a warning will be printed and the metadata will not be used.

src/poetry/repositories/http_repository.py
radoering: I think we can keep it as is for now. The current approach has the advantage that it does not change our logic (which wheel's metadata we chose) so we get the same results no matter if we can download metadata or if we have to download the wheel/sdist.

Further, I suppose that in 99 % of use cases it's not relevant because either all files of a release have PEP 658 metadata or none of them. (I may be wrong, but in that case we can adjust it later.)

src/poetry/repositories/link_sources/json.py
```diff
@@ -28,7 +28,28 @@ def _link_cache(self) -> LinkCache:
...
+                if metadata_key in file:
+                    metadata_value = file[metadata_key]
+                    if metadata_value and isinstance(metadata_value, dict):
+                        # The interface of poetry.core.packages.utils.link.Link
+                        # is currently limited to strings with one hash.
+                        if sha256 := metadata_value.get("sha256"):
+                            metadata = f"sha256={sha256}"
+                        else:
+                            key, value = next(iter(metadata_value.items()))
+                            metadata = f"{key}={value}"
```
radoering, line 45: This can be improved with python-poetry/poetry-core#664

radoering:
I fixed the parsing of `data-dist-info-metadata` for special cases, added support for PEP 714 (renaming to `data-core-metadata`), added tests and applied some review feedback. (Separate commits to make it traceable for now. I suppose in the end we'll squash it anyway.)

Further, I added PEP 691 support as fallback for PyPI (last commit). We could make it a separate PR, but I think the change is still small enough especially when squashed because there's some refactoring of the previous commits included.

dimbleby:
A couple of doubts.  
- if I read this right, poetry is downloading all of the metadatas even though it is definitely going to use at most one of them?  Seems wasteful
- I was thinking about this PEP last week and I'm not sure when it is going to be helpful for poetry...

What I mean is: I expect that on pypi the json API almost always agrees with one of the metadatas.  So if poetry has decided that the json data doesn't contain any requires-dist - then likely neither will any of the metadatas, and it is going to have to download a whole distribution anyway.  ie also inspecting the metadata is an extra step that probably won't help.

The exceptional case where this can help is for packages that (i) do have dependencies (ii) uploaded their sdist first.  (Because this hits the warehouse oddity in which the json api is populated from the first uploaded package).  Then the json api will show no requires-dist, but metadata from one of the wheels will fill that gap without downloading the whole distribution.

However this case is rare: the popular CI tools and workflows upload wheels first.  (dulwich is a [counterexample](https://github.com/jelmer/dulwich/issues/975))

All of which is I guess a long-winded way of asking whether you have any evidence about the practical effect of this?  Is it a lot of extra round trips to pypi for no benefit, or does it save lots of wheel downloads.  And since the answer is likely "a bit of both" - how does that balance work out?

radoering:
> All of which is I guess a long-winded way of asking whether you have any evidence about the practical effect of this? Is it a lot of extra round trips to pypi for no benefit, or does it save lots of wheel downloads? And since the answer is likely "a bit of both" - how does that balance work out?

Those are some very good points. I haven't done any performance tests yet but I completely agree that it is necessary.

> if I read this right, poetry is downloading all of the metadatas even though it is definitely going to use at most one of them? Seems wasteful

I overlooked that. Metadata fetching should definitely be lazy.

> I expect that on pypi the json API almost always agrees with one of the metadatas. So if poetry has decided that the json data doesn't contain any requires-dist - then likely neither will any of the metadatas, and it is going to have to download a whole distribution anyway. ie separately inspecting the metadata is an extra step that probably won't help.

IIUC the huge difference is that we don't trust the (non-standard) json API if requires-dist is null, but no requires-dist in the metadatas means that there are no dependencies for sure so that we do not have to download the distributions. Or did I misunderstand the PEP? I suppose the current implementation will not download any distributions if the relevant metadatas say that there are no dependencies.

dimbleby:
I think you are right about trusting the metadata when it says "no dependencies" - and I agree that this should be a big win.

(We don't trust sdists when their PKG-INFO says "no dependencies" and I had misremembered / confused myself).

src/poetry/repositories/http_repository.py
```diff
@@ -85,21 +85,76 @@ def _cached_or_downloaded_file(self, link: Link) -> Iterator[Path]:
...
+                        return None
+
+                distribution.parse(response.content)
+                return PackageInfo(
+                    name=distribution.name,
+                    version=distribution.version,
+                    summary=distribution.summary,
+                    requires_dist=list(distribution.requires_dist),
+                    requires_python=distribution.requires_python,
+                )
```
radoering, line 127: TODO: Use `packaging.metadata` instead of `pkginfo`. See `PackageInfo.from_wheel_metadata()` in #8815

radoering:
blocked by #8940

radoering:
This should be ready now. I'll probably check again later this week but in case anyone wants to take a look...

radoering (approved):

abn:
@Secrus @radoering @dimbleby thanks for taking this over :tada:

Issue #520, opened 2018-10-18
Poetry fails to install p4python due to read-only files


epage:
- [x] I am on the [latest](https://github.com/sdispater/poetry/releases/latest) Poetry version.
- [x] I have searched the [issues](https://github.com/sdispater/poetry/issues) of this repo and believe that this is not a duplicate.
- [x] If an exception occurs when executing a command, I executed it again in debug mode (`-vvv` option).

- **OS version and name**: Windows 10
- **Poetry version**: poetry 0.12.2
- **Link of a [pyproject.toml Gist](https://gist.github.com/epage/5f28e3b1e5eeb9a30697363e369a5fde)
- **Link of a [backtrace Gist](https://gist.github.com/epage/2584ad981ff5d9f175d55212b0192987)

## Issue

In digging into the problem, it seems that p4python's files are all marked read-only, causing windows to error when trying to delete them via `shutil.rmtree` which is invoked by poetry's custom temp directory handling.

nagulv:
I face the same issue. 
Python 3.6.6 (v3.6.6:4cf1f54eb7, Jun 27 2018, 03:37:03) [MSC v.1900 64 bit (AMD64)] on win32

`pip install p4python` works:
```
pip install p4python
Collecting p4python
  Using cached https://files.pythonhosted.org/packages/e2/eb/9f948e36dc489299a918e8de305fa598a2759798a6e72ca3d5cbe2dbe862/p4python-2017.2.1615960-cp36-cp36m-win_amd64.whl
Installing collected packages: p4python
Successfully installed p4python-2017.2.1615960

```

Adding p4python via pipenv works as well.

I'm not sure why adding it via poetry retrieves  p4python-2017.2.1615960.tar.gz and then throws the permissioning error:

[https://gist.github.com/nagulv/9be7c38d8796b1b1609d45041eaf5f30](p4python backtrace Gist)

epage:
I decided to dig a little more into why this happens.

So it looks like poetry went to pypi to get the `info.required_dists` but p4python doesn't have that information in pypi's database

- p4python example: https://pypi.org/pypi/p4python/2018.2.1743033/json
- example of `info.required_dists`: https://pypi.org/pypi/black/json

So poetry looks at all of the wheels.  It will only look at the wheels if
- there is a universal wheel
- there is a platform wheel and no sdist

Because p4python is not universal and has an sdist, the sdist is favored.

[...]

PR #5510, opened 2022-04-28
rename safe_rmtree and ensure remove_directory is used instead of shutil.rmtree

abn:
This change renames `safe_rmtree` into a more appropriate name with  relevant arguments. We ensure that the force behaviour is opt-in and not the default to avoid ambiguity.

This combines the work done in https://github.com/python-poetry/poetry/commit/c12d86b7 and #520.

### DIFF
--- a/src/poetry/installation/executor.py
+++ b/src/poetry/installation/executor.py
@@ -25,7 +25,7 @@ from poetry.utils._compat import decode
 from poetry.utils.authenticator import Authenticator
 from poetry.utils.env import EnvCommandError
 from poetry.utils.helpers import pluralize
-from poetry.utils.helpers import safe_rmtree
+from poetry.utils.helpers import remove_directory
 from poetry.utils.pip import pip_install
 
 
@@ -488,7 +488,7 @@ class Executor:
         if package.source_type == "git":
             src_dir = self._env.path / "src" / package.name
             if src_dir.exists():
-                safe_rmtree(str(src_dir))
+                remove_directory(src_dir, force=True)
 
         try:
             return self.run_pip("uninstall", package.name, "-y")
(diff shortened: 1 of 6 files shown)


src/poetry/installation/executor.py
```diff
@@ -588,7 +588,7 @@ def _install_git(self, operation: Install | Update) -> int:
 
         src_dir = self._env.path / "src" / package.name
         if src_dir.exists():
-            safe_rmtree(str(src_dir))
+            remove_directory(str(src_dir))
```
finswimmer, line 591: Any reason to cast to string? And shouldn't there be a `force=True` just like in the other cases?

src/poetry/installation/pip_installer.py
```diff
@@ -128,7 +128,7 @@ def remove(self, package: Package) -> None:
         if package.source_type == "git":
             src_dir = self._env.path / "src" / package.name
             if src_dir.exists():
-                safe_rmtree(str(src_dir))
+                remove_directory(src_dir, force=True)
```
finswimmer, line 131: Any reason to cast to string?

src/poetry/utils/env.py
```diff
@@ -341,7 +341,7 @@ def remove_distribution_files(self, distribution_name: str) -> list[Path]:
                     file.unlink()
 
             if distribution._path.exists():
-                shutil.rmtree(str(distribution._path))
+                remove_directory(str(distribution._path), force=True)
```
finswimmer, line 344: Any reason to cast to string?

finswimmer (changes requested):

src/poetry/utils/env.py
abn: It's a simple path, which mypy does not like.

src/poetry/installation/pip_installer.py
abn: ?

finswimmer (approved):

Issue #5516, opened 2022-04-28
Installing JAX with gpu/tpu support using poetry


pablo2909:
- [x] I have searched the [issues](https://github.com/python-poetry/poetry/issues) of this repo and believe that this is not a duplicate.
- [x] I have searched the [documentation](https://python-poetry.org/docs/) and believe that my question is not covered.

## Issue

Hi everyone, 

I am installing JAX using poetry. I run the command `poetry add jax` and it works fine but this installs the cpu version, as expected. To install the gpu/tpu version of JAX the documentation indicates that I have to run:

`pip install --upgrade "jax[cuda]" -f https://storage.googleapis.com/jax-releases/jax_releases.html`
`pip install "jax[tpu]>=0.2.16" -f https://storage.googleapis.com/jax-releases/libtpu_releases.html`

I understand I could run this specific command in my environment, but if I do this, I believe, it is not handled well by poetry. Is there a more `poetry` way of installing JAX for gpu/tpu ?

Thank you for the help :)

abn:
Poetry only supports PyPI and PEP 503 simple API repositories (nexus, gitlab packages, artifactory etc). An example similar to jax is pytorch. The latter recently started providing a PEP 503 repository making things easier.

Note that this still has annoyances see https://github.com/python-poetry/poetry/issues/4231#issuecomment-1111400536.

Supporting "find-links" style sources is something that will probably happen in the future. However, note that poetry will have to download all files listed to fetch hashes and inspect metadata.

All that said, in the mean time you provide url dependencies pointing to the wheels you want. Ideally with platform markers?

Hope this helps.

abn:
Can you please try the fix at #5517?

### Using pipx
```sh
pipx install --suffix=@5517 'poetry @ git+https://github.com/python-poetry/poetry.git@refs/pull/5517/head'
```

### Using a container (podman | docker)
```sh
podman run --rm -i --entrypoint bash docker.io/python:3.10 <<EOF
set -xe
python -m pip install -q git+https://github.com/python-poetry/poetry.git@refs/pull/5517/head
poetry new foobar
pushd foobar
poetry source add jax https://storage.googleapis.com/jax-releases/jax_releases.html
poetry add -vvv --lock --source jax jaxlib
cat poetry.lock
EOF
```

pablo2909:
Thank you for writing that ! 

I'm not sure how pipx works. If I run this command without a container, what would happen to my installed version of `poetry` ?
Thank you for the help

[...]

PR #5517, opened 2022-04-29
add initial support for single page repositories (sources)

abn:
This change allows users to make use of find-links style single page
sources that do not present a PEP 503 simple API repository.

Resolves: #5516

### Example Configuration
```toml
[[tool.poetry.source]]
name = "jax"
url = "https://storage.googleapis.com/jax-releases/jax_releases.html"
```
### Using pipx
```sh
pipx install --suffix=@5517 'poetry @ git+https://github.com/python-poetry/poetry.git@refs/pull/5517/head'
```

```sh
podman run --rm -i --entrypoint bash docker.io/python:3.10 <<EOF
set -xe
python -m pip install -q git+https://github.com/python-poetry/poetry.git@refs/pull/5517/head
poetry new foobar
pushd foobar
poetry source add jax https://storage.googleapis.com/jax-releases/jax_releases.html
poetry add -vvv --lock --source jax jaxlib
cat poetry.lock
EOF
```

### DIFF
--- a/src/poetry/factory.py
+++ b/src/poetry/factory.py
@@ -159,7 +159,7 @@ class Factory(BaseFactory):
             logger.debug("Disabling source caches")
 
         for source in sources:
-            repository = cls.create_legacy_repository(
+            repository = cls.create_package_source(
                 source, config, disable_cache=disable_cache
             )
             is_default = bool(source.get("default", False))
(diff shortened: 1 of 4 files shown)


abn:
Change was verified as working in #5516.

neersighted (commented):
I feel like the name of `create_legacy_repository` might be worth changing, but otherwise this LGTM.

abn:
> I feel like the name of `create_legacy_repository` might be worth changing, but otherwise this LGTM.

Sounds like a good plan.

neersighted (approved):

PR #5518, opened 2022-04-29
improve http request handling for sources and support multiple paths on same netloc

abn:
This change refactors HTTP repository source implementations. The following changes have been made.

- CacheControl cache now lives within Authenticator.
- Authenticator manages unique sessions for individual netloc.
- CacheControl usage now respects disable cache parameter in repos.
- Certificate and authentication logic is now managed solely within
  Authenticator for source repositories taking advantage of recent
  enhancements.
- Support same netloc with multiple path's in Authenticator.
- Clean up and unifiy redundant logic in credential access.

These changes should allow for better handling of cases like those described in #3041. Additionally, this forms the foundation for unifying HTTP specific logic within the code base and possibly allowing for migration of requests etc. if/when required.

Additionally, also cleans up some type hinting issues pertaining to http repo only attributes.

Co-Authors: @Darsstar @agni-sairent

Closes: #5108
Closes: #5463

### DIFF
--- a/src/poetry/utils/authenticator.py
+++ b/src/poetry/utils/authenticator.py
@@ -1,13 +1,15 @@
 from __future__ import annotations
 
 import contextlib
+import dataclasses
+import functools
 import logging
 import time
 import urllib.parse
 
+from os.path import commonprefix
 from typing import TYPE_CHECKING
 from typing import Any
-from typing import Iterator
 
 import requests
 import requests.auth
@@ -20,6 +22,7 @@ from poetry.exceptions import PoetryException
 from poetry.locations import REPOSITORY_CACHE_DIR
 from poetry.utils.helpers import get_cert
 from poetry.utils.helpers import get_client_cert
+from poetry.utils.password_manager import HTTPAuthCredential
 from poetry.utils.password_manager import PasswordManager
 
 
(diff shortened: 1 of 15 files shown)


abn:
@Darsstar :cake: Thank you for testing that much appreciated and for the #5463. 

@agni-sairent would you mind if I pulled in your changes into this PR so its all tidy? I am guessing the use case things like gitlab etc where there are project specific credentials?

agni-sairent:
@abn go for it! I don't mind at all :smile: 
You're exactly right, our use case is multiple gitlab repos/package registries with different credentials.

abn:
@Darsstar @agni-sairent I have landed that change at https://github.com/abn/poetry/pull/1 for now - might need a couple of iterations. Feedback/testing appreciated. I just reused tests from #5463 - well, whatever was net new.

KundaPanda:
I am looking forward to seeing either one of these PRs merged, so I went forward and tried your version but discovered some odd behaviour.

- If I specify a custom source, which only contains specific packages, and I try to install another package from pypi (in my case, pytz), poetry cannot resolve it. (`Because test depends on pytz (^2022.1), which doesn't exist, version solving failed.`) This issue disappears once I have removed the source from pyproject.toml. After adding custom source(s) back, the issue arises again. All custom repositories are set as secondary, and adding an official pypi source as default did not help either.

- Custom packages need to be specified with source; otherwise, poetry only looks on pypi.org and fails with 404.

src/poetry/utils/authenticator.py
```diff
@@ -100,31 +212,57 @@ def request(self, method: str, url: str, **kwargs: Any) -> requests.Response:
...
+    def post(self, url: str, **kwargs: Any) -> requests.Response:
+        return self.request("post", url, **kwargs)
 
-        credentials: tuple[str | None, str | None] = self._credentials.get(
-            netloc, (None, None)
+    def _get_credentials_for_url(self, url: str) -> HTTPAuthCredential:
+        repository = self.get_repository_config_for_url(url)
+
+        credential = (
+            repository.get_http_credentials(password_manager=self._password_manager)
```
Darsstar, line 239: How about adding `_get_credentiald_for_repository(self, repository)` which caches the result in `self._credentials[repository.url]`?
That should save keyring lookups when downloading multiple deps from the same repo.

tests/utils/test_authenticator.py
```diff
@@ -339,14 +337,182 @@ def test_authenticator_uses_certs_from_config_if_not_provided(
...
+
+    authenticator = Authenticator(config, NullIO())
+
+    authenticator.request(
+        "get", "https://foo.bar/org/_packaging/GUID/pypi/simple/a/1.0.0/a-1.0.0.whl"
+    )
+    request = http.last_request()
+
+    basic_auth = base64.b64encode(b"foo:bar").decode()
+    assert request.headers["Authorization"] == f"Basic {basic_auth}"
```
Darsstar, line 518: This should define multiple repositories. Something like this:
```suggestion
    config.merge(
        {
            "repositories": {
                "alpha": {
                    "url": "https://foo.bar/org-alpha/_packaging/feed/pypi/simple/"
                },
                "beta": {
                    "url": "https://foo.bar/org-beta/_packaging/feed/pypi/simple/"
                },
            },
            "http-basic": {
                "alpha": {"username": "foo", "password": "bar"},
                "beta": {"username": "baz", "password": "qux"},
            },
        }
    )

    authenticator = Authenticator(config, NullIO())

    authenticator.request(
        "get",
        "https://foo.bar/org-alpha/_packaging/GUID/pypi/simple/a/1.0.0/a-1.0.0.whl",
    )
    request = http.last_request()

    basic_auth = base64.b64encode(b"foo:bar").decode()
    assert request.headers["Authorization"] == f"Basic {basic_auth}"

    authenticator.request(
        "get",
        "https://foo.bar/org-beta/_packaging/GUID/pypi/simple/b/1.0.0/a-1.0.0.whl",
    )
    request = http.last_request()

    basic_auth = base64.b64encode(b"baz:qux").decode()
    assert request.headers["Authorization"] == f"Basic {basic_auth}"

```

You could do a second request to the aplha repo if you want

Darsstar (changes requested):
Almost there! Lots of nice refactoring!

Darsstar (dismissed):

neersighted (approved):

Issue #1243, opened 2019-07-23
Issue with installing GitPython on Windows 10 / Python 3.7


aaront:
- [x] I am on the [latest](https://github.com/sdispater/poetry/releases/latest) Poetry version.
- [x] I have searched the [issues](https://github.com/sdispater/poetry/issues) of this repo and believe that this is not a duplicate.
- [x] If an exception occurs when executing a command, I executed it again in debug mode (`-vvv` option).

- **OS version and name**: Windows 10
- **Poetry version**: 0.12.17
- **Link of a [Gist](https://gist.github.com/) with the contents of your pyproject.toml file**: https://gist.github.com/aaront/9c1156015284a29f7c2bb5cd9435fdc8

## Issue
When trying to install the GitPython dependency on Windows 10 (with Python 3.7) in a project, I get the following error:

```
[PermissionError]
[WinError 5] Access is denied: 'C:\\Users\\aaron\\AppData\\Local\\Temp\\tmplx22xv9u\\unpacked\\gitdb-0.6.4\\gitdb\\test\\fixtures\\objects\\88\\8401851f15db0eed60eb1bc29dec5ddcace911'
```

(I've logged the output of `poetry install -vvv` in the gist: https://gist.github.com/aaront/9c1156015284a29f7c2bb5cd9435fdc8 )

Repeated attempts to install the package yield the same result (even attempted on 2 different machines). Also attempted to install the `gitdb` dependency itself first, to the same issue.

Installing this inside Ubuntu/WSL works without issues:

```
Creating virtualenv gitypython-test-py3.7 in /home/aaron/.cache/pypoetry/virtualenvs
Updating dependencies
Resolving dependencies... (1.1s)

Writing lock file


Package operations: 14 installs, 0 updates, 0 removals

  - Installing zipp (0.5.2)
  - Installing importlib-metadata (0.18)
  - Installing smmap2 (2.0.5)
  - Installing atomicwrites (1.3.0)
  - Installing attrs (19.1.0)
  - Installing ddt (1.2.1)
  - Installing gitdb (0.6.4)
  - Installing gitdb2 (2.0.5)
  - Installing more-itertools (7.2.0)
  - Installing pluggy (0.12.0)
  - Installing py (1.8.0)
  - Installing six (1.12.0)
  - Installing gitpython (2.1.12)
  - Installing pytest (3.10.1)
  - Installing gitypython-test (0.1.0)
```

aaront:
[...]

PR #5522, opened 2022-04-30
retire temporary_directory()

dimbleby:
removing poetry's own `temporary_directory()` implementation, prefer `tempfile.TemporaryDirectory()`.

There are already examples of `TemporaryDirectory()` in the codebase so this isn't breaking new ground.  I assume the other was legacy of python2 support or somesuch.

### DIFF
--- a/src/poetry/utils/helpers.py
+++ b/src/poetry/utils/helpers.py
@@ -7,12 +7,10 @@ import stat
 import tempfile
 
 from collections.abc import Mapping
-from contextlib import contextmanager
 from pathlib import Path
 from typing import TYPE_CHECKING
 from typing import Any
 from typing import Callable
-from typing import Iterator
 
 
 if TYPE_CHECKING:
@@ -31,15 +29,6 @@ def canonicalize_name(name: str) -> str:
 
 def module_name(name: str) -> str:
     return canonicalize_name(name).replace(".", "_").replace("-", "_")
-
-
-@contextmanager
-def temporary_directory(*args: Any, **kwargs: Any) -> Iterator[str]:
-    name = tempfile.mkdtemp(*args, **kwargs)
-
-    yield name
-
-    remove_directory(name, force=True)
 
 
 def get_cert(config: Config, repository_name: str) -> Path | None:
@@ -114,6 +103,7 @@ def get_package_version_display_string(
     package: Package, root: Path | None = None
 ) -> str:
     if package.source_type in ["file", "directory"] and root:
+        assert package.source_url is not None
         path = Path(os.path.relpath(package.source_url, root.as_posix())).as_posix()
         return f"{package.version} {path}"
 
--- a/src/poetry/inspection/info.py
+++ b/src/poetry/inspection/info.py
@@ -7,6 +7,7 @@ import tarfile
 import zipfile
 
 from pathlib import Path
+from tempfile import TemporaryDirectory
 from typing import TYPE_CHECKING
 from typing import Iterator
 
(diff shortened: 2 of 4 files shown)


src/poetry/utils/helpers.py
```diff
@@ -33,15 +31,6 @@ def module_name(name: str) -> str:
     return canonicalize_name(name).replace(".", "_").replace("-", "_")
 
 
-@contextmanager
-def temporary_directory(*args: Any, **kwargs: Any) -> Iterator[str]:
-    name = tempfile.mkdtemp(*args, **kwargs)
-
-    yield name
-
-    remove_directory(name, force=True)
```
abn, line 42: Note the force implementation in `remove_directory`. The reason why it exists is twofolds.

1. Git directories can be read only causing issues on windows.  #1243
2. Certain sdists can have similar issues. #520.

Will probably need to verify that the temfile implementation handles that well.

dimbleby: I'd have much more confidence in the language implementation of `TemporaryDirectory()` than in a home-grown version when it comes to dealing with edge cases!

And indeed it looks to do all that remove_directory() does and more: https://github.com/python/cpython/blob/e91dee87edcf6dee5dd78053004d76e5f05456d4/Lib/tempfile.py#L808-L836

abn: I agree the system implementation is preferred. However, I'd want to make sure we are not regressing here somehow.

dimbleby: Well, there's a testcase over in poetry-core for this.  So the pipeline from https://github.com/python-poetry/poetry-core/pull/337 should tell us.

dimbleby: Colour me surprised, recent releases behave as expected but Windows and python 3.7 do indeed go wrong.

So I guess this will have to wait until such time as python 3.7 is out of support,

dimbleby: https://bugs.python.org/issue26660

abn: We should add some comments around it and possibly move it into `_compat` to signal it's intent .

neersighted:
I ran into this when ripping out Python 2.7 support -- which is why it got left behind, but without comments evidently. I would suggest making this a wrapper for stdlib version that only uses the home-grown version on Python 3.7. That way the Python version can be grepped in the future and we can robustly test the stdlib version for future inclusion.

neersighted:
Going to merge this in Poetry and link my suggestion in the PR where it actually matters.

neersighted (approved):

dimbleby:
@neersighted I don't think this wanted merging yet - apologies for leaving it in a state where this was not clear

- While `poetry.core.utils.helpers.temporary_directory` still exists, after merging this MR there are no users of it outside of the poetry-core unit tests.
- even though this repository had no unit test proving the point, I suppose it still had a reason for using `temporary_directory` (which is faulty on Windows python 3.7)

I agree that a single implementation of `temporary_directory()` in just one of the repositories would be enough: but that implementation does want using, until we can drop python 3.7

neersighted:
Hmm, I'm not fully convinced that we need it in Poetry given our particular uses of TemporaryDirectory -- can we replicate the failing test on Python 3.7 in Poetry itself? I was rather under the impression that we don't trigger the bug given our usage in this repo.

dimbleby:
I'm afraid that trying to do that is more trouble than I can be bothered with!

I don't know what would have changed to stop poetry from hitting this since all the previous reports.  eg `_get_info_from_wheel()` and `_get_info_from_sdist()` can still potentially download and extract archives containing read-only things.

If you're somehow convinced that it's not possible to this this problem any more then I guess that's fine by me, but I'd think that it's sensible to err on the side of caution, we can tidy up this nonsense in a year or so when dropping support for python 3.7.

PR #5523, opened 2022-04-30
refactor(provider): move code to merge dependencies by constraint to separate method

radoering:
Just a simple refactoring to reduce the length of `Provider.complete_package()`. Move some of its code into a new method `_merge_dependencies_by_constraint` in analogy to the existing method `_merge_dependencies_by_marker`.


branchvincent (approved):
nice!

PR #5524, opened 2022-04-30
mypy: fix "unused type ignore" issue on windows

radoering:
At the moment the mypy pre-commit hooks fails on windows with:

```
src\poetry\utils\appdirs.py:211: error: Unused "type: ignore" comment
```

### DIFF
--- a/src/poetry/utils/appdirs.py
+++ b/src/poetry/utils/appdirs.py
@@ -8,9 +8,6 @@ import os
 import sys
 
 
-WINDOWS = sys.platform.startswith("win") or (sys.platform == "cli" and os.name == "nt")
-
-
 def expanduser(path: str) -> str:
     """
     Expand ~ and ~user constructions.
@@ -44,7 +41,7 @@ def user_cache_dir(appname: str) -> str:
 
     OPINION: This function appends "Cache" to the `CSIDL_LOCAL_APPDATA` value.
     """
-    if WINDOWS:
+    if sys.platform == "win32":
         # Get the base path
         path = os.path.normpath(_get_win_folder("CSIDL_LOCAL_APPDATA"))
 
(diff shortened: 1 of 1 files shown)


dimbleby:
poetry.utils._compat has `WINDOWS = sys.platform == "win32"`.  With that in mind I am pretty suspicious about whether the more elaborate definition in this file is actually useful.

It looks as though `sys.platform == "cli"` was a thing IronPython did, but not since https://github.com/IronLanguages/ironpython3/issues/395, which seems to have gone into their version 3.4.0 - so that surely is not something we need to worry about any more.

Therefore suggest it would be cleaner to change all of the `if WINDOWS` in this file to `if sys.platform == "win32"`, and also put everything from the definition of `_get_win_folder_from_registry()` downwards inside an `if sys.platform == "win32"` block?  That would satisfy mypy I think.

branchvincent:
> it would be cleaner to change all of the `if WINDOWS` in this file to `if sys.platform == "win32"`

agreed, this code originated from [appdirs](https://github.com/ActiveState/appdirs), which has been [superseded](https://github.com/ActiveState/appdirs/issues/170) by [platformdirs](https://github.com/platformdirs/platformdirs) which does exactly that

https://github.com/platformdirs/platformdirs/blob/60fc40c294d90f7687f779487b8c19679fb73255/src/platformdirs/__init__.py#L20-L21

dimbleby:
platformdirs is already in poetry's dependency tree (via virtualenv), perhaps it would be better just to use that directly, rather than maintaining an old copy of it

radoering:
You are probably right but that requires further investigation. The docstring says _"modified to suit our purposes"_. I suppose we should check what modifications were made and if these are obsolete when using `platformdirs`.

For now, I'm satisfied with fixing mypy but feel free to check if `platformdirs` can be used safely and draft a PR.

dimbleby:
https://github.com/python-poetry/poetry/pull/5527

per comment in that MR, this _does_ change the behaviour on MacOS, in a way that seems to be a straight bug fix.  I look forward to learning whether poetry feels able to take such things.

branchvincent (approved):

PR #5527, opened 2022-05-01
Replace poetry.utils.appdirs with platformdirs

dimbleby:
As noted in https://github.com/python-poetry/poetry/pull/5524#issuecomment-1114195782, platformdirs is already in the poetry dependency tree (via virtualenv), so it's not clear why we have our own copy of an old version of it.

This first commit uses the github pipelines to verify that platformdirs is indeed compatible, at least with all the places where we run pipelines.  If that passes I'll follow up to make the changes that actually remove poetry.utils.appdirs


dimbleby:
pipelines reveal that platformdirs moves the config directory on MacOS from `/Users/runner/Library/Application Support/pypoetry` to `/Users/runner/Library/Preferences/pypoetry`.

This seems to be a straight bug fix, but it is a change that will want highlighting in any changelog.  See also discussion in platformdirs at https://github.com/platformdirs/platformdirs/issues/47.

I'll follow up now with the commit that switches to using platformdirs (and actually also another commit that prefers to use `Path` over `str`).

Then y'all can decide whether you actually want to take this fix, or prefer to keep the old code and behaviour.

tests/conftest.py
```diff
@@ -193,11 +193,11 @@ def config(
 
 @pytest.fixture()
 def config_dir(tmp_dir: str) -> str:
```
abn, line 195:
```suggestion
def config_dir(tmp_dir: str) -> Path:
```

src/poetry/locations.py
```diff
@@ -4,21 +4,20 @@
...
+from platformdirs import user_cache_path
+from platformdirs import user_config_path
+from platformdirs import user_data_path
 
 
-CACHE_DIR = user_cache_dir("pypoetry")
-DATA_DIR = user_data_dir("pypoetry")
-CONFIG_DIR = user_config_dir("pypoetry")
+CACHE_DIR = user_cache_path("pypoetry", appauthor=False)
+CONFIG_DIR = user_config_path("pypoetry", appauthor=False, roaming=True)
```
abn, line 13: For this release we might want to issue a deprecation log for osx users and suggest user moves the directory if it already exists.

Something like:

```py

if osx:
    _LEGACY_PATH = CONFIG_DIR.parent.parent.joinpath("Application Support").joinpath("pypoetry")
    
    if _LEGACY_PATH.exists():
        logger.warn("Legacy configuration directory exists at %s, reusing this directory. Consider moving this to %s, as support for legacy directory will be removed in an upcoming release.", _LEGACY_PATH, CONFIG_DIR)
        CONFIG_DIR = _LEGACY_PATH
```

dimbleby: Seems reasonable.  I've pushed a commit that's actually a bit more careful even than that, checking explicitly for `config.toml` or `auth.toml`.  

The reason is that `../ApplicationSupport/pypoetry` remains the correct directory for application data - and while `DATA_DIR` seems unused (and is removed in this MR) I'm cautious of the possibility that it might exist anyway, for non-config purposes..

abn (approved):

Issue #4752, opened 2021-11-14
RFC: Packaging 21+ and dropping support for managing Python 2.7 environments


neersighted:
Time marches on, and the EOL for Python 2.7 is nearly two years in the past. While Poetry 1.2 (which drops support for Python 2.7/3.5) is not yet out of alpha, we have one more additional wrinkle to consider regarding Python 2.7.

Namely, Poetry intended to keep supporting management of Python 2.7 environments, even if it required Python 3.6+ to run. However, due to upstream changes in Packaging that we want to be able to track, it looks like we are going to be unable to fulfill that promise going forward.

Issue #4264 is tracking support for more modern Packaging versions -- to fix this issue/follow upstream, we will have to drop support for Python 2.
PR #4749 implements support for more modern Packaging versions by both bumping the version constraint as well as adapting to changes in upstream's package layout.

Given the time that has gone by since the official end of Python 2.7 support, I'd like to collect feedback on the idea of dropping Python 2.7 management in the next alpha release. To do so, we would simply merge #4749.

Users who still require Python 2.7 support will be able to fall back to the 1.1.x line of releases (currently 1.1.11) which will continue to support Python 2.

gunungpw:
I think it's reasonable for poetry to drop support python 2.7, because lack of man power/maintainer or little interest from user base itself. If no one from maintainer or user want to maintain the part regarding python 2.7, this change was the best decision from the perspective of maintainability for poetry going forward. Also pip itself closing issue regarding python 2 related problem as `won't fix`.

Thanks, maintainers and contributors of Poetry, I love this project.
Hope you all have a nice day

abn:
@neersighted a quick search on source graph says there are around 234 projects out there that reference python 2.7.

https://sourcegraph.com/search?q=context:global+file:pyproject.toml+%28content:%22python+%3D+%5C%22%5E2.7%22+OR+content:%22python+%3D+%5C%22%7E2.7%22+OR+content:%22python+%3D+%5C%22%3E%3D2.7%22%29+-repo:%5Egithub%5C.com/python-poetry/.*+&patternType=literal

Secrus:
IMO let the dead remain dead. Python 2.7 is not supported for some time now and if you ask me, the less support in tooling it has, the more reason for projects to migrate to Python 3.x. Poetry has already a great amount of issues, lets limit them to 3.x only. 

Same as gunungpw, I would like to thank all maintainers and contributors for their work on Poetry, I absolutely love it

[...]

PR #5529, opened 2022-05-01
Remove mentions of Py2.7 from docs

Secrus:
# Pull Request Check List

Resolves: #4752 

As decided in #4752, support for managing python 2.7 environments is removed from `poetry`. This PR implements necessary changes in documentation, tests and error handling (as mentioned [here](https://github.com/python-poetry/poetry/issues/4752#issuecomment-1112816106))


src/poetry/utils/env.py
```diff
@@ -1637,9 +1640,12 @@ def get_pip_command(self, embedded: bool = False) -> list[str]:
         ]
 
     def get_supported_tags(self) -> list[Tag]:
-        output = self.run_python_script(GET_SYS_TAGS)
+        if self.get_version_info() < (3, 0, 0):
+            raise PoetryException("Poetry does not support python 2.7 environments")
```
abn, line 1644: We can use richer text here, even indicating the plugin. Either via a simple console exception or via `logger.error`.

src/poetry/utils/env.py
```diff
@@ -1637,9 +1640,12 @@ def get_pip_command(self, embedded: bool = False) -> list[str]:
...
+        else:
+            output = self.run_python_script(GET_SYS_TAGS)
```
abn, line 1646:
```suggestion

        output = self.run_python_script(GET_SYS_TAGS)
```

src/poetry/utils/env.py
```diff
@@ -1637,9 +1640,12 @@ def get_pip_command(self, embedded: bool = False) -> list[str]:
...
 
-        return [Tag(*t) for t in json.loads(output)]
+            return [Tag(*t) for t in json.loads(output)]
```
abn, line 1648:
```suggestion
        return [Tag(*t) for t in json.loads(output)]
```

src/poetry/utils/env.py
Secrus: By richer text you mean more info (including info about plugin)?

nikiink:
As this is a documentation change env.py should be removed. This file is also outdated in pull request.

Secrus:
> As this is a documentation change env.py should be removed. This file is also outdated in pull request.

The scope of this PR changed in the process and the old title is misleading. I am still working on it.

PR #5531, opened 2022-05-02
enable strictness flags for mypy

dimbleby:
Enable strict mypy checking, and fix errors arising.

### DIFF
--- a/.flake8
+++ b/.flake8
@@ -5,9 +5,6 @@ ban-relative-imports = true
 # flake8-use-fstring: https://github.com/MichaelKim0407/flake8-use-fstring#--percent-greedy-and---format-greedy
 format-greedy = 1
 inline-quotes = double
-# Allow omission of a return type hint for __init__ if at least one argument is annotated
-# used by flake8-annotations
-mypy-init-return = true
 enable-extensions = TC, TC1
 type-checking-exempt-modules = typing, typing-extensions
 eradicate-whitelist-extend = ^-.*;
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -137,6 +137,7 @@ module = [
   'html5lib.*',
   'jsonschema.*',
   'pexpect.*',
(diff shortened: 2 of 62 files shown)


dimbleby:
closer now that #5485 is merged, I think we need
- #5279 so that the pipeline does its typechecking in the proper environment
- poetry-core to be published with py.typed, so that we recognise types from that project

dimbleby:
this looks as though it has got worse ince #5279 was merged, but that's only because we now run mypy on all platforms.  We're now just waiting on publishing a `poetry-core` that includes `py.typed`.

Maybe it's worthwhile updating this MR to treat `poetry-core` as untyped, and then reverse those changes when the next version of that project is published.

dimbleby:
> Maybe it's worthwhile updating this MR to treat poetry-core as untyped

I've done this, in the hope of getting the improved mypy settings merged sooner rather than later.

When the next poetry-core is published, we'll get a handful of warnings about unnecessary `# ignore: type[...]` markers, which can be resolved simply by removing them.

neersighted (approved):
This looks really good, and the un-cargo-culted mypy options are a breath of fresh air. I agree with the incremental approach -- better to get this work in and prevent the need to endlessly rebase than to wait for the perfect one-shot change.

neersighted:
@dimbleby Needs another rebase, sorry!

dimbleby:
More conflicts, and also a new mistake to fix... any chance we can get this merged?

neersighted (approved):
Let's ship it :)

neersighted:
Oh bloody hell more conflicts... Sorry! It looks good to merge as soon as the next rebase is done.

dimbleby:
@neersighted this is not currently conflicted, it should merge cleanly

neersighted:
Ah, looks like my local pull wasn't up to date. Here we go!

PR #5554, opened 2022-05-06
refactor dependency specification parsing logic

abn:
This change moves, cleans up and refactors dependency specification parsing logic from `InitCommand` to `poetry.utils.dependency_specification`. This is done to improve usability and maintainability of this logic.

This has been bugging me for a while now, so decided to clean it up a bit. 

Additionally, with this change, users can now add dependencies using valid PEP 508 strings.

```console
$ poetry add 'cachecontrol[filecache] (>=0.12.9,<0.13.0); python_version >= "3.6"' --lock

Updating dependencies
Resolving dependencies... (0.5s)

Writing lock file
$ cat pyproject.toml | grep -i cachecontrol
CacheControl = {version = ">=0.12.9,<0.13.0", markers = "python_version >= \"3.6\"", extras = ["filecache"]}
```

PR #5558, opened 2022-05-07
factory: reuse dependency specification utils

abn:


abn:
@neersighted since you reviewed #5554

src/poetry/utils/dependency_specification.py
```diff
@@ -9,11 +9,13 @@
 from typing import TYPE_CHECKING
 from typing import Dict
 from typing import List
+from typing import TypeVar
 from typing import Union
 from typing import cast
 
 from poetry.core.packages.dependency import Dependency
 from poetry.core.packages.vcs_dependency import VCSDependency
+from tomlkit.items import InlineTable
```
neersighted, line 18: This is a little gross, but it's no worse than the tight coupling with logic in the factory we had before -- we can always revisit later.

src/poetry/utils/dependency_specification.py
```diff
@@ -22,7 +24,7 @@
     from poetry.utils.env import Env
 
 
-DependencySpec = Dict[str, Union[str, Dict[str, Union[str, bool]], List[str]]]
+DependencySpec = Dict[str, Union[str, bool, Dict[str, Union[str, bool]], List[str]]]
```
neersighted, line 27: I think this might deserve promotion to a full type -- at the very least, how about using the new Union syntax (`foo | bar`) for better readability?

src/poetry/utils/dependency_specification.py
```diff
@@ -167,7 +172,8 @@ def pep508_to_dependency_specification(requirement: str) -> DependencySpec | Non
 
     with contextlib.suppress(ValueError):
         dependency = Dependency.create_from_pep_508(requirement)
-        specification = dependency_to_specification(dependency)
+        specification: DependencySpec = {}
```
neersighted, line 175: Can the type not be declared on the same line, or am I missing something?

src/poetry/utils/dependency_specification.py
abn: There is something weird going on with that, even with annotations imported bot ide and type checks complain about that.

src/poetry/utils/dependency_specification.py
abn: Yeah I am not a but fan of this one, but at present we need that for clarity.  Would have been better if tomlkit's typing was better. Let me know if you have any particular ideas to improve this in this PR.

src/poetry/utils/dependency_specification.py
neersighted: Odd -- my understanding is that the Python runtime won't evaluate anything inside the brackets until you try to access it at runtime -- it's basically a string. It it mypy that complains, or just PyCharm?

src/poetry/utils/dependency_specification.py
abn: Since i am using generics for the factory, mypy complains. This is more to appease the mighty mypy.

src/poetry/utils/dependency_specification.py
abn: Maybe just pycharm; let me double check

neersighted (approved):
All my complaints are minor typing ones -- and mostly about readability. They can always be revisited later, and in no way affect the code quality which is good :)

PR #5565, opened 2022-05-08
doc: improve repository documentation

abn:


docs/repositories.md
```diff
@@ -21,24 +22,282 @@ on PyPI.
...
+[Upload API](https://warehouse.pypa.io/api-reference/legacy.html#upload-api) endpoint for your
+[publishable repository](#publishable-repository).
+
+```bash
+poetry config repositories.foo https://pypi.example.org/legacy/
+```
+
+{{% note %}}
+
+If you need to use a different credential to your [package source](#package-sources), then it is
```
neersighted, line 64:
```suggestion
If you need to use a different credential for your [package source](#package-sources), then it is
```

docs/repositories.md
```diff
@@ -21,24 +22,282 @@ on PyPI.
...
-This will set the url for repository `foo` to `https://foo.bar/simple/`.
+There can be more than one secondary package source.
+
+{{% note %}}
+
+Package sources, even when configured as secondary, will be searched when Poetry looks up packages
+despite the package having been found in the default source or another configured source.
 
-### Configuring credentials
+If you wish to avoid this, you may explicitly specify which source too look in for a particular
```
neersighted, line 169:
```suggestion
If you wish to avoid this, you may explicitly specify which source to search in for a particular
```

docs/repositories.md
```diff
@@ -21,24 +22,282 @@ on PyPI.
...
+poetry source add --secondary foo https://foo.bar/simple/
 ```
 
-This will set the url for repository `foo` to `https://foo.bar/simple/`.
+There can be more than one secondary package source.
+
+{{% note %}}
+
+Package sources, even when configured as secondary, will be searched when Poetry looks up packages
+despite the package having been found in the default source or another configured source.
```
neersighted, line 167:
```suggestion
All package sources (including secondary sources) will be searched during the package lookup process. These network requests will occur for all sources, regardless of if the package is found at one or more sources.
```

docs/repositories.md
```diff
@@ -21,24 +22,282 @@ on PyPI.
...
+
+{{% /note %}}
+
+### Supported Package Sources
+
+#### Python Package Index (PyPI)
+
+Poetry interacts with [PyPI](https://pypi.org) via its
+[JSON API](https://warehouse.pypa.io/api-reference/json.html). This is used to retrieve a requested
+package's versions, metadata, files etc.
```
neersighted, line 184:
```suggestion
package's versions, metadata, files, etc.
```

docs/repositories.md
```diff
@@ -21,24 +22,282 @@ on PyPI.
...
+
+{{% note %}}
+
+If the the package's published metadata is invalid, Poetry will download available bdist/sdist to
```
neersighted, line 188:
```suggestion
If the the package's published metadata is invalid, Poetry will download the available bdist/sdist to
```

docs/repositories.md
```diff
@@ -21,24 +22,282 @@ on PyPI.
...
+
+If any source within a project is configured with `default = true`, The implicit `pypi` source will
+be disabled and not used for any packages.
+
+{{% /warning %}}
+
+#### Simple API Repository
+
+Poetry can fetch and install package dependencies from public or private custom repositories that
+implement the simple API repository as described in [PEP 503](https://peps.python.org/pep-0503/).
```
neersighted, line 216:
```suggestion
implement the simple repository API as described in [PEP 503](https://peps.python.org/pep-0503/).
```

docs/repositories.md
```diff
@@ -120,53 +383,27 @@ poetry config certificates.foo.cert /path/to/ca.pem
...
+## Caches
 
-Now that you can publish to your private repository, you need to be able to
-install dependencies from it.
+Poetry employs multiple caches for package sources in order to improve user experience and avoid duplicate network
+requests.
 
-For that, you have to edit your `pyproject.toml` file, like so
+The first level cache is a [Cache-Control](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Cache-Control)
+cache for almost all HTTP requests.
```
neersighted, line 392:
```suggestion
header for almost all HTTP requests.
```

neersighted (changes requested):

docs/repositories.md
abn: This one is as intended as it is "simple API" repository.

docs/repositories.md
abn: Much better

docs/repositories.md
abn: This should be "based cache" as we are using the cache control package with the requests session.

docs/repositories.md
neersighted: I do think that's a lot less clear, and it doesn't match the title of the PEP. At the very least "simple API repository specification" would make it more clear.

docs/repositories.md
neersighted: Ah, then linking to docs for the header is a bit misleading.

abn, line 392:
```suggestion
header based cache for almost all HTTP requests.
```

neersighted (dismissed):

neersighted:
pre-commit.ci autofix

neersighted (approved):

PR #5570, opened 2022-05-08
chore(inspection): remove type errors

branchvincent:
Removes `poetry.inspection.*` from mypy's ignored list

Relates-to: https://github.com/python-poetry/poetry/issues/5017

### DIFF
--- a/src/poetry/inspection/info.py
+++ b/src/poetry/inspection/info.py
@@ -10,7 +10,11 @@ import zipfile
 from pathlib import Path
 from tempfile import TemporaryDirectory
 from typing import TYPE_CHECKING
+from typing import Any
+from typing import Callable
+from typing import ContextManager
 from typing import Iterator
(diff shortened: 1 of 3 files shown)


src/poetry/inspection/info.py
```diff
@@ -450,7 +457,7 @@ def _get_poetry_package(path: Path) -> ProjectPackage | None:
         return None
 
     @classmethod
-    def _pep517_metadata(cls, path: Path) -> PackageInfo:
+    def _pep517_metadata(cls, path: Path) -> PackageInfo | None:
```
abn, line 460: Is raising exception the same as returning `None`?

branchvincent: Looks like you already addressed this as part of #5601 💯

neersighted (approved):

neersighted (approved):

PR #5581, opened 2022-05-09
improve git http authentication via repository configuration

abn:
**replace Factory.create_config() w/ Config.create()**

Prior to this change when `Config` was initialised for non-command use, user `config.toml` and `auth.toml` files were not loaded. This caused unintended side effects when configuration look up were performed from the `Authenticator` and other parts of the code.

**ensure git repository authn uses exact urls**

Since git repository authentication is a special case of repository configuration, the existing assumptions around path matching do not apply. In order to prevent unexpected behaviour due to similar path matching, git authentication will use exact url matching.

**git: ignore http auth for ssh url**

This change ensures that http-basic auth credentials are only passed to dulwich when the remote url uses http/https schemes.

In addition to the above, it is now ensured that username/password  parameters are not passed through to dulwich unless both username and password are configured explicitly. This is to ensure that dulwich does not bail out if it detects a username in the url (eg: `ssh://git@github.com`).

Relates-to: #5567 (this feature relies on config being available)

### DIFF
--- a/src/poetry/factory.py
+++ b/src/poetry/factory.py
@@ -3,6 +3,7 @@ from __future__ import annotations
 import contextlib
 import logging
 import re
+import warnings
 
 from typing import TYPE_CHECKING
 from typing import Any
@@ -14,14 +15,11 @@ from poetry.core.toml.file import TOMLFile
 from tomlkit.toml_document import TOMLDocument
 
 from poetry.config.config import Config
-from poetry.config.file_config_source import FileConfigSource
-from poetry.locations import CONFIG_DIR
 from poetry.packages.locker import Locker
 from poetry.packages.project_package import ProjectPackage
 from poetry.plugins.plugin import Plugin
 from poetry.plugins.plugin_manager import PluginManager
 from poetry.poetry import Poetry
-from poetry.utils.dependency_specification import dependency_to_specification
 
 
 try:
@@ -65,7 +63,12 @@ class Factory(BaseFactory):  # type: ignore[misc]
         )
 
         # Loading global configuration
-        config = self.create_config(io)
+        with warnings.catch_warnings():
+            # this is preserved to ensure export plugin tests pass in ci,
+            # once poetry-plugin-export version is updated to use one that do not
+            # use Factory.create_config(), this can be safely removed.
+            warnings.filterwarnings("ignore", category=DeprecationWarning)
+            config = self.create_config()
 
         # Loading local configuration
         local_config_file = TOMLFile(base_poetry.file.parent / "poetry.toml")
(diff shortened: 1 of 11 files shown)


.github/workflows/main.yml
```diff
@@ -108,4 +108,7 @@ jobs:
           ref: refs/tags/${{ steps.poetry-plugin-export-version.outputs.version }}
 
       - name: Run pytest (poetry-plugin-export)
-        run: poetry run python -m pytest -p no:sugar -q poetry-plugin-export/tests/
+        run: |
+          # patch test mock for removed method
```
neersighted, line 112: I'd really rather not merge this with the hack present -- is there a way around it?

neersighted (changes requested):
Only had time to glance at one thing for now.

.github/workflows/main.yml
abn: Unfortunately not anything trivial. We test against the test suite of the installed version of the plugin. So, we either have to redo the fix so that create_factory is still the config provide, and wait for a forwards compatible plugin release. 

That said, I don't think the hack is too bad. Especially considering the alternative and since it's not patching anything other than the test mock.

abn: This is what a candidate fix might look like for this. Let me know what you think.

```diff
diff --git a/src/poetry/factory.py b/src/poetry/factory.py
index 3614324b..5d35d000 100644
--- a/src/poetry/factory.py
+++ b/src/poetry/factory.py
@@ -3,6 +3,7 @@ from __future__ import annotations
 import contextlib
 import logging
 import re
+import warnings
 
 from typing import TYPE_CHECKING
 from typing import Any
@@ -62,7 +63,8 @@ class Factory(BaseFactory):  # type: ignore[misc]
         )
 
         # Loading global configuration
-        config = Config.create()
+        with warnings.filterwarnings("ignore", category=DeprecationWarning):
+            config = self.create_config()
 
         # Loading local configuration
         local_config_file = TOMLFile(base_poetry.file.parent / "poetry.toml")
@@ -111,6 +113,16 @@ class Factory(BaseFactory):  # type: ignore[misc]
     def get_package(cls, name: str, version: str) -> ProjectPackage:
         return ProjectPackage(name, version, version)
 
+    @classmethod
+    def create_config(cls, io: IO | None = None) -> Config:
+        if io is not None:
+            logger.debug("Ignoring provided io when creating config.")
+        warnings.warn(
+            "Use of Factory.create_config() is deprecated, use Config.create() instead",
+            DeprecationWarning,
+        )
+        return Config.create()
+
     @classmethod
     def configure_sources(
         cls,
```

neersighted: This looks a lot more palatable to me. Maybe throw a `TODO` comment onto it if you think that would make it easier to remember to remove it -- otherwise that LGMT.

abn: Cool I will apply it then

src/poetry/utils/authenticator.py
```diff
@@ -267,6 +269,19 @@ def _get_credentials_for_url(self, url: str) -> HTTPAuthCredential:
 
         return credential
 
+    def get_credentials_for_git_url(self, url: str) -> HTTPAuthCredential:
+        parsed_url = urllib.parse.urlsplit(url)
+
+        if parsed_url.scheme not in {"http", "https"}:
+            return HTTPAuthCredential()
```
neersighted, line 276: Is there a good reason to return an empty credential over None?

src/poetry/factory.py
```diff
@@ -65,7 +63,12 @@ def create_poetry(
         )
 
         # Loading global configuration
-        config = self.create_config(io)
+        with warnings.catch_warnings():
+            # this is preserved to ensure export plugin tests pass in ci,
+            # once poetry-plugin-export version is updated to use one that do not
+            # use Factory.create_config(), this can be safely removed.
+            warnings.filterwarnings("ignore", category=DeprecationWarning)
+            config = self.create_config()
```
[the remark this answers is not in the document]
abn: Since the export plugin test suite mocks Factory.create_config() to inject config into test poetry instance, we need to use this. Otherwise it's back to sed patch.

src/poetry/utils/authenticator.py
abn: Additional guards on consumers is one reason. And this is a bit more cleaner than returning none here imho. It also keeps interfaces uniform when interacting with the authenticator.

neersighted (approved):

Issue #5526, opened 2022-05-01
Mixing POETRY_HTTP_BASIC_  env vars with CLI args does not work on poetry publish


dszakallas:
- [x] I am on the [latest](https://github.com/python-poetry/poetry/releases/latest) Poetry version.
- [x] I have searched the [issues](https://github.com/python-poetry/poetry/issues) of this repo and believe that this is not a duplicate.
- [x] If an exception occurs when executing a command, I executed it again in debug mode (`-vvv` option).

- **OS version and name**: Debian
- **Poetry version**: 1.1.13

## Issue

`poetry publish` does not consider `--username` when supplied as a standalone CLI argument (i.e without `--password`).

**Env vars only**  :heavy_check_mark: 
```bash
poetry config repositories.myrepo $REPO_URL \
  && POETRY_HTTP_BASIC_MYREPO_USERNAME=$USERNAME \
         POETRY_HTTP_BASIC_MYREPO_PASSWORD=$PASSWORD \
         poetry publish -r myrepo
```

**CLI args only** :heavy_check_mark: 
```bash
poetry config repositories.myrepo $REPO_URL \
  && poetry publish -r myrepo -u $USERNAME -p $PASSWORD
```

**Mixed** ❌
```bash
poetry config repositories.myrepo $POETRY_HTTP_BASIC_MYREPO_PASSWORD=$PASSWORD \
  && poetry publish -r myrepo -u $USERNAME

No suitable keyring backends were found
Using a plaintext file to store and retrieve credentials
Username: 
  RuntimeError
  Aborted
```
This issue is particularly easy to work around, however still got me off-guard, as I expected that environment variables can be mixed with CLI args.

abn:
This should be an easy fix at

https://github.com/python-poetry/poetry/blob/529281e8cc24146585d66585edbd553a8bbc10e5/src/poetry/publishing/publisher.py#L72

```diff
-                    username = auth.username
+                   username = username or auth.username
```

Should do the trick. Leaving this as a good first issue as it will also need some test cases to be added.

PR #5584, opened 2022-05-09
Use env vars if either `--username` or `--password` (or both) is missing

dszakallas:
# Pull Request Check List

Resolves: #5526 

This PR changes authentication logic for publishing:
- If both `username` and `password` is missing, consider PyPI token env var.
- If no PyPI token and either `username` or `password` is missing, consider http basic env var for the missing field.

Previously, env vars were only considered if both CLI options where missing. This PR effectively makes it possible to mix the CLI options with the env vars.

This way both
```
poetry config repositories.myrepo $REPO_URL \
  && POETRY_HTTP_BASIC_MYREPO_PASSWORD=$PASSWORD \
     poetry publish -r myrepo -u $USERNAME

poetry config repositories.myrepo $REPO_URL \
  && POETRY_HTTP_BASIC_MYREPO_USERNAME=$USERNAME \
     poetry publish -r myrepo -p $PASSWORD
```
will work. I am doubtful about whether it is worth supporting the second use case.

### DIFF
--- a/src/poetry/publishing/publisher.py
+++ b/src/poetry/publishing/publisher.py
@@ -54,9 +54,9 @@ class Publisher:
                 raise RuntimeError(f"Repository {repository_name} is not defined")
 
         if not (username and password):
-            # Check if we have a token first
             token = None
             if not (username or password):
+                # If both are missing, check if we have a token first
                 token = self._authenticator.get_pypi_token(repository_name)
 
             if token:
--- a/src/poetry/publishing/publisher.py
+++ b/src/poetry/publishing/publisher.py
@@ -55,7 +55,10 @@ class Publisher:
 
         if not (username and password):
             # Check if we have a token first
-            token = self._authenticator.get_pypi_token(repository_name)
+            token = None
+            if not (username or password):
+                token = self._authenticator.get_pypi_token(repository_name)
+
             if token:
                 logger.debug("Found an API token for %s.", repository_name)
                 username = "__token__"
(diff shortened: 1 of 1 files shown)


src/poetry/publishing/publisher.py
```diff
@@ -57,8 +57,11 @@ def publish(
                 raise RuntimeError(f"Repository {repository_name} is not defined")
 
         if not (username and password):
-            # Check if we have a token first
-            token = self._authenticator.get_pypi_token(repository_name)
+            token = None
+            if not (username or password):
```
abn, line 61: I don't think this is desired.

src/poetry/publishing/publisher.py
```diff
@@ -69,8 +72,8 @@ def publish(
                     logger.debug(
                         f"Found authentication information for {repository_name}."
                     )
-                    username = auth.username
-                    password = auth.password
+                    username = username or auth.username
+                    password = password or auth.password
```
abn, line 76: Unsure if we really have this scenario.

src/poetry/publishing/publisher.py
dszakallas: The original behavior would overwrite any value passed via CLI, if either is missing and a token is found (e.g if a username is passed and a token is found, then it would discard the username). Do we want to consider token auth if there's a username? I am happy to change back to the original behavior, although I find it a bit unintuitive.

src/poetry/publishing/publisher.py
dszakallas: This looks very niche to me as well. Do you think we should not consider this case?

neersighted:
I'm still not sure on this myself, but I would like to see tests (plus dropping the second case) to consider this for merging.

Issue #5599, opened 2022-05-12
Allow `poetry add` to force build dependencies from source


ajatkj:
- [x] I have searched the [issues](https://github.com/python-poetry/poetry/issues) of this repo and believe that this is not a duplicate.
- [x] I have searched the [documentation](https://python-poetry.org/docs/) and believe that my question is not covered.

## Feature Request

Can we have an option in `poetry add` to build a dependency from source instead of just downloading the existing package for the platform?
I am talking something equivalent to `pip3 install --no-binary :all: <package> --ignore-installed`.

Current work-around is to use `poetry run pip3 install ....` and then do `poetry add --lock` to update the lock file. It will be good to have a flag in `add` to do this easily.

abn:
`--no-binary` for add, update or lock does not make sense. The only case where i makes sense is for the install command.

If you need to avoid installation during add you can do the following (with #5600).

```sh
poetry add --lock package
poetry install --no-binary package
```

clintonroy:
If it's not remembered in pyproject.toml, every person building the project will have to be told to do the --no-binary install option though, no?

abn:
Perhaps. However, this is a user preference and not a project preference I believe. I'm unsure if a project level preference for this really makes sense. That would mean we lock only source distribution etc (which is weird when a bdist is available) and further there is no way to put this information in the project's metadata.

ajatkj:
@abn what you say does make sense as this is not exactly a project level preference. The work around to use `poetry install --no-binary` after `poetry add --lock <package>` does the work for me. I will be closing this ticket now. Thanks for everyone's comments.

pmav99:
@abn I am afraid that this can be a project preference and/or even necessity. A common issue with geospatial packages is that they bundle e.g. GEOS inside the wheel. When you need more than one of these packages and the GEOS binaries are different you run into trouble. E.g.: https://github.com/geopandas/geopandas/issues/2355

To replicate with `poetry`:

```
poetry init -n
poetry add shapely pygeos geopandas
python -c 'import geopandas'
```
will result in performance issues (and/or strange [bugs](https://github.com/geopandas/geopandas/issues/2198))
```
/tmp/asdf/.venv/lib/python3.9/site-packages/geopandas/_compat.py:111: UserWarning: The Shapely GEOS version (3.10.2-CAPI-1.16.0) is incompatible with the GEOS version PyGEOS was compiled with (3.10.1-CAPI-1.16.0). Conversions between both will be slow.
[...]

Issue #1316, opened 2019-08-22
How to add a dependency with  --no-binary option


dazza-codes:
- [x] I have searched the [issues](https://github.com/sdispater/poetry/issues) of this repo and believe that this is not a duplicate.
- [x] I have searched the [documentation](https://poetry.eustace.io/docs/) and believe that my question is not covered.

## Question

How is a dependency added so that poetry will run something like this:
```
pip3 install -U 'numpy >=1.17,<2.0' --no-binary numpy
```

What does that look like in the `pyproject.toml` file?

PS.  See a work-around using poetry export in https://github.com/python-poetry/poetry/issues/365#issuecomment-792054069

chipfranzen:
This is still an issue. The spatial library `shapely` requires installing with a `no-binaries` flag on macOS. See this issue: https://github.com/Toblerity/Shapely/issues/416

geckon:
I'm also interested in this.

Eisbrenner:
Cartopy is another library which requires the no binary flag, since I work with geospatial data I hardly want to miss out on it and would really appreciate an implementation into poetry.

As of now, I gave up on poetry and am back to Conda which has working versions of the Cartopy in its registries.

abn:
This is an interesting case. I think the more, elegant solution would be to be able to specify a per-dependency requirement. The issue however, is that this is ususally a PEP 517 frontend task, which generally falls outside the scope of poetry's core features. That said, since we are moving more towards being more independent of tools like `pip` for installation, specifying formats for a dependency to be installed might become more relevant. 

We might need to discuss where we draw the line first, before enabling a feature like this. I do think think this is unavoidable.

For now, a workaround might be to use a url dependency pointing to an sdist along with a platform marker.

sondrelg:
I'm having this same issue.

@abn would you be able to provide a working example of your suggested workaround?

sbreidbach-edr:
We're in the same position with `pygeos` - we've tried setting `PIP_NO_BINARY` as suggested in #365 but still face the same errors.

abn:
> @abn would you be able to provide a working example of your suggested workaround?

```toml
shapely = { url = "https://files.pythonhosted.org/packages/42/f3/0e1bc2c4f15e05e30c6b99322b9ddaa2babb3f43bc7df2698efdc1553439/Shapely-1.7.1.tar.gz" }
```

[...]

Issue #365, opened 2018-08-01
Equivalent of pip install --no-binary


spapanik:
- [x] I have searched the [issues](https://github.com/sdispater/poetry/issues) of this repo and believe that this is not a duplicate.
- [x] I have searched the [documentation](https://poetry.eustace.io/docs/) and believe that my question is not covered.

## Question

How can I specify that I don't want the binary package to be installed, ie. the equivalent of pip install --no-binary?

jambonrose:
I would appreciate some clarity on this front as well. The currently recommended installation of `psycopg2` is:

```
python3 -m pip install psycopg2>=2.7,<2.8 --no-binary psycopg2
```

What is the right way to install `psycopg2` with Poetry?

Thanks!

rectalogic:
This seems to work:

`PIP_NO_BINARY="psycopg2" poetry install`

RevolutionTech:
Is there a solution with Poetry that involves a change to the `pyproject.toml` or something similar?

The team I work on is trying to install `psycopg2` with `--no-binary`, and although I appreciate that your solution works well @rectalogic, it would be ideal if the solution wasn't co-ordinating that all developers on the team add an environment variable whenever executing `poetry install` (or update their local `~./bashrc`, etc.).

janw:
I agree with @RevolutionTech that this should be an option for dependencies in `pyproject.toml`, a `no-binary` flag. If this is still something to be done, I'd be happy to come up with an implementation.

RevolutionTech:
In the interest of full transparency I should mention that this is no longer an urgent need for the team I work on since modern versions of `psycopg2` do not install the binary version by default anyways.

That said, I still think that this would be a really great feature, and I am sure there are other packages besides `psycopg2` where this presents a problem.

b-long:
Related / duplicate: https://github.com/python-poetry/poetry/issues/1316

b-long:
I ran into an issue with the `neuralcoref` library ([github](https://github.com/huggingface/neuralcoref), [pypi](https://pypi.org/project/neuralcoref/)) which was solved by changing my workflow to `PIP_NO_BINARY="neuralcoref" poetry install` .  

A few questions: 
1. Is this currently the best practice?
2. Is there any method of configuring this via pyproject.toml, poetry.lock, or poetry.toml ?
3. Are there any discussions in the Poetry namespace or https://github.com/pypa namespace to follow which might provide an answer to # 1 or # 2 ?

[...]

PR #5600, opened 2022-05-13
install: add --no-binary option

abn:
This changeset introduces a familiar option from `pip` for Poetry's install command. Users can now use `--no-binary :all:|:none:|package` to indicate they do not want to instal lthe binary distribution of packages.

Resolves: #1316
Resolves: #5599
Resolves: #365

### DIFF
--- a/docs/cli.md
+++ b/docs/cli.md
@@ -223,6 +223,7 @@ option is used.
 * `--default`: Only include the main dependencies. (**Deprecated**)
 * `--sync`: Synchronize the environment with the locked packages and the specified groups.
 * `--no-root`: Do not install the root package (your project).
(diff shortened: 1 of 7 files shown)


kasteph (approved):

fredrikaverpil:
@abn this is awesome for `grpcio` 🎉

abn:
@fredrikaverpil be sure to check #5609.

fredrikaverpil:
> @fredrikaverpil be sure to check #5609. 

@abn Ah, I see. Thanks!
Looks like a better implementation to me, as it becomes opt-in. For example a missing ARM build doesn't require building on Intel too by default. 👍

PR #5602, opened 2022-05-13
ci: allow doc builds from forks

abn:


abn:
Based on https://github.com/python-poetry/poetry-website-ci-build-test/pull/1 it seems reasonably to assume that the configuration here is correct. However, since `pull_request_target` runs against the base ref and this is fixing the docs workflow on master, the trigger does not happen as expected until this is merged.

neersighted (approved):

Issue #3855, opened 2021-03-29
Poetry doesn't try public pypi when private pypi included


damienrj:
- [ X ] I am on the [latest](https://github.com/python-poetry/poetry/releases/latest) Poetry version.
- [ X ] I have searched the [issues](https://github.com/python-poetry/poetry/issues) of this repo and believe that this is not a duplicate.
- [ X ] If an exception occurs when executing a command, I executed it again in debug mode (`-vvv` option).

-  macOS : 11.1
- **Poetry version**: 1.1.5

## Issue
When using a private pypi repo, the public repo is no longer being checked.  I can get everything working again by adding

```
[[tool.poetry.source]]
name = "pypi-public"
url = "https://pypi.org/simple/"
```

But the wasn't needed in the past.

sinoroc:
Any details? Is your private repository set as "secondary"? Why do you think the public PyPI is not checked anymore, does your project have dependencies that should be downloaded from the public PyPI?

damienrj:
Seems to be related to https://github.com/python-poetry/poetry/issues/3306 

Yeah, there are packages that are not on our private repository that then fail because it tries to pull from only the private one.  Adding the public one as a source is a work around that unblocked me.

caniko:
> Any details? Is your private repository set as "secondary"? Why do you think the public PyPI is not checked anymore, does your project have dependencies that should be downloaded from the public PyPI?

I have this exact issue, you can see my pyproject.toml file [here](https://github.com/python-poetry/poetry/issues/2339#issuecomment-871199027)

damienrj:
The way we have been solving it is by
```[[tool.poetry.source]]
name = "private"
url = "https://private/simple"
secondary = true


[[tool.poetry.source]]
name = "pypi-public"
url = "https://pypi.org/simple/"
```

Otherwise if you look at the lockfile the source for all packages is our private repo. This puts more work on our private repo because there isn't any reason to check it for the public packages since we are not running a full clone of pypi.

dwyatte:
@damienrj given this is related to https://github.com/python-poetry/poetry/issues/3306, do you want to check to see if the behavior is still present in https://github.com/python-poetry/poetry/releases/tag/1.2.0a1? Seems 3306 (PR https://github.com/python-poetry/poetry/pull/3406) made the release notes there and may be cherry-picked for an upcoming patch release per https://github.com/python-poetry/poetry/pull/4241

jleclanche:
[...]

PR #5605, opened 2022-05-13
Update repositories.md

jonapich:
# Pull Request Check List

Resolves: some discussion in #3855 

### DIFF
--- a/docs/repositories.md
+++ b/docs/repositories.md
@@ -29,7 +29,7 @@ By default, Poetry discovers and installs packages from [PyPI](https://pypi.org)
 install a dependency to your project for a [simple API repository](#simple-api-repository)? Let's
 do it.
 
-First, [configure](#project-configuration) the [package source](#package-source) as a secondary repository to your
+First, [configure](#project-configuration) the [package source](#package-source) as a [secondary package source](#secondary-package-sources) to your
 project.
 
 ```bash
@@ -185,9 +185,9 @@ poetry add --source internal-pypi httpx
 ```
 
 ```toml
-[tool.poetry.requirements]
-httpx = { version = "*", source = "internal-pypi" }
-urllib3 = "*"
+[tool.poetry.dependencies]
+...
+httpx = { version = "^0.22", source = "internal-pypi" }
 
 [[tool.poetry.source]]
 name = "internal-pypi"
--- a/docs/repositories.md
+++ b/docs/repositories.md
@@ -177,8 +177,7 @@ All package sources (including secondary sources) will be searched during the pa
 process. These network requests will occur for all sources, regardless of if the package is
 found at one or more sources.
 
-If you wish to avoid this, you may explicitly specify which source to search in for a particular
-package.
+In order to limit the search for a specific package to a particular package source, you can explicitly specify what source to use.
 
 ```bash
 poetry add --source internal-pypi httpx
--- a/docs/repositories.md
+++ b/docs/repositories.md
@@ -126,7 +126,7 @@ Any package source defined like this takes precedence over [PyPI](https://pypi.o
 
 If you prefer to disable [PyPI](https://pypi.org) completely, you may choose to set one of your package sources to be the [default](#default-package-source).
 
-To enable a package source only for a specific dependency, see [Secondary Package Sources](#secondary-package-sources).
+If you prefer to specify a package source for a specific dependency, see [Secondary Package Sources](#secondary-package-sources).
 
 {{% /note %}}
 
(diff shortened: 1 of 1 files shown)


docs/repositories.md
```diff
@@ -112,9 +112,27 @@ default = false
...
+[[tool.poetry.source]]
+name = "foo"
+url = "https://foo.bar/simple/"
+secondary = true
+```
+
 {{% warning %}}
 
-If package sources are defined for a project, these will take precedence over
+Reminder: If package sources are defined for a project, these will take precedence over
```
abn, line 135: I think this is redundant over the existing warning box.

docs/repositories.md
```diff
@@ -112,9 +112,27 @@ default = false
 secondary = false
 ```
 
+When a package source is defined like this, it takes precedence over [PyPI](https://pypi.org).
```
abn, line 115: This is redundant I feel as this is said just below as well.

docs/repositories.md
```diff
@@ -112,9 +112,27 @@ default = false
...
+```toml
+[tool.poetry.requirements]
+my-lib = { version = "*", source = "foo" }
+urllib3 = "*"
+
+[[tool.poetry.source]]
+name = "foo"
+url = "https://foo.bar/simple/"
+secondary = true
+```
```
abn, line 131: I'd move this below the warning for continuity. And refer to the default section. Rephrase it something like -

```md
{{% note %}}
If you prefer to disable PyPI, you may choose to set one of your package sources to be the [default](#default-package-source).
{{% /note %}}
```

And move the exampe to the secondary package source section maybe?

abn (changes requested):

jonapich:
@abn I changed a few things. I added the `--secondary` to the first example too since that's what the wording was implying. 

However, this part in the secondary source section is basically false:

```
All package sources (including secondary sources) will be searched during the package lookup process. These network requests will occur for all sources, regardless of if the package is found at one or more sources.

If you wish to avoid this, you may explicitly specify which source to search in for a particular package.
```

You do not "avoid this" by adding the source. If you add the source to a package, the other dependencies _still use the secondary sources_. I would rephrase to:

```
You can force a dependency to use a specific package source (exclusively? or it will look it up first? not sure) by specifying which source to search in for a particular package.
```

abn (dismissed):

docs/repositories.md
```diff
@@ -112,23 +119,26 @@ default = false
...
 
-See [Supported Package Sources](#supported-package-sources) for source type specific information.
 
 {{% note %}}
 
-If your package source requires [credentials](#configuring-credentials) or
-[certificates](#certificates), please refer to the relevant sections below.
+If you prefer to disable [PyPI](https://pypi.org) completely, you may choose to set one of your package sources to be the [default](#default-package-source).
+
+To enable a package source only for a specific dependency, see [Secondary Package Sources](#secondary-package-sources).
```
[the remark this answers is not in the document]
jonapich: I think this suggestion makes it unclear what this is about. I rephrased to:

> If you prefer to specify a package source for a specific dependency, see [Secondary Package Sources](#secondary-package-sources).

docs/repositories.md
```diff
@@ -171,7 +181,18 @@ If you wish to avoid this, you may explicitly specify which source to search in
 package.
```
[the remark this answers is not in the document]
neersighted: `source =` should restrict this to only that repository -- if we're doing otherwise, that's a bug.

docs/repositories.md
```diff
@@ -112,23 +119,26 @@ default = false
...
 ```
 
-{{% warning %}}
-
-If package sources are defined for a project, these will take precedence over
-[PyPI](https://pypi.org). If you do not want this to be the case, you should declare **all** package
-sources to be [secondary](#secondary-package-sources).
-
-{{% /warning %}}
+Any package source defined like this takes precedence over [PyPI](https://pypi.org).
```
neersighted, line 122:
```suggestion
Any package source not marked as `secondary` will take precedence over [PyPI](https://pypi.org).
```

docs/repositories.md
```diff
@@ -167,11 +177,21 @@ All package sources (including secondary sources) will be searched during the pa
 process. These network requests will occur for all sources, regardless of if the package is
 found at one or more sources.
 
-If you wish to avoid this, you may explicitly specify which source to search in for a particular
-package.
+In order to limit the search for a specific package to a particular package source, you can explicitly specify what source to use.
```
neersighted, line 180:
```suggestion
In order to limit the search for a specific package to a particular package repository, you can specify the source explicitly. This is strongly suggested for all private packages to avoid dependency confusion attacks.
```

neersighted (changes requested):

docs/repositories.md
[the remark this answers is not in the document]
jonapich: It's difficult for me to assert this, because my private `pypi` is set to redirect to `pypi.org` on missing packages.

Since this closely touches security, there really should be a unit test that asserts it.

neersighted (approved):

PR #5609, opened 2022-05-14
config: introduce `installer.no-binary`

abn:
This change replaces the `--no-binary` option introduced in #5600 as the original implementation could cause inconsistent results when the add, update or lock commands were used.

This implementation makes use of a new configuration `installer.no-binary` to allow for user specification of sdist preference for select packages.

References: #5600 https://github.com/python-poetry/poetry/issues/5599#issuecomment-1126681242
Relates-to: #1316 #5599 #365

### DIFF
--- a/docs/cli.md
+++ b/docs/cli.md
@@ -223,7 +223,6 @@ option is used.
 * `--default`: Only include the main dependencies. (**Deprecated**)
 * `--sync`: Synchronize the environment with the locked packages and the specified groups.
 * `--no-root`: Do not install the root package (your project).
-* `--no-binary`: Do not use binary distributions for packages matching given policy. Use package name to disallow a specific package; or `:all:` to disallow and `:none:` to force binary for all packages.
 * `--dry-run`: Output the operations but do not execute anything (implicitly enables --verbose).
 * `--extras (-E)`: Features to install (multiple values allowed).
 * `--no-dev`: Do not install dev dependencies. (**Deprecated**)
@@ -232,11 +231,6 @@ option is used.
 
 {{% note %}}
 When `--only` is specified, `--with` and `--without` options are ignored.
-{{% /note %}}
-
-{{% note %}}
-The `--no-binary` option will only work with the new installer. For the old installer,
-this is ignored.
 {{% /note %}}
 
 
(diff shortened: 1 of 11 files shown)


abn:
Documentation: https://website-1xp7qvswl-python-poetry.vercel.app/docs/configuration/#installerno-binary

pmav99:
@abn I tested it. As far as I can tell it works great and the docs are clear. Kudos!

As I mentioned in the issue, my only concert is whether`poetry.toml` is the proper place to introduce an option about the installation of specific packages. A possible alternative could be to have an extra section in `pyproject.toml`. E.g.
```toml
[tool.poetry.no-binary]
packages = [
    "shapely",
    "pygeos",
]
```
From a UX point of view, both approaches should be practically equivalent. The `poetry.toml` one has the advantage that the option can be set via the CLI, too. The `pyproject.toml` avoids the proliferation of configuration options in `poetry config`.

That being said, you have a much better understanding of the "spirit" of poetry's configuration so I trust your judgement.

abn:
Right now I think this is the best route as the alternative might set incorrect expectations and cause unintended side-effects. Nothing stopping us from iterating on this. But watch this space :).

neersighted (approved):

sneakers-the-rat:
out of curiosity, can I ask why not have it as an option on a specific dependency? so you could do something like specifying `no-binary` for a particular platform, etc.?

```toml
[tool.poetry.dependencies]
numpy = [
  { version = "^1.20.0", no_binary = true, markers="sys_platform == 'win32'" },
  { version = "^1.20.0", markers="sys_platform != 'win32'" }
]
```

abn:
Part of the reasons are mentioned in https://github.com/python-poetry/poetry/issues/5599#issuecomment-1126684119. I do not think this is completely infeasible, however not something we would want to implement right now. And I am still not convinced this is upto the project to decide as this.

In addition to the above, the solution in this PR also applies to indirect dependencies as well. Because, often times the issues faced by users are to do with a package that are not a dependency of the package itself. Ofcourse, you can add it explicitly to the package. So, the inconsistencies it can introduce and lack of an implementation that addresses all concerns are reasons why the dependency specific configuration is not prefered at this point in time.

rjurney:
@abn this isn't working in 1.2b3 for me...

Issue #5103, opened 2022-01-26
Poetry init -n causes crashes on poetry >= 1.2.0a2


max-wittig:
- [x] I am on the [latest](https://github.com/python-poetry/poetry/releases/latest) Poetry version.
- [x] I have searched the [issues](https://github.com/python-poetry/poetry/issues) of this repo and believe that this is not a duplicate.
- [x] If an exception occurs when executing a command, I executed it again in debug mode (`-vvv` option).

- **OS version and name**: debian:stable docker image (12.1 macOS / Containerd)
- **Poetry version**: 1.2.0a2 | latest master (8d9fdade0f9f241ef62629c693417b59d87445ee)
- **Link of a [Gist](https://gist.github.com/) with the contents of your pyproject.toml file**: no pyproject.toml file

## Issue

1. Run poetry init -n and observe the crash on version `>= 1.2.0a2`

```sh
root@4a73891740bc:/# poetry init -n

  IndexError

  tuple index out of range

  at ~/.local/share/pypoetry/venv/lib/python3.10/site-packages/poetry/layouts/layout.py:91 in get_package_include
       87│ 
       88│     def get_package_include(self) -> Optional["InlineTable"]:
       89│         package = inline_table()
       90│ 
    →  91│         include = self._package_path_relative.parts[0]
       92│         package.append("include", include)
       93│ 
       94│         if self.basedir != Path():
       95│             package.append("from", self.basedir.as_posix())
root@4a73891740bc:/# poetry init -n -vvv

  Stack trace:

  9  ~/.local/share/pypoetry/venv/lib/python3.10/site-packages/cleo/application.py:330 in run
      328│ 
      329│             try:
    → 330│                 exit_code = self._run(io)
      331│             except Exception as e:
      332│                 if not self._catch_exceptions:

  8  ~/.local/share/pypoetry/venv/lib/python3.10/site-packages/poetry/console/application.py:174 in _run
      172│         self._load_plugins(io)
      173│ 
    → 174│         return super()._run(io)
      175│ 
      176│     def _configure_io(self, io: "IO") -> None:

  7  ~/.local/share/pypoetry/venv/lib/python3.10/site-packages/cleo/application.py:425 in _run
      423│                 io.set_input(ArgvInput(argv))
      424│ 
    → 425│         exit_code = self._run_command(command, io)
      426│         self._running_command = None
      427│ 

[...]

PR #5612, opened 2022-05-15
tolerate `poetry init` in `/`

dimbleby:
# Pull Request Check List

Resolves: #5103

Not so easy to add a testcase for this one: I can arrange that we go through the fixed code branch, but then the unit test actually tries to create a project in `/`, which it usually doesn't have permission to do (and would be undesirable even if it did)


neersighted (changes requested):
Throw a comment in explaining why we're checking this (maybe a link to the issue, even) and I think we're good.

neersighted (approved):

PR #5613, opened 2022-05-15
pipeline fix: handle importlib typing at python 3.7

dimbleby:
Kinda ridiculous, but happens in few enough places that it's just about tolerable.

Looks like the types on the backported importlib metadata aren't quite the same as the types on the 3.8-and-onwards version.  So I've written the same code twice but ignoring different errors...

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -127,6 +127,18 @@ module = [
 ]
 ignore_errors = true
 
+# use of importlib-metadata backport at python3.7 makes it impossible to
+# satisfy mypy without some ignores: but we get a different set of ignores at
+# different python versions.
+#
+# <https://github.com/python/mypy/issues/8823>, meanwhile suppress that
+# warning.
+[[tool.mypy.overrides]]
+module = [
+  'poetry.repositories.installed_repository',
+]
+warn_unused_ignores = false
+
 [[tool.mypy.overrides]]
 module = [
   'cachecontrol.*',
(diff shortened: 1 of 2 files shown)


dimbleby:
hmm, that didn't go as well as I'd hoped

Secrus:
I'd say it would be ok to leave it be as it is for now, at least until the core team decides to drop python 3.7 (it still has about 1y till EOL).

dimbleby:
gonna have to do something about the failing pipelines.

_Maybe_ it's ok not to run the mypy checks at py37...

dimbleby:
Different strategy: sprinkle ignores around liberally so that both 3.7 and 3.8 are satisfied; and then suppress the warnings that the 3.7-ignores are not needed at 3.8, and vice versa.

relevant issues: https://github.com/python/mypy/issues/8823, https://github.com/python/importlib_metadata/pull/342

Secrus:
> Different strategy: sprinkle ignores around liberally so that both 3.7 and 3.8 are satisfied; and then suppress the warnings that the 3.7-ignores are not needed at 3.8, and vice versa.
> 
> relevant issues: [python/mypy#8823](https://github.com/python/mypy/issues/8823), [python/importlib_metadata#342](https://github.com/python/importlib_metadata/pull/342)

That looks much better imho. I wonder if there should be extra comment with those ignore comments, just to explain why it is that way, so someone overeager doesn't go and fight to fix it loosing their time. Other than that, it looks good to me.

abn (approved):

abn:
Ideally yes additional comment would be great. But I am going to merge this right now to prevent future PRs from failing and get the master branch passing.

Issue #3735, opened 2021-02-26
`poetry/__init__.py` is provided by both poetry and poetry-core


nchepanov:
- [x] I am on the [latest](https://github.com/python-poetry/poetry/releases/latest) Poetry version.
- [x] I have searched the [issues](https://github.com/python-poetry/poetry/issues) of this repo and believe that this is not a duplicate.
- [x] If an exception occurs when executing a command, I executed it again in debug mode (`-vvv` option).

- **OS version and name**: N/A
- **Poetry version**: 1.1.4
- **Poetry-core version**: 1.0.2
- **Link of a [Gist](https://gist.github.com/) with the contents of your pyproject.toml file**: N.A.

## Issue

It appears that both `poetry-core-1.0.2.tar.gz` and `poetry-1.1.4.tar.gz` attempt to install `poetry/__init__.py`. 

While this may not be a problem now, since `pip` will silently overwrite one file with another https://github.com/pypa/pip/issues/4625

It does become a problem when the two are repackaged via a more strict package management system, e.g. in our internal debian distribution both binary packages would contain `$prefix/lib/python3.9/site-packages/poetry/__init__.py`. Consequently, they cannot be installed with something like `apt-get install`: 

```
.$prefix/lib/python3.9/site-packages/poetry/__init__.py duplicated when installing python3.9-poetry and python3.9-poetry-core.
```

It appears that this can be solved by treating `poetry/` as   [PEP420](https://www.python.org/dev/peps/pep-0420/) implicit namespace by removing `__init__.py` from both the `poetry` and `poetry.core` packages.

This can be done without affecting any users. I'm happy to submit a PR if needed.

finswimmer:
Hello @nchepanov,

the current version of poetry support python2 and python3. Namespace packages were introduced somewhere in python3. This is why we need to use [pkgutil-style namespace packages](https://packaging.python.org/guides/packaging-namespace-packages/#pkgutil-style-namespace-packages).

With the next minor release (1.2), we are dropping python2 support. So I guess we can switch to real namespace packages then.

fin swimmer

neersighted:
Hi @nchepanov, would you like to create that PR you mentioned? `master` is now the head for 1.2 and can drop Python 2 support. Note you'll also need to submit a PR to poetry-core (but you likely already realized that).

nchepanov:
Hi @neersighted I'm afraid I'm not going to have time for this in the near future, thank you for asking!
[...]

PR #5614, opened 2022-05-16
Remove `src/poetry/__init__.py` to convert to namespace package

lafrenierejm:
# Pull Request Check List

Resolves: #3735

- [ ] Added **tests** for changed code. (Not applicable)
- [ ] Updated **documentation** for changed code. (Not applicable)

The corresponding change was made in https://github.com/python-poetry/poetry-core/commit/fa93845ed48f8ba392da3628fd5718a84ebae5d6.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -105,6 +105,7 @@ lines_between_types = 1
 lines_after_imports = 2
 src_paths = ["src", "tests"]
 extend_skip = ["setup.py"]
+known_third_party = ["poetry.core"]
 
 
 [tool.black]
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -117,6 +117,9 @@ force-exclude = '''
 
 [tool.mypy]
 files = "src"
+mypy_path = "src"
+namespace_packages = true
+explicit_package_bases = true
 show_error_codes = true
 strict = true
 enable_error_code = [
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -36,7 +36,6 @@ repos:
           - flake8-bugbear==22.4.25
           - flake8-comprehensions==3.10.0
           - flake8-eradicate==1.2.1
-          - flake8-no-pep420==2.3.0
           - flake8-quotes==3.3.1
           - flake8-simplify==0.19.2
           - flake8-tidy-imports==4.8.0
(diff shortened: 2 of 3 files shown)


lafrenierejm:
Any ideas why pre-commit is failing?

dimbleby:
pre-commit is failing because it explicitly bans implicit-namespack-packages: https://github.com/python-poetry/poetry/blob/f33c9db2bbc083f3147dec0fc171df820baa4390/.pre-commit-config.yaml#L39.

If this MR goes ahead you'll also need to fix things that are broken by implicit namespaces.  mypy for instance, I think you'd want to set all of namespace_packages, explicit_package_bases, mypy_path - all as in poetry-core.

However, since poetry-core has already done this there is already no collision of `__init__.py`.  So perhaps #3735 can simply be closed without further action?

lafrenierejm:
> If this MR goes ahead you'll also need to fix things that are broken by implicit namespaces.  mypy for instance, I think you'd want to set all of namespace_packages, explicit_package_bases, mypy_path - all as in poetry-core.

Thanks for the tip. I have resolved the errors that CI was raising. This PR _should_ now be mergeable if the maintainers so choose.

> However, since poetry-core has already done this - https://github.com/python-poetry/poetry-core/pull/251 - there is already no collision of `__init__.py`.  So perhaps #3735 can simply be closed without further action?

AFAIK you are correct that there's no longer any technical reason this PR would be required. My personal preference would still be to take ahead and make this change so the codebase is a proper PEP420 namespace package rather than implicitly relying on poetry-core's already having done so. If we don't proceed with these changes and a third package is ever added to the `poetry` namespace, the same errors would arise again.

pyproject.toml
```diff
@@ -124,6 +125,9 @@ enable_error_code = [
     "redundant-expr",
     "truthy-bool",
 ]
+namespace_packages = true
+mypy_path = "src"
```
neersighted, line 129: I'd move mypy_path to be next to files, and put namespace_packages and explicit_package_bases right below it. After that, LGTM.

neersighted (changes requested):

pyproject.toml
lafrenierejm: Force-pushed https://github.com/python-poetry/poetry/commit/4e32bade3c76fca133f7a82795f686dedc3bc679 with these changes.

neersighted (approved):

Issue #5611, opened 2022-05-14
Suggest using the `py` launcher when installing Poetry globally on Windows


johnthagen:
- [x] I have searched the [issues](https://github.com/python-poetry/poetry/issues) of this repo and believe that this is not a duplicate.

## Issue

I am starting to use Poetry and would like to try to improve some rough edges I have hit before recommending it to novice users.

On Windows, the [official installation instructions](url) state:

```ps
(Invoke-WebRequest -Uri https://raw.githubusercontent.com/python-poetry/poetry/master/get-poetry.py -UseBasicParsing).Content | python -
```

And similar for the [`master` branch](https://python-poetry.org/docs/master/#installation):

```ps
(Invoke-WebRequest -Uri https://install.python-poetry.org -UseBasicParsing).Content | python -
```

The problem is that the most common way to install/invoke Python on _Windows_ for launching is through the `py` launcher. Adding `python` to the `PATH` is a box that is unchecked by default. This means from an average user on Windows their first experience with trying to install Poetry the recommended way will be failure.

I recommend that the default instructions `| py -` or at the very least mention that you might have to do this prominently.

abn:
Thanks for highlighting this. I'm not too familiar with the windows ecosystem to comment on how relevant this is. 

Might be worth a doc update. A generic note that the python executable could be different with this as an example might be good. Happy to review PRs.

johnthagen:
I asked Steve Dower what his thoughts were on this
- https://discuss.python.org/t/microsoft-store-package-does-not-add-the-py-exe-launcher/538/10?u=johnthagen

johnthagen:
[According to Steve Dower](https://discuss.python.org/t/microsoft-store-package-does-not-add-the-py-exe-launcher/538/11?u=johnthagen), 95% of Windows users download from python.org, so let's target `py` as the primary installation method.

PR #5618, opened 2022-05-16
Update Windows install instructions to target the py launcher

johnthagen:
# Pull Request Check List

Resolves: #5611

### DIFF
--- a/docs/_index.md
+++ b/docs/_index.md
@@ -43,8 +43,13 @@ curl -sSL https://install.python-poetry.org | python3 -
 
 **windows powershell install instructions**
 ```powershell
-(Invoke-WebRequest -Uri https://install.python-poetry.org -UseBasicParsing).Content | python -
+(Invoke-WebRequest -Uri https://install.python-poetry.org -UseBasicParsing).Content | py -
 ```
+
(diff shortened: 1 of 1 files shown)


johnthagen:
@abn Could you review this?

Secrus (approved):

docs/_index.md
```diff
@@ -43,9 +43,14 @@ curl -sSL https://install.python-poetry.org | python3 -
 
 **windows powershell install instructions**
 ```powershell
-(Invoke-WebRequest -Uri https://install.python-poetry.org -UseBasicParsing).Content | python -
+(Invoke-WebRequest -Uri https://install.python-poetry.org -UseBasicParsing).Content | py -
```
abn, line 46: Should it be the other way around? python by default and note to suggest use of py?

Also does Windows have an equivalent of `py || python`? Not even sure if that would work.

abn: Just saw your comment on the issue.

Secrus: `py` is default runner when installing from `python.org`.

abn (approved):

PR #5627, opened 2022-05-17
chore(console): remove type errors

branchvincent:
Removes `poetry.console.commands.init` from mypy's ignored list

Relates-to: https://github.com/python-poetry/poetry/issues/5017

### DIFF
--- a/src/poetry/layouts/layout.py
+++ b/src/poetry/layouts/layout.py
@@ -3,6 +3,7 @@ from __future__ import annotations
 from pathlib import Path
 from typing import TYPE_CHECKING
 from typing import Any
+from typing import Mapping
 
 from tomlkit import dumps
 from tomlkit import inline_table
@@ -50,8 +51,8 @@ class Layout:
         author: str | None = None,
         license: str | None = None,
         python: str = "*",
-        dependencies: dict[str, str] | None = None,
-        dev_dependencies: dict[str, str] | None = None,
+        dependencies: dict[str, str | Mapping[str, Any]] | None = None,
+        dev_dependencies: dict[str, str | Mapping[str, Any]] | None = None,
     ) -> None:
         self._project = canonicalize_name(project).replace(".", "-")
         self._package_path_relative = Path(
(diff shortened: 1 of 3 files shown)


src/poetry/layouts/layout.py
```diff
@@ -50,8 +51,8 @@ def __init__(
         author: str | None = None,
         license: str | None = None,
         python: str = "*",
-        dependencies: dict[str, str] | None = None,
-        dev_dependencies: dict[str, str] | None = None,
+        dependencies: dict[str, str | Mapping[str, Any]] | None = None,
```
Secrus, line 54: why do you use `dict[]` here and `Dict[]` in `init.py` `Requirements` type declaration? shouldn't this be in the same notation?

dimbleby: mypy seems unhappy with the lower-case version when defining a type alias, we have a couple of similar examples here and in poetry-core eg https://github.com/python-poetry/poetry/blob/d0aa6e870772c39c941eff030852e63e63b22965/src/poetry/utils/dependency_specification.py#L27

dimbleby: (wow, what a monster that is...! I'd have settled for `dict[str, Any]` here.)

branchvincent: The difference here is because a type alias is executed at runtime:
```python
name = Dict[str, str]
name = dict[str, str] # syntax error for <3.10
```

whereas a type hint isn't, since we've opted in to [PEP 563](https://peps.python.org/pep-0563/) (with `from __future__ import annotations`):

```python
name: dict[str, str]
name: 'dict[str, str]' # equivalent representation
```

I actually didn't handle this on my first push, so you can see the syntax it caused [here](https://github.com/python-poetry/poetry/runs/6474204828?check_suite_focus=true#step:13:6865)

Secrus (approved):

src/poetry/layouts/layout.py
neersighted: Ah, that makes sense and I technically knew that -- but forgot it after not looking at this for a couple months...

neersighted (approved):

PR #5630, opened 2022-05-18
ci: update to Ubuntu 22.04 and macOS 12

neersighted:
Testing against newer versions will help us keep abreast of any upcoming issues. The older versions can be added as partial matrix entries if desirable, but I do not believe this to be necessary.

Also re-order mypy in the workflow so that type errors fail fast (before tests are run).

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -31,11 +31,11 @@ jobs:
         python-version: ["3.7", "3.8", "3.9", "3.10"]
         include:
           - os: Ubuntu
-            image: ubuntu-latest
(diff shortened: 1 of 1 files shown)


.github/workflows/main.yml
```diff
@@ -81,15 +81,15 @@ jobs:
       - name: Install dependencies
         run: poetry install
 
+      - name: Run mypy
+        run: poetry run mypy
```
branchvincent, line 85: nice! was about to do the same

branchvincent (approved):

Issue #5625, opened 2022-05-17
`poetry show` `--why` option requires `--tree` 


Secrus:
- [x] I have searched the [issues](https://github.com/python-poetry/poetry/issues) of this repo and believe that this is not a duplicate.

## Issue
`--why` option on `poetry show` command is dependent on `--tree` option, but documentation doesn't reflect that. `--why` should be listed as suboption under `--tree`, have note about it in its description or be denoted the same way `--only` is (note box).

tgolsson:
It shouldn't, but I agree I didn't write docs clearly. You should be able to show `--why` for the full table as well; and it'll include it in the table. So it's either package + tree + why or why without either.

Secrus:
@tgolsson can it be assumed, that you will make a PR fixing the docs?

tgolsson:
Yes, sorry! Wanted to make one yesterday but got caught up in other things.

Secrus:
> Yes, sorry! Wanted to make one yesterday but got caught up in other things.

No worries. Just wanted to know whether to expect anything. Take your time

PR #5444, opened 2022-04-12
add `--why` option to `poetry show`

tgolsson:
# Pull Request Check List

Resolves: #5396

- [x] Added **tests** for changed code.
- [x] Updated **documentation** for changed code. (none found for poetry show)

This adds a `--why` option which when combined with `--tree` and a single package starts the tree at the parent level to show why a specific dependency is included. Optimally this'd walk the tree to the root; but that requires a graph-like structure to easily do/filter, and feels out of scope for this PR. 

# Example output

Without `--why`:
```
$ poetry show requests --tree
requests 2.27.1 Python HTTP for Humans.
├── certifi >=2017.4.17
├── charset-normalizer >=2.0.0,<2.1.0
├── idna >=2.5,<4
└── urllib3 >=1.21.1,<1.27
```

With `--why`:
```
$ poetry show requests --tree --why
cachecontrol 0.12.10 httplib2 caching for requests
└── requests *
    ├── certifi >=2017.4.17
    ├── charset-normalizer >=2.0.0,<2.1.0
    ├── idna >=2.5,<4
    └── urllib3 >=1.21.1,<1.27
requests-toolbelt 0.9.1 A utility belt for advanced users of python-requests
└── requests >=2.0.1,<3.0.0
    ├── certifi >=2017.4.17
    ├── charset-normalizer >=2.0.0,<2.1.0
    ├── idna >=2.5,<4
    └── urllib3 >=1.21.1,<1.27
```

On a package with no parent:

```
$ poetry show --tree cachy --why
Package cachy is a direct dependency.
cachy 0.3.0 Cachy provides a simple yet effective caching library.
```

Forgetting to specify package:
```
$ poetry show --tree --why
Error: --why requires specifying a package name.
```

PR #5634, opened 2022-05-18
docs: list `--why` as a suboption of `--tree` in `poetry show`

kosmv:
Hi!
Here's my attempt for #5625

### DIFF
--- a/docs/cli.md
+++ b/docs/cli.md
@@ -444,12 +444,12 @@ required by
 ### Options
 
 * `--without`: The dependency groups to ignore.
-* `--why`: Include reverse dependencies where applicable.
 * `--with`: The optional dependency groups to include.
 * `--only`: The only dependency groups to include.
 * `--default`: Only include the main dependencies. (**Deprecated**)
 * `--no-dev`: Do not list the dev dependencies. (**Deprecated**)
 * `--tree`: List the dependencies as a tree.
+    * `--why`: When listing the tree for a single package, start from parents.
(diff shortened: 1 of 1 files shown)


tgolsson:
See my comment on original issue; the docs are unclear but this is even less correct. :-)

Secrus:
@kosmv thank you for your contribution. 

@tgolsson Can you please assist with clearing up how this should work? its cool feature, but in my testing it was hard to find how it should be used. Thank you in advance.

kosmv:
@tgolsson Oh right, my bad. I didn't understand why `--why` can be used alone.

tgolsson:
@Secrus Unless @kosmv wants to finish this I'll make a PR tonight :-) 

The intended behavior is as follows:

* You can use --why when showing table or when showing a tree for a single package.
* If you use --why for the table it'll include a `from a, b, c` column
* If you use --why for a single-package tree it'll root the tree at the parent packages instead of the queried package.

This comment contains the expected two use-cases: https://github.com/python-poetry/poetry/pull/5444#issuecomment-1098263805.

 The CLI command line is just utterly wrong; I think I had some rebase problems and didn't double-check that well enough. The docs one should probably just specify where it's applicable and not just "when applicable".

kosmv:
@tgolsson Sorry for not reading your original pull request properly. 
My terminal width was reduced, so I couldn't see the `from package-name` column when I tried your feature.

I won't finish this, you can make your PR :)

PR #5635, opened 2022-05-18
more typechecking

dimbleby:
Picking off some type errors that will be exposed when poetry-core with its py.typed is published.

The additional parameter and return value on `EditableBuilder.build()` go unused but are needed to match the parent `Builder` class.  Maybe that suggests that the type signature on the parent is wrong? But probably it just doesn't matter.


Secrus (dismissed):

dimbleby:
I've now made the changes in `poetry.installation`: give or take the usual code review, this should be good to merge

Secrus (approved):

neersighted (approved):

Issue #5633, opened 2022-05-18
Cache clear fails without waiting for input


provinzkraut:
- [ ] I am on the [latest](https://github.com/python-poetry/poetry/releases/latest) Poetry version.
- [x] I have searched the [issues](https://github.com/python-poetry/poetry/issues) of this repo and believe that this is not a duplicate.
- [x] If an exception occurs when executing a command, I executed it again in debug mode (`-vvv` option).

- **OS version and name**: Ubuntu 21.10
- **Poetry version**: 1.2.0b1
- **Link of a [Gist](https://gist.github.com/) with the contents of your pyproject.toml file**: None needed

## Issue

When running `poetry cache clear <cache name> --all` fails with the following error:

```
Delete 2812 entries? (yes/no) [no] 
AttributeError

'NoneType' object has no attribute 'readline'

at ~/.local/share/pypoetry/venv/lib/python3.10/site-packages/cleo/io/inputs/input.py:77 in read_line
     73│         """
     74│         if not self._interactive:
     75│             return default
     76│ 
  →  77│         return self._stream.readline(length)
     78│ 
     79│     def close(self) -> None:
     80│         """
     81│         Closes the input.
```

It does display the `Delete 2812 entries? (yes/no) [no] ` line, however it does not wait for an input and defaults to `no`.

Secrus:
Confirmed. It does the same thing on my machine and master branch version from May 17th

volker48:
Hitting the same issue. Here is the stack trace with verbose logging enabled

```% poetry -vvv cache clear pypi --all
Delete 430 entries? (yes/no) [no]
  Stack trace:

  13  ~/.asdf/installs/python/3.10.4/lib/python3.10/site-packages/cleo/application.py:330 in run
       328│
       329│             try:
     → 330│                 exit_code = self._run(io)
       331│             except Exception as e:
       332│                 if not self._catch_exceptions:

  12  ~/.asdf/installs/python/3.10.4/lib/python3.10/site-packages/poetry/console/application.py:174 in _run
       172│         self._load_plugins(io)
       173│
     → 174│         return super()._run(io)
       175│
       176│     def _configure_io(self, io: IO) -> None:

  11  ~/.asdf/installs/python/3.10.4/lib/python3.10/site-packages/cleo/application.py:425 in _run
       423│                 io.set_input(ArgvInput(argv))
       424│
     → 425│         exit_code = self._run_command(command, io)
       426│         self._running_command = None
       427│

[...]

PR #5647, opened 2022-05-19
Add hook to update io.input.stream to sys.stdin if it is None.

chadac:
# Pull Request Check List

Resolves: #5633 

### DIFF
--- a/src/poetry/console/commands/command.py
+++ b/src/poetry/console/commands/command.py
@@ -24,7 +24,8 @@ class Command(BaseCommand):  # type: ignore[misc]
     def run(self, io: IO) -> int:
         if io.input.stream is None:
             io.input.set_stream(sys.stdin)
-        return super().run(io)
+        status_code: int = super().run(io)
+        return status_code
 
     @property
     def poetry(self) -> Poetry:
--- a/src/poetry/console/commands/command.py
+++ b/src/poetry/console/commands/command.py
@@ -8,8 +8,6 @@ from cleo.exceptions import ValueException
 
 
 if TYPE_CHECKING:
-    from cleo.io.io import IO
-
     from poetry.console.application import Application
     from poetry.poetry import Poetry
 
@@ -18,17 +16,6 @@ class Command(BaseCommand):  # type: ignore[misc]
     loggers: list[str] = []
 
     _poetry: Poetry | None = None
-
-    def run(self, io: IO) -> int:
-        """
-        Temporarily fix for issue where io.input.stream is unset by cleo.
-        """
-        # FIXME: Remove method after upstream merge for cleo
-        # https://github.com/sdispater/cleo/pull/135
-        if io.input.stream is None and self.application._input is not None:
-            io.input.set_stream(self.application._input.stream)
-        status_code: int = super().run(io)
-        return status_code
 
     @property
     def poetry(self) -> Poetry:
--- a/src/poetry/console/commands/command.py
+++ b/src/poetry/console/commands/command.py
@@ -22,6 +22,11 @@ class Command(BaseCommand):  # type: ignore[misc]
     _poetry: Poetry | None = None
 
     def run(self, io: IO) -> int:
+        """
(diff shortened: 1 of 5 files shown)


src/poetry/console/commands/command.py
```diff
@@ -17,6 +21,17 @@ class Command(BaseCommand):  # type: ignore[misc]
...
 
+    def run(self, io: IO) -> int:
+        """
+        Temporarily fix for issue where io.input.stream is unset by cleo.
+        """
+        # FIXME: Remove method after upstream merge for cleo
+        # https://github.com/sdispater/cleo/pull/135
+        if io.input.stream is None:
+            io.input.set_stream(sys.stdin)
+        status_code: int = super().run(io)
```
Secrus, line 32: Is that extra variable needed? can't we just `return super().run(io)`?

chadac: mypy complains about the type for `super.run()` being `Any` if I return it directly, so I wrapped it into an explicit cast here. I could swap to a mypy flag if that works better, though.

Secrus: No, lets leave like this.

Secrus (approved):

Secrus (approved):

abn:
Is this something we can add a meaningful test for?

chadac:
@abn 

I could create a new `test_command.py` with these tests or add them to an existing file:

```
def test_command_with_no_io_stream(mocker: MockerFixture):
    command = Command()
    mocker.patch.object(
        BaseCommand,
        "run",
        return_value=[1]
    )
    io = IO(input=Input(), output=None, error_output=None)
    io.input.set_stream(None)
    command.run(io)
    assert io.input.stream == sys.stdin


def test_command_with_existing_io_stream(mocker: MockerFixture):
    command = Command()
    mocker.patch.object(
        BaseCommand,
        "run",
        return_value=[1]
    )
    io = IO(input=Input(), output=None, error_output=None)
    io.input.set_stream(sys.stdout)
    command.run(io)
    assert io.input.stream == sys.stdout
```

What would y'all recommend?

abn:
I was wondering if we could create a more functional test. Similar to one of the other interactive command test cases.

chadac:
Sure, lemme take a crack at that. There currently aren't any tests for `cache clear` at the moment, I can see if I can get it fully covered in the integration tests.

chadac:
I've gone ahead and added some tests for the `cache clear` command, but incidentally it actually doesn't capture the issue this PR fixes. The main issue is that the `CommandTester` in cleo don't actually trigger the problematic code in `cleo.application.Application:L425`. It's possible that this could be caught by migrating to using `ApplicationTester` instead, although I don't know what y'all think about that.

neersighted:
> I've gone ahead and added some tests for the `cache clear` command, but incidentally it actually doesn't capture the issue this PR fixes. The main issue is that the `CommandTester` in cleo don't actually trigger the problematic code in `cleo.application.Application:L425`. It's possible that this could be caught by migrating to using `ApplicationTester` instead, although I don't know what y'all think about that. It would be a divergence from what most of the other commands do.

I would rather have a good regression test for this, even if it's a bit divergent from the other tests.

chadac:
Sounds good, I've gone ahead and swapped it over. I I did have to add one attribute to the `Application` class in order to ensure that the "original" input object is sufficiently preserved, so that the command code can property specify either sys.stdin or the proper test StringIO stream. I'm open to other ideas, though. It's technically less brittle now as the hardcoded sys.stdin is gone, but technically it should be unneeded when cleo merges in this PR and has the same tests.

branchvincent:
@chadac we plan to have a new `cleo` release soon with your merged fix. Could we update this pr to temporarily depend on cleo's `master` branch to verify your tests also pass here? Then we can update again once it's released

neersighted:
The Cleo release is now in master -- this can just be converted to a regression test only.

neersighted (approved):

PR #5648, opened 2022-05-19
ensure provider respects subdirectory when merging and imrove subdirectory support for vcs deps

abn:


src/poetry/mixology/version_solver.py
```diff
@@ -42,16 +43,23 @@ class DependencyCache:
...
+            tuple[str, str | None, str | None, str | None, str | None],
+            list[DependencyPackage],
         ] = {}
-
-    @functools.lru_cache(maxsize=128)
-    def search_for(self, dependency: Dependency) -> list[DependencyPackage]:
+        self.search_for = functools.lru_cache(
+            maxsize=128
+            if metadata.version("poetry-core")  # type: ignore[no-untyped-call]
+            != "1.1.0a7"
```
abn, line 52: I don't like this but necessary evil since the hash for dependency is incorrect otherwise.

src/poetry/puzzle/provider.py
```diff
@@ -230,7 +236,13 @@ def search_for_vcs(self, dependency: VCSDependency) -> list[Package]:
         Basically, we clone the repository in a temporary directory
         and get the information we need by checking out the specified reference.
         """
-        if dependency in self._deferred_cache:
+        # we ensure subdirectory match here as workaround until poetry-core is updated
+        # to >1.1.0a7
+        if (
+            dependency in self._deferred_cache
+            and self._deferred_cache[dependency].source_subdirectory
```
abn, line 243: This is another case that will be removed once core is upgraded.

Secrus: The same as comment above, please add a `#FIXME` comment.

Secrus (changes requested):

Secrus (dismissed):

radoering:
@abn I took the liberty of adding a small improvement to avoid weird error messages like "Because myapp depends on both demo (1.2.3) and demo (1.2.3), version solving failed."

Otherwise looks fine. 👍 Can be merged if CI passes and the change is OK for you.

radoering (approved):

PR #5650, opened 2022-05-19
poc: rely on poetry to manage pep517 build env

abn:


neersighted:
This looks really promising -- I've always hated the Python-in-strings in that file, and this lets use further re-use and exercise our internal installer while also moving away from strings-as-code.

abn:
The only reason I am not aggressive on this yet, is that this seems to, for some reason, take a longer time per environment. Need to get some metrics on that before I can be really sure though. But there are areas of improvements here.

Secrus:
Quick note, `pypa` is hcanging `pep517` name to `pyproject-hooks` https://github.com/pypa/pyproject-hooks might be worth changing it when this PR will be revived.

PR #5654, opened 2022-05-20
mixology: choose direct ref deps when merging

abn:


Secrus (approved):

src/poetry/mixology/term.py
```diff
@@ -151,21 +151,45 @@ def difference(self, other: Term) -> Term | None:
...
+        if (
+            not compatible
+            and not self._is_direct_reference(self.dependency)
+            and self._is_direct_reference(other)
+        ):
+            # we do this here to indicate direct reference dependencies are compatible
+            # with NVR dependencies
+            return bool(
+                self.dependency.complete_name == other.complete_name
+                and self.dependency.constraint.allows_all(other.constraint)
```
radoering, line 174:
```suggestion
```
Checking the constraint is probably not necessary here because it is done later anyway. (`is_same_package_as()` does not consider the constraint, too.) Although it may not change anything, it seems more suitable not to do it here.

src/poetry/mixology/term.py
```diff
@@ -151,21 +151,45 @@ def difference(self, other: Term) -> Term | None:
...
+            )
+
         return compatible
```
radoering, line 177: That is a place where `is_same_package_as()` is expected to be non-commutative (which seems really bad design to me). We should refactor it further so it still works when we make `is_same_package_as()` commutative in the future. Assuming, we don't need the constraint check here:

By the way, what does NVR stand for?

```suggestion
        return (
            self.dependency.is_root
            or other.is_root
            or other.is_same_package_as(self.dependency)
            or (
                # we do this here to indicate direct reference dependencies are
                # compatible with NVR dependencies
                self.dependency.complete_name == other.complete_name
                and self._is_direct_reference(self.dependency)
                != self._is_direct_reference(other)
            )
        )
```

radoering (changes requested):

dimbleby:
It's a bit unfortunate that it's possible to make commits here that break things when run against the latest `poetry-core`.  

Perhaps, just as `poetry-core` has pipelines that run against latest `poetry`, `poetry` should have pipelines that run against the latest `poetry-core`.

Or perhaps this doesn't happen very often and that's just not worthwhile

src/poetry/mixology/term.py
```diff
@@ -151,21 +151,45 @@ def difference(self, other: Term) -> Term | None:
         """
         return self.intersect(other.inverse)
 
+    @staticmethod
+    def _is_direct_reference(dependency: Dependency) -> bool:
+        return dependency.source_type in ["directory", "file", "url", "git"]
```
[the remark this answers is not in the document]
abn: Yeah, I will raise a new PR after this on core to address the changes I mentioned on discord.

src/poetry/mixology/term.py
abn: I feel that that is incorrect. We do not want `demo (>=1.1.0)` to return compatible for `demo (1.0.0 git)`. I'd rather keep it in here and not rely on the later checks. Unless there is a pressing reason to remove it, I'd much rather keep it for correctness.

src/poetry/mixology/term.py
abn: Ack. 

NVR = Name, Version, Release. Old habbits. :)

src/poetry/mixology/term.py
radoering: But `demo (>=1.1.0)` is already compatible to `demo (<1.0)`. Removing the check allows the simplification in the comment below.

radoering: At the moment, I think this method is not supposed to consider versions. That may seem weird at first glance, but at least it's consistent for now since versions are handled in the callers. It may even be wrong to handle it here that way because the handling depends on if the term is positive or negative.

abn: In that case we should just replace `is_same_package_as` with `self.dependency.complete_name == other.complete_name` right?

abn: This seems to not be the case with the current core. 

I have applied your suggested change for now. We can rework it if required after the next core release.

radoering: `is_same_package_as()` (which does not consider versions) is required in some cases. `complete_name` is not enough:
If both dependencies are direct origin (or legacy?) dependencies, `is_same_package_as()` is required.
If both dependencies are PyPI dependencies, `is_same_package_as()` and `complete_name` are the same (does not matter which one to use).
If one dependency is a direct origin dependency and the other is a PyPI or legacy dependency, `complete_name` must be used instead of `is_same_package_as()`.

I'm only a bit unsure if two legacy dependencies from different sources or a PyPI and a legacy dependency are "compatible" or not. Perhaps, these are rare edge cases that can be revisited later.

radoering (approved):

radoering:
@dimbleby Not sure if it's worth. In some cases it is useful. In other cases it complicates making tests forward compatible. At least, when applying the method that is favoured at the moment as in: https://github.com/python-poetry/poetry/blob/ed7eac3914911ca4e793353841efabe68b80c685/tests/puzzle/test_solver.py#L1387-L1390

That would require a version bump in core and a new compat variable for each such change. On the other side that may even be cleaner than now and we could use `.dev` or `.post` versions for that.

PR #5660, opened 2022-05-22
tests: Update generated PKG-INFO fixture

andersk:
Since commit f5cbba2bca2293b6e7b0fed6fd8b506203c969f4 (#4766), `test_executor_should_write_pep610_url_references_for_git` has been writing a different version of this file than the one checked into the repository, leaving a dirty working tree. Fix it and add a CI test.

(Here’s how the new CI test [fails](https://github.com/andersk/poetry/runs/6540382659) without the fix.)

# Pull Request Check List


branchvincent (approved):
thanks for the fix and the test 😄

Issue #5598, opened 2022-05-12
Incomplete error message in exception


ElricleNecro:
- [x] I am on the [latest](https://github.com/python-poetry/poetry/releases/latest) Poetry version.
- [x] I have searched the [issues](https://github.com/python-poetry/poetry/issues) of this repo and believe that this is not a duplicate.
- [x] If an exception occurs when executing a command, I executed it again in debug mode (`-vvv` option).

- **OS version and name**: MacOS 12.3.1
- **Poetry version**: 1.1.13
- **Link of a [Gist](https://gist.github.com/) with the contents of your pyproject.toml file**: I can't share it (at least the part with the bug).

## Issue
When catching exception, `poetry` does not give any useful information as it discards the original error message. As an example, poetry can't access the keyring on my mac, but it it doesn't tell me why, just:
```python
  KeyRingError

  Unable to retrieve the password for poetry-repository-iktos from the key ring

  at ~/.poetry/lib/poetry/utils/password_manager.py:40 in get_password
       36│         try:
       37│             return keyring.get_password(name, username)
       38│         except (RuntimeError, keyring.errors.KeyringError):
       39│             raise KeyRingError(
    →  40│                 "Unable to retrieve the password for {} from the key ring".format(name)
       41│             )
       42│
       43│     def set_password(self, name, username, password):
       44│         if not self.is_available():
```
without useful information in debug mode. If the message from the original exception was also printed, the errors would more useful.
It could be done using the `raise ... from ...` syntax, for example:
```python
try:
    return keyring.get_password(name, username)
except (RuntimeError, keyring.errors.KeyringError) as err:
    raise KeyRingError(
        "Unable to retrieve the password for {} from the key ring".format(name)
    ) from err
```
or by including the original error message, when available, to the new exception message.

In my case, the original error message (the one from `keyring.errors.KeyringError`) would tell me that Apple does not like my python interpreter because it is not codesign...

vighnesh-kadam:
Hello! im interested in working on this issue , can you assign it to me please?

Secrus:
@vighnesh-kadam thank you. Assigned the issue to you.

PR #5663, opened 2022-05-22
Incomplete error message in exception

vighnesh-kadam:
# Pull Request Check List

Resolves: #5598 

### DIFF
--- a/src/poetry/utils/password_manager.py
+++ b/src/poetry/utils/password_manager.py
@@ -69,10 +69,10 @@ class KeyRing:
 
         try:
             return keyring.get_password(name, username)
-        except (RuntimeError, keyring.errors.KeyringError):
+        except (RuntimeError, keyring.errors.KeyringError) as err:
             raise KeyRingError(
                 f"Unable to retrieve the password for {name} from the key ring"
-            )
+            ) from err
 
     def set_password(self, name: str, username: str, password: str) -> None:
         if not self.is_available():
--- a/tests/fixtures/git/github.com/demo/demo/demo.egg-info/PKG-INFO
+++ b/tests/fixtures/git/github.com/demo/demo/demo.egg-info/PKG-INFO
@@ -6,8 +6,5 @@ Home-page: https://github.com/demo/demo
 Author: Sébastien Eustace
 Author-email: sebastien@eustace.io
 License: MIT
-Platform: UNKNOWN
 Provides-Extra: foo
 Provides-Extra: bar
-
-UNKNOWN
--- a/tests/fixtures/git/github.com/demo/demo/demo.egg-info/PKG-INFO
+++ b/tests/fixtures/git/github.com/demo/demo/demo.egg-info/PKG-INFO
@@ -6,5 +6,8 @@ Home-page: https://github.com/demo/demo
 Author: Sébastien Eustace
 Author-email: sebastien@eustace.io
 License: MIT
+Platform: UNKNOWN
 Provides-Extra: foo
 Provides-Extra: bar
+
+UNKNOWN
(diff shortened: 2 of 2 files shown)


vighnesh-kadam:
i dont know why that one test failed ? Is it because of my local python3 version? currently it is 3.8.10

tests/fixtures/git/github.com/demo/demo/demo.egg-info/PKG-INFO
```diff
@@ -6,9 +6,5 @@ Home-page: https://github.com/demo/demo
 Author: Sébastien Eustace
 Author-email: sebastien@eustace.io
 License: MIT
-Platform: UNKNOWN
 Provides-Extra: foo
 Provides-Extra: bar
-
-UNKNOWN
-
```
Secrus, line 14: These are changes made by flaky test and shouldn't be part of your changes. Please revert that file.

src/poetry/utils/password_manager.py
```diff
@@ -69,10 +69,10 @@ def get_password(self, name: str, username: str) -> str | None:
 
         try:
             return keyring.get_password(name, username)
-        except (RuntimeError, keyring.errors.KeyringError):
+        except (RuntimeError, keyring.errors.KeyringError) as err:
             raise KeyRingError(
                 f"Unable to retrieve the password for {name} from the key ring"
-            )
+            ) from err
```
Secrus, line 75: I would like to see a test for that.

Secrus (changes requested):

tests/fixtures/git/github.com/demo/demo/demo.egg-info/PKG-INFO
vighnesh-kadam: Okay i will revert the file

src/poetry/utils/password_manager.py
vighnesh-kadam: I did not get what you meant by this ?  i just do the `poetry run pytest tests/`  on the files ,  do you want me to do anything more for this specific file?

Secrus: Yes. I think there should be a test written to check if a proper exception is raised.

vighnesh-kadam: Alright! I will do it .

neersighted: A proper test should fail if your change is not applied, and succeed when it is applied. I agree that we'll need a regression test to validate/merge this.

eamanu:
Hi @vighnesh-kadam  seems that only tests is missing?, Are still interested in this PR?

vighnesh-kadam:
> Hi @vighnesh-kadam  seems that only tests is missing?, Are still interested in this PR?

Yes I'm very much interested in completing this!! Last time i triee i was not able to clear the tests on ci/cd . Can you help me out with the next steps?

eamanu:
Well, I'm looking some points here. 
First of one, this change in the code should make some tests fails, but no tests fail. So the options are a) your change has not effect, or b) we need improve the tests.
The second point to think is that the issue is talking about error message in general, not only the manage of the KeyRingError. I guess that the author of the issue used that KeyRingError just an example. So, I suppose that need more than one line to change.

vighnesh-kadam:
> Well, I'm looking some points here. 
> First of one, this change in the code should make some tests fails, but no tests fail. So the options are a) your change has not effect, or b) we need improve the tests.
> The second point to think is that the issue is talking about error message in general, not only the manage of the KeyRingError. I guess that the author of the issue used that KeyRingError just an example. So, I suppose that need more than one line to change. 
> 

Sorry for the late reply , actually i wont be able to contribute to this and be active for a month or so due to personal commitments . Thank you for giving me thIs opportunity!!

Secrus:
@vighnesh-kadam are you still interested in finishing this?

vighnesh-kadam:
> @vighnesh-kadam are you still interested in finishing this? 

Thank you for the opportunity!! But I won't be able to commit to this at the moment.

ralbertazzi:
@Secrus do we really need a test case for a simple raise from? This feels a no-brain merge to me.

Secrus:
Fixed in another commit.

PR #5670, opened 2022-05-23
support for md5-hash pypiserver

ombschervister:
If you are using your own pypiserver which returns md5 hashes, then the poetry.lock file contains md5 hashes. The new version of poetry now throws a RuntimeError for these hashes because it only checks for sha256


b2metric (approved):
ok

Secrus:
superseded by #8118

PR #5672, opened 2022-05-23
config: rework eval order and respect env vars

abn:
This change ensures that we respect `POETRY_CONFIG_DIR` as well as
ensure environment variables are respected for `POETRY_CACHE_DIR`.

This change also fixes an issue with repository cache directories not
respecting base cache directory configuration changes.

### DIFF
--- a/src/poetry/console/commands/cache/list.py
+++ b/src/poetry/console/commands/cache/list.py
@@ -1,7 +1,6 @@
 from __future__ import annotations
 
-import os
-
+from poetry.config.config import Config
 from poetry.console.commands.command import Command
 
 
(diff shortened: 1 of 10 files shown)


src/poetry/console/commands/cache/list.py
```diff
@@ -11,10 +11,8 @@ class CacheListCommand(Command):
     description = "List Poetry's caches."
 
     def handle(self) -> int | None:
-        from poetry.locations import REPOSITORY_CACHE_DIR
-
-        if os.path.exists(str(REPOSITORY_CACHE_DIR)):
-            caches = sorted(REPOSITORY_CACHE_DIR.iterdir())
+        if os.path.exists(str(self.poetry.config.repository_cache_directory)):
```
Secrus, line 14: `repository_cache_directory` is `Path`, so we can do: 
```suggestion
        if self.poetry.config.repository_cache_directory.exists():
```
Looks much cleaner imho

Secrus (changes requested):

src/poetry/console/commands/cache/list.py
abn: Good catch.

Secrus (approved):

neersighted (approved):

PR #5662, opened 2022-05-22
Bump `flake8` additional dependencies

mkniewallner:
# Pull Request Check List
- [ ] ~Added **tests** for changed code.~ Not applicable
- [ ] ~Updated **documentation** for changed code.~ Not applicable

Similarly to https://github.com/python-poetry/poetry-core/pull/362, update flake8 dependencies.

https://github.com/python-poetry/poetry/commit/b1a6ef5b63f36d17f69e2487085d6d33f860601a updates some assignments to avoid unnecessary `map`, according to the new [C417 rule from flake8-comprehensions](https://github.com/adamchainz/flake8-comprehensions/blob/6e319a14ca056bf38e3fad16862033d1c3a1d0a5/README.rst#c417-unnecessary-map-usage---rewrite-using-a-generator-expressionlistsetdict-comprehension).

https://github.com/python-poetry/poetry/commit/356766efaca264ddba3484dc2b00ee811079326d is a bit more tricky. It updates misuses of `lru_cache` that prevent class instances from being garbage collected when it is applied on class instances methods.
This is reported by the new [B019 rule of flake8-bugbear](https://github.com/PyCQA/flake8-bugbear/pull/218).
Some resources about this:
- https://rednafi.github.io/reflections/dont-wrap-instance-methods-with-functoolslru_cache-decorator-in-python.html
- https://www.youtube.com/watch?v=sVjtp6tGo0g

Changelogs:
- [flake8-annotations](https://github.com/sco1/flake8-annotations/blob/ea1e59eb02b0ddc1eb854aea23ea9e74a84deed3/CHANGELOG.md)
- [flake8-bugbear](https://github.com/PyCQA/flake8-bugbear/blob/0d54ee6d14367e444664fa38b8522776aaaee5df/README.rst#change-log)
- [flake8-comprehensions](https://github.com/adamchainz/flake8-comprehensions/blob/05067c706835799ec2fdce403efc2ce115bd8466/HISTORY.rst)
- [flake8-eradicate](https://github.com/wemake-services/flake8-eradicate/blob/275c8963543fbc805fd6d59bbc0d51aecdf65dfc/CHANGELOG.md)
- [flake8-no-pep420](https://github.com/adamchainz/flake8-no-pep420/blob/da3808adca131a8135857a63d83b106d27a4af75/HISTORY.rst)
[...]

PR #5673, opened 2022-05-23
update poetry-core to 1.1.0b1

abn:
Requires: https://github.com/python-poetry/poetry-plugin-export/pull/61

### DIFF
--- a/src/poetry/mixology/version_solver.py
+++ b/src/poetry/mixology/version_solver.py
@@ -19,7 +19,6 @@ from poetry.mixology.result import SolverResult
 from poetry.mixology.set_relation import SetRelation
 from poetry.mixology.term import Term
 from poetry.packages import DependencyPackage
-from poetry.utils._compat import metadata
 
 
 if TYPE_CHECKING:
@@ -47,13 +46,7 @@ class DependencyCache:
             list[DependencyPackage],
         ] = {}
 
-        # TODO: re-enable cache when poetry-core upgrade is completed
-        self.search_for = functools.lru_cache(
-            maxsize=128
-            if metadata.version("poetry-core")  # type: ignore[no-untyped-call]
-            != "1.1.0a7"
-            else 0
-        )(self._search_for)
+        self.search_for = functools.lru_cache(maxsize=128)(self._search_for)
 
     def _search_for(self, dependency: Dependency) -> list[DependencyPackage]:
         key = (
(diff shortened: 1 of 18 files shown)


Secrus (approved):

src/poetry/mixology/version_solver.py
```diff
@@ -47,13 +46,7 @@ def __init__(self, provider: Provider) -> None:
...
         ] = {}
 
-        # TODO: re-enable cache when poetry-core upgrade is completed
-        self.search_for = functools.lru_cache(
-            maxsize=128
-            if metadata.version("poetry-core")  # type: ignore[no-untyped-call]
-            != "1.1.0a7"
-            else 0
-        )(self._search_for)
+        self.search_for = functools.lru_cache(maxsize=128)(self._search_for)
```
neersighted, line 49: Can we use the decorator syntax again? 😄

neersighted (approved):

src/poetry/mixology/version_solver.py
Secrus: nope, this causes memory leaks

Secrus: also `flake8-bugbear` will point it out as memory leak

abn: What @Secrus said, this was actually a fix for memory leak :D we could have done the conditional on decorator as well if we wanted.

mkniewallner: Yep, since https://github.com/python-poetry/poetry/pull/5662, using a decorator on class instances methods is reported by `flake8-bugbear` as a potential bad thing to do, as this prevents objects from being garbage collected correctly.

The rationale is explained in those 2 resources:
- https://rednafi.github.io/reflections/dont-wrap-instance-methods-with-functoolslru_cache-decorator-in-python.html
- https://www.youtube.com/watch?v=sVjtp6tGo0g

PR on `flake8-bugbear` side, for even more background: https://github.com/PyCQA/flake8-bugbear/pull/218.

abn:
Converting to draft to debug issue on windows.

abn:
Working around `tomlkit` regression on Windows. 

https://github.com/python-poetry/poetry/pull/5673/commits/f29ca4ebc1f1d012bea9a193cdd526f16d0ca876

This should be reverted prior to 1.2 hopefully.

neersighted (approved):

PR #5678, opened 2022-05-24
schema: validate source objects in poetry

abn:
This change is intended to help migrate validation of sections in pyproject.toml that are not relevant to poetry-core into poetry. This allows for poetry frontend specific configuration to be introduced without dependent changes in poetry-core.

Related-to: https://github.com/python-poetry/poetry-core/pull/369

This will make changes like https://github.com/python-poetry/poetry/pull/5442 easier.


Secrus (approved):

src/poetry/factory.py
```diff
@@ -311,3 +312,13 @@ def create_pyproject_from_package(
...
 
         return cast(TOMLDocument, pyproject)
+
+    @classmethod
+    def validate(
+        cls, config: dict[str, Any], strict: bool = False
+    ) -> dict[str, list[str]]:
+        results = super().validate(config, strict)
+
+        results["errors"].extend(validate_object(config.get("source", []), "source"))
```
neersighted, line 322: I'm not as excited about this hardcoding approach -- how do we feel about overriding the schema with an extended version using `allOf`?

https://json-schema.org/understanding-json-schema/reference/combining.html

neersighted (changes requested):

src/poetry/factory.py
abn: If I am groking your suggestion correctly, I reckon that's unnecessary complexity given the schemas will live in different repos and original validation logic lives in core. 

If the explicit config section by section validation is not preferred, we could avoid that hard coding by using a unified schema. But that I reckon negatively impacts maintainability by ending up with a large schema.

Personally, I don't think it is required. I can take a look at how this might come out in code if you reckon there is definite value in exploring it.

neersighted: I'm of two minds here on what would be 'cleaner' -- do all the validation up front in Poetry, extending the core schema, so that it all happens in once place with one set of messages? This would mean that core's validation code only really runs when used as a pep517 backend by another frontend.

The other option is the one you've presented here, where the validation is split. I personally would find the large schema easier to reason about compared to many small separate schemas for each table -- when I want to think about the file format I want to think about it holistically instead of having to tab through disparate definitions in my editor.

I can definitely be convinced to go with the approach in this PR, but I am curious if we should outright reject the frontend doing all the validation.

abn: The initial reason why I started down the path of this change was to try reduce vendored dependencies in `poetry-core`. The initial target was `jsonschema`. My view is that core never really modifies the `pyproject.toml`, so validation at the core level seems - unnecessary. That said there is some value in it to allow for builds to fail early etc in cases of bad configuration.

And this is the only case where I would say doing all the validation in the frontend might make sense. I am not a fan of both doing full validation, which I know is not what you are recommending.

I am not opposed to a single schema either. Both layouts have their advantages. The intention of this change is simply to let the frontend validation things it extends. I really do not want to end up in a scenario where

```toml
package = { version = "1.0.0", frontened-option = false }
```

and

```toml
package = { version = "1.0.0" }
```

are considered valid by the frontend, whereas only the second option is valid in the backend. This will lead to scenarios where `poetry build` will succeed but `pip wheel --no-deps` will fail. This will could be the case if we went the route of combining schemas.

abn: @neersighted I had a look at how we could do this in a clean manner. But the best I could come up with is something like this.

```py
    schema = {
        "allOf": []
    }

    for base_dir in [CORE_SCHEMA_DIR, SCHEMA_DIR]:
        for schema_file in Path(base_dir).glob("*.json"):
            schema["allOf"].append({
                "$ref": schema_file.as_uri()
            })
```

But this also mens we either need to split out validation in core or repeat the logc in the frontend. Particularly when `strict` mode is used and the addtional validation for readmes and scripts. This is not ideal. And also important to note that this can lead to more human errors and also can cause issues in some environments when the schema discovery might not work as expected.

So, right now, unless there is a cleaner proposal, I reckon the approach proposed in this PR is probably the better option.  We can pull it into a single schema (ie. frontend properties) however, that is not my preference as this enables reuse of config sections easier.

PR #5680, opened 2022-05-24
source show: include pypi and additional information about order

abn:
```console
$ poetry source show -v
 name       : pypi-simple             
 url        : https://pypi.org/simple 
 default    : no                      
 secondary  : no                      

 name       : pypi                     
 url        : https://pypi.org/simple/ 
 default    : no                       
 secondary  : yes                      

Poetry will search all of the above package sources when searching for packages.

All other things equal, for a give package that is available from multiple sources, it will be preferred in the order listed above.

For example, if a package foo is available in both pypi-simple and pypi, Poetry will select pypi-simple.

You can override this, by explicitly specify the source when adding the package.

    $ poetry add --source pypi foo

Alternatively, you can set all your custom sources as secondary.

```

### DIFF
--- a/src/poetry/console/commands/source/show.py
+++ b/src/poetry/console/commands/source/show.py
@@ -2,6 +2,7 @@ from __future__ import annotations
 
 from cleo.helpers import argument
 
+from poetry.config.source import Source
 from poetry.console.commands.command import Command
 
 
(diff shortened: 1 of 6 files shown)


src/poetry/console/commands/source/show.py
```diff
@@ -18,42 +19,87 @@ class SourceShowCommand(Command):
...
+            False: "no",
+        }
+
+        table = self.table(style="compact")
+        rows = [
+            ["<info>name</>", f" : <c1>{source.name}</>"],
+            ["<info>url</>", f" : {source.url}"],
+            [
+                "<info>default</>",
+                f" : {bool_string.get(source.default, False)}",
```
Secrus, line 34: Shouldn't this `False` be `"no"`? And the same in line 38?

Secrus (changes requested):

src/poetry/console/commands/source/show.py
neersighted: `False` is the default value if `source.default` is `None`.

Secrus: ok, but if I understand this right, we want to display `default: yes` or `default: no`. With `False`, it will be `default: False`.

neersighted: Ah, fair point... `bool_string.get(bool(source.default))` might lead to the desired behavior.

abn:
Closing in favour of #9974.

Issue #5537, opened 2022-05-04
Export fails to take proper care of extra in deps with more than one path


shaib:
- [x] I am on the [latest](https://github.com/python-poetry/poetry/releases/latest) Poetry version.
- [x] I have searched the [issues](https://github.com/python-poetry/poetry/issues) of this repo and believe that this is not a duplicate.
- [x] If an exception occurs when executing a command, I executed it again in debug mode (`-vvv` option).

- **OS version and name**: Debian testing (bookworm, i.e. on the way to 12.0)
- **Poetry version**: 1.1.13 installed from pip
- **Link of a [Gist](https://gist.github.com/) with the contents of your pyproject.toml file**: https://gist.github.com/shaib/3bd014dd9cfe39faf23dcd1363da9a2d

## Issue
Starting from scratch, adding one dependency which requires `google-api-core[grpc]` and another which constrains `google-api-core` (the real project is of course significantly more complex and the limit is derived), all is well as long as we're in poetry; but when exported, even though all the chain of dependencies seems to be exported, it seems that the extra designation itself is not; and the result is, that when we try to `pip install` from the requirements file, we get
```
ERROR: In --require-hashes mode, all requirements must have their versions pinned with ==. These do not:
    google-api-core[grpc]<2.0.0dev,>=1.14.0 from ... (from firebase-admin==4.5.3->-r requirements.txt (line 13))
```

All the details -- a log of the session, the `poetry add -vvv` output, the pyproject.toml, poetry.lock and requirements.txt are in the gist listed above.

(edit: Running, in the above, the command
```
$ sed -i 's/google-api-core/google-api-core[grpc]/g' requirements.txt
```
allows pip to install everything)

#4912 seems to be closely related but I believe it is not the same issue.

Secrus:
Hi. As with release 1.2, `export` is gonna be moved to plugin [`poetry-plugin-export`](https://github.com/python-poetry/poetry-plugin-export), could you please check if this issue is still present with poetry 1.2 and plugin? If so, please report it in the plugin repository (linked above).

johnthagen:
I hit a similar issue in Docker running `poetry export` with Poetry 1.1.13:

```
#0 167.7 Collecting cffi>=1.0.1
#0 167.7 ERROR: In --require-hashes mode, all requirements must have their versions pinned with ==. These do not:
[...]

PR #5688, opened 2022-05-25
exporting packages with features

dimbleby:
Fix for #5537, see also https://github.com/python-poetry/poetry-plugin-export/pull/65

Have `get_project_dependency_packages()` return packages with features suitably set, so that the exporter can record them.  Features are accumulated as we walk the tree.

With that set of fixes, [this test](https://github.com/python-poetry/poetry-plugin-export/blob/db1db238d521462960616463741f20c63c9cb6b7/tests/test_exporter.py#L2140-L2143) sees this diff
```patch
-localstack-ext==1.0.0 ; {MARKER_PY36}
-localstack==1.0.0 ; {MARKER_PY36}
+localstack-ext[bar]==1.0.0 ; {MARKER_PY36}
+localstack[foo]==1.0.0 ; {MARKER_PY36}
```
which is https://github.com/python-poetry/poetry-plugin-export/pull/66

Direct manipulation of `_extras` is not great, we should add `set_extras(extras)` or similar in `poetry-core`.  That's a little exciting - `_extras` gets hashed so careless callers can go wrong - but it's not worse than `set_constraint()` as also used by the locker (and, similarly, is safe to do in this code).

Also we can remove `requires_extras` from poetry-core, it's unused.

### DIFF
--- a/src/poetry/packages/locker.py
+++ b/src/poetry/packages/locker.py
@@ -261,18 +261,22 @@ class Locker:
             if not locked_package:
                 raise RuntimeError(f"Dependency walk failed at {requirement}")
 
+            if requirement.extras:
+                locked_package = locked_package.with_features(requirement.extras)
+
             # create dependency from locked package to retain dependency metadata
             # if this is not done, we can end-up with incorrect nested dependencies
             constraint = requirement.constraint
             marker = requirement.marker
-            extras = requirement.extras
             requirement = locked_package.to_dependency()
             requirement.marker = requirement.marker.intersect(marker)
 
             requirement.set_constraint(constraint)
 
             for require in locked_package.requires:
-                if require.in_extras and extras.isdisjoint(require.in_extras):
+                if require.in_extras and locked_package.features.isdisjoint(
+                    require.in_extras
+                ):
                     continue
 
                 require = deepcopy(require)
@@ -369,9 +373,6 @@ class Locker:
             project_requires=selected,
             locked_packages=repository.packages,
         ):
-            for extra in dependency.extras:
-                package.requires_extras.append(extra)
-
             yield DependencyPackage(dependency=dependency, package=package)
 
     def set_lock_data(self, root: Package, packages: list[Package]) -> bool:


dimbleby:
I don't know what order these things can all be committed in to keep the tests passing at all times!

src/poetry/packages/locker.py
```diff
@@ -268,6 +268,7 @@ def __walk_dependencies(
             extras = requirement.extras
             requirement = locked_package.to_dependency()
             requirement.marker = requirement.marker.intersect(marker)
+            requirement._extras = extras
```
abn, line 271:
```suggestion
            requirement._extras = extras
            requirement._features = frozenset(extras)
```

In deference to the change proposed in https://github.com/python-poetry/poetry-core/pull/373/files#r882064718 this should allow us to update to next core without issues.

src/poetry/packages/locker.py
```diff
@@ -283,12 +284,12 @@ def __walk_dependencies(
...
-            if key not in nested_dependencies:
-                nested_dependencies[key] = requirement
-            else:
-                nested_dependencies[key].marker = nested_dependencies[key].marker.union(
+            combined_requirement = nested_dependencies.setdefault(key, requirement)
+            if combined_requirement is not requirement:
+                combined_requirement.marker = combined_requirement.marker.union(
                     requirement.marker
                 )
+                combined_requirement._extras |= requirement.extras
```
abn, line 292: Similar as above.

abn (changes requested):
Thanks @dimbleby for picking this up. Some comments in relation to fixing the duplicated attributes in core's `Dependency` implementation.

src/poetry/packages/locker.py
dimbleby: The intention is that this MR fixes things with the current poetry-core, and when poetry-core exposes a better API we use it here - which I would expect to do with a later MR.

If you want to block this behind publishing poetry-core improvements I guess that's fine by me, then we can do it in a single commit here.

abn:
One possible order for these changes could be as follows.

1. Add a ` || true` to the export test command in ci so the workflow does not bail
2. Relock the dev dependency in the export plugin in https://github.com/python-poetry/poetry-plugin-export/pull/66
3. Make a new release of the export plugin
4. Update plugin version on poetry and undo (1).

In the long term we need to fix this coupling somehow :(

dimbleby:
Long-term, I suppose that this repository should simply not know anything about poetry-plugin-export; because that plugin should be no more special than any other.

(Also that would fix the circular dependency where both projects depend on one another).

Perhaps we can just do that now, must we test the export plugin from here?

src/poetry/packages/locker.py
abn: I am simply anticipating a forward port required to get the downstream tests passing. And suggesting that we remove `requirement._extras = extras` once the new API is available.

I'd rather not block this behind a new core release and have this fixed in the next poetry pre-release that will likely happen prior to next core release.

abn:
> Perhaps we can just do that now, must we test the export plugin from here?

It is possible yes. However, might require a deprecation cycle for the command itself. The issue is `export` command has been a core poetry command, and changes in poetry can break the export command. The tests were introduced because of one such case a while back where `poetry export` broke badly.

dimbleby:
reworked fix is now simpler, and exports like this in that existing testcase:
```patch
 localstack-ext==1.0.0 ; {MARKER_PY36}
-localstack==1.0.0 ; {MARKER_PY36}
+localstack-ext[bar]==1.0.0 ; {MARKER_PY36}
+localstack[foo]==1.0.0 ; {MARKER_PY36}
```
ie we get both localstack-ext and localstack-ext[bar]

This is cosmetically ugly when, as here, both variations have the same markers. But in cases where they have different markers eg some extra is required on windows only - this seems like it is giving the right answers where the previous try didn't.

abn (approved):

PR #5697, opened 2022-05-26
entrypoints 0.4

dimbleby:
entrypoints 0.4 was released in February.


abn (approved):
Approved. We should drop entrypoints for importlib.metadata soon anyway though.

Issue #5682, opened 2022-05-25
Current master falls back to PyPi when using private repo.


voney:
Following on from #5651

The original issue was that poetry could not handle repositories that hand back CDN links for packages with authenticationbaked into the URL because it would still add the base repository index auth details to the request headers.

After trying the lastest version from git I get a new issue.

```
(.venv) [daniel@dv-xps-fedora Davis_state_consumer]$ poetry@git add --source myget 'redacted'
Using version ^2.0.6 for redacted

Updating dependencies
Resolving dependencies... (4.2s)

  HTTPError

  404 Client Error: Not Found for url: https://pypi.org/pypi/redacted_dependency/json/

  at ~/.local/pipx/venvs/poetry@git/lib64/python3.10/site-packages/requests/models.py:960 in raise_for_status
      956│         elif 500 <= self.status_code < 600:
      957│             http_error_msg = u'%s Server Error: %s for url: %s' % (self.status_code, reason, self.url)
      958│ 
      959│         if http_error_msg:
    → 960│             raise HTTPError(http_error_msg, response=self)
      961│ 
      962│     def close(self):
      963│         """Releases the connection back to the pool. Once this method has been
      964│         called the underlying ``raw`` object must not be accessed again.
```

That's after adding the myget repo using:

```
poetry@git source add myget https://www.myget.org/F/redacted/python/
```

_Originally posted by @voney in https://github.com/python-poetry/poetry/issues/5651#issuecomment-1136564646_

abn:
Following from #5651. Any chance you can provide the logs for the same command with `-vvv`?

voney:
Sorry for the double tickets, I'll comment here going forward.

```
(.venv) [daniel@dv-xps-fedora my_project]$ poetry@git add --source myget 'redacted' -vvv
Loading configuration file /home/daniel/.config/pypoetry/config.toml
Loading configuration file /home/daniel/.config/pypoetry/auth.toml
Adding repository myget (https://www.myget.org/F/myrepo/python)
Using virtualenv: /home/daniel/myrepo/my_project/.venv
Project environment contains an empty path in sys_path, ignoring.
[keyring.backend] Loading KWallet
[keyring.backend] Loading SecretService
[keyring.backend] Loading Windows
[keyring.backend] Loading chainer
[keyring.backend] Loading libsecret
[keyring.backend] Loading macOS
Creating new session for www.myget.org
[...]

PR #5698, opened 2022-05-26
pypi: do not raise exception on 404

abn:
Resolves: #5682

Writing a test case here is significantly more effort as this will require some form if setup where we use http mocks instead of current  `MockRepository._get` monkey patching.

### DIFF
--- a/src/poetry/repositories/pypi_repository.py
+++ b/src/poetry/repositories/pypi_repository.py
@@ -243,12 +243,16 @@ class PyPiRepository(HTTPRepository):
 
     def _get(self, endpoint: str) -> dict[str, Any] | None:
         try:
-            json_response = self.session.get(self._base_url + endpoint)
+            json_response = self.session.get(
+                self._base_url + endpoint, raise_for_status=False
+            )
         except requests.exceptions.TooManyRedirects:
             # Cache control redirect loop.
             # We try to remove the cache and try again
             self.session.delete_cache(self._base_url + endpoint)
-            json_response = self.session.get(self._base_url + endpoint)
+            json_response = self.session.get(
+                self._base_url + endpoint, raise_for_status=False
+            )
 
         if json_response.status_code == 404:
             return None
--- a/src/poetry/repositories/pypi_repository.py
+++ b/src/poetry/repositories/pypi_repository.py
@@ -254,7 +254,7 @@ class PyPiRepository(HTTPRepository):
                 self._base_url + endpoint, raise_for_status=False
             )
 
-        if json_response.status_code == 404:
+        if json_response.status_code != 200:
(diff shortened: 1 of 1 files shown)


Secrus (approved):

src/poetry/repositories/pypi_repository.py
```diff
@@ -243,12 +243,16 @@ def _get_release_info(
 
     def _get(self, endpoint: str) -> dict[str, Any] | None:
         try:
-            json_response = self.session.get(self._base_url + endpoint)
+            json_response = self.session.get(
+                self._base_url + endpoint, raise_for_status=False
```
dimbleby, line 247: This changes the behaviour for all errors, some of which might plausibly contain a json body.  In that case the eventual failure is likely to come at some later and more confusing point.

Suggest that still want to raise_for_status on non-404 errors.

abn: We can get `raise_for_status` to accept a filter I guess. But it should also be noted that this is he behaviour for legacy anyway right now. Maybe logging the errors and not bailing out is a better compromise.

abn:
```console
Creating new session for pypi.org
[urllib3.connectionpool] Starting new HTTPS connection (1): pypi.org:443
[urllib3.connectionpool] https://pypi.org:443 "GET /pypi/fooobaaar/json/ HTTP/1.1" 404 1979
Source (PyPI): No packages found for fooobaaar >=0.1.0,<0.2.0
Falling back to installed packages to discover metadata for fooobaaar
Found 0 compatible packages for fooobaaar
   1: fact: no versions of fooobaaar match >=0.1.0,<0.2.0
   1: conflict: no versions of fooobaaar match >=0.1.0,<0.2.0
```

This is what the logs look like with the fix.

neersighted (dismissed):

src/poetry/repositories/pypi_repository.py
```diff
@@ -243,12 +243,16 @@ def _get_release_info(
...
         except requests.exceptions.TooManyRedirects:
             # Cache control redirect loop.
             # We try to remove the cache and try again
             self.session.delete_cache(self._base_url + endpoint)
-            json_response = self.session.get(self._base_url + endpoint)
+            json_response = self.session.get(
+                self._base_url + endpoint, raise_for_status=False
+            )
 
         if json_response.status_code == 404:
```
abn, line 257: Thinking a bit about what @dimbleby said, it is possible that we can mitigate the "bad data" scenario by doing something like this. 


```suggestion
        if not 200 <= json_response.status_code <  300:
```

neersighted (approved):

PR #5704, opened 2022-05-27
provider: allow fallback to installed packages

abn:
Split out from https://github.com/python-poetry/poetry/pull/5450

### DIFF
--- a/src/poetry/puzzle/provider.py
+++ b/src/poetry/puzzle/provider.py
@@ -35,6 +35,7 @@ from poetry.mixology.term import Term
 from poetry.packages import DependencyPackage
 from poetry.packages.package_collection import PackageCollection
 from poetry.puzzle.exceptions import OverrideNeeded
+from poetry.repositories.exceptions import PackageNotFound
 from poetry.utils.helpers import download_file
 from poetry.vcs.git import Git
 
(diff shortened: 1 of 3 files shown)


src/poetry/puzzle/provider.py
```diff
@@ -478,15 +516,25 @@ def complete_package(self, package: DependencyPackage) -> DependencyPackage:
...
+                        repository=package.dependency.source_name,
+                    ),
+                )
+            except PackageNotFound as e:
+                for pkg in self.search_for_installed_packages(package.dependency):
+                    package = DependencyPackage(
+                        package.dependency,
+                        pkg,
+                    )
+                    break
```
neersighted, line 535: I'm not quite sure what this break is for -- if we just intend to get the first package in the list, a call to `next()` might be more explicit.

abn: More to trigger the else block in the for loop, we can also do `next` and then catch the `StopIteration`. For me this is a bit cleaner; but I am happy to switch to `next` too.

neersighted: Ah, I missed the for-else -- I think I would prefer a next/StopIteration as the pattern is less likely to get lost in indentation noise (like just happened to me) and the intent is more explicit.

abn: Fine! Here I was ... all excited I could use a `for/else` :D

neersighted (approved):

PR #5708, opened 2022-05-27
docs: Improve contributing docs

juhoautio:
Switching to directly paste-able commands by removing the extra `'$ '` prefixes.

## Question about the file structure

Is it really so that `CONTRIBUTING.md` and `docs/contributing.md` (which have the same content with slightly different formatting syntax) need to be kept in sync manually? If yes, may I add comments to in these files to mention that?

## Question about local dev setup

Is there a known requirement for minimum Python version for developing poetry? If yes, I can document it as well. I was trying to run `poetry run pre-commit run --all-files` with Python 3.7 and stumbled on this:

```
    ERROR: Could not find a version that satisfies the requirement flake8-type-checking==1.5.0 (from versions: 0.1.0, 0.1.1, 0.1.2, 0.1.3, 0.1.4, 0.1.5, 1.0.0, 1.0.1, 1.0.2, 1.0.3)
    ERROR: No matching distribution found for flake8-type-checking==1.5.0
```

I switched to Python 3.8 and also activated `poetry shell` – then the `poetry run pre-commit run --all-files` succeeded. But I'm not sure is this about Python version or something else – please advise what is the recommended setup for local dev?

# Pull Request Check List

(Keeping these from the PR template in case I end up making code changes, too):

### DIFF
--- a/docs/contributing.md
+++ b/docs/contributing.md
@@ -1,3 +1,5 @@
+<!-- The content of this file is manually kept in sync with ../CONTRIBUTING.md. There are formatting differences because this file uses Hugo templating. -->
+
 ---
 title: "Contributing to Poetry"
 draft: false
--- a/docs/contributing.md
+++ b/docs/contributing.md
@@ -122,8 +122,8 @@ You will need Poetry to start contributing on the Poetry codebase. Refer to the 
 You will first need to clone the repository using `git` and place yourself in its directory:
 
 ```bash
-$ git clone git@github.com:python-poetry/poetry.git
-$ cd poetry
+git clone git@github.com:python-poetry/poetry.git
+cd poetry
 ```
 
 {{% note %}}
@@ -135,15 +135,15 @@ Now, you will need to install the required dependency for Poetry and be sure tha
 tests are passing on your machine:
 
 ```bash
-$ poetry install
-$ poetry run pytest tests/
+poetry install
+poetry run pytest tests/
 ```
 
 Poetry uses [mypy](https://github.com/python/mypy) for typechecking, and the CI
 will fail if it finds any errors.  To run mypy locally:
 
 ```bash
-$ poetry run mypy
+poetry run mypy
 ```
 
 Poetry uses the [black](https://github.com/psf/black) coding style and you must ensure that your
@@ -156,13 +156,13 @@ To make sure that you don't accidentally commit code that does not follow the co
 install a pre-commit hook that will check that everything is in order:
 
 ```bash
-$ poetry run pre-commit install
+poetry run pre-commit install
 ```
 
 You can also run it anytime using:
 
 ```bash
-$ poetry run pre-commit run --all-files
+poetry run pre-commit run --all-files
 ```
 
 Your code must always be accompanied by corresponding tests, if tests are not present your code
--- a/docs/contributing.md
+++ b/docs/contributing.md
@@ -1,7 +1,3 @@
-<!--
-The content of this file is manually kept in sync with ../CONTRIBUTING.md. There are formatting differences because this file uses Hugo templating.
--->
-
 ---
 title: "Contributing to Poetry"
 draft: false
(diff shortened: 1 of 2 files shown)


Secrus:
@juhoautio `CONTRIBUTING.md` and `docs/contributing.md` are different mostly because of Hugo templating used in `docs/` version. As to comment about manual sync, if you feel the need, it's okay, but add it in `CONTRIBUTING.md` as the `docs/` version is being rendered into the website. 

You should have no problem with running Poetry code on Python 3.7, pre-commit however is slightly not in sync as you noticed. That was missed probably because the `pre-commit` workflow runs separately from tests. Personally, I see no reason to stay on old versions and I work on 3.10. You may add info about contributions requiring 3.8.

juhoautio:
> As to comment about manual sync, if you feel the need, it's okay, but add it in CONTRIBUTING.md as the docs/ version is being rendered into the website.

I would prefer a visible note in `CONTRIBUTING.md` and a hidden comment that is not showed in the user-facing rendered website in the `docs/` version. The reason why I'd like to add those is that I spent some time looking for some kind of generator/converter because I was not expecting that two files with the ~same content are maintained manually :)

neersighted:
I'm waiting on this until https://github.com/python-poetry/website/pull/37 is done as we can take advantage of the new clipboard button in this PR.

docs/contributing.md
```diff
@@ -1,3 +1,5 @@
+<!-- The content of this file is manually kept in sync with ../CONTRIBUTING.md. There are formatting differences because this file uses Hugo templating. -->
```
juhoautio, line 1: This could be mentioned:

> When you make a PR, a github action runs automatically and deploys a test website using your branch.

But it would be good to know also how to build & view the docs locally. For example adding this comment here in a "hugo template file", I'm not sure how it's going to look like when it's live before the gh action has finished deploying the changes that I pushed.

Secrus: You can clone the repo, use provided README file to set it up (granted, it's not the best manual, I am working on something better) and see it locally.

docs/contributing.md
```diff
@@ -1,4 +1,6 @@
-<!-- The content of this file is manually kept in sync with ../CONTRIBUTING.md. There are formatting differences because this file uses Hugo templating. -->
```
juhoautio: With a single line comment this hit a known Hugo bug: https://github.com/gohugoio/hugo/issues/7092.

Splitting to more than a single line seems to work.

docs/contributing.md
juhoautio: Thanks!

I had to run `npm install` before `npm run dev` – so that's missing from https://github.com/python-poetry/website#local-development.

I get this in the terminal:
```
rollup v2.50.5
bundles src/app.js → assets/assets/app.js...
```

Where can I view the docs?

neersighted: It should be in the `public/` folder as a pile of files -- that being said, note that the action requires someone to add the `Documentation` tag, and presently that has to be a member of the Poetry team. It's really meant as a way for the team to view the rendered docs without a local clone, as opposed to a way to give feedback to the PR author.

We should focus on documenting how to build the website locally for this use case.

juhoautio: I think I figured it out. Previously I was trying `poetry run python bin/website build --local ./poetry` which lead into some errors, and I ignored them – I shouldn't have :) Of course I didn't have `./poetry `, I have `../poetry`.

So, this worked:
```
cd website
poetry run python bin/website build --local ../poetry
hugo server -w
```

> Web Server is available at http://localhost:1313/ (bind address 127.0.0.1)

And that answered my question about "Where can I view the docs?".

I can volunteer to document this in the contributing docs if it helps, let me know.

Secrus: checkout latest update to `package.json` and `npm scripts` provided in website. you can now just run `npm run dev` and both npm and hugo will run. There is also `Makefile` providing ready to use scripts for that. I don't see need for all this to be in Poetry contributing.md. Those are separate projects.

neersighted (approved):
Going to go ahead and merge this now and let @Secrus send a follow-up with the Clipboard button as that website PR is getting bogged down.

docs/contributing.md
juhoautio: Related: https://github.com/python-poetry/website/pull/48

Issue #1854, opened 2020-01-09
`poetry new` to support --dependency & --dev-dependency


juhoautio:
- [x] I have searched the [issues](https://github.com/python-poetry/poetry/issues) of this repo and believe that this is not a duplicate.
- [x] I have searched the [documentation](https://python-poetry.org/docs/) and believe that my question is not covered.

## Feature Request

Could these options be added to the `poetry new` command (non-interactive):

- `--dependency`: Package to require with a version constraint
- `--dev-dependency`: Development packages

The `poetry init` command (interactive) already supports these flags.

Both with `multiple=True`.

----

Maybe also these (`init` supports but `new` doesn't):
- `--description`: Description of the package.
- `--author`: Author of the package.

----

I can implement & make a PR if this makes sense to you.

Secrus:
That could be something to be considered for future release.

juhoautio:
Are you open to reviewing a pull request that implements this?

Secrus:
> Are you open to reviewing a pull request that implements this?

If you are willing to make one, sure. However, I can't guarantee that it will be part of the next release.

juhoautio:
@Secrus I'm hoping to get feedback on my PR, I wonder if you could help with that?

I tried to tag @neersighted, but no response so far: https://github.com/python-poetry/poetry/pull/5710#issuecomment-1152708230.

Thanks!

finswimmer:
I think we should make `poetry new` an extension of `poetry init`. This is why I'm closing this in favor of #2563.

Issue #2563, opened 2020-06-16
Why does `poetry new` not call `poetry init` to build the `pyproject.toml` file interactively?


tsela:
- [x] I have searched the [issues](https://github.com/python-poetry/poetry/issues) of this repo and believe that this is not a duplicate.
- [x] I have searched the [documentation](https://python-poetry.org/docs/) and believe that my question is not covered.

## Feature Request

Hi,

I've recently switched from Pipenv to Poetry and so far this is working fine. It's a bit of an annoyance that VSCode does not recognise Poetry virtualenvs automatically like it does with Pipenv, but it's not Poetry's fault and I have a workaround for that anyway. However, there is something I don't understand about the behaviour of `poetry new`.

Right now, when I want to create a new project, I need to:

1. Run `poetry new` to create the folder structure for the project with the right names for the folder and package;
2. Delete the automatically created `pyproject.toml` file in order for the next step to work;
3. run `poetry init` to create the `pyproject.toml` I actually want interactively;
4. run `poetry install --no-root` to initialise the virtual environment.

The problem is that the `project.toml` created by `poetry new` lacks various fields like the license and description, and it's a bit weird to have to fill them by hand after `poetry new` when the `poetry init` command exists and works so well. But I can't just run `poetry init` after `poetry new` because it complains about an already existing `pyproject.toml`.

In fact, while I understand why one would want `poetry init` as a separate command, I don't understand why `poetry new` doesn't automatically call `poetry init` once it created the folder structure, so that one can build the `pyproject.toml` interactively. I understand why someone would not always want to do that, but `poetry` provides the `--no-interaction` option for that. So why isn't the default behaviour of `poetry new` to call `poetry init`, so that one can create the folder structure and the wanted `pyproject.toml` file in one go? I would expect that to be the most common workflow when creating a new project, so I find it weird that it's not supported out of the box and that I have to go through a weird uncomfortable step (having to erase an automatically created file) in order to make it work.

So I guess my feature request would be to make `poetry new` run `poetry init` to create the `pyproject.toml` file, either by default (with the `--no-interaction` option to bypass it) or using an option (maybe `--init`) if you really do not want to change the default behaviour. What do you think?

raffraffraff:
[...]

PR #5710, opened 2022-05-27
new command to support more options

juhoautio:
Adding all the missing options that are asked interactively by the init command.

Also improving & fixing the doc of init command to match.

**Question:** How to validate the rendering of `docs/cli.md` (especially the `relref`)?
-> Answered in https://github.com/python-poetry/poetry/pull/5708#discussion_r885083371

**Question:** Is it on purpose that `mypy` is not part of the pre-commit hooks? My 1st push failed on CI because of a mypy error because I didn't realize I should've ran it manually.
-> Answered in https://github.com/python-poetry/poetry/pull/5710#issuecomment-1146531780

# Pull Request Check List

Resolves: #1854


neersighted:
> **Question:** Is it on purpose that `mypy` is not part of the pre-commit hooks? My 1st push failed on CI because of a mypy error because I didn't realize I should've ran it manually.

Yes, it is for now. mypy was part of our pre-commit hooks but we can't run it in the correct environment without duplicating our list of dependencies. We are probably going to set it up as a tox environment and then suggest running tox to trigger pytest and mypy.

neersighted (changes requested):
I'm not very enthusiastic about subclassing the init command -- instead, I would prefer to see the common parts of both commands factored out into shared helpers. Also, let's drop the `-l` short option for `--license` as it's inconsistent and it was never documented.

juhoautio:
There are some existing commands that inherit the InitCommand. Should those be also detached by extracting shared helpers? Maybe not yet in this PR though.

juhoautio:
> Also, let's drop the -l short option for --license as it's inconsistent and it was never documented.

Added in commit https://github.com/python-poetry/poetry/pull/5710/commits/7d578a10d3d0f1074f94a48fe3ea90f67ddebfa1.

> I would prefer to see the common parts of both commands factored out into shared helpers.

Done in commit https://github.com/python-poetry/poetry/pull/5710/commits/a09c47b38a15c2803c8bec159f857258eac80838.

juhoautio:
@neersighted is this ok? or something to improve still?

juhoautio:
@neersighted I tried to address the review comments but looks like this PR was forgotten and there are some conflicts to be resolved. Those should be easy to fix, but would it be possible to have a review for the commits that I added since you last checked this?

neersighted:
Hi @juhoautio -- the maintainers are busy people doing this in their spare time. While an occasional ping (especially to those who have already interacted with a PR/issue) is not unwelcome, pinging people who have not expressed interest in an issue or reviewing code, without a prior understanding of what they are interested in helping with, is generally considered rude. Likewise, pinging repeatedly is generally considered rude.

Additionally, Github as a 'request review/re-review' feature that is generally used for this and makes it easier to reason about what reviews are left on your to-do list as a contributor with many priorities. 

Most reviewers will not take a look at code that is conflicted and failing tests as non-mergable code must, by its nature, change (often substantially) in order to be accepted. Thus reviewing it is generally a waste of time for all parties involved. I am sorry that I did not comment that here -- hopefully that explanation/expectation helps in the future when you request reviews.

juhoautio:
Sorry! I had not noticed that there's a `Re-request review` feature, or at least didn't remember. I'll use that from now on.

I'm also very sorry that I ended up using desperate means like pinging another maintainer. In fact I wasn't sure how to best raise my question about what I should be doing next without appearing rude, yet I failed at it. In the end, I now have the guidance that I was looking for. Thanks a lot for that.

juhoautio:
The test failure with `Tests / macOS / 3.11-dev (pull_request) ` doesn't seem to be related to my changes. The same failure seems to be happening at least in one other open PR.

neersighted (changes requested):
Overall looks good to me, but I'm not sure about the addition of `--dev-dependency` while we're moving away from `--dev`. I think we should either generalize it to any group, or just drop it in general as it's going to be pretty niche during `new`.

juhoautio:
Should the existing `--dev-dependency` option of `init` be deprecated?

----

How about having a new option `--group-dependency` for both `init` and `new`? The value format could be `GROUP:DEPENDENCY`.

e.g. `dev:requests:^2.10.0` or `dev:requests=2.11.1` or `dev:requests`.

----

> or just drop it in general as it's going to be pretty niche during new

The ability to set dev dependencies with `new` is needed for my use case. Or then I would like to have an option for `add` that allows only modifying `pyproject.toml` without updating `poetry.lock`.

neersighted:
I think the best move is to drop the `--dev-dependency` change from this PR, and then introduce a new PR deprecating `--dev-dependency` on `init` and adding `--group-dependency` using the design you've proposed.

src/poetry/console/commands/new.py
```diff
@@ -88,3 +134,12 @@ def handle(self) -> int:
         )
 
         return 0
+
+    def _get_pool(self) -> Pool:
+        if isinstance(self, EnvCommand):
```
neersighted, line 139: This check makes no sense given the type of the command.

src/poetry/utils/requirements.py
```diff
@@ -0,0 +1,133 @@
...
+            )
+
+            requirement["name"] = name
+
+        result.append(requirement)
+
+    return result
+
+
+def create_pool() -> Pool:
```
neersighted, line 126: Let's drop this helper and keep things inline -- it's less messy than stashing this helper in such an obscure place, and this isn't the PR to solve the ergonomics of creating the pool in anyway. If you want to clean things up and be a bit more specific, you can just use `PyPiRepository.search()` instead of creating a pool with one repository.

neersighted (changes requested):
Some more tweaks -- there is a lot of moving code around in this PR in the same commit it was changed, so it's a bit hard to review.

I think we should block this on https://github.com/python-poetry/poetry/pull/6669 as the new `AbstractRepository` introduced in that PR will make all the functions that currently take `Pool` (which just has `PyPiRepository` in it anyway) much cleaner and allow for dropping `Pool` until such time that we actually support searching sources other than PyPI.

src/poetry/utils/requirements.py
juhoautio: > If you want to clean things up and be a bit more specific, you can just use PyPiRepository.search() instead of creating a pool with one repository.

I'm all for cleaning things up, but not sure what it means to replace this with `PyPiRepository.search()` -> moving the code inline. Also, good to avoid additional refactoring in this PR's scope.

juhoautio:
> there is a lot of moving code around in this PR in the same commit it was changed, so it's a bit hard to review.

This is unfortunate, yes. The commit to blame for most of the refactoring is https://github.com/python-poetry/poetry/pull/5710/commits/a09c47b38a15c2803c8bec159f857258eac80838. In that commit I am only moving the existing code to functions but not changing it otherwise. 

> I think we should block this on #6669 as the new `AbstractRepository` introduced in that PR will make all the functions that currently take `Pool` (which just has `PyPiRepository` in it anyway) much cleaner and allow for dropping `Pool` until such time that we actually support searching sources other than PyPI.

Looks like that PR was already merged 🎉 I'm sorry but I couldn't figure out what it would mean in practice for this PR, how to replace `Pool` with `AbstractRepository` in these functions? I was about to take `AbstractRepository` in `find_best_version_for_package` but it ends up using `VersionSelector(pool)`. Is it possible to change the signature of `VersionSelector` to take `AbstractRepository` instead of `Pool`?

src/poetry/console/commands/show.py
```diff
@@ -528,7 +528,7 @@ def find_latest_package(
                     return provider.search_for_direct_origin_dependency(dep)
 
         name = package.name
-        selector = VersionSelector(self.poetry.pool)
+        selector = VersionSelector(self.poetry.pool.repositories[0])
```
juhoautio, line 531: What would be the right way to handle this? Moving from Pool to AbstractRepository.

finswimmer:
@juhoautio: Sorry I'm late to the party. I think we should rework the implementation of `poetry new` and `poetry init` in that way, that `poetry new` becomes an extended version of  `poetry init`. Doing this we can makes sure, that `new` have at least the same options as `init`, even if we introduce a new option.

This is why I closed the issue you've linked in your PR description. The new "major" ticket should be #2563 now.

juhoautio:
@finswimmer is someone (you?) going to work on implementing #2563 some time soon? If yes, does that means this PR is rejected after all? But if no one will work on #2563, I would still hope to have this PR merged with the current scope ie. just adding some missing options & fixing some docs without any bigger refactoring.

finswimmer:
Hey @juhoautio ,

this doesn't mean your PR is rejected. The scope is not that far away from #2563 and I was hoping that you maybe like to go that extra mile. :) 

fin swimmer

juhoautio:
I could work on it if you'd like, but if the design changes that much, I think starting from a clean slate with a new PR could be better. I would hope to have some guidance on the desired approach for implementing it. Please let me know.

I'm still wondering if this PR could be finished because it's almost there. Maybe I'm also missing what additional use cases the new alternative will enable compared to this.

Secrus:
Superseded by #9088

PR #5711, opened 2022-05-28
Fix complicated export case

dimbleby:
Related to https://github.com/python-poetry/poetry-core/pull/381 (on which, as currently written, it relies) and https://github.com/python-poetry/poetry-plugin-export/issues/32.

I don't yet have a testcase proving the point for poetry-export-plugin, but will get there eventually...

### DIFF
--- a/src/poetry/packages/locker.py
+++ b/src/poetry/packages/locker.py
@@ -5,7 +5,6 @@ import logging
 import os
 import re
 
-from copy import deepcopy
 from hashlib import sha256
 from pathlib import Path
 from typing import TYPE_CHECKING
@@ -339,7 +338,7 @@ class Locker:
                     for region_marker in region_markers:
                         marker = region_marker.intersect(base_marker)
                         if not marker.is_empty():
-                            require2 = deepcopy(require)
+                            require2 = require.clone()
                             require2.marker = marker
                             dependencies.append(require2)
 
@@ -391,7 +390,7 @@ class Locker:
         if project_python_marker is not None:
             marked_requires: list[Dependency] = []
             for require in project_requires:
-                require = deepcopy(require)
+                require = require.clone()
                 require.marker = require.marker.intersect(project_python_marker)
                 marked_requires.append(require)
             project_requires = marked_requires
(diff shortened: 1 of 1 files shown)


radoering:
What I don't like is that there seems to be no usage and no test for `get_project_dependency_packages()` in this repo. (Of course, that's not your fault but a result of moving export to its own repo.) I think it might really help, if there was a unit test here. However, I don't know how much effort is required for that and it may be too much to ask. (Anyway, I don't see a unit test in this repo being a blocker for merging this.)

dimbleby:
This is of course blocked on there being a poetry-core that provides `constraint_regions()`, happy to wait for that rather than temporarily re-implement it here.

I've made a couple more fixes along the way though.

First, tighten up the code that selects a locked package so that it never can choose overlaps
- this code previously said "if I've already selected a version of this package whose markers overlap with the current markers, then I must make the same choice again"
- but it now also says "if there is more than one previously selected version of this package whose markers overlap with the current markers, then I have gone wrong"
- that is, it detects the case where we've already chosen package p1 with markers m1, p2 with markers m2 (not overlapping m1) and we encounter a requierment for p with markers that overlap both m1 and m2

Alas that, combined with marker fixes in poetry-core, caused one of the poetry-export-plugin testcases to fail.  Getting a truly reliable export requires a more sophisticated algorithm than we have I think: probably backtracking on previously made choices when we hit the sort of problem just described

I don't see that happening any time soon.  So meanwhile I have made the hokey change to prefer earlier versions of packages rather than later.  If you're feeling generous you could call this a heuristic improvement; if ungenerous then you might say that it's just an unprincipled change to make a particular testcase pass.

dimbleby:
releases of poetry-core makes this ready to merge; but it's hard to negotiate the trip hazards set up by the cross-repository testing in the workflows.

The workflow identifies that this is "breaking" the released version of the plugin; though it's really fixing it.  https://github.com/python-poetry/poetry-plugin-export/pull/68 is the corresponding UT change.

dimbleby:
Some notes on disentangling the cross-repository mess in https://github.com/python-poetry/poetry/pull/5980#issuecomment-1184881387

TLDR suggest moving all the remaining export-specific code over to the export plugin.

That would put the code and its tests in the same place, which seems much more sensible.

dimbleby:
This should be ready to merge, modulo "what do we do about cross-repository testing".

radoering:
Superseded by: python-poetry/poetry-plugin-export#94

Issue #5714, opened 2022-05-29
Cannot generate lock file when specifying multiple platform-dependent wheels for a package


alasdairtran:
- [x] I am on the [latest](https://github.com/python-poetry/poetry/releases/latest) Poetry version on the master branch.
- [x] I have searched the [issues](https://github.com/python-poetry/poetry/issues) of this repo and believe that this is not a duplicate.
- [x] If an exception occurs when executing a command, I executed it again in debug mode (`-vvv` option).

- **OS version and name**: Ubuntu 18.04.6 LTS
- **Poetry version**: current master branch
- **Link of a [Gist](https://gist.github.com/) with the contents of your pyproject.toml file**: See below

## Issue

Here's the `pyproject.toml` lock file to reproduce the error

```toml
[tool.poetry]
name = "test"
version = "0.1.0"
description = ""
authors = []

[tool.poetry.dependencies]
python = "~3.10.4"
jaxlib = [
    {url = "https://storage.googleapis.com/jax-releases/cuda11/jaxlib-0.3.10+cuda11.cudnn82-cp310-none-manylinux2014_x86_64.whl", platform = "linux"},
    {url = "https://storage.googleapis.com/jax-releases/mac/jaxlib-0.3.10-cp310-none-macosx_10_9_x86_64.whl", platform = "darwin" },
]

[build-system]
requires = ["poetry-core"]
build-backend = "poetry.masonry.api"

```

Using the latest commit from master:

```
pipx install --suffix=@master 'poetry@git+https://github.com/python-poetry/poetry.git@master'
poetry@master lock
```

generating the lock file fails with error

```
  SolverProblemError

  Because test depends on both jaxlib (0.3.10+cuda11.cudnn82 url) and jaxlib (0.3.10 url), version solving failed.

  at ~/.local/pipx/venvs/poetry@master/lib/python3.10/site-packages/poetry/puzzle/solver.py:159 in _solve
      155│             packages = result.packages
      156│         except OverrideNeeded as e:
      157│             return self.solve_in_compatibility_mode(e.overrides, use_latest=use_latest)
      158│         except SolveFailure as e:
    → 159│             raise SolverProblemError(e)
      160│
      161│         combined_nodes = depth_first_search(PackageNode(self._package, packages))
      162│         results = dict(aggregate_package_nodes(nodes) for nodes in combined_nodes)
      163│
```

[...]

Issue #5447, opened 2022-04-14
Poetry doesn't install the correct dependency according to its supported Python version


roipoussiere:
- [x] I am on the [latest](https://github.com/python-poetry/poetry/releases/latest) Poetry version.
- [x] I have searched the [issues](https://github.com/python-poetry/poetry/issues) of this repo and believe that this is not a duplicate.
- [x] If an exception occurs when executing a command, I executed it again in debug mode (`-vvv` option).

- **OS version and name**: ubuntu 21.10
- **Poetry version**: 1.1.13
- **Link of a [Gist](https://gist.github.com/) with the contents of your pyproject.toml file**: https://gist.github.com/roipoussiere/94e123906e86448e755f04a1ad798d77

## Issue

I have a wheel dependency on specific url for many platforms (Linuw/macOS/Windows) and many Python version (3.8, 3.9, 3.10).

Despite the fact I use Python 3.9, when I try `poetry install`, the python 3.8 version of the wheel is downloaded.

I also tried to:
- replace `python = "^3.8"` by `python = "^3.9"` (same thing);
- replace `python = "~3.8",  markers = "platform_system == 'Linux'",` by `markers = "python_version ~= '3.8' and platform_system == 'Linux'",` (same thing).

Maybe related to #5378 or #5205, I don't think it's the same issue but I'm not sure.

kmsquire:
I've seen a similar problem with something like the following:

```
package = [
  { platform = "linux", python = "~3.7", url = "https://some.url.com/path/to/package-cp37-cp37-linux_x86_64.whl" },
  { platform = "linux", python = "~3.8", url = "https://some.url.com/path/to/package-cp38-cp38-linux_x86_64.whl" },
  { platform = "linux", python = "~3.9", url = "https://some.url.com/path/to/package-cp39-cp39-linux_x86_64.whl" },
  { platform = "darwin", python = "~3.7", url = "https://some.url.com/path/to/package-cp37-cp37-macosx_10_14_x86_64.whl" },
  { platform = "darwin", python = "~3.8", url = "https://some.url.com/path/to/package-cp38-cp38-macosx_10_14_x86_64.whl" },
  { platform = "darwin", python = "~3.9", url = "https://some.url.com/path/to/package-cp39-cp39-macosx_10_14_x86_64.whl" },
]
```

In this case, poetry attempts to install only the first URL--if I look at `poetry.lock`, I see something like

```
[[package]]
name = "package"
version = "1.2.3"
[...]

Issue #4810, opened 2021-11-22
Incorrect locking of platform-specific wheel dependencies


nyuszika7h:
- [x] I am on the [latest](https://github.com/python-poetry/poetry/releases/latest) Poetry version.
- [x] I have searched the [issues](https://github.com/python-poetry/poetry/issues) of this repo and believe that this is not a duplicate.
- [x] If an exception occurs when executing a command, I executed it again in debug mode (`-vvv` option).

- **OS version and name**: Debian 11
- **Poetry version**: 1.1.1 / 1.2.0a2
- **Link of a [Gist](https://gist.github.com/) with the contents of your pyproject.toml file**: https://gist.github.com/nyuszika7h/3b26db37f0cdfe6ecc7f5bd9e2522c06

## Issue

When given the above `pyproject.toml` file, running `poetry lock` on Linux generates the following `poetry.lock`:
```toml
[[package]]                                      
name = "test"                                  
version = "1.0"                                  
description = ""                                 
category = "main"                                
optional = false                                 
python-versions = "*"                            
                                                 
[package.source]                                 
type = "file"                                    
url = "wheels/test-1.0-cp34-abi3-win_amd64.whl"
```

Also, `poetry export` outputs:
```
test @ file:///home/nyuszika7h/example/wheels/test-1.0-cp34-abi3-win_amd64.whl; sys_platform == "win32" or sys_platform == "linux" or sys_platform == "darwin"                               
```

The expected behavior would be to include 3 separate entries, one for each platform.

Interestingly enough, `poetry install` is smart enough to pick the correct version for the platform despite the incorrect locking. But the lock file is incomplete, and the requirements.txt output is just completely wrong.

nyuszika7h:
If the dependencies are provided as URLs rather than local files, it seems like it's even worse: `poetry install` actually picks the incorrectly locked Windows version for Linux as well.

teaVeloper:
I also have a similar issue with `poetry 1.2.0b1` locking some dependencies as windows and creating a `requirements.txt` mentioning windows, and thus install failed on a system I can only use `requirements.txt` - `poetry 1.1.13` did resolve the same `pyproject.toml` correctly.

If helpful, I can create an Issue for that with more details.

PR #5715, opened 2022-05-29
Support for multiple-constraint direct origin dependencies with same version

radoering:
Resolves: #4810
Resolves: #5447
Resolves: #5714

Prior to #5640 dependencies with same version from different sources were merged, which led to incorrect lock files. However, according to #4810 at least in some cases the installer was able to mitigate the incorrect lock files by installing a suitable wheel anyway.

With #5640 instead of generating incomplete lock files a solver error is raised. Although, in terms of correctness that's an improvement, it could also be considered a minor regression by some users. This PR, improves the handling of multiple-constraint direct origin (path, url, git) dependencies fundamentally.

The first commit of this PR enables the provider/solver to handle multiple-constraint direct origin dependencies correctly by considering them as duplicates and not merging them by version anyway. However, that is not enough to create correct lock files because the installer discards direct origin dependencies with same name and version as well.

In order to enable the installer to handle this kind of dependencies a new Repository type, which can distinguish between multiple direct origin dependencies with the same version, is introduced.

Further, a consistent order of direct origin dependencies in the lock file is ensured.

### DIFF
--- a/src/poetry/installation/installer.py
+++ b/src/poetry/installation/installer.py
@@ -12,6 +12,7 @@ from poetry.installation.pip_installer import PipInstaller
 from poetry.repositories import Pool
 from poetry.repositories import Repository
 from poetry.repositories.installed_repository import InstalledRepository
+from poetry.repositories.lockfile_repository import LockfileRepository
 from poetry.utils.extras import get_extra_package_names
 from poetry.utils.helpers import canonicalize_name
 from poetry.utils.helpers import pluralize
@@ -107,9 +108,7 @@ class Installer:
             self._write_lock = False
             self._execute_operations = False
 
-        local_repo = Repository()
-
-        return self._do_install(local_repo)
+        return self._do_install()
 
     def dry_run(self, dry_run: bool = True) -> Installer:
         self._dry_run = dry_run
@@ -204,14 +203,14 @@ class Installer:
         ):
             ops = solver.solve(use_latest=[]).calculate_operations()
 
-        local_repo = Repository()
-        self._populate_local_repo(local_repo, ops)
+        lockfile_repo = LockfileRepository()
+        self._populate_lockfile_repo(lockfile_repo, ops)
 
-        self._write_lock_file(local_repo, force=True)
+        self._write_lock_file(lockfile_repo, force=True)
 
         return 0
 
-    def _do_install(self, local_repo: Repository) -> int:
+    def _do_install(self) -> int:
         from poetry.puzzle.solver import Solver
 
         locked_repository = Repository()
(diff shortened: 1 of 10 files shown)


src/poetry/puzzle/provider.py
```diff
@@ -612,38 +612,28 @@ def complete_package(self, package: DependencyPackage) -> DependencyPackage:
...
                 dependencies.append(deps[0])
                 continue
 
-            extra_keys = ", ".join(k for k in key[1:] if k is not None)
-            dep_name = f"{key[0]} ({extra_keys})" if extra_keys else key[0]
-
             self.debug(f"<debug>Duplicate dependencies for {dep_name}</debug>")
 
-            deps = self._merge_dependencies_by_marker(deps)
+            _deps: list[Dependency] = []
```
neersighted, line 627: Not really a fan of this name -- maybe `indirect_origin_deps`or `non_direct_origin_deps` would help make this clearer?

src/poetry/puzzle/provider.py
```diff
@@ -612,38 +612,28 @@ def complete_package(self, package: DependencyPackage) -> DependencyPackage:
...
+            direct_origin_deps: list[Dependency] = []
+            for dep in deps:
+                if dep.is_direct_origin():
+                    direct_origin_deps.append(dep)
+                else:
+                    _deps.append(dep)
+            deps = self._merge_dependencies_by_marker(_deps)
```
neersighted, line 634: Also not a fan of this style of chaining -- I'd rather just wrap the calls.

src/poetry/installation/installer.py
```diff
@@ -204,14 +203,14 @@ def _do_refresh(self) -> int:
         ):
             ops = solver.solve(use_latest=[]).calculate_operations()
 
-        local_repo = Repository()
+        local_repo = LockfileRepository()
```
neersighted, line 206: I think the introduction of a new type here makes the code/intention a fair bit more obvious as an ancillary benefit. I think we should probably go ahead and change the use of `local_repo` to `lockfile_repo` in this file to capitalize on that further.

neersighted (changes requested):

src/poetry/puzzle/provider.py
radoering: Not sure I understood. Please have a look if my change covers what you meant.

neersighted: Yep, that's what I meant!

neersighted (approved):
LGTM, but one last question -- does the base `Repository` class have any direct instances anymore, or could we migrate it to be an abstract base class?

radoering:
The base class still has several (more than 10) direct instances.

Issue #3766, opened 2021-03-10
`poetry update --dry-run` modifies poetry.lock


martinmo:
- [x] I am on the [latest](https://github.com/python-poetry/poetry/releases/latest) Poetry version.
- [x] I have searched the [issues](https://github.com/python-poetry/poetry/issues) of this repo and believe that this is not a duplicate.
- [x] If an exception occurs when executing a command, I executed it again in debug mode (`-vvv` option).

- **OS version and name**: macOS 10.15.7
- **Poetry version**: 1.1.5
- **Link of a [Gist](https://gist.github.com/) with the contents of your pyproject.toml file**: https://gist.github.com/martinmo/bf32a5e6278f9edc74eb3f7787e5b372

## Issue

When `poetry update` is called with the `--dry-run` argument, the expected outcome is that the file `poetry.lock` remains unmodified.

However, a regression introduced in poetry version 1.1.1 (it seems, via PR #3034) modifies this file regardless of the `--dry-run` flag.

To reproduce using the above gist:

```
$ git clone https://gist.github.com/bf32a5e6278f9edc74eb3f7787e5b372.git poetry-bug && cd poetry-bug
$ poetry -V
Poetry version 1.1.5
$ poetry config --list
cache-dir = "/Users/martin/Library/Caches/pypoetry"
experimental.new-installer = false
installer.parallel = false
virtualenvs.create = true
virtualenvs.in-project = true
virtualenvs.path = "{cache-dir}/virtualenvs"  # /Users/martin/Library/Caches/pypoetry/virtualenvs
$ poetry update --dry-run
Updating dependencies
Resolving dependencies... (0.1s)

Writing lock file

Package operations: 1 install, 0 updates, 0 removals

  - Installing defusedxml (0.7.1)
$ git status
On branch master
Your branch is up to date with 'origin/master'.

Changes not staged for commit:
  (use "git add <file>..." to update what will be committed)
  (use "git restore <file>..." to discard changes in working directory)
        modified:   poetry.lock

no changes added to commit (use "git add" and/or "git commit -a")
```

With poetry 1.1.0, it works as expected:

```
$ git checkout poetry.lock
$ pipx install --suffix @1.1.0 'poetry==1.1.0'
$ poetry@1.1.0 -V
Poetry version 1.1.0
$ poetry@1.1.0 update --dry-run
Updating dependencies
Resolving dependencies... (0.1s)

Package operations: 1 install, 0 updates, 0 removals

[...]

Issue #3666, opened 2021-02-09
poetry remove --dry-run modifies poetry.lock


samestep:
- [x] I am on the [latest](https://github.com/python-poetry/poetry/releases/latest) Poetry version.
- [x] I have searched the [issues](https://github.com/python-poetry/poetry/issues) of this repo and believe that this is not a duplicate.
- [x] If an exception occurs when executing a command, I executed it again in debug mode (`-vvv` option).

- **OS version and name**: macOS Big Sur Version 11.2
- **Python version**: Python 3.8.3
- **Poetry version**: Poetry version 1.1.4
- **Link of a [Gist](https://gist.github.com/) with the contents of your pyproject.toml file**: https://gist.github.com/samestep/7fab4df48b7b4f89ea5fe5c29f7d4692

## Issue

As the title states, running `poetry remove --dry-run` modifies `poetry.lock`. Example:
```sh
poetry new remove-dry-run-bug
cd $_
sed -i '' 's/^authors = \[".*"\]/authors = []/' pyproject.toml
cat pyproject.toml # see gist
poetry add pyyaml
cp poetry.lock{,.old}
poetry remove --dry-run pyyaml
git diff --no-index poetry.lock{.old,} | tee poetry.lock.diff # see gist
```
This causes subsequent commands (such as `poetry show`) to fail.

getup8:
Yeah I'm on the latest version of poetry and just ran:

`poetry add djangorestframework@latest --dry-run` just to test what it would do and it updated `poetry.lock` (it did not update the actual package locally).

This seems pretty bad no?

PR #3767, opened 2021-03-10
Don't change poetry.lock during `poetry update --dry-run`

martinmo:
# Pull Request Check List

Resolves: #3766

- [x] Added **tests** for changed code.
- [x] Updated **documentation** for changed code.

AFAIK, there is no relevant documentation to be changed, am I right?

PR #5718, opened 2022-05-29
Ensure `--dry-run` option doesn't update `poetry.lock` for `add`, `update` and `remove`

mkniewallner:
# Pull Request Check List

Resolves: #3666
Resolves: #3766

Update logic of `_write_lock_file` to ensure that `--dry-run` option of `add`, `update` and `remove` commands leave `poetry.lock` intact.

This supersedes #3767, since it has become stale, while adding missing tests for `--dry-run` arguments on the commands.

- https://github.com/python-poetry/poetry/commit/a0c605a5824a0ec4538056c64f40a78313e628b7 adds tests to show that [they currently fail](https://github.com/python-poetry/poetry/runs/6644967841?check_suite_focus=true)
- other commits fix the issue to make them pass

### DIFF
--- a/src/poetry/installation/installer.py
+++ b/src/poetry/installation/installer.py
@@ -337,7 +337,7 @@ class Installer:
         # Execute operations
         return self._execute(ops)
 
-    def _write_lock_file(self, repo: Repository, force: bool = True) -> None:
+    def _write_lock_file(self, repo: Repository, force: bool = False) -> None:
         if force or (self._update and self._write_lock):
             updated_lock = self._locker.set_lock_data(self._package, repo.packages)
 
--- a/src/poetry/installation/installer.py
+++ b/src/poetry/installation/installer.py
@@ -338,7 +338,7 @@ class Installer:
         return self._execute(ops)
 
     def _write_lock_file(self, repo: Repository, force: bool = False) -> None:
-        if force or (self._update and self._write_lock):
+        if self._write_lock and (force or self._update):
             updated_lock = self._locker.set_lock_data(self._package, repo.packages)
 
             if updated_lock:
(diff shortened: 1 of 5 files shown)


src/poetry/installation/installer.py
```diff
@@ -337,8 +337,8 @@ def _do_install(self, local_repo: Repository) -> int:
         # Execute operations
         return self._execute(ops)
 
-    def _write_lock_file(self, repo: Repository, force: bool = True) -> None:
-        if force or (self._update and self._write_lock):
+    def _write_lock_file(self, repo: Repository, force: bool = False) -> None:
+        if self._write_lock and (force or self._update):
```
mkniewallner, line 341: This was a suggestion from @radoering [here](https://github.com/python-poetry/poetry/pull/3767/files#r842891790), and looking at the usage of the method, it does seem that the only place where we use `force` leaves `_write_lock` to `True`, so this seems to make sense.

tests/console/commands/test_add.py
```diff
@@ -1999,34 +2014,54 @@ def test_add_with_lock_old_installer(
...
     )
-    original_content = app.poetry.file.read()
+    original_pyproject_content = poetry_with_up_to_date_lockfile.file.read()
+    original_lockfile_content = poetry_with_up_to_date_lockfile._locker.lock_data
 
     repo.add_package(get_package("cachy", "0.2.0"))
+    repo.add_package(get_package("docker", "4.3.1"))
 
-    tester.execute("cachy --dry-run")
+    tester.execute("cachy")
```
mkniewallner, line 2033: Updated the test here because since we're testing a keyboard interrupt, it may be nice to test that the non-dry-run variant prevents updates on the files.

neersighted (approved):

PR #3676, opened 2021-02-10
Feat: Added option to disable SSL verify

maayanbar13:
# Pull Request Check List
Resolves: #1556
 
* [x]  Added **tests** for changed code.
* [x]  Updated **documentation** for changed code.

 I'm opening this new pull request to replace [/pull/2912](https://github.com/python-poetry/poetry/pull/2912) since @Celeborn2BeAlive is no longer following the issue.

Issue #1556, opened 2019-11-08
Option to disable SSL verify


pawamoy:
- [x] I have searched the [issues](https://github.com/sdispater/poetry/issues) of this repo and believe that this is not a duplicate.
- [x] I have searched the [documentation](https://poetry.eustace.io/docs/) and believe that my question is not covered.

## Feature Request

I'm trying to use poetry in a corporate environment. We have a private server and index for packages, and conda is setup to not verify SSL. Unfortunately, I didn't find a similar option or configuration for Poetry, so when I try to install a package with Poetry, it fails (SSLError).

I managed to get it to work by changing two lines in

https://github.com/sdispater/poetry/blob/51c7042160a74adf14038460468e5e5a72b0d965/poetry/repositories/legacy_repository.py#L415-L426

...to this:

```diff
     def _download(self, url, dest):  # type: (str, str) -> None
-        r = self._session.get(url, stream=True)
+        r = self._session.get(url, stream=True, verify=False)
         with open(dest, "wb") as f:
             for chunk in r.iter_content(chunk_size=1024):
                 if chunk:
                     f.write(chunk)

     def _get(self, endpoint):  # type: (str) -> Union[Page, None]
         url = self._url + endpoint
-        response = self._session.get(url)
+       response = self._session.get(url, verify=False)
         if response.status_code == 404:
             return

         return Page(url, response.content, response.headers)
```

Obviously we would use a value specified in the `config.toml` instead of a literal `False`.

pawamoy:
I gues other modifications will be needed to be able to publish as well.

pawamoy:
After talking with some colleagues, it seems that a better solution is to install the Certificate Authorities (CA) of your corporation on your system and configure your tools to use it, instead of disabling SSL verification (which is bad?).

There is a great answer on how to do this for Windows or Linux on this [stackoverflow post](https://stackoverflow.com/questions/39356413/how-to-add-a-custom-ca-root-certificate-to-the-ca-store-used-by-pip-in-windows) 

All this should be dealt with by the system administrators I guess, instead of the developers themselves on each of their machines... but hey, at least I learned something!

I'm leaving this issue open since there were some upvotes, but I don't consider it myself a priority anymore.

blunt1973:
really need this option too

enicklas:
We have exactly the same issue. It would be great to have this option, similar to pip's trusted-host

evstratbg:
+

[...]

PR #5719, opened 2022-05-29
config: allow bool values for repo cert

abn:
boolean values in addition to certificate paths. This allows for repositories to skip TLS certificate validation for cases where self-signed certificats are used by package sources.

In addition to the above, the certificate configuration handling has now been delegated to a dedicated dataclass.

Closes: #3676
Resolves: #1556

### DIFF
--- a/tests/installation/test_pip_installer.py
+++ b/tests/installation/test_pip_installer.py
@@ -14,6 +14,7 @@ from poetry.core.packages.package import Package
 from poetry.installation.pip_installer import PipInstaller
 from poetry.repositories.legacy_repository import LegacyRepository
 from poetry.repositories.pool import Pool
+from poetry.utils.authenticator import RepositoryCertificateConfig
 from poetry.utils.env import NullEnv
 
 
@@ -21,6 +22,7 @@ if TYPE_CHECKING:
     from pytest_mock import MockerFixture
 
     from poetry.utils.env import VirtualEnv
+    from tests.conftest import Config
 
 
 @pytest.fixture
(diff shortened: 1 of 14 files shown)


neersighted (approved):
LGTM -- I especially like that PyPI is tamperproof the way you've added this, as well as the dataclass encapsulation of invariants/configuration.

neersighted:
Things look a little awkward with the docs nesting, but I can live with it:

image

Secrus (dismissed):

abn:
> Things look a little awkward with the docs nesting, but I can live with it:

Yeah. But couldn't think of a better way of grouping it.

tests/installation/test_pip_installer.py
```diff
@@ -204,3 +206,32 @@ def copy_only(source: Path, dest: Path) -> None:
...
+        source_url=default.url,
+    )
+
+    installer.install(foo)
+
+    assert len(null_env.executed) == 1
+    cmd = null_env.executed[0]
+    assert "--trusted-host" in cmd
+    cert_index = cmd.index("--trusted-host")
+    # Need to do the str(Path()) bit because Windows paths get modified by Path
```
Secrus, line 236: Does this comment belong here? I don't see such code anywhere around

neersighted (approved):

tests/installation/test_pip_installer.py
neersighted: This is an orphan after the fix for the failing Windows tests -- good eye.

abn: It's copied over from #3676, will remve it.

Secrus (approved):

neersighted (approved):

Dimaqa:
Still have error with version 1.2.0b2

```
Could not find a suitable TLS CA certificate bundle, invalid path: false
```

markfaine:
This doesn't resolve the problem,  because false is added as a string in auth.toml

```
  [certificates]
  [certificates.cp_util]
  cert = "false"
```

It should be:

```
  [certificates]
  [certificates.cp_util]
  cert = false
```
I suppose this command is the culprit.  It's probably casting the command line argument to a string since it would have always been a string prior to this commit. 

`poetry config certificates.cp_util.cert false`

PR #5738, opened 2022-05-31
Refactor `show` command a bit

mkniewallner:
# Pull Request Check List

Refactor `show` command a tiny bit to separate the different cases into dedicated functions, since right now `handle` method is 300 lines long.
This also has the advantage of making some imports necessary only on a specific variant of the command.

### DIFF
--- a/src/poetry/console/commands/show.py
+++ b/src/poetry/console/commands/show.py
@@ -165,9 +165,7 @@ lists all packages available."""
             ["<info>description</>", f" : {pkg.description}"],
         ]
 
-        table = self.table(style="compact")
-        table.add_rows(rows)
-        table.render()
+        self.table(rows=rows, style="compact").render()
 
         if pkg.requires:
             self.line("")
--- a/src/poetry/console/commands/show.py
+++ b/src/poetry/console/commands/show.py
@@ -115,15 +115,7 @@ lists all packages available."""
 
         # Show tree view if requested
         if self.option("tree") and package is None:
-            requires = root.all_requires
-            packages = locked_repo.packages
-            for p in packages:
-                for require in requires:
-                    if p.name == require.name:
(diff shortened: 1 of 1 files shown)


neersighted (dismissed):

neersighted:
Looks like this needs a rebase for current master.

mkniewallner:
Since the rebase, `reverse_deps` method was added, and I'm wondering if it should not be moved outside of the command module (in `utils` for instance), since it could be useful in other parts of the code.

At least maybe marking it as protected would avoid importing it from outside the module, if we want to keep it there?

neersighted:
@mkniewallner can we rebase this so we can merge it for 1.3? Also, (though it doesn't have to be part of this PR), how feasible would it be to add symbols in addition to colors to communicate the status of table entries, without splitting the logic and rendering code apart?

mkniewallner:
> @mkniewallner can we rebase this so we can merge it for 1.3?

Will do that tomorrow, I kinda forgot about this one since the last rebase.

> Also, (though it doesn't have to be part of this PR), how feasible would it be to add symbols in addition to colors to communicate the status of table entries, without splitting the logic and rendering code apart?

Good question, I'll have to dig a bit more into the codebase. There are other things that could be nice to have in the command output (liked yanked packages, if some are installed), so maybe we should rethink the command output as a whole, since there may be more information we want to convey compared to previous versions.

neersighted:
Sounds good. Long term I'd like to split the traversal/logic and rendering code apart so we can have color-only, color-with-symbols, symbol-only, and JSON output on every command. But that would be a much more ambitious refactor, and this is good incremental progress in cleaning up the beast of a command that `show` is.

mkniewallner:
Rebased, so this should be ready to review again.

Issue #5729, opened 2022-05-31
How do I make part of and auto-enable plugins in my project


Maciej-Zwolinski:
- [x] I have searched the [issues](https://github.com/python-poetry/poetry/issues) of this repo and believe that this is not a duplicate.

## Issue

system: Ubuntu-20.04
poetry: 'master'

I need `poetry-dotenv-plugin` to be a part of my project (i need to set up environment variables to point to proper installation of CUDA, and don't want the user to have to do that manually or modify .bashrc in any way). The documentation is silent on how to make a plugin part of the project's dependency - simply adding `poetry-dotenv-plugin = '*'` in `pyproject.toml`'s dependencies section and running `poetry run python3 -c 'import os; print(os.environ)'` does not return modified environment. What is an expected syntax here?

Secrus:
Hi. There is info in docs about how to add plugins to your poetry instance: https://python-poetry.org/docs/master/cli/#self-add. Could you confirm that it works for you?

Maciej-Zwolinski:
Hi, I want the `poetry install` command to add plugins specified (how, I do not know, that's what I'm asking) in my `pyproject.toml`. 

Is that possible? Otherwise a user would need to run `poetry install && poetry self add poetry-dotenv-plugin` to be able to fully utilize the package.

Maciej-Zwolinski:
Ok, thank you, that's very helpful. 

Is there a way to define plugins in a config of some sort so they can be enabled with a single `poetry` command?

danielknell:
looks like the PoC in  https://github.com/python-poetry/poetry/pull/5740 would solve this?

Kobold:
@Maciej-Zwolinski Would love to see this too!

> Plugins are not part of the package you are making, they are for your Poetry instance.

@Secrus The use case I, @Maciej-Zwolinski, and I suspect others are looking address are workflow issues across all developers and all Poetry instances across a project. Management of common tasks ([nat-n/poethepoet](https://github.com/nat-n/poethepoet)), common environments ([mpeteuil/poetry-dotenv-plugin](https://github.com/mpeteuil/poetry-dotenv-plugin)), precommit ([vstrimaitis/poetry-pre-commit-plugin](https://github.com/vstrimaitis/poetry-pre-commit-plugin)) and such are things individual developers may want to customize — but most projects want to have shared common ways to address these needs.

[...]

Issue #5037, opened 2022-01-14
(🎁) Support isolated dependencies that install like pipx


KotlinIsland:
- [x] I have searched the [issues](https://github.com/python-poetry/poetry/issues) of this repo and believe that this is not a duplicate.
- [x] I have searched the [documentation](https://python-poetry.org/docs/) and believe that my question is not covered.

## Feature Request

pseudo pyproject.toml:
```toml
[tool.poetry.dev-dependencies]
pytest = "6.0.0"
black = { version = "^21.12b0", isolated = true }
```
OR:
```toml
[tool.poetry.group.test]
pytest = "6.0.0"
black = { version = "^21.12b0", isolated = true }
```

I have dev dependencies that are never imported by my source code or my test code, they act more as standalone cli applications. I still want to the version to be locked down, and provide a one click install for developers working with my project, but I don't want the possibility of that dependency interfering with dependency resolution or accidentally becoming imported due to it being in the pythonpath.

When declared with the optional `isolated` properties, I want Poetry to install the dependency like [pipx](https://github.com/pypa/pipx) would, completely isolated from the other dependencies.

branchvincent:
There's a draft PR at #5740, feedback is welcome there

rockandska:
This is how I expect to see this feature works and explained it originally in #7421 

```toml
[tool.poetry.dependencies]
python = ">= 3.7, <= 3.11"

[tool.poetry.group.lint]
optional = true
# depends_on by default is [ 'main' ]
depends_on = []
           
[tool.poetry.group.lint.dependencies]
python = ">= 3.7, <= 3.8"
flake8 = "*"

[tool.poetry.group.typing]
optional = true
depends_on = [ 'main' ]

[tool.poetry.group.typing.dependencies]
mypy = "*"
```

In this example :
- lint is only intended to be executed with python 3.7 or 3.8
- typing is intended to be executed with the same python versions the project has
- lint is isolated, so dependencies required by flake8 (`importlib-metadata<5`) doesn't influence other packages in other groups if they don't depend on lint
- `poetry install --with lint` will fail because `lint` does not have main in `depends_on`
- when doing `poetry install --only typing`, it will install `main` too since it depends on it

PR #5740, opened 2022-05-31
poc: isolated project plugins & tools

abn:
This is a proof of concept for discussions and feedback only.

### Rationale
In a lot of case, a project might require a plugin that is otherwise not required anywhere else for a user. This can in theory be defined within the `pyproject.toml` as opposed to being manual instructions for the user.

### Rationale
Projects usually require additional tools like `pre-commit` and `tox`. Today, these tools are part of the project's dependency graph. While this is useful for cases like `pytest`, it is not so much the case for `tox` and/or `pre-commit` that are used as cli tools and not used within the project code.

## Implementation Details
In this poc, when `poetry install` is run, we 

1. create a `.poetry` directory if either `tools` or `plugins` are defined.
2. create a plugins directory (`.poetry/plugins`) to which plugin wheels will be installed (without polluting the project venv or the global system site)
3. create a `tools/<name>` virtual environment with only the tool and any additional dependencies specified.
4. create a `.poetry/config.toml` to store hashes, sytem python version and poetry version to detect changes.

## Example Configuration
In this poc we use `overlay` section to avoid validation errors.

```toml
[overlay]
plugins = [
    "poethepoet"
]

[overlay.tools]
tox = [
    "tox",
    "tox-venv",
]
pre-commit = [
    "pre-commit"
]
```

This can use used as follows.

```sh
poetry install
poetry poe --help
poetry run tox --version
poetry run pre-commit --version
```

## Limitations
1. We use `pip install` to demonstrate capability. This can be replaced with Poetry's internals.
2. Plugins are all installed to a single directory (this can be changed)

### DIFF
--- a/src/poetry/plugins/plugin_manager.py
+++ b/src/poetry/plugins/plugin_manager.py
@@ -42,6 +42,10 @@ class PluginManager:
     def get_plugin_entry_points(
         self, env: Env | None = None
     ) -> list[entrypoints.EntryPoint]:
+        from poetry.utils.env import EnvManager
+
+        EnvManager.load_project_plugins()
+
         entry_points: list[entrypoints.EntryPoint] = entrypoints.get_group_all(
             self._group, path=env.sys_path if env else sys.path
         )
--- a/src/poetry/utils/env.py
+++ b/src/poetry/utils/env.py
@@ -1,9 +1,11 @@
 from __future__ import annotations
 
 import base64
+import dataclasses
 import hashlib
 import itertools
 import json
+import logging
 import os
 import platform
 import re
@@ -34,6 +36,7 @@ from poetry.core.toml.file import TOMLFile
 from poetry.core.utils.helpers import temporary_directory
 from virtualenv.seed.wheels.embed import get_embed_wheel
 
+from poetry.__version__ import __version__
 from poetry.utils._compat import decode
 from poetry.utils._compat import encode
 from poetry.utils._compat import list_to_shell_command
@@ -217,6 +220,8 @@ if site.check_enableusersite() and hasattr(obj, "install_usersite"):
 
 print(json.dumps(paths))
 """
+
+logger = logging.getLogger(__name__)
 
 
 class SitePackages:
@@ -504,6 +509,11 @@ class InvalidCurrentPythonVersionError(EnvError):
         super().__init__(message)
 
 
+@dataclasses.dataclass
+class OverlayConfig:
+    poetry: Poetry
+
+
 class EnvManager:
     """
     Environments manager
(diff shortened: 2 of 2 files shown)


Kobold:
Looking forward to seeing how the Poetry plugin concept evolves! This PR makes a lot of sense, although the implementation details are pretty fancy. I made a comment on issue #5729 on how a blessed `npm run` equivalent may be a 70% solution to this with a much smaller surface area commitment from Poetry:

https://github.com/python-poetry/poetry/issues/5729#issuecomment-1237511646

aucampia:
Would really love to have this, for now I'm using tox for this ([ref](https://github.com/nipunn1313/mypy-protobuf/issues/442#issuecomment-1380405613)) but it is quite hacky.

aldencolerain:
I really like the idea of having plugins on a per-project basis specified in `pyproject.toml` or `poetry.toml`.  The plugin system is fantastic, however we also commonly have projects that require additional tooling: ci, tasks, env, etc.  It quickly becomes complicated when these need to be documented for every project and setup globally.

kubijo:
Is there any idea of further movement / next steps on this?

DaStoll:
Would absolutely love this feature!

radoering:
tl;dr: I split off the plugins part so that you can find a full blown implementation for project plugins in #9547.

Long story:

I took a closer look at this and I think there are some not so small TODOs to bring this over the finish line:

- In the PoC, plugins and tools are only strings that are passed to pip. However, I think it makes more sense to allow the same syntax as for `tool.poetry.dependencies` when installing via poetry instead of pip. (In my opinion, we definitively need the option to specify sources!)
- What about locking? In the PoC, there is no locking. I think for plugins that is ok. For tools, I'm not that sure. Without the PoC, you put tools in some dependency group so that they are locked. In this PoC, you trade locked tools in a common project venv for unlocked tools in isolated venvs.
- Further, I think there is some overlap between tools and isolated groups as described in #5037. Using isolated groups for tools combined with  #9427 would have the nice side effect that we could put locked depenencies for the tools (resp. isolated groups) into poetry.lock without the risk of mixing a tool's dependencies with other dependencies.

I think we should split this in two parts: project plugins on the one side and tools on the other for the following reasons:

- Although both are similar enough in this PoC, they are different enough that concentrating on one of both reduces the necessary work to bring it over the finish line a lot.
- Considering that we want to remove `poetry-plugin-export` from poetry's dependencies and project plugins may mitigate the consequences, some of us consider project plugins as a blocker for 2.0 so it has a higher priority for me. (Tools can still be added in a minor release later.)

MrLoh:
@radoering is there any indication when the new `tool.poetry.requires-plugins` functionality will be released? It seems like it is not available in 1.8.4 yet.

radoering:
See #9448, it should be only a matter of weeks.

abn:
@radoering thanks for running with it btw :purple_heart: and I agree with your assessment, the PoC was done around 2 years ago - a lot has also changed under the hood that makes more sense to split the objectives. I will close this once we have something on the tools front.

Issue #5020, opened 2022-01-10
Support save prefix configuration


barraponto:
- [x] I have searched the [issues](https://github.com/python-poetry/poetry/issues) of this repo and believe that this is not a duplicate.
- [x] I have searched the [documentation](https://python-poetry.org/docs/) and believe that my question is not covered.

## Feature Request

Right now, installing any dependency will save it to pyproject.toml with a caret `^` symbol, indicating it would allow minor updates but not major ones. That's a nice policy, but maybe users want to default to a stricter policy (`~`, allowing only patch updates) or a looser one (`>=`, effectively limiting only on a lower boundary). NPM allows for this with a [save-prefix](https://docs.npmjs.com/cli/v8/using-npm/config#save-prefix) configuration. I think we should do the same.

bmarroquin:
Poetry already supports this with the [add command](https://python-poetry.org/docs/cli/#add) but it could be documented better.

`
poetry add pendulum@~2.0.5
poetry add pendulum@^2.0.5
poetry add pendulum>=2.0.5
`

zEdS15B3GCwq:
Changing the default is not the same as changing the behaviour for individual added packages. Imagine you're adding a dozen, and need to specify each individually.

I like the idea of an option like this. I think, however, there should be 2 options, one for dependencies in general, and one for the Python version. Minor versions of Python have pretty large differences so packages are typically bound to minor and not major versions. Having caret (^) as default for the python dependency (e.g. `python = [^3.10]`) often causes trouble because of this.

<details>
  <summary>Simple example how it goes wrong (click to expand)</summary>
  
```
(test) C:\Stuff\Code\test>pip list
Package    Version
---------- -------
pip        22.1.2
setuptools 62.3.2

(test) C:\Stuff\Code\test>poetry init

This command will guide you through creating your pyproject.toml config.

Package name [test]:
Version [0.1.0]:
Description []:
Author [me <me@me.me>, n to skip]:  n
License []:
Compatible Python versions [^3.10]:

Would you like to define your main dependencies interactively? (yes/no) [yes] no
Would you like to define your development dependencies interactively? (yes/no) [yes] no
Generated file

[tool.poetry]
name = "test"
version = "0.1.0"
description = ""
authors = ["Your Name <you@example.com>"]
readme = "README.md"

[tool.poetry.dependencies]
python = "^3.10"


[build-system]
requires = ["poetry-core"]
build-backend = "poetry.core.masonry.api"


[...]

PR #5741, opened 2022-06-01
Config option to set default python version prefix

zEdS15B3GCwq:
Resolves: #5020 partially

Python minor versions often introduce significant changes. As such, many packages are tied to specific minor versions, e.g. currrent `numba` requires Python >=3.7, <3.11.

By default, `poetry` is more permissive - new projects are initialised with the caret (^) dependency marker for Python, allowing minor version changes. This can quickly lead to failed solutions as the project version range (e.g. ^3.10 meaning >=3.10, <4.0) is wider than what some packages allow (see `numba`).

I think it would be reasonable to change the whole default behaviour, but that would be a major change and I'm sure there's a good reason why it's ^ now. Still, it would be good to at least be able to configure the default.

This new option, `default-python-prefix` allows the default marker to be changed. The options are, currently, limited to **^** and **~** as I was trying to avoid introducing anything complicated (asterisk etc.).

#5020 is a broader request for an option that would affect all packages, so this is just a partial solution, but I believe the Python version should be handled separately anyway.

<details>
  <summary> minimalistic example: project based on Python 3.10 </summary>

```
(poetry_dev) C:\Lib\poetry>poetry --version
Poetry (version 1.2.0b2.dev0)

(poetry_dev) C:\Lib\poetry>poetry config --list
cache-dir = "C:\\Users\\R\\AppData\\Local\\pypoetry\\Cache"
default-python-prefix = null
experimental.new-installer = true
experimental.system-git-client = false
installer.max-workers = null
installer.no-binary = null
installer.parallel = true
virtualenvs.create = false
virtualenvs.in-project = null
virtualenvs.options.no-setuptools = false
virtualenvs.path = "C:\\Users\\R\\Virtualenvs"
virtualenvs.prefer-active-python = false

(poetry_dev) C:\Lib\poetry>poetry config default-python-prefix ~

(poetry_dev) C:\Lib\poetry>cd ..
(poetry_dev) C:\Lib>mkdir test
(poetry_dev) C:\Lib>cd test

(poetry_dev) C:\Lib\test>poetry init

This command will guide you through creating your pyproject.toml config.

Package name [test]:
Version [0.1.0]:
Description []:
Author [me <me@me.me>, n to skip]:  n
License []:
Compatible Python versions [~3.10]:

Would you like to define your main dependencies interactively? (yes/no) [yes] n
Would you like to define your development dependencies interactively? (yes/no) [yes] n
Generated file

[tool.poetry]
name = "test"
version = "0.1.0"
description = ""
readme = "README.md"

[tool.poetry.dependencies]
python = "~3.10"


[build-system]
requires = ["poetry-core"]


Do you confirm generation? (yes/no) [yes]

(poetry_dev) C:\Lib\test>poetry config default-python-prefix *

  RuntimeError

  "*" is an invalid value for default-python-prefix

  at C:\Lib\poetry\src\poetry\console\commands\config.py:323 in _handle_single_value
      319│             raise RuntimeError("You can only pass one value.")
      320│
      321│         value = values[0]
      322│         if not validator(value):
    → 323│             raise RuntimeError(f'"{value}" is an invalid value for {key}')
      324│
      325│         source.add_property(key, normalizer(value))
      326│
      327│         return 0
```
</details>

Yes, the Python version specification can be changed during the initialisation process, but some people may want to *always* use a stricter specification, and setting it manually each time always makes me think why there's no option to change this.

Edit: removed formatting inside code section


src/poetry/console/commands/config.py
```diff
@@ -252,8 +257,8 @@ def handle(self) -> int | None:
                     password = self.secret("Password:")
                 elif len(values) != 2:
                     raise ValueError(
-                        "Expected one or two arguments "
-                        f"(username, password), got {len(values)}"
+                        "Expected one or two arguments (username, password), got"
```
Secrus, line 260: Unrelated and not needed

src/poetry/console/commands/init.py
```diff
@@ -154,7 +154,14 @@ def handle(self) -> int:
...
         if not python:
             current_env = SystemEnv(Path(sys.executable))
-            default_python = "^" + ".".join(
+
+            from poetry.config.config import Config
+
+            config = Config.create()
+            prefix = config.get("default-python-prefix")
+            if prefix is None:
+                prefix = "^"
```
Secrus, line 163: I would prefer to see it as a method. `handle` methods are already pretty big, let's avoid putting even more into them.

src/poetry/console/commands/init.py
```diff
@@ -327,8 +334,8 @@ def _determine_requirements(
                 # no constraint yet, determine the best version automatically
                 if package and "version" not in constraint:
                     question = self.create_question(
-                        "Enter the version constraint to require "
-                        "(or leave blank to use the latest version):"
+                        "Enter the version constraint to require (or leave blank to use"
```
Secrus, line 337: Unrelated and less readable

src/poetry/console/commands/init.py
```diff
@@ -451,8 +458,8 @@ def _validate_author(self, author: str, default: str) -> str | None:
         m = AUTHOR_REGEX.match(author)
         if not m:
             raise ValueError(
-                "Invalid author string. Must be in the format: "
-                "John Smith <john@example.com>"
+                "Invalid author string. Must be in the format: John Smith"
```
Secrus, line 461: Unrelated and less readable

tests/console/commands/test_init.py
```diff
@@ -822,10 +822,8 @@ def test_init_non_interactive_existing_pyproject_add_dependency(
     repo.add_package(get_package("foo", "1.19.2"))
 
     tester.execute(
-        "--author 'Your Name <you@example.com>' "
-        "--name 'my-package' "
-        "--python '^3.6' "
-        "--dependency foo",
+        "--author 'Your Name <you@example.com>' --name 'my-package' --python '^3.6'"
```
Secrus, line 825: Unrelated and much less readable

tests/test_factory.py
```diff
@@ -304,8 +304,8 @@ def test_validate_fails():
     content["this key is not in the schema"] = ""
 
     expected = (
-        "Additional properties are not allowed "
-        "('this key is not in the schema' was unexpected)"
+        "Additional properties are not allowed ('this key is not in the schema' was"
```
Secrus, line 307: Unrelated and less readable

Secrus (changes requested):

src/poetry/console/commands/config.py
```diff
@@ -227,8 +232,8 @@ def handle(self) -> int | None:
                 return 0
 
             raise ValueError(
-                "You must pass the url. "
-                "Example: poetry config repositories.foo https://bar.com"
+                "You must pass the url. Example: poetry config repositories.foo"
```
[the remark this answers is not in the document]
zEdS15B3GCwq: Thanks for the quick response. I'll look into why these unrelated changes are in the code. I didn't do them manually, perhaps one of the pre-commit hooks (black?) did. It was sloppy of me not to notice.

src/poetry/console/commands/init.py
zEdS15B3GCwq: OK

tests/test_factory.py
zEdS15B3GCwq: I'll try to revert these formatting changes.

src/poetry/console/commands/init.py
zEdS15B3GCwq: I'm pushing a new commit that does this in its own method.

neersighted:
> @neersighted OK, I got what you mean, a change in the default would be the solution I prefer too.
> 
> I'm not sure that the meaning of "application" vs "library" init is very self-explanatory, though.
> 
> Do you plan to integrate the changes you mention into 1.1 as well? If not, how about adding this option there, while 1.2 and up would provide a different solution?

Just realized I never got back to you. What I mean by application vs library is the two dominant types of Python codebase have different priorities (e.g. application wants locked deps, library wants flexible ranges) and sane defaults (such as Python constraint, etc) -- it may make sense to add a flag to `poetry init` that sets some of these defaults (such as Python constraint prefix), or it may make sense to add a new field to `[tool.poetry]` that changes defaults in Poetry.

Regardless, we definitely won't be adding this as a new feature to 1.1 -- that branch is bugfixes only, and a new feature like this is out of scope (not to mention the fact that we have finite resources and are trying to direct all energies to 1.2).

zEdS15B3GCwq:
> Just realized I never got back to you. What I mean by application vs library is the two dominant types of Python codebase have different priorities (e.g. application wants locked deps, library wants flexible ranges) and sane defaults (such as Python constraint, etc) -- it may make sense to add a flag to `poetry init` that sets some of these defaults (such as Python constraint prefix), or it may make sense to add a new field to `[tool.poetry]` that changes defaults in Poetry.

While I understand the general idea, the *library* approach would only work if the project has no dependencies with stricter Python version requirements. My main issue with "^" is that it includes yet non-existent minor versions, and a lot of libraries are restricted to a few minor versions, most likely not including future ones. Just as in my example above, `numba` is limited to a few latest **existing** minor versions. The moment a library depends on a package like that, `poetry` will fail due to a mismatch in Python versions.

I think there's a theoretical problem with how `poetry` considers Python version requirements. A project may support many different versions, but that doesn't mean that all dependencies support all of those versions. And in case of an actual installation of dependencies, poetry should find a solution that *works now* instead of trying to find one that fulfills all theoretical, *mathematical* requirements. For example, my library supports all Python versions 3.x, but right now Python 3.10 is active and I want to install into this environment - in such a case, `poetry` shouldn't complain that my project allows 3.9999999 (because of "^3.x" requirement), it should realise that a solution is necessary for 3.10. What my project supports and what the solution is needed for are two different things. Perhaps I don't know poetry well enough and this is a non-issue, it's just me who's using it the wrong way?

PR #5745, opened 2022-06-01
tests: enable parallel test execution

abn:


.github/workflows/main.yml
```diff
@@ -88,13 +88,13 @@ jobs:
...
+        run: poetry run python -m pytest -n auto -p no:sugar -q tests/
 
       - name: Run pytest (integration suite)
         env:
           POETRY_TEST_INTEGRATION_GIT_USERNAME: ${GITHUB_ACTOR}
           POETRY_TEST_INTEGRATION_GIT_PASSWORD: ${{ secrets.GITHUB_TOKEN }}
-        run: poetry run python -m pytest -p no:sugar -q --integration tests/integration
+        run: poetry run python -m pytest -n auto -p no:sugar -q --integration tests/integration
 
       - name: Get Plugin Version (poetry-plugin-export)
```
mkniewallner, line 99: Should we also use parallelism on `poetry-plugin-export` tests?
Though, similarly to integration tests, they take [~15 seconds](https://github.com/python-poetry/poetry/runs/6694707029?check_suite_focus=true#step:18:9) to run without it, which is already pretty fast, so maybe not worth doing so?

pyproject.toml
```diff
@@ -75,6 +75,7 @@ pytest = "^7.1"
 pytest-cov = "^3.0"
 pytest-mock = "^3.5"
 pytest-sugar = "^0.9"
+pytest-xdist = "^2.5"
```
mkniewallner, line 78: Could it be worth including `psutil` extra, as specified in [pytest-xdist documentation](https://pytest-xdist.readthedocs.io/en/latest/index.html#installation)?

```suggestion
pytest-xdist = { version = "^2.5", extras = ["psutil"] }
```

The benefits are not listed in the documentation, but as explained in https://github.com/pytest-dev/pytest-xdist/issues/676, `psutil` is useful for detecting logical CPUs, which may (or may not?) be useful on GitHub Actions.

abn: Wasn't sure if it was needed. Will have to dig into a bit more.

.github/workflows/main.yml
abn: Yeah. The overhead didn't seem worth it.

abn: And need to make sure those tests work correctly under xdist.

.cirrus.yml
```diff
@@ -27,7 +27,7 @@ test_task:
     - poetry config virtualenvs.in-project true
   test_script:
     - poetry install
-    - poetry run pytest -q --junitxml=junit.xml tests
+    - poetry run pytest -n auto -q --junitxml=junit.xml tests
```
neersighted, line 30: I think this change landed in the wrong commit

neersighted (changes requested):
The first commit is invasive enough that I don't really see a point in having two separate changes -- let's just squash them down.

.cirrus.yml
abn: Let's squash it.

abn:
Feel free to squash on merge.

mkniewallner (approved):
Besides the fact that tests will now run faster locally and on the CI, those are also some really nice changes for the test suite itself. Good job :slightly_smiling_face:

neersighted (approved):

Issue #5597, opened 2022-05-12
install-poetry.py crashes on Python 3.11 beta 1


davidfstr:
Repro steps:

```
$ wget https://raw.githubusercontent.com/python-poetry/poetry/master/install-poetry.py
$ python3 install-poetry.py 
[...]
Installing Poetry (1.1.13): Creating environment
Traceback (most recent call last):
  File "/Users/davidf/Projects/trycast/install-poetry.py", line 908, in <module>
    sys.exit(main())
             ^^^^^^
  File "/Users/davidf/Projects/trycast/install-poetry.py", line 876, in main
    return installer.run()
           ^^^^^^^^^^^^^^^
  File "/Users/davidf/Projects/trycast/install-poetry.py", line 507, in run
    self.install(version)
    ^^^^^^^^^^^^^^^^^^^^^
  File "/Users/davidf/Projects/trycast/install-poetry.py", line 528, in install
    with self.make_env(version) as env:
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Library/Frameworks/Python.framework/Versions/3.11/lib/python3.11/contextlib.py", line 137, in __enter__
    return next(self.gen)
           ^^^^^^^^^^^^^^
  File "/Users/davidf/Projects/trycast/install-poetry.py", line 600, in make_env
    raise e
    ^^^^^^^
  File "/Users/davidf/Projects/trycast/install-poetry.py", line 586, in make_env
    yield VirtualEnvironment.make(env_path)
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/davidf/Projects/trycast/install-poetry.py", line 292, in make
    builder.ensure_directories(target)
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Library/Frameworks/Python.framework/Versions/3.11/lib/python3.11/venv/__init__.py", line 119, in ensure_directories
    if os.pathsep in env_dir:
       ^^^^^^^^^^^^^^^^^^^^^
TypeError: argument of type 'PosixPath' is not iterable
```

**Your environment**

* CPython 3.11 beta 1

I'm shepherding a fix to upstream Python 3.11: https://github.com/python/cpython/issues/92675

davidfstr:
[...]

PR #5749, opened 2022-06-01
Test against Python 3.11

mkniewallner:
# Pull Request Check List

- ~[ ] Added **tests** for changed code.~ Not applicable
- ~[ ] Updated **documentation** for changed code.~ Not applicable

Since Python 3.11 reached a beta state, it would be nice to start running the test suite against it, to ensure that Poetry works as expected on this version when the final version is out.

It was not really feasible to do that earlier with `3.11.0b1` because of https://github.com/python-poetry/poetry/issues/5597 (and the associated upstream issue), and would have required a workaround with `3.11.0b2` because of https://github.com/pytest-dev/pytest/issues/10008.

`3.11.0b3` was [released](https://www.python.org/downloads/release/python-3110b3/) quickly after `3.11.0b2`, which fixes the problem in `3.11.0b2`, so we can now test against Python 3.11.

The good news is that the tests [all pass](https://github.com/python-poetry/poetry/runs/6715337505?check_suite_focus=true#step:14:1) on Python 3.11 :tada: (though there are some deprecation warnings).

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -28,7 +28,7 @@ jobs:
     strategy:
       matrix:
         os: [Ubuntu, macOS, Windows]
-        python-version: ["3.7", "3.8", "3.9", "3.10"]
+        python-version: ["3.7", "3.8", "3.9", "3.10", "3.11.0-beta.3"]
         include:
           - os: Ubuntu
             image: ubuntu-22.04
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -28,7 +28,7 @@ jobs:
     strategy:
       matrix:
         os: [Ubuntu, macOS, Windows]
-        python-version: ["3.7", "3.8", "3.9", "3.10", "3.11.0-beta.3"]
+        python-version: ["3.7", "3.8", "3.9", "3.10", "3.11-dev"]
         include:
           - os: Ubuntu
             image: ubuntu-22.04
--- a/.github/workflows/skip.yml
+++ b/.github/workflows/skip.yml
@@ -28,6 +28,6 @@ jobs:
     strategy:
       matrix:
         os: [Ubuntu, macOS, Windows]
-        python-version: ["3.7", "3.8", "3.9", "3.10", "3.11.0-beta.3"]
+        python-version: ["3.7", "3.8", "3.9", "3.10", "3.11-dev"]
     steps:
       - run: exit 0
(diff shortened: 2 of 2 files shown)


mkniewallner:
> @mkniewallner
> 
> > Once `3.11.0b3` is available on GitHub Actions, I'll update the PR to use it and remove the workaround.
> 
> You can use the `-dev` suffix, i.e. `3.11-dev`: https://github.com/actions/setup-python#available-versions-of-python

Interesting, I was not aware of that, thanks!

Though by using `-dev`, won't this will still look for the latest alpha/beta release available in https://github.com/actions/python-versions/blob/main/versions-manifest.json, or is it doing something else?

The downside I can see with using `-dev` is that having a new alpha/beta release could out of the blue break the CI workflow (although the upside is that you automatically pick up the last version without needing to change anything, which is nice if a new version works as expected).

edgarrmondragon:
> Though by using `-dev`, won't this will still look for the latest alpha/beta release available in https://github.com/actions/python-versions/blob/main/versions-manifest.json, or is it doing something else?

It's doing just doing that, but the advantage is you don't need to go back and update the version in your workflows. Once `3.11.0b3` is available in `python-versions`, `3.11-dev` will point to it.

.github/workflows/main.yml
```diff
@@ -28,7 +28,7 @@ jobs:
     strategy:
       matrix:
         os: [Ubuntu, macOS, Windows]
-        python-version: ["3.7", "3.8", "3.9", "3.10"]
+        python-version: ["3.7", "3.8", "3.9", "3.10", "3.11.0-beta.3"]
```
neersighted, line 31: I'd like to use `3.11-dev` as our version here unless there's a compelling reason to pin to one beta version -- a moving target is arguably a better smoke test for us than a pinned release.

.github/workflows/skip.yml
```diff
@@ -28,6 +28,6 @@ jobs:
     strategy:
       matrix:
         os: [Ubuntu, macOS, Windows]
-        python-version: ["3.7", "3.8", "3.9", "3.10"]
+        python-version: ["3.7", "3.8", "3.9", "3.10", "3.11.0-beta.3"]
```
neersighted, line 31: As above.

neersighted (changes requested):

neersighted (approved):

PR #5754, opened 2022-06-02
Typechecking tweaks

dimbleby:
Just tightening up a handful of annotations

### DIFF
--- a/src/poetry/puzzle/provider.py
+++ b/src/poetry/puzzle/provider.py
@@ -12,7 +12,6 @@ from collections import defaultdict
 from contextlib import contextmanager
 from pathlib import Path
 from typing import TYPE_CHECKING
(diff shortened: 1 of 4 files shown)


src/poetry/puzzle/provider.py
```diff
@@ -125,7 +125,7 @@ def __init__(
         self,
         package: Package,
         pool: Pool,
-        io: Any,
+        io: NullIO,
```
neersighted, line 128: I'm reasonably sure this should be a `cleo.io.IO` -- unless for some reason we never pass a real IO object in?

dimbleby: I agree, I guess mypy doesn't notice because cleo isn't exporting its types

PR #5756, opened 2022-06-02
cleo-typechecking

dimbleby:
fixing some - but by no means all - of the errors that would be exposed if cleo revealed its types.  On a `Command`

- `self._io` may be None, there's a property that gives the guaranteed-not-None thing
- `self.handle()` is supposed to return an `int`

### DIFF
--- a/src/poetry/console/commands/about.py
+++ b/src/poetry/console/commands/about.py
@@ -15,7 +15,7 @@ class AboutCommand(Command):
 
     description = "Shows information about Poetry."
 
-    def handle(self) -> None:
+    def handle(self) -> int:
         from poetry.utils._compat import metadata
 
         # The metadata.version that we import for Python 3.7 is untyped, work around
@@ -34,3 +34,5 @@ Poetry-Core Version: {version('poetry-core')}</info>
 See <fg=blue>https://github.com/python-poetry/poetry</> for more information.</comment>\
 """
         )
+
+        return 0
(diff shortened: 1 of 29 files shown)


neersighted (approved):

neersighted:
It's also worth noting that Cleo is now under the Poetry org, and we're going to try to get its typing tightened and exported if you'd like to help there :)

neersighted:
> > It's also worth noting that Cleo is now under the Poetry org, and we're going to try to get its typing tightened and exported if you'd like to help there :)
> 
> On the whole I think that cleo (and tomlkit) are mis-steps, not bringing enough value compared to the much more popular alternatives being used by the rest of the python ecosystem. I don't consider that time spent polishing either of them would be time well spent.
> 
> Ripping cleo out of poetry is probably too much at this point, but maybe tomlkit would be feasible...

I will have to disagree re: cleo, as Cleo is unique (at least at the time it was created) for being a CLI toolkit that has first-class null-io/testing support as well as a really easy to use markup/styling system. Poetry is the only major user, but it's stable enough and adds enough value that I think it's more than worth keeping, unless someone has something that supports every major feature and improves upon them.

Re: TomlKit, I'll also have to disagree there as there is not another TOML implementation in the ecosystem that preserves formatting/order to allow ergonomic combinations of hand-editing and programmatic manipulation of a TOML file. TomlKit is also used by other projects in the ecosystem (most notably PDM, but also other smaller downstream consumers) and is actively maintained by outside contributors as well.

dimbleby:
Reasonable people can disagree on these things!  Either way, I have no plan to get involved in cleo typechecking

Issue #4903, opened 2021-12-16
Can not install 'stylus': EnvCommandError


amhndu:
- [x] I am on the [latest](https://github.com/python-poetry/poetry/releases/latest) Poetry version.
- [x] I have searched the [issues](https://github.com/python-poetry/poetry/issues) of this repo and believe that this is not a duplicate.
- [x] If an exception occurs when executing a command, I executed it again in debug mode (`-vvv` option).

- **OS version and name**: OpenSUSE Tumbleweed with Python 3.8.12, also tested on Ubuntu 20.04
- **Poetry version**: 1.1.12
- **Link of a [Gist](https://gist.github.com/) with the contents of your pyproject.toml file**: https://gist.github.com/amhndu/4f2a92a1687f6dfac0e458f61e4c0d08

## Issue

I can not install [stylus](https://pypi.org/project/stylus/) with poetry on Linux. (tried on both opensuse and ubuntu). It appears to try installing some macos package and fails. Using regular `pip install stylus` however works. Also, explicitly adding the source tar.gz link works.

Steps to reproduce:
```sh
$ mkdir stylusmvp && cd stylusmvp
$ poetry init -n
$ poetry add stylus
```

Verbose logs: [here](https://gist.github.com/amhndu/a90e3faa2aaa06428935c96ba1a3f1e1)

finswimmer:
Hello @amhndu,

seems to the same issue as in #3649.

Can you test my PR #3656?

```
pix install poetry --suffix=@fix 'poetry @ git+https://github.com/finswimmer/poetry.git@issue/3649-package-types'
poetry@fix init -n
poetry@fix add stylus
```

amhndu:
That works! 
(Aside: had to remove that suffix flag).

Closing this issue since it's a duplicate.

Issue #3649, opened 2021-02-05
poetry unable to install hbmqtt


FlorianLudwig:
- [x] I am on the [latest](https://github.com/python-poetry/poetry/releases/latest) Poetry version.
- [x] I have searched the [issues](https://github.com/python-poetry/poetry/issues) of this repo and believe that this is not a duplicate.
- [x] If an exception occurs when executing a command, I executed it again in debug mode (`-vvv` option).

- **OS version and name**: Linux
- **Poetry version**: 1.1.4
- **Link of a [Gist](https://gist.github.com/) with the contents of your pyproject.toml file**: 

## Issue

poetry is unable to install hbmqtt.

```
# poetry add -vvv hbmqtt

[...]

      Traceback (most recent call last):
        File "<string>", line 1, in <module>
        File "/usr/lib64/python3.9/tokenize.py", line 392, in open
          buffer = _builtin_open(filename, 'rb')
      FileNotFoundError: [Errno 2] No such file or directory: '/tmp/pip-req-build-38j5rmex/setup.py'
```

Doing `pip install hbmqtt` works fine.

flyte:
Thanks for issuing the PR to fix this, but in the mean time, is there a workaround available at all? I'd like to publish a new version of my project to PyPI but without changing back to using `setup.py` I can't.

FlorianLudwig:
To be honest, I didn't spent time investigating a workaround as my workaround is to fork hbmqtt :]

PR #3656, opened 2021-02-07
fix: only consider package links for sdist and bdist_wheels

finswimmer:
Only follow and lock links for packages of type `sdist` or `bdist_wheel` in PyPi repository.

Once accepted and merge this should go to master as well.

Closes: https://github.com/python-poetry/poetry/issues/3649
Closes: https://github.com/python-poetry/poetry/issues/4903

# Pull Request Check List

- [ ] Added **tests** for changed code.
- [ ] Updated **documentation** for changed code.

PR #5767, opened 2022-06-04
fix: only consider package links for sdist and bdist_wheels

finswimmer:
Only follow and lock links for packages of type `sdist` or `bdist_wheel` in PyPi repository.

Closes: https://github.com/python-poetry/poetry/issues/3649
Closes: https://github.com/python-poetry/poetry/issues/4903

(This is a port of https://github.com/python-poetry/poetry/pull/3656.)

### DIFF
--- a/src/poetry/repositories/pypi_repository.py
+++ b/src/poetry/repositories/pypi_repository.py
@@ -25,10 +25,11 @@ cache_control_logger.setLevel(logging.ERROR)
 
 logger = logging.getLogger(__name__)
 
-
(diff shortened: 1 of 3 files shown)


neersighted (changes requested):
LGTM, but I'd like to see at least one or two regression tests validating this behavior in both `find_links_for_package` and `_get_release_info` -- assuming that we can mock those without things getting too tangled.

neersighted (approved):

Issue #5727, opened 2022-05-31
poetry add "requests[security]" fails on Poetry 1.2.0b1


thailow:
If I run `poetry add "requests[security]"` on Poetry 1.2.0b1, it fails with this message:

 Package 'requests' is listed as a dependency of itself.

I downgraded to Poetry 1.1.13 where it works. Thanks for all your work!

finswimmer:
Hello @thailow,

I guess your project itself is named `requests`. Have a look at your `pyproject.toml` at the keyword `name` in the `[tool.poetry]` section.

Naming the project the same as a dependency is not allowed.

fin swimmer

finswimmer:
Hey, 

then please show your `pyproject.toml` so that one can reproduce your issue.

fin swimmer

thailow:
Sure, here it is:

`
[tool.poetry]
name = "cross_piano_client"
version = "0.1.27"
description = ""
authors = ["Full Name <full.name@email.com>"]

[tool.poetry.dependencies]
python = "3.9.13"
dependency-injector = "^4.38.0"
deprecated = "^1.2.13"
requests = {extras = ["security"], version = "^2.27.1"}

[tool.poetry.dev-dependencies]
pytest = "^7.1.2"

[build-system]
requires = ["poetry-core>=1.0.0"]
build-backend = "poetry.core.masonry.api"

[[tool.poetry.source]]
name = "gitlab_shared"
url = "https://gitlab.com/api/v4/groups/1241251/-/packages/pypi/simple"
`

dimbleby:
Still all works fine for me...

Hard to know what to suggest in that case, maybe clear out your cache and see if that helps?  `rm -fr ~/.cache/pypoetry` or local equivalent

thailow:
Strange.. I am managing the .venv in the projects (virtualenvs.in-project = true), hence, should not rely too much on any cache. Is it possible that there is some interference with pyenv? 

There is some other issue which I always face (and I do not know it this may be related as well). When I do a `poetry install`, it does not pick up the exact Python version specified in the `pyproject.toml` (and also not pyenv's version in .python-version). Rather, I always need to do a `poetry env use 3.9.13` or similar to enforce this.

dimbleby:
afaik the way that you create your virtual environment has nothing to do with poetry's cache.

klnrdknt:
[...]

PR #5769, opened 2022-06-04
Fix the solver adding a package as a dependency of itself when extras are involved

Darsstar:
# Pull Request Check List

Resolves: #5727 

### DIFF
--- a/src/poetry/puzzle/solver.py
+++ b/src/poetry/puzzle/solver.py
@@ -164,21 +164,20 @@ class Solver:
             if package.features:
                 for _package in packages:
                     if (
-                        _package.name == package.name
-                        and not _package.is_same_package_as(package)
+                        not _package.features
+                        and _package.name == package.name
                         and _package.version == package.version
                     ):
                         for dep in package.requires:
-                            if dep.is_same_package_as(_package):
+                            # Prevent adding base package as a dependency to itself
+                            if _package.name == dep.name:
                                 continue
 
                             if dep not in _package.requires:
                                 _package.add_dependency(dep)
-
-                continue
-
-            final_packages.append(package)
-            depths.append(results[package])
+            else:
+                final_packages.append(package)
+                depths.append(results[package])
 
         # Return the packages in their original order with associated depths
         return final_packages, depths
--- a/tests/puzzle/test_solver.py
+++ b/tests/puzzle/test_solver.py
@@ -793,6 +793,38 @@ def test_solver_finds_extras_next_to_non_extras(
     )
 
 
+def test_solver_merge_extras_into_base_package_multiple_repos_fixes_5727(
+    solver: Solver, repo: Repository, pool: Pool, package: ProjectPackage
+):
+    package.add_dependency(
+        Factory.create_dependency("A", {"version": "*", "source": "legacy"})
+    )
+    package.add_dependency(Factory.create_dependency("B", {"version": "*"}))
+
+    package_a = get_package("A", "1.0")
+    package_a.extras = {"foo": []}
+
+    repo.add_package(package_a)
+
+    package_b = Package("B", "1.0", source_type="legacy")
+    package_b.add_dependency(package_a.with_features(["foo"]).to_dependency())
+
+    package_a = Package("A", "1.0", source_type="legacy")
+    package_a.extras = {"foo": []}
+
+    repo = Repository("legacy")
+    repo.add_package(package_a)
+    repo.add_package(package_b)
+
+    pool.add_repository(repo)
+
+    transaction = solver.solve()
+
+    ops = transaction.calculate_operations(synchronize=True)
+
+    assert len(ops[0].package.requires) == 0, "a should not require itself"
+
+
 def test_solver_returns_prereleases_if_requested(
     solver: Solver, repo: Repository, package: ProjectPackage
 ):


src/poetry/puzzle/solver.py
```diff
@@ -164,12 +164,13 @@ def _solve(
...
-                        _package.name == package.name
-                        and not _package.is_same_package_as(package)
+                        not _package.features
+                        and package.provides(_package)
                         and _package.version == package.version
                     ):
                         for dep in package.requires:
-                            if dep.is_same_package_as(_package):
+                            # Prevent adding base package as a dependency to itself
+                            if _package.provides(dep):
```
neersighted, line 173: @radoering @dimbleby PTAL -- I can see from reading the code that this is indeed correct given the current implementation. However, I'm wondering if this is another issue with `is_same_package_as` being brittle, or just a misuse. If I understand the recent discussion and refactoring correctly, this is just a misuse and broke as a result of fixing the overly-aggressive merging of packages.

radoering: LGTM. This is a place where we don't care for `source_type`, so clearly a misuse of `is_same_package_as`.

Maybe, we can change L179-L182 to use `else` instead of `continue`. Imho, that improves readability.

tests/puzzle/test_solver.py
```diff
@@ -793,6 +793,51 @@ def test_solver_finds_extras_next_to_non_extras(
...
+        )
+    )
+    package.add_dependency(
+        Factory.create_dependency(
+            "B",
+            {
+                "version": "*",
+            },
+        )
+    )
```
radoering, line 815: Can we make this a bit more dense? (Just remove the trailing commas and let pre-commit do its work.)

radoering (changes requested):

tests/puzzle/test_solver.py
```diff
@@ -1858,8 +1889,11 @@ def test_solver_does_not_trigger_conflict_for_python_constraint_if_python_requir
     check_solver_result(transaction, [{"job": "install", "package": package_a}])
 
 
-def test_solver_does_not_trigger_conflict_for_python_constraint_if_python_requirement_is_compatible_multiple(  # noqa: E501
-    solver: Solver, repo: Repository, package: Package
+def test_solver_does_not_trigger_conflict_for_python_constraint_if_python_requirement_is_compatible_multiple(
+    # noqa: E501
+    solver: Solver,
+    repo: Repository,
+    package: Package,
```
neersighted, line 1896:
```suggestion
    package: Package
```

Not sure why this trailing comma snuck in, but let's drop it.

neersighted (changes requested):

neersighted (approved):

src/poetry/puzzle/solver.py
```diff
@@ -164,21 +164,20 @@ def _solve(
             if package.features:
                 for _package in packages:
                     if (
-                        _package.name == package.name
-                        and not _package.is_same_package_as(package)
+                        not _package.features
```
dimbleby, line 167: Seems like something is redundant here: `package.provides()` checks the compatibility of features so either we didn't need to check `not _package.features` or we only needed to check that the names (and versions) match, right?

Similarly, since we know that `_package` has no features the later check that uses `_package.provides()` is also just a check on the names, I think?

neersighted: I interpreted the line as being an early return -- yes there's a redundant check but that's because we're re-using `package.provides()` but we only need half the functionality. That being said, I'm not nearly as much of an expert on this code so I'm going to refer to your expertise.

Do you have any reservations about the design, or is it more about clarity of intent given the overlapping checks?

dimbleby: I'm mostly trying to check that I understand it myself!  

I think this code is searching for "base" packages: that is feature-less packages that are otherwise the same as `package`.  If that's right then I would find it slightly clearer to go
```patch
                 for _package in packages:
                     if (
                         not _package.features
-                        and package.provides(_package)
+                        and _package.name == package.name
                         and _package.version == package.version
                     ):
                         for dep in package.requires:
                             # Prevent adding base package as a dependency to itself
-                            if _package.provides(dep):
+                            if _package.name == dep.name:
                                 continue

                             if dep not in _package.requires:
```

so that I don't have to go and check what `package.provides()` means.

But this is picking nits: the change looks plausible, it has a test, it doesn't obviously break anything, it's good for me.

neersighted: Appreciate your insight as always. While I don't mind the use of provides here, I am going to agree that your version makes clear the meaning of this code at a glance. I think that given the issues we've had with confusion over `Package.__eq__`, `Package.__hash__`, `Package.provides()` and `Package.is_same_package_as`, we're better off spelling things out explicitly here to prevent regressions (or more likely confusion, given that we do have some high quality tests with this change). Would you mind submitting your suggestions as a review so that @Darsstar can apply them easily?

src/poetry/puzzle/solver.py
```diff
@@ -164,21 +164,20 @@ def _solve(
...
+                        and package.provides(_package)
```
dimbleby, line 168:
```suggestion
                        and _package.name == package.name
```

src/poetry/puzzle/solver.py
```diff
@@ -164,21 +164,20 @@ def _solve(
...
                         and _package.version == package.version
                     ):
                         for dep in package.requires:
-                            if dep.is_same_package_as(_package):
+                            # Prevent adding base package as a dependency to itself
+                            if _package.provides(dep):
```
dimbleby, line 173:
```suggestion
                            if _package.name == dep.name:
```

src/poetry/puzzle/solver.py
Darsstar: > Maybe, we can change L179-L182 to use `else` instead of `continue`. Imho, that improves readability.

Done

neersighted (approved):

radoering (approved):

PR #5770, opened 2022-06-04
solver: fix special case where a direct origin dependency without extras ...

radoering:
… is requested by the project and the same dependency with extras is requested by another dependency

# Pull Request Check List

Resolves: #5311

### DIFF
--- a/src/poetry/mixology/partial_solution.py
+++ b/src/poetry/mixology/partial_solution.py
@@ -4,6 +4,7 @@ from typing import TYPE_CHECKING
 
 from poetry.mixology.assignment import Assignment
 from poetry.mixology.set_relation import SetRelation
+from poetry.mixology.term import Term
 
 
 if TYPE_CHECKING:
@@ -11,7 +12,6 @@ if TYPE_CHECKING:
     from poetry.core.packages.package import Package
 
     from poetry.mixology.incompatibility import Incompatibility
-    from poetry.mixology.term import Term
 
 
 class PartialSolution:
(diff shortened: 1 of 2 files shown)


neersighted (changes requested):
I'd rather hold off merging this until we bump the core version -- could you rebase this onto poetry-core main in the mean time, and then once we bump core we can switch back?

src/poetry/mixology/partial_solution.py
```diff
@@ -213,6 +214,18 @@ def relation(self, term: Term) -> str:
...
+        positive = self._positive.get(term.dependency.name)
+        if positive is not None:
+            # TODO: replace deepcopy with dep.with_features() when
+            # https://github.com/python-poetry/poetry-core/pull/392 is available
+            dep = deepcopy(positive.dependency)
+            dep._features = frozenset(term.dependency.features)
+            # TODO: remove https://github.com/python-poetry/poetry-core/pull/375
+            dep._extras = frozenset(term.dependency.features)
+            positive_with_features = Term(dep, is_positive=True)
+            self._positive[term.dependency.complete_name] = positive_with_features
```
dimbleby, line 226: this doesn't look like a method that I would expect to be making updates to `self`: is there some other moment at which this should have happened?

radoering: Maybe, you're right. I am not completely satisfied with the solution myself. Will take another look at it.

radoering: Moved it to `_register`. Looks better, doesn't it?

dimbleby: Yes, this looks a much better place for it - thanks!

suggestion: we only need to look for `old_positive_without_features` if the assignment dependency _has_ features: else - which would usually be the case - we're just looking up the same thing twice.  So
```patch
         """
         name = assignment.dependency.complete_name
         old_positive = self._positive.get(name)
-        if old_positive is None:
+        if old_positive is None and assignment.dependency.features:
             old_positive_without_features = self._positive.get(
                 assignment.dependency.name
             )
```
?

neersighted:
> #5782

I think you might have taken a wrong turn -- please elaborate if you think these issues are somehow related (and use a regular comment instead of a review comment, please).

neersighted (approved):

Issue #4864, opened 2021-12-02
Install only dev dependencies


juliuskittler:
- [x] I have searched the [issues](https://github.com/python-poetry/poetry/issues) of this repo and believe that this is not a duplicate.
- [x] I have searched the [documentation](https://python-poetry.org/docs/) and believe that my question is not covered.

## Feature Request

The documentation mentions "The `--dev-only` option is now deprecated. You should use the `--only dev` notation instead." (https://python-poetry.org/docs/master/cli/)

However, this functionality is not actually available in poetry 1.1.11 (see `poetry install --help`). It would be great if you could bring it back (there have been other issues related to this).

finswimmer:
The documentation is a little bit misleading here. The `--dev-only` option was never available in a release, but was already available in the master branch. Because it's a feature we will probably not backport this to the 1.1 branch.

PR #5771, opened 2022-06-04
Remove references to `--dev-only` install option

lafrenierejm:
According to @finswimmer in https://github.com/python-poetry/poetry/issues/4864#issuecomment-1002465303:
> The `--dev-only` option was never available in a release, but was already available in the master branch. Because it's a feature we will probably not backport this to the 1.1 branch.

Given that the option was only ever present when installing Poetry from master and was never available as part of a release, I think it's clearer to just remove all references to it rather than include a deprecation notice that would only be relevant to a tiny fraction of Poetry users.

# Pull Request Check List

Resolves: #4864

### DIFF
--- a/docs/cli.md
+++ b/docs/cli.md
@@ -167,10 +167,6 @@ It's also possible to only install specific dependency groups by using the `only
 poetry install --only test,docs
 ```
 
-{{% note %}}
-The `--dev-only` option is now deprecated. You should use the `--only dev` notation instead.
-{{% /note %}}
-
 See [Dependency groups]({{< relref "managing-dependencies#dependency-groups" >}}) for more information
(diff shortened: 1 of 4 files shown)


neersighted (approved):

lafrenierejm:
@neersighted Can you trigger a redeploy of the docs? It looks like the removed option is still present on the website.

neersighted:
The workflow will run tonight -- we haven't added a manual trigger, though you're welcome to PR a `workflow_dispatch` event onto our workflows if you really feel strongly about maintainers being able to deploy early :smile:

https://github.com/python-poetry/website/blob/main/.github/workflows/deploy-production.yml#L5

PR #5778, opened 2022-06-05
uses python-finder during env activate

### LINKED ISSUE #3520: Natively support `py` launcher on Windows

bmarroquin:
# Pull Request Check List

Resolves #3520

While resolving for python versions in the PATH, python-finder executes a python command to get the python version. It will execute the command for every supported python implementation in the path until it finds the right one. 

On windows, pythons registered per pep-514 take priority over path pythons. When checking registered pythons, it does not execute the version command. It does succeed in finding pythons installed through the MS Store (they are in the path).  Py.exe also prefers registered pythons over MS Store pythons, so the behavior here is consistent. To be clear this doesn't use py.exe. pyenv-win creates a python.bat shim in the path. The bat files are recognized as other pythons in the path so it just works.

I added an experimental flag to enable or disable the python-finder.

I haven't tested this with linux but I figured the implementation wouldn't change much so creating the draft PR. I also expected to just work since python-finder covers this testing.

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -10,14 +10,14 @@ python-versions = ">=2.7, !=3.0.*, !=3.1.*, !=3.2.*, !=3.3.*"
 name = "attrs"
 version = "21.4.0"
 description = "Classes Without Boilerplate"
-category = "dev"
+category = "main"
 optional = false
 python-versions = ">=2.7, !=3.0.*, !=3.1.*, !=3.2.*, !=3.3.*, !=3.4.*"
 
 [package.extras]
-dev = ["coverage[toml] (>=5.0.2)", "hypothesis", "pympler", "pytest (>=4.3.0)", "six", "mypy", "pytest-mypy-plugins", "zope.interface", "furo", "sphinx", "sphinx-notfound-page", "pre-commit", "cloudpickle"]
-docs = ["furo", "sphinx", "zope.interface", "sphinx-notfound-page"]
-tests = ["coverage[toml] (>=5.0.2)", "hypothesis", "pympler", "pytest (>=4.3.0)", "six", "mypy", "pytest-mypy-plugins", "zope.interface", "cloudpickle"]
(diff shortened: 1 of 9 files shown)


Diggsey:
> pyenv-win creates a python.bat shim in the path. The bat files are recognized as other pythons in the path so it just works.

FWIW, this doesn't work for me - either because poetry only recognises `.exe` files, or it tries to execute the files with no-extension that pyenv-win also creates, and then gives an error.

bmarroquin:
@Diggsey, could you provide more info on your python installations. Python-finder executes a "get version" command using the full path. It also checks for both `python.exe` and `python.bat` on windows. `poetry env use 3.x` should just work.

Diggsey:
@bmarroquin, sorry maybe I misunderstood - I thought you were saying that poetry works *without this PR* for python installed via `pyenv-win` because it creates appropriate stubs in the PATH. For me that's not the case. It probably works just fine *with* the changes in this PR.

Secrus:
@bmarroquin thank you for your effort. Since there were many changes to the codebase in the meantime, I will be taking this feature on myself, with a fresh start.

PR #5781, opened 2022-06-06
Fix headers in documentation

mkniewallner:

`Options` headers of `self` commands are currently at the wrong level:
Screenshot from 2022-06-06 09-56-08

Also update headers to remove backticks, in order to be consistent with other commands that don't have them, though if you feel like all commands should have backticks instead, I'd be happy to update that.


mkniewallner:
btw right now the sidebar navigation only displays up to 3 levels of headers. I'm wondering if we should update the website to display 4, but that may clutter a bit the navigation on some pages.

Issue #3377, opened 2020-11-18
Immitate `pip install --no-deps` / Add `--only-root`


Natureshadow:
- [x] I have searched the [issues](https://github.com/python-poetry/poetry/issues) of this repo and believe that this is not a duplicate.
- [x] I have searched the [documentation](https://python-poetry.org/docs/) and believe that my question is not covered.

## Feature Request

It would be great if Poetry could support installing only the root of the project, but leave all dependencies untouched, much like the `--no-deps` option to pip does.

The use case is a development environment where the code being worked on is spread across several packages, and I want to keep several of them editable in the same poetry environment without having to temporarily change/remove a dependency from `pyproject.toml`.

Poetry ould optionally still check whether the installed packages fulfill the dependencies or not, but with such an option that developers will explicitly use, this is not necessary I suppose.

sinoroc:
To make sure I understand, you have tried the following and it does not fit your workflow, right?

```
[tool.poetry.dependencies]
# ...
library = "^0.1.0"

[tool.poetry.dev-dependencies]
library = {path = "../library", develop = true}
```

Natureshadow:
Well, this is a nice hack, but it would require one of manipulating pyproject.toml every time I start developing on such a situation, or forcing relative paths on every developer in the project (given that the library is, or might be, in a different Git repository/whatever).

wom:
I'd like to add a +1 here; sometimes upstream packages are too conservative when pinning dependencies. I'd rather avoid having to add a `pip install --no-dep <package==...` after my initial  `poetry install`.

cglacet:
Does anyone found a way to prevent install sub-dependencies? It would also be interesting when trying to generate documentation for another package (where you only want to be able to read the package's code).

1oglop1:
I have a use case for this as well. 
I'm developing IaC with Pulumi [python] and at the moment `pulumi_aws` package is auto-generated from the terraform package.

Then when it comes to describing complex resources eg. `pulumi_aws.cloudfront.Distribution` which require a lot of nested objects however many of them are also just plain strings that allow only specific values to be set eg: [viewer_protocol_policy](https://www.pulumi.com/docs/reference/pkg/aws/cloudfront/distribution/#viewer_protocol_policy_python) which allowed values are:  `allow-all`, `https-only`, or `redirect-to-https` [API Reference](https://docs.aws.amazon.com/cloudfront/latest/APIReference/API_CacheBehavior.html).

[...]

PR #5783, opened 2022-06-06
Add `--only-root` option to install command, excluding all dependencies.

chadac:
Per the related issue, added the `--only-root` option to the `InstallCommand` (as hinted in the issue) to exclude all dependency groups from install.

This is already [expected behavior on the part of the installer](https://github.com/python-poetry/poetry/blob/master/tests/installation/test_installer.py#L383) so no additional tests are needed to ensure this works with an empty list of dependency groups.

# Pull Request Check List

Resolves: #3377

### DIFF
--- a/docs/cli.md
+++ b/docs/cli.md
@@ -167,6 +167,12 @@ It's also possible to only install specific dependency groups by using the `only
 poetry install --only test,docs
 ```
 
+To only install the project itself with no dependencies, use the `only-root` flag.
+
+```bash
+poetry install --only-root
+```
+
 See [Dependency groups]({{< relref "managing-dependencies#dependency-groups" >}}) for more information
 about dependency groups.
 
--- a/docs/cli.md
+++ b/docs/cli.md
@@ -167,7 +167,7 @@ It's also possible to only install specific dependency groups by using the `only
 poetry install --only test,docs
 ```
 
-To only install the project itself with no dependencies, use the `only-root` flag.
+To only install the project itself with no dependencies, use the `--only-root` flag.
 
 ```bash
 poetry install --only-root
--- a/docs/cli.md
+++ b/docs/cli.md
@@ -226,6 +226,7 @@ option is used.
 * `--without`: The dependency groups to ignore.
 * `--with`: The optional dependency groups to include.
 * `--only`: The only dependency groups to include.
+* `--only-root`: Install only the root project, exclude all dependencies.
 * `--default`: Only include the main dependencies. (**Deprecated**)
 * `--sync`: Synchronize the environment with the locked packages and the specified groups.
 * `--no-root`: Do not install the root package (your project).
--- a/docs/managing-dependencies.md
+++ b/docs/managing-dependencies.md
@@ -159,7 +159,7 @@ poetry install --only main
 {{% /note %}}
 
 {{% note %}}
-If you only want to install the project root, and no other dependencies, you can use
+If you want to install the project root, and no other dependencies, you can use
 the `--only-root` option.
 
 ```bash
--- a/docs/managing-dependencies.md
+++ b/docs/managing-dependencies.md
@@ -158,6 +158,15 @@ poetry install --only main
 ```
 {{% /note %}}
 
+{{% note %}}
+If you only want to install the project root, and no other dependencies, you can use
+the `--only-root` option.
+
+```bash
+poetry install --only-root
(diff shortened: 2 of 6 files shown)


radoering:
Out of curiosity: What happens if I run `poetry install --only-root --no-root`?

chadac:
> Out of curiosity: What happens if I run `poetry install --only-root --no-root`?

Haha, good question :) I tried it out, looks like it's effectively a noop minus creating the lockfile if it isn't present.

neersighted:
> > Out of curiosity: What happens if I run `poetry install --only-root --no-root`?
> 
> Haha, good question :) I tried it out, looks like it's effectively a noop minus creating the lockfile if it isn't present.

I'd like to see a meaningful message for the user when this case happens, e.g. https://github.com/python-poetry/poetry/blob/6d01f074f10ce08af33ce59aa0c5065dac24244d/src/poetry/console/commands/install.py#L77

abn:
Is this something that should apply to all group enabled commands? For example, this won't make much sense for update or export.

chadac:
> > > Out of curiosity: What happens if I run `poetry install --only-root --no-root`?
> > 
> > 
> > Haha, good question :) I tried it out, looks like it's effectively a noop minus creating the lockfile if it isn't present.
> 
> I'd like to see a meaningful message for the user when this case happens, e.g.
> 
> https://github.com/python-poetry/poetry/blob/6d01f074f10ce08af33ce59aa0c5065dac24244d/src/poetry/console/commands/install.py#L77-L83

Good point, I've gone ahead and added logic to catch this similar to that case, and set it up to exit when both options are present.

Should I also have this exit when both `--only-root` and `--with`/`--without`/`--only` are also passed? Right now it's currently a warning.



> Is this something that should apply to all group enabled commands? For example, this won't make much sense for update or export.

True, I've gone ahead and migrated it into `install.py` and modified the logic accordingly.

src/poetry/console/commands/install.py
```diff
@@ -82,6 +96,24 @@ def handle(self) -> int:
             )
             return 1
 
+        if self.option("only-root") and any(
+            [self.option(key) for key in {"with", "without", "only"}]
```
neersighted, line 100:
```suggestion
            self.option(key) for key in {"with", "without", "only"}
```

You should be able to drop the brackets here and just let this become a generator expression instead of a list comprehension.

neersighted:
> Should I also have this exit when both `--only-root` and `--with`/`--without`/`--only` are also passed? Right now it's currently a warning.

I personally prefer an error as well, because an inattentive developer may munge a command line out of history. If they don't notice a warning message they may never understand the unexpected/inconsistent results they get until much later. I'd rather fail first as what the user is asking us to do is nonsensical and we should ask them to re-evaluate their request.

(this is all hypothetical of course, I would never make a silly mistake like re-using a command from history and combining options in a nonsensical manner :smile:)

chadac:
> I personally prefer an error as well, because an inattentive developer may munge a command line out of history. If they don't notice a warning message they may never understand the unexpected/inconsistent results they get until much later. I'd rather fail first as what the user is asking us to do is nonsensical and we should ask them to re-evaluate their request.
> 
> (this is all hypothetical of course, I would never make a silly mistake like re-using a command from history and combining options in a nonsensical manner smile)

Yeah, that's definitely true. Plus I imagine swapping warnings to errors in the future is a whole lot harder than the other way around... might need to keep that in mind for some of my own tools lol

docs/cli.md
```diff
@@ -167,6 +167,12 @@ It's also possible to only install specific dependency groups by using the `only
 poetry install --only test,docs
 ```
 
+To only install the project itself with no dependencies, use the `only-root` flag.
```
neersighted, line 170:
```suggestion
To only install the project itself with no dependencies, use the `--only-root` flag.
```

docs/managing-dependencies.md
```diff
@@ -158,6 +158,15 @@ poetry install --only main
 ```
 {{% /note %}}
 
+{{% note %}}
+If you only want to install the project root, and no other dependencies, you can use
```
neersighted, line 162:
```suggestion
If you want to install the project root, and no other dependencies, you can use
```

neersighted (changes requested):

neersighted (approved):

PR #5789, opened 2022-06-06
ci: use composite action for poetry setup

abn:


Secrus:
Wouldn't it be better to put action in separate repo under `python-poetry` org? That would separate concerns better I think.

neersighted:
> Wouldn't it be better to put action in separate repo under `python-poetry` org? That would separate concerns better I think.

I don't think so as this is intended for use in this repo only. I have been meaning to compare the use of a reusable workflow with this method, as the reusable workflow won't be exposed out of this repo and thus won't tempt people to depend on it.

Secrus:
Superseded by the recent CI changes

PR #5805, opened 2022-06-08
direct access for performance and typechecking

dimbleby:
Revisiting https://github.com/python-poetry/poetry/pull/2149, which seems to have stalled or maybe got into some merge conflict horror.

As it goes, I can't reproduce anything like the performance gains reported over there, which is a shame because I'd really have liked to!

- in the example of the original report, resolution is simply much faster in the first place.  It's hard to tell whether that's because poetry has changed, or because the web of dependencies has changed such that this particular example is easier to resolve.  I am seeing times nearer 2 seconds than 200 seconds.
- in some examples I do have with slower resolves, I just don't see meaningful gains.

However: this surely can't _hurt_ performance...

Even if this doesn't turn out to be a big performance win though, I think that it is still good discipline.  In particular, removing the `__setattr__()` and `__getattr_()` tightens up the typechecking: we no longer have to wait to runtime to find that we have gone wrong.

Actually I'd have been a little scared to make this change without good mypy coverage: but that and the test suite makes me pretty confident that this is safe.

### DIFF
--- a/src/poetry/puzzle/provider.py
+++ b/src/poetry/puzzle/provider.py
@@ -491,42 +491,47 @@ class Provider:
             for dep in dependencies
         ]
 
-    def complete_package(self, package: DependencyPackage) -> DependencyPackage:
-        if package.package.is_root():
-            package = package.clone()
-            requires = package.package.all_requires
-        elif not package.package.is_root() and package.package.source_type not in {
+    def complete_package(
+        self, dependency_package: DependencyPackage
+    ) -> DependencyPackage:
+        package = dependency_package.package
+        dependency = dependency_package.dependency
+
+        if package.is_root():
+            dependency_package = dependency_package.clone()
+            package = dependency_package.package
+            dependency = dependency_package.dependency
+            requires = package.all_requires
+        elif package.source_type not in {
             "directory",
             "file",
             "url",
             "git",
         }:
             try:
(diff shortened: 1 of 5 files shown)


src/poetry/puzzle/provider.py
```diff
@@ -508,10 +511,10 @@ def incompatibilities_for(
         ]
 
     def complete_package(self, package: DependencyPackage) -> DependencyPackage:
-        if package.is_root():
+        if package.package.is_root():
```
neersighted, line 514: Let's do the same `package = dependency_package.package` clarification here, like we did above, instead of typing `package.package` everywhere.

neersighted (changes requested):

src/poetry/puzzle/provider.py
dimbleby: it gets a lot more awkward in this one, because we create new `package = ...` in various branches and I was afraid of messing it up

neersighted: I'd probably call the root one `package =` then and use a new name in each branch for clarity.

dimbleby:
just doing a review of my open MRs and I think this one is good to merge.  (But also it's not urgent; it can wait until after 1.2.0)

neersighted (approved):

PR #5817, opened 2022-06-09
poc: refactor: tests/console/commands/test_show.py

chadac:
I noticed this file had a lot of repeated code due to the need to manually specify each lock file. Added a few methods to drop this down from 1.9k to about 896 LOC, while keeping the tests readable. Putting this up as a POC since I'd love feedback on what can be improved/dropped/maybe the refactor isn't needed in the first place :) I can break it out into multiple PRs for merging

Changes:
* Move common packages (cachy, pendulum, pytest) into fixtures
* Merged some test functions using `pytest.mark.parametrize`
* Add `mock_lock_data`, for building the lockfile based on a list of packages. This cut down most of the repeated code, and helped catch a couple errors (such as a mistaken package dependency added in `test_show_tree_why_package` that led to no functional difference)
* Extend `get_package` to include category and description
* Add `mock_poetry_dependency` to capture the common pattern of having Poetry depend on package X and no version earlier.

# Pull Request Check List

Resolves: N/A

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -3,7 +3,7 @@ ci:
 
 repos:
   - repo: https://github.com/pre-commit/pre-commit-hooks
-    rev: v4.2.0
+    rev: v4.3.0
     hooks:
       - id: trailing-whitespace
       - id: end-of-file-fixer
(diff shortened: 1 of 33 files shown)


Secrus:
Stalled

PR #5818, opened 2022-06-09
fix: fixed experimental.system-git-client reported as unknown setting

finswimmer:
# Pull Request Check List
This PR fixed in issue were the config option `experimental.system-git-client reported` was reported as unknown, when trying to set it via `poetry config`.

It also updates the docs to include this option. 

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -157,6 +157,20 @@ Defaults to one of the following directories:
 - Windows: `C:\Users\<username>\AppData\Local\pypoetry\Cache`
 - Unix:    `~/.cache/pypoetry`
 
+### `experimental.system-git-client`
+
+**Type**: boolean
+
+*Introduced in 1.2.0*
+
+Use system git client backend for git related tasks.
+
+Poetry uses `dulwich` by default for git related tasks to not rely on the availability of a git client.
+
+If you encounter any problems with it, set to `true` to use the system git backend.
+
+Defaults to `false`.
+
 ### `installer.parallel`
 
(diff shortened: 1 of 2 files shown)


docs/configuration.md
```diff
@@ -157,6 +157,18 @@ Defaults to one of the following directories:
 - Windows: `C:\Users\<username>\AppData\Local\pypoetry\Cache`
 - Unix:    `~/.cache/pypoetry`
 
+## `experimental.system-git-client`
```
mkniewallner, line 160:
```suggestion
### `experimental.system-git-client`
```

docs/configuration.md
```diff
@@ -157,6 +157,18 @@ Defaults to one of the following directories:
...
+
+**Type**: boolean
+
```
mkniewallner, line 163: Should we add:
```suggestion

*Introduced in 1.2.0*

```
like we have [here](https://github.com/python-poetry/poetry/blob/c68011d8e5e3172181ea65c556b66a393c09e17c/docs/configuration.md?plain=1#L186)?

mkniewallner (changes requested):

docs/configuration.md
finswimmer: Good catch :+1:

docs/configuration.md
finswimmer: Done

mkniewallner (approved):

neersighted (approved):

Issue #5779, opened 2022-06-06
Poetry export does not respect groups defined in lockfile


jkgenser:
- [x] I am on the [latest](https://github.com/python-poetry/poetry/releases/latest) Poetry version.
- [x] I have searched the [issues](https://github.com/python-poetry/poetry/issues) of this repo and believe that this is not a duplicate.
- [x] If an exception occurs when executing a command, I executed it again in debug mode (`-vvv` option).

- **OS version and name**: Ubuntu 22.04 
- **Poetry version**: Poetry (version 1.2.0b2.dev0), Master@865aa94dd3393076f790d3351f6ae18483673e38
- **Link of a [Gist](https://gist.github.com/) with the contents of your pyproject.toml file**: 

## Issue
When running poetry export from the following pyproject.toml and poetry.lock, specifying various combinations of requirements do not result in an optional group excluded. The difference between the main group and the test group is that one of them specifies an extra.

### Project installing
- [pyproject.toml](https://github.com/IndicoDataSolutions/poetry-export/blob/40064828acc373af0b00769a351208dc5e31a53d/pyproject.toml)
- [poetry.lock](https://github.com/IndicoDataSolutions/poetry-export/blob/40064828acc373af0b00769a351208dc5e31a53d/poetry.lock)

dimbleby:
It looks as though parsing the lockfile does not set `.in_extras` for the pytest dependency, which I think is what I would have expected to happen.  That not-in-extras package is created [here](https://github.com/python-poetry/poetry/blob/374cff5153d7175fdc14171fab4f15df72763094/src/poetry/packages/locker.py#L182).

The export can be fixed by an uglier check that `pytest` is an extra, like this:

```patch
diff --git a/src/poetry/packages/locker.py b/src/poetry/packages/locker.py
index c2a6c35c..d2b3db63 100644
--- a/src/poetry/packages/locker.py
+++ b/src/poetry/packages/locker.py
@@ -319,8 +319,9 @@ class Locker:
             requirement.set_constraint(constraint)

             for require in locked_package.requires:
-                if require.in_extras and locked_package.features.isdisjoint(
-                    require.in_extras
+                if require.is_optional() and not any(
+                    require in locked_package.extras[feature]
+                    for feature in locked_package.features
                 ):
                     continue
```

But I'm not sure whether that's just putting a sticky plaster over the `in_extras` problem, and we'd be better arranging that `pytest` has `.in_extras` set all along.

[...]

PR #5819, opened 2022-06-09
don't export unwanted extras

dimbleby:
Fixes https://github.com/python-poetry/poetry/issues/5779, tested in https://github.com/python-poetry/poetry-plugin-export/pull/73.

Cross-repository unit testing continues to be hairy, good luck finding the right sequence to merge things and keep pipelines happy.

It looks as though `in_extras` is only set for packages in the top-level project extras.  Per https://github.com/python-poetry/poetry/issues/5779#issuecomment-1148011863 it's not clear to me whether that's expected, but the pragmatic thing to do seems to be to do the checking differently here, in a way that works for both cases.

### DIFF
--- a/src/poetry/packages/locker.py
+++ b/src/poetry/packages/locker.py
@@ -272,8 +272,9 @@ class Locker:
             requirement.constraint = constraint
 
             for require in locked_package.requires:
-                if require.in_extras and locked_package.features.isdisjoint(
-                    require.in_extras
+                if require.is_optional() and not any(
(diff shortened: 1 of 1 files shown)


dimbleby:
> good luck finding the right sequence to merge things and keep pipelines happy.

maybe it can be done if I split the export plugin fix into two commits...

dimbleby:
release of poetry-plugin-export 1.0.5 allows the pipeline here to go green, this should be good to merge

PR #5820, opened 2022-06-09
replace entrypoints use with importlib.metadata

abn:


pyproject.toml
```diff
@@ -50,7 +50,6 @@ cachecontrol = { version = "^0.12.9", extras = ["filecache"] }
 cachy = "^0.3.0"
 cleo = "^1.0.0a5"
 crashtest = "^0.3.0"
-entrypoints = "^0.4"
```
Secrus, line 53: if you are removing this, remove `types-entrypoints` from dev section too

src/poetry/console/commands/self/show/plugins.py
```diff
@@ -8,15 +8,32 @@
...
-    plugins: list[EntryPoint] = dataclasses.field(default_factory=list)
-    application_plugins: list[EntryPoint] = dataclasses.field(default_factory=list)
+    plugins: list[metadata.EntryPoint] = dataclasses.field(default_factory=list)
+    application_plugins: list[metadata.EntryPoint] = dataclasses.field(
+        default_factory=list
+    )
+
+    def append(self, entry_point: metadata.EntryPoint) -> None:
+        from poetry.plugins.application_plugin import ApplicationPlugin
+        from poetry.plugins.plugin import Plugin
```
Secrus, line 26: any particular reason for these imports to be here and not at the top of the file?

src/poetry/plugins/plugin_manager.py
```diff
@@ -41,11 +39,16 @@ def load_plugins(self, env: Env | None = None) -> None:
...
-        return entry_points
+    ) -> list[metadata.EntryPoint]:
+        return [
+            ep
+            for ep in metadata.entry_points(group=self._group)
+            if env is None
+            or (
+                ep.dist is not None
+                and env.site_packages.find_distribution(ep.dist.name) is not None
+            )
```
Secrus, line 50: I would rewrite this to a normal loop. in its current form, this is barely readable.

Secrus (changes requested):

src/poetry/console/commands/self/show/plugins.py
abn: it is mostly done to defer the import so the load times are quicker; for this change I am simply maintaining the status quo. Any changes to import ordering should be dealt with outside of this PR i reckon.

pyproject.toml
abn: Good catch.

src/poetry/plugins/plugin_manager.py
abn: Split the conditional into another method for now.

Secrus (approved):

Issue #5761, opened 2022-06-03
Confusing error reporting for faulty git dependencies


harahu:
- [X] I am on the [latest](https://github.com/python-poetry/poetry/releases/latest) Poetry version.
- [X] I have searched the [issues](https://github.com/python-poetry/poetry/issues) of this repo and believe that this is not a duplicate.
- [X] If an exception occurs when executing a command, I executed it again in debug mode (`-vvv` option).

- **OS version and name**: Ubuntu 21.10
- **Poetry version**: 1.1.13

## Issue
I tried to add `pendulum` as a git dependency, but I am pointing to a rev where `pendulum`s `pyproject.toml` file is in an invalid state:
```toml
[tool.poetry.dependencies]
pendulum = {git = "https://github.com/sdispater/pendulum.git", rev = "7e3880abbc51387f75644754f6f437dc515ed7e3"}
```

Running `poetry update` then leads to the following error:

```
Updating dependencies
Resolving dependencies... (1.3s)

  RuntimeError

  The Poetry configuration is invalid:
    - Additional properties are not allowed ('group' was unexpected)
  

  at ~/.poetry/lib/poetry/_vendor/py3.8/poetry/core/factory.py:43 in create_poetry
       39│             message = ""
       40│             for error in check_result["errors"]:
       41│                 message += "  - {}\n".format(error)
       42│ 
    →  43│             raise RuntimeError("The Poetry configuration is invalid:\n" + message)
       44│ 
       45│         # Load package
       46│         name = local_config["name"]
       47│         version = local_config["version"]

```

Upon seeing this error, my initial reaction was to assume that I had messed up _my own_ poetry configuration, but this isn't the case. This error originates from the `pendulum` resolution process.

From a user perspective, it would be helpful if I was told this in the error message, so that I know where to look for issues. I think it could suffice to just prepend the error message above with:
```
Error when resolving dependency `pendulum`:
```

harahu:
This one might be closable now? Haven't checked.

neersighted:
The specific error is resolved.

PR #5834, opened 2022-06-11
feat: fallback to gather metadata via pep517 if reading as Poetry project raises RuntimeError

finswimmer:
At the moment Poetry gives up gather metadata from a Poetry project during installation/locking if parsing the `pyproject.toml` raises a `RuntimeError`. This currently happens if one tries to add a Poetry project whose `pyproject.toml` contains group-dependencies introduced in Poetry 1.2, but the Poetry in use is <1.2. Similar scenarios are possible in the future whenever something changed in the appearance of the `pyproject.toml`.

With this PR we catch the `RuntimeError` and fallback to pep517 to gather the metadata, making the process more robust.

Once accepted I would suggest to backport it to Poetry 1.1 and cut a new release immediately, hoping to make the transition period between 1.1 and 1.2 more smooth, by avoiding things like https://github.com/python-poetry/poetry/issues/5761.

### DIFF
--- a/tests/inspection/test_info.py
+++ b/tests/inspection/test_info.py
@@ -131,6 +131,24 @@ def test_info_from_poetry_directory():
     demo_check_info(info)
 
 
+def test_info_from_poetry_directory_fallback_on_poetry_create_error(
+    mocker: MockerFixture,
+):
+    mock_create_poetry = mocker.patch(
+        "poetry.inspection.info.Factory.create_poetry", side_effect=RuntimeError
+    )
+    mock_get_poetry_package = mocker.spy(PackageInfo, "_get_poetry_package")
+    mock_get_pep517_metadata = mocker.patch(
+        "poetry.inspection.info.get_pep517_metadata"
+    )
+
+    PackageInfo.from_directory(FIXTURE_DIR_INSPECTIONS / "demo_poetry_package")
+
+    assert mock_create_poetry.call_count == 1
+    assert mock_get_poetry_package.call_count == 1
+    assert mock_get_pep517_metadata.call_count == 1
+
+
 def test_info_from_requires_txt():
     info = PackageInfo.from_metadata(
         FIXTURE_DIR_INSPECTIONS / "demo_only_requires_txt.egg-info"
--- a/tests/installation/test_executor.py
+++ b/tests/installation/test_executor.py
@@ -704,3 +704,59 @@ def test_executor_should_be_initialized_with_correct_workers(
     executor = Executor(tmp_venv, pool, config, io)
 
     assert executor._max_workers == expected_workers
+
+
+def test_executer_fallback_on_poetry_create_error(
+    mocker: MockerFixture,
+    config: Config,
+    pool: Pool,
+    io: BufferedIO,
+    tmp_dir: str,
+    mock_file_downloads: None,
+    env: MockEnv,
+):
+    mock_pip_install = mocker.patch("poetry.installation.executor.pip_install")
+    mock_sdist_builder = mocker.patch("poetry.core.masonry.builders.sdist.SdistBuilder")
+    mock_editable_builder = mocker.patch(
+        "poetry.masonry.builders.editable.EditableBuilder"
+    )
+    mock_create_poetry = mocker.patch(
+        "poetry.factory.Factory.create_poetry", side_effect=RuntimeError
+    )
+
+    config.merge({"cache-dir": tmp_dir})
+
+    executor = Executor(env, pool, config, io)
+
+    directory_package = Package(
+        "simple-project",
+        "1.2.3",
+        source_type="directory",
+        source_url=Path(__file__)
+        .parent.parent.joinpath("fixtures/simple_project")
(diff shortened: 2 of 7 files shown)


tests/inspection/test_info.py
```diff
@@ -131,6 +131,27 @@ def test_info_from_poetry_directory():
...
+def test_info_from_poetry_directory_fallback_on_poetry_create_error(
+    mocker: MockerFixture,
+):
+    import poetry.inspection.info
+
+    mocker.patch(
+        "poetry.inspection.info.Factory.create_poetry", side_effect=RuntimeError
+    )
+    mock_get_poetry_package = mocker.spy(
+        poetry.inspection.info.PackageInfo, "_get_poetry_package"
```
radoering, line 143:
```suggestion
        PackageInfo, "_get_poetry_package"
```

Can't we just use `PackageInfo` so that the local import is not required?

tests/inspection/test_info.py
```diff
@@ -131,6 +131,27 @@ def test_info_from_poetry_directory():
     demo_check_info(info)
 
 
+def test_info_from_poetry_directory_fallback_on_poetry_create_error(
+    mocker: MockerFixture,
+):
+    import poetry.inspection.info
+
+    mocker.patch(
```
radoering, line 139:
```suggestion
    mock_create_poetry = mocker.patch(
```
We should check that `create_poetry` is called once later. Otherwise the test will still succeed if the `pyproject.toml` is not found because `is_poetry_project()` returns False in that case. The same applies to the tests in `test_executor` and `test_pip_installer`.

radoering (changes requested):

dimbleby:
`RuntimeError` is a rather broad exception, and the scope at which it is suppressed is rather large in two of the three changed files.  This risks masking errors other than the one you are aiming at.

- too late to help if this does get back-applied to poetry 1.1: but can poetry-core be taught to throw a more specific exception on failing to parse the `pyproject.toml`?
- can the scope of the exception handling be narrowed here?

radoering:
> * too late to help if this does get back-applied to poetry 1.1: but can poetry-core be taught to throw a more specific exception on failing to parse the `pyproject.toml`?

I agree we should do this at least for future versions.

> * can the scope of the exception handling be narrowed here?

Had the same thought and then forgot to mention it. It probably can since we are only expecting an exception from `Factory.create_poetry()`. At least that's my understanding.

dimbleby:
re throwing a more precise exception: there's an existing `PyProjectException` which would probably be fine both [here](https://github.com/python-poetry/poetry-core/blob/a43b3b5027d33cccc58cc981192ae1e4d8b4ab42/src/poetry/core/factory.py#L56) and [here](https://github.com/python-poetry/poetry-core/blob/a43b3b5027d33cccc58cc981192ae1e4d8b4ab42/src/poetry/core/factory.py#L448), these should be the two that are relevant I think.

finswimmer:
> * can the scope of the exception handling be narrowed here?

Probably it can, but it looks a bit ugly:

```py
try:
    package_poetry = Factory().create_poetry(pyproject.file.path.parent)
except RuntimeError:
    pass
else:
    builder: Builder
    if package.develop and not package_poetry.package.build_script:
        from poetry.masonry.builders.editable import EditableBuilder
        ...
```

Is it worth it?  :thinking:

dimbleby:
```python
package_poetry = None
if pyproject.is_poetry_project():
    with contextlib.suppress(RuntimeError):
        package_poetry = Factory().create_poetry(pyproject.file.path.parent)

if package_poetry is not None:
    etc
```
?

finswimmer:
Much nicer :+1: 

I decided for:

```py
if pyproject.is_poetry_project():
    try:
        package_poetry = Factory().create_poetry(pyproject.file.path.parent)
    except RuntimeError:
        package_poetry = None

    if package_poetry is not None:
        ...
```

tests/inspection/test_info.py
finswimmer: Done

tests/inspection/test_info.py
finswimmer: Done

tests/installation/test_executor.py
```diff
@@ -648,3 +648,59 @@ def test_executor_should_be_initialized_with_correct_workers(
...
+    tmp_dir: str,
+    mock_file_downloads: None,
+    env: MockEnv,
+):
+    mock_pip_install = mocker.patch("poetry.installation.executor.pip_install")
+    mock_sdist_builder = mocker.patch("poetry.core.masonry.builders.sdist.SdistBuilder")
+    mock_editable_builder = mocker.patch(
+        "poetry.masonry.builders.editable.EditableBuilder"
+    )
+    mock_creat_poetry = mocker.patch(
```
radoering, line 667:
```suggestion
    mock_create_poetry = mocker.patch(
```

finswimmer: oops ...

radoering (approved):

PR #5840, opened 2022-06-13
refactor: use contextlib.suppress instead of except pass

kkirsche:
# Pull Request Check List

Resolves: N/A

This merge request converts `try` / `except` blocks with no `except` action to use `contextlib.suppress` to silence a specific error, instead of passing in an exception handler.

#### Example Before

```python
try:
    travel_world(days=80)
except DistractionError:
    pass
```

#### Example After

```python
import contextlib

with contextlib.suppress(DistractionError):
    travel_world(days=80)
```

#### Explanation
The context manager slightly shortens the code and significantly clarifies the author's intention to ignore the specific errors. The standard library feature was introduced following a discussion, where the consensus was that

> A key benefit here is in the priming effect for readers... The with statement form makes it clear before you start reading the code that certain exceptions won't propagate.

This ignores a few locations that use `try` / `except` / `else`

### DIFF
--- a/get-poetry.py
+++ b/get-poetry.py
@@ -30,7 +30,7 @@ import tarfile
 import tempfile
 
 from contextlib import closing
-from contextlib import contextmanager
+from contextlib import contextmanager, suppress
 from functools import cmp_to_key
 from gzip import GzipFile
 from io import UnsupportedOperation
@@ -46,10 +46,8 @@ except ImportError:
     from urllib2 import Request
     from urllib2 import urlopen
 
-try:
+with suppress(NameError):
     input = raw_input
-except NameError:
-    pass
 
 
 try:
--- a/get-poetry.py
+++ b/get-poetry.py
@@ -30,7 +30,7 @@ import tarfile
 import tempfile
 
 from contextlib import closing
-from contextlib import contextmanager, suppress
+from contextlib import contextmanager
 from functools import cmp_to_key
 from gzip import GzipFile
 from io import UnsupportedOperation
@@ -46,8 +46,10 @@ except ImportError:
     from urllib2 import Request
     from urllib2 import urlopen
 
-with suppress(NameError):
+try:
     input = raw_input
+except NameError:
+    pass
 
 
 try:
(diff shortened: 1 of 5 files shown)


get-poetry.py
```diff
@@ -30,7 +30,7 @@
 import tempfile
 
 from contextlib import closing
-from contextlib import contextmanager
+from contextlib import contextmanager, suppress
```
branchvincent, line 33: this install script is deprecated, but still technically supports python2. let's revert this change, but the rest look good

branchvincent (changes requested):

get-poetry.py
neersighted: To clarify further, we're considering it frozen and not subject to modification outside critical defects like a security issue. Even code formatting is not being applied.

neersighted (changes requested):
I'd rather import all of contextlib and invoke the context manager as `contextlib.suppress` unless you see a compelling reason not to -- it's more explicit given that `suppress` is not a common short import name.

kkirsche:
Including feedback from @neersighted in 32c34884a63b3fd2ca3e90f867b6b364ca33a4fc
. If there is a reason to revert this, we can use git revert accordingly 👍  hopefully this will help ensure that your time is respected reviewing this.

branchvincent (approved):
@kkirsche thank you!

Callek:
@branchvincent @kkirsche - IMO, Probably shouldn't have touched the pyyaml and ansible `setup.py` files in the test folder I don't think.

kkirsche:
Why, @Callek?

kkirsche:
I'll submit a similar PR upstream to normalize them, if applicable to their supported versions

Callek:
> Why, @Callek?
> 
> EDIT: Oh, it looks like you're vendoring them. Didn't realize that, sorry. I thought they were just local forks

Yea, because they are vendored -- but I should be super explicit, I'm not a maintainer on poetry nor on the two projects I mentioned there. So my comments are merely as a passerby.

neersighted:
> @branchvincent @kkirsche - IMO, Probably shouldn't have touched the pyyaml and ansible `setup.py` files in the test folder I don't think.

Good catch, thanks!

Issue #5849, opened 2022-06-15
poetry.lock at ~/.config/pypoetry incorrect when installing local plugin via wheel due to symbolic links


spoorn:
- [ x ] I am on the [latest](https://github.com/python-poetry/poetry/releases/latest) Poetry version.
- [ x ] I have searched the [issues](https://github.com/python-poetry/poetry/issues) of this repo and believe that this is not a duplicate.
- [ x ] If an exception occurs when executing a command, I executed it again in debug mode (`-vvv` option).

- **OS version and name**: CentOS 8/Amazon Linux 2
- **Poetry version**: 1.2.0b2
- **Link of a [Gist](https://gist.github.com/) with the contents of your pyproject.toml file**: 

## Issue

While investigating https://github.com/python-poetry/poetry/issues/5440, I found out that if I tried to install a plugin using a path that contained a symlink e.g. 

`poetry self add /mysymlink/workspace/myuser/myproject/dist/myproject-1.0.0-py3-none-any.whl`

where __mysymlink__ resolves to say `home/myuser`, poetry would fail to install the plugin saying FileNotFound.

If I had instead used the [real physical path](https://unix.stackexchange.com/questions/63580/resolving-symbolic-links-pwd) using `pwd -P`, so I'd do 

`poetry self add /home/myuser/workspace/myuser/myproject/dist/myproject-1.0.0-py3-none-any.whl`

it works just fine.

The problem is, if I originally used the symlink which caused the plugin install via a local wheel file to fail, the second option using the real physical path would also fail afterwards.  Even if I had used the real physical path initially, if I remake a new virtualenv and try again, it also fails.  After some digging, I found out when poetry tries to get the locked_repository here: https://github.com/python-poetry/poetry/blob/master/src/poetry/packages/locker.py#L454, it's reading from poetry.lock/pyproject.toml in `~/.config/pypoetry`.

The `~/.config/pypoetry/poetry.lock` has an incorrect __relative__ path to the local wheel file:

```
...
[package.source]
type = "file"
url = "../../../../home/myuser/workspace/myuser/myproject/dist/myproject-1.0.0-py3-none-any.whl"
...
```

If I try to reinstall the plugin, it says the file is not found:

```
...
File /local/home/myuser/workspace/myuser/myproject/dist/myproject-1.0.0-py3-none-any.whl does not exist
```

It looks like it is missing an extra "../".

[...]

PR #5850, opened 2022-06-15
Resolve symlinks in locker.lock path to real physical paths

spoorn:
Fixes https://github.com/python-poetry/poetry/issues/5849

# Pull Request Check List

Resolves: #5849

### DIFF
--- a/tests/packages/test_locker.py
+++ b/tests/packages/test_locker.py
@@ -2,6 +2,8 @@ from __future__ import annotations
 
 import json
 import logging
+import os
+import sys
 import tempfile
 import uuid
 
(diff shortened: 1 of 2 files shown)


spoorn:
Having some trouble getting the new test to pass on Windows.

Windows Python < 3.8 does not resolve symlinks at all using `os.path.realpath()`: https://docs.python.org/3/library/os.path.html#os.path.realpath

The suggestion is instead to use `os.readlink()`, but that does not give the correct path for Python >= 3.8 as it resolves to the substitute path which contains a prefix `\\\\?\\`, instead of the display path, due to https://bugs.python.org/issue42957

Few options we can take:
1. Use `os.readlink()` and remove prefix "\\\\?\\"
2. Check for python version.  If < 3.8, use `os.readlink()`, else use `os.path.realpath()`
3. Disable the test for Python < 3.8, so only Windows 3.7 will not resolve symlinks

spoorn:
Could I get a review for this?

tests/packages/test_locker.py
```diff
@@ -753,3 +754,77 @@ def test_content_hash_with_legacy_is_compatible(
...
+        symlink_path = Path(d1).joinpath("testsymlink")
+        with tempfile.TemporaryDirectory(dir=d1) as d2, tempfile.TemporaryDirectory(
+            dir=d1
+        ) as d4, tempfile.TemporaryDirectory(dir=d2) as d3, tempfile.NamedTemporaryFile(
+            dir=d4
+        ) as source_file, tempfile.NamedTemporaryFile(
+            dir=d3
+        ) as lock_file:
+            lock_file.close()
+            os.symlink(Path(d3), symlink_path)
```
radoering, line 785:
```suggestion
            try:
                os.symlink(Path(d3), symlink_path)
            except OSError:
                pass
                if sys.platform == "win32":
                    # os.symlink requires either administrative privileges or developer
                    # mode on Win10, throwing an OSError if neither is active.
                    # Test is not possible in that case.
                    return
                raise
```

`os.symlink` requires os.symlink requires either administrative privileges or developer mode on Win10. We have a workaround to avoid symlinks in https://github.com/python-poetry/poetry/blob/83e6c318f59a77e83553c9805562490a4e5f8811/tests/helpers.py#L75-L92

However, I don't think it makes sense to use this function here. Thus, maybe it's best to "skip" the test for windows when `os.symlink` fails.

tests/packages/test_locker.py
```diff
@@ -753,3 +754,77 @@ def test_content_hash_with_legacy_is_compatible(
...
+version = "1.2.3"
+description = ""
+category = "main"
+optional = false
+python-versions = "*"
+
+[package.source]
+type = "file"
+url = "{Path(os.path.relpath(Path(source_file.name).resolve().as_posix(),
+         Path(Path(lock_file.name).parent).resolve().as_posix())).as_posix()}"
```
radoering, line 817:
```suggestion
url = "{
    Path(
        os.path.relpath(
            Path(source_file.name).resolve().as_posix(),
            Path(Path(lock_file.name).parent).resolve().as_posix(),
        )
    ).as_posix()
}"
```
formatting for readability

radoering (changes requested):
Just some minor nitpicks in the test.

tests/packages/test_locker.py
spoorn: Done

tests/packages/test_locker.py
spoorn: Done

tests/packages/test_locker.py
```diff
@@ -753,3 +755,92 @@ def test_content_hash_with_legacy_is_compatible(
...
+        ) as d4, tempfile.TemporaryDirectory(dir=d2) as d3, tempfile.NamedTemporaryFile(
+            dir=d4
+        ) as source_file, tempfile.NamedTemporaryFile(
+            dir=d3
+        ) as lock_file:
+            lock_file.close()
+            try:
+                os.symlink(Path(d3), symlink_path)
+            except OSError:
+                pass
```
radoering, line 789:
```suggestion
```

my fault

spoorn: Done

radoering (approved):

Issue #5720, opened 2022-05-30
Dependency Audit


danieleades:
We should audit Poetry's dependencies to ensure they are rock solid.

Some of the dependencies could do with a refresh-

| Dependency | maintained?/last updated | notes |
|------------------|-------------------------------------|--------|
| ~~cachy~~ | ~~7/8/19~~ | ~~some tidying required. No type annotations. Deps out of date~~ |
| ~~entrypoints~~ | ~~unmaintained~~ | ~~this dep should be replaced with `importlib-metadata`/`importlib.metadata`~~ |
| importlib-metadata | actively maintained | required for python 3.7 *only*. Should be removed when 3.7 is end-of-life |

there may be others worth looking at

onerandomusername:
Is it possible for poetry to vendor some of its dependencies? I don't think poetry exactly needs 50 dependencies installed, as sometimes those dependencies break

danieleades:
> Is it possible for poetry to vendor some of its dependencies? I don't think poetry exactly needs 50 dependencies installed, as sometimes those dependencies break

i count 39 non-dev transitive dependencies.

I don't think that's necessarily an issue, since they are managed and isolated. I'm more interested in ensuring the foundation packages used by Poetry are solid, then in changing the way those packages are distributed. Are there specific dependencies that you think should be vendored? if so, why?

danieleades:
> @danieleades I agree with you on `entrypoints`. When it comes to `importlib-metadata`, it will be kept until we drop python 3.7 and that will be 1yr+ as that's how long will 3.7 be officially supported by PSF.
> 
> When it comes to `cachy`, you can try finding some alternative and prepare PR for migrating it. However, as `cachy` is made by sdispater, I am quite sure it's tailored for Poetry and it might be hard to find good alternative.
> 
> Long story short, PRs are welcome :D

I've got a few open PRs to refresh `cachy`, though I don't know how active the maintenance is.
Agree that we can't touch `importlib-metadata` for a while, listing here for awareness.
Had a quick look at removing `entrypoint`, but it's non-trivial. Hopefully there'll be some time in my future where I can have a proper crack at it, but it won't be in the next few weeks.

Secrus:
> > @danieleades I agree with you on `entrypoints`. When it comes to `importlib-metadata`, it will be kept until we drop python 3.7 and that will be 1yr+ as that's how long will 3.7 be officially supported by PSF.
[...]

PR #5868, opened 2022-06-18
poc: replace cachy with cache module

chadac:
Related to the discussion in #5720.

This is a test to see what would be needed to pull out of cachy to get it into Poetry entirely. The result is a new utils module, `poetry.utils.cache`, that is about 250 LOC. I also went ahead and made a couple of small improvements:

* Broke up cachy into simpler classes, mainly to push stuff out of config and into typed variables. Now caches are declared separately -- `FileCache`, `DictCache`
* Add types to methods and generics
* Update the file format. Before it was the expiration time concatenated with the payload, now it's purely JSON encoded. Added compatibility tests and checks to ensure the migration won't break existing caches.

# Pull Request Check List

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -51,7 +51,6 @@ poetry-core = "^1.3.2"
 poetry-plugin-export = "^1.1.2"
 "backports.cached-property" = { version = "^1.0.2", python = "<3.8" }
 cachecontrol = { version = "^0.12.9", extras = ["filecache"] }
-cachy = "^0.3.0"
 cleo = "^1.0.0a5"
 crashtest = "^0.3.0"
 dulwich = "^0.20.46"
@@ -80,6 +79,8 @@ urllib3 = "^1.26.0"
 pre-commit = "^2.6"
 
 [tool.poetry.group.test.dependencies]
+# Cachy frozen to test backwards compatibility for `poetry.utils.cache`.
+cachy = "0.3.0"
 deepdiff = "^5.0"
 flatdict = "^4.0.1"
 httpretty = "^1.0"
--- a/src/poetry/repositories/legacy_repository.py
+++ b/src/poetry/repositories/legacy_repository.py
@@ -90,23 +90,19 @@ class LegacyRepository(HTTPRepository):
         if not constraint.is_any():
             key = f"{key}:{constraint!s}"
 
-        if self._cache.store("matches").has(key):
-            versions = self._cache.store("matches").get(key)
-        else:
-            page = self._get_page(f"/{name}/")
-            if page is None:
-                self._log(
-                    f"No packages found for {name}",
-                    level="debug",
-                )
-                return []
+        page = self._get_page(f"/{name}/")
+        if page is None:
+            self._log(
+                f"No packages found for {name}",
+                level="debug",
+            )
+            return []
 
-            versions = [
-                (version, page.yanked(name, version))
-                for version in page.versions(name)
-                if constraint.allows(version)
-            ]
-            self._cache.store("matches").put(key, versions, 5)
+        versions = [
+            (version, page.yanked(name, version))
+            for version in page.versions(name)
+            if constraint.allows(version)
+        ]
 
         return [
             Package(
--- a/src/poetry/repositories/pypi_repository.py
+++ b/src/poetry/repositories/pypi_repository.py
@@ -100,13 +100,7 @@ class PyPiRepository(HTTPRepository):
         The information is returned from the cache if it exists
         or retrieved from the remote server.
         """
-        if self._disable_cache:
-            return self._get_package_info(name)
-
-        package_info: dict[str, Any] = self._cache.store("packages").remember_forever(
-            name, lambda: self._get_package_info(name)
-        )
-        return package_info
+        return self._get_package_info(name)
 
     def _find_packages(
         self, name: NormalizedName, constraint: VersionConstraint
@@ -129,15 +123,11 @@ class PyPiRepository(HTTPRepository):
         if not constraint.is_any():
             key = f"{key}:{constraint!s}"
 
-        if self._cache.store("matches").has(key):
-            versions = self._cache.store("matches").get(key)
-        else:
-            versions = [
-                (version, json_page.yanked(name, version))
-                for version in json_page.versions(name)
-                if constraint.allows(version)
-            ]
-            self._cache.store("matches").put(key, versions, 5)
+        versions = [
+            (version, json_page.yanked(name, version))
+            for version in json_page.versions(name)
+            if constraint.allows(version)
+        ]
 
         pretty_name = json_page.content["name"]
         packages = [
--- a/src/poetry/utils/cache.py
+++ b/src/poetry/utils/cache.py
@@ -1,0 +1,192 @@
+from __future__ import annotations
+
+import contextlib
+import dataclasses
+import hashlib
+import json
+import shutil
+import time
+
+from pathlib import Path
+from typing import Any
+from typing import Callable
+from typing import Generic
+from typing import TypeVar
+
+
+# Used by Cachy for items that do not expire.
+MAX_DATE = 9999999999
+T = TypeVar("T")
+
+
+def decode(string: bytes, encodings: list[str] | None = None) -> str:
+    """
+    Compatiblity decode function pulled from cachy.
+
+    :param string: The byte string to decode.
+    :param encodings: List of encodings to apply
+    :return: Decoded string
+    """
+    if encodings is None:
+        encodings = ["utf-8", "latin1", "ascii"]
+
+    for encoding in encodings:
+        with contextlib.suppress(UnicodeDecodeError):
+            return string.decode(encoding)
+
+    return string.decode(encodings[0], errors="ignore")
+
+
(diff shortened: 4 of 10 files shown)


src/poetry/utils/cache.py
```diff
@@ -0,0 +1,250 @@
...
+def _expiration(minutes: int) -> int:
+    """
+    Calculates the time in seconds since epoch that occurs 'minutes' from now.
+
+    :param minutes: The number of minutes to count forward
+    """
+    return round(time.time()) + minutes * 60
+
+
+@dataclass
```
danieleades, line 66: could you use an `abc.ABC` here?

pyproject.toml
```diff
@@ -78,6 +77,7 @@ pytest-randomly = "^3.10"
 pytest-sugar = "^0.9"
 pytest-xdist = { version = "^2.5", extras = ["psutil"] }
 pre-commit = "^2.6"
+cachy = "^0.3.0"
```
danieleades, line 80: is this intentional?

src/poetry/utils/cache.py
chadac: Sure, added `abc.ABC` and namespaced `dataclasses.dataclass` for the same convention.

src/poetry/repositories/legacy_repository.py
```diff
@@ -45,14 +47,13 @@ def find_packages(self, dependency: Dependency) -> list[Package]:
 
         ignored_pre_release_versions = []
 
-        if self._cache.store("matches").has(key):
-            versions = self._cache.store("matches").get(key)
-        else:
+        versions = self._match_cache.get(key)
```
abn, line 50: Doesn't this move the cache from a file cache into an in-memory cache? That would be undesired.

abn: Also, I am not sure I like having the cache interface split out into specific caches; the way it is done today all caches related to the repository is stored under the same top-level dir iirc. We will want to retain this.

src/poetry/repositories/pypi_repository.py
```diff
@@ -42,6 +43,7 @@ def __init__(
 
         self._base_url = url
         self._fallback = fallback
+        self._package_cache: DictCache[dict[str, Any]] = DictCache()
```
abn, line 46: Same as above (`self._match_cache`), spliting this out into individual child caches might be undesired.

src/poetry/repositories/pypi_repository.py
```diff
@@ -152,7 +154,7 @@ def get_package_info(self, name: str) -> dict[str, Any]:
         if self._disable_cache:
             return self._get_package_info(name)
 
-        package_info: dict[str, Any] = self._cache.store("packages").remember_forever(
+        package_info: dict[str, Any] = self._package_cache.remember(
```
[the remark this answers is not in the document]
abn: This use is bot read (when cache exists) and write (when it does not exist).

src/poetry/utils/cache.py
```diff
@@ -0,0 +1,250 @@
+from __future__ import annotations
```
abn, line 1: Let's move this module into a top-level package `poetry.cache`. Ideally with relevant bits split out into its own modules.

src/poetry/utils/cache.py
```diff
@@ -0,0 +1,250 @@
...
+
+    def has(self, key: str) -> bool:
+        """
+        Determine if a file exists and has not expired in the cache.
+
+        :param key: The cache key
+
+        :returns: True if the key exists in the cache
+        """
+        return self.get(key) is not None
```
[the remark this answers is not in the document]
abn: Unsure if file stat is sufficient as file timestamps are not reliable across filesystems and/or cases when cache is restored under different environments (eg: CI).

abn (changes requested):

abn:
Have not gone through the cache logic itself, but left some high level comments regarding the change.

src/poetry/utils/cache.py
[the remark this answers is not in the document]
chadac: Right now the actual filesystem-based cache only saves entries indefinitely, so no timestamp checking is needed either. At this moment, it would be functionally equivalent to check the filesystem for the existence of this file.

This can be confirmed by checking caches directly as well:

    find ~/.cache/pypoetry/cache/repositories/pypi -not -type d -not -path "*/_http/*" -exec sh -c "echo -n '{} '; head -c 10 {}; echo" \;

The header of each file is the unix timestamp when the cache expires -- which is 9999999999 everywhere.

src/poetry/repositories/legacy_repository.py
chadac: Currently the match and package caches are actually both in-memory, see [this line.](https://github.com/python-poetry/poetry/blob/master/src/poetry/repositories/cached.py#L36). The only in-file cache is for release metadata.

I could definitely combine these all back together, but I was wondering if, since _match_cache and _package_cache are already in-memory and only used in these subclasses, perhaps they should stay split? Unless there are plans for the cache to make use of all three stores in the future.

dimbleby: I'm pretty sure that the `packages` cache is more or less worthless: the request to the server has `cachecontrol` taking care of caching already, at best the extra layer is saving us re-parsing the JSON in the response.

The 'matches' cache is similar, though perhaps it saves _slightly_ more work.

git history shows that `cachy` was introduced first, and `cachecontrol` followed later.  The caching provided by `cachecontrol` makes these two in-memory caches near-enough redundant I reckon.

Still, I'd be tempted to get rid of them both altogether for the simplification - unless testing shows that they really make a meaningful difference.

chadac: I could definitely go ahead and do a profile to compare these two. This would really help reduce the size of the caching module overall, as we could then compress most of this logic into a single class.

Secrus:
@chadac any progress on this?

chadac:
@Secrus Apologies, got pulled into some busy separate stuff that has occupied my time. I'll commit to get some of those profiling stats up for next week so to help reach a decision on one of those threads above (for `DictCache`).

chadac:
Posted an update which cuts down a lot of the junk. I did do a test on the proposed `match` and `package` caches referenced above. For `PyPiRepository` I found relatively little difference, so I focused testing on `LegacyRepository` with a sample package of about 10 packages. With caching, average install time took about 55.2s +/- 1.5s, while without caching it took around 56.1s +/- 2.7s. This was for a sample size of 10, each run inside a Docker container under the same conditions. Since this is well within the margin of error I believe it's not a significant difference, but I'm increasing the sample size now to increase confidence. I'll get a proper hypothesis test once ready to officially confirm.

src/poetry/utils/cache.py
```diff
@@ -0,0 +1,171 @@
...
+
+    @property
+    def expired(self) -> bool:
+        """
+        Return true if the cache item has exceeded its expiration period.
+        """
+        return self.expires is not None and time.time() >= self.expires
+
+
+@dataclasses.dataclass
```
neersighted, line 92: I don't see a good reason to not make this `frozen=True` as well -- changing the Path or hash of a cache object during runtime seems unnecessary.

tests/console/commands/cache/test_clear.py
```diff
@@ -33,8 +33,10 @@ def test_cache_clear_all(
 
     assert exit_code == 0
     assert tester.io.fetch_output() == ""
-    # ensure directory is empty
-    assert not any((repository_cache_dir / repository_one).iterdir())
+    # ensure directory is empty or doesn't exist
+    assert not (repository_cache_dir / repository_one).exists() or not any(
```
neersighted, line 37: Store this in a variable instead of joining the path twice.

tests/utils/test_cache.py
```diff
@@ -0,0 +1,160 @@
...
+            "stores": {
+                "cache": {"driver": "file", "path": str(repository_cache_dir / "cache")}
+            },
+        }
+    )
+    return patch_cachy(cache)
+
+
+@pytest.fixture
+def file_cache(repository_cache_dir: Path) -> FileCache:
```
neersighted, line 69: Let's call this `poetry_file_cache` or `builtin_file_cache` for clarity.

neersighted (changes requested):

tests/utils/test_cache.py
chadac: Changed it to `poetry_file_cache`, added a `remember` test as well.

src/poetry/utils/cache.py
chadac: Added, the more immutable stuff there is, the better :)

chadac:
I did end up running a more complete performance comparison for the package & match cache. The results mostly agreed with what was shown earlier; when comparing 50 identical runs between the two, the improvement in performance was about 0.5%, which was well within the margin of error (difference in average: 0.3s, with standard deviation of +/- 3.2s). The student t test reports a p-value of around 0.6, indicating that there was very little certainty that this adds any improvement at all.

chadac:
Sounds good, lemme refactor the commits.

neersighted (approved):

Issue #2415, opened 2020-05-13
Poetry downloading same wheels multiple times within a single invocation


bb:
- [x] I am on the [latest](https://github.com/python-poetry/poetry/releases/latest) Poetry version.
- [x] I have searched the [issues](https://github.com/python-poetry/poetry/issues) of this repo and believe that this is not a duplicate.
- [ ] If an exception occurs when executing a command, I executed it again in debug mode (`-vvv` option).

- **OS version and name**: macOS 10.14.6
- **Poetry version**: 1.0.5
- **Link of a [Gist](https://gist.github.com/) with the contents of your pyproject.toml file**: https://gist.github.com/bb/501f33ad3f35eb8c26ce2513ca6074c8

## Issue

When adding a new dependency, it is downloaded multiple times; I observed three downloads, two of those are unneccessary.

Starting with a `pyproject.toml` as in the Gist given above, I run

    poetry add https://github.com/oroszgy/spacy-hungarian-models/releases/download/hu_core_ud_lg-0.3.1/hu_core_ud_lg-0.3.1-py3-none-any.whl

Then I see the following output (XXX added as markers for explanation below):

    Updating dependencies XXX
    Resolving dependencies... (276.1s)

    Writing lock file
    XXX

    Package operations: 0 installs, 7 updates, 0 removals

      - Updating certifi (2019.11.28 -> 2020.4.5.1)
      - Updating urllib3 (1.25.8 -> 1.25.9)
      - Updating asgiref (3.2.3 -> 3.2.7)
      - Updating pytz (2019.3 -> 2020.1)
      - Updating django (3.0.4 -> 3.0.6)
      - Updating hu-core-ud-lg (0.3.1 -> 0.3.1 https://github.com/oroszgy/spacy-hungarian-models/releases/download/hu_core_ud_lg-0.3.1/hu_core_ud_lg-0.3.1-py3-none-any.whl)
    XXX  - Updating psycopg2-binary (2.8.4 -> 2.8.5)

At the positions where the marker `XXX` is inserted, the same 1.3GB download is done again and again.

Similar, when adding another package later, again `XXX` marks the cursor position when the big download is done:

    $ poetry add djangorestframework

    Using version ^3.11.0 for djangorestframework

    Updating dependencies
    Resolving dependencies... (0.4s)

    Writing lock file
    XXX

    Package operations: 1 install, 1 update, 0 removals

      - Installing djangorestframework (3.11.0)
[...]

PR #5871, opened 2022-06-18
Use locally cached wheels during install

dimbleby:
Related to #2415, though it only addresses one of the many repeated downloads which that issue reports.

If we have a url dependency pointing at a wheel eg
```
hu-core-ud-lg = {url = "https://github.com/oroszgy/spacy-hungarian-models/releases/download/hu_core_ud_lg-0.3.1/hu_core_ud_lg-0.3.1-py3-none-any.whl"}
```
then poetry performs the download during `poetry install`, even though it has a perfectly good cached version in `~/.cache/pypoetry/artifacts`.

`get_cached_archive_for_link` looks kinda confused: returning the original input `Link` to mean "no cached archive found" is odd.  So I have tweaked it to return `None` instead which seems a lot clearer.  

It's always possible that I am the confused one, certainly I am struggling to understand what the intention of the changed code can have been.  So do please think about that when reviewing!

### DIFF
--- a/src/poetry/installation/chef.py
+++ b/src/poetry/installation/chef.py
@@ -25,15 +25,10 @@ class Chef:
             Path(config.get("cache-dir")).expanduser().joinpath("artifacts")
         )
 
-    def get_cached_archive_for_link(self, link: Link) -> Link:
-        # If the archive is already a wheel, there is no need to cache it.
-        if link.is_wheel:
-            return link
-
+    def get_cached_archive_for_link(self, link: Link) -> Link | None:
         archives = self.get_cached_archives_for_link(link)
-
         if not archives:
-            return link
+            return None
 
         candidates: list[tuple[float | None, Link]] = []
         for archive in archives:
@@ -54,7 +49,7 @@ class Chef:
             )
 
         if not candidates:
-            return link
+            return None
 
         return min(candidates)[1]
 
--- a/src/poetry/installation/chef.py
+++ b/src/poetry/installation/chef.py
@@ -6,13 +6,12 @@ import json
 from pathlib import Path
 from typing import TYPE_CHECKING
 
-from poetry.core.packages.utils.link import Link
-
 from poetry.installation.chooser import InvalidWheelName
 from poetry.installation.chooser import Wheel
 
 
 if TYPE_CHECKING:
+    from poetry.core.packages.utils.link import Link
 
     from poetry.config.config import Config
     from poetry.utils.env import Env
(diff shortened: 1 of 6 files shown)


src/poetry/installation/chef.py
```diff
@@ -41,15 +41,10 @@ def should_prepare(self, archive: Path) -> bool:
     def is_wheel(self, archive: Path) -> bool:
         return archive.suffix == ".whl"
 
-    def get_cached_archive_for_link(self, link: Link) -> Link:
-        # If the archive is already a wheel, there is no need to cache it.
-        if link.is_wheel:
-            return link
-
+    def get_cached_archive_for_link(self, link: Link) -> Link | None:
```
radoering, line 44: Can we make `get_cached_archive_for_link` return a `Path` instead of a `Link` (same for `get_cached_archives_for_link`)? That should make things even clearer.

src/poetry/installation/chef.py
```diff
@@ -41,15 +41,10 @@ def should_prepare(self, archive: Path) -> bool:
     def is_wheel(self, archive: Path) -> bool:
         return archive.suffix == ".whl"
 
-    def get_cached_archive_for_link(self, link: Link) -> Link:
-        # If the archive is already a wheel, there is no need to cache it.
-        if link.is_wheel:
-            return link
```
radoering, line 47: Considering the comment, I can only imagine that might be a shortcut for local wheels? Maybe, you can examine if the change has some side effects on wheels as path dependencies. 🙏

dimbleby: The only place this is called is in the executor `_download_link()` code path, which means that it is never relevant for path dependencies.

I'm quite unsure about what this is supposed to be doing - `Chef` isn't a name that helps much.  At first I'd thought it was going to be some general purpose artifact cache, but its extremely limited scope is puzzling to me.

src/poetry/installation/chef.py
dimbleby: pushed a second commit that does this.  It does indeed tidy things up, but it also makes the MR quite a bit more substantial

src/poetry/installation/chef.py
abn: I need to look at this when I have more time. But I suspect the original intent was to cache the environment specific when built. Useful when the link is an sdist.

dimbleby: https://github.com/python-poetry/poetry/pull/5871#issuecomment-1168805110

radoering (approved):
Convincing enough for me. I see no reason to put this PR off any longer.

Issue #5116, opened 2022-01-27
Allow `poetry add` with a `--frozen` flag to freeze current/specified version of dependency  


aalok-sathe:
- [x] I have searched the [issues](https://github.com/python-poetry/poetry/issues) of this repo and believe that this is not a duplicate.
- [x] I have searched the [documentation](https://python-poetry.org/docs/) and believe that my question is not covered.

## Feature Request

It would be useful to be able to freeze the version of a dependency at the same time as `add`. Some dependencies can be volatile, and even with the expectation that a minor revision should not chage the interface of any dependency, to certain projects, this risk may not be tolerable. Whereas we can already manually edit the `pyproject.toml` file to achieve this, an option to the `add` command to by default add a specific version `3.1.4` rather than `^3.1.4` or no version at all, would be useful.

Being able to do this retroactively could also be useful, using, say, `poetry freeze numpy`, which will freeze the `numpy` dependency to its current version in the lock file.

clintonroy:
You can already do this.

poetry add pytest==6.2.4

for example.

PR #5874, opened 2022-06-18
Better document exact requirement specification

mkniewallner:
# Pull Request Check List

Resolves: #5116

Following https://github.com/python-poetry/poetry/issues/5116, this PR adds an example for exact requirements to both the dependency specification and `add` command documentations, in order to better reflect this possibility.

### DIFF
--- a/docs/cli.md
+++ b/docs/cli.md
@@ -288,18 +288,28 @@ poetry will choose a suitable one based on the available package versions.
 poetry add requests pendulum
 ```
 
-You also can specify a constraint when adding a package, like so:
+You can also specify a constraint when adding a package:
 
 ```bash
+# Allow >=2.0.5, <3.0.0 versions
 poetry add pendulum@^2.0.5
+
+# Allow >=2.0.5, <2.1.0 versions
+poetry add pendulum@~2.0.5
+
+# Allow >=2.0.5 versions, without upper bound
 poetry add "pendulum>=2.0.5"
-poetry add "pendulum==2.0.5"
+
+# Allow only 2.0.5 version
+poetry add pendulum==2.0.5
 ```
 
 If you try to add a package that is already present, you will get an error.
 However, if you specify a constraint, like above, the dependency will be updated
-by using the specified constraint. If you want to get the latest version of an already
-present dependency you can use the special `latest` constraint:
+by using the specified constraint.
+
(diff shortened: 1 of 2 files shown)


docs/dependency-specification.md
```diff
@@ -74,10 +74,13 @@ Here are some examples of inequality requirements:
...
 
 You can specify the exact version of a package.
+
+`==1.2.3` is an example of an exact version specification.
+
 This will tell Poetry to install this version and this version only.
 If other dependencies require a different version, the solver will ultimately fail and abort any install or update procedures.
 
-#### Multiple requirements
+### Multiple requirements
```
mkniewallner, line 83: Wait, I thought that it was incorrectly indented, because it does not relate at all to `Exact requirements` section, but re-reading the doc, the fact that it's an H4 may be correct, and it should just be moved under [Inequality requirements](https://python-poetry.org/docs/dependency-specification/#inequality-requirements) instead.

mkniewallner: Added https://github.com/python-poetry/poetry/pull/5874/commits/53374dfff4ac7ce05cc87ed9e9b3006069c232a7 to change that, though I would prefer to have confirmation that this is the correct thing to do.

docs/cli.md
```diff
@@ -293,6 +293,7 @@ You also can specify a constraint when adding a package, like so:
 ```bash
 poetry add pendulum@^2.0.5
 poetry add "pendulum>=2.0.5"
+poetry add "pendulum==2.0.5"
```
Secrus, line 296: Not sure if this belongs here. Those two existing entries are the same, but the one you added has different behavior. I would either move it to a separate block or remove it entirely.

Secrus (changes requested):

docs/cli.md
mkniewallner: They are not exactly the same, since `^2.0.5` will pick up `>=2.0.5,<3.0.0` while `>=2.0.5` will not have an upper bound and could pick up `3.0.0` for instance.

I've tried something in https://github.com/python-poetry/poetry/pull/5874/commits/51c975daae2f6b69c0e11b60f2ee1803d9586044 to better detail the possibilities, let me know what you think.

Secrus (approved):

PR #5878, opened 2022-06-19
Include strtobool in utils.extras to speed up the shell command

tddschn:
# Pull Request Check List

- [ ] Updated **documentation** for changed code. (The change doesn't reflect in the current doc)

# What does this PR do?

This line `from distutils.util import strtobool` in  `console.commands.shell` is slow and makes poetry shell command slow, since importing `distutils.util` also does a bunch of other things. 

`strtobool` is a very simple function (~10 lines), if poetry includes this function itself, for example, putting it in utils.extras.py and use from poetry.utils.extras import strtobool, poetry shell would run faster.

[Discord discussion link](https://discord.com/channels/487711540787675139/974839878669987840/988024065933594704)

### DIFF
--- a/src/poetry/utils/extras.py
+++ b/src/poetry/utils/extras.py
@@ -10,7 +10,6 @@ if TYPE_CHECKING:
     from typing import Mapping
 
     from poetry.core.packages.package import Package
-    from typing_extensions import Literal
 
 
 def get_extra_package_names(
@@ -67,7 +66,7 @@ def get_extra_package_names(
     return _extra_packages(extra_package_names)
 
 
-def strtobool(val: str) -> Literal[0, 1]:
+def str_to_bool(val: str) -> bool:
     """Convert a string representation of truth to true (1) or false (0).
 
     True values are 'y', 'yes', 't', 'true', 'on', and '1'; false values
@@ -76,8 +75,8 @@ def strtobool(val: str) -> Literal[0, 1]:
     """
     val = val.lower()
     if val in ("y", "yes", "t", "true", "on", "1"):
-        return 1
+        return True
     elif val in ("n", "no", "f", "false", "off", "0"):
-        return 0
+        return False
     else:
         raise ValueError(f"invalid truth value {val!r}")
--- a/src/poetry/utils/extras.py
+++ b/src/poetry/utils/extras.py
@@ -10,6 +10,7 @@ if TYPE_CHECKING:
     from typing import Mapping
 
     from poetry.core.packages.package import Package
+    from typing_extensions import Literal
 
 
 def get_extra_package_names(
@@ -64,3 +65,19 @@ def get_extra_package_names(
                     yield dependency_package_name
 
     return _extra_packages(extra_package_names)
+
+
+def strtobool(val: str) -> Literal[0, 1]:
+    """Convert a string representation of truth to true (1) or false (0).
+
+    True values are 'y', 'yes', 't', 'true', 'on', and '1'; false values
+    are 'n', 'no', 'f', 'false', 'off', and '0'.  Raises ValueError if
+    'val' is anything else.
+    """
+    val = val.lower()
+    if val in ("y", "yes", "t", "true", "on", "1"):
+        return 1
+    elif val in ("n", "no", "f", "false", "off", "0"):
+        return 0
+    else:
+        raise ValueError(f"invalid truth value {val!r}")
--- a/src/poetry/utils/extras.py
+++ b/src/poetry/utils/extras.py
@@ -76,7 +76,6 @@ def str_to_bool(val: str) -> bool:
     val = val.lower()
     if val in ("y", "yes", "t", "true", "on", "1"):
         return True
-    elif val in ("n", "no", "f", "false", "off", "0"):
+    if val in ("n", "no", "f", "false", "off", "0"):
         return False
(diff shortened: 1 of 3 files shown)


src/poetry/utils/extras.py
```diff
@@ -64,3 +65,19 @@ def _extra_packages(package_names: Iterable[str]) -> Iterator[str]:
                     yield dependency_package_name
 
     return _extra_packages(extra_package_names)
+
+
+def strtobool(val: str) -> Literal[0, 1]:
```
Secrus, line 70: If we are creating our own version, I would prefer it not to be exact copy. 
- name should be `str_to_bool`, because its more readable
- if the name is "..to_bool", then it should return bool (so returns should be `True` and `False`)

Secrus (changes requested):

radoering:
`utils/extras.py` does not seem to be a suitable place to me because that file is about package extras. Maybe, put it in `utils/shell.py`?

Can you provide some simple measurements? Startup time with `distutils` import vs. startup time without it?

mkniewallner:
I'm wondering if having `strtobool` conversion is actually necessary, since we only have one usage of it and I'm not sure it really makes sense.

Looking back at the history, it was added in https://github.com/python-poetry/poetry/pull/201, and the justification for using it is [in this thread](https://github.com/python-poetry/poetry/pull/201/files/c5a1c18cb6b775b1437c7224148cb5fc7a7e32f6#r196260004).

Now looking at the code itself, it seems that we explicitly set `POETRY_ACTIVE` to `"1"`: https://github.com/python-poetry/poetry/blob/0ff5a6a06f97f79740ed2f8c1aa68c079865190a/src/poetry/console/commands/shell.py#L37-L38.

We can definitely confirm that by running:
```console
$ poetry shell
Spawning shell within <REDACTED>

$ echo $POETRY_ACTIVE
1
```

Even if setting a value before entering the shell, we then erase it with `"1"`:
```console
$ POETRY_ACTIVE=invalid_value

$ poetry shell
Spawning shell within <REDACTED>

$ echo $POETRY_ACTIVE
1
```

So the only ways to end up with other values would be if the user explicitly rewrites `POETRY_ACTIVE` value once in a shell through `poetry shell`, or if a shell command rewrites the value, which I'm not sure are cases worth handling.

tddschn:
@mkniewallner Maybe we can change `strtobool(environ.get("POETRY_ACTIVE", "0"))` to just `environ.get("POETRY_ACTIVE")`?

tddschn:
@radoering 

Importing time:
```
$ python3 << EOF
import time

t0 = time.time()
from distutils.util import strtobool
t1 = time.time()
print(t1 - t0)
EOF

0.52388596534729

$ python3 << EOF
import time

t0 = time.time()
from poetry.utils.extras import str_to_bool
t1 = time.time()
print(t1 - t0)
EOF

0.00874018669128418
```

mkniewallner:
> @mkniewallner Yes, it's only used once and `shell.py` doesn't need such a function to work.
> 
> Maybe we can change `strtobool(environ.get("POETRY_ACTIVE", "0"))` to just `environ.get("POETRY_ACTIVE", False)`?

Explicitly checking for `"1"` for truthiness may be safer, but yeah, this would be the idea.

Though I'm not part of the core team, it was simply a possible suggestion based on my limited knowledge of the codebase. We may prefer to keep the current behavior as a safety measure.

src/poetry/utils/extras.py
```diff
@@ -64,3 +64,19 @@ def _extra_packages(package_names: Iterable[str]) -> Iterator[str]:
...
+    are 'n', 'no', 'f', 'false', 'off', and '0'.  Raises ValueError if
+    'val' is anything else.
+    """
+    val = val.lower()
+    if val in ("y", "yes", "t", "true", "on", "1"):
+        return True
+    elif val in ("n", "no", "f", "false", "off", "0"):
+        return False
+    else:
+        raise ValueError(f"invalid truth value {val!r}")
```
mkniewallner, line 82: The `else` is not needed, since all previous branches return something.
A similar logic applies for `elif`, we can use `if` since the previous condition returns something already.
```suggestion
    if val in ("n", "no", "f", "false", "off", "0"):
        return False
    raise ValueError(f"invalid truth value {val!r}")
```

radoering:
I don't think that `POETRY_ACTIVE` is an environment variable a user should tamper, so I'm fine with an exact check.

@neersighted What do you think? Is it sufficient to check `POETRY_ACTIVE` to be present since nobody else then poetry itself should set/remove it or should we keep the current distinction between true and false values?

neersighted:
> I don't think that `POETRY_ACTIVE` is an environment variable a user should tamper, so I'm fine with an exact check.
> 
> @neersighted What do you think? Is it sufficient to check `POETRY_ACTIVE` to be present since nobody else then poetry itself should set/remove it or should we keep the current distinction between true and false values?

My instincts say that it's fine to merely check for a non-empty value -- it's an internal implementation detail of Poetry and typical convention among command line tools is to treat any non-zero value as true.

I would advise against checking for presence alone as you can run into confusion from shell-specific semantics -- e.g. `POETRY_ACTIVE="" poetry` vs `POETRY_ACTIVE= poetry` have different meaning in Bourne-derived shells.

Issue #2825, opened 2020-08-17
Poetry update hangs when prompted for ssh key passphrase


danihodovic:
Re-opening https://github.com/python-poetry/poetry/issues/446 which was automatically closed.

> I successfully added private git repository using poetry add --git ssh://git@gitlab_url/username/repository_name.git package_name. Then I ran poetry update, but it hanged forever on cloning repository (there: https://github.com/sdispater/poetry/blob/master/poetry/vcs/git.py#L99). The problem is that I have to authenticate using my ssh key for cloning repository, git asks for the passphrase, but I cannot enter it.

gehring:
I ran into a similar issue when authenticating with https, e.g., `poetry add git+https://github.com/username/my_private_repo.git`. Oddly enough, running it with `-vvv` allowed me to run it to completion but I had to type in my login info 4 different times!

I had to type it in 1) at the start, 2) while "resolving dependencies", 3) in between "writing lock file" and "Package operations: 1 install, [...]", and 4) at the very end after the final " - Installing my_private_package ..." message.

Izoelektro:
Confirming that this is still an issue as of Poetry version 1.1.3. 

It seems as if the poetry CLI/prompt overshadows the Git one - I am assuming this because for me the poetry prompt (without verbosity) keeps flashing and at times shows the Git prompt. I have even managed to make it work simply by guessing when the Git prompt appeared and typing in my username and passphrase multiple times.

This would explain the current issue (#446) as well as #863 (which was observed to be related already). Funnily enough, they have a similar issue over at [pipenv](https://github.com/pypa/pipenv/issues/3333). Seeing as the CL package used by poetry is built by hand as it is, I am guessing it should not be *that* hard to solve this (although I am far from capable of doing it). I see why it would not be of high priority to do so however.

abn:
It seems that it might be senisble to disable interactions for the git client and document that cases that need credentials should really use https://git-scm.com/docs/gitcredentials.

danihodovic:
@abn Can't you forward stdin / stdout between the git subprocess to the user process when the password prompt is triggered?

abn:
I suspect that will cause more problems than it solves. For example, when multiple operations happen in parallel. Better to keep this simple.

danihodovic:
Usually the ssh key is prompted for once and saved in the ssh-agent.

abn:
[...]

PR #5880, opened 2022-06-19
tell git not to use terminal prompts

### LINKED ISSUE #3222: Poetry inheriting issue for git-client on github.com
- [X] I am on the [latest](https://github.com/python-poetry/poetry/releases/latest) Poetry version. 

- **OS version and name**: Linux Mint 20, Py3.8.2 virtualenv
- **Poetry version**: 1.1.2

## Issue
While `installing` or `locking`, if the github git repo is wrong (e.g. returns a 404 in a browser), then poetry (sometimes) shows an authentication error and ask for the username for that url. 

```Updating dependencies
Resolving dependencies...
   1: fact: actions-gateway is 0.7.0
   1: derived: actions-gateway
Username for 'https://github.com':
```
The pyproject.toml has a git dependency like
```
Flask-Pika = { git = "https://github.com/rienafairefr/flask_pika.git", rev= "b2b4d68186c52ae034b39f4fb56fe86786b3a055"}
```
The typo is hard to see, it should be `flask-pika` instead of `flask_pika`

If the command is run without verbose output, then the "Username for 'https://github.com':" is sometimes shown only for a fraction of a second, so the command may never terminate and it's hard to know why.

Not sure poetry can or should mitigate the problem that comes from a lower level.

The problem comes (pretty sure) from github.com returning a 401 when it should return a 404:
```
GET /inexistent-user/inexistent-repo/info/refs?service=git-upload-pack
Host github.com
User-Agent: git/inexistent-version
```
gives us
```
HTTP/1.1 401 Authorization Required
Server: GitHub Babel 2.0
```
This makes the git client (which is called in a subprocess by poetry) to ask for authentication. 

setting the GIT_ASKPASS variable to false while caling `git` is an option, the credentials to use for a git dependency should be provided by poetry, not leaving `git` to figure it out by itself

dimbleby:
Fixes #3222

See https://git-scm.com/docs/git#Documentation/git.txt-codeGITTERMINALPROMPTcode

This environment variable is available since git 2.3 (2015) but should be harmless even in the unlikely event that someone has an older version than that.

### DIFF
--- a/src/poetry/vcs/git/system.py
+++ b/src/poetry/vcs/git/system.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
+import os
 import subprocess
 
 from typing import TYPE_CHECKING
@@ -49,8 +50,14 @@ class SystemGit:
             ) + args
 
         git_command = find_git_command()
+        env = os.environ.copy()
+        env["GIT_TERMINAL_PROMPT"] = "0"
         return (
-            subprocess.check_output(git_command + list(args), stderr=subprocess.STDOUT)
+            subprocess.check_output(
+                git_command + list(args),
+                stderr=subprocess.STDOUT,
+                env=env,
+            )
             .decode()
             .strip()
         )


dimbleby:
probably also fixes #2825

dimbleby:
Behaviour before on `poetry lock` is to sit there forever at "Resolving dependencies", after is:
```
$ poetry lock
Updating dependencies
Resolving dependencies... (0.6s)

Failed to clone https://github.com/rienafairefr/flask_pika.git, check your git configuration and permissions for this repository.
```

Secrus (approved):

Issue #3352, opened 2020-11-11
Poetry install is stuck on specific dependency with "pending..."


cpvandehey:
- [x] I am on the [latest](https://github.com/python-poetry/poetry/releases/latest) Poetry version.
- [x] I have searched the [issues](https://github.com/python-poetry/poetry/issues) of this repo and believe that this is not a duplicate.
- [x] If an exception occurs when executing a command, I executed it again in debug mode (`-vvv` option).

- **OS version and name**: 10.15.7
- **Poetry version**: 1.1.4, 1.1.3, 1.0.9

## Issue

Hello poetry crew. Loving the work that yall do!

I am having issues with poetry while running `poetry install` in one of my projects on one specific machine.

It appears to get stuck on one dependency during the install and it will hang forever. When I kill the process, I see this in my shell (hinting at a race condtion/locking problem):

```
Error in atexit._run_exitfuncs:
Traceback (most recent call last):
  File "/Users/{}/.pyenv/versions/3.6.8/lib/python3.6/concurrent/futures/thread.py", line 40, in _python_exit
    t.join()
  File "/Users/{}/.pyenv/versions/3.6.8/lib/python3.6/threading.py", line 1056, in join
    self._wait_for_tstate_lock()
  File "/Users/{}/.pyenv/versions/3.6.8/lib/python3.6/threading.py", line 1072, in _wait_for_tstate_lock
    elif lock.acquire(block, timeout):
KeyboardInterrupt
```

I am on the latest poetry (1.1.4) and have tried 1.1.3 as well with the same results (_all of which while running python 3.6.8 (I did use 3.6.5 as well)_). 

After downgrade to 1.0.9, I no longer faced this issue. I am assuming that there was an update to the way poetry can download multiple deps at the same time. Anyone else facing locking issues? At the very least, poetry should recognize a pending download with no progress being made to exit and fail early, but it simply hung and would never exit (left it going for at least 30 mins).

sinoroc:
Maybe it would be helpful to see the `pyproject.toml`. Is that on macos? Were you able to identify if it hangs always on the same dependency or not?

cpvandehey:
Unfortunately, I cannot share the pyproject.toml as it contains a lot of proprietary names.

> Is that on macos? 
As noted above, mac OS 10.15.7 (catalina)

> Were you able to identify if it hangs always on the same dependency or not?
It does indeed hang on the same dependency every time.

One other interesting thing to note:
[...]

PR #5881, opened 2022-06-19
add timeouts to all requests

dimbleby:
https://requests.readthedocs.io/en/latest/user/advanced/#timeouts

> Most requests to external servers should have a timeout attached, in case the server is not responding in a timely manner. By default, requests do not time out unless a timeout value is set explicitly. Without a timeout, your code may hang for minutes or more.

Seems like not a great default.

I've added a timeout to all the requests that I could find.  Probably fixes some of the examples of poetry being "slow" or "stuck" that have been reported eg #3352, though it's hard to be sure.

### DIFF
--- a/src/poetry/publishing/uploader.py
+++ b/src/poetry/publishing/uploader.py
@@ -22,6 +22,7 @@ from requests_toolbelt.multipart import MultipartEncoderMonitor
 from urllib3 import util
 
 from poetry.__version__ import __version__
+from poetry.utils.constants import REQUESTS_TIMEOUT
 from poetry.utils.patterns import wheel_file_re
 
 
@@ -262,7 +263,7 @@ class Uploader:
                         data=monitor,
                         allow_redirects=False,
                         headers={"Content-Type": monitor.content_type},
-                        timeout=5,
+                        timeout=REQUESTS_TIMEOUT,
                     )
                 if resp is None or 200 <= resp.status_code < 300:
                     bar.set_format(
@@ -321,7 +322,7 @@ class Uploader:
             data=encoder,
             allow_redirects=False,
             headers={"Content-Type": encoder.content_type},
-            timeout=5,
+            timeout=REQUESTS_TIMEOUT,
         )
 
         resp.raise_for_status()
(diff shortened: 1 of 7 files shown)


mkniewallner:
Should we make this configurable, like [pip does](https://pip.pypa.io/en/stable/cli/pip/?highlight=timeout#cmdoption-timeout) for instance?
An arbitrary timeout may negatively impact users with unstable connections (especially on mobile networks, or on the move).

I would also suggest to define the default timeout value in a constant to avoid hard coding the value everywhere.

dimbleby:
I'm probably against making this configurable, just another confusing thing for users to worry about.

Not sure I can be bothered about making all these value use a constant either.  I don't know that there's any particularly compelling reason to think that we'd always want them all to have the same value anyway.  Hmm, maybe.

I could see a case for picking a bigger number than 5 though.  5 seconds of inactivity is certainly a slow server, but perhaps not slow enough to completely give up on.  Anyone want to propose a number?

neersighted:
> I'm probably against making this configurable, just another confusing thing for users to worry about.
> 
> Not sure I can be bothered about making all these value use a constant either. I don't know that there's any particularly compelling reason to think that we'd always want them all to have the same value anyway. Hmm, maybe.
> 
> I could see a case for picking a bigger number than 5 though. 5 seconds of inactivity is certainly a slow server, but perhaps not slow enough to completely give up on. Anyone want to propose a number? Perhaps pip's default of 15 is a sensible lead to follow.

I would not be opposed to configuring this through an environment variable, though given that our existing config machinery reads from `poetry.toml` and environment variables it might be less confusing to allow for both. It's definitely not a knob I'm eager to add, but I do see it having practical use in some scenarios (hence an environment variable as a 'hidden' knob).

dimbleby:
I'll make it a constant and if y'all are motivated to make that configurable later then at least it'll only be one place to change.

mkniewallner:
> I'm probably against making this configurable, just another confusing thing for users to worry about.
> 
I don't think this would confuse users, since the idea would be that most users won't even need to worry about configuring the timeout (since we have a sane default), but they would still have a knob to turn if they ever need to.

Case in point, I almost never use the `--timeout` option from `pip`, but it happened that I needed to use it while working on a train, with an ever-changing mobile connection through an hotspot and great chances of having timeouts, so I was glad to have the option when I needed to (even if most of the time, I don't have to worry about it).

dimbleby:
I won't be adding code to make this configurable.

Feel free to do as you see fit with this MR: merge it and add to it, or merge it and leave it alone, or reject it.

neersighted (approved):

Issue #5902, opened 2022-06-23
Poetry repeadetly re-downloads url dependencies during dependency solving


Jinior:
- [x] I am on the [latest](https://github.com/python-poetry/poetry/releases/latest) Poetry version.
- [x] I have searched the [issues](https://github.com/python-poetry/poetry/issues) of this repo and believe that this is not a duplicate.
- [x] If an exception occurs when executing a command, I executed it again in debug mode (`-vvv` option).

- **OS version and name**: ArchLinux with kernel 5.18.5-arch1-1
- **Poetry version**: version 1.2.0b3.dev0
- **Link of a [Gist](https://gist.github.com/) with the contents of your pyproject.toml file**: [https://gist.github.com/Jinior/3dbc15a9118dc3dd470c3f2c9ad33124](https://gist.github.com/Jinior/3dbc15a9118dc3dd470c3f2c9ad33124)

## Issue

When solving dependencies poetry repeadetly re-downloads url dependencies. As can be seen in the attached pyproject.toml I have a url dependency which is downloaded many times during solving. This not only makes the resolving slower, but also makes it impossible in some cases because the server will give "429 too many tries" error after a while.

When running the `poetry lock -vvv` command the output shows that the package is constantly redownloaded.

I suspect the bug is due to cache misses in the deferred cache of the Provider class.

dimbleby:
duplicate #2415

Jinior:
It is indeed a duplicate. When I initially read #2415 I thought it was about re-downloading a package during installing while it was already downloaded during locking; while my issue is about re-downloading packages during only locking. But I now see that #2415 covers both cases.

PR #5903, opened 2022-06-23
Use hash of dependency instead of object in a provider's deferred cache

Jinior:
# Pull Request Check List

Resolves: #5902

- [n.a.] Updated **documentation** for changed code. 

This is my first pull request so apologies if I did something wrong, I am open to feedback!

As described in the referred issue poetry often repeatedly re-downloads packages when resolving dependencies. This is due a bug which causing poetry thinking cached packages are not cached yet. The providers cache is a dictionary which uses a Dependency object a key (and the package as the value). During a dictionary lookup python checks equivalence on both the hash and __eq__ for keys, but in some cases the Dependency object can change breaking the __eq__ check. 

This for example happens with the URL dependency, where the package version in only known after downloading which leads to the following:

1. we check if the dependency is in the cache using the URLDependency as the key
2. we download the package
3. we check the package version from the downloaded file
4. we update the version in the dependency object
5. we store the package into the cache using the updated dependency object

This is shown in the code snippet below:
```python
if dependency in self._deferred_cache:
    return [self._deferred_cache[dependency]]

package = self.get_package_from_url(dependency.url)

self.validate_package_for_dependency(dependency=dependency, package=package)

# truncated irrelevant stuff for this example

dependency._constraint = package.version
dependency._pretty_constraint = package.version.text

self._deferred_cache[dependency] = package
```

When searching for this dependency again we check the cache dictionary with a new URLDependency as a key, which has not been updated with the version number yet. We have missed the cache.

Adding the package to the cache before updating the dependency object is not an elegant solution in my opinion because that would make a package search with the updated dependency object miss the cache.

This pull request solves the problem by using the hashes of the Dependency objects as keys in the cache dictionary. The currently implemented __hash__ methods already uniquely describe the package they download. (For example the URLDependency hash is made up of the url and package name; the VCS cache hash is made up of the package name, repository, branch, tag, and revision).

Another approach would have been to re-implement the __eq__ method but I think this is also the wrong approach, because an URL dependency with an unbounded version is different than a URL dependency with a set version.

### DIFF
--- a/src/poetry/puzzle/provider.py
+++ b/src/poetry/puzzle/provider.py
@@ -288,7 +288,7 @@ class Provider:
         dependency._source_resolved_reference = package.source_resolved_reference
         dependency._source_subdirectory = package.source_subdirectory
 
-        self._deferred_cache[dependency] = package
+        self._deferred_cache[hash(dependency)] = package
 
         return [package]
 
@@ -315,8 +315,8 @@ class Provider:
         )
 
     def search_for_file(self, dependency: FileDependency) -> list[Package]:
-        if dependency in self._deferred_cache:
-            _package = self._deferred_cache[dependency]
+        if hash(dependency) in self._deferred_cache:
+            _package = self._deferred_cache[hash(dependency)]
 
             package = _package.clone()
         else:
@@ -325,7 +325,7 @@ class Provider:
             dependency._constraint = package.version
             dependency._pretty_constraint = package.version.text
 
-            self._deferred_cache[dependency] = package
+            self._deferred_cache[hash(dependency)] = package
 
         self.validate_package_for_dependency(dependency=dependency, package=package)
 
@@ -352,8 +352,8 @@ class Provider:
         return package
 
     def search_for_directory(self, dependency: DirectoryDependency) -> list[Package]:
-        if dependency in self._deferred_cache:
-            _package = self._deferred_cache[dependency]
+        if hash(dependency) in self._deferred_cache:
+            _package = self._deferred_cache[hash(dependency)]
 
             package = _package.clone()
         else:
@@ -362,7 +362,7 @@ class Provider:
             dependency._constraint = package.version
             dependency._pretty_constraint = package.version.text
 
-            self._deferred_cache[dependency] = package
+            self._deferred_cache[hash(dependency)] = package
 
         self.validate_package_for_dependency(dependency=dependency, package=package)
 
@@ -378,8 +378,8 @@ class Provider:
         return PackageInfo.from_directory(path=directory).to_package(root_dir=directory)
 
     def search_for_url(self, dependency: URLDependency) -> list[Package]:
-        if dependency in self._deferred_cache:
-            return [self._deferred_cache[dependency]]
+        if hash(dependency) in self._deferred_cache:
+            return [self._deferred_cache[hash(dependency)]]
 
         package = self.get_package_from_url(dependency.url)
 
@@ -396,7 +396,7 @@ class Provider:
         dependency._constraint = package.version
         dependency._pretty_constraint = package.version.text
 
-        self._deferred_cache[dependency] = package
+        self._deferred_cache[hash(dependency)] = package
 
         return [package]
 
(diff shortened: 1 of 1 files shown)


radoering:
Thanks, for your contribution and the detailed information. I'm not sure whether the `__eq__` approach isn't better. Maybe, `__eq__` should not consider the constraint for direct origin dependencies since the constraint is always implicitly given? Two direct origin dependencies with the same origin can't have different constraints, can they? The result should be the same as using the hash in `_deferred_cache`.

Maybe, we should try if replacing

```
return super().__eq__(other) and self._constraint == other.constraint
```

in

https://github.com/python-poetry/poetry-core/blob/4f9c35272cf53f56596c8a71175da6e50bb71b13/src/poetry/core/packages/dependency.py#L577

by

```
return super().__eq__(other) and (
    self._constraint == other.constraint or self.is_direct_origin()
)
```

does the trick, too. Do you want to try and create an alternative PR if it works so we can decide which path to take?

dimbleby:
@radoering yes this addresses a different one of the downloads.  I think we have (at least!):

- during locking poetry twice goes through `search_for_url()`, this MR is aimed at making sure that one download deals with both of those

- during installation poetry again downloads wheels, #5871 is aimed at arranging that if it already has the wheel in its cache then it can reuse it

Ideally poetry should also arrange that the `search_for_url()` download goes into the cache so that it can be reused during installation.

Jinior:
> Thanks, for your contribution and the detailed information. I'm not sure whether the `__eq__` approach isn't better. Maybe, `__eq__` should not consider the constraint for direct origin dependencies since the constraint is always implicitly given? Two direct origin dependencies with the same origin can't have different constraints, can they? The result should be the same as using the hash in `_deferred_cache`.
> 
> Maybe, we should try if replacing
> 
> ```
> return super().__eq__(other) and self._constraint == other.constraint
> ```
> 
> in
> 
> https://github.com/python-poetry/poetry-core/blob/4f9c35272cf53f56596c8a71175da6e50bb71b13/src/poetry/core/packages/dependency.py#L577
> 
> by
> 
> ```
> return super().__eq__(other) and (
>     self._constraint == other.constraint or self.is_direct_origin()
> )
> ```
> 
> does the trick, too. Do you want to try and create an alternative PR if it works so we can decide which path to take?

@radoering thank you for the suggestion. I shall create a second PR today. 
You are correct that two direct origin dependencies should have the the same constraints. But theoretically it could happen when a branch or path gets updated during locking but that is an unlikely situation.

I am unsure how `__eq__` will be used in the future.  I can imagine that a developer would assume that two dependencies with different constraints will not test equal. Perhaps they would be testing if the direct origin has changed. 

What also comes to mind now is that we only update the constraint field in the dependency if it is not found in the cache; when we do have a cache hit we immediately return the package. (at least for the URLDependency). Should we change that?

radoering:
Additional tests are usually a good idea. 😄

A simple unit test in `test_provider.py` could look like this:

- mock the method that downloads the file
- search for dependency
- change constraint of dependency
- search for dependency (with changed constraint) again
- check that "download method" was only called once

You might also write a test for `version_solver` that triggers the repeated downloads and check that the "download method" is only called once. Although this makes the reason for the test more explicit, I think a simple unit test with a comment about the use case might be enough.

radoering:
Using the hash as key probably makes `_deferred_cache` unusable for pure `Dependency` (because `constraint` is not considered in hash) and `VCSDependency` (because neither `source_reference` nor `source_resolved_reference` is considered in hash). It seems the former is not relevant because `_deferred_cache` is only used for direct origin dependencies. However, the latter seems to be relevant. I suppose, using the hash as key will break use cases were the same repo is used twice with different references (branch, tag, ...). Thus, python-poetry/poetry-core#405 might be the better alternative.

radoering:
Due to the shortcomings mentioned in my previous comment I'm closing this one in favor of python-poetry/poetry-core#405. A test checking that a dependency is only downloaded once is still appreciated. Please create a separate PR if you want to write such a test and note that the test should fail until a new poetry-core version is released and used in master of poetry.

PR #5904, opened 2022-06-23
refactor the search for direct-origin dependencies

dimbleby:
refactoring suggested by #5903. 

- all the `search_for_xxx()` methods returned a list of length one, which is a bit weird
- the whole pattern is repeated three times, let's just do it once.

IMO it's kinda ugly that `search_for_xxx(dependency)` makes updates to `dependency`, that doesn't match my intuition for what such a method ought to do.  But trying to refactor that broke more tests than I care to investigate.

In particular this rearrangement ensures that the updates that we make to the input dependency are the same regardless of whether we got there via the `_deferred_cache` or not.

### DIFF
--- a/src/poetry/puzzle/provider.py
+++ b/src/poetry/puzzle/provider.py
@@ -221,6 +221,42 @@ class Provider:
         )
         return packages
 
+    def search_for_direct_origin_dependency(self, dependency: Dependency) -> Package:
+        cached = self._deferred_cache.get(dependency)
+        if cached is not None:
+            return cached
+
+        if dependency.is_vcs():
+            dependency = cast(VCSDependency, dependency)
+            package = self.search_for_vcs(dependency)
+            dependency._source_reference = package.source_reference
+            dependency._source_resolved_reference = package.source_resolved_reference
+            dependency._source_subdirectory = package.source_subdirectory
+
+        elif dependency.is_file():
+            dependency = cast(FileDependency, dependency)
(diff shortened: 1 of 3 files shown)


src/poetry/puzzle/provider.py
```diff
@@ -221,6 +221,42 @@ def search_for_installed_packages(
         )
         return packages
 
+    def search_for_direct_origin_dependency(self, dependency: Dependency) -> Package:
+        cached = self._deferred_cache.get(dependency)
+        if cached is not None:
```
Jinior, line 226: Should we also update dependency._constraint and dependency._pretty_constraint here?

dimbleby: oh dear, that was even half of my point in the first place...

radoering (commented):
Have you thought about the following changes in behavior? (Not sure about the consequences or if it's relevant at all.)

`package` is not cloned anymore for file and directory dependencies. Thus, `package.root_dir`, `package.files`, `package.develop` are set according to the first dependency.

It seems like VCS dependencies were only put into the cache but never taken from it. Seems like a bug, but had the same effect as cloning. Now, `package.develop` is set according to the first dependency.

dimbleby:
@radoering yes I noticed those things, but couldn't think of a reason that they should matter - so chose to trust the test suite when it said that it saw no problem.

radoering:
Me neither. Should be fine.

radoering (approved):

PR #5910, opened 2022-06-25
every Repository is named

dimbleby:
Per the FIXME that is removed by this MR, nameless repositories are undesirable because they'll confuse a`Pool` of repositories.

I've made sure that all repositories are named:
- sometimes by giving repositories somewhat artificial names, but always "poetry-" something so as to make clashes with user-defined repositories unlikely
- sometimes by recognising that where we were using a nameless repository, all that we really needed was a list of packages.

### DIFF
--- a/src/poetry/console/commands/debug/resolve.py
+++ b/src/poetry/console/commands/debug/resolve.py
@@ -85,7 +85,7 @@ class DebugResolveCommand(InitCommand):
 
         pool = self.poetry.pool
 
-        solver = Solver(package, pool, Repository(), Repository(), self.io)
+        solver = Solver(package, pool, [], [], self.io)
 
         ops = solver.solve().calculate_operations()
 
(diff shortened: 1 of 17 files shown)


dimbleby:
ugh, tests failing because of poetry-export-plugin, see https://github.com/python-poetry/poetry-plugin-export/pull/79

radoering (approved):
New release of poetry-plugin-export required for tests to pass.

dimbleby:
> New release of poetry-plugin-export required for tests to pass.

indeed.  Also #5711, #5819 are even more tangled (and #5711 also needs poetry-core to be released before it can progress).

Cross-repository testing is awkward.

dimbleby:
release of export plugin 1.0.5 should give us green pipelines here and allow this to merge

neersighted (approved):

PR #5918, opened 2022-06-26
remove a lot of casts

dimbleby:
I reckon `cast` should be the last resort when appeasing the typechecker, mostly because it leaves you in a bad place for refactoring: once a `cast` is in the codebase it's all too easy to do some refactoring such that it's no longer true, but nothing can tell you so.

So I've gone and looked for places where `cast` can sensibly be removed altogether, or - in a couple of places - replaced by assertions.

I've also realised that if you `cast("Foo", foo)` instead of `cast(Foo, foo)` then you only need to import `Foo` during typechecking, which should be a small win.

### DIFF
--- a/src/poetry/console/application.py
+++ b/src/poetry/console/application.py
@@ -14,7 +14,6 @@ from cleo.events.console_events import COMMAND
 from cleo.events.event_dispatcher import EventDispatcher
 from cleo.exceptions import CleoException
 from cleo.formatters.style import Style
-from cleo.io.inputs.argv_input import ArgvInput
 from cleo.io.null_io import NullIO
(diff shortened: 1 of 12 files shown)


dimbleby:
Hmm, 3.11-dev tests look a bit flakey - certainly I don't think my MR is responsible for their failure.

Perhaps it was too soon to add them to the matrix?  Or perhaps Someone would like to figure out what the problem is...?

neersighted (approved):
LGTM -- regarding the tests, they're optional in the matrix and I believe the issue is a bad dev tag/pip release combo. Ignoring it for now.

Issue #5331, opened 2022-03-18
uninstall / reinstall not working due to missing pyvenv.cfg


bradfordgp:
OS:  Windows 11
Python version:  3.10.3
Initial poetry install via Powershell command:  ` (Invoke-WebRequest -Uri https://install.python-poetry.org -UseBasicParsing).Content | python -`

Had some problems with initial install, so ran uninstall with no problems.  However, on attempting to reinstall, got the following message:
```
Installing Poetry (1.1.13)
Installing Poetry (1.1.13): Creating environment
Actual environment location may have moved due to redirects, links or junctions.
  Requested location: "C:\Users\321gp\AppData\Roaming\pypoetry\venv\Scripts\python.exe"
  Actual location:    "C:\Users\321gp\AppData\Local\Packages\PythonSoftwareFoundation.Python.3.10_qbz5n2kfra8p0\LocalCache\Roaming\pypoetry\venv\Scripts\python.exe"
Actual environment location may have moved due to redirects, links or junctions.
  Requested location: "C:\Users\321gp\AppData\Roaming\pypoetry\venv\Scripts\python.exe"
  Actual location:    "C:\Users\321gp\AppData\Local\Packages\PythonSoftwareFoundation.Python.3.10_qbz5n2kfra8p0\LocalCache\Roaming\pypoetry\venv\Scripts\python.exe"
Installing Poetry (1.1.13): An error occurred. Removing partial environment.
Poetry installation failed.
See C:\Users\321gp\poetry-installer-error-2yr835li.log for error logs.
```
Error log contains the following:
```
No pyvenv.cfg file

Traceback:

  File "<stdin>", line 879, in main
  File "<stdin>", line 515, in run
  File "<stdin>", line 536, in install
  File "C:\Program Files\WindowsApps\PythonSoftwareFoundation.Python.3.10_3.10.1008.0_x64__qbz5n2kfra8p0\lib\contextlib.py", line 135, in __enter__
    return next(self.gen)
  File "<stdin>", line 608, in make_env
  File "<stdin>", line 594, in make_env
  File "<stdin>", line 326, in make
  File "<stdin>", line 349, in pip
  File "<stdin>", line 346, in python
  File "<stdin>", line 339, in run
```
Checking the links listed, `C:\Users\... \Roaming\pypoetry\venv\` does not exist.

Suggestions on how to correct this?

yodaldevoid:
[...]

PR #5931, opened 2022-06-29
fix: fix venv paths redirected by Python from MS Store

finswimmer:
Running Poetry with Python installed from the Microsoft Store have to tackle multiple issues.

The system's Python interpreter is located at something like `C:\Program Files\WindowsApps\PythonSoftwareFoundation.Python.3.9_3.9.3568.0_x64__qbz5n2kfra8p0`. But the user is not allowed to run any executable from there. Instead one should use the binaries locates at `C:\Users\username\AppData\Local\Microsoft\WindowsApps\PythonSoftwareFoundation.Python.3.9_qbz5n2kfra8p0\`.

Another problem us, that applications from the Microsoft Store are not allowed to write directly into `%LocalAppdata%`, instead any attempt to do this is redirected to a sandbox folder located at `C:\Users\username\AppData\Local\Packages\PythonSoftwareFoundation.Python.3.9_qbz5n2kfra8p0`.

The first problem described above is an issue, when trying to find the interpreter, that should be used for creating a new venv. This can be solved by replacing the `%ProgramFiles%` part of the path by `%LocalAppdata/Microsoft%`.

The second problem described above is an issue, when a new venv should be created at the default location under `C:\Users\klare\AppData\Local\pypoetry\Cache\virtualenvs`. Fortunately when one try to access this place during runtime, the `Path` object is a symlink that can be resolved and than be passed to the subprocess call that creates the venv.

This `install-poetry.py` script is affected by the same problems. A possible fix is prepared at https://github.com/python-poetry/install.python-poetry.org/pull/18

Resolves: #5331

### DIFF
--- a/src/poetry/utils/env.py
+++ b/src/poetry/utils/env.py
@@ -36,10 +36,12 @@ from poetry.core.toml.file import TOMLFile
 from poetry.core.utils.helpers import temporary_directory
 from virtualenv.seed.wheels.embed import get_embed_wheel
 
+from poetry.utils._compat import WINDOWS
 from poetry.utils._compat import decode
 from poetry.utils._compat import encode
 from poetry.utils._compat import list_to_shell_command
 from poetry.utils._compat import metadata
+from poetry.utils.helpers import get_real_windows_path
 from poetry.utils.helpers import is_dir_writable
 from poetry.utils.helpers import paths_csv
 from poetry.utils.helpers import remove_directory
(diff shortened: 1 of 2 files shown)


neersighted (changes requested):
It looks like the typing is a bit of a mess -- the `get_real_windows_path` function takes a Path but immediately coerces it back to a string. Why don't we make it accept a Union of str or Path since that is what goes on internally anyway?

finswimmer:
> It looks like the typing is a bit of a mess -- the `get_real_windows_path` function takes a Path but immediately coerces it back to a string. Why don't we make it accept a Union of str or Path since that is what goes on internally anyway?

Good idea :+1:

neersighted (approved):

PR #5944, opened 2022-07-02
[1.1] Recommend new installer for `1.1` documentation

mkniewallner:
# Pull Request Check List

We often encounter new issues about `get-poetry.py` installer, despite it being deprecated and replaced by `install.python-poetry.org`.
Part of this is because while the documentation mentions that the installer is deprecated, the installation instructions of `1.1` documentation still use `get-poetry.py`.

Since `install.python-poetry.org` can install `1.1.x` just fine, this PR updates the instructions to recommend it in place of `get-poetry.py`.
This should hopefully reduce issues related to `get-poetry.py`.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -3,6 +3,9 @@ repos:
     rev: 19.10b0
     hooks:
       - id: black
+        # Pinning click because of https://github.com/psf/black/issues/2964
+        additional_dependencies: [click==8.0.4]
+
 
   - repo: https://gitlab.com/pycqa/flake8
     rev: 3.8.3
--- a/docs/_index.md
+++ b/docs/_index.md
@@ -32,19 +32,25 @@ Poetry provides a custom installer that will install `poetry` isolated
 from the rest of your system by vendorizing its dependencies. This is the
 recommended way of installing `poetry`.
 
-{{% note %}}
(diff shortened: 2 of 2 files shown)


finswimmer:
Hey,

unfortunately this is not that easy. Poetry 1.1 still supports running with Python 2.7, but the new install script does not support this anymore.

To avoid having two install script instructions in the same docs, we just linked to the new instructions for the new installer in master.

mkniewallner:
> unfortunately this is not that easy. Poetry 1.1 still supports running with Python 2.7, but the new install script does not support this anymore.

What about a note/warning before the instructions to link to https://github.com/python-poetry/poetry/blob/28dc4c4e99b28a479c29f96b20e3767c41082701/docs/_index.md#installation for Python 2.7 users?

Not the cleanest solution, but given the, I presume, small number of people still using Python 2.7, they would still have proper instructions to install Poetry, so this could be a good tradeoff if it could avoid us more issues that mention `get-poetry.py`.

johnthagen:
It sounds like this won't be merged right away, but if it were to, could we have a discussion about

- #5917 

In doing some testing of the new installer on Windows, I found that the new script doesn't add `poetry` to the `PATH`, which feels like a major user experience regression, especially for novice users and those new to Poetry.

PR #5946, opened 2022-07-03
Small poetry init fixes

wieczorek1990:
* Fixed invalid authors field population
* Fixed no info about skipping while adding packages

### DIFF
--- a/src/poetry/console/commands/init.py
+++ b/src/poetry/console/commands/init.py
@@ -281,7 +281,7 @@ You can specify a package in the following forms:
                 ):
                     self.line(f"Adding <info>{package}</info>")
                     result.append(constraint)
-                    package = self.ask("\nAdd a package:")
+                    package = self.ask("\nAdd a package (leave blank to skip):")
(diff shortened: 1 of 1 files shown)


dimbleby:
Why removing apostrophes - do they cause a problem?

They're quite common in some names - eg Ronnie O'Sullivan.  Why should they not be allowed in poetry projects?

wieczorek1990:
Removing only from the beginning and ending.
Now instead of:
Łukasz Wieczorek ***@***.***>
I get:
’Łukasz Wieczorek’ ***@***.***’>
It is a bug and does cause error when publishing.

W dniu niedz., 3.07.2022 o 15:15 David Hotham ***@***.***>
napisał(a):

wieczorek1990:
I had bad data inputed in `.gitconfig` with single quotation marks.
The problem is not existent.
Now the patch only fixes messages in `poetry init` to add info about skipping.

mkniewallner (approved):
Thanks!

PR #5963, opened 2022-07-07
Copyedit faq.md

dymil:
"sometimes, with some specific set of dependencies" is shortened and "no choice but downloading [...] and inspect" is made parallel/more idiomatic

# Pull Request Check List

Resolves: N/A

- [N/A] Added **tests** for changed code.

### DIFF
--- a/docs/faq.md
+++ b/docs/faq.md
@@ -14,12 +14,12 @@ menu:
 ### Why is the dependency resolution process slow?
 
 While the dependency resolver at the heart of Poetry is highly optimized and
-should be fast enough for most cases, sometimes, with some specific set of dependencies,
+should be fast enough for most cases, with certain dependencies
 it can take time to find a valid solution.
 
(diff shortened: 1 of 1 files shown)


docs/faq.md
```diff
@@ -14,12 +14,12 @@ menu:
 ### Why is the dependency resolution process slow?
 
 While the dependency resolver at the heart of Poetry is highly optimized and
-should be fast enough for most cases, sometimes, with some specific set of dependencies,
```
Secrus, line 17: This should be left as it is. The new version describes a single dependency, while the old one (correctly) describes that there are sets of dependencies that cause slow resolution.

dymil: The original sounded wrong to me, as "some specific set of dependencies" sounds like it's referring to one specific set. I committed a tweaked version which would be a sidegrade from "some specific _sets_ of dependencies"

Secrus (approved):

Issue #5970, opened 2022-07-08
Breaking change in Warehouse breaks poetry lock file creation


cyraxjoe:
- [x] I am on the [latest](https://github.com/python-poetry/poetry/releases/latest) Poetry version.
- [x] I have searched the [issues](https://github.com/python-poetry/poetry/issues) of this repo and believe that this is not a duplicate.
- [x] If an exception occurs when executing a command, I executed it again in debug mode (`-vvv` option).

- **Linux/Debian bookworm**: 
- **1.2.0b2**: 
- **Sample [pyproject](https://gist.github.com/cyraxjoe/9fb478d87ef69bef839f771e0f1561bf)**: 

## Issue

The warehouse project (backend of pypi.org) applied a [breaking change ](https://github.com/pypi/warehouse/pull/11775) that affects `poetry lock` the `releases` key is no longer present in the responses, required for the [pypi_repository](https://github.com/python-poetry/poetry/blob/master/src/poetry/repositories/pypi_repository.py#L207) implementation.

It can be easily verified by removing the local cache and inspecting the generated lock file, it would have an empty `metadata.files` property, something like: 
```
[[package]]
name = "beautifulsoup4"
version = "4.11.1"
description = "Screen-scraping library"
category = "main"
optional = false
python-versions = ">=3.6.0"

[package.dependencies]
soupsieve = ">1.2"

[package.extras]
html5lib = ["html5lib"]
lxml = ["lxml"]

[[package]]
name = "soupsieve"
version = "2.3.2.post1"
description = "A modern CSS selector implementation for Beautiful Soup."
category = "main"
optional = false
python-versions = ">=3.6"

[metadata]
lock-version = "1.1"
python-versions = "^3.10"
content-hash = "36fc53f6339e375f6295378c7a8ecea11e3a04ba0adcbfa7aeb812e30500aad0"

[metadata.files]
beautifulsoup4 = []
soupsieve = []
```

Kurt-von-Laven:
I can confirm that this issue is also affecting Poetry 1.1.13 as it is taking down all of our builds.

novotl:
👍 We're seeing the same with Poetry 1.1.8

dimbleby:
I suspect a lot of people are about to learn a hard lesson about the dangers of relying on public repositories....

```diff
diff --git a/src/poetry/repositories/pypi_repository.py b/src/poetry/repositories/pypi_repository.py
index 0e5beab6..142b29d8 100644
[...]

Issue #5967, opened 2022-07-07
Poetry not passing through SHAs for certain package versions.


TNonet:
- [x] I am on the [latest](https://github.com/python-poetry/poetry/releases/latest) Poetry version.
- [x] I have searched the [issues](https://github.com/python-poetry/poetry/issues) of this repo and believe that this is not a duplicate.
- [x] If an exception occurs when executing a command, I executed it again in debug mode (`-vvv` option).

- **OS version and name**: 'macOS-12.0.1-arm64-arm-64bit'
- **Poetry version**:  1.1.13
- **Link of a [Gist](https://gist.github.com/) with the contents of your pyproject.toml file**: (https://gist.github.com/TNonet/c3774fdc624cd2d183997e18c9a453d4)

## Issue

Poetry is not passing through SHA values of packages in the `poetry.lock` file.

MWE:
```bash
poetry new test-sha
cd test-sha
poetry add requests==2.27.1
```
```
Updating dependencies
Resolving dependencies... (0.1s)

Writing lock file

Package operations: 5 installs, 0 updates, 0 removals

  • Installing certifi (2022.6.15)
  • Installing charset-normalizer (2.0.12)
  • Installing idna (3.3)
  • Installing urllib3 (1.26.10)
  • Installing requests (2.27.1)
```

Results in a `poetry.lock` file that is missing hashes.
```
...
packaging = []
pluggy = [
    {file = "pluggy-0.13.1-py2.py3-none-any.whl", hash = "sha256:966c145cd83c96502c3c3868f50408687b38434af77734af1e9ca461a4081d2d"},
    {file = "pluggy-0.13.1.tar.gz", hash = "sha256:15b2acde666561e1298d71b523007ed7364de07029219b604cf808bfa1c765b0"},
]
py = []
pyparsing = [
    {file = "pyparsing-3.0.9-py3-none-any.whl", hash = "sha256:5026bae9a10eeaefb61dab2f09052b9f4307d44aee4eda64b309723d8d206bbc"},
    {file = "pyparsing-3.0.9.tar.gz", hash = "sha256:2b020ecf7d21b687f219b71ecad3631f644a47f01403fa1d1036b0c6416d70fb"},
]
pytest = [
    {file = "pytest-5.4.3-py3-none-any.whl", hash = "sha256:5c0db86b698e8f170ba4582a492248919255fcd4c79b1ee64ace34301fb589a1"},
[...]

PR #5972, opened 2022-07-08
fix(repositories): use `urls` for versioned JSON API instead of `releases`

mkniewallner:
# Pull Request Check List

Resolves: #5967 
Resolves: #5970 

When retrieving package metadata for a specific version from `pypi.org` JSON API, Poetry currently relies on a `releases` keys to look for the correct version based on the URL.
Following https://github.com/pypi/warehouse/pull/11775, `releases` has been removed from `/pypi/<package_name>/<version>/json` endpoints, so Poetry can't pick up this metadata anymore.

There's no real reason to use `releases`, since `urls` is also available, and has the advantage of directly returning the version we are looking for (so instead of checking for `data["releases"][version]`, we directly access the version we are looking for in `data["urls"]`).

Note: Given that this makes Poetry locking mechanism broken on all released versions (not only on `master`), this would probably require a backport to the `1.1` branch, even if it is mostly unmainted. Started 
https://github.com/mkniewallner/poetry/commits/fix/fix-releases-key-pypi-repository-1.1 for that.


mkniewallner:
Tests pass, which means we definitely lack some test coverage for this.

finswimmer:
> Tests pass (minus the flaky 3.11-dev on macOS), which means we definitely lack some test coverage for this.

I think the `test_package` test in `test_pypi_repository.py` is a good candidate for extending the test. We have to check for the content of the `files`  attribute.

mkniewallner:
> Probably the json files under `tests/repositories/fixtures/pypi.org/json` must be adopted.

Agree, though `releases` is also used in https://github.com/python-poetry/poetry/blob/584aaee5f8da9f950f98d1ef0ebc06fe835e5d96/src/poetry/repositories/pypi_repository.py#L66
So we will probably need to change that too.

dimbleby:
> will probably need to change that too.

I don't think so - I think that is looking at the reponse from the non-versioned page, which continues to include this information

mkniewallner:
> I don't think so - I think that is looking at the reponse from the non-versioned page, which continues to include this information

Oh yeah, https://github.com/pypi/warehouse/pull/11775/files#diff-beb4788f749d9d561005b6542d39b9f7d6f0ed2e7e9bd8b614673294aab9b488R188-R189 you're right indeed, `/pypi/<project_name>/json` is untouched, my bad for overlooking that.

mkniewallner:
Updated the PR to:
- add a dedicated version fixture for `requests` in https://github.com/python-poetry/poetry/pull/5972/commits/9f29869d7377086015e1348af766962ebe6d7d66 (so it uses this file instead of [the global one](https://github.com/python-poetry/poetry/blob/9cf87a285a2d3fbb0b9fa621997b3acc3631ed24/tests/repositories/fixtures/pypi.org/json/requests.json), according to [this code](https://github.com/python-poetry/poetry/blob/9f29869d7377086015e1348af766962ebe6d7d66/tests/repositories/test_pypi_repository.py#L45-L50))
- remove all `releases` key from the versioned fixtures in https://github.com/python-poetry/poetry/pull/5972/commits/6b40ebc59761f19be95d14126ef4639ab236ec18

finswimmer (approved):
LGTM and also tested this locally :medal_sports: :rocket: 

Can you please prepare a backport to 1.1?

mkniewallner:
> Can you please prepare a backport to 1.1?

Yes, PR is https://github.com/python-poetry/poetry/pull/5973, just waiting for CI to pass.

mbelang:
Updated to 1.1.14 and still doing it. Is there something to modify to actually fix the issue locally?

mkniewallner:
> Updated to 1.1.14 and still doing it. Is there something to modify to actually fix the issue locally?

Yes, you have to clear Poetry's cache first:
```bash
poetry cache clear pypi --all
```

mbelang:
Yeah it works thx. I found the answer in one of the issues.

dwangfaulkner-fn:
Sorry to jump onto this thread but is there anywhere to see when a new beta release will be cut with this very critical fix? We'd prefer to use a released version instead of building from main, as I'm sure many other folks would.

mkniewallner:
> Sorry to jump onto this thread but is there anywhere to see when a new beta release will be cut with this very critical fix? We'd prefer to use a released version instead of building from main, as I'm sure many other folks would.

A new beta for `1.2` has been released: https://github.com/python-poetry/poetry/releases/tag/1.2.0b3.

Issue #5586, opened 2022-05-10
1.2.0 Release


abn:
Trying to gather the list of actual blockers for 1.2.0 final release. If there is a PR open that absolutely must be included in the 1.2.0 release, please link them here. Issues, unless they are deemed critical (ie. regression that fundamentally breaks workflows), are unlikely to be considered at this stage.

A link and a short note on why you think is important would suffice. If you agree that a linked issue/pr in the comments is important use :+1:  else use :-1:  to vote.

Let's avoid bike shedding in here so that the maintainers can focus on the signal.

# Blockers
These are a subset of the [1.2 Milestone](https://github.com/python-poetry/poetry/milestone/5).

- https://github.com/python-poetry/poetry-core/pull/348
- https://github.com/python-poetry/poetry/pull/5673
- https://github.com/python-poetry/poetry/pull/5450
- https://github.com/python-poetry/poetry/pull/5581
- https://github.com/python-poetry/poetry/issues/4242
- https://github.com/python-poetry/poetry-plugin-export/pull/56
- https://github.com/python-poetry/poetry/pull/5587
- https://github.com/python-poetry/poetry/pull/5590
- https://github.com/python-poetry/poetry/pull/6234
- https://github.com/python-poetry/poetry/issues/5593
- https://github.com/python-poetry/poetry/issues/5633
- https://github.com/python-poetry/poetry/issues/5675
- https://github.com/python-poetry/poetry/issues/5682
- https://github.com/python-poetry/poetry/issues/4718
- https://github.com/python-poetry/poetry/issues/5715
- https://github.com/python-poetry/poetry/issues/5727
- https://github.com/python-poetry/poetry/pull/5725
- https://github.com/python-poetry/poetry/issues/5331
- https://github.com/python-poetry/poetry/issues/6054
- https://github.com/python-poetry/poetry/issues/6184

# Nice to haves
- https://github.com/python-poetry/poetry/pull/5811
- https://github.com/python-poetry/poetry/pull/5172
- https://github.com/python-poetry/poetry/pull/5451
- https://github.com/python-poetry/poetry/pull/5442
- https://github.com/python-poetry/poetry/pull/5401
- https://github.com/python-poetry/poetry/pull/5385
- https://github.com/python-poetry/poetry/pull/5718
- https://github.com/python-poetry/poetry/pull/5841
- https://github.com/python-poetry/poetry/pull/5935

# General Notes
## PEP 610 Changes

[...]

PR #5980, opened 2022-07-09
remove circular dependency on export plugin

dimbleby:
The circular dependency has been bothering me for a while, but https://github.com/python-poetry/poetry/issues/6441 was the nudge to do something about it.

poetry should not depend on the plugin: the plugin should depend on poetry.

This means that docs that talk about `poetry export` arguably don't belong here at all: but mostly they already say "this is actually provided by poetry-plugin-export" so I've just beefed that up slightly.

I'm about to find out what I've broken in the github workflows by removing this dependency...


finswimmer:
I'm courious how this should work. IIRC removing the plugin essentially means removing the export command from the default installation. This is something we really like to do in the future, but for now this would be a breaking change.

dimbleby:
Sure, people who want `poetry export` will need to install it themselves.

If we want `poetry export` to be a core command plugin: then move it back in-tree in this repository.  But if we want it to be a plugin, then let's treat it as a plugin.

dimbleby:
I can see we're going to need consensus here... somehow.

While we're thinking about disentangling these two projects, here are some more thoughts.

I'd be inclined to transfer the bits of export-specific code that linger in this repository over to the plugin entirely.  That is: `get_project_dependency_packages()`, and its children in the call graph.

That would allow deleting the horrid cross-repository github workflow, making it much easier to make fixes to the export function.

It would also make it easier to add enhancements if the export plugin was more involved in walking the dependency tree eg https://github.com/python-poetry/poetry-plugin-export/issues/91 would like access to information that is known inside that code.

To be clear, I don't intend to do any of that any time soon.  But this seems as good a place as any to dump some notes.

dimbleby:
It's probably desirable to take a view on what the Right Thing To Do is before 1.2.0.  

That doesn't necessarily require that anything change - I'd hate to block 1.2.0! - but if we're backing ourselves into a corner where we're stuck with an unwanted circular dependency that's hard to get rid of: it would be as well to do that with eyes wide open.

muhammedseydali (dismissed):

maresb:
@drhagen above:

> I only want Poetry 1.2 to be able to be installable with conda.

@neersighted in https://github.com/python-poetry/poetry/issues/5586#issuecomment-1231945773:

> [The circular dependency issue] is known and we do not plan to address it as part of 1.2.

These two stances seem incompatible.

neersighted:
Yes, this PR is not scoped for 1.2. @abn has a plan to migrate away from this circular dependency, re-using much of @dimbleby's work. However, it's simply too big a change for 1.2, both in terms of adding additional risk, and in terms of @abn's present bandwidth.

xylar:
@neersighted, hmm.  That will potentially leave a lot of packages uninstallable on conda-forge.  One option is that we can patch on conda-forge to remove `poetry-plugin-export` from `poetry` to avoid this issue until it gets resolved in a longer-term way.  This is what I've done in https://github.com/conda-forge/poetry-feedstock/pull/71

joshgc:
Is this PR still a possibility? I'm also having issues building poetry now, but with bazel instead of conda.

radoering:
> Is this PR still a possibility?

Short-term: no. Mid-term: maybe. See #8562

johnthagen:
Would it be possible to remove the circular dependency but still install the plugin by default? Being able to have Poetry interact with `pip` through `requirements.txt` files is still a really nice default feature to have. For example, tools like `nox-poetry` depend on this:

- https://github.com/cjolowicz/nox-poetry/blob/0612f2fabad3f3a3e661db5e0f254000341e92f1/src/nox_poetry/poetry.py#L135

Not having to install a plugin for this kind of functionality (especially since it's a different workflow based on if you are using `pipx` or not, warnings about Windows, etc) would be convenient if at all possible.

johnthagen:
Idea: Could the export plugin be an `extras`? So that a user could install it with `pipx` in a single convenient command with:

```
pipx install poetry[export-plugin]
```

radoering:
I don't think an extra changes much. We still have a circular dependency (only if the extra is used but nevertheless...). I think a general solution for plugins that are required for a specific project as demanded in #5729 might make more sense. It looks like there is already a POC for project plugins: #5740. Maybe, someone picks this up.

johnthagen:
Okay, wow yes #5729 / #5740 would be a really great way to solve this issue! The specific challenge that this PR will introduce for us is a bootstrapping problem. Whenever we share a Poetry project with another developer, they already need to:

1. Install `pipx`
2. Install Poetry
3. Use the project (Poetry/nox etc.)

Now we'd need to inform them that they need to:

1. Install `pipx`
2. Install Poetry
3. Install the `export` plugin
4. Use the project (Poetry/nox etc)

We've found that these bootstrapping type steps are fraught with developers making small mistakes and thus, we actually avoid Poetry plugins more than we otherwise would because getting them installed everywhere correctly is tricky to document and get everyone to manually install.

Hope this feedback is useful!

xylar:
I'm still in favor of eliminating this circular dependency but I just wanted to point to my comment here:
https://github.com/python-poetry/poetry/issues/6441#issuecomment-2058832773
It seems we *can* handle circular dependencies in conda packages, it's just a little awkward.

radoering:
After having merged #9547, I think there is consensus that we can move on here.

docs/cli.md
```diff
@@ -774,17 +774,9 @@ This command exports the lock file to other formats.
...
-In order to avoid a breaking change and make your automation forward-compatible,
-please install poetry-plugin-export explicitly.
-See [Using plugins]({{< relref "plugins#using-plugins" >}}) for details on how to install a plugin.
-{{% /warning %}}
-
 {{% note %}}
-This command is also available as a pre-commit hook.
-See [pre-commit hooks]({{< relref "pre-commit-hooks#poetry-export" >}}) for more information.
+This command is provided by the [Export Poetry Plugin](https://github.com/python-poetry/poetry-plugin-export)
+and is also available as a pre-commit hook. See [pre-commit hooks]({{< relref "pre-commit-hooks#poetry-export" >}}) for more information.
```
radoering, line 779: Maybe, we should make this a warning and put it at top of the `export` section.

johnthagen: Suggestion: Since this will be a big change (removing the export plugin by default) perhaps it's worth adding some explicit docs on how to add this back in a project-specific plugin dependency (e.g. #9547)? I suspect this will be a very common need for many users to figure out.

radoering: Good idea. Here is the relevant section in the docs: https://github.com/python-poetry/poetry/blob/b61e6bd974b7c584b12fa43fe983087cb507c94d/docs/plugins.md?plain=1#L258-L270

We can add a link or even a code snippet with `poetry-plugin-export` to be more explicit.

dimbleby:
I've rebased, am happy to accept pull requests on my fork with tweaks / allow maintainers to push / have someone else follow up separately with further change

docs/cli.md
radoering: @johnthagen I added some additional information. Please take a look.

johnthagen: @radoering The new docs in c3d14384ffc4e3cf00c03639ed330ed0952d14af look great.

radoering (approved):

Issue #5852, opened 2022-06-15
Resolver prioritizes dependencies


FredrikWallstrom:
- [x] I have searched the [issues](https://github.com/python-poetry/poetry/issues) of this repo and believe that this is not a duplicate.
- [x] I have searched the [documentation](https://python-poetry.org/docs/) and believe that my question is not covered.

## Feature Request
I have one dependency in my project that acts like a wrapper and bundles several dependencies. I’m continuously releasing new versions of this wrapper. The problem I have is that the sub-dependencies that is installed by the wrapper have strict versions constraints.

Example:

```
Main project:
[tool.poetry.dependencies]
wrapper = ">=2.0.0,<2.100.0"
flake8 = "^3"

wrapper: 
[tool.poetry.dependencies]
flake8 = "3.0.0"
```

Let's say there are 100 available versions of the wrapper package, all of them have flake8 locked to version 3.0.0. The resolver will start resolving from the flake8 dependency because it has [fewest available versions](https://github.com/python-poetry/poetry/blob/3f54fdc96432a7313ea00df52ba6808465519b10/src/poetry/mixology/version_solver.py#L378). It will start with the latest flake8 version and check for a solution, it wont find a solution because the requirement is not fulfilled by any version of the wrapper. The resolver will continue until it reaches flake8 version 3.0.0.

This is of course a stupid example but imagine what will happen with the resolve time if the bundler bundles more dependencies and the main project depends on additional dependencies where sub-dependencies conflicts with wrappers' dependencies.

In this example and also in my real project, it would be beneficial if the resolver started to resolve from the wrapper dependency instead of picking the dependency with the fewest available versions. I know it is not recommended to have strict dependencies like this but restructure my project setup requires too much of an effort.  

I have been digging but I haven’t find a way around this, is there something I can do to reduce the resolve time?
 What about adding a flag where you can tell the resolver to start resolve from a specific dependency, like this?

`wrapper = {"version" = ">=2.0.0,<2.100.0", prioritize = true}`

dimbleby:
#5217 was similar, in that one many hundreds of versions of awscli were incompatible with recent pyyaml: so it was unfortunate that poetry tried to fix pyyaml first.

Best advice I have is to restrict the space that you have to search: eg prefer `wrapper = ">= 2.100.0"`. Or explicitly require lower flake8 from your main project - or make `wrapper` compatible with recent flake8!

[...]

PR #6011, opened 2022-07-14
feat: add support to specify solver's resolve order

FredrikWallstrom:
# Pull Request Check List

Resolves: https://github.com/python-poetry/poetry/issues/5852
Requires: https://github.com/python-poetry/poetry-core/pull/416

### DIFF
--- a/docs/dependency-specification.md
+++ b/docs/dependency-specification.md
@@ -195,6 +195,18 @@ These activate extra defined for the dependency, to configure an optional depend
 for extras in your project refer to [`extras`]({{< relref "pyproject#extras" >}}).
 {{% /note %}}
 
+## Dependency `resolve-order`
+
+You can specify the solver's resolve order as shown here.
+
+```toml
+[tool.poetry.dependencies]
+gunicorn = { version = "^20.1", resolve-order = 1 }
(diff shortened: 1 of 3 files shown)


neersighted:
I have to admit, I am incredibly dubious on the value here, as this seems like an incredible footgun that only improve some pathological cases. Many of those pathological cases will be self-inflicted, and the real solution is to tighten constraints to reduce the search area. Better tooling for visualizing the search area and targeting packages for constraint adjustment is probably the best long term solution here (as the current state of the art is Poetry veterans eyeballing things based on PyPI releases).

The integer appears to be entirely a lie, as it does not influence solver order beyond the starting node. As `min(resolve_order)` is essentially what happens, this could just be more honest as a boolean (and the docs don't make clear what the intention of an integer is, further supporting this conclusion).

I think ultimately a bit of the X-Y problem is happening here -- I think the real request is for better docs and tooling to understand the solver search space, and manipulate it to get the desired end result (of versions + reduced resolve time).

FredrikWallstrom:
I'll see what you mean. Doesn't feel like none of you are convinced about this, so I will close it. Thank you for taking your time.

PR #6016, opened 2022-07-14
disambiguate direct-origin dependencies in show-outdated

dimbleby:
if you have, say:
```toml
[tool.poetry.dependencies]
poetry = "^1.2.0b3"

poetry = {git = "https://github.com/python-poetry/poetry.git"}
```
and go `poetry show --outdated` then you hit the runtime error `Unknown direct dependency type None`.

The reason is that poetry appears twice, once as a direct dependency and once not.  And we happen to search for the direct dependency using the one that isn't, if you see what I mean.

This MR also improves the error message by saying which package is the problematic one.

### DIFF
--- a/src/poetry/console/commands/show.py
+++ b/src/poetry/console/commands/show.py
@@ -505,7 +505,7 @@ lists all packages available."""
             requires = root.all_requires
 
             for dep in requires:
-                if dep.name == package.name:
+                if dep.name == package.name and dep.source_type == package.source_type:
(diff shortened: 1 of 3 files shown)


mkniewallner (commented):
Since we have a way to reproduce the problem, should we add a non-regression test?

dimbleby:
sure, go for it

neersighted (approved):

PR #6021, opened 2022-07-16
Misc fixes

dimbleby:
A series of commits tightening up the type-checking in various places

### DIFF
--- a/src/poetry/console/commands/cache/clear.py
+++ b/src/poetry/console/commands/cache/clear.py
@@ -46,7 +46,7 @@ class CacheClearCommand(Command):
                     f"Add the --all option if you want to clear all {parts[0]} caches"
                 )
 
-            if not os.path.exists(str(cache_dir)):
+            if not os.path.exists(cache_dir):
                 self.line(f"No cache entries for {parts[0]}")
                 return 0
 
--- a/src/poetry/utils/authenticator.py
+++ b/src/poetry/utils/authenticator.py
@@ -131,18 +131,10 @@ class Authenticator:
             self._get_repository_config_for_url
         )
 
-    @property
-    def cache(self) -> FileCache | None:
-        return self._cache_control
-
-    @property
-    def is_cached(self) -> bool:
-        return self._cache_control is not None
-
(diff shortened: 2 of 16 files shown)


src/poetry/console/commands/cache/clear.py
```diff
@@ -46,7 +46,7 @@ def handle(self) -> int:
                     f"Add the --all option if you want to clear all {parts[0]} caches"
                 )
 
-            if not os.path.exists(str(cache_dir)):
+            if not os.path.exists(cache_dir):
```
Secrus, line 49: Why not use `pathlib.Path.exists()`? Seems cleaner.

src/poetry/utils/authenticator.py
```diff
@@ -202,18 +194,17 @@ def request(
...
+        proxies: dict[str, str] = kwargs.get("proxies", {})
+        stream: bool | None = kwargs.get("stream")
 
         certs = self.get_certs_for_url(url)
-        verify = kwargs.get("verify") or certs.cert or certs.verify
-        cert = kwargs.get("cert") or certs.client_cert
+        verify: bool | str | Path = kwargs.get("verify") or certs.cert or certs.verify
+        verify = str(verify) if isinstance(verify, Path) else verify
 
+        cert: str | Path | None = kwargs.get("cert") or certs.client_cert
```
Secrus, line 204:
```suggestion
        cert: str | Path | None = kwargs.get("cert", certs.client_cert)
```
Isn't this the same?

Secrus (changes requested):

src/poetry/utils/authenticator.py
dimbleby: not _quite_ the same, if `cert` is somehow an empty string...

Secrus: Oh, right, didn't think about this case.

Secrus (approved):
LGTM, but let's get a second opinion

dimbleby:
are we waiting for a second opinion from anyone in particular?  do they know that?

branchvincent (approved):

PR #6022, opened 2022-07-17
use packaging.utils.canonicalize_name()

dimbleby:
`poetry` currently has two different versions of `canonicalize_name()` floating around, which is slightly absurd - and what's worse, they don't even agree.

Since https://github.com/python-poetry/poetry-core/pull/328, `poetry-core` has an implementation that agrees with the implementation in `packaging` - but `poetry` has its own implementation that continues to disagree with that.  Sometimes `poetry` uses one of these, and sometimes the other.

Let's treat the version from `packaging` as the canonical `canonicalize_name()`, and get rid of _both_ of the re-implementations.

I don't know that this fixes - or breaks! - anything in particular; the unit tests don't detect a difference.  But two differing implementations is surely an accident waiting to happen.

Aside from being one-true-implementation, a potential advantage of the `packaging` version is that it returns a `NormalizedName`: which is a newtype around string, rather than a plain string.  That opens the door to future work in which the typechecker can verify that places which expect canonicalized names do indeed get them.

(Will follow up in `poetry-core`, but this one will need merging first since `poetry` sometimes uses the `poetry-core` version).

### DIFF
--- a/src/poetry/config/config.py
+++ b/src/poetry/config/config.py
@@ -10,8 +10,8 @@ from pathlib import Path
 from typing import TYPE_CHECKING
 from typing import Any
 
+from packaging.utils import canonicalize_name
 from poetry.core.toml import TOMLFile
-from poetry.core.utils.helpers import canonicalize_name
 
 from poetry.config.dict_config_source import DictConfigSource
 from poetry.config.file_config_source import FileConfigSource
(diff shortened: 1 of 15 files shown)


dimbleby:
> I wonder if it makes sense to re-export the method in poetry-core? 

This seems excessive to me.  Any change to this method in `packaging` would be enormously breaking and not PEP-503 compliant.  

This method hasn't been changed (bar type annotations) since introduction at https://github.com/pypa/packaging/pull/60, and I don't expect that it ever will be.

Issue #6054, opened 2022-07-23
poetry 1.2.0b3 lock fails with no output


vstrimaitis:
- [ ] I am on the [latest](https://github.com/python-poetry/poetry/releases/latest) Poetry version.
- [x] I have searched the [issues](https://github.com/python-poetry/poetry/issues) of this repo and believe that this is not a duplicate.
- [x] If an exception occurs when executing a command, I executed it again in debug mode (`-vvv` option).

- **OS version and name**: macOS Monterey 12.4
- **Poetry version**: 1.2.0b3
- **Link of a [Gist](https://gist.github.com/) with the contents of your pyproject.toml file**: [click](https://gist.github.com/vstrimaitis/f9511a4b43a771b6bd08e3adc8f5defe)

## Issue

I've discovered a case where the latest pre-release version of Poetry (1.2.0b3) fails to lock dependencies in some cases. Unfortunately, I haven't been able to track down why exactly this happens, but I do have a concrete example (linked above).

Steps to reproduce:

- Install Poetry 1.2.0b3 (I installed it with `pipx install --suffix @1.2.0b3 'poetry==1.2.0b3'` to distinguish from my other installations)
- Create a `pyproject.toml` with `kfp = "1.8.12"` and `gsutil = "^5.6.0"`
- Run `poetry@1.2.0b3 lock`

The output I get is:

```
$ poetry@1.2.0b3 lock
Updating dependencies
Resolving dependencies... (0.1s)



$
```

Running the same command with `-vvv`:

```
$ poetry@1.2.0b3 lock -vvv
Loading configuration file /Users/vstrimaitis/Library/Preferences/pypoetry/config.toml
Using virtualenv: /Users/vstrimaitis/dev/random/reproduce-poetry-lock-bug/.venv
Project environment contains an empty path in sys_path, ignoring.
Updating dependencies
Resolving dependencies...
   1: fact: reproduce-poetry-lock-bug is 0.1.0
   1: derived: reproduce-poetry-lock-bug
   1: fact: reproduce-poetry-lock-bug depends on kfp (1.8.12)
   1: fact: reproduce-poetry-lock-bug depends on gsutil (^5.6.0)
   1: selecting reproduce-poetry-lock-bug (0.1.0)
   1: derived: gsutil (>=5.6.0,<6.0.0)
   1: derived: kfp (==1.8.12)
[keyring.backend] Loading KWallet
[keyring.backend] Loading SecretService
[keyring.backend] Loading Windows
[...]

PR #6058, opened 2022-07-24
revert #5770, provide new fix

dimbleby:
# Pull Request Check List

Resolves: #6054

I think that the fix made at #5770 was not good.  (I did review it at the time without seeing problems, so no blame!)

As I now understand it, the problem that it was trying to solve was:
- we register a direct-origin assignment for `foo`
- we then register a regular assignment for `foo[extras]`
- we'd like for the result to be that all `foo`-related assignments find the direct-origin package
- but that's not what happens

#5770 fixed this by finding the `foo` assignment at the second bullet, and then the intersection of a direct-origin and not-direct-origin term always favours the direct-origin.

I think this is probably unreliable: eg in the first bullet we might have registered a direct-origin assignment for `foo[other-extras]`.  Or bullets 1 and 2 might happen the other way round.

Apart from that, #6054 exposed a problem where we registered regular assignments for `foo` and `foo[extras]`, but with incompatible version ranges.  Before #5770 that was fine and is unwound later in the search, but after #5770 we try to update the `foo` assignment and go wrong.

This MR undoes the fix of #5770 and tries another.  I have arranged that:
- we always try to resolve direct-origin dependencies first
- the provider remembers direct-origin dependencies and returns them again if asked for the same package but with different features

That fixes the original #5311 because now the direct-origin version of the `foo` package is the only one that we ever see; and, per the new testcase, unbreaks #6054

### DIFF
--- a/src/poetry/puzzle/provider.py
+++ b/src/poetry/puzzle/provider.py
@@ -140,6 +140,7 @@ class Provider:
         self._load_deferred = True
         self._source_root: Path | None = None
         self._installed_packages = installed if installed is not None else []
+        self._direct_origin_packages: dict[str, Package] = {}
 
     @property
     def pool(self) -> Pool:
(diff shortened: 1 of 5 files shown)


src/poetry/puzzle/provider.py
```diff
@@ -268,18 +269,33 @@ def search_for(self, dependency: Dependency) -> list[DependencyPackage]:
...
+        #
+        # We rely on the VersionSolver resolving direct-origin dependencies first.
+        direct_origin_package = self._direct_origin_packages.get(dependency.name)
+        if direct_origin_package is not None:
+            package = direct_origin_package.with_features(dependency.features)
+            packages = (
+                [package]
+                if package.satisfies(dependency, ignore_source_type=True)
+                else []
             )
```
radoering, line 287:
```suggestion
            packages = (
                [direct_origin_package]
                if dependency.constraint.allows(direct_origin_package.version)
                else []
            )
```

If I am right, `search_for` did always return packages without features/extras so far and `complete_package` added these. Maybe, we should keep this pattern?

radoering (commented):
I agree that resolving direct origin dependencies first seems to be sensible. I just noticed a minor inconsistency (see comment below).

src/poetry/puzzle/provider.py
dimbleby: I'm ok with that

radoering:
I just added some dedicated tests for `search_for` (some general tests and some related to this change).

radoering (approved):

PR #6059, opened 2022-07-24
fix: debug message while adding packages

m9810223:
# Pull Request Check List

Resolves: #issue-number-here


neersighted (commented):
Hi @m9810223 -- it's unclear what this change is intended to solve/fix. If you think adding a colon to the existing logging messages significantly clarifies them, could you explain why?

PR #6062, opened 2022-07-24
allow use of not-embedded pip

### LINKED ISSUE #6060: using embedded pip is slow
Related to #6047 

Installing a large number of packages with the embedded pip takes noticeably longer than using a more regular pip - I see the install stage go from ~90s to ~60s in a particular project, if I tweak poetry so that it doesn't use embedded pip.  I suppose that reaching inside a wheel and executing pip from there is more expensive.

Probably there are reasons for choosing embedded pip - presumably it leaves poetry less exposed to whatever foolishness the user has managed to achieve with their own pip.  And perhaps not every environment has a non-embedded pip in it.

On the other hand this is a noticeable performance penalty.  As someone who doesn't habitually mess up their environment, I'd be happy for poetry to use the pip that is already in it.

Perhaps it would be reasonable to use the embedded pip only as a last resort, if no other was available?

This changed at #4011, where the implementation of `run_pip()` gained an `embedded=True` parameter.

dimbleby:
Fixes #6060 

The code still falls back to embedded pip if no other pip can be found - see the implementation of `Env.pip()`.

(Tests change because we save ourselves a call to `get_embedded_wheel()` which causes a python script to be run to get environment info.)

I realise that the discussion in #6060 hasn't yet reached a conclusion but this is trivial enough that we might as well have it available.

### DIFF
--- a/src/poetry/utils/env.py
+++ b/src/poetry/utils/env.py
@@ -1425,7 +1425,7 @@ class Env:
         return self._run(cmd, **kwargs)
 
     def run_pip(self, *args: str, **kwargs: Any) -> int | str:
-        pip = self.get_pip_command(embedded=True)
+        pip = self.get_pip_command()
         cmd = pip + list(args)
         return self._run(cmd, **kwargs)
 
--- a/tests/inspection/test_info.py
+++ b/tests/inspection/test_info.py
@@ -226,7 +226,7 @@ def test_info_setup_missing_mandatory_should_trigger_pep517(
     except PackageInfoError:
         assert spy.call_count == 3
     else:
-        assert spy.call_count == 2
+        assert spy.call_count == 1
 
 
 def test_info_prefer_poetry_config_over_egg_info():


dimbleby:
seeing #6205 reminded me of this one.  

Certainly this doesn't bring such large gains as are claimed there but I reckon it does take about 50% off a typical "poetry install" (with warm caches, downloaded artifacts).

neersighted (approved):
I'm fine with this, but I'd like to hear from a few others.

radoering:
I thought, it might make a difference when updating pip as a locked dependency (at least on Windows). But it seems it makes no difference if the embedded pip or the pip inside the venv is used. I always get the following error:
```
ERROR: To modify pip, please run the following command: ...\python.exe -m pip install ...
```

So I'm fine with it, too.

brandon-leapyear:
Note: it seems like this PR fixed a bug in 1.2.0 where an upgrade to `virtualenv` would fail `poetry self update` when trying to call the embedded pip (`virtualenv` just released a new version today, which is causing the issue).

Repro:
```bash
docker run --rm -it python:3.10 bash

# in docker
python3 -m venv venv
venv/bin/pip install poetry==1.2.0 virtualenv==20.16.5
venv/bin/poetry self update
```
Fails with:
```
  Command ['/venv/bin/python3', '/venv/lib/python3.10/site-packages/virtualenv/seed/wheels/embed/pip-22.2.2-py3-none-any.whl/pip', 'install', '--disable-pip-version-check', '--prefix', '/venv', '--upgrade', '--no-deps', '/root/.cache/pypoetry/artifacts/8a/85/69/120758446f5ee47cfa1c7c237156c8eaf982af2f9f5778dd1aa1395273/poetry-1.2.2-py3-none-any.whl'] errored with the following return code 2, and output: 
  /venv/bin/python3: can't open file '/venv/lib/python3.10/site-packages/virtualenv/seed/wheels/embed/pip-22.2.2-py3-none-any.whl/pip': [Errno 2] No such file or directory
```

Opening an issue doesn't seem appropriate since this is now fixed, but maybe a test should be added to ensure this doesn't regress?

PR #6083, opened 2022-07-27
more typechecking

dimbleby:
Fixing most, though not quite all, of the mypy errors that would be revealed if cleo had a `py.typed`.

### DIFF
--- a/src/poetry/console/application.py
+++ b/src/poetry/console/application.py
@@ -38,13 +38,13 @@ if TYPE_CHECKING:
     from poetry.poetry import Poetry
 
 
-def load_command(name: str) -> Callable[[], type[Command]]:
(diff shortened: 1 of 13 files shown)


Secrus:
@dimbleby Cleo could use your proficiency with types. Right now I am working on it mostly alone, could use a hand.

dimbleby:
@Secrus I may be about to offer a small MR to help make poetry happier with cleo's types but, on the whole, I am against doing work in cleo.

Not looking to repeat that discussion, but my view remains as at https://github.com/python-poetry/poetry/pull/5756#issuecomment-1145032810

niharikavb (approved):

neersighted (approved):

PR #6084, opened 2022-07-27
build: refresh lockfile to remove unused dependencies and vulnerable version of `cryptography`

md384:
# Pull Request Check List

Resolves: NA

The lockfile for the `1.1.X` versions of poetry contains a vulnerable version of the `cryptography` package. See
https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2020-36242 and 
https://nvd.nist.gov/vuln/detail/CVE-2020-36242.

This PR is the changes after running ~`poetry lock --no-update`~ `poetry update cryptography`. It is unclear why `poetry lock --no-update` did not work - perhaps an existing bug with poetry 1.1.X.

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -51,7 +51,7 @@ redis = ["redis (>=2.10.5)"]
 
 [[package]]
 name = "cachecontrol"
-version = "0.12.10"
+version = "0.12.11"
 description = "httplib2 caching for requests"
 category = "main"
 optional = false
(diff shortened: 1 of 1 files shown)


md384:
The patched version of cryptography only supports 2.7 and >=3.6 and all versions after the patched version only support >=3.6 so there is no possible patched version that we can install and still support py3.5. I will close this PR.

PR #6085, opened 2022-07-28
Turn spaces in project name to underscore.

Ninzero:
# Pull Request Check List

Currently, if the project name you typed contains whitespaces, the package name will also contain whitespaces. For example:

```
poetry new "New Project"
```

will cause test file contain:

```
from new project import __version__
```

which will cause error. In this case, we should automatically change project name to `new_project`.

I know that set `--name` argument can solve this problem. But I think maybe this process can be automated.

I have not include a tests for this PR. Mainly because I have not figured out a suitable testing method, also because this is a draft PR. If anyone have good idea on how to test, please help, thanks.

No documentation change required for this PR.

### DIFF
--- a/src/poetry/console/commands/new.py
+++ b/src/poetry/console/commands/new.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
+import re
 import sys
 
 from contextlib import suppress
@@ -48,6 +49,7 @@ class NewCommand(Command):
         name = self.option("name")
         if not name:
             name = path.name
+        name = re.sub(r"\s+", "_", name)
 
         if path.exists() and list(path.glob("*")):
             # Directory is not empty. Aborting.


neersighted:
This change as-is is much too specific and hardcoded to the particular issue you sought to solve. We should likely normalize all project names during `poetry new` (e.g. see the `canonicalize_name()` function in `packaging.utils` that was most recently discussed in https://github.com/python-poetry/poetry/pull/6022), or reject them outright if they don't meet validity criteria.

neersighted (changes requested):
This change as-is is much too specific and hardcoded to the particular issue you sought to solve. We should likely normalize all project names during `poetry new` (e.g. see the `canonicalize_name()` function in `packaging.utils` that was most recently discussed in https://github.com/python-poetry/poetry/pull/6022), or reject them outright if they don't meet validity criteria.

Ninzero:
I see. Since [PEP 426](https://peps.python.org/pep-0426/#name) only indicated which characters are allowed in package names. Maybe we should reject all package name with any invalid characters in it (by throwing an error)?

neersighted:
Closing as this is not mergable; a check for invalid characters in a pretty name (`name =` is a pretty name and not a canonicalized name) would likely be welcome, however.

PR #6095, opened 2022-07-29
Update `flake8` additional dependencies

mkniewallner:
# Pull Request Check List

Update `flake8` additional dependencies and update code to comply with:
- https://github.com/PyCQA/flake8-bugbear/releases/tag/22.7.1 that added late binding check as `B023`
- https://github.com/snok/flake8-type-checking/releases/tag/v2.0.0 that now uses `TC002` for third-party imports, and `TC003` for built-in imports

### DIFF
--- a/src/poetry/plugins/application_plugin.py
+++ b/src/poetry/plugins/application_plugin.py
@@ -24,4 +24,6 @@ class ApplicationPlugin(BasePlugin):
     def activate(self, application: Application) -> None:
         for command in self.commands:
             assert command.name is not None
-            application.command_loader.register_factory(command.name, lambda: command())
+            application.command_loader.register_factory(
+                command.name, lambda: command()  # noqa: B023
+            )
--- a/src/poetry/puzzle/provider.py
+++ b/src/poetry/puzzle/provider.py
@@ -670,10 +670,10 @@ class Provider:
             #   - {<Package foo (1.2.3): {"bar": <Dependency bar (<2.0)>}
 
             def fmt_warning(d: Dependency) -> str:
-                marker = d.marker if not d.marker.is_any() else "*"
+                dependency_marker = d.marker if not d.marker.is_any() else "*"
                 return (
                     f"<c1>{d.name}</c1> <fg=default>(<c2>{d.pretty_constraint}</c2>)</>"
-                    f" with markers <b>{marker}</b>"
+                    f" with markers <b>{dependency_marker}</b>"
                 )
 
             warnings = ", ".join(fmt_warning(d) for d in deps[:-1])
--- a/tests/compat.py
+++ b/tests/compat.py
@@ -4,7 +4,7 @@ from __future__ import annotations
 try:
     import zipp  # nopycln: import
 except ImportError:
-    import zipfile as zipp  # noqa: F401, TC002
+    import zipfile as zipp  # noqa: F401, TC003
 
 try:
     from typing import Protocol  # nopycln: import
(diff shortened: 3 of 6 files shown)


src/poetry/puzzle/provider.py
```diff
@@ -670,10 +670,9 @@ def complete_package(
             #   - {<Package foo (1.2.3): {"bar": <Dependency bar (<2.0)>}
 
             def fmt_warning(d: Dependency) -> str:
-                marker = d.marker if not d.marker.is_any() else "*"
                 return (
                     f"<c1>{d.name}</c1> <fg=default>(<c2>{d.pretty_constraint}</c2>)</>"
-                    f" with markers <b>{marker}</b>"
+                    f" with markers <b>{'*' if d.marker.is_any() else d.marker}</b>"
```
mkniewallner, line 675: While this change is not strictly needed, it raised a `B023` as well.

As `marker` is also defined outside of `fmt_warning`, I believe it confused the plugin (because renaming `marker` to `dependency_marker` made it pass).

Since the variable is only used once, I though it wasn't worth having one anyway, but if we prefer the current version, it's fine by me to just rename `marker` to `dependency_marker` instead.

src/poetry/plugins/application_plugin.py
```diff
@@ -24,4 +24,6 @@ def commands(self) -> list[type[Command]]:
     def activate(self, application: Application) -> None:
         for command in self.commands:
             assert command.name is not None
-            application.command_loader.register_factory(command.name, lambda: command())
+            application.command_loader.register_factory(
+                command.name, lambda: command()  # noqa: B023
```
mkniewallner, line 28: This one can, I think, be considered as a false positive, since we do want late binding here.

src/poetry/puzzle/provider.py
Secrus: I think it's good the way it is

tests/compat.py
```diff
@@ -4,7 +4,7 @@
 try:
     import zipp  # nopycln: import
```
Secrus, line 5: While you are at it, you could rewrite it to `if sys.version < (3, x)`, that way it will be recognized by `pyupgrade` ([link](https://github.com/asottile/pyupgrade#python2-and-old-python3x-blocks)) and will be removed when we upgrade Python versions.

Secrus (changes requested):

src/poetry/puzzle/provider.py
mkniewallner: Alright, squashed into https://github.com/python-poetry/poetry/pull/6095/commits/a84feef0fd66ec25f3c0f4ef9a5c738d39bbf83a by only renaming `marker` to `dependency_marker`.

tests/compat.py
mkniewallner: Nice, I didn't know that `pyupgrade` handled that, thanks!

tests/compat.py
```diff
@@ -1,10 +1,12 @@
 from __future__ import annotations
 
+import sys
 
-try:
-    import zipp  # nopycln: import
-except ImportError:
-    import zipfile as zipp  # noqa: F401, TC002
+
+if sys.version_info < (3, 8):
```
mkniewallner, line 6: Based on https://github.com/python-poetry/poetry/blob/c8ae868472d9e6e4fc281adc08f2d14e374d35c8/pyproject.toml#L84

Took the opportunity to export as `zipfile` rather than `zipp` as well, since `zipfile` is the one in stdlib, and `zipp` will most likely go away when we will remove support for `3.7`.

Secrus (approved):

PR #6097, opened 2022-07-30
Remove dead code

dimbleby:
I thought I'd have a go with https://pypi.org/project/vulture/

It reports rather a lot of false positives and I doubt that it's going to be worth the trouble of maintaining an allowlist and running this in pipelines: but as a one-off thing I reckon it did find a fair amount that can sensibly be removed - per this MR.

Less code = less bugs!


ngnpope:
This might also be useful: https://github.com/asottile/dead

neersighted:
I'm dubious on all code removed here being dead (e.g. I used some of the removed `Env` methods in #6458), but I'm all for trimming out cruft... I think in the medium term we need to start marking methods as internal or external more rigorously (such as in docstrings since `_` prefixing is really not an indicator of private or public API status), or simply defined a `poetry.api` module that re-exports everything we intend to let outside consumers rely on.

dimbleby:
> I'm dubious on all code removed here being dead

18 months later it all still is unused

Secrus (approved):

Secrus:
@dimbleby if you could rebase the changes with the latest `main`, I will be happy to merge this

Issue #6109, opened 2022-08-03
Windows drive mappings break virtual environment name


adamgreg:
- [x] I am on the [latest](https://github.com/python-poetry/poetry/releases/latest) Poetry version.
- [x] I have searched the [issues](https://github.com/python-poetry/poetry/issues) of this repo and believe that this is not a duplicate.
- [x] If an exception occurs when executing a command, I executed it again in debug mode (`-vvv` option).

- **OS version and name**: Windows 10 x64
- **Poetry version**: Reproduced in 1.1.14 and 1.2.0b3

## Issue
Windows drive mappings are not resolved when hashing the current working directory to generate a virtual environment name. 

This means that a virtual environment that was created when working under a mapped drive is not re-used when working under the real un-mapped directory.

I'll explain the problem this has caused me. I work under a mapped drive and installed git pre-commit hooks that run commands such as `pylint` from the development environment using `poetry run`. When the hooks were automatically run by git, the current working directory was always the "real" directory. Poetry would automatically create and use a new empty environment that did not contain the necessary executables, leading to very confusing errors.

It can be fixed with a single call to `os.path.realpath()` in `EnvManager.generate_env_name()`. I will create a pull request.

I've checked in Linux, and could not reproduce the same problem using symbolic links.

Example:

```powershell
cd c:\parent\my_project
poetry install
poetry env list  # Poetry-managed virtual environment listed
subst z: c:\parent
cd z:\my_project
poetry env list  # No environments listed. "poetry run" commands will fail. 
```

PR #6110, opened 2022-08-03
Resolve cwd using os.path.realpath() when generating environment names

adamgreg:
# Pull Request Check List

Resolves: #6109

### DIFF
--- a/src/poetry/utils/env.py
+++ b/src/poetry/utils/env.py
@@ -1145,7 +1145,7 @@ class EnvManager:
     def generate_env_name(cls, name: str, cwd: str) -> str:
         name = name.lower()
         sanitized_name = re.sub(r'[ $`!*@"\\\r\n\t]', "_", name)[:42]
-        normalized_cwd = os.path.normcase(cwd)
+        normalized_cwd = os.path.realpath(os.path.normcase(cwd))
         h_bytes = hashlib.sha256(encode(normalized_cwd)).digest()
         h_str = base64.urlsafe_b64encode(h_bytes).decode()[:8]
 
--- a/src/poetry/utils/env.py
+++ b/src/poetry/utils/env.py
@@ -1145,7 +1145,7 @@ class EnvManager:
     def generate_env_name(cls, name: str, cwd: str) -> str:
         name = name.lower()
         sanitized_name = re.sub(r'[ $`!*@"\\\r\n\t]', "_", name)[:42]
-        normalized_cwd = os.path.realpath(os.path.normcase(cwd))
+        normalized_cwd = os.path.normcase(os.path.realpath(cwd))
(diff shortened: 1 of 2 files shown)


adamgreg:
I wasn't sure how to add a suitable new unit test, sorry.

src/poetry/utils/env.py
```diff
@@ -1145,7 +1145,7 @@ def get_base_prefix(cls) -> Path:
     def generate_env_name(cls, name: str, cwd: str) -> str:
         name = name.lower()
         sanitized_name = re.sub(r'[ $`!*@"\\\r\n\t]', "_", name)[:42]
-        normalized_cwd = os.path.normcase(cwd)
+        normalized_cwd = os.path.realpath(os.path.normcase(cwd))
```
radoering, line 1148:
```suggestion
        normalized_cwd = os.path.normcase(os.path.realpath(cwd))
```

It should be the other way around to ensure that case is normalized.

radoering (changes requested):

radoering:
> I wasn't sure how to add a suitable new unit test, sorry.

You probably have to mock `realpath()`.

src/poetry/utils/env.py
adamgreg: Good catch, thanks.

I suspect it's a bit redundant, because `realpath()` does seem to return the canonical case for the path (on NTFS at least), with an uppercase drive letter, so it should already be consistent. Calling `normcase()` last should keep the majority of environment names unchanged by this PR,  though, so worth doing.

adamgreg:
> > I wasn't sure how to add a suitable new unit test, sorry.
> 
> You probably have to mock `realpath()`.

`realpath()` returns its argument unchanged if the path does not exist, so I don't think it's necessary to mock it unless we want to count the calls.

I meant to say I can't think of a way to unit test that this actually fixes the bug, short of mapping a drive on the test system.

radoering:
Let's say you have `path1` and `path2`. You can mock `realpath` to return `path1` for both (i.e. `path2` is a symlink pointing to `path1`). Then, `generate_env_name()` should return the same value no matter if it's called with `path1` or `path2`. A test case similar to https://github.com/python-poetry/poetry/blob/62701fe89fa444452e0f9d1d77d0b653d67d1f18/tests/utils/test_env.py#L1356-L1362 should be sufficient (just with mocking and without platform distinction).

Alternatively, instead of mocking you could create a symlink similar to https://github.com/python-poetry/poetry/blob/62701fe89fa444452e0f9d1d77d0b653d67d1f18/tests/packages/test_locker.py#L760 although that's probably a bit more overhead.

radoering:
pre-commit.ci autofix

radoering (approved):

PR #6128, opened 2022-08-07
move export code into export plugin

radoering:
follow up of python-poetry/poetry-plugin-export#94

Requires a release of poetry-plugin-export with the mentioned change

### DIFF
--- a/src/poetry/packages/locker.py
+++ b/src/poetry/packages/locker.py
@@ -5,7 +5,6 @@ import logging
 import os
 import re
 
-from copy import deepcopy
 from hashlib import sha256
 from pathlib import Path
 from typing import TYPE_CHECKING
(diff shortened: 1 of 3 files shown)


src/poetry/packages/locker.py
```diff
@@ -206,175 +197,6 @@ def locked_repository(self) -> Repository:
...
-            )
-
-        nested_dependencies = cls.__walk_dependencies(
-            dependencies=project_requires,
-            packages_by_name=packages_by_name,
-        )
-
-        return nested_dependencies.items()
-
-    def get_project_dependency_packages(
```
mkniewallner, line 327: A quick GitHub search shows that this is at least used by one popular project: https://github.com/cloud-custodian/cloud-custodian/blob/318f2b4b1c81b4ffae3474c92038ad328ce00dab/tools/dev/poetrypkg.py#L221-L223.
Not that this holds us from removing this method, but we should probably mention this removal in 1.2 release notes.

mkniewallner (approved):

src/poetry/packages/locker.py
radoering: Good point. I agree we should mention that the functionality has been moved to `poetry-plugin-export`.

I was just thinking about whether we should deprecate the methods and call the functions in the plugin but I don't think that's necessary since we moved the complete export code between 1.1 and 1.2 so that can't be imported as before either. Further, calling export code here is counterproductive for disentangling `poetry` and `poetry-plugin-export`. Thus, a notice in the changelog must suffice.

PR #6130, opened 2022-08-07
Improve handling of extras

maksbotan:
# Pull Request Check List

This changeset improves caching of dependencies with extras in two ways:

* `foo[extra]` -> `foo` dependency preserves information about `foo`'s source. If it comes from secondary repo, poetry will not look for it on PyPI
* Do not resolve optional dependencies that are not requested: if we depend on `foo[extra1]`, do not look into dependencies of `foo[extra2]`. This will save time when `extra2` contains some heavy-to-resolve dependencies like VCS ones.

I'm all in for adding tests, but I'd need guidance how to proceed.

When this PR is merged, I'll backport it to 1.1 branch as well.

### DIFF
--- a/src/poetry/puzzle/provider.py
+++ b/src/poetry/puzzle/provider.py
@@ -248,8 +248,23 @@ class Provider:
         )
         return packages
 
-    def search_for_direct_origin_dependency(self, dependency: Dependency) -> Package:
+    def _get_from_deferred_cache(self, dependency: Dependency) -> Package | None:
         package = self._deferred_cache.get(dependency)
+        if package is None:
+            return None
+
+        if dependency.is_vcs():
+            dependency._source_reference = package.source_reference
+            dependency._source_resolved_reference = package.source_resolved_reference
+            dependency._source_subdirectory = package.source_subdirectory
+
+        dependency._constraint = package.version
+        dependency._pretty_constraint = package.version.text
+
+        return package
+
+    def search_for_direct_origin_dependency(self, dependency: Dependency) -> Package:
+        package = self._get_from_deferred_cache(dependency)
         if package is not None:
             pass
 
@@ -572,20 +587,7 @@ class Provider:
             dependency = dependency_package.dependency
             requires = package.requires
 
-        if self._load_deferred:
-            # Retrieving constraints for deferred dependencies
-            for r in requires:
-                if r.is_direct_origin():
-                    locked = self.get_locked(r)
-                    # If lock file contains exactly the same URL and reference
-                    # (commit hash) of dependency as is requested,
-                    # do not analyze it again: nothing could have changed.
-                    if locked is not None and locked.package.is_same_package_as(r):
-                        continue
-                    self.search_for_direct_origin_dependency(r)
-
-        optional_dependencies = []
-        _dependencies = []
+        optional_dependencies: list[str] = []
 
         # If some extras/features were required, we need to
         # add a special dependency representing the base package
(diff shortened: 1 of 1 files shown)


maksbotan:
@radoering hi! I've also updated this one, now tests pass.

maksbotan:
I will add a test for first change — check that poetry does not try to query PyPI for locked package with extras and secondary source.

However I'm a bit at lost about the second one. What is the cleanest way to test that change?

LarsDu:
Does this PR resolve the following issue?
https://github.com/python-poetry/poetry/issues/6419

maksbotan:
@LarsDu do you have reproduction steps for that issue? I can check if my branch fixes it.

radoering:
Maybe, we can split this in two PRs? The first change seems quite clear and low risk (and easier to test). The second one has more potential for negative side effects. We should be able to proceed quite fast with the first one but I probably need some time to think about the second one.

maksbotan:
@radoering no problem, I'll split
What are your thoughts about required test cases?

radoering:
I have not yet an idea for the second one. For the first one something like this might be sufficient (`test_provider.py`):

```
@pytest.mark.parametrize("source_name", [None, "repo"])
def test_complete_package_with_extras_preserves_source_name(
    provider: Provider, repository: Repository, source_name: str | None
) -> None:
    package_a = Package("A", "1.0")
    package_b = Package("B", "1.0")
    dep = get_dependency("B", "^1.0", optional=True)
    package_a.add_dependency(dep)
    package_a.extras = {"foo": [dep]}
    repository.add_package(package_a)
    repository.add_package(package_b)

    dependency = Dependency("A", "1.0", extras=["foo"])
    if source_name:
        dependency.source_name = source_name

    complete_package = provider.complete_package(
        DependencyPackage(dependency, package_a)
    )

    requires = complete_package.package.all_requires
    assert len(requires) == 2
    assert requires[0].name == "a"
    assert requires[0].source_name == source_name
    assert requires[1].name == "b"
    assert requires[1].source_name is None

```

maksbotan:
@radoering I've split the first change into #6472, along with your test.

LarsDu:
@maksbotan I'll look into seeing if this branch resolves the issue sometime this week. We are primarily concerned with case 2:

When we have multiple teams working out of the same repo with different extra dependencies, we want to avoid having the solver run across all extras as dependency conflicts can become extremely problematic.

Having different groups of extras will necessitate making unit tests aware of dependencies, but that is a reasonable tradeoff.

@radoering I believe this may cause side-effects as virtually all groups are currently having `poetry` solve across the union of all base + extras dependencies?
Perhaps making this behavior optional would be the way forward?

neersighted:
This PR solves a different issue from what you are interested in @LarsDu -- the idea here is to not solve for extras that are not defined on a dependency of your project.

If your project (the 'root package') defines multiple extras, they must not conflict and the solver will ensure that all extras are installable at the same time. This is currently how Poetry's solver works and you are expressing a feature request found at #6419.

maksbotan:
@radoering I've updated this PR to include only the second one of my changes.

However, my initial pain point is solved by my other PR — the one for reusing locked git repo dependency. Previously poetry used to refetch git repo that is a transitive dependency under extra I don't actually depend on. Now that refetching is gone, this one is not so important for me.

What do you think about the change in general?

radoering:
@maksbotan Finally, I had the time to take a closer look. I think I found a simpler solution to solve this issue. See #6615.

radoering:
First part split out to #6472, second part superseded by #6615.

PR #6131, opened 2022-08-07
Do not fetch VCS deps when reference didn't change

maksbotan:
# Pull Request Check List

If we have `root` -> `foo` -> `VCSDependency(bar, ref="sha256")`, and lock file contains `bar @ sha256` and user ran `poetry lock --no-update`, do not fetch `bar` repo from git: as commit hash is explicitly locked, nothing could have changed.

As with #6130, I'll backport this to 1.1 once merged. And I'll need guidance on adding tests.

### DIFF
--- a/src/poetry/puzzle/provider.py
+++ b/src/poetry/puzzle/provider.py
@@ -581,6 +581,12 @@ class Provider:
             # Retrieving constraints for deferred dependencies
             for r in requires:
                 if r.is_direct_origin():
+                    locked = self.get_locked(r)
+                    # If lock file contains exactly the same URL and reference
+                    # (commit hash) of dependency as is requested,
+                    # do not analyze it again: nothing could have changed.
+                    if locked is not None and locked.package.is_same_package_as(r):
+                        continue
                     self.search_for_direct_origin_dependency(r)
 
         optional_dependencies = []
--- a/tests/puzzle/test_solver.py
+++ b/tests/puzzle/test_solver.py
@@ -3454,6 +3454,63 @@ def test_solver_keeps_multiple_locked_dependencies_for_same_package(
     )
 
 
+@pytest.mark.parametrize("is_locked", [False, True])
+def test_solver_does_not_update_ref_of_locked_vcs_package(
+    package: ProjectPackage,
+    repo: Repository,
+    pool: Pool,
+    io: NullIO,
+    is_locked: bool,
+):
+    locked_ref = "123456"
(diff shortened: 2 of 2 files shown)


radoering (commented):
Although I think the change is functionally fine, I don't like it from an architectural point of view. I have thought about it for some time and finally created #6204. If this refactoring makes it to master, this PR can be adjusted accordingly.

radoering:
Since #6204 has been merged you may rebase this PR. The required change should be much simpler now.

Re test: This change also fixes that the locked hash of a vcs dependency referencing a branch is updated when running `poetry lock --no-update`. I can provide a test for this later this week.

maksbotan:
Thanks! I've folloed #6204, it's great that it's merged.

I'll try to update this in a few days.

maksbotan:
@radoering i've updated my branch to current master, incorporating your refactoring changes. Thanks!

neersighted (dismissed):
LGTM, but tests are needed. The existing tests for the provider should be good guidance @maksbotan -- but if you're still stuck on tests, @radoering are you still volunteering to get this one over the finish line?

maksbotan:
I'd love to write a test, but I need an advice on what actually __test for__. Is there a way to check that `provider` does not actually call anything git-related (perhaps `provider.get_package_from_vcs`)? Maybe it can be done through mocker?

radoering:
I added a test with a slightly different focus. It checks that a locked vcs dependency referencing a branch is not updated when running `poetry lock --no-update`.

@maksbotan You can probably add a similar test with a locked vcs dependency referencing a hash and check via a mock that `get_package_from_vcs` is not called.

@neersighted Do you really want to backport this? Afaik it's not a regression and it would require to backport #6204 or make significant changes in the backport...

maksbotan:
Thanks!
What backport are talking about? I'm not sure I'm following

radoering:
The label `backport/1.2` has been set but I don't think we'll backport it.

maksbotan:
Ah, I see, thanks. Without backport this will only hit 1.3, right?

onerandomusername:
I'd like to see this backported if the 1.3 release is going to be anything like 1.2

neersighted:
That's fair -- I had it in my head this was a regression, but the truth is that 1.1 exhibited this behavior as well.

maksbotan:
@radoering I've added a new test, hope it's ok

radoering (approved):
LGTM 👍

PR #6157, opened 2022-08-12
Fix repository cache inconsistency issue

radoering:
Fixes an inconsistency when caching matches of a `LegacyRepository`. Without the fix `find_packages()` might return an empty list on second call for packages with only pre-releases because only (released) `versions` are cached but not `ignored_pre_release_versions` (which cannot be ignored at all events).

### DIFF
--- a/src/poetry/repositories/legacy_repository.py
+++ b/src/poetry/repositories/legacy_repository.py
@@ -46,7 +46,9 @@ class LegacyRepository(HTTPRepository):
         ignored_pre_release_versions = []
 
         if self._cache.store("matches").has(key):
-            versions = self._cache.store("matches").get(key)
+            versions, ignored_pre_release_versions = self._cache.store("matches").get(
+                key
+            )
         else:
             page = self._get_page(f"/{dependency.name.replace('.', '-')}/")
             if page is None:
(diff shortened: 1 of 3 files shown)


dimbleby:
I fixed the same thing incidentally in #6081; the changes to the legacy repository there could be taken standalone.

Our fixes are quite different: I was mostly trying to make the legacy and pypi repositories more alike and it was almost accidental that I discovered this bug.  The fix here is certainly more targeted, but it does indeed make the repository classes diverge yet further.

I'm pretty suspicious that this cache is more or less worthless anyway - https://github.com/python-poetry/poetry/pull/5868#discussion_r931507885 - probably it could be removed without anyone noticing the difference...

radoering:
I like making `find_packages` of the different repositories more alike. I'd like to extract the changes to `LegacyRepository` from #6081. Do you want to create a PR or are you fine if I adopt them?

There is just one thing I can't assess yet: You removed some `replace('.', '-')`. Can you explain the reason?

> I'm pretty suspicious that this cache is more or less worthless anyway - [#5868 (comment)](https://github.com/python-poetry/poetry/pull/5868#discussion_r931507885) - probably it could be removed without anyone noticing the difference...

Same feeling, but not sure enough...

dimbleby:
I'm more than fine with you adopting those changes, help yourself.

I removed the `replace('.', '-')` stuff because it is made redundant by #6022: canonicalized names have already made this replacement.

radoering:
Preferred alternative PR: #6165

@dimbleby: Please take a look.

radoering:
Superseded by #6165

Issue #6144, opened 2022-08-09
Poetry relies on tomlkit `0.11.{2/3}` which has a memory leak


TNonet:
- [x] I am on the [latest](https://github.com/python-poetry/poetry/releases/latest) Poetry version.
- [x] I have searched the [issues](https://github.com/python-poetry/poetry/issues) of this repo and believe that this is not a duplicate.
- [x] If an exception occurs when executing a command, I executed it again in debug mode (`-vvv` option).

- **OS version and name**: >>>cat /etc/os-release
NAME="Ubuntu"
VERSION="20.04.4 LTS (Focal Fossa)"
ID=ubuntu
ID_LIKE=debian
PRETTY_NAME="Ubuntu 20.04.4 LTS"
VERSION_ID="20.04"
HOME_URL="https://www.ubuntu.com/"
SUPPORT_URL="https://help.ubuntu.com/"
BUG_REPORT_URL="https://bugs.launchpad.net/ubuntu/"
PRIVACY_POLICY_URL="https://www.ubuntu.com/legal/terms-and-policies/privacy-policy"
VERSION_CODENAME=focal
UBUNTU_CODENAME=focal
- **Poetry version**: Multiple
- **Link of a [Gist](https://gist.github.com/) with the contents of your pyproject.toml file**: 

## Issue
TL;DR: `poetry export` now seems to use 5x+ more memory than it used to.

In a CircleCI workflow, we run `poetry export`. This has worked for months, but sometime between Friday (8/5/2022) and Monday (8/8/2022) morning, something changed and caused `poetry export` to use significantly more memory.

Using this [cimg/python:3.9 docker image](https://hub.docker.com/layers/python/cimg/python/3.9/images/sha256-d38e7296a2d042bd84e1a3fdc9ebaac897714b391072bc2293dd7b91b178b3ad?context=explore) which has poetry 1.1.13 installed, we install poetry 1.1.14 using `curl -sSL https://raw.githubusercontent.com/python-poetry/poetry/master/install-poetry.py | python3 - --version 1.1.14`.  

Then running `poetry export` produces the following allocations (using https://github.com/bloomberg/memray):
```
⠋ Calculating high watermark... ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━   Memray ERROR: Failed to process memory record

Allocation metadata
-------------------
[...]

PR #6160, opened 2022-08-13
avoid tomlkit versions with memory leak

radoering:
# Pull Request Check List

Resolves: #6144


mkniewallner (dismissed):
Just a suggestion to add a comment linking to https://github.com/sdispater/tomlkit/issues/225 to better explain why we want to avoid those versions, but LGTM otherwise.

mkniewallner (approved):

PR #6163, opened 2022-08-14
mypy: adapt type hints to change in poetry-core#442

radoering:
Necessary changes when updating poetry-core to a version containing python-poetry/poetry-core#442

### DIFF
--- a/src/poetry/utils/extras.py
+++ b/src/poetry/utils/extras.py
@@ -9,6 +9,7 @@ if TYPE_CHECKING:
     from collections.abc import Sequence
     from typing import Mapping
 
+    from packaging.utils import NormalizedName
     from poetry.core.packages.package import Package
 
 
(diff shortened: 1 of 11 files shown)


src/poetry/utils/extras.py
```diff
@@ -43,7 +44,7 @@ def get_extra_package_names(
     # keep record of packages seen during recursion in order to avoid recursion error
     seen_package_names = set()
 
-    def _extra_packages(package_names: Iterable[str]) -> Iterator[str]:
```
dimbleby, line 46:
```
   def _extra_packages(
        package_names: Iterable[NormalizedName],
    ) -> Iterator[NormalizedName]:
```
and remove the redundant `canonicalize_name()` at line 52

src/poetry/utils/extras.py
```diff
@@ -9,12 +9,13 @@
...
     from typing import Mapping
 
+    from packaging.utils import NormalizedName
     from poetry.core.packages.package import Package
 
 
 def get_extra_package_names(
     packages: Sequence[Package],
-    extras: Mapping[str, list[str]],
+    extras: Mapping[str, list[NormalizedName]],
```
dimbleby, line 18: If we believe this change, then the call to `canonicalize_name` a few lines later is redundant and should be removed

I'm not completely sure that I do believe it: some callers populate this parameter by reading from the locker: is `lock_data.get("extras", {})` definitely returning `NormalizedName`s?

I think the answer is that it is if the lockfile was written by a sufficiently recent poetry, possibly it isn't otherwise.

dimbleby: (not to mention that the text in the locker is often like
```toml
[package.extras]
foo = ["foo[bar] (>=1.5)"]
```
surely that doesn't work in this code at all (regardless of typing)?!?

radoering: Actually, poetry 1.2 should be able to read lock files generated by poetry 1.1. Thus, I think we should keep it.

dimbleby: both poetry 1.1 and 1.2 write un-normalized strings to the lockfile and so the annotation is flat wrong.

radoering: Ahh, now I see.

dimbleby (approved):

neersighted (approved):

PR #6164, opened 2022-08-14
Use flake8-pie

AlpAribal:
Relates-to: #4776 

Adding the hook leads to four new warnings for the repo. Each warning was handled in a separate commit.
- [PIE781](https://github.com/sbdchd/flake8-pie#pie781-assign-and-return): assign-and-return
  ~~All instances of this warning were fixed~~ [Edit: This warning is ignored altogether as it does not play well with mypy and there is not 100% agreement with the enforced style.] 
- [PIE803](https://github.com/sbdchd/flake8-pie#pie803-prefer-logging-interpolation): prefer-logging-interpolation
  This check produces false positives (`debug()` calls are always flagged as if they belong to a logger). PR ignores such instances and fixes the rest.
- [PIE786](https://github.com/sbdchd/flake8-pie#pie786-precise-exception-handlers): precise-exception-handlers
  PR ignores instances where the intention is indeed to catch any exceptions and fixes the rest.
- [PIE798](https://github.com/sbdchd/flake8-pie#pie798-no-unnecessary-class): no-unnecessary-class
  All instances are ignored via an additional entry in flake8 config.

I am aware that some ignore/fix decisions can be considered opinionated, and am happy to make changes where necessary.

### DIFF
--- a/src/poetry/installation/executor.py
+++ b/src/poetry/installation/executor.py
@@ -757,7 +757,7 @@ class Executor:
                         writer.writerow([str(path), "", ""])
 
     def _create_git_url_reference(self, package: Package) -> dict[str, Any]:
-        reference = {
+        return {
             "url": package.source_url,
             "vcs_info": {
                 "vcs": "git",
@@ -768,17 +768,13 @@ class Executor:
         if package.source_subdirectory:
             reference["subdirectory"] = package.source_subdirectory
 
-        return reference
-
     def _create_url_url_reference(self, package: Package) -> dict[str, Any]:
         archive_info = {}
 
         if package.name in self._hashes:
             archive_info["hash"] = self._hashes[package.name]
 
-        reference = {"url": package.source_url, "archive_info": archive_info}
-
-        return reference
+        return {"url": package.source_url, "archive_info": archive_info}
 
     def _create_file_url_reference(self, package: Package) -> dict[str, Any]:
         archive_info = {}
--- a/src/poetry/installation/executor.py
+++ b/src/poetry/installation/executor.py
@@ -757,7 +757,7 @@ class Executor:
                         writer.writerow([str(path), "", ""])
 
     def _create_git_url_reference(self, package: Package) -> dict[str, Any]:
-        return {
+        reference = {
             "url": package.source_url,
             "vcs_info": {
                 "vcs": "git",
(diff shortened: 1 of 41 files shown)


dimbleby:
as the pipelines are showing you, pie781 is at odds with appeasing mypy.  IMO it's more important that mypy be happy.

(I believe that there's work going on to improve the typing eg in cleo, so perhaps we'll one day be able to avoid the mypy-appeasement and just write the code more naturally)

AlpAribal:
> as the pipelines are showing you, pie781 is at odds with appeasing mypy. IMO it's more important that mypy be happy.

How would you recommend tackling this issue? Should I ignore pie781 for good or ignore mypy type warnings for these cases?

dimbleby:
> How would you recommend tackling this issue? 

I don't much mind either way.   The main thing is that I think that the mypy checking is valuable, so making that happy should 'win'.

Personally I rather like the style that pie781 objects to so I'd probably suppress it altogether, but this is certainly a matter of taste.  

If you think it's generally a valuable warning then I guess ignore it explicitly in the places where we have deliberately gone against it for the purpose of typechecking.

src/poetry/installation/executor.py
```diff
@@ -757,7 +757,7 @@ def _save_url_reference(self, operation: Operation) -> None:
                         writer.writerow([str(path), "", ""])
 
     def _create_git_url_reference(self, package: Package) -> dict[str, Any]:
-        reference = {
+        return {
```
dimbleby, line 760: this one has gone wrong: we now can never hit the `if package.source_subdirectory:` branch

AlpAribal:
> Personally I rather like the style that pie781 objects to so I'd probably suppress it altogether, but this is certainly a matter of taste.

I see, I do not have a strong preference either. I will ignore the warning.

src/poetry/puzzle/solver.py
```diff
@@ -99,7 +99,7 @@ def solve_in_compatibility_mode(
         depths = []
         for override in overrides:
             self._provider.debug(
-                "<comment>Retrying dependency resolution "
+                "<comment>Retrying dependency resolution "  # noqa: PIE803
```
neersighted, line 102: If we're going to embrace the style changes proposed by flake8-pie, I think we should do so consistently here rather than using `# noqa`.

neersighted (changes requested):

src/poetry/puzzle/solver.py
neersighted: I realize now I'm not sure where the `debug()` method on `_provider` comes from -- if there's no built in interpolation, fair enough (but add a comment explaining why). Seems like this might be a fairly coarse lint.

neersighted (approved):

PR #6165, opened 2022-08-14
Unify find_packages of different repository types

radoering:
Moved the repository specific stuff into a `_find_packages` method so that the prerelease handling (which probably is too strict btw, but that's something for another PR) and logging is not duplicated for all repository types.

Supersedes #6157


dimbleby:
Looks good to me!

mkniewallner (approved):
Nice one :+1:

Issue #6153, opened 2022-08-11
Inconsistent Sort Direction in Lock File


Kurt-von-Laven:
- [x] I am on the [latest](https://github.com/python-poetry/poetry/releases/latest) Poetry version.
- [x] I have searched the [issues](https://github.com/python-poetry/poetry/issues) of this repo and believe that this is not a duplicate.
- [x] If an exception occurs when executing a command, I executed it again in debug mode (`-vvv` option).

- **OS version and name**: Ubuntu 20.04 LTS
- **Python version**: 3.10.5
- **Poetry version**: 1.1.14
- **Link of a [Gist](https://gist.github.com/) with the contents of your pyproject.toml file**: [pyproject.toml](https://github.com/ScribeMD/pre-commit-action/blob/007ef3b9cf1594655d356c2e210343f70ae2fb81/pyproject.toml)

## Issue

Today (August 11th, 2022), between 12:48 and 18:25 UTC `poetry lock --no-update` began producing inconsistent output on the same `pyproject.toml`, breaking all of our CI pipelines. When run locally on Ubuntu 22.04 LTS, `poetry lock --no-update` continues to consistently produce this [poetry.lock](https://github.com/ScribeMD/pre-commit-action/blob/007ef3b9cf1594655d356c2e210343f70ae2fb81/poetry.lock). This is consistent with past runs on Ubuntu 22.04 LTS in CI and hence what I perceive to be the correct output. [Here](https://github.com/ScribeMD/pre-commit-action/runs/7796869979?check_suite_focus=true) is a CI workflow on Ubuntu 20.04 LTS showing the diff (reproduced below) between the old (correct) and new (incorrect) output. The new output appears consistent across CI runs. Note that the new order of lines within `[package.extras]` is reverse alphabetical order, and the new order of dependencies within each line is reversed with respect to the original order. One possible root cause may be that the lock command's sorting behavior is not fully self-contained (i.e., it's system dependent), and GitHub Actions rolled out a seemingly innocuous change to their ubuntu-20.04 image that exposed this preexisting bug. Note also that two spaces were added before `metadata.python-versions`.

```diff
diff --git a/poetry.lock b/poetry.lock
index f9756ea..26afd4e 100644
--- a/poetry.lock
+++ b/poetry.lock
@@ -146,8 +146,8 @@ optional = false
 python-versions = ">=3.7"
 
 [package.extras]
[...]

PR #6169, opened 2022-08-14
Sort extras and extras dependencies in lockfile

mkniewallner:
# Pull Request Check List

Resolves: #6153

This PR sorts extras under `[package.extras]` in the lockfile, as well as their dependencies.

Note: https://github.com/python-poetry/poetry/blob/ed88e34dd769019d365140e4d2526a248a872ab5/src/poetry/packages/locker.py#L591-L592 suggests to use `to_pep_508` once https://github.com/python-poetry/poetry-core/pull/102 is merged, but doing so, even with `with_extras=False`, would add environment markers, which I'm not sure is the desired behaviour, so I preferred to leave that untouched.
If it is the expected behaviour, I'd be happy to do the change in a follow-up PR.

### DIFF
--- a/src/poetry/packages/locker.py
+++ b/src/poetry/packages/locker.py
@@ -587,7 +587,7 @@ class Locker:
 
         if package.extras:
             extras = {}
-            for name, deps in package.extras.items():
+            for name, deps in sorted(package.extras.items()):
                 # TODO: This should use dep.to_pep_508() once this is fixed
                 # https://github.com/python-poetry/poetry-core/pull/102
                 extras[name] = sorted(
--- a/src/poetry/packages/locker.py
+++ b/src/poetry/packages/locker.py
@@ -590,10 +590,10 @@ class Locker:
             for name, deps in package.extras.items():
                 # TODO: This should use dep.to_pep_508() once this is fixed
                 # https://github.com/python-poetry/poetry-core/pull/102
-                extras[name] = [
+                extras[name] = sorted(
                     dep.base_pep_508_name if not dep.constraint.is_any() else dep.name
                     for dep in deps
-                ]
+                )
 
             data["extras"] = extras
 
--- a/tests/installation/test_installer.py
+++ b/tests/installation/test_installer.py
@@ -14,7 +14,6 @@ from cleo.io.io import IO
 from cleo.io.null_io import NullIO
 from cleo.io.outputs.buffered_output import BufferedOutput
 from cleo.io.outputs.output import Verbosity
-from deepdiff import DeepDiff
 from poetry.core.packages.dependency_group import MAIN_GROUP
 from poetry.core.packages.dependency_group import DependencyGroup
 from poetry.core.packages.package import Package
@@ -1164,7 +1163,7 @@ def test_installer_with_pypi_repository(
 
     expected = fixture("with-pypi-repository")
 
-    assert not DeepDiff(expected, locker.written_data, ignore_order=True)
+    assert expected == locker.written_data
 
 
 def test_run_installs_with_local_file(
--- a/tests/installation/fixtures/with-pypi-repository.test
+++ b/tests/installation/fixtures/with-pypi-repository.test
@@ -7,7 +7,7 @@ optional = false
 python-versions = "*"
 
 [package.extras]
-dev = ["coverage", "hypothesis", "pympler", "pytest", "six", "zope-interface", "sphinx", "zope-interface"]
+dev = ["coverage", "hypothesis", "pympler", "pytest", "six", "sphinx", "zope-interface", "zope-interface"]
 docs = ["sphinx", "zope-interface"]
 tests = ["coverage", "hypothesis", "pympler", "pytest", "six", "zope-interface"]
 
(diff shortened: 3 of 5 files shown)


src/poetry/packages/locker.py
```diff
@@ -587,12 +587,12 @@ def _dump_package(self, package: Package) -> dict[str, Any]:
 
         if package.extras:
             extras = {}
-            for name, deps in package.extras.items():
+            for name, deps in sorted(package.extras.items()):
                 # TODO: This should use dep.to_pep_508() once this is fixed
                 # https://github.com/python-poetry/poetry-core/pull/102
                 extras[name] = [
-                    dep.base_pep_508_name if not dep.constraint.is_any() else dep.name
-                    for dep in deps
```
dimbleby, line 595: simply
```patch
-                extras[name] = [
+                extras[name] = sorted(
                     dep.base_pep_508_name if not dep.constraint.is_any() else dep.name
                     for dep in deps
-                ]
+                )
```
surely?

mkniewallner: Indeed, thanks. Amended in https://github.com/python-poetry/poetry/pull/6169/commits/ecbd171fe4984fd1b800503314d81be9b4c277be.

tests/installation/test_installer.py
```diff
@@ -1164,7 +1163,7 @@ def test_installer_with_pypi_repository(
 
     expected = fixture("with-pypi-repository")
 
-    assert not DeepDiff(expected, locker.written_data, ignore_order=True)
+    assert expected == locker.written_data
```
radoering, line 1166: Can we apply the same change in `test_installer_old.py`?

mkniewallner: Done in https://github.com/python-poetry/poetry/pull/6169/commits/92ed15f1f5a2a29c07baabedcaed53739efa9084, thanks.

radoering (approved):

Kurt-von-Laven:
@radoering, would you consider accepting a pull request to cherry pick this fix to Poetry 1.1? The issue it fixes is presently breaking Dependabot for all users (in addition to all of our CI pipelines).

radoering:
Not sure if that's enough for another 1.1 release. We will probably only do another 1.1 release if poetry breaks for the majority of users like the breaking warehouse change, which was the reason for 1.1.14.

Is it really breaking Dependabot or is it just adding noise to the PRs? Actually, I think some shuffling in addition to the real change is annoying but not a blocker.

Kurt-von-Laven:
I get what you are saying about Dependabot continuing to function. The extra pull request noise isn't a major concern in and of itself. There is presently no single correct lock file though as the output of `poetry lock --no-update` is inconsistent, which complicates automatically checking the lock file in CI and will lead to a lot of time wasting merge conflicts on high velocity projects. As a practical consequence of there being no single correct lock file, folks will disable or refrain from implementing automated checking of lock files in CI. Without such checking, they will be relying on human reviewers to catch issues in a machine-generated file that is collapsed in pull requests by default on GitHub at least. This is a serious security concern for the Python ecosystem if left unaddressed for a protracted period of time as it will facilitate the creation of malicious pull requests that modify the lock file in ways that are not consistent with the `pyproject.toml` reviewed (and comprehended) by the human. I am saying all of this on the potentially erroneous assumption that a stable 1.2 release is at least a few months away.

radoering:
@finswimmer Since #5834 proposes to do another 1.1 release: Do you think we should include this one as well if we really do another release? It seems like a quite riskless change and #6153 could be considered as an (externally triggered) regression. 🤷

mkniewallner:
> @finswimmer Since #5834 proposes to do another 1.1 release: Do you think we should include this one as well if we really do another release? It seems like a quite riskless change and #6153 could be considered as an (externally triggered) regression. shrug

+1 on backporting, if we are certain to do another 1.1 release, since this is pretty much harmless. I can prepare a backport PR just in case.

dimbleby:
there are zillions of bug fixes that could be backported, this one isn't special, just release 1.2 already!  Then no-one will need the backport.

Kurt-von-Laven:
This is the only Poetry 1.1 bug I'm aware of that knocks over many CI pipelines in a way where the path-of-least-resistance workaround is to disable (or not implement) an important security check (namely that `poetry lock --no-update` doesn't modify `poetry.lock`). It also creates friction with Dependabot in a way that's cumbersome to deal with in an automated fashion. (python-poetry/poetry#6201 is similar but at least has the simpler workaround of removing indentation from `pyproject.toml` that is more likely to be more widely adopted if necessary.) Even once 1.2 is stable, there will probably be teams that remain on 1.1 for one reason or another given that there is a [1,022 commit delta between 1.1 and master](https://github.com/python-poetry/poetry/tree/d2abf921069208d1e568c5cad4a8c0213d56a969), which is 42% of the [2,460 commits presently on master](https://github.com/python-poetry/poetry/tree/8df4614dd6201c24892acc2316b680e085276965). I don't mean this to say that releasing 1.2 shouldn't be a very high priority, but my impression is that the maintainers already strongly agree with us on that point.

dimbleby:
Folk who remain on 1.1 are going to be out of support soon anyway, the question is whether this fix is important enough to divert maintainer bandwidth from the more important task of releasing 1.2.

IMO it's not, releasing 1.2 should be the sole priority right now, anyone who could be making that happen and is instead doing other poetry work is doing it wrong.

But if folk want to spend their time doing things that I consider less useful: well I'm not king of the world and can't stop them!

PR #6178, opened 2022-08-16
Fix an infinite loop in the solver

dimbleby:
I tried bumping crashtest to 0.4.0 in poetry itself.  This can't work because cleo requires a lower version than that, but instead of failure I hit an infinite loop.

Turned out there are two separate bugs here, both with the effect that version 1.0.0a5 of cleo is retried even after it has been ruled out because of this incompatibility

- first, we go into `_get_locked()` with a constraint that should explicitly rule out 1.0.0a5 - `>1.0.0a5,<2.0.0` - but because of the prerelease we decide that 1.0.0a5 is fine after all.  So far as I can see the code that I've removed just doesn't want to be there: anyway removing it doesn't break any existing tests and helps this one...

- then we go into `search_for_installed_packages()` again having started with a constraint that should rule out 1.0.0a5: but in here we pay no attention to that constraint at all.  So we again rediscover the already ruled-out version.  I've fixed this one by respecting the dependency in that function.

### DIFF
--- a/src/poetry/mixology/version_solver.py
+++ b/src/poetry/mixology/version_solver.py
@@ -513,11 +513,9 @@ class VersionSolver:
         locked = self._locked.get(dependency.name, [])
         for dependency_package in locked:
             package = dependency_package.package
-            if (allow_similar or dependency.is_same_package_as(package)) and (
-                dependency.constraint.allows(package.version)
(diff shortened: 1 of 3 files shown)


radoering (approved):
LGTM 👍

Issue #6182, opened 2022-08-17
`poetry install` ignore build script failure


touilleMan:
- [x] I am on the [latest](https://github.com/python-poetry/poetry/releases/latest) Poetry version.
- [x] I have searched the [issues](https://github.com/python-poetry/poetry/issues) of this repo and believe that this is not a duplicate.
- [x] If an exception occurs when executing a command, I executed it again in debug mode (`-vvv` option).

- **OS version and name**: all
- **Poetry version**: 1.2.0b3

## Issue

When using a build script in the `pyproject.toml` (i.e. the `script = "build.py"` in `[tool.poetry.build]`), the return status from the build script is ignored by the `poetry install` command:

```bash
$ poetry install -E backend -E core
Installing dependencies from lock file

No dependencies to install or update

Preparing build environment with build-system requirements poetry-core>=1.0.0, setuptools, maturin~=0.13, PyQt5~=5.15, Babel~=2.10, docutils~=0.17
PYTHON_EXECUTABLE_PATH=/tmp/tmp44hka6ee/.venv/bin/python
Launching poetry build.py script
Traceback (most recent call last):
  File "/mnt/c/Users/gbleu/source/repos/parsec-cloud/parsec-cloud/build.py", line 74, in <module>
    build()
  File "/mnt/c/Users/gbleu/source/repos/parsec-cloud/parsec-cloud/build.py", line 39, in build
    raise SystemError("D'oh !")
SystemError: D'oh !
Installing the current project: parsec-cloud (v2.11.1+dev)
$ echo $?
0
```

The expected behavior would be to have the `poetry install` command stopped and return an error status, this is for instance the behavior of `poetry build --wheel`

```bash
$ poetry build --format wheel
Preparing build environment with build-system requirements poetry-core>=1.0.0, setuptools, maturin~=0.13, PyQt5~=5.15, Babel~=2.10, docutils~=0.17
Building parsec-cloud (v2.11.1+dev)
PYTHON_EXECUTABLE_PATH=/tmp/tmp5ef_ve_y/.venv/bin/python
Launching poetry build.py script
Traceback (most recent call last):
  File "/mnt/c/Users/gbleu/source/repos/parsec-cloud/parsec-cloud/build.py", line 74, in <module>
    build()
  File "/mnt/c/Users/gbleu/source/repos/parsec-cloud/parsec-cloud/build.py", line 39, in build
    raise SystemError("D'oh !")
SystemError: D'oh !

[...]

PR #6183, opened 2022-08-17
Fix EditableBuilder ignoring error in build script

touilleMan:
# Pull Request Check List

Resolves: #6182


qualiaa:
I believe this will fix the issue, but what was the purpose of including `call=True` in the first place? If it serves some deeper purpose, then an alternative fix would be to throw an exception if the return value of `env.run` is non-zero (this is the returncode of `call`: https://github.com/python-poetry/poetry/blob/c8945eb110aeda611cc6721565d7ad0c657d453a/src/poetry/utils/env.py#L1539)

dimbleby:
superseded by #7812, please close

PR #6185, opened 2022-08-17
installer: use locked version of vcs dependency with extras instead of latest

radoering:
# Pull Request Check List

Resolves: #6184

### DIFF
--- a/tests/installation/test_installer.py
+++ b/tests/installation/test_installer.py
@@ -2473,3 +2473,45 @@ def test_installer_should_use_the_locked_version_of_git_dependencies(
         source_reference="master",
         source_resolved_reference="123456",
     )
+
+
+@pytest.mark.parametrize("is_locked", [False, True])
+def test_installer_should_use_the_locked_version_of_git_dependencies_with_extras(
+    installer: Installer,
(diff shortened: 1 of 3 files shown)


tests/installation/test_installer.py
```diff
@@ -2473,3 +2473,45 @@ def test_installer_should_use_the_locked_version_of_git_dependencies(
...
+    package: ProjectPackage,
+    repo: Repository,
+    is_locked: bool,
+):
+    if is_locked:
+        locker.locked(True)
+        locker.mock_lock_data(fixture("with-vcs-dependency-with-extras"))
+        expected_reference = "123456"
+    else:
+        expected_reference = "9cf87a285a2d3fbb0b9fa621997b3acc3631ed24"
```
mkniewallner, line 2491: nit: Given our growing usage of this value in tests, a follow-up PR that defines it in a reusable constant may make it clearer where it comes from (I'd be happy to do the change).

mkniewallner (approved):

tests/installation/test_installer.py
radoering: Sounds good.

Issue #5142, opened 2022-02-02
Invalid hashes when running multiple Poetry installs simultaneously


hopper-signifyd:
- [x] I am on the [latest](https://github.com/python-poetry/poetry/releases/latest) Poetry version.
- [x] I have searched the [issues](https://github.com/python-poetry/poetry/issues) of this repo and believe that this is not a duplicate.
- [x] If an exception occurs when executing a command, I executed it again in debug mode (`-vvv` option).

- **OS version and name**: Linux (run using Docker)
- **Poetry version**: 1.1.12
- **Link of a [Gist](https://gist.github.com/) with the contents of your pyproject.toml file**: https://gist.github.com/hopper-signifyd/96e8e36f5c91f7d32e109182b8b7b62a

^(please note that projects are in sibling directories named project_a and project_b)

## Issue
When running multiple poetry installs simultaneously with a shared Poetry cache directory, the operation commonly fails with the infamous "invalid hashes" error. This is common in CI and monorepo environments. Here's a sample:

```
RuntimeError

  Invalid hashes (sha256:c7a7026632f45188f4a4548cc308c5c0683d9b8259da5cbfe0301f7527843eb4) for pandas (1.0.5) using archive pandas-1.0.5-cp36-cp36m-manylinux1_x86_64.whl. Expected one of
[omitted the other hashes for the sake of brevity]
sha256:faa42a78d1350b02a7d2f0dbe3c80791cf785663d6997891549d0f86dc49125e.

  at ~/.local/share/pypoetry/venv/lib/python3.9/site-packages/poetry/installation/executor.py:627 in _download_link
      623│                     )
      624│                 )
      625│
      626│             if archive_hashes.isdisjoint(hashes):
    → 627│                 raise RuntimeError(
      628│                     "Invalid hashes ({}) for {} using archive {}. Expected one of {}.".format(
      629│                         ", ".join(sorted(archive_hashes)),
      630│                         package,
      631│                         archive_path.name,
```

After receiving this error, if I run `find . -name pandas-1.0.5-cp36-cp36m-manylinux1_x86_64.whl`, and then run a checksum on the file, I usually get a SHA256 from the "expected" list in the error message. If I don't get an expected hash, it seems to be related to another poetry install process that's still running and downloading that artifact.

My current working theory is something like this:
* Poetry install process A checks cache for an arbitrary package (say pandas, since that's what's in the example error above). The process get a cache miss and starts downloading pandas.
[...]

PR #6186, opened 2022-08-17
Fix a race condition in simultaneous poetry installations updating the cache

zweger:
If one poetry installation is writing to the cache while a second installer is attempting to read that cache file, the second installation will fail because the in-flight cache file is invalid while it is still being written to by the first process. 

This PR resolves this issue by having Poetry write to a temporary file in the cache directory first, and then rename the file after it's written, which is ~atomic.

# Pull Request Check List

Resolves: #5142 

I'm not sure how to test this change, as the conditions which cause this bug to appear are a little hard to reproduce.


radoering:
Makes sense to me. I just wonder if it's wise to import a helper function from `requests.utils`. It's probably not part of the public API, is it? Maybe, we should add "our own" `atomic_open()` (to `poetry.utils.helpers` maybe)?

I don't know how to test this either. Maybe, we can go without a test here.

@neersighted What do you think?

neersighted:
I'm wondering if #6471 is the better solution -- I agree that we should yoink `atomic_open` into this repo (with a comment for attribution/licensing) if we go forward with this. However, as either would likely be part of 1.3 exclusively, I think making the OS work for us instead of against us is the better long-term path forward.

dimbleby:
#6471 and this are independent fixes, it would make perfect sense to merge them both

(#6471 relates to files owned by cachecontrol's filecache, this one relates to a file written to by poetry regardless of whether caching is enabled or not)

neersighted:
Fair enough, and good point

zweger:
Hi, thanks for the feedback. I've updated this PR to include `atomic_open` in `poetry.utils.helpers`. Regarding testing, I will note that while poetry doesn't have a test for this particular situation (concurrent installations / ctrl+c during install), poetry's current tests do at least cover this code.

neersighted (approved):

PR #6191, opened 2022-08-18
Replace `strtobool` by a simpler check in `shell` command

mkniewallner:
# Pull Request Check List

Closes #5878.

Alternative implementation to https://github.com/python-poetry/poetry/pull/5878 that takes into accounts comments made in the PR, mostly based on [this comment](https://github.com/python-poetry/poetry/pull/5878#issuecomment-1159795698) that suggests to only check for non-empty values.

Since there were no tests at all for `shell` command, this PR also adds some.

_Note: For some reason, `SHELL` environment variable is not set when running tests on FreeBSD through Cirrus CI, so https://github.com/python-poetry/poetry/pull/6191/commits/7668ced985fff74479fab254135a6c8162e60320 explicitly sets it, otherwise shell tests fail._


tests/console/commands/test_shell.py
```diff
@@ -0,0 +1,84 @@
...
+    assert tester.io.fetch_output() == expected_output
+    assert tester.status_code == 0
+
+
+@pytest.mark.parametrize(
+    ("poetry_active", "real_prefix", "prefix", "expected"),
+    [
+        (None, None, "", False),
+        ("", None, "", False),
+        (" ", None, "", True),
```
mkniewallner, line 55: Should we `strip()` `POETRY_ACTIVE` to remove white spaces, as a safety measure?

src/poetry/console/commands/shell.py
```diff
@@ -21,10 +20,7 @@ def handle(self) -> int:
...
 
         # Check if it's already activated or doesn't exist and won't be created
-        venv_activated = strtobool(environ.get("POETRY_ACTIVE", "0")) or getattr(
-            sys, "real_prefix", sys.prefix
-        ) == str(self.env.path)
-        if venv_activated:
+        if self._is_venv_activated():
             self.line(
                 f"Virtual environment already activated: <info>{self.env.path}</>"
             )
```
mkniewallner, line 26: Not related to those changes, but the line after this one makes the method return `0`, so the command will have an exit code `0` in the shell.
Is this the correct thing to do? I would have expected the command to end up with an exit code `1` if the shell is already active.

tests/console/commands/test_shell.py
neersighted: That's never been convention when testing if a variable is non-empty in shell, so I don't think it's necessary in Poetry.

src/poetry/console/commands/shell.py
neersighted: I think the intention is to be a soft no-op as opposed to a failure -- we could always revisit this later, but I don't think it's a major issue.

neersighted (approved):
LGTM -- I like this approach quite a bit. I'm not sure why you need to manually define `SHELL` in Cirrus, but that also looks good to me.

Issue #6018, opened 2022-07-15
`poetry env remove` removes wrong environment


KDMichaelis:
- [x] I am on the [latest](https://github.com/python-poetry/poetry/releases/latest) Poetry version.
- [x] I have searched the [issues](https://github.com/python-poetry/poetry/issues) of this repo and believe that this is not a duplicate.
- [ ] If an exception occurs when executing a command, I executed it again in debug mode (`-vvv` option).

- **OS version and name**: Ubuntu 20.04
- **Poetry version**: 1.1.14

## Issue

I created two poetry environments, `project-tmp-1BotOSEK-py3.9` and `tmp-project-yuY8E-Dp-py3.9`. In my project only the first one is available and activated:
```
$ poetry env list
project-tmp-1BotOSEK-py3.9 (Activated)
```
I am trying to delete the other environment using the absolute path according to the [documentation](https://python-poetry.org/docs/managing-environments/#deleting-the-environments) as follows:
```
$ poetry env remove /home/user/.cache/pypoetry/virtualenvs/tmp-project-yuY8E-Dp-py3.9/bin/python
```
However, this deletes the wrong environment (the activated one) instead:
```
Deleted virtualenv: /home/user/.cache/pypoetry/virtualenvs/project-tmp-1BotOSEK-py3.9
```
I would expect that using the absolute path exactly the specified environment would be deleted instead of the activated one, including any data connected to it (e.g. the entry in `envs.toml`). It actually seems that the entry for the correct (specified) environment was deleted but the wrong (activated) environment is erased instead, leading to inconsistent information in `envs.toml`.
If for some reason it is desired that a virtualenv cannot be deleted from a project where it's not available (or while another one is activated) the command should at least fail when trying to do so.
As a side note, I am using direnv and the command [layout_poetry](https://github.com/direnv/direnv/wiki/Python#poetry) defined in `~/.direnvrc` and called in `.envrc` within my project.

dhvcc:
Are there any steps to reproduce this? I imagine renaming your directory/project and creating a new venv?

dhvcc:
@python-poetry/triage
Hi, I'd like to take on this issue. The solution is pretty easy - just to extract real venv path with something like that `import sys; print(sys.prefix)` and removing it instead of **always** generating venv name from current project.

This issue is easily reproducable:
1. Create a dir `mkdir dir_name` and run `poetry init & poetry install` (remember this venv's path)
[...]

PR #6195, opened 2022-08-18
Update env remove logic

dhvcc:
# Pull Request Check List

Resolves: #6018 

- [X] Updated **documentation** for changed code. (no documentation change is needed I think)

### Changes

1. Added a check so that if `python` argument is a file (then it should be a python path) - extract it's venv name and raise `IncorrectEnvError` if it doesn't belong to this project
    **Before**
    ```
    └❯ poetry env remove ~/.cache/pypoetry/virtualenvs/different-project-OKfJHH_5-py3.10/bin/python
    /bin/sh: 1: different-project-OKfJHH_5-py3.10: not found

    Deleted virtualenv: ~/.cache/pypoetry/virtualenvs/poetry-4pWfmigs-py3.10
    ```
    Removes current project's env, which is wrong.
    **After**
    ```
    └❯ poetry env remove ~/.cache/pypoetry/virtualenvs/different-project-OKfJHH_5-py3.10/bin/python

    Env different-project-OKfJHH_5-py3.10 doesn't belong to this project.
    ```
2. Added the exact same check as before ^, but for cases where env name is passed.
    **Before**
    ```
    └❯ poetry env remove different-project-OKfJHH_5-py3.10      
     /bin/sh: 1: different-project-OKfJHH_5-py3.10: not found

     Command different-project-OKfJHH_5-py3.10 -c "import sys; print('.'.join([str(s) for s in sys.version_info[:3]]))" errored with the following return code 127, and output: 
    ```
    Errors while trying to exec env name as an interpreter, error is not clear.
    **After**
    ```
    └❯ poetry env remove different-project-OKfJHH_5-py3.10                      

    Env different-project-OKfJHH_5-py3.10 doesn't belong to this project.
    ```
3. Added a couple of tests for **new** and for **old** scenarios which weren't tested.
4. Added `venv_name` fixture for `tests/utils` directory to use in `test_env`. Also replaced some of `"simple-project"` hardcoded value to use `poetry.package.name`

It's up to maintainers to choose what they want for this project - I'm happy either way if we at least fix the bug. I can remove/change any of the stuff I added on top of the fix. But yeah I just decided that if we fix the bug, we might also make some improvements/changes in this area of code. Any thoughts on this are welcome, thanks!

### DIFF
--- a/src/poetry/utils/env.py
+++ b/src/poetry/utils/env.py
@@ -175,6 +175,7 @@ print('.'.join([str(s) for s in sys.version_info[:3]]))
 GET_PYTHON_VERSION_ONELINER = (
     "\"import sys; print('.'.join([str(s) for s in sys.version_info[:3]]))\""
 )
+GET_ENV_PATH_ONELINER = '"import sys; print(sys.prefix)"'
 
 GET_SYS_PATH = """\
 import json
@@ -461,6 +462,12 @@ class EnvError(Exception):
     pass
 
 
+class IncorrectEnvError(EnvError):
+    def __init__(self, env_name: str) -> None:
+        message = f"Env {env_name} doesn't belong to this project."
+        super().__init__(message)
+
+
 class EnvCommandError(EnvError):
     def __init__(self, e: CalledProcessError, input: str | None = None) -> None:
         self.e = e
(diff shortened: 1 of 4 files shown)


neersighted (approved):

neersighted:
I think this code could use a general refactor/cleanup to make it clear we now support 'identifiers' that can be Python version numbers, Python executable paths, or environment names (as well as maybe a docs pass). If you'd like to do that, it would be great -- if not, this is still incremental progress.

dhvcc:
@neersighted yeah, I agree this could be a bit clearer in the code. I'd like to try to clean this up, wondering if there's any plan/ideas for it right now?

neersighted:
Personally I would rework the code to actually be a simple `env.remove()` (actually using `self` to find the environment to remove) and then add functionality to `EnvManager` to find/create an `Env` instance for the identifier in question.

neersighted:
If you're interested in working in this area still, #2748 might provide some more ideas for improvements to the CLI experience.

Issue #2288, opened 2020-04-09
Support `--no-compile` when installing a package


laike9m:
- [x] I have searched the [issues](https://github.com/python-poetry/poetry/issues) of this repo and believe that this is not a duplicate.
- [x] I have searched the [documentation](https://python-poetry.org/docs/) and believe that my question is not covered.

## Feature Request
Pip supports the [`--no-compile`](https://pip.pypa.io/en/stable/reference/pip_install/#cmdoption-no-compile) option when installing a package. I want to install a [library](https://github.com/thautwarm/moshmosh) that requires this flag, but find it hard to achieve using Poetry.

KotlinIsland:
> #2288 is solved here too now.

\- https://github.com/python-poetry/poetry/pull/6205#issuecomment-1356133534

@Secrus Is this resolved? Can this be closed?

radoering:
Yep, "no-compile" is the default now unless `installer.modern-installation` is set to `false`. See https://python-poetry.org/docs/cli/#install

PR #6205, opened 2022-08-20
Faster build and installation of packages

sdispater:
This PR (which is an updated version of #4213) is the first step towards getting rid of subprocess calls to `pip` for package installations and removals.

The wheel building is handled by https://github.com/pypa/build and the wheel installation is handled by https://github.com/pypa/installer (with modifications to be more lenient).

Note that the wheel building is **not** parallelized due to race conditions that I am yet to track down (it seems like the issue comes from `build` when calling the backend hooks). This is not a blocker though.

Overall this speeds up quite a bit the installation times, especially once wheels are cached.

An example with the installation of 115 packages (warm cache and artifacts already downloaded):

Without the new builder/installer: `226.24s user 15.26s system 488% cpu 49.413 total`
With the new builder/installer: `11.52s user 3.08s system 117% cpu 12.452 total`

Note that this can be deactivated by setting `experimental.installer.modern-installation`.

# Pull Request Check List

Resolves: https://github.com/python-poetry/roadmap/issues/1


src/poetry/factory.py
```diff
@@ -110,23 +112,23 @@ def get_package(cls, name: str, version: str) -> ProjectPackage:
...
         cls,
-        poetry: Poetry,
-        sources: list[dict[str, str]],
         config: Config,
-        io: IO,
+        io: IO | None = None,
         disable_cache: bool = False,
-    ) -> None:
-        if disable_cache:
-            logger.debug("Disabling source caches")
```
radoering, line 122: The debug message got lost by accident, didn't it?

src/poetry/factory.py
```diff
@@ -110,23 +112,23 @@ def get_package(cls, name: str, version: str) -> ProjectPackage:
...
-            )
-            is_default = bool(source.get("default", False))
-            is_secondary = bool(source.get("secondary", False))
+        if io is None:
+            io = NullIO()
+
+        pool = Pool()
+
+        for source in config.get("sources", {}).values():
+            repository = cls.create_package_source(source, config)
```
radoering, line 129:
```suggestion
            repository = cls.create_package_source(
                source, config, disable_cache=disable_cache
            )
```
It seems that got lost, too.

src/poetry/installation/executor.py
```diff
@@ -88,6 +92,11 @@ def __init__(
         self._shutdown = False
         self._hashes: dict[str, str] = {}
 
+    def set_chef(self, chef: Chef) -> Executor:
+        self._chef = chef
+
+        return self
+
```
radoering, line 99: `set_chef` is only required for tests. It's probably a matter of taste, but I wouldn't introduce a setter just for a test but set the attribute `_chef` directly in the test. I don't think it's too bad to access internals in the unit test of the corresponding module.

src/poetry/installation/chef.py
```diff
@@ -17,12 +32,175 @@
...
+                destination = self.get_cache_directory_for_link(Link(archive.as_uri()))
+
+            destination.mkdir(parents=True, exist_ok=True)
+
+            return self._prepare(
+                sdist_dir,
+                destination,
+            )
+
+    def should_prepare(self, archive: Path) -> bool:
```
radoering, line 198:
```suggestion
    def _should_prepare(self, archive: Path) -> bool:
```
only used internally

src/poetry/installation/chef.py
```diff
@@ -17,12 +32,175 @@
...
+        return archive.is_dir() or not self.is_wheel(archive)
+
+    @classmethod
+    def is_wheel(cls, archive: Path) -> bool:
```
radoering, line 202:
```suggestion
    def _is_wheel(cls, archive: Path) -> bool:
```
only used internally

radoering (commented):
Some first (mostly superficial) feedback.

It seems the commits are not cut optimally: The commit "Add a wheel builder" already requires `WheelInstaller` which is added only later in "Add a wheel installer". (At least if we keep the commits separated when merging it may be worth to fix that.)

radoering:
Rebased, resolved merge conflicts, swapped the builder and the installer commit (because the builder commit depends on the installer commit) and addressed some of the review comments.

While executing the tests locally, I noticed that e.g. `test_execute_prints_warning_for_yanked_package` does not install in a temp folder anymore but in poetry's own venv. However, I'm not sure if that is an issue of the new `WheelInstaller` or `MockEnv`. It seems that the pip installer respects the temp path via https://github.com/python-poetry/poetry/blob/3a961dfb67ccc214fa101f5c1a187e2ed077673b/src/poetry/utils/pip.py#L34 but the new `WheelInstaller` derives the destination from https://github.com/python-poetry/poetry/blob/3a961dfb67ccc214fa101f5c1a187e2ed077673b/src/poetry/utils/env.py#L1598
which does not return a temp path.

dimbleby:
I experimented with the arbitrary large-ish install at https://github.com/lincolnloop/python-package-manager-shootout

It looks as though, at least, in that case, the lost parallelism costs more than the faster build buys.

Before:
```bash
$ time poetry install
...
real    1m9.535s
user    5m2.576s
sys     0m40.103s
```
After:
```bash
$ time poetry install
...
real    2m5.472s
user    3m2.383s
sys     1m17.763s
```
so it really has done ~60% of the work - but it has taken nearly twice as long anyway

(both with an empty cache.  With a populated cache the new code is unambiguously faster.  I guess populated cache is more common on a user's own computer, but empty cache is more likely eg in a CI pipeline)

src/poetry/installation/chef.py
```diff
@@ -2,27 +2,206 @@
...
+    def _prepare(
+        self, directory: Path, destination: Path, *, editable: bool = False
+    ) -> Path:
+        from subprocess import CalledProcessError
+
+        from poetry.utils.env import EnvManager
+        from poetry.utils.env import VirtualEnv
+
+        with self._lock, temporary_directory() as tmp_dir:
+            EnvManager.build_venv(tmp_dir, executable=self._env.python, with_pip=True)
```
abn, line 120: We can use `poetry.utils.env.ephemeral_environment` here.

abn: In theory, we do not need `pip` nor `setuptools` in there. Also see #5650 for example usage. The major issue with this approach is that it seems that setting up and installing the environments per-package is hefty. However, this change could contribute to this performance no longer being an issue. Additionally, we should cache these environments globally probably keyed with python version and build requires.

src/poetry/installation/chef.py
```diff
@@ -2,27 +2,206 @@
...
+
+
 class Chef:
     def __init__(self, config: Config, env: Env) -> None:
+        self._config = config
         self._env = env
         self._cache_dir = (
             Path(config.get("cache-dir")).expanduser().joinpath("artifacts")
         )
+        self._lock = threading.Lock()
```
abn, line 96: @sdispater do we need locking here? In theory we already know all the `sdists` we need to build, why not just fan it out via multiprocess. Probably will have to build all required environments before hand if we want to cache/re-use though. But that should all be easily computed.

src/poetry/installation/wheel_installer.py
```diff
@@ -0,0 +1,101 @@
...
+        stream: BinaryIO,
+        is_executable: bool,
+    ) -> RecordEntry:
+        from installer.records import Hash
+        from installer.records import RecordEntry
+        from installer.utils import copyfileobj_with_hashing
+        from installer.utils import make_file_executable
+
+        target_path = os.path.join(self.scheme_dict[scheme], path)
+        if os.path.exists(target_path):
```
abn, line 42: nitpick; we should really use `pathlib` where ever we can.

src/poetry/installation/chef.py
abn: It would be good if you can pull up the `poetry.utils.pep517` module into this change since that would be reusable across inspection as well.

src/poetry/installation/chef.py
dimbleby: naively removing this lock causes things to go wrong eg while it more than recovers the lost performance as described in https://github.com/python-poetry/poetry/pull/6205#issuecomment-1304814783, we also hit
```
 • Installing python-u2flib-server (5.0.1): Failed

  FileNotFoundError

  [Errno 2] No such file or directory: '/tmp/tmps4b9ofmc/petname-2.6'

  at ~/.virtualenvs/poetry/lib/python3.10/site-packages/build/__init__.py:149 in _working_directory
      145│
      146│     try:
      147│         yield
      148│     finally:
    → 149│         os.chdir(current)
      150│
      151│
      152│ def _validate_source_directory(srcdir: PathType) -> None:
      153│     if not os.path.isdir(srcdir):
```

https://github.com/python-poetry/poetry/pull/4213#issuecomment-1029786297 presumably?

radoering: I think the reason is that `build` changes the working directory, which is not thread-safe: https://github.com/pypa/build/blob/63342ce0adc2333123aa63bf6b03c5d2235af324/src/build/__init__.py#L140-L149

With that knowledge we can use the lock more precisely which results in a significant performance improvement. Since this is a threading issue, multiprocessing should also be possible and might even be faster.

radoering:
> I experimented with the arbitrary large-ish install at https://github.com/lincolnloop/python-package-manager-shootout

I followed up on this and created a pyproject.toml with the following three dependencies, which only provide a source distribution:

```
maxminddb = "2.2.0"
petname = "2.6"
python-u2flib-server = "5.0.1"
```

Before each `poetry install`, I removed the three wheels from the artifacts cache. Results:

poetry 1.2.2

```
real    0m5.761s
user    0m9.343s
sys     0m0.929s
```

before using lock more precisely

```
real    0m11.785s
user    0m7.965s
sys     0m1.177s
```

after using lock more precisely

```
real    0m6.837s
user    0m9.108s
sys     0m1.899s
```

Although not as fast as with the old installer, still a significant improvement.

konstin:
The code does compute hashes, but does not compile bytecode, right? I've benchmarked my own wheel installer implementation against pip 22.2.2, poetry 1.2.2 and this branch (`poetry@installer`) using [this sample project](https://github.com/konstin/poc-monotrail/tree/a47b4b953e5042fbc0a5ac86376864cbc0e94702/test-data/poetry/data-science) and [this script](https://github.com/konstin/poc-monotrail/blob/a47b4b953e5042fbc0a5ac86376864cbc0e94702/benchmark.sh) with this branch patched in.

```
test-data/poetry/data-science -E tqdm_feature
pip 22.2.2 from /home/konsti/monotrail/.venv/lib/python3.8/site-packages/pip (python 3.8)
Poetry (version 1.2.2)
Benchmark 1: .venv/bin/pip install -q -r requirements-benchmark.txt
  Time (mean ± σ):     11.868 s ±  0.246 s    [User: 9.395 s, System: 1.324 s]
  Range (min … max):   11.589 s … 12.698 s    20 runs
 
Benchmark 2: poetry install -q --no-root --only main -E tqdm_feature
  Time (mean ± σ):     11.784 s ±  0.410 s    [User: 23.019 s, System: 3.801 s]
  Range (min … max):   10.058 s … 12.069 s    20 runs
 
  Warning: Statistical outliers were detected. Consider re-running this benchmark on a quiet PC without any interferences from other programs. It might help to use the '--warmup' or '--prepare' options.
 
Benchmark 3: poetry@installer install -q --no-root --only main -E tqdm_feature
  Time (mean ± σ):      3.680 s ±  0.059 s    [User: 3.553 s, System: 0.949 s]
  Range (min … max):    3.637 s …  3.899 s    20 runs
 
Benchmark 4: /home/konsti/monotrail/target/release/monotrail poetry-install -E tqdm_feature
  Time (mean ± σ):      4.885 s ±  0.122 s    [User: 12.715 s, System: 2.162 s]
  Range (min … max):    4.676 s …  5.082 s    20 runs
 
Summary
  'poetry@installer install -q --no-root --only main -E tqdm_feature' ran
    1.33 ± 0.04 times faster than '/home/konsti/monotrail/target/release/monotrail poetry-install -E tqdm_feature'
    3.20 ± 0.12 times faster than 'poetry install -q --no-root --only main -E tqdm_feature'
    3.22 ± 0.08 times faster than '.venv/bin/pip install -q -r requirements-benchmark.txt'
```

If i also skip bytecode compilation on pip and mine i got those numbers for a short run (3 repetitions):

```
test-data/poetry/data-science -E tqdm_feature
pip 22.2.2 from /home/konsti/monotrail/.venv/lib/python3.8/site-packages/pip (python 3.8)
Poetry (version 1.2.2)
Benchmark 1: .venv/bin/pip install --no-compile -q -r requirements-benchmark.txt
  Time (mean ± σ):      6.238 s ±  0.172 s    [User: 4.419 s, System: 0.899 s]
  Range (min … max):    6.070 s …  6.413 s    3 runs
 
Benchmark 2: poetry install -q --no-root --only main -E tqdm_feature
  Time (mean ± σ):     10.266 s ±  0.008 s    [User: 18.091 s, System: 2.818 s]
  Range (min … max):   10.259 s … 10.275 s    3 runs
 
Benchmark 3: poetry@installer install -q --no-root --only main -E tqdm_feature
  Time (mean ± σ):      3.560 s ±  0.081 s    [User: 3.396 s, System: 0.901 s]
  Range (min … max):    3.495 s …  3.650 s    3 runs
 
Benchmark 4: /home/konsti/monotrail/target/release/monotrail poetry-install -E tqdm_feature
  Time (mean ± σ):      5.648 s ±  1.659 s    [User: 14.845 s, System: 2.623 s]
  Range (min … max):    3.872 s …  7.157 s    3 runs
 
Benchmark 5: /home/konsti/monotrail/target/release/monotrail poetry-install --no-compile -E tqdm_feature
  Time (mean ± σ):      2.070 s ±  0.226 s    [User: 3.513 s, System: 1.018 s]
  Range (min … max):    1.935 s …  2.330 s    3 runs
 
  Warning: Statistical outliers were detected. Consider re-running this benchmark on a quiet PC without any interferences from other programs. It might help to use the '--warmup' or '--prepare' options.
 
Summary
  '/home/konsti/monotrail/target/release/monotrail poetry-install --no-compile -E tqdm_feature' ran
    1.72 ± 0.19 times faster than 'poetry@installer install -q --no-root --only main -E tqdm_feature'
    2.73 ± 0.85 times faster than '/home/konsti/monotrail/target/release/monotrail poetry-install -E tqdm_feature'
    3.01 ± 0.34 times faster than '.venv/bin/pip install --no-compile -q -r requirements-benchmark.txt'
    4.96 ± 0.54 times faster than 'poetry install -q --no-root --only main -E tqdm_feature'
```

radoering:
> The code does compute hashes, but does not compile bytecode, right?

Good catch. It seems we did compare apples to oranges so far (at least to some degree). If I'm not mistaken, the "old" installer did compile bytecode but this one doesn't. I wasn't aware of that and I assume it's not intentional (but actually I don't know).

src/poetry/installation/wheel_installer.py
```diff
@@ -0,0 +1,101 @@
...
+import platform
+import sys
+
+from pathlib import Path
+from typing import TYPE_CHECKING
+
+from installer.destinations import SchemeDictionaryDestination  # type: ignore[import]
+from installer.sources import WheelFile  # type: ignore[import]
+
+from poetry import __version__
```
radoering, line 13:
```suggestion
from poetry.__version__ import __version__
```

I just noticed, that the version in the INSTALLER file is not printed correctly. We should fix this during the next rebase.

Secrus:
Two side-notes about dependencies: 
- `pep517` is now known as [`pyproject-hooks`](https://github.com/pypa/pyproject-hooks) and neither the old nor new version is defined in `pyproject.toml` as a top-level dependency (used in chef implementation)
- `installer` just got a new version [0.6.0](https://pypi.org/project/installer), which exports typing information so `type: ignore` comments should no longer be needed

radoering:
Rebased on current master and
* updated build and installer
* removed `type: ignore` comments no longer needed for installer
* replaced `pep517` usage by `pyproject-hooks` and added it as explicit dependency
* fixed version in INSTALLER file
* added tests for `WheelInstaller`
* added `--compile` option to `poetry install` (@konstin We decided to make bytecode compilation opt-in.)

Secrus:
#2288 is solved here too now.

src/poetry/installation/chef.py
radoering: I adapted it to use `ephemeral_environment`. Pulling `poetry.utils.pep517` in is a bit too much for me (at least for now). This should probably not block this PR and we can refactor it later in a follow-up PR.

pyproject.toml
```diff
@@ -50,21 +50,24 @@ python = "^3.7"
...
 filelock = "^3.8.0"
 html5lib = "^1.0"
 importlib-metadata = { version = ">=4.4,<6", python = "<3.10" }
+installer = "^0.6.0"
 jsonschema = "^4.10.0"
 keyring = "^23.9.0"
 lockfile = "^0.12.2"
 # packaging uses calver, so version is unclamped
-packaging = ">=20.4"
+packaging = ">=20.4,<22.0"  # TODO remove upper bound when poetry-core with vendored packaging 22.0 is available
```
dimbleby, line 66: why?

if it is necessary for poetry and poetry-core to agree on their packaging version then this ought to be an exact pin.  (I don't think this is necessary)

radoering: Without it `test_prepare_directory_with_extensions` fails on macOS:

```
 >       assert wheel.name == f"extended-0.1-{env.supported_tags[0]}.whl"
E       AssertionError: assert 'extended-0.1...16_x86_64.whl' == 'extended-0.1..._0_x86_64.whl'
E         - extended-0.1-cp39-cp39-macosx_12_0_x86_64.whl
E         ?                                ^ ^
E         + extended-0.1-cp39-cp39-macosx_10_16_x86_64.whl
E         ?                                ^ ^^
```

Maybe, it's just an issue of the test. I'm not sure if changing the test to not expect the first element but just any element of `supported_tags` makes sense?

dimbleby: Ah I see, poetry-core is building a platform-specific wheel using its packaging, and poetry tests are verifying the answer with a different packaging.

Meh, you're probably right just to do something pragmatic and not worry about it too much.

src/poetry/installation/chef.py
dimbleby: build 0.10 now published, should include your fix at https://github.com/pypa/build/pull/536

can we remove this lock altogether now?

radoering: Yes, I removed it.

radoering:
Rebased on current master and
* updated `build`
* removed the `threading.Lock` from `Chef` because it's not necessary anymore with the latest version of `build`
* moved setting from `experimental.installer.modern-installation` to `installer.modern-installation`
  (I think it's not good practice to set an experimental option to true by default. Either you call it experimental and don't enable it by default or you don't call it experimental. Further, it's a source of confusion that the current stable installer is `experimental.new-installer`.)

dimbleby:
What's the experience like in this branch when encountering a package that is not correctly pep-517 packaged and fails to build?

A fairly common issue on this tracker is people reporting such things as though it were poetry's fault.  

Currently the error looks like: poetry is calling out to pip, pip produces a big long error message.  pip's error message includes a note that "this is likely not a problem with pip" - but that certainly doesn't stop folk from coming here and reporting that they think there's a poetry problem.

This MR seems like a good opportunity to improve poetry's error presentation in this case, preferably also with a message indicating that this is likely not a problem in poetry but rather in the package being installed.

radoering:
Good point. It gave similar verbose output as pip but of course no note that this is probably not a problem with poetry. I added such a disclaimer. PTAL if the message can be improved.

src/poetry/installation/chef.py
```diff
@@ -2,28 +2,209 @@
...
+                    )
+            except BuildBackendException as e:
+                disclaimer = (
+                    "Note: This error originates from the build backend,"
+                    " and is likely not a problem with poetry"
+                    " but the concerning package not supporting PEP 517 builds."
+                    " You can verify this by running"
+                    " 'pip wheel --use-pep517 <concerning package>'."
+                )
+                message_parts = [disclaimer, str(e)]
```
dimbleby, line 147: Suggest that the disclaimer should come last, not first.  The build error message is quite likely to be a wall of text from gcc or whatever, some people will definitely not bother to scroll back, whatever shows up in the last few lines of a message gets more attention.

(pip agrees)
```
blah blah blah
...
  note: This error originates from a subprocess, and is likely not a problem with pip.
```

Also `concerning package` is awkward.  I realise that the package name isn't currently available to this code - but I'd suggest that it's worth rearranging things so that this error can be explicit about which package poetry is failing on (and pointing the finger at!)

src/poetry/installation/executor.py
```diff
@@ -298,6 +299,18 @@ def _execute_operation(self, operation: Operation) -> None:
...
+                    if isinstance(e, ChefBuildError):
+                        name = operation.package.pretty_name
+                        io.write_line("")
+                        io.write_line(
+                            "<info>"
+                            "Note: This error originates from the build backend,"
+                            " and is likely not a problem with poetry"
+                            f" but with {name} not supporting PEP 517 builds."
+                            " You can verify this by running"
+                            f" 'pip wheel --use-pep517 {name}'."
```
neersighted, line 311: This should probably also include the concrete version number if there's any way we can get at it.

radoering (approved):

PR #6210, opened 2022-08-22
feat!: remove deprecated CLI options

finswimmer:
This PR removes CLI options that were introduced during development of Poetry 1.2 but get replaced until today:

* `poetry plugin` --> use `poetry self add` instead
* `--default` option for group commands --> use `--with main` instead


neersighted (approved):

mkniewallner (approved):

mkniewallner (changes requested):
I guess we should also remove mentions of `--default` from the documentation in those places:

- `install`: https://github.com/python-poetry/poetry/blob/a3aafa840de950f81e99553e52190a54c94d6bce/docs/cli.md?plain=1#L230
- `update`: https://github.com/python-poetry/poetry/blob/a3aafa840de950f81e99553e52190a54c94d6bce/docs/cli.md?plain=1#L271
- `show`: https://github.com/python-poetry/poetry/blob/a3aafa840de950f81e99553e52190a54c94d6bce/docs/cli.md?plain=1#L474

finswimmer:
Excellent catch @mkniewallner :+1:

docs/cli.md
```diff
@@ -471,7 +469,6 @@ required by
 * `--why`: When showing the full list, or a `--tree` for a single package, display why a package is included.
 * `--with`: The optional dependency groups to include.
 * `--only`: The only dependency groups to include.
-* `--default`: Only include the main dependencies. (**Deprecated**)
 * `--no-dev`: Do not list the dev dependencies. (**Deprecated**)
```
Secrus, line 472: Should we remove that as well, since it also states it's deprecated?

mkniewallner (approved):

PR #6212, opened 2022-08-22
Adds --assume-yes to cache clear command to suppress prompts

colindean:
I had a perceived need to blow away my cache entirely and did so with:

    for cache in `poetry cache list`; do poetry cache clear $cache --all; done

but was annoyed by having to answer _yes_ for each of the _many_ caches I had. This PR adds an `--assume-yes|-y` command like `apt`'s that skips prompts.

Potential future work: Add a special syntax `poetry cache clear '*' --all --assume-yes` or `poetry cache implode` that will clear all entries in all caches.

# Pull Request Check List

### DIFF
--- a/docs/cli.md
+++ b/docs/cli.md
@@ -728,6 +728,8 @@ To only remove a specific package from a cache, you have to specify the cache en
 poetry cache clear pypi:requests:2.24.0
 ```
 
+To suppress prompts and proceed with the action without interaction, pass `--assume-yes` or `-y`.
+
 ## source
 
 The `source` namespace regroups sub commands to manage repository sources for a Poetry project.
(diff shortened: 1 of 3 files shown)


neersighted:
Hi @colindean, thanks for the PR! We actually already have a global `--no-interaction, -n` flag that is meant to do the exact same thing. It works by making every call to `confirm()` etc use the default value. It's broken for this command -- I discussed it with @Secrus a couple weeks ago, but I got side-tracked and never investigated personally.

Any chance you two could sync up and you could convert this PR to one that fixes the existing option?

Secrus:
I was summoned. 
Like @neersighted stated, we already have `--no-interaction` flag, coming for all commands from [`cleo`](https://github.com/python-poetry/cleo), our CLI framework. Current bug is that here: https://github.com/python-poetry/poetry/blob/0686427ce175e14b69c8ffeca4515cce79453047/src/poetry/console/commands/cache/clear.py#L58
you need to add `default=True` argument to `self.confirm` method call, for `cleo` to handle it properly as "if `--no-interactions`, agree". That's a small, easy fix.

colindean:
Thanks for #6338! I couldn't get around to acting until today and I'm glad the better solution is in, now.

Issue #6213, opened 2022-08-22
Improve consistency of Pool().remove_repository()


jre21:
- [x] I have searched the [issues](https://github.com/python-poetry/poetry/issues) of this repo and believe that this is not a duplicate.
- [x] I have searched the [documentation](https://python-poetry.org/docs/) and believe that my question is not covered.

## Feature Request
`Pool().remove_repository()` is currently difficult to use correctly in poetry plugins because it leaves internal variables in an inconsistent state.  This makes it difficult to correctly write plugins which can both add and remove repositories to/from poetry's pool.  I've submitted my desired changes as #6214

PR #6214, opened 2022-08-22
Ensure repository deletion is consistent

jre21:
This change improves the consistency of `Pool().remove_repository()` to make it easier to write poetry plugins which mutate the repository pool.

1.  Deleting an element from the middle of `Pool()._repositories` decrements the index of later entries.  Update `Pool()._lookup` to reflect this.

2.  If a primary repository is deleted, decrement `Pool()._secondary_start_idx` to ensure that any additional primary repositories are still added before all secondary repositories.

3.  If the default repository is deleted, reset `Pool()._default` so a new one can be added.

# Pull Request Check List

Resolves: #6213

- [ ] Updated **documentation** for changed code.  n/a: this change just updates an API to better match existing documentation

### DIFF
--- a/tests/repositories/test_pool.py
+++ b/tests/repositories/test_pool.py
@@ -102,9 +102,16 @@ def test_remove_default_repository():
     pool.add_repository(repo1)
     pool.add_repository(repo2)
     pool.add_repository(default, default=True)
+
+    assert pool.has_default()
+
     pool.remove_repository("default")
+
+    assert not pool.has_default()
+
     pool.add_repository(new_default, default=True)
 
+    assert pool.has_default()
     assert pool.repositories[0] is new_default
     assert not pool.has_repository("default")
 
--- a/tests/repositories/test_pool.py
+++ b/tests/repositories/test_pool.py
@@ -74,3 +74,64 @@ def test_repository_with_normal_default_and_secondary_repositories():
     assert pool.repository("foo") is repo1
     assert pool.repository("bar") is repo2
     assert pool.has_default()
+
+
+def test_remove_repository():
+    repo1 = LegacyRepository("foo", "https://foo.bar")
+    repo2 = LegacyRepository("bar", "https://bar.baz")
+    repo3 = LegacyRepository("baz", "https://baz.quux")
+
+    pool = Pool()
+    pool.add_repository(repo1)
+    pool.add_repository(repo2)
+    pool.add_repository(repo3)
+    pool.remove_repository("bar")
+
+    assert pool.repository("foo") is repo1
+    assert not pool.has_repository("bar")
+    assert pool.repository("baz") is repo3
+
+
+def test_remove_default_repository():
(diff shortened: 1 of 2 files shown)


jre21:
@neersighted Any chance this can get backported to the 1.2 branch, since it's a useful improvement to the new plugin functionality?

neersighted:
> @neersighted Any chance this can get backported to the 1.2 branch, since it's a useful improvement to the new plugin functionality?

We never accept feature backports, and 1.2 is reaching stabilization, so this will likely make it in 1.3. One could argue it's a correctness/soundness fix that deserves to make it into 1.2, but it's a higher risk change and I don't want to delay the 1.2 release unless @radoering and @dimbleby are confident in it.

jre21:
As far as i can tell, poetry itself never calls remove_repository(), so this would just be a change that affects plugins -- and i would definitely make the correctness/soundness argument since the function's existing behavior leaves the pool in an inconsistent state in many circumstances.

dimbleby:
It's understandable that folk would ask for backports given the enormous amount of time that it has taken to release 1.2 - that sets expectations about when the next release will happen.  But a much better thing would be for the project to shift to a mode where it released a _lot_ more frequently.

As it goes I agree that poetry itself never even calls the method being changed, so it is unlikely to break anything, so "OK fine" is probably a safe answer for this one.

Still I would much prefer the answer to be "no, but we promise that the next release will follow shortly"...!

neersighted:
The intention is definitely for frequent minor releases with patch releases only for major regressions that make the previous minor unusable compared to its predecessor version. Please don't get scared of 'wait for 1.3' as that is intended to be short -- 1.2 was only heroic to release because of the multi-year gap and massive code churn.

I'm still open for arguments for inclusion in 1.2 even if I agree with @dimbleby and would prefer to see this in 1.3.

tests/repositories/test_pool.py
```diff
@@ -74,3 +74,64 @@ def test_repository_with_normal_default_and_secondary_repositories():
...
+    new_default = LegacyRepository("new_default", "https://new.default.com")
+
+    pool = Pool()
+    pool.add_repository(repo1)
+    pool.add_repository(repo2)
+    pool.add_repository(default, default=True)
+    pool.remove_repository("default")
+    pool.add_repository(new_default, default=True)
+
+    assert pool.repositories[0] is new_default
```
radoering, line 108:
```suggestion
    pool.add_repository(default, default=True)

    assert pool.has_default()

    pool.remove_repository("default")

    assert not pool.has_default()

    pool.add_repository(new_default, default=True)

    assert pool.has_default()
    assert pool.repositories[0] is new_default
```

radoering:
Since it's a simple bugfix not adding new functionality, it's test coverage is quite good and it's not even used in poetry itself (lowest possible risk), I'd have no problem to accept a 1.2 backport after this PR is merged. However, I have no strong feelings about it.

radoering (approved):

PR #6218, opened 2022-08-23
docs: mention how to add a git dependency with a subdirectory

radoering:
Currently, git dependencies with subdirectories are only mentioned in https://python-poetry.org/docs/master/dependency-specification/#git-dependencies so you can see the syntax in pyproject.toml.

It might be useful to add an example how to `poetry add` such a dependency.

### DIFF
--- a/docs/cli.md
+++ b/docs/cli.md
@@ -348,7 +348,13 @@ poetry add git+ssh://github.com/sdispater/pendulum.git#develop
 poetry add git+ssh://github.com/sdispater/pendulum.git#2.0.5
 ```
 
-or make them point to a local directory or file:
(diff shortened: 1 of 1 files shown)


mkniewallner (approved):

mkniewallner:
I guess this is something that could be backported to `1.2`?

If so, since we will probably have multiple documentation updates that could end up being backported, we might want to cherry-pick what is of interest to `1.2` in one batch.

neersighted:
Let's bundle them up into one PR that maintainers can cherry-pick into -- doc updates are a good backport candidate as they require no subsequent RCs.

Issue #6222, opened 2022-08-23
1.2.0rc1 - `poetry env` - Current Python version conflict with pre-existing virtualenv


xkortex:
- [x] I am on the [latest](https://github.com/python-poetry/poetry/releases/latest) Poetry version.
- [x] I have searched the [issues](https://github.com/python-poetry/poetry/issues) of this repo and believe that this is not a duplicate.
- [x] If an exception occurs when executing a command, I executed it again in debug mode (`-vvv` option).
Possibly related: 
- https://github.com/python-poetry/poetry/issues/6051
- https://github.com/python-poetry/poetry/issues/6141
- https://github.com/python-poetry/poetry/issues/1735 (maybe, bit of a stretch, but I am having issues with poetry env)

- **OS version and name**: macOS catalina 10.15.7 (19H1922)
- **Poetry version**: 1.2.0rc.1
- **Link of a [Gist](https://gist.github.com/) with the contents of your pyproject.toml file**: 
```
[tool.poetry]
name = "test-project"
version = "0.1.0"
description = ""
authors = <redacted>
readme = "README.md"
packages = [{include = "test_project"}]

[tool.poetry.dependencies]
python = "^3.8"  # also tried this process with ">3.8,<3.9" to try to force it not to use 3.9

[build-system]
requires = ["poetry-core"]
build-backend = "poetry.core.masonry.api"
```
- **config**
```
❯ cat ~/Library/Preferences/pypoetry/config.toml
[virtualenvs]
create = false

[experimental]
new-installer = true

[installer]
parallel = true
```

## Issue
Apologies in advance if this is a PEBCAK, but I'm following the docs as closely as possible and it's not conforming with my expectations, so I think there is a mismatch somewhere. I've been experimenting with the 1.2 rc because I'm super excited for it, and running into some friction. I'm interested in running [poetry-version-plugin](https://github.com/tiangolo/poetry-version-plugin)

I've just installed poetry 1.2.0rc1 via  https://install.python-poetry.org (md5sum `6687215717deb5c6d8b86bbca2c043d8`)

```
❯ python3 ~/install-poetry.py --version 1.2.0rc1  # python3 points to ../Cellar/python@3.9/3.9.12/bin/python3
...
It will add the `poetry` command to Poetry's bin directory, located at:

/Users/mike/Library/Python/3.9/bin
...
❯ ln -svf /Users/mike/Library/Python/3.9/bin/poetry ~/.local/bin/poetry
[...]

PR #6225, opened 2022-08-24
fix: skip configure env for SelfCommand

finswimmer:
At the moment a venv is created on running a `poetry self` command. Even if it is not used in the command itself it's unnecessary and lead to problems as reported in #6222. This PR fixes this.

Resolves: #6222

### DIFF
--- a/src/poetry/console/application.py
+++ b/src/poetry/console/application.py
@@ -276,9 +276,10 @@ class Application(BaseApplication):  # type: ignore[misc]
         self, event: ConsoleCommandEvent, event_name: str, _: Any
     ) -> None:
         from poetry.console.commands.env_command import EnvCommand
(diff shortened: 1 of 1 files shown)


Secrus (approved):

dimbleby:
Shouldn't the `SelfCommand` instead be made not an `EnvCommand`?

finswimmer:
> Shouldn't the SelfCommand instead be made not an EnvCommand?

`SelfCommand` and its derived classes for the several `self` commands inherit `EnvCommand` to mirror the functionality from the command without `self`.

> Or perhaps the whole notion of `EnvCommand` wants abandoning and commands should instead get a `requires_venv(self) -> bool` or similar.

Yes, I had something similar in mind some time ago. But this would require a bigger refactoring. Let's do it at a later point.

PR #6237, opened 2022-08-25
repository: keep (uncanonicalized) pretty_name

radoering:
Lockfiles created with 1.2rc1 are not compatible with 1.1 if there is a dependency with a dot in its name (e.g. `discord.py`), because 1.1 applies a half-hearted normalization not converting dot to dash (writing `discord.py` in the lockfile) and 1.2rc1 does a complete normalization so that `discord.py` is normalized to `discord-py`, which is written to the lockfile.

However, when looking at

https://github.com/python-poetry/poetry/blob/4392bd6e7accf77a46d474a4767c9dd48da1ad89/src/poetry/packages/locker.py#L388

we can see that the intention has always been to write pretty name (instead of the canonicalized name) into the lockfile. Considering that, half-hearted normalization vs. complete normalization is not the actual issue.

The actual issue is that pretty name got lost unintionally in the process of looking up dependencies in repositories. This PR ensures that pretty name is not lost anymore.


dimbleby:
interesting

canonicalization still creeps in for extras somewhere, I haven't tried to figure out where.  Eg if you add `pyjwt` to a project you get these extras:
```toml
dev = ["coverage[toml] (==5.0.4)", "cryptography (>=3.3.1)", "mypy", "pre-commit", "pytest (>=6.0.0,<7.0.0)", "sphinx", "sphinx-rtd-theme", "zope-interface"]
docs = ["sphinx", "sphinx-rtd-theme", "zope-interface"]
```

although `pyjwt` declared a dependency on `zope.interface` (and that's what poetry 1.1 wrote in the lockfile)

src/poetry/packages/locker.py
```diff
@@ -412,10 +412,7 @@ def _dump_package(self, package: Package) -> dict[str, Any]:
             for name, deps in sorted(package.extras.items()):
                 # TODO: This should use dep.to_pep_508() once this is fixed
                 # https://github.com/python-poetry/poetry-core/pull/102
```
neersighted, line 414: Is this comment still relevant?

neersighted (dismissed):
I'm not the biggest fan of the inline canonicalization -- I'd rather see a variable to hold the canon form. That being said, LGTM even if you disagree.

src/poetry/repositories/repository.py
```diff
@@ -139,10 +140,10 @@ def find_links_for_package(self, package: Package) -> list[Link]:
         return []
 
     def package(
-        self, name: NormalizedName, version: Version, extras: list[str] | None = None
+        self, name: str, version: Version, extras: list[str] | None = None
     ) -> Package:
         for package in self.packages:
-            if name == package.name and package.version == version:
+            if canonicalize_name(name) == package.name and package.version == version:
```
dimbleby, line 146: this one wants pulling out of the loop, no need to canonicalize repeatedly

src/poetry/packages/locker.py
radoering: I will remove it. `to_pep_508()` appends markers that are probably not required in that place. At least not something like `extra == "foo"`

neersighted (approved):
LGTM

dimbleby (approved):

PR #6252, opened 2022-08-27
skip cloning badly defined submodules

dimbleby:
eg with this dependency
```
pyscf = { git = "https://github.com/pyscf/pyscf", tag = "v2.0.1"}
```
the cloning of submodules fails - because that project has messed up and not committed the file saying what revision they want of the `doc` submodule.

Legacy git client just silently carries on, so I've done the same here.

So far as I can tell this code currently does not have any test coverage, I must say that I don't feel particularly motivated to do anything about that: if anyone is more disciplined than I am then please have at it...

### DIFF
--- a/src/poetry/vcs/git/backend.py
+++ b/src/poetry/vcs/git/backend.py
@@ -334,7 +334,7 @@ class Git:
             url: bytes
             path: bytes
             submodules = parse_submodules(config)
-            for path, url, _ in submodules:
+            for path, url, name in submodules:
                 path_relative = Path(path.decode("utf-8"))
                 path_absolute = repo_root.joinpath(path_relative)
 
@@ -345,6 +345,12 @@ class Git:
                     try:
                         revision = repo.open_index()[path].sha.decode("utf-8")
                     except KeyError:
+                        logger.debug(
+                            "Skip submodule %s in %s, path %s not found",
+                            name,
+                            repo.path,
+                            path,
+                        )
                         continue
 
                 cls.clone(
--- a/src/poetry/vcs/git/backend.py
+++ b/src/poetry/vcs/git/backend.py
@@ -342,7 +342,10 @@ class Git:
                 source_root.mkdir(parents=True, exist_ok=True)
 
                 with repo:
-                    revision = repo.open_index()[path].sha.decode("utf-8")
+                    try:
+                        revision = repo.open_index()[path].sha.decode("utf-8")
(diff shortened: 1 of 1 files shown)


src/poetry/vcs/git/backend.py
```diff
@@ -342,7 +342,10 @@ def _clone_submodules(cls, repo: Repo) -> None:
                 source_root.mkdir(parents=True, exist_ok=True)
 
                 with repo:
-                    revision = repo.open_index()[path].sha.decode("utf-8")
+                    try:
+                        revision = repo.open_index()[path].sha.decode("utf-8")
+                    except KeyError:
```
Secrus, line 347: I'd add some `logger.debug()` message, just for ease of debugging this in the future.

radoering:
> So far as I can tell this code currently does not have any test coverage

Tests are in https://github.com/python-poetry/poetry/blob/master/tests/integration/test_utils_vcs_git.py. And there is even a test for cloning submodules. Maybe, we should move the tests. (I didn't find it right away either.) 🤔

dimbleby:
Those tests only use the legacy client.

dimbleby:
Aha, then IIUC this gets tested by merging https://github.com/python-poetry/test-fixture-vcs-repository/pull/1

dimbleby:
merge of https://github.com/python-poetry/test-fixture-vcs-repository/pull/1 probably requires this to be merged if the pipeline is ever to pass again

probably also forces some sort of action on 1.2.0?  

- simplest is to back-apply this, it really looks super-safe
- alternative would be to do something complicated where, I dunno, the tests on the 1.2.0 branch are changed so that they point at a version of the test-fixtures repository that doesn't include the bad module; or something like that

neersighted:
> merge of [python-poetry/test-fixture-vcs-repository#1](https://github.com/python-poetry/test-fixture-vcs-repository/pull/1) probably requires this to be merged if the pipeline is ever to pass again
> 
> probably also forces some sort of action on 1.2.0?
> 
>     * simplest is to back-apply this, it really looks super-safe
> 
>     * alternative would be to do something complicated where, I dunno, the tests on the 1.2.0 branch are changed so that they point at a version of the test-fixtures repository that doesn't include the bad module; or something like that

Go ahead and backport; I have no issue with pushing this to 1.2 for CI reasons.

neersighted (approved):

PR #6256, opened 2022-08-28
Misc fixes

dimbleby:
A collection of small clean-up commits that I've been accumulating, none of them interesting enough to be worth an MR alone - but perhaps I now have just about enough of them to be worthwhile.

No functional change or fixes here, just tidier code.

### DIFF
--- a/src/poetry/utils/extras.py
+++ b/src/poetry/utils/extras.py
@@ -17,7 +17,7 @@ def get_extra_package_names(
     packages: Sequence[Package],
     extras: Mapping[str, list[str]],
     extra_names: Sequence[str],
-) -> Iterable[str]:
+) -> Iterable[NormalizedName]:
     """
     Returns all package names required by the given extras.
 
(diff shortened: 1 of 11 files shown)


Secrus (approved):

src/poetry/utils/extras.py
```diff
@@ -4,20 +4,20 @@
...
 
 
 def get_extra_package_names(
-    packages: Sequence[Package],
+    packages: Iterable[Package],
     extras: Mapping[str, list[str]],
-    extra_names: Sequence[str],
-) -> Iterable[str]:
+    extra_names: Collection[str],
+) -> Iterable[NormalizedName]:
```
radoering, line 20: Maybe, we should remove the type names from the docstring:
"A collection of packages, such as ..." -> "Packages, such as ..."
"A list of strings ..." -> "Strings ..."

(I wonder if the param descriptions add any value. Maybe not for `packages`, but maybe for `extras`.)

dimbleby: I'm happy enough treating the types and the text as complementary documentation, it doesn't much bother me if the words are slightly imprecise - in the way that ordinary English often is - when we have the annotations making things explicit.

(But it wouldn't bother me if anyone did change this)

radoering (approved):

PR #6274, opened 2022-08-31
chore: drop tox

neersighted:
Drop tox as we have not made use of it in the development of Poetry for a long time.

Additionally, switch to the new `tool.poetry.group` syntax.


Secrus:
I am not sure about dropping tox. I has its use in local testing on multiple versions.

pyproject.toml
```diff
@@ -71,8 +71,7 @@ xattr = { version = "^0.9.7", markers = "sys_platform == 'darwin'" }
 urllib3 = "^1.26.0"
 dulwich = "^0.20.44"
 
-[tool.poetry.dev-dependencies]
-tox = "^3.18"
+[tool.poetry.group.test.dependencies]
```
Secrus, line 74: I don't like just changing labels like this. If we want to make use of groups, we should split the deps properly (mypy and pre-commit are not "test" deps)

Secrus (changes requested):

pyproject.toml
neersighted: That was a brainfart -- it was supposed to be dev

neersighted:
> I am not sure about dropping tox. I has its use in local testing on multiple versions.

Personally I find it to be of relatively low value as the matrix is wide with OS versions, so you really need to rely on CI to test widely. If we really are going to make it part of the repo tooling we need to maintain it, document it, and make it the default option for contributors -- multiple methods to run tests locally is confusing at best and harmful at worst.

Secrus (approved):

Issue #6275, opened 2022-08-31
exporting packages to stdout adds 'creating virtualenv ...' in v 1.2.0


jegger:
- [x] I have searched the [issues](https://github.com/python-poetry/poetry/issues) of this repo and believe that this is not a duplicate.
- [x] I have searched the [documentation](https://python-poetry.org/docs/) and believe that my question is not covered.

## Issue
Not sure if that should be considered a bug - maybe more an annoying change. However it broke our pipelines this morning as the new version 1.2.0 came out :)
I don't mind if you close this in case you think that it should be like that. Maybe it is helpful for others which use poetry in the same way we do.

We pipe the output of package export to pip for installation. 
E.g. `poetry export -f requirements.txt | poetry run pip install -r /dev/stdin`

In case there is not virtualenv created yet, poetry prints to stdout `Creating virtualenv foobar-pyXX in /foo/bar` which will result in pip trying to install a package called "Creating virtualenv...", which then ofc. fails.

### Workaround
As a workaround, we can simply export the dependencies to a file, and use pip to query the file instead of stdin.
`poetry export -f requirements.txt --output requirements.txt && poetry run pip install -r requirements.txt`

Thanks!

neersighted:
Likely related to the plugin migration. Going to tag in @finswimmer as he was doing work on environment commands (and implicit environment creation) recently.

dimbleby:
`-o requirements.txt` was always the contract: IMO this is indeed "annoying change" rather than "bug".

That `poetry export` creates a virtual environment is covered at https://github.com/python-poetry/poetry-plugin-export/issues/78.  You can run the command with `POETRY_VIRTUALENVS_CREATE=false` to avoid that, if you care to do so.

dimbleby:
Also `--quiet`, probably

alex:
FWIW `-o` doesn't seem to be compatible with using the results with pip's `-c`: https://github.com/pyca/cryptography/pull/7553 (the certbot-josepy job)

alex:
This was working under the previous version of poetry with the ` > constraints.txt` operation. This thread suggests that the redirect was never supported, and `-o` should be used. Perhaps `constraints.txt` was never supported in either of them, but just happened to work?

dimbleby:
I don't know how to be clearer so I guess I'll just repeat myself.  The output format is (and always was) `requirements.txt`, not `constraints.txt`.

adferrand:
[...]

PR #6282, opened 2022-08-31
not every GroupCommand needs to be an EnvCommand

dimbleby:
Relating to #6275 and https://github.com/python-poetry/poetry-plugin-export/issues/78

Not every `GroupCommand` needs to be an `EnvCommand` (though it's true that every `GroupCommand` in this repository is also an `EnvCommand`, so this refactoring changes nothing here).

The intention is to make a change to the export plugin so that the `ExportCommand` can be a `GroupCommand` but not an `EnvCommand`: it does need access to the group options, but it does not need a virtual environment.

### DIFF
--- a/src/poetry/console/commands/group_command.py
+++ b/src/poetry/console/commands/group_command.py
@@ -5,7 +5,7 @@ from typing import TYPE_CHECKING
 from cleo.helpers import option
 from poetry.core.packages.dependency_group import MAIN_GROUP
 
-from poetry.console.commands.env_command import EnvCommand
+from poetry.console.commands.command import Command
 
 
 if TYPE_CHECKING:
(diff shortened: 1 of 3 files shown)


dimbleby:
Arguably not every `InstallerCommand` should need to be an `EnvCommand` eg `SelfCommand` shouldn't, per #6225.  Perhaps an improved version of that fix would be to take `EnvCommand` off the `InstallerCommand` and add it back only on the commands that actually need it.

Another day, maybe...

dimbleby:
actually let's do that while we're here, you can always reject the second commit if it's too much

dimbleby:
ugh no, the `SelfCommand` is messier than I thought, let's just do the one thing after all.

neersighted (approved):

PR #6287, opened 2022-08-31
The old script installs the old poetry instead of breaking

jonapich:
Instead of crashing, the old script will fallback to a compatible version and issue a warning.


jonapich:
It is my wish that one of the maintainers include the correct wording/colors/style and url.

PR #6290, opened 2022-08-31
Update type hints of `register_factory` to be compatible with examples

kkirsche:
The example:
https://python-poetry.org/docs/plugins/#application-plugins

Creates a `cleo.commands.command` from the factory, which is not compatible with the type `poetry.console.commands.command`. This returns a union type to address this, assuming this isn't an error in the documentation.

Affects: Poetry 1.2.x

# Pull Request Check List

Resolves: #issue-number-here

### DIFF
--- a/src/poetry/console/command_loader.py
+++ b/src/poetry/console/command_loader.py
@@ -14,6 +14,7 @@ if TYPE_CHECKING:
 
     Command = Union[CleoCommand, PoetryCommand]
 
+
 class CommandLoader(FactoryCommandLoader):  # type: ignore[misc]
     def register_factory(
         self, command_name: str, factory: Callable[[], Command]
--- a/src/poetry/console/command_loader.py
+++ b/src/poetry/console/command_loader.py
@@ -8,7 +8,6 @@ from cleo.loaders.factory_command_loader import FactoryCommandLoader
 
 if TYPE_CHECKING:
     from collections.abc import Callable
-    from typing import Union
 
     from cleo.commands.command import Command
 
--- a/src/poetry/console/command_loader.py
+++ b/src/poetry/console/command_loader.py
@@ -10,11 +10,7 @@ if TYPE_CHECKING:
     from collections.abc import Callable
     from typing import Union
 
-    from cleo.commands.command import Command as CleoCommand
-
-    from poetry.console.commands.command import Command as PoetryCommand
-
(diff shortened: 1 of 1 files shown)


kkirsche:
Test error appears to be unrelated to this change:

```
Error: test_info_setup_missing_mandatory_should_trigger_pep517[version]

assert 3 == 2
 +  where 3 = <function run at 0x7fc2211dfc10>.call_count
```

dimbleby:
I think the fix you want is simpler - won't this do the job?

```patch
-    from poetry.console.commands.command import Command
+    from cleo.commands.command import Command
```

kkirsche:
It will fix the error, but as the Poetry team has chosen not to use docstrings to capture the intent of the registration function, I can't say with any confidence which is right — the documentation or the code, so the types were updated to include both.

If the team can address the lack of clarity, I can update it accordingly.

dimbleby:
pretty sure I added the original type annotation myself and if so I'm confident it was just carelessness that caused me to choose the  more specific type when the more general would have done.

Your variation is I think a strictly worse version: the code is still required to accept a cleo Command because of the union, and since the poetry Command _is_ a cleo Command, the union is just a noisier way of saying the same thing.

kkirsche:
> pretty sure I added the original type annotation myself and if so I'm confident it was just carelessness that caused me to choose the more specific type when the more general would have done.
> 
> Your variation is I think a strictly worse version: the code is still required to accept a cleo Command because of the union, and since the poetry Command _is_ a cleo Command, the union is just a noisier way of saying the same thing.

No worries, I'll update it then. Though this reads as an aggressive response to what simply is a lack of clarity

neersighted:
> > pretty sure I added the original type annotation myself and if so I'm confident it was just carelessness that caused me to choose the more specific type when the more general would have done.
> > Your variation is I think a strictly worse version: the code is still required to accept a cleo Command because of the union, and since the poetry Command _is_ a cleo Command, the union is just a noisier way of saying the same thing.
> 
> No worries, I'll update it then. Though this reads as an aggressive response to what simply is a lack of clarity

I think that might just be the lack of inflection in text -- it reads as a pretty dry technical evaluation to me. Thanks for this typing fix!

neersighted (approved):

kkirsche:
Probably, wasn't meant as an accusation, simply context that I misinterpreted it. Thanks for everyone's time.

PR #6304, opened 2022-08-31
B023 is real

dimbleby:
Here's some sample code approximating the poetry code:
```python
#!/usr/bin/env python3


def one() -> int:
    return 1


def two() -> int:
    return 2


def three() -> int:
    return 3


factories = [one, two, three]

registered = []
for factory in factories:
    registered.append(lambda: factory())

for registered_factory in registered:
    print(f"result is {registered_factory()}")
```
output is
```
result is 3
result is 3
result is 3
```
which is exactly the gotcha that flake8-bugbear's B023 was trying to warn about.

I've applied one of the workarounds that various parts of the internet recommend, and you can verify for yourself if you're so inclined that doing the same in the toy script gives the expected output.

### DIFF
--- a/src/poetry/plugins/application_plugin.py
+++ b/src/poetry/plugins/application_plugin.py
@@ -1,5 +1,7 @@
 from __future__ import annotations
 
+import functools
+
 from typing import TYPE_CHECKING
 
 from poetry.plugins.base_plugin import BasePlugin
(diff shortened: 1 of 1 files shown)


dimbleby:
Although come to think of it this particular case could surely be more compactly fixed by not lambda-ing at all, `register_factory(command.name, command)` should do the job

dimbleby:
I've pushed the more compact fix in a second commit so that this discussion, such as it is, still makes sense.  Feel free to squash if and when merging.

neersighted (approved):

neersighted (approved):

PR #6308, opened 2022-08-31
Support venv/src/ not being a valid git dir

Alphadelta14:
# Pull Request Check List

After updating to latest 1.2, we ran into an issue where we had a lot of pre-existing /venv/src directories that we kept license metadata in for distribution which were suddenly being parsed by Poetry as git repos (and obviously failing).

This just adds a simple lookahead to make sure that we aren't trying to create git repositories where they don't belong.

We will be moving our license files into /venv/share/doc in the future, but I can't imagine we were the only folks to have run into this.


abn:
This is no longer an issue. See https://github.com/python-poetry/poetry/issues/6322#issuecomment-2599133247.

PR #6309, opened 2022-08-31
Reflect 1.2 in README and CI

mkniewallner:
# Pull Request Check List

Update README to showcase dependency groups and update CI to use `--only main` instead of `--no-dev` for docs. Also remove obsolete matrix options in main workflow.

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -51,7 +51,7 @@ jobs:
       - name: Fetch Documentation
         run: |
           python -m pip install poetry
-          poetry install --no-dev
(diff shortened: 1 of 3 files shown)


mkniewallner:
Unfortunately, because of [this](https://github.com/python-poetry/poetry/blob/7fcb6439815a458884f55fb807121a3463cf4c64/.github/workflows/docs.yml#L9-L17), it looks like it's not possible to run documentation preview workflow if the workflow is updated from a fork.

neersighted (approved):

PR #6315, opened 2022-09-01
ci: add 3.11 to the Python matrix

neersighted:
*- ci: use 3.10 in release workflow
- ci: add 3.11 to the Python matrix


neersighted:
After discussion we're going to test against RCs again.

Secrus (approved):

PR #6318, opened 2022-09-01
docs: fix example for --with option combined with --without

morpser:
The current "Managing Dependencies" documentation contains an error at
[Installing group dependencies](https://python-poetry.org/docs/managing-dependencies/#installing-group-dependencies)

Within the warning concerning the combined use of  `--with` and `--without` the parameters were swapped.

- [x] Updated **documentation**

### DIFF
--- a/docs/managing-dependencies.md
+++ b/docs/managing-dependencies.md
@@ -142,7 +142,7 @@ poetry install --with docs
 
 {{% warning %}}
 When used together, `--without` takes precedence over `--with`. For example, the following command
-will only install the dependencies specified in the `test` group.
+will only install the dependencies specified in the optional `test` group.
 
 ```bash
 poetry install --with test,docs --without docs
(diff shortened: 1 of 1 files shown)


docs/managing-dependencies.md
```diff
@@ -145,7 +145,7 @@ When used together, `--without` takes precedence over `--with`. For example, the
 will only install the dependencies specified in the `test` group.
```
finswimmer, line 145: I would suggest to make clear that the `test` group in this example must be an optional group. Maybe like this:

```suggestion
will only install the dependencies specified in the optional `test` group.
```

finswimmer (changes requested):
Good catch :+1: 

I added one suggestion for further improvement of this section.

finswimmer (approved):
LGTM :1st_place_medal:

PR #6323, opened 2022-09-01
Exit with non-zero code on error

baryluk:


baryluk:
mac ci failure looks to be totally unrelated.

neersighted:
Thanks for the PR @baryluk -- this sat for a bit because there's been ongoing discussion about changes in the `get-poetry.py` script -- I've rolled up your changes into #6378 and will close this when/if we merge that PR.

PR #6324, opened 2022-09-01
Add no interaction flag to 1.1 documentation

ZhiliangWu:
A small PR to update documentation of poetry 1.1 regarding the global option `-n`

### DIFF
--- a/docs/cli.md
+++ b/docs/cli.md
@@ -25,6 +25,7 @@ then `--help` combined with any of those can give you more information.
 * `--ansi`: Force ANSI output.
 * `--no-ansi`: Disable ANSI output.
 * `--version (-V)`: Display this application version.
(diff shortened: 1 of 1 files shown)


neersighted:
Hi @ZhiliangWu! Thanks for contributing! Unfortunately the 1.1 branch is no longer accepting contributions, and is scheduled to be removed from this repo. While technically we could accept docs commits without making a new release, it would both send the wrong message, and lead to lost commits when we eventually drop the branch. We are doing our best to triage and address regressions in 1.2, and the stable release is expected to improve rapidly as we triage and fix issues with it. If you can switch to 1.2 you should, and if you experience regressions you should do your best to create a high-quality reproduction so our team of volunteers can address it.

PR #6325, opened 2022-09-01
fix: publisher reported successfully finding credentials for PyPI even when they did not exist

neersighted:
Minor cleanup in the authentication code to prevent spuriously reporting finding PyPI credentials due to a special case.
It's also possible to return `None` in the PyPI special case itself -- suggestions are welcome.


neersighted:
Converting to draft as this is harder to do cleanly than I hoped -- a more comprehensive refactor is in order

Issue #6335, opened 2022-09-01
`poetry lock --no-update` updates dependencies if custom source specified


maxim-krikun:
- [x] I am on the [latest](https://github.com/python-poetry/poetry/releases/latest) Poetry version.
- [x] I have searched the [issues](https://github.com/python-poetry/poetry/issues) of this repo and believe that this is not a duplicate.
- [x] If an exception occurs when executing a command, I executed it again in debug mode (`-vvv` option).

- **OS version and name**: Windows 10 (21H2)
- **Poetry version**: 1.2.0
- **Link of a [Gist](https://gist.github.com/maxim-krikun/ad33f3abbf8e8dab09246c03648ee843) with the contents of your pyproject.toml file**: 

## Issue

Using Poetry 1.1.14 and lock file `poetry_1.1.14.lock`:

```
❯ poetry lock --no-update
Resolving dependencies...
```

The lock file remains unchanged. Then upgrade Poetry up to 1.2.0.
Using Poetry 1.2.0 and lock file `poetry_1.1.14.lock`:

```
❯ poetry lock --check
poetry.lock is consistent with pyproject.toml.
❯ poetry lock --no-update
Resolving dependencies...

Writing lock file
```

It produces a new lock file `poetry_1.2.0.lock`.
As you can see some versions of the dependencies have been updated.
That is, the changes were not only cosmetic.
I'd also like to note:
- if you don't use a custom source (just remove source block from `pyproject.toml` and repeat repro steps), everything works as it should;
- command for checking lock file returned a positive result. So there must be a bug somewhere, either in the check or in the locking.

dimbleby:
https://github.com/python-poetry/poetry/blob/a613347bf0a9fd7b7ec782cd45222b3a0f0e7401/src/poetry/puzzle/provider.py#L846 I think is the line that refuses to acknowledge that the locked package is the same as the dependency - because the dependency has no `source_type` or `source_url`, whereas the locked package does.

poetry 1.1 used to ask a slightly different question: https://github.com/python-poetry/poetry/blob/46bf7fd3d1797b75ad9b07adc280039d93414e07/poetry/mixology/version_solver.py#L469.  

I think that it is https://github.com/python-poetry/poetry-core/pull/325 which introduced the changed behaviour, though refactorings since then have made the trail not quite straightforward to follow.

I guess we need a more delicate check on the source type here.

dimbleby:
[...]

PR #6336, opened 2022-09-01
No unwanted update of packages from legacy source

dimbleby:
If a dependency doesn't specify a source type, then a locked package
from any source can satisfy it

# Pull Request Check List

Resolves: #6335

I'm pretty sure this is reinstating the earlier behaviour - see https://github.com/python-poetry/poetry/issues/6335#issuecomment-1234666240 for analysis


dimbleby:
(test failure looks like noise)

radoering (approved):

PR #6345, opened 2022-09-02
docs: added ci recommendations

finswimmer:
Added some recommendations about how to install Poetry into an CI environment.

### DIFF
--- a/docs/_index.md
+++ b/docs/_index.md
@@ -28,7 +28,7 @@ If you are viewing documentation for the development branch, you may wish to ins
 See the **advanced** installation instructions to use a preview or alternate version of Poetry.
 {{% /note %}}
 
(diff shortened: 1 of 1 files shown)


docs/_index.md
```diff
@@ -274,10 +274,82 @@ Poetry will be available at `$VENV_PATH/bin/poetry` and can be invoked directly
...
+```
+virtualenv /opt/poetry
+/opt/poetry/bin/pip install poetry==1.2.0
+/opt/poetry/bin/poetry --version
+```
+
+{{% note %}}
+If you install Poetry via `pip`, ensure you have Poetry installed into an isolated environment that is **not the same**
+as the target environment managed by Poetry. If Poetry and your project are installed into the same environment, Poetry
+is likely to upgrade or uninstall its down dependencies (causing hard-to-debug and understand errors).
```
neersighted, line 330:
```suggestion
is likely to upgrade or uninstall its own dependencies (causing hard-to-debug and understand errors).
```

neersighted (approved):

Issue #2412, opened 2020-05-13
Install from git may be slow when cloning huge repository


Inndy:
- [x] I have searched the [issues](https://github.com/python-poetry/poetry/issues) of this repo and believe that this is not a duplicate.
- [x] I have searched the [documentation](https://python-poetry.org/docs/) and believe that my question is not covered.

## Feature Request

Install from git should not clone all commit history. Can we just make a shallow clone instead?

markopy:
+1 for this.

It makes sense to just check out the branch actually needed and without all the history. This can reduce the cloning time enormously for large repositories with a lot of history.

This seem like it would be a fairly simple change with something like this added to the git invoke:
```
--branch <rev> --single-branch --depth 1
```

Kache:
Ran into this exact issue here: https://github.com/python-poetry/poetry/issues/2094#issuecomment-1203314891

gaborfekete-greehill:
I had a similar issue when a branch used submodules and the other not, and the non-default would have been the preferred one: #9612

PR #6348, opened 2022-09-02
feat: use shallow git clone and shallow submodule clone

roelofvandijk:
# Pull Request Check List

Resolves: #2412
Relates to: #2094 

This should speed up `git clone` for all large repositories.

I am unsure what extra tests would be required for this.
I am also not sure if there is any case in which a non-shallow clone of a repository would be required, for instance if tags are needed.

Note that `--depth 1` automatically implies `--single-branch` per the [git docs](https://git-scm.com/docs/git-clone#Documentation/git-clone.txt---depthltdepthgt).

### DIFF
--- a/docs/managing-dependencies.md
+++ b/docs/managing-dependencies.md
@@ -142,10 +142,10 @@ poetry install --with docs
 
 {{% warning %}}
 When used together, `--without` takes precedence over `--with`. For example, the following command
-will only install the dependencies specified in the `test` group.
+will only install the dependencies specified in the optional `test` group.
 
 ```bash
-poetry install --with docs --without test,docs
+poetry install --with test,docs --without docs
 ```
 {{% /warning %}}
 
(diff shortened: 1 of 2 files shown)


roelofvandijk:
It seems like the `--depth` parameter is not recognized by the `git` command that is used in the tests. Is this a (very) old version of git? That could also cause some slowness for git clone.

neersighted:
I'm not 100% sure this is something universally desirable to apply, and it will be need to done to Dulwich as well as the system git client. There should also be some fallback if it is not supported.

mtkennerly:
I have [a plugin](https://github.com/mtkennerly/poetry-dynamic-versioning) that does VCS-based dynamic versioning, and this would break the handling for dependencies that also use the plugin. If you make this the default, please also add an option for Git dependencies to specify a custom depth, including the full history. For example:

```toml
[tool.poetry.dependencies]
requests = { git = "https://github.com/requests/requests.git", depth = 0 }
```

PR #6372, opened 2022-09-02
Simplify extras handling

dimbleby:
Another random bit of code-tidying:

- remove some dead code that constructs an unused dictionary
- simplify (considerably) the code that walks the dependency tree looking for packages introduced by extras

I think there's a bug in that code actually, in the (unlikely) case that a package appears twice in the tree (eg at different versions or with different sources) and the different versions of the packages would have given different answers.  I don't plan to do anything about this.


LarsDu:
I was looking through this and super puzzled by the unused dicts as well! Kudos

radoering (approved):

PR #6376, opened 2022-09-02
fix virtualenv constraint

dimbleby:
https://github.com/python-poetry/poetry-core/pull/461 reveals that poetry's own `pyproject.toml` contained an invalid constraint:
```
Could not parse version constraint: (>=20.4.3
```
oops.

I still think that poetry-core fix is good and desirable, but if this is an indication of the likely influx of issues... well that's something to think about.

Anyway this MR is clearly good, let's fix the error here.

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -931,7 +931,7 @@ testing = ["func-timeout", "jaraco.itertools", "pytest (>=6)", "pytest-black (>=
 [metadata]
 lock-version = "1.1"
 python-versions = "^3.7"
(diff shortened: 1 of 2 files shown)


dimbleby:
just to confirm:
```
curl -s https://pypi.org/pypi/poetry/1.2.0/json | jq ".info.requires_dist[3]"
"virtualenv"
```
ie poetry's metadata is asking for any version of virtualenv at all, not the constrained version that it had thought it was asking for

neersighted (approved):

Issue #6349, opened 2022-09-02
Poetry does not add all hashes for multiple constraint dependences


intgr:
- [x] I am on the [latest](https://github.com/python-poetry/poetry/releases/latest) Poetry version.
- [x] I have searched the [issues](https://github.com/python-poetry/poetry/issues) of this repo and believe that this is not a duplicate.
- [x] If an exception occurs when executing a command, I executed it again in debug mode (`-vvv` option).

- **OS version and name**: Linux, win32
- **Poetry version**: 1.2.0
- **Link of a [Gist](https://gist.github.com/) with the contents of your pyproject.toml file**: https://gist.github.com/intgr/efc96c44c3919189c3757275f0239eb6

## Issue
The same example as in #6327. If package to be installed depends on the OS:
```toml
python-ldap = [
  { platform = "linux", version = "3.4.0" },
  { platform = "win32", url = "https://download.lfd.uci.edu/pythonlibs/archived/python_ldap-3.4.0-cp310-cp310-win_amd64.whl" },
]
```

Running `poetry lock` on Linux only adds hashes for the Linux version of the package. poetry.lock only contains:
```toml
python-ldap = [
    {file = "python-ldap-3.4.0.tar.gz", hash = "sha256:60464c8fc25e71e0fd40449a24eae482dcd0fb7fcf823e7de627a6525b3e0d12"},
]
```

And this is despite the fact that on Linux, `poetry lock` every time re-downloads the Windows package again.
```
Resolving dependencies... Downloading https://download.lfd.uci.edu/pythonlibs/archived/python_ldap-3.4.0-cp310-cp310-win_amd64.whl   9%
```

Using this lockfile, `poetry install` on Windows causes the following error (note, this has different version numbers, I don't have Windows myself to reproduce cleanly)
```
 • Updating python-ldap (3.3.1 .../AppData/Local/pypoetry/Cache/artifacts/df/70f0/b890a4118f494cdc75350af78a5bd590e52b9edd2788ca3e01001d1ef1/python_ldap-3.3.1-cp310-cp310-win_amd64.whl -> 3.3.1 .../python_ldap-3.3.1-cp310-cp310-win_amd64.whl)

 RuntimeError
[...]

Issue #6327, opened 2022-09-01
poetry lock adds duplicate hashes when using multiple constraint dependences


intgr:
- [x] I am on the [latest](https://github.com/python-poetry/poetry/releases/latest) Poetry version.
- [x] I have searched the [issues](https://github.com/python-poetry/poetry/issues) of this repo and believe that this is not a duplicate.
- [x] If an exception occurs when executing a command, I executed it again in debug mode (`-vvv` option).

- **OS version and name**: macOS
- **Poetry version**: 1.2.0
- **Link of a [Gist](https://gist.github.com/) with the contents of your pyproject.toml file**: https://gist.github.com/intgr/efc96c44c3919189c3757275f0239eb6

## Issue

With a dependency specification like
```toml
python-ldap = [
  { platform = "linux", version = "3.4.0" },
  { platform = "win32", url = "https://download.lfd.uci.edu/pythonlibs/archived/python_ldap-3.4.0-cp310-cp310-win_amd64.whl" },
]
```

Every time I run `poetry lock --no-update`, Poetry adds duplicated hashes for `python-ldap` to the `poetry.lock` file.

After the first lock command, the content ends with:
```toml
python-ldap = [
    {file = "python-ldap-3.4.0.tar.gz", hash = "sha256:60464c8fc25e71e0fd40449a24eae482dcd0fb7fcf823e7de627a6525b3e0d12"},
]
```
After the second run, it's:
```toml
python-ldap = [
    {file = "python-ldap-3.4.0.tar.gz", hash = "sha256:60464c8fc25e71e0fd40449a24eae482dcd0fb7fcf823e7de627a6525b3e0d12"},
    {file = "python-ldap-3.4.0.tar.gz", hash = "sha256:60464c8fc25e71e0fd40449a24eae482dcd0fb7fcf823e7de627a6525b3e0d12"},
]
```
Etc, every run adds yet another entry there.

The contents of the lockfile should be deterministic.

radoering:
All locked filenames/hashes are duplicated once for each direct origin dependency among the multiple constraints dependencies. This issue doesn't occur in poetry 1.1 because direct origin dependencies are always updated (even with `--no-update` flag). I created a PR to fix this regression.

intgr:
Thanks! I can confirm that this is fixed in Poetry master branch.

PR #6389, opened 2022-09-03
locker: do not assign unrelated filenames/hashes to direct origin dependencies

radoering:
Fix for duplicate (or outdated) hashes of multiple constraint dependencies

Resolves: #6327
Resolves: #6349

Fixes a regression in `poetry lock --no-update` where filenames and hashes of multiple constraints dependencies are always kept/duplicated.

Background: There is only one `metadata.files` entry for all multiple constraints dependencies in poetry.lock. When reading the lock file we can't always decide for sure which filename/hash belongs to which of the multiple constraints dependencies so we just add all filenames/hashes for all packages. For non-direct origin dependencies, this is fixed later in `complete_package()` when the package is looked up in the repository. For direct origin dependencies this isn't fixed, so all entries are kept and duplicated once for each direct origin dependency. Even if one of the multiple constraints dependencies has been deleted from pyproject.toml, its filenames/hashes are kept if there is still one direct origin dependency.

(This issue did not occur in poetry 1.1.x because poetry was not able to recognize if a direct origin dependency had already been included in the lockfile and thus always updated it when running `poetry lock --no-update`.)

Update: Another manifestation of the fixed bug is that url, git or directory dependencies of multiple constraints dependencies can't be installed if there is a non-direct origin or file dependency among the multiple constraints dependencies.

### DIFF
--- a/src/poetry/packages/locker.py
+++ b/src/poetry/packages/locker.py
@@ -128,9 +128,18 @@ class Locker:
                 # Old lock so we create dummy files from the hashes
                 hashes = cast("dict[str, Any]", metadata["hashes"])
                 package.files = [{"name": h, "hash": h} for h in hashes[name]]
+            elif source_type in {"git", "directory", "url"}:
+                package.files = []
             else:
                 files = metadata["files"][name]
-                package.files = files
+                if source_type == "file":
+                    filename = Path(url).name
+                    package.files = [item for item in files if item["file"] == filename]
+                else:
+                    # Strictly speaking, this is not correct, but we have no chance
+                    # to always determine which are the correct files because the
+                    # lockfile doesn't keep track which files belong to which package.
+                    package.files = files
 
             package.python_versions = info["python-versions"]
             extras = info.get("extras", {})
--- a/tests/packages/test_locker.py
+++ b/tests/packages/test_locker.py
@@ -409,6 +409,120 @@ git-package-subdir = []
     assert package.source_subdirectory == "subdir"
 
 
+def test_locker_properly_assigns_metadata_files(locker: Locker) -> None:
+    """
+    For multiple constraints dependencies, there is only one common entry in
+    metadata.files. However, we must not assign all the files to each of the packages
+    because this can result in duplicated and outdated entries when running
+    `poetry lock --no-update` and hash check failures when running `poetry install`.
+    """
+    content = """\
+[[package]]
+name = "demo"
+version = "1.0"
+description = ""
+category = "main"
+optional = false
+python-versions = "*"
+develop = false
+
+[[package]]
+name = "demo"
+version = "1.0"
+description = ""
+category = "main"
+optional = false
+python-versions = "*"
+develop = false
+
+[package.source]
+type = "git"
+url = "https://github.com/demo/demo.git"
+reference = "main"
+resolved_reference = "123456"
+
+[[package]]
+name = "demo"
+version = "1.0"
+description = ""
+category = "main"
+optional = false
+python-versions = "*"
+develop = false
+
+[package.source]
+type = "directory"
+url = "./folder"
+
+[[package]]
+name = "demo"
+version = "1.0"
+description = ""
+category = "main"
+optional = false
+python-versions = "*"
+develop = false
+
+[package.source]
+type = "file"
+url = "./demo-1.0-cp39-win_amd64.whl"
+
+[[package]]
+name = "demo"
+version = "1.0"
+description = ""
+category = "main"
+optional = false
+python-versions = "*"
+develop = false
+
+[package.source]
+type = "url"
+url = "https://example.com/demo-1.0-cp38-win_amd64.whl"
+
+[metadata]
+lock-version = "1.1"
+python-versions = "*"
+content-hash = "115cf985d932e9bf5f540555bbdd75decbb62cac81e399375fc19f6277f8c1d8"
+
+[metadata.files]
+# metadata.files are only tracked for non-direct origin and file dependencies
+demo = [
+    {file = "demo-1.0-cp39-win_amd64.whl", hash = "sha256"},
+    {file = "demo-1.0.tar.gz", hash = "sha256"},
+    {file = "demo-1.0-py3-none-any.whl", hash = "sha256"},
+]
+"""
+    locker.lock.write(tomlkit.parse(content))
+
+    repository = locker.locked_repository()
+    assert len(repository.packages) == 5
+    assert {package.source_type for package in repository.packages} == {
+        None,
+        "git",
+        "directory",
+        "file",
+        "url",
+    }
+    for package in repository.packages:
+        if package.source_type is None:
+            # non-direct origin package contains all files
+            # with the current lockfile format we have no chance to determine
+            # which files are correct, so we keep all for hash check
+            # correct files are set later in Provider.complete_package()
+            assert package.files == [
+                {"file": "demo-1.0-cp39-win_amd64.whl", "hash": "sha256"},
+                {"file": "demo-1.0.tar.gz", "hash": "sha256"},
+                {"file": "demo-1.0-py3-none-any.whl", "hash": "sha256"},
+            ]
+        elif package.source_type == "file":
+            assert package.files == [
+                {"file": "demo-1.0-cp39-win_amd64.whl", "hash": "sha256"}
+            ]
+        else:
+            package.files = []
+
+
 def test_lock_packages_with_null_description(locker: Locker, root: ProjectPackage):
     package_a = get_package("A", "1.0.0")
     package_a.description = None


src/poetry/puzzle/provider.py
```diff
@@ -546,35 +546,53 @@ def complete_package(
...
+                            package.version,
+                            extras=list(dependency.extras),
+                            repository=dependency.source_name,
+                        ),
                     )
-                except StopIteration:
-                    raise e from e
+                except PackageNotFound as e:
+                    try:
+                        dependency_package = next(
```
neersighted, line 587: I know this is a generator expression and thus evaluated lazily, but it seems more straightforward to construct this such that the `next()` is called on `self.search_for_installed_packages` and we make the `DependencyPackage` from the result of that call.

Also, is there a reason `pkg.to_dependency()` wouldn't work here?

src/poetry/puzzle/provider.py
```diff
@@ -546,35 +546,53 @@ def complete_package(
...
-                        repository=dependency.source_name,
-                    ),
-                )
-            except PackageNotFound as e:
+        else:
+            if package.source_type == "file":
+                if len(package.files) > 1:
+                    # a locked package might contain further files
+                    # for multiple-constraints dependencies -> prune
+                    path = urllib.parse.urlsplit(package.source_url).path
```
neersighted, line 554: I'm guessing `urllib` isn't properly typed?

neersighted (changes requested):

dimbleby:
isn't the real problem that all the files in the metadata in poetry.lock get attached to all packages?

What if we just... don't do that?  ie remove the code in the locker that sets `packages.files` from metadata.  Aren't the correct values going to be picked up later anyway?

(`complete_package()` is one of the biggest and ugliest methods in the codebase, this... isn't helping that!)

dimbleby:
hmm, ignoring the metadata might work during solving but it also means that we don't do any checking of hashes during install.

Well OK then, how about fixing the broken file format instead?  We've learned that the files belong on the packages, let's write that in the lockfile?

Should be possible to do in a way that can still read old lockfiles: if the lockfile contains per-package files then use them, else fall back to the merged metadata

neersighted:
If we're talking about changing the lockfile, that would have to go into 1.3. This as a correctness fix for a regression would be eligible for 1.2... So maybe the correct solution is both?

I defer to you both for your expertise on the moving pieces of the solver -- my knowledge is still a combination of surface level and function level.

dimbleby:
why would it have to go into 1.3, if the new code is capable of reading the old lockfile?

because it's a "large" change?  or because of some compatibility rule that I don't know?

neersighted:
Because incrementing the lockfile version in a patch release is not backwards compatible or expected, let alone the scale of risk for a 'stable' branch.

dimbleby:
it can be made backwards compatible, I don't know about "expected"

neersighted:
Would we write the hashes twice, one to the new location, and one to the old? I still don't think that's appropriate for a patch release, especially given the intention is to release 1.3 the backed up PRs as we can get them stabilized and over the finish line "soon..."

A new lock file format that is backwards-compatible in 1.3 (so that 1.2 can read them), followed by 1.4 generating lock files that 1.3 can read (but not 1.2) sounds like the preferred approach for this particular migration to me anyway.

dimbleby:
yes, for backwards compatibility the approach would be to write in both places, new code would prefer to read from the right place, old code could continue to read from the less good place.

While I see what this fix is trying to do I am quite uncomfortable with it: conceptually it's patching up an earlier mistake instead of fixing that mistake at root, and practically it's a lot of ugly code in an already horribly ugly place.

I'm tempted to assert that the bug being fixed isn't important enough to be worth that, so long as there's a plan that does better in 1.3 and 1.4.  But it would be reasonable to disagree.

dimbleby:
for clarity, #6393 is the proposal I had in mind

radoering:
> Well OK then, how about fixing the broken file format instead? We've learned that the files belong on the packages, let's write that in the lockfile?

Yes, we should fix the format long-term.

> I'm tempted to assert that the bug being fixed isn't important enough to be worth that, so long as there's a plan that does better in 1.3 and 1.4. But it would be reasonable to disagree.

I disagree. The bug is a regression making `poetry lock --no-update` unusable for some (not so uncommon) use cases. I strongly advocate a backport and I think my fix is probably minimally invasive.

> for clarity, #6393 is the proposal I had in mind

I haven't looked closely yet, but this is probably the solution for 1.3.

@neersighted If this workaround is only for 1.2, should we merge it into master first even if it is replaced by #6393 quickly after or should we only create a PR for 1.2?

src/poetry/puzzle/provider.py
radoering: I haven't changed anything here, just restructured the `if-else`. I don't want to include unrelated cleanups here because as discussion shows this fix will not survive long on master.

neersighted:
I think it is important to keep a roughly unified history -- nothing should go directly into 1.2, except as a backport would be my general stance. I'd be fine merging this, and then immediately ripping it out.

That being said, I think we should try and establish some testing around lockfile compatibility before we bump the format, and also add some code to 1.2 that will print a useful error message when it encounters a lockfile beyond what is known to be compatible.

src/poetry/puzzle/provider.py
radoering: It returns bytes if you feed it with bytes.

dimbleby:
> I think my fix is probably minimally invasive.

I don't think it is!  #6393 suggests a different approach which seems like a truer halfway house: how about, when reading the lockfile, we don't attach file to url packages at all?  (also other direct origin types)

That's a fix that's much closer to the root of the problem, if not as accurate as getting it right per package, and I think will be significantly more compact than this fix.

dimbleby:
> I think we should try and establish some testing around lockfile compatibility before we bump the format, and also add some code to 1.2 that will print a useful error message when it encounters a lockfile beyond what is known to be compatible.

This already exists (and is tested) https://github.com/python-poetry/poetry/blob/5997a605e55c9cbf08e70d60cd1980e45d94574b/src/poetry/packages/locker.py#L289-L301

radoering:
> how about, when reading the lockfile, we don't attach files to url packages at all? (also other direct origin types)

I also thought about that alternative and decided for changing `complete_package() ` because that 's the place where `files` is fixed for non-direct origin dependencies. However, that was before the lockfile change loomed on the horizon. Now, I agree that handling it when reading the lockfile might be better. I'll convert the PR to that variant.

Btw, not attaching `files` is only correct for url, git and directory dependencies. For file dependencies the correct list item has to be found.

src/poetry/puzzle/provider.py
dimbleby, line 554: does urllib need to get involved at all?  Why isn't this just `Path(package.source_url)` - that's a pattern that is used elsewhere

dimbleby:
> Btw, not attaching files is only correct for url, git and directory dependencies. For file dependencies the correct list item has to be found.

agreed, just lazy language on my part!

dimbleby:
Also I don't know why poetry doesn't put a hash on url dependencies: if it's a good idea to calculate and store the hash for file dependencies then why not url dependencies too?  But that can definitely wait for another day.

src/poetry/puzzle/provider.py
radoering: At first, I thought hashes are not only locked for file dependencies but also for url dependencies (which turned out to be wrong). For file dependencies only, we don't need urllib.

src/poetry/packages/locker.py
```diff
@@ -129,8 +129,23 @@ def locked_repository(self) -> Repository:
...
-                files = metadata["files"][name]
-                package.files = files
+                if source_type in {"git", "directory", "url"}:
+                    package.files = []
+                else:
+                    files = metadata["files"][name]
+                    if source_type == "file":
+                        if len(files) > 1:
+                            filename = Path(url).name
+                            for file_item in files:
```
dimbleby, line 139: suggest
```python
filename = Path(url).name
package.files = [item for item in files if item["file"] == filename]
```
in the expectation that there is only one match, but if multiple then we don't know which is right so had better take them all

src/poetry/packages/locker.py
```diff
@@ -129,8 +129,23 @@ def locked_repository(self) -> Repository:
                 hashes = cast("dict[str, Any]", metadata["hashes"])
                 package.files = [{"name": h, "hash": h} for h in hashes[name]]
             else:
-                files = metadata["files"][name]
-                package.files = files
+                if source_type in {"git", "directory", "url"}:
```
dimbleby, line 132: pedantic: unnecessary identation here.
```python
elif source_type in {"git", "directory", "url"}:
...
```
would be ever so slightly neater

src/poetry/packages/locker.py
```diff
@@ -129,8 +129,23 @@ def locked_repository(self) -> Repository:
...
+                        if len(files) > 1:
+                            filename = Path(url).name
+                            for file_item in files:
+                                if file_item["file"] == filename:
+                                    package.files = [file_item]
+                                    break
+                    else:
+                        # Strictly speaking, this is not correct, but we have no chance
+                        # to always determine which are the correct files.
+                        # This is fixed later in Provider.complete_package() by looking
```
dimbleby, line 146: this is only true in code paths that perform solving.  Eg in `poetry install` there is no lookup in the repository, right?

radoering: `poetry install` performs a solver run, too.
https://github.com/python-poetry/poetry/blob/5997a605e55c9cbf08e70d60cd1980e45d94574b/src/poetry/installation/installer.py#L312-L316

dimbleby: I'm pretty sure it doesn't reach complete_package and update the `files` on things that it read from the locker.

If it did, that would defeat the point of storing the hashes in the first place.  The whole purpose is to check that the hashes for the files that are installed are the same as the hashes in the lockfile: if we always refreshed those values before making the check then this would all be a big old waste of time.

radoering: Of course, you're right. `complete_package()` is called but it does not do anything because in this solver run the only repository in the pool is the repository created from the lockfile. I'll adapt the comment.

dimbleby: Suggest that it's sufficient to comment this code, without reference to other code.  ie what you're saying in your first sentence is enough: this lockfile doesn't accurately record which files go with which packages, and this is our best effort at reconstructing that.

Other parts of the codebase can look after themselves!

src/poetry/packages/locker.py
```diff
@@ -128,9 +128,21 @@ def locked_repository(self) -> Repository:
                 # Old lock so we create dummy files from the hashes
                 hashes = cast("dict[str, Any]", metadata["hashes"])
                 package.files = [{"name": h, "hash": h} for h in hashes[name]]
+            elif source_type in {"git", "directory", "url"}:
+                package.files = []
             else:
                 files = metadata["files"][name]
-                package.files = files
+                if source_type == "file":
+                    if len(files) > 1:
```
dimbleby, line 136: sorry I wasn't clear earlier - suggest remove this check too.  At best it saves us iterating through a list of length 0 or 1, at worst it causes us to put a bad file here and later refuse to install the package because of hash mismatch.

dimbleby (approved):

dimbleby:
macos pipeline failure looks like https://github.com/pypa/pip/issues/11352, if so will presumably be fixed when a pip fix makes its way out

neersighted (approved):
This is a whole lot cleaner -- thanks @radoering and @dimbleby for the excellent work cleaning up this change. I agree with #6393 being the best way forward long term, but I am very much in favor of not changing the lockfile format in 1.2 (especially given incidental/whatever possible compatibility with 1.1 is desirable).

PR #6393, opened 2022-09-04
feat: lock file format 2.0 (was: store package files on the package in lockfile)

### LINKED ISSUE #6327: poetry lock adds duplicate hashes when using multiple constraint dependences

### LINKED ISSUE #6349: Poetry does not add all hashes for multiple constraint dependences

dimbleby:
counter-proposal to #6389, per https://github.com/python-poetry/poetry/pull/6389#issuecomment-1236210737

~doesn't quite fix #6327, we continue to write the unwanted extra metadata.files line to the lockfile until some future point where we stop writing that section at all~ does fix #6327 after all now I think harder about it, the unwanted files and hashes don't get wrongly attached to the URL package after this

also fixes #6349, installing a URL package doesn't care any more about files and hashes from other packages.


src/poetry/packages/locker.py
```diff
@@ -241,6 +250,7 @@ def set_lock_data(self, root: Package, packages: list[Package]) -> bool:
             "lock-version": self._VERSION,
             "python-versions": root.python_versions,
             "content-hash": self._content_hash,
+            # TODO stop writing files here, this is deprecated.
```
radoering, line 253: If we bump the `lock-version`, there is no reason to keep writing files here, isn't it?

radoering (changes requested):
We should add a test poetry can still read the old format. And while we're on it, we can add a test for the even older "only hash" format, too.

src/poetry/packages/locker.py
dimbleby: Depends whether we're ready to make a breaking change or not.

If we stop writing files here then the version should go 2.0, so that down-level poetry knows that it can't read it.  

I was expecting that we'd go 1.2 now, and 2.0 at some later point, but am happy to be told otherwise.

dimbleby:
> We should add a test poetry can still read the old format. And while we're on it, we can add a test for the even older "only hash" format, too.

There are already tests using the `old_lock` fixture.  

Given that even that doesn't use the even-older format I think it might actually be about time to remove the code that reads it.  Looks like it was removed at b1c4c68d08f0efeeac7f8efeaba2f26816cf5c57, three years ago, it's probably reasonable for the next poetry release to finally drop support for that - there must be approximately zero examples left in the eild.

src/poetry/packages/locker.py
radoering: Ah ok, I thought that any version bump makes it unreadable for older poetry. (I didn't look into the code.)

I have concerns that the lockfile size will increase significantly if we write the information twice. It seems we can only choose between the devil and the deep blue sea... I think we have to discuss that among maintainers.

dimbleby:
it turns out poetry's own unit tests use the old lockfile format, I've updated them to the 2019 vintage in this MR and also in https://github.com/python-poetry/poetry-plugin-export/pull/119

Updating them to use the format proposed here is more tedious, I'm inclined to say that we can live without doing that.  Anyway it might be some time before I can work up enthusiasm for the task...

dimbleby:
it occurs to me that another thing that could be done to soften the blow of changing the file format would be to backport the reading (but not the writing) part of the change to 1.2.

Maybe that would make it more palatable to do the breaking change all in one go, with no intermediate write-everything-in-two-places

- poetry 1.2.0 and 1.2.1 would still see a breaking change and refuse to use a poetry 1.3-written lockfile
- but poetry 1.2.2 (say) could use that lockfile just fine

being forced to `1.2.n` is likely more tolerable to some than being forced to `1.3`.  And if that `1.2.n` were released some moderate period ahead of `1.3.0`, then some of those who would have been annoyed by a breaking change in format instead won't even notice it.

radoering:
I like the idea. 👍

dimbleby:
pushed a commit to go straight to 2.0 in this MR, and remove the old `metadata.files` completely

#6608 is the proposed backport to 1.2.

radoering (approved):

PR #6404, opened 2022-09-04
chore(typing): replace typing.* types with native ones

Secrus:
Simple replacement of `typing` based container types with native ones (available via `from __future__ import annotations`).

### DIFF
--- a/src/poetry/console/commands/init.py
+++ b/src/poetry/console/commands/init.py
@@ -5,9 +5,7 @@ import sys
 from pathlib import Path
 from typing import TYPE_CHECKING
 from typing import Any
-from typing import Dict
 from typing import Mapping
-from typing import Union
 
 from cleo.helpers import option
 from packaging.utils import canonicalize_name
@@ -25,7 +23,7 @@ if TYPE_CHECKING:
 
     from poetry.repositories import Pool
 
-Requirements = Dict[str, Union[str, Mapping[str, Any]]]
+Requirements = dict[str, str | Mapping[str, Any]]
 
 
 class InitCommand(Command):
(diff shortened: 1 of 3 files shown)


Secrus:
> per the pipelines, this doesn't work - which is why it wasn't done in the first place

Yeah, we already discussed this with @neersighted on Discord.

neersighted:
Yeah, talked about typing syntax and support tools (MyPy, PyCharm) and how to configure them offline. I do wonder if we can stringify our type declarations here, and if so if it is desirable to do so to simplify onto one syntax. I could push a commit up and check pipelines, but you might already have knowledge/opinions @dimbleby?

dimbleby:
I remember being slightly surprised that what Secrus has tried here didn't work, and then deciding that it wasn't worth worrying about.  That's all I've got, sorry!

neersighted:
Fair enough -- and for posterity, I'll repost my Discord conversation here:
* `from __future__ import annotations` works by making annotations lazy-evaluated. By default they are parsed into the AST and thus all names need to be resolvable. When you import this `__future__` item, they are instead treated as strings (as if you quoted them), and the type-checking tool will evaluate them at type-checking time.
* This only works for annotations, and not for direct assignment (e.g. defining a TypeVar, a type alias, a named Union, etc) -- thus it doesn't work on any Python before 3.10 (which introduces the primitive-objects-as-types paradigm).
* This wasn't caught locally as @Secrus was using Python 3.10 locally -- to do typing work on Poetry, you want to work on the least common denominator, so you should use Python 3.7 to run PyCharm/mypy/pyright.
* I think we could quote this type definition and everything will just work (as mypy understands the new syntax regardless of Python version) -- but I'm not sure we want to do so/if we do, we should make that the project standard and enforce it with pre-commit. That's why this PR is still open, though I don't care that much, so maybe we juts close it and forget about it :laughing:

neersighted:
Going to go ahead and close this since no one seems interested in applying the new style as strings.

Issue #6138, opened 2022-08-08
Cannot have pip as a requirement on windows


vxgmichel:
- [x] I am on the [latest](https://github.com/python-poetry/poetry/releases/latest) Poetry version.
- [x] I have searched the [issues](https://github.com/python-poetry/poetry/issues) of this repo and believe that this is not a duplicate.
- [x] If an exception occurs when executing a command, I executed it again in debug mode (`-vvv` option).

- **OS version and name**: Windows 10
- **Poetry version**: 1.2.0b3
- **Contents of pyproject.toml file**:
  ```toml
  [tool.poetry]
  name = "some-api"
  version = "1.0.0"
  description = ""
  authors = ["redacted"]

  [tool.poetry.dependencies]
  python = "~3.9"
  pip = "^22"
  ```

## Issue

It seems like `pip` cannot be listed as requirement on windows since the pip commands are run as `pip install ...` instead of `python -m pip install ...`. On windows, running commands such as `pip install pip` produces an error of the type:
```
ERROR: To modify pip, please run the following command:
  C:\...\python.exe -m pip install pip
```

Find bellow the full trace corresponding to pyproject.toml above:

```
λ poetry install
Installing dependencies from lock file

Package operations: 0 installs, 1 update, 0 removals

  • Updating pip (22.1.2 -> 22.2.2): Failed

  CalledProcessError

  Command 'C:\Users\User\AppData\Local\pypoetry\Cache\virtualenvs\some-api-PA_BOgSJ-py3.9\Scripts\python.exe C:\Users\User\AppData\Roaming\pypoetry\venv\lib\site-packages\virtualenv\seed\wh
eels\embed\pip-22.2.1-py3-none-any.whl\pip install --disable-pip-version-check --prefix C:\Users\User\AppData\Local\pypoetry\Cache\virtualenvs\some-api-PA_BOgSJ-py3.9 --upgrade --no-deps C:
\Users\User\AppData\Local\pypoetry\Cache\artifacts\80\e1\ce\21d858207e1a33dc020819410970a306662661b1529bb85f459633a476\pip-22.2.2-py3-none-any.whl' returned non-zero exit status 1.

  at C:\Python39\lib\subprocess.py:528 in run
       524│             # We don't call process.wait() as .__exit__ does that for us.
       525│             raise
       526│         retcode = process.poll()
       527│         if check and retcode:
    →  528│             raise CalledProcessError(retcode, process.args,
[...]

PR #6430, opened 2022-09-06
Handle locked pip version on Windows

radoering:
Resolves: #6138

Use `poetry -m pip ...` instead of calling pip directly so that pip can update itself on Windows.

Just to be sure that this has no influence on performance, a little test:

```
setup = "import subprocess"

variants = {
    'embedded': [PYTHON_PATH, EMBEDDED_PIP_PATH, "-V"],  # fallback + prior #6062
    'bin': [PYTHON_PATH, PIP_PATH, "-V"],  # current
    'module': [PYTHON_PATH, "-m", "pip", "-V"],  # new
}

for name, cmd in variants.items():
    stmt = f"subprocess.check_output({cmd})"
    print(f"{name}: {timeit(stmt, setup, number=5)}")
```

Sample output:
```
embedded: 5.4903443
bin: 1.8143028999999995
module: 1.7508081999999998
```

As expected it makes no difference whether pip is called directly or as module.

### DIFF
--- a/src/poetry/utils/env.py
+++ b/src/poetry/utils/env.py
@@ -1393,7 +1393,7 @@ class Env:
         raise NotImplementedError()
 
     def get_pip_command(self, embedded: bool = False) -> list[str]:
-        raise NotImplementedError()
+        return [self.python, self.pip_embedded if embedded else self.pip]
 
     def get_supported_tags(self) -> list[Tag]:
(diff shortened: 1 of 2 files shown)


neersighted (approved):
LGTM -- glad this was as simple as I asserted it was :laughing:

Issue #6436, opened 2022-09-07
Dependency resolution way slower for packages with large number of releases from legacy secondary repositories 


MasterNayru:
- [x] I am on the [latest](https://github.com/python-poetry/poetry/releases/latest) Poetry version.
- [x] I have searched the [issues](https://github.com/python-poetry/poetry/issues) of this repo and believe that this is not a duplicate.
- [x] If an exception occurs when executing a command, I executed it again in debug mode (`-vvv` option).

- **OS version and name**: Fedora Linux 36, Debian 11
- **Poetry version**: Poetry v1.2.0
- **Link of a [Gist](https://gist.github.com/) with the contents of your pyproject.toml file**: https://gist.github.com/MasterNayru/d08b08c6ae86656faee171dcf387c0af

## Issue
I am seeing a significant speed regression when Poetry attempts to resolve dependencies from a secondary repo with many releases. To replicate this issue so that the Poetry team can see what I am talking about, my gist configures pypi.org as a secondary repo and points the pyproject.toml to packages with many releases. I have not been seeing this issue in v1.1 and do believe this was introduced relatively recently with the changes to support detecting yanked releases.

It looks like with Poetry v1.2 there was a change in logic to how the links to packages in a legacy repository are obtained. Poetry, as of v1.2, this line tells Poetry to reach out to a legacy repository and grab every package published for that package: https://github.com/python-poetry/poetry/blob/master/src/poetry/repositories/legacy_repository.py#L84. It then checks each package to see whether the package has been yanked or not, and in doing so loops through every published package again, and reparses the page and creates package objects for every link on the page for every package linked on the page in the repo. I hope I am describing this clearly enough, but basically because the links are stored in a list, it seems like Poetry has no option but to keep continually looping through the links over and over again to reprocess them to work out which links belong to a particular release version.

To put numbers on how drastic an effect this has on Poetry's performance, with the linked pyproject.toml, it takes somewhere in the order of 10 to 11 minutes to resolve these dependencies. If the secondary repo is removed, it takes about 25 seconds. I was able to make changes to the code in a fork of mine to make it use a dict instead of a list iterator to try and store links by package name and version, but I struggled to update the tests to make sure that I didn't break any other parts of Poetry in doing so. With the changes I had made, I was able to resolve dependencies using a secondary repo with the above config in Poetry v1.2 in the 25 seconds or so that I was expecting. 

[...]

PR #6442, opened 2022-09-07
fix: performance regression when parsing links from legacy repositories

radoering:
Resolves: #6436

Measurements of `poetry lock` with warm cache with example pyproject.toml from #6436:

|test case|time in s|peak memory usage in MB|
|---|---|---|
|legacy repository (before)|422|113|
|legacy repository (after)|3|118|
|pypi repository|1|92|

Usage of `@cached_property` would be nice, but unfortunately is not available in Python 3.7 and `@property` combined with `@cached` as proposed in https://docs.python.org/3/library/functools.html#functools.cached_property results in memory leaks (flake8-bugbear B019).

### DIFF
--- a/src/poetry/repositories/link_sources/base.py
+++ b/src/poetry/repositories/link_sources/base.py
@@ -1,10 +1,12 @@
 from __future__ import annotations
 
+import functools
 import logging
 import re
 
-from abc import abstractmethod
 from typing import TYPE_CHECKING
+from typing import DefaultDict
+from typing import List
 
 from packaging.utils import canonicalize_name
 from poetry.core.packages.package import Package
@@ -19,6 +21,8 @@ if TYPE_CHECKING:
 
     from packaging.utils import NormalizedName
     from poetry.core.packages.utils.link import Link
+
+    LinkCache = DefaultDict[NormalizedName, DefaultDict[Version, List[Link]]]
 
 
 logger = logging.getLogger(__name__)
@@ -39,21 +43,16 @@ class LinkSource:
 
     def __init__(self, url: str) -> None:
         self._url = url
+        self._get_link_cache_wrapper = functools.lru_cache(maxsize=1)(
+            self._get_link_cache
+        )
 
     @property
     def url(self) -> str:
         return self._url
 
-    def versions(self, name: str) -> Iterator[Version]:
-        name = canonicalize_name(name)
-        seen: set[Version] = set()
-
-        for link in self.links:
-            pkg = self.link_package_data(link)
-
-            if pkg and pkg.name == name and pkg.version not in seen:
-                seen.add(pkg.version)
-                yield pkg.version
+    def versions(self, name: NormalizedName) -> Iterator[Version]:
+        yield from self._link_cache[name]
 
     @property
     def packages(self) -> Iterator[Package]:
@@ -64,9 +63,10 @@ class LinkSource:
                 yield pkg
 
     @property
-    @abstractmethod
     def links(self) -> Iterator[Link]:
-        raise NotImplementedError()
+        for links_per_version in self._link_cache.values():
+            for links in links_per_version.values():
+                yield from links
 
     @classmethod
     def link_package_data(cls, link: Link) -> Package | None:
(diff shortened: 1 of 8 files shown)


MasterNayru:
This is like what I did but way better. Woohoo! I am so happy to have been even the slightest bit helpful in making Poetry a better tool for everyone. 

I have run this against other projects that I was seeing performance issues with internally and can see that performance in dependency resolution has dramatically improved.

I can't stress enough the importance of having this performance issue fixed and in a release branch (whether it be this PR or some other one). The performance drop with the logic as it stands on v1.2.0 just makes using poetry in environments where a secondary repo is used practically unusable. I have had two backends show this same performance issue, one being pypicloud which seems to direct clients to pypi.org when it does not have a package, and AWS CodeArtifact, which can be configured to pull packages from pypi.org itself. I would just hate to have people advocate against tools like this or be hesitant to upgrade.

src/poetry/repositories/link_sources/base.py
```diff
@@ -39,21 +40,16 @@ class LinkSource:
...
         self._url = url
+        self._get_link_cache_wrapper = functools.lru_cache(maxsize=1)(
+            self._get_link_cache
+        )
 
     @property
     def url(self) -> str:
         return self._url
 
     def versions(self, name: str) -> Iterator[Version]:
```
dimbleby, line 51: I reckon this takes a `NormalizedName` already so can be annotated as such, saving the need to re-canonicalize.

src/poetry/repositories/link_sources/html.py
```diff
@@ -44,7 +52,14 @@ def links(self) -> Iterator[Link]:
                 if link.ext not in self.SUPPORTED_FORMATS:
                     continue
 
-                yield link
+                pkg = self.link_package_data(link)
+                if pkg:
+                    links[pkg.name][pkg.version].append(link)
+                else:
+                    # dummy name and version
```
dimbleby, line 59: This seems weird.  Is it useful?  Can this branch just be removed?

MasterNayru: I believe this branch was needed to be able to pass tests where the links provided in the backend were invalid, and thus package objects with no name/version were causing issues. I wasn't comfortable enough with exactly why that was being tested enough to remove them.

radoering: The question is: "Do we want to return **all** links when calling `LinkSource.links` or only links representing valid package names? At the moment, we return all links. If we want to keep this behavior the branch is necessary.

radoering:
As proposed by @neersighted on discord, I added `backports.cached-property` in order to support `cached_property` on Python 3.7. This simplifies things a bit. It's in a separate commit, so we can decide if it's worth it.

src/poetry/repositories/link_sources/html.py
```diff
@@ -25,8 +29,13 @@ def __init__(self, url: str, content: str) -> None:
 
         self._parsed = html5lib.parse(content, namespaceHTMLElements=False)
 
-    @property
-    def links(self) -> Iterator[Link]:
+    @cached_property
+    def _link_cache(
+        self,
+    ) -> defaultdict[NormalizedName, defaultdict[Version, list[Link]]]:
```
neersighted, line 35: I find this really messy -- would you object to making this a named type (`LinkCache` maybe)?

neersighted (commented):
LGTM overall, I have high confidence in backporting this -- one minor nit that we can skip if you don't like it.

src/poetry/repositories/link_sources/html.py
```diff
@@ -25,8 +31,11 @@ def __init__(self, url: str, content: str) -> None:
 
         self._parsed = html5lib.parse(content, namespaceHTMLElements=False)
 
-    @property
-    def links(self) -> Iterator[Link]:
+    @cached_property
+    def _link_cache(self) -> LinkCache:
+        links: defaultdict[
+            NormalizedName, defaultdict[Version, list[Link]]
+        ] = defaultdict(lambda: defaultdict(list))
```
neersighted, line 38:
```suggestion
        links: LinkCache = defaultdict(lambda: defaultdict(list))
```

neersighted (changes requested):

neersighted (approved):

radoering:
pre-commit.ci autofix

neersighted (approved):

PR #6450, opened 2022-09-08
Add pre-commit hook to check poetry lock

nzig:
This PR fixes a few things with the repo's pre-commit hooks:
* ~~`files` regex - escape `.`.~~
* ~~Add `files` to `poetry-lock`, so that it doesn't run on every commit~~
* ~~Remove `language_version: python3`.~~
  * ~~Including this overrides a top-level [`default_language_version`](https://pre-commit.com/#top_level-default_language_version) directive, for no real benefit.~~
 * Add a hook that runs `poetry lock --check`

Edit: removed features that are already upstream.


.pre-commit-hooks.yaml
```diff
@@ -3,17 +3,24 @@
...
   description: run poetry lock to update lock file
   entry: poetry lock
   language: python
-  language_version: python3
   pass_filenames: false
+  files: ^pyproject\.toml|poetry\.lock$
+
+- id: poetry-lock-check
+  name: poetry-lock-check
+  description: run poetry lock to check lock file is up-to-date with pyproject.toml
```
mkniewallner, line 19:
```suggestion
  description: check that poetry lock file is consistent with pyproject.toml
```

mkniewallner (approved):

mkniewallner (approved):

mkniewallner:
@neersighted I'd like to see that backported to `1.2` branch as well, unless you think that only the hook fixes (and not the new hook) should be backported to `1.2`. Any opinion?

neersighted:
I'm not 100% sure on this just yet -- or rather I'm not sure on the future of pre-commit hooks in this repo given that upstream is incompatible with our release cycle. But I suppose that should not necessarily stop us from improving the hooks in this repo -- it's just that I hesitate to show more support/commitment for them when they may not be viable in the long run.

mkniewallner:
> I'm not 100% sure on this just yet -- or rather I'm not sure on the future of pre-commit hooks in this repo given that upstream is incompatible with our release cycle. But I suppose that should not necessarily stop us from improving the hooks in this repo -- it's just that I hesitate to show more support/commitment for them when they may not be viable in the long run.

Makes sense. I've removed my review and added `status/needs-consensus` label to prevent any accidental merge.

Happy to discuss that further if needed.

nzig:
> I'm not 100% sure on this just yet -- or rather I'm not sure on the future of pre-commit hooks in this repo given that upstream is incompatible with our release cycle. But I suppose that should not necessarily stop us from improving the hooks in this repo -- it's just that I hesitate to show more support/commitment for them when they may not be viable in the long run.

Could you clarify what you mean by "upstream is incompatible with our release cycle"? You mean that pre-commit hooks are tied to git refs rather than Poetry releases? Or something else I'm misunderstanding?

finswimmer:
> Could you clarify what you mean by "upstream is incompatible with our release cycle"?

I guess @neersighted is talking about the fact, that `pre-commit autoupdate` does not work for the pre-commit hooks provided in this repo. See the FAQ about this (https://python-poetry.org/docs/pre-commit-hooks/#why-does-pre-commit-autoupdate-not-update-to-the-latest-version) and the linked issues there.

nzig:
I see. Thanks for clarifying.

Prior to this repo having pre-commit hooks defined, I set up https://github.com/medigateio/poetry-pre-commit for doing the same thing internally at my company (also using the 3rd-party `poetry-lock-check` for getting `poetry lock check` in 1.1). You could create such a repo with a history like `pre-commit` expects to get around that problem. I'd be happy to help!

nzig:
Hey @neersighted , what are your thoughts on merging this?
It seems the current `.pre-commit-hooks.yaml` continues to be [maintained](https://github.com/python-poetry/poetry/commits/main/.pre-commit-hooks.yaml) despite no solution for `pre-commit autoupdate`, so maybe we could merge these fixes?

I'm also happy to help with setting up a separate repo for the pre-commit hooks to fix the autoupdate issue, if you'd like.

nzig:
Actually, I see that maybe all of the features of this PR were added upstream. The only thing that remains is a hook that runs `pre-commit check --lock`, but you already have one that runs `pre-commit check` so it's probably not needed.

I'll close this PR.

PR #6458, opened 2022-09-09
poc: foreign pip

neersighted:
This is a take on an alternate solution to our current embedded/in-venv pip woes: use a 'foreign' pip installed in Poetry's env to manage files in the project venv.

This works, except for scripts as `--target` does not install scripts properly, and `--prefix` cannot be used with the wheel platform overrides. However, if pip grows support for `--prefix` instead of `--target` in our use case, this could be a viable transitional approach for 1.2.

Tests are going to be totally busted as I have not upgraded any of them for the new code yet.

### DIFF
--- a/src/poetry/inspection/info.py
+++ b/src/poetry/inspection/info.py
@@ -23,7 +23,7 @@ from poetry.core.utils.helpers import temporary_directory
 from poetry.core.version.markers import InvalidMarker
 
 from poetry.utils.env import EnvCommandError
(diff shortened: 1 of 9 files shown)


neersighted:
After discussion with pip devs, https://github.com/pypa/pip/pull/11320 is merged and will be released soon, and will allow this to go forward without any of the drawbacks (having to gather wheel tags/not getting scripts) as we can just pass the path to the target interpreter to pip.

PR #6472, opened 2022-09-10
Preserve package's source when it has extras

maksbotan:
# Pull Request Check List

This PR splits out simple part of #6130. Once this merged I'll update that PR to contain only second (more controversial) fix.

### DIFF
--- a/src/poetry/puzzle/provider.py
+++ b/src/poetry/puzzle/provider.py
@@ -603,7 +603,15 @@ class Provider:
             )
             package = dependency_package.package
             dependency = dependency_package.dependency
-            _dependencies.append(package.without_features().to_dependency())
+            new_dependency = package.without_features().to_dependency()
+
+            # When adding dependency foo[extra] -> foo, preserve foo's source, if it's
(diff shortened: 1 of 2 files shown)


src/poetry/puzzle/provider.py
```diff
@@ -603,7 +603,18 @@ def complete_package(
...
+            new_dependency = package.without_features().to_dependency()
+
+            # When adding dependency foo[extra] -> foo, preserve foo's source, if it's
+            # specified. This prevents us from trying to get foo from PyPI
+            # when user explicitly set repo for foo[extra].
+            if (
+                not new_dependency.source_name
+                and dependency_package.dependency.source_name
+            ):
+                new_dependency.source_name = dependency_package.dependency.source_name
```
dimbleby, line 615:
```suggestion
            if not new_dependency.source_name and dependency.source_name:
                new_dependency.source_name = dependency.source_name
```

maksbotan: Makes sense, thanks!

neersighted (approved):

Issue #6466, opened 2022-09-09
Issue with filename for wheel with epoch version


wenkokke:
- [x] I am on the [latest](https://github.com/python-poetry/poetry/releases/latest) Poetry version.
- [x] I have searched the [issues](https://github.com/python-poetry/poetry/issues) of this repo and believe that this is not a duplicate.
- [x] If an exception occurs when executing a command, I executed it again in debug mode (`-vvv` option).

- **OS version and name**: macOS 11
- **Poetry version**: 1.2.0
- **Link of a [Gist](https://gist.github.com/) with the contents of your pyproject.toml file**: n/a

## Issue

When using a version number such as "1!1.0.0", the wheels generated by poetry have a filename with "1_1.0.0". This causes pip install to crash with the following error:
"ERROR: Could not build wheels for *, which is required to install pyproject.toml-based projects".

I believe that the following regular expression is at fault as it replaces the `!` with `_`:
https://github.com/python-poetry/poetry/blob/6fe7f54a83c59131d4e291084c3496a0c513cfaa/poetry/masonry/builders/wheel.py#L197-L203

dimbleby:
That seems to be what the spec says should be done: https://peps.python.org/pep-0427/#escaping-and-unicode

> Each component of the filename is escaped by replacing runs of non-alphanumeric characters with an underscore 

(I think you are confusing hyphens and underscores in your report, I don't think hyphens should be introduced)

wenkokke:
Yeah, it's an underscore, that was a typo.

Unfortunately, that does break installing with the latest pip.

dimbleby:
I dunno, I'm curious about whether PEP-427 is really the final word here.  If it is then I guess you should report this to pip.

However I could believe that there's more to be said about this and that poetry might not be right.

Would be interesting to know what other tooling does when building a wheel with a version that contains an epoch, or to see any published examples on pypi of such a wheel...

wenkokke:
I've also submitted this to pip, hopefully we can figure out which implementation is correct.

wenkokke:
> The rule you referenced from PEP 427 is outdated. The most up-to-date and canonical specification of wheel file names (among other things) [can be found on packaging.python.org](https://packaging.python.org/en/latest/specifications/binary-distribution-format/):
> 
> > Version numbers should be normalised according to [PEP 440](https://peps.python.org/pep-0440/). Normalised version numbers cannot contain `-`.
> 
[...]

PR #6476, opened 2022-09-11
don't use normalize_version()

dimbleby:
When you already have a `Version` in hand, `normalize_version(version.text)` is a very roundabout way of calling `version.to_string()`: it re-parses the version text to give you the same `Version` you already had and then calls `to_string()` on that.

https://github.com/python-poetry/poetry-core/blob/37cee90a5dd4c7ee2c5ee836216ba813242b3ade/src/poetry/core/utils/helpers.py#L27-L28

Then calling `escape_version()` on such a version is actually counter-productive, per #6466.

Similar changes can and should be made over in poetry-core, but it should be safe to merge this before that is done.

### DIFF
--- a/src/poetry/publishing/uploader.py
+++ b/src/poetry/publishing/uploader.py
@@ -11,7 +11,6 @@ import requests
 
 from poetry.core.masonry.metadata import Metadata
 from poetry.core.masonry.utils.helpers import escape_name
-from poetry.core.masonry.utils.helpers import escape_version
 from requests import adapters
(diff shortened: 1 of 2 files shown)


dimbleby:
test failure is not the usual 3.11 one, but it still looks like noise

neersighted (approved):

radoering:
Requires a backport to avoid deprecation warnings after merging python-poetry/poetry-core#469

PR #6480, opened 2022-09-11
chore(pre-commit): check lock file consistency

mkniewallner:
# Pull Request Check List

Eat our own dogfood by checking that Poetry's own lock file is consistent with `pyproject.toml`.

A warning is already shown if we `poetry install` with an inconsistent lock file, but since this won't stop the dependencies from being installed, resulting in a green CI, running `poetry lock --check` will more explicitly show the issue and prevent any merge.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -89,3 +89,11 @@ repos:
     rev: v2.20.0
     hooks:
       - id: validate_manifest
+
+  - repo: https://github.com/python-poetry/poetry
+    rev: 1.2.0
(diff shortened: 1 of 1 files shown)


mkniewallner:
If all the changes in https://github.com/python-poetry/poetry/pull/6450 are accepted and released under a new version (let's assume 1.2.1), the hook could be simplified to:
```yaml
  - repo: https://github.com/python-poetry/poetry
    rev: 1.2.1
    hooks:
      - id: poetry-lock-check
```

Issue #6301, opened 2022-08-31
poetry 1.2.0 can't install packages from private pypi servers supporting only MD5 hashes


jclerman:
- [x] I am on the [latest](https://github.com/python-poetry/poetry/releases/latest) Poetry version.
- [x] I have searched the [issues](https://github.com/python-poetry/poetry/issues) of this repo and believe that this is not a duplicate.
- [x] If an exception occurs when executing a command, I executed it again in debug mode (`-vvv` option).

- **OS version and name**: MacOS Monterey (12.5.1)
- **Poetry version**: 1.2.0

## Issue
After upgrading from poetry 1.1.15 to 1.2.0, I can no longer `poetry install` in my dev environment - it fails on a single package.  I've investigated and it **looks** like the problem is that with the new `poetry`, only SHA256 hashes are stored for packages in `poetry.lock`, even when the serving PyPi repository only supplies MD5 hashes.

Somehow, even though the PyPi repo serves MD5 (only) hashes, the new poetry stores a SHA256 hash in `poetry.lock` - then when I try to do `poetry install`, the MD5 hash from the retrieved package of course doesn't match the hash in the lock file, so I get:

```
Retrieved digest for link sirona-xyz-utils-0.10.2.tar.gz(md5:XXXXX) not in poetry.lock metadata ['sha256:YYYYY', 'sha256:ZZZZZ']

  at ~/Library/Application Support/pypoetry/venv/lib/python3.7/site-packages/poetry/installation/chooser.py:145 in _get_links
      141│             selected_links.append(link)
      142│
      143│         if links and not selected_links:
      144│             raise RuntimeError(
    → 145│                 f"Retrieved digest for link {link.filename}({h}) not in poetry.lock"
      146│                 f" metadata {hashes}"
      147│             )
      148│
      149│         return selected_links
```
(I've obscured the actual hashes and package-name above).

In our case, the PyPi repo server is Nexus, and not the very latest Nexus which does serve SHA256 hashes - that was only released this month, Aug 2022.

jclerman:
I can work around the issue by manually installing the offending package(s) via `pip`, but that doesn't address the real issue here.

neersighted:
Duplicate #4523

jclerman:
@neersighted, is this really a duplicate?  The behavior I've noted here is a regression from 1.1.15 to 1.2.0.

andrecp:
[...]

PR #6490, opened 2022-09-12
Feat: support multiple hash algorithms

Ckarles:
# Pull Request Check List

Resolves: #6301 #4085

### DIFF
--- a/docs/cli.md
+++ b/docs/cli.md
@@ -144,7 +144,7 @@ This ensures that everyone using the library will get the same versions of the d
 
 If there is no `poetry.lock` file, Poetry will create one after dependency resolution.
 
-If you want to exclude one or more dependency group for the installation, you can use
+If you want to exclude one or more dependency groups for the installation, you can use
 the `--without` option.
 
 ```bash
@@ -199,6 +199,12 @@ Pass `--all-extras` to install all defined extras for a project.
 poetry install --extras "mysql pgsql"
 poetry install -E mysql -E pgsql
 poetry install --all-extras
+```
+
+Extras are not sensitive to `--sync`.  Any extras not specified will always be removed.
+
+```bash
+poetry install --extras "A B"  # C is removed
 ```
 
 By default `poetry` will install your project's package every time you run `install`:
--- a/docs/managing-dependencies.md
+++ b/docs/managing-dependencies.md
@@ -197,7 +197,7 @@ poetry install --sync
 ```
 
 The `--sync` option can be combined with any [dependency groups]({{< relref "#dependency-groups" >}}) related options
-to synchronize the environment with specific groups. Note that extras are separate.  Any 
+to synchronize the environment with specific groups. Note that extras are separate.  Any
 extras not selected for install are always removed, regardless of `--sync`.
 
 ```bash
@@ -212,6 +212,6 @@ The `--sync` option replaces the `--remove-untracked` option which is now deprec
 
 ## Layering optional groups
 
-When you omit the `--sync` option, you can install any subset of optional groups without removing 
+When you omit the `--sync` option, you can install any subset of optional groups without removing
 those that are already installed.  This is very useful, for example, in multi-stage
 Docker builds, where you run `poetry install` multiple times in different build stages.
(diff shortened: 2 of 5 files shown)


dimbleby:
Pretty sure that this is at best a part of the solution: it's not enough to verify an md5 hash against a value in a lockfile, you also need persuade poetry to put a non-sha256 hash into the lockfile in the first place.

Also you are likely calculating the same hash again and again, which is undesirable

neersighted:
I have a PR cooking locally that attempts to solve this holistically -- however, I confess that reworking hash handling in a way that we can backport is rather difficult, and it may be easier to just forward-port the change from the 1.1 branch and consider the holistic rework something for 1.3 only.

Ckarles:
Thx for the quick review, I've been late to the party.

Please tell me if I'm missing something resuming that issue:
- Poetry is computing sha256 hashes to store in the lock file

Ckarles:
Thx for the quick review, I've been late to the party.

Please help me figure out if I'm reading this correctly:

When the lock file contains the package with its file+hash:
- find a link on the repo <- fails if the lockfile `hash_name:hash` is not equal to the one on the repo `Chooser._get_links()`
- validate the package by comparing lockfile hash with the computed hash from the downloaded archive `Executor._validate_archive_hash()`

When the lock file does not contains the package, the lockfile will add the hash of the package according to the specific repository implementation, e.g.
- for pypi repo, the hash is implied sha256 and extracted from the json pypi response `file_info["digests"]["sha256"]`
- for legacy repo, `hashname:hash` is extracted from `url.hash` and written as-is in the lockfile.

If I understand correctly, the issues triggered by 1.2.0 regarding hashes are:
1. If the hash from the lockfile is not with the same algo than the one supported on the repo, Poetry can't find a valid archive to download.
2. sha256 use is hardcoded for archive validation with lockfile.
3. There could be artifact repository implementations that Poetry does not support ?

`2` has been addressed in 1.1 branch.
`1` has no perfect solution afaik, on the top of my head. Is it safe to assume that, with current 1.2 implementation, if the hash provided by the repo doesn't change, the lockfile continues to be compatible ? If it is the case, as long as there is no breaking change on the artifact repo, there is no need to re-generate a lockfile, and if there is breaking change on the repo, it will be obligatory to re-generate a lock file. If the previous statement is correct, it seems to be an acceptable behavior (as long as it is advertised properly during a failing `install`).

timbedard:
The problem (with Nexus at least) is that the repo hashes _do_ change for _all versions_ of a package whenever any new version is pushed for that package.

JuanFranDevo (approved):

Fabhi:
Why aren't we merging it?

Ckarles:
As others have said, I don't think this is good enough.

That particular piece of code has already been written on a better way in 1.1 branch, also other matters need to be addressed, particularly regarding the chooser and the pipy adapter.

I'll complete this MR before next week with a more solid implementation. It will not cover any scenario but will at least provide guidance in case the pull doesn't work (instructing to perform another lock).

Fabhi:
> As others have pointed, I don't think this is good enough.
> 
> That particular piece of code has already been written on a better way in 1.1 branch, also other matters need to be addressed, particularly regarding the chooser and the pipy adapter.
> 
> I'll complete this MR before next week with a more solid implementation. It will not cover any scenario but will at least provide guidance in case the install from non-sha256 repos doesn't work (instructing to perform another lock). From what I've seen a better management of hashes requires some refactoring and changes to poetry-core (which is not the focus on that MR).

Wonderful! Let me know how I can help. I don't have much experience with contributing to open source, but I am good enough with python.

cpnielsen:
1. Based on @Ckarles comment, will this PR fix [the issue with md5-only hashed package](https://github.com/python-poetry/poetry/issues/6301) or is it only partially adressing it?
2. If someone can give me some pointers, I'd be happy to look into creating a PR that fixes this in poetry-core as/if necessary. We are "stuck" on an older version of Nexus that is beyond our control, and we cannot use poetry >=1.2 anywhere due to this bug.

Secrus:
Superseded by #8118

PR #6505, opened 2022-09-13
keyring and virtualenv are not using calver

dimbleby:
`keyring` had a `>=` dependency on the grounds that it used calendar versioning: but since they have reached `23.x.x` that's clearly not true.  

`virtualenv` had a `>=` dependency, presumably for similar reasons.  It looks as though they maybe did use calendar versioning up to 2020, but they've made plenty of releases in the last couple of years and only reached 20.16.5.

I am not claiming that these projects follow semver in a way that can be relied on: but that's just how life is in the python world.  Eg the [virtualenv changelog](https://virtualenv.pypa.io/en/latest/changelog.html) regularly includes 'Features' in patch releases.  

So far as adherence to semver goes, poetry and poetry-core are likely living in a glass house and should not throw any stones.

However I am claiming that it is inconsistent to use `>=` for these dependencies when, so far as I can tell, they are as close to using semver as any other project.

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -956,7 +956,7 @@ testing = ["func-timeout", "jaraco.itertools", "pytest (>=6)", "pytest-black (>=
 [metadata]
 lock-version = "1.1"
 python-versions = "^3.7"
-content-hash = "0cbb342d793a9fec2c99fbcc07434f320c618054a30dfc4af44896ebc3e9f676"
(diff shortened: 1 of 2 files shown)


neersighted (approved):

neersighted:
LGTM after a re-lock

neersighted:
Once again conflicted by your last PR

neersighted (approved):

PR #6509, opened 2022-09-14
fix: poetry-core>=1.1.0 is needed as build-system requirement

finswimmer:
Because Poetry itself uses dependency groups now, the minimum required `poetry-core` version for the `[build-system]` is `1.1.0`.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -95,7 +95,7 @@ poetry = "poetry.console.application:main"
 
 
 [build-system]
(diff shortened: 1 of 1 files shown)


onerandomusername:
As an aside: this should be added to the documentation and perhaps even to poetry's interactive jnterface when groups are in use, telling the user to update their build dependency requirements.

neersighted (approved):

PR #6512, opened 2022-09-14
fix: wrong nested table level when overwriting pyproject

frostming:
# Pull Request Check List

Previously, we rely on the `display_name` stored with `tomlkit.Table`, which is an internal implementation detail, and it breaks https://github.com/sdispater/tomlkit/pull/234. This PR corrects the nested table level, because `preserved` map comes from the poetry config table.

### DIFF
--- a/src/poetry/console/commands/self/self_command.py
+++ b/src/poetry/console/commands/self/self_command.py
@@ -75,7 +75,7 @@ class SelfCommand(InstallerCommand):
         content = Factory.create_pyproject_from_package(package=package)
 
         for key in preserved:
-            content[key] = preserved[key]
+            content["tool"]["poetry"][key] = preserved[key]
 
         self.system_pyproject.write_text(content.as_string(), encoding="utf-8")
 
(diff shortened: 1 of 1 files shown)


dimbleby:
I'm trying to move in the direction of wearing the hair shirt when it comes to dealing with tomlkit's horrible typing eg https://github.com/python-poetry/poetry-core/pull/468 (among other things) does this in what looks like a similar case:

```patch
-           return cast(Container, self.data["tool"]["poetry"])
+           tool = self.data["tool"]
+           assert isinstance(tool, dict)
+           config = tool["poetry"]
+           assert isinstance(config, Table)
+           return config
```
... which is awful, but at least we know what we're dealing with

(I see you're active at tomlkit, if you're able to do anything to make the types more usable that would be lovely)

frostming:
> living with the awkwardness when it comes to dealing with tomlkit's horrible typing

The awkwardness is because Python's dict is flexible, which can contain any value, and tomlkit.Table is just like a dict. How can it be better typed?

dimbleby:
I'm afraid I'm offering a diagnosis, rather than a solution.  

tomlkit's types are borderline unusable at the moment: per this MR, many common uses require you either to `# type: ignore` or to do something equally ugly.

In my opinion this simply isn't providing any help to users - it's making their lives harder, and their code not more likely to be correct.

I consider the current typing so not-useful that I'd actually prefer tomlkit did not expose `py.typed` at all.

radoering (approved):

Issue #6521, opened 2022-09-15
Poetry fails to run pip which picks `user` option from user config


sergei-dyshel:
- [x] I am on the [latest](https://github.com/python-poetry/poetry/releases/latest) Poetry version.
- [x] I have searched the [issues](https://github.com/python-poetry/poetry/issues) of this repo and believe that this is not a duplicate.
- [x] If an exception occurs when executing a command, I executed it again in debug mode (`-vvv` option).

- **OS version and name**:  MacOS Big Sur 11.6
- **Poetry version**: 1.2.0 (installed with Homebrew)
- **Link of a [Gist](https://gist.github.com/) with the contents of your pyproject.toml file**: https://github.com/sergei-dyshel/itunes-tag-sync/blob/master/pyproject.toml

## Issue
I have pip configured to install to user directory:
```
# pip.conf
[global]
user = yes
```

When I install project with poetry I get the following error:
```
# poetry install
  Command ['/Users/sergei/Library/Caches/pypoetry/virtualenvs/itunes-tag-sync-_c69w8Vl-py3.10/bin/python', '/usr/local/Cellar/poetry/1.2.0_1/libexec/lib/python3.10/site-packages/virtualenv/seed/wheels/embed/pip-22.2.2-py3-none-any.whl/pip', 'install', '--disable-pip-version-check', '--prefix', '/Users/sergei/Library/Caches/pypoetry/virtualenvs/itunes-tag-sync-_c69w8Vl-py3.10', '--no-deps', '/Users/sergei/Library/Caches/pypoetry/artifacts/e1/09/c0/6a6df3ef412eb19fcda1b353ea54e3b48f9c518348fadd0967a962cbb6/pyparsing-3.0.9-py3-none-any.whl'] errored with the following return code 1, and output:
  ERROR: Can not combine '--user' and '--prefix' as they imply different installation locations
```

So it looks like Poetry is picking `user` option from the config and it conflicts with the option passed on command line.

## Expected behaviour
Poetry ignores user config or at least potentially conflicting parts of it (e.g. `user` option).

## Workaround
Use env variable to override `user` during Poetry run.
```
PIP_USER=0 poetry install
```

dimbleby:
duplicate #1214

sergei-dyshel:
@dimbleby 
I think you misunderstood the issue here and there's not duplicate here.

#1214 is talking about making Poetry into a global (or rather user-specific) package manager that installs packages to user directory.

[...]

PR #6531, opened 2022-09-15
fix: run pip with --isolated to avoid bad local config

neersighted:
Resolves: #6521

### DIFF
--- a/src/poetry/utils/pip.py
+++ b/src/poetry/utils/pip.py
@@ -25,7 +25,13 @@ def pip_install(
     # either the virtual environment or the virtualenv package embedded wheel. Version
     # checks are a wasteful network call that adds a lot of wait time when installing a
     # lot of packages.
-    args = ["install", "--disable-pip-version-check", "--prefix", str(environment.path)]
+    args = [
+        "install",
+        "--disable-pip-version-check",
+        "--isolated",
+        "--prefix",
+        str(environment.path),
+    ]
 
     if not is_wheel and not editable:
         args.insert(1, "--use-pep517")


Secrus (approved):

HansAarneLiblik:
When you use `--isolated`, does this also ignore `PIP_INDEX_URL` env variable ??

strangemonad:
@HansAarneLiblik, the pip docs call this case out explicitly https://pip.pypa.io/en/stable/cli/pip/#cmdoption-isolated

VolvoxGlobator:
Just a note that this also (obviously) makes pip ignore the PIP_CACHE_DIR :-/

neersighted:
`PIP_CACHE_DIR` does nothing with Poetry as we use our own system for caching.

VolvoxGlobator:
But the pip caches the built wheel file to its cache. Which by this --isolated flag is reverted to default location

neersighted:
It shouldn't be, see #3439.

VolvoxGlobator:
This is the output we get during one of the craches - note that PIP_CACHE_DIR is present and is pointing elsewhere
(the crash itself we suspect to be caused by how pip, not poetry, handles access to cache from multiple instances of pip running simultaneously

```
  Building wheels for collected packages: future

    Building wheel for future (pyproject.toml): started

    Building wheel for future (pyproject.toml): finished with status 'done'

    Created wheel for future: filename=future-0.18.2-py3-none-any.whl size=491086 sha256=a7afdfefd81979c60312ec0fbada3b92361d7159c2b783f3c791cedf5de0bfac

    Stored in directory: c:\windows\system32\config\systemprofile\appdata\local\pip\cache\wheels\e6\98\20\6f689e25af7abda5431cc4480b86e91d9242e2a6211188f2d9
```

The cache location is default (in this case for system user)

neersighted:
Odd, I think pip behavior may have changed. However, you won't be able to set `PIP_CACHE_DIR` per Poetry worker anyway, so the better move here is to set `poetry config installer.parallel false` to turn off concurrency. Also make sure that the pip in use by Poetry (the one in the target environment) is up-to-date.

damienrj:
It looks like this might have broken SSL in 1.3.x #7513

PR #6541, opened 2022-09-16
canonicalized extra names

dimbleby:
companion to https://github.com/python-poetry/poetry-core/pull/476

### DIFF
--- a/src/poetry/installation/installer.py
+++ b/src/poetry/installation/installer.py
@@ -63,7 +63,7 @@ class Installer:
 
         self._whitelist: list[NormalizedName] = []
 
-        self._extras: list[str] = []
+        self._extras: list[NormalizedName] = []
 
         if executor is None:
             executor = Executor(
@@ -175,7 +175,7 @@ class Installer:
         return self
 
     def extras(self, extras: list[str]) -> Installer:
-        self._extras = extras
+        self._extras = [canonicalize_name(extra) for extra in extras]
 
         return self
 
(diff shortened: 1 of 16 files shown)


src/poetry/installation/installer.py
```diff
@@ -260,7 +260,11 @@ def _do_install(self) -> int:
                 )
 
             for extra in self._extras:
-                if extra not in self._locker.lock_data.get("extras", {}):
+                locker_extras = (
+                    canonicalize_name(extra)
+                    for extra in self._locker.lock_data.get("extras", {})
+                )
+                if extra not in locker_extras:
```
radoering, line 267: That doesn't work, does it? (Creating a generator expression and checking if something is in it.)

dimbleby: yes?
```
>>> foo = (f for f in [1,2,3])
>>> foo
<generator object <genexpr> at 0x7f8c83eb0040>
>>> 2 in foo
True
```

radoering: Ah ok, it works because it is only accessed once. I wasn't aware of that. Maybe, we should make `locker_extras` a set and create it once outside of the loop?

dimbleby: yes, that's an improvement.

The elephant in the room here is that I've no idea how this gets merged while the cross-repository CI is so entangled but sure, let's make this better...

neersighted: Core merges are not that hard -- usually we make the PR depend on the core PR branch to pass CI. Once merged and released in core, the PR is rebased to depend on the released tag instead.

radoering: The alternative is to make poetry so robust that it can handle both normalized and unnormalized extras. Even if it's feasible, I don't think it's worth it if we agree to include python-poetry/poetry-core#476 in the next core release. I don't even think this requires a new poetry-core major version. I'd consider it more of a bugfix. By the way, it would feel weird to include a new poetry-core major version in a poetry patch version (re backport label).

neersighted: I'd be happy to land the core change as a core minor version and then backport it in a 1.2 patch release myself.

PR #6555, opened 2022-09-18
tidying up

dimbleby:
refresh lockfile

it looks as though the latest lock was done with a cache that is missing some recent uploads


dimbleby:
re mypy my error, I'd simply forgotten to `install` - will undo that part!

Secrus (approved):

Issue #6495, opened 2022-09-12
poetry shell in git bash on Windows broken since 1.2.0


Godsmith:
- [x ] I am on the [latest](https://github.com/python-poetry/poetry/releases/latest) Poetry version.
- [x ] I have searched the [issues](https://github.com/python-poetry/poetry/issues) of this repo and believe that this is not a duplicate.
- [ x] If an exception occurs when executing a command, I executed it again in debug mode (`-vvv` option).

- **OS version and name**: Windows 11, git 2.33.1.windows.1
- **Poetry version**: 1.2.0
- **Link of a [Gist](https://gist.github.com/) with the contents of your pyproject.toml file**:  Not relevant

## Issue
Since poetry 1.2.0, `poetry shell` does not work on git bash in Windows.

```
$ poetry shell
Spawning shell within C:\Users\filip\AppData\Local\pypoetry\Cache\virtualenvs\dinrplan-9Ih6xyvL-py3.10
/usr/bin/bash: /K: No such file or directory
```

Downgrading to 1.1.5 works:

```
filip@Marwin MINGW64 ~/workspace/python/dinrplan (master)
$ poetry --version
Poetry version 1.1.15

filip@Marwin MINGW64 ~/workspace/python/dinrplan (master)
$ poetry shell
Spawning shell within C:\Users\filip\AppData\Local\pypoetry\Cache\virtualenvs\dinrplan-T1RTjXIi-py3.9

filip@Marwin MINGW64 ~/workspace/python/dinrplan (master)
$ which python
/c/Users/filip/AppData/Local/pypoetry/Cache/virtualenvs/dinrplan-T1RTjXIi-py3.9/Scripts/python
```

Looking at the code, I believe 57450f2717 broke the git bash support.

neersighted:
I think your identification of the point of change is incorrect, but we definitely need to check `$SHELL` instead of blindly trusting that the user is using CMD/Powershell.

neersighted:
`poetry shell` is just a convenience tool to start a subshell and activate an environment -- you can activate the env yourself with more typing (find the path with `poetry env info --path`).

bmarroquin:
The identification point is correct. For windows, we are only considering Powershell or cmd.exe. I'll add a third case to  fallback to the previous behavior. The prompt modification won't happen in this situtation.

Here is what shellingham reports
```
$ .venv/Scripts/python.exe -i
[...]

PR #6560, opened 2022-09-19
adds fallback behavior for non-ms shells

bmarroquin:
# Pull Request Check List

Resolves: #6495

Confirmed working on MS Powershell, Powershell 7, cmd and git bash

### DIFF
--- a/src/poetry/utils/shell.py
+++ b/src/poetry/utils/shell.py
@@ -78,12 +78,22 @@ class Shell:
         if sys.platform == "win32":
             if self._name in ("powershell", "pwsh"):
                 args = ["-NoExit", "-File", str(activate_path)]
-            else:
+            elif self._name == "cmd":
                 # /K will execute the bat file and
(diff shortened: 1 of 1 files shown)


neersighted (changes requested):
A couple thoughts -- `args` should be hoisted instead of being part of an `else` branch.

However, I'm wondering about this whole approach. Doesn't `pexpect` work on Windows? Why not use the same code path we use on Unix OSes as much as possible, and keep the special-casing to Windows Powershell and cmd.exe?

bmarroquin:
I'll give it a shot

bmarroquin:
`pexpect's` windows support is missing the `interact` method. Without that, you can't run a sub shell or reuse any of the unix code path. I tried a few things but none of them worked.

neersighted (approved):
LGTM -- this class is heavily overloaded and needs to be refactored, but this was a regression in 1.2 and we shouldn't let the refactor block the fix.

PR #6561, opened 2022-09-19
fix: fix failing tests due to updated list of available pythons

finswimmer:
Fixes tests, that will fail when changes in https://github.com/python-poetry/poetry-core/pull/477 are merged.

(Checks in the first commit shows that should work)

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -525,9 +525,9 @@ importlib-metadata = {version = ">=1.7.0", markers = "python_version < \"3.8\""}
 
 [package.source]
 type = "git"
-url = "https://github.com/dimbleby/poetry-core.git"
-reference = "canonicalize-extras"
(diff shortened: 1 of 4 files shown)


mkniewallner (approved):

neersighted (changes requested):
It feels like it would make more sense to interrogate core for the Python versions and check that they are all reflected in the markers. Are we opposed to doing that, instead of hardcoding the output?

radoering:
I took a quick look and it seems "interrogating core for the versions" works only for the test in `test_editable_builder.py`. In `test_factory.py` it's not all versions from `AVAILABLE_PYTHONS`, but only the versions which are supported by the project. So it's actually reimplementing `all_classifiers` (or comparing it to itself). Thus, maybe just keep it hardcoded in both places.

finswimmer:
I came to the same conclusion as @radoering, but forget to answer.

radoering (approved):

Issue #6201, opened 2022-08-20
Inconsistent Indentation in Lock File


Kurt-von-Laven:
- [x] I am on the [latest](https://github.com/python-poetry/poetry/releases/latest) Poetry version.
- [x] I have searched the [issues](https://github.com/python-poetry/poetry/issues) of this repo and believe that this is not a duplicate.
- [x] If an exception occurs when executing a command, I executed it again in debug mode (`-vvv` option).

- **OS version and name**: Ubuntu 20.04 LTS
- **Poetry version**: 1.1.14
- **Link to the contents of your pyproject.toml file**: [pyproject.toml](https://github.com/ScribeMD/slack-templates/blob/2d3d47617fcf07427bf59867d7ecff56c544ae5a/pyproject.toml)

## Issue
`poetry lock --no-update` produces inconsistent output on the same `pyproject.toml`. When run locally on Ubuntu 22.04 LTS, `poetry lock --no-update` continues to consistently produce this [poetry.lock](https://github.com/ScribeMD/slack-templates/blob/2d3d47617fcf07427bf59867d7ecff56c544ae5a/poetry.lock). This is consistent with past runs on Ubuntu 22.04 LTS in CI and hence what I perceive to be the correct output. [Here](https://github.com/ScribeMD/slack-templates/runs/7927992847?check_suite_focus=true) is a CI workflow on Ubuntu 20.04 LTS showing the diff (reproduced below) between the old (correct) and new (incorrect) output. The new output appears consistent across CI runs. Note that the new order of lines within `[package.extras]` is reverse alphabetical order, and the new order of dependencies within each line is reversed with respect to the original order. Both of those sorting issues were the focus of the closely related issue #6153 and have already been addressed (but not yet released) in #6169. The focus of this issue is the two spaces added before `metadata.python-versions`.

```diff
diff --git a/poetry.lock b/poetry.lock
index def2008..4e50b11 100644
--- a/poetry.lock
+++ b/poetry.lock
@@ -7,7 +7,7 @@ optional = false
 python-versions = "*"
 
 [package.extras]
-test = ["coverage", "flake8", "pexpect", "wheel"]
+test = ["wheel", "pexpect", "flake8", "coverage"]
 
 [[package]]
 name = "astroid"
@@ -31,10 +31,10 @@ optional = false
 python-versions = ">=2.7, !=3.0.*, !=3.1.*, !=3.2.*, !=3.3.*, !=3.4.*"
 
 [package.extras]
[...]

PR #6562, opened 2022-09-19
Use tomli for reading the lock file

dimbleby:
Pros and cons in this one, opinions are invited.

tomlkit is significantly slower than tomli / tomli-w.  And also, at least in a project that uses type-checking, it's a pain in the neck.

Even for a modestly sized project like poetry itself, reading `poetry.lock` on my laptop takes 0.6 - 0.7 seconds (timings obtained with some `time.time()` statements inserted into the codebase).  `tomli` manages it in ~0.02 seconds.

On larger projects the difference is more significant, in my experimentation that ratio of of 30-ish remains consistent.

Admittedly this is only likely to save folk a small number of seconds on any given command.  But fast tools are nicer to use than slow tools.

Against this: we already have `tomlkit` in the dependency tree, adding `tomli` and `tomli-w` looks a bit weird.  

The difference is that `pyproject.toml` is human-readable and it's highly desirable to maintain formatting and comments etc in interactions with it (`poetry add`, `poetry remove`).  Much as I'd like to use `tomli` and `tomli-w` everywhere: they just don't offer that feature.

`poetry.lock` however is machine-generated and machine-read and there is no reason to pay the tomlkit performance penalty in reading and writing it.

(I've also removed the code that re-reads the lockfile after writing it, "checking lock file data consistency".  Not clear to me what that was for: do you trust your toml writer or not?  That same code could be removed independent of the switch away from tomlkit)

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -951,7 +951,7 @@ testing = ["flake8 (<5)", "func-timeout", "jaraco.functools", "jaraco.itertools"
 [metadata]
 lock-version = "1.1"
 python-versions = "^3.7"
-content-hash = "f1ebc914a6733399d6dfb591c7c0e1aaef549a809a99c092a6fa04db0c209819"
+content-hash = "15420e602f85e17a36f549d671458e0439697ce2337b85023c9309d49b74f5e4"
 
 [metadata.files]
 attrs = [
--- a/poetry.lock
+++ b/poetry.lock
@@ -524,15 +524,15 @@ importlib-metadata = {version = ">=1.7.0", markers = "python_version < \"3.8\""}
 
 [[package]]
 name = "poetry-plugin-export"
-version = "1.1.2"
+version = "1.2.0"
 description = "Poetry plugin to export the dependencies to various formats"
 category = "main"
 optional = false
 python-versions = ">=3.7,<4.0"
 
 [package.dependencies]
-poetry = ">=1.2.0,<2.0.0"
-poetry-core = ">=1.1.0,<2.0.0"
+poetry = ">=1.2.2,<2.0.0"
+poetry-core = ">=1.3.0,<2.0.0"
 
 [[package]]
 name = "pre-commit"
@@ -951,7 +951,7 @@ testing = ["flake8 (<5)", "func-timeout", "jaraco.functools", "jaraco.itertools"
 [metadata]
 lock-version = "1.1"
 python-versions = "^3.7"
-content-hash = "15420e602f85e17a36f549d671458e0439697ce2337b85023c9309d49b74f5e4"
+content-hash = "703e97d65aae512130b6a43ef7c82a27c914b9f4023f5267824b54e901d4bde3"
 
 [metadata.files]
 attrs = [
@@ -1364,8 +1364,8 @@ poetry-core = [
     {file = "poetry_core-1.3.2-py3-none-any.whl", hash = "sha256:ea0f5a90b339cde132b4e43cff78a1b440cd928db864bb67cfc97fdfcefe7218"},
 ]
 poetry-plugin-export = [
-    {file = "poetry-plugin-export-1.1.2.tar.gz", hash = "sha256:5e92525dd63f38ce74a51ed68ea91d753523f21ce5f9ef8d3b793e2a4b2222ef"},
-    {file = "poetry_plugin_export-1.1.2-py3-none-any.whl", hash = "sha256:946e3313b3d00c18fb9a50522e9d5e6a7e111beaba8d6ae33297662fc2070ac1"},
+    {file = "poetry_plugin_export-1.2.0-py3-none-any.whl", hash = "sha256:109fb43ebfd0e79d8be2e7f9d43ba2ae357c4975a18dfc0cfdd9597dd086790e"},
+    {file = "poetry_plugin_export-1.2.0.tar.gz", hash = "sha256:9a1dd42765408931d7831738749022651d43a2968b67c988db1b7a567dfe41ef"},
 ]
 pre-commit = [
     {file = "pre_commit-2.20.0-py2.py3-none-any.whl", hash = "sha256:51a5ba7c480ae8072ecdb6933df22d2f812dc897d5fe848778116129a681aac7"},
--- a/poetry.lock
+++ b/poetry.lock
@@ -820,14 +820,6 @@ optional = false
 python-versions = ">=3.7"
 
 [[package]]
-name = "tomli-w"
-version = "1.0.0"
-description = "A lil' TOML writer"
-category = "main"
-optional = false
-python-versions = ">=3.7"
-
-[[package]]
 name = "tomlkit"
 version = "0.11.5"
 description = "Style preserving TOML library"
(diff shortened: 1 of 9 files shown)


neersighted:
Another argument against introducing tomli is we'll need to vendor it into poetry-core for locked dist/locked build support. 

I think the reason for reading the lock file back is because we may write escape characters or similar in strings (since we are accepting untrusted inputs from PyPI/other package indexes + setuptools and other build-systems) that make the file invalid (even if they are valid Python strings). tomli-w encourages something similar (namely, it makes no promises that it writes valid toml and suggests reading back to validate toml), but to be fair, I'm not sure it's 100% necessary.

dimbleby:
poetry-core doesn't need to rewrite pyproject.toml, so the other way to go would be to remove tomlkit there altogether

dimbleby:
per CI the way I've done this is also breaking for plugins, apparently the export plugin at least was relying on `locker.lock.path` which is now just `locker.path`.

dimbleby:
comment about poetry-core and vendoring got me thinking.  The full program for tomlkit-minimization - if that is indeed a desirable path, which perhaps is what this MR is really asking - would go something like this:

- copy poetry-core's `TOMLFile` into poetry
- poetry only needs to use `TOMLFile` (hence tomlkit) for interactions with `pyproject.toml`
  - probably that doesn't matter for anything except `poetry.lock`, other files are small enough that I guess the performance difference is likely tiny
- have poetry-core switch to reading `pyproject.toml` (and `poetry.lock`, if that day comes) via `tomli`
- deprecate / remove the poetry-core version of the `TOMLFile`
- (at this point poetry-core is vendoring `tomli` but not `tomlkit`)
- eventually, `tomllib` is in the python standard library and support for python <3.11 is dropped.  Then poetry-core doesn't vendor anything toml related.

mkniewallner:
I may not have enough knowledge to be 100% sure of that, but I am also under the assumption that poetry-core is never doing any write operations on TOML files.

So if we confirm that, I'm +1 on the plan highlighted by @dimbleby. https://github.com/python-poetry/poetry-core/pull/445 would probably be good to have as well (issue it fixes is highlighted [here](https://github.com/python-poetry/poetry/issues/6201#issuecomment-1221303348), for the context).

dimbleby:
I see that this one has been added to a 1.4 milestone.  A couple of thoughts:

- I don't know whether that's code for "some far distant future" or (say) "November".  It would be lovely if the meaning of such labels was shared somewhere

- In case it does mean "some far distant future", I'd note that I don't think there's anything much to be scared of in this one
  - obviously it needs to get over "is this even the right direction of travel?"
  - but if that standard is met then I see this as pretty straightforward, it's the expansion to https://github.com/python-poetry/poetry-core/pull/483 / #6616 that seems more exciting
  - as well as being a noticeable performance win, this fixes #6201

neersighted:
1.4 is next next release. The reason for this is mostly because of how invasive this is and the need to make a hard breaking change in core. I'd rather make changes in core more gradually in order to avoid branching it as much as possible.

I could be convinced otherwise, but personally I'd rather try and get the new installer + `pip --python` landed and then worry about this after we've got those complete.

dimbleby:
>  how invasive this is and the need to make a hard breaking change in core

I'm glad I spoke then, because I think we're at cross purposes.  This one does not make or require a change in core, but stands alone.  It's the follow-up https://github.com/python-poetry/poetry-core/pull/483 and #6616 that change poetry-core

neersighted:
Ah, fair point. I was under the impression that you intended #6616 to be a replacement/successor to this PR and commented on the wrong one. Am I correct in that it's a PR of a branch rebased on top of this, and you plan to keep it rebased so this can land first?

In that case I don't have major objections -- I'll try to complete a review of this PR sooner than later.

dimbleby:
re "1.4 is next next release".  Of course I understand that 1.4 is two more than 1.2, that wasn't what I was asking!  It would be good to understand whether that is more likely to be (say) November, or in some far-distant future (think of the gap betwewen 1.1 and 1.2 and double it).  

Eg that would help me to decide about whether it's even worth quibbling about how soon this gets merged.

(FWIW I'd like to see poetry turn out a 1.3 pretty darned soon, if only as a demonstration to the world - and to maintainers! - that shorter gaps are the new normal).

Anyway back on topic: yes, the reason I kept #6616 separate (and in two commits) was in anticipation that it might be reaonsable to merge this one sooner, and that one later.

neersighted:
It's very much a fluid thing, but the intention is to make releases on the order of months. I would invite you to participate in the Discord channel that you have been invited to if you'd like to join said discussions.

Certainly, we're carrying several important fixes in master right now, and it might be very nice to cut a 1.3, maybe even land invasive changes to core (this, extras and sdist normalization) in 1.4, and then scope the installer work for a 1.5. There's plenty of ways to skin the cat and it's mostly deciding what people have the energy for given limited volunteer bandwidth.

radoering:
I know @dimbleby is going to kill me for this and the exact format of the lock file is not relevant for most people but what about using `tomli` for reading but keeping `tomlkit` for writing? (I tried a bit and it seems straight forward.)

I see at least two advantages:

- We have more control about the format. E.g. using `tomli-w` will just revert #6734. (It doesn't look like that in the test but I suppose that's only because the test uses names and hashes that are significantly shorter than in reality. In a real world example, the output of tomli-w is like in the verbose example of the mentioned PR.) Further, `tomli-w` will write the dependencies of a package more verbose in some cases, e.g. I don't like something like https://github.com/python-poetry/poetry/blob/314d76e35e13ce82cec1b7fb7c2d5b09c44c45b6/tests/packages/test_locker.py#L216-L222
- Only `tomli` will be included in the standard library. `tomli-w` will remain an additional dependency.

Further, I don't think there is a relevant performance difference for writing the lock file in real world examples. (Running `poetry lock` on a small example takes 0.4 s on my PC, no matter if I use `tomli-w` or `tomlkit` for writing. I suppose in a larger example the time for writing the file can be neglected.)

dimbleby:
Control over the formatting is a double-edged sword, eg it's tomlkit's 'control' that exposes us to such as #6201.  Personally I see no reason to care about fine-grained control of how lists are presented: both formats are plenty readable enough.

However I must admit that I feel a strong mistrust of tomlkit and consequently perhaps I'm not as rational as I'd like to be about it...!  

As I said at the start of this thread, there are pros and cons, and the additional tomli-w dependency is admittedly among the cons.

I think the case is stronger over in poetry-core, where removing tomlkit in favour of something that is now in the standard library (since python 3.11) seems a clear win.  Also poetry-core doesn't need to write (except in unit test) so that in some sense simplifies the discussion.

I think tomlkit _is_ also slow at writing, though I don't have numbers to hand.  I agree that it's debatable how important half a second here or there is - though I do think that if you cut such gaps in the right places, it can make a real difference to the feel of a command line tool.

So anyway my preference would be that this stands as it is.  

However, I wouldn't want the perfect to be the enemy of the good: if keeping tomlkit for writing gets this over the line then let's do that, and I can try again at stepping away from it some other time...

dimbleby:
(I just want to acknowledge that in ten minutes experimentation I was unable to make tomlkit slow at writing, which modestly strengthens the case for not bothering with tomli-w)

dimbleby:
I pushed a commit that removes the tomlkit-using re-read of the lockfile after writing it.  That _is_ slow, per top of thread.  (Probably it's that which had left me with the apparently mistaken impression that writing the file with tomlkit is slow.)

dimbleby:
it maybe doesn't make much sense in the context of a solve that has likely taken much longer, but I genuinely find - at least now that I know about it! - that the gap between "Resolving dependencies" completing and "Writing lock file" is noticeable, and that it feels nicer to lose it.

(Kinda weird that we print "Writing lock file" after it has already been written, but whatever).

Anyway I also have never seen this check do anything other than pass.

radoering (approved):

PR #6576, opened 2022-09-20
chore(ci): setup certificates for 3.11 on macOS

mkniewallner:
CI is failing on Python 3.11.0rc2 for macOS because of missing certificates. For some reasons, they don't seem to be correctly installed, so we manually bundle them with `certifi` and symlink them.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -52,7 +52,7 @@ jobs:
         id: full-python-version
         run: echo ::set-output name=version::$(python -c "import sys; print('-'.join(str(v) for v in sys.version_info))")
 
-      # Certificates are not correctly installed for 3.11-dev.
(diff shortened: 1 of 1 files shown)


Secrus (dismissed):

.github/workflows/main.yml
```diff
@@ -52,6 +52,11 @@ jobs:
         id: full-python-version
         run: echo ::set-output name=version::$(python -c "import sys; print('-'.join(str(v) for v in sys.version_info))")
 
+      # Certificates are not correctly installed for 3.11-dev.
```
neersighted, line 55:
```suggestion
      # XXX: Certificates are not correctly installed for 3.11-dev -- remove when fixed.
```

neersighted (approved):

PR #6601, opened 2022-09-22
fix-3155: add coverage to install command --dry-run flag

DavidBord:
Resolves: #3155


neersighted (approved):
Looks good over all, thanks! Since the issue you linked is a permanent issue, you'll want to tag it with something like `Relates-to` or `Meta` or something, instead of `Resolves` (which will close it).

PR #6602, opened 2022-09-22
ci: remove cache health check

mkniewallner:
# Pull Request Check List

Cache is never reused on macOS because we use `timeout` to validate cache healthiness, but `timeout` is not available on macOS. So as the command fail, we execute the condition after `||` that deletes the virtual env, meaning we never use the cache on macOS (this can easily be seen [here](https://github.com/python-poetry/poetry/actions/runs/3107116318/jobs/5034833860) for instance).

I personally never came across a busted cache, so this PR suggests to remove the health check entirely (it is regularly updated anyway whenever there's a new Python patch version, or a dependencies update).

Alternatively, if you think that the health check is still necessary, removing the `timeout` could be enough, but I'd personally be onboard with removing the check entirely, and if we ever spot issues about busted caches, add it back.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -74,14 +74,9 @@ jobs:
 
       - name: Set up cache
         uses: actions/cache@v3
-        id: cache
         with:
           path: .venv
           key: venv-${{ runner.os }}-${{ steps.full-python-version.outputs.version }}-${{ hashFiles('**/poetry.lock') }}
-
-      - name: Ensure cache is healthy
-        if: steps.cache.outputs.cache-hit == 'true'
-        run: timeout 10s poetry run pip --version || rm -rf .venv
 
       # XXX: https://github.com/pypa/pip/issues/11352 causes random failures -- remove once fixed in a release.
       - name: Upgrade pip on 3.11 for macOS


Secrus (approved):

neersighted:
This is mostly a defense against Poetry breaking its installer temporarily due to a bad change at the same time a new cache is generated. Is this a huge issue? No, but it doesn't hurt.

Why not define a function for macOS that is compatible:

```sh
function timeout() {
  perl -e 'alarm shift; exec @ARGV' "$@"
}
```

mkniewallner:
> This is mostly a defense against Poetry breaking its installer temporarily due to a bad change at the same time a new cache is generated. Is this a huge issue? No, but it doesn't hurt.
> 
> Why not define a function for macOS that is compatible:
> 
> ```shell
> function timeout() {
>   perl -e 'alarm shift; exec @ARGV' "$@"
> }
> ```

I get that checking the cache healthiness could be nice to keep as a safeguard, but I'm not sure to exactly understand how the timeout helps. Aren't the steps (cache creation/retrieval included) all run sequentially?

Given our workflow, I would expect the cache to be retrieved, and only once it's fully loaded (or if we have a cache miss), the workflow would run `poetry run pip --version` that we define in the step right after.

neersighted:
They are sequential -- the `timeout` is really there in case `poetry run pip` hangs for some reason. I suppose we could just drop it and hope for the best -- I think @abn added it after some specific pathological behavior at some point in the past.

mkniewallner:
> They are sequential -- the `timeout` is really there in case `poetry run pip` hangs for some reason. I suppose we could just drop it and hope for the best -- I think @abn added it after some specific pathological behavior at some point in the past.

Let's play safe and keep it then: https://github.com/python-poetry/poetry/pull/6618.

PR #6604, opened 2022-09-23
Fix for inconsistent docs for multiple constraint dependencies (GH6603)

### LINKED ISSUE #6603: DOC: dependency example not consistent with text (multiple constraint dependencies)

## Issue

The docs for [Multiple Constraint Dependencies](https://python-poetry.org/docs/1.1/dependency-specification/#multiple-constraints-dependencies) was updated for version 1.2 .  It was previously:

_Let’s say you have a dependency on the package foo which is only compatible with Python <3.0 up to version 1.9 and compatible with Python 3.4+ from version 2.0: you would declare it like so:_

```
[tool.poetry.dependencies]
foo = [
    {version = "<=1.9", python = "^2.7"},
    {version = "^2.0", python = "^3.4"}
]
```

In the 1.2 docs the dependency specs have been updated without the corresponding updates required to the text above it.  I.e. 1.2 docs are:

_Let’s say you have a dependency on the package foo which is only compatible with Python <3.0 up to version 1.9 and compatible with Python 3.4+ from version 2.0: you would declare it like so:_

```
[tool.poetry.dependencies]
foo = [
    {version = "<=1.9", python = "^3.6"},
    {version = "^2.0", python = "^3.8"}
]
```

venaturum:
Fixed #6603 

Doc change only

### DIFF
--- a/docs/dependency-specification.md
+++ b/docs/dependency-specification.md
@@ -301,15 +301,15 @@ pathlib2 = { version = "^2.2", markers = "python_version <= '3.4' or sys_platfor
 Sometimes, one of your dependency may have different version ranges depending
 on the target Python versions.
 
-Let’s say you have a dependency on the package foo which is only compatible
-with Python >=3.6,<3.8 up to version 1.9 and compatible with Python 3.8+ from version 2.0:
+Let's say you have a dependency on the package `foo` which is only compatible
+with Python 3.6-3.7 up to version 1.9, and compatible with Python 3.8+ from version 2.0:
 you would declare it like so:
 
 ```toml
 [tool.poetry.dependencies]
 foo = [
-    {version = "<=1.9", python = "^3.6"},
-    {version = "^2.0", python = "^3.8"}
+    {version = "<=1.9", python = ">=3.6,<3.8"},
+    {version = "^2.0", python = ">=3.8"}
 ]
 ```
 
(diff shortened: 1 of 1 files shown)


docs/dependency-specification.md
```diff
@@ -301,8 +301,8 @@ pathlib2 = { version = "^2.2", markers = "python_version <= '3.4' or sys_platfor
 Sometimes, one of your dependency may have different version ranges depending
 on the target Python versions.
 
-Let's say you have a dependency on the package `foo` which is only compatible
-with Python <3.0 up to version 1.9 and compatible with Python 3.4+ from version 2.0:
+Let’s say you have a dependency on the package foo which is only compatible
```
neersighted, line 304:
```suggestion
Let’s say you have a dependency on the package `foo`, which is only compatible
```

docs/dependency-specification.md
```diff
@@ -301,8 +301,8 @@ pathlib2 = { version = "^2.2", markers = "python_version <= '3.4' or sys_platfor
...
+with Python >=3.6,<3.8 up to version 1.9 and compatible with Python 3.8+ from version 2.0:
```
neersighted, line 305:
```suggestion
with Python 3.6-3.7 up to version 1.9, and compatible with Python 3.8+ from version 2.0:
```

neersighted (changes requested):
Could you also please change the constraints to be `>=3.6,<3.8` and `>=3.8`?

venaturum:
@neersighted Hi Bjorn, requested changes made 👍

neersighted (approved):

Issue #1917, opened 2020-01-18
Keyring errors during non-publishing operations


wahuneke:
I gather that keyring integration was added as a convenience feature so that you can publish packages using keys stored in your keyring.  This ticket describes that feature:

https://github.com/python-poetry/poetry/issues/210

But it appears that poetry tries to access the keyring even for install operations. In one of my first times using poetry -  I, as a security paranoid person, decided to abort installation of a py package I wanted because the installation process seemed to be wanting to access my keyring.

I think people should be careful about granting programs access to things they should not need to access. This behavior is an example of a problem that makes that hard to achieve.

For me, I worked around the problem by pip uninstalling the 'keyring' package from that virt env.

In the future, I think it would be beneficial to poetry's adoption and to its users, if users are not prompted for keyring unless keys should actually be needed (for a 'publish' operation).

Thanks! I'm happy to be trying out poetry.

sdispater:
Do you have private repositories present in your `pyproject.toml` file?

I think you are affected by the issue that is fixed by #1892. The fix will be available in the next bug fix release.

colatkinson:
I was affected by something similar on the latest release (1.0.5), but I'm not sure if it's entirely the same.

The keyring will be hit for private repos if there is an `auth.toml` with only the username set, even if credentials are provided via environment variables.

I believe the relevant code is [here](https://github.com/python-poetry/poetry/blob/4687ef89f69cf9e7b1ba5cc8678f54db25be3290/poetry/utils/password_manager.py#L156). In this case, `auth` will be truthy, causing the keyring path to be taken.

In our case, the "fix" was to remove any leftover `auth.toml` files in the environment. Manually uninstalling keyring had the same effect.

I was wondering your thoughts on adding a config variable that would prevent any keyring hits altogether (by e.g. unconditionally marking the keyring as unavailable). This would have largely the same effect as @wahuneke's workaround, but might be a bit cleaner for CI pipelines and such.

asandeep:
+1 to this. I am facing similar issue while integrating poetry with Jenkins. Even though private repo credentials were provided via environment variables, poetry still tries to look into keyring for credentials.

dariobig:
+1 running into this issue while working over ssh

bdowling:
fwiw, I came across this issue and agree with the concerns mentioned above.  

[...]

PR #6612, opened 2022-09-23
fix: remove exception when keyring is locked 

MatthieuBizien:
# Pull Request Check List

Resolves: #1917

### DIFF
--- a/src/poetry/utils/password_manager.py
+++ b/src/poetry/utils/password_manager.py
@@ -48,7 +48,11 @@ class PoetryKeyring:
         import keyring
 
         for name in names:
-            credential = keyring.get_credential(name, username)
+            try:
+                credential = keyring.get_credential(name, username)
+            except keyring.errors.KeyringLocked:
+                logger.debug("Keyring %s is locked", name)
+                credential = None
             if credential:
                 return HTTPAuthCredential(
                     username=credential.username, password=credential.password


neersighted:
Thanks for starting on this! There are many more errors Keyring has been throwing (one of the reasons no one has started on this yet) -- see all the dupes of #1917 for situations we should handle.

We also most definitely need a unit/regression test for this.

real-yfprojects:
> Thanks for starting on this! There are many more errors Keyring has been throwing (one of the reasons no one has started on this yet) -- see all the dupes of #1917 for situations we should handle.

I was looking through all linked issues for Exceptions poetry could catch to prevent this.  I bumped into some problems. Sometimes the keyring raises a RuntimeError. I guess discarding that isn't a problem. However how to catch `win32ctypes.pywin32.pywintypes.error` which is a direct subclass of `Exception`?
There even was an incident where `keyring` raised a `ModuleNotFoundError`. IMO the last two are issues of `keyring` and should be handled by that library. However we could also simply catch all `Exceptions`.

dimbleby:
I would agree with you: catching `keyring.errors.KeyringError` seems like the right thing to do, and if keyring raises other errors then that feels to me like a bug that should be raised against that project.

That ought to catch the great majority of cases; and if keyring does take the view that it can raise any old exception and callers are supposed to handle them all - well that will be a simple enough thing to change later.

real-yfprojects:
Ok, so I opened an updated PR: #8227

Issue #4236, opened 2021-06-29
Request "git+http" scheme for private repo installation


avnovoselov:
- [x] I have searched the [issues](https://github.com/python-poetry/poetry/issues) of this repo and believe that this is not a duplicate.
- [x] I have searched the [documentation](https://python-poetry.org/docs/) and believe that my question is not covered.

## Feature Request

I have private gitlab server (available locally at http://gitlab.company-owned-network). Private repos that need to be installed with poetry are stored in gitlab.

Small improvement at `poetry/console/commands/init.py:409` - add "git+http" scheme.

current:

```python
if url_parsed.scheme in ["git+https", "git+ssh"]:
```

new:

```python
if url_parsed.scheme in ["git+https", "git+http", "git+ssh"]:
```

I ready to make pull request.

PR #6619, opened 2022-09-24
fix: missing git+http:// scheme support

moriyoshi:
# Pull Request Check List

Resolves: #4236

### DIFF
--- a/docs/dependency-specification.md
+++ b/docs/dependency-specification.md
@@ -176,7 +176,6 @@ how [repository credentials]({{< relref "repositories#configuring-credentials" >
 ```bash
 poetry config repositories.git-org-project https://github.com/org/project.git
 poetry config http-basic.git-org-project username token
-poetry add git+http://private-repo.example.com/project.git
 poetry add git+https://github.com/org/project.git
 ```
 
--- a/docs/dependency-specification.md
+++ b/docs/dependency-specification.md
@@ -176,6 +176,7 @@ how [repository credentials]({{< relref "repositories#configuring-credentials" >
 ```bash
 poetry config repositories.git-org-project https://github.com/org/project.git
 poetry config http-basic.git-org-project username token
+poetry add git+http://private-repo.example.com/project.git
 poetry add git+https://github.com/org/project.git
 ```
 
--- a/src/poetry/utils/dependency_specification.py
+++ b/src/poetry/utils/dependency_specification.py
@@ -29,6 +29,8 @@ if TYPE_CHECKING:
 DependencySpec = Dict[str, Union[str, bool, Dict[str, Union[str, bool]], List[str]]]
 BaseSpec = TypeVar("BaseSpec", DependencySpec, InlineTable)
 
+GIT_URL_SCHEMES = {"git+http", "git+https", "git+ssh"}
+
 
 def dependency_to_specification(
     dependency: Dependency, specification: BaseSpec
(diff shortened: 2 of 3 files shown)


neersighted:
I'm not sure about this change -- historically we have only allowed secure transports for Git, and both upstream Git and pip takes steps to discourage their use:

image

Generally, Poetry tries to make things as secure and reproducible as possible, by enforcing check-summing, encrypted transports, etc.

What is gained here that can't be solved by maintaining an intranet CA or using Let's Encrypt?

moriyoshi:
> I'm not sure about this change -- historically we have only allowed secure transports for Git, and both upstream Git and pip take steps to discourage their use:
> 
> image
> 
> Generally, Poetry tries to make things as secure and reproducible as possible, by enforcing check-summing, encrypted transports, etc.
> 
> What is gained here that can't be solved by maintaining an intranet CA or using Let's Encrypt?

I am completely aware of it. While insecure transport should be discouraged whenever possible, we might well happen to use the repo which is not exposed to the internet and neither secured.  Aside from the ethics, poetry-core actually supports the scheme (see [this](https://github.com/python-poetry/poetry-core/blob/4535724acb356a9da2fa976a243e5f63bef892f8/tests/vcs/test_vcs.py#L31)).

src/poetry/utils/dependency_specification.py
```diff
@@ -64,7 +64,7 @@ def _parse_dependency_specification_url(
     if not (url_parsed.scheme and url_parsed.netloc):
         return None
 
-    if url_parsed.scheme in ["git+https", "git+ssh"]:
+    if url_parsed.scheme in ["git+http", "git+https", "git+ssh"]:
```
neersighted, line 67: Let's make this (and the list below for URL deps) a file-level constant set (e.g. `GIT_URL_SCHEMES = {}`, `URL_SCHEMES = {}`).

docs/dependency-specification.md
```diff
@@ -172,6 +172,7 @@ how [repository credentials]({{< relref "repositories#configuring-credentials" >
 ```bash
 poetry config repositories.git-org-project https://github.com/org/project.git
 poetry config http-basic.git-org-project username token
+poetry add git+http://private-repo.example.com/project.git
```
neersighted, line 175: This is not the right place to demonstrate this at all -- it would be better off somewhere else (e.g. `basic-usage` or `commands`), or just not mentioned at all.

neersighted (dismissed):

src/poetry/utils/dependency_specification.py
moriyoshi: It looks like simply `startswith("git+")` would be enough:
https://github.com/python-poetry/poetry-core/blob/e72d3e497c7d9298bdabae179e8c58e58a82903b/src/poetry/core/packages/dependency.py#L425

neersighted: `git+` is not good enough as we don't support `git+file://` -- the original approach of a set is much better.

moriyoshi: Yes, it's much better, but I don't think it makes much sense as long as the underlying library doesn't do in that way. I think it should surely be addressed, but I guess it's not what should be done in this PR.

moriyoshi: The rationale is that it's even better off sharing the same logic between poetry and poetry-core, and that should be done in the separate PR. I don't want to dig this even further, like factoring out the list never used in other places.

neersighted: This is still the correct place to make this check; the code in poetry-core is generic URL parsing and not opinionated on what URLs are valid for a Python package vs. Git in general.

I am fine to move that logic into poetry-core, but I am not fine with removing the check from Poetry before that code is ready. Even though this is a fairly off-the-beaten path concern, making things more permissive and restricting it again later is not a viable approach.

PR #6621, opened 2022-09-25
normalized sdist names

dimbleby:
See https://github.com/python-poetry/poetry-core/pull/484


dimbleby:
I've left in the noise where we took equivalent-but-different decisions because I am too lazy to change such things, if you care then feel free to update

neersighted (approved):

PR #6632, opened 2022-09-26
repositories/http: support algorithms in hashlib.algorithms_guaranteed

vfazio:
PEP 503 says:

```
Repositories SHOULD choose a hash function from one of the ones guaranteed to be available via the hashlib module in the Python standard library (currently md5, sha1, sha224, sha256, sha384, sha512). The current recommendation is to use sha256.
```

It should make sense, then, to just check that the value returned is in `hashlib.algorithms_guaranteed` instead of a short subset of hashes.

Otherwise, the subset of hashes should be extended to those directly mentioned in the PEP though that list was compiled 7 years ago and does not reflect algorithms guaranteed to be present.

# Pull Request Check List

Resolves: #6301

### DIFF
--- a/src/poetry/repositories/http.py
+++ b/src/poetry/repositories/http.py
@@ -230,7 +230,7 @@ class HTTPRepository(CachedRepository, ABC):
 
             if not link.hash or (
                 link.hash_name is not None
-                and link.hash_name not in ("sha256", "sha384", "sha512")
+                and link.hash_name not in hashlib.algorithms_guaranteed
                 and hasattr(hashlib, link.hash_name)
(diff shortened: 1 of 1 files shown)


neersighted:
This does not resolve #6301 -- hash handling code is much more involved than this and there are no tests that actually make sure the intended fix really happens. I would suggest instead syncing with the authors of other attempts (e.g. #6490, #4740, #5326) to maybe combine efforts or see what is needed.

I have my own refactor spinning locally as well -- it's much more holistic and security-focused than a straight forward-port, but may take more time to cook.

Issue #2446, opened 2020-05-20
Support search in LegacyRepository


taranlu-houzz:
- [x] I am on the [latest](https://github.com/python-poetry/poetry/releases/latest) Poetry version.
- [x] I have searched the [issues](https://github.com/python-poetry/poetry/issues) of this repo and believe that this is not a duplicate.
- [x] If an exception occurs when executing a command, I executed it again in debug mode (`-vvv` option).

- **OS version and name**: macOS 10.14.6
- **Poetry version**: Poetry version 1.0.5
- **Link of a [Gist](https://gist.github.com/) with the contents of your pyproject.toml file**: https://gist.github.com/taranlu-houzz/d85b1ff99c8a0610ae7ed6911263038d

## Issue
When I use `poetry search <package from private repo>` it does not show up in the search results. However, using `poetry add <package from private repo>` works just fine.

## Config Notes
- I have a local pypi-server running using the package `pypiserver`.
- The package I am trying to search for is several levels down in a namespace package.
- This project is also a package that lives several levels down in a namespace package.
- I am able to publish to this local repo just fine using `poetry publish` with the appropriate flags.

jduprey:
I have seen this too... it was a while ago, but I think my colleague dug into poetry and found that they pypi URL was hardcoded for search in poetry.. maybe.. that is from memory so I could be wrong.

Would love to see search fixed.  Poetry add/install seems to work fine with private repos now.

WYishai:
@jduprey, it looks like you're right.
The relevant code is here:
https://github.com/python-poetry/poetry/blob/master/src/poetry/console/commands/search.py#L16
the `search` command creates a `PyPiRepository` with its default values, which cause it to use the global PyPi:
https://github.com/python-poetry/poetry/blob/master/src/poetry/repositories/pypi_repository.py#L52

Maybe a good-enough solution is to add an optional flag that specifies the remote repository URL. It should be very easy and simple to write (just a ~5 lines of code in the `search` command), and I think that in the cases you are using the `search` command you're probably not in the context of a specific repository...

taranlu-houzz:
@WYishai While that would work, it seems like a bit of a half-solution to me. In order to work with the private repo, I have to explicitly add that as a source in the `pyproject.toml`. To me, that indicates that the added source should always be considered, since that is what happens when using `poetry add`.

fbrodrigorezino:
[...]

PR #6669, opened 2022-10-01
Refactor repositories pool

b-kamphorst:
# Pull Request Check List

Relates-to: #3155

I refactored `Pool` in preparation of tackling https://github.com/python-poetry/poetry/issues/6713. Most notably, this PR uses an enum to significantly reduce the bookkeeping effort for the various types of repositories (primary / default / secondary).

Two considerations:
- One may argue whether the `RepositoryPriority` (name open for suggestions) should be an attribute of a `Repository` instead. I'm willing to have a look at that, but as that change will scatter all over the (test) code I decided to present this suggestion first.
- `OrderedDict` is used just to satisfy the test `test_repository_ordering` in `tests/repositories/test_pool.py`. Although I appreciate the deterministic approach, I would like to verify whether this ordering is actually relevant. In particular, was the order only verified to ensure proper ordering + clustering between default/primary/secondary, or also proper ordering within each cluster? Phrased differently, does the user expect that the order in which primary sources A and B are added through the CLI affects the resolution order? They might and I'm happy to leave it as is, but otherwise we can use normal dicts.

### DIFF
--- a/src/poetry/repositories/abstract_repository.py
+++ b/src/poetry/repositories/abstract_repository.py
@@ -6,9 +6,9 @@ from typing import TYPE_CHECKING
 
 
 if TYPE_CHECKING:
+    from poetry.core.constraints.version import Version
     from poetry.core.packages.dependency import Dependency
     from poetry.core.packages.package import Package
-    from poetry.core.semver.version import Version
 
 
 class AbstractRepository(ABC):
--- a/src/poetry/repositories/abstract_repository.py
+++ b/src/poetry/repositories/abstract_repository.py
@@ -1,0 +1,37 @@
+from __future__ import annotations
+
+from abc import ABC
+from abc import abstractmethod
+from typing import TYPE_CHECKING
+
+
+if TYPE_CHECKING:
+    from poetry.core.packages.dependency import Dependency
+    from poetry.core.packages.package import Package
+    from poetry.core.semver.version import Version
+
+
+class AbstractRepository(ABC):
+    def __init__(self, name: str) -> None:
+        self._name = name
+
+    @property
+    def name(self) -> str:
+        return self._name
+
+    @abstractmethod
+    def find_packages(self, dependency: Dependency) -> list[Package]:
+        ...
+
+    @abstractmethod
+    def search(self, query: str) -> list[Package]:
+        ...
+
+    @abstractmethod
+    def package(
+        self,
+        name: str,
+        version: Version,
+        extras: list[str] | None = None,
+    ) -> Package:
+        ...
--- a/src/poetry/repositories/legacy_repository.py
+++ b/src/poetry/repositories/legacy_repository.py
@@ -33,6 +33,18 @@ class LegacyRepository(HTTPRepository):
 
         super().__init__(name, url.rstrip("/"), config, disable_cache)
 
+    @property
+    def packages(self) -> list[Package]:
+        # LegacyRepository._packages is not populated and other implementations
+        # implicitly rely on this (e.g. Pool.search via
+        # LegacyRepository.search). To avoid special-casing Pool or changing
+        # behavior, we stub and return an empty list.
+        #
+        # TODO: Rethinking search behaviour and design.
+        # Ref: https://github.com/python-poetry/poetry/issues/2446 and
+        # https://github.com/python-poetry/poetry/pull/6669#discussion_r990874908.
+        return []
+
     def package(
         self, name: str, version: Version, extras: list[str] | None = None
     ) -> Package:
--- a/src/poetry/repositories/pool.py
+++ b/src/poetry/repositories/pool.py
@@ -3,6 +3,7 @@ from __future__ import annotations
 import enum
 
 from collections import OrderedDict
+from dataclasses import dataclass
 from enum import IntEnum
 from typing import TYPE_CHECKING
 
@@ -26,6 +27,12 @@ class Priority(IntEnum):
     SECONDARY = enum.auto()
 
 
+@dataclass(frozen=True)
+class PrioritizedRepository:
+    repository: Repository
+    priority: Priority
+
+
 class Pool(AbstractRepository):
     def __init__(
         self,
@@ -33,9 +40,7 @@ class Pool(AbstractRepository):
         ignore_repository_names: bool = False,
     ) -> None:
         super().__init__("poetry-pool")
-        self._repositories: OrderedDict[
-            str, tuple[Repository, RepositoryPriority]
-        ] = OrderedDict()
+        self._repositories: OrderedDict[str, PrioritizedRepository] = OrderedDict()
         self._ignore_repository_names = ignore_repository_names
 
         if repositories is None:
@@ -46,8 +51,10 @@ class Pool(AbstractRepository):
     @property
     def repositories(self) -> list[Repository]:
         unsorted_repositories = self._repositories.values()
-        sorted_repositories = sorted(unsorted_repositories, key=lambda p: p[1].value)
-        return [repo for (repo, _) in sorted_repositories]
+        sorted_repositories = sorted(
+            unsorted_repositories, key=lambda prio_repo: prio_repo.priority
+        )
+        return [prio_repo.repository for prio_repo in sorted_repositories]
 
     def has_default(self) -> bool:
         return self._contains_priority(Priority.DEFAULT)
@@ -57,7 +64,7 @@ class Pool(AbstractRepository):
 
     def _contains_priority(self, priority: Priority) -> bool:
         return any(
-            repo_prio is priority for _, repo_prio in self._repositories.values()
+            prio_repo.priority is priority for prio_repo in self._repositories.values()
         )
 
     def has_repository(self, name: str) -> bool:
@@ -66,7 +73,7 @@ class Pool(AbstractRepository):
     def repository(self, name: str) -> Repository:
         name = name.lower()
         if self.has_repository(name):
-            return self._repositories[name][0]
+            return self._repositories[name].repository
         raise IndexError(f'Repository "{name}" does not exist.')
 
     def add_repository(
@@ -89,7 +96,9 @@ class Pool(AbstractRepository):
             priority = Priority.DEFAULT
         elif secondary:
             priority = Priority.SECONDARY
-        self._repositories[repository_name] = (repository, priority)
+        self._repositories[repository_name] = PrioritizedRepository(
+            repository, priority
+        )
         return self
 
     def remove_repository(self, name: str) -> Pool:
--- a/src/poetry/repositories/pool.py
+++ b/src/poetry/repositories/pool.py
@@ -1,5 +1,9 @@
 from __future__ import annotations
 
+import enum
+
+from collections import OrderedDict
+from enum import IntEnum
 from typing import TYPE_CHECKING
 
 from poetry.repositories.exceptions import PackageNotFound
@@ -13,6 +17,14 @@ if TYPE_CHECKING:
     from poetry.repositories.repository import Repository
 
 
+class Priority(IntEnum):
+    # The order of the members below dictates the actual priority. The first member has
+    # top priority.
+    DEFAULT = enum.auto()
+    PRIMARY = enum.auto()
+    SECONDARY = enum.auto()
+
+
 class Pool:
     def __init__(
         self,
@@ -20,20 +32,15 @@ class Pool:
         ignore_repository_names: bool = False,
     ) -> None:
         self._name = "poetry-pool"
+        self._repositories: OrderedDict[
+            str, tuple[Repository, RepositoryPriority]
+        ] = OrderedDict()
+        self._ignore_repository_names = ignore_repository_names
 
         if repositories is None:
             repositories = []
-
-        self._lookup: dict[str, int] = {}
-        self._repositories: list[Repository] = []
-        self._default = False
-        self._has_primary_repositories = False
-        self._secondary_start_idx: int | None = None
-
         for repository in repositories:
             self.add_repository(repository)
-
-        self._ignore_repository_names = ignore_repository_names
 
     @property
     def name(self) -> str:
@@ -41,25 +48,29 @@ class Pool:
 
     @property
     def repositories(self) -> list[Repository]:
-        return self._repositories
+        unsorted_repositories = self._repositories.values()
+        sorted_repositories = sorted(unsorted_repositories, key=lambda p: p[1].value)
+        return [repo for (repo, _) in sorted_repositories]
 
     def has_default(self) -> bool:
-        return self._default
+        return self._contains_priority(Priority.DEFAULT)
 
     def has_primary_repositories(self) -> bool:
-        return self._has_primary_repositories
+        return self._contains_priority(Priority.PRIMARY)
+
+    def _contains_priority(self, priority: Priority) -> bool:
+        return any(
+            repo_prio is priority for _, repo_prio in self._repositories.values()
+        )
 
     def has_repository(self, name: str) -> bool:
-        return name.lower() in self._lookup
+        return name.lower() in self._repositories
 
     def repository(self, name: str) -> Repository:
         name = name.lower()
-
-        lookup = self._lookup.get(name)
-        if lookup is not None:
-            return self._repositories[lookup]
-
-        raise ValueError(f'Repository "{name}" does not exist.')
+        if self.has_repository(name):
+            return self._repositories[name][0]
+        raise IndexError(f'Repository "{name}" does not exist.')
 
     def add_repository(
         self, repository: Repository, default: bool = False, secondary: bool = False
@@ -68,69 +79,26 @@ class Pool:
         Adds a repository to the pool.
         """
         repository_name = repository.name.lower()
-        if repository_name in self._lookup:
-            raise ValueError(f"{repository_name} already added")
+        if self.has_repository(repository_name):
+            raise ValueError(
+                f"A repository with name {repository_name} was already added."
+            )
 
+        if default and self.has_default():
+            raise ValueError("Only one repository can be the default.")
+
+        priority = Priority.PRIMARY
         if default:
-            if self.has_default():
-                raise ValueError("Only one repository can be the default")
-
-            self._default = True
-            self._repositories.insert(0, repository)
-            for name in self._lookup:
-                self._lookup[name] += 1
-
-            if self._secondary_start_idx is not None:
-                self._secondary_start_idx += 1
-
-            self._lookup[repository_name] = 0
+            priority = Priority.DEFAULT
         elif secondary:
-            if self._secondary_start_idx is None:
-                self._secondary_start_idx = len(self._repositories)
-
-            self._repositories.append(repository)
-            self._lookup[repository_name] = len(self._repositories) - 1
-        else:
-            self._has_primary_repositories = True
-            if self._secondary_start_idx is None:
-                self._repositories.append(repository)
-                self._lookup[repository_name] = len(self._repositories) - 1
-            else:
-                self._repositories.insert(self._secondary_start_idx, repository)
-
-                for name, idx in self._lookup.items():
-                    if idx < self._secondary_start_idx:
-                        continue
-
-                    self._lookup[name] += 1
-
-                self._lookup[repository_name] = self._secondary_start_idx
-                self._secondary_start_idx += 1
-
+            priority = Priority.SECONDARY
+        self._repositories[repository_name] = (repository, priority)
         return self
 
-    def remove_repository(self, repository_name: str) -> Pool:
-        if repository_name is not None:
-            repository_name = repository_name.lower()
-
-        idx = self._lookup.get(repository_name)
-        if idx is not None:
-            del self._repositories[idx]
-            del self._lookup[repository_name]
-
-            if idx == 0:
-                self._default = False
-
-            for name in self._lookup:
-                if self._lookup[name] > idx:
-                    self._lookup[name] -= 1
-
-            if (
-                self._secondary_start_idx is not None
-                and self._secondary_start_idx > idx
-            ):
-                self._secondary_start_idx -= 1
-
+    def remove_repository(self, name: str) -> Pool:
+        if not self.has_repository(name):
+            raise IndexError(f"Pool can not remove unknown repository '{name}'.")
+        del self._repositories[name.lower()]
         return self
 
     def package(
@@ -138,60 +106,32 @@ class Pool:
         name: str,
         version: Version,
         extras: list[str] | None = None,
-        repository: str | None = None,
+        repository_name: str | None = None,
     ) -> Package:
-        if repository is not None:
-            repository = repository.lower()
+        if repository_name and not self._ignore_repository_names:
+            return self.repository(repository_name).package(
+                name, version, extras=extras
+            )
 
-        if (
-            repository is not None
-            and repository not in self._lookup
-            and not self._ignore_repository_names
-        ):
-            raise ValueError(f'Repository "{repository}" does not exist.')
-
-        if repository is not None and not self._ignore_repository_names:
-            return self.repository(repository).package(name, version, extras=extras)
-
-        for repo in self._repositories:
+        for repo in self.repositories:
             try:
-                package = repo.package(name, version, extras=extras)
+                return repo.package(name, version, extras=extras)
             except PackageNotFound:
                 continue
-
-            return package
-
         raise PackageNotFound(f"Package {name} ({version}) not found.")
 
     def find_packages(self, dependency: Dependency) -> list[Package]:
-        repository = dependency.source_name
-        if repository is not None:
-            repository = repository.lower()
+        repository_name = dependency.source_name
+        if repository_name and not self._ignore_repository_names:
+            return self.repository(repository_name).find_packages(dependency)
 
-        if (
-            repository is not None
-            and repository not in self._lookup
-            and not self._ignore_repository_names
-        ):
-            raise ValueError(f'Repository "{repository}" does not exist.')
-
-        if repository is not None and not self._ignore_repository_names:
-            return self.repository(repository).find_packages(dependency)
-
-        packages = []
-        for repo in self._repositories:
+        packages: list[Package] = []
+        for repo in self.repositories:
             packages += repo.find_packages(dependency)
-
         return packages
 
     def search(self, query: str) -> list[Package]:
-        from poetry.repositories.legacy_repository import LegacyRepository
-
-        results = []
-        for repository in self._repositories:
-            if isinstance(repository, LegacyRepository):
-                continue
-
+        results: list[Package] = []
+        for repository in self.repositories:
             results += repository.search(query)
-
         return results
--- a/src/poetry/repositories/pool.py
+++ b/src/poetry/repositories/pool.py
@@ -6,6 +6,7 @@ from collections import OrderedDict
 from enum import IntEnum
 from typing import TYPE_CHECKING
 
+from poetry.repositories.abstract_repository import AbstractRepository
 from poetry.repositories.exceptions import PackageNotFound
 
 
@@ -25,13 +26,13 @@ class Priority(IntEnum):
     SECONDARY = enum.auto()
 
 
-class Pool:
+class Pool(AbstractRepository):
     def __init__(
         self,
         repositories: list[Repository] | None = None,
         ignore_repository_names: bool = False,
     ) -> None:
-        self._name = "poetry-pool"
+        super().__init__("poetry-pool")
         self._repositories: OrderedDict[
             str, tuple[Repository, RepositoryPriority]
         ] = OrderedDict()
@@ -41,10 +42,6 @@ class Pool:
             repositories = []
         for repository in repositories:
             self.add_repository(repository)
-
-    @property
-    def name(self) -> str:
-        return self._name
 
     @property
     def repositories(self) -> list[Repository]:
--- a/src/poetry/repositories/pool.py
+++ b/src/poetry/repositories/pool.py
@@ -3,7 +3,6 @@ from __future__ import annotations
 from typing import TYPE_CHECKING
 
 from poetry.repositories.exceptions import PackageNotFound
-from poetry.repositories.repository import Repository
 
 
 if TYPE_CHECKING:
@@ -11,14 +10,16 @@ if TYPE_CHECKING:
     from poetry.core.packages.dependency import Dependency
     from poetry.core.packages.package import Package
 
+    from poetry.repositories.repository import Repository
 
-class Pool(Repository):
+
+class Pool:
     def __init__(
         self,
         repositories: list[Repository] | None = None,
         ignore_repository_names: bool = False,
     ) -> None:
-        super().__init__("poetry-pool")
+        self._name = "poetry-pool"
 
         if repositories is None:
             repositories = []
@@ -33,6 +34,10 @@ class Pool(Repository):
             self.add_repository(repository)
 
         self._ignore_repository_names = ignore_repository_names
+
+    @property
+    def name(self) -> str:
+        return self._name
 
     @property
     def repositories(self) -> list[Repository]:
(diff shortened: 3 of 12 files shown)


b-kamphorst:
I also removed the inheritance from `Repository` without removing any functionality. The only thing that `Pool` actually used was the `name` property. As it appears that this is not used anywhere, I made `Pool` nameless.

radoering:
> Phrased differently, does the user expect that the order in which primary sources A and B are added through the CLI affects the resolution order?

I expect that the order of primary sources in `pyproject.toml` is relevant. I didn't find anything in the docs, but I think we should keep it deterministic.

b-kamphorst:
Hi @radoering, thank you for your swift consideration.

> > Phrased differently, does the user expect that the order in which primary sources A and B are added through the CLI affects the resolution order?
> 
> I expect that the order of primary sources in `pyproject.toml` is relevant. I didn't find anything in the docs, but I think we should keep it deterministic.

That is very reasonable. I've clarified the resolution order in the docs. Please have a look at the proposed change. I'm also looking forward to your other thoughts on the PR.

radoering:
I just ran `test_pool.py` with coverage. It seems there is no coverage for `has_primary_repositories`, `find_packages` and `search` and `package` is only tested superficially. It would be nice to add some tests to gain more confidence in the refactoring.

I still need to take a closer look, but at first glance it looks like a neat simplification. 👍

However, I noticed that without `_lookup` and the Boolean flags there is more iterating in some methods/properties. I don't know how often this code is called during dependency resolution and if this could be a performance bottleneck.

src/poetry/repositories/pool.py
```diff
@@ -1,58 +1,66 @@
...
+
+class RepositoryPriority(Enum):
+    # The order of the members below dictates the actual priority. The first member has
+    # top priority.
+    DEFAULT = auto()
+    PRIMARY = auto()
+    SECONDARY = auto()
+
+
+class Pool:
```
neersighted, line 27:
```suggestion
class RepositoryPool:
```

Let's go all the way with breaking the false inheritance and name this clearly.

src/poetry/repositories/pool.py
```diff
@@ -62,134 +70,62 @@ def add_repository(
...
-                    self._lookup[name] -= 1
-
-            if (
-                self._secondary_start_idx is not None
-                and self._secondary_start_idx > idx
-            ):
-                self._secondary_start_idx -= 1
-
+        for repo in self.repositories:
+            if repo.name.lower() == repository_name.lower():
```
neersighted, line 90: While we are refactoring, I think we should hoist any normalization up and out (or make it explicit that names are not normalized) -- having this only in the remove code is surprising.

neersighted (changes requested):

src/poetry/repositories/pool.py
dimbleby: relevant: https://github.com/python-poetry/poetry/issues/6670#issuecomment-1264443658

src/poetry/repositories/pool.py
neersighted: Looking at this closer, we do have a lot of shared methods. Let's make a common `AbstractRepository` ABC that both `RepositoryPool` and `Repository` can implement.

src/poetry/repositories/pool.py
neersighted: Given that comment, I think we should normalize names uniformly.

src/poetry/repositories/pool.py
b-kamphorst: It turns out the number of shared methods is quite small: only `find_packages` and `search` overlap (assuming we don't put private methods in the ABC). The method `package` also exists in both, but the signature is different.

neersighted: Hmm, we could make `package` common with the use of kwargs. I'm rather inclined to do that because the idea of reasoning about these repos based on ABC instead of concrete type (when possible) is quite attractive. 

Also, I noticed for some reason `CachedRepository` does inherit from ABC without having any `@abstractmethod`s -- we should probably clean that up as well (e.g. add abstract declarations common between implementations).

b-kamphorst:
> I just ran `test_pool.py` with coverage. It seems there is no coverage for `has_primary_repositories`, `find_packages` and `search` and `package` is only tested superficially. It would be nice to add some tests to gain more confidence in the refactoring.

We now have 100% test coverage for `test_repository_pool.py` 😃 

> However, I noticed that without `_lookup` and the Boolean flags there is more iterating in some methods/properties. I don't know how often this code is called during dependency resolution and if this could be a performance bottleneck.

I assumed that we wouldn't do this too often, but it was quite easy to make this constant-time lookup (at the slight expense of clarity).

src/poetry/repositories/pool.py
b-kamphorst: Done.

src/poetry/repositories/pool.py
b-kamphorst: I didn't find a way to do this without `# type: ignore[override]`, but perhaps you know a better way?

`CachedRepository` has a single abstract method named `_get_release_info`.

b-kamphorst:
Apparently all tests failed due to an external import by `poetry_plugin_export`. I temporarily fixed it in ea72f31d4cc1ac07f8106a77ee8b7b0c9b75eb4b, but I can imagine you want to revert the whole renaming of `repositories/http.py`.

src/poetry/repositories/pool.py
dimbleby: in your current draft, only the `AbstractRepository` has kwargs, none of the concrete classes do.  So just remove the kwargs from the abstract class, and also remove all the `# type: ignore[override]` comments?

b-kamphorst:
> Apparently all tests failed due to an external import by `poetry_plugin_export`. I temporarily fixed it in [ea72f31](https://github.com/python-poetry/poetry/commit/ea72f31d4cc1ac07f8106a77ee8b7b0c9b75eb4b) and [9aec915](https://github.com/python-poetry/poetry/commit/9aec91518956c3a3564a89e1eaf517313c1468d3), but I can imagine you want to revert the whole renaming of `repositories/http.py` and `repositories/pool.py`.

This is clearly breaking, I'll revert the changed names tomorrow.

src/poetry/repositories/pool.py
b-kamphorst: The `RepositoryPool` does, that's the issue..

dimbleby: then add kwargs to the `RepositoryPool` only, and not to the abstract base class

dimbleby: I re-reviewed the code and stand by plan A: remove the kwargs from the abstract class and remove all of the `# type: ignore[override]` comments.

neersighted:
@b-kamphorst I'm okay with the rename personally -- the plugin is coupled to implementation details and we can work around that with a 3-PR-dance or just some maintainer magic (but we should break the rename out into a separate PR).

Regarding `search`, it's because only PyPI supports search as of now via the deprecated XML-RPC API, and not via PEP 508. Any vestiges of search on non-PyPI repositories is a holdover from when PyPI was not special-cased to use the JSON API: https://warehouse.pypa.io/api-reference/xml-rpc.html

radoering:
I'm strongly for breaking all the renaming out in a separate PR with **renaming only and no other changes**.

> We now have 100% test coverage for `test_repository_pool.py` 😃

Sounds great. 🎉 

> I assumed that we wouldn't do this too often, but it was quite easy to make this constant-time lookup (at the slight expense of clarity).

The only way to be sure is doing some measurements. (On the one hand I don't assume it is called too often or has a significant influence, but on the other hand I can imagine that some of the methods are called for each package lookup.) If others don't mind we can also take the risk.

docs/repositories.md
```diff
@@ -124,8 +124,7 @@ default = false
 secondary = false
 ```
 
-Any package source not marked as `secondary` will take precedence over [PyPI](https://pypi.org).
-
+The resolution order for searching packages is the following. First, the optional `default` package source is considered. Second, the `primary` (non-default, non-secondary) package sources are considered in order of appearance in your [`pyproject.toml`]({{< relref "pyproject" >}}) file. Third, unless another `default` package source has been configured, the [PyPI](https://pypi.org) repository is considered. Note that all non-secondary packages take precedence over [PyPI](https://pypi.org). Finally, the `secondary` package sources are considered in order of appearance.
```
neersighted, line 127: This isn't accurate -- `default` is mentioned twice and the first mention is incorrect. In addition, I would double quote (`"`) "primary" (instead of writing it as `primary`) to make it clear that it's just the internal nickname and not a TOML-level option.

src/poetry/console/commands/show.py
```diff
@@ -194,11 +194,11 @@ def _display_packages_information(
 
         from poetry.puzzle.solver import Solver
         from poetry.repositories.installed_repository import InstalledRepository
-        from poetry.repositories.pool import Pool
+        from poetry.repositories.repository_pool import RepositoryPool
         from poetry.utils.helpers import get_package_version_display_string
 
         locked_packages = locked_repository.packages
-        pool = Pool(ignore_repository_names=True)
+        pool = RepositoryPool(ignore_repository_names=True)
```
neersighted, line 201: Out of scope/for another time, but I am fascinated that we're using `RepositoryPool` here instead of just using `LockedRepository` directly.

src/poetry/repositories/repository_pool.py
```diff
@@ -0,0 +1,136 @@
...
+        for repo in self.repositories:
+            packages += repo.find_packages(dependency)
+        return packages
+
+    def search(self, query: str) -> list[Package]:
+        from poetry.repositories.legacy_repository import LegacyRepository
+
+        results: list[Package] = []
+        for repository in self.repositories:
+            if isinstance(repository, LegacyRepository):
```
neersighted, line 133: Instead of special casing here, it's likely better/cleaner to just return an empty list in `LegacyRepository`.

neersighted (changes requested):
Looking a lot more promising overall -- I haven't had time to look at everything exhaustively, but I left some feedback.

On the rename, I suggest we either to that in a separate PR, or you raise some DeprecationWarnings when the old names are imported.

I'd also like to see the kwargs change -- @dimbleby described it accurately, and it would remove the need for `type: ignore[override]` while better encapsulating how these classes are used.

radoering:
> On the rename, I suggest we either to that in a separate PR, or you raise some DeprecationWarnings when the old names are imported.

When renaming you have to cut the commits carefully in order to avoid "breaking" git history. We shouldn't just squash the commits. See python-poetry/poetry-core#482 for example. That's why I suggest doing it in a separate PR.

docs/repositories.md
radoering: It might be inaccurate, but I don't think the first mention is incorrect. IIUC, the second mention is not always correct.

According to https://github.com/python-poetry/poetry/blob/23c4f9b34bf897948b85b44e351d40d064f7efb5/src/poetry/factory.py#L144-L153

- If there is an explicit default, PyPI is not used.
- If there is no explicit default
  - and you don't have any primary repositories, PyPI is the default.
  - and you have primary repositories there is no default. PyPI is **not** the default, but the  secondary with the lowest priority.

The last case is not described properly.

neersighted: I think we're both agreeing but in different ways... Here is how I would describe the resolution logic:

* First, "primary" sources are considered in the order they are defined. These are entries in the pyproject.toml without `default = true` or `secondary = true`. This is equivalent to pip's `--extra-index-url`.
* Second, the `default` source is considered. If the default source is not overidden in pyproject.toml, this is PyPI. This is equivalent to pip's `--index-url`.
* Finally, all `secondary` sources are considered in the order they are defined. This is a Poetry invention.

radoering: Not entirely correct: If there is a default source (explicit or implicit), the default source is always the first primary source. (The default source is always considered first.)

neersighted: I don't disagree on a source code level, but I am referring to the practical behavior an end user sees:

Given a primary source, `click` and friends will be locked from the primary source:
```toml
[tool.poetry.dependencies]
click = "^8.1.3"

[[tool.poetry.source]]
name = "piwheels"
url = "https://www.piwheels.org/simple"
```

Given a secondary source, `click` and friends will be locked from the default source (PyPI):
```toml
[tool.poetry.dependencies]
click = "^8.1.3"

[[tool.poetry.source]]
name = "piwheels"
url = "https://www.piwheels.org/simple"
secondary = true
```

Skimming the current code, I'm actually a bit confused on how this is implemented, but it is what we currently do.

neersighted: Indeed, salting some prints into `add_repository` yields the following: 
```
piwheels 0
pypi 1
[<poetry.repositories.legacy_repository.LegacyRepository object at 0x7fdebc55dae0>, <poetry.repositories.pypi_repository.PyPiRepository object at 0x7fdebc5be3e0>]
```

I do think that the logic is tied into a hard to decipher pretzel, and the `RepositoryPriorities` enum in this PR needs to change.

radoering: In a discussion on Discord we came to the conclusion that

> * First, "primary" sources are considered in the order they are defined. These are entries in the pyproject.toml without `default = true` or `secondary = true`. This is equivalent to pip's `--extra-index-url`.
> 
> * Second, the `default` source is considered. If the default source is not overidden in pyproject.toml, this is PyPI. This is equivalent to pip's `--index-url`.
> 
> * Finally, all `secondary` sources are considered in the order they are defined. This is a Poetry invention.

describes how it should be. However, it isn't the current behavior. So let's drop the documentation for now, do only the refactoring here and add documentation + bugfix to comply with the description above in a follow-up PR because it's easier to do after the refactoring than before.

Sorry for the back and forth. We really appreciate your contribution. 🙏

src/poetry/repositories/pool.py
b-kamphorst: I've processed the change. I guess it is a matter of preference -- as I user I like to be informed of potential deviations from the ABC (API to me), but it did get quite ugly for little benefit.

docs/repositories.md
b-kamphorst: Ha don't worry about it! It took me quite some time to get familiar with enough code to work on this and the PyPI insertion and default/primary/secondary sources got me confused -- glad to see I'm not the only one. I'd be more than happy to change the behavior into something that can be explained easily (but I figured that a pure refactor should not change the current behavior). I'll take this up in my attempt at https://github.com/python-poetry/poetry/pull/5984.

src/poetry/repositories/repository_pool.py
b-kamphorst: Fixed. We might even clean up `LegacyRepository` further by removing all references to `self._packages` -- they seem to never get populated.

b-kamphorst:
It appears that master (44a89cbd7530bc3c50c87f672c221f61125faf54) breaks all tests?

neersighted:
> It appears that master ([44a89cb](https://github.com/python-poetry/poetry/commit/44a89cbd7530bc3c50c87f672c221f61125faf54)) breaks all tests?

You've got a typing issue.

src/poetry/repositories/pool.py
neersighted: Alternatively, we could just ignore `kwargs` on every implementation except RepositoryPool, if that makes you feel better :laughing:

b-kamphorst:
> > It appears that master ([44a89cb](https://github.com/python-poetry/poetry/commit/44a89cbd7530bc3c50c87f672c221f61125faf54)) breaks all tests?
> 
> You've got a typing issue.

Fixed that. Locally the tests also failed in Sint places but apparently that is resolved... 

I prepared another PR with the name changes. I'll present it when this PR is closed.

Thanks for your time, I enjoy the collaboration!

b-kamphorst:
Hi @radoering, @neersighted, are there any other changes that need to be processed here (at some point @neersighted mentioned that the `RepositoryPriorities` needs to change, but that discussion was dropped for this PR) or can the PR be merged?

src/poetry/repositories/legacy_repository.py
```diff
@@ -33,6 +33,10 @@ def __init__(
 
         super().__init__(name, url.rstrip("/"), config, disable_cache)
 
+    @property
+    def packages(self) -> list[Package]:
+        return []
```
radoering, line 38: Why that?

src/poetry/repositories/abstract_repository.py
```diff
@@ -0,0 +1,37 @@
...
+
+from abc import ABC
+from abc import abstractmethod
+from typing import TYPE_CHECKING
+
+
+if TYPE_CHECKING:
+    from poetry.core.packages.dependency import Dependency
+    from poetry.core.packages.package import Package
+    from poetry.core.semver.version import Version
```
radoering, line 11:
```suggestion
    from poetry.core.constraints.version import Version
```

src/poetry/repositories/legacy_repository.py
radoering: Can we add a comment in the code why we are just returning an empty list please?

src/poetry/repositories/pool.py
```diff
@@ -1,59 +1,77 @@
...
 if TYPE_CHECKING:
     from poetry.core.constraints.version import Version
     from poetry.core.packages.dependency import Dependency
     from poetry.core.packages.package import Package
 
+    from poetry.repositories.repository import Repository
 
-class Pool(Repository):
+
+class RepositoryPriority(IntEnum):
```
neersighted, line 21: I like the introduction of the dataclass -- with that present, I think this can be named cleaner as `Priority`.

src/poetry/repositories/pool.py
```diff
@@ -1,59 +1,77 @@
...
+    # The order of the members below dictates the actual priority. The first member has
+    # top priority.
+    DEFAULT = auto()
+    PRIMARY = auto()
+    SECONDARY = auto()
+
+
+@dataclass
```
neersighted, line 29: I think we should either make this `frozen=true` or `NamedTuple` as there is no reason I see to leave it mutable.

src/poetry/repositories/pool.py
```diff
@@ -1,59 +1,77 @@
...
+class PrioritizedRepository:
+    repository: Repository
+    priority: RepositoryPriority
+
+
+class Pool(AbstractRepository):
```
neersighted, line 35: It looks like the name got reverted to `Pool`. Sorry if I was unclear before -- I'd like to keep the name `RepositoryPool` in this PR, even if we do the file-level renames in another.

src/poetry/repositories/pool.py
```diff
@@ -1,59 +1,77 @@
...
-        self._lookup: dict[str, int] = {}
-        self._repositories: list[Repository] = []
-        self._default = False
-        self._has_primary_repositories = False
-        self._secondary_start_idx: int | None = None
+        super().__init__("poetry-repository-pool")
+        self._repositories: OrderedDict[str, PrioritizedRepository] = OrderedDict()
+        self._ignore_repository_names = ignore_repository_names
 
+        repositories = repositories or []
```
neersighted, line 45: This is pretty messy -- it doesn't look like we ever call `Pool(None)`, so it should be perfectly safe to change this to a default value without any other code changes.

src/poetry/repositories/pool.py
```diff
@@ -1,59 +1,77 @@
...
+        return self._contains_priority(RepositoryPriority.PRIMARY)
+
+    def _contains_priority(self, priority: RepositoryPriority) -> bool:
+        return any(
+            prio_repo.priority == priority for prio_repo in self._repositories.values()
+        )
 
     def has_repository(self, name: str) -> bool:
-        return name.lower() in self._lookup
+        return name.lower() in self._repositories
```
neersighted, line 69: We could do this in another PR, but in the long term I'd like to add a `RepositoryName` type that represents a normalized name, and centralize all normalization code there.

src/poetry/repositories/pool.py
```diff
@@ -1,59 +1,77 @@
...
 
     def repository(self, name: str) -> Repository:
         name = name.lower()
-
-        lookup = self._lookup.get(name)
-        if lookup is not None:
-            return self._repositories[lookup]
-
+        if name in self._repositories:
```
neersighted, line 73: I'd rather see this use `has_repository()`.

src/poetry/repositories/pool.py
```diff
@@ -63,133 +81,61 @@ def add_repository(
         Adds a repository to the pool.
         """
         repository_name = repository.name.lower()
-        if repository_name in self._lookup:
-            raise ValueError(f"{repository_name} already added")
-
-        if default:
-            if self.has_default():
-                raise ValueError("Only one repository can be the default")
+        if repository_name in self._repositories:
```
neersighted, line 84: Same as above, I'd rather see this use `has_repository()`.

src/poetry/repositories/pool.py
```diff
@@ -63,133 +81,61 @@ def add_repository(
...
-                for name, idx in self._lookup.items():
-                    if idx < self._secondary_start_idx:
-                        continue
-
-                    self._lookup[name] += 1
-
-                self._lookup[repository_name] = self._secondary_start_idx
-                self._secondary_start_idx += 1
-
+            repository_priority = RepositoryPriority.PRIMARY
```
neersighted, line 98: I'd rather use `priority` here and hoist this out of the else to be on top.

src/poetry/repositories/pool.py
```diff
@@ -63,133 +81,61 @@ def add_repository(
...
-                if self._lookup[name] > idx:
-                    self._lookup[name] -= 1
-
-            if (
-                self._secondary_start_idx is not None
-                and self._secondary_start_idx > idx
-            ):
-                self._secondary_start_idx -= 1
-
+        del self._repositories[repository_name.lower()]
```
neersighted, line 105: The old function had error checking -- this can now throw an `IndexError` -- we should probably check for existence first since we silently swallowed errors before.

neersighted (changes requested):
Overall looks very promising. I have some concerns about the specific implementation, but otherwise the new tests look good and the architecture seems sound. Good work cleaning this up!

src/poetry/repositories/legacy_repository.py
neersighted: Or a docstring :)

src/poetry/repositories/pool.py
neersighted: (and we should also have a test for deleting a non-extant repository)

src/poetry/repositories/legacy_repository.py
b-kamphorst: Looking into this again, I wonder why we return an empty list indeed. The rationale for changing this is to remove the special casing in `Pool.search`. However, `LegacyRepository.search` only browses `LegacyRepository.packages` and inspects their names (one may debate whether that is 'searching', but it is what happens). As such, it appears that *even if `LegacyRepository.packages` is populated*, this should not be a heavy task. Should we undo the change to `LegacyRepository`? If not, what is the rationale to be provided in the docstring?

Related (though a bit of a sidestep): it appears that the `search` command (console/commands/search) considers `pypi` only and ignores all other repositories (even if `pypi` is disabled). Is that the intended behaviour? It seems odd.

src/poetry/repositories/pool.py
b-kamphorst: That will still break [the plugin test code](https://github.com/python-poetry/poetry-plugin-export/blob/main/tests/conftest.py#L16), so I think it is best to combine the file+ object renaming in a single PR. Do you agree?

src/poetry/repositories/pool.py
b-kamphorst: We do call `Pool()` in many places. Which default value are you thinking of (assuming you don't want to have a mutable object as default)?

src/poetry/repositories/pool.py
b-kamphorst: See [278c432](https://github.com/python-poetry/poetry/pull/6669/commits/278c4325e3bb6a4a11e220fd18aab83217583b44) for an attempt. I'm not too happy with it, but let me know what you think.

b-kamphorst: Ah again this breaks plugin...

b-kamphorst:
> Once review is addressed, I'd like to see cleaner history -- it would be nice to introduce the new tests and then the refactor (assuming the tests pass without the refactor -- I didn't look closely to see if they would run against the old implementation).

This took me a lot of time, but at least I learned a bit more about git (plenty to learn left). The added tests are now in the first commit (not 100% at that commit just yet).

neersighted:
I personally would drop the `RepositoryName` commits from this PR and move them to a subsequent PR -- I think that we might want to think a little more about the design/implementation and it clutters this one too much.

src/poetry/repositories/pool.py
neersighted: Ah, I missed that -- fair enough!

src/poetry/repositories/pool.py
neersighted: Ah, shoot, I apparently have spent too much time in other languages (*cough*Ruby*cough*) -- I appear to have forgotten about mutable default values last night. My bad :laughing:

src/poetry/repositories/pool.py
```diff
@@ -1,59 +1,79 @@
...
-        self._default = False
-        self._has_primary_repositories = False
-        self._secondary_start_idx: int | None = None
+        super().__init__("poetry-repository-pool")
+        self._repositories: OrderedDict[
+            RepositoryName, PrioritizedRepository
+        ] = OrderedDict()
+        self._ignore_repository_names = ignore_repository_names
 
+        repositories = repositories or []
```
neersighted, line 48: I'd rather write this as

```python
if repositories is None:
    repositories = []
```

It's more correct as if we use a boolean evaluation, an empty list is false-y. Using the caller's list isn't necessary here, but it doesn't hurt to use the more correct form.

src/poetry/repositories/legacy_repository.py
neersighted: It's indeed a good question. I believe currently this is because `poetry search` only ever historically searched PyPI, and when we refactored repositories the first time (the old code was quite jumbled), `search` was implemented on the base `Repository` class even if it doesn't make a ton of sense on `LegacyRepository` as we only populate it on demand.

If we teach the search command to use `LegacyRepository` and use some sort of lightweight `PackageVersion` structure that doesn't require inspection like creating a `Package`, this is probably worth revisiting (see https://github.com/python-poetry/poetry/issues/2446) -- but that's work for another day. For now a regular comment instead of docstring explaining we're stubbing this out to avoid changing behavior/it's a TODO to rethink search so all source types can support it likely makes more sense.

neersighted:
@b-kamphorst I've pushed up a rebase of your branch that squashes the changes down into mergable commits -- I think this is good to go once we address the final review commits, as long as we keep the history clean.

src/poetry/repositories/pool.py
```diff
@@ -1,59 +1,78 @@
...
-        return self._default
+        return self._contains_priority(Priority.DEFAULT)
 
     def has_primary_repositories(self) -> bool:
-        return self._has_primary_repositories
+        return self._contains_priority(Priority.PRIMARY)
+
+    def _contains_priority(self, priority: Priority) -> bool:
+        return any(
+            prio_repo.priority == priority for prio_repo in self._repositories.values()
```
neersighted, line 66:
```suggestion
            prio_repo.priority is priority for prio_repo in self._repositories.values()
```

Enums should be compared using identity, as per https://docs.python.org/3/library/enum.html#comparisons

src/poetry/repositories/pool.py
```diff
@@ -1,59 +1,78 @@
...
+    from poetry.repositories.repository import Repository
 
-class Pool(Repository):
+
+class Priority(IntEnum):
+    # The order of the members below dictates the actual priority. The first member has
+    # top priority.
+    DEFAULT = auto()
+    PRIMARY = auto()
+    SECONDARY = auto()
```
neersighted, line 26:
```suggestion
    DEFAULT = enum.auto()
    PRIMARY = enum.auto()
    SECONDARY = enum.auto()
```

Very minor nit, but I prefer `enum.auto()` here as `auto()` is a very generic function name.

src/poetry/repositories/pool.py
```diff
@@ -1,59 +1,78 @@
...
-
         for repository in repositories:
             self.add_repository(repository)
 
-        self._ignore_repository_names = ignore_repository_names
-
     @property
     def repositories(self) -> list[Repository]:
-        return self._repositories
+        unsorted_repositories = list(self._repositories.values())
```
neersighted, line 52: We're creating a list here when the view object will work just fine with `sorted()`.

src/poetry/repositories/pool.py
```diff
@@ -1,59 +1,78 @@
...
+
+
+class Pool(AbstractRepository):
     def __init__(
         self,
         repositories: list[Repository] | None = None,
         ignore_repository_names: bool = False,
     ) -> None:
-        super().__init__("poetry-pool")
+        super().__init__("poetry-repository-pool")
```
neersighted, line 41:
```suggestion
        super().__init__("poetry-pool")
```

Let's keep this consistent with the class name -- it can change back in the rename PR.

neersighted (changes requested):
Final batch of requested changes -- should be good to go after we discuss/address these.

src/poetry/repositories/pool.py
```diff
@@ -1,59 +1,78 @@
...
     def repository(self, name: str) -> Repository:
         name = name.lower()
-
-        lookup = self._lookup.get(name)
-        if lookup is not None:
-            return self._repositories[lookup]
-
+        if self.has_repository(name):
+            return self._repositories[name].repository
         raise ValueError(f'Repository "{name}" does not exist.')
```
neersighted, line 76: This should probably be an `IndexError` instead.

src/poetry/repositories/pool.py
```diff
@@ -63,133 +82,62 @@ def add_repository(
...
+                "A repository with (case-independent) name"
+                f" {repository_name} was already added"
+            )
 
-            self._default = True
-            self._repositories.insert(0, repository)
-            for name in self._lookup:
-                self._lookup[name] += 1
+        if default and self.has_default():
+            raise ValueError("Only one repository can be the default")
```
neersighted, line 92:
```suggestion
            raise ValueError("Only one repository can be the default.")
```

src/poetry/repositories/pool.py
```diff
@@ -63,133 +82,62 @@ def add_repository(
...
-        if repository_name in self._lookup:
-            raise ValueError(f"{repository_name} already added")
-
-        if default:
-            if self.has_default():
-                raise ValueError("Only one repository can be the default")
+        if self.has_repository(repository_name):
+            raise ValueError(
+                "A repository with (case-independent) name"
+                f" {repository_name} was already added"
```
neersighted, line 88:
```suggestion
                "A repository with name {repository_name}"
                f" was already added."
```

I think we can drop the "case-independent" bit as we're printing the name we're using in the exception.

b-kamphorst:
> @b-kamphorst I've pushed up a rebase of your branch that squashes the changes down into mergable commits -- I think this is good to go once we address the final review comments, as long as we keep the history clean.

I have processed the suggested changes (in a separate commit for your convenience). If they are in line with your expectations, feel free to squash the final commit into another commit and make this fly :)

neersighted (approved):
LGTM. FWIW, I would suggest getting familiar with interactive rebases and how to leverage the `edit` rebase command (as well as the `git reset` CLI command) to manipulate the history of branches. When you're doing larger, more ambitious PRs like this, you generally want to take the approach you took and make the history as atomic/easy to follow for reviewers as possible.

Then, when it's time for merge, you'll want to squash your commits down. In the simple/smaller PR case a squash of everything is just fine. But in a larger PR like this it's helpful to capture some of the evolutionary steps as commits in history -- which is when you want to master that process of munging history during an interactive rebase.

Issue #6638, opened 2022-09-27
Automated migration to groups during `poetry add --group dev` 


tomwatson1024:
Running `poetry add --group dev` (or `poetry add --dev`) with a dependency that already exists in the `[tool.poetry.dev-dependencies]` section adds a new entry in `[tool.poetry.group.dev.dependencies` that may conflict with the existing entry.

It's easily fixed (updating `[tool.poetry.dev-dependencies]` to `[tool.poetry.group.dev.dependencies]`, and merging the sections if the extra entry has already been added) but it makes the Poetry 1.1 -> 1.2 upgrade painful if you've got lots of repos.

```Dockerfile
# syntax=docker/dockerfile:1.3-labs
FROM python:3.8
RUN pip install poetry==1.2.1
WORKDIR /src
COPY <<"EOF" pyproject.toml
[tool.poetry]
name = ""
version = "0.1.0"
description = ""
authors = [""]

[tool.poetry.dependencies]
python = "^3.8"

[tool.poetry.dev-dependencies]
tomli = "1.0.0"

EOF

# Using `poetry add --group dev` and `poetry add --dev` produce the same result.
RUN poetry add --group dev tomli@2.0.0
RUN cat pyproject.toml && poetry lock
```

produces

```
 > [6/6] RUN cat pyproject.toml && poetry lock:
#14 0.534 [tool.poetry]
#14 0.534 name = ""
#14 0.534 version = "0.1.0"
#14 0.534 description = ""
#14 0.534 authors = [""]
#14 0.534
#14 0.534 [tool.poetry.dependencies]
#14 0.534 python = "^3.8"
#14 0.534
#14 0.534 [tool.poetry.dev-dependencies]
#14 0.534 tomli = "1.0.0"
#14 0.534
#14 0.534 [[tool.poetry.source]]
#14 0.534 name = 'simple'
#14 0.534 url = 'https://pypi.org/simple'
#14 0.534 default = true
#14 0.534
#14 0.534
#14 0.534 [tool.poetry.group.dev.dependencies]
#14 0.534 tomli = "2.0.0"
#14 0.534
#14 1.122 Updating dependencies
#14 1.123 Resolving dependencies...
#14 1.133
#14 1.133 Because  depends on both tomli (2.0.0) and tomli (1.0.0), version solving failed.
```

edit: Removed unnecessary `[tool.poetry.source]` section in Dockerfile.

neersighted:
[...]

PR #6677, opened 2022-10-02
Migrate dev-dependencies when 'add'  is done

rzuckerm:
# Pull Request Check List

Resolves: #6638


src/poetry/console/commands/add.py
```diff
@@ -153,6 +162,11 @@ def handle(self) -> int:
 
         if not packages:
             self.line("Nothing to add.")
+
+            if packages_migrated and not self.option("dry-run"):
```
neersighted, line 166: I'm not sure about writing when we have nothing to add -- it seems more useful to only run when a change is actually made to me.

src/poetry/console/commands/add.py
```diff
@@ -144,6 +145,14 @@ def handle(self) -> int:
 
             section = poetry_content["group"][group]["dependencies"]
 
+            if group == "dev" and "dev-dependencies" in poetry_content:
+                self.line_error(
+                    "<warning>The dev-dependencies are deprecated. "
+                    "Migrating dev-dependencies to group.dev.dependencies</warning>"
```
neersighted, line 151:
```suggestion
                    "<warning>The dev-dependencies section is deprecated."
                    " Automatically rewriting dev-dependencies as group.dev.dependencies</warning>"
```

src/poetry/console/commands/add.py
```diff
@@ -285,3 +299,22 @@ def notify_about_existing_packages(self, existing_packages: list[str]) -> None:
         for name in existing_packages:
             self.line(f"  • <c1>{name}</c1>")
         self.line(self._hint_update_packages)
+
+    def migrate_dev_dependencies(self, poetry_content: dict[str, Any]) -> None:
+        from poetry.factory import Factory
+
+        dev_group = poetry_content["group"]["dev"]["dependencies"]
```
neersighted, line 306: This seems more elaborate than is needed. Is there a reason that we can't do a more direct assignment?

neersighted (changes requested):

src/poetry/console/commands/add.py
rzuckerm: I noticed that with older versions of poetry the dependencies in `pyproject.toml` were not always in canonical form -- e.g, `Sphinx` and `PyGitHub`. `poetry add` doesn't seem to recognize these.

rzuckerm: I'm still wondering if there should be a command to explicitly do the migration.

rzuckerm: That command would be able to deal with things like making sure the dependencies are in canonical form.

rzuckerm: I think this may be necessary since the group `dev` doesn't seem to exist when `dev-dependencies` are used:
```python
(Pdb) 
> /home/rzuckerm/git/public/poetry/src/poetry/console/commands/add.py(234)handle()
-> self.poetry.package.dependency_group(group).remove_dependency(
(Pdb) self.poetry.package.dependency_group(group)
*** ValueError: The dependency group "dev" does not exist.
```

neersighted: I'm really wondering what's in `poetry_content` -- can we perform a direct reassignment instead of trying to re-create it from scratch?

rzuckerm: For this test:
```
test_add_existing_constraint_old_dev_section_migrates[cachy]
```
This is what is in `poetry_content`:
```python
$ poetry run pytest -vvl -k old_dev_section
Test session starts (platform: linux, Python 3.10.6, pytest 7.1.3, pytest-sugar 0.9.5)
cachedir: .pytest_cache
Using --randomly-seed=3658096042
rootdir: /home/rzuckerm/git/public/poetry
plugins: xdist-2.5.0, forked-1.4.0, cov-3.0.0, randomly-3.12.0, mock-3.8.2, sugar-0.9.5
collecting ... 
>>>>>>>>>>>>>>>>>>> PDB set_trace (IO-capturing turned off) >>>>>>>>>>>>>>>>>>>>
> /home/rzuckerm/git/public/poetry/src/poetry/console/commands/add.py(301)migrate_dev_dependencies()
-> dev_group = poetry_content["group"]["dev"]["dependencies"]
(Pdb) poetry_content
{'name': 'simple-project', 'version': '1.2.3', 'description': 'Some description.', 'authors': ['Sébastien Eustace <sebastien@eustace.io>'], 'license': 'MIT', 'readme': ['README.rst'], 'homepage': 'https://python-poetry.org', 'repository': 'https://github.com/python-poetry/poetry', 'documentation': 'https://python-poetry.org/docs', 'keywords': ['packaging', 'dependency', 'poetry'], 'classifiers': ['Topic :: Software Development :: Build Tools', 'Topic :: Software Development :: Libraries :: Python Modules'], 'dependencies': {'python': '~2.7 || ^3.4'}, 'scripts': {'foo': 'foo:bar', 'baz': 'bar:baz.boom.bim', 'fox': 'fuz.foo:bar.baz'}, 'dev-dependencies': {'cachy': '^0.2.0'}, 'group': {'dev': {'dependencies': {}}}}
```

Also, as I showed above, the `dev` group does not yet exist. If I just do a direct assignment, then my new tests fail.

src/poetry/console/commands/add.py
```diff
@@ -285,3 +293,22 @@ def notify_about_existing_packages(self, existing_packages: list[str]) -> None:
...
+        dev_dependencies = poetry_content["dev-dependencies"]
+        for name, constraint in dev_dependencies.items():
+            canonicalized_name = canonicalize_name(name)
```
dimbleby, line 303: I don't see any reason to canonicalize names as part of this migration.  

- At best it's a completely separate concern than where the dependencies are declared
- but anyway non-canonical names are fine here

dimbleby: Also I think this mechanism is going to lose any comments that people have left in their dependency listing.

The code should surely move the dependencies wholesale from one place to the other, rather than reconstructing them one at a time - then I expect it should be possible to preserve comments.

neersighted: Agreed, and what I was trying to get at in https://github.com/python-poetry/poetry/pull/6677#discussion_r985172404. I still don't see why this isn't possible, but I haven't had time to pull this down locally and play with it myself.

rzuckerm: Non-canonical names are definitely an issue. Older versions of poetry did not handle this properly. For instance, `porg add sphinx` would add `Sphinx = "^5.3.2"` to `pyproject.toml`. If I don't use canonical names, then there is a possibility of duplicate entries. I observed this in one of my new tests:

```toml
[tool.poety.group.dev.dependencies]
Cachy = "^0.2.0"
cachy = "^0.2.0"
```

Not good ☹️

dimbleby: if I have a `pyproject.toml` that asks for `Cachy` then I will be surprised and perhaps disappointed if this migration changes that to `cachy` - it simply shouldn't need to do that.

I don't understand where your two `cachy`s come from.  Are you trying to `poetry add cachy` to a project that already holds `Cachy`?  or what?

Conceptually this MR should be simpler than you are making it I think:

- if `poetry.group.dev.dependencies` and `poetry.dev-dependencies` are both present then bail, perhaps with a log saying that the user probably wants to sort that out manually
- it `poetry.dev-dependencies` exists, then pop that out of the dictionary and re-insert it as `poetry-group.dev.dependencies`
- that's it

rzuckerm: Non-canonical package names come from older versions of poetry. For example, `sphinx` is added as `Sphinx`:
```toml
pytest = "^5.2"
Sphinx = "^5.2.3"
```

dimbleby: yes, but so what?  that doesn't answer either (i) why you are getting two names for `cachy` or (ii) why this migration should canonicalize those names

dimbleby: I think I see the confusion in your unit tests.

After going `app.poetry.file.write(content)` you also need to `app.reset_poetry()` or else explicitly update the `poetry` object that the tests are passing around with the new dependency.

Once you do that you'll find that you don't need to do the same from within poetry src.

Then indeed the entire implementation of `migrate_dev_dependencies` can be
```python
        dev_dependencies = poetry_content.pop("dev-dependencies")
        poetry_content["group"]["dev"]["dependencies"] = dev_dependencies
```
(which is hardly worth a separate method IMO)

rzuckerm: @dimbleby As I said earlier, the `dev` group doesn't get created when there is the `dev-dependencies` section. The rest of the `add` command seems to need that to function properly. Therefore, I think I still have to have the loop. As for why the two `cachy`'s, I have seen something similar when migrating from 1.1x to 1.2. I'm trying to come up with a better example to illustrate this.

Thanks for the tip on `app.reset_poetry()`. I'll try that.

rzuckerm: Here's what I'm seeing with the double `cachy`:
```python
tests/console/commands/test_add.py:2213: AssertionError


>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>> traceback >>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>

name = 'Cachy'
app = <tests.helpers.PoetryTestApplication object at 0x7f30401fcf40>
repo = <tests.helpers.TestRepository object at 0x7f303f833130>
tester = <cleo.testers.command_tester.CommandTester object at 0x7f303f809ab0>

    @pytest.mark.parametrize("name", ["cachy", "Cachy"])
    def test_add_existing_constraint_old_dev_section_migrates(
        name: str, app: PoetryTestApplication, repo: TestRepository, tester: CommandTester
    ):
        # Note: Not testing non-canonical names since this is currently broken
        # for this scenario
    
        repo.add_package(get_package("cachy", "0.1.0"))
        repo.add_package(get_package("cachy", "0.2.0"))
    
        content = app.poetry.file.read()
        content["tool"]["poetry"]["dev-dependencies"] = {name: "^0.2.0"}
        app.poetry.file.write(content)
        app.reset_poetry()
    
        tester.execute("cachy^0.2.0 -G dev")
    
        warning = """\
    The dev-dependencies section is deprecated. \
    Automatically rewriting dev-dependencies as \
    group.dev.dependencies
    """
        expected = """\
    
    Updating dependencies
    Resolving dependencies...
    
    Writing lock file
    
      • Installing cachy (0.2.0)
    """
    
        assert tester.io.fetch_error() == warning
    
        content = app.poetry.file.read()["tool"]["poetry"]
    
        assert "dev-dependencies" not in content
        assert "cachy" in content["group"]["dev"]["dependencies"]
>       assert content["group"]["dev"]["dependencies"] == {"cachy": "^0.2.0"}
E       AssertionError: assert {'Cachy': '^0.2.0', 'cachy': '^0.2.0'} == {'cachy': '^0.2.0'}
E         Common items:
E         {'cachy': '^0.2.0'}
E         Left contains 1 more item:
E         {'Cachy': '^0.2.0'}
E         Full diff:
E         - {'cachy': '^0.2.0'}
E         + {'Cachy': '^0.2.0', 'cachy': '^0.2.0'}
```

The call to `app.reset_poetry()` didn't help for the test that where `Cachy = "^0.2.0"` is in `[tool.poetry.dev-dependencies]`. Also, I see other writes to `pyproject.toml` in this file, and none of them are calling that method. For example, [this](https://github.com/python-poetry/poetry/blob/master/tests/console/commands/test_add.py#L1000).

rzuckerm:
@dimbleby OK, I simplified the migrate logic as suggested. I also removed all the non-canonical name tests since, I believe, that is a separate issue that I seem to have trouble reproducing.

rzuckerm:
Any opinions on where in the documentation I need to add something about automatically migrating `dev-depdendencies` to `group.dev.dependencies`? Would I add this to the CLI section for the `add` command? Would I also need to add this to the `Managing Dependencies` section (since there's already a note there about `dev-dependencies`?

src/poetry/console/commands/add.py
```diff
@@ -285,3 +293,7 @@ def notify_about_existing_packages(self, existing_packages: list[str]) -> None:
         for name in existing_packages:
             self.line(f"  • <c1>{name}</c1>")
         self.line(self._hint_update_packages)
+
+    def migrate_dev_dependencies(self, poetry_content: dict[str, Any]) -> None:
```
neersighted, line 297: This is now what I had in mind earlier. However, this method shouldn't be exported -- so it should either be `_migrate_dev_dependencies`, or just inlined (which I think is better given how trivial it is).

neersighted (changes requested):

neersighted:
> Any opinions on where in the documentation I need to add something about automatically migrating `dev-depdendencies` to `group.dev.dependencies`? Would I add this to the CLI section for the `add` command? Would I also need to add this to the `Managing Dependencies` section (since there's already a note there about `dev-dependencies`?

I would suggest you pick a canonical source to describe the migration in (e.g. in `Managing Dependencies`, and then add `{{ note }}`s where appropriate linking back.

dimbleby:
You should also handle the case where both `poetry.group.dev.dependencies` and `poetry.dev-dependencies` are present - currently you overwrite `poetry.group.dev.dependencies`, which isn't what is wanted. 

As above I'd probably just bail out in this case: but if you are motivated to test whether it works then maybe this will do the job
```python
poetry_content["group"]["dev"]["dependencies"].update(dev_dependencies)
```

The tests that you've added don't test the things that I'd be worried about.  Mostly they are testing that `poetry add` still adds the expected things, which this was never likely to break.  I'd prefer to see them testing that the `pyproject.toml` file is updated in the expected way, including preservation of comments.  But maybe this is OK, maintainers will say.

To be honest I'm also not entirely sure that this whole flow is the right one, it feels a bit odd to me to do this only in such specific circumstances and without being asked to do it.  But I don't know that I care enough, I'll again leave it to others to worry about such things.

NB you need review from someone with the power to merge - which is not me.

rzuckerm:
@dimbleby Preservation of comments looks like it is going to be extremely difficult. It seems like `.update()` doesn't preserve comments at all ☹️

rzuckerm:
Hmm, the tests seem to be a bit flaky on Windows for python 3.9. I don't think my change has caused this.

dimbleby:
does just moving the dictionary around preserve comments?  I'd hope that it does.  If `update()` doesn't, then that seems like a good reason not to do it (ie not to attempt to merge the two sections if they both exist).

re "without being asked to do it" I didn't mean that there should be an interactive prompt, I was thinking that maybe an explicit "poetry migrate" or whatever would make sense.  

re "such specific circumstances": it's not clear to me why we only fix up the sections when adding a dev dependency.  Why not also when removing a dev dependency, or when adding a non-dev dependency, or when re-locking?  I don't know what the right answers are, perhaps this is fine.  DId you consider such choices?

rzuckerm:
@dimbleby I actually prefer the idea of a `migrate` command. However, the issue was that the `add` command did something unexpected when `dev-dependencies` was present. Please see #6638 and comment on it. Maybe I jumped the gun trying my hand at implementing a solution. I'd like to hear from the maintainers on what they think before proceeding any further on the PR.

dimbleby:
another 
approach would be to perform this migration whenever we write to pyproject.toml, I think this means in https://github.com/python-poetry/poetry-core/blob/d007227f1320421a46cae6aae50296945433b9c0/src/poetry/core/pyproject/toml.py#L98

rzuckerm:
> another approach would be to perform this migration whenever we write to pyproject.toml, I think this means in https://github.com/python-poetry/poetry-core/blob/d007227f1320421a46cae6aae50296945433b9c0/src/poetry/core/pyproject/toml.py#L98

Interesting idea, but it doesn't fix the problem that `add` doesn't respect `dev-dependencies` which was the original issue. I'm still waiting for the maintainers to weigh in on this.

dimbleby:
Yeah, always making the change on write isn't quite good enough because it fails in exactly the `add` case that this MR is trying to address.  (By the time we come to write, poetry will have created both sections in the configuration)

What would work, perhaps, would be always to make the adjustment on _read_.  Then the migrated configuration would be written back, if it happened to be one of the commands that writes configuration.

rzuckerm:
I like that idea. Now, the issue would be how to do this without losing comments or formatting.

dimbleby:
>  how to do this without losing comments or formatting.

this should Just Work, tomlkit ought to preserve such things automatically.  Did you test it?

rzuckerm:
I just tried a quick experiment, and direct assignment does seem to preserve comments and formatting. It may be possible to combine `dev-dependencies` with an existing `group.dev.dependencies` using some combination of `tomlkit.parse` and the `as_string` method.

Issue #6633, opened 2022-09-26
`poetry build` fails sdist build when using multiple `readme` entries


ekralc:
- **Poetry version**: 1.2.1
- **Python version**: 3.8.10
- **OS version and name**: Ubuntu 20.04 (running from a Docker container on macOS 12.4)
- **pyproject.toml**: https://gist.github.com/ekralc/c97f8880a94e54a901d7562b3eaebe92

- [x] I am on the [latest](https://github.com/python-poetry/poetry/releases/latest) stable Poetry version, installed using a recommended method.
- [x] I have searched the [issues](https://github.com/python-poetry/poetry/issues) of this repo and believe that this is not a duplicate.
- [x] I have consulted the [FAQ](https://python-poetry.org/docs/faq/) and [blog](https://python-poetry.org/blog/) for any relevant entries or release notes.
- [x] If an exception occurs when executing a command, I executed it again in debug mode (`-vvv` option) and have included the output below.

## Issue

According to the current documentation, you can add multiple entries for `readme` in `pyproject.toml`
https://github.com/python-poetry/poetry/blob/31ef8d781607505d9bd581647df056a1a637ddea/docs/pyproject.md?plain=1#L129-L131

The current example in `README.md` also includes multiple files:
https://github.com/python-poetry/poetry/blob/424ea0c59005df761a9b4fc542c04d513e402046/README.md?plain=1#L32-L33

However `poetry build` fails when using this pattern.

When using an array of readme entries, `poetry build` throws the following error:
```
Building test (0.1.0)
  - Building sdist

unhashable type: 'Array'
```

As a new user, this error was a bit confusing...so I had a look at the verbose output:

```
Building test (0.1.0)
  - Building sdist
  - Adding: /test/test/__init__.py

  Stack trace:

  10  /usr/local/lib/python3.8/dist-packages/cleo/application.py:329 in run
       327│
       328│             try:
     → 329│                 exit_code = self._run(io)
       330│             except Exception as e:
       331│                 if not self._catch_exceptions:

   9  /usr/local/lib/python3.8/dist-packages/poetry/console/application.py:185 in _run
       183│         self._load_plugins(io)
       184│
     → 185│         exit_code: int = super()._run(io)
       186│         return exit_code
       187│

   8  /usr/local/lib/python3.8/dist-packages/cleo/application.py:423 in _run
       421│             io.input.set_stream(stream)
       422│
     → 423│         exit_code = self._run_command(command, io)
[...]

PR #6678, opened 2022-10-02
Add tests for pyproject with multiple readme files

wagnerluis1982:
# Pull Request Check List

**tests/masonry/builders/test_editable_builder.py**
- Checks if content of all readme files are printed to the metadata

**tests/console/commands/test_build.py** (new file)
- Checks sdist file exists and is not empty
- Checks wheel file exists and is not empty
- Checks inside sdist tar file if contains all readme files

Closes: #6633
Depends on: python-poetry/poetry-core#551

- This PR only add **tests**


wagnerluis1982:
Question to maintainers, does poetry have any integration test for `poetry build` command? I was not able to found any.

radoering:
Probably not. (It should be located at `tests/console/commands`.)

Since the test in this PR is passing even without the poetry-core fix, I assume an integration test is required to provoke the issue?

wagnerluis1982:
@radoering Exactly, that's why I was looking for an integration test for `build`.

I will create a new test module then.

wagnerluis1982:
Just added a test to verify `poetry build` command.

NOTE: the pipeline is going to fail until python-poetry/poetry-core#486 is merged.

wagnerluis1982:
Hi @radoering, after you merged python-poetry/poetry-core#486, I expected the tests to not fail (I just rebased after `master`).

I expected the pipeline to use latest poetry-core version. Does that happen?

radoering:
No, it just uses the locked version of poetry-core (no magic). Normally, we just update the locked version after a core release.

wagnerluis1982:
> No, it just uses the locked version of poetry-core (no magic). Normally, we just update the locked version after a core release.

No magic 🙁

In that case, it would be advisable to reopen issue #6633 to signalize users, IMHO.

wagnerluis1982:
@Secrus the tests here are going to fail while next poetry-core isn't released.

wagnerluis1982:
@radoering when you have a chance, please check this.

Now that poetry-core 1.5.0 is locked, the tests are passing.

tests/masonry/builders/test_editable_builder.py
```diff
@@ -287,6 +296,44 @@ def test_builder_installs_proper_files_when_packages_configured(
...
+Description-Content-Type: text/x-rst
+
+Single Python
+=============
+
+Changelog
+=========
+
+"""
+    assert metadata == dist_info.joinpath("METADATA").read_text(encoding="utf-8")
```
radoering, line 334:
```suggestion
    assert dist_info.joinpath("METADATA").read_text(encoding="utf-8") == metadata
```
We prefer actual-expected order.

tests/fixtures/with_multiple_readme_files/pyproject.toml
```diff
@@ -0,0 +1,19 @@
+[tool.poetry]
+name = "my-package"
+version = "0.1"
+description = "Some description."
+authors = [
+    "Wagner Macedo <wagnerluis1982@gmail.com>"
```
radoering, line 6: Can you replace author information by Sébastiens or some synthetic data, please? We want to avoid introducing any new real-world information in test data.

radoering (changes requested):
Just some minor nitpicks.

tests/masonry/builders/test_editable_builder.py
wagnerluis1982: Got it. Done :slightly_smiling_face:

tests/fixtures/with_multiple_readme_files/pyproject.toml
wagnerluis1982: Makes sense, I liked this.

Done!

radoering (approved):

Issue #5205, opened 2022-02-16
Multi-architecture wheel selection across disperate sources (with markers)


merberich:
- [x] I am on the [latest](https://github.com/python-poetry/poetry/releases/latest) Poetry version.
- [x] I have searched the [issues](https://github.com/python-poetry/poetry/issues) of this repo and believe that this is not a duplicate.
- [x] If an exception occurs when executing a command, I executed it again in debug mode (`-vvv` option).

- **OS version and name**: Ubuntu 18.04 on both aarch64 and x86_64
- **Poetry version**: 1.1.12
- **Link of a [Gist](https://gist.github.com/) with the contents of your pyproject.toml file**: (see below for variants)

## Issue
I'm trying to build a project that will be used in multi-architecture environments (aarch64 and x86_64 specifically). One of my project's dependencies (`jaxlib`) only offers prebuilt wheels for x86_64 via PiPy. As a solution, I've built and hosted the wheel for aarch64 separately, which necessitates adding secondary dependency sources in `pyproject.toml` like so:
```toml
[tool.poetry]
name = "temp"
version = "0.1.0"
description = ""
authors = ["merberich <merberich@gmail.com>"]

[tool.poetry.dependencies]
python = "^3.9"
cupy-cuda102 = "^10.1.0"
jaxlib = [
  { version = "0.1.71" },
  { url = "<valid-url-to-aarch64-wheel-host>" }
]
jax = "0.2.20"

[tool.poetry.dev-dependencies]

[build-system]
requires = ["poetry-core>=1.0.0"]
build-backend = "poetry.core.masonry.api"
```

Building a new `poetry.lock` for this project produces `jaxlib` references:
```
cat poetry.lock | grep jaxlib
cpu = ["jaxlib (==0.1.71)"]
cuda102 = ["jaxlib (==0.1.71+cuda102)"]
cuda111 = ["jaxlib (==0.1.71+cuda111)"]
minimum-jaxlib = ["jaxlib (==0.1.69)"]
tpu = ["jaxlib (==0.1.71)", "libtpu-nightly (==0.1.dev20210809)", "requests"]
name = "jaxlib"
jaxlib = [
    {file = "jaxlib-0.1.71-cp37-none-macosx_10_9_x86_64.whl", hash = "sha256:ad36895ceb68782bb74f0657da127cece06c19ce1e72ad9c660f79bce549618e"},
[...]

PR #6679, opened 2022-10-02
provider: do not merge dependencies from different sources

radoering:
# Pull Request Check List

Resolves: #5205

Especially https://github.com/python-poetry/poetry/issues/5205#issuecomment-1063671665

Side effect: the order of dependencies will no longer get mixed up (see adapted test in `test_solver.py`)

### DIFF
--- a/src/poetry/puzzle/provider.py
+++ b/src/poetry/puzzle/provider.py
@@ -670,19 +670,29 @@ class Provider:
 
             self.debug(f"<debug>Duplicate dependencies for {dep_name}</debug>")
 
-            non_direct_origin_deps: list[Dependency] = []
-            direct_origin_deps: list[Dependency] = []
-            for dep in deps:
-                if dep.is_direct_origin():
-                    direct_origin_deps.append(dep)
-                else:
-                    non_direct_origin_deps.append(dep)
-            deps = (
-                self._merge_dependencies_by_constraint(
-                    self._merge_dependencies_by_marker(non_direct_origin_deps)
+            # Group dependencies for merging.
+            # We must not merge dependencies from different sources!
+            dep_groups = self._group_by_source(deps)
+            deps = []
+            for group in dep_groups:
+                # In order to reduce the number of overrides we merge duplicate
+                # dependencies by constraint. For instance, if we have:
+                #   - foo (>=2.0) ; python_version >= "3.6" and python_version < "3.7"
+                #   - foo (>=2.0) ; python_version >= "3.7"
+                # we can avoid two overrides by merging them to:
+                #   - foo (>=2.0) ; python_version >= "3.6"
+                # However, if we want to merge dependencies by constraint we have to
+                # merge dependencies by markers first in order to avoid unnecessary
+                # solver failures. For instance, if we have:
+                #   - foo (>=2.0) ; python_version >= "3.6" and python_version < "3.7"
+                #   - foo (>=2.0) ; python_version >= "3.7"
+                #   - foo (<2.1) ; python_version >= "3.7"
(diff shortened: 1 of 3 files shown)


src/poetry/puzzle/provider.py
```diff
@@ -947,6 +940,22 @@ def debug(self, message: str, depth: int = 0) -> None:
...
 
+    def _group_by_source(
+        self, dependencies: Iterable[Dependency]
+    ) -> list[list[Dependency]]:
+        groups: list[list[Dependency]] = []
+        for dep in dependencies:
+            for group in groups:
+                if (
+                    dep.is_same_source_as(group[0])
+                    and dep.source_name == group[0].source_name
```
neersighted, line 951: I find it somewhat surprising that `is_same_source_as` doesn't compare `source_name` -- however, I expect that that is for a good reason?

In general, that method could use a good docstring and more detailed comments -- however, that is a discussion for a different day (and repo).

neersighted (approved):

src/poetry/puzzle/provider.py
```diff
@@ -947,6 +940,22 @@ def debug(self, message: str, depth: int = 0) -> None:
 
             self._io.write(debug_info)
 
+    def _group_by_source(
```
neersighted, line 943: I just realized we don't have a docstring here -- could you add one describing the intention of this method and how it's meant to be used?

The battle against no source code level documentation depends on each PR :smile:

src/poetry/puzzle/provider.py
```diff
@@ -670,19 +670,12 @@ def complete_package(
...
-            direct_origin_deps: list[Dependency] = []
-            for dep in deps:
-                if dep.is_direct_origin():
-                    direct_origin_deps.append(dep)
-                else:
-                    non_direct_origin_deps.append(dep)
-            deps = (
-                self._merge_dependencies_by_constraint(
-                    self._merge_dependencies_by_marker(non_direct_origin_deps)
+            dep_groups = self._group_by_source(deps)
```
neersighted, line 673: Similarly, a comment describing the intended merge logic (and why) would be valuable here. This refactoring is much more readable than the old version, and a comment will make it so even new contributors can digest it.

neersighted (changes requested):

src/poetry/puzzle/provider.py
radoering: At first, I found it surprising, too. However, after further thinking it makes sense because `is_same_source_as` is a method of `PackageSpecification` and `source_name` is an attribute of Dependency which is a subclass of `PackageSpecification`. We could override `is_same_source_as` for `Dependency` to consider `source_name` if a Dependency is compared to another `Dependency` and not consider it if a `Dependency` is compared to a `Package` or `PackageSpecification`.

neersighted: Ah, that makes sense -- I think this change would be easier to reason about if we got that into core first.

radoering: python-poetry/poetry-core#488

radoering: It seems like we can't change that easily. See failing downstream tests in the core PR. Sometimes `source_name` is not relevant because dependencies created from packages via `to_dependency()` can't have a source name and the solver still has to consider such dependencies equal to dependencies with source name.

neersighted: Looks like a battle for another day, then. Could you update your core PR to instead explain that gotcha in a comment somewhere for future reference?

neersighted (approved):

PR #1954, opened 2020-01-28
#453 Add command to check poetry.lock freshness

cliebBS:
Fixes #453 

This PR adds a new command, `poetry lock --check`, which returns `0` if `poetry.lock` is consistent with `pyproject.toml` and `1` if it is not.  This is useful for CI/CD tooling to have a very fast way of checking the lockfile before building without the re-resolution of the dependency tree that `poetry lock` performs.

# Pull Request Check List

This is just a reminder about the most common mistakes. Please make sure that you tick all *appropriate* boxes.  But please read our [contribution guide](https://python-poetry.org/docs/contributing/) at least once, it will save you unnecessary review cycles!

- [x] Added **tests** for changed code.
- [x] Updated **documentation** for changed code.

**Note**: If your Pull Request introduces a new feature or changes the current behavior, it should be based
on the `develop` branch. If it's a bug fix or only a documentation update, it should be based on the `master` branch.

If you have *any* questions to *any* of the points above, just **submit and ask**!  This checklist is here to *help* you, not to deter you from contributing!

PR #6704, opened 2022-10-03
Document that `poetry lock --check` validates pyproject.toml

johnthagen:
Document that the behavior of `poetry lock --check` also essentially runs `poetry check`:

```
$ poetry lock --check

  - 'name' is a required property
  - Additional properties are not allowed ('nam' was unexpected)
```

### DIFF
--- a/docs/cli.md
+++ b/docs/cli.md
@@ -630,7 +630,7 @@ poetry lock
 
 ### Options
 
-* `--check`: Verify that `poetry.lock` is consistent with `pyproject.toml`
+* `--check`: Verify that `poetry.lock` is consistent with `pyproject.toml` and validates the structure of the `pyproject.toml` file.
 * `--no-update`: Do not update locked versions, only refresh lock file.
 
 ## version


neersighted:
This is more a consequence of the fact that loading the config (basically `Poetry()`/the old `Factory.create_poetry()`) runs the validation as it creates a full application instance.

I'm saying that most commands will validate pyproject.toml since we don't read it lazily to populate the internal representation, but I'm not sure there's a good way to communicate that. This is correct, just not exhaustive.

johnthagen:
My motivation is to be able to point co-workers at a canonical, single command to validate Poetry config and lockfile freshness. I think ideally I agree with Sebastian's comment from last year that this would have been implemented as `poetry check --lock`:

- https://github.com/python-poetry/poetry/pull/1954#issuecomment-800096650

In terms of the status quo, should developers not rely on the side effect that `poetry lock --check` also invokes the internals of `poetry check`? Would the recommendation in CI be running both:

```
poetry check
poetry lock --check
```

I'd like be sure I'm not recommending relying on some internal behavior of Poetry.

neersighted:
I wouldn't mind deprecating `lock --check` in favor of `check --lock` and moving all the code there/making the deprecated usage simply an alias. I think that would be the most obvious and clean solution for Poetry going forward.

johnthagen:
Closing this in favor of

- #6756

johnthagen:
In case anyone comes across this, I learned today that `poetry lock --check` does _not_ in fact validate everything that `poetry check` does. For example, Trove classifiers are validated in `poetry check` but not `poetry lock --check`.

So, to get the full amount of checking, a user currently needs to run two steps in CI:

```
poetry check
poetry lock --check
```

As mentioned above, this UX would be improved with:

- #6756

Issue #6705, opened 2022-10-03
self commands including 'add <plugin>' should facilitate using an index URL


colindean:
- [X] I have searched the [issues](https://github.com/python-poetry/poetry/issues) of this repo and believe that this is not a duplicate.
- [X] I have searched the [FAQ](https://python-poetry.org/docs/faq/) and general [documentation](https://python-poetry.org/docs/) and believe that my question is not already covered.

## Feature Request

I'm preparing for the eventuality of the export plugin not being installed by default. When running `poetry self add poetry-plugin-export`, I noticed that the files were being downloaded from pythonhosted.org instead of my company's repos. I'd prefer that poetry retrieve artifacts from my company's internal repos.

### Expectation

⭐ ⭐ ⭐ ⭐ ⭐  = A command line option would be great:

    poetry self add poetry-plugin-export --index-url https://example.com/proxies/pypi/simple

⭐ ⭐ ⭐ ⭐  = But I could also see this being a one-time setup thing:

    poetry self config pip.global.index-url https://example.com/proxies/pypi/simple

⭐ ⭐ ⭐  = but this _probably_ works for now until something nicer exists:

    ${POETRY_HOME}/bin/pip3 config --site set global.index-url https://example.com/proxies/pypi/simple

But I don't have a good way for `poetry` to tell me its venv path for scripting, e.g. something like

    POETRY_HOME=$(poetry self venv)

or to run something in poetry's venv with a command like

    poetry self run pip config …

### Related:

* https://github.com/python-poetry/poetry/pull/4944
* https://github.com/python-poetry/poetry/issues/6529

neersighted:
Keep in mind that `pip` is an implementation detail that we are slowly removing from the codebase, so any attempt to support this cannot couple to any details of pip directly.

I think in general the aim is going to be allowing plugins to be installed outside the main Poetry environment (e.g. somehow have a `required-plugins` section in pyproject.toml) and make them use the sources configured there. But some facilities to manage sources for the global/instance pyproject.toml certainly wouldn't hurt either.

colindean:
> allowing plugins to be installed outside the main Poetry environment (e.g. somehow have a required-plugins section in pyproject.toml

This is _definitely_ the more desirable solution for me. I think we discussed this on Discord a while back!

Maybe this invocation would adequately hide pip:

    poetry self config plugins.index-url https://example.com/proxies/pypi/simple

[...]

PR #6708, opened 2022-10-04
Adds 'self source add|remove|show' for managing Poetry plugin sources

colindean:
This reuses the existing command infra provided by the `source` command, using the SelfCommand infra to change which pyproject.toml is the target.

I tried to write some tests for this, but realized that the tests would be modifying the developer's valued Poetry system pyproject.toml. I think it might be OK to rely on the existing source add, remove, and show tests, but I'll share my test code in a comment on the PR for this branch.

## Pull Request Check List

Resolves: #6705 

### DIFF
--- a/src/poetry/console/commands/self/self_command.py
+++ b/src/poetry/console/commands/self/self_command.py
@@ -57,7 +57,7 @@ class SelfCommand(InstallerCommand):
     def activated_groups(self) -> set[str]:
         return {self.default_group}
 
-    def generate_system_pyproject(self) -> None:
+    def generate_system_pyproject(self) -> str:
         preserved = {}
 
         if self.system_pyproject.exists():
@@ -77,11 +77,16 @@ class SelfCommand(InstallerCommand):
         for key in preserved:
             content["tool"]["poetry"][key] = preserved[key]  # type: ignore[index]
 
-        self.system_pyproject.write_text(content.as_string(), encoding="utf-8")
+        return content.as_string()
+
+    def overwrite_system_pyproject(self) -> None:
+        self.system_pyproject.write_text(
+            self.generate_system_pyproject(), encoding="utf-8"
+        )
 
     def reset_poetry(self) -> None:
         with directory(self.system_pyproject.parent):
-            self.generate_system_pyproject()
+            self.overwrite_system_pyproject()
             self._poetry = Factory().create_poetry(
                 self.system_pyproject.parent, io=self.io, disable_plugins=True
             )
--- a/src/poetry/console/commands/self/source/add.py
+++ b/src/poetry/console/commands/self/source/add.py
@@ -1,0 +1,19 @@
+from __future__ import annotations
+
+from poetry.console.commands.self.self_command import SelfCommand
+from poetry.console.commands.source.add import SourceAddCommand
+
+
+class SelfSourceAddCommand(SelfCommand, SourceAddCommand):
+    name = "self source add"
+    description = "Add additional sources to Poetry's runtime environment."
+    options = [
+        o for o in SourceAddCommand.options if o.name in {"default", "secondary"}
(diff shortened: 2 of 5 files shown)


colindean:
This was a stab at a test for `self source add` but I realized that I couldn't easily avoid modifying the developer's in-use Poetry system configuration.

Some of the minor refactorings in this PR resulted from attempting to support an in-memory modification of the config, but I realized I couldn't do that without what seemed like a lot more refactoring.

```python

import pytest

from poetry.console.commands.self.self_command import SelfCommand
from poetry.factory import Factory
from tests.console.commands.self.utils import get_self_command_dependencies

from ..source.conftest import source_one, source_existing


if TYPE_CHECKING:
    from cleo.testers.command_tester import CommandTester

    from poetry.config.source import Source
    from poetry.poetry import Poetry
    from tests.types import CommandTesterFactory, ProjectFactory


@pytest.fixture
def poetry_with_source(project_factory: ProjectFactory) -> Poetry:
    self_cmd = SelfCommand()
    return project_factory(pyproject_content=self_cmd.generate_system_pyproject())


@pytest.fixture
def tester(
        command_tester_factory: CommandTesterFactory, poetry_with_source: Poetry
) -> CommandTester:
    return command_tester_factory("self source add", poetry=poetry_with_source)


def assert_source_added(
        tester: CommandTester,
        poetry: Poetry,
        source_existing: Source,
        source_added: Source,
) -> None:
    assert (
            tester.io.fetch_output().strip()
            == f"Adding source with name {source_added.name}."
    )
    # poetry.pyproject.poetry_config.reload()
    sources = poetry.get_sources()
    assert sources == [source_existing, source_added]
    assert tester.status_code == 0


def test_source_add_simple(
        tester: CommandTester,
        source_one: Source,
        source_existing: Source,
        poetry_with_source: Poetry,
):
    tester.execute(f"{source_one.name} {source_one.url}")
    assert_source_added(tester, poetry_with_source, source_existing, source_one)

```

src/poetry/console/commands/self/self_command.py
```diff
@@ -57,7 +57,7 @@ def default_group(self) -> str:
     def activated_groups(self) -> set[str]:
         return {self.default_group}
 
-    def generate_system_pyproject(self) -> None:
+    def generate_system_pyproject(self) -> str:
```
colindean, line 60: This was a refactor to support generating the system pyproject.toml and modifying it in memory that turned out not to be true. I left the refactor because it felt cleaner afterward.

src/poetry/console/commands/self/source/add.py
```diff
@@ -0,0 +1,19 @@
+from __future__ import annotations
+
+from poetry.console.commands.self.self_command import SelfCommand
+from poetry.console.commands.source.add import SourceAddCommand
+
+
+class SelfSourceAddCommand(SelfCommand, SourceAddCommand):
```
colindean, line 7: I followed the pattern established in `src/poetry/console/commands/self/add.py`.

src/poetry/console/commands/self/source/add.py
```diff
@@ -0,0 +1,19 @@
...
+    name = "self source add"
+    description = "Add additional sources to Poetry's runtime environment."
+    options = [
+        o for o in SourceAddCommand.options if o.name in {"default", "secondary"}
```
colindean, line 11: I think this options filter might not be necessary since we want these commands to be cloned 100%.

abn:
@colindean is this something you are still interested in?

I am wondering if we need to add more commands for this. Can using `pipx inject` be the right way for this use case? `pipx` already supports private repos via `pip` and since `pipx` is our preferred way of installation right now, it might make sense to favour that.

Secrus:
Stalled

Issue #6668, opened 2022-10-01
Poetry stuck when installing GCP keyring plugin


snthibaud:
- **Poetry version**: 1.2.1
- **Python version**: 3.9.2
- **OS version and name**: Debian GNU/Linux 11 (bullseye)
- **pyproject.toml**: not relevant

- [x] I am on the [latest](https://github.com/python-poetry/poetry/releases/latest) stable Poetry version, installed using a recommended method.
- [x] I have searched the [issues](https://github.com/python-poetry/poetry/issues) of this repo and believe that this is not a duplicate.
- [x] I have consulted the [FAQ](https://python-poetry.org/docs/faq/) and [blog](https://python-poetry.org/blog/) for any relevant entries or release notes.
- [x] If an exception occurs when executing a command, I executed it again in debug mode (`-vvv` option) and have included the output below.

## Issue
I have been trying to install the GCP keyring plugin, but the installer got stuck. It has been resolving for an entire night, so I reran it with -vvv and noticed it stops after indicating it is creating a new session for files.pythonhosted.org . This is the full output:
```

xyz@xyz:~ $ poetry self add keyrings.google-artifactregistry-auth -vvv
[keyring.backend] Loading KWallet
[keyring.backend] Loading SecretService
[keyring.backend] Loading Windows
[keyring.backend] Loading chainer
[keyring.backend] Loading libsecret
[keyring.backend] Loading macOS
Creating new session for pypi.org
Source (PyPI): 6 packages found for keyrings-google-artifactregistry-auth *
Using version ^1.1.1 for keyrings.google-artifactregistry-auth

Updating dependencies
Resolving dependencies...
   1: fact: poetry-instance is 1.2.1
   1: derived: poetry-instance
   1: fact: poetry-instance depends on poetry (1.2.1)
   1: fact: poetry-instance depends on keyrings.google-artifactregistry-auth (^1.1.1)
   1: selecting poetry-instance (1.2.1)
   1: derived: keyrings.google-artifactregistry-auth (>=1.1.1,<2.0.0)
   1: derived: poetry (==1.2.1)
Creating new session for pypi.org
Source (PyPI): 1 packages found for poetry 1.2.1
Source (PyPI): 1 packages found for keyrings-google-artifactregistry-auth >=1.1.1,<2.0.0
   1: fact: keyrings-google-artifactregistry-auth (1.1.1) depends on pluggy (*)
[...]

PR #6724, opened 2022-10-06
Pip no input

dimbleby:
Sometimes pip will ask for user input.

But in the context of poetry: the user never gets to see that request, and so it just looks as though poetry has hung.

I think this may have been what was going on in #6668 

Anyway passing `--no-input` seems sensible and safe.

### DIFF
--- a/src/poetry/utils/pip.py
+++ b/src/poetry/utils/pip.py
@@ -29,6 +29,7 @@ def pip_install(
         "install",
         "--disable-pip-version-check",
         "--isolated",
+        "--no-input",
         "--prefix",
         str(environment.path),
     ]


neersighted:
@radoering I'm personally not sure on backporting this to 1.2 -- I understand pragmatism in getting it out sooner than later in case we can't cut 1.3 this weekend, but also I think we should do `--isolated` at the same time if we do so.

neersighted (approved):

radoering:
I have no strong feelings about it. I just thought, it's little risk but we can also remove the labels.

Just to be clear, IMO this weekend is no option for 1.3. We should cut 1.3 only some time after 1.2.2. Thus, I'd hope for 1.2.2 this weekend and 1.3 some weeks later.

neersighted:
Yeah, I realize now that due to the new lock file format we'll need to try to announce it loudly and give it some soak time. Your call, I just think we should also do #6531 at the same time if we do.

radoering:
I'm not sure if #6531 is more risk since it's more like a change in behavior. So maybe postpone it to 1.3. Thus, I would pass the ball back to you. 😉

PR #6641, opened 2022-09-27
Add note about publishable repositories to publish

StarrFox:
This pull request adds a note under publish about publishing repositories.

relevant discord message thread: https://discord.com/channels/487711540787675139/487711540787675143/1024346792793485412

light mode:
image
dark mode:
image

PR #6557, opened 2022-09-18
docs: rephrase plugin configuration in docs

Secrus:
# Pull Request Check List

Resolves: #6121

- [ ] Added **tests** for changed code. (**does not include code changes**)
- [x] Updated **documentation** for changed code.

PR #6731, opened 2022-10-07
release: bump version to 1.2.2

radoering:
### Added

- Add forward compatibility for lock file format 2.0, which will be used by Poetry 1.3 ([#6608](https://github.com/python-poetry/poetry/pull/6608)).

### Changed

- Allow `poetry lock` to re-generate the lock file when invalid or incompatible ([#6753](https://github.com/python-poetry/poetry/pull/6753)).

### Fixed

- Fix an issue where the deprecated JSON API was used to query PyPI for available versions of a package ([#6081](https://github.com/python-poetry/poetry/pull/6081)).
- Fix an issue where versions were escaped wrongly when building the wheel name ([#6476](https://github.com/python-poetry/poetry/pull/6476)).
- Fix an issue where the installation of dependencies failed if pip is a dependency and is updated in parallel to other dependencies ([#6582](https://github.com/python-poetry/poetry/pull/6582)).
- Fix an issue where the names of extras were not normalized according to PEP 685 ([#6541](https://github.com/python-poetry/poetry/pull/6541)).
- Fix an issue where sdist names were not normalized ([#6621](https://github.com/python-poetry/poetry/pull/6621)).
- Fix an issue where invalid constraints, which are ignored, were only reported in a debug message instead of a warning ([#6730](https://github.com/python-poetry/poetry/pull/6730)).
- Fix an issue where `poetry shell` was broken in git bash on Windows ([#6560](https://github.com/python-poetry/poetry/pull/6560)).

### Docs

- Rework the README and contribution docs ([#6552](https://github.com/python-poetry/poetry/pull/6552)).
- Fix for inconsistent docs for multiple-constraint dependencies ([#6604](https://github.com/python-poetry/poetry/pull/6604)).
- Rephrase plugin configuration ([#6557](https://github.com/python-poetry/poetry/pull/6557)).
- Add a note about publishable repositories to `publish` ([#6641](https://github.com/python-poetry/poetry/pull/6641)).
- Fix the path for lazy-loaded bash completion ([#6656](https://github.com/python-poetry/poetry/pull/6656)).
- Fix a reference to the invalid option `--require` ([#6672](https://github.com/python-poetry/poetry/pull/6672)).
- Add a PowerShell one-liner to the basic usage section ([#6683](https://github.com/python-poetry/poetry/pull/6683)).
- Fix the minimum poetry version in the example for plugins ([#6739](https://github.com/python-poetry/poetry/pull/6739)).

### poetry-core ([`1.3.2`](https://github.com/python-poetry/poetry-core/releases/tag/1.3.2))

- Add `3.11` to the list of available Python versions ([#477](https://github.com/python-poetry/poetry-core/pull/477)).
- Fix an issue where caret constraints of pre-releases with a major version of 0 resulted in an empty version range ([#475](https://github.com/python-poetry/poetry-core/pull/475)).

### poetry-plugin-export ([`^1.1.2`](https://github.com/python-poetry/poetry-plugin-export/releases/tag/1.1.2))

- Add support for exporting `constraints.txt` files ([#128](https://github.com/python-poetry/poetry-plugin-export/pull/128)).
- Fix an issue where a relative path passed via `-o` was not interpreted relative to the current working directory ([#130](https://github.com/python-poetry/poetry-plugin-export/pull/130)).


neersighted:
Blocking this a change in the export plugin -- we discussed the fact that the author of the `constraints.txt` PR actually won't be able to use the feature as they use editable deps and those currently trigger a `RuntimeError`. We've decided it would be preferable to write a warning to stderr and continue (both for this user and in the general case) -- this has to happen before the release to avoid it being a breaking change.

neersighted:
Per the call today, we also established at at least while we're including poetry-plugin-export as a built-in component, we should roll up the effective changes between the version bumps.

We should also roll up user-facing changes for core based on where the pin moves.

I think these should each be their own sub-section -- once this PR is updated with that changelog information we can roll it up into the release announcement as well: https://github.com/python-poetry/website/pull/93

radoering:
I only added changelog entries for poetry-core and poetry-export-plugin PRs that don't have a poetry companion PR with a similar changelog entry.

neersighted:
Re-worked the format a little, PTAL.

Website PR is prepped and just needs the final changelog: https://github.com/python-poetry/website/pull/93

CHANGELOG.md
```diff
@@ -1,8 +1,51 @@
 # Change Log
 
+## [1.2.2] - 2022-10-09
```
mkniewallner, line 3:
```suggestion
## [1.2.2] - 2022-10-10
```

mkniewallner (approved):

neersighted (approved):

PR #1001, opened 2019-03-31
fix: RunCommand calling scripts with incorrect executable path

chdsbd:
Fixes GH-965

Calling `sys.argv` should run the same program as the currently running program. To make calling Poetry scripts through RunCommand match this behavior, we must set `sys.argv[0]` to be the full path of the executable.

This change makes the behavior of calling a script through `poetry run` the same as calling a script directly from the .venv/bin.

# Pull Request Check List

This is just a reminder about the most common mistakes. Please make sure that you tick all *appropriate* boxes.  But please read our [contribution guide](https://poetry.eustace.io/docs/contributing/) at least once, it will save you unnecessary review cycles!

- [x] Added **tests** for changed code.
- [x] Updated **documentation** for changed code.

**Note**: If your Pull Request introduces a new feature or changes the current behavior, it should be based
on the `develop` branch. If it's a bug fix or only a documentation update, it should be based on the `master` branch.

If you have *any* questions to *any* of the points above, just **submit and ask**!  This checklist is here to *help* you, not to deter you from contributing!

Issue #965, opened 2019-03-16
RunCommand incorrectly sets sys.argv for Poetry scripts


chdsbd:
- [x] I am on the [latest](https://github.com/sdispater/poetry/releases/latest) Poetry version.
- [x] I have searched the [issues](https://github.com/sdispater/poetry/issues) of this repo and believe that this is not a duplicate.
- [x] If an exception occurs when executing a command, I executed it again in debug mode (`-vvv` option).

- **OS version and name**: macOS 10.13
- **Poetry version**: 0.12.11
- **Link of a [Gist](https://gist.github.com/) with the contents of your pyproject.toml file**: https://github.com/chdsbd/poetry-repro/blob/master/pyproject.toml

## Issue
sys.argv varies when running package through `poetry run` vs `poetry shell`

- `poetry run cli`, `sys.argv == ['cli']`
- `poetry shell && cli`, `sys.argv == ['/full/path/to/package/.venv/bin/cli']`

__EDIT:__ I've looked into this a little bit more and it seems like [this RunCommand code](https://github.com/sdispater/poetry/blob/a17a99462e111d3cab9361db32245743b825f27f/poetry/console/commands/run.py#L22-L41) is the root of the issue. We are assigning incorrect values to `sys.argv` for the executable. I believe that `sys.argv` should be args that, if called, will run the same program as the running program.

`poetry shell` is going to always work correctly because it doesn't touch Poetry. The virtualenv is added to the path so we're calling directly to the executable in the virtualenv bin.

For the fix, I think we want to lookup the full path to the Poetry script in virtualenv bin and use that executable path as the first arg in `sys.argv`. I think that should fix this issue and make the behavior of calling scripts through `poetry run` equal to `poetry shell`.

dimaqq:
I think I've hit this issue with `tornado` dev mode hot reload (i.e. tornado restarts when app source code change is detected).
The trouble there is that tornado tries to restart the app as `/path-to-system/python this-app`

chdsbd:
Yes. I hit this with the Django hot reload server.

chdsbd:
@sdispater I've looked into this more and updated the issue with more information. I believe there is bug with `RunCommand` and poetry scripts.

chdsbd:
I've opened PR #1001 to fix this.

chdsbd:
This is still an issue. I have a PR open (#1001) that will fix this.

PR #6737, opened 2022-10-07
Fix incorrect sys.argv[0] path when calling project scripts

wagnerluis1982:
# Pull Request Check List

Resolves: #965 
Closes: #1001 

This PR supersedes PR #1001. I am reusing the code with the PR author's blessing, @chdsbd (see https://github.com/python-poetry/poetry/pull/1001#issuecomment-1267808776). The work I did here is the following:

1. I rebased the original code against the latest master and resolved the conflicts.
2. I handled the request for change by @finswimmer (see https://github.com/python-poetry/poetry/pull/1001#pullrequestreview-359861877)
3. I rewrote the test to use a real pyproject.toml and added a second test case

---

Description copied from PR #1001:
> Calling sys.argv should run the same program as the currently running program. To make calling Poetry scripts through RunCommand match this behavior, we must set sys.argv[0] to be the full path of the executable.
>
> This change makes the behavior of calling a script through poetry run the same as calling a script directly from the .venv/bin.


wagnerluis1982:
Test on Windows is tricky :sweat_smile:

src/poetry/console/commands/run.py
```diff
@@ -44,7 +44,13 @@ def _module(self) -> Module:
 
         return module
 
-    def run_script(self, script: str | dict[str, str], args: str) -> int:
+    def run_script(self, script: str | dict[str, str], args: list[str]) -> int:
+        # Calling `sys.argv` should run the same program as the currently
```
neersighted, line 48: Let's redo this comment as a docstring, e.g. 

```python
"""
This function runs the script, setting argv[0] because...
"""
```

src/poetry/console/commands/run.py
```diff
@@ -44,7 +44,13 @@ def _module(self) -> Module:
...
+        # running program. To make calling Poetry scripts through RunCommand
+        # match this behavior, we must set `sys.argv[0]` to be the full path of
+        # the executable.
+        args = [self.env._bin(args[0]), *args[1:]]
```
neersighted, line 52: Not a huge fan of using the `_bin` function here -- could we maybe refactor the `Env` family to have `get_bin` or something else appropriate?

neersighted (dismissed):

src/poetry/console/commands/run.py
wagnerluis1982: Good point. I sticked to `get_bin_path`, it is (almost) self explainable.

src/poetry/console/commands/run.py
wagnerluis1982: I redid as the following:

```python
"""Runs an entry point script defined in the section ``[tool.poetry.scripts]``.

When a script exists in the venv bin folder, i.e. after ``poetry install``,
then ``sys.argv[0]`` must be set to the full path of the executable, so
``poetry run foo`` and ``poetry shell``, ``foo`` have the same ``sys.argv[0]``
that points to the full path.

Otherwise (when an entry point script does not exist), ``sys.argv[0]`` is the
script name only, i.e. ``poetry run foo`` has ``sys.argv == ['foo']``.
"""
```

tests/console/commands/test_run.py
```diff
@@ -146,3 +148,38 @@ def test_run_script_exit_code(
...
+    """
+    cli_script = "foo"
+
+    if installed_script:
+        if WINDOWS:
+            cli_script += ".exe"
+        cli_script = tmp_venv._bin_dir / cli_script
+        cli_script.touch()
+        if WINDOWS:
+            cli_script = str(cli_script).replace("\\", "\\\\")
```
dimaqq, line 174: r-strings could be helpful here.

tests/console/commands/test_run.py
```diff
@@ -146,3 +148,38 @@ def test_run_script_exit_code(
...
+    poetry = Factory().create_poetry(
+        Path(__file__).parent.parent.parent / "fixtures" / "simple_project"
+    )
+    env_execute_mock = mocker.patch("poetry.utils.env.Env.execute", return_value=0)
+    tester = command_tester_factory("run", poetry, environment=tmp_venv)
+
+    tester.execute("foo status")
+
+    expected_sys_argv = f"sys.argv = ['{cli_script}', 'status']"
+    assert expected_sys_argv in env_execute_mock.call_args_list[0][0][2]
```
dimaqq, line 185: This is a little magic for my liking:
* is more than one call expected?
* what are the other arguments?

How about `assert mock.call_args.args == [ANY, ANY, expected]` instead ?

dimaqq: Another option is `m.assert_called_once_with([ANY, ANY, exp])`

dimaqq: Oh, I missed the substring match... 
I guess that invalidates my suggestions but also shows how current version confused other developers.

dimaqq (dismissed):
I'm very glad that someone took this up ❤️

tests/console/commands/test_run.py
wagnerluis1982: I feel the same. For that, I used once the following nice test lib https://callee.readthedocs.io

With it, I could rewrite the assert as
```
m.assert_called_once_with(ANY, ANY, Contains(expected_sys_argv))
```

Anyway, I will think about on how to improve this piece.

wagnerluis1982: @dimaqq I improved the test a bit

- Now I am using `mock.call_args[0][2]` to get the argument. It's better than `mock.call_args_list[0][0][2]`. Unfortunately I cannot use `mock.call_args.args[2]`, because `call_args.args` is available from Python 3.8, but poetry still requires Python 3.7 compatibility.
- I extracted the expression to a variable and added a comment explaining what is this.

wagnerluis1982: I rewrote the test using `assert_called_once_with`, it becomes better, I admit :smile:

dimaqq (dismissed):

radoering:
Seems like #6971 is a duplicate of this one. In #6971, I thought it would make sense to use `script_dirs`. I assume, most of the time `bin_path` and `script_dirs[0]` is the same. I'm wondering which approach is the correct one if they are not?

wagnerluis1982:
@radoering, #6971 assumes the script is always installed to the bin/Scripts folder, which will cause an issue if you add an entry to `[tool.poetry.scripts]` after you run `poetry install`.

This assumption disrupts the existing behavior.

wagnerluis1982:
Apart from that, this PR has tests :smirk_cat:

wagnerluis1982:
@radoering I didn't proper answered you regarding the `script_dirs`:

On this PR, I am not concerned about locations. I rely locations to `Env._bin` method (that I turned public and renamed to `Env.get_bin_path`).

But checking the underlying implementation of the turned-public method, it makes use of `Env._bin_dir` not `Env.script_dirs`.

radoering:
I still think `script_dirs` is more correct. I suppose it may be possible that `_bin_dir` is read-only and scripts are installed in another directory. By the way, `Env._bin`/`Env.get_bin_path` does not work for Windows because scripts are not installed as ".exe" but ".cmd". I adapted the test so that a real script is installed by poetry to avoid making false assumptions.

> #6971 assumes the script is always installed to the bin/Scripts folder, which will cause an issue if you add an entry to `[tool.poetry.scripts]` after you run `poetry install`.

This is very valuable input. With that information I'm quite confident we can just iterate over `script_dirs` and if we don't find the script it should be fine to keep argv0 as is.

Please take a look at my changes and let me know if you have any concerns.

wagnerluis1982:
Hi @radoering, thank you for the update.

Regarding `script_dirs` I am still unsure about it. From what I can see in the code `script_dirs[0]` is always the target folder for installed scripts.

It seems to me you're trying to solve the generic case, i.e. run any command, but `run_script` method is used only for scripts in the `[tool.poetry.scripts]`.

My reasoning is the following: in the `RunCommand.handle` there is a condition (lines 27, 28) which uses `run_script` when the script exists in the `[tool.poetry.scripts]`, otherwise calls `self.env.execute` without any preprocessing.

https://github.com/python-poetry/poetry/blob/ed5441f5659bd66b56ce475504be98031907621c/src/poetry/console/commands/run.py#L22-L34

radoering:
The root project is built/installed here: https://github.com/python-poetry/poetry/blob/ed5441f5659bd66b56ce475504be98031907621c/src/poetry/console/commands/install.py#L182

And here you can see that `script_dirs` is searched for a writable directory to install the scripts: https://github.com/python-poetry/poetry/blob/ed5441f5659bd66b56ce475504be98031907621c/src/poetry/masonry/builders/editable.py#L160-L172

wagnerluis1982:
Makes sense. I'm convinced!

PS: I don't quite understand how `script_path.exists()` is working on Windows, given you're running without extension in the tests 😱

radoering:
It's a bit weird. 😄 There are two files on Windows. One without extension and one with ".cmd". And even when executing without an extension the ".cmd" seems to be used. Strictly speaking, we had to check for the existence of the ".cmd", but since we are creating both files, it's probably not that relevant. I assume we could also add the ".cmd" for argv0. Might be cleaner. 🤷

https://github.com/python-poetry/poetry/blob/ed5441f5659bd66b56ce475504be98031907621c/src/poetry/masonry/builders/editable.py#L184-L211

tests/console/commands/test_run.py
```diff
@@ -146,3 +146,40 @@ def test_run_script_exit_code(
...
+
+    install_tester = command_tester_factory(
+        "install",
+        poetry=poetry_with_scripts,
+        environment=tmp_venv,
+    )
+    assert install_tester.execute() == 0
+    if not installed_script:
+        for path in tmp_venv.script_dirs[0].glob("check-argv0*"):
+            path.unlink()
```
wagnerluis1982, line 179: Here the test would be more clear if I don't install at all when `installed_script == True`.

I actually tried to refactor this way, but for some reason, the test for "not installed" fails...

Here what I tried
```py
    if installed_script:
        install_tester = command_tester_factory(
            "install",
            poetry=poetry_with_scripts,
            environment=tmp_venv,
        )
        assert install_tester.execute() == 0
```

tests/console/commands/test_run.py
```diff
@@ -146,3 +146,40 @@ def test_run_script_exit_code(
...
+
+    tester = command_tester_factory(
+        "run", poetry=poetry_with_scripts, environment=tmp_venv
+    )
+    argv1 = "absolute" if installed_script else "relative"
+    assert tester.execute(f"check-argv0 {argv1}") == 0
```
wagnerluis1982, line 185: I'd prefer, instead of passing the expectation as argument, to the script return an exit code for each type, e.g. 1 is absolute, 2 is relative.

wagnerluis1982:
I pushed a commit to include `.cmd` on Windows (tests are passing).

Also, I added a few minor contextual comments to the code.

tests/console/commands/test_run.py
wagnerluis1982: Never mind this one. For some reason, I believed that I could use `poetry run` before issue a `poetry install`.

tests/console/commands/test_run.py
wagnerluis1982: Disconsider this comment, I just realized that applying this will make the test confused.

src/poetry/console/commands/run.py
```diff
@@ -44,7 +45,25 @@ def _module(self) -> Module:
...
+        Otherwise (when an entry point script does not exist), ``sys.argv[0]`` is the
+        script name only, i.e. ``poetry run foo`` has ``sys.argv == ['foo']``.
+        """
+        for script_dir in self.env.script_dirs:
+            script_path = script_dir / args[0]
+            if WINDOWS:
+                script_path = script_path.with_suffix(".cmd")
+            if script_path.exists():
+                args[0] = str(script_path)
+                break
```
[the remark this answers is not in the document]
radoering: Never thought about that. It's safe as long nobody else relies on args being unchanged. I think that's the case, here. However, looking into the cleo code it seems it just returns it's internal list (not a copy) and changing it doesn't seem correct.

Please go ahead with the copy approach. That way we stay safe even if cleo decides to return a tuple instead of a list.

tests/console/commands/test_run.py
radoering: Yep, not installing at all was my first approach. However, it seems the root package has to be installed. Only the script itself needn't to be installed.

src/poetry/console/commands/run.py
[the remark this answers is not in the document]
wagnerluis1982: Great, just done!

radoering (approved):

src/poetry/console/commands/run.py
```diff
@@ -44,7 +45,25 @@ def _module(self) -> Module:
...
+    def run_script(self, script: str | dict[str, str], args: list[str]) -> int:
+        """Runs an entry point script defined in the section ``[tool.poetry.scripts]``.
+
+        When a script exists in the venv bin folder, i.e. after ``poetry install``,
+        then ``sys.argv[0]`` must be set to the full path of the executable, so
+        ``poetry run foo`` and ``poetry shell``, ``foo`` have the same ``sys.argv[0]``
+        that points to the full path.
+
+        Otherwise (when an entry point script does not exist), ``sys.argv[0]`` is the
+        script name only, i.e. ``poetry run foo`` has ``sys.argv == ['foo']``.
```
davidism, line 57: This is incorrect. `sys.argv[0]` should be the absolute path to the imported file, regardless of installation status.

wagnerluis1982:
I understand you want to use `sys.argv[0]` to make a self call to the
script.

Now I wonder how useful would be a `sys.argv[0]` pointing to a imaginary
non-existing location (which would be the case when the script is not
installed).

davidism:
It should probably be `["-c"]` instead of the name of the Poetry script if it's not installed. That mirrors Python's `sys.argv` behavior, no tool inspecting `sys.argv` would know what to do with the name on its own.

`sys.orig_argv` can be used (and shouldn't be modified by Poetry) to get the original command line that would produce the same execution.

wagnerluis1982:
@davidism I think is sort of pointless to discuss this here, in a closed PR. Better to focus the discussion on the open issue #7528

Issue #1196, opened 2019-06-26
`poetry lock` should overwrite broken `poetry.lock` file, not error out.


dimaqq:
```
… ~/my_project> poetry shell

[EmptyKeyError]
Empty key at line 36 col 0

shell

```

The actual reason is that I have a merge conflict in `pyproject.toml`.
I wish `poetry` told me that! 🙏 

Also, if the merge conflict is in `poetry.lock` the best action to resolve it is seemingly to run `poetry lock` (assuming project is close to `HEAD` and not some old branch merged with another old branch).

(poetry versions tested: `1.0.0a2` `1.0.0a4` `0.12.16`)

dimaqq:
```
… ~/myproj> git rebase origin/HEAD
...
M	poetry.lock
Falling back to patching base and 3-way merge...
Auto-merging poetry.lock
CONFLICT (content): Merge conflict in poetry.lock
error: Failed to merge in the changes.
...
… ~/myproj> poetry lock

[EmptyKeyError]
Empty key at line 1896 col 0
```

Still a problem with `poetry==1.0.0b2`

eikevons:
Would a `--force` flag help here, that ignores any existing content in `poetry.lock`. That way we avoid silently overwriting inconsistent lock states.

sinoroc:
@dimaqq  Does the issue still exist in newer poetry versions (>=1.1)?

eikevons:
I can confirm that a broken `poetry.lock` still makes `poetry lock` to throw an error.

In a fresh Python 3.8 with poetry 1.1.4 I get (I broke `poetry.lock` by emptying all content)
```
 poetry lock

  NonExistentKey

  'Key "metadata" does not exist.'

  at ~/.poetry/lib/poetry/_vendor/py3.8/tomlkit/container.py:576 in __getitem__
      572│             key = Key(key)
      573│ 
      574│         idx = self._map.get(key, None)
      575│         if idx is None:
    → 576│             raise NonExistentKey(key)
      577│ 
      578│         if isinstance(idx, tuple):
      579│             # The item we are getting is an out of order table
      580│             # so we need a proxy to retrieve the proper objects
```

Steps to reproduce with `docker`:
- create `test.sh`:
  ```sh
  #!/bin/sh

  set -xe

  apt-get update -q
  apt-get install --yes -q curl
  curl -sSL https://raw.githubusercontent.com/python-poetry/poetry/master/get-poetry.py | python
[...]

PR #6753, opened 2022-10-09
feat(locker): `poetry lock` works if an invalid/incompatible lock file exists

radoering:
After having created a lock file 2.0, running `poetry lock` with poetry 1.2.1 results in the following output:

```
The lock file is not compatible with the current version of Poetry.
Upgrade Poetry to be able to read the lock file or, alternatively, regenerate the lock file with the `poetry lock` command.
```

Ironically, the error message proposes to run `poetry lock` which results in this error message.

Further, it doesn't make sense that `poetry lock` fails because it creates a new lock file from scratch (in contrast to `poetry lock --no-update`).

Update: Running `poetry lock` is now also possible if there is a broken lock file.

Resolves: #1196

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -12,6 +12,7 @@ repos:
       - id: check-case-conflict
       - id: check-json
       - id: check-toml
+        exclude: tests/fixtures/invalid_lock/poetry\.lock
       - id: check-yaml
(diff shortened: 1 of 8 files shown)


neersighted:
Does this resolve #1196, or can it be extended to do so?

radoering:
It does. I added a test.

PR #6760, opened 2022-10-10
chore: add missing CHANGELOG entry for 1.1.15

neersighted:


radoering:
Just noticed there is a "#" missing before the "Fixed" of 1.1.14. Do we want to fix it on the fly?

neersighted:
Let's go ahead and do so.

radoering (approved):

Issue #3060, opened 2020-10-02
📚 Contribute to documentation and frequently asked questions


abn:
Hi there! This issue is a permanent issue dedicated as a placeholder to for documentation contributions. If you have questions talk to the community on our [discord server](https://discordapp.com/invite/awxPgve).

If you have a question about a particular issue, or pull rqeuest just tag @python-poetry/triage in a comment.

## Reference
- [Contribution Guide](https://github.com/python-poetry/poetry/blob/master/CONTRIBUTING.md#contributing-to-documentation)
- [Open Documentation Issues](https://github.com/python-poetry/poetry/labels/Documentation)
- [Documentation Source](https://github.com/python-poetry/poetry/tree/master/docs)
- [Opportunities to add an FAQ](https://github.com/python-poetry/poetry/issues?q=is%3Aissue+label%3Acandidate%2Ffaq+)
- [FAQ Source](https://github.com/python-poetry/poetry/blob/master/docs/docs/faq.md)

PR #6762, opened 2022-10-10
fixed a grammatic error in the documentation

sadatyussuf:
# Pull Request Check List

Resolves: #3060 

### DIFF
--- a/docs/contributing.md
+++ b/docs/contributing.md
@@ -1,5 +1,5 @@
 ---
-title: "Contributing to Poetry"
+title: 'Contributing to Poetry'
 draft: false
 type: docs
 layout: single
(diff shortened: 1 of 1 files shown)


neersighted:
Hi @sadatyussuf, thanks for your interest in contributing! Unfortunatly, this PR is not mergable:

* Most of it is changing syntax for seemingly no reason -- without tooling to enforce one syntax or another, a PR that just changes the docs with zero substance is unlikely to be accepted.
* The only grammar change is incorrect.
* The PR claims it `Resolves:` the permanent issue when it should be `Relates-to:`.

If you're interested in making these syntax changes (and dropping the bad grammar change), feel free to use pre-commit to enforce some normalization of our markdown files. But until then, I think we'll decline any syntax-only changes.

PR #6770, opened 2022-10-11
Add changelog URL to package metadata

intgr:
When Renovate bot creates an automated pull request to update Poetry, it can automatically include the link to changelog so it's easier to find out what changed.

The URL name has to be one of: 'changelog', 'change log', 'changes', 'release notes', 'news', "what's new" (case insensitive)

https://github.com/renovatebot/renovate/blob/c3a87b687ed44a1a100e0e4f91b81e22b7c32482/lib/modules/datasource/pypi/index.ts#L130-L140

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -38,7 +38,7 @@ classifiers = [
 ]
 
 [tool.poetry.urls]
-"Release notes" = "https://python-poetry.org/history/"
+changelog = "https://python-poetry.org/history/"
 
 [tool.poetry.build]
 generate-setup-file = false
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -37,6 +37,9 @@ classifiers = [
     "Topic :: Software Development :: Libraries :: Python Modules"
 ]
 
+[tool.poetry.urls]
+"Release notes" = "https://python-poetry.org/history/"
+
 [tool.poetry.build]
 generate-setup-file = false
 
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -38,7 +38,7 @@ classifiers = [
 ]
 
 [tool.poetry.urls]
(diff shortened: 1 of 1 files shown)


pyproject.toml
```diff
@@ -37,6 +37,9 @@ classifiers = [
     "Topic :: Software Development :: Libraries :: Python Modules"
 ]
 
+[tool.poetry.urls]
+"Release notes" = "https://python-poetry.org/history/"
```
neersighted, line 41:
```suggestion
changelog = "https://python-poetry.org/history/"
```

neersighted (changes requested):
I'm not sure this is the canonical URL we want to use -- we've discussed trying to have better automation around changelogs as currently we have to do a lot of juggling of text, as well as exposing all projects on the website. But I suppose it should still be available as a URL and is probably better than linking to a Markdown file or Github releases.

pyproject.toml
```diff
@@ -37,6 +37,9 @@ classifiers = [
     "Topic :: Software Development :: Libraries :: Python Modules"
 ]
 
+[tool.poetry.urls]
+changelog = "https://python-poetry.org/history/"
```
intgr, line 41: @neersighted I don't mind lowercase personally, but I think it's common to use capitalized words in the urls metadata, because the keys are displayed on PyPI website. I don't know if the website also does capitalization on its own.

E.g. https://pypi.org/project/mypy/

image

intgr: The example in packaging docmentation also uses capitalized keys: https://packaging.python.org/en/latest/guides/distributing-packages-using-setuptools/#project-urls

neersighted: Capital C works for me, then.

neersighted (approved):

Issue #6775, opened 2022-10-11
Improve description of the cache clear subcommand


ottobackwards:
Poetry: 1.2.2
Python 3.9
MacOS 11.7

- **Poetry version**: 
- **Python version**: 
- **OS version and name**: 
- **pyproject.toml**: 

- [x] I am on the [latest](https://github.com/python-poetry/poetry/releases/latest) stable Poetry version, installed using a recommended method.
- [x] I have searched the [issues](https://github.com/python-poetry/poetry/issues) of this repo and believe that this is not a duplicate.
- [x] I have consulted the [FAQ](https://python-poetry.org/docs/faq/) and [blog](https://python-poetry.org/blog/) for any relevant entries or release notes.
- [ ] If an exception occurs when executing a command, I executed it again in debug mode (`-vvv` option) and have included the output below.

## Issue
When you do `poetry` and it lists the commands, the entry for cache is
```bash

 cache
  cache clear        Clears Poetry's cache.
```

This implies that there is a `Poetry cache` when in fact there are multiple poetry caches, and a name must be specified.
The wording for this description should be something like
`Clear a Poetry cache by name ( poetry cache list )`

PR #6776, opened 2022-10-11
improve cache clear description

ottobackwards:
# Pull Request Check List

Resolves: #6775

- [-] Added **tests** for changed code.
- [-] Updated **documentation** for changed code.

### DIFF
--- a/src/poetry/console/commands/cache/clear.py
+++ b/src/poetry/console/commands/cache/clear.py
@@ -12,7 +12,7 @@ from poetry.console.commands.command import Command
 
 class CacheClearCommand(Command):
     name = "cache clear"
-    description = "Clears Poetry's cache."
+    description = "Clears a Poetry cache by name (see cache list)."
 
     arguments = [argument("cache", description="The name of the cache to clear.")]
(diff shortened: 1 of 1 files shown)


src/poetry/console/commands/cache/clear.py
```diff
@@ -12,7 +12,7 @@
 
 class CacheClearCommand(Command):
     name = "cache clear"
-    description = "Clears Poetry's cache."
+    description = "Clears a Poetry cache by name (see cache list)."
```
neersighted, line 15:
```suggestion
    description = "Clears a Poetry cache by name."
```

I'm not a fan of this parenthetical, since we don't do anything like it anywhere else. Since this PR is related to the presentation during `poetry list`, it feels unnecessary as `cache list` will be right next to it, anyway.

neersighted (changes requested):
LGTM with one tweak.

src/poetry/console/commands/cache/clear.py
ottobackwards: ok

neersighted (approved):

PR #6781, opened 2022-10-12
ci: lint audit and improvements

neersighted:
A series of audits of our linting settings, plus incremental improvements, including some new plugins. It's easy to drop commits if anything is controversial, but I think most of these should be very palatable.

- chore(flake8): audit settings and add documentation lints
- chore(flake8): update plugin versions
- chore(pyupgrade): modify excludes for easy grepping
- chore(pycln): store config in pyproject.toml
- chore(isort): simplify config and centralize in pyproject.toml
- chore(pre-commit): add additional (cheap) hooks
- Revert "fix: use importlib_metadata on python <3.10"
- fix(pyproject): make 3.7-only backports more explicit
- chore(flake8): add logging-format and string-format plugins
- chore(flake8): add deprecated plugin
- chore(flake8): add builtins and warnings plugins
- chore(flake8): add debugger and print plugins
- refactor(installed_repository): drop Python 2.7 compatibility
- chore(flake8): drop broken-line and annotations plugins


.flake8
```diff
@@ -1,36 +1,44 @@
...
 min_python_version = 3.7.0
-max-line-length = 88
-ban-relative-imports = true
-# flake8-use-fstring: https://github.com/MichaelKim0407/flake8-use-fstring#--percent-greedy-and---format-greedy
-format-greedy = 1
-inline-quotes = double
-enable-extensions = TC, TC1
-type-checking-exempt-modules = typing, typing-extensions
-eradicate-whitelist-extend = ^-.*;
+enable-select =
```
finswimmer, line 3: I cannot find this option :thinking: Do you mean [extend-select](https://flake8.pycqa.org/en/latest/user/options.html#cmdoption-flake8-extend-select)? Or `enable-extensions` as described in https://pypi.org/project/flake8-type-checking/?

.pre-commit-config.yaml
```diff
@@ -5,76 +5,91 @@ repos:
   - repo: https://github.com/pre-commit/pre-commit-hooks
     rev: v4.3.0
     hooks:
-      - id: trailing-whitespace
+      - id: fix-byte-order-marker
+      - id: mixed-line-ending
+        args: [--fix=lf]
```
finswimmer, line 10: AFAIK this is problematic when working on windows and linux. By default git changes the line endings on checkout on windows to whatever windows is used to (I always forget how it is called) and will change it to what's the default on unix on commit. pre-commit runs before that on commit, so it will always fix line ending on windows.

https://github.com/pre-commit/pre-commit-hooks#mixed-line-ending

src/poetry/puzzle/provider.py
```diff
@@ -664,7 +664,9 @@ def complete_package(
                 dependencies.append(deps[0])
                 continue
 
-            self.debug(f"<debug>Duplicate dependencies for {dep_name}</debug>")
+            self.debug(
+                f"<debug>Duplicate dependencies for {dep_name}</debug>"  # noqa: G004
```
finswimmer, line 668: Does it make sense to use `flake8-logging-format` if we suppress most of the errors?

I'm also a bit ambivalent about disallowing f-strings in logging messages. We neither have shown that this is a performance issue nor we have a logging handler that like to process the arguments.

finswimmer (changes requested):

.flake8
neersighted: I do mean `extend-select` -- looks like this silently breaks typechecking messages :laughing:

.pre-commit-config.yaml
neersighted: Ah, fair, I guess we should use auto because of git's `core.autocrlf` (I just disable it, personally).

src/poetry/puzzle/provider.py
finswimmer: Oh, I didn't recognize it is `self.debug` and not `logger.debug` :see_no_evil:

pyproject.toml
```diff
@@ -49,14 +49,14 @@ python = "^3.7"
...
-"backports.cached-property" = { version = "^1.0.2", python = "<3.8" }
+"backports.cached-property" = { version = "^1.0.2", python = "3.7" }
 cachecontrol = { version = "^0.12.9", extras = ["filecache"] }
 cleo = "^1.0.0a5"
 crashtest = "^0.3.0"
 dulwich = "^0.20.46"
 filelock = "^3.8.0"
 html5lib = "^1.0"
-importlib-metadata = { version = "^4.4", python = "<3.10" }
+importlib-metadata = { version = "^4.4", python = "3.7" }
```
radoering, line 59: Sure about that?

<html><body>


importlib_metadata | stdlib
-- | --
4.8 | 3.11
4.4 | 3.10
1.4 | 3.8


</body>
</html>
(from https://pypi.org/project/importlib-metadata/)

If we really want to do that, we'll have to change https://github.com/python-poetry/poetry/blob/0b71e4381445bf6b6c239bf0d88c8f5dfb4a32bd/src/poetry/utils/_compat.py#L10-L11

pyproject.toml
```diff
@@ -49,14 +49,14 @@ python = "^3.7"
 
 poetry-core = "^1.3.2"
 poetry-plugin-export = "^1.1.2"
-"backports.cached-property" = { version = "^1.0.2", python = "<3.8" }
+"backports.cached-property" = { version = "^1.0.2", python = "3.7" }
```
radoering, line 52: I like <3.8 even if we don't support 3.6. I suppose that's a matter of taste. 🤷

neersighted: The intention is to make it grepable -- I agree that `<3.8` is generally preferable, but in this specific case of covering one version I think it makes it easier to remove later.

.pre-commit-config.yaml
```diff
@@ -5,76 +5,90 @@ repos:
...
 
   - repo: https://github.com/asottile/yesqa
     rev: v1.4.0
     hooks:
       - id: yesqa
         additional_dependencies: &flake8_deps
-          - flake8-annotations==2.9.0
-          - flake8-broken-line==0.5.0
-          - flake8-bugbear==22.7.1
+          - flake8-annotations==2.9.1
```
finswimmer, line 53: We have enable `strict` mode in `mypy`. This enables  `--disallow-untyped-defs`, `--disallow-incomplete-defs`, `--check-untyped-defs`. I think this makes `flake8-annotations obsolete. (https://mypy.readthedocs.io/en/stable/command_line.html#cmdoption-mypy-disallow-untyped-defs)

.pre-commit-config.yaml
```diff
@@ -5,76 +5,90 @@ repos:
...
+          - flake8-broken-line==0.6.0
```
finswimmer, line 54: I'm not 100% sure, but I think that if the code is formatted by `black`, those things that `flake8-broken-line` checks will never appear.

neersighted: It looks like you are correct: https://black.readthedocs.io/en/stable/the_black_code_style/current_style.html, ctrl-f `backslashes`.

.pre-commit-config.yaml
```diff
@@ -5,76 +5,90 @@ repos:
...
+      - id: name-tests-test
+        args: [--pytest-test-first]
+        exclude: (compat|helpers|types|utils)\.py|fixtures/[^/]+
+      - id: no-commit-to-branch
 
   - repo: https://github.com/pre-commit/pygrep-hooks
     rev: v1.9.0
     hooks:
+      - id: python-check-blanket-noqa
+      - id: python-check-blanket-type-ignore
```
finswimmer, line 36: In the `mypy` config we have enabled `ignore-without-code`. So this check here would redundant.

neersighted: It is strictly speaking so, but I think this is nice as a way to give early feedback (many contributors forget to run mypy locally).

.pre-commit-config.yaml
neersighted: I think you're right, and it cleans things up nicely.

.flake8
```diff
@@ -1,36 +1,39 @@
...
     PIE781,
-    # PIE798 no-unnecessary-class: Consider using a module for namespacing instead
-    PIE798,
 per-file-ignores =
-    # TC002: Move third-party import '...' into a type-checking block
-    __init__.py:TC002,
-    # ANN201: Missing return type annotation for public function
-    tests/test_*:ANN201
-    tests/**/test_*:ANN201
+    # PIE798: no-unnecessary-class: Consider using a module for namespacing instead
```
neersighted, line 16: This is a target for future removal (Git implementations should probably implement a protocol or inherit from an ABC, and should be swapped by class instead of hardcoded).

.pre-commit-config.yaml
neersighted: I think the argument for removing flake8-annotations is that we have to suppress/special case a decent number of issues it reports; whereas this is free and doesn't require any special casing.

Secrus:
I am going to close this. It's based on `flake8`, which we dropped and has a lot of conflicts. A lot of the `flake8` plugins used here can be found in `ruff`.

Issue #6722, opened 2022-10-06
Environment variable POETRY_EXPERIMENTAL_SYSTEM_GIT_CLIENT not recognised during poetry install


lhutzelmann:
- **Poetry version**: 1.2.1
- **Python version**: 3.9.10
- **OS version and name**: Ubuntu 22.04
- **pyproject.toml**: See below

- [x] I am on the [latest](https://github.com/python-poetry/poetry/releases/latest) stable Poetry version, installed using a recommended method.
- [x] I have searched the [issues](https://github.com/python-poetry/poetry/issues) of this repo and believe that this is not a duplicate.
- [x] I have consulted the [FAQ](https://python-poetry.org/docs/faq/) and [blog](https://python-poetry.org/blog/) for any relevant entries or release notes.
- [x] If an exception occurs when executing a command, I executed it again in debug mode (`-vvv` option) and have included the output below.

## Issue
My pyproject.toml contains a dependency to a company internal git repository such as following example:
```
[tool.poetry.dependencies]
mylib = {git = "https://SOME_GIT_CLONE_URL"}
```
This is only relevant to ensure that the configuration is asked whether the system git client should be used or not.
When setting the environment variable `POETRY_EXPERIMENTAL_SYSTEM_GIT_CLIENT` to `true` and then running a `poetry install` the environment variable is ignored and the internal git client library is used.
When setting the configuration `experimental.system-git-client` via `poetry config`to true and then running the same installation command the system git client is used now.
It seems that in the method `is_using_legacy_client()` of module `poetry.vcs.git.backend` the following line does not catch the value from the environment variable (due to the two separate .get() calls that do not fit to the single environment variable name).
```
Config.create().get("experimental", {}).get("system-git-client", False)
```
At other places the following line is used to get the "new-installer" from "experimental" values instead (which works AFAIK):
```
poetry.config.get("experimental.new-installer", False)
```
Maybe this can be fixed in a similar way?
Best regards
Lars

ramvikrams:
can i take up this issue and do the changes

ramvikrams:
this **Config.create().get("experimental", {}).get("system-git-client", False)** has to be changed to **poetry.config.get("experimental.new-installer", False)** ,am I correct?

neersighted:
> can i take up this issue and do the changes
Nobody is going to tell you you can't -- the worst that could happen is someone else beats you to the punch after you've started work.

[...]

PR #6783, opened 2022-10-12
fix(git): `experimental.system-git-client` can't be set using environment variable

miles170:
# Pull Request Check List

Resolves: #6722

### DIFF
--- a/tests/integration/test_utils_vcs_git.py
+++ b/tests/integration/test_utils_vcs_git.py
@@ -111,6 +111,13 @@ def remote_default_branch(remote_default_ref: bytes) -> str:
     return remote_default_ref.decode("utf-8").replace("refs/heads/", "")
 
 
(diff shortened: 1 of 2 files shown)


tests/integration/test_utils_vcs_git.py
```diff
@@ -110,6 +110,10 @@ def remote_default_ref(_remote_refs: FetchPackResult) -> bytes:
 def remote_default_branch(remote_default_ref: bytes) -> str:
     return remote_default_ref.decode("utf-8").replace("refs/heads/", "")
 
+def test_use_system_git_client_from_environment_variables():
```
neersighted, line 113: Let's add a docstring linking back to the original issue and labeling this a regression test (we test the primary functionality of env var -> config else where and it will help explain why this orphan test exists).

neersighted (changes requested):

neersighted (approved):

Issue #6784, opened 2022-10-12
Encoding failure is not caught and propagated


kristang:
- **Poetry version**:  `1.2.1` and `1.3.0.dev01`
- **Python version**: `3.9.13`
- **OS version and name**: `Windows 11 Enterprise (build 22000.856)`
- **pyproject.toml**: 

```toml
[tool.poetry]
name = "test-poetry-master"
version = "0.1.0"
description = ""
authors = ["kristang"]
readme = "README.md"

[[tool.poetry.source]]
name = "private-pypi"
url = "https://private/pypi/private-pypi/simple/"
secondary = true

[tool.poetry.dependencies]
python = "^3.9"

[tool.poetry.dev-dependencies]
pytest = "^7.1.1"
mypy = "^0.982"

[build-system]
requires = ["poetry-core"]
build-backend = "poetry.core.masonry.api"
```

- [x] I am on the [latest](https://github.com/python-poetry/poetry/releases/latest) stable Poetry version, installed using a recommended method.
- [x] I have searched the [issues](https://github.com/python-poetry/poetry/issues) of this repo and believe that this is not a duplicate.
- [x] I have consulted the [FAQ](https://python-poetry.org/docs/faq/) and [blog](https://python-poetry.org/blog/) for any relevant entries or release notes.
- [x] If an exception occurs when executing a command, I executed it again in debug mode (`-vvv` option) and have included the output below.
#### Additionally
- [x] Discussed the issue  with @neersighted as well as @finswimmer and @radoering on the Poetry Discord.
## Issue

When adding a secondary, private source and attempting to authenticate via basic-http, `poetry` silently ignores a `UnicodeEncodingError` originating from a failed latin1 encoding of a `str`. 

Instead, poetry moves on to the public [pypi.org](https://pypi.org/) feed and informs the user that the searched package is not available - even if it is available on [pypi.org](https://pypi.org/)

The error stems from loading in a user/password from Windows Certificate Manager which appears to be corrupted and cannot be encoded with 'latin1'

The source of the encoding error is `poetry` invoking `_basic_auth_str` from `auth.py` in the `requests` package:
https://github.com/psf/requests/blob/7104ad4b135daab0ed19d8e41bd469874702342b/requests/auth.py#L56-L60

The originating call in `poetry` is `request` in `utils.authenticator`. 

[...]

PR #6786, opened 2022-10-12
Remove try-except ValueError to allow UnicodeEncodeError to propagate

kristang:
UnicodeEncodeError is currently silently caught and not passed through to end-user when encountered.

UnicodeEncodeError is a subclass of UnicodeError and thus a subclass of ValueError which is currently caught in try-except ValueError in `mixology/version_solver.py`:

https://github.com/python-poetry/poetry/blob/0ffe91c0c8e2edde987cf2213d402614a9c2a6a6/src/poetry/mixology/version_solver.py#L398-L401

# Pull Request Check List

Resolves: #6784

### DIFF
--- a/src/poetry/mixology/version_solver.py
+++ b/src/poetry/mixology/version_solver.py
@@ -395,10 +395,8 @@ class VersionSolver:
                 if locked:
                     return is_specific_marker, Preference.LOCKED, 1
 
-            try:
-                num_packages = len(self._dependency_cache.search_for(dependency))
-            except ValueError:
-                num_packages = 0
+            num_packages = len(self._dependency_cache.search_for(dependency))
+
             if num_packages < 2:
                 preference = Preference.NO_CHOICE
             elif use_latest:


dimbleby:
I mean 2018... whatever!

This change was at my suggestion so I'm just doubling down rather than adding new opinion, but I'm good with this.  If and when reports roll in of things that this used to catch then they can be fixed: but that the poetry test suite does not hit any reassures me that this at worst very far from the main line.

Actually once you believe in this MR you might as well go further, the same `ValueError` is caught a few lines later and if you believe that's unhittable too then lots of code can disappear.

radoering:
If it's from 2018 and it's not clearly evident that a test was added these days, I'm not very confident that there is something in the test suite. On the one hand, removing it is the best way to get feedback. On the other hand, it's quite disruptive. Since I'm normally a cautious person I'd probably go with

> It seems like we should refactor the exceptions thrown by poetry-core in general to avoid a generic `ValueError` (we should probably subclass it)

Nevertheless, if other maintainers want to tread the other path, that's fine with me too.

dimbleby:
I don't believe this ValueError handling is connected to version or constraint parsing, if it ever was

- the constraint on the dependency that we have in hand has long since been parsed
- versions on the packages that we are finding are handled during the search eg [here](https://github.com/python-poetry/poetry/blob/c2b1fb3d17f9da1bfc8dc483fbb8d916917c03f0/src/poetry/repositories/pypi_repository.py#L87) and [here](https://github.com/python-poetry/poetry/blob/1215ba8f339a03701d4f87105c3bb4c54f34b7b0/src/poetry/repositories/link_sources/base.py#L87)
  - and this is how it has to be, else a single badly versioned package would prevent all versions of that package from being installed

By all means tighten up the exceptions in poetry-core and elsewhere, but I do not think this fix needs to be blocked on that

kristang:
Would the pseudo-fix I suggested in the issue I created be more palatable? I'm not familiar enough with your development practices, but I understand the hesitation with just up and removing a try-except. 

```python 
if credential.username is not None or credential.password is not None:
    try:
        request = requests.auth.HTTPBasicAuth(
            credential.username or "", credential.password or ""
        )(request)
    except UnicodeEncodeError as e:
        raise PoetryException(e)
```

dimbleby:
IMO that's a step in the wrong direction rather than a step in the right direction.  The current try-except is either too broad or - as I contend - altogether unwanted, and working around that by cluttering up the rest of the code with this sort of thing is undesirable

kristang:
Completely fair.

neersighted:
I'm inclined to go with @dimbleby's version as I don't see any evidence that the try/catch in this code is doing anything useful. I do think that we should open an issue to subclass exceptions in poetry-core and make them more specific.

kristang:
Should I just close this and let it be replaced by #6790?

radoering:
Superseded by #6790

PR #6803, opened 2022-10-14
Rename repository modules

b-kamphorst:
Relates-to: https://github.com/python-poetry/poetry/issues/3155

Follow-up of https://github.com/python-poetry/poetry/pull/6669 regarding the [proposed renaming](https://github.com/python-poetry/poetry/pull/6669#issuecomment-1264767316) of modules in `poetry/repositories`.

A subsequent PR for poetry-plugin-export is in preparation: https://github.com/python-poetry/poetry-plugin-export/pull/151.


b-kamphorst:
@neersighted let's make this fly. For now, I've added `DeprecationWarnings` that indicate support until poetry v1.4.0 (as to not be breaking for the next poetry minor release, 1.3.0). I noted that the deprecation warning for the `poetry.core.semver.version:Version` import did not state a version, but it might be good to do so.

You also [mentioned](https://github.com/python-poetry/poetry/pull/6669#discussion_r990364440) the desire for a `RepositoryName` type that represents a normalized name. My (ugly) attempt at doing so broke the plugin and would be less ugly if rooted in `poetry-core`. Not sure if we want to pick this up, and if so, where.

Looking forward to your thoughts.

radoering (changes requested):
I propose to divide the rename and the deprecation into two commits as done in python-poetry/poetry-core#482. That way, it's easier to follow changes in git history and doing rebases of PRs changing code in one of the renamed modules.

radoering:
And it simplifies the review. 😉

b-kamphorst:
> I propose to divide the rename and the deprecation into two commits as done in [python-poetry/poetry-core#482](https://github.com/python-poetry/poetry-core/pull/482). That way, it's easier to follow changes in git history and doing rebases of PRs changing code in one of the renamed modules.

Done!

src/poetry/repositories/pool.py
```diff
@@ -0,0 +1,27 @@
...
+        repositories: list[Repository] | None = None,
+        ignore_repository_names: bool = False,
+    ) -> None:
+        warnings.warn(
+            "Object Pool from poetry.repositories.pool is renamed and scheduled for"
+            " removal in poetry release 1.4.0. Please migrate to RepositoryPool from"
+            " poetry.repositories.repository_pool.",
+            DeprecationWarning,
+            stacklevel=2,
+        )
```
radoering, line 26: IMO the deprecation warning should be at module level because not only the class but the module is deprecated.

radoering (changes requested):
Is there a reason for not having a deprecation warning for `cached`? (And for the exclamation mark in the commit message. `refactor!: rename cached to cached_repository`?)

As for the wording of the deprecation warning (if we want to add a "scheduled for removal notice" and if it will be 1.4), I'll leave that decision to @neersighted. Afaik we don't have a schedule yet for how long we keep deprecated code.

b-kamphorst:
> Is there a reason for not having a deprecation warning for `cached`? (And for the exclamation mark in the commit message. `refactor!: rename cached to cached_repository`?)

Not really -- I hoped that this was used only internally and therefore we may not have to remain compatible. But that was just guessing so I'll change it (probably tomorrow). I guess in general with Python it is not quite clear what the API is -- harder to know what you can change without issues to others.

> As for the wording of the deprecation warning (if we want to add a "scheduled for removal notice" and if it will be 1.4), I'll leave that decision to @neersighted. Afaik we don't have a schedule yet for how long we keep deprecated code.

If this is part of the public API, I'd expect deprecation in a major release. But I'm not sure how the poetry team applies versioning. We can leave it open but I think it is nice to at some point (timely) inform users when it is completely removed. Looking forward to your suggestions.

src/poetry/repositories/pool.py
b-kamphorst: I would agree, but since `poetry.repositories.pool` is also imported in `poetry.repositories.__init__`, this will raise the warning at any `poetry.repositories(.*)` import. As a user I would only expect to see a `DeprecationWarning` if I import the thing that is deprecated, so therefore I opted for this suboptimal approach (also: no warning for typing with deprected `Pool`). 

The rationale for importing `Pool` in `__init__.py` is that `poetry-plugin-import` depends on it (imports `Pool` from poetry.repositories`).

b-kamphorst:
@radoering, I've added the deprecation warning for `poetry/repositories/cached`. I think all that is left to finalize is the content of the warning message depending on your and @neersighted's thoughts?

tests/repositories/test_repository_pool.py
```diff
@@ -4,33 +4,41 @@
...
 
 
+def test_pool_raises_package_not_found_when_no_package_is_found() -> None:
+    pool = RepositoryPool()
+    pool.add_repository(Repository("repo"))
+
+    with pytest.raises(PackageNotFound):
+        pool.package("foo", Version.parse("1.0.0"))
+
+
```
radoering, line 30: That's a duplicated test, isn't it?

radoering (changes requested):
We decided that the deprecation warnings are fine as they are. I just noticed a test that seems to be duplicated (see separate comment). Apart from that I think this PR is ready to merge.

tests/repositories/test_repository_pool.py
b-kamphorst: It is.

radoering (approved):

PR #6805, opened 2022-10-15
match requests pool_maxsize to num workers

adriangb:
This avoids trashing connections just to immediately re-create them when `num-worker` > 10. This should provide a pretty solid speedup on beefy machines.

I'm not attaching any tests because this would be hard to test and if it doesn't crash with an unknown keyword argument or something it means it's probably going to be doing what we expect.


neersighted (approved):

adriangb:
Thank you for pushing this across the finish line @neersighted !

Issue #6807, opened 2022-10-15
Poetry v1.2.2 does not work with `packaging==20.4` due to a bad `cherry-pick`


aloisklink:
- **Poetry version**: v1.2.2
- **Python version**: 3.8.10
- **OS version and name**: Ubuntu 20.04
- **pyproject.toml**: Any `pyproject.toml` (you can create a blank one with `poetry init` and it has the same error, so does `poetry run pytest`)

- [ ] I am on the [latest](https://github.com/python-poetry/poetry/releases/latest) stable Poetry version, installed using a recommended method.
  - I'm on v1.2.2 (current latest), but I installed via `pip3 install poetry`.
- [x] I have searched the [issues](https://github.com/python-poetry/poetry/issues) of this repo and believe that this is not a duplicate.
- [x] I have consulted the [FAQ](https://python-poetry.org/docs/faq/) and [blog](https://python-poetry.org/blog/) for any relevant entries or release notes.
- [x] If an exception occurs when executing a command, I executed it again in debug mode (`-vvv` option) and have included the output below.

## Issue

It looks like the `1.2` branch has a bad cherry-pick, https://github.com/python-poetry/poetry/commit/04c5cbf043610dd5d89b7d980bf8a137948c4fa5

This line was not in the original commit and is not in the `master` branch: https://github.com/python-poetry/poetry/blob/04c5cbf043610dd5d89b7d980bf8a137948c4fa5/src/poetry/installation/installer.py#L6

On v20.4 of `packaging`, this import is only available `if typing.TYPE_CHECKING`, which is correct on the Poetry `master` branch:

https://github.com/python-poetry/poetry/blob/16046d9ac9b72a49e1bc4618fb686695cc64821c/src/poetry/installation/installer.py#L21-L25

This causes the following error when running `poetry install`:

```
  ImportError

  cannot import name 'NormalizedName' from 'packaging.utils' (/home/user/.local/lib/python3.8/site-packages/packaging/utils.py)

  at ~/.local/lib/python3.8/site-packages/poetry/installation/installer.py:6 in <module>
        2│ 
        3│ from typing import TYPE_CHECKING
        4│ 
        5│ from cleo.io.null_io import NullIO
    →   6│ from packaging.utils import NormalizedName
        7│ from packaging.utils import canonicalize_name
        8│ 
        9│ from poetry.installation.executor import Executor
       10│ from poetry.installation.operations import Install
```

### Reproducing error in Poetry repo

[...]

PR #6808, opened 2022-10-15
[1.2] fix: fix support for packaging v20.4

aloisklink:
**This bug only affects the `1.2` branch (caused by a bad cherry-pick from `master`)**

Poetry v1.2.2 does not work with packaging v20.4 (which is an allowed version of packaging in the pyproject.toml file).

This is because the `from packaging.utils import NormalizedName` line before [packaging v20.5][1] only works if `typing.TYPE_CHECKING` is True.

This is not an issue on the `master` branch, since there this import is already hidden behind an `if TYPE_CHECKING:` guard. However, the v1.2 branch never cherry-picked the appropriate commit adding this change.

See the current `master` branch, which works fine:
https://github.com/python-poetry/poetry/blob/16046d9ac9b72a49e1bc4618fb686695cc64821c/src/poetry/installation/installer.py#L21-L26

[1]: https://github.com/pypa/packaging/commit/eca30e0435f99084682d529d997f5c5fe1cd43fb

Fixes: https://github.com/python-poetry/poetry/commit/04c5cbf043610dd5d89b7d980bf8a137948c4fa5 (this was the broken cherry-pick)

# Pull Request Check List

Resolves: #6807

  - The only way to test this is to install `packaging` v20.4, and I can't see an easy way to do this.

### DIFF
--- a/src/poetry/installation/installer.py
+++ b/src/poetry/installation/installer.py
@@ -3,7 +3,6 @@ from __future__ import annotations
 from typing import TYPE_CHECKING
 
 from cleo.io.null_io import NullIO
-from packaging.utils import NormalizedName
 from packaging.utils import canonicalize_name
 
 from poetry.installation.executor import Executor
@@ -24,6 +23,7 @@ if TYPE_CHECKING:
     from collections.abc import Sequence
 
     from cleo.io.io import IO
+    from packaging.utils import NormalizedName
     from poetry.core.packages.project_package import ProjectPackage
 
     from poetry.config.config import Config


neersighted (approved):

radoering (approved):
There might not be another 1.2 release. Nevertheless I'm going to merge this so it's fixed at least in the 1.2 branch.

rossmotley:
To help others - to work around this you can update the version of packaging with `pip3 install --upgrade packaging`

PR #6824, opened 2022-10-17
Relay return code as exit code when running script

### LINKED ISSUE #6801: Exit code relay fails when run command is a script
- **Poetry version**: `1.2.2`
- **Python version**: `3.10.6`
- **OS version and name**: macOS 12.5.1
- **pyproject.toml**: Relevant details below

```
[tool.poetry.scripts]
toolname = "toolname.__main__:main"
```

## Issue

Hi there,

When using `poetry run` to run a script defined in the [`scripts`](https://python-poetry.org/docs/pyproject/#scripts) section of `pyproject.toml` the exit code of the script is not relayed as the final exit code of `poetry` itself. I originally noticed https://github.com/python-poetry/poetry/issues/5773 and https://github.com/python-poetry/poetry/issues/2369 and was wondering why it wasn't working for me.

Consider the following cases:

```
$ poetry run python -m toolname --version
0.5.0
$ echo $?
1
```

```
$ poetry run toolname --version
0.5.0
$ echo $?
0
```

It looks like https://github.com/python-poetry/poetry/pull/4456 did not fix https://github.com/python-poetry/poetry/issues/2369 because these are different execution paths. Looking at [`commands/run.py`](https://github.com/python-poetry/poetry/blob/16046d9ac9b72a49e1bc4618fb686695cc64821c/src/poetry/console/commands/run.py#L61), running a script (as defined in `pyproject.toml`) is a different execution path than running non-scripts. As mentioned in https://github.com/python-poetry/poetry/issues/2369, I believe the fix would be:

`f"import_module('{module}').{callable_}()"` -> `f"sys.exit(import_module('{module}').{callable_}())"`

I made this change locally and it fixed this issue for me.

Alternatively, I believe another fix would be removing the `RunCommand.run_script` special-case code. Since `poetry` knows where the script is (`$VIRTUALENV/bin/`), it could include this directory in the `$PATH` when executing the command. `poetry` would then execute the script generated by [`builders/editable.py`](https://github.com/python-poetry/poetry/blob/master/src/poetry/masonry/builders/editable.py#L31).

There may be reasons these two code paths are distinct that I don't understand, but if they can be unified to a single path that would ease maintenance burden and help alleviate situations like this where one path is updated but the other is forgotten.

rc-mattschwager:
# Pull Request Check List

Resolves: #6801 

I didn't see any relevant tests for the `run_script` functionality, so I'm not sure how best to test it. I did confirm the fix by running the relevant functionality locally and noting the correct exit code:

```
$ poetry run toolname --version
0.5.0
$ echo $?
1
```

```
$ poetry run python -m toolname --version
0.5.0
$ echo $?
1
```

Let me know the best way to add an automated test 👍

### DIFF
--- a/src/poetry/console/commands/run.py
+++ b/src/poetry/console/commands/run.py
@@ -58,7 +58,7 @@ class RunCommand(EnvCommand):
             "import sys; "
             "from importlib import import_module; "
             f"sys.argv = {args!r}; {src_in_sys_path}"
-            f"import_module('{module}').{callable_}()"
+            f"sys.exit(import_module('{module}').{callable_}())"
         ]
 
         return self.env.execute(*cmd)
--- a/tests/console/commands/test_run.py
+++ b/tests/console/commands/test_run.py
@@ -1,5 +1,7 @@
 from __future__ import annotations
 
+import subprocess
+
 from typing import TYPE_CHECKING
 
 import pytest
@@ -127,7 +129,12 @@ def test_run_script_exit_code(
     poetry_with_scripts: Poetry,
     command_tester_factory: CommandTesterFactory,
     tmp_venv: VirtualEnv,
+    mocker: MockerFixture,
 ) -> None:
+    mocker.patch(
+        "os.execvpe",
+        lambda file, args, env: subprocess.call([file] + args[1:], env=env),
+    )
     install_tester = command_tester_factory(
         "install",
         poetry=poetry_with_scripts,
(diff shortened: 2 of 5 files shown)


finswimmer (dismissed):
Thanks a lot for contribution :+1: 

The change looks reasonable to me. Could you please try to add a test for it?

fin swimmer

radoering (approved):

rc-mattschwager:
Thanks for writing a test, that looks like a doozy!

Issue #5989, opened 2022-07-11
Installing poetry using python2.7 in docker image


elados93:
- [X] I am on the [latest](https://github.com/python-poetry/poetry/releases/latest) Poetry version.
- [X] I have searched the [issues](https://github.com/python-poetry/poetry/issues) of this repo and believe that this is not a duplicate.
- [ ] If an exception occurs when executing a command, I executed it again in debug mode (`-vvv` option).

- **OS version and name**: MacOS Monterey 12.4
- **Poetry version**: 1.1.14
- **Link of a [Gist](https://gist.github.com/) with the contents of your pyproject.toml file**: 

## Issue
I'm trying to write a `Dockerfile` for my setup, installing poetry using python2.7.18. I'm installing poetry the recommended way which according to docs should support python2.7 & 3.5. Lately, I encountered `UnicodeDecodeError` & noticed https://github.com/python-poetry/poetry/issues/3672 so I added `PYTHONIOENCODING` although it didn't seems to help.

### Dockerfile
```
# latest python2 version
FROM python:2.7.18-slim

# pin poetry version
ENV POETRY_VERSION="1.1.14"

# https://github.com/python-poetry/poetry/issues/3672
ENV PYTHONIOENCODING=utf8

# install poetry using official website recomended way
RUN apt-get update; apt-get install curl -y
RUN curl -sSL https://raw.githubusercontent.com/python-poetry/poetry/master/get-poetry.py | python -
```

### Build Error
```
 => ERROR [3/8] RUN curl -sSL https://raw.githubusercontent.com/python-poetry/poetry/master/get-poetry.py | python -               1.1s
------                                                                                                                                  
 > [3/8] RUN curl -sSL https://raw.githubusercontent.com/python-poetry/poetry/master/get-poetry.py | python -:                          
#6 1.031 Retrieving Poetry metadata                                                                                                     
#6 1.032 Traceback (most recent call last):                                                                                             
#6 1.032   File "<stdin>", line 1117, in <module>                                                                                       
#6 1.032   File "<stdin>", line 1113, in main
#6 1.032   File "<stdin>", line 351, in run
#6 1.032   File "<stdin>", line 414, in get_version
#6 1.032 UnicodeDecodeError: 'ascii' codec can't decode byte 0xc3 in position 20: ordinal not in range(128)
```
 what am I missing here?

Thanks.

dimbleby:
[...]

Issue #6377, opened 2022-09-02
Removal of get-poetry.py


neersighted:
This issue tracks the removal of `get-poetry.py` from this repository. After discussion triggered by the release of 1.2.0, and an attempt to bring awareness to the deprecation of this script, it has been decided that we will remove `get-poetry.py` from this repo on or after **January 1, 2023**.

A PR that changes `get-poetry.py` to have the following flow will be merged:

- When run without an explicit version, 1.2.x will be selected (as it is the latest) and the installer will fail as it cannot install 1.2.0a1 or newer.
- Any explicit version can be selected, and the installer will succeed for versions it understands.
- If `GET_POETRY_IGNORE_DEPRECATION=1` in set the environment, the latest installable version will be used even when no explicit version is selected.
- In all cases a deprecation message will be printed, and warnings will be issued when uninstallable versions of Poetry are skipped.
- A target removal date of January 1, 2023 will be printed, with a link to this issue, and a blog post linking here as well.

The `get-poetry.py` script is considered frozen, and will not receive fixes if it is broken by external factors.

For previous discussion of this process, see #6314.

neersighted:
For those following along at home, the PR is merged and the above behaviors are accurate for `get-poetry.py`. No more changes are planned for this script before removal.

BurekPrime:
@neersighted
I would like to ask you if changing:
`curl -sSL https://raw.githubusercontent.com/python-poetry/poetry/master/get-poetry.py | python -`
to:
`curl -sSL https://raw.githubusercontent.com/python-poetry/poetry/901bdf0491005f1b3db41947d0d938da6838ecb9/get-poetry.py | python -`
will still work after 01.01.2023?

To give some context I develop a platform that is scheduled to be closed, somewhere in February 2023 and all of our Python component build pipelines depend on the old install script still working (they broke, but after your fix they are working again, so thank you for that). I'm only interested in the poetry installation working up util then (in case of some hot fix release) and don't want to introduce to much changes. Can I count that the master branch won't be rebased or something and that the script from the commit will be still working? You mentioned that script can break due to "external factors", could you please elaborate.
Sorry if I picked up a wrong channel for such question, but I'm not aware of a more appropriate place.
Thanks for the answer in advance.

neersighted:
The commit will remain in history -- Poetry as a project has a linear history and we do not rebase or drop commits from the stable/protected branches. In short: you will be able to keep using the script from that URL indefinitely.

[...]

PR #6830, opened 2022-10-18
Fix Python 2 installation script by using UTF-8 to decode in get-poetry.py

chrisburr:
# Pull Request Check List

Resolves: #5989

I know Python 2.7 is no longer supported but I think the fix is sufficiently trivial that it might be worth considering regardless?

### DIFF
--- a/get-poetry.py
+++ b/get-poetry.py
@@ -412,7 +412,7 @@ class Installer:
 
         print(colorize("info", "Retrieving Poetry metadata"))
 
-        metadata = json.loads(self._get(self.METADATA_URL).decode())
+        metadata = json.loads(self._get(self.METADATA_URL).decode("utf-8"))
 
         def _compare_versions(x, y):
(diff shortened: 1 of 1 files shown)


finswimmer:
Hey @chrisburr,

thanks a lot for your contribution :+1: We really appreciate it :pray: 

However, we will not merge your suggestion for several reasons:

- The `get-poetry.py` script is deprecated for more than one year.
- This installer cannot be used to install the current stable version of Poetry. 
- Only the outdated and no longer supported 1.1 version can be installed with it.
- Any Poetry version >=1.8 can be installed with the [recommended installer](https://python-poetry.org/docs/#installing-with-the-official-installer).
- We will remove this installer script soon from the repository #6377 

Making changes to this script might send the false signal to other users, as they might think it is under active development.

If you still need this installer for any reasons, it is [recommended](https://python-poetry.org/docs/#ci-recommendations) to store a local copy of it.

Best regards,
fin swimmer

Issue #6831, opened 2022-10-18
Poetry adds conflicting requirements when cases are mixed


adamchainz:
- **Poetry version**: 1.2.2
- **Python version**: 3.10.8
- **OS version and name**: macOS 12.6
- **pyproject.toml**: see below

- [x] I am on the [latest](https://github.com/python-poetry/poetry/releases/latest) stable Poetry version, installed using a recommended method.
- [x] I have searched the [issues](https://github.com/python-poetry/poetry/issues) of this repo and believe that this is not a duplicate.
- [x] I have consulted the [FAQ](https://python-poetry.org/docs/faq/) and [blog](https://python-poetry.org/blog/) for any relevant entries or release notes.
- [x] If an exception occurs when executing a command, I executed it again in debug mode (`-vvv` option) and have included the output below.

## Issue

Take this minimal `pyproject.toml` config for poetry:

```toml
[tool.poetry]
name = "example"
version = "0"
description = ""
authors = []

[tool.poetry.dependencies]
python = "^3.10"
Django = "^4.0.8"


[build-system]
requires = ["poetry-core"]
build-backend = "poetry.core.masonry.api"
```

Then run a command to upgrade Django:

```console
$ poetry add "Django@4.1.2"
Creating virtualenv example-i90kWyrF-py3.10 in /Users/chainz/Library/Caches/pypoetry/virtualenvs

Updating dependencies
Resolving dependencies... (0.5s)

Writing lock file

Package operations: 3 installs, 0 updates, 0 removals

  • Installing asgiref (3.5.2)
  • Installing sqlparse (0.4.3)
  • Installing django (4.1.2)
```

Poetry adds a *second* lower-cased dependency on Django to its configuration:

```toml
[tool.poetry]
name = "example"
version = "0"
description = ""
authors = []

[tool.poetry.dependencies]
python = "^3.10"
Django = "^4.0.8"
django = "4.1.2"


[build-system]
requires = ["poetry-core"]
build-backend = "poetry.core.masonry.api"
```

This difference lead to a crash when building a project Docker container:

```
 > [stage-1  7/18] RUN   poetry config virtualenvs.create false &&   poetry install:
#17 0.749 Skipping virtualenv creation, as specified in config file.
#17 1.881 Installing dependencies from lock file
#17 3.363
#17 3.363 Because example depends on both Django (^4.1.2) and django (4.0.8), version solving failed.
```

[...]

PR #6832, opened 2022-10-18
fix: ensure no duplicate package names are stored in pyproject.toml

finswimmer:
When upgrading dependencies via `poetry add <package>@latest` it may result in multiple entries in `pyproject.toml` if the name returned by the API differs from the name given in the `pyproject.toml`, e.g. different upper/lower cases.

This PR fixes this by keeping the name given in the `pyproject.toml` and updating the version constraint.

Resolves: #6831

### DIFF
--- a/src/poetry/console/commands/add.py
+++ b/src/poetry/console/commands/add.py
@@ -215,7 +215,15 @@ The add command adds required packages to your <comment>pyproject.toml</> and in
 
             constraint_name = _constraint["name"]
             assert isinstance(constraint_name, str)
-            section[constraint_name] = constraint
+
+            canonical_constraint_name = canonicalize_name(constraint_name)
+
+            for key in section:
+                if canonicalize_name(key) == canonical_constraint_name:
+                    section[key] = constraint
+                    break
+            else:
+                section[constraint_name] = constraint
 
             with contextlib.suppress(ValueError):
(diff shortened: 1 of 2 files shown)


src/poetry/console/commands/add.py
```diff
@@ -215,6 +215,11 @@ def handle(self) -> int:
 
             constraint_name = _constraint["name"]
             assert isinstance(constraint_name, str)
+
+            for key in list(section.keys()):
```
dimbleby, line 219: pedantry warning, none of this really matters

- if you're willing to assume that the `pyproject.toml` wasn't broken before we started, then 
```python
            for key in section:
                if canonicalize_name(key) == canonicalize_name(constraint_name):
                    del section[key]
                    break
```
should be sufficient, both avoiding the `list()` / `.keys()` ugliness, and avoiding doing more work than strictly necessary

- canonicalizing the name every time round the loop is ever so slightly inefficient, that could be pulled into a local variable

finswimmer: Thanks a lot for your suggestions. I don't like the `list()` as well :smiley:

dimbleby (dismissed):

Secrus (approved):

neersighted (approved):

PR #6843, opened 2022-10-20
Always re-solve path dependencies

### LINKED ISSUE #3783: poetry lock --no-update does not reflect directory packages changes
- **OS version and name**: MacOS 10.15.7 
- **Poetry version**: 1.1.5

## Issue
When using packages with the directory source type, I would expect that running `poetry lock --no-update` inside the package would update the `poetry.lock` file with changes for all directory packages being used. 

The issue can be reproduced with:
```
poetry new lib
poetry new project

# project depends on lib (directory package)
cd project && poetry add ../lib

# lib is updated with a new dependency
cd ../lib && poetry add arrow

# we try to update the project with --no-update (we do not want to run poetry update that will update too much)
cd ../project
poetry lock --no-update

# the lock file is not updated correctly
grep arrow poetry.lock  # lock file is not updated
```

To my knowledge the only workaround are:
1) `poetry update` (which naturally updates more than we want here – i.e. unrelated dependencies)
2) `poetry update <dir_package1> <dir_package2>...` (cumbersome if you have a lot of directory packages – especially painful in a big mono repository)

I believe the suggested behavior makes sense for both editable and non-editable directory packages.

adriangb:
Resolves #3783

This resolves an issue poetry lock --no-update does not recognize updates made to a path dependency because it treats it as immutable

@neersighted this works in practice on a real project, but I'm a bit at a loss as to where to add this. I think it should go in `test_solver.py`, but I think I'd have to re-create the dependencies manually by adding them to the repo, which seems tedious. Might need a bit of guidance there.


tests/console/commands/test_show.py
```diff
@@ -1041,10 +1041,7 @@ def test_show_outdated_local_dependencies(
 
     tester.execute("--outdated")
 
-    expected = """\
-cachy              0.2.0                       0.3.0
-project-with-setup 0.1.1 ../project_with_setup 0.1.2 ../project_with_setup
-"""
+    expected = """cachy 0.2.0 0.3.0 Cachy package"""
```
adriangb, line 1044: This one is interesting. I guess if path dependencies are "special" in the sense that we always update them, we don't show them? I'm going to check what happens with VCS dependencies

dimbleby:
cf #3783

adriangb:
@dimbleby thank you for linking to the related issues. Please let me know what you think this PR needs, in particular if we want any tests I would appreciate some guidance along the lines of "please write a test similar to this existing test in this test file" since I'm having a little bit of trouble figuring out what the testing strategy is for the project.

tests/console/commands/test_show.py
radoering: This looks like a regression to me. Probably, we still want to show them.

radoering:
A test may look similar to https://github.com/python-poetry/poetry/blob/5c578a2988b1725ba5042f198c9bb3efa0c77db8/tests/puzzle/test_solver.py#L3467-L3468
(except that the directory dependency is updated even if a prior version is locked).

adriangb:
@radoering test added, thanks for the pointer!

tests/console/commands/test_show.py
adriangb: Do we want all path dependencies to always show up as outdated? For me updated/not-updated for path dependencies is more "undefined" than "yes/no".

adriangb: Alright I think I was able to address this!

src/poetry/console/commands/show.py
```diff
@@ -211,7 +211,7 @@ def _display_packages_information(
         with solver.use_environment(self.env):
             ops = solver.solve().calculate_operations()
 
-        required_locked_packages = {op.package for op in ops if not op.skipped}
+        required_locked_packages = {op.package.name for op in ops if not op.skipped}
```
radoering, line 214: This change will introduce a bug that is unfortunately not discovered by our test suite. I just created #6920 to increase test coverage.

radoering (changes requested):
After taking a closer look at the `show` command, I came to the conclusion that we don't want to re-solve path dependencies in any case. We only want to re-solve when refreshing the lock file but not when solving for a specific environment for `show`.

I think we have to try another approach: We shouldn't touch `get_locked()` but add path dependencies to `use_latest` when refreshing the lock file. If you want to try that approach, this should be the place to add packages with source type "directory" to `use_latest`:
https://github.com/python-poetry/poetry/blob/4a07b5e0243bb8879dd6725cb901d9fa0f6eb182/src/poetry/installation/installer.py#L207

adriangb:
@radoering ok I think I got that approach to work, thanks for the feedback!

radoering (approved):

radoering (approved):

adriangb:
@radoering thank you!

skovhus:
Thanks for fixing this @adriangb!

Issue #668, opened 2018-11-23
Do not validate path dependencies when they are not selected for installation


spither:
- [x] I am on the [latest](https://github.com/sdispater/poetry/releases/latest) Poetry version.
- [x] I have searched the [issues](https://github.com/sdispater/poetry/issues) of this repo and believe that this is not a duplicate.
- [x] If an exception occurs when executing a command, I executed it again in debug mode (`-vvv` option).

- **OS version and name**: Ubuntu 18.10
- **Poetry version**: 0.12.10
- **Link of a [Gist](https://gist.github.com/) with the contents of your pyproject.toml file**: https://gist.github.com/spither/47f144b1f355990d562b5ba92bf15fe3

## Issue

I cannot run `poetry install --no-dev` away from my development environment because I have dev-only path based dependencies.

    > poetry install --no-dev
                                        
    [ValueError]             
    Directory ../dev-only does not exist  

The error is correct, the directory doesn't exist, but it's only listed as a dev-dependency and I'm trying to install `--no-dev`.  I was not expecting any dev-dependency checks to take place.

In my case I'm trying to install the non-dev dependencies for a Django site deployment.  The path based dev-dependencies only exist in development, not on the deployment system.

sdispater:
Poetry locks everything, including dev dependencies, before installing to ensure that there is no conflict between dev and runtime dependencies, so every dependency must be resolvable.

spither:
This happens with an existing lock file in place.

Is resolving dev dependencies (especially with an existing lock file) really needed when `--no-dev` is specified?  Is there anything I can do to avoid it?

I don't have a need or desire to copy these dependencies to my deployment system.  They wouldn't/shouldn't be used for anything there and might add confusion by being present.

rphes:
I've also run into this issue, my use case being that I want to use a local (and editable) version of a package during development and a commit from a git repository for deployment situations.

jjuris:
Hello,

I also found in this situation when I used poetry in my project. As a workaround, I simply commented out lines under development dependencies section in the project, but it is not nice solution -> un/commenting lines in configuration file is pretty annoying.

@sdispater, I would like to help you solve this problem.

The simplest solution would be "Ignore dev dependencies when --no-dev option is used."

If it is ok for you let me know and I will try to find it and fix it. If you have better idea, let me know and we can discuss it. 

Have a nice 

Day

Jozef Juris

jjuris:
Hello,

[...]

PR #6845, opened 2022-10-20
Add option to skip installing directory dependencies

### LINKED ISSUE #3671: poetry install with nested packages with "--no-root-nested" or similar option to improve docker builds
- [ x] I have searched the [issues](https://github.com/python-poetry/poetry/issues) of this repo and believe that this is not a duplicate.
- [ x] I have searched the [documentation](https://python-poetry.org/docs/) and believe that my question is not covered.

## Feature Request

Right now it is possible to install only the dependencies of a package with the `poetry install --no-root --no-dev` options. This is great to improve docker builds (cache behaviour for example).

it would be great to extend that functionality to nested packages. Something like

`poetry install --no-root --no-dev --no-root-nested`

which would tell poetry to only install the *dependencies* of the nested package, but
not the package itself. This would greatly improve the caching behaviour of docker builds as we would only need
to copy  the pyproject.toml and poetry.lock files into the docker machine which causes much better caching behaviour.

### LINKED ISSUE #6850: Monorepo support using groups and path dependencies
Following up on https://discord.com/channels/487711540787675139/1032534656777715732/1032549486230253580

Summary: I think we can make some relatively simple and universally (as in not only beneficial to monorepos) changes to the existing `path` dependencies that will make them much easier to use for monorepos. There are also some more complex feature requests to handle but I think those can be tabled for future feature work.

I had to migrate a Python monorepo at work from an ad-hoc structure to something with a more explicit separation of applications and components. I explored several options including monorepo build tools (Bazel, Pants, etc.) and Python package managers (Hatch, PDM and Poetry). I won't get into the details but the TLDR is that the build tools are too complex for a simple Python-only monorepo and the other package managers have missing features. There are several ways to structure monorepos with Poetry, but after surveying a lot of blog posts, issues and plugins I came across this one: https://github.com/martinxsliu/poetry_workspace_plugin (@martinxsliu if you're out there thank you and feel free to chime in). While I don't plan on using that Plugin necessarily (I don't need most of the functionality it provides) it had some interesting ideas. The main thing I changed was using Poetry's existing `group` feature to organize the path dependencies to projects so that even without a plugin you can get basic functionality.

## The good

So the final pattern I landed on is this: https://github.com/adriangb/python-monorepo/tree/main/poetry

TLDR is:

```toml
# /pyproject.toml
[tool.poetry.group.app.dependencies]
namespace-app = { path = "workspaces/app", develop = true}

[tool.poetry.group.lib.dependencies]
namespace-lib = { path = "workspaces/lib", develop = true}

[tool.poetry.group.lib-test.dependencies]
namespace-lib = { path = "workspaces/lib", develop = true, extras=["test"]}

# /{workspace,projects}/app/pyproject.toml
[tool.poetry.dependencies]
namespace-lib = { path = "../lib"}
```

So you can then do:

```shell
poetry install --only app
```

Which will install `app`, `lib` and any 3rd party dependencies!

### `poetry lock --no-update` doesn't work

This is alluded to in the plugin linked to above: https://github.com/martinxsliu/poetry_workspace_plugin/blob/master/poetry_workspace/plugin.py#L83-L107

Even worse: if you delete any path dependencies (delete the folder, delete it from `pyproject.toml`) `poetry lock --no-update` straight up crashes because `DirectoryDependency` is used to load the lock file (which still has a reference to the deleted dependency) and unlike VCS and other dependencies [it validates that the path exists in `__init__`](https://github.com/python-poetry/poetry-core/blob/main/src/poetry/core/packages/directory_dependency.py#L37-L38). This means your options are to (a) run `poetry lock` and update all 3rd party dependencies (b) delete it from pyproject.toml first, run `poetry --lock-update`, delete the folder and run `poetry lock --no-update` again) or (c) edit `poetry.lock` before running `poetry lock --no-update`. Especially with `--no-update` becoming the default at some point, this is problematic.

I've submitted a couple of PRs to fix this:

 - [poetry-core#507](https://github.com/python-poetry/poetry-core/pull/507) and https://github.com/python-poetry/poetry/pull/6844 move the burden of validating that a path dependency exists from the "loading the lock file" stage to "re-locking" stage which allows running `poetry lock --no-update` after deleting a path dependency
 - https://github.com/python-poetry/poetry/pull/6843 forces re-exploration of all path dependencies instead of assuming they haven't changed, which lets us pick up changes to transitive dependencies of path deps

### No caching in Docker builds because source is required to install 3rd party dependencies

When you're building a Dockerfile that uses Poetry it's best to use `--no-root` to install 3rd party dependencies first (which only requires a `pyproject.toml` and a `poetry.lock` lockfile) so that the layer which installs 3rd party dependencies gets cached across builds (it's often both the slowest and least frequently changed step). This breaks down if you have any path dependencies because you'll need to copy them over in order to install 3rd party dependencies (even using groups there's no way to skip the path dependencies but install its transitive 3rd party dependencies).

I opened https://github.com/python-poetry/poetry/pull/6845 as a way to address this. I'm not sure if `--no-path` is the best way; I'm open to other options, but the basic idea is there.

### No way to build sdists/wheels for projects with path dependencies

In my example repo you can't run `poetry build` on `cli` because it depends on `lib`. One might want to do this if e.g. you have a backend and a client you are publishing on PyPi that share a common component/library. There are several open issues around this, which I won't get into detail on. My thoughts is that it would be nice to have a feature like `namespace-lib = { path = "../lib", wheel = { version = "~={}.{}.{}"} }` where the `{}` in `version` means "insert the current version" or `namespace-lib = { path = "../lib", wheel = { target = "path/inside/wheel" } }`. I think this quite complex of a topic that needs more thought. [Even Cargo hasn't fully figured this out](https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#multiple-locations) since they don't support inheriting the version from workspace dependencies (their [`overrides` feature](https://doc.rust-lang.org/cargo/reference/overriding-dependencies.html#working-with-an-unpublished-minor-version) may be of interest but is really beyond the scope of this issue). I think this can be solved but I would table it for now since it needs more thought and will likely be more invasive to the codebase.

#1168 also is related to this particular use case and might be an alternative path forward.

#### No support for dependency groups in path dependencies

It would be interesting to have some special handling for Poetry path dependencies so that we can "propagate" groups from path dependencies to workspace/root projects (for example, choosing to install the `test` group in a path dependency). I think this can be done but I don't think it should be a blocker for other stuff, also going to table this for now.

### Boilerplate

There's a lot boilerplate involved in the whole `groups` and path deps thing, especially if you get into listing extras. I think this is where a plugin might shine: it could completely replace these sections with a `[tool.workspaces]` section or just serve to keep them in sync with the filesystem (to avoid manually writing the boilerplate).

adriangb:
This serves the same purpose as `--only-root` but for path deps (which just like the "root" will bust caches if any source file changes).

If/once https://github.com/python-poetry/poetry-core/pull/520 gets merged this will allow writing a Dockerfile along the lines of:

```
FROM python

COPY pyproject.toml poetry.lock .
RUN pip install poetry && poetry install --no-root --no-path
COPY src/ ./src
COPY folder-of-path-deps/ folder-of-path-deps/
RUN poetry install
```

I chose `--no-path` because of it's parallels to `--no-root`, but I'm open to other options.

Closes #6845
Closes #6850 
Closes #3671


docs/faq.md
```diff
@@ -152,3 +152,37 @@ This is done so to be compliant with the broader Python ecosystem.
...
+```text
+FROM python
+COPY pyproject.toml poetry.lock .
+RUN pip install poetry && poetry install --no-root --no-path
+COPY src/ ./src
+RUN poetry install --no-dev
+```
+
+The two key options we are using here are `--no-root` (skips installing the project source) and `--no-path` (skips installing any local path dependencies, you can skip this if you don't have any).
+[More information on the options available for `poetry install`]({{< relref "cli#install" >}}).
```
adriangb, line 188: I'm just kinda throwing this out here. I know we probably need more documentation on Docker best practices in general but I think that's way beyond the scope of this PR. Even this little bit is maybe pushing it, but I'm hoping that it's okay to put in here if I restrict the scope to just the impact of `--no-root` and `--no-path` (as opposed to jumping into multistage builds and whatnot).

tests/console/commands/test_install.py
```diff
@@ -195,3 +195,15 @@ def test_dry_run_populates_installer(tester: CommandTester, mocker: MockerFixtur
...
+def test_no_path_is_passed_to_installer(tester: CommandTester, mocker: MockerFixture):
+    """
+    The --no-root options is passed to the installer.
+    """
+
+    mocker.patch.object(tester.command.installer, "run", return_value=1)
+
+    tester.execute("--no-path")
+
+    assert tester.command.installer._skip_path is True
```
adriangb, line 209: I don't really like this but it's exactly what all of the other tests do so 🤷🏻‍♂️

tests/installation/test_installer.py
```diff
@@ -60,12 +60,12 @@ class Executor(BaseExecutor):
...
         super().__init__(*args, **kwargs)
 
-        self._installs: list[DependencyPackage] = []
+        self._installs: list[Package] = []
         self._updates: list[DependencyPackage] = []
         self._uninstalls: list[DependencyPackage] = []
 
     @property
-    def installations(self) -> list[DependencyPackage]:
+    def installations(self) -> list[Package]:
```
adriangb, line 68: These types seem to be wrong so I'm fixing them. I bet it's the same for `_updates` but I wasn't touching them in this PR so I decided to leave em alone.

adriangb:
I think this fixes https://github.com/python-poetry/poetry/issues/668

adriangb:
@radoering while we wait for review on https://github.com/python-poetry/poetry-core/pull/520, would you mind taking a look at this and giving some thoughts? Thanks!

radoering:
If I understand correctly, the main motivation for this feature is the Docker cache. Before taking a look into the implementation details, I'd like to here the opinions of other maintainers if this feature makes sense for them. Especially @neersighted because he has much more experience with Docker than I do.

Just one question: Does it make sense to lump directory and file dependencies together (since both are path dependencies) or should it only consider directory dependencies?

adriangb:
> Just one question: Does it make sense to lump directory and file dependencies together (since both are path dependencies) or should it only consider directory dependencies?

Great question. I’m not super familiar with workflows around file dependencies, I’ve never used them. The fundamental question as far as caches are concerned is “does this likely change every build (source code) or only rarely (dependencies)?”.
The main use case I associate file dependencies with is vendoring, in which case they’re not changing any more often than other leaf dependencies. What makes directory dependencies special is that they’re often actually your own source code.
We could always add another flag allowing users to choose depending on their use case. It’s a bit more complexity but avoids painting ourselves into a corner if we miss some use cases.

neersighted:
I'm not convinced this is useful, as in the use-case shown here (not building a package, just using Poetry to set up an environment), groups could easily be used. I don't think path dependencies should be special at all; encouraging users to think of them as 'special' takes us down a path we have strictly wanted to avoid (e.g. leaking metadata beyond the Core Metadata spec between layers of the dependency tree because Poetry is used all the way down), and reduces the interoperability of Poetry with the rest of the ecosystem.

I think that documenting the use of this pattern, but with groups, can help the people who are trying to do monorepos today, and that a workspace-like approach like suggested in #2270 is what we should explore, instead of trying to tear down the Core Metadata/PEP 517 wall between Poetry projects.

adriangb:
> in the use-case shown here (not building a package, just using Poetry to set up an environment), groups could easily be used

could you explain how? I use dependency groups in conjunction with path dependencies in a real world monorepo and it works great aside from docker caching. I don’t see how “using dependency groups” solves anything?

neersighted:
Hmm, I did not realize that this PR was targeted at the pattern in your example repo there. While I brought up the overall direction for better supporting monorepos as kind of a extension of my thinking re: this PR, I did not realize that is at the core of what you want to do.

I think that makes my objection firmer -- special support for using path dependencies in this way (with a superproject that delegates to subprojects) is not desirable because it is ultimately incomplete unless we tear down the metadata barrier, and all the designs along that line have been rejected to date. It's possible we decide that the subproject approach is not tenable and go that way after all, but I think it would have to be comprehensive and not piecemeal. It would probably be better if it was a new dependency type if that were the case, instead of overloading `path =`, but I digress.

Regarding the use case you have, I'm afraid I still don't see how what you want to do is not possible with dependency groups. Could one not do the following (with all your groups set to optional)?

``` dockerfile
FROM python:3.11

COPY pyproject.toml poetry.lock .
RUN pip install poetry && poetry install --no-root
COPY src/ ./src
RUN poetry install --only-root
COPY folder-of-path-deps/ folder-of-path-deps/
RUN poetry install --no-root --with <path-dep-groups>
```

adriangb:
> special support for using path dependencies in this way (with a superproject that delegates to subprojects) is not desirable because it is ultimately incomplete unless we tear down the metadata barrier, and all the designs along that line have been rejected to date

I haven’t heard any objection to my design. If you can think of any issues let's discuss in #6850.

> I think it would have to be comprehensive and not piecemeal

Yes this is a small isolated piece, but that is by design. Part of the attractive of this design is that it can be implemented in small incremental steps, most of which except maybe this one are already completed and we’re beneficial to Poetry regardless of the acceptance of the proposed monorepo design. So I wouldn’t really call it “piecemeal”.

> I'm afraid I still don't see how what you want to do is not possible with dependency groups. Could one not do the following (once https://github.com/python-poetry/poetry-core/pull/520 is in, with all your groups set to optional)?

I don’t think things will work properly without this PR but I’m away from a computer to confirm, I can test later or feel free to give it a crack with my example repo.

adriangb:
Ok so I had a chance to look at your example and test in my example repo.

The issue with the solution you proposed is that you never install the transitive 3rd party dependencies (which may even be all 3rd party dependencies if there is no "root" pacakge/dependencies).

I updated my example repo to use this branch (https://github.com/adriangb/python-monorepo/commit/0250ddac28d8cad20381948efad6eae3061f99be) and tested that indeed doing `poetry install --only app --no-path` gives me transitive dependencies of `app` (in particular, `app` depends on `lib` and `lib` depends on `pydantic` and indeed `pydantic` gets installed).

So at least as proposed dependency groups are not an alternative to this PR.

adriangb:
pre-commit.ci autofix

adriangb:
@radoering I renamed `path` to `directory` as per your feedback in #6850

adriangb:
I think this will also be useful in CI workflows (so not Docker but similar) since the same principle of transitive deps changing less than source code apply.

taynaud:
Maybe this change is more about top level paths dependancy than directory or files (i.e from the project.toml rather than transitive path dependancies) ? In this case, the option could be --no-root-paths or --no-top-level-paths ?

adriangb:
It does skip transitive dir dependencies

mwgamble:
> The issue with the solution you proposed is that you never install the transitive 3rd party dependencies (which may even be all 3rd party dependencies if there is no "root" pacakge/dependencies), so there's no layer that has "all of the things that are expensive to install but don't change often" which is what we want for good caching.

Can confirm, this behaviour is essential for making this work in a useful manner. I tried the package groups functionality when I was trying to make this work nicely, and it just didn't work properly at all because of the transitive dependencies issue.

src/poetry/console/commands/install.py
```diff
@@ -30,6 +30,17 @@ class InstallCommand(InstallerCommand):
         option(
             "no-root", None, "Do not install the root package (the current project)."
         ),
+        option(
+            "no-directory",
+            None,
+            (
+                "Do not install any directory path dependencies (ones using `package ="
+                ' { path = "..." }`\'; useful to install dependencies without source'
```
radoering, line 38:
```suggestion
                "Do not install any directory path dependencies; useful to install dependencies without source'
```

I'd leave that away and expect a user who doesn't know what it is to not use this option or consult the documentation.

tests/console/commands/test_install.py
```diff
@@ -347,3 +347,15 @@ def test_install_path_dependency_does_not_exist(
     else:
         with pytest.raises(ValueError, match="does not exist"):
             tester.execute(options)
+
+
+def test_no_path_is_passed_to_installer(tester: CommandTester, mocker: MockerFixture):
```
radoering, line 352:
```suggestion
def test_no_directory_is_passed_to_installer(tester: CommandTester, mocker: MockerFixture):
```

tests/console/commands/test_install.py
```diff
@@ -347,3 +347,15 @@ def test_install_path_dependency_does_not_exist(
...
+    """
+    The --no-root options is passed to the installer.
```
radoering, line 354:
```suggestion
    The --no-directory option is passed to the installer.
```

tests/console/commands/test_install.py
radoering: Can you move this somewhere near the other "pass to the installer" tests, e.g. after `test_compile_option_is_passed_to_the_installer` and add a negative test via `pytest.mark.parametrize`  (`_skip_directory` is `False` if option is not passed)

src/poetry/puzzle/transaction.py
```diff
@@ -27,7 +27,10 @@ def __init__(
         self._root_package = root_package
 
     def calculate_operations(
-        self, with_uninstalls: bool = True, synchronize: bool = False
+        self,
+        with_uninstalls: bool = True,
+        synchronize: bool = False,
+        skip_directory: bool = False,
```
radoering, line 33:
```suggestion
        *,
        skip_directory: bool = False,
```

tests/installation/test_installer.py
```diff
@@ -1283,6 +1283,49 @@ def test_run_installs_with_local_poetry_directory_transitive(
     assert installer.executor.installations_count == 6
 
 
+def test_run_installs_with_local_poetry_directory_transitive_no_path(
+    installer: Installer,
+    locker: Locker,
+    repo: Repository,
+    package: ProjectPackage,
+    tmpdir: Path,
```
radoering, line 1291:
```suggestion
```

unused

tests/installation/test_installer.py
```diff
@@ -1283,6 +1283,49 @@ def test_run_installs_with_local_poetry_directory_transitive(
     assert installer.executor.installations_count == 6
 
 
+def test_run_installs_with_local_poetry_directory_transitive_no_path(
```
radoering, line 1286: Can you integrate this with `pytest.mark.parametrize` in the previous test?

`skip_directory=False` -> `directory_installs`, `installations_count == 6`
`skip_directory=True` -> `not directory_installs`, `installations_count == ...`

radoering (changes requested):
On the whole, the feature makes sense to me for the mentioned use case and the change is quite minimal. So I think we can merge it after some minor changes.

docs/cli.md
```diff
@@ -225,6 +225,12 @@ If you want to skip this installation, use the `--no-root` option.
 poetry install --no-root
 ```
 
+Similar to `--no-root` you can use `--no-directory` to skip path dependencies:
```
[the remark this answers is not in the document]
adriangb: 469d246b

tests/console/commands/test_install.py
adriangb: donzo

tests/installation/test_installer.py
adriangb: did it

adriangb:
@rodering, I addressed all of the feedback and got the branch in sync with the default branch. Apologies if the commits got messy, I'm working from plane wifi.

radoering (approved):

adriangb:
Thank you for pushing this across the finish line Randy!

Issue #6788, opened 2022-10-12
AssertionError from Poetry install


jace-ys:
- **Poetry version**: 1.2.1
- **Python version**: 3.8.10
- **OS version and name**: macOS 12.6
- **pyproject.toml**: 
```
[tool.poetry]
name = ""
version = "0.1.0"
description = ""
authors = []
readme = "README.md"
packages = []

[tool.poetry.dependencies]
python = "^3.8"
click = "8.1.3"
hvac = "0.9.2"
uvloop = "0.14.0"
protobuf = "3.20.2"
prometheus-async = "18.3.0"
prometheus-client = "0.2.0"
sanic-validation = "0.4.4"
urllib3 = "1.26.9"
sanic = "20.12.6"
monzo-cli = { path = "wheels/Monzo_CLI-1.1.0-py3-none-any.whl" }
detect-secrets = "0.12.7"

[tool.poetry.group.test.dependencies]
multidict = "5.0.0"
pytest = "5.3.1"
pytest-sanic = "1.1.2"
mock = "3.0.5"

[build-system]
requires = ["poetry-core"]
build-backend = "poetry.core.masonry.api"

```

- [x] I am on the [latest](https://github.com/python-poetry/poetry/releases/latest) stable Poetry version, installed using a recommended method.
- [x] I have searched the [issues](https://github.com/python-poetry/poetry/issues) of this repo and believe that this is not a duplicate.
- [x] I have consulted the [FAQ](https://python-poetry.org/docs/faq/) and [blog](https://python-poetry.org/blog/) for any relevant entries or release notes.
- [x] If an exception occurs when executing a command, I executed it again in debug mode (`-vvv` option) and have included the output below.

## Issue

Hey 👋🏻 

I have a Docker image which runs `poetry install` and this works completely fine on my local machine (M1 MacBook) but for some reason fails occasionally on my production build pipeline (running on Linux x86_64). I can't seem to figure out why it's happening, so was wondering if anyone here might have an idea?

Dockerfile:
```
FROM ubuntu:20.04

# Poetry configuration
ENV POETRY_VERSION=1.2.1 \
    POETRY_HOME="/opt/poetry" \
    POETRY_VIRTUALENVS_IN_PROJECT=true \
    POETRY_INSTALLER_MAX_WORKERS=10

ENV PATH="$PATH:$POETRY_HOME/bin"

WORKDIR /app

RUN apt-get update
RUN apt-get install -y --no-install-recommends \
    build-essential \
[...]

PR #6855, opened 2022-10-21
Handle distribution.files being None

jace-ys:
# Pull Request Check List

Resolves: #6788 

We discovered that in some unknown situations, `distribution.files` can be `None`, causing the assertion to fail. To handle `distribution.files` being `None` more gracefully, we treat it as if it were an empty list.


jace-ys:
Closing in favour of #6877 (just a change in branch of my fork)

PR #6863, opened 2022-10-21
Allow Git to request user credentials if specifically desired

andrewschueler:
This PR is related to [#2825](https://github.com/python-poetry/poetry/issues/2825).

This PR addresses the need to let the user decide if they need to submit user credentials when cloning Git repositories. Most users don't want to force a user to enter credentials. This PR does not change that functionality. This PR simply enables the user to decide for themselves if they want `poetry` to request credentials or not when cloning repositories with Git.

I would suspect that better documentation should be included regarding this issue. I am happy to add documentation, I am not sure where to add said documentation.

Example python program that demonstrates the desired functionality:
```
import os
env = os.environ.copy()
try:
    result = env["GIT_TERMINAL_PROMPT"]
    print(f"GIT_TERMINAL_PROMPT state:\t{result}")
except:
    # Preserve default behavior
    result = "0"
    env["GIT_TERMINAL_PROMPT"] = result
    print(f"Default GIT_TERMINAL_PROMPT set to \"0\"")
```
Define the environment variable `GIT_TERMINAL_PROMPT` to a value of `0` or `1` to see the behavior change w.r.t. not defining the variable at all.


andrewschueler:
I fixed the issue pointed out by `pre-commit.ci` is fixed but the test still fails. The exception captured is specific to the type of exception this situation would encounter.

dimbleby:
Did you test this?  I don't think it will work.

IIRC the flow that you are trying to restore just didn't work: unless running poetry with verbose output - and why would you do that? - the user doesn't even see the prompt, and so assumes that poetry has hung.

andrewschueler:
I have tested the original code being set to "0" as well as removing the offending line of code `env['GIT_TERMINAL_PROMPT`] = "0"` and the code worked as expected. The git cloning process was interrupted by the subprocess asking for a password. Upon typing the password, the cloning process resumed. I should also mention that the `toml` file had `parallel = false`.

andrewschueler:
I have tested this PR in a private environment that cannot be shared due to the nature of the git repositories in question (they aren't hosted on github). I am still thinking of a simple test case for this PR that can be shows publicly.

andrewschueler:
I was able to run this fix on poetry 1.2.2 today and the authentication request popped up as expected.

andrewschueler:
If a user adds a dependency, then runs `poetry update` on this PR, the user has to add `-vvv` to get the prompt to show up.

PleahMaCaka (approved):

abn:
I do not think this is desired as this can open up a can of worms. It is better to fail on failed access attempts requiring user user to unlock or setup credentials. Please refer to the [documentation](https://python-poetry.org/docs/dependency-specification#git-dependencies) on instructions to add credentials.

```sh
poetry config repositories.git-org-project https://github.com/org/project.git
poetry config http-basic.git-org-project username token
poetry add git+https://github.com/org/project.git
```

Alternatively, enable and use `gitcredentials`.

PR #6877, opened 2022-10-24
Handle distribution.files being None

jace-ys:
# Pull Request Check List

Resolves: #6788 

We discovered that in some unknown situations, `distribution.files` can be `None`, causing the assertion to fail. To handle `distribution.files` being `None` more gracefully, we treat it as if it were an empty list.

### DIFF
--- a/src/poetry/utils/env.py
+++ b/src/poetry/utils/env.py
@@ -329,7 +329,7 @@ class SitePackages:
             name=distribution_name, writable_only=writable_only
         ):
             if not distribution.files:
-                break
+                continue
 
             for file in distribution.files:
                 if file.name.endswith(suffix):
@@ -344,7 +344,7 @@ class SitePackages:
             name=distribution_name, writable_only=writable_only
         ):
             if not distribution.files:
-                break
+                continue
 
             for file in distribution.files:
                 if file.name == name:
@@ -377,7 +377,7 @@ class SitePackages:
             name=distribution_name, writable_only=True
         ):
             if not distribution.files:
-                break
+                continue
 
             for file in distribution.files:
                 path = Path(
--- a/src/poetry/utils/env.py
+++ b/src/poetry/utils/env.py
@@ -328,8 +328,8 @@ class SitePackages:
         for distribution in self.distributions(
             name=distribution_name, writable_only=writable_only
         ):
-            assert distribution.files is not None
-            for file in distribution.files:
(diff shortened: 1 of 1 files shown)


neersighted (changes requested):
I do think this change is a bit clunky, I'd much rather do something like:

```py
if not distribution.files:
    break
```

Otherwise, the spirit of the thing looks good to me.

jace-ys:
Thanks for the suggestion @neersighted, definitely much neater than what I had - fixed it 👍🏻

jace-ys:
Good catch! Just updated it.

dimbleby:
this looks suspect in the third place you've done it: we now don't execute the code after the `for file in distribution.files`, which we previously did for an empty list.

fwiw I preferred the original version of this MR that went `files = [] if distribution.files is None else distribution.files`...

jace-ys:
@dimbleby Yeah good point. I think `files = [] if distribution.files is None else distribution.files...` has the benefit that it would easily solve the problem you highlighted.

I don't write Python much so don't have strong opinions on which approach is preferred - happy to defer to you folks!

jace-ys:
@neersighted Are you happy for me to revert it back to the original `files = [] if distribution.files is None else distribution.files` version?

Otherwise, I think the other option is to nest the inner for loo:
```
if distribution.files is not None:
    for file in distribution.files:
...
```

neersighted:
Ah, good catch, I skimmed too quickly there as well :laughing:

Yeah, I'd rather just indent/nest the for loop than create an empty list, but that's my personal preference.

jace-ys:
I've reverted it back to the original version I had 👍🏻

neersighted (approved):

leocwolter:
Hey, are there any workarounds we can use before this is released?

ffreemt:
Yeah, any workarounds? Am also struggling with a similar situation.

leocwolter:
Hey, I just noticed that this only happens when installing poetry via apt.
If you install it using the official installation script it works fine
(that was the case for me at least)

neersighted:
> Hey, are there any workarounds we can use before this is released?
#7137 is under review, we should be releasing shortly.

> Hey, I just noticed that this only happens when installing poetry via apt.

Versions of Poetry distributed by distro packagers are unsupported and carry untested patches. Poetry's test suite is stripped by Debian and not run to validate their changes. We can provide no support for distro-packaged versions of Poetry.

Issue #6713, opened 2022-10-05
Replace the `secondary` source type with more granular types


neersighted:
Poetry currently will look up packages in every source, even when unnecessary due to a `source =`  constraint on a dependency, or if the dependency was already found. This behavior often surprises users (though it was originally modeled on pip). Ostensibly, `secondary` is the solution to this -- however, it merely pushes the repository further back in the search order and does little to address the user complaints/confusion. As such, we should deprecate it and replace it with new options that better match user expectations.

I think, to preserve backwards compatibility, the long-term path forward is probably two different options (in my example, `supplemental` and `private`). I would propose something like the following, as an overview of behaviors:

* No options means that the repository overlays PyPI, but we will still fall back to PyPI if necessary (e.g. suitable for internal mirrors), `pip --extra-index-url`-style. This is known internally as primary, and is the current behavior.
* `default = true` means that the repository replaces PyPI, `pip --index-url`-style. This is the current behavior.
* `supplemental = true` means that this source is only consulted after a lookup in the `default` (implicitly PyPI unless configured otherwise) and primary sources fail.
* `private = true` means that the repo will only be considered for packages that are explicitly configured with `source =`, and should be preferred by users for private packages to avoid dependency confusion attacks.

`secondary = true` would be kept around as a deprecated option (likely with a warning), and would maintain the legacy behavior of being searched exhaustively for backwards compatibility.

Originally https://github.com/python-poetry/poetry/pull/5984#issuecomment-1237245571

neersighted:
#6669 represented incremental progress towards this goal, and the author is interested in implementing this proposal.

MasterNayru:
Would it be fair say that a contributing factor to issues that users have when managing multiple repositories is that Poetry assumes the worst case scenario when it comes to how to obtain metadata from non-PyPI repositories? Lookups to PyPI are done through APIs and happen way quicker than the HTML parsing that happens with the other repositories. Obviously, Poetry has gone down that route to ensure that things work and don't just explode because some backend didn't whatever APIs PyPI is currently exposing.

Just seems to me that while querying repos unnecessarily is worth avoiding, a real pain point is how these other package repositories end up being utilised.

neersighted:
I did not reference it here as it is out of scope for this issue; however, yes, that is a problem. However, it is an ecosystem problem:

[...]

PR #6879, opened 2022-10-24
6713 Introduce supplemental package source

b-kamphorst:
# Pull Request Check List

Resolves: https://github.com/python-poetry/poetry/issues/6713
Relates to: https://github.com/python-poetry/poetry/pull/7658

### DIFF
--- a/src/poetry/console/commands/source/add.py
+++ b/src/poetry/console/commands/source/add.py
@@ -115,7 +115,7 @@ class SourceAddCommand(Command):
             priority = Priority[priority_str.upper()]
 
         if priority is Priority.SECONDARY:
-            allowed_prios = filter(lambda p: p is not Priority.SECONDARY, Priority)
+            allowed_prios = (p for p in Priority if p is not Priority.SECONDARY)
             self.line_error(
                 "<warning>Warning: Priority 'secondary' is deprecated. Consider"
                 " changing the priority to one of the non-deprecated values:"
--- a/src/poetry/console/commands/source/add.py
+++ b/src/poetry/console/commands/source/add.py
@@ -114,6 +114,14 @@ class SourceAddCommand(Command):
         else:
             priority = Priority[priority_str.upper()]
 
+        if priority is Priority.SECONDARY:
+            allowed_prios = filter(lambda p: p is not Priority.SECONDARY, Priority)
+            self.line_error(
+                "<warning>Warning: Priority 'secondary' is deprecated. Consider"
+                " changing the priority to one of the non-deprecated values:"
+                f" {', '.join(repr(p.name.lower()) for p in allowed_prios)}.</warning>"
+            )
+
         sources = AoT([])
         new_source = Source(name=name, url=url, priority=priority)
         is_new_source = True
--- a/src/poetry/factory.py
+++ b/src/poetry/factory.py
@@ -156,11 +156,11 @@ class Factory(BaseFactory):
                     priority = Priority.SECONDARY
 
             if priority is Priority.SECONDARY:
-                allowed_prios = filter(lambda p: p is not Priority.SECONDARY, Priority)
+                allowed_prios = (p for p in Priority if p is not Priority.SECONDARY)
                 warning = (
-                    "Found deprecated priority 'secondary' in pyproject.toml"
-                    f" configuration for source {source.get('name')}. Consider"
-                    " changing the priority to one of the non-deprecated values:"
+                    "Found deprecated priority 'secondary' for source"
+                    f" '{source.get('name')}' in pyproject.toml. Consider changing the"
+                    " priority to one of the non-deprecated values:"
                     f" {', '.join(repr(p.name.lower()) for p in allowed_prios)}."
                 )
                 io.write_error_line(f"<warning>Warning: {warning}</warning>")
--- a/src/poetry/factory.py
+++ b/src/poetry/factory.py
@@ -155,6 +155,16 @@ class Factory(BaseFactory):
                 elif source.get("secondary"):
                     priority = Priority.SECONDARY
 
+            if priority is Priority.SECONDARY:
+                allowed_prios = filter(lambda p: p is not Priority.SECONDARY, Priority)
+                warning = (
+                    "Found deprecated priority 'secondary' in pyproject.toml"
+                    f" configuration for source {source.get('name')}. Consider"
+                    " changing the priority to one of the non-deprecated values:"
+                    f" {', '.join(repr(p.name.lower()) for p in allowed_prios)}."
+                )
+                io.write_error_line(f"<warning>Warning: {warning}</warning>")
+
             if io.is_debug():
                 message = f"Adding repository {repository.name} ({repository.url})"
                 if priority is Priority.DEFAULT:
--- a/src/poetry/json/schemas/poetry.json
+++ b/src/poetry/json/schemas/poetry.json
@@ -45,6 +45,7 @@
             "primary",
             "default",
             "secondary",
+            "supplemental",
             "explicit"
           ],
           "description": "Declare the priority of this repository."
--- a/src/poetry/repositories/repository_pool.py
+++ b/src/poetry/repositories/repository_pool.py
@@ -187,11 +187,13 @@ class RepositoryPool(AbstractRepository):
 
         packages: list[Package] = []
         for repo in self.repositories:
+            if packages and self.get_priority(repo.name) is Priority.SUPPLEMENTAL:
+                break
             packages += repo.find_packages(dependency)
         return packages
 
     def search(self, query: str) -> list[Package]:
         results: list[Package] = []
-        for repository in self.repositories:
-            results += repository.search(query)
+        for repo in self.repositories:
+            results += repo.search(query)
         return results
--- a/src/poetry/repositories/repository_pool.py
+++ b/src/poetry/repositories/repository_pool.py
@@ -28,6 +28,7 @@ class Priority(IntEnum):
     DEFAULT = enum.auto()
     PRIMARY = enum.auto()
     SECONDARY = enum.auto()
+    SUPPLEMENTAL = enum.auto()
     EXPLICIT = enum.auto()
 
 
--- a/tests/test_factory.py
+++ b/tests/test_factory.py
@@ -402,10 +402,9 @@ def test_poetry_with_non_default_multiple_sources_pypi(
     error = io.fetch_error()
     assert (
         error.strip()
-        == "<warning>Warning: Found deprecated priority 'secondary' in pyproject.toml"
-        " configuration for source foo. Consider changing the priority to one of the"
-        " non-deprecated values: 'default', 'primary', 'supplemental',"
-        " 'explicit'."
+        == "<warning>Warning: Found deprecated priority 'secondary' for source 'foo' in"
+        " pyproject.toml. Consider changing the priority to one of the"
+        " non-deprecated values: 'default', 'primary', 'supplemental', 'explicit'."
     )
 
 
--- a/tests/test_factory.py
+++ b/tests/test_factory.py
@@ -400,7 +400,13 @@ def test_poetry_with_non_default_multiple_sources_pypi(
     expected = ["bar", "PyPI", "baz", "foo"]
     assert [repo.name for repo in poetry.pool.repositories] == expected
     error = io.fetch_error()
-    assert error == ""
+    assert (
+        error.strip()
+        == "<warning>Warning: Found deprecated priority 'secondary' in pyproject.toml"
+        " configuration for source foo. Consider changing the priority to one of the"
+        " non-deprecated values: 'default', 'primary', 'supplemental',"
+        " 'explicit'."
+    )
 
 
 def test_poetry_with_no_default_source(fixture_dir: FixtureDirGetter) -> None:
--- a/tests/test_factory.py
+++ b/tests/test_factory.py
@@ -298,7 +298,7 @@ def test_poetry_with_non_default_secondary_source_legacy(
     assert poetry.pool.get_priority("PyPI") is Priority.DEFAULT
     assert poetry.pool.has_repository("foo")
     assert isinstance(poetry.pool.repository("foo"), LegacyRepository)
-    assert [repo.name for repo in poetry.pool.repositories] == ["PyPI", "foo"]
+    assert {repo.name for repo in poetry.pool.repositories} == {"PyPI", "foo"}
 
 
 def test_poetry_with_non_default_secondary_source(
@@ -311,7 +311,7 @@ def test_poetry_with_non_default_secondary_source(
     assert poetry.pool.get_priority("PyPI") is Priority.DEFAULT
     assert poetry.pool.has_repository("foo")
     assert isinstance(poetry.pool.repository("foo"), LegacyRepository)
-    assert [repo.name for repo in poetry.pool.repositories] == ["PyPI", "foo"]
+    assert {repo.name for repo in poetry.pool.repositories} == {"PyPI", "foo"}
 
 
 def test_poetry_with_non_default_multiple_secondary_sources_legacy(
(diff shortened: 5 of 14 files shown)


b-kamphorst:
This is still work in progress. In particular, the following items are on my TODO list:
- [ ] add tests to `tests/test_factory` (incl. adding fixtures)
- [ ] add tests to `tests/commands/source/test_add.py`
- [ ] add behaviour to `source add`
- [ ] update documentation `poetry source add`

For backwards compatibility I complemented the `default` and `secondary` keys for sources in `pyproject.toml` with `supplemental` and `private`. However it seems more maintainable to introduce a non-boolean key `priority`. How do you feel about the suggestion (and its relation to this PR)?

neersighted:
I don't think integer priorities make sense as there are specific semantics beyond order that some of the flags indicate (e.g. `default` and `private` have meanings beyond the lookup order).

b-kamphorst:
> I don't think integer priorities make sense as there are specific semantics beyond order that some of the flags indicate (e.g. `default` and `private` have meanings beyond the lookup order).

I agree. I was thinking of literals; e.g. 

```toml
[[tool.poetry.source]]
name = "foo"
type = "supplemental " # type = "primary" if unspecified
```

instead of 

```toml
[[tool.poetry.source]]
name = "foo"
default = false
secondary = false
supplemental = true
private = false
```

neersighted:
Ah, that makes more sense -- I read that as integer, but I see that you didn't explicitly state that.

I'm not necessarily opposed, but I think we would need to think about the design and how it would interact with older Poetry versions (maybe it's time to add a `minimum-poetry-version` gate?), as well as the potential `indexed = true` from #5442 and if we can easily validate the string-constants and provide helpful errors (I think our JSON schema lets us do this trivially).

dimbleby:
Haven't got any further than starting to read the docs yet: but it looks as though `private` is not a good name hree.  Other types of source might be private, and `private` sources might not be private.  The name is completely disconnected from the meaning of the word!

`explicit` maybe?

dimbleby:
this reminds me that I still think it's a good idea to spell out what key use cases you're trying to cover here, and how this approach would meet them - https://github.com/python-poetry/poetry/issues/6713#issuecomment-1270562226

in particular two of the cases mentioned there exactly show what I mean about `private` vs "private"

- a private mirror from which I want to get everything is private but not `private`
- the pytorch repository from which folk want to get pytorch only is `private` but not private

b-kamphorst:
> Haven't got any further than starting to read the docs yet: but it looks as though `private` is not a good name here. Other types of source might be private, and `private` sources might not be private. The name is completely disconnected from the meaning of the word!
> 
> `explicit` maybe?

I quite like `explicit`, I'll rename if @neersighted agrees.

b-kamphorst:
I've continued the effort through the cli and factory. I expect that there is some test to actually verify that the correct information ends up in `pyproject.toml`, but I have not found it yet. Feel free to point me in the right direction!

Feedback is welcome. Note that there are some open discussion that need to be concluded, e.g. https://github.com/python-poetry/poetry/pull/6879#issuecomment-1296348489 and https://github.com/python-poetry/poetry/issues/6713#issuecomment-1296350113.

b-kamphorst:
Tests for `poetry-plugin-export` are failing (as in my previous MR). The obvious conflicts are in the code ([here](https://github.com/python-poetry/poetry-plugin-export/blob/dd00a2fa908158f0c9a5d1f92c12f235f54c8a85/src/poetry_plugin_export/exporter.py#L189)) and tests ([here](https://github.com/python-poetry/poetry-plugin-export/blob/dd00a2fa908158f0c9a5d1f92c12f235f54c8a85/tests/test_exporter.py#L1851) and [here](https://github.com/python-poetry/poetry-plugin-export/blob/dd00a2fa908158f0c9a5d1f92c12f235f54c8a85/tests/test_exporter.py#L1859)). These can easily be fixed, but we end up with a circular dependency (plugin will fail for current version of poetry, new poetry breaks plugin). Thoughts?

NTS: also add tests for `Priority.EXPLICIT` in `poetry-export`.

radoering:
Is there a discussion somewhere why we should change the priority of `default` and `primary`?

> I expect that there is some test to actually verify that the correct information ends up in `pyproject.toml`, but I have not found it yet.

I'd just deliberately introduce a bug, run the tests locally and see if some tests fail. If all tests pass, then there is no (sufficient) test. 😉

> Note that there are some open discussion that need to be concluded, e.g. [#6879 (comment)](https://github.com/python-poetry/poetry/pull/6879#issuecomment-1296348489) and [#6713 (comment)](https://github.com/python-poetry/poetry/issues/6713#issuecomment-1296350113).

> explicit maybe?

IMO, `explicit` is fine.

> is supplemental behaviour actually breaking for secondary? If secondary sources are considered after all other (primary, default) sources, then the resolution behaviour should be identical (apart from making / timing of queries). If not, what am I missing?

With `secondary`, you can get an additional version of a package, which is not available in primary sources and resolution might choose exactly that version.

src/poetry/repositories/repository_pool.py
```diff
@@ -77,7 +85,9 @@ def repository(self, name: str) -> Repository:
         raise IndexError(f'Repository "{name}" does not exist.')
 
     def add_repository(
-        self, repository: Repository, default: bool = False, secondary: bool = False
+        self,
+        repository: Repository,
+        priority: Priority = Priority.PRIMARY,
```
radoering, line 90: If I remember correctly, this function is used by plugin authors (not only poetry-plugin-export). Maybe, we should deprecate the parameters, keep them functional and add `priority` as keyword-only argument for now.

neersighted:
Just checking in to say sorry for going AWOL; this takes more energy than I have the ability to contribute now due to other commitments. You're in good hands with @radoering and I'll try to provide input again when I am able.

b-kamphorst:
> Is there a discussion somewhere why we should change the priority of `default` and `primary`?

This is actually a bug. `primary` sources are supposed to behave as the pip `--extra-index-url`, which have priority over `--index-url` -- our `default`. However, it turns out that the bug fix is incompatible with the `poetry-plugin-export` ([this line](https://github.com/python-poetry/poetry-plugin-export/blob/dd00a2fa908158f0c9a5d1f92c12f235f54c8a85/src/poetry_plugin_export/exporter.py#L189)). The method `RepositoryPool.get_priority` will make it easy to change that line in a separate PR.

I propose to merge this PR in `poetry` without fixing the priority, then updating `poetry-plugin-export` to not depend on the order of repositories in `pool.repositories` but rather just verify the priority of a given repo, and then fix the bug in `poetry`. 

> I'd just deliberately introduce a bug, run the tests locally and see if some tests fail. If all tests pass, then there is no (sufficient) test. 😉

TODO on my side.

> IMO, `explicit` is fine.

OK!

> With `secondary`, you can get an additional version of a package, which is not available in primary sources and resolution might choose exactly that version.

Ah you are right. I got confused between the version pickers (there is one [here](https://github.com/python-poetry/poetry/blob/daa87b2db053319ebbef7709fd0fd7721a46eb5b/src/poetry/version/version_selector.py) and another [here](https://github.com/python-poetry/poetry/blob/daa87b2db053319ebbef7709fd0fd7721a46eb5b/src/poetry/puzzle/provider.py#L332)) and I mistakenly thought that one of them did not sort and as such would just pick the first returned version (which should be a non-secondary source when available).

b-kamphorst:
> Just checking in to say sorry for going AWOL; this takes more energy than I have the ability to contribute now due to other commitments. You're in good hands with @radoering and I'll try to provide input again when I am able.

No worries, I appreciate both of your feedback and hope that I can help poetry with this PR :)

b-kamphorst:
> I'd just deliberately introduce a bug, run the tests locally and see if some tests fail. If all tests pass, then there is no (sufficient) test. 😉

Done -- all works!

@radoering I think this PR is ready for another round of review. I'll probably have to remove the DEFAULT-PRIMARY-changing commits and you may prefer some other tidying up as well. Just let me know what you need for approval.

radoering:
> This is actually a bug. `primary` sources are supposed to behave as the pip `--extra-index-url`, which have priority over `--index-url` -- our `default`.

IMO that's not that clear. Actually, there is no warranted order of `--extra-index-url` and `--index-url`. IIUC, pip considers all sources equal and it's an implementation detail which source is chosen first. See https://github.com/pypa/pip/issues/9610#issuecomment-778796972 for example. That's why I doubt whether we should change order at all.

> I propose to merge this PR in `poetry` without fixing the priority

I absolutely agree on this point.

I'll try to do a review this week or next.

radoering (commented):
I haven't looked at the tests yet, but I think there are some implementation details that need to be worked out first.

Coming back to the `default` priority: I think the issue is that at the moment (state of current master) we are inconsistent between implementation and documentation:

|implementation|documentation|
|---|---|
|If there is no explicit `default`, **PyPI is only the `default` if there is no `primary`**, otherwise it's `secondary`. **`default` is the highest priority.** In other words, implicit PyPI always comes after `primary` but is not always the `default`.|It's a bit difficult to find but it says that primary sources take precedence over PyPI and that the default source takes precedence over `secondary` sources. It does not clearly say but implies that **if there is no explicit `default`, PyPI always is the `default`**. If you combine everything, it seems **`primary` sources should take precedence over the `default` source** (even if it's not PyPI).|

This inconsistence makes it difficult to decide if changes made in this PR are correct or not. I'm still not sure what to consider the source of truth. I think it's difficult to teach the current implemented behavior. Thus, it might make sense to consider the documentation as correct and change the implementation so that `default` is in between `primary` and `secondary`. If we consider changing the order a bugfix, we should do this in a separate PR (maybe even before this PR). I'll try to discuss this in the next maintainer meeting and give you some feedback.

src/poetry/config/source.py
```diff
@@ -7,8 +7,10 @@
 class Source:
     name: str
     url: str
-    default: bool = dataclasses.field(default=False)
-    secondary: bool = dataclasses.field(default=False)
+    default: bool = False
+    secondary: bool = False
+    supplemental: bool = False
+    explicit: bool = False
```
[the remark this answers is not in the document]
b-kamphorst: Well that was quite a headache, but educative as well. I look forward to your thoughts on the last two commits, which introduce the enum.

src/poetry/factory.py
```diff
@@ -125,31 +126,32 @@ def configure_sources(
             repository = cls.create_package_source(
                 source, config, disable_cache=disable_cache
             )
-            is_default = bool(source.get("default", False))
-            is_secondary = bool(source.get("secondary", False))
+            priority = Priority[source.get("priority", Priority.PRIMARY.name).upper()]
+            if source.get("default"):
```
b-kamphorst, line 130: Does this need to display a warning so that users can migrate their existing `pyproject.toml`? We now only display warnings when calling `poetry source add`. `Source` also generates warnings, but `Factory` does not call that with legacy parameters.

src/poetry/json/schemas/poetry.json
```diff
@@ -39,7 +39,17 @@
         },
         "secondary": {
           "type": "boolean",
-          "description": "Declare this repository as secondary, i.e. it will only be looked up last for packages."
+          "description": "Declare this repository as secondary, i.e. default repositories take precedence."
+        },
+        "priority": {
```
b-kamphorst, line 44: TODO: Ensure priority mutually exclusive with default and secondary.

src/poetry/factory.py
```diff
@@ -125,31 +126,32 @@ def configure_sources(
...
         if poetry.pool.has_default():
             if io.is_debug():
                 io.write_line("Deactivating the PyPI repository")
         else:
             from poetry.repositories.pypi_repository import PyPiRepository
 
-            default = not poetry.pool.has_primary_repositories()
             poetry.pool.add_repository(
-                PyPiRepository(disable_cache=disable_cache), default, not default
+                PyPiRepository(disable_cache=disable_cache), priority=Priority.DEFAULT
```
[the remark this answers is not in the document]
b-kamphorst: Hi @radoering, do you have any updates on the requested behaviour of the PyPI repository?

src/poetry/factory.py
b-kamphorst: I have added a warning.

src/poetry/factory.py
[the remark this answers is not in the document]
radoering: Unfortunately, not yet. I wanted to discuss this topic in the maintainer meeting, which is scheduled every two weeks, but the last meeting was cancelled. Thus, I'll try in the next meeting.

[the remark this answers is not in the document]
radoering: See https://github.com/python-poetry/poetry/issues/6713#issuecomment-1435487741 for the results of the discussion in our meeting. I'd like to hear your opinion on the two variants.

radoering:
May you split the introduction of `explicit` into a separate PR as first step? I think that's the most uncontroversial part since it's not influenced by order or priority of any other repository. (Maybe, we can switch from flags to one setting as proposed in https://github.com/python-poetry/poetry/pull/6879#discussion_r1089703393 in this PR, too.)

Another reason I like to see `explicit` first is that I'm not quite convinced by #7430 and want to try an alternative approach that requires `explicit`.

Secrus:
Is this still a valid PR since #7658 was merged? If not, please close.

radoering:
#7658 was only part of this one. Introducing `supplementary` is still open. However, it might be easier to close this one and open a new PR? Further, it might make sense to wait for #7801 to avoid conflicts.

radoering:
@b-kamphorst Since #7801 has been merged, we can go for `supplemental` now.

b-kamphorst:
> @b-kamphorst Since #7801 has been merged, we can go for `supplemental` now.

I've rebased the MR on master. Most looks fine to me, but surely a fresh review is welcome.

tests/test_factory.py
```diff
@@ -298,7 +298,7 @@ def test_poetry_with_non_default_secondary_source_legacy(
     assert poetry.pool.get_priority("PyPI") is Priority.DEFAULT
     assert poetry.pool.has_repository("foo")
     assert isinstance(poetry.pool.repository("foo"), LegacyRepository)
-    assert [repo.name for repo in poetry.pool.repositories] == ["PyPI", "foo"]
+    assert {repo.name for repo in poetry.pool.repositories} == {"PyPI", "foo"}
```
radoering, line 301: What is the reason for this change? Since repositories are ordered by convention, we might as well check the correct order.

src/poetry/console/commands/source/add.py
```diff
@@ -114,6 +114,14 @@ def handle(self) -> int:
         else:
             priority = Priority[priority_str.upper()]
 
+        if priority is Priority.SECONDARY:
+            allowed_prios = filter(lambda p: p is not Priority.SECONDARY, Priority)
```
radoering, line 118:
```suggestion
            allowed_prios = (p for p in Priority if p is not Priority.SECONDARY)
```

src/poetry/factory.py
```diff
@@ -155,6 +155,16 @@ def create_pool(
                 elif source.get("secondary"):
                     priority = Priority.SECONDARY
 
+            if priority is Priority.SECONDARY:
+                allowed_prios = filter(lambda p: p is not Priority.SECONDARY, Priority)
```
radoering, line 159:
```suggestion
                allowed_prios = (p for p in Priority if p is not Priority.SECONDARY)
```

src/poetry/factory.py
```diff
@@ -155,6 +155,16 @@ def create_pool(
...
+                warning = (
+                    "Found deprecated priority 'secondary' in pyproject.toml"
+                    f" configuration for source {source.get('name')}. Consider"
```
radoering, line 162:
```suggestion
                    "Found deprecated priority 'secondary' for source"
                    f" {source.get('name')} in pyproject.toml. Consider"
```

radoering (changes requested):
Looks good. Just some minor nitpicks.

tests/test_factory.py
b-kamphorst: We validate the order in `test_repository_pool.py`. My rationale here is that the (tests in) `factory` should not have to change if we decide to change the repository order in `RepositoryPool` as this is an implementation detail of the object that it creates. #7658 therefore introduces set comparisons in `test_factory` (6x) but failed to do this consistently (3 remaining cases of list comparisons) which I corrected here. Alternatively, we can replace the set comparisons by list comparisons. Please let me know which option you prefer.

radoering (approved):

PR #6882, opened 2022-10-25
Switch from -dev to released 3.11 version

Secrus:
As in title, switching Python 3.11 in CI from beta to full release


mkniewallner (commented):
We can also remove https://github.com/python-poetry/poetry/blob/0ffe91c0c8e2edde987cf2213d402614a9c2a6a6/.github/workflows/main.yml#L86-L89 since `pip==22.3`, which is [bundled in 3.11.0](https://github.com/python/cpython/blob/v3.11.0/Lib/ensurepip/__init__.py#L14), fixes the issue.

mkniewallner (approved):

Issue #6887, opened 2022-10-25
Poetry add help discrepancy


tjzegmott:
- [x] I have searched the [issues](https://github.com/python-poetry/poetry/issues) of this repo and believe that this is not a duplicate.

## Issue

There is a discrepancy between the documentation when comparing `poetry add --help` and the webpage https://python-poetry.org/docs/cli/#add

The discrepancy is for adding a git repo using ssh:
- webpage:
    ```
    poetry add git+ssh://git@github.com/sdispater/pendulum.git
    # or alternatively:
    poetry add git+ssh://git@github.com:sdispater/pendulum.git
    ```
- CLI help:
    ```
     - A git SSH url (git+ssh://github.com/python-poetry/poetry.git)
    ```

The CLI help is missing the "git@" part.

dimbleby:
submit a pull request!

PR #6889, opened 2022-10-25
docs(poetry-add-help): resolve discrepancy between cli help and webpa…

tjzegmott:
…ge docs

# Pull Request Check List

Resolves: #6887 


neersighted:
I actually think we should drop the SSH lines entirely -- a "Git URL" encompasses multiple protocols, such as HTTPS, HTTP (see #6619), Git (the wire-protocol), and SSH. We could add a list of accepted protocols as Poetry does not (currently) accept HTTP, and we do not intend to support `git+file://` or `git+git://` if you think being explicit is better; however, the latter are so rarely used that I think the error message might be sufficient.

neersighted (changes requested):
As per my PR comment.

PR #6929, opened 2022-10-31
fix: remove side effects from tests

chadac:
# Pull Request Check List

Found some cases where tests were writing to the Poetry test runtime environment rather than a mocked local directory. This is an issue, since it meant that some tests were technically stateful when rerunning locally.

Added in placeholders for those tests. Tested that this succeeds by using a build where my Poetry venv was made entirely read-only.

Resolves: N/A

### DIFF
--- a/tests/installation/test_installer.py
+++ b/tests/installation/test_installer.py
@@ -174,8 +174,8 @@ def locker(project_root: Path) -> Locker:
 
 
 @pytest.fixture()
-def env() -> NullEnv:
-    return NullEnv()
+def env(tmp_dir: str) -> NullEnv:
+    return NullEnv(path=Path(tmp_dir))
 
 
 @pytest.fixture()
--- a/tests/installation/test_installer.py
+++ b/tests/installation/test_installer.py
@@ -1151,12 +1151,13 @@ def test_installer_with_pypi_repository(
     locker: Locker,
     installed: CustomInstalledRepository,
     config: Config,
+    env: NullEnv,
 ):
     pool = RepositoryPool()
     pool.add_repository(MockRepository())
 
     installer = Installer(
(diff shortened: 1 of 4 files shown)


radoering (commented):
Nice. 👍

Is there a reason that `NullEnv()` without a tmp path is still used instead of the fixture once in each installer test module?

chadac:
> Nice. +1
> 
> Is there a reason that `NullEnv()` without a tmp path is still used instead of the fixture once in each installer test module?

Good catch. I missed those instances, and they must not cause any side effects since they weren't causing my RO tests to complain. I've removed those instantiations you mentioned, plus any others I could find.

neersighted:
I'd rather new uses start using the built-in pytest fixtures, which provide a `Path`: https://docs.pytest.org/en/7.1.x/how-to/tmp_path.html

chadac:
Nice, migrated all uses here to tmp_path as well. On that note... there isn't much left using `tmp_dir`. Would it be all right if I threw up an issue and assigned myself to it to knock that out?

tests/installation/test_installer.py
```diff
@@ -174,8 +174,8 @@ def locker(project_root: Path) -> Locker:
 
 
 @pytest.fixture()
-def env() -> NullEnv:
-    return NullEnv()
+def env(tmp_path: Path) -> NullEnv:
+    return NullEnv(path=Path(tmp_path))
```
neersighted, line 178: You can pass `tmp_path` directly now instead of constructing a new `Path` from it.

neersighted:
> [...] there isn't much left using `tmp_dir`. Would it be all right if I threw up an issue and assigned myself to it to knock that out?

Please do!

radoering (approved):

PR #6936, opened 2022-11-02
example won't work with poetry (1.2.2) install

wensenfriendandextra:
in toml file
[tool.poetry.dependencies]
pendulum = "^2.1"

This will cause "pendumlum (^2.1) which doesn't match any versions, version solving failed."

need to specify a full version like "^2.1.1"

# Pull Request Check List

Resolves: #issue-number-here

### DIFF
--- a/docs/basic-usage.md
+++ b/docs/basic-usage.md
@@ -82,7 +82,7 @@ If you want to add dependencies to your project, you can specify them in the `to
 
 ```toml
 [tool.poetry.dependencies]
-pendulum = "^2.1"
+pendulum = "^2.1.1"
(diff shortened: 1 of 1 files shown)


finswimmer:
Hello @wensenfriendandextra,

thanks a lot for your contribution. Unfortunately I'm not able to reproduce your issue. So I would suggest, that you open an issue ticket first and provide a reproducible example there. Than we can further debug where your problem comes from.

fin swimmer

Secrus:
I am going to close this since the problem doesn't reproduce and PR is currently unnecessary.

PR #6966, opened 2022-11-04
pip install --no-input everywhere

coobas:
This is an extension of https://github.com/python-poetry/poetry/pull/6724. I think `pip install` invoked by poetry should never ask for user input. 

Motivation is that it happened to myself and a number of colleagues many times that poetry got seemingly stuck while it was just waiting for a user input because of a private pypi repository that needed authentication.

I hope this is a valuable contribution to a tool I like a lot and would like to use more and more :)

### DIFF
--- a/src/poetry/inspection/info.py
+++ b/src/poetry/inspection/info.py
@@ -593,6 +593,7 @@ def get_pep517_metadata(path: Path) -> PackageInfo:
                 "install",
                 "--disable-pip-version-check",
                 "--ignore-installed",
+                "--no-input",
                 *PEP517_META_BUILD_DEPS,
             )
             venv.run(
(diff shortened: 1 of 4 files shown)


finswimmer (changes requested):
Thanks a lot for your contribution :+1: Overall it looks good to me.

I wonder if `--no-input` should be added here as well :thinking: https://github.com/python-poetry/poetry/blob/3a961dfb67ccc214fa101f5c1a187e2ed077673b/src/poetry/installation/pip_installer.py#L49

coobas:
> Thanks a lot for your contribution 👍 Overall it looks good to me.
> 
> I wonder if `--no-input` should be added here as well 🤔
> 
> https://github.com/python-poetry/poetry/blob/3a961dfb67ccc214fa101f5c1a187e2ed077673b/src/poetry/installation/pip_installer.py#L49

👍  I think so - I've added that. Thanks your prompt review!

finswimmer (approved):
LGTM :+1:

Issue #4870, opened 2021-12-08
Poetry fails to resolve dependencies for multiple Python versions


Darel13712:
- **OS version and name**: MacOS 10.15.7
- **Poetry version**: 1.1.12
- **pyproject.toml**: [Gist](https://gist.github.com/Darel13712/83c1aaafc9815fe323dbb938a4263757)
- **Python**: 3.9.7

## Issue
My project has a dependency on numpy and [lightautoml](https://github.com/sberbank-ai-lab/LightAutoML)
Lightautoml's pyproject has dependencies for numpy, and I am able to resolve their dependencies:
```
python = ">=3.6.1, <3.10"
numpy = [
  {version = ">=1.20.0", python = ">=3.7"},
  {version = "*", python = "<3.7"}
]
```
However, if I create my own pyproject, that depends both on numpy and lightautoml, dependencies cannot be resolved:
```
[tool.poetry.dependencies]
python = ">=3.6.2, <3.10"
lightautoml = "==0.3.2"
numpy = [
  {version = ">=1.20.0", python = ">=3.7"},
  {version = "*", python = "<3.7"}
]
```
It seems that I do not add any new requirements in my pyproject, but if i try to run `poetry lock`, I get 

```
The current project's Python requirement (>=3.6.2,<3.10) is not compatible with some of the required packages Python requirement:
    - numpy requires Python >=3.7,<3.11, so it will not be satisfied for Python >=3.6.2,<3.7
    - numpy requires Python >=3.7, so it will not be satisfied for Python >=3.6.2,<3.7
    - numpy requires Python >=3.7, so it will not be satisfied for Python >=3.6.2,<3.7
    - numpy requires Python >=3.7, so it will not be satisfied for Python >=3.6.2,<3.7
    - numpy requires Python >=3.7, so it will not be satisfied for Python >=3.6.2,<3.7
    - numpy requires Python >=3.7, so it will not be satisfied for Python >=3.6.2,<3.7
    - numpy requires Python >=3.7, so it will not be satisfied for Python >=3.6.2,<3.7
    - numpy requires Python >=3.7,<3.11, so it will not be satisfied for Python >=3.6.2,<3.7
    - numpy requires Python >=3.7,<3.11, so it will not be satisfied for Python >=3.6.2,<3.7
  
[...]

PR #6969, opened 2022-11-04
Handle restricted dependencies as implicit multiple-constraints dependencies

### LINKED ISSUE #5506: Solver breaks faced with related dependencies that are both conditional

- **OS version and name**:  macOS Monterey 12.3.1
- **Poetry version**:  1.1.13
- **Link of a [Gist](https://gist.github.com/) with the contents of your pyproject.toml file**: [Here](https://gist.github.com/anthonymichaelclark/dad50a1444e523846d2ccd9dace4f817)

## Issue
The solver breaks when faced with dependencies A and B that are both conditional but where A depends on B only sometimes.

Consider these lines from the attached `pyproject.toml`
```
dbt = { version = "0.19.2", markers = "sys_platform != 'darwin'" }
dbt-bigquery = [
    { version = "0.19.2", markers = "sys_platform != 'darwin'" },
    { version = "^1.0.0", markers = "sys_platform == 'darwin' "},
]
```
The intent here is "Install dbt==0.19.2 and dbt-bigquery==0.19.2 when `sys_platform` is not `'darwin'`; else install `dbt-bigquery==^1.0.0`. When trying to install on a machine using `sys_platform == 'darwin'` I get:

```
Updating dependencies
Resolving dependencies... (12.4s)

  SolverProblemError

  Because analytics-dbt depends on dbt (0.19.2) which depends on dbt-bigquery (0.19.2), dbt-bigquery is required.
  So, because analytics-dbt depends on dbt-bigquery (^1.0.0), version solving failed.

  at /opt/homebrew/Cellar/poetry/1.1.13/libexec/lib/python3.10/site-packages/poetry/puzzle/solver.py:241 in _solve
      237│             packages = result.packages
      238│         except OverrideNeeded as e:
      239│             return self.solve_in_compatibility_mode(e.overrides, use_latest=use_latest)
      240│         except SolveFailure as e:
    → 241│             raise SolverProblemError(e)
      242│ 
      243│         results = dict(
      244│             depth_first_search(
      245│                 PackageNode(self._package, packages), aggregate_package_nodes

```

radoering:
# Pull Request Check List

Resolves: #5506

Although I think that this PR makes the solver more correct it comes with a massive performance regression that is far from acceptible.

I carried out some measurements with example pyproject.toml files from other PRs. If locking succeeds without this PR, the same lock file is generated with this PR, it just takes longer...

Times for `poetry lock` with a warm cache:

|---|---|---|
|#3367|0.8 s|2.8 s|
|#4670|1.8 s|4.2 s|
|#4870|71 s| 11800 s (not a typo)|
|#5506|error after 4.5 s|1090 s|
|shootout example|3.9 s|250 s|

Number of overrides:

|`pyproject.toml` from ...| number of overrides without PR | number of overrides with PR |
|---|---|---|
|#3367|4|10|
|#4670|16|19|
|#4870|46|4179|
|#5506|-|288|
|shootout example|0|69|

The data shows that the time seems to correlate with the number of overrides. Thus, I assume a more sophisticated algorithm to reduce the number of overrides or even a complete overhaul of how to handle multiple-constraints dependencies might be necessary. I can imagine to make the `VersionSolver` marker aware so that a version conflict is only a conflict if the intersection of markers is not empty. This way, overrides would not be necessary anymore and everything could be solved at once. However, that's probably a huge task.

### DIFF
--- a/src/poetry/puzzle/provider.py
+++ b/src/poetry/puzzle/provider.py
@@ -570,6 +570,9 @@ class Provider:
                         continue
                     self.search_for_direct_origin_dependency(dep)
 
+        active_extras = None if package.is_root() else dependency.extras
+        _dependencies = self._add_implicit_dependencies(_dependencies, active_extras)
+
         dependencies = self._get_dependencies_with_overrides(
             _dependencies, dependency_package
         )
@@ -606,7 +609,6 @@ class Provider:
 
             # For dependency resolution, markers of duplicate dependencies must be
             # mutually exclusive.
-            active_extras = None if package.is_root() else dependency.extras
             deps = self._resolve_overlapping_markers(package, deps, active_extras)
 
             if len(deps) == 1:
@@ -856,6 +858,42 @@ class Provider:
             and (not self._env or marker.validate(self._env.marker_env))
         )
 
+    def _add_implicit_dependencies(
+        self,
+        dependencies: Iterable[Dependency],
+        active_extras: Collection[NormalizedName] | None,
+    ) -> list[Dependency]:
+        """
+        This handles an edge case where the dependency is not required for a subset
+        of possible environments. We have to create such an implicit "not required"
+        dependency in order to not miss other dependencies later.
+        For instance:
+            • foo (1.0) ; python == 3.7
+            • foo (2.0) ; python == 3.8
+            • bar (2.0) ; python == 3.8
+            • bar (3.0) ; python == 3.9
+        the last dependency would be missed without this,
+        because the intersection with both foo dependencies is empty.
+
+        A special case of this edge case is a restricted dependency with a single
+        constraint, see https://github.com/python-poetry/poetry/issues/5506
+        for details.
+        """
+        by_name: dict[str, list[Dependency]] = defaultdict(list)
+        for dep in dependencies:
+            by_name[dep.name].append(dep)
+        for _name, deps in by_name.items():
+            marker = marker_union(*[d.marker for d in deps])
+            if marker.is_any():
+                continue
+            inverted_marker = marker.invert()
+            if self._is_relevant_marker(inverted_marker, active_extras):
+                # Set constraint to empty to mark dependency as "not required".
+                inverted_marker_dep = deps[0].with_constraint(EmptyConstraint())
+                inverted_marker_dep.marker = inverted_marker
+                deps.append(inverted_marker_dep)
+        return [dep for deps in by_name.values() for dep in deps]
+
     def _resolve_overlapping_markers(
         self,
         package: Package,
@@ -878,7 +916,20 @@ class Provider:
         dependencies = self._merge_dependencies_by_constraint(dependencies)
 
         new_dependencies = []
+
+        # We have can sort out the implicit "not required" dependency determined
+        # in _add_implicit_dependencies, because it does not overlap with
+        # any other dependency for sure.
+        for i, dep in enumerate(dependencies):
+            if dep.constraint.is_empty():
+                new_dependencies.append(dependencies.pop(i))
+                break
+
         for uses in itertools.product([True, False], repeat=len(dependencies)):
+            if not any(uses):
+                # handled by the implicit "not required" dependency
+                continue
+
             # intersection of markers
             # For performance optimization, we don't just intersect all markers at once,
             # but intersect them one after the other to get empty markers early.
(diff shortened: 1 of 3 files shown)


jeertmans:
Thanks for linking this to #8670 @radoering :-)

Maybe we should rewrite Poetry in Rust if speed is an issue ^^'

Jokes aside, having a resolving time this long is really an issue..

src/poetry/puzzle/provider.py
```diff
@@ -856,6 +858,42 @@ def _is_relevant_marker(
...
+        for _name, deps in by_name.items():
+            marker = marker_union(*[d.marker for d in deps])
+            if marker.is_any():
+                continue
+            inverted_marker = marker.invert()
+            if self._is_relevant_marker(inverted_marker, active_extras):
+                # Set constraint to empty to mark dependency as "not required".
+                inverted_marker_dep = deps[0].with_constraint(EmptyConstraint())
+                inverted_marker_dep.marker = inverted_marker
+                deps.append(inverted_marker_dep)
```
jorenham, line 894: These loops could be merged if 1) you use `itertools.groupby`, with e.g. `operators.attrgetter('name')` as key function, and 2) turn this method into a generator (e.g. with a `yield from` in the first `if` statement, and `yield` in the second).
This way you can avoid creating temporary lists altogether, for a significant speedup.

src/poetry/puzzle/provider.py
```diff
@@ -856,6 +858,42 @@ def _is_relevant_marker(
...
+        constraint, see https://github.com/python-poetry/poetry/issues/5506
+        for details.
+        """
+        by_name: dict[str, list[Dependency]] = defaultdict(list)
+        for dep in dependencies:
+            by_name[dep.name].append(dep)
+        for _name, deps in by_name.items():
+            marker = marker_union(*[d.marker for d in deps])
+            if marker.is_any():
+                continue
```
jorenham, line 888: Is `marker_union` also needed when e.g. `len(deps) == 1`? 
Because, at a glance, `marker_union` looks like a rather expensive function call.

src/poetry/puzzle/provider.py
```diff
@@ -570,6 +570,9 @@ def complete_package(
                         continue
                     self.search_for_direct_origin_dependency(dep)
 
+        active_extras = None if package.is_root() else dependency.extras
+        _dependencies = self._add_implicit_dependencies(_dependencies, active_extras)
```
jorenham, line 574: Since `_dependencies` is only used once, it's probably better to skip the variable assignment, by inlining in into the `_add_implicit_dependencies` call

src/poetry/puzzle/provider.py
```diff
@@ -878,7 +916,20 @@ def _resolve_overlapping_markers(
         dependencies = self._merge_dependencies_by_constraint(dependencies)
 
         new_dependencies = []
+
+        # We have can sort out the implicit "not required" dependency determined
+        # in _add_implicit_dependencies, because it does not overlap with
+        # any other dependency for sure.
+        for i, dep in enumerate(dependencies):
+            if dep.constraint.is_empty():
+                new_dependencies.append(dependencies.pop(i))
```
jorenham, line 925: The `list.pop` method can be very slow operation, and I think that it can be avoided here, by using a "blacklist" approach, e.g.

```python
blacklist = set()
for dep in dependencies:
    if dep.constraint.is_empty():
        blacklist.add(dep)
        break
```

Then later on in `itertools.product` use 
`repeat=len(dependencies) - len(blacklist)`.
And when looping over `dep in dependencies` again, simply skip it if `dep in blacklist`.

This avoids the `list.pop` operation, which has a time-complexity `O(n)`, by relying on `set.__contains__`, which is only `O(1)`.

tests/puzzle/test_solver.py
```diff
@@ -2101,6 +2101,54 @@ def test_duplicate_path_dependencies_same_path(
...
 
 
+@pytest.mark.parametrize(
+    "marker1, marker2",
+    [
+        ("python_version < '3.7'", "python_version >= '3.7'"),
+        ("sys_platform == 'linux'", "sys_platform != 'linux'"),
+        (
+            "python_version < '3.7' and sys_platform == 'linux'",
+            "python_version >= '3.7' and sys_platform == 'linux'",
```
jorenham, line 2111: I don't think `python<3.7` is relevant anymore

jorenham (commented):
I left some notes on potential performance improvements; perhaps it could help speed things up :)

dimbleby:
> I left some notes on potential performance improvements; perhaps it could help speed things up :)

I think it is likely that you are micro-optimizing essentially irrelevant parts of the code.  If you want to make performance improvements - recommend that the first thing to do is to profile, so that you spend your time optimizing the right things

But perhaps I am wrong, and you are now seeing results much better than those in the comment at the top of the thread?  If so - submit a merge request!

jorenham:
> > I left some notes on potential performance improvements; perhaps it could help speed things up :)
> 
> I think it is likely that you are micro-optimizing essentially irrelevant parts of the code. 

I don't agree that improvements to the runtime complexity are the same as "micro-optimizing". 

Plus, my suggestions will also result in fewer lines of code, without harming readability. So even if the performance benefits are minimal, at the very least there are no disadvantages.

radoering:
Superseded by #10944.

(Adding overrides on demand instead of eagerly makes much more sense from a performance perspective.)

Issue #2435, opened 2020-05-17
poetry run script doesn't run the actual executable script


k4ml:
- [x] I am on the [latest](https://github.com/python-poetry/poetry/releases/latest) Poetry version.
- [x] I have searched the [issues](https://github.com/python-poetry/poetry/issues) of this repo and believe that this is not a duplicate.
- [x] If an exception occurs when executing a command, I executed it again in debug mode (`-vvv` option).

- **OS version and name**: Linux x1 5.5.13-1-MANJARO #1 SMP PREEMPT Wed Mar 25 17:14:28 UTC 2020 x86_64 GNU/Linux
- **Poetry version**: 1.0.5
- **Link of a [pyproject.toml](https://github.com/k4ml/cutepony/blob/master/pyproject.toml)

## Issue
I'm trying to run django management command not through `python manage.py command` as usual, so that I can have single entry-point to my django app. I prefer to write my cli using Click instead of django management command framework which having to many boilerplate. This basically works perfectly fine. Example of the script is https://github.com/k4ml/cutepony/blob/master/cutepony/cli/cutepony.py. I can run any command with the following:-

```
poetry run cutepony manage <any_django_command>
```

The problem is with runserver command when running with autoreload (the default). The way autoreload is implemented is that it will re-execute the command using subprocess and to do that, it will re-collect the original command, using sys.executable and sys.argv. But using `poetry run` above, the resulting command that django will see is like:-

```
['/home/kamal/.cache/pypoetry/virtualenvs/cutepony-U1m74Vcq-py3.8/bin/python', 'cutepony', 'manage', 'runserver']
```
and will fail with the following error:-

```
/home/kamal/.cache/pypoetry/virtualenvs/cutepony-U1m74Vcq-py3.8/bin/python: can't find '__main__' module in 'cutepony'
```
This is because it try to run `cutepony` as python module. The correct command that django should see is:-

```
['/home/kamal/.cache/pypoetry/virtualenvs/cutepony-U1m74Vcq-py3.8/bin/python', '/home/kamal/.cache/pypoetry/virtualenvs/cutepony-U1m74Vcq-py3.8/bin/cutepony', 'manage', 'runserver']
```
The fix for this seem quite simple, in `poetry/console/commands/run.py:RunCommand.run_script()`, we can fix args[0] to be an absolute path to the script executable.

wagnerluis1982:
This is a duplicate of #965

PR #6971, opened 2022-11-04
Make run_script args[0] the absolute path to the script being called

logan-bobo:
# Pull Request Check List

Resolves: #2435

- [] Added **tests** for changed code. 
- [x] Updated **documentation** for changed code <- This would not produce a change in the documentation as it is not altering the core way the `run` command functions only deal with edge cases of its use. 

When a poetry user wants to run tools with auto-reload such as the kind found in Django's `runserver` through the `poetry run` functionality the auto-reload will collect the original command again in our case it works as `poetry run script` will look up in the config where the script is to be run from however when called again will produce an error as found in #2435. This PR aims to resolve that issue.

### DIFF
--- a/src/poetry/console/commands/run.py
+++ b/src/poetry/console/commands/run.py
@@ -47,10 +47,7 @@ class RunCommand(EnvCommand):
 
     def run_script(self, script: str | dict[str, str], args: list[str]) -> int:
         env = EnvManager(self.poetry).get()
-        if env.platform == "win32":
-            args[0] = rf"{env.path}\Scripts\{args[0]}"
-        else:
-            args[0] = f"{env.path}/bin/{args[0]}"
+        args[0] = rf"{env.script_dirs[0]}/{args[0]}"
 
         if isinstance(script, dict):
             script = script["callable"]
--- a/src/poetry/console/commands/run.py
+++ b/src/poetry/console/commands/run.py
@@ -5,6 +5,7 @@ from typing import TYPE_CHECKING
 from cleo.helpers import argument
 
 from poetry.console.commands.env_command import EnvCommand
+from poetry.utils.env import EnvManager
 
 
 if TYPE_CHECKING:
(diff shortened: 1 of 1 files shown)


src/poetry/console/commands/run.py
```diff
@@ -44,7 +45,10 @@ def _module(self) -> Module:
 
         return module
 
-    def run_script(self, script: str | dict[str, str], args: str) -> int:
+    def run_script(self, script: str | dict[str, str], args: list[str]) -> int:
+        env = EnvManager(self.poetry).get()
+        args[0] = f"{env.path}/bin/{args[0]}"
```
Secrus, line 50: I am 90% sure that the path will be different for windows. IIRC Windows uses `Scripts` directory in venvs (please verify it).

radoering: This looks like we should add a test that fails if the path is wrong. (In general, we should always try to add a test for a bugfix.)

Further, it might be better to use `Env.script_dirs` than building the path manually. However, I'm not sure which directory to choose if there is more than one.

logan-bobo: I will take a look in to `env.script_dirs` and also look to add tests to this PR thanks for the feedback 😅

logan-bobo: @radoering it looks as if changing to `env.script_dirs[0]` works on Linux but not on windows so I am going to leave the paths as `{env.path}\Scripts\{args[0]}` for Windows and `{env.path}/bin/{args[0]}` for Linux and MacOS as we know this is where the script will exist defined in `tool.poetry.scripts`

radoering: What's the issue with `script_dirs` on Windows?

What I'm actually wondering is if we have to check which folder is the correct one if there are multiple entries in `script_dirs` or can we be sure that it's `script_dirs[0]` (or maybe `script_dirs[-1]`)? It might be relevant if `env` is a `VirtualEnv` or a `SystemEnv`.

radoering: According to https://github.com/python-poetry/poetry/blob/ba97fea031aecd5fd31878e15ece24b3ddc55752/src/poetry/masonry/builders/editable.py#L164-L172 scripts are written to the first writable folder in `script_dirs`. Thus, we really should use `script_dirs` and have to check which is the correct folder.

PabloEmidio (approved):

radoering:
Closing this one in favor of #6737 because #6737 has been created before this one and progressed further.

PR #6977, opened 2022-11-05
docs: add advice for maintaining a plugin

radoering:


docs/plugins.md
```diff
@@ -249,3 +249,18 @@ If you want to uninstall a plugin, you can run:
 ```shell
 $POETRY_HOME/bin/pip uninstall poetry-plugin
 ```
+
+
+## Maintaining a plugin
+
+When writing a plugin, you will probably access internals of Poetry, since there is no
+official stable API. Although we try our best to deprecate methods first, before
```
Secrus, line 257: "stable *public* API" maybe? and I would add that there is no official plugin API *yet*.

radoering: Changed it to "stable public API". Doesn't that include plugin API?

Secrus (approved):

Issue #6893, opened 2022-10-26
poetry env info has wrong virtualenv info


adammarples:
- **Poetry version**: (version 1.2.2)
- **Python version**: (version 3.9.11)
- **OS version and name**: macOS 13.0
- **pyproject.toml**: https://gist.github.com/adammarples/256b00800dbecacccde7ec8700807c94

- [x] I am on the [latest](https://github.com/python-poetry/poetry/releases/latest) stable Poetry version, installed using a recommended method.
- [x] I have searched the [issues](https://github.com/python-poetry/poetry/issues) of this repo and believe that this is not a duplicate.
- [x] I have consulted the [FAQ](https://python-poetry.org/docs/faq/) and [blog](https://python-poetry.org/blog/) for any relevant entries or release notes.
- [x] If an exception occurs when executing a command, I executed it again in debug mode (`-vvv` option) and have included the output below.

## Issue
Creating a fresh poetry environment, the info returned by `poetry env info` does not match the virtualenv created

1. create a fresh dir
```bash
mkdir sample && cd sample
```

2. check that my active python is as I want (3.10.3 installed using asdf)
```bash
> ~/sample $ python --version
Python 3.10.3
> ~/sample $ which python
/Users/adammarples/.asdf/installs/python/3.10.3/bin/python
```
2.5 make sure prefer-active-python is set to true
```bash
> ~/sample $ poetry config virtualenvs.prefer-active-python true
```
3. poetry init, and manually changed suggested python compatible version from 3.9 to 3.10
```bash
> ~/sample $ poetry init
Compatible Python versions [^3.9]:  ^3.10
```

4. poetry env info, no path so far, but declares that the Virtualenv is Python version 3.9.11
```bash
> ~/sample $ poetry env info
Virtualenv
Python:         3.9.11
Implementation: CPython
Path:           NA
Executable:     NA
```

5. poetry update seems to create the correct virtualenv (3.10)
```bash
> ~/sample $ poetry update
Creating virtualenv sample-vPyTu8GS-py3.10 in /Users/xxxx/Library/Caches/pypoetry/virtualenvs
> ~/sample $ poetry run python --version
Python 3.10.3
> ~/sample $  poetry run which python
/Users/adammarples/Library/Caches/pypoetry/virtualenvs/sample-vPyTu8GS-py3.10/bin/python
```

6. poetry env info, still no path, but declares that the Virtualenv is Python version 3.9.11
```bash
> ~/sample $ poetry env info
[...]

Issue #5947, opened 2022-07-03
`poetry env info` does not respect `virtualenvs.prefer-active-python = true` setting


nirvana-msu:
- [x] I am on the [latest](https://github.com/python-poetry/poetry/releases/latest) Poetry version.
- [x] I have searched the [issues](https://github.com/python-poetry/poetry/issues) of this repo and believe that this is not a duplicate.
- [x] If an exception occurs when executing a command, I executed it again in debug mode (`-vvv` option).

- **OS version and name**: WSL, Ubuntu 20.04
- **Poetry version**: 1.2.0b1

## Issue
I'm relying on `pyenv` to specify which base Python version Poetry should be using. It was incredibly confusing, coming from an old version of Poetry, to find out that there's now  a `virtualenvs.prefer-active-python` setting which defaults to `false` - so has to be set to `true`  to use Poetry with `pyenv`. While that fixes e.g. `poetry install`,  `poetry env info` still ignores `pyenv` even with this setting. To get it to provide information about my actual venv I had to wrap the call with `poetry run` (for it to pick up pyenv python correctly), i.e. `poetry run poetry env info --path`.

As a side note, this new `prefer-active-python` setting defauling to `false` is such an incredible gotcha. I would think the vast majority of (linux) users would use Poetry alongside `pyenv` (and everyone was using `get-poetry.py` installer in the past as it was the recommended one), and this will get them all.

finswimmer:
Hello @nirvana-msu,

can you please give an example about the output you see and the one you would expect?


fin swimmer

matbun:
Try installing poetry from the `get-poetry.py` installer.

1. Clear poetry with 

```bash
curl -sSL https://install.python-poetry.org | python - --uninstall
rm -rf .cache/pypoetry
```

2. Re-install it with the "deprecated" script
```bash
curl -sSL https://raw.githubusercontent.com/python-poetry/poetry/master/get-poetry.py | python -
```

and let us know if something changes

jacopok:
@finswimmer I have encountered the same issue, so I can give an example of unexpected behavior.

**Setup**

My system python is 3.9.11.
I take the following `pyproject.toml`, trying to use python 3.7 for my project:
```toml
[tool.poetry]
name = "example_project"
version = "0.0.1"
description = "Demonstrating an issue."
license = "GNU GPL3"
authors = ["Jacopo Tissino <jacopo@tissino.it>"]

[tool.poetry.dependencies]
[...]

PR #6986, opened 2022-11-06
fix: take virtualenvs.prefer-active-python into account on EnvManager.get()

finswimmer:
Until now `virtualenvs.prefer-active-python` wasn't respected by the `EnvManager` when it tries to get the venv. This is fixed by the PR.

I took the chance to refactor `EnvManager` in that way, that `IO` is now set during initializing a new object. Thus it is not necessary to pass it in into the several methods.

# Pull Request Check List

Resolves: #6893
Resolves: #5947

### DIFF
--- a/src/poetry/utils/env.py
+++ b/src/poetry/utils/env.py
@@ -20,6 +20,7 @@ from pathlib import Path
 from subprocess import CalledProcessError
 from typing import TYPE_CHECKING
 from typing import Any
+from typing import cast
 
 import packaging.tags
 import tomlkit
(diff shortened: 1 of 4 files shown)


src/poetry/utils/env.py
```diff
@@ -533,26 +535,48 @@ def _full_python_path(self, python: str) -> str:
...
+                        list_to_shell_command(
+                            [executable, "-c", GET_PYTHON_VERSION_ONELINER]
+                        ),
+                        shell=True,
+                    ).strip()
+                )
+
+                version_info = [int(v) for v in python_patch.split(".")[:3]]
+
+        return version_info
```
mkniewallner, line 577:
```suggestion
    def _get_python_version(self) -> tuple[int, int, int]:
        if self._poetry.config.get("virtualenvs.prefer-active-python"):
            executable = self._detect_active_python()

            if executable:
                python_patch = decode(
                    subprocess.check_output(
                        list_to_shell_command(
                            [executable, "-c", GET_PYTHON_VERSION_ONELINER]
                        ),
                        shell=True,
                    ).strip()
                )

                return tuple(int(v) for v in python_patch.split(".")[:3])

        return sys.version_info[:3]
```
Using a tuple would more clearly indicate that we expect 3 items.

finswimmer: I implemented a version that returns a tuple. `mypy` was a little bit in the way here ...

neersighted (approved):
I'd rather we had one source of truth for this/it feels like the abstraction is leaking even more, but this is a fix for a real issue and it doesn't make the amount of work needed to refactor this file greater.

Issue #6994, opened 2022-11-08
Permission denied: '/usr/src/sqlalchemy'


auxsvr:
- **Poetry version**: 1.2.2
- **Python version**: 3.10.5
- **OS version and name**: openSUSE Tumbleweed
- **pyproject.toml**: 
```
[tool.poetry]
name = "test"
version = "0.1.0"
description = "test"
authors = ["test"]
readme = "README.md"

[tool.poetry.dependencies]
SQLAlchemy = {git = "https://github.com/sqlalchemy/sqlalchemy.git", branch = "main"}

[build-system]
requires = ["setuptools", "poetry-core>=1.0.0"]
build-backend = "poetry.core.masonry.api"
```

- [x] I am on the [latest](https://github.com/python-poetry/poetry/releases/latest) stable Poetry version, installed using a recommended method.
- [ ] I have searched the [issues](https://github.com/python-poetry/poetry/issues) of this repo and believe that this is not a duplicate.
- [x] I have consulted the [FAQ](https://python-poetry.org/docs/faq/) and [blog](https://python-poetry.org/blog/) for any relevant entries or release notes.
- [x] If an exception occurs when executing a command, I executed it again in debug mode (`-vvv` option) and have included the output below.

This looks like a duplicate of https://github.com/python-poetry/poetry/issues/6941 for the runtime case, but I was told to open a new issue.

## Issue
Simply run `poetry lock` with the `pyproject.toml` above. This used to work before upgrading to 1.2.2. The error message is:

> [Errno 13] Permission denied: '/usr/src/sqlalchemy'
<details>
<summary>The backtrace:</summary>

```
poetry lock -vvv
Loading configuration file /home/petros/.config/pypoetry/config.toml
Skipping virtualenv creation, as specified in config file.
Updating dependencies
Resolving dependencies...
   1: fact: test is 0.1.0
   1: derived: test
   1: Version solving took 0.003 seconds.
   1: Tried 1 solutions.

  Stack trace:

  22  /usr/lib/python3.10/site-packages/cleo/application.py:329 in run
       327│ 
       328│             try:
    → 329│                 exit_code = self._run(io)
       330│             except Exception as e:
       331│                 if not self._catch_exceptions:

  21  /usr/lib/python3.10/site-packages/poetry/console/application.py:185 in _run
       183│         self._load_plugins(io)
       184│ 
    → 185│         exit_code: int = super()._run(io)
       186│         return exit_code
       187│ 

[...]

PR #7001, opened 2022-11-10
Improve support without virtualenv

auxsvr:
# Pull Request Check List

Resolves: #6994

The change looks trivial, so I only did a couple of tests, with virtualenv and without. Beware, I'm not familiar with the codebase!

### DIFF
--- a/src/poetry/installation/executor.py
+++ b/src/poetry/installation/executor.py
@@ -27,6 +27,7 @@ from poetry.installation.operations import Update
 from poetry.utils._compat import decode
 from poetry.utils.authenticator import Authenticator
 from poetry.utils.env import EnvCommandError
+from poetry.utils.helpers import get_src_dir
 from poetry.utils.helpers import pluralize
 from poetry.utils.helpers import remove_directory
 from poetry.utils.pip import pip_install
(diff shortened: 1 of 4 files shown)


dimbleby:
I make no comment on the wisdom or correctness of this change - to be honest I'm suspicious about it, but working without a virtual environment very much doesn't interest me so I'll leave that to others.

However the code is kinda icky: instead of horrid `isinstance()` checks, you should implement a method on the `Env`, with an override on the `SystemEnv` or whatever as needed

auxsvr:
As `isinstance` determines if `SystemEnv` is used, are you suggesting that the method return true if bound to `SystemEnv`, otherwise false? Why should we encode this piece of information in two different places in the class definitions? Using `isinstance` may be frowned upon, but the only alternative in my humble opinion in this case seems to be `functools.singledispatch`, which requires major refactoring or redesign of the code.

neersighted:
The idea is to add a new `@property` on the `Env` classes, instead of trying to `isinstance()` them later, you just call `envinstance.src_path`.

abn:
Stale issue. If there is a real need for this, please re-create a new PR.

PR #7012, opened 2022-11-11
docs: update Windows cache path to match 1.2

neersighted:


Leon0402:
Thank you! Can confirm this is correct for my machine.

radoering (approved):

Issue #6499, opened 2022-09-13
Git dependencies' submodules with relative URLs handled incorrectly (regression from 1.1)


adamgreig:
- [x] I am on the [latest](https://github.com/python-poetry/poetry/releases/latest) Poetry version.
- [x] I have searched the [issues](https://github.com/python-poetry/poetry/issues) of this repo and believe that this is not a duplicate.
- [x] If an exception occurs when executing a command, I executed it again in debug mode (`-vvv` option).

- **OS version and name**: Ubuntu 20.04
- **Poetry version**: 1.2.0

## Issue

I have a project with a git dependency. The dependency uses git submodules, and the dependency's `.gitmodules` file contains a submodule using a relative URL, e.g.:

```
[submodule "foo/bar"]
    path = foo/bar
    url = ../bar.git
```

When Poetry attempts to install the dependency, it clones the repository and then attempts to clone each submodule, but it passes the raw relative URL from the configuration to `_clone()` and thus `_fetch_remote_refs()`, which calls Dulwich's `get_transport_and_path()`, which decides the URL is for a local repository so it attempts to find `../bar.git` on the filesystem, which fails. I think it should instead detect relative URLs and append them to the root repository's URL so that Dulwich will realise it's a remote repo and return the appropriate client.

If I clone the dependency using git on the command line, with `--recurse-submodules`, git correctly uses a relative URL to fetch the submodule. If I force Poetry to use the legacy system git instead of Dulwich, it doesn't seem to clone the submodules at all, so also doesn't run into trouble (in this case the dependency's submodule isn't needed to install the dependency, just for testing). Similarly this also worked OK on poetry 1.1 which I think used the system git to recursively clone submodules.

It looks like this is from #5428.

Here's a representative traceback:

<details><summary>Traceback</summary>
<pre>
$ poetry lock -vvv
Using virtualenv: /home/x/.cache/pypoetry/virtualenvs/baz-Nz-qS6ut-py3.8
Updating dependencies
Resolving dependencies...
   1: fact: baz is 0.1.0
   1: derived: baz
Cloning ssh://git@github.com/x/foo.git at 'master' to /home/x/.cache/pypoetry/virtualenvs/baz-Nz-qS6ut-py3.8/src/foo
   1: Version solving took 1.496 seconds.
   1: Tried 1 solutions.

  Stack trace:

  28  ~/.local/lib/python3.8/site-packages/cleo/application.py:329 in run
       327│ 
       328│             try:
[...]

PR #7017, opened 2022-11-12
Correctly parse Git submodule URLs

evanrittenhouse:
# Pull Request Check List

Resolves: https://github.com/python-poetry/poetry/issues/6499

This PR adds support for correctly parsing relative submodule URLs.

### DIFF
--- a/src/poetry/vcs/git/backend.py
+++ b/src/poetry/vcs/git/backend.py
@@ -325,7 +325,7 @@ class Git:
         return local
 
     @classmethod
-    def _clone_submodules(cls, repo: Repo, main_url: str) -> None:
+    def _clone_submodules(cls, repo: Repo) -> None:
         """
         Helper method to identify configured submodules and clone them recursively.
         """
@@ -440,7 +440,7 @@ class Git:
         try:
             if not cls.is_using_legacy_client():
                 local = cls._clone(url=url, refspec=refspec, target=target)
-                cls._clone_submodules(repo=local, main_url=url)
+                cls._clone_submodules(repo=local)
                 return local
         except HTTPUnauthorized:
             # we do this here to handle http authenticated repositories as dulwich
--- a/src/poetry/vcs/git/backend.py
+++ b/src/poetry/vcs/git/backend.py
@@ -333,7 +333,7 @@ class Git:
         modules_config = repo_root.joinpath(".gitmodules")
 
         # A relative URL by definition starts with ../ or ./
-        relative_submodule_regex = re.compile(r"^(\.{1,2})/?")
+        relative_submodule_regex = re.compile(r"^\.{1,2}/")
 
         if modules_config.exists():
             config = ConfigFile.from_path(str(modules_config))
--- a/src/poetry/vcs/git/backend.py
+++ b/src/poetry/vcs/git/backend.py
@@ -324,12 +324,15 @@ class Git:
         return local
 
     @classmethod
-    def _clone_submodules(cls, repo: Repo) -> None:
+    def _clone_submodules(cls, repo: Repo, main_url: str) -> None:
         """
         Helper method to identify configured submodules and clone them recursively.
         """
         repo_root = Path(repo.path)
         modules_config = repo_root.joinpath(".gitmodules")
+
+        # A relative URL by definition starts with ../ or ./
+        relative_submodule_regex = re.compile("^(\\.\\./|\\./),?")
 
         if modules_config.exists():
             config = ConfigFile.from_path(str(modules_config))
@@ -337,9 +340,27 @@ class Git:
             url: bytes
             path: bytes
             submodules = parse_submodules(config)
+
             for path, url, name in submodules:
                 path_relative = Path(path.decode("utf-8"))
                 path_absolute = repo_root.joinpath(path_relative)
+
+                url_string = url.decode()
+                final_url = url_string
+                submodule_is_relative = bool(
+                    relative_submodule_regex.search(url_string)
+                )
+                if submodule_is_relative:
+                    # Find a root list of url sections to mutate according
+                    # to the relative url
+                    url_sections = [section + "/" for section in main_url.split("/")]
+                    directories_upward = url_string.count("../")
+                    # Walk up the main URL until we know where to insert
+                    # the submodule URL
+                    url_portion_remaining = "".join(url_sections[:-directories_upward])
+                    # Insert the submodule URL - the last segment of url_string.split()
+                    # is the path relative to the main URL
+                    final_url = url_portion_remaining + url_string.split("../")[-1]
 
                 source_root = path_absolute.parent
                 source_root.mkdir(parents=True, exist_ok=True)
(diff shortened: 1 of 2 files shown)


evanrittenhouse:
@neersighted FYI

src/poetry/vcs/git/backend.py
```diff
@@ -321,23 +321,44 @@ def _clone(cls, url: str, refspec: GitRefSpec, target: Path) -> Repo:
...
 
             url: bytes
             path: bytes
             submodules = parse_submodules(config)
+
             for path, url, name in submodules:
                 path_relative = Path(path.decode("utf-8"))
                 path_absolute = repo_root.joinpath(path_relative)
 
+                url_string = url.decode()
```
abn, line 345: Thinking out loud here. Can't we simplify all by just doing something like this?

```py
from urllib.parse import urljoin

url_string = urljoin(f"{Git.get_remote_url(repo)}/", url.decode("utf-8"))
```
What I am not sure of is if `./foo.git` is ever a valid submodule. I am assuming here that git allows `./` only to enable relative paths such as `./../foo.git`. Because the above, if used with `./foo.git` in a superproject `https://github.com/baz/bar.git` will result in `https://github.com/baz/bar.git/foo.git`.

neersighted: This is more along the lines of what I was thinking, and would handle more cases... Tests for `./` and `./..` would be good to have as well (I for one have used `./` in a real project).

abn: @neersighted I tried to get `./foo.git` working on a project but event the git client does not like it when trying to do `git submodule init`. Can you share an example?

neersighted: Ah, I just realized the example I remembered uses `/proc/self/cwd` -- in other words, Linux-specific hacks.

evanrittenhouse: Implemented these changes locally. @neersighted I've also raised a [PR](https://github.com/python-poetry/test-fixture-vcs-repository/pull/6) in the test fixtures repo for the `./../foo.git` testcase. I'll update this PR when that one goes through.

evanrittenhouse: @neersighted Bumping this

gbooth27:
any updates on this? I am running into this issue

src/poetry/vcs/git/backend.py
```diff
@@ -319,23 +320,35 @@ def _clone(cls, url: str, refspec: GitRefSpec, target: Path) -> Repo:
...
-    def _clone_submodules(cls, repo: Repo) -> None:
+    def _clone_submodules(cls, repo: Repo, main_url: str) -> None:
         """
         Helper method to identify configured submodules and clone them recursively.
         """
         repo_root = Path(repo.path)
         modules_config = repo_root.joinpath(".gitmodules")
 
+        # A relative URL by definition starts with ../ or ./
+        relative_submodule_regex = re.compile("^(\\.\\./|\\./),?")
```
jelmer, line 331: This seems unnecessary, since urljoin will already ignore the first argument if the second argument is already an absolute URL:

```
>>> from urllib.parse import urljoin
>>> urljoin('https://example.com/some/base/', '../bar')
'https://example.com/some/bar'
>>> urljoin('https://example.com/some/base/', 'https://voorbeeld.info/blah/')
'https://voorbeeld.info/blah/'
```

jelmer (changes requested):

jelmer:
> any updates on this? I am running into this issue

IIRC this is stalled on https://github.com/python-poetry/test-fixture-vcs-repository/pull/6

src/poetry/vcs/git/backend.py
evanrittenhouse: Resolving as I implemented this change and tests pass.

src/poetry/vcs/git/backend.py
```diff
@@ -331,16 +332,28 @@ def _clone_submodules(cls, repo: Repo) -> None:
...
             for path, url, name in submodules:
                 path_relative = Path(path.decode("utf-8"))
                 path_absolute = repo_root.joinpath(path_relative)
 
+                url_string = url.decode("utf-8")
+                final_url = url_string
+                submodule_is_relative = bool(
+                    relative_submodule_regex.search(url_string)
+                )
+                if submodule_is_relative:
```
neersighted, line 354: I think we ought to make this a one-liner instead of using a temporary var.

src/poetry/vcs/git/backend.py
```diff
@@ -331,16 +332,28 @@ def _clone_submodules(cls, repo: Repo) -> None:
...
             url: bytes
             path: bytes
             submodules = parse_submodules(config)
+
             for path, url, name in submodules:
                 path_relative = Path(path.decode("utf-8"))
                 path_absolute = repo_root.joinpath(path_relative)
 
+                url_string = url.decode("utf-8")
+                final_url = url_string
```
neersighted, line 350: If we're just going to clobber final_url anyway, it feels like we should just update url_string in place instead of introducing another variable.

src/poetry/vcs/git/backend.py
```diff
@@ -331,16 +332,28 @@ def _clone_submodules(cls, repo: Repo) -> None:
         repo_root = Path(repo.path)
         modules_config = repo_root.joinpath(".gitmodules")
 
+        # A relative URL by definition starts with ../ or ./
+        relative_submodule_regex = re.compile("^(\\.\\./|\\./),?")
```
dimbleby, line 336: whats the `,?` for?

The tests pass without it, so if it's doing something useful then you're missing a test.

A more direct translation of your words into regex would be `re.compile(r"^\.{1,2}/")`

src/poetry/vcs/git/backend.py
evanrittenhouse: Resolved in latest commit.

src/poetry/vcs/git/backend.py
evanrittenhouse: Resolved in latest commit.

src/poetry/vcs/git/backend.py
evanrittenhouse: Yup, good catch - thanks. Resolved in latest commit.

src/poetry/vcs/git/backend.py
```diff
@@ -331,16 +332,24 @@ def _clone_submodules(cls, repo: Repo) -> None:
         repo_root = Path(repo.path)
         modules_config = repo_root.joinpath(".gitmodules")
 
+        # A relative URL by definition starts with ../ or ./
+        relative_submodule_regex = re.compile(r"^(\.{1,2})/?")
```
dimbleby, line 336: why are we capturing the dots?  what is the question mark for?

To reiterate: ny previous suggestion was `r"^\.{1,2}/"`

evanrittenhouse: Sorry, misread it. Done.

itsthejoker:
Add me to the list of folks who are waiting on this; I've got a dependency I can install with pip but not poetry, which has kind of stalled me.

radoering (approved):

Issue #7006, opened 2022-11-11
Remove `NullLocker` and satisfy `mypy` 0.990


moonsikpark:
Mypy introduced python/mypy#12118, and it wails `empty-body` on functions with empty body like this. https://github.com/python-poetry/poetry/blob/b28339d8b76500025fadcd23c87debfc368b609c/src/poetry/packages/locker.py#L469-L471

It seems there are no references to `NullLocker`, so I am wondering whether we can remove `NullLocker` and bump the `mypy` version to 0.990. If that's OK, I'm ready to submit a PR.

abn:
This was originally used in the update command, which is not the case anymore.

Based on a quick scan of https://cs.github.com/?q=NullLocker+path%3A*.py it seems there is no downstream usage on GitHub that we are aware of that will break implicitly. 

So, I suspect we can drop this without any side-effects. I doubt there is much value in employing a deprecation cycle here. 

We should call it out in the change log though. There are tools out there that use poetry as a library.

PR #7025, opened 2022-11-14
Fix: Remove `NullLocker`

moonsikpark:
# Pull Request Check List

Resolves: #7006 

`mypy` introduced python/mypy#12118, and it fails with `empty-body` on functions with empty body like this `NullLocker` implementation.

https://github.com/python-poetry/poetry/blob/b28339d8b76500025fadcd23c87debfc368b609c/src/poetry/packages/locker.py#L469-L471

It seems there is no downstream usage of this class on GitHub that we are aware of that will break implicitly. This PR deletes `NullLocker` and bump the version of `mypy` to 0.990, which is the latest.

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -396,7 +396,7 @@ python-versions = "*"
 
 [[package]]
 name = "mypy"
-version = "0.982"
+version = "0.990"
 description = "Optional static typing for Python"
 category = "dev"
 optional = false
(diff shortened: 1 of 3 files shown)


abn:
@moonsikpark this would be easier to merge if you can squash the commits into two please?

1. `refactor: remove unused NullLocker`
2. `chore: update mypy >= 0.990`

moonsikpark:
@abn No problem.

Secrus (approved):

Issue #7024, opened 2022-11-13
VCS dependency with extras requires explicit branch (due to double clone)


neersighted:
- **Poetry version**: 1.2.2
- **Python version**: 3.10.8
- **OS version and name**: Arch Linux, Windows 10/11, macOS 13

- [x] I am on the [latest](https://github.com/python-poetry/poetry/releases/latest) stable Poetry version, installed using a recommended method.
- [x] I have searched the [issues](https://github.com/python-poetry/poetry/issues) of this repo and believe that this is not a duplicate.
- [x] I have consulted the [FAQ](https://python-poetry.org/docs/faq/) and [blog](https://python-poetry.org/blog/) for any relevant entries or release notes.
- [x] If an exception occurs when executing a command, I executed it again in debug mode (`-vvv` option) and have included the output below.

## Issue
When using a VCS dependency with extras and the Dulwich Git client, an explicit branch must be provided. The following pyproject.toml will fail to clone:

```toml
[tool.poetry]
name = "extra-second-clone"
version = "0.1.0"
description = ""
authors = ["Bjorn Neergaard <bjorn@neersighted.com>"]
readme = "README.md"
packages = [{include = "extra_second_clone"}]

[tool.poetry.dependencies]
python = "^3.10"
poetry-core = {git = "https://github.com/python-poetry/poetry-core", extras = ["demo"]}


[build-system]
requires = ["poetry-core"]
build-backend = "poetry.core.masonry.api"
```
```
$ poetry lock -vvv
Loading configuration file /home/neersighted/.config/pypoetry/config.toml
Loading configuration file /home/neersighted/.config/pypoetry/auth.toml
Trying to detect current active python executable as specified in the config.
Found: /home/neersighted/.local/share/pyenv/versions/3.11.0/bin/python
Virtualenv extra-second-clone-IgA_y9qo-py3.11 already exists.
Using virtualenv: /home/neersighted/.cache/pypoetry/virtualenvs/extra-second-clone-IgA_y9qo-py3.11
Project environment contains an empty path in sys_path, ignoring.
Updating dependencies
Resolving dependencies...
   1: fact: extra-second-clone is 0.1.0
   1: derived: extra-second-clone
[keyring.backend] Loading KWallet
[keyring.backend] Loading SecretService
[keyring.backend] Loading Windows
[keyring.backend] Loading chainer
[keyring.backend] Loading libsecret
[keyring.backend] Loading macOS
No suitable keyring backend found
[...]

PR #7028, opened 2022-11-14
fix: make git ref resolution less greedy for HEAD

abn:
This change ensures that when ref `HEAD` is specified, it is not incorrectly resolved to `refs/head/HEAD`. `HEAD` is not treated as a branch, and correctly resolved from the ref spec.

Resolves: #7024
Closes: #6874

### DIFF
--- a/tests/integration/test_utils_vcs_git.py
+++ b/tests/integration/test_utils_vcs_git.py
@@ -8,6 +8,7 @@ from hashlib import sha1
 from pathlib import Path
 from typing import TYPE_CHECKING
 from typing import Iterator
+from typing import TypedDict
 from urllib.parse import urlparse
 from urllib.parse import urlunparse
 
@@ -37,6 +38,15 @@ if TYPE_CHECKING:
 # these tests are integration as they rely on an external repository
 # see `source_url` fixture
 pytestmark = pytest.mark.integration
+
+
+class GitCloneKwargs(TypedDict):
+    name: str | None
+    branch: str | None
+    tag: str | None
+    revision: str | None
+    source_root: Path | None
+    clean: bool
 
 
 @pytest.fixture(autouse=True)
@@ -133,7 +143,11 @@ def test_git_local_info(
         assert info.revision == remote_refs.refs[remote_default_ref].decode("utf-8")
 
 
+@pytest.mark.parametrize(
+    "specification", [{}, {"revision": "HEAD"}, {"branch": "HEAD"}]
+)
 def test_git_clone_default_branch_head(
+    specification: GitCloneKwargs,
     source_url: str,
     remote_refs: FetchPackResult,
     remote_default_ref: bytes,
@@ -142,7 +156,7 @@ def test_git_clone_default_branch_head(
     spy = mocker.spy(Git, "_clone")
     spy_legacy = mocker.spy(Git, "_clone_legacy")
 
-    with Git.clone(url=source_url) as repo:
+    with Git.clone(url=source_url, **specification) as repo:
         assert remote_refs.refs[remote_default_ref] == repo.head()
 
     spy_legacy.assert_not_called()
--- a/src/poetry/vcs/git/backend.py
+++ b/src/poetry/vcs/git/backend.py
@@ -134,7 +134,9 @@ class GitRefSpec:
 
     @property
     def is_ref(self) -> bool:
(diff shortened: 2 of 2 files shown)


neersighted (approved):
This looks a bit better than the first attempt (and has tests), but I am a bit wary of it still. Do you have any knowledge of where the duplicate clone (with extras) comes from/how we can prevent it?

dimbleby:
We clone the package twice in spite of the cache on https://github.com/python-poetry/poetry/blob/25767bcb4022a4a75e9c0f7c0eae5c8b8a72e12c/src/poetry/puzzle/provider.py#L88-L96 because the first time round `rev` is `None` and the second time it is `"HEAD"`.

From that point of view the alternative fix that I tried in https://github.com/python-poetry/poetry/pull/6874#issue-1419899629 does better, but perhaps it's the wrong direction or has other disadvantages.

neersighted:
Right, that makes more sense. What do we think about introducing a new datatype here to avoid the overloaded strings and to encapsulate the `HEAD == None` logic? The tests look good; we just need to realize that `None` and `HEAD` are the same thing (and to avoid the entropy of a string conversion from `HEAD` causing the cache to miss and clone a second time).

abn:
@neersighted I kind of feel that adding a new datatype here might be overkill at present, but happy to be told otherwise. Alternatively, we could do  something like this.

<details>
<summary><b>cache-fix.patch</b></summary>
<p>

```diff
diff --git a/src/poetry/puzzle/provider.py b/src/poetry/puzzle/provider.py
index 4d2e2d30..97abfb8c 100644
--- a/src/poetry/puzzle/provider.py
+++ b/src/poetry/puzzle/provider.py
@@ -86,7 +86,7 @@ class Indicator(ProgressIndicator):  # type: ignore[misc]
 
 
 @functools.lru_cache(maxsize=None)
-def _get_package_from_git(
+def _get_package_from_git_cached(
     url: str,
     branch: str | None = None,
     tag: str | None = None,
@@ -118,6 +118,24 @@ def _get_package_from_git(
     return package
 
 
+def _get_package_from_git(
+    url: str,
+    branch: str | None = None,
+    tag: str | None = None,
+    rev: str | None = None,
+    subdirectory: str | None = None,
+    source_root: Path | None = None,
+) -> Package:
+    return _get_package_from_git_cached(
+            url,
+            branch if branch != "HEAD" else None,
+            tag,
+            rev if rev != "HEAD" else None,
+            subdirectory,
+            source_root,
+        )
+
+
 class Provider:
     UNSAFE_PACKAGES: set[str] = set()
 
```

</p>
</details>

Either way, I can raise another PR to deal with the double clone issue. Let's solve the functionality issue by merging this?

dimbleby:
cf https://github.com/python-poetry/poetry-plugin-export/issues/172 in which the `HEAD` reference is again escaping where it oughtn't.

I wonder whether `HEAD` simply isn't a good reference to use and this fix is just papering over that (and requiring more of the same elsewhere).  

Perhaps the alternative at https://github.com/python-poetry/poetry/pull/6874#issue-1419899629 is better after all - though I seem to recall there was more unit test noise to work through in that case.

Kuba314:
Is there any plan to merge / fix this?

tests/integration/test_utils_vcs_git.py
```diff
@@ -39,6 +40,15 @@
 pytestmark = pytest.mark.integration
 
 
+class GitCloneKwargs(TypedDict):
```
abn, line 43: Kind of feel this was over kill, but seems this was the only way to stop mypy complaining. Suggestions welcome. 

```console
tests/integration/test_utils_vcs_git.py:149: error: Argument 2 to "clone" of "Git" has incompatible type "**dict[str, str]"; expected "Path | None"  [arg-type]
tests/integration/test_utils_vcs_git.py:149: error: Argument 2 to "clone" of "Git" has incompatible type "**dict[str, str]"; expected "bool"  [arg-type]
Found 2 errors in 1 file (checked 330 source files)
```

abn:
@Secrus I have rebased this and fixed the mypy issue.

tests/integration/test_utils_vcs_git.py
Secrus: I like this. The more type checking we have in place, the more potential issues we can catch before they blow up.

Secrus (approved):
LGTM

Issue #3012, opened 2020-10-01
KeyRingError: Failed to create the collection: Prompt dismissed..


AbstractUmbra:
- [x] I am on the [latest](https://github.com/python-poetry/poetry/releases/latest) Poetry version.
- [x] I have searched the [issues](https://github.com/python-poetry/poetry/issues) of this repo and believe that this is not a duplicate.
- [x] If an exception occurs when executing a command, I executed it again in debug mode (`-vvv` option).

- **OS version and name**: Debian 10
- **Poetry version**: Poetry version 1.0.10
- **Link of a [Gist](https://gist.github.com/) with the contents of your pyproject.toml file**: [Link!](https://gist.github.com/AbstractUmbra/13915cd8f7d7235ab0e6940562afce2e)

## Issue

Hello!
Just starting out using poetry to build and publish a package - when trying to configure my PyPI token with `poetry config pypi-token.pypi <redacted token>` I get the following exception:

```py
Loading configuration file /home/penumbra/.config/pypoetry/config.toml

[KeyRingError]
Unable to store the password for poetry-repository-pypi in the key ring: Failed to create the collection: Prompt dismissed..

Traceback (most recent call last):
  File "/home/penumbra/.poetry/lib/poetry/_vendor/py3.8/clikit/console_application.py", line 131, in run
    status_code = command.handle(parsed_args, io)
  File "/home/penumbra/.poetry/lib/poetry/_vendor/py3.8/clikit/api/command/command.py", line 120, in handle
    status_code = self._do_handle(args, io)
  File "/home/penumbra/.poetry/lib/poetry/_vendor/py3.8/clikit/api/command/command.py", line 171, in _do_handle
    return getattr(handler, handler_method)(args, io, self)
  File "/home/penumbra/.poetry/lib/poetry/_vendor/py3.8/cleo/commands/command.py", line 92, in wrap_handle
    return self.handle()
  File "/home/penumbra/.poetry/lib/poetry/console/commands/config.py", line 216, in handle
    password_manager.set_pypi_token(m.group(2), token)
  File "/home/penumbra/.poetry/lib/poetry/utils/password_manager.py", line 139, in set_pypi_token
    self.keyring.set_password(name, "__token__", token)
  File "/home/penumbra/.poetry/lib/poetry/utils/password_manager.py", line 55, in set_password
    raise KeyRingError(
```

I have searched for this exception and found no other issues. similar. Any help is appreciated!

abn:
[...]

PR #7037, opened 2022-11-15
Fall back when keyring is locked

remram44:
# Pull Request Check List

Resolves: #1917

# Description

Fall back if keyring is locked.

This avoids Poetry crashing with KeyringLocked if no keyring is available or the user doesn't unlock it, in situation where no credentials are needed.

If credentials were needed and the keyring is locked, the user will get a permission error, which I believe is perfectly fine.

I think Poetry shouldn't request access to the keyring before the remote server has actually asked for credentials, but this can be done in a subsequent change. #1917 is affecting many users.


src/poetry/utils/password_manager.py
```diff
@@ -46,9 +46,14 @@ def get_credential(
             return default
 
         import keyring
+        import keyring.errors
```
neersighted, line 49: This is unnecessary.

dimbleby:
see #6612

src/poetry/utils/password_manager.py
```diff
@@ -46,9 +46,14 @@ def get_credential(
...
 
         for name in names:
-            credential = keyring.get_credential(name, username)
+            try:
+                credential = keyring.get_credential(name, username)
+            except keyring.errors.KeyringLocked:
+                self._is_available = False
+                break
```
neersighted, line 56: This flow-control doesn't make sense (we shouldn't abort trying all `name`s) -- instead, defaulting credential to `None` and logging make more sense.

neersighted (changes requested):
Needs tests (and mostly looks like a less correct #6612 -- @
MatthieuBizien, are you still working on that PR)?

We also do not accept PRs against the 1.2 branch -- this change must go into master to be accepted.

src/poetry/utils/password_manager.py
remram44: If the keyring is locked because the user canceled, asking them again and again is not very good UX. A single `poetry install` command might prompt the user a hundred times...

felix-ht:
i did some reading in the code and the root cause of this issue seems to be that any request made by poetry tries to get auth information from the keyring even if the config of the `PasswordManager` has no indication that the requets needs auth in the first place. This is also the reason why any operation that does any kind of requests fails, with the keyring errors.

The solution should be to only check the keyring if the config of the `PasswordManager` holds a entry but this entry has no password stored.

This will most likely need more changes than this PR.

src/poetry/utils/password_manager.py
```diff
@@ -46,9 +46,14 @@ def get_credential(
             return default
 
         import keyring
+        import keyring.errors
 
         for name in names:
-            credential = keyring.get_credential(name, username)
+            try:
+                credential = keyring.get_credential(name, username)
+            except keyring.errors.KeyringLocked:
```
felix-ht, line 54: This is missing the https://github.com/python-poetry/poetry/issues/3012 error which is common for headless servers.

remram44: I feel like no retry should happen, even with different service names, when getting a `KeyringLocked`, but I am not sure that behavior is appropriate for all the other `KeyringError` subclasses.

radoering:
Closing because #8227 has been merged.

PR #7038, opened 2022-11-15
make keyring fallback optional

### LINKED ISSUE #1917: Keyring errors during non-publishing operations

felix-ht:
Fixes #1917 - the root cause of this issue seems to be that any request made by poetry tries to get auth information from the keyring even if the config of the PasswordManager has no indication that the requets needs auth in the first place. This is also the reason why any operation that does any kind of requests fails, with the keyring errors.
This was caused by some code that was bypassing the normal keyring opreations, which i removed.

I updated the test so that they still pass using the normal poetry keyring repo. I also made the keyring protected so access to it is discouraged.

# Pull Request Check List

Resolves: #issue-number-here

### DIFF
--- a/src/poetry/console/commands/config.py
+++ b/src/poetry/console/commands/config.py
@@ -245,6 +245,8 @@ To remove a repository (repo is a short alias for repositories):
             if self.option("unset"):
                 if m.group(1) == "http-basic":
                     password_manager.delete_http_password(m.group(2))
+                if m.group(1) == "http-fallback":
+                    password_manager.delete_http_fallback(m.group(2))
                 elif m.group(1) == "pypi-token":
                     password_manager.delete_pypi_token(m.group(2))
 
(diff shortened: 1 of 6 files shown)


neersighted:
I'm not sure this approach is good; some indexes rate-limit users based on credentials (e.g. have lower rate limits for unauthenticated users). Always providing credentials, instead of waiting for an error and retrying, seems to be a better behavior as it generates fewer network requests.

felix-ht:
@neersighted now i see - the assumption is that it should use the keyring even if there is no config - in case there just happens to be a suitable config in the normal namespace of the keyring

felix-ht:
while this certainly makes sense to have wouldn’t this very specifc feature - legacy support for keyring entries not created by poetry not be the perefect candiate for a config flag?

felix-ht:
allright i changed this PR so that the fallback is now an option that can be configured. @neersighted would be great if you could take a look. 

(I would assume that this would require some more documentation before this can merged)

felix-ht:
don't thing that this will get merged, so coling the pr

Issue #7043, opened 2022-11-16
Improve `poetry init` validate


nahco314:
- [x] I have searched the [issues](https://github.com/python-poetry/poetry/issues) of this repo and believe that this is not a duplicate.
- [x] I have searched the [FAQ](https://python-poetry.org/docs/faq/) and general [documentation](https://python-poetry.org/docs/) and believe that my question is not already covered.

## Feature Request

It appears that `poetry init` does not validate the value to the (Python or dependency) version.
I think it should do an easy to perform format check so that simple mistakes can be noticed and corrected, and ask for re-entry if they are wrong.

Example:
![image](https://user-images.githubusercontent.com/73940286/202058387-93309b4b-efa9-4393-807e-0602a3e4de99.png)

nahco314:
I'm new to OSS and didn't know how to do it. Thank you!

PR #7044, opened 2022-11-16
Improve poetry init validate

nahco314:
# Pull Request Check List

Resolves: #7043

I did not find any mention of `poetry init` input validation in the existing docs, so I decided it was not worth mentioning.


src/poetry/console/commands/init.py
```diff
@@ -448,6 +453,15 @@ def _format_requirements(self, requirements: list[dict[str, str]]) -> Requiremen
 
         return requires
 
+    @staticmethod
+    def _validate_pep440_version(version: str) -> str:
+        try:
```
dimbleby, line 458: this try-except (also the other) is pointless.  If all you do in the `except` branch is re-raise the exception, you might as well simply not catch it in the first place.

nahco314: I thought it would be better to clarify that the function to parse will generate an error if the format is incorrect.
Would it be better to simply call the function and write the intent in a comment?

nahco314: fixed.

abn:
@nahco314 I know this has been open a while, but if you are still interested let us know. Might be preferable to raise a new PR as this one is fairly stale.

PR #7072, opened 2022-11-22
fix: consistently retry on error codes and respect Retry-After header

nejch:
Closes https://github.com/python-poetry/poetry/issues/7679.

This applies a common list of status codes to retry between `poetry publish` and `poetry install` . Previously, `500, 501, 502, 503` was used for publish and `502, 503, 504` for install.

Additionally, 02ffff88238a6d74c739349177c9147489fdb3af adds support for the `Retry-After` header to properly handle 429 responses (https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/429).

Note: this does not touch the subprocess `pip install` commands, I haven't checked how pip handles retries.

~~Edit: marking this as ready even though there are some failures as I can't reproduce them locally, in case anyone has ideas already. Will try to check once I can get to a Windows machine.~~

# Pull Request Check List

- [x] Updated **documentation** for changed code. (Not applicable from what I can see)


src/poetry/utils/constants.py
```diff
@@ -3,3 +3,8 @@
 
 # Timeout for HTTP requests using the requests library.
 REQUESTS_TIMEOUT = 15
+
+RETRY_AFTER_HEADER = "retry-after"
+
+# Server response codes to retry requests on.
+STATUS_FORCELIST = [429, 500, 501, 502, 503, 504]
```
nejch, line 10: Notes:
* I'm not sure about the 501, but it was already in the uploader code so kept it for compatibility
* Potentially we could add cloudflare 52x codes since it powers so much of the internet, even if it's not standard. See [Cloudflare docs](https://support.cloudflare.com/hc/en-us/articles/206973867-Status-code-metrics-in-Cloudflare-Site-Analytics#h_3VPW9Ekh0e580wUvfawvQv), [another example](https://github.com/python-gitlab/python-gitlab/blob/1020ce965ff0cd3bfc283d4f0ad40e41e4d1bcee/gitlab/client.py#L25)

src/poetry/utils/authenticator.py
```diff
@@ -259,21 +262,29 @@ def request(
...
         # this should never really be hit under any sane circumstance
         raise PoetryException("Failed HTTP {} request", method.upper())
 
+    def _get_backoff(self, response: requests.Response | None, attempt: int) -> float:
+        if response is not None:
+            retry_after = response.headers.get(RETRY_AFTER_HEADER, "")
+            if retry_after:
+                return float(retry_after)
+
+        return 0.5 * attempt
```
nejch, line 286: mypy wasn't too happy about my try/except and duck typing here, hence the if/else checks. Happy to change this though.

nejch:
Alright. I finally got to a Windows machine today and fixed the test. This should be ready for review.

/cc @novas0x2a @dimbleby you were active in the related issue, if you have any tips on moving forward here, just let me know ;) not sure if it this hasn't been reviewed because of the CI failure or more general maintainer shortage.

dimbleby:
if I knew how to persuade folk to hit the merge button I would gladly share that secret!

nejch:
> if I knew how to persuade folk to hit the merge button I would gladly share that secret!

No worries, people have their own lives and priorities, just wanted to check if some of you more regular contributors might have a flow that works well ;)

src/poetry/utils/authenticator.py
radoering: Should be fine as is.

nejch:
> Due to the limited time, I only take a closer look at a PR I'm not personally interested in if the PR looks good at first glance and/or seems to solve an issue that hits many users.
> 
> Some positive criteria:
> 
> * pipelines do not fail
> * test that fails without the change was added
> * change does not conflict with main branch
> * a regular contributor (such as dimbleby) reviewed/approved the PR
> * users claim they have tested the change and it fixes the issue for them
> 
> Even if you created the perfect PR, you can be unlucky if all maintainers are busy with other things at the time. In that case it might make sense to ask on our Discord server after a month or so.

Thanks for taking a look @radoering! Should be ready for another round now, I've left the cloudfare codes out so I can look mocking custom codes in httpretty separately.

As this is mostly dealing with transient errors, it might be a bit harder for users to reproduce the condition reliably, but if any users here would like to give it a shot let us know :bow:

radoering (approved):

Issue #6009, opened 2022-07-14
How can I modify REQUESTS_TIMEOUT?


LGDHuaOPER:
- [✓] I have searched the [issues](https://github.com/python-poetry/poetry/issues) of this repo and believe that this is not a duplicate.
- [✓] I have searched the [documentation](https://python-poetry.org/docs/) and believe that my question is not covered.

## Issue

The default value of REQUESTS_TIMEOUT is 15. How can I modify REQUESTS_TIMEOUT? Thanks.

Secrus:
We are trying to avoid adding too many configuration options. Could you please provide an example of when this could be useful?

LGDHuaOPER:
I'm certainly happy to provide PR. But then again, there are indeed cases where this configuration is useful.

For example, when I use docker to build a python program image based on the poetry v1.2.0b3 package manager and have configured the image source.

```toml
[[tool.poetry.source]]
name = "tsinghua"
url = "https://pypi.tuna.tsinghua.edu.cn/simple"
default = true

[[tool.poetry.source]]
name = "aliyun"
url = "https://mirrors.aliyun.com/pypi/simple"
secondary = true
```

Then there is the following timeout error.

```text
Installing collected packages: pip
Successfully installed pip-22.1.2
Removing intermediate container d5771fcc3a03
 ---> 25df640a8c73
Step 26/28 : RUN python3 -m pip install --user poetry==1.2.0b3
 ---> Running in de7717e00f0c
Collecting poetry==1.2.0b3
  WARNING: Retrying (Retry(total=4, connect=None, read=None, redirect=None, status=None)) after connection broken by 'ReadTimeoutError("HTTPSConnectionPool(host='files.pythonhosted.org', port=443): Read timed out. (read timeout=15)")': /packages/35/ed/a7d18cbff756519db3b73c7fe9a9955d8e51629c0ae0315a8977ba44aef0/poetry-1.2.0b3-py3-none-any.whl
  WARNING: Retrying (Retry(total=3, connect=None, read=None, redirect=None, status=None)) after connection broken by 'ReadTimeoutError("HTTPSConnectionPool(host='files.pythonhosted.org', port=443): Read timed out. (read timeout=15)")': /packages/35/ed/a7d18cbff756519db3b73c7fe9a9955d8e51629c0ae0315a8977ba44aef0/poetry-1.2.0b3-py3-none-any.whl
[...]

PR #7081, opened 2022-11-23
Add experimental POETRY_REQUESTS_TIMEOUT option

Secrus:
# Pull Request Check List

Resolves: #6009

Not sure if it should be tested, since it is only a stdlib function usage.

### DIFF
--- a/docs/faq.md
+++ b/docs/faq.md
@@ -223,3 +223,10 @@ RUN poetry install --no-dev
 
 The two key options we are using here are `--no-root` (skips installing the project source) and `--no-directory` (skips installing any local directory path dependencies, you can omit this if you don't have any).
 [More information on the options available for `poetry install`]({{< relref "cli#install" >}}).
+
+
+### My requests are timing out!
+
+Poetry's default HTTP request timeout is 15 seconds, the same as `pip`.
+Similar to `PIP_REQUESTS_TIMEOUT`, the **experimental** environment variable `POETRY_REQUESTS_TIMEOUT`
+can be set to alter this value.
--- a/src/poetry/utils/constants.py
+++ b/src/poetry/utils/constants.py
@@ -1,8 +1,10 @@
 from __future__ import annotations
 
+import os
+
 
 # Timeout for HTTP requests using the requests library.
-REQUESTS_TIMEOUT = 15
+REQUESTS_TIMEOUT = int(os.getenv("POETRY_REQUESTS_TIMEOUT", 15))
 
 RETRY_AFTER_HEADER = "retry-after"
 


mbbyn:
> Is there any appetite to merge this change? It would really help anyone trying to download larger packages that are being scanned by corporate proxies. Without being able to extend this timeout, downloading packages like pyspark or pytorch results in a failure.

We faced this issue for a while and eventually became too painful to ignore, so we set up a devpi server internally with docker (first we set up pypiserver, but devpi was better for mirroring pypi.org). You may or may not be able to do this, but do consider it even if deployed locally it has proven to be a worthwhile 2-day investment.

ralbertazzi:
Although I'm not facing this issue, it seems more than reasonable to me that users have a way to configure network timeouts. This PR looks a no brainer to me, we have real users waiting for it, and there's already been consensus on making this a quick temporary solution. But it's of no use if it's not _quick_ :) I would encourage a merge 🙏 @radoering @dimbleby

dimbleby:
(no point summoning me, I do not have the commit bit)

docs/faq.md
```diff
@@ -152,3 +152,16 @@ This is done so to be compliant with the broader Python ecosystem.
...
+
+### My requests are timing out!
+
+Poetry's default HTTP request time out is 15 seconds, the same as `pip`.
+Similar to `PIP_REQUESTS_TIMEOUT`, the experimental environment variable `POETRY_REQUESTS_TIMEOUT`
+can be set to alter this value. As this environment variable is **unstable**, it may be removed in the future, but
+not before being replaced with a permanent alternative.
+
+Note that it is almost always better to debug network stability than to blindly increase
+the request timeout -- the timeout controls how long it Poetry waits to establish a HTTP
```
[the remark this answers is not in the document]
radoering: I agree, we should be more accurate here. I'm not even sure if debugging network stability really helps. I'm not sure about this, but from observation I believe a read timeout can also occur when the server is overloaded. IMO we'd better omit this paragraph entirely if we don't know how to write it correctly.

radoering (changes requested):
> This PR looks a no brainer to me

It does, but there are at least two things to consider:

- Do we really want to introduce a config option that can only be configured via an environment variable and not via `poetry config`? (We have already discussed that we accept that in this case.)
- Documentation should be accurate. (see comments)

docs/faq.md
```diff
@@ -152,3 +152,16 @@ This is done so to be compliant with the broader Python ecosystem.
...
 For example, if Poetry builds a distribution for a project that uses a version that is not valid according to
 [PEP 440](https://peps.python.org/pep-0440), third party tools will be unable to parse the version correctly.
+
+
+### My requests are timing out!
+
+Poetry's default HTTP request time out is 15 seconds, the same as `pip`.
+Similar to `PIP_REQUESTS_TIMEOUT`, the experimental environment variable `POETRY_REQUESTS_TIMEOUT`
+can be set to alter this value. As this environment variable is **unstable**, it may be removed in the future, but
+not before being replaced with a permanent alternative.
```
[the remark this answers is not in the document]
Secrus: Dropped as suggested

docs/faq.md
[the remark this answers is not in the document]
Secrus: Deleted. Didn't know how to rephrase it better (or what to suggest instead).

docs/faq.md
```diff
@@ -223,3 +223,10 @@ RUN poetry install --no-dev
 
 The two key options we are using here are `--no-root` (skips installing the project source) and `--no-directory` (skips installing any local directory path dependencies, you can omit this if you don't have any).
 [More information on the options available for `poetry install`]({{< relref "cli#install" >}}).
+
+
+### My requests are timing out!
+
+Poetry's default HTTP request time out is 15 seconds, the same as `pip`.
```
radoering, line 230:
```suggestion
Poetry's default HTTP request timeout is 15 seconds, the same as `pip`.
```

Just noticed in the generated docs.

docs/faq.md
```diff
@@ -223,3 +223,10 @@ RUN poetry install --no-dev
...
+Similar to `PIP_REQUESTS_TIMEOUT`, the experimental environment variable `POETRY_REQUESTS_TIMEOUT`
+can be set to alter this value. This environment variable is **unstable**.
```
radoering, line 232:
```suggestion
can be set to alter this value.
```

The previous sentence already says _"experimental environment variable"_. You could emphasize **experimental** if you like. IMO, that's sufficient. The last sentence could also be understood to mean that setting the environment variable does not work reliably.

radoering (changes requested):

docs/faq.md
Secrus: Fixed

docs/faq.md
Secrus: Done

radoering (approved):

PR #7084, opened 2022-11-23
chore: added flake8-class-attributes-order

finswimmer:
No functional changes.

Added [flake8-class-attributes-order](https://pypi.org/project/flake8-class-attributes-order/) to pre-commit checks and move attributes according to errors reported.

The desired order  of class attributes is of course arguable. I configured it in the way I'm used to it and was already used in various places.

### DIFF
--- a/.flake8
+++ b/.flake8
@@ -8,6 +8,23 @@ inline-quotes = double
 enable-extensions = TC, TC1
 type-checking-strict = true
 eradicate-whitelist-extend = ^-.*;
+class_attributes_order =
+    constant,
+    field,
+    nested_class,
+    __new__,
+    __init__,
+    __post_init__,
+    private_property_method
+    property_method,
+    private_static_method,
+    static_method,
+    private_class_method,
+    class_method,
+    private_method,
+    method,
+    magic_method
+
 extend-ignore =
     # E203: Whitespace before ':' (pycqa/pycodestyle#373)
     E203,
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -35,6 +35,7 @@ repos:
           - flake8-annotations==2.9.0
           - flake8-broken-line==0.5.0
           - flake8-bugbear==22.7.1
+          - flake8-class-attributes-order==0.1.3
           - flake8-comprehensions==3.10.0
           - flake8-eradicate==1.3.0
           - flake8-quotes==3.3.1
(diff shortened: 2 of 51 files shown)


neersighted (commented):
I don't mind making things uniform, but the massive amount of code movement/churn makes me nervous given the large backlog of PRs. Certainly, navigation is not bad when using tooling/IDEs as it is, but this makes things better for non-IDE users and reduces bikeshedding. So overall I'm of two minds, and would like to hear from others.

dimbleby:
> would like to hear from others.

I considered commenting on this earlier but decided that I didn't care enough.  So while I'm going to be minus one on this, it's a weak minus one that I declare only because you asked.

My hesitation is mostly just that I feel this is more likely to be irritating than helpful - I expect to get very little value out of methods etc being in some particular order; I expect to be ever so slightly bothered by the linter nagging me about this thing that I just don't care about.

The point about annoying all the folk with pending merge requests is also valid.

Secrus:
I am not too fond of this too. With how we have our code organized right now, it adds a lot of merging noise and there is a very small gain margin since most (if not all) modern editors/IDEs support easy code navigation.

radoering:
I tend to say this brings more damage than good. Rebasing pending PRs will probably be a pain. Further, if private methods are made public or static methods are converted to class methods they have to be moved around, which is annoying when investigating the history or reviewing PRs especially if there is a change in the method and a necessary move (because of the pre-commit hook) in one commit.

finswimmer:
Hm, ok. Looks like I cannot get a majority for this change.

Yes, the need for rebasing a lot of open PR is a pain. But this always happens if one introduce/enforce a new style to the code. We had this in the past already, when we dropped the `dev` branch, split out code to `poetry-core`, introduce type hints to the code base. Surprisingly I can't remember that this was a big deal to contributors.

> if private methods are made public or static methods are converted to class methods they have to be moved around

This is what I would do with these method even without a linter which reminds me to do this :)

Whether it brings value to the code base or not, might be rated differently by each contributor. That's the typically discussion about the value of style guides at all. My point of view is, that code is read more often than it is written. So everything, that makes the way of how it is written more consistency, is a good thing.

I will leave this PR open for some time, if someone likes to change its mind. :D Otherwise I will close it in some days.

Secrus:
> We had this in the past already, when we dropped the dev branch, split out code to poetry-core, introduce type hints to the code base. Surprisingly I can't remember that this was a big deal to contributors.

Sure, but the proposed changes are much more bikeshedding than the ones you mention.

Issue #6972, opened 2022-11-04
1.3.0 Release


neersighted:
This issue tracks the progress of version 1.3.0 of Poetry. Please surface any blockers or concerns you have regarding this minor release here:

## Blockers

* Changelog
* Release notes (blog)
* poetry-plugin-export 1.2.0 (python-poetry/poetry-plugin-export#158)
* https://github.com/python-poetry/poetry/pull/6950

## Nice to have

* https://github.com/python-poetry/poetry/pull/6810
* ~~https://github.com/python-poetry/poetry-core/pull/517~~
* #6966
* #6824
* #6808 
* #6790 
* #6723 
* ~~#6709~~ 
* https://github.com/python-poetry/poetry-core/pull/512
* https://github.com/python-poetry/poetry-core/pull/487
* #6747
* ~~https://github.com/python-poetry/cleo/issues/248~~

## Important changes to surface

(This is intended to be the basis for the blog post and should cover new features or any manual steps/incompatibility when upgrading)

* The next release will turn off generation of `setup.py` by default (https://github.com/python-poetry/poetry-core/pull/318)

finswimmer:
Nice to have: https://github.com/python-poetry/poetry/pull/6810

dimbleby:
At time of writing, there are quite a lot of straightforward pull requests queued up.  I don't think any of them should block a release - I'm in favour of releasing more often! - but also I don't know any reason they shouldn't be merged and included.

Trying to err on the side of caution, I'd still suggest that the following look ready to go:

- #6966
- #6824
- #6808 
- #6790 
- #6723 
- #6709

dimbleby:
#6747 is maybe also desirable (needs a poetry-core release first)

neersighted:
Edited; I'll start taking a look at finishing/merging the low hanging fruit later today.

rickzinit:
any possibility to release v.1.3.0 this week?

neersighted:
As ever, soon/when we can.

neersighted:
Regarding #6709, without CI to validate it, and with other work needed to migrate Poetry to the (eminent) 1.0 release, I am inclined to defer it to reduce risk and my need to manually validate the change.

https://github.com/python-poetry/cleo/issues/248 would be nice, but we can bump Cleo in a patch release and @Secrus has limited bandwidth to get #6709 reviewed/Poetry rebased on top of the breaking changes in Cleo 1.0.

neersighted:
[...]

PR #7098, opened 2022-11-25
Discard merged any marker constraints

radoering:
Resolves an issue revealed in https://github.com/python-poetry/poetry/pull/6950#issuecomment-1327494436

Based on #4590, after adapting the marker of "any marker dependencies", these dependencies have to be discarded if the marker is empty, not compatible with the project's python constraint or (during install) not compatible with the environment.

### DIFF
--- a/src/poetry/puzzle/provider.py
+++ b/src/poetry/puzzle/provider.py
@@ -59,6 +59,18 @@ if TYPE_CHECKING:
 logger = logging.getLogger(__name__)
 
 
+class IncompatibleConstraintsError(Exception):
+    """
+    Exception when there are duplicate dependencies with incompatible constraints.
+    """
+
+    def __init__(self, package: Package, *dependencies: Dependency) -> None:
+        constraints = "\n".join(dep.to_pep_508() for dep in dependencies)
+        super().__init__(
+            f"Incompatible constraints in requirements of {package}:\n{constraints}"
+        )
+
+
 class Indicator(ProgressIndicator):
     CONTEXT: str | None = None
 
@@ -740,7 +752,7 @@ class Provider:
                 f"<warning>Different requirements found for {warnings}.</warning>"
             )
 
-            deps = self._handle_any_marker_dependencies(deps)
+            deps = self._handle_any_marker_dependencies(package, deps)
 
             overrides = []
             overrides_marker_intersection: BaseMarker = AnyMarker()
@@ -975,7 +987,7 @@ class Provider:
         return deps
 
     def _handle_any_marker_dependencies(
-        self, dependencies: list[Dependency]
+        self, package: Package, dependencies: list[Dependency]
     ) -> list[Dependency]:
         """
         We need to check if one of the duplicate dependencies
@@ -999,11 +1011,16 @@ class Provider:
         any_markers_dependencies = [d for d in dependencies if d.marker.is_any()]
         other_markers_dependencies = [d for d in dependencies if not d.marker.is_any()]
 
-        for dep_any in any_markers_dependencies:
+        if any_markers_dependencies:
             for dep_other in other_markers_dependencies:
-                dep_other.constraint = dep_other.constraint.intersect(
-                    dep_any.constraint
-                )
+                new_constraint = dep_other.constraint
+                for dep_any in any_markers_dependencies:
+                    new_constraint = new_constraint.intersect(dep_any.constraint)
+                if new_constraint.is_empty():
+                    raise IncompatibleConstraintsError(
+                        package, dep_other, *any_markers_dependencies
+                    )
+                dep_other.constraint = new_constraint
 
         marker = other_markers_dependencies[0].marker
         for other_dep in other_markers_dependencies[1:]:
--- a/src/poetry/puzzle/provider.py
+++ b/src/poetry/puzzle/provider.py
@@ -740,55 +740,7 @@ class Provider:
                 f"<warning>Different requirements found for {warnings}.</warning>"
             )
 
-            # We need to check if one of the duplicate dependencies
-            # has no markers. If there is one, we need to change its
-            # environment markers to the inverse of the union of the
-            # other dependencies markers.
-            # For instance, if we have the following dependencies:
-            #   - ipython
-            #   - ipython (1.2.4) ; implementation_name == "pypy"
-            #
-            # the marker for `ipython` will become `implementation_name != "pypy"`.
-            #
-            # Further, we have to merge the constraints of the requirements
-            # without markers into the constraints of the requirements with markers.
-            # for instance, if we have the following dependencies:
-            #   - foo (>= 1.2)
-            #   - foo (!= 1.2.1) ; python == 3.10
-            #
-            # the constraint for the second entry will become (!= 1.2.1, >= 1.2)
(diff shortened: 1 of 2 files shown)


neersighted (changes requested):

src/poetry/puzzle/provider.py
```diff
@@ -1021,3 +973,70 @@ def _merge_dependencies_by_marker(
...
+                #   - foo (2.0) ; python == 3.8
+                #   - bar (2.0) ; python == 3.8
+                #   - bar (3.0) ; python == 3.9
+                #
+                # the last dependency would be missed without this,
+                # because the intersection with both foo dependencies is empty
+                inverted_marker_dep = dependencies[0].with_constraint(EmptyConstraint())
+                inverted_marker_dep.marker = inverted_marker
+                dependencies.append(inverted_marker_dep)
+        else:
```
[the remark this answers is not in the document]
radoering: `any_markers_dependencies` and `other_markers_dependencies` is used unconditionally in line 1000 and that's always required so we can't move the computation. Intersecting the constraints is always necessary, no matter if the inverted marker is empty or not.

Further, we can't use a set because dependencies can be equal even if they have different markers (and here we care about markers).

dimbleby:
I have doubts about correctness, especially in cases where the requirements actually are incompatible.

For instance consider
```toml
requests = [
  { version = "=2.24.0" },
  { version = "=2.25.0", markers = "implementation_name == 'pypy'" }
]
```
the result of this is that 2.24.0 is chosen unconditionally, which one way or the other surely wasn't what the user intended.

It would certainly be simpler to make the user responsible for ensuring that their markers are non-overlapping, rather than trying to infer what they probably meant.

That is, perhaps the simpler approach is just to search for, and refuse to process, overlapping markers.

Alternatively poetry could search for overlapping markers and reject them if and only if they have contradictory constraints - but still do something like what this MR does when their constraints are compatible.  Feels like it might be getting a bit complicated though...

radoering:
> I have doubts about correctness, especially in cases where the requirements actually are incompatible.

Good catch. (This has already been an issue before this PR.) I just added a fix for that, too.

> It would certainly be simpler to make the user responsible for ensuring that their markers are non-overlapping, rather than trying to infer what they probably meant.

It would, but I'd rather support compatible overlapping markers if there isn't any PEP that says overlapping markers are forbidden.

> also the special-casing of "any" bothers me. That's not the only way to write overlapping markers.

That's right. At the moment we are just dealing with one more common case of overlapping markers. Or more exactly with two cases because `_merge_dependencies_by_marker` is handling the special case of completely overlapping markers. I already have a follow up PR in mind to deal with overlapping markers in general, i.e. generalizing `_handle_any_marker_dependencies` and removing `_merge_dependencies_by_marker` (and probably also some of the code in between the two calls).

src/poetry/puzzle/provider.py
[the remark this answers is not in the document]
neersighted: Okay, fair enough. I wonder if the equality situation here was a mistake (and we should include markers) -- but given we are deliberately rewriting markers, maybe not?

neersighted:
In general, I think we're committed to this formulation of markers, as to do otherwise would be a breaking change, and likely break existing projects. I'm reasonably sure our extras implementation relies on part of this as well. Overall, I think trying to refactor this to be cleaner and better abstracted/architected over time is the only way forward.

src/poetry/puzzle/provider.py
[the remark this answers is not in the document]
radoering: Difficult to say. Changing `Dependency.__eq__` to consider the `marker` property seems to break some downstream tests, but only because we don't care for markers in our expectations. Searching for usage of `Dependency` in sets or as key in dicts, I only found `Provider._deferred_cache` which might get a bit worse at caching if we consider markers. At a quick glance, I couldn't find any other obvious side effects but of course that doesn't mean it's safe. There can still be simple comparisons that rely on equality as is.

I don't think rewriting markers necessarily is a reason for not considering markers in equality. It's just a reason to not include it in the hash.

src/poetry/puzzle/provider.py
```diff
@@ -1021,3 +985,77 @@ def _merge_dependencies_by_marker(
...
+                #   - bar (2.0) ; python == 3.8
+                #   - bar (3.0) ; python == 3.9
+                #
+                # the last dependency would be missed without this,
+                # because the intersection with both foo dependencies is empty.
+                inverted_marker_dep = dependencies[0].with_constraint(EmptyConstraint())
+                inverted_marker_dep.marker = inverted_marker
+                dependencies.append(inverted_marker_dep)
+        else:
+            # Attention: Iterating over any_markers_dependencies and removing them
```
dimbleby, line 1056: doesn't this code have the same problem, then?

eg if the list is `[foo (with a marker that's not any), foo (with a marker that is any)]`, then this code will remove the first of those elements from the list, right?

perhaps there's some reason that can't happen?

dimbleby:
this one makes myhead hurt! perhaps it would be sensible to unblock #6950 and therefore the 1.3.0 release by rewriting the markers in that one so that they don't need this fix?

src/poetry/puzzle/provider.py
radoering: Good catch. The workaround is non-sense. Actually, this case probably never occurs because we merge dependencies by constraint earlier. Nevertheless, I think it wouldn't be correct to use `remove` here. Now, it should be correct.

radoering:
I was hoping that we were pretty far along with this review. Actually, this is just a small fix. Partial support for overlapping markers was not introduced here, but was already present before this PR. Anyway, IMO this doesn't really block the release as long as nobody has updated the changelog. 😉

radoering:
That's right. I just mean reworking #6950 is just a matter of minutes but even with that the changelog has to be updated and that's the real work. We might as well wait with reworking #6950 until that's the only thing the release is waiting for. (As you have probably already noticed, I'd prefer to keep #6950 as is which requires this PR.)

dimbleby:
> ... until that's the only thing the release is waiting for

So far as I can tell from https://github.com/python-poetry/poetry/issues/6972, #6950 _is_ the only thing the release is waiting for

src/poetry/puzzle/provider.py
```diff
@@ -1021,3 +985,76 @@ def _merge_dependencies_by_marker(
...
+                # the last dependency would be missed without this,
+                # because the intersection with both foo dependencies is empty.
+                inverted_marker_dep = dependencies[0].with_constraint(EmptyConstraint())
+                inverted_marker_dep.marker = inverted_marker
+                dependencies.append(inverted_marker_dep)
+        else:
+            # Attention: We can't use dependencies.remove() because markers are not
+            # considered in Dependency.__eq__() so that the wrong dependency
+            # might be removed.
+            dependencies = [dep for dep in dependencies if not dep.marker.is_any()]
```
dimbleby, line 1059: this is already `other_markers_dependencies` isn't it?

dimbleby:
left a note suggesting a little tidying, otherwise looks good to me

src/poetry/puzzle/provider.py
radoering: It seems I was too focussed on what's not working to see the simplest solution.

neersighted (approved):
LGTM!

Issue #7073, opened 2022-11-22
[Feature Request] `poetry init`  and `poetry new` could/should respect `virtualenvs.prefer-active-python` configuration option


david-vicente:
### Discussed in https://github.com/orgs/python-poetry/discussions/7059

<div type='discussions-op-text'>

<sup>Originally posted by **david-vicente** November 19, 2022</sup>
Consider the following situation: I'm using `pyenv` to manage my python versions and want to start a new project using `poetry`.

I thought that if I ran

``` shell
$ pyenv versions
* system
  3.11.0 
```

``` shell
$ pyenv shell 3.11.0
```

Then using poetry to start a new project, like

``` shell
$ poetry new myproject
```

would create a new folder with all the relevant stuff inside, and file `pyproject.toml` would be

``` toml
[tool.poetry]
name = "myproject"
version = "0.1.0"
description = ""
authors = ["Me <my@email.com>"]
readme = "README.md"

[tool.poetry.dependencies]
python = "^3.11"


[build-system]
requires = ["poetry-core"]
build-backend = "poetry.core.masonry.api"
```
notice the python version. However this is not the case as `poetry new <project name>` seems to ignore the active python version and go with the system one. 

Is this the intended behavior? Shouldn't `poetry new` and `poetry init` take into consideration the active python version?

I'm aware of other issues with similar questions:

- #5252 for example is about the command `poetry install` considering the active python version, so it regards the virtualenvs managed by poetry but not the `pyproject.toml` file.

</div>

PR #7100, opened 2022-11-26
feat: respect `virtualenvs.prefer-active-python` on `poetry new` and `init`

finswimmer:
# Pull Request Check List

Resolves: #7073

### DIFF
--- a/src/poetry/console/commands/init.py
+++ b/src/poetry/console/commands/init.py
@@ -164,7 +164,7 @@ The <c1>init</c1> command creates a basic <comment>pyproject.toml</> file in the
             default_python = (
                 "^"
                 + EnvManager.get_python_version(
-                    precious=2,
+                    precision=2,
                     prefer_active_python=config.get("virtualenvs.prefer-active-python"),
                     io=self.io,
                 ).to_string()
--- a/src/poetry/console/commands/init.py
+++ b/src/poetry/console/commands/init.py
@@ -1,7 +1,5 @@
 from __future__ import annotations
 
-import sys
-
 from pathlib import Path
 from typing import TYPE_CHECKING
 from typing import Any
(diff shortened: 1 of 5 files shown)


src/poetry/console/commands/new.py
```diff
@@ -71,8 +70,16 @@ def handle(self) -> int:
...
+        poetry_config = Config.create()
+        default_python = "^" + ".".join(
+            str(v)
+            for v in EnvManager.get_python_version(
+                prefer_active_python=poetry_config.get(
+                    "virtualenvs.prefer-active-python"
+                ),
+                io=self.io,
+            )[:2]
+        )
```
Secrus, line 82: Could you put it into a utils function? we could reduce duplication (command classes are already a mess)

finswimmer: I refactored it a bit, so this splitting and joining stuff happens in `EnvManager.get_python_version`.

src/poetry/console/commands/init.py
```diff
@@ -161,10 +160,16 @@ def handle(self) -> int:
...
         python = self.option("python")
         if not python:
-            current_env = SystemEnv(Path(sys.executable))
-            default_python = "^" + ".".join(
-                str(v) for v in current_env.version_info[:2]
+            config = Config.create()
+            default_python = (
+                "^"
+                + EnvManager.get_python_version(
+                    precious=2,
```
radoering, line 167: I stumbled a bit on the variable name `precious`. I wonder if there might be a better name or if that's a common name I just don't know?

finswimmer: The two hardest things in programming are naming and caching ;) So if you have a better name I'm glad to here it.

finswimmer:
> LGTM in general. I'm just wondering if turning methods of `EnvManager` into staticmethods is necessary / improves anything or if we could just create instances of `EnvManager` in the `init` and `new` command so we don't have to pass `io` as parameter of the methods?

`EnvManager` needs a `Poetry` instance if we create a new instance of it. In case of `init` and `new` we don't have a `Poetry` instance yet.

src/poetry/console/commands/init.py
radoering: What about `precision`? That's what we call it when we deal with versions in poetry-core.

finswimmer: I'm quite sure I wanted to name it like this :see_no_evil: :upside_down_face:

radoering (approved):

src/poetry/console/commands/new.py
vocqan-2gIbdy-pipguf: - [Python] -

PR #7110, opened 2022-11-27
Add additional method for CI installation using `pip-tools`

nikitabarskov:
I add the additional way to install `poetry` in CI environments.

It uses `pip-tools`, and `requirements.txt` with hashes. So, the level of reproducibility is quite high.

I've developed it with my colleague @aucampia, and we widely use it with a high level of success, so I decided to share it with the Poetry community.

I'm glad to hear your feedback!

### DIFF
--- a/docs/_index.md
+++ b/docs/_index.md
@@ -331,6 +331,40 @@ as the target environment managed by Poetry. If Poetry and your project are inst
 is likely to upgrade or uninstall its own dependencies (causing hard-to-debug and understand errors).
 {{% /note %}}
 
+**Using pip-tools**
+
+You can use `pip-tools` to generate `requirements.txt` for your poetry.
+
+The first thing you need is to create `requirements-poetry.in` with
+the following content:
+
+```text title="requirement-poetry.in"
+poetry==<poetry-version>
+```
+
+Then with `pip-tools` you can generate a `requirements-poetry.txt` file
+with hashes:
+
+```shell
+pip-compile --generate-hashes --rebuild --resolver backtracking --output-file requirements-poetry.txt requirements-poetry.in
+```
+
+The output file will contain all required dependencies for a given `poetry`
+version.
+
+Since Poetry has to be installed in its own virtual environment, you need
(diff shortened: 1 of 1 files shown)


Secrus:
This goes against the recommended way of working with Poetry, which is to install Poetry into a separate environment from your project.

Secrus:
Ok, I am puzzled. Why would you mix Poetry with pip-tools? What is the reason for it?

nikitabarskov:
> Ok, I am puzzled. Why would you mix Poetry with pip-tools? What is the reason for it?

There was no way to have an analog of `lockfile` when I wanted to install Poetry. To some extent, it might be considered an over-engineered solution.

But basically, the usage of `pip-tools` allows me to generate a `requirements.txt` with hashes containing dependencies only for Poetry. Then I use this `requirements.txt` to install Poetry within my CI/CD pipelines or development images.

You can take a look at the public example I have at my current workplace:

- the Dockerfile https://github.com/coopnorge/engineering-docker-images/blob/main/images/poetry-python3.11/context/Dockerfile

- and the commands https://github.com/coopnorge/engineering-docker-images/blob/main/images/poetry-python3.11/context/Makefile#L14-L23 I used to generate `requirements-poetry.txt` https://github.com/coopnorge/engineering-docker-images/blob/main/images/poetry-python3.11/context/requirements-poetry.txt

nikitabarskov:
As a side effect, with this approach later you can configure @dependabot or @renovatebot to update the `requirements-poetry.in` at almost 0 costs :)

Secrus:
Ok. While we are grateful for your contribution, we don't know `pip-tools` and don't want to carry the support burden when we introduce this in docs. We plan on providing a `requirements.txt`-like lockfile for Poetry installation but based on `pip` alone.

nikitabarskov:
> Ok. While we are grateful for your contribution, we don't know `pip-tools` and don't want to carry the support burden when we introduce this in docs. We plan on providing a `requirements.txt`-like lockfile for Poetry installation but based on `pip` alone.

This is excellent news for us! We also want to reduce the maintenance burden for our reproducible builds.

PR #7121, opened 2022-11-30
Calculate and store hash for url dependencies

dunkmann00:
This performs the same hashing process as is done on file dependencies, for url dependencies. This requires https://github.com/python-poetry/poetry-core/pull/535 to work.

I haven't updated the tests yet as I would first like to get feedback from the maintainers.

# Pull Request Check List

Resolves: #7122

### DIFF
--- a/src/poetry/installation/executor.py
+++ b/src/poetry/installation/executor.py
@@ -28,6 +28,7 @@ from poetry.utils._compat import decode
 from poetry.utils.authenticator import Authenticator
 from poetry.utils.env import EnvCommandError
 from poetry.utils.helpers import atomic_open
+from poetry.utils.helpers import get_file_hash
 from poetry.utils.helpers import pluralize
 from poetry.utils.helpers import remove_directory
 from poetry.utils.pip import pip_install
@@ -667,7 +668,7 @@ class Executor:
     @staticmethod
     def _validate_archive_hash(archive: Path, package: Package) -> str:
         file_dep = FileDependency(package.name, archive)
-        archive_hash: str = "sha256:" + file_dep.hash()
+        archive_hash: str = "sha256:" + get_file_hash(file_dep.full_path)
         known_hashes = {f["hash"] for f in package.files}
 
         if archive_hash not in known_hashes:
--- a/src/poetry/installation/executor.py
+++ b/src/poetry/installation/executor.py
@@ -15,7 +15,6 @@ from typing import TYPE_CHECKING
 from typing import Any
 
 from cleo.io.null_io import NullIO
-from poetry.core.packages.file_dependency import FileDependency
 from poetry.core.packages.utils.link import Link
 from poetry.core.pyproject.toml import PyProjectTOML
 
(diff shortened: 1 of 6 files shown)


dimbleby:
thanks!

I don't think that attaching the `hash()` method to the url dependency is the right way:

- you don't use `self` in that method at all, which is a clue that it doesn't particularly belong there
- you haven't achieved any code re-use, this essentially duplicates the `hash()` method on the file dependency

I would suggest that the method on the file dependency be deprecated or removed, and you replace it with a standalone utility function that just takes a `Path` and calculates the dependency.

Also you could put that utility in poetry proper rather than poetry-core, which will allow you to make the unit tests work without having cross-repository problems to work through

dunkmann00:
Okay, I moved `hash()` into `utils.helpers` and updated any calls to use this new function.

The tests still fail since the lockfile that's generated has the hashes in it and doesn't match the test fixtures. Is there a recommended way to generate the new fixture test files?

dimbleby:
> Is there a recommended way to generate the new fixture test files?

With a grand total of four failures it shouldn't be very tedious just to fix up those testcases manually,

`pytest -k <test-name> -vv` will give a fairly readable diff to tell you what's changed so that's easy to copy into the test script (if you're happy with it)

dunkmann00:
Fixed test failures and added in tests for the new `get_file_hash()` function. The tests are pulled over from what used to be in `poetry-core`.

dimbleby (approved):

src/poetry/installation/executor.py
```diff
@@ -667,7 +668,7 @@ def _download_link(self, operation: Install | Update, link: Link) -> Path:
     @staticmethod
     def _validate_archive_hash(archive: Path, package: Package) -> str:
         file_dep = FileDependency(package.name, archive)
-        archive_hash: str = "sha256:" + file_dep.hash()
+        archive_hash: str = "sha256:" + get_file_hash(file_dep.full_path)
```
radoering, line 671: Can we just pass `archive` to `get_file_hash()` and skip creating a `FileDependency`?

dunkmann00: **Short answer:** 
AFAICT that would be fine. Locally I made the change and it passes the tests and works when I tried installing something.

**Longer answer:**
Looking into it, the first thing that jumps out is that `FileDependency` does some checking around the `path` to make sure it is valid and also guarantees that `full_path` is absolute. 

Looking at [_download_link](https://github.com/python-poetry/poetry/blob/8c2113ce9b3c279337830fbce97ec413e8e152f4/src/poetry/installation/executor.py#L643), which is the only place `_validate_archive_hash` is called, `archive` seems like it will always be absolute, and will always be a valid file. So I think the checks in `FileDependency` are probably redundant...but it might be good if someone else takes a look to make sure I'm not missing something.

dunkmann00: I've looked this over again and do believe there are no problems doing this, so I've made the changes.

radoering (approved):

Issue #7123, opened 2022-11-30
Replace special chars with underscores for environment variables with credentials


bryant-finney:
- [x] I am on the [latest](https://github.com/python-poetry/poetry/releases/latest) stable Poetry version, installed using a recommended method.
- [x] I have searched the [issues](https://github.com/python-poetry/poetry/issues) of this repo and believe that this is not a duplicate.
- [x] I have searched the [FAQ](https://python-poetry.org/docs/faq/) and general [documentation](https://python-poetry.org/docs/) and believe that my question is not already covered.

- **Poetry version**: [>=1.0.0](https://github.com/python-poetry/poetry/releases/tag/1.0.0)
- **Python version**: `3`
- **OS version and name**: N/A
- **`pyproject.toml`**: [bryant-finney/pyproject.toml](https://gist.github.com/bryant-finney/c2537462fea99f1217a7ab113c56edd6)

## Request for Enhancement

Currently, I am unable to use environment variables to authenticate for installing packages from a private internal PyPI repository `example.com`.

In the linked [`pyproject.toml`](https://gist.github.com/bryant-finney/c2537462fea99f1217a7ab113c56edd6), the source repository named `example.com` provides a private package named `the-package`:

```toml
# ...

[[tool.poetry.source]]
name = "example.com"
url = "https://example.com/pypi/simple"
default = false
secondary = false

# ...
```

From my development machine, I can authenticate with the [example.com](https://https://example.com/pypi/simple) repository and install `the-package`:

```sh
❯ poetry config http-basic.example.com $EXAMPLE_USER $EXAMPLE_PASSWORD

❯ poetry add --group dev the-package
Using version ^1.2.3 for the-package

Updating dependencies
Resolving dependencies... (6.7s)

Writing lock file

Package operations: 8 installs, 0 updates, 0 removals

  • Installing click-didyoumean (0.3.0)
  • Installing click-plugins (1.1.1)
  • Installing click-repl (0.2.0)
[ ... ]
```

I also expected to use the environment variables `POETRY_HTTP_BASIC_EXAMPLE_COM_USERNAME` and `POETRY_HTTP_BASIC_EXAMPLE_COM_PASSWORD` for installations during Docker image builds.

However, this does not work: [`poetry.config.config:230`](https://github.com/python-poetry/poetry/blob/dcd48c8df6d22246c21c0243fd387e3a9b189f93/src/poetry/config/config.py#L230) only replaces `-`s with `_`s, so it does not match these variables.

[...]

PR #7127, opened 2022-11-30
Support repo names containing special characters

bryant-finney:
# Pull Request Check List

Resolves: #7123


src/poetry/config/config.py
```diff
@@ -227,7 +227,7 @@ def get(self, setting_name: str, default: Any = None) -> Any:
                 if repositories:
                     return repositories
 
-            env = "POETRY_" + "_".join(k.upper().replace("-", "_") for k in keys)
+            env = "POETRY_" + "_".join(re.sub(r"[\W]", "_", k.upper()) for k in keys)
```
dimbleby, line 230: per https://github.com/python-poetry/poetry/issues/7123#issuecomment-1332587118, this doesn't do anything for people who retrieve their configuration through the configuration file rather than through environment variables.  

Eg dots remain problematic, per the code a dozen lines up

PR #7136, opened 2022-12-03
simplify marker simplification

dimbleby:
The simplification and generalization of marker parsing and merging in https://github.com/python-poetry/poetry-core/pull/530 results in markers being parsed to disjunctive normal form.  

Obviously needs coordination with https://github.com/python-poetry/poetry-core/pull/530


radoering:
Can we just make the test forward compatible like `assert marker in [variant1, variant2]` to keep coordination simple? I think we can even keep this after the next core release so we don't have to adapt the test again if the marker changes to the original variant for some reason in the future.

radoering (approved):

Issue #7159, opened 2022-12-09
30X upload redirect bug, fails but reports success 


lineman60:
- **Poetry (version 1.2.2)**: 
- **3.11**: 
- **Debian GNU/Linux 10 (buster), 4.19.0-22-amd64**: 
- **[pyproject.toml](https://gist.github.com/lineman60/65984a1f9b7f35214bdd1c61f1a40e3a) **: 

- [x] I am on the [latest](https://github.com/python-poetry/poetry/releases/latest) stable Poetry version, installed using a recommended method.
- [x] I have searched the [issues](https://github.com/python-poetry/poetry/issues) of this repo and believe that this is not a duplicate.
- [x] I have consulted the [FAQ](https://python-poetry.org/docs/faq/) and [blog](https://python-poetry.org/blog/) for any relevant entries or release notes.
- [x] If an exception occurs when executing a command, I executed it again in debug mode (`-vvv` option) and have included the output below.

## Issue

When publishing with poetry and the response is in the 300 range, (other then 301) the upload reports success but upload does not succeed. This also happened with 308. 

```
$ poetry publish --repository foo -vvv
Adding repository foo (http://x.x.x.173:8090/redirect)

[keyring.backend] Loading KWallet
[keyring.backend] Loading SecretService
[keyring.backend] Loading Windows
[keyring.backend] Loading chainer
[keyring.backend] Loading libsecret
[keyring.backend] Loading macOS
No suitable keyring backend found
No suitable keyring backends were found
Keyring is not available, credentials will be stored and retrieved from configuration files as plaintext.
Found authentication information for foo.
Publishing poetry-demo (0.1.0) to foo
 - Uploading poetry_demo-0.1.0-py3-none-any.whl 0%[urllib3.connectionpool] Starting new HTTP connection (1): x.x.x.173:8090
 - Uploading poetry_demo-0.1.0-py3-none-any.whl 100%[urllib3.connectionpool] http://x.x.x.173:8090 "POST /redirect HTTP/1.1" 302 0

 - Uploading poetry_demo-0.1.0.tar.gz 100%[urllib3.connectionpool] http://x.x.x.173:8090 "POST /redirect HTTP/1.1" 302 0
```

I have a PR written to fix this, will submit in conjunction with this.  
PR: https://github.com/python-poetry/poetry/pull/7160

PR #7160, opened 2022-12-09
uploader: fix HTTP status code handling for redirects

lineman60:
# Pull Request Check List

Resolves: #7159

- [N/A ] Updated **documentation** for changed code.


src/poetry/publishing/uploader.py
```diff
@@ -261,12 +261,12 @@ def _upload_file(
                         headers={"Content-Type": monitor.content_type},
                         timeout=REQUESTS_TIMEOUT,
                     )
-                if resp is None or 200 <= resp.status_code < 300:
+                if resp is None or 200 <= resp.status_code < 299:
```
dimbleby, line 264: you have introduced an out by one error here and elsewhere (what happens for 299? etc)

src/poetry/publishing/uploader.py
```diff
@@ -275,7 +275,9 @@ def _upload_file(
                         "Redirects are not supported. "
                         "Is the URL missing a trailing slash?"
                     )
-                elif resp.status_code == 400 and "was ever registered" in resp.text:
+                elif (
```
dimbleby, line 278: this code branch looks as though it was aimed at the specific behaviour of a particular server (almost surely pypi).  I doubt that broadening the range of status codes is useful

src/poetry/publishing/uploader.py
```diff
@@ -261,12 +261,12 @@ def _upload_file(
...
                         timeout=REQUESTS_TIMEOUT,
                     )
-                if resp is None or 200 <= resp.status_code < 300:
+                if resp is None or 200 <= resp.status_code < 299:
                     bar.set_format(
                         f" - Uploading <c1>{file.name}</c1> <fg=green>%percent%%</>"
                     )
                     bar.finish()
-                elif resp.status_code == 301:
+                elif 300 <= resp.status_code < 399:
```
hutson, line 269:
```suggestion
                elif 300 <= resp.status_code < 400:
```

src/poetry/publishing/uploader.py
hutson, line 278: Seems unrelated to addressing the behavior around 3xx codes - https://github.com/python-poetry/poetry/issues/7159

lineman60: Removed this change.

src/poetry/publishing/uploader.py
lineman60: Thanks, fixed!

src/poetry/publishing/uploader.py
lineman60: updated to 400

dimbleby:
This looks harmless to me, but I don't follow why this fix is necessary.  

Wouldn't the code previously have gone through the `else ... resp.raise_for_status()` branch?  and then through the `except ... requests.HTTPError` branch, raising an `UploadError`?

ie the command should have failed all along, just without the text about "Redirects are not supported..."?

tests/publishing/test_uploader.py
```diff
@@ -65,6 +65,24 @@ def test_uploader_properly_handles_nonstandard_errors(
...
+        (303, "See Other"),
+        (302, "Found"),
+        (301, "Moved Permanently"),
+        (300, "Multiple Choices"),
+    ],
+)
+def test_uploader_properly_handles_redirects(
+    http: type[httpretty.httpretty], uploader: Uploader, status: int, body: str
+):
+    http.register_uri(http.POST, "https://foo.com", status=status, body=body)
```
radoering, line 83: Is something missing here? The test succeeds even without the change.

lineman60: @radoering  I for got the assert, should be fixed now.

radoering (approved):

Issue #5429, opened 2022-04-08
Wrong dependencies resolving for SQLAlchemy[postgresql_asyncpg]


evgen231:
- [x] I am on the [latest](https://github.com/python-poetry/poetry/releases/latest) Poetry version.
- [x] I have searched the [issues](https://github.com/python-poetry/poetry/issues) of this repo and believe that this is not a duplicate.
- [x] If an exception occurs when executing a command, I executed it again in debug mode (`-vvv` option).

- **OS version and name**: MacOS 12.2.1 on Apple M1 chip
- **Poetry version**: 1.1.13
- **Link of a [Gist](https://gist.github.com/) with the contents of your pyproject.toml file**: https://gist.github.com/evgen231/1cda84fbf99b826f9ed611d969fdd01e

## Issue

After installing [SQLAlchemy](https://github.com/sqlalchemy/sqlalchemy/) with postgresql_asyncpg extra there is no greenlet dependency.

```bash
# poetry update
Creating virtualenv example-OWA4dUH3-py3.9 in /Users/.../Library/Caches/pypoetry/virtualenvs
Updating dependencies
Resolving dependencies... (0.1s)

Writing lock file

Package operations: 2 installs, 0 updates, 0 removals

  • Installing asyncpg (0.25.0)
  • Installing sqlalchemy (1.4.35)
# poetry run pip freeze
asyncpg @ .../asyncpg-0.25.0.tar.gz
SQLAlchemy @ .../SQLAlchemy-1.4.35.tar.gz
```

Using pip:
```bash
# python -m pip install 'sqlalchemy[postgresql_asyncpg]'
...
# python -m pip freeze
asyncpg==0.25.0
greenlet==1.1.2
SQLAlchemy==1.4.35
```

evgen231:
Same thing with poetry 1.2.0b1.

finswimmer:
Hey @evgen231,

`sqlalchemy` has this marker for `greenlet`:

```
greenlet != 0.4.17;python_version>='3' and (platform_machine=='aarch64' or (platform_machine=='ppc64le' or (platform_machine=='x86_64' or (platform_machine=='amd64' or (platform_machine=='AMD64' or (platform_machine=='win32' or platform_machine=='WIN32'))))))
```

So my guess is, that your platform is not included here and it is intended that `greenlet` should not be installed.

fin swimmer

evgen231:
But pip does. So it's a bug in the pip and SQLAlchemy dependencies?

[...]

PR #7173, opened 2022-12-11
Miscellaneous fixes

dimbleby:
A small collection of miscellaneous fixes, none of which have individually seemed to me worth the bother of a pull request.  But I've now got enough of them queued up that, well, here we are.

Hopefully the commits and their messages are self-explanatory, these are all intended as uncontroversial.  But if some of them are for some reason harder to merge than others then say so and this can be broken up into more PRs.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -56,7 +56,7 @@ crashtest = "^0.4.1"
 dulwich = "^0.20.46"
 filelock = "^3.8.0"
 html5lib = "^1.0"
-importlib-metadata = { version = "^4.4", python = "<3.10" }
+importlib-metadata = { version = ">=4.4,<6", python = "<3.10" }
 jsonschema = "^4.10.0"
 keyring = "^23.9.0"
 lockfile = "^0.12.2"
@@ -88,14 +88,14 @@ pre-commit = "^2.6"
 [tool.poetry.group.test.dependencies]
 # Cachy frozen to test backwards compatibility for `poetry.utils.cache`.
 cachy = "0.3.0"
-deepdiff = "^5.0"
+deepdiff = "^6.2"
 flatdict = "^4.0.1"
 httpretty = "^1.0"
 pytest = "^7.1"
 pytest-cov = "^4.0"
 pytest-mock = "^3.9"
 pytest-randomly = "^3.12"
-pytest-xdist = { version = "^2.5", extras = ["psutil"] }
+pytest-xdist = { version = "^3.1", extras = ["psutil"] }
 zipp = { version = "^3.4", python = "<3.8" }
 
 [tool.poetry.group.typing.dependencies]
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -163,8 +163,6 @@ enable_error_code = [
 [[tool.mypy.overrides]]
 module = [
   'poetry.console.commands.self.show.plugins',
-  'poetry.installation.executor',
-  'poetry.mixology.version_solver',
   'poetry.plugins.plugin_manager',
   'poetry.repositories.installed_repository',
   'poetry.utils.env',
--- a/poetry.lock
+++ b/poetry.lock
@@ -70,14 +70,14 @@ redis = ["redis (>=3.3.6,<4.0.0)"]
 
 [[package]]
 name = "certifi"
-version = "2022.9.24"
+version = "2022.12.7"
 description = "Python package for providing Mozilla's CA Bundle."
 category = "main"
 optional = false
 python-versions = ">=3.6"
 files = [
-    {file = "certifi-2022.9.24-py3-none-any.whl", hash = "sha256:90c1a32f1d68f940488354e36370f6cca89f0f106db09518524c88d6ed83f382"},
-    {file = "certifi-2022.9.24.tar.gz", hash = "sha256:0d9c601124e5a6ba9712dbc60d9c53c21e34f5f641fe83002317394311bdce14"},
+    {file = "certifi-2022.12.7-py3-none-any.whl", hash = "sha256:4ad3232f5e926d6718ec31cfc1fcadfde020920e278684144551c91769c7bc18"},
+    {file = "certifi-2022.12.7.tar.gz", hash = "sha256:35824b4c3a97115964b408844d64aa14db1cc518f6562e8d7261699d1350a9e3"},
 ]
 
 [[package]]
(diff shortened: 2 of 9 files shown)


src/poetry/repositories/abstract_repository.py
```diff
@@ -30,7 +31,7 @@ def search(self, query: str) -> list[Package]:
     @abstractmethod
     def package(
         self,
-        name: str,
+        name: NormalizedName,
```
radoering, line 34: I'm wondering that mypy doesn't complain. There are more subclasses that still take pretty name and iirc some need the pretty name.

src/poetry/puzzle/provider.py
```diff
@@ -572,7 +572,7 @@ def complete_package(
                 dependency_package = DependencyPackage(
                     dependency,
                     self._pool.package(
-                        package.pretty_name,
+                        package.name,
```
radoering, line 575: Obviously, there is no test, but I expect this to cause some name changes in the lock file (again). (See other comment.) I'd like to avoid that if there is no good reason.

src/poetry/repositories/abstract_repository.py
dimbleby: Could you point out what you mean?  Like mypy, I didn't notice anything (though per the commit it's some weeks since I looked at this and so it's not fresh in my mind).

src/poetry/puzzle/provider.py
dimbleby: why would this cause a change in the lock file?

src/poetry/repositories/abstract_repository.py
radoering: https://github.com/python-poetry/poetry/blob/0c8e5b43d30d86d6ba7dc9b3068db26042e9cc53/src/poetry/repositories/cached_repository.py#L70-L78

https://github.com/python-poetry/poetry/blob/0c8e5b43d30d86d6ba7dc9b3068db26042e9cc53/src/poetry/repositories/legacy_repository.py#L48-L72

https://github.com/python-poetry/poetry/blob/0c8e5b43d30d86d6ba7dc9b3068db26042e9cc53/src/poetry/repositories/repository_pool.py#L110-L127

src/poetry/puzzle/provider.py
radoering: Because the pretty name is "lost" and thus the canonicalized name is written in the lock file.

src/poetry/repositories/abstract_repository.py
dimbleby: ah, somehow machine and man both missed that the legacy repository uses the un-normalized name: [here](https://github.com/python-poetry/poetry/blob/0c8e5b43d30d86d6ba7dc9b3068db26042e9cc53/src/poetry/repositories/cached_repository.py#L77).

I'll remove this change from the collection

src/poetry/puzzle/provider.py
```diff
@@ -945,19 +945,9 @@ def _merge_dependencies_by_constraint(
...
-                    # No marker or only extras
-                    continue
-
-                new_markers.append(marker)
-
-            if not new_markers:
-                continue
 
+        for constraint, _deps in by_constraint.items():
+            new_markers = [d.marker for d in _deps]
```
radoering, line 950: Does that maybe even fix #5429? I'd appreciate if you could check and if that's the case split this change into a separate PR and add a test. 🙏 

By the way, line 952 can probably be simplified to `dep.marker = MarkerUnion.of(*new_markers)`. If I don't miss something the marker of the first dependency is considered twice unnecessarily at the moment.

dimbleby: I considered MarkerUnion.of() but it's heading in the wrong direction with respect to https://github.com/python-poetry/poetry-core/pull/530, which removes much of the simplification from the `of()` methods

radoering: From an API perspective, I wonder if it might make sense to have two versions of `of()` (better names welcome). One that we use internally for cnf/dnf and one (that basically just does the `dnf(cnf())` trick) that can be used in cases like this. It seems more natural to use something like `of()` here. Or maybe just add a parameter to `of()` which is actively set by cnf/dnf?

dimbleby: so far as I can see this doesn't touch #5429

https://github.com/python-poetry/poetry-core/pull/530 is more-or-less making `XXXMarker.of()` a private method.  

I'm going to tweak this particular line so that it goes `AnyMarker().union(MarkerUnion(*new_markers)` which is certainly a bit artificial.  But I think it's the most efficient thing to do, or anyway will be if-and-when https://github.com/python-poetry/poetry-core/pull/530 merges.

I wouldn't be opposed to some new API allowing union-of-lots-of-markers to be more conveniently constructed in one go, but also I'm not especially excited about it.

radoering: I took a closer look. It's part of the issue. See #7175

dimbleby: Oh, how satisfying.  Nice.

pyproject.toml
```diff
@@ -56,7 +56,7 @@ crashtest = "^0.4.1"
 dulwich = "^0.20.46"
 filelock = "^3.8.0"
 html5lib = "^1.0"
-importlib-metadata = { version = "^4.4", python = "<3.10" }
+importlib-metadata = { version = "^5.1", python = "<3.10" }
```
radoering, line 59: According to https://pypi.org/project/importlib-metadata/, 4.6 is included in python 3.10. If 4.6 is enough for Python 3.10, why should we require 5.1 for python < 3.10?

dimbleby: no big deal either way, but IMO it's as well for application-type code to stay current. What is gained by allowing old versions of importlib-metadata?

Indeed we're currently explicitly excluding new versions of importlib-metadata, which seems a reasonable thing to happen by accident, but a strange thing to choose deliberately.

radoering: Yeah, `^4.4` isn't good, either. Probably, it should be `>=4.4,<6`.

Re allowing old versions: If you install poetry in its own virtual env (as recommended) it doesn't matter. But then we could also pin dependencies. 🤷

dimbleby: sure, it's a can of worms.  But for applications - and poetry really is intended to be installed in its own virtual environment - I don't see meaningful upside in allowing a broad range.  There certainly is downside: it would be some sort of commitment to support old versions of importlib-metadata (snd there is no testing to verify that this will work).

However I don't think this is important enough that I care to worry about it: if you're asking me to make a change, please say so.

radoering: > There certainly is downside: it would be some sort of commitment to support old versions of importlib-metadata (snd there is no testing to verify that this will work).

Personally, I like to keep constraints as broad as possible and only restrict them if a version is known to be incompatible. (At least, experienced users can easily exclude a bad version by themselves, but cannot allow an excluded version as easy.)

It might be a while before we find the time to discuss our update strategy internally.  So yes, maybe just change it and I'll merge it.

radoering (approved):

PR #7175, opened 2022-12-11
solver: fix a special case with unconditional dependency of an extra

radoering:
# Pull Request Check List

Resolves: #5429


dimbleby:
I can't see what's wrong with this and the testcase is convincing - and that's about as much as I hope for when making changes like this!

I might have preferred to save a level of indentation by setting `_dep.marker = dep.marker` unconditionally, but of course it really doesn't matter.

mkniewallner (approved):
Tested this PR over #5429 by requiring `sqlalchemy[postgresql_asyncpg]` then `sqlalchemy[asyncio]`, and it was able to include `greenlet` for both. So with this + the tests, pretty confident about the fix in this PR.

PR #7169, opened 2022-12-10
Add lockfile as an explicit dependency

orf:
Resolves #7171

Whilst exploring why https://github.com/Homebrew/homebrew-core/pull/117779 is failing, I saw the following traceback:

```
  File "/usr/local/Cellar/poetry/1.3.0/libexec/lib/python3.11/site-packages/poetry/puzzle/provider.py", line 37, in <module>
    from poetry.vcs.git import Git
  File "/usr/local/Cellar/poetry/1.3.0/libexec/lib/python3.11/site-packages/poetry/vcs/git/__init__.py", line 3, in <module>
    from poetry.vcs.git.backend import Git
  File "/usr/local/Cellar/poetry/1.3.0/libexec/lib/python3.11/site-packages/poetry/vcs/git/backend.py", line 21, in <module>
    from poetry.utils.authenticator import get_default_authenticator
  File "/usr/local/Cellar/poetry/1.3.0/libexec/lib/python3.11/site-packages/poetry/utils/authenticator.py", line 15, in <module>
    import lockfile
ModuleNotFoundError: No module named 'lockfile'
```

The `lockfile` module was missing from the resources block when it should have been present. However, Poetry is using the `lockfile` package directly despite it being brought in via an transient dependency:

https://github.com/python-poetry/poetry/blob/aa48815d2253e318d6e70c9a15fc4def0b48c870/src/poetry/utils/authenticator.py#L15

This MR adds it as a direct dependency.

PR #7178, opened 2022-12-12
release: bump version to 1.3.1

neersighted:
### Fixed

- Fix an issue where an explicit dependency on `lockfile` was missing, resulting in a broken Poetry in rare circumstances ([7169](https://github.com/python-poetry/poetry/pull/7169)).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,13 @@
 # Change Log
 
 
+## [1.3.1] - 2022-12-12
+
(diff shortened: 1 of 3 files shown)


Secrus:
Did you mean to include dependencies update?

neersighted:
Yes, the re-lock is because the backport PR that added the explicit dep had a stale lockfile.

Secrus (approved):

Issue #7158, opened 2022-12-08
Poetry selects wrong Python version if Poetry's installation differs from `python3` (or `python3` changes)


ShadowLNC:
- **Poetry version**: 1.2.2
- **Python version**: CPython 3.9.12
- **OS version and name**: MacOS (darwin) 12.6.1 Monterey (ARM architecture)
- **pyproject.toml**: See below

- [x] I am on the [latest](https://github.com/python-poetry/poetry/releases/latest) stable Poetry version, installed using a recommended method.
- [x] I have searched the [issues](https://github.com/python-poetry/poetry/issues) of this repo and believe that this is not a duplicate. Related but not the same as far as I can tell:
    #7075
    #6978
    #6893 
- [x] I have consulted the [FAQ](https://python-poetry.org/docs/faq/) and [blog](https://python-poetry.org/blog/) for any relevant entries or release notes.
- [x] (N/A) If an exception occurs when executing a command, I executed it again in debug mode (`-vvv` option) and have included the output below.

## Issue

If Poetry's internal virtual environment (used to execute Poetry) doesn't match the version resolved by the `python3` command, then Poetry doesn't select the correct Python version (defined in `pyproject.toml`) for installation when creating a new virtual environment for a project.

There are two different ways that I've found to achieve this scenario.

### Scenario 1:

1. Install Python 3.10 (or 3.9 or 3.8, just referring to 3.10 from now on) from https://python.org (using standard installer)
1. (Now `$PATH` resolves `python3` to `3.10`)
1. Run `curl -sSL https://install.python-poetry.org | python3 -` as officially supported installation mechanism
1. Install Python 3.11 from https://python.org
1. (Now `$PATH` resolves `python3` to `3.11`)

### Scenario 2:

1. Install Python 3.9 (or any version other than 3.11) from https://python.org (using standard installer)
1. Install Python 3.11 from https://python.org
1. (Now `$PATH` resolves `python3` to `3.11`)
1. Run `curl -sSL https://install.python-poetry.org | python3.9 -` 

### Outcome

Relaunch the terminal to ensure `$PATH` changes take effect.

```sh
$ poetry install --no-root
The currently activated Python version 3.9.12 is not supported by the project (~3.11).
Trying to find and use a compatible version. 
Using python3 (3.11.0)
Creating virtualenv poetrytest in /Users/scott/Documents/automation/poetrytest/.venv
Installing dependencies from lock file

Package operations: 1 install, 0 updates, 0 removals

[...]

PR #7221, opened 2022-12-19
fix(env): get full python path for appropriate executable

finswimmer:
Passing `python3` to `virtualenv` will result in a wrong Python version, if `virtualenv` is installed in a venv with a different Python Version associated with the `python3` command. So it is necessary to get the full python path for the executable before passing it to `virtualenv`.

During the investigation of this bug I recognize that `_detect_active_python` catches a wrong Exception. I fixed this in this PR as well.

# Pull Request Check List

Resolves:  #7158

### DIFF
--- a/src/poetry/utils/env.py
+++ b/src/poetry/utils/env.py
@@ -995,7 +995,7 @@ class EnvManager:
                     self._io.write_error_line(
                         f"Using <c1>{python}</c1> ({python_patch})"
                     )
-                    executable = python
+                    executable = self._full_python_path(python)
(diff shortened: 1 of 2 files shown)


radoering (approved):

rpf3:
@radoering will this be released into a new version soon?

radoering:
You can follow the progress of the next release in #7357. The exact release date depends on the available time of (other) maintainers.

rpf3:
Thanks for the update.  I'll track it there.

PR #7232, opened 2022-12-22
Fix 1.3 performance regression when locking with PyPI repository

radoering:
I noticed a significant performance regression when locking the shootout example with a warm cache. The regression is caused by the removal of cachy (#5868) and has been fixed only for legacy repositories in #6932. I refactored `_get_page` a bit so that the same fix applies to PyPI repositories.

poetry | time for `poetry lock` with warm cache
--- | ---
1.2.2 | 17 s
1.3.1 | 38 s
this PR | 17 s


dimbleby:
Huh, I'd considered this but convinced myself that the cachecontrol cache was fast enough.  Can't argue with facts though, nice win.

`assert isinstance(json_page, SimpleJsonPage)` alerts my smell-detector, there are a couple of curious things going on here

- we only need specifically a `SimpleJsonPage` so that we can go ` pretty_name = json_page.content["name"]`.  However since the name that the JSON API returns is normalized this is pointless: we're not succeeding in extracting a pretty name anyway and might as well not bother
- then it would be enough to `assert json_page is not None`: which raises the question of why the other `_get_page` methods return an optional thing whereas the pypi one returns a definite thing.

neersighted (approved):

PR #7241, opened 2022-12-23
poetry check command accepts --directory argument

marmig0404:
# Pull Request Check List

Resolves: #7240

~~I'm not sure about how to add some tests for my code, but~~ I built a local version and the [output](https://gist.github.com/marmig0404/5db87fc624571df4ae65f1885f5c10ed) confirms the `poetry check` command functions as documented by the global options section on the poetry website.


neersighted (changes requested):
I'd prefer to use the value constructed during Application init over duplicating the logic here -- can we get at that value from this command?

marmig0404:
> the value constructed during Application init

i achieved this using the path from `self.poetry.file.path` in the check command.

this now has some maybe unexpected behavior where running `poetry check --directory bar` while in `foo/` with a `foo/pyproject.toml`, the poetry check will run on the file in `foo`, not the specified directory.

should this be expected?

neersighted:
Yes, the intention is always to run in the context of the closest pyproject.toml. This mirrors what happens normally when you run Poetry with a CWD set, and the semantics of similar commands like Git.

src/poetry/console/commands/check.py
```diff
@@ -61,7 +59,8 @@ def handle(self) -> int:
         from poetry.factory import Factory
 
         # Load poetry config and display errors, if any
-        poetry_file = Factory.locate(Path.cwd())
+        project_path = self.poetry.file.path
+        poetry_file = Factory.locate(project_path)
         config = PyProjectTOML(poetry_file).poetry_config
```
finswimmer, line 64: Can't we use `self.poetry.file.path` directly as input for `PyProjectTOML`? :thinking:

finswimmer (changes requested):

src/poetry/console/commands/check.py
marmig0404: Seems reasonable enough, project_path and poetry_file lead to the same path no matter if the file was specified by argument or determined implicitly. I'll test that change some more and push it up.

Issue #7239, opened 2022-12-22
Pre-commit hook skips because pyproject.toml is in subdirectory


marmig0404:
- **Poetry version**: Poetry (version 1.3.1)
- **Python version**: Python 3.11.0
- **OS version and name**: Windows 11
- **pyproject.toml**: https://gist.github.com/marmig0404/cf3c44f35ad86502ac82f104e77aab73

- [x] I am on the [latest](https://github.com/python-poetry/poetry/releases/latest) stable Poetry version, installed using a recommended method.
- [x] I have searched the [issues](https://github.com/python-poetry/poetry/issues) of this repo and believe that this is not a duplicate.
- [x] I have consulted the [FAQ](https://python-poetry.org/docs/faq/) and [blog](https://python-poetry.org/blog/) for any relevant entries or release notes.

## Issue
In my project, I have a poetry-check pre-commit hook (also shared in [gist](https://gist.github.com/marmig0404/cf3c44f35ad86502ac82f104e77aab73)) which gets skipped when ran, saying: 
    `poetry-check.........................................(no files to check)Skipped`

I am almost entirely sure that this is because my `pyproject.toml` is in a sub-directory of my repo as only part of my project relies on it.

I would assume this is an easy fix, changing the files parameter in [`.pre-commit-hooks.yaml`](https://github.com/python-poetry/poetry/blob/master/.pre-commit-hooks.yaml) to match sub-directories too.

Or I might just be missing an argument in my hook to specify the directory, but none are mentioned on the poetry website where the hook is documented.

dimbleby:
> I would assume this is an easy fix

did you try it?

clone poetry, make the hook work, submit an MR!  (or close this if it turns out not to be so easy after all)

marmig0404:
> did you try it?

I figured I would pose the issue to see if I'm just missing an argument or something.
Currently investigating making an MR, I'm not too familiar with writing hooks...

marmig0404:
I think this goes much deeper than just the hook...

I can change the hook to match pyproject.toml in any sub-directory with `files: .*pyproject.toml`, but because the hook runs `poetry check` at the top level, the `pypoject.toml` wont be found, even if the hook is triggered.

`poetry check` could be made more extensible if it took the `--directory` argument as specified by [global options](https://python-poetry.org/docs/cli/#global-options) for poetry commands, but atm it doesn't look like it does. 

I would be looking for some support before going down this route. This is kinda stopped in its tracks unless the above step is done.

[...]

PR #7242, opened 2022-12-23
pre-commit matches pyproject.toml in subdirs

marmig0404:
# Pull Request Check List

Resolves: #7239

- [x] ~~Added **tests** for changed code.~~ No testing for pre-commit hooks available
- [x] Depends on #7241

### DIFF
--- a/.pre-commit-hooks.yaml
+++ b/.pre-commit-hooks.yaml
@@ -5,7 +5,7 @@
   language: python
   language_version: python3
   pass_filenames: false
-  files: ^pyproject.toml$
+  files: .*pyproject.toml
 
 - id: poetry-lock
   name: poetry-lock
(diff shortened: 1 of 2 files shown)


.pre-commit-hooks.yaml
```diff
@@ -5,7 +5,7 @@
   language: python
   language_version: python3
   pass_filenames: false
-  files: ^pyproject.toml$
+  files: .*pyproject.toml
```
wagnerluis1982, line 8: Minor, but I think this may match files like `p-pyproject.toml` or `pyproject.toml.old`

I'm assuming this is regex, so how about if you change to something like the following?
```
^(.*/)?pyproject.toml$
```

marmig0404: This is great, I knew my implementation was simplistic, but I wasn't sure how to improve it.
Thank you 😄

wagnerluis1982: Glad that worked ☺️

radoering (approved):

Issue #5066, opened 2022-01-19
Surprising behaviour (and missing warning?) with overlaping python versions


DMRobertson:
- [x] I am on the [latest](https://github.com/python-poetry/poetry/releases/latest) Poetry version.
- [x] I have searched the [issues](https://github.com/python-poetry/poetry/issues) of this repo and believe that this is not a duplicate.
  - closest I could find was https://github.com/python-poetry/poetry/issues/4959 but there are no markers here; just using the python version syntax only.
- [x] If an exception occurs when executing a command, I executed it again in debug mode (`-vvv` option).

- **OS version and name**: Linux, Fedora 35
- **Poetry version**: $ poetry --version
Poetry version 1.1.12
- **Link of a [Gist](https://gist.github.com/) with the contents of your pyproject.toml file**: 

## Issue

When experimenting with how poetry lock files handle different python versions, I tried `poetry install`ing from the following `pyproject.toml`.

```toml
[tool.poetry]
name = "example-poetry"
version = "0.1.0"
description = ""
authors = ["David Robertson <david.m.robertson1@gmail.com>"]

[tool.poetry.dependencies]
python = "^3.10"

[tool.poetry.dev-dependencies]
pytest = "^5.2"
pillow = [
	{ version = "^9", python = "^3.9" },
	{ version = "^7", python = "^3.7" },
]

[build-system]
requires = ["poetry-core>=1.0.0"]
build-backend = "poetry.core.masonry.api"
```

I hadn't noticed that the project was scoped to `python ^= 3.10`. Running this on Python 3.10, I expected the constraing `python ^3.9` to beat `python ^3.7`, and so expected poetry to select  `pillow ^9` at install time. What I saw instead was that poetry reported installing _two_ copies of pillow:

```
$ poetry install
Creating virtualenv example-poetry-ZY5Szwk5-py3.10 in /home/dmr/.cache/pypoetry/virtualenvs
Updating dependencies
Resolving dependencies... (1.5s)

Writing lock file

Package operations: 10 installs, 0 updates, 0 removals

  • Installing pyparsing (3.0.6)
  • Installing attrs (21.4.0)
  • Installing more-itertools (8.12.0)
  • Installing packaging (21.3)
  • Installing pluggy (0.13.1)
  • Installing py (1.11.0)
  • Installing wcwidth (0.2.5)
  • Installing pillow (7.2.0)
  • Installing pillow (9.0.0)
  • Installing pytest (5.4.3)

[...]

PR #7257, opened 2022-12-27
full support for duplicate dependencies with overlapping markers

radoering:
Resolves: #5066

Requires: python-poetry/poetry-core#530, python-poetry/poetry-core#546, python-poetry/poetry-core#547

- replace `_merge_dependencies_by_marker` and `_handle_any_marker_dependencies` by the more general `_resolve_overlapping_markers`
- raise `IncompatibleConstraintsError` if there are conflicts (due to incompatible version constraints or source types) in the constraints of a package

Tested with [opencv](https://pypi.org/project/opencv-python/4.6.0.66/), which has the most duplicate dependencies with overlapping markers I know about:
```
numpy (>=1.13.3) ; python_version < "3.7"
numpy (>=1.21.2) ; python_version >= "3.10"
numpy (>=1.21.2) ; python_version >= "3.6" and platform_system == "Darwin" and platform_machine == "arm64"
numpy (>=1.19.3) ; python_version >= "3.6" and platform_system == "Linux" and platform_machine == "aarch64"
numpy (>=1.14.5) ; python_version >= "3.7"
numpy (>=1.17.3) ; python_version >= "3.8"
numpy (>=1.19.3) ; python_version >= "3.9"
```

Especially the markers with `platform_system` and `platform_machine`, which overlap with most other markers make this example quite complex.

~~TODO: Investigation of the influence on the performance for (more) real-world examples~~

### DIFF
--- a/src/poetry/puzzle/provider.py
+++ b/src/poetry/puzzle/provider.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
+import itertools
 import logging
 import re
 import time
@@ -13,10 +14,10 @@ from typing import cast
 from cleo.ui.progress_indicator import ProgressIndicator
 from poetry.core.constraints.version import EmptyConstraint
 from poetry.core.constraints.version import Version
+from poetry.core.constraints.version import VersionRange
 from poetry.core.packages.utils.utils import get_python_constraint_from_marker
 from poetry.core.version.markers import AnyMarker
-from poetry.core.version.markers import EmptyMarker
-from poetry.core.version.markers import MarkerUnion
+from poetry.core.version.markers import union as marker_union
 
 from poetry.mixology.incompatibility import Incompatibility
 from poetry.mixology.incompatibility_cause import DependencyCause
@@ -60,10 +61,22 @@ class IncompatibleConstraintsError(Exception):
     Exception when there are duplicate dependencies with incompatible constraints.
     """
 
-    def __init__(self, package: Package, *dependencies: Dependency) -> None:
-        constraints = "\n".join(dep.to_pep_508() for dep in dependencies)
+    def __init__(
+        self, package: Package, *dependencies: Dependency, with_sources: bool = False
+    ) -> None:
+        constraints = []
+        for dep in dependencies:
+            constraint = dep.to_pep_508()
+            if dep.is_direct_origin():
+                # add version info because issue might be a version conflict
+                # with a version constraint
+                constraint += f" ({dep.constraint})"
+            if with_sources and dep.source_name:
+                constraint += f" ; source={dep.source_name}"
+            constraints.append(constraint)
         super().__init__(
-            f"Incompatible constraints in requirements of {package}:\n{constraints}"
+            f"Incompatible constraints in requirements of {package}:\n"
+            + "\n".join(constraints)
         )
 
 
@@ -591,53 +604,13 @@ class Provider:
 
             self.debug(f"<debug>Duplicate dependencies for {dep_name}</debug>")
 
-            # Group dependencies for merging.
-            # We must not merge dependencies from different sources!
-            dep_groups = self._group_by_source(deps)
-            deps = []
-            for group in dep_groups:
-                # In order to reduce the number of overrides we merge duplicate
-                # dependencies by constraint. For instance, if we have:
-                #   • foo (>=2.0) ; python_version >= "3.6" and python_version < "3.7"
-                #   • foo (>=2.0) ; python_version >= "3.7"
-                # we can avoid two overrides by merging them to:
-                #   • foo (>=2.0) ; python_version >= "3.6"
-                # However, if we want to merge dependencies by constraint we have to
-                # merge dependencies by markers first in order to avoid unnecessary
-                # solver failures. For instance, if we have:
-                #   • foo (>=2.0) ; python_version >= "3.6" and python_version < "3.7"
-                #   • foo (>=2.0) ; python_version >= "3.7"
-                #   • foo (<2.1) ; python_version >= "3.7"
-                # we must not merge the first two constraints but the last two:
-                #   • foo (>=2.0) ; python_version >= "3.6" and python_version < "3.7"
-                #   • foo (>=2.0,<2.1) ; python_version >= "3.7"
-                deps += self._merge_dependencies_by_constraint(
-                    self._merge_dependencies_by_marker(group)
-                )
+            # For dependency resolution, markers of duplicate dependencies must be
+            # mutually exclusive.
+            deps = self._resolve_overlapping_markers(package, deps)
+
             if len(deps) == 1:
                 self.debug(f"<debug>Merging requirements for {deps[0]!s}</debug>")
                 dependencies.append(deps[0])
-                continue
-
-            # We leave dependencies as-is if they have the same
-            # python/platform constraints.
-            # That way the resolver will pickup the conflict
-            # and display a proper error.
-            seen = set()
-            for dep in deps:
-                pep_508_dep = dep.to_pep_508(False)
-                if ";" not in pep_508_dep:
-                    _requirements = ""
-                else:
-                    _requirements = pep_508_dep.split(";")[1].strip()
-
-                if _requirements not in seen:
-                    seen.add(_requirements)
-
-            if len(deps) != len(seen):
-                for dep in deps:
-                    dependencies.append(dep)
-
                 continue
 
             # At this point, we raise an exception that will
@@ -665,8 +638,6 @@ class Provider:
                 f"<warning>Different requirements found for {warnings}.</warning>"
             )
 
-            deps = self._handle_any_marker_dependencies(package, deps)
-
             overrides = []
             overrides_marker_intersection: BaseMarker = AnyMarker()
             for dep_overrides in self._overrides.values():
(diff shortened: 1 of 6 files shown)


radoering:
Times for `poetry lock` with a warm cache:

|---|---|---|
|#4670|1 s|1 s|
|#4870|58 s|58 s|
|shootout example|16 s|10 s|
|large internal project|210 s|53 s|

Number of overrides:

|`pyproject.toml` from ...| number of overrides without PR | number of overrides with PR |
|---|---|---|
|#4670|24|24|
|#4870|52|52|
|shootout example|28|18|
|large internal project|93|15|

Considering the examples, the performance of our marker arithmetic seems to be fast enough to handle this PR: There is no measurable performance regression even in cases where the number of overrides does not change.

neersighted:
PTAL @dimbleby

neersighted (approved):

src/poetry/puzzle/provider.py
```diff
@@ -845,118 +817,121 @@ def _merge_dependencies_by_constraint(
...
+            by_constraint: dict[VersionConstraint, list[Dependency]] = defaultdict(list)
+            for dep in group:
+                by_constraint[dep.constraint].append(dep)
+            for constraint, deps in by_constraint.items():
+                new_markers = [dep.marker for dep in deps]
+                dep = deps[0]
+                dep.marker = marker_union(*new_markers)
+                by_constraint[constraint] = [dep]
+
+            merged_dependencies += [value[0] for value in by_constraint.values()]
```
dimbleby, line 839:
```suggestion

            for deps in by_constraint.values():
                new_markers = [dep.marker for dep in deps]
                dep = deps[0]
                dep.marker = marker_union(*new_markers)
                merged_dependencies.append(dep)
```

src/poetry/puzzle/provider.py
```diff
@@ -845,118 +817,121 @@ def _merge_dependencies_by_constraint(
...
+                        raise IncompatibleConstraintsError(
+                            package, dep, specific_source_dependency, with_sources=True
+                        )
+                    specific_source_dependency = dep
+                constraint = constraint.intersect(dep.constraint)
+            if constraint.is_empty():
+                # conflict in overlapping area
+                raise IncompatibleConstraintsError(package, *used_dependencies)
+
+            if set(uses) == {False}:
```
dimbleby, line 909:
```suggestion
            if not any(uses):
```

dimbleby (commented):
Looks good I think, though my head now hurts a little...

Feels as though there's some danger of this being horribly inefficient if there are lots of overlapping markers to resolve.  But probably folk mostly don't do that and certainly the examples showing that this is typically a win by reducing overrides is encouraging.  So I guess we hope that this continues to hold when exposed to more examples from the outside world...

Made a couple of more or less pedantic code-improving suggestions.

dimbleby:
Here's an example in which `_resolve_overlapping_markers()` has to do too much work:
```toml
[tool.poetry.dependencies]
python = "^3.5"
oldest-supported-numpy = "^2022.11.19"
```
see https://github.com/scipy/oldest-supported-numpy/blob/main/setup.cfg

radoering:
I can't get a solution (within a reasonable amount of time) with or without this PR. Thus, probably not a blocker?

dimbleby:
I'm not sure how bad I think this is.  I agree that with that example I am not getting a solution either before or after, or anyway not in an amount of time that I am willing to wait.

You can use the python requirement as a knob to adjust "how many combinations of markers are there to work through?", and therefore how slow this is.  eg `python = "^3.8"` is just about tolerable.

However it is all made a bit muddier by finding that the overlapping markers produce some incompatible requirements, which the new code discovers and the old code does not.  (I've opened an issue at `oldest-supported-numpy`.)  

So that's an example where this change is "breaking" (defensible as "fixing" really, but expect bug reports anyway!) - and because it's breaking I'm not sure what the performance impact really is for that case.

Not sure this is helping much: I think this is an improvement, but I sure would be happier if it could be made faster!

radoering:
After fixing the requirements of `oldest-supported-numpy` (cf scipy/oldest-supported-numpy#71) and with python-poetry/poetry-core#608, which is required to get sensible results for this package anyway, I tried some ranges with and without the PR:

Python constraint|without PR|with PR
---|---|---
^3.10|1 s|1 s
^3.9|44 s|13 s
\>=3.8,<3.10|> 1.000 s|69 s
\>=3.5,<3.8|858 s|318 s

Larger ranges still took too long. Maybe, I was just lucky but the results are pretty clear...

dimbleby:
Interesting, do the changes you have made to oldest-supported-numpy have a negative effect on poetry performance?

without the PR and also without those changes I see:
| Python constraint | Time |
|-------------------|------|
| >=3.5,<3.8 | 5s |
| >=3.8,<3.10 | 3s |
| ^3.9 | 3s |
| ^3.10 | 1s |

where the first of those fails to find a solution and some of the others ought to (because the code without the PR is not correct); but nevertheless these times are _much_ faster than you are reporting.

I'm all in favour of concluding that the new code is both more correct and also faster than the old code, that would be lovely.  I'm just a bit confused about why we are seeing such different results.

radoering:
The changes to oldest-supported-numpy and python-poetry/poetry-core#608 both have a negative impact on performance.

dimbleby:
I haven't profiled, so this might be a bad or irrelevant idea...

I wonder whether `_is_relevant_marker()` is slow: this bit looks like quite a lot of work:
```python
            and self._python_constraint.allows_any(
                get_python_constraint_from_marker(marker)
            )
```
thought: the project python constraint could be transformed into an equivalent marker (once, up-front), and then intersected with `used_marker_intersection`.  If that intersection is (not) empty then `used_marker_intersection` is (not) irrelevant to the current project.

As I say, I haven't tested whether that's definitely a slow bit that is worth optimising: and anyway it isn't completely obvious that this would even be an improvement.  So don't block the merge on this idea!

dimbleby:
oh, but a better idea: the calculation of the intersection is currently rather slow.  Much faster is to arrange it as something like this:
```
              markers = (
                  dep.marker if use else dep.marker.invert()
                  for use, dep in zip(uses, dependencies)
              )
              used_marker_intersection: BaseMarker = AnyMarker()
              for m in markers:
                  used_marker_intersection = used_marker_intersection.intersect(m)
```
presumably because most of the time the intersection collapses to the empty marker pretty quickly; and once that has happened we can avoid the work of DNF-ing and suchlike on the remaining markers.

dimbleby:
and in fact that observation can be encouraged by putting the positive markers early on in the intersection, like this (could do with prettying up!):
```python
              markers = (
                  dep.marker if use else dep.marker.invert()
                  for use, dep in sorted(
                      zip(uses, dependencies), key=lambda ud: ud[0], reverse=True
                  )
              )
```
That's relying on the heuristic that if there are a lot of markers present, then surely they most don't intersect one another.  But that seems quite plausible, it would be quite hard to write them anyway

radoering:
`_is_relevant_marker()` seems to be irrelevant. 😉

The other suggestion looks very promising. (I have to run some additional tests later.)

dimbleby:
An easy win in `_merge_dependencies_by_constraint()`:
```diff
             for deps in by_constraint.values():
-                new_markers = [dep.marker for dep in deps]
                 dep = deps[0]
-                dep.marker = marker_union(*new_markers)
+                if len(deps) > 1:
+                    new_markers = (dep.marker for dep in deps)
+                    dep.marker = marker_union(*new_markers)
```
no need to perform all the dnf-and-cnf complexity when trying to merge a list that only contained a single marker in the first place

radoering:
I applied both suggestions and checked that there is no negative impact on the examples in https://github.com/python-poetry/poetry/pull/7257#issuecomment-1575046996.

Measuring https://github.com/python-poetry/poetry/pull/7257#issuecomment-1597713699 again:

Python constraint|before optimization|after optimization 1|after optimization 2
---|---|---|---
^3.9|13 s|10 s|4 s
\>=3.8,<3.10|69 s|22 s|3 s
\>=3.5,<3.8|318 s|15 s|9 s

The only drawback of the second optimization is that the locked markers become more verbose. See `test_solver_duplicate_dependencies_with_overlapping_markers_complex`. (Of course, the performance improvement is more important than consise markers in the lock file.)

dimbleby:
Seems like what we're seeing here is that the performance of marker operations is still expensive, when they get sufficiently complicated.  Nearly all of the time in the solving is now taken up with performing the single intersection of the final combination of markers (the case where all `uses` are `False`).  Improvements recorded above are great: but with broader python constraints this package is still too slow to solve.

`oldest-supported-numpy` is certainly more complicated than most, but it is a real found-in-the-wild example.  I don't think that it's necessary to completely solve that to merge this! Especially given the evidence that before this MR poetry was also nowhere near to finding a solution, so this isn't making things worse.

However, if you have any good ideas then this still would be worth addressing....

I suspect that we are doing a lot of repeated work to normalize the same markers again and again.  I don't think I'm going to get to trying anything out soon, but I wonder if there are wins to be had either by caching answers, or having markers remember that they are already in DNF (therefore another dnf() can be made a no-op).

radoering:
> but I wonder if there are wins to be had either by caching answers, or having markers remember that they are already in DNF

On their own, both appear to have a significant effect on performance. However, when caching answers, remembering that a marker is already DNF seems to only have a very small effect not worth the more complex code. There might be some cases where it's still relevant but I think I'll wait until we find one in the wild.

Updated measurements:

Python constraint|without python-poetry/poetry-core#609|with python-poetry/poetry-core#609
---|---|---
^3.9|4 s|1 s
\>=3.8,<3.10|3 s|1 s
\>=3.5,<3.8|9 s| 2 s

Broader Python constraints are still too slow...

Then, I switched from scipy/oldest-supported-numpy#71 to scipy/oldest-supported-numpy#74 (with https://github.com/scipy/oldest-supported-numpy/pull/74#discussion_r1239887547) because that's more likely to be merged:

Python constraint|without python-poetry/poetry-core#609|with python-poetry/poetry-core#609
---|---|---
^3.8|40 s|2 s
^3.7|too long|32 s

neersighted (approved):

PR #7286, opened 2023-01-03
Add clarifications for `virtualenvs.path` settings

spirit-q2:


PabloEmidio (dismissed):

neersighted (changes requested):
The first note is misleading, and should probably be combined with the second to be something more like:

> This setting controls the global virtual environment storage path. It most likely will not be useful at the local level. To store virtual environments in the project root, see `virutualenvs.path`.

"unpredictable behavior" is true as there is currently a bug with a relative path, but it makes more sense to fix that bug than it does to add nonspecific documentation warnings unlikely to be removed when the bug is fixed.

spirit-q2:
updated

neersighted (approved):

Issue #7247, opened 2022-12-24
Pre-commit hook for check and lock not working in subdirectories.


gnkow:
- [x] I am on the [latest](https://github.com/python-poetry/poetry/releases/latest) stable Poetry version, installed using a recommended method.
- [x] I have searched the [issues](https://github.com/python-poetry/poetry/issues) of this repo and believe that this is not a duplicate.
- [x] I have consulted the [FAQ](https://python-poetry.org/docs/faq/) and [blog](https://python-poetry.org/blog/) for any relevant entries or release notes.
- [x] If an exception occurs when executing a command, I executed it again in debug mode (`-vvv` option) and have included the output below.

## Issue
I've a multi-root project directory such as:

```
/api (poetry/python)
     /pyproject.toml
/api2 (non-python)
.pre-commit-config.yml
```

My .pre-commit-config.yml states:
```
repos:
  - repo: https://github.com/python-poetry/poetry
    rev: 1.3.1
    hooks:
      - id: poetry-check
      - id: poetry-lock
```

Whenever I try to use the pre-commit hook I get this:

```
poetry-check.........................................(no files to check)Skipped
poetry-lock..............................................................Failed
- hook id: poetry-lock
- exit code: 1

Poetry could not find a pyproject.toml file in <ROOT> or its parents
```

Is there a way to run the pre-commit from the `api` folder?

gnkow:
Closing due to duplicate https://github.com/python-poetry/poetry/issues/7239

rubenvdg:
Actually, this issue rightfully mentions that `poetry-lock` hook doesn't work if your poetry project is not in the root of your project and https://github.com/python-poetry/poetry/issues/7239 does not mention that. Therefore, I think this is not a duplicate (but a superset of https://github.com/python-poetry/poetry/issues/7239).

Maybe it'd be good to re-open?

xinbinhuang:
I think the problem is that the current the `.pre-commit-config.yaml` hardcoded the `files: ^poetry.lock$`, which means that it'll only trigger the hook when the files are in the root directory. I created a PR to fix it: #7247

PR #7295, opened 2023-01-04
Pass all `pyproject.toml` and `poetry.lock` from subdirectories

xinbinhuang:
relates: https://github.com/python-poetry/poetry/issues/7239  #7247

This change pass all `pyproject.toml` and `poetry.lock` from subdirectories. In a monorepo setup, `pyproject.toml` and `poetry.lock` may not exists in the root folder. Since users need to configure the `--directory` flag to control which folder to run the check on, I'll be easier if the hook by default supply all files so users don't need to configure bot the flag and the `files` field. Also, the default hardcoded `files` is a bit hidden and hard to find without looking at the source code.

# Pull Request Check List

### DIFF
--- a/.pre-commit-hooks.yaml
+++ b/.pre-commit-hooks.yaml
@@ -5,7 +5,7 @@
   language: python
   language_version: python3
   pass_filenames: false
-  files: ^pyproject.toml$
+  files: .*/pyproject.toml$
 
 - id: poetry-lock
   name: poetry-lock
@@ -22,5 +22,5 @@
   language: python
   language_version: python3
   pass_filenames: false
-  files: ^poetry.lock$
+  files: .*/poetry.lock$
   args: ["-f", "requirements.txt", "-o", "requirements.txt"]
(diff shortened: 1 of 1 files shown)


neersighted:
Partial duplicate of #7242, should refer to #7239 as the target issue to close.

xinbinhuang:
Hi @neersighted, I believe #7239 is for a different thing - poetry check doesn't respect the `--directory` flag. But this is about pre-commit hook misconfiguration

neersighted:
It's for the same thing, the pre-commit hook. The fact that `--check` doesn't respect the global option was discovered there, but is not what the issue originally tracked (as indicated by the issue title).

xinbinhuang:
> neersighted

sure thing. I've updated the description to link to that issue, but I don't think this PR alone is enough to close it. Because `poetry check` still can't work with sub-directories.

xinbinhuang:
ah that makes sense. Can you help review this PR? are there any concerns on merging it or let me know if you have any comment you want me to address.

PR #7304, opened 2023-01-05
pkginfo has published types, fix resulting errors

dimbleby:
The latest version of pkginfo includes type stubs

This reveals that the `platform` on poetry's `PackageInfo` was mis-typed: sometimes it's a `str | None`, but when set via information from `pkginfo` it's actually a `Sequence[str]`

This has caused no real problems because the field isn't used anywhere... so I've just removed the dead code.

In re-locking I've also taken a shellingham post-release: regular 1.5.0 is yanked.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -63,7 +63,7 @@ lockfile = "^0.12.2"
 # packaging uses calver, so version is unclamped
 packaging = ">=20.4"
 pexpect = "^4.7.0"
-pkginfo = "^1.5"
+pkginfo = "^1.9.4"
 platformdirs = "^2.5.2"
 requests = "^2.18"
 requests-toolbelt = ">=0.9.1,<0.11.0"
@@ -174,7 +174,6 @@ module = [
   'cachecontrol.*',
   'lockfile.*',
   'pexpect.*',
-  'pkginfo.*',
   'requests_toolbelt.*',
   'shellingham.*',
   'virtualenv.*',
--- a/src/poetry/repositories/cached_repository.py
+++ b/src/poetry/repositories/cached_repository.py
@@ -22,7 +22,7 @@ if TYPE_CHECKING:
 
 
 class CachedRepository(Repository, ABC):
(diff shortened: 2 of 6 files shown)


dimbleby:
Although "platform" is unused, it was written to cache.  So I've bumped the cache version: else when the new code encounters an old cache it won't know what to do with the unexpected "platform" value.

(Alternatively could have left the cache version alone and explicitly dropped unexpected "platform" values.  That has the advantage of re-using existing caches - but seems a little less honest, the cache format really has changed)

pyproject.toml
```diff
@@ -63,7 +63,7 @@ lockfile = "^0.12.2"
 # packaging uses calver, so version is unclamped
 packaging = ">=20.4"
 pexpect = "^4.7.0"
-pkginfo = "^1.5"
+pkginfo = "^1.9.4"
```
Secrus, line 66: You forgot to remove `pkginfo` entry from mypy ignore list (line 177)

src/poetry/repositories/cached_repository.py
```diff
@@ -22,7 +22,7 @@
 
 
 class CachedRepository(Repository, ABC):
-    CACHE_VERSION = parse_constraint("1.1.0")
+    CACHE_VERSION = parse_constraint("2.0.0")
```
Secrus, line 25: Could you add info about this to your original PR description? It would make it easier for us to put an info about this in release notes (I assume people may want to remove their old caches)

Secrus (changes requested):

src/poetry/repositories/cached_repository.py
dimbleby: I wouldn't expect to need to spell this out in the release note.  

The last time this was changed was with the addition of support for yanked releases, it wasn't called out then.

Secrus: Fine then.

Secrus (approved):

PR #7305, opened 2023-01-05
don't set package pretty_version

dimbleby:
The pretty_version on the Package is pointless, it can always just be taken from the regular version's `.text`.

This is the only place in the codebase that sets it; and clearly it's not doing anything useful.

This is a prelude to corresponding tidying in poetry-core

### DIFF
--- a/src/poetry/packages/locker.py
+++ b/src/poetry/packages/locker.py
@@ -119,7 +119,6 @@ class Locker:
             package = Package(
                 name,
                 info["version"],
-                info["version"],
                 source_type=source_type,
                 source_url=url,
                 source_reference=source.get("reference"),


dimbleby:
test failure looks spurious:
```
      � Installing cleo (0.6.5)
  +   � Installing tomlkit (0.5.5)
      � Installing pendulum (1.4.4)
  -   � Installing tomlkit (0.5.5)
      � Installing demo (0.1.2 9cf87a2)
```
hard to believe that this MR changes that but only on Windows 3.10

(also this is clearly an over-specified test, though I'm not proposing to do anything about that since I assume the problem will go away on retry)

radoering (approved):

PR #6471, opened 2022-09-10
Use filelock to lock the file cache

dimbleby:
Fixes #6030, which I suspect is a significant contributor to "poetry hangs" complaints.

`cachecontrol` uses `lockfile` to create lock files.  In the event that poetry exits while one of those files exists (eg because the impatient user goes Ctrl-C) the lockfile is not removed.  That causes subsequent runs to block when trying to download the thing that it had been in the middle of dowloading.

`FileLock` from the `filelock` package does not have this undesirable property, when the process exits the lock is released and the next run _can_ acquire it.

(`filelock` was already in the poetry dependency tree via `virtualenv`). 

Note to anyone who tests this: `filelock` also doesn't remove the `.lock` file when it is done.  So if you test this branch and then revert to a version of poetry using `lockfile`, you will experience the original issue.  You'll want to do something like `find ~/.cache/pypoetry -name '*.lock' -delete` to get going again.

PR #7310, opened 2023-01-06
Drop lockfile as direct dependency

Secrus:
# Pull Request Check List

Direct dependency on `lockfile` wasn't really necessary, I have brought all the necessary code into one class building upon what was done in #6471

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -1798,4 +1798,4 @@ testing = ["flake8 (<5)", "func-timeout", "jaraco.functools", "jaraco.itertools"
 [metadata]
 lock-version = "2.0"
 python-versions = "^3.7"
-content-hash = "bbc71952916c7ee1c404e4a85e611189347955d63d5560b5bdb23073a07afde6"
+content-hash = "fc092106a32b55aa2ca36ebb60a1e7281440dbf6f3f0489ec75f5ac3f773695d"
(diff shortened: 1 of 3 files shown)


dimbleby:
What's the motivation?

What I mean is: given that cachecontrol is pulling in the `lockfile` dependency anyway, does this buy anything?  Or does it just take some messy code that we're indirectly using anyway, and duplicate it but make it also poetry's responsibility?

If and when cachecontrol ever (re)releases the version that includes my MR to drop `lockfile`, the code here should be removed altogether.  But I'm not seeing a future in which poetry does have this code, and also can actually eliminate `lockfile` from the dependency tree.

Secrus:
Actually, I was looking into the `cachecontrol` code, and having this, we could hack the `lockfile` extra away, since `cachecontrol` only uses `lockfile` if lock class wasn't provided in `FileCache` constructor.

neersighted:
I think this is good. I need to finish my draft of the writeup for the issue that caused 1.3.1 -- but in short, we can't reliably depend on cachecontrol to pull in lockfile, but it's silly just to pull it in to inherit from the class, when it's so minimal.

In short: not every index supports yanked packages and I'd like to see this in the codebase so we are indifferent to the presence of `lockfile`.

dimbleby:
To my mind, simply removing a top-level dependency isn't achieving anything very useful..  

And given that for the moment `lockfile` is an unavoidable indirect dependency, it seems to me more reasonable to use it than to duplicate it.

Still I suppose it's not worth fighting over a small amount of noise.

dimbleby:
On reflection I think I ought to be firmer in saying what I think about this: which is that it seems pointless and that I am against it.

Of course y'all are welcome to do what you think is right regardless: but it seems silly to have an opinion and not so say so.

PR #7312, opened 2023-01-06
Update workflows

Secrus:
# Pull Request Check List

I have noticed that the release workflow uses an older script contained in this repository, which is no longer being updated.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -19,7 +19,8 @@ jobs:
           python-version: "3.10"
 
       - name: Install Poetry
-        run: python install-poetry.py -y
+        run: |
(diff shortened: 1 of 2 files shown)


.github/workflows/release.yml
```diff
@@ -19,7 +19,8 @@ jobs:
           python-version: "3.10"
 
       - name: Install Poetry
-        run: python install-poetry.py -y
+        run: |
+          curl -sL https://install.python-poetry.org | python - -y
```
radoering, line 23: In poetry-core we use `curl -sSL`. Makes sense, doesn't it?

Secrus: Sure, this was missing in the main workflow as well, so added it there too.

radoering (approved):

PR #7339, opened 2023-01-11
Add poetry version used to generate lock file

pfouque:
# Pull Request Check List

Resolves: https://github.com/python-poetry/poetry/pull/2773#issuecomment-1353657313

Like in this comment, we would like to ensure the committed lock file has been generated with a specific version.

Following #2773 This PR simply add the poetry version used to generate the lock file (with the same fingerprint than `poetry --version`)

From
```
# This file is automatically @generated by Poetry and should not be changed by hand.
```
To
```
# This file is automatically @generated by Poetry (version 1.3.1) and should not be changed by hand.
```

### DIFF
--- a/src/poetry/packages/locker.py
+++ b/src/poetry/packages/locker.py
@@ -25,6 +25,7 @@ from tomlkit import document
 from tomlkit import inline_table
 from tomlkit import table
 
+from poetry.__version__ import __version__
 from poetry.utils._compat import tomllib
 
 
@@ -40,8 +41,8 @@ if TYPE_CHECKING:
 logger = logging.getLogger(__name__)
 _GENERATED_IDENTIFIER = "@" + "generated"
 GENERATED_COMMENT = (
-    f"This file is automatically {_GENERATED_IDENTIFIER} by Poetry and should not be"
-    " changed by hand."
+    f"This file is automatically {_GENERATED_IDENTIFIER} by Poetry"
+    f" {__version__} and should not be changed by hand."
 )
 
 
--- a/tests/packages/test_locker.py
+++ b/tests/packages/test_locker.py
@@ -17,6 +17,7 @@ from poetry.core.constraints.version import Version
 from poetry.core.packages.package import Package
 from poetry.core.packages.project_package import ProjectPackage
 
+from poetry.__version__ import __version__
 from poetry.factory import Factory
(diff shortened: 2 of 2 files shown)


radoering:
It might make more sense to add it in the metadata section so it can be read more easily by other tools?

radoering:
Thinking about it again, the advantage of the comment is that the lock file is not rewritten if there are no changes in the actual content. So there are no unnecessary diffs.

pfouque:
@radoering metadata is a great idea! 
There is already `python-version` and `lock-version`, adding `poetry-version` in-between would make a lot of sense...

neersighted:
I am a big -1 on `poetry-version` -- the reason being that there are two uses for this information. The first is diagnostic/informational; a comment here is better to avoid conflicts and unnecessary churn, as well as to discourage the second use.

The second category of users want to infer things about the contents of the lock file based on the Poetry version. That is in correct, and is in fact what the lock file version aims to solve. The solution here is to start writing down documentation on the lock file, which also helps explain when things change despite no bump (usually due to bug fixes in Poetry).

src/poetry/packages/locker.py
```diff
@@ -40,8 +41,8 @@
 logger = logging.getLogger(__name__)
 _GENERATED_IDENTIFIER = "@" + "generated"
 GENERATED_COMMENT = (
-    f"This file is automatically {_GENERATED_IDENTIFIER} by Poetry and should not be"
-    " changed by hand."
+    f"This file is automatically {_GENERATED_IDENTIFIER} by Poetry (version"
+    f" {__version__}) and should not be changed by hand."
```
radoering, line 45:
```suggestion
    f"This file is automatically {_GENERATED_IDENTIFIER} by Poetry"
    f" {__version__} and should not be changed by hand."
```

Should be sufficient and is consistent with what we write into the INSTALLER file in #6205.

pfouque:
@neersighted After some tests, I understand your point about "lock-version" and I agree it doesn't bring much value in metadata.
Thanks @radoering, I do not have a strong opinion on the exact format, my proposition was based on the output of "poetry --version", but whatever the format is, the aim is to give access to this information for diagnostic.

radoering (approved):

pfouque:
🙏 Thanks!

PR #7341, opened 2023-01-11
Docs: whitelist_externals was replaced by allowlist_externals

jkawamoto:
tox replaced `whitelist_externals` with `allowlist_externals` and removed `whitelist_externals`.

This PR fixes the docs so that it uses the new option.

See also: https://tox.wiki/en/3.18.1/changelog.html#v3-18-0-2020-07-23

### DIFF
--- a/docs/faq.md
+++ b/docs/faq.md
@@ -81,7 +81,7 @@ Thus, dependencies are resolved by `pip`.
 isolated_build = true
 
 [testenv]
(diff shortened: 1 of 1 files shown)


jkawamoto:
It looks like it was fixed in #4574, but somehow docs still use the old option.

Secrus (approved):

Issue #5003, opened 2022-01-07
option to fail if `poetry.lock` doesn't match `pyproject.toml`


DetachHead:
- [X] I have searched the [issues](https://github.com/python-poetry/poetry/issues) of this repo and believe that this is not a duplicate.
- [X] I have searched the [documentation](https://python-poetry.org/docs/) and believe that my question is not covered.
## steps
1: add dependencies and run `poetry update`
2: change dependencies and run `poetry install`
## expected
The command fails as you will install dependencies that are not reflecting the project configuration

## actual
All you get is a warning
## Feature Request
currently when you run `poetry install` with an outdated lock file, you get this warning:
```
Warning: The lock file is not up to date with the latest changes in pyproject.toml. You may be getting outdated dependencies. Run update to update them.
```
it would be nice if there was an option to make this an error instead of a warning. in my case i want the CI to fail if `poetry.lock` is outdated. 

the warning can be easily missed when running in CI

glencairn:
In my opinion, this feature would be counterintuitive and it wouldn't add much value. Lock files are used as the ultimate source of truth when installing/updating your project's dependencies. The fact that a lock file exists in your project, and that it's contents contain older or outdated versions of dependencies is perfectly valid and should never be considered an error. I think adding a command line option to remove the assertions made by a lock file would be confusing.

The way I see it, poetry already gives you the tools to figure out if your locked environment is out of date. For example, you could update your CI to run `poetry show --outdated` and capture the standard output. If it's empty, you're good to go. If it's not, then you have outdated packages.

DetachHead:
> In my opinion, this feature would be counterintuitive and it wouldn't add much value.

I disagree. If I change/add a dependency in `pyproject.toml` and forget to run `poetry update`, I’ll experience unexpected behaviour where the cause won’t be immediately obvious. Other similar tools such as npm and gradle fail when the lock file is outdated

> I think adding a command line option to remove the assertions made by a lock file would be confusing

I’m not asking for an option to remove the lock file’s assertions, only an option to fail if `pyproject.toml` contradicts those assertions

glencairn:
No, if you choose to commit your lock file this is not unexpected behavior and it's certainly not an error. As per the [documentation](https://python-poetry.org/docs/basic-usage/#installing-with-poetrylock)...

[...]

PR #7350, opened 2023-01-13
Add --check-lock flag to poetry install

ashokdelphia:
Add a flag to allow aborting installation if the lock file is out of sync.

I'd like to ensure that when I'm using 'poetry install' in a CI pipeline that the lock file is up to date.

This is possible at present by using:

    poetry lock --check && poetry install ...

But I think it's a common enough use that providing it within the install command is still useful.

# Pull Request Check List

Resolves: #5003.


radoering:
Superseded by #8737. With Poetry 1.8, this is default behavior now.

PR #7356, opened 2023-01-15
installer: remove old installer and config setting `experimental.new-installer`

radoering:
Since the current default installer is stable and we will introduce a new default installer in #6205, there is no reason to keep the old installer anymore.

In other words, the setting `experimental.new-installer` (default: true) is removed. Despite the setting's name the "new installer" has not been experimental anymore for a while now.

Update: The old installer will be officially deprecated in 1.4, so it will be not be removed before 1.5.


SadPencil:
Please don't remove the old installer until https://github.com/python-poetry/poetry/issues/1632 is solved. This is the only nice workaround to the issue.

dimbleby:
upgrading virtualenv (and therefore also platformdirs #7640) is blocked on completing this deprecation.  IMO we should get this removal completed sooner rather than later.

Folk who are relying on the old installer can either use old versions of poetry - or (even better) will finally be motivated to submit a merge request making the new installer work for them

radoering:
> Please don't remove the old installer until #1632 is solved. Using `pip` is the only nice workaround to the issue.

I'm sorry but as it seems, we can't afford anymore to keep the old installer as it blocks dependency updates, which fix relevant bugs.

> Folk who are relying on the old installer can either use old versions of poetry - or (even better) will finally be motivated to submit a merge request making the new installer work for them

In addition, there seems to be a plugin that may already cover this use case: https://github.com/arcesium/poetry-plugin-pypi-mirror

Secrus (approved):

PR #7358, opened 2023-01-15
installer: deprecate old installer (setting `experimental.new-installer` to false)

radoering:
Maybe, we should deprecate the old installer officially (we didn't announce it, did we?) or at least make the deprecation more prominent by printing a warning before removing it some minor releases later in #7356.


neersighted (approved):

dlqqq:
It seems that it was known in advance that this would break compatibility when installing certain packages. In the future, can we please refrain from doing so without bumping the major version? Releasing breaking changes without bumping the major version damages trust with users. Yes, I understand technically the specification validates the current approach, but the fact is that a significant number of packages just became uninstallable overnight.

radoering:
The deprecation of the old installer (this PR) is probably not breaking anything (because it's only printing a deprecation warning). The installer deprecated in this PR was **not** the default installer in poetry 1.3, 1.2 or 1.1.

Poetry 1.1 -1.3 has two installers, poetry 1.4 has three:

- deprecated old installer (`experimental.new-installer=false`)
- previous default installer (`experimental.new-installer=true`, `installer.modern-installation=false`)
- new "modern" installer (`experimental.new-installer=true`, `installer.modern-installation=true`)

I assume you are referring to the introduction of the new modern installer. Before the 1.4 release, we tested the new installer with many packages and were able to install all of them. Thus, it was not known in advance.

You can always switch back to the previous default installer, which is not deprecated, by setting `installer.modern-installation` to `false`.

PR #7359, opened 2023-01-15
Spread envs into separate modules

Secrus:
# Pull Request Check List

A starting point for much-needed refactoring of envs handling. There are no logic changes, and everything should be 100% backward compatible.


tests/utils/test_env.py
```diff
@@ -876,7 +876,7 @@ def err_on_rm_venv_only(path: Path | str, *args: Any, **kwargs: Any) -> None:
             remove_directory(path)
 
     m = mocker.patch(
-        "poetry.utils.env.remove_directory", side_effect=err_on_rm_venv_only
```
neersighted, line 879: This should probably just mock `poetry.utils.helpers.remove_directory` as that is where the method actually comes from.

neersighted (dismissed):
Overall looks like a good direction, though we could split things further into `env.base`, `env.generic`, etc etc if we want to try and increase clarity further.

I also think `env.shared` should probably be called `env.stringpy` or something more descriptive.

tests/utils/test_env.py
Secrus: I have tried that, but if I do so, the test fails with `FAILED tests/utils/test_env.py::test_remove_keeps_dir_if_not_deleteable - AssertionError: remove_directory(PosixPath('/tmp/poetry_1s94p2i5/simple-project-qCH79UIk-py3.6')) call not found`, didn't really know how to properly debug this, so decided to leave it as it is.

Secrus:
@neersighted I renamed modules, is that what you meant?

src/poetry/utils/env/generic_env.py
```diff
@@ -0,0 +1,102 @@
...
+        if not self._is_windows:
+            return os.execvpe(command[0], command, env=env)
+
+        exe = subprocess.Popen(command, env=env, **kwargs)
+        exe.communicate()
+
+        return exe.returncode
+
+    def _run(self, cmd: list[str], **kwargs: Any) -> str:
+        return super()._run(cmd, **kwargs)
```
radoering, line 99: Changing this from `return super(VirtualEnv, self)._run(cmd, **kwargs)` is a behavioral change. Before we called `Env._run()`, now we are calling `VirtualEnv._run()`. Are you sure that's ok? I'm not...

src/poetry/utils/env/__init__.py
```diff
@@ -0,0 +1,129 @@
...
+    "GET_SYS_PATH",
+    "GET_SYS_TAGS",
+    "GET_ENV_PATH_ONELINER",
+    "GET_PYTHON_VERSION_ONELINER",
+    "GET_PATHS_FOR_GENERIC_ENVS",
+    "EnvCommandError",
+    "IncorrectEnvError",
+    "InvalidCurrentPythonVersionError",
+    "NoCompatiblePythonVersionFound",
+    "PythonVersionNotFound",
```
radoering, line 118: `EnvError` is missing.

radoering (changes requested):

src/poetry/utils/env/generic_env.py
Secrus: yeah, reverted, I thought it was leftover after dropping Py2 support.

src/poetry/utils/env/__init__.py
Secrus: Added

radoering (approved):

PR #7369, opened 2023-01-19
drop flatdict dependency

martin-kokos:
# Pull Request Check List

In the spirit of dependency cleanup #5720 this PR replaces the `flatdict` dependency with a new function. Flatdict doesn't appear to be maintained and was used only once (in a test). I also believe poetry being a package manager should have as little dependencies as possible.
I am trying to get poetry included in the package repository of Gentoo Linux, and not having to include the `flatdict` dependency would be really nice.

### DIFF
--- a/tests/helpers.py
+++ b/tests/helpers.py
@@ -9,7 +9,6 @@ import urllib.parse
 
 from pathlib import Path
 from typing import TYPE_CHECKING
-from typing import Any
 
 from poetry.core.packages.package import Package
 from poetry.core.packages.utils.link import Link
@@ -28,6 +27,8 @@ from poetry.utils._compat import metadata
 
 if TYPE_CHECKING:
     from collections.abc import Iterator
+    from typing import Any
+    from typing import Mapping
 
     from poetry.core.constraints.version import Version
     from poetry.core.packages.dependency import Dependency
@@ -284,3 +285,33 @@ def mock_metadata_entry_points(
         "entry_points",
         return_value=[make_entry_point_from_plugin(name, cls, dist)],
     )
+
+
+def flatten_dict(obj: Mapping[str, Any], delimiter: str = ".") -> Mapping[str, Any]:
+    """
+    Flatten a nested dict.
+
+    A flatdict replacement.
+
+    :param obj: A nested dict to be flattened
+    :delimiter str: A delimiter used in the key path
+    :return: Flattened dict
+    """
+
+    def recurse_keys(obj: Mapping[str, Any]) -> Iterator[tuple[list[str], Any]]:
+        """
+        A recursive generator to yield key paths and their values
+
+        :param obj: A nested dict to be flattened
+        :return:  dict
+        """
+        if isinstance(obj, dict):
+            for key in obj.keys():
+                for leaf in recurse_keys(obj[key]):
+                    leaf_path, leaf_value = leaf
+                    leaf_path.insert(0, key)
+                    yield (leaf_path, leaf_value)
+        else:
+            yield ([], obj)
+
+    return {delimiter.join(path): value for path, value in recurse_keys(obj)}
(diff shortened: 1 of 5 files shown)


src/poetry/utils/flatdict.py
```diff
@@ -0,0 +1,32 @@
+from __future__ import annotations
+
+from typing import Any
+from collections.abc import Mapping
+
+
+def flatten_dict(obj: Mapping[str, Any], delimiter: str=".") -> Mapping[str, Any]:
```
dimbleby, line 7: apart from the typing errors - per the pipeline - this doesn't belong in src, it's used in unit test only.

(for the same reason, end users never saw this dependency anyway - which isn't anything against this merge request exactly; but does say that if getting this clean is causing you much trouble, it may not be worth too much effort)

martin-kokos: Good point. Where would you suggest I put it. test/helpers ?

Secrus: Yes, tests/helpers is a much better choice.

tests/utils/test_flatdict.py
```diff
@@ -0,0 +1,24 @@
+from __future__ import annotations
+
+from poetry.utils.flatdict import flatten_dict
+
+
+def test_flatten_dict() -> None:
```
Secrus, line 6: since we are keeping the function, I would like to see more tests (different separators for example). Just because we only use `.` for now, doesn't mean we won't use the full capabilities of the function in the future.

Secrus (changes requested):

tests/utils/test_flatdict.py
martin-kokos: The test does use a non-default delimiter `:`

src/poetry/utils/flatdict.py
radoering: Now, it's duplicated.

neersighted (changes requested):
LGTM with a relock (and satisfying pre-commit, of course).

Secrus:
Also requires fixes from pre-commit

martin-kokos:
> LGTM with a relock (and satisfying pre-commit, of course).

I've fixed outstanding pre-commit hook issues. I also did run `poetry lock --no-updates`, if that is what you meant.

neersighted:
Sorry, when I said relock, I meant 'rebase onto master, resolving the resulting conflict by locking' -- it tends to be shorthand for the regular contributors, but I realize that was rather opaque 😆

martin-kokos:
Please note, in `tests/helpers.py` I've have moved the types into the `if TYPE_CHECKING` section. Branch was rebased to master and did `poetry lock --no-updates` again.

tests/helpers.py
```diff
@@ -284,3 +285,33 @@ def mock_metadata_entry_points(
...
+    A flatdict replacement.
+
+    :param obj: A nested dict to be flattened
+    :delimiter str: A delimiter used in key path
+    :return: Flattened dict
+    """
+
+    def recurse_keys(obj: Mapping[str, Any]) -> Iterator[tuple[list[str], Any]]:
+        """
+        A recursive generater to yield key paths and their values
```
neersighted, line 303:
```suggestion
        A recursive generator to yield key paths and their values
```

tests/helpers.py
```diff
@@ -284,3 +285,33 @@ def mock_metadata_entry_points(
...
+
+
+def flatten_dict(obj: Mapping[str, Any], delimiter: str = ".") -> Mapping[str, Any]:
+    """
+    Flatten a nested dict.
+
+    A flatdict replacement.
+
+    :param obj: A nested dict to be flattened
+    :delimiter str: A delimiter used in key path
```
neersighted, line 297:
```suggestion
    :delimiter str: A delimiter used in the key path
```

neersighted (changes requested):
Found two minor nits, otherwise looks ready for merge.

Secrus (approved):

neersighted (approved):

martin-kokos:
Thanks for the patience.

Issue #7374, opened 2023-01-20
Root extra dependencies should be sorted


yeger00:
- **Poetry version**: 1.3.1 (tested also on 1.4.0.dev0)
- **Python version**: python3.10.8
- **OS version and name**: MacOS 12.5.1
- **pyproject.toml**: 

- [x] I am on the [latest](https://github.com/python-poetry/poetry/releases/latest) stable Poetry version, installed using a recommended method.
- [x] I have searched the [issues](https://github.com/python-poetry/poetry/issues) of this repo and believe that this is not a duplicate.
- [x] I have consulted the [FAQ](https://python-poetry.org/docs/faq/) and [blog](https://python-poetry.org/blog/) for any relevant entries or release notes.
- [ ] If an exception occurs when executing a command, I executed it again in debug mode (`-vvv` option) and have included the output below.

## Issue
Hello,
First of all, that you for your work in maintaining poetry.

It is recommended to commit to `poetry.lock` file to the repository [link](https://python-poetry.org/docs/basic-usage/#commit-your-poetrylock-file-to-version-control). Since this is the case, the lock file to be determine as possible, in order to be able to perform diff in a PR and have better control on what I'm merging.

Here is an example of updating one library and locking the file. Instead of 1 line of change I receive many changes of order.

![image](https://user-images.githubusercontent.com/4067567/213636148-a0a8ccac-7769-49d4-9019-37561d3de53e.png)

PR #7375, opened 2023-01-20
Order root extra dependencies

yeger00:
# Pull Request Check List

Resolves: #7374

- [ ] Updated **documentation** for changed code. - No need

Hello,
This PR solves issue [7374](https://github.com/python-poetry/poetry/issues/7374).
I added a test that validates the bug in the first commit, and fixed it in the 3rd commit.

Thanks,
Avi

### DIFF
--- a/src/poetry/packages/locker.py
+++ b/src/poetry/packages/locker.py
@@ -244,7 +244,7 @@ class Locker:
 
         if root.extras:
             lock["extras"] = {
-                extra: [dep.pretty_name for dep in sorted(deps, key=lambda dep: dep.pretty_name)]
+                extra: sorted(dep.pretty_name for dep in deps)
                 for extra, deps in sorted(root.extras.items())
             }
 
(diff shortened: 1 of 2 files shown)


src/poetry/packages/locker.py
```diff
@@ -244,7 +244,7 @@ def set_lock_data(self, root: Package, packages: list[Package]) -> bool:
 
         if root.extras:
             lock["extras"] = {
-                extra: [dep.pretty_name for dep in deps]
+                extra: [dep.pretty_name for dep in sorted(deps, key=lambda dep: dep.pretty_name)]
```
dimbleby, line 247: quibble: this feels like a complicated way of writing `sorted(dep.pretty_name for dep in deps)`

yeger00: Thanks for the feedback. I changed to your suggestion (and also fixed the flake error)

radoering (approved):

PR #7387, opened 2023-01-21
remove strange VENDORS handling in installed repository

dimbleby:
#7386 caused me to look harder, and not for the first time I find myself confused about what some code is trying to do...

So far as I can see, the code that I have removed hardcodes knowledge of poetry's own source code into its search for installed packages - regardless of the environment in which it is actually running.  And then the unit test patches that `_VENDORS` in a way that can't possibly happen outside of the unit test, allowing it to find the attrs package in the `installed` fixture's vendors directory.

I'm very open to having it explained to me that I have completely the wrong end of the stick.  But this seems to be really quite peculiar.

### DIFF
--- a/src/poetry/repositories/installed_repository.py
+++ b/src/poetry/repositories/installed_repository.py
@@ -20,15 +20,6 @@ if TYPE_CHECKING:
     from poetry.utils.env import Env
 
 
-_VENDORS = Path(__file__).parent.parent.joinpath("_vendor")
-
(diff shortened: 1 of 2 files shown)


neersighted:
This is a relic of the old install method where we packaged things up into bespoke tarballs, and no longer necessary since we install into a standard Python environment.

neersighted (approved):

Issue #7392, opened 2023-01-23
Ensure poetry.lock modification time is newer than pyproject.toml, after relevant commands


Kache:
- [x] I have searched the [issues](https://github.com/python-poetry/poetry/issues) of this repo and believe that this is not a duplicate.
- [x] I have searched the [FAQ](https://python-poetry.org/docs/faq/) and general [documentation](https://python-poetry.org/docs/) and believe that my question is not already covered.

## Feature Request

After doing operations like `add`/`remove`/`update` that both updates `pyproject.toml` and refreshes `poetry.lock`, please complete the operation leaving `poetry.lock`'s modification time to be newer than `pyproject.toml`'s.

Not only is this conceptually proper, it means Makefiles can be used to implement "auto-`poetry install` if needed" behavior, e.g.:
 * https://github.com/python-poetry/poetry/issues/75#issuecomment-399514444
 * https://medium.com/aigent/makefiles-for-python-and-beyond-5cf28349bf05

The comment in #75 suggests this already used to be the case at some point.

dimbleby:
As far as I know, poetry has always left updating `pyproject.toml` to last, waiting to check that the change actually succeeded.

The code is at https://github.com/python-poetry/poetry/blob/33242c22906ec84c7f526537bfd9a98415192d13/src/poetry/console/commands/add.py#L264-L266 (and similar for `poetry remove`)

I have no idea whether others are likely to agree with you that your proposed change is a good idea, but preparing a pull request that shuffle this code around is likely your best shot at getting it to happen.

Kache:
Would it really take shuffling code? I was thinking these operations could just finish off with a `touch poetry.lock` equivalent at the end.

dimbleby:
Sure, sounds plausible.  But still - if you want this to happen, recommend that submitting an MR is the most likely way.

Kache:
Threw together a quick draft, the change is rather straightforward, but there's a testing caveat: https://github.com/python-poetry/poetry/pull/7401

radoering:
Can you further explain what you want to derive from a pyproject.toml with a newer modification date than poetry.lock? Did you consider that poetry is not the only tool that uses the pyproject.toml?

If you only want to check if pyproject.toml and poetry.lock are in sync, you can run `poetry lock --check`.

Secrus:
I agree with @radoering here.

Kache:
[...]

Issue #75, opened 2018-04-30
Detect pyproject.toml changes and automaticaly run `poetry install` if needed


mtkennerly:
When using Stack (Haskell) and Cargo (Rust), if you manually add a new dependency in package.yaml or cargo.toml respectively, then the next invocation of `stack test` or `cargo test` will detect the change and install the new dependency, whereas `poetry run pytest` could fail on the dependency missing. It would be great if Poetry could auto-detect the change instead of having to manually run `poetry install` again. Of course, for an individual user, it's easy to go through the `poetry add` interface instead, but it's more problematic when working on a team since other people won't immediately know after `git pull` whether there were updates requiring a `poetry install`.

cauebs:
Maybe warn the user instead of automatically installing. But I agree that detecting changes would be useful.

jacebrowning:
I've achieved this behavior with a `Makefile`:

```
.PHONY: all
all: install

.PHONY: install
install: .venv/flag
.venv/flag: pyproject.lock
	@ poetry config settings.virtualenvs.in-project true
	poetry develop
	@ touch $@

pyproject.lock: pyproject.toml
	poetry lock

###############################################################################

.PHONY: test
test: install
	poetry run pytest
``` 

but I would be in favor of `poetry` doing this automatically.

cauebs:
This is more important than I initially thought, as it's causing some very confusing things, e.g.: #292.

+ How do we detect changes in `pyproject.toml`? Check the file's metadata for last modified time? Hash its contents?
+ Where do we store these stamps? In `pyproject.lock`
+ Then, whenever the lock file is used, that should be checked, right?

I can give this one a go, but I'll definitely need some pointers from @sdispater.

ranisalt:
> Check the file's metadata for last modified time?

This is faster and less resource-consuming, but there are known issues when depending on modification times.

I think that just as pyproject contains hashes to deps, a hash to pyproject.toml content is fine too.

sdispater:
Poetry already tracks changes and discrepancies between `pyproject.toml` and `pyproject.lock`. 
That's why you can sometimes see a warning when executing `install`.

Here is the corresponding line: https://github.com/sdispater/poetry/blob/master/poetry/installation/installer.py#L164

A similar approach could be used elsewhere if needed.

And I think displaying a warning (like in `install`) is preferable instead of installing automatically.

cauebs:
Well, then why didn't he get a warning in #292?

[...]

PR #7401, opened 2023-01-24
feat: Support "automated install" usage via `make`

Kache:
A common annoyance when managing deps is having long sessions of
repeated iteration loops of `add`, `remove`, tweaking `pyproject.toml` +
`lock --no-update` + `install --sync`, re-running tests, trying to find
a configuration that works.

GNU `make` can implement "automated install" behavior, simplifying the
loop to: "save any change, `make`, repeat if needed". Consider:

```make
VENV := .venv
PYTHON_VERSION := $(shell cat .python-version)

PYTHON_VERSIONS_PATH ?= ~/.pyenv/versions
PYTHON_INSTALL_CMD ?= pyenv install

.PHONY: test
test: $(VENV)/install.sentinel  # tests depend on venv state
	poetry run pytest

$(VENV)/install.sentinel: poetry.lock $(VENV)/bin/python  # venv state depends on poetry.lock and venv
	poetry install --sync
	@ touch $@

.poetry.lock: pyproject.toml  # poetry.lock depends on pyproject.toml
	poetry lock --check && touch poetry.lock || poetry lock --no-update

$(VENV)/bin/python: $(PYTHON_VERSIONS_PATH)/$(PYTHON_VERSION)/bin/python
	@ [[ "$(shell python --version)" = "Python $(PYTHON_VERSION)" ]] || { \
		echo 'Expected Python $(PYTHON_VERSION) in $$PATH' >& 2; \
		exit 1; \
	}
	poetry env use python

$(PYTHON_VERSIONS_PATH)/$(PYTHON_VERSION)/bin/python:
	$(PYTHON_INSTALL_CMD) $(PYTHON_VERSION)
```

This greatly simplifies handling many workflow states. In all cases:

 * The lockfile changed and venv needs to sync
 * I just freshly downloaded the project
 * The required Python version isn't installed
 * I want to run tests
 * I changed pyproject.toml

Just run `make`.

These changes to `add` and `remove` speed up the "dep search iteration
loop" by avoiding unnecessary `lock --check` and `lock --no-update`
calls after changing `pyproject.toml`

Resolves: #7392

# Pull Request Check List

Resolves: #7392

### DIFF
--- a/src/poetry/console/commands/add.py
+++ b/src/poetry/console/commands/add.py
@@ -241,10 +241,7 @@ The add command adds required packages to your <comment>pyproject.toml</> and in
                 )
             )
 
-        # Refresh the locker
-        self.poetry.set_locker(
-            self.poetry.locker.__class__(self.poetry.locker.lock, poetry_content)
-        )
+        self.poetry.locker.refresh(poetry_content)
         self.installer.set_locker(self.poetry.locker)
 
         # Cosmetic new line
@@ -264,6 +261,7 @@ The add command adds required packages to your <comment>pyproject.toml</> and in
         if status == 0 and not self.option("dry-run"):
             assert isinstance(content, TOMLDocument)
             self.poetry.file.write(content)
+            self.installer.touch_lockfile()
 
         return status
 
--- a/src/poetry/console/commands/remove.py
+++ b/src/poetry/console/commands/remove.py
@@ -105,10 +105,7 @@ list of installed packages
                 "The following packages were not found: " + ", ".join(sorted(not_found))
             )
 
-        # Refresh the locker
-        self.poetry.set_locker(
-            self.poetry.locker.__class__(self.poetry.locker.lock, poetry_content)
-        )
+        self.poetry.locker.refresh(poetry_content)
         self.installer.set_locker(self.poetry.locker)
 
         self.installer.set_package(self.poetry.package)
@@ -122,6 +119,7 @@ list of installed packages
         if not self.option("dry-run") and status == 0:
             assert isinstance(content, TOMLDocument)
             self.poetry.file.write(content)
+            self.installer.touch_lockfile()
 
         return status
 
--- a/src/poetry/installation/installer.py
+++ b/src/poetry/installation/installer.py
@@ -158,6 +158,9 @@ class Installer:
 
         return self
 
+    def touch_lockfile(self) -> None:
+        self._locker.lock.touch()
+
     def is_updating(self) -> bool:
         return self._update
 
--- a/src/poetry/packages/locker.py
+++ b/src/poetry/packages/locker.py
@@ -9,6 +9,7 @@ from hashlib import sha256
 from pathlib import Path
 from typing import TYPE_CHECKING
 from typing import Any
+from typing import TypeVar
 from typing import cast
 
 from packaging.utils import canonicalize_name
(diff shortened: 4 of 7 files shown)


src/poetry/console/commands/version.py
```diff
@@ -8,14 +8,14 @@
...
-from poetry.console.commands.command import Command
+from poetry.console.commands.installer_command import InstallerCommand
 
 
 if TYPE_CHECKING:
     from poetry.core.constraints.version import Version
 
 
-class VersionCommand(Command):
+class VersionCommand(InstallerCommand):
```
radoering, line 18: I don't like that. `InstallerCommand` requires the creation of a virtualenv. We shouldn't make commands `InstallerCommand` if it's not required. Is touching the lock file really required here to cover your use case? (Bumping the version does not change the lock file.)

Kache: This feature is about keeping `poetry.lock`'s modification time newer than `pyproject.toml` so that external tooling like Make is able to discern whether or not the lockfile is out of date. https://github.com/python-poetry/poetry/issues/7392 explains the use case.

Even though the lockfile contents aren't touched here, pyproject.toml is, meaning that Make would think the lockfile is out of date.

I suppose `VersionCommand` could alternatively write to `pyproject.toml` without changing its modificaiton time (perhaps by resetting it)?

radoering: > I suppose `VersionCommand` could alternatively write to `pyproject.toml` without changing its modificaiton time (perhaps by resetting it)?

That seems also wrong to me. `VersionCommand` changes pyproject.toml and thus should change the modification date of pyproject.toml. However, it doesn't change the lock file and thus shouldn't change the modification date of the lock file. Adding quirks to suffice the `make` use case doesn't seem very attractable to me.

Kache: I'm up for only applying this to `add` and `remove` -- I'll make the change.

Any thoughts on testing? This [section in `commands/add.py`](https://github.com/python-poetry/poetry/blob/85ef11bc6f0cc0dad9c630351688ae4b98ea34be/src/poetry/console/commands/add.py#L244-L248) makes it "impossible" for `TestLocker` to have `self._write == True` (to materialize test lockfile to disk & direct test modification time).

Kache:
@dimbleby @radoering I've added tests, would you please take a look?

Also:
 * Is there any documentation that needs updating?
 * I'll be interested in backporting this to onto `1.2`, anything I should know?

neersighted:
We're not accepting backports against 1.2, and new features are not likely to ever get backports -- so this would go into 1.4, which is imminent/currently in the release process.

Kache:
> We're not accepting backports against 1.2, and new features are not likely to ever get backports

😞 Any chance this could be considered a "bug" b/c https://github.com/python-poetry/poetry/issues/75#issuecomment-399514444 seems to suggest this behavior used to be the case, long ago?

Secrus:
> > We're not accepting backports against 1.2, and new features are not likely to ever get backports
> 
> disappointed Any chance this could be considered a "bug" b/c [#75 (comment)](https://github.com/python-poetry/poetry/issues/75#issuecomment-399514444) seems to suggest this behavior used to be the case, long ago?

This is far from being considered a bug, and the issue you are pointing at is also a "feature request", so the answer is no.

PR #7409, opened 2023-01-25
Use py.exe to find python on Windows

bruchar1:
On Windows, python installs the py.exe launcher as a reliable way to find a given installed version of python.

This PR adds the possibillity to specify a python version (e.g. `poetry env use 3.10`) on Windows.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,5 @@
 *.pyc
+__pycache__/
 
 # Packages
 *.egg
--- a/src/poetry/utils/env.py
+++ b/src/poetry/utils/env.py
@@ -522,12 +522,18 @@ class EnvManager:
         self._io = io or NullIO()
 
     def _full_python_path(self, python: str) -> str:
+        command = [python, "-c", '"import sys; print(sys.executable)"']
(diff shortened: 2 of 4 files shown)


neersighted:
We tried this once, I do not recall if it was merged and reverted or simply never merged. The takeaway was that pythonfinder or something similar (pure Python) should be used instead of `py.exe` which really designed for our use case.

cc @finswimmer @abn

bruchar1:
Fixes #3520. Related to #5778.

I'm really not sure what is bad about calling `py.exe`, since `poetry env use` already calls `python` using a subprocess. I only changed the name of the called executable.

I'm not against the use of `pythonfinder`, but imho, this solution is just simpler.

bruchar1:
Also related to #4217.

TBBle:
The prior PR for this approach that I could find is #4217. The comments seem to cover the main concerns with this approach, and was closed in favour of a pythonfinder-based approach, which appeared in #5778 but apparently never went anywhere.

Secrus:
Stalled

Issue #7408, opened 2023-01-25
test_run_has_helpful_error_when_command_not_found is locale dependent


bruchar1:
My Windows installation is in French. When I run tests, the test_run_has_helpful_error_when_command_not_found test fails because the output is in French, not is English, on my machine.

Detecting current language is probably not trivial.

Maybe the test should only ensure that `nonexistent-command` is in the output.

dimbleby:
probably a merge request fixing it would be welcome

PR #7411, opened 2023-01-25
Make test locale independent

bruchar1:
Resolves: #7408 

Just test that the command name is in the output. THis should be sufficient as we already tested the command result code.

### DIFF
--- a/tests/console/commands/test_run.py
+++ b/tests/console/commands/test_run.py
@@ -65,26 +65,23 @@ def test_run_keeps_options_passed_before_command(
 def test_run_has_helpful_error_when_command_not_found(
     app_tester: ApplicationTester, env: MockEnv, capfd: pytest.CaptureFixture[str]
 ):
+    nonexistent_command = "nonexistent-command"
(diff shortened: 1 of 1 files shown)


tests/console/commands/test_run.py
```diff
@@ -65,26 +65,23 @@ def test_run_keeps_options_passed_before_command(
...
 
-    assert env.executed == [["nonexistent-command"]]
+    assert env.executed == [[nonexistent_command]]
     assert app_tester.status_code == 1
     if WINDOWS:
         # On Windows we use a shell to run commands which provides its own error
         # message when a command is not found that is not captured by the
         # ApplicationTester but is captured by pytest, and we can access it via capfd.
         # The expected string in this assertion assumes Command Prompt (cmd.exe) is the
         # shell used.
```
radoering, line 79: The last sentence of this comment should be changed as well. Something like "The exact error message depends on the system language. Thus, we check only for the name of the command."

radoering (approved):

Issue #6934, opened 2022-11-01
Update tests to use `tmp_path` over `tmp_dir`


chadac:
As mentioned by @neersighted in https://github.com/python-poetry/poetry/pull/6929#issuecomment-1298611340, pytest now includes a fixture `tmp_path` that is of the `Path` type, which is preferred over the legacy `tmp_dir` of a `str` type.

There are only a few places which still use `tmp_dir`, so this should be doable in a single PR.

PR #7129, opened 2022-12-01
Removing tmp_dir test fixture in favor of pytest's tmp_path.

upsideon:
# Pull Request Check List

Resolves: #6934

These changes remove the custom `tmp_dir` test fixture in favor of pytest's built-in `tmp_path` fixture for creating temporary directories. Only test code has been updated, so there was no need to add additional tests or documentation.

- [x] Added **tests** for changed code.
- [ ] Updated **documentation** for changed code.

PR #7412, opened 2023-01-25
tests: stop using tmp_dir fixture in favor of pytest.tmp_path fixture

martin-kokos:
# Pull Request Check List

Resolves: #6934 

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -5,7 +5,6 @@ import os
 import re
 import shutil
 import sys
-import tempfile
 
 from contextlib import suppress
 from pathlib import Path
@@ -28,7 +27,6 @@ from poetry.repositories import RepositoryPool
 from poetry.utils.env import EnvManager
 from poetry.utils.env import SystemEnv
 from poetry.utils.env import VirtualEnv
-from poetry.utils.helpers import remove_directory
 from tests.helpers import MOCK_DEFAULT_GIT_REVISION
 from tests.helpers import TestLocker
 from tests.helpers import TestRepository
@@ -169,8 +167,8 @@ def with_chained_null_keyring(mocker: MockerFixture) -> None:
 
 
 @pytest.fixture
-def config_cache_dir(tmp_dir: str) -> Path:
-    path = Path(tmp_dir) / ".cache" / "pypoetry"
+def config_cache_dir(tmp_path: Path) -> Path:
+    path = tmp_path / ".cache" / "pypoetry"
     path.mkdir(parents=True)
     return path
 
@@ -219,8 +217,10 @@ def config(
 
 
 @pytest.fixture()
-def config_dir(tmp_dir: str) -> Path:
-    return Path(tempfile.mkdtemp(prefix="poetry_config_", dir=tmp_dir))
+def config_dir(tmp_path: Path) -> Path:
+    path = tmp_path / "config"
+    path.mkdir()
+    return path
 
 
 @pytest.fixture(autouse=True)
@@ -296,18 +296,8 @@ def fixture_dir(fixture_base: Path) -> FixtureDirGetter:
 
 
 @pytest.fixture
-def tmp_dir() -> Iterator[str]:
-    dir_ = tempfile.mkdtemp(prefix="poetry_")
-    path = Path(dir_)
-
-    yield path.resolve().as_posix()
-
-    remove_directory(path, force=True)
-
-
-@pytest.fixture
-def tmp_venv(tmp_dir: str) -> Iterator[VirtualEnv]:
-    venv_path = Path(tmp_dir) / "venv"
+def tmp_venv(tmp_path: Path) -> Iterator[VirtualEnv]:
+    venv_path = tmp_path / "venv"
 
     EnvManager.build_venv(venv_path)
 
@@ -348,14 +338,14 @@ def repo(http: type[httpretty.httpretty]) -> TestRepository:
 
 @pytest.fixture
 def project_factory(
-    tmp_dir: str,
+    tmp_path: Path,
     config: Config,
     repo: TestRepository,
     installed: Repository,
     default_python: str,
     load_required_fixtures: None,
 ) -> ProjectFactory:
-    workspace = Path(tmp_dir)
+    workspace = tmp_path
 
     def _factory(
         name: str | None = None,
@@ -380,9 +370,7 @@ def project_factory(
                 project_dir.mkdir(parents=True, exist_ok=True)
 
             if pyproject_content:
-                with project_dir.joinpath("pyproject.toml").open(
-                    "w", encoding="utf-8"
-                ) as f:
+                with (project_dir / "pyproject.toml").open("w", encoding="utf-8") as f:
                     f.write(pyproject_content)
         else:
             assert name is not None
@@ -446,10 +434,10 @@ def set_simple_log_formatter() -> None:
 
 
 @pytest.fixture
-def fixture_copier(fixture_base: Path, tmp_dir: str) -> FixtureCopier:
+def fixture_copier(fixture_base: Path, tmp_path: Path) -> FixtureCopier:
     def _copy(relative_path: str, target: Path | None = None) -> Path:
-        path = fixture_base.joinpath(relative_path)
-        target = target or Path(tmp_dir, relative_path)
+        path = fixture_base / relative_path
+        target = target or (tmp_path / relative_path)
         target.parent.mkdir(parents=True, exist_ok=True)
 
         if target.exists():
--- a/tests/console/commands/env/conftest.py
+++ b/tests/console/commands/env/conftest.py
@@ -2,7 +2,6 @@ from __future__ import annotations
 
 import os
 
-from pathlib import Path
 from typing import TYPE_CHECKING
 
 import pytest
@@ -12,6 +11,7 @@ from poetry.utils.env import EnvManager
 
 if TYPE_CHECKING:
     from collections.abc import Iterator
+    from pathlib import Path
 
     from tests.helpers import PoetryTestApplication
 
@@ -25,8 +25,8 @@ def venv_name(app: PoetryTestApplication) -> str:
 
 
 @pytest.fixture
-def venv_cache(tmp_dir: str) -> Path:
-    return Path(tmp_dir)
+def venv_cache(tmp_path: Path) -> Path:
+    return tmp_path
 
 
 @pytest.fixture(scope="module")
@@ -49,7 +49,7 @@ def venvs_in_cache_dirs(
 ) -> list[str]:
     directories = []
     for version in python_versions:
-        directory = venv_cache.joinpath(f"{venv_name}-py{version}")
+        directory = venv_cache / f"{venv_name}-py{version}"
         directory.mkdir(parents=True, exist_ok=True)
         directories.append(directory.name)
     return directories
--- a/tests/console/commands/test_new.py
+++ b/tests/console/commands/test_new.py
@@ -155,18 +155,20 @@ def test_command_new(
     package_path: str,
     include_from: str | None,
     tester: CommandTester,
-    tmp_dir: str,
+    tmp_path: Path,
 ):
-    path = Path(tmp_dir) / directory
-    options.append(path.as_posix())
+    path = tmp_path / directory
+    options.append(str(path))
     tester.execute(" ".join(options))
     verify_project_directory(path, package_name, package_path, include_from)
 
 
 @pytest.mark.parametrize(("fmt",), [(None,), ("md",), ("rst",), ("adoc",), ("creole",)])
-def test_command_new_with_readme(fmt: str | None, tester: CommandTester, tmp_dir: str):
+def test_command_new_with_readme(
+    fmt: str | None, tester: CommandTester, tmp_path: Path
+):
     package = "package"
-    path = Path(tmp_dir) / package
+    path = tmp_path / package
     options = [path.as_posix()]
 
     if fmt:
@@ -191,7 +193,7 @@ def test_respect_prefer_active_on_new(
     config: Config,
     mocker: MockerFixture,
     tester: CommandTester,
-    tmp_dir: str,
+    tmp_path: Path,
 ):
     from poetry.utils.env import GET_PYTHON_VERSION_ONELINER
 
@@ -208,8 +210,8 @@ def test_respect_prefer_active_on_new(
     config.config["virtualenvs"]["prefer-active-python"] = prefer_active
 
     package = "package"
-    path = Path(tmp_dir) / package
-    options = [path.as_posix()]
+    path = tmp_path / package
+    options = [str(path)]
     tester.execute(" ".join(options))
 
     pyproject_file = path / "pyproject.toml"
--- a/tests/installation/test_executor.py
+++ b/tests/installation/test_executor.py
@@ -91,8 +91,8 @@ class Chef(BaseChef):
 
 
 @pytest.fixture
-def env(tmp_dir: str) -> MockEnv:
-    path = Path(tmp_dir) / ".venv"
+def env(tmp_path: Path) -> MockEnv:
+    path = tmp_path / ".venv"
     path.mkdir(parents=True)
 
     return MockEnv(path=path, is_venv=True)
@@ -169,19 +169,16 @@ def mock_file_downloads(
 
 
 @pytest.fixture
-def copy_wheel(tmp_dir: Path, fixture_dir: FixtureDirGetter) -> Callable[[], Path]:
+def copy_wheel(tmp_path: Path, fixture_dir: FixtureDirGetter) -> Callable[[], Path]:
     def _copy_wheel() -> Path:
         tmp_name = tempfile.mktemp()
-        Path(tmp_dir).joinpath(tmp_name).mkdir()
+        (tmp_path / tmp_name).mkdir()
 
         shutil.copyfile(
-            (
-                fixture_dir("distributions") / "demo-0.1.2-py2.py3-none-any.whl"
-            ).as_posix(),
-            (Path(tmp_dir) / tmp_name / "demo-0.1.2-py2.py3-none-any.whl").as_posix(),
+            fixture_dir("distributions") / "demo-0.1.2-py2.py3-none-any.whl",
+            tmp_path / tmp_name / "demo-0.1.2-py2.py3-none-any.whl",
         )
-
-        return Path(tmp_dir) / tmp_name / "demo-0.1.2-py2.py3-none-any.whl"
+        return tmp_path / tmp_name / "demo-0.1.2-py2.py3-none-any.whl"
 
     return _copy_wheel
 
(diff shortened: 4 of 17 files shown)


martin-kokos:
rebased to master

tests/conftest.py
```diff
@@ -292,15 +291,6 @@ def _fixture_dir(name: str) -> Path:
     return _fixture_dir
 
 
-@pytest.fixture
-def tmp_dir() -> Iterator[str]:
-    dir_ = tempfile.mkdtemp(prefix="poetry_")
```
eamanu, line 297: we want to remove the poetry_ prefix?

tests/console/commands/env/conftest.py
```diff
@@ -25,8 +25,8 @@ def venv_name(app: PoetryTestApplication) -> str:
 
 
 @pytest.fixture
-def venv_cache(tmp_dir: str) -> Path:
-    return Path(tmp_dir)
+def venv_cache(tmp_path: Path) -> Path:
```
eamanu, line 28: if TYP_CHECKING is False, it wont fail?

tests/conftest.py
```diff
@@ -453,10 +443,10 @@ def set_simple_log_formatter() -> None:
 
 
 @pytest.fixture
-def fixture_copier(fixture_base: Path, tmp_dir: str) -> FixtureCopier:
+def fixture_copier(fixture_base: Path, tmp_path: Path) -> FixtureCopier:
     def _copy(relative_path: str, target: Path | None = None) -> Path:
         path = fixture_base.joinpath(relative_path)
```
eamanu, line 448: is not the same that write `fixture_base / relative_path`?

tests/console/commands/test_new.py
```diff
@@ -155,18 +155,20 @@ def test_command_new(
...
-    path = Path(tmp_dir) / directory
+    path = tmp_path / directory
     options.append(path.as_posix())
     tester.execute(" ".join(options))
     verify_project_directory(path, package_name, package_path, include_from)
 
 
 @pytest.mark.parametrize(("fmt",), [(None,), ("md",), ("rst",), ("adoc",), ("creole",)])
-def test_command_new_with_readme(fmt: str | None, tester: CommandTester, tmp_dir: str):
+def test_command_new_with_readme(
```
eamanu, line 167: There's 2 new lines that are not necessary, I'd keep the code in one line as before.

tests/installation/test_executor.py
```diff
@@ -153,26 +153,23 @@ def callback(
...
-            .joinpath(tmp_name)
+            tmp_path.joinpath(tmp_name)
             .joinpath("demo-0.1.2-py2.py3-none-any.whl")
             .as_posix(),
         )
 
-        return (
-            Path(tmp_dir).joinpath(tmp_name).joinpath("demo-0.1.2-py2.py3-none-any.whl")
-        )
+        return tmp_path.joinpath(tmp_name).joinpath("demo-0.1.2-py2.py3-none-any.whl")
```
eamanu, line 172: I think is more readable write:

`tmp_path / tmp_name / "demo-0.1.2-py2.py3-none-any.whl"`

eamanu:
There's an old PR https://github.com/python-poetry/poetry/pull/7129 but From last year and it was not updated.

tests/conftest.py
martin-kokos: Good question. I did not find any place that would rely on that prefix. The pytest's `tmp_path` fixture ensures there are no path collisions by using it's own (per-run) pytest prefix.

eamanu:
As @radoering noted the current CI tests in windows has a similar fail with the other PR https://github.com/python-poetry/poetry/pull/7129.

tests/console/commands/test_new.py
martin-kokos: That's `black` splitting the lines.

tests/console/commands/env/conftest.py
martin-kokos: `Path` is not used outside of type annotations and `TYPE_CHECKING` is false at runtime when tests are run.

martin-kokos:
I have added some code for debugging the failed Windows pipelines because the error output is truncated.
The problem appears to be the `pytest.tmp_path` path prefix making the file paths too long.
```
Cloning into 'C:/Users/runneradmin/AppData/Local/Temp/pytest-of-runneradmin/pytest-0/popen-gw0/test_system_git_called_when_co0/.cache/pypoetry/src/test-fixture-vcs-repository/submodules/sample-namespace-packages'...
fatal: cannot write keep file 'C:/Users/runneradmin/AppData/Local/Temp/pytest-of-runneradmin/pytest-0/popen-gw0/test_system_git_called_when_co0/.cache/pypoetry/src/test-fixture-vcs-repository/.git/modules/submodules/sample-namespace-packages/objects/pack/pack-c638ec21e7fcb21acf4e0b89cefcdb31ab0adc29.keep': Filename too long
```
https://github.com/python-poetry/poetry/actions/runs/4364500429/jobs/7631897284#step:14:2832

eamanu:
Take a look here https://github.com/pytest-dev/pytest/issues/8999

martin-kokos:
Just on a curious note. I am getting a failing test in the pipeline which errors out on code that doesn't exist at this point in time.

The error:
```
==================================== ERRORS ====================================
_______ ERROR at setup of test_execute_prints_warning_for_invalid_wheels _______
[gw1] linux -- Python 3.7.16 /home/runner/work/poetry/poetry/.venv/bin/python
file /home/runner/work/poetry/poetry/tests/installation/test_executor.py, line 341
  def test_execute_prints_warning_for_invalid_wheels(
E       fixture 'tmp_dir' not found
>       available fixtures: artifact_cache, auth_config_source, cache, capfd, capfdbinary, caplog, capsys, capsysbinary, class_mocker, config, config_cache_dir, config_dir, config_source, config_virtualenvs_path, copy_wheel, cov, current_env, current_python, default_python, doctest_namespace, download_mock, dummy_keyring, env, environ, fixture_base, fixture_copier, fixture_dir, git_mock, http, installed, io, io_decorated, io_not_decorated, isolate_environ, load_required_fixtures, mock_file_downloads, mock_user_config_dir, mocked_open_files, mocker, module_mocker, monkeypatch, no_cover, package_mocker, pep517_metadata_mock, pool, project_factory, project_root, pytestconfig, record_property, record_testsuite_property, record_xml_attribute, recwarn, repo, required_fixtures, session_mocker, set_simple_log_formatter, testrun_uid, tmp_path, tmp_path_factory, tmp_venv, tmpdir, tmpdir_factory, wheel, with_chained_fail_keyring, with_chained_null_keyring, with_fail_keyring, with_null_keyring, with_simple_keyring, worker_id
>       use 'pytest --fixtures [testpath]' for help on them.

/home/runner/work/poetry/poetry/tests/installation/test_executor.py:341
```
https://github.com/python-poetry/poetry/actions/runs/4583142389/jobs/8093814943?pr=7412#step:14:2715

The file at the current commit of the pipeline run:
https://github.com/python-poetry/poetry/blob/39662eee491b537aebda22e504577176361db3be/tests/installation/test_executor.py#L341

dimbleby:
> I am getting a failing test in the pipeline which errors out on code that doesn't exist at this point in time.

looks like the pipeline is smart enough to checkout the merge commit that would happen, if this MR were to be merged.  (That sounds like what you'd want, else the pipeline could pass but then fail on the master branch as soon as it was merged).

You're going to want to merge or rebase from master sooner or later anyway, suggesst you do that.

martin-kokos:
> looks like the pipeline is smart enough to checkout the merge commit that would happen, if this MR were to be merged. 

I did not expect that. Thanks.
I just did, a second time, but will do again.

martin-kokos:
So, fixed those test failing due to long path names on Windows.

eamanu:
Take a look the pre-commit.ci @martin-kokos

martin-kokos:
Wow. Didn't know the CI list is a scrolly frame. Thought I just didn't have permission to see all the CIs.
The black was fixed. Not sure how it got past as I do have pre-commit installed.

tests/conftest.py
```diff
@@ -52,6 +52,21 @@
     from tests.types import ProjectFactory
 
 
+# Enable handling long path names (260) in windows
+# https://docs.microsoft.com/en-us/windows/win32/fileio/naming-a-file#maximum-path-length-limitation
+if sys.platform.startswith("win32"):
```
radoering, line 57: We should probably put this into a fixture with `autouse=True`

tests/conftest.py
```diff
@@ -52,6 +52,21 @@
...
+    import subprocess
+
+    git_command = find_git_command()
+    env = os.environ.copy()
+    env["GIT_TERMINAL_PROMPT"] = "0"
+    args = ["config", "--system", "core.longpaths", "true"]
```
radoering, line 63: IIUC, we are changing a git setting globally? Can we do it locally (`--local`)?

tests/installation/test_executor.py
```diff
@@ -165,19 +165,20 @@ def callback(
 
 
 @pytest.fixture
-def copy_wheel(tmp_dir: Path, fixture_dir: FixtureDirGetter) -> Callable[[], Path]:
+def copy_wheel(tmp_path: Path, fixture_dir: FixtureDirGetter) -> Callable[[], Path]:
     def _copy_wheel() -> Path:
         tmp_name = tempfile.mktemp()
-        Path(tmp_dir).joinpath(tmp_name).mkdir()
+        tmp_path.joinpath(tmp_name).mkdir()
```
radoering, line 171:
```suggestion
        (tmp_path /tmp_name).mkdir()
```

tests/installation/test_executor.py
```diff
@@ -165,19 +165,20 @@ def callback(
...
 
         shutil.copyfile(
             (
                 fixture_dir("distributions") / "demo-0.1.2-py2.py3-none-any.whl"
             ).as_posix(),
-            (Path(tmp_dir) / tmp_name / "demo-0.1.2-py2.py3-none-any.whl").as_posix(),
+            tmp_path.joinpath(tmp_name)
+            .joinpath("demo-0.1.2-py2.py3-none-any.whl")
+            .as_posix(),
```
radoering, line 179:
```suggestion
            tmp_path / tmp_name / "demo-0.1.2-py2.py3-none-any.whl",
```

I suppose we can remove the `as_posix()` in the line above, too.

radoering (changes requested):

tests/conftest.py
martin-kokos: local doesn't work. I don't know how the test pipeline si setup. It took me a while to figure out it needs to be set more broadly. Alternatively, the option could be set somewhere where the test environment of CI. I'd need some help with that.

tests/conftest.py
martin-kokos: :heavy_check_mark:

tests/conftest.py
dimbleby: it's not ok for running test scripts to have the side-effect of changing the user's global git configuration, though!

- can a way be found to make the paths shorter in the first place?
- if this has to be system configuration, it should probably be run from the CI workflow yaml definition - so that it only affects CI

martin-kokos: That's a very good point. I don't know why I was thinking only in the context of the CI.

Getting the path length just bellow the limit is just hiding it for the next guy on linux who won't know why its failing on windows. It's failing on a git subprocess, so the error is not super easy to spot, especially since the subprocess stderr is truncated.

I have come across an interesting fact; that Python for windows installer has an option to [Enable long paths on windows via a registry key when the installer is run](https://stackoverflow.com/questions/51019926/python-installer-for-windows-disable-path-length-limit-option-not-available). This leads me to believe that the complication is somewhat common and avoidable with that without drawbacks as the registry key help text doesn't mention any.

I will look into moving it to CI.

martin-kokos: I've put the long pathname config into CI.

I think that the reason --local doesn't work is because during tests, different repos are handled and --local sets it only for the current poetry repo which is not where the issue manifests itself.

tests/console/commands/env/conftest.py
```diff
@@ -36,7 +36,7 @@ def python_versions() -> list[str]:
 
 @pytest.fixture
 def venvs_in_cache_config(app: PoetryTestApplication, venv_cache: Path) -> None:
-    app.poetry.config.merge({"virtualenvs": {"path": str(venv_cache)}})
+    app.poetry.config.merge({"virtualenvs": {"path": venv_cache}})
```
radoering, line 39: Why? It probably doesn't matter, but in config paths are strings, aren't they?

tests/installation/test_executor.py
```diff
@@ -1193,7 +1190,7 @@ def test_executor_fallback_on_poetry_create_error_without_wheel_installer(
 
     config.merge(
         {
-            "cache-dir": tmp_dir,
+            "cache-dir": tmp_path,
```
radoering, line 1193:
```suggestion
            "cache-dir": str(tmp_path),
```

for consistency

tests/utils/test_env.py
```diff
@@ -267,9 +267,9 @@ def test_activate_fails_when_python_cannot_be_found(
     if "VIRTUAL_ENV" in os.environ:
         del os.environ["VIRTUAL_ENV"]
 
-    os.mkdir(os.path.join(tmp_dir, f"{venv_name}-py3.7"))
+    os.mkdir(tmp_path / f"{venv_name}-py3.7")
 
-    config.merge({"virtualenvs": {"path": str(tmp_dir)}})
+    config.merge({"virtualenvs": {"path": tmp_path}})
```
radoering, line 272:
```suggestion
    config.merge({"virtualenvs": {"path": str(tmp_path)}})
```

tests/installation/test_executor.py
```diff
@@ -201,15 +198,15 @@ def test_execute_executes_a_batch_of_operations(
...
+    tmp_path: Path,
     mock_file_downloads: None,
     env: MockEnv,
     copy_wheel: Callable[[], Path],
     fixture_dir: FixtureDirGetter,
 ):
     wheel_install = mocker.patch.object(WheelInstaller, "install")
 
-    config.merge({"cache-dir": tmp_dir})
+    config.merge({"cache-dir": tmp_path})
```
radoering, line 209:
```suggestion
    config.merge({"cache-dir": str(tmp_path)})
```

radoering (changes requested):
Just some minor nitpicks for consistency. I think after that we are ready to merge. 😄

tests/installation/test_executor.py
martin-kokos: :heavy_check_mark:

tests/installation/test_executor.py
martin-kokos: :heavy_check_mark:

dimbleby:
#7784 kicks in and tells us something about the flakey CI!

https://github.com/python-poetry/poetry/actions/runs/4669647501/jobs/8268387674?pr=7412
```
error: Invalid distribution name or version syntax: poetry-config-5k08imt--0.1.0
```
(not the error I was expecting, but should be a good clue anyway)

radoering:
```
test_info_setup_missing_mandatory_should_trigger_pep517[name]
```

It's the test case with the missing name in `setup.py`.

```
error: Invalid distribution name or version syntax: poetry-config-5k08imt--0.1.0
```

 Is it possible that the name of some temporary directory is used as package name and that if we are unlucky this name ends with a dash so that we get an invalid name? (Can't look for myself right now.)

martin-kokos:
> #7784 kicks in and tells us something about the flakey CI!

I have rebased to pull it in.

dimbleby:
> I have rebased to pull it in.

As previously discussed - https://github.com/python-poetry/poetry/pull/7412#issuecomment-1493409146 - the pipeline was doing that already

radoering:
It seems that the failure is really caused by this PR:

After setting a breakpoint at https://github.com/python-poetry/poetry/blob/5a9da19fa85631f46a5f1edc9e61c2b65255dab2/tests/inspection/test_info.py#L236

I noticed that `source_dir` is empty without this PR, but contains two directories `.cache` and something like `poetry_config_jz_t48_l` with this PR.

If there is no name in `setup.py` and there is exactly one directory (directories starting with a dot don't count as it seems) next to `setup.py`, setuptools assumes that's the name. If there are two folders setuptools raises an exception. If there is no folder, setuptools sets the name to `UNKNOWN`.

Thus, the question is what change of this PR creates these two directories?

radoering:
OK, it's clear now: `config_cache_dir` and `config_dir` in `conftest.py` used `tmp_dir` before and `source_dir` in `test_info.py` used `tmp_path`, so we had two different directories and now it's all the same.

I suppose we should change https://github.com/python-poetry/poetry/blob/5a9da19fa85631f46a5f1edc9e61c2b65255dab2/tests/inspection/test_info.py#L39

```diff
-    return Path(tmp_path.as_posix())
+    return tmp_path / "source"
```

tests/conftest.py
```diff
@@ -219,8 +218,8 @@ def config(
 
 
 @pytest.fixture()
-def config_dir(tmp_dir: str) -> Path:
-    return Path(tempfile.mkdtemp(prefix="poetry_config_", dir=tmp_dir))
+def config_dir(tmp_path: Path) -> Path:
+    return Path(tempfile.mkdtemp(prefix="poetry_config_", dir=tmp_path))
```
radoering, line 222:
```suggestion
    path = tmp_path / "config"
    path.mkdir()
    return path
```

I don't think we need a random name here since we create it in the `tmp_path` for each test.

martin-kokos:
> OK, it's clear now: `config_cache_dir` and `config_dir` in `conftest.py` used `tmp_dir` before and `source_dir` in `test_info.py` used `tmp_path`, so we had two different directories and now it's all the same.
> 
> I suppose we should change
> 
> https://github.com/python-poetry/poetry/blob/5a9da19fa85631f46a5f1edc9e61c2b65255dab2/tests/inspection/test_info.py#L38-L39
> 
> ```diff
> -    return Path(tmp_path.as_posix())
> +    path = tmp_path / "source"
> +    path.mkdir()
> +    return path
> ```

Thanks a lot for the investigation. I don't think I would've been able to figure it out on my own, especially on Win. Is there a way to trigger CI re-run on GH, to make sure its not passing by chance since it was manifesting itself randomly?

tests/conftest.py
martin-kokos: :heavy_check_mark: 
Should I review the usage of tempfile in tests and use tmp_path where possible as well?

tests/console/commands/env/conftest.py
radoering, line 39:
```suggestion
    app.poetry.config.merge({"virtualenvs": {"path": str(venv_cache)}})
```

I'd say let's keep `str` for consistency. By the way that will trigger CI. 😉 (One extra run is enough for me since I'm quite confident about the fix.)

tests/conftest.py
radoering: It's ok for me to put this of until later. You can see if it's worth in a follow up PR if you like.

tests/console/commands/env/conftest.py
martin-kokos: :heavy_check_mark:

radoering (approved):

PR #7437, opened 2023-01-30
test: remove cachy dependency and cache compatiblity tests

martin-kokos:
# Pull Request Check List

Resolves: kind of related to #5720

Followup of https://github.com/python-poetry/poetry/pull/5868 by chadac which I am happy for. This PR removes `cachy` dependency and compatibility tests that were put in place for the transition from old `cachy.CacheManager` to the new `utils.cache.FileCache` and . Cachy is an abandoned package.

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -62,22 +62,6 @@ filecache = ["lockfile (>=0.9)"]
 redis = ["redis (>=2.10.5)"]
 
 [[package]]
-name = "cachy"
-version = "0.3.0"
-description = "Cachy provides a simple yet effective caching library."
-optional = false
-python-versions = ">=2.7, !=3.0.*, !=3.1.*, !=3.2.*, !=3.3.*"
-files = [
-    {file = "cachy-0.3.0-py2.py3-none-any.whl", hash = "sha256:338ca09c8860e76b275aff52374330efedc4d5a5e45dc1c5b539c1ead0786fe7"},
-    {file = "cachy-0.3.0.tar.gz", hash = "sha256:186581f4ceb42a0bbe040c407da73c14092379b1e4c0e327fdb72ae4a9b269b1"},
-]
-
-[package.extras]
-memcached = ["python-memcached (>=1.59,<2.0)"]
-msgpack = ["msgpack-python (>=0.5,<0.6)"]
-redis = ["redis (>=3.3.6,<4.0.0)"]
-
-[[package]]
 name = "certifi"
 version = "2023.5.7"
 description = "Python package for providing Mozilla's CA Bundle."
@@ -1688,4 +1672,4 @@ testing = ["big-O", "flake8 (<5)", "jaraco.functools", "jaraco.itertools", "more
 [metadata]
 lock-version = "2.0"
 python-versions = "^3.8"
-content-hash = "2a028eba8da6043cb897458153fc28cadc9a7b9184fe8edaee48e3e5a8468f06"
+content-hash = "fab5b1b0800c476f29b2f064c2c664cde4bcf7b8fbe98c58feb529d0935f98fb"
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -182,7 +182,6 @@ warn_unused_ignores = false
 [[tool.mypy.overrides]]
 module = [
     'cachecontrol.*',
-    'cachy.*',
     'deepdiff.*',
     'httpretty.*',
     'keyring.*',
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -66,8 +66,6 @@ urllib3 = "^1.26.0"
 pre-commit = "^2.6"
 
 [tool.poetry.group.test.dependencies]
-# Cachy frozen to test backwards compatibility for `poetry.utils.cache`.
-cachy = "0.3.0"
 deepdiff = "^6.3"
 httpretty = "^1.0"
 pytest = "^7.1"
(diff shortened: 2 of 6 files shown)


neersighted:
I'm not sure this is something we want; at least not for now. `cachy` is only used at test-time, and I'd rather have the backwards compat tests for a few more releases to ensure we don't break something. We have cache versioning -- the reason for this code is to make sure we don't choke on the on-disk representation which could be spectacular given that we don't have defensive logic protecting everywhere we use the cache.

To remove cachy, I think I'd prefer to see tests that ensure we gracefully handle a 'corrupt' cache.

martin-kokos:
> I'm not sure this is something we want; at least not for now. `cachy` is only used at test-time, and I'd rather have the backwards compat tests for a few more releases to ensure we don't break something. We have cache versioning above the on-disk format -- the reason for this code is to make sure we don't choke on the on-disk representation which could be spectacular given that we don't have defensive logic protecting everywhere we use the cache.
> 
> To remove cachy, I think I'd prefer to see tests that ensure we gracefully handle a 'corrupt' cache.

Alright. I didn't realize how long term the cache actually is. So my understanding now is that a cache generated by older poetry version can be used by current poetry version. I will attempt to look into that graceful handling.

In the meanwhile, would it be acceptable to modify the tests such that cachy test would be skipable? At this time cachy is imported in conftest, so I don't think there is any good way to achieve that.

Secrus:
> Alright. I didn't realize how long term the cache actually is. So my understanding now is that a cache generated by older poetry version can be used by current poetry version. I will attempt to look into that graceful handling.
> 
> In the meanwhile, would it be acceptable to modify the tests such that cachy test would be skipable? At this time cachy is imported in conftest, so I don't think there is any good way to achieve that.

Would you care to explain why is this such a great matter to you?

neersighted:
It's not that the cache is long-lived, is that historically a invalid cache file jams Poetry until the cache is cleared. As such we need to make sure to feed our replacement for Cachy garbage input and make sure that it gracefully handles corrupt files.

martin-kokos:
> > Alright. I didn't realize how long term the cache actually is. So my understanding now is that a cache generated by older poetry version can be used by current poetry version. I will attempt to look into that graceful handling.
> > In the meanwhile, would it be acceptable to modify the tests such that cachy test would be skipable? At this time cachy is imported in conftest, so I don't think there is any good way to achieve that.
> 
> Would you care to explain why is this such a great matter to you?

I'd like to include poetry in the Gentoo Linux package repo https://github.com/gentoo/gentoo/pull/28843
And because it is desirable to have packages pass their tests in the environment they are installed in, test dependencies need to be satisfied. That would include cachy, but it being an abandoned package and not very popular, I'd like to avoid pulling it into the package repo. It would be acceptable to skip the test involving cachy, but as it is now, I don't think it is possible in a nice way it being imported in conftest. A last resort would be file patching at install time, but that's best to be avoided, and I'd much rather invest the effort in poetry towards as #5720 says *rock solid dependencies*.

neersighted:
Unfortunately, I don't think making our test suite less robust to mitigate distro packaging choices is tenable here in the upstream. If you need to bypass that test, patching it out is your best bet. Or if you want to do something mergable, I'd suggest a robust set of tests with a corpus of good and bad inputs (and maybe even some fuzzing) to ensure that we no longer hang on an invalid cache would be good.

The cachy tests don't prevent the invalid cache issue today, but they do help us avoid making it worse. If we are resilient to an outdated/corrupt on-disk format, that will give us the confidence to drop this set of tests from the test suite.

martin-kokos:
> 

I wouldn't mean to making the tests lest robust in favor of distro packaging. That's why I said, Id rather put the effort into poetry, however unless I am missing something, the new `utils.cache.FileCache` is already being used now https://github.com/python-poetry/poetry/pull/5868 and the tests that this PR removes merely a cleanup of tests for the transition, proving the interfaces of the two cache implementations is the same.

I can understand the need for a more robust cache, but that concern would belong to #5868 (never mind the fact cachy wasn't in the first place, as you've mentioned) or filed as a separate issue.

I've created https://github.com/python-poetry/poetry/pull/7453 in an attempt to make Cache more resilient to cache file corruption.

martin-kokos:
Rebased and resolved conflicts.

Is there any update on the opinion of not wanting to remove tests dependent on cachy just yet?
If not, would it be acceptable to make cachy on *optional* dependency achieved with something like [pytest.importskip](https://docs.pytest.org/en/latest/reference/reference.html#pytest-importorskip)

poetry.lock
```diff
@@ -1,4 +1,4 @@
-# This file is automatically @generated by Poetry 1.4.0 and should not be changed by hand.
```
Secrus, line 1: Please revert any changes not connected with the removal of `cachy`.

martin-kokos: Since removing it from pyproject.toml, the lock file needs to be regenerated, no?

Secrus: yes, but as you can see, you not only removed cachy, but also updated other dependencies, which shouldn't be the case (not as a part of this PR at least)

martin-kokos: Ahh. Did `poetry lock --no-update` and fixed. Thanks.

martin-kokos:
Rebased and resolved conflicts.

Secrus:
@martin-kokos could you rebase and resolve conflicts? I guess it's time to merge this

martin-kokos:
:raised_hands: Rebased. Thanks for the ping.
If you could, please, check a newly added commit resolving some typing checks.

Secrus (approved):

Issue #6937, opened 2022-11-02
`poetry install` checks `README.md`, but `poetry check` does not.


YDX-2147483647:
- **Poetry version**: 1.2.2
- **Python version**:  3.10.8
- **OS version and name**: Windows 10 (21H1) and Raspberry Pi OS (based on Debian 11 Bullseye)
- **pyproject.toml**: default – `poetry new <package_name>` can fire this issue.

- [x] I am on the [latest](https://github.com/python-poetry/poetry/releases/latest) stable Poetry version, installed using a recommended method.
- [x] I have searched the [issues](https://github.com/python-poetry/poetry/issues) of this repo and believe that this is not a duplicate.
- [x] I have consulted the [FAQ](https://python-poetry.org/docs/faq/) and [blog](https://python-poetry.org/blog/) for any relevant entries or release notes. (*Note*: I’ve only checked FAQ and blog 1.2.x)
- [x] If an exception occurs when executing a command, I executed it again in debug mode (`-vvv` option) and have included the output below.

## Issue

<u>`poetry install` checks `README.md`, but `poetry check` does not.</u>

>  To be rigorous, it checks [`tool.poetry.readme`](https://python-poetry.org/docs/pyproject/#readme) in `pyproject.toml`.

## Is it an issue?

- **All commands work**: `poetry install` raises a `FileNotFoundError`, but it works.

- [Doc](https://python-poetry.org/docs/cli/#install): The `install` command reads `pyproject.toml`, resolves the dependencies, and installs them.

  It should not care about `README`.

- [Doc](https://python-poetry.org/docs/cli/#check): The `check` command validates the structure of  `pyproject.toml`.

  Although `README` is absent, the TOML structure is valid. So this might be expected.

## To Reproduce

TL; DR:

1. Create a project.
2. `rm README.md`.
3. `poetry install`: No such file or directory: `README.md`.

### 1. Create a project

```shell
$ poetry new why-read-me
$ cd why-read-me

$ cat pyproject.toml
# There's a default `README.md`.
[tool.poetry]
name = "why-read-me"
……
readme = "README.md"
……
```

### 2. `poetry install` checks `README.md`

```shell
$ mv .\README.md read-me-whatever.md
# Let's steal the README.

$ poetry install
# The `install` command checks the README, and raises an error.
# However, it successfully installs packages. (If I add some packages in pyproject.toml, it will be installed here.)
Using virtualenv: ……
Installing dependencies from lock file

[...]

Issue #6756, opened 2022-10-09
Replace `poetry lock --check` with `poetry check --lock`


johnthagen:
- [x] I have searched the [issues](https://github.com/python-poetry/poetry/issues) of this repo and believe that this is not a duplicate.
- [x] I have searched the [FAQ](https://python-poetry.org/docs/faq/) and general [documentation](https://python-poetry.org/docs/) and believe that my question is not already covered.

## Feature Request

There is an important need for development teams to validate their Poetry configuration files and lockfile freshness in CI. This should ideally be done in a single, easy to understand command.

The following PR added `poetry lock --check`:

- #1954

The problem is that this is a somewhat strange UI, especially given that the separate `poetry check` command already exists (and existed). Sebastian noted this and asked for a revert (which must never have happened?):

- https://github.com/python-poetry/poetry/pull/1954#issuecomment-800096650

I propose, similar to Sebastian, that `poetry lock --check` be deprecated and its functionality moved to `poetry check --lock`. This would be easier to teach and explain to new Poetry users and would keep all "file validation logic" under `poetry check`.

`poetry check --lock` would therefor be the superset of everything that is checked in `poetry check` and `poetry lock --check` so that a user can run a single command in CI and get all validation performed.

@neersighted seemed to agree in:

- https://github.com/python-poetry/poetry/pull/6704#issuecomment-1272216665

PR #2881, opened 2020-09-02
Check project against valid and deprecated trove classifiers.

kdeldycke:
Inspects trove classifiers on `check` CLI command calls, and look for unrecognized and deprecated categories.

Adds dependency https://github.com/pypa/trove-classifiers, a package published and maintained by PyPA that is cataloguing all classifiers. This is the canonical source of all trove definitions.

# Pull Request Check List

Resolves: #2579

- [x] Added **tests** for changed code.
- [x] Updated **documentation** for changed code.

PR #7444, opened 2023-01-31
`poetry check` command validates readme (files must exist)

YDX-2147483647:
Let `poetry check` command checks existence of referenced readme files.

Reason:

- Prevent typo.
- Help to debug `poetry install`. (See #6937)

## Pull Request Check List

Resolves: #6937

### [Removed] `warnings: list[str] = []` in `CheckCommand.validate_readme`

To be consistent with `validate_classifiers` from #2881 and pass `mypy`.

### On case-sensitivity

I try to raise warnings for inconsistent cases, but I cannot find any good way to detect the _real_ cases…

- `str(Path.resolve) == str(Path.absolute)` is tricky and doesn’t work for symbolic links.

  Details: [`pathlib` doc](https://docs.python.org/3/library/pathlib.html#pathlib.Path.absolute) and [`os.path` implementation](https://github.com/python/cpython/blob/43af2dbb54785e34f67813eee08156647f340b64/Lib/ntpath.py).

- [`Path.iterdir`](https://docs.python.org/3/library/pathlib.html#pathlib.Path.iterdir) or [`os.listdir`](https://docs.python.org/3/library/os.html#os.listdir) can be too heavy.

---

## Update on 2023-06-10

I guess there will be a big change in `poetry check`  in near future. (#6756) This PR might be obsolete after #8015 is merged.

2023-06-14: #8015 was merged and conflicts were resolved.


src/poetry/console/commands/check.py
```diff
@@ -57,6 +63,22 @@ def validate_classifiers(
...
+        """Check existence of referenced readme files"""
+
+        errors = []
+        warnings: list[str] = []
+
+        readmes = [readme] if isinstance(readme, str) else readme
+
+        for name in readmes:
+            if not (poetry_file.parent / name).exists():
+                errors.append(f"Declared README file must exist: found {name}")
```
radoering, line 78:
```suggestion
                errors.append(f"Declared README file does not exist: {name}")
```

radoering (changes requested):
LGTM. Just a minor nitpick: I'd invert the error message.

YDX-2147483647:
Hi! I agree with you, and I've committed the change.

As I mentioned above, this PR might conflict with #8015, as they both change `poetry check`. Feel free to review this PR later if #8015 is more important/urgent.

YDX-2147483647:
Sorry tests failed because I forgot to update the error message there. Now fixed.

radoering:
#8015 is merged now. Can you please resolve the conflicts?

YDX-2147483647:
Sure!

I've resolved the conflicts, and `poetry run pytest -k check` (with empty `$TERM`) passed. Now waiting for the CI.

src/poetry/console/commands/check.py
```diff
@@ -73,6 +79,22 @@ def validate_classifiers(
 
         return errors, warnings
 
+    def validate_readme(
+        self, readme: str | list[str], poetry_file: Path
+    ) -> tuple[list[str], list[str]]:
+        """Check existence of referenced readme files"""
+
+        errors = []
+        warnings: list[str] = []
```
eamanu, line 88: `warnings` is not used. Why it is created?

radoering: Probably, so that it has the same return values as `validate_classifiers()`. On the other side, we could also omit warnings as long as we don't need it. No strong feelings about it. 🤷

YDX-2147483647: Yes, as I mentioned above, there're 2 reasons:

- to be consistent with `validate_classifiers()`.
- for future extensibility. Say, Python 3.14 introduces a cheap way to detect the real cases of files, then we can raise warnings for inconsistent cases.

I have no strong idea either, should I remove it?

radoering: Maybe, just remove it for now and put a `_` in front of the validate-methods (`_validate_classifiers`, `_validate_readme`) to make clear they are internal and the interface might be changed in future.

If we introduce the third, fourth or fifth validate method, we should think about a common interface (not only return values, but also parameters) and calling the methods in a loop anyway.

YDX-2147483647: Make sense. I did try to make it into a loop (in January), but I found it hard to unify parameters.

radoering (approved):

PR #7453, opened 2023-02-01
Cache: Add logic and tests for FileCache handling corrupt files

martin-kokos:
# Pull Request Check List

Resolves: ~#issue-number-here~

Poetry has started using its own cache implementation https://github.com/python-poetry/poetry/pull/5868
This PR adds logic and test for gracefully handling corruption of test files.


martin-kokos:
rebased to baster

tests/utils/test_cache.py
```diff
@@ -192,3 +195,70 @@ def test_cachy_compatibility(
 
     assert cachy_file_cache.get("key3") == test_str
     assert cachy_file_cache.get("key4") == test_obj
+
+
+def test_missing_cache_file(repository_cache_dir: Path, mocker: MockerFixture) -> None:
```
chadac, line 200: You should be able to use the `poetry_file_cache` fixture here rather than initializing a new cache manually.

tests/utils/test_cache.py
```diff
@@ -192,3 +195,70 @@ def test_cachy_compatibility(
...
+        repository_cache_dir
+        / "cache"
+        / "81/74/09/96/87/a2/66/21/8174099687a26621f4e2cdd7cc03b3dacedb3fb962255b1aafd033cabe831530"  # noqa: E501
+    )
+    assert key1_path.exists()
+
+    # original content: 9999999999"value"
+
+    write_modes = {"str": "w", "bytes": "wb"}
+    with open(key1_path, write_modes[corrupt_payload.__class__.__name__]) as f:
```
chadac, line 258: This should work if it's changed to `write_modes = {str: "w", bytes: "wb"}` and `type(corrupt_payload)` right? That seems a bit more readable.

chadac (changes requested):

src/poetry/utils/cache.py
```diff
@@ -172,7 +184,12 @@ def _get_payload(self, key: str) -> T | None:
             return None
 
         with open(path, "rb") as f:
-            payload = self._deserialize(f.read())
+            file_content = f.read()
+
+        try:
+            payload = self._deserialize(file_content)
```
chadac, line 190: Thinking about this again.. I wonder if we can skip the custom error altogether and simply run the forget+return `None` routine here. `_get_payload` already manages forgetting in case the payload is expired. I'm not too opinionated about this, though.

chadac (dismissed):
Leaving comments with stylistic suggestions, but this looks good to me regardless.

src/poetry/utils/cache.py
martin-kokos: I suppose it is a little extra. Removed.

chadac (dismissed):

martin-kokos:
Rebased. Resolved conflicts. ... bump? can this be merged?

radoering (approved):

Issue #6925, opened 2022-10-30
Poetry is incompatible with unstable-tagged Python builds (Invalid PEP 440 version: '3.Y.Z+')


dimaqq:
I'v just installed Python3.11 from deadsnakes/nightly.

Poetry can't use it, because the Python version has a `+` in it.

`Invalid PEP 440 version: '3.11.0+'` is all I get 😭 

P.S. if the wise men declare that deadsnakes is at fault here, I'll happily bug dead snakes.

neersighted:
Hmm, it looks like this could be a parse error with the Python version numbers provided by Deadsnakes? Looking closer at #6334, it looks like this might be a dupe. That issue was closed as a environment issue caused by bad versions in the target environment, but looking closer, that only represents the comments. The issue body, like this one, seems to be about a Python version.

Deadsnakes versions tend to be like `3.11.0+jammy1`, which should be valid; some investigation of the actual version numbers from `sys` would be quite helpful. We might have to add some logic to strip extra tokens from Python versions in any case -- they really don't make sense for interpreter version comparisons as we don't support anything other than simple three-part-tuple versions anyway.

dimaqq:
```
>>> import sys
>>> sys.version
'3.11.0+ (main, Oct 30 2022, 08:56:33) [GCC 9.4.0]'
>>> sys.version_info
sys.version_info(major=3, minor=11, micro=0, releaselevel='final', serial=0)
```

neersighted:
PTAL @radoering @dimbleby

radoering:
In the stack trace you can see that `python_full_version` is the culprit. According to [PEP 508](https://peps.python.org/pep-0508/#environment-markers) `python_full_version` is equal to `platform.python_version()`, which is `3.11+` in this case. Although I could not find a direct reference, the sample values for `python_full_version` in PEP 508 indicate that it should be PEP 440 compliant.

<details>
  <summary>Stack trace</summary>
  
  ```
  Finding the necessary packages for the current system

  Stack trace:

  16  ~/.local/pipx/venvs/poetry/lib/python3.8/site-packages/cleo/application.py:329 in run
        exit_code = self._run(io)

  15  ~/.local/pipx/venvs/poetry/lib/python3.8/site-packages/poetry/console/application.py:185 in _run
        exit_code: int = super()._run(io)

  14  ~/.local/pipx/venvs/poetry/lib/python3.8/site-packages/cleo/application.py:423 in _run
[...]

PR #7462, opened 2023-02-02
Fix python version parsing for deadsnakes python versions

GeorgeSaussy:
This PR will allows poetry to a single trailing "+" character at the end of a PEP 440 compatible Python version string. This will allow Poetry to handle the nightly builds from CPython.

Note: The version string must be either exactly compatible with PEP 440 or it must be compatible except for a trailing "+" character while also not having a local version segment.

Resolves: #233

### DIFF
--- a/src/poetry/utils/env.py
+++ b/src/poetry/utils/env.py
@@ -134,9 +134,13 @@ else:
     iver = "0"
     implementation_name = platform.python_implementation().lower()
 
-    
-_pep_440_and_deadsnakes_regex = "^\\d+\\.([a-zA-Z\\d]+(\\.[a-zA-Z\\d]+)*)(\\+([a-zA-Z\\d]+(\\.[a-zA-Z\\d]+)*)?)?$"
-_clean_python_version = re.compile(_pep_440_and_deadsnakes_regex).match(platform.python_version()).group()
+_primary_version_regex = "\\d+\\.([a-zA-Z\\d]+(\\.[a-zA-Z\\d]+)*)"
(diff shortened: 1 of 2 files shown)


src/poetry/utils/env.py
```diff
@@ -1678,6 +1688,17 @@ def get_marker_env(self) -> dict[str, Any]:
...
+        local_segment_regex = "(\\+([a-zA-Z\\d]+(\\.[a-zA-Z\\d]+)*)?)?$"
+        pep_440_and_deadsnakes_regex = f"^{primary_version_regex}{local_segment_regex}$"
+        match = re.compile(pep_440_and_deadsnakes_regex).match(
+            platform.python_version()
+        )
+        if match is None:
+            raise ValueError("Python version must be PEP 440 compatible.")
+        clean_python_version = match.group()
+        clean_python_version = clean_python_version.rstrip("+")
+
```
radoering, line 1701: That feels a little over the top to me. I'd keep it simple and just strip the trailing plus. Afterwards, validation can be done via `poetry.core.constraints.version.Version.parse()` instead of the regexes.

I'm a bit torn if it's even necessary to validate the version here, since we parse it later anyway and already fail with a quite good mesage as can be seen in https://github.com/python-poetry/poetry/issues/6925#issue-1429055137. Of course, the message could be improved to say that the invalid version is the python version.

lazka:
This makes it sound like it is specific to deadsnakes, but it's the normal version scheme used for Python built from git, and not deadsnakes specific.

Issue #3013, opened 2020-10-01
poetry install fails because a space in path to env


cruzelante098:
- [ ] I am on the [latest](https://github.com/python-poetry/poetry/releases/latest) Poetry version.

  > I'm on 1.1.0rc1

- [x] I have searched the [issues](https://github.com/python-poetry/poetry/issues) of this repo and believe that this is not a duplicate. 

  > There are issues that deal with this error but are one or two years old, are closed or marked as fixed and the error output message is different.

- [X] If an exception occurs when executing a command, I executed it again in debug mode (`-vvv` option).

- **OS version and name**:  Windows 10 v2004
- **Poetry version**:  1.1.0rc1
- **Link of a [Gist](https://gist.github.com/) with the contents of your pyproject.toml file**: https://gist.github.com/cruzelante098/4a46e9387a50c7ea339adaac5421001f

## Issue
`poetry install` throws the next error:

```
  EnvCommandError

  Command "C:\Users\John Smith\AppData\Local\pypoetry\Cache\virtualenvs\dge-bot-KUpJK1g6-py3.7\Scripts\pip.exe" install --no-deps C:\Users\John Smith\AppData\Local\pypoetry\Cache\artifacts\61\a4\b1\1096e79aeb46c7d99c2c17427ce252e25c6bf439f2b2747770cbcda517\zipp-3.2.0-py3-none-any.whl errored with the following return code 1, and output:       
  ERROR: Invalid requirement: 'C:\\Users\\John'
  Hint: It looks like a path. It does exist.
  WARNING: You are using pip version 20.2.2; however, version 20.2.3 is available.
  You should consider upgrading via the 'C:\Users\John Smith\AppData\Local\pypoetry\Cache\virtualenvs\dge-bot-KUpJK1g6-py3.7\Scripts\python.exe -m pip install --upgrade pip' command.


  at c:\users\john smith\appdata\local\programs\python\python37\lib\site-packages\poetry\utils\env.py:944 in _run
       940│                 output = subprocess.check_output(
       941│                     cmd, stderr=subprocess.STDOUT, **kwargs
       942│                 )
       943│         except CalledProcessError as e:
    →  944│             raise EnvCommandError(e, input=input_)
       945│
       946│         return decode(output)
       947│
       948│     def execute(self, bin, *args, **kwargs):

```

The error seems to be caused because there's a space in my username, that's included in the path to the virtualenv.

[...]

Issue #6848, opened 2022-10-20
Poetry init not working when path contains certain characters


PabloRuizCuevas:
- **1.2.2**: 
- **3.8-2.10**: 
- **Windows 10**: 
- **None**: 

- [X] I am on the [latest](https://github.com/python-poetry/poetry/releases/latest) stable Poetry version, installed using a recommended method.
- [X] I have searched the [issues](https://github.com/python-poetry/poetry/issues) of this repo and believe that this is not a duplicate.
- [X] I have consulted the [FAQ](https://python-poetry.org/docs/faq/) and [blog](https://python-poetry.org/blog/) for any relevant entries or release notes.
- [X] If an exception occurs when executing a command, I executed it again in debug mode (`-vvv` option) and have included the output below.

## Issue
`poetry init --vvv` or in general any poetry command don't work and raise this error:
`Unexpected UTF-8 BOM (decode using utf-8-sig): line 1 column 1 (char 0)`

The only clue I have with this issue is that the username and therefore the path contains a special character "è" 
such as :`"C:\Users\MyUsèrWithRareCharactèr\Project`

And if the path is written somewhere and parsed maybe gives an error like in this SO [issue:](https://stackoverflow.com/questions/13156395/python-load-json-file-with-utf-8-bom-header)

On the other hand the --vvv didn't help to solve it as it logs the same with or without --vvv so it may be a very early problem during execution.

Any idea, besides creating a new user?

Azleur:
I ran into a similar issue with a username containing an ampersand (&). Windows 10, Python 3.10, Poetry 1.2.1.

In my case, the Python executable is in a "safe" directory, but Poetry failed when trying to access cached data due to the ampersand.  I could partially solve the issue by setting the cache-dir to a path without special characters, but Poetry still fails if it needs to invoke a compiler (same underlying reason: ampersand in a path that is not quoted).

The feeling I get as an external observer is that Poetry doesn't quote path strings when calling system commands, and that the issue would be solved if it did.

Azleur:
[poetry-test.zip](https://github.com/python-poetry/poetry/files/10582612/poetry-test.zip)

I managed to reproduce the bug in a minimal setting. All you need is a cache-dir with an ampersand.

Azleur:
I believe the problem can be fixed by modifying `poetry.utils._compat.list_to_shell_command()`. I tried the following version locally and it seemed to work:

```python
def list_to_shell_command(cmd: list[str]) -> str:
[...]

PR #7471, opened 2023-02-04
don't set shell=True on subprocess calls

dimbleby:
When poetry executes subprocesses, it sets `shell=True`, concatenates the list of command elements into a single string, and makes an unconvincing try at escaping the components.  See eg #6848.

I don't know what the point is in any of this.  Surely it's simpler not to create a shell, and to let `subprocess` take care of arguments.

On the other hand I'm ony testing on Linux, and - I'm not sure - it may be that the unit tests mock this part sufficiently thoroughly that making them pass doesn't reveal very much about the real-life behaviour. 

So I thought I'd see if either the pipeline or a reviewer can tell me why this doesn't work.

### DIFF
--- a/src/poetry/utils/_compat.py
+++ b/src/poetry/utils/_compat.py
@@ -60,19 +60,11 @@ def to_str(string: str) -> str:
     return decode(string)
 
 
-def list_to_shell_command(cmd: list[str]) -> str:
-    return " ".join(
-        f'"{token}"' if " " in token and token[0] not in {"'", '"'} else token
-        for token in cmd
-    )
-
-
 __all__ = [
     "WINDOWS",
     "cached_property",
     "decode",
     "encode",
-    "list_to_shell_command",
     "metadata",
     "to_str",
     "tomllib",
--- a/src/poetry/utils/env.py
+++ b/src/poetry/utils/env.py
@@ -40,7 +40,6 @@ from virtualenv.seed.wheels.embed import get_embed_wheel
 from poetry.utils._compat import WINDOWS
 from poetry.utils._compat import decode
 from poetry.utils._compat import encode
-from poetry.utils._compat import list_to_shell_command
 from poetry.utils._compat import metadata
 from poetry.utils.helpers import get_real_windows_path
 from poetry.utils.helpers import is_dir_writable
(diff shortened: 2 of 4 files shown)


YDX-2147483647:
Just FYI:

- `list_to_shell_command` was lead in [598503](https://github.com/dimbleby/poetry/commit/598503ec271cd322976461f3100fc99318c84aca#diff-58970a6af112ab5dce31955e7d73936f8ad104ef7ae7d1ecf677f729e60206f2) to fix #3013.

- [Python doc on security considerations for `subprocess`](https://docs.python.org/3/library/subprocess.html#security-considerations):

  > `subprocess.run` will never implicitly call a system shell. This means that all characters (including shell metacharacters) can _safely_ be passed to child processes.
  >
  > If the shell is invoked explicitly (`shell=True`), it is _the application’s responsibility_ to ensure that all whitespace and metacharacters are quoted appropriately to avoid [shell injection](https://en.wikipedia.org/wiki/Shell_injection#Shell_injection) vulnerabilities. On [some platforms](https://docs.python.org/3/library/shlex.html#shlex-quote-warning), it is possible to use [shlex.quote()](https://docs.python.org/3/library/shlex.html#shlex.quote) for this escaping.

dimbleby:
in fact `list_to_shell_command` was introduced a couple of years earlier than that, at 58dbca462

however I think we are agreeing that a more reliable way of dealing with special characters is to let `subprocess` handle them, so it's all good

radoering:
There is the following paragraph in the docs:

> If shell is True, the specified command will be executed through the shell. This can be useful if you are using Python primarily for the enhanced control flow it offers over most system shells and still want convenient access to other shell features such as shell pipes, filename wildcards, environment variable expansion, and expansion of ~ to a user’s home directory.

Maybe, that's relevant for `_run`?

dimbleby:
Happy to hear about and discuss examples where this makes a negative difference; hard, though, to give a useful answer "maybe"...

radoering:
It seems `Env._run()` is never called with user input. (When calling `poetry run ...`, `Env.execute()` is called.) Thus, I assume the changes in this PR should be fine if we don't use pathes with something like `~` or `%appdata%` internally.

radoering (approved):

PR #7473, opened 2023-02-05
Cache git dependencies as wheels

maksbotan:
Currently, poetry install will clone, build and install every git dependency when it's not present in the environment. This is OK for developer's machines, but not OK for CI - there environment is always fresh, and installing git dependencies takes significant time on each CI run, especially if the dependency has C extensions that need to be built.

This commit builds a wheel for every git dependency that has precise reference hash in lock file and is not required to be in editable mode, stores that wheel in a cache dir and will install from it instead of cloning the repository again.

# Pull Request Check List

This is a copy of #6896, which got automatically closed when `feature/wheel-installer-and-builder` branch was merged and deleted.

@neersighted @radoering may I ping you about this feature? 🙏 

If it's OK, I'll try to write a test.

### DIFF
--- a/src/poetry/installation/chef.py
+++ b/src/poetry/installation/chef.py
@@ -94,8 +94,8 @@ class Chef:
             return archive
 
         if archive.is_dir():
-            tmp_dir = tempfile.mkdtemp(prefix="poetry-chef-")
-            return self._prepare(archive, Path(tmp_dir), editable=editable)
+            destination = output_dir or Path(tempfile.mkdtemp(prefix="poetry-chef-"))
+            return self._prepare(archive, destination=destination, editable=editable)
 
         return self._prepare_sdist(archive, destination=output_dir)
 
--- a/src/poetry/installation/executor.py
+++ b/src/poetry/installation/executor.py
@@ -653,7 +653,8 @@ class Executor:
             )
 
         archive = self._prepare_archive(operation, output_dir=output_dir)
-        package._source_url = original_url
+        if not package.develop:
+            package._source_url = original_url
 
         if output_dir is not None and output_dir.is_dir():
             # Mark directories with cached git packages, to distinguish from
@@ -893,12 +894,12 @@ class Executor:
 
         url_reference: dict[str, Any] | None = None
 
-        if package.source_type == "git":
+        if package.source_type == "git" and not package.develop:
             url_reference = self._create_git_url_reference(package)
+        elif package.source_type in ("directory", "git"):
+            url_reference = self._create_directory_url_reference(package)
         elif package.source_type == "url":
             url_reference = self._create_url_url_reference(package)
-        elif package.source_type == "directory":
-            url_reference = self._create_directory_url_reference(package)
         elif package.source_type == "file":
             url_reference = self._create_file_url_reference(package)
 
--- a/src/poetry/installation/executor.py
+++ b/src/poetry/installation/executor.py
@@ -529,7 +529,7 @@ class Executor:
         cleanup_archive: bool = False
         if package.source_type == "git":
             archive = self._prepare_git_archive(operation)
-            cleanup_archive = True
+            cleanup_archive = operation.package.develop
         elif package.source_type == "file":
             archive = self._prepare_archive(operation)
         elif package.source_type == "directory":
@@ -584,7 +584,9 @@ class Executor:
 
             raise
 
-    def _prepare_archive(self, operation: Install | Update) -> Path:
+    def _prepare_archive(
+        self, operation: Install | Update, *, output_dir: Path | None = None
+    ) -> Path:
         package = operation.package
         operation_message = self.get_operation_message(operation)
 
@@ -603,12 +605,28 @@ class Executor:
 
         self._populate_hashes_dict(archive, package)
 
-        return self._chef.prepare(archive, editable=package.develop)
+        return self._chef.prepare(
+            archive, editable=package.develop, output_dir=output_dir
+        )
 
     def _prepare_git_archive(self, operation: Install | Update) -> Path:
         from poetry.vcs.git import Git
 
         package = operation.package
+        assert package.source_url is not None
+
+        if package.source_resolved_reference and not package.develop:
+            # Only cache git archives when we know precise reference hash,
+            # otherwise we might get stale archives
+            cached_archive = self._artifact_cache.get_cached_archive_for_git(
+                package.source_url,
+                package.source_resolved_reference,
+                package.source_subdirectory,
+                env=self._env,
+            )
+            if cached_archive is not None:
+                return cached_archive
+
         operation_message = self.get_operation_message(operation)
 
         message = (
@@ -616,7 +634,6 @@ class Executor:
         )
         self._write(operation, message)
 
-        assert package.source_url is not None
         source = Git.clone(
             url=package.source_url,
             source_root=self._env.path / "src",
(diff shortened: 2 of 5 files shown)


maksbotan:
@sdispater may I ask your attention about this? Since 1.4.0 was released, this change can be incorporated into master. IMO this is an important improvement.

src/poetry/installation/executor.py
```diff
@@ -582,14 +582,25 @@ def _prepare_git_archive(self, operation: Install | Update) -> Path:
         from poetry.vcs.git import Git
 
         package = operation.package
+        reference = package.source_resolved_reference or package.source_reference
```
radoering, line 585: Can we be sure that reference is always a hash or tag and never a branch?

src/poetry/installation/executor.py
```diff
@@ -582,14 +582,25 @@ def _prepare_git_archive(self, operation: Install | Update) -> Path:
...
+        assert package.source_url is not None
+        assert reference is not None
+
+        cache_dir = self._chef.get_cache_directory_for_git(
+            package.source_url, reference
+        )
+        cache_dir.mkdir(parents=True, exist_ok=True)
+        cached_archive = next(cache_dir.glob("*.whl"), None)
+        if cached_archive is not None and not operation.package.develop:
+            return cached_archive
```
radoering, line 595: We should extract this to method in `ArtifactCache` similar to `get_cached_archive_for_link`.

src/poetry/installation/executor.py
```diff
@@ -604,7 +615,13 @@ def _prepare_git_archive(self, operation: Install | Update) -> Path:
 
         package._source_url = original_url
 
-        return archive
+        if operation.package.develop:
+            return archive
+        else:
+            cached_archive = cache_dir / archive.name
+            shutil.copy(archive, cached_archive)
```
radoering, line 622: Can we move instead of copy? Alternatively, we could pass an optional destination to `_prepare_archive`.

src/poetry/installation/chef.py
```diff
@@ -256,6 +256,12 @@ def get_cache_directory_for_link(self, link: Link) -> Path:
             self._env.marker_env["interpreter_version"].split(".")[:2]
         )
 
+        return self._get_directory_from_hash(key_parts)
+
+    def get_cache_directory_for_git(self, url: str, ref: str) -> Path:
```
radoering, line 261: `subdirectory` is relevant, too

radoering (changes requested):
Please note #7621 and see my review comments.

Further, I wonder if it might make sense to mark the cache directories of git dependencies somehow for debugging purposes, e.g. by creating some `created_from_git_dependency` file? This would also give you an easy way to get rid of all cached wheels for git dependencies without nuking your complete artifacts cache. 🤔

src/poetry/installation/executor.py
maksbotan: I ended up using the second suggestions, refactoring artifacts cache a bit, making git cache use the same logic to find correct archive as is done for ordinary packages. This is more safe then just grabbing the first wheel found in the cache directory.

src/poetry/installation/chef.py
maksbotan: Thanks!

maksbotan:
Thank you for the review, @radoering. I've addressed your comments in a separate commit, I'll squash everything before the merge.

> Further, I wonder if it might make sense to mark the cache directories of git dependencies somehow for debugging purposes, e.g. by creating some `created_from_git_dependency` file? This would also give you an easy way to get rid of all cached wheels for git dependencies without nuking your complete artifacts cache. 🤔

I think this would make sense. What is a proper way to touch a file in `poetry`, or should I just grab `shutil` or something for that?

maksbotan (commented):
Some tests fail after my refactoring, I'll look at that later, if the overall approach is ok.

src/poetry/installation/executor.py
radoering: Sounds good to me.

radoering:
> What is a proper way to touch a file in `poetry`, or should I just grab `shutil` or something for that?

[path.touch()](https://docs.python.org/3/library/pathlib.html#pathlib.Path.touch)

maksbotan:
@radoering i've addressed your comments and fixed failing tests.

src/poetry/installation/executor.py
```diff
@@ -603,34 +604,66 @@ def _prepare_archive(self, operation: Install | Update) -> Path:
...
 
     def _prepare_git_archive(self, operation: Install | Update) -> Path:
         from poetry.vcs.git import Git
 
         package = operation.package
+        reference = package.source_resolved_reference or package.source_reference
+        assert package.source_url is not None
+        assert reference is not None
+
+        if package.source_resolved_reference is not None:
```
radoering, line 619:
```suggestion
        if package.source_resolved_reference:
```

just in case it's an empty string

src/poetry/installation/executor.py
```diff
@@ -603,34 +604,66 @@ def _prepare_archive(self, operation: Install | Update) -> Path:
...
             f"  <fg=blue;options=bold>•</> {operation_message}: <info>Cloning...</info>"
         )
         self._write(operation, message)
 
-        assert package.source_url is not None
         source = Git.clone(
             url=package.source_url,
             source_root=self._env.path / "src",
-            revision=package.source_resolved_reference or package.source_reference,
+            revision=reference,
```
radoering, line 641: We can revert this change (and storing reference in a variable some lines above) since we only use `source_resolved_reference` for caching now.

src/poetry/installation/executor.py
```diff
@@ -603,34 +604,66 @@ def _prepare_archive(self, operation: Install | Update) -> Path:
...
         )
 
         # Now we just need to install from the source directory
         original_url = package.source_url
         package._source_url = str(source.path)
 
-        archive = self._prepare_archive(operation)
+        output_dir = None
+        if (
+            package.source_resolved_reference is not None
```
radoering, line 650:
```suggestion
            package.source_resolved_reference
```

src/poetry/utils/cache.py
```diff
@@ -231,35 +231,68 @@ def get_cache_directory_for_link(self, link: Link) -> Path:
...
 
         split_key = [key[:2], key[2:4], key[4:6], key[6:]]
-
         return self._cache_dir.joinpath(*split_key)
 
+    def get_cache_directory_for_git(
+        self, url: str, ref: str, subdirectory: str | None
+    ) -> Path:
+        return self._get_directory_from_hash(
+            {"url": url, "ref": ref, "subdirectory": subdirectory}
```
radoering, line 250: Since we add subdirectory only if it's present for links, I'd do the same here.

src/poetry/utils/cache.py
```diff
@@ -289,6 +322,9 @@ def get_cached_archive_for_link(
     def _get_cached_archives_for_link(self, link: Link) -> list[Path]:
```
radoering, line 322: Not used anymore. Can be removed.

radoering (changes requested):
LGTM after some minor changes and with some tests.

- in `test_cache.py`, we should add tests for `get_cache_directory_for_git` and `get_cached_archive_for_git`
- in `test_executor.py`, we probably should extend `test_executor_should_write_pep610_url_references_for_git` similar to `test_executor_should_write_pep610_url_references_for_wheel_urls` or `test_executor_should_write_pep610_url_references_for_non_wheel_urls`

maksbotan (commented):
Hi @radoering!

I've addressed your comments and added some tests. I did not add a test for `get_cached_archive_for_git`, as it is a wrapper around `_get_cached_archive`, which is tested elsewhere anyway. Moreover, there was no test for `get_cached_archive_for_link` anyway.

radoering:
> Moreover, there was no test for `get_cached_archive_for_link` anyway.

Actually, there are two tests for `get_cached_archive_for_link`. 😉 Apart from that, you're right. I just added a small smoke test that checks that none of the internal assertions is raised.

Further, I fixed the lost cleanup for editable git installs. Thereby, I noticed that the content of PEP 610 `direct_url.json` had been wrong for editable git installs and fixed that, too. Please take a look at my commits and verify that they do not break anything for you.

maksbotan:
@radoering thank you! 

I've run a few tests with my production use case and this branch seems to work correctly.

Should I rebase these commits now? Or perhaps squash into one?

radoering (approved):

maksbotan:
Thank you! Could you give some estimate when this can hit a release? I understand that it would be in 1.5.x series?

radoering:
It will be in 1.5. I wouldn't expect it in April. 1.4.2 was released about a week ago and we didn't do too much for 1.5 until then.

maksbotan:
Thanks.

Issue #7460, opened 2023-02-02
"hash" value in file direct_url.json uses incorrect separator


chopeen:
- **Poetry version**: 1.3.2
- **Python version**: 3.8.16
- **OS version and name**: Ubuntu 20.04.5 LTS (Focal Fossa)
- **pyproject.toml**: https://gist.github.com/chopeen/ab18e5490fad88698ac6ed2dd5ba4435

- [x] I am on the [latest](https://github.com/python-poetry/poetry/releases/latest) stable Poetry version, installed using a recommended method.
- [x] I have searched the [issues](https://github.com/python-poetry/poetry/issues) of this repo and believe that this is not a duplicate.
- [x] I have consulted the [FAQ](https://python-poetry.org/docs/faq/) and [blog](https://python-poetry.org/blog/) for any relevant entries or release notes.
- [x] If an exception occurs when executing a command, I executed it again in debug mode (`-vvv` option) and have included the output below.

## Issue

After running `poetry install`, there is a file `direct_url.json` under `en_core_web_sm-3.5.0.dist-info`, with the following contents (below I formatted it for readability):

```
{
  "url": "https://github.com/explosion/spacy-models/releases/download/en_core_web_sm-3.5.0/en_core_web_sm-3.5.0.tar.gz",
  "archive_info": {
    "hash": "sha256:63d38fecdd4290635c7af4d4f6da50902bdc6c1732ce416b55c2b76c4b0c4626"
  }
}
```

The **colon** `:` separator after `sha256` is does not comply with the [PyPA spec](https://packaging.python.org/en/latest/specifications/direct-url/#specification:~:text=A%20deprecated%20hash%20key%20(type%20string)%20MAY%20be%20present%20for%20backwards%20compatibility%20purposes%2C%20with%20value%20%3Chash%2Dalgorithm%3E%3D%3Cexpected%2Dhash%3E.):

> A deprecated hash key (type `string`) MAY be present for backwards compatibility purposes, with value `<hash-algorithm>=<expected-hash>`.

It should be **equals** sign.

Because of that issue, pip v23.0 crashes when trying to list packages in a Poetry environment that contains archives installed from URL - see https://github.com/pypa/pip/issues/11773 for details.

dimbleby:
```patch
diff --git a/src/poetry/installation/executor.py b/src/poetry/installation/executor.py
index 0496a5f8..e5582413 100644
--- a/src/poetry/installation/executor.py
[...]

PR #7475, opened 2023-02-05
Writes `hashes` in direct_url.json for archives installed from URL

chopeen:
[PyPA spec](https://packaging.python.org/en/latest/specifications/direct-url/#:~:text=When%20url%20refers%20to%20a%20source%20archive%20or%20a%20wheel%2C%20the%20archive_info%20key%20MUST) says:

> - A `hashes` key SHOULD be present as a dictionary mapping a hash name to a hex encoded digest of the file.
> - A deprecated `hash` key (type `string`) MAY be present for backwards compatibility purposes, with value `<hash-algorithm>=<expected-hash>`.

Initially, I added both `hashes` and deprecated `hash`. However, the latter was causing the `mypy` check to fail. Nothing major though, I could fix it, but I am not sure if there is any actual backwards compatibility reason that would justify the extra complexity of having both `hashes` and `hash` with the same file digest in two different formats.

Please advise - is there a reason to keep the deprecated `hash` key?

Resolves: #7460


sbidoul:
> Please advise - is there a reason to keep the deprecated hash key?

`hashes` was added to the spec very recently (1 week ago, actually).

OTOH, since the `hash` key emitted by poetry was incorrect and nobody complained, I suppose we can assume that it was not used, or parsed in a non spec-compliant way. Either way, consumers would need to be updated, so I think it's fine to drop it in favor of the new `hashes` key.

src/poetry/installation/executor.py
```diff
@@ -906,3 +898,12 @@ def _create_directory_url_reference(self, package: Package) -> dict[str, Any]:
             "url": Path(package.source_url).as_uri(),
             "dir_info": dir_info,
         }
+
+    def _get_archive_info(self, package: Package) -> dict[str, Any]:
+        archive_info = {}
```
radoering, line 903: Can add a docstring with a link to https://packaging.python.org/en/latest/specifications/direct-url/ and info that it supersedes PEP 610, please?

tests/installation/test_executor.py
```diff
@@ -701,12 +701,27 @@ def test_executor_should_write_pep610_url_references_for_urls(
         source_type="url",
         source_url="https://files.pythonhosted.org/demo-0.1.0-py2.py3-none-any.whl",
     )
+    # Set package.files so the executor will attempt to hash the package
```
radoering, line 704: Can you extend `test_executor_should_write_pep610_url_references_for_files` in a similar way, please?

radoering (changes requested):
Thanks for your contribution. 😃

> Please advise - is there a reason to keep the deprecated hash key?

It seems we didn't read back the hash key ourselves. Further, I'm not aware that anybody consumed our (non spec-compliant) `hash` key. Thus, I agree it's probably fine to drop it.

tests/installation/test_executor.py
chopeen: For this to work, I needed to populate the `_hashes` dictionary inside function `_prepare_archive`. Hope that's fine.

radoering: LGTM. Apparently, we did write the hash only for URL dependencies but not for file dependencies so far because of the missing part in `_prepare_archive`.

radoering (approved):

Issue #6799, opened 2022-10-14
Poetry ignoring credentials in source url?


keriksson-rosenqvist:
### Discussed in https://github.com/orgs/python-poetry/discussions/6629

<div type='discussions-op-text'>

<sup>Originally posted by **keriksson-rosenqvist** September 26, 2022</sup>
I am trying to set up a poetry environment which pulls a certain package from a private pypi-like registry. As I am sharing the project with others via a git repository, I want the setup to be contained within the pyproject.toml file, i.e. not rely on configs, environment variables, or keyring settings that may differ between user devices.
I have therefore added a `source` with the basic http credentials included in the url. The exact same url is able to install the package and it's dependencies through `pip install -i https://<username:password>@[REDACTED]/pypi/simple/ my-private-package`

```
[tool.poetry.dependencies]
my-private-package = { version = "^2.0.0", source = "my_private_package"}
...
[[tool.poetry.source]]
name = "my_private_package"
url = "https://<username:password>@[REDACTED]/pypi/simple/"
secondary = true
```

I've set `secondary = true` as that lets dependencies be installed from the normal pypi, otherwise poetry also fails to install them.  

According to [this issue](https://github.com/python-poetry/poetry/issues/4389), others have been seemingly been able to get it working with the credentials in the url, however with the setup above raises the following error for me.
```
$ poetry update
Updating dependencies
Resolving dependencies... (1.7s)

401 Client Error:  for url: https://[REDACTED]/pypi/download/my-private-package/2/my_private_package-2.0.0-py3-none-any.whl#sha256=[REDACTED]
```
Is this a bug, am I setting it up wrong, or has there been a change to the usage since the mentioned issue was able to get it to work?</div>

kristang:
Try running `poetry update -vvv` and post the output (remember to redact anything you don't want public - like your feed address) 

Does `pip install -i https://[REDACTED]/pypi/simple/ my-private-package` fail with a 401?

keriksson-rosenqvist:
@kristang I've made an MWE using a new `hello-world` package in my private pipy registry. Running `poetry update -vvv` gives the following trace:  
```
Loading configuration file [REDACTED]/.config/pypoetry/config.toml
Loading configuration file [REDACTED]/.config/pypoetry/auth.toml
Adding repository hello_world (https://<username:password>@[REDACTED]/hello_world/pypi/simple) and setting it as secondary
Using virtualenv: [REDACTED]/poetry_credential_mwe/.venv
[...]

PR #7486, opened 2023-02-08
Read credentials from repository URL

squirrel532:
# Pull Request Check List

Resolves: #6799

This PR allows user to encode their credential of a private repository into URL.


src/poetry/utils/authenticator.py
```diff
@@ -91,7 +91,12 @@ class AuthenticatorRepositoryConfig:
 
     def __post_init__(self) -> None:
         parsed_url = urllib.parse.urlsplit(self.url)
-        self.netloc = parsed_url.netloc
+        # Remove username:password from netloc string
+        if "@" in parsed_url.netloc:
+            self.netloc = parsed_url.netloc.rsplit("@", 1)[-1]
+        else:
+            self.netloc = parsed_url.netloc
+
```
nfnvsc, line 99:
```suggestion
        self.netloc = parsed_url.netloc
```
I think that in order to ensure the consistency of netloc you shouldn't remove the login info from it.

src/poetry/utils/authenticator.py
```diff
@@ -104,10 +109,26 @@ def http_credential_keys(self) -> list[str]:
...
+        if "@" in parsed_url.netloc:
+            # Split from the right because that's how urllib.parse.urlsplit()
+            # behaves if more than one @ is present (which can be checked using
+            # the password attribute of urlsplit()'s return value).
+            auth, netloc = parsed_url.netloc.rsplit("@", 1)
+            # Split from the left because that's how urllib.parse.urlsplit()
+            # behaves if more than one : is present (which again can be checked
+            # using the password attribute of the return value)
+            user, password = auth.split(":", 1) if ":" in auth else (auth, "")
+            credential = HTTPAuthCredential(user, password)
```
nfnvsc, line 123:
```suggestion
        # try with the URL
        if "@" in self.netloc:
            # Split from the right because that's how urllib.parse.urlsplit()
            # behaves if more than one @ is present (which can be checked using
            # the password attribute of urlsplit()'s return value).
            auth, netloc = self.netloc.rsplit("@", 1)
            # Split from the left because that's how urllib.parse.urlsplit()
            # behaves if more than one : is present (which again can be checked
            # using the password attribute of the return value)
            user, password = auth.split(":", 1) if ":" in auth else (auth, "")
            credential = HTTPAuthCredential(
                    urllib.parse.unquote(user),
                    urllib.parse.unquote(password),
                )
```
You can then use the netloc from the class itself to retrive auth data.

src/poetry/utils/authenticator.py
```diff
@@ -104,10 +109,26 @@ def http_credential_keys(self) -> list[str]:
...
+        else:
+            credential = HTTPAuthCredential(None, None)
+
+        if credential.username is None and credential.password is None:
+            # try with the repository name via the password manager
+            credential = HTTPAuthCredential(
+                **(password_manager.get_http_auth(self.name) or {})
+            )
```
nfnvsc, line 131:
```suggestion
        else:
            # try with the repository name via the password manager
            credential = HTTPAuthCredential(
                **(password_manager.get_http_auth(self.name) or {})
            )
```
No need to create a new `HTTPAuthCredential(None, None)`

src/poetry/utils/authenticator.py
```diff
@@ -365,6 +386,8 @@ def get_credentials_for_url(self, url: str) -> HTTPAuthCredential:
                     urllib.parse.unquote(password),
                 )
 
+        print(url)
+        print(self._credentials)
```
nfnvsc, line 390:
```suggestion
```

nfnvsc (commented):
I would then suggest to make the repository verification:
https://github.com/python-poetry/poetry/blob/f92370e470b898ad4de3f0994d9a4a6a29e1213f/src/poetry/utils/authenticator.py#L455
To this:
`if repository.netloc.split("@")[-1] == parsed_url.netloc:`

So that we ensure that repositories ignore eventual authentication in the netloc.

src/poetry/utils/authenticator.py
```diff
@@ -82,10 +82,25 @@ def http_credential_keys(self) -> list[str]:
     def get_http_credentials(
         self, password_manager: PasswordManager, username: str | None = None
     ) -> HTTPAuthCredential:
-        # try with the repository name via the password manager
-        credential = HTTPAuthCredential(
-            **(password_manager.get_http_auth(self.name) or {})
-        )
+        # try with the URL
+        if "@" in self.netloc:
```
amp-matthew-wallace, line 86: This logic already exists in `get_credentials_for_url()` and shouldn't need to be copied.

Secrus:
Stalled

Issue #395, opened 2018-08-15
If installation of a dependency fails, it remains in the lock file until it is updated with `poetry lock`


13steinj:
- [X] I am on the [latest](https://github.com/sdispater/poetry/releases/latest) Poetry version.
- [X] I have searched the [issues](https://github.com/sdispater/poetry/issues) of this repo and believe that this is not a duplicate.
- [X] If an exception occurs when executing a command, I executed it again in debug mode (`-vvv` option).

- **OS version and name**: Windows 10 64-bit
- **Poetry version**: 0.11.4
- **Link of a [Gist](https://gist.github.com/) with the contents of your pyproject.toml file**: not really applicable

## Issue
When installation of a dependency fails, such as for example, it requires a system component that you do not have to complete the build process, the package/hash combo is still added to the pyproject.lock file, even though the dependency has not installed.

The pyproject.toml file *does not* incorrectly retain the dependency.

Example: `poetry add -D yappi` without the MSVC++ 14.0 build tools causes a VenvCommandError. This is fine and expected. However, after running this, the following lines was added to my pyproject.lock:

```toml
[[package]]
category = "dev"
description = "Yet Another Python Profiler"
name = "yappi"
optional = false
platform = "UNKNOWN"
python-versions = "*"
version = "0.98"
```

and 

```
yappi = ["5f657129e1b9b952379ffbc009357d0dcdb58c50f3bfe88ffbb992e4b27b263c"]
```

My pyproject.toml was unaffected, as mentioned.

Running `poetry lock` clears this anomaly.

sdispater:
I agree this is an undesired behavior.

The installer should be more clever when rollbacking if an error occurs during installation.

wagnerluis1982:
I can reproduce this bug on latest (1.2.1). I tried to install a difficult dependency on Linux, called `pyaudio`.

```console
$ poetry add pyaudio
Using version ^0.2.12 for PyAudio

Updating dependencies
Resolving dependencies... (0.5s)

Writing lock file

Package operations: 1 install, 0 updates, 0 removals

  • Installing pyaudio (0.2.12): Failed

...stack tracebacks...
```

```console
$ git st -s
 M poetry.lock
```
```patch
$ git diff
diff --git a/poetry.lock b/poetry.lock
index f1d2d23..a26c47d 100644
--- a/poetry.lock
+++ b/poetry.lock
@@ -224,6 +224,17 @@ category = "dev"
 optional = false
[...]

PR #7498, opened 2023-02-10
Only write lock file when installation is success

wagnerluis1982:
# Pull Request Check List

Resolves: #395 

## How it appear to users

The user experience is almost the same, but "Writing lock file" only appears in the end of the process, and only if the installation process succeeds.

### On success ("Writing lock file" displays after installation)
```console
$ poetry add jinja2      
Using version ^3.1.2 for jinja2

Updating dependencies

  • Installing markupsafe (2.1.2)
  • Installing jinja2 (3.1.2)

Writing lock file
```

### On failure ("Writing lock file" is not displayed)
```console
$ poetry add pyaudio
Using version ^0.2.13 for pyaudio

Updating dependencies

  • Installing pyaudio (0.2.13): Failed

  ChefBuildError

  Backend subprocess exited when trying to invoke build_wheel

  ...bdist logs omitted for brevity...

Note: This error originates from the build backend, and is likely not a problem with poetry but with pyaudio (0.2.13) not supporting PEP 517 builds. You can verify this by running 'pip wheel --use-pep517 "pyaudio (==0.2.13)"'.

```

### DIFF
--- a/src/poetry/installation/installer.py
+++ b/src/poetry/installation/installer.py
@@ -291,12 +291,10 @@ class Installer:
         lockfile_repo = LockfileRepository()
         self._populate_lockfile_repo(lockfile_repo, ops)
 
-        if self._update:
+        if self._lock and self._update:
+            # If we are only in lock mode, no need to go any further
             self._write_lock_file(lockfile_repo)
-
-            if self._lock:
-                # If we are only in lock mode, no need to go any further
-                return 0
+            return 0
 
         if self._groups is not None:
             root = self._package.with_dependency_groups(list(self._groups), only=True)
@@ -362,7 +360,13 @@ class Installer:
         self._filter_operations(ops, lockfile_repo)
 
         # Execute operations
-        return self._execute(ops)
+        status = self._execute(ops)
+
+        if status == 0 and self._update:
+            # Only write lock file when installation is success
+            self._write_lock_file(lockfile_repo)
+
+        return status
 
     def _write_lock_file(self, repo: LockfileRepository, force: bool = False) -> None:
         if self._write_lock and (force or self._update):
--- a/tests/console/commands/test_add.py
+++ b/tests/console/commands/test_add.py
@@ -2233,7 +2233,6 @@ Resolving dependencies...
 
   BuildError
 
-Writing lock file
 """
 
     assert poetry_with_up_to_date_lockfile.file.read() == original_pyproject_content
--- a/tests/console/commands/test_add.py
+++ b/tests/console/commands/test_add.py
@@ -2159,7 +2159,8 @@ def test_add_keyboard_interrupt_restore_content(
     tester = command_tester_factory("add", poetry=poetry_with_up_to_date_lockfile)
 
     mocker.patch(
-        "poetry.installation.installer.Installer.run", side_effect=KeyboardInterrupt()
+        "poetry.installation.installer.Installer._execute",
+        side_effect=KeyboardInterrupt(),
     )
     original_pyproject_content = poetry_with_up_to_date_lockfile.file.read()
     original_lockfile_content = poetry_with_up_to_date_lockfile._locker.lock_data
--- a/tests/console/commands/test_add.py
+++ b/tests/console/commands/test_add.py
@@ -16,6 +16,8 @@ from tests.helpers import get_package
 
 
 if TYPE_CHECKING:
+    from typing import Any
+
     from cleo.testers.command_tester import CommandTester
     from pytest_mock import MockerFixture
 
@@ -2200,3 +2202,40 @@ def test_add_with_dry_run_keep_files_intact(
     assert (
         poetry_with_up_to_date_lockfile._locker.lock_data == original_lockfile_content
     )
+
+
+def test_add_should_not_change_lock_file_when_dependency_installation_fail(
+    poetry_with_up_to_date_lockfile: Poetry,
+    repo: TestRepository,
+    command_tester_factory: CommandTesterFactory,
+    mocker: MockerFixture,
+):
+    tester = command_tester_factory("add", poetry=poetry_with_up_to_date_lockfile)
+
+    repo.add_package(get_package("docker", "4.3.1"))
+    repo.add_package(get_package("cachy", "0.2.0"))
+
+    original_pyproject_content = poetry_with_up_to_date_lockfile.file.read()
+    original_lockfile_content = poetry_with_up_to_date_lockfile.locker.lock_data
+
+    def error(_: Any) -> int:
+        tester.io.write("\n  BuildError\n\n")
+        return 1
+
+    mocker.patch("poetry.installation.installer.Installer._execute", side_effect=error)
+    tester.execute("cachy")
+
+    expected = """\
+Using version ^0.2.0 for cachy
+
+Updating dependencies
+Resolving dependencies...
+
+  BuildError
+
+Writing lock file
+"""
+
+    assert poetry_with_up_to_date_lockfile.file.read() == original_pyproject_content
+    assert poetry_with_up_to_date_lockfile.locker.lock_data == original_lockfile_content
+    assert tester.io.fetch_output() == expected
--- a/tests/console/commands/test_add.py
+++ b/tests/console/commands/test_add.py
@@ -72,11 +72,11 @@ Using version ^0.2.0 for cachy
 Updating dependencies
 Resolving dependencies...
 
-Writing lock file
-
 Package operations: 1 install, 0 updates, 0 removals
 
   • Installing cachy (0.2.0)
+
+Writing lock file
 """
 
     assert tester.io.fetch_output() == expected
@@ -102,11 +102,11 @@ Using version ^0.2.0 for cachy
 Updating dependencies
 Resolving dependencies...
 
-Writing lock file
-
 Package operations: 1 install, 0 updates, 0 removals
 
   • Installing cachy (0.2.0)
+
+Writing lock file
 """
     assert tester.io.fetch_output() == expected
     assert tester.command.installer.executor.installations_count == 1
@@ -121,11 +121,11 @@ Package operations: 1 install, 0 updates, 0 removals
 Updating dependencies
 Resolving dependencies...
 
-Writing lock file
-
 Package operations: 1 install, 0 updates, 0 removals
 
   • Installing cachy (0.1.0)
+
+Writing lock file
 """
     assert tester.io.fetch_output() == expected
 
@@ -169,11 +169,11 @@ def test_add_equal_constraint(
 Updating dependencies
 Resolving dependencies...
 
-Writing lock file
-
 Package operations: 1 install, 0 updates, 0 removals
 
   • Installing cachy (0.1.0)
+
+Writing lock file
 """
 
     assert tester.io.fetch_output() == expected
@@ -193,11 +193,11 @@ def test_add_greater_constraint(
 Updating dependencies
 Resolving dependencies...
 
-Writing lock file
-
 Package operations: 1 install, 0 updates, 0 removals
 
   • Installing cachy (0.2.0)
+
+Writing lock file
 """
 
     assert tester.io.fetch_output() == expected
@@ -227,12 +227,12 @@ def test_add_constraint_with_extras(
 Updating dependencies
 Resolving dependencies...
 
-Writing lock file
-
 Package operations: 2 installs, 0 updates, 0 removals
 
   • Installing msgpack-python (0.5.3)
   • Installing cachy (0.1.0)
+
+Writing lock file
 """
 
     assert tester.io.fetch_output() == expected
@@ -257,12 +257,12 @@ def test_add_constraint_dependencies(
 Updating dependencies
 Resolving dependencies...
 
-Writing lock file
-
 Package operations: 2 installs, 0 updates, 0 removals
 
   • Installing msgpack-python (0.5.3)
   • Installing cachy (0.2.0)
+
+Writing lock file
 """
 
     assert tester.io.fetch_output() == expected
@@ -287,12 +287,12 @@ def test_add_git_constraint(
 Updating dependencies
 Resolving dependencies...
 
-Writing lock file
-
 Package operations: 2 installs, 0 updates, 0 removals
 
   • Installing pendulum (1.4.4)
   • Installing demo (0.1.2 9cf87a2)
+
+Writing lock file
 """
 
     assert tester.io.fetch_output() == expected
@@ -323,12 +323,12 @@ def test_add_git_constraint_with_poetry(
 Updating dependencies
 Resolving dependencies...
 
-Writing lock file
-
 Package operations: 2 installs, 0 updates, 0 removals
 
   • Installing pendulum (1.4.4)
   • Installing demo (0.1.2 9cf87a2)
+
+Writing lock file
 """
 
     assert tester.io.fetch_output() == expected
@@ -356,14 +356,14 @@ def test_add_git_constraint_with_extras(
 Updating dependencies
 Resolving dependencies...
 
-Writing lock file
-
 Package operations: 4 installs, 0 updates, 0 removals
 
   • Installing cleo (0.6.5)
   • Installing pendulum (1.4.4)
   • Installing tomlkit (0.5.5)
   • Installing demo (0.1.2 9cf87a2)
+
+Writing lock file
 """
 
     assert tester.io.fetch_output().strip() == expected.strip()
@@ -402,11 +402,11 @@ def test_add_git_constraint_with_subdirectory(
 Updating dependencies
 Resolving dependencies...
 
-Writing lock file
-
 Package operations: 1 install, 0 updates, 0 removals
 
   • Installing two (2.0.0 9cf87a2)
+
+Writing lock file
 """
     assert tester.io.fetch_output().strip() == expected.strip()
     assert tester.command.installer.executor.installations_count == 1
@@ -446,12 +446,12 @@ def test_add_git_ssh_constraint(
 Updating dependencies
 Resolving dependencies...
 
-Writing lock file
-
 Package operations: 2 installs, 0 updates, 0 removals
 
   • Installing pendulum (1.4.4)
   • Installing demo (0.1.2 9cf87a2)
+
+Writing lock file
 """
 
     assert tester.io.fetch_output() == expected
@@ -493,12 +493,12 @@ def test_add_directory_constraint(
 Updating dependencies
 Resolving dependencies...
 
-Writing lock file
-
 Package operations: 2 installs, 0 updates, 0 removals
 
   • Installing pendulum (1.4.4)
   • Installing demo (0.1.2 {app.poetry.file.parent.joinpath(path).resolve().as_posix()})
+
+Writing lock file
 """
 
     assert tester.io.fetch_output() == expected
@@ -534,12 +534,12 @@ def test_add_directory_with_poetry(
 Updating dependencies
 Resolving dependencies...
 
-Writing lock file
-
 Package operations: 2 installs, 0 updates, 0 removals
 
   • Installing pendulum (1.4.4)
   • Installing demo (0.1.2 {app.poetry.file.parent.joinpath(path).resolve().as_posix()})
+
+Writing lock file
 """
 
     assert tester.io.fetch_output() == expected
@@ -566,12 +566,12 @@ def test_add_file_constraint_wheel(
 Updating dependencies
 Resolving dependencies...
 
-Writing lock file
-
 Package operations: 2 installs, 0 updates, 0 removals
 
   • Installing pendulum (1.4.4)
   • Installing demo (0.1.0 {app.poetry.file.parent.joinpath(path).resolve().as_posix()})
+
+Writing lock file
 """
 
     assert tester.io.fetch_output() == expected
@@ -602,12 +602,12 @@ def test_add_file_constraint_sdist(
 Updating dependencies
 Resolving dependencies...
 
-Writing lock file
-
 Package operations: 2 installs, 0 updates, 0 removals
 
   • Installing pendulum (1.4.4)
   • Installing demo (0.1.0 {app.poetry.file.parent.joinpath(path).resolve().as_posix()})
+
+Writing lock file
 """
 
     assert tester.io.fetch_output() == expected
@@ -642,12 +642,12 @@ def test_add_constraint_with_extras_option(
 Updating dependencies
 Resolving dependencies...
 
-Writing lock file
-
 Package operations: 2 installs, 0 updates, 0 removals
 
   • Installing msgpack-python (0.5.3)
   • Installing cachy (0.2.0)
+
+Writing lock file
 """
 
     assert tester.io.fetch_output() == expected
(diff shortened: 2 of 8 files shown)


wagnerluis1982:
## Summary

This fixes the leaking `poetry.lock` left changed incorrectly when `poetry add` installation fails (does not affect when using option `--lock`).

## Code changes

The rollback happens in reaction to `Installer.run` status. When exit status is not good (i.e. different of zero), it removes the dependencies being added from the relevant objects (`DependencyGroup` and `Locker` data) and the installer runs again in lock only mode.

## Comments

In order to accomplish, I had to make a few refactorings.

### src/poetry/packages/locker.py

In the following, the locker object is "copied" using the constructor whatever the class is. This was done this way to leverage other implementations, e.g. `TestLocker` for the test cases. The problem here is `TestLocker` contains a few artificial state (such as `_write` and `_locked`) that is needed for my test case.

https://github.com/python-poetry/poetry/blob/7b40c844f523246738850b5ec9537dcf58ac7043/src/poetry/console/commands/add.py#L244-L247

So, I introduced to the `Locker` a method `copy_with` with same signature of constructor.

### tests/helpers.py

I overrode the new `copy_with` in the `TestLocker`, but also copying the artificial state.

### tests/console/commands/test_add.py

I added a call to "poetry lock" in the fixture `poetry_with_up_to_date_lockfile`, in order to the lock file uses the latest syntax of poetry.

wagnerluis1982:
My investigation towards this bug also raised one question about the **validity** of one existing test in `tests/console/commands/test_add.py`.

### test_add_keyboard_interrupt_restore_content

In the following, how can the lock be ever changed if `Installer.run` (which is the one method which changes the lock file) is never even called (it is mocked here). So there is nothing really to be restored.

https://github.com/python-poetry/poetry/blob/8558f0cb94c11d5d1f901900d4a6ea8a8167bd35/tests/console/commands/test_add.py#L2150-L2172

dimbleby:
#395 was scoped only at not leaving the lock file inconsistent with the pyproject.toml.  This goes further in also rolling back changes to the virtual environment.

I'm not sure whether this is a good idea.  Though I can see how this might be desirable, it adds complexity and edge cases. (What if the rollback itself fails?  What if the user had previously gone `pip install some-tool` in the environment and the dependencies being removed had overlap with `some-tool`'s dependencies?  in that second case, the environment after rollback won't be restored to its original state)

The failed install left the user with no hint that the lockfile might be inconsistent with pyproject.toml.  No question that #395 identifies a bug here that wants fixing.

However, the failed install made plenty of logs saying "Installing package-foo" etc and so it should be obvious to a user that their environment has changed.   Given the difficulty of reliably restoring exactly the previous environment, it's simpler and safer to let the user decide what to do about that.

I'd suggest limiting the scope of this MR to leaving `pyproject.toml` and `poetry.lock` in a good state

wagnerluis1982:
> I'd suggest limiting the scope of this MR to leaving `pyproject.toml` and `poetry.lock` in a good state

I don't doubt this PR requires some polishing, but this is exactly what I'm doing here, no? In case of installation failure, I'm not removing any package, not really. I'm cleaning up the objects in the scope of `poetry add` and running installer in **lock only mode** to keep the lock consistent.

wagnerluis1982:
Oh, now I realized the source of confusion. Here I'm not in no way making changes to environment.

I mean, AFAIK installer in lock only mode don't change the environment, so I'm relying on it.

dimbleby:
Oh, perhaps I misread.  Still I am surprised that this fix requires any re-locking: I'd expect to arrange things so that the file isn't written (or is explicitly rewritten with a saved copy of the old file) in case of failure - more like the handling of `pyproject.toml`.

Also maybe cf #7401 - it seems to be uncertain whether anyone likes that enough to merge it as a whole, but the `locker.refresh()` method there looks like it might be an improvement that could be useful in this one too.

dimbleby:
What about `poetry update`?  Should that update the lockfile, if installing updated packages fails?  

Maybe you already looked into this, perhaps a better place to catch all such examples would be the Installer's `_do_install()` method.  That could defer writing the lockfile until it knew that it had successfully executed operations.

wagnerluis1982:
> What about `poetry update`? Should that update the lockfile, if installing updated packages fails?
> 
> Maybe you already looked into this, perhaps a better place to catch all such examples would be the Installer's `_do_install()` method. That could defer writing the lockfile until it knew that it had successfully executed operations.

Good points. Excellent points, actually. This will require a further investigation, and probably a more complex change. At least I already have the test 😸

dimbleby:
looking at your latest draft, I'd definitely suggest that simpler than backing up and restoring the lockfile would be to refactor `_do_install()` so that the write happens later, and only on success.

`do_install()` is certainly kinda messy today, if you can find a sensible way to extract something out of it so that it becomes a bit mroe readable that would probably be welcome too!

wagnerluis1982:
> looking at your latest draft, I'd definitely suggest that simpler than backing up and restoring the lockfile would be to refactor `_do_install()` so that the write happens later, and only on success.

Thank you very much for looking at the draft and for the suggestion.

Lazy as I am, I was avoiding to make any rewrite, but I was definitely starting to think along these lines 😸

> `do_install()` is certainly kinda messy today, if you can find a sensible way to extract something out of it so that it becomes a bit mroe readable that would probably be welcome too!

I have a few ideas I'll try this week.

But `do_install` mess isn't the only problem here. To work with the `Locker` is a lot confusing.

wagnerluis1982:
@dimbleby when you have a chance, check my latest update.

I am applying a delay in the write to lock file by creating an internal `Locker` wrapper that holds the lock data in memory is signalized that lock data is good to be dumped to file.

wagnerluis1982:
Remaining to do:
- [ ] Add test to `test_installer`
- [ ] Add warning log informing the lock file changes were not applied
- [ ] Think about the refactoring in `_do_install`

I need advice on the warning log as I am not sure what to write :thought_balloon:

wagnerluis1982:
I was also thinking there is a need to check if lock file is writable. Before my changes, it's known in advance.

Now we install and only in the end we get aware of the issue...

wagnerluis1982:
Okay, I feel this is ready for a review.

There are only 2 things I am uncertain:
1. The message "Writing lock file" is now a lie, since the lock is not anymore written at that place, but if moved to the end, it will change the existing output, where "Writing lock file" appears before install the dependencies. Since I am not sure this is desirable, I kept the output as is.
2. Related to the previous, I think we can keep the existing "Writing lock file" and add on the rollback a message "Rolling back lock file". Despite the fact the lock file was never written on installation file, I feel that is a good message to appear.

@dimbleby I am still thinking on a possible refactor on `_do_install`, but, if I do, I will do in context of a separate PR (that code is currently driving me nut).

tests/console/commands/test_add.py
```diff
@@ -32,16 +32,25 @@
...
-    return project_factory(
+    poetry = project_factory(
         name="foobar",
         pyproject_content=(source / "pyproject.toml").read_text(encoding="utf-8"),
         poetry_lock_content=(source / "poetry.lock").read_text(encoding="utf-8"),
     )
 
+    # ensure poetry.lock with updated syntax
+    repo.add_package(get_package("docker", "4.3.1"))
+    command_tester_factory("lock", poetry=poetry).execute()
```
wagnerluis1982, line 50: This change was required because the existing lock file is using an older version of the lock file.

I need to have the same version in order to compare that the lock file was untouched.

tests/helpers.py
```diff
@@ -188,7 +187,7 @@ def reset_poetry(self) -> None:
 
 class TestLocker(Locker):
     def __init__(self, lock: str | Path, local_config: dict) -> None:
-        self._lock = TOMLFile(lock)
+        self._lock = Path(lock)
```
wagnerluis1982, line 190: This change was required, because there is no reason to keep `TOMFile` when `Locker._lock` is actually a `Path`.

No test is relying on `_lock` be a `TOMLFile`.

tests/installation/test_installer.py
```diff
@@ -102,17 +102,18 @@ def load(
 
 class Locker(BaseLocker):
     def __init__(self, lock_path: str | Path) -> None:
-        self._lock = TOMLFile(Path(lock_path).joinpath("poetry.lock"))
+        self._lock = Path(lock_path.joinpath("poetry.lock"))
```
wagnerluis1982, line 105: This change was required, because there is no reason to keep `TOMFile` when `Locker._lock` is actually a `Path`.

No test is relying on `_lock` be a `TOMLFile`.

tests/installation/test_installer.py
```diff
@@ -102,17 +102,18 @@ def load(
...
         self._written_data = None
         self._locked = False
+        self._lock_data = None
```
wagnerluis1982, line 108: Reason: some tests were failing due to the lack of this property (that exists on parent `Locker`)

tests/installation/test_installer.py
```diff
@@ -102,17 +102,18 @@ def load(
...
         self._content_hash = self._get_content_hash()
 
     @property
     def written_data(self) -> dict | None:
         return self._written_data
 
     def set_lock_path(self, lock: str | Path) -> Locker:
-        self._lock = TOMLFile(Path(lock).joinpath("poetry.lock"))
+        self._lock = Path(lock.joinpath("poetry.lock"))
```
wagnerluis1982, line 116: Reason: the same as above in the `__init__`

tests/installation/test_installer_old.py
```diff
@@ -54,7 +54,7 @@ def load(
 
 class Locker(BaseLocker):
     def __init__(self, lock_path: str | Path) -> None:
-        self._lock = TOMLFile(Path(lock_path).joinpath("poetry.lock"))
+        self._lock = Path(lock_path.joinpath("poetry.lock"))
```
wagnerluis1982, line 57: This change was required, because there is no reason to keep `TOMFile` when `Locker._lock` is actually a `Path`.

No test is relying on `_lock` be a `TOMLFile`.

wagnerluis1982:
I just added a "Rolling back lock file" message. Check in the link below to see what you think.

In addition, this is a demo that shows the lock file was not changed :smile: 

---

[asciicast](https://asciinema.org/a/zsEWM1nhG2W1wlwSDVN3OFS1I)

dimbleby:
I don't much like it!  This adds a pile of code that I don't think ought to be needed.

- _If_ the _TransactionalLocker is a useful API, let's just change the Locker rather than wrapping it up
- But is it needed?  can't you just rearrange `_do_install()` so that the call to write the lockfile happens later (and only on success)?

wagnerluis1982:
> I don't much like it! This adds a pile of code that I don't think ought to be needed.
> 
> * _If_ the _TransactionalLocker is a useful API, let's just change the Locker rather than wrapping it up
> * But is it needed?  can't you just rearrange `_do_install()` so that the call to write the lockfile happens later (and only on success)?

It is a good question. I tried to rearrange `_do_install`, but every attempt fails. Because the remaining operations in `_do_install` (and `_execute` I think) relies on the updated information provided by `locker.lock_data`.

Right now the `Locker` API has no way to write the `_lock_data` without writing to the file. So, I felt that I should keep the `Locker` intact (apart from the little refactoring) and created the wrapper.

All in all, I am open to discussions, here or in Discord :slightly_smiling_face:

wagnerluis1982:
Update: I incorporated the transactional state into `Locker`.

All existing `Locker` operations was kept the same behavior. If `start_transaction` is called, the lock data is not written until `commit`.

I kept the implementation as simple as possible.

wagnerluis1982:
I realized that indeed the lock file can be written after installation without any transaction mechanism. I let the failures in my initial investigations misled me, which turned to my current solution. Ashamed :no_mouth: 

Now I have a working solution that do that, but there is a side effect in the output: "Writing lock file" message is shifted to the end (see below).

Is this desired???

**Current output**
```console
Using version ^1.1.2 for foo

Updating dependencies
Resolving dependencies...

Writing lock file

  - Installing foo (1.1.2)
```

**New output writing lock file in the end**
```console
Using version ^1.1.2 for foo

Updating dependencies
Resolving dependencies...

  - Installing foo (1.1.2)

Writing lock file
```

radoering:
IMO the "writing lock file" message should be printed when the lock file is written. If it is written before installation it should be printed before. If it is written after installation it should be printed afterwards.

wagnerluis1982:
In that case, I will publish the solution postponing lock file writing to the end.

wagnerluis1982:
Solution published!

Once this gets approved, it will also need update in the website:
image

This would be here, right?

https://github.com/python-poetry/website/blob/9aab35bcc573f353c81d69e9d92412d0c1133669/layouts/home.html#L65-L80

wagnerluis1982:
This proper lock file writing caught a test case with a wrong assertion.

I moved back to `master` and set output on the following (rather than `NullIO`) and found a solver issue. Since test case isn't checking output, it was missed!

https://github.com/python-poetry/poetry/blob/3e32a4f83e89b2b2bede68e0d375a70931b04a8e/tests/installation/test_installer.py#L1147-L1166

```console
Updating dependencies
Resolving dependencies...

Package operations: 7 installs, 0 updates, 0 removals

  • Installing six (1.11.0)
  • Installing attrs (17.4.0)
  • Installing more-itertools (4.1.0)
  • Installing pluggy (0.6.0)
  • Installing py (1.5.3)
  • Installing setuptools (39.2.0)

  SolveFailure

  Because -root- depends on wheel (*) which doesn't match any versions, version solving failed.

  at src/poetry/mixology/version_solver.py:349 in _resolve_conflict
      345│             )
      346│             self._log(f'! which is caused by "{most_recent_satisfier.cause}"')
      347│             self._log(f"! thus: {incompatibility}")
      348│ 
    → 349│         raise SolveFailure(incompatibility)
      350│ 
      351│     def _choose_package_version(self) -> str | None:
      352│         """
      353│         Tries to select a version of a required package.

  SolverProblemError

  Because -root- depends on wheel (*) which doesn't match any versions, version solving failed.

  at src/poetry/puzzle/solver.py:163 in _solve
      159│             packages = result.packages
      160│         except OverrideNeeded as e:
      161│             return self._solve_in_compatibility_mode(e.overrides)
      162│         except SolveFailure as e:
    → 163│             raise SolverProblemError(e)
      164│ 
      165│         combined_nodes = depth_first_search(PackageNode(self._package, packages))
      166│         results = dict(aggregate_package_nodes(nodes) for nodes in combined_nodes)
      167│ 
```

wagnerluis1982:
I pasted on last comment, by mistake, an incorrect output (missing "Writing lock file" line), I edited it to be right.

dimbleby:
sounds like the next step is to figure out why that testcase is hitting a solver failure.  Maybe it's right to do so and there's just something about the test setup that's wrong, maybe it's a real bug.

dimbleby:
pretty sure that `-root-` comes from https://github.com/python-poetry/poetry/blob/40061f9d1351a9f76d6bdc1db8b3d903864c1373/src/poetry/installation/chef.py#L69, and I expect the dependency on `wheel` is via the default build requirement for something that is being prepared by that testcase.

radoering:
This is because the pool with the mock repository is not generated via `create_pool()` from the config, but created directly and passed to the installer. Before #7560 this pool wasn't used to create the isolated environment for building dependencies but a new pool was created from the config, which does not contain the mock repository. (In real-world scenarios both pools were created from the same config and thus were equal.) With #7560 the pool is passed from the installer to the chef.

Thus, it's just an issue of the test that #7560 plus your change revealed.

wagnerluis1982:
> Thus, it's just an issue of the test that #7560 plus your change revealed.

Thank you for the explanation.

Now I'm trying to figure out how to fix the test, but so far no relevant progress, I don't understand how the solver operates at all.

radoering:
I assume you "just" have to add two json files for `wheel` in https://github.com/python-poetry/poetry/tree/master/tests/repositories/fixtures/pypi.org/json

See https://github.com/python-poetry/poetry/blob/master/tests/repositories/fixtures/pypi.org/json/setuptools.json and https://github.com/python-poetry/poetry/blob/master/tests/repositories/fixtures/pypi.org/json/setuptools/39.2.0.json for example.

That's where the ancient setuptools 39.2.0 in https://github.com/python-poetry/poetry/pull/7498#issuecomment-1450943161 comes from.

wagnerluis1982:
Thank, @radoering, I got the idea now.

But this is getting weirder, I added the `wheel` json files as instructed and the test produced a very confused output.

Notice it was logged "Installing setuptools (39.2.0)", but I have a dependency on setuptools??? It makes no sense.

```console
Updating dependencies
Resolving dependencies...

Writing lock file

Package operations: 7 installs, 0 updates, 0 removals

  • Installing six (1.11.0)
  • Installing attrs (17.4.0)
  • Installing more-itertools (4.1.0)
  • Installing pluggy (0.6.0)
  • Installing py (1.5.3)
  • Installing setuptools (39.2.0)

  SolveFailure

  Because -root- depends on setuptools (>=40.8.0) which doesn't match any versions, version solving failed.
```

wagnerluis1982:
I also made a different investigation by switching off my network and noticed this test is requiring network to work. See below.

Was that supposed to be???

```console
Updating dependencies
Resolving dependencies...

Writing lock file

Package operations: 7 installs, 0 updates, 0 removals

  • Installing six (1.11.0)

  gaierror

  [Errno -3] Temporary failure in name resolution

  at /usr/lib/python3.8/socket.py:918 in getaddrinfo
      914│     """
      915│     # We override this function since we want to translate the numeric family
      916│     # and socket type values to enum constants.
      917│     addrlist = []
    → 918│     for res in _socket.getaddrinfo(host, port, family, type, proto, flags):
      919│         af, socktype, proto, canonname, sa = res
      920│         addrlist.append((_intenum_converter(af, AddressFamily),
      921│                          _intenum_converter(socktype, SocketKind),
      922│                          proto, canonname, sa))

  NewConnectionError

  <urllib3.connection.HTTPSConnection object at 0x7f973c75afd0>: Failed to establish a new connection: [Errno -3] Temporary failure in name resolution

  at ~/.cache/pypoetry/virtualenvs/poetry-Iqs9p7vq-py3.8/lib/python3.8/site-packages/urllib3/connection.py:186 in _new_conn
      182│                 % (self.host, self.timeout),
      183│             )
      184│ 
      185│         except SocketError as e:
    → 186│             raise NewConnectionError(
      187│                 self, "Failed to establish a new connection: %s" % e
      188│             )
      189│ 
      190│         return conn

  MaxRetryError

  HTTPSConnectionPool(host='files.pythonhosted.org', port=443): Max retries exceeded with url: /packages/67/4b/141a581104b1f6397bfa78ac9d43d8ad29a7ca43ea90a2d863fe3056e86a/six-1.11.0-py2.py3-none-any.whl (Caused by NewConnectionError('<urllib3.connection.HTTPSConnection object at 0x7f973c75afd0>: Failed to establish a new connection: [Errno -3] Temporary failure in name resolution'))

  at ~/.cache/pypoetry/virtualenvs/poetry-Iqs9p7vq-py3.8/lib/python3.8/site-packages/urllib3/util/retry.py:592 in increment
      588│             history=history,
      589│         )
      590│ 
      591│         if new_retry.is_exhausted():
    → 592│             raise MaxRetryError(_pool, url, error or ResponseError(cause))
      593│ 
      594│         log.debug("Incremented Retry for (url='%s'): %r", url, new_retry)
      595│ 
      596│         return new_retry

  ConnectionError

  HTTPSConnectionPool(host='files.pythonhosted.org', port=443): Max retries exceeded with url: /packages/67/4b/141a581104b1f6397bfa78ac9d43d8ad29a7ca43ea90a2d863fe3056e86a/six-1.11.0-py2.py3-none-any.whl (Caused by NewConnectionError('<urllib3.connection.HTTPSConnection object at 0x7f973c75afd0>: Failed to establish a new connection: [Errno -3] Temporary failure in name resolution'))

  at ~/.cache/pypoetry/virtualenvs/poetry-Iqs9p7vq-py3.8/lib/python3.8/site-packages/requests/adapters.py:565 in send
      561│             if isinstance(e.reason, _SSLError):
      562│                 # This branch is for urllib3 v1.22 and later.
      563│                 raise SSLError(e, request=request)
      564│ 
    → 565│             raise ConnectionError(e, request=request)
      566│ 
      567│         except ClosedPoolError as e:
      568│             raise ConnectionError(e, request=request)
      569│
```

wagnerluis1982:
I didn't include the wheel files. They are in another branch that I am experimenting.

- https://github.com/wagnerluis1982/poetry/blob/613fda0a9f61c4d0260158077dd6a6fba0dc9f1d/tests/repositories/fixtures/pypi.org/json/wheel.json
- https://github.com/wagnerluis1982/poetry/blob/613fda0a9f61c4d0260158077dd6a6fba0dc9f1d/tests/repositories/fixtures/pypi.org/json/wheel/0.31.0.json

dimbleby:
> I have a dependency on setuptools??? It makes no sense.

like the earlier dependency on wheel, this will be a default build requirement for a package that needs building and that does not specify its build requirements.

certainly it's undesirable for these tests to rely on having an internet connection

wagnerluis1982:
> like the earlier dependency on wheel, this will be a default build requirement for a package that needs building and that does not specify its build requirements.

I will take your word.

Still, don't you think that's weird the solver to show to the user
```
  • Installing setuptools (39.2.0)
```
and in the end to display
```
  SolveFailure

  Because -root- depends on setuptools (>=40.8.0) which doesn't match any versions, version solving failed.
```

dimbleby:
pretty sure it's just an unlucky coincidence that this testcase happens to pick up a dependency on setuptools 39.2.0, while some other dependency has (by default) a build-time dependency on setuptools 40.8.0.  

Agree that it's not completely obvious that this is what is happening, if you haven't got enough to fix in this MR already feel free also to make the messages better!

radoering:
For example, we should replace `-root-` with the actual name of the package we are trying to build if possible. Originally, I thought this comes from https://github.com/python-poetry/poetry/blob/40061f9d1351a9f76d6bdc1db8b3d903864c1373/src/poetry/installation/chef.py#L69 but it's spelled differently so it may be unrelated...

dimbleby:
it definitely comes from there, `-root-` is `__root__` after normalisation

wagnerluis1982:
@dimbleby thanks. I'm still interested. Will do the rebase ASAP.

src/poetry/installation/installer.py
```diff
@@ -291,12 +291,10 @@ def _do_install(self) -> int:
         lockfile_repo = LockfileRepository()
         self._populate_lockfile_repo(lockfile_repo, ops)
 
-        if self._update:
+        if self._lock and self._update:
```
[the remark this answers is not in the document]
wagnerluis1982: It would make sense, however it's not true, since you can run `poetry lock --no-update`, which ends up in `self._lock = True` and `self._update = False`.

Relevant extracts below from **commands/lock.py** and **installer.py**

https://github.com/python-poetry/poetry/blob/161b19cb4e4686fc5a0a7925001534a87f6c4052/src/poetry/console/commands/lock.py#L55

https://github.com/python-poetry/poetry/blob/161b19cb4e4686fc5a0a7925001534a87f6c4052/src/poetry/installation/installer.py#L153-L161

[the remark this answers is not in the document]
wagnerluis1982: Perhaps will be clearer if I swap the conditions?

```suggestion
        if self._update and self._lock:
```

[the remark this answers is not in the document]
radoering: Arghh, too many options and a confusing mapping. I always forget one or the other. Good catch! 👍

I don't think the order makes a difference.

radoering (approved):

Issue #7488, opened 2023-02-08
Why is it unnecessary for libraries to commit the lock file?


iainelder:
https://github.com/python-poetry/poetry/blob/2c610de46805f6f0a31aacaf394c598a3bd9e122/docs/basic-usage.md?plain=1#L231

Why not?

The preceding paragraph basically says committing the lock file makes it easier to reproduce the same experience for all users of the code.

Why would it be different for libraries?

The botocove package uses Poetry to manage its dependencies. botocove is a library for querying AWS, intended to be used in scripts. We are now [wondering whether to keep the lock file or not](https://github.com/connelldave/botocove/pull/59#issuecomment-1422411473).

dimbleby:
You committing the lockfile will make no difference at all to the users of your library, who will consume your library via pypi.  The metadata on pypi is derived from the requirements in pyproject.toml.

iainelder:
Thanks, @dimbleby . That makes sense.

For us library developers, I think it still makes sense to keep the lockfile for the sake of reproducibility.

Is there something else I'm missing?

dimbleby:
keeping the lockfile locally is a trade-off: it's true that you get reproducibility, but it also means that if some new dependency has been released that breaks things for consumers of your library, you're likely to be the last to find out about it.

I guess either choice is legit, but the usual advice is for libraries not to commit lockfiles.

eg here's Rust saying the same thing about Cargo.lock https://doc.rust-lang.org/cargo/faq.html#why-do-binaries-have-cargolock-in-version-control-but-not-libraries

iainelder:
Thanks for the Cargo docs reference. That helps a lot because the Cargo.lock file plays the same role.

Your comment, though, is the one that changes my mind:

> if some new dependency has been released that breaks things for consumers of your library, you're likely to be the last to find out about it.

For my project I had suggested periodically explicitly updating versions declared in the lock file and in the pre-commit config (another place to specify build dependencies).

But now it seems that by discarding the lockfile, I can get that behavior automatically at least for the lock file.

Thanks for your help.

How can we adapt your comment for the Poetry documentation?

neersighted:
You'll see that Poetry does the same thing, incidentally -- we commit our lock files in both poetry-core (consumed as a library only) and Poetry (an application installed with normal version constraints). It keeps our CI reproducible/prevents random changes in other projects from blocking development, but we still have to refresh our lock file regularly and stay up-to-date to make sure we don't break suddenly for users.

[...]

PR #7506, opened 2023-02-12
Explain whether to commit poetry.lock

iainelder:
Summarizes discussion from https://github.com/python-poetry/poetry/issues/7488.

Resolves: #7488

### DIFF
--- a/docs/basic-usage.md
+++ b/docs/basic-usage.md
@@ -219,7 +219,11 @@ but they may not all be at the very latest available versions
 (some dependencies listed in the `poetry.lock` file may have released newer versions since the file was created).
 This is by design, it ensures that your project does not break because of unexpected changes in dependencies.
 
-### Commit your `poetry.lock` file to version control
+### Committing your `poetry.lock` file to version control
+
+#### As an application developer
+
+Application developers commit `poetry.lock` to get more reproducible builds.
 
 Committing this file to VC is important because it will cause anyone who sets up the project
 to use the exact same versions of the dependencies that you are using.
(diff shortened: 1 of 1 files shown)


docs/basic-usage.md
```diff
@@ -227,9 +231,15 @@ Even if you develop alone, in six months when reinstalling the project you can f
...
-{{% /note %}}
+#### As a library developer
+
+Library developers have more to consider. Your users are application developers, and your library will run in a Python environment you don't control.
+
+The application ignores your library's lock file. It can use whatever dependency version meets the constraints in your `pyproject.toml`. The application will probably use the latest compatible dependency version. If your library's `poetry.lock` falls behind some new dependency version that breaks things for your users, you're likely to be the last to find out about it.
+
+To work with your library as users are likely to, omit the `poetry.lock` file.
+
+If you still want the reproducibility benefits, consider a regular refresh of `poetry.lock` to stay up-to-date and reduce the risk of sudden breakage for users.
```
radoering, line 242: Another benefit of committing the lock file is that your CI will run faster. (If you have many dependencies locking can take a while.)

radoering, line 242:
```suggestion
A simple way to avoid such a scenario is to omit the `poetry.lock` file. However, by doing so, you sacrifice reproducibility and performance to a certain extent. Without a lockfile, it can be difficult to find the reason for failing tests, because in addition to obvious code changes an unnoticed library update might the culprit. Further, Poetry has to lock before installing a dependency if `poetry.lock` has been omitted. Depending on the number of dependencies, locking may take a significant amount of time.

If you do not want to give up the reproducibility and performance benefits, consider a regular refresh of `poetry.lock` to stay up-to-date and reduce the risk of sudden breakage for users.
```

iainelder:
Thanks @radoering for adding more details.

radoering (approved):

PR #7517, opened 2023-02-15
docs: document Poetry's own versioning scheme

radoering:
Some people expect poetry to adhere to strict semver, presumably because it used to promote semver (or even still promotes it in the docs). During the last releases, we realized that strict semver does not work well for poetry (anymore). Because of its large user base, even small changes not considered harmful can be breaking for somebody. Sticking to strict semver and always bumping the major version does not seem desirable/reasonable.

Since a versioning scheme that is more like Python's versioning than strict semver has emerged during the last releases and has been mentioned in the chat sometimes, we might as well document it somewhere. (I didn't come up with a better place than the FAQ.)

Further, if poetry itself does not adhere to strict semver, we shouldn't aggressively promote it in the docs.

### DIFF
--- a/docs/faq.md
+++ b/docs/faq.md
@@ -33,6 +33,28 @@ Once Poetry has cached the releases' information on your machine, the dependency
 will be much faster.
 {{% /note %}}
 
+### What kind of versioning scheme does Poetry use for itself?
+
+Poetry uses "major.minor.micro" version identifiers as mentioned in
+[PEP 440](https://peps.python.org/pep-0440/#final-releases).
+
+Version bumps are done similar to Python's versioning:
+* A major version bump (incrementing the first number) is only done for breaking changes
+  if a deprecation cycle is not possible and many users have to perform some manual steps
+  to migrate from one version to the next one.
+* A minor version bump (incrementing the second number) may include new features as well
+  as new deprecations and drop features deprecated in an earlier minor release.
+* A micro version bump (incrementing the third number) usually only includes bug fixes.
+  Deprecated features will not be dropped in a micro release.
+
+### Why does Poetry not adhere to semantic versioning?
+
+Because of its large user base, even small changes not considered relevant by most users
+can turn out to be a breaking change for some users in hindsight.
+Sticking to strict [semantic versioning](https://semver.org) and (almost) always bumping
+the major version instead of the minor version does not seem desirable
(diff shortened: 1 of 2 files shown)


dimbleby:
Updating "how poetry the application is versioned" feels a lot less controversial than updating "how python projects should think about versioning in general" - the latter has been quite the can of worms!

It's probably not the worst idea for poetry to revisit its wider stance on versioning.  (Whatever the docs say, defaulting to upper bounds for everything is a strong commitment to believing in semantic versioning; I for one have mixed feelings about this). 

But do you intend that discussion to be in scope for this merge request?  Or is this supposed to be limited to describing the policy for poetry's own version?

radoering:
The main purpose of this PR is to describe the policy for poetry's own version. However, it felt kind of contradictive to claim that strict semver doesn't work for us and advocate for semver some sections later. Thus, I tried to change the concerning parts of the docs to be more neutral.

I agree that `poetry add` defaulting to caret constraints is somehow connected. On the other hand, it's just a default and a weaker commitment than saying "you should use upper bounds" in the docs (at least in my opinion).

It might be the next step to change this default. (I'm not a big fan of this default myself.) However, I think this can be discussed later (not in the scope of this PR).

dimbleby:
Yeah, I'm not objecting exactly to anything here.  Rather just cautioning that - for better or worse - it risks encouraging a discussion that has not happened for a while. 

eg I wouldn't be at all surprised to see someone showing up with an argument along the lines of "how come poetry is defaulting the world into semantic versioning when even your own docs don't encourage it any more?"

docs/faq.md
```diff
@@ -33,21 +33,54 @@ Once Poetry has cached the releases' information on your machine, the dependency
...
+However, when defining an upper bound, users of your package are not able to update
+a dependency beyond the upper bound even if it does not break anything
+and is fully compatible with your package.
+You have to release a new version of your package with an increased upper bound first.
+
+If your package will be used as a library in other packages, it might be better to avoid
+upper bounds and thus unnecessary dependency conflicts (unless you already know for sure
+that the next release of the dependency will break your package).
+If your package will be used as an application, it might be worth to define an upper bound.
+
```
neersighted, line 83: This should probably be a separate commit in the same PR.

neersighted (dismissed):

neersighted (approved):

PR #7560, opened 2023-02-25
fix: don't merge sources into config

radoering:
Fixes a bug introduced via https://github.com/python-poetry/poetry/commit/55127c893db5bf8ab33b494e0c4adc8c9b09c458 in #6205. Merging the sources from pyproject.toml into the global configuration implicitly allows to define sources globally (in an unintuitive way, but nevertheless...), which was surely not intended.

Instead of merging sources into the global config and creating a new repository pool for each install of an sdist from the global config, we can just reuse the pool passed to the executor. This should be a minor performance improvement.

### DIFF
--- a/tests/console/commands/self/conftest.py
+++ b/tests/console/commands/self/conftest.py
@@ -1,6 +1,7 @@
 from __future__ import annotations
 
 from typing import TYPE_CHECKING
+from typing import Any
 from typing import Callable
 
 import pytest
@@ -14,6 +15,8 @@ from poetry.utils.env import EnvManager
 
 
 if TYPE_CHECKING:
+    from collections.abc import Iterable
+
     import httpretty
 
     from cleo.io.io import IO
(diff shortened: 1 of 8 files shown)


tests/console/commands/self/conftest.py
```diff
@@ -44,9 +47,9 @@ def pool(repo: TestRepository) -> RepositoryPool:
 
 def create_pool_factory(
     repo: Repository,
-) -> Callable[[Config, IO, bool], RepositoryPool]:
+) -> Callable[[Config, Iterable[str, Any], IO, bool], RepositoryPool]:
     def _create_pool(
-        config: Config, io: IO, disable_cache: bool = False
+        config: Config, sources: Iterable[str, Any], io: IO, disable_cache: bool = False
```
dimbleby, line 52: wrong type

radoering:
I added a test that fails without this fix. (That's how I noticed the issue in the first place.)

neersighted (approved):

Issue #7567, opened 2023-02-27
Crash with "too many values to unpack" when adding package with `poetry init`


ssteinerx:
- **Poetry version**:  1.3.2
- **Python version**: 3.11.2
- **OS version and name**: Ubuntu 22.10
- **pyproject.toml**: not created, crash
 

- [x] I am on the [latest](https://github.com/python-poetry/poetry/releases/latest) stable Poetry version, installed using a recommended method.
- [x] I have searched the [issues](https://github.com/python-poetry/poetry/issues) of this repo and believe that this is not a duplicate.
- [x] I have consulted the [FAQ](https://python-poetry.org/docs/faq/) and [blog](https://python-poetry.org/blog/) for any relevant entries or release notes.
- [x] If an exception occurs when executing a command, I executed it again in debug mode (`-vvv` option) and have included the output below.
```
Package to add or search for (leave blank to skip): now we crash
Invalid package definition.

Package to add or search for (leave blank to skip): maybe now
Adding maybe now

Add a package (leave blank to skip): how about this

  Stack trace:

  12  .venv/lib/python3.11/site-packages/cleo/application.py:327 in run
       325│ 
       326│             try:
     → 327│                 exit_code = self._run(io)
       328│             except BrokenPipeError:
       329│                 # If we are piped to another process, it may close early and send a

  11  .venv/lib/python3.11/site-packages/poetry/console/application.py:190 in _run
       188│         self._load_plugins(io)
       189│ 
     → 190│         exit_code: int = super()._run(io)
       191│         return exit_code
       192│ 

  10  .venv/lib/python3.11/site-packages/cleo/application.py:431 in _run
       429│             io.input.interactive(interactive)
       430│ 
     → 431│         exit_code = self._run_command(command, io)
       432│         self._running_command = None
       433│ 

   9  .venv/lib/python3.11/site-packages/cleo/application.py:473 in _run_command
       471│ 
       472│         if error is not None:
     → 473│             raise error
       474│ 
       475│         return terminate_event.exit_code

   8  .venv/lib/python3.11/site-packages/cleo/application.py:457 in _run_command
       455│ 
       456│             if command_event.command_should_run():
     → 457│                 exit_code = command.run(io)
       458│             else:
       459│                 exit_code = ConsoleCommandEvent.RETURN_CODE_DISABLED

   7  .venv/lib/python3.11/site-packages/cleo/commands/base_command.py:119 in run
       117│         io.input.validate()
       118│ 
[...]

PR #7569, opened 2023-02-27
Fix: Misconfiguration in ask dependency interactive

PabloEmidio:
Fixing a misunderstand of buitin str function: [maxsplit can return maxsplit+1 items](https://docs.python.org/3.8/library/stdtypes.html#str.split)

Resolves: #7567

---

Although, this alteration resolves the broken message reported, we'd be better to treat the string using regex and validating the received result.

### DIFF
--- a/src/poetry/console/commands/init.py
+++ b/src/poetry/console/commands/init.py
@@ -292,6 +292,11 @@ You can specify a package in the following forms:
             )
             question.set_validator(self._validate_package)
 
+            follow_up_question = self.create_question(
+                "\nAdd a package (leave blank to skip):"
+            )
+            follow_up_question.set_validator(self._validate_package)
+
             package = self.ask(question)
             while package:
                 constraint = self._parse_requirements([package])[0]
@@ -303,9 +308,7 @@ You can specify a package in the following forms:
                 ):
                     self.line(f"Adding <info>{package}</info>")
                     result.append(constraint)
-
-                    self.line("")
-                    package = self.ask(question)
+                    package = self.ask(follow_up_question)
                     continue
 
                 canonicalized_name = canonicalize_name(constraint["name"])
@@ -373,8 +376,7 @@ You can specify a package in the following forms:
                     result.append(constraint)
 
                 if self.io.is_interactive():
-                    self.line("")
-                    package = self.ask(question)
+                    package = self.ask(follow_up_question)
 
             return result
 
--- a/src/poetry/console/commands/init.py
+++ b/src/poetry/console/commands/init.py
@@ -303,7 +303,9 @@ You can specify a package in the following forms:
                 ):
                     self.line(f"Adding <info>{package}</info>")
                     result.append(constraint)
-                    package = self.ask("\nAdd a package (leave blank to skip):")
+
+                    self.line("")
+                    package = self.ask(question)
                     continue
 
                 canonicalized_name = canonicalize_name(constraint["name"])
@@ -371,7 +373,8 @@ You can specify a package in the following forms:
                     result.append(constraint)
 
                 if self.io.is_interactive():
-                    package = self.ask("\nAdd a package (leave blank to skip):")
+                    self.line("")
+                    package = self.ask(question)
 
             return result
 
--- a/tests/console/commands/test_init.py
+++ b/tests/console/commands/test_init.py
@@ -595,9 +595,6 @@ pytest = "^3.6.0"
 def test_interactive_with_wrong_dependency_inputs(
     tester: CommandTester, repo: TestRepository
 ):
-    repo.add_package(get_package("pendulum", "2.0.0"))
-    repo.add_package(get_package("pytest", "3.6.0"))
-
     inputs = [
         "my-package",  # Package name
         "1.2.3",  # Version
@@ -608,9 +605,6 @@ def test_interactive_with_wrong_dependency_inputs(
         "",  # Interactive packages
         "foo 1.19.2",
         "pendulum 2.0.0 foo",  # Package name and constraint (invalid)
-        "pendulum 2.0.0",  # Package name and constraint (invalid)
-        "pendulum 2.0.0",  # Package name and constraint (invalid)
-        "pendulum 2.0.0",  # Package name and constraint (invalid)
         "pendulum@^2.0.0",  # Package name and constraint (valid)
         "",  # End package selection
         "",  # Interactive dev packages
--- a/tests/console/commands/test_init.py
+++ b/tests/console/commands/test_init.py
@@ -606,6 +606,7 @@ def test_interactive_with_wrong_dependency_inputs(
         "MIT",  # License
         "^3.8",  # Python
         "",  # Interactive packages
+        "foo 1.19.2",
         "pendulum 2.0.0 foo",  # Package name and constraint (invalid)
         "pendulum 2.0.0",  # Package name and constraint (invalid)
         "pendulum 2.0.0",  # Package name and constraint (invalid)
@@ -633,6 +634,7 @@ packages = [{include = "my_package"}]
 
 [tool.poetry.dependencies]
 python = "^3.8"
+foo = "1.19.2"
 pendulum = "^2.0.0"
 
 [tool.poetry.group.dev.dependencies]
--- a/src/poetry/utils/dependency_specification.py
+++ b/src/poetry/utils/dependency_specification.py
@@ -114,7 +114,7 @@ def _parse_dependency_specification_simple(
     require: DependencySpec = {}
 
     if " " in pair:
-        name, version = pair.split(" ", 2)
+        name, version = pair.split(" ", 1)
         extras_m = re.search(r"\[([\w\d,-_]+)\]$", name)
         if extras_m:
             extras = [e.strip() for e in extras_m.group(1).split(",")]


radoering:
Can you add a test, please?

PabloEmidio:
@radoering debbuging the source code, I find the follow behavior as used this PR change:

```python
Package to add or search for (leave blank to skip): now we crash
> /home/utrape/Documents/programming/contributions/poetry/src/poetry/console/commands/init.py(484)_validate_package()
-> if package and len(package.split()) > 2:
(Pdb) c
Invalid package definition.
Package to add or search for (leave blank to skip): about now
> /home/utrape/Documents/programming/contributions/poetry/src/poetry/console/commands/init.py(484)_validate_package()
-> if package and len(package.split()) > 2:
(Pdb) c
Adding about now

Add a package (leave blank to skip): here we go
Adding here we go
```

Clearly the validation has not been called at some asking.

Working on it.

PabloEmidio:
I modify an existing test to add the scenary reported. When runned against master generate the follow output:

```python
=========================================================================== FAILURES ============================================================================
_________________________________________________________ test_interactive_with_wrong_dependency_inputs _________________________________________________________
[gw3] linux -- Python 3.11.2 /home/utrape/.cache/pypoetry/virtualenvs/poetry-X0uBKQkB-py3.11/bin/python

tester = <cleo.testers.command_tester.CommandTester object at 0x7f2628c88ed0>, repo = <tests.helpers.TestRepository object at 0x7f2628c49950>

    def test_interactive_with_wrong_dependency_inputs(
        tester: CommandTester, repo: TestRepository
    ):
        repo.add_package(get_package("pendulum", "2.0.0"))
        repo.add_package(get_package("pytest", "3.6.0"))
    
        inputs = [
            "my-package",  # Package name
            "1.2.3",  # Version
            "This is a description",  # Description
            "n",  # Author
            "MIT",  # License
            "^3.8",  # Python
            "",  # Interactive packages
            "foo 1.19.2",
            "pendulum 2.0.0 foo",  # Package name and constraint (invalid)
            "pendulum 2.0.0",  # Package name and constraint (invalid)
            "pendulum 2.0.0",  # Package name and constraint (invalid)
            "pendulum 2.0.0",  # Package name and constraint (invalid)
            "pendulum@^2.0.0",  # Package name and constraint (valid)
            "",  # End package selection
            "",  # Interactive dev packages
            "pytest 3.6.0 foo",  # Dev package name and constraint (invalid)
            "pytest 3.6.0",  # Dev package name and constraint (invalid)
            "pytest@3.6.0",  # Dev package name and constraint (valid)
            "",  # End package selection
            "\n",  # Generate
        ]
>       tester.execute(inputs="\n".join(inputs))

/home/utrape/Documents/programming/contributions/poetry/tests/console/commands/test_init.py:623: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
/home/utrape/.cache/pypoetry/virtualenvs/poetry-X0uBKQkB-py3.11/lib/python3.11/site-packages/cleo/testers/command_tester.py:88: in execute
    self._status_code = self._command.run(self._io)
/home/utrape/.cache/pypoetry/virtualenvs/poetry-X0uBKQkB-py3.11/lib/python3.11/site-packages/cleo/commands/base_command.py:119: in run
    status_code = self.execute(io)
/home/utrape/.cache/pypoetry/virtualenvs/poetry-X0uBKQkB-py3.11/lib/python3.11/site-packages/cleo/commands/command.py:62: in execute
    return self.handle()
/home/utrape/Documents/programming/contributions/poetry/src/poetry/console/commands/init.py:207: in handle
    self._format_requirements(self._determine_requirements([]))
/home/utrape/Documents/programming/contributions/poetry/src/poetry/console/commands/init.py:297: in _determine_requirements
    constraint = self._parse_requirements([package])[0]
/home/utrape/Documents/programming/contributions/poetry/src/poetry/console/commands/init.py:438: in _parse_requirements
    return [
/home/utrape/Documents/programming/contributions/poetry/src/poetry/console/commands/init.py:439: in <listcomp>
    parse_dependency_specification(
/home/utrape/Documents/programming/contributions/poetry/src/poetry/utils/dependency_specification.py:218: in parse_dependency_specification
    or _parse_dependency_specification_simple(requirement)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

requirement = 'pendulum 2.0.0 foo'

    def _parse_dependency_specification_simple(
        requirement: str,
    ) -> DependencySpec | None:
        extras: list[str] = []
        pair = re.sub("^([^@=: ]+)(?:@|==|(?<![<>~!])=|:| )(.*)$", "\\1 \\2", requirement)
        pair = pair.strip()
    
        require: DependencySpec = {}
    
        if " " in pair:
>           name, version = pair.split(" ", 2)
E           ValueError: too many values to unpack (expected 2)

/home/utrape/Documents/programming/contributions/poetry/src/poetry/utils/dependency_specification.py:117: ValueError
```

src/poetry/console/commands/init.py
```diff
@@ -303,7 +303,9 @@ def _determine_requirements(
                 ):
                     self.line(f"Adding <info>{package}</info>")
                     result.append(constraint)
-                    package = self.ask("\nAdd a package (leave blank to skip):")
+
```
dimbleby, line 306: Do I understand correctly that the current code uses different text for the second and subsequent packages, and you have changed it so that it uses the same text for all?

I wonder if that might be a little confusing as a UI: it could feel as though the first package you add "didn't work".  Anyway it seems like a behaviour change that's unrelated to what you're wanting to fix

tests/console/commands/test_init.py
```diff
@@ -606,6 +606,7 @@ def test_interactive_with_wrong_dependency_inputs(
         "MIT",  # License
         "^3.8",  # Python
         "",  # Interactive packages
+        "foo 1.19.2",
```
dimbleby, line 609: what is going on with this testcase?  why is "foo 1.19.2" valid but "pendulum 2.0.0" is invalid?  why do is "pendulum 2.0.0" tested three times?

src/poetry/console/commands/init.py
PabloEmidio: Yeah, I reused the created question that is set the validation. I did not find the text behavior a change to be confusing, but it's not difficult to add an ask with the validate set using the text in the diff above.

tests/console/commands/test_init.py
PabloEmidio: Yeah, now it's bugging me.
This changes was add in https://github.com/python-poetry/poetry/pull/5838 with by @mkniewallner with @neersighted review. Maybe one of them could bring us some understanding.

PabloEmidio: Another point is that the `foo` package is not added in repo.add_package above. And testing locally with `repo._packages.clear()` before testing seems to not cause effect. Is it the `TestRepository` being considered at all?

neersighted: Looks like a simple oversight, please feel free to clean it up.

src/poetry/console/commands/init.py
PabloEmidio: @dimbleby could you give follow up feedback?

dimbleby: I don't really have anything to add to what I said before

- IMO the new behaviour is a little more confusing, I could imagine that for a new user the experience of being asked to add a package, providing a name, and then being asked the exact same question might feel as though it just hadn't worked
- regardless of whether it really is more confusing or not (reasonable people might disagree on such things, and I don't feel especially strongly about it anyway): it's a behaviour change that is unrelated to the fix that you're making

So I'd prefer to see that change undone.

However I'm not going to lose any sleep over it either way.

PabloEmidio:
Could anyone give any thought about it's something missing or it can be merged?

radoering:
Since there is a fix in `parse_dependency_specification` in might make sense to extend `test_parse_dependency_specification`.

radoering:
I think poetry should not crash with a ValueError. (I didn't check if such an error is caught somewhere.) The best response would be to tell the user that the dependency specification is invalid and ask again for a package.

PabloEmidio:
> I think poetry should not crash with a ValueError. (I didn't check if such an error is caught somewhere.) The best response would be to tell the user that the dependency specification is invalid and ask again for a package.

@radoering, I suggest the `ValueError` because the functions that called this code raise ValueError in x scenarios.

We could catch the ValueError in the `_determine_requirements` and ask again:

```python
# src/poetry/console/commands/init.py:L302
try:
    constraint = self._parse_requirements([package])[0]
except ValueError as err:
    self.line_error(f"<error>{str(err)}</error>")
    package = self.ask(question)
    continue
```

radoering:
It seems `_validate_package` already handles this. If I provoke an `Invalid package definition.`, poetry asks for the next package. Maybe, we can just do the complete parsing, i.e. call `_parse_requirements`, in `_validate_package`?

PabloEmidio:
@radoering exactly. The central idea is the requirement already was validate before been passed to `parse_dependency_specification`. But, it has been the central idea for while and we got 2 issue about it. And anyhow, the quote bellow continue to be true and the right way to split in max 2 itens is passing maxsplit=1.

> Fixing a misunderstand of buitin str function: [maxsplit can return maxsplit+1 items](https://docs.python.org/3.8/library/stdtypes.html#str.split)

PabloEmidio:
looking the application as a whole, the currently code is going to work property. But looking only to `_parse_dependency_specification_simple` and knowing what return expect, the function does not attend its purpose.

radoering (approved):
Now, I realize that due to the fix in `init.py` the fix in `dependency_specification.py` does not matter anymore (but is more correct nevertheless) because there will always be just one space.

With that knowledge I think this PR is fine as is.

Issue #7575, opened 2023-02-28
"subdirectory" argument ignored


oyvinev:
- **Poetry version**: 1.4.0
- **Python version**: 3.8-10
- **OS version and name**: Ubuntu 20.04
- **pyproject.toml**: https://gist.github.com/oyvinev/4df478d8dee2fbdce3128043d4666468

- [x] I am on the [latest](https://github.com/python-poetry/poetry/releases/latest) stable Poetry version, installed using a recommended method.
- [x] I have searched the [issues](https://github.com/python-poetry/poetry/issues) of this repo and believe that this is not a duplicate.
- [x] I have consulted the [FAQ](https://python-poetry.org/docs/faq/) and [blog](https://python-poetry.org/blog/) for any relevant entries or release notes.
- [x] If an exception occurs when executing a command, I executed it again in debug mode (`-vvv` option) and have included the output below.

```
Using virtualenv: /home/oyvinev/.cache/pypoetry/virtualenvs/gpbuilder-rFATLHJD-py3.8
Installing dependencies from lock file

Finding the necessary packages for the current system

Package operations: 1 install, 0 updates, 0 removals, 1 skipped

  • Installing numpy (1.16.6): Pending...
  • Installing numpy (1.16.6): Skipped for the following reason: Already installed
  • Installing pyd4 (0.3.6.2 39001f4): Pending...
  • Installing pyd4 (0.3.6.2 39001f4): Cloning...
  • Installing pyd4 (0.3.6.2 39001f4): Building...
  • Installing pyd4 (0.3.6.2 39001f4): Preparing...
  • Installing pyd4 (0.3.6.2 39001f4): Failed

  Stack trace:

  10  ~/.local/lib/python3.8/site-packages/poetry/installation/executor.py:271 in _execute_operation
       269│ 
       270│             try:
     → 271│                 result = self._do_execute_operation(operation)
       272│             except EnvCommandError as e:
       273│                 if e.e.returncode == -2:

   9  ~/.local/lib/python3.8/site-packages/poetry/installation/executor.py:357 in _do_execute_operation
       355│             return 0
       356│ 
     → 357│         result: int = getattr(self, f"_execute_{method}")(operation)
       358│ 
       359│         if result != 0:

   8  ~/.local/lib/python3.8/site-packages/poetry/installation/executor.py:477 in _execute_install
       475│ 
       476│     def _execute_install(self, operation: Install | Update) -> int:
     → 477│         status_code = self._install(operation)
[...]

PR #7580, opened 2023-03-01
fix: append subdirectory to archive path (#7575)

harrymander:
Resolves: #7575

From @nikowatari's suggestion (https://github.com/python-poetry/poetry/issues/755#issuecomment-1449275865) with added test.

### DIFF
--- a/src/poetry/installation/executor.py
+++ b/src/poetry/installation/executor.py
@@ -569,6 +569,8 @@ class Executor:
 
         assert package.source_url is not None
         archive = Path(package.source_url)
+        if package.source_subdirectory:
+            archive = archive / package.source_subdirectory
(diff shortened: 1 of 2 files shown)


radoering (changes requested):
LGTM.

In addition, I think we can remove https://github.com/python-poetry/poetry/blob/3e32a4f83e89b2b2bede68e0d375a70931b04a8e/src/poetry/installation/executor.py#L589-L596

because this part was supposed to consider the subdirectory but apparently it does nothing.

radoering:
Actually, if we remove this code, we don't need `_prepare_directory_archive` at all, but can call `_prepare_archive` directly. 🤔

harrymander:
Yep those lines do nothing... have removed

radoering (approved):

PR #7594, opened 2023-03-02
Fix certain hash miss matches in case of non weel installations

quantum-byte:
## Addressed Issue:

1. Have a a package which provides both wheel and non wheel archives (like a `tar.gz`)
2. Force the use of the non wheel archive during installation:
```bash
export POETRY_INSTALLER_NO_BINARY=dagster-docker; poetry install dagster-docker
```
3. Get error like this:
```bash
 RuntimeError

  Hash for dagster-docker (0.17.20) from archive dagster_docker-0.17.20-py3-none-any.whl not found in known hashes (was: sha256:f0369c37d4328bd140c727494f65b88c49df0999c1981912bc68274e8412d793)
```

~~This is due to the original wheel archive from the registry being overwritten in the poetry artifact cache by the newly created wheel archive from the sdist `tar.gz` archive. The new wheel archive though has minimal changes (bdist version etc.) and causes the sha265 hash miss match.~~ (Overwrite does not happen but the wrong hash is checked. See [comment](https://github.com/python-poetry/poetry/pull/7594#issuecomment-1454792582) )

## Changes to fix the issue:

- ~~Adjusted Chef.prepare() to always write to a temporary directory. Otherwise we would overwrite an already existing cached wheel. This could then make the installation fail due to miss matched hashes.~~
- ~~Changed Executor._download_link() to validate package hashes via _populate_hashes_dict before building a new wheel in case we use a non wheel archive.~~

~~This eliminates the caching of the built wheel for non wheel archives but i did not see an easy way to keep this caching and fix the issue.  Do you think this will impact performance in a meaningfull way ?~~

## Pull Request Check List

Resolves: Unclear - I have not created a specific issue for my problem but it could address some issues related to hash miss matches.

### DIFF
--- a/src/poetry/installation/chef.py
+++ b/src/poetry/installation/chef.py
@@ -154,8 +154,6 @@ class Chef:
             return path
 
     def _prepare_sdist(self, archive: Path, destination: Path) -> Path:
-        from poetry.core.packages.utils.link import Link
-
         suffix = archive.suffix
         context: Callable[
             [str], AbstractContextManager[zipfile.ZipFile | tarfile.TarFile]
--- a/src/poetry/installation/chef.py
+++ b/src/poetry/installation/chef.py
@@ -86,8 +86,6 @@ class IsolatedEnv(BaseIsolatedEnv):
 
 
 class Chef:
-    tmp_dir_prefix = "poetry-chef-"
-
     def __init__(self, config: Config, env: Env, pool: RepositoryPool) -> None:
         self._env = env
         self._pool = pool
@@ -102,7 +100,7 @@ class Chef:
             return archive
 
         if archive.is_dir():
-            tmp_dir = tempfile.mkdtemp(prefix=self.tmp_dir_prefix)
+            tmp_dir = tempfile.mkdtemp(prefix="poetry-chef-")
             return self._prepare(archive, Path(tmp_dir), editable=editable)
 
         return self._prepare_sdist(archive, destination=output_dir)
--- a/src/poetry/installation/chef.py
+++ b/src/poetry/installation/chef.py
@@ -86,6 +86,8 @@ class IsolatedEnv(BaseIsolatedEnv):
 
 
 class Chef:
+    tmp_dir_prefix = "poetry-chef-"
+
     def __init__(self, config: Config, env: Env, pool: RepositoryPool) -> None:
         self._env = env
         self._pool = pool
@@ -93,18 +95,15 @@ class Chef:
             Path(config.get("cache-dir")).expanduser().joinpath("artifacts")
         )
 
-    def prepare(
-        self, archive: Path, output_dir: Path | None = None, *, editable: bool = False
-    ) -> Path:
+    def prepare(self, archive: Path, *, editable: bool = False) -> Path:
         if not self._should_prepare(archive):
             return archive
 
+        tmp_dir = tempfile.mkdtemp(prefix=self.tmp_dir_prefix)
         if archive.is_dir():
-            tmp_dir = tempfile.mkdtemp(prefix="poetry-chef-")
-
             return self._prepare(archive, Path(tmp_dir), editable=editable)
 
-        return self._prepare_sdist(archive, destination=output_dir)
+        return self._prepare_sdist(archive, destination=Path(tmp_dir))
 
     def _prepare(
         self, directory: Path, destination: Path, *, editable: bool = False
@@ -154,7 +153,7 @@ class Chef:
 
             return path
 
-    def _prepare_sdist(self, archive: Path, destination: Path | None = None) -> Path:
+    def _prepare_sdist(self, archive: Path, destination: Path) -> Path:
         from poetry.core.packages.utils.link import Link
 
         suffix = archive.suffix
@@ -180,9 +179,6 @@ class Chef:
                 sdist_dir = archive_dir / archive.name.rstrip(suffix)
                 if not sdist_dir.is_dir():
                     sdist_dir = archive_dir
-
-            if destination is None:
-                destination = self.get_cache_directory_for_link(Link(archive.as_uri()))
 
             destination.mkdir(parents=True, exist_ok=True)
 
--- a/src/poetry/installation/chef.py
+++ b/src/poetry/installation/chef.py
@@ -199,7 +199,7 @@ class Chef:
     def _is_wheel(cls, archive: Path) -> bool:
         return archive.suffix == ".whl"
 
-    def get_cached_archive_for_link(self, link: Link, strict: bool) -> Path | None:
+    def get_cached_archive_for_link(self, link: Link, *, strict: bool) -> Path | None:
         archives = self.get_cached_archives_for_link(link)
         if not archives:
             return None
--- a/src/poetry/installation/chef.py
+++ b/src/poetry/installation/chef.py
@@ -95,15 +95,17 @@ class Chef:
             Path(config.get("cache-dir")).expanduser().joinpath("artifacts")
         )
 
-    def prepare(self, archive: Path, *, editable: bool = False) -> Path:
+    def prepare(
+        self, archive: Path, output_dir: Path | None = None, *, editable: bool = False
+    ) -> Path:
         if not self._should_prepare(archive):
             return archive
 
-        tmp_dir = tempfile.mkdtemp(prefix=self.tmp_dir_prefix)
         if archive.is_dir():
+            tmp_dir = tempfile.mkdtemp(prefix=self.tmp_dir_prefix)
             return self._prepare(archive, Path(tmp_dir), editable=editable)
 
-        return self._prepare_sdist(archive, destination=Path(tmp_dir))
+        return self._prepare_sdist(archive, destination=output_dir)
 
     def _prepare(
         self, directory: Path, destination: Path, *, editable: bool = False
@@ -153,7 +155,9 @@ class Chef:
 
             return path
 
-    def _prepare_sdist(self, archive: Path, destination: Path) -> Path:
+    def _prepare_sdist(self, archive: Path, destination: Path | None = None) -> Path:
+        from poetry.core.packages.utils.link import Link
+
         suffix = archive.suffix
         context: Callable[
             [str], AbstractContextManager[zipfile.ZipFile | tarfile.TarFile]
@@ -177,6 +181,9 @@ class Chef:
                 sdist_dir = archive_dir / archive.name.rstrip(suffix)
                 if not sdist_dir.is_dir():
                     sdist_dir = archive_dir
+
+            if destination is None:
+                destination = self.get_cache_directory_for_link(Link(archive.as_uri()))
 
             destination.mkdir(parents=True, exist_ok=True)
 
--- a/src/poetry/installation/chef.py
+++ b/src/poetry/installation/chef.py
@@ -209,8 +209,7 @@ class Chef:
                 # prioritized archive type.
                 if link.filename == archive.name:
                     return archive
-                else:
-                    continue
+                continue
             if archive.suffix != ".whl":
                 candidates.append((float("inf"), archive))
                 continue
--- a/src/poetry/installation/chef.py
+++ b/src/poetry/installation/chef.py
@@ -199,13 +199,20 @@ class Chef:
     def _is_wheel(cls, archive: Path) -> bool:
         return archive.suffix == ".whl"
 
-    def get_cached_archive_for_link(self, link: Link) -> Path | None:
+    def get_cached_archive_for_link(self, link: Link, strict: bool) -> Path | None:
         archives = self.get_cached_archives_for_link(link)
         if not archives:
             return None
 
         candidates: list[tuple[float | None, Path]] = []
         for archive in archives:
+            if strict:
+                # in strict mode return the original cached archive instead of the
+                # prioritized archive type.
+                if link.filename == archive.name:
+                    return archive
+                else:
+                    continue
             if archive.suffix != ".whl":
                 candidates.append((float("inf"), archive))
                 continue
--- a/src/poetry/installation/executor.py
+++ b/src/poetry/installation/executor.py
@@ -697,6 +697,7 @@ class Executor:
     def _download_link(self, operation: Install | Update, link: Link) -> Path:
         package = operation.package
 
+        output_dir = self._chef.get_cache_directory_for_link(link)
         archive = self._chef.get_cached_archive_for_link(link)
         if archive is None:
             # No cached distributions was found, so we download and prepare it
@@ -712,7 +713,10 @@ class Executor:
 
                 raise
 
-        self._populate_hashes_dict(archive, package)
+        # validate the original (downloaded or cached) archive only if we can. e.g. if
+        # we use the cached wheel or the actual downloaded archive
+        if package.files and link.filename == archive.name:
+            self._validate_archive_hash(archive, package)
 
         if archive.suffix != ".whl":
             message = (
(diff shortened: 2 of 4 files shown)


radoering:
> This eliminates the caching of the built wheel for non wheel archives but i did not see an easy way to keep this caching and fix the issue. Do you think this will impact performance in a meaningfull way ?

It will have a significant impact.

radoering (changes requested):
After thinking about the issue and trying a few things, I suspect the solution is much simpler: Don't move the `_populate_hashes_dict()`, but make it conditional:

```
if archive.suffix != ".whl":
    ...

elif link.is_wheel:
    self._populate_hashes_dict(archive, package)
```

We only want to populate the dict with (and check the hash of) the wheel, if it was downloaded from somewhere and not built from an sdist by poetry.

radoering:
> This is due to the original wheel archive from the registry being overwritten in the poetry artifact cache by the newly created wheel archive from the sdist `tar.gz` archive.

By the way, that's not true. Example:

Cache after installing tomli from wheel:

```
$ find ~/.cache/pypoetry/artifacts -name tomli*
/home/randy/.cache/pypoetry/artifacts/62/12/b6/6db9ebb9c8e1a6c5aa8a92ae73098d8119816b5e8507490916621bc305/tomli-2.0.1-py3-none-any.whl
```

Cache after installing tomli from sdist afterwards:

```
$ find ~/.cache/pypoetry/artifacts -name tomli*
/home/randy/.cache/pypoetry/artifacts/aa/63/d4/79512b2ef845b845ecbc074b5689753b3fc1b4ec2048cfff310fb36ea0/tomli-2.0.1-py3-none-any.whl
/home/randy/.cache/pypoetry/artifacts/aa/63/d4/79512b2ef845b845ecbc074b5689753b3fc1b4ec2048cfff310fb36ea0/tomli-2.0.1.tar.gz
/home/randy/.cache/pypoetry/artifacts/62/12/b6/6db9ebb9c8e1a6c5aa8a92ae73098d8119816b5e8507490916621bc305/tomli-2.0.1-py3-none-any.whl
```

Artifact cache folders are per link. A downloaded sdist and a downloaded wheel have different links and thus are stored in different folders. The self-built wheel is put in the same folder as the sdist.

quantum-byte:
> > This is due to the original wheel archive from the registry being overwritten in the poetry artifact cache by the newly created wheel archive from the sdist `tar.gz` archive.
> 
> By the way, that's not true. Example:

Wow my bad. I checked the file hashes for the filename but i missed comparing the path that it was actually the same file that i was looking at. I even saw the url logic for the cache path generation :sweat_smile:. So much for jumping to conclusions.

This makes your solution work and we get the caching for generated wheels back :partying_face:.


Though i am not sure if its a good idea to skip validation for everything besides wheels. Should poetry not validate the downloaded and (cached) tar.gz file if there is a hash from the repo available ? Or is that done somewhere else that i did not see.

I provided a solution which still includes validation of non_wheel packages if possible. But then i am not sure what hash should be recorded in this case (the original e.g. [tar.gz hash](https://github.com/python-poetry/poetry/pull/7594/commits/de71a165665a25fe14b1569ab252d05aef0179b5), [the generated whl hash](https://github.com/python-poetry/poetry/pull/7594/commits/d8e45fbbe808628633c9e37e5624a6400a2a7905) or none) ?

radoering:
> Though i am not sure if its a good idea to skip validation for everything besides wheels. Should poetry not validate the downloaded and (cached) tar.gz file if there is a hash from the repo available ? Or is that done somewhere else that i did not see.
> 
> I provided a solution which still includes validation of non_wheel archives if possible. But then i am not sure what hash should be recorded in this case (the original e.g. [tar.gz hash](https://github.com/python-poetry/poetry/pull/7594/commits/de71a165665a25fe14b1569ab252d05aef0179b5), [the generated whl hash](https://github.com/python-poetry/poetry/pull/7594/commits/d8e45fbbe808628633c9e37e5624a6400a2a7905) or none) ?

Good point. I've noticed this too, but would have put it off until later if you hadn't brought it up. I think you are on the right track so we can also tackle it now. 😄  It does not make sense to record the hash of the generated wheel because this hash is not known by anybody else and the wheel is only intermediate. We always have to record the original url and hash (tar.gz in this case).

If I see it correctly, there is still one flaw: The hash of the tar.gz is only recorded if there is not yet a cached wheel. In order to fix this, I assume we need two versions of `get_cached_archive_for_link()`: one that prefers cached wheels and one that only returns the original archive. So maybe, let's introduce a parameter `strict` where `strict=False` is the current implementation and `strict=True` returns the original archive or None if it is not cached.

Then, we can use `strict=True` in https://github.com/python-poetry/poetry/blob/f27308face9b09da8b5bbd59175b1b46392eb8d5/src/poetry/installation/executor.py#L696

afterwards always call `_populate_hashes_dict()` (because we always have the original distribution) and at last call `get_cached_archive_for_link()` with `strict=False` to get a cached wheel if there exists one.

tests/installation/test_executor.py
```diff
@@ -739,6 +776,42 @@ def test_executor_should_write_pep610_url_references_for_urls(
     verify_installed_distribution(tmp_venv, package, expected_url_reference)
 
 
+def test_executor_should_not_write_pep610_url_references_for_non_wheel_urls(
+    tmp_venv: VirtualEnv,
+    pool: RepositoryPool,
+    config: Config,
+    io: BufferedIO,
+    mock_file_downloads: None,
+):
```
radoering, line 785: You can verify that the hash is not written if a wheel is cached by extending your test case as follows:

(The mock has to be adapted after implementing the `strict` behavior.)

```suggestion
@pytest.mark.parametrize("cached", [False, True])
def test_executor_should_write_pep610_url_references_for_non_wheel_urls(
    tmp_venv: VirtualEnv,
    pool: RepositoryPool,
    config: Config,
    io: BufferedIO,
    mock_file_downloads: None,
    mocker: MockerFixture,
    fixture_dir: FixtureDirGetter,
    cached: bool,
):
    if cached:
        link_cached = fixture_dir("distributions") / "demo-0.1.0-py2.py3-none-any.whl"
        mocker.patch(
            "poetry.installation.chef.Chef.get_cached_archive_for_link",
            return_value=link_cached,
        )
```

quantum-byte:
> > Though i am not sure if its a good idea to skip validation for everything besides wheels. Should poetry not validate the downloaded and (cached) tar.gz file if there is a hash from the repo available ? Or is that done somewhere else that i did not see.
> > I provided a solution which still includes validation of non_wheel archives if possible. But then i am not sure what hash should be recorded in this case (the original e.g. [tar.gz hash](https://github.com/python-poetry/poetry/pull/7594/commits/de71a165665a25fe14b1569ab252d05aef0179b5), [the generated whl hash](https://github.com/python-poetry/poetry/pull/7594/commits/d8e45fbbe808628633c9e37e5624a6400a2a7905) or none) ?
> 
> Good point. I've noticed this too, but would have put it off until later if you hadn't brought it up. I think you are on the right track so we can also tackle it now. smile It does not make sense to record the hash of the generated wheel because this hash is not known by anybody else and the wheel is only intermediate. We always have to record the original url and hash (tar.gz in this case).
> 
> If I see it correctly, there is still one flaw: The hash of the tar.gz is only recorded if there is not yet a cached wheel. In order to fix this, I assume we need two versions of `get_cached_archive_for_link()`: one that prefers cached wheels and one that only returns the original archive. So maybe, let's introduce a parameter `strict` where `strict=False` is the current implementation and `strict=True` returns the original archive or None if it is not cached.

Ok that looks like a good idea. I dont have the time today but will have a look tomorrow to get a implementation setup up and improve the tests to cover it.

tests/installation/test_executor.py
quantum-byte: Done

quantum-byte:
@radoering I did find some time already and the implementation including tests is complete as far as i can see. Let me know if there is still something missing and how to proceed from here :).

tests/installation/test_executor.py
```diff
@@ -739,6 +787,60 @@ def test_executor_should_write_pep610_url_references_for_urls(
...
+    if cached:
+        link_cached = fixture_dir("distributions") / "demo-0.1.0.tar.gz"
+        original_func = Chef.get_cached_archive_for_link
+
+        def mock_get_cached_archive_for_link(self: Self, link: Link, strict: bool):
+            if link.filename == "demo-0.1.0.tar.gz":
+                return link_cached
+            else:
+                return original_func(self, link, strict)
```
radoering, line 809: That doesn't work as expected. The mock is always called with the tar.gz. It should be something like:

```suggestion
            if strict:
                return cached_sdist
            else:
                return cached_wheel
```

tests/installation/test_executor.py
```diff
@@ -739,6 +787,60 @@ def test_executor_should_write_pep610_url_references_for_urls(
...
+@pytest.mark.parametrize("cached", [False, True])
+def test_executor_should_write_pep610_url_references_for_non_wheel_urls(
+    tmp_venv: VirtualEnv,
+    pool: RepositoryPool,
+    config: Config,
+    io: BufferedIO,
+    mock_file_downloads: None,
+    mocker: MockerFixture,
+    fixture_dir: FixtureDirGetter,
+    cached: bool,
```
radoering, line 799:
```suggestion
    cached_wheel: bool,
```

Maybe, we should name it more clear because it doesn't matter if the sdist is cached, only if a generated wheel is cached or not.

radoering (changes requested):

src/poetry/installation/executor.py
```diff
@@ -729,9 +743,9 @@ def _populate_hashes_dict(self, archive: Path, package: Package) -> None:
     @staticmethod
     def _validate_archive_hash(archive: Path, package: Package) -> str:
         archive_hash: str = "sha256:" + get_file_hash(archive)
-        known_hashes = {f["hash"] for f in package.files}
+        known_hashes = {f["hash"] for f in package.files if f["file"] == archive.name}
 
-        if archive_hash not in known_hashes:
+        if known_hashes and archive_hash not in known_hashes:
```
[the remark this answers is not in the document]
quantum-byte: I am not sure about this. Is there a standard in pypy that a package downloaded from a url needs to provide a hash ? 

In case there is no hash we would run into an empty set here. On second though, it would already fail in line 746 accessing `f["hash"]`

I can remove it, if we can assume that every package.files entry has a hash

src/poetry/installation/chef.py
```diff
@@ -198,13 +199,20 @@ def _should_prepare(self, archive: Path) -> bool:
     def _is_wheel(cls, archive: Path) -> bool:
         return archive.suffix == ".whl"
 
-    def get_cached_archive_for_link(self, link: Link) -> Path | None:
+    def get_cached_archive_for_link(self, link: Link, strict: bool) -> Path | None:
```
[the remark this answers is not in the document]
quantum-byte: Oh wow, did not know about this python feature. Thanks :smile:

tests/installation/test_executor.py
quantum-byte: How about `is_artifact_cached` since its not only about wheels.

tests/installation/test_executor.py
quantum-byte: You are right. I was about to write a reasoning why it would not but then it dawned on me.

Originally i thought since its a full sdist install, it needs further dependencies like 'poetry-core' during the `chef._prepare` call. This causes additional calls to `mock_get_cached_archive_for_link` for those deps which need to be handled.

I can also see two other ways work:

1. Mocking away all other required deps here with specific filename matches (in my view brittle and hard to maintain)
2. Also mocking away the `chef._prepare` the call. Then it would similiarily become less integration test and more proper unit test, but the `coverage` of the testcase would go down. (Also did not think of that one yesterday).

But if you are fine with mocking that specific case and avoiding the actual build process of the wheel, its good enough for me.

src/poetry/installation/executor.py
```diff
@@ -514,6 +515,10 @@ def _install(self, operation: Install | Update) -> int:
         else:
             archive = self._download(operation)
 
+        tmp_path_prefix = str(Path(tempfile.gettempdir(), self._chef.tmp_dir_prefix))
+        if str(archive).startswith(tmp_path_prefix):
+            cleanup_archive = True
+
```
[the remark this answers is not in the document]
quantum-byte: You are right. Its no longer needed. While checking i also fixed the tests that where still leaving some generated files laying around.

Command to check / find them (on linux):
```bash
find /tmp/ -type f -iwholename *poetry-chef-*
```

src/poetry/installation/executor.py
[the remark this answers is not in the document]
radoering: We assumed it before this PR. If there the set of known hashes is empty, the hash is not known and we should fail for security reasons. I wouldn't weaken that constraint if it is not necessary.

tests/installation/test_executor.py
radoering: Mocking `chef._prepare` should be fine since preparing sdists is tested in `test_chef.py`. If we mock it, we can add an additional check that `chef._prepare` is only called if artifacts are not cached and is not called if there is a cached wheel.

tests/installation/test_executor.py
radoering: Ah, I thought caching is only turned on/off for the generated wheel and the sdist is always cached, but it is also turned on/off for the sdist. In that case `cached` would have been fine, but `is_artifact_cached` is better. 😄

src/poetry/installation/executor.py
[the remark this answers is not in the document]
quantum-byte: Done

tests/installation/test_executor.py
```diff
@@ -846,38 +1007,6 @@ def test_executor_should_write_pep610_url_references_for_git_with_subdirectories
     )
 
 
-def test_executor_should_use_cached_link_and_hash(
```
quantum-byte, line 849: This test is replaced by the more complete `test_executor_should_write_pep610_url_references_for_non_wheel_urls`

tests/installation/test_executor.py
quantum-byte: I improved the tests and made them more comprehensive. I hope its still readable / understandable.

src/poetry/installation/chef.py
```diff
@@ -198,13 +197,20 @@ def _should_prepare(self, archive: Path) -> bool:
...
             return None
 
         candidates: list[tuple[float | None, Path]] = []
         for archive in archives:
+            if strict:
+                # in strict mode return the original cached archive instead of the
+                # prioritized archive type.
+                if link.filename == archive.name:
+                    return archive
+                else:
```
dimbleby, line 212: redundant `else`

tests/installation/test_executor.py
```diff
@@ -737,6 +793,111 @@ def test_executor_should_write_pep610_url_references_for_urls(
...
+)
+def test_executor_should_write_pep610_url_references_for_non_wheel_urls(
+    tmp_venv: VirtualEnv,
+    pool: RepositoryPool,
+    config: Config,
+    io: BufferedIO,
+    mock_file_downloads: None,
+    mocker: MockerFixture,
+    fixture_dir: FixtureDirGetter,
+    is_sdist_artifact_cached: bool,
```
radoering, line 826:
```suggestion
    is_sdist_cached: bool,
```

consistent to `is_wheel_cached`

tests/installation/test_executor.py
```diff
@@ -737,6 +793,111 @@ def test_executor_should_write_pep610_url_references_for_urls(
...
+                    return cached_sdist
+                else:
+                    return None
+            else:
+                if is_wheel_cached:
+                    return cached_wheel
+                elif is_sdist_artifact_cached:
+                    return cached_sdist
+                else:
+                    return None
```
radoering, line 854:
```suggestion
            if is_sdist_cached:
                return cached_sdist
            if is_wheel_cached and not strict:
                return cached_wheel
            return None
```

radoering (changes requested):

src/poetry/installation/chef.py
quantum-byte: Fixed

tests/installation/test_executor.py
quantum-byte: Fixed

tests/installation/test_chef.py
```diff
@@ -38,20 +41,80 @@ def setup(mocker: MockerFixture, pool: RepositoryPool) -> None:
...
+        Factory.create_pool(config),
+    )
+
+    mocker.patch.object(
+        chef, "get_cached_archives_for_link", return_value=available_packages
+    )
+
+    archive = chef.get_cached_archive_for_link(Link(link), strict=strict)
+
+    assert None is archive
```
radoering, line 88:
```suggestion
    assert archive is None
```

radoering (changes requested):

radoering (approved):

JoseHernandezTR:
Hello, would anything related to this fix cause this side effect?

_WheelFileValidationError
67	
68	  ["In /root/.cache/pypoetry/artifacts/23/11/79/b24ac52cf28d681dbe7414498ba63b1521494ec3907b422567db0dd630/pydata_sphinx_theme-0.13.1-py3-none-any.whl, hash / size of pydata_sphinx_theme/__init__.py didn't match RECORD", "In /root/.cache/pypoetry/artifacts/23/11/79/b24ac52cf28d681dbe7414498ba63b1521494ec3907b422567db0dd630/pydata_sphinx_theme-0.13.1-py3-none-any.whl, hash / size of pydata_sphinx_theme/assets/scripts/bootstrap.js didn't match RECORD"
...

This is confirmed to happen with 1.4.1 but not with 1.4.0. Thanks!

heimalne:
Probably, here's the issue https://github.com/python-poetry/poetry/issues/7691 and the relevant MR to fix it https://github.com/python-poetry/poetry/pull/7694

PR #7595, opened 2023-03-03
feat: cache url dependencies during dependency resolution

ralbertazzi:
# Pull Request Check List

Resolves: #2415

This PR exploits the already existing `Authenticator` object with file-cache functionalities to create a dedicated `url` cache for URL dependencies. Such cache is used at dependency resolution time.

Note how the solution isn't still perfect, as there's an extra file transfer performed by Poetry from the cached file to a temporary directory. Still, by setting an appropriate `chunk_size` the performance improvement is extremely visible even with huge dependencies (dear PyTorch) where the "download' goes from minutes to a few seconds.

Also, there's quite a `disable_cache` forwarding around the codebase to make sure that the Authenticator's cache is disabled upon user request.

- [ ] Added **tests** for changed code. -> Might need support for it. Also there's currently no tests for URL dependencies in `test_provider.py`, while there are tests for most other kinds (git, file, ..)
- ~[ ] Updated **documentation** for changed code.~ Not needed IMO

### DIFF
--- a/src/poetry/puzzle/provider.py
+++ b/src/poetry/puzzle/provider.py
@@ -33,6 +33,7 @@ from poetry.packages import DependencyPackage
 from poetry.packages.package_collection import PackageCollection
 from poetry.puzzle.exceptions import OverrideNeeded
 from poetry.repositories.exceptions import PackageNotFound
+from poetry.utils.authenticator import Authenticator
 from poetry.utils.helpers import download_file
 from poetry.utils.helpers import get_file_hash
 from poetry.vcs.git import Git
@@ -54,6 +55,7 @@ if TYPE_CHECKING:
     from poetry.core.packages.url_dependency import URLDependency
     from poetry.core.packages.vcs_dependency import VCSDependency
     from poetry.core.version.markers import BaseMarker
+    from requests import Session
 
     from poetry.repositories import RepositoryPool
     from poetry.utils.env import Env
@@ -144,6 +146,7 @@ class Provider:
         *,
         installed: list[Package] | None = None,
         locked: list[Package] | None = None,
+        disable_cache: bool = False,
     ) -> None:
         self._package = package
         self._pool = pool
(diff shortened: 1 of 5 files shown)


src/poetry/puzzle/provider.py
```diff
@@ -458,7 +460,7 @@ def get_package_from_url(cls, url: str) -> Package:
         file_name = os.path.basename(urllib.parse.urlparse(url).path)
         with tempfile.TemporaryDirectory() as temp_dir:
             dest = Path(temp_dir) / file_name
-            download_file(url, dest)
+            download_file(url, dest, session=self._url_authenticator, chunk_size=1024 * 1024)
```
ralbertazzi, line 463: The increased `chunk_size` is needed to ensure a fast read from the file cache when there's a cache hit. The default `chunk_size = 1024` would make the read pretty slow

src/poetry/puzzle/provider.py
```diff
@@ -159,6 +160,7 @@ def __init__(
         self._direct_origin_packages: dict[str, Package] = {}
         self._locked: dict[NormalizedName, list[DependencyPackage]] = defaultdict(list)
         self._use_latest: Collection[NormalizedName] = []
+        self._url_authenticator = Authenticator(cache_id="url")
```
ralbertazzi, line 163: This assumes that there should be no repository called "url". If we think this could cause potential problems we can also rename it to `_url`

radoering:
As already mentioned by dimbleby in the related issue and the other (closed) PR, I also think that using the artifacts cache is the way to go. That avoids copying to a temporary directory and not only reuses the cache during dependency resolution but also for installation.

If anybody is motivated enough, I think the steps should be as follows:

1. make [`get_cached_archives_for_link()`](https://github.com/python-poetry/poetry/blob/f27308face9b09da8b5bbd59175b1b46392eb8d5/src/poetry/installation/chef.py#L240) independent from `interpreter_name` and `interpreter_version`
2. add an `artifacts_cache_directory` property to `config.py` analoguous to [`repository_cache_directory()`](https://github.com/python-poetry/poetry/blob/f27308face9b09da8b5bbd59175b1b46392eb8d5/src/poetry/config/config.py#L212)
3. move `get_cached_archives_for_link()` and `get_cache_directory_for_link()` from `chef.py` to `utils/cache.py`
4. adapt `get_package_from_url()` in `provider.py` to use the artifacts cache

If you want to split it up, steps 1-3 can be done in a first PR (refactoring only) and step 4 in a follow up PR (enhancement).

ralbertazzi:
Hi @radoering , I tried to address your first 3 steps in this PR ;) https://github.com/python-poetry/poetry/pull/7621

ralbertazzi:
Closed in favour of https://github.com/python-poetry/poetry/pull/7621

Issue #6198, opened 2022-08-19
Regression in namespace package metadata on poetry-core 1.1.0rc1


ariebovenberg:
- [x] ~~I am on the [latest](https://github.com/python-poetry/poetry/releases/latest) Poetry version.~~ n/a
- [x] I have searched the [issues](https://github.com/python-poetry/poetry/issues) of this repo and believe that this is not a duplicate.
- [x] ~~If an exception occurs when executing a command, I executed it again in debug mode (`-vvv` option).~~ n/a

- **OS version and name**: MacOS 12.5.1
- **Poetry version**: poetry-core 1.1.0rc1
- **Link of a [Gist](https://gist.github.com/) with the contents of your pyproject.toml file**: see below

## Issue

An addition in `poetry-core` between 1.1.0a7 and 1.1.0b1 introduced a bug in package metadata name for namespaced packages:

```python
>>> from importlib import metadata
>>> metadata('mynamespace.foomodule')
importlib.metadata.PackageNotFoundError: mynamespace.foomodule
```

The package is instead registered as `mynamespace-foomodule`. This is a particular problem since `importlib.metadata` is the canonical way to single-source the `__version__`

This bug occurs on Python 3.8 and 3.9 (but not 3.10 insterestingly enough). The change that introduced the bug: https://github.com/python-poetry/poetry-core/pull/328/. Ever since this change, `.` is normalized to `-` in package names, which is incorrect for namespace packages. 

### `pyproject.toml`
```toml
[tool.poetry]
name = "mynamespace.foomodule"
version = "0.1.0"
description = ""
authors = ["Your Name <you@example.com>"]
readme = "README.md"
packages = [
    { include = "mynamespace"},
]

[tool.poetry.dependencies]
python = "^3.8"

[build-system]
requires = ["poetry-core==1.1.0rc1"]
build-backend = "poetry.core.masonry.api"
```

### files

```
mynamespace/
├── __init__.py  # contains: __path__ = __import__('pkgutil').extend_path(__path__, __name__)
└── foomodule
    └── __init__.py  # empty
```

### Then, run:

```bash
# make sure you're on python 3.8 or 3.9
pip install .
# this raises an error, but shouldn't
python -c "from importlib import metadata; metadata.version('mynamespace.foomodule')"
```

## What to do?

I don't mind submitting a fix in `poetry-core`, but I'm of course missing a lot of context on why the change was made in the first place.

dimbleby:
[...]

PR #7607, opened 2023-03-05
Clarify dots canonicalisation change is breaking

PeterJCLaw:
The handling of package names within the Python packaging ecosystem is unfortunately not in a great state at the moment. There exist a large number of packages which contain characters which (some of) the specs indicate are invalid, yet there is not a clear migration path nor destination at the moment.

Poetry's approach is to normalise project names towards their "canonical" form, meaning that existing projects on PyPI which use the more relaxed name forms will be silently renamed if they move to (recent versions of) Poetry. As well as changing the name of these packages (which maintainers are unlikely to expect nor desire) this unfortunately ends up breaking introspection via `importlib.metadata` in Python 3.8 and 3.9 which are not aware of the canonicalisation rules which Poetry is using.

Adding this compatibility note informs maintainers so that they can decide how (and if) they want this name canonicalisation to happen.

See also https://github.com/python-poetry/poetry/issues/6198, https://github.com/python-poetry/poetry-core/pull/560.


dimbleby:
per https://github.com/python-poetry/poetry-core/pull/560#issuecomment-1662689185: https://github.com/pypi/warehouse/issues/10030 was closed - fixed apparently - a couple of weeks ago, which should make this unwanted.

radoering:
I'm closing it since 1.2 was released quite a while ago and I don't think it's worth to still update the changelog. Especially, if we can fix / change back now.

(Accidentally clicked on the update button.)

PR #7608, opened 2023-03-05
docs:mention enabled venv in virtualenvs.create configuration

kamilswiec:
## Issue
[Docs](https://python-poetry.org/docs/configuration/#virtualenvscreate) do not mention that if you have `VIRTUAL_ENV` set with path to venv created inside (if path does not have venv direcotries there poetry simply ignores that variable), poetry will install packages into it. Documentation only mentions that it will detect venvs found in `{cache-dir}/virtualenvs` or `{project-dir}/.venv` and then fallback to systems python environment . Change that detects venv by variable is mentioned in [0.12.4](https://github.com/python-poetry/poetry/releases/tag/0.12.4) - 2018-10-21. 

Reason to mention that is to be aware that it is possible to install all packages in custom directory using environment variable. I use it myself to install dependencies in one stage of docker build and then copy them to another stage. 

### Pull Request Check List

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -280,8 +280,8 @@ Use parallel execution when using the new (`>=1.1.0`) installer.
 Create a new virtual environment if one doesn't already exist.
 
 If set to `false`, Poetry will not create a new virtual environment. If it detects a virtual environment
-in `{cache-dir}/virtualenvs` or `{project-dir}/.venv` it will install dependencies into them, otherwise it will install
-dependencies into the systems python environment.
+in `{cache-dir}/virtualenvs` or `{project-dir}/.venv` or in `VIRTUAL_ENV` it will install dependencies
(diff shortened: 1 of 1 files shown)


neersighted:
I don't think `VIRTUAL_ENV` deserves special acknowledgement in Poetry docs -- it's better to state that Poetry detects an already virtual environment. You can activate a virtual environment by setting `VIRTUAL_ENV`, but that is an implementation detail of `venv` and `virtualenv`.

kamilswiec:
> I don't think `VIRTUAL_ENV` deserves special acknowledgement in Poetry docs -- it's better to state that Poetry detects an already virtual environment. You can activate a virtual environment by setting `VIRTUAL_ENV`, but that is an implementation detail of `venv` and `virtualenv`.

Sounds resonable, I changed description to only mention it detects enabled venv not only certain paths.

neersighted (dismissed):

PR #7621, opened 2023-03-07
refactor: extract cache utilities

ralbertazzi:
# Pull Request Check List

As discuss with @radoering in https://github.com/python-poetry/poetry/pull/7595, this is a first refactor PR that tries to extract cache utilities outside of Chef. The end goal is to resolve finally https://github.com/python-poetry/poetry/issues/2415.

- Added an `artifacts_cache_directory` property to config.py analoguous to `repository_cache_directory`
- Made `get_cached_archives_for_link()` independent from interpreter_name and interpreter_version
- Moved `get_cached_archives_for_link` and `get_cache_directory_for_link` from `chef.py` to `utils/cache.py`

I ended up moving the recently touched `get_cached_archive_for_link` too, although it looks to me that it's not such a generic utility, but it's mostly related to the installation step. The fact that I had to refactor a bit the `installer` package to avoid circular imports makes this evident. Still, I believed `chef.py` wasn't the best place to keep that function since the other ones have been moved. I'm happy to receive feedback on this and other possible solutions. Most likely I would put  `get_cached_archive_for_link` directly inside `Executor`( which is the only user of that function), but I wanted to get your feedback first.

### DIFF
--- a/src/poetry/config/config.py
+++ b/src/poetry/config/config.py
@@ -213,6 +213,10 @@ class Config:
         return Path(self.get("cache-dir")) / "cache" / "repositories"
 
     @property
+    def artifacts_cache_directory(self) -> Path:
+        return Path(self.get("cache-dir")) / "artifacts"
+
+    @property
     def virtualenvs_path(self) -> Path:
         path = self.get("virtualenvs.path")
         if path is None:
--- a/src/poetry/config/config.py
+++ b/src/poetry/config/config.py
@@ -210,11 +210,11 @@ class Config:
 
     @property
     def repository_cache_directory(self) -> Path:
-        return Path(self.get("cache-dir")) / "cache" / "repositories"
+        return Path(self.get("cache-dir")).expanduser() / "cache" / "repositories"
 
     @property
     def artifacts_cache_directory(self) -> Path:
-        return Path(self.get("cache-dir")) / "artifacts"
+        return Path(self.get("cache-dir")).expanduser() / "artifacts"
 
     @property
     def virtualenvs_path(self) -> Path:
--- a/src/poetry/installation/executor.py
+++ b/src/poetry/installation/executor.py
@@ -20,6 +20,8 @@ from poetry.core.packages.utils.link import Link
 from poetry.installation.chef import Chef
 from poetry.installation.chef import ChefBuildError
 from poetry.installation.chooser import Chooser
+from poetry.installation.chooser import InvalidWheelName
+from poetry.installation.chooser import Wheel
 from poetry.installation.operations import Install
 from poetry.installation.operations import Uninstall
 from poetry.installation.operations import Update
@@ -27,6 +29,8 @@ from poetry.installation.wheel_installer import WheelInstaller
 from poetry.puzzle.exceptions import SolverProblemError
 from poetry.utils._compat import decode
 from poetry.utils.authenticator import Authenticator
+from poetry.utils.cache import get_cache_directory_for_link
+from poetry.utils.cache import get_cached_archives_for_link
 from poetry.utils.env import EnvCommandError
 from poetry.utils.helpers import atomic_open
 from poetry.utils.helpers import get_file_hash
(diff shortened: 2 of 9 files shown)


ralbertazzi:
I went ahead at moving `get_cached_archive_for_link` inside `Executor`

src/poetry/config/config.py
```diff
@@ -212,6 +212,10 @@ def _get_environment_repositories() -> dict[str, dict[str, str]]:
     def repository_cache_directory(self) -> Path:
         return Path(self.get("cache-dir")) / "cache" / "repositories"
 
+    @property
+    def artifacts_cache_directory(self) -> Path:
+        return Path(self.get("cache-dir")) / "artifacts"
```
radoering, line 217:
```suggestion
        return Path(self.get("cache-dir")).expanduser() / "artifacts"
```

Not sure if it's necessary, but it's probably safer to keep this. Can you please add this in `repository_cache_directory` as well?

src/poetry/installation/executor.py
```diff
@@ -909,3 +923,40 @@ def _get_archive_info(self, package: Package) -> dict[str, Any]:
             archive_info["hashes"] = {algorithm: value}
 
         return archive_info
+
+    @staticmethod
+    def _get_cached_archive_for_link(
+        env: Env, cache_dir: Path, link: Link, *, strict: bool
+    ) -> Path | None:
```
radoering, line 930: Thinking a bit further, in the follow up PR, we will probably want to call this method (with `strict=True`) in the `Provider`. Thus, it might make sense to move this method to `cache.py`, too. Further, `env` is only required if `strict=False`. I think we should make `env` a keyword-only parameter with default `None` and add an `assert strict or env is not None` at the beginning of the method.

src/poetry/utils/cache.py
```diff
@@ -196,3 +201,35 @@ def _deserialize(self, data_raw: bytes) -> CacheItem[T]:
         data = json.loads(data_str[10:])
         expires = int(data_str[:10])
         return CacheItem(data, expires)
+
+
+def get_cached_archives_for_link(cache_dir: Path, link: Link) -> list[Path]:
```
radoering, line 206: Since all of these methods require the `cache_dir`, which will not change in one invocation of poetry, I think I'd like it better to introduce a class `ArtifactCache` that takes the cache directory in its `__init__()` and make all these functions instance methods of this class.

radoering (changes requested):

src/poetry/installation/executor.py
ralbertazzi: I agree in general, and that's how I first refactored it. But since `get_cached_archive_for_link` depends from classes in `installation` (`Wheel` and `InvalidWheelName`) I ended up in a circular import issue that could be solved only by emptying `poetry/installation/__init__.py`. At that point though other things, such as the `poetry-export-plugin`, started failing... While I know that it's not the main codebase's responsibility to make sure that plugins don't break, I felt that it would be better avoiding possibly breaking plugins if there isn't a really strong reason for doing that.

I propose you a couple alternative solutions resulting `get_cached_archive_for_link` in `utils.cache.py`, let me know which one you prefer:
1. `installation` is not touched, we `import poetry.installation.choose.Wheel` _inside_ the function to avoid circular import
1. `poetry.installation.chooser.Wheel` and `poetry.installation.chooser.InvalidWheelName` are moved into a separate utils file `poetry.utils.wheel`
1. Duplicate the small wheel logic we need, thus making `cache.py` and `Wheel` completely decoupled
```python
# NOW
from poetry.installation.chooser import InvalidWheelName
from poetry.installation.chooser import Wheel

try:
    wheel = Wheel(archive.name)
except InvalidWheelName:
    continue
```

radoering: A fourth alternative would be to split `get_cached_archive_for_link` into a strict version that is moved to `utils.cache` and a non-strict version that is kept in `executor`.

However, I think your variant 2 sounds good.

ralbertazzi: I was just performing the change to variant 2 :D Now I guess I'm just missing the class-based refactor. I'll ask you for another review once I closed all comments. Thanks a lot!

radoering (approved):

PR #7637, opened 2023-03-11
package category is not useful

dimbleby:
companion to https://github.com/python-poetry/poetry-core/pull/561


dimbleby:
the continuing entanglement of this repository with poetry-plugin-export seems to make https://github.com/python-poetry/poetry-plugin-export/pull/188 a prerequisite for merging this.

It sure would be nice to get #5980 done

dimbleby:
I don't think a version bump is necessary - afaics all versions of poetry from 1.2.0 onwards default the category ` package.category = info.get("category", "main")` and will therefore have no objection to it being absent.

(versions earlier than that already can't read the current lockfile format)

radoering (approved):

eloquence:
@dimbleby 

I just want to note for the record that this change led to a fair amount of downstream breakage. GitHub's Dependabot poetry lockfile parsing broke for weeks (https://github.com/dependabot/dependabot-core/pull/7418) and PyUp Safety is still broken (https://github.com/pyupio/safety/issues/479) - there are probably others. I think as Poetry has become rightly widespread, predictability of file formats is becoming more important. 

I don't know what the right solution is to mitigate such breakage (more strict semantic versioning? longer deprecation timelines/warnings?) but as a downstream user affected by both breakages I did want to at least note it.

dimbleby:
poetry's lockfile has never been part of its API: it's an internal implementation detail.

if you know of projects that rely on its format - please encourage them not to do so!

eloquence:
I would suggest reviewing whether that stance is sustainable.  GitHub's Dependabot, for example, is hardly a minor third party use case - it's a project many thousands of repositories rely on to monitor their projects for security vulnerabilities. And I don't see how it can do its job _without_ parsing the lockfile, after all, it's only the lockfile that gives a vulnerability analysis tool like Dependabot the _exact_ information it needs to know whether the _actually used_ versions of dependencies are vulnerable.

PR #7658, opened 2023-03-17
6713 Introduce explicit package source

b-kamphorst:
# Pull Request Check List

Partially addresses #6713, specifically addresses https://github.com/python-poetry/poetry/pull/6879#issuecomment-1459581673.


b-kamphorst:
@radoering I prepared a PR for adding `explicit` sources as requested in https://github.com/python-poetry/poetry/pull/6879#issuecomment-1459581673. The commit history is identical apart from the removal of `supplemental`-related changes in every commit. If you want I can clean up the commit history, but considering that we have been working on this already for quite a while and you may want to squash the changes anyway, I'll only make the effort if you explicitly (pun intended) request it.

src/poetry/factory.py
```diff
@@ -132,31 +134,44 @@ def create_pool(
...
         if pool.has_default():
             if io.is_debug():
                 io.write_line("Deactivating the PyPI repository")
         else:
             from poetry.repositories.pypi_repository import PyPiRepository
 
-            default = not pool.has_primary_repositories()
             pool.add_repository(
-                PyPiRepository(disable_cache=disable_cache), default, not default
+                PyPiRepository(disable_cache=disable_cache), priority=Priority.DEFAULT
```
b-kamphorst, line 174: TODO: make SECONDARY.

dimbleby:
if `default` is deprecated, what is the intended replacement for that?  ie how do I now disable use of pypi and insist that everything come from my private repository?

`primary` doesn't seem to do it, or anyway the docs only say that primary sources are considered before the default source: they don't give me any way to remove that default source.  (It's possible that this is a documentation gap or even a reading comprehension gap).

I've said this a couple of times in related MRs and issues but to avoid such gaps can I once again encourage development of a list or table of: some common use cases for multiple repositories, and how those use cases are achieved by any new proposal.

I think https://github.com/python-poetry/poetry/issues/6713#issuecomment-1270562226 was my first try at asking for this, and should provide a starting point.

radoering:
> if `default` is deprecated, what is the intended replacement for that? ie how do I now disable use of pypi and insist that everything come from my private repository?

I've not reviewed this PR yet, but this PR should only introduce a new source type called `explicit` and instead of having a flag for each type like

```
secondary=false
default=false
explicit=true
```

in pyproject.toml, just define one property, e.g. `priority="default"` or `priority="explicit"`.

Deprecating `default` sources is not in scope of this PR. #7430 is one approach. I have another approach in mind, which allows to set the priority of PyPI and thus covers more use cases. Then, you can disable it by setting its priority to `"explicit"`. IMO `default` is so broken it is difficult to fix and it's not really required as a priority.

docs/dependency-specification.md
```diff
@@ -258,7 +258,9 @@ you can use the `source` property:
 [[tool.poetry.source]]
 name = "foo"
 url = "https://foo.bar/simple/"
+default = false
 secondary = true
+explicit = false
```
radoering, line 263:
```suggestion
priority = "secondary"
```

replace deprecated syntax

docs/repositories.md
```diff
@@ -122,10 +122,19 @@ name = "foo"
 url = "https://foo.bar/simple/"
 default = false
 secondary = false
+explicit = false
```
radoering, line 125: Shouldn't it generate something with `priority = ...`?

docs/repositories.md
```diff
@@ -122,10 +122,19 @@ name = "foo"
...
 ```
 
-Any package source not marked as `secondary` will take precedence over [PyPI](https://pypi.org).
+If all keys are `false`, the source is considered a "primary" source that takes precedence over default, secondary and explicit sources.
 
+Package sources are considered in the following order:
+1. "primary" sources,
+2. [default source](#default-package-source),
+3. [secondary sources](#secondary-package-sources),
```
radoering, line 133: Not quite correct. We shouldn't care about `default` too much here because I don't want to change default handling in this PR.

src/poetry/console/commands/source/add.py
```diff
@@ -28,20 +29,31 @@ class SourceAddCommand(Command):
             (
                 "Set this source as the default (disable PyPI). A "
                 "default source will also be the fallback source if "
-                "you add other sources."
+                "you add other sources. (DEPRECATED)"
```
radoering, line 32:
```suggestion
                "you add other sources. (<warning>Deprecated</warning>)"
                " Use --priority instead."
```

Same for `secondary`.

src/poetry/repositories/repository_pool.py
```diff
@@ -71,13 +76,23 @@ def has_repository(self, name: str) -> bool:
...
+            return self._repositories[name]
         raise IndexError(f'Repository "{name}" does not exist.')
 
     def add_repository(
-        self, repository: Repository, default: bool = False, secondary: bool = False
+        self,
+        repository: Repository,
+        default: bool = False,
+        secondary: bool = False,
+        priority: Priority = Priority.PRIMARY,
```
radoering, line 95:
```suggestion
        secondary: bool = False,
        *,
        priority: Priority = Priority.PRIMARY,
```

(makes removing `default` and `secondary` easier in the future)

radoering (commented):
Some first feedback.

docs/repositories.md
b-kamphorst: I think I have made the MR compatible with the current behavior (as described in https://github.com/python-poetry/poetry/pull/6879#pullrequestreview-1273887207).
- `Priority.DEFAULT` is the highest priority.
- PyPI is disabled if another source has priority `DEFAULT`.
- PyPI has priority `DEFAULT` if no other sources are `DEFAULT` _or_ `PRIMARY`.
- PyPI has priority 'SECONDARY' if no other sources are `DEFAULT`, but at least one is `PRIMARY`.

As we now have several types of source priorities, it feels right to still mention the default and PyPI in the documentation. I kept it very minimal for now in line with your comment.

Please carefully review the latest commits.

src/poetry/console/commands/source/add.py
b-kamphorst: Changed into `<warning>Deprecated, use --priority</warning>`

src/poetry/json/schemas/poetry.json
```diff
@@ -39,7 +39,16 @@
...
+          "description": "Declare this repository as secondary, i.e. default repositories take precedence."
+        },
+        "priority": {
+          "enum": [
+            "primary",
+            "default",
+            "secondary",
+            "explicit"
+          ],
+          "description": "Declare the priority repository of this repository."
```
TCherici, line 51: I think this should be `Declare the priority repository of this repository.`

b-kamphorst: Indeed, thank you for pointing that out!

src/poetry/json/schemas/poetry.json
```diff
@@ -39,7 +39,16 @@
         },
         "secondary": {
           "type": "boolean",
-          "description": "Declare this repository as secondary, i.e. it will only be looked up last for packages."
+          "description": "Declare this repository as secondary, i.e. default repositories take precedence."
```
radoering, line 42:
```suggestion
          "description": "Declare this repository as secondary, i.e. default repositories take precedence. (deprecated, see priority)"
```

same for `default`

src/poetry/console/commands/source/add.py
```diff
@@ -50,8 +69,30 @@ def handle(self) -> int:
             )
             return 1
 
+        if is_default or is_secondary:
+            if priority is not None:
+                self.line_error(
+                    "Priority was passed through both --priority and a deprecated flag"
+                    " (--default or --secondary). Please only provide one of these."
```
radoering, line 76:
```suggestion
                    "<error>Priority was passed through both --priority and a deprecated flag"
                    " (--default or --secondary). Please only provide one of these.</error>"
```
(red color)

src/poetry/console/commands/source/add.py
```diff
@@ -50,8 +69,30 @@ def handle(self) -> int:
...
+                )
+                return 1
+            else:
+                self.line(
+                    "Priority was set through both a deprecated flag (--default or"
+                    " --secondary). Consider using --priority next time."
+                )
```
radoering, line 83:
```suggestion
                self.line_error(
                    "<warning>Warning: Priority was set through a deprecated flag (--default or"
                    " --secondary). Consider using --priority next time.</warning>"
                )
```
(yellow color plus explicit word "Warning" for colorless IOs)

radoering (changes requested):

src/poetry/console/commands/source/add.py
b-kamphorst: Should this then also start with "Error", in line with https://github.com/python-poetry/poetry/pull/7658/files#r1155309417? I do note that other occurrences of `self.line_error("<error>...` do also not start with "Error:", but I don't know whether that is on purpose.

radoering: We are not very consistent about that. You always have to consider that you can't rely just on colors. It's probably good practice to start with `"Error:"` (at least if it's not absolutely clear from the message itself).

radoering: On the other hand, the text of uncaught exceptions is only printed in red without any prefix. Thus, I'd say `Error:` or not should be clarified in general and that's something for another PR.

b-kamphorst:
I am quite sure that I addressed all feedback, but GitHub seems to think I have one open requested change (conversation view). In files view it states that everything is resolved. Either way: if there are remaining changes please let me know so that we can wrap this up :)

radoering:
I squashed the changes into two commits and added some minor changes:
- added some tests for the last bit of coverage and fixed a special case for updating an existing source via `source add`
- reverted removal of `has_primary_repositories()` (Might be used by plugins. I wondered if we should deprecate it, but decided against it for now.)
- added a docstring to `RepositoryPool.repositories` and another property `RepositoryPool.all_repositories`. I wondered if `repositories` should actually return all repositories (including explicit ones), but decided against it (see docstring for reasoning).

@b-kamphorst Please let me know if the last two points make sense to you?

b-kamphorst:
> * reverted removal of `has_primary_repositories()` (Might be used by plugins. I wondered if we should deprecate it, but decided against it for now.)

It might indeed be, so reverting the removal is completely justified. Perhaps deprecation is something to reconsider when you decided on the future of "default" repositories, specifically if that also deprecates `RepositoryPool.has_default`.

> * added a docstring to `RepositoryPool.repositories` and another property `RepositoryPool.all_repositories`. I wondered if `repositories` should actually return all repositories (including explicit ones), but decided against it (see docstring for reasoning).

I think this makes sense. If we could design this from scratch we'd probably opt to rename `RepositoryPool.repositories` to something that signals it returns a selection / applies a filter, but we can't. `RepositoryPool.repositories_unfiltered` is longer but perhaps `repositories`/`repositories_unfiltered` is slightly more informative than `repositories`/`all_repositories`? I don't have strong feelings about this though.

radoering (approved):
@b-kamphorst Thanks for your feedback and your patience. ❤️ 

> `RepositoryPool.repositories_unfiltered` is longer but perhaps `repositories`/`repositories_unfiltered` is slightly more informative than `repositories`/`all_repositories`? I don't have strong feelings about this though.

For the record: I'll stick to `all_repositories` here.

Issue #7669, opened 2023-03-17
Support reading metadata from PEP 621 compliant pyproject.toml


ughstudios:
Flit-core uses a dependencies block to determine requirements. 

https://flit.pypa.io/en/stable/pyproject_toml.html?highlight=dependencies#dependencies

But if you look at utils/setup_reader.py -> read_setup_cfg() it only checks for "python_requires", "install_requires...etc"

Could we add support for other backends?

Also it looks like it's only able to get the requirements from setup.py or setup.cfg, why doesn't it read pyproject.toml?

ughstudios:
Why can't poetry grab the info for the package from pyproject.toml? Why is it only gathering this data from setup.py or setup.cfg?

dimbleby:
duplicate #3332, please close

finswimmer:
I've edited the title of this ticket, because it was misleading. Poetry is able to gather metadata from flit and other, as long as they are PEP 518 compliant. Until now `setup.py` was a first class citizen, to avoid unnecessary builds to get the final metadata. For all other build-backend Poetry falls back to build the wheel file to have access to metadata it needs.

Miuler:
What's new?

maribedran:
The lack of support for the standard `pyproject.toml` format breaks the usage of [`zest.releaser`](https://zestreleaser.readthedocs.io/en/latest/index.html) in Poetry projects. The library looks for the `[project]` section on the `pyproject.toml` file as specified by PEP 621 to determine the project's version number. While Poetry expects this attribute, as well as `name`, to be under `[tool.poetry]`.

Secrus:
@maribedran support for PEP 621 in Poetry was recently merged. This issue concerns reading package metadata from editable projects included in your Poetry projects.

PR #7670, opened 2023-03-17
Add support for pyproject.toml files when adding --editable projects …

ughstudios:
There is currently a problem where setup_reader.py can only read setup.cfg or setup.py files for adding local packages. This adds support for flitcore and other backends that are defined in pyproject.toml files

Resolves: #7669

### DIFF
--- a/src/poetry/utils/setup_reader.py
+++ b/src/poetry/utils/setup_reader.py
@@ -2,11 +2,18 @@ from __future__ import annotations
 
 import ast
 
+from enum import Enum
 from configparser import ConfigParser
 from pathlib import Path
 from typing import Any
 
+from poetry.core.toml.file import TOMLFile, TOMLDocument
 from poetry.core.constraints.version import Version
+
+
+class SupportedBackends(Enum):
+    UNSUPPORTED = 'UNSUPPORTED'
+    FLIT_CORE = 'flit_core.buildapi'
 
 
 class SetupReader:
@@ -22,7 +29,7 @@ class SetupReader:
         "python_requires": None,
     }
 
-    FILES = ["setup.py", "setup.cfg"]
+    FILES = ["setup.py", "setup.cfg", "pyproject.toml"]
 
     @classmethod
     def read_from_directory(cls, directory: str | Path) -> dict[str, Any]:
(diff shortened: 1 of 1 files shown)


ughstudios:
> I'd anti-recommend using `tomlkit` to parse pyproject.toml here. poetry already has `tomllib` in its dependencies - that'll be faster and also sidestep the type-checking issues currently visible in the pipeline

The codebase I work on is only 3.10, do you mind if I add a fallback to use tomlkit for versions <3.11? tomllib is only available in 3.11.

dimbleby:
> The codebase I work on is only 3.10, do you mind if I add a fallback to use tomlkit for versions <3.11? tomllib is only available in 3.11.
```python
from poetry.utils._compat import tomllib
```

ughstudios:
Is there a way to run these tests at my desk so I don't keep spamming you? lol

ughstudios:
I'm running through all of the tests in the test suite. Are any of the tests known to be broken? It seems like my change has broken like 10+ tests :(

dimbleby:
If you've figured out how to run the tests then you can very easily verify for yourself that they pass without your changes.

On the plus side this probably points you in the right direction for figuring out where to add new tests.

src/poetry/inspection/info.py
```diff
@@ -498,6 +499,7 @@ def from_directory(cls, path: Path, disable_build: bool = False) -> PackageInfo:
                     else:
                         info = get_pep517_metadata(path)
                 except PackageInfoError:
+                    info = cls.from_pyproject_toml(path)
```
FFY00, line 502: IMO, instead of putting this here, we should probably create a `from_project_static_files` which tries `from_pyproject_toml` and then `from_setup_files`, and then replace all existing `from_setup_files` usages with `from_project_static_files`.

FFY00 (approved):
Looks good. I think the test failures are just due to the test suite relying too much on mocks, making it fragile when we touch the internals. I am also getting them in #7703, but haven't gotten around to look at them.

src/poetry/inspection/info.py
ughstudios: I'll have to do this on Monday when I'm back at my desk. Are you one of the maintainers of poetry? It says "FFY00 approved these changes 3 days ago."

Secrus:
Stalled

PR #7673, opened 2023-03-18
Misc tidying

dimbleby:
A series of commits cleaning up the code or making teensy corrections

most substantially, there are a couple of commits here that move further in the direction of preferring `pathlib.Path` as the internal representation when dealing with paths; aiming to convert to string only when running into third-party libraries that require that.


Secrus (dismissed):

Secrus:
Please resolve conflicts

radoering (approved):

PR #7674, opened 2023-03-18
drop support for python 3.7

dimbleby:
python 3.7 eol is approaching fast, let's drop support


Secrus:
Maintainers have decided, that we want to keep support for Python 3.7 until it's officially dropped.

dimbleby:
heh, I will have a small wager with myself about whether that's before or after the release of poetry 1.5.0

Secrus:
> heh, I will have a small wager with myself about whether that's before or after the release of poetry 1.5.0

Did you win? 😛

PR #7680, opened 2023-03-18
prefer stdlib TemporaryDirectory

dimbleby:
dropping python 3.7 per #7674 moves poetry past the Windows python 3.7 bugs that made the homegrown `temporary_directory()` worthwhile


eamanu (approved):
LGTM, we just need to wait for 3.7 drop :-)

dimbleby:
> LGTM, we just need to wait for 3.7 drop :-)

In fact https://github.com/python-poetry/poetry-core/pull/460 seems to leave poetry committed to its own implementation until support for python 3.9 (or windows!) can be dropped

eamanu:
@dimbleby  try to avoid force push

dimbleby:
@eamanu thanks but no thanks.  

per my last comment this is unlikely to be merged for quite some time and it is absolutely more sensible that when that time comes it will be best represented as a single commit on top of `main`.

All due respect but you have now made two low-value comments on this MR, please stop.

dimbleby:
with support for python 3.9 now dropped, `ignore_cleanup_errors` is always available in the standard library `TemporaryDirectory` - so homegrown workarounds are no longer needed

PR #7693, opened 2023-03-20
feat: let Provider use ArtifactCache

ralbertazzi:
# Pull Request Check List

Follow-up of https://github.com/python-poetry/poetry/pull/7621. Closes https://github.com/python-poetry/poetry/issues/2415

@radoering most of the PR is about initializing the `ArtifactCache` and passing it around the codebase whenever it's needed. Let me know if you prefer a different approach. For instance I saw usages of `Config.create()` around the codebase (such as in `CachedRepository`) that would simplify the changes a lot, as one would simply need to create an `ArtifactCache` right inside `get_package_from_url`. On the other side I like the DI approach as it allows for better modularity and testability, as the cost of more verbose code. I'll not fight this one though, I'll gladly go for the option you recommend ;) 

- [ ] Added **tests** for changed code. -> I'll do it as soon as we agree on the best way to refactor the code
- ~[ ] Updated **documentation** for changed code.~ Don't think it's needed

### DIFF
--- a/src/poetry/console/commands/debug/resolve.py
+++ b/src/poetry/console/commands/debug/resolve.py
@@ -113,7 +113,7 @@ class DebugResolveCommand(InitCommand):
 
         if self.option("install"):
             env = EnvManager(self.poetry).get()
-            pool = RepositoryPool()
+            pool = RepositoryPool(config=self.poetry.config)
             locked_repository = Repository("poetry-locked")
             for op in ops:
                 locked_repository.add_package(op.package)
--- a/src/poetry/puzzle/provider.py
+++ b/src/poetry/puzzle/provider.py
@@ -107,6 +107,7 @@ class Provider:
     ) -> None:
         self._package = package
         self._pool = pool
+        self._direct_origin = DirectOrigin(self._pool.artifact_cache)
         self._io = io
         self._env: Env | None = None
         self._python_constraint = package.python_constraint
@@ -311,7 +312,7 @@ class Provider:
         Basically, we clone the repository in a temporary directory
         and get the information we need by checking out the specified reference.
         """
-        package = DirectOrigin.get_package_from_vcs(
+        package = self._direct_origin.get_package_from_vcs(
             dependency.vcs,
             dependency.source,
             branch=dependency.branch,
@@ -330,7 +331,7 @@ class Provider:
 
     def _search_for_file(self, dependency: FileDependency) -> Package:
         dependency.validate(raise_error=True)
-        package = DirectOrigin.get_package_from_file(dependency.full_path)
+        package = self._direct_origin.get_package_from_file(dependency.full_path)
 
         self.validate_package_for_dependency(dependency=dependency, package=package)
 
@@ -348,7 +349,7 @@ class Provider:
 
     def _search_for_directory(self, dependency: DirectoryDependency) -> Package:
         dependency.validate(raise_error=True)
-        package = DirectOrigin.get_package_from_directory(dependency.full_path)
+        package = self._direct_origin.get_package_from_directory(dependency.full_path)
 
         self.validate_package_for_dependency(dependency=dependency, package=package)
 
@@ -360,7 +361,7 @@ class Provider:
         return package
 
     def _search_for_url(self, dependency: URLDependency) -> Package:
-        package = DirectOrigin.get_package_from_url(dependency.url)
+        package = self._direct_origin.get_package_from_url(dependency.url)
 
         self.validate_package_for_dependency(dependency=dependency, package=package)
 
--- a/src/poetry/puzzle/provider.py
+++ b/src/poetry/puzzle/provider.py
@@ -1,16 +1,11 @@
 from __future__ import annotations
 
-import functools
 import logging
-import os
 import re
-import tempfile
 import time
-import urllib.parse
 
 from collections import defaultdict
 from contextlib import contextmanager
-from pathlib import Path
 from typing import TYPE_CHECKING
 from typing import cast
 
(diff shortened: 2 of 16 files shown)


src/poetry/console/commands/debug/resolve.py
```diff
@@ -82,8 +83,11 @@ def handle(self) -> int:
         )
 
         pool = self.poetry.pool
+        artifact_cache = ArtifactCache(
```
ralbertazzi, line 86: Something I'm not entirely convinced about is the frequent repetition of this in this PR
```python
ArtifactCache(
    cache_dir=self.poetry.config.artifacts_cache_directory
)
```

On the other side I couldn't find a place where to define this once 🤔 The only place I actually thought was the `Poetry` object but I first wanted to get feedback on this, as I kind of consider that class an important one. On the other side I think having access to the `ArtifactCache` in `Poetry` would increase usability in potential plugins

src/poetry/puzzle/provider.py
```diff
@@ -454,15 +458,33 @@ def _search_for_url(self, dependency: URLDependency) -> Package:
         return package
 
     @classmethod
-    def get_package_from_url(cls, url: str) -> Package:
+    def get_package_from_url(
```
ralbertazzi, line 461: The function still allows `None` because of `poetry/console/commands/init.py`. I don't exactly know when that `PyProjectException` can be raised (I just copied it from a few lines above). If the exception can never be raised (but I doubt) then I would gladly simplify the codebase and remove the tempdir usage

radoering:
> For instance I saw usages of `Config.create()` around the codebase (such as in `CachedRepository`) that would simplify the changes a lot

... and probably introduce bugs. This only considers the global config and neglect the local config. (I'm aware that there are such bugs in the code base.) Here you can see what's necessary to create a config that considers global and local settings https://github.com/python-poetry/poetry/blob/71b3d6e030c879f0b4b8e3a08eac2b38a402b6bd/src/poetry/factory.py#L62-L71

ralbertazzi:
@radoering forgive me for summoning you here, I'm just unsure whether your comment is actually part of the review - suggesting that I should go for a simpler refactor that uses `Config.create()` but that also considers the local config - or if it was just an agreement comment and I should actually wait for your proper review

radoering:
Rather agreement. I just wanted to prevent you from doing a refactoring and calling `Config.create()` in various places. Even if the local config is considered, I don't think it's a good idea to read the config files several times. So nothing to do for now.

src/poetry/console/commands/debug/resolve.py
radoering: It might make sense to make it an attribute of `RepositoryPool` because both are often passed around together. The only thing that makes me hesitate is that strictly speaking direct origin dependencies are not provided by a repository. At the moment, I tend to do it regardless of this fact.

src/poetry/puzzle/provider.py
radoering: `PyProjectException` is raised if there is an issue with `pyproject.toml`. I think you can remove the exception handling for `ArtifactCache` in `init.py`.

ralbertazzi: Several tests of `init` where indeed failing when moving the artifact_cache fetching out of the try/except, so I went for keeping it

src/poetry/console/commands/debug/resolve.py
ralbertazzi: I did perform the change. It's still not extremely clean due to the fact that `RepositoryPool` is initialized more than once in the codebase, but it definitely has a smaller impact in term of PR changes. LMK if this is what you were envisioning. If so, I'll proceed writing some tests

ralbertazzi:
@radoering while we finalize the form of this PR, would it be possible to include https://github.com/python-poetry/poetry/pull/7621 in the next Poetry release? It should be a harmless change but it would allow me to write a very easy plugin using the ArtifactCache and start using it ASAP in my organization. I'd make sure the plugin will be compatible with these changes ;)

radoering:
#7621 will be included in the next minor release since it has already been merged into master. However, the next release is not yet imminent so this PR might make it in, too.

src/poetry/console/commands/debug/resolve.py
radoering: I had to do some trial and error for myself to figure out what is probably the best. I decided to pass the `config` to the `RepositoryPool`. This way, we can make sure that we always have an artifact cache without adding a mandatory argument to `RepositoryPool`. By the way, that's analogous to `cache_dir` in `CachedRepository`.

Further, I inserted two additional refactoring commits before your change. I moved all `get` methods of the different direct origin variants from `provider` into a separate module.

If you think the changes are good, I would be happy if you can write some tests. (I think we need at least some test for `get_package_from_url` that checks that `download_file` is only called if the artifact is not cached. That's probably similar to one of the tests in `test_executor.py`.)

src/poetry/puzzle/provider.py
radoering: Ah, that's because there may be no pyproject.toml when running `poetry init`. I handled that as well.

ralbertazzi:
Thanks for the great contribution, I definitely like the changes!
I also added a few tests to cover `DirectOrigin`.

Although I'm eager to see this PR merged, I just thought that Poetry supports a `--no-cache` argument, while this PR will _always_ cache URL dependencies regardless of the parameter value. To be fair the documentation mentions it more as a debugging parameter, which is the reason why I'd still be in favour of merging it and see if anybody complains through an issue (I'm sure this change would get more praises than complaints).

radoering:
> Although I'm eager to see this PR merged, I just thought that Poetry supports a `--no-cache` argument, while this PR will _always_ cache URL dependencies regardless of the parameter value.

I also thought about it. However, as far as I know `--no-cache` as well as the `poetry cache` command only handle `<cache-dir>/cache` and ignore `<cache-dir>/artifacts`. That's why I think we should _not_ take `--no-cache` into account here. Improving the documentation, let alone extending the command/option to consider the artifact cache is something for another PR.

radoering (approved):

Issue #7698, opened 2023-03-21
Python extensions that print very long build logs to stdout causes poetry install to hang 


fzahle:
- **Poetry version**: 1.4.1
- **Python version**: 3.11.2 (Homebrew)
- **OS version and name**: macOS 12.2.1
- **pyproject.toml**: https://gist.github.com/fzahle/3f0ae0237b325e736246bd467e56e6d5

- [ x] I am on the [latest](https://github.com/python-poetry/poetry/releases/latest) stable Poetry version, installed using a recommended method.
- [x ] I have searched the [issues](https://github.com/python-poetry/poetry/issues) of this repo and believe that this is not a duplicate.
- [x ] I have consulted the [FAQ](https://python-poetry.org/docs/faq/) and [blog](https://python-poetry.org/blog/) for any relevant entries or release notes.
- [x ] If an exception occurs when executing a command, I executed it again in debug mode (`-vvv` option) and have included the output below.

## Issue

I have a package with an extension I compile with CMake via `build.py`, which happens to produce an enormous amount of warnings. The code compiles fine, and executing `poetry build` works, but `poetry install` hangs during the call to the `build.py` script. I've isolated the problem to this change by @dimbleby: https://github.com/python-poetry/poetry/commit/012fcb99b8b7561ccbe138f6097cfbd0d96bb39c#diff-a0b29e9c4cccbaeea08eefe3abe15e6cffe0fdcb998ad87a4a39bfeaa5aec27fR1540 where `stdout=subprocess.PIPE` was added to the subprocess call. I believe the issue is exactly what is written in the warning box of the `subprocess.call` doc string (https://docs.python.org/3/library/subprocess.html#subprocess.call), that using PIPE can fill the OS pipe buffer causing the process not to return. Indeed, removing `subprocess.PIPE` from the subprocess call in L1540 fixes the problem.

In the gist the problem can be reproduced with a very simple example where `build.py` prints 10000 lines. Changing it to print only 10 lines `install` works.

I've also reproduced the issue in a Docker container. I used `quay.io/pypa/manylinux2014_x86_64:latest`.

dimbleby:
not sure why you've summoned me, I don't seem to have had anything to do with that change!

either way, I expect a pull request would be welcome

PR #7699, opened 2023-03-21
remove stdout=subprocess.PIPE from env._call, see #7698

fzahle:
# Pull Request Check List

Resolves: #7698

I have zero insight in the code base and unfortunately can't see how I would add a test for this fix. But a minimal example can be found here https://gist.github.com/fzahle/3f0ae0237b325e736246bd467e56e6d5, that could be the basis for a test.


fzahle:
pre-commit.ci autofix

radoering (approved):

Issue #7046, opened 2022-11-16
package installation failure


sileht:
- **Poetry version**: 1.2.2
- **Python version**: 3.10.8
- **OS version and name**: Ubuntu 20.04
- **pyproject.toml**:  https://gist.github.com/sileht/1ea186354c1df26c4cbed623b6582157

- [X] I am on the [latest](https://github.com/python-poetry/poetry/releases/latest) stable Poetry version, installed using a recommended method.
- [X] I have searched the [issues](https://github.com/python-poetry/poetry/issues) of this repo and believe that this is not a duplicate.
- [X] I have consulted the [FAQ](https://python-poetry.org/docs/faq/) and [blog](https://python-poetry.org/blog/) for any relevant entries or release notes.
- [X] If an exception occurs when executing a command, I executed it again in debug mode (`-vvv` option) and have included the output below.

## Issue

poetry failed to install packages from time to time with the error below.
I suspect it only occurs with the package is not pep518 compatible and pip requires building a wheel.
I'm not sure if it's directly linked to poetry, or if this is just poetry that triggers a pip/setuptools bug?

It's not 100% reproductive, but it occurs a lot inside our GitHub workflows.

```
#39 18.26   • Installing urllib3 (1.26.12)
#39 19.62   • Installing asyncpg (0.27.0)
#39 19.62   • Installing beautifulsoup4 (4.11.1)
#39 19.62   • Installing cattrs (22.2.0)
#39 19.62   • Installing click (8.1.3)
#39 19.62   • Installing bytecode (0.13.0)
#39 19.62   • Installing cryptography (38.0.3)
#39 21.00   • Installing ddsketch (2.0.4)
#39 21.03   • Installing envier (0.4.0)
#39 21.03   • Installing fastapi (0.86.0)
#39 21.17   • Installing forbiddenfruit (0.1.4)
#39 21.47   • Installing greenlet (1.1.3.post0)
#39 21.94   • Installing h2 (4.1.0)
#39 22.40   • Installing httpcore (0.15.0)
#39 22.50   • Installing httptools (0.5.0)
#39 22.54   • Installing itsdangerous (2.1.2)
#39 22.89   • Installing jsonschema (4.17.0)
#39 23.34   • Installing markupsafe (2.1.1)
#39 23.83   • Installing passlib (1.7.4)
#39 23.87   • Installing python-dotenv (0.21.0)
#39 23.89   • Installing python-json-logger (2.0.4)
[...]

PR #7701, opened 2023-03-22
executor: update setuptools/wheel first before other (parallel) installs

sileht:
If a package is build from source because no pep517 build system are
indicated in pyproject.toml or the file is missing, pep517 will fallback
to setuptools and wheel installed within the virtualenv to build it.

eg:

https://github.com/pypa/pip/blob/22.3.1/src/pip/_vendor/pep517/wrappers.py#L167
https://github.com/pypa/pip/blob/22.3.1/src/pip/_vendor/pep517/wrappers.py#L220
https://github.com/pypa/pip/blob/22.3.1/src/pip/_vendor/pep517/wrappers.py#L273

Resolves: #7046

### DIFF
--- a/src/poetry/installation/executor.py
+++ b/src/poetry/installation/executor.py
@@ -157,9 +157,10 @@ class Executor:
         self._sections = {}
         self._yanked_warnings = []
 
-        # pip has to be installed first without parallelism if we install via pip
+        # pip/setuptools/wheel has to be installed first without parallelism
+        # if we install via pip
         for i, op in enumerate(operations):
-            if op.package.name == "pip":
+            if op.package.name in ("setuptools", "pip", "wheel"):
                 wait([self._executor.submit(self._execute_operation, op)])
                 del operations[i]
                 break


sileht:
> Why?

I added the missing issue in the description, eg: Resolves: https://github.com/python-poetry/poetry/issues/7046

sileht:
What is the minimal version of pip poetry support?

Because it looks like the build-system fallback has been dropped in 23.0.

dimbleby:
I still don't understand what the connection is between the installation of `setuptools` in the target environment, and the installation of `setuptools` in the temporary environment that pip uses to build (eg) forbiddenfruit as at #7046.   

Surely they are completely different `setuptools`?

sileht:
> I still don't understand what the connection is between the installation of `setuptools` in the target environment, and the installation of `setuptools` in the temporary environment that pip uses to build (eg) forbiddenfruit as at #7046.
> 
> Surely they are completely different `setuptools`?

I continue to dig, and I don't get it either.

The interesting logs are:
```
 • Updating setuptools (63.2.0 -> 65.5.1)  -----> Update of setuptools inside the target venv.
```
I was misled by this, and as you seem to confirm, this is unrelated to the issue we see later in the logs.

Then pip fails:
```
Command ['/venv/bin/python', '-m', 'pip', 'install', '--use-pep517', '--disable-pip-version-check', '--prefix', '/venv', '--no-deps', '/root/.cache/pypoetry/artifacts/91/e3/3a/ee514b36115af9f8b73c52022cf789cee2e2d0603ada6f03311b316242/forbiddenfruit-0.1.4.tar.gz']' returned non-zero exit status 2.
...
Installing build dependencies: started
Installing build dependencies: finished with status 'done' <=========== HERE a new setuptools is supposed to be installed in the temporary pep517 venv
Getting requirements to build wheel: started
Getting requirements to build wheel: finished with status 'done'
....
File "<frozen importlib._bootstrap>", line 688, in _load_unlocked
File "<frozen importlib._bootstrap_external>", line 883, in exec_module
 File "<frozen importlib._bootstrap>", line 241, in _call_with_frames_removed
File "/tmp/pip-build-env-k0jhdblz/overlay/lib/python3.10/site-packages/setuptools/__init__.py", line 8, in <module>
import _distutils_hack.override  # noqa: F401
ModuleNotFoundError: No module named '_distutils_hack.override' <======= HERE setuptools looks partially installed.
```

I'm going to close this PR. 
This clearly looks like a pip issue.

dimbleby:
I could _just about_ believe an explanation that said that pip somehow re-uses the nearest `setuptools` when creating the isolated environment in which to peform the build.

However I've no reason to think that's true, it's pure speculation.

If you're hitting the issue frequently enough that this is worth your trouble, perhaps you can also test the change to see whether - experimentally - it seems to help or not.

Issue #6154, opened 2022-08-12
poetry install does not honor `build-system.requires`


sharvil:
- [ ] I am on the [latest](https://github.com/python-poetry/poetry/releases/latest) Poetry version.
- [x] I have searched the [issues](https://github.com/python-poetry/poetry/issues) of this repo and believe that this is not a duplicate.
- [x] If an exception occurs when executing a command, I executed it again in debug mode (`-vvv` option).

- **OS version and name**: Ubuntu 20.04.4
- **Poetry version**: 1.2.0b2
- **Link of a [Gist](https://gist.github.com/) with the contents of your pyproject.toml file**:
<details>
<summary>pyproject.toml</summary>

```
[tool.poetry]
name = "meow"
version = "0.1.0"
description = ""
authors = [""]
build = "build.py"

[tool.poetry.dependencies]
python = "^3.9"
SoundFile = "0.10.*"
crcmod = "^1.7"
librosa = ">=0.8.0"
numba = "^0.56.0"
protobuf = "3.14.0"
tensorboard = "^2.10.0"
torchaudio = {version = "=0.12.1+cu113", source="torch"}
torch = {version = "=1.12.1+cu113", source="torch"}

[[tool.poetry.source]]
name = "torch"
url = "https://download.pytorch.org/whl/cu113"
secondary = true

[tool.poetry.dev-dependencies]

[build-system]
requires = ["poetry-core>=1.0.0", "setuptools>=40.6.0", "wheel", "numpy", "pybind11"]
build-backend = "poetry.core.masonry.api"
```

</details>

## Issue

Running `poetry build` with the above `pyproject.toml` does the right thing:
```
Preparing build environment with build-system requirements poetry-core>=1.0.0, setuptools>=40.6.0, wheel, numpy, pybind11
[...snip...]
```

Running `poetry install` fails because it attempts to build the project without installing `build-system.requires` packages.

I would have expected the install command to create an isolated environment, install the build requirements and perform the build in that isolated environment, and then install the produced artifact in my current environment.

bilelomrani1:
Same issue on macOS 12.5. A dirty hack is to add the following at the top of `build.py`:
```python
import subprocess

try:
    import pybind11
except ImportError:
    subprocess.run(["pip", "install", "pybind11"])
```

neersighted:
[...]

Issue #7702, opened 2023-03-22
Poetry doesn't install `build-system.requires` from dependencies not using Poetry.


adamgreig:
- **Poetry version**: 1.4.1 and current master 71b3d6e
- **Python version**: 3.10.6
- **OS version and name**: Ubuntu 22.04
- **pyproject.toml**: https://github.com/adamgreig/poetry_build_requires_bug/blob/master/poetry_bug_2/pyproject.toml

- [x] I am on the [latest](https://github.com/python-poetry/poetry/releases/latest) stable Poetry version, installed using a recommended method.
- [x] I have searched the [issues](https://github.com/python-poetry/poetry/issues) of this repo and believe that this is not a duplicate.
- [x] I have consulted the [FAQ](https://python-poetry.org/docs/faq/) and [blog](https://python-poetry.org/blog/) for any relevant entries or release notes.
- [x] If an exception occurs when executing a command, I executed it again in debug mode (`-vvv` option) and have included the output below.

## Issue

Poetry doesn't install entries in `[build-system.requires]` from dependencies not using Poetry. In my case a dependency has a build system requirement on `setuptools_scm`, which must be installed for `setup.py` to run successfully. `setuptools_scm` is listed in the dependency's `[build-system.requires]` but isn't installed by Poetry, so the overall installation fails.

Adding `setuptools_scm` to the dependency's actual dependencies list causes it to be installed first, and then the build succeeds.

I ran into this issue trying to install [amaranth](https://github.com/amaranth-lang/amaranth) but have reproduced the error in a minimal test case [here](https://github.com/adamgreig/poetry_build_requires_bug/). Run `poetry install` in `poetry_bug_2/` to see the error.

It seems like the issue (thanks to @whitequark for investigating) is here:

https://github.com/python-poetry/poetry/blob/71b3d6e030c879f0b4b8e3a08eac2b38a402b6bd/src/poetry/inspection/info.py#L599

which should possibly include the extra requirements:

```diff
-                *PEP517_META_BUILD_DEPS,
+                *(PEP517_META_BUILD_DEPS + PyProjectTOML(path.joinpath("pyproject.toml")).build_system.requires),
```

I'm happy to prepare a PR with that change if you would like.

Full error output is below but the pertinent error is:

```

Unable to determine package info for path: /home/adam/Projects/tmp/poetry_bug/poetry_bug_1

Fallback egg_info generation failed.

[...]

PR #7703, opened 2023-03-22
inspection.info: simplify get_pep517_metadata

FFY00:
Resolves: #6154, #7702

This patch simplifies `get_pep517_metadata`, makes it more efficient, and make sure we handle the build dependencies corrently on legacy (`setup.py`) projects.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -181,6 +181,7 @@ module = [
     'cachecontrol.*',
     'lockfile.*',
     'pexpect.*',
+    'pyproject_hooks.*',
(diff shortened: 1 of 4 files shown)


dimbleby:
related #7670, which takes a complementary approach: if poetry can understand pep-621 compliant pyproject.toml well enough to read the metadata from it directly then there's no need to install in this code path at all.

probably there's room for both, I've no idea whether that other has stalled or is only resting.

Secrus:
Superseded by #9168

Issue #7718, opened 2023-03-24
Poetry caching issue


Syffe:
- **Poetry version**: 1.4.1
- **Python version**: 3.11.0
- **OS version and name**: Ubuntu 20.04.6 LTS
- **pyproject.toml**: https://gist.github.com/Syffe/80bc7fe55dc2dcb7d59dd295b8e9c7c0

- [x] I am on the [latest](https://github.com/python-poetry/poetry/releases/latest) stable Poetry version, installed using a recommended method.
- [x] I have searched the [issues](https://github.com/python-poetry/poetry/issues) of this repo and believe that this is not a duplicate.
- [x] I have consulted the [FAQ](https://python-poetry.org/docs/faq/) and [blog](https://python-poetry.org/blog/) for any relevant entries or release notes.
- [x] If an exception occurs when executing a command, I executed it again in debug mode (`-vvv` option) and have included the output below.

## Issue
Poetry seems to be flaky when executing our test suite.
We're running a GitHub Actions workflow where we use poetry to install our dependencies based on pyproject.toml 

```
• Installing python-lsp-jsonrpc (1.0.0)
  • Installing pyyaml (6.0)
  • Installing restructuredtext-lint (1.4.0)
  • Installing rich (13.3.2)
  • Installing ruamel-yaml (0.17.21)
  • Installing setproctitle (1.2.3)
  • Installing sphinx (6.1.3)
  • Installing sqlalchemy (2.0.7)

  StopIteration

  0

  at /opt/hostedtoolcache/Python/3.11.2/x64/lib/python3.11/json/decoder.py:353 in raw_decode
      349│         have extraneous data at the end.
      350│ 
      351│         """
      352│         try:
    → 353│             obj, end = self.scan_once(s, idx)
      354│         except StopIteration as err:
      355│             raise JSONDecodeError("Expecting value", s, err.value) from None
      356│         return obj, end
      357│ 

The following error occurred when trying to handle this error:


  Stack trace:

  27  /opt/hostedtoolcache/Python/3.11.2/x64/lib/python3.11/site-packages/poetry/installation/executor.py:272 in _execute_operation
        result = self._do_execute_operation(operation)

  26  /opt/hostedtoolcache/Python/3.11.2/x64/lib/python3.11/site-packages/poetry/installation/executor.py:374 in _do_execute_operation
        result: int = getattr(self, f"_execute_{method}")(operation)

[...]

PR #7728, opened 2023-03-28
fix: don't fail if cache is corrupted

sileht:
Related #7718

### DIFF
--- a/src/poetry/utils/cache.py
+++ b/src/poetry/utils/cache.py
@@ -4,6 +4,7 @@ import contextlib
 import dataclasses
 import hashlib
 import json
+import logging
 import shutil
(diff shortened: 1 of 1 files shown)


sileht:
The goal is to show which cache key/path is corrupted to help to debug and get more information in future issues opened on this subject.

Tell me if the new behavior looks good, or if you prefer to keep the failure as before and convert the JSONDecodeError exception into a custom Exception that contains the key and the path of the problèmatic cache entry.

(I will add tests once we are ok with the solution).

dimbleby:
duplicates #7453

radoering:
Closing in favor of #7453

Issue #5885, opened 2022-06-21
Add more context to Poetry error messages


MSLeiter:
- [x] I have searched the [issues](https://github.com/python-poetry/poetry/issues) of this repo and believe that this is not a duplicate.
- [x] I have searched the [documentation](https://python-poetry.org/docs/) and believe that my question is not covered.

## Feature Request

I'm using Poetry 1.1.13.

I've recently gotten some Poetry errors for the first time. When Poetry works, it just works and it's great, but when Poetry has an issue it does not provide any context - at least with the 2 errors I've seen so far. This makes it difficult to understand what the issue is and how to even find the general thing that is causing the issue.

For example, I was getting a "ValueError - not enough values to unpack (expected 2, got 1)" for a script entry that was not complete in pyproject.toml, but Poetry only gives the quoted error statement, without mentioning that it was trying to parse the [tool.poetry.scripts] section of pyproject.toml, or that this was during pyproject.toml parsing at all. Preferrably it would also give the line number where the error occurred.

The other error was attempting to parse a non-semver version that was installed to the venv - again, no context to understand what Poetry was trying to do at the time. Thankfully I had just installed that alternate package, and the version name itself was clear, but I didn't know Poetry could fail just parsing the version of an item that was installed to the venv, that Poetry didn't install itself or was mentioned in pyproject.toml at all as a dependency (I hadn't done anything with it yet), so it was originally quiet stunning to get a message that Poetry could not parse a version when I hadn't changed my pyproject.toml at all.

## Request:

Please add more context to the error messages, where possible. We need to at least know what Poetry was doing generally, or what line of which file it was parsing, etc. to be able to decipher what is going on.

Some possibilities:
1. If there is a parse error, print the filename and line number (if line number is possible), or print the text that was attempting to be parsed (i.e. perhaps the raw line of text prior to parsing). With the raw text, we can search our files to find the problem location.
2. Currently the errors only show the final piece of the exception call stack - maybe Poetry should print the entire call stack? I realize often the additional steps are not needed and take up a lot of space in the shell, but sometimes they are required to understand what went wrong.

Secrus:
a lot was improved in the incoming 1.2 release. You can check it out from `master` branch

dimbleby:
When hitting an exception the stack is available by running with `-vv`.

You didn't say how to reproduce either of the errors that you're complaining about.

- The version parsing error I think I recognise and it is already covered by #4597 (please submit a pull request!)
[...]

PR #7756, opened 2023-04-01
Catch problems with [tool.poetry.scripts] entries

christokur:
# Pull Request Check List

Resolves: #5885 (partially)

Problem encountered:
Output: ` too many values to unpack (expected 2)` for current project's pyproject.toml i.e. editable install failed
As mentioned in #5885 these failures are cryptic and hard to diagnose. For example, I tried to run poetry via the debugger but then encounter a lot of challenges with `subprocess` invocations using the debugger options etc.
I eventually tracked down the error in my pyproject.toml to this pattern
```toml

[tool.poetry.scripts]
command = "module::main"

```

Solution:
At a low level I did not wish to substantially change the way the low level code deal with exceptions but I did want to add some context value to the error message the user will see. My proposed change to the `_add_script` method follows this idea.
I created a test to reproduce the problem, verified that the code changes fixes the problem and added a simple fixture for script related problems like the one I encountered and there are many more test cases we can add there


src/poetry/masonry/builders/editable.py
```diff
@@ -174,7 +174,11 @@ def _add_scripts(self) -> list[Path]:
         scripts = entry_points.get("console_scripts", [])
         for script in scripts:
             name, script = script.split(" = ")
-            module, callable_ = script.split(":")
+            try:
+                module, callable_ = script.split(":")
+            except ValueError as exc:
+                msg = f"{exc.args}  - Failed to parse script entry point '{script}'"
+                raise ValueError(msg) from exc
```
radoering, line 181:
```suggestion
                raise ValueError(f"Failed to parse script entry point '{script}'") from exc
```

radoering: > Several lint tools recommend against interpolation when instantiating Exceptions.

Sorry, I don't follow. Can you give an example?

My reasoning is that you get the original exception via `from exc`. Thus, you don't need `exc.args` it in the derived exception.

christokur: Poetry does not display the stack trace so you have to add the exception info to the message so the user can see what the symptom and cause is in one message

christokur: Also the example
```
resources.py:45: [EM102] Exception must not use an f-string literal, assign to variable first
```

radoering: Thanks for the example. I read the [reasoning to EM102](https://pypi.org/project/flake8-errmsg/0.4.0/) and I'm not sure I'm convinced (especially since we are not printing the traceback by default). Normally, we are using string literals and f-strings without assigning it to a variable first, so there is no reason to deviate from that here.

> Poetry does not display the stack trace so you have to add the exception info to the message so the user can see what the symptom and cause is in one message

You will see the traceback by running poetry with `-vvv`. When not in verbose mode, we shouldn't print the original error message since it is not very helpful anyway. If you think that `f"Failed to parse script entry point '{script}'"` is not enough information, then what about `f"Failed to parse script entry point '{script}', ':' not found"` or similar?

By the way, we have the same script parsing with the same issue in https://github.com/python-poetry/poetry/blob/161b19cb4e4686fc5a0a7925001534a87f6c4052/src/poetry/console/commands/run.py#L73

christokur: This also catches cases where someone does not place any ":"
Are we overthinking this?

radoering: Good point, my proposal is not accurate enough. However, I still stand by my initial statement.

IMO `('not enough values to unpack (expected 2, got 1)',)  - Failed to parse script entry point '...'` is not a good error message. Without any context `not enough values to unpack` does not really help. `Failed to parse script entry point '...'` should be good enough. If you want more details you have to run `poetry install -v`. That way you will get the `not enough values to unpack` with the line were it happened.

sneakers-the-rat:
came across https://github.com/python-poetry/poetry/issues/5885 and went to see if anyone had done this PR before submitting it myself, wondering why this was closed? seems like a no brainer. Re the prior discussion, it seems like it would be simple to wrap the whole block of three string splits with try: except: and just say `"expected a stript of the form entrypoint = module.submodule:function, got {whatever was input}"`

I'm pretty sure the "don't fstring in messages" is more for log messages, where you save time by not interpolating if eg. the message won't be processed due to the loglevel

radoering:
I assume the author lost interest. It shouldn't be too difficult to push it across the finish line if you address the review feedback.

sneakers-the-rat:
Cool, done https://github.com/python-poetry/poetry/pull/8898

Issue #7075, opened 2022-11-22
Poetry not using the correct python version


shanewazabbas:
```
> poetry shell
The currently activated Python version 3.9.14 is not supported by the project (>=3.10.1,<3.11).
Trying to find and use a compatible version. 
Using python3 (3.10.7)
Skipping virtualenv creation, as specified in config file.
Spawning shell within /Users/shaneabbas/.pyenv/versions/3.9.14
```

It says using 3.10.7 but the last line says otherwise. 
If I do poetry install, it chooses to still use 3.9.14

Keep in mind that I'm using pyenv and local `.python-version` file
```
> python --version
Python 3.10.7
```

```
> cat .python-version 
3.10.7
```

```
> poetry env info

Virtualenv
Python:         3.9.14
Implementation: CPython
Path:           NA
Executable:     NA

System
Platform:   darwin
OS:         posix
Python:     3.9.14
Path:       /Users/shaneabbas/.pyenv/versions/3.9.14
Executable: /Users/shaneabbas/.pyenv/versions/3.9.14/bin/python3.9
```

There is nothing crazy going on here on my end. How are we not able to handle just having the current python version be respected for poetry? I'm on a mac

shanewazabbas:
Also this doesn't work:
```
> poetry env use $(pyenv which python3.10) -vvv
Loading configuration file /Users/shaneabbas/Library/Preferences/pypoetry/config.toml
Loading configuration file /Users/XXX/poetry.toml
Skipping virtualenv creation, as specified in config file.
Using virtualenv: /Users/shaneabbas/.pyenv/versions/3.9.14
```

This is rather frustrating...

clintonroy:
I really don't understand what pyenv is, I've never used it, and thus I'm not well placed to try to help here.

Buuuut, if poetry isn't making the virtual environment, that means the user is, and  that means it's on you?

shanewazabbas:
> I really don't understand what pyenv is

Then look it up...
There's even documentation about pyenv from poetry: https://python-poetry.org/docs/managing-environments/#switching-between-environments

shanewazabbas:
I have and it still doesn't work...

shanewazabbas:
Even then the original output I posted is clearly a bug. It shouldn't be saying it's using python3 (3.10.7) when it actually isn't.
```
> poetry shell
The currently activated Python version 3.9.14 is not supported by the project (>=3.10.1,<3.11).
Trying to find and use a compatible version. 
[...]

PR #7771, opened 2023-04-06
use shutil.which() to detect the active python

dimbleby:
Resolves: #7772

https://github.com/python-poetry/poetry/issues/7075#issuecomment-1497136923 says that `subprocess.call("python")` on windows doesn't give the same answer as calling `python` from the command line does (in particular, since dropping `shell=True` on the `subprocess` call).

dropping `shell=True` still seems to me to be highly desirable, so this MR uses `shutil.which("python")` to figure out which python would have been found by typing `python` directly.

I'm not working on windows so can't readily verify that this works and it's not clear to me how to write a test script either.  Such things certainly seem desirable, but it might need someone else to pick up the baton...

### DIFF
--- a/src/poetry/utils/env.py
+++ b/src/poetry/utils/env.py
@@ -9,6 +9,7 @@ import os
 import platform
 import plistlib
 import re
+import shutil
 import subprocess
 import sys
 import sysconfig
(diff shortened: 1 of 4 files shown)


dimbleby:
second commit probably fixes #7772

xyxel:
shutil.which() works fine. However current fix doesn't work, because
`python_path = self._full_python_path(python)`
in activate method should be updated too. 
Maybe it's better to add shutil.which to _full_python_path method instead of using it outside?

xyxel:
Now it works. Thank you!

radoering (approved):

PR #7774, opened 2023-04-07
handle importlib-metadata deprecation

dimbleby:
https://github.com/python/importlib_metadata/blob/main/CHANGES.rst#v520
https://github.com/python/importlib_metadata/pull/416

### DIFF
--- a/src/poetry/repositories/installed_repository.py
+++ b/src/poetry/repositories/installed_repository.py
@@ -257,9 +257,8 @@ class InstalledRepository(Repository):
                 if path in skipped:
                     continue
 
-                try:
-                    name = canonicalize_name(distribution.metadata["name"])
-                except TypeError:
+                name = distribution.metadata.get("name")  # type: ignore[attr-defined]
+                if name is None:
                     logger.warning(
                         (
                             "Project environment contains an invalid distribution"
@@ -270,6 +269,8 @@ class InstalledRepository(Repository):
                     )
                     skipped.add(path)
                     continue
+
+                name = canonicalize_name(name)
 
                 if name in seen:
                     continue


src/poetry/repositories/installed_repository.py
```diff
@@ -258,8 +258,11 @@ def load(cls, env: Env, with_dependencies: bool = False) -> InstalledRepository:
                     continue
 
                 try:
-                    name = canonicalize_name(distribution.metadata["name"])
-                except TypeError:
+                    name = distribution.metadata["name"]
```
radoering, line 261: I was wondering where this warning came from, but was too lazy to go into the matter. Since we are handling None as well as KeyError now, we should filter this deprecation warning, shouldn't we?

dimbleby: I wondered about that but I figure no-one will ever remember to remove the filter, and the warning will go away on its own in due course.

(it wasn't a very good deprecation, IMO, since importlib-metadata hasn't provided an alternative that avoids the warning - there ought really to be a `distribution.metadata.get("name")` available.  Anyway...)

radoering: However, it will take quite a while: https://github.com/python/importlib_metadata/blob/700f2c7d74543e3695163d5487155b92e6f04d65/importlib_metadata/_adapters.py#L11

It seems even though https://github.com/python/importlib_metadata/pull/416 has been merged into main it's not in main!?

I'd prefer a filter and an explanatory comment. Even if we forget to remove it, I assume it will not really hurt. (Just added it.) Does that make sense to you?

dimbleby: Turns out there's actually a perfectly good `.get()` available, it's just that importlib-metadata hides it from mypy.

I've raised https://github.com/python/importlib_metadata/issues in the hope of getting them to admit that it's there.

If importlib-metadata seems amenable to that then I'd prefer to use `.get()` here, suppressing the mypy warning.  (Then when a new importlib-metadata comes out, we'll be alerted that we have an unused ignore)

dimbleby: there's also `__contains__` so we can ask if `"name" in distribution.metadata` without triggering warnings or needing suppressions.  So that's probably cleanest for now.

(Two lookups where one would have done does slightly bother me, but I'm sure it's completely negligible in practice.)

dimbleby:
slightly surprised - in a good way - to see that importlib-metadata MR being merged already.  So let's go back to the "use `.get()` and suppress mypy for now" plan.

radoering (approved):

PR #7777, opened 2023-04-07
tighten up some unused typechecking ignores

dimbleby:
we'll see whether the pipeline agrees, but I think that most of the `# type: ignore` markers relating to importlib-metadata are unnecessary

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -169,7 +169,6 @@ enable_error_code = [
 # warning.
 [[tool.mypy.overrides]]
 module = [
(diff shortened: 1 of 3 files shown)


dimbleby:
the pipeline agrees with only about 50% of what I thought was safe to remove, which doesn't leave very much in this MR but I guess I opened it now so here it is

dimbleby:
abandoning in favour of #7779 which includes this and some other tidy-ups that I've accumulated

PR #7797, opened 2023-04-13
Shell clear

yelircaasi:
Resolves: NA

- [✔] Added **tests** for changed code.
- [✔] Updated **documentation** for changed code.

Typically, `poetry shell` leaves lines of output that are redundant in typical use. There should be an option to have `poetry shell` drop the user into a clean shell (i.e. no previous output). This is a feature that I have been missing in my daily workflow with Poetry.

I have elected to do this via an environment variable, `$POETRY_SHELL_CLEAR`. The advantage of this approach is that this is the simplest to implement; only two lines of code have been added in addition to small testing and documentation changes.

An alternative would be to add a config option, such as `virtualenv.clear-upon-activation`, which I would be willing to implement if that is preferred.

### DIFF
--- a/docs/cli.md
+++ b/docs/cli.md
@@ -613,6 +613,8 @@ according to the `$SHELL` environment variable,
 within the virtual environment.
 If one doesn't exist yet, it will be created.
 
+If `$POETRY_SHELL_CLEAR` is set, all previous terminal output will be cleared.
+
 ```bash
 poetry shell
 ```
--- a/src/poetry/utils/shell.py
+++ b/src/poetry/utils/shell.py
@@ -112,7 +112,7 @@ class Shell:
                 f"{self._get_source_command()} {shlex.quote(str(activate_path))}"
             )
 
-        if os.environ.get("POETRY_CLEAR_SHELL"):
+        if os.environ.get("POETRY_SHELL_CLEAR"):
             c.sendline("clear")
 
         def resize(sig: Any, data: Any) -> None:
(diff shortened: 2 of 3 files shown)


yelircaasi:
> If I were you and I really cared about having a clear terminal when I activated a virtual environment I'd solve this outside of poetry eg maybe set up an alias to go `source path-to-venv/bin/activate ; clear` and not bother with `poetry shell` at all.

That's the exact problem this is supposed to solve. Because `poetry shell` drops you into a subshell, the `clear` part of the command has no power there, and the output remains. This is the case in zsh, anyway, and whether I use and alias or a function makes no difference. 

This wouldn't matter at all, except that Poetry includes some very unnecessary output and it gets tedious having to always run `clear`.

yelircaasi:
> Also your testcase is still not testing your code: the proof of which is that the testcase passes if I remove your code entirely.

It's testing that the code doesn't break any of the pre-exisiting tests, even when the env var is set and the code executes. I'm actually not sure how to verify from within pytest that output is cleared.

> I did not suggest that you set up an alias to poetry shell ; clear, I suggested source path-to-venv/bin/activate ; clear

For a number of reasons, this isn't a great alternative. There are real benefits to working in a subshell, and there are also issues with the prompt when `.../activate` is sourced instead of running `poetry shell`.

I agree this isn't a make-or-break feature for Poetry, but this is something that I wish for on a daily basis, having already tried every alternative I could think of, and finding none of them to be satisfactory.

dimbleby:
Another problem with this MR is that it doesn't work on windows

Secrus:
Ok, I am pulling a plug on this.

PR #7766, opened 2023-04-04
handle distutils deprecation

dimbleby:
`distutils` is going away in python 3.12.

I don't find this comment very clear
```
# We can't use sysconfig.get_paths() because
# on some distributions it does not return the proper paths
# (those used by pip for instance).
```
but it was written three years ago now, we can always hope that things have moved on far enough to make it no longer a thing that needs to be worried about.

Well anyway let's start by seeing how the pipelines like it if we don't use `distutils`...

Issue #2487, opened 2020-06-03
differing case for package name confuses poetry


davidszotten:
- [x] I am on the [latest](https://github.com/python-poetry/poetry/releases/latest) Poetry version.
- [x] I have searched the [issues](https://github.com/python-poetry/poetry/issues) of this repo and believe that this is not a duplicate.
- [x] If an exception occurs when executing a command, I executed it again in debug mode (`-vvv` option).

 i'm not sure exactly how to describe this issue so apologies if a duplicate issue exists

- **OS: OSX 10.15.4**
- **Poetry: 1.0.5**

## Issue
trying to add the same package with differing case seems to confuse poetry into thinking that they're different causing it to then fail to resolve two different version constraints

```
cd <empty folder>
poetry new .
poetry add django==2  # note all lowercase
# ... installs

$ poetry add Django==3  # note capitalised

Updating dependencies
Resolving dependencies... (0.0s)

[SolverProblemError]
Because poetrytest depends on both django (2) and Django (3), version solving failed.
```

danihodovic:
Experiencing the same issue with Poetry 1.1.5 during an upgrade from Django 3.0 to Django 3.2. 

```  
SolverProblemError

  Because findwork.dev depends on both django (3.0) and Django (3.2), version solving failed.


```

mkniewallner:
This has been fixed in the to-be-released Poetry `1.2` version.

Tried with [1.2.0b2](https://github.com/python-poetry/poetry/releases/tag/1.2.0b2) for instance:
```console
$ poetry new .
Created package foo in .

$ poetry add django==2
Creating virtualenv foo-V-ORBwh6-py3.9 in <REDACTED>

Updating dependencies
Resolving dependencies... (0.5s)

Writing lock file

Package operations: 2 installs, 0 updates, 0 removals

  • Installing pytz (2022.1)
  • Installing django (2.0)

$ poetry add Django==3

Updating dependencies
Resolving dependencies... (0.3s)

Writing lock file

Package operations: 2 installs, 1 update, 0 removals

  • Installing asgiref (3.5.2)
  • Installing sqlparse (0.4.2)
  • Updating django (2.0 -> 3.0)

$ poetry add django==4

Updating dependencies
Resolving dependencies... (0.1s)

Writing lock file

Package operations: 0 installs, 1 update, 1 removal

  • Removing pytz (2022.1)
  • Updating django (3.0 -> 4.0)

$ cat pyproject.toml
...
[...]

PR #7803, opened 2023-04-16
python 3.12 in pipelines

dimbleby:
expecting this to fail per #7766, let's check that's true and see whether anything else shows up

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -28,7 +28,7 @@ jobs:
     strategy:
       matrix:
         os: [Ubuntu, macOS, Windows]
-        python-version: ["3.8", "3.9", "3.10", "3.11"]
+        python-version: ["3.8", "3.9", "3.10", "3.11", "3.12-dev"]
         include:
           - os: Ubuntu
             image: ubuntu-22.04
@@ -117,7 +117,8 @@ jobs:
           ref: refs/tags/${{ steps.poetry-plugin-export-version.outputs.version }}
 
       - name: Run pytest (poetry-plugin-export)
-        run: poetry run pytest -v poetry-plugin-export/tests/
+        working-directory: ./poetry-plugin-export
+        run: poetry run -C .. pytest -v
 
       - name: Check for clean working tree
         run: |
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -375,6 +375,12 @@ If set to `true` the `--no-setuptools` parameter is passed to `virtualenv` on cr
 means when a new virtual environment is created, `setuptools` will not be installed in the environment. Poetry, for its
 internal operations, does not require `setuptools` and this can safely be set to `true`.
 
+For environments using python 3.12 or later, `virtualenv` defaults to not
+installing `setuptools` when creating a virtual environment.
+In such environments this poetry configuration option therefore has no effect:
+`setuptools` is not installed either way.
+If your project relies on `setuptools`, you should declare it as a dependency.
+
 {{% warning %}}
 Some development tools like IDEs, make an assumption that `setuptools` (and other) packages are always present and
 available within a virtual environment. This can cause some features in these tools to not work as expected.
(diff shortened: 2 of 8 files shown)


dimbleby:
not sure what's with the ubuntu and macos pipeline failure now, but the good news is that the `__getattr__` that seems to be recursed on does not exist in the main branch of poetry-core - so with any luck releasing that will solve this.

(There's a corresponding `__getattr__` now in poetry: I want to remove that too - but doing so only becomes possible when poetry-core is released.  So probably clean 3.12 pipelines is just blocked on a poetry-core release)

I've no idea what the windows failure is about

dimbleby:
getting closer

- virtualenv behaves differently at python 3.12, per their changelog:

> Do not install wheel and setuptools seed packages for Python 3.12+. To restore the old behaviour use:
>
>    for wheel use VIRTUALENV_WHEEL=bundle environment variable or --wheel=bundle CLI flag,
>    for setuptools use VIRTUALENV_SETUPTOOLS=bundle environment variable or --setuptools=bundle CLI flag.
>
> By [@chrysle](https://github.com/chrysle). ([#2487](https://github.com/pypa/virtualenv/issues/2487))

so I suppose code needs writing to deal with that.  (I'm not in any hurry to get there, if anyone would like to contribute then please don't wait on me)

and there's still something screwy on windows which I haven't looked at but I imagine is likely someone else's bug that will be sorted in due course.  Probably.

dimbleby:
looks like whatever the screwy thing was on windows has now gone away, so all that is left is the virtualenv behaviour change

dimbleby:
re the virtualenv behaviour change I've taken the view that poetry should roll with it and head in the same direction of not installing setuptools and wheel in new environments, certainly on new python.

then there's a bit of a rework to make sure that they _are_ installed in the one case where poetry itself wants them

a behaviour change here is that poetry no longer automatically installs `wheel` by default alongside `pip`, this really shouldn't be necessary.  (By way of comparison, `python -m venv` does not install `wheel`).  I suppose we could continue to install `wheel` at python < 3.12, but that seems likely to cause more confusion than it solves.

radoering:
> re the virtualenv behaviour change I've taken the view that poetry should roll with it and head in the same direction of not installing setuptools and wheel in new environments, certainly on new python.

I agree.

> I suppose we could continue to install `wheel` at python < 3.12, but that seems likely to cause more confusion than it solves.

Not sure about that. I have no idea if there are users who rely on it but I suppose from a user perspective it would be more like a clean cut to keep setuptools and wheel for < 3.12 and drop it only for >= 3.12. On the other side, the current design is quite weird because we don't have a `no-wheel` config option and introducing it now does not seem very alluring so maybe just risk the hassle.

dimbleby:
I'm pretty sure that putting `setuptools` and `wheel` into the project environment shouldn't be useful for anyone in the modern world of isolated builds.  (Except those who have forgotten to declare the dependency and have been getting away with it because of this behaviour: but those people should fix their dependencies.)

In the long run it's probably desirable to lose any options or code that make a special case of those packages, but I've tried to rein in any such ambitions in this merge request...

3.12.0 is due at the start of October.  While there's obviously no obligation to be instantly ready for that, it is approaching time to make any decisions that need making.

Secrus:
My take on that would be that we should only leave `pip` in the project environment, starting from now. Since this is now a default for Python 3.12+, I don't see why we should stick to a different set of rules. I would add a note about that in the next release blog post and changelog and add deprecation/future warning to the environment creation code, that we will be removing setuptools and wheel installation in project venv for older pythons in let's say, 2 releases.

radoering:
> I would add a note about that in the next release blog post and changelog

Agreed. We should mention the difference between Python 3.12 and older versions, as well as the removal of `wheel` for older versions (if we merge the PR as is).

> and add deprecation/future warning to the environment creation code, that we will be removing setuptools and wheel installation in project venv for older pythons in let's say, 2 releases.

I don't think it's necessary to change the behavior for `setuptools`. Of course, we could, but I wouldn't spend the effort but just let the old behavior fade out together with the older Python versions.

`wheel` is already removed (as per this PR) and it would require additional effort to keep the old behavior for older Python versions. I don't think the changed behavior is a big thing, so I'd probably go with dimbleby and risk this change.

radoering (approved):

PR #7820, opened 2023-04-20
unit test type annotations

dimbleby:
continuing to chip away at the mypy errors in the unit tests

### DIFF
--- a/tests/console/commands/test_add.py
+++ b/tests/console/commands/test_add.py
@@ -4,12 +4,14 @@ import sys
 
 from pathlib import Path
 from typing import TYPE_CHECKING
+from typing import cast
 
 import pytest
 
 from poetry.core.constraints.version import Version
 from poetry.core.packages.package import Package
 
+from poetry.console.commands.installer_command import InstallerCommand
 from poetry.puzzle.exceptions import SolverProblemError
 from poetry.repositories.legacy_repository import LegacyRepository
 from tests.helpers import get_dependency
@@ -21,6 +23,7 @@ if TYPE_CHECKING:
 
     from cleo.testers.command_tester import CommandTester
     from pytest_mock import MockerFixture
+    from tomlkit import TOMLDocument
 
     from poetry.poetry import Poetry
     from poetry.utils.env import MockEnv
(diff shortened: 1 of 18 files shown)


tests/repositories/test_legacy_repository.py
```diff
@@ -477,7 +477,7 @@ def __init__(
 def test_get_200_returns_page(http: type[httpretty.httpretty]) -> None:
     repo = MockHttpRepository({"/foo/": 200}, http)
 
-    assert repo.get_page("foo")
+    _ = repo.get_page("foo")
```
Secrus, line 480: Why did you remove `assert`?

tests/helpers.py
```diff
@@ -178,6 +178,8 @@ def reset_poetry(self) -> None:
 
 
 class TestLocker(Locker):
+    __test__ = False
```
Secrus, line 181: What is this for?

tests/console/commands/test_add.py
```diff
@@ -1239,38 +1293,16 @@ def test_add_to_section_that_does_no_exist_yet(
...
 
-    content = app.poetry.file.read()["tool"]["poetry"]
+    pyproject: dict[str, Any] = app.poetry.file.read()
+    content = pyproject["tool"]["poetry"]
 
     assert "cachy" in content["group"]["dev"]["dependencies"]
     assert content["group"]["dev"]["dependencies"]["cachy"] == "^0.2.0"
 
 
-def test_add_preferes_stable_releases(
```
Secrus, line 1250: Why is this test removed?

tests/repositories/test_legacy_repository.py
dimbleby: iirc the API changed here so that this method never returns `None` any more and the assert is therefore always true

(it instead raises an exception to signal failure)

tests/helpers.py
dimbleby: else pytest thinks that this class is defining testcases

tests/console/commands/test_add.py
dimbleby: it's a copy of `test_add_prefers_stable_releases()`, but with a typo in the name

tests/helpers.py
Secrus: Could you add a short comment explaining that to avoid similar confusion in the future?

dimbleby: sure

Secrus (approved):

PR #7822, opened 2023-04-21
Add 'prefer-active-environment' setting

Roang-zero1:
This setting allows one to choose which environment should be preferred.
When this setting is true an active virtual environment will be chosen above the current poetry environment.

# Pull Request Check List

Resolves: #2888

## Tests

I've updated the configuration tests, but am unsure how to properly test with an external virtual environment, so I have not yet added any tests there.


radoering:
Preferring an activated virtual environment is already Poetry's default behavior. See https://python-poetry.org/docs/managing-environments/

If there is a bug in the detection of an activated virtual environment, I suppose this bug should be fixed instead of introducing a new setting.

Roang-zero1:
Ok then I will modify the PR accordingly.
Just to make sure I am not misunderstanding:
[Line 731:  if not in_venv or env is not None:](https://github.com/python-poetry/poetry/blob/735235a08fce654f924b1c50cd9a613abd6aab62/src/poetry/utils/env.py#L731) currently prefers the poetry environment that it selected at [Line 718](https://github.com/python-poetry/poetry/blob/735235a08fce654f924b1c50cd9a613abd6aab62/src/poetry/utils/env.py#L718).

What you are saying is that this should prefer the venv from the environment settings?

Roang-zero1:
Just for clarification, the tests and the docs are not in line:
- Docs say:  "To achieve this, it will first check if it’s currently running inside a virtual environment. If it is, it will use it directly without creating a new one."
- But the tests explicitly test for the opposite behavior. See [tests/utils/test_env.py#L607](https://github.com/python-poetry/poetry/blob/735235a08fce654f924b1c50cd9a613abd6aab62/tests/utils/test_env.py#L607) for an example

radoering:
Phew. If I understand correctly, the issue only occurs after a virtualenv env has been created by poetry so that an `envs.toml` exists. In that case, poetry does not prefer anymore an activated virtualenv but uses the environment from the `envs.toml`?

Roang-zero1:
Yes, that is correct.

We have four states:

- Virtual environment activated, but no `envs.toml`: Will use the virtual enviroment
- Virtual environment activated, and `envs.toml`: Will use the environment from `envs.toml` <- This is the behavior I want to change
  Either we introduce the setting (current PR), or we change the default behavior to always prefer the activated environment (this may be a breaking change for some workflows)
- No virtual environment, and `envs.toml` exists: Will use the environment from `envs.toml`
- No virtual environment, and `envs.toml` is empty: Will create a new environment

Reasoning behind this request:
I am using tox to run my tests and want to be able to use poetry to manage my tox virtual environment.
```ini
[testenv]
allowlist_externals=
    poetry
commands_pre =
    poetry install --no-root --only main, test --sync
commands=
    pytest <module>
````
With the current behavior, this will sync the environment from `envs.toml` to test.
This would be fine if it was the only environment, but it isn't so each test will sync to a different state.
At the moment there is no way to force the use of the active environment, so I would need to export multiple requirements and use pip to install them...

**Comment by radoering** [comment]
After some trial and error, I think I understand the reasoning for the current behavior (and I think it makes sense). Further, I don't think a setting is necessary. Nevertheless, it's probably at least a documentation issue. (PRs improving the documentation are always welcome.)

If you want to use an activated virtual environment even if there is an entry in `envs.toml` (i.e. you have called `poetry env use` before), just call `poetry env use system`.

Example (starting without an `envs.toml`):
```
$ poetry env info -p
/home/randy/.cache/pypoetry/virtualenvs/active-env-ViNcRw5I-py3.8
$ . venv/bin/activate
(venv) $ poetry env info -p
/home/randy/active_env/venv
(venv) $ poetry env use python3.9
Using virtualenv: /home/randy/.cache/pypoetry/virtualenvs/active-env-ViNcRw5I-py3.9
(venv) $ poetry env info -p
/home/randy/.cache/pypoetry/virtualenvs/active-env-ViNcRw5I-py3.9
(venv) $ poetry env use python3.8
Using virtualenv: /home/randy/.cache/pypoetry/virtualenvs/active-env-ViNcRw5I-py3.8
(venv) $ poetry env info -p
/home/randy/.cache/pypoetry/virtualenvs/active-env-ViNcRw5I-py3.8
(venv) $ poetry env use system
Deactivating virtualenv: /home/randy/.cache/pypoetry/virtualenvs/active-env-ViNcRw5I-py3.8
(venv) $ poetry env info -p
/home/randy/active_env/venv
```

Roang-zero1:
Yes that works, thanks!

PR #7833, opened 2023-04-24
Update locker.py

clicking:
This fixes "files" KeyError

# Pull Request Check List

Resolves: #issue-number-here

### DIFF
--- a/src/poetry/packages/locker.py
+++ b/src/poetry/packages/locker.py
@@ -145,7 +145,7 @@ class Locker:
                 package.files = [{"name": h, "hash": h} for h in hashes[name]]
             elif source_type in {"git", "directory", "url"}:
                 package.files = []
-            else:
+            elif "files" in metadata:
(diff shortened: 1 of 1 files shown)


clicking:
I was trying to add a new version of whl file that I generated.
The package "exceptiongroup" did not allow the installation to continue. I don't know if it's because of something in the package or because of something inside the .lock file

When editing this file I thought that somewhere else there would be an error but no. Apparently it is not necessary for the files attribute to be assigned and everything still worked correctly.
I assume that as in the previous IFs this one could also become an ifelse

dimbleby:
this shouldn't be necessary, the existing code is already good for lockfiles written by any even slightly recent poetry.

I expect that whatever state you have gotten into can be resolved by re-locking.  If that's not so, please raise a bug, providing all the details asked for by the template (in particular, a means to reproduce the problem).

PR #7845, opened 2023-04-27
more type-checking

dimbleby:
still not sure how valuable it is to be annotating the unit tests, but this is getting close to clean now.  If you continue to encourage me by merging these MRs I suspect I'll end up completing the job.

### DIFF
--- a/src/poetry/repositories/pypi_repository.py
+++ b/src/poetry/repositories/pypi_repository.py
@@ -147,7 +147,7 @@ class PyPiRepository(HTTPRepository):
 
     def _get_release_info(
         self, name: NormalizedName, version: Version
-    ) -> dict[str, str | list[str] | None]:
+    ) -> dict[str, Any]:
         from poetry.inspection.info import PackageInfo
 
         self._log(f"Getting info for {name} ({version}) from PyPI", "debug")
--- a/tests/console/commands/self/test_add_plugins.py
+++ b/tests/console/commands/self/test_add_plugins.py
@@ -1,11 +1,13 @@
 from __future__ import annotations
 
 from typing import TYPE_CHECKING
+from typing import Any
 
 import pytest
 
 from poetry.core.packages.package import Package
 
+from poetry.console.commands.add import AddCommand
 from poetry.console.commands.self.self_command import SelfCommand
 from poetry.factory import Factory
 from tests.console.commands.self.utils import get_self_command_dependencies
@@ -29,7 +31,7 @@ def assert_plugin_add_result(
     constraint: str | dict[str, str],
 ) -> None:
     assert tester.io.fetch_output() == expected
-    dependencies = get_self_command_dependencies()
+    dependencies: dict[str, Any] = get_self_command_dependencies()
 
     assert "poetry-plugin" in dependencies
     assert dependencies["poetry-plugin"] == constraint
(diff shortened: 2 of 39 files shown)


tests/puzzle/test_solver.py
```diff
@@ -2165,7 +2183,8 @@ def test_solver_does_not_raise_conflict_for_locked_conditional_dependencies(
     repo.add_package(package_a)
     repo.add_package(package_b)
 
-    solver._locked = Repository("locked", [package_a])
+    dep_package_a = DependencyPackage(dependency_a, package_a)
```
dimbleby, line 2186: the changes here are maybe the only interesting point.  I've no idea what this testcase was trying to prove, or whether it is succeeding in proving it.  I've just replaced definiteliy-broken code with not-definitely broken code: but since the testcase passes with or without this I suspect it doesn't mean much.

On the plus side this is perhaps an argument in favour of typechecking the unit tests.  Presumably this test did something useful at some point, and it has been allowed to degrade because no-one noticed that `solver._locked` simply isn't a thing.  That wouldn't have happened, if the tests were being checked.

Secrus (dismissed):

tests/console/commands/self/test_add_plugins.py
```diff
@@ -128,14 +130,11 @@ def test_add_with_git_constraint_with_extras(
...
 
-    assert_plugin_add_result(
-        tester,
-        expected,
-        {
-            "git": "https://github.com/demo/poetry-plugin.git",
-            "extras": ["foo"],
-        },
-    )
+    constraint: dict[str, Any] = {
```
Secrus, line 133: couldn't this `Any` be made into `str | list[str]`?

src/poetry/repositories/pypi_repository.py
```diff
@@ -147,7 +147,7 @@ def find_links_for_package(self, package: Package) -> list[Link]:
 
     def _get_release_info(
         self, name: NormalizedName, version: Version
-    ) -> dict[str, str | list[str] | None]:
```
Secrus, line 150: was this type that bad? I feel that `Any` is always worse choice over more concrete type

dimbleby: this one was definitely wrong iirc, there was some point at which one of the values was a list of dictionaries or something like that.

Try changing it back and see what additional warning comes up if you really want to know!

tests/console/commands/self/test_add_plugins.py
dimbleby: that leads to a small rabbit-hole where it disagrees with `assert_plugin_add_result`.

I have a few more type annotation changes coming, I'd suggest merge this as it is now and I pinky-promise to do more with it in the next (final?) round

src/poetry/repositories/pypi_repository.py
dimbleby: https://github.com/python-poetry/poetry/blob/978d435be25b799fb6d107c1fb1739eb858057c1/tests/repositories/test_pypi_repository.py#L353-L359 is the testcase that reveals the old type to be a lie

tests/console/commands/self/test_add_plugins.py
dimbleby: actually one last push and we're done...

Secrus (approved):

PR #7850, opened 2023-04-29
Expand ruff usage

Secrus:
# Pull Request Check List

Migrated `pyupgrade`, `isort`, and `pygrep-hooks` to be used via `ruff` rules. At the same time deleted `pycln` since its usage was redundant with other tools.


dimbleby:
Now seems as good a time as any to float an idea: I don't particularly like poetry's non-default isort settings.

Definitely not a big deal, having these things autoformatted is much more important than the exact details of how they are autoformatted.  But if others would also prefer simply to use the default settings - perhaps this MR is a good excuse to make the switch.

finswimmer:
I don't care much about the exact sorting of imports either. The only thing a really like to see/keep is one line between the types.

radoering:
As it seems nothing changes if I comment out `section-order`.

What about `force-single-line = true`? E.g. without this setting:

```diff
- from poetry.core.masonry.api import build_sdist
- from poetry.core.masonry.api import build_wheel
- from poetry.core.masonry.api import get_requires_for_build_sdist
- from poetry.core.masonry.api import get_requires_for_build_wheel
- from poetry.core.masonry.api import prepare_metadata_for_build_wheel
+ from poetry.core.masonry.api import (
+     build_sdist,
+     build_wheel,
+     get_requires_for_build_sdist,
+     get_requires_for_build_wheel,
+     prepare_metadata_for_build_wheel,
+ )
```

As long as the imports are short enough (which is always true for IIRC), I think I'd slightly prefer to keep this setting.

Secrus:
I actually like the current settings (and it's not like it changes frequently, so we set it once and just use it)

dimbleby:
meh, I don't want everyone wasting their time and energy on bikeshedding this!  

If y'all had gone "me too! let's do it", then let's do it: but I suspect we're going to have nearly as many preferences as voices, and no-one who really cares very much - which doesn't sound like a good reason to change anything.

pyproject.toml
```diff
@@ -156,17 +159,21 @@ extend-exclude = [
...
-force_single_line = true
-combine_as_imports = true
-lines_between_types = 1
-lines_after_imports = 2
-src_paths = ["src", "tests"]
-extend_skip = ["setup.py"]
-known_third_party = ["poetry.core"]
+[tool.ruff.isort]
+force-single-line = true
+combine-as-imports = true
```
radoering, line 164:
```suggestion
```

This doesn't do anything since we force single line, does it?

pyproject.toml
```diff
@@ -156,17 +159,21 @@ extend-exclude = [
...
+known-first-party = ["poetry"]
+known-third-party = ["poetry.core"]
+required-imports = ["from __future__ import annotations"]
+section-order = [
+    "future",
+    "standard-library",
+    "third-party",
+    "first-party",
+    "local-folder"
+]
```
radoering, line 176:
```suggestion
```

This seems to be the default order, so we don't have to configure it.

Secrus: According to the [documentation](https://beta.ruff.rs/docs/settings/#section-order) by default this value is empty.

radoering: The [documentation of isort](https://pycqa.github.io/isort/docs/configuration/custom_sections_and_ordering.html) seems to be better in this point:

> You can change the section order with sections option from the default of:
> 
> `FUTURE,STDLIB,THIRDPARTY,FIRSTPARTY,LOCALFOLDER`

I'm not good at Rust but after some trial and error and code spelunking, I'd say ruff does the same:
[ImportType Enum](https://github.com/charliermarsh/ruff/blob/64b7280eb824d0e5f9da887e82dcda53838dd38d/crates/ruff/src/rules/isort/categorize.rs#L35C1-L41)
[Default order 1](https://github.com/charliermarsh/ruff/blob/64b7280eb824d0e5f9da887e82dcda53838dd38d/crates/ruff/src/rules/isort/settings.rs#LL343C83-L343C83)
[Default order 2](https://github.com/charliermarsh/ruff/blob/64b7280eb824d0e5f9da887e82dcda53838dd38d/crates/ruff/src/rules/isort/settings.rs#L351-L353)

radoering (approved):

PR #7852, opened 2023-04-30
tests: ensure forward compatibility with python-poetry/poetry-core#402

radoering:
With python-poetry/poetry-core#402 the constraint `">= 2.8.1, == 2.8.*"` is parsed to `">=2.8.1,<2.9.dev0"` instead of `">=2.8.1,<2.9.0"`.

All three are equivalent according to PEP 440.

I added some more options (considering zero padding) to make the test more robust to possible future changes.

### DIFF
--- a/tests/utils/test_dependency_specification.py
+++ b/tests/utils/test_dependency_specification.py
@@ -99,7 +99,7 @@ if TYPE_CHECKING:
                 {
                     "name": "requests",
                     "markers": 'python_version < "2.7"',
-                    "version": ">=2.8.1,!=2.8.*",
+                    "version": ">=2.8.1,==2.8.*",
                     "extras": ["security", "tests"],
                 },
             ),
--- a/tests/utils/test_dependency_specification.py
+++ b/tests/utils/test_dependency_specification.py
@@ -20,54 +20,73 @@ if TYPE_CHECKING:
 
 
 @pytest.mark.parametrize(
-    ("requirement", "specification"),
+    ("requirement", "expected_variants"),
     [
         (
             "git+https://github.com/demo/demo.git",
-            {"git": "https://github.com/demo/demo.git", "name": "demo"},
+            ({"git": "https://github.com/demo/demo.git", "name": "demo"},),
         ),
         (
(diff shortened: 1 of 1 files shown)


tests/utils/test_dependency_specification.py
```diff
@@ -69,12 +71,38 @@
...
+                {
+                    "name": "requests",
+                    "markers": 'python_version < "2.7"',
+                    "version": ">=2.8.1,<2.9.0.dev0",
+                    "extras": ["security", "tests"],
+                },
+                {
+                    "name": "requests",
+                    "markers": 'python_version < "2.7"',
+                    "version": ">=2.8.1,!=2.8.*",
```
Secrus, line 102: I don't understand how this works. "Greater or equal 2.8.1 and not matching 2.8.*"?

tests/utils/test_dependency_specification.py
```diff
@@ -119,8 +147,13 @@ def _mock(self: Path) -> bool:
 
     mocker.patch("pathlib.Path.exists", _mock)
 
-    assert not DeepDiff(
-        RequirementsParser(artifact_cache=artifact_cache).parse(requirement),
-        specification,
-        ignore_order=True,
+    if isinstance(specification, dict):
+        specification = [specification]
```
Secrus, line 151: Could you just change the data in parametrize so we can avoid this?

Secrus (commented):
LGTM, just one question and one nitpick

tests/utils/test_dependency_specification.py
radoering: Good catch, it doesn't. 😉 It has to be `">=2.8.1,==2.8.*"`. (Right now it's just not relevant because neither version of poetry-core generates this string.) Fixed.

tests/utils/test_dependency_specification.py
radoering: No strong feelings about this. I changed it in a separate commit. (In the end we'll squash the commits anyway.)

Secrus (approved):

Issue #7819, opened 2023-04-20
CalledProcessError Immediately after install when installing dynamic versioning plugin


LukeDev42:
- **Poetry version**: 1.4.2
- **Python version**: 3.11.1
- **OS version and name**: Windows 10; 19044.2846
- **pyproject.toml**: N/A

- [ x ] I am on the [latest](https://github.com/python-poetry/poetry/releases/latest) stable Poetry version, installed using a recommended method.
- [ x ] I have searched the [issues](https://github.com/python-poetry/poetry/issues) of this repo and believe that this is not a duplicate.
- [ x ] I have consulted the [FAQ](https://python-poetry.org/docs/faq/) and [blog](https://python-poetry.org/blog/) for any relevant entries or release notes.
- [ ] If an exception occurs when executing a command, I executed it again in debug mode (`-vvv` option) and have included the output below.

## Issue
I had a problem updating my project's dependencies. It returned `CalledProcessError` when I ran `poetry install`. I tried running it again and got a ModuleNotFound error for poetry.console. I then tried a reinstall of poetry. 

I use `pipx` to manage my poetry installation so I ran:
```
pipx install poetry
poetry --version
   Poetry (version 1.4.2)
poetry self add "poetry-dynamic-versioning[plugin]"
Using version ^0.21.4 for poetry-dynamic-versioning

Updating dependencies
Resolving dependencies...

Package operations: 4 installs, 14 updates, 0 removals

  • Updating charset-normalizer (3.1.0 -> 3.0.1)
  • Updating more-itertools (9.1.0 -> 9.0.0)
  • Updating urllib3 (1.26.15 -> 1.26.14)
  • Updating zipp (3.15.0 -> 3.12.0)

  CalledProcessError

  Command '['C:\\Users\\user\\.local\\pipx\\venvs\\poetry\\Scripts\\python.exe', 'C:\\Users\\user\\.local\\pipx\\venvs\\poetry\\Lib\\site-packages\\virtualenv\\seed\\wheels\\embed\\pip-23.1-py3-none-any.whl\\pip', 'uninstall', 'charset-normalizer', '-y']' returned non-zero exit status 2.

  at C:\Program Files\Python311\Lib\subprocess.py:571 in run
       567│             # We don't call process.wait() as .__exit__ does that for us.
       568│             raise
       569│         retcode = process.poll()
       570│         if check and retcode:
    →  571│             raise CalledProcessError(retcode, process.args,
[...]

Issue #7610, opened 2023-03-06
Poetry self update fails on Windows with Access Denied, leaving poetry in a broken state


iTitus:
- **Poetry version**: Current version is 1.4.0, but it has happened to me multiple times when updating poetry 
- **Python version**: System Python is 3.10.5 
- **OS version and name**: Windows 10 
- **pyproject.toml**: N/A 

- [x] I am on the [latest](https://github.com/python-poetry/poetry/releases/latest) stable Poetry version, installed using a recommended method. (not really applicable)
- [x] I have searched the [issues](https://github.com/python-poetry/poetry/issues) of this repo and believe that this is not a duplicate. (there are similar issues, but none have a solution/activity)
- [x] I have consulted the [FAQ](https://python-poetry.org/docs/faq/) and [blog](https://python-poetry.org/blog/) for any relevant entries or release notes.
- [x] If an exception occurs when executing a command, I executed it again in debug mode (`-vvv` option) and have included the output below. (output included below, but not verbose)

## Issue

This has happened to me for multiple version updates in a row now.
Whenever upgrading poetry with `poetry self update` pip fails with an OSError and leaves Poetry in a broken state.
A reinstall fixes that.

There are two problems here:
1. Pip crashes due to an OSError: this does not happen when normally running pip or using poetry to install/update venvs, only when upgrading poetry itself
1. Poetry does not rollback any changes after pip crashes, thus leaving itself in a broken and unusable state

```
poetry self update
Updating Poetry version ...

Using version ^1.4.0 for poetry

Updating dependencies
Resolving dependencies...

Writing lock file

Package operations: 4 installs, 4 updates, 0 removals

  • Installing colorama (0.4.6)
  • Updating poetry-core (1.4.0 -> 1.5.1)
  • Installing pyproject-hooks (1.0.0)
  • Installing build (0.10.0)
  • Updating dulwich (0.20.50 -> 0.21.3)
  • Updating poetry-plugin-export (1.2.0 -> 1.3.0)
  • Installing installer (0.6.0)

  CalledProcessError

[...]

PR #7872, opened 2023-05-03
deprecate `poetry self` commands

dimbleby:
Of course I am not expecting this to be merged without some discussion!  But I think it's a discussion that's worth having, so let's create a space in which to have it.

A thing that sometimes happens on this issue tracker is putting right users who have installed poetry itself in the project under management.  Of course that goes wrong: among other things, running `poetry update` (or whatever) can cause packages to be updated while poetry itself is actively using them; if you get unlucky then poetry just won't work any more.

But so far as I can see `poetry self` commands all are built on this exact mistake: the `poetry` binary is managing its own environment.  And indeed there are reports from people who do things like `poetry self update` and see just the same sort of problem: #7610, #7819

I don't understand how this can be expected to work, and I have no idea how it can be fixed.  

What's more - perhaps this wasn't true a few years back - there are fine alternatives available, most obviously `pipx` based installations.

I also think in general that `poetry` has a larger surface area than is good for the health of the project - see the state of the issue backlog - and that `poetry self` is a relatively bug-rich part of the codebase, certainly compared to the value that it brings.  So to my mind it's a great candidate for establishing that it is possible to remove code that is more burden than help.

Opinions welcome...

### DIFF
--- a/src/poetry/console/commands/self/self_command.py
+++ b/src/poetry/console/commands/self/self_command.py
@@ -124,6 +124,10 @@ class SelfCommand(InstallerCommand):
         #
         # This method **should not** be overridden in child classes as it may have
         # unexpected consequences.
+        self.line_error(
+            "<warning>`poetry self` commands are deprecated and will be removed in a"
+            " future release</>"
+        )
 
         self.reset()
 


johnthagen:
Would this effectively deprecate the official installer as well? Since `poetry self update` is the only way to update Poetry when installed that way?

dimbleby:
The installer is perfectly capable of updating an existing poetry installation, `poetry self update` is not needed.

johnthagen:
Okay, so would the workflow be to upgrade the officially installed Poetry that a user would go back to https://python-poetry.org/docs/#installing-with-the-official-installer, copy the command for their platform, and run it again?

To me, this would move me to use `pipx` instead to use a cross-platform way to update Poetry rather than use the official installer since the update process for the official installer would require a bit more work than `poetry self update`.

Note, I'm not saying that `poetry self` should or shouldn't be deprecated (I've had _plenty_ of issues with `poetry self update` myself). Just wanted to discuss the knock on effects this could have.

(Slightly off topic) The bigger issue (not the Poetry dev's fault) is that Python really needs to bundle something like `pipx` in its official distribution to solve this bootstrapping issue for Poetry and other CLI tools. Rust does this with `cargo install` and NPM does this with `npx`, by being providing this by default.

I've brought this up to the core team before

- https://discuss.python.org/t/wanting-a-singular-packaging-tool-vision/21141
- https://discuss.python.org/t/wanting-a-singular-packaging-tool-vision/21141/24

dimbleby:
pipx is a fine solution and the way that I install poetry myself.  For users who want to do fancy things with plugins (which is not me) it would probably be a particularly good choice because `pipx inject`.

For those wanting to avoid pipx for whatever reason, running the installer both to install or upgrade seems fine.  

There is no shortage of alternatives to `poetry self update`, and deprecating it need have no implications for the status of the installer.

radoering:
Not sure if we should deprecate all `self` commands. `poetry self show plugins` for example still makes sense to me.

> And as such does that mean that if you want to install plugins then you'll need to not use the installer

If you use the installer, you can still install plugins via pip as mentioned [here](https://python-poetry.org/docs/master/plugins/#with-pip).

> and pipx becomes the default for installation type?

see #8090

amp-matthew-wallace:
>> And as such does that mean that if you want to install plugins then you'll need to not use the installer

> If you use the installer, you can still install plugins via pip as mentioned [here](https://python-poetry.org/docs/master/plugins/#with-pip).

[Poetry's current documentation](https://python-poetry.org/docs/master/plugins/#the-self-add-command) states that `poetry self add poetry-plugin` "...is the easiest way and should account for all the ways Poetry can be installed."

In my opinion, removing this simple option for installing plugins would be a step backwards in the usability of the tool. I have the same use-case as @mjam03 and also use Google Artifact Registry for private repos.

I don't have nearly enough context on the general issues with `poetry self`, so maybe this change can't really be avoided. I just think that having separate plugin install commands depending on how poetry is installed makes the tool more difficult to use.

adam-grant-hendry:
> In my opinion, removing this simple option for installing plugins would be a step backwards in the usability of the tool.

I second this

Secrus:
Ok, I am going to wear my "maintainer" hat and make a decision on that. No. While I understand the `self` commands might be buggy, we should fix them instead of deprecating/removing.

dimbleby:
fair enough - though 

> I don't understand how this can be expected to work, and I have no idea how it can be fixed.

remains true...

Issue #1853, opened 2020-01-09
Config: Disable environment prompt when using shell


MooseV2:
- [x] I have searched the [issues](https://github.com/python-poetry/poetry/issues) of this repo and believe that this is not a duplicate.
- [x] I have searched the [documentation](https://python-poetry.org/docs/) and believe that my question is not covered.

## Feature Request

Recently, a feature was added that allows `poetry shell` to modify the prompt (PS1 in some shells) to show the current environment. See #1642.

However, this conflicts with prompts that already show the environment, causing text duplication.
<img width="443" alt="poetry" src="https://user-images.githubusercontent.com/6157758/72076531-a6ff9280-32c3-11ea-8496-9f543ca7a2be.png">

A configuration option to prevent the prompt from being modified would resolve this.

### My configuration
OS: macOS Catalina 10.15.1
Shell: Fish 3.0.2
Poetry: 1.0.0

MooseV2:
Ah, upon further research, I have discovered the environment variable, `VIRTUAL_ENV_DISABLE_PROMPT`.

Because Poetry uses Virtualenv internally, exporting this with `export VIRTUAL_ENV_DISABLE_PROMPT=1` (or in Fish, `set -x VIRTUAL_ENV_DISABLE_PROMPT 1`) appears to resolve this issue.

Perhaps this is a documentation issue instead? People who are using Poetry as a turnkey solution shouldn't have to dig into the internal workings (ie virtualenv).

Otherwise, a config variable could set `VIRTUAL_ENV_DISABLE_PROMPT` before spawning the virtualenv when `poetry shell` is issued.

grisaitis:
I second this. 

I would also like a config variable to prevent the `PS1` modification. Or at least document it. 

In my situation, I use [bash-it](https://github.com/Bash-it/bash-it) to have a cute shell prompt. `poetry shell` totally vanquishes this.

Where would this setting be (@MooseV2)? in the `[tools.poetry]` section of `pyproject.toml`? `[tools.poetry.env]`?

grisaitis:
Sorry, I'm an idiot - I just needed to move bash configurations from `.bash_profile` to `.bashrc`. Dotfile drama. `poetry shell` creates a new bash shell, but it's not a "login" shell. So `.bashrc` is sourced but `.bash_profile` is not. This is unlike a new terminal window on a Mac (which _is_ a login shell).

grisaitis:
[...]

PR #7874, opened 2023-05-04
Add details to basic usage docs about setting `VIRTUAL_ENV_DISABLE_PROMPT`

savannahostrowski:
# Pull Request Check List

Resolves: #1853

I don't think that these tasks apply for documentation changes. Happy to make updates for tone, more details, links to virtualenv docs, etc. 

### DIFF
--- a/docs/basic-usage.md
+++ b/docs/basic-usage.md
@@ -123,9 +123,6 @@ To take advantage of this, simply activate a virtual environment using your pref
 any Poetry commands that expect to manipulate an environment.
 {{% /note %}}
 
-If you'd like to prevent `poetry shell` from modifying your shell prompt on virtual environment activation, you should
-set `VIRTUAL_ENV_DISABLE_PROMPT=1` as an environment variable before running the command.
-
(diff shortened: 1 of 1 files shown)


radoering:
IMO, we should move this either into https://python-poetry.org/docs/basic-usage/#activating-the-virtual-environment or https://python-poetry.org/docs/cli/#shell. Currently, it's in a section that does not even mention `poetry shell` and IIUC this hint is only relevant for `poetry shell`.

savannahostrowski:
Ah, yeah. You're right - I've gone ahead and moved it lower down in the activating the virtual env section of the basic usage doc. It seemed to fit a little more naturally there.

radoering (changes requested):
Now, the paragraph is in between "manually activating the virtualenv" and "deactivating a manually activated virtualenv". IMO, it should be moved up one paragraph (before manually activating the virtualenv).

savannahostrowski:
Sure thing! Done

radoering (approved):

Issue #7883, opened 2023-05-05
`poetry shell` does not send a newline to fish


NiklasKappel:
- **Poetry version**: Poetry (version 1.4.2)
- **Python version**: 3.11.3
- **OS version and name**: Fedora 38 Silverblue
- **pyproject.toml**: https://github.com/python-poetry/poetry/blob/master/pyproject.toml

- [x] I am on the [latest](https://github.com/python-poetry/poetry/releases/latest) stable Poetry version, installed using a recommended method.
- [x] I have searched the [issues](https://github.com/python-poetry/poetry/issues) of this repo and believe that this is not a duplicate.
- [x] I have consulted the [FAQ](https://python-poetry.org/docs/faq/) and [blog](https://python-poetry.org/blog/) for any relevant entries or release notes.
- [x] If an exception occurs when executing a command, I executed it again in debug mode (`-vvv` option) and have included the output below.

## Issue

Steps to reproduce:
```
fish
cd poetry/
poetry shell
```
Observed behavior ("|" represents cursor position):
```
[niklas@toolbox poetry]$ poetry shell
Spawning shell within /var/home/niklas/.cache/pypoetry/virtualenvs/poetry-Xjwftcq8-py3.11
[niklas@toolbox poetry]$ source /var/home/niklas/.cache/pypoetry/virtualenvs/poetry-Xjwftcq8-py3.11/bin/activate.fish|
```
Expected behavior:
```
[niklas@toolbox poetry]$ poetry shell
Spawning shell within /var/home/niklas/.cache/pypoetry/virtualenvs/poetry-Xjwftcq8-py3.11
[niklas@toolbox poetry]$ source /var/home/niklas/.cache/pypoetry/virtualenvs/poetry-Xjwftcq8-py3.11/bin/activate.fish
(poetry-py3.11) [niklas@toolbox poetry]$ |
```
In fish, the source command is not executed automatically when running `poetry shell`. The reason is that fish does not count `os.linesep` as a return/enter when the `sendline` method of pexpect is used. Compare [here](https://github.com/pexpect/pexpect/issues/755#issuecomment-1536157035).

PR #7884, opened 2023-05-05
send \r to fish when using poetry shell to execute source command

NiklasKappel:
# Pull Request Check List

Resolves: #7883

The fix to issue #7883 is to send "\r" explicitly. Compare [here](https://github.com/pexpect/pexpect/issues/755#issuecomment-1536157035). I don't think a documentation update is necessary. I did not add a test, because the change is small and I could not find an existing test for the poetry shell module.

### DIFF
--- a/src/poetry/utils/shell.py
+++ b/src/poetry/utils/shell.py
@@ -107,12 +107,13 @@ class Shell:
             if self._name == "zsh":
                 # Under ZSH the source command should be invoked in zsh's bash emulator
                 c.sendline(f"emulate bash -c '. {shlex.quote(str(activate_path))}'")
-        elif self._name == "fish":
-            # Under fish "\r" should be sent explicitly
-            c.sendline(f"source {shlex.quote(str(activate_path))}\r")
         else:
+            cmd = f"{self._get_source_command()} {shlex.quote(str(activate_path))}"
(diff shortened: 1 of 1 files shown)


src/poetry/utils/shell.py
```diff
@@ -107,6 +107,9 @@ def activate(self, env: VirtualEnv) -> int | None:
             if self._name == "zsh":
                 # Under ZSH the source command should be invoked in zsh's bash emulator
                 c.sendline(f"emulate bash -c '. {shlex.quote(str(activate_path))}'")
+        elif self._name == "fish":
+            # Under fish "\r" should be sent explicitly
+            c.sendline(f"source {shlex.quote(str(activate_path))}\r")
         else:
             c.sendline(
                 f"{self._get_source_command()} {shlex.quote(str(activate_path))}"
```
radoering, line 115:
```suggestion
        else:
            cmd = f"{self._get_source_command()} {shlex.quote(str(activate_path))}"
            if self._name == "fish":
                # Under fish "\r" should be sent explicitly
                cmd += "\r"
            c.sendline(
                cmd
```

and revert the change in `_get_source_command()`

radoering (approved):

PR #7888, opened 2023-05-07
tidy pyproject.toml

danieleades:
teeny tiny nits in pyproject.toml

- alphabetises some config
- makes use of newlines between sections consistent
- consistently pads inline comments
- flattens some short lists into one-liners
- adds trailing commas
- makes indentation consistent

the flattening of short lists might not be desirable, depending on your tastes (i let an autoformatter make that call), but i think the rest is probably not too controversial


Secrus:
Please revert the changes to lists with more than one item. Otherwise, LGTM.

Secrus (approved):

PR #7896, opened 2023-05-09
feat: run tasks defined ini [tool.poetry.tasks]

finswimmer:
Example `pyproject.toml`:

```toml
[tool.poetry]
name = "tasks"
version = "0.1.0"
description = ""
authors = ["finswimmer <finswimmer77@gmail.com>"]
readme = "README.md"

[tool.poetry.tasks]
version = "python --version"
test = "pytest tests"

[tool.poetry.dependencies]
python = "^3.11"
pytest = "^7.3.1"


[build-system]
requires = ["poetry-core"]
```

Example usage:

```bash
poetry run version
``` 

```bash
poetry run test
```


dimbleby:
Acknowledging that clearly this is a thing that you think would be useful, and also that there have been feature requests for this sort of thing so you are not alone - I would want to discourage poetry from going in this direction.

It seems to me that poetry already has more features (and bugs) than maintainers and contributors can keep up with - see the state of this issue tracker for evidence.  Adding "be a task-runner" to the things that poetry does is likely to make that even more true, and dilute the available resource still further.

Being a task-runner is both no small matter, and already well-served by other solutions: poethepoet is the obvious example in the poetry ecosystem (lots of non-poetry based approaches are also possible).  Duplicating the work of that plugin seems to me both to be quite a lot of work, and not necessary - it already exists!

Of course people will work on the things that interest them.  When I occasionally wish for a better poetry project, "more features" is not what comes to mind.  But, that's just one person's opinion...

radoering:
I completely agree with dimbleby, but I may be in the minority among maintainers.

clintonroy:
I certainly would use it, and would appreciate it being built in. Maybe in time there will be something like this upstream, but..that could be a long time in waiting.

dimbleby:
> Maybe in time there will be something like this upstream

what do you mean by "upstream"?

Did you try `poethepoet`?  why not, or what's wrong with it?  

I've never used it myself so maybe there are good reasons to avoid it.  But it sounds as though you are saying "I would use a solution if it existed"; while so far as I know a solution _does_ exist!

Issue #7588, opened 2023-03-01
`poetry update --lock` prints operations that don't run.


hwmq:
- **Poetry version**: 1.4.0
- **Python version**: 3.10.6
- **OS version and name**: Ubuntu 22.04
- **pyproject.toml**: N/A

- [x] I am on the [latest](https://github.com/python-poetry/poetry/releases/latest) stable Poetry version, installed using a recommended method.
- [x] I have searched the [issues](https://github.com/python-poetry/poetry/issues) of this repo and believe that this is not a duplicate.
- [x] I have consulted the [FAQ](https://python-poetry.org/docs/faq/) and [blog](https://python-poetry.org/blog/) for any relevant entries or release notes.
- [x] If an exception occurs when executing a command, I executed it again in debug mode (`-vvv` option) and have included the output below.

## Issue
`poetry update --lock` prints operations that don't run. The following gives the gist...
```
$ poetry update --lock mypackage
Updating dependencies
Resolving dependencies... (0.9s)

Writing lock file
  • Installing a_different_package (0.0.0): Pending...
  • Updating some_other_package (0.0.0 -> 1.0.0): Pending...
# ... and so on ...
$ # and back to my shell.
```
To be clear, `some_other_package` is not updated in the lock file.

Essentially, the printed operations correspond to what *would* happen in the poetry env if `--lock` were not specified. However since `--lock` is specified, it doesn't make sense to show this output, because these things do not happen.

We can see this here (tag 1.4.0)
https://github.com/python-poetry/poetry/blob/0e72a55c43a993ec0258facec23416c9212964ba/src/poetry/console/commands/update.py#L54

Poetry actually ends up reaching this line, which doesn't end up doing anything because `self._execute_operations is False`, but does print the output like above.
https://github.com/python-poetry/poetry/blob/0e72a55c43a993ec0258facec23416c9212964ba/src/poetry/installation/executor.py#L187

I wonder if when `--lock` is specified, should we also call `self.installer.lock()`? That way, the return in this block will execute. Currently it does not because `self._lock is False`
https://github.com/python-poetry/poetry/blob/0e72a55c43a993ec0258facec23416c9212964ba/src/poetry/installation/installer.py#L294-L299

`add` already does this.
https://github.com/python-poetry/poetry/blob/0e72a55c43a993ec0258facec23416c9212964ba/src/poetry/console/commands/add.py#L257-L258

hwmq:
Proposed fix.
```
[...]

PR #7915, opened 2023-05-13
fix: do not print unwanted operations on --lock

ralbertazzi:
Resolves: #7588

### DIFF
--- a/src/poetry/installation/installer.py
+++ b/src/poetry/installation/installer.py
@@ -53,7 +53,6 @@ class Installer:
         self._requires_synchronization = False
         self._update = False
         self._verbose = False
-        self._write_lock = True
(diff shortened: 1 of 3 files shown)


src/poetry/installation/installer.py
```diff
@@ -53,7 +53,6 @@ def __init__(
         self._requires_synchronization = False
         self._update = False
         self._verbose = False
-        self._write_lock = True
```
ralbertazzi, line 56: Another unnecessary state variable goes away, as basically `_write_lock = not dry_run`

radoering: Nice. 👍

radoering (approved):
LGTM. Just added a test.

PR #7916, opened 2023-05-13
perf: use cached file for wheel inspection

ralbertazzi:
Another PR that goes in the way of improving the Poetry experience when dealing with huge wheels (did anyone say PyTorch? 👀 #6409 )

I'm hosting huge wheels on a repository that is not PEP 658 compliant. What happens is that Poetry will rightfully want to [get package metadata from the first wheel](https://github.com/python-poetry/poetry/blob/3bed8d67cb16f6e016e8cac0905bb7a4677868c6/src/poetry/repositories/http_repository.py#L210). The package is correctly cached on filesystem in the repository cache (the whole `Authenticator` + `FileCache` flow) so that future downloads are not needed again.

However, when a wheel is already cached on filesystem the inspection code performs an unnecessary data transfer operation from the cache location to a temporary directory. As the speed of this operation is controlled by the low `chunk_size: int = 1024` value, we spend a gigantic amount of time waiting for something that we already have. Console logs also incorrectly show that a package is being downloaded, while in reality no network operation is happening.

This PR fixes this behaviour by accessing the cached file directly, and relying on the download to temporary file only if the file is not cached (or the cache is disabled). In order to do so it exploits the public `url_to_file_path` utility from `cachecontrol`. I thought it was best leaving any direct `cachecontrol` dependency inside the `Authenticator`, which is the reason why I'm touching two classes.

### DIFF
--- a/src/poetry/repositories/http_repository.py
+++ b/src/poetry/repositories/http_repository.py
@@ -79,10 +79,13 @@ class HTTPRepository(CachedRepository):
         from poetry.inspection.info import PackageInfo
 
         wheel_name = urllib.parse.urlparse(url).path.rsplit("/")[-1]
+
+        filepath = self._authenticator.get_cached_file_for_url(url)
+        if filepath:
+            return PackageInfo.from_wheel(filepath)
+
         self._log(f"Downloading wheel: {wheel_name}", level="debug")
-
         filename = os.path.basename(wheel_name)
-
         with temporary_directory() as temp_dir:
             filepath = Path(temp_dir) / filename
             self._download(url, filepath)
@@ -95,10 +98,12 @@ class HTTPRepository(CachedRepository):
         sdist_name = urllib.parse.urlparse(url).path
         sdist_name_log = sdist_name.rsplit("/")[-1]
 
+        filepath = self._authenticator.get_cached_file_for_url(url)
+        if filepath:
+            return PackageInfo.from_wheel(filepath)
+
         self._log(f"Downloading sdist: {sdist_name_log}", level="debug")
-
         filename = os.path.basename(sdist_name)
-
         with temporary_directory() as temp_dir:
             filepath = Path(temp_dir) / filename
             self._download(url, filepath)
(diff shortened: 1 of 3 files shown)


src/poetry/utils/authenticator.py
```diff
@@ -463,6 +466,13 @@ def _get_certs_for_url(self, url: str) -> RepositoryCertificateConfig:
             return selected.certs(config=self._config)
         return RepositoryCertificateConfig()
 
+    def get_cached_file_for_url(self, url: str) -> Path | None:
+        if self._disable_cache:
```
dimbleby, line 470: if you instead ask `if self._cache_control is None:` - which is also done elsewhere in this code - then the later cast will become unnecessary

(generally prefer to avoid casting.  If and when someone submits an MR that makes the cast incorrect, mypy has no way to say so.  So if there was not a simple way around this, would prefer to assert that self._cache_control is not None)

ralbertazzi: Hadn't realized that `disable_cache` implies `self._cache_control is None`, although that's pretty obvious now I must admit :D thanks!

src/poetry/repositories/http_repository.py
```diff
@@ -79,10 +79,13 @@ def _get_info_from_wheel(self, url: str) -> PackageInfo:
         from poetry.inspection.info import PackageInfo
 
         wheel_name = urllib.parse.urlparse(url).path.rsplit("/")[-1]
-        self._log(f"Downloading wheel: {wheel_name}", level="debug")
 
         filename = os.path.basename(wheel_name)
+        filepath = self._authenticator.get_cached_file_for_url(url)
+        if filepath:
+            return PackageInfo.from_wheel(filepath)
```
radoering, line 86: We should apply the same optimization to `_get_info_from_sdist()`, probably even everywhere where `_download()` is used (i.e. also in `_links_to_data()`), shouldn't we?

ralbertazzi: Thanks for the suggestion, I should have surely detected that myself 😅

I took the liberty to refactor a bit the code to avoid lots of duplications, which ended up in a nice context manager and code easier to follow IMO. Looking forward to see this performance improvement in... 1.5.0? 👀

ralbertazzi: Further exploring the `PackageInfo.from_*` helper methods it looks to me like we could just invoke `PackageInfo.from_path` for both wheel and sdist (`DirectOrigin` already does that), thus removing the need of having two separate `_get_info_from_wheel` and `_get_info_from_sdist`. I'll let you decide on this

radoering: `PackageInfo.from_wheel()` returns an empty `PackageInfo` if there is a `ValueError`, `PackageInfo.from_path()` will probably not. I don't know what the implications of that are. Thus, maybe keep it as is for now...

radoering (approved):

PR #7917, opened 2023-05-13
feat: add --lock option to remove command

ralbertazzi:
# Pull Request Check List

Resolves: #3971

Needs also https://github.com/python-poetry/poetry/pull/7915 to avoid the logging bug

### DIFF
--- a/src/poetry/console/commands/remove.py
+++ b/src/poetry/console/commands/remove.py
@@ -35,6 +35,7 @@ class RemoveCommand(InstallerCommand):
                 "(implicitly enables --verbose)."
             ),
         ),
+        option("lock", None, "Do not perform operations (only update the lockfile)."),
     ]
 
     help = """The <info>remove</info> command removes a package from the current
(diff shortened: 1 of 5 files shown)


radoering:
We should add a test that checks:

- `remove` without `--lock` removes the dependency
- `remove --lock` does not remove the dependency

src/poetry/console/commands/remove.py
```diff
@@ -108,15 +109,13 @@ def handle(self) -> int:
             )
 
         # Refresh the locker
-        self.poetry.set_locker(
-            self.poetry.locker.__class__(self.poetry.locker.lock, poetry_content)
```
ralbertazzi, line 112: There you go with the tests :) In order to make them work I had to work around this line that was effectively resetting the mocked attributes of the `TestLocker`

ralbertazzi: If you like it I will apply it to the `AddCommand` too

radoering: Looks better to me than before.

radoering:
Now, we only have to mention the option in the [docs](https://python-poetry.org/docs/cli/#options-5).

radoering (approved):

Issue #7228, opened 2022-12-20
nushell: `poetry shell` does not activate virtualenv


dead10ck:
- **Poetry version**: Poetry (version 1.3.1)
- **Python version**:
  ```
  ❯ : poetry debug info
  Poetry
  Version: 1.3.1
  Python:  3.11.0

  Virtualenv
  Python:         3.11.0
  Implementation: CPython
  Path:           /home/dead10ck/.cache/pypoetry/virtualenvs/mastodon-ansible-SPRg-5UF-py3.11
  Executable:     /home/dead10ck/.cache/pypoetry/virtualenvs/mastodon-ansible-SPRg-5UF-py3.11/bin/python
  Valid:          True

  System
  Platform:   linux
  OS:         posix
  Python:     3.11.0
  Path:       /usr
  Executable: /usr/bin/python3.11
  ```
- **OS version and name**: Fedora Workstation 37
- **pyproject.toml**:
  ```toml
  [tool.poetry]
  name = "mastodon-ansible"
  version = "1.0.0"
  description = ""
  authors = ["Skyler Hawthorne"]

  [tool.poetry.dependencies]
  python = "^3.10"
  ansible = "^6.1.0"
  boto3 = "^1.24.42"

  [tool.poetry.dev-dependencies]

  [build-system]
  requires = ["poetry-core>=1.0.0"]
  build-backend = "poetry.core.masonry.api"
  ```

- [x] I am on the [latest](https://github.com/python-poetry/poetry/releases/latest) stable Poetry version, installed using a recommended method.
- [x] I have searched the [issues](https://github.com/python-poetry/poetry/issues) of this repo and believe that this is not a duplicate.
- [x] I have consulted the [FAQ](https://python-poetry.org/docs/faq/) and [blog](https://python-poetry.org/blog/) for any relevant entries or release notes.
- [x] If an exception occurs when executing a command, I executed it again in debug mode (`-vvv` option) and have included the output below.

## Issue

After doing `poetry install`, when running `poetry shell`, a new shell launches, but it does not activate the overlay of the virtualenv. e.g.

```
mastodon-ansible/ansible on  main [!] is 📦 v1.0.0 via 🐍 v3.11.0 on ☁  strln-prod
❯ : poetry shell
Spawning shell within /home/dead10ck/.cache/pypoetry/virtualenvs/mastodon-ansible-SPRg-5UF-py3.11

[...]

PR #7919, opened 2023-05-14
fix: fix `poetry shell` for nushell

miikka:
Here are two fixes to `poetry shell` under nushell:

1. the virtualenv activation command was not actually being sent. The commit 28d0175 accidentally removed the sendline command for nu, and nu also requires an explicit `\r` like fish.
2. the correct activation command is `overlay use`, not `source`. For reference, see the header comment in virtualenv's [`activate.nu`](https://github.com/pypa/virtualenv/blob/64bec9be8d412d7e7439987f565f7070e66cdf5a/src/virtualenv/activation/nushell/activate.nu#L2).

Tested manually with nushell 0.78.0. There do not seem to be any unit tests for the shell activation logic, so I didn't add any new ones.

# Pull Request Check List

Resolves: #7228

### DIFF
--- a/src/poetry/utils/shell.py
+++ b/src/poetry/utils/shell.py
@@ -104,13 +104,14 @@ class Shell:
 
         if self._name in ["zsh", "nu"]:
             c.setecho(False)
-            if self._name == "zsh":
-                # Under ZSH the source command should be invoked in zsh's bash emulator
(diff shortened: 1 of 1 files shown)


radoering (approved):

dead10ck:
Just want to say thank you! This has been bugging me for a little while. I appreciate your fix!

Issue #7309, opened 2023-01-06
`poetry add --lock` updates all dependencies


jvenant:
- [x] I have searched the [issues](https://github.com/python-poetry/poetry/issues) of this repo and believe that this is not a duplicate.
- [x] I have searched the [FAQ](https://python-poetry.org/docs/faq/) and general [documentation](https://python-poetry.org/docs/) and believe that my question is not already covered.

## Feature Request

Hi,
In my current workflow, I install my python module in a docker image. I copy the project sources (with pyproject.toml poetry.lock) in the image and do a `poetry install`. it's simple and it works perfectly
But sometime, I'd like to test just one module upgrade 
So In a job preceding the build, I'd like to update the pyproject and lock files with just this module to a specific version. And eventually its own dependencies if the current constraints doesn't match.
I didn't find any commands and/or options to do that :
* `poetry update <module>` only get the latest version
* `poetry add <module@version>` also do the install
* `poetry add <module@version> --lock` do a full lock update

What I'm looking for is something like a `poetry add <module@version> --lock --no-update`
But this add command option doesn't currently exist.

I made my own plugin duplicating the add command and adding a `--no-update` option to add. It's very simple

I could do a PR with the change.
But I'm actually not sure if it is a feature you want to have
Maybe it is some kind of "anti-pattern" or maybe there is an other approach to achieve what I try to do

finswimmer:
Hey @jvenant,

`poetry add` already only updates existing dependencies if it's necessary, because the current version would be in conflict with the requirements of the new dependencies.

If you believe this is not the case, please provide a full reproducible example.

fin swimmer

jvenant:
Sorry, I'm not sure I'm using the good terms then.
To be very clear, what I'm talking about is this :
https://github.com/python-poetry/poetry/blob/ef89e90fc1305e07a62e9e715c91a66c8d7425f7/src/poetry/console/commands/add.py#L256
I sometime need to set `self.installer.update` to `False`

radoering:
It's not a full update due to the whitelist:

https://github.com/python-poetry/poetry/blob/ef89e90fc1305e07a62e9e715c91a66c8d7425f7/src/poetry/console/commands/add.py#L260

An example (pyproject.toml/poetry.lock) where your change matters would be really helpful.

jvenant:
sorry for the answer delay 

I was writing my answer and the walk-through to reproduce the behavior
[...]

PR #7920, opened 2023-05-14
fix: poetry add --lock does not perform a full update

ralbertazzi:
# Pull Request Check List

Resolves: #7309 

~It builds on top of https://github.com/python-poetry/poetry/pull/7917 . Once merged I'll rebase accordingly.~ ✅ 


tests/fixtures/up_to_date_lock/pyproject.toml
```diff
@@ -6,7 +6,7 @@ authors = ["Poetry Developer <developer@python-poetry.org>"]
 
 [tool.poetry.dependencies]
 python = "^3.8"
-docker = "4.3.1"
+docker = ">=4.3.1"
```
ralbertazzi, line 9: This was necessary for `test_add_does_not_update_locked_dependencies` to let the solver try to (wrongly) update from `4.3.1` to `4.3.2` when `--lock` is specified. I verified that the test was failing with the previous implementation and passes now

src/poetry/console/commands/add.py
```diff
@@ -266,8 +264,7 @@ def handle(self) -> int:
         self.installer.dry_run(self.option("dry-run"))
         self.installer.verbose(self.io.is_verbose())
         self.installer.update(True)
-        if self.option("lock"):
-            self.installer.lock()
```
ralbertazzi, line 270: In another PR I might remove `Installer.lock()` entirely as it's quite confusing. Its behaviour is in fact not in sync with the `--lock` options of the Poetry commands

radoering: Yeah, quite confusing. IIRC, it's only set for `poetry lock`.

tests/console/commands/test_add.py
```diff
@@ -1506,3 +1511,29 @@ def test_add_extras_only_accepts_one_package(
...
+def test_add_does_not_update_locked_dependencies(
+    repo: TestRepository,
+    poetry_with_up_to_date_lockfile: Poetry,
+    tester: CommandTester,
+    command_tester_factory: CommandTesterFactory,
+    command: str,
+) -> None:
+    tester = command_tester_factory("add", poetry=poetry_with_up_to_date_lockfile)
+    docker_locked = get_package("docker", "4.3.1")
+    dockr_new = get_package("docker", "4.3.2")
```
radoering, line 1526:
```suggestion
    docker_new = get_package("docker", "4.3.2")
```

src/poetry/console/commands/add.py
ralbertazzi: It is now ;)

tests/console/commands/test_add.py
```diff
@@ -1506,3 +1511,29 @@ def test_add_extras_only_accepts_one_package(
             str(e.value)
             == "You can only specify one package when using the --extras option"
         )
+
+
+@pytest.mark.parametrize("command", ["foo", "foo --lock"])
```
radoering, line 1516:
```suggestion
@pytest.mark.parametrize("command", ["foo", "foo --lock"])
@pytest.mark.parametrize(("locked", "expected_docker"), [(True, "4.3.1"), (False, "4.3.2")])
```

Thinking about the `poetry.locker.locked(True)`, I just came up with an idea to make the test more robust.

ralbertazzi: Great suggestion! I just added it ;)

radoering (approved):

PR #7922, opened 2023-05-14
Poetry core update

dimbleby:
updating poetry-core and making a series of tidy-ups, mostly as unblocked by that update

probably best reviewed commit by commit

### DIFF
--- a/src/poetry/repositories/installed_repository.py
+++ b/src/poetry/repositories/installed_repository.py
@@ -247,9 +247,7 @@ class InstalledRepository(Repository):
                 continue
 
             for distribution in sorted(
-                metadata.distributions(  # type: ignore[no-untyped-call]
-                    path=[entry],
-                ),
+                metadata.distributions(path=[entry]),
                 key=lambda d: str(d._path),  # type: ignore[attr-defined]
             ):
                 path = Path(str(distribution._path))  # type: ignore[attr-defined]
--- a/src/poetry/toml/file.py
+++ b/src/poetry/toml/file.py
@@ -1,5 +1,7 @@
 from __future__ import annotations
 
+import warnings
+
 from typing import TYPE_CHECKING
 from typing import Any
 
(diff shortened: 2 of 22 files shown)


src/poetry/repositories/installed_repository.py
```diff
@@ -247,7 +247,7 @@ def load(cls, env: Env, with_dependencies: bool = False) -> InstalledRepository:
                 continue
 
             for distribution in sorted(
-                metadata.distributions(  # type: ignore[no-untyped-call]
+                metadata.distributions(
```
radoering, line 250: most of them now probably fit into a single line

src/poetry/toml/file.py
```diff
@@ -34,8 +33,5 @@ def read(self) -> TOMLDocument:
         except (ValueError, TOMLKitError) as e:
             raise TOMLError(f"Invalid TOML file {self.path.as_posix()}: {e}")
 
-    def __getattr__(self, item: str) -> Any:
```
radoering, line 37: I suppose this will break some plugins. I'd like to deprecate it first.

dimbleby: this was previously available via the poetry-core object, and you've just released poetry-core 1.6.0 on which this no longer exists...

ie any plugins that were relying on this were going to have to switch from the `poetry.core.toml.file` to the `poetry.toml.file` anyway, IMO they might just as well make both changes at the same time

(probably not worth much of an argument, if you _really_ want this back but deprecated then say so)

radoering: It's not the import, I'm concerned about. It's `PyProjectTOML.file`. Correct me if I'm wrong, but plugins relying on `PyProjectTOML.file.parent` will not have to change anything if we keep `__getattr__()`.

dimbleby: my point is they already will have had to change the import, and in that context it's not too objectionable also to have them conform to a very slightly changed API.  

The code that you are asking to be deprecated is in some sense brand new - it hasn't even been released yet - so it's a little unexpected to be deprecated immediately.

But like I say I don't care much to argue about it

radoering: You may not have to change the import if there is no import. Let's just look at the [docs](https://python-poetry.org/docs/plugins/#generic-plugins). You have an instance of the `Poetry` class in your hand and just could do `poetry.pyproject.file.parent`. I suppose many users will not import `PyProjectTOML` or `TOMLFile` but just use the instance they get from the Poetry object.

ralbertazzi:
It probably didn't need to be said, but I'll say it anyways since this PR is not included among the requirements for 1.5: please make it part of the release 😄 Since this is the first release that, among performance improvements and better ways to define sources, should robustly support PyTorch, it would be a bit not to include https://github.com/python-poetry/poetry-core/pull/579 (which is part of `poetry-core` 1.6) in the release

radoering (approved):

PR #7924, opened 2023-05-14
Rewrite HTML parsers and drop html5lib

Secrus:
# Pull Request Check List

This PR implements custom parsers for HTML contents and drops `html5lib`, an old and barely used dependency, which has not really been maintained for some time.

### DIFF
--- a/src/poetry/mixology/version_solver.py
+++ b/src/poetry/mixology/version_solver.py
@@ -56,12 +56,21 @@ class DependencyCache:
         )
 
         packages = self.cache.get(key)
-        if packages is None:
-            packages = self.provider.search_for(dependency)
-        else:
+
+        if packages:
             packages = [
                 p for p in packages if dependency.constraint.allows(p.package.version)
             ]
+
+        # provider.search_for() normally does not include pre-release packages
+        # (unless requested), but will include them if there are no other
+        # eligible package versions for a version constraint.
+        #
+        # Therefore, if the eligible versions have been filtered down to
+        # nothing, we need to call provider.search_for() again as it may return
+        # additional results this time.
+        if not packages:
+            packages = self.provider.search_for(dependency)
 
         self.cache[key] = packages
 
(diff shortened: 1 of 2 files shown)


tests/repositories/parsers/test_pypi_search_parser.py
```diff
@@ -0,0 +1,131 @@
...
+
+from pathlib import Path
+
+import pytest
+
+from poetry.repositories.parsers.pypi_search_parser import Result
+from poetry.repositories.parsers.pypi_search_parser import SearchResultParser
+
+
+FIXTURES_DIRECTORY = Path(__file__).parent.parent / "fixtures" / "pypi.org" / "search"
```
ralbertazzi, line 11: You could use `fixture_dir: FixtureDirGetter` to achieve the same

ralbertazzi (dismissed):

tests/repositories/parsers/test_pypi_search_parser.py
Secrus: Can't really, cause this goes 2 `.parent` up, and `fixture_dir` goes only 1 `.parent` up.

radoering: Ah, these are not `tests/fixtures` but `tests/repositories/fixtures`. Then, it's just that...

src/poetry/repositories/parsers/html_page_parser.py
```diff
@@ -0,0 +1,18 @@
...
+    def __init__(self) -> None:
+        super().__init__()
+        self.base_url: str | None = None
+        self.anchors: list[dict[str, str | None]] = []
+
+    def handle_starttag(self, tag: str, attrs: list[tuple[str, str | None]]) -> None:
+        if tag == "base" and self.base_url is None:
+            base_url = dict(attrs).get("href")
+            if base_url is not None:
+                self.base_url = base_url
```
radoering, line 16: What's the `base_url` about? Is it used anywhere? Can we remove this code block and the attribute?

Secrus: It's for compatibility with HTML specs. I've added the handling for it and the proper tests.

radoering (approved):

PR #7932, opened 2023-05-16
fix: subprocess doesn't allow Path on windows, 3.7

branchvincent:
# Pull Request Check List

Due to https://bugs.python.org/issue31961, `subprocess.run` on Python 3.7 + Windows does not accept a `Path` as an argument.

Discovered from https://github.com/python-poetry/install.python-poetry.org/pull/90, where our installer smoke test fails: [`argument of type 'WindowsPath' is not iterable`](https://github.com/python-poetry/install.python-poetry.org/actions/runs/4973814297/jobs/8899895899?pr=90#step:8:111). Our existing tests here didn't catch this since `subprocess.run` is mocked

### DIFF
--- a/src/poetry/utils/env.py
+++ b/src/poetry/utils/env.py
@@ -544,7 +544,7 @@ class EnvManager:
             if executable:
                 python_patch = decode(
                     subprocess.check_output(
-                        [executable, "-c", GET_PYTHON_VERSION_ONELINER],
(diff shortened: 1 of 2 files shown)


Secrus:
@dimbleby could you include dropping these changes in your "drop python 3.7" PRs?

Secrus (approved):

PR #7938, opened 2023-05-17
feat: skip python-incompatible wheels inspection and locking

ralbertazzi:
# Pull Request Check List

Resolves: #7935

- [ ] Added **tests** for changed code.
- [ ] Updated **documentation** for changed code.
- [ ] Cache is invalidated on `poetry lock` and `poetry lock --no-update`
- [ ] Applies to PyPI too

Would love some preliminary feedback on this 🙏

PR #7939, opened 2023-05-17
feat: do not expose cache_version outside of CachedRepository

ralbertazzi:
# Pull Request Check List

I believe there's a current leak of abstraction, where the `cached_version` used in `CachedRepository` is spread among multiple files in the codebase. The assumption behind it is that the cache version should be associated to a `PackageInfo`, while I believe that it should be better associated with the generic data stored inside the CachedRepository, regardless of what that data is.

I needed it for https://github.com/python-poetry/poetry/pull/7938 as I'd like to store the python constraint inside the cache file (somewhat I prefer it compared to making the python constraint part of the file name, which is less robust due to special characters and whatnot). Still, I believe this should be merged regardless of the destiny of the other PR.


ralbertazzi:
I realize caching needs to get smarter as the cache can be used by projects with diverse Python constraint 🤔

src/poetry/inspection/info.py
```diff
@@ -87,23 +86,17 @@ def __init__(
...
         self.name = other.name or self.name
         self.version = other.version or self.version
         self.summary = other.summary or self.summary
         self.requires_dist = other.requires_dist or self.requires_dist
         self.requires_python = other.requires_python or self.requires_python
         self.files = other.files or self.files
-        self._cache_version = other.cache_version or self._cache_version
         return self
 
     def asdict(self) -> dict[str, Any]:
```
Secrus, line 102: since we are now in Python 3.8 era, could you change this typing to use `TypedDict`? you could then eliminate `Any` in `CachedRepository._cache_func()`

Secrus:
Stalled

PR #7953, opened 2023-05-19
fix: fix exception on invalid url dependency for current env

ralbertazzi:
The `Executor` first downloads and caches the original artifact, then tries to fetch a higher priority cached artifact (i.e. a wheel instead of a sdist) to speed up the installation, falling back on the previously downloaded original artifact. Therefore it was assumed that whatever cached artifact gets returned, it will not be None.

The assumption however is wrong since the cached artifact search is filtered for the given environment, i.e. if even the original artifact is invalid for the current environment than `artifact` will actually be None, raising a quite mysterious exception. This can happen for instance in this case:
```toml
[tool.poetry.dependencies]
python = "~3.9"

# Requesting a Python 3.10 wheel on a Python 3.9 environment
torch = { url = "https://download.pytorch.org/whl/torch-1.13.1-cp310-cp310-manylinux2014_aarch64.whl" }
```

The fix simply transforms that mysterious exception into a more self-explanatory one.

### DIFF
--- a/src/poetry/installation/executor.py
+++ b/src/poetry/installation/executor.py
@@ -780,9 +780,15 @@ class Executor:
             strict=False,
             env=self._env,
         )
-        # 'archive' can at this point never be None. Since we previously downloaded
-        # an archive, we now should have something cached that we can use here
-        assert archive is not None
+        if archive is None:
+            # Since we previously downloaded an archive, we now should have
+            # something cached that we can use here. The only case in which
+            # archive is None is if the original archive is not valid for the
+            # current environment.
+            raise RuntimeError(
+                f"Package {link.url} cannot be installed in the current environment"
+                f" {self._env.marker_env}"
+            )
 
         if archive.suffix != ".whl":
             message = (


dimbleby:
this isn't a `ValueError`, no-one has provided an unsuitable value anywhere.  

Really it's just a plain bug: if left unfixed then suggest that `RuntimeError` is more accurate.

ralbertazzi:
@dimbleby I think I didn't get what bug exactly are you referring to, isn't this a user error?

radoering:
Same here. Do you think it's a bug that it's possible to lock something like this and the error only pops up in the executor?

ralbertazzi:
While technically you could avoid locking my example in the first place, I think there's other situations in which you really can't know until you install (different os, different arch, ..), so I think this error handling is still valid. Plus, it feels consistent with the current "lock everything" behaviour of Poetry.

dimbleby:
I definitely think that this error popping up in `download_link()` is strange.  

`_download_link()` seems confused about what it is for.  It mixes up the straightforward "download an archive, possibly getting it from cache" with some sort of "but actually maybe not the archive that I was asked for, and I'll check compatibility with the current environment, but _only_ if I look in the cache".  I'm not sure how best to disentangle this: but it certainly is not pretty.

I'm fine with this MR being an incremental improvement: but it's very strange to me that checking an archive for suitability for the current environment happens (i) during download and (ii) only if the archive was in the cache.

(What's the behaviour in that repro with `--no-cache`? Perhaps this cache just doesn't respect `--no-cache`...)

ralbertazzi:
> Perhaps this cache just doesn't respect --no-cache

It doesn't. https://github.com/python-poetry/poetry/pull/7693#issuecomment-1529075338

radoering:
I agree that `_download_link()` does more than you would expect it to do but since that has already been the case in 1.4, we can defer cleaning this up.

> (ii) only if the archive was in the cache

The check is always executed, no matter if the artifact was in the cache before. If it was not in the cache it is downloaded and put into the cache before this check.

radoering (approved):

PR #7963, opened 2023-05-19
preserve relative-path at `poetry run`

### LINKED ISSUE #7959: Poetry run no longer executes local bash scripts
- **Poetry version**: 1.5.0
- **Python version**: 3.11.3
- **OS version and name**:  macOS 13.2.1 (22D68)
- **pyproject.toml**: [Basic pyproject.toml as created by `poetry new poetry-bug`](https://gist.github.com/oefe/1c8f02fa54de74b7f68807d024e5035d)

## Issue

In Poetry 1.4.2, you could execute local shell scripts using `poetry run`, e.g. `poetry run ./pyright.sh`.

This no longer works with Poetry 1.5

```
>ls -l *.sh
-rwxr-xr-x  1 user  staff  1871 May 19 17:49 pyright.sh*

>head -n 1 pyright.sh
#!/bin/bash

>poetry ./pyright.sh
The command "./pyright.sh" does not exist.

>poetry pyright.sh
The command "./pyright.sh" does not exist.
```

The changelog mentions #7606, but this should only affect python scripts listed in `[tool.poetry.scripts]`, which is not the case here.

So I would expect that this still works.

## Workaround

Explicitly calling `bash` with the script as argument, i.e. `poetry run bash ./pyright.sh` still works

dimbleby:
fixes #7959

a regression introduced in the effort to use `Path` for paths.  Turns out that `Env._bin()` isn't really returning a path at all, it's returning a thing that you pass to `subprocess.Popen()`

In particular, `Path("./foo.py")` is the same as `Path("foo.py")`

### DIFF
--- a/src/poetry/utils/env.py
+++ b/src/poetry/utils/env.py
@@ -1243,7 +1243,7 @@ class Env:
         """
         Path to current python executable
         """
-        return self._bin(self._executable)
+        return Path(self._bin(self._executable))
 
     @property
     def marker_env(self) -> dict[str, Any]:
@@ -1311,7 +1311,7 @@ class Env:
         """
         # we do not use as_posix() here due to issues with windows pathlib2
         # implementation
-        path = self._bin(self._pip_executable)
+        path = Path(self._bin(self._pip_executable))
         if not path.exists():
             return self.pip_embedded
         return path
@@ -1431,7 +1431,7 @@ class Env:
         raise NotImplementedError()
 
     def get_pip_command(self, embedded: bool = False) -> list[str]:
-        if embedded or not self._bin(self._pip_executable).exists():
+        if embedded or not Path(self._bin(self._pip_executable)).exists():
             return [str(self.python), str(self.pip_embedded)]
         # run as module so that pip can update itself on Windows
         return [str(self.python), "-m", "pip"]
@@ -1461,7 +1461,7 @@ class Env:
             # embedded pip when pip is not available in the environment
             return self.get_pip_command()
 
-        return [str(self._bin(bin))]
+        return [self._bin(bin)]
 
     def run(self, bin: str, *args: str, **kwargs: Any) -> str:
         cmd = self.get_command_from_bin(bin) + list(args)
@@ -1541,7 +1541,7 @@ class Env:
                 self._script_dirs.append(self.userbase / self._script_dirs[0].name)
         return self._script_dirs
 
-    def _bin(self, bin: str) -> Path:
+    def _bin(self, bin: str) -> str:
         """
         Return path to the given executable.
         """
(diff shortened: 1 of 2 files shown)


src/poetry/utils/env.py
```diff
@@ -1541,7 +1541,7 @@ def script_dirs(self) -> list[Path]:
                 self._script_dirs.append(self.userbase / self._script_dirs[0].name)
         return self._script_dirs
 
-    def _bin(self, bin: str) -> Path:
+    def _bin(self, bin: str) -> str:
```
radoering, line 1544: Maybe, we should mention why we can't return a `Path` in the docstring.

dimbleby: suggestions welcome.  I find this such a confused function that I don't know how to write a sensible sentence describing it!

radoering:
```
Return path to the given executable as required by subprocess.POpen().

Caution: We can't return Path objects because we need to preserve
the leading "./" in relative paths (e.g. if "bin" is a shell script).
```

dimbleby: seems to me that belongs in `get_command_from_bin()` if anywhere; and that the testcase more than has it covered.  

Eg an alternative fix that I considered was for `_bin()` to continue to return a `Path` but actually look for the executable with `shutil.which()` - though then you have to account for the case where it doesn't exist earlier than it is currently done.

radoering: Yeah, probably the test case is enough. I'm just wondering if the test case needs some explanation? What if someone thinks `assert command == ["foo.py"]` is just as good as `assert command == ["./foo.py"]`? Both paths are not absolute paths...

dimbleby: Of course testcases are always vulnerable to someone deciding they were wrong all along, but this one is called `test_command_from_bin_preserves_relative_path()`.  

If that's not enough to give pause for thought when removing the './' - I would expect at that point to `git blame` and find this discussion - then I doubt that putting a comment in the code under test is going to stop them.

I'll add an link from the testcase to the original issue, maybe that'll save someone a `git blame` one day.

radoering: Probably, I'm overthinking this. We'll see. Thank you.

radoering (approved):

djch:
Nice to see this was addressed already 🙂

I'm hoping this one can go out in a bug fix release ASAP. I've got a lot of scripts in a lot of Django projects which call `./manage.py`.

In the mean time, can I run `poetry self update *something*` to install poetry from `master` and get this fix?

radoering:
> In the mean time, can I run `poetry self update *something*` to install poetry from `master` and get this fix?

Not sure about `poetry self` but if you use pipx, you can install the 1.5 branch (from which 1.5.1 will be built) via

```
pipx install git+https://github.com/python-poetry/poetry.git@1.5
```

Issue #7722, opened 2023-03-27
Multiple environment markers doesn't work


marhoy:
- **Poetry version**: 1.4.1
- **Python version**: 3.10.7
- **OS version and name**: Ubuntu 22.04.2 and Debian 11

- [x] I am on the [latest](https://github.com/python-poetry/poetry/releases/latest) stable Poetry version, installed using a recommended method.
- [x] I have searched the [issues](https://github.com/python-poetry/poetry/issues) of this repo and believe that this is not a duplicate.
- [x] I have consulted the [FAQ](https://python-poetry.org/docs/faq/) and [blog](https://python-poetry.org/blog/) for any relevant entries or release notes.
- [x] If an exception occurs when executing a command, I executed it again in debug mode (`-vvv` option) and have included the output below.

## Issue

I'm trying to install a different version of `torch`, depending on the CPU. If I'm on the Raspberry Pi, I want to use the wheel from download.pytorch.org, otherwise I can install from pypi. (The reason for this is a cuda-dependency-issue, but that's not the point here).

I have the following in my `pyproject.toml`
```
torch = [
    { markers = "platform_machine == 'aarch64'", url = "https://download.pytorch.org/whl/torch-1.13.1-cp310-cp310-manylinux2014_aarch64.whl"},
    { markers = "platform_machine == 'x86_64'", version = "^1.13.1"},
]
```

The problem is that the marker-logic doesn't work: Poetry just installs the *first* of the two alternatives, and ignores the markes.

Am I doing something wrong?

radoering:
> duplicate #5506 I expect

I don't think so. #5506 is about one multiple-constraints dependency and one restricted (single constraint) dependency, which somehow depend on each other.

At first glance this looks as if it should work except when local version identifiers are involved. Then, it could be a duplicate of some other issue.

@marhoy Which versions are locked. Can you share your lock file?

dimbleby:
you should provide an explicit `source = "pypi"` in the one you want from pypi, so that poetry knows that it cannot be satisfied by the url whl.

That's still not enough, though, this fix looks like it should finish the job:
```patch
diff --git a/src/poetry/puzzle/provider.py b/src/poetry/puzzle/provider.py
index ed88b57d..79271517 100644
--- a/src/poetry/puzzle/provider.py
+++ b/src/poetry/puzzle/provider.py
@@ -320,12 +320,10 @@ class Provider:
         #
         # We rely on the VersionSolver resolving direct-origin dependencies first.
[...]

PR #7973, opened 2023-05-20
fix: explicit source dependency is not satisfied by direct origin

ralbertazzi:
Resolves: #7722

cc @dimbleby

### DIFF
--- a/tests/puzzle/test_provider.py
+++ b/tests/puzzle/test_provider.py
@@ -113,7 +113,7 @@ def test_search_for(
             Dependency("foo", ">=2"),
             URLDependency("foo", SOME_URL),
             [Package("foo", "3")],
-            [],
+            [Package("foo", "3")],
         ),
         (
             Dependency("foo", ">=1", extras=["bar"]),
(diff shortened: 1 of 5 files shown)


ralbertazzi:
There seems to be a weird issue with pytest fixtures in the failing test 🤔

tests/puzzle/test_provider.py
```diff
@@ -113,7 +113,7 @@ def test_search_for(
             Dependency("foo", ">=2"),
             URLDependency("foo", SOME_URL),
             [Package("foo", "3")],
-            [],
+            [Package("foo", "3")],
```
ralbertazzi, line 116: Looks like correct behaviour to me: the dependency is not satisfied by the URL package and so it falls back to checking the repository (which exactly the proposed change)

dimbleby:
So far as the code change goes I'm basically reviewing my own suggestion here.  Looks fine to me, but this is not an independent review.

I think I would have expected to write a testcase here that more closely follows the actual problem as laid out in the issue report.  Probably something in `test_installer.py` that sets things up and runs `poetry install` and checks that the expected thing happens.

I always find those testcases unpleasant to write, which is partly what put me off doing anything with this one in the first place.  If you can convince yourself - and someone with the commit bit - that what you have is sufficient, then your life will be easier.

ralbertazzi:
I added a test, LMK if that's what you were looking for ;)

radoering (approved):

PR #7977, opened 2023-05-20
fix check for in-project-venv

### LINKED ISSUE #7957: Poetry 1.5 fails to install dependencies
- **Poetry version**: 1.5
- **Python version**: 3.10
- **OS version and name**:  Ubuntu 20.04

- [X ] I am on the [latest](https://github.com/python-poetry/poetry/releases/latest) stable Poetry version, installed using a recommended method.
- [ X] I have searched the [issues](https://github.com/python-poetry/poetry/issues) of this repo and believe that this is not a duplicate.
- [X ] I have consulted the [FAQ](https://python-poetry.org/docs/faq/) and [blog](https://python-poetry.org/blog/) for any relevant entries or release notes.
- [ X] If an exception occurs when executing a command, I executed it again in debug mode (`-vvv` option) and have included the output below.

## Issue

When running `Poetry install` with the latest version (1.5) that was just released, I am getting `No such file or directory 'python'`.

The exact same project etc works fine with version 1.4.2.

dimbleby:
looks like #7792 was a bit careless, confusing two things

- it rightly replaced two "is in-project-venv enabled?" checks with "is in-project-venv enabled, OR not explicitly configured but the directory exists?"
- it wrongly replaced two "is in-project-venv enabled AND the directory already exists?" checks in the same way

this undoes those mistakes

fixes #7957

### DIFF
--- a/src/poetry/utils/env.py
+++ b/src/poetry/utils/env.py
@@ -695,7 +695,7 @@ class EnvManager:
 
         if not in_venv or env is not None:
             # Checking if a local virtualenv exists
-            if self.use_in_project_venv():
+            if self.in_project_venv_exists():
                 venv = self.in_project_venv
 
                 return VirtualEnv(venv)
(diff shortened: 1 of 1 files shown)


radoering (commented):
Is a test feasible?

dimbleby:
unclear how to test.  

`test_activate_with_in_project_setting_does_not_fail_if_no_venvs_dir()` executes the code path but the reason that it shows no problem is that the mocking `check_output_wrapper()` does not know that `python -c ...` is trying to call a non-existent python.

radoering (approved):

manugarri:
maybe a way to test this would be to have an end to end test? create a new poetry environment with the flag set to true?

PR #7978, opened 2023-05-21
Fix dependency cache inconsistency with prerelease versions

chriskuehl:
This PR fixes an inconsistency identified in https://github.com/python-poetry/poetry/pull/7950#issuecomment-1555344115 where in some situations Poetry will not select a prerelease package version even though it satisfies all constraints.

Poetry is supposed to consider prerelease package versions if no other package versions satisfy the constraints. I believe this works if the constrained dependency is at the top level, but not necessarily if the constraint is introduced at a later level. Currently, whether it works or not depends on some arbitrary factors like the order dependencies are evaluated in and if the solver recently backtracked.

### Regression test

I introduced a regression test which fails on `master`:

```
____ test_solver_with_dependency_and_prerelease_sub_dependencies_increasing_constraints _____
  File "/home/ckuehl/proj/poetry/src/poetry/puzzle/solver.py", line 155, in _solve
    result = resolve_version(self._package, self._provider)
  File "/home/ckuehl/proj/poetry/src/poetry/mixology/__init__.py", line 18, in resolve_version
    return solver.solve()
  File "/home/ckuehl/proj/poetry/src/poetry/mixology/version_solver.py", line 111, in solve
    self._propagate(next)
  File "/home/ckuehl/proj/poetry/src/poetry/mixology/version_solver.py", line 150, in _propagate
    root_cause = self._resolve_conflict(incompatibility)
  File "/home/ckuehl/proj/poetry/src/poetry/mixology/version_solver.py", line 351, in _resolve_conflict
    raise SolveFailure(incompatibility)
poetry.mixology.failure.SolveFailure: Because no versions of a match !=1.0
 and a (1.0) depends on B (>0.9.0), every version of a requires B (>0.9.0).
So, because no versions of b match >0.9.0
 and root depends on A (*), version solving failed.

During handling of the above exception, another exception occurred:

  File "/home/ckuehl/proj/poetry/tests/puzzle/test_solver.py", line 1222, in test_solver_with_dependency_and_prerelease_sub_dependencies_increasing_constraints
    transaction = solver.solve()
  File "/home/ckuehl/proj/poetry/src/poetry/puzzle/solver.py", line 72, in solve
    packages, depths = self._solve()
  File "/home/ckuehl/proj/poetry/src/poetry/puzzle/solver.py", line 161, in _solve
    raise SolverProblemError(e)
poetry.puzzle.exceptions.SolverProblemError: Because no versions of a match !=1.0
 and a (1.0) depends on B (>0.9.0), every version of a requires B (>0.9.0).
So, because no versions of b match >0.9.0
 and root depends on A (*), version solving failed.
```

The test passes if I disable the `DependencyCache` (replace the contents of `DependencyCache._search_for()` with `return self.provider.search_for(dependency)`, and also passes with the patch provided in this PR.

Additionally, the reproduction using `black<22` from https://github.com/python-poetry/poetry/pull/7950#issuecomment-1555344115 also succeeds for me now.

### DIFF
--- a/tests/puzzle/test_solver.py
+++ b/tests/puzzle/test_solver.py
@@ -1183,6 +1183,64 @@ def test_solver_with_dependency_and_prerelease_sub_dependencies(
     )
 
 
+def test_solver_with_dependency_and_prerelease_sub_dependencies_increasing_constraints(
+    solver: Solver,
+    repo: Repository,
+    package: ProjectPackage,
+    mocker: MockerFixture,
+) -> None:
+    """Regression test to ensure the solver eventually uses pre-release
+    dependencies if the package is progressively constrained enough.
+
+    This is different from test_solver_with_dependency_and_prerelease_sub_dependencies
+    above because it also has a wildcard dependency on B at the root level.
+    This causes the solver to first narrow B's candidate versions down to
+    {0.9.0} at an early level, then eventually down to the empty set once A's
+    dependencies are processed at a later level.
+
+    Once the candidate version set is narrowed down to the empty set, the
+    solver should re-evaluate available candidate versions from the source, but
+    include pre-release versions this time as there are no other options.
+    """
+    # Note: The order matters here; B must be added before A or the solver
+    # evaluates A first and we don't encounter the issue. This is a bit
+    # fragile, but the mock call assertions ensure this ordering is maintained.
+    package.add_dependency(Factory.create_dependency("B", "*"))
+    package.add_dependency(Factory.create_dependency("A", "*"))
+
+    package_a = get_package("A", "1.0")
+    package_a.add_dependency(Factory.create_dependency("B", ">0.9.0"))
+
+    repo.add_package(package_a)
+    repo.add_package(get_package("B", "0.9.0"))
+    package_b = get_package("B", "1.0.0.dev4")
+    repo.add_package(package_b)
+
+    search_for_spy = mocker.spy(solver._provider, "search_for")
+    transaction = solver.solve()
+
+    check_solver_result(
+        transaction,
+        [
+            {"job": "install", "package": package_b},
+            {"job": "install", "package": package_a},
+        ],
+    )
+
+    # The assertions below aren't really the point of this test, but are just
+    # being used to ensure the dependency resolution ordering remains the same.
+    search_calls = [
+        call.args[0]
+        for call in search_for_spy.mock_calls
+        if call.args[0].name in ("a", "b")
+    ]
+    assert search_calls == [
+        Dependency("a", "*"),
+        Dependency("b", "*"),
+        Dependency("b", ">0.9.0"),
+    ]
+
+
 def test_solver_circular_dependency(
     solver: Solver, repo: Repository, package: ProjectPackage
 ) -> None:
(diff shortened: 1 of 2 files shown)


tests/puzzle/test_solver.py
```diff
@@ -1183,6 +1184,65 @@ def test_solver_with_dependency_and_prerelease_sub_dependencies(
...
+    This is different from test_solver_with_dependency_and_prerelease_sub_dependencies
+    above because it also has a wildcard dependency on B at the root level.
+    This causes the solver to first narrow B's candidate versions down to
+    {0.9.0} at an early level, then eventually down to the empty set once A's
+    dependencies are processed at a later level.
+
+    Once the candidate version set is narrowed down to the empty set, the
+    solver should re-evaluate available candidate versions from the source, but
+    include pre-release versions this time as there are no other options.
+    """
```
chriskuehl, line 1202: Sorry for all the verbosity—I was having a tough time  finding words to clearly describe the situation I'm trying to test. Open to suggestions to improve this!

tests/puzzle/test_solver.py
```diff
@@ -1183,6 +1184,65 @@ def test_solver_with_dependency_and_prerelease_sub_dependencies(
...
+    # Note: The order matters here; B must be added before A or the solver
+    # evaluates A first and we don't encounter the issue. This is a bit
+    # fragile, but the mock call assertions ensure this ordering is maintained.
+    package.add_dependency(Factory.create_dependency("B", "*"))
+    package.add_dependency(Factory.create_dependency("A", "*"))
```
chriskuehl, line 1207: Let me know if you prefer me to remove the assertions at the bottom of this test. My concern was that this test is a bit brittle (e.g. simply swapping the order of these two lines is otherwise enough to make it pass on master) and the assertions protect against that.

tests/puzzle/test_solver.py
```diff
@@ -1183,6 +1184,65 @@ def test_solver_with_dependency_and_prerelease_sub_dependencies(
...
+
+    package_a = get_package("A", "1.0")
+    package_a.add_dependency(Factory.create_dependency("B", ">0.9.0"))
+
+    repo.add_package(package_a)
+    repo.add_package(get_package("B", "0.9.0"))
+    package_b = get_package("B", "1.0.0.dev4")
+    repo.add_package(package_b)
+
+    with mock.patch.object(
```
radoering, line 1217: Typically, we use the mocker fixture (add `mocker: MockerFixture` to the test function's parameters).

Then, this can be used instead of `mock.patch.object()`:

```
search_for_spy = mocker.spy(solver._provider, "search_for")
```

tests/puzzle/test_solver.py
radoering: Many of our tests are not documented well and sometimes you even have to guess the objective of the test. IMO, the description is fine here considering it's a quite complex use case.

tests/puzzle/test_solver.py
radoering: I like the assertions and the explanatory comments. It's always good to make it harder to render a test case useless.

radoering (changes requested):
LGTM. 👍 Just something about our (implicit) internal coding style considering tests.

tests/puzzle/test_solver.py
chriskuehl: Fixed now.

radoering (approved):

Issue #7983, opened 2023-05-22
Modern installer eats a lot of RAM when installing big wheels


ralbertazzi:
- **Poetry version**: 1.5.0
- **Python version**: 3.9.8
- **OS version and name**: Ubuntu 20.04 (also observed on other platforms)
- **pyproject.toml**:
```toml
[tool.poetry]
authors = ["Author <foo@bar.com>"]
description = "Project"
name = "project"
version = "1.0.0"

[tool.poetry.dependencies]
python = "~3.9"

torch = { url = "https://download.pytorch.org/whl/cu113/torch-1.11.0%2Bcu113-cp39-cp39-linux_x86_64.whl" }
```

- [x] I am on the [latest](https://github.com/python-poetry/poetry/releases/latest) stable Poetry version, installed using a recommended method.
- [x] I have searched the [issues](https://github.com/python-poetry/poetry/issues) of this repo and believe that this is not a duplicate.
- [x] I have consulted the [FAQ](https://python-poetry.org/docs/faq/) and [blog](https://python-poetry.org/blog/) for any relevant entries or release notes.
- [x] If an exception occurs when executing a command, I executed it again in debug mode (`-vvv` option) and have included the output below.

## Issue

Poetry's modern installer is using a lot of RAM when installing big wheels, such as the PyTorch one (>1 GB) that I provided in the example. The RAM usage starts increasing while Poetry is `Installing...` the wheel, and it spikes over 10 GB of RAM used while performing the operation. The RAM usage goes back to acceptable values before the `Installing...` operation is done.

Deactivating the modern installer (`poetry config installer.modern-installation false`) fixes the problem. The RAM usage does not go beyond 1 GB there.

I haven't debugged deeply enough to tell if it's a Poetry issue or an `pypa/installer` issue, but for sure Poetry is affected by it. This is bad for CI environments where resources are more limited and currently cause `poetry install` to go OOM. 

cc #6409

ralbertazzi:
Found the culprit being in [validate_record](https://github.com/python-poetry/poetry/blob/88cb8266679287af123cf71daaa07544711cd2bc/src/poetry/installation/wheel_installer.py#L104). More specifically, it looks like the garbage collector can't stand behind [self._zipfile.read(item)](https://github.com/pypa/installer/blob/main/src/installer/sources.py#L290). The RAM increases while the line is executed but it also seems to go quite down by placing a breakpoint on the next line.

[...]

PR #7987, opened 2023-05-23
fix: disable wheel content validation

ralbertazzi:
# Pull Request Check List

Closes: https://github.com/python-poetry/poetry/issues/7983

First commit is a temporary solution to https://github.com/python-poetry/poetry/issues/7983 until https://github.com/pypa/installer/pull/183 gets released.

~Second commit is a permanent solution that also provides faster installation times.~

### DIFF
--- a/src/poetry/installation/wheel_installer.py
+++ b/src/poetry/installation/wheel_installer.py
@@ -101,7 +101,10 @@ class WheelInstaller:
     def install(self, wheel: Path) -> None:
         with WheelFile.open(wheel) as source:
             try:
-                source.validate_record()
+                # Content validation is temporarily disabled because of
+                # pypa/installer's out of memory issues with big wheels. See
+                # https://github.com/python-poetry/poetry/issues/7983
+                source.validate_record(validate_contents=False)
             except _WheelFileValidationError as e:
                 self.invalid_wheels[wheel] = e.issues
             install(
--- a/tests/installation/test_executor.py
+++ b/tests/installation/test_executor.py
@@ -335,6 +335,7 @@ def test_execute_prints_warning_for_yanked_package(
         assert error.count("yanked") == 0
 
 
+@pytest.mark.skip(reason="https://github.com/python-poetry/poetry/issues/7983")
 def test_execute_prints_warning_for_invalid_wheels(
     config: Config,
     pool: RepositoryPool,


ralbertazzi:
I added another commit where I achieve what I was saying here: https://github.com/python-poetry/poetry/issues/7983#issuecomment-1558645745

This:
- Requires no release of pypa/installer
- Makes the installation process faster, as we compute file hashes just once
- Does not lose any validation capabilities. They are just handled a bit later

Cons: some pypa/installer code spilling into the Poetry codebase

radoering:
Not sure if the second commit is worth it or if we should just turn off validation until there is a solution in `installer` (or we really need validation for some reason). For a patch release I'd probably go with the first commit only (and include the second commit in the next minor if we want to include it).

> First commit is a temporary solution to https://github.com/python-poetry/poetry/issues/7983 until https://github.com/pypa/installer/pull/183 gets released.

The installer PR only reduces the memory consumption, the hash will still be computed twice, won't it?

src/poetry/installation/wheel_installer.py
```diff
@@ -66,12 +87,52 @@ def for_source(self, source: WheelFile) -> WheelDestination:
...
+        }
+        for item in self._source._zipfile.infolist():
+            record_args = record_mapping.pop(item.filename, None)
+            if not record_args:
+                continue
+
+            file_record = RecordEntry.from_elements(*record_args)
+            computed_record = next(
+                record for _, record in records if record.path == item.filename
+            )
```
radoering, line 112: This doesn't always work. I tested a large project and got a `StopIteration` for some packages.

You can reproduce it by running `poetry add fonttools` for example.

What's even worth: The package is installed, but the `RECORD` file is not written.

`poetry run pip uninstall fonttools` now throws `ERROR: Cannot uninstall fonttools 4.39.4, RECORD file not found. Hint: The package was installed by Poetry 1.6.0.dev0.` and `poetry run pip install fonttools` gives `Requirement already satisfied`.

You can only "manually" uninstall the package.

src/poetry/installation/wheel_installer.py
```diff
@@ -66,12 +87,52 @@ def for_source(self, source: WheelFile) -> WheelDestination:
...
+                )
+
+    def finalize_installation(
+        self,
+        scheme: Scheme,
+        record_file_path: str,
+        records: Iterable[tuple[Scheme, RecordEntry]],
+    ) -> None:
+        self._validate_hash_and_size(records)
+        return super().finalize_installation(scheme, record_file_path, records)
```
radoering, line 134:
```suggestion
        try:
            self._validate_hash_and_size(records)
        finally:
            return super().finalize_installation(scheme, record_file_path, records)
```

I wonder if we should do something like that to make sure the `RECORD` file is written despite of bugs in `_validate_hash_and_size()`. However, when I tried this, the exception was swallowed, which not good either. Thus, we would have to put more effort here...

radoering (changes requested):

ralbertazzi:
@radoering yes, I agree the PR still needs some touches. The way I got it is that we wouldn't have merged the additional work on top of the simple `validate_contents=False`, am I right?

radoering:
Yes, I think for 1.5.1 the first commit is fine. The second commit needs more work and testing. You can split the second commit into a separate PR if you want to keep working on it.

radoering (approved):

PR #7992, opened 2023-05-23
7988 fix resolving subdependencies from explicit source

b-kamphorst:
# Pull Request Check List

Resolves: #7988

### DIFF
--- a/tests/puzzle/test_solver.py
+++ b/tests/puzzle/test_solver.py
@@ -3108,6 +3108,42 @@ def test_solver_ignores_explicit_repo_for_transient_dependencies(
         solver.solve()
 
 
+def test_solver_finds_inter_dependencies_in_explicit_repo(
+    package: ProjectPackage,
+    repo: Repository,
+    io: NullIO,
+) -> None:
+    # dependency of root with explicit source
+    package.add_dependency(
+        Factory.create_dependency("demo", {"version": "*", "source": "repo"})
+    )
+    demo = Package("demo", "0.1.2")
+    repo.add_package(demo)
+
+    # transient dependency of demo with explicit source
+    # also specified as dependency of root with explicit source
+    package.add_dependency(
+        Factory.create_dependency("pendulum", {"version": "*", "source": "repo"})
+    )
+    demo.add_dependency(Factory.create_dependency("pendulum", "*"))
+    pendulum = get_package("pendulum", "2.0.3")
+    repo.add_package(pendulum)
+
+    pool = RepositoryPool()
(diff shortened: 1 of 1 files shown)


b-kamphorst:
Tagging @dimbleby for his involvement in https://github.com/orgs/python-poetry/discussions/7988.

Tagging @radoering for his involvement in [#7658](https://github.com/python-poetry/poetry/pull/7658), [#6879](https://github.com/python-poetry/poetry/pull/6879), and [#7801](https://github.com/python-poetry/poetry/pull/7801) -- my guess is that you might be familiar with the downloading of packages and might be interested specifically in https://github.com/orgs/python-poetry/discussions/7988#discussioncomment-5981635.

My best guess at this point is that the issue is caused by the following elements:
1. a `Package` instantiated by [`PyPiRepository`](https://github.com/python-poetry/poetry/blob/1.5.0/src/poetry/repositories/pypi_repository.py) is unaware of its `source` (e.g. `Package.source_type` and `Package.source_reference` are both `None`);
2. a `Package` derived from a `Dependency` through `RepositoryPool.find_packages(dependency)` is (therefore also) unaware of its source -- even if that source is explicitly requested in `Dependency.source_reference`;
2. the [`Chooser`](https://github.com/python-poetry/poetry/blob/1.5.0/src/poetry/installation/chooser.py#L83) performs some fuzzy logic to get the links to a package. It appears to be impossible for that logic to have access to an explicit source. As I understand now, the Chooser:
    - tries to extract a package `source_reference` ONLY IF a `source_type` is provided (as stated, both are `None` if the source is a `PyPiRepository`);
    - if it fails, grabs the `pypi` repository regardless of other sources that are defined or the priority of the pypi repository;
    - if that fails yet again, grabs the first repository returned by `RepositoryPool.repositories` (can never have `Priority.EXPLICIT`).

This line of reasoning explains why the puzzle is solved successfully (uses [`VersionSolver._dependency_cache.search_for(dependency,...)`](https://github.com/python-poetry/poetry/blob/1.5.0/src/poetry/mixology/version_solver.py#L415) and therefore maintains source information). It would also explain why PyPI is searched. It does not explain why the `supplemental` option works as this repository seems to be ignored by the Chooser as well. Therefore I must be missing something still. Looking forward to your thoughts.

dimbleby:
do we even have a repro at this point?

The discussion seems to be about locking rather than installing, it doesnt look as though anything should get as far as the `Chooser`.  

I would recommend a healthy degree of scepticism about whether what is being claimed is actually the case - folk misreport their misunderstandings on this tracker all the time.  If you have made a good conscience best effort to reproduce and not managed it, you can go back and ask for a better reproduction.

Issue #7961, opened 2023-05-19
Poetry now does not add the package.dependencies


martingaldeca:
- **Poetry version**:1.5.0
- **Python version**: >3.10
- **OS version and name**: Ubuntu 22.04
- **pyproject.toml**: private

## Issue
I was using the poetry version 1.4.2, when adding a new package from a private pypi it adds also the package dependencies well as you can see here: (I remove the private values, but I promise that are the same for the other example)
![1 14 2](https://github.com/python-poetry/poetry/assets/36915710/94c76007-7475-4e1a-a84d-55cfc26a38aa)

When using poetry version 1.5.0, for the same package, poetry did not added the dependencies, also the description was not added and the python version not specified:
![1 15 0](https://github.com/python-poetry/poetry/assets/36915710/2c502799-c7e8-4fea-9725-4b4f11af75fb)

So in that case for us this is not a minor update, this is a major update as it is breaking some of our dependencies.

dimbleby:
I use private dependencies and repositories, and poetry 1.5.0 works fine

without a way to reproduce this there's not going to be very much that anyone can do.

cquick01:
Also seeing the same behavior on 1.5.0. We have several Python packages hosted in a private repo, and when trying to add them to a new project with `poetry add ...`, the package itself is installed but no dependencies are resolved. Note that the dependencies that are missing are just normal public PyPi packages.

I have my sources defined as follows, if it helps.

```yaml
[[tool.poetry.source]]
name = "private_pypi"
url = "https://pypi.domain.tld/simple/"
priority = "supplemental"

[[tool.poetry.source]]
name = "PyPI"
priority = "primary"
```

dimbleby:
Works fine for me using `pypi-server` as my private repository:
```
$ tree ~/packages/
/home/dch/packages/
└── simple
    └── cq-test-poetry
        └── cq_test_poetry-0.1.0-py3-none-any.whl

2 directories, 1 file
```

```
$ pypi-server run --hash-algo sha256 -p 8080 ~/packages/
```

```toml
[tool.poetry.dependencies]
python = "^3.10"

[[tool.poetry.source]]
name = "private_pypi"
url = "http://localhost:8080/simple/"
priority = "supplemental"

[[tool.poetry.source]]
name = "PyPI"
priority = "primary"
```
```
$ poetry --version
Poetry (version 1.5.0)

$ poetry add cq-test-poetry
Using version ^0.1.0 for cq-test-poetry

[...]

PR #7995, opened 2023-05-24
fix: revert #7916

ralbertazzi:
# Pull Request Check List

Resolves: #7961 

Reverting #7916 (funny how it's the same issue numbers reverted) as the cached file from `cachecontrol` contains the entire HTTP response, not just its content. Terribly sorry for that, I don't know how this slipped when debugging.

There's still probably some optimizations we could do but we should prioritize making Poetry correct again. The PR reverts the old unwanted behaviours:
- An unwanted file copy operation
- An unwanted _slow_ file copy operation, as we perform it with a relative small chunk size
- An unwanted "Downloading <wheel>" log when in reality we are not

### DIFF
--- a/src/poetry/repositories/http_repository.py
+++ b/src/poetry/repositories/http_repository.py
@@ -76,15 +76,11 @@ class HTTPRepository(CachedRepository):
 
     @contextmanager
     def _cached_or_downloaded_file(self, link: Link) -> Iterator[Path]:
-        filepath = self._authenticator.get_cached_file_for_url(link.url)
-        if filepath:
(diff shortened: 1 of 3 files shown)


ralbertazzi:
As far as I understand, I should be able to re-achieve the same behaviour by using https://cacheyou.readthedocs.io/en/latest/storage.html#separatebodyfilecache. I'll give it a try once #https://github.com/python-poetry/poetry/pull/7997 is merged. Clearly not a blocker for merging this in the meantime.

radoering (approved):

PR #7997, opened 2023-05-24
Replace cachecontrol with a fork

Secrus:
CacheControl package was largely unmaintained and made us haul around a deprecated dependency, `lockfile`. Due to recent changes connected to `urllib3` 2.0 release, a fork was made and is now maintained by @frostming. This fork merges some long-waiting PRs of cachecontrol, one of them being typing support, so we no longer need to ignore typing errors from that library.

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -42,23 +42,23 @@ typing = ["importlib-metadata (>=5.1)", "mypy (==0.991)", "tomli", "typing-exten
 virtualenv = ["virtualenv (>=20.0.35)"]
 
 [[package]]
-name = "cachecontrol"
-version = "0.12.11"
+name = "cacheyou"
+version = "23.2"
 description = "httplib2 caching for requests"
 optional = false
-python-versions = ">=3.6"
+python-versions = ">=3.7"
 files = [
-    {file = "CacheControl-0.12.11-py2.py3-none-any.whl", hash = "sha256:2c75d6a8938cb1933c75c50184549ad42728a27e9f6b92fd677c3151aa72555b"},
-    {file = "CacheControl-0.12.11.tar.gz", hash = "sha256:a5b9fcc986b184db101aa280b42ecdcdfc524892596f606858e0b7a8b4d9e144"},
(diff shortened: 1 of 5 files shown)


dimbleby:
should be able to remove filelock as a direct dependency too

I had also seen this and was holding off on making this pull request until https://github.com/frostming/cacheyou/issues/10 was resolved - though I guess if it's not causing us any problems perhaps it's no big deal

Secrus:
> should be able to remove filelock as a direct dependency too
> 
> I had also seen this and was holding off on making this pull request until [frostming/cacheyou#10](https://github.com/frostming/cacheyou/issues/10) was resolved - though I guess if it's not causing us any problems perhaps it's no big deal

Nice, didn't notice the `filelock` part. As to the issue, it's more important for distro packagers than for us, so I don't see it as an issue.

Secrus:
`cachecontrol` might be moved to PSF org, so I am drafting this for now until we see how this unfolds.

ralbertazzi:
Another +1 for considering the migration: I took a look again at https://github.com/python-poetry/poetry/pull/7916 (which was broken and reverted by https://github.com/python-poetry/poetry/pull/7995) and in looks like we could easily reintroduce the performance improvement by using a `SeparateBodyFileCache` instead of a basic `FileCache` implementation. However, `SeparateBodyFileCache` is badly broken on the latest pypi release of `cachecontrol`, used by Poetry (while issues seem fixed on master).

dimbleby:
https://github.com/ionrock/cachecontrol/issues/300 contains discussion about moving cachecontrol to a new home.

It's hard to tell, from the outside, whether that is active and about to happen: or has stalled in the same way as everything else in that repository.

Secrus:
I am gonna give them some more time. There might be some internal discussion about the topic.

dimbleby:
there's a cachecontrol 0.13.0

Secrus:
Yeah, I will make the changes tonight

Secrus:
> So the idea is to stick with `psf/cachecontrol` right?

Yes, since it is now maintained by the PSF and a dedicated team, not only the author.

Secrus:
superseded by #8055

PR #8001, opened 2023-05-24
[1.5] perf: Improve dependency resolution performance by not clearing caches when backtracking

chriskuehl:
Followed the instructions from https://github.com/python-poetry/poetry/pull/7950#issuecomment-1561615484 to fix the backport PR.


radoering (approved):
Thank you.

PR #8020, opened 2023-05-26
fix urljoin for arbitrary URL protocols

rmorshea:
This works around an issue with urllib.parse.urljoin where it only handles relative URLs for protocols contained in urllib.parse.uses_relative. As it happens common protocols used with git, like ssh or git+ssh are not in that list.

For more info: https://github.com/python-poetry/poetry/issues/6499#issuecomment-1564712609

# Pull Request Check List

Resolves: #6499


dimbleby:
I can't tell what this fixes, a test case showing the point would be helpful

rmorshea:
@dimbleby, I'll add a test, but for now, the problem this solves is described [here](https://github.com/python-poetry/poetry/issues/6499#issuecomment-1564712609).

rmorshea:
@dimbleby, I've added tests. Doing so required a bit of refactoring to make doing so easier though. The core the of the change, as before, is `urlpathjoin`, which appropriately handles relative paths for URLs with `ssh://` schemas. All other changes are related to testing this.

radoering (approved):

PR #8024, opened 2023-05-27
Add Dev Container configuration 

savannahostrowski:
This PR adds a [Dev Container](https://containers.dev/) configuration to Poetry. By adding this to the repo, contributors will be able to open the Poetry codebase in a Dev Container (via Visual Studio Code or Codespaces, for example) and have the local development environment setup for them per https://python-poetry.org/docs/contributing/.

I haven't updated the documentation in this PR but can if maintainers would like me to. No tests applicable here.

### DIFF
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -1,0 +1,10 @@
+
+ARG IMAGE="python:3.11"
+FROM --platform=amd64 mcr.microsoft.com/vscode/devcontainers/${IMAGE}
+RUN apt-get update \
+    && export DEBIAN_FRONTEND=noninteractive \
+    && apt-get -y install --no-install-recommends build-essential libssl-dev gdb cmake
+
+ENV POETRY_HOME="/opt/poetry"
+ENV PATH="$POETRY_HOME/bin:$PATH"
+RUN curl -sSL https://install.python-poetry.org | python3 -
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -1,4 +1,3 @@
-
 ARG IMAGE="python:3.11"
 FROM --platform=amd64 mcr.microsoft.com/vscode/devcontainers/${IMAGE}
 RUN apt-get update \
--- a/.devcontainer/setup.sh
+++ b/.devcontainer/setup.sh
@@ -1,0 +1,4 @@
+poetry install
+poetry run pytest
+poetry run mypy
+poetry run pre-commit install
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -1,30 +1,32 @@
 {
-	"name": "Poetry",
-	"build": {
-        "dockerfile": "Dockerfile",
-        "context": "..",
-        "args": {
-            "IMAGE": "python:3.11"
-        }
-    },
-	"postCreateCommand": "bash .devcontainer/setup.sh",
-	"customizations": {
-		"vscode": {
-			"settings": {
-				"python.formatting.blackPath": "black",
-				"python.formatting.provider": "black",
-				"python.testing.pytestEnabled": true,
-				"python.testing.pytestPath": "pytest",
-				"python.editor.codeActionsOnSave": {"source.fixAll": true},
-				"python.testing.pytestArgs": [
-					"tests"
-				]
-			},
-			"extensions": [
-				"ms-python.python",
-				"ms-python.black-formatter",
-                "charliermarsh.ruff"
-			]
-		}
-	}
+  "name": "Poetry",
+  "build": {
+    "dockerfile": "Dockerfile",
+    "context": "..",
+    "args": {
+      "IMAGE": "python:3.11"
+    }
+  },
+  "postCreateCommand": "bash .devcontainer/setup.sh",
+  "customizations": {
+    "vscode": {
+      "settings": {
+        "python.formatting.blackPath": "black",
+        "python.formatting.provider": "black",
+        "python.testing.pytestEnabled": true,
+        "python.testing.pytestPath": "pytest",
+        "python.editor.codeActionsOnSave": {
+          "source.fixAll": true
+        },
+        "python.testing.pytestArgs": [
+          "tests"
+        ]
+      },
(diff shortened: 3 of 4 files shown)


.devcontainer/Dockerfile
```diff
@@ -0,0 +1,10 @@
+ARG IMAGE=bullseye
```
eamanu, line 1: using bulleye is for any particular reason?

.devcontainer/setup.sh
```diff
@@ -0,0 +1,4 @@
+poetry install
```
eamanu, line 1: I don't know the Devcontainer workflow. The poetry run pre-commit install confuse me. I don't know if the purpose of setup.sh is run before start working in the new feature/bug/etc, in that way the pre-commit install make sense for me, but the other commands don't. 
Or when I finished to work in my new feature/bug/etc the setup.sh run, in that case I don't know if I want run the pre-commit install.
IMO it would be great write the new workflow (obviously if is necessary) in the documentation.

In any case, I like this idea.

.devcontainer/Dockerfile
savannahostrowski: Reading over things again, I realized I could simplify some things/ensure it's using the latest version of Python. I'm using the devcontainer images that Microsoft publishes to MCR.

.devcontainer/setup.sh
savannahostrowski: So, the `setup.sh` script runs just one time after the image has been created + the editor has launched (see postCreateCommand [here](https://containers.dev/implementors/json_reference/)).

The commands in this script are those which are denoted in https://python-poetry.org/docs/contributing/#local-development. You probably don't **need** to run `poetry run mypy` or `poetry run pytest` but I've included them to make sure everything is working as expected (doesn't really hurt to run them). 

This script/configuration doesn't change anything about the core development workflow as prescribed in the docs today.

savannahostrowski:
Just wanted to ping @Secrus on this since I chatted with them on Discord. Any feedback would be appreciated!

.devcontainer/Dockerfile
```diff
@@ -0,0 +1,9 @@
+FROM mcr.microsoft.com/devcontainers/python:3
```
joshspicer, line 1: I'd suggest pinning to a base OS/distro so that this this dev container continues to work even if the upstream OS changes (this happened recently with [the new version of debian](https://github.com/devcontainers/images/issues/607).  We often update our images fairly quickly with new distro releases, some upstream tools take a bit of time.

```suggestion
FROM mcr.microsoft.com/devcontainers/python:3-bullseye
```

or
```suggestion
FROM mcr.microsoft.com/devcontainers/python:3-bookworm
```

savannahostrowski: Ouuu thank you! Will do!

.devcontainer/Dockerfile
```diff
@@ -0,0 +1,9 @@
+FROM mcr.microsoft.com/devcontainers/python:3-bookworm
+
+RUN export DEBIAN_FRONTEND=noninteractive \
+     && apt-get update && apt-get install -y xdg-utils \
+     && apt-get clean -y && rm -rf /var/lib/apt/lists/*
+
+ENV POETRY_HOME="/opt/poetry"
+ENV PATH="$POETRY_HOME/bin:$PATH"
+RUN curl -sSL https://install.python-poetry.org | python3 -
```
abn, line 9:
```suggestion
RUN pipx install poetry
```

.devcontainer/Dockerfile
```diff
@@ -0,0 +1,9 @@
+FROM mcr.microsoft.com/devcontainers/python:3-bookworm
+
+RUN export DEBIAN_FRONTEND=noninteractive \
+     && apt-get update && apt-get install -y xdg-utils \
+     && apt-get clean -y && rm -rf /var/lib/apt/lists/*
```
neersighted, line 5: Why do we need `xdg-utils`? Does this add anything for users of a container used only via TTY?

.devcontainer/Dockerfile
```diff
@@ -0,0 +1,9 @@
+FROM mcr.microsoft.com/devcontainers/python:3-bookworm
```
neersighted, line 1: Any chance we can use a standard `python` base image here? If not, what does the Microsoft-supplied base image add that enhances use of devcontainers?

.devcontainer/Dockerfile
neersighted, line 9: We don't really want to depend on the script here, and we're not actively encourage it's use either. Instead a manual install flow is perfectly suitable, in my opinion:

```suggestion
ENV POETRY_HOME=/opt/poetry
RUN python3 -m venv $POETRY_HOME && \
    $POETRY_HOME/bin/pip install poetry

ENV VIRTUAL_ENV=/opt/poetry-env
RUN python3 -m venv $VIRTUAL_ENV  
    
ENV PATH="$VIRTUAL_ENV/bin:$POETRY_HOME/bin:$PATH"
```

This way we have a virtualenv holding Poetry (like the script creates, but explicitly controlled), and a separate virtual env for the project. This is a generic template following our recommendations that can be duplicated by other projects making use of Poetry + Dev Containers as well.

.devcontainer/setup.sh
```diff
@@ -0,0 +1,4 @@
...
+poetry run pytest
+poetry run mypy
+poetry run pre-commit install
```
neersighted, line 4: We should drop these lines; if the project is in a failing state we shouldn't cause issues with Dev Container creation... Imagine if a user modifies some files such that ruff is not happy, and then tries to activate Dev Containers with their editor -- now they have to stash or similar to avoid a failure here (and not all users will understand that).

neersighted (changes requested):

.devcontainer/Dockerfile
neersighted: It looks like this might relate to https://github.com/devcontainers/images/issues/885 (not sure how VSCode wires that up -- maybe a script that emits an escape sequence?)... I'd rather let upstream figure this out (if we keep the Microsoft image) than try and solve it downstream.

PR #8027, opened 2023-05-28
feat: perform a single hash validation of wheel files

ralbertazzi:
Re-introduce wheel content file validation in a more performant way, aka computing (and validating) hash _once_ while installing the wheel (and verifying the wheel at the end of the installation process).

Related to #7983 and #7987

- [ ] Address https://github.com/python-poetry/poetry/pull/7987 comments

### DIFF
--- a/src/poetry/installation/wheel_installer.py
+++ b/src/poetry/installation/wheel_installer.py
@@ -5,9 +5,13 @@ import sys
 
 from pathlib import Path
 from typing import TYPE_CHECKING
+from typing import Collection
+from typing import Iterable
 
 from installer import install
 from installer.destinations import SchemeDictionaryDestination
+from installer.records import RecordEntry
+from installer.records import parse_record_file
 from installer.sources import WheelFile
 from installer.sources import _WheelFileValidationError
 
@@ -18,7 +22,6 @@ from poetry.utils._compat import WINDOWS
 if TYPE_CHECKING:
     from typing import BinaryIO
 
-    from installer.records import RecordEntry
     from installer.scripts import LauncherKind
     from installer.utils import Scheme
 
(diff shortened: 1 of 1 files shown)


dimbleby:
I do not like the amount of pypa/installer that is being duplicated by poetry already, and this isn't helping!

Are you sure you wouldn't prefer to fix up your MR in that project - I see that the pipeline is currently failing - and wait for it to be released?  Then just _use_ pypa/installer instead of re-implementing it?

src/poetry/installation/wheel_installer.py
```diff
@@ -66,12 +87,52 @@ def for_source(self, source: WheelFile) -> WheelDestination:
...
+        }
+        for item in self._source._zipfile.infolist():
+            record_args = record_mapping.pop(item.filename, None)
+            if not record_args:
+                continue
+
+            file_record = RecordEntry.from_elements(*record_args)
+            computed_record = next(
+                record for _, record in records if record.path == item.filename
+            )
```
radoering, line 112: This doesn't always work. I tested a large project and got a `StopIteration` for some packages.

You can reproduce it by running `poetry add fonttools` for example.

What's even worth: The package is installed, but the `RECORD` file is not written.

`poetry run pip uninstall fonttools` now throws `ERROR: Cannot uninstall fonttools 4.39.4, RECORD file not found. Hint: The package was installed by Poetry 1.6.0.dev0.` and `poetry run pip install fonttools` gives `Requirement already satisfied`.

You can only "manually" uninstall the package.

src/poetry/installation/wheel_installer.py
```diff
@@ -66,12 +87,52 @@ def for_source(self, source: WheelFile) -> WheelDestination:
...
+                )
+
+    def finalize_installation(
+        self,
+        scheme: Scheme,
+        record_file_path: str,
+        records: Iterable[tuple[Scheme, RecordEntry]],
+    ) -> None:
+        self._validate_hash_and_size(records)
+        return super().finalize_installation(scheme, record_file_path, records)
```
radoering, line 134:
```suggestion
        try:
            self._validate_hash_and_size(records)
        finally:
            return super().finalize_installation(scheme, record_file_path, records)
```

I wonder if we should do something like that to make sure the `RECORD` file is written despite of bugs in `_validate_hash_and_size()`. However, when I tried this, the exception was swallowed, which not good either. Thus, we would have to put more effort here...

radoering (changes requested):
I have copied my comments from the other PR so they will not be forgotten.

I agree that we should prefer a solution in installer if possible. Another approach to avoid calculating the hashes twice might be to add a "use hashes from `RECORD` file" parameter to the `install()` function that can be set if validation was successful.

By the way, what's the performance penalty for calculating the hashes twice for a big wheel like torch?

Issue #7459, opened 2023-02-02
poetry install hugely slowed down by deepcopy operations


ralbertazzi:
- **Poetry version**: 1.3.2
- **Python version**: 3.9.11
- **OS version and name**: MacOS 12.5.1
- **pyproject.toml**: not shared

- [x] I am on the [latest](https://github.com/python-poetry/poetry/releases/latest) stable Poetry version, installed using a recommended method.
- [x] I have searched the [issues](https://github.com/python-poetry/poetry/issues) of this repo and believe that this is not a duplicate.
- [x] I have consulted the [FAQ](https://python-poetry.org/docs/faq/) and [blog](https://python-poetry.org/blog/) for any relevant entries or release notes.
- [x] If an exception occurs when executing a command, I executed it again in debug mode (`-vvv` option) and have included the output below.

## Issue

`poetry install` takes considerable time (70 seconds) to run what should be a no-op. In fact:
- All packages to be installed are already cached
- All packages are already installed in the virtual environment. The command terminates with `No dependencies to install or update`

Although I cannot share the `pyproject.toml` and `poetry.lock` files as they contain private dependencies (so you wouldn't be able to reproduce it), I can share a profiling of `poetry` that obtained running cProfile ([link here](https://github.com/ralbertazzi/poetry/blob/master/test.profile)). I'll share a few links that obtained plotting with `snakeviz` the result of
> /path/to/pypoetry/venv/bin/python" -m cProfile -o test.profile -m poetry install

As you can see lots of time seems to be spent in `deepcopy` operations in `poetry-core`.

I'll try to create a fully reproducible use case. In the meantime I can also share that:
- The project contains a great number of dependencies (around 200), some of them related to ML (PyTorch, TensorFlow, ..).
- The same issue happens in our CI systems running on Linux
- I experienced the same issue on both Poetry 1.2.2 ad Poetry 1.3.2

Thanks a lot for your support and for maintaining this amazing tool ❤️ 

<img width="951" alt="Screenshot 2023-02-02 at 15 40 25" src="https://user-images.githubusercontent.com/8771599/216355294-e7ec95bb-fde7-49ec-8b9d-5d093ef57f51.png">
<img width="951" alt="Screenshot 2023-02-02 at 15 40 41" src="https://user-images.githubusercontent.com/8771599/216355317-41eb52cc-61ea-4b6b-ab4b-815544d1dda4.png">
[...]

PR #8031, opened 2023-05-29
perf: remove unnecessary deepcopy

ralbertazzi:
Related to: https://github.com/python-poetry/poetry/issues/7459

This removes an unnecessary `deepcopy` of a `Package` once returned by `Repository`. The object returned is in fact [used in a read-only way](https://github.com/python-poetry/poetry/blob/198d6ff8caf22215f85e055578fa4b6c4526d061/src/poetry/puzzle/provider.py#L461).
Here's the performance comparison by timing the [solver](https://github.com/python-poetry/poetry/blob/198d6ff8caf22215f85e055578fa4b6c4526d061/src/poetry/installation/installer.py#L312) on a `poetry install --dry-run`:
- A mildly complex project where the Solver still has to perform a good amount of calculations before installing dependencies: from `3.4s` to `2.4s`
- Poetry itself: from `0.1s` to `0.07s`

That's about a 30% performance boost.

Feedback welcome on how to possible test this 🤷 Although I think the long-run solution would be to refactor the `poetry-core` classes in a more read-only / frozen way

### DIFF
--- a/src/poetry/repositories/repository.py
+++ b/src/poetry/repositories/repository.py
@@ -115,6 +115,6 @@ class Repository(AbstractRepository):
         canonicalized_name = canonicalize_name(name)
         for package in self.packages:
             if canonicalized_name == package.name and package.version == version:
-                return package.clone()
+                return package
 
         raise PackageNotFound(f"Package {name} ({version}) not found.")


radoering:
The more complex the project, the smaller the effect. Tested with a larger project: `17 s` -> `15 s`

Nevertheless, probably significant enough to give up the additional safety net.

> Feedback welcome on how to possible test this 🤷 Although I think the long-run solution would be to refactor the `poetry-core` classes in a more read-only / frozen way

I don't think a sensible test is possible. We won't realize it until we do write at some point.

Secrus:
Ok, let's merge it and hope we don't break something :laughing:

Secrus (approved):

Issue #6091, opened 2022-07-29
Poetry should write .pth files in sys.path with the locale-defined encoding


shoucandanghehe:
- [x] I am on the [latest](https://github.com/python-poetry/poetry/releases/latest) Poetry version.
- [x] I have searched the [issues](https://github.com/python-poetry/poetry/issues) of this repo and believe that this is not a duplicate.
- [x] If an exception occurs when executing a command, I executed it again in debug mode (`-vvv` option).

- **OS version and name**:  Windows 11 22H2 22622.436
- **Poetry version**:  1.1.14
- **Link of a [Gist](https://gist.github.com/) with the contents of your pyproject.toml file**:  https://gist.github.com/shoucandanghehe/84fa45ddecba402d284bd22e7e141c11

## Issue

This looks similar to #5985, but it should not be the same error
I tried removing all python environments and reinstalling poetry, but it didn't work
Below is the error log
```
nonebot-plugin-tetris-stats on  main [!⇡] via 🐍 v3.10.5 
❯ poetry install -vvv

  Stack trace:

  12  ~\AppData\Roaming\pypoetry\venv\lib\site-packages\clikit\console_application.py:131 in run
       129│             parsed_args = resolved_command.args
       130│
     → 131│             status_code = command.handle(parsed_args, io)
       132│         except KeyboardInterrupt:
       133│             status_code = 1

  11  ~\AppData\Roaming\pypoetry\venv\lib\site-packages\clikit\api\command\command.py:120 in handle
       118│     def handle(self, args, io):  # type: (Args, IO) -> int
       119│         try:
     → 120│             status_code = self._do_handle(args, io)
       121│         except KeyboardInterrupt:
       122│             if io.is_debug():

  10  ~\AppData\Roaming\pypoetry\venv\lib\site-packages\clikit\api\command\command.py:163 in _do_handle
       161│         if self._dispatcher and self._dispatcher.has_listeners(PRE_HANDLE):
       162│             event = PreHandleEvent(args, io, self)
     → 163│             self._dispatcher.dispatch(PRE_HANDLE, event)
       164│
       165│             if event.is_handled():

   9  ~\AppData\Roaming\pypoetry\venv\lib\site-packages\clikit\api\event\event_dispatcher.py:22 in dispatch
        20│
        21│         if listeners:
     →  22│             self._do_dispatch(listeners, event_name, event)
        23│
        24│         return event

[...]

PR #8041, opened 2023-05-30
Encode using locale when writing .pth files

savannahostrowski:
# Pull Request Check List

Resolves: #6091 

~- [ ] Updated **documentation** for changed code.~ Not required

From what I could glean from the issue/reading through the code, fixing this issue looks like a 1-line change. Re. testing, I wasn't really sure how to explicitly test this beyond just updating the existing `file.open()` in an existing test.


savannahostrowski:
@python-poetry/triage Happy to address any feedback here :)

radoering:
The test only tests the change if you check out the poetry repo in a folder with a letter that is different from utf-8 in the preferred encoding (like "ö"). To test it reproducibly, we probably had to create a directory with such a letter in tests/fixtures. However, I don't know if this causes issues for systems with a preferred encoding that does not contain this letter.

Thus, maybe not worth it. I think I'll just wait a couple of days if there are other opinions/ideas, and if not, merge it as is.

radoering (approved):

PR #8045, opened 2023-05-30
feat: add ARG001 rule

ralbertazzi:
As this affects the whole codebase it would be great to have a quick feedback cycle to avoid lots of painful rebases 🙏 

This kind of rules doesn't obviously play well with pytest fixtures, that's why many inline ignores. On the other side, it helped detecting **lots** of unneeded fixtures.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -102,6 +102,7 @@ unfixable = [
 target-version = "py38"
 line-length = 88
 extend-select = [
+    "ARG", # flake8-unused-arguments
     "B",   # flake8-bugbear
     "C4",  # flake8-comprehensions
     "ERA", # flake8-eradicate/eradicate
@@ -127,6 +128,12 @@ extend-exclude = [
     # External to the project's coding standards
     "tests/**/fixtures/*",
 ]
+
+[tool.ruff.extend-per-file-ignores]
+"tests/installation/test_chooser.py" = ["ARG001"]      # rule gets too noisy
+"tests/integration/test_utils_vcs_git.py" = ["ARG001"] # rule gets too noisy
+"tests/utils/test_authenticator.py" = ["ARG001"]       # rule gets too noisy
+"tests/utils/test_password_manager.py" = ["ARG001"]    # rule gets too noisy
 
 [tool.ruff.flake8-tidy-imports]
 ban-relative-imports = "all"
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -102,7 +102,7 @@ unfixable = [
 target-version = "py38"
 line-length = 88
 extend-select = [
-    "ARG001", # flake8-unused-arguments - Rules other than ARG001 result in false positives only
+    "ARG", # flake8-unused-arguments
     "B",   # flake8-bugbear
     "C4",  # flake8-comprehensions
     "ERA", # flake8-eradicate/eradicate
@@ -130,7 +130,7 @@ extend-exclude = [
 ]
 
 [tool.ruff.extend-per-file-ignores]
-"tests/*.py" = ["ARG001"] # rule gets too noisy with pytest fixtures
+"tests/*.py" = ["ARG"] # rule gets too noisy with pytest fixtures and mocks
 
 [tool.ruff.flake8-tidy-imports]
 ban-relative-imports = "all"
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -102,7 +102,7 @@ unfixable = [
 target-version = "py38"
 line-length = 88
 extend-select = [
-    "ARG", # flake8-unused-arguments
+    "ARG001", # flake8-unused-arguments - Rules other than ARG001 result in false positives only
     "B",   # flake8-bugbear
     "C4",  # flake8-comprehensions
     "ERA", # flake8-eradicate/eradicate
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -130,10 +130,7 @@ extend-exclude = [
 ]
 
 [tool.ruff.extend-per-file-ignores]
-"tests/installation/test_chooser.py" = ["ARG001"]      # rule gets too noisy
-"tests/integration/test_utils_vcs_git.py" = ["ARG001"] # rule gets too noisy
-"tests/utils/test_authenticator.py" = ["ARG001"]       # rule gets too noisy
-"tests/utils/test_password_manager.py" = ["ARG001"]    # rule gets too noisy
+"tests/*.py" = ["ARG001"] # rule gets too noisy with pytest fixtures
 
 [tool.ruff.flake8-tidy-imports]
 ban-relative-imports = "all"
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -360,7 +360,6 @@ def project_factory(
     repo: TestRepository,
     installed: Repository,
     default_python: str,
-    load_required_fixtures: None,
 ) -> ProjectFactory:
     workspace = tmp_path
 
--- a/tests/console/commands/cache/test_list.py
+++ b/tests/console/commands/cache/test_list.py
@@ -18,7 +18,7 @@ def tester(command_tester_factory: CommandTesterFactory) -> CommandTester:
 
 def test_cache_list(
     tester: CommandTester,
-    mock_caches: None,  # noqa: ARG001
+    mock_caches: None,
     repository_one: str,
     repository_two: str,
 ) -> None:
--- a/tests/console/commands/cache/test_list.py
+++ b/tests/console/commands/cache/test_list.py
@@ -6,8 +6,6 @@ import pytest
 
 
 if TYPE_CHECKING:
-    from pathlib import Path
-
     from cleo.testers.command_tester import CommandTester
 
     from tests.types import CommandTesterFactory
@@ -19,7 +17,10 @@ def tester(command_tester_factory: CommandTesterFactory) -> CommandTester:
 
 
 def test_cache_list(
-    tester: CommandTester, mock_caches: None, repository_one: str, repository_two: str
+    tester: CommandTester,
+    mock_caches: None,  # noqa: ARG001
+    repository_one: str,
+    repository_two: str,
 ) -> None:
     tester.execute()
 
@@ -31,7 +32,7 @@ def test_cache_list(
     assert tester.io.fetch_output() == expected
 
 
-def test_cache_list_empty(tester: CommandTester, repository_cache_dir: Path) -> None:
+def test_cache_list_empty(tester: CommandTester) -> None:
     tester.execute()
 
     expected = """\
--- a/tests/console/commands/env/helpers.py
+++ b/tests/console/commands/env/helpers.py
@@ -22,7 +22,7 @@ def build_venv(path: Path | str, **_: Any) -> None:
 def check_output_wrapper(
     version: Version = VERSION_3_7_1,
 ) -> Callable[[list[str], Any, Any], str]:
-    def check_output(cmd: list[str], *args: Any, **kwargs: Any) -> str:
+    def check_output(cmd: list[str], *_: Any, **__: Any) -> str:
         # cmd is a list, like ["python", "-c", "do stuff"]
         python_cmd = cmd[2]
         if "sys.version_info[:3]" in python_cmd:
--- a/tests/console/commands/self/conftest.py
+++ b/tests/console/commands/self/conftest.py
@@ -17,8 +17,6 @@ from poetry.utils.env import EnvManager
 if TYPE_CHECKING:
     from collections.abc import Iterable
 
-    import httpretty
-
     from cleo.io.io import IO
     from pytest_mock import MockerFixture
 
@@ -43,12 +41,7 @@ def pool(repo: TestRepository) -> RepositoryPool:
 def create_pool_factory(
     repo: Repository,
 ) -> Callable[[Config, Iterable[dict[str, Any]], IO, bool], RepositoryPool]:
-    def _create_pool(
-        config: Config,
-        sources: Iterable[dict[str, Any]] = (),
-        io: IO | None = None,
-        disable_cache: bool = False,
-    ) -> RepositoryPool:
+    def _create_pool(*_: Any, **__: Any) -> RepositoryPool:
         pool = RepositoryPool()
         pool.add_repository(repo)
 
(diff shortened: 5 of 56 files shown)


tests/helpers.py
```diff
@@ -156,13 +156,13 @@ def _do_execute_operation(self, operation: Operation) -> int:
 
         return rc
 
-    def _execute_install(self, operation: Operation) -> int:
+    def _execute_install(self, _: Operation) -> int:
```
radoering, line 159: I don't like these changes. IMO if a method is overridden, the names of arguments shouldn't be changed.

tests/conftest.py
```diff
@@ -360,7 +360,6 @@ def project_factory(
     repo: TestRepository,
     installed: Repository,
     default_python: str,
-    load_required_fixtures: None,
```
radoering, line 363: Is this really unused? Then, we don't need `load_required_fixtures()` at all. Although, `required_fixtures()` returns an empty list here, it's overridden in `test_add.py` for example. Apparently, the tests don't fail, so maybe we can just remove all of this?

radoering: On the other hand it might be dangerous to remove this because it was introduced in  #5745. Thus, a passing test is not prove enough because it probably avoids race conditions.

tests/helpers.py
finswimmer: > IMO if a method is overridden, the names of arguments shouldn't be changed.

Glad to see, there's someone with the same opinion than me :smile:

tests/conftest.py
ralbertazzi: I tried removing more than this and tests for the add and show commands started in fact breaking. I didn't experience any race condition while removing this line though and re-running the whole suite several times. But if you think it's best to keep it I won't complain about an extra line, just say so :)

tests/helpers.py
ralbertazzi: I generally agree with you, I mustn't have noticed that this was overridden! To be fair, there are already some instances on this around the poetry codebase..

ralbertazzi: I must have changed it when I initially enabled all ARG rules, and then I reverted everything other than ARG001 because others were in fact too noisy when overriding methods

Secrus:
Honestly, I don't like this. The amount of false positives that we have to silence is just too high for my taste. I am all for doing some cleaning, but to keep this rule enabled in the long run, not really.

finswimmer:
> Honestly, I don't like this. The amount of false positives that we have to silence is just too high for my taste. I am all for doing some cleaning, but to keep this rule enabled in the long run, not really.

Maybe a good compromise would  be to activate this rule only for the package code, but not for the tests?

Secrus:
> Maybe a good compromise would be to activate this rule only for the package code, but not for the tests?

For me, tests are an integral part of code, so treating them differently would be weird imho. 

Disclaimer: I won't block this PR if others think it's a good change. Just sounding out my concerns.

ralbertazzi:
I may have missed an obvious thing, but I didn't find a way to disable rules on entire folders on ruff 🤔 
If you don't like having the rule enabled, one solution could also be to revert the enabling while still merging the improvements that it brought

tests/conftest.py
radoering: Ah, `load_required_fixtures()` is marked with `autouse=True`, which means this change makes sense. 😀

radoering:
> Honestly, I don't like this. The amount of false positives that we have to silence is just too high for my taste. I am all for doing some cleaning, but to keep this rule enabled in the long run, not really.

That was actually my first thought as well.

> one solution could also be to revert the enabling while still merging the improvements that it brought

Sounds good to me.

radoering:
> Maybe a good compromise would be to activate this rule only for the package code, but not for the tests?

Would also be ok for me.

> I may have missed an obvious thing, but I didn't find a way to disable rules on entire folders on ruff 🤔

Did you try file patterns with `*`?

ralbertazzi:
Silly me.. it's actually possible to disable the check on the entire `tests` folder. To summarize:
- We keep `ARG001` enabled on the main codebase
- We keep `ARG001` disabled on `tests`, while still merging the cleanup performed while keeping this rule enabled

We now have a total of 2 additional `# noqa` pragmas in total. Seems small enough to call it not noisy :)

pyproject.toml
```diff
@@ -102,6 +102,7 @@ unfixable = [
 target-version = "py38"
 line-length = 88
 extend-select = [
+    "ARG001", # flake8-unused-arguments - Rules other than ARG001 result in false positives only
```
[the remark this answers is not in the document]
ralbertazzi: Sorry, I might have explained myself incorrectly. They are true positives but provide meaningless action (like a #noqa pragma), so I don't see value in keeping them enabled.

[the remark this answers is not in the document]
dimbleby: Also I think the warnings can mostly be suppressed by using an `_` for the unused parameter name, which hardly clutters the code at all - makes it clearer, if anything.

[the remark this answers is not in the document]
ralbertazzi: This is kind of the opposite of https://github.com/python-poetry/poetry/pull/8045#discussion_r1211081164 though. Hard to find consensus 😄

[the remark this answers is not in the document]
dimbleby: personally I find `_operation` as a convention meaning "`operation`, but not used" to be more than acceptable - it preserves enough of the original to be not confusing, it acknowledges that the value is not used and it makes clear that someone has thought about it and had some reason to leave the unused parameter there anyway.

The exact means by which warnings for intentional unused parameters is secondary though: my main point is that _if_ we think this is a valuable warning, then it is strange to think that it is only valuable in certain types of function.  For me, it should be all of `ARG`, or none.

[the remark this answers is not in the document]
radoering: > personally I find `_operation` as a convention meaning "`operation`, but not used" to be more than acceptable

I don't think so. The parameter name is part of the interface (except for positional-only parameters). Changing the name in an overridden method changes the interface and thereby hurts the Liskov substitution principle.

> my main point is that _if_ we think this is a valuable warning, then it is strange to think that it is only valuable in certain types of function.

On first thought: yes. However, Liskov is not relevant for functions. That's why there will be less "false positives" than for methods and so on. IMO, it's perfectly valid to say `ARG001` is the only one that does not do more damage than good. As already mentioned before I'm also fine with just merging thte improvements and not enabling the check at all.

ralbertazzi:
I enabled all ARG rules on the main codebase while keeping them disabled on tests. I'm more in favour of [radoering's comment](https://github.com/python-poetry/poetry/pull/8045#discussion_r1212538158) on preserving names to keep the Liskov principle valid rather than modifying variable names with `_`.

If this version doesn't find consensus yet, then I'll probably revert enabling ARG rules everywhere and just propose the improvements.

tests/console/commands/cache/test_list.py
```diff
@@ -19,7 +17,10 @@ def tester(command_tester_factory: CommandTesterFactory) -> CommandTester:
 
 
 def test_cache_list(
-    tester: CommandTester, mock_caches: None, repository_one: str, repository_two: str
```
radoering, line 22: This change can be reverted.

tests/console/commands/env/helpers.py
```diff
@@ -22,7 +22,7 @@ def build_venv(path: Path | str, **_: Any) -> None:
 def check_output_wrapper(
     version: Version = VERSION_3_7_1,
 ) -> Callable[[list[str], Any, Any], str]:
-    def check_output(cmd: list[str], *args: Any, **kwargs: Any) -> str:
```
radoering, line 25: I think I'd revert that, too

tests/console/commands/self/conftest.py
```diff
@@ -43,12 +41,7 @@ def pool(repo: TestRepository) -> RepositoryPool:
 def create_pool_factory(
     repo: Repository,
 ) -> Callable[[Config, Iterable[dict[str, Any]], IO, bool], RepositoryPool]:
-    def _create_pool(
-        config: Config,
-        sources: Iterable[dict[str, Any]] = (),
-        io: IO | None = None,
-        disable_cache: bool = False,
-    ) -> RepositoryPool:
+    def _create_pool(*_: Any, **__: Any) -> RepositoryPool:
```
radoering, line 44: Can we have `*args: Any, **kwargs: Any` here, too?

tests/console/test_application.py
```diff
@@ -40,7 +40,9 @@ def with_add_command_plugin(mocker: MockerFixture) -> None:
     mock_metadata_entry_points(mocker, AddCommandPlugin)
 
 
-def test_application_with_plugins(with_add_command_plugin: None) -> None:
+def test_application_with_plugins(
```
radoering, line 43: Can be reverted.

tests/console/test_application.py
```diff
@@ -50,7 +52,9 @@ def test_application_with_plugins(with_add_command_plugin: None) -> None:
     assert tester.status_code == 0
 
 
-def test_application_with_plugins_disabled(with_add_command_plugin: None) -> None:
+def test_application_with_plugins_disabled(
```
radoering, line 55: Can be reverted.

tests/console/test_application.py
```diff
@@ -60,7 +64,9 @@ def test_application_with_plugins_disabled(with_add_command_plugin: None) -> Non
     assert tester.status_code == 0
 
 
-def test_application_execute_plugin_command(with_add_command_plugin: None) -> None:
+def test_application_execute_plugin_command(
```
radoering, line 67: Can be reverted.

tests/installation/test_chooser.py
```diff
@@ -46,7 +46,7 @@ def env() -> MockEnv:
 @pytest.fixture()
 def mock_pypi(http: type[httpretty.httpretty]) -> None:
     def callback(
-        request: HTTPrettyRequest, uri: str, headers: dict[str, Any]
+        _: HTTPrettyRequest, uri: str, headers: dict[str, Any]
```
radoering, line 49: revert

tests/installation/test_chooser.py
```diff
@@ -73,7 +73,7 @@ def callback(
 @pytest.fixture()
 def mock_legacy(http: type[httpretty.httpretty]) -> None:
     def callback(
-        request: HTTPrettyRequest, uri: str, headers: dict[str, Any]
+        _: HTTPrettyRequest, uri: str, headers: dict[str, Any]
```
radoering, line 76: revert

tests/installation/test_chooser.py
```diff
@@ -93,7 +93,7 @@ def callback(
 @pytest.fixture()
 def mock_legacy_partial_yank(http: type[httpretty.httpretty]) -> None:
     def callback(
-        request: HTTPrettyRequest, uri: str, headers: dict[str, Any]
+        _: HTTPrettyRequest, uri: str, headers: dict[str, Any]
```
radoering, line 96: revert

tests/puzzle/test_provider.py
```diff
@@ -37,7 +37,7 @@
 
 
 class MockEnv(BaseMockEnv):
-    def run(self, bin: str, *args: str, **kwargs: Any) -> str:
+    def run(self, *_: str, **__: Any) -> str:
```
radoering, line 40: revert or `*args: str, **kwargs: Any`

tests/repositories/test_single_page_repository.py
```diff
@@ -34,7 +35,7 @@ def _get_page(self, name: NormalizedName) -> SimpleRepositoryPage:
         with fixture.open(encoding="utf-8") as f:
             return SimpleRepositoryPage(self._url, f.read())
 
-    def _download(self, url: str, dest: Path) -> None:
+    def _download(self, *_: Any, **__: Any) -> None:
```
radoering, line 38: revert or `*args: Any, **kwargs: Any`

tests/utils/test_env.py
```diff
@@ -190,7 +190,7 @@ def build_venv(path: Path | str, **__: Any) -> None:
 def check_output_wrapper(
     version: Version = VERSION_3_7_1,
 ) -> Callable[[list[str], Any, Any], str]:
-    def check_output(cmd: list[str], *args: Any, **kwargs: Any) -> str:
+    def check_output(cmd: list[str], *_: Any, **__: Any) -> str:
```
radoering, line 193: revert

tests/utils/test_env.py
```diff
@@ -936,7 +916,7 @@ def test_remove_keeps_dir_if_not_deleteable(
         side_effect=check_output_wrapper(Version.parse("3.6.6")),
     )
 
-    def err_on_rm_venv_only(path: Path, *args: Any, **kwargs: Any) -> None:
+    def err_on_rm_venv_only(path: Path, *_: Any, **__: Any) -> None:
```
radoering, line 919: revert

tests/utils/test_env.py
```diff
@@ -1626,7 +1586,7 @@ def test_create_venv_accepts_fallback_version_w_nonzero_patchlevel(
 
     poetry.package.python_versions = "~3.5.1"
 
-    def mock_check_output(cmd: str, *args: Any, **kwargs: Any) -> str:
+    def mock_check_output(cmd: str, *_: Any, **__: Any) -> str:
```
radoering, line 1589: revert

radoering (changes requested):
IMO the current version is fine. We have only 9 `# noqa: ARG...` now. If we find that the rules are too annoying, we can still disable them in the future.

I'd just revert some changes in the tests that don't make sense anymore (formatting and `_` / `__` arguments).

tests/installation/test_chooser.py
silviamichele: yep, request is better to describe the param

Secrus:
@ralbertazzi are you still interested in this? If not, please close.

abn:
@ralbertazzi thank you for taking the time to work on this, propose this, and worth through feedback. However, at this point I do not think we would want this change unfortunately. In a code base like Poetry, the the noise a rule like this produces is quite high. Things might change in the future, but considering how most of this code base is written, I would not encourage this rule being activated.

For now closing. Once again, we really appreciate your time and effort.

ralbertazzi:
No worries, thank you for your steady contributions :)

Issue #7936, opened 2023-05-16
 'HTTPResponse' object has no attribute 'strict'


aakashrshah:
- **Poetry version**: Poetry (version 1.4.2)
- **Python version**: Python:  3.10.10
- **OS version and name**: macOS 13.2
- **pyproject.toml**: https://gist.github.com/aakashrshah/16d755f1bfacdcee1bb2cd2a66a409f0

- [x] I am on the [latest](https://github.com/python-poetry/poetry/releases/latest) stable Poetry version, installed using a recommended method.
- [x] I have searched the [issues](https://github.com/python-poetry/poetry/issues) of this repo and believe that this is not a duplicate.
- [x] I have consulted the [FAQ](https://python-poetry.org/docs/faq/) and [blog](https://python-poetry.org/blog/) for any relevant entries or release notes.
- [x] If an exception occurs when executing a command, I executed it again in debug mode (`-vvv` option) and have included the output below.

## Issue
```

  AttributeError

  'HTTPResponse' object has no attribute 'strict'

  at ~/Library/Caches/pypoetry/virtualenvs/rubber-duck-py3.10/lib/python3.10/site-packages/cachecontrol/serialize.py:54 in dumps
       50│                 ),
       51│                 u"status": response.status,
       52│                 u"version": response.version,
       53│                 u"reason": text_type(response.reason),
    →  54│                 u"strict": response.strict,
       55│                 u"decode_content": response.decode_content,
       56│             }
       57│         }
       58│ 
```

I am not able to install or add packages and blocked significantly. I do not see this issue on internet. Am i using a wrong python version? this is the latest version of poetry , shall i revert?

dimbleby:
this is https://github.com/ionrock/cachecontrol/issues/292

however since poetry requirements do not allow urllib 2.0, this shows that you, or whatever you used to install poetry, have upgraded something that you should not have done

please close

dimbleby:
PS I do not know why you write here "I do not see this issue on internet." when, as I now see, you have already commented on that other issue

Secrus:
Poetry has `urllib3` pinned below 2.0, so you had to mess something up on your end (for example by installing your project and poetry into the same environment). In any case, this is not Poetry's fault.

aakashrshah:
Good by me if we want to close it. Currently I removed the line with strict in cache control. 

I tried it on python 3.10, 3.11 with latest poetry and latest pip.

I don’t know what did I mess up on my end. 

Thanks though.

[...]

PR #8055, opened 2023-06-01
feat: update cachecontrol to 0.13.0

ralbertazzi:
# Pull Request Check List

Closes: https://github.com/python-poetry/poetry/pull/7997

@Secrus didn't want to step on your toes, just provide help :)


popadi:
Initially I hoped that a new cachecontrol release will silently fix the issue described [here](https://github.com/python-poetry/poetry/issues/7936) (without having to update to a newer version of poetry OR to pin urllib3). However, because the toml file has `^0.12.9` for all releases and because the new cachecontrol was release as `0.13.0`, this won't happen. 

What will be the poetry versions that will be lucky to benefit from this?

ralbertazzi:
> without having to update to a newer version of poetry OR to pin urllib3

I don't think anyone will release a fix for older versions of Poetry due to the limited time that maintainers can invest. The issue has been closed since fixed in latest versions, please update to 1.5

dimbleby:
recent versions of poetry all pin to urrlib3<2, therefore if correctly installed they are not exposed to that issue anyway.

however this reminds me that this MR should allow that constraint to be relaxed, urllib3 version 3 is now out.

ralbertazzi:
Could you point me to the release? I can only see 2.0.2

dimbleby:
yes sorry, I meant 2.  but either way, this MR should allow that to be relaxed

pyproject.toml
```diff
@@ -60,7 +58,6 @@ tomlkit = ">=0.11.4,<1.0.0"
 trove-classifiers = ">=2022.5.19"
 virtualenv = "^20.22.0"
 xattr = { version = "^0.10.0", markers = "sys_platform == 'darwin'" }
-urllib3 = "^1.26.0"
```
dimbleby, line 63: poetry still has a direct dependency on `urllib3` - https://github.com/python-poetry/poetry/blob/50e61e696bb0ae61c148d570f1c204a667464f1c/src/poetry/publishing/uploader.py#L20 - so this should not be removed.

ralbertazzi: Seemed really minor for me to be kept but I understand the reasoning, better to be safe than sorry!

ralbertazzi: Would you prefer to relax the constraint or pin it to `^2.0.2` to enforce the usage of a more recent version? I'd prefer the second but I also understand the Poetry direct dependency on urllib3 works ok with <2 too..

dimbleby: so far as I can see that import is only used to set up a `Retry`, which is configured to retry only for GETs; and the uploader never does a GET, it only POSTs.

So if that's right then that little pile of code could be deleted anyway.   But double-check, I only skimmed it..!

ralbertazzi: I think you are totally right. Wouldn't it be better to then fix it to "POST" if the original code was to retry uploads?

dimbleby: meh, if the code has been there for this long without anyone feeling the need to have working retries - I'd just delete it.

ralbertazzi: I was just writing that 😄 Makes sense

Secrus (approved):

frostming:
> However, because the toml file has `^0.12.9` for all releases and because the new cachecontrol was release as `0.13.0`, this won't happen.

I've released v0.12.13 with the necessary fixes

PR #8063, opened 2023-06-05
perf: let Transaction.calculate_operations run in O(N)

ralbertazzi:
Hard to call it a `perf` commit, as we are talking about a few milliseconds, but I think it also makes the code more Pythonic

### DIFF
--- a/src/poetry/puzzle/transaction.py
+++ b/src/poetry/puzzle/transaction.py
@@ -31,7 +31,7 @@ class Transaction:
         if installed_packages is None:
             installed_packages = []
 
-        self._installed_packages = installed_packages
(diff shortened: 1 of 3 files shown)


radoering (approved):
Rebased on current main with some minor adaptions:
I only adopted the change for `_installed_packages` because that is the only list without duplicate package names. Building a `dict` of `_current_packages` and `_result_packages` is dangerous because it drops distinct entries with the same name. Further, there is no need for a `dict` for `_current_packages` and (with changes on the main branch, which happened in the meantime) there is no need for a `dict` for `_result_packages` anymore.

I agree that making `_installed_packages` a `dict` improves maintainability.

PR #8078, opened 2023-06-08
Access keyring only when needed.

### LINKED ISSUE #1917: Keyring errors during non-publishing operations

real-yfprojects:
# Pull Request Check List

Resolves: #1917

***

Requesting a network page is first tried without retrieving authentication from the keyring. Credentials provided by other means are used already. If no authentication was found and the server answers with 401 or 403 the keyring is queried for credentials. This fixes #1917.

* src/poetry/utils/constants.py (STATUS_AUTHLIST): A list of HTTP errors requesting authentication. Currently 401 and 403.

* src/poetry/utils/authenticator.py (Authenticator.request): Retrieve credentials with disabled keyring first. On auth error try again with enabled keyring.

* src/poetry/utils/authenticator.py (Authenticator.get_credentials_for_url): Add `keyring` option.
* src/poetry/utils/authenticator.py (Authenticator._get_credentials_for_url: ^^^ same as above
* src/poetry/utils/authenticator.py (Authenticator._get_credentials_for_repository): ^^^
* src/poetry/utils/authenticator.py (AuthenticatorRepositoryConfig.get_http_credentials): ^^^
* src/poetry/utils/password_manager.py (PasswordManager.get_http_auth): ^^^

### DIFF
--- a/src/poetry/utils/authenticator.py
+++ b/src/poetry/utils/authenticator.py
@@ -18,11 +18,13 @@ import requests.exceptions
 
 from cachecontrol import CacheControlAdapter
 from cachecontrol.caches import FileCache
+from requests import Request
 
 from poetry.config.config import Config
 from poetry.exceptions import PoetryException
 from poetry.utils.constants import REQUESTS_TIMEOUT
 from poetry.utils.constants import RETRY_AFTER_HEADER
+from poetry.utils.constants import STATUS_AUTHLIST
 from poetry.utils.constants import STATUS_FORCELIST
 from poetry.utils.password_manager import HTTPAuthCredential
 from poetry.utils.password_manager import PasswordManager
@@ -80,14 +82,17 @@ class AuthenticatorRepositoryConfig:
         return [self.url, self.netloc, self.name]
 
     def get_http_credentials(
-        self, password_manager: PasswordManager, username: str | None = None
+        self,
+        password_manager: PasswordManager,
+        username: str | None = None,
+        keyring: bool = True,
     ) -> HTTPAuthCredential:
         # try with the repository name via the password manager
         credential = HTTPAuthCredential(
-            **(password_manager.get_http_auth(self.name) or {})
+            **(password_manager.get_http_auth(self.name, keyring=keyring) or {})
         )
 
-        if credential.password is None:
+        if credential.password is None and keyring:
             # fallback to url and netloc based keyring entries
             credential = password_manager.keyring.get_credential(
                 self.url, self.netloc, username=credential.username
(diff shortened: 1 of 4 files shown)


real-yfprojects:
There is already #7037 and #6612 that try to fix this issue. However they seem to take a different approach. Instead of fixing that poetry tries to get credential although it doesn't need them, they only fix that poetry complains about a locked keyring.

real-yfprojects:
Oh I forgot to fix some existing tests. However I am unsure what I should add new tests for?
I also couldn't find any documentation that has to be updated.

real-yfprojects:
I just noticed #7038 does something similar. So what has to be changed so that poetry behaves as you wish?

dimbleby:
imo #6612 and #7037 look like a more pragmatic approach

in particular #6612 doesn't seem like it needs to be much better to be mergeable: catch a slightly more general exception and add a unit test and - noting that I don't have the power to make such decisions - it would look to me to be good to go

real-yfprojects:
Quoting from #7037:

> I think Poetry shouldn't request access to the keyring before the remote server has actually asked for credentials, but this can be done in a subsequent change. https://github.com/python-poetry/poetry/issues/1917 is affecting many users.

That's what I thought too. So I wrote this patch. From my point of view this PR can be merged immediately.

As @remram44 said, fixing poetries elaborate keyring usage is inevitable. I don't want to deny a keyring request every time I run poetry.

real-yfprojects:
> From my point of view this PR can be merged immediately.

I just noticed ci is failing for some reason. Didn't it pass yesterday when I finished work on this PR?

dimbleby:
So far as I can see #7038 was not merged because hitting every API twice - first to try without authentication and then with - was considered not OK.  SInce this does exactly the same I don't know why it wouldn't meet the same fate.

real-yfprojects:
> First to try without authentication and then with - was considered not OK. SInce this does exactly the same I don't know why it wouldn't meet the same fate.

Well then discuss how poetry should behave instead - here or in a separate issue.

dimbleby:
... that's exactly what https://github.com/python-poetry/poetry/pull/8078#issuecomment-1584169367 was doing!

real-yfprojects:
> ... that's exactly what [#8078 (comment)](https://github.com/python-poetry/poetry/pull/8078#issuecomment-1584169367) was doing!

Great! However I didn't notice you were discussing which poetry behaviour would be favourable. You seemed to discuss which PR would be more pragmatic to merge or a better fix for #1917. I would like to separate these two discussions. While #6612 seems better suited for a quick fix and I advocate merging it while I don't mind when this PR isn't merged in its current form, we shouldn't leave sight of the bigger picture. 

Regarding #6612 I will take a look whether I can take the branch, finish the patch and open a new PR.

felix-ht:
> So far as I can see #7038 was not merged because hitting every API twice - first to try without authentication and then with - was considered not OK. SInce this does exactly the same I don't know why it wouldn't meet the same fate.

They main issue is that peotry tries to get a password for any http request from the keyring. It just looks at the keyring and checks if there is a password for the netloc of the current request. This automatic checking happens for any and all http requets. 

There is also no duplicated api requets. If you do not either configure a http_fallback (use the netloc password store in the keyring) or explicitly set a password through poetry. The fetching data from the url will just fail with unauthorised.

Having some poeple properly configure there password proteced routes vs breaking the workflow for most users seems like a very good trade off to me. 

I don't get why nobody seems to care about the most upvoted issue of poetry :/   - i certainly wont be making another contribution

real-yfprojects:
> I don't get why nobody seems to care about 9th most upvoted issues of poetry :/

At least none of the maintainers although this would be exactly where a maintainer has to step in. Instead they seem to dodge the hard decisions.

dimbleby:
per https://github.com/python-poetry/poetry/pull/8078#issuecomment-1584169367: IMO a moderately improved #6612 ought to be mergeable.

(so far as I know no-one has offered such an MR)

real-yfprojects:
> per https://github.com/python-poetry/poetry/pull/8078#issuecomment-1584169367: IMO a moderately improved #6612 ought to be mergeable.
> 
> 
> 
> (so far as I know no-one has offered such an MR)

Well I have offered help to improve set PR. However I stumbled upon an issue I need help with before writing the code. See https://github.com/python-poetry/poetry/pull/6612#issuecomment-1584318886

PR #8090, opened 2023-06-11
docs: make pipx the preferred (first mentioned) installation method

radoering:
As discussed in the chat some time ago, most of the maintainers and frequent contributors use pipx instead of the official installer. There is probably no consensus to deprecate the installer so this PR at least makes pipx the installation method that is mentioned at first in the docs.

In addition to this restructuring, I made the most important thing that has to be considered independent from the installation method - an isolated environment - more prominent. Analogously, I moved the most important CI recommendation - version pinning - from bottom to top.

### DIFF
--- a/docs/_index.md
+++ b/docs/_index.md
@@ -23,18 +23,85 @@ on Linux, macOS and Windows.
 
 ## Installation
 
+{{% warning %}}
+Poetry should always be installed in a dedicated virtual environment to isolate it from the rest of your system.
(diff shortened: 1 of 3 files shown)


Secrus:
Since we are switching to recommending `pipx`, I think alternatives should be mentioned first in plugins docs and the `self` namespace (since it breaks if used on non-installer installations).

Secrus (approved):

PR #8118, opened 2023-06-21
Support non SHA256/SHA384/SHA512 HTTPRepository

gaborbernat:
Resolves https://github.com/python-poetry/poetry/issues/6301

### DIFF
--- a/src/poetry/installation/chooser.py
+++ b/src/poetry/installation/chooser.py
@@ -8,6 +8,7 @@ from typing import Any
 
 from poetry.config.config import Config
 from poetry.config.config import PackageFilterPolicy
+from poetry.repositories.http_repository import HTTPRepository
 from poetry.utils.wheel import Wheel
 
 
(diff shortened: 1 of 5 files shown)


gaborbernat:
@radoering would appreciate if you could take a look here 😊 thanks!

avicennax:
Would it be naive to assume we can easily update / rename this function and leverage `hashlib.md5` with a hash name check?

radoering:
> @radoering would appreciate if you could take a look here 😊 thanks!

LGTM. IMO, after adding a test in `test_chooser.py`, it could be merged.

> Would it be naive to assume we can easily update / rename this function and also support md5 via hashlib.md5 with a hash name check?

See https://github.com/python-poetry/poetry/issues/6301#issuecomment-1604555396

gaborbernat:
@radoering all done now 👍

radoering (approved):

gaborbernat:
@radoering could we get a release cut for this? Thanks!

radoering:
I don't think I'll find the time within the next weeks. Thus, probably not before August.

radoering:
Unfortunately, I can't. You can track the progress in #8232. I'll skim the issues for blockers, review some PRs and wait for some reviews for myself over the next few weeks and afterwards start preparing the release. Thus, my estimation is still August.

threat0day:
@radoering  OK thanks)

Issue #6666, opened 2022-10-01
virtualenv.in-project=true has no effect if project has an existing virtualenv


s7m4b4:
- **Poetry version**: 1.2.1
- **Python version**: 3.10.6
- **OS version and name**: Fedora 36
- https://gist.github.com/s7m4b4/c449b62e3767dfd6c518502a12bf9788

- [x] I am on the [latest](https://github.com/python-poetry/poetry/releases/latest) stable Poetry version, installed using a recommended method.
- [x] I have searched the [issues](https://github.com/python-poetry/poetry/issues) of this repo and believe that this is not a duplicate.
- [x] I have consulted the [FAQ](https://python-poetry.org/docs/faq/) and [blog](https://python-poetry.org/blog/) for any relevant entries or release notes.
- [ ] If an exception occurs when executing a command, I executed it again in debug mode (`-vvv` option) and have included the output below.

## Issue
`virtualenvs.path` will always takes precedence once a project's virtualenv has been initialised under `{cache-dir}/virtualenvs`.
This means you can only convert to using `virtualenvs.in-project` if the original virtualenv directory is deleted - just setting `virtualenvs.in-project` to true is not enough.
This is likely intended, however it can be a bit misleading if you're unfamiliar with Poetry. It would be nice if something like the following output was displayed when adding packages -
`Installing package to /path/to/exmaple-project/virtualenv`
EDIT: I just realised the above is provided when using -v. Might still be nice to provide it by default though?

Steps -
1. Create a new Poetry project with `virtualenvs.in-project` unset
2. Install a package to populate the cache
3. Remove the package
4. Set `virtualenvs.in-project` to `true`
5. Install a package

lovelyyfiaaa:
Hiiii~~~

I have reproduced this on Fedora 36 (Toolbox)!!

![Screenshot from 2022-10-02 09-15-28](https://user-images.githubusercontent.com/66000635/193435085-df71ba9d-57fa-4031-bfa3-a8b21fc46819.png)
![Screenshot from 2022-10-02 09-19-27](https://user-images.githubusercontent.com/66000635/193435087-6c9b6970-c8ea-4935-a15d-ebe3f6eab32e.png)
![Screenshot from 2022-10-02 09-19-41](https://user-images.githubusercontent.com/66000635/193435088-b3e84204-ee14-427e-ab83-cd8bf7b71c4a.png)
[...]

PR #8126, opened 2023-06-23
Clarify use of virtualenvs.in-project setting.

Davi-Gray:
# Pull Request Check List

Resolves: #6666

- [ ] Added **tests** for changed code. (N/A)

Added documentation for case discussed in #6666. Hopefully this helps clear things up.

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -319,6 +319,9 @@ If set to `true`, the virtualenv will be created and expected in a folder named
 If a virtual environment has already been created for the project under `{cache-dir}/virtualenvs`, setting this variable to `true` will not cause `poetry` to create or use a local virtual environment.
 
 In order for this setting to take effect for a project already in that state, you must delete the virtual environment folder located in `{cache-dir}/virtualenvs`.
+
+You can find out where the current project's virtual environment (if there is one) is stored
+with the command `poetry env info --path`.
 {{% /note %}}
 
 If set to `false`, `poetry` will ignore any existing `.venv` directory.
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -309,11 +309,17 @@ might contain additional Python packages as well.
 
 Create the virtualenv inside the project's root directory.
 
-If not set explicitly, `poetry` by default will create virtual environment under
-`{cache-dir}/virtualenvs` or use the `{project-dir}/.venv` directory when one is available.
+If not set explicitly, `poetry` by default will create a virtual environment under
+`{cache-dir}/virtualenvs` or use the `{project-dir}/.venv` directory if one already exists.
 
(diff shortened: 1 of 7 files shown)


docs/configuration.md
```diff
@@ -309,12 +309,18 @@ might contain additional Python packages as well.
...
+If not set explicitly, `poetry` by default will create a virtual environment under
+`{cache-dir}/virtualenvs` or use the `{project-dir}/.venv` directory if one already exists.
 
 If set to `true`, the virtualenv will be created and expected in a folder named
 `.venv` within the root directory of the project.
 
+{{% note %}}
+If a virtual environment has already been created for the project under `{cache-dir}/virtualenvs`, setting this variable to `true` will not cause `poetry` to create or use a local virtual environment.
+
+In order for this setting to take effect for a project already in that state, you must delete the virtual environment folder located in `{cache-dir}/virtualenvs`.
```
radoering, line 321: Should we add that you can get the exact path via `poetry env info --path` (if there is such an environment)?

Davi-Gray: That seems reasonable. Should I do a new pull request? Or is there a way to add that change into this one? 

On a related note, a couple things I came across while testing this.
1. You can't use 'poetry env remove' to get rid of the non-local virtualenv folder, at least in my (admittedly extremely limited) testing. I didn't  recommend this in the note, so no action needed here. Just mentioning it.

2. While it is possible to override the configuration of virtualenvs.in-project by setting the environment variable POETRY_VIRTUALENVS_IN_PROJECT, it's *not* possible to get the default behavior by doing so - because default behavior (of using local .venv if present, creating/using cache-dir virtualenv otherwise) only occurs when the setting is unset. It's also not possible to override global configuration to use default behavior by using a local configuration (because "poetry config virtualenvs.in-project --unset --local" just leaves the global setting in effect).
I think this is a bug, but maybe not worth addressing since nobody has run into problems because of it (?).

radoering: > That seems reasonable. Should I do a new pull request? Or is there a way to add that change into this one?

You can just add another commit to the git branch of your PR.

Re 1.:
This might be a bug. It may depend on the setting of `virtualenv.in-project` and the existence of a local .venv. If there is no issue yet and you can provide a sequence of commands to reproduce it, you can create an issue.

Re 2.:
I think it's debatable whether this is a bug or just a missing feature. IMO, if you want to override a global specification or set a specific behavior, you can decide if you want to have in-project venvs or not, hence, set it to true or false. The default behavior is only a fallback if no decision has been made.

Davi-Gray: With further testing I wasn't able to reproduce #1. And you're right about #2. I don't think there's any point in trying to make an obscure case work when the same task is easily achievable through existing routes.

radoering (approved):

Issue #8160, opened 2023-07-06
Suddenly require Rust and Cargo installed?


oc-ben-ellis:
- **Poetry version**: 1.5.1
- **Python version**: 3.11.4
- **OS version and name**: WSL2.0 on Windows 11 running Ubuntu
- **pyproject.toml**: Not relevant.

- [*] I am on the [latest](https://github.com/python-poetry/poetry/releases/latest) stable Poetry version, installed using a recommended method.
- [*] I have searched the [issues](https://github.com/python-poetry/poetry/issues) of this repo and believe that this is not a duplicate.
- [*] I have consulted the [FAQ](https://python-poetry.org/docs/faq/) and [blog](https://python-poetry.org/blog/) for any relevant entries or release notes.
- [*] If an exception occurs when executing a command, I executed it again in debug mode (`-vvv` option) and have included the output below.

## Issue
Recently we've been getting the below error in our CI environments. It appears there is now a requirement for rust and cargo to be installed in order to run poetry on alpine linux? 

To reproduce,

`docker run -it --rm alpine:latest /bin/sh -c "apk add python3 py3-pip && pip install poetry"`

We get this error,

```
Collecting rpds-py>=0.7.1 (from jsonschema<5.0.0,>=4.10.0->poetry)
  Downloading rpds_py-0.8.4.tar.gz (15 kB)
  Installing build dependencies ... done
  Getting requirements to build wheel ... done
  Preparing metadata (pyproject.toml) ... error
  error: subprocess-exited-with-error

  × Preparing metadata (pyproject.toml) did not run successfully.
  │ exit code: 1
  ╰─> [6 lines of output]

      Cargo, the Rust package manager, is not installed or is not on PATH.
      This package requires Rust and Cargo to compile extensions. Install it through
      the system's package manager or via https://rustup.rs/

      Checking for Rust toolchain....
      [end of output]

  note: This error originates from a subprocess, and is likely not a problem with pip.
error: metadata-generation-failed

× Encountered error while generating package metadata.
╰─> See above for output.

note: This is an issue with the package mentioned above, not pip.
hint: See above for details.
```

Workaround is to install rust and cargo,

`docker run -it --rm alpine:latest /bin/sh -c "apk add rust cargo python3 py3-pip && pip install poetry"`

Installs succesfully.

oc-ben-ellis:
[...]

PR #8161, opened 2023-07-06
Restrict jsonschema

Secrus:
# Pull Request Check List

Resolves: #8160 

This restricts `jsonschema` to versions below the problematic ones.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -41,7 +41,8 @@ crashtest = "^0.4.1"
 dulwich = "^0.21.2"
 importlib-metadata = { version = ">=4.4", python = "<3.10" }
 installer = "^0.7.0"
-jsonschema = "^4.10.0"
+# jsonschema 4.18 uses Rust-based libraries which causes issues when building from source
+jsonschema = ">=4.10.0,<4.18.0"
 keyring = "^23.9.0"
 # packaging uses calver, so version is unclamped
 packaging = ">=20.4"
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -23,7 +23,8 @@ test_task:
     - POETRY_HOME=/opt/poetry
     - $PYTHON -m venv $POETRY_HOME
     - $POETRY_HOME/bin/pip install --upgrade pip setuptools wheel
-    - $POETRY_HOME/bin/pip install poetry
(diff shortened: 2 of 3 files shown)


Secrus:
@radoering or @neersighted could you please take a look, backport to `1.5` branch, and do a release of this as a hotfix?

oc-ben-ellis:
Are hotfixes required for older versions? Just curious but doesn't affect me directly as I'm on 1.5.1.

Secrus:
> Are hotfixes required for older versions? Just curious but doesn't affect me directly as I'm on 1.5.1.

The tricky part is that it's fine for already installed versions, but if you were to reinstall Poetry, it would pull `jsonschema 4.18` which requires Rust toolchain. For most systems it will be fine, but as presented in the issue, MUSL based OS will have problems.

oc-ben-ellis:
Looks like this version has a dependency on rust also?

`error: can't find Rust compiler`

oc-ben-ellis:
Separate from this PR but maybe worth adding alpine linux to the list of distros to test?

pyproject.toml
```diff
@@ -41,7 +41,7 @@ crashtest = "^0.4.1"
 dulwich = "^0.21.2"
 importlib-metadata = { version = ">=4.4", python = "<3.10" }
 installer = "^0.7.0"
-jsonschema = "^4.10.0"
+jsonschema = ">=4.10.0,<4.18.0"
```
radoering, line 44: Can you add a comment, why we are restricting it?

radoering (commented):
For the BSD tests you probably can just add `jsonschema<4.18.0` to https://github.com/python-poetry/poetry/blob/d31dfa8312e72f6f16fefb5203a5e3bcac7daece/.cirrus.yml#L26

pyproject.toml
Secrus: Done

radoering (approved):

johnthagen:
Has anyone asked for MUSL wheels?

Last I tried to install Poetry on Alpine (year or so ago?) it also failed due to cryptography missing the Rust tool chain, but these days it looks like they publish MUSL wheels?

- https://pypi.org/project/cryptography/41.0.1/#files

Perhaps `jsonschema` would too?

PR #8171, opened 2023-07-08
docs: Added PyPI source pyproject.toml example to repositories.md.

ghammock:
# Pull Request Check List

- [x] Unit tests (`poetry run pytest`)
- [x] Type checking (`poetry run mypy`)
- [x] Pre-commit hooks (`poetry run pre-commit run --all-files`)

## Description

I was trying to verify the best way to use a private repository as a primary, but keep PyPI as a supplemental source in case the private repository didn't have a recent pull from PyPI upstream.  The documentation has the CLI configuration (`poetry source add --priority=primary PyPI`), but not an example for use with `pyproject.toml`.  I added the TOML table for using PyPI as a supplemental source with the note that the URL parameter is invalid in that circumstance.  (`poetry check` replies with `The PyPI repository cannot be configured with a custom url.` if you include the `url` parameter).

## Supplemental Notes

Poetry is the best package manager for Python.  It's so much better than any of the others.  Thanks!


docs/repositories.md
```diff
@@ -174,6 +174,18 @@ poetry source add --priority=primary PyPI
 
 This way, the priority of PyPI can be set in a fine-granular way.
 
+The equivalent specification in `pyproject.toml` is:
+
+```toml
+[[tool.poetry.source]]
+name = "pypi"
+priority = "supplemental"
```
radoering, line 182:
```suggestion
priority = "primary"
```

This should be "primary" to be equivalent to the previous command, shouldn't it?

ghammock: True.  I'll change it.

docs/repositories.md
```diff
@@ -174,6 +174,18 @@ poetry source add --priority=primary PyPI
...
+
+```toml
+[[tool.poetry.source]]
+name = "pypi"
+priority = "primary"
+```
+
+**Omit the `url` when specifying PyPI as a supplemental source.** Because PyPI is internally configured
+with Poetry, the PyPI repository cannot be configured with a given URL.  Remember, you can always use
+`poetry check` to ensure the validity of the `pyproject.toml` file.
```
radoering, line 187:
```suggestion
**Omit the `url` when specifying PyPI explicitly.** Because PyPI is internally configured
with Poetry, the PyPI repository cannot be configured with a given URL. Remember, you can always use
`poetry check` to ensure the validity of the `pyproject.toml` file.
```

PR #8176, opened 2023-07-10
Added downgrading note when updating to an older version

trag1c:
# Pull Request Check List

Resolves: #1394 


johnthagen:
I was just thinking the other day that making downgrades more obvious would be a really useful feature. Thank you for working on this.

johnthagen:
One idea I had for this is that the new version could also be shown in yellow when downgrading so that it pops nicely for the user.

So in this image below, `8.1.3` would be colored yellow:

Screenshot 2023-07-17 at 11 03 38 AM

PR #8218, opened 2023-07-21
Remove package pyproject line if package is module to resolve #7609

### LINKED ISSUE #7609: default pyproject.toml from poetry init -n improperly handles src layout; reports installation success even though it cannot actually be imported
- **Poetry version**: 1.4.0
- **Python version**:
```
Poetry
Version: 1.4.0
Python:  3.10.9

Virtualenv
Python:         3.9.11
```
- **OS version and name**: 5.13.19-2-MANJARO
- **pyproject.toml** (see below)

## Issue

Trying to make simple hello-word package in src layout. The default pyproject.toml created with `poetry init -n` does not let me install it properly.

To begin with, my file structure is like this:

```
$ tree
.
├── README.md
└── src
    └── fancy_project
        ├── __init__.py
        └── something.py

3 directories, 3 files
```

The `__init__.py` file is blank. The `something.py` file contains this:

```python
def hello():
    print("this is a fancy project!")
```

When I run `poetry init -n`, it generates the following (sanitized) `pyproject.toml`.

```toml
[tool.poetry]
name = "fancy-project"
version = "0.1.0"
description = ""
authors = ["aaaa <aaaa@bbbb.com>"]
readme = "README.md"
packages = [{include = "fancy_project"}]

[tool.poetry.dependencies]
python = "^3.9"

[build-system]
requires = ["poetry-core"]
```

When I run `poetry install`, I get this error:

```
$ poetry install
Updating dependencies

Writing lock file

/tmp/fancy-project/fancy_project does not contain any element
```

When I amend `pyproject.toml` to change the `include = "fancy_project"` to `include = "src"`, the `poetry install` command seems to succeed:

```
$ poetry install

Installing the current project: fancy-project (0.1.0)
```

However, the package is not actually importable:

```bash
$ poetry run python -c 'from fancy_project.something import hello; hello()'
  File "<string>", line 1, in <module>
ModuleNotFoundError: No module named 'fancy_project'
```

I tried changing the include line to say `include = "src/fancy_project"`. It similarly reports a successful install, but the import fails just the same.

I tried removing the `packages` line from `pyproject.toml` entirely. This time it works:

```
$ poetry install

Installing the current project: fancy-project (0.1.0)
$ poetry run python -c 'from fancy_project.something import hello; hello()'
this is a fancy project!
```

But I don't understand why. I don't remember ever having to muck around with the `packages` line; everything "just worked" the last time I tried this with a src layout.

I believe this is at least one bug here:

- the default `poetry init -n` should make a pyproject.toml that can work with src layout
- `poetry install` should not say that an installation succeeded when it actually didn't
- the docs for `packages` don't clear up any of this confusion
- the "does not contain any element" error message does not make any sense

(PS: I ran `rm -rf .venv` before every `poetry install` command, to start fro ma clean slate each time)

wdonofrio:
# Pull Request Check List

Resolves: #7609

While it does resolve the issue, I'm a little worried it completely removed the package pyproject line in `poetry init`. I'm not familiar with how this line is meant to be used on `init`, so please provide feedback or make edits. 

Tests were modified to support this change, which removes the package line from pyproject on init.
I don't think a documentation change should be made - but I do think the error message should be clarified for the user, since I believe it still occurs if a package can't be found. This is more involved though as the `raise` is coming from poetry-core.

### DIFF
--- a/tests/console/commands/test_init.py
+++ b/tests/console/commands/test_init.py
@@ -86,7 +86,6 @@ description = "This is a description"
 authors = ["Your Name <you@example.com>"]
 license = "MIT"
 readme = "README.md"
-packages = [{include = "my_package"}]
 
 [tool.poetry.dependencies]
 python = "~2.7 || ^3.6"
@@ -169,7 +168,6 @@ description = "This is a description"
 authors = ["Your Name <you@example.com>"]
 license = "MIT"
 readme = "README.md"
-packages = [{include = "my_package"}]
 
 [tool.poetry.dependencies]
 python = "~2.7 || ^3.6"
@@ -218,7 +216,6 @@ description = "This is a description"
 authors = ["Your Name <you@example.com>"]
 license = "MIT"
 readme = "README.md"
-packages = [{include = "my_package"}]
 
 [tool.poetry.dependencies]
 python = "~2.7 || ^3.6"
@@ -249,7 +246,6 @@ version = "1.2.3"
 description = ""
 authors = ["Your Name <you@example.com>"]
 readme = "README.md"
-packages = [{{include = "my_package"}}]
 
 [tool.poetry.dependencies]
 python = "^{python}"
@@ -290,7 +286,6 @@ description = "This is a description"
 authors = ["Your Name <you@example.com>"]
 license = "MIT"
 readme = "README.md"
-packages = [{include = "my_package"}]
 
 [tool.poetry.dependencies]
 python = "~2.7 || ^3.6"
@@ -383,7 +378,6 @@ description = "This is a description"
 authors = ["Your Name <you@example.com>"]
 license = "MIT"
 readme = "README.md"
-packages = [{include = "my_package"}]
 
 [tool.poetry.dependencies]
 python = "~2.7 || ^3.6"
@@ -429,7 +423,6 @@ description = "This is a description"
 authors = ["Your Name <you@example.com>"]
 license = "MIT"
 readme = "README.md"
-packages = [{include = "my_package"}]
 
 [tool.poetry.dependencies]
 python = "~2.7 || ^3.6"
@@ -481,7 +474,6 @@ description = "This is a description"
 authors = ["Your Name <you@example.com>"]
 license = "MIT"
 readme = "README.md"
-packages = [{include = "my_package"}]
 
 [tool.poetry.dependencies]
 python = "~2.7 || ^3.6"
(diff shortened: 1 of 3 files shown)


tests/console/commands/test_init.py
```diff
@@ -1043,7 +1025,17 @@ def test_package_include(
         ),
     )
 
-    packages = "" if include is None else f'packages = [{{include = "{include}"}}]\n'
+    def module_name(name: str) -> str:
+        return name.replace("_", "-")
```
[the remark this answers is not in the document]
wdonofrio: You're right. Thanks for the callout!

I realized my mistake and refactored to leverage poetry-core directly.

tests/console/commands/test_new.py
```diff
@@ -57,9 +57,7 @@ def verify_project_directory(
 
     packages = poetry.local_config.get("packages")
 
-    if not packages:
-        assert poetry.local_config.get("name") == package_include.get("include")
```
radoering, line 61: We probably should not remove this test but adapt it. `module_name()`?

radoering (changes requested):

tests/console/commands/test_new.py
wdonofrio: Agreed. Thanks for the feedback! I have adapted the test.

tests/console/commands/test_new.py
```diff
@@ -55,10 +57,11 @@ def verify_project_directory(
     else:
         package_include = {"include": package_path.parts[0]}
 
+    name = poetry.local_config.get("name")
     packages = poetry.local_config.get("packages")
 
     if not packages:
-        assert poetry.local_config.get("name") == package_include.get("include")
+        assert module_name(name) == package_include.get("include") if name else ""
```
radoering, line 64:
```suggestion
        assert module_name(name) == package_include.get("include")
```

`name` is always set isn't it?

If I don't miss anything the expression wouldn't work if it wasn't set because there is a missing bracket. It had to be:

```
assert module_name(name) == (package_include.get("include") if name else "")
```

wdonofrio: `name` is always set because `module_name()` expects a `str`. Therefore your suggestion does not pass mypy.

However, I think this may be a better approach. What do you think?
```python
    name = poetry.local_config.get("name", "")
    packages = poetry.local_config.get("packages")

    if not packages:
        assert module_name(name) == package_include.get("include")
```

radoering: LGTM 👍

PR #8227, opened 2023-07-24
fix: remove exception when keyring is locked

real-yfprojects:
Resolves: https://github.com/python-poetry/poetry/issues/1917
This inherits and therefore closes #6612.

* src/poetry/utils/password_manager.py (PoetryKeyring.get_credential): Catch `KeyringError` and `RuntimeError` when accessing keyring.

### DIFF
--- a/src/poetry/utils/password_manager.py
+++ b/src/poetry/utils/password_manager.py
@@ -48,7 +48,11 @@ class PoetryKeyring:
         import keyring
 
         for name in names:
-            credential = keyring.get_credential(name, username)
+            try:
+                credential = keyring.get_credential(name, username)
+            except keyring.errors.KeyringLocked:
+                logger.debug("Keyring %s is locked", name)
+                credential = None
             if credential:
                 return HTTPAuthCredential(
                     username=credential.username, password=credential.password
--- a/src/poetry/utils/password_manager.py
+++ b/src/poetry/utils/password_manager.py
@@ -47,12 +47,18 @@ class PoetryKeyring:
 
         import keyring
 
+        from keyring.errors import KeyringError
+        from keyring.errors import KeyringLocked
+
         for name in names:
+            credential = None
             try:
                 credential = keyring.get_credential(name, username)
-            except keyring.errors.KeyringLocked:
+            except KeyringLocked:
                 logger.debug("Keyring %s is locked", name)
-                credential = None
+            except (KeyringError, RuntimeError):
+                logger.debug("Accessing keyring %s failed", name, exc_info=True)
+
             if credential:
                 return HTTPAuthCredential(
                     username=credential.username, password=credential.password
--- a/tests/utils/test_authenticator.py
+++ b/tests/utils/test_authenticator.py
@@ -94,6 +94,19 @@ def test_authenticator_uses_username_only_credentials(
     assert request.headers["Authorization"] == "Basic Zm9vMDAxOg=="
 
 
+def test_authenticator_ignores_locked_keyring(
+    mock_config: Config,
+    mock_remote: None,
+    http: type[httpretty.httpretty],
+    with_locked_keyring: None,
+) -> None:
+    authenticator = Authenticator(mock_config, NullIO())
+    authenticator.request("get", "https://foo001@foo.bar/files/foo-0.1.0.tar.gz")
+    request = http.last_request()
+
+    assert request.headers["Authorization"] == "Basic Zm9vMDAxOg=="
+
+
 def test_authenticator_uses_password_only_credentials(
     mock_config: Config, mock_remote: None, http: type[httpretty.httpretty]
 ) -> None:
--- a/tests/utils/test_authenticator.py
+++ b/tests/utils/test_authenticator.py
@@ -1,6 +1,7 @@
 from __future__ import annotations
 
 import base64
+import logging
 import re
 import uuid
 
@@ -20,6 +21,7 @@ from poetry.utils.authenticator import RepositoryCertificateConfig
 
 
 if TYPE_CHECKING:
+    from _pytest.logging import LogCaptureFixture
     from _pytest.monkeypatch import MonkeyPatch
     from pytest_mock import MockerFixture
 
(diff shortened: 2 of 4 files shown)


src/poetry/utils/password_manager.py
```diff
@@ -47,8 +47,18 @@ def get_credential(
...
 
+        from keyring.errors import KeyringError
+        from keyring.errors import KeyringLocked
+
         for name in names:
-            credential = keyring.get_credential(name, username)
+            credential = None
+            try:
+                credential = keyring.get_credential(name, username)
+            except KeyringLocked:
```
dimbleby, line 57: is `KeyringLocked` really worth a special case?  every other access of keyring in this file goes
```python
except (RuntimeError, keyring.errors.KeyringError):
```
and I feel it would be hard to argue with consistency,

(Personally I am am not a fan of catching `RuntimeError` and would prefer to have people raise bugs against keyring if it raises random exceptions; but that isn't the view taken in the rest of the code)

dimbleby (commented):
review of #6612 indicated that unit test would be required, which seems reasonable.

I'd suggest adding a `LockedBackend` to `tests/conftest.py` - cf the existing `DummyBackend` etc - and writing a test that uses that in `test_authenticator.py`.

src/poetry/utils/password_manager.py
real-yfprojects: KeyLocked is an expected exception resulting from specific user interaction. That's why we wouldn't want to print a traceback. Traceback are for unexpected errors that might need further action by the user. That's the case for all other exceptions thrown by keyring.

dimbleby: sure, I wouldn't print a traceback for any of them

real-yfprojects: Well, for RuntimeError a traceback could be relevant for debugging. I think the same applies to KeyringError.

dimbleby: well, I don't care enough to argue about it, happy to leave that to someone with the commit bit.  (but you'll have fewer testcases to write if you handle all the exceptions in the same way!)

radoering: IMO, it makes sense to print the traceback (or at least the original error) for unexpected error messages. So I think that's ok with me.

As already mentioned, we should add tests.

real-yfprojects:
Ok so I added a few tests but I am very unsure whether I did it correctly and sufficiently. Please review.

dimbleby:
tests look plausible to me.  I'm not a fan of the many imports of `KeyringLocked` in `conftest.py` - just import it once and be done! - but that's not a big deal.

real-yfprojects:
> I'm not a fan of the many imports of `KeyringLocked` in `conftest.py` - just import it once and be done!

Me neither, but I thought there must a reason the existing code only imports keyring on demand instead of adding the imports to the top of the file.

dimbleby:
There''s already a keyring import at the top of the file, I see no reason to be afraid of another

https://github.com/python-poetry/poetry/blob/2b16b2fcf15da2a69e118a637ab4ee4fcb6a234c/tests/conftest.py#L17

real-yfprojects:
Ok, fixed

radoering:
> Ok so I added a few tests but I am very unsure whether I did it correctly and sufficiently. Please review.

You only tested one of your `except` branches. If you add another test with `KeyringError` instead of `KeyRingLocked` in `test_password_manager.py`, it will be good enough IMO.

tests/utils/test_authenticator.py
```diff
@@ -94,6 +94,32 @@ def test_authenticator_uses_username_only_credentials(
...
+    mock_config: Config,
+    mock_remote: None,
+    http: type[httpretty.httpretty],
+    with_erroneous_keyring: None,
+) -> None:
+    authenticator = Authenticator(mock_config, NullIO())
+    authenticator.request("get", "https://foo001@foo.bar/files/foo-0.1.0.tar.gz")
+    request = http.last_request()
+
+    assert request.headers["Authorization"] == "Basic Zm9vMDAxOg=="
```
real-yfprojects, line 120: Tbh I don't know what this assert is actually testing, I just copied it from the other tests. However the `Authorization` is the same for the other tests using the same URL. Additionally I don't understand why some tests use different URLs. While the tests seem to be working I cannot be sure they are correct.

radoering: I suppose the assert is just some sort of smoke test.

However, it seems the changed code is never called here (same for `test_authenticator_ignores_locked_keyring`) because there is a username in the URL. It would be best to check that the appropriate messages are logged so that we can be sure the relevant code is really called.

radoering (changes requested):

radoering (approved):

arneyjfs:
### Had to manually set backend keyring for private and public repos

Fantastic to see keyring improvements being released in this version. However I still had a convoluted route to getting authentication working when trying to install packages from google artifact registry over SSH (5 hours in total). I know 1.7.0 is not released yet so there are still some bugs expected but since this issue is closed I thought I would log here all the loops I had to jump through in case it is helpful in finding any more bugs that might have been missed. The crux of the issue was I had to set the backend manually depending on if I was installing from public or private repos.

**Install process**
1. Start out with poetry 1.6.1 through normal install methods 
```shell
curl -sSL https://install.python-poetry.org | python3 -
```  
> Cannot directly install 1.7.0 from git as I end up with a `module not found: cleo` error (separate issue)

2. upgrade to 1.7.0.dev0
```shell
curl -sSL https://install.python-poetry.org | python3 - --git https://github.com/python-poetry/poetry.git@master
```  

3. Install `keyrings.google-artifactregistry-auth` latest version (1.1.2 in my case)

```shell
poetry self add keyrings.google-artifactregistry-auth
```  

4. This was the particularly tricky part to work out... I eventually found that I have to explicitly set the keyring backends to get a full install of both packages from my private Artefact Registry repo, and public the PyPi public repo. If I set `export PYTHON_KEYRING_BACKEND=keyring.backends.fail.Keyring` then obviously I cannot authenticate to google since I need the keyrings.gauth.GooglePythonAuth for that. However if I set `export PYTHON_KEYRING_BACKEND=keyrings.gauth.GooglePythonAuth` then all other packages not from my python repo hang forever (presumably waiting to be authed or something?). Therefore the below worked for me:

    4.1 Remove private packages from pyproject.toml, then explicitly remove keyring backend and install public packages:
   
   ```shell
    export PYTHON_KEYRING_BACKEND=keyring.backends.fail.Keyring
    poetry install
   ```

    4.2 explicitly set keyring backend to google auth, and install the remaining private packages:
   ```shell
    export PYTHON_KEYRING_BACKEND=keyrings.gauth.GooglePythonAuth
    poetry add  --source <repo-id> <package-name>
   ```

This then completed successfully for me. I hope it is helpful. Please let me know if any more info is required or if I am making a silly mistake. Thanks!

**System:**
Ubuntu 22.04.3 LTS **Over SSH**
keyrings-google-artifactregistry-auth 1.1.2
Poetry 1.7.0.dev0

Issue #7881, opened 2023-05-05
Intermittent Hash Validation Failures


rohanp-eiq:
- **Poetry version**: 1.4.2
- **Python version**: 3.10.6
- **OS version and name**: Ubuntu 22.04.2 LTS (running in a docker image based on this)
- **pyproject.toml**: https://gist.github.com/rohanp-eiq/3171fb1561994642b2580313b1939a29

- [x] I am on the [latest](https://github.com/python-poetry/poetry/releases/latest) stable Poetry version, installed using a recommended method.
- [x] I have searched the [issues](https://github.com/python-poetry/poetry/issues) of this repo and believe that this is not a duplicate.
- [x] I have consulted the [FAQ](https://python-poetry.org/docs/faq/) and [blog](https://python-poetry.org/blog/) for any relevant entries or release notes.
- [x] If an exception occurs when executing a command, I executed it again in debug mode (`-vvv` option) and have included the output below.

## Issue

Hi, we are running into an issue where we intermittently see libraries failing to install with the command: `poetry install -vvv --with dev,test --sync` with the following error (torch is an example, this has failed on other libraries):

```Stack trace:

  7  ~/.local/pipx/venvs/poetry/lib/python3.10/site-packages/poetry/installation/executor.py:280 in _execute_operation
      278│ 
      279│             try:
    → 280│                 result = self._do_execute_operation(operation)
      281│             except EnvCommandError as e:
      282│                 if e.e.returncode == -2:

  6  ~/.local/pipx/venvs/poetry/lib/python3.10/site-packages/poetry/installation/executor.py:382 in _do_execute_operation
      380│             return 0
      381│ 
    → 382│         result: int = getattr(self, f"_execute_{method}")(operation)
      383│ 
      384│         if result != 0:

  5  ~/.local/pipx/venvs/poetry/lib/python3.10/site-packages/poetry/installation/executor.py:502 in _execute_install
      500│ 
      501│     def _execute_install(self, operation: Install | Update) -> int:
    → 502│         status_code = self._install(operation)
      503│ 
      504│         self._save_url_reference(operation)

  4  ~/.local/pipx/venvs/poetry/lib/python3.10/site-packages/poetry/installation/executor.py:540 in _install
      538│             archive = self._download_link(operation, Link(package.source_url))
      539│         else:
    → 540│             archive = self._download(operation)
      541│ 
      542│         operation_message = self.get_operation_message(operation)

[...]

Issue #8232, opened 2023-07-24
1.6.0 Release


radoering:
This issue tracks the progress of version 1.6.0 of Poetry. Please surface any blockers or concerns you have regarding this minor release here:

## Blockers
* #8255
  or #8256
*  #8304
* python-poetry/poetry-core#626
  * python-poetry/poetry-core#609
  * python-poetry/poetry-core#489
  * update vendored dependencies, cf #8223
* Changelog #8333
* Release notes (blog) python-poetry/website#132

## Nice to have
* python-poetry/poetry-plugin-export#227

## Important changes to surface

(This is intended to be the basis for the blog post and should cover new features or any manual steps/incompatibility when upgrading)

* official Poetry badge https://github.com/python-poetry/poetry/issues/8232#issuecomment-1653531661
* Support non SHA256/SHA384/SHA512 HTTPRepository ?
* Poetry 1.6 will drop support for Python 3.7 as runtime.

johnthagen:
It would be fun in the blog post for the 1.6.0 release to highlight the new official Poetry badge:

- https://github.com/python-poetry/website/pull/119

[![Poetry](https://img.shields.io/endpoint?url=https://python-poetry.org/badge/v0.json)](https://python-poetry.org/)

```
[![Poetry](https://img.shields.io/endpoint?url=https://python-poetry.org/badge/v0.json)](https://python-poetry.org/)
```

PR #8235, opened 2023-07-24
fix: cache corruption due to broken connections, where of the wheel is only partially downloaded

realabja:
This issue occurs mostly when having unstable connections and package is only half downloaded and it causes cache corruption. Raising a runtime error will fix the issue as it does not cause hash issues while retrying the install

# Pull Request Check List

Resolves: #7881

### DIFF
--- a/src/poetry/installation/executor.py
+++ b/src/poetry/installation/executor.py
@@ -870,6 +870,15 @@ class Executor:
             with self._lock:
                 progress.finish()
 
+        # Validate expected byte size based on http headers
+        if wheel_size is not None and done != int(wheel_size):
+            raise RuntimeError(
+                "Downloaded wheel does not match expected size, "
+                f" was {done} expected {wheel_size}. "
+                "This means probably connection was interrupted "
+                "and might corrupt the cache."
+            )
+
         return archive
 
     def _should_write_operation(self, operation: Operation) -> bool:


dimbleby:
https://github.com/urllib3/urllib3/pull/2514 defaults `enforce_content_length` to `True` for urllib3 2.0, and the main branch of poetry allows urllib3 2.0.

So this should be fixed automatically in the next release without needing to write any code?

realabja:
@dimbleby but does poetry include urllib3 >= 2.0 ? Would it be acceptable to have a hard check by poetry as well as through urllib3, it wouldn't cause any conflict I suppose

dimbleby:
the main branch of poetry doesn't require any particular version of urllib3 because it is only an indirect dependency; but urrlib3 2.04 is what you get if you install poetry from that main branch today

realabja:
@dimbleby Just tried the main branch and it solves the issue perfectly

realabja:
I will close this as it resolves itself with next release

realabja:
@dimbleby  when do you think  next release happens?

dimbleby:
no idea, I have no more influence on that than you do.  Watch https://github.com/python-poetry/poetry/issues/8232 I guess

realabja:
@dimbleby  yep just saw there is an issue for it

Issue #7685, opened 2023-03-19
poetry add with --index-url option


alanwilter:
- [x] I have searched the [issues](https://github.com/python-poetry/poetry/issues) of this repo and believe that this is not a duplicate.
- [x] I have searched the [FAQ](https://python-poetry.org/docs/faq/) and general [documentation](https://python-poetry.org/docs/) and believe that my question is not already covered.

## Feature Request

I do this with pip:

```bash
pip3 install torch --index-url https://download.pytorch.org/whl/cpu
```

I can't find an equivalent in `poetry`. Suggestion:

```bash
poetry add torch --index-url https://download.pytorch.org/whl/cpu
```

dimbleby:
https://python-poetry.org/docs/repositories/, please close

alanwilter:
Thanks!

alanwilter:
Ok, I got it, I just have to wait it searches...

I was having difficult to understand this bit:

> Any package source not marked as secondary will take precedence over PyPI.

I still don't understand that statement, but I glad I got it to work as I wanted.

```toml
[tool.poetry.dependencies]
torch = {version = "^2.0.0+cpu", source = "pytorch-cpu"}
...

[tool.poetry.group.dev.dependencies]
torch = {version = "^2.0.0", source = "pypi"}
...
```

hyliu1989:
Thought I could share some successful experience on the same issue.

After updating poetry to 1.5.1, there is a new package URL specification priority call explicit and that works more clear to me.
```
poetry source add --priority=explicit pytorch-gpu-src https://download.pytorch.org/whl/cu118
poetry add --source pytorch-gpu-src torch torchvision torchaudio
```

I can successfully generate a pyproject.toml with the following section
```
[tool.poetry.dependencies]
python = ">=3.9, <3.11"
torch = {version = "^2.0.1+cu118", source = "pytorch-gpu-src"}
torchvision = {version = "^0.15.2+cu118", source = "pytorch-gpu-src"}
torchaudio = {version = "^2.0.2+cu118", source = "pytorch-gpu-src"}
```

Hope this helps for people revisiting here.

Martins6:
I'm going to say what I think everyone that came to this issue is thinking. The answer from @hyliu1989 should be on the documentation of `poetry add` (https://python-poetry.org/docs/cli/#add) to give more visibility to users that use PyTorch and other specific repositories.

dimbleby:
[...]

PR #8246, opened 2023-07-27
Add a documentation for adding other online source for a dependency.

hyliu1989:
# Pull Request Check List

Resolves: [#issue-number-here](https://github.com/python-poetry/poetry/issues/7685#issuecomment-1654355849)

This PR updates the documentation for command `add --source` for adding a dependency package with other online source (i.e., other than PyPi but not a git source). It includes a summary sentence and a two-line example that contains `source add` and `add` commands.


radoering (commented):
I think I'd prefer to just add a short sentence with a link to https://python-poetry.org/docs/repositories/#installing-from-private-package-sources instead of having almost the same in two places.

Issue #8191, opened 2023-07-12
improve performance of version solver for dependencies with many versions


ahvigil:
- [x] I have searched the [issues](https://github.com/python-poetry/poetry/issues) of this repo and believe that this is not a duplicate.
- [x] I have searched the [FAQ](https://python-poetry.org/docs/faq/) and general [documentation](https://python-poetry.org/docs/) and believe that my question is not already covered.

## Feature Request

The current dependency solver can perform suboptimally for packages with many frequent releases creating long chains to explore. `botocore` in particular seems to be problematic and can produce dead ends in the graph of possible package combinations that poetry takes _a long time_ to resolve out of.

See: 

- https://github.com/python-poetry/poetry/issues/832#issuecomment-613618715
- https://github.com/python-poetry/poetry/issues/7858
- https://github.com/python-poetry/poetry/issues/7875
- https://github.com/python-poetry/poetry/issues/7856

The workarounds for these are generally of the "pin package X" variety, which works well but doesn't make for the best user experience and adds requirements to a pyproject.toml that a developer doesn't necessarily care about. The documentation section on ["why-is-the-dependency-resolution-process-slow"](https://python-poetry.org/docs/faq/#why-is-the-dependency-resolution-process-slow) also suggest adding pins as a solution for speed up resolution. 

Investigate improving the performance of the version solver in these cases, preferably without too much impact on non-pathologic cases.

dimbleby:
the code you've linked has nothing much to do with the main part of poetry's search, which is neither depth-first nor breadth-first.

what you want to think about is changing the heuristic by which poetry chooses the next package to resolve - https://github.com/python-poetry/poetry/blob/01657ddeca182259e4350794a4f378241336085f/src/poetry/mixology/version_solver.py#L443-L475

eg most of the cases that currently go wrong are made much faster by reversing the preference for packages with fewer available versions.  However: there's every chance that making that change will introduce _new_ pathological cases - there is no silver bullet.

Feel free to experiment.

ahvigil:
thats a great pointer, thanks! i must admit I'm not too familiar with the poetry codebase except as a user, seems I was looking in the wrong place.

ahvigil:
using the [backtracking unit test cases](https://github.com/python-poetry/poetry/blob/master/tests/mixology/version_solver/test_backtracking.py) as a model I came up with a case that reproduces the performance issue I'm circling around
```
def test_i_can_reproduce_the_issue(
[...]

PR #8255, opened 2023-07-30
solver: tune heuristics for choosing the next dependency to resolve …

radoering:
… in a way that dependencies that are not required by another unsatisfied dependency are resolved first

Alternative: #8256

# Pull Request Check List

Resolves: (partially<sup>1</sup>) boto3/botocore vs. urllib3 issue
Closes: #8256

Currently, we choose to resolve dependencies with less possible versions first. We can improve the heuristics, which dependency to resolve first, by considering relations between dependencies:

If dependency `A` depends on dependency `B`, we should resolve dependency `A` first no matter which dependency has more versions. Of course, different versions of a package can have different dependencies, but most often dependencies are at least similar and don't change much between versions. So I think the new heuristics will not worsen the performance in most cases (and of course improve it in some cases) - at least if the heuristics itself can be calculated fast enough.

Example: If latest `A` requires `B<2`, then older versions of `A` will typically not allow newer versions of `B` and if they do it's only because the incompatibility had not been known at the time they were released.

Some measurements (with warm cache):

|---|---|---|
|shootout example|225 s|8 s|
|shootout example with additional `urllib3<2` constraint|8 s|8 s|
|#4670|1 s|1 s|
|#4870|70 s|118 s|
|large internal project|64 s|52 s|

As can be seen, the performance of the shootout example improves dramatically. The performance of the #4870 example is worse, however, we also get some different (probably better) results:

|#4870 without PR|#4870 with PR|
|---|---|
|Sphinx 3.5.3 (requires docutils>=0.12)<br/>docutils 0.18.1|Sphinx 4.3.2 (requires docutils>=0.14,<0.18)<br/>docutils 0.17.1|

The solution with the PR is probably better because an older Sphinx version probably does not work with a newer docutils version. The constraint is just missing because the incompatibility was not known when the older Sphinx version was released. Since that's a typical issue, **the new heuristics may even lead to less surprising results for inexperienced users if there are several possible solutions.**

<sup>1</sup> This PR can solve the boto3/botocore vs. urllib3 issue in some cases (like the shootout example), but not in all cases. That's because boto3 does not depend directly on urllib3 but on botocore, which depends on urllib3. With this PR we only look forward one level. That means it doesn't help if we have to choose between boto3 and urllib3 with no other dependency that depends on urllib3 in the list of unsatisfied dependencies yet as in the example in #7950 with boto3 and urllib3 as the only top level dependencies.

### DIFF
--- a/src/poetry/mixology/version_solver.py
+++ b/src/poetry/mixology/version_solver.py
@@ -22,6 +22,7 @@ from poetry.mixology.term import Term
 
 
 if TYPE_CHECKING:
+    from poetry.core.constraints.version import VersionConstraint
     from poetry.core.packages.project_package import ProjectPackage
 
     from poetry.packages import DependencyPackage
@@ -139,6 +140,13 @@ class VersionSolver:
             int, set[Incompatibility]
         ] = collections.defaultdict(set)
         self._solution = PartialSolution()
+
+        # package name -> package.requires names for an arbitrary version of the package
+        # This is only used to determine which package to resolve next.
+        self._requiresHeuristics: dict[str, set[str]] = {}
+        self._get_package_for_dependency = functools.lru_cache(maxsize=None)(
+            self._get_package_for_dependency_uncached
+        )
 
     @property
     def solution(self) -> PartialSolution:
@@ -415,6 +423,30 @@ class VersionSolver:
 
         raise SolveFailure(incompatibility)
 
+    def _get_package_for_dependency_uncached(
+        self,
+        dependency: Dependency,
+        # Constraint is unused but required to calculate the cache key,
+        # because two direct origin dependencies with the same origin are equal
+        # independent of the constraint. If neglected, we run into
+        # https://github.com/python-poetry/poetry/issues/7398 again.
+        constraint: VersionConstraint,
+    ) -> DependencyPackage | None:
+        locked = self._provider.get_locked(dependency)
+        if locked is None:
+            packages = self._dependency_cache.search_for(
+                dependency, self._solution.decision_level
+            )
+            package = next(iter(packages), None)
+
+            if package is None:
+                return None
+        else:
+            package = locked
+
+        # complete package so that package.requires are set
+        return self._provider.complete_package(package)
+
     def _choose_package_version(self) -> str | None:
         """
         Tries to select a version of a required package.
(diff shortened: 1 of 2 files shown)


dimbleby:
Have you thought about the approach being explored in #8191, which simply reverses the preference for "packages with fewer versions"?

That is tempting to me in that it is a single character fix - just add a minus sign.  And presumably it will also handle the cases described in your footnote.

I suppose all heuristics are in the end subject to bad cases - I'm pretty sure that the problem ought to be NP-hard.  While I see the logic in trying to force conflicts sooner rather than later: perhaps that heuristic is optimising for a case that turns out to be not very important in the python ecosystem, whereas "packages with most versions" would be helpful in the world as it actually is.

My experience is that: in most solutions most packages are at the latest allowed version anyway.  So simply eliminating the known pathological case might be a pretty sensible try.

radoering:
No, I haven't really thought about it. I just assumed that there is a good reason for choosing dependencies with less versions first - there's even a test (`test_traverse_into_package_with_fewer_versions_first`). On the other hand, the code is quite old so that we won't find out if there were some real-world examples that benefit from that decision.

Choosing packages with more versions first, of course, solves the boto3/urllib3 issue completely and also the Sphinx/docutils example. Testing the projects from the description, I can measure a slight performance regression for the shootout example with the additional urllib constraint: When choosing packages with less versions first it takes 8 s, when choosing packages with more versions first it takes 10 s. Since we have no example yet where it's worse, maybe, we should risk it...

dimbleby:
the original algorithm describes "fewest versions first" - https://github.com/dart-lang/pub/blob/master/doc/solver.md#decision-making

> Pub chooses the latest matching version of the package with the fewest versions that match the outstanding constraint. This tends to find conflicts earlier if any exist, since these packages will run out of versions to try more quickly. But there's likely room for improvement in these heuristics.

I suspect the testcase was written just because that's how the algorithm is described.  While that text acknowledges room for improvement, it would be amusing if that improvement was doing the exact opposite!

The way I'm thinking about it is that - in cases where there are real conflicts to resolve - reversing this heuristic probably loses some average performance, but helps with worst-case performance

- most of the time it makes sense to try and find dead-ends sooner rather than later, per the original heuristic
- but we have this known expensive case where poetry reaches a dead-end (eg with urllib3==2.0) but can't tell that it's a dead end without working its way through a zillion versions of eg botocore, and there it helps a _lot_ to do things the other way round

But also most of the time there just aren't that many conflicts anyway - as I say, mostly most things end up at the latest version - so my hope is that the damage this does to the average case isn't much to worry about.

However I've a feeling that I could tell myself a story justifying almost any heuristic!  Perhaps there's no way really to know except to ship it and see what the new bad cases are...

PR #8256, opened 2023-07-30
solver: invert heuristics for choosing the next dependency to resolve…

radoering:
… so that dependencies with more versions are resolved first

Although the original algorithm proposes to choose dependencies with fewer versions first, we don't know about any real-world examples where this is important. However, we know about real-world examples (especially boto3 vs. urllib3, but also Sphinx vs. docutils) where it's better to choose dependencies with more versions first.

Alternative to #8255

# Pull Request Check List

Resolves: #8191, boto3/botocore vs. urllib3 issue
Closes: #8255
see https://github.com/python-poetry/poetry/pull/8255#issuecomment-1657198242

### DIFF
--- a/tests/mixology/version_solver/test_backtracking.py
+++ b/tests/mixology/version_solver/test_backtracking.py
@@ -67,7 +67,7 @@ def test_backjumps_after_partial_satisfier(
     add_to_repo(repo, "y", "1.0.0")
     add_to_repo(repo, "y", "2.0.0")
 
-    check_solver_result(root, provider, {"c": "1.0.0", "y": "2.0.0"}, tries=2)
+    check_solver_result(root, provider, {"c": "1.0.0", "y": "2.0.0"}, tries=4)
 
 
 def test_rolls_back_leaf_versions_first(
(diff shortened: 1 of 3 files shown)


tests/mixology/version_solver/test_backtracking.py
```diff
@@ -132,30 +132,30 @@ def test_backjump_to_nearer_unsatisfied_package(
     )
 
 
-def test_traverse_into_package_with_fewer_versions_first(
+def test_traverse_into_package_with_more_versions_first(
```
dimbleby, line 135: is this testcase worth keeping?

if what we are learning is that it's one of those testcases that just follows whatever the code does, then it's not really providing useful protection against regressions

perhaps it _would_ be worth adding the testcase given in https://github.com/python-poetry/poetry/issues/8191#issuecomment-1633408719: that actually would provide some defense against accidentally reintroducing this particular perf problem

Secrus (approved):

Tenzer:
Thank you for this! Last week I updated transient dependencies for a project using `poetry update <package>`, and the command took over 3 minutes each time. With Poetry 1.6.0 it instead takes 8 seconds 😄

nambrosini17:
Awesome work! I was about to drop poetry since it took ~10 minutes for every dependency change. Updating it to check how much it improves :rocket:

PR #8280, opened 2023-08-08
Clarify docs

temeddix:
# Pull Request Check List

I've discovered that there is plenty of room for improvement in the documentation.  Considerable portion of Python coders are not professional programmers, so we need more guides about environment variables and PATH. Also, More easier description and crucial information about using Poetry on Windows and Python from MS Store were needed.

I am very much experienced with conda and pip, but the first impression of Poetry was quite frustrating. Adding this information to the docs will reduce the barrier considerably.

### DIFF
--- a/docs/_index.md
+++ b/docs/_index.md
@@ -67,8 +67,13 @@ Note: On some systems, `python` may still refer to Python 2 instead of Python 3.
 ```
 
 {{% note %}}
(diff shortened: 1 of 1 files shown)


temeddix:
If this PR is merged, I will try to provide further improvements with Poetry experience for those using Python from Microsoft Store. This issue has existed for too long since #5331.

temeddix:
Isn't there anybody to review?

Issue #8288, opened 2023-08-09
Mark tests using internet with @pytest.mark.network 


hrnciar:
- [x] I have searched the [issues](https://github.com/python-poetry/poetry/issues) of this repo and believe that this is not a duplicate.
- [x] I have searched the [FAQ](https://python-poetry.org/docs/faq/) and general [documentation](https://python-poetry.org/docs/) and believe that my question is not already covered.

## Feature Request

Hello,

I am maintaining `poetry` in Fedora, recently we noticed the number of tests requiring internet access has increased. Since internet access is not enabled while building the package we need to skip those tests. I am just working on updating `poetry` to `1.5.1` and it looks like this:

~~~~
%pytest -k "not executor and \
   not test_chef and \
   not test_installer_with_pypi_repository and \
   not test_info_setup_complex and not test_info_setup_complex_pep517_legacy and \
   not test_info_setup_missing_mandatory_should_trigger_pep517 and \
   not test_builder_setup_generation_runs_with_pip_editable"
~~~~

It would be nice to mark those tests to be able to disable them all at once with `pytest -m "not network"`. 
I've already prepared PR for this, I'll open it soon.

eamanu:
Hi @hrnciar

I maintain poetry in Debian. Leave this comment here to don't add no related code comment in PR. 

This is a very important feature. Are you still interested on it? I looked that you marked only 5 tests, I think that there are more  tests, but not sure.

hrnciar:
Hi, thank you for pinging me. It seems like I left out some tests in `test_chef.py`. Please try to see if the updated PR works for you in Debian.  I am getting a successful build in Fedora when building without internet access.

PR #8289, opened 2023-08-09
Add network mark for pytest

hrnciar:
# Pull Request Check List

Resolves: #8288 


dimbleby:
I expect that in most or all of these it's undesired that the test relies on network access, and the preferred fix should be to add more things to tests/repositories/fixtures so that this is not the case

radoering:
> I expect that in most or all of these it's undesired that the test relies on network access, and the preferred fix should be to add more things to `tests/repositories/fixtures` so that this is not the case

I totally agree that we should avoid network access if possible. I suppose most (or at least some) of the network accesses are caused by pip subprocesses. Thus, I wonder what should be done in such a case:

- Mock calls to pip with the risk that we don't notice if something changes in the interface of pip or we introduce a bug that passes wrong arguments to pip?
- Start a local package index server and pass `--index-url` to pip calls?
- ...?

By the way, it looks like the tests with network access are all among the slowest tests (at least on my computer) so a change might also speed up the test suite.

radoering:
I tend to think that we can also mark the tests that require network access if it's too difficult to avoid network access. (Of course, the long-term goal would still be to avoid network access.)

radoering:
@hrnciar Did you test that this solves your issue? It seems that `pytest -m "not network"` does not work due to https://github.com/python-poetry/poetry/blob/fc573231e46ae2a90cfceba0c0406183b536bb47/tests/conftest.py#L67-L68

hrnciar:
@radoering I am sorry for the late response, I was swamped with other things.

> @hrnciar Did you test that this solves your issue? It seems that `pytest -m "not network"` does not work due to
> 
> https://github.com/python-poetry/poetry/blob/fc573231e46ae2a90cfceba0c0406183b536bb47/tests/conftest.py#L67-L68

I was under the impression that it solved my problem, but seems like you are right about `conftest.py`. I've adjusted the PR, could you please take another look?

tests/installation/test_executor.py
```diff
@@ -51,6 +51,8 @@
     from poetry.utils.env import VirtualEnv
     from tests.types import FixtureDirGetter
 
+pytestmark = pytest.mark.network
+
```
radoering, line 55:
```suggestion
```

radoering (approved):

Issue #8303, opened 2023-08-12
Environment markers with different sources for a dependency


sassanh:
- **Poetry version**: 1.5.1
- **Python version**: 3.11.4
- **OS version and name**: macOS Sonoma 14.0
- **pyproject.toml**: -

- [x] I am on the [latest](https://github.com/python-poetry/poetry/releases/latest) stable Poetry version, installed using a recommended method.
- [x] I have searched the [issues](https://github.com/python-poetry/poetry/issues) of this repo and believe that this is not a duplicate.
- [x] I have consulted the [FAQ](https://python-poetry.org/docs/faq/) and [blog](https://python-poetry.org/blog/) for any relevant entries or release notes.
- [x] If an exception occurs when executing a command, I executed it again in debug mode (`-vvv` option) and have included the output below.

## Issue
I have read https://github.com/python-poetry/poetry/issues/7300, https://github.com/python-poetry/poetry/issues/5205 and https://github.com/python-poetry/poetry/issues/6710. I discussed it in Discord but didn't get any answer. The issue is if I need to install a package from different sources based on the platform, poetry doesn't behave as expected.
So imagine I have this dependency:

```toml
[[tool.poetry.source]]
name = "piwheels"
url = "https://www.piwheels.org/simple/"
priority = "primary"

[[tool.poetry.source]]
name = "PyPI"
priority = "primary"

[tool.poetry.dependencies]
kivy = [
  { version = "^2.2.1", markers = "platform_machine == 'aarch64'", source = "piwheels" },
  { version = "^2.2.1", markers = "platform_machine != 'aarch64'", source = "PyPI" }
]
```

I expect it to install from piwheels on RPi and install from PyPI on my macOS, but it doesn't. It creates this in lock file:

```toml
[[package]]
name = "kivy"
version = "2.2.1"
description = "A software library for rapid development of hardware-accelerated multitouch applications."
optional = false
python-versions = ">=3.7"
files = [
    {file = "Kivy-2.2.1-cp310-cp310-macosx_10_9_universal2.whl", hash = "sha256:5aa8a0ca0708dd6edb0127c91712db7b8aad3aea3a79d498df145b8a7ee0aee6"},
    {file = "Kivy-2.2.1-cp310-cp310-manylinux_2_17_aarch64.manylinux2014_aarch64.whl", hash = "sha256:af45f433208c16a4d0e755fafac5f2238560880a475a8a041ca9d99695a6b166"},
[...]

PR #8304, opened 2023-08-13
installer: respect source if the same version of a package has been locked from different sources

radoering:
# Pull Request Check List

Resolves: #8303

fix from https://github.com/python-poetry/poetry/issues/8303#issuecomment-1676175485 enriched with a testcase (`test_installer_distinguishes_locked_packages_with_same_version_by_source` in `test_installer.py`).

### Notable changes other testcases

I changed `test_explicit_source_dependency_with_direct_origin_dependency` from unlocked to locked because our test setup in this module does not handle locking good enough anymore: `repo` is just a `Repository` but had to be a `LegacyRepository`. However, that would require more effort to change. Since the test has been added in #7973 and locking is already tested in the other test case (in `test_solver.py`) added in the same PR, it does not give any additional value to keep an unlocked version of the test.

### DIFF
--- a/src/poetry/console/commands/show.py
+++ b/src/poetry/console/commands/show.py
@@ -211,8 +211,7 @@ lists all packages available."""
         from poetry.utils.helpers import get_package_version_display_string
 
         locked_packages = locked_repository.packages
-        pool = RepositoryPool(ignore_repository_names=True, config=self.poetry.config)
-        pool.add_repository(locked_repository)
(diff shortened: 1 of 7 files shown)


dimbleby:
Thanks for adding the testcase!

I think the RepositoryPool ought to lose the ability to ignore repository names altogether - I'll send an MR to your branch with a small proposed refactor

dimbleby:
https://github.com/python-poetry/poetry/pull/8305

Secrus (approved):

PR #8311, opened 2023-08-14
FIPS compliance during publish

vpbonev:
As md5 is only used to get additional hash alongside with sha256, a passing of usedforsecurity=False is required. This option was introduced in Python 3.9 so we need to make sure we check whether that version is present

# Pull Request Check List

Resolves: #8310 

### DIFF
--- a/src/poetry/publishing/uploader.py
+++ b/src/poetry/publishing/uploader.py
@@ -114,8 +114,12 @@ class Uploader:
         file_type = self._get_type(file)
 
         blake2_256_hash = hashlib.blake2b(digest_size=256 // 8)
+        # Enable FIPS support if Python version is greater or equal to 3.9
+        if sys.version_info >= (3, 9):
+            md5_hash = hashlib.md5(usedforsecurity=False)
+        else:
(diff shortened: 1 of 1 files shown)


dimbleby:
you forgot to import sys

but: this 100% _is_ used for security, surely?  the whole point of the hash is so that downloaders can later verify that the downloaded package is the same as the one that was uploaded.

agree that this is not actually secure - which is why sha256 is preferred - but `usedforsecurity=False` here is just not true.

compare what twine did - https://github.com/pypa/twine/pull/367 - which is simply not provide the md5 hash when compiled in FIPS mode

jeeftor:
This would be nice to have

dimbleby:
@jeeftor per previous comment, this merge request is quite wrong - and I see no sign that the author intends to fix it.

If _you_ care about this then recommend making a new merge request yourself.

jeeftor:
I was hoping my comment would enhance them to make progress. I just turned off FIPS mode last night and poetry was moving along well enough ... :(

Secrus:
Superseded by #9101

Issue #2863, opened 2020-08-28
pre-commit hooks to always be sync after a git checkout


tetienne:
- [x] I have searched the [issues](https://github.com/python-poetry/poetry/issues) of this repo and believe that this is not a duplicate.
- [x] I have searched the [documentation](https://python-poetry.org/docs/) and believe that my question is not covered.

## Issue

Hi, 

I have several projects where we update regularly the dependencies thank to [dependabot](dependabot.com/). My collaborators and I have to think to redo `poetry update` each time we `git pull` or `git checkout`.

So I wrote a small pre-commit hook to ease this process: https://github.com/tetienne/pre-commit-hooks

It will be interesting to host such hooks directly under https://github.com/python-poetry/poetry or perhaps under a dedicated project https://github.com/python-poetry/pre-commit-hooks.

What do you think?

finswimmer:
Hey @tetienne,

sounds like in interesting idea. In the meantime we provide some pre-commit hooks and I think your suggestion would fit quite nice. Do you mind providing a PR for it?

fin swimmer

tetienne:
Thx for this ping. I didn't notice there was now some hooks hosted here. That's great. 
Sure I will try to open a PR with my suggestion.

Secrus:
@tetienne are you still interested in doing this?

alsmnn:
@Secrus are you already on this? I just wanted to create a PR.

PR #8327, opened 2023-08-17
hooks: add poetry-update hook

alsmnn:
When switching branches keeping dependencies in sync with the lock file can become cumbersome, because it is easy to forget. Therefore poetry supplies a pre-commit hook, that can take care of it. After switching branches or checking out a file `poetry install --sync` is invoked and keeps everything nice and tidy.
To make use of that, you need to either supply `default_install_hook_types` inside of your `.pre-commit-config.yaml` or you need to manually install them with: `pre-commit install --install-hooks -t pre-commit -t post-checkout -t post-merge`.

# Pull Request Check List

Resolves: #2863 

### DIFF
--- a/.pre-commit-hooks.yaml
+++ b/.pre-commit-hooks.yaml
@@ -31,4 +31,3 @@
   pass_filenames: false
   stages: [post-checkout, post-merge]
   always_run: true
-  args: ["--sync"]
--- a/.pre-commit-hooks.yaml
+++ b/.pre-commit-hooks.yaml
@@ -25,7 +25,7 @@
 
 - id: poetry-install
   name: poetry-install
-  description: Synchronize the Poetry dependencies with the lock file if out of sync
+  description: run poetry install to install dependencies from the lock file
   entry: poetry install
   language: python
   pass_filenames: false
--- a/.pre-commit-hooks.yaml
+++ b/.pre-commit-hooks.yaml
@@ -23,8 +23,8 @@
   files: ^(.*/)?poetry\.lock$
   args: ["-f", "requirements.txt", "-o", "requirements.txt"]
 
-- id: poetry-sync
-  name: poetry-sync
+- id: poetry-install
+  name: poetry-install
   description: Synchronize the Poetry dependencies with the lock file if out of sync
   entry: poetry install
   language: python
--- a/.pre-commit-hooks.yaml
+++ b/.pre-commit-hooks.yaml
@@ -31,3 +31,4 @@
   pass_filenames: false
   stages: [post-checkout, post-merge]
   always_run: true
+  args: ["--sync"]
--- a/.pre-commit-hooks.yaml
+++ b/.pre-commit-hooks.yaml
@@ -22,3 +22,12 @@
   pass_filenames: false
   files: ^(.*/)?poetry\.lock$
   args: ["-f", "requirements.txt", "-o", "requirements.txt"]
+
+- id: poetry-update
+  name: poetry-update
+  description: Update the Poetry dependencies if out of sync
+  entry: poetry install --sync
+  language: python
+  pass_filenames: false
+  stages: [post-checkout, post-merge]
+  always_run: true
--- a/.pre-commit-hooks.yaml
+++ b/.pre-commit-hooks.yaml
@@ -23,10 +23,10 @@
   files: ^(.*/)?poetry\.lock$
   args: ["-f", "requirements.txt", "-o", "requirements.txt"]
 
-- id: poetry-update
-  name: poetry-update
-  description: Update the Poetry dependencies if out of sync
-  entry: poetry install --sync
+- id: poetry-sync
+  name: poetry-sync
+  description: Synchronize the Poetry dependencies with the lock file if out of sync
+  entry: poetry install
   language: python
   pass_filenames: false
   stages: [post-checkout, post-merge]
--- a/docs/pre-commit-hooks.md
+++ b/docs/pre-commit-hooks.md
@@ -83,9 +83,9 @@ hooks:
     args: ["--dev", "-f", "requirements.txt", "-o", "requirements.txt"]
 ```
 
-## poetry-sync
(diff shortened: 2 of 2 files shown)


.pre-commit-hooks.yaml
```diff
@@ -22,3 +22,12 @@
   pass_filenames: false
   files: ^(.*/)?poetry\.lock$
   args: ["-f", "requirements.txt", "-o", "requirements.txt"]
+
+- id: poetry-update
+  name: poetry-update
```
radoering, line 27: IMO, this ID/name is a bit misleading since the hook does not run `poetry update`. Maybe, `poetry-sync` would be a better name?

finswimmer: 💯

.pre-commit-hooks.yaml
```diff
@@ -22,3 +22,12 @@
...
+  description: Update the Poetry dependencies if out of sync
+  entry: poetry install --sync
```
finswimmer, line 29: The `--sync` argument should be moved to the [args](https://pre-commit.com/#hooks-args) key. Doing so users can decide to overwrite it.

I wonder if want to use `--sync` by default at all. I personally would expect this, but having a look into responses in our issue tracker, there seems to be a lot of people who don't agree. @radoering, @Secrus, @neersighted Any opinions?

finswimmer (changes requested):

.pre-commit-hooks.yaml
Secrus: Yeah, I would remove it. Personally, I rarely use `--sync` option.

alsmnn: So, if I remove some dependencies and invoke `poetry install` does it result in the same virtual environment as invoking `poetry install --sync`? If so, what is the correct use case for the `--sync` option?

finswimmer: `poetry install` without `--sync` will not remove any dependencies. This is useful for people who have installed additional things in the venv via `pip` and want to keep these things. Personally, I clearly prefer using `--sync`. :smiley:

alsmnn: I would like to put the `--sync` option into the args key, because for me it is the intended behaviour. If I change branches and there is a different lock file, i want to have a clean  slate

radoering: I always use `--sync` so that's fine for me. 😉

We could also add two hooks `poetry-install` (without sync) and `poetry-sync` (with sync). Maybe, we just add `poetry-sync` now and poetry-install later only if someone requests it?

docs/pre-commit-hooks.md
```diff
@@ -83,6 +83,18 @@ hooks:
     args: ["--dev", "-f", "requirements.txt", "-o", "requirements.txt"]
 ```
 
+## poetry-update
+
+The `poetry-update` hook calls the `poetry install --sync` command
```
radoering, line 88: The name has to be updated here, too.

.pre-commit-hooks.yaml
finswimmer: To be consistent with the existing pre-commit hooks, I think it would be a good idea to call this hook `poetry-install` (because it runs `poetry install`). One can add any argument they want  (in you use case `--sync`) to projects pre-commit config then.

radoering: That's fine with me, too. It has the benefit that it's more flexible and we only need one hook. 🤷

.pre-commit-hooks.yaml
```diff
@@ -22,3 +22,13 @@
...
+
+- id: poetry-install
+  name: poetry-install
+  description: Synchronize the Poetry dependencies with the lock file if out of sync
+  entry: poetry install
+  language: python
+  pass_filenames: false
+  stages: [post-checkout, post-merge]
+  always_run: true
+  args: ["--sync"]
```
finswimmer, line 34: I still think, in respect to the principle of least surprise, we shouldn't add any default arguments. Users might be confused here two times. One time because when reading "poetry-install" they think of the default behavior of `poetry install`. And a second time, when a project don't want to use `--sync`  they have to overwrite the `args` within the pre-commit config with an empty list. This might look unnecessary to people reading this config.

alsmnn: Now I got you and thanks for the patience.

docs/pre-commit-hooks.md
```diff
@@ -83,6 +83,18 @@ hooks:
     args: ["--dev", "-f", "requirements.txt", "-o", "requirements.txt"]
 ```
 
+## poetry-install
+
+The `poetry-install` hook calls the `poetry install --sync` command
```
radoering, line 88:
```suggestion
The `poetry-install` hook calls the `poetry install` command
```

docs/pre-commit-hooks.md
```diff
@@ -83,6 +83,18 @@ hooks:
...
+to make sure the installed dependencies match the packages defined in `poetry.lock`.
```
radoering, line 89: ... and this still describes `poetry install --sync` so maybe just say that "all locked packages are installed".

alsmnn:
Anything else I could improve?

.pre-commit-hooks.yaml
```diff
@@ -22,3 +22,12 @@
   pass_filenames: false
   files: ^(.*/)?poetry\.lock$
   args: ["-f", "requirements.txt", "-o", "requirements.txt"]
+
+- id: poetry-install
+  name: poetry-install
+  description: Synchronize the Poetry dependencies with the lock file if out of sync
```
radoering, line 28: Another spot that reads more like `poetry install --sync`. Considering the descriptions of the other hooks, maybe:

```suggestion
  description: run poetry install to install dependencies from the lock file
```

radoering (approved):
LGTM. If @finswimmer agrees we can merge it.

finswimmer (approved):

Issue #7990, opened 2023-05-23
Ship Poetry releases with pinned dependencies


johnthagen:
- [x] I have searched the [issues](https://github.com/python-poetry/poetry/issues) of this repo and believe that this is not a duplicate.
- [x] I have searched the [FAQ](https://python-poetry.org/docs/faq/) and general [documentation](https://python-poetry.org/docs/) and believe that my question is not already covered.

## Feature Request

I propose that Poetry itself is distributed with locked/pinned dependencies for each release. This makes each Poetry release reproducible and consistent for all users. This treats Poetry more like an application than a library/package, which is how users interact with it.

Currently Poetry gets installed as a Python package with bounds on its dependencies rather than being pinned/locked. This means that if User A installs Poetry version `X`, and later User B installs version `X`, and User C upgrades to version `X` from version `W`, all three of these users can result in different Poetry installations with subtly different dependencies.

This can lead to some hard to undderstand/fix issues such as:

- #7344

Within teams, it can also make it harder to debug why one developer's Poetry is acting in an odd way with the same version of Poetry than anothers, because their transitive dependencies Poetry is riding on top of are different.

I've noticed that if you simply run `poetry update` to update Poetry [as described in the docs](https://python-poetry.org/docs/1.4#installing-with-the-official-installer), then your dependencies tend to lag behind what you'd get if you did a fresh install or instead ran the more verbose `poetry self lock && poetry self update`.

I believe this could also help make triaging tickets better, as maintainers could know that if a user reports an issue with Poetry version `X`, the maintainer can reproduce the exact transitive dependencies that user has.

In this idea, we could remove `poetry self lock`, `poetry self install`, etc. and simply use `poetry update` to look for a new Poetry version, and install it and it's locked dependencies into the environment.

finswimmer:
Hey @johnthagen,

this have been requested already earlier. While we agree this would be useful, we haven't found a working solution here. The main problems are poetry plugins. If we pin Poetry's own dependencies, these versions might be incompatible to the version a plugin needs.

Suggestions on how to solve this are welcome.

fin swimmer

johnthagen:
Thanks for the feedback. That is a good point.

Would it help at all if there was a scheme for poetry plugins to target supported Poetry versions? Basically, put the responsibility on the plugin to have dependency compatibility with a specific version of Poetry.

[...]

PR #8336, opened 2023-08-21
Use requests^2.26

ysk24ok:
[src/poetry/utils/helpers.py](https://github.com/python-poetry/poetry/blob/1.6.0/src/poetry/utils/helpers.py) imports `atomic_open` from `requests.utils` but the function exists only with requests 2.26.0 or newer.

:o: https://github.com/psf/requests/blob/v2.26.0/requests/utils.py 
:x: https://github.com/psf/requests/blob/v2.25.1/requests/utils.py

# Pull Request Check List

Resolves: #issue-number-here

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -50,7 +50,7 @@ pexpect = "^4.7.0"
 pkginfo = "^1.9.4"
 platformdirs = "^3.0.0"
 pyproject-hooks = "^1.0.0"
-requests = "^2.18"
+requests = "^2.26"
 requests-toolbelt = ">=0.9.1,<2"
 shellingham = "^1.5"
 tomli = { version = "^2.0.1", python = "<3.11" }


ysk24ok:
Would it be possible to release the next patch version (1.16.1) after this PR is merged as this is blocking us?

gsingh1:
All our builds our failing since the new release too - I'll take this opportunity to start pinning poetry installs to versions 🙃

gsingh1:
Opened an issue for this here - https://github.com/python-poetry/poetry/issues/8338

dimbleby:
> this is blocking us

fix looks fine, but I'm curious as to how this is blocking.  poetry is certainly forgetting to force you to update requests - but it isn't _blocking_ you from doing so

klen:
As example: I use `pip install poetry` and `actions/setup-python` with poetry cache in my github actions. And from poetry 1.6 it breaks for the step `actions/setup-python` when the action trying to do `poetry config --list`

dimbleby:
ah, presumably you're picking up an older `requests` from your system packages

Couple of points then

- all documented methods for installing poetry result in it being in its own virtual environment: just `pip install poetry` is not one of those methods.  So you shouldn't be stuck with the system requests at all
- either way, nothing is stopping you from upgrading requests anyway

again this fix looks fine and I assume it'll be merged in due course; but no-one need be "blocked" by it.

(as others have noted, pinning dependencies in your CI pipelines was obviously a good idea anyway)

radoering (approved):

samypr100:
Likely related to https://github.com/python-poetry/poetry/issues/7990, or at least it would help here

allanlewis:
For future reference, `requests.utils.atomic_open` was added in #psf/requests#5707 via psf/requests@24630743cecd3d0e9db7bee7e47381dbadf2d789.

Issue #6035, opened 2022-07-19
Poetry wants to reinstall every system package to the venv when using virtualenvs.options.system-site-packages


awoimbee:
- [x] I am on the [latest](https://github.com/python-poetry/poetry/releases/latest) Poetry version.
- [x] I have searched the [issues](https://github.com/python-poetry/poetry/issues) of this repo and believe that this is not a duplicate.
- [x] If an exception occurs when executing a command, I executed it again in debug mode (`-vvv` option).

- **Poetry version**: 1.2.0b3

## Issue

When I set `virtualenvs.options.system-site-packages` to `true`, my program running inside the poetry venv has access to the system deps (good). But when I run `poetry install`, poetry wants to re-download and re-install everything (bad).

Either I use `virtualenvs.options.system-site-packages true` and poetry should use the system deps (unless not possible, eg not if version mismatch). Or I use `virtualenvs.options.system-site-packages false` and system deps are never used.

Currently If I want to use a dependency from the system site-packages I have to remove it from pyproject.toml and poetry.lock before `poetry install`.

## To reproduce

```bash
docker run -it --rm python:3.10 bash

pip install poetry==1.2.0b3
poetry config virtualenvs.options.system-site-packages true
cat <<EOT > pyproject.toml
[tool.poetry]
name = "gros-zozo"
version = "0.0.0"
description = "saussice"
authors = ["superman"]

[tool.poetry.dependencies]
python = "^3.8"
six = "*"

[build-system]
requires = ["poetry>=1.1"]
build-backend = "poetry.masonry.api"
EOT
poetry install --dry-run
```
output is:
```log
Creating virtualenv gros-zozo-il7asoJj-py3.10 in /root/.cache/pypoetry/virtualenvs
Updating dependencies
Resolving dependencies... (0.3s)

Package operations: 0 installs, 1 update, 0 removals

  • Updating six (1.16.0 /usr/local/lib/python3.10/site-packages -> 1.16.0)
```

Here I use `--dry-run` for the example, but without `--dry-run` poetry downloads the packages then installs then to the venv (but I already have these packages ! I don't want to re-download everything !).

see https://github.com/python-poetry/poetry/issues/1393#issuecomment-1189239876

awoimbee:
1. I tracked down where poetry triggers an update:
[...]

PR #8359, opened 2023-08-24
Consider system packages as installed if the venv includes them

abompard:
This changesets adds a getter to `env.virtual_env.VirtualEnv` to check whether it has access to system packages, and overrides the `is_path_relative_to_lib` method to take it into account.

This will prevent Poetry from reinstalling system packages in the venv when they are already installed with a compatible version.

Resolves: #6035

### DIFF
--- a/src/poetry/utils/env/virtual_env.py
+++ b/src/poetry/utils/env/virtual_env.py
@@ -3,6 +3,7 @@ from __future__ import annotations
 import json
 import os
 import re
+import sys
 
 from contextlib import contextmanager
 from copy import deepcopy
@@ -20,6 +21,7 @@ from poetry.utils.env.script_strings import GET_PATHS
 from poetry.utils.env.script_strings import GET_PYTHON_VERSION
 from poetry.utils.env.script_strings import GET_SYS_PATH
 from poetry.utils.env.script_strings import GET_SYS_TAGS
+from poetry.utils.env.system_env import SystemEnv
 
 
 if TYPE_CHECKING:
@@ -133,3 +135,14 @@ class VirtualEnv(Env):
 
     def _updated_path(self) -> str:
         return os.pathsep.join([str(self._bin_dir), os.environ.get("PATH", "")])
+
+    def includes_system_site_packages(self) -> bool:
+        pyvenv_cfg = self._path / "pyvenv.cfg"
+        return "include-system-site-packages = true" in pyvenv_cfg.read_text()
+
+    def is_path_relative_to_lib(self, path: Path) -> bool:
+        system_env = SystemEnv(Path(sys.prefix))
+        return (
+            self.includes_system_site_packages()
+            and system_env.is_path_relative_to_lib(path)
+        ) or super().is_path_relative_to_lib(path)
--- a/tests/utils/test_env.py
+++ b/tests/utils/test_env.py
@@ -18,6 +18,7 @@ from poetry.factory import Factory
 from poetry.repositories.installed_repository import InstalledRepository
 from poetry.toml.file import TOMLFile
 from poetry.utils._compat import WINDOWS
+from poetry.utils._compat import metadata
 from poetry.utils.env import GET_BASE_PREFIX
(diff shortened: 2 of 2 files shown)


abompard:
The `macOS / 3.11` error looks like a temporary CI error.

src/poetry/utils/env/virtual_env.py
```diff
@@ -133,3 +135,14 @@ def temp_environ(self) -> Iterator[None]:
...
+    def includes_system_site_packages(self) -> bool:
+        pyvenv_cfg = self._path / "pyvenv.cfg"
+        return "include-system-site-packages = true" in pyvenv_cfg.read_text()
+
+    def is_path_relative_to_lib(self, path: Path) -> bool:
+        system_env = SystemEnv(Path(sys.prefix))
+        return (
+            self.includes_system_site_packages()
+            and system_env.is_path_relative_to_lib(path)
+        ) or super().is_path_relative_to_lib(path)
```
radoering, line 148:
```suggestion
        return super().is_path_relative_to_lib(path) or (
            self.includes_system_site_packages()
            and SystemEnv(Path(sys.prefix)).is_path_relative_to_lib(path)
        )
```

That way we don't have to create a SystemEnv and call additional methods in the default case (where system site packages are not included).

src/poetry/utils/env/virtual_env.py
```diff
@@ -133,3 +135,14 @@ def temp_environ(self) -> Iterator[None]:
 
     def _updated_path(self) -> str:
         return os.pathsep.join([str(self._bin_dir), os.environ.get("PATH", "")])
+
+    def includes_system_site_packages(self) -> bool:
+        pyvenv_cfg = self._path / "pyvenv.cfg"
+        return "include-system-site-packages = true" in pyvenv_cfg.read_text()
```
radoering, line 141: Maybe, we should make this a `cached_property` since this should not change during one invocation of poetry. Now, we are reading `pyvenv.cfg` once for each package.

radoering (commented):
IMO, it would be nice to have a test in `test_installed_repository.py` that checks that a package from system site packages does not have `source_type == "directory"`.

abompard:
> IMO, it would be nice to have a test in `test_installed_repository.py` that checks that a package from system site packages does not have `source_type == "directory"`.

I thought that as well, but in that file the virtual environment is entirely mocked, so I thought it wouldn't be terribly useful as I didn't change any of the calls in `repository`. What do you think?

radoering:
I think it makes sense if we set up the test a little differently. Please take a look at the test I just added.

abompard:
> I think it makes sense if we set up the test a little differently. Please take a look at the test I just added.

Oh yeah that's a good idea! Thanks :-)

src/poetry/utils/env/virtual_env.py
```diff
@@ -133,3 +136,14 @@ def temp_environ(self) -> Iterator[None]:
 
     def _updated_path(self) -> str:
         return os.pathsep.join([str(self._bin_dir), os.environ.get("PATH", "")])
+
+    @cached_property
+    def includes_system_site_packages(self) -> bool:
+        pyvenv_cfg = self._path / "pyvenv.cfg"
+        return "include-system-site-packages = true" in pyvenv_cfg.read_text()
```
radoering, line 143: This line seems a bit fragile so I wondered if `pyvenv.cfg` is some well defined format. At first, I did not find a specification, but the `site` module being more robust considering spaces and case-sensitivity:

https://github.com/python/cpython/blob/042aa88bcc6541cb8b312f1119452f7a58a5b4df/Lib/site.py#L515-L522

Then, I found PEP 405, which describes how the `site` module parses it so I think we should be more robust, too.

radoering (changes requested):

radoering (approved):

niniack:
Is the expected behavior that poetry will not try to update system packages, as long as they satisfy the dependency resolution?

PR #8363, opened 2023-08-24
Optionally only install the group holding the incoming dependencies

rjyunis:
This pull request adds the `--only` flag to `poetry add`. The rationale is to counteract the following situation:

```
poetry install --without=somegroup
poetry add new-dependency
# Poetry now installs new-dependency as well as dependencies in somegroup
```

By invoking `--only` the group of the incoming dependencies is passed to the install command via `self.installer.only_groups(group)`.

This is the best solution/simplicity ratio I have found to mimic the current workaround
```
poetry add new-dependency --lock
poetry install --without=somegroup
```

Tests incoming.
Doc updates incoming. 

# Pull Request Check List

Resolves: #issue-number-here

### DIFF
--- a/src/poetry/console/commands/add.py
+++ b/src/poetry/console/commands/add.py
@@ -68,7 +68,7 @@ class AddCommand(InstallerCommand, InitCommand):
             " --verbose).",
         ),
         option("lock", None, "Do not perform operations (only update the lockfile)."),
-        option("only" None, "Do not install additional packages from the toml."),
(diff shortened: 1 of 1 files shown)


abn:
This is stale atm. I am not sure if this is still needed either. If it is needed and someone decides to implement it, I suspect the better way to approach this is to alter the `add` default implementation such that it does the "workaround" in that it does a lock, then only activates dependencies in the given package's dependency graph.

Issue #1132, opened 2019-05-28
Support for application-first, library-first, and non-package modes that change Poetry defaults


thejohnfreeman:
- [x] I have searched the [issues](https://github.com/sdispater/poetry/issues) of this repo and believe that this is not a duplicate.
- [x] I have searched the [documentation](https://poetry.eustace.io/docs/) and believe that my question is not covered.

I'd like to be able to use Poetry to manage a virtual environment for a project that I never intend to package or publish.

My particular use case is that I want to use Python for testing a project written for another language. I just want to use Python as a cross-platform scripting language that comes with a nice test framework (pytest). I don't want to have to give my project a name or version in `pyproject.toml`, but Poetry refuses to create a virtual environment or install any dependencies until they are given. Pipenv doesn't have this limitation (because it doesn't try to handle packaging, which would require a name and version), so I'm using it in the meantime. I love Poetry, and I would rather use one tool, so please consider limiting name and version checks to the methods that absolutely need them.

jacebrowning:
The `--no-interaction` mode (`-n`) will infer the name from the directory:

```
$ cd <my_project>
$ poetry init -n
$ poetry add pytest
$ poetry run pytest
```

This has worked for my applications projects (e.g. Django) that will never be packaged.

thejohnfreeman:
Yes, and the version is defaulted. In my use case, this is sitting in a project that has a name and version in another language's ecosystem. Choosing a name is not difficult. There are two good reasons I don't want to have to give a version, though:

- I don't want the "dummy version" in `pyproject.toml` to confuse newcomers into thinking that there is a Python package cohabiting with the rest of the project. The project has a version that is tracked in another file specific to the other language's ecosystem.
- I don't want me or someone else to accidentally run any commands that might try to package or publish this project to PyPI. I want Poetry to warn that packaging is not possible for this project.

And If I'm not giving a version to Poetry, I might as well not give a name either.

mwchase:
As a general workaround for the second point, you could add a trove classifier starting with "Private ::", per https://github.com/pypa/warehouse/pull/5440

shelper:
I also want the option to configure the project that poetry manages as a non-package. 
not always want to package the project, e.g., could be a ML or DS project

neersighted:
Renamed to better capture the intent and make this the round-up issue for related duplicates.

johnthagen:
[...]

PR #8369, opened 2023-08-26
install: add warning if current project cannot be installed

radoering:
Companion to python-poetry/poetry-core#629

- `ModuleOrPackageNotFound` is now raised in `EditableBuilder.build()` instead of `EditableBuilder.__init__()`
- The existence of a specified readme is now checked before the existence of the module. -> Just not declare a non-existing readme in the test.

**Update:** In the second commit, I changed the current behavior so that errors are no longer swallowed silently when the installation of the current project fails. I only added a warning and did not change the return code to avoid a breaking change. We can decide if we only want to go with the first commit or both commits.

Pro second commit: Currently, the user will not know if installing the current project succeeds or fails.

Contra second commit: Users who don't use `--no-root` and don't care about the current project being installed or not, might be annoyed by the warning.

### DIFF
--- a/src/poetry/console/commands/install.py
+++ b/src/poetry/console/commands/install.py
@@ -162,10 +162,16 @@ dependencies and not including the current project, run the command with the
         # i.e. during EditableBuilder.build().
         try:
             builder = EditableBuilder(self.poetry, self.env, self.io)
-        except ModuleOrPackageNotFound:
+        except (ModuleOrPackageNotFound, FileNotFoundError) as e:
             # This is likely due to the fact that the project is an application
             # not following the structure expected by Poetry
             # If this is a true error it will be picked up later by build anyway.
+            self.line_error(
+                f"The current project could not be installed: <error>{e}</error>\n"
+                "If you do not want to install the current project"
+                " use <c1>--no-root</c1>",
+                style="warning",
+            )
             return 0
 
         log_install = (
@@ -185,10 +191,17 @@ dependencies and not including the current project, run the command with the
 
         try:
             builder.build()
-        except ModuleOrPackageNotFound:
+        except (ModuleOrPackageNotFound, FileNotFoundError) as e:
             # This is likely due to the fact that the project is an application
             # not following the structure expected by Poetry.
             # No need for an editable install in this case.
+            self.line("")
+            self.line_error(
+                f"The current project could not be installed: <error>{e}</error>\n"
+                "If you do not want to install the current project"
+                " use <c1>--no-root</c1>",
+                style="warning",
+            )
             return 0
 
         if overwrite:
--- a/src/poetry/console/commands/install.py
+++ b/src/poetry/console/commands/install.py
@@ -155,6 +155,11 @@ dependencies and not including the current project, run the command with the
         if self.option("no-root"):
             return 0
 
+        # Prior to https://github.com/python-poetry/poetry-core/pull/629
+        # the existence of a module/package was checked when creating the
+        # EditableBuilder. Afterwards, the existence is checked after
+        # executing the build script (if there is one),
+        # i.e. during EditableBuilder.build().
         try:
             builder = EditableBuilder(self.poetry, self.env, self.io)
         except ModuleOrPackageNotFound:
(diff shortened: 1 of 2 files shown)


rragundez:
Can you update the branch? I guess that will give it more chance to be reviewed

Secrus (approved):

rragundez:
Hi @Secrus can you have a look at this one as well? https://github.com/python-poetry/poetry-core/pull/629

silverwind:
I'm confused about this new warning. What does "install the current project" mean in practice? Does it copy or symlink the current project's files into the virtualenv so that it can be imported by name?

If above is true, I think this behaviour should be removed and users should just be required to install a self-dependency, eliminating this confusing warning for users who don't need this feature.

radoering:
It does an editable install. More specifically, it creates a `*.pth` file in the virtualenv, which probably is comparable to a symlink.

The warning means that your package is probably broken. If you don't care and don't like to add `--no-root`, you are probably interested in a "non-package" mode as described in #1132.

silverwind:
Yes, my "package" contains no code, it's only used to run binaries. I guess I will run with `--no-root` until https://github.com/python-poetry/poetry/issues/1132 is fixed.

adriangb:
Should this be triggered if `packages = []` in `pyproject.toml`? I'd think not but it seems to.

radoering:
`packages = []` is not a reliable indicator to determine if this check should be triggered. If it's empty, there's a [default](https://github.com/python-poetry/poetry-core/blob/8cd4f5d94e2fea1632c1d6ac671922330bbefbfd/src/poetry/core/masonry/utils/module.py#L36). I think the only way is to decide is an explicit attribute, see #1132.

adriangb:
I'm surprised by that. I'd think if the user explicitly specifies `packages = []` then Poetry should respect that and only use a default if no `packages` is supplied at all. That's how it seemed to work from a users perspective prior to this PR (if Poetry tried to install some default and failed silently I don't know). In fact the [docs](https://python-poetry.org/docs/pyproject/#packages) said and currently say:

> Using packages disables the package auto-detection feature meaning you have to explicitly specify the “default” package.

radoering:
> Using packages disables the package auto-detection feature meaning you have to explicitly specify the “default” package.

I suppose the author of this part of the docs did not expect that somebody interprets "using packages" as setting it to an empty list. 😄

> I'd think if the user explicitly specifies packages = [] then Poetry should respect that

On the one hand "yes", but on the other hand: does it make sense to build a distribution without any package/module or do you only use `packages = []` if you never intend to build an sdist/wheel?

In the latter case, I think it's probably better to introduce an explicit attribute to switch between package and non-package mode (cf https://github.com/python-poetry/poetry/issues/1132#issuecomment-1793690699).

adriangb:
> does it make sense to build a distribution without any package/module or do you only use packages = [] if you never intend to build an sdist/wheel?

I don't think it makes sense to build a distribution without packages, although I suppose it shouldn't be disallowed since it's a technically valid distribution and metapackages are a thing. But that's not my use case so I don't want to opine too much. It does make a lot of sense to want to _install_ a meta project with no source code of its own. So at the very least, this error should either be a warning or only apply to building a distribution.

> In the latter case, I think it's probably better to introduce an explicit attribute to switch between package and non-package mode (cf https://github.com/python-poetry/poetry/issues/1132#issuecomment-1793690699).

Perhaps. But my stance is going to be that this was a breaking change on a non major version release and that the change itself is questionable in the first place so it should be at least partially reverted (maybe to only apply to building distributions).

silverwind:
> this error should either be a warning

It is a warning as exit code is 0, but I agree the text output looks deceptively like an error.

adriangb:
Also it seems to me like the goal was to prevent users from shooting themselves in the foot if they named their source code folder wrong or similar. Even if we made an exception for `packages = []` the benefit of this change would be preserved.

radoering:
> this was a breaking change

I don't consider printing a warning a breaking change. (Every change is a breaking change for someone. If we move on this level, every new version is a major version.)

> Also it seems to me like the goal was to prevent users from shooting themselves in the foot if they named their source code folder wrong or similar.

That and in more complicated cases, editable builds can just fail and without this change you will just notice that something does not work in your environment although `poetry install` "did not fail".

> Even if we made an exception for packages = [] the benefit of this change would be preserved.

I agree that we can make an exception for `packages = []` and clarify the docs (although this will be a breaking change for someone even if the current behavior can be interpreted as a bug). 😉

adriangb:
I thought it was an error. I agree a warning is not a breaking change, but still a confusing annoyance.

nikobockerman:
This warning is printed in red so it definitely looks like an error. I also got confused and thought it was an actual error.

tueda:
After getting there and reading this issue, I understand this error-ish warning that contains the format `<error>{e}</error>` (leading to the red text) is actually not an error. Then the question is how the user can distinguish red text for a warning from that for an error. Maybe simply printing the word `warning` would be helpful.

Issue #8375, opened 2023-08-28
Support automatic Python 3.12 PyPI trove classifier


jakob-keller:
- [x] I have searched the [issues](https://github.com/python-poetry/poetry/issues) of this repo and believe that this is not a duplicate.
- [x] I have searched the [FAQ](https://python-poetry.org/docs/faq/) and general [documentation](https://python-poetry.org/docs/) and believe that my question is not already covered.

## Feature Request

Python 3.12 [will be released later this year](https://peps.python.org/pep-0693/) and is already [available as a release candidate](https://pythoninsider.blogspot.com/2023/08/python-3120-release-candidate-1-released.html):

> We strongly encourage maintainers of third-party Python projects to prepare their projects for 3.12 compatibilities during this phase, and where necessary publish Python 3.12 wheels on PyPI to be ready for the final release of 3.12.0.

This would be made easier if poetry-core would automatically add the appropriate PyPI trove classifier.

I will draft a trivial PR in a minute.

PR #8376, opened 2023-08-28
Prepare tests for Python 3.12 PyPI trove classifier

jakob-keller:
# Pull Request Check List

https://github.com/python-poetry/poetry-core/pull/631 adresses #8375, but currently fails poetry downstream tests.

This PR prepares poetry unit tests for https://github.com/python-poetry/poetry-core/pull/631.

### DIFF
--- a/tests/masonry/builders/test_editable_builder.py
+++ b/tests/masonry/builders/test_editable_builder.py
@@ -149,6 +149,7 @@ Classifier: Programming Language :: Python :: 3.8
 Classifier: Programming Language :: Python :: 3.9
 Classifier: Programming Language :: Python :: 3.10
 Classifier: Programming Language :: Python :: 3.11
+Classifier: Programming Language :: Python :: 3.12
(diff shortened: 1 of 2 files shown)


jakob-keller:
Tests are expected to fail. This PR is intended to work in conjuncture with https://github.com/python-poetry/poetry-core/pull/631.

radoering:
I tend to think that we should make these tests more generic so we avoid the dance for each new Python version:
We could change the expectation in the tests to all Python versions in `Package.AVAILABLE_PYTHONS` that are allowed by the test's Python constraint.

jakob-keller:
> I tend to think that we should make these tests more generic so we avoid the dance for each new Python version:
> We could change the expectation in the tests to all Python versions in `Package.AVAILABLE_PYTHONS` that are allowed by the test's Python constraint.

You're absolutely right. I'll try to address that tonight.

jakob-keller:
Superseded by #8397

Issue #8382, opened 2023-08-29
0-byte files getting unpacked


jonathanmorley:
- **Poetry version**: 1.6.1
- **Python version**: 3.9
- **OS version and name**: Ubuntu 22.04
- **pyproject.toml**: See https://github.com/cvent/poetry-empty-file-repro/blob/main/pyproject.toml

- [ ] I am on the [latest](https://github.com/python-poetry/poetry/releases/latest) stable Poetry version, installed using a recommended method.
- [ ] I have searched the [issues](https://github.com/python-poetry/poetry/issues) of this repo and believe that this is not a duplicate.
- [ ] I have consulted the [FAQ](https://python-poetry.org/docs/faq/) and [blog](https://python-poetry.org/blog/) for any relevant entries or release notes.
- [ ] If an exception occurs when executing a command, I executed it again in debug mode (`-vvv` option) and have included the output below.

## Issue

When running `poetry install` against a project, it sometimes 'corrupts' the installation of packages by creating empty, 0-byte files, when they should be full of content.

I have managed to create a minimal-reproduction repository, but in testing, it seems extremely sensitive to alterations (pyproject dependencies, lockfile, etc).

It seems to happen independently of creating a local venv.

I have not observed this happening in other packages, but also havent checked.
`pip install jira` appears to work fine consistently, and `poetry add jira` in an empty project too, so I don't think its a problem with the package itself.

dimbleby:
looks like `jira` and `jira-python` collide.  Nothing much poetry can do about that, please close.

jonathanmorley:
Would you mind expanding on how they collide? How could you tell? Could poetry be able to identify this collision and raise an error? As it stands, poetry appears to succeed installing them both together

dimbleby:
both packages expect to be used as `import jira`, both packages contain `jira/__init__.py`.  At installation sometimes one wins and sometimes the other.

You can inspect the packages that you download eg from https://pypi.org/project/jira-python/#files to verify this.

No, poetry can't know this in advance.

Please close.

PR #8383, opened 2023-08-29
simplify WheelDestination

dimbleby:
`WheelDestination.for_source()` is in my opinion odd: it looks as though some code was attached to it just because it could be.

- we construct a placeholder `WheelDestination` with a bunch of values that will be re-used for all wheels being installed
- but then every time we actually want a `WheelDestination` we construct a new one, never meaningfully using the placeholder object for its intended purpose - ie as `WheelDestination` - at all.

Might as well simply put the common reusable values on the `WheelInstaller` directly.

---

Actually my original hope had been to remove the `WheelDestination` altogether.  Its purpose seems to be to contain an override of `write_to_fs()` that tolerates duplicate files, which - while also being what `pip` does - allows confusion such as #8382.

The comment explaining why this is necessary is wrong, properly written native namespace packages - including `poetry` and `poetry-core` - do not provide duplicate `__init__.py` (or anything else).

However distributions of other styles of namespace packages are apparently supposed to provide identical duplicate `__init__.py`.  (I suppose a possible future fix would be to allow overwrites only if the contents were identical.)


src/poetry/installation/wheel_installer.py
```diff
@@ -85,18 +74,13 @@ def __init__(self, env: Env) -> None:
                 script_kind = "win-arm64" if sys.maxsize > 2**32 else "win-arm"
             else:
                 script_kind = "win-amd64" if sys.maxsize > 2**32 else "win-ia32"
+        self.script_kind = script_kind
```
radoering, line 77:
```suggestion
        self._script_kind = script_kind
```
to make the interface clearer (same for `bytecode_optimization_levels`)

radoering (approved):

PR #8389, opened 2023-08-30
Limit default max-workers to 8

coltonb-arcadia:
# Pull Request Check List

Resolves: #3219

When `os.cpu_count()` reports a large number of cores, "Connection reset by peer" errors seem to be raised fairly often because of the resulting huge number of requests being made. It struck me as more intuitive for the end user to use a safer value as the default maximum, rather than letting it be unbound to whatever the executing machine reports.

With this change, a power user can still configure a larger number of max workers if it is desirable and choose the upper limit themselves.

I do not have any specific data on the number of workers to cause this error to start cropping up due to its non-deterministic nature, so 8 was chosen conservatively. 5 might be best to make the default behavior be only a single core.

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -179,7 +179,7 @@ If you encounter any problems with it, set to `true` to use the system git backe
 
 **Type**: `int`
 
-**Default**: `number_of_cores + 4`
+**Default**: `min(number_of_cores + 4, 8)`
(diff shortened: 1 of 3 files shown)


coltonb-arcadia:
Closing this until confirmed this mitigates the issue.

PR #8394, opened 2023-09-01
bugfix: user-agent python-requests

mtricolici98:
Replaced user agent to avoid bot protection on private repositories.

For lock and install commands, the requests library was used, resulting in HTTP requests with 'python-requests' User-agent being sent. 

It is possible that HTTP Servers (such as OPNSens in my case) have 'bot protection' to automatically block "bot-like" user agents. 

This quick fix comes as a counter measure for the following problem. 

# Pull Request Check List

Resolves: No tracked issue

- No code tests updated
- No documentation updates necessary


src/poetry/utils/authenticator.py
```diff
@@ -193,7 +195,8 @@ def authenticated_url(self, url: str) -> str:
     def request(
         self, method: str, url: str, raise_for_status: bool = True, **kwargs: Any
     ) -> requests.Response:
-        headers = kwargs.get("headers")
+        headers = kwargs.get("headers", {}) or {}
```
dimbleby, line 198: `or {}` isn't doing anything useful

cf https://github.com/python-poetry/poetry/blob/02448cf7f184dea204156f7dcb620a4f01a0068e/src/poetry/publishing/uploader.py#L82, which does this on the session - it would be consistent here to do the work in `get_session()`.

dimbleby (commented):
this all seems a bit silly to me: a repository that doesn't allow downloads from the requests user agent is broken and should be fixed (or reconfigured)

however it also seems harmless enough I suppose

mtricolici98:
> this all seems a bit silly to me: a repository that doesn't allow downloads from the requests user agent is broken and should be fixed (or reconfigured)
> 
> however it also seems harmless enough I suppose

I do agree, it took me a while to figure out that the 403 I was receiving wasn't from bad credentials or authorization but from the user agent. It seems like our repository is proxied through a nginx configuration that blocks some "bot like" user agents, as to "protect" from vulnerability scans or other bot activity.  A quick `curl` with 2 different user agent headers cleared that one for me. However the dev-ops team on my project is reluctant to disable the bot protection.
Thus I created a quite silly plugin to inject the ua headers, and thought it might be fine to also solve this at the source, especially since poetry provides his own user agent during the publishing step, it makes sense that for consistency it will also use the `poetry` UA.

src/poetry/utils/authenticator.py
mtricolici98: I think `or {}` covers the potential of there being a `headers` key in the `dict` with a `None` value. 


I will apply the UA to the session itself. Thank you for the suggestion.

src/poetry/utils/authenticator.py
```diff
@@ -134,9 +136,11 @@ def __init__(
             self._get_repository_config_for_url
         )
         self._pool_size = pool_size
+        self.user_agent = user_agent("poetry", __version__)
```
radoering, line 139: I'd either make it a property (as in `Uploader`) or prepend an underscore (`self._user_agent`).

PR #8401, opened 2023-09-04
🧹 clean `poetry.vcs.git.system` module

tzing:
Clean up duplicated logic in `poetry.vcs.git.system` module and add test for it.

### Pull Request Check List

Relates-to: #3155


tests/vcs/git/test_git_system.py
```diff
@@ -0,0 +1,92 @@
...
+    )
+
+    # add redundant commit
+    stdout = subprocess.check_output(
+        [
+            "git",
+            "commit",
+            "--allow-empty",
+            "--message=test commit",
+        ],
```
radoering, line 55:
```suggestion
            "--no-gpg-sign",  # avoid user interaction in case signing is turned on
        ],
```

radoering (changes requested):
The `clone` test takes quite long. It might make sense to create a new temporary repo instead of using the poetry repo? (I could imagine it may be faster to clone a repo with only some commits.)

Btw, I think we should name the test file `test_system.py` according to the module we are testing. It's in the git folder anyway.

tzing:
> The clone test takes quite long. It might make sense to create a new temporary repo instead of using the poetry repo?

Okay I can try.

Convert to draft and will work on it later.

radoering (approved):

PR #8478, opened 2023-09-27
poetry shell: fix nushell

dead10ck:
Presently, there are two bugs with `poetry shell` on nushell:

1. The `overlay use` command that poetry sends to the subshell does not get run, but merely appears on the command line as if the user had typed it themselves. Hitting the enter key is still necessary to activate the virtualenv.
2. In the subshell, whenever the user uses a tool that requires interactive keyboard input, the input is not echoed on the screen. This is because the current code specifically and explicitly turns off the terminal echo mode. It is unclear why.

This change fixes both of these problems by changing the invocation of `nu` to run the `overlay use` command with the `-e` flag, which runs the given command and then starts an interactive shell.

Disabling echo mode is also removed.

### DIFF
--- a/src/poetry/utils/shell.py
+++ b/src/poetry/utils/shell.py
@@ -97,12 +97,16 @@ class Shell:
         import shlex
 
         terminal = shutil.get_terminal_size()
+        cmd = f"{self._get_source_command()} {shlex.quote(str(activate_path))}"
+
         with env.temp_environ():
+            args = ["-e", cmd] if self._name == "nu" else ["-i"]
+
             c = pexpect.spawn(
-                self._path, ["-i"], dimensions=(terminal.lines, terminal.columns)
+                self._path, args, dimensions=(terminal.lines, terminal.columns)
             )
 
-        if self._name in ["zsh", "nu"]:
+        if self._name in ["zsh"]:
             c.setecho(False)
 
         if self._name == "zsh":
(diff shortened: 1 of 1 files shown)


src/poetry/utils/shell.py
```diff
@@ -97,22 +97,27 @@ def activate(self, env: VirtualEnv) -> int | None:
...
+        if self._name in ["zsh"]:
             c.setecho(False)
 
         if self._name == "zsh":
             # Under ZSH the source command should be invoked in zsh's bash emulator
             c.sendline(f"emulate bash -c '. {shlex.quote(str(activate_path))}'")
         elif self._name == "xonsh":
             c.sendline(f"vox activate {shlex.quote(str(env.path))}")
+        elif self._name == "nu":
+            pass
```
radoering, line 118: Can you add an explanatory comment for the `pass`, please?

src/poetry/utils/shell.py
```diff
@@ -97,22 +97,27 @@ def activate(self, env: VirtualEnv) -> int | None:
...
         else:
-            cmd = f"{self._get_source_command()} {shlex.quote(str(activate_path))}"
-            if self._name in ["fish", "nu"]:
+            if self._name in ["fish"]:
                 # Under fish and nu "\r" should be sent explicitly
```
radoering, line 121: The comment should be updated, too.

radoering (changes requested):

dead10ck:
Done and done, thanks for the review!

src/poetry/utils/shell.py
```diff
@@ -97,23 +97,31 @@ def activate(self, env: VirtualEnv) -> int | None:
...
         if self._name == "zsh":
             # Under ZSH the source command should be invoked in zsh's bash emulator
             c.sendline(f"emulate bash -c '. {shlex.quote(str(activate_path))}'")
         elif self._name == "xonsh":
             c.sendline(f"vox activate {shlex.quote(str(env.path))}")
+
+        # if this is nu, we don't want to send the activation command to the
+        # command line since we already ran it via the shell's invocation
+        elif self._name == "nu":
+            pass
```
radoering, line 121:
```suggestion
        elif self._name == "nu":
            # If this is nu, we don't want to send the activation command to the
            # command line since we already ran it via the shell's invocation.
            pass
```

radoering (approved):

Issue #8491, opened 2023-09-29
Documentation for POETRY_REPOSITORIES_<NAME> missing _URL suffix


SamambaMan:
Hello!

The documentation is not precise on how to configure repositories via variables.
In the docs on https://github.com/python-poetry/poetry/blob/081146956349e24d901c5187685df345e6621a98/docs/configuration.md?plain=1#L444 the variable should be configured as `POETRY_REPOSITORIES_<NAME>` but in the code on https://github.com/python-poetry/poetry/blob/081146956349e24d901c5187685df345e6621a98/src/poetry/config/config.py#L200 expects a pattern like:
`pattern = re.compile(r"POETRY_REPOSITORIES_(?P<name>[A-Z_]+)_URL")`
With _URL suffix.
So configuring as documented won't work.

Thanks and keep up the good work

SamambaMan:
Done, thanks

PR #8492, opened 2023-09-29
Fix POETRY_REPOSITORIES_ variable docs

SamambaMan:
POETRY_REPOSITORIES_<NAME>_URL wrong in the documentation

# Pull Request Check List

Resolves: #8491

No changes are required to the code.

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -441,7 +441,7 @@ The variables `project_name` and `python_version` are available for formatting.
 
 **Type**: `string`
 
-**Environment Variable**: `POETRY_REPOSITORIES_<NAME>`
+**Environment Variable**: `POETRY_REPOSITORIES_<NAME>_URL`
 
 Set a new alternative repository. See [Repositories]({{< relref "repositories" >}}) for more information.
 
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -445,7 +445,7 @@ The variables `project_name` and `python_version` are available for formatting.
 
 Set the repository URL for `<name>`.
 
-See [Repositories]({{< relref "repositories" >}}) for more information.
+See [Publishable Repositories]({{< relref "repositories#publishable-repositories" >}}) for more information.
 
 ### `http-basic.<name>.[username|password]`:
 
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -437,19 +437,21 @@ If set to `false`, Python version used during Poetry installation is used.
 Format string defining the prompt to be displayed when the virtual environment is activated.
 The variables `project_name` and `python_version` are available for formatting.
 
-### `repositories.<name>`
+### `repositories.<name>.url`
 
 **Type**: `string`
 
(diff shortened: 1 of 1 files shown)


dimbleby:
it's not so simple as that: `POETRY_REPOSITORIES_<NAME>_URL` is specifically for setting the url associated with a repository ie it corresponds to `repositories.<name>.url`.

I can believe that there are clarity improvements to be made here and it would be good to see that: but I don't think this MR as it stands helps with that

SamambaMan:
What do you suggest?
I merely wrote the same as it in the code (i think) for doing as the manual won't work.

Do you want me to explain that the <name> should match naming in sources?

dimbleby:
the current doc has a heading `repositories.name`.  You are reporting - I think rightly - that as an environment variable this isn't very useful because you never want to configure only a repository name: you want to configure sub-properties of the repository (in particular, the URL).

However your MR leaves that heading alone, so now the doc is not self-consistent.

I realise that disentangling this is more than the four character fix you were hoping for!

SamambaMan:
Oh, i get it now.
This should cover other parameters as priority as well, right?
I'm hovering the code and this is not supported code wise yet also, right?

dimbleby:
As I understand it, global repository configuration is only interesting during publishing, so eg priority is irrelevant - I think url is the only interesting property.

SamambaMan:
So...? :D

dimbleby:
So: put some or more of that in the docs!

SamambaMan:
@dimbleby check if this is better.
Also included fixes for basic auth as tested in https://github.com/python-poetry/poetry/blob/47a3a19c442c21aa5b2836ea092dda8ae4d13151/tests/utils/test_authenticator.py#L355

SamambaMan:
Can I improve this PR in any way?
Let me know

docs/configuration.md
```diff
@@ -437,19 +437,21 @@ If set to `false`, Python version used during Poetry installation is used.
...
-### `repositories.<name>`
+### `repositories.<name>.url`
 
 **Type**: `string`
 
-**Environment Variable**: `POETRY_REPOSITORIES_<NAME>`
+**Environment Variable**: `POETRY_REPOSITORIES_<NAME>_URL`
 
-Set a new alternative repository. See [Repositories]({{< relref "repositories" >}}) for more information.
+Set a new alternative repository URL for `name` repository
```
radoering, line 446:
```suggestion
Set the repository URL for `<name>`.
```

docs/configuration.md
```diff
@@ -437,19 +437,21 @@ If set to `false`, Python version used during Poetry installation is used.
...
 
-### `http-basic.<name>`:
+See [Repositories]({{< relref "repositories" >}}) for more information.
```
radoering, line 448:
```suggestion
See [Publishable Repositories]({{< relref "repositories#publishable-repositories" >}}) for more information.
```

radoering (changes requested):

radoering (approved):

PR #8502, opened 2023-10-03
fix: Correct error message when pypi-token.pypi has incorrect number of arguments.

radoshi:
# Pull Request Check List

Resolves: #8503

### DIFF
--- a/src/poetry/console/commands/config.py
+++ b/src/poetry/console/commands/config.py
@@ -122,6 +122,9 @@ To remove a repository (repo is a short alias for repositories):
 
         # show the value if no value is provided
         if not self.argument("value") and not self.option("unset"):
+            if setting_key.split(".")[0] in self.LIST_PROHIBITED_SETTINGS:
+                raise ValueError(f"Expected a value for {setting_key} setting.")
+
             m = re.match(r"^repos?(?:itories)?(?:\.(.+))?", self.argument("key"))
             value: str | dict[str, Any]
             if m:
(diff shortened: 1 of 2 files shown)


src/poetry/console/commands/config.py
```diff
@@ -138,6 +138,8 @@ def handle(self) -> int:
 
                 self.line(str(value))
             else:
+                if setting_key.split(".")[0] in self.LIST_PROHIBITED_SETTINGS:
+                    raise ValueError(f"Expected a value for {setting_key} setting.")
```
radoering, line 142: Can we move this up, so we don't have to do the regex matching, which should not be relevant in this case?

tests/console/commands/test_config.py
```diff
@@ -402,6 +402,16 @@ def test_set_pypi_token_unsuccessful_multiple_values(
     assert str(e.value) == "Expected only one argument (token), got 2"
 
 
+def test_set_pypi_token_no_values(
+    tester: CommandTester,
+) -> None:
+    print("Running this test")
```
radoering, line 408:
```suggestion
```

radoshi: Doh. Silly. Fixed.

src/poetry/console/commands/config.py
radoshi: Yeah good point. It'll make the code cleaner. Fixed.

radoering (approved):

Issue #7611, opened 2023-03-06
Installing a package that depends on setuptools fails intermittently: `Backend 'setuptools.build_meta:__legacy__' is not available.`


MycroftKang:
- **Poetry version**: 1.4.0
- **Python version**: 3.10.9
- **OS version and name**: Microsoft Windows Server 2022 10.0.20348 Datacenter (`windows-2022` GitHub runner)

- [x] I am on the [latest](https://github.com/python-poetry/poetry/releases/latest) stable Poetry version, installed using a recommended method.
- [x] I have searched the [issues](https://github.com/python-poetry/poetry/issues) of this repo and believe that this is not a duplicate.
- [ ] I have consulted the [FAQ](https://python-poetry.org/docs/faq/) and [blog](https://python-poetry.org/blog/) for any relevant entries or release notes.
- [ ] If an exception occurs when executing a command, I executed it again in debug mode (`-vvv` option) and have included the output below.

## Issue

Installing a package that depends on setuptools fails intermittently. This bug does not occur in poetry 1.3 and is found in poetry 1.4.

```
ChefBuildError

  Backend 'setuptools.build_meta:__legacy__' is not available.

  at C:\Program Files (x86)\pipx\venvs\poetry\lib\site-packages\poetry\installation\chef.py:152 in _prepare
      148| 
      149|                 error = ChefBuildError("\n\n".join(message_parts))
      150| 
      151|             if error is not None:
    > 152|                 raise error from None
      153| 
      154|             return path
      155| 
      156|     def _prepare_sdist(self, archive: Path, destination: Path | None = None) -> Path:

Note: This error originates from the build backend, and is likely not a problem with poetry but with docopt (0.6.2) not supporting PEP 517 builds. You can verify this by running 'pip wheel --use-pep517 "docopt (==0.6.2)"'.
```

<details>
<summary>View full logs</summary>
<pre><code>
    $ poetry install
    Installing dependencies from lock file

    Package operations: 150 installs, 2 updates, 0 removals

        Installing certifi (2022.12.7)
        Installing charset-normalizer (2.0.12)
        Installing idna (3.4)
        Installing pyasn1 (0.4.8)
        Installing urllib3 (1.26.14)
        Installing cachetools (5.3.0)
        Installing oauthlib (3.2.2)
        Installing pyasn1-modules (0.2.8)
        Installing requests (2.28.2)
        Installing six (1.16.0)
        Installing rsa (4.9)
        Installing catalogue (2.0.8)
[...]

Issue #8483, opened 2023-09-28
1.7.0 Release


radoering:
This issue tracks the progress of version 1.7.0 of Poetry. Please surface any blockers or concerns you have regarding this minor release here:

## Blockers
* #8399 (just waits for the poetry-core release)
* #8479
* #8517 
* python-poetry/poetry-core#656
* Changelog #8599 
* Release notes (blog) python-poetry/website#138

## Nice to have
* #8227
* python-poetry/poetry-plugin-export#243
* python-poetry/cleo#380

## Important changes to surface

(This is intended to be the basis for the blog post and should cover new features or any manual steps/incompatibility when upgrading)

* official support for Python 3.12, mention behavior change for implicit setuptools and wheel, see https://github.com/python-poetry/poetry/pull/7803#issuecomment-1704318275
* improved handling of dependencies with extras during dependency resolution #8399
* #8517
* #8562

MadhavVij:
Do we have a timeline for when's the release planned for?

radoering:
As always, when it's ready. (Maybe next week.)

For those waiting for Python 3.12 support, afaik it's not that Poetry 1.6 won't work with Python 3.12, it's just that it wasn't tested and there will be some minor adjustments in the next release:

- As can be seen in #7803, tests just passed without any adjustments except for a minor detail regarding the implicit `setuptools` and `wheel` included in the virtual environments. Anyways, relying on these implicit packages is bad practice. If you depend on setuptools or wheel, you should make it an explicit dependency.
- The second notable difference is that the current Poetry will not automatically add the `Programming Language :: Python :: 3.12` classifier when building a distribution. However, you can just add it manually to `tool.poetry.classifiers` in `pyproject.toml` if you can't wait to build a distribution with that classifier.

Apart from that, I'm not aware of any differences regarding Python 3.12.

jakob-keller:
It would be nice to get #8559 into the 1.7.0 release. Anything I can do to facilitate that?

PR #8517, opened 2023-10-07
Fix race condition in artifact cache

radoering:
# Pull Request Check List

Resolves: #7611

Fixes the race condition identified in https://github.com/python-poetry/poetry/issues/7611#issuecomment-1747836233

The first two commits are refactorings to reduce duplicated code and let the artifact cache handle downloads of archives that will be stored in the artifact cache. The third commit is the actual fix, which uses per-artifact locks.

### DIFF
--- a/src/poetry/installation/executor.py
+++ b/src/poetry/installation/executor.py
@@ -15,7 +15,6 @@ from typing import Any
 
 from cleo.io.null_io import NullIO
 from poetry.core.packages.utils.link import Link
-from requests.utils import atomic_open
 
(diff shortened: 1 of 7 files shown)


thesct22:
Patiently working for this to be available
😁
Do you guys have an estimated update on when (if) this will be patched through as a new release?

radoering:
It's planned for 1.7 (see #8483). I'll probably start with the preparations for the release after this one has been reviewed and merged.

neersighted (approved):

PR #8523, opened 2023-10-09
fix: error when lockfile without metadata entry is encountered

iloveitaly:
Without this, you'll get a cryptic 'metadata' in red when running `poetry install`. This can occur if a poetry.lock file is empty

Not sure of the best way to write a test for this, point me in the right direction and I'll write a test.

# Pull Request Check List

### DIFF
--- a/tests/packages/test_locker.py
+++ b/tests/packages/test_locker.py
@@ -647,11 +647,11 @@ reference = "legacy"
     with pytest.raises(RuntimeError) as e:
         _ = locker.lock_data
 
-        assert (
-            "The lock file does not have a metadata entry.\nRegenerate the lock file with"
-            " the `poetry lock` command."
-            in str(e.value)
-        )
+    assert (
+        "The lock file does not have a metadata entry.\nRegenerate the lock file with"
+        " the `poetry lock` command."
+        in str(e.value)
+    )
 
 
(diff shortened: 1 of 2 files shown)


iloveitaly:
added test + fixed linting

tests/packages/test_locker.py
```diff
@@ -622,6 +622,38 @@ def test_reading_lock_file_should_raise_an_error_on_invalid_data(
...
+url = "https://foo.bar"
+reference = "legacy"
+"""
+    with locker.lock.open("w", encoding="utf-8") as f:
+        f.write(content)
+
+    with pytest.raises(RuntimeError) as e:
+        _ = locker.lock_data
+
+    assert (
```
eamanu, line 650: I'd run the assert in the `with pytest.raises` block.

eamanu (approved):

tests/packages/test_locker.py
iloveitaly: good catch, fixed!

radoering: IMO, only the code that is expected to raise the exception should be in the `with pytest.raises`. That's also how we did it in all other tests, so please revert.

iloveitaly: reverted!

radoering (approved):

PR #8600, opened 2023-10-31
remove special cases for setuptools and wheel

dimbleby:
`poetry install --sync` will now remove those packages, if they are not required by the project being managed.

consistent with the modern world of isolated builds, #7803, etc: continuing to remove special cases for `setuptools` and `wheel`.

### DIFF
--- a/src/poetry/puzzle/transaction.py
+++ b/src/poetry/puzzle/transaction.py
@@ -98,14 +98,9 @@ class Transaction:
                 result_package_names = {
                     result_package.name for result_package, _ in self._result_packages
                 }
-                # We preserve pip/setuptools/wheel when not managed by poetry, this is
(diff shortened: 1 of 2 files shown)


radoering (approved):

johnthagen:
Note, this is likely an "issue" with PyCharm, but for your awareness 1.7.0 is currently pretty disruptive for PyCharm users because it assumes that `setuptools` will be in the environment on Python <3.12.

- https://youtrack.jetbrains.com/issue/PY-48909/PyCharm-assumes-the-presence-of-setuptools-and-pip#focus=Comments-27-8357081.0-0

dimbleby:
you can always add setuptools as a dev dependency, which apparently for you it is

Issue #2048, opened 2020-02-15
Tab completion with Prezto (Zsh)


jayBana:
- [x] I have searched the [issues](https://github.com/python-poetry/poetry/issues) of this repo and believe that this is not a duplicate.
- [x] I have searched the [documentation](https://python-poetry.org/docs/) and believe that my question is not covered.

## Issue
Did anyone manage to get tab completion working with [Prezto](https://github.com/sorin-ionescu/prezto) for ZSH? If so, please advise.

Is there a Prezto plugin for it?

Palisand:
```shell
poetry completions zsh > ~/.zprezto/modules/completion/external/src/_poetry
```
worked for me.

PR #8607, opened 2023-11-01
Adding additional information for completions in Prezto

predmijat:
Resolves: [#2048](https://github.com/python-poetry/poetry/issues/2048)

I had to perform this additional step in order to get completions working with Prezto (capital 'P' according to the official repository).

### DIFF
--- a/docs/_index.md
+++ b/docs/_index.md
@@ -398,7 +398,7 @@ plugins(
 	)
 ```
 
-#### prezto
+#### Prezto
 
 ```zsh
 poetry completions zsh > ~/.zprezto/modules/completion/external/src/_poetry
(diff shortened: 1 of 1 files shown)


docs/_index.md
```diff
@@ -403,12 +403,13 @@ plugins(
...
 
-#### prezto
+#### Prezto
 
 ```zsh
 poetry completions zsh > ~/.zprezto/modules/completion/external/src/_poetry
 ```
 
 {{% note %}}
 You may need to restart your shell in order for these changes to take effect.
```
radoering, line 413: I assume this note was meant for all types of shells. Maybe, we should put it at the top (before the sections for indiviual shells).

Secrus:
I am not sure we should provide that info. Why for one type of shell framework only?

docs/_index.md
```diff
@@ -398,12 +398,13 @@ plugins(
...
+If completions still don't work, try removing `~/.cache/prezto/zcompcache` and starting a new shell.
```
abn, line 409:
```suggestion
If completions still do not work, you can try removing the `~/.cache/prezto/zcompcache` file and restarting your shell.
```

abn:
Created at #10070. Thank you @predmijat , I have retained your commit authorship.

PR #8621, opened 2023-11-04
installer: remove unrequested extras if `requires_synchronization` is set

radoering:
**tl;dr**

- fix an issue where `poetry install --sync` does not remove unrequested `extras` (funny enough `poetry install` without `--sync` does remove them)
- increase test coverage of the `installer` regarding `extras`

**Long story**

Originally, I was just poking around with https://github.com/python-poetry/poetry-core/pull/613#issuecomment-1694697769. However, I noticed that all `installer` tests still passed if one of the following lines were removed:

- https://github.com/python-poetry/poetry/blob/f5b022843660988daaf4153cf3058a33b389b4f6/src/poetry/installation/installer.py#L386-L388
- https://github.com/python-poetry/poetry/blob/f5b022843660988daaf4153cf3058a33b389b4f6/src/poetry/installation/installer.py#L394-L395

Thus, I decided to improve the test coverage before trying to fix bugs. I merged some quite similar tests and used `pytest.mark.parametrize` to cover the missing cases. However, https://github.com/python-poetry/poetry/blob/f5b022843660988daaf4153cf3058a33b389b4f6/src/poetry/installation/installer.py#L386-L388 was still irrelevant.

After some code spelunking, I suppose that these lines became irrelevant with [ea8fb8c](https://github.com/python-poetry/poetry/commit/ea8fb8c6679f1b9afee4ef141467acc615eca71f#diff-95c2e34175a95676a55830f8442a68da156ae6a31dfec68317a2fd746be2243f). However, that introduced the mentioned bug. After adding `@pytest.mark.parametrize("do_sync", [False, True])` to the test and fixing the bug, the mentioned lines of code are relevant again.

### DIFF
--- a/src/poetry/installation/installer.py
+++ b/src/poetry/installation/installer.py
@@ -267,7 +267,7 @@ class Installer:
             ops = self._get_operations_from_lock(locked_repository)
 
         lockfile_repo = LockfileRepository()
-        self._populate_lockfile_repo(lockfile_repo, ops)
+        uninstalls = self._populate_lockfile_repo(lockfile_repo, ops)
 
         if not self.executor.enabled:
             # If we are only in lock mode, no need to go any further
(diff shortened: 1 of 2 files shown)


Secrus (approved):
LGTM

Issue #1537, opened 2019-11-04
Add support for disabling package publish


iamsauravsharma:
- [X] I have searched the [issues](https://github.com/sdispater/poetry/issues) of this repo and believe that this is not a duplicate.
- [X] I have searched the [documentation](https://poetry.eustace.io/docs/) and believe that my question is not covered.

## Feature Request

Currently poetry doesn't support disabling publishing of package to remote repository. Add `publish` key similar to `publish` key present over [cargo manifest](https://doc.rust-lang.org/cargo/reference/manifest.html#the-publish--field-optional) for rust crates. `publish` key can have two types of value either bool or list of repository name. The `publish` field can be used to prevent a package from being published to a remote repository (like PYPI) by mistake, for instance to keep a package private in a company. 
```toml
[tool.poetry]
#....
publish=false
```
```toml
[tool.poetry]
#....
publish=["some-remote-repository-name"]
```

515hikaru:
I think this proposal is nice.

I set the following for packages that should not be made public.

```toml
[tool.poetry]
exclude = ["**/*"]
```

With this workaround, even if `poetry publish` is executed by mistake, the source code is not published. However, since `poetry build` includes pyproject.toml, setup.py, and PKG-INFO in the tar.gz file, such metadata may be disclosed.

This workaround is not good. I think you need an option to make it impossible to publish.

alkasm:
This is very trivial to implement. However the more important question here is the API / user experience. I'd be happy to push out a PR for this with some agreement from core devs on API. I have written down a host of options, there may be others as well. 

---

### Option 1: Borrow Rust Cargo's manifest `publish` 

This is the option specified in the OP, using the same [`publish` field](https://doc.rust-lang.org/cargo/reference/manifest.html#the-publish-field) as Rust's manifest which can be either `false` or an array of repo names, e.g. `["repo-name-1", "repo-name-2"]`. This follows an established standard from another packaging tool which uses TOML, and is easy to understand and document, but has a number of disadvantages. It's annoying that the `publish` option may have multiple types, but more importantly, it forces repetition of allowable repo names (leading to possible errors) and implicitly overrides default behavior. An empty list `publish=[]` would likely do the same thing as `false` leading to documentation confusion and multiple ways to do the same thing. The field name `publish` is intuitive for the boolean case, but unintuitive for the array.

Other similar options might be to provide an `allowed-repos = []` array or `disallowed-repos = []` array.

### Option 2: Override the PyPI URL

[...]

PR #8650, opened 2023-11-11
Introduce non-package-mode

radoering:
# Pull Request Check List

Resolves: #1132

Requires: python-poetry/poetry-core#661

Introduce non-package mode

- add a new attribute ~`mode`~ `package-mode` in `tool.poetry`
- ~`mode` can be either `package` (default) or `non-package`~
- `package-mode` can be either `true` (default) or `false`
- in non-package mode:
  - metadata like `name` and `version` is not required
  - the root package is never installed (same as `--no-root`)
  - building and publishing is not possible

[Docs preview](https://website-jeevwq2md-python-poetry.vercel.app/docs/basic-usage/#operating-modes)

Example Config:

```
[tool.poetry]
package-mode = false

[tool.poetry.dependencies]
python = "^3.8"
cleo = ">=2.0.0"
```

You can install this branch with pipx as follows:
```
pipx install --suffix _npm git+https://github.com/radoering/poetry.git@non-package-mode
```

This will install poetry with the suffix "_npm" so that you have to call `poetry_npm` instead of `poetry`. (The suffix is arbitrary, I chose "npm" as an abbreviation for non-package-mode. 😉)

### DIFF
--- a/docs/basic-usage.md
+++ b/docs/basic-usage.md
@@ -76,6 +76,29 @@ cd pre-existing-project
 poetry init
 ```
 
+### Operating modes
+
+Poetry can be operated in two different modes. The default mode is the **package mode**, which is the right mode
+if you want to package your project into an sdist or a wheel and perhaps publish it to a package index.
+In this mode, some metadata such as `name` and `version`, which are required for packaging, are mandatory.
+Further, the project itself will be installed in editable mode when running `poetry install`.
+
+If you want to use Poetry only for dependency management but not for packaging, you can use the **non-package mode**:
+
+```toml
+[tool.poetry]
+mode = "non-package"
+```
+
+In this mode, metadata such as `name` and `version` are optional.
+Therefore, it is not possible to build a distribution or publish the project to a package index.
+Further, when running `poetry install`, Poetry does not try to install the project itself,
+but only its dependencies (same as `poetry install --no-root`).
+
+{{% note %}}
+In the [pyproject section]({{< relref "pyproject" >}}) you can see which fields are required in package mode.
+{{% /note %}}
+
 ### Specifying dependencies
 
 If you want to add dependencies to your project, you can specify them in the `tool.poetry.dependencies` section.
--- a/docs/basic-usage.md
+++ b/docs/basic-usage.md
@@ -87,7 +87,7 @@ If you want to use Poetry only for dependency management but not for packaging, 
 
 ```toml
 [tool.poetry]
-mode = "non-package"
+package-mode = false
 ```
 
 In this mode, metadata such as `name` and `version` are optional.
--- a/docs/pyproject.md
+++ b/docs/pyproject.md
@@ -13,9 +13,19 @@ menu:
 
 The `tool.poetry` section of the `pyproject.toml` file is composed of multiple sections.
 
+## mode
+
+The mode of the project, either `"package"` (default) or `"non-package"`. **Optional**
+
+See [basic usage]({{< relref "basic-usage#operating-modes" >}}) for more information.
+
+```toml
+mode = "non-package"
+```
+
 ## name
 
-The name of the package. **Required**
+The name of the package. **Required in package mode**
 
 This should be a valid name as defined by [PEP 508](https://peps.python.org/pep-0508/#names).
 
@@ -26,7 +36,7 @@ name = "my-package"
 
 ## version
 
-The version of the package. **Required**
+The version of the package. **Required in package mode**
 
 This should be a valid [PEP 440](https://peps.python.org/pep-0440/) string.
 
@@ -43,7 +53,7 @@ If you would like to use semantic versioning for your project, please see
 
 ## description
 
-A short description of the package. **Required**
+A short description of the package. **Required in package mode**
 
 ```toml
 description = "A short description of the package."
@@ -81,7 +91,7 @@ If your project is proprietary and does not use a specific licence, you can set 
 
 ## authors
 
-The authors of the package. **Required**
+The authors of the package. **Required in package mode**
 
 This is a list of authors and should contain at least one author. Authors must be in the form `name <email>`.
 
--- a/docs/pyproject.md
+++ b/docs/pyproject.md
@@ -13,14 +13,14 @@ menu:
 
 The `tool.poetry` section of the `pyproject.toml` file is composed of multiple sections.
 
-## mode
+## package-mode
 
-The mode of the project, either `"package"` (default) or `"non-package"`. **Optional**
+Whether Poetry operates in package mode (default) or not. **Optional**
 
 See [basic usage]({{< relref "basic-usage#operating-modes" >}}) for more information.
 
 ```toml
-mode = "non-package"
+package-mode = false
 ```
 
 ## name
--- a/src/poetry/console/commands/build.py
+++ b/src/poetry/console/commands/build.py
@@ -49,6 +49,10 @@ class BuildCommand(EnvCommand):
             builder(self.poetry, executable=executable).build(target_dir)
 
     def handle(self) -> int:
+        if not self.poetry.is_package_mode:
+            self.line_error("Building a package is not possible in non-package mode.")
+            return 1
+
         with build_environment(poetry=self.poetry, env=self.env, io=self.io) as env:
             fmt = self.option("format") or "all"
             dist_dir = Path(self.option("output"))
--- a/src/poetry/console/commands/install.py
+++ b/src/poetry/console/commands/install.py
@@ -187,7 +187,7 @@ you can set the "package-mode" to false in your pyproject.toml file.
             # No need for an editable install in this case.
             self.line("")
             self.line_error(
-                f"The current project could not be installed: <error>{e}</error>\n"
+                f"Warning: The current project could not be installed: {e}\n"
                 "If you do not want to install the current project"
                 " use <c1>--no-root</c1>.\n"
                 "If you want to use Poetry only for dependency management"
--- a/src/poetry/console/commands/install.py
+++ b/src/poetry/console/commands/install.py
@@ -79,7 +79,7 @@ dependencies and not including the current project, run the command with the
 <info> poetry install --no-root</info>
 
 If you want to use Poetry only for dependency management but not for packaging,
-you can set the operating mode to "non-package" in your pyproject.toml file.
+you can set the "package-mode" to false in your pyproject.toml file.
 """
 
     _loggers = ["poetry.repositories.pypi_repository", "poetry.inspection.info"]
--- a/src/poetry/console/commands/install.py
+++ b/src/poetry/console/commands/install.py
@@ -77,6 +77,9 @@ dependencies and not including the current project, run the command with the
 <info>--no-root</info> option like below:
 
 <info> poetry install --no-root</info>
+
+If you want to use Poetry only for dependency management but not for packaging,
+you can set the operating mode to "non-package" in your pyproject.toml file.
 """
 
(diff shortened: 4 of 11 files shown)


ilyagr:
For people who want to try this out, here is what the config looks like at the moment:

```toml
[tool.poetry]
mode = "non-package"

[tool.poetry.dependencies]
python = "^3.8"
cleo = "^0.6"
pendulum = { git = "https://github.com/sdispater/pendulum.git", branch = "2.0" }
```
@radoering , it might make sense to copy this to PR description and collapse my comment. Thank you for working on this!

ab:
Thanks for working on this! I tested it locally and everything worked as expected.

I know you left a TODO for docs, but it might be nice to update the help text in places like:
https://github.com/python-poetry/poetry/blob/5e0ff80925dacef4b3d9ff6e7b6505c60b4ee152/src/poetry/console/commands/install.py#L75-L79

docs/pyproject.md
```diff
@@ -13,9 +13,17 @@ menu:
 
 The `tool.poetry` section of the `pyproject.toml` file is composed of multiple sections.
 
+## mode
+
+The mode of the project, either `"package"` (default) or `"non-package"`. **Optional**
```
ilyagr, line 18: It might be nice to link to the section in `basic-usage.md` from here, or to some other explanation of what this means.

Alternatively, you could shorten the description of non-package mode there and move it here.

Secrus:
I wonder if we actually need to introduce that "mode" part. We could just have placeholder values for when the config is missing `name` and `version` and refuse to do packaging-related stuff (like artifact building) and project installation when those are not set properly. What do you think?

mathewvaughan:
> I wonder if we actually need to introduce that "mode" part. We could just have placeholder values for when the config is missing `name` and `version` and refuse to do packaging-related stuff (like artifact building) and project installation when those are not set properly. What do you think?

What you're saying might make sense for when those things aren't set, but it's not very explicit and relies on the user knowing implicitly that 'omitting name and version will activate non-package mode'. 

Personally, I do use 'name' and 'version' even when I'm not building a package - and I feel like an explicit boolean setting of 'package=false' would be clearer.

Secrus:
> Personally, I do use 'name' and 'version' even when I'm not building a package - and I feel like an explicit boolean setting of 'package=false' would be clearer.

Ok, so what use would you have for that mode if you are setting the name and version anyway?

mathewvaughan:
> > Personally, I do use 'name' and 'version' even when I'm not building a package - and I feel like an explicit boolean setting of 'package=false' would be clearer.
> 
> Ok, so what use would you have for that mode if you are setting the name and version anyway?

It's the name and version of the thing I'm building, but it's not a package, it's a web service

Secrus:
> It's the name and version of the thing I'm building, but it's not a package, it's a web service

Ok, but as you can see in PR description, other than removing the requirement for name and version, what that mode introduces is lock on building and publishing (which I would say is on the user to decide, no real need to hardlock it) and installation of the project is using `--no-root` by default (which I would say isn't that much effort to type in and we could have Poetry detect that name and version are placeholder values and default to `--no-root` anyway, with a proper message). So out of those three changes, I would say only the `name`/`version` is meaningful and for your case, it's already handled since you set it anyway.

mathewvaughan:
> > It's the name and version of the thing I'm building, but it's not a package, it's a web service
> 
> Ok, but as you can see in PR description, other than removing the requirement for name and version, what that mode introduces is lock on building and publishing (which I would say is on the user to decide, no real need to hardlock it) and installation of the project is using `--no-root` by default (which I would say isn't that much effort to type in and we could have Poetry detect that name and version are placeholder values and default to `--no-root` anyway, with a proper message). So out of those three changes, I would say only the `name`/`version` is meaningful and for your case, it's already handled since you set it anyway.

I'm here because poetry kept asking me to specify 'no-root'. It's a nuisance to have to specify that every time I run install and also has no obvious semantic link to the concept of 'I'm actually not building a package, thanks'. I was merely providing an answer to your question 'Do we really need the mode part?' and I'm explaining why I think we do. 

I can appreciate that blocking the publishing/building if there is no name/version would be sufficient to stop web services being published as packages, but as someone who uses poetry to build web services it would be nice for there to be an obvious way to configure poetry to not expect me to jump through hoops designed for package development.

I think `mode` isn't clear enough still, though. It feels too generic a term. I think it would be better to use `package=true/false` and default it to true during `poetry init`, providing an option to set it to false.

radoering:
I tend to say, it's good to have an explicit way because naming and versioning your project even if you don't intend to build is a legit use case (especially if we support PEP 621 in the future so that name and version are not poetry specific keys). So yes, at the moment it's only about `--no-root` or not, which can't be decided implicitly, but I think always having to add `--no-root` is annoyance enough. Further, I'm not sure whether other differences (which nobody is currently thinking about) might be added in the future.

However, we can think about making it a tri-state with the default (not explicitly set): package mode if name and version is defined, otherwise non-package mode. I'm just not sure if the little bit of comfort is worth making it more complicated (not only to implement but also to understand as a normal user)?

> I think mode isn't clear enough still, though. It feels too generic a term. I think it would be better to use package=true/false and default it to true during poetry init, providing an option to set it to false.

I'm not sure if `package` is a good idea because we already have [packages](https://python-poetry.org/docs/pyproject/#packages). Nevertheless, I'm open to suggestions for other names.

The option in `poetry init` makes sense (if we have an explicit setting) and might be added in this PR or later.

mathewvaughan:
It's a fair concern to worry about collisions on the names, but whilst looking at how the `packages` option is used I've started to believe the similarity is actually quite natural. Thinking about how we want want these options to interact:

- `package=true` and no `packages`? Then we look for a directory that has the correct package name.
- `package=true` with packages? We simply respect the dirnames within `packages`
- `package=false` with no `packages`. You're probably building a web service so we wont fuss about `--no-root`
- `package=false` but you've configured `packages`? Go home, you're drunk. (And we maybe should warn/error here?)

I mean, even if we just went with `package_mode=true/false`? I'm just wondering if we otherwise open ourselves to having `mode` become a bloated config option.

> The option in `poetry init` makes sense (if we have an explicit setting) and might be added in this PR or later.

I really would be curious what the lowest friction default choice would be for first-time setup? Either way, I've never contributed to an open source project before but `--no-root` has been bothering me at work so with your blessing I'd like to contribute here - Perhaps I could raise a separate PR with the `poetry init` changes, ready to be merged after this branch?

radoering:
I'm quite convinced now that this should be an explicit option. This leaves naming as an open topic. Let's try to get a picture of the general mood. (I have no idea how well this will work, but it's worth a try. 😅)

Please react with the respective emoji if you have an opinion:

- 🚀 `mode` with values `"package"` (default) and `"non-package"`
- 🎉 `package` with values `true` (default) and `false` (consider that there is already [packages](https://python-poetry.org/docs/pyproject/#packages), which however will only be used in package mode)
- ❤️ `package-mode` with values `true` (default) and `false`
- 👀 if you are waiting for this feature but do not care about the name

If you have two favorites and dislike the third option, just vote for your two favorites.

Disclaimer: This vote does not result in a binding decision. We may choose another name if we think there is a good reason.

branchvincent:
> consider that there is already [packages](https://python-poetry.org/docs/pyproject/#packages), which however will only be used in package mode

I'll offer one more option 😅 : `packages = false`

It re-uses the existing field, avoids the possibility of conflicting settings like `package = False` and `packages = [...]`, and also mirrors cargo's [`publish`](https://doc.rust-lang.org/cargo/reference/manifest.html#the-publish-field) field (which can be `false` or an array of packages to publish)

radoering:
> I'll offer one more option 😅 : `packages = false`
> 
> It re-uses the existing field, avoids the possibility of conflicting settings like `package = False` and `packages = [...]`, and also mirrors cargo's [`publish`](https://doc.rust-lang.org/cargo/reference/manifest.html#the-publish-field) field (which can be `false` or an array of packages to publish)

One thing to consider: `packages = []` might make sense to define a meta package without modules/packages but only dependencies. (Afaik that does not work at the moment and I'm not sure if that's a valid use case but I'd assume it is until the opposite is proven.) So `packages = []` and `packages = false` would have a different meaning which might be difficult to understand?

silverwind:
I too would prefer to just re-use the existing `packages` field where any empty/falsy value is taken as "there are no packages", e.g. `[]`, `False`, `None`. Imho, this non-package mode should also be the default when `packages` is not specified.

johnthagen:
> Imho, this non-package mode should also be the default when packages is not specified.

I think this would break simple projects that only have a single package that is implicitly discovered. For example, the following repo builds a package, but it is implicitly found in `src/fact` based on convention and the `fact` name.

- https://github.com/johnthagen/python-blueprint/blob/main/pyproject.toml

radoering:
FYI: I switched the PR from `mode` to `package-mode` as this seems to be more popular. Further, with no prospect of introducing further modes a Boolean option is sufficient.

johnthagen:
> building and publishing is not possible

A nice side effect of this change will be that for non-packages, we will also get a clean solution for private-ness as well 🚀 

- #1537

nextmat:
Thanks for all the work on this all, I'm really excited to see this get merged. 😁 

It looks like this is scheduled to be part of the 1.8 release. Reviewing the [1.8 issue](https://github.com/python-poetry/poetry/issues/8770) it seems like the related PRs are mostly ready to go, I was curious about likely timeline to the release or if there is anything we can do to help?

radoering:
> it seems like the related PRs are mostly ready to go

That's right. They are mainly waiting for the poetry-core release. I'm just waiting for some reviews before creating the release PR. After the poetry-core release we can do the final adjustments in the open PRs mentioned in #8770.

docs/basic-usage.md
```diff
@@ -76,6 +76,29 @@ cd pre-existing-project
...
+if you want to package your project into an sdist or a wheel and perhaps publish it to a package index.
+In this mode, some metadata such as `name` and `version`, which are required for packaging, are mandatory.
+Further, the project itself will be installed in editable mode when running `poetry install`.
+
+If you want to use Poetry only for dependency management but not for packaging, you can use the **non-package mode**:
+
+```toml
+[tool.poetry]
+package-mode = false
+```
```
silverwind, line 91: Will the setting work in `poetry.toml` as well? If so, I would add an example for  that as well.

radoering: I don't think so. Actually, I don't think it would be a good idea to put it in `poetry.toml` because it's clearly a project setting and not a user setting. (It's not like user A wants to have this setting for project X while user B does not want this setting for project X.)

silverwind: I meant `poetry.toml` inside a project directory. I did not test it yet, but I assume settings in `poetry.toml` are loaded the same as ones in `pyproject.toml`'s `tool.poetry` section.

Secrus (approved):
LGTM

johnthagen:
@armanjal It's generally best to open a new issue to report an issue like this. You can still reference this MR using `#8650`.

armanjal:
> @armanjal It's generally best to open a new issue to report an issue like this. You can still reference this MR using `#8650`.

Sure , thanks!

thinh9e:
Can we add a new parameter in `poetry new` and `poetry init` commands so that we can set it to `false` when creating a new project?

neersighted:
Please open a feature request issue -- merged PRs are not a good venue for discussion, as both initial comments and any follow-up are easy to lose track of in the noise. Creating a new issue, with a descriptive title and body both makes clear what the current (and ideally only) topic of discussion is, and makes returning to the conversation in the future easier.

Issue #8631, opened 2023-11-07
Poetry fails to use the currently activated environnement


laclouis5:
- **Poetry version**: 1.7.0
- **Python version**: 3.11
- **OS version and name**: Ubuntu 22.04.3 LTS
- **pyproject.toml**: Empty TOML as created by `poetry new <project>` 

- [x] I am on the [latest](https://github.com/python-poetry/poetry/releases/latest) stable Poetry version, installed using a recommended method.
- [x] I have searched the [issues](https://github.com/python-poetry/poetry/issues) of this repo and believe that this is not a duplicate.
- [x] I have consulted the [FAQ](https://python-poetry.org/docs/faq/) and [blog](https://python-poetry.org/blog/) for any relevant entries or release notes.
- [x] If an exception occurs when executing a command, I executed it again in debug mode (`-vvv` option) and have included the output below.

## Issue
In some circonstances Poetry fails to use the activated virtual environnement (for instance a Conda environnement). In my experiments, this seems to occur when there is an hyphen (`-`) in the name of the folder of the project (for instance `poetry-project`).

### To Reproduce

```shell
poetry new poetry-project
cd poetry-project

conda create -n poetry-project python=3.11 -y
conda activate poetry-project

poetry env info
```

This returns:

```
Virtualenv
Python:         3.10.11
Implementation: CPython
Path:           NA
Executable:     NA
```

This works correctly if I change the folder name to use an underscore(`_`) instead of the hyphen (`-`), for instance `poetry_project`.

dimbleby:
poetry just looks at environment variables to find the current virtual environment, I don't see anything special about hyphens

https://github.com/python-poetry/poetry/blob/f004a70260404560b89334271641a7093ada6551/src/poetry/utils/env/env_manager.py#L252-L256

so you should start there to understand what's actually happening

dimbleby:
yeah, does not reproduce.

I'm starting with this docker file:
```dockerfile
# syntax=docker/dockerfile:1.5
FROM continuumio/miniconda3:23.9.0-0

RUN apt-get update && \
    apt-get install \
      --yes \
      --no-install-recommends \
      curl && \
    curl \
      --silent \
      --show-error \
      --location \
      --output install-poetry.py \
      https://install.python-poetry.org && \
    python3 install-poetry.py

ENV PATH="/root/.local/bin:$PATH"
```

Then the sequence of commands that you give ends with:
```console
[...]

PR #8665, opened 2023-11-13
Fix for the removal of all the environments using the `--all` option

laclouis5:
Resolves: #8631 

I didn't added new tests because one is already here for this command (there is a test for `poetry env remove --all`) and don't think new documentation is required since this is a simple bug fix. I ran all the tests and linting/formatting tools listed in the [contribution](https://python-poetry.org/docs/contributing#code-contributions) guidelines before submitting this PR.

The previous code did't remove the TOML sections in `envs.toml` when removing all virtual environment using `poetry env remove --all`. I implemented a new method on the `EnvManager` for this specific case. I also renamed the previous method `EnvManager.remove_venv` to the more explicit name `EnvManager.remove_venv_dir` and updated the references.

I struggled a little to understand if the environments listed in `envs.toml` should be kept in sync with the virtual environments present in the parent folder. I supposed the negative since I think that this assertion is never checked in the tests, thus I haven't modified the tests to check this. 

Feel free to suggest improvements.


dimbleby:
I never use `poetry env` and I'm not too sure what `envs.toml` is used for in all honesty.  Could you simply have deleted that file from the "remove all" branch?

well this looks plausible too, hopefully someone who remembers something about this code will take a look

laclouis5:
Looks like it is used pervasively through the code dealing with envs to me. I don't think it should be deleted.

laclouis5:
Ah, there is a misunderstanding I think! The command `poetry env remove --all` does not remove all the existing Poetry environments but only all the environments associated with the *current* project.

src/poetry/utils/env/env_manager.py
```diff
@@ -405,10 +405,34 @@ def remove(self, python: str) -> Env:
...
 
-        self.remove_venv(venv_path)
+        self.remove_venv_dir(venv_path)
 
         return VirtualEnv(venv_path, venv_path)
 
+    def remove_venv(self, venv: VirtualEnv) -> Env:
+        venv_path = self._poetry.config.virtualenvs_path
+        cwd = self._poetry.file.path.parent
+        envs_file = TOMLFile(venv_path / self.ENVS_FILE)
```
radoering, line 415:
```suggestion
        envs_file = TOMLFile(venv.path.parent / self.ENVS_FILE)
```

Shouldn't that be enough to get the location?

src/poetry/utils/env/env_manager.py
```diff
@@ -405,10 +405,34 @@ def remove(self, python: str) -> Env:
...
+        base_env_name = self.generate_env_name(self._poetry.package.name, str(cwd))
+
+        env_path = venv.path
+        _, minor, _ = venv.get_version_info()
+
+        name = f"{base_env_name}-py{minor}"
```
radoering, line 421: That's wrong. It results in `py9` for Python 3.9.

src/poetry/utils/env/env_manager.py
```diff
@@ -405,10 +405,34 @@ def remove(self, python: str) -> Env:
...
+
+        if not env_path.exists():
+            raise ValueError(f'<warning>Environment "{name}" does not exist.</warning>')
```
radoering, line 424: Do we need this warning? At least raising a `ValueError` seems to be wrong, especially for a warning.

src/poetry/utils/env/env_manager.py
```diff
@@ -405,10 +405,34 @@ def remove(self, python: str) -> Env:
...
+
+        if envs_file.exists():
+            envs = envs_file.read()
+            if base_env_name in envs:
+                del envs[base_env_name]
+                envs_file.write(envs)
```
radoering, line 430: It may not make a difference for `remove --all` but considering that this method removes exactly one venv, we should only remove the entry if the Python version matches the version of the venv we are removing.

src/poetry/utils/env/env_manager.py
```diff
@@ -405,10 +405,34 @@ def remove(self, python: str) -> Env:
...
+
+        self.remove_venv_dir(env_path)
+
+        return VirtualEnv(env_path, env_path)
```
radoering, line 434: Why do we return a venv?

radoering (changes requested):
> I didn't added new tests because one is already here for this command (there is a test for poetry env remove --all)

I don't follow this argument. If I understand correctly, this PR fixes a bug. In that case, there should be a test that fails without the fix. If there isn't such a test, then it should be added. (It shouldn't be too difficult to create an `envs.toml` in a test and check that the concerning section is removed.)

src/poetry/utils/env/env_manager.py
laclouis5: This now returns `py3.9` like in the `.remove()` method.

src/poetry/utils/env/env_manager.py
laclouis5: `ValueError` is the error type that was chosen for this kind of warning I guess. I sticked to the behavior of the original `.remove()` method which uses `ValueError` in this exact same way.

src/poetry/utils/env/env_manager.py
laclouis5: Again, I sticked to what appears to be the norm in the Poetry repository. The original `.remove()` method returns a virtual env, thus I kept that design parttern here too.

src/poetry/utils/env/env_manager.py
laclouis5: Indeed, this may no make a difference. The check is only there if there is an inconsistency between the TOML env file and the virtual env folders.

laclouis5:
> > I didn't added new tests because one is already here for this command (there is a test for poetry env remove --all)
> 
> I don't follow this argument. If I understand correctly, this PR fixes a bug. In that case, there should be a test that fails without the fix. If there isn't such a test, then it should be added. (It shouldn't be too difficult to create an `envs.toml` in a test and check that the concerning section is removed.)

I wasn't able to tell if consistency between the TOML file and the env folders has to be strictly enforced since no test for the remove command seems to validate this assertion. I was simply conservative and preferred not to change the tests. The TOML could be out of sync with the env folders for many reasons, for instance a user manually removing them from the cache because they take too much space on their device. 

I checked the tests for the create command (triggered by `env use`) and it appears that *some* kind of consistency is checked. The new environment should be listed in the TOML but the rest of the TOML isn't checked for consistency.

I you confirm that doing basic consistency checks makes sense, i'll considerer adding tests for that.

radoering:
Ah, now I understand: the new method is just a partial copy of the existing `remove` method, to be more precise it's more or less https://github.com/python-poetry/poetry/blob/0578ed81787449d79430360de143d916006e028b/src/poetry/utils/env/env_manager.py#L339-L359

In that case, we should avoid code duplication. Can you please do a small refactoring so that `remove` uses the new method, too. Further, the new method should not return a venv since it already takes the venv as argument. (It makes more sense that `remove` returns the venv because `remove` has to determine the venv itself.)

> I was simply conservative and preferred not to change the tests. The TOML could be out of sync with the env folders for many reasons, for instance a user manually removing them from the cache because they take too much space on their device.

We can't expect the envs.toml to be consistent, but we should keep it consistent when running a Poetry command.

> I you confirm that doing basic consistency checks makes sense, i'll considerer adding tests for that.

I think a test that checks that the entry is removed when calling `remove --all` makes sense. I even think we need two tests:

- the entry references an existing venv (removal can be handled in `remove_venv`)
- the entry references a non-existing venv (removal has to be handled outside of `remove_venv` because `remove_venv` removes exactly one venv and, thus, should only remove the entry if it matches the venv's python version)

laclouis5:
I got some time to advance on this issue.

first, I don't think it would be straightforward to make `remove` to use `remove_venv`. There is a ton of logic in the `remove` (there re many nested if-statements and a code path even spin a subprocess) and I don't feel confident trying to stuff `remove_env` in here.

Second, I tried to implement the tests for `remove --all` but I stumbled on another issue: the current test suite does not seem to check anything related to the `envs.toml`, at least in the env removal tests. **The `env.toml` file is literally not present from the temporary env folder created to test Poetry (with the sample `simple-project`).** 

I start to feel very confused about the purpose of that file and what is its expected state but I wasn't able to find any specification about it. Where can I find the documentation about Poetry's internals?

radoering:
> The current test suite does not seem to check anything related to the `envs.toml`, at least in the env removal tests.

In some places, tests are more incomplete than in others. There is a permanent issue to improve the tests suite.

> **The `env.toml` file is literally not present from the temporary env folder created to test Poetry (with the sample `simple-project`).**

The file is probably not present because we do not always create a real virtualenv if we do not need one for a unit test. If we need an `envs.toml` file for a unit test we can just create it.

> I start to feel very confused about the purpose of that file and what is its expected state but I wasn't able to find any specification about it. Where can I find the documentation about Poetry's internals?

Afaik, there is not such documentation. You have to look into the code.

I have just added some tests, improved the (in-code) documentation a bit and tried another approach, which I think covers more cases: #8831

Issue #8646, opened 2023-11-10
[1.7] `[Errno 2] No such file or directory: '/usr/pyvenv.cfg'` when running `poetry add` in `nvidia/cuda:...ubuntu22.04` image


CallumJHays:
- **Poetry version**: 1.7.0
- **Python version**: 3.10.8
- **OS version and name**: Ubuntu 22.04. base docker image: `nvidia/cuda:11.8.0-cudnn8-runtime-ubuntu22.04`
- **pyproject.toml**: https://gist.github.com/CallumJHays/af8252187ea4f7ed783d35ef19229c72 (+Dockerfile)

- [x] I am on the [latest](https://github.com/python-poetry/poetry/releases/latest) stable Poetry version, installed using a recommended method.
- [x] I have searched the [issues](https://github.com/python-poetry/poetry/issues) of this repo and believe that this is not a duplicate.
- [x] I have consulted the [FAQ](https://python-poetry.org/docs/faq/) and [blog](https://python-poetry.org/blog/) for any relevant entries or release notes.
- [x] If an exception occurs when executing a command, I executed it again in debug mode (`-vvv` option) and have included the output below.

## Issue

I'm running into a `[Errno 2] No such file or directory: '/usr/pyvenv.cfg'` error when trying to run `poetry add` or `poetry install` within a docker container based on `nvidia/cuda:11.8.0-cudnn8-runtime-ubuntu22.04`.

Debugging Details:
- The issue does not occur in `1.6.1` - only `1.7.0`
- The issue does not occur with a different base image (ie `FROM python:3.10`)
   - clearly something in the nvidia image is causing the problems - but not sure what.
- The issue does not occur unless `poetry config virtual-envs.create false`

 I included a `Dockerfile` in my attached gist, with the build args POETRY_VERSION and POETRY_VIRTUALENVS_CREATE to help isolate the issue. The error can be demonstrated with the following command:

 ```bash
 docker build . --build-arg POETRY_VERSION=1.7.0 --build-arg POETRY_VIRTUALENVS_CREATE=false
 ```

The docker build fails on the last step (`RUN poetry add six -vvv`), and produces the following output. PS: the package used to test (`six`) is arbitrary - it fails for any package I've tried

```python
  > [7/7] RUN poetry add six -vvv:                                                                                                
0.452 Loading configuration file /root/.config/pypoetry/config.toml                                                              
0.463 Skipping virtualenv creation, as specified in config file.                                                                 
0.531                                                                                                                            
[...]

PR #8667, opened 2023-11-14
release: bump version to 1.7.1

radoering:
### Fixed

- Fix an issue where sdists that call CLI tools of their build requirements could not be installed ([#8630](https://github.com/python-poetry/poetry/pull/8630)).
- Fix an issue where sdists with symlinks could not be installed due to a broken tarfile datafilter ([#8649](https://github.com/python-poetry/poetry/pull/8649)).
- Fix an issue where `poetry init` failed when trying to add dependencies ([#8655](https://github.com/python-poetry/poetry/pull/8655)).
- Fix an issue where `poetry install` failed if `virtualenvs.create` was set to `false` ([#8672](https://github.com/python-poetry/poetry/pull/8672)).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,16 @@
 # Change Log
 
+
+## [1.7.1] - 2023-11-16
+
+### Fixed
+
+- Fix an issue where sdists that call CLI tools of their build requirements could not be installed ([#8630](https://github.com/python-poetry/poetry/pull/8630)).
+- Fix an issue where sdists with symlinks could not be installed due to a broken tarfile datafilter ([#8649](https://github.com/python-poetry/poetry/pull/8649)).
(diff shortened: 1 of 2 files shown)


Secrus (dismissed):

dimbleby:
not that I'm volunteering to do it myself, but #8646 looks like a 1.7.0 regression that shouldn't be too hard to fix (by adopting some default behaviour when pyvenv.cfg is not present): if that's near the top of anyone's list then perhaps worth waiting for it

radoering:
> not that I'm volunteering to do it myself, but #8646 looks like a 1.7.0 regression that shouldn't be too hard to fix (by adopting some default behaviour when pyvenv.cfg is not present): if that's near the top of anyone's list then perhaps worth waiting for it

 #8672

Secrus (approved):

PR #8668, opened 2023-11-14
Loosen platformdirs dependency

stefanor:
4.x is backwards compatible with 3.x except that `site_cache_dir` has moved to /var/cache. Poetry doesn't use this.

https://github.com/platformdirs/platformdirs/releases/tag/4.0.0


dimbleby:
if you instead of doing a full re-lock went `poetry lock --no-update` then I expect you wouldn't get merge conflicts

NeilGirdhar:
@dimbleby This PR fixes the Poetry source dependency itself—it's not about any project.

dimbleby:
@NeilGirdhar I don't know what you think I wrote but you are talking past it

this merge request currently has merge conflicts, I expect those would be avoided if `poetry.lock` had been updated by `poetry lock --no-update`

NeilGirdhar:
Oh, sorry, I misunderstood your comment.

stefanor:
Rebased with `poetry lock --no-update`.

My expectation was that any change to pyproject would change `content-hash` and so conflicts were inevitable. Such is life with lock files...

radoering:
> My expectation was that any change to pyproject would change `content-hash` and so conflicts were inevitable. Such is life with lock files...

That's correct. Can you please re-lock (`--no-update`) with the latest poetry version (1.7.1)? (If the lockfile does not change just change the hash and do again.)

stefanor:
> That's correct. Can you please re-lock (`--no-update`) with the latest poetry version (1.7.1)? (If the lockfile does not change just change the hash and do again.)

Done. The content is as-before, with a different version in the comment. :shrug:

radoering:
> Done. The content is as-before, with a different version in the comment. 🤷

That's fine. We just want to make sure, we always use the latest version in the project.

Issue #8250, opened 2023-07-28
Proportion of Time Spent on Cloning is Significant during poetry lock --no-update


higumachan:
- [x] I have searched the [issues](https://github.com/python-poetry/poetry/issues) of this repo and believe that this is not a duplicate.
- [x] I have searched the [FAQ](https://python-poetry.org/docs/faq/) and general [documentation](https://python-poetry.org/docs/) and believe that my question is not already covered.

## Feature Request

Title: Proportion of Time Spent on Cloning is Significant during `poetry lock --no-update`

Body:

While running Poetry and cProfile locally, I noticed that a significant amount of time is spent on the cloning process, as depicted in the image below.

![image](https://github.com/python-poetry/poetry/assets/1011298/1603c4be-27d4-4b87-a52a-11197afd8529)

I made some modifications to replace the clone operation with a shallow copy, copying only necessary attributes. (I adjusted it to pass the tests by defining "necessary attributes".)

As a result, I was able to significantly reduce the time spent on the cloning operation, as can be seen in the following image. In addition, the processing time also improved significantly:

![image](https://github.com/python-poetry/poetry/assets/1011298/7c45f11c-4531-4627-a3fd-5689d8db6175)

From
```
Resolving dependencies... (58.9s)
```
to
```
Resolving dependencies... (34.5s)
```

Would you be interested in this improvement? If so, I would like to create a PR right away.

## environment

poetry: e8ca4b32ceeb1648eaba6bed49e8b36dcfbf0cc0
poetry-core: f09d84a4c6ad6d1117998ff9d46949e1bd819f5b -> 86a0ec388eb7581c3c3b3d0f78848aa85f434893

dimbleby:
duplicate #7459

dimbleby:
I guess that whether a pull request is desirable or not will depend on what it looks like.  Something that's obviously safe and a performance win is obviously wanted; something that trades performance for maintainability might not be

PR #8671, opened 2023-11-14
treat collection types in package as immutable

dimbleby:
addresses #8250, #7459: but the interesting part of the fix is in poetry-core and I'll make more comments there

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -923,13 +923,13 @@ testing = ["pytest", "pytest-benchmark"]
 
 [[package]]
 name = "poetry-core"
-version = "1.8.1"
+version = "1.9.0"
 description = "Poetry PEP 517 Build Backend"
 optional = false
 python-versions = ">=3.8,<4.0"
 files = [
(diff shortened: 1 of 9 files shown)


dimbleby:
https://github.com/python-poetry/poetry-core/pull/663 is the companion MR and, not for the first time, it will be awkward to find a sequence of merges that keeps both pipelines happy throughout

radoering:
@dimbleby I suppose we have to update poetry-core in this PR to pass mypy. Do you have the time to take care of it?

dimbleby:
gonna need to upgrade poetry core for lots of reasons though, right?  I'd think that wants handling in a separate MR and then this one can simply merge / rebase

either way I wont' be doing anything before the weekend, probably can find time then though.

radoering:
> I'd think that wants handling in a separate MR and then this one can simply merge / rebase

I assumed that mypy would fail after upgrading poetry-core without the changes of this PR but I'll try if it's possible to do it separately or if other things break. (I have the hope that all other things are just deprecation warnings we can address later.)

radoering (approved):

Issue #8674, opened 2023-11-14
The `pre-commit-hooks` page should remove the use of `poetry-lock` as it is deprecated.


iamibi:
- [x] I have searched the [issues](https://github.com/python-poetry/poetry/issues) of this repo and believe that this is not a duplicate.

## Issue

The `pre-commit-hooks` page currently shares the information on using `poetry-lock` plugin which has be deprecated. The `poetry-check` plugin instead can be used with the argument `args: ["--lock"]` as suggested by the following deprecation warning
```
poetry lock --check is deprecated, use `poetry check --lock` instead.
```

dimbleby:
`poetry lock` is not deprecated.

(Only `poetry lock --check` is deprecated, but the hook is more general than that)

PR #8675, opened 2023-11-14
Update `pre-commit-hooks` documentation

iamibi:
# Pull Request Check List

Resolves: https://github.com/python-poetry/poetry/issues/8674

### DIFF
--- a/docs/pre-commit-hooks.md
+++ b/docs/pre-commit-hooks.md
@@ -58,7 +58,7 @@ This hook is provided by the [Export Poetry Plugin](https://github.com/python-po
 {{% /warning %}}
 
 {{% note %}}
-It is recommended to run the [`poetry-lock`](#poetry-lock) hook prior to this one.
+It is recommended to run the [`poetry-lock`](#poetry-lock) or [`poetry-check`](#poetry-check) with argument `--lock` hook prior to this one.
 {{% /note %}}
 
 ### Arguments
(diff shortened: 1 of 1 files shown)


docs/pre-commit-hooks.md
```diff
@@ -58,7 +48,7 @@ This hook is provided by the [Export Poetry Plugin](https://github.com/python-po
 {{% /warning %}}
 
 {{% note %}}
-It is recommended to run the [`poetry-lock`](#poetry-lock) hook prior to this one.
```
radoering, line 61: IMO that's the only part that might be changed. Instead of recommending only `poetry-lock`, we may recommend two alternatives: `poetry-lock` or `poetry-check` with `--lock`.

radoering (changes requested):
See https://github.com/python-poetry/poetry/issues/8674#issuecomment-1811980720

docs/pre-commit-hooks.md
iamibi: gotcha, will update the same.

docs/pre-commit-hooks.md
```diff
@@ -58,7 +58,7 @@ This hook is provided by the [Export Poetry Plugin](https://github.com/python-po
...
+It is recommended to run the [`poetry-lock`](#poetry-lock) or [`poetry-check`](#poetry-check) with argument `--lock` hook prior to this one.
```
radoering, line 61:
```suggestion
It is recommended to run the [`poetry-lock`](#poetry-lock) hook or [`poetry-check`](#poetry-check) with argument `--lock` prior to this one.
```
?

radoering (approved):

PR #8701, opened 2023-11-23
fix requests auto decode file

codeskyblue:
only happens when Content-Encoding: gzip

# Pull Request Check List

Resolves: #4523


dimbleby:
poetry downloads probably millions of tar.gz files every day without issue: the real problem here is that your server is misconfigured.

the proposed fix will break things for people whose servers are correctly configured and who are (admittedly unnecessarily) zipping the tar.gz for download.  pip tried the same fix years ago and saw exactly that - https://github.com/pypa/pip/issues/1435

Having said that, they do now seem to have landed in a place where they try to compensate for misconfigured servers by using `Accept-Encoding: Identity` - see all of the code in https://github.com/pypa/pip/blob/main/src/pip/_internal/network/utils.py

personally I would prefer that you corrected your server configuration and we left poetry alone, but if you want to pursue this path here then there is more to do.

pyproject.toml
```diff
@@ -61,6 +61,7 @@ xattr = { version = "^0.10.0", markers = "sys_platform == 'darwin'" }
 
 [tool.poetry.group.dev.dependencies]
 pre-commit = ">=2.10"
+pytest-httpserver = "^1.0.8"
```
dimbleby, line 64: should be in group.test.dependencies presumably

codeskyblue:
> poetry downloads probably millions of tar.gz files every day without issue: the real problem here is that your server is misconfigured.
> 
> the proposed fix will break things for people whose servers are correctly configured and who are (admittedly unnecessarily) zipping the tar.gz for download. pip tried the same fix years ago and saw exactly that - [pypa/pip#1435](https://github.com/pypa/pip/pull/1435)
> 
> Having said that, they do now seem to have landed in a place where they try to compensate for misconfigured servers by using `Accept-Encoding: Identity` - see all of the code in https://github.com/pypa/pip/blob/main/src/pip/_internal/network/utils.py
> 
> personally I would prefer that you corrected your server configuration and we left poetry alone, but if you want to pursue this path here then there is more to do.
> 
> edit: on reflection I think that just setting `Accept-Encoding: Identity` - and making no other change - should be reasonable. `whl` and `tar.gz` files are already compressed so telling servers not to compress them again during transfer is just saving compute on both sides

updated code according to your request.

dimbleby:
looks ok to me (once you fix up the linting).  

Does it work in getting your server to behave itself?

codeskyblue:
> looks ok to me (once you fix up the linting).
> 
> Does it work in getting your server to behave itself?

I fix this issue by remove Content-Encoding from my server (flask based pypi mirror which is not support Accept-Encoding: Identity)

codeskyblue:
how to fix the linting? any tools?

dimbleby:
> how to fix the linting?

Per the errors!  ruff and black

pyproject.toml
```diff
@@ -70,6 +70,7 @@ pytest-cov = "^4.0"
 pytest-mock = "^3.9"
 pytest-randomly = "^3.12"
 pytest-xdist = { version = "^3.1", extras = ["psutil"] }
+pytest-httpserver = "^1.0.8"
```
Secrus, line 73: Do we need the extra dependency? we already have `httpretty`, isn't that enough?

codeskyblue:
To make the pr simple, I removed the tests code

Secrus:
> To make the pr simple, I removed the tests code

No chance for a PR to be merged without tests of any kind. If you don't want to write tests for that, close this PR.

radoering:
pre-commit.ci autofix

radoering (approved):

Issue #8528, opened 2023-10-10
xattr/lib_build.h file not getting installed with poetry installation


WeirdWolves:
- **Poetry version**: 1.6.1
- **Python version**: 3.8.14
- **OS version and name**: MacOS Ventura 13.5.2
- **pyproject.toml**: 

- [x] I am on the [latest](https://github.com/python-poetry/poetry/releases/latest) stable Poetry version, installed using a recommended method.
- [x] I have searched the [issues](https://github.com/python-poetry/poetry/issues) of this repo and believe that this is not a duplicate.
- [x] I have consulted the [FAQ](https://python-poetry.org/docs/faq/) and [blog](https://python-poetry.org/blog/) for any relevant entries or release notes.
- [x] If an exception occurs when executing a command, I executed it again in debug mode (`-vvv` option) and have included the output below.

## Issue
 Trying to downgrade numpy version from 1.24 to 1.19.5. It says xattr/lib_build.h file not found. Reinstalling poetry does not resolve the error.
 ImportError

  dlopen(/Users/nabekhan/Library/Application Support/pypoetry/venv/lib/python3.8/site-packages/_cffi_backend.cpython-38-darwin.so, 0x0002): symbol not found in flat namespace '_ffi_prep_closure'

  at ~/Library/Application Support/pypoetry/venv/lib/python3.8/site-packages/xattr/lib.py:7 in <module>
        3│ 
        4│ from .compat import fs_encode
        5│ 
        6│ try:
    →   7│     from ._lib import lib, ffi
        8│ except ImportError:
        9│     from .lib_build import ffi, c_source
       10│     lib = ffi.verify(c_source)
       11│ 

The following error occurred when trying to handle this error:

  Stack trace:

  12  ~/Library/Application Support/pypoetry/venv/lib/python3.8/site-packages/cleo/application.py:327 in run
       325│ 
       326│             try:
     → 327│                 exit_code = self._run(io)
       328│             except BrokenPipeError:
       329│                 # If we are piped to another process, it may close early and send a

  11  ~/Library/Application Support/pypoetry/venv/lib/python3.8/site-packages/poetry/console/application.py:190 in _run
       188│         self._load_plugins(io)
       189│ 
     → 190│         exit_code: int = super()._run(io)
       191│         return exit_code
       192│ 

  10  ~/Library/Application Support/pypoetry/venv/lib/python3.8/site-packages/cleo/application.py:431 in _run
       429│             io.input.interactive(interactive)
[...]

PR #8702, opened 2023-11-23
Bump xattr version for darwin

mrmeszaros:
Resolves: #8528 

Bumping the version fixes the missing `xattr/lib_build.h` file issue.

I have one question: is this lib only needed for osx / darwin?
If `yes` - I would like to remove the TODO comment
If `not` - I would like to test this for other `sys_platform`s


Secrus:
There is a `TODO` comment [here](https://github.com/python-poetry/poetry/blob/e6c1c99e3f966cc8c3c8af049136861d57b1c95f/src/poetry/utils/env/env_manager.py#L674) that explains why this comment is there. If you want to explore that further, feel free to do it and submit a PR, but it would be out of scope for this PR here.

Secrus:
@mrmeszaros I have opened #8801 since I had some problems fixing issues in your PR. I added you as co-author to commit in my PR, I hope you don't mind.

PR #8706, opened 2023-11-24
Increase verboseness of `poetry update` documentation

DhenPadilla:
# Pull Request Check List

Resolves: #8707

- [x] Added **tests** for changed code [None needed]
- [x] Updated **documentation** for changed code. [No code changed]

### DIFF
--- a/docs/cli.md
+++ b/docs/cli.md
@@ -290,8 +290,10 @@ poetry update requests toml
 Note that this will not update versions for dependencies outside their
 [version constraints]({{< relref "dependency-specification#version-constraints" >}})
 specified in the `pyproject.toml` file.
-In other terms, `poetry update foo` will be a no-op if the only latest version for `foo` lies outside the constraint specified for `foo` in the `pyproject.toml` file. (More info on constraints can be found in our (Version Requirements)[https://python-poetry.org/docs/dependency-specification/#version-constraints].)
-In order for `foo` to be updated, you must update the constraint. You can do this using the `add` command.
+In other terms, `poetry update foo` will be a no-op if the version constraint
+specified for `foo` is `~2.3` or `2.3` and `2.4` is available.
+In order for `foo` to be updated, you must update the constraint, for example `^2.3`.
+You can do this using the `add` command.
 
 ### Options
 
--- a/docs/cli.md
+++ b/docs/cli.md
@@ -289,8 +289,7 @@ poetry update requests toml
 
 Note that this will not update versions for dependencies outside their (SemVer) version constraints specified
 in the `pyproject.toml` file.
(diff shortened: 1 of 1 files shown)


docs/cli.md
```diff
@@ -287,10 +287,11 @@ If you just want to update a few packages and not all, you can list them as such
 poetry update requests toml
 ```
 
-Note that this will not update versions for dependencies outside their version constraints specified
-in the `pyproject.toml` file. In other terms, `poetry update foo` will be a no-op if the version constraint
-specified for `foo` is `~2.3` or `2.3` and `2.4` is available. In order for `foo` to be updated, you must
-update the constraint, for example `^2.3`. You can do this using the `add` command.
+Note that this will not update versions for dependencies outside their (SemVer) version constraints specified
```
dimbleby, line 290: most of the additions here are unhelpful for people who have written their constraints not using the caret.  

Reproducing a tiny fragment of https://python-poetry.org/docs/dependency-specification/#version-constraints here is redundant - and too hyper-focused on the particular misunderstanding that you happened to have.

Suggest that simply cross-referencing from the words "version constraints" to https://python-poetry.org/docs/dependency-specification/#version-constraints should be sufficient

DhenPadilla: Done!

dimbleby: the addition of "(Semver)" is still unwanted: only caret constraints are semver constraints

what I had in mind was simply 

> Note that this will not update versions for dependencies outside their [version constraints]({{ cross-reference here}}) specified ...

where an internal cross-reference is the right way to do that rather than pointing directly at  python-poetry.org (but I'm not familiar with the exact syntax needed so you'll have either to figure it out yourself or take a couple of tries to get it right, sorry!)

docs/cli.md
```diff
@@ -287,10 +287,10 @@ If you just want to update a few packages and not all, you can list them as such
...
+in the `pyproject.toml` file.
```
radoering, line 291:
```suggestion
Note that this will not update versions for dependencies outside their
[version constraints]({{< relref "dependency-specification#version-constraints" >}})
specified in the `pyproject.toml` file.
```

docs/cli.md
```diff
@@ -287,10 +287,10 @@ If you just want to update a few packages and not all, you can list them as such
...
+In other terms, `poetry update foo` will be a no-op if the only latest version for `foo` lies outside the constraint specified for `foo` in the `pyproject.toml` file. (More info on constraints can be found in our (Version Requirements)[https://python-poetry.org/docs/dependency-specification/#version-constraints].)
+In order for `foo` to be updated, you must update the constraint. You can do this using the `add` command.
```
radoering, line 293:
```suggestion
In other terms, `poetry update foo` will be a no-op if the version constraint
specified for `foo` is `~2.3` or `2.3` and `2.4` is available.
In order for `foo` to be updated, you must update the constraint, for example `^2.3`.
You can do this using the `add` command.
```

radoering (changes requested):

radoering (approved):

Issue #8718, opened 2023-11-27
Add 'dev' as version bump rule for developmental releases. 


gaileyleseman:
- [x] I have searched the [issues](https://github.com/python-poetry/poetry/issues) of this repo and believe that this is not a duplicate.
- [x] I have searched the [FAQ](https://python-poetry.org/docs/faq/) and general [documentation](https://python-poetry.org/docs/) and believe that my question is not already covered.

## Feature Request

I would like to use `poetry version dev` to automatically increment my versions according to the PEP-440 rules for developmental releases. 

Expected behaviour: 

| rule       | before  | after   |
| ---------- |---------|---------|
| dev        | 1.0.2   | 1.0.2.dev0 |
| dev        | 1.0.2.dev0 | 1.0.2.dev1 |
| dev        | 1.0.2a0 | 1.0.2a0.dev0 |

dimbleby:
that's not an increment, that's a decrement

radoering:
@gaileyleseman If you are still interested in this, please fix the expected behavior in the issue description.

vlaszdunov:
what if the dev state is added as a `--dev` flag?

### examples:

| command | before | after |
| :--- | :--- | :---|
| `poetry version major --dev` | 1.1.0 | 2.0.0.dev0 |
| `poetry version minor --dev` | 1.1.0 | 1.2.0.dev0 |
| `poetry version patch --dev` | 1.1.0 | 1.1.1.dev0 |
| `poetry version premajor --dev` | 1.1.0 | 2.0.0.a0.dev0 |
| `poetry version preminor --dev` | 1.1.0 | 1.2.0.a0.dev0 |
| `poetry version prepatch --dev` | 1.1.0 | 1.1.1.a0.dev0 |

In this case command `poetry version dev` will bump .dev version:
| command | before | after |
| :--- | :--- | :---|
| `poetry version dev` | 1.1.0.dev0 | 1.1.0.dev1 |
| `poetry version dev` | 1.1.0.dev1 | 1.1.0.dev2 |

expected behaviors in other cases:
| command | before | after |
| :--- | :--- | :---|
| `poetry version prerelease --dev` | 1.1.0 | 1.1.1.a0.dev0 |
| `poetry version prerelease --next-phase` | 1.1.0.a2.dev1 | 1.1.0.a2 |
| `poetry version prerelease --next-phase --dev` | 1.1.0.a2.dev1 | 1.1.0.b0.dev0 |

what do you think about it?

[...]

PR #8719, opened 2023-11-27
Added 'dev' as a version bump rule to increment versions with .devM

gaileyleseman:
This adds `dev` as a version bump rule for [developmental releases](https://peps.python.org/pep-0440/#developmental-releases). This was already implemented in `poetry-core`, I've just added the option to the cli. 

This is my first contribution, I hope I followed the correct procedure!

# Pull Request Check List

Resolves: #8718 

### DIFF
--- a/docs/cli.md
+++ b/docs/cli.md
@@ -675,7 +675,7 @@ bump rule is provided.
 
 The new version should be a valid [PEP 440](https://peps.python.org/pep-0440/)
 string or a valid bump rule: `patch`, `minor`, `major`, `prepatch`, `preminor`,
-`premajor`, `prerelease`.
+`premajor`, `prerelease`, `dev`.
 
 {{% note %}}
 
@@ -697,6 +697,9 @@ The table below illustrates the effect of these rules with concrete examples.
 | prerelease | 1.0.2   | 1.0.3a0 |
 | prerelease | 1.0.3a0 | 1.0.3a1 |
 | prerelease | 1.0.3b0 | 1.0.3b1 |
+| dev        | 1.0.2   | 1.0.2.dev0 |
+| dev        | 1.0.2.dev0 | 1.0.2.dev1 |
+| dev        | 1.0.2a0 | 1.0.2a0.dev0 |
 
 The option `--next-phase` allows the increment of prerelease phase versions.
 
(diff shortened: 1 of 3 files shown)


tests/console/commands/test_version.py
```diff
@@ -46,6 +46,9 @@ def tester(command_tester_factory: CommandTesterFactory) -> CommandTester:
         ("1.2.3beta1", "prerelease", "1.2.3b2"),
         ("1.2.3b1", "prerelease", "1.2.3b2"),
         ("1.2.3", "prerelease", "1.2.4a0"),
+        ("1.2.3", "dev", "1.2.3.dev0"),
```
dimbleby, line 49: if this is actually what it is doing then it is bugged, because the dev release is earlier than the non-dev release

gaileyleseman: I see, I'm experimenting with automatically bumping versions through ci/cd pipelines, for which the `poetry version` command could be very useful. 

I've updated the logic and the docs/tests such that the version number always moves forward.

docs/cli.md
```diff
@@ -697,6 +701,9 @@ The table below illustrates the effect of these rules with concrete examples.
 | prerelease | 1.0.2   | 1.0.3a0 |
 | prerelease | 1.0.3a0 | 1.0.3a1 |
 | prerelease | 1.0.3b0 | 1.0.3b1 |
+| dev        | 1.0.2   | 1.0.3a0.dev0 |
+| dev        | 1.1.0a0 | 1.1.0a0.dev0 |
+| dev        | 1.0.2a0 | 1.0.2a0.dev0 |
```
radoering, line 706: Still a decrement. (`1.1.0a0.dev0` comes before `1.1.0a0`.)

src/poetry/console/commands/version.py
```diff
@@ -122,6 +123,15 @@ def increment_version(
...
                 new = parsed.next_patch().first_prerelease()
+        elif rule == "dev":
+            if parsed.is_stable():
+                new = self.increment_version(
+                    version, "prerelease", next_phase
+                ).first_devrelease()
+            elif parsed.is_devrelease():
+                new = parsed.next_devrelease()
+            else:
+                new = parsed.first_devrelease()
```
radoering, line 134: That's a decrement.

radoering (changes requested):
In general, I think we should define the target behavior in #8718 first because it's ambigious. For example, we could also bump `1.0.2` to `1.0.3.dev0` instead of `1.0.3a0.dev0`. (Both are correct.)

PR #8733, opened 2023-12-01
Add shadowing warning to supplemental source docs

JockeTF:
We may want people to consider what happens if someone publishes a new package to PyPI which matches one in their supplemental source. There were a few dependency confusion vulnerabilities a while back that were caused by something similar. I presume lock files help to some extent though.

# Pull Request Check List

Resolves: None

### DIFF
--- a/docs/repositories.md
+++ b/docs/repositories.md
@@ -232,6 +232,14 @@ poetry source add --priority=supplemental https://foo.bar/simple/
 
 There can be more than one supplemental package source.
 
+{{% warning %}}
+
+Take into account that someone could publish a new package to a primary source
+which matches a package in your supplemental source. They could coincidentally
+or intentionally replace your dependency with something you did not expect.
+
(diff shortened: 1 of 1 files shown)


docs/repositories.md
```diff
@@ -232,6 +232,12 @@ poetry source add --priority=supplemental https://foo.bar/simple/
 
 There can be more than one supplemental package source.
 
+{{% warning %}}
+
+Take into account that someone could publish a new package to PyPI which matches one in your supplemental source.
```
radoering, line 237:
```suggestion
Take into account that someone could publish a new package to a primary source which matches one in your supplemental source.
```

I'd put it more generally.

JockeTF: Okay! How about this?

radoering: I think the second sentence is a valuable addition. 👍

radoering (approved):

PR #8737, opened 2023-12-04
(🎁) upgrade warning about inconsistent lock file to an error

KotlinIsland:
# Pull Request Check List

- Resolves: #5003

This is an extremely common issue that I see people running into all the time, there is a warning at the top of the output about the invalid state of the project, but it is instantly hidden by a wall of other output.

Finally poetry says that the version doesn't exist. Confusingly, poetry is only referring to versions **in the lock file** that don't exist.

- [🚀] Added **tests** for changed code.
- [🚀] Updated **documentation** for changed code.


src/poetry/mixology/incompatibility.py
```diff
@@ -413,7 +413,7 @@ def _try_requires_forbidden(
             cause: PythonCause = latter.cause
             buffer.append(f"which requires Python {cause.python_version}")
         elif isinstance(latter.cause, NoVersionsCause):
-            buffer.append("which doesn't match any versions")
+            buffer.append("which doesn't match any versions in the lock file")
```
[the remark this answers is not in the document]
KotlinIsland: Okay, I have just updated the warning about an inconsistent lock file to an error.

src/poetry/installation/installer.py
```diff
@@ -245,12 +245,12 @@ def _do_install(self) -> int:
...
-                self._io.write_error_line(
-                    "<warning>"
-                    "Warning: poetry.lock is not consistent with pyproject.toml. "
+                raise ValueError(
+                    "<error>"
+                    "Error: poetry.lock is not consistent with pyproject.toml. "
                     "You may be getting improper dependencies. "
                     "Run `poetry lock [--no-update]` to fix it."
-                    "</warning>"
+                    "</error>"
```
radoering, line 253: I assume we don't need `<error>` and `Error:` when raising an error. Further, we can remove `"You may be getting improper dependencies."` because now you can't install anymore.

ilyagr:
Will this allow running Poetry 1.7 (or 1.8) on a lockfile generated by Poetry 1.6? Currently, that generates a warning since the lockfile doesn't match exactly what Poetry 1.7 would generate; I'd hate for that to become an error.

BTW, I agree that erroring out when the lockfile is incompatible with `pyproject.toml` is a good idea.

radoering:
> that generates a warning since the lockfile doesn't match exactly what Poetry 1.7 would generate

It does not generate _this_ warning, does it? Actually, I'm not aware that it generates a warning at all. The last update of the lockfile version was in Poetry 1.3.

ilyagr:
> Actually, I'm not aware that it generates a warning at all.

Hmm... I was sure I saw it, but I couldn't actually reproduce it. I will let you know if I find it again.

For reference, here is what the diff looks like between a lockfile generated by Poetry 1.3 and Poetry 1.7.1. The differences are slight, but this is what I *thought* (seemingly incorrectly?) Poetry was complaining about.

```diff
diff --git a/poetry.lock b/poetry.lock
index e933ba2ee6...955471c81b 100644
--- a/poetry.lock
+++ b/poetry.lock
@@ -1,10 +1,9 @@
-# This file is automatically @generated by Poetry and should not be changed by hand.
+# This file is automatically @generated by Poetry 1.7.1 and should not be changed by hand.

 [[package]]
 name = "babel"
 version = "2.14.0"
 description = "Internationalization utilities"
-category = "dev"
 optional = false
 python-versions = ">=3.7"
 files = [
```

(there are then many repeats of the `-category = "dev"` line).

radoering:
`category` is not written anymore because it has not been read even longer. This should not produce any warning since we decided not to bump the lockfile version. See #7637 for details.

The only risk in this PR is that users edited the pyproject.toml, did not lock because they knew that their change did not affect the lockfile except for the hash and deliberately ignored this warning. This workflow will not work anymore after this change. I suppose we can/should take that risk to improve the feedback for the majority of users.

ilyagr:
> The only risk in this PR is that users edited the pyproject.toml, did not lock because they knew that their change did not affect the lockfile except for the hash and deliberately ignored this warning.

I did not realize that `poetry.lock` contained a hash of the `pyproject.toml`, so changing a comment in the latter invalidates the former. This might be the warning that confused me.

>  I suppose we can/should take that risk to improve the feedback for the majority of users.

This seems like a reasonable tradeoff. One way to make it friendlier is to mention this possibility in the error message, saying something "pyproject.toml changed since poetry.lock was generated". When it says that "poetry.lock is incompatible with pyproject.toml", I assumed it checked that the versions of packages in poetry.lock don't satisfy the version specifiers in pyproject.toml.

radoering:
> I did not realize that `poetry.lock` contained a hash of the `pyproject.toml`, so changing a comment in the latter invalidates the former. This might be the warning that confused me.

It's not a hash of the entire `pyproject.toml`. Only fields that are relevant for locking are considered.

> "pyproject.toml changed since poetry.lock was generated"

Maybe we should say "relevant fields of pyproject.toml changed ..."?

ilyagr:
> Maybe we should say "relevant fields of pyproject.toml changed ..."?

Maybe "Dependencies changed in pyproject.toml since poetry.lock was generated"?

src/poetry/installation/installer.py
```diff
@@ -245,12 +245,9 @@ def _do_install(self) -> int:
             locked_repository = self._locker.locked_repository()
 
             if not self._locker.is_fresh():
-                self._io.write_error_line(
-                    "<warning>"
-                    "Warning: poetry.lock is not consistent with pyproject.toml. "
-                    "You may be getting improper dependencies. "
+                raise ValueError(
+                    "poetry.lock is not consistent with pyproject.toml. "
```
radoering, line 249: @KotlinIsland What's your opinion on @ilyagr 's proposal?
Change the message to `"Dependencies changed in pyproject.toml since poetry.lock was generated. "` or something similar or keep as is?

Strictly speaking, it's not only dependencies (see https://github.com/python-poetry/poetry/blob/9756b57995bd5cf5e0b7f89bd08a7c71ae6c7d98/src/poetry/packages/locker.py#L54-L60), so `"Relevant fields"` might be more correct although all fields are related to dependencies...

Secrus: I'd say that "relevant fields" is not verbose enough for end users. If we were to use that, I think those fields would have to be named in the error message.

radoering: And "Dependencies"? Not accurate enough or sufficient? (I wonder if it will be a bit cumbersome to name all fields?)

KotlinIsland: I think the current message is sufficient. If it's desired to change it, perhaps that could be done in a follow-up PR.

I'd rather merge it sooner than later, and the wording can be changed in a follow-up PR if needed.

ilyagr: I have no objection to doing this in a follow-up PR, but here's another thought I had:

> `pyproject.toml` changed significantly since `poetry.lock` was generated

radoering (approved):

PR #8748, opened 2023-12-05
Add note to help resolve confusion around `poetry run` and external env management

jamesowers-cohere:
Add a note to help resolve confusion around `poetry run` and external env management.


radoering (commented):
I'd probably insert the note after the text.

radoering (approved):

Issue #8756, opened 2023-12-07
Broken verbose/debug logging for poetry build with a custom build script


vit-zikmund:
- **Poetry version**: 1.7.1
- **Python version**: 3.11
- **OS version and name**: Fedora 38
- **pyproject.toml**: 
  <details><summary>hidden here</summary>

  ```toml
  [tool.poetry]
  name = "poetry-custom-build"
  version = "0.1.0"

  [tool.poetry.dependencies]
  python = "^3.10"

  [tool.poetry.build]
  script = "build.py"  # there's nothing in it

  [build-system]
  requires = ["poetry-core>=1.0.0", 'tomli; python_version < "3.11"']
  build-backend = "poetry.core.masonry.api"
  ```
</details>

- [x] I am on the [latest](https://github.com/python-poetry/poetry/releases/latest) stable Poetry version, installed using a [recommended method](https://python-poetry.org/docs/#installation).
- [x] I have searched the [issues](https://github.com/python-poetry/poetry/issues) of this repo and believe that this is not a duplicate.
- [x] I have consulted the [FAQ](https://python-poetry.org/docs/faq/) and [blog](https://python-poetry.org/blog/) for any relevant entries or release notes.
- [x] If an exception occurs when executing a command, I executed it again in debug mode (`-vvv` option) and have included the output below.

## Issue
I'm not getting any debug `-vvv` logs from the `poetry build` command when using a custom build script with additional dependencies. The wheel builder source code obviously has some `logger.debug()` calls, but they're not getting  through.

Note, I was able to pinpoint the root cause. PR is on its way.

PR #8760, opened 2023-12-08
Fix debug logging for `build` with a build script

vit-zikmund:
# Pull Request Check List

Resolves: #8756 

  Haven't seen any existing tests for log output :thinking: should I hack it nevertheless?
  This shouldn't need any change.

This change aims to fix the muted debug level logs for builds with a custom build script.

Trying to understand in depth what the optional build script can do, I ran into a very scarce output despite using `-vvv`. Roaming the building/packaging code, I found out there's definitely some debug logging that might come in handy but is not getting out.

First I found a [piece of logging](https://github.com/python-poetry/poetry/blob/6f9de73b7525504f21a6eaf46e66bbd49708bd7b/src/poetry/utils/env/__init__.py#L69-L98) (https://github.com/python-poetry/poetry/commit/a7a42905 by @abn) that lacked a newline. I tried to fix that in line with what I understood the original code tries to do (I still have a feeling I didn't hit the exact spot, but it sure works better now). It also seemed like a lost opportunity not to log the pip call there, so I added it for debug. It looks quite good as is, but we it might want to prefix those lines with something to make it clear it's a 3rd party log. Either way, this doesn't really tackle the original problem...

I found that despite the `poetry-core` builders' loggers [have their log level set properly](https://github.com/python-poetry/poetry/blob/6f9de73b7525504f21a6eaf46e66bbd49708bd7b/src/poetry/console/application.py#L267-L279), they have no handlers of their own, so rely on the parent's (root logger's) custom handler [set a couple lines above](https://github.com/python-poetry/poetry/blob/master/src/poetry/console/application.py#L261). This works nice, until something touches the root logger, which is exactly what happens by default in the [`virtualenv.cli_run` call](https://github.com/python-poetry/poetry/blob/6f9de73b7525504f21a6eaf46e66bbd49708bd7b/src/poetry/utils/env/env_manager.py#L671). With the default `setup_logging=True` it pretty much resets the root logger level to `WARNING` (or to its desired verbosity in `args`, but that's not being used). I'm setting the `setup_logging=False`, which then adopts the root logger as is. That fixes the original problem, now every Poetry's `info` and `debug` logs pass through again, but also all `info` and `debug` logs from `virtualenv` hitch a ride, which is not the prettiest thing under the sun. But it's the debug logs we're after, right, so this might be a good thing!

Either way, this surely requires at least your aesthetic opinion, as I'm not aware of any logging guideline you stand by. Thank you for helping me out.

PS: I think there's a way how to avoid the `virutalenv`'s verbose logs.
- One way is by setting the custom handler on the particular named log handlers and setting back `setup_logging=True` for `virtualenv`. But that tramples the root logger, which definitely sucks and seems hardly as a conceptual thing to do.
- ~~Another way might be to monkey-patch the `virtualenv.report.LOGGER` (root logger) to some logger we control.~~ Won't work, they directly use `logging.info` etc. for logging :disappointed: that always goes back to the root logger.

### DIFF
--- a/src/poetry/console/logging/io_formatter.py
+++ b/src/poetry/console/logging/io_formatter.py
@@ -1,10 +1,10 @@
 from __future__ import annotations
 
 import logging
-import os
 import sys
 import textwrap
 
+from pathlib import Path
 from typing import TYPE_CHECKING
 
 from poetry.console.logging.filters import POETRY_FILTER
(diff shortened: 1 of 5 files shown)


vit-zikmund:
I've eventually realized improving the third-party decoration logic in `IOFormatter` could make this look acceptable. The last commit bears a take on that.

radoering:
Thank you for going down this rabbit hole and providing such a detailed explanation.

> I've eventually realized improving the third-party decoration logic in IOFormatter could make this look acceptable. The last commit bears a take on that.

Trying `poetry build` with the example from the linked issue, I get `[filelock:filelock]` and `[virtualenv]` on Linux but `[lib:filelock]` and `[lib]` on Windows. So probably something we should improve (and write a unit test for).

Without having taken a closer look, I'd suppose to have one test for the third party logging logic in `io_formatter` and one test for the `virtualenv.cli_run()` thing.

> Haven't seen any existing tests for log output

There are tests where we check the output `via io.fetch_output()` and `io.fetch_error()`. (Not sure if that's what you meant.)

vit-zikmund:
Good points and pointers @radoering , thanks for giving it a whirl under Windows. I'll see what I can do, despite Win not being readily available around my premises.

radoering (approved):

vit-zikmund:
Thank you @radoering for finishing up this PR, it's an honor. I'm sorry I wasn't able to be of a better further assistance.

Issue #8779, opened 2023-12-11
`poetry init` does not allow `é` in the authors field


jmg-duarte:
- **Poetry version**: 1.7.1
- **Python version**: 3.12.0
- **OS version and name**: MacOS Sonoma 14.1.1
- **pyproject.toml**: NA

- [X] I am on the [latest](https://github.com/python-poetry/poetry/releases/latest) stable Poetry version, installed using a [recommended method](https://python-poetry.org/docs/#installation).
- [X] I have searched the [issues](https://github.com/python-poetry/poetry/issues) of this repo and believe that this is not a duplicate.
- [X] I have consulted the [FAQ](https://python-poetry.org/docs/faq/) and [blog](https://python-poetry.org/blog/) for any relevant entries or release notes.
- [X] If an exception occurs when executing a command, I executed it again in debug mode (`-vvv` option) and have included the output below.

## Issue

I've read https://github.com/python-poetry/poetry/issues/798 and https://github.com/python-poetry/poetry/issues/4258, those issues deal with "post-`init`", my issue is _with_/_about_ `init`.

As stated in the title, `poetry init` does not allow for `é` to exist in the authors field.
```
Author [José Duarte <redacted>, n to skip]:  
Invalid author string. Must be in the format: John Smith <john@example.com>
```
If I try `Jose...` instead, it works though.

I didn't try other characters, but I'd suspect they also fail.

dimbleby:
For me, é fails but é succeeds.  yuk.

I suspect those of us with ascii names are less motivated to figure out the difference between these things.  Merge request welcome, I imagine.

jmg-duarte:
Could you share the code points? My computer doesn't catch a difference between them when pasted in Python:
```python
ord("é".encode("utf-8")) # first one
>>> b'\xc3\xa9'
ord("é".encode("utf-8")) # second one
>>> b'\xc3\xa9'
```

With the original regex:
https://github.com/python-poetry/poetry-core/blob/56c03764e1359615adfcacab12ababd7c57a0e90/src/poetry/core/packages/package.py#L38-L40

[...]

PR #8798, opened 2023-12-15
normalize gitconfig's author field to UTF-8 NFC

jmg-duarte:
# Pull Request Check List

Resolves: #8779

Not sure if the checklist is applicable to this one, let me know!

### DIFF
--- a/src/poetry/console/commands/init.py
+++ b/src/poetry/console/commands/init.py
@@ -142,7 +142,7 @@ The <c1>init</c1> command creates a basic <comment>pyproject.toml</> file in the
         question = self.create_question(
             f"Author [<comment>{author}</comment>, n to skip]: ", default=author
         )
-        question.set_validator(lambda v: self._validate_author(v, author))
+        question.set_validator(lambda v: InitCommand._validate_author(v, author))
         author = self.ask(question)
 
         authors = [author] if author else []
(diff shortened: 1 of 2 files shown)


radoering:
Can you add a test, please?
(In poetry-core, there is already a test for `combine_unicode` so I suppose it should be possible to add a test for `init`, here.)

jmg-duarte:
> Can you add a test, please?

Added. Had to make `_validate_author` static, but it's inline with other validations.

I'm currently just testing for the offending character though, let me know if you know of more troublemakers I should add!

src/poetry/console/commands/init.py
```diff
@@ -142,7 +142,7 @@ def handle(self) -> int:
         question = self.create_question(
             f"Author [<comment>{author}</comment>, n to skip]: ", default=author
         )
-        question.set_validator(lambda v: self._validate_author(v, author))
+        question.set_validator(lambda v: InitCommand._validate_author(v, author))
```
radoering, line 145: No need to change this line. (We can access a static method via `self`.)

tests/console/commands/test_init.py
```diff
@@ -993,6 +993,24 @@ def test_validate_package_invalid(name: str) -> None:
         assert InitCommand._validate_package(name)
 
 
+@pytest.mark.parametrize(
+    "author",
+    [
+        # As MacOS saves "é" — 65 CC 81
+        "José Duarte",
+        # As written by a MacOS keyboard — C3 A9
+        "José Duarte",
```
radoering, line 1002: It seems like both strings are the same (C3 A9). The test succeeds even after reverting your change.

radoering (changes requested):

src/poetry/console/commands/init.py
jmg-duarte: TIL

radoering (approved):

Issue #8704, opened 2023-11-24
`poetry check` does not consider the package source


lucemia:
- [x] I have searched the [issues](https://github.com/python-poetry/poetry/issues) of this repo and believe that this is not a duplicate.
- [x] I have searched the [FAQ](https://python-poetry.org/docs/faq/) and general [documentation](https://python-poetry.org/docs/) and believe that my question is not already covered.

## Feature Request

It would be excellent if `poetry check` could encompass common errors? when defining package sources.

### Source is undefined
- if there is a package point to not existing source

```yaml
[tool.poetry]
name = "PythonProjects"
version = "2.0.0"
homepage = "https://github.com/roghu/py3_projects"
license = "MIT"
readme = "README.md"
authors = ["Dependabot <support@dependabot.com>"]
description = "Various small python projects."

[tool.poetry.dependencies]
python = "^3.7"
requests = { version = "2.18.0", source = "custom-1" }

[[tool.poetry.source]]
name = "custom"
url = "https://some.internal.registry.com/pypi/"
```

- poetry install will failed directly

```
(pythonprojects-py3.11) ➜  dependabot-source-constraint git:(main) ✗ poetry install
Warning: In a future version of Poetry, PyPI will be disabled automatically if at least one custom primary source is configured. In order to avoid a breaking change and make your pyproject.toml forward compatible, add PyPI explicitly via 'poetry source add pypi'. By the way, this has the advantage that you can set the priority of PyPI as with any other source.
Updating dependencies
Resolving dependencies... (0.0s)

Repository "custom-1" does not exist.
```

### Package's source is not consistence 
- The behavior is undefined when a package is defined with different source configurations.

```yaml
[tool.poetry]
name = "PythonProjects"
version = "2.0.0"
homepage = "https://github.com/roghu/py3_projects"
license = "MIT"
readme = "README.md"
authors = ["Dependabot <support@dependabot.com>"]
description = "Various small python projects."


[tool.poetry.dependencies]
python = "^3.7"

requests = { version = "2.18.0", source = "custom-2" }
[tool.poetry.group.dev]
optional = true

[tool.poetry.group.dev.dependencies]
requests = { version = "2.18.0", extras=["dev"], source = "custom-3" }

[tool.poetry.group.test]
optional = true

[tool.poetry.group.test.dependencies]
requests = { version = "2.18.0", extras=["test"], source = "custom-1" }
 

[[tool.poetry.source]]
name = "custom-1"
[...]

PR #8802, opened 2023-12-18
fix #8614 source constriants failed with extras

### LINKED ISSUE #8514: Failed to `poetry add` with false version conflict
- **Poetry version**: Poetry (version 1.6.1)
- **Python version**: Python:         3.9.16
- **OS version and name**: macOS 13.4.1
- **pyproject.toml**: 
```
[tool.poetry]
name = "test"
version = "0.1.0"
description = ""
authors = ["Hikaru <hkak03key@gmail.com>"]
readme = "README.md"

[tool.poetry.dependencies]
python = "^3.9"


[build-system]
requires = ["poetry-core"]
```

## Issue

I tried to install "sqlalchemy-bigquery" but failed with version conflict.
But I think this is the false error.

<details>
<summary>log</summary>

```
$ poetry add "sqlalchemy-bigquery"
Using version ^1.8.0 for sqlalchemy-bigquery

Updating dependencies
Resolving dependencies... (0.0s)

The current project's Python requirement (>=3.9,<4.0) is not compatible with some of the required packages Python requirement:
  - sqlalchemy-bigquery requires Python >=3.7, <3.12, so it will not be satisfied for Python >=3.12,<4.0

Because no versions of sqlalchemy-bigquery match >1.8.0,<2.0.0
 and sqlalchemy-bigquery (1.8.0) requires Python >=3.7, <3.12, sqlalchemy-bigquery is forbidden.
So, because test depends on sqlalchemy-bigquery (^1.8.0), version solving failed.

  • Check your dependencies Python requirement: The Python requirement can be specified via the `python` or `markers` properties

    For sqlalchemy-bigquery, a possible solution would be to set the `python` property to ">=3.9,<3.12"

    https://python-poetry.org/docs/dependency-specification/#python-restricted-dependencies,
    https://python-poetry.org/docs/dependency-specification/#using-environment-markers
```
</details>

<details>
<summary>log (debug)</summary>

```
$ poetry add -vvv "sqlalchemy-bigquery"
Using virtualenv: /Users/murakami_hikaru/Library/Caches/pypoetry/virtualenvs/test-2WwrecSa-py3.9
[keyring.backend] Loading KWallet
[keyring.backend] Loading SecretService
[keyring.backend] Loading Windows
[keyring.backend] Loading chainer
[keyring.backend] Loading libsecret
[keyring.backend] Loading macOS
Creating new session for pypi.org
Source (PyPI): 20 packages found for sqlalchemy-bigquery *
Using version ^1.8.0 for sqlalchemy-bigquery

Updating dependencies
Resolving dependencies...
   1: fact: test is 0.1.0
   1: derived: test
   1: fact: test depends on sqlalchemy-bigquery (^1.8.0)
   1: selecting test (0.1.0)
   1: derived: sqlalchemy-bigquery (>=1.8.0,<2.0.0)
Source (PyPI): 1 packages found for sqlalchemy-bigquery >=1.8.0,<2.0.0
   1: fact: sqlalchemy-bigquery (1.8.0) requires Python >=3.7, <3.12
   1: derived: not sqlalchemy-bigquery (==1.8.0)
Source (PyPI): 0 packages found for sqlalchemy-bigquery >1.8.0,<2.0.0
Falling back to installed packages to discover metadata for sqlalchemy-bigquery
Found 0 compatible packages for sqlalchemy-bigquery
   1: fact: no versions of sqlalchemy-bigquery match >1.8.0,<2.0.0
   1: conflict: no versions of sqlalchemy-bigquery match >1.8.0,<2.0.0
   1: ! sqlalchemy-bigquery (>1.8.0,<2.0.0) is partially satisfied by not sqlalchemy-bigquery (==1.8.0)
   1: ! which is caused by "sqlalchemy-bigquery (1.8.0) requires Python >=3.7, <3.12"
   1: ! thus: sqlalchemy-bigquery is forbidden
   1: ! sqlalchemy-bigquery (>=1.8.0,<2.0.0) is satisfied by sqlalchemy-bigquery (>=1.8.0,<2.0.0)
   1: ! which is caused by "test depends on sqlalchemy-bigquery (^1.8.0)"
   1: ! thus: version solving failed
   1: Version solving took 0.013 seconds.
   1: Tried 1 solutions.

  Stack trace:

  4  ~/.anyenv/envs/pyenv/versions/3.9.16/lib/python3.9/site-packages/poetry/puzzle/solver.py:155 in _solve
      153│
      154│         try:
    → 155│             result = resolve_version(self._package, self._provider)
      156│
      157│             packages = result.packages

  3  ~/.anyenv/envs/pyenv/versions/3.9.16/lib/python3.9/site-packages/poetry/mixology/__init__.py:18 in resolve_version
       16│     solver = VersionSolver(root, provider)
       17│
    →  18│     return solver.solve()
       19│

  2  ~/.anyenv/envs/pyenv/versions/3.9.16/lib/python3.9/site-packages/poetry/mixology/version_solver.py:163 in solve
      161│             next: str | None = self._root.name
      162│             while next is not None:
    → 163│                 self._propagate(next)
      164│                 next = self._choose_package_version()
      165│

  1  ~/.anyenv/envs/pyenv/versions/3.9.16/lib/python3.9/site-packages/poetry/mixology/version_solver.py:202 in _propagate
      200│                     # where that incompatibility will allow us to derive new assignments
      201│                     # that avoid the conflict.
    → 202│                     root_cause = self._resolve_conflict(incompatibility)
      203│
      204│                     # Back jumping erases all the assignments we did at the previous

  SolveFailure

  The current project's Python requirement (>=3.9,<4.0) is not compatible with some of the required packages Python requirement:
    - sqlalchemy-bigquery requires Python >=3.7, <3.12, so it will not be satisfied for Python >=3.12,<4.0

  Because no versions of sqlalchemy-bigquery match >1.8.0,<2.0.0
   and sqlalchemy-bigquery (1.8.0) requires Python >=3.7, <3.12, sqlalchemy-bigquery is forbidden.
  So, because test depends on sqlalchemy-bigquery (^1.8.0), version solving failed.

  at ~/.anyenv/envs/pyenv/versions/3.9.16/lib/python3.9/site-packages/poetry/mixology/version_solver.py:416 in _resolve_conflict
      412│             )
      413│             self._log(f'! which is caused by "{most_recent_satisfier.cause}"')
      414│             self._log(f"! thus: {incompatibility}")
      415│
    → 416│         raise SolveFailure(incompatibility)
      417│
      418│     def _choose_package_version(self) -> str | None:
      419│         """
      420│         Tries to select a version of a required package.

  Stack trace:

  11  ~/.anyenv/envs/pyenv/versions/3.9.16/lib/python3.9/site-packages/cleo/application.py:327 in run
       325│
       326│             try:
     → 327│                 exit_code = self._run(io)
       328│             except BrokenPipeError:
       329│                 # If we are piped to another process, it may close early and send a

  10  ~/.anyenv/envs/pyenv/versions/3.9.16/lib/python3.9/site-packages/poetry/console/application.py:190 in _run
       188│         self._load_plugins(io)
       189│
     → 190│         exit_code: int = super()._run(io)
       191│         return exit_code
       192│

   9  ~/.anyenv/envs/pyenv/versions/3.9.16/lib/python3.9/site-packages/cleo/application.py:431 in _run
       429│             io.input.interactive(interactive)
       430│
     → 431│         exit_code = self._run_command(command, io)
       432│         self._running_command = None
       433│

   8  ~/.anyenv/envs/pyenv/versions/3.9.16/lib/python3.9/site-packages/cleo/application.py:473 in _run_command
       471│
       472│         if error is not None:
     → 473│             raise error
       474│
       475│         return terminate_event.exit_code

   7  ~/.anyenv/envs/pyenv/versions/3.9.16/lib/python3.9/site-packages/cleo/application.py:457 in _run_command
       455│
       456│             if command_event.command_should_run():
     → 457│                 exit_code = command.run(io)
       458│             else:
       459│                 exit_code = ConsoleCommandEvent.RETURN_CODE_DISABLED

   6  ~/.anyenv/envs/pyenv/versions/3.9.16/lib/python3.9/site-packages/cleo/commands/base_command.py:119 in run
       117│         io.input.validate()
       118│
     → 119│         status_code = self.execute(io)
       120│
       121│         if status_code is None:

   5  ~/.anyenv/envs/pyenv/versions/3.9.16/lib/python3.9/site-packages/cleo/commands/command.py:62 in execute
        60│
        61│         try:
     →  62│             return self.handle()
        63│         except KeyboardInterrupt:
        64│             return 1

   4  ~/.anyenv/envs/pyenv/versions/3.9.16/lib/python3.9/site-packages/poetry/console/commands/add.py:267 in handle
       265│         self.installer.whitelist([r["name"] for r in requirements])
       266│
     → 267│         status = self.installer.run()
       268│
       269│         if status == 0 and not self.option("dry-run"):

   3  ~/.anyenv/envs/pyenv/versions/3.9.16/lib/python3.9/site-packages/poetry/installation/installer.py:104 in run
       102│             self.verbose(True)
       103│
     → 104│         return self._do_install()
       105│
       106│     def dry_run(self, dry_run: bool = True) -> Installer:

   2  ~/.anyenv/envs/pyenv/versions/3.9.16/lib/python3.9/site-packages/poetry/installation/installer.py:241 in _do_install
       239│                 source_root=self._env.path.joinpath("src")
       240│             ):
     → 241│                 ops = solver.solve(use_latest=self._whitelist).calculate_operations()
       242│         else:
       243│             self._io.write_line("Installing dependencies from lock file")

   1  ~/.anyenv/envs/pyenv/versions/3.9.16/lib/python3.9/site-packages/poetry/puzzle/solver.py:72 in solve
        70│         with self._progress(), self._provider.use_latest_for(use_latest or []):
        71│             start = time.time()
     →  72│             packages, depths = self._solve()
        73│             end = time.time()
        74│

  SolverProblemError

  The current project's Python requirement (>=3.9,<4.0) is not compatible with some of the required packages Python requirement:
    - sqlalchemy-bigquery requires Python >=3.7, <3.12, so it will not be satisfied for Python >=3.12,<4.0

  Because no versions of sqlalchemy-bigquery match >1.8.0,<2.0.0
   and sqlalchemy-bigquery (1.8.0) requires Python >=3.7, <3.12, sqlalchemy-bigquery is forbidden.
  So, because test depends on sqlalchemy-bigquery (^1.8.0), version solving failed.

  at ~/.anyenv/envs/pyenv/versions/3.9.16/lib/python3.9/site-packages/poetry/puzzle/solver.py:161 in _solve
      157│             packages = result.packages
      158│         except OverrideNeeded as e:
      159│             return self._solve_in_compatibility_mode(e.overrides)
      160│         except SolveFailure as e:
    → 161│             raise SolverProblemError(e)
      162│
      163│         combined_nodes = depth_first_search(PackageNode(self._package, packages))
      164│         results = dict(aggregate_package_nodes(nodes) for nodes in combined_nodes)
      165│

  • Check your dependencies Python requirement: The Python requirement can be specified via the `python` or `markers` properties

    For sqlalchemy-bigquery, a possible solution would be to set the `python` property to ">=3.9,<3.12"

    https://python-poetry.org/docs/dependency-specification/#python-restricted-dependencies,
    https://python-poetry.org/docs/dependency-specification/#using-environment-markers
```
</details>

If using venv directly, I can install sqlalchemy-bigquery.

<details>
<summary>log</summary>

```
$ python -m venv .venv
$ . .venv/bin/activate
$ pip install "sqlalchemy-bigquery"
Collecting sqlalchemy-bigquery
  Using cached sqlalchemy_bigquery-1.8.0-py2.py3-none-any.whl (33 kB)
Collecting packaging
  Using cached packaging-23.2-py3-none-any.whl (53 kB)
Collecting google-api-core!=2.0.*,!=2.1.*,!=2.2.*,!=2.3.0,<3.0.0dev,>=1.31.5
  Downloading google_api_core-2.12.0-py3-none-any.whl (121 kB)
     ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 121.4/121.4 KB 7.9 MB/s eta 0:00:00
Collecting sqlalchemy<2.0.0dev,>=1.2.0
  Using cached SQLAlchemy-1.4.49.tar.gz (8.5 MB)
  Preparing metadata (setup.py) ... done
Collecting google-auth<3.0.0dev,>=1.25.0
  Downloading google_auth-2.23.2-py2.py3-none-any.whl (181 kB)
     ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 182.0/182.0 KB 15.0 MB/s eta 0:00:00
Collecting google-cloud-bigquery<4.0.0dev,>=2.25.2
  Downloading google_cloud_bigquery-3.12.0-py2.py3-none-any.whl (220 kB)
     ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 220.4/220.4 KB 17.4 MB/s eta 0:00:00
Collecting protobuf!=3.20.0,!=3.20.1,!=4.21.0,!=4.21.1,!=4.21.2,!=4.21.3,!=4.21.4,!=4.21.5,<5.0.0.dev0,>=3.19.5
  Downloading protobuf-4.24.4-cp37-abi3-macosx_10_9_universal2.whl (409 kB)
     ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 409.4/409.4 KB 19.0 MB/s eta 0:00:00
Collecting requests<3.0.0.dev0,>=2.18.0
  Using cached requests-2.31.0-py3-none-any.whl (62 kB)
Collecting googleapis-common-protos<2.0.dev0,>=1.56.2
  Downloading googleapis_common_protos-1.60.0-py2.py3-none-any.whl (227 kB)
     ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 227.6/227.6 KB 10.2 MB/s eta 0:00:00
Collecting cachetools<6.0,>=2.0.0
  Downloading cachetools-5.3.1-py3-none-any.whl (9.3 kB)
Collecting pyasn1-modules>=0.2.1
  Using cached pyasn1_modules-0.3.0-py2.py3-none-any.whl (181 kB)
Collecting rsa<5,>=3.1.4
  Using cached rsa-4.9-py3-none-any.whl (34 kB)
Collecting grpcio<2.0dev,>=1.47.0
  Downloading grpcio-1.59.0-cp39-cp39-macosx_10_10_universal2.whl (9.6 MB)
     ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 9.6/9.6 MB 25.5 MB/s eta 0:00:00
Collecting google-cloud-core<3.0.0dev,>=1.6.0
  Downloading google_cloud_core-2.3.3-py2.py3-none-any.whl (29 kB)
Collecting proto-plus<2.0.0dev,>=1.15.0
  Downloading proto_plus-1.22.3-py3-none-any.whl (48 kB)
     ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 48.1/48.1 KB 6.4 MB/s eta 0:00:00
Collecting google-resumable-media<3.0dev,>=0.6.0
  Downloading google_resumable_media-2.6.0-py2.py3-none-any.whl (80 kB)
     ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 80.3/80.3 KB 12.8 MB/s eta 0:00:00
Collecting python-dateutil<3.0dev,>=2.7.2
  Using cached python_dateutil-2.8.2-py2.py3-none-any.whl (247 kB)
Collecting grpcio-status<2.0.dev0,>=1.33.2
  Downloading grpcio_status-1.59.0-py3-none-any.whl (14 kB)
Collecting google-crc32c<2.0dev,>=1.0
  Using cached google_crc32c-1.5.0-cp39-cp39-macosx_10_9_universal2.whl (32 kB)
Collecting pyasn1<0.6.0,>=0.4.6
  Using cached pyasn1-0.5.0-py2.py3-none-any.whl (83 kB)
Collecting six>=1.5
  Using cached six-1.16.0-py2.py3-none-any.whl (11 kB)
Collecting idna<4,>=2.5
  Using cached idna-3.4-py3-none-any.whl (61 kB)
Collecting urllib3<3,>=1.21.1
  Downloading urllib3-2.0.6-py3-none-any.whl (123 kB)
     ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 123.8/123.8 KB 20.5 MB/s eta 0:00:00
Collecting charset-normalizer<4,>=2
  Downloading charset_normalizer-3.3.0-cp39-cp39-macosx_11_0_arm64.whl (117 kB)
     ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 117.8/117.8 KB 15.7 MB/s eta 0:00:00
Collecting certifi>=2017.4.17
  Using cached certifi-2023.7.22-py3-none-any.whl (158 kB)
Using legacy 'setup.py install' for sqlalchemy, since package 'wheel' is not installed.
Installing collected packages: urllib3, sqlalchemy, six, pyasn1, protobuf, packaging, idna, grpcio, google-crc32c, charset-normalizer, certifi, cachetools, rsa, requests, python-dateutil, pyasn1-modules, proto-plus, googleapis-common-protos, google-resumable-media, grpcio-status, google-auth, google-api-core, google-cloud-core, google-cloud-bigquery, sqlalchemy-bigquery
  Running setup.py install for sqlalchemy ... done
Successfully installed cachetools-5.3.1 certifi-2023.7.22 charset-normalizer-3.3.0 google-api-core-2.12.0 google-auth-2.23.2 google-cloud-bigquery-3.12.0 google-cloud-core-2.3.3 google-crc32c-1.5.0 google-resumable-media-2.6.0 googleapis-common-protos-1.60.0 grpcio-1.59.0 grpcio-status-1.59.0 idna-3.4 packaging-23.2 proto-plus-1.22.3 protobuf-4.24.4 pyasn1-0.5.0 pyasn1-modules-0.3.0 python-dateutil-2.8.2 requests-2.31.0 rsa-4.9 six-1.16.0 sqlalchemy-1.4.49 sqlalchemy-bigquery-1.8.0 urllib3-2.0.6
WARNING: You are using pip version 22.0.4; however, version 23.2.1 is available.
You should consider upgrading via the '/Users/murakami_hikaru/Workspace/adhoc/2023-10-06/.venv/bin/python -m pip install --upgrade pip' command.

```
</details>

### LINKED ISSUE #8614: Package source constraints are silently ignored when extras and transitive dependencies are involved
- **Poetry version**: 1.6.1
- **Python version**: 3.11.4
- **OS version and name**: Ubuntu 23.04
- **pyproject.toml**: https://gist.github.com/latk/958e918b2ce4d96956a901263d38e955#file-pyproject-toml

## Issue

We use Poetry to manage multiple private packages, and have recently started migrating to setting our sources to `priority = "explicit"` as encouraged by the Poetry docs. This has been working fine, until Poetry **silently** stopped installing one of our internal dependencies. I have not been able to pinpoint the issue in Poetry's puzzle/solver, but it seems this bug depends on three factors:

* using sources with `priority = "explicit"` (but otherwise leaving the default PyPI source)
* having transitive dependencies to a package that only exists in the private source
* having these dependencies differ in required "extras"

I don't know of a way to quickly spin up a private package repository so I wasn't able to create a demo/test case, but I will walk us through all relevant details of the scenario:

We have three packages:

* `private-app` depends on `private-other` and `private-lib[extra]`
* `private-other` depends on `private-lib` (no extras!)
* `private-lib` is a library with extras. This is the library that goes missing during the solving process.

The `private-app` pyproject.toml has the following dependency specification:

```toml
[[tool.poetry.source]]
name = "myrepo"
url = "https://myrepo.example/"
priority = "explicit"

[tool.poetry.dependencies]
python = "~3.11"
private-lib = { version = "^5", source = "myrepo", extras = ["extra"] }
private-other = { version = "^1.6.3", source = "myrepo" }
```

This worked fine while the source was configured as `secondary = true`, but setting it to `priority = "explicit"` and adding the `source` constraints on the deps started creating breakage.

While trying to debug this without having created a venv with `poetry install`, Poetry instead reported an error that no matching version for `private-lib` could be found. Unfortunately I can no longer repro because we have already mitigated the problem, and this seems depends on the unavailability of a suitable version in the private package sources? Sorry.

When running `poetry update` or `poetry lock` on an *existing* poetry-installed venv, things would "succeed", but create a `poetry.lock` that contained the following entry for the private-lib:

```toml
[[package]]
name = "private-lib"
version = "5.17.0"
description = ""
optional = false
python-versions = "*"
files = []
```

Nothing other than the name has been redacted here. **There is no `source`, and `files` are empty.** So Poetry knows about the dependency, but finds no installation candidates, and thus silently skips installation. This then broke our CI systems when trying to install dependencies from this lockfile.

I have captured a log for this session `poetry -vvv update`, minus all PyPI-based dependencies as they distract from the bug: https://gist.github.com/latk/958e918b2ce4d96956a901263d38e955#file-log-txt

Here is an excerpt of that already-condensed log highlighting how the private-lib is resolved:

```
   1: fact: private-app depends on private-lib (^5)
   1: derived: private-lib[extra] (>=5,<6)
Source (myrepo): 31 packages found for private-lib >=5,<6
   1: fact: private-other1 (1.6.3) depends on private-lib (*)
   1: selecting private-other1 (1.6.3)
   1: derived: private-lib
Source (PyPI): No packages found for private-lib
Source (PyPI): 0 packages found for private-lib *
Falling back to installed packages to discover metadata for private-lib
Found 1 compatible packages for private-lib
Source (PyPI): Getting info for private-lib (5.17.0) from PyPI
Falling back to installed packages to discover metadata for private-lib
Found 1 compatible packages for private-lib
   1: selecting private-lib (5.17.0)
   1: fact: private-lib (5.17.0) depends on private-lib (5.17.0)
   1: selecting private-lib[extra] (5.17.0)

Finding the necessary packages for the current system
Source (myrepo): 1 packages found for private-other1 >=1.6.3,<2.0.0
Source (myrepo): 1 packages found for private-other2 >=2.1.1,<3.0.0
Source (myrepo): 1 packages found for private-lib >=5,<6
Source (PyPI): 1 packages found for private-lib *
Source (myrepo): 1 packages found for private-lib *

Writing lock file
```

What we are seeing here is that when the `private-other1` dependencies are resolved, the `private-lib` is looked up in PyPI rather than through its `explicit` source. Nothing is found there. Thus, Poetry falls back to looking at locally installed packages. Since the venv already exists, it is found, albeit without any installation candidates (since it's already installed). At the end, one `private-lib` package was found each for `myrepo` (correct) and `PyPI` (incorrect, this is the installed package). When writing the lockfile, it seems that the PyPI/local info is used, thus leading to the empty `files = []` array.

Regardless of the problematic poetry.lock file being generated, this violates the expectations about top-level dependencies with a `source` option: if I specify a `source`, then I'm expecting that metadata about that package will only ever be retrieved from that source. That this logfile shows `PyPI` being contacted is **arguably a security issue** (but not actively exploitable, so whatever…).

## Attempted workarounds

I have found some techniques to mitigate the observed problems, which might aid with debugging.

* Explicitly updating that dependency via `poetry update private-lib`. This leads to a working poetry.lock file, but things will break again with the next `poetry update`.

* Setting the private repository to `priority = "supplemental"`. This "works" because when the PyPI lookup for `private-lib` fails, `myrepo` will be checked next, finding the correct package data. This avoids the problematic fallback to the installed version of the package. However, this would only mask the bug, and despite the explicit `source = "myrepo"` would not protect against dependency confusion attacks!

* Stop using the `extras` in Python packages. This is what we are doing, at least until this Poetry bug is fixed.

* Include the top-level dependency on the `private-lib` both with and without extras. I believe (see discussion below) that this primes Poetry's dependency solver cache to avoid the PyPI lookup.

  ```toml
  [tool.poetry.dependencies]
  python = "~3.11"
  private-lib = [
    { version = "^5", source = "myrepo" },
    { version = "^5", source = "myrepo", extras = ["extra"] },
  ]
  ```

  This doesn't seem to depend on the order of entries.

## Interpretation and preliminary analysis

I've been poking a bit at the Poetry source code and now have a basic understanding of Poetry's dependency solver. I think the following is happening, but am not entirely sure:

When a dependency has extras, a virtual dependency is created which inherits the correct source:

https://github.com/python-poetry/poetry/blob/a029b367a88abafb1fda4349b0f1e763df61f583/src/poetry/puzzle/provider.py#L532-L536

Thus, we actually have the following facts in our dependency graph:

1. `private-app` depends on `private-other` with `source = "myrepo"`
2. `private-app` depends on `private-lib[extra]` with `source = "myrepo"`
3. `private-other` depends on `private-lib`
4. `private-lib[extra]` depends on `private-lib` with `source = "myrepo"`

The 3rd fact cannot have a `source`, because sources are purely a Poetry feature, and not part of Python package metadata.

Instead, the source for a transitive dependency can be set by specifying it as a top-level dependency in the `pyproject.toml`. Later, dependency resolutions are resolved by name in the dependency cache. I think that happens here:

https://github.com/python-poetry/poetry/blob/a029b367a88abafb1fda4349b0f1e763df61f583/src/poetry/mixology/version_solver.py#L486-L488

https://github.com/python-poetry/poetry/blob/a029b367a88abafb1fda4349b0f1e763df61f583/src/poetry/mixology/version_solver.py#L96-L114

Since the `complete_name` is used as the key, `private-lib[extra]` has a different cache entry from `private-lib`, but that is of course necessary.

What is not obvious is that the contents of the cache are sensitive to the order in which dependencies are resolved:

* If the above fact 3 is resolved before fact 4, we will see the problem explained in this issue, as the `private-lib` gets wrongly associated with the PyPI source (and this is inherited while resolving fact 4).
* If fact 4 were to be resolved first, then everything would work and fact 3 would inherit the correct source. This is what I'm achieving in one of the workarounds by specifying both `private-lib` and `private-lib[extra]` in the Poetry dependencies.

I'm not sure how this could be solved. Maybe:

* when a virtual `dep[extras] -> dep` dependency is created, ensure that `dep` is resolved *immediately* to prime the cache with the correct source. However, this seems fragile, and wouldn't be a systematic solution to this kind of problem.

* keep an explicit name -> source mapping, which is used to look up the appropriate source for transitive dependencies. However, this would fail for configurations with constraints like:

  ```toml
  some-dep = [
    { ..., source = "repo-a", python = ">=3.6,<3.8" },
    { ..., source = "repo-b", python = ">=3.8" },
  ]
  ```

* cache package data at the correct place. Currently, information about dependencies is cached in two places during resolution. First, the aforementioned `DependencyCache`. This will then eventually load data from the `RepositoryPool`, which will load data from the correct repository (in particular, the explicit `myrepo` if the source name is set to such):

  https://github.com/python-poetry/poetry/blob/a029b367a88abafb1fda4349b0f1e763df61f583/src/poetry/repositories/repository_pool.py#L209-L219

  Each repository in the pool might have its own cache (though this doesn't seem to prevent the repeated failing PyPI lookups visible in the logfile…).

  It seems to me that this problem might disappear if there were a cache *around* the RepositoryPool? After all, the repositories do not care about `extras`, so a top-level dependency on `private-lib[extra]` with `source = "myrepo"` would prime a repository pool cache for any subsequent `private-lib` lookup.

lucemia:
# Pull Request Check List

Resolves: #8614 
- try to solve the #8614
- Introduce a `dependency_source_mapping` parameter for the `RepositoryPool` class, which is derived directly from the `pyproject.toml` file.

### DIFF
--- a/src/poetry/poetry.py
+++ b/src/poetry/poetry.py
@@ -41,7 +41,18 @@ class Poetry(BasePoetry):
 
         self._locker = locker
         self._config = config
-        self._pool = RepositoryPool(config=config)
+
+        local_config = local_config or {}
+        dependency_source_cache = {}
+
+        for group in [*local_config.get("group", {}).values(), local_config]:
+            for name, dependency in group.get("dependencies", {}).items():
+                if isinstance(dependency, dict) and "source" in dependency:
+                    dependency_source_cache[name] = dependency["source"]
+
+        self._pool = RepositoryPool(
+            config=config, dependency_source_mapping=dependency_source_cache
+        )
         self._plugin_manager: PluginManager | None = None
         self._disable_cache = disable_cache
 
--- a/src/poetry/repositories/repository_pool.py
+++ b/src/poetry/repositories/repository_pool.py
@@ -45,6 +45,7 @@ class RepositoryPool(AbstractRepository):
         repositories: list[Repository] | None = None,
         ignore_repository_names: object = _SENTINEL,
         *,
+        dependency_source_mapping: dict[str, str] | None = None,
         config: Config | None = None,
     ) -> None:
         super().__init__("poetry-repository-pool")
@@ -57,6 +58,10 @@ class RepositoryPool(AbstractRepository):
 
         self._artifact_cache = ArtifactCache(
             cache_dir=(config or Config.create()).artifacts_cache_directory
+        )
+
+        self._dependency_source_mapping: dict[str, str] = (
+            dependency_source_mapping or {}
         )
 
         if ignore_repository_names is not _SENTINEL:
@@ -210,6 +215,11 @@ class RepositoryPool(AbstractRepository):
         repository_name = dependency.source_name
         if repository_name:
             return self.repository(repository_name).find_packages(dependency)
+
+        if dependency.name in self._dependency_source_mapping:
+            return self.repository(
+                self._dependency_source_mapping[dependency.name]
+            ).find_packages(dependency)
 
         packages: list[Package] = []
         for repo in self.repositories:
--- a/tests/repositories/test_repository_pool.py
+++ b/tests/repositories/test_repository_pool.py
@@ -333,6 +333,23 @@ def test_pool_find_packages_in_specified_repository() -> None:
     assert returned_packages_unavailable == []
 
 
+def test_pool_find_packages_with_dependency_source_mapping() -> None:
+    package_foo1 = get_package("foo1", "1.1.1")
+    package_foo2 = get_package("foo2", "1.1.1")
+
+    bar = Repository("bar", [package_foo1, package_foo2])
+    pool = RepositoryPool(dependency_source_mapping={"foo1": "bar"})
+    pool.add_repository(bar, priority=Priority.EXPLICIT)
+
+    available_dependency = get_dependency("foo1", "^1.0.0")
+    returned_packages_available = pool.find_packages(available_dependency)
+    unavailable_dependency = get_dependency("foo2", "^1.0.0")
+    returned_unavailable_dependency = pool.find_packages(unavailable_dependency)
+
+    assert returned_packages_available == [package_foo1]
+    assert returned_unavailable_dependency == []
+
+
 def test_search_no_legacy_repositories() -> None:
     package_foo1 = get_package("foo", "1.0.0")
     package_foo2 = get_package("foo", "2.0.0")


latk:
Thank you so much for trying to fix this! It seems that this proposed code is evaluated quite early in the Poetry lifecycle, so that you're working with the raw TOML structure:

```python
        for group in [*local_config.get("group", {}).values(), local_config]:
            for name, dependency in group.get("dependencies", {}).items():
                if isinstance(dependency, dict) and "source" in dependency:
                    dependency_source_cache[name] = dependency["source"]
```

This is good, because it runs before packages can be looked up.

But I'm not sure whether it supports all documented dependency specification forms, in particular [multiple constraint dependencies](https://python-poetry.org/docs/dependency-specification/#multiple-constraints-dependencies) where the entry might be a list, not a dict.

The docs give this example:

```toml
[tool.poetry.dependencies]
foo = [
    {version = "<=1.9", python = ">=3.6,<3.8"},
    {version = "^2.0", python = ">=3.8"}
]
```

In #8614, I noted that each candidate's source might differ. For example:

```toml
some-dep = [
  { ..., source = "repo-a", python = ">=3.6,<3.8" },
  { ..., source = "repo-b", python = ">=3.8" },
]
```

I'm not sure whether Poetry actually supports this.

* If yes, I don't think the name–source mapping suggested in this PR would work. We'd need a better "key" than the dependency name.
* If it isn't supported, this mapping should at least check that there's no contradictory source information. For security-relevant features like `source`, I think a hard error is preferable to silently proceeding with an unclear choice.

lucemia:
Thank you for your feedback. Yes, I did forget about handling multiple constraint dependencies, I will add support for them.

As you pointed out, the current pull request doesn't completely resolve the issue if we don't ensure the consistency of source information. I'm also addressing the related problem of "Package's source inconsistency" described in https://github.com/python-poetry/poetry/issues/8704. However, I still don't have a clear idea of how to develop it correctly. Did poetry already handle this issue somewhere? I don't have an idea yet.

The primary goal of the pull request is to provide a partial solution that is at least on par with the current situation. This is because the mapping comes into play only when the repository_name is not provided.

radoering:
I'm not sure if this is the right path, but since I have no other solution in mind, I'll give some feedback:

> I'm not sure whether Poetry actually supports this.

It does. `name` + `markers` (including `python` and `platform`) should be sufficient. (IIRC, `platform` and `python` is integrated in `markers` internally.) However, markers can't be used as a simple key because the lookup might not be with the original marker but a subset of it. Further, we don't pass markers to the `RepositoryPool` at the moment, so this might not be the best place for the lookup...

> working with the raw TOML structure

It's probably better to use the parsed dependencies in `package.requires` because `python` and `platform` have already been merged with `markers` and multiple-constraints dependencies should be parsed at that point.

radoering:
After thinking about it again, I believe that this issue should be addressed in the solver instead of the repository pool. The repository pool is not aware of markers (and probably should not be) so that it is not possible to handle multiple constraints with different sources.

I gave it a try in #8835 and I am not aware of any drawbacks yet.

Issue #8512, opened 2023-10-06
poetry-export pre-commit hook doesn't fail the commit/push


mil-ad:
The poetry-expory hook is quite useful but shouldn't it return an error when it actually detects any changes? I want to use it alongside poetry-lock hook which does in fact fails after updating the lock file, and forces you to commit the changed lock file. I basically want to make sure that the poetry.lock and requirements.txt files are always in sync when commits are pushed. Right now the poetry-export does correctly update requirements.txt but doesn't block the push.

xqm32:
Hi @mil-ad, here's a workaround for you: https://github.com/python-poetry/poetry-plugin-export/issues/237#issuecomment-1769923575.

noamraph:
I think that actually there is no bug. It turns out that pre-commit checks if the content was changed, and doesn't require a nonzero return code. The check is only done on committed files - this is why I was confused to think that it generally ignores modifications done by the hook. But from what I see, after committing the file, everything works as expected.

See: https://github.com/pre-commit/pre-commit.com/pull/912#issuecomment-1864668427

PR #8810, opened 2023-12-19
.pre-commit-hooks.yaml: Add --check-change to poetry-export

noamraph:
# Pull Request Check List

Resolves: #8512 , https://github.com/python-poetry/poetry-plugin-export/issues/237

Depends on: https://github.com/python-poetry/poetry-plugin-export/pull/252

https://github.com/python-poetry/poetry-plugin-export/pull/252 adds a --check-change option to `poetry export`. This change adds this argument to the pre-commit hook, in order to make the hook behave as expected and return nonzero if it changed files.

### DIFF
--- a/.pre-commit-hooks.yaml
+++ b/.pre-commit-hooks.yaml
@@ -21,7 +21,7 @@
   language: python
   pass_filenames: false
   files: ^(.*/)?poetry\.lock$
-  args: ["-f", "requirements.txt", "-o", "requirements.txt"]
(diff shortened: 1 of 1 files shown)


noamraph:
I think that this change isn't needed - it turns out that pre-commit checks if the content was changed, and doesn't require a nonzero return code.

See: https://github.com/pre-commit/pre-commit.com/pull/912#issuecomment-1864668427

Issue #8811, opened 2023-12-20
How to use different Versions of poetry?


MaKaNu:
- [x] I have searched the [issues](https://github.com/python-poetry/poetry/issues) of this repo and believe that this is not a duplicate.

## Issue
I am usually on the edge and utilize poetry 1.7.1 at the moment.

Recently I needed to lower my support for a package down to 3.6.9. I actually don't want to, but some old ubuntu 2018.04 forced me.

By doing so I realized 1.7.1 does not support 3.6.9. So I installed 1.2.0a2, the most recent Version available for 3.6.9 it seems.

So for the moment the newest release is not available for me. I use pyenv to control the python versions I have in use.

What is the best practice to have access to both versions?

dimbleby:
Best practice is to read the docs, which discuss this.

Please close.

MaKaNu:
@dimbleby 
Are you talking about Managing environments?

If yes this was not a solution for my problem.

If I have installed python3.6 via pyenv and selected 3.6 I get an Error for the venv creation:

```bash
SyntaxError: future feature annotations is not defined
```
which is reasoanble since this future is only available from 3.7

OR is the solution `pipx`? I only used the official installer in the past.

radoering:
I use pipx with suffixes as described in https://python-poetry.org/docs/#installing-with-pipx

Suprithvarma1:
by using pipx 
$ pipx install --suffix "@1.4" poetry==1.4.2
$ pipx install --suffix "@1.3" poetry==1.3.5

MaKaNu:
Okay this explains my issues I had. The main reason I and I think many others have is the installation of a lower version if the latest is no longer supported by an older python version.

Is it somehow reasonable to control the suffix with pyenv? So as example, if I set the version of python to 3.6 in a repo dir poetry is binded to poetry version suffix I define?

MaKaNu:
Additionally I checked that auto-completion in this case only work for the none suffix version

dimbleby:
do whatever you like, if something around pyenv works for you then go for it.  The poetry docs provide a solution already.

this really isn't a poetry question at all, you would have the same question for any python tool where you wanted to have more than one version available: poetry is only an example of that.

MaKaNu:
I agree. I would describe my issue from what I've learned in this discussion as following:

without pipx experience and usage of official installer (which I used and I think many first time users also) it is not very clear in the documentation that pipx is the solution to go for multiple versions of poetry.

[...]

PR #8814, opened 2023-12-21
[Docs]: Add note for parallel installation

MaKaNu:
- add note for installation advice

# Pull Request Check List

Resolves: #8811

### DIFF
--- a/docs/_index.md
+++ b/docs/_index.md
@@ -175,7 +175,7 @@ You can also install Poetry from a `git` repository by using the `--git` option:
 ```bash
 curl -sSL https://install.python-poetry.org | python3 - --git https://github.com/python-poetry/poetry.git@master
 ````
-If you want to install different versions of poetry in parallel, a good approach is the installation with pipx and suffix.
+If you want to install different versions of Poetry in parallel, a good approach is the installation with pipx and suffix.
 
 {{< /step >}}
 {{< step >}}
--- a/docs/_index.md
+++ b/docs/_index.md
@@ -42,6 +42,10 @@ See the **advanced** installation instructions to use a preview or alternate ver
 [`pipx`](https://github.com/pypa/pipx) is used to install Python CLI applications globally while still isolating them in virtual environments.
 `pipx` will manage upgrades and uninstalls when used to install Poetry.
 
+{{% note %}}
+pipx is the installation approach which the user should choose to install multiple versions of pipx. 
+{{% /note %}}
+
 {{< steps >}}
 {{< step >}}
 **Install Poetry**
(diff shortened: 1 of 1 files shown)


docs/_index.md
```diff
@@ -42,6 +42,10 @@ See the **advanced** installation instructions to use a preview or alternate ver
 [`pipx`](https://github.com/pypa/pipx) is used to install Python CLI applications globally while still isolating them in virtual environments.
 `pipx` will manage upgrades and uninstalls when used to install Poetry.
 
+{{% note %}}
+pipx is the installation approach which the user should choose to install multiple versions of pipx.
```
dimbleby, line 46: even after you fix the brain fart, it is not clear to me that this is an improvement.  

This question is on your mind and so you think that the docs should emphasise it; but 99.9% of readers do not want to install multiple versions of poetry, and the remaining 0.1% can read how to do so in the very next section anyway.

ie the danger here is that this makes the docs better specifically for you - but a little worse for almost everyone.

MaKaNu: You refer to 99.9% of user which use pipx as the main installation method? If so I agree. So a better place would be inside the official installer section as a specific subsection?

I can only imagine, but for many users which used like me the official install and reached my point, it might be unclear that pipx is the solution.

dimbleby: I refer to the 99.9% of users who don't want to install more than one version of poetry.

pipx is already documented as the primary installation method, and that section already clearly describes its use to install more than one version of poetry.  

No matter what the docs say, some people just won't read them - so at some point you hit diminishing returns:
- people who didn't read the docs didn't read the bit that you've added either
- people who did read the docs already had the answer

Anyway I don't hold the commit bit, I'll leave it to someone who does to take a view on whether this is worthwhile

MaKaNu: Thanks for clarification.

So for the 0.1% of the users (me included) who does not install with pipx in the first place and used the official installer and red the documentation only for the official installer, the hint in advanced section might be helpful.

Speaking for myself this sentence would have been useful for me.

docs/_index.md
```diff
@@ -175,6 +175,8 @@ You can also install Poetry from a `git` repository by using the `--git` option:
 ```bash
 curl -sSL https://install.python-poetry.org | python3 - --git https://github.com/python-poetry/poetry.git@master
 ````
+If you want to install different versions of poetry in parallel, a good approach is the installation with pipx and suffix.
```
radoering, line 178:
```suggestion
If you want to install different versions of Poetry in parallel, a good approach is the installation with pipx and suffix.
```

docs/_index.md
radoering: IMO, the note makes sense at the updated location - at least for users who are too fixated on the official installer.

radoering (approved):

PR #8815, opened 2023-12-21
feature: add new "lazy-wheel" config option

radoering:
# Pull Request Check List

_With #5509, this is only relevant for indexes that do not support PEP 658 - thus, probably for most indexes except for PyPI - or old wheels without PEP 658 metadata._

This PR adds a config option `solver.lazy-wheel`. If active and a server supports [HTTP range requests](https://developer.mozilla.org/en-US/docs/Web/HTTP/Range_requests), we do not have to download entire wheels to fetch metadata but can just download the METADATA file from the remote wheels with a couple of range requests. Especially with slow network connections this setting can speed up dependency resolution significantly. (If the cache has already been filled or the server does not support HTTP range requests, this setting makes no difference.)

This PR is based on pypa/pip#12208, which I learned about at PackagingCon, so all credits to the authors of this pip PR. 👏 I just had to make the interface fit for Poetry, write some tests and fix some special cases I encountered.

I measured the time of `poetry lock` of a largish project, once from a machine with a slow network connection (1.5 MB/s, 40 ms ping) and once from a machine with a fast network connection (95 MB/s, 1 ms ping) to the package index.

|operation|slow network|fast network|
|---|---|---|
|cold cache without PR|825 s|75 s|
|cold cache with PR|190 s|60 s|
|warm cache without/with PR|24 s|16 s|

What `lazy-wheel` basically does is:

- try a range request with a negative offset in order to fetch the central directory of the wheel zip file
- if negative range requests are not supported, send a HEAD request to get the size of the wheel and afterwards send a normal range request to fetch the last bytes of the wheel
- if the central directory is quite large we may need an additional range request to fetch the rest of the central directory
- look for the position of the METADATA file in the central directory and download it in one range request

In order to not try range requests again and again in vain, we keep track if a domain supports range requests at all and especially with negative offsets.

In contrast to the pip PR, we only fetch the `METADATA` file instead of the entire `.dist-info` directory. On the machine with a slow network connection this made a difference of 190 s to 330 s.

Further, I added some handling for special cases I encountered while trying range requests with different servers:

|Server|Accept-Ranges|Negative Offset|Negative Offset > entire wheel|
|---|---|---|---|
|pypi.org|bytes|501 Unsupported client range|501 Unsupported client range|
|Internal devpi server 1|bytes<sup>1</sup>|206 Partial content|⚠️ 200 OK, entire wheel|
|Internal devpi server 2|-|200 OK, entire wheel|200 OK, entire wheel|
|Internal Artifactory|bytes|206 Partial content, ⚠️ data from start of file|206 Partial content, entire wheel|
|Internal GitLab|bytes|206 Partial content|206 Partial content, entire wheel|
|piwheels.org|bytes|206 Partial content|206 Partial content, entire wheel|
|download.pytorch.org|bytes|206 Partial content|206 Partial content, entire wheel|

<sup>1</sup> "Devpi server 1" hosts several indexes including a PyPI mirror. I noticed that range requests were supported for wheels that have been downloaded before, but were not supported for wheels that have not been downloaded before. In other words, a server might supports range requests for some but not all wheels. (This is handled in `HTTPRepository`.)

The behavior of the different servers can be interpreted as follows:

- The internal GitLab, piwheels.org and download.pytorch.org behave as you would expect from a server that supports range requests with negative offsets.
- PyPI behaves as you would expect from a server that supports range requests but not negative offsets.
- The internal devpi server 2 behaves as you would expect from a server that does not support range requests.
- The internal devpi server 1 has a quirk when the negative offset is greater than the entire wheel. It does return `200 OK` instead of `206 Partial content`, so you have to inspect the `Accept-Ranges` header to determine if it supports range requests or not.
- The internal Artifactory has the most unexpected behavior. It does not support negative offsets but interprets it as positive. E.g. `-100` is handled as `0-100`.

In case you were wondering at the beginning why we should introduce a config option and not just always use range requests if possible, it's probably clearer now: Although various servers were tested and special cases were implemented, it's not unlikely that there are still servers with unexpected behavior that is not handled well.


src/poetry/inspection/lazy_wheel.py
```diff
@@ -0,0 +1,762 @@
...
+
+    def __init__(self, location: str, name: str):
+        self.location = location
+        self.name = name
+
+    def __str__(self) -> str:
+        return f"Wheel '{self.name}' located at {self.location} is invalid."
+
+
+class MemoryWheel(Distribution):
```
Secrus, line 61: Could you use [`packaging.metadata`](https://packaging.pypa.io/en/stable/metadata.html) instead of `pkginfo`? I plan on dropping `pkginfo` in favor of `packaging` quite soon.

radoering: tldr: I don't think so.

Having just tried it out, I like that it simplifies the interface because we can get rid of `MemoryWheel`.

However, I firmly believe that **we should not partially switch from pkginfo to packaging.metadata**.

`pkginfo` just keeps the string values. This results in `version` being a string and `requires_dist` being a list of strings and so on. `packaging.metadata` does some additional parsing so `version` is a `Version` and `requires_dist` is a list of `Requirement` objects, `requires_python` is a `SpecifierSet`.

And here is the issue:
A `requries_python` like `">=2.7, !=3.0.*, !=3.1.*, !=3.2.*"` is passed as is by `pkginfo` but converted to `"!=3.0.*,!=3.1.*,!=3.2.*,>=2.7"` by `packaging.metadata`. (It's equivalent but not equal.)

If I only use `packaging.metadata` here, you will get different (though equivalent) lock files depending on whether you have activated `lazy-wheel` or not. Even worse: it will not depend on the current setting, but on the setting when the data was cached. So even `poetry lock --no-update` may result in different files on different machines if the setting was different when the metadata had been cached.

Thinking about this, **we probably should even increment the cache version, when switching from `pkginfo` to `packaging.metadata`**.

If you are interested, I put a commit switching to `packaging.metadata` on a separate branch: https://github.com/radoering/poetry/commit/f0bcee4809bf7b019f68c1399b9c81178ded1247

radoering: Rereading the docs, a solution might be to use `packaging.metadata.parse_email()` and use the resulting `RawMetadata` dict instead of the parsed `Metadata` from `packaging.metadata.from_email()`. That way, the transition might be transparent. I'll try this tomorrow.

radoering: `RawMetadata` seems to be the right way for us. The PR is void of `pkginfo` now.

src/poetry/inspection/lazy_wheel.py
```diff
@@ -0,0 +1,755 @@
...
+
+
+class UnsupportedWheel(Exception):
+    """Unsupported wheel."""
+
+
+class InvalidWheel(Exception):
+    """Invalid (e.g. corrupt) wheel."""
+
+    def __init__(self, location: str, name: str):
```
trag1c, line 54:
```suggestion
    def __init__(self, location: str, name: str) -> None:
```

src/poetry/inspection/lazy_wheel.py
```diff
@@ -0,0 +1,755 @@
...
+        self.location = location
+        self.name = name
+
+    def __str__(self) -> str:
+        return f"Wheel '{self.name}' located at {self.location} is invalid."
```
trag1c, line 59:
```suggestion
        return f"Wheel {self.name!r} located at {self.location} is invalid."
```

src/poetry/inspection/lazy_wheel.py
```diff
@@ -0,0 +1,755 @@
...
+
+    def writelines(self, lines: Iterable[Any]) -> None:
+        raise NotImplementedError
+
+    def __enter__(self) -> ReadOnlyIOWrapper:
+        self._file.__enter__()
+        return self
+
+    def __exit__(self, *exc: Any) -> None:
+        self._file.__exit__(*exc)
```
trag1c, line 244:
```suggestion
    def __exit__(
        self,
        exc_type: type[BaseException] | None,
        exc_value: BaseException | None,
        traceback: TracebackType | None,
    ) -> bool:
    	return self._file.__exit__(exc_type, exc_value, traceback)
```
(`TracebackType` requires importing from `types`)

same applies to L301

src/poetry/inspection/lazy_wheel.py
```diff
@@ -0,0 +1,755 @@
...
+    def _prefetch_metadata(self, name: str) -> str:
+        """Locate the *.dist-info/METADATA entry from a temporary ``ZipFile`` wrapper,
+        and download it.
+
+        This method assumes that the *.dist-info directory (containing e.g. METADATA) is
+        contained in a single contiguous section of the zip file in order to ensure it
+        can be downloaded in a single ranged GET request."""
+        logger.debug("begin prefetching METADATA for %s", name)
+
+        metadata_regex = re.compile(r"^[^/]*\.dist-info/METADATA$")
```
trag1c, line 714: Is there any reason not to compile this earlier and have this pattern as a constant?

tests/inspection/test_lazy_wheel.py
```diff
@@ -0,0 +1,324 @@
...
+        [HTTPrettyRequest, str, Dict[str, Any]], HttPrettyResponse
+    ]
+
+    class RequestCallbackFactory(Protocol):
+        def __call__(
+            self,
+            *,
+            accept_ranges: str | None = "bytes",
+            negative_offset_error: tuple[int, bytes] | None = None,
+        ) -> HttPrettyRequestCallback: ...
```
trag1c, line 42: For consistency with `httpretty` and compliance with PEP8 naming:
```suggestion
    HTTPrettyResponse = Tuple[int, Dict[str, Any], bytes]  # status code, headers, body
    HTTPrettyRequestCallback = Callable[
        [HTTPrettyRequest, str, Dict[str, Any]], HTTPrettyResponse
    ]

    class RequestCallbackFactory(Protocol):
        def __call__(
            self,
            *,
            accept_ranges: str | None = "bytes",
            negative_offset_error: tuple[int, bytes] | None = None,
        ) -> HTTPrettyRequestCallback: ...
```

tests/inspection/test_lazy_wheel.py
```diff
@@ -0,0 +1,324 @@
...
+            if start < 0:
+                # wheel is smaller than initial chunk size
+                return 200, response_headers, wheel_bytes
+            end = total_length - 1
+            body = wheel_bytes[offset:]
+    else:
+        # range with start and end
+        rng_parts = rng.split("-")
+        start = int(rng_parts[0])
+        end = int(rng_parts[1])
```
trag1c, line 86:
```suggestion
        start, end = map(int, rng_parts)
```
Or if `rng_parts` isn't guaranteed to be of length 2:
```suggestion
        start, end, *_ = map(int, rng_parts)
```

tests/inspection/test_lazy_wheel.py
```diff
@@ -0,0 +1,324 @@
...
+        negative_offset_error: tuple[int, bytes] | None = None,
+    ) -> HttPrettyRequestCallback:
+        def handle_request(
+            request: HTTPrettyRequest, uri: str, response_headers: dict[str, Any]
+        ) -> HttPrettyResponse:
+            name = Path(urlparse(uri).path).name
+
+            wheel = Path(__file__).parent.parent.joinpath(
+                "repositories/fixtures/pypi.org/dists/" + name
+            )
```
trag1c, line 106:
```suggestion
            wheel = Path(__file__).parents[1].joinpath(
                "repositories/fixtures/pypi.org/dists/" + name
            )
```

tests/inspection/test_lazy_wheel.py
```diff
@@ -0,0 +1,324 @@
...
+            del response_headers["status"]
+
+            if request.method == "HEAD":
+                return build_head_response(
+                    accept_ranges, len(wheel_bytes), response_headers
+                )
+
+            rng = request.headers.get("Range", "")
+            if rng:
+                rng = rng.split("=")[1]
```
trag1c, line 127:
```suggestion
            rng = request.headers.get("Range", "=").split("=")[1]
```

tests/inspection/test_lazy_wheel.py
```diff
@@ -0,0 +1,324 @@
...
+    assert len(latest_requests) == expected_requests
+
+
+@pytest.mark.parametrize("negative_offset_as_positive", [False, True])
+def test_metadata_from_wheel_url_smaller_than_initial_chunk_size(
+    http: type[httpretty.httpretty],
+    handle_request_factory: RequestCallbackFactory,
+    negative_offset_as_positive: bool,
+) -> None:
+    domain = f"tiny-wheel-{str(negative_offset_as_positive).lower()}.com"
```
trag1c, line 229:
```suggestion
    domain = f"tiny-wheel-{str(negative_offset_as_positive).casefold()}.com"
```

tests/repositories/test_http_repository.py
```diff
@@ -0,0 +1,150 @@
...
+        repo._supports_range_requests[domain] = supports_range_requests
+
+    info = repo._get_info_from_wheel(url)
+    assert info.name == "poetry-core"
+    assert info.version == "1.5.0"
+    assert info.requires_dist == [
+        'importlib-metadata (>=1.7.0) ; python_version < "3.8"'
+    ]
+
+    if lazy_wheel and (supports_range_requests or supports_range_requests is None):
```
trag1c, line 71:
```suggestion
    if lazy_wheel and supports_range_requests is not False:
```

trag1c (changes requested):

src/poetry/inspection/lazy_wheel.py
radoering: Good catch. I have always thought that mypy catches this but it looks like there is an exception for `__init__`. 😲

src/poetry/inspection/lazy_wheel.py
radoering: I just removed the apostrophes.

src/poetry/inspection/lazy_wheel.py
radoering: That does not work for me. mypy fails if I try this.

src/poetry/inspection/lazy_wheel.py
radoering: No. Changed it.

src/poetry/inspection/lazy_wheel.py
trag1c: This should work then:
```suggestion
    def __exit__(
        self,
        exc_type: type[BaseException] | None,
        exc_value: BaseException | None,
        traceback: TracebackType | None,
    ) -> None:
    	self._file.__exit__(exc_type, exc_value, traceback)
```
Seems like Python itself doesn't exactly follow the standard signature 😄

radoering: That works. 👍

src/poetry/console/commands/config.py
```diff
@@ -68,15 +68,16 @@ def unique_config_values(self) -> dict[str, tuple[Any, Any]]:
             ),
             "virtualenvs.path": (str, lambda val: str(Path(val))),
             "virtualenvs.prefer-active-python": (boolean_validator, boolean_normalizer),
+            "virtualenvs.prompt": (str, lambda val: str(val)),
```
trag1c, line 71: I realize you just moved the line but why is this `lambda val: str(val)` instead of `str` like the first element?

trag1c (approved):

src/poetry/console/commands/config.py
radoering: Probably, a copy & paste (& edit) of another line. Since I had already touched the line anyway and there's no risk in the change, I adjusted it straight away.

src/poetry/inspection/lazy_wheel.py
```diff
@@ -0,0 +1,772 @@
...
+        Further, caching partial requests might cause issues:
+        https://github.com/pypa/pip/pull/8716
+        """
+        # "no-cache" is the correct value for "up to date every time", so this will also
+        # ensure we get the most recent value from the server:
+        # https://developer.mozilla.org/en-US/docs/Web/HTTP/Caching#provide_up-to-date_content_every_time
+        return {"Accept-Encoding": "identity", "Cache-Control": "no-cache"}
+
+
+class LazyWheelOverHTTP(LazyHTTPFile):
```
Secrus, line 522: Do we need all those inheritance levels? From what I have seen, we only use `LazyWheelOverHTTP`. Maybe you could squash some classes together so we can avoid unnecessary code?

radoering: Good point. I thought the same at first, but then I got used to it because the scope of the classes is clearly separated. Rethinking it, you are probably right that we will never need that many classes. I squashed some of the classes so we now have 3 classes instead of 5. I think that's a good compromise between the number of classes and the length of each class.

I added the changes in separate commits so it's easier to proceed with the review.

Secrus (approved):
LGTM, one small nitpick left.

src/poetry/inspection/lazy_wheel.py
```diff
@@ -0,0 +1,694 @@
...
+    def _merge(
+        self, start: int, end: int, left: int, right: int
+    ) -> Iterator[tuple[int, int]]:
+        """Return an iterator of intervals to be fetched.
+
+        Args:
+            start (int): Start of needed interval
+            end (int): End of needed interval
+            left (int): Index of first overlapping downloaded data
+            right (int): Index after last overlapping downloaded data
```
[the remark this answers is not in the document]
radoering: done

Secrus (approved):

Issue #8816, opened 2023-12-22
Unclarity with `poetry show --why`


ValentinCalomme:
> This is in the context of me using `poetry==1.7.1`

Hi! was going through the documentation and saw the `--why` flag for the `poetry show` option.

Currently, the only explanation as to what the flag does is: 

> --why: When showing the full list or a --tree for a single package, display why a package is included.

I'd say what it means is super clear. Of course, I tried out the flag, which is clearer when used with the `--tree` option. Something like this would be more obvious to me:

> --why: When showing the full list, or a --tree for a single package, display whether a package is a direct dependency or which packages require it

Thanks a lot, looking forward to version 1.8!

dimbleby:
To improve docs or help text - submit a merge request!

ValentinCalomme:
I would love to submit a merge request, but I would like to know whether my interpretation is correct to begin with 😊

PR #8817, opened 2023-12-22
Update description of --why flag for poetry show

ValentinCalomme:
Resolves: #8816 

Changes:
- Update description in `docs/cli.md`
- Kept the description in the `docs` in sync with the `cli` by updating `src/poetry/console/commands/show.py`

Motivation:

The current description was unclear. "display why a package is included" could suggest that users could get back a motivation or rationale behind why the package is added. But that's not the case.

`--why` returns whether a package is a direct dependency or whether it is itself a dependency of another package in the project.

### DIFF
--- a/docs/cli.md
+++ b/docs/cli.md
@@ -503,7 +503,7 @@ required by
 ### Options
 
 * `--without`: The dependency groups to ignore.
-* `--why`: When showing the full list, or a `--tree` for a single package, display why a package is included.
+* `--why`: When showing the full list, or a `--tree` for a single package, display whether they are direct dependency or required by other packages.
 * `--with`: The optional dependency groups to include.
 * `--only`: The only dependency groups to include.
 * `--no-dev`: Do not list the dev dependencies. (**Deprecated**, use `--without dev` or `--only main` instead)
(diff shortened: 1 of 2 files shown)


docs/cli.md
```diff
@@ -503,7 +503,7 @@ required by
 ### Options
 
 * `--without`: The dependency groups to ignore.
-* `--why`: When showing the full list, or a `--tree` for a single package, display why a package is included.
+* `--why`: When showing the full list, or a `--tree` for a single package, display whether they are direct dependency or required by other packages.
```
radoering, line 506:
```suggestion
* `--why`: When showing the full list, or a `--tree` for a single package, display whether they are a direct dependency or required by other packages.
```

src/poetry/console/commands/show.py
```diff
@@ -48,8 +48,8 @@ class ShowCommand(GroupCommand, EnvCommand):
         option(
             "why",
             None,
-            "When showing the full list, or a <info>--tree</info> for a single"
-            " package, also display why it's included.",
+            "When showing the full list, or a <info>--tree</info> for a single package,"
+            " display whether they are direct dependency or required by other packages",
```
radoering, line 52:
```suggestion
            " display whether they are a direct dependency or required by other packages",
```

radoering (approved):

PR #8831, opened 2023-12-29
env: delete entry in envs.toml when running `poetry env remove --all`

radoering:
# Pull Request Check List

Resolves: #8631
Closes: #8665

Alternative approach to #8665. Just remove the entry in envs.toml when running `poetry env remove --all`. This even covers the case that the entry refers to a virtualenv that had been removed manually before.

The actual change is in `remove.py`. Changes in `env_manager.py` are just refactorings to avoid code duplication and document the `envs.toml` file a little bit better.

### DIFF
--- a/src/poetry/utils/env/env_manager.py
+++ b/src/poetry/utils/env/env_manager.py
@@ -9,6 +9,7 @@ import shutil
 import subprocess
 import sys
 
+from functools import cached_property
 from pathlib import Path
 from subprocess import CalledProcessError
 from typing import TYPE_CHECKING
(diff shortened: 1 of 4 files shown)


src/poetry/utils/env/env_manager.py
```diff
@@ -121,11 +160,20 @@ def in_project_venv(self) -> Path:
         venv: Path = self._poetry.file.path.parent / ".venv"
         return venv
 
-    def activate(self, python: str) -> Env:
+    @cached_property
+    def envs_file(self) -> EnvsFile:
         venv_path = self._poetry.config.virtualenvs_path
-        cwd = self._poetry.file.path.parent
+        return EnvsFile(venv_path / self.ENVS_FILE)
```
Secrus, line 166: Could we just use the `self._poetry.config.virtualenvs_path` instead of adding a variable?

radoering: Done.

Secrus (approved):

Issue #4055, opened 2021-05-10
Option to force Poetry to create a virtual environment, even if a virtual env is active


chernals:
I am using Poetry from within a `conda` environment; with Poetry being installed by conda (`poetry` is present in the `environment.yaml` file). This might not be ideal but for a specific setup this seems to work well.

When running `poetry install`, a `venv` is not created because Poetry detect that a virtual environment (the Conda one) is already active.

I believe that we should be able to force Poetry to create a virtual environment anyway: the Conda environment is more "physical" than virtual in this case, as it replaces the system Python.

The same issue occurs when doing `poetry shell` from within the Conda environment, after creating the virtual environment with the following workaround: I deactivate the Conda environment, use the full path to Poetry (in the `bin` directory of the Conda environment) and do `poetry install`.

chernals:
This is related to ISSUE-1724 and to ISSUE-4050, but in the form of an explicit request to add an option.

mforbes:
What would really be nice is a way to have an active underlying conda environment upon which poetry could create a virtual environment for installing dependencies while maintaining access to the underlying conda environment.

A use-case is a group of students working on a shared computer (like an HPC cluster): I would like to create a fairly full featured shared conda environment (called `work` here) which students have read access to, but not write.  Currently poetry will treat this active conda environment as a virtual environment, but fail because it has no write access.

The behaviour I would like (if possible) is for poetry to be able to install or upgrade any missing dependencies as needed, akin to how `pip install --user` would, but to install these into a custom virtual environment rather than the users `~/.local`.

I know that I could do this by manipulating paths `PYTHONUSERBASE` etc. but it would be really nices if this could "just work" in a way that consistent with general poetry usage.

Thus, daily work looks like:

```bash
conda activate work  # Activate read-only conda environment
poetry shell         # Creates a new virtual environment with access to `work`
python ...           # Works for anything using just `work`
poetry add ...       # If something is not in `work` (currently fails because of permissions)
poetry install       # Installed into the virtual environment like pip install --user
```

This allows students to get to work as quickly as possible, allowing us to provide most
of what they need in the `work` environment, but providing them a way to install other
libraries if needed.  Everything new needed is recorded in `pyproject.toml`, so in
combination with the environment file for the `work` environment, this satisfies the
basic requirements for reproducibility.

When preparing for release, one would add the minimal set of missing dependencies
supplied by work into a clean virtual environment:

```bash
poetry env use 3.8
[...]

PR #8833, opened 2023-12-29
(#4055) Update warning documentation (include virtualenv notice)

aesteve:
Mention that virtual environment in which poetry is installed should not be activated

See #4055 (only the doc part)


radoering:
pre-commit.ci autofix

radoering (approved):

PR #8834, opened 2023-12-30
solver: improve verbose output for dependencies with extras

radoering:

When running `poetry lock -vv` with a dependency with extras `foo[bar]`:

- without PR: `foo (1.0) depends on foo (1.0)`
- with PR: `foo[bar] (1.0) depends on foo (1.0)`

### DIFF
--- a/src/poetry/mixology/incompatibility.py
+++ b/src/poetry/mixology/incompatibility.py
@@ -432,7 +432,8 @@ class Incompatibility:
 
         if term.dependency.source_type:
             return str(term.dependency)
-        return f"{term.dependency.pretty_name} ({term.dependency.pretty_constraint})"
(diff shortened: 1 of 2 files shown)


src/poetry/mixology/incompatibility.py
```diff
@@ -432,7 +432,8 @@ def _terse(self, term: Term, allow_every: bool = False) -> str:
 
         if term.dependency.source_type:
             return str(term.dependency)
-        return f"{term.dependency.pretty_name} ({term.dependency.pretty_constraint})"
+        complete_pretty_name = str(term.dependency).split(" (")[0]
```
radoering, line 435: TODO: Use `complete_pretty_name` from python-poetry/poetry-core#677

Secrus (approved):

PR #8843, opened 2024-01-03
Fix git clone test on non-English locale

kkhramova:
# Pull Request Check List

-----
When running tests using a non-English locale `vcs.git.test_system.TestSystemGit` test fails because git uses localized messages in its output.  

To reproduce run:

```sh
LANG=uk_UA.UTF-8 poetry run pytest
```

This PR fixes the issue by ignoring any output from git commands.

### DIFF
--- a/src/poetry/vcs/git/system.py
+++ b/src/poetry/vcs/git/system.py
@@ -15,18 +15,18 @@ if TYPE_CHECKING:
 
 class SystemGit:
     @classmethod
-    def clone(cls, repository: str, dest: Path) -> str:
+    def clone(cls, repository: str, dest: Path) -> None:
         cls._check_parameter(repository)
 
-        return cls.run("clone", "--recurse-submodules", "--", repository, str(dest))
+        cls.run("clone", "--recurse-submodules", "--", repository, str(dest))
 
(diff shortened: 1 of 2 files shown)


dimbleby:
IMO it doesnt make a lot of sense for `SystemGit.run()` (and `.clone()` and `.checkout()`) to return strings.

Suggest that it would be preferable to use `subprocess.check_call()` - sending stdout to devnull - instead of `subprocess.check_output()`, then the locale is of no interest to anyone.

kkhramova:
I made the changes, @dimbleby. 
Can you please take another look?

dimbleby:
> Suggest that it would be preferable to use subprocess.check_call() - sending stdout to devnull - instead of subprocess.check_output()

kkhramova:
Thank you, @dimbleby 
Now subprocess.check_output() is changed to subprocess.check_call()

dimbleby:
and how about redirecting stdout?  does the code as written write the output to the terminal?

kkhramova:
stdout is now redirected to devnull.
@dimbleby thank you for your patient guidance.

radoering (approved):

Issue #7778, opened 2023-04-08
unsupported hash type sha


gygabyte017:
- **Poetry version**: 1.4.2
- **Python version**: 3.8.16
- **OS version and name**: Windows Server 2019
- **pyproject.toml**: none

- [x] I am on the [latest](https://github.com/python-poetry/poetry/releases/latest) stable Poetry version, installed using a recommended method.
- [x] I have searched the [issues](https://github.com/python-poetry/poetry/issues) of this repo and believe that this is not a duplicate.
- [x] I have consulted the [FAQ](https://python-poetry.org/docs/faq/) and [blog](https://python-poetry.org/blog/) for any relevant entries or release notes.
- [x] If an exception occurs when executing a command, I executed it again in debug mode (`-vvv` option) and have included the output below.

## Issue

When installing the package `swat` I get this error: `unsupported hash type sha`

```
>poetry add swat@latest
Using version ^1.12.1 for swat

Updating dependencies
Resolving dependencies... Downloading https://files.pythonhosted.org/packages/3c/c8/5d6bb20ba20d8770e933d8c0fe956fdc11940af2c39f5b23e02196e0395c/swat-1.12.1-0-cp27-cp27m-macosx_10_9_x86_64.whl#sha256=30ad0ce7e69fa40e19433aa8327b55ae1c0f092fResolving dependencies... (1.6s)

Writing lock file

Package operations: 1 install, 0 updates, 0 removals

  • Installing swat (1.12.1): Failed

  ValueError

  unsupported hash type

  at D:\miniconda\lib\hashlib.py:157 in __hash_new
      153│         # It does neither support keyed blake2 nor advanced features like
      154│         # salt, personal, tree hashing or SSE.
      155│         return __get_builtin_constructor(name)(data, **kwargs)
      156│     try:
    → 157│         return _hashlib.new(name, data)
      158│     except ValueError:
      159│         # If the _hashlib module (OpenSSL) doesn't support the named
      160│         # hash, try using our builtin implementations.
      161│         # This allows for SHA224/256 and SHA384/512 support even though

The following error occurred when trying to handle this error:


  ValueError

  unsupported hash type sha

  at D:\miniconda\lib\hashlib.py:120 in __get_builtin_constructor
      116│     constructor = cache.get(name)
      117│     if constructor is not None:
      118│         return constructor
      119│
    → 120│     raise ValueError('unsupported hash type ' + name)
      121│
      122│
      123│ def __get_openssl_constructor(name):
[...]

PR #8851, opened 2024-01-05
Added archive validation for all available algorithms in hashlib.

GMouzourou:
Resolves: #4578 

May also contribute to #7881 , as it's unclear if the known hashes being checked are all sha256.

### DIFF
--- a/src/poetry/installation/executor.py
+++ b/src/poetry/installation/executor.py
@@ -797,7 +797,7 @@ class Executor:
             f["hash"] for f in package.files if f["file"] == archive.name
         }
         hash_types: set[str] = {t.split(":")[0] for t in known_hashes}
-        hash_type: str | None = get_highest_priority_hash_type(hash_types)
+        hash_type: str | None = get_highest_priority_hash_type(hash_types, archive.name)
 
         if hash_type is None:
             raise RuntimeError(
--- a/src/poetry/installation/executor.py
+++ b/src/poetry/installation/executor.py
@@ -793,11 +793,9 @@ class Executor:
 
     @staticmethod
     def _validate_archive_hash(archive: Path, package: Package) -> str:
-        known_hashes: set[str] = {
-            f["hash"] for f in package.files if f["file"] == archive.name
-        }
-        hash_types: set[str] = {t.split(":")[0] for t in known_hashes}
-        hash_type: str | None = get_highest_priority_hash_type(hash_types, archive.name)
+        known_hashes = {f["hash"] for f in package.files if f["file"] == archive.name}
+        hash_types = {t.split(":")[0] for t in known_hashes}
+        hash_type = get_highest_priority_hash_type(hash_types, archive.name)
 
         if hash_type is None:
             raise RuntimeError(
@@ -805,7 +803,7 @@ class Executor:
                 f" {archive.name} found (known hashes: {known_hashes!s})"
             )
 
-        archive_hash: str = f"{hash_type}:{get_file_hash(archive, hash_type)}"
+        archive_hash = f"{hash_type}:{get_file_hash(archive, hash_type)}"
 
         if archive_hash not in known_hashes:
             raise RuntimeError(
--- a/src/poetry/installation/executor.py
+++ b/src/poetry/installation/executor.py
@@ -3,6 +3,7 @@ from __future__ import annotations
 import contextlib
 import csv
 import functools
+import hashlib
 import itertools
 import json
 import threading
@@ -792,16 +793,22 @@ class Executor:
 
     @staticmethod
     def _validate_archive_hash(archive: Path, package: Package) -> str:
-        archive_hash: str = "sha256:" + get_file_hash(archive)
         known_hashes = {f["hash"] for f in package.files if f["file"] == archive.name}
+        hash_types = {
+            t.split(":")[0]
+            for t in known_hashes
+            if t.split(":")[0] in hashlib.algorithms_available
+        }
+        archive_hashes = {f"{t}:{get_file_hash(archive, t)}" for t in hash_types}
+        matching_hashes = known_hashes.intersection(archive_hashes)
 
-        if archive_hash not in known_hashes:
+        if len(matching_hashes) < 1:
             raise RuntimeError(
                 f"Hash for {package} from archive {archive.name} not found in"
-                f" known hashes (was: {archive_hash})"
+                f" known hashes (was: {archive_hashes!s})"
             )
 
-        return archive_hash
+        return matching_hashes.pop()
 
     def _download_archive(
         self,
--- a/src/poetry/installation/executor.py
+++ b/src/poetry/installation/executor.py
@@ -3,7 +3,6 @@ from __future__ import annotations
 import contextlib
 import csv
 import functools
-import hashlib
 import itertools
 import json
 import threading
@@ -32,6 +31,7 @@ from poetry.utils.authenticator import Authenticator
 from poetry.utils.env import EnvCommandError
 from poetry.utils.helpers import Downloader
 from poetry.utils.helpers import get_file_hash
+from poetry.utils.helpers import get_highest_priority_hash_type
 from poetry.utils.helpers import pluralize
 from poetry.utils.helpers import remove_directory
 from poetry.utils.pip import pip_install
@@ -793,22 +793,27 @@ class Executor:
 
     @staticmethod
     def _validate_archive_hash(archive: Path, package: Package) -> str:
-        known_hashes = {f["hash"] for f in package.files if f["file"] == archive.name}
-        hash_types = {
-            t.split(":")[0]
-            for t in known_hashes
-            if t.split(":")[0] in hashlib.algorithms_available
+        known_hashes: set = {
+            f["hash"] for f in package.files if f["file"] == archive.name
         }
-        archive_hashes = {f"{t}:{get_file_hash(archive, t)}" for t in hash_types}
-        matching_hashes = known_hashes.intersection(archive_hashes)
+        hash_types: set = {t.split(":")[0] for t in known_hashes}
+        hash_type: str | None = get_highest_priority_hash_type(hash_types)
 
-        if len(matching_hashes) < 1:
+        if hash_type is None:
             raise RuntimeError(
-                f"Hash for {package} from archive {archive.name} not found in"
-                f" known hashes (was: {archive_hashes!s})"
+                f"No usable hash type(s) for {package} from archive"
+                f" {archive.name} found (known hashes: {known_hashes!s})"
             )
 
-        return matching_hashes.pop()
+        archive_hash: str = f"{hash_type}:{get_file_hash(archive, hash_type)}"
+
+        if archive_hash not in known_hashes:
+            raise RuntimeError(
+                f"Hash for {package} from archive {archive.name} not found in"
+                f" known hashes (was: {archive_hash})"
+            )
+
+        return archive_hash
 
     def _download_archive(
         self,
--- a/src/poetry/installation/executor.py
+++ b/src/poetry/installation/executor.py
@@ -793,10 +793,10 @@ class Executor:
 
     @staticmethod
     def _validate_archive_hash(archive: Path, package: Package) -> str:
-        known_hashes: set = {
+        known_hashes: set[str] = {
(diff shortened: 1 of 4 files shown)


radoering:
Some thoughts:

- Regardless of the changes in this PR, it looks like we are checking that the hash of an archive matches the hash of any known archive. Actually, we should check that it matches the hash of the correct archive since we know the name of the archive, shouldn't we?
- With this PR we calculate multiple hashes if we have several files with different hash types. What's the performance impact? (This would be a non-issue if we'd fix the above point first.)

dimbleby:
This certainly has nothing to do with #7778

GMouzourou:
Hi @radoering,

When calculating the known hashes, we filter for `if f["file"] == archive.name`. So we only check hashes for the file given.

What would impact performance is if we had multiple hashes for the same given file. This can be seen on the line were we generate `archive_hashes`. This could be changed to just picking the first hash type from `hash_types` if required.

radoering:
Thanks for the hint, I overlooked this. So in theory, the same archive could appear several times (with different hash types) in the list. (I'm not sure whether this can happen in practice but let's assume it can.)

> This could be changed to just picking the first hash type from `hash_types` if required.

I don't think picking the first hash type is good enough. We do not want to rely on md5 if there are more secure hashes. I suppose, we have to put a list of preferred hash types somewhere (maybe in `poetry.utils.helpers`). Then, we can use the first hash type in the list that is available and just fallback to a "random" choice if no hash type is in our list.

I assume normally there shouldn't be other hashes than these in the lockfile anyway so this might all be a theoretical issue after all:
https://github.com/python-poetry/poetry/blob/022308e9c4ce1f9dec17fa57d44fa9ab9a0c4a16/src/poetry/repositories/http_repository.py#L227-L232

GMouzourou:
I've created a `get_highest_priority_hash_type` function in `poetry.utils.helpers` as suggested. So now only one hash algorithm is used per archive (the most cryptographically secure). If you have issue with the priority of the algorithms, let me know the priority you'd like and I'll update the list.

I've also added tests for multiple hash types of the same file, multiple files in the same package, and an unsupported hash type.

radoering:
I think the approach makes sense in general. We just have to clarify some details.

I don't really like the repetetiveness of `get_highest_priority_hash_type`(). Further, you might get the log message that the hash is not in the priority list if it is in the list but not in `hashlib.algorithms_available`. I would probably rewrite it like this:

```
def get_preferred_available_hash_type(hash_types: set[str]) -> str | None:
    if not hash_types:
        return None

    # ordered by preference (most preferred first)
    preferred_hashes = ('sha512', 'sha256', ...)
    for chosen in preferred_hashes:
        if chosen in hash_types:
            if chosen in hashlib.algorithms_available:
                return chosen
            logger.debug("Hash type %s not available in hashlib", chosen)

    # do not consider hashes that are not available in hashlib again
    hash_types = hash_types - set(preferred_hashes)

    if hash_types:
        logger.debug("No hash type of %s in preferred hashes list", hash_types)
        # choose any available hash type
        for chosen in hash_types:
            if chosen in hashlib.algorithms_available:
                return chosen
            logger.debug("Hash type %s not available in hashlib", chosen)

    return None
```

We definitively want a parameterized unit test for this function that checks the special cases (regardless of the exact implementation).

> the most cryptographically secure

That's probably fine. I don't expect performance to be an issue but if it was that would also be a factor to consider...

> If you have issue with the priority of the algorithms, let me know the priority you'd like and I'll update the list.

LGTM but I'm not an expert. I suppose we can change it later if someone with more expertise comes along. Just one thing: I'd probably omit md5 (and maybe sha1?) because they are known to be insecure. If we omit them from the preferred/priority list, we still have a chance to choose a better algorithm, which is unknown to us. Not something you should rely on but anyway...

GMouzourou:
I didn't particularly like that big block of if statements either. I did originally write it as a match\case, but then realised poetry supports versions of python older than 3.10.

I think I got the gist of where you were going in your code snippet. I pulled out the creation of the prioritised hash types into a global, as I didn't like the idea of doing all that collection manipulation for every archive and to only get the same answer every time. Also python `set` is an unordered collection, so you wouldn't have got your intended behaviour.

As to your MD5 and SHA1 comment, preferring other algorithms is a must (which is what this implementation does). Omitting them from the list will remove the ability to ever use them. I took it that was not your intent, but if you did want to prohibit the use of archives with only an MD5 or SHA1 hash then I will remove them. That does tend to force people to point to git repositories to avoid any checks (or to lesser dependency management tools 😝), which is obviously worse.

radoering:
> I pulled out the creation of the prioritised hash types into a global, as I didn't like the idea of doing all that collection manipulation for every archive and to only get the same answer every time.

Makes sense.

> Also python `set` is an unordered collection, so you wouldn't have got your intended behaviour.

Not sure what your are referring to. The preferred/prioritized hashes must be ordered, but the `hash_types` we pass to the function can be a set.

> Omitting them from the list will remove the ability to ever use them.

Not when using my code snippet. 😉 My idea is that we still choose an available hash type even if none is in our preferred list. That part is missing now. I'll add a comment in the code to make it clearer.

> Is there a good mechanism to warn users if we have had to check an archive using MD5 or SHA1?

We could have another list/set of deprecated hash types. However, I'm not sure if it's worth it.

src/poetry/utils/helpers.py
```diff
@@ -37,6 +37,28 @@
...
+        "shake_128",
+        "blake2s",
+        "blake2b",
+        "sha1",
+        "md5",
+    ]
+    if t in hashlib.algorithms_available
+]
+
+
```
radoering, line 61:
```suggestion

non_prioritized_hash_types = hashlib.algorithms_available - set(prioritized_hash_types)

```

src/poetry/utils/helpers.py
```diff
@@ -312,3 +334,14 @@ def extractall(source: Path, dest: Path, zip: bool) -> None:
...
+def get_highest_priority_hash_type(hash_types: set[str]) -> str | None:
+    if not hash_types:
+        return None
+
+    for prioritised_hash_type in prioritised_hash_types:
+        if prioritised_hash_type in hash_types:
+            return prioritised_hash_type
+
+    return None
```
radoering, line 347:
```suggestion

    logger.debug('None of the hash types %s is in prioritized_hash_types', hash_types)

    for hash_type in hash_types:
        if hash_type in non_prioritized_hash_types:
            return hash_type

    logger.warning('None of the hash types %s is available in hashlib', hash_types)

    return None
```

If there is no prioritized hash type we just "randomly" choose an available hash type. That's why md5 and sha1 can be omitted in `prioritized_hash_types`.

If none of the hash types is available we probably should log a warning.

GMouzourou: Cool, I see what your saying now. I'll create a commit for this...

GMouzourou:
So I've:

- Removed MD5 and SHA1 from the prioritised list.
- Introduced your `non_prioritized_available_hash_types` and the processing that goes with it.

I didn't include the warning log you had as this catered for in `executor.py`.

GMouzourou:
> > Also python `set` is an unordered collection, so you wouldn't have got your intended behaviour.
> 
> Not sure what your are referring to. The preferred/prioritized hashes must be ordered, but the `hash_types` we pass to the function can be a set.

I thought `preferred_hashes = ('sha512', 'sha256', ...)` was a `set` not a `tuple`. My bad.

tests/installation/test_executor.py
```diff
@@ -1456,3 +1456,149 @@ def test_other_error(
     output = io.fetch_output().strip()
     assert output.startswith(expected_start)
     assert output.endswith(expected_end)
+
+
+def test_executor_known_hash_sha512(
```
radoering, line 1461: Is it necessary to have three similar test cases or is one parameterized test case sufficient?

tests/installation/test_executor.py
```diff
@@ -1456,3 +1456,149 @@ def test_other_error(
...
+
+    executor = Executor(tmp_venv, pool, config, io)
+    return_code = executor.execute([Install(package)])
+    distributions = list(tmp_venv.site_packages.distributions(name=package.name))
+    assert len(distributions) == 0
+
+    output = io.fetch_output()
+    error = io.fetch_error()
+    assert return_code == 1, f"\noutput: {output}\nerror: {error}\n"
+    assert "pytest" not in error
```
radoering, line 1602:
```suggestion
```
?

src/poetry/utils/helpers.py
```diff
@@ -312,3 +336,25 @@ def extractall(source: Path, dest: Path, zip: bool) -> None:
                 archive.extractall(dest, filter="data")
             else:
                 archive.extractall(dest)
+
+
+def get_highest_priority_hash_type(
```
radoering, line 341: Wants a (parameterized) test in `test_helpers` to cover the different cases.

radoering (changes requested):

tests/installation/test_executor.py
radoering: If you now add `packages_files` and `expected_url_reference` as parameters (like [that](https://github.com/python-poetry/poetry/blob/e31a9fc3ce5a7fc72eefa22040676bc66a35e5ad/tests/installation/test_executor.py#L247-L279)) that's what I meant. 😅

src/poetry/utils/helpers.py
radoering: That's still open. I know it is implicitly (partially) tested in `test_executor.py`. Nevertheless, we should add an explicit test for this function in `test_helpers.py`.

src/poetry/installation/executor.py
```diff
@@ -792,8 +793,19 @@ def _populate_hashes_dict(self, archive: Path, package: Package) -> None:
 
     @staticmethod
     def _validate_archive_hash(archive: Path, package: Package) -> str:
-        archive_hash: str = "sha256:" + get_file_hash(archive)
-        known_hashes = {f["hash"] for f in package.files if f["file"] == archive.name}
+        known_hashes: set[str] = {
+            f["hash"] for f in package.files if f["file"] == archive.name
+        }
+        hash_types: set[str] = {t.split(":")[0] for t in known_hashes}
+        hash_type: str | None = get_highest_priority_hash_type(hash_types, archive.name)
```
radoering, line 800:
```suggestion
        known_hashes = {f["hash"] for f in package.files if f["file"] == archive.name}
        hash_types = {t.split(":")[0] for t in known_hashes}
        hash_type = get_highest_priority_hash_type(hash_types, archive.name)
```

We don't need these annotations since the type is clear.

src/poetry/installation/executor.py
```diff
@@ -792,8 +793,19 @@ def _populate_hashes_dict(self, archive: Path, package: Package) -> None:
...
+
+        if hash_type is None:
+            raise RuntimeError(
+                f"No usable hash type(s) for {package} from archive"
+                f" {archive.name} found (known hashes: {known_hashes!s})"
+            )
+
+        archive_hash: str = f"{hash_type}:{get_file_hash(archive, hash_type)}"
```
radoering, line 808:
```suggestion
        archive_hash = f"{hash_type}:{get_file_hash(archive, hash_type)}"
```

src/poetry/utils/helpers.py
```diff
@@ -36,6 +37,29 @@
...
+        "sha384",
+        "sha256",
+        "sha224",
+        "shake_256",
+        "shake_128",
+        "blake2s",
+        "blake2b",
+    ]
+    if t in hashlib.algorithms_available
+])
```
radoering, line 58:
```suggestion
prioritised_hash_types: tuple[str, ...] = tuple(
    t
    for t in [
        "sha3_512",
        "sha3_384",
        "sha3_256",
        "sha3_224",
        "sha512",
        "sha384",
        "sha256",
        "sha224",
        "shake_256",
        "shake_128",
        "blake2s",
        "blake2b",
    ]
    if t in hashlib.algorithms_available
)
```

src/poetry/utils/helpers.py
radoering, line 341: Can you move the function up, so it's after `get_file_hash`?

radoering (changes requested):
Just some minor nitpicks.

radoering (approved):

Issue #8698, opened 2023-11-21
Default for the include flag is sdist only, not sdist AND wheel


rdauncey53:
- [x] I have searched the [issues](https://github.com/python-poetry/poetry/issues) of this repo and believe that this is not a duplicate.

## Issue

I was hitting a bunch of issues trying to include resource files in my Poetry build using the `include` key. It turns out that despite [the docs](https://python-poetry.org/docs/pyproject/#include-and-exclude) saying that

> If no format is specified, it will default to include both sdist and wheel.

It in fact seems to default to only `sdist`. The workaround was to change

```
include = "my_folder"
```

to

```
include = [
    { path = "my_folder", format = ["sdist", "wheel"] }
]
```

In addition to this, I find the documentation unclear / challenging in terms of understanding how to access these files when pulling the package in as a dependency. If the files aren't included in the module, I am struggling to get `importlib.resources` to work.

If it is required that files the package needs to access live _inside_ the package folder, please could this be added to the documentation? If not, could you add some guidance on this?

dimbleby:
the best way to make docs changes happen is to submit merge requests!

your question about where files should live seems like more of a question about how python modules work at all, than about how poetry builds packages, I'm doubtful that poetry should take responsibility for documenting that.

rdauncey53:
Fair enough - I think the reason for not making the change myself is because I don't know what the correct default behaviour should be. Happy to make the change if it is just change to "If no format is specified, it will default to include only sdist".

On the general Python questions, yes I agree this is largely outside of the Poetry remit. I think the poetry-relevant bit (which I admit I didn't make clear in my original post) is that Poetry will include those files in the `site-packages` folder upon Poetry install, which feels like a slightly odd design - I'm sure there is a good reason for that, but this behaviour isn't particularly well documented and makes working with the include flag difficult. I only figured things out because I found this (closed) Poetry [issue](https://github.com/python-poetry/poetry/issues/2015).

dimbleby:
I assume that making the documentation match the code is uncontroversial and just a clear improvement: even if a later fix decides to change both.

If you or anyone else _does_ think that it is the code that is wrong I suppose [this](https://github.com/python-poetry/poetry-core/blob/1ee19a572484b4bc3929d9229763d3a2b6735aba/src/poetry/core/masonry/builders/builder.py#L168) is the relevant line

PR #8852, opened 2024-01-05
Fix documentation for default `format` option for `include` and `exclude`

zanieb:
The existing documentation is incorrect, which confused me when attempting to `include` a directory.

I verified this behavior by building a package with `include` and `exclude` and inspecting the tarballs created.

I'd appreciate if someone pointed me to the code for this — I tried to find it but could not :)

# Pull Request Check List

Closes https://github.com/python-poetry/poetry/issues/8698

### DIFF
--- a/docs/pyproject.md
+++ b/docs/pyproject.md
@@ -279,7 +279,9 @@ include = [
 ]
 ```
 
-If no format is specified, it will default to include both `sdist` and `wheel`.
+If no format is specified, `include` defaults to only `sdist`.
+
+In constrast, `exclude` defaults to both `sdist` and `wheel`.
 
 ```toml
 exclude = ["my_package/excluded.py"]
--- a/docs/pyproject.md
+++ b/docs/pyproject.md
@@ -281,7 +281,7 @@ include = [
 
 If no format is specified, `include` defaults to only `sdist`.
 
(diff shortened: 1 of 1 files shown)


Secrus:
This code lives in `poetry-core` repo. Namely [here](https://github.com/python-poetry/poetry-core/blob/main/src/poetry/core/masonry/builders/sdist.py) and [here](https://github.com/python-poetry/poetry-core/blob/main/src/poetry/core/masonry/builders/wheel.py).

zanieb:
Ah interesting thank you! No wonder I couldn't find it with a search :)

dimbleby:
cf #8698

zanieb:
As far as I can tell, this should be ready to review — I agree with the discussion in the issue that the documentation should be corrected separately from considering if this behavior is correct.

docs/pyproject.md
```diff
@@ -279,7 +279,9 @@ include = [
 ]
 ```
 
-If no format is specified, it will default to include both `sdist` and `wheel`.
+If no format is specified, `include` defaults to only `sdist`.
+
+In constrast, `exclude` defaults to both `sdist` and `wheel`.
```
radoering, line 284:
```suggestion
In contrast, `exclude` defaults to both `sdist` and `wheel`.
```

radoering (approved):

Issue #8892, opened 2024-01-19
Encoding Warnings with PEP 597 Warnings Enabled


danyeaw:
- **Poetry version**: 1.7.1
- **Python version**: 3.11.7 and 3.12.1
- **OS version and name**: Windows 11 / macOS 12.7.2  / Linux Fedora 39
- **pyproject.toml**: https://github.com/gaphor/gaphor/blob/main/pyproject.toml

- [X] I am on the [latest](https://github.com/python-poetry/poetry/releases/latest) stable Poetry version, installed using a [recommended method](https://python-poetry.org/docs/#installation).
- [X] I have searched the [issues](https://github.com/python-poetry/poetry/issues) of this repo and believe that this is not a duplicate.
- [X] I have consulted the [FAQ](https://python-poetry.org/docs/faq/) and [blog](https://python-poetry.org/blog/) for any relevant entries or release notes.
- [X] If an exception occurs when executing a command, I executed it again in debug mode (`-vvv` option) and have included the output below.

## Issue

With [PEP 597](https://peps.python.org/pep-0597/) Encoding Warnings enabled using PYTHONWARNDEFAULTENCODING=true, Poetry prints encoding warnings while doing a `poetry install`.

In a future version of Python, UTF-8 will be the default encoding. It is also easy to assume that the default encoding right now is UTF-8, which can introduce bugs for Windows users.

If there is interest in fixing this, I would be glad to submit a PR to resolve these warnings. We could also enable the Encoding Warnings on Poetry's GitHub Actions to avoid further introduction of code without encodings specified.

```
poetry install --only main,packaging --no-interaction
C:\hostedtoolcache\windows\Python\3.11.7\x64\Lib\site-packages\poetry\utils\env\base_env.py:350: EncodingWarning: 'encoding' argument not specified.
  output = subprocess.check_output(
C:\hostedtoolcache\windows\Python\3.11.7\x64\Lib\site-packages\poetry\repositories\installed_repository.py:61: EncodingWarning: 'encoding' argument not specified
  with pth_file.open() as f:
Installing dependencies from lock file

Package operations: 7 installs, 2 updates, 0 removals

  � Updating pycairo (1.25.1 C:/gtk-build/gtk/x64/release/pycairo-1.25.1-cp311-cp311-win_amd64.whl -> 1.25.1)
  � Installing altgraph (0.17.4)
  � Installing pefile (2023.2.7)
[...]

PR #8893, opened 2024-01-21
Fix encoding warnings with PEP 597 enabled

danyeaw:
# Pull Request Check List

Resolves: #8892 

This PR fixes Encoding Warnings with PEP 597 warnings enabled and enables the warnings in CI. There are also a few warnings with poetry-core and cleo, and I can submit PRs for those repos as well.

### DIFF
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -13,6 +13,9 @@ defaults:
   run:
     shell: bash
 
+env:
+  PYTHONWARNDEFAULTENCODING: 'true'
+
 jobs:
   changes:
     name: Detect changed files
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -13,9 +13,6 @@ defaults:
   run:
     shell: bash
 
-env:
-  PYTHONWARNDEFAULTENCODING: 'true'
-
 jobs:
   changes:
     name: Detect changed files
--- a/src/poetry/repositories/installed_repository.py
+++ b/src/poetry/repositories/installed_repository.py
@@ -13,6 +13,7 @@ from poetry.core.packages.utils.utils import url_to_path
 from poetry.core.utils.helpers import module_name
 
 from poetry.repositories.repository import Repository
+from poetry.utils._compat import getencoding
 from poetry.utils._compat import metadata
 
 
@@ -58,7 +59,7 @@ class InstalledRepository(Repository):
             if not pth_file.exists():
                 continue
 
-            with pth_file.open() as f:
+            with pth_file.open(encoding=getencoding()) as f:
                 for line in f:
                     line = line.strip()
                     if line and not line.startswith(("#", "import ", "import\t")):
--- a/src/poetry/utils/env/base_env.py
+++ b/src/poetry/utils/env/base_env.py
@@ -330,8 +330,9 @@ class Env:
                 subprocess.check_call(cmd, stderr=stderr, env=env, **kwargs)
                 output = ""
             else:
+                encoding = "locale" if sys.version_info >= (3, 10) else None
                 output = subprocess.check_output(
-                    cmd, stderr=stderr, env=env, text=True, **kwargs
+                    cmd, stderr=stderr, env=env, text=True, encoding=encoding, **kwargs
                 )
         except CalledProcessError as e:
             raise EnvCommandError(e)
--- a/.github/workflows/.tests-matrix.yaml
+++ b/.github/workflows/.tests-matrix.yaml
@@ -22,6 +22,9 @@ defaults:
   run:
     shell: bash
 
+env:
+  PYTHONWARNDEFAULTENCODING: 'true'
+
 jobs:
   mypy:
     name: mypy
--- a/src/poetry/masonry/builders/editable.py
+++ b/src/poetry/masonry/builders/editable.py
@@ -3,7 +3,6 @@ from __future__ import annotations
 import csv
 import hashlib
 import json
-import locale
 import os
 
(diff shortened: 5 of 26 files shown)


src/poetry/utils/env/base_env.py
```diff
@@ -348,7 +348,7 @@ def _run(self, cmd: list[str], **kwargs: Any) -> str:
                 output = ""
             else:
                 output = subprocess.check_output(
-                    cmd, stderr=stderr, env=env, text=True, **kwargs
+                    cmd, stderr=stderr, env=env, text=True, encoding="utf-8", **kwargs
```
dimbleby, line 351: this one looks suspect - why should we assume that the subprocess is producing utf8 output?

looks like callers to this arrive via `run_python_scipt()`, which so far as I can see does not do anything to force utf8

danyeaw: Hi @dimbleby, thanks for the feedback! I think you are right, I updated it so that it enforces the locale encoding.

radoering:
pre-commit.ci autofix

src/poetry/repositories/installed_repository.py
```diff
@@ -58,7 +58,7 @@ def get_package_paths(cls, env: Env, name: str) -> set[Path]:
             if not pth_file.exists():
                 continue
 
-            with pth_file.open() as f:
+            with pth_file.open(encoding="utf-8") as f:
```
radoering, line 61: That might be wrong: #8041

radoering (changes requested):
There are still some warnings we should address.

src/poetry/repositories/installed_repository.py
danyeaw: Thanks for the feedback, that is now fixed!

danyeaw:
We are down to 1 warning:

```
tests/console/commands/test_init.py::test_respect_prefer_active_on_init[True-1.1]
  /home/dan/Projects/poetry/tests/console/commands/test_init.py:1091: EncodingWarning: 'encoding' argument not specified.
    result: str = orig_check_output(cmd, *_, **__)
```

I'm not sure what to do about that one since it is a mocked out version of check_output.

danyeaw:
Actually 1 more as well:

```
tests/vcs/git/test_system.py::TestSystemGit::test_clone_success
  /home/dan/Projects/poetry/src/poetry/vcs/git/system.py:43: EncodingWarning: 'encoding' argument not specified.
    subprocess.check_call(
```

I'm not sure how to apply encodings to a check_call.

danyeaw:
Hi @radoering, I rebased this on top of the main branch. Unfortunately mypy gives a `Unexpected keyword argument "encoding" for "check_call"`.

radoering:
If `check_call` supports this keyword and mypy thinks it does not, this is probably a mypy/typeshed bug and we should just add a `type: ignore` (and even better additionally create an issue/PR at https://github.com/python/typeshed).

What surprises me a little is that I no longer see any encoding warnings in the test output in the CI. Am I overlooking something? Does enabling the default encoding warning does not work anymore after the rebase?

danyeaw:
Hi @radoering, thanks again for the feedback. I created https://github.com/python/typeshed/issues/12084 and added a type ignore.

dimbleby:
https://docs.github.com/en/enterprise-cloud@latest/actions/using-workflows/reusing-workflows#limitations

> Any environment variables set in an env context defined at the workflow level in the caller workflow are not propagated to the called workflow

https://github.com/python-poetry/poetry/pull/9534

radoering:
@dimbleby Thanks for finding that out.

@danyeaw Can you adopt the change from #9534 to see which warnings still need to be addressed, please?

.github/workflows/tests.yaml
```diff
@@ -13,6 +13,9 @@ defaults:
   run:
     shell: bash
 
+env:
+  PYTHONWARNDEFAULTENCODING: 'true'
```
dimbleby, line 17: Remove this one?

danyeaw: Thanks, done!

danyeaw:
Is this one ready to merge now?

radoering:
> Is this one ready to merge now?

We cannot do much about warnings in `site-packages` but there are still warnings that come from code in this repo. We should fix all warnings that come directly from this repo before enabling encoding warnings in the main branch. Please also take a look at the test runs for the different operating systems. Some warnings only show up on one of them.

dimbleby:
@danyeaw are you intending to pick off the remaining warnings?

If not then I suggest that this is a case where three quarters of a fix is better than no fix at all, might as well merge anyway.

danyeaw:
@dimbleby, I'm probably done working on this one to be honest. I think this gets things most of the way there, and eventually Poetry could adopt the Ruff PLW1514 rule https://docs.astral.sh/ruff/rules/unspecified-encoding/, and remove the encoding warning environmental variable.

radoering:
I will take a look at it next week and probably either deactivate the warning only merging the fixes or fix the remaining warnings.

radoering:
I think all warnings in the poetry code base are fixed now. In other words, the only existing warnings are in site packages.

@dimbleby Can you please take a look at the last commit? If you do not spot a bug I will merge this.

dimbleby:
Looks like that should resolve the warnings, without changing the behaviour - right?

Seems sensible.  I find it more than confusing to work out what the "correct" values are supposed to be!  But I don't recall any recent issues relating to funky locales and non-ascii paths - so let's not disturb the balance.

radoering (approved):

danyeaw:
Hi @radoering, thanks for the teamwork on this.

Issue #8902, opened 2024-01-23
Console scripts dependant on extras is officially discouraged


couling:
- [x] I have searched the [issues](https://github.com/python-poetry/poetry/issues) of this repo and believe that this is not a duplicate.

### Context

Console scripts dependent on extras is [officially discouraged](https://packaging.python.org/en/latest/specifications/entry-points/#data-model) and may not work in the wider eco python system:

> Using extras for an entry point is no longer recommended. Consumers should support parsing them from existing distributions, but may then ignore them. New publishing tools need not support specifying extras. The functionality of handling extras was tied to setuptools’ model of managing ‘egg’ packages, but newer tools such as pip and virtualenv use a different model.

Poetry has a [documented feature](url) to support scripts dependent on extras:
> To specify a script that [depends on an extra](https://python-poetry.org/docs/pyproject/#extras), you may provide an entry as an inline table:
>
> ```python
> [tool.poetry.scripts]
> devtest = { reference = "mypackage:test.run_tests", extras = ["test"], type = "console" }
> ```

When people have raised issues around this feature, it's been flagged that using the feature may be inadvisable (examples: [here](https://github.com/python-poetry/poetry/issues/8899#issuecomment-1906031337) and [here](https://github.com/python-poetry/poetry/issues/6892#issuecomment-1295924362)).  It seems odd to warn people that a documented feature is discouraged only when they raise an issue that the feature is broken in some way.

### Problem & Suggestion

Poetry documentation is suggesting something that is discouraged in the wider Python eco-system.  It would make sense to either:
 - Add a documentation warning with external documentation link to highlight that the feature is discouraged
 - Remove the documentation on the feature entirely. Perhaps this would require a decision to deprecate and remove the feature too.

dimbleby:
please submit a merge request!

PR #8903, opened 2024-01-24
Remove reference to old scripts extras feature

Secrus:
Resolves: #8902

### DIFF
--- a/docs/pyproject.md
+++ b/docs/pyproject.md
@@ -349,13 +349,6 @@ my_package_cli = 'my_package.console:run'
 
 Here, we will have the `my_package_cli` script installed which will execute the `run` function in the `console` module in the `my_package` package.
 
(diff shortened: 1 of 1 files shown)


radoering:
We probably should print a deprecation warning somewhere as proposed in the issue.

Secrus:
> We probably should print a deprecation warning somewhere as proposed in the issue.

Yeah, that would be a separate PR in core, as that's where the code is.

radoering:
Do you want to create that PR? I feel like we should undocument and deprecate it at the same time. It don't like the idea to make this an undocumented but not yet deprecated feature.

radoering:
I just created python-poetry/poetry-core#690 so we might include it in the next release.

radoering (approved):

PR #8910, opened 2024-01-26
Allow disabling keyring

Popkornium18:
This PR allows disabling the keyring completely. I have found myself often in situations where the keyring caused a lot of trouble. Checking whether the keyring was available would cause poetry to hang indefinetely. Rather than fixing broken keyring setups on various Linux systems I would much rather just disable the keyring altogether and accept the risk of plaintext credentials.


src/poetry/utils/password_manager.py
```diff
@@ -240,3 +230,17 @@ def delete_http_password(self, name: str) -> None:
...
+
+    def get_credential(
+        self, *names: str, username: str | None = None
+    ) -> HTTPAuthCredential:
+        if self.use_keyring:
+            return self.keyring.get_credential(*names, username=username)
+        else:
+            return HTTPAuthCredential(username=username, password=None)
+
+    def get_password(self, name: str, username: str) -> str | None:
```
radoering, line 242: This method is unused or did I miss something?

docs/configuration.md
```diff
@@ -506,3 +506,13 @@ repository.
 Set client certificate for repository `<name>`.
 See [Repositories - Configuring credentials - Custom certificate authority]({{< relref "repositories#custom-certificate-authority-and-mutual-tls-authentication" >}})
 for more information.
+
+### `keyring.enabled`:
+
+**Type**: `boolean`
+
+**Default**: `true`
+
```
radoering, line 515: "Environment variable" is missing.

docs/faq.md
```diff
@@ -154,6 +154,15 @@ required variables explicitly or `passenv = "*"` to forward all of them.
 Linux systems may require forwarding the `DBUS_SESSION_BUS_ADDRESS` variable to allow access to the system keyring,
 though this may vary between desktop environments.
 
+Alternatively you can disable the keyring completely:
```
radoering, line 157:
```suggestion
Alternatively, you can disable the keyring completely:
```

docs/repositories.md
```diff
@@ -500,6 +500,12 @@ If a system keyring is available and supported, the password is stored to and re
 
 Keyring support is enabled using the [keyring library](https://pypi.org/project/keyring/). For more information on supported backends refer to the [library documentation](https://keyring.readthedocs.io/en/latest/?badge=latest).
 
+If you do not want to use the keyring you can tell Poetry to disable it and store the credentials in plaintext config files:
```
radoering, line 503:
```suggestion
If you do not want to use the keyring, you can tell Poetry to disable it and store the credentials in plaintext config files:
```

radoering (changes requested):
Looks good in general. I wonder if we should add some kind of tests in `password_manager` to check that keyring is not used when calling certain methods or if this does not add much value?

tests/utils/test_password_manager.py
```diff
@@ -330,3 +330,14 @@ def test_get_pypi_token_with_env_var_not_available(
     result_token = manager.get_pypi_token(repo_name)
 
     assert result_token is None
+
+
+def test_disabled_keyring_is_unavailable(
```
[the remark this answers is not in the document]
Popkornium18: I'll move it

src/poetry/utils/password_manager.py
Popkornium18: Yes, i removed it

Popkornium18:
I made all the changes you requestet @radoering.

> Looks good in general. I wonder if we should add some kind of tests in password_manager to check that keyring is not used when calling certain methods or if this does not add much value?

What certain methods would this be?

radoering:
It's probably just like "when `use_keyring` is not set, then none of the methods of `PasswordManager` tries to access keyring. I can imagine a test that replaces keyring with a mock, then calls each method and afterwards checks that the mock was never called.

Popkornium18:
> It's probably just like "when `use_keyring` is not set, then none of the methods of `PasswordManager` tries to access keyring. I can imagine a test that replaces keyring with a mock, then calls each method and afterwards checks that the mock was never called.

I implemented such a test and actually found an issue with it, so I guess it is necessary. Not sure what you think of the attempt to detect code-changes that were not reflected in the test. Looks weird, but I think it might be worth it. Better to have one more place to make a change than a bug.

tests/utils/test_password_manager.py
```diff
@@ -330,3 +330,32 @@ def test_get_pypi_token_with_env_var_not_available(
...
+    config.config["http-basic"] = {"onlyuser": {"username": "user"}}
+
+    manager = PasswordManager(config)
+    num_public_functions = len([f for f in dir(manager) if not f.startswith("_")])
+    if num_public_functions != 10:
+        pytest.fail(
+            f"A function was added to or removed from the {PasswordManager.__name__} "
+            "class without reflecting this change in this test."
+        )
+
```
radoering, line 348: Maybe, we can add the following line in this test, too:

```
with pytest.raises(PoetryKeyringError):
        _ = manager.keyring
```

and then add a comment like "No method must try to access keyring. If it does a PoetryKeyringError will be raised." to make clear why there is no assert in the following lines.

radoering (changes requested):
> I implemented such a test and actually found an issue with it, so I guess it is necessary.

Nice.

> Not sure what you think of the attempt to detect code-changes that were not reflected in the test. Looks weird, but I think it might be worth it. Better to have one more place to make a change than a bug.

I completely agree. I don't think it's an issue that the unit test of `password_manager` has to be adapted if `password_manager` is changed.

The test looks good to me. Just a small suggestion to make the idea behind the test more clear.

tests/utils/test_password_manager.py
Popkornium18: If I add this to this test, I would remove the other test, because config test just becomes a subset of this test.
What about a comment that states that the test in test_config makes sure that access `.keyring` raises an error and we just check that accessing the important functions of the password manager don't raise an exception.

radoering: I would probably even do it the other way around and check for the exception only in `test_password_manager`. The goal of `test_config` is only to check that the config option is propagated correctly. It's sufficient that `use_keyring` is `False` before and `True` after setting the option. All implications of `use_keyring` being `False` should be checked in `test_password_manager`.

Apart from that, I think even if there is some duplication, we should keep both tests because it corresponds to our (unwritten) test concept. If we introduce a new config option:

- We want a test in `test_config` that just checks that the option is propagated correctly. Other effects of the option do not matter here.
- We want a test of the changed behavior in the unit test of the influenced module. In this test it does not matter how the behavior is triggered (via config option or directly if there is an attribute that can be set). Just setting `password_manager.use_keyring = False` would be fine if that's sufficient. (Not saying you should do this, it just does not matter.)

Popkornium18: Ok, I changed it

radoering (approved):

PR #8921, opened 2024-01-29
Change docs to use `--only main` syntax, and add missing argument to Dockerfile example

shawnz:
This PR describes a suggestion which I believe improves upon two small issues I have with the docs regarding the `--without dev`/`--only main` option for the `poetry install` command.

1) The Commands page disagrees with the deprecation warning when using the deprecated `--no-dev` option: the Commands page suggests using `--without dev` instead, but the deprecation warning suggests using `--only main` here: https://github.com/python-poetry/poetry/blob/ed1dcddebb6337997829df5b6c73b020cdcdee99/src/poetry/console/commands/group_command.py#L86

    To resolve this, I updated the Commands page to suggest both options. Personally, I think `--only main` is preferrable overall because it works even if the dev dependency group doesn't exist, but it might be helpful for readers to know that both options are viable.

2) The FAQ entry about avoiding Docker cache busting doesn't apply either of the `--without dev` or `--only main` options to the first install statement, but rather only the second install statement. Based on my testing, this causes dev dependencies to get installed, which I think is not intended.

    To resolve this I added `--only main` to the first install statement, and also changed the second install statement to use the `--only main` syntax rather than the `--without dev` syntax for the reasons described above.

### DIFF
--- a/docs/cli.md
+++ b/docs/cli.md
@@ -153,7 +153,7 @@ poetry install --without test,docs
 ```
 
 {{% note %}}
-The `--no-dev` option is now deprecated. You should use the `--without dev` notation instead.
+The `--no-dev` option is now deprecated. You should use the `--without dev` or `--only main` notation instead.
 {{% /note %}}
 
 You can also select optional dependency groups with the `--with` option.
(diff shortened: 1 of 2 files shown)


shawnz:
Based on an initial positive response that `--only main` is preferable here, I just added a follow up commit which swaps the order of all the examples so that `--only main` is suggested first before `--without dev`

PR #8927, opened 2024-01-30
Update managing-environments.md to have Poetry init before Poetry install

AFirooz:
Add poetry initiation

# Pull Request Check List

Resolves: #issue-number-here


dimbleby:
I dont think this change makes sense, this section of the doc is describing work on an existing project.

`poetry init` and then immediately `poetry install` almost certainly isnt what anyone wants to do anyway, they likely want to add packages, write a README, etc

radoering:
> I dont think this change makes sense, this section of the doc is describing work on an existing project.

I completely agree with this.

Nevertheless, thank you for your contribution.

PR #8947, opened 2024-02-05
Correct tests for handling of paths to files where paths to directories are expected

Cypher1:
I think these tests are testing slightly incorrect behaviour. Fixing them should unblock another change I'd like to suggest but we should also ensure we aren't testing behaviours that aren't intended.


radoering:
Good catch. Did you notice that there is a third test that failed in https://github.com/python-poetry/poetry-core/actions/runs/7782092252/job/21217798889?pr=695? It looks like `with_empty_repositories_key` does not exist at all. We probably can just use another fixture without repositories.

Cypher1:
Ah! Thanks for the catch. I went to bed before I could see the results.

Cypher1:
I thought it wise to add that missing fixture, could you check that I've understood the test's intent?

In particular: I used the local_config fixture as a starting point, so I might have missed something that should be removed from that.

Cypher1:
@radoering I'm wondering if that test actually needs to exist. The empty repository key that it checks is enforced by the schema validation in Factory.

I've updated it to check that schema validation is performed, but this may be redundant or the behaviour may be wrong (as the previously tested behaviour may be desired). Or perhaps the 'empty_repository_key' actually meant an unset repository key and I've gone in the wrong direction.

Thanks in advance

radoering:
The test was introduced in #7910 so the goal was to increase test coverage. If I remove the test, the line coverage for `config.py` decreases from 95% to 93%, so I think it makes sense to have such a test. The test is not about [repository](https://python-poetry.org/docs/pyproject/#repository) but about [repositories](https://python-poetry.org/docs/repositories/#publishable-repositories) and especially just tests the command if there are no repositories. (That's why the tests works if some abitrary sample project without repositories is chosen.) Thus, you took the wrong turn.

radoering (approved):

PR #8953, opened 2024-02-07
Fix documentation command for adding a supplemental source 

alexkiro:
The command requires both a name and an URL

### DIFF
--- a/docs/repositories.md
+++ b/docs/repositories.md
@@ -247,7 +247,7 @@ You can configure a package source as a supplemental source with `priority = "su
 source configuration.
 
 ```bash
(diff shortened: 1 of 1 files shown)


docs/repositories.md
```diff
@@ -247,7 +247,7 @@ You can configure a package source as a supplemental source with `priority = "su
 source configuration.
 
 ```bash
-poetry source add --priority=supplemental https://foo.bar/simple/
+poetry source add --priority=supplemental foo https://foo.bar/simple/
```
radoering, line 250: There is the same issue in line 229, isn't it?

radoering (approved):

PR #8961, opened 2024-02-10
Add a dir generated by `pytest` to .gitignore

ilyagr:
I'm guessing we don't want it in the repo, though adding
it would be an alternative.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -31,6 +31,7 @@ MANIFEST.in
 /docs/site/*
 /tests/fixtures/simple_project/setup.py
 /tests/fixtures/project_with_extras/setup.py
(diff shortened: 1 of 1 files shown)


radoering:
It would be even better if it was not created in the tests fixtures directory in the first place because it could influence subsequent calls of the test. In other words, the test should create a copy of the fixture in tmp and work on that copy. This should be the culprit: https://github.com/python-poetry/poetry/blob/5f1d1a797b2e016bc643f40d40b43a9cdbee7448/tests/installation/test_chef.py#L174-L182

ilyagr:
Thanks for looking at this! I might or might not get to fixing this further at some point.

Secrus:
I am going to close this since the actual change that has to be done is completely different.

Issue #8965, opened 2024-02-12
`poetry-core` does not install local dependency in editable mode


bersbersbers:
- **Poetry version**: poetry_core-1.9.0-py3-none-any.whl
- **Python version**: 3.12.2
- **OS version and name**: Windows 11 22H2

- [x] I am on the [latest](https://github.com/python-poetry/poetry/releases/latest) stable Poetry version, installed using a [recommended method](https://python-poetry.org/docs/#installation).
- [x] I have searched the [issues](https://github.com/python-poetry/poetry/issues) of this repo and believe that this is not a duplicate.
- [x] I have consulted the [FAQ](https://python-poetry.org/docs/faq/) and [blog](https://python-poetry.org/blog/) for any relevant entries or release notes.
- [x] If an exception occurs when executing a command, I executed it again in debug mode (`-vvv` option) and have included the output below.

## Issue

```
C:\Code\MyTestParent
│   pyproject.toml
│   TestParent.py
└───MyTestChild
        pyproject.toml
        TestChild.py
```

Both `.py` files are empty. The `pyproject.toml` files are:

`MyTestParent`:
```toml
[build-system]
requires = ["poetry-core>=1.2.0"]
build-backend = "poetry.core.masonry.api"

[tool.poetry]
name = "TestParent"
version = "0.0.0"
description = ""
authors = []

[tool.poetry.dependencies]
python = ">=3.12"
TestChild = { path = "MyTestChild", develop = true }
```

`MyTestChild`:
```toml
[build-system]
requires = ["poetry-core>=1.2.0"]
build-backend = "poetry.core.masonry.api"

[tool.poetry]
name = "TestChild"
version = "0.0.0"
description = ""
authors = []

[tool.poetry.dependencies]
python = ">=3.12"
```

Now, I can install `TestParent` in editable mode, but `TestChild` is not installed in editable mode:

```
$ pip install -e . && (pip list | findstr Test)
TestChild  0.0.0
TestParent 0.0.0   C:\Code\MyTestParent
```

dimbleby:
there is no way to express in `TestParent`'s metadata that `TestChild` is editable, this is expected

please close

bersbersbers:
> there is no way to express in `TestParent`'s metadata that `TestChild` is editable

I think I don't get it. What is the purpose of the `develop` key in `TestParent`s `TestChild` dependency then, if not to make it editable?

dimbleby:
[...]

PR #8971, opened 2024-02-14
Docs: state a limitation of `develop` attribute

bersbersbers:
Resolves: #8965


dimbleby:
I would guess that this will confuse more people than it helps!

radoering (approved):

PR #8984, opened 2024-02-17
chore: update dependencies

radoering:


dimbleby:
fyi I have https://github.com/jaraco/keyring/pull/666 and https://github.com/jaraco/jaraco.classes/pull/13 open trying to fix up the faulty types in those two projects.  I wouldn't wait on them though: in poetry it only makes a difference to type annotations in unit test so certainly not worth holding up a release

Not sure about the changes in site_packages.py, if importlib-metadata doesn't promise to return a `Path` then I'd have thought the code was fine as it was?

(`find_distribution_files_with_suffix()` seems to be unused so that could be removed altogether, that's one less to worry about...)

radoering:
> Not sure about the changes in site_packages.py, if importlib-metadata doesn't promise to return a `Path` then I'd have thought the code was fine as it was?

`importlib-metadata` promises to return a [`SimplePath`](https://github.com/python/importlib_metadata/blob/v7.0.1/importlib_metadata/_meta.py#L50), which is a `Protocol` that just provides a subset of `Path`. According to mypy we cannnot create a `Path` from a `SimplePath` (that's probably true). Thus, the old code will only work if we get a real `Path` - or strictly speaking a `PathLike` or `str`. I assume it's fine to expect a real path since we are working with a venv. Or do you think it would be better to keep the old code and just add a `#type: ignore`?

dimbleby:
Ah, different python versions - this returns `Pathlike[str]` according to typeshed.

I'd think the right thing to do is to raise a pull request at importlib-metadata, changing the return type to be a plain `Path`.

Presumably they are currently trying to avoid committing to a particular type, but hopefully should be open to agreeing that in fact returning their own odd half-`Path` is unhelpful for callers.

(Meanwhile if you are looking to cut a release... do whatever)

dimbleby:
ugh, though https://github.com/python/importlib_metadata/pull/480

> in fact could be a `zipfile.Path` or other `SimplePath`

but I think this over-shot: it makes sense for the abstract `Distribution` to commit only to an abstract `Path`-ish, but a `PathDistribution` really should return a concrete `Path`

Secrus (approved):

Issue #8987, opened 2024-02-18
Time to deprecate the non-modern installer?


dimbleby:
I'm not seeing any issue reports complaining about the modern installer.  It would be good to remove it from the codebase.

Suggest it's about time to deprecate its use, probably with a pointer to an issue in this repository that says something along the lines of: poetry intends to remove the pip-based installer, if you have a compelling reason for using it then please open an issue and we can talk about it: but you should expect that it is going away.

abn:
This is done now. Since version 1.8.0 configuring `installer.modern-installation=false` is deprecated. Let's aim to remove it after the next minor release.

PR #8988, opened 2024-02-18
deprecate "installer.modern-installation false"

dimbleby:
As proposed at #8987

I started with `warnings.warn` but something about all the cleo io cleverness results in those messages disappearing.  Anyway this deprecation is more user-facing than developer-facing, so regular warning output is probably fine.

I'd be happy to see this deprecation in the upcoming release, might as well get it out there and see who squeals.  But no big deal if it misses the cut.

### DIFF
--- a/src/poetry/installation/executor.py
+++ b/src/poetry/installation/executor.py
@@ -70,7 +70,7 @@ class Executor:
         )
         if not self._use_modern_installation:
             self._io.write_line(
-                "<warning>Setting `installer.modern-installation` to `false` "
+                "<warning>Warning: Setting `installer.modern-installation` to `false` "
                 "is deprecated.</>"
             )
             self._io.write_line(
--- a/src/poetry/installation/executor.py
+++ b/src/poetry/installation/executor.py
@@ -68,6 +68,17 @@ class Executor:
         self._use_modern_installation = config.get(
             "installer.modern-installation", True
         )
+        if not self._use_modern_installation:
+            self._io.write_line(
+                "<warning>Setting `installer.modern-installation` to `false` "
+                "is deprecated.</>"
+            )
+            self._io.write_line(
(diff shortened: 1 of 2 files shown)


radoering (dismissed):

src/poetry/installation/executor.py
```diff
@@ -68,6 +68,17 @@ def __init__(
         self._use_modern_installation = config.get(
             "installer.modern-installation", True
         )
+        if not self._use_modern_installation:
+            self._io.write_line(
```
radoering, line 72: One thing: Shouldn't we use `write_error_line()` for warnings?

dimbleby: some do, some don't: if there is a rule then it is not consistently followed.  The stdout from these commands is not really suitable for parsing anyway so I don't suppose it much matters.

I'm not at the right computer to do anything about this now, feel free do any of tweak it yourself / ask me to do it later / not bother.

radoering: You are probably right that it is inconsistent and does not matter.

Interestingly enough, I see random test failures on Windows if I write the warning to error out. It's always a PermissionError in https://github.com/python-poetry/poetry/blob/0eedea4e1411667b525c234a363f8b7eb5e8fc4b/tests/conftest.py#L356

but in different tests, for example one or two of

```
tests/console/commands/self/test_add_plugins.py::test_add_with_constraint
tests/console/commands/self/test_add_plugins.py::test_add_with_git_constraint_with_extras
tests/console/commands/self/test_remove_plugins.py::test_remove_installed_package
tests/console/commands/self/test_remove_plugins.py::test_remove_installed_package_dry_run
tests/installation/test_executor.py::test_executor_should_write_pep610_url_references_for_editable_git
```

Maybe, a reason to just keep it as is, even if this is a bit unsettling...

Issue #8991, opened 2024-02-19
Dependency specification: inaccurate description of caret requirement


TanguyP:
- [x] I have searched the [issues](https://github.com/python-poetry/poetry/issues) of this repo and believe that this is not a duplicate.

## Issue
On the _Dependency specification_ page, the [Caret requirements](https://python-poetry.org/docs/dependency-specification#caret-requirements) section reads:
> An update is allowed if the new version number does not modify the left-most non-zero digit in the major, minor, patch grouping.

But this seems partly inaccurate, as it means that specifying `^0.2.3` as a requirement allows version `2.2.3` which has the same left-most non-zero digit i.e. 2 - but that's clearly not the case, as illustrated in the examples table further down, which mentions that a requirement of `^0.2.3` will allow versions `>=0.2.3 <0.3.0`.

Or maybe the person who wrote the above sentence meant that the digit _and position_ of the left-most non-zero digit shouldn't change? In that case, I would argue that the phrasing can be improved.

dimbleby:
The misunderstanding that you propose feels unlikely!

but if you have ideas to improve the docs then - by a very large margin - the best way to do that is to submit a merge request with the change you would like to see.

ulgens:
I think the term we need here is "significant figures" https://en.wikipedia.org/wiki/Significant_figures "0" means different things for the logic if it's in "012.34" or "0.1234", but the referenced part of the doc mentions it like there is only one case.

TanguyP:
> I think the term we need here is "significant figures" https://en.wikipedia.org/wiki/Significant_figures "0" means different things for the logic if it's in "012.34" or "0.1234", but the referenced part of the doc mentions it like there is only one case.

Thanks! But I believe this comment belongs more in the PR, would you mind copying/moving it there (if you have sufficient permissions)? Then we can have a more extended discussion with a clearer format :)

abn:
I think this is overkill and suggestion in #8992 does not really make much of a dent in improving the status quo imho. I appreciate that the current version can be confusing, but I do think it gets the point across just fine.

And further, I think any concerns raised about the "confusion" is mitigated by the additional examples provided in the same line.

>  If instead we had specified the version string as `^0.1.13`, poetry would update to `0.1.14` but not `0.2.0`. `0.0.x` is not considered compatible with any other version.

ulgens:
@radoering Can you please add a short comment why this got closed? Thanks in advance.

radoering:
[...]

PR #8992, opened 2024-02-19
docs: Dependency specification: improve description of caret requirement

TanguyP:
I'm open to suggestions for improving the wording still, but I believe this version already improves the current documentation. I've included both a succinct ("the less significant numbers") and a more formal definition.

# Pull Request Check List

Resolves: #8991

- [ ] Added **tests** for changed code. _(not applicable)_
- [ ] Updated **documentation** for changed code. _(not applicable)_

### DIFF
--- a/docs/dependency-specification.md
+++ b/docs/dependency-specification.md
@@ -18,7 +18,7 @@ of the dependency and on the optional constraints that might be needed for it to
 
 ### Caret requirements
 
-**Caret requirements** allow [SemVer](https://semver.org/) compatible updates to a specified version. An update is allowed if the new version number does not modify the left-most non-zero digit in the major, minor, patch grouping. For instance, if we previously ran `poetry add requests@^2.13.0` and wanted to update the library and ran `poetry update requests`, poetry would update us to version `2.14.0` if it was available, but would not update us to `3.0.0`. If instead we had specified the version string as `^0.1.13`, poetry would update to `0.1.14` but not `0.2.0`. `0.0.x` is not considered compatible with any other version.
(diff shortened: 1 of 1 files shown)


docs/dependency-specification.md
```diff
@@ -18,7 +18,7 @@ of the dependency and on the optional constraints that might be needed for it to
...
+**Caret requirements** allow [SemVer](https://semver.org/) compatible updates to a specified version. An update is allowed if the new version only modifies the less significant numbers (i.e. numbers to the right of the left-most non-zero number) in the major, minor, patch grouping. For instance, if we previously ran `poetry add requests@^2.13.0` and wanted to update the library and ran `poetry update requests`, poetry would update us to version `2.14.0` if it was available, but would not update us to `3.0.0`. If instead we had specified the version string as `^0.1.13`, poetry would update to `0.1.14` but not `0.2.0`. `0.0.x` is not considered compatible with any other version.
```
radoering, line 21: Is "less significant" a common term? (I know "least significant" but haven't heard "less significant" before.)

Alternatively, we could just write "... only modifies the numbers to the right ...".

TanguyP: For what it's worth, I've done some Google searches: "less significant numbers" yields ~38k results, "least significant number" ~61k, i.e. the former is used less but both figures have the same order of magnitude.

Strictly speaking, using "less significant numbers" on its own is incorrect, because I'm not mentioning less significant _than what_. However, the rationale for this phrasing is that "numbers to the right of the left-most non-zero number" is quite a mouthful that most people will need to read at least twice before understanding it; so having a less precise, more intuitive explanation first can help quickly grasp the idea before getting into the technical details.

What do you think?

radoering:
Taking https://github.com/python-poetry/poetry/issues/8991#issuecomment-1974079665 into account I'll close this since I am not very convinced myself. Even though the change might be more correct, I think it is more difficult to understand and as already mentioned in https://github.com/python-poetry/poetry/issues/8991#issuecomment-1952879853 the misunderstanding feels very unlikely.

Nevertheless, thanks for your efforts.

PR #9006, opened 2024-02-24
doc: add pytorch example for explicit source

abn:
Closes: #8246

This is a rework of #8246.


radoering:
I think the original issue was that there is no hint about explicit sources in https://python-poetry.org/docs/cli/#add so we should at least add a sentence and a link to https://python-poetry.org/docs/repositories/#explicit-package-sources. Otherwise LGTM.

Secrus (approved):

Secrus (approved):

PR #9019, opened 2024-02-25
Trust the setup reader when it finds no requirements

dimbleby:
follow up to #9000 

the setup reader should now be failing if it is not confident about the answers that it is giving: so when it says that a package has no requirements we can believe it rather than assuming that we need to build that package to make sure.

gives a nice speed-up in the cases where it happens eg `petname` is a package out in the wild that has the necessary properties if you want to try it: on my computer `poetry add --lock --no-cache petname` goes from about 8 seconds to about 1 second

### DIFF
--- a/src/poetry/inspection/info.py
+++ b/src/poetry/inspection/info.py
@@ -369,16 +369,9 @@ class PackageInfo:
             name=result.get("name"),
             version=result.get("version"),
             summary=result.get("description", ""),
-            requires_dist=requirements or None,
+            requires_dist=requirements,
             requires_python=python_requires,
         )
-
-        if not (info.name and info.version) and not info.requires_dist:
-            # there is nothing useful here
-            raise PackageInfoError(
-                path,
-                "No core metadata (name, version, requires-dist) could be retrieved.",
-            )
 
         return info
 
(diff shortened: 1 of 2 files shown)


src/poetry/inspection/info.py
```diff
@@ -369,17 +369,10 @@ def from_setup_files(cls, path: Path) -> PackageInfo:
             name=result.get("name"),
             version=result.get("version"),
             summary=result.get("description", ""),
-            requires_dist=requirements or None,
+            requires_dist=requirements,
```
radoering, line 372: IIRC, this is written to the cache. It shouldn't matter if it's an empty list or None, should it?

dimbleby: maybe?  if it doesn't  matter then it doesn't matter: but an empty list seems a clearer representation of what we understand at this point, I don't see why this code would prefer `None`.

perhaps the field shouldn't even be allowed to be `None` on the `PackageInfo`.

abn: If i recall correct, the setup reader cannot be trusted when the `setup.py` files are complex. The `None` state was, if i recall correct, used to indicate whether we are confident enough to use the value or resolution must continue (see http/pypi repo code). 

If we are happy to say that we are confident the setup reader has provided the correct value, this should be okay and an empty list is probably more accurate.

It might be that we just need to someone factor in complexity of the setup file to use `[]` or `None`.

dimbleby: you probably want to catch up with #9000, which is intended to have arranged the setup reader so that either
- it is confident that it has recognised one of the ways to write setup.py that it does recognise
- or it throws a SetupReaderError (and we do not reach this branch)

abn: Awesome, yeah missed that. With that, I think the fix here is good.

abn (approved):

PR #9024, opened 2024-02-25
ci: enable Python 3.12 for FreeBSD

abn:


Secrus:
We are sunsetting BSD CI

Issue #9026, opened 2024-02-25
Installing private packages from GAR in Docker build fails with 500 Internal Server Error on Poetry 1.8.0


ntaylorwss:
- **Poetry version**: 1.8.0
- **Python version**: 3.12.0
- **OS version and name**: Docker image python@sha256:d0c9bf03d80a3c00963f1b973760094c0b2070638fa64dd4f67d2247c0110efc
- **pyproject.toml**: https://gist.github.com/ntaylorwss/69cc7b9836047c5f69fbd263191d1210

- [x] I am on the [latest](https://github.com/python-poetry/poetry/releases/latest) stable Poetry version, installed using a [recommended method](https://python-poetry.org/docs/#installation).
- [x] I have searched the [issues](https://github.com/python-poetry/poetry/issues) of this repo and believe that this is not a duplicate.
- [x] I have consulted the [FAQ](https://python-poetry.org/docs/faq/) and [blog](https://python-poetry.org/blog/) for any relevant entries or release notes.
- [x] If an exception occurs when executing a command, I executed it again in debug mode (`-vvv` option) and have included the output below.

## Issue

I have a Python service whose pyproject.toml includes some dependencies from a private Google Artifact Registry Python repository. Until yesterday, this all worked fine. Poetry is installed in the container via:

```
# install poetry
RUN apt-get update \
    && apt-get install -y curl gnupg gcc \
    && curl -sSL https://install.python-poetry.org | python - \
    && poetry self add keyrings.google-artifactregistry-auth
```

and this keyring, given an environment with active application credentials, is able to authenticate me to my GAR repository and install the required dependencies. A build ran this morning and failed with the following error:

```
14.39   ValueError
14.39
14.39   Package('my-package', '0.0.0') is not in list
14.39
14.39   at /opt/poetry/venv/lib/python3.12/site-packages/poetry/repositories/legacy_repository.py:66 in package
14.39        62│         Note that this will be cached so the subsequent operations
14.39        63│         should be much faster.
14.39        64│         """
14.39        65│         try:
14.39     →  66│             index = self._packages.index(Package(name, version))
14.39        67│
14.39        68│             return self._packages[index]
14.39        69│         except ValueError:
14.39        70│             package = super().package(name, version, extras)
14.39
14.39 The following error occurred when trying to handle this error:
14.39
14.39
14.39   Stack trace:
14.39
14.39   34  /opt/poetry/venv/lib/python3.12/site-packages/cleo/application.py:327 in run
[...]

PR #9030, opened 2024-02-26
lazy-wheel: handle internal server error and others

radoering:
if the server returns an internal server error (500), it probably does not support negative offsets

# Pull Request Check List

Resolves: #9026 #9030

### DIFF
--- a/src/poetry/inspection/lazy_wheel.py
+++ b/src/poetry/inspection/lazy_wheel.py
@@ -608,22 +608,10 @@ class LazyWheelOverHTTP(LazyFileOverHTTP):
             # Initial range request for just the end of the file.
             file_length, tail = self._try_initial_chunk_request(initial_chunk_size)
         except HTTPError as e:
+            # Our initial request using a negative byte range was not supported.
             resp = e.response
             code = resp.status_code if resp is not None else None
-            # Our initial request using a negative byte range was not supported.
-            if code in [codes.not_implemented, codes.method_not_allowed]:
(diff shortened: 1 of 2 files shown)


Secrus (dismissed):

src/poetry/inspection/lazy_wheel.py
```diff
@@ -611,13 +611,18 @@ def _extract_content_length(
             resp = e.response
             code = resp.status_code if resp is not None else None
             # Our initial request using a negative byte range was not supported.
-            if code in [codes.not_implemented, codes.method_not_allowed]:
+            if code in {
```
abn, line 614: Should/can this be inverted? ie. `if code not in { ...}`

Secrus:
Take a look at #9033, maybe we can fix that as well?

src/poetry/inspection/lazy_wheel.py
abn: Yeah. Given we cannot guarantee good behavior, it might be safer.

abn (approved):

david-nano:
@abn  what version will include this fix?

abn:
@david-nano we are hoping we can get a 1.8.1 release out sooner rather than later. Depends on how much time we are able to find.

Symbianx:
We also just found this problem and it's breaking all service builds. Is there any chance we can get a patch released soon?

radoering:
#9036

Issue #2117, opened 2020-03-02
poetry env use X.Y fails on Windows


absassi:
- [x] I am on the [latest](https://github.com/python-poetry/poetry/releases/latest) Poetry version.
- [x] I have searched the [issues](https://github.com/python-poetry/poetry/issues) of this repo and believe that this is not a duplicate.
- [ ] If an exception occurs when executing a command, I executed it again in debug mode (`-vvv` option).

- **OS version and name**: Windows 10
- **Poetry version**: 1.0.5
- **Link of a [Gist](https://gist.github.com/) with the contents of your pyproject.toml file**: not needed

## Issue

The shortcut command `poetry env use X.Y` to select the environment for Python X.Y, as documented in [Managing environments](https://python-poetry.org/docs/managing-environments/#switching-between-environments), is broken in Windows. One is required instead to find, copy and paste the full path to Python executable.

Note that the documentation even says "especially Windows where pyenv is not available".

```
>poetry env use 3.7
'python3.7' is not recognized as an internal or external command,
operable program or batch file.
```

The reason is quite simple: CPython installation in Windows doesn't provide `pythonX.Y.exe`, so even if you have all of them in `PATH`, Poetry can only find the first one. They do, however, provide registry keys with installation locations, which Poetry can use to resolve `X.Y` to the Python interpreter path.

This piece of code should find the path to Python, given the version number as a `X.Y` string:

    import os
    try:
        import winreg
    except ImportError:
        import _winreg as winreg

    def find_python_interpreter(version):
        for hkey, prefix in (
            (winreg.HKEY_CURRENT_USER, "Software"),
            (winreg.HKEY_CURRENT_USER, "Software\\WOW6432Node"),
            (winreg.HKEY_LOCAL_MACHINE, "Software"),
            (winreg.HKEY_LOCAL_MACHINE, "Software\\WOW6432Node"),
        ):
            try:
                value = winreg.QueryValue(hkey, "{}\\Python\\PythonCore\\{}\\InstallPath".format(prefix, version))
            except FileNotFoundError:
                continue
            python = os.path.join(value, "python.exe")
            if os.path.exists(python):
                return python
        return None

*I have adjusted the code above to make it compatible with Python 2, but I have not tested afterwards, so please forgive any mistakes.*

konserw:
I've also been hit by this bug. Can this proposed solution be merged?

shelper:
same here, it has been  over a year, seems still not been fixed...

[...]

PR #9050, opened 2024-02-27
refactor(env_manager): split out python detection

finswimmer:
At the moment the `env_manager` is more or less a big ball of mud, mainly due to violating the single responsible principle. Beside the original purpose - handling environments - it also includes lot of logic to find the desired python interpreter.

This PR refactor the env_manager by split out the logic for finding the path to the desired python interpreter.

### DIFF
--- a/src/poetry/utils/env/python_manager.py
+++ b/src/poetry/utils/env/python_manager.py
@@ -11,8 +11,10 @@ from typing import TYPE_CHECKING
 from cleo.io.null_io import NullIO
 from cleo.io.outputs.output import Verbosity
 from poetry.core.constraints.version import Version
+from poetry.core.constraints.version import parse_constraint
 
 from poetry.utils._compat import decode
+from poetry.utils.env.exceptions import NoCompatiblePythonVersionFound
 from poetry.utils.env.script_strings import GET_PYTHON_VERSION_ONELINER
 
 
@@ -20,6 +22,7 @@ if TYPE_CHECKING:
     from cleo.io.io import IO
 
     from poetry.config.config import Config
+    from poetry.poetry import Poetry
 
 
 class Python:
@@ -119,3 +122,46 @@ class Python:
             return cls(executable=active_python)
 
         return cls.get_system_python()
+
+    @classmethod
+    def get_compatible_python(cls, poetry: Poetry, io: IO | None = None) -> Python:
+        io = io or NullIO()
+        supported_python = poetry.package.python_constraint
+        python = None
+
+        for suffix in [
+            *sorted(
+                poetry.package.AVAILABLE_PYTHONS,
+                key=lambda v: (v.startswith("3"), -len(v), v),
+                reverse=True,
+            ),
+            "",
+        ]:
+            if len(suffix) == 1:
+                if not parse_constraint(f"^{suffix}.0").allows_any(supported_python):
+                    continue
+            elif suffix and not supported_python.allows_any(
+                parse_constraint(suffix + ".*")
+            ):
+                continue
+
+            python_name = f"python{suffix}"
+            if io.is_debug():
+                io.write_error_line(f"<debug>Trying {python_name}</debug>")
+
+            executable = cls._full_python_path(python_name)
+            if executable is None:
+                continue
+
+            candidate = cls(executable)
+            if supported_python.allows(candidate.patch_version):
+                python = candidate
+                io.write_error_line(
+                    f"Using <c1>{python_name}</c1> ({python.patch_version})"
+                )
+                break
+
+        if not python:
+            raise NoCompatiblePythonVersionFound(poetry.package.python_versions)
+
+        return python
--- a/src/poetry/utils/env/python_manager.py
+++ b/src/poetry/utils/env/python_manager.py
@@ -102,6 +102,14 @@ class Python:
         return cls(executable=sys.executable)
 
     @classmethod
+    def get_by_name(cls, python_name: str) -> Python | None:
+        executable = cls._full_python_path(python_name)
+        if not executable:
+            return None
+
+        return cls(executable=executable)
+
+    @classmethod
     def get_preferred_python(cls, config: Config, io: IO | None = None) -> Python:
         io = io or NullIO()
 
--- a/src/poetry/utils/env/python_manager.py
+++ b/src/poetry/utils/env/python_manager.py
@@ -46,3 +46,7 @@ class Python:
     @cached_property
     def minor_version(self) -> Version:
         return Version.from_parts(major=self.version.major, minor=self.version.minor)
+
+    @classmethod
+    def get_system_python(cls) -> Python:
+        return cls(executable=sys.executable)
--- a/src/poetry/utils/env/python_manager.py
+++ b/src/poetry/utils/env/python_manager.py
@@ -1,0 +1,48 @@
+from __future__ import annotations
+
+import subprocess
+import sys
+
+from functools import cached_property
+from pathlib import Path
+
+from poetry.core.constraints.version import Version
+
+from poetry.utils._compat import decode
+from poetry.utils.env.script_strings import GET_PYTHON_VERSION_ONELINER
+
+
+class Python:
+    def __init__(self, executable: str | Path, version: Version | None = None) -> None:
+        self.executable = Path(executable)
+        self._version = version
+
+    @property
+    def version(self) -> Version:
+        if not self._version:
+            if self.executable == Path(sys.executable):
+                python_version = ".".join(str(v) for v in sys.version_info[:3])
+            else:
+                encoding = "locale" if sys.version_info >= (3, 10) else None
+                python_version = decode(
+                    subprocess.check_output(
+                        [str(self.executable), "-c", GET_PYTHON_VERSION_ONELINER],
+                        text=True,
+                        encoding=encoding,
+                    ).strip()
+                )
+            self._version = Version.parse(python_version)
+
+        return self._version
+
+    @cached_property
+    def patch_version(self) -> Version:
+        return Version.from_parts(
+            major=self.version.major,
+            minor=self.version.minor,
+            patch=self.version.patch,
+        )
+
+    @cached_property
+    def minor_version(self) -> Version:
+        return Version.from_parts(major=self.version.major, minor=self.version.minor)
--- a/src/poetry/utils/env/python_manager.py
+++ b/src/poetry/utils/env/python_manager.py
@@ -1,15 +1,25 @@
 from __future__ import annotations
 
+import shutil
 import subprocess
 import sys
 
 from functools import cached_property
 from pathlib import Path
+from typing import TYPE_CHECKING
 
+from cleo.io.null_io import NullIO
+from cleo.io.outputs.output import Verbosity
 from poetry.core.constraints.version import Version
 
 from poetry.utils._compat import decode
 from poetry.utils.env.script_strings import GET_PYTHON_VERSION_ONELINER
+
+
+if TYPE_CHECKING:
+    from cleo.io.io import IO
+
+    from poetry.config.config import Config
 
 
 class Python:
@@ -47,6 +57,57 @@ class Python:
     def minor_version(self) -> Version:
         return Version.from_parts(major=self.version.major, minor=self.version.minor)
 
+    @staticmethod
+    def _full_python_path(python: str) -> Path | None:
+        # eg first find pythonXY.bat on windows.
+        path_python = shutil.which(python)
+        if path_python is None:
+            return None
+
+        try:
+            encoding = "locale" if sys.version_info >= (3, 10) else None
+            executable = subprocess.check_output(
+                [path_python, "-c", "import sys; print(sys.executable)"],
+                text=True,
+                encoding=encoding,
+            ).strip()
+            return Path(executable)
+
+        except subprocess.CalledProcessError:
+            return None
+
+    @staticmethod
+    def _detect_active_python(io: IO) -> Path | None:
+        io.write_error_line(
+            "Trying to detect current active python executable as specified in"
+            " the config.",
+            verbosity=Verbosity.VERBOSE,
+        )
+
+        executable = Python._full_python_path("python")
+
+        if executable is not None:
+            io.write_error_line(f"Found: {executable}", verbosity=Verbosity.VERBOSE)
+        else:
+            io.write_error_line(
+                "Unable to detect the current active python executable. Falling"
+                " back to default.",
+                verbosity=Verbosity.VERBOSE,
+            )
+
+        return executable
+
     @classmethod
     def get_system_python(cls) -> Python:
         return cls(executable=sys.executable)
+
+    @classmethod
+    def get_preferred_python(cls, config: Config, io: IO | None = None) -> Python:
+        io = io or NullIO()
+
+        if config.get("virtualenvs.prefer-active-python") and (
+            active_python := Python._detect_active_python(io)
+        ):
+            return cls(executable=active_python)
+
+        return cls.get_system_python()
--- a/tests/utils/env/test_env_manager.py
+++ b/tests/utils/env/test_env_manager.py
@@ -1070,7 +1070,7 @@ def test_create_venv_uses_patch_version_to_detect_compatibility(
 
     m.assert_called_with(
         config_virtualenvs_path / f"{venv_name}-py{version.major}.{version.minor}",
-        executable=None,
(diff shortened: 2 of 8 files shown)


src/poetry/utils/env/python_manager.py
```diff
@@ -0,0 +1,154 @@
...
+
+
+if TYPE_CHECKING:
+    from cleo.io.io import IO
+
+    from poetry.config.config import Config
+    from poetry.poetry import Poetry
+
+
+class Python:
```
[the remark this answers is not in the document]
finswimmer: Not really, because objects of this class don't "manage" python. It's a value object represented by the path to the executable and its version.

src/poetry/utils/env/python_manager.py
```diff
@@ -0,0 +1,154 @@
...
+                verbosity=Verbosity.VERBOSE,
+            )
+
+        return executable
+
+    @staticmethod
+    def get_system_python() -> Python:
+        return Python(executable=sys.executable)
+
+    @staticmethod
```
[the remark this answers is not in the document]
finswimmer: :thinking: I'm more used to `@staticmethod` for methods that serves as a factory. But it looks like `@classmethod` is more common and we are doing so in other place. So I'm fine with changing it.

src/poetry/utils/env/python_manager.py
```diff
@@ -0,0 +1,158 @@
...
+import subprocess
+import sys
+
+from functools import cached_property
+from pathlib import Path
+from typing import TYPE_CHECKING
+
+from cleo.io.null_io import NullIO
+from cleo.io.outputs.output import Verbosity
+from poetry.core.constraints.version import Version
```
Secrus, line 13: I am not sure if we should use this as a version in the long run. We had some reports that it doesn't allow proper Python versions to be used. I am not sure if it's better to change it now or wait for the next stage of changes.

src/poetry/utils/env/python_manager.py
```diff
@@ -0,0 +1,163 @@
...
+                "Unable to detect the current active python executable. Falling"
+                " back to default.",
+                verbosity=Verbosity.VERBOSE,
+            )
+
+        return executable
+
+    @classmethod
+    def get_system_python(cls) -> Python:
+        return cls(executable=sys.executable)
```
TBBle, line 105: Is this the right name for this function? If Poetry is installed in a venv, e.g., using pipx, it'll pick up the pipx venv's Python binary, which seems surprising as a "system" Python version, which to me would imply "shared instance, outside any venv".

I don't see this function being used excepts in the tests, so I'm not clear what the intended behaviour is, but perhaps `get_running_python` or `get_poetry_python` would be more-accurate? (`get_preferred_python` open-codes the same behaviour so it will also pick up a pipx-poetry Python executable and not a system-shared instance, but it only appears to be used to extract the version and create new venvs, so that seems like the expected/intended behaviour there.)

finswimmer: Good catch :+1: I guess my initial thought was, that there is a need to get the Python executable that poetry itself uses. But obviously the only place where this is required is within `get_preferred_python`, where we decide whether we use this executable or the one used in the current shell. We can either keep this method and used in within `get_preffered_python` to make the choice options more explicit or remove it.

src/poetry/utils/env/python_manager.py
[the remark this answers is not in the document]
Secrus: The file name is `python_manager` though.

src/poetry/utils/env/python_manager.py
```diff
@@ -0,0 +1,163 @@
...
+    @cached_property
+    def minor_version(self) -> Version:
+        return Version.from_parts(major=self.version.major, minor=self.version.minor)
+
+    @staticmethod
+    def _full_python_path(python: str) -> Path | None:
+        # eg first find pythonXY.bat on windows.
+        path_python = shutil.which(python)
+        if path_python is None:
+            return None
```
Secrus, line 68: nit: 
```suggestion
        if path_python := shutil.which(python) is None:
            return None
```

src/poetry/utils/env/python_manager.py
```diff
@@ -0,0 +1,163 @@
...
+    from poetry.poetry import Poetry
+
+
+class Python:
+    def __init__(self, executable: str | Path, version: Version | None = None) -> None:
+        self.executable = Path(executable)
+        self._version = version
+
+    @property
+    def version(self) -> Version:
```
Secrus, line 34: I think this could be a `cached_property`

src/poetry/utils/env/python_manager.py
```diff
@@ -0,0 +1,163 @@
...
+
+    @classmethod
+    def get_system_python(cls) -> Python:
+        return cls(executable=sys.executable)
+
+    @classmethod
+    def get_by_name(cls, python_name: str) -> Python | None:
+        _executable = cls._full_python_path(python_name)
+        if not _executable:
+            return None
```
Secrus, line 111:
```suggestion
        if not (_executable := cls._full_python_path(python_name)):
            return None
```

also, not sure if we need the `_` prefix there

src/poetry/utils/env/python_manager.py
```diff
@@ -0,0 +1,163 @@
...
+
+        return cls(executable=_executable)
+
+    @classmethod
+    def get_preferred_python(cls, config: Config, io: None | IO = None) -> Python:
+        io = io or NullIO()
+        _executable = Path(sys.executable)
```
Secrus, line 118: Does it have to be `_` prefixed?

tests/utils/env/test_env_manager.py
```diff
@@ -1068,7 +1068,7 @@ def test_create_venv_uses_patch_version_to_detect_compatibility(
 
     m.assert_called_with(
         config_virtualenvs_path / f"{venv_name}-py{version.major}.{version.minor}",
-        executable=None,
+        executable=Path(sys.executable),
```
Secrus, line 1071: Why did that change? Did you change something in the algorithm?

tests/utils/test_python_manager.py
```diff
@@ -0,0 +1,116 @@
...
+
+@pytest.mark.skipif(sys.platform != "win32", reason="Windows only")
+def test_detect_active_python_with_bat(tmp_path: Path) -> None:
+    """On Windows pyenv uses batch files for python management."""
+    python_wrapper = tmp_path / "python.bat"
+    wrapped_python = Path(r"C:\SpecialPython\python.exe")
+    encoding = "locale" if sys.version_info >= (3, 10) else None
+    with python_wrapper.open("w", encoding=encoding) as f:
+        f.write(f"@echo {wrapped_python}")
+    os.environ["PATH"] = str(python_wrapper.parent) + os.pathsep + os.environ["PATH"]
```
Secrus, line 104: `os.pathsep.join(...)` maybe?

Secrus (commented):
LGTM, some minor issues here and there

src/poetry/utils/env/python_manager.py
radoering: At work, we consider it bad style to use the walrus operator in an `if` statement even though the assigned variable is used outside of the `if` later. In other words, I do not like the suggested change.

src/poetry/utils/env/python_manager.py
radoering: At work, we consider it bad style to use the walrus operator in an `if` statement even though the assigned variable is used outside of the `if` later. In other words, I do not like the suggested change.

src/poetry/utils/env/python_manager.py
finswimmer: `cached_property` is available from Python 3.8. Because Poetry's minimal supported Python version is 3.8 as well, we can do so :+1:

src/poetry/utils/env/python_manager.py
finswimmer: We have this is rule as well at work :smiley:

src/poetry/utils/env/python_manager.py
finswimmer: Probably not :shrug:

tests/utils/env/test_env_manager.py
finswimmer: Previously the `create_venv` assumed, that unless an `executable` was given or we detected the active python, the "default" python was used. Now we use an instance of `Python` in any way. See https://github.com/python-poetry/poetry/pull/9050/files#diff-37f4bd765a025d97c1128f60e349dc94b7df596ac9c9dd8293b7b4ad4e6bafb3R405

tests/utils/test_python_manager.py
finswimmer: I just copied this test over from `test_env_manager.py`. If we want to change this test, I would prefer using `Path`. But I have no strong feelings about whether we want to touch this test or not :man_shrugging:

src/poetry/utils/env/python_manager.py
finswimmer: I see now, why I haven't used `cached_property`. We need `self._version` anyway, because we can setup the version during `intit`.

tests/utils/test_python_manager.py
Secrus: let's leave it for now, we can go back to tests quality later

src/poetry/utils/env/python_manager.py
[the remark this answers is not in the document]
finswimmer: I can rename the file to `python.py`, but I'm not sure if this might lead to any conflicts.

src/poetry/utils/env/python_manager.py
finswimmer: I kept this method and used in `get_preffered_python`.

Regarding the name: @Secrus & @radoering Are you fine with  `get_system_python` or do you prefer any other name?

src/poetry/utils/env/python_manager.py
finswimmer: Removed it.

src/poetry/utils/env/python_manager.py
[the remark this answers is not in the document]
Secrus: `python_detector.py` and `PythonDetector`? Still not the best name, but says more about the purpose of this class. It's a bikeshedding discussion though. If you want to, just resolve this convo and leave it as is.

src/poetry/utils/env/python_manager.py
radoering: I understand the confusion.

On the one hand, `SystemEnv` is the env Poetry is installed in. `poetry env use system` tells Poetry to use it's own Python version as default for new venvs. Thus, in our wording `system` is Poetry's own env.

One the other hand, `virtualenvs.options.system-site-packages` does not refer to Poetry's own env. And most people will probably not associate `system` with Poetry's own env.

Not sure, if `get_poetry_python` is a better name or if it can be confused with the project's python. At least, `system` is quite consistent in our code. No strong opinion on this. 🤷

Secrus: For me `system python` was always confusing, but I agree with @radoering that it is quite consistent in the code now, so we can leave it as is.

Secrus (dismissed):
LGTM, let's resolve last conversation and can be merged

tests/utils/test_python_manager.py
```diff
@@ -0,0 +1,116 @@
...
+
+    active_python = Python._detect_active_python(NullIO())
+
+    assert active_python == wrapped_python
+
+
+def test_python_find_compatible(project_factory: ProjectFactory) -> None:
```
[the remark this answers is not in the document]
finswimmer: Interesting finding :+1: I believe it's not a problem with the test. The test discovers a bug/suboptimal implementation. 

The logic within `get_compatible_python()` assumes that there as an executable with a version number in the name (e.g. `python3`, `python312`). The assumption is obviously not correct. We didn't run into any trouble until now, because this logic is only triggered if the "system" or "preferred" python isn't compatible. So an executable `python` was checked before.

We could add  `python` to  the list of executables that are tried. But where? At the beginning or at the end?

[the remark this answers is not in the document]
radoering: > We could add python to the list of executables that are tried. But where? At the beginning or at the end?

I like adding it but do not have a strong opinion where to add it.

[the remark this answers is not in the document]
finswimmer: I have added it to the end.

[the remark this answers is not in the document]
TBBle: I think this test has a high chance of flaking on random Windows systems until we have #2117 resolved (i.e. sometime after this PR) since a 'python.exe' in the path on Windows has a reasonable chance of being the Store-launcher version (which we would hopefully _reject_ if it was chosen...) or simply not existing at all.

We _could_ also try 'py.exe' after 'python.exe' which would give you _some_ PEP-514 version of Python, if possible. I just noticed on my machine that it picked up the Python 3.13.0rc1 experimental free-threaded build in preference to either the traditional 3.13.0rc1, or (released) 3.12, so yeah... it's a lexicographical version ordering. Surprise!

But I don't think we try 'py.exe' now, so I don't think that change would make sense in this refactoring anyway.

Is there value in marking this test "May fail due to system configuration on Windows, see #2117" or the like? (I am _assuming_ that the implementation for #2117 will work for both `py use env` and `get_compatible_python`...)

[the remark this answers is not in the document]
Secrus: Ok, but from what I understood, this PR is just refactoring, not bug fixing. If there are bugs that were present previously, we can take care of those in a separate PR.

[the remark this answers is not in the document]
TBBle: Yes, that's why my suggestion was to simply annotate the test, since it's a _newly distinguished function_ in this refactoring, and the new test for it has flake-potential in rational-but-uncontrolled environments, and a possibly-related existing bug report where we can track and resolve (or at least improve) that.

src/poetry/utils/env/python_manager.py
finswimmer: Good to see that we are on the same boat :smile: So I would leave it at it is for now.

tests/utils/test_python_manager.py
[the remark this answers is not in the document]
finswimmer: Added a comment as suggested.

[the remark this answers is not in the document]
TBBle: Just looking at the comment, I think it might have conflated a few things:
* The Store-launcher that [is bundled in Windows was python.exe in the path](https://learn.microsoft.com/windows/python/faqs#why-does-running-python-exe-open-the-microsoft-store-). I expect that's rejected by logic that checks the version of the discovered Python binary, but would pop up the store UI in the meantime.
* A store-installed Python will put functioning python.exe and python3.exe in the path (apparently, I didn't test this), causing this test to work correctly, and giving the most POSIX-like experience in this respect.
* A python.org installer may or may not put python.exe in the path (not sure of the default off-hand, I always have it off personally), but definitely won't put python3.exe in the path.
* A python.org installer would normally install a py.exe which behaves link the normal /usr/bin/python symlink on POSIX machines, as far as `py -c` goes, anyway.

So I think
```python
    # Note: This test may fail on Windows systems using Python from the Microsoft Store,
    # as the executable is named `py.exe`, which is not currently recognized by
    # Python.get_compatible_python. This issue will be resolved in #2117.
```

maybe should be

```python
    # Note: This test may fail on Windows systems using Python from the python.org installer,
    # as the included `python.exe` may not be on the PATH, and hence not seen by
    # Python.get_compatible_python. This issue will be resolved in #2117.
    # Note: On Windows systems with nothing else putting a `python.exe` in the PATH,
    # this test may open the Microsoft Store, see
    # https://learn.microsoft.com/windows/python/faqs#why-does-running-python-exe-open-the-microsoft-store-
```

(I added the second note because #2117 won't address that, so it'll have to be left behind when #2117 is done and the first note is removed. There's basically nothing we can do for that case except special case that exact executable, perhaps by full path in `_full_python_path`, and that's something we can fuss about with later if it's interesting.)

That said, looking at the rest of the comment, if we can guarantee that the Poetry-running Python executable will be selected in this case, maybe the comment could be more-succinctly summarised to just that, or even asserted somehow? Although with a quick look, I don't see how this happens? `get_compatible_python` doesn't appear to try the result of `get_system_python` before it goes searching the PATH with `_full_python_path`, so if our currently-executing python binary isn't in the PATH, it won't be found, I think.

Secrus (approved):

Issue #9042, opened 2024-02-27
Poetry 1.8.0 does not tolerate new sources in pyproject.toml


aabmets:
### Description

I'm using Poetry in a Jenkins pipeline to install packages from a private repository for the pytest stage:
```Groovy
withDockerContainer(image:'*********.com/python:3.12.1-slim-bookworm', args:'-u root') {
    sh "pip install poetry==${env.POETRY_VERSION} " +
        "--trusted-host ${env.TRUSTED_PIP_HOST} " +
        "--index-url ${env.PIP_REGISTRY} " +
        "--root-user-action ignore " +
        "--no-cache-dir"
    sh "poetry config virtualenvs.create false"
    sh "poetry source add --priority=primary privsrc ${env.PIP_REGISTRY}"
    sh "poetry install --with dev"
    sh "pytest"
}
```

The pipeline broke, because I hadn't pinned poetry version to 1.7.1 and it automatically used the latest version 1.8.0, which poetry had released on 25th of February. The pipeline works as expected with Poetry version 1.7.1.

The line `poetry install --with dev` causes poetry to throw the following error:
```bash
pyproject.toml changed significantly since poetry.lock was last generated.  
Run `poetry lock [--no-update]` to fix the lock file.
```

Inserting the suggested command `poetry lock --no-update` after the `poetry source add` command will produce the following error instead:
```bash
17:08:37  poetry lock --no-update
17:08:37  Skipping virtualenv creation, as specified in config file.
17:08:37  Updating dependencies
17:08:37  Resolving dependencies...
17:08:47  
17:08:47  'content-range'
```

#### Option A:
```markdown
Downgrade to 1.7.1
```

#### Option B:
```bash session
poetry config solver.lazy-wheel false
poetry source add --priority=primary <your private source>
poetry lock --no-update
poetry install [--with dev]
```

### Poetry Installation Method

pip

### Operating System

Linux

### Poetry Version

1.8.0

### Poetry Configuration

```bash session
10:20:34  poetry config --list
10:20:35  cache-dir = "/root/.cache/pypoetry"
10:20:35  experimental.system-git-client = false
10:20:35  installer.max-workers = null
10:20:35  installer.modern-installation = true
10:20:35  installer.no-binary = null
10:20:35  installer.parallel = true
10:20:35  keyring.enabled = true
10:20:35  repositories.privsrc.url = "https://********.com/repository/pypi.org/simple"
10:20:35  solver.lazy-wheel = true
[...]

PR #9051, opened 2024-02-28
harden lazy wheel wheel error handling

abn:
This change attempts to provide a catch-all safety net for circumstances where an index server is behaving badly when ranged requests are made.

Additionally, this change also helps guard against  missing "Content-Range" headers in specific scenarios.

Resolves: #9042

### DIFF
--- a/src/poetry/inspection/lazy_wheel.py
+++ b/src/poetry/inspection/lazy_wheel.py
@@ -41,15 +41,19 @@ if TYPE_CHECKING:
 logger = logging.getLogger(__name__)
 
 
-class HTTPRangeRequestUnsupported(Exception):
+class LazyWheelUnsupportedError(Exception):
+    """Raised when a lazy wheel is unsupported."""
+
+
+class HTTPRangeRequestUnsupported(LazyWheelUnsupportedError):
     """Raised when the remote server appears unable to support byte ranges."""
 
 
-class UnsupportedWheel(Exception):
+class UnsupportedWheel(LazyWheelUnsupportedError):
     """Unsupported wheel."""
 
 
-class InvalidWheel(Exception):
+class InvalidWheel(LazyWheelUnsupportedError):
     """Invalid (e.g. corrupt) wheel."""
 
     def __init__(self, location: str, name: str) -> None:
@@ -85,6 +89,24 @@ def metadata_from_wheel_url(
         # themselves are invalid, not because we've messed up our bookkeeping
         # and produced an invalid file.
         raise InvalidWheel(url, name)
+    except Exception as e:
+        if isinstance(e, LazyWheelUnsupportedError):
+            # this is expected when the code handles issues with lazy wheel metadata retrieval correctly
+            raise e
+
+        logger.debug(
+            "There was an unexpected %s when handling lazy wheel metadata retrieval for %s from %s: %s",
+            type(e).__name__,
+            name,
+            url,
+            e,
+        )
+
+        # Catch all exception to handle any issues that may have occurred during
+        # attempts to use Lazy Wheel.
+        raise LazyWheelUnsupportedError(
+            f"Attempts to use lazy wheel metadata retrieval for {name} from {url} failed"
+        ) from e
 
 
 class MergeIntervals:
@@ -590,6 +612,12 @@ class LazyWheelOverHTTP(LazyFileOverHTTP):
                 f"did not receive partial content: got code {code}"
             )
 
+        if "Content-Range" not in tail.headers:
+            raise LazyWheelUnsupportedError(
+                f"file length cannot be determined for {self._url}, "
+                f"did not receive content range header from server"
+            )
+
         file_length = self._parse_full_length_from_content_range(
             tail.headers["Content-Range"]
         )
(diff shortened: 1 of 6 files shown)


tests/conftest.py
```diff
@@ -513,3 +519,9 @@ def venv_flags_default() -> dict[str, bool]:
         "no-pip": False,
         "no-setuptools": False,
     }
+
+
+@pytest.fixture(autouse=(os.name == "nt"))
+def httpretty_windows_mock_urllib3_wait_for_socket(mocker: MockerFixture) -> None:
+    # this is a workaround for https://github.com/gabrielfalcao/HTTPretty/issues/442
+    mocker.patch("urllib3.util.wait.select_wait_for_socket", retruns=True)
```
radoering, line 527: Is `retruns` a typo?

src/poetry/inspection/lazy_wheel.py
```diff
@@ -85,6 +89,14 @@ def metadata_from_wheel_url(
         # themselves are invalid, not because we've messed up our bookkeeping
         # and produced an invalid file.
         raise InvalidWheel(url, name)
+    except Exception as e:
+        if isinstance(e, LazyWheelUnsupportedError):
```
radoering, line 93: We probably should at least add a debug log with the error message.

radoering (changes requested):

tests/conftest.py
abn: Hah. Yes. Surprised it still just worked lol.

src/poetry/inspection/lazy_wheel.py
abn: It will pop up in the stack trace anyway. But yeah adding debug log cannot hurt.

aabmets:
Great job! Thank you a lot for fixing this! :)

src/poetry/inspection/lazy_wheel.py
```diff
@@ -614,7 +642,10 @@ def _extract_content_length(
...
             # This indicates that the requested range from the end was larger than the
             # actual file size: https://www.rfc-editor.org/rfc/rfc9110#status.416.
-            if code == codes.requested_range_not_satisfiable:
+            if (
+                code == codes.requested_range_not_satisfiable
+                and "Content-Range" in resp.headers
+            ):
                 # In this case, we don't have any file content yet, but we do know the
                 # size the file will be, so we can return that and exit here.
                 assert resp is not None
```
dimbleby, line 651: this assertion is now too late to be useful: we've already used `resp` in the previous line

dimbleby:
https://github.com/python-poetry/poetry/issues/9056 looks as though it provdes another instance of the LazyWheel code making optimistic assertions about server behaviour, dunno whether it makes sense to handle that in this pull request or separately

> 526│                 assert response_length == min(initial_chunk_size, ret_length)

src/poetry/inspection/lazy_wheel.py
abn: I guess we can we can move it into the if condition instead.

abn:
> dunno whether it makes sense to handle that in this pull request or separately

I reckon another PR might be better largely because I do not want to hold this up too long since this will in-effect solve that as well, or make it not impact UX.  But might be good to have a targeted fix for defense-in-depth measure.

src/poetry/inspection/lazy_wheel.py
dimbleby: if `resp` was `None` then `code` would also be `None` so I guess the extra check really is unnecessary, I suppose I don't care either way

src/poetry/inspection/lazy_wheel.py
```diff
@@ -85,6 +89,24 @@ def metadata_from_wheel_url(
         # themselves are invalid, not because we've messed up our bookkeeping
         # and produced an invalid file.
         raise InvalidWheel(url, name)
+    except Exception as e:
+        if isinstance(e, LazyWheelUnsupportedError):
+            # this is expected when the code handles issues with lazy wheel metadata retrieval correctly
+            raise e
+
+        logger.debug(
```
radoering, line 97: Since you already mentioned it in the chat:

> It is a bit unfortunate that we will end up not really getting reports of lazy wheel not working if there is a bug

We could make this a warning to increase visibility (of course only for unexpected exceptions, see other comment). What do you think?

src/poetry/inspection/lazy_wheel.py
```diff
@@ -85,6 +89,24 @@ def metadata_from_wheel_url(
         # themselves are invalid, not because we've messed up our bookkeeping
         # and produced an invalid file.
         raise InvalidWheel(url, name)
+    except Exception as e:
+        if isinstance(e, LazyWheelUnsupportedError):
+            # this is expected when the code handles issues with lazy wheel metadata retrieval correctly
+            raise e
```
radoering, line 95: I think it might be good to add a debug log for known exceptions because otherwise it might be difficult to tell why exactly lazy-wheel failed (e.g. no content-range).

abn: This will basically be raised as it previously used to. In this scenario we are simply re-raising the range/invalid wheel exception. Don't think the log would add anything meaningful. 

The catch-all is if it isn't a known error.

src/poetry/inspection/lazy_wheel.py
abn: I do not think that adds anything as it will become background noise for most people dealing with this issue as they do not control the ability to manage the index server. It will become an annoying nag rather than serve the purpose of improving their experience.

We can think about how best to balance these concerns in a separate PR I reckon. Might need an option to hide the warning etc.

radoering (approved):

PR #9064, opened 2024-02-29
build: allow specifying local version label

abn:


src/poetry/console/commands/build.py
```diff
@@ -45,6 +51,13 @@ def _build(
         else:
             raise ValueError(f"Invalid format: {fmt}")
 
+        local_version_label = self.option("local-version")
+
+        if local_version_label:
```
Secrus, line 56:
```suggestion
        if local_version_label := self.option("local-version"):
```

Secrus (changes requested):
Just one issue, otherwise LGTM

src/poetry/console/commands/build.py
abn: Ah forgot we are `^3.8` now :)

Secrus (approved):

Issue #1329, opened 2019-08-25
Make poetry build remove older versions of packages in the dist directory


dclong:
- [x] I have searched the [issues](https://github.com/sdispater/poetry/issues) of this repo and believe that this is not a duplicate.
- [x] I have searched the [documentation](https://poetry.eustace.io/docs/) and believe that my question is not covered.

## Feature Request
Make `poetry build` remove older versions of packages (built before) in the dist directory. For example, support the current version of the project/package is 0.4.3 and there is a previous built package with the version 0.4.2. `poetry build` should remove the previous built package with version 0.4.2 after generating the package with version 0.4.3.

wagnerluis1982:
I feel the request is valid, but it shouldn't be set as default operation.

I think a new argument would be the best idea here.

Something like the following would make sense...
```
poetry build --clean
```

LukeSavefrogs:
> I think a new argument would be the best idea here.
> 
> Something like the following would make sense...
> ```
> $ poetry build --clean
> ```

I agree, that would be the best.

LaoshuBaby:
Issue that may related to this request: #3331

PR #9067, opened 2024-02-29
build: add --clean option

abn:
Resolves: #1329


Secrus (approved):

abn:
@Secrus can you review again please, expanded the test suite a bit.

tests/console/commands/test_build.py
```diff
@@ -82,6 +83,35 @@ def test_build_with_local_version_label(
...
+def test_build_with_clean(
+    tmp_tester: CommandTester, tmp_project_path: Path, tmp_poetry: Poetry, clean: bool
+) -> None:
+    dist_dir = tmp_project_path.joinpath("dist")
+    dist_dir.joinpath("hello").touch(exist_ok=True)
+
+    tmp_tester.execute("--clean" if clean else "")
+    build_artifacts = tuple(dist_dir.glob("*"))
+
+    assert len(build_artifacts) == 2 if clean else 3
```
Secrus, line 96: Could we put the amount of artifacts and `--clean` param (line 93) in parametrize? I don't like the idea of conditionals in tests. It's a nitpick though, if you disagree, disregard it.

Secrus (approved):

tests/console/commands/test_build.py
abn: In principle I agree. However since the test is what is creating the files, I think the code block needs to specify it rather than a parameter.  We can make it more clearer if you prefer? For example do one of he following.

1. Move the ternary conditional into an explicit block.
2. Make a count of the existing files in the directory and then assert it is `count + 2`.

But I suspect neither would be qualitatively better.

Issue #9072, opened 2024-02-29
Install error when installing a non-package in package-mode gives invalid instructions


andyclegg:
### Description

Create an empty project with `poetry init`, then attempt to install it. Without the `README.md` file being created, the installation will fail. When installation fails without configuring `package-mode`, Poetry 1.8.1 emits the following error message:

```
Warning: The current project could not be installed: [Errno 2] No such file or directory: '/README.md'
If you do not want to install the current project use --no-root.
If you want to use Poetry only for dependency management but not for packaging, you can set the operating mode to "non-package" in your pyproject.toml file.
In a future version of Poetry this warning will become an error!
```

The message instructs to 'set the operating mode to "non-package" in your pyproject.toml file', implying using 'non-package = true' or similar, whereas the correct key is 'package-mode = false'.

### Workarounds

The correct key can be inferred from a search.

### Poetry Installation Method

pipx

### Operating System

Linux Mint 21.3

### Poetry Version

1.8.1

### Poetry Configuration

```bash session
cache-dir = "/home/andy/.cache/pypoetry"
experimental.system-git-client = false
installer.max-workers = null
installer.modern-installation = true
installer.no-binary = null
installer.parallel = true
keyring.enabled = true
solver.lazy-wheel = true
virtualenvs.create = true
virtualenvs.in-project = null
virtualenvs.options.always-copy = false
virtualenvs.options.no-pip = false
virtualenvs.options.no-setuptools = false
virtualenvs.options.system-site-packages = false
virtualenvs.path = "{cache-dir}/virtualenvs"  # /home/andy/.cache/pypoetry/virtualenvs
virtualenvs.prefer-active-python = false
virtualenvs.prompt = "{project_name}-py{python_version}"
warnings.export = true
```

### Example pyproject.toml

```TOML
[tool.poetry]
name = "foo"
version = "0.1.0"
description = ""
authors = ["Andy"]
readme = "README.md"

[tool.poetry.dependencies]
python = "^3.9"


[build-system]
requires = ["poetry-core"]
build-backend = "poetry.core.masonry.api"
```

### Poetry Runtime Logs

```bash session
$ poetry -vvv install
Loading configuration file /home/andy/.config/pypoetry/config.toml
Using virtualenv: /home/andy/.cache/pypoetry/virtualenvs/foo-IVQOPcWJ-py3.9
Installing dependencies from lock file

Finding the necessary packages for the current system

Installing the current project: foo (0.1.0)

[...]

PR #9073, opened 2024-02-29
Improve error message when installing non-package in package-mode

andyclegg:
# Pull Request Check List

Resolves: #9072 

I have not added tests or documentation for the changed code, as this change only modifies an error string which I believe is currently not tested, and effectively *is* documentation. Happy to update the PR if needed!

### DIFF
--- a/src/poetry/console/commands/install.py
+++ b/src/poetry/console/commands/install.py
@@ -191,8 +191,8 @@ you can set the "package-mode" to false in your pyproject.toml file.
                 "If you do not want to install the current project"
                 " use <c1>--no-root</c1>.\n"
                 "If you want to use Poetry only for dependency management"
-                " but not for packaging, you can use non-package mode by setting "
-                '"package-mode = false" in your pyproject.toml file.\n'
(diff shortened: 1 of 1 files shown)


src/poetry/console/commands/install.py
```diff
@@ -191,8 +191,8 @@ def handle(self) -> int:
                 "If you do not want to install the current project"
                 " use <c1>--no-root</c1>.\n"
                 "If you want to use Poetry only for dependency management"
-                " but not for packaging, you can set the operating mode to "
-                '"non-package" in your pyproject.toml file.\n'
+                " but not for packaging, you can use non-package mode by setting "
+                '"package-mode = false" in your pyproject.toml file.\n'
```
abn, line 195:
```suggestion
                " but not for packaging, you can disable package mode by setting"
                " <c1>package-mode = false</> in your pyproject.toml file.\n"
```

Bit nit picky, but might be better to stick to the terminology in the docs "disable package mode" instead of "use non-package mode".

abn (changes requested):

PR #9078, opened 2024-02-29
trust empty requires dist with modern metadata

dimbleby:
when modern metadata says that it has no requires-dist, we can trust it

I bet it doesn't often matter:
- https://github.com/pypi/warehouse/pull/13606 was only just released
- as of today presumably more or less no-one (poetry included) is yet publishing metadata versioned at 2.2 or higher
- and probably nearly all newly published packages also publish wheels anyway.

still, when it happens it happens, and avoiding unnecessary builds is a good win

### DIFF
--- a/src/poetry/inspection/info.py
+++ b/src/poetry/inspection/info.py
@@ -15,6 +15,8 @@ from typing import Sequence
 import pkginfo
 
 from build import BuildBackendException
+from poetry.core.constraints.version import Version
+from poetry.core.factory import Factory
 from poetry.core.packages.dependency import Dependency
 from poetry.core.packages.package import Package
 from poetry.core.pyproject.toml import PyProjectTOML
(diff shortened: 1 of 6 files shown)


src/poetry/inspection/info.py
```diff
@@ -246,9 +247,18 @@ def _from_distribution(
         :param dist: The distribution instance to parse information from.
         """
         requirements = None
+        dynamic_metadata = False
+
+        if dist.metadata_version is not None:
+            metadata_version = Version.parse(dist.metadata_version)
+            dynamic_metadata_introduced = Version.parse("2.2")
```
eamanu, line 254: I'd create this as a global variable

dimbleby:
prior to their 1.10, the `pkginfo` parsing of files with metadata 2.3 failed to return any useful values.

this is more or less a proof that no-one was declaring metadata version 2.3: because if they did then things would surely go wrong.

However I see that `maturin` has just merged a change such that their backend will start declaring metadata version 2.3, and it looks as though they are about to cut a 1.5.0 release, and so fairly soon there will likely be a fresh supply of metadata-2.3 packages.  Once that happens I would not be surprised to see a little flurry of issues for which the answer will be "upgrade `pkginfo` in your poetry installation".

I've bumped the `pkginfo` requirement in this MR to 1.10, and made one of the test fixtures use metadata version 2.3.

dimbleby:
I see that #9130 rediscovered some of the above

abn (approved):

PR #9080, opened 2024-03-01
docs: add git subdirectory example to `poetry add`

estyrke:
I managed to find the correct incantation by trial and error, but thought it'd be nice to have an example in the help message. I noticed afterwards that it is already described in the actual documentation (https://python-poetry.org/docs/dependency-specification/#git-dependencies).


abn (approved):
The information is also available in the docs. 

https://python-poetry.org/docs/cli/#add

Issue #8943, opened 2024-02-04
Generate a README.md when using poetry init


Heus-Sueh:
- [x] I have searched the [issues](https://github.com/python-poetry/poetry/issues) of this repo and believe that this is not a duplicate.
- [x] I have searched the [FAQ](https://python-poetry.org/docs/faq/) and general [documentation](https://python-poetry.org/docs/) and believe that my question is not already covered.

## Feature Request

generate a generic or empty readme.md file at the end of poetry init, this will prevent the user from encountering an error due to the lack of readme.md when using `poetry install`

example:

```
{project_name}
============

**This app was generated by `Poetry`**

{description_of_project}

.. _`Poetry`: https://python-poetry.org/

```

dimbleby:
```console
$ poetry init --help                                                          
Description:
  Creates a basic pyproject.toml file in the current directory.
```
sounds like you want `poetry new`

Heus-Sueh:
> ```
> $ poetry init --help                                                          
> Description:
>   Creates a basic pyproject.toml file in the current directory.
> ```
> 
> sounds like you want `poetry new`

the `new` command does this automatically, I know, but `init` should do the same thing when you finish filling in the information or at least be optional

dimbleby:
`poetry init` is very clear that it is intended to create `pyproject.toml` and `poetry new` is very clear that it is intended to create a new project.

if you want to create more than only `pyproject.toml`, then you want `poetry new`.  

I might agree that having both of these commands is confusing: but that is not what you are reporting.  (I am fairly sure that there are other issues saying that!)  However: in a world where `poetry new` and `poetry init` both exist - this is working as designed.

Heus-Sueh:
> `poetry init` is very clear that it is intended to create `pyproject.toml` and `poetry new` is very clear that it is intended to create a new project.
> 
> if you want to create more than only `pyproject.toml`, then you want `poetry new`.
> 
> I might agree that having both of these commands is confusing: but that is not what you are reporting. (I am fairly sure that there are other issues saying that!) However: in a world where `poetry new` and `poetry init` both exist - this is working as designed.

For me, `poetry init` would be to initialize a new project, but as you explained in the docs it says that it only generates a `pyproject.toml`, a very strange design choice.

I would close this issue, but I'll wait for others to give their opinions here

AllanHOlesenBW:
[...]

PR #9088, opened 2024-03-02
cli/new: support interactive config

abn:
This change allows new command to be used interactively similar to the init command. Additionally, this also allows for configuration of description, author, python and dependencies via command line options.

Closes: #8943 #1854 #5710

### DIFF
--- a/docs/cli.md
+++ b/docs/cli.md
@@ -102,11 +102,17 @@ my-package
 
 ### Options
 
+* `--interactive (-i)`: Allow interactive specification of project configuration.
 * `--name`: Set the resulting package name.
 * `--src`: Use the src layout for the project.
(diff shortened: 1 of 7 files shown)


docs/cli.md
```diff
@@ -102,11 +102,17 @@ my-package
 
 ### Options
 
+* `--interactive (-i)`: Allow interactive specification of project configuration.
```
Secrus, line 105: Why do we need this over `io.is_interactive()`?

abn: For `poetry new` non-interactive is the current expectation, even if `--no-interaction` is not explicitly set. So we need a way to trigger the "interactive" configuration when someone needs it to bring parity with `poetry init`.

Secrus:
Is this supposed to replace #5710?

Secrus (approved):

PR #9094, opened 2024-03-02
faster PEP517 builds

dimbleby:
minor simplifications and tweaks to pep517 builds

- update version of `build` used during metadata extraction
- if pep517 metadata extraction fails, don't try fallback to `setup.py egg_info`
  - eggs are long deprecated
  - using setup.py directly from the command line is deprecated
  - if pep517 builds fail then poetry will later refuse to install the package in any case
- no need to install setuptools or wheel in the build environments, if they are needed as build requirements then they will be installed as build requirements
- need pip in the build environment for metadata extraction because we run `venv.run_pip()`, but don't need it in the build environment during final build

### DIFF
--- a/src/poetry/inspection/info.py
+++ b/src/poetry/inspection/info.py
@@ -4,7 +4,6 @@ import contextlib
 import functools
 import glob
 import logging
-import os
 
 from pathlib import Path
 from typing import TYPE_CHECKING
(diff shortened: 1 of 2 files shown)


abn:
For context around the deprecation of the `python setup.py <cmd>`, future visitors can check https://packaging.python.org/en/latest/discussions/setup-py-deprecated/

abn (approved):

PR #9101, opened 2024-03-03
Introduce HashManager to work around FIPS

Secrus:
Resolves: #8310

### DIFF
--- a/src/poetry/publishing/hash_manager.py
+++ b/src/poetry/publishing/hash_manager.py
@@ -1,0 +1,65 @@
+from __future__ import annotations
+
+import hashlib
+import io
+
+from contextlib import suppress
+from typing import TYPE_CHECKING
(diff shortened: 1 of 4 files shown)


radoering (changes requested):
Can we put `HashManager` in a separate module and name the test module after it (for example `hash_manager.py` and `test_hash_manager.py`)? IMO, that makes it easier to find relevant unit tests.

eamanu (approved):

Secrus:
> Can we put `HashManager` in a separate module and name the test module after it (for example `hash_manager.py` and `test_hash_manager.py`)? IMO, that makes it easier to find relevant unit tests.

Done. I forgot to change the test suite name from the old, initial implementation.

radoering (approved):

abn (approved):

PR #9107, opened 2024-03-03
ci/gha: simplify poetry setup and caching

abn:


.github/workflows/main.yml
```diff
@@ -70,23 +58,10 @@ jobs:
...
-      - name: Ensure cache is healthy
-        if: steps.cache.outputs.cache-hit == 'true'
         run: |
-          # `timeout` is not available on macOS, so we define a custom function.
-          [ "$(command -v timeout)" ] || function timeout() { perl -e 'alarm shift; exec @ARGV' "$@"; }
-
-          # Using `timeout` is a safeguard against the Poetry command hanging for some reason.
-          timeout 10s poetry run pip --version || rm -rf .venv
+          poetry config virtualenvs.in-project true
+          poetry config virtualenvs.options.no-pip true
```
radoering, line 63: Not so sure about `no-pip`. We still depend on pip for uninstalling (and thereby also for updating) packages. I suppose we still fall back to the embedded pip wheel if there is no pip in the venv? Using the wheel directly proved brittle in the past. On the other hand, we can still change it if CI becomes flaky...

abn: The configuration is really for the bootstrapped Poetry installation. We should not (and do not) rely on the managed project's virtual environment having  `pip` available.

We switched to embedded wheel solution since there were a tonne of issues regarding the `pip` version in the creating virtual environment not being right, or non-deterministic and hence causing breakages.

I am curious, however, if there has been issues recently that was caused by the embedded wheel. It should not be any more brittle than relying on the `pip` installed by the user in the virtualenv. 

All that said, as you say, we can change it if there are issues that arise later.

neersighted (approved):

.github/workflows/main.yml
radoering: Since #6062 (Poetry 1.2.1), we prefer the in-venv-pip over the embedded wheel.

Further, using the embedded wheel directly is not supported by pip https://github.com/pypa/pip/issues/11622#issuecomment-1327720454 and according to the linked PEP even discouraged.

abn: Interesting. I will read up on those, seems I missed those changes while on hiatus.

neersighted:
Automated backport PRs opened:
* https://github.com/python-poetry/poetry/pull/9224

PR #9108, opened 2024-03-03
Drop direct dependency on crashtest

Secrus:
`crashtest` was used in quite a limited scope in Poetry. Since the next version of Cleo will also drop this requirement, I decided to remove it completely from Poetry.


src/poetry/mixology/failure.py
```diff
@@ -91,7 +103,31 @@ def write(self) -> str:
...
 
             buffer.append(message)
+        if required_python_version_notification:
+            # Add suggested solution
+            links = ",".join(
+                f"\n    <fg=blue>{link}</>"
+                for link in [
+                    "https://python-poetry.org/docs/dependency-specification/#python-restricted-dependencies",
+                    "https://python-poetry.org/docs/dependency-specification/#using-environment-markers",
+                ]
```
abn, line 113:
```suggestion
                f"\n    <fg=blue>https://python-poetry.org/docs/dependency-specification/#{section}</>"
                for section in [
                    "python-restricted-dependencies",
                    "using-environment-markers",
                ]
```

abn (approved):

PR #9113, opened 2024-03-05
Remove unnecessary explicit use of 'cwd'

Cypher1:
This removes an explicit override of the `cwd` argument to `Factory.create_poetry` which already computes the current working directory.

This de-risks changes currently being considered https://github.com/python-poetry/poetry-core/pull/695 and avoids a runtime import of `pathlib`.


abn:
I don't think this change is required. This makes things less explicit. And relies on implicit behavior.

And as mentioned in the referenced PR, that change might not be accepted.

Cypher1:
@abn I agree that it is less explicit, but the current implementation is explicitly misleading.

In the current implementation the directory used is not the current working directory, but ANY directory including or above the current working directory containing a `pyproject.toml`.

This has bitten me and probably others multiple times when using poetry in projects that have subprojects. I'd like to address this and this is one of the instances where the assumption is baked in. This should help poetry support monorepos in future, but for now is just updating the code to match the documentation.

abn:
> In the current implementation the directory used is not the current working directory, but ANY directory including or above the current working directory containing a `pyproject.toml`.

This is by design. Similar to git.

> This has bitten me and probably others multiple times when using poetry in projects that have subprojects. I'd like to address this and this is one of the instances where the assumption is baked in. This should help poetry support monorepos in future, but for now is just updating the code to match the documentation.

I am not sure I quite follow. Can you clarify how your change makes things better? If you actually have a subproject, your current directory pyproject should be picked up, if not, the direct parents are searched. I might be missing something obvious here, so please feel free to let me know.

eamanu:
IMO this change is not necessary right now, I think is better be explicit.
> This should help poetry support monorepos in future, but for now is just updating the code to match the documentation.

But maybe would be better introduce this change with a feature to be better look the benefits of remove the explicit cwd.

Cypher1:
> > In the current implementation the directory used is not the current working directory, but ANY directory including or above the current working directory containing a `pyproject.toml`.
> 
> This is by design. Similar to git.

I think you misunderstand me.

I totally support the *default* behaviour matching git as you say. I'm specifically describing the behaviour when an explicit `-C` argument is passed in.

This behaviour leads to poetry running from parent directories that are not described by the `-C` argument, even when the `-C` argument is not an existing directory.

My belief is that poetry should return at minimum warn the user when the user has requested a project directory that does not exist, and preferrably should not run the command (which may be destructive to a project that the user did not intend to modify).

Thanks for reconsidering this and for asking further when I was not sufficiently clear earlier.

abn:
@Cypher1 unfortunately, right now, I cannot see the need for this change. We appreciate the time and effort you have put into this pull request, but given that this is not solving an issue, I am going to close this. As far as the team is concerned the behavior today is by design.

If you would like to change the behavior, I would suggest a feature request with a clear description of what is being solved. As far as I can tell, you want poetry to stop walking up the parent tree - this is a breaking change to the cli and will need to be discussed.

PR #9132, opened 2024-03-08
Add plumbing for legacy repository search and refactor repository fixtures

abn:
*Add support for simple search in legacy repositories.*

Relates-to: https://github.com/python-poetry/poetry/issues/2446

*Remove unwanted stubbing of `LegacyRepository.packages`*

This change removes a stubbing change added in 46ae4f5. This change was incorrect, as the base class already sets package list to an empty list.

The test case was also incorrect as it forced addition to internal package list.

This change introduces httpretty based mocking for tests using mocked legacy or pypi repositories. This also fixes various issues with tests incorrectly relying on remote versions for metadata and distributions instead of expected mocked versions. Additionally, we also auto enables PyPI repository mocking to ensure no external requests are made without explicit handling. Further, a new `generate.py` script has now been added to help regenerate existing repository json and distribution fixtures; or add/remove release versions to the fixtures.

These changes also lead to the fixing of an incorrect handling of sdist inspection.


abn:
Windows failures are due to path parsing and fixture look up. Will fix that later.

tests/repositories/fixtures/pypi.org/generate.py
```diff
@@ -0,0 +1,730 @@
...
+
+    def copy_as_is(self, link: Link) -> ReleaseFileMetadata:
+        dst = FIXTURE_PATH_REPOSITORIES_PYPI / "dists" / link.filename
+        logger.info(
+            "Preserving release file from %s to %s",
+            link.url,
+            dst.relative_to(FIXTURE_PATH_REPOSITORIES_PYPI),
+        )
+
+        # noinspection PyProtectedMember
```
Secrus, line 529: Do we need this? I don't think ruff understands that.

tests/repositories/fixtures/pypi.org/generate.py
```diff
@@ -0,0 +1,730 @@
...
+        dst = FIXTURE_PATH_REPOSITORIES_PYPI / "stubbed" / link.filename
+        is_protected = self.should_preserve_file_content_check(link)
+
+        logger.info(
+            "Stubbing release file from %s to %s",
+            link.url,
+            dst.relative_to(FIXTURE_PATH_REPOSITORIES_PYPI),
+        )
+
+        # noinspection PyProtectedMember
```
Secrus, line 545: ditto

tests/repositories/fixtures/pypi.org/generate.py
```diff
@@ -0,0 +1,730 @@
...
```
Secrus, line 571: ditto

tests/repositories/fixtures/pypi.org/generate.py
```diff
@@ -0,0 +1,730 @@
...
+            path.relative_to(json_fixture_dir),
+        )
+        return data
+
+    def _get_remote_json_data(
+        self, name: str, version: str | None = None
+    ) -> dict[str, Any]:
+        if version is None:
+            logger.info("Fetching remote json via https://pypi.org/simple/%s", name)
+            # noinspection PyProtectedMember
```
Secrus, line 626: ditto

tests/repositories/fixtures/pypi.org/generate.py
```diff
@@ -0,0 +1,730 @@
...
+                headers={"Accept": "application/vnd.pypi.simple.v1+json"},
+            )
+        else:
+            logger.info(
+                "Fetching remote json via https://pypi.org/pypi/%s/%s/json",
+                name,
+                version,
+            )
+
+            # noinspection PyProtectedMember
```
Secrus, line 638: ditto

tests/repositories/fixtures/pypi.org/generate.py
```diff
@@ -0,0 +1,730 @@
...
+
+        data: dict[str, Any] = {}
+
+        if path.exists():
+            data = json.loads(path.read_text())
+
+        if prefer_remote:
+            remote_data = self._get_remote_json_data(name, version)
+
+            if remote_data is not None:
```
Secrus, line 612:
```suggestion
            if (remote_data := self._get_remote_json_data(name, version)) is not None:
```

Secrus (changes requested):

tests/repositories/fixtures/pypi.org/generate.py
```diff
@@ -0,0 +1,730 @@
...
+
+    def process(self) -> None:
+        for file in FIXTURE_PATH_REPOSITORIES_PYPI.joinpath("stubbed").iterdir():
+            if file.is_file():
+                file.unlink()
+
+        files = []
+
+        for dist in self.release_specs:
+            files.extend(factory.add_release_file(dist))
```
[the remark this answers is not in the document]
abn: Ha, think i moved the code from `__main__` to here at some point and it just worked because well ... global.

src/poetry/inspection/info.py
```diff
@@ -490,7 +490,10 @@ def from_directory(cls, path: Path, disable_build: bool = False) -> PackageInfo:
         else:
             info = cls.from_metadata_directory(path)
 
-            if not info or info.requires_dist is None:
+            # unfortunately, we cannot yet trust that an empty or None requires_dist
```
dimbleby, line 493: This

radoering: I agree. In https://github.com/python-poetry/poetry/pull/9019#discussion_r1501861491 and the following we concluded that `None` means "I don't know" and `[]` means we are sure that there are no dependencies so this change really feels like a step backwards.

abn: Let me check the test cases again, but disabling net access ended up using the mocked http calls. Which led to this code path being triggered causing metadata to be not correctly discovered since the metadata did not contain a requires-dist. 

But, either way, let me see if I can revert this.

abn: Okay, seems the problem fixed itself with the refreshed test data. So, dropping the commit for now.

Secrus (approved):

PR #9099, opened 2024-03-02
remove the setup reader

dimbleby:
fixes #9066, probably others

the more I think about it the more I reckon that the setup reader and mechanics around it are buggy beyond repair.

eg I realised today that even though the setup reader might correctly identify that some piece of metadata is not set in either `setup.py` or `setup.cfg` - it might still be completely wrong to declare that the value is missing, because this code knows nothing about PEP621-style `pyproject.toml`.

indeed, an approximately empty `setup.py` is [recommended](https://packaging.python.org/en/latest/guides/modernize-setup-py-project/#what-if-something-that-can-not-be-changed-expects-a-setup-py-file) by the "how to modernize" guide as "perfectly fine".

as discussed in #9066, the python ecosystem is in a different place than it was when this was written: pretty much all of the popular packages are providing wheels these days and so the value in trying so hard to avoid building from source is much reduced.

(It might be worth someone making something along the lines of #7670 work at some point.  That tries to read metadata from a PEP-621 `pyproject.toml` without performing a build.  In the current ecosystem, if you can get all that you need that way then you likely should.)

meanwhile let's rip out the code that gives wrong answers, and just take the performance hit.

PR #9148, opened 2024-03-13
tests: replace unused setup.py files

abn:
This change should reduce the performance hit we take once we merge #9099.


abn:
The change previously failing on windows has been isolated into #9149.

Secrus (approved):

PR #9149, opened 2024-03-13
test: enforce tmp working directory

abn:
This change ensures that working director when executing a test is set to `tmp_path` fixture value.

Additionally, we introduce the `set_project_context` fixture that allows for test cases to switch cwd to the required fixture project (copied to `tmp_path` by default) to ensure correct source files are used. Previously, some tests accidentally picked up incorrect pyproject.toml files as a result.

### DIFF
--- a/tests/helpers.py
+++ b/tests/helpers.py
@@ -338,3 +338,32 @@ def http_setup_redirect(
             status=status_code,
             body=redirect_request_callback,
         )
+
+
+@contextlib.contextmanager
+def switch_working_directory(path: Path, remove: bool = False) -> Iterator[Path]:
+    original_cwd = Path.cwd()
+    os.chdir(path)
+
+    with contextlib.suppress(Exception) as exception:
+        yield path
+
+    os.chdir(original_cwd)
+
+    if remove:
+        shutil.rmtree(path, ignore_errors=True)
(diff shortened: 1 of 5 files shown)


tests/helpers.py
```diff
@@ -338,3 +338,32 @@ def redirect_request_callback(
             status=status_code,
             body=redirect_request_callback,
         )
+
+
+@contextlib.contextmanager
+def switch_working_directory(path: Path, remove: bool = False) -> Iterator[Path]:
```
Secrus, line 344: How about using the existing code [here](https://github.com/python-poetry/poetry/blob/64d8931538baaf91e4021ad245389c544e6e788c/src/poetry/utils/helpers.py#L65) instead of introducing new one?

abn: The functionality is different, and the "remove" use case is only required for the test suite.

Secrus: ok, now the question remains, where is a test for a test helper? `test_test_helpers.py` maybe?

abn: We can potentially deal with in a separate issue? But I feel like that is overkill since the code is not overly complex.

Secrus (approved):

Issue #9120, opened 2024-03-05
`poetry lock --no-update` fails on FIPS-enabled system when the only known hash for a dependency is `md5`


PabloAlexis611:
### Description

Using custom sources for a private repo (private mirror of PyPI), we need to perform a `poetry lock --no-update` command so that the `poetry.lock` file gets regenerated with the new source URL data. This stems from internet-enabled development, but air-gapped builds.

There might be the case that one of the dependencies pulled from this private mirror of PyPI only contains an `md5` known hash.

This then attempts to generate an MD5 hash on a FIPS-enabled system, causing failure.

### Workarounds

To manually patch the `src/poetry/repositories/http_repository.py` file at https://github.com/python-poetry/poetry/blob/28d5c007d8a73fac466deedd3b691fbf14bd9fff/src/poetry/repositories/http_repository.py#L377

with the following content:
``` python
known_hash = getattr(hashlib, hash_name)(usedforsecurity=False) if hash_name else None
```

Which is **not** what I'd want to do in a FIPS-enabled system.

### Poetry Installation Method

pipx

### Operating System

RedHat 8.9

### Poetry Version

Poetry (version 1.8.1)

### Poetry Configuration

```bash session
cache-dir = "/home/appuser/.cache/pypoetry"
certificates.pypy_example_url_org.cert = "/path/to/cert.pem"
experimental.system-git-client = false
installer.max-workers = null
installer.modern-installation = true
installer.no-binary = null
installer.parallel = true
keyring.enabled = true
solver.lazy-wheel = true
virtualenvs.create = true
virtualenvs.in-project = null
virtualenvs.options.always-copy = false
virtualenvs.options.no-pip = false
virtualenvs.options.no-setuptools = false
virtualenvs.options.system-site-packages = false
virtualenvs.path = "{cache-dir}/virtualenvs"  # /home/someuser/.cache/pypoetry/virtualenvs
virtualenvs.prefer-active-python = false
virtualenvs.prompt = "{project_name}-py{python_version}"
warnings.export = true
```

### Python Sysconfig

```bash session
Platform: "linux-x86_64"
Python version: "3.10"
Current installation scheme: "posix_prefix"

Paths: 
        data = "/usr/local"
        include = "/usr/local/include/python3.10"
        platinclude = "/usr/local/include/python3.10"
        platlib = "/usr/local/lib/python3.10/site-packages"
        platstdlib = "/usr/local/lib/python3.10"
[...]

PR #9152, opened 2024-03-14
Improve error handling on FIPS systems

PabloAlexis611:
# Description
This PR improves error handling, particularly in FIPS-enabled systems that are unable to calculate MD5 hashes used for security reasons. If the provided known hash in a link for a dependency is MD5, this PR makes it so that FIPS systems can fallback to SHA256 instead.

# Pull Request Check List

Resolves: #9120 

### DIFF
--- a/tests/repositories/test_http_repository.py
+++ b/tests/repositories/test_http_repository.py
@@ -153,3 +153,33 @@ def test_get_info_from_wheel_state_sequence(mocker: MockerFixture) -> None:
     repo._get_info_from_wheel(link)
     assert mock_metadata_from_wheel_url.call_count == 5
     assert mock_download.call_count == 4
+
+
+@pytest.mark.parametrize(
+    "mock_hashes",
+    [
+        None,
+        {"sha256": "e216b70f013c47b82a72540d34347632c5bfe59fd54f5fe5d51f6a68b19aaf84"},
+        {"md5": "be7589b4902793e66d7d979bd8581591"},
+    ],
+)
+def test_calculate_sha256(
+    mocker: MockerFixture, mock_hashes: dict[str, Any] | None
+) -> None:
+    filename = "poetry_core-1.5.0-py3-none-any.whl"
+    filepath = MockRepository.DIST_FIXTURES / filename
+    mock_download = mocker.patch(
+        "poetry.repositories.http_repository.download_file",
+        side_effect=lambda _, dest, *args, **kwargs: shutil.copy(filepath, dest),
+    )
+    domain = "foo.com"
+    link = Link(f"https://{domain}/{filename}", hashes=mock_hashes)
+    repo = MockRepository()
+
+    calculated_hash = repo.calculate_sha256(link)
+
+    assert mock_download.call_count == 1
+    assert (
+        calculated_hash
+        == "sha256:e216b70f013c47b82a72540d34347632c5bfe59fd54f5fe5d51f6a68b19aaf84"
+    )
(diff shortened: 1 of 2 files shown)


tests/repositories/test_http_repository.py
```diff
@@ -153,3 +153,33 @@ def test_get_info_from_wheel_state_sequence(mocker: MockerFixture) -> None:
     repo._get_info_from_wheel(link)
     assert mock_metadata_from_wheel_url.call_count == 5
     assert mock_download.call_count == 4
+
+
+@pytest.mark.parametrize(
+    "mock_hashes",
+    [
+        None,
+        {"sha256": "e216b70f013c47b82a72540d34347632c5bfe59fd54f5fe5d51f6a68b19aaf84"},
```
abn, line 162: Shouldn't this be a case where repo does not provide a sha256?

tests/repositories/test_http_repository.py
```diff
@@ -153,3 +153,33 @@ def test_get_info_from_wheel_state_sequence(mocker: MockerFixture) -> None:
...
+    filepath = MockRepository.DIST_FIXTURES / filename
+    mock_download = mocker.patch(
+        "poetry.repositories.http_repository.download_file",
+        side_effect=lambda _, dest, *args, **kwargs: shutil.copy(filepath, dest),
+    )
+    domain = "foo.com"
+    link = Link(f"https://{domain}/{filename}", hashes=mock_hashes)
+    repo = MockRepository()
+
+    calculated_hash = repo.calculate_sha256(link)
```
abn, line 179: Shouldn't we force/mock the md5 error?

tests/repositories/test_http_repository.py
PabloAlexis611: Since there were no unit tests for this function as far as I could see, the unit test will run with 3 cases (no hashes in link, sha256 in link, and md5 in link)

tests/repositories/test_http_repository.py
PabloAlexis611: @abn Agreed with this.

Added test in 823fc6df712cc74e89441e7d3561c705ba75d7d9.

PabloAlexis611: @abn let me know if there' any other review action I need to take

tests/repositories/test_http_repository.py
```diff
@@ -153,3 +153,71 @@ def test_get_info_from_wheel_state_sequence(mocker: MockerFixture) -> None:
...
+    domain = "foo.com"
+    link = Link(
+        f"https://{domain}/{filename}",
+        hashes={"md5": "be7589b4902793e66d7d979bd8581591"},
+    )
+    repo = MockRepository()
+
+    calculated_hash = repo.calculate_sha256(link)
+
+    assert suppressed_value_error
```
radoering, line 217: Is this really necessary? Isn't it enough to check `mock_hashlib.call_count`? (This nonlocal thing feels a bit like a code smell.)

PabloAlexis611: Since the error is suppressed there's no way to know to know that the ValueError was raised (i.e., `with pytest.raises(ValueError)`).

It _should_ indeed be enough to assert that `mock_hashlib.call_count` was called though, with the assumption that the mock is working as expected - which is what I was avoiding by introducing a `nonlocal` variable (see [PEP 3104](https://peps.python.org/pep-3104/)) to test the error was raised but suppressed.

I can remove it though - just let me know if you still think it's unnecessary (:

Btw, just renamed this mocked function to `mock_hashlib_md5` to be more specific.

radoering: After sleeping on it, I think I'm ok with it. Maybe, just rename `suppressed_value_error` to `raised_value_error` because that's what we actually care about?

PabloAlexis611: Done @radoering

radoering (approved):

PR #9163, opened 2024-03-15
deps: update build

abn:


dimbleby:
pedantically: that does not relax the constraint, `^1.0.3` already allowed `1.1.1`.  So actually it narrows the constraint!

pyproject.toml
```diff
@@ -33,7 +33,7 @@ python = "^3.8"
 
 poetry-core = "1.9.0"
 poetry-plugin-export = "^1.7.0"
-build = "^1.0.3"
+build = "^1.1"
```
radoering, line 36: Shall we require `"^1.1.1"` because `1.1.0` is yanked?

abn: Sounds reasonable

radoering (approved):

PR #9164, opened 2024-03-15
remove advice re seeding keyring backends

dimbleby:
I can't make sense of the advice that this pull request removes: so this is an invitation either to remove it or to explain it!

Surely all that matters to poetry is that the relevant keyring backend is in its own virtual environment?  What reason would there be for also seeding the project virtual environment with that backend?

this comes from https://github.com/python-poetry/poetry/pull/4086 so I guess I should tag @Darsstar, though of course three years later it would be completely reasonable to have forgotten or stopped caring.

even if right in principle the doc is not enormously helpful.  As far as I am aware the imagined plug-in remains imaginary - and I would not know from reading this text what I was supposed to do anyway.  So I might encourage removing it in any case.

### DIFF
--- a/docs/repositories.md
+++ b/docs/repositories.md
@@ -541,12 +541,6 @@ Microsoft's [artifacts-keyring](https://pypi.org/project/artifacts-keyring/) get
 valid credentials. It will need to be properly installed into Poetry's virtualenv,
 preferably by installing a plugin.
 
-If you are letting Poetry manage your virtual environments you will want a virtualenv
-seeder installed in Poetry's virtualenv that installs the desired keyring backend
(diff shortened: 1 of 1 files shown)


Darsstar:
Seeding the venv isn't relevent with the [modern installer](https://python-poetry.org/blog/announcing-poetry-1.4.0/#faster-installation-of-packages), but as long as the old installer is still supported it should probably be updated instead of removed.

dimbleby:
... because the pip-based installer is using pip from the project environment: and the issue is not how to get poetry to use the backend, it is how to get that pip to use the backend?

thanks, that just about makes sense.  

given that the pip-based installer is now deprecated I would vote for removing this advice anyway - I am currently in a `$WORK` conversation where it is confusing folk, so I know the downside to it being there, and I think the number of people benefitting from it is very likely zero.

radoering (approved):

PR #9175, opened 2024-03-19
tests/repo: generate metadata file fixtures

abn:
This change ensures that all dist fixtures also have associated metadata files for testing. This takes care of most if not all of the review comments raised in https://github.com/python-poetry/poetry/pull/9170#pullrequestreview-1944491228.


tests/repositories/fixtures/pypi.org/generate.py
```diff
@@ -518,6 +523,30 @@ def bdist_check(filename: str) -> bool:
...
+            return None
+
+        metadata, _ = parse_email(response.content)
+        content = response.content.decode(encoding="utf-8").replace(
+            metadata["description"], ""
+        )
+
+        FIXTURE_PATH_REPOSITORIES_PYPI.joinpath(
+            "metadata", f"{link.filename}.metadata"
+        ).write_text(content, encoding="utf-8")
```
dimbleby, line 548: `write_bytes()`

probably everything is utf8 encoded all along so it works out fine, but from a hash-preserving point of view working with bytes throughout should be reliable

abn: The dropping of the description is not security related, never was. We really just don't need that information and it's frustrating when it shows up in searches. 

As for hashes, are metadata file hashes used anywhere? The tests passing seem to indicate they are not.

The binary file hashes are recomputed because we empty out the source. And the new hashes are used in fixtures.

dimbleby: hashes are used in the branch that relies on metadata instead of the json api - if the hash doesn't match the value that the simpl page gives, then it is ignored

abn: Ah okay. Updated to ensure hashes are updated along with binary hashes.

abn: Will drop the version specific json files and their generation once #9170 is merged.

radoering (approved):

Issue #1444, opened 2019-10-07
Installing *.whl from GitHub release url from private repo


hc2683:
- [ x] I have searched the [issues](https://github.com/sdispater/poetry/issues) of this repo and believe that this is not a duplicate.
- [ x] I have searched the [documentation](https://poetry.eustace.io/docs/) and believe that my question is not covered.

## Feature Request

I think it would be immensely useful for development purposes for it to be possible to add the ability to install from a GitHub release. It doesn't fall under the banner of the git feature because we don't want to actually copy or compile the repository, and the url= functionality in project.toml doesn't work because that doesn't pass the correct GitHub credentials (when required).

Since Github releases ask for compiled binaries, installation files etc., being able to install from these links would be immensely useful when in the development stage (where using a public or private repo is not yet possible).

conraddd:
Any update on this feature request? It would be very useful if poetry supports github private repo's release wheel file.

ruohola:
Any update on this?

el-iot:
I would also be very interested in this :+1:

Philippe-Neveux:
Hello everyone, this feature also interested me :)

PR #9202, opened 2024-03-22
utils: ensure Downloader defaults to Authenticator

abn:
This change allows correct usage of http credentials when downloading files.

Resolves: #1444

```sh
poetry config repositories.gh-release-your-org-project https://github.com/your-org/project/releases/download/
poetry config http-basic.gh-release-your-org-project username password
poetry add https://github.com/your-org/project//releases/download/1.1.2/project-1.1.2-py3-none-any.whl
```

### Testing #1444
Modify commands to your specific case.

#### Using pipx
```sh
pipx install --suffix=@9202 'poetry @ git+https://github.com/python-poetry/poetry.git@refs/pull/9202/head'
```

#### Using a container (podman | docker)
```sh
podman run --rm -i --entrypoint bash docker.io/python:3.12 <<EOF
set -xe
python -m pip install --disable-pip-version-check -q git+https://github.com/python-poetry/poetry.git@refs/pull/9202/head

poetry config repositories.gh-release-your-org-project https://github.com/your-org/project/releases/download/
poetry config http-basic.gh-release-your-org-project username password

poetry new foobar
pushd foobar
poetry add [pycowsay](https://github.com/your-org/project//releases/download/1.1.2/project-1.1.2-py3-none-any.whl)
EOF
```

### DIFF
--- a/src/poetry/packages/direct_origin.py
+++ b/src/poetry/packages/direct_origin.py
@@ -9,6 +9,7 @@ from poetry.core.packages.utils.link import Link
 
 from poetry.inspection.info import PackageInfo
 from poetry.inspection.info import PackageInfoError
+from poetry.utils.authenticator import get_default_authenticator
 from poetry.utils.helpers import download_file
 from poetry.utils.helpers import get_file_hash
 from poetry.vcs.git import Git
@@ -56,6 +57,7 @@ def _get_package_from_git(
 class DirectOrigin:
     def __init__(self, artifact_cache: ArtifactCache) -> None:
         self._artifact_cache = artifact_cache
+        self._authenticator = get_default_authenticator()
 
     @classmethod
     def get_package_from_file(cls, file_path: Path) -> Package:
@@ -74,10 +76,13 @@ class DirectOrigin:
     def get_package_from_directory(cls, directory: Path) -> Package:
         return PackageInfo.from_directory(path=directory).to_package(root_dir=directory)
 
+    def _download_file(self, url: str, dest: Path) -> None:
+        download_file(url, dest, session=self._authenticator)
+
     def get_package_from_url(self, url: str) -> Package:
         link = Link(url)
         artifact = self._artifact_cache.get_cached_archive_for_link(
-            link, strict=True, download_func=download_file
+            link, strict=True, download_func=self._download_file
         )
 
         package = self.get_package_from_file(artifact)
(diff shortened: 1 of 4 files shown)


radoering:
Actually, (without trying it) I suppose #1444 has already been solved some time ago (or this change does not solve it) because it looks like we always pass a session to `Downloader` so this might be a change only affecting tests. I wonder if we should just disallow to pass no session because we don't want to use it in practice or if we want to keep it just to simplify tests?

radoering:
Argh, I missed one dynamic usage that does not pass a session:

https://github.com/python-poetry/poetry/blob/main/src/poetry/packages/direct_origin.py#L80

However, we could think about changing that.

abn:
We do not always pass in a session. 

https://github.com/python-poetry/poetry/blob/9b3893a46f20d422ae1912109428b00bcce320b5/src/poetry/packages/direct_origin.py#L80

This is what caused #1444 afaict.

abn:
I don't think changing it is trivial. And we want to keep allowing the utility function to optionally pass in a session.

radoering:
> I don't think changing it is trivial.

What about creating an `Authenticator` in `DirectOrigin.__init__` and then
```
        download_func = functools.partial(download_file, session=self._authenticator)
        artifact = self._artifact_cache.get_cached_archive_for_link(
            link, strict=True, download_func=download_func
        )
```

> And we want to keep allowing the utility function to optionally pass in a session.

We probably should avoid using it without passing a session for performance reasons but if you think it makes sense to keep it for convenience, I'm ok with it.

abn:
> What about creating an Authenticator in DirectOrigin.__init__

Not opposed to that. But, I do wonder if that actually gains us much since the cache already will be set after the first download and the url never changes. 

Reckon we should anyway? 

> We probably should avoid using it without passing a session for performance reasons but if you think it makes sense to keep it for convenience, I'm ok with it.

I am not so sure there is a discernable performance hit for one-shot downloads. But yes, I do think keeping it for convenience is worthwhile.

radoering:
> But, I do wonder if that actually gains us much since the cache already will be set after the first download and the url never changes.

One `DirectOrigin` instance is used for all URL downloads during resolving so it's not just one URL. It's probably not that much of a gain but it shouldn't hurt since it will be similar to what `Executor` is doing.

abn:
@radoering made trhe changes, used `get_default_authenticator` too.

radoering (approved):

Issue #9195, opened 2024-03-22
Better handling of unrecognized metadata versions


jfly:
### Issue Kind

Change in current behaviour

### Description

See https://github.com/jfly/2024-03-19-poetry-build-deps-issue for details of a saga I went through this week to figure out why poetry recently stopped installing the transitive dependencies of hatchling.

tl;dr: Hatchling now uses metadata version 2.3, and existing installations of poetry may or may not handle metadata version 2.3 well, depending on if they were installed before or after 2024-03-03 (when [pkginfo 1.10.0](https://pypi.org/project/pkginfo/1.10.0/) was released).

### Impact

The next 1.8 release of poetry will be guaranteed to handle metadata 2.3 correctly thanks to https://github.com/python-poetry/poetry/pull/9131. However, upgrading to a fixed version of poetry isn't enough, as your poetry cache may already be poisoned with bad information.

In my option, it would be better if poetry just crashed clearly when presented with unrecognized metadata. I believe the current `pkginfo` api doesn't make this possible to do, but I've asked upstream if this is something they'd be willing to expand their API to support: https://bugs.launchpad.net/pkginfo/+bug/2058697.

Is this something poetry would be interested in handling differently? If so, please chime in on [this bug report](https://bugs.launchpad.net/pkginfo/+bug/2058697) upstream with pkginfo.

### Workarounds

N/A

jfly:
For the record, the maintainer of `pkginfo` is open to expanding the existing `Dsitribution` api to allow tools like poetry to do something more nuanced here. See https://bugs.launchpad.net/pkginfo/+bug/2058697/comments/2

dimbleby:
Not sure that this issue is tracking anything useful in this repository.  If the `pkginfo` API changes then I expect that poetry will either pick that up; or use something else (I see that a recent PR preferred `packaging.metadata`).

`pkginfo` throwing an exception on unrecognised metadata sounds fine, but for it to make a meaningful difference to end users someone will still have to write code here to turn that into a helpful error message.  

You could do that today if you wanted to: where the code currently says "Unable to create a package with no name" you could just as well add a hint that it probably is because of faulty or unexpected metadata.

jfly:
> Not sure that this issue is tracking anything useful in this repository

That's fair. Perhaps I should have filed this as a discussion, but to try to state this as a concrete bug/user experience issue, I might rephrase this as "poetry silently ignores transitive dependencies of package A when package A uses an unrecognized metadata format".

[...]

PR #9203, opened 2024-03-22
inspection: error on unsupported metadata versions

abn:
Resolves: https://github.com/python-poetry/poetry/issues/9195

This change also handles a previously unidentified bug. Previously, when processing wheel files, `PackageInfo` incorrectly
returned an empty instance. This change ensures a `PackageInfoError` is correctly raised. See 36bf3251f548d27a81677be32a87013a4b41f101.


radoering (approved):

src/poetry/inspection/info.py
```diff
@@ -266,6 +266,11 @@ def _from_distribution(
 
         :param dist: The distribution instance to parse information from.
         """
+        if dist.metadata_version not in pkginfo.distribution.HEADER_ATTRS:
+            # This check can be replaced once upstream implements strict parsing
+            # https://bugs.launchpad.net/pkginfo/+bug/2058697
+            raise ValueError("Unknown metadata version")
```
jfly, line 272: I think it would be nice to include the metadata version number here: https://github.com/python-poetry/poetry/pull/9226.

jfly (commented):
Thanks, @abn <3 I love this change. Fail-fast!

I tested this out with the [test harness I put together](https://github.com/jfly/2024-03-19-poetry-build-deps-issue?tab=readme-ov-file#2023-03-24), and it works great!

```
$ docker compose run poetry-1.9.x.prerelease-pkginfo-1.9.6 -c 'cd backend-with-mydevpi && poetry lock -vv'
...
Using virtualenv: /opt/venv
Updating dependencies
Resolving dependencies...
   1: fact: backend-with-mydevpi is 0.1.0
   1: derived: backend-with-mydevpi
   1: fact: backend-with-mydevpi depends on hatchling (*)
   1: selecting backend-with-mydevpi (0.1.0)
   1: derived: hatchling
   1: Version solving took 0.070 seconds.
   1: Tried 1 solutions.

  ValueError

  Package('hatchling', '1.22.4') is not in list

  at /opt/poetry/lib/python3.11/site-packages/poetry/repositories/legacy_repository.py:57 in package
       53│         Note that this will be cached so the subsequent operations
       54│         should be much faster.
       55│         """
       56│         try:
    →  57│             index = self._packages.index(Package(name, version))
       58│
       59│             return self._packages[index]
       60│         except ValueError:
       61│             package = super().package(name, version, extras)

...

PackageInfoError

Unable to determine package info for path: /tmp/tmpdkj0o8_8/hatchling-1.22.4-py3-none-any.whl

Unknown metadata version

at /opt/poetry/lib/python3.11/site-packages/poetry/inspection/info.py:554 in from_wheel
    550│         try:
    551│             wheel = pkginfo.Wheel(str(path))
    552│             return cls._from_distribution(wheel)
    553│         except ValueError as e:
  → 554│             raise PackageInfoError(path, e)
    555│
    556│     @classmethod
    557│     def from_bdist(cls, path: Path) -> PackageInfo:
    558│         """
```

PR #9226, opened 2024-03-24
include metadata version in "unknown metadata" error message

jfly:
I thought this would be nice just to help people debug things in the future.

# Pull Request Check List

refs: https://github.com/python-poetry/poetry/issues/9195, https://github.com/python-poetry/poetry/pull/9203/

- [x] Added **tests** for changed code.
- [ ] Updated **documentation** for changed code.

PR #9229, opened 2024-03-24
[1.8 backport] inspection: error on unsupported metadata versions

neersighted:
Backport #9203 to 1.8.

---

Resolves: https://github.com/python-poetry/poetry/issues/9195

This change also handles a previously unidentified bug. Previously, when processing wheel files, `PackageInfo` incorrectly
returned an empty instance. This change ensures a `PackageInfoError` is correctly raised. See 36bf3251f548d27a81677be32a87013a4b41f101.


neersighted:
Note that this is without the test as it did not apply cleanly; let me know if you want me to try to bring it forward as well.

abn (approved):

jfly (changes requested):
I don't see the `if dist.metadata_version not in pkginfo.distribution.HEADER_ATTRS` check that's a key piece of https://github.com/python-poetry/poetry/pull/9203/. Is that intentionally omitted here?

src/poetry/inspection/info.py
```diff
@@ -245,6 +245,11 @@ def _from_distribution(
 
         :param dist: The distribution instance to parse information from.
         """
+        if dist.metadata_version not in pkginfo.distribution.HEADER_ATTRS:
+            # This check can be replaced once upstream implements strict parsing
+            # https://bugs.launchpad.net/pkginfo/+bug/2058697
+            raise ValueError("Unknown metadata version")
```
jfly, line 251: nit: consider including the version number here as in https://github.com/python-poetry/poetry/pull/9226

jfly (approved):
I haven't manually tested this version with my [test harness](https://github.com/jfly/2024-03-19-poetry-build-deps-issue), but the code LGTM!

radoering (approved):

PR #9235, opened 2024-03-24
ci: better and DRYer matrixes

neersighted:
Hopefully drastically increase the maintainability of our CI workflows by DRYing up common bits, and taking better advantage of parallelism.

Note some big changes:
* The main 'tests' workflow is now much wider and more parallel, and should surface meaningful results faster (e.g. mypy and pytest in parallel)
* We now have a single 'Status' job intended to be used as a required check; this means we no longer need to keep updating branch protection rules
* Likewise, skipped jobs should now actually show that they were skipped in the UI
* pytest-xdist is now using 'logical' CPU cores as well (the distinction is less meaningful with VM vCPUs, so we should use all of them)
* mypy data is cached, resulting in a 5x speedup on Windows
* We're using the 'Poetry managed' venv for now, as this is all that setup-python can cache

### DIFF
--- a/.github/actions/bootstrap-poetry/action.yml
+++ b/.github/actions/bootstrap-poetry/action.yml
@@ -1,0 +1,34 @@
+name: Bootstrap Poetry
+description: Configure the environment with the specified Python and Poetry version.
+
+inputs:
+  python-version:
+    description: Desired Python version
+    default: "3.12"
+  python-latest:
+    description: Use an uncached Python if a newer match is available
+    default: 'false'
+  poetry-spec:
+    description: pip-compatible installation specification to use for Poetry
+    default: 'poetry'
+
+runs:
+  using: composite
+  steps:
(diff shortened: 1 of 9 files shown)


.github/actions/bootstrap-poetry/action.yml
```diff
@@ -0,0 +1,34 @@
+name: Bootstrap Poetry
```
abn, line 1: Can we also use the cache action to cache `~/.cache/pypoetry/{artifacts,cache}` in this action till we can hopefully alter `setup-python` upstream?

abn: Actually scratch that we might not need it right now because the installs are really fast already with the default venv caching.

neersighted: Yes, but I want to keep that separate for now

abn (approved):

PR #9237, opened 2024-03-25
ci/release: implement trusted publishing

neersighted:
Completely rewrite the release workflow. There are a couple key differences now:
* Releases are triggered by GitHub releases; which can refer to an existing tag, or create one on the fly.
* Credentials are no longer needed to perform a release, instead we use the PyPI trusted publishing flow.
* The build and upload tasks are split to ensure the credentials are safeguarded.
* Uploads require approval from one other maintainer (beyond the person who triggered the build).
* We now also upload release artifacts to the GitHub release.

Note that PyPI will only accept uploads from the 'pypi' environment, triggered in this workflow file.

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -19,8 +19,19 @@ jobs:
           path: dist/
           if-no-files-found: error
 
-  upload:
-    name: Upload
+  upload-github:
+    name: Upload (GitHub)
+    runs-on: ubuntu-latest
+    needs: build
+    steps:
+      - uses: actions/download-artifact@v4
+        with:
+          name: distfiles
+
+      - run: gh release upload ${{ github.event.release.tag_name }} dist/*.{tar.gz,whl}
+
+  upload-pypi:
+    name: Upload (PyPI)
     runs-on: ubuntu-latest
     environment: pypi
     permissions:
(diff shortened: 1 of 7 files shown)


.github/workflows/release.yaml
```diff
@@ -1,36 +1,48 @@
 name: Release
 
 on:
-  push:
-    tags:
-      - '*.*.*'
+  release:
+    types: [published]
```
abn, line 5: So, would have to create the release manually after a tag is pushed?

abn: So release workflow would be

1. Create release PR (version bump, changelog updates).
2. Merge.
3. Create release via UI w/ changelog.

I am guessing later we can do a workflow similar to backport where we can label it as a release so it triggers a creation of a tag from the version and corresponding release.

neersighted: The release can create the tag; I am going to suggest that we restrict tags such that they can only be created from the web UI.

.github/workflows/release.yaml
```diff
@@ -1,36 +1,49 @@
...
-  release:
-    name: Release
+  build:
+    name: Build
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v4
 
-      - uses: ./.github/actions/bootstrap-poetry
+      - run: pipx run build
```
abn, line 14: This is neat

abn (approved):

PR #9239, opened 2024-03-25
ci: use optimized caching strategy for Poetry

neersighted:
Cache our artifact/metadata cache instead of relying on the virtualenv-based strategy of setup-python, as well as misc CI tweaks.

### DIFF
--- a/.github/actions/poetry-install/action.yml
+++ b/.github/actions/poetry-install/action.yml
@@ -1,0 +1,50 @@
+name: Poetry Install
+description: Run `poetry install` with optional artifact and metadata caching
+
+inputs:
+  args:
+    description: Arguments for `poetry install`
+  cache:
+    description: Enable transparent Poetry artifact and metadata caching
+    default: 'true'
+
+outputs:
+  cache-hit:
+    description: Whether an exact cache hit occured
+    value: ${{ steps.cache.outputs.cache-hit }}
+
+runs:
+  using: composite
+  steps:
+    - run: printf 'cache-dir=%s\n' "$(poetry config cache-dir)" >> $GITHUB_OUTPUT
+      id: poetry-config
+      shell: bash
+
+    # Bust the cache every 24 hours to prevent it from expanding over time.
+    - run: printf 'date=%s\n' "$(date -I)" >> $GITHUB_OUTPUT
+      id: get-date
+      if: inputs.cache == 'true'
+      shell: bash
+
+    - uses: actions/cache@v4
+      id: cache
+      if: inputs.cache == 'true'
+      with:
+        path: |
+          ${{ steps.poetry-config.outputs.cache-dir }}/artifacts
+          ${{ steps.poetry-config.outputs.cache-dir }}/cache
+        key: poetry-${{ steps.get-date.outputs.date }}-${{ runner.os }}-${{ hashFiles('pyproject.toml', 'poetry.lock') }}
+        # The cache is cross-platform, and other platforms are used to seed cache misses.
+        restore-keys: |
+          poetry-${{ steps.get-date.outputs.date }}-${{ runner.os }}-
+          poetry-${{ steps.get-date.outputs.date }}-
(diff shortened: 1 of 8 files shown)


.github/actions/poetry-install/action.yml
```diff
@@ -0,0 +1,42 @@
...
+    value: ${{ steps.cache.outputs.cache-hit }}
+
+runs:
+  using: composite
+  steps:
+    - run: printf 'cache-dir=%s\n' "$(poetry config cache-dir)" >> $GITHUB_OUTPUT
+      id: poetry-config
+      shell: bash
+
+    - uses: actions/cache@v4
```
abn, line 23: Can't we move this to bootstrap? Because we really do not want caches changing during the execution of the job right? And the call to cache action every step is rather overkill.

.github/workflows/tests.yaml
```diff
@@ -62,6 +58,18 @@ jobs:
...
+    name: Check build
+    runs-on: ubuntu-latest
+    if: needs.changes.outputs.src == 'true'
+    needs: lockfile
+    steps:
+      - uses: actions/checkout@v4
+
+      - run: pipx run build
+
+      - run: pipx run twine check --strict dist/*
```
abn, line 71: I am a bit conflicted about this. We should actually rely on build producing valid artifacts and not actually need twine here.

If we do these, we are better off just doing `pipx run poetry build`.

.github/workflows/tests.yaml
```diff
@@ -62,6 +58,18 @@ jobs:
 
       - run: poetry check --lock
 
+  build:
+    name: Check build
+    runs-on: ubuntu-latest
+    if: needs.changes.outputs.src == 'true'
```
abn, line 64: Don't think checking on src changes is valuable. Only thing that really impacts a build to produce bad artifacts is changes to `pyproject.toml` right?

abn: I would ask if this is really adding any value for us especially if `poetry check` passes.

.github/actions/poetry-install/action.yml
neersighted: We can't do this in `bootstrap-poetry`, because composite actions trigger post-run actions when they complete. In other words, the cache would never cache because `poetry install` hasn't happened yet.

With the new `poetry-install` composite action, we actually finalize the cache after `poetry install` and before proceeding with the rest of the job.

I'm not sure what you mean about excessive; we only perform this step when `poetry install` happens.

.github/workflows/tests.yaml
neersighted: This is intended to test the PEP 517/pyproject-hooks flow. We don't use `poetry build` here as we have plenty of builder tests, but our build flow special-cases/partially ignores `build-system.requires`.

This step is really here to ensure other spec-compliant tools can build us.

.github/workflows/tests.yaml
neersighted: You're right, this should be `project`.

.github/workflows/tests.yaml
```diff
@@ -62,6 +58,18 @@ jobs:
...
+    name: Check external build
+    runs-on: ubuntu-latest
+    if: needs.changes.outputs.project == 'true'
+    needs: lockfile
+    steps:
+      - uses: actions/checkout@v4
+
+      - run: pipx run build
+
+      - run: pipx run twine check --strict dist/*
```
abn, line 71:
```suggestion
      - run: pipx run twine check --strict dist/*
      
      - run: pipx install --suffix=@build  --strict dist/poetry-*.tar.gz
      
      - run: poetry@build about | grep $(poetry version --short)
```

neersighted: `pipx install` will build a second time from sdist, which seems wasteful. The version check is simple, but seems like it might be appropriate.

abn (approved):

PR #9251, opened 2024-03-29
refactor: use `typing_extensions.Self`

bswck:
Used `typing_extensions.Self` instead of `(self: T, ...) -> T`.
`typing_extensions` has a special stub in typeshed and is seen by mypy even if not installed as a direct/transitive dependency, so there's no need to change anything in dependency specs.
`typing_extensions` was used instead of `typing` because the `Self` type was introduced in 3.11 with PEP 673 and `typing_extensions` offers a backport for 3.8+, which is our case.

`Operation.skip()` and `Operation.unskip()` have an anti-pattern, i.e. they return `Self` despite operating in-place.
This will be addressed in a separate PR; I need an idea how to address the API incompatibility that would bring.

### DIFF
--- a/src/poetry/inspection/lazy_wheel.py
+++ b/src/poetry/inspection/lazy_wheel.py
@@ -14,7 +14,6 @@ from typing import TYPE_CHECKING
 from typing import Any
 from typing import BinaryIO
 from typing import ClassVar
-from typing import TypeVar
 from typing import cast
 from urllib.parse import urlparse
 from zipfile import BadZipFile
@@ -34,6 +33,7 @@ if TYPE_CHECKING:
 
     from packaging.metadata import RawMetadata
     from requests import Session
+    from typing_extensions import Self
 
     from poetry.utils.authenticator import Authenticator
 
@@ -168,9 +168,6 @@ class MergeIntervals:
         yield from self._merge(start, end, left, right)
 
 
-T = TypeVar("T", bound="ReadOnlyIOWrapper")
-
-
 class ReadOnlyIOWrapper(BinaryIO):
     """Implement read-side ``BinaryIO`` methods wrapping an inner ``BinaryIO``.
 
@@ -181,7 +178,7 @@ class ReadOnlyIOWrapper(BinaryIO):
     def __init__(self, inner: BinaryIO) -> None:
         self._file = inner
 
-    def __enter__(self: T) -> T:
+    def __enter__(self) -> Self:
         self._file.__enter__()
         return self
 
@@ -286,9 +283,6 @@ class ReadOnlyIOWrapper(BinaryIO):
         raise NotImplementedError
 
 
-U = TypeVar("U", bound="LazyFileOverHTTP")
-
-
 class LazyFileOverHTTP(ReadOnlyIOWrapper):
     """File-like object representing a fixed-length file over HTTP.
 
@@ -311,7 +305,7 @@ class LazyFileOverHTTP(ReadOnlyIOWrapper):
         self._session = session
         self._url = url
 
-    def __enter__(self: U) -> U:
+    def __enter__(self) -> Self:
         super().__enter__()
         self._setup_content()
         return self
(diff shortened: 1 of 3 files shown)


dimbleby:
> `typing_extensions` has a special stub in typeshed and is seen by mypy even if not installed as a direct/transitive dependency, so there's no need to change anything in dependency specs.

citation required

`mypy` has a direct dependency on `typing-extensions` so you will get away with it.  But eg if you uninstall `mypy` (and `typing-extensions`) and try `pyright` instead you will see the warning `warning: Import "typing_extensions" could not be resolved from source (reportMissingModuleSource)
`

if poetry imports from typing-extensions then it should express that dependency directly.

dimbleby:
> From the [official typing_extensions README](https://github.com/python/typing_extensions#overview) ...

no part of what you quote says "therefore you do not need `typing-extensions` to be available in your environment.",  

Most of the rest of what you write confuses stubs for typing-extensions with the typing-extensions package,

I think it is - for the moment, just about - true that it would be sufficient to put `typing-extensions` in the group of typing dependencies in pyproject.toml.  However even that makes me uncomfortable

- are future contributions going to remember that they are only allowed to import typing-extensions in `if TYPE_CHECKING:` blocks? eg typing-extensions includes plenty of decorators, it would be easy to be confused about things going wrong if we started to use one of those
- it just is not the done thing.  Possibly you can offer counter-examples, but I see plenty of projects depending on typing-extensions - including poetry itself until recently (5bad73cd93ee6d111ed723fe57e0cc8ece678cd6)

why even worry about adding the dependency?

bswck:
> no part of what you quote says "therefore you do not need typing-extensions to be available in your environment.",

It's a conclusion that isn't explicitly written.

> Most of the rest of what you write confuses stubs for typing-extensions with the typing-extensions package

Sorry, but I'm afraid you've completely missed my point.

> are future contributions going to remember that they are only allowed to import typing-extensions in if TYPE_CHECKING: blocks? eg typing-extensions includes plenty of decorators, it would be easy to be confused about things going wrong if we started to use one of those.
When `typing-extensions` is needed _at runtime_ and `typing_extensions` is unavailable _at runtime_, CI is going to fail.

> why even worry about adding the dependency?
Because it is unnecessary. Poetry already relies on a lot of packages.

If you're still hesitant that installing `typing-extensions` is **completely unnecessary**, here's my short conversation on the topic with @AlexWaygood, a [CPython Core Developer, typeshed & typing_extensions maintainer and mypy contributor](https://github.com/AlexWaygood#about-me):
image

I hope that that is convincing enough, if I'm not.

dimbleby:
as we have established, actual typecheckers do not approve of this approach: `warning: Import "typing_extensions" could not be resolved from source (reportMissingModuleSource)`

As I say, declaring the dependency only in the typing dependency group would I think be sufficient, though I still do not like it.

CI might or might not catch missing imports.  Eg it is very plausible to fail to spot a missing import of package that is made transitively available by a test dependency.

AlexWaygood:
> as we have established, actual typecheckers do not approve of this approach: `warning: Import "typing_extensions" could not be resolved from source (reportMissingModuleSource)`

I don't _think_ that warning would impact users of poetry? But yeah, it's true that that's a lint that pyright has and mypy doesn't

> * are future contributions going to remember that they are only allowed to import typing-extensions in `if TYPE_CHECKING:` blocks? eg typing-extensions includes plenty of decorators, it would be easy to be confused about things going wrong if we started to use one of those

FWIW, while @bswck is correct that it is _strictly speaking_ unnecessary to list `typing_extensions` as a dependency, I also think this is a very valid point. Importing things in if TYPE_CHECKING blocks can be useful, but it can also make code more fragile and harder to understand; that's a trade-off you'll have to make for yourselves. We work hard at `typing_extensions` to minimise breakage between versions, and it's a single-module pure-Python package, so it shouldn't be a huge dependency to add by any means. 

Anyway, I think both ways of doing it are workable, and it's up to you folks to weigh the tradeoff appropriately :-) ping me again if you need me!

dimbleby:
pyright lint seems to have been in response to a bug report saying "you are failing to detect that typing-extensions is missing" - https://github.com/microsoft/pyright/issues/7318

possibly that fix was an overshoot and they could be convinced that the lint is only desirable for imports made outside of `if TYPE_CHECKING:`

Secrus (approved):

PR #9253, opened 2024-03-30
ci: sync with poetry-core changes

neersighted:
Fixes and improvements discovered when working on python-poetry/poetry-core#717


Secrus:
Stalled

Issue #8766, opened 2023-12-08
Poetry should handle the version of python like it handles everything else


mmr-crexi:
Right now, when I start a new project, I can choose a python version.  Say I choose python ^3.12.  Oh no!  I just realized that triton and pytorch don't yet use python 3.12!  No problem, I'll just downgrade my python version in my poetry pyproject.toml file. If I do that, then I get this error:

`Current Python version (3.12) is not allowed by the project (3.11.5).`

This is nonsense.  Just downgrade the python version, treat it like any other library. I can solve the problem by deleting the directory and restarting the entire project, but that's just awful UX.  

Poetry creates a virtual environment for use when it is first run, and then that python version appears to be locked for all time and I'm having a hell of a time changing it. When people on my team are off by third digit version numbers (3.11.5 vs 3.11.6), well, that means that each of us have created incompatible python versions and we can't run each other's code.

Please, please, please, please, just allow poetry to treat python like any other library, like it does when the project is first created.

dimbleby:
what you suggest is just not possible: poetry doesn't get to choose the python version in the same way that it gets to choose the version of every requirement.

if you have a project that has set `requires-python` to an exact version of python then the solution is simple: don't do that!

mmr-crexi:
Not possible?  I'm not asking for reversing gravity or time travel, I'm asking for the ability to hard reset a poetry installation so that I can start entirely from scratch.  That is, kill everything from the ground up, and start again with a fresh python installation before going to other libraries. I can by-hand wipe virtual environments that have been set aside in directories that are clearly not intended for user intervention, but even that doesn't clear the warning that the python version in my pyproject.toml doesn't match what was in the first iteration of the file. 

The minor/minor version number problem arises when I have 3.11.6 and someone else on my team has 3.11.5, and then we say it should be version 3.11, and then when poetry lock files move around, that disagreement also causes error messages to be sprayed out about python version incompatibility.

This is only a problem because some of the libraries I want to use (triton, pytorch, etc) are not yet compatible with python 3.12, but Poetry wants to install 3.12, so I find myself having to jump through a huge series of hoops to downgrade to 3.11.

I see a lot of conversations around using pyenv, but frankly, I was under the impression that poetry is supposed to be a one-stop shop.  If I'm going to use pyenv, then why not just use virtualenvs, pip, and a makefile that can run `make clean` to wipe the venv and start anew when I need to?

dimbleby:
Not gonna happen.  

[...]

PR #9256, opened 2024-03-31
docs: clarify python version usage

DbCrWk:
# Summary
This proposed change to documentation clarifies how Poetry does and does not manage Python. I am proposing this change because of some confusion expressed in #8766.

# Pull Request Check List

Resolves: #8766

### DIFF
--- a/docs/basic-usage.md
+++ b/docs/basic-usage.md
@@ -63,8 +63,29 @@ Similarly, the traditional `MANIFEST.in` file is replaced by the `tool.poetry.re
 `tool.poetry.exclude` sections. `tool.poetry.exclude` is additionally implicitly populated by your `.gitignore`. For
 full documentation on the project format, see the [pyproject section]({{< relref "pyproject" >}}) of the documentation.
 
+### Setting a Python Version
+
+{{% note %}}
+Unlike with other packages, Poetry will not automatically install a python interpreter for you.
+If you want to run Python files in your package like a script or application, you must _bring your own_ python interpreter to run them.
+{{% /note %}}
+
 Poetry will require you to explicitly specify what versions of Python you intend to support, and its universal locking
 will guarantee that your project is installable (and all dependencies claim support for) all supported Python versions.
+Again, it's important to remember that -- unlike other dependencies -- setting a Python version is merely specifying which versions of Python you intend to support.
+
+For example, in this `pyproject.toml` file:
+
+```toml
+[tool.poetry.dependencies]
+python = "^3.7.0"
+```
+
+we are allowing any version of Python 3 that is greater than `3.7.0`.
+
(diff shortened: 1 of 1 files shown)


DbCrWk:
One thing: I was thinking about adding some language to explain _why_ this is the case, namely:
 1. For technical reasons
 2. There are some benefits if you want to have multiple environments with different Python versions for the same project.

docs/basic-usage.md
```diff
@@ -63,8 +63,29 @@ Similarly, the traditional `MANIFEST.in` file is replaced by the `tool.poetry.re
...
+### Setting a Python Version
+
+{{% note %}}
+Unlike with other packages, Poetry will not automatically install a python interpreter for you.
+If you want to run Python files in your package like a script or application, you must _bring your own_ python interpreter to run them.
+{{% /note %}}
+
 Poetry will require you to explicitly specify what versions of Python you intend to support, and its universal locking
 will guarantee that your project is installable (and all dependencies claim support for) all supported Python versions.
+Again, it's important to remember that---unlike other dependencies---setting a Python version is merely specifying which versions of Python you intend to support.
```
jsimonlane, line 75: You could combine your above revision with with this one more concisely by saying,

"Critically, setting a Python version is merely specifying which versions of Python you intend to _support._ To run your application or library, you must _bring your own_ python interpreter (for instance using `pyenv`)."

That would make it more concise.

jsimonlane:
Really helpful update. If I'd had read these when I was installing Poetry, it would have helped me a lot.

docs/basic-usage.md
```diff
@@ -5,8 +5,8 @@ type: docs
 layout: single
 
 menu:
-  docs:
-    weight: 10
+    docs:
+        weight: 10
```
radoering, line 9: That's inconsistent with other files now. I'd leave it as is in this PR.

docs/basic-usage.md
radoering, line 75:
```suggestion
Again, it's important to remember that -- unlike other dependencies -- setting a Python version is merely specifying which versions of Python you intend to support.
```

Shouldn't there be some spaces and less "-"?

radoering (changes requested):
LGTM in general. Just some small nitpicks.

docs/basic-usage.md
DbCrWk: For now, I think it would be good to have an explicit note in a callout box. I agree that your phrasing is more concise, but I think this will be a bit more visible.

DbCrWk: I'll go with what you wrote. (I think there might be some differences in how people use en-dashes vs. em-dashes, so I'll defer.)

radoering (approved):

PR #9304, opened 2024-04-12
Move to native tags gathering

Secrus:
# Pull Request Check List

Resolves: None

~~- [ ] Added **tests** for changed code.~~ New code doesn't change the behavior
~~- [ ] Updated **documentation** for changed code.~~ No user-facing changes

The scripts connected to envs bothered me for quite a while. I have found a way to remove `GET_SYS_TAGS` script, by replacing it with a native implementation. I have also included a few small improvements.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -42,7 +42,7 @@ importlib-metadata = { version = ">=4.4", python = "<3.10" }
 installer = "^0.7.0"
 keyring = "^24.3.1"
 # packaging uses calver, so version is unclamped
-packaging = ">=23.1"
+packaging = ">=24.0"
 pexpect = "^4.7.0"
 pkginfo = "^1.10"
 platformdirs = ">=3.0.0,<5"
--- a/src/poetry/utils/env/script_strings.py
+++ b/src/poetry/utils/env/script_strings.py
@@ -1,31 +1,5 @@
 from __future__ import annotations
 
-import packaging.tags
-
-
-GET_SYS_TAGS = f"""
-import importlib.util
-import json
-import sys
-
-from pathlib import Path
-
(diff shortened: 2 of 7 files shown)


pyproject.toml
```diff
@@ -42,7 +42,7 @@ importlib-metadata = { version = ">=4.4", python = "<3.10" }
 installer = "^0.7.0"
 keyring = "^24.3.1"
 # packaging uses calver, so version is unclamped
-packaging = ">=23.1"
+packaging = ">=24.0"
```
radoering, line 45: Is there a specific reason for this? (Nothing in the changelog caught my eye.)

src/poetry/utils/env/script_strings.py
```diff
@@ -54,12 +28,7 @@ def interpreter_version():
 
 
 def _version_nodot(version):
-    if any(v >= 10 for v in version):
-        sep = "_"
-    else:
-        sep = ""
-
-    return sep.join(map(str, version))
+    return "".join(map(str, version))
```
radoering, line 31: Had this always been wrong and we just did not use it?

Secrus: I was producing a wrong nodot version. I spent some time debugging why there was a difference between old and new implementation and that was the culprit. This is the implementation that `packaging` is using for that function.

pyproject.toml
Secrus: Probably overzealous of me. I can revert that if you want.

dimbleby: since poetry-core vendors `packaging` and puts its vendored copy at the front of the `sys.path`, it makes no difference what version poetry asks for...

radoering (approved):

Issue #3248, opened 2020-10-19
breaking change: invert `lock --no-update` to `lock --update` (or similar)


seansfkelley:
- [x] I have searched the [issues](https://github.com/python-poetry/poetry/issues) of this repo and believe that this is not a duplicate.
- [x] I have searched the [documentation](https://python-poetry.org/docs/) and believe that my question is not covered.

## Feature Request

Now that https://github.com/python-poetry/poetry/issues/1614 is fixed and released (yay!), I'd like to propose that for v2, there is a breaking change to make Poetry more opinionated and preserve the versions in the lockfile by default, only updating when explicitly asked to (such as with the `update` command or some kind of `--update` flag to `lock`). Doing so follows the patterns in analogous tools like Yarn and Cargo and adheres to the principle of least surprise.

See also the philosophically-related https://github.com/python-poetry/poetry/issues/2370.

Michal-Szelag-Sonos:
+1 for this feature.  Our project is pretty big where many folks, with various skills, contribute and when dependencies move in unexpected watts, it creates lots of waves.  It's better to be explicit in such situations and do exactly what the user is requesting.

artemBeletzky:
@seansfkelley Hi! Does it mean that by using "update command or some kind of --update flag to lock" I'll be able to change current versions of packages in my .lock file?

nyuszika7h:
If this is implemented, please also make `poetry add` not try to automatically update unrelated dependencies.

nishantvarma:
@nyuszika7h , `poetry add` already behaves like that no? This can be tested like this:

Current SQLAlchemy's version is `1.4.36`.
Create a `pyproject.toml` with `1.4.35` and generate a lock file.
Update `pyproject.toml` with `^1.4.0`.
Now do `poetry add hello-world`.

This doesn't touch the SQLAlchemy version.

When you **manually** add `hello-world` to pyproject.toml and do a `poetry install`, it refuses to do so stating that it is **not** in the lock file. Now, we need to lock it again and that is where the problem happens!

If this understanding is correct, we will never get into this issue `add` or `remove`, but only via manual update -- and that too only because of the need to explicitly lock. Also, I think it makes sense for both to behave the same way which is what this requirement is about.

AleCandido:
Actually, what would be the difference between:
- `poetry lock --update`: generate the lock file, updating the dependencies
- `poetry update`: update the dependencies, generating the lock file

[...]

Issue #9136, opened 2024-03-10
Poetry 2.0 removals and other breaking changes


radoering:
This issue tracks possible removals and other breaking changes we want to include _**if**_  the next release (or the one after that) will be a major release. (That does not mean that the changes listed here cannot be included in a minor release after a decent deprecation period.)

Please add a comment if you have concerns or would like to include another pending breaking change that is not on the list.

### Background & Motivation

As support for PEP 621 is slowly taking shape (see  python-poetry/poetry#9135 to track the current state) and this will be a big, significant change (even if the old way will still be supported), this might be the time for a major version bump. Even if we try to avoid obvious breaking changes, just not ignoring the `project` section anymore will be a subtle breaking change for anyone who is already using this section (for whatever reason). Further, since supporting PEP 621 requires massive changes in many areas of the code, the risk of unintentional breaking changes is higher than usual.

Long story short: If we do a major version bump anyway, we should make use of it and remove things / add breaking changes we were holding back so far.

### Candidates

Disclaimer: decision is not final and might be subject to change

change|decision|details
---|---|---
remove dependency on `poetry-plugin-export`| ✅ | python-poetry/poetry#5980, ~although python-poetry/poetry#5740 would be nice~  python-poetry/poetry#9547 is sufficient
remove deprecated source types `secondary` and `default`| ✅ | python-poetry/poetry#9698
remove support of scripts with extras| ✅ | already done in python-poetry/poetry-core#708
change `poetry add --optional` to require an extra| ✅ | required for PEP 621 support (#9135)
fail if project cannot be installed| ✅ |#9333
remove `poetry shell`| ✅ | will be moved to a plugin, #9763
remove `poetry self`| ⛔ | python-poetry/poetry#7872
make `lock --no-update` the default| ✅ | python-poetry/poetry#9327
~make `install --sync` the default~ introduce `poetry sync`| ✅ | ~#9855 or~ #9801
remove deprecated command line options| ✅⛔ | python-poetry/poetry#9732 (revoked deprecation of `--dev`, removed the rest)
remove `no-setuptools` config option | ✅ | python-poetry/poetry#9331
make include behavior more consistent | ✅ | python-poetry/poetry#9691
remove ~(or deprecate?)~ support for `x` instead of `*` as wildcard| ✅ | python-poetry/poetry#5990
prefer active python by default | ✅ | #9786
rename `experimental.system-git-client` to `system-git-client` | ✅ | #9787
[...]

Issue #3527, opened 2021-01-02
Support lock dependencies with lowest solvable versions


JorgeGarciaIrazabal:
- [x] I have searched the [issues](https://github.com/python-poetry/poetry/issues) of this repo and believe that this is not a duplicate.
- [x] I have searched the [documentation](https://python-poetry.org/docs/) and believe that my question is not covered.

### Context
When I am implementing a library, it would be super useful to run my tests with all the supported dependencies defined with the constrains in pyproject.toml. This will prevent errors in production when, for example, I use some new features of a dependency and forget to update the dependencies constrains (this is hard to identify when you are coding).

Because running the tests with all the possible combinations of all your supported libraries is virtually impossible, maybe running them with the lowest versions solvable solution and also the highest (I think this is the default behavior)

### Proposal
 - add the flag `--use-lowest-versions` to the cli entrypoints (`install`, `sync`, `lock`, `shell`,  and `run`) 
 - these commands will create and/or use a different  lock file `poetry.lowest.lock` 
 - a different venv will be created by just adding `lowest` after the project name like this:  `{project_name}-lowest-{id}-{python-version}`

### Main simple use case
In ci do the following steps

- Install environments
```
>>> poetry install 
>>> poetry install --use-lowest-versions
```
- run tests for highest versions
```
poetry run pytest  --junit-xml pytest.xml
```
- run tests for lowest versions
```
poetry run --use-lowest-versions pytest  --junit-xml pytest.xml
```

Note: I would appreciate if anyone has a better name for the flag (naming is the hardest part :P)

Also, I looked a the code and it doesn't seems to be very hard to implement, I can create a PR if you like the idea

sinoroc:
You might want to look into [_tox_](https://pypi.org/project/tox/), it is very good at this kind of things.

JorgeGarciaIrazabal:
> You might want to look into [_tox_](https://pypi.org/project/tox/), it is very good at this kind of things.

Thank you for the rapid response @sinoroc. 

tox could definitely help to run the tests, but this feature is for defining the dependencies instead. (As far as I know, you have to provide the dependencies to tox)

This feature will simplify the definition of an environment with the lowest possible dependency versions which could potentially be an environment where final users could install the package with `pip install {my_library}`. 

So for example, if my library support `flask = "^1.0.0"`  and I start using a feature that was only implemented in flask `1.1.0`, this problem will be hidden as you will normally test your library with an environment with the latest versions solved with `poetry lock`. 

[...]

PR #9327, opened 2024-04-21
lock: make `--no-update` the default behavior and introduce `--regenerate` for the previous default behavior

radoering:
# Pull Request Check List

Resolves: #3248
Relates-to: #9136

⚠️ Breaking change. Requires a major version bump. ⚠️

The most difficult decision (if we want to let this change happen) is the name of the new option (old default behavior).

**The old flag was `--no-update`. Why not call the new flag `--update`?**

Because it's not exactly what it is. It does not update the locked package versions but ignores the lock file and creates a new one from scratch. The result is the same as an update because our (only) locking strategy is "use-latest" but in case we introduce alternative locking strategies in the future (cf #3527) the result will be different. Further, by not calling it `--update` the difference to `poetry update --lock` becomes clearer.

### DIFF
--- a/src/poetry/console/commands/lock.py
+++ b/src/poetry/console/commands/lock.py
@@ -18,7 +18,10 @@ class LockCommand(InstallerCommand):
 
     options: ClassVar[list[Option]] = [
         option(
-            "no-update", None, "Do not update locked versions, only refresh lock file."
+            "regenerate",
+            None,
+            "Ignore existing lock file"
+            " and overwrite it with a new lock file created from scratch.",
         ),
         option(
             "check",
@@ -34,6 +37,8 @@ The <info>lock</info> command reads the <comment>pyproject.toml</> file from the
 current directory, processes it, and locks the dependencies in the\
  <comment>poetry.lock</>
 file.
+By default, packages that have already been added to the lock file before
+will not be updated.
 
 <info>poetry lock</info>
 """
@@ -57,6 +62,6 @@ file.
             )
             return 1
 
-        self.installer.lock(update=not self.option("no-update"))
+        self.installer.lock(update=self.option("regenerate"))
 
         return self.installer.run()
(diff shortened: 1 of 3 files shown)


src/poetry/console/commands/lock.py
```diff
@@ -18,7 +18,7 @@ class LockCommand(InstallerCommand):
 
     options: ClassVar[list[Option]] = [
         option(
-            "no-update", None, "Do not update locked versions, only refresh lock file."
+            "new", None, "Ignore existing lock file and create a new one from scratch."
```
Secrus, line 21: how about `--refresh`?

radoering: Maybe. 😃

At first I wondered whether it is unambiguous enough but thinking a bit longer about the gist of the word I start to like it.

rbebb: `--regenerate` might be a good option as well!

rbebb:
I believe that it's the right move to switch the default behavior of poetry lock to poetry lock --no-update. More often than not, users are looking to update a few packages rather than all of them. If a user updates all of the packages, the user has an increased risk of introducing a package update that contains breaking changes.

This change will result in a more user friendly tool since the behavior will be more clear and therefore less experienced users will have less confusion!

src/poetry/console/commands/lock.py
radoering: Considering the 👍, `--regenerate` might be the best choice.

@Secrus Do you have a strong opinion?

Secrus: Sounds good to me

rbebb:
The proposed changes look good to me! Is it a good time to mark this draft as ready for review?

radoering:
This PR is marked as draft to avoid that it is merged by accident because we only want to do this with a major version bump and it is very likely but not absolutely clear that the next version will be 2.0. I will mark it as ready as soon as this is clarified. The actual review shoud be quite easy because this is a simple change.

Secrus (approved):

PR #9331, opened 2024-04-23
remove no-setuptools config option

dimbleby:
it is currently mildly inconsistent that poetry defaults to installing setuptools in newly created environments, but then removing it at `poetry install --sync`

the direction of travel is that setuptools should be named as an explicit dependency if required, let's go there
- remove the config option
- don't install setuptools

maybe one for #9136 if that's going to be a thing, though personally I would not be too frightened of releasing this anyway, so long as it gets a suitable mention in the changelog


eamanu (approved):

edmorley:
Given that `poetry install --sync` immediately removes setuptools from any Poetry-created venv, this seems fine to merge before Poetry v2? And it would be in alignment with:
- https://github.com/python/cpython/pull/101039
- https://github.com/pypa/get-pip/pull/218
- https://github.com/astral-sh/uv/pull/1613

Secrus (approved):

Secrus:
@dimbleby could you please rebase?

dimbleby:
> @dimbleby could you please rebase?

@Secrus there are no merge conflicts, you can just hit the "rebase and merge" button

Issue #8637, opened 2023-11-09
`poetry install` exit code 0 (success) even if readme file does not exist and fails to install the package


roy-ht:
- **Poetry version**: Poetry (version 1.7.0)
- **Python version**: Python:  3.10.9
- **OS version and name**: Ubuntu 22.04
- **pyproject.toml**: 
```
[tool.poetry]
name = "test"
version = "1.0.0"
readme = "README_does_not_exist"
authors = ["xxx <xxx@example.com>"]
description = ""

[tool.poetry.dependencies]
python = ">=3.10,<3.11"

[build-system]
requires = ["poetry-core"]
build-backend = "poetry.core.masonry.api"
```

- [x] I am on the [latest](https://github.com/python-poetry/poetry/releases/latest) stable Poetry version, installed using a recommended method.
- [x] I have searched the [issues](https://github.com/python-poetry/poetry/issues) of this repo and believe that this is not a duplicate.
- [x] I have consulted the [FAQ](https://python-poetry.org/docs/faq/) and [blog](https://python-poetry.org/blog/) for any relevant entries or release notes.
- [x] If an exception occurs when executing a command, I executed it again in debug mode (`-vvv` option) and have included the output below.

## Issue

with Poetry 1.6.1, `poetry install` command fails with exit code != 0:

```
❯ poetry161 install
Installing dependencies from lock file

[Errno 2] No such file or directory: '/home/jovyan/repo/sarasvati/temp/README_does_not_exist'
```

but Poetry 1.7.0 do not fail:

```
❯ poetry install   
Installing dependencies from lock file

Installing the current project: test (1.0.0)
The current project could not be installed: [Errno 2] No such file or directory: '/home/jovyan/repo/sarasvati/temp/README_does_not_exist'
If you do not want to install the current project use --no-root
```

I guess this behavior was introduced by https://github.com/python-poetry/poetry/pull/8369 .

In daily development, it's ok, one can see the warning messages.

But in my usecase, poetry install in Dockerfile,
`docker build` doesn't fail if i forget to copy README like `COPY README.md`.

This result is very confusing, and I expect to return failure exit code if attempting to install package was failed.

Thanks! :)

radoering:
So with Poetry 1.6, `poetry install` fails if the readme is missing, but does not fail if the complete package is missing. Quite inconsistent, isn't it?

[...]

PR #9333, opened 2024-04-24
Fail command install if cannot be installed

eamanu:
An error code is returned if the installation command fails. Also the style is changed from warning to error.

Tests was updated.

# Pull Request Check List

Resolves: #8637 


radoering (changes requested):
LGTM. I'm just blocking this in case we will release a 1.9 before 2.0. (Currently, I do not think we will but who knows.) Considering the churn the warning caused, I think it would be too early to do this in a minor release.

radoering (approved):

Issue #9334, opened 2024-04-24
poetry install crashes due to a race condition and unreported KeyError


vfazio:
### Description

There is a subtle and difficult to reproduce race condition in poetry when using decorated output (default) and parallel installs (default) which leads to poetry randomly failing to install dependencies without displaying the cause of the error.

The gist is this:

Multiple install tasks are queued up:

```python
src/poetry/installation/executor.py
                tasks.append(self._executor.submit(self._execute_operation, operation))

            try:
                wait(tasks)
```

When `self._execute_operation` runs we have branching logic:

```python
            if self.supports_fancy_output():
                if id(operation) not in self._sections and self._should_write_operation(
                    operation
                ):
                    with self._lock:
                        self._sections[id(operation)] = self._io.section()
                        self._sections[id(operation)].write_line(
                            f"  <fg=blue;options=bold>-</> {op_message}:"
                            " <fg=blue>Pending...</>"
                        )
            else:
                if self._should_write_operation(operation):
                    if not operation.skipped:
                        self._io.write_line(
                            f"  <fg=blue;options=bold>-</> {op_message}"
                        )
</snip>
            try:
                result = self._do_execute_operation(operation)                        
```

If fancy output is _not_ supported, a line is written to the output, otherwise a new section is added to our dictionary of sections and then we try to perform the operation.

The operation runs via _do_execute_operation -> _execute_{op}
```
  File "/opt/poetry/venv/lib/python3.12/site-packages/poetry/installation/executor.py", line 303, in _execute_operation
    result = self._do_execute_operation(operation)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/poetry/venv/lib/python3.12/site-packages/poetry/installation/executor.py", line 416, in _do_execute_operation
    result: int = getattr(self, f"_execute_{method}")(operation)
                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/poetry/venv/lib/python3.12/site-packages/poetry/installation/executor.py", line 541, in _execute_install
    status_code = self._install(operation)
                  ^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/poetry/venv/lib/python3.12/site-packages/poetry/installation/executor.py", line 586, in _install
    self._write(operation, message)
```

The operation prints out a message:

```python
        message = (
[...]

PR #9335, opened 2024-04-24
fix race condition and KeyError exception in executor

vfazio:
Previously, it was possible that Executor.supports_fancy_output would flip-flop between True and False if a thread was updating a section. This could lead to crashes when an operation got different reponses as it made progress.

Executor.supports_fancy_output reflects the value of the underlying cleo Formatter used by the Output object. The Formatter is shared by any SectionOutputs derived by that Output object.

If a thread (tA) is in the middle of Formatter.remove_format, the flag to show decorator support is temporarily toggled off and then restored, which opens a window of time where another thread could get an "incorrect" answer when queried via `supports_fancy_output`.

If a parallel thread (tB) queries `supports_fancy_output` and sees it's False, the operation would not get added to the Executor's _sections dictionary. If tB's operation progressed after tA has restored the decorator value and attempts to write out progress information it will call `Executor._write`, see that `supports_fancy_output` is now True and attempt to find the operation in the _sections dictionary, however there will not be an entry for that operation due to the earlier query that returned False.

This causes tB to throw a KeyError and causes the install to shutdown.

Now, the Executor queries and caches whether the Output is decorated during init. This value is used in `supports_fancy_output` so as to not be affected by changes to the underlying Formatter object during section updates.

# Pull Request Check List

Resolves: #9334 

### DIFF
--- a/src/poetry/installation/executor.py
+++ b/src/poetry/installation/executor.py
@@ -104,6 +104,10 @@ class Executor:
         self._shutdown = False
         self._hashes: dict[str, str] = {}
 
+        # Cache whether decorated output is supported.
+        # https://github.com/python-poetry/cleo/issues/423
+        self._decorated_output: bool = self._io.output.is_decorated()
+
     @property
     def installations_count(self) -> int:
         return self._executed["install"]
@@ -121,7 +125,7 @@ class Executor:
         return self._enabled
 
     def supports_fancy_output(self) -> bool:
-        return self._io.output.is_decorated() and not self._dry_run
+        return self._decorated_output and not self._dry_run
 
     def disable(self) -> Executor:
         self._enabled = False


vfazio:
Note that 1.8.x is also affected. I'd like to maybe see this get cherry-picked to that branch as our CI is currently running into this issue.

src/poetry/installation/executor.py
```diff
@@ -104,6 +104,10 @@ def __init__(
         self._shutdown = False
         self._hashes: dict[str, str] = {}
 
+        # Cache whether decorated output is supported.
+        # https://github.com/python-poetry/poetry/issues/9334
```
Secrus, line 108: Could you switch the linked issue to the one in cleo?

vfazio: sure, i can even remove the comment and link to the issue in the commit itself so as to not clutter the code if that's preferred?

Secrus: I would prefer it as a comment in the code. It's more visible that way

vfazio: changed this and rebased on main

Secrus (approved):

src/poetry/installation/executor.py
vfazio: resolving this per approval

PR #9341, opened 2024-04-27
ci: restore old backport workflow because the overhauled workflow does not work

radoering:
It fails with "failed to fork: HTTP 403: Resource not accessible by integration". See https://github.com/python-poetry/poetry/actions/runs/8849739498 for example.

Let's restore the old workflow until we know what to do to fix the new workflow.


.github/workflows/backport.yaml
```diff
@@ -15,26 +15,19 @@ jobs:
     if: >
       github.event.pull_request.merged
       && (
-          (github.event.action == 'closed' && contains(github.event.pull_request.labels.*.name, 'backport/'))
```
radoering, line 18: I reverted this part of the trigger because it resulted in not triggering when the PR was closed.

Secrus (approved):

PR #9413, opened 2024-05-13
Replace --python with --try-first-with

hexmode:
If a tool like asdf provides a python that is first in the path, virtualenv does not use it even though it is the one
used to invoke poetry.  Using --try-first-with (introduced in 20.3.0) allows poetry to point to the prefered python.

See https://virtualenv.pypa.io/en/latest/changelog.html#v20-3-0-2021-01-10
Resolves: #9278

### DIFF
--- a/src/poetry/utils/env/env_manager.py
+++ b/src/poetry/utils/env/env_manager.py
@@ -681,7 +681,7 @@ class EnvManager:
         args = [
             "--no-download",
             "--no-periodic-update",
-            "--python",
+            "--try-first-with",
             executable_str or sys.executable,
         ]
 


radoering:
@hexmode Considering https://github.com/python-poetry/poetry/issues/9278#issuecomment-2131968874, do you think this still makes sense?

Maybe, we should use `--try-first-with` instead of `--python` anyway because we always pass a path and do not need any discovery?

hexmode:
I would have to re-create my scenario, but, yes, using `--try-first-with` makes sense to me.

radoering (approved):

PR #9422, opened 2024-05-17
Downloader support resume from connection reset

austinzh:
# Pull Request Check List

Resolves: #3219
Resolves: #9551

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -273,6 +273,19 @@ values, usage instructions and warnings.
 
 Use parallel execution when using the new (`>=1.1.0`) installer.
 
+### `installer.max-retries`
+
+**Type**: `int`
+
+**Default**: `0`
+
+**Environment Variable**: `POETRY_INSTALLER_MAX_RETRIES`
+
+*Introduced in 1.9.0*
+
+Set the maximum number of retries in an unstable network when using the new (`>=1.9.0`) installer.
+This setting only valid if the server support HTTP range requests.
+
 ### `solver.lazy-wheel`
 
 **Type**: `boolean`
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -273,13 +273,13 @@ values, usage instructions and warnings.
 
 Use parallel execution when using the new (`>=1.1.0`) installer.
 
-### `installer.max-retries`
+### `requests.max-retries`
 
 **Type**: `int`
 
 **Default**: `0`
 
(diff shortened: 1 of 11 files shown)


radoering (changes requested):
`download_file()` is used in some more places (`HTTPRepository`, `DirectOrigin`), which might not be relevant for installing but only for locking. Maybe, connection resets are not that relevant for locking. On the other side, it feels more consistent to also retry in this case. What do you think?

`HTTPRepository` already has the config available, so it is easy to use the setting. For `DirectOrigin`, we had to add another parameter.

austinzh:
> `download_file()` is used in some more places (`HTTPRepository`, `DirectOrigin`), which might not be relevant for installing but only for locking. Maybe, connection resets are not that relevant for locking. On the other side, it feels more consistent to also retry in this case. What do you think?
> 
> `HTTPRepository` already has the config available, so it is easy to use the setting. For `DirectOrigin`, we had to add another parameter.

Sure. I can add parameter to them, It's my first time working in this repo, I agree that even they are not much needed for retry it's better to make it all consistent.

radoering:
> But for the config, should we still use `installer.max-retries`? I think it's still valid name right? Or we add `solver.max-retries` ?

Good question. Thinking about it, `installer.max-retries` might have been a bit ambiguous from the start: We only retry the download not the complete installation. I think we should introduce a new section, maybe call it `network` or `requests`? (We already have a [`POETRY_REQUESTS_TIMEOUT`](https://python-poetry.org/docs/faq/#my-requests-are-timing-out), which is not yet available via config but only via environment variable; anyway, I think that one would fit into the same section.)

austinzh (commented):
Change config to "requests.max-retries"

SmartManoj:
Please reopen https://github.com/python-poetry/poetry/issues/3219 to close the duplicate of #9551

radoering:
> Please reopen #3219 to close the duplicate of #9551

I do not think reopening #3219 makes sense because it is was closed a long time ago. Generally, we do not need an issue if we already have a PR. However, since some users only search issues, I think it is ok to keep #9551 open until this one is merged.

radoering:
> Does this PR need any change?

I think all my comments have been addressed but I still have to review the changes. I won't have time for the next two weeks but feel free to remind me end of August if I have not come back to this issue by then.

radoering (approved):

Issue #9432, opened 2024-05-21
poetry build -C ./relative/path doesnt work properly


OhioDschungel6:
### Description

When your projekt looks like the following:
.
├── poetry.lock
├── pyproject.toml
└── src
    └── testproject
        └── __init__.p
the command poetry build -C "." doesn't work properly.
In general relative paths are not resolved properly in this case (you have to specify "packages = [{include = "testproject", from="src"}]" in your pyproject.toml).

### Workarounds

Use absolute paths or use "poetry build" without -C

### Poetry Installation Method

other

### Operating System

Windows 11

### Poetry Version

1.8.3

### Poetry Configuration

```bash session
Clean install. No further configuration.
```

### Example pyproject.toml

```TOML
[tool.poetry]
name = "testproject"
version = "0.1.0"
description = ""
authors = ["Your Name <you@example.com>"]
packages = [{include = "testproject", from="src"}]

[tool.poetry.dependencies]
python = "^3.11"


[build-system]
requires = ["poetry-core"]
build-backend = "poetry.core.masonry.api"
```

### Poetry Runtime Logs

```bash session
Command:  poetry -vvv build --directory "."
Building testproject (0.1.0)
  - Building sdist
  - Adding: src\testproject\__init__.py
  - Adding: pyproject.toml
  - Built testproject-0.1.0.tar.gz
  - Building wheel
  - Adding: src\testproject\__init__.py

  Stack trace:

  12  ~\scoop\apps\poetry\1.8.3\venv\Lib\site-packages\cleo\application.py:327 in run
       325│
       326│             try:
     → 327│                 exit_code = self._run(io)
       328│             except BrokenPipeError:
       329│                 # If we are piped to another process, it may close early and send a

  11  ~\scoop\apps\poetry\1.8.3\venv\Lib\site-packages\poetry\console\application.py:190 in _run
       188│         self._load_plugins(io)
       189│
     → 190│         exit_code: int = super()._run(io)
       191│         return exit_code
       192│

  10  ~\scoop\apps\poetry\1.8.3\venv\Lib\site-packages\cleo\application.py:431 in _run
       429│             io.input.interactive(interactive)
       430│
     → 431│         exit_code = self._run_command(command, io)
       432│         self._running_command = None
       433│

   9  ~\scoop\apps\poetry\1.8.3\venv\Lib\site-packages\cleo\application.py:473 in _run_command
       471│
       472│         if error is not None:
     → 473│             raise error
       474│
[...]

PR #9433, opened 2024-05-21
Fix relative build path issue

OhioDschungel6:
# Pull Request Check List

Resolves: #9432


OhioDschungel6:
Hey, this is my first time contributing.
I have not changed the docs, because the fixed behavior is intended and was already documented.
I was not able to write a reproducing test in "test_build.py" as this seems to circumvent the "application.py" module. 
Also a nested src folder is needed as shown in the issue.
The simplest way to reproduce the issue is to check out the poetry repo and try: "poetry build -C ."

radoering:
I added a test, which was a bit tricky for sure. You have to choose `ApplicationTester` instead of `CommandTester` (in comparison to the other tests in `test_build.py`) because the latter initializes Poetry before the directory is passed. But even when using the correct tester class, at first, I was not able to reproduce the issue because of #9444.

OhioDschungel6:
@radoering nice! thanks for the help

Issue #1488, opened 2019-10-20
keep old line breaker


Trim21:
- [X] I have searched the [issues](https://github.com/sdispater/poetry/issues) of this repo and believe that this is not a duplicate.
- [X] I have searched the [documentation](https://poetry.eustace.io/docs/) and believe that my question is not covered.

## Feature Request

I'm using poetry on windows or Linux.

So I may lock deps both on windows and Linux. But the line endings of `poetry.lock` and `pyproject.toml` are different on Linux and windows. 

I'm considering if poetry could check original line endings and keep it when generating a new lock file.

Trim21:
Or make it `LF` in any system to make sure it's same on all system.

tigerhawkvok:
A config option to always use one or the other would work, too. I always want LF endings, never CRLF, even on Windows.

Trim21:
Looks like this feature has already been added. so closing it.

Issue #7527, opened 2023-02-17
Preserve line breaks in poetry.lock (regression)


extrwi:
- **Poetry version**: 1.3.2
- **Python version**: 3.11.1
- **OS version and name**: Windows 10

- [x] I am on the [latest](https://github.com/python-poetry/poetry/releases/latest) stable Poetry version, installed using a recommended method.
- [x] I have searched the [issues](https://github.com/python-poetry/poetry/issues) of this repo and believe that this is not a duplicate.
- [ ] I have consulted the [FAQ](https://python-poetry.org/docs/faq/) and [blog](https://python-poetry.org/blog/) for any relevant entries or release notes.
- [ ] If an exception occurs when executing a command, I executed it again in debug mode (`-vvv` option) and have included the output below.

## Issue

When generating poetry.lock in mixed environments (e.g. Windows and Linux), CRLF gets replaced by LF and vice versa. This had already been reported and supposedly fixed before (https://github.com/python-poetry/poetry/issues/1488), but probably regressed during the introduction of tomli (https://github.com/python-poetry/poetry/pull/6562). I couldn't find any special treatment for line endings of poetry.lock in poetry's code.

dimbleby:
there has been no recent change here, poetry continues to use tomlkit to write its lockfiles.

The tomlkit API offers no control so far as I can see over this, so if there's something important to you here you'll need to raise it at that project.

https://github.com/sdispater/tomlkit/pull/200 and https://github.com/sdispater/tomlkit/pull/201 look like the last time anyone worried about this

dimbleby:
Maybe this has changed actually, perhaps reading the lockfile via tomlkit allowed preservation of the existing line endings where reading it through tomli discards that information.

IMO this would be a reason to move further away from tomlkit and write the lockfile with tomli-w (which uses the same line ending on all platforms).

radoering:
> IMO this would be a reason to move further away from tomlkit and write the lockfile with tomli-w (which uses the same line ending on all platforms).

I don't think that's reason enough. We discussed that we don't want to introduce two libraries for writing toml files if there is not a really good reason. And we are tied to tomlkit for modifying pyproject.toml via `poetry add`. IMO the more promising approach is to extend the tomlkit API to force specific line endings so we can keep the existing line endings (if they are consistent) even if we use tomllib for reading.

> When generating poetry.lock in mixed environments (e.g. Windows and Linux), CRLF gets replaced by LF and vice versa.

[...]

PR #9468, opened 2024-06-03
fix: keep lock file eol

trim21:
# Pull Request Check List

Resolves: https://github.com/python-poetry/poetry/issues/1488 again

https://github.com/python-poetry/poetry/issues/1488 once be fixed but it's broken again.

appearly `TOMLFile` need to call `read` method to get line encoding, otherwise it just use default os eol.

### DIFF
--- a/src/poetry/packages/locker.py
+++ b/src/poetry/packages/locker.py
@@ -291,6 +291,8 @@ class Locker:
 
     def _write_lock_data(self, data: TOMLDocument) -> None:
         lockfile = TOMLFile(self.lock)
+        if self.lock.exists():
+            lockfile.read() # get original line ending
         lockfile.write(data)
 
         self._lock_data = None


Secrus:
Could you please make the fix in [`tomlkit`](https://github.com/python-poetry/tomlkit) instead? Sounds like it belongs there more than Poetry code.

trim21:
> Could you please make the fix in [`tomlkit`](https://github.com/python-poetry/tomlkit) instead? Sounds like it belongs there more than Poetry code.

tomlkit already fixed this, many times, I think.

https://github.com/python-poetry/tomlkit/pulls?q=is%3Apr+line+ending+is%3Aclosed

 they even have test for this

Secrus:
Well, if it was fixed in tomlkit, we wouldn't need this patch, right? For me, this sounds like a workaround at best.

trim21:
> Well, if it was fixed in tomlkit, we wouldn't need this patch, right? For me, this sounds like a workaround at best.

Actually, I'm not sure if this is a bug of tomlkit, or poetry didn't use it as expected.

trim21:
but you are right, this can be fiexd on both side. tomlkit could get line ending then `TomlFile` object is created.

tests/packages/test_locker.py
```diff
@@ -1231,3 +1231,20 @@ def test_lockfile_is_not_rewritten_if_only_poetry_version_changed(
         content = f.read()
 
     assert content == old_content
+
+
+def test_lockfile_keep_eol(locker: Locker, root: ProjectPackage) -> None:
+    sep = "\n" if os.sep == "\r\n" else "\r\n"
```
radoering, line 1237:
```suggestion
    sep = "\n" if os.linesep == "\r\n" else "\r\n"
```
?

radoering (changes requested):
I suppose this has a negative impact on performance but I hope this is neglectable.

Secrus:
I still believe that this should be fixed in `tomlkit` not here

trim21:
> > I still believe that this should be fixed in `tomlkit` not here
> 
> I do not think that is possible because since #6562 we use `tomli` to read the lock file so that `tomlkit` cannot know what type of line endings are in the lock file. If we want to keep existing line endings we either have to read with `tomlkit` (what this PR is doing) or determine the type of line ending by ourself and somehow set it in `tomlkit`. I think I would prefer reading with `tomlkit` (as it is done in this PR) if it was not a performance issue.
> 
> #6562 suggests that there is a relevant performance impact:
> 
> > Even for a modestly sized project like poetry itself, reading poetry.lock on my laptop takes 0.6 - 0.7 seconds (timings obtained with some time.time() statements inserted into the codebase). tomli manages it in ~0.02 seconds.
> 
> Thus, maybe we should really check the timings and investigate the alternative approach.

this only impact perf when it write lockfile, which mean only `add` and `lock`. other command like `install`, `update` won't be affected.

radoering:
> this only impact perf when it write lockfile, which mean only add and lock. other command like install, update won't be affected. normally add/lock will spend more time on resolving.

Good point. I think I will do some simple performance measurements later.

We could basically implement part of the logic of https://github.com/python-poetry/tomlkit/pull/201/files by ourselves, just using `TOMLDocument.as_string()` instead of `TOMLFile.write()`. However, that might not be worth it depending on what the timing measurements will reveal.

src/poetry/packages/locker.py
```diff
@@ -291,6 +293,10 @@ def _should_write(self, lock: TOMLDocument) -> bool:
 
     def _write_lock_data(self, data: TOMLDocument) -> None:
         lockfile = TOMLFile(self.lock)
+        if self.lock.exists():
+            # get original line ending.
+            with contextlib.suppress(TOMLError):
+                lockfile.read()
```
radoering, line 299:
```suggestion
        if self.lock.exists():
            # The following code is roughly equivalent to
            # • lockfile = TOMLFile(self.lock)
            # • lockfile.read()
            # • lockfile.write(data)
            # However, lockfile.read() takes more than half a second even
            # for a modestly sized project like Poetry itself and the only reason
            # for reading the lockfile is to determine the line endings. Thus,
            # we do that part for ourselves here, which only takes about 10 ms.

            # get original line endings
            with open(self.lock, encoding="utf-8", newline="") as f:
                line = f.readline()
            linesep = "\r\n" if line.endswith("\r\n") else "\n"

            # enforce original line endings
            content = data.as_string()
            if linesep == "\n":
                content = content.replace("\r\n", "\n")
            elif linesep == "\r\n":
                content = re.sub(r"(?<!\r)\n", "\r\n", content)
            with open(self.lock, "w", encoding="utf-8", newline="") as f:
                f.write(content)

        else:
            lockfile = TOMLFile(self.lock)
            lockfile.write(data)
```

radoering (changes requested):
I measured the timing for Poetry itself, and the penalty for reading the lock file with `tomlkit` is still about 0.7 s. Even though that is only done when the lock file is actually written, it still feels like a significant regression for an edge use case (i.e. keep line endings). IMO, we have two options:

- Just do not support this use case and write `\r\n` on Windows and `\n` on Linux. You can still enforce one type of line endings in your git repo via .gitattributes (cf https://github.com/python-poetry/poetry/issues/7527#issuecomment-1494170636)
- Implement the "keep eol" logic from `tomlkit` by ourselves. See suggested change.

src/poetry/packages/locker.py
dimbleby: or, abandon the notion that a particular formatting-style for lockfiles is important and write them with tomli-w - which always uses `\n` exclusively and therefore can never flip-flop.

I realise that I've suggested this in the past and I guess we likely continue to disagree on the merits of this - but I continue to favour eliminating tomlkit wherever it is possible to do so.

dimbleby:
> IMO, we have two options:

or, abandon the notion that a particular formatting-style for lockfiles is important and write them with tomli-w - which always uses `\n` exclusively and therefore can never flip-flop.

I realise that I've suggested this in the past and I guess we likely still disagree on the merits of this - but I continue to favour eliminating tomlkit wherever it is possible to do so.

dimbleby:
to make the tomli-w comparison concrete I opened https://github.com/python-poetry/poetry/pull/9637

trim21:
im OK with forcing lf, I just want eol don't change from lf to crlf or crlf to lf.

radoering (approved):

PR #9475, opened 2024-06-06
fix(check): don't raise error on pypi reference

Secrus:
# Pull Request Check List

Resolves: #9436

### DIFF
--- a/src/poetry/console/commands/check.py
+++ b/src/poetry/console/commands/check.py
@@ -94,7 +94,7 @@ class CheckCommand(Command):
 
     def _validate_dependencies_source(self, config: dict[str, Any]) -> list[str]:
         """Check dependencies's source are valid"""
-        sources = {k["name"] for k in config.get("source", [])}
+        sources = {repository.name for repository in self.poetry.pool.all_repositories}
 
         dependency_declarations: list[
             dict[str, str | dict[str, str] | list[dict[str, str]]]
--- a/tests/console/commands/test_check.py
+++ b/tests/console/commands/test_check.py
@@ -137,33 +137,13 @@ def test_check_invalid_dep_name_same_as_project_name(
         ),
         new_callable=mocker.PropertyMock,
     )
-    tester.execute("--lock")
-    fastjsonschema_error = "data must contain ['description'] properties"
-    custom_error = "The fields ['description'] are required in package mode."
-    expected_template = """\
-Error: Project name (invalid) is same as one of its dependencies
-Error: Unrecognized classifiers: ['Intended Audience :: Clowns'].
-Error: Declared README file does not exist: never/exists.md
-Error: Invalid source "exists" referenced in dependencies.
-Error: Invalid source "not-exists" referenced in dependencies.
-Error: Invalid source "not-exists2" referenced in dependencies.
-Error: poetry.lock was not found.
-Warning: A wildcard Python dependency is ambiguous.\
- Consider specifying a more explicit one.
-Warning: The "pendulum" dependency specifies the "allows-prereleases" property,\
- which is deprecated. Use "allow-prereleases" instead.
-Warning: Deprecated classifier 'Natural Language :: Ukranian'.\
- Must be replaced by ['Natural Language :: Ukrainian'].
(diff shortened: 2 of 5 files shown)


src/poetry/console/commands/check.py
```diff
@@ -95,6 +95,7 @@ def _validate_readme(self, readme: str | list[str], poetry_file: Path) -> list[s
     def _validate_dependencies_source(self, config: dict[str, Any]) -> list[str]:
         """Check dependencies's source are valid"""
         sources = {k["name"] for k in config.get("source", [])}
+        sources.add("pypi")
```
radoering, line 98: PyPI might be deactivated. I think we should use `self.poetry.pool.all_repositories` to get the names of the sources.

radoering (changes requested):

cjalvis (approved):

tests/console/commands/test_check.py
```diff
@@ -254,3 +306,25 @@ def test_check_lock_up_to_date(
...
+
+
+def test_check_does_not_error_on_pypi_reference(
+    command_tester_factory: CommandTesterFactory,
+    poetry_with_pypi_reference: Poetry,
+) -> None:
+    tester = command_tester_factory("check", poetry=poetry_with_pypi_reference)
+    status_code = tester.execute("")
+
+    # Warnings about deprecated syntax are expected
```
radoering, line 318: Shouldn't we just avoid deprecated syntax in this test because it distracts?

tests/console/commands/test_check.py
```diff
@@ -111,19 +127,55 @@ def test_check_valid_legacy(
     assert tester.io.fetch_error() == expected
 
 
-def test_check_invalid(
+def test_check_invalid_dep_name_same_as_project_name(
```
radoering, line 130: What is the reason for splitting (almost duplicating) the test and how is it related to the fix?

Secrus: The old one was checking the dependency name vs project name. This is not possible with a fixture being a Poetry instance, so I moved this to another test to keep this case tested.

tests/console/commands/test_check.py
Secrus: Yeah, right.

tests/console/commands/test_check.py
radoering: So if we do not introduce the fixture, we do not have to split the test? Then, why introduce a fixture in the first place? (Further, it still seems unrelated to the test case for the fix.)

tests/fixtures/pypi_reference/poetry.lock
```diff
@@ -0,0 +1,232 @@
+# This file is automatically @generated by Poetry 1.8.3 and should not be changed by hand.
```
radoering, line 1: I think we do not need the lockfile for the test.

tests/console/commands/test_check.py
Secrus: We need to introduce the fixture because otherwise (with patching pyproject file) we get the wrong list of sources for the project.

radoering: I still do not get why this is an issue. We now have two almost similar tests (`test_check_invalid_dep_name_same_as_project_name `and `test_check_invalid`). Both provide the same errors except for that the first one provides two more errors:

- `Error: Project name (invalid) is same as one of its dependencies`
- `Error: Invalid source "exists" referenced in dependencies.`

And both tests are unrelated to the issue that is resolved by this PR. This is tested in `test_check_does_not_error_on_pypi_reference`, which uses another fixture.

radoering: After trying it out myself, I now understand why you had to introduce the fixture: The change requires that sources are parsed, which is not done without the fixture. Otherwise, all sources (also existing ones) will be recognized as non-existent.

I added a commit to reduce the duplication between the two tests to a minimum. Please take a look if this is fine for you.

Secrus: Looks fine, thanks for the commit.

radoering (approved):

PR #9478, opened 2024-06-07
update dependencies

dimbleby:
updating dependencies.  

minor annoyances:
- unwanted setuptools dependency via dulwich, will be resolved in their next release per https://github.com/jelmer/dulwich/pull/1329
- missing type information in pkginfo https://bugs.launchpad.net/pkginfo/+bug/2068777

I've switched dev dependencies to use `>=`, not sure whether this is going to be controversial, can undo it if so.  My reasoning is that it makes no difference to consumers of poetry, and no difference either to developers since the policy is to commit the lock file - except that it becomes just fractionally easier to take updates, and minimizes potential conflicts between dev and non-dev requirements


radoering (approved):
I agree with omitting upper bounds for dev dependencies.

I just added `setuptools` for older Python versions in the `dev` group because PyCharm still does not play well with missing `setuptools` for Python 3.9 and older.

PR #9491, opened 2024-06-15
Tests: add tests for vcs.git.backend

indrajithi:
# Pull Request Check List

Related-to: [#3155](https://github.com/python-poetry/poetry/issues/3155)

- [X] Added **tests** for `poetry.vcs.git.backend`

### DIFF
--- a/tests/vcs/git/test_backend.py
+++ b/tests/vcs/git/test_backend.py
@@ -1,9 +1,15 @@
 from __future__ import annotations
 
+from unittest.mock import MagicMock
+
 import pytest
 
+from dulwich.repo import Repo
+
 from poetry.vcs.git.backend import Git
+from poetry.vcs.git.backend import annotated_tag
 from poetry.vcs.git.backend import is_revision_sha
+from poetry.vcs.git.backend import urlpathjoin
 
 
 VALID_SHA = "c5c7624ef64f34d9f50c3b7e8118f7f652fddbbd"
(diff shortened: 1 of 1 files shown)


tests/vcs/git/test_backend.py
```diff
@@ -41,3 +47,25 @@ def test_invalid_revision_sha_max_len() -> None:
 def test_get_name_from_source_url(url: str) -> None:
     name = Git.get_name_from_source_url(url)
     assert name == "poetry"
+
+
+def test_annotated_tag() -> None:
```
radoering, line 52: Can you make the test `parametrized` to test passing `bytes` as well as `str`, please?

tests/vcs/git/test_backend.py
```diff
@@ -41,3 +47,25 @@ def test_invalid_revision_sha_max_len() -> None:
...
+def test_get_remote_url() -> None:
+    repo = MagicMock(spec=Repo)
+    repo.get_config.return_value.get.return_value = (
+        b"https://github.com/python-poetry/poetry.git"
+    )
+
+    assert Git.get_remote_url(repo) == "https://github.com/python-poetry/poetry.git"
+
+
+def test_urlpathjoin() -> None:
```
radoering, line 66: Can you make the test `parametrized` to add an example with a trailing slash, e.g. `"ssh://git@github.com/org/repo/"`, please?

radoering (changes requested):

tests/vcs/git/test_backend.py
indrajithi: CMMIR, 
```
    @staticmethod
    def get_name_from_source_url(url: str) -> str:
        return re.sub(r"(.git)?$", "", url.rstrip("/").rsplit("/", 1)[-1])
```
 Does not accept bytes as url right? So passing bytes as url will fail unless we change this implementation.

tests/vcs/git/test_backend.py
indrajithi: I have made the changes. LMK if I missed something or need to make any further changes. @radoering

radoering (approved):

PR #9523, opened 2024-07-01
python 3.13 in workflows

dimbleby:
time to find out what is broken at python 3.13


dimbleby:
could be worse:

- mysterious failure to install on macos
  - I guess something is unavailable as a wheel and fails to build?
  - but the output is not enlightening and I do not care to dig into it
  - with any luck this will solve itself automatically, whenever the problematic dependency is sorts itself out... fingers crossed
- and some encoding warnings, presumably per https://github.com/python-poetry/poetry/pull/8893

radoering:
> and some encoding warnings

Can you tell me where I see these warnings? I thought it should be visible for example in the Ubuntu Python 3.13 pytest job when I expand `Run poetry run pytest --integration -v` and scroll down but there is only a `datetime` warning. (Still wondering why I cannot find any warnings in #8893, cf https://github.com/python-poetry/poetry/pull/8893#issuecomment-2132228330)

dimbleby:
I might have mixed up poetry and poetry-core, `PYTHONWARNDEFAULTENCODING` is only turned on in poetry-core

eg in the poetry-core pull request https://github.com/python-poetry/poetry-core/actions/runs/9750266660/job/26909309129?pr=742

> Run poetry run python -m pytest -p no:sugar --integration -q tests/integration
/home/runner/.local/share/pypoetry/venv/lib/python3.13/site-packages/poetry/utils/env/base_env.py:338: EncodingWarning: 'encoding' argument not specified.
  output = subprocess.check_output(

but this is also true in the 3.12 test so perhaps nothing much to do with 3.13

dimbleby:
re-ran following the release of cffi 1.17.0, pretty much working now.

Just one over-specified testcase that mysteriously is ever so slightly different on macos python 3.13.

Aditya-1313 (approved):

radoering:
> Just one over-specified testcase that mysteriously is ever so slightly different on macos python 3.13.

I assume we should just make the test more robust to allow any order in the output because `cleo`, `tomlkit` and `pendulum` are all installed in parallel.

dimbleby:
or, maybe, make the tests run with a worker pool of size one so that they are not exposed to this sort of thing

(I did take a brief look back in August and remember being puzzled as to what was going on, not expecting to try again though)

radoering (approved):

PR #9534, opened 2024-07-07
enable encoding warnings

dimbleby:
verifying that this behaves as expected in github workflows


Secrus:
@dimbleby considering there is already #8893 open, do we need this PR or can we include these changes in the other one?

PR #9535, opened 2024-07-07
fix: specify venv activator as init script when starting PowerShell

lewis-yeung:
This PR is to fix a bug (not on Windows but at least on Linux): `poetry shell` sends the command line that invokes the venv activator script into a newly spawned PowerShell instance, while the command will not be automatically executed.


lewis-yeung:
Any comments?

Secrus:
Further development of the `shell` command was moved to [`poetry-plugin-shell`](https://github.com/python-poetry/poetry-plugin-shell), please open PR in that repo.

PR #9542, opened 2024-07-10
Docs/docker best practices

gianfa:
# Pull Request Check List

Resolves: Discussion #1879 

This PR is intended to begin the section on Docker Best Practices in Documentation.
It will serve as a collection point for all best practices (BPs) that will be recommended.
Hopefully, BPs, in order to be included and accepted here, will have to be accompanied by a concrete example that reproduces and demonstrates their effectiveness and by due references where possible.

The following contributions are proposed in this PR:

- `docs/docker-best-practices.md`. Introduction file and list of best practices.
- `docker-examples/`. Folder with practical examples to get you started. This already comes with two examples, also referenced in the docker-best-practices.md.

### DIFF
--- a/docker-examples/minimum-poetry/Dockerfile
+++ b/docker-examples/minimum-poetry/Dockerfile
@@ -1,0 +1,18 @@
+FROM python:3.11-slim
+
+ARG POETRY_VERSION=1.8
+
+# See https://python-poetry.org/docs/#ci-recommendations
+RUN pip install "poetry==${POETRY_VERSION}"
+
+WORKDIR /app
+
+# --- Reproduce the environment ---
+# You can comment the following two lines if you prefer to manually install
+#   the dependencies from inside the container.
+COPY pyproject.toml .
+
+RUN poetry install
+# --- --- --- --- --- --- --- --- ---
+
+ENTRYPOINT ["bash"]
--- a/docker-examples/poetry-multistage/Dockerfile
+++ b/docker-examples/poetry-multistage/Dockerfile
@@ -1,0 +1,38 @@
+FROM python:3.11-slim as builder
+
+# --- Install Poetry ---
+ARG POETRY_VERSION=1.8
+
+ENV POETRY_HOME=/opt/poetry
+ENV POETRY_NO_INTERACTION=1
+ENV POETRY_VIRTUALENVS_IN_PROJECT=1
+ENV POETRY_VIRTUALENVS_CREATE=1
+ENV PYTHONDONTWRITEBYTECODE=1
+ENV PYTHONUNBUFFERED=1
+# Tell Poetry where to place its cache and virtual environment
+ENV POETRY_CACHE_DIR=/opt/.cache
+
+RUN pip install "poetry==${POETRY_VERSION}"
+
+WORKDIR /app
+
+# --- Reproduce the environment ---
+# You can comment the following two lines if you prefer to manually install
+#   the dependencies from inside the container.
+COPY pyproject.toml .
+
+# Install the dependencies and clear the cache afterwards.
+#   This may save some MBs.
+RUN poetry install --no-root && rm -rf $POETRY_CACHE_DIR
+
+# Now let's build the runtime image from the builder.
+#   We'll just copy the env and the PATH reference.
+FROM python:3.11-slim as runtime
+
+ENV VIRTUAL_ENV=/app/.venv
+ENV PATH="/app/.venv/bin:$PATH"
+
+COPY --from=builder ${VIRTUAL_ENV} ${VIRTUAL_ENV}
+
+
+ENTRYPOINT ["bash"]
--- a/docker-examples/minimum-poetry/.gitignore
+++ b/docker-examples/minimum-poetry/.gitignore
@@ -1,0 +1,1 @@
+poetry.lock
--- a/docker-examples/minimum-poetry/README.md
+++ b/docker-examples/minimum-poetry/README.md
@@ -1,0 +1,38 @@
+# Minimum Poetry image
+
+## Description
+
+Minimum-poetry is the minimum-constructible image containing poetry, from an official python base image.
+
+Expected size: ~218 MB, virtual env layer excluded.
+
+## Use cases
+
+This image is especially useful when you don't yet have in mind a clear idea of the environment requirements you need, but need a reproducible first starting point for a python development environment. It' a quick and easy way to start.
+
+## How to use it
+
+Run the following commmands from the *minimum-poetry* folder. They are just an example of how to use it. You can wrrite your custom commands according to Docker API. For more information about Docker please see the [official documentation](https://docs.docker.com/).
+
+### Build the image
+
+```bash
+# This will build the image
+
+TAG="minimum-poetry:0.1.0"
+docker build \
+    -t $TAG \
+    --build-arg POETRY_VERSION="1.8.3" \
+    "."
+```
+
+### Run the container
+
+```bash
+# This will run the container
+
+docker run \
+    --rm -it \
+    -v ${PWD}:/app/shared \
+    "$TAG"
+```
--- a/docker-examples/minimum-poetry/pyproject.toml
+++ b/docker-examples/minimum-poetry/pyproject.toml
@@ -1,0 +1,18 @@
+[tool.poetry]
+name = "minimum-poetry"
+version = "0.1.0"
+description = "A minimum functioning example of Docker image based on Poetry"
+authors = ["gianfa <gian.angelini@hotmail.com>"]
+readme = "README.md"
+
+[tool.poetry.dependencies]
+python = "^3.9"
+
+numpy = "^2.0.0"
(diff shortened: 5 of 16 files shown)


docker-examples/minimum-poetry/Dockerfile
```diff
@@ -0,0 +1,18 @@
...
+
+# See https://python-poetry.org/docs/#ci-recommendations
+RUN pip install "poetry==${POETRY_VERSION}"
+
+WORKDIR /app
+
+# --- Reproduce the environment ---
+# You can comment the following two lines if you prefer to manually install
+#   the dependencies from inside the container.
+COPY pyproject.toml .
```
radoering, line 13: Shouldn't we copy also copy `poetry.lock` for reproducibility?

docker-examples/minimum-poetry/Dockerfile
```diff
@@ -0,0 +1,18 @@
...
+
+RUN poetry install
```
radoering, line 15:
```suggestion
RUN poetry install --no-root
```
Otherwise it might fail.

docker-examples/poetry-multistage/Dockerfile
```diff
@@ -0,0 +1,38 @@
...
+ENV POETRY_CACHE_DIR=/opt/.cache
+
+RUN pip install "poetry==${POETRY_VERSION}"
+
+WORKDIR /app
+
+# --- Reproduce the environment ---
+# You can comment the following two lines if you prefer to manually install
+#   the dependencies from inside the container.
+COPY pyproject.toml .
```
radoering, line 22: Shouldn't we copy also copy `poetry.lock` for reproducibility?

radoering (commented):
Some first feedback.

One general point: I am wondering why the project itself is not copied (after `poetry install`)? It feels like the examples stop halfway through.

docker-examples/minimum-poetry/Dockerfile
gianfa: This is an important point.
Short answer. Yes, let's add it.

As much as I understand the function of the two (pyproject.toml, poetry.lock), in collaborative development, one can rely on even just the pyproject.toml to keep a general idea of the project's dependencies and configurations, in order to let the developers add what they need in terms of libraries and versions; moreover we can always get the latest versions at build time.
I didn't put the poetry.lock because this Dockerfile was for
> when you don't yet have in mind a clear idea of the environment requirements you need

In fact, on second thought, in this case, if we show a “minimal” example and want to take for granted the function of exactly reproducing an environment ,it becomes necessary. Thank you for pointing this out to me

docker-examples/minimum-poetry/Dockerfile
gianfa: It actually worked for me without it, but I added it anyway as  you suggested, as it doesn't go against the main function. Thanks

docker-examples/poetry-multistage/Dockerfile
gianfa: Done, like for https://github.com/python-poetry/poetry/pull/9542#discussion_r1797755173. Thanx

radoering:
> I wonder if we could make a repo in our org to store those examples instead of putting them here. @radoering what do you think?

We could. Do you propose to only put the `docker-examples` directory into that repo and keep the docs here? This could be annoying because worst case you have to update two repos to update something. The other option would be to convert `docs/docker-best-practices.md` to the readme of the new repo and just put a link to the repo in the docs of poetry.

abn:
There are some things that irk me in this. But first, is this something we are considering putting into the docs?

I agree with Secrus that the example code etc should either be inline in doc or in a separate repo. I personally think we should not need to be the care takers of that, as that is out of scope for Poetry as a project.

docker-examples/poetry-multistage/Dockerfile
```diff
@@ -0,0 +1,37 @@
+FROM python:3.11-slim as builder
+
+# --- Install Poetry ---
+ARG POETRY_VERSION=1.8
+
+ENV POETRY_HOME=/opt/poetry
+ENV POETRY_NO_INTERACTION=1
```
fechnert, line 7: Is this ENV variable available? It's not documented and in [this discussion](https://github.com/orgs/python-poetry/discussions/8977) it was mentioned that it's not a real thing.

gianfa: Just removed. Thank you for pointing out!

docker-examples/poetry-multistage/Dockerfile
```diff
@@ -0,0 +1,36 @@
...
+WORKDIR /app
+
+# --- Reproduce the environment ---
+# You can comment the following two lines if you prefer to manually install
+#   the dependencies from inside the container.
+COPY pyproject.toml poetry.lock /app/
+
+# Install the dependencies and clear the cache afterwards.
+#   This may save some MBs.
+RUN poetry install --no-root && rm -rf $POETRY_CACHE_DIR
```
iTrooz, line 25: I wonder if `rm -rf $POETRY_CACHE_DIR` is necessary since this is the builder stage, that shouldn't (?) be present in the output image at all

docker-examples/poetry-multistage/Dockerfile
```diff
@@ -0,0 +1,36 @@
+FROM python:3.11-slim as builder
```
iTrooz, line 1: Consider using `python:3.12-alpine` instead (124MB vs 48.3MB)

gianfaProm: Alpine is a good choice to make _the lightest poetry image_. Thank you.
However, I would not recommend it regularly, since it does not seem to be the top choice in terms of security and stability (e.g.  https://pythonspeed.com/articles/alpine-docker-python/).

Please feel free to share more information if you feel differently

gianfa: Thank you very much for sharing this suggestion. In fact Alpine may be great for a “the lightest poetry image” version.
In general, I would not recommend Alpine, since it does not seem the best in terms of security and stability (see e.g. https://pythonspeed.com/articles/alpine-docker-python/).

Please feel free to share if you have different thoughts.

iTrooz: That's interesting. No, I don't have any other argument other than "I've heard alpine small", and the image size comparison I shared earlier.

I'm not sure why you mention security though ? I didn't see anything like that in your article

gianfa: You are right.
My answer comes from some rules of thumb like:

> prefer *debian* for stability, prefer *slim* for smallness, use *alpine* when you want something just essential, but consider that you will have to compile everything yourself, with possible errors.

Anyway, it was enough to search on docker hub, to compare for example on python 3.11:
| image | Main CVEs | compressed size|
| - | - | - |
|[python:3.11-slim-bookworm](https://hub.docker.com/layers/library/python/3.11-slim-bookworm/images/sha256-ea2dcd43dcf76b35fec9976bb89e9e6e4d688ec95710b5eb3a9d65e6123f5b11) |1 H| ~46MB |
|[python:3.11-alpine3.21](https://hub.docker.com/layers/library/python/3.11-alpine3.21/images/sha256-41713a75be5fddc59fec56ffa9d94cb803a4d2a1ca124593801865a6e1ca11b3) |1 H| ~20MB|

So I would say that from the security point of view there are no particular concerns. Finally, alpine team also state that [they care about safety](https://alpinelinux.org/about/#:~:text=who%20appreciate%20security) :) .
Thank you again

docker-examples/poetry-multistage/Dockerfile
Manezki: Having the `rm -rf $POETRY_CACHE_DIR` should keep the intermediate stage smaller.
With bigger dependencies (pytorch) the cache impact is notable.

docker-examples/poetry-multistage/Dockerfile
```diff
@@ -0,0 +1,36 @@
...
+
+# --- Install Poetry ---
+ARG POETRY_VERSION=1.8
+
+ENV POETRY_HOME=/opt/poetry
+ENV POETRY_VIRTUALENVS_IN_PROJECT=1
+ENV POETRY_VIRTUALENVS_CREATE=1
+ENV PYTHONDONTWRITEBYTECODE=1
+ENV PYTHONUNBUFFERED=1
+# Tell Poetry where to place its cache and virtual environment
```
Manezki, line 11: Noticed that this seems wrong, since `POETRY_VIRTUALENVS_IN_PROJECT=1` is set
```suggestion
# Tell Poetry where to place its cache
```

docker-examples/poetry-multistage/Dockerfile
dolfinus: Why not using:
```
RUN --mount=type=cache,target=/opt/.cache \
    poetry install
```
This allows saving poetry package cache in local docker builder, which can be reused later. And it is not included into final image.

Secrus:
Stalled

Issue #3316, opened 2020-11-02
Method for enforcing a minimum version of Poetry


PetterS:
Poetry 1.1 is not compatible with 1.0, so we need to enforce Poetry version for our development team one way or another.

There is this Stack Overflow question: https://stackoverflow.com/questions/64003868/how-to-enforce-a-minimum-python-poetry-version which currently is unanswered.

Is there a way of specifying a version of Poetry in pyproject.toml that will fail if the executing version of poetry does not match? What effect does `[build-system] requires` have?

abn:
At present, `poetry` does not enforce a version required as in theory it is not necessary per say at the moment as it is the lock file that is not forwards compatible.

```console
root@cc2fc2786035:/foobar# cat poetry.lock | grep lock-version
lock-version = "1.1"
root@cc2fc2786035:/foobar# pip install -q --force poetry==1.0.10
root@cc2fc2786035:/foobar# poetry install
The lock file might not be compatible with the current version of Poetry.
Upgrade Poetry to ensure the lock file is read properly or, alternatively, regenerate the lock file with the `poetry lock` command.
...
```

That said, the lock version is the closest thing we have at the moment. As you can see above, lock files generated since `1.0.10` if add and verify `lock-version` metadata in the `poetry.lock` file. Version `1.1.0` onwards the `lock-version` is `1.1` (note that this is not tied to `poetry` version, but to the lock file format).

Lock file generated using `poetry>=1.1.0` will raise a warning if used with `poetry==1.0.10`. For now, in cases where poetry version used to generate the lock file needs to be enforced, you could possible verify this by checking if `poetry lock --no-update` generates a diff.

As for `build-system.requires`, this does not have any bearing on how `poetry` operates as it is relevant only to PEP 517 scenarios. This informaiton is used when a PEP 517 front-end builds the project and needs to know what dependencies to install when doing so. For example, this is used when you do `pip install /path/to/poetry/project`. And this is now expecte to use `poetry-core` and not `poetry`.

PetterS:
Right, it is the format of the lock file and possibly the virtualenv that is the issue.

But alright, it makes upgrading from 1.0 to 1.1 harder but not impossible.

I have found that if the system version of poetry is 1.0 and 1.1 is added as a dep into the pyproject.toml, the newer version will be used for the project from that point on. That feels a bit nonstandard though.

abn:
[...]

PR #9547, opened 2024-07-14
Project plugins and forcing a minimum poetry version

radoering:
# Pull Request Check List

Related-to: #5740 (refines the plugin part, omits the tools part)
Resolves: #5729
Resolves: #3316

Introduces `tool.poetry.requires-poetry` and `tool.poetry.requires-plugins`.

Example:

```toml
[tool.poetry]
requires-poetry = ">=2.0"

[tool.poetry.requires-plugins]
poetry-plugin-bundle = ">=1.3"
poetry-plugin-export = ">=1.7.0"
```

`tool.poetry.requires-poetry` enforces a Poetry version and raises an error if the used version is not allowed, e.g.:

```
This project requires Poetry <2.0, but you are using Poetry 2.0
```

(Of course, this only works for future Poetry versions because older versions do not know this setting.)

In `tool.poetry.requires-plugins` you can define plugins that are required for the project. If these plugins are not installed in Poetry's own environment, they are installed only for the project into `.poetry/plugins` inside the project folder.

<details>
<summary>rejected first draft</summary>
Introduces a `tool.poetry.self` section in pyproject.toml. Example:

```toml
[tool.poetry.self]
version = ">=2.0"

[tool.poetry.self.plugins]
poetry-plugin-bundle = ">=1.3"
poetry-plugin-export = ">=1.7.0"
```
</details>


Secrus:
Some bikeshedding from my side. I don't like the new `self` section. When I was working on those issues, I was adding fields directly under the `tool.poetry` namespace. We already have `self` referring to the global Poetry installation and IMHO this will be confusing if one `self` was global and one per project. My proposal would be:
`tool.poetry.self.version` -> `tool.poetry.requires-poetry` (mimicking `requires-python` from PEP 621 spec)
`tool.poetry.self.plugins` -> `tool.poetry.project-plugins`

radoering:
> We already have self referring to the global Poetry installation and IMHO this will be confusing if one self was global and one per project.

Yeah, that has already come to my mind, but I couldn't think of anything better. At least, I like the grouping of required version and required plugins into one section.

> `tool.poetry.self.version` -> `tool.poetry.requires-poetry` (mimicking `requires-python` from PEP 621 spec)

That is an option. The drawback is that it is not grouped anymore with the required plugins into one section.

> `tool.poetry.self.plugins` -> `tool.poetry.project-plugins`

Not sure about that because plugins are only installed if they are not available in the global Poetry installation. (I think that makes more sense than just always installing them into the project.) Maybe, `requires-plugins`? I think it is more like a detail that required plugins that are not available are installed automatically only for the project. We could even introduce a setting to turn this off and just fail if the required plugins are not available (as for the version). (I do not think that we need such a setting but that can change if we get feedback after releasing this into the wild.)

Secrus:
> That is an option. The drawback is that it is not grouped anymore with the required plugins into one section.
Do we need those two features to be grouped together? I don't think so, but maybe you have some good reasons for that

radoering:
It just belongs together because both are requirements for Poetry it*self*.

By the way, when running `poetry self show plugins`, both, global plugins and project plugins will be displayed. Actually, it shouldn't even matter if a plugin is installed globally or in the project folder (except for debugging purposes if something does not work).

Actually, I still think `tool.poetry.self.version` and `tool.poetry.self.plugins` is not a bad option. On the other hand, I probably could also get used to `tool.poetry.requires-poetry` and `tool.poetry.requires-plugins`.

johnthagen:
@radoering Thanks for the ping.

First, a quick question. Is it assumed that `[project]` support will be merged as well before this?

- #9135

The reason I ask, is that that PR would allow users to move a lot of metadata _out_ of `[tool.poetry]`. That then makes `[tool.poetry]` a cleaner space to be able to focus on these kinds of internal Poetry-specific metadata (for example `package-mode = false` etc.) and that could sway my thoughts.

radoering:
Although this could come before `[project]` support, I think both will be released together. However, even with `[project]` support, the fields in `[tool.poetry]` are still there (although many of them deprecated) so at least we cannot reuse names. (But I understand that there will be less in `[tool.poetry]` when using the `[project]` section.)

johnthagen:
> are still there (although many of them deprecated) so at least we cannot reuse names

That is a good point to consider. After there any fields that we know currently we'd need to use?

I do not have a super strong opinion, but given that `package-mode` is already under `[tool.poetry]`, I think it is best to go with option #2 requires field

```toml
[project]
# ...

[tool.poetry]
# ...
package-mode = false
requires-poetry = ">=2.0"

[tool.poetry.requires-plugins]
poetry-plugin-bundle = ">=1.3"
poetry-plugin-export = ">=1.7.0"
```

To me this seems like the best long term option.

radoering:
@johnthagen Thanks for your input. One thing that came to my mind: Is `requires-plugins` unambiguous enough or is `requires-poetry-plugins` better?

johnthagen:
@radoering `requires-plugins` seems fine to me given that it's already nested inside:

```toml
[tool.poetry.requires-plugins]
```

The `.poetry` part of that to me implies that we're already talking about poetry-specific configuration (especially once everyone can move all of the wheel metadata out of this into `[project]`).

radoering:
I updated the PR from `tool.poetry.self.version` and `tool.poetry.self.plugins` to `tool.poetry.requires-poetry` and `tool.poetry.requires-plugins`.

src/poetry/plugins/plugin_manager.py
```diff
@@ -80,4 +100,209 @@ def _load_plugin_entry_point(self, ep: metadata.EntryPoint) -> None:
                 "The Poetry plugin must be an instance of Plugin or ApplicationPlugin"
             )
 
-        self.add_plugin(plugin())
+        self._add_plugin(plugin())
+
+
+class ProjectPluginCache:
+    PATH = Path(".poetry") / "plugins"
```
Secrus, line 107: Please add writing `.gitginore` file to the `.poetry` directory by default so that we can avoid people committing it to their repositories.

src/poetry/repositories/installed_repository.py
```diff
@@ -239,7 +242,7 @@ def load(cls, env: Env, with_dependencies: bool = False) -> InstalledRepository:
         seen = set()
         skipped = set()
 
-        for entry in reversed(env.sys_path):
+        for entry in env.sys_path:
```
Secrus, line 245: Any special reason for that removal?

radoering: I think it is wrong. Actually, I was wondering why it should be reversed order. The correct order is required to overwrite the version of a global plugin. The following test (I added) will fail with reversed order: `test_loads_in_correct_sys_path_order`

src/poetry/plugins/plugin_manager.py
radoering: done

src/poetry/plugins/plugin_manager.py
```diff
@@ -23,53 +47,49 @@ class PluginManager:
...
-            ep
-            for ep in metadata.entry_points(group=self._group)
-            if self._is_plugin_candidate(ep, env)
-        ]
+    def activate(self, *args: Any, **kwargs: Any) -> None:
+        for plugin in self._plugins:
+            plugin.activate(*args, **kwargs)
 
-    def add_plugin(self, plugin: Plugin) -> None:
+    def _add_plugin(self, plugin: Plugin) -> None:
```
Secrus, line 85: Nit: any particular reason for making this private?

src/poetry/plugins/plugin_manager.py
```diff
@@ -80,4 +100,219 @@ def _load_plugin_entry_point(self, ep: metadata.EntryPoint) -> None:
...
+
+class ProjectPluginCache:
+    PATH = Path(".poetry") / "plugins"
+
+    def __init__(self, poetry: Poetry, io: IO) -> None:
+        self._poetry = poetry
+        self._io = io
+        self._path = poetry.pyproject_path.parent / self.PATH
+        self._config_file = self._path / "config.toml"
+        self._gitignore_file = self._path / ".gitignore"
```
Secrus, line 114: I think this should point to `.poetry/.gitignore` so that if we use `.poetry` for another purpose in the future, we don't have to multiply the `.gitignore` files.

src/poetry/plugins/plugin_manager.py
```diff
@@ -80,4 +100,219 @@ def _load_plugin_entry_point(self, ep: metadata.EntryPoint) -> None:
...
+
+        # force new package to be installed in the project cache instead of Poetry's env
+        poetry_env.paths["platlib"] = str(self._path)
+        poetry_env.paths["purelib"] = str(self._path)
+
+        # Build installed repository from locked packages so that plugins
+        # that may be overwritten are not included.
+        repo = Repository("poetry-repo")
+        for package in locked_packages:
+            repo.add_package(package)
```
Secrus, line 284:
```suggestion
        repo = Repository("poetry-repo", locked_packages)
```
We can also eliminate that variable all together and put `Repository("poetry-repo", locked_packages)` directly in the `Installer` init in line 295.

Secrus (commented):
Some more comments after another review pass.

src/poetry/plugins/plugin_manager.py
radoering: done

src/poetry/plugins/plugin_manager.py
radoering: done

Secrus (approved):

Issue #2310, opened 2020-04-14
Improve script support


abn:
### Rationale
Poetry, today, only allows for the handling of [console_scripts](https://python-packaging.readthedocs.io/en/latest/command-line-scripts.html#the-console-scripts-entry-point). A common use case for python packages is to make available shell scripts in the user's path on package installation. The latter is accomplishe by setuptools using the [scripts](https://python-packaging.readthedocs.io/en/latest/command-line-scripts.html#the-scripts-keyword-argument) keyword argument to `setup()`.

### Specification
In order to achive this, we need to expand the capabilities offered by the [scripts section](https://python-poetry.org/docs/pyproject/#scripts) of the poetry configuration. Today, we expect `console_scripts` to be configured as follows.

```toml
[tool.poetry.scripts]
poetry = 'poetry.console:run'
```

This should be extended to support inline tables as shown here.

```toml
[tool.poetry.scripts]
my-script = {type = "file", path = "bin/my-script.sh"}
my-python-script = {type = "entrypoint", name = "package.module:function"}
```

While it is possible to detect if a given value is a method or script, the team feels that using an inline table here will allow for future enhancements to this feature. And furthermore, it allows us to be explicit rather than implicit with respect to the intention.

An implementation of this feature should ensure the following.

* Installation of a project configured so, should make available `my-script` and `my-python-script` in the `PATH`.
* At the initial implementation only `file` and `entrypoint` will be allowed as legal vlues for `type`.
* An entry of the `file` type should only specify the `path` property. The value of `path` must be an existing file.
* An entry of the `entrypoint` type should only have `name` property specified. The value of which must be a valid entrypoint function.

### Backwards compatibility
To ensure older configuration will still work, string values must also be accepted as a script value. This will default to treating it as a script of type `entrypoint`. The following will be equivalent.

```toml
[tool.poetry.scripts]
poetry = 'poetry.console:run'
```

```toml
[tool.poetry.scripts]
poetry = {type = "entrypoint", name = "poetry.console:run"}
```

### Related Issues
- #241
- #1023
- #1504

funkyfuture:
fyi, i started drafting another proposal. hope to be finished by tomorrow if not today.

funkyfuture:
a question unrelated to my proposal, how would you accomplish to drop the file extension here?

```toml
[tool.poetry.scripts]
my-script = {path = "bin/my-script.sh"}
```

i'm assuming that you propose to install that file as `my-script`.

[...]

PR #9549, opened 2024-07-15
Enable GUI scripts

IgnaciodelaTorreArias:
Resolves: #2310 

## FEATURE:  ability to define gui scripts 

The implementation is based on the official specification by @abn [#2310](https://github.com/python-poetry/poetry/issues/2310)

### Created a script to be used with Windows gui_scripts, when using poetry install. 

1. The script starts a new window with an empty title "", in this new cmd pythonw is executed with the python script that is created for unix systems as argument (just as in console scripts), any other arguments are forwarded (just as in console scripts), [The start command](https://learn.microsoft.com/en-us/windows-server/administration/windows-commands/start)
2.  Since pythonw doesn't starts/need a command window only the GUI window is opened, if the python script is defined correctly on the scripts section.
3. After starting pythonw the WINDOWS_GUI_CMD_TEMPLATE finishes, this means that if {script-name}.cmd  was executed by clicking on it this will close the command window that was opened.

**Questions:**
To be able to run pythonw i used string functions [split, replace, join] modifying the python path provided, i don't know if this is correct?
Using pythonw means this only works with the CPython implementation, is this OK?

### Updated documentation corresponding to the scripts section.

Based on [PyPA entry-points specification](https://packaging.python.org/en/latest/specifications/entry-points/) defined console scripts and gui scripts.
Also added a warning deprecation for 

```toml
[tool.poetry.scripts]
my_package_cli = 'my_package.console:run'
```

this is because of a [TODO](https://github.com/python-poetry/poetry-core/blob/06d72bc0f1f05086a1cc0e760075c519af33db5b/src/poetry/core/masonry/builders/builder.py#L301)

### Example

```toml
[tool.poetry.scripts]
my_pachkage_gui = {reference='my_package.gui:run', type='gui'}
```

### testing

- Tested manually with a repository a created with this purpose [DearPyGUI-example-poetry](https://github.com/IgnaciodelaTorreArias/DearPyGUI-example-poetry)
For this testing i also needed to modify poetry-core, i will make a pull request after this one is merged.

### DIFF
--- a/docs/pyproject.md
+++ b/docs/pyproject.md
@@ -366,10 +366,24 @@ This section describes the scripts or executables that will be installed when in
 
 ```toml
 [tool.poetry.scripts]
-my_package_cli = 'my_package.console:run'
+my_package_cli = {reference='my_package.console:run', type='console'}
+my_pachkage_gui = {reference='my_package.gui:run', type='gui'}
 ```
 
 Here, we will have the `my_package_cli` script installed which will execute the `run` function in the `console` module in the `my_package` package.
+
+valid script types:
+
+* console. Usable as a command in a system shell after the package is installed
+* gui. Differs only on windows, where they can be started without a console window, using pythonw instead of python, this means they cannot use standard streams unless application code redirects them. On non windows platforms they are treated as console scripts.
+
+{{% warning %}}
+string defined scripts are deprecated:
+```toml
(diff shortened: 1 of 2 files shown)


dimbleby:
haven't looked at the code but, any implementation for this should have corresponding changes in poetry-core for building wheels

Issue #9537, opened 2024-07-08
Poetry lock does  not include explicit sources for conditional packages


Holt59:
### Description

Following `pyproject.toml` from https://github.com/python-poetry/poetry/issues/6409 to allow user to install CPU or GPU versions of torch via `poetry install -E cpu --with cpu` or `poetry install -E cuda --with cuda`.

CUDA installation works fine, but CPU does not, throws a `Repository "pytorch-cpu" does not exist.`. When checking the `poetry.lock` file, the explicit source `pytorch-cpu` is missing and I cannot find a way to include it.

### Workarounds

No workaround for now, I am completely stuck.

### Poetry Installation Method

install.python-poetry.org

### Operating System

Ubuntu 24.04

### Poetry Version

1.8.3

### Poetry Configuration

```bash session
cache-dir = "/home/XXX/.cache/pypoetry"
experimental.system-git-client = false
installer.max-workers = null
installer.modern-installation = true
installer.no-binary = null
installer.parallel = true
keyring.enabled = false
repositories.pytorch-cpu.url = "https://download.pytorch.org/whl/cpu"
repositories.pytorch-cuda.url = "https://download.pytorch.org/whl/cu118"
solver.lazy-wheel = true
virtualenvs.create = true
virtualenvs.in-project = null
virtualenvs.options.always-copy = false
virtualenvs.options.no-pip = false
virtualenvs.options.no-setuptools = false
virtualenvs.options.system-site-packages = false
virtualenvs.path = "{cache-dir}/virtualenvs"  # /home/XXX/.cache/pypoetry/virtualenvs
virtualenvs.prefer-active-python = false
virtualenvs.prompt = "{project_name}-py{python_version}"
warnings.export = true
```

### Example pyproject.toml

```TOML
[tool.poetry]
name = "pkg"
version = "0.1.0"
description = ""
authors = ["XXX <xxx@xxx.xxx>"]
readme = "README.md"

[[tool.poetry.source]]
name = "pytorch-cuda"
url = "https://download.pytorch.org/whl/cu118"
priority = "explicit"

[[tool.poetry.source]]
name = "pytorch-cpu"
url = "https://download.pytorch.org/whl/cpu"
priority = "explicit"

[tool.poetry.dependencies]
python = "^3.12"

[tool.poetry.group.cpu]
optional = true

[tool.poetry.group.cpu.dependencies]
[...]

Issue #7748, opened 2023-03-31
Extra marker is ignored in `poetry install`, but respected in `pip install`


Omegastick:
- **Poetry version**: 1.4.1
- **Python version**: 3.9
- **OS version and name**: Ubuntu 22.04 on WSL2
- **pyproject.toml**: https://gist.github.com/Omegastick/713164402b450b16b35c84f425db0a29

- [X] I am on the [latest](https://github.com/python-poetry/poetry/releases/latest) stable Poetry version, installed using a recommended method.
- [X] I have searched the [issues](https://github.com/python-poetry/poetry/issues) of this repo and believe that this is not a duplicate.
- [X] I have consulted the [FAQ](https://python-poetry.org/docs/faq/) and [blog](https://python-poetry.org/blog/) for any relevant entries or release notes.
- [X] If an exception occurs when executing a command, I executed it again in debug mode (`-vvv` option) and have included the output below.

## Issue
If my understanding is correct, a dependency like this should install the PyPI repository version of the dependency if installed as `package_a[repo]`, and the path dependency if installed as `package_a{local]`:
```toml
[tool.poetry.dependencies]
package-b = [
    {version = "^0.1.0", markers = "extra == 'repo'", optional = true},
    {path = "../package_b", develop = true, markers = "extra == 'local'", optional = true},
]

[tool.poetry.extras]
repo = ["package-b"]
local = ["package-b"]
```
`pip install -e .[local]` installs the path dependency, and `pip install -e .[repo]` fails because `package_b` doesn't exist (as it should). `poetry install -E local` and `poetry install -E repo` both install the path dependency, though.

## Use case
I have quite a few repos, all using Poetry, that I would like to install locally with a single `poetry install` at the top of the dependency tree.

Example dependency tree:
```
package_a -> package_b -> package_c
```

If you alter the `package_b` dependency specification to add the right extras:
```toml
package-b = [
    {version = "^0.1.0", extras = ["repo"], markers = "extra == 'repo'", optional = true},
    {path = "../package_b", extras = ["local"], develop = true, markers = "extra == 'local'", optional = true},
]
```

You could use `poetry install -E local` to install the whole tree for local development. `pip install -e .[local]` seems to work this way, unless I'm missing something.

I'm aware that including path dependencies outside of groups specifies the full path (not relative) in `PKG-INFO`. That's not a problem for us, because the packages are only being uploaded to a private PyPI.

dimbleby:
[...]

Issue #6419, opened 2022-09-06
Support for conflicting dep extras, when they are exclusive of each other.


erezsh:
- [x] I have searched the [issues](https://github.com/python-poetry/poetry/issues) of this repo and believe that this is not a duplicate.
- [x] I have searched the [documentation](https://python-poetry.org/docs/) and believe that my question is not covered.

### Background
I am working on a project called data-diff (https://github.com/datafold/data-diff), which supports a lot of optional 3rd party libraries, mostly connectors to different databases. We provide them as "extras", so that users can install them easily if they wish. For example, they might run `pip install data-diff[mysql, snowflake, bigquery]`.

For the most part, these libraries play nice with each other, except that sometimes they don't. For example, adding the `databricks` connector (https://github.com/databricks/databricks-sql-python) is somehow causing a conflict with `snowflake` on numpy/pandas/etc., which can only be resolved, if at all, by taking all these libraries years back, which is bad for both features and security.

This kind of conflict resolution really isn't necessary for our use case, because it's very unlikely for someone to install both `databricks` and `snowflake`, and even when they do, we would much prefer to give them a special error message, over the alternative which is removing `databricks` from "extras" entirely.

It would be great if poetry offered some solution to this situation, where extras conflict, but don't have to co-exist.

### Request (simple)

One way to do it would be to add an option to make some extras exclusive of one another, so they can't be installed together, but also don't affect each other's dependencies.

Maybe add the option to make dependency groups exclusive. Anything within a groups can be installed together, but deps from separate groups cannot.

### Request (ideal)

What I really want in this situation, is to be able to install dependencies in their own "virtual env", so that when they import their dependencies, they import them from their own private cache.

That would allow all these 3rd party libraries to play nice with each other, no matter what.

This might be beyond the scope of poetry, in more ways than one. But I also think it could be useful for a lot of other scenarios..

LarsDu:
Having the **exact same** issue!

We would like to have mutually exclusive extras groups that do not trigger the solver collectively. This type of feature would be a great boon for teams that operate out of the same repo, but may require different packages.

LarsDu:
Also if anyone could provide information on how extras propagate into the solver, that would be extremely helpful (for making a potential PR)

neersighted:
Related to (but not the same as #1168)

martinitus:
[...]

Issue #834, opened 2019-01-25
How can I specify differing sets of dependency extras for each of my project's extras?


kurtiss:
- [x] I have searched the [issues](https://github.com/sdispater/poetry/issues) of this repo and believe that this is not a duplicate.
- [x] I have searched the [documentation](https://poetry.eustace.io/docs/) and believe that my question is not covered.

## Question
How can I specify differing sets of dependency extras for each of my project's extras? This is something that is relatively trivial to do with setup.py. To clarify, I would like to be able to specify the following dependency graph:

> myproject[feature-a] depends on [dependency-one, dependency-two[feature-a]]
> myproject[feature-b] depends on [dependency-one, dependency-two[feature-b]]

I'll propose a couple ways that pyproject.toml might model these graphs, but the core value of these examples is to highlight an apparent deficiency with the current model. 

1) By allowing dependency aliases that specify distinct sets of extras. For example:

```
[tool.poetry.dependencies]
…
dependency-two-a = { git = "git+ssh://git@company.githost.io/projects/dependency-two.git",optional = true,extras = ["feature-a"], provides="dependency-two" }
dependency-two-b = { git = "git+ssh://git@company.githost.io/projects/dependency-two.git",optional = true,extras = ["feature-b"], provides="dependency-two" }
…

[tool.poetry.extras]
feature-a = ["dependency-one", "dependency-two-a"]
feature-b = ["dependency-one", "dependency-two-b"]
```

2) By treating the set of extras in the dependencies section as a superset that can be further filtered down when the dependency is referenced in the extras section. For example:

```
[tool.poetry.dependencies]
…
dependency-two = { git = "git+ssh://git@company.githost.io/projects/dependency-two.git",optional = true,extras = ["feature-a", "feature-b"]}
…

[tool.poetry.extras]
feature-a = ["dependency-one", "dependency-two[feature-a]"]
feature-b = ["dependency-one", "dependency-two[feature-b]"]
```

Thanks for your time!

kurtiss:
@sdispater Gently bumping this – is this use case compelling enough to warrant some attention?

kurtiss:
Bumping again, because the issue is outstanding and marks a flaw in the way dependencies are modeled, preventing reasonable use of this application. Hope somebody can take a look at this.

wshayes:
Looks like you can install multiple dependencies in a single set using the following approach:

```
poetry add -E api fastapi
poetry add -E api gunicorn
poetry add -E api uvicorn
[...]

PR #9553, opened 2024-07-15
Pass Install Extras to Markers

reesehyde:
# Pull Request Check List

Resolves:
- #834 
- #6409 
- #6419
- #7748  
- #9537 

This PR exposes the `extras` supplied at install to the 'extra' marker for dependencies, taking up @radoering's [offer](https://github.com/python-poetry/poetry/issues/6409#issuecomment-1799026967) to create a PR implementing this functionality to allow switching torch installation versions based on a `cuda` extra.

It is a duplicate of the feature in #613 but it appears that PR is no longer active, with the last updates in August 2023 and an unanswered question about timeline from March 2024.

See issues and unit tests for more details but the idea is to, among other things, enable exclusive extras like so:
```toml
[tool.poetry]
package-mode = false

[tool.poetry.dependencies]
python = "^3.10"
click = [
    { markers = "extra == 'one' and extra != 'two' and extra != 'three'", version = "8.1.1", optional = true},
    { markers = "extra != 'one' and extra == 'two' and extra != 'three'", version = "8.1.2", optional = true},
    { markers = "extra != 'one' and extra != 'two' and extra == 'three'", version = "8.1.3", optional = true}
 ]

[tool.poetry.extras]
one = ["click"]
two = ["click"]
three = ["click"]
```

@radoering it looks like you're very familiar with this issue, would you be the right person to review? :pray:  
I'm sure I'm missing something here but look forward to iterating!

### DIFF
--- a/src/poetry/puzzle/provider.py
+++ b/src/poetry/puzzle/provider.py
@@ -134,7 +134,7 @@ class Provider:
         self._direct_origin_packages: dict[str, Package] = {}
         self._locked: dict[NormalizedName, list[DependencyPackage]] = defaultdict(list)
         self._use_latest: Collection[NormalizedName] = []
-        self._active_root_extras = frozenset(active_root_extras) if active_root_extras is not None else frozenset()
+        self._active_root_extras = frozenset(active_root_extras) if active_root_extras is not None else None
 
         self._explicit_sources: dict[str, str] = {}
         for package in locked or []:
--- a/src/poetry/puzzle/provider.py
+++ b/src/poetry/puzzle/provider.py
@@ -7,7 +7,7 @@ import time
 
 from collections import defaultdict
 from contextlib import contextmanager
-from typing import TYPE_CHECKING
+from typing import TYPE_CHECKING, Any
 from typing import ClassVar
 from typing import cast
 
@@ -117,6 +117,7 @@ class Provider:
         *,
         installed: list[Package] | None = None,
         locked: list[Package] | None = None,
+        active_root_extras: Collection[NormalizedName] | None = None,
     ) -> None:
         self._package = package
         self._pool = pool
@@ -133,6 +134,7 @@ class Provider:
         self._direct_origin_packages: dict[str, Package] = {}
         self._locked: dict[NormalizedName, list[DependencyPackage]] = defaultdict(list)
         self._use_latest: Collection[NormalizedName] = []
+        self._active_root_extras = frozenset(active_root_extras) if active_root_extras is not None else frozenset()
 
         self._explicit_sources: dict[str, str] = {}
         for package in locked or []:
@@ -457,7 +459,12 @@ class Provider:
             for dep in dependencies
             if dep.name not in self.UNSAFE_PACKAGES
             and self._python_constraint.allows_any(dep.python_constraint)
-            and (not self._env or dep.marker.validate(self._env.marker_env))
+            and (not self._env or dep.marker.validate(
+                self._marker_values(
+                    self._active_root_extras if dependency_package.package.is_root()
+                    else dependency_package.dependency.extras)
+                )
+            )
         ]
         dependencies = self._get_dependencies_with_overrides(_dependencies, package)
 
@@ -541,7 +548,8 @@ class Provider:
             if dep.name in self.UNSAFE_PACKAGES:
                 continue
 
-            if self._env and not dep.marker.validate(self._env.marker_env):
+            active_extras = self._active_root_extras if package.is_root() else dependency.extras
+            if self._env and not dep.marker.validate(self._marker_values(active_extras)):
                 continue
 
             if not package.is_root() and (
@@ -601,7 +609,7 @@ class Provider:
 
             # For dependency resolution, markers of duplicate dependencies must be
             # mutually exclusive.
-            active_extras = None if package.is_root() else dependency.extras
+            active_extras = self._active_root_extras if package.is_root() else dependency.extras
             deps = self._resolve_overlapping_markers(package, deps, active_extras)
 
             if len(deps) == 1:
@@ -860,7 +868,7 @@ class Provider:
                 get_python_constraint_from_marker(marker)
             )
             and (active_extras is None or marker.validate({"extra": active_extras}))
-            and (not self._env or marker.validate(self._env.marker_env))
+            and (not self._env or marker.validate(self._marker_values(active_extras)))
         )
 
     def _resolve_overlapping_markers(
@@ -953,3 +961,17 @@ class Provider:
         # dependencies by constraint again. After overlapping markers were
         # resolved, there might be new dependencies with the same constraint.
         return self._merge_dependencies_by_constraint(new_dependencies)
+
+    def _marker_values(self, extras: set[NormalizedName] | None = None) -> dict[str, Any]:
+        """
+        Marker values, per:
+        1. marker_env of `self._env`
+        2. 'extras' will be added to the 'extra' marker if not already present
+        """
+        result = self._env.marker_env.copy()
+        if extras is not None:
+            if 'extra' not in result.keys():
+                result['extra'] = extras
+            else:
+                result['extra'] = set(result['extra']).union(extras)
+        return result
--- a/src/poetry/puzzle/provider.py
+++ b/src/poetry/puzzle/provider.py
@@ -962,16 +962,16 @@ class Provider:
         # resolved, there might be new dependencies with the same constraint.
         return self._merge_dependencies_by_constraint(new_dependencies)
 
-    def _marker_values(self, extras: set[NormalizedName] | None = None) -> dict[str, Any]:
+    def _marker_values(self, extras: Collection[NormalizedName] | None = None) -> dict[str, Any]:
         """
         Marker values, per:
         1. marker_env of `self._env`
         2. 'extras' will be added to the 'extra' marker if not already present
         """
-        result = self._env.marker_env.copy()
+        result = self._env.marker_env.copy() if self._env is not None else {}
         if extras is not None:
             if 'extra' not in result.keys():
-                result['extra'] = extras
+                result['extra'] = set(extras)
             else:
                 result['extra'] = set(result['extra']).union(extras)
         return result
--- a/docs/dependency-specification.md
+++ b/docs/dependency-specification.md
@@ -300,6 +300,51 @@ via the `markers` property:
 pathlib2 = { version = "^2.2", markers = "python_version <= '3.4' or sys_platform == 'win32'" }
 ```
 
+### `extra` environment marker
+
+Poetry populates the `extra` marker with each of the selected extras for the parent declaring the dependency. For 
+example, consider the following dependency in your root package:
+```toml
+[tool.poetry.dependencies]
+pathlib2 = { version = "^2.2", markers = "extra == 'paths' and sys_platform == 'win32'", optional = true}
+```
+`pathlib2` will be installed when you install your package with `--extras paths` on a `win32` machine. 
+You'll also need to [define the `paths` extra in your project](./pyproject.md#extras).
+
+#### Exclusive extras
+
+Keep in mind that all combinations of extras available in your project need to be compatible with each other. This 
+means that in order to use differing or incompatible versions across different extras, you need to make your extra 
+markers *exclusive*. For example, the following installs PyTorch from one source repository with CPU versions when 
+the `cpu` extra is specified, while the other installs from another repository with a separate version set for GPUs
+when the `cuda` extra is specified:
+
+```toml
+[tool.poetry.dependencies]
+torch = [
+    { markers = "extra == 'cpu' and extra != 'cuda'", version = "2.3.1+cpu", source = "pytorch-cpu", optional = true},
+    { markers = "extra != 'cpu' and extra == 'cuda'", version = "2.3.1+cu118", source = "pytorch-cu118", optional = true},
+ ]
+
+[tool.poetry.extras]
+cpu = ["torch"]
+cuda = ["torch"]
+
+[[tool.poetry.source]]
+name = "pytorch-cpu"
+url = "https://download.pytorch.org/whl/cpu"
+priority = "explicit"
+
+[[tool.poetry.source]]
+name = "pytorch-cu118"
+url = "https://download.pytorch.org/whl/cu118"
+priority = "explicit"
+```
+
+For the `cpu` extra marker, we have to specify `"extra == 'cpu' and extra != 'cuda'"` and vice-versa for the `cuda` 
+extra marker. There is no version that satisfies both constraints, so installing with `--extras cpu --extras gpu` 
+would match no cases and not install `torch`.
+
 ## Multiple constraints dependencies
 
 Sometimes, one of your dependency may have different version ranges depending
--- a/docs/dependency-specification.md
+++ b/docs/dependency-specification.md
@@ -313,21 +313,20 @@ You'll also need to [define the `paths` extra in your project](./pyproject.md#ex
 
 #### Exclusive extras
 
-Keep in mind that all combinations of extras available in your project need to be compatible with each other. This 
-means that in order to use differing or incompatible versions across different extras, you need to make your extra 
-markers *exclusive*. For example, the following installs PyTorch from one source repository with CPU versions when 
-the `cpu` extra is specified, while the other installs from another repository with a separate version set for GPUs
-when the `cuda` extra is specified:
+Keep in mind that all combinations of possible extras available in your project need to be compatible with each other. 
+This means that in order to use differing or incompatible versions across different combinations, you need to make your 
+extra markers *exclusive*. For example, the following installs PyTorch from one source repository with CPU versions 
+when the `cuda` extra is *not* specified, while the other installs from another repository with a separate version set 
+for GPUs when the `cuda` extra *is* specified:
 
 ```toml
 [tool.poetry.dependencies]
 torch = [
-    { markers = "extra == 'cpu' and extra != 'cuda'", version = "2.3.1+cpu", source = "pytorch-cpu", optional = true},
-    { markers = "extra != 'cpu' and extra == 'cuda'", version = "2.3.1+cu118", source = "pytorch-cu118", optional = true},
+    { markers = "extra != 'cuda'", version = "2.3.1+cpu", source = "pytorch-cpu", optional = true},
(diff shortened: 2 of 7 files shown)


radoering:
I will try to take a close look at the end of the week.

src/poetry/puzzle/provider.py
```diff
@@ -541,7 +555,12 @@ def complete_package(
             if dep.name in self.UNSAFE_PACKAGES:
                 continue
 
-            if self._env and not dep.marker.validate(self._env.marker_env):
+            active_extras = (
+                self._active_root_extras if package.is_root() else dependency.extras
```
radoering, line 559: Seems like the non root part is not tested yet. (Tests succeed when replacing `dependency.extras` with `None`.) Maybe, this path can be triggered by taking one of your tests and introducing an intermediate package?

src/poetry/puzzle/provider.py
```diff
@@ -457,7 +462,16 @@ def incompatibilities_for(
...
+            and (
+                not self._env
+                or dep.marker.validate(
+                    self._marker_values(
+                        self._active_root_extras
+                        if dependency_package.package.is_root()
+                        else dependency_package.dependency.extras
+                    )
+                )
+            )
```
radoering, line 474: It looks like this change is not required by any test. To be honest, I have currently no idea if this change is redundant or if a test is missing...

radoering (changes requested):
This looks really good. Nice work. 👍

It looks like some of the changes are not tested yet or may be redundant. See separate comments for details.

reesehyde:
Thank you so much for the great feedback here! All the comments make sense to me, I'm working through it and will submit updates and responses later this week

GeeCastro:
How would this differentiate between the `extras` for the project itself and the `extras` for a dependency package?

1. For the project the scenario would be "I want to install my project with cuda124 extra" which I guess you could interface with the -E/--extra option
```bash
poetry install -E cuda124
```
and define the in the package
```toml
[tool.poetry.extras]
cuda124 = ["torch"]
```
Here your suggestion would add 
```toml
[tool.poetry.dependencies]
python = "^3.10"
torch = [
    { markers = "extra == 'cuda124'", version = "2.4.0+cu124", optional = true},
                ^^^^^^
```
2. And a package-specific extra for example 
```bash
poetry add "pandas[aws]"
```
which would show in the `extras` field as follows 
```toml
[tool.poetry.dependencies]
python = "^3.11"
pandas = {extras = ["excel"], version = "^2.2.2"}
          ^^^^^
```


Would the differentiation be made on the following?
1. project extras in the `marker` string
2. package's extras in the extras field 

If so, I must have misunderstood the docs because I thought these 2 syntaces express the same requirements/constraints

reesehyde:
I apologize for the long delay here, I had to pause this before I managed to finish all the updates. But it hasn't left my mind, and I'll be able to resume the updates in a couple weeks and will push them up and respond to everything in mid-September! Thanks for the patience

mg515:
@reesehyde Any update on this? Would love to have this feature available. ^^

src/poetry/puzzle/provider.py
reesehyde: Ah yeah I wasn't sure what the best policy was here: I just updated the places where I saw markers being evaluated to include extras. What I found is that of the three changes in `provider.py`, any one of them is sufficient to pass the first round of tests. So I think there is some redundancy, but my thought is to not eliminate anything yet due to the issue with uncovered around not resolving conflicts with their own different extras.

src/poetry/puzzle/provider.py
reesehyde: Great catch! While trying to add a test for this I uncovered an issue referenced in a top-level comment. The breaking test I've added doesn't include the intermediate package for simplicity but I can add it as we get to the bottom of that.

reesehyde (commented):
I apologize for the long delay here, I've been working on the suggested updates but uncovered an issue in the process. I've [added a couple breaking tests](https://github.com/python-poetry/poetry/pull/9553/commits/fd1c1dc9d13b03ee5fadf90640f0804c887f79ed) which will hopefully help illuminate it. But I'm a little stuck on the best course of action, so would love a push in the right direction if you're able @radoering.

It seems that part of the reason the initial tests passed is because they used conflicting versions of `torch`, which works only because they then go through the process of [merging duplicates](https://github.com/python-poetry/poetry/blob/ca8917781e033cd3f35c6713d4d1bb3f5db34bdb/src/poetry/puzzle/provider.py#L590-L595) since the package name is the same in all cases.

However, dependencies with _their own differing extras_ (e.g. `torch[cpu]` and `torch[cuda]`) are not considered duplicates, so they result in a `SolverProblemError`.
Simply considering them duplicates creates its own issues (e.g. see the [test_solver_resolves_duplicate_dependency_in_extra](https://github.com/python-poetry/poetry/blob/ca8917781e033cd3f35c6713d4d1bb3f5db34bdb/tests/puzzle/test_solver.py#L4722) test). So we need some way to do [overlapping marker resolution](https://github.com/python-poetry/poetry/blob/ca8917781e033cd3f35c6713d4d1bb3f5db34bdb/src/poetry/puzzle/provider.py#L866) in these cases without breaking other solving cases.

Another possibility would be scoping this PR down to not support conflicting dependency extras, but I'm not certain how that would look.

reesehyde:
> How would this differentiate between the extras for the project itself and the extras for a dependency package?

@GeeCastro this is a great question, you're right that they are basically the same syntax! The problem comes up when you have _conflicting extras_, e.g. if the `cuda` extra requires `torch == 1.0.0+cuda` while the `cpu` extra requires `torch == 1.0.0+cpu`. Packages must be resolvable with every possible combination of extras, so if you tried to use both (`--extra cpu --extra cuda`) it would be impossible since `1.0.0+cuda` conflicts with `1.0.0+cpu`. Markers provide a way to say "`cuda` extra requires `torch == 1.0.0+cuda`, **but only if the `cpu` extra isn't also present**" so they don't conflict.

tests/puzzle/test_solver.py
```diff
@@ -4707,6 +4707,47 @@ def test_solver_resolves_duplicate_dependency_in_extra(
...
+def test_solver_resolves_duplicate_dependency_in_root_extra(
+    package: ProjectPackage,
+    pool: RepositoryPool,
+    repo: Repository,
+    io: NullIO,
+    with_extra: bool,
+) -> None:
+    """
+    Without extras, a newer version of A can be chosen than with root extras.
+    """
```
radoering, line 4720: This is only a theoretical issue. In other words, the setup of the test cannot happen when invoking poetry.

First of all, you cannot create this issue with Poetry 1.8 because if you try something like:

```
[tool.poetry.dependencies]
python = "^3.9"
tomli = [
    { version = ">=1" },
    { version = "^1", optional = true },
]

[tool.poetry.extras]
foo = [ "tomli" ]
```

... both dependencies are in the extra. However, with the main branch (by the way, please rebase your branch when it suits you) and PEP 621 support you can define:

```
[project]
# ...
requires-python = ">=3.9"
dependencies = [
    "tomli>=1",
]

foo = [
    "tomli>=1,<2",
]
```

When you debug this, you will notice that the tomli dependency from the extra has no `extra == "foo"` marker. Generally speaking, dependencies of root extras do not have the extra in the marker but only the attribute `_in_extras` set. Thus, a more realistic test will look like that (and does not fail):

```
def test_solver_resolves_duplicate_dependency_in_root_extra(
    package: ProjectPackage,
    pool: RepositoryPool,
    repo: Repository,
    io: NullIO,
) -> None:
    """
    Without extras, a newer version of A could be chosen than with root extras.
    However, only the older version is chosen.
    """
    package_a1 = get_package("A", "1.0")
    package_a2 = get_package("A", "2.0")

    dep = get_dependency("A", ">=1.0")
    package.add_dependency(dep)

    dep_extra = get_dependency("A", "^1.0", optional=True)
    dep_extra._in_extras = ["foo"]
    package.extras = {canonicalize_name("foo"): [dep_extra]}
    package.add_dependency(dep_extra)

    repo.add_package(package_a1)
    repo.add_package(package_a2)

    solver = Solver(package, pool, [], [], io)
    transaction = solver.solve()

    check_solver_result(transaction, ([{"job": "install", "package": package_a1}]))
```

I only checked this test for now and will take a look at the installer tests later.

src/poetry/puzzle/provider.py
reesehyde: Confirmed that this change is redundant: either this change or the one on lines 558 are sufficient. Happy to leave this in or to remove it since we haven't found a case depending on this case alone.

In the one or two cases I looked at the other change alone took the same number of steps as with this change, while this change alone took more steps. But I'm not sure if there are some cases I didn't find on a glance where having both changes may reduce the number of steps required.

reesehyde (commented):
@radoering thank you again for all the help here! I've pushed updates, responded to all threads, and rebased on `main`.

I believe everything has test coverage now, there are just small open questions on:
1. [Whether to leave redundant changes](https://github.com/python-poetry/poetry/pull/9553#discussion_r1685431088) in both places or isolate to one place
    - Either the [L465 change](https://github.com/python-poetry/poetry/blob/4734c0f03352e7ba42359eef57cc2808e59a73c8/src/poetry/puzzle/provider.py#L465-L474) or [L558 change](https://github.com/python-poetry/poetry/blob/4734c0f03352e7ba42359eef57cc2808e59a73c8/src/poetry/puzzle/provider.py#L558-L564) are sufficient to pass the new `test_installer` tests, but not sure if it's worth including both for "marker value correctness" or in cases where having both may shorten the solving steps required.
1. [Whether to leave or remove](https://github.com/python-poetry/poetry/pull/9553#discussion_r1685428038) the changes and test which rely on root extra dependencies having the extra marker value even though it's not generally populated in those cases
    - I added test coverage for the [L623 change](https://github.com/python-poetry/poetry/blob/4734c0f03352e7ba42359eef57cc2808e59a73c8/src/poetry/puzzle/provider.py#L623-L625) but you mentioned that test scenario may be unrealistic

**Edit:** The bug referenced was still present, but I added an improved test for it and finally [found a fix](https://github.com/python-poetry/poetry/pull/9553#discussion_r1842900077). Final open question is:

 3. A review of [that new logic](https://github.com/python-poetry/poetry/pull/9553/commits/1326833b925820e2451a7c8a42b5e25c60bb2670)

src/poetry/puzzle/provider.py
reesehyde: > that breaking test relied on conflicting versions of the same package between extras which I don't think is a realistic situation that needs to be covered.

Not sure what I was thinking here, that's absolutely an important case in transitive dependencies as well as root (e.g. what if your dependency has a `--cuda` flag to switch Pytorch versions, and you want your own `--cuda` flag to select between the conflicting cpu and cuda versions in the transitive Pytorch dependency).

This was in fact [the bug identified earlier](https://github.com/python-poetry/poetry/pull/9553#pullrequestreview-2341388336). I added a better test based on the earlier one to illustrate it, which covers conflicting dependencies in root and non-root cases. It would pass in the root case (due to duplicate resolution) but fail in the non-root case.

I then **pushed a fix** in [the latest commit](https://github.com/python-poetry/poetry/pull/9553/commits/1326833b925820e2451a7c8a42b5e25c60bb2670), all tests pass but could use validation that it's not unwise: I updated the duplicate resolution logic to _raise `OverrideNeededError`s_ when duplicates conflict even with different extras, but _not perform overlapping marker resolution_ in those "same package with different extras" cases.

I found that in that case overlapping marker resolution results in incomplete lockfiles which don't contain all possible package versions, while the override exceptions are needed to prevent solver failure.

radoering:
pre-commit.ci autofix

src/poetry/puzzle/provider.py
radoering: > I found that in that case overlapping marker resolution results in incomplete lockfiles which don't contain all possible package versions, while the override exceptions are needed to prevent solver failure.

Yes, overlapping marker resolution is not prepared for duplicates with different extras.

> I updated the duplicate resolution logic to _raise `OverrideNeededError`s_ when duplicates conflict even with different extras, but _not perform overlapping marker resolution_ in those "same package with different extras" cases.

Unfortunately, it is not that simple. The updated logic will only produce correct results if markers are already mutually exclusive. We have to make sure that overrides are not triggered if markers do overlap. I just added a commit with a refined logic. Please take a look.

src/poetry/puzzle/provider.py
```diff
@@ -601,7 +620,9 @@ def complete_package(
 
             # For dependency resolution, markers of duplicate dependencies must be
             # mutually exclusive.
-            active_extras = None if package.is_root() else dependency.extras
+            active_extras = (
+                self._active_root_extras if package.is_root() else dependency.extras
```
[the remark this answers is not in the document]
radoering: I added another comment concerning the test. I think we can keep the test with slight modifications.

radoering (changes requested):
I think we are slowly approaching the finish line. 😄

I addressed all open questions in the relevant threads.

src/poetry/puzzle/provider.py
reesehyde: Thank you for digging in and explaining here, that makes sense to me! Reverted

tests/installation/fixtures/with-conflicting-dependencies-root-extras.test
```diff
@@ -0,0 +1,24 @@
+[[package]]
```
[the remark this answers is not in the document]
reesehyde: Ah yep, I removed the redundant test but not fixture — removed!

tests/puzzle/test_solver.py
```diff
@@ -4765,6 +4765,46 @@ def test_solver_resolves_duplicate_dependency_in_extra(
...
+    dep_extra = get_dependency("A", "^1.0", optional=True)
+    dep_extra.marker = parse_marker("extra == 'foo'")
+    package.extras = {canonicalize_name("foo"): [dep_extra]}
+    package.add_dependency(dep_extra)
+
+    repo.add_package(package_a1)
+    repo.add_package(package_a2)
+
+    solver = Solver(package, pool, [], [], io, active_root_extras=extra)
+    transaction = solver.solve()
```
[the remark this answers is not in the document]
reesehyde: Great point, makes sense to me. Updated!

radoering:
Trying out the examples, I unfortunately discovered some issues:

1. `tool.poetry.extras` will be deprecated in favor of `project.optional-dependencies` in the next version. I updated the examples to the new syntax and thereby noticed a bug in poetry-core, which is especially relevant for custom sources: python-poetry/poetry-core#792
I added a commit to temporarily use my branch with the fix here.
2. In general, the examples did only work for me when setting `poetry config installer.re-resolve false`. Without this setting I always get solver issues when trying to install the examples.

I did not find a solution for the second issue. In general, the issue is that on the one hand we want to filter unused extras in `Provider` to avoid conflicts but on the other hand we do not want to filter them to uninstall unwanted extras when calculating the operations in `Transaction`.

Unfortunately, `poetry config installer.re-resolve false` will not be the default in the next version (but may become the default in long-term) so I am wondering if we just add a warning in the docs that exclusive extras will only work with `poetry config installer.re-resolve false`?

reesehyde:
Thanks for finding and fixing that bug @radoering!

> In general, the examples did only work for me when setting poetry config installer.re-resolve false. Without this setting I always get solver issues when trying to install the examples.

On my investigation here I think the only thing that doesn't work are conflicts between "no extras" and "some extra" when installing the "no extras" case with re-resolve. For example, this can fail:

```toml
[tool.poetry]
package-mode = false

[project]
name = "torch-test"
dependencies = [
    "torch (==2.3.1+cpu) ; extra != 'cuda'",
]

cuda = [
    "torch (==2.3.1+cu118)",
]

[tool.poetry.dependencies]
python = "^3.10"
torch = [
    { markers = "extra != 'cuda'", source = "pytorch-cpu"},
    { markers = "extra == 'cuda'", source = "pytorch-cuda"},
 ]

[[tool.poetry.source]]
name = "pytorch-cpu"
url = "https://download.pytorch.org/whl/cpu"
priority = "explicit"

[[tool.poetry.source]]
name = "pytorch-cuda"
url = "https://download.pytorch.org/whl/cu118"
priority = "explicit"

[build-system]
requires = ["poetry-core>=2.0.0"]
```
When installing with no extras and `re-resolve = true` we get:
```
Incompatible constraints in requirements of torch-test (0):
torch (==2.3.1+cu118) ; extra == "cuda" ; source=pytorch-cuda
torch (==2.3.1+cpu) ; extra != "cuda" ; source=pytorch-cpu
```
However it works with `--extras cuda` or with `re-resolve = false`.

But I think importantly the workaround is just to use exclusive extras, rather than the `cuda` extra that conflicts with the base. i.e. this works even with `re-resolve = true` with no extras, `cuda`, and `cpu` extras:

```toml
[tool.poetry]
package-mode = false

[project]
name = "torch-test"

cpu = [
    "torch (==2.3.1+cpu)",
]
cuda = [
    "torch (==2.3.1+cu118)",
]

[tool.poetry.dependencies]
python = "^3.10"
torch = [
    { markers = "extra == 'cpu' and extra != 'cuda'", source = "pytorch-cpu"},
    { markers = "extra != 'cpu' and extra == 'cuda'", source = "pytorch-cuda"},
 ]

[[tool.poetry.source]]
name = "pytorch-cpu"
url = "https://download.pytorch.org/whl/cpu"
priority = "explicit"

[[tool.poetry.source]]
name = "pytorch-cuda"
url = "https://download.pytorch.org/whl/cu118"
priority = "explicit"

[build-system]
requires = ["poetry-core>=2.0.0"]
```

This would also explain why all the tests succeed even with your new `config_installer_reresolve` auto-use fixture, as there is no "extra conflicts with non-extra" case tested.

Updated the example, let me know what you think.

radoering:
Thanks for the update @reesehyde!

I did not realize that the second example will work if I leave out the `extra != ...` in `project.optional-dependencies`.

I updated the docs with your updated examples (with some minor adaptions):
- `build-system` is not required with `package-mode = false`
- `tool.poetry.dependencies.python` can be replaced with `project.requires-python`
- I put `tool.poetry` after `project`

Further, I added a warning that the first example will only work with `installer.re-resolve` set to `false`.

reesehyde:
Awesome, thanks so so much for all the help over these past months @radoering! 🙏 

I may look more into whether it's possible to support `re-resolve = true` in a future PR but agree it's not worth holding this one up for it. All those changes look good to me, let me know whether we need anything else or this is is ready to merge

radoering (approved):
I have tried everything so far and skimmed over the code again. I think it is ready to merge now. Thanks @reesehyde for your persistent work on this topic.

PR #9558, opened 2024-07-16
default python requirement to >=

dimbleby:
I'm mostly of the view that there is absolutely no sign of python 4.0 happening in the foreseeable future, so it should make no difference to anyone whether they have this upper bound on their python requirement or not.

However, it continues to be a source of noise that this is a thing: and it is true that - in combination with its own solver - poetry's upper bounds becomes self-propagating.  Which, if nothing else, becomes an opinionated stance - on a matter which I think hardly deserves an opinion.

I expect that the transition may lead to a small bump in that noise, as resolution on new projects will now be more likely to hit conflicts when pulling in dependencies that have the cap.

Still, on the whole I think that no cap here is a better default and the world will be ever so slightly a better place if poetry makes the change.

(Upper bounds on regular dependencies are a whole different kettle of fish, out of scope here).

### DIFF
--- a/src/poetry/console/commands/init.py
+++ b/src/poetry/console/commands/init.py
@@ -173,7 +173,7 @@ The <c1>init</c1> command creates a basic <comment>pyproject.toml</> file in the
         if not python:
             config = Config.create()
             python = (
-                "^"
+                ">="
                 + EnvManager.get_python_version(
                     precision=2,
                     prefer_active_python=config.get("virtualenvs.prefer-active-python"),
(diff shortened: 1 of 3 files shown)


Secrus:
I like this (I would be all for removing the `^` marker completely at some point). Poetry 2.0 is a good place to make such a change.

radoering:
I did this change in #9135 anyway so I am fine with it. (It cannot hurt to keep it separate in this PR so we do not forget to mention it in the changelog.)

> I would be all for removing the ^ marker completely at some point). Poetry 2.0 is a good place to make such a change.

I am neutral to the first part but strongly disagree with the second sentence. Removing it will break poetry-core quite hard. (Harder than I like for 2.0) IMO, that is something that should be deprecated first for quite a while before finally removing it.

Secrus:
@radoering you misunderstood me. The Poetry 2.0 comment was only about the change in this PR.

Issue #9510, opened 2024-06-21
Missing documentation for `type = "file"` under `tool.poetry.scripts` 


RobertoRoos:
### Issue Kind

Missing documentation

### Existing Link

https://python-poetry.org/docs/pyproject/#scripts

### Description

It seems usage like
```
myscript = { reference = "some_binary.exe", type = "file" }
```
is missing from the documentation. Is this feature still experimental?

I posted about it in this StackOverflow answer: https://stackoverflow.com/questions/77893374/how-to-configure-pip-to-install-a-script-to-the-environment-bin-folder
I can't find where I first found this config option.

PR #9572, opened 2024-07-23
(Docs) Added type = "file" reference `[tool.poetry.scripts]`

RobertoRoos:
# Pull Request Check List

Resolves: #9510

### DIFF
--- a/docs/pyproject.md
+++ b/docs/pyproject.md
@@ -723,6 +723,7 @@ for more information on other keys and specifying version ranges.
 ### scripts
 
 **Deprecated**: Use `project.scripts` instead.
+Use `[tool.poetry.scripts]` only for scripts of type "file", because those are not supported in `[project.scripts]`.
 
 This section describes the scripts or executables that will be installed when installing the package
 
@@ -742,12 +743,12 @@ When a script is added or updated, run `poetry install` to make them available i
 my_executable = { reference = "some_binary.exe", type = "file" }
 ```
 
-With this configuration, Poetry will look for the referenced file in the active project and then copy it to the OS installation directory.
+This tells Poetry to include the specified file, relative to your project directory, in distribution builds. It will then be copied to the appropriate installation directory for your operating system when your package is installed.
 
 * On Windows the file is placed in the `Scripts/` directory.
 * On *nix system the file is placed in the `bin/` directory.
 
-## `extras`
+### `extras`
 
 **Deprecated**: Use `project.optional-dependencies` instead.
 
--- a/docs/pyproject.md
+++ b/docs/pyproject.md
@@ -737,7 +737,17 @@ Here, we will have the `my_package_cli` script installed which will execute the 
 When a script is added or updated, run `poetry install` to make them available in the project's virtualenv.
 {{% /note %}}
 
-### extras
+```toml
+[tool.poetry.scripts]
+my_executable = { reference = "some_binary.exe", type = "file" }
+```
+
+With this configuration, Poetry will look for the referenced file in the active project and then copy it to the OS installation directory.
+
+* On Windows the file is placed in the `Scripts/` directory.
+* On *nix system the file is placed in the `bin/` directory.
+
+## `extras`
 
 **Deprecated**: Use `project.optional-dependencies` instead.
 


docs/pyproject.md
```diff
@@ -375,6 +375,15 @@ Here, we will have the `my_package_cli` script installed which will execute the
...
 {{% /note %}}
 
+```toml
+[tool.poetry.scripts]
+my_executable = { reference = "some_binary.exe", type = "file" }
+```
+
+With this configuration, the file `some_binary.exe` is looked for in the directory of the active `pyproject.toml` and copied into the installation folder.
+On Windows, the `.exe` extension is added to the file and the file will be placed in the `Scripts/` directory.
+On Linux no extension is added and the destination is `bin/`.
```
nbaju1, line 385:
```suggestion
* On Windows, the `.exe` extension is added to the file, and is placed in the `Scripts/` directory.
* On Linux no extension is added, and is placed in the `bin/` directory.
```

For better readability, perhaps?

dimbleby: This is ungrammatical.  "The '.exe' extension... is placed in the 'Scripts/' directory" is not what you intend.

evilhamsterman:
@RobertoRoos the `type` option has more possible values than just `type="file"` there's also at least `type="console"` which at one point was at least mentioned in the documentation but was later removed even though it appears to still be a valid value https://github.com/orgs/python-poetry/discussions/8436

docs/pyproject.md
evilhamsterman: I agree the readability of this is not optimal. From the way it is currently described the full paragraph seems to imply that on Windows you'd end up with `some_file.exe.exe`

> With this configuration, the file `some_binary.exe` is looked for in the directory....On Windows, the `.exe` extension is added to the file

Perhaps something like this
> Poetry will look for the referenced file in the directory of the active project and then copied to the OS installation directory.
> * On Windows the `.exe` extension is added to the file and it is copied to the Python Scripts directory
> * On *nix systems the file will be copied to the environment `bin` directory

yeoffrey:
> @RobertoRoos the `type` option has more possible values than just `type="file"` there's also at least `type="console"` which at one point was at least mentioned in the documentation but was later removed even though it appears to still be a valid value https://github.com/orgs/python-poetry/discussions/8436

If this PR just wants to focus on the `file` option, I'm fine to add the `type` option in a follow up, which would close #9608

docs/pyproject.md
RobertoRoos: Updates the text. The `.exe` bit was weird indeed, it should read clearly now.

yeoffrey:
What is the status of this PR? Can we merge this?

docs/pyproject.md
```diff
@@ -737,7 +737,17 @@ Here, we will have the `my_package_cli` script installed which will execute the
 When a script is added or updated, run `poetry install` to make them available in the project's virtualenv.
 {{% /note %}}
 
-### extras
+```toml
```
abn, line 740: Can you also update the `deprecation` notice above to include something like "This section should only be used for scripts of type 'file'".

RobertoRoos: I've added that text.
But I don't actually understand what that means, has the `scripts` section been deprecated entirely?

docs/pyproject.md
```diff
@@ -737,7 +738,17 @@ Here, we will have the `my_package_cli` script installed which will execute the
...
+[tool.poetry.scripts]
+my_executable = { reference = "some_binary.exe", type = "file" }
+```
+
+With this configuration, Poetry will look for the referenced file in the active project and then copy it to the OS installation directory.
+
+* On Windows the file is placed in the `Scripts/` directory.
+* On *nix system the file is placed in the `bin/` directory.
+
+## `extras`
```
abn, line 751:
```suggestion
### `extras`
```

docs/pyproject.md
```diff
@@ -737,7 +738,17 @@ Here, we will have the `my_package_cli` script installed which will execute the
 When a script is added or updated, run `poetry install` to make them available in the project's virtualenv.
 {{% /note %}}
 
-### extras
+```toml
+[tool.poetry.scripts]
+my_executable = { reference = "some_binary.exe", type = "file" }
+```
+
+With this configuration, Poetry will look for the referenced file in the active project and then copy it to the OS installation directory.
```
abn, line 746:
```suggestion
This tells Poetry to include the specified file, relative to your project directory, in distribution builds. It will then be copied to the appropriate installation directory for your operating system when your package is installed. 
```

docs/pyproject.md
abn: Might need to reword it if it is not clear I guess :) 

PEP 621 introduced both cli and gui scripts in the spec, but lacks the `type = "file"` functionality. So, the prefered way of doing cli/gui scripts is the `project.scripts` section.

PR #9580, opened 2024-07-25
`--no-truncate` flag in show command

### LINKED ISSUE #9597: Get more consistent output from `poetry show`
### Issue Kind

Change in current behaviour

### Description

Currently, the output of `poetry show` depends on the width of the terminal of the caller. This means that some information might get truncated if the terminal is not wide enough. While this makes for a more more readable output for the user, it becomes less ideal when calling the command from a script, where the notion of terminal width does not make sense. 

In order to get a more consistent output from the `poetry show` command, there should be a `--no-truncate` flag added to the command. When the flag is set, the show command does not take the width of the terminal into account when deciding the level of detail to print.

### Impact

If `poetry show` is called from a script that processes the output downstream, it needs the command to give a consistent output.

### Workarounds

Currently, it is possible to set an arbitrarily large fake terminal width, in order to achieve to ensure nothing is truncated. E.g., setting environment variable `COLUMNS={large number}`. However, this is very much a hack and does not seem like the right solution.

ch1nq:
Fixes #9597

In order to get a more consistent output from the `poetry show` command, this PR introduces a `--no-truncate` flag to the command. When the flag is set the show command does not take the width of the terminal into account when deciding the level of detail to print.

# Pull Request Check List

### DIFF
--- a/docs/cli.md
+++ b/docs/cli.md
@@ -1061,6 +1061,7 @@ required by
 * `--outdated (-o)`: Show the latest version but only for packages that are outdated.
 * `--all (-a)`: Show all packages (even those not compatible with current system).
 * `--top-level (-T)`: Only show explicitly defined packages.
+* `--no-truncate`: Do not truncate the output based on the terminal width.
 
 {{% note %}}
 When `--only` is specified, `--with` and `--without` options are ignored.
--- a/src/poetry/console/commands/show.py
+++ b/src/poetry/console/commands/show.py
@@ -1,6 +1,6 @@
 from __future__ import annotations
 
-import math
+import sys
 
 from typing import TYPE_CHECKING
 from typing import ClassVar
@@ -237,7 +237,7 @@ lists all packages available."""
         show_all = self.option("all")
         show_top_level = self.option("top-level")
         width = (
-            math.inf
+            sys.maxsize
             if self.option("no-truncate")
             else shutil.get_terminal_size().columns
         )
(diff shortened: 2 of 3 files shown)


abn (commented):
Why not just set `width = math.inf if self.option("--no-truncate") else ...`? simplifies the code changes a lot.

abn:
Also, you might need to re-do the cli doc changes, since we reodered that doc.

ch1nq:
> @ch1nq is this something you'd like to continue working on?

Thanks for the review - I will apply the suggested changes

ch1nq:
@abn I have gone with `sys.maxsize` instead of `math.inf`, since we need it to be an int. This is because we use the value as a slice index later in the code.

docs/cli.md
```diff
@@ -210,6 +210,7 @@ If necessary the build process happens in an isolated environment.
 * `--local-version (-l)`: Add or replace a local version label to the build (deprecated).
 * `--output (-o)`: Set output directory for build artifacts. Default is `dist`.
 * `--config-settings=<key>=<value> (-c)`: Config settings to be passed to the build back-end. (multiple allowed)
+* `--no-truncate`: Do not truncate the output based on the terminal width.
```
abn, line 213: I think this is in the wrong place.

abn (changes requested):

docs/cli.md
ch1nq: Good catch! Must have been moved when I merged in `main`

abn (dismissed):

radoering (approved):

Issue #9161, opened 2024-03-14
Add test coverage for tests/helpers.py


abn:
As requested by @Secrus in https://github.com/python-poetry/poetry/pull/9149#discussion_r1525588514.

dimbleby:
I don't understand.  If there is code in tests/helpers.py that is not executed, just remove it already?

DarkTechPirate:
No what are even supposed to do ?

abn:
@dimbleby I think @Secrus is asking for test helpers to have functional test coverage. But yes, if we are not using a helper we should remove it. Although, I do not think that is the case right now.

dimbleby:
I dont buy it: who writes tests for their tests?  what then, shall we write tests for the tests for the tests too?

test code automatically gets coverage by being executed in the test suite, else delete it

abn:
I'll let @Secrus decide what they want to do with this one as I just created the issue based on his request. :wink:

Julynx:
> I dont buy it: who writes tests for their tests? what then, shall we write tests for the tests for the tests too?
> 
> test code automatically gets coverage by being executed in the test suite, else delete it

This is a horrible take. Tests should NEVER dictate which code is in your application by their coverage.

Imagine you have code that looks like this:

```python
if len(text) > MAX_LEN:
    raise ValueError("Text too long.")

do_something()
```

And you're getting coverage 90% in your tests because they are crap and are only sending text shorter than MAX_LEN, so your if branch is never triggered.

So, by your logic we should go "lets remove the code that isnt covered then". 

```python
do_something()
```

Boom, 100% coverage.

And now a user will send text longer than MAX_LEN, and your app will crash because it wasn't expecting it.

Coverage is meant for you to implement all possible cases and then ensure you test all of them in your tests. Its not something you can solve by removing those cases and hoping they never happen in production. 

TLDR. To get coverage to 100% you make decent tests instead of removing random important lines of code from your application.

dimbleby:
@Julynx you have completely misunderstood what is being discussed here

Julynx:
> @Julynx you have completely misunderstood what is being discussed here

Fair, I apologize for the misunderstading and I'm deleting my previous reply. Thanks for your patience.

j7an:
@python-poetry/triage main branch has merged `test_helpers.py`. This issue should be closed as resolved.

PR #9624, opened 2024-08-13
Improve test-coverage tests/helpers.py

dasschnee:
# Pull Request Check List

Related-to: #9161

### DIFF
--- a/tests/test_helpers.py
+++ b/tests/test_helpers.py
@@ -1,6 +1,8 @@
 from __future__ import annotations
 
-from tests.helpers import flatten_dict
+import os
+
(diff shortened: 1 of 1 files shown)


dimbleby:
per discussion in #9161, writing test for your tests still seems to me to be bizarre: I would prefer to see that issue simply rejected.

dasschnee:
@dimbleby, while I understand your point, I must say, that the `helpers` are not tests, they are dependencies for tests. Therefore, it makes total sense to me to test those dependencies. Even though they are dependencies for tests

radoering (approved):

PR #9631, opened 2024-08-15
feat: add strategy=lowest

TiansuYu:
# Pull Request Check List

Resolves: #3527

I have tried to follow the suggestion from [here](https://github.com/python-poetry/poetry/issues/3527#issuecomment-753531696) to come up with an solution. From what I have observed, it does not result in a resolution using the lowest version, but behaves the same as `strategy=highest`.

To reproduce, run `poetry install` then the dev version should be available at `.venv/bin/poetry`. 

Try lock with an experiement project (e.g. with only one dependency, say `requests`):
```shell
.venv/bin/poetry lock --strategy=lowest --directory <some experimental project dir>
```

Then check what the result is, by `poetry show --directory xyz` (either using the dev one or a prod one): 
```shell
certifi            2024.7.4 Python package for providing Mozilla's CA Bundle.
charset-normalizer 3.3.2    The Real First Universal Charset Detector. Open, modern...
idna               3.7      Internationalized Domain Names in Applications (IDNA)
requests           2.32.3   Python HTTP for Humans.
urllib3            2.2.2    HTTP library with thread-safe connection pooling, file ...
```

That means the native approach does not work. As a next step, I would have a deeper look at `version_solver.py` to understand if we could change the heuristics on version solving (not sure if I need to fully understand the Pubgrub algo though).

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "poetry"
-version = "1.9.0.dev0"
+version = "1.9.0.dev1"
 description = "Python dependency management and packaging made easy."
 authors = ["Sébastien Eustace <sebastien@eustace.io>"]
 maintainers = [
--- a/src/poetry/console/commands/lock.py
+++ b/src/poetry/console/commands/lock.py
@@ -1,12 +1,12 @@
 from __future__ import annotations
 
-from typing import TYPE_CHECKING
 from typing import ClassVar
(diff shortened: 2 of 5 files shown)


Secrus:
Since you marked it as `WIP`, I converted the PR to draft.

TiansuYu:
Currently, I have very limited capacity on this, so dont expect this to be moving forward any time soon.

PR #9606, opened 2024-08-06
Apply assorted repo-review suggestions

DimitriPapadopoulos:
Apply some of these suggestions:
https://learn.scientific-python.org/development/guides/repo-review/?repo=python-poetry%2Fpoetry&branch=main

# Pull Request Check List

- [ ] Added **tests** for changed code.
- [ ] Updated **documentation** for changed code.

PR #9633, opened 2024-08-16
remove sp-repo-review pre-commit hook because it does not work on Windows

radoering:
Remove the hook as per https://github.com/python-poetry/poetry/pull/9606#issuecomment-2291676090 and https://github.com/python-poetry/poetry/pull/9606#issuecomment-2292278736 because it does not work on Windows.

See https://github.com/scientific-python/cookie/issues/473

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -31,8 +31,3 @@ repos:
     hooks:
       - id: ruff
       - id: ruff-format
-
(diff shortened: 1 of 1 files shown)


Secrus (approved):

henryiii:
Latest sp-repo-review (technically repo-review, but due to pre-commit's caching, I'd recommend updating sp-repo-review) no longer requires PYTHONUTF8 to be set on Windows!

Issue #8662, opened 2023-11-13
No file/folder found for package 


Miguerfi:
- [X] I have searched the [issues](https://github.com/python-poetry/poetry/issues) of this repo and believe that this is not a duplicate.

## Issue
```Installing the current project: api_meucx (1.2)
The current project could not be installed: No file/folder found for package api-meucx
If you do not want to install the current project use --no-root
```

### how to reproduce:
``` poetry install ```

the error may be here:
![image](https://github.com/python-poetry/poetry/assets/91486755/74414306-3fae-49bc-8a26-e9e54a792f99)

### what i tried to do:
- [X] change the project name in pyproject.toml
- [X] reinstall poetry
- [x] reinstall project - delete and git clone again.
- [X] change path name.

**Comment by radoering** [comment]
And do you have an `api_meucx` folder with Python files or an `api_meucx.py` in your project folder?

**Comment by Miguerfi** [comment]
no, its a django project.

**Comment by Miguerfi** [comment]
and the  `api_meucx` is the base dir. inside have the poetry filess

**Comment by Miguerfi** [comment]
hmhmh None?

**Comment by dimbleby** [comment]
then please close the issue report

**Comment by lexhl** [comment]
I don't understand what is going on here; I am sure my project used to work but now it doesn't. What does it mean - that I now have to have my source files in a sub-directory with the same name as my project?

**Comment by radoering** [comment]
That's just a warning. If your project used to work, it will probably still work despite of this warning. It means that you either have to make your project packageable or use `--no-root`.

Or ignore the warning for now and wait for #8650.

**Comment by ughstudios** [comment]
this is a weird error happening for me as well.

**Comment by Miguerfi** [comment]
hi guys i've solved the problem renaming the project name in `.toml_` to the project core folder. 
dont use uppercase to the project folder name. and dont use special symbols like "-" or "_"

**Comment by mathewvaughan** [comment]
Facing a similar issue - just want to check my understanding. Here's my minimum repro:

- I create a directory called 'test', and I cd into it.

I run `poetry init` and just accept all the defaults.

I run `poetry install`. 

I see the following: 

```
Installing the current project: test (0.1.0)
The current project could not be installed: No file/folder found for package test
If you do not want to install the current project use --no-root
```

I've not had this warning before - but to my understanding my root is the parent directory 'test' and this isn't being respected by the `poetry install` command?

kfsone:
How to recreate this in a most confusing way.

[...]

PR #9651, opened 2024-08-25
Improve error message to highlight solution to user configuration err…

beattyml1:
…or highlighted in issue 8662

# Pull Request Check List

Resolves: #8662

Provides additional help text error message for `Warning: The current project could not be installed`. I know the issue above is closed but I think this is also very helpful for helping people to understand issue. If you don't agree I won't raise a fuss just thought I'd offer to add add the help text to the error message that would have helped me.

- [N/A] Added **tests** for changed code. 
  - Checked for existing code for tests related to this error message
  - No control flow to change just a error message text 
  - Tests still pass on local


src/poetry/console/commands/install.py
```diff
@@ -201,6 +201,8 @@ def handle(self) -> int:
                 "If you want to use Poetry only for dependency management"
                 " but not for packaging, you can disable package mode by setting"
                 " <c1>package-mode = false</> in your pyproject.toml file.\n"
+                "If you did intend to install the current project please make sure"
+                " you set the packages setting in your pyproject.toml file.\n"
```
radoering, line 205:
```suggestion
                "If you did intend to install the current project you may need"
                " to set `packages` in your pyproject.toml file.\n"
```

I agree it makes sense to provide additional information. I would only rephrase it slightly because setting `packages` is only required if your package name does not match your project name.

radoering (approved):

Issue #6958, opened 2022-11-03
`NotGitRepository` error when installing multiple packages from one git repository


gnuletik:
- **Poetry version**: 1.2.2
- **Python version**: 3.10.8
- **OS version and name**: macOS 13.0
- **pyproject.toml**: https://gist.github.com/gnuletik/8d876426a36b9bfefee4327823c1459b

- [x] I am on the [latest](https://github.com/python-poetry/poetry/releases/latest) stable Poetry version, installed using a recommended method.
- [x] I have searched the [issues](https://github.com/python-poetry/poetry/issues) of this repo and believe that this is not a duplicate.
- [x] I have consulted the [FAQ](https://python-poetry.org/docs/faq/) and [blog](https://python-poetry.org/blog/) for any relevant entries or release notes.
- [x] If an exception occurs when executing a command, I executed it again in debug mode (`-vvv` option) and have included the output below.

## Issue

It seems that a race condition occurs when installing two packages:
- from the same git repository
- with a different subdirectory
- on a non-default git branch

Repro:
```bash
cd /tmp
git clone https://github.com/gnuletik/poetry-lib-monorepo-issue
cd poetry-lib-monorepo-issue
poetry install
```

It fails with 
```
Package operations: 2 installs, 0 updates, 0 removals

  • Installing package1 (0.1.0 c6f487b): Failed

  NotGitRepository

  No git repository was found at /private/tmp/test-poetry/.venv/src/poetry-multipackages-example

  at /opt/homebrew/Cellar/poetry/1.2.2/libexec/lib/python3.10/site-packages/dulwich/repo.py:1090 in __init__
      1086│             elif (os.path.isdir(os.path.join(root, OBJECTDIR))
      1087│                     and os.path.isdir(os.path.join(root, REFSDIR))):
      1088│                 bare = True
      1089│             else:
    → 1090│                 raise NotGitRepository(
      1091│                     "No git repository was found at %(path)s" % dict(path=root)
      1092│                 )
      1093│
      1094│         self.bare = bare

The following error occurred when trying to handle this error:
```

NB: output of `poetry install -vvv` can be found here: https://gist.github.com/gnuletik/ddcb05ff3467f022f9d3540f379763df

Please note that subsequent calls **may** succeed but a fresh install  (after a `poetry env remove --all`) always fails.

pneb:
[...]

PR #9658, opened 2024-08-30
Avoid race conditions when installing multiple dependencies from the same git repository.

gustavgransbo:
Multiple installs from the same git repository causes a race condition when the repository is cloned.
This change only allows one operation per repository to be executed by the parallel workers.
Any extra operation on the same git repository will be performed serially.
Since the git repository is cached after the first operation, this is fast.

Building on the [setup](https://github.com/python-poetry/poetry/issues/6958#issuecomment-2229914607) of @JonathanRayner, I was able to consistently reproduce the error in [this repository](https://github.com/gustavgransbo/some_other_repo).

Thank you @JonathanRayner for the hard work!

# Pull Request Check List

Resolves: #6958 

The race condition occurs inside [poetry.vcs.git.Git.clone](https://github.com/python-poetry/poetry/blob/5d19ab0d8e9b734e3452c6e1210a7f2501709397/src/poetry/vcs/git/backend.py#L412), this method is mocked in all tests: [conftest.py](https://github.com/python-poetry/poetry/blob/5d19ab0d8e9b734e3452c6e1210a7f2501709397/tests/conftest.py#L328).
This makes the actual race condition hard to test.

Also, since the race condition required quite a bit of external setup, I think it could be challenging to add a good test for this.

This change should not require any new documentation.

### DIFF
--- a/src/poetry/installation/executor.py
+++ b/src/poetry/installation/executor.py
@@ -160,13 +160,11 @@ class Executor:
                     operation.package.develop
                     and operation.package.source_type in {"directory", "git"}
                 )
-
+                # Skipped operations are safe to execute in parallel
                 if operation.skipped:
-                    # Skipped operations are safe to execute in parallel
-                    tasks.append(
-                        self._executor.submit(self._execute_operation, operation)
-                    )
-                elif is_parallel_unsafe:
+                    is_parallel_unsafe = False
+
+                if is_parallel_unsafe:
                     serial_operations.append(operation)
                 elif operation.package.source_type == "git":
                     # Git operations on the same repository should be executed serially
--- a/src/poetry/installation/executor.py
+++ b/src/poetry/installation/executor.py
@@ -176,15 +176,14 @@ class Executor:
                         self._executor.submit(self._execute_operation, operation)
                     )
 
+            def _serialize(
+                repository_serial_operations: list[Operation],
+            ) -> None:
+                for operation in repository_serial_operations:
+                    self._execute_operation(operation)
+
             # For each git repository, execute all operations serially
             for repository_git_operations in serial_git_operations.values():
-
-                def _serialize(
-                    repository_serial_operations: list[Operation],
-                ) -> None:
-                    for operation in repository_serial_operations:
-                        self._execute_operation(operation)
-
                 tasks.append(
                     self._executor.submit(
                         _serialize,
--- a/src/poetry/installation/executor.py
+++ b/src/poetry/installation/executor.py
@@ -6,6 +6,7 @@ import itertools
 import json
 import threading
 
+from collections import defaultdict
 from concurrent.futures import ThreadPoolExecutor
 from concurrent.futures import wait
 from pathlib import Path
@@ -144,23 +145,10 @@ class Executor:
         for _, group in groups:
             tasks = []
             serial_operations = []
-            git_repositories_to_clone = set()
+            serial_git_operations = defaultdict(list)
             for operation in group:
                 if self._shutdown:
                     break
-
-                # It's not safe to clone the same git repository multiple times in parallel
-                if operation.package.source_type == "git" and not operation.skipped:
-                    # If this is the first operation cloning this repository,
-                    # allow it to be executed in parallel
-                    if operation.package.source_url not in git_repositories_to_clone:
-                        git_repositories_to_clone.add(operation.package.source_url)
-
-                    # If another operation is already cloning this repository,
-                    # it must be executed serially
-                    else:
-                        serial_operations.append(operation)
-                        continue
 
                 # Some operations are unsafe, we must execute them serially in a group
                 # https://github.com/python-poetry/poetry/issues/3086
@@ -172,11 +160,41 @@ class Executor:
                     operation.package.develop
                     and operation.package.source_type in {"directory", "git"}
                 )
-                if not operation.skipped and is_parallel_unsafe:
-                    serial_operations.append(operation)
-                    continue
 
-                tasks.append(self._executor.submit(self._execute_operation, operation))
+                if operation.skipped:
+                    # Skipped operations are safe to execute in parallel
+                    tasks.append(
+                        self._executor.submit(self._execute_operation, operation)
+                    )
+                elif is_parallel_unsafe:
+                    serial_operations.append(operation)
+                elif operation.package.source_type == "git":
+                    # Git operations on the same repository should be executed serially
+                    serial_git_operations[operation.package.source_url].append(
+                        operation
+                    )
+                else:
+                    tasks.append(
+                        self._executor.submit(self._execute_operation, operation)
+                    )
+
+            # For each git repository, execute all operations serially
+            for repository_git_operations in serial_git_operations.values():
+
+                def _serialize(
+                    repository_serial_operations: list[Operation],
+                ) -> None:
+                    for operation in repository_serial_operations:
+                        self._execute_operation(operation)
+
+                tasks.append(
+                    self._executor.submit(
+                        functools.partial(
+                            _serialize,
+                            repository_serial_operations=repository_git_operations,
+                        )
+                    )
+                )
 
             try:
                 wait(tasks)
--- a/src/poetry/installation/executor.py
+++ b/src/poetry/installation/executor.py
@@ -178,10 +178,17 @@ class Executor:
 
             # For each git repository, execute all operations serially
             for repository_git_operations in serial_git_operations.values():
+
+                def _serialize(
+                    repository_serial_operations: list[Operation],
+                ) -> None:
+                    for operation in repository_serial_operations:
+                        self._execute_operation(operation)
+
                 tasks.append(
                     self._executor.submit(
-                        self._serialize_operations,
-                        serial_operations=repository_git_operations,
+                        _serialize,
+                        repository_serial_operations=repository_git_operations,
                     )
                 )
 
@@ -876,11 +883,3 @@ class Executor:
             archive_info["hashes"] = {algorithm: value}
 
         return archive_info
-
-    def _serialize_operations(
-        self,
-        serial_operations: list[Operation],
-    ) -> None:
-        """Execute operations serialy"""
-        for operation in serial_operations:
-            self._execute_operation(operation)
--- a/src/poetry/installation/executor.py
+++ b/src/poetry/installation/executor.py
@@ -139,13 +139,12 @@ class Executor:
         self._sections = {}
         self._yanked_warnings = []
 
-        git_repositories_to_clone = set()
-
         # We group operations by priority
         groups = itertools.groupby(operations, key=lambda o: -o.priority)
         for _, group in groups:
             tasks = []
             serial_operations = []
+            git_repositories_to_clone = set()
             for operation in group:
                 if self._shutdown:
                     break
--- a/src/poetry/installation/executor.py
+++ b/src/poetry/installation/executor.py
@@ -139,6 +139,8 @@ class Executor:
         self._sections = {}
         self._yanked_warnings = []
 
+        git_repositories_to_clone = set()
+
         # We group operations by priority
         groups = itertools.groupby(operations, key=lambda o: -o.priority)
         for _, group in groups:
@@ -147,6 +149,19 @@ class Executor:
             for operation in group:
                 if self._shutdown:
                     break
+
+                # It's not safe to clone the same git repository multiple times in parallel
+                if operation.package.source_type == "git" and not operation.skipped:
+                    # If this is the first operation cloning this repository,
+                    # allow it to be executed in parallel
+                    if operation.package.source_url not in git_repositories_to_clone:
+                        git_repositories_to_clone.add(operation.package.source_url)
+
+                    # If another operation is already cloning this repository,
+                    # it must be executed serially
+                    else:
+                        serial_operations.append(operation)
+                        continue
 
                 # Some operations are unsafe, we must execute them serially in a group
                 # https://github.com/python-poetry/poetry/issues/3086
--- a/src/poetry/installation/executor.py
+++ b/src/poetry/installation/executor.py
@@ -189,10 +189,8 @@ class Executor:
 
                 tasks.append(
                     self._executor.submit(
-                        functools.partial(
-                            _serialize,
-                            repository_serial_operations=repository_git_operations,
-                        )
+                        _serialize,
+                        repository_serial_operations=repository_git_operations,
                     )
                 )
 
--- a/src/poetry/installation/executor.py
+++ b/src/poetry/installation/executor.py
@@ -178,17 +178,10 @@ class Executor:
 
             # For each git repository, execute all operations serially
             for repository_git_operations in serial_git_operations.values():
-
-                def _serialize(
-                    repository_serial_operations: list[Operation],
-                ) -> None:
-                    for operation in repository_serial_operations:
-                        self._execute_operation(operation)
-
                 tasks.append(
                     self._executor.submit(
-                        _serialize,
-                        repository_serial_operations=repository_git_operations,
+                        self._serialize_operations,
+                        serial_operations=repository_git_operations,
                     )
                 )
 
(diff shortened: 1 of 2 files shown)


seeker25:
Thank you so much

radoering (commented):
LGTM. Just a minor nitpick.

I thought a bit about how to write a unit test but even with mocking I could not come up with a good test. Maybe, we can just add kind of a smoke test like `test_executor_should_append_subdirectory_for_git` but with two packages from the same git repo and check that both are still installed. At least, the order should be fix now, since they are installed serially.

seeker25:
A workaround for now:

`poetry config installer.max-workers 1`

src/poetry/installation/executor.py
```diff
@@ -148,6 +150,19 @@ def execute(self, operations: list[Operation]) -> int:
                 if self._shutdown:
                     break
 
+                # It's not safe to clone the same git repository multiple times in parallel
+                if operation.package.source_type == "git" and not operation.skipped:
+                    # If this is the first operation cloning this repository,
+                    # allow it to be executed in parallel
+                    if operation.package.source_url not in git_repositories_to_clone:
```
seeker25, line 157: Does this work on forked vs non-forked?
https://github.com/repo1/code
https://github.com/repo2/code

src/poetry/installation/executor.py
```diff
@@ -148,6 +150,19 @@ def execute(self, operations: list[Operation]) -> int:
...
+                        git_repositories_to_clone.add(operation.package.source_url)
+
+                    # If another operation is already cloning this repository,
+                    # it must be executed serially
+                    else:
+                        serial_operations.append(operation)
+                        continue
+
```
TheSven73, line 165: This is a really cool fix! Two nits, though:
- it makes the loop logic much harder to read / understand, conceptually
- the serialization seems too strict: I suspect it's sufficient to serialize git ops on the same repo

I wonder if this could be made much simpler by bucketizing the git ops, and submitting them in a serialized fashion? e.g.
```python
serial_operations = []
git_serial_ops = collections.defaultdict(list)
```
Then right before `tasks.append`:
```python
if not operation.skipped and operation.package.source_type == "git":
    git_serial_ops[operation.package.source_url].append(operation)
    continue
```
Then right before the wait, submit the serialized ops:
```python
# git operations on the same repo url should execute serially
for ops in git_serial_ops.values():
    def serialize() -> None:
        for op in ops:
            self._execute_operation(op)
    tasks.append(self._executor.submit(serialize))
```

src/poetry/installation/executor.py
gustavgransbo: Good catch, it does not! The race condition stems from having multiple installs that try to clone to the same local directory. The name for this directory is created [here](https://github.com/python-poetry/poetry/blob/7181557d01cc8c6e4eb15014a1e736fcfaea3a15/src/poetry/vcs/git/backend.py#L426).

In your example, both repositories would get the same `target`: `<source_root>/code`.

But as you point out, I would not catch this in my checks for identical source_urls, since the forks have different urls.

src/poetry/installation/executor.py
```diff
@@ -139,6 +139,8 @@ def execute(self, operations: list[Operation]) -> int:
         self._sections = {}
         self._yanked_warnings = []
 
+        git_repositories_to_clone = set()
```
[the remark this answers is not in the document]
gustavgransbo: Good call, I'll move it!

src/poetry/installation/executor.py
gustavgransbo: I like this suggestion! I think the performance win is marginal, since the first operation on each repository is already done in parallel, **but** it turns two levels of conditionals into one, and to me this makes the loop easier to read.

Should I update this PR with the new proposed logic?

gustavgransbo: I created a separate branch where I implemented this logic to make it easier to tell what it would look like: https://github.com/python-poetry/poetry/commit/8d15e5df192c13b8747e35d87c757e5bc6252be4

Tell me if you would like me to merge that branch into this one.

TheSven73: Glad you liked it ! However I am just a poetry community member, and certainly not a maintainer. It's up to people such as @abn to decide which approach is preferable. I believe he's also the original author of the code in question.

Regarding the https://github.com/python-poetry/poetry/commit/8d15e5df192c13b8747e35d87c757e5bc6252be4 branch: nice, I like how you've refactored the operation logic. I don't think you need `functools.partial`: you can just define `_serialize` as a closure, which will capture and store `self` and `ops` in the function itself, like so:
```python
for ops in serial_git_operations.values():

  def _serialize() -> None:
    for op in ops:
      self._execute_operation(op)

  tasks.append(self._executor.submit(_serialize))
```

gustavgransbo: I did write it like that at first, but was sternly warned by Ruff to not use a [loop variable in a function](https://docs.astral.sh/ruff/rules/function-uses-loop-variable/)!

TheSven73: Nice gotcha! That just shows the value of linter tools to keep us on the python straight-and-narrow :)

How about this, would it be simpler than using `functools.partial` ?
```python
for ops in serial_git_operations.values():

  def _serialize(ops=ops) -> None:
    for op in ops:
      self._execute_operation(op)

  tasks.append(self._executor.submit(_serialize))
```

radoering: > Tell me if you would like me to merge that branch into this one.

I took a quick glance and do not have any clear preference as to which approach is easier to understand. Since the second approach is functionally better in certain situations we should probably choose it.

src/poetry/installation/executor.py
TheSven73: Or 3. fix the race condition ?

From the little I understand, it appears that the root cause is concurrent execution of `poetry.vcs.git.Git.clone()` in the same physical directory. Neither the `git` executable, nor dulwich support concurrent execution in the same directory. Does this sound right?

What if we prevent concurrent git execution in the same directory? We could use `threading.Lock`.

In `vcs/git/backend.py` this would look like:
```python
from collections import defaultdict
from threading import Lock

_clone_lock: dict[Path, Lock] = defaultdict(Lock)

@classmethod
def clone(
    cls,
    url: str,
    name: str | None = None,
    branch: str | None = None,
    tag: str | None = None,
    revision: str | None = None,
    source_root: Path | None = None,
    clean: bool = False,
) -> Repo:
    source_root = source_root or cls.get_default_source_root()
    source_root.mkdir(parents=True, exist_ok=True)

    name = name or cls.get_name_from_source_url(url=url)
    target = source_root / name
    refspec = GitRefSpec(branch=branch, revision=revision, tag=tag)

    # Prohibit concurrent clones in the same target Path
    with cls._clone_lock[target]:
      # rest of clone follows, only the indentation changes
```
This will deadlock if `clone()` calls back into `clone()` _using the same directory_. And although recursive calls do happen via submodules, it doesn't look like they use the same directory.

I tried this out on the sample repo in https://github.com/python-poetry/poetry/issues/6958 and it seems to fix the issue, but I'm not familiar enough with the poetry code to know if this would be a viable approach.

src/poetry/installation/executor.py
gustavgransbo: I switched to this solution.

However I did not remove the `functools.partial`, as the solutions are equivalent, and I personally have nothing against using `functools`. If someone feels strongly about this I will of course update it.

src/poetry/installation/executor.py
gustavgransbo: Nice! Fixing the problem at it's root would of course be great.

An unfortunate functional drawback would be that when multiple installs from the same repository are executed in parallel, they would all execute past the [cache lookup](https://github.com/python-poetry/poetry/blob/258ac696efb4c4c6abfc3352036a1bd1f7135d03/src/poetry/installation/executor.py#L620), and then proceed to wait right before the clone.

I have not dug deep into this, but it seems like this would cause the repository to be downloaded multiple times instead of just once. With the solution in this PR, it's only downloaded once.

TheSven73: Good point. My suggestion in https://github.com/python-poetry/poetry/pull/9658#discussion_r1751154938 could potentially work, but would cause a performance hit for sure.

So how about this: git operations with the same url name (from `Git.get_name_from_source_url()`) might conflict, because they're cloned in `<source_root>/src/<url name>`. So we serialize all git operations which share an url name.

For example, on top of poetry main:
```diff
diff --git a/src/poetry/installation/executor.py b/src/poetry/installation/executor.py
index 196cdd5a..ff439d50 100644
--- a/src/poetry/installation/executor.py
+++ b/src/poetry/installation/executor.py
@@ -32,6 +32,7 @@ from poetry.utils.helpers import pluralize
 from poetry.utils.helpers import remove_directory
 from poetry.utils.isolated_build import IsolatedBuildError
 from poetry.utils.isolated_build import IsolatedBuildInstallError
+from poetry.vcs.git import Git
 
 
 if TYPE_CHECKING:
@@ -144,6 +145,7 @@ class Executor:
         for _, group in groups:
             tasks = []
             serial_operations = []
+            git_operations = []
             for operation in group:
                 if self._shutdown:
                     break
@@ -162,8 +164,32 @@ class Executor:
                     serial_operations.append(operation)
                     continue
 
+                # Defer git operations, they might need to be serialized
+                if operation.package.source_type == "git" and not operation.skipped:
+                    git_operations.append(operation)
+                    continue
+
                 tasks.append(self._executor.submit(self._execute_operation, operation))
 
+            def url_name_from_operation(operation: Operation) -> str | None:
+                if url := operation.package.repository_url:
+                    return Git.get_name_from_source_url(url)
+                return None
+
+            # git operations on the same url name end up being cloned in the same directory,
+            # so these must execute serially or they'll interfere with each other.
+            for _, git_ops_group in itertools.groupby(
+                git_operations,
+                key=url_name_from_operation,
+            ):
+                ops = list(git_ops_group)
+
+                def _serialize(ops: list[Operation] = ops) -> None:
+                    for op in ops:
+                        self._execute_operation(op)
+
+                tasks.append(self._executor.submit(_serialize))
+
             try:
                 wait(tasks)
```

src/poetry/installation/executor.py
TheSven73: Is `partial` really needed in this case? I.e. is your code:
```python
tasks.append(
    self._executor.submit(
        functools.partial(
            _serialize,
            repository_serial_operations=repository_git_operations,
        )
    )
)
```
equivalent to
```python
tasks.append(self._executor.submit(_serialize, repository_git_operations))
```
?

src/poetry/installation/executor.py
gustavgransbo: I'm leaning towards ensuring that forked repositories (actually any repository that happen to have the same name) are not cloned into the same target directory. I'd have to examine if there are any implicit dependencies on the name of the target folder though.

If that route seems acceptable, I would also consider implementing it in a separate PR, since it could be viewed as a separate fix.

What do you think?

radoering: A separate PR sounds right to me.

I am wondering what the use case behind this issue is? If you have two forks with the same name the two packages that are built from the forks are likely to have the same name, right? If you have two packages with the same name from different repos you will get a conflict because you can only install one package with a specific name into an environment.

radoering:
@gustavgransbo @TheSven73 Just to be sure before I take another detailed look and potentially merge it: If we outsource https://github.com/python-poetry/poetry/pull/9658#discussion_r1745925411 into a separate issue/PR, this PR is ready from your end, isn't it?

gustavgransbo:
> @gustavgransbo @TheSven73 Just to be sure before I take another detailed look and potentially merge it: If we outsource [#9658 (comment)](https://github.com/python-poetry/poetry/pull/9658#discussion_r1745925411) into a separate issue/PR, this PR is ready from your end, isn't it?

Yes, from my perspective it's ready to be merged!

src/poetry/installation/executor.py
```diff
@@ -158,11 +160,41 @@ def execute(self, operations: list[Operation]) -> int:
...
+                    repository_serial_operations: list[Operation],
+                ) -> None:
+                    for operation in repository_serial_operations:
+                        self._execute_operation(operation)
+
+                tasks.append(
+                    self._executor.submit(
+                        functools.partial(
+                            _serialize,
+                            repository_serial_operations=repository_git_operations,
```
TheSven73, line 194: Is `functools.partial` an unnecessary complication here? Isn't this code just equivalent to
```python
tasks.append(self.executor.submit(_serialize, repository_git_operations))
```
?

gustavgransbo: You are right! Fixed it now.

src/poetry/installation/executor.py
```diff
@@ -158,11 +160,39 @@ def execute(self, operations: list[Operation]) -> int:
...
-                    continue
+                elif operation.package.source_type == "git":
+                    # Git operations on the same repository should be executed serially
+                    serial_git_operations[operation.package.source_url].append(
+                        operation
+                    )
+                else:
+                    tasks.append(
+                        self._executor.submit(self._execute_operation, operation)
+                    )
```
TheSven73, line 179: Something feels weird here - the first `if` and final `else` clauses have identical code:
```python
if operation.skipped:
  tasks.append(self._executor.submit())
<snip bunch of elif clauses>
else:
  tasks.append(self._executor.submit())
```
Maybe it would make sense to fold the `operation.skipped` into `is_parallel_unsafe`? Like so, simplest I can think of:
```python
is_parallel_unsafe = operation.job_type == "uninstall" or (
    operation.package.develop
    and operation.package.source_type in {"directory", "git"}
)
# Skipped operations are safe to execute in parallel
if operation.skipped:
  is_parallel_unsafe = False
```
Then the if/else becomes simpler:
```python
if is_parallel_unsafe:
  # run serially
elif operation.package.source_type == "git":
  # Git operations on the same repository should be executed serially
else:
  # execute in parallel
```

gustavgransbo: Agreed, nice suggestion!

src/poetry/installation/executor.py
```diff
@@ -158,11 +160,37 @@ def execute(self, operations: list[Operation]) -> int:
...
-                tasks.append(self._executor.submit(self._execute_operation, operation))
+            # For each git repository, execute all operations serially
+            for repository_git_operations in serial_git_operations.values():
+
+                def _serialize(
+                    repository_serial_operations: list[Operation],
+                ) -> None:
+                    for operation in repository_serial_operations:
+                        self._execute_operation(operation)
+
```
TheSven73, line 187: Minor nit: given that you chose to pass the serial ops as a parameter, can `_serialize` be moved out of the loop?

TheSven73 (approved):
lgtm
needs owner / maintainer approval

src/poetry/installation/executor.py
TheSven73: I'm sorry for being unclear in the above comment. I meant: move `_serialize` out of the loop, but still inside the `execute` function.

I am not convinced that making `_serialize` a member function provides value. It's a helper function very closely coupled with the parallel execution happening in `execute`.

src/poetry/installation/executor.py
```diff
@@ -855,3 +876,11 @@ def _get_archive_info(self, package: Package) -> dict[str, Any]:
...
 
         return archive_info
+
+    def _serialize_operations(
+        self,
+        serial_operations: list[Operation],
+    ) -> None:
+        """Execute operations serialy"""
+        for operation in serial_operations:
+            self._execute_operation(operation)
```
TheSven73, line 886: I'm sorry for being unclear in https://github.com/python-poetry/poetry/pull/9658#discussion_r1776862695. I meant: move `_serialize` out of the loop, but still inside the `execute` function.

I am not convinced that making `_serialize` a member function provides value. It's a helper function very closely coupled with the parallel execution happening in `execute`.

gustavgransbo: Ah, I misinterpreted you. Fixed!

TheSven73 (approved):
lgtm
needs owner / maintainer approval

radoering (approved):
Thanks for the profound review @TheSven73 and thanks for keeping up with it @gustavgransbo .

TheSven73:
Apologies for the time-fragmented review. And thanks so much @gustavgransbo for tackling this hard problem !!

gustavgransbo:
Thanks a lot for the thorough review @TheSven73! It’s a privilege to receive such well thought out suggestions, and the resulting solution was much better for it.

Issue #9249, opened 2024-03-29
Add truststore support to use system certificate store


nullSoup:
### Issue Kind

Brand new capability

### Description

Poetry should have the option or default to pulling certificates from the system certificate store like pip now does with use-feature=trustore. Without this, Poetry is extremely difficult to use in corporate environments with IT systems where TLS inspection breaks the use of the default certificates.

This was already mentioned by @Secrus in [https://github.com/orgs/python-poetry/discussions/8729](url) as in his plans, but I wanted to raise the idea’s profile by making a Feature Request after spending half a day struggling with the problem.

### Impact

Adding this would make Poetry go from essentially broken behind corporate IT systems to just working for these users.

### Workarounds

It is possible to deactivate certificate verification at the cost of extra project setup step, nasty warnings, and confused users. Also theoretically, a manual cert file file can be exported and referenced with the REQUESTS_CA_BUNDLE environment variable. Both solutions are costly and painful. Native Python ssl already defaults to using the system certificate store, and most users would benefit by having something that just works.

dimbleby:
> ... a formal Feature Request ...

it is possible you misunderstand how things get done in poetry.  There is no committee evaluating feature requests and scheduling an army of developers to do work: things are implemented because someone who wants them shows up and implements them.

if that is you - be the change you want to see! - I expect a pull request would be welcome

Secrus:
This is still on my todo list, it's just a matter of priorities and other things having to be done before this could happen. Also a reminder, that `truststore` is only available on Python 3.10+, so that feature would cover only 60% of our supported Python versions (which lowers the priority of that feature on my list).

nullSoup:
Appreciate the response. Had no intention to act entitled about feature requests. I would contribute a PR if I had the skills, but unfortunately I couldn’t even figure out how to properly dump the correct certificates from my Windows system certificate store to take advantage of the REQUESTS_CA_STORE environment variable workaround, so coding the feature correctly is probably not something it makes sense to take on personally. I’m just an ordinary fan of the poetry tool for managing my dependencies who got caught with hours of confusion when IT quietly added TLS inspection causing everything to mysteriously break.

It’s a good point about truststore requiring Python 3.10+. I don’t have a technical understanding why the best fix adopted by pip is this extra Python 3.10+ truststore package. I just figured that since the —use-feature=truststore solution was deployed (and worked for me) in pip, it would make sense in poetry as well.

[...]

PR #9685, opened 2024-09-12
feat: Support OS truststore for the TLS certificate verification

Ousret:
This commit add support for loading the OS truststore root certificates in addition to the "legacy" certifi bundle. It will come in handy for every user behind a company proxy or just using a private PyPI warehouse.

The underlying library is called "wassima" and works without touching any Python internals (aka. SSLContext monkeypatch) and is compatible from Python 3.7 and greater. It's about a year old, and count approx 100k pull per month (very modest).

Its main strength is to be extremely simple. I remain available to answer any concerns.

# Pull Request Check List

Resolves: https://github.com/python-poetry/poetry/issues/9249


dimbleby:
for comparison

- pdm uses `truststore` and requires it explicitly to be injected as `pdm self add truststore`
- pip uses `truststore` but only if you provide the command line flag asking for it
- uv also requires a flag to use the system trust store (asserting that on macOS, reading the system trust store "incurs a significant delay")

notable that wassima is wrapping compiled code which makes it less portable - though I see that there is a good range of wheels available - and perhaps also a little harder to trust.

I would think that the barriers to merging this would be:

- (nothing personal!  but it needs asking for this sort of thing) do we trust wassima?
- what platforms does wassima not support, is that going to be a problem?
- should this be always on?

perhaps it is something that could be delivered as a plugin, so the user experience would be somewhat as it is in pdm: users would be expected to install the plugin in their poetry environment if they want to use it.

that way this project could duck those difficult questions for a little longer

Secrus:
To be honest, I don't like this. Seems like the library is quite fresh and wasn't battle-tested before. I don't feel like Poetry is a project that wants to be a testing ground for libraries. I would much rather see us use `truststore`, enabling it basing on the underlying Python version (`truststore` only supports Python 3.10+). Also, I don't think we need a library that wraps a compiled code unless there is no other viable option (in this case, there is `truststore`).
So, from my side, the question is: @Ousret, are you willing to work on this feature, but basing it on `truststore` instead of your own library?

Ousret:
Hello,

@dimbleby 

> what platforms does wassima not support, 

It support Windows, MacOS, Linux, and FreeBSD. Wheels are available for every platform but FreeBSD due to restriction by the PyPI (FreeBSD does not have a concept of manylinux). Soon to be supported in Android. iOS is unlikely to be supported. So, beside what is listed, it won't work. But it cover easily 98% of the traffic (according to the public dataset of "downloads")

> is that going to be a problem?

Not at all. It was developed with the main idea that it will never require a compilation toolchain and rather fallback on certifi if installed.

> should this be always on?

Absolutely, yes. There's no reason not to trust the root CA carefully handled by the operating system. It's just wierd to issue a command that basically says:

```shell
$ poetry --please-trust-my-trusted-os-store
```

---

@Secrus 

> To be honest, I don't like this. Seems like the library is quite fresh and wasn't battle-tested before.

Not at all. The core logic is battle tested, and actually, more battle tested than its counterpart. rustls-native-certs is five years old and truststore has 2 years.

So, the question is, can we, as engineer look at fifty-ich lines of code and assess if it's a okay solution to start on? Do we need "other" people to "run it" so that we gain confidence? Do "other" people say the same thing, therefor locked in circular loop?

rustls-native-certs is trusted. "wassima" is rather new, okay, but I am not new myself, and can see what projects I maintain.

> Also, I don't think we need a library that wraps a compiled code unless there is no other viable option (in this case, there is truststore).

truststore isn't as viable as you think it is. if I started "wassima" it was because of issues caused by it. The monkeypatching of the ssl standard library is causing weird side effect depending on numerous factors, it also have an unconsistent behavior depending on the platform (e.g. MacOS, and Windows, they use the native OS certificate validation (they don't extract the root CA, but rather send the validation process to the OS), whereas Linux still is based on Python native capabilities).

> are you willing to work on this feature, but basing it on truststore instead of your own library?

unfortunately no.

If the final answer is no, then I'll invite you to close this PR.

Thanks for the response.

regards.

dimbleby:
> > should this be always on?
> Absolutely, yes

Yet all the other projects disagree, for varying reasons...  are uv's performance concerns real?

I am less allergic than Secrus to compiled code: if it's a good solution then it's a good solution.

I notice you have suppressed type warnings.  Prefer instead to add annotations.

Ousret:
> Yet all the other projects disagree

Not at all. for example, recently, pip changed to truststore+certifi by default as I am proposing here but with "wassima"+certifi.

> are uv's performance concerns real?

In our era where an energetic crisis is happening, along with global warming, I think speed and cpu time saving are vital.

> I notice you have suppressed type warnings. Prefer instead to add annotations.

Yes, but Requests is untyped, and the typeshed does not have type for this method. I am uncomfortable setting types, but if you want to, I can, for sure.

regards,

dimbleby:
> In our era where an energetic crisis is happening, along with global warming, I think speed and cpu time saving are vital.

a curious response, but insofar as it addresses the question I think you are agreeing that there _is_ a performance problem?

dimbleby:
> typeshed does not have type for this method

if it makes you happier - I'd encourage you to submit a pull request there too, then!

Ousret:
>  I think you are agreeing that there is a performance problem?

There is, but the debate is way larger than this and I am not the expert on the broader field.

> if it makes you happier - I'd encourage you to submit a pull request there too, then!

does this mean you are considering this PR for merge?

regards,

dimbleby:
I don't have the power to merge or reject this PR, I am just a humble contributor

Improving typeshed is desirable regardless of whether this is merged, if you've chanced across a gap that it would be helpful to fill then I'd encourage you to do that either way

When I ask about the performance impact I think you know that I am not looking for a debate about global warming.  I am asking whether poetry users will notice a meaningful difference.  uv's user-facing documentation says that this is something they think is significant, at least on macos - are they right?  is it also significant in the poetry context?

(uv is faster than poetry in general, it's plausible that a performance penalty that is important to them is not important to poetry)

Ousret:
> Improving typeshed is desirable regardless of whether this is merged

It's debatable, because requests maintainers don't approve the typing established in typeshed, for various valid reasons.

> When I ask about the performance impact I think you know that I am not looking for a debate about global warming.

I understand, clearly, but myself was surprised with the question as this PR isn't concerned with that topic.

> I am asking whether poetry users will notice a meaningful difference.

None. As long as we still load certifi, which is the case for that PR, the performance will be the same.
If we were to stop from loading certifi, and solely use the in-memory store from the OS, a noticeable performance bump (expressed in hundred of ms) will be noticed. Mostly UX / feel reactive kind of things.

> uv's user-facing documentation says that this is something they think is significant, at least on macos - are they right?

yes. I don't have the data in-hand, but my experience with Python natively accessing truststore through ext. func calls are slow. idk the factor.

> uv is faster than poetry in general, it's plausible that a performance penalty that is important to them is not important to poetry

I don't know how to answer this last point.

---

Finally, let's wait for guidance / dismissal / approval from a maintainer with access before doing any further effort.

regards,

Ousret:
gentle ping @Secrus 

If this PR can't be accepted, feel free to close it. 

Side note: I've added / injected the OS truststore (if available) in the dulwich client (if http transport of course).

PR #9692, opened 2024-09-17
feat: drop support for Python 3.8

finswimmer:
Python 3.8 will be EOL on October. Let's drop the support than for running Poetry itself with it.


TheSven73:
From `py3.9` onwards, we can make some `executor.py` code simpler:
```python
            if self._shutdown:
                # Cancelling further tasks from being executed
                [task.cancel() for task in tasks]
                self._executor.shutdown(wait=True)
```
=>
```python
            if self._shutdown:
                self._executor.shutdown(wait=True, cancel_futures=True)
```

radoering (approved):

PR #9700, opened 2024-09-22
fix: lock all extra markers in dependencies if package is required by multiple extras

### LINKED ISSUE #9636: Poetry does not install a non-optional package from lock file
### Description

Using `pyproject.toml` like this:
```toml
[tool.poetry]
name = "poetry-sandbox"
version = "0.1.0"
description = ""
authors = ["rinaldv"]
readme = "README.md"

[tool.poetry.dependencies]
python = "^3.10"
burr = {extras = ["tracking-client-s3", "tracking-server-s3", "cli"], version = "^0", allow-prereleases = true }

[build-system]
requires = ["poetry-core"]
```

Use `poetry lock` to generate the lock file (see in attachments, [poetry.lock](https://github.com/user-attachments/files/16641153/poetry.lock.txt)). Notice that it contains the following snippet about non-optional `requests` package:
```toml
[[package]]
name = "requests"
version = "2.32.3"
description = "Python HTTP for Humans."
optional = false
python-versions = ">=3.8"
files = [
    {file = "requests-2.32.3-py3-none-any.whl", hash = "sha256:70761cfe03c773ceb22aa2f671b4757976145175cdfca038c02654d061d6dcc6"},
    {file = "requests-2.32.3.tar.gz", hash = "sha256:55365417734eb18255590a9ff9eb97e9e1da868d4ccd6402399eaf68af20a760"},
]

[package.dependencies]
certifi = ">=2017.4.17"
charset-normalizer = ">=2,<4"
idna = ">=2.5,<4"
urllib3 = ">=1.21.1,<3"

[package.extras]
socks = ["PySocks (>=1.5.6,!=1.5.7)"]
use-chardet-on-py3 = ["chardet (>=3.0.2,<6)"]

[[package]]
name = "s3transfer"
```

Do `poetry install` and notice that `requests` is not installed. Logs for `-vvv` are attached, see [poetry-install.log](https://github.com/user-attachments/files/16641377/poetry-install.log).
Expected behavior - it should be installed, because it seems to be resolved this way already into the lock file, and before it was resolved as dependency for `burr[cli]`.

@radoering mentioned in discord that this might be related to poetry#9427 and `installer.re-resolve` but this is reported for `1.8.3` version

### Workarounds

Adding missing dependencies manually into `pyproject.toml`

pipx

### Operating System

MacOS 13.6.7 

### Poetry Version

1.8.3

### Poetry Configuration

```bash session
cache-dir = "/Users/rgareev/Library/Caches/pypoetry"
experimental.system-git-client = false
installer.max-workers = null
installer.modern-installation = true
installer.no-binary = null
installer.parallel = true
keyring.enabled = true
solver.lazy-wheel = true
virtualenvs.create = true
virtualenvs.in-project = null
virtualenvs.options.no-setuptools = false
virtualenvs.path = "{cache-dir}/virtualenvs"  # /Users/rgareev/Library/Caches/pypoetry/virtualenvs
virtualenvs.prefer-active-python = false
warnings.export = true
```

### Python Sysconfig

_No response_

### Example pyproject.toml

```TOML
[tool.poetry]
name = "poetry-sandbox"
version = "0.1.0"
description = ""
authors = ["rinaldv"]
readme = "README.md"

[tool.poetry.dependencies]
python = "^3.10"
burr = {extras = ["tracking-client-s3", "tracking-server-s3", "cli"], version = "^0", allow-prereleases = true }

[build-system]
requires = ["poetry-core"]
```

### Poetry Runtime Logs

```bash session
[poetry-install.log](https://github.com/user-attachments/files/16641377/poetry-install.log)
```

dimbleby:
fixes #9636 

could do with a testcase really but I doubt that I will have the energy for it.  Maybe you will consider this "obviously correct" and low enough risk to merge anyway, maybe it will have to wait for someone who is more motivated to write that testcase

either way I figured this was more useful as a pull request than not

### DIFF
--- a/src/poetry/puzzle/solver.py
+++ b/src/poetry/puzzle/solver.py
@@ -179,16 +179,14 @@ class Solver:
                             if _package.name == dep.name:
                                 continue
 
-                            try:
-                                index = _package.requires.index(dep)
-                            except ValueError:
-                                _package.add_dependency(dep)
-                            else:
-                                _dep = _package.requires[index]
-                                if _dep.marker != dep.marker:
-                                    # marker of feature package is more accurate
-                                    # because it includes relevant extras
-                                    _dep.marker = dep.marker
+                            # Avoid duplication.
+                            if any(
+                                _dep == dep and _dep.marker == dep.marker
+                                for _dep in _package.requires
+                            ):
+                                continue
+
+                            _package.add_dependency(dep)
             else:
                 final_packages.append(package)
                 depths.append(results[package])
--- a/tests/puzzle/test_solver.py
+++ b/tests/puzzle/test_solver.py
@@ -663,6 +663,63 @@ def test_solver_returns_extras_when_multiple_extras_use_same_dependency(
     assert ops[0].package.marker.is_any()
 
 
+def test_solver_locks_all_extras_when_multiple_extras_require_same_dependency(
+    solver: Solver,
+    repo: Repository,
+    package: ProjectPackage,
+) -> None:
+    """
+    - root depends on A[extra-b1] and C
+    - C depends on A[extra-b2]
+    - B is required by both extras
+    -> the locked dependency A on B must have both extra markers
+    """
+    package_a = get_package("A", "1.0")
+    package_b = get_package("B", "1.0")
+    package_c = get_package("C", "1.0")
+
+    dep_b1 = get_dependency("B", "*", optional=True)
+    dep_b1.marker = parse_marker("extra == 'extra-b1'")
+
+    dep_b2 = get_dependency("B", "*", optional=True)
+    dep_b2.marker = parse_marker("extra == 'extra-b2'")
+
+    package_a.extras = {
+        canonicalize_name("extra-b1"): [dep_b1],
+        canonicalize_name("extra-b2"): [dep_b2],
+    }
+    package_a.add_dependency(dep_b1)
+    package_a.add_dependency(dep_b2)
+
+    package.add_dependency(
+        get_dependency("A", {"version": "*", "extras": ["extra-b1"]})
+    )
+    package.add_dependency(get_dependency("C", "*"))
+    package_c.add_dependency(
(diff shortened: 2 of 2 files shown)


radoering (changes requested):
We should add a test for that. In case you do not want to write one, I will take a look next week.

dimbleby:
I have been sitting on this for ages without writing a testcase, apologies for leaving the job half done but I think it is unlikely to happen

src/poetry/puzzle/solver.py
```diff
@@ -179,16 +179,14 @@ def _solve(self) -> tuple[list[Package], list[int]]:
...
-                                    # because it includes relevant extras
-                                    _dep.marker = dep.marker
+                            # Avoid duplication.
+                            if any(
+                                _dep == dep and _dep.marker == dep.marker
+                                for _dep in _package.requires
+                            ):
+                                continue
+
+                            _package.add_dependency(dep)
```
radoering, line 189:
```suggestion
                            try:
                                index = _package.requires.index(dep)
                            except ValueError:
                                _package.add_dependency(dep)
                            else:
                                _dep = _package.requires[index]
                                if _dep.marker != dep.marker:
                                    # merge markers
                                    _dep.marker = _dep.marker.union(dep.marker)
```

That seems to be an alternative solution that results in

```
requests = {version = "*", optional = true, markers = "extra == \"tracking-server\" or extra == \"cli\""}
```

whereas the original fix results in

```
requests = [
    {version = "*", optional = true, markers = "extra == \"cli\""},
    {version = "*", optional = true, markers = "extra == \"tracking-server\""},
]
```

Do you think one is better than the other or is it just a matter of taste?

dimbleby: I spotted that possibility and decided that it didn't much matter, perhaps it is a little more cautious to preserve the separate dependencies as we find them rather than mangle them.  

I find that the way that I have written it is a little clearer, but certainly not in an important way, probably it is mostly a matter of taste.

radoering (approved):

Issue #7592, opened 2023-03-02
Unable to use Flask debugger


delucca:
- **Poetry version**: 1.4.0
- **Python version**: 3.10
- **OS version and name**: Arch Linux

- [x] I am on the [latest](https://github.com/python-poetry/poetry/releases/latest) stable Poetry version, installed using a recommended method.
- [x] I have searched the [issues](https://github.com/python-poetry/poetry/issues) of this repo and believe that this is not a duplicate.
- [x] I have consulted the [FAQ](https://python-poetry.org/docs/faq/) and [blog](https://python-poetry.org/blog/) for any relevant entries or release notes.
- [x] If an exception occurs when executing a command, I executed it again in debug mode (`-vvv` option) and have included the output below.

## Issue
Following up [on this issue](https://github.com/mfussenegger/nvim-dap-python/issues/85), for some reason nvim-dap is not being able to attach a debugger if I'm running an application using Poetry

If I run using Python directly it works, but using Poetry it automatically closes.

You can see the details on the original issue.

delucca:
@dimbleby this isn't using poetry run :)

Even in a Poetry shell, the issue is the same. If I install the application with Poetry, Flask debugger doesnt work for some reason (with Nvim)

dimbleby:
all you have given us here is a link to another issue, in which the configuration shows that you _are_ calling `poetry run`

```
	program = "${env:PYTHON_VENV_PATH}/bin/poetry",
	args = {
		"run",
		"flask",
		"--no-debug",
		"run",
	},
```

if you're now saying that's not what you're doing then I don't even

delucca:
@dimbleby as I've mentioned, the same thing happens if you open a shell instead of using `poetry run` (as I've mentioned in the issue)

dimbleby:
do you mean `poetry shell`?  I wouldn't do that either: just activate the virtual environment in the regular way and run things directly.

This report is too involved and too confused: if you can reduce it to something that someone can reproduce without having to install neovim and dap and flask then you'll have a better chance of someone taking an interest in it.

delucca:
@dimbleby you can use the same example from the original issue (on the last comment from the nvim-dap-python author)

I can create a github repo if you want, but essentially you just need to launch any simple Python application. Unfortunately I don't see any possible alternative of exploring this issue without having to install `nvim` and `nvim-dap`, since the issue is specifically a problem with debugging in that specific tool 😄

dimbleby:
[...]

PR #9708, opened 2024-09-25
docs: How to debug your main script from VS Code

mmartinortiz:
Resolves: #7592 

After having some difficulties to debug the main script of my program while using VS Code, I found [this answer](https://github.com/python-poetry/poetry/issues/7592#issuecomment-1897482128) that helped me to solve my issue. I thought that would be worth for others to have this piece of information within Poetry's documentation.

### DIFF
--- a/docs/faq.md
+++ b/docs/faq.md
@@ -271,3 +271,50 @@ The two key options we are using here are `--no-root` (skips installing the proj
 Poetry's default HTTP request timeout is 15 seconds, the same as `pip`.
 Similar to `PIP_REQUESTS_TIMEOUT`, the **experimental** environment variable `POETRY_REQUESTS_TIMEOUT`
 can be set to alter this value.
+
+### How to debug the main script of your project using Poetry, VS Code and `launch.json`
+
+To debug a project that is managed by `poetry` and has a main script, you need to install `poetry` as a development dependency and use the `poetry run` command to execute the script via the `launch.json` file, and configure `poetry` to create the virtual environment in the project directory.
+
+Make sure that `poetry` creates the virtual environment in your project directory. If that is not your case, you will have to recreate your virtual environment after the configuration options is changed.
+
+```bash
+poetry config virtualenvs.in-project true
+```
+
+Given that your `pyproject.toml` has the following entry:
+
+```toml
+[tool.poetry.scripts]
+my_program = "my_package.my_program:main"
+```
+
+Add `poetry` as a development dependency:
+
+```bash
+poetry add --group dev poetry
+```
+
+Then, create a `launch.json` file (if it does not exists already) in the `.vscode` directory of your project with the following content:
+
+```json
+{
+    "version": "0.2.0",
+    "configurations": [
+        {
+            "name": "Python: Debug my_program",
+            "type": "python",
+            "request": "launch",
+            "console": "integratedTerminal",
+            "module": "poetry",
+            "args": [
+                "run",
+                "my_program",
+                // Add other arguments for your program here if the are required
+            ]
+        }
+    ]
+}
+```
+
+Now you can start the debugger in VS Code and it will run starting from the main script of your project.


radoering:
I do not think we should recommend adding poetry to the dev dependencies of the project because this can lead to unexpected behavior. I am not a VS Code user, but I assume you can just configure the correct interpreter (the one returned by `poetry env info`) and run your entry point directly with this interpreter to debug it.

mmartinortiz:
> I am not a VS Code user, but I assume you can just configure the correct interpreter (the one returned by poetry env info) and run your entry point directly with this interpreter to debug it.

This is one of the multiple things I tried. Since `poetry` does not install the package being developed into the virtual environment, the interpreter cannot resolve the imports made from your own module. (At least, that is how understood it while I tried to make it work)

I found some threads in StackOverflow ([this one is just an example](https://stackoverflow.com/questions/69106483/python-project-with-poetry-how-to-debug-it-in-visual-studio-code)) about how to get VS Code debug working for a project managed with `poetry` and I never found a working solution, except for this one, that install `poetry` as development dependency.

I can understand that it is not neither an elegant or robust solution and I'm very happy to document a more elegant method :-)

radoering:
> Since `poetry` does not install the package being developed into the virtual environment,

Poetry installs the package into the virtual environment in editable mode (except for when you set `package-mode = false`). You should find a *.pth file named after your project in the site packages of the virtual environment after running `poetry install`.

mmartinortiz:
The documentation states that package mode is the default behavior. This is true for my project; however, unless I install poetry as a development dependency, I encounter a 'Module not found' error, and I can't get debugging to work in VS Code.

Clearly, my assumption about the reason for the module not being found was incorrect. What could be then the reason for the debugger to not find the error?

TheSven73:
vscode debugging works for me without having to install poetry as a development dependency.

First, make sure vscode is using the venv which poetry has created. Run `poetry env info` and copy the location of the executable. Then, in vscode, run "Python: Select Interpreter" -> "Enter Interpreter path" -> paste the executable location. The bottom right of the vscode window should now show something like this 
Screenshot 2024-09-25 at 11 13 14 PM


My main function is located in `src/hypermodern_python/console.py`:
```python
import click

@click.command()
def main():
    click.echo('Hello, world!')


if __name__ == "__main__":
    main()
```
Next, create a `launch.json` which specifies the module which contains main:
```json
{
    "version": "0.2.0",
    "configurations": [
        {
            "name": "My script",
            "type": "debugpy",
            "request": "launch",
            "cwd": "${workspaceFolder}",
            "module": "hypermodern_python.console",
            "justMyCode": false,
        }
    ]
}
```
Place a breakpoint, run "Debug: Start Debugging" in vscode, and the breakpoint should be hit.
Screenshot 2024-09-25 at 11 12 00 PM

mmartinortiz:
Thanks @TheSven73 , with your `launch.json` file I managed to debug my main script. I'm not sure how it different to other attempts, maybe it is just combination of `justMyCode` and `module`

I've updated my pull request with your instructions.

TheSven73:
> I'm not sure how it different to other attempts

Not sure either, hard to tell without a clear description of your attempt(s) (config+code).

Secrus:
Drive-by question: if we were to include that in docs, what about other editors? Why should we have a special treatment for VS Code?

mmartinortiz:
> Drive-by question: if we were to include that in docs, what about other editors? Why should we have a special treatment for VS Code?

From my point of view, this is not giving a special treatment to VS Code. 

When I found myself in need to debug a project managed with Poetry, setting it up in VS Code was not as straight forward as it is for projects not managed with Poetry. After searching for "how to do it" I found more people straggled with this, proposing different alternatives.

After I succeeded, and given that more people struggled with this, I thought would be worth to share the "how to" together with poetry's documentation.

TheSven73:
I'm not convinced we have a clear definition yet of the problem we're trying to solve.

I just tried the [official Microsoft tutorial to debug Flask](https://code.visualstudio.com/docs/python/tutorial-flask) and it works like a charm with Poetry. The only difference is that the developer must point vscode to the venv created by poetry, instead of hand-rolling their own using `python -m venv` and `pip`.

So I suspect that this is not actionable, unless and until someone in the linked issue(s) can provide a clear and reproducible problem statement. Other than: "must tell the user to point vscode to the poetry venv", but isn't that true for other IDEs as well?

Secrus:
Stalled

PR #9723, opened 2024-09-29
Fix race condition with forked monorepos

TheSven73:
There is a race condition when installing packages from multiple forked monorepos. Fix it.

See [discussion](https://github.com/orgs/python-poetry/discussions/9718#discussioncomment-10785589)

### DIFF
--- a/src/poetry/installation/executor.py
+++ b/src/poetry/installation/executor.py
@@ -33,6 +33,7 @@ from poetry.utils.helpers import pluralize
 from poetry.utils.helpers import remove_directory
 from poetry.utils.isolated_build import IsolatedBuildError
 from poetry.utils.isolated_build import IsolatedBuildInstallError
+from poetry.vcs.git import Git
 
 
 if TYPE_CHECKING:
@@ -44,6 +45,12 @@ if TYPE_CHECKING:
     from poetry.installation.operations.operation import Operation
     from poetry.repositories import RepositoryPool
     from poetry.utils.env import Env
+
+
+def _package_get_name(package: Package) -> str | None:
+    if url := package.repository_url:
+        return Git.get_name_from_source_url(url)
+    return None
 
 
 class Executor:
@@ -167,8 +174,9 @@ class Executor:
                 if is_parallel_unsafe:
                     serial_operations.append(operation)
                 elif operation.package.source_type == "git":
-                    # Git operations on the same repository should be executed serially
-                    serial_git_operations[operation.package.source_url].append(
+                    # Serially execute git operations that get cloned to the same directory,
+                    # to prevent multiple parallel git operations in the same repo.
+                    serial_git_operations[_package_get_name(operation.package)].append(
                         operation
                     )
                 else:
@@ -604,8 +612,6 @@ class Executor:
         )
 
     def _prepare_git_archive(self, operation: Install | Update) -> Path:
-        from poetry.vcs.git import Git
-
         package = operation.package
         assert package.source_url is not None
 
--- a/tests/installation/test_executor.py
+++ b/tests/installation/test_executor.py
@@ -1137,6 +1137,49 @@ def test_executor_should_install_multiple_packages_from_same_git_repository(
     assert archive_arg == tmp_venv.path / "src/demo/subdirectories/package_b"
 
 
+@pytest.mark.xfail
+def test_executor_should_install_multiple_packages_from_forked_git_repository(
+    mocker: MockerFixture,
+    tmp_venv: VirtualEnv,
+    pool: RepositoryPool,
+    config: Config,
+    artifact_cache: ArtifactCache,
+    io: BufferedIO,
+    wheel: Path,
+) -> None:
+    package_a = Package(
+        "one",
+        "1.0.0",
+        source_type="git",
+        source_reference="master",
+        source_resolved_reference="123456",
+        source_url="https://github.com/demo/subdirectories.git",
+        source_subdirectory="one",
+    )
+    package_b = Package(
+        "two",
+        "2.0.0",
+        source_type="git",
+        source_reference="master",
+        source_resolved_reference="123456",
+        source_url="https://github.com/forked_demo/subdirectories.git",
+        source_subdirectory="two",
+    )
+
+    chef = Chef(artifact_cache, tmp_venv, Factory.create_pool(config))
+    chef.set_directory_wheel(wheel)
+    prepare_spy = mocker.spy(chef, "prepare")
+
+    executor = Executor(tmp_venv, pool, config, io)
+    executor._chef = chef
+    executor.execute([Install(package_a), Install(package_b)])
+
+    # Verify that the repo for package_a is not re-used for package_b.
+    # both repos must be cloned serially into separate directories.
+    # If so, executor.prepare() will be called twice.
+    assert prepare_spy.call_count == 2
+
+
 def test_executor_should_write_pep610_url_references_for_git_with_subdirectories(
     tmp_venv: VirtualEnv,
     pool: RepositoryPool,
--- a/tests/installation/test_executor.py
+++ b/tests/installation/test_executor.py
@@ -1091,7 +1091,7 @@ def test_executor_should_append_subdirectory_for_git(
     executor.execute([Install(package)])
 
     archive_arg = spy.call_args[0][0]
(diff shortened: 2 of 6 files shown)


TheSven73:
Not sure how to add reviewers / watchers. @radoering @gustavgransbo @seeker25

tests/installation/test_executor.py
```diff
@@ -1050,37 +1050,6 @@ def test_executor_should_write_pep610_url_references_for_editable_git(
     assert not (prepare_spy.spy_return.parent / ".created_from_git_dependency").exists()
 
 
-def test_executor_should_append_subdirectory_for_git(
```
radoering, line 1053: Why removing the test and not adapting it?

src/poetry/installation/executor.py
```diff
@@ -167,8 +174,9 @@ def execute(self, operations: list[Operation]) -> int:
                 if is_parallel_unsafe:
                     serial_operations.append(operation)
                 elif operation.package.source_type == "git":
-                    # Git operations on the same repository should be executed serially
-                    serial_git_operations[operation.package.source_url].append(
+                    # Serially execute git operations that get cloned to the same directory,
+                    # to prevent multiple parallel git operations in the same repo.
+                    serial_git_operations[_package_get_name(operation.package)].append(
```
radoering, line 179: If for some reason `_package_get_name()` returns `None`, this should still be safe because all operations with `None` are serialized, correct?

src/poetry/installation/executor.py
```diff
@@ -33,6 +33,7 @@
 from poetry.utils.helpers import remove_directory
 from poetry.utils.isolated_build import IsolatedBuildError
 from poetry.utils.isolated_build import IsolatedBuildInstallError
+from poetry.vcs.git import Git
```
radoering, line 36: I assume the `Git` import was local for a reason. It is probably slow and should be avoided in case we do not need it.

radoering (changes requested):
LGTM. 👍
Just some minor nitpicks.

tests/installation/test_executor.py
TheSven73: I removed the test because I don't understand (yet) why it's there, and this is a WIP. I now pushed a passing version.

Can you help understand why the test is there? It seems to verify an implementation detail (dir layout for emulated git client) yet not the behaviour in question. The test was added as part of a fix for https://github.com/python-poetry/poetry/issues/7575 **but** when I run the example linked in the issue, the behaviour seems broken. Did things regress?
```sh
$ poetry install

  - Installing numpy (1.16.6): Failed

  ChefBuildError

  Backend subprocess exited when trying to invoke build_wheel
  
  Running from numpy source directory.
  /tmp/tmp8jilduae/numpy-1.16.6/numpy/distutils/misc_util.py:476: SyntaxWarning: "is" with 'str' literal. Did you mean "=="?
    return is_string(s) and ('*' in s or '?' is s)
    File "/usr/local/py-utils/venvs/poetry/lib/python3.12/site-packages/pyproject_hooks/_in_process/_in_process.py", line 373, in <module>
      main()
    File "/usr/local/py-utils/venvs/poetry/lib/python3.12/site-packages/pyproject_hooks/_in_process/_in_process.py", line 357, in main
      json_out["return_val"] = hook(**hook_input["kwargs"])
                               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    File "/usr/local/py-utils/venvs/poetry/lib/python3.12/site-packages/pyproject_hooks/_in_process/_in_process.py", line 271, in build_wheel
      return _build_backend().build_wheel(
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    File "/tmp/tmphipzhe0d/.venv/lib/python3.12/site-packages/setuptools/build_meta.py", line 421, in build_wheel
      return self._build_with_temp_dir(
             ^^^^^^^^^^^^^^^^^^^^^^^^^^
    File "/tmp/tmphipzhe0d/.venv/lib/python3.12/site-packages/setuptools/build_meta.py", line 403, in _build_with_temp_dir
      self.run_setup()
    File "/tmp/tmphipzhe0d/.venv/lib/python3.12/site-packages/setuptools/build_meta.py", line 503, in run_setup
      super().run_setup(setup_script=setup_script)
    File "/tmp/tmphipzhe0d/.venv/lib/python3.12/site-packages/setuptools/build_meta.py", line 318, in run_setup
      exec(code, locals())
    File "<string>", line 419, in <module>
    File "<string>", line 398, in setup_package
    File "/tmp/tmp8jilduae/numpy-1.16.6/numpy/distutils/core.py", line 26, in <module>
      from numpy.distutils.command import config, config_compiler, \
    File "/tmp/tmp8jilduae/numpy-1.16.6/numpy/distutils/command/config.py", line 19, in <module>
      from numpy.distutils.mingw32ccompiler import generate_manifest
    File "/tmp/tmp8jilduae/numpy-1.16.6/numpy/distutils/mingw32ccompiler.py", line 34, in <module>
      from distutils.msvccompiler import get_build_version as get_build_msvc_version
  ModuleNotFoundError: No module named 'distutils.msvccompiler'
  

  at /usr/local/py-utils/venvs/poetry/lib/python3.12/site-packages/poetry/installation/chef.py:164 in _prepare
      160│ 
      161│                 error = ChefBuildError("\n\n".join(message_parts))
      162│ 
      163│             if error is not None:
    → 164│                 raise error from None
      165│ 
      166│             return path
      167│ 
      168│     def _prepare_sdist(self, archive: Path, destination: Path | None = None) -> Path:

Note: This error originates from the build backend, and is likely not a problem with poetry but with numpy (1.16.6) not supporting PEP 517 builds. You can verify this by running 'pip wheel --no-cache-dir --use-pep517 "numpy (==1.16.6)"'.
```

radoering: The test checks that the path passed to `Chef.prepare()` is the path of the root directory of the Python package. In this test, the Python package is not in the root of the git repo but in a subdirectory. Therefore, the path must be the path of the subdirectory. Otherwise, the subsequent build would fail.

> when I run the example linked in the issue, the behaviour seems broken. Did things regress?

I assume the example fails because numpy 1.16.6 is not compatible with latest setuptools but claims to be compatible. In other words, it is broken because of a recent setuptools release.

> The test was added as part of a fix for https://github.com/python-poetry/poetry/issues/7575

If you remove https://github.com/python-poetry/poetry/blob/8622273717a686ba34b3e2288ec8192fdb1c5e85/src/poetry/installation/executor.py#L595-L596
which has been added in the PR with the fix, you will see the test fail.

src/poetry/installation/executor.py
TheSven73: Would you like me to make the import local to the helper function?

tests/installation/test_executor.py
TheSven73: Thanks for clarifying. Does this mean that the modified test in this MR is ok to merge?

tests/installation/test_executor.py
```diff
@@ -1130,11 +1130,49 @@ def test_executor_should_install_multiple_packages_from_same_git_repository(
     executor._chef = chef
     executor.execute([Install(package_a), Install(package_b)])
 
-    archive_arg = spy.call_args_list[0][0][0]
-    assert archive_arg == tmp_venv.path / "src/demo/subdirectories/package_a"
+    assert spy.call_count == 2
 
-    archive_arg = spy.call_args_list[1][0][0]
-    assert archive_arg == tmp_venv.path / "src/demo/subdirectories/package_b"
```
radoering, line 1137: I'd still keep these checks (and just remove the "demo" part).

src/poetry/installation/executor.py
radoering: > The import in question seems to be already imported in our main branch:

Interesting. That seems to be the import in `direct_origin.py`.

> Would you like me to make the import local to the helper function?

However, even when making this import local, it does not seem to have a significant effect so I suppose we can keep it global. We probably should investigate import times a bit in general before the next release. There seem to be some quick wins.

> time poetry run poetry > /dev/null

You will get better measurements, if you activate the venv manually and than just run `time poetry > /dev/null` so there is less irrelevant code included in the measurement. Further, it might make sense to check the time for different commands, which are expected to be fast, e.g. `poetry -V`.

tests/installation/test_executor.py
TheSven73: Done.

radoering (approved):

PR #9732, opened 2024-10-05
remove deprecated cli options and methods, revoke deprecation of `--dev`

radoering:
There are at least some users who will miss `--dev/-D`, see https://github.com/orgs/python-poetry/discussions/7530.

Personally, I still tend to remove it.

### DIFF
--- a/docs/cli.md
+++ b/docs/cli.md
@@ -448,6 +448,7 @@ about dependency groups.
 ### Options
 
 * `--group (-G)`: The group to add the dependency to.
+* `--dev (-D)`: Add package as development dependency. (shortcut for `-G dev`)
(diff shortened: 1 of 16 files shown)


Secrus:
I would leave the `--dev/-D` option for `add` (maybe `remove`) as an alias for `--group dev`. It's convenient and since we still support `dev-dependencies`, it will be confusing to a lot of people when the option disappears.

radoering:
Then, should we undeprecate it?

Secrus:
> Then, should we undeprecate it?

I would say yes, for the commands I mentioned.

Secrus (approved):

PR #9744, opened 2024-10-09
feat(install): add `--all-groups` flag to install all dependency groups

schneebuzz:
This PR adds the `--all-groups` flag to `poetry install` which includes all the dependency groups (including optional groups) for installation. This is a very common use case for projects with multiple optional groups and allows the user to *install everything* without specifying each group.

This feature has been discussed in the past. See e.g. #7605 and #5461. I gave it another shot with an implementation that respects the existing options and flags.

I decided not to go with `--with=all` or similar and instead use the explicit flag `--all-groups`, consistent with the `--all-extras` flag. A naming alternative would be `--with-all`.

Options like
- `--all-groups --without=docs` are supported,
- `--all-groups --with=foo` ignores the `with` with a warning, 
- and `--all-groups --only=foo` results in an error.

# Pull Request Check List

### DIFF
--- a/docs/cli.md
+++ b/docs/cli.md
@@ -168,6 +168,12 @@ You can also select optional dependency groups with the `--with` option.
 poetry install --with test,docs
 ```
 
+To install all dependency groups including the optional groups, use the ``--all-groups`` flag.
+
+```bash
+poetry install --all-groups
+```
+
 It's also possible to only install specific dependency groups by using the `only` option.
 
 ```bash
@@ -261,12 +267,14 @@ poetry install --compile
 * `--dry-run`: Output the operations but do not execute anything (implicitly enables --verbose).
 * `--extras (-E)`: Features to install (multiple values allowed).
 * `--all-extras`: Install all extra features (conflicts with --extras).
+* `--all-groups`: Install dependencies from all groups (conflicts with --only).
 * `--compile`: Compile Python source files to bytecode.
 * `--no-dev`: Do not install dev dependencies. (**Deprecated**, use `--only main` or `--without dev` instead)
 * `--remove-untracked`: Remove dependencies not presented in the lock file. (**Deprecated**, use `--sync` instead)
 
 {{% note %}}
 When `--only` is specified, `--with` and `--without` options are ignored.
+When `--all-groups` is specified, the `--with` option is ignored.
 {{% /note %}}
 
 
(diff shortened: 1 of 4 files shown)


schneebuzz:
Please review, feedback welcome.

Secrus (approved):
LGTM. @radoering, care to take a look?

radoering:
I will take a look tomorrow.

radoering (commented):
LGTM. There is just one thing: https://github.com/python-poetry/poetry-plugin-export/pull/294 introduces `--all-groups` to `poetry export` (not yet released). It is the same name; that is good. However, in `poetry export` `--all-groups` conflicts with `--with` and `--without`. I have no clear preference, but we should make it consistent.

schneebuzz:
> LGTM. There is just one thing: [python-poetry/poetry-plugin-export#294](https://github.com/python-poetry/poetry-plugin-export/pull/294) introduces `--all-groups` to `poetry export` (not yet released). It is the same name; that is good. However, in `poetry export` `--all-groups` conflicts with `--with` and `--without`. I have no clear preference, but we should make it consistent.

Thanks for reviewing.
Even though the *all but without* could be handy, I aligned with the referenced PR since it leaves less room for confusion.

docs/cli.md
```diff
@@ -254,14 +260,16 @@ poetry install --compile
...
 * `--extras (-E)`: Features to install (multiple values allowed).
-* `--all-extras`: Install all extra features (conflicts with --extras).
+* `--all-extras`: Install all extra features (conflicts with `--extras`).
+* `--all-groups`: Install dependencies from all groups (conflicts with `--only`, `--with`, and `--without`).
 * `--compile`: Compile Python source files to bytecode.
 * `--remove-untracked`: Remove dependencies not presented in the lock file. (**Deprecated**, use `--sync` instead)
 
 {{% note %}}
 When `--only` is specified, `--with` and `--without` options are ignored.
+When `--all-groups` is specified, the `--with` option is ignored.
```
radoering, line 272: That is not up to date anymore.

radoering (changes requested):

docs/cli.md
schneebuzz: Apologies for missing that one. Fixed.

radoering (approved):

PR #9748, opened 2024-10-10
Try finding commit using short SHA if it is not on HEAD

mikhainin:
# Pull Request Check List

If a dependency specifies a git revision using a short hash, Dulwich does not provide an option to directly address it (see discussion in #9560); however, we can still list revisions and find what we are looking for.

Resolves: #9560

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -189,7 +189,7 @@ addopts = [ "-n", "logical", "-ra", "--strict-config", "--strict-markers" ]
 testpaths = ["tests"]
 markers = [
     "network: mark tests that require internet access",
-    "skip_git_mock: makrs test should not auto-apply git_mock"
+    "skip_git_mock: mark tests that should not auto-apply git_mock"
 ]
 log_cli_level = "INFO"
 xfail_strict = true
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -189,6 +189,7 @@ addopts = [ "-n", "logical", "-ra", "--strict-config", "--strict-markers" ]
 testpaths = ["tests"]
 markers = [
     "network: mark tests that require internet access",
+    "skip_git_mock: makrs test should not auto-apply git_mock"
 ]
 log_cli_level = "INFO"
 xfail_strict = true
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -35,7 +35,7 @@ poetry-core = { git = "https://github.com/python-poetry/poetry-core.git", branch
 build = "^1.2.1"
 cachecontrol = { version = "^0.14.0", extras = ["filecache"] }
 cleo = "^2.2.1"
-dulwich = "^0.22.1"
+dulwich = "^0.22.6"
 fastjsonschema = "^2.18.0"
 importlib-metadata = { version = ">=4.4", python = "<3.10" }
 installer = "^0.7.0"
--- a/src/poetry/vcs/git/backend.py
+++ b/src/poetry/vcs/git/backend.py
@@ -18,7 +18,6 @@ from dulwich.config import ConfigFile
 from dulwich.config import parse_submodules
 from dulwich.errors import NotGitRepository
 from dulwich.index import IndexEntry
-from dulwich.objects import Commit
 from dulwich.refs import ANNOTATED_TAG_SUFFIX
 from dulwich.repo import Repo
 
@@ -55,14 +54,14 @@ class GitRefSpec:
     tag: str | None = None
     ref: bytes = dataclasses.field(default_factory=lambda: b"HEAD")
 
-    def resolve(self, remote_refs: FetchPackResult) -> None:
+    def resolve(self, remote_refs: FetchPackResult, repo: Repo) -> None:
         """
         Resolve the ref using the provided remote refs.
         """
-        self._normalise(remote_refs=remote_refs)
+        self._normalise(remote_refs=remote_refs, repo=repo)
         self._set_head(remote_refs=remote_refs)
 
-    def _normalise(self, remote_refs: FetchPackResult) -> None:
+    def _normalise(self, remote_refs: FetchPackResult, repo: Repo) -> None:
         """
         Internal helper method to determine if given revision is
             1. a branch or tag; if so, set corresponding properties.
@@ -99,7 +98,12 @@ class GitRefSpec:
             for sha in remote_refs.refs.values():
                 if sha.startswith(short_sha):
                     self.revision = sha.decode("utf-8")
-                    break
+                    return
+
+            # no heads with such SHA, let's check all objects
+            for sha in repo.object_store.iter_prefix(short_sha):
+                self.revision = sha.decode("utf-8")
+                return
 
     def _set_head(self, remote_refs: FetchPackResult) -> None:
         """
@@ -270,7 +274,7 @@ class Git:
         )
 
         try:
-            refspec.resolve(remote_refs=remote_refs)
+            refspec.resolve(remote_refs=remote_refs, repo=local)
         except KeyError:  # branch / ref does not exist
             raise PoetryConsoleError(
                 f"Failed to clone {url} at '{refspec.key}', verify ref exists on"
@@ -314,37 +318,12 @@ class Git:
             if isinstance(e, AssertionError) and "Invalid object name" not in str(e):
                 raise
 
-            short_ref_found = False
-
-            if refspec.is_sha_short:
-                commit = cls._find_object_by_sha_prefix(local, refspec.key.encode())
-                if commit is not None:
-                    logger.debug(
-                        "\nResolved short SHA <c2>%s</c2> as <c2>%s</c2>",
-                        refspec.key,
-                        commit.id,
-                    )
-                    local.refs[b"HEAD"] = commit.id
-                    with local:
-                        local.reset_index()
-                    short_ref_found = True
-
-            if not short_ref_found:
-                raise PoetryConsoleError(
-                    f"Failed to clone {url} at '{refspec.key}', verify ref exists on"
-                    " remote."
-                )
+            raise PoetryConsoleError(
+                f"Failed to clone {url} at '{refspec.key}', verify ref exists on"
+                " remote."
+            )
 
         return local
-
-    @classmethod
-    def _find_object_by_sha_prefix(cls, repo: Repo, sha_prefix: bytes) -> Commit | None:
-        for sha in repo.object_store:
-            if sha.startswith(sha_prefix):
-                obj = repo.object_store[sha]
-                if isinstance(obj, Commit):
-                    return obj
-        return None
 
     @classmethod
     def _clone_submodules(cls, repo: Repo) -> None:
--- a/src/poetry/vcs/git/backend.py
+++ b/src/poetry/vcs/git/backend.py
@@ -18,6 +18,7 @@ from dulwich.config import ConfigFile
 from dulwich.config import parse_submodules
 from dulwich.errors import NotGitRepository
 from dulwich.index import IndexEntry
+from dulwich.objects import Commit
 from dulwich.refs import ANNOTATED_TAG_SUFFIX
 from dulwich.repo import Repo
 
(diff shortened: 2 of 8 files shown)


mikhainin:
It seemed to work in my case, but I would appreciate it if someone looked into this and told me if it makes sense.
Happy to add tests - I didn't find them for this functionality but my guess it should be somewhere in `tests/vcs/git`?

schneebuzz:
> I didn't find them for this functionality but my guess it should be somewhere in `tests/vcs/git`

Some tests for the `..backend.Git.clone` tests are in `tests/integration/test_utils_vcs_git.py`.

radoering:
@jelmer I'd appreciate if you could take a look if it makes sense (from a dulwich point of view).

src/poetry/vcs/git/backend.py
```diff
@@ -313,21 +314,47 @@ def _clone(cls, url: str, refspec: GitRefSpec, target: Path) -> Repo:
...
+            if refspec.is_sha_short:
+                commit = cls._find_object_by_sha_prefix(local, refspec.key.encode())
+                if commit is not None:
+                    logger.debug(
+                        "\nResolved short SHA <c2>%s</c2> as <c2>%s</c2>",
+                        refspec.key,
+                        commit.id,
+                    )
+                    local.refs[b"HEAD"] = commit.id
+                    try:
```
jelmer, line 328: What is this try for? If the object was found in the local repo, why would it be missing here?

mikhainin: Hmm, this is a good question. It probably was me being too cautious. Let me remove it.

radoering:
> Happy to add tests - I didn't find them for this functionality but my guess it should be somewhere in `tests/vcs/git`?

We have tests in [tests/integration/test_utils_vcs_git.py](https://github.com/python-poetry/poetry/blob/main/tests/integration/test_utils_vcs_git.py) and in [main/tests/vcs/git/test_backend.py](https://github.com/python-poetry/poetry/blob/main/tests/vcs/git/test_backend.py). Fast tests that do not require an external repository can be put into the latter.

> I can see some integration tests against https://github.com/python-poetry/test-fixture-vcs-repository but every commit there seems to have its own head/tag which does not allow to reproduce the initial issue...

I have no idea if it is feasible, but maybe we can create a local repository in the test?

jelmer:
Creating a local repository in e.g. a tempdir should be fast and easy with Dulwich - Repo.init(path) will do what you need.

src/poetry/vcs/git/backend.py
jelmer: You've retrieved the commit object, so that will definitely be present. 

One of git's basic invariants (ensured by e.g. fetch, etc) is that if you have the commit object you also have the other objects it refers to*.

[*] This isn't strictly true under some circumstances, e.g. shallows - but that's not relevant here.

src/poetry/vcs/git/backend.py
```diff
@@ -313,21 +314,47 @@ def _clone(cls, url: str, refspec: GitRefSpec, target: Path) -> Repo:
...
+                raise PoetryConsoleError(
+                    f"Failed to clone {url} at '{refspec.key}', verify ref exists on"
+                    " remote."
+                )
 
         return local
 
+    @classmethod
+    def _find_object_by_sha_prefix(cls, repo: Repo, sha_prefix: bytes) -> Commit | None:
+        for sha in repo.object_store:
```
jelmer, line 351: With the upcoming release of Dulwich you can also use repo.object_store.iter_prefix, which avoids scanning the entire repository.

mikhainin: Shall we wait till the next release and use the new method?

mikhainin:
I added test, I had to introduce a new marker "skip_git_mock" not to mock "git clone" as that one is not ready to mock local clone. 

The rest was copied from `tests/vcs/git/test_system.py`: I moved fixtures into separate files so they can be re-used and added a "middle commit" so the issue was finally reproduced.

jelmer:
> I added the test, I had to introduce a new marker "skip_git_mock" not to mock "git clone" as that one is not ready to mock local clone. 
> 
> The rest was copied from `tests/vcs/git/test_system.py`: I moved fixtures into separate files so they can be re-used and added a "middle commit" so the issue was finally reproduced.

LGTM, for what it's worth.

pyproject.toml
```diff
@@ -193,6 +193,7 @@ addopts = [ "-n", "logical", "-ra", "--strict-config", "--strict-markers" ]
 testpaths = ["tests"]
 markers = [
     "network: mark tests that require internet access",
+    "skip_git_mock: makrs test should not auto-apply git_mock"
```
radoering, line 196:
```suggestion
    "skip_git_mock: mark tests that should not auto-apply git_mock"
```

tests/conftest.py
```diff
@@ -578,3 +584,49 @@ def project_context(project: str | Path, in_place: bool = False) -> Iterator[Pat
...
+    repo.stage(["foo"])
+
+    init_commit = repo.do_commit(
+        committer=b"User <user@example.com>",
+        author=b"User <user@example.com>",
+        message=b"init",
+        no_verify=True,
+    )
+
+    # one commit which is not "head
```
radoering, line 605:
```suggestion
    # one commit which is not "head"
```

tests/conftest.py
```diff
@@ -578,3 +584,49 @@ def project_context(project: str | Path, in_place: bool = False) -> Iterator[Pat
...
+    repo.stage(["bar"])
+    middle_commit = repo.do_commit(
+        committer=b"User <user@example.com>",
+        author=b"User <user@example.com>",
+        message=b"extra",
+        no_verify=True,
+    )
+
+    # extra commit
+    (tmp_path / "foo").write_text("foo", encoding="utf-8")
```
radoering, line 616: Is this an empty commit because it writes the same as the first commit? (It probably does not matter, I am just wondering.)

tests/vcs/git/test_backend.py
```diff
@@ -75,3 +82,31 @@ def test_urlpathjoin(url: str, expected_result: str) -> None:
     path = "../other-repo"
     result = urlpathjoin(url, path)
     assert result == expected_result
+
+
+@pytest.mark.skipif(GIT_NOT_INSTALLLED, reason="These tests requires git cli")
```
radoering, line 87: Do they? I thought they just use `dulwich` and do not need git installed (in comparison to `SystemGit`)?

radoering (changes requested):

tests/conftest.py
mikhainin: Thanks for flagging this, I added a new file instead

tests/vcs/git/test_backend.py
mikhainin: I thought the fixture uses `git cli` but now I see it's wasn't the case. Removed this `skipif`

tests/vcs/git/test_backend.py
```diff
@@ -75,3 +82,30 @@ def test_urlpathjoin(url: str, expected_result: str) -> None:
     path = "../other-repo"
     result = urlpathjoin(url, path)
     assert result == expected_result
+
+
+class TestBackendGit:
```
radoering, line 87: I think we do not need this class anymore (without the `skipif`). There are already other tests for the `Git` class, which are top level. (In general, we tend to not create test classes but just test functions.)

tests/vcs/git/git_fixture.py
```diff
@@ -0,0 +1,21 @@
...
+import typing
+
+
+if typing.TYPE_CHECKING:
+    from pathlib import Path
+
+    import dulwich.repo
+
+
+GIT_NOT_INSTALLLED = shutil.which("git") is None
```
radoering, line 13: I think we can put this constant back into `test_system` because it is unlikely that it is relevant for other tests than for `SystemGit`.

tests/conftest.py
```diff
@@ -578,3 +584,49 @@ def project_context(project: str | Path, in_place: bool = False) -> Iterator[Pat
             yield path
 
     return project_context
+
+
+@pytest.fixture()
+def temp_repo(tmp_path: Path) -> TempRepoFixture:
```
radoering, line 590: This fixture is only used in `test_backend` and `test_system`, isn't it? Then, we can put it in a `conftest.py` in `tests/vcs/git`.

radoering (changes requested):

src/poetry/vcs/git/backend.py
radoering: I do not think we have to wait unless @jelmer says there will be a release in the next days / few weeks. (It will take at least some weeks until the next Poetry release and I think it makes sense to include this with or without the improvement.)

tests/conftest.py
mikhainin: Yes. However, it will interfere with Ruff's rules as ruff will be removed as an "unused import".
https://github.com/astral-sh/ruff/issues/3295
https://github.com/astral-sh/ruff/issues/8796

We can mark the import with `noqa` but I wasn't sure it was the best approach

radoering: We do not have to import a fixture from a `conftest.py` to use it or do you mean something else?

tests/vcs/git/test_backend.py
mikhainin: removed the class

tests/vcs/git/git_fixture.py
mikhainin: done

tests/conftest.py
mikhainin: ah, sure, moved to `tests/vcs/git/conftest.py`

src/poetry/vcs/git/backend.py
jelmer: FWIW There's a new release out with this API now.

jelmer:
```suggestion
        for sha in repo.object_store.iter_prefix(sha_prefix):
```

mikhainin: I can see that this is implemented in **dulwich** `0.22.5`, which drops support of Python 3.8.
It is a reasonable change, but this pull-request branch still requires compatibility with `3.8`. I'm not sure it must be changed as a part of this PR...

@radoering?

Secrus: We have dropped Python 3.8 on the main branch already, so that should not be an issue

src/poetry/vcs/git/backend.py
```diff
@@ -313,21 +314,38 @@ def _clone(cls, url: str, refspec: GitRefSpec, target: Path) -> Repo:
...
```
jelmer, line 342:
```suggestion
        for sha in repo.object_store.iter_prefix(prefix):
```

jelmer (approved):

src/poetry/vcs/git/backend.py
radoering: @mikhainin As Secrus said. Just rebase your branch on main. That is required before merging anyway.

mikhainin: Hmm, this direct replace causes:
```

~/src/poetry/src/poetry/vcs/git/backend.py:342: in _find_object_by_sha_prefix
    for sha in repo.object_store.iter_prefix(sha_prefix):
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

self = <DiskObjectStore('/private/var/folders/yq/nvkqzc2s60vdd86cg6_79dgh0000gn/T/pytest-of-mikhailgalanin/pytest-87/popen-gw0/test_short_sha_not_in_head0/test-repo/clone-test/.git/objects')>
prefix = b'895d5f'

    def iter_prefix(self, prefix):
        if len(prefix) < 2:
            yield from super().iter_prefix(prefix)
            return
        seen = set()
        dir = prefix[:2].decode()
        rest = prefix[2:].decode()
>       for name in os.listdir(os.path.join(self.path, dir)):
E       FileNotFoundError: [Errno 2] No such file or directory: '/private/var/folders/yq/nvkqzc2s60vdd86cg6_79dgh0000gn/T/pytest-of/pytest-87/popen-gw0/test_short_sha_not_in_head0/test-repo/clone-test/.git/objects/89'

.../virtualenvs/poetry-GxhiNo8B-py3.11/lib/python3.11/site-packages/dulwich/object_store.py:1042: FileNotFoundError



```

src/poetry/vcs/git/backend.py
```diff
@@ -313,21 +314,38 @@ def _clone(cls, url: str, refspec: GitRefSpec, target: Path) -> Repo:
...
-                refspec.key,
-                e,
-            )
+            short_ref_found = False
 
-            raise PoetryConsoleError(
-                f"Failed to clone {url} at '{refspec.key}', verify ref exists on"
-                " remote."
-            )
+            if refspec.is_sha_short:
```
abn, line 319: @mikhainin can we move the new logic into an `else` block under the for loop here;

https://github.com/python-poetry/poetry/blob/6f6fd7012983a2e749c6030c1f5f155fd4397058/src/poetry/vcs/git/backend.py#L95-L101

And as @jelmer suggested, we can simplify it to something like 

```py
            else:
                for sha in repo.object_store.iter_prefix(prefix):
                    self.revision = sha.decode("utf-8")
                    break
```

Assuming we get a dulwich release with the fix soon.

abn (changes requested):

src/poetry/vcs/git/backend.py
jelmer: I'll see if I can do a release this weekend,

abn:
This will likely also require #9849 if we intend to update dulwich.

src/poetry/vcs/git/backend.py
jelmer: 0.22.6 is now out with the fix

dimbleby:
> This will likely also require #9849 if we intend to update dulwich.

Since recent dulwich releases are bumping only the patch number, poetry users are already getting the latest dulwich.

Are you saying that dulwich has released a breaking change in a patch release?

src/poetry/vcs/git/backend.py
dimbleby: No it isn't!  I guess your release pipeline failed?

jelmer:
I'm curious what change is, we haven't had reports of API breakages.

src/poetry/vcs/git/backend.py
mikhainin: Right, I moved the logic to `_normalise()` - it now looks even cleaner.
Waiting for **0.22.6** to appear and then will push the change.

abn (approved):

src/poetry/vcs/git/backend.py
Secrus: Dulwich 0.22.6 is out

abn:
@jelmer To close the loop regarding @dimbleby 's question; I have raised a dulwich issue for triage with a reproducer.

Cross-post: https://github.com/jelmer/dulwich/issues/1443

radoering (approved):

Issue #9749, opened 2024-10-10
`poetry show [package]` 'required by' output is confusing


elliot-100:
### Description

Here's an example.

Extract from my project's `poetry.lock`:
```
[[package]]
name = "playwright"
version = "1.47.0"
description = "A high-level API to automate web browsers"
optional = false
python-versions = ">=3.8"
files = [--snipped--]

[package.dependencies]
greenlet = "3.0.3"
pyee = "12.0.0"
```

```
> poetry show greenlet
 name         : greenlet                                      
 version      : 3.0.3                                         
 description  : Lightweight in-process concurrent programming 

required by
 - playwright 3.0.3
```
Playwright's latest version is 1.47; it looks like the locked version of greenlet is being shown here instead.

Notes:
- Re 'How did you install Poetry?' I am not sure of the answer - is there a way to verify?
- I will try to check on other platforms

### Workarounds

Examining `poetry.lock` and not trusting `poetry show [package]`.

### Poetry Installation Method

install.python-poetry.org

### Operating System

Windows 11

### Poetry Version

1.8.3

### Poetry Configuration

```bash session
cache-dir = "C:\\Users\\--user-redacted--\\AppData\\Local\\pypoetry\\Cache"
experimental.system-git-client = false
installer.max-workers = null
installer.modern-installation = true
installer.no-binary = null
installer.parallel = true
keyring.enabled = true
repositories.testpypi.url = "https://test.pypi.org/legacy/"
solver.lazy-wheel = true
virtualenvs.create = true
virtualenvs.in-project = null
virtualenvs.options.always-copy = false
virtualenvs.options.no-pip = false
virtualenvs.options.no-setuptools = false
virtualenvs.options.system-site-packages = false
virtualenvs.path = "{cache-dir}\\virtualenvs"  # C:\Users\--user-redacted--\AppData\Local\pypoetry\Cache\virtualenvs
virtualenvs.prefer-active-python = false
virtualenvs.prompt = "{project_name}-py{python_version}"
warnings.export = false
```
```


### Python Sysconfig

```bash session
Platform: "win-amd64"
Python version: "3.12"
Current installation scheme: "venv"

Paths: 
        data = "C:\Users\--user-redacted--\AppData\Local\pypoetry\Cache\virtualenvs\britishcycling-clubs-_kmXzTV9-py3.12"
        include = "C:\Users\--user-redacted--\AppData\Local\Programs\Python\Python312\Include"
[...]

PR #9750, opened 2024-10-10
feat: in `show {package}` output, clarify meaning of 'required by' entries

elliot-100:
# Pull Request Check List

Resolves: #9749

- [x] Updated **documentation** for changed code

Example of actual output before:
```console
> poetry show requests
 name         : requests
 version      : 2.32.3
 description  : Python HTTP for Humans.

dependencies
 - certifi >=2017.4.17
 - charset-normalizer >=2,<4
 - idna >=2.5,<4
 - urllib3 >=1.21.1,<3

required by
 - cachecontrol >=2.16.0
```

After: 
```console 
> poetry show requests
 name         : requests
 version      : 2.32.3
 description  : Python HTTP for Humans.

dependencies
 - certifi >=2017.4.17
 - charset-normalizer >=2,<4
 - idna >=2.5,<4
 - urllib3 >=1.21.1,<3

required by
 - cachecontrol requires requests >=2.16.0
```

### DIFF
--- a/docs/cli.md
+++ b/docs/cli.md
@@ -515,7 +515,7 @@ dependencies
  - pytzdata >=2017.2.2
 
 required by
- - calendar requires pendulum >=1.4.0
+ - calendar requires >=1.4.0
 ```
 
 ### Options
--- a/docs/cli.md
+++ b/docs/cli.md
@@ -515,7 +515,7 @@ dependencies
  - pytzdata >=2017.2.2
 
 required by
- - calendar >=1.4.0
+ - calendar requires pendulum >=1.4.0
 ```
 
 ### Options
--- a/src/poetry/console/commands/show.py
+++ b/src/poetry/console/commands/show.py
@@ -200,7 +200,9 @@ lists all packages available."""
             self.line("")
             self.line("<info>required by</info>")
             for parent, requires_version in required_by.items():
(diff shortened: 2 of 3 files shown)


elliot-100:
pre-commit.ci autofix

docs/cli.md
```diff
@@ -515,7 +515,7 @@ dependencies
  - pytzdata >=2017.2.2
 
 required by
- - calendar >=1.4.0
+ - calendar requires pendulum >=1.4.0
```
radoering, line 518:
```suggestion
 - calendar requires >=1.4.0
```

Should we omit the package name? (It will be the same in each line if it is required by multiple packages.)

@Secrus What is your opinion?

Secrus: Yeah, I think it's fine without the package name. Maybe we can add some explanation for this output in the docs too?

elliot-100: OK, I've reverted the addition of the package name, so the PR now just adds the word 'requires' for each item in `Required by`.

I am not sure what I would add to docs for this change - there's already amended sample output, and now the output is more self-explanatory.

(I do think `show` documentation can be improved overall and I am drafting PR #9752 on that - I think I'm going to reduce scope though)

Secrus (approved):

PR #9763, opened 2024-10-12
Replace `poetry shell` command with `poetry env activate`

Secrus:
# Pull Request Check List

Resolves: personal grudge towards `poetry shell`


src/poetry/console/commands/env/activate.py
```diff
@@ -0,0 +1,64 @@
...
+        if (activation_script := env.bin_dir / filename).exists():
+            if WINDOWS:
+                return f"{self.quote(str(activation_script), shell)}"
+            return f"{command} {self.quote(str(activation_script), shell)}"
+        return ""
+
+    @staticmethod
+    def quote(command: str, shell: str) -> str:
+        if shell in ["powershell", "pwsh"] or WINDOWS:
+            return "{}".format(command.replace("'", "''"))
```
radoering, line 63: I suppose this should handle paths with `'`? However, it does not seem to work (on Windows) and it also does not work for paths with spaces (which is far more common and thereby more important).

On Windows, for `cmd` you just have to put the command in double quotes (`f'"{command}"'`) and for Powershell you need something like `f'& "{command}"'`.

tests/console/commands/env/test_activate.py
```diff
@@ -0,0 +1,74 @@
...
+
+    line = tester.io.fetch_output().split("\n")[0]
+    assert line.startswith(command)
+    assert line.endswith(f"activate{ext}")
+
+
+@pytest.mark.parametrize(
+    "shell, command, ext",
+    (
+        ("pwsh", ".", "Activate.ps1"),
```
radoering, line 55:
```suggestion
        ("cmd", ".", "activate"),
        ("pwsh", ".", "Activate.ps1"),
```

src/poetry/console/commands/env/activate.py
```diff
@@ -0,0 +1,64 @@
...
+
+        if command := self.get_activate_command(env):
+            self.line(command)
+            return 0
+        else:
+            raise ShellNotSupportedError(
+                "Discovered shell doesn't have an activator in virtual environment"
+            )
+
+    def get_activate_command(self, env: Env) -> str:
```
radoering, line 38:
```suggestion
    def _get_activate_command(self, env: Env) -> str:
```

I'd add an underscore to make clear it is internal.

src/poetry/console/commands/env/activate.py
```diff
@@ -0,0 +1,64 @@
...
+            command, filename = "source", "activate"
+
+        if (activation_script := env.bin_dir / filename).exists():
+            if WINDOWS:
+                return f"{self.quote(str(activation_script), shell)}"
+            return f"{command} {self.quote(str(activation_script), shell)}"
+        return ""
+
+    @staticmethod
+    def quote(command: str, shell: str) -> str:
```
radoering, line 61:
```suggestion
    def _quote(command: str, shell: str) -> str:
```

I'd add an underscore to make clear it's internal

docs/cli.md
```diff
@@ -648,29 +648,6 @@ poetry run my-script
 
 Note that this command has no option.
 
-## shell
```
radoering, line 651: Maybe, we should not remove it (now) but write that it has been outsourced into a plugin and we recommend to use `env activate` instead. (We did not remove `export` from the docs. I hope this will save us some support requests because some people will look here.)

docs/managing-environments.md
```diff
@@ -76,6 +76,41 @@ special `system` Python version to retrieve the default behavior:
 poetry env use system
 ```
 
+## Activating the environment
+
+`poetry env activate` command prints the activate command to the console. This way you won't leave the current shell.
+You can then feed the output to `eval` to activate the environment. This way is the closest to manually activating the environment.
```
radoering, line 82: I would restructure the text a bit and do not refer to `eval` but "the eval command of your shell" (or similar). Proposal:

```suggestion
The `poetry env activate` command prints the activate command of the virtual environment to the console.
You can run the output command manually or feed it to the eval command of your shell to activate the environment.
This way you won't leave the current shell.

File: docs/managing-environments.md [hunk 7]
```diff
@@ -76,6 +76,41 @@ special `system` Python version to retrieve the default behavior:
...
+
+{{% note %}}
+Looking for `poetry shell`? It was moved to a plugin: [`poetry-plugin-shell`](https://github.com/python-poetry/poetry-plugin-shell)
+{{% /note %}}
```
radoering, line 86: I'd either put this at the beginning or the end and not between the text and the examples how to feed the output to the eval command of different shells.

docs/managing-environments.md
```diff
@@ -76,6 +76,41 @@ special `system` Python version to retrieve the default behavior:
...
+
+{{< tabs tabTotal="3" tabID1="bash-csh-zsh" tabID2="fish" tabID3="powershell" tabName1="Bash/Zsh/Csh" tabName2="Fish" tabName3="Powershell" >}}
+
+{{< tab tabID="bash-csh-zsh" >}}
+
+```bash
+$ eval $(poetry env activate)
+(test-project-for-test) $  # Virtualenv entered
```
radoering, line 94: This line is missing for `fish` and `powershell`.

radoering (changes requested):

src/poetry/console/commands/env/activate.py
Secrus: I changed it like you said. I don't have access to a windows machine, so if you could test it on yours I would appreciate it

radoering (approved):

Issue #9769, opened 2024-10-14
Problems with wheel_file_re


AlekseyLobanov:
### Description

Hello.

I found that performance of the current `wheel_file_old` may be really bad O(N^4) on some inputs.  It may be not a serious problem but for end user it will be almost impossible to find and troubleshoot.

After performance fix I also found `.dist-info` extension is working as expected.

We can verify performance with this POC

```python
import re
import time
from contextlib import contextmanager
import sys

wheel_file_old = re.compile(
    r"^(?P<namever>(?P<name>.+?)-(?P<ver>\d.*?))"
    r"(-(?P<build>\d.*?))?"
    r"-(?P<pyver>.+?)"
    r"-(?P<abi>.+?)"
    r"-(?P<plat>.+?)"
    r"\.whl|\.dist-info$",
    re.VERBOSE,
)
wheel_file_re_new= re.compile(
    r"^(?P<namever>(?P<name>[^-]+)-(?P<ver>[^-]*))"
    r"(-(?P<build>\d[^-]*))?"
    r"-(?P<pyver>[^-]+)"
    r"-(?P<abi>[^-]+)"
    r"-(?P<plat>[^-]+)"
    r"(?:\.whl|\.dist-info)$",
    re.VERBOSE,
)


@contextmanager
def duration(text):
    begin_at = time.monotonic()
    yield
    delta = time.monotonic() - begin_at
    print(f"{text:20} takes {delta:0.6f}s")


pattern = '0-0' + '-' * int(sys.argv[1])

with duration("Current (old) re"):
    wheel_file_old.match(pattern)
with duration("New re"):
    wheel_file_re_new.match(pattern)
```

Example output
```
➜ python poetry-poc.py 250
Current (old) re     takes 0.942238s
New re               takes 0.000003s
```

### Workarounds

- Performance issues should not be a problem for almost all valid wheel names
- Do not use wheels with `.dist-info` extension

### Poetry Installation Method

pip

### Operating System

Ubtuntu

### Poetry Version

2.0.0.dev0

### Poetry Configuration

```bash session
cache-dir = "/home/alex/.cache/pypoetry"
experimental.system-git-client = false
installer.max-workers = null
installer.no-binary = null
installer.only-binary = null
installer.parallel = true
keyring.enabled = true
requests.max-retries = 0
solver.lazy-wheel = true
virtualenvs.create = true
virtualenvs.in-project = null
virtualenvs.options.always-copy = false
[...]

PR #9770, opened 2024-10-14
Fix wheel_file_re RegExp

AlekseyLobanov:
# Pull Request Check List

Resolves: #9769

- For performance POC check related issue
- All interfaces are the same
- It covers two problems but I believe that they are close enough (few characters in the single RE)

### DIFF
--- a/tests/utils/test_patterns.py
+++ b/tests/utils/test_patterns.py
@@ -42,6 +42,12 @@ def test_wheel_file_re(filename: str, expected: dict[str, str | None]) -> None:
     assert groups == expected
 
 
+@pytest.mark.parametrize("filename", [".dist-info"])
+def test_wheel_file_re_not_matches(filename: str) -> None:
+    match = patterns.wheel_file_re.match(filename)
+    assert match is None
+
+
 @pytest.mark.parametrize(
     ["filename", "expected"],
     [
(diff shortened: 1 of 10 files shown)


dimbleby:
I have no idea why the code thinks that `dist-info` is a valid wheel extension, it probably is just legacy wrongness that should be removed.

If it has never even worked in all these years anyway, that's pretty good evidence that it is safe to remove it.

radoering (dismissed):
I agree. Let's remove the `dist-info` part.

AlekseyLobanov:
I removed `.dist-info` extension from RE. No problems with tests.

tests/utils/test_patterns.py
```diff
@@ -42,6 +54,12 @@ def test_wheel_file_re(filename: str, expected: dict[str, str | None]) -> None:
     assert groups == expected
 
 
+@pytest.mark.parametrize("filename", [".dist-info"])
+def test_wheel_file_re_not_matches(filename: str) -> None:
+    match = patterns.wheel_file_re.match(filename)
+    assert match is None
+
+
```
radoering, line 62: This test is no longer of much use, is it? Thus, I think we can remove it.

AlekseyLobanov: Yes, agree. Test has been removed.

Secrus (approved):

Issue #9782, opened 2024-10-19
Document PEP 440 (Compatible Release) specifiers


srittau:
### Issue Kind

Missing documentation

### Existing Link

https://python-poetry.org/docs/dependency-specification/

### Description

poetry seems to support [PEP 440 "Compatible Release"](https://peps.python.org/pep-0440/#compatible-release) specifiers. Would you accept a PR add a section about them?

radoering:
> Would you accept a PR add a section about them?

I do not see why we should not. 😃

PR #9783, opened 2024-10-19
Document compatible release requirements

srittau:
# Pull Request Check List

Resolves: #9782

### DIFF
--- a/docs/dependency-specification.md
+++ b/docs/dependency-specification.md
@@ -109,6 +109,20 @@ Here are some more examples of caret requirements and the versions that would be
 | ^0.0        | >=0.0.0 <0.1.0   |
 | ^0          | >=0.0.0 <1.0.0   |
 
+### Compatible release requirements
+
+**Compatible release requirements** specify a minimal version with the ability to update to later versions of the same level.
+For example, if you specify a major, minor, and patch version, only patch-level changes are allowed.
+If you only specify a major, and minor version, then minor- and patch-level changes are allowed.
+
+`~=1.2.3` is an example of a tilde requirement.
+
+| Requirement | Versions allowed |
+| ----------- | ---------------- |
+| ~=1.2.3     | >=1.2.3 <1.3.0   |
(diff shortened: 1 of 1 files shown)


docs/dependency-specification.md
```diff
@@ -109,6 +109,20 @@ Here are some more examples of caret requirements and the versions that would be
...
+For example, if you specify a major, minor, and patch version, only patch-level changes are allowed.
+If you only specify a major, and minor version, then minor- and patch-level changes are allowed.
+
+`~=1.2.3` is an example of a tilde requirement.
+
+| Requirement | Versions allowed |
+| ----------- | ---------------- |
+| ~=1.2.3     | >=1.2.3 <1.3.0   |
+| ~=1.2       | >=1.2.0 <2.0.0   |
+| ~=1         | >=1.0.0          |
```
radoering, line 124: That is neither what the standard says nor what we do:

> This operator MUST NOT be used with a single segment version number such as ~=1.

```
$ from poetry.core.constraints.version import parse_constraint
$ str(parse_constraint("~=1"))
'>=1,<1.1'
```

I think we should drop this example and fix our parser.

srittau: Could catch, I missed that.

docs/dependency-specification.md
```diff
@@ -109,6 +109,19 @@ Here are some more examples of caret requirements and the versions that would be
 | ^0.0        | >=0.0.0 <0.1.0   |
 | ^0          | >=0.0.0 <1.0.0   |
 
+### Compatible release requirements
+
+**Compatible release requirements** specify a minimal version with the ability to update to later versions of the same level.
+For example, if you specify a major, minor, and patch version, only patch-level changes are allowed.
+If you only specify a major, and minor version, then minor- and patch-level changes are allowed.
+
+`~=1.2.3` is an example of a tilde requirement.
```
radoering, line 118:
```suggestion
`~=1.2.3` is an example of a compatible release requirement.
```

PR #9786, opened 2024-10-20
feat(config)!: prefer active python by default

finswimmer:
From now on Poetry prefers the currently active Python by default. To restore the old behavior the option `virtualenvs.use-poetry-python` was introduced and must be set to `true` (default: `false`).

`virtualenvs.prefer-active-python` is removed as an option.

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -423,9 +423,11 @@ This setting controls the global virtual environment storage path. It most likel
 
 **Type**: `boolean`
 
-**Default**: `false`
+**Default**: `true`
 
 **Environment Variable**: `POETRY_VIRTUALENVS_PREFER_ACTIVE_PYTHON`
+
+*Default changed to `true` in 2.0.0*
 
 *Introduced in 1.2.0*
 
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -58,8 +58,8 @@ virtualenvs.options.always-copy = true
 virtualenvs.options.no-pip = false
 virtualenvs.options.system-site-packages = false
 virtualenvs.path = "{cache-dir}/virtualenvs"  # /path/to/cache/directory/virtualenvs
-virtualenvs.prefer-active-python = false
 virtualenvs.prompt = "{project_name}-py{python_version}"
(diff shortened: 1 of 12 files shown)


Secrus:
I was actually considering if we could remove that option or invert it so that one can choose to use system Python instead of the active one. That seems more instinctive behaviour than what we have currently

finswimmer:
> I was actually considering if we could remove that option or invert it 

I wouldn't remove it, but inverting the default would be fine (But I'm biased here, because that's the config I change first in an new installation ;)

Secrus:
Well, we could take advantage of the fact that we are making a breaking release and change that to `use-system-python` that would force usage of the Python version Poetry was installed with. That would give us an open interface to introduce `findpython` backend without breaking changes but as an alternative (until deemed stable).

finswimmer:
I'm fine with renaming, but I think `use-system-python` is confusing for the users. While it is consistent with the code base, I guess most users would expect that the Python of their system is meant. Maybe `use-application-python` or `use-poetry-python`?

Secrus:
> I'm fine with renaming, but I think `use-system-python` is confusing for the users. While it is consistent with the code base, I guess most users would expect that the Python of their system is meant. Maybe `use-application-python` or `use-poetry-python`?

I am fine with `use-poetry-python` and a proper docs entry explaining that.

Secrus (dismissed):

docs/configuration.md
```diff
@@ -445,6 +432,19 @@ If set to `false`, Python version used during Poetry installation is used.
...
+**Type**: `boolean`
+
+**Default**: `false`
+
+**Environment Variable**: `POETRY_VIRTUALENVS_USE_POETRY_PYTHON`
+
+*Introduced in 2.0.0*
+
+By default, Poetry will use the activated Python version to create a new virtual environment.
+If set to `true`, Python version used during Poetry installation is used.
```
radoering, line 446:
```suggestion
If set to `true`, the Python version used during Poetry installation is used.
```

Secrus (dismissed):

radoering (approved):

PR #9787, opened 2024-10-21
feat(config)!: rename experimental.system-git-client to system-git-client

finswimmer:
`experimental.system-git-client` was introduced in 1.2.0. Let's stop calling it "experimental".

### DIFF
--- a/src/poetry/config/config.py
+++ b/src/poetry/config/config.py
@@ -119,9 +119,6 @@ class Config:
             "prefer-active-python": False,
             "prompt": "{project_name}-py{python_version}",
         },
-        "experimental": {
-            "system-git-client": False,
-        },
         "requests": {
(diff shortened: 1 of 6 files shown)


src/poetry/config/config.py
Secrus, line 1: I wonder if we could add some feature to config to make the transition easier for users. Right now, if someone has the "old" version of the option, they would have to manually set the new option too. Any ideas?

finswimmer: I thought about it as well. Because I couldn't come up immediately with a solution, I decided to say: It's a major version bump with breaking changes and users have to check there config for necessary changes.

If we have strong feelings about the user experience here, I have to think about it more deeply.

Secrus (approved):

PR #9805, opened 2024-10-28
Add support for system truststore

### LINKED ISSUE #9249: Add truststore support to use system certificate store

Secrus:
# Pull Request Check List

Resolves: #9249


src/poetry/console/application.py
```diff
@@ -342,6 +344,15 @@ def _load_plugins(self, io: IO | None = None) -> None:
...
         self._plugins_loaded = True
 
+    @staticmethod
+    def _load_system_truststore() -> None:
+        from poetry.utils.ssl_truststore import is_truststore_enabled
+
+        if is_truststore_enabled():
+            import truststore
+
+            truststore.inject_into_ssl()
```
abn, line 354: shouldn't this be wrapped in try/catch for ImportError? We should not need the `ssl_truststore` module.

abn (changes requested):
You also might want to update the `http` fixture in `tests/conftest.py` to be something like this:

```py
@pytest.fixture
def http(mocker: MockerFixture) -> Iterator[type[httpretty.httpretty]]:
    mocker.patch("truststore.inject_into_ssl")
    httpretty.reset()
    with httpretty.enabled(allow_net_connect=False, verbose=True):
        yield httpretty
```

src/poetry/config/config.py
```diff
@@ -135,6 +135,7 @@ class Config:
         "keyring": {
             "enabled": True,
         },
+        "system-truststore": True,
```
abn, line 138: Default to `False`.

src/poetry/console/application.py
Secrus: As you can see, there are several checks done in `ssl_truststore` module. I wanted to isolate it and avoid cluttering the `Application` with it.

src/poetry/config/config.py
Secrus: I can agree, but if we go with `False`, we should set a date for when it goes `True` by default.

src/poetry/console/application.py
abn: As I understood it those checks are redundant. If truststore cannot be used an import error will be raised.

src/poetry/config/config.py
Secrus: Switched and added a TODO with a flip date proposal.

src/poetry/console/application.py
Secrus: Simple `ImportError` is too little in my taste. The `_is_truststore_available()` function has various checks and logs a proper message.

abn: It might be, but it does not add anything in this context.

https://truststore.readthedocs.io/en/latest/

> If Truststore can’t work for a given platform due to APIs not being available then at import time the exception ImportError will be raised with an informative message

The only case where it makes sense is if we need to programmatically check if truststore is available. If that is the case, we might not want `inject_into_ssl` anyway.

Also note;
> The call to truststore.inject_into_ssl() should be called as early as possible in your program as modules that have already imported ssl.SSLContext won’t be affected.

This means this might be too late as imports to repos and authenticator would have already happened.

ikappaki:
HI @Secrus, Is there much left to address in this PR or would you like any assistance? I just wanted to give a gentle nudge since it's fundamental for expanding Poetry's support to a wider range of environments.

thanks

Secrus:
@ikappaki what is left is the testing. There is no test to check which truststore is used and I didn't have time to come up with one.

Issue #8761, opened 2023-12-08
Keyring access requested although non-publishing operations do not need any credentials.


real-yfprojects:
- **Poetry version**:  1.8.0.dev0 (6f9de73)
- **Python version**: 3.8.10
- **OS version and name**: Kubuntu
- **pyproject.toml**: Any minimal working configuration, no dependencies are required.

- [x] I am on the [latest](https://github.com/python-poetry/poetry/releases/latest) stable Poetry version, installed using a [recommended method](https://python-poetry.org/docs/#installation).
- [x] I have searched the [issues](https://github.com/python-poetry/poetry/issues) of this repo and believe that this is not a duplicate.
- [x] I have consulted the [FAQ](https://python-poetry.org/docs/faq/) and [blog](https://python-poetry.org/blog/) for any relevant entries or release notes.
- [x] If an exception occurs when executing a command, I executed it again in debug mode (`-vvv` option) and have included the output below.

## Issue

In its essence this is the same as #1917 and closely linked to #8623.

### Can you summarize the problem?
For any http request made by poetry it will hit the keyring for possible credentials to send along (presumably to support private repositories and registries). However this has the consequents that poetry needs keyring access for almost all operations although most operations do not require any credentials. This is bad. Quoting from https://github.com/python-poetry/poetry/issues/1917#issue-551829162:

> I think people should be careful about granting programs access to things they should not need to access. This behavior is an example of a problem that makes that hard to achieve.

### Why was #1917 closed?

Very good question. The maintainers apparently thought fixing the symptoms would also fix the underlying problem. See https://github.com/python-poetry/poetry/pull/8078#issuecomment-1584169367 and #8227. As a matter of fact the underlying issue now reappears in a different disguise.

### What's the new symptom?

#8227 fixed the problem where poetry just exited when the keyring access was denied. In the current version it ignores the denial and sends the http request without credentials.
However for the next http request it will try to access the keyring again. For a command like `poetry add pytest` this will result in uncountable popups requesting keyring access that one has to deny (hit *cancel* in my case) one after another. This makes poetry unusable without granting keyring access. It is safe to say that #8227 practically didn't fix anything.

### What would be an actual fix?

I can already imagine the maintainers putting forward fixes for this new problem instead of fixing the underlying issue. 
The sane way of fixing this goes like that:
[...]

Issue #8623, opened 2023-11-04
Regression: Poetry 1.7 hangs instead of asking to unlock keyring


ilyagr:
- **Poetry version**: 1.7.0
- **Python version**:  3.11.6
- **OS version and name**: Debian trixie (testing) running in a Chrome OS Linux container
- **pyproject.toml**: https://gist.github.com/ilyagr/1bebabebc93e6662d139009d19ed8c1b/edit (includes poetry.lock and detailed command-line output as well).

- [x] I am on the [latest](https://github.com/python-poetry/poetry/releases/latest) stable Poetry version, installed using a recommended method.
- [x] I have searched the [issues](https://github.com/python-poetry/poetry/issues) of this repo and believe that this is not a duplicate.
- [x] I have consulted the [FAQ](https://python-poetry.org/docs/faq/) and [blog](https://python-poetry.org/blog/) for any relevant entries or release notes.
- [x] If an exception occurs when executing a command, I executed it again in debug mode (`-vvv` option) and have included the output below.

## Issue

This is a follow-up to https://github.com/python-poetry/poetry/pull/1917 and https://github.com/python-poetry/poetry/pull/8227. I also mentioned this briefly in
https://github.com/python-poetry/poetry/issues/1917#issuecomment-1793322478

The bug happens when I run either `poetry install` or `poetry lock`.

- **Poetry 1.6.1 behavior**: a Gnome (gnome-keyring or seahorse, I'm unsure) window opens and asks to unlock a keyring. If I click cancel (refuse to unlock the keyring), poetry fails with an exception as described in #8227.

- **Current Poetry 1.7 behavior**: no Gnome window opens, poetry just freezes for a while.
 
  With Poetry 1.7, I also tried rebooting. After a reboot, a Gnome window does appear. After clicking cancel (refuse to unlock the keyring), it reappears. After clicking cancel ~10 more times, it stops reappearing and hangs again.

`rm -r ~/.cache/pypoetry` as suggested in https://github.com/python-poetry/poetry/issues/1917#issuecomment-1770241003 did not help, with or without reinstalling `poetry` (via `pipx`, in my case)

## Expected behavior

poetry does not try to unlock the keyring unless it is performing an operation that can't succeed without access to the keyring (e.g., already failed without access to the keyring).

[...]

PR #9820, opened 2024-11-01
disable keyring per default and only install via an extra

radoering:
# Pull Request Check List

Resolves: #8623
Resolves: #8761

### disable keyring per default

I think the majority of Poetry users do not use/need keyring. Therefore, changing the default of `keyring.enabled` from `true` to `false` avoids issues like #8623 and #8761. In case, someone stores credentials with keyring disabled/unavailable, we have already warned that credentials will be stored in plain text. I extended this warning so it tells the user that they should consider using keyring.

### keyring as an extra

Further, I moved keyring to an extra because it is responsible for one third of our dependencies on Linux (and still a quarter on Windows)! Without keyring, there are 13 dependencies less on Linux and 9 dependencies less on Windows.

#### Why care about the number of dependencies?

The installation of Poetry is about 1-2 seconds faster and about 18 MB smaller without keyring. Maybe not much, but especially installation speed might still be relevant (at least nice) in CI pipelines.

#### And the other side of the coin?

We have to make sure that Poetry still works (and keeps working) without keyring. There must not be an accidental use of keyring anywhere in the code!

I have taken a few measures that should ensure that no accidental use of keyring slips in in the future:
- disallow global keyring imports in `src` via [TID253](https://docs.astral.sh/ruff/rules/banned-module-level-imports/)
- restrict keyring imports to one module (`password_manager.py`) via [TID251](https://docs.astral.sh/ruff/rules/banned-api/)
- extended `test_password_manager.py::test_disabled_keyring_never_called` to `test_password_manager.py::test_disabled_or_unavailable_keyring_never_called`


radoering:
> I am a bit conflicted here.

I can fully understand. It took me a while to get to the point where I thought it is better to deactivate it per default.

> The question I guess, is if security considerations are getting in the way of ux without much value gained.

I think we have reached this point because users run into keyring issues even when installing from sources without any authentication. Especially less experienced / new users are not able to identify and work around this issue quickly.

> The other rationales, like the dependency numbers and sizes etc, I do not think really applies here.

Less dependencies -> less issues. 😉

I still like the extra but this is not a hill I want to die on.

@Secrus If I remember correctly, you are in the "reduce number of dependencies" team. Do you have a strong opinion on this one?

radoering:
Closing as even the weakened variant #9866 does not reach consensus.

PR #9830, opened 2024-11-05
feat(cli): add option to migrate outdated configs

finswimmer:
This PR add a new option `--migrate` to `poetry config`. This option enables users to migrate explicit set configs that have been renamed or changed in other ways in new Poetry versions.

image

Some question:

Only `FileConfigSource` has been updated in that way, that if a property is removed, it is checked whether the section to which this property belongs is now empty and should be removed from the file. Do we want the same behavior for `DictConfigSource`?

It's probably possible to detect on runtime if the Config contains outdated options. Should we inform users, that they should run `poetry config --migrate` to fix it? (I thinks if yes, I will do it in a separate PR.)


abn:
@finswimmer can/should we add a confirmation step? Alternatively, preserve the old file with a new suffix (`.old`)?

finswimmer:
@abn I've added a confirmation step.

Secrus (approved):

Issue #7897, opened 2023-05-09
Actually switch directory with --directory/-C


brandonchinn178:
- [x] I have searched the [issues](https://github.com/python-poetry/poetry/issues) of this repo and believe that this is not a duplicate.
- [x] I have searched the [FAQ](https://python-poetry.org/docs/faq/) and general [documentation](https://python-poetry.org/docs/) and believe that my question is not already covered.

## Feature Request

Previous issues have been raised asking about the behavior of `--directory`: (https://github.com/python-poetry/poetry/issues/7363, https://github.com/python-poetry/poetry/issues/7507). They have both been closed as "working as expected", so this issue is an official request to change the behavior OR change the flag ~OR update the docs~.

Most GNU tools like `make` or `git` will make `git -C dir ...` equivalent to `(cd dir && git ...)`. Using the same flag implies the same behavior, when it's not currently the case:

> this is expected. Poetry just loads the pyproject.toml from the given directory and does not switch to the directory.
> https://github.com/python-poetry/poetry/issues/7363#issuecomment-1386833452

Ideally, the behavior would be changed to actually change the directory; I don't see why a simple addition of `os.chdir` in the beginning of the CLI main function would break things.

dimbleby:
You don't need to make an "official request" to fix the docs, just submit a merge request - make it happen yourself!

(Reporters in both of the issues that you linked were also invited to fix the docs)

brandonchinn178:
I know the issues were closed as "working-as-designed", but I don't actually think that's the case. The original PR (https://github.com/python-poetry/poetry/pull/6810) says it closes https://github.com/python-poetry/poetry/issues/2179, but that issue is asking for the same behavior as this issue, which isn't solved by the current behavior of `-C`

Makman2:
Imo it should not change the directory, because this would defy the whole option at all. It is especially useful to run a script entrypoint via `poetry run your-script` from a different folder and still be able to pass files relative to your current directory as arguments that are not relative to the pyproject.toml's directory.

brandonchinn178:
@Makman2 I assume you meant `poetry -C ../other/dir/ run entry-point`?

I don't think that should be a supported usecase. If you're using another poetry project as a tool for another project, you should add it as a dependency, which will register the entrypoint in the second project so you can just do `poetry run entry-point project2-file` in the second project

brandonchinn178:
[...]

PR #9831, opened 2024-11-06
feat(cli)!: actually switch directory with --directory/-C 

finswimmer:
# Pull Request Check List

`--directory` now switch the current working folder, so that relative paths are resolved relative to the given folder.

To retain the old behavior the `--project` flag is introduced.

Resolves: #7897

### DIFF
--- a/docs/cli.md
+++ b/docs/cli.md
@@ -29,7 +29,8 @@ then `--help` combined with any of those can give you more information.
 * `--no-interaction (-n)`: Do not ask any interactive question.
 * `--no-plugins`: Disables plugins.
 * `--no-cache`: Disables Poetry source caches.
-* `--directory=DIRECTORY (-C)`: The working directory for the Poetry command (defaults to the current working directory).
+* `--directory=DIRECTORY (-C)`: The working directory for the Poetry command (defaults to the current working directory). All command-line arguments will be resolved relative to the given directory.
+* `--project=PROJECT (-P)`: Specify another path as the project root. All command-line arguments will be resolved relative to the current working directory or directory specified using `--directory` option if used.
 
 
 ## new
--- a/src/poetry/console/application.py
+++ b/src/poetry/console/application.py
@@ -20,6 +20,7 @@ from cleo.formatters.style import Style
 from poetry.__version__ import __version__
 from poetry.console.command_loader import CommandLoader
 from poetry.console.commands.command import Command
+from poetry.utils.helpers import directory
 
 
 if TYPE_CHECKING:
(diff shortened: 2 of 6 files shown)


radoering (commented):
Can we add some sort of test for the new option?

finswimmer:
> Can we add some sort of test for the new option?

I really would like to add some tests. But until now I wasn't able to do so because I don't know how.

The scenario I have in mind goes like this:

* install a Poetry project with a `tool.poetry.scripts` entrypoint
* This entrypoint simply prints out the current working directory
* run `poetry run` from outside the project folder with either `--directory` or `--project` and compare the output of the script with the expected

So any help is appreciated :smiley:

finswimmer:
Thanks @radoering. :+1:  I played around a while, but due to the creation of venv I find it to complicated. In finally choose the `version` command to demonstrate what should happen when using the parameters.

docs/cli.md
```diff
@@ -29,7 +29,8 @@ then `--help` combined with any of those can give you more information.
 * `--no-interaction (-n)`: Do not ask any interactive question.
 * `--no-plugins`: Disables plugins.
 * `--no-cache`: Disables Poetry source caches.
-* `--directory=DIRECTORY (-C)`: The working directory for the Poetry command (defaults to the current working directory).
+* `--directory=DIRECTORY (-C)`: The working directory for the Poetry command (defaults to the current working directory). All command-line arguments will be resolved relative to the given directory.
+* `--project=PROJECT (-P)`: Specify another path as the project root. All command-line arguments will be resolved relative to the current working directory.
```
abn, line 33:
```suggestion
* `--project=PROJECT (-P)`: Specify another path as the project root. All command-line arguments will be resolved relative to the current working directory or directory specified using `--directory` option if used.
```

finswimmer: :ok_hand:

src/poetry/console/application.py
```diff
@@ -47,6 +50,17 @@ def _load() -> Command:
     return _load
 
 
+@contextlib.contextmanager
+def switch_working_directory(path: Path) -> Iterator[Path]:
```
Secrus, line 54: There is already `poetry.utils.helpers:directory` for that. However, I would consider changing the name of the helper to `chdir` or to your function name to make it more meaningful.

Secrus (commented):
What is the expected behaviour when user passes both `--project` and `--directory`? I feel like there should be some more info on that matter, it doesn't feel intuitive right now and the docs are quite vague on the difference between those two.

radoering:
> What is the expected behaviour when user passes both `--project` and `--directory`?

I did not check, but my expectation is that the working directory is changed but the pyproject.toml is searched in `--project`.

> I feel like there should be some more info on that matter, it doesn't feel intuitive right now and the docs are quite vague on the difference between those two.

I agree it makes sense to add a sentence about this special case. Maybe, we should even add a test case where both parameters are used.

src/poetry/console/application.py
finswimmer: Good catch :+1:

finswimmer:
> I agree it makes sense to add a sentence about this special case. Maybe, we should even add a test case where both parameters are used.

I've added a test case. Any suggestion for extending the docs?

radoering:
> Any suggestion for extending the docs?

Maybe an info box below the global options. However, while thinking about what to write I did not come up with something that is saying more than what has already been added in https://github.com/python-poetry/poetry/pull/9831#discussion_r1845246462. Maybe, that is sufficient.

radoering (approved):

PR #9849, opened 2024-11-16
vcs: use peeled ref when retrieving tag revision

abn:
This is required to support incoming changes introduced in dulwich >=0.22.2. This change preserves current behaviour and ensures locks do not change when dulwich is updated.

Note without this change upgrading dulwich will fail integration testing.

### DIFF
--- a/src/poetry/vcs/git/backend.py
+++ b/src/poetry/vcs/git/backend.py
@@ -176,7 +176,7 @@ class Git:
     @staticmethod
     def get_revision(repo: Repo) -> str:
         with repo:
-            return repo.head().decode("utf-8")
+            return repo.get_peeled(b"HEAD").decode("utf-8")
 
     @classmethod
     def info(cls, repo: Repo | Path) -> GitRepoLocalInfo:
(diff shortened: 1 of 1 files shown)


radoering:
Do we also have to replace `head()` at this place?

https://github.com/python-poetry/poetry/blob/0ebfd630945d81e1562f29ee30d78fd0021c1145/src/poetry/vcs/git/backend.py#L437

abn:
Yes we should. Good catch, think Ctrl+F skills failed me :D

radoering (approved):

jelmer:
What Dulwich change is this referring to?

src/poetry/vcs/git/backend.py
```diff
@@ -176,7 +176,7 @@ def get_remote_url(repo: Repo, remote: str = "origin") -> str:
     @staticmethod
     def get_revision(repo: Repo) -> str:
         with repo:
-            return repo.head().decode("utf-8")
+            return repo.get_peeled(b"HEAD").decode("utf-8")
```
abn, line 179: @jelmer the return value of `head()` changed for tags. Previously the peeled value was returned, upgrading to 0.22.5 started breaking our integration tests.

dimbleby: bound to be https://github.com/jelmer/dulwich/pull/1244, I bet

jelmer: That might be related to the symref regression that was fixed in 0.22.5.

abn: The issue was reproduced with 0.22.5 locally.

abn: To close the loop I have raised a dulwich issue for triage with a reproducer.

Cross-post: https://github.com/jelmer/dulwich/issues/1443

Issue #9079, opened 2024-02-29
`poetry install` authentication issue with http-basic private index (in devcontainer)


ckchow:
### Description

Using `poetry install` in a context without a keyring backend (in this case a devcontainer), trying to pull packages from a private http-basic secured pypi, causes 401 authorization in the install process. `poetry lock` works, and certain GETs earlier in `poetry install` succeed as well.

[devcontainer.json](https://github.com/python-poetry/poetry/files/14454672/devcontainer.json)
[dockerfile.txt](https://github.com/python-poetry/poetry/files/14454691/dockerfile.txt)

### Workarounds

`PYTHON_KEYRING_BACKEND=keyring.backends.null.Keyring poetry install -vvv` works

### Poetry Installation Method

pipx

### Operating System

Linux 43d0d0761105 6.5.0-15-generic #15-Ubuntu SMP PREEMPT_DYNAMIC Tue Jan  9 22:39:36 UTC 2024 x86_64 GNU/Linux

### Poetry Version

1.8.1, 1.7.1

### Poetry Configuration

```bash session
cache-dir = "/root/.cache/pypoetry"
experimental.new-installer = true
experimental.system-git-client = false
installer.max-workers = null
installer.modern-installation = true
installer.no-binary = null
installer.parallel = true
keyring.enabled = true
repositories.appml-pypi.url = "https://<redacted>/artifactory/api/pypi/appml-pypi/simple/"
solver.lazy-wheel = true
virtualenvs.create = true
virtualenvs.in-project = null
virtualenvs.options.always-copy = false
virtualenvs.options.no-pip = false
virtualenvs.options.no-setuptools = false
virtualenvs.options.system-site-packages = false
virtualenvs.path = "{cache-dir}/virtualenvs"  # /root/.cache/pypoetry/virtualenvs
virtualenvs.prefer-active-python = false
virtualenvs.prompt = "{project_name}-py{python_version}"
warnings.export = true
```
```


### Python Sysconfig

```bash session
Platform: "linux-x86_64"
Python version: "3.8"
Current installation scheme: "posix_prefix"

Paths: 
        data = "/usr/local"
        include = "/usr/local/include/python3.8"
        platinclude = "/usr/local/include/python3.8"
        platlib = "/usr/local/lib/python3.8/site-packages"
        platstdlib = "/usr/local/lib/python3.8"
        purelib = "/usr/local/lib/python3.8/site-packages"
        scripts = "/usr/local/bin"
[...]

Issue #2538, opened 2020-06-10
Private registry dependencies using API token


danieleades:
- [ ] I am on the [latest](https://github.com/python-poetry/poetry/releases/latest) Poetry version.
- [x] I have searched the [issues](https://github.com/python-poetry/poetry/issues) of this repo and believe that this is not a duplicate.
- [x] If an exception occurs when executing a command, I executed it again in debug mode (`-vvv` option).

- ubuntu 18.04
- poetry 1.0.5

## Issue

Apologies for the brevity of this description, i'm not in front of my machine right now. There may be a very simple answer to this question.

Are private pypi registries accessible using the API token, rather than username/password? I see that API token support for the public pypi registry was added in #1275, was this intended to also provide support for non-public registries?

- we have an Artifactory-hosted pypi registry
- we can use poetry with this registry using username and password authentication
- we cannot use API token authentication with this registry
- we have security concerns with using username/password authentication
- other tools (such as twine) *are* able to access this registry using the API token, rather than username/password

I guess i'm asking if some part of the API token authentication implementation is hardcoded to the public pypi registry?

If it *is* intended that this should work, what diagnostic information can I provide, or debugging steps can I follow?

(note that i'm lagging a couple of releases on the poetry version. Please let me know if there's any changes that could affect this that are undocumented in the changelog)

Persedes:
we have a similar setup and poetry (>=1) works fine with a PAT. 
You can use the following command:

`poetry config http-basic.foo $PAT "" (where pat goes in place of the user name)`

https://python-poetry.org/docs/repositories/

danieleades:
@Persedes if that works that is a massive help!

I'm going to tentatively say that this is still a bug. there's a documented solution using the api token that doesn't work-

`poetry config pypi-token.$REPO $TOKEN`

and an undocumented solution which allegedly *does* work

`poetry config http-basic.$REPO $TOKEN ""`

jdeyton:
@danieleades I'm of the opinion the documentation is a little unclear. However it seems to work for multiple users:

There are steps provided here in #910, specifically https://github.com/python-poetry/poetry/issues/910#issuecomment-484963046

For what it's worth, I was able to get a poetry to publish to a pywharf container backed by the local filesystem using this approach, with the exception that I provide the repo name as the username, e.g., with the repo "mypypi" and secret "foo":
```
user@host > poetry config repositories.mypypi http://localhost:8888/simple/                                                                                                                                                          
[...]

PR #9850, opened 2024-11-16
Fix regression when using empty username/password

abn:
This change fixes a regression in the password manager that disallowed the use of empty username or password as required by some self-hosted repositories.

Relates-to: #9079 #2538

### DIFF
--- a/docs/repositories.md
+++ b/docs/repositories.md
@@ -511,6 +511,17 @@ You can prevent this by adding double dashes to prevent any following argument f
 poetry config -- http-basic.pypi myUsername -myPasswordStartingWithDash
 ```
 
+{{% note %}}
+In some cases like that of [Gemfury](https://gemfury.com/help/errors/repo-url-password/) repositories, it might be
+required to set an empty password. This is supported by Poetry.
+
+```bash
+poetry config http-basic.foo <TOKEN> ""
+```
+
+**Note:** Usernames cannot be empty.  Attempting to use an empty username can result in an unpredictable failure.
+{{% /note %}}
+
 ## Certificates
 
 ### Custom certificate authority and mutual TLS authentication
--- a/src/poetry/utils/password_manager.py
+++ b/src/poetry/utils/password_manager.py
@@ -195,18 +195,25 @@ class PasswordManager:
     def get_http_auth(self, repo_name: str) -> dict[str, str | None] | None:
         username = self._config.get(f"http-basic.{repo_name}.username")
         password = self._config.get(f"http-basic.{repo_name}.password")
-        if not username and not password:
+
+        # we only return None if both values are None or ""
+        # password can be None at this stage with the username ""
+        if (username is password is None) or (username == password == ""):
             return None
 
         if not password:
             if self.use_keyring:
                 password = self.keyring.get_password(repo_name, username)
-            else:
+            elif not username:
+                # at this tage if username is "" or None, auth is invalid
                 return None
 
+        if not username and not password:
+            return None
+
         return {
-            "username": username,
-            "password": password,
+            "username": username or "",
+            "password": password or "",
         }
 
     def set_http_password(self, repo_name: str, username: str, password: str) -> None:
--- a/src/poetry/utils/password_manager.py
+++ b/src/poetry/utils/password_manager.py
@@ -192,29 +192,17 @@ class PasswordManager:
 
         self.keyring.delete_password(repo_name, "__token__")
 
-    def get_http_auth(self, repo_name: str) -> dict[str, str | None] | None:
+    def get_http_auth(self, repo_name: str) -> HTTPAuthCredential:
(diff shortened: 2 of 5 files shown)


abn:
We might need to disallow empty usernames. 

See: jaraco/keyring#668 jaraco/keyring#687

abn:
Keeping the commits separate.

src/poetry/utils/password_manager.py
```diff
@@ -195,18 +195,19 @@ def delete_pypi_token(self, repo_name: str) -> None:
...
+
+        if not username:
             return None
 
         if not password:
             if self.use_keyring:
                 password = self.keyring.get_password(repo_name, username)
-            else:
+            elif not username:
                 return None
```
dimbleby, line 206: Unhittable, we would have bailed already

docs/repositories.md
```diff
@@ -511,6 +511,17 @@ You can prevent this by adding double dashes to prevent any following argument f
...
 
+{{% note %}}
+In some cases like that of [Gemfury](https://gemfury.com/help/errors/repo-url-password/) repositories, it might be
+required to set an empty password. This is supported by Poetry.
+
+```bash
+poetry config http-basic.foo <TOKEN> ""
+```
+
+However, it is important to note that empty usernames are not supported. And attempts use it will fail unpredictably.
```
radoering, line 522:
```suggestion
However, it is important to note that empty usernames are not supported - and attempts to use it will fail unpredictably.
```

Maybe?

src/poetry/utils/password_manager.py
abn: Not sure that is accurate. When `username="bar"`,  `password=""` and `self.use_keyring == False`, we want to skip to return and not return `None`.

dimbleby: when `username="bar"` we will not go into the code branch `elif not username:`

that branch is unhittable because you already made that check and bailed five lines earlier

docs/repositories.md
abn: Good catch. I originally meant to write "Any attempts". Rewrote it to be more direct.

src/poetry/utils/password_manager.py
dimbleby, line 206: previous comment was "resolved" too soon and I do not have power to unresolve it, so leaving another!

this code is indeed unhittable.

or if I am wrong: please write a unit test covering it

abn: @dimbleby I thought I responded with the case. The test case already exists https://github.com/python-poetry/poetry/pull/9850/files#diff-9290d504b4b225fb1efa7f54d44cc4ed335659c90d0c6ca558176d326b3d1355R138.

You can switch that to an `else` statement and run the test case, it will fail.

dimbleby: the `return None`  is not hit by your testcase, as you easily can verify.  Just remove both lines of code.

abn: Fair enough. I can drop both lines :smile:

abn: Dropped and simplified the if block now that I follow you.

src/poetry/utils/password_manager.py
```diff
@@ -195,18 +195,16 @@ def delete_pypi_token(self, repo_name: str) -> None:
...
-                password = self.keyring.get_password(repo_name, username)
-            else:
-                return None
+        if not password and self.use_keyring:
+            password = self.keyring.get_password(repo_name, username)
 
         return {
             "username": username,
-            "password": password,
+            "password": password or "",
```
dimbleby, line 207: the type annotation on the return value for this function can be tightened up now I think - the values in this dictionary are guaranteed strings

dimbleby: actually perhaps it should have been an `HTTPAuthCredential` all along, though perhaps we are beginning to get into feature creep for this pull request

abn: While it is a bit of a feature creep, I reckon it might be a useful one. I'll try going down the rabbit hole to see how far I can go without it becoming a much bigger change.

abn: Turns out sending empty string when password is None is probably a bad thing. But fixed the method to return `HTTPAuthCredential` now. Further tightening might require a slightly more involved PR.

dimbleby: Yeah.  I would think the `HTTPAuthCredential` should guarantee a non-empty - or at least, not-`None` - username.  But I can definitely see that being a change that begins to touch more than you might want to right now!

abn: Yeah will look into the side-effects of that a bit later. For now this should suffice.

dimbleby (approved):

Secrus (approved):

PR #9851, opened 2024-11-17
ignore installed packages during solving

### LINKED ISSUE #8328: use of installed packages means poetry lock does not fully respect designated source for dependencies
- **Poetry version**: 1.5.1
- **Python version**: 3.10.12
- **OS version and name**: MacOS 12.6.8
- **pyproject.toml**: 

- [x] If an exception occurs when executing a command, I executed it again in debug mode (`-vvv` option) and have included the output below. **N/A**

## Issue

When I started writing this, I thought the issue was different - now I think I see what's happening.

In a clean project for which `poetry.lock` has never been run before, `poetry` works as expected.

However, the results of `poetry lock` seem to be influenced by poetry's cache (somehow), since in an existing project, where poetry has successfully done `lock` before, it does not seem to be possible to tell `poetry` to limit where it looks for a given dependency.

Let's say we have the following `pyproject.toml` in a clean, new project, and that `my_dep` is only available in private repo `privaterepo`:

```
[tool.poetry]
name = "my-project"
version = "0.1.0"
description = ""
authors = ["Jeff Lerman <jeff.lerman@somewhere.com>"]
readme = "README.md"

[tool.poetry.dependencies]
python = "^3.10.0"
my-dep = {version = "^0.8.2", source = "PyPI"}

[[tool.poetry.source]]
name = "private_repo"
url = "https://repo.shared.privaterepo.com/repository/pypi-privaterepo/simple"
priority = "explicit"

[[tool.poetry.source]]
name = "PyPI"
priority = "supplemental"

[build-system]
requires = ["poetry-core>=1.4.0"]
```

`poetry lock` correctly tells us: 
```
Because my-project depends on my-dep (^0.8.2) which doesn't match any versions, version solving failed.
```

But now if we:
1. change the `pyproject.toml` to indicate the correct source (`private_repo`)
2. run `poetry lock` (completes successfully)
3. run `poetry install` (this is key!!)
4. change `pyproject.toml` back to the wrong source (PyPI)
5. run `poetry lock`

Then there is no failure anymore.  We get the following clause in the resulting `poetry.lock`:
```
[[package]]
name = "my-dep"
version = "0.8.3"
description = "this is a very private package"
optional = false
python-versions = "*"
files = []
```

So, `poetry` is seeing the dependency in its local cache (somehow), using that to decide that it's a valid package, but not getting any metadata for it except the version & description - and populating an awfully sparse-looking `poetry.lock` clause.

That means there's no way to tell poetry, once a package has been found and installed (presumably, from any repo) that now we want to install it from a specific private repo - also no way to ensure that we've completely locked down the source for packages that we want to only ever come from our private repo.

I noticed this when trying to upgrade existing projects to the new `poetry` 1.5.x style of specifying sources, and not getting failures when I should have been.

dimbleby:
fixes #8328, #9774

Currently-installed packages have no place in the locking process.  Folk who are doing development against unreleased projects can and should use a path dependency, or a git dependency, or a dependency on a private index, or a privately built wheel

### DIFF
--- a/src/poetry/puzzle/provider.py
+++ b/src/poetry/puzzle/provider.py
@@ -27,7 +27,6 @@ from poetry.packages import DependencyPackage
 from poetry.packages.direct_origin import DirectOrigin
 from poetry.packages.package_collection import PackageCollection
 from poetry.puzzle.exceptions import OverrideNeededError
-from poetry.repositories.exceptions import PackageNotFoundError
 from poetry.utils.helpers import get_file_hash
 
 
@@ -115,7 +114,6 @@ class Provider:
         pool: RepositoryPool,
         io: IO,
         *,
-        installed: list[Package] | None = None,
         locked: list[Package] | None = None,
     ) -> None:
         self._package = package
@@ -129,7 +127,6 @@ class Provider:
         self._deferred_cache: dict[Dependency, Package] = {}
         self._load_deferred = True
         self._source_root: Path | None = None
-        self._installed_packages = installed if installed is not None else []
         self._direct_origin_packages: dict[str, Package] = {}
         self._locked: dict[NormalizedName, list[DependencyPackage]] = defaultdict(list)
         self._use_latest: Collection[NormalizedName] = []
@@ -205,36 +202,6 @@ class Provider:
                 f" package's name: {package.name}"
             )
 
-    def search_for_installed_packages(
-        self,
-        dependency: Dependency,
-    ) -> list[Package]:
-        """
-        Search for installed packages, when available, that satisfy the given
-        dependency.
-
-        This is useful when dealing with packages that are under development, not
-        published on package sources and/or only available via system installations.
-        """
-        if not self._installed_packages:
-            return []
-
-        logger.debug(
-            "Falling back to installed packages to discover metadata for <c2>%s</>",
-            dependency.complete_name,
-        )
-        packages = [
-            package
-            for package in self._installed_packages
-            if package.satisfies(dependency, ignore_source_type=True)
-        ]
-        logger.debug(
-            "Found <c2>%d</> compatible packages for <c2>%s</>",
-            len(packages),
-            dependency.complete_name,
-        )
-        return packages
-
     def search_for_direct_origin_dependency(self, dependency: Dependency) -> Package:
         package = self._deferred_cache.get(dependency)
         if package is not None:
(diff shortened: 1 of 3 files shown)


abn:
@dimbleby doesn't this break the case where system site packages are used for packages that are only available via os package management? While this is not an explicitly supported case, we have been relaxed on such scenarios - this change will start breaking those cases. Might not be a bad thing, just raising this as I have seen this workaround being used in the wild by some clients of mine - particularly for packages with native extensions.

In a away would be good to see if anyone complains and if so, we might need to figure out a better way to support a "system"or "external" source more explicitly when the behavior is desired.

dimbleby:
>  doesn't this break the case where system site packages are used for packages that are only available via os package management? 

in such cases: just do not declare the dependency

indeed - if such cases even exist - declaring a dependency on such a package would open the door to horrors where I publish a package with a name that deliberatey clashes with your system package.  Then poetry will upgrade that package and break your system.

abn:
Can you add a small FAQ entry for this (why installed packages are not used) as well so we can point to it when inevitably it comes up in an issue?

dimbleby:
I am not sure that I believe there even are any such cases: in that context I expect zero questions about it rather than frequent questions!

Do you have any examples?  Without a concrete example I fear that a FAQ would have to be written rather abstractly...

I think I would suggest deferring FAQ entries until such time as the question is frequently asked.

dimbleby:
Ah, perhaps we are at cross-purposes.  I doubt that the os-package-manager-only case exists, but maybe you think a FAQ is worthwhile anyway for the more likely case of local development?

Even then I would propose starting with a changelog / release note entry and promote that only if and when there are frequent questions.  But I can leave some helpful text here for when that changelog comes, perhaps something like:

> Poetry no longer pays attention to packages installed during your environment during locking.  If you have declared a dependency on package that is unpublished, but available only in your local environment, you should do one of:
> - remove that dependency from `pyproject.toml`
> - declare it instead in a way that does not rely on a public index, for instance as a path dependency

abn:
Sorry, should have been a bit clearer there.

I was thinking about the "developing against unreleased packages" case, not the system package case. The latter is definitely rarer, I agree. 

I am okay with deferring it if you reckon that is not a common enough scenario.

abn (approved):
Approving this as I concur with the rationale that installed packages should not be a source of truth when solving. Doing so is decidedly an anti-pattern.

If, in the future, there is a legitimate reason to use a source for package metadata that is not already a valid source, this must be implemented as a new source type with explicit details around how it gets handled.

PR #9855, opened 2024-11-18
install: make `--sync` the default and introduce `--keep-untracked`

radoering:
# Pull Request Check List

Closes #9801

Alternative approach after the discussion in #9801.


radoering:
I am closing this since @finswimmer expressed his doubts as to whether it is worth it and I am not convinced myself because I am expecting complaints about unexpected uninstalls. In my opinion, #9801 makes more sense.

PR #9861, opened 2024-11-20
env: fix check whether path is relative to system site packages

radoering:
We are looking into the wrong environment.

We must not use `SystemEnv`, which is Poetry's own env, but `env.parent_env`, which is the base env.

# Pull Request Check List

### DIFF
--- a/src/poetry/utils/env/virtual_env.py
+++ b/src/poetry/utils/env/virtual_env.py
@@ -3,7 +3,6 @@ from __future__ import annotations
 import json
 import os
 import re
-import sys
 
 from contextlib import contextmanager
 from copy import deepcopy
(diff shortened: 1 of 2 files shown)


Secrus (approved):
The code seems fine to me, but I am not sure I get what is wrong here. Could you add some more context on what the bug is?

abn (approved):

radoering:
> Could you add some more context on what the bug is?

The method is used to determine if a package is in the env's site packages or - if the env is a virtual env created with `include-system-site-packages = true` - in the env's or system site packages. The bug is that `SystemEnv(Path(sys.prefix)` is Poetry's own env and not the env where the system site packages are (in the context of the virtual env).

PR #9863, opened 2024-11-20
install: make `--sync` compatible with `virtualenvs.options.system-site-packages = true`

radoering:
# Pull Request Check List

More or less required for #9855

Without this fix, we try to uninstall untracked system site packages when running `poetry install --sync`.

### DIFF
--- a/src/poetry/installation/installer.py
+++ b/src/poetry/installation/installer.py
@@ -39,7 +39,7 @@ class Installer:
         locker: Locker,
         pool: RepositoryPool,
         config: Config,
-        installed: Repository | None = None,
+        installed: InstalledRepository | None = None,
         executor: Executor | None = None,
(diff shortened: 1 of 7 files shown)


radoering:
> This seems to have a side-effect when poetry installs a managed package to system-site. It does also removes previously installed packages.

This PR does not change anything about that. You can reproduce the same behavior with Poetry 1.8.4.

If you replace `poetry config virtualenvs.create false` with `poetry config virtualenvs.options.system-site-packages true` you will see what this PR changes.

abn:
You are right. Looks like the removal of system packages when create env is false is an existing issue. I have verified that that this change does functionally fix `--sync` from being too eager in removals when a virtualenv is used with system site packages.

abn (approved):

PR #9864, opened 2024-11-21
doc: add documentation for using extension modules

abn:


smith120bh:
@abn : I spotted this PR from your post on the issue, which I've been following. This looks great! Just one thing I noticed - this statement looks mutually contradictory to me? I've personally not tried putting a build script in a subdirectory, but I had thought that it needed to be in the root directory, next to pyproject.toml.
image

docs/building-extension-modules.md
```diff
@@ -0,0 +1,338 @@
...
+{{% /warning %}}
+
+Poetry allows a project developer to introduce support for, build and distribute native extensions within their project.
+In order to achieve this, at the highest level, the following steps are required.
+
+{{< steps >}}
+{{< step >}}
+**Add Build Dependencies**
+
+The build dependencies, in this context, refer to those Python packages that required in order to successfully execute
```
radoering, line 29:
```suggestion
The build dependencies, in this context, refer to those Python packages that are required in order to successfully execute
```

docs/building-extension-modules.md
```diff
@@ -0,0 +1,338 @@
...
+your build script. Common examples include `cython`, `meson`, `maturin`, `setuptools` etc., depending on how your
+extension is built.
+
+{{% note %}}
+You must assume that only python built-ins are available by default in a build environment. This means, if you use need
```
radoering, line 34:
```suggestion
You must assume that only Python built-ins are available by default in a build environment. This means, if you need
```

docs/building-extension-modules.md
```diff
@@ -0,0 +1,338 @@
...
+even packages like `setuptools`, it must be explicitly declared.
+{{% /note %}}
+
+The necessary build dependencies must be added to the `build-system.requires` section of your `pyproject.toml` file.
+
+```toml
+[build-system]
+requires = ["poetry-core", "setuptools", "cython"]
```
radoering, line 42: Do we want to use unrestricted build requirements here or do we want to set upper/lower bounds (cf #9812)?

docs/building-extension-modules.md
```diff
@@ -0,0 +1,338 @@
...
+poetry add --group=build setuptools cython
+```
+{{% /note %}}
+
+{{< /step >}}
+
+{{< step >}}
+**Add Build Script**
+
+The build script can be free-form Python script that uses any dependency specified in the previous step. This can be
```
radoering, line 60:
```suggestion
The build script can be a free-form Python script that uses any dependency specified in the previous step. This can be
```

docs/building-extension-modules.md
```diff
@@ -0,0 +1,338 @@
...
+{{% /warning %}}
+
+```toml
+[tool.poetry]
+...
+packages = [
+    { include = "package", from = "src" }
+]
+include = [
+    { path = "src/package/**/*.so", format = "wheel" },
```
radoering, line 92: Does this work? Or is the `*.so` put into a `src` folder into the wheel? I remember such issues because we do not support `from` and `to` here as we do for `packages`.

docs/building-extension-modules.md
```diff
@@ -0,0 +1,338 @@
...
+{{< tabs tabTotal="2" tabID1="cython-pyproject" tabName1="pyproject.toml" tabID2="cython-build-script" tabName2="build-extension.py">}}
+
+{{< tab tabID="cython-pyproject" >}}
+
+```toml
+[tool.poetry.build]
+script = "build-extension.py"
+
+[build-system]
+requires = ["poetry-core", "cython"]
```
radoering, line 125:
```suggestion
requires = ["poetry-core", "cython", "setuptools"]
```

radoering (commented):
Some first feedback. (I have not looked at the examples in details yet.)

You did not mention `generate-setup-file`. (Personally, I have no idea if/when you need this setting.) Should it be mentioned?

Secrus:
> You did not mention `generate-setup-file`. (Personally, I have no idea if/when you need this setting.) Should it be mentioned?

That's a good thing. I'd personally be glad to see that option gone, since it ties us to `setuptools`. It was useful when PEP 517 was still new and not all tools supported build backends different than setuptools, but nowadays it has limited use.

docs/building-extension-modules.md
abn: Well that is up to the user. And honestly keeping this doc up to date every time we release core is a burden without much gain. But I am not firm on best approach here. Not sure which is more likely: someone copy pasting outdated version of this or someone being bitten by a major version upgrade breakage.

abn:
Yes. The non-mention of that option is intentional. Today, it defaults to false, and we may remove it in the future. Especially as the setup file generation is full of holes.

docs/building-extension-modules.md
abn: You are probably correct. I'll see if I can fix that if it's the case.

docs/building-extension-modules.md
abn: I have added a note about this in the doc.

docs/building-extension-modules.md
abn: Yeah, the current include/exclude stuff is a bit of a mess. I have simplified the doc to work with what we have right now.

abn: The following seems to do the right things for now.

```
{ path = "src/package/**/*.so", format = "wheel" },
```

abn:
```console
$ unzip *.whl
Archive:  poetry_cython_example-0.1.0-cp313-cp313-manylinux_2_40_x86_64.whl
  inflating: package/__init__.py     
  inflating: package/example.cpython-313-x86_64-linux-gnu.so  
  inflating: package/example.pyx     
  inflating: poetry_cython_example-0.1.0.dist-info/METADATA  
  inflating: poetry_cython_example-0.1.0.dist-info/WHEEL  
  inflating: poetry_cython_example-0.1.0.dist-info/RECORD 
```

radoering: I debugged this and it only works because the `*.so` file has already been picked up by the package include (with the correct target directory) so that the other include (with the wrong target directory) is a NOP (because we do not include files twice). Otherwise, it would end in the wrong directory.

In other words, you can remove the `include` and get the same result.

abn: Well it does ignore the c files being generated though, so this means the package directory is not necessarily all included. Or are you saying `*.so` files are treated differently?

radoering: Ah, I did not consider that the `*.so` is in your `.gitignore`. That makes the difference. This is really weird: The `*.so` file would be ignored without the `include` because of `.gitignore`. However, the `include` only ensures that the `*.so` is not ignored but it is not used itself, but the `*.so` is picked up from the package include. Thus, the example makes sense, especially with `.gitignore`.

abn: I guess we should look at the include/exclude stuff at some point.

radoering (approved):
I have only skimmed over the examples and have not tried them out, but since there is a warning at the beginning of the section, I think it is ok to publish it and wait for feedback.

Btw, do you want to squash the commits? They have the same message and I do not see that keeping them separate adds any value.

abn:
Squash is fine. I think I forgot to squash the last push.

PR #9879, opened 2024-11-29
chore: update hugo (must be in sync with website repo) and simplify/fix trigger condition

radoering:
The update is required because of python-poetry/website#159

The `paths-ignore` is unnecessarily complicated and can be replace with `paths`.

The `if` condition was just wrong.

### DIFF
--- a/.github/workflows/docs.yaml
+++ b/.github/workflows/docs.yaml
@@ -3,18 +3,16 @@ name: Documentation Preview
 on:
   pull_request:
     # allow repository maintainers to modify and test workflow
-    paths-ignore:
-      - "**"
-      - "!.github/workflows/docs.yaml"
+    paths:
+      - ".github/workflows/docs.yaml"
   pull_request_target:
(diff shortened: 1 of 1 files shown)


.github/workflows/docs.yaml
```diff
@@ -67,7 +65,7 @@ jobs:
           # Build website assets (CSS/JS).
           npm ci && npm run prod
           # Build the static website.
-          hugo -v --minify
+          hugo --minify --logLevel info
```
branchvincent, line 68: might be easier to just use the npm-managed binary? I think the `peaceiris/actions-hugo` action predated our use of https://www.npmjs.com/package/hugo-bin:
```suggestion
          npx hugo --minify --logLevel info
```

radoering: Good catch. Tested it in python-poetry/website#162 and adopted it.

branchvincent (approved):

Issue #9670, opened 2024-09-04
Implement PEP 639


befeleme:
### Issue Kind

Change in current behaviour

### Description

See: https://peps.python.org/pep-0639

Additional context:

- The full implementation, including the Core Metadata bump to 2.4, will require also the support on the side of PyPI first: 
https://github.com/pypi/warehouse/issues/16620

- There's a plan to add validation data and parser to packaging:
https://github.com/pypa/packaging/pull/799 + https://github.com/pypa/packaging/pull/828

### Impact

poetry now uses the `license` field from the custom `[tool.poetry]` table - this will enable unification under the specified main `[project]` table of pyproject.toml and map to a new core metadata field.
This also brings in a standardized way of specifying the license files in the package metadata, which will help the downstream consumers locate them unambiguously.

### Workarounds

poetry now includes both the license expression (as SPDX) and the license files in the distribution. This will just make the declarations and metadata unified across many packaging tools.

Secrus:
We will wait for packaging to support it and use their implementation.

carlwgeorge:
This looks to be implemented in packaging 24.2.

https://github.com/pypa/packaging/releases/tag/24.2

dekim24:
Hi - we started getting the following error when trying to update `ruff` to the latest version `0.8.1` which was released very recently. Is there some work that needs to be done for poetry to support packages using the new metadata 2.4 version?

```Unable to determine package info for path: /var/folders/j3/kcklg1d50637nkxf8v4py65h0000gp/T/tmpcybtv8zi/ruff-0.8.1-py3-none-linux_armv6l.whl

Unknown metadata version: 2.4
```

martynvdijke:
We have the same problem

andresdelfino:
urllib3 2.3.0 released today uses metadata version 2.4.

edgarrmondragon:
> urllib3 2.3.0 released today uses metadata version 2.4.

@andresdelfino what problem is that causing for you? and have you seen https://github.com/astral-sh/ruff/issues/14681#issuecomment-2512302023? It seems upgrading `pkginfo` fixes the metadata issues.

andresdelfino:
> > urllib3 2.3.0 released today uses metadata version 2.4.
> 
[...]

PR #9881, opened 2024-12-02
fix: workaround unsupported metadata version 2.4 in pkginfo

finswimmer:
`pkginfo` will show a warning if it detects metadata version:

```
NewMetadataVersion: New metadata version (2.4) higher than latest supported version: parsing as 2.3
  warnings.warn(NewMetadataVersion(self.metadata_version))
```

I decided to not suppress this message.

Related to https://github.com/python-poetry/poetry/issues/9670#issuecomment-2510587030


radoering (approved):

MichaReiser (commented):
This fix might no longer be required. A new `pkginfo` version was released just a few minutes ago that adds support for 2.4.

What would be nice in general is if poetry wouldn't fail when seeing a newer minor version and instead just warns about it.

dimbleby:
>  poetry wouldn't fail when seeing a newer minor version and instead just warns about it.

that already is the behaviour, since pkginfo 1.11.0

radoering:
> This fix might no longer be required. A new pkginfo version was released just a few minutes ago that adds support for 2.4.

Thanks for the info.

> What would be nice in general is if poetry wouldn't fail when seeing a newer minor version and instead just warns about it.

Poetry did not fail before #9203. However, this resulted in #9195, which is worse than failing. It seems like `pkginfo` handles unknown METADATA versions better now so maybe we can (partially) revert #9203.

dimbleby:
poetry only fails on unexpected metadata versions in some of the code paths, which I think adds to the general confusion (including my own)

eg so far as I can see the `_get_info_from_metadata()` path does not care about metadata version.

Probably this is why some people saw a problem and some did not - they either were or were not using pypi / a private repository.

broodjeaap:
For anyone looking for a quick fix, you can downgrade poetry with:
`poetry self update 1.8.2`

dimbleby:
it's upgrading pkginfo that helps, not downgrading poetry

Buedenbender:
In my case it was resolved by 
```bash
pip install --upgrade pkginfo
pip install --upgrade poetry
pip install --upgrade pip
```

Aftwards i was able to successfully install ruff in my poetry env via
`poetry add -G dev ruff`

PR #9887, opened 2024-12-04
Make poetry export tests optional

Secrus:
Let's make `poetry-plugin-export` tests optional in CI. Most changes we make won't affect the plugin, and for situations where there is a risk, we can check by applying a label and running extra pipelines.

### DIFF
--- a/.github/workflows/.tests-matrix.yaml
+++ b/.github/workflows/.tests-matrix.yaml
@@ -75,7 +75,7 @@ jobs:
   pytest-export:
     name: pytest (poetry-plugin-export)
     runs-on: ${{ inputs.runner }}
-    if: inputs.run-pytest-export
+    if: ${{ inputs.run-pytest-export && contains(github.event.pull_request.labels.*.name, 'test poetry export')}}
     steps:
       - uses: actions/checkout@0ad4b8fadaa221de15dcec353f45205ec38ea70b # v4.1.4
         with:


radoering:
I do not have a strong opinion, just some thoughts:

1. It is not easy to know if a change will affect the export plugin. When I remember my PRs that let the plugin tests fail, I probably did not expect an influence on the export plugin for the majority of them; in other words, I would not have applied the label. Thus, making the tests optional, will result in failures caused by previous, already merged changes, where we did not apply the label.
2. I have already wondered whether we want to keep the export tests after 2.0, with the export plugin not being installed per default anymore. Why executing export tests but not bundle tests (or shell tests)? Probably, because the export plugin is more widely used and thereby more important.

Making the tests optional might be a good compromise.

.github/workflows/.tests-matrix.yaml
```diff
@@ -75,7 +75,7 @@ jobs:
   pytest-export:
     name: pytest (poetry-plugin-export)
     runs-on: ${{ inputs.runner }}
-    if: inputs.run-pytest-export
+    if: ${{ inputs.run-pytest-export && contains(github.event.pull_request.labels.*.name, 'test poetry export')}}
```
radoering, line 78: The name of the label is not consistent with our naming scheme (no spaces, but slashes). Maybe, just `test/export`?

Secrus:
@radoering While I agree with most of your concerns, I'd say that it would be better to test compatibility on the plugin side and fix stuff if needed rather than slowing down the CI here. We could make export (bundle and shell, too) tests run as part of a release pipeline only so that we can ensure that the plugins work with the new release and have plugins run compatibility pipelines in their repos.

PR #9888, opened 2024-12-06
info: require `pkginfo >= 1.12` for METADATA 2.4 support and loosen check for unknown METADATA versions

radoering:
# Pull Request Check List

Resolves: #9885

I think we should do a 1.8 release, which requires `pkginfo >= 1.12` to reduce support workload.

Further, after some testing I agree to @Secrus' proposal to loosen the check for unknown METADATA versions.

We introduced this check because `pkginfo < 1.11.0` did not handle unknown versions well:

```
>>> print(pkginfo.Wheel("demo_metadata_version_299-0.1.0-py2.py3-none-any.whl").requires_dist)
()
```

However, `pkginfo >= 1.11.0` handles it better:

```
>>> print(pkginfo.Wheel("demo_metadata_version_299-0.1.0-py2.py3-none-any.whl").requires_dist)
['cleo; extra == "foo"', 'pendulum (>=1.4.4)', 'tomlkit; extra == "bar"']
NewMetadataVersion: New metadata version (2.99) higher than latest supported version: parsing as 2.3
```


Secrus (approved):
LGTM, thank you for tackling this issue.

PR #9899, opened 2024-12-12
fix: properly close streaming requests when not completely consumed

Tasssadar:
https://requests.readthedocs.io/en/latest/user/advanced/#body-content-workflow

> If you set stream to True when making a request, Requests cannot release the connection back to the pool unless you consume all the data or call [Response.close](https://requests.readthedocs.io/en/latest/api/#requests.Response.close). This can lead to inefficiency with connections.

In our case, with combination with Artifactory, these open connections are stuck in the middle of a download, because Artifactory does not support the Range requests for lazy wheel.

Not sure if I need to write test for this, let me know if so.

### DIFF
--- a/src/poetry/inspection/lazy_wheel.py
+++ b/src/poetry/inspection/lazy_wheel.py
@@ -428,15 +428,20 @@ class LazyFileOverHTTP(ReadOnlyIOWrapper):
         headers["Range"] = f"bytes={start}-{end}"
         logger.debug("streamed bytes request: %s", headers["Range"])
         self._request_count += 1
+
         response = self._session.get(self._url, headers=headers, stream=True)
-        response.raise_for_status()
-        if int(response.headers["Content-Length"]) != (end - start + 1):
-            raise HTTPRangeRequestNotRespectedError(
-                f"server did not respect byte range request: "
-                f"requested {end - start + 1} bytes, got "
-                f"{response.headers['Content-Length']} bytes"
-            )
-        return response
+        try:
+            response.raise_for_status()
+            if int(response.headers["Content-Length"]) != (end - start + 1):
+                raise HTTPRangeRequestNotRespectedError(
+                    f"server did not respect byte range request: "
+                    f"requested {end - start + 1} bytes, got "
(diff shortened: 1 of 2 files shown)


Tasssadar:
I just found it this does not really matter if we were using just `requests`, because finalizers would take care of closing the responses - but there is also a bug in `cachecontrol`, which creates circular references and prevents the finalizers from working: https://github.com/psf/cachecontrol/pull/344

src/poetry/inspection/lazy_wheel.py
```diff
@@ -428,15 +428,20 @@ def _stream_response(self, start: int, end: int) -> Response:
...
+        try:
+            response.raise_for_status()
+            if int(response.headers["Content-Length"]) != (end - start + 1):
+                raise HTTPRangeRequestNotRespectedError(
+                    f"server did not respect byte range request: "
+                    f"requested {end - start + 1} bytes, got "
+                    f"{response.headers['Content-Length']} bytes"
+                )
+            return response
+        except BaseException:
```
radoering, line 442: In `lazy_wheel`, you are using `BaseException`. In `helpers`, you are using `Exception`. Is there a reason for that?

radoering (commented):
Good catch. 👍

> Not sure if I need to write test for this, let me know if so.

Do you think a test makes sense? If the only way to test it, is to `assert` that `close()` is called, it may not be worth it.

src/poetry/inspection/lazy_wheel.py
Tasssadar: Hm, yeah, will change it to BaseException when I get back to PC on monday.

Tasssadar:
> Good catch. 👍
> 
> > Not sure if I need to write test for this, let me know if so.
> 
> Do you think a test makes sense? If the only way to test it, is to `assert` that `close()` is called, it may not be worth it.

Fixed the exception type. In my opinion, the test do not make much sense, as they would be pretty fragile and would not check any new places where `requests` usage could be added in the future.

racinmat:
Hi, when can we expect this being released to pypi?

PR #9902, opened 2024-12-14
Always create virtual env if virtualenvs.force-create is True

### LINKED ISSUE #4055: Option to force Poetry to create a virtual environment, even if a virtual env is active

jasondamour:
# Pull Request Check List

Resolves: #4055

### DIFF
--- a/src/poetry/console/commands/config.py
+++ b/src/poetry/console/commands/config.py
@@ -73,6 +73,7 @@ To remove a repository (repo is a short alias for repositories):
         unique_config_values = {
             "cache-dir": (str, lambda val: str(Path(val))),
             "virtualenvs.create": (boolean_validator, boolean_normalizer),
+            "virtualenvs.force-create": (boolean_validator, boolean_normalizer),
             "virtualenvs.in-project": (boolean_validator, boolean_normalizer),
             "virtualenvs.options.always-copy": (boolean_validator, boolean_normalizer),
             "virtualenvs.options.system-site-packages": (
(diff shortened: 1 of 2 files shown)


radoering:
As it is implemented, when `virtualenvs.force-create` is set:

- Poetry still uses an existing in-project `.venv` (and does not create a new venv) if `virtualenvs.in-project` is not set to `false`. That is probably fine.
- If `virtualenvs.create` is set to `false`, the system env will be used. No venv will be created. Not sure if that makes sense.

@finswimmer Pinging you since you have done some refactoring of the env management recently. What do you think of such an option, especially in combination with the existing options, cf https://github.com/python-poetry/poetry/issues/4055#issuecomment-2209552367.

As an alternative to a new option we could also allow the value `"force"` for `virtualenvs.create`. Then, we do not have to think about combinations like `virtualenvs.create = false` and `virtualenvs.force-create = true`.

After having decided what exactly this option should look like, we definitively need some tests and docs.

jasondamour:
@radoering thanks for the quick (weekend) review! I agree with you on everything.

> As an alternative to a new option we could also allow the value "force" for virtualenvs.create.

I thought about this too, and decided against it to make it a "safer" change to merge. But if the maintainers agree with it, I'll implement that instead.

Additionally, I would like to see this released in 1.X, is there anything I need to do regarding the branch to do that?

PR #9933, opened 2025-01-04
chore: update dependencies and fix warnings

radoering:


tests/masonry/builders/test_editable_builder.py
```diff
@@ -11,7 +11,7 @@
 import pytest
 
 from cleo.io.null_io import NullIO
-from deepdiff import DeepDiff
+from deepdiff import DeepDiff  # type: ignore[attr-defined]
```
dimbleby, line 14:
```python
from deepdiff.diff import DeepDiff
```

pyproject.toml
```diff
@@ -72,6 +72,7 @@ pytest-cov = ">=4.0"
 pytest-mock = ">=3.9"
 pytest-randomly = ">=3.12"
 pytest-xdist = { version = ">=3.1", extras = ["psutil"] }
+urllib3 = "<2.3"  # see https://github.com/gabrielfalcao/HTTPretty/issues/484
```
dimbleby, line 75: this crisis was bound to come: I would not bet on there being a new httpretty any time soon...

radoering: Fortunately, it is only a dev dependency. That is why I would say it is not that relevant for the release. In the medium term, of course, we have to replace it.

finswimmer (approved):

pyproject.toml
Romain-Geissler-1A: Indeed httpretty seems stalled. Some projects seems to move their dependency from httpretty to mocket which seems to have mocks specifically aiming at httpretty compatibility.

PR #9936, opened 2025-01-04
Remove outdated `poetry lock --no-update` suggestion

edgarrmondragon:
# Pull Request Check List

Resolves: #issue-number-here

### DIFF
--- a/src/poetry/console/commands/check.py
+++ b/src/poetry/console/commands/check.py
@@ -158,7 +158,7 @@ class CheckCommand(Command):
         if self.poetry.locker.is_locked() and not self.poetry.locker.is_fresh():
             check_result["errors"] += [
                 "pyproject.toml changed significantly since poetry.lock was last generated. "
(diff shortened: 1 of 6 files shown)


src/poetry/console/commands/check.py
```diff
@@ -158,7 +158,7 @@ def handle(self) -> int:
         if self.poetry.locker.is_locked() and not self.poetry.locker.is_fresh():
             check_result["errors"] += [
                 "pyproject.toml changed significantly since poetry.lock was last generated. "
-                "Run `poetry lock [--no-update]` to fix the lock file."
+                "Run `poetry lock` to fix the lock file."
```
edgarrmondragon, line 161: Or alternatively

```suggestion
                "Run `poetry lock [--regenerate]` to fix the lock file."
```

radoering: I think it is sufficient to mention only the option that will result in the fewest changes.

radoering (approved):
Good catch. Thanks.

PR #9937, opened 2025-01-05
chore: use project section

radoering:
I checked that this change produces the exact same metadata as before.

Further, I added `requires-poetry` because with this change we have to use 2.0 for Poetry itself.


abn (approved):

pyproject.toml
```diff
@@ -1,59 +1,68 @@
...
+    "poetry-core (==2.0.0)",
+    "build (>=1.2.1,<2.0.0)",
+    "cachecontrol[filecache] (>=0.14.0,<0.15.0)",
+    "cleo (>=2.1.0,<3.0.0)",
+    "dulwich (>=0.22.6,<0.23.0)",
+    "fastjsonschema (>=2.18.0,<3.0.0)",
+    "importlib-metadata (>=4.4) ; python_version < '3.10'",
+    "installer (>=0.7.0,<0.8.0)",
+    "keyring (>=25.1.0,<26.0.0)",
+        # packaging uses calver, so version is unclamped
```
Secrus, line 16: Indent to deep

pyproject.toml
```diff
@@ -1,59 +1,68 @@
...
+    "packaging (>=24.0)",
+    "pkginfo (>=1.12,<2.0)",
+    "platformdirs (>=3.0.0,<5)",
+    "pyproject-hooks (>=1.0.0,<2.0.0)",
+    "requests (>=2.26,<3.0)",
+    "requests-toolbelt (>=1.0.0,<2.0.0)",
+    "shellingham (>=1.5,<2.0)",
+    "tomli (>=2.0.1,<3.0.0) ; python_version < '3.11'",
+    "tomlkit (>=0.11.4,<1.0.0)",
+        # trove-classifiers uses calver, so version is unclamped
```
Secrus, line 26: ditto indent

pyproject.toml
```diff
@@ -1,59 +1,68 @@
-[tool.poetry]
+[project]
 name = "poetry"
 version = "2.0.0"
 description = "Python dependency management and packaging made easy."
-authors = ["Sébastien Eustace <sebastien@eustace.io>"]
+requires-python = ">=3.9,<4.0"
+dependencies = [
+    "poetry-core (==2.0.0)",
```
Secrus, line 7: 1. why are we pinning to a version? IIRC before 2.0 we had a `^` pin on the core.
2. are the parenthesis fine with the spec? (First I see those used)

pyproject.toml
abn: Shouldn't pre-commit have caught this?

pyproject.toml
```diff
@@ -1,59 +1,68 @@
...
-    "Secrus <b.sokorski@gmail.com>",
-]
-license = "MIT"
+    { name = "Arun Babu Neelicattu", email = "arun.neelicattu@gmail.com" },
+    { name = "Bjorn Neergaard", email = "bjorn@neersighted.com" },
+    { name = "Branch Vincent", email = "branchevincent@gmail.com" },
+    { name = "Randy Döring", email = "radoering.poetry@gmail.com" },
+    { name = "Steph Samson", email = "hello@stephsamson.com" },
+    { name = "finswimmer", email = "finswimmer77@gmail.com" },
+    { name = "Secrus", email = "b.sokorski@gmail.com" },
```
Secrus, line 41:
```suggestion
    { name = "Bartosz Sokorski", email = "b.sokorski@gmail.com" },
```
just keeping it consistent with other projects :wink:

pyproject.toml
radoering: 1. We do this since Poetry 1.3. IIRC, the idea behind this was that we will do a Poetry release anyway even if there is just a poetry-core bugfix release.
2. I think so. That is what Poetry procudes for core metadata.

pyproject.toml
Secrus: `check-toml` only validates the syntax, nothing else. An indented comment is still a valid TOML.

radoering:
It seems like we have two issues in the test environment with Poetry 2.0:

- Python versions are not detected: https://github.com/python-poetry/poetry/actions/runs/12619806450/job/35164781538?pr=9937
- A test with `system-site-packages` does not work anymore (still unclear why the poetry version influences the test): https://github.com/python-poetry/poetry/actions/runs/12619806450/job/35164781646?pr=9937

By the way, this using the project section is also required for CI because we do `poetry check` and expect no warnings. Otherwise, no tests are executed: See early failure in #9938

pyproject.toml
```diff
@@ -1,59 +1,68 @@
-[tool.poetry]
+[project]
 name = "poetry"
 version = "2.0.0"
 description = "Python dependency management and packaging made easy."
-authors = ["Sébastien Eustace <sebastien@eustace.io>"]
+requires-python = ">=3.9,<4.0"
+dependencies = [
```
abn, line 6: I dislike this more every time i see it :sob: Part of me wants to keep the old style deps for the Poetry project itself. But, we should be good ecosystem members. **sigh**.

abn (approved):

pyproject.toml
radoering: Same here. We could declare `dependencies` as `dynamic` and keep it. (This does not trigger a deprecation warning.) However, this is intended more for internal projects with custom sources. I agree with _"But, we should be good ecosystem members."_

abn: I mean, dynamic defeats the static parsing advantage. I do not think the `dynamic` should only be promoted for the internal project cases, it should really be first class citizens too (at least from Poetry's perspective). Personally, most of my projects will likely retain the Poetry syntax.

dimbleby:
please read the docs

abn:
Relevant doc is at https://python-poetry.org/docs/pyproject/#classifiers.

PR #9941, opened 2025-01-05
fix(tests): ensure only mocked system site is used

abn:
Previously, on systems where there existed multiple system-site locations the `test_system_site_packages` test case failed due to it incorrectly assuming only one site needed removal when patching the environment's `sys_path`. This change ensures all paths outside the `tmp_path` is ignored.


radoering (approved):

radoering (dismissed):
Looks like FreeBSD fails after removing the workaround.

abn:
I wrongly assumed the fix was no longer needed based on the previous failure. It seems that with the change the sqlite3 package correctly goes to the system site. I am guessing this should fix it, but lets see if the most recent change fixes it.

radoering (approved):

Issue #9943, opened 2025-01-05
`poetry sync --only=main` doesn't remove package from other groups like `poetry install --sync --only=main` do


trim21:
### Description

I'm not sure if `--with` and `--without` works as expected, but `--only` doesn't work as expected.

```text
(test) ❯❯ ~\..\test (master) uvx poetry sync --only=main
Installing dependencies from lock file

No dependencies to install or update

Installing the current project: project (0.0.0)
(test) ❯❯ ~\..\test (master) uvx poetry install --sync --only=main
The `--sync` option is deprecated and slated for removal in the next minor release after June 2025, use the `poetry sync` command instead.
Installing dependencies from lock file

Package operations: 0 installs, 0 updates, 16 removals

  - Removing botocore-stubs (1.35.92)
  - Removing google-re2-stubs (0.1.1): Cancelled
  - Removing iniconfig (2.0.0  - Removing iniconfig (2.0.0)
  - Removing lxml-stubs (0.5.1)
  - Removing msgpack-types (0.5.0)
  - Removing mypy (1.14.1)
  - Removing mypy-extensions (1.0.0)
  - Removing objprint (0.3.0)
  - Removing pluggy (1.5.0)
  - Removing pytest (8.3.4)
  - Removing types-awscrt (0.23.6)
  - Removing types-pyyaml (6.0.12.20241230)
  - Removing types-requests (2.32.0.20241016)
  - Removing types-six (1.17.0.20241205)
  - Removing types-tqdm (4.67.0.20241221)
  - Removing viztracer (1.0.0)
  ...
```

### Workarounds

use `poetry install --sync --only=main`

### Poetry Installation Method

other

### Operating System

win32

### Poetry Version

Poetry (version 2.0.0)

### Poetry Configuration

```bash session
cache-dir = "D:\\.cache\\pypoetry"
installer.max-workers = 10
installer.no-binary = null
installer.only-binary = null
installer.parallel = true
installer.re-resolve = true
keyring.enabled = true
requests.max-retries = 0
solver.lazy-wheel = true
system-git-client = false
virtualenvs.create = true
virtualenvs.in-project = true
virtualenvs.options.always-copy = false
virtualenvs.options.no-pip = false
virtualenvs.options.system-site-packages = false
virtualenvs.path = "{cache-dir}\\virtualenvs"  # D:\.cache\pypoetry\virtualenvs
virtualenvs.prompt = "{project_name}-py{python_version}"
[...]

PR #9946, opened 2025-01-05
fix: sync command should really add `--sync` option to installer

trim21:
# Pull Request Check List

Resolves: https://github.com/python-poetry/poetry/issues/9943

### DIFF
--- a/src/poetry/console/commands/sync.py
+++ b/src/poetry/console/commands/sync.py
@@ -36,5 +36,5 @@ you can set the "package-mode" to false in your pyproject.toml file.
 """
 
     @property
-    def _with_synchronization(self) -> int:
+    def _with_synchronization(self) -> bool:
         return True
--- a/src/poetry/console/commands/sync.py
+++ b/src/poetry/console/commands/sync.py
@@ -35,5 +35,6 @@ If you want to use Poetry only for dependency management but not for packaging,
 you can set the "package-mode" to false in your pyproject.toml file.
 """
 
-    def handle(self) -> int:
-        return self._handle_install(with_synchronization=True)
+    @property
+    def _with_synchronization(self) -> int:
+        return True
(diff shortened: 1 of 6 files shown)


radoering:
> not sure how to test this...

You can mock `installer` and check that `requires_synchronization` is called with `True`.

tests/console/commands/test_sync.py
```diff
@@ -28,3 +31,14 @@ def test_sync_option_is_passed_to_the_installer() -> None:
 def test_sync_option_not_available(tester: CommandTester) -> None:
     with pytest.raises(CleoNoSuchOptionError):
         tester.execute("--sync")
+
+
+def test_synced_installer(tester: CommandTester, mocker: MockerFixture) -> None:
```
radoering, line 36: Can you add the same test for `self sync` in [tests/console/commands/self/test_sync.py](https://github.com/python-poetry/poetry/blob/main/tests/console/commands/self/test_sync.py) (and thereby check if the fix also works for `self sync`), please?

radoering (changes requested):

src/poetry/console/commands/sync.py
```diff
@@ -34,3 +34,6 @@ class SyncCommand(InstallCommand):
 If you want to use Poetry only for dependency management but not for packaging,
 you can set the "package-mode" to false in your pyproject.toml file.
 """
+
+    def handle(self) -> int:
+        return self._handle_install(True)
```
Secrus, line 39:
```suggestion
        return self._handle_install(with_synchronization=True)
```
A small nitpick, but let's avoid nameless boolean calls.

tests/console/commands/test_sync.py
Secrus: It would be nice if you did. The tests are practically the same and it would allow us to avoid having to do the extra work of doing it ourselves.

radoering: Fair enough. I extended the fix so that it also works for `self sync`.

radoering (approved):

Issue #9972, opened 2025-01-06
mark dry run mode in poetry output


piotr-kubiak:
### Issue Kind

Change in current behaviour

### Description

Currently, there is no difference in output between `poetry install --dry-run` and `poetry install`.

```
$ poetry install --dry-run
Installing dependencies from lock file

Package operations: 4 installs, 0 updates, 0 removals

  - Installing mdurl (0.1.2)
  - Installing pygments (2.18.0)
  - Installing markdown-it-py (3.0.0)
  - Installing rich (13.9.4)

Installing the current project: my-package (0.1.0)

$ poetry install
Installing dependencies from lock file

Package operations: 4 installs, 0 updates, 0 removals

  - Installing mdurl (0.1.2)
  - Installing pygments (2.18.0)
  - Installing markdown-it-py (3.0.0)
  - Installing rich (13.9.4)

Installing the current project: my-package (0.1.0)
```

### Impact

This may be confusing, since observing only the output of poetry command, there is no way of telling if it was actually executed, or only simulated.

### Workarounds

You need to observe both the command used (to check for `dry-run` flag) and the output (to see an actual output) in order to check if for example a package was installed. This makes scripting poetry output more challenging.

PR #9973, opened 2025-01-06
mark dry run in Poetry output

piotr-kubiak:
# Pull Request Check List

A proposed fix for #9972.


abn:
If I recall correctly, the output is intentionally similar. I don't think there is a need for this change. Happy to hear from others before deciding.

If we do proceed, I do think the message will have to change.

piotr-kubiak:
@abn Sorry to hear that. I explained my rationale for this change in the original issue #9972.

radoering:
I do not have a strong opinion. I see benefits on both sides.

Secrus:
I don't mind adding some markers about the simulation mode, but I agree the message should be different.

piotr-kubiak:
I am more than happy to change the message, but not sure what it should be.
Here are some examples in other tools:

```
$ rsync --dry-run -av foo/ bar/
sending incremental file list
./

sent 95 bytes  received 19 bytes  228.00 bytes/sec
total size is 0  speedup is 0.00 (DRY RUN)

$ pip install --dry-run six
Collecting six
  Using cached six-1.17.0-py2.py3-none-any.whl.metadata (1.7 kB)
Using cached six-1.17.0-py2.py3-none-any.whl (11 kB)
Would install six-1.17.0

$ git clean -fd --dry-run
Would remove bar/
Would remove foo/

```

Issue #9971, opened 2025-01-06
`poetry check`: distinguish warnings from errors?


seub:
### Issue Kind

Brand new capability

### Description

`poetry check` exits with error code `1` even when there are only warnings and no hard errors.

This is standard UNIX behavior, but it makes it impossible to distinguish warnings from hard errors.

### Example scenario

When upgrading Poetry to 2.0, my project.toml file caused `poetry check` to raise warnings, and exit with exit code `1`:

<img width="1107" alt="Screenshot 2025-01-06 at 3 39 20 PM" src="https://github.com/user-attachments/assets/a7dd23e5-f22a-4e19-bf99-910b38a05cd7" />

I don't have a problem with that, but it seems inconvenient that there is no way to check for warnings vs hard errors.

### Suggested solutions

Here are two ideas:

1. Exit with error code `2` when there are only warnings and no hard errors
2. Add a flag `--soft-warnings` causing `poetry check` to pass (exit code `0`) when there are only warnings and no hard errors

### Workarounds

Suboptimal workaroung:

```
poetry check || echo "Some errors or warnings were raised by 'poetry check'"
```

gridhead:
As mentioned before here https://github.com/python-poetry/poetry/issues/9965#issuecomment-2574687981,

> I'm going to chime in having the same observations as @Zlopez here with a couple of my projects. 

> If those are warnings - a non-zero exit code coming as a result of an impending release breaks things.

> Context. https://github.com/fedora-infra/mdapi/actions/runs/12646184815/job/35236500953

> ![image](https://github.com/user-attachments/assets/bc9b883b-9d2f-45db-a40d-61696889b25e)

> Context. https://github.com/fedora-infra/firmitas/actions/runs/12624139412/job/35173984552#step:5:21

> ![image](https://github.com/user-attachments/assets/d21fe651-a60e-40c4-8e5b-2bc6964af2fd)

przemyslaw-sitek:
> `poetry check` exits with error code `1` even when there are only warnings and no hard errors.
> 
> This is standard UNIX behavior, but it makes it impossible to distinguish warnings from hard errors.

I disagree. Most tools report success exit code if no errors were encountered, and tools that do, that feature requires explicit opt-in. For example `gcc` returns 0 exit code unless `-Werror` (_Make all warnings into errors_) option was explicitly provided.

ralphie0112358:
Thanks for the quick action on this one.  When could we expect poetry release with the fix?

PR #9983, opened 2025-01-07
fix(console): do not fail on warnings on check command

finswimmer:
# Pull Request Check List

Old behavior is retained by new parameter `--strict`.

Resolves: #9971

## Summary by Sourcery

Add the `--strict` option to the `check` command to make it fail if warnings are reported. By default, the command will only report warnings and exit with a success code.

New Features:
- Introduce the `--strict` option to the `poetry check` command.

Tests:
- Add tests for the `--strict` option.

### DIFF
--- a/src/poetry/console/commands/check.py
+++ b/src/poetry/console/commands/check.py
@@ -29,6 +29,11 @@ class CheckCommand(Command):
             "Checks that <comment>poetry.lock</> exists for the current"
             " version of <comment>pyproject.toml</>.",
         ),
+        option(
+            "strict",
+            None,
+            "Fail if check reports warnings.",
+        ),
     ]
 
     def _validate_classifiers(
@@ -161,10 +166,15 @@ class CheckCommand(Command):
                 "Run `poetry lock` to fix the lock file."
             ]
 
+        return_code = 0
+
+        if check_result["errors"] or (
+            check_result["warnings"] and self.option("strict")
+        ):
+            return_code = 1
+
         if not check_result["errors"] and not check_result["warnings"]:
             self.info("All set!")
-
-            return 0
 
         for error in check_result["errors"]:
             self.line_error(f"<error>Error: {error}</error>")
@@ -172,4 +182,4 @@ class CheckCommand(Command):
         for error in check_result["warnings"]:
             self.line_error(f"<warning>Warning: {error}</warning>")
 
-        return 1
+        return return_code
--- a/tests/console/commands/test_check.py
+++ b/tests/console/commands/test_check.py
@@ -76,15 +76,26 @@ All set!
     assert tester.io.fetch_output() == expected
 
 
+@pytest.mark.parametrize(
+    ["args", "expected_status"],
+    [
+        ([], 0),
+        (["--strict"], 1),
+    ],
+)
 def test_check_valid_legacy(
-    mocker: MockerFixture, tester: CommandTester, fixture_dir: FixtureDirGetter
+    args: list[str],
+    expected_status: int,
+    mocker: MockerFixture,
+    tester: CommandTester,
+    fixture_dir: FixtureDirGetter,
 ) -> None:
     mocker.patch(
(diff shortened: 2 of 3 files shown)


tests/console/commands/test_check.py
```diff
@@ -76,15 +76,26 @@ def test_check_valid(tester: CommandTester) -> None:
     assert tester.io.fetch_output() == expected
```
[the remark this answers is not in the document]
finswimmer: Is covered by other tests.

dimbleby:
I'm inclined to agree with the camp that thinks that warnings should not have resulted in exit code 1 in the first place.  Warnings are only warnings, only errors are errors.

so I'd vote that way rather than throw more flags at it

finswimmer:
> Perhaps a `--strict` flag would make sense, so that people can choose to error on warnings, but I think the current default is the wrong way round

I agree with you. But @Secrus and I had a short conversation about it on discord. Changing the current default, would be a breaking change. Of course one could argue, the current behavior is a bug and the change is only a fix ... Would like to hear opinions from the other maintainers.

dimbleby:
"breaking" only in rather a weak sense of not failing where it could have done.  I doubt that this will actually *break* anything for anyone, though it might cause them to not notice warnings.

imo 2.0 is sufficiently recent and the current behaviour is sufficiently unpopular that it is ok to rip the bandage and do what we think is right.  Having to wait for poetry 3.0 to address this paper cut of backwards defaults seems worse in the long run.

(unless the team anticipates more regular major releases and 3.0 is relatively close...)

radoering:
Pedantically speaking, it feels like a breaking change. However, I agree that in the long run, it is better to consider it as bugfix and do it the way it feels right (`--strict`).

finswimmer:
Ok, I changed the behavior.

src/poetry/console/commands/check.py
```diff
@@ -161,15 +166,20 @@ def handle(self) -> int:
...
+        return_code = 0
+
+        if check_result["errors"] or (
+            check_result["warnings"] and self.option("strict")
+        ):
+            return_code = 1
 
-            return 0
+        if return_code == 0:
+            self.info("All set!")
```
radoering, line 177: Should this be printed based on the return code or only if there are no warnings and errors?

finswimmer: Good question :man_shrugging: I don't know. Frontend isn't my strength :grin:

radoering: I just noticed sourcery had the same concern 😆: https://github.com/python-poetry/poetry/pull/9983#discussion_r1905345114

I think the "All set!" message is just there because otherwise there would be no output if there are no errors and no warnings.

abn:
@sourcery-ai review

src/poetry/console/commands/check.py
jeremiah-snee-openx: One possible risk of removing `All Set!` from output with warnings and no errors is if anyone is checking logs for an all set in their pipeline. 

The config even with warnings is still valid so I think leaving this after the warnings should be fine.

jeremiah-snee-openx (approved):
LGTM

src/poetry/console/commands/check.py
abn: I agree with sourcery that we should only present "All set!" despite the strictness, when there are no check results at all.

finswimmer: Done.

tests/console/commands/test_check.py
```diff
@@ -76,15 +76,26 @@ def test_check_valid(tester: CommandTester) -> None:
...
 
 
+@pytest.mark.parametrize(
+    ["args", "expected_status"],
+    [
+        ([], 0),
```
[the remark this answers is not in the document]
abn: You are hallucinating, there is no such option as `--with-errors` for the `check` command.

Issue #9975, opened 2025-01-07
Poetry removes installed extras when adding a new dependency


pronovic:
### Description

I've been converting my projects to Poetry v2.0.0.  I noticed that Poetry now removes extra dependencies (`[project.optional-dependencies]`) from the virtualenv when adding a new dependency.  This is surprising behavior, and I don't ever remember seeing it with earlier versions of Poetry.  If it's intentional, it seems like it should be documented.

I think what I'm demonstrating below should be easy to reproduce in pretty much any project.  However, if you want a fully reproducible example, follow the instructions at [pronovic/cookiecutter-pypi](https://github.com/pronovic/cookiecutter-pypi) to create a sample repo &mdash; just use all of the defaults.  That generates a project with the `pyproject.toml` I've attached.  Initialize the git repository per the repo's README.md (needed due to the poetry-dynamic-versioning plugin), and then follow the steps below.

If you start with `poetry sync`, you get the plugin and dev dependencies installed as expected:

```
Git Status: [ master ✲ ✓ ]
mars:~/projects/repos/sample-project> poetry sync
Creating virtualenv sample-project in /home/pronovic/projects/repos/sample-project/.venv
Ensuring that the Poetry plugins required by the project are available...
The following Poetry plugins are required by the project but are not installed in Poetry's environment:
  - poetry-dynamic-versioning[plugin] (>=1.5.0,<2.0.0)
Installing Poetry plugins only for the current project...
Updating dependencies
Resolving dependencies... (1.2s)

Package operations: 4 installs, 0 updates, 0 removals

  - Installing markupsafe (3.0.2)
  - Installing dunamai (1.23.0): Installing...
  - Installing dunamai (1.23.0)
  - Installing jinja2 (3.1.5)
  - Installing poetry-dynamic-versioning (1.5.0): Installing...
  - Installing poetry-dynamic-versioning (1.5.0)

Writing lock file

Updating dependencies
Resolving dependencies... (1.1s)

Package operations: 28 installs, 0 updates, 0 removals

  - Installing distlib (0.3.9)
  - Installing filelock (3.16.1)
  - Installing iniconfig (2.0.0)
  - Installing packaging (24.2)
  - Installing platformdirs (4.3.6)
  - Installing pluggy (1.5.0)
  - Installing astroid (3.3.8)
  - Installing cfgv (3.4.0)
  - Installing click (8.1.8)
  - Installing dill (0.3.9)
  - Installing identify (2.6.5)
[...]

PR #10016, opened 2025-01-11
fix: `update`, `add` and `remove` shall not uninstall extra dependencies

radoering:
With this change unrequested extras dependencies will also be kept when running `install` and are only removed when running `sync`!

# Pull Request Check List

Resolves: #9975
Requires: #10014

## Summary by Sourcery

Bug Fixes:
- Dependencies with extras are no longer uninstalled when those extras are not explicitly requested during installation or update.

### DIFF
--- a/docs/cli.md
+++ b/docs/cli.md
@@ -204,11 +204,13 @@ poetry install -E mysql -E pgsql
 poetry install --all-extras
 ```
 
-Any extras not specified will always be removed.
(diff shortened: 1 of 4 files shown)


radoering:
> * Please update the documentation to reflect this behavior change regarding how extra dependencies are handled during install/update/sync operations.

Good point.

abn (approved):

Issue #9978, opened 2025-01-07
poetry 2 install command can not find pyproject.toml with --only and --directory parameters


konstantin-mueller:
### Description

I have a pyproject.toml with `package-mode = false` and I'm copying it into a docker image. I installed the dependencies like this in previous versions of poetry:
```
poetry install --only main --no-cache -C /tmp
```
/tmp contains the pyproject.toml and poetry.lock files.

The command fails with poetry 2. Error message:
```
Poetry could not find a pyproject.toml file in / or its parents
```
The current directory during docker build is /. Poetry should change the directory to /tmp because of the  `-C /tmp` parameter.

If I change the order of the parameters like this:
```
poetry install -C /tmp --only main --no-cache
```
then poetry successfully installs the dependencies.

### Workarounds

Specify -C parameter before the --only parameter

### Poetry Installation Method

pip

### Operating System

Alpine 3.21

### Poetry Version

2.0.0

### Poetry Configuration

```bash session
cache-dir = "/home/user/.cache/pypoetry"
installer.max-workers = null
installer.no-binary = null
installer.only-binary = null
installer.parallel = true
installer.re-resolve = true
keyring.enabled = true
requests.max-retries = 0
solver.lazy-wheel = true
system-git-client = false
virtualenvs.create = true
virtualenvs.in-project = null
virtualenvs.options.always-copy = false
virtualenvs.options.no-pip = false
virtualenvs.options.system-site-packages = false
virtualenvs.path = "{cache-dir}/virtualenvs"  # /home/user/.cache/pypoetry/virtualenvs
virtualenvs.prompt = "{project_name}-py{python_version}"
virtualenvs.use-poetry-python = false
```

### Poetry Runtime Logs

```bash session
Stack trace:

  12  /opt/doc_venv/lib/python3.13/site-packages/cleo/application.py:327 in run
       325│ 
       326│             try:
     → 327│                 exit_code = self._run(io)
       328│             except BrokenPipeError:
       329│                 # If we are piped to another process, it may close early and send a

  11  /opt/doc_venv/lib/python3.13/site-packages/poetry/console/application.py:236 in _run
       234│ 
       235│         with directory(self._working_directory):
     → 236│             exit_code: int = super()._run(io)
       237│ 
       238│         return exit_code

  10  /opt/doc_venv/lib/python3.13/site-packages/cleo/application.py:431 in _run
       429│             io.input.interactive(interactive)
       430│ 
[...]

PR #10021, opened 2025-01-11
fix(application): handle global options correctly

abn:
Previously, the application implementation did not handle global options correctly when positioned after command options. This lead to incorrect cached working directory values prior to switching application context.

Resolves: #9978

## Summary by Sourcery

Bug Fixes:
- Fixed an issue where global options were not handled correctly when positioned after command options, leading to incorrect cached working directory values.

### DIFF
--- a/src/poetry/console/application.py
+++ b/src/poetry/console/application.py
@@ -4,7 +4,6 @@ import logging
 import re
 
 from contextlib import suppress
-from functools import cached_property
 from importlib import import_module
 from pathlib import Path
 from typing import TYPE_CHECKING
@@ -25,6 +24,7 @@ from poetry.utils.helpers import directory
 
 if TYPE_CHECKING:
     from collections.abc import Callable
+    from typing import Any
 
     from cleo.events.event import Event
     from cleo.io.inputs.argv_input import ArgvInput
@@ -103,6 +103,8 @@ class Application(BaseApplication):
         self._disable_plugins = False
         self._disable_cache = False
         self._plugins_loaded = False
+        self._working_directory = Path.cwd()
+        self._project_directory = self._working_directory
 
         dispatcher = EventDispatcher()
         dispatcher.add_listener(COMMAND, self.register_command_loggers)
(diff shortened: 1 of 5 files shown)


abn:
The issue reported in #9978 can be reproduced and the fix in this PR tested using the following snippet.

```sh
podman run --rm -i --entrypoint bash python:latest <<EOF
set -x
python -m pip install --disable-pip-version-check --root-user-action ignore -q poetry
pushd /tmp
poetry new foobar
poetry add -C /tmp/foobar httpx
poetry show --only main -C /tmp/foobar || :
python -m pip install --disable-pip-version-check --root-user-action ignore -q --force 'git+https://github.com/python-poetry/poetry.git@refs/pull/10021/head'
poetry show --only main -C /tmp/foobar
EOF
```

abn:
@sourcery-ai review

src/poetry/utils/helpers.py
```diff
@@ -258,6 +258,18 @@ def paths_csv(paths: list[Path]) -> str:
...
+def ensure_path(path: str | Path, is_directory: bool = False) -> Path:
+    if isinstance(path, str):
+        path = Path(path)
+
+    if path.exists() and path.is_dir() == is_directory:
+        return path
+
+    raise ValueError(
+        f"Specified path '{path}' is not a valid {'directory' if is_directory else 'file'}."
+    )
```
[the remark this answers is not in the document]
abn: Unnecessary complexity for not much gain.

src/poetry/console/application.py
```diff
@@ -237,6 +228,52 @@ def _run(self, io: IO) -> int:
...
+
+        if not io.input.has_parameter_option(values):
+            return default
+
+        if option.is_flag():
+            return True
+
+        return io.input.parameter_option(values=values, default=default)
+
+    def _configure_custom_application_options(self, io: IO | None) -> None:
```
Secrus, line 250: None-able `io` makes no sense. you are calling this method from `_run`, where you have your IO configured, and `_run` is being passed an IO object. So you are always passing `IO` object in, never `None`

src/poetry/console/application.py
```diff
@@ -237,6 +228,52 @@ def _run(self, io: IO) -> int:
 
         return exit_code
 
+    def _option_get_value(self, io: IO, name: str, default: Any) -> Any:
```
Secrus, line 231: nitpick: we could probably do `default: T` and `cast()` return to T. It's not worth the work, I'm just mentioning it because I hate seeing `Any` used.

abn: I originally had that, but mypy does not like that since default passed to parameter_option is of type Any.

src/poetry/console/application.py
abn: Fair. I think I instinctively did a `if io is None` check in the code since the `self._io` is `IO | None`. Then sourcery complained about it, so I added this. Now removing it and the `if io is None` check.

abn:
Changed the path resolution for `--project` to avoid doing a hard cwd switch.

Secrus (approved):

PR #10026, opened 2025-01-12
refactor(env): move mock and null envs to tests

abn:
## Summary by Sourcery

Tests:
- Refactor mock and null environments into helper functions for testing.


abn:
@sourcery-ai review

abn:
Seems plugins are using this in their tests suites as well. Better to leave things where they are for now, until we figure out a cleaner way of doing this such that we do not break things for third-party plugins.

Issue #10024, opened 2025-01-12
poetry env activate does not work on windows when python is not in environment


teddy661:
### Description

Upgraded to poetry 2.0.1 attempted to launch a shell using poetry env activate and receive the following error message for powershell and cmd: 

Python installation options are:  for all users; Do not add Python to environment variables ; 

the command poetry shell worked correctly for this configuration

PWSH
PS C:\Users\xxx.xxx\Desktop\automation_python> poetry env activate
Python was not found; run without arguments to install from the Microsoft Store, or disable this shortcut from Settings > Manage App Execution Aliases.
& "C:\Users\xxx.xxx\Desktop\automation_python\.venv\Scripts\Activate.ps1"
PS C:\Users\xxx.xxx\Desktop\automation_python>

CMD
C:\Users\xxx.xxx\Desktop\automation_python>poetry env activate
Python was not found; run without arguments to install from the Microsoft Store, or disable this shortcut from Settings > Manage App Execution Aliases.
"C:\Users\xxx.xxx\Desktop\automation_python\.venv\Scripts\activate.bat"

C:\Users\xxx.xxx\Desktop\tautomation_python>

### Workarounds

Calling the scripts directly from the shell works. 

CMD
C:\Users\xxx.xxx\Desktop\automation_python>"C:\Users\xxx.xxx\Desktop\automation_python\.venv\Scripts\activate.bat"

(tautomation-python-py3.13) C:\Users\xxx.xxx\Desktop\automation_python>

PWSH
PS C:\Users\xxx.xxx\Desktop\automation_python> & "C:\Users\xxx.xxx\Desktop\tc_dev_automation_python\.venv\Scripts\Activate.ps1"
(automation-python-py3.13) PS C:\Users\xxx.xxx\Desktop\automation_python>

### Poetry Installation Method

install.python-poetry.org

### Operating System

Windows 11 10.0.26100

### Poetry Version

Poetry (version 2.0.1)

### Poetry Configuration

```bash session
cache-dir = "C:\\Users\\xxx.xxx\\AppData\\Local\\pypoetry\\Cache"
installer.max-workers = null
installer.no-binary = null
installer.only-binary = null
installer.parallel = true
installer.re-resolve = true
keyring.enabled = true
requests.max-retries = 0
solver.lazy-wheel = true
system-git-client = false
virtualenvs.create = true
virtualenvs.in-project = null
virtualenvs.options.always-copy = false
virtualenvs.options.no-pip = false
[...]

PR #10027, opened 2025-01-13
doc(cli): add missing docs for env command

abn:
This helps mitigate issues like #10024.

## Summary by Sourcery

Documentation:
- Add documentation for the `env activate`, `env info`, `env list`, `env remove`, and `env use` commands.

### DIFF
--- a/docs/cli.md
+++ b/docs/cli.md
@@ -890,6 +890,58 @@ associated with a specific project.
 
 See [Managing environments]({{< relref "managing-environments" >}}) for more information about these commands.
 
+### env activate
+
+The `env activate` command prints the command to activate a virtual environment in your current shell.
+
+{{% note %}}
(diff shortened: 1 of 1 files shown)


docs/cli.md
```diff
@@ -890,6 +890,58 @@ associated with a specific project.
...
+
+* `--all`: Remove all managed virtual environments associated with the project.
+
+### use
+
+The `env use` command activates or creates a new virtualenv for the current project.
+
+#### Arguments
+
+* `python`: The python executable to use.
```
Secrus, line 943: I'd add a note that it can be either a version number (`3.9`, `3.12`) or a path to the binary.

radoering: Then, you have to add that it cannot be a version number on Windows.

Secrus (approved):

PR #10060, opened 2025-01-16
Apply use-pathlib rules

Secrus:
Since we use `os.path.join` as a string glue method, I silenced those cases.


Secrus:
@sourcery-ai review

src/poetry/config/config.py
```diff
@@ -110,7 +110,7 @@ class Config:
         "virtualenvs": {
             "create": True,
             "in-project": None,
-            "path": os.path.join("{cache-dir}", "virtualenvs"),
+            "path": os.path.join("{cache-dir}", "virtualenvs"),  # noqa: PTH118
```
abn, line 113: Any reason why we can't replace this?

src/poetry/utils/env/env_manager.py
```diff
@@ -493,14 +493,14 @@ def create_venv(
         # but others can symlink *to* the venv Python,
         # so we can't just use sys.executable.
         # So we just check every item in the symlink tree (generally <= 3)
-        p = os.path.normcase(sys.executable)
+        p = Path(os.path.normcase(sys.executable))
```
abn, line 496: Is normcase required with Path?

src/poetry/utils/env/env_manager.py
```diff
@@ -493,14 +493,14 @@ def create_venv(
...
         paths = [p]
-        while os.path.islink(p):
-            p = os.path.normcase(os.path.join(os.path.dirname(p), os.readlink(p)))
+        while p.is_symlink():
+            p = Path(os.path.normcase(p.parent / p.readlink()))
             paths.append(p)
 
         p_venv = os.path.normcase(str(venv))
-        if any(p.startswith(p_venv) for p in paths):
+        if any(str(p).startswith(p_venv) for p in paths):
```
abn, line 503: Shouldn't this use relative?

tests/console/commands/self/test_add_plugins.py
```diff
@@ -194,7 +194,7 @@ def test_add_existing_plugin_warns_about_no_operation(
     installed: TestRepository,
 ) -> None:
     pyproject = SelfCommand.get_default_system_pyproject_file()
-    with open(pyproject, "w", encoding="utf-8", newline="") as f:
+    with pyproject.open("w", encoding="utf-8", newline="") as f:
```
abn, line 197: Can we use write_text? Similar in a few other places.

abn (commented):
Some initial comments.

tests/console/commands/self/test_add_plugins.py
radoering: Only after dropping support for Python 3.9, I assume:

> Changed in version 3.10: The newline parameter was added.

https://docs.python.org/3/library/pathlib.html#pathlib.Path.write_text

src/poetry/config/config.py
Secrus: from what I understand it would be `str(Path(...))` which doesn't look like a good use of resources, while `os.path` operates on strings all the way

tests/console/commands/self/test_add_plugins.py
Secrus: Exactly this. I can add a TODO for when we drop 3.9 to change that

src/poetry/config/config.py
abn: Add a comment maybe to avoid someone swapping it later.

But I reckon since it's not always a valid path (eg: template) it's okay.

PR #10062, opened 2025-01-16
fix(keyring): make keyring unlock thread safe

abn:
When using `poetry install`, Poetry executor prefers to execute  operations using a `concurrent.futures.ThreadPoolExecutor`. This can lead to multiple credential store unlock requests to be dispatched simultaneously.

If the credential store is locked, and a user either intentionally or unintentionally dismisses the prompt to provide the store password; or the dbus messages fail to launch the prompter the Poetry user can experience, what can appear as a "hang". This in fact can be several threads competing with each other waiting for a dbus event indefinitely; this is a consequence of how Poetry uses keyring.

This change introduces the following:

- pre-flight check for installer commands that ensures keyring unlocks only once for the duration of the command
- improved logging of keyring unlock event and/or failure
- thread safe caching of `PasswordManager.<use_keyring|keyring>`
- the `@atomic_cached_property` decorator

This change does not address the following:

- handling cases where dbus message fails or prompter is blocked
- authentication of a locked keyring in a non-tty session

I am also open to consider an alternative for `@atomic_cached_property` as I am not so comfortable having to implement something like this unless we really need it.

Resolves: #8623

## Summary by Sourcery

Make keyring unlock thread-safe and improve logging.

Bug Fixes:
- Fix a hang that could occur when multiple threads attempt to unlock the keyring simultaneously.

Enhancements:
- Add a pre-flight check to ensure keyring unlocks only once during a command.
- Improve logging of keyring unlock events and failures.
- Introduce thread-safe caching for `PasswordManager` properties.

Tests:
- Update tests to reflect changes in keyring handling.

### DIFF
--- a/src/poetry/console/commands/installer_command.py
+++ b/src/poetry/console/commands/installer_command.py
@@ -4,9 +4,12 @@ from typing import TYPE_CHECKING
 
 from poetry.console.commands.env_command import EnvCommand
 from poetry.console.commands.group_command import GroupCommand
+from poetry.utils.password_manager import PoetryKeyring
 
 
 if TYPE_CHECKING:
+    from cleo.io.io import IO
+
     from poetry.installation.installer import Installer
 
 
@@ -30,3 +33,7 @@ class InstallerCommand(GroupCommand, EnvCommand):
 
     def set_installer(self, installer: Installer) -> None:
         self._installer = installer
+
+    def execute(self, io: IO) -> int:
+        PoetryKeyring.preflight_check(io, self.poetry.config)
+        return super().execute(io)
--- a/src/poetry/utils/password_manager.py
+++ b/src/poetry/utils/password_manager.py
@@ -7,11 +7,15 @@ import logging
 from contextlib import suppress
 from typing import TYPE_CHECKING
 
+from poetry.config.config import Config
+from poetry.utils.threading import atomic_cached_property
+
 
 if TYPE_CHECKING:
-    from keyring.backend import KeyringBackend
+    import keyring.backend
 
-    from poetry.config.config import Config
+    from cleo.io.io import IO
+
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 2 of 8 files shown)


src/poetry/utils/threading.py
```diff
@@ -0,0 +1,68 @@
...
+
+T = TypeVar("T")
+C = TypeVar("C", bound=object)
+
+
+class AtomicCachedProperty(functools.cached_property[T]):
+    def __init__(self, func: Callable[[C], T]) -> None:
+        super().__init__(func)
+        self._semaphore = threading.BoundedSemaphore()
+        self._locks: dict[object, threading.Lock] = {}
```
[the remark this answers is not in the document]
abn: Great suggestion. Done.

abn:
Updated the implementation to only unlock for installer commands.

src/poetry/console/commands/installer_command.py
```diff
@@ -30,3 +33,7 @@ def installer(self) -> Installer:
 
     def set_installer(self, installer: Installer) -> None:
         self._installer = installer
+
+    def execute(self, io: IO) -> int:
+        PoetryKeyring.preflight_check(io)
```
radoering, line 38:
```suggestion
        PoetryKeyring.preflight_check(io, self.poetry.config)
```

Otherwise, we are neglecting the local config IIRC.

src/poetry/utils/password_manager.py
```diff
@@ -138,20 +173,30 @@ def backend_is_valid(backend: KeyringBackend) -> bool:
...
             logger.debug("No valid keyring backend was found")
             return False
-        else:
-            logger.debug(f"Using keyring backend {backend.name!r}")
-            return True
+
+        logger.debug(f"Using keyring backend {backend.name!r}")
+
+        try:
+            # unfortunately there is no clean of checking if keyring is unlocked
```
radoering, line 180:
```suggestion
            # unfortunately there is no clean way of checking if keyring is unlocked
```

tests/utils/test_authenticator.py
```diff
@@ -96,29 +96,39 @@ def test_authenticator_ignores_locked_keyring(
...
 ) -> None:
     caplog.set_level(logging.DEBUG, logger="poetry.utils.password_manager")
+    spy_get_credential = mocker.spy(PoetryKeyring, "get_credential")
+    spy_get_password = mocker.spy(PoetryKeyring, "get_password")
     authenticator = Authenticator()
     authenticator.request("get", "https://foo.bar/files/foo-0.1.0.tar.gz")
 
     request = http.last_request()
     assert request.headers["Authorization"] is None
-    assert "Keyring foo.bar is locked" in caplog.messages
```
radoering, line 106: Shouldn't there still be a message that the keyring is locked in the output?

radoering (changes requested):

tests/utils/test_authenticator.py
abn: Yeah "Accessing keyring failed during availability check" will add.

radoering (approved):

Issue #10085, opened 2025-01-21
Poetry 2 POETRY_HTTP_BASIC_ issues


black-snow:
### Description

This is going to be a very bad bug report as I have little information to provide, sadly.

When I upgrade from 1.8.4 to 2.0.1 my pipeline fails to pull from a private registry. It's an Azure DevOps pipeline and I use the PipAuthenticate task and then export the password as POETRY_HTTP_BASIC_registry-name_PASSWORD.

I also provide the password via env locally and that works fine with Poetry 2. I am clueless to why it breaks the pipeline - perhaps someone has an idea.

### Workarounds

none

### Poetry Installation Method

pip

### Operating System

some Ubuntu, I guess

### Poetry Version

2.0.1

### Poetry Configuration

```bash session
the default, I guess - can print it out if it helps
```

### Example pyproject.toml

```TOML

```

### Poetry Runtime Logs

<details>
  <summary>poetry-runtime.log</summary>
  

  ```
Source (X): Authorization error accessing https://pkgs.dev.azure.com/X/_packaging/X/pypi/simple/X/

  RuntimeError

  Unable to find installation candidates for X (0.0.1rc5)

  at .venv/lib/python3.13/site-packages/poetry/installation/chooser.py:86 in choose_for
       82│ 
       83│             links.append(link)
       84│ 
       85│         if not links:
    →  86│             raise RuntimeError(f"Unable to find installation candidates for {package}")
       87│ 
       88│         # Get the best link
       89│         chosen = max(links, key=lambda link: self._sort_key(package, link))
       90│ 

Cannot install X.

  ```
</details>

black-snow:
@abn I suspected it might be that but was honestly too lazy to actually test it (since it was working without it before ...?). I set it to some dummy value and that seems to do the trick.

... thanks for your time!

(I'm still curious if there was a change in Poetry 2 that explains why it worked previously but does no longer do so.)

abn:
The change was introduced in #9850 which was released in 2.0.0. See also https://github.com/python-poetry/poetry/pull/9850#issuecomment-2480809622.

I have also raised #10088 to revert most of that change and to allow empty usernames in the next release. But this is up for discussion.

PR #10088, opened 2025-01-21
feat(auth): allow passwords with empty usernames

abn:
This change largely reverts the decision made in b544ed5 as it seems that it is commonplace, albeit not best practice, for private indices to require or document that users use an empty username. The change in behaviour seems to surprise users migrating from 1.x (#10085).

## Summary by Sourcery

Allow empty usernames for private repositories.

Bug Fixes:
- Fix unexpected behavior when migrating from Poetry 1.x with private indices requiring empty usernames.

Enhancements:
- Improve compatibility with private indices that require or document empty usernames for authentication.

### DIFF
--- a/src/poetry/utils/password_manager.py
+++ b/src/poetry/utils/password_manager.py
@@ -27,7 +27,7 @@ class HTTPAuthCredential:
 
 
 class PoetryKeyring:
-    # some private sources expect tokens to be provided as password with that can be empty
+    # some private sources expect tokens to be provided as passwords with empty userames
     # we use a fixed literal to ensure that this can be stored in keyring (jaraco/keyring#687)
     #
     # Note: If this is changed, users with passwords stored with empty usernames will have to
--- a/src/poetry/utils/password_manager.py
+++ b/src/poetry/utils/password_manager.py
@@ -27,6 +27,13 @@ class HTTPAuthCredential:
 
 
 class PoetryKeyring:
+    # some private sources expect tokens to be provided as password with that can be empty
+    # we use a fixed literal to ensure that this can be stored in keyring (jaraco/keyring#687)
+    #
+    # Note: If this is changed, users with passwords stored with empty usernames will have to
+    # re-add the config.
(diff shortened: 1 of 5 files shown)


tests/utils/test_password_manager.py
```diff
@@ -65,8 +64,8 @@ def test_get_http_auth(
 
     if is_valid:
         assert auth is not None
-        assert auth.username == username
-        assert auth.password == password
+        assert auth.username == (username or None)
+        assert auth.password == (password or None)
```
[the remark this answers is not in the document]
abn: The test case is not covering the input validity, but rather functional validity. Current cases are sufficient.

src/poetry/utils/password_manager.py
```diff
@@ -27,6 +27,13 @@ class HTTPAuthCredential:
 
 
 class PoetryKeyring:
+    # some private sources expect tokens to be provided as password with that can be empty
```
radoering, line 30: > with that can be empty

Is there a word (like "username") missing here?

tests/utils/test_authenticator.py
```diff
@@ -170,7 +171,7 @@ def test_authenticator_ignores_empty_strings_as_default_username(
     authenticator.request("get", "https://foo.bar/files/foo-0.1.0.tar.gz")
 
     request = http.last_request()
-    assert request.headers["Authorization"] is None
+    assert request.headers["Authorization"] == "Basic OmJhcg=="
```
radoering, line 174:
```suggestion
    basic_auth = base64.b64encode(b":bar").decode()
    assert request.headers["Authorization"] == f"Basic {basic_auth}"
```

I prefer the previous version that avoids the "random" string.

radoering (changes requested):

src/poetry/utils/password_manager.py
abn, line 30:
```suggestion
    # some private sources expect tokens to be provided as passwords with empty userames
```

radoering (approved):

PR #9392, opened 2024-05-06
remove pip-based installer

dimbleby:
following through on #8987 

remaining uses of pip in poetry

- installation of the managed package, when that package has a build script: [here](https://github.com/python-poetry/poetry/blob/43dd8cce4f04120ee698609fa0a0a66063e80dcd/src/poetry/masonry/builders/editable.py#L102)
- uninstalling packages: [here](https://github.com/python-poetry/poetry/blob/43dd8cce4f04120ee698609fa0a0a66063e80dcd/src/poetry/installation/executor.py#L592)

desirable to remove both of these in the long run...

Issue #9263, opened 2024-04-03
Parallelize removing deps in `poetry intall --sync`


adriangb:
### Issue Kind

Change in current behaviour

### Description

Currently `poetry install --sync` removes dependencies one by one. It's also somewhat slow for each one (~1s, I'd think it should me ms). Uninstalling should be done in parallel and hopefully can be sped up.

### Impact

This enables a workflow where you cache all of the dependencies for a monorepo and then uninstall the ones you don't need in each services' Docker layer.

### Workarounds

None that I know of. Just don't use the pattern.

adriangb:
I see now that the source of this is https://github.com/python-poetry/poetry/issues/2658, but is seems like that was never fully investigated and a lot has changed in the internals of poetry since 2020 (e.g. not calling pip via subprocess, I think?). It seems prime candidate to re-evaluate.

dimbleby:
This was certainly not an accident.
https://github.com/python-poetry/poetry/blob/bd029ac0564b7822efbc903320e24ad021567a48/src/poetry/installation/executor.py#L190-L192

Uninstall is one of the few places where poetry does still call pip via a subprocess.  I expect a pull request to reimplement uninstallation more directly would be desirable - but more work than just "don't act serially"

adriangb:
Yeah I saw those lines, I was just hoping that "reimplement uninstallation more directly" had already been done so this would be easy. I guess not. Seems like that's what this issue is asking for then.

Secrus:
> @dimbleby if I wanted to start working on this, could you give me some pointers as to where to begin? I tried looking at how `install` does things but it's a lot of code. Some more general guidance around what to replace pip with would be helpful; I know at some point the internals of pip were broken out into libraries and I assume that's what's being used elsewhere but I don't know much more than that.

I have started some PoC work on this by taking out the uninstaller from `pip` (there is no library separate for that yet) and putting it [here](https://github.com/python-poetry/poetry/compare/main...Secrus:poetry:uninstaller). Didn't have much time to come back to it recently, but it *might* be a good start. Although I am still thinking that it could be a separate library one day.

dimbleby:
I guess conceptually the answer is: look at the package dist-info directory in the venv, and remove the items named in the `RECORD` file.  Though perhaps there are edge cases where things are more complicated than that eg packages installed from git maybe? not sure

[...]

PR #10091, opened 2025-01-21
feat(env): prefer embedded pip for poetry use

abn:
Presently, when Poetry dispatches pip commands for uninstallation, this uses the pip version installed in the active virtual environment. This can potentially cause issues like #10089.

This change, switches to preferring the embedded pip wheel and makes preferring the in-env version the exception, while preferring it when a user uses `poetry run pip`.

See-also: #4011

## Summary by Sourcery

Bug Fixes:
- Fixes potential issues caused by using the virtual environment's pip version when uninstalling packages via Poetry.

### DIFF
--- a/src/poetry/utils/env/base_env.py
+++ b/src/poetry/utils/env/base_env.py
@@ -269,11 +269,11 @@ class Env(ABC):
     @abstractmethod
     def get_marker_env(self) -> dict[str, Any]: ...
 
-    def get_pip_command(self, embedded: bool = False) -> list[str]:
-        if embedded or not Path(self._bin(self._pip_executable)).exists():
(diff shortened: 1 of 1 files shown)


dimbleby:
cf #6060, #6062

really what we want is to stop using pip altogether, iirc the remaining cases are:
- uninstall
- certain cases where the local project has a build step

uninstall is clearly the more important, there is discussion in #9263

radoering (changes requested):
I agree with dimbleby that using the embedded pip is known to cause issues and should only be done as fallback. Actually, we made sure that pip was updated before all other packages in the past. This got removed accidentally in #9392. I think we have to revert the part of #9392.

PR #10092, opened 2025-01-22
feat(cli): fallback to build-in poetry-core on build

finswimmer:
# Pull Request Check List

When we cannot find a `[build-system]` section or a `build-backend` within this section in a `pyproject.toml`  we assume `setuptools` as the build backend This is fine according to PEP 517.

While this approach is valid for dependencies it would be surprising on `poetry build`, now that an isolated build happens if the build-system is not `poetry-core`.

Before #10059 `poetry build` had created the package even if no build backend was defined in the `pyproject.toml`. Now it would fallback to `setuptools` which would results in a broken package.

This PR checks if a there is a `[build-system]` section and has a `build-backend` defined in the pyproject.toml and will use the build-in `poetry-core` if none is define. In addition a warning is presented to the user to fix it.

## Summary by Sourcery

Tests:
- Added a test to verify that the build command fails when no build system is defined in the pyproject.toml file.

## Summary by Sourcery

Tests:
- Added a test to verify that the build command fails when no build system is defined in the `pyproject.toml` file.


src/poetry/console/commands/build.py
```diff
@@ -114,13 +114,22 @@ def _get_builder(self) -> Callable[..., None]:
 
         return self._build
 
+    def _has_build_backend_defined(self) -> bool:
+        return "build-backend" in self.poetry.file.read().get("build-system", {})
```
abn, line 118: Would be good to move this to be something like `self.poetry.pytpoject.build_system.is_fallback`.

But that will need an additional property in https://github.com/python-poetry/poetry-core/blob/d97577e664d59173482a78f70e8f743cc5468851/src/poetry/core/pyproject/tables.py#L12

But for now something like this should work. 

```suggestion
        return "build-backend" in self.poetry.pyproject.data.get("build-system", {})
```

finswimmer:
@sourcery-ai review

src/poetry/console/commands/build.py
```diff
@@ -114,13 +114,22 @@ def _get_builder(self) -> Callable[..., None]:
...
+
     def build(self, options: BuildOptions) -> int:
         if not self.poetry.is_package_mode:
             self.io.write_error_line(
                 "Building a package is not possible in non-package mode."
             )
             return 1
 
+        if not self._has_build_backend_defined():
```
abn, line 127: Won't this also be a surprise to users as this breaks previous behavior in the same way setuptools are being used? Not sure how many people leave out the section intentionally. 

That said, I'd also say that the current use of setuptools is in accordance with PEP518. 

> Tools should not require the existence of the [build-system] table. A pyproject.toml file may be used to store configuration details other than build-related data and thus lack a [build-system] table legitimately. If the file exists but is lacking the [build-system] table then the default values as specified above should be used. If the table is specified but is missing required fields then the tool should consider it an error.

src/poetry/console/commands/build.py
finswimmer: Ah, I wasn't aware of the `data` attribute of `pyproject` :+1: 

I thought about a flag as well. But I don't wanted to change `poetry-core` before. And we would pollute again `poetry-core` with something we only need in `poetry`. Actually we only need it for `poetry build`.

src/poetry/console/commands/build.py
finswimmer: > Won't this also be a surprise to users as this breaks previous behavior in the same way setuptools are being used?

It will surprise people for sure. But now they get an error message. Without it `poetry build` would build a package, but it won't contain the data people expect. So "fail fast" is a better surprise ;)

dimbleby:
> we assume setuptools as the build backend This is fine according to PEP 517.

it's not only "fine" - it's what PEP517 says tools should do.

>  Now it would fallback to setuptools which would results in a broken package.

only results in a broken package if the pyproject.toml is not understood be setuptools, presumably?  In the new PEP621 world that is not going to be a problem.

finswimmer:
> In the new PEP621 world that is not going to be a problem.

In the PEP 621 world, yes. But if the project doesn't (fully) use the `project` section it is a problem.

As an alternative we could trigger a non-isolated build if no build-backend is given and )optional)  show a warning that it would be good idea to add one to the `pyproject.toml`. Than poetry would behave just like before #10059.

src/poetry/console/commands/build.py
abn: The problem is, how do you determine if they intended to use setuptools by relying on PEP518?

It is not unreasonable to imagine that Poetry is used for dev but the package needs setuptools for build.

abn:
Allowing the build to continue into isolated with a warning is safe enough, I reckon.

Let's not add more logic unless we need it. 

That said I can already see the issues titled "why is poetry using setuptools" etc.

radoering:
I am in favor of being backward compatible to the point of using `poetry-core` by default if no backend is defined. Printing a warning is fine.

It will be far more common that people have no backend defined and still used poetry build than that people that have no backend defined will be wondering that `poetry-core` is used instead of `setuptools`.

abn:
I am okay with defaulting to `poetry-core` too to maintain backwards compat.

Although I was looking forward to just being able to build arbitrary setuptool projects (which a lot of times do not have the `build-system` specified by simply calling `poetry build`.  But that is a niche use case.

finswimmer:
@sourcery-ai review

src/poetry/console/commands/build.py
```diff
@@ -91,6 +91,15 @@ def _requires_isolated_build(self) -> bool:
...
         :returns: True if an isolated build is required, False otherwise.
         """
+        if not self._has_build_backend_defined():
+            self.io.write_error_line(
+                "<warning>WARNING: No build backend defined. Please define one in the pyproject.toml.\n"
+                "Fallback to the build-in `poetry-core` version.\n"
+                "This warning will turn into an error in the future.\n"
+                "More details can be found at https://python-poetry.org/docs/main/libraries/#packaging</>"
+            )
+            return False
```
abn, line 101:
```suggestion
        if not self._has_build_backend_defined():
            self.io.write_error_line(
                "<warning><b>WARNING</>: No build backend defined. Please define one in the <c1>pyproject.toml</>.\n"
                "Falling back to using the built-in `poetry-core` version.\n"
                "This warning will turn into an error in the future.\n"
                "More details can be found at https://python-poetry.org/docs/libraries/#packaging</>"
            )
            return False
```

dimbleby:
I think this is a wrong direction, the specs are explicit about what the behaviour should be in case there is no backend defined.  

It is not "fall back to poetry-core", and it is not "report an error".  It is "use `setuptools.build_meta:__legacy__`".

Anyway you will want to have an answer ready for when folk turn up complaining that poetry is not behaving as it is supposed to.

dimbleby:
eg `poetry build` will now give different results than building with any other front end.  (Consider someone installing from a source distribution on PyPI.)

finswimmer:
> eg poetry build will now give different results than building with any other front end. (Consider someone installing from a source distribution on PyPI.)

This was always the case until now, because `poetry build` always used `poetry-core` to build the package, no matter if and what build-system was given in the `pyproject.toml`.

It's arguable to not throw in error in the future, if no build-system is given and instead use `setuptools`. But doing it now would be a breaking change and we should give people time to fix there packages and/or make the transition to PEP 621.

dimbleby:
> This was always the case until now

this is not correct, I think there are two cases and we are perhaps talking slightly past one another

In the case of a `pyproject.toml` with metadata defined in `tool.poetry`, the only build-backend that ever would have worked was `poetry-core`.  You are right that `poetry build` always used that backend regardless.  Other front-ends would almost certainly fail to produce anything useful if the backend was not specified,

In the case of a PEP-621 variant `pyproject.toml`, poetry is unambiguously going against the spec by falling back to `poetry-core`.  And in this case other front-ends _will_ use `setuptools`, and they might or might not get a wheel, and by not going with what the spec says we are introducing this confusion.

In the `tool.poetry` case, there is an argument for falling back to `poetry-core`: I agree anyway that this is preserving the old behaviour (though not necessarily that this is a good thing!)

In the PEP-621 case I really think that there is not a good argument for falling back to `poetry-core` and even that it is clearly the wrong thing to do.

radoering:
I think there is one (and probably the only) good argument: backwards compatibility. As of now (no matter if PEP 621 or not) all released versions do use `poetry-core` as build-backend, no matter what is defined.

In my opinion, it still makes sense to default to `poetry-core` with a warning that this is deprecated and you should define a build-backend and in future we will fall back to setuptools because that is what the spec says.

I do not think that it is that important to be compatible with projects that do not define a build-backend (and expect `setuptools`) - famous last words. If you manage a project with poetry you should define a build-backend. Maybe, `poetry check` should give an error/warning if `package-mode` is `true` and no build-backend is defined.

dimbleby:
> If you manage a project with poetry you should define a build-backend.

Sure - we can go further - if you manage a project with anything at all, you should define a build backend.  

So far as I know no-one thinks that omitting a build backend is the right thing to do, even the PEP517 stuff about how to handle that is (as I understand it) really aimed at projects that have not _yet_ added a build backend.

The question is whether we are justified in overruling the specs and choosing a different fallback than we "should".

Defaulting to poetry-core clearly made sense pre-PE517 and pre-PEP621.  If nothing else: poetry was here first, and it is hardly poetry's fault if the specs retrospectively asked it to do something else.

But anyway the poetry-core default made sense where there was no reasonable expectation of successfully building a poetry project except with poetry-core; and where there was definitely no hope of using poetry to build a project that used another backend.

But the world has moved on: 
- poetry has adopted the standard format, it is now totally reasonable to use poetry for project management while choosing a different backend eg for projects with build scripts
- on the main branch poetry is now build-system-agnostic
 
IMO those changes really both drive big holes through "default to poetry-core": it is time to get in line with the standards.

(I think I've said my piece - possibly too many times, sorry! - and will leave it there)

abn:
I agree with 

> In my opinion, it still makes sense to default to poetry-core with a warning that this is deprecated and you should define a build-backend and in future we will fall back to setuptools because that is what the spec says.

And that is my vote for now. We do a hard deprecation. 

1. Issue a warning during build in the next release and fallback **iff** no build-backend is defined.
2. Call attention to this in the blog post.
3. Next release conform to spec (because in the long run this is the right thing to do).

src/poetry/console/commands/build.py
```diff
@@ -114,6 +123,9 @@ def _get_builder(self) -> Callable[..., None]:
 
         return self._build
 
+    def _has_build_backend_defined(self) -> bool:
+        return "build-backend" in self.poetry.pyproject.data.get("build-system", {})
```
abn, line 127: Maybe move that inline as we will remove it right after the next release?

finswimmer: I think there is no need for it. We will either revert this commit completely or we can reuse it, to still emit a warning if falling back happens.

abn: /shrug okay

radoering (approved):

PR #10097, opened 2025-01-23
feat(python): use findpython

abn:
With this change Poetry will use [findpython](https://github.com/frostming/findpython) for Python version discovery. This helps remove Poetry's own custom "best effort" Python version discovery and hopefully also resolve a lot of issues related to the discovery of available Python versions in a given environment.

This change largely focuses on the Python Manager and not the Environment Manager. The latter could also be improved once this has landed. 

### Using pipx
```sh
pipx install --suffix=@10097 'poetry @ git+https://github.com/python-poetry/poetry.git@refs/pull/10097/head'
alias poetry=poetry@10097
poetry new foobar
cd foobar
poetry install
poetry env use 3.13
```

```sh
podman run --rm -i --entrypoint bash docker.io/python:latest <<EOF
set -xe 
python -m pip install --disable-pip-version-check -q git+https://github.com/python-poetry/poetry.git@refs/pull/10097/head
poetry new foobar
cd foobar
poetry env use 3.13
EOF
```

## Summary by Sourcery

Tests:
- Use mocked Python versions in tests.

## Summary by Sourcery

Tests:
- Use mocked Python versions in tests.

### DIFF
--- a/src/poetry/utils/env/python/manager.py
+++ b/src/poetry/utils/env/python/manager.py
@@ -1,7 +1,6 @@
 from __future__ import annotations
 
 import contextlib
-import shutil
 import sys
 
 from functools import cached_property
@@ -17,28 +16,15 @@ from cleo.io.null_io import NullIO
 from cleo.io.outputs.output import Verbosity
 from poetry.core.constraints.version import Version
 
-from poetry.utils.env.exceptions import NoCompatiblePythonVersionFoundError
+from poetry.utils.env.python.exceptions import NoCompatiblePythonVersionFoundError
+from poetry.utils.env.python.providers import ShutilWhichPythonProvider
 
 
 if TYPE_CHECKING:
-    from collections.abc import Iterable
-
     from cleo.io.io import IO
-    from typing_extensions import Self
 
     from poetry.config.config import Config
     from poetry.poetry import Poetry
-
-
-class ShutilWhichPythonProvider(findpython.BaseProvider):  # type: ignore[misc]
-    @classmethod
-    def create(cls) -> Self | None:
-        return cls()
-
-    def find_pythons(self) -> Iterable[findpython.PythonVersion]:
-        if path := shutil.which("python"):
-            return [findpython.PythonVersion(executable=Path(path))]
-        return []
 
 
 class Python:
--- a/src/poetry/utils/env/python_manager.py
+++ b/src/poetry/utils/env/python_manager.py
@@ -1,52 +1,85 @@
 from __future__ import annotations
 
+import contextlib
 import shutil
-import subprocess
 import sys
 
 from functools import cached_property
 from pathlib import Path
 from typing import TYPE_CHECKING
+from typing import cast
+from typing import overload
+
+import findpython
+import packaging.version
 
 from cleo.io.null_io import NullIO
 from cleo.io.outputs.output import Verbosity
 from poetry.core.constraints.version import Version
-from poetry.core.constraints.version import parse_constraint
 
-from poetry.utils._compat import decode
 from poetry.utils.env.exceptions import NoCompatiblePythonVersionFoundError
-from poetry.utils.env.script_strings import GET_PYTHON_VERSION_ONELINER
 
 
 if TYPE_CHECKING:
+    from collections.abc import Iterable
+
     from cleo.io.io import IO
+    from typing_extensions import Self
 
     from poetry.config.config import Config
     from poetry.poetry import Poetry
 
 
+class ShutilWhichPythonProvider(findpython.BaseProvider):  # type: ignore[misc]
+    @classmethod
+    def create(cls) -> Self | None:
+        return cls()
+
+    def find_pythons(self) -> Iterable[findpython.PythonVersion]:
+        if path := shutil.which("python"):
+            return [findpython.PythonVersion(executable=Path(path))]
+        return []
+
+
 class Python:
-    def __init__(self, executable: str | Path, version: Version | None = None) -> None:
-        self.executable = Path(executable)
-        self._version = version
+    @overload
+    def __init__(self, *, python: findpython.PythonVersion) -> None: ...
+
+    @overload
+    def __init__(
+        self, executable: str | Path, version: Version | None = None
+    ) -> None: ...
+
+    # we overload __init__ to ensure we do not break any downstream plugins
+    # that use the this
+    def __init__(
+        self,
+        executable: str | Path | None = None,
+        version: Version | None = None,
+        python: findpython.PythonVersion | None = None,
+    ) -> None:
+        if python and (executable or version):
+            raise ValueError(
+                "When python is provided, neither executable or version must be specified"
+            )
+
+        if python:
+            self._python = python
+        elif executable:
+            self._python = findpython.PythonVersion(
+                executable=Path(executable),
+                _version=packaging.version.Version(str(version)) if version else None,
+            )
+        else:
+            raise ValueError("Either python or executable must be provided")
+
+    @property
+    def executable(self) -> Path:
+        return cast(Path, self._python.interpreter)
 
     @property
     def version(self) -> Version:
-        if not self._version:
-            if self.executable == Path(sys.executable):
-                python_version = ".".join(str(v) for v in sys.version_info[:3])
-            else:
-                encoding = "locale" if sys.version_info >= (3, 10) else None
-                python_version = decode(
-                    subprocess.check_output(
-                        [str(self.executable), "-c", GET_PYTHON_VERSION_ONELINER],
-                        text=True,
-                        encoding=encoding,
-                    ).strip()
-                )
-            self._version = Version.parse(python_version)
-
-        return self._version
+        return Version.parse(str(self._python.version))
 
     @cached_property
     def patch_version(self) -> Version:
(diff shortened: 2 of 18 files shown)


tests/utils/test_python_manager.py
```diff
@@ -85,39 +94,51 @@ def test_get_preferred_python_use_poetry_python_disabled_fallback(
...
     os.environ["PATH"] = str(python_wrapper.parent) + os.pathsep + os.environ["PATH"]
 
-    active_python = Python._detect_active_python(NullIO())
+    python = Python.get_active_python()
+    assert python is not None
 
-    assert active_python == wrapped_python
+    # TODO: Asses if Poetry needs to discover real path in these cases as
+    # this is not a symlink and won't be handled by findpython
+    assert python.executable.as_posix() == python_wrapper.as_posix()
```
abn, line 128: @python-poetry/core as things stand findpython returns an object like this:

```
<PythonVersion executable=WindowsPath('C:/Users/.../python.bat'), version=<Version('3.12.2')>, architecture='64bit', implementation='CPython', major=3, minor=12, patch=2>
```

It correctly detects the version, but does not understand that this is a wrapper script. How do we want to handle this scenario?

I am guessing this will cause issues for folks that use a Python version installation manager that use `python.bat` to always have the same invocation while switching versions. But this also seems to be the case for cases where `python` is a symlink. 

For the latter we can use `findpython.PythonVersion.real_path` to reslove the path or simply `python.executable.resolve()`.

abn:
@sourcery-ai review

src/poetry/utils/env/python_manager.py
```diff
@@ -1,52 +1,88 @@
...
+    def create(cls) -> Self | None:
+        return cls()
+
+    def find_pythons(self) -> Iterable[findpython.PythonVersion]:
+        if path := shutil.which("python"):
+            return [findpython.PythonVersion(executable=Path(path))]
+        return []
+
+
 class Python:
```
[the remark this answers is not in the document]
abn: This would be the desired solution. Unfortunately, we need to maintain backwards compatibility. Hence we are using the overload.

src/poetry/utils/env/python_manager.py
```diff
@@ -1,52 +1,88 @@
...
+    ) -> None: ...
+
+    # we overload __init__ to ensure we do not break any downstream plugins
+    # that use the this
+    def __init__(
+        self,
+        executable: str | Path | None = None,
+        version: Version | None = None,
+        python: findpython.PythonVersion | None = None,
+    ) -> None:
```
finswimmer, line 59: I would like to keep `findpython` as an implementation detail of the `Python` class, therefor we should avoid `findpython` in the public "API" (like here in `__init__()`).

As far as I can see passing `python` on init is only done by several classmethods. Those method could pass the executable instead and the `PythonVersion` object is then (re)created. (Yes, recreating might be some overhead but it shouldn't be that much and for the sake of separating the implementation details I clearly prefer that). Not sure if the version should than also be passed through, because it looks like it's lazy loaded in `PythonVersion`.

src/poetry/utils/env/python_manager.py
```diff
@@ -1,52 +1,88 @@
...
-                encoding = "locale" if sys.version_info >= (3, 10) else None
-                python_version = decode(
-                    subprocess.check_output(
-                        [str(self.executable), "-c", GET_PYTHON_VERSION_ONELINER],
-                        text=True,
-                        encoding=encoding,
-                    ).strip()
-                )
-            self._version = Version.parse(python_version)
+    def python(self) -> findpython.PythonVersion:
```
finswimmer, line 76: Is this property used somewhere outside the tests? :thinking:

src/poetry/utils/env/python_manager.py
```diff
@@ -60,66 +96,62 @@ def patch_version(self) -> Version:
...
-        if not executable:
-            return None
+        if Path(python_name).exists():
+            try:
+                # if it is a path try assuming it is an executable
+                python = cls.from_executable(python_name)
+                assert python.version is not None
+                return cls.from_executable(python_name)
+            except (AssertionError, ValueError):
+                pass
```
finswimmer, line 137:
```suggestion
            with contextlib.suppress(AssertionError, ValueError):
                # if it is a path try assuming it is an executable
                python = cls.from_executable(python_name)
                assert python.version is not None
                return cls.from_executable(python_name)
```

src/poetry/utils/env/python_manager.py
```diff
@@ -129,39 +161,12 @@ def get_compatible_python(cls, poetry: Poetry, io: IO | None = None) -> Python:
         supported_python = poetry.package.python_constraint
         python = None
```
finswimmer, line 162: Seems to be unused now.

src/poetry/utils/env/python_manager.py
```diff
@@ -1,52 +1,88 @@
...
+    from collections.abc import Iterable
+
     from cleo.io.io import IO
+    from typing_extensions import Self
 
     from poetry.config.config import Config
     from poetry.poetry import Poetry
 
 
+class ShutilWhichPythonProvider(findpython.BaseProvider):  # type: ignore[misc]
```
finswimmer, line 32: Nice thing. How did you find out about this?

src/poetry/utils/env/python_manager.py
```diff
@@ -60,66 +96,62 @@ def patch_version(self) -> Version:
...
-        )
+    @classmethod
+    def get_active_python(cls) -> Python | None:
+        for python in ShutilWhichPythonProvider().find_pythons():
+            return cls(python=python)
 
-        executable = Python._full_python_path("python")
+        # fallback to findpython
+        if python := findpython.find():
+            return cls(python=python)
```
finswimmer, line 106: I don't think falling back to `findpython.find()` is correct here as it will return a "random" python. This method here is used by `get_preferred_python()` which should try to return either the active python in the current shell or, if this fails, the Python Poetry itself uses. With the fallback here it will only return Poetry's Python by accident.

src/poetry/utils/env/python_manager.py
abn: It will trigger unnecessary subprocess overheads and is not recommended. I don't think there is anything wrong with using it as an init var.

src/poetry/utils/env/python_manager.py
abn: By looking in the code heh.

src/poetry/utils/env/python_manager.py
abn: I guess the question would what we mean by active python. If want to say only return active if a binary named "python" is available, then maybe falling back to `findpython.find("python")` is more appropriate.

If we mean to say "any python available in the environment" then `find()` is probably fine.

In the first case, if no `python` binary is found, Poetry's python is returned.

In the second case, yes, we will return `python3.x`, for example, if its in the path otherwise nothing will be returned.

I am okay with it being `findpython.find("python")`.

finswimmer: I think there are different things here:

1. `get_active_python()` should only return whatever `python` resolves to in the current shell (that's essential to work with pyenv). If it cannot resolve it, it should return `None` and should not return any fallback.

2. It's the job of `get_preferred_python()`  to try to detect the active python (if not disabled in the config) and implement a fallback if there are any problems.

> we should only switch to poetry system python in the "worst" case scenario; which at this stage is "no python" was found anywhere. Since find() returns the highest version found we should be okay I reckon 🤔 

This would be a change to the behavior Poetry had since its beginnings. It was always either the current active one or the one Poetry itself uses. I see no reason why we should change it. Choosing the highest version found is not really a reliable/reproducible behavior for the end user.

abn: > If it cannot resolve it, it should return None and should not return any fallback.

Might be out of scope for this change, but I would like to know what this means for non *nix users. Because resolving binary named `python` might return nothing even if they have an active Python in the environment. Which would be a bug.

> It's the job of get_preferred_python() to try to detect the active python

In that case the naming is incorrect. We should just rename `get_active_python` to something like `get_default_python`.

> Choosing the highest version found is not really a reliable/reproducible behavior for the end user.

Unfortunately neither is picking the whatever `python` is found in the path because that usually is a symlink that can keep changing.

src/poetry/utils/env/python_manager.py
abn: Resolving this, I do not think I want to remove the ability to pass in an already instantiated `PythonVersion` instance.

abn:
@sourcery-ai review

src/poetry/utils/env/python/manager.py
```diff
@@ -0,0 +1,197 @@
...
+        )
+
+    @classmethod
+    def get_by_name(cls, python_name: str) -> Python | None:
+        if Path(python_name).exists():
+            with contextlib.suppress(AssertionError, ValueError):
+                # if it is a path try assuming it is an executable
+                python = cls.from_executable(python_name)
+                assert python.version is not None
+                return cls.from_executable(python_name)
```
finswimmer, line 137: Why we need the `assert`? And even if it is required, can't we return `python` instead of running `from_executabl` again? :thinking:

abn: We need the assert because sometimes given path might not be a valid python and we cannot be sure unless we execute something. Version allows this and ensures it's a Python we can get version from.

And yes we can rerun the instance. It just got missed in the refactor.

abn: Yeah, that assert was an artifact of the first iteration. Removed it now and changed it to suppressing just the ValueError. This should be sufficient for our needs.

finswimmer (approved):

Issue #10103, opened 2025-01-24
Clarify whether the python version needs to be specified in [tool.poetry.dependencies]


felixgwilliams:
### Issue Kind

Unclear documentation

### Existing Link

https://python-poetry.org/docs/pyproject/#dependencies-and-dependency-groups

### Description

According to the documentation on the `tool.poetry.dependencies` section:

> Be aware that declaring the python version for which your package is compatible is mandatory:

However, since v2, it is possible to use `project.requires-python` to specify the python version and poetry can accept a `tool.poetry.dependencies` section without a python version specified and can take the python version declaration from `requires-python`.

Can the documentation be updated to reflect the current behavior?

dimbleby:
Submit a pull request!

felixgwilliams:
Submitted in #10104

PR #10104, opened 2025-01-24
docs: add additional information on specifying required python versions

felixgwilliams:
The docs say that the python version must be declared in the `tool.poetry.dependencies` section

> Be aware that declaring the python version for which your package is compatible is mandatory:

This is no longer true, as the version can be declared in `project.requires-python` only.
I just removed the note saying it is mandatory.

# Pull Request Check List

Resolves: #10103

## Summary by Sourcery

Documentation:
- Remove the note stating that declaring the Python version in `tool.poetry.dependencies` is mandatory.

### DIFF
--- a/docs/pyproject.md
+++ b/docs/pyproject.md
@@ -708,6 +708,7 @@ python = "^3.7"
 
 {{% note %}}
 If you specify the compatible python versions in both `tool.poetry.dependencies` and in `project.requires-python`, then Poetry will use the information in `tool.poetry.dependencies` for locking, but the python versions must be a subset of those allowed by `project.requires-python`.
+
 For example, the following is invalid and will result in an error, because versions `4.0` and greater are allowed by `tool.poetry.dependencies`, but not by `project.requires-python`.
 
 ```toml
--- a/docs/pyproject.md
+++ b/docs/pyproject.md
@@ -699,6 +699,27 @@ If you have multiple repositories configured, you can explicitly tell poetry whe
 requests = { version = "^2.13.0", source = "private" }
 ```
 
+You may also specify your project's compatible python versions in this section, instead of or in addition to `project.requires-python`.
+
+```toml
+[tool.poetry.dependencies]
+python = "^3.7"
+```
+
+{{% note %}}
+If you specify the compatible python versions in both `tool.poetry.dependencies` and in `project.requires-python`, then Poetry will use the information in `tool.poetry.dependencies` for locking, but the python versions must be a subset of those allowed by `project.requires-python`.
+For example, the following is invalid and will result in an error, because version `3.14` is allowed by `tool.poetry.dependencies`, but not by `project.requires-python`.
+
+```toml
+[project]
+# ...
+requires-python = ">=3.8,<4.0"
+
+[tool.poetry.dependencies]
+python = ">=3.8"
+```
+{{% /note %}}
+
 You can organize your dependencies in [groups]({{< relref "managing-dependencies#dependency-groups" >}})
 to manage them in a more granular way.
 
--- a/docs/pyproject.md
+++ b/docs/pyproject.md
@@ -708,7 +708,7 @@ python = "^3.7"
 
 {{% note %}}
 If you specify the compatible python versions in both `tool.poetry.dependencies` and in `project.requires-python`, then Poetry will use the information in `tool.poetry.dependencies` for locking, but the python versions must be a subset of those allowed by `project.requires-python`.
-For example, the following is invalid and will result in an error, because version `3.14` is allowed by `tool.poetry.dependencies`, but not by `project.requires-python`.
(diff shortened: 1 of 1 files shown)


docs/pyproject.md
```diff
@@ -695,16 +695,6 @@ If you have multiple repositories configured, you can explicitly tell poetry whe
 requests = { version = "^2.13.0", source = "private" }
 ```
 
-{{% note %}}
```
abn, line 698: This is still required for projects that do not use the 621 project section.

abn: Maybe just modify to clarify that?

abn: That is unexpected then. I would imagine that since 2.0 supports <2.0 `pyproject.toml` it should in theory enforce that python is present in deps.

abn: @radoering is `(>=2.7,<2.8 || >=3.4)` defaults intended?

felixgwilliams: This is what my pyproject file looks like

```toml
[project]
name = "poetry-no-version"
version = "0.1.0"
description = ""
authors = [
    {name = "Felix Williams",email = "felixgwilliams@users.noreply.github.com"}
]
readme = "README.md"
dependencies = [
    "pytest (>=1)"
]


[build-system]
requires = ["poetry-core>=2.0.0,<3.0.0"]
```

I added pytest via `poetry add pytest>=1` with no problems. This is on poetry v2.0.1

abn: @felixgwilliams the case I am referring to is this:

```toml
[tool.poetry]
name = "poetry-no-version"
version = "0.1.0"
description = ""
authors = ["..."]
readme = "README.md"

[build-system]
requires = ["poetry-core"]
```

But just tested it seems to not care `python` is not specified anyway unless you do a `poetry check` using `<2.0.0`. So I guess this change is okay. /shrug

felixgwilliams: From what I can tell:

1. The python version requirement may be specified in `project.requires-python` or in `tool.poetry.dependencies`
2. If the python requirement is specified in both places
    1. Poetry will use the specification in `tool.poetry.dependencies` for locking
    2. The `tool.poetry.dependencies` specification must be at least as restrictive as the specification in `project.requires-python`
3. If the python version is not specified anywhere, the default of  `~2.7 || >=3.4` will be used

What do you think?

felixgwilliams:
I just added some more words about how the python version specifications interact. I think what I wrote is correct but let me know if it's not or if it's too long

docs/pyproject.md
radoering: > What do you think?

I think your observation is correct.

> It is not a good default

I agree. I think we should discourage not defining `project.requires-python` or `tool.poetry.dependency.python`, maybe at least print a warning in `poetry check`, but that is something for another PR.

docs/pyproject.md
```diff
@@ -695,14 +695,25 @@ If you have multiple repositories configured, you can explicitly tell poetry whe
...
 ```
+
+{{% note %}}
+If you specify the compatible python versions in both `tool.poetry.dependencies` and in `project.requires-python`, then Poetry will use the information in `tool.poetry.dependencies` for locking, but the python versions must be a subset of those allowed by `project.requires-python`.
+For example, the following is invalid and will result in an error, because version `3.14` is allowed by `tool.poetry.dependencies`, but not by `project.requires-python`.
+
+```toml
+[project]
+# ...
+requires-python = ">=3.8,<4.0"
```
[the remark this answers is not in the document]
felixgwilliams: Thanks for spotting! 🤦 I updated the text to match the example

docs/pyproject.md
```diff
@@ -695,14 +695,25 @@ If you have multiple repositories configured, you can explicitly tell poetry whe
...
+You may also specify your project's compatible python versions in this section, instead of or in addition to `project.requires-python`.
 
 ```toml
 [tool.poetry.dependencies]
 python = "^3.7"
 ```
+
+{{% note %}}
+If you specify the compatible python versions in both `tool.poetry.dependencies` and in `project.requires-python`, then Poetry will use the information in `tool.poetry.dependencies` for locking, but the python versions must be a subset of those allowed by `project.requires-python`.
+For example, the following is invalid and will result in an error, because versions `4.0` and greater are allowed by `tool.poetry.dependencies`, but not by `project.requires-python`.
```
abn, line 707:
```suggestion
If you specify the compatible python versions in both `tool.poetry.dependencies` and in `project.requires-python`, then Poetry will use the information in `tool.poetry.dependencies` for locking, but the python versions must be a subset of those allowed by `project.requires-python`.

For example, the following is invalid and will result in an error, because versions `4.0` and greater are allowed by `tool.poetry.dependencies`, but not by `project.requires-python`.
```

Minor nitpick.

radoering (approved):

Issue #7646, opened 2023-03-14
using `poetry add 'package[all]'` doesn't install extra dependencies (happens only with 'all')


aberges-grd:
- **Poetry version**: 1.4.0
- **Python version**: 3.8.10
- **OS version and name**: Ubuntu 22.04
- **pyproject.toml**: 

- [X] I am on the [latest](https://github.com/python-poetry/poetry/releases/latest) stable Poetry version, installed using a recommended method.
- [X] I have searched the [issues](https://github.com/python-poetry/poetry/issues) of this repo and believe that this is not a duplicate.
- [ ] I have consulted the [FAQ](https://python-poetry.org/docs/faq/) and [blog](https://python-poetry.org/blog/) for any relevant entries or release notes.
- [ ] If an exception occurs when executing a command, I executed it again in debug mode (`-vvv` option) and have included the output below.

## Issue

The issue is fairly simple. On a new poetry project, I used (I will report with the package used in case it helps):

```
poetry add 'polars[all]'
```

poetry installs polars but no extras. Meanwhile, on an isolated environment, using `pip install 'polars[all]'` does install every extra dependency listed by that package.

However, poetry works well if I specify which extras to install, e.g.:

```
poetry add 'polars[numpy,pandas,pyarrow,connectorx]'
```

works.

dimbleby:
previously rejected, for better or worse, at #3369

rtest2344:
Adding here in addition to #8496:

Documentation states that this should work [here](https://python-poetry.org/docs/cli#add).

clintonroy:
Poetry groups are (at the moment) a poetry specific concept that does not propagate to the published wheel. They're for things like development dependencies, cicd pipeline stuff, documentation tools etc.

Extras are an upstream python wheel thing, that get published to the wheel metadata.

color-science has groups, it does not have extras.

potentially here, poetry could/should warn that the requested extras don't exist.

dimbleby:
fixed by #10106

PR #10106, opened 2025-01-24
fix(puzzle): handle self-referential extras

abn:
The issue this is trying to fix can be reproduced as follows.

```
poetry new foobar
cd foobar
poetry add -E all pbs-installer
```

This should do the same thing as `poetry add -E download -E install pbs-installer`, but it does not at the moment.

### Using pipx
```sh
pipx install --suffix=@10106 'poetry @ git+https://github.com/python-poetry/poetry.git@refs/pull/10106/head'
alias poetry=poetry@10106
poetry new foobar
cd foobar
poetry add -E all pbs-installer
```

```sh
podman run --rm -i --entrypoint bash docker.io/python:latest <<EOF
set -xe
python -m pip install --disable-pip-version-check -q git+https://github.com/python-poetry/poetry.git@refs/pull/10106/head
poetry new foobar
pushd foobar
poetry add -E all pbs-installer
EOF
```

## Summary by Sourcery

Tests:
- Add tests for resolving self-referential extras dependency resolution.

## Summary by Sourcery

Fix the resolution of self-referential extras to consider nested extras.

Bug Fixes:
- Fixed an issue where self-referential extras were not resolved correctly when nested extras were involved.

Tests:
- Added tests for resolving self-referential extras dependency resolution.


tests/puzzle/test_solver.py
```diff
@@ -512,6 +513,76 @@ def test_solver_returns_extras_if_requested(
...
+        "A", {"version": "*", "extras": ["download", "install"]}
+    )
+    dep_nested = Factory.create_dependency("A", {"version": "*", "extras": ["all"]})
+
+    # Setting up "all" extra to include both "download" and "install"
+    package_a.extras = {
+        canonicalize_name("all"): [dep_all],
+        canonicalize_name("download"): [dep_download],
+        canonicalize_name("install"): [dep_install],
+        canonicalize_name("nested"): [dep_nested],
```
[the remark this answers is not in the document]
abn: This test case is not looking at circular dependency resolution but rather extras that refer to other extras in the same package.

src/poetry/puzzle/provider.py
```diff
@@ -469,7 +469,14 @@ def complete_package(
                 if extra not in package.extras:
                     continue
 
-                optional_dependencies += [d.name for d in package.extras[extra]]
+                for extra_dependency in package.extras[extra]:
+                    if extra_dependency.name == dependency.name:
+                        # If an extra/feature refers to another in the base package,
+                        # resolve the dependencies of that extra/feature.
+                        for feature in extra_dependency.features:
+                            for pd in package.extras.get(feature, []):
```
dimbleby, line 477: What if the self-reference goes a layer deeper still? I think this wants to be a depth- first search (or some other graph traversal)

src/poetry/puzzle/provider.py
```diff
@@ -469,7 +469,14 @@ def complete_package(
...
+                                _dependencies.append(pd)
```
dimbleby, line 478: I think this adds `pd` to `_dependencies` too soon.  The code a few lines below will check markers etc before copying from `requires` to `_dependencies`, this skips any such checks

dimbleby: I think there is likely also a problem later, in the block `# For normal dependency resolution, we have to make sure that root extras ...`

`dep.in_extras` only reflects the the "direct" extras that require `dep`.  But we should now also understand the top-level extras that might require `dep` through the indirection

dimbleby:
putting my review together I think this patch improves on what you currently have
```patch
index 1d081a3c..6793ba20 100644
@@ -458,19 +458,29 @@ class Provider:
             dependency = dependency_package.dependency
             requires = package.requires

-        optional_dependencies = []
+        found_extras = set()
+        optional_dependencies = set()
         _dependencies = []
-
-        # If some extras/features were required, we need to
-        # add a special dependency representing the base package
-        # to the current package
         if dependency.extras:
-            for extra in dependency.extras:
-                if extra not in package.extras:
+            # Find all the optional dependencies that are wanted - taking care to allow
+            # for self-referential extras.
+            stack = list(dependency.extras)
+            while stack:
+                extra = stack.pop()
+                if extra in found_extras:
                     continue
+                found_extras.add(extra)

-                optional_dependencies += [d.name for d in package.extras[extra]]
+                extra_dependencies = package.extras.get(extra, [])
+                for extra_dependency in extra_dependencies:
+                    if extra_dependency.name == dependency.name:
+                        stack += list(extra_dependency.features)
+                    else:
+                        optional_dependencies.add(extra_dependency.name)

+            # If some extras/features were required, we need to
+            # add a special dependency representing the base package
+            # to the current package
             dependency_package = dependency_package.with_features(
                 list(dependency.extras)
             )
@@ -505,10 +515,7 @@ class Provider:

             if not package.is_root() and (
                 (dep.is_optional() and dep.name not in optional_dependencies)
-                or (
-                    dep.in_extras
-                    and not set(dep.in_extras).intersection(dependency.extras)
-                )
+                or (dep.in_extras and not set(dep.in_extras).intersection(found_extras))                                                                                    ):
                 continue
```

but I think it still is problematic per https://github.com/python-poetry/poetry/pull/10106#discussion_r1929562873.

dimbleby:
also linking #7646

src/poetry/puzzle/provider.py
abn: Seems we do not encounter this case due to https://github.com/python-poetry/poetry/blob/603646c64601eb49886a97f6e930aba2cc520475/src/poetry/packages/locker.py#L577.

Which i suspect is a bug in itself. Seems [#6237](https://github.com/python-poetry/poetry/commit/5f2d2b866379a6f4f6779d382fde8aee9a2c8498#diff-b0671e034884534eca80d94078a8d744a370d79b9cab043257523440ba887f62R413) swallowed a [TODO](https://github.com/python-poetry/poetry/blob/9fb6c95d3784b79891747a2fbf05a1b95e1aa138/src/poetry/packages/locker.py#L413C45-L414) we had.

abn:
@sourcery-ai review

abn:
There is an intermittent failure on CI due to the way the fixtures are implemented, will need to fix that before merge.

dimbleby (approved):
I'm more or less reviewing my own suggestion here.  So: yes, looks good to me - but don't over-weight that opinion!

src/poetry/puzzle/provider.py
```diff
@@ -460,18 +460,29 @@ def complete_package(
...
+                extra = stack.pop()
+                if extra in found_extras:
                     continue
+                found_extras.add(extra)
 
-                optional_dependencies += [d.name for d in package.extras[extra]]
+                extra_dependencies = package.extras.get(extra, [])
+                for extra_dependency in extra_dependencies:
+                    if extra_dependency.name == dependency.name:
+                        stack += list(extra_dependency.features)
```
radoering, line 480:
```suggestion
                        stack += list(extra_dependency.extras)
```

Can we stick with `extras` here? (This should be the same.)

radoering (changes requested):
The implementation makes sense to me. I will take a look at the tests later this week. Then, I think we can merge it.

src/poetry/puzzle/provider.py
abn: Aren't features "activated extras" as opposed to "available extras"?

dimbleby: https://github.com/python-poetry/poetry-core/blob/afaa1285f160f0bad2ee9c3aa6133498ce3245af/src/poetry/core/packages/dependency.py#L197-L200

looks like a distinction without difference

src/poetry/puzzle/provider.py
abn: #10119

abn:
@dimbleby sorry about the multiple threads, figured I might as well bring the conversation here from https://github.com/python-poetry/poetry/pull/10124#issuecomment-2622318547 as I was able to resolve the test case issues.

We might be talking about two distinct cases. The scenario you have mentioned is fixed, I have also updated the [`test_solver_resolves_self_referential_extras`](https://github.com/python-poetry/poetry/pull/10106/files#diff-c3ec7c37bcb989d25a2fa5c7d1d5ad1496b9f70ad02982330ab6478c4738de7aR518-R579) case to include just that. Let me know if you were thinking something else there.

The case I was referring to is the [`test_solver_resolves_self_referential_extras_markers`](https://github.com/python-poetry/poetry/pull/10106/files#diff-c3ec7c37bcb989d25a2fa5c7d1d5ad1496b9f70ad02982330ab6478c4738de7aR582-R619) case. Where the self-referencing extra as a python version constraint. This however, I suspect can be left out of scope for this PR as it is not a regression but an existing issue.

Let me know if you think the scenario itself is invalid.

dimbleby:
> Let me know if you think the scenario itself is invalid.

While I do not see a good use case for it, it looks valid (and I had not thought of it!)

Probably not hard to fix?  factor out the three checks (on python constraint.allows_any(), UNSAFE_PACKAGES, and marker.validate()) that cause us not to transfer from `requires` to `_dependencies`.  Then make that same check before adding to the stack of extras.

(While we are talking about that code... UNSAFE_PACKAGES is completely dead and can safely be removed, and the calculation of marker_values could safely be done outside rather than inside the loop)

radoering (approved):
LGTM

PR #10110, opened 2025-01-25
tests: adapt tests to cosmetic changes caused by poetry-core#826

radoering:
see issue title

The `poetry-plugin-export` failures are caused by cosmetic changes. The tests are fixed in https://github.com/python-poetry/poetry-plugin-export/pull/321. We should probably merge this PR with failing downstream tests and switch from the released poetry-core version to the main branch afterwards. (I changed the CI to use the main branch of `poetry-plugin-export` instead the latest release.)

## Summary by Sourcery

Tests:
- Adapt tests to match the simplified dependency markers generated by poetry-core.

### DIFF
--- a/tests/puzzle/test_solver.py
+++ b/tests/puzzle/test_solver.py
@@ -1989,71 +1989,37 @@ def test_solver_duplicate_dependencies_with_overlapping_markers_complex(
         ],
     )
     opencv_requires = {dep.to_pep_508() for dep in ops[-1].package.requires}
-    expectation = (
-        {  # concise solution, but too expensive
-            (
-                "numpy (>=1.21.2) ;"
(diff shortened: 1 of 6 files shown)


tests/puzzle/test_solver.py
```diff
@@ -1989,71 +1989,37 @@ def test_solver_duplicate_dependencies_with_overlapping_markers_complex(
         ],
     )
     opencv_requires = {dep.to_pep_508() for dep in ops[-1].package.requires}
-    expectation = (
```
radoering, line 1992: Notably, I removed the distinction between "concise solution" and "current solution" because we have reached a solution which is even more concise than the "concise solution".

abn (approved):

abn (approved):

PR #10124, opened 2025-01-29
fix(puzzle): handle self-referential extras

abn:
This is an alternative proposed by @dimbleby for #10106. 

Raising this as a draft to preserve the proposed chang as I had to revert to my original approach to ensure tests passed once #10119 was merged to `main`.

## Summary by Sourcery

Tests:
- Add tests for resolving self-referential extras with and without markers, and at different dependency levels.


dimbleby:
I'm not sure what you're trying to do in the failing test, so far as I can tell the code is right and the test is wrong.

I think the sort of test you want is something like:
- one of the dependencies eg `download-package` has an additional marker eg `python_version < "3.10"`
- provide an environment to the solver during the test ie `with solver.use_environment(...)`
- confirm that when the marker is false in the environment, the extra is not installed

abn:
> I'm not sure what you're trying to do in the failing test, so far as I can tell the code is right and the test is wrong.

Your quite possibly correct. I only looked at it briefly, and the test setup is incorrect. That said, there is also the issue of the marker not being propagated or being accounted for if it is on an extra dependency; which I had hoped #10119 would fix - but I reckon that was incomplete as well.

abn:
Sorry about the multiple threads, decided to go back to #10106 to keep things together. https://github.com/python-poetry/poetry/pull/10106#issuecomment-2622616413

Issue #9908, opened 2024-12-18
Adding new source should set explicit flag by default


ggolda:
### Issue Kind

Change in current behaviour

### Description

According to the documentation ([Poetry Repositories Documentation](https://python-poetry.org/docs/repositories#primary-package-sources)) and current Poetry behavior, when a new source is added, it is automatically assigned a primary priority, causing PyPI to be ignored.

I believe this behavior is counterintuitive and should not be the default. In my opinion, it is error-prone and can lead to issues like the one my team encountered. We were using Poetry for the first time, and as a result, all of our dependencies were being loaded from an external mirror instead of PyPI.

I understand that it is our fault and such behavior is documented, but it is not what you would expect from a package manager. Usually at least in my experience with other languages and package managers (such as npm, composer, or even go get) by default they prioritize the main package repository and fallback to the supplementary if it is explicitly stated.

So in my opinion, a better alternative is to assign an `explicit` priority to added sources. Existing system is more than flexible enough to handle other requirements if such arise. 

### Impact

Improve user experience for users who do not read every section of documentation before starting to use the tool. Make it similar to average expectation of users coming from another languages. 

### Workarounds

This should be a default behavior in my opinion:

`poetry source add --priority=explicit https://path/to/source`

abn:
I tend to agree here, atleast in that primary shouldn't be the default. I suspect the assumption here was that internal projects typically have a private primary source that proxied PyPI and this was the larger use cases.

Would be good to hear from others as well before we change anything.

radoering:
At least, the current default is secure in the sense that you will not use PyPI accidentally when configuring other sources.

I do not have a strong opinion on changing it. If we want to change it, we should consider if `supplemental` would be a better default than `explicit`.

abn:
Using `supplemental` is what I was thinking.

The default right now is very much a "surprise". Even for myself, when working on the source command recently was a bit surprised PyPI disappeared when I added a source. Took me a few seconds to connect the dots. 

On thinking back, everytime I needed to add a new source has almost always been a supplemental source with PyPI as primary. 

So, my vote would be to change it and use supplemental unless someone has an objection to such a change.

PR #10134, opened 2025-01-31
chore(cli): deprecate default source priority

abn:
Relates-to: #9908

## Summary by Sourcery

Deprecate the default source priority for `poetry source add` from "primary" to "supplemental".  Emit a warning when using the default priority to inform users of the upcoming change.

Enhancements:
- Update the help text of the `poetry source add` command to reflect the deprecation of the default source priority and the future change to "supplemental".

Documentation:
- Document the `--priority` option for the `poetry source add` command and mention the possible values: primary, supplemental, and explicit.

### DIFF
--- a/src/poetry/console/commands/source/add.py
+++ b/src/poetry/console/commands/source/add.py
@@ -43,7 +43,8 @@ class SourceAddCommand(Command):
             "p",
             "Set the priority of this source. One of:"
             f" {', '.join(p.name.lower() for p in Priority)}. Defaults to"
-            f" {Priority.PRIMARY.name.lower()}.",
+            f" {Priority.PRIMARY.name.lower()}, but will switch to "
+            f"{Priority.SUPPLEMENTAL.name.lower()} in a later release.",
             flag=False,
         ),
     ]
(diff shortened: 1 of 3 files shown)


tests/console/commands/source/test_add.py
```diff
@@ -32,7 +32,12 @@ def assert_source_added(
 ) -> None:
     assert tester.io.fetch_error().strip() == ""
     assert (
-        tester.io.fetch_output().strip()
+        tester.io.fetch_output()
+        .strip()
+        .replace(
+            "The default priority will change to supplemental in a future release.\n",
+            "",
+        )
```
radoering, line 40: I would prefer testing that the warning is included.

src/poetry/console/commands/source/add.py
```diff
@@ -70,6 +71,10 @@ def handle(self) -> int:
             return 1
 
         if priority_str is None:
+            self.io.write_line(
```
radoering, line 74:
```suggestion
            self.io.write_error_line(
```

?

radoering (changes requested):

tests/console/commands/source/test_add.py
abn: 😅 I just took the quick way out. Will add a separate test case for it.

radoering (approved):

Issue #10148, opened 2025-02-05
Add ability to customize the Poetry project directory (where plugins are installed)


bcapitanu:
### Issue Kind

Brand new capability

### Description

This is a follow-up feature request from https://github.com/python-poetry/poetry/issues/10147#issuecomment-2637110495
Currently, Poetry uses a hardcoded folder `.poetry/` for storing project-related data in the current project root folder. 
For example `.poetry/plugins/` is used for storing any plugins installed by the plugin manager, as defined in the `[tool.poetry.requires-plugins]` section.

This feature request asks for the capability to customize the location of that `.poetry/` folder to help resolve issues related to multiple parallel invocations of `poetry` (via `tox`, in my case). 

### Impact

The impact is minimal, but important for users who run into the same kind of issue I did.

### Workarounds

The possible workarounds aren't ideal:
- don't use `tox` with parallel `poetry` invocations
- install plugins in the `poetry` environment, and don't rely on `requires-plugins` configuration
- use Poetry 1.x instead of 2.x

bcapitanu:
This is a super simple PR that won't affect anyone that doesn't use this "feature".  I was thinking to add a configuration option alongside the ones in `poetry config --list`, but I think it may be overkill for a feature that impacts a minuscule proportion of users (my estimate) that may need to use it (and makes the PR more complicated). 

Feel free to rename the env var name if what I chose isn't accurate.

PR #10149, opened 2025-02-05
Add dotpoetry folder override

bcapitanu:
# Pull Request Check List

Resolves: #10148

## Summary by Sourcery

Allow overriding the default ".poetry" project directory using the POETRY_PROJECT_DIR environment variable.

Enhancements:
- Add support for overriding the default project directory.

Documentation:
- Document the new POETRY_PROJECT_DIR environment variable and its usage.

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -159,9 +159,9 @@ You can override the Cache directory by setting the `POETRY_CACHE_DIR` environme
 
 ### Project Directory
 
-By default, the Poetry project directory is set to `.poetry/` in the project root. It is used by Poetry for storing project-specific 
-data. For example, the required plugins defined in `pyproject.toml` are installed by default in the `.poetry/plugins/` directory. 
-You can override the Poetry project directory by setting the `POETRY_PROJECT_DIR` environment variable. Note that it must be a relative 
+By default, the Poetry project directory is set to `.poetry/` in the project root. It is used by Poetry for storing project-specific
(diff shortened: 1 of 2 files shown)


dimbleby:
Is this really what you want?  

You are going to have to do something ugly with your tox cofiguration to pass a different environment variable each time, and then you are going to end up with a pile of identical directories containing copies of plugins.  It all feels like a bad solution?

Workarounds suggested elsewhere sound preferable to me.  Or can you just do something before running tox, to persuade poetry to install plugins? then none of your parallel invocations should need to repeat that and so they won't interfere with each other.

bcapitanu:
Passing a diff env var in tox is easy -- that's what I'd do:
```
[testenv]
description = run unit tests
setenv =
    POETRY_PROJECT_DIR=.tox.{envname}/.poetry/
allowlist_externals =
    mkdir
    poetry
commands =
    poetry sync --no-ansi --only=main,test
    poetry run ...
```

the fact that it would install multiple copies of plugins (we use only 1) isn't a concern either -- it's installed in ephemeral storage for that CI pipeline run and it's a small amount of data.

But overall I agree with your sentiment... this feels hacky and I'm not too happy with it either.
I just wanted to spend the minimal amount of time making changes that would "resolve" my issue.

Maybe it's worth for the Poetry team to think about a better solution to this.  Being able to run multiple `tox` / `poetry` commands in parallel sounds like a useful capability to have.

PR #10157, opened 2025-02-07
fix considered python constraint for overrides

radoering:
Previously, the Python constraint of the root package has always been used to check for incompatibilities, even if the Python constraint had effectively been restricted by overrides before. This could result in false conflicts à la "The current project's supported Python range is not compatible with some of the required packages Python requirement."

# Pull Request Check List

Resolves an issue introduced (or maybe just revealed) by #10141

## Summary by Sourcery

Fix an issue where the root package's Python constraint was always used for incompatibility checks, even when overrides had restricted it, leading to false conflict errors.

Bug Fixes:
- Fixed false "The current project's supported Python range is not compatible" errors.

Tests:
- Added tests to cover the fix for the Python constraint issue.

### DIFF
--- a/src/poetry/puzzle/provider.py
+++ b/src/poetry/puzzle/provider.py
@@ -125,7 +125,7 @@ class Provider:
         self._direct_origin = DirectOrigin(self._pool.artifact_cache)
         self._io = io
         self._env: Env | None = None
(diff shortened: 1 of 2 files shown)


radoering:
@sourcery-ai review

abn (approved):

PR #10164, opened 2025-02-10
Silently allow --no-update for backward compatibility

leonardgerardatomicmachinescom:
# Pull Request Check List

Resolves: #issue-number-here


radoering:
Not sure if this still makes sense after such a long time and so few complaints (resp. so few "thumbs up" on this PR). I think we should have printed a deprecation warning instead of just removing it in the first place but now it is probably just fine to leave it alone?

leonardgerardatomicmachinescom:
Consequence for us is that we stuck to poetry v1 and will probably move to some other toolchain rather than upgrade to v2.

Note that some other cli breakage led to the full decision, a major one being v2 `poetry -C <dir>` not behaving like all other linux tools (does not change the CWD).

radoering:
> Consequence for us is that we stuck to poetry v1 and will probably move to some other toolchain rather than upgrade to v2.

Does that mean that this PR would still help you if included in the next minor release or that it does not matter to you? In case, you are still interested in getting this merged, would it be fine for you if we print a deprecation warning if `--no-update` is used?

> a major one being v2 `poetry -C <dir>` not behaving like all other linux tools (does not change the CWD).

I think that one was fixed in #9831 with Poetry 2.0.0.

PR #10232, opened 2025-02-28
fix format and add links for contribute.md

pierceroberts:
# Pull Request Check List

Resolves: #issue-number-here

### DIFF
--- a/docs/contributing.md
+++ b/docs/contributing.md
@@ -28,23 +28,22 @@ related reports.
 
 #### Before submitting a bug report
 
-* **Check the [FAQ]** for a list of common questions and problems.
-* **Check the [blog]** for release notes from recent releases, including steps for upgrading and known issues.
-* **Check that your issue does not already exist** in the [issue tracker].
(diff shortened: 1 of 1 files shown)


pierceroberts:
So the `<ins>` make it more distinguishable with the amount of text and is a preference, I do think it makes it easier to find.

As for the links I was simply making them work and cleaning up the inconsistencies. There were some that went to links and some that just went to the bottom. And it made more sense to have one way.

radoering:
I think you are on the wrong track. The files in the docs folder are not optimized for being viewed via github but for generating our [documentation](https://python-poetry.org/docs/contributing/).

pierceroberts:
Oh interesting....
How does that process work @radoering ?
I thats the case, we can revert some of the changes. I think the links at the bottom can at least get updated.

radoering:
> How does that process work @radoering ?

The source code is here: https://github.com/python-poetry/website

See https://github.com/python-poetry/website?tab=readme-ov-file#local-development

Secrus (changes requested):

PR #10256, opened 2025-03-06
fix: :bug: handle pwsh correctly under linux

Katze719:
# Pull Request Check List

Resolves: #10253 

## Summary by Sourcery

Bug Fixes:
- Fixes an issue where the wrong activation script was being used for PowerShell (pwsh) under Linux.

### DIFF
--- a/src/poetry/console/commands/env/activate.py
+++ b/src/poetry/console/commands/env/activate.py
@@ -47,7 +47,11 @@ class EnvActivateCommand(EnvCommand):
         elif shell in ["csh", "tcsh"]:
             command, filename = "source", "activate.csh"
         elif shell in ["powershell", "pwsh"]:
-            command, filename = ".", "Activate.ps1"
+            if WINDOWS:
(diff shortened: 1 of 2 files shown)


src/poetry/console/commands/env/activate.py
```diff
@@ -47,7 +47,11 @@ def _get_activate_command(self, env: Env) -> str:
         elif shell in ["csh", "tcsh"]:
             command, filename = "source", "activate.csh"
         elif shell in ["powershell", "pwsh"]:
-            command, filename = ".", "Activate.ps1"
+            if WINDOWS:
+                command, filename = ".", "Activate.ps1"
+            else:
+                command, filename = ".", "activate.ps1"
+
```
radoering, line 54: I think we can just make it lower case. Actually, it seems it is also lower case on Windows (but it does not matter because Windows does not care about case).

radoering (approved):

PR #10276, opened 2025-03-15
provider: do not iterate over sets for more deterministic results

radoering:
Remove potential sources of non-determinism.

### DIFF
--- a/src/poetry/puzzle/provider.py
+++ b/src/poetry/puzzle/provider.py
@@ -366,7 +366,7 @@ class Provider:
 
         self.validate_package_for_dependency(dependency=dependency, package=package)
 
-        for extra in dependency.extras:
+        for extra in sorted(dependency.extras):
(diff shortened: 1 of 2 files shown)


src/poetry/puzzle/provider.py
```diff
@@ -497,16 +497,14 @@ def complete_package(
...
                     else:
                         optional_dependencies.add(extra_dependency.name)
 
             # If some extras/features were required, we need to add a special dependency
             # representing the base package to the current package.
 
-            dependency_package = dependency_package.with_features(
-                list(dependency.extras)
-            )
+            dependency_package = dependency_package.with_features(dependency.extras)
```
Secrus, line 507: I guess no need for sorted/list here since it doesn't change the outcome of the method?

Secrus (approved):
Just a small question, otherwise LGTM

Issue #10231, opened 2025-02-27
Poetry V2: Change ONLY_BINARY/NO_BINARY implementation so that you can ONLY_BINARY=:all: NO_BINARY=one_package


oc-will-holtam:
### Issue Kind

Change in current behaviour

### Description

Hi, when using pip, you can use only_binary and no_binary interchangeably as part of the same configuration.

For example, when running `pip install -r requirements --only-binary=:all: --no-binary=numpy` all packages in my requirements file will be installed using binaries, with the exception of numpy.

I want to be able to specify something similar to `installer.only-binary=:all: installer.no-binary=package1`.

### Impact

This is to mitigate against possible supply chain attacks, whereby a malicious developer has provided a malicious sdist for one of the packages we use, and released it as a new version of a package we use, but hasn't provided a binary, enabling arbitrary code execution to be run when the malicious package is built from source.

The reason for providing exceptions, is that there are some packages/dependencies we have to use, which never include binaries in their releases, but we want those to be the exceptions rather than the rule.

### Workarounds

The only way I can see of achieving this functionality currently, is to provide a really long list of packages (except the one you want to build from source) to the POETRY_INSTALLER_ONLY_BINARY environment variable. Which isn't practical, and is difficult to maintain.

dimbleby:
code is [here](https://github.com/python-poetry/poetry/blob/26da130da263f668a5674a40a30cb1465c605a55/src/poetry/installation/chooser.py#L63-L95), simplest way to make a proposal is likely by making a pull request

yokomotod:
+1.  I submitted a PR.

@dimbleby Thank you for pointing out the location in the code, it was very helpful

PR #10278, opened 2025-03-15
fix: installer.no-binary takes precedence over installer.only-binary

yokomotod:
# Pull Request Check List

Resolves: #10231

- [*] Added **tests** for changed code.
- [*] Updated **documentation** for changed code.

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -221,6 +221,8 @@ This configuration is ignored when `installer.parallel` is set to `false`.
 
 When set, this configuration allows users to disallow the use of binary distribution format for all, none or specific packages.
 
+If both `installer.no-binary` and `installer.only-binary` configurations match a package, `installer.no-binary` takes precedence.
+
 | Configuration          | Description                                                |
 |------------------------|------------------------------------------------------------|
 | `:all:` or `true`      | Disallow binary distributions for all packages.            |
@@ -268,6 +270,8 @@ across all your projects if incorrectly set.
 
 When set, this configuration allows users to enforce the use of binary distribution format for all, none or
 specific packages.
+
+If both `installer.no-binary` and `installer.only-binary` configurations match a package, `installer.no-binary` takes precedence.
 
 {{% note %}}
 Please refer to [`installer.no-binary`]({{< relref "configuration#installerno-binary" >}}) for information on allowed
--- a/src/poetry/installation/chooser.py
+++ b/src/poetry/installation/chooser.py
@@ -85,14 +85,18 @@ class Chooser:
                 continue
 
             if link.is_sdist and not self._only_binary_policy.allows(package.name):
-                logger.debug(
-                    "Skipping source distribution for %s as requested in only binary policy for"
-                    " package (%s)",
-                    link.filename,
-                    package.name,
-                )
-                sdists_skipped += 1
-                continue
+                if not self._no_binary_policy.allows(package.name):
+                    # no-binary policy takes precedence over only-binary policy
+                    pass
+                else:
+                    logger.debug(
+                        "Skipping source distribution for %s as requested in only binary policy for"
+                        " package (%s)",
+                        link.filename,
+                        package.name,
+                    )
+                    sdists_skipped += 1
+                    continue
 
             links.append(link)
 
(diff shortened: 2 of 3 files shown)


dimbleby:
And what if someone wants `NO_BINARY=:all: ONLY_BINARY=some_package`?  Why is the proposed order of precedence better?

src/poetry/installation/chooser.py
```diff
@@ -85,14 +85,18 @@ def choose_for(self, package: Package) -> Link:
...
             if link.is_sdist and not self._only_binary_policy.allows(package.name):
-                logger.debug(
-                    "Skipping source distribution for %s as requested in only binary policy for"
-                    " package (%s)",
-                    link.filename,
-                    package.name,
-                )
-                sdists_skipped += 1
-                continue
+                if not self._no_binary_policy.allows(package.name):
```
dimbleby, line 88: Don't write:
```python
if not condition:
    pass
else:
    do something
```

Prefer simply
```python
if condition:
    do something
```

yokomotod:
> And what if someone wants NO_BINARY=:all: ONLY_BINARY=some_package? Why is the proposed order of precedence better?

pip seems to have same precedence

```
$ pip install --dry-run --no-cache-dir --only-binary=requests --no-binary=:all: requests
Collecting requests
  Downloading requests-2.32.3.tar.gz (131 kB)

.......

Would install requests-2.32.3
```

dimbleby:
> pip seems to have same precedence

poetry is not pip, pip is not a standard.  Perhap if you reported that example to pip, they'd treat it as a bug.

The question is: why is the proposed precedence _better_?

yokomotod:
I see.

The reason for this precedence is to achieve `ONLY_BINARY=:all: NO_BINARY=one_package` as described in #10231. We can use `NO_BINARY` as an exclusion of `ONLY_BINARY=:all:`.  
I'm not sure of the motivation or use case of `NO_BINARY=:all: ONLY_BINARY=some_package`.  
So, I think this precedence is more useful and better.

On the other hand, I agree that it's a bit weird that `ONLY_BINARY=some_package NO_BINARY=some_package` results in `NO_BINARY=some_package`.

If we don't care about pip, I think another option is to "lower the priority of `:all:`". So, obviously conflicted options like `ONLY_BINARY=some_package NO_BINARY=some_package`, `ONLY_BINARY=:all: NO_BINARY=:all:` keep the current behavior.

What do you think?

dimbleby:
That sounds preferable to me, I would expect the treatment to be as near as possible symmetrical.

But I will say that this is not a feature I expect ever to use, so I am not personally especially invested either way

src/poetry/installation/chooser.py
yokomotod: fixed

yokomotod:
@dimbleby changed to lower `:all:` precedence

src/poetry/installation/chooser.py
```diff
@@ -61,7 +61,10 @@ def choose_for(self, package: Package) -> Link:
             links_seen += 1
 
             if link.is_wheel:
-                if not self._no_binary_policy.allows(package.name):
+                if not self._no_binary_policy.allows(package.name) and (
+                    not self._only_binary_policy.has_exact_package(package.name)
+                    or self._no_binary_policy.has_exact_package(package.name)
+                ):
```
dimbleby, line 67: This check should be redundant, we already did `_no_binary_policy_allows()`.  (And similarly later, on the binary policy)

yokomotod: @dimbleby 

`self._no_binary_policy.has_exact_package()` is required.  
`not self._no_binary_policy.allows()` means it includes exact package name or `:all:`. We have to distinguish it with `has_exact_package()`.

I agree that it isn't readable and fixed readability: https://github.com/python-poetry/poetry/pull/10278/commits/82792e646f6a3bc013f517e38c3d8e870015c3d3

docs/configuration.md
```diff
@@ -269,6 +271,8 @@ across all your projects if incorrectly set.
 When set, this configuration allows users to enforce the use of binary distribution format for all, none or
 specific packages.
 
+If both `installer.no-binary` and `installer.only-binary` are set, explicit package names will take precedence over `:all:`.
+
```
radoering, line 275: This should be redundant because of the note below.

radoering:
```suggestion
```

radoering (approved):

PR #10340, opened 2025-04-16
executor: propagate original error even if pkg.to_dependency() fails

lotheac:
this fixes the following regression, introduced somewhere between 1.8.4 and 2.0.1:

a git dependency which fails wheel build does not output the underlying error that caused the wheel build failure. eg.

	podman run -ti -w /tmp --entrypoint /bin/sh --rm python:3.13.3-alpine3.21 -c "pip install poetry && poetry new . && poetry add 'python-kadmin @ git+https://github.com/authentik-community/python-kadmin.git@v0.2.0'"

the root cause is that the exception handling in _execute_operation() calls pkg.to_dependency(), which fails like:

	ValueError: Invalid git url "/root/.cache/pypoetry/virtualenvs/tmp-6WcazSRI-py3.13/src/python-kadmin"

and the underlying build error is therefore never output.

# Pull Request Check List


radoering (changes requested):
At first, thanks for providing an exact repro of the issue. 👍 

Although the change clearly improves the behavior, I think it is more a workaround than a valid fix. The suppressed exception is the real bug. When testing the repro with the change, you can see the following line at the end of the output:

```
You can verify this by running pip wheel --no-cache-dir --use-pep517 "/root/.cache/pypoetry/virtualenvs/tmp-6WcazSRI-py3.13/src/python-kadmin".
```

which is wrong because it references the cache directory instead of the git URL.

The reason is probably the following hack:

https://github.com/python-poetry/poetry/blob/297ae7d6ab98f9cadca27d1e4f702aa25e475bd6/src/poetry/installation/executor.py#L674

We overwrite the source URL and reset it later:
https://github.com/python-poetry/poetry/blob/297ae7d6ab98f9cadca27d1e4f702aa25e475bd6/src/poetry/installation/executor.py#L685-L686

Maybe, it is enough if we put some lines in a try-except and always reset the source URL in case of an exception.

Further, we should probably add a test. Maybe, the following test can be extended (or at least used as a starting point): https://github.com/python-poetry/poetry/blob/297ae7d6ab98f9cadca27d1e4f702aa25e475bd6/tests/installation/test_executor.py#L1370

The test will probably be more complicated than the fix. I may take a look for myself later.

lotheac:
@radoering thank you for your review.

I'm not super familiar with the poetry code, so I would appreciate it if you could take a look at a better fix. If that's ok with you, I think we can close this PR.

PR #10361, opened 2025-04-25
fix `get_supported_tags` in cases where the target env has a different platform

radoering:
# Pull Request Check List

 #9304 introduced a regression where wheels for the wrong platform were installed in rare cases. This includes:
- using a 32 Bit Python on Windows (and having Poetry installed with a normal 64 Bit Python)
- using an emulated aarch64 Python on an x86_64 Linux (and having Poetry installed "normally")

This fix basically falls back to determine the platform tags inside the target venv if `sysconfig.get_platform()` of the target venv is different from the value from Poetry's own venv.

## Summary by Sourcery

Ensure correct platform tags are detected for target environments with different platforms or architectures than the host.

Bug Fixes:
- Fix incorrect platform tag detection when the target environment's platform differs from Poetry's host environment (e.g., 32-bit vs 64-bit Windows, x86_64 vs aarch64 Linux), preventing incorrect wheel installations by querying the target environment directly for its platform information when necessary.
- Include the target environment's `sysconfig.get_platform()` value in its markers for comparison.

Tests:
- Add tests to verify correct tag generation in cross-platform and cross-architecture scenarios.

### DIFF
--- a/src/poetry/utils/env/virtual_env.py
+++ b/src/poetry/utils/env/virtual_env.py
@@ -3,6 +3,7 @@ from __future__ import annotations
 import json
 import os
 import re
+import sysconfig
 
 from contextlib import contextmanager
 from copy import deepcopy
@@ -15,6 +16,7 @@ from poetry.utils.env.base_env import Env
 from poetry.utils.env.script_strings import GET_BASE_PREFIX
 from poetry.utils.env.script_strings import GET_ENVIRONMENT_INFO
 from poetry.utils.env.script_strings import GET_PATHS
+from poetry.utils.env.script_strings import GET_PLATFORMS
 from poetry.utils.env.script_strings import GET_SYS_PATH
 
 
@@ -54,6 +56,7 @@ class VirtualEnv(Env):
         python = self.version_info[:3]
         interpreter_name = self.marker_env["interpreter_name"]
         interpreter_version = self.marker_env["interpreter_version"]
+        sysconfig_platform = self.marker_env["sysconfig_platform"]
 
         if interpreter_name == "pp":
             interpreter = "pp3"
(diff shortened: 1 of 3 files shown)


tests/utils/env/test_env.py
```diff
@@ -92,16 +93,24 @@ def test_env_commands_with_spaces_in_their_arg_work_as_expected(
...
+
+    # determine expected tags before patching sysconfig!
+    expected_tags = list(packaging.tags.sys_tags())
+
+    if differing_platform:
+        mocker.patch("sysconfig.get_platform", return_value="some_other_platform")
 
-    assert venv.get_supported_tags() == list(packaging.tags.sys_tags())
+    assert venv.get_supported_tags() == expected_tags
+    assert run_python_script_spy.call_count == (2 if differing_platform else 1)
```
Secrus, line 113: Instead of inline condition, can you add an extra variable and set it in the `if` in line 109, so that we have a single conditioning place?

src/poetry/utils/env/virtual_env.py
```diff
@@ -62,9 +65,29 @@ def get_supported_tags(self) -> list[Tag]:
...
+        #  This one is also different for a 32 Bit and a 64 Bit Python on Windows,
+        #  but it is the same for x86_64 Linux and aarch64 Linux!
+        if sysconfig_platform != sysconfig.get_platform():
+            # Relevant for the following use cases, for example:
+            # - using a 32 Bit Python on a 64 Bit Windows
+            # - using an emulated aarch Python on an x86_64 Linux
+            output = self.run_python_script(GET_PLATFORMS)
+            platforms = json.loads(output)
+        else:
+            platforms = None
```
Secrus, line 82:
```suggestion
        platforms = None
        if sysconfig_platform != sysconfig.get_platform():
            # Relevant for the following use cases, for example:
            # - using a 32 Bit Python on a 64 Bit Windows
            # - using an emulated aarch Python on an x86_64 Linux
            output = self.run_python_script(GET_PLATFORMS)
            platforms = json.loads(output)

```

tests/utils/env/test_env.py
radoering: done

Secrus (approved):

PR #10374, opened 2025-05-04
chore: update dependencies (especially poetry-core) ...

radoering:
...  and set upper bound for importlib-metadata, which is only required for Python 3.9

## Summary by Sourcery

Chores:
- Upgrade poetry-core from version 2.1.2 to 2.1.3

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -68,14 +68,14 @@ virtualenv = ["virtualenv (>=20.0.35)"]
 
 [[package]]
 name = "cachecontrol"
-version = "0.14.2"
+version = "0.14.3"
 description = "httplib2 caching for requests"
 optional = false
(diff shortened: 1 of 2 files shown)


Secrus (approved):

somyarj1637:
Hi @radoering 
Can you explain why and where does the changes breaks because of which you need to set a upper-bound for `importlib_metadata` for Python3.9 ?

radoering:
@somyarj1637 Just take a look at the changes. There is a comment:

```
    # <8.7 because .metadata() (and Distribution.metadata) can now return None,
    # which requires some adaptions to our code.
```

Further, it is only Python 3.9 because we do not need `importlib-metadata` for later Python versions.

Issue #10396, opened 2025-05-19
test_no_additional_output_in_verbose_mode fails on Linux


hrnciar:
### Description

Hello,

poetry 2.1.3 doesn't build on Fedora Rawhide. The newly added `test_no_additional_output_in_verbose_mode` fails with:

````
=================================== FAILURES ===================================
_________________ test_no_additional_output_in_verbose_mode[] __________________
[gw0] linux -- Python 3.13.3 /usr/bin/python3

tmp_venv = VirtualEnv("/tmp/pytest-of-mockbuild/pytest-0/popen-gw0/test_no_additional_output_in_v0/venv")
mocker = <pytest_mock.plugin.MockerFixture object at 0x7fd5f4bd5d90>
app_tester = <cleo.testers.application_tester.ApplicationTester object at 0x7fd5f58eb2f0>
verbosity = ''

    @pytest.mark.parametrize("verbosity", ["", "-v", "-vv", "-vvv"])
    def test_no_additional_output_in_verbose_mode(
        tmp_venv: VirtualEnv,
        mocker: MockerFixture,
        app_tester: ApplicationTester,
        verbosity: str,
    ) -> None:
        mocker.patch("shellingham.detect_shell", return_value=("pwsh", None))
        mocker.patch("poetry.utils.env.EnvManager.get", return_value=MockEnv(is_venv=True))
    
        # use an AppTester instead of a CommandTester to catch additional output
        app_tester.execute(f"env activate {verbosity}")
    
        lines = app_tester.io.fetch_output().splitlines()
>       assert len(lines) == 1
E       assert 0 == 1
E        +  where 0 = len([])

/builddir/build/BUILD/poetry-2.1.3-build/poetry-2.1.3/tests/console/commands/env/test_activate.py:95: AssertionError
------------------------------ Captured log setup ------------------------------
INFO     virtualenv.discovery.builtin:builtin.py:76 find interpreter for spec PythonSpec(path=/usr/bin/python3)
INFO     virtualenv.discovery.builtin:builtin.py:83 proposed PythonInfo(spec=CPython3.13.3.final.0-64, exe=/usr/bin/python3, platform=linux, version='3.13.3 (main, May  6 2025, 00:00:00) [GCC 15.1.1 20250425 (Red Hat 15.1.1-1)]', encoding_fs_io=utf-8-utf-8)
[...]

PR #10397, opened 2025-05-19
Run test_no_additional_output_in_verbose_mode only on Windows

hrnciar:
# Pull Request Check List

Resolves: #10396


tests/console/commands/env/test_activate.py
```diff
@@ -79,6 +79,7 @@ def test_env_activate_prints_correct_script_on_windows(
 
 
 @pytest.mark.parametrize("verbosity", ["", "-v", "-vv", "-vvv"])
+@pytest.mark.skipif(not WINDOWS, reason="Only Windows shells")
```
radoering, line 82: I suppose there is no shell activator script that is available on all Linux systems? I assume that the test fails because `_get_activate_command` checks if an `activate.ps1` file exists, which does not in your case? (Interesting enough it seems to exist on Ubuntu - and it also works on macOS in our CI.)

I would prefer not restricting the test to Windows. Maybe, we can just patch `_get_activate_command` to just return an arbitrary string?

radoering (changes requested):

tests/console/commands/env/test_activate.py
hrnciar: So the files are present even on Fedora:
~~~~
(Pdb) os.listdir("/tmp/pytest-of-root/pytest-26/test_no_additional_output_in_v0/venv/bin")
['python', 'python3', 'python3.13', 'activate', 'activate.csh', 'activate.fish', 'activate.nu', 'activate.ps1', 'activate_this.py']
~~~~
I think the test is missing `mocker.patch("poetry.utils.env.EnvManager.get", return_value=tmp_venv)`, with this line the test passes.
Although I am not sure how come it works on Ubuntu a MacOS.

radoering: That might also be a solution. I think it works because `MockEnv` works for our test setup, but obviously not for your setup. Maybe, it is not a matter of OS, but you do something different than our CI (or other contributors). On my system, it looks into the directory of Poetry's own venv, which also has these files.

Nevertheless, if `tmp_venv` works for you and our CI that is fine. It is a bit more expensive than `MockEnv`, but also more realistic.

radoering:
pre-commit.ci fix

Issue #8864, opened 2024-01-10
`poetry init --author` doesn't seem to allow multiple authors on Windows


DVFrankieHomewood:
- [x] I have searched the [issues](https://github.com/python-poetry/poetry/issues) of this repo and believe that this is not a duplicate.
- [x] I have searched the [FAQ](https://python-poetry.org/docs/faq/) and general [documentation](https://python-poetry.org/docs/) and believe that my question is not already covered.

## Feature Request

Hi all,

Been creating a couple scripts to spin up development environments automatically using poetry in the terminal.

Noticed that there is no `--authors` flag for `poetry init` and `poetry init --author` doesn't seem to accept lists no matter how I seem to input it, figured it'd be a pretty quick implementation to allow some sort of delimiter to input more than one author at a time or add any existing solutions to this issue into the docs.

All the best,
Frankie

kaminskj:
I'd like to Claim this issue.

After looking over _poetry/src/poetry/console/commands/init.py_, It seems this issue hasn't been fixed yet. I don't have much experience contributing to OSS but I will do my best! 

The _pyproject.toml_ file already has _authors_ defined as a list, So it seems only the _init.py_ file needs to be updated. Now it seems there are a couple options to make this change.

1. add a --coauthor or --authors Option as _DVFrankieHomewood_ mentioned, that allows adding/appending additional authors
2. add a prompt/loop to the --author option, that asks if co-authors need to be added
3. update --author to accept lists

Currently I'm planning on option 2, and adding a loop to the --author command, that prompts the user for more authors. This doesnt require any updates to the commands list in documentation, and it also seems more user friendly then copying lists into the console. @python-poetry/triage Please let me know if there is a preferred method for how information is prompted by the user, otherwise I will continue with option 2.

Secrus:
@kaminskj I think the best way would be to go with option 3. Option 2 would be tricky to handle in non-interactive settings. The way I see it, CLI should allow for something like `poetry init --author author_1 --author author_2`.

dimbleby:
Something about this issue is astonishingly attractive to everyone and their bots.  

Humans and AI alike: check for open pull requests already addressing this before opening another one.

PR #10402, opened 2025-05-20
poetry init --author doesn't seem to allow multiple authors on Windows #8864

kaminskj:
# Pull Request Check List

Resolves: #8864

Changed Authors option in init.py to multiple=True. This allows the user to input –author name1 –author name2, to the nth author name. This does change the output of this option from a string to a list, which impacts downstream code.

For init.py changed line 151, author, Is now authors as this variable corresponds to a potential list of authors instead of a string. Since we are now handling a list, line 157 has to change the vcs.config author created into a list, so downstream code will work with it. 

Updated the is_interactive string, so multiple authors can be displayed (lines 164 to 168). Lines 166 to 169 handle user inputs to the is_interactive. If the user inputs nothing, then displayed authors should stay the same. If user enters “n” then author reverts to defaults. This is defined in the tests, removing this will cause errors where authors in the final file won't be set to defaults. Lines 172 and 173 handle the original line 165. This may be a different issue, not sure if user entering "n" should mean no to current input or no to entering a different author.

The layout function  (248 init.py) now takes an entire list of authors, before it only took index 0. This required changes to layout, as it wasn’t coded to handle multiple author inputs. 

Updated layout.py to take list inputs, line64, previously it only took str inputs. Added a handle for if user enters multiple or just 1 author as an --author input (lines 88 to 92 layout.py).

Now when creating the layout, a for loop is needed to iterate over the list of multiple authors the only change is wrapping the existing code handling authors in a for loop iterating over each author in the list (lines 148 to 157 layout.py)

In test_init.py added a new test to check for multiple author inputs. Simply copied test_predefined to create a 2 author input. The only change is line 911 and 912, and then changes to the expected output for the check (lines 926 and 927). The remaining code is copied from test_predefined.

Poetry run pytest successfully completes with only a threading error which I think is due to issues with my local environment setup ->   tests/utils/test_threading.py::test_threading_atomic_cached_property_different_instances[value_functools_cache-24] - AssertionError: assert 11 == 24

------ Only Added 1 new test, that adds 2 authors, existing test check for 1 author, did not test 3 or more authors
- [No] Updated **documentation** for changed code. 
------Code doesn't require a documentation change. Currently --author command is documented, and it seems the method of adding multiple commands (e.g --author name1 --author name2... --author nameX, is standard for these types of commands) Documentation could be added for extra clarification but doesn't seem required currently.

### DIFF
--- a/src/poetry/console/commands/init.py
+++ b/src/poetry/console/commands/init.py
@@ -163,13 +163,13 @@ The <c1>init</c1> command creates a basic <comment>pyproject.toml</> file in the
             )
             question.set_validator(lambda v: self._validate_author(v, authors))
             author = self.ask(question)
-            if author == author_str:
+            if author == author_str: # user entered nothing, dont change authors
                 author = ""
-            if author is None:
-                authors = author
+            if author is None: # none is returned if user enters "n",
+                authors = author # author gets overwritted by default settings defined in question.set
 
 
-        authors = [author] if author else authors
+        authors = [author] if author else authors # checks if author = None or ""
         authors = authors if authors else []
 
         license_name = self.option("license")
--- a/src/poetry/console/commands/init.py
+++ b/src/poetry/console/commands/init.py
@@ -236,7 +236,6 @@ The <c1>init</c1> command creates a basic <comment>pyproject.toml</> file in the
 
             self.line("")
 
-
         layout_ = layout(layout_name)(
             name,
             version,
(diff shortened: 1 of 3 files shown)


src/poetry/console/commands/init.py
```diff
@@ -148,21 +150,28 @@ def _init_pyproject(
...
             question = self.create_question(
-                f"Author [<comment>{author}</comment>, n to skip]: ", default=author
+                f"Author [<comment>{author_str}</comment>, n to skip]: ",
+                default=author_str,
             )
-            question.set_validator(lambda v: self._validate_author(v, author))
+            question.set_validator(lambda v: self._validate_author(v, authors))
             author = self.ask(question)
+            if author == author_str:  # user entered nothing, dont change authors
+                author = ""
```
radoering, line 170: It looks like `author` is not used after this line. Why do we set it here?

ammargit93: Why not introduce a new `--authors` flag for multiple authors?

PR #10405, opened 2025-05-20
fix: `ensure_project_plugins` is not called when `--no-plugins` flag is true

dfavato:
# Pull Request Check List

This PR fixes when the user wants to run `poetry install` but does not want to install the plugins.

Many poetry plugins ([shell](https://github.com/python-poetry/poetry-plugin-shell), [export](https://github.com/python-poetry/poetry-plugin-export), ...) are usually intended for development environments only. Ensuring plugin installation whenever `poetry install` is run means that all environments need to be compatible, not only with project dependencies, but also with plugins (which might not always be the case).

For instance, I'm trying to install a project in Databricks, but I can't because the plugin is incompatible.
image

However, if I manually edit the `pyproject.toml` and remove the plugins, I can install it.
image

Currently, the `--no-plugins` [documentation](https://python-poetry.org/docs/cli/#global-options) is minimal; however, in IMHO, `--no-plugins` should prevent the installation.

## Summary by Sourcery

Modify the installer command to skip plugin setup when the --no-plugins flag is used

Bug Fixes:
- Respect the --no-plugins flag by not calling PluginManager.ensure_project_plugins during install

Tests:
- Add parameterized test to verify ensure_project_plugins is called only when plugins are enabled

### DIFF
--- a/src/poetry/console/commands/install.py
+++ b/src/poetry/console/commands/install.py
@@ -111,7 +111,8 @@ you can set the "package-mode" to false in your pyproject.toml file.
 
         from poetry.masonry.builders.editable import EditableBuilder
 
-        PluginManager.ensure_project_plugins(self.poetry, self.io)
+        if not self.option("no-plugins"):
+            PluginManager.ensure_project_plugins(self.poetry, self.io)
 
         if self.option("extras") and self.option("all-extras"):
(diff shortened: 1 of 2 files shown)


radoering (approved):
makes sense

Issue #10429, opened 2025-06-13
Unexpected Behavior: `poetry new .`  Acts Like `poetry init` Without Creating Project Structure


SoSaymon:
### Description

# Unexpected Behavior: `poetry new .` Acts Like `poetry init` Without Creating Project Structure

## Description

Running `poetry new .` inside a newly created and empty directory behaves like `poetry init`, rather than creating the expected full project structure. This behavior is undocumented and may be confusing to users expecting the typical `src/` and `tests/` layout.

## Steps to Reproduce

```powershell
> mkdir test-poetry
> cd test-poetry
> poetry new .
> ls
```

### Actual Output

```powershell
Mode                 LastWriteTime         Length Name
----                 -------------         ------ ----
-a----         6/13/2025   7:24 PM            391 pyproject.toml
```

### Expected Output

```powershell
Mode                 LastWriteTime         Length Name
----                 -------------         ------ ----
d-----         6/13/2025   7:53 PM                src
d-----         6/13/2025   7:53 PM                tests
-a----         6/13/2025   7:53 PM            379 pyproject.toml
-a----         6/13/2025   7:53 PM              0 README.md
```

Or alternatively, a clear error indicating that creating a project in `.` is not supported.

## Discussion

This behavior appears to be an undocumented overlap between `poetry new` and `poetry init`. The resulting project is incomplete (no `src/`, `tests/`, or `README.md`). Additionally, `poetry init` does not support the `--src` flag, creating a functional gap.

## Suggested Solutions

1. **Document** the current behavior explicitly in the CLI help and documentation.
2. **Raise an error** when `poetry new .` is used, similar to how it errors on non-empty directories.
3. **Update behavior** so that `poetry new .` generates a full project scaffold in the current directory.

## Environment

* **OS**: Windows 10 Pro
* **Poetry version**: 2.1.3

If this behavior is intentional, it would be helpful to document it explicitly under `poetry new` usage

### Workarounds

To mitigate this error today you need to use `poetry new <project_name>` in the parent directory. 

### Poetry Installation Method

install.python-poetry.org

### Operating System

Windows 10 Pro

### Poetry Version

Poetry (version 2.1.3)

### Poetry Configuration

```bash session
cache-dir = "C:\\Users\\Szymon BUAS\\AppData\\Local\\pypoetry\\Cache"
data-dir = "C:\\Users\\Szymon BUAS\\AppData\\Roaming\\pypoetry"
installer.max-workers = null
[...]

PR #10431, opened 2025-06-14
fix: unexpected behavior of new command in existing empty directories

SoSaymon:
# Pull Request Check List

Resolves: #10429

Fix `poetry new` behavior when used in an existing empty directory

This pull request addresses unexpected behavior when running `poetry new` in an already existing empty directory, including the current directory (`.`). It ensures that `poetry new` handles these cases correctly, while maintaining the expected behavior of `poetry init`.

The fix introduces a new argument, `create_layout_if_empty`, to the `_init_pyproject` method in the `InitCommand` class. Additionally, it adds a conditional check to determine whether a full project layout should be created or only a `pyproject.toml` file.

## Summary by Sourcery

Ensure 'poetry new' generates full project layout in existing empty directories while 'poetry init' only creates a pyproject.toml in such cases

Bug Fixes:
- Fix 'poetry new' not creating project structure in existing empty directories

Enhancements:
- Introduce 'allow_layout_creation_on_empty' flag in init_pyproject to control layout creation in empty dirs
- Enable the new flag in the 'new' command to trigger full layout generation in empty dirs

Tests:
- Add tests verifying that 'poetry init' only creates a pyproject.toml in empty or non-empty directories without extra files
- Add tests verifying that 'poetry new' generates full project structure in existing empty directories and when using '.'


radoering:
I think we should prefer to fix/change the behavior rather than document it. See https://github.com/python-poetry/poetry/issues/10429#issuecomment-2980945151.

SoSaymon:
@sourcery-ai review

SoSaymon:
I’ve renamed the flag to something more descriptive.

I decided not to extract the not any(project_path.iterdir()) check into a helper like Path.is_empty(), since it's only used once and introducing a helper felt unnecessary in this context.

Regarding the suggestion about dotfiles: in this case, directories containing dotfiles are deliberately treated as non-empty, which is intentional.

Happy to revisit either the helper function or the dotfile treatment if maintainers feel strongly about it.

radoering (approved):
LGTM

PR #10432, opened 2025-06-16
Use authenticated_url when accessing HTTP BASIC repository

AndyLem:
# Pull Request Check List

Resolves: Authorization error accessing <url> when using http-basic repository and credentials configuration 

## Summary by Sourcery

Bug Fixes:
- Fix authorization failures by constructing request URLs with credentials when using HTTP BASIC repositories.

### DIFF
--- a/src/poetry/repositories/http_repository.py
+++ b/src/poetry/repositories/http_repository.py
@@ -401,7 +401,7 @@ class HTTPRepository(CachedRepository):
         return None
 
     def _get_response(self, endpoint: str) -> requests.Response | None:
-        url = self._url + endpoint
+        url = self.authenticated_url + endpoint
         try:
             response: requests.Response = self.session.get(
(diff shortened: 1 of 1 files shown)


src/poetry/repositories/http_repository.py
```diff
@@ -401,7 +401,7 @@ def calculate_sha256(self, link: Link) -> str | None:
         return None
 
     def _get_response(self, endpoint: str) -> requests.Response | None:
-        url = self._url + endpoint
+        url = self.authenticated_url + endpoint
```
radoering, line 404: I wonder if this is always correct? There is also https://github.com/python-poetry/poetry/blob/fee7d0d2a0d3170cac456f6505471278ccd09ee6/src/poetry/utils/authenticator.py#L196-L198

which seems to handle HTTP basic auth by adding a header if I understand correctly.

Issue #10170, opened 2025-02-11
Prevent poetry from creating projects with invalid project names


Ronimsenn:
### Issue Kind

Change in current behaviour

### Description

Using Poetry 2.0.1 on Windows 11.

When creating a poetry project with a `+` in the project name - might be the same for other invalid characters -, generation runs smoothly:

```console
D:\Development\test> poetry init -v

This command will guide you through creating your pyproject.toml config.

Package name [test]:  test+test
Version [0.1.0]:
Description []:
Author [Your Name <you@example.com>, n to skip]:  n
License []:
Trying to detect current active python executable as specified in the config.
Found: D:\Program Files\Python313\python.EXE
Compatible Python versions [>=3.13]:

Would you like to define your main dependencies interactively? (yes/no) [yes] no
Would you like to define your development dependencies interactively? (yes/no) [yes] no
Generated file

[project]
name = "test+test"
version = "0.1.0"
description = ""
authors = [
    {name = "Your Name",email = "you@example.com"}
]
readme = "README.md"
requires-python = ">=3.13"
dependencies = [
]

[build-system]
requires = ["poetry-core>=2.0.0,<3.0.0"]
build-backend = "poetry.core.masonry.api"

Do you confirm generation? (yes/no) [yes] yes
```

Then, when trying to add a new dependency with `poetry add`, it fails due to an invalid project name:
```console
D:\Development\test> poetry add requests

The Poetry configuration is invalid:
  - project.name must match pattern ^([a-zA-Z\d]|[a-zA-Z\d][\w.-]*[a-zA-Z\d])$
```

### Impact

Prevent the user from creating projects with invalid names.
While the current behaviour is not in any way problematic or breaking, it's unexpected and should be improved in my opinion.

### Workarounds

Use a project name with no `+` :)

rbogart1990:
@Ronimsenn , The issue is that `poetry add` performs validation on the project name, while `poetry init` does not. This allows a project with characters like `+` in the name to be created via `poetry init`, but later causes an error when running `poetry add` due to the stricter validation rules.

I think that we can change this by adding the same validation step that runs during `poetry add` to the `poetry init` process. This will keep the two commands synced up. 

I am interested in working on this! Please feel free to assign the issue to me.

Ronimsenn:
Hey @rbogart1990,
[...]

PR #10457, opened 2025-07-01
Validate user input from init command

rbogart1990:
This PR introduces validation for user input values during the `poetry init` command. It uses the rules defined in the [project-schema.json](https://github.com/python-poetry/poetry-core/blob/002aa3e16f98d21645bb9a45f698b55adc40f317/src/poetry/core/json/schemas/project-schema.json#L10C1-L14C7) schema. These rules are already enforced when running the `poetry add` command, so this change extends the same validation rules to the `poetry init` command.

Previously, user inputs were not validated against these schema rules, allowing the creation of Poetry projects with invalid or disallowed names (e.g. "new+project"). 

This change ensures that inputs conform to the schema before proceeding, preventing invalid project configurations.

# Pull Request Check List

Resolves: #10170

## Summary by Sourcery

Validate user inputs against the project schema in the 'poetry init' command and abort on validation errors.

New Features:
- Add schema-based validation of project metadata in the init command

Enhancements:
- Extend existing validation rules from 'poetry add' to the 'poetry init' command

Tests:
- Add parametrized tests for valid and invalid project names during initialization

### DIFF
--- a/tests/console/commands/test_init.py
+++ b/tests/console/commands/test_init.py
@@ -7,7 +7,6 @@ import textwrap
 
 from pathlib import Path
 from typing import TYPE_CHECKING
-from typing import Any
 
 import pytest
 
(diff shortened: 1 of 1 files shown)


src/poetry/console/commands/init.py
```diff
@@ -265,6 +267,14 @@ def _init_pyproject(
 
             return 1
 
+        # Validate fields before creating pyproject.toml file. If any validations fail, throw an error.
+        # Convert TOML string to a TOMLDocument (a dict-like object) for validation.
+        pyproject_dict = parse(pyproject.data.as_string())
+        validation_results = self._validate(pyproject_dict)
```
radoering: `Factory.validate(pyproject.data)` should suffice.

src/poetry/console/commands/init.py
```diff
@@ -265,6 +267,14 @@ def _init_pyproject(
...
+        if validation_results.get("errors"):
+            self.line_error(f"<error>Validation failed: {validation_results}</error>")
+            return 1
```
radoering: The output is not so nicely formatted. It contains a raw dict, e.g.:
```
Validation failed: {'errors': ['project.name must match pattern ^([a-zA-Z\\d]|[a-zA-Z\\d][\\w.-]*[a-zA-Z\\d])$'], 'warnings': []}
```

You may take a look at https://github.com/python-poetry/poetry-core/blob/002aa3e16f98d21645bb9a45f698b55adc40f317/src/poetry/core/factory.py#L53-L60 for better formatting.

src/poetry/console/commands/init.py
```diff
@@ -533,3 +543,13 @@ def _get_pool(self) -> RepositoryPool:
             self._pool.add_repository(PyPiRepository(pool_size=pool_size))
 
         return self._pool
+
+    @staticmethod
+    def _validate(pyproject_data: dict[str, Any]) -> dict[str, Any]:
```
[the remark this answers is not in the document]
radoering: Poetry does not support Python 3.8 anymore.

radoering (changes requested):

src/poetry/console/commands/init.py
rbogart1990: @radoering , thanks for the feedback about this. 

Would this format be acceptable?
```
Validation failed:
  - project.name must match pattern ^([a-zA-Z\d]|[a-zA-Z\d][\w.-]*[a-zA-Z\d])$
```

radoering: Absolutely.

rbogart1990: Thank you!

rbogart1990: @radoering , change made.

src/poetry/console/commands/init.py
rbogart1990: @radoering , thanks for the suggestion. I updated the code to call `Factory.validate(pyproject.data)` directly. All tests are passing!

I kept it inside of `InitCommand()._validate()` so that I can unit-test it.

src/poetry/console/commands/init.py
Secrus: You must be kidding. Error messages should be readable and clear for users. Imagine some beginner developer getting hit by that error message. Hell, 9/10 seasoned developers would have a hard time understanding that error message without some sort of regex explorer. This should be a clear message, that a 5 year old is able to understand. I would split the checks and make a separate clear message for each condition, gather the errors and list them clearly. Poetry's motto is "Python packaging and dependency management made easy". Getting hit with a regex is not "easy" by any means.

radoering: My comment was just about formatting the output. At this point, we get a dict with lists of messages and the output should just be nicely formatted. The content of the single messages has been created before and is fixed at this point.

In my opinion, it is too much effort to improve each possible message that comes from schema validation - or at least this is clearly out of scope of this PR. The message is the same message if you run `poetry check` on such an invalid pyproject.toml.

src/poetry/console/commands/init.py
radoering: Hmm, I see. So we are actually just unit testing `Factory.validate()`, which does not give much value because it is (or at least should be) already tested in `test_factory.py`.

Further, my comment suggests that converting `pyproject.data` to string and parsing it is unnecessary. You probably also do this just for unit testing to have a clear interface? I do not like that the production code becomes more complicated just for unit testing.

I think you should check the other (non-interactive) unit tests in `test_init.py` and use one of these as base for your unit test so that you do not need a separate method to test. Further, this is probably the wrong place to test the regex in detail because as mentioned the validation is part of `Factory`. Here, we should just check that validation is called and gives a nicely formatted output.

PR #10469, opened 2025-07-18
feat: adds an option to generate a project with pacakge mode = false

danielepintore:
# Pull Request Check List

Resolves: #9447 

This small pull request adds an option to disable package mode: `-N` and `----disable-package-mode`.
When this options are enabled optional field in the `pyproject.toml` are skipped if the user leaves the default value

### DIFF
--- a/src/poetry/console/commands/init.py
+++ b/src/poetry/console/commands/init.py
@@ -58,6 +58,7 @@ class InitCommand(Command):
             flag=False,
             multiple=True,
         ),
+        option("disable-package-mode", "-N", "Disables package mode.", flag=True),
         option("license", "l", "License of the package.", flag=False),
     ]
 
@@ -141,9 +142,12 @@ The <c1>init</c1> command creates a basic <comment>pyproject.toml</> file in the
 
         if is_interactive:
             question = self.create_question(
-                f"Version [<comment>{version}</comment>]: ", default=version
+                f"Version [<comment>{version}</comment>]: ",
+                default=version if not self.option("disable-package-mode") else "",
             )
             version = self.ask(question)
+        else:
+            version = version if not self.option("disable-package-mode") else ""
 
         description = self.option("description") or ""
         if not description and is_interactive:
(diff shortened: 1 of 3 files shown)


src/poetry/console/commands/init.py
```diff
@@ -58,6 +58,7 @@ class InitCommand(Command):
             flag=False,
             multiple=True,
         ),
+        option("disable-package-mode", "-N", "Disables package mode.", flag=True),
```
radoering, line 61: I am wondering about the short name. Does it stand for "non-package-mode"? If that is the case, should we name the option `non-package-mode`?

src/poetry/console/commands/init.py
```diff
@@ -141,9 +142,12 @@ def _init_pyproject(
 
         if is_interactive:
             question = self.create_question(
-                f"Version [<comment>{version}</comment>]: ", default=version
+                f"Version [<comment>{version}</comment>]: ",
```
radoering, line 145: If package-mode is disabled, this now shows a version, which is not the default. We should probably set the version some lines above depending on the `disable-package-mode` option.

tests/console/commands/test_init.py
```diff
@@ -1207,3 +1207,108 @@ def test_init_does_not_create_project_structure_in_non_empty_directory(
...
+    tester.execute("-N", inputs="\n".join(inputs))
+
+    assert (source_dir / "pyproject.toml").exists()
+    toml_content = (source_dir / "pyproject.toml").read_text(encoding="utf-8")
+    assert "version = " not in toml_content
+    assert "description =" not in toml_content
+    assert "authors =" not in toml_content
+
+
+def test_init_with_no_package_mode_flag_active_remove_optional_fields_with_email(
```
radoering, line 1289: This should test all fields that may be removed, not just `authors` but also `version` and `description`.

deronnax:
It has not moved forward for a month. If @danielepintore can't finish it, I volunter to take it over.

PR #10491, opened 2025-08-03
fix: adjust virtualenv constraint in pyproject.toml to <= 20.32.0

yonathan-wolloch-lendbuzz:
This is introduced as virtualenv introduced a breaking change to poetry in v20.33.0.
fixes #10490 

# Pull Request Check List

Resolves: #issue-number-here

- [ ] Added **tests** for changed code.
- [ ] Updated **documentation** for changed code.

PR #10493, opened 2025-08-04
release: bump version to 2.1.4

radoering:
### Changed

- Require `virtualenv<20.33` to work around an issue where Poetry uses the wrong Python version ([#10491](https://github.com/python-poetry/poetry/pull/10491)).
- Improve the error messages for the validation of the `pyproject.toml` file ([#10471](https://github.com/python-poetry/poetry/pull/10471)).

### Fixed

- Fix an issue where project plugins were installed even though `poetry install` was called with `--no-plugins` ([#10405](https://github.com/python-poetry/poetry/pull/10405)).
- Fix an issue where dependency resolution failed for self-referential extras with duplicate dependencies ([#10488](https://github.com/python-poetry/poetry/pull/10488)).

### Docs

- Clarify how to include files that were automatically excluded via VCS ignore settings ([#10442](https://github.com/python-poetry/poetry/pull/10442)).
- Clarify the behavior of `poetry add` if no version constraint is explicitly specified ([#10445](https://github.com/python-poetry/poetry/pull/10445)).


esafak:
@radoering Please can you separately try reverting #9413 to fix it, so I know if something needs to be done at virtualenv's end?

radoering:
@esafak I will try - but probably not until the end of the week. (I still have to understand/reproduce the issue on my machine.)

branchvincent (approved):

radoering:
@esafak Using `--python` seems to work. See https://github.com/python-poetry/poetry/pull/10506. Thus, it looks like there is nothing to do at virtualenv's end.

Issue #10490, opened 2025-08-03
bug: poetry is using pipx python instead of venv python as result of new version in virtualenv v20.33.0


yonathan-wolloch-lendbuzz:
### Description

poetry is using the wrong python version when running in github action. it uses the machine default version instead of the venv python version.
```
Using virtualenv: /home/runner/.cache/pypoetry/virtualenvs/lb-kafka-C-kvVqEG-py3.13
....
error:
Current Python version (3.12.3) is not allowed by the project (~=3.13).
Please change python executable via the "env use" command.
Error: Process completed with exit code 1.
```

### Workarounds

locking virtualenv==20.32.0 when installing poetry using constraints.txt as pip args fixes the issue.

### Poetry Installation Method

pipx

### Operating System

Ubuntu

### Poetry Version

2.1.3

### Poetry Configuration

```bash session
cache-dir = "/home/runner/.cache/pypoetry"
data-dir = "/home/runner/.local/share/pypoetry"
installer.max-workers = null
installer.no-binary = null
installer.only-binary = null
installer.parallel = true
installer.re-resolve = true
keyring.enabled = true
python.installation-dir = "{data-dir}/python"  # /home/runner/.local/share/pypoetry/python
requests.max-retries = 0
solver.lazy-wheel = true
system-git-client = false
virtualenvs.create = true
virtualenvs.in-project = null
virtualenvs.options.always-copy = false
virtualenvs.options.no-pip = false
virtualenvs.options.system-site-packages = false
virtualenvs.path = "{cache-dir}/virtualenvs"  # /home/runner/.cache/pypoetry/virtualenvs
virtualenvs.prompt = "{project_name}-py{python_version}"
virtualenvs.use-poetry-python = false
```

### Python Sysconfig

<details>
  <summary>sysconfig.log</summary>
  

  ```
  Paste the output of 'python -m sysconfig', over this line.
  ```
</details>

### Example pyproject.toml

```TOML

```

### Poetry Runtime Logs

<details>
  <summary>poetry-runtime.log</summary>
  

  ```
  Stack trace:

  8  /opt/pipx/venvs/poetry/lib/python3.12/site-packages/cleo/application.py:327 in run
      325│ 
      326│             try:
    → 327│                 exit_code = self._run(io)
      328│             except BrokenPipeError:
      329│                 # If we are piped to another process, it may close early and send a

[...]

PR #10506, opened 2025-08-09
allow virtualenv >= 20.33.0 again

radoering:
I was able to reproduce the issue with `virtualenv>=20.33.0` reported in #10490. Using `--python` as proposed in https://github.com/python-poetry/poetry/issues/10490#issuecomment-3148615579 seems to fix the issue. Thus, we can allow newer virtualenv versions again.

## Summary by Sourcery

Allow support for virtualenv >=20.33.0 by removing the version upper bound and updating the invocation flag for environment creation

Bug Fixes:
- Restore compatibility with virtualenv >=20.33.0 by removing the <20.33.0 upper version constraint in pyproject.toml

Enhancements:
- Use --python flag instead of deprecated --try-first-with when invoking virtualenv during environment setup


ahasha-ml:
Our team is having CI issues related to this bug.  Is there a timeframe for when it will be addressed?

radoering:
@ahasha-ml I do not understand how this can still cause issues because we addressed the bug in a quickfix by not allowing `virtualenv>=20.33.0` in Poetry 2.1.4 (see #10491). That is why this PR is not considered "urgent".

This PR will be included in the next minor release, which might be released in September or October.

ArcticLampyrid:
FYI, I infer that #9413 was merged as a workaround for https://github.com/pypa/virtualenv/issues/2285 (later fixed by https://github.com/pypa/virtualenv/pull/2821 and included in [virtualenv 20.29.1](https://github.com/pypa/virtualenv/releases/tag/20.29.1) released on Jan 18, 2025).
However, due to https://github.com/pypa/virtualenv/pull/2921, we encountered this bug.

Since virtualenv v20.29.1 is quite new, do we still need to maintain compatibility for users who install Poetry via system package managers (which may still ship an older virtualenv)?

radoering:
@ArcticLampyrid You mean that we should keep using `--try-first-with` with older virtualenv versions? That might make sense as long as we allow these older versions. On the other side, as stated in https://github.com/python-poetry/poetry/issues/10490#issuecomment-3148615579, the issue that #9413 tried to fix was not well understood and it did not influence many users. In other words, `--python` will probably also work with older virtualenv versions for most users.

PR #10509, opened 2025-08-18
chore: update locked dependencies

radoering:
## Summary by Sourcery

Update locked dependencies and adjust test type ignore directives

Tests:
- Remove obsolete type ignore comments on assignments in puzzle and incompatibility tests
- Add a type ignore for no-any-return in the helpers test

Chores:
- Refresh poetry.lock with updated dependency versions

### DIFF
--- a/tests/helpers.py
+++ b/tests/helpers.py
@@ -265,7 +265,7 @@ def make_entry_point_from_plugin(
     )
 
     if dist:
-        ep = ep._for(dist)  # type: ignore[attr-defined]
+        ep = ep._for(dist)  # type: ignore[attr-defined,no-untyped-call]
         return ep
 
     return ep


tests/helpers.py
```diff
@@ -266,7 +266,7 @@ def make_entry_point_from_plugin(
 
     if dist:
         ep = ep._for(dist)  # type: ignore[attr-defined,no-untyped-call]
-        return ep
+        return ep  # type: ignore[no-any-return]
```
dimbleby: suggestion: prefer to change the previous line, as
```python
ep: metadata.EntryPoint = ep._for(dist)
```
which is more likely to catch errors if and when `ep._for()` ever gains type annotations

Issue #7832, opened 2023-04-24
(🐞) No error message when `FileLocked`  appears for `poetry show --outdated`


KotlinIsland:
- **Poetry version**:  1.4.2
- **Python version**:  3.11.2
- **OS version and name**:  Windows 10
- **pyproject.toml**:  

- [🚀] I am on the [latest](https://github.com/python-poetry/poetry/releases/latest) stable Poetry version, installed using a recommended method.
- [🚀] I have searched the [issues](https://github.com/python-poetry/poetry/issues) of this repo and believe that this is not a duplicate.
- [🚀] I have consulted the [FAQ](https://python-poetry.org/docs/faq/) and [blog](https://python-poetry.org/blog/) for any relevant entries or release notes.
- [🚀] If an exception occurs when executing a command, I executed it again in debug mode (`-vvv` option) and have included the output below.

## Issue

```
👉 poetry show --outdated

(b'C:\\Users\\amongus\\AppData\\Local\\pypoetry\\Cache\\src\\pylint-module-boundaries\\.git\\refs\\remotes\\origin\\fix-multiple-rules', b'C:\\Users\\amongus\\AppData\\Local\\pypoetry\\Cache\\src\\pylint-module-boundaries\\.git\\refs\\remotes\\origin\\fix-multiple-rules.lock')                                                                                                                                                                           
👉  poetry show --outdated -v
Using virtualenv: C:\Users\amongus\AppData\Local\pypoetry\Cache\virtualenvs\myhr-a5tggzGn-py3.11

  FileExistsError

  [Errno 17] File exists: b'C:\\Users\\amongus\\AppData\\Local\\pypoetry\\Cache\\src\\pylint-module-boundaries\\.git\\refs\\remotes\\origin\\fix-multiple-rules.lock'

  at ~\AppData\Roaming\pypoetry\venv\Lib\site-packages\dulwich\file.py:147 in __init__
      143│             self._lockfilename = self._filename + b".lock"
      144│         else:
      145│             self._lockfilename = self._filename + ".lock"
      146│         try:
    → 147│             fd = os.open(
      148│                 self._lockfilename,
      149│                 os.O_RDWR | os.O_CREAT | os.O_EXCL | getattr(os, "O_BINARY", 0),
      150│                 mask,
      151│             )

The following error occurred when trying to handle this error:


  FileLocked

[...]

PR #10532, opened 2025-09-06
git: be more robust against dangling lock files

radoering:
# Pull Request Check List

Resolves: #7832

## Summary by Sourcery

Make Git clone more robust against stale lock files by retrying ref imports and deleting dangling locks

Bug Fixes:
- Prevent clone failures due to existing lock files in refs directory

Enhancements:
- Retry importing refs up to twice when encountering a FileLocked error
- Remove stale lock files automatically before retrying ref imports

Tests:
- Add test to verify that cloning removes existing lock files and succeeds
- Include a tag in the temporary Git repository fixture for testing locked tags

### DIFF
--- a/src/poetry/vcs/git/backend.py
+++ b/src/poetry/vcs/git/backend.py
@@ -17,6 +17,7 @@ from dulwich.client import get_transport_and_path
 from dulwich.config import ConfigFile
 from dulwich.config import parse_submodules
 from dulwich.errors import NotGitRepository
+from dulwich.file import FileLocked
 from dulwich.index import IndexEntry
 from dulwich.refs import ANNOTATED_TAG_SUFFIX
 from dulwich.repo import Repo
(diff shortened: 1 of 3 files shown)


dimbleby:
automatically removing a lock file surely undermines the point of the lock file?

radoering:
> automatically removing a lock file surely undermines the point of the lock file?

Sure, but that is what many users (including myself) do manually if they encounter this state. Of course, it would be better if git/dulwich could handle it by itself. But if they do not, removing it may be a practical way to deal with it. I assume the lock file should not be there at this point anyway. Otherwise, we should at least give a better error message like "try again in some seconds/minutes and if you still experience this issue remove the lock file manually."

Maybe, @jelmer can assess if we risk some other (more serious) issues if we remove the lock file.

dimbleby:
Agree that a better error message would be desirable.  Also agree that if dulwich lock files can dangle then that suggests a bug in dulwich.

>  I assume the lock file should not be there at this point

I expect that the lock file serves some purpose - and that poetry assuming that dulwich has gone wrong would defeat that purpose. 

A user is much better placed to decide whether they are indeed in the middle of performing two operations at once, or whatever, and to make a judgment about whether the lock file is really dangling - or is just doing its job.

jelmer:
I've added a comment in https://github.com/python-poetry/poetry/issues/7832#issuecomment-3263681559 - I think we should try to understand first why this is happening. The lock files aren't specific to Dulwich.

radoering:
Superseded by #10535

PR #10567, opened 2025-09-24
Fix Python executable resolution when 'python' not found in virtual environment

deronnax:
## Problem

Poetry fails with `[Errno 2] No such file or directory: 'python'` when trying to run Python scripts in virtual environments where the 'python' executable is not available in the system PATH.

This is particularly problematic because:
1. Poetry downloads and manages its own Python installations
2. Poetry creates proper symlinks in its managed Python directories  
3. But when Poetry's `_bin()` method can't find an executable in the virtual environment, it falls back to returning just the executable name (e.g., 'python')
4. This causes subprocess calls to look for 'python' in the system PATH instead of using Poetry's own managed Python installations

## Root Cause

In `src/poetry/utils/env/base_env.py`, the `_bin()` method has this problematic fallback:

```python
if not bin_path.exists():
    # ... Windows-specific checks ...
    return bin  # ❌ Returns just 'python' instead of absolute path
```

When `run_python_script()` calls `self.run(self._executable, ...)`, it goes through:
1. `run()` → `get_command_from_bin()` → `_bin()`
2. If `_bin()` can't find the executable, it returns just the name
3. subprocess then looks for 'python' in system PATH and fails

## Solution

This PR adds a fallback mechanism in `_bin()` that:
1. Detects when the requested executable is 'python' or starts with 'python'
2. Attempts to find a compatible Python in Poetry's managed installations
3. Returns the absolute path to Poetry's managed Python if found
4. Only falls back to the original behavior if no Poetry-managed Python is available

## Code Changes

The fix adds this logic to `_bin()` method:

```python
# If the executable is 'python' and not found in the virtual environment,
# try to find it in Poetry's managed Python installations
if bin == "python" or bin.startswith("python"):
    from poetry.utils.env.python import Python
    try:
        # Try to find a compatible Python version
        for python in Python.find_poetry_managed_pythons():
            python_path = python.executable
            if python_path and python_path.exists():
                return str(python_path)
    except Exception:
        # If we can't find Poetry-managed Python, fall back to system lookup
        pass
```

## Impact

This change:
-  Fixes the 'Errno 2 No such file or directory python' error
-  Allows Poetry to use its own managed Python installations as intended
-  Maintains backward compatibility
-  Only affects the fallback case when executables aren't found in virtual environments
-  Gracefully degrades to original behavior if Poetry-managed Python lookup fails

Resolves the design inconsistency where Poetry manages its own Python versions but doesn't use them when subprocess calls need to find Python executables.

## Summary by Sourcery

Add fallback mechanism to resolve 'python' executables via Poetry-managed installations when missing in virtual environments

Bug Fixes:
- Fix Errno 2 'python' not found error by locating and using Poetry-managed Python installations when the executable is absent

Enhancements:
- Fall back to original behavior if no Poetry-managed Python is available

### DIFF
--- a/src/poetry/utils/env/base_env.py
+++ b/src/poetry/utils/env/base_env.py
@@ -484,6 +484,21 @@ class Env(ABC):
                 if bin_path.exists():
                     return str(bin_path)
 
+            # If the executable is 'python' and not found in the virtual environment,
+            # try to find it in Poetry's managed Python installations
+            if bin == "python" or bin.startswith("python"):
+                from poetry.utils.env.python import Python
+
+                try:
+                    # Try to find a compatible Python version
+                    for python in Python.find_poetry_managed_pythons():
+                        python_path = python.executable
+                        if python_path and python_path.exists():
+                            return str(python_path)
+                except Exception:
+                    # If we can't find Poetry-managed Python, fall back to system lookup
+                    pass
+
             return bin
 
         return str(bin_path)


deronnax:
I have doubt on this solution, I will come back later with a fix I am more sure about. Closing meanwhile.

finswimmer:
Hey @deronnax,

please always mention a related issue ticket, so we can easily follow which problem you are trying to solve. If no ticket exists yet, please open one and wait, if any of us maintainer confirms that there is a bug, before creating a pull request. Otherwise you might be disappointed, if we later find out, that there is no issue and something is be design how it is and now you've already put afford and time into something.

But thanks a lot that you are willing to contribute!

fin swimmer

deronnax:
Duely noted. ~In this case, it was a bug I ran into several time and it determisticaly reprodicible, so I took the liberty to directly do the PR.
I will open an issue, meanwhile.~
Ahahaha, can't reproduce anymore. I don't understand. I see this error from times to times, but I can't pin it. Probably happens when shell environment has been tampered (it happens e.g when you delete a virtualenv that is activated in the current shell), but in this case it's not a Poetry bug (is it even a bug?)

Issue #9393, opened 2024-05-06
`poetry install` run under pre-commit hook does not apply changes.


thisiswhereitype:
### Description

`pre-commit run poetry-install` passed but acted as no-op.

The expected package was not in the venv.
A regular: `poetry install` noticed the missing package.

### Workarounds

Manually poetry install.

### Poetry Installation Method

pipx

### Operating System

Win 10

### Poetry Version

v1.8.2

### Poetry Configuration

```bash session
> poetry config --list
cache-dir = "~\\AppData\\Local\\pypoetry\\Cache"
experimental.system-git-client = false
installer.max-workers = null
installer.modern-installation = true
installer.no-binary = null
installer.parallel = true
keyring.enabled = true
solver.lazy-wheel = true
virtualenvs.create = true
virtualenvs.in-project = true
virtualenvs.options.always-copy = false
virtualenvs.options.no-pip = false
virtualenvs.options.no-setuptools = false
virtualenvs.options.system-site-packages = false
virtualenvs.path = "{cache-dir}\\virtualenvs"  # ~\AppData\Local\pypoetry\Cache\virtualenvs
virtualenvs.prefer-active-python = false
virtualenvs.prompt = "{project_name}-py{python_version}"
warnings.export = true
```

### Python Sysconfig

```bash session
> python -m sysconfig 
Platform: "win-amd64"
Python version: "3.12"
Current installation scheme: "venv"

Paths:
        data = "<repo>\.venv"
        include = "~\scoop\apps\python\3.12.3\Include"
        platinclude = "~\scoop\apps\python\3.12.3\Include"
        platlib = "<repo>\.venv\Lib\site-packages"
        platstdlib = "<repo>\.venv\Lib"
        purelib = "<repo>\.venv\Lib\site-packages"
        scripts = "<repo>\.venv\Scripts"
        stdlib = "~\scoop\apps\python\3.12.3\Lib"

Variables:
        BINDIR = "<repo>\.venv\Scripts"
        BINLIBDEST = "<repo>\.venv\Lib"
        EXE = ".exe"
        EXT_SUFFIX = ".cp312-win_amd64.pyd"
        INCLUDEPY = "~\scoop\apps\python\3.12.3\Include"
        LIBDEST = "~\scoop\apps\python\3.12.3\Lib"
        TZPATH = ""
        VERSION = "312"
        VPATH = "..\.."
        abiflags = ""
        base = "<repo>\.venv"
[...]

PR #10574, opened 2025-09-30
Fix poetry-install pre-commit hook configuration

abn:
- Change language from python to system for poetry-install hook
- Add default args: ["--sync"] to poetry-install hook
- Update documentation with warning about versions up to 2.2.1

Resolves: #9393

### DIFF
--- a/.pre-commit-hooks.yaml
+++ b/.pre-commit-hooks.yaml
@@ -18,7 +18,8 @@
   name: poetry-install
   description: run poetry install to install dependencies from the lock file
   entry: poetry install
-  language: python
+  language: system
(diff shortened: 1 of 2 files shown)


.pre-commit-hooks.yaml
```diff
@@ -18,7 +18,8 @@
   name: poetry-install
   description: run poetry install to install dependencies from the lock file
   entry: poetry install
-  language: python
+  language: system
+  args: ["--sync"]
```
radoering, line 22: This is unlucky because we deprecated `--sync` in favor of `poetry sync`.

Issue #10586, opened 2025-10-16
`poetry remove` crashes when removing a package from a legacy dependency group (PEP 735) containing `include-group`


Wellshh:
### Description

When attempting to remove a package from a dependency group that is defined using the PEP 735 format, the `poetry remove` command crashes and throws `AttributeError: 'InlineTable' object has no attribute 'split'` if nested group is used (the groups's dependency list contains { include-group = "..."} ), which is supported for Poetry 2.2.1: [Including dependencies from other groups](https://python-poetry.org/docs/managing-dependencies/#including-dependencies-from-other-groups). 

### Analysis
The root cause appears to be in the implementation of the remove command. When processing a dependency group defined under the legacy [dependency-groups] key, the code iterates through a list that can contain mixed types (package strings and include-group tables). The command's logic does not filter out the non-string table objects before passing them to Dependency.create_from_pep_508, which expects a string.

### Workarounds

May use the poetry style of nested poetry group:
```toml
[tool.poetry.group.test.dependencies]
pytest = "^8.0.0"

[tool.poetry.group.lint.dependencies]
ruff = "^0.11"

[tool.poetry.group.dev]
include-groups = [
    "test",
    "lint",
]
```

### Poetry Installation Method

pipx

### Operating System

MacOS Tahoe 26.0.1

### Poetry Version

Poetry (version 2.2.1)

### Poetry Configuration

```bash session
cache-dir = "/Users/wells/Library/Caches/pypoetry"
data-dir = "/Users/wells/Library/Application Support/pypoetry"
installer.max-workers = null
installer.no-binary = null
installer.only-binary = null
installer.parallel = true
installer.re-resolve = true
keyring.enabled = true
python.installation-dir = "{data-dir}/python"  # /Users/wells/Library/Application Support/pypoetry/python
requests.max-retries = 0
solver.lazy-wheel = true
system-git-client = false
virtualenvs.create = true
virtualenvs.in-project = null
virtualenvs.options.always-copy = false
virtualenvs.options.no-pip = false
virtualenvs.options.system-site-packages = false
virtualenvs.path = "{cache-dir}/virtualenvs"  # /Users/wells/Library/Caches/pypoetry/virtualenvs
virtualenvs.prompt = "{project_name}-py{python_version}"
virtualenvs.use-poetry-python = false
```

### Python Sysconfig

<details>
  <summary>sysconfig.log</summary>
  

  ```
  Paste the output of 'python -m sysconfig', over this line.
  ```
[...]

PR #10587, opened 2025-10-16
fix error of removing packages if nested groups (include-group) are used in pep735 style

Wellshh:
# Pull Request Check List

Resolves: #10586 

Tests:
- Add parameterized tests for removing packages from nested dependency groups in both PEP735 and legacy poetry styles

### DIFF
--- a/tests/console/commands/test_remove.py
+++ b/tests/console/commands/test_remove.py
@@ -754,6 +754,7 @@ foobar = [
 ]
 """
     else:
+        # This also checks that the include-groups section is not removed even if its explicit dependencies are empty
         assert "dependencies" not in content["group"]["foobar"]
         expected = """\
 [tool.poetry.group.bar.dependencies]


tests/console/commands/test_remove.py
```diff
@@ -668,3 +668,107 @@ def test_remove_with_lock_does_not_perform_uninstall_op(
...
+
+    app.poetry.package.add_dependency(Factory.create_dependency("foo", "^2.0.0"))
+    app.poetry.package.add_dependency(
+        Factory.create_dependency("foo", "^2.0.0", groups=["bar"])
+    )
+    app.poetry.package.add_dependency(
+        Factory.create_dependency("baz", "^1.0.0", groups=["foobar"])
+    )
+
+    tester.execute("foo")
```
radoering: Shouldn't we try to remove a dependency from the group with the include, i.e. remove `baz` from `foobar`?

Reading #10586:

> When attempting to remove a package from a dependency group that is defined using the PEP 735 format, the `poetry remove` command crashes  [...] if nested group is used (the groups's dependency list contains { include-group = "..."} )

(Actually, I think this test shows another issue: The group `bar` is removed, but `foobar` still includes `bar`, which probably will result in an error the next time you run poetry.)

Wellshh: Hi @radoering,

Thank you so much for your review and valuable feedback! This is my first time contributing to an open-source project, and I'm learning a lot from your guidance.

After thinking about your suggestion, I completely agree. You've made a great point. Testing the removal of the dependency directly from the group that uses include-group is indeed a more direct, clear, and robust approach. It makes the test's purpose much more explicit. I will update the test case as you recommended and push a new commit shortly.

Regarding the other potential issue you mentioned—and I'm glad you brought it up, as I was thinking about it too:

> (Actually, I think this test shows another issue: The group bar is removed, but foobar still includes bar, which probably will result in an error the next time you run poetry.)

When I was working on this, I noticed that poetry removes a group when its last dependency is gone. However, it leaves a "dangling" include-group entry pointing to the now-deleted group.

My initial thought was that perhaps this was intentional, to preserve the user's original dependency structure design. I personally find it a bit disruptive when the tool removes an entire group I've defined. But on further reflection, as you pointed out, this dangling reference will almost certainly cause errors later and feels more like a bug—perhaps due to logic not being fully synchronized between poetry and poetry-core.

This leads me to a question for you: Would you prefer that I open a separate issue to track this "dangling include" problem, or should I try to fix it within this same PR? I'm happy to do whichever is best for the project's workflow.

Wellshh: Also an add-on, there seems to be inconsistency between the two styles' behavior after removing dependencies:

**PEP 735 format (`[dependency-groups]`):**
- When all dependencies are removed from a group, the group still remains containing the 'include-group' entry (because `include-group` is included into `standard_section`):
```toml
# Original
[dependency-groups]
bar = [
    "foo (>=2.0,<3.0)",
]
foobar = [
    { include-group = "bar" },
    "baz (>=1.0,<2.0)",
]
```
```toml
# After remove
[dependency-groups]
bar = [
    "foo (>=2.0,<3.0)",
]
foobar = [
    { include-group = "bar" },
]
```
**Legacy format (`[tool.poetry.group]`):**
- When all dependencies are removed, the group is deleted even if it contains `include-groups` under `tool.poetry.group.xx`(because they are ignored when `poetry_groups_content` is created):
```toml
# Original
[tool.poetry.group.bar.dependencies]
foo = "^2.0.0"

[tool.poetry.group.foobar]
include-groups = [
    "bar",
]
[tool.poetry.group.foobar.dependencies]
baz = "^1.0.0"
```
```toml
# After remove
[tool.poetry.group.bar.dependencies]
foo = "^2.0.0"
```

radoering: > This leads me to a question for you: Would you prefer that I open a separate issue to track this "dangling include" problem, or should I try to fix it within this same PR?

I think it makes sense to handle this in a separate PR.

> My initial thought was that perhaps this was intentional, to preserve the user's original dependency structure design. I personally find it a bit disruptive when the tool removes an entire group I've defined.

I think without `include-group` it did not matter that much because a group without dependencies did not do anything and if you added a dependency later the group was recreated. However, with `include-groups` it might be better to not remove the group (at least if it is included somewhere) because the include will not be readded. (I am still not 100% sure about the desired behavior. That is also why I think we should tackle this in a separate PR.)

> When all dependencies are removed, the group is deleted even if it contains include-groups under tool.poetry.group.xx(because they are ignored when poetry_groups_content is created):

Good catch. The group should not be removed. If you have a simple fix for this, feel free to add it to this PR.

radoering (approved):

Issue #10559, opened 2025-09-19
Show path to poetry self venv python


jaredahern:
### Issue Kind

Change in current behaviour

### Description

There doesn't appear to be a way to get the path to the python used for the poetry installation. One option could be to implement a `poetry self env info` command. However, perhaps the easiest way to expose this information would be to add it to `poetry debug info`. Either or both (or probably something else) would work.

If I've just missed this capability, please let me know!

### Impact

Knowledge of the python location can be needed, e.g. for patching the venv to handle self-signed certificates. 

### Workarounds

Use `which poetry` can start the search, but that doesn't necessarily point to the real location of the poetry executable, and its sibling python. If installed with `pipx` there is probably a way to do this using that tool.

radoering:
> One option could be to implement a `poetry self env info` command. However, perhaps the easiest way to expose this information would be to add it to `poetry debug info`.

Both sounds reasonable and would probably be accepted.

jaredahern:
Thanks! I'll try to propose a PR.

dimbleby:
Why do you want this information?  It is only an implementation detail that poetry is written in python at all, I don't know why the interpreter being used should be of interest to users?

Asking the same question another way: do you raise the same issue against every program that happens to be written in python?  Why (or why not)?

jaredahern:
Fair points, @dimbleby. The use case I have in mind is installing certificates to handle things like SSL inspection in a corporate environment. In my experience, reliably addressing the various ways that certs could need to be hooked into python often requires things to be configured each affected venv. Since poetry may need to reach out to various repos and across network boundaries, a venv hosting poetry is of interest.

I would say no, I wouldn't typically raise this against other python programs. However, my thought process is:

1. Poetry has an official installer, outside the usual process for installing Python tools. It places poetry in a separate venv, and can be controlled by POETRY_HOME. Depending on how it winds up in someone's PATH (e.g. are there symlinks?), the venv python location might not be immediately obvious in the future.
2. Poetry already provides debug information about the project venv and the base python for the project venv, so it doesn't seem that this is too far outside the current scope.

Perhaps there is an obvious solution outside of Poetry that I am missing. I'm very happy to get a suggestion!

FWIW, I think this could possibly just be a one-line addition to print `sys.executable` in `src/poetry/console/commands/debug/info.py`, plus likely a test change (I haven't looked yet). I don't think finding the env or checking `is_sane` need to be considered, but could be. No docs appear to be impacted, but info about this could be added.

[...]

PR #10588, opened 2025-10-19
Show Python path for Poetry in `debug info`

jaredahern:
# Pull Request Check List

Resolves: #10559

# Summary

Adds the executable and environment path for the Python executable running Poetry to `poetry self debug`.

I believe the `sys` calls are correct, but happy to adjust as needed. Also, I did not see any docs that required updating.

## Summary by Sourcery

Include the Python environment path and executable in the output of `poetry self debug`.

Enhancements:
- Add 'Path' and 'Executable' fields under the Poetry section in debug info
- Adjust label alignment for Version and Python fields

Tests:
- Add tests to verify the display of Poetry's path and executable in debug info

### DIFF
--- a/src/poetry/console/commands/debug/info.py
+++ b/src/poetry/console/commands/debug/info.py
@@ -2,6 +2,8 @@ from __future__ import annotations
 
 import sys
 
+from pathlib import Path
+
 from poetry.console.commands.command import Command
 
 
@@ -19,8 +21,8 @@ class DebugInfoCommand(Command):
                 [
                     f"<info>Version</info>:    <comment>{self.poetry.VERSION}</>",
                     f"<info>Python</info>:     <comment>{poetry_python_version}</>",
-                    f"<info>Path</info>:       <comment>{sys.prefix}</>",
-                    f"<info>Executable</info>: <comment>{sys.executable or 'Unknown'}</>",
+                    f"<info>Path</info>:       <comment>{Path(sys.prefix)}</>",
+                    f"<info>Executable</info>: <comment>{Path(sys.executable) if sys.executable else 'Unknown'}</>",
                 ]
             )
         )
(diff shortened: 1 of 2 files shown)


radoering (commented):
LGTM except for the failing Windows tests. Can you take a look, please?

jaredahern:
Hmm. I took a look at the errors, and tried to resolve them. Not sure if I totally have. Here's the original error:

```
    Poetry
    Version:    2.2.1
    Python:     3.10.11
    Path:       /poetry/prefix
    Executable: /poetry/prefix/bin/python

    Virtualenv
    Python:         3.7.0
    Implementation: CPython
    Path:           \prefix
  - Executable:     /poetry/prefix/bin/python
  ?                 ^      ^      ^   ^
  + Executable:     \poetry\prefix\bin\python
  ?                 ^      ^      ^   ^
    Valid:          True

    Base
    Platform:   darwin
    OS:         posix
    Python:     3.10.11
    Path:       \base\prefix
    Executable: python
```

So here, the error appears to be with the output from the `env info` command, which I haven't touched, and directly copied the test for. It confuses me that the related `test_env_info_displays_complete_info` test does not fail.

That being said, I tried to be more careful with paths, to better support Windows. Happy to take pointers. I rebased onto main and added a commit. There are now different failing tests, which seem to be unrelated to the changes here. Any thoughts on how I should proceed?

radoering:
> So here, the error appears to be with the output from the `env info` command, which I haven't touched, and directly copied the test for. It confuses me that the related `test_env_info_displays_complete_info` test does not fail.

You mock `sys.executable` in the new test. Your mock uses forward slashes on Windows. `test_env_info_displays_complete_info` just uses the real `sys.executable`, which has backslashes.

> There are now different failing tests, which seem to be unrelated to the changes here. Any thoughts on how I should proceed?

Just wait and re-run. That was https://github.com/actions/runner-images/issues/13341

PR #10596, opened 2025-10-24
fix: respect local config file in `poetry python` commands

finswimmer:
https://github.com/python-poetry/poetry/pull/10595 should be merged first, otherwise we are unable to set the relevant configs anyway.

# Pull Request Check List

Resolves: #issue-number-here

## Summary by Sourcery

Enhancements:
- Add PoetryPythonPathProvider.base_installation_dir to load and merge local poetry.toml into the config for determining python installation paths

### DIFF
--- a/src/poetry/utils/env/python/providers.py
+++ b/src/poetry/utils/env/python/providers.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
+import contextlib
 import dataclasses
 import shutil
 import sysconfig
@@ -12,6 +13,7 @@ import findpython
 from findpython.providers.path import PathProvider
 
 from poetry.config.config import Config
+from poetry.toml import TOMLFile
 from poetry.utils._compat import WINDOWS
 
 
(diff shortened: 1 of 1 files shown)


src/poetry/utils/env/python/providers.py
```diff
@@ -41,9 +43,25 @@ def find_python_by_name(cls, name: str) -> findpython.PythonVersion | None:
...
 class PoetryPythonPathProvider(PathProvider):  # type: ignore[misc]
+    @classmethod
+    def base_installation_dir(cls) -> Path:
+        from poetry.factory import Factory
+
+        config = Config.create()
+
+        # Let's check if there is a local config file.
+        with contextlib.suppress(RuntimeError):
+            pyproject = Factory.locate()
```
radoering, line 54: I believe, we have to pass the project directory to `base_installation_dir()` and call `Factory.locate()` with it. Otherwise, `--project` and `--directory` will not work. In `Command` classes, we can call  `self.get_application().project_directory`. Then, we will probably have to pass it to all relevant methods.

radoering (changes requested):

src/poetry/utils/env/python/providers.py
finswimmer: Good catch :+1: I forgot these command line arguments. Looks like it is not trivial to pass it to all relevant methods :thinking: I will have a closer look at it the next days.

Issue #10602, opened 2025-10-29
Poetry crashes when using poetry publish --skip-existing with Nexus OSS


kisev:
### Description

I use Poetry with Nexus OSS 3.75.1-01 and the --skip-existing flag doesn't work as expected.

Currently supported return codes and error messages are - [here](https://github.com/python-poetry/poetry/blob/6f31f56c3ecb3357d75cd58496e215a6a67e79bc/src/poetry/publishing/uploader.py#L327)

I'm not sure which Nexus OSS version introduced the new error message, but in my setup I get `400 cannot be updated` instead of the expected `400 updating asset`.

In my opinion, Poetry should support the new error message as well (without changing the old behavior).

### Workarounds

- Find the Nexus OSS version that introduced the new error message and downgrade to it (not possible in my case)
- Implement a check before running `poetry publish` for Nexus OSS PyPi repositories with immutable settings enabled

### Poetry Installation Method

pip

### Operating System

Ubuntu 20.04

### Poetry Version

2.2.1

### Poetry Configuration

```bash session
cache-dir = "/home/kisev/.cache/pypoetry"
data-dir = "/home/kisev/.local/share/pypoetry"
installer.max-workers = null
installer.no-binary = null
installer.only-binary = null
installer.parallel = true
installer.re-resolve = true
keyring.enabled = true
python.installation-dir = "{data-dir}/python"  # /home/kisev/.local/share/pypoetry/python
repositories.pypi-proxy.url = "https://***/repository/pypi-proxy/simple"
requests.max-retries = 0
solver.lazy-wheel = true
system-git-client = false
virtualenvs.create = true
virtualenvs.in-project = null
virtualenvs.options.always-copy = false
virtualenvs.options.no-pip = false
virtualenvs.options.system-site-packages = false
virtualenvs.path = "{cache-dir}/virtualenvs"  # /home/kisev/.cache/pypoetry/virtualenvs
virtualenvs.prompt = "{project_name}-py{python_version}"
virtualenvs.use-poetry-python = false
```

### Example pyproject.toml

```TOML

```

### Poetry Runtime Logs

<details>
  <summary>poetry-runtime.log</summary>
  

  ```
# Private data replaced by "***"
$ poetry publish --repository nexus --username "***" --password "***" --skip-existing -vvv
Adding repository pypi-proxy (https://***/repository/pypi-proxy/simple) and setting it as primary
Deactivating the PyPI repository

Publishing *** (0.4.3) to nexus
[...]

PR #10603, opened 2025-10-29
Support new Nexus OSS versions for --skip-existing argument logic

kisev:
# Pull Request Check List

Resolves: #10602

## Summary by Sourcery

Enhancements:
- Treat HTTP 400 responses containing "cannot be updated" as file-exists errors


radoering (changes requested):
Can you add a test with the full text here, please?

https://github.com/python-poetry/poetry/blob/477b54b26167933f636d54142d2e5cc91284ee8d/tests/publishing/test_uploader.py#L125-L129

PR #10623, opened 2025-11-17
feat: Add default-group-optionality configuration option

### LINKED ISSUE #10550: Allow setting default optionality for dependency groups
### Issue Kind

Brand new capability

### Description

Now that PEP 735 `[dependency-groups]` is supported, the `[tool.poetry.group.GROUPNAME]` tables can largely be omitted from `pyproject.toml`. However, it appears that in Poetry 2.2.0 they continue to have the same behavior as in previous versions, defaulting to `optional = false`. This is a reasonable choice, but given the ease by which (e.g.) a single group can be defined and requested for installation via command line, package developers may wish to declare most or all groups optional.

To accomplish this, propose defining a Boolean `tool.poetry.default-group-optionality` key in the configuration, itself having a default `false` value. This would expose and allow configuration of the default value for `optional` for all groups.

Alternatively, perhaps this key could be under `tool.poetry.group` somehow, but it is not immediately clear how it would be differentiated from a group name. Additionally, there may well be a clearer key name.

### Impact

This would allow package developers to fully omit the `[tool.poetry.group.GROUPNAME]` tables from their `pyproject.toml` config. For configs with a large number of optional dependency groups, changing the default optionality could eliminate substantial extra complexity.

For comparison, I believe that `pip` only installs specifically-requested groups (see https://pip.pypa.io/en/latest/user_guide/#dependency-groups).

This change should be backwards compatible.

### Workarounds

Specify the following for each optional group:

```toml
[tool.poetry.group.GROUPNAME]
optional = true
```

3015pavan:
This commit implements a new configuration option `default-group-optionality` that allows users to treat all dependency groups as optional by default.

When set to `true`, Poetry will only install the implicit `main` group by default, and other groups must be explicitly requested using `--with` or `--only` flags. This eliminates the need to mark each group as `optional = true` in pyproject.toml for projects with many optional dependency groups.

Changes:
- Add `default-group-optionality` boolean config option (default: false)
- Modify GroupCommand.non_optional_groups to respect the configuration
- Add comprehensive tests for the new functionality
- Update documentation with usage examples

Fixes #10550

# Pull Request Check List

Resolves: #issue-number-here

## Summary by Sourcery

Add a new `default-group-optionality` configuration option to allow treating all dependency groups as optional by default and update related command logic, documentation, and tests.

New Features:
- Introduce `default-group-optionality` boolean config option to toggle default optionality of dependency groups

Enhancements:
- Update GroupCommand to respect the `default-group-optionality` setting when determining which groups to install

Documentation:
- Document the `default-group-optionality` option in the configuration guide with usage examples

Tests:
- Add comprehensive tests for the new `default-group-optionality` behavior, including install scenarios and config commands

### DIFF
--- a/src/poetry/console/commands/group_command.py
+++ b/src/poetry/console/commands/group_command.py
@@ -46,6 +46,10 @@ class GroupCommand(Command):
     @property
     def non_optional_groups(self) -> set[str]:
         # TODO: this should move into poetry-core
+        default_optional = self.poetry.config.get("default-group-optionality", False)
+        if default_optional:
+            # When default-group-optionality is True, all groups are optional
+            return set()
         return {
             group.name
             for group in self.poetry.package._dependency_groups.values()
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -611,3 +611,38 @@ for more information.
 Enable the system keyring for storing credentials.
 See [Repositories - Configuring credentials]({{< relref "repositories#configuring-credentials" >}})
 for more information.
+
+### `default-group-optionality`
+
+**Type**: `boolean`
+
+**Default**: `false`
+
+**Environment Variable**: `POETRY_DEFAULT_GROUP_OPTIONALITY`
+
+*Introduced in 2.3.0*
+
+When set to `true`, all dependency groups (except the implicit `main` group) are treated as optional by default.
+This means that `poetry install` will only install the main dependencies unless groups are explicitly requested
+with `--with` or `--only` options.
+
+This setting allows developers to omit the `optional = true` declaration from `pyproject.toml` for most
+dependency groups, reducing configuration verbosity when working with projects that have many optional groups.
+
+{{% note %}}
+This is a user-specific configuration and does not affect the `pyproject.toml` file or how dependencies
+are resolved. It only changes which groups are installed by default.
+{{% /note %}}
+
+**Example usage:**
+
+```bash
+# Enable default-group-optionality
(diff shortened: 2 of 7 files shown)


3015pavan:
Hey! I saw issue #10550 and thought this would be useful. Let me know if you'd like any changes! 😊

src/poetry/console/commands/group_command.py
```diff
@@ -46,6 +46,10 @@ def _group_dependency_options() -> list[Option]:
     @property
     def non_optional_groups(self) -> set[str]:
         # TODO: this should move into poetry-core
+        default_optional = self.poetry.config.get("default-group-optionality", False)
+        if default_optional:
+            # When default-group-optionality is True, all groups are optional
```
radoering, line 51: That should not be true. What if a group is explicitly set to non-optional?

I think this is the wrong place to apply such a setting. It had to be applied [here](https://github.com/python-poetry/poetry-core/blob/2548a4c47e172d298c2c34d0e511e5c589397de5/src/poetry/core/factory.py#L474-L477) and [here](https://github.com/python-poetry/poetry-core/blob/2548a4c47e172d298c2c34d0e511e5c589397de5/src/poetry/core/factory.py#L494-L497), in other words, in `poetry-core`.

However, since I am still not sure if we want to have such a feature I requested the opinion of another maintainer in https://github.com/python-poetry/poetry/issues/10550#issuecomment-3732348589.

PR #10648, opened 2025-12-09
Fix: Improve error message for invalid template variables in virtualenvs.prompt

### LINKED ISSUE #10635: Error message for incorrect configs is unclear
### Issue Kind

Change in current behaviour

### Description

The messages for errors in the global configuration could be somewhat clearer (and reported immediately as such).

## What I did
I changed the `prompt` in my `~/.config/pypoetry/config.toml` to a malformed value
```
[virtualenvs]
prompt = "{project_name} [{python-version}]"
```
(sic, it should say `python_version` but I did not notice that).

## What I got
Any operation resulting in dealing with a new venv (`poetry init`, `poetry install`, `poetry env use python3.13`...) would suddenly fail, with nothing but an empty line and a message
```
"python-version"
```
in red letters. Deleting and reinstalling `poetry` via `curl`, with `pipx` or any other way did not help. However, other active venvs continued working fine. After finally finding the root cause and changing to `python_version`, everything works again and the prompt is indeed changed according to my preference.

## What I would expect
A clearer message stating the origin of the problem in the config file (like `Undefined template variable "python-version" in "virtualenvs.prompt" read from "~/.config/pypoetry/config.toml"`) , or at least something saying that the error comes from the global configuration, and not from Python or local `pyproject.toml`. Ideally this should come as soon as possible after the wrong config is saved.

What made it particularly bad was that the time between me changing it and the first symptoms was fairly long. I did not create any new venvs inbetween, and old ones continued working without problems even with an incorrect config. Only when I recreated them during unrelated troubleshooting the problem surfaced. I suppose that it does not affect venvs that already existed at the time the wrong config was introduced, only new ones.

### Impact

I spent over two hours trying to find a trivial configuration error, and a long time after I introduced it. No amount of searching or consulting with a LLM would help as there was no error message.

### Workarounds

Probably no workarounds, just pay very close attention to what you are doing.

JaviMaligno:
Fixes #10635
Closes #10649

## Summary by Sourcery

Improve validation and error reporting for virtual environment prompt templates when creating virtualenvs.

Bug Fixes:
- Raise a clear PoetryRuntimeError when the virtualenvs.prompt template contains unknown variables instead of failing with a generic KeyError.

Tests:
- Add a regression test verifying that an invalid virtualenvs.prompt template variable produces a descriptive error message including the invalid name and config key.

### DIFF
--- a/tests/utils/env/test_env_manager.py
+++ b/tests/utils/env/test_env_manager.py
@@ -1307,14 +1307,9 @@ def test_generate_env_name_uses_real_path(
 
 
 def test_create_venv_invalid_prompt_template_variable(
-    manager: EnvManager,
-    poetry: Poetry,
-    config: Config,
-    mocked_python_register: MockedPythonRegister,
+    manager: EnvManager, poetry: Poetry, config: Config
 ) -> None:
     config.merge({"virtualenvs": {"prompt": "{project_name}-{invalid_var}"}})
-
-    mocked_python_register("3.9.0", make_system=True)
 
     with pytest.raises(PoetryConsoleError) as exc_info:
         manager.create_venv()
@@ -1326,14 +1321,9 @@ def test_create_venv_invalid_prompt_template_variable(
 
 
 def test_create_venv_malformed_prompt_template(
-    manager: EnvManager,
-    poetry: Poetry,
-    config: Config,
-    mocked_python_register: MockedPythonRegister,
+    manager: EnvManager, poetry: Poetry, config: Config
 ) -> None:
     config.merge({"virtualenvs": {"prompt": "{project_name"}})  # Missing closing brace
-
-    mocked_python_register("3.9.0", make_system=True)
 
     with pytest.raises(PoetryConsoleError) as exc_info:
         manager.create_venv()


aryanyk:
Hey @JaviMaligno! 👋  
Thanks for jumping in.

I’ve already opened a PR for this issue and am working through the review feedback right now. To avoid overlapping work, it might be easier to tackle another open issue unless the maintainers prefer otherwise.

Appreciate the enthusiasm!

src/poetry/utils/env/env_manager.py
```diff
@@ -441,10 +443,20 @@ def create_venv(
...
-                project_name=self._poetry.package.name or "virtualenv",
-                python_version=python.minor_version.to_string(),
-            )
+            try:
+                venv_prompt = venv_prompt.format(
+                    project_name=self._poetry.package.name or "virtualenv",
+                    python_version=python.minor_version.to_string(),
+                )
+            except KeyError as e:
+                raise PoetryRuntimeError(
```
radoering: The error messages are not displayed in red. Maybe, just use another exception type.

tests/utils/env/test_env_manager.py
```diff
@@ -1302,3 +1306,38 @@ def test_generate_env_name_uses_real_path(
     venv_name1 = EnvManager.generate_env_name("simple-project", "the_real_dir")
     venv_name2 = EnvManager.generate_env_name("simple-project", "linked_dir")
     assert venv_name1 == venv_name2
+
+
+def test_create_venv_invalid_prompt_template_variable(
+    manager: EnvManager,
+    poetry: Poetry,
+    config: Config,
+    mocker: MockerFixture,
```
radoering: `mocker` seems to be unused.

tests/utils/env/test_env_manager.py
```diff
@@ -1302,3 +1306,38 @@ def test_generate_env_name_uses_real_path(
...
+    mocked_python_register: MockedPythonRegister,
+) -> None:
+    config.merge({"virtualenvs": {"prompt": "{project_name}-{invalid_var}"}})
+
+    mocked_python_register("3.9.0", make_system=True)
```
radoering: We do not need `mocked_python_register,` do we?

JaviMaligno:
Thanks for the review feedback! I've addressed all the comments:

**Changes in this commit:**

1. **Exception type changed**: Replaced `PoetryRuntimeError` with `PoetryConsoleError` so error messages now display in red (as `PoetryConsoleError` is handled by cleo's standard exception handler).

2. **Removed unused test parameters**: Removed the unused `mocker: MockerFixture` parameter from both test functions.

3. **Pre-commit fixes**: Ran ruff which fixed 5 linting errors (removed unused imports: `Config`, `NullIO`, `virtualenv`) and reformatted the code.

Let me know if there's anything else that needs to be addressed!

PR #10649, opened 2025-12-09
Improve error message for invalid virtualenvs.prompt config

aryanyk:
# Pull Request Check List

Resolves: #Resolves: #10635

### Summary
Improve error message when an invalid template variable is used in
`virtualenvs.prompt` configuration.

### Details
Previously an invalid placeholder such as `{python-version}` resulted in
a cryptic KeyError. This change catches the error and raises a clearer
PoetryError explaining the config source and how to fix it.

### Testing
- Reproduced and validated locally using WSL (Ubuntu)

error_regen

image

## Summary by Sourcery

Bug Fixes:
- Prevent cryptic KeyError when virtualenv prompt configuration contains an unknown template variable by raising a descriptive PoetryError instead.

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -2,24 +2,23 @@
 
 [[package]]
 name = "anyio"
-version = "4.10.0"
+version = "4.12.0"
 description = "High-level concurrency and networking framework on top of asyncio or Trio"
 optional = false
 python-versions = ">=3.9"
 groups = ["main"]
 files = [
-    {file = "anyio-4.10.0-py3-none-any.whl", hash = "sha256:60e474ac86736bbfd6f210f7a61218939c318f43f9972497381f1c5e930ed3d1"},
-    {file = "anyio-4.10.0.tar.gz", hash = "sha256:3f3fae35c96039744587aa5b8371e7e8e603c0702999535961dd336026973ba6"},
+    {file = "anyio-4.12.0-py3-none-any.whl", hash = "sha256:dad2376a628f98eeca4881fc56cd06affd18f659b17a747d3ff0307ced94b1bb"},
+    {file = "anyio-4.12.0.tar.gz", hash = "sha256:73c693b567b0c55130c104d0b43a9baf3aa6a31fc6110116509f27bf75e21ec0"},
 ]
 
 [package.dependencies]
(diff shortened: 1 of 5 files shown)


aryanyk:
Thanks for the feedback! Makes sense — I’ll update the error message to include the supported template variables and reference the concrete config source so it’s easier for users to fix. Will push an update shortly.

aryanyk:
I’ve updated the error message to include the supported template variables and more concrete config hints, and applied Ruff formatting. Let me know if you’d like the wording tweaked further.

aryanyk:
Hi @radoering ,

I’ve pushed the requested changes and all CI checks are green now ✅  
Could you please review the PR when you get a chance?

Thanks for your time and guidance!

src/poetry/utils/env/env_manager.py
```diff
@@ -441,10 +442,20 @@ def create_venv(
...
-            )
+            try:
+                venv_prompt = venv_prompt.format(
+                    project_name=self._poetry.package.name or "virtualenv",
+                    python_version=python.minor_version.to_string(),
+                )
+            except KeyError as e:
+                raise PoetryError(
+                    f'Invalid template variable "{e.args[0]}" in '
+                    'global configuration option "virtualenvs.prompt".\n\n'
```
radoering, line 453: It is not necessarily in the global configuration. It can also be in the [local configuration](https://python-poetry.org/docs/configuration/#local-configuration).

src/poetry/utils/env/env_manager.py
```diff
@@ -441,10 +442,20 @@ def create_venv(
...
+                    "Supported variables are: project_name, python_version.\n"
+                    "You can update this setting via:\n"
+                    "  - ~/.config/pypoetry/config.toml\n"
```
radoering, line 456: That is only the path for Linux and only the global configuration. It probably makes more sense to recommend setting it via `poetry config [--local] virtualenvs.prompt <value>`.

radoering (commented):
Unfortunately, someone else created a similar PR shortly before you: #10648.

Since that PR covers an additional error, adds tests and was the first one, I tend to prefer it - if the author addresses the review feedback.

Thank you anyway.

PR #10656, opened 2025-12-11
tests: make test independent of git setting of default branch

radoering:
## Summary by Sourcery

Ensure git backend tests use a repository with a fixed 'main' default branch and update related expectations to be independent of external git default-branch settings.

Tests:
- Initialize temporary git test repositories with 'main' as the default branch to decouple tests from environment-specific git settings.
- Adjust clone error expectation to reference 'refs/heads/main' instead of 'refs/heads/master' in git backend tests.


radoering:
pre-commit.ci autofix

Issue #6742, opened 2022-10-08
Poetry self add command does not work with directories


wieczyk:
According to `poetry self add --help` it should work with directories, but it does not.

I want to split my package into many packages (let us say I am going to use monorepo approach) and wanted to develop a couple of plugins for poetry. I wanted to store those plugins inside the repository and develop them together with my other packages.
Unfortunately when I am doing `poetry self add ../mypoetryplugin` I see an error. It looks like poetry cannot resolve package. Unfortunately `-vvv` is not informative, it just prints stacktrace, but no additional information about resolution, anything I could follow and analyze to understand the issue.

- Operating system: Ubuntu 20.10
- Poetry debug output:

```
Poetry
Version: 1.2.1
Python:  3.10.6

Virtualenv
Python:         3.10.6
Implementation: CPython
Path:           NA
Executable:     NA

System
Platform:   linux
OS:         posix
Python:     3.10.6
Path:       /home/wieczyk/.pyenv/versions/3.10.6
Executable: /home/wieczyk/.pyenv/versions/3.10.6/bin/python3.10
```

Reproduction is quite simple, we can use any public poetry plugin to reproduce issue.

```
$ git clone https://github.com/tiangolo/poetry-version-plugin.git
$ poetry new myproject
$ cd myproject
$ poetry self add ../poetry-version-plugin

Could not find a matching version of package ../poetry-version-plugin/
```

neersighted:
Duplicate #6529 -- PRs welcome :smile:

wieczyk:
One question, I think more things is wrong or I do not understand something.

I created plugin and installed it in my system. It is visible via `pip list`. But then when I am doing `poetry self add mypluginpkg` it still throws the same error.

neersighted:
`pip` and `poetry` are both manipulating the list of packages installed into Poetry's environment in this case. `pip install` doesn't make a plugin visible to `poetry self add`, it is a substitute/semi-equivalent command.

Issue #6529, opened 2022-09-15
Installing plugin as dist or local path into Poetry's environment


jo014670:
- [x] I am on the [latest](https://github.com/python-poetry/poetry/releases/latest) Poetry version.
- [x] I have searched the [issues](https://github.com/python-poetry/poetry/issues) of this repo and believe that this is not a duplicate.
- [x] If an exception occurs when executing a command, I executed it again in debug mode (`-vvv` option).

- **OS version and name**: Windows 10
- **Poetry version**: 1.2.0
- **Link of a [Gist](https://gist.github.com/) with the contents of your pyproject.toml file**: N/A

## Issue
There does not appear to be a way to install a custom plugin locally without involving PyPi. I've tried the following commands:

Within the plugin directory:
```
poetry self add <plugin name>
```

Result:
```
'NoneType' object has no attribute 'startswith'
```

Within the dist directory after building:
```
poetry self add <wheel file>
```

Result:
```
Could not parse version constraint: <wheel file>
```

I have been able to work around this by using the system `pip` to install the plugin, but this is an ugly workaround for an issue that makes testing incredibly difficult.

neersighted:
Currently `poetry self add` is designed solely for convenience/the end user. For development, I would suggest using the copy of `pip` bundled in Poetry's environment.

e.g.

```sh
~/.local/share/pypoetry/venv/bin/pip install <path to wheel>
```

This could easily become a docs ask -- I don't think we'll change how `poetry self add` requires a constraint as it is fundamentally based on regular `poetry add` and manages an internal pyproject.toml (so that one can configure plugins as part of their dotfiles and have Poetry install them in one command).

jo014670:
This is what I do:

```
<user directory>/AppData/Roaming/pypoetry/venv/Scripts/pip.exe install <path to wheel>
```

Which succeeds, but:

```
poetry show <plugin>
```

Result:
```
Package <plugin> not found
```

neersighted:
You should be using `poetry self show` to introspect Poetry's internal environment

jo014670:
Yes, good point. I do that, and:

```
Error: poetry.lock not found. Run `poetry lock` to create it.
```

So I run `poetry self lock` with success. However:

```
> poetry self show <plugin name>
Package <plugin name> not found
```

Also worth noting is using the workaround approach, the custom commands show up in `poetry list`, however using Poetry's internal environment, it does not.

neersighted:
[...]

PR #10661, opened 2025-12-16
Add support for processing .pth files in project plugins

BlackGad:
Handle .pth files for plugins in editable mode.

# Pull Request Check List

Related to: #6742, #6529

## Summary by Sourcery

Handle plugin .pth files when loading plugins so editable/develop-mode plugins have their dependencies available.

New Features:
- Process .pth files for plugin distributions when loading plugin entry points to make editable-mode plugin dependencies available.

Documentation:
- Document support for installing plugins in editable mode using path dependencies with develop = true, including an example in the plugin configuration docs.

Tests:
- Add tests covering plugin loading behavior when a .pth file is present, absent, or when the distribution has no files.

### DIFF
--- a/src/poetry/plugins/plugin_manager.py
+++ b/src/poetry/plugins/plugin_manager.py
@@ -63,7 +63,7 @@ class PluginManager:
 
         plugin_path = pyproject_toml.parent / ProjectPluginCache.PATH
         if plugin_path.exists():
-            # insert at the beginning to allow overriding with (compatible) versions
+            # insert at the beginning to allow overriding dependencies
             EnvManager.get_system_env(naive=True).sys_path.insert(0, str(plugin_path))
             # process .pth files (among other things)
             addsitedir(str(plugin_path))


BlackGad:
@radoering Hi! Can you make review of this PR?

BlackGad:
Any update on this?

docs/plugins.md
```diff
@@ -263,13 +263,16 @@ in the `tool.poetry.requires-plugins` section of the pyproject.toml file:
 ```toml
 [tool.poetry.requires-plugins]
 my-application-plugin = ">1.0"
+custom-plugin = {path = "custom_plugin/src", develop = true}
```
radoering:
```suggestion
custom-plugin = {path = "custom_plugin", develop = true}
```

Normally, you use the project path, no matter if the project uses source layout or flat layout. I tried with poetry-plugin-export and it seems to work with src.

src/poetry/plugins/plugin_manager.py
```diff
@@ -89,9 +90,27 @@ def _add_plugin(self, plugin: Plugin) -> None:
...
+        return pth_files
+
     def _load_plugin_entry_point(self, ep: metadata.EntryPoint) -> None:
         logger.debug("Loading the %s plugin", ep.name)
 
+        # In case the plugin installed in editable/develop mode, we need to
+        # process its .pth files to ensure that its dependencies are available.
+        for pth_path in self._get_pth_files(ep.dist):
+            logger.debug("Processing .pth file: %s", pth_path)
+            addpackage(str(pth_path.parent), pth_path.name, None)
```
radoering: I am wondering whether we really should use `addpackage()` because it seems to be undocumented: It is not mentioned on https://docs.python.org/3/library/site.html.

A documented method that uses `addpackage()` is [addsitedir()](https://docs.python.org/3/library/site.html#site.addsitedir). Have you considered using it [here](https://github.com/python-poetry/poetry/blob/e8a05b7aea2bbf4a06f4e4886dd5353ff5f13a1f/src/poetry/plugins/plugin_manager.py#L65) instead of using `addpackage()`?

The reason why we did not use it is given in https://github.com/python-poetry/poetry/pull/9547#discussion_r1778673450:

> `site.addsitedir` appends, but we want to insert it at the beginning. (At least if we want to allow overwrites. See other comment.)
> 
> At first, I additionally called `site.addsitedir()` but I removed it because it breaks many tests and I did not find a use case where it is necessary. (That is just a testing issue, but it is annoying anyway. That is also why we do not use `sys.path` but `Env.sys_path`. I think, in production it does not make a difference but it is relevant for tests.)

Since we have a use case now, this should be reconsidered. I did a quick test: Fixing the testing issue seems to be as simple as adding the following fixture to `test_plugin_manager.py`: 

```
@pytest.fixture(autouse=True)
def mock_sys_path(mocker: MockerFixture) -> None:
    sys_path_copy = sys.path.copy()
    mocker.patch("poetry.plugins.plugin_manager.sys.path", new=sys_path_copy)
```

What do you think?

docs/plugins.md
BlackGad: I usually place pyproject.toml files under `src` folder.

src/poetry/plugins/plugin_manager.py
BlackGad: Good point.

I originally used `addpackage()` because it felt more exact for this case, and since not very familiar with the whole Poetry codebase, I tried to keep the change as small as possible.

Your suggestion to use `addsitedir()` makes sense.

I have updated the code to use `addsitedir()` and adjusted the unit tests. Thanks for the help!

Odechree:
I do not have permission to be able to execute this poll request or I would even though I know what it implies so how do you want me to go about it?

BlackGad:
@radoering Thank you!

Issue #10217, opened 2025-02-23
Make `poetry init` not add readme section if it doesn't detect a readme


iTrooz:
### Issue Kind

Change in current behaviour

### Description

When creating a project using `poetry init`, poetry always generates a `readme` key, even if there is no readme file in the repository. That then makes `poetry install` fail, forcing the user to either create a readme or delete the key.

I think it would be better if poetry detected if the file existed, and only added the `readme` key if the README.md file actually exists

### Impact

Minimal impact ?

### Workarounds

removing `readme` key after doing poetry ini( which is what I do now)

ammargit93:
I’m planning to work on this and submit a PR. Commenting here to avoid duplicate work. @python-poetry/triage
The plan is to update `poetry init` to only add the readme field if the README file actually exists, and add a regression test to ensure `poetry install` doesn’t fail.

PR #10679, opened 2026-01-01
init: only add readme key if README file exists

ammargit93:
### Problem
When running `poetry init`, the generated `pyproject.toml` always includes a `readme` key, even if there is no README file in the project.  
This can cause `poetry install` to fail, forcing the user to either create a README file or manually remove the key.

### Solution
Check if the README file exists before adding the `readme` key in `pyproject.toml`.  
- If `README.md` (or the specified format) exists, the `readme` key is added.  
- If not, the key is omitted.

This change fixes [Issue #10217](https://github.com/python-poetry/poetry/issues/10217).

### Impact
Minimal: only affects newly created projects with `poetry init`.  
Does not change behavior for existing projects or other keys.

### Testing
Added unit tests in `tests/layouts/test_layout.py`:
- `test_readme_key_present_if_readme_exists`
- `test_readme_key_removed_if_readme_missing`

All tests pass locally on Windows.

Screenshot 2026-01-01 183427

### Workarounds
Before this fix, users had to manually remove the `readme` key after running `poetry init`.

## Summary by Sourcery

Conditionally include the `readme` entry in newly generated `pyproject.toml` files based on the presence of a README file in the project directory.

Bug Fixes:
- Prevent `poetry init` from generating a `readme` key when no README file exists, avoiding subsequent installation failures.

Tests:
- Add layout tests to verify the `readme` key is omitted when the README file is missing and present when it exists.

### DIFF
--- a/tests/console/commands/test_init.py
+++ b/tests/console/commands/test_init.py
@@ -1214,4 +1214,4 @@ def test_init_does_not_add_readme_key_when_readme_missing(
     tester.execute(interactive=False)
     # Assert
     pyproject = (tmp_path / "pyproject.toml").read_text(encoding="utf-8")
-    assert 'readme = "README.md"' not in pyproject
+    assert "readme =" not in pyproject


ammargit93:
Hi
I wanted to note that after implementing the change to conditionally add the readme key only if a README file exists, I ran into a large number of test failures. Even though the code change itself is small, many existing tests—especially in `tests/console/commands/test_init.py` and `tests/console/commands/test_new.py`—assume that the readme key always exists after `poetry init`.

After modifying the `test_init.py` i reduced the failed tests from 34 to 15, any help on this matter would be greatly appreciated.

radoering:
> Even though the code change itself is small, many existing tests—especially in `tests/console/commands/test_init.py` and `tests/console/commands/test_new.py`—assume that the readme key always exists after `poetry init`.

It makes sense that (many) tests in `test_init.py` have to be adapted. If there is no such test yet, a test with an existing readme should be added so that a project.toml with an readme entry is still created in at least one test.

As far as I know, `poetry new` creates a readme file so that the readme entry in the pyproject.toml should still be created. If the check for an existing readme happens before the readme file is created and thus the readme entry in the pyproject.toml file is not created that would be a bug that should be fixed. If this still works, it might just be an issue of the tests.

ammargit93:
Hello @radoering, really sorry for the messy commit history. The important changes in this PR are:

- `4d5e1de` — updated and wrote tests to respect updated init behavior
- `fb66c1d` — apply pre-commit fixes

Regarding the code:  
- I considered that `poetry new` creates a README file, so the `readme` entry in `pyproject.toml` should still be created.  
- Introduced a `command_name` parameter in `generate_project_content(self, command_name: str | None = None)` to check if the command is `'new'` or `'init'`.

The other commits are mainly minor formatting/linting adjustments from pre-commit.

Environment:
Windows 11

Could you please review this PR? Thanks!

src/poetry/layouts/layout.py
```diff
@@ -160,7 +159,14 @@ def generate_project_content(self) -> TOMLDocument:
...
 
-        project_content["readme"] = f"README.{self._readme_format}"
+        readme_path = self.basedir / f"README.{self._readme_format}"
+        if command_name == "new":
+            project_content["readme"] = f"README.{self._readme_format}"
+        elif command_name == "init":
+            if readme_path.exists():
+                project_content["readme"] = f"README.{self._readme_format}"
+            else:
+                project_content.remove("readme")
```
radoering: I wondered why the distinction between `init` and `new` is necessary because actually the readme has already been created by `poetry new` at this point. It is necessary because `readme_path` is relative to the project directory, which is not necessarily the current working directory. It normally is the current working directory for `poetry init` but normally not for `poetry new`. However, you can also do `poetry init --project subdirectory`. Then, this approach probably fails.

Long story short: I think the distinction on `command_name` is not sufficient and should be replaced by passing the `path`, building the absolute `readme_path` and checking this for existence. Then, there should be no need to distinguish between `init` and `new`.

tests/console/commands/test_init.py
```diff
@@ -64,8 +64,13 @@ def tester(patches: None) -> CommandTester:
 def test_basic_interactive(
     tester: CommandTester, init_basic_inputs: str, init_basic_toml: str
 ) -> None:
+    # Remove the readme line
+    lines = init_basic_toml.splitlines()
+    lines = [line for line in lines if not line.strip().startswith("readme =")]
+    init_basic_toml_no_readme = "\n".join(lines)
```
radoering: Can you extract this into a `init_basic_toml_no_readme` fixture and use it in all places where it is necessary?

tests/console/commands/test_new.py
```diff
@@ -191,7 +191,13 @@ def test_command_new_with_readme(
     poetry = verify_project_directory(path, package, Path("src") / package)
     project_section = poetry.pyproject.data["project"]
     assert isinstance(project_section, dict)
-    assert project_section["readme"] == f"README.{fmt or 'md'}"
+
+    readme_file = path / f"README.{fmt or 'md'}"
+    readme_file.touch()
```
radoering:
```suggestion
```

We want to check that the `readme_file` is created automatically, right?

.gitignore
```diff
@@ -17,3 +17,6 @@
 
 .venv
 /poetry.toml
+
+# ignore pycache dirs
+__pycache__/
```
radoering: Unrelated changes should be avoided in a PR. (Is this even necessary since we ignore *.pyc files anyway?)

radoering (changes requested):

src/poetry/layouts/layout.py
ammargit93: Thanks for your feedback, you're right, I should've resolved the paths instead of checking for commands. In this case it becomes easier to simply pass in the project path. 

this earlier logic was correct ?
```
readme_path = self.basedir / f"README.{self._readme_format}"
if readme_path.exists():
    project_content["readme"] = f"README.{self._readme_format}"
else:
    project_content.remove("readme")
```

tests/console/commands/test_init.py
ammargit93: Sure, will do

tests/console/commands/test_new.py
ammargit93: My bad regarding that, I'll update the test accordingly.

src/poetry/layouts/layout.py
radoering: I think so (except that `readme_path` is relative).

I did not check whether `"readme"` already exists in `project_content` and needs to be removed but I assume that this is the case.

.gitignore
ammargit93: Ok

Issue #10610, opened 2025-11-03
Poetry doesn't recognize the virtualenv it created; insists the virtualenv path is "."


brad-eisenhauer:
### Description

Acutal behavior:

```shell
$ poetry env info --path
.
$ poetry env remove
Deleted virtualenv: /Users/beise/dev/<redacted>/.venv
$ poetry env use python3.12
Creating virtualenv <redacted> in /Users/beise/dev/<redacted>/.venv
Using virtualenv: .
$ poetry install
The currently activated Python version 3.9.4 is not supported by the project (>=3.10,<4).
Trying to find and use a compatible version. 
Using python3.14 (3.14.0)
The virtual environment found in . seems to be broken.
Recreating virtualenv <redacted> in /Users/beise/dev/<redacted>/.venv
Installing dependencies from lock file
...
```

Expected behavior: poetry should show the virtualenv created in `.venv` as the virtual environment and install dependencies into it when running `poetry install`.

I attempted a workaround by deleting the project directory and re-cloning from the repository, but the result was the same after this operation.

### Workarounds

n/a

### Poetry Installation Method

pipx

### Operating System

Mac Sequoia 15.7.1

### Poetry Version

Poetry (version 2.2.1)

### Poetry Configuration

```bash session
cache-dir = "/Users/beise/Library/Caches/pypoetry"
data-dir = "/Users/beise/Library/Application Support/pypoetry"
installer.max-workers = null
installer.no-binary = null
installer.only-binary = null
installer.parallel = true
installer.re-resolve = true
keyring.enabled = true
python.installation-dir = "{data-dir}/python"  # /Users/beise/Library/Application Support/pypoetry/python
requests.max-retries = 0
solver.lazy-wheel = true
system-git-client = false
virtualenvs.create = true
virtualenvs.in-project = true
virtualenvs.options.always-copy = false
virtualenvs.options.no-pip = false
virtualenvs.options.system-site-packages = false
virtualenvs.path = "{cache-dir}/virtualenvs"  # /Users/beise/Library/Caches/pypoetry/virtualenvs
virtualenvs.prompt = "{project_name}-py{python_version}"
virtualenvs.use-poetry-python = false
```

### Python Sysconfig

<details>
  <summary>sysconfig.log</summary>
  

  ```
Platform: "macosx-10.9-x86_64"
Python version: "3.9"
Current installation scheme: "posix_prefix"

Paths: 
	data = "/Library/Frameworks/Python.framework/Versions/3.9"
[...]

PR #10689, opened 2026-01-14
Fix(env): Prioritize in-project venv over VIRTUAL_ENV

Rahul-IIT-B:
# Pull Request Check List

Resolves: #10610

When `virtualenvs.in-project = true`, Poetry should prioritize the local `.venv` even if `VIRTUAL_ENV` is set (e.g. when running via `pipx`). This PR modifies `EnvManager` to check for the in-project virtual environment before checking environment variables.
### Changes
- Modified `EnvManager.get()` to check `in_project_venv_exists()` earlier.
- Added regression tests in `tests/utils/env/test_env_manager_priority.py`.
### Verification
- Added regression tests pass interactions with `VIRTUAL_ENV` and `.`
- Existing tests pass.

## Summary by Sourcery

Prioritize in-project virtual environments over externally provided VIRTUAL_ENV when resolving the active environment.

Bug Fixes:
- Ensure EnvManager.get() selects an existing in-project .venv even when VIRTUAL_ENV is set, such as when Poetry is run via pipx or with VIRTUAL_ENV='.'.

Tests:
- Add regression tests to verify EnvManager prefers in-project .venv over VIRTUAL_ENV, including cases with VIRTUAL_ENV set to a separate path or to '.' and after `poetry env use`.

### DIFF
--- a/tests/utils/env/test_env_manager_priority.py
+++ b/tests/utils/env/test_env_manager_priority.py
@@ -1,9 +1,11 @@
 """
 Regression tests for Poetry issue #10610.
 """
+
 from __future__ import annotations
 
 import os
+
 from pathlib import Path
 from typing import TYPE_CHECKING
 from unittest.mock import MagicMock
@@ -16,6 +18,7 @@ from poetry.utils.env import VirtualEnv
 
 if TYPE_CHECKING:
     from pytest_mock import MockerFixture
+
     from poetry.poetry import Poetry
     from tests.conftest import Config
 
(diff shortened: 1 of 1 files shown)


rahul-singh-ai:
Hi maintainers,

I see that the CI check `Tests / Windows (Python 3.11) / pytest (pull_request)` is failing, but the failure appears to be unrelated to my changes:

**Failing test:** `test_call_does_not_block_on_full_pipe[sys.stderr]` in [tests/utils/env/test_env.py]

**My changes:**
- Modified [src/poetry/utils/env/env_manager.py] (EnvManager priority logic)
- Added [tests/utils/env/test_env_manager_priority.py] (new regression tests)

The failing test is in a completely different file and tests subprocess pipe handling, which is unrelated to the EnvManager environment selection logic I modified.

I've fixed the pre-commit formatting issues (ruff and trailing whitespace), and my new regression tests all pass successfully.

Could you please re-run the CI when you have a chance? I believe this is a flaky test failure.

Thank you!

dimbleby:
what makes you believe that this is related to #10610?

I don't see any particular reason to prefer either the environment variable or the directory, in case of conflict - but changing it is going to break anyone who has been relying on the existing behaviour

radoering:
I do not think this is a desired change. `VIRTUAL_ENV` is an activated virtual environment which has priority by design. In other words, this is not a bug.

> Poetry should prioritize the local `.venv` even if `VIRTUAL_ENV` is set (e.g. when running via `pipx`).

This does not make sense to me. As far as I know, `VIRTUAL_ENV` is not set when running a tool via pipx. At least, `.venv` is used when running poetry via pipx (and no venv has been activated before).

rahul-singh-ai:
Thank you for the feedback! After re-reading issue #10610 more carefully, I realize I misunderstood the root cause.

You're absolutely right that `VIRTUAL_ENV` should have priority when a user has explicitly activated an environment. The actual bug reported in #10610 is that `poetry env info --path` returns `.` instead of the correct `.venv` path when **no** environment is activated.

I apologize for the confusion. I'll close this PR and take another look at the issue to understand the real problem better.

Thanks for taking the time to review and explain this!

PR #10701, opened 2026-01-21
Update to dulwich 1.0.0

hroncok:
From the release notes:

> From here on, Dulwich will not break backwards compatibility until 2.0

https://github.com/jelmer/dulwich/commit/ca9de6da80e2dab04f0da4c260afc09ae24bc1c1

# Pull Request Check List

- There is no changed code.
- There is no changed code.

## Summary by Sourcery

Build:
- Broaden dulwich version constraint from <0.26.0 to <2 to support dulwich 1.0 and later 1.x releases.


hroncok:
> Given Dulwich’s backward-compatibility guarantee only starts at 1.0.0, consider bumping the lower bound to >=1.0.0

Happy to do that if the poetry maintainers want to. But it works with the previous version as well.

radoering:
I think it is fine to only bump the upper bound. Apparently, there is no breaking change that hits us between 0.25 and 1.0 and I do not expect more 0.x versions in the future.

jelmer:
> I think it is fine to only bump the upper bound. Apparently, there is no breaking change that hits us between 0.25 and 1.0 and I do not expect more 0.x versions in the future.

Indeed, there will not be any more 0.x versions with one caveat: if there are security issues or grave C Git interoperability issues (new release of C Git write repositories in a way that breaks Dulwich) I will consider releasing another 0.25.x - but also without making any backwards incompatible changes.

Issue #4117, opened 2021-05-30
Improve documentation on the behaviour of update and install


chernals:
I believe that the documentation could be improved regarding what `update` and `install` are doing.

To me it is not really clear, for example:

- when doing update are the new packages installed? is the `pyproject.toml` file update with latest version numbers?
- when should one do `install` ? Only the first time?

Maybe we could collect here answers to questions like the two above and I'll be happy to make a PR for the doc.

finswimmer:
The purpose of `poetry install` is to:

- create a venv if None exists
- resolve dependency and write the exact version to `poetry.lock` if None exists
- install the locked dependencies in the venv
- install the current project in editable mode

A `poetry install` is only necessary if you:
- create the venv the first time
- change any metadata of the project, like version number
- add "scripts" entry points to the `pyproject.toml`

The purpose of `poetry update` is to:

- update all dependencies to the current version within in the `pyproject.toml` specified version range
- write the updated version to `poetry.lock`
- the version numbers in `pyproject.toml` remain untouched

chernals:
Thanks !

How does `poetry update` behave when "extra" dependencies are declared or with respect to development dependencies? Does it follow what `poetry install` did when last executed? (for example if `poetry install --no-dev` then `poetry update` will not touch the developement dependencies, if `poetry install` (without `-E foo`0 then `poetry update` will not touch the `foo` extra dependencies?

mrkeuz:
Same issue with `poetry install `. Always confused with this command help.

Help also should explain:
- where dependencies will be installed - exactly into virtual environment (even poetry shell virtual env not activated). Not to local python dependencies. Even poetry shell / virtual env not activated.

I understand now that it assumed and expected but not clear from docs. But at first I tried to install project into system via `poetry install`. Exactly because poor documented.

rafrafek:
Are dev (or other groups) dependencies installed by default?

neersighted:
> Are dev (or other groups) dependencies installed by default?

The best intro to the feature is the [release announcement](https://python-poetry.org/blog/announcing-poetry-1.2.0/#dependency-groups) which goes into detail with examples. Yes, they are installed by default using Poetry (but not captured by tools that build/install as a Python package).

rafrafek:
> Yes, they are installed by default using Poetry (but not captured by tools that build/install as a Python package).

Can we add this information to the docs here: https://python-poetry.org/docs/cli/#install

neersighted:
[...]

PR #10713, opened 2026-02-01
docs: clarify install vs update command differences

abdullahmujahidali:
# Pull Request Check List

Resolves: #4117

## Summary
Improves documentation for `install` and `update` commands to address common user confusion.

## Changes
- Added note in `install` section explaining when to use `install` vs `update`
- Clarified that `update` also installs packages after updating the lock file
- Added note that `update` does **not** modify `pyproject.toml`

## Questions Addressed (from issue)
- When doing `update`, are new packages installed?-> Yes
- Is `pyproject.toml` updated? -> No, only `poetry.lock`
- When should one do `install`? → After cloning or for reproducible installs

## Summary by Sourcery

Clarify the behavioral differences between the `poetry install` and `poetry update` commands in the CLI documentation.

Documentation:
- Document when to use `poetry install` versus `poetry update`, emphasizing reproducible installs and lockfile creation.
- Clarify that `poetry update` both updates and installs dependencies while only modifying `poetry.lock` and not `pyproject.toml`, and direct users to `add` for changing version constraints.

### DIFF
--- a/docs/cli.md
+++ b/docs/cli.md
@@ -475,9 +475,19 @@ poetry init
 The `install` command reads the `pyproject.toml` file from the current project,
 resolves the dependencies, and installs them.
 
+```bash
+poetry install
+```
+
+If there is a `poetry.lock` file in the current directory,
+it will use the exact versions from there instead of resolving them.
+This ensures that everyone using the library will get the same versions of the dependencies.
+
+If there is no `poetry.lock` file, Poetry will create one after dependency resolution.
+
 {{% note %}}
 **When to use `install` vs `update`:**
-- Use `poetry install` to install dependencies as specified in `poetry.lock` (or resolve dependencies and create the lock file if it's missing).
+- Use `poetry install` to install dependencies as specified in `poetry.lock` (or resolve dependencies and create the lock file if it is missing).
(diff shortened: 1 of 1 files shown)


docs/cli.md
```diff
@@ -475,6 +475,14 @@ poetry init
 The `install` command reads the `pyproject.toml` file from the current project,
 resolves the dependencies, and installs them.
 
+{{% note %}}
+**When to use `install` vs `update`:**
+- Use `poetry install` to install dependencies as specified in `poetry.lock` (or resolve dependencies and create the lock file if it's missing).
+  This is what you run after cloning a project or when you want reproducible installs.
```
radoering: Especially for reproducible installs `sync` is even better than `install`.

docs/cli.md
```diff
@@ -475,6 +475,14 @@ poetry init
...
+- Use `poetry update` when you want to update dependencies to their latest versions (within constraints)
```
radoering: Which constraints? Maybe, better:

```suggestion
- Use `poetry update` when you want to update dependencies to their latest versions (within the constraints from the `pyproject.toml`)
```

Issue #10536, opened 2025-09-09
Clarify "poetry lock" suggestion provided for `poetry self` subcommands


shellydavid:
### Issue Kind

Change in current behaviour

### Description

When a lock file is outdated, the provided suggestion is to run `poetry lock`. The same suggestion applies for `poetry self` subcommands - ex `poetry self sync`:

*"pyproject.toml changed significantly since poetry.lock was last generated. Run `poetry lock` to fix the lock file"*

For `poetry self` subcommands however, the fix would be `poetry self lock`:
- `poetry lock` references `pyproject.toml` from the current project/directory. 
- `poetry self lock` references `pyproject.toml` from the Poetry installation directory:
    - macOS - `~/Library/Application\ Support/pypoetry`
    - Linux - `~/.config/pypoetry`
    - Windows - `%APPDATA%\pypoetry`

Assuming the user is likely to be in a different directory, they might see *"Poetry could not find a pyproject.toml file"* or a lock file from an unrelated project gets updated. `poetry lock` updates the expected lock file in this case only if you're in the Poetry installation directory - it does also create a virtualenv and I'm not sure if this may present other unintended consequences.

---

**Proposal**
- An outdated lock file from a regular command continues to suggest `poetry lock`
- An outdated lock file from a `poetry self` subcommand suggests `poetry self lock`

I did not find an existing issue on this, apologies in advance if I missed a duplicate issue!

### Impact

Could reduce potential confusion/errors for the user

### Workarounds

- Running `poetry self lock`
- Running `poetry lock` as suggested, but specifically in the Poetry installation configuration directory
    - macOS - `~/Library/Application\ Support/pypoetry`
    - Linux - `~/.config/pypoetry`
    - Windows - `%APPDATA%\pypoetry`

gabriel-trigo:
@radoering created PR

LouisLau-art:
Opened PR: python-poetry/poetry#10715

This updates the outdated-lock suggestion for `poetry self ...` commands to recommend `poetry self lock` (instead of `poetry lock`).

PR #10715, opened 2026-02-06
fix: suggest `poetry self lock` for `self` commands

LouisLau-art:
Fixes python-poetry/poetry#10536.

When an install/sync operation is run via `poetry self ...`, the outdated-lock message currently suggests `poetry lock`, which points at the current project rather than Poetry’s system project.

This changes the suggestion to `poetry self lock` when the installer is operating on Poetry’s system project.

Tests:
- `python -m pytest -o addopts= tests/installation/test_installer.py::test_not_fresh_lock_self_project`

## Summary by Sourcery

Adjust installer lock-file guidance to suggest the appropriate command when working on Poetry’s own system project.

Bug Fixes:
- Ensure outdated-lock error messages for Poetry’s system project instruct users to run `poetry self lock` instead of `poetry lock`.

Tests:
- Add coverage verifying that installer errors for the self project recommend `poetry self lock` when the lock file is not fresh.

### DIFF
--- a/src/poetry/installation/installer.py
+++ b/src/poetry/installation/installer.py
@@ -98,14 +98,6 @@ class Installer:
 
         return self
 
-    def _lock_fix_command(self) -> str:
-        # `poetry self` commands operate on Poetry's own system project. When the lock
-        # file is outdated, users should run `poetry self lock` rather than `poetry lock`.
-        if self._package.name == POETRY_SYSTEM_PROJECT_NAME:
-            return "poetry self lock"
-
-        return "poetry lock"
-
     def run(self) -> int:
         # Check if refresh
         if not self._update and self._lock and self._locker.is_locked():
@@ -388,6 +380,14 @@ class Installer:
             self._write_lock_file(solved_packages)
 
         return status
+
+    def _lock_fix_command(self) -> str:
+        # `poetry self` commands operate on Poetry's own system project. When the lock
+        # file is outdated, users should run `poetry self lock` rather than `poetry lock`.
+        if self._package.name == POETRY_SYSTEM_PROJECT_NAME:
+            return "poetry self lock"
+
+        return "poetry lock"
 
     def _write_lock_file(
         self,
--- a/src/poetry/console/commands/show.py
+++ b/src/poetry/console/commands/show.py
@@ -93,12 +93,6 @@ lists all packages available."""
 
     colors: ClassVar[list[str]] = ["cyan", "yellow", "green", "magenta", "blue"]
 
-    def _lock_create_command(self) -> str:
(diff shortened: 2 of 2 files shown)


LouisLau-art:
Good point — updated to avoid the magic string by centralizing the system project name as a constant on `Installer`.

radoering:
Please make sure that the CI does not fail. (PRs that pass the CI are more likely to be reviewed. This also applies to your other PRs.)

LouisLau-art:
Good call. I’ll rebase and fix the failing CI jobs on this PR (and clean up the other open Poetry PRs as well) before requesting more review.

src/poetry/installation/installer.py
```diff
@@ -33,6 +33,7 @@
 
 
 class Installer:
+    _POETRY_SYSTEM_PROJECT_NAME = "poetry-instance"
```
radoering: Now, we have two places where the constant is defined. We should define the constant in one place.

LouisLau-art:
Quick follow-up: acknowledged on CI quality. I’ll post an update on this PR after I rebase and get all required checks green, then request another review.

LouisLau-art:
Thanks for the review — I pushed an update to also handle `poetry self show`.

What changed:
- `ShowCommand` now chooses the lock-create hint based on system project context.
- For self context, missing lock now suggests `poetry self lock` (instead of `poetry lock`).
- Added regression test: `tests/console/commands/self/test_show.py::test_self_show_errors_without_lock_file`.

Validation:
- `python -m pytest -o addopts='' tests/installation/test_installer.py::test_not_fresh_lock_self_project tests/console/commands/self/test_show.py::test_self_show_errors_without_lock_file -q`

Could you please take another look?

LouisLau-art:
Follow-up pushed to fix the failing `pre-commit.ci` check.

- Applied `ruff format` on `src/poetry/installation/installer.py` (no logic change).
- Re-ran targeted tests:
  - `python -m pytest -o addopts= tests/installation/test_installer.py::test_not_fresh_lock_self_project tests/console/commands/self/test_show.py::test_self_show_errors_without_lock_file -q`

Could you please re-check when CI completes?

src/poetry/installation/installer.py
LouisLau-art: Addressed in 38d02e44. I moved the system project name to `poetry.utils.constants.POETRY_SYSTEM_PROJECT_NAME` and reused it from `Installer`, `ShowCommand`, `SelfCommand`, and the installer test so the value is defined in one place only.

LouisLau-art:
Follow-up update: branch is rebased, the constant is centralized, and all required checks are green now. Could you take another look when you have time?

PR #10716, opened 2026-02-06
fix(env): support posix-compliant shells in `env activate` on Windows

### LINKED ISSUE #10395: `poetry env activate` does not work in Bash for Windows
### Description

### Workarounds

pipx

### Operating System

Windows

### Poetry Version

2.1.3

### Poetry Configuration

```bash session
Not relevant
```

### Python Sysconfig

<details>
  

  ```
Not relevant
  ```
</details>

### Example pyproject.toml

```TOML

```

### Poetry Runtime Logs

<details>
  

  ```
Not relevant
  ```
</details>

LouisLau-art:
## Summary
- make `env activate` decide output format by detected shell instead of host OS
- keep shell-prefixed output (`source ...`) for POSIX shells even on Windows
- preserve direct invocation output for `cmd`, `powershell`, and `pwsh`
- add a regression test for Windows + bash shell detection

Fixes #10395.

## Testing
- `pytest -q -o addopts='' tests/console/commands/env/test_activate.py`

## Summary by Sourcery

Adjust `env activate` to select activation command format based on detected shell rather than host OS, with improved Windows handling and updated tests.

Bug Fixes:
- Ensure POSIX-style shells on Windows receive shell-prefixed activation commands instead of Windows-specific invocation format.
- Preserve correct direct invocation behavior for `cmd`, `powershell`, and `pwsh` by refining command and quoting logic.
- Raise a shell not supported error for unsupported C-shell variants on Windows instead of producing invalid activation commands.

Tests:
- Extend activation command tests to cover POSIX shells on Windows and validate Windows-specific quoting and command behavior.

### DIFF
--- a/tests/console/commands/env/test_activate.py
+++ b/tests/console/commands/env/test_activate.py
@@ -4,6 +4,7 @@ from typing import TYPE_CHECKING
 
 import pytest
 
+from poetry.console.commands.env.activate import ShellNotSupportedError
 from poetry.utils._compat import WINDOWS
 
 
@@ -44,10 +45,15 @@ def test_env_activate_prints_correct_script(
     mocker.patch("shellingham.detect_shell", return_value=(shell, None))
     mocker.patch("poetry.utils.env.EnvManager.get", return_value=tmp_venv)
 
-    tester.execute()
+    if WINDOWS and shell in {"csh", "tcsh"}:
+        with pytest.raises(ShellNotSupportedError):
+            tester.execute()
 
-    line = tester.io.fetch_output().rstrip("\n")
-    assert line == f"{command} {tmp_venv.bin_dir.as_posix()}/activate{ext}"
+    else:
+        tester.execute()
+
+        line = tester.io.fetch_output().rstrip("\n")
+        assert line == f"{command} {tmp_venv.bin_dir.as_posix()}/activate{ext}"
 
 
 @pytest.mark.parametrize(


LouisLau-art:
Thanks for the detailed review — agreed on all points.

I’ll update this PR accordingly:
- use shell-specific command tokening so PowerShell uses `&` and cmd stays empty
- remove the `_quote`-side `&` handling
- extend the existing parameterized tests instead of adding a separate one

I’ll push a follow-up revision that addresses the Windows failures.

LouisLau-art:
Addressed the requested changes and pushed `5f427522`.

What changed:
- `powershell` / `pwsh` now use `&` as the command token (instead of `.`)
- `cmd` now uses an empty command token (so output is just the quoted `.bat` path)
- removed `&` prefixing from `_quote` (so command tokening and quoting are separated)
- folded the Windows-bash case into the existing Windows parameterized test (removed the standalone test)

Validation run:
- `poetry run pytest -o addopts='' tests/console/commands/env/test_activate.py -q`
- result: `11 passed, 4 skipped`

LouisLau-art:
Pushed another follow-up: `abe08c02`.

Root cause of the Windows CI failures was assertion mismatch in the test expectations (actual output used single-quoted paths for pwsh/powershell/bash via `shlex.quote`).

Updates:
- adjusted Windows test expectations to match actual quoting behavior:
  - pwsh/powershell: `& '{path}'`
  - bash: `source '{path}'`
- small Ruff cleanup (`SIM102`) in `_quote`

Local verification:
- `pre-commit run --files src/poetry/console/commands/env/activate.py tests/console/commands/env/test_activate.py`
- `poetry run pytest -o addopts='' tests/console/commands/env/test_activate.py -q`
- result: `11 passed, 4 skipped`

LouisLau-art:
All requested fixes are now in and CI is fully green on commit `abe08c02` (including the Windows pytest matrix).

Could you take another look when you have a moment?

radoering:
@sourcery-ai review

PR #10722, opened 2026-02-08
Fix poetry env activate for Bash on Windows

### LINKED ISSUE #10395: `poetry env activate` does not work in Bash for Windows

veeceey:
## Summary

Fixes #10395

This PR fixes the `poetry env activate` command to work correctly with Bash (and similar shells) on Windows.

## Problem

The command `eval $(poetry env activate)` didn't work in Bash running on Windows (e.g., Git Bash, WSL) because the output was missing the `source` command prefix. The code incorrectly assumed that all shells on Windows should skip the command prefix, but this is only true for PowerShell and cmd.

## Changes

- Changed the logic to set an empty command prefix (`""`) only for PowerShell and cmd shells
- Modified the output logic to check `if not command:` instead of `if WINDOWS:`
- All POSIX-style shells (bash, zsh, etc.) now correctly get the "source" prefix regardless of OS

## Testing

- All existing tests pass
- The fix ensures that bash/zsh on Windows will output: `source /path/to/activate`
- PowerShell/cmd on Windows will continue to output just: `/path/to/activate` (or with appropriate quoting)

## Summary by Sourcery

Validate package names for `poetry update`, ensure deterministic dependency constraint ordering in the lockfile, and fix environment activation output for POSIX shells on Windows.

New Features:
- Reject `poetry update` invocations that specify packages not declared as project dependencies.
- Support parsing package arguments with version constraints when validating update targets.

Bug Fixes:
- Ensure `poetry env activate` prints a `source` prefix for POSIX-style shells on Windows while keeping bare script paths for PowerShell and cmd.
- Stabilize the ordering of dependency constraint entries in the lockfile for consistent output.

### DIFF
--- a/src/poetry/console/commands/env/activate.py
+++ b/src/poetry/console/commands/env/activate.py
@@ -48,17 +48,20 @@ class EnvActivateCommand(EnvCommand):
         elif shell in ["csh", "tcsh"]:
             command, filename = "source", "activate.csh"
         elif shell in ["powershell", "pwsh"]:
-            command, filename = "", "activate.ps1"
+            command, filename = ".", "activate.ps1"
         elif shell == "cmd":
-            command, filename = "", "activate.bat"
+            command, filename = ".", "activate.bat"
         elif shell in ["bash", "mksh", "zsh"]:
             command, filename = "source", "activate"
         else:
             command, filename = ".", "activate"
 
         if (activation_script := env.bin_dir / filename).exists():
-            # Only omit the command prefix for PowerShell and cmd on Windows
-            if not command:
+            # On Windows, PowerShell and cmd don't need a command prefix
+            # because _quote already handles invocation (e.g. '& "path"'
+            # for PowerShell). Other shells (bash, zsh, etc.) still need
+            # their prefix even on Windows, which is the fix for #10395.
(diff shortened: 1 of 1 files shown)


dimbleby:
you have made a mess of your branches, this pull request is mixed up with changes from another

veeceey:
I’ll correct it

________________________________

veeceey:
Fixed - rebased the branch so it only contains the changes for this PR. Sorry about the branch mix-up.

veeceey:
Verified the condition now correctly limits the prefix omission to Windows PowerShell/cmd/pwsh shells only. Bash/zsh on Windows retain their 'source' prefix as intended. All 11 activate tests pass.

dimbleby:
cf #10716

veeceey:
Closing in favor of #10716 which addresses the same issue (#10395) with a cleaner approach. Thanks for pointing this out @dimbleby!

PR #10732, opened 2026-02-13
fix(publish): normalize legacy repository URL trailing slash

### LINKED ISSUE #6687: Omitting trailing slash on some PEP 508 repositories prevents inferring the correct upload URL
- **Poetry version**: 1.2.1
- **Python version**: 3.8.10
- **OS version and name**: Linux Mint 20.3 Una
- **pyproject.toml**: not applicable

- [ ] If an exception occurs when executing a command, I executed it again in debug mode (`-vvv` option) and have included the output below. (not applicable)

## Issue
Apologies for reporting essentially two issues at once, but trying to figure out the first led directly into the second for me.

The first is the same basic issue reported in #3600, now in 1.2.1 - configuring `poetry config repositories.testpypi https://test.pypi.org/legacy` without the trailing slash caused silent failures.

The thing is, I get silent failures this way **even when** using `twine` (or `poetry run twine`, as described in Issue #742) to upload instead would succeed. The only symptom is that the package does not appear on TestPyPI.

Second issue: after adding the trailing slash and trying again, I got noisy failures instead -
```
$ poetry config repositories.testpypi https://test.pypi.org/legacy/
$ poetry config pypi-token.testpypi <api token omitted>
$ poetry publish -r testpypi

Publishing <package name omitted> to testpypi
 - Uploading <file name omitted> FAILED

HTTP Error 403: Invalid or non-existent authentication information. See https://
test.pypi.org/help/#invalid-auth for more information. | b'<html>\n <head>\n  <t
itle>403 Invalid or non-existent authentication information. See https://test.py
pi.org/help/#invalid-auth for more information.\n \n <body>\n  <h1>403 Invalid o
r non-existent authentication information. See https://test.pypi.org/help/#inval
id-auth for more information.\n  Access was denied to this resource.<br/><br/>\n
Invalid or non-existent authentication information. See https://test.pypi.org/he
lp/#invalid-auth for more information.\n\n\n \n'
```
As far as I can tell, `poetry config pypi-token.testpypi <API token>` *doesn't actually do anything*, despite everything I can find in the documentation and third-party guides. There is no update to the content of `~/.config/pypoetry/config.toml`, and I get the error shown above when trying to use `poetry publish`. (If it is supposed to write information somewhere else, I have no idea how to verify that.) It seems I am forced to specify my actual username and password explicitly on the command line. Even setting the username to `__token__` and password to the API token value, as described [in PyPI's help](https://test.pypi.org/help/#apitoken), does not seem to work.

LouisLau-art:
Fixes #6687.

When a repository upload URL ends with /legacy (no trailing slash), some servers respond with a redirect that can cause uploads to silently fail. Poetry now normalizes legacy upload URLs to ensure they end with /legacy/.

Test:
- python -m pytest -o addopts='' tests/publishing/test_publisher.py

## Summary by Sourcery

Normalize legacy repository upload URLs used during publishing to avoid issues caused by missing trailing slashes.

Bug Fixes:
- Ensure legacy repository URLs ending with /legacy are normalized to include a trailing slash before publishing to prevent redirect-related upload failures.

Tests:
- Add a publishing test verifying that legacy repository URLs without a trailing slash are normalized before invoking the uploader.

### DIFF
--- a/src/poetry/publishing/publisher.py
+++ b/src/poetry/publishing/publisher.py
@@ -1,6 +1,7 @@
 from __future__ import annotations
 
 import logging
+import urllib.parse
 
 from typing import TYPE_CHECKING
 
@@ -16,6 +17,16 @@ if TYPE_CHECKING:
     from poetry.poetry import Poetry
 
 logger = logging.getLogger(__name__)
+
+
+def _normalize_legacy_repository_url(url: str) -> str:
+    parsed = urllib.parse.urlsplit(url)
+
+    if parsed.path.endswith("/legacy") and not parsed.path.endswith("/legacy/"):
+        parsed = parsed._replace(path=f"{parsed.path}/")
+        return urllib.parse.urlunsplit(parsed)
+
+    return url
 
 
 class Publisher:
@@ -52,6 +63,7 @@ class Publisher:
             url = self._poetry.config.get(f"repositories.{repository_name}.url")
             if url is None:
                 raise RuntimeError(f"Repository {repository_name} is not defined")
+            url = _normalize_legacy_repository_url(url)
 
         if not (username and password):
             # Check if we have a token first
--- a/tests/publishing/test_publisher.py
+++ b/tests/publishing/test_publisher.py
@@ -82,6 +82,31 @@ def test_publish_can_publish_to_given_repository(
     assert f"Publishing {project_name} ({uploader_version}) to foo" in io.fetch_output()
 
 
+def test_publish_normalizes_legacy_repository_url_without_trailing_slash(
+    fixture_dir: FixtureDirGetter, mocker: MockerFixture, config: Config
+) -> None:
+    uploader_auth = mocker.patch("poetry.publishing.uploader.Uploader.auth")
+    uploader_upload = mocker.patch("poetry.publishing.uploader.Uploader.upload")
+
+    poetry = Factory().create_poetry(fixture_dir("sample_project"))
+    poetry._config = config
+    poetry.config.merge(
+        {
+            "repositories": {"testpypi": {"url": "https://test.pypi.org/legacy"}},
+            "http-basic": {"testpypi": {"username": "foo", "password": "bar"}},
+        }
+    )
+
+    publisher = Publisher(poetry, NullIO())
+    publisher.publish("testpypi", None, None)
+
+    assert uploader_auth.call_args == [("foo", "bar")]
+    assert uploader_upload.call_args == [
+        ("https://test.pypi.org/legacy/",),
+        {"cert": True, "client_cert": None, "dry_run": False, "skip_existing": False},
+    ]
+
+
 def test_publish_raises_error_for_undefined_repository(
     fixture_dir: FixtureDirGetter, config: Config
 ) -> None:


src/poetry/publishing/publisher.py
```diff
@@ -52,6 +63,7 @@ def publish(
             url = self._poetry.config.get(f"repositories.{repository_name}.url")
             if url is None:
                 raise RuntimeError(f"Repository {repository_name} is not defined")
+            url = _normalize_legacy_repository_url(url)
```
radoering, line 66: Wouldn't this be good enough?

```suggestion
            if url.endswith("/legacy"):
                url += "/"
```

dimbleby:
Is there definitely a problem that needs solving here?   Do we have a reproducer?

I see that the issue report was made in 2022.  If there is a real problem here, with a trivial fix available, it is surprising that it should have survived all this time.

LouisLau-art:
@dimbleby Yes, I believe there is a real reproducer here, and it is the first half of #6687 (the trailing-slash part, not the auth confusion that came later in the same report).

The minimal case is:

```bash
poetry config repositories.testpypi https://test.pypi.org/legacy
poetry publish -r testpypi
```

The original reporter notes that this silently fails, while the same flow works once the URL is changed to `https://test.pypi.org/legacy/`. They also mention that `twine` succeeds against the same endpoint, which matches the idea that normalizing the configured legacy upload URL is sufficient here.

So this PR is intentionally very narrow: it only normalizes configured repository URLs that end in `/legacy` to `/legacy/` before upload. It does not try to address the token/auth behavior discussed later in the issue.

My guess for why this survived for so long is that it only affects explicitly configured legacy endpoints without the trailing slash; users who rely on the default PyPI/TestPyPI setup, or who already configure `/legacy/`, would never hit it.

If you'd prefer, I can also add a slightly higher-level regression test, but I started with the smallest change at the point where Poetry consumes the configured repository URL.

dimbleby:
I see what the original report says.  Did you verify that it is true today?

LouisLau-art:
Yes, I verified it again on the current code.

I tested on `origin/main` at commit `029685cf` (checked on 2026-03-03), with a minimal local repro that mocks `Uploader.upload` so no network is involved.

Result:
- config URL `https://test.pypi.org/legacy` -> upload URL `https://test.pypi.org/legacy`
- config URL `https://test.pypi.org/legacy/` -> upload URL `https://test.pypi.org/legacy/`

So today it still passes the configured URL through verbatim, and the missing trailing slash case is still present.

dimbleby:
Yes, but my question is whether this is actually an issue.  

Does pypi - or testpypi - really reject uploads when the user does not provide a trailing slash?  This should be answered by experiment.

dimbleby:
Either way I don't know that it is up to poetry to guess that users meant something different than what they said.  

The endpoint [is documented](https://docs.pypi.org/api/upload/)  as <https://upload.pypi.org/legacy/>.  Probably it's normal that if you configure the wrong endpoint, it won't work.

radoering:
I just tried with `test.pypi.org`. If the trailing slash is missing, I get the following error message:

```
HTTP Error 400: Bad Request | b'<html>\n <head>\n  <title>400 Bad Request\n \n <body>\n  <h1>400 Bad Request\n  The server could not comply with the request since it is either malformed or otherwise incorrect.<br/><br/>\nPOST body may not contain duplicate keys (URL: &#x27;https://test.pypi.org/legacy&#x27;)\n\n\n \n'
```

From a user's point of view at least a better error message would be nice if you just forget a slash.

> Either way I don't know that it is up to poetry to guess that users meant something different than what they said.

Valid point. I think if there always has to be a trailing slash we can as well add it if it is missing. However, I am not sure if this is specific to PyPI or also valid for other indices...

radoering:
Superseded by #10785

PR #10735, opened 2026-02-13
fix: poetry env activate works in Bash for Windows

### LINKED ISSUE #10395: `poetry env activate` does not work in Bash for Windows

by22Jy:
## Summary

- `poetry env activate` now correctly outputs the activation command for Bash on Windows
- Previously, it only output the path without `source` prefix, breaking `eval $(poetry env activate)`

## Context

Fixes #10395

On Windows, the `poetry env activate` command was checking for the OS instead of the shell type. This meant that Bash (e.g., Git Bash) users could not use `eval $(poetry env activate)` because the command was missing the `source` prefix.

## Changes

- Changed the condition from `if WINDOWS:` to `if shell in ["powershell", "pwsh", "cmd"] and WINDOWS:`
- PowerShell and cmd still get the path-only output for `Invoke-Expression`
- All other shells (including Bash on Windows) now get the correct `source <path>` format

## Testing

On Windows with Git Bash, `poetry env activate` now outputs:
```
source "/path/to/venv/Scripts/activate"
```

Instead of just:
```
"/path/to/venv/Scripts/activate"
```

## Summary by Sourcery

Bug Fixes:
- Ensure `poetry env activate` outputs a `source`-prefixed activation command for Bash and other non-PowerShell shells on Windows, restoring support for `eval $(poetry env activate)` on those shells.

### DIFF
--- a/src/poetry/console/commands/env/activate.py
+++ b/src/poetry/console/commands/env/activate.py
@@ -43,7 +43,7 @@ class EnvActivateCommand(EnvCommand):
     def _get_activate_command(self, env: Env, shell: str) -> str:
         # Normalize shell name to lowercase for case-insensitive comparison
         shell = shell.lower()
-        
+
         if shell == "fish":
             command, filename = "source", "activate.fish"
         elif shell == "nu":


dimbleby:
Why is this better, or worse,  than #10716?

by22Jy:
Hi @dimbleby, thank you for the feedback!

I apologize for the overlap with #10716 - I am relatively new to open source and did not notice the existing PR when I submitted mine.

Looking at both implementations:

**Similarities:**
- Both address #10395 by checking the shell type instead of just the OS
- Both ensure `source` prefix is included for POSIX shells on Windows

**Differences:**
- My PR (#10735) adds `shell = shell.lower()` for case-insensitive comparison and checks `if shell in ["powershell", "pwsh", "cmd"] and WINDOWS:`
- #10716 appears to have a more comprehensive approach with a regression test

I think #10716 is more complete with the regression test. I am happy to close this PR in favor of #10716 if the maintainers prefer that approach.

Thank you for your patience as I learn the contribution workflow!

radoering:
already fixed in #10716

PR #10742, opened 2026-02-20
Check _shutdown flag before executing serial operations

bysiber:
## Problem

In `Executor.execute()`, when a parallel task fails and sets `self._shutdown = True`, the serial operations in the same priority group still execute unconditionally.

`concurrent.futures.wait()` does not re-raise exceptions from worker threads — it waits for all futures to complete and returns. After it returns, the loop over `serial_operations` runs every serial operation without checking `_shutdown`. The shutdown check only happens after the `try`/`except` block.

```python
try:
    wait(tasks)                          # parallel task set _shutdown = True
    for operation in serial_operations:
        self._execute_operation(operation)  # runs anyway!
except KeyboardInterrupt:
    self._shutdown = True

if self._shutdown:  # too late — serial ops already ran
    ...
```

This matters when parallel installs and serial uninstalls are in the same priority group. A failed compilation in a parallel install should stop the whole process, but instead all uninstalls still proceed, potentially leaving the environment in a worse state than before.

## Fix

Check `self._shutdown` before each serial operation:

```python
for operation in serial_operations:
    if self._shutdown:
        break
    self._execute_operation(operation)
```

## Summary by Sourcery

Guard serial installation operations with the executor shutdown flag and document a separate sdist archive filename handling issue and its fix proposal.

Bug Fixes:
- Prevent serial operations from running after a parallel task failure by checking the executor shutdown flag before each serial operation.

Documentation:
- Add a markdown document describing a bug in sdist archive directory name derivation caused by misuse of str.rstrip and outlining a suffix-matching based fix.

### DIFF
--- a/src/poetry/installation/executor.py
+++ b/src/poetry/installation/executor.py
@@ -268,6 +268,8 @@ class Executor:
             section.write(line)
 
     def _execute_operation(self, operation: Operation) -> None:
+        if self._shutdown:
+            return
         try:
             op_message = self.get_operation_message(operation)
             if self.supports_fancy_output():
--- a/src/poetry/installation/executor.py
+++ b/src/poetry/installation/executor.py
@@ -225,6 +225,8 @@ class Executor:
                 wait(tasks)
 
                 for operation in serial_operations:
+                    if self._shutdown:
+                        break
                     self._execute_operation(operation)
 
             except KeyboardInterrupt:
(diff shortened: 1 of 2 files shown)


radoering (commented):
Wouldn't it make sense to put this check at the beginning of `_execute_operation` so that waiting tasks are also aborted early?

radoering (approved):

PR #10751, opened 2026-02-28
Fix #8797: validate version constraint in interactive init flow

### LINKED ISSUE #8797: Interactive dependency prompt accepts invalid version strings
- **Poetry version**: 1.2.2
- **Python version**: 3.10.10
- **OS version and name**: macOS Sonoma 14.1.2
- **pyproject.toml**: https://gist.github.com/nk9/4fa6fe37487d50308deacd425c8ade16

- [x] I am on the [latest](https://github.com/python-poetry/poetry/releases/latest) stable Poetry version, installed using a [recommended method](https://python-poetry.org/docs/#installation).
- [ ] If an exception occurs when executing a command, I executed it again in debug mode (`-vvv` option) and have included the output below.

## Issue
When using the interactive package adding flow, it's possible to enter a version which isn't semver compatible. For example, I missed that I was being asked for a version, and instead typed the name of the next package I wanted to add:

```
Add a package (leave blank to skip): flake8
Found 20 packages matching flake8
Showing the first 10 matches

Enter package # to add, or the complete package name if it is not listed []:
 [ 0] flake8
 [ 1] flake9
 [ 2] flake518
 [ 3] flake99
 [ 4] flake64
 [ 5] flake82
 [ 6] flake8-2020
 [ 7] flake8-zale
 [ 8] flake8-return
 [ 9] flake2lint
 [10]
 > 0
Enter the version constraint to require (or leave blank to use the latest version): isort
```

Poetry dutifully put `"isort"` as the version for the flake8 package, which of course doesn't exist. So when I tried to remove the dependency, it wouldn't let me:

```
$ poetry remove --group dev flake8
Could not parse version constraint: isort
```

## Proposed solution

I believe it's theoretically possible for packages to use non-semver strings for their versions, but in practice I believe that is vanishingly rare. So while it may not be possible to **require** a semver string for the interactive flow, it seems like Poetry should be skeptical of a non-semver string and ask for confirmation that the user really wants to do that. Anyway, when using the interactive flow, I wouldn't expect Poetry to let me create a file which it can't then parse correctly.

gcomneno:
Closes #8797

Adds validation of version constraint in interactive init flow using parse_constraint.

If the user enters an invalid version constraint, an error is shown and the prompt is repeated.

Includes regression test.

## Summary by Sourcery

Validate dependency version constraints entered in the interactive init flow and ensure invalid inputs are rejected with a clear error message.

Bug Fixes:
- Prevent acceptance of invalid dependency version constraints during the interactive init workflow by validating user input before applying it.

Tests:
- Add a regression test covering rejection of invalid version constraints and acceptance of a subsequent valid constraint in the interactive init flow.

### DIFF
--- a/src/poetry/console/commands/init.py
+++ b/src/poetry/console/commands/init.py
@@ -11,6 +11,8 @@ from typing import ClassVar
 
 from cleo.helpers import option
 from packaging.utils import canonicalize_name
+from poetry.core.constraints.version import parse_constraint
+from poetry.core.constraints.version.exceptions import ParseConstraintError
 from tomlkit import inline_table
 
 from poetry.console.commands.command import Command
@@ -395,6 +397,14 @@ The <c1>init</c1> command creates a basic <comment>pyproject.toml</> file in the
                             f"Using version <b>{package_constraint}</b> for"
                             f" <c1>{package}</c1>"
                         )
+                    else:
+                        try:
+                            parse_constraint(package_constraint)
+                        except ParseConstraintError:
+                            self.line_error(
+                                "<error>Invalid version constraint. Please enter a valid version specifier.</error>"
+                            )
+                            continue
 
                     constraint["version"] = package_constraint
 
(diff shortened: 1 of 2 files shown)


gcomneno:
Hi! I’m reducing the number of OSS efforts I actively maintain and won’t be able to follow up on this PR. I’m closing it to avoid leaving it hanging. Thanks for your time and for maintaining the project!

PR #10757, opened 2026-03-05
docs: clarify configuration source precedence

### LINKED ISSUE #5029: [docs] configuration documentation does not specify the precedence of each method

## Issue
https://python-poetry.org/docs/configuration/ specifies which methods are availible to configure poetry, but does not specify the precedence of the options.

e.g. if a value is set via both `poetry config`, `poetry config --local`, and an environment variable, the documentation does not specify which method will be picked.

(local trumping global is obvious, but local vs env is less so)

looking at the code it looks like the sources have the following priority (top wins):

- environment var
- `./poetry.toml`
- `~/.config/pypoetry/auth.toml`
- `~/.config/pypoetry/config.toml`

it would be nice to have this formalised so we can be confident it wont change without warning.

mitre88:
## Summary
- Document configuration source precedence in `docs/configuration.md`.
- Clarify that environment variables override local config, global config, and default values.
- Add a note that credential values may also be read from `auth.toml` and keyring, while environment variables still override file-based values.

Resolves: #5029

## Summary by Sourcery

Clarify configuration source precedence and credential lookup behavior in the configuration documentation.

Documentation:
- Document the precedence order between environment variables, local configuration, global configuration, and default values in the configuration guide.
- Explain how repository credentials can be sourced from auth.toml and the system keyring, while environment variables still take precedence.

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -40,20 +40,6 @@ If a setting is defined in both `poetry.toml` (local/project) and `config.toml` 
 the local/project configuration takes precedence over the global configuration.
 {{% /note %}}
 
-## Configuration sources
-
-When a setting is set in multiple places, Poetry applies the following precedence
-(from highest to lowest):
-
-1. Environment variables (for example, `POETRY_VIRTUALENVS_CREATE`)
-2. The local `poetry.toml` file (created with `poetry config --local`)
-3. The global `config.toml` file
-4. The setting's default value
-
-For repository credentials (`http-basic.*`, `pypi-token.*`), Poetry may also read
-from `auth.toml` and the system keyring. Environment variables still take
-precedence over file-based values.
-
 {{% warning %}}
 Be mindful when checking in this file into your repository since it may contain user-specific or sensitive information.
 {{% /warning %}}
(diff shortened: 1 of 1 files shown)


docs/configuration.md
```diff
@@ -40,6 +40,20 @@ If a setting is defined in both `poetry.toml` (local/project) and `config.toml`
 the local/project configuration takes precedence over the global configuration.
 {{% /note %}}
 
+## Configuration sources
```
radoering: I wonder if this section should be placed after "Using environment variables" because here environment variables were not even introduced.

By the way, the section is placed in the middle of another section at the moment. The warning that follows belongs to "Local configuration".

docs/configuration.md
```diff
@@ -40,6 +40,20 @@ If a setting is defined in both `poetry.toml` (local/project) and `config.toml`
...
+When a setting is set in multiple places, Poetry applies the following precedence
+(from highest to lowest):
+
+1. Environment variables (for example, `POETRY_VIRTUALENVS_CREATE`)
+2. The local `poetry.toml` file (created with `poetry config --local`)
+3. The global `config.toml` file
+4. The setting's default value
+
+For repository credentials (`http-basic.*`, `pypi-token.*`), Poetry may also read
+from `auth.toml` and the system keyring. Environment variables still take
```
radoering: We should add a link to where this is explained like:

```suggestion
from `auth.toml` and the system keyring - for details see <link>. Environment variables still take
```

radoering (changes requested):

mitre88:
Gentle ping — this PR is ready for review. All CI checks pass. Would appreciate a look when possible. Thanks!

radoering:
@mitre88 Do you see my review feedback from March 15?

mitre88:
Hi @radoering — sorry for the delay! I've addressed your review feedback:

1. **Moved 'Configuration sources' section** after 'Using environment variables' so the precedence order is clear
2. **Moved the warning** back under 'Local configuration' where it belongs
3. **Added a link** to the Repositories docs for auth.toml/keyring details

Please take another look when you get a chance. Thanks!

docs/configuration.md
```diff
@@ -40,6 +40,20 @@ If a setting is defined in both `poetry.toml` (local/project) and `config.toml`
...
+precedence over file-based values.
+
 {{% warning %}}
 Be mindful when checking in this file into your repository since it may contain user-specific or sensitive information.
```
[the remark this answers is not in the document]
radoering: unrelated

radoering (approved):

PR #10758, opened 2026-03-05
fix: warn when group name is normalized in --with/--without/--only options

### LINKED ISSUE #10580: Add error/warning when a group specified in `poetry install --with` does not exist or a group name is modified in the lock file
### Issue Kind

Change in current behaviour

### Description

When upgrading poetry to 2.2.0 and beyond, names of dependency groups are now silently normalized: https://github.com/python-poetry/poetry/pull/10387 which can act cause a breaking change.

Updating a lock file after the upgrade breaks any automation that depends on groups with names subject to normalization.

Personal example. I bumped a dependency in pyproject.toml and updated the lock file. Because I happened to have the latest poetry,  while the lock in the repo was generated with `2.1.4`, the normalization changed the names of the groups in the lock file. This was overlooked in the review process, as the changes in the lock file tend to be. As a result, after CI has "successfully" installed the dependencies and deployed the code, some dependencies were missing. It wasn't exactly obvious why at the time.

I think it would be a good idea to either:
- output a warning during `poetry lock` when a dependency name is normalized
- return an error when using `--with` option when specified group name does not exist in the lock file

### Impact

Introduces guardrails to avoid breaking behaviour. Communicates clearly the modifications applied.

### Workarounds

1. Make sure the names of the dependency groups in the lock file are the same as the ones in the toml.
2. Rename the dependency groups in the toml

srikaaviya:
Resolves: #10580

Since Poetry 2.2.0, dependency group names are silently normalized in the
lock file (e.g. `Dev-Tools` → `dev-tools`). This could cause CI pipelines
to silently skip installing a group when using `--with Dev-Tools`.

This PR adds two guardrails:
1. **Normalization warning**: When `--with`/`--without`/`--only` is passed
   a group name that differs from its normalized form, Poetry now emits a warning suggesting t
he user update their command to use the normalized name.

2. **Validation fix**: [_validate_group_options()] now normalizes group names
   before checking `has_dependency_group()`, so `Dev-Tools`, `dev_tools`, and `dev-tools` all correctly resolve to the same group.

## Tests added
- [test_non_normalized_group_name_emits_warning] - verifies a warning is
  emitted for `--with`, `--without`, and `--only` when a non-normalized name is passed. 
- [test_normalized_group_name_does_not_raise] - verifies no `GroupNotFoundError`
  is raised when a valid group name is passed in a non-normalized form.

## Summary by Sourcery

Warn when dependency group CLI options use names that normalize differently and ensure normalized names are used for validation.

Bug Fixes:
- Normalize group names before validating dependency groups so differently formatted names resolve to existing groups.

Enhancements:
- Emit a warning when dependency group names passed to --with/--without/--only are normalized to a different name, guiding users toward the canonical form.

Tests:
- Add tests to verify a warning is emitted for non-normalized group names across --with/--without/--only options.
- Add a test ensuring non-normalized but valid group names do not raise GroupNotFoundError.

### DIFF
--- a/src/poetry/console/commands/group_command.py
+++ b/src/poetry/console/commands/group_command.py
@@ -103,6 +103,16 @@ class GroupCommand(Command):
             key: {canonicalize_name(group) for group in key_groups}
             for key, key_groups in groups.items()
         }
+        # Warn if any group name was normalized
+        for key_groups in groups.values():
+            for group in key_groups:
+                normalized = canonicalize_name(group)
+                if normalized != group:
+                    self.line_error(
+                        f"<warning>Group '{group}' was normalized to '{normalized}'."
+                        " Consider updating your command to use the normalized name.</warning>"
+                    )
+
         norm_default_groups = {canonicalize_name(name) for name in self.default_groups}
 
         return norm_groups["only"] or norm_default_groups.union(
@@ -121,7 +131,9 @@ class GroupCommand(Command):
         invalid_options = defaultdict(set)
         for opt, groups in group_options.items():
             for group in groups:
-                if not self.poetry.package.has_dependency_group(group):
+                if not self.poetry.package.has_dependency_group(
+                    canonicalize_name(group)
+                ):
                     invalid_options[group].add(opt)
         if invalid_options:
             message_parts = []
--- a/src/poetry/console/commands/group_command.py
+++ b/src/poetry/console/commands/group_command.py
@@ -103,11 +103,13 @@ class GroupCommand(Command):
             key: {canonicalize_name(group) for group in key_groups}
             for key, key_groups in groups.items()
         }
-        # Warn if any group name was normalized
+        # Warn if any group name was normalized (deduplicate warnings)
+        warned: set[str] = set()
         for key_groups in groups.values():
             for group in key_groups:
                 normalized = canonicalize_name(group)
-                if normalized != group:
+                if normalized != group and group not in warned:
+                    warned.add(group)
                     self.line_error(
                         f"<warning>Group '{group}' was normalized to '{normalized}'."
                         " Consider updating your command to use the normalized name.</warning>"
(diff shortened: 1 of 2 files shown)


src/poetry/console/commands/group_command.py
```diff
@@ -121,7 +133,9 @@ def _validate_group_options(self, group_options: dict[str, set[str]]) -> None:
         invalid_options = defaultdict(set)
         for opt, groups in group_options.items():
             for group in groups:
-                if not self.poetry.package.has_dependency_group(group):
+                if not self.poetry.package.has_dependency_group(
+                    canonicalize_name(group)
```
radoering, line 137: This should be redundant because `has_dependency_group` already calls `canonicalize_name`.

src/poetry/console/commands/group_command.py
```diff
@@ -103,6 +103,18 @@ def activated_groups(self) -> set[NormalizedName]:
             key: {canonicalize_name(group) for group in key_groups}
             for key, key_groups in groups.items()
         }
+        # Warn if any group name was normalized (deduplicate warnings)
```
radoering, line 106: Not sure if this makes sense because the group name is just normalized anyway.

radoering (commented):
I am not sure if this solves an actual issue. Maybe, I am missing something. I also commented on the issue this PR is trying to solve.

src/poetry/console/commands/group_command.py
srikaaviya: Hi @radoering — thanks for the review. I agree the warning doesn't really belong in `activated_groups()` since normalization works there.

Would a warning in `locker.py` make more sense instead? Specifically in `set_lock_data()`, when group names from `pyproject.toml` are being normalized before being written to the lock file?

src/poetry/console/commands/group_command.py
srikaaviya: @radoering, Its redundant. will remove the change.

radoering:
I do not think we need a warning at all. The reported issue was with an ancient version of Poetry (see https://github.com/python-poetry/poetry/issues/10580#issuecomment-4113338760) so that I conclude that there is no real issue with current versions. Thanks for your effort anyway.

Issue #10364, opened 2025-04-29
Windows installation folder using Python from Windows Store (3.12)


Houseton:
### Issue Kind

Error in existing documentation

### Existing Link

https://python-poetry.org/docs/#installing-with-the-official-installer

### Description

Identical as per this comment on closed issue:
https://github.com/python-poetry/poetry/issues/8566#issuecomment-1776294473

**System**: Windows 11 Pro 24H3
**Build**: Version	10.0.26100 Build 26100
**Python Version**: 3.12 (Microsoft Store)

I did a complete fresh initial installation of Poetry using the information provided on the [website](https://python-poetry.org/docs/#installing-with-the-official-installer).

The information for the PATH that shows in Terminal/PS is incorrect as well as the information on the [website](https://python-poetry.org/docs/#installing-with-the-official-installer). poetry.exe is not installed in `%appdat%\Python\scripts` when you are using Python Interpreter from the Microsoft Store. 

For me, just like the commenter in that issue, it is installed in a different location:
`%localappdata%\Packages\PythonSoftwareFoundation.Python.3.12_qbz5n2kfra8p0\LocalCache\Roaming\Python\Scripts`

It seems the Microsoft Store version of the Python Interpreter installs Poetry in different places namely:
- %localappdata%\Packages\PythonSoftwareFoundation.Python.**[Version]**_qbz5n2kfra8p0\LocalCache\Roaming\pypoetry\venv\Scripts
   - This is where the `venv` folder is installed
- %localappdata%\Packages\PythonSoftwareFoundation.Python.**[Version]**_qbz5n2kfra8p0\LocalCache\Roaming\Python\Scripts
   - This is where the `poetry.exe` is installed

Judging from the comment, the ID tag in the path seems to be the same `_qbz5n2kfra8p0` for the 3.11 and 3.12 installation.

Issue #8566, opened 2023-10-23
The Windows path in the installation tutorial is incorrect.


brokedarius:
It should be:

%appdata%\Roaming\pypoetry\venv\Scripts\

%APPDATA%\Roaming\Python\Scripts

but the tutorial indicates:

%APPDATA%\pypoetry\venv\Scripts\poetry

%APPDATA%\Python\Scripts

nothing big but I thought this would be relevant for beginners like me.

radoering:
AFAIK, on a typical Windows machine `%APPDATA%` is the same as `C:\Users\<username>\AppData\Roaming` (and there is no other `Roaming` directory inside). Maybe, you have an uncommon setup?

artsakenos:
I want to report a further related issue.
In my case I have this settings
* Python 3.11.6 (installed from windows store)
* pip 23.2.1 

The installation process seems to go smoothly:
* (Invoke-WebRequest -Uri https://install.python-poetry.org -UseBasicParsing).Content | python -
        * It will add the `poetry` command to Poetry's bin directory, located at:
        * C:\Users\andre\AppData\Roaming\Python\Scripts
        * You can uninstall at any time by executing this script with the --uninstall option, and these changes will be reverted.
        * Poetry (1.6.1) is installed now. Great!
        * To get started you need Poetry's bin directory (C:\Users\andre\AppData\Roaming\Python\Scripts) in your `PATH` environment variable.
        * Alternatively, you can call Poetry explicitly with `C:\Users\andre\AppData\Roaming\Python\Scripts\poetry`.
 
But poetry is nowhere to be found. After searching in all Local and Roaming folders I found it at 
* C:\Users\andre\AppData\Local\Packages\PythonSoftwareFoundation.Python.3.11_qbz5n2kfra8p0\LocalCache\Roaming\pypoetry\venv\Scripts
* C:\Users\andre\AppData\Local\Packages\PythonSoftwareFoundation.Python.3.11_qbz5n2kfra8p0\LocalCache\Roaming\Python\Scripts

this makes me think that the target poetry folder is strictly related to the pip script source folder, which in case of windows app store python is put in some random location.

brokedarius:
> AFAIK, on a typical Windows machine `%APPDATA%` is the same as `C:\Users\<username>\AppData\Roaming` (and there is no other `Roaming` directory inside). Maybe, you have an uncommon setup?

I have a fairly normal setup, latest version of python (non ms-store version) and latest pip.

[...]

PR #10759, opened 2026-03-06
docs: add note about Microsoft Store Python PATH on Windows

KellyBallinger05:
Resolves: #10364

The installation docs state that on Windows, the `poetry` binary is installed to `%APPDATA%\Python\Scripts`. However, when Python is installed  via the Microsoft Store, the binary is instead placed at:

`%LOCALAPPDATA%\Packages\PythonSoftwareFoundation.Python<Version>_qbz5n2kfra8p0\LocalCache\Roaming\Python\Scripts`

This has caused confusion for multiple users (see also #8566 - which is now closed). This PR adds  a note to the "Add Poetry to your PATH" step in the official installer section, consistent in style with the existing Microsoft Store note in step 1 of the same section.

## Summary by Sourcery

Documentation:
- Add a note to the installation guide describing the Microsoft Store-specific path where the Poetry binary is installed on Windows.

### DIFF
--- a/docs/_index.md
+++ b/docs/_index.md
@@ -195,7 +195,8 @@ If you have installed Python through the Microsoft Store, the `poetry` binary
 will be installed to a different location, for example:
 
 ```
-%LOCALAPPDATA%\Packages\PythonSoftwareFoundation.Python.3.12_qbz5n2kfra8p0\LocalCache\Roaming\Python\Scripts
+%LOCALAPPDATA%\Packages\PythonSoftwareFoundation.Python.3.12_qbz5n2kfra8p0
+\LocalCache\Roaming\Python\Scripts
 ```
 
 Replace `3.12` with your installed Python version and `qbz5n2kfra8p0` with your suffix.


docs/_index.md
```diff
@@ -190,6 +190,15 @@ The installer creates a `poetry` wrapper in a well-known, platform-specific dire
 - `%APPDATA%\Python\Scripts` on Windows.
 - `$POETRY_HOME/bin` if `$POETRY_HOME` is set.
 
+{{% note %}}
+If you have installed Python through the Microsoft Store, the `poetry` binary
+will be installed to a different location:
+
+`%LOCALAPPDATA%\Packages\PythonSoftwareFoundation.Python.3.12_qbz5n2kfra8p0\LocalCache\Roaming\Python\Scripts`
+
+Replace `3.12` with your installed Python version.
```
radoering: The suffix `_qbz5n2kfra8p0` probably is not fix, is it?

Maybe, we should add "for example" and something at the end?

```suggestion
will be installed to a different location, for example:

`%LOCALAPPDATA%\Packages\PythonSoftwareFoundation.Python.3.12_qbz5n2kfra8p0\LocalCache\Roaming\Python\Scripts`

Replace `3.12` with your installed Python version and `qbz5n2kfra8p0` with your suffix.
```

KellyBallinger05: Yes, good catch. My apologies for not clarifying that there will be a variable suffix attached depending on the user. Committing the suggestion.

PR #10763, opened 2026-03-08
feat: deferred updates

### LINKED ISSUE #10646: Support deferred updates / min release age
### Issue Kind

Brand new capability

### Description

Hello,

Thank you for your great work with poetry

An effective security method is to defer the use of dependencies: give it some time between the release time and the application in your project (depending on the environment etc)

poetry could implement that feature, perhaps through a dedicated new command line argument (and/or a default configuration in pyproject.toml)

The current behavior would be kept by default

The argument would be a number of days (strictly > 0)

poetry would exclude all packages whose `upload-time` is lower than `now - timedelta(days=number_of_days)`

Dependencies resolution would then process as usual

(this is a bit like the processing of the "yanked" package status)

Best regards,

### Impact

This would help people increase the security of their project (optionally)

### Workarounds

None known

marabb01:
Adds installer.minimum-release-age and installer.minimum-release-age-exclude config options. Versions released more recently than the configured threshold are filtered out during dependency resolution, guarding against supply chain attacks on freshly published packages (inline with pnpm's minimumReleaseAge).

Upload time is sourced from the PEP 691 Simple API page already fetched for version listing, so no extra HTTP requests are made. HTML-only index pages fail open (no upload_time available). Glob patterns are supported in the exclude list.

# Pull Request Check List

Resolves: #10646
Also discussed here: https://github.com/orgs/python-poetry/discussions/10555

## Summary by Sourcery

Introduce configurable minimum release age filtering for dependency resolution to avoid using very recently published package versions.

New Features:
- Add installer.minimum-release-age config option to restrict dependency resolution to package versions older than a configured age.
- Add installer.minimum-release-age-exclude config option supporting glob patterns to bypass minimum-age checks for selected packages.
- Apply minimum release age checks in HTTP-based repositories when selecting candidate versions, failing open when upload time data is unavailable.

Tests:
- Add unit tests covering release age calculation, minimum-release-age filtering behavior, exclusion patterns, and fail-open semantics when upload time metadata is missing.

### DIFF
--- a/src/poetry/config/config.py
+++ b/src/poetry/config/config.py
@@ -405,11 +405,10 @@ class Config:
             return int_normalizer
 
         if name == "installer.minimum-release-age-exclude":
-            return lambda val: (
-                [v.strip() for v in val.split(",")]
-                if isinstance(val, str)
-                else list(val)
-            )
+            return lambda val: [
+                str(v).strip()
+                for v in (val.split(",") if isinstance(val, str) else val)
+            ]
 
         if name in ["installer.no-binary", "installer.only-binary"]:
             return PackageFilterPolicy.normalize
--- a/src/poetry/config/config.py
+++ b/src/poetry/config/config.py
@@ -170,6 +170,8 @@ class Config:
             "no-binary": None,
             "only-binary": None,
             "build-config-settings": {},
+            "minimum-release-age": None,
+            "minimum-release-age-exclude": [],
         },
         "python": {"installation-dir": os.path.join("{data-dir}", "python")},
         "solver": {
@@ -398,8 +400,16 @@ class Config:
         if name in {
             "installer.max-workers",
             "requests.max-retries",
+            "installer.minimum-release-age",
         }:
             return int_normalizer
+
+        if name == "installer.minimum-release-age-exclude":
+            return lambda val: (
+                [v.strip() for v in val.split(",")]
+                if isinstance(val, str)
+                else list(val)
+            )
 
         if name in ["installer.no-binary", "installer.only-binary"]:
             return PackageFilterPolicy.normalize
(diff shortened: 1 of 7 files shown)


src/poetry/config/config.py
```diff
@@ -398,9 +400,16 @@ def _get_normalizer(name: str) -> Callable[[str], Any]:
         if name in {
             "installer.max-workers",
             "requests.max-retries",
+            "installer.minimum-release-age",
```
acdha, line 403: The additions in this file have the same code as `console.commands.config` which seems like a recipe for unexpected behaviour. For example, this version does not apply the console version's check that the minimum release age is greater than zero.

src/poetry/config/config.py
```diff
@@ -398,9 +400,16 @@ def _get_normalizer(name: str) -> Callable[[str], Any]:
...
         }:
             return int_normalizer
 
+        if name == "installer.minimum-release-age-exclude":
+            return lambda val: [
+                str(v).strip()
+                for v in (val.split(",") if isinstance(val, str) else val)
```
acdha, line 410: Since the definition of this option is an optional list of glob separated strings, it's not clear that type check here is helpful and it could lead to some oddities (e.g. if I give it a list with a `NoneType` or `bool` the result will be `[…, "None", …]`. It feels like this would be cleaner broken out into a function which could be used in both locations which returns an empty list when given a false-y value, does the string check to convert a single string to a list (esp. from `POETRY_INSTALLER_MINIMUM_RELEASE_AGE_EXCLUDE`), and then does `filter(None, map(str.strip))` before returning it.

radoering:
Thanks for your contribution. However, I think that there are too many issues that still had to be addressed so that I decided to create a PR myself: #10824

Secrus:
Superseded by #10824

PR #10766, opened 2026-03-10
Add tests for `switch_working_directory` helper

SergioChan:
## Summary
- add test that `switch_working_directory()` restores the original cwd when an exception is raised
- add test that `remove=True` deletes the temporary working directory after exiting the context manager

Related to python-poetry/poetry#9161

## Summary by Sourcery

Add coverage for the switch_working_directory helper behavior.

Tests:
- Add a test ensuring switch_working_directory restores the original working directory when an exception is raised inside the context.
- Add a test ensuring switch_working_directory removes the temporary working directory when used with remove=True.

### DIFF
--- a/tests/test_helpers.py
+++ b/tests/test_helpers.py
@@ -2,8 +2,11 @@ from __future__ import annotations
 
 import os
 
+import pytest
+
 from tests.helpers import flatten_dict
 from tests.helpers import isolated_environment
+from tests.helpers import switch_working_directory
 
 
 def test_flatten_dict() -> None:
@@ -44,3 +47,25 @@ def test_isolated_environment_updates_environ() -> None:
     with isolated_environment(environ={"NEW_VAR": "new_value"}):
         assert os.environ["NEW_VAR"] == "new_value"
     assert "NEW_VAR" not in os.environ
+
+
+def test_switch_working_directory_restores_original_cwd_on_error(tmp_path) -> None:
+    original_cwd = os.getcwd()
+
+    with pytest.raises(RuntimeError):
+        with switch_working_directory(tmp_path):
+            assert os.getcwd() == str(tmp_path)
+            raise RuntimeError("boom")
+
+    assert os.getcwd() == original_cwd
+
+
+def test_switch_working_directory_remove_deletes_directory(tmp_path) -> None:
+    temp_dir = tmp_path / "temp-working-dir"
+    temp_dir.mkdir()
+
+    with switch_working_directory(temp_dir, remove=True):
+        assert os.getcwd() == str(temp_dir)
+        assert temp_dir.exists()
+
+    assert not temp_dir.exists()
--- a/tests/test_helpers.py
+++ b/tests/test_helpers.py
@@ -55,7 +55,10 @@ def test_isolated_environment_updates_environ() -> None:
     assert "NEW_VAR" not in os.environ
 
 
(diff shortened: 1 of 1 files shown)


SergioChan:
Follow-up pushed in `a7f78f6` to address failing mypy checks.

### Change
- Added explicit type annotations for `tmp_path` fixture arguments in:
  - `test_switch_working_directory_restores_original_cwd_on_error`
  - `test_switch_working_directory_remove_deletes_directory`
- Added `from pathlib import Path` import.

### Targeted validation
- `python -m mypy --follow-imports=silent --ignore-missing-imports tests/test_helpers.py` ✅

SergioChan:
Pushed follow-up commit 0facb5c2 to fix the failing `pre-commit.ci` check.

What changed (auto-fixes from the repo hooks):
- adjusted typing import style to satisfy Ruff (`TYPE_CHECKING` + deferred `Path` import)
- applied Ruff formatting to the new tests
- simplified the exception test context usage with combined context managers

Validation:
- `python -m pre_commit run --files tests/test_helpers.py`

radoering (changes requested):
The two tests seem arbitrary. A complete test should test both aspects (cwd is changed (back) and cwd is only deleted if `remove=True`) - both under normal conditions and if there is an error. I think one parameterized test makes sense.

SergioChan:
Thanks for the clear review guidance.

I’ll rework this into more complete coverage: both normal/error paths, and delete behavior for `remove=True` vs `remove=False` (likely via parametrization), then push an update.

SergioChan:
Thanks for the concrete feedback — I pushed a follow-up in commit aa866ed0 that rewrites the coverage into one parametrized test over (remove, raise_error).

What this now verifies in one place:
- cwd changes inside the context
- cwd is restored after exit
- directory removal behavior is correct (
emove=True removes, 
emove=False keeps)
- both normal and error paths are covered

Targeted validation run locally:
- python -m pytest --noconftest tests/test_helpers.py -k switch_working_directory -q (4 passed)

If you want, I can split this back into two parametrized tests (restore/remove) for readability, but I kept it as one per your suggestion.

SergioChan:
Pushed follow-up commit edc596c6 to address the failing pre-commit.ci check.\n\n### What changed\n- Reformatted the long pytest.mark.parametrize decorator in 	ests/test_helpers.py (Ruff format).\n- Reworked the nested with in the error-path branch into a single multi-context with (...) block to satisfy Ruff SIM117.\n\n### Validation\n- python -m pre_commit run --files tests/test_helpers.py ✅\n

SergioChan:
Quick follow-up on the requested rework: the PR now has a single parametrized test covering both normal/error paths and remove=true/false behavior, plus a pre-commit clean-up in edc596c6.\n\n@radoering when you have a moment, could you take another look and let me know if this now matches what you had in mind?

SergioChan:
Follow-up pushed: I rebased this branch onto the current main to clear the **BEHIND** state after the latest upstream changes.\n\nValidation after rebase:\n- python -m pytest --noconftest tests/test_helpers.py -k switch_working_directory -q ✅ (4 passed)\n\n@radoering when you have time, could you take another look at the latest revision?

SergioChan:
Thanks for the detailed review. I updated the tests into a single parameterized matrix to cover both behaviors (cwd restore + conditional directory removal) across normal and error paths.\n\nIf you have a minute, could you please take another look and let me know if you’d like this split differently?

radoering (approved):

PR #10782, opened 2026-03-24
feat: add --non-package flag to init and new commands

20xic:
# Add `--non-package` flag to `init` and `new` commands

## Summary

This PR adds a `--non-package` flag to the `poetry init` and `poetry new` commands, allowing users to initialize a project with `package-mode = false` set in `pyproject.toml` directly from the CLI.

## Motivation

Currently, to create a non-package project (e.g., an application, a script collection, or a project that only uses Poetry for dependency management), users must:

1. Run `poetry init` or `poetry new`
2. Manually edit `pyproject.toml` to add `package-mode = false` under `[tool.poetry]`
3. Optionally remove the now-unnecessary `[build-system]` section

This is a common workflow — many Python projects are applications rather than distributable packages. The `--non-package` flag streamlines this by handling everything in a single step.

## Usage

```bash
# Initialize a non-package project in the current directory
poetry init --non-package

# Create a new non-package project
poetry new my-app --non-package

# Combine with other flags
poetry init --non-package --name my-app --dependency requests --dependency click
```

### `src/poetry/layouts/layout.py`
- Added `package_mode: bool = True` parameter to `Layout.__init__()`
- `generate_project_content()`: when `package_mode=False`, sets `package-mode = false` in `[tool.poetry]` instead of generating the `packages` key, and skips the `[build-system]` section entirely
- `create()`: skips package directory (`__init__.py`) and `tests/` creation when `package_mode=False`; README is still created

### `src/poetry/console/commands/init.py`
- Added `--non-package` option to `InitCommand.options`
- In `_init_pyproject()`: when `--non-package` is set, interactive prompts for package-specific fields (name, version, description, author, license, python version) are skipped — sensible defaults are used instead
- Passes `package_mode=not non_package` to the `Layout` constructor

### `src/poetry/console/commands/new.py`
- Added `"non-package"` to the set of options inherited from `InitCommand`

## Example output

Running `poetry init --non-package --name my-app` produces:

```toml
[project]
name = "my-app"
version = "0.1.0"
description = ""
authors = [
    {name = "John", email = "john@example.com"}
]
requires-python = ">=3.12"
dependencies = []

[tool.poetry]
package-mode = false
```

Note the absence of `[build-system]` and `packages` — neither is needed for non-package projects.

## Tests

Added 13 new tests across two test files:

**`test_init.py`** (7 tests):
- Non-interactive mode produces correct `package-mode = false` and no `[build-system]`
- `[project]` section is still generated with name/version
- Works correctly with `--dependency` flag
- Interactive mode skips package-specific prompts
- No `packages` key in `[tool.poetry]`
- Does not create `src/` or `tests/` directories
- Regression: normal `init` (without `--non-package`) is unaffected

**`test_new.py`** (6 tests):
- Creates `pyproject.toml` without `[build-system]`
- Does not create `src/`, `tests/`, or `__init__.py` (README is still created)
- Works with `--name` flag
- No `packages` key in output
- Regression: normal `new` still creates full project structure
- Works in existing empty directories

## Summary by Sourcery

Add support for initializing and creating non-package projects via a new CLI flag and adjust project layout generation accordingly.

New Features:
- Introduce a --non-package flag for the poetry init command to generate non-package projects with package-mode disabled.
- Support the --non-package flag in the poetry new command to create non-package project skeletons without package structure or tests.

Enhancements:
- Update project layout generation to conditionally omit package directories, tests, packages configuration, and build-system sections when package mode is disabled.

Tests:
- Add init command tests covering non-package behavior, interaction with dependencies, interactive flows, and regression for standard package mode.
- Add new command tests verifying non-package project structure, pyproject.toml content, behavior with custom names, existing directories, and regression for standard behavior.

### DIFF
--- a/src/poetry/console/commands/init.py
+++ b/src/poetry/console/commands/init.py
@@ -58,6 +58,12 @@ class InitCommand(Command):
             multiple=True,
         ),
         option("license", "l", "License of the package.", flag=False),
+        option(
+            "non-package",
+            None,
+            "Initialize the project in non-package mode"
+            " (sets package-mode = false in pyproject.toml).",
+        ),
     ]
 
     help = """\
@@ -98,6 +104,7 @@ The <c1>init</c1> command creates a basic <comment>pyproject.toml</> file in the
         from poetry.pyproject.toml import PyProjectTOML
 
         is_interactive = self.io.is_interactive() and allow_interactive
+        non_package = self.option("non-package")
 
         pyproject = PyProjectTOML(project_path / "pyproject.toml")
 
@@ -130,7 +137,7 @@ The <c1>init</c1> command creates a basic <comment>pyproject.toml</> file in the
         if not name:
             name = project_path.name.lower()
 
-            if is_interactive:
+            if is_interactive and not non_package:
                 question = self.create_question(
                     f"Package name [<comment>{name}</comment>]: ", default=name
                 )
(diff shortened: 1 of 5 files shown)


radoering (commented):
I wonder whether we should really skip name, version, description, author, license, and python versions in interactive non-package mode? Python versions should be relevant even in non-package mode. Regarding the other fields, you can only guess if the user wants to specify them or not.

PR #10787, opened 2026-03-28
Update init & new commands for PEP 639 (License)

dunkmann00:
# Pull Request Check List

Resolves: N/A

- [x] Updated **tests** for changed code.

Poetry was already updated to support this in
https://github.com/python-poetry/poetry/pull/10413. But the layout needed to be updated for it to get generated in the `init` and `new` commands.

This also updates the relevant tests.

## Summary by Sourcery

Align project layout and init/new command outputs with PEP 639 license field expectations.

Bug Fixes:
- Update generated pyproject.toml license field to use a simple string instead of a nested table for license text.
- Adjust interactive init prompt to remove the empty default indicator for the license question.

Tests:
- Update init and new command tests to assert the new license string format in generated pyproject.toml files.

### DIFF
--- a/src/poetry/console/commands/init.py
+++ b/src/poetry/console/commands/init.py
@@ -166,7 +166,7 @@ The <c1>init</c1> command creates a basic <comment>pyproject.toml</> file in the
 
         license_name = self.option("license")
         if not license_name and is_interactive:
-            license_name = self.ask(self.create_question("License []: ", default=""))
+            license_name = self.ask(self.create_question("License: ", default=""))
 
         python = self.option("python")
         if not python:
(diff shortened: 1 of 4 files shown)


src/poetry/console/commands/init.py
```diff
@@ -166,7 +166,7 @@ def _init_pyproject(
 
         license_name = self.option("license")
         if not license_name and is_interactive:
-            license_name = self.ask(self.create_question("License []: ", default=""))
```
radoering, line 169: I think this still makes sense because `[]` just means that there is no default. E.g., it is also `Description []:`

dunkmann00: That makes sense, my mistake, I'll put it back!

radoering (approved):

PR #10797, opened 2026-03-29
fix: ensure consistent column widths in python list command

BlocksecPHD:
Set explicit minimum column widths for Version, Implementation, and Manager columns to ensure consistent formatting across all Python versions, including Python 3.14 free-threaded builds.

Fixes #10796

**Issue:** Test `test_list_poetry_managed[False]` fails on Python 3.14.3 (free-threaded) due to whitespace formatting. The test expects format like `"3.10.8  PyPy           Poetry"` but gets `"3.10.8   PyPy           Poetry"` (extra space before PyPy).

**Root Cause:** The column width calculation in cleo's Table class appears to calculate differently on Python 3.14, resulting in an extra space being added to column padding.

**Fix:** Added explicit minimum column widths using `table.set_column_width()`:
- Version column: 7 characters
- Implementation column: 14 characters  
- Manager column: 7 characters

This ensures consistent formatting by using known minimums rather than relying on dynamic calculation.

**Testing:** All 15 tests in `tests/console/commands/python/test_python_list.py` pass.

## Summary by Sourcery

Bug Fixes:
- Fix whitespace-sensitive output formatting for the python list command by enforcing minimum column widths for key columns.

### DIFF
--- a/src/poetry/console/commands/python/list.py
+++ b/src/poetry/console/commands/python/list.py
@@ -95,6 +95,13 @@ class PythonListCommand(Command):
                 "<fg=magenta;options=bold>Path</>",
             ]
         )
+        # Set minimum column widths to ensure consistent formatting across Python versions
+        # Version: max len is "X.Y.Zt" (7 chars) or "Version" (7 chars)
+        # Implementation: max len is "Implementation" (14 chars)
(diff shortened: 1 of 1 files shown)


radoering:
Thanks for your contribution. However, this is just an issue of the test that should be fixed in the test.

Superseded by #10817

PR #10816, opened 2026-04-01
fix: use SeparateBodyFileCache for file caching

JPK64:
Use `SeparateBodyFileCache` instead of `FileCache` for caching. `FileCache` attempts to serialize the HTTP response body along with the rest of the response which can cause a `ValueError: memoryview is too large` for files larger than 2 GiB. `cachecontrol` provides `SeparateBodyFileCache` which is better suited for caching large responses and does not cause this issue.

# Pull Request Checklist

Resolves: #10814

## Summary by Sourcery

Bug Fixes:
- Prevent failures when caching HTTP responses for files larger than 2 GiB by avoiding serialization of large response bodies.

### DIFF
--- a/src/poetry/utils/authenticator.py
+++ b/src/poetry/utils/authenticator.py
@@ -119,7 +119,7 @@ class Authenticator:
             SeparateBodyFileCache(
                 self._config.repository_cache_directory
                 / (cache_id or "_default_cache")
-                / "_http"
+                / "_http2"
             )
             if not disable_cache
             else None
--- a/src/poetry/utils/authenticator.py
+++ b/src/poetry/utils/authenticator.py
@@ -119,7 +119,7 @@ class Authenticator:
             SeparateBodyFileCache(
                 self._config.repository_cache_directory
                 / (cache_id or "_default_cache")
-                / "_http2"
+                / "_http_"
             )
             if not disable_cache
             else None
--- a/src/poetry/utils/authenticator.py
+++ b/src/poetry/utils/authenticator.py
@@ -115,6 +115,10 @@ class Authenticator:
             dict[str, AuthenticatorRepositoryConfig] | None
         ) = None
         self._password_manager = PasswordManager(self._config)
+
+        # Poetry < 2.4: FileCache -> directory "_http"
+        # Poetry >= 2.4: SeparateBodyCache -> directory "_http_"
+        # See https://github.com/python-poetry/poetry/pull/10816 for details.
         self._cache_control = (
             SeparateBodyFileCache(
                 self._config.repository_cache_directory
@@ -124,6 +128,7 @@ class Authenticator:
             if not disable_cache
             else None
         )
+
         self.get_repository_config_for_url = functools.lru_cache(maxsize=None)(
             self._get_repository_config_for_url
         )
--- a/src/poetry/utils/authenticator.py
+++ b/src/poetry/utils/authenticator.py
@@ -17,7 +17,7 @@ import requests.auth
 import requests.exceptions
 
 from cachecontrol import CacheControlAdapter
-from cachecontrol.caches import FileCache
+from cachecontrol.caches import SeparateBodyFileCache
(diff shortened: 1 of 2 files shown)


JPK64:
Switching the `Authenticator`s `_cache_control` from `FileCache` to `SeparateBodyFileCache` fixes #10814. However, it also breaks any already existing cache files that contain the body in the response instead of a separate file. What's more, it raises a `PoetryRuntimeError` stating that the package source is not reachable due to the "incomplete reads" that result from there not being a separate body file to read from.

Although all of these issues can be fixed by clearing the cache, I am not keen on breaking the caches of an entire user base. I could probably implement some kind of cache converter that converts from the old cache format to the new cache format, or implement a proxy that uses the correct cache implementation for reading from the cache and just the `SeparateBodyFileCache` for writing to the cache. I am not sure which approach is best, which is why I first created this pull request as a draft to get some input on how to handle this gracefully.

radoering:
> However, it also breaks any already existing cache files that contain the body in the response instead of a separate file. What's more, it raises a PoetryRuntimeError stating that the package source is not reachable due to the "incomplete reads" that result from there not being a separate body file to read from.

What are the steps to reproduce this issue? In a quick test, Poetry seems to lock fine after switching from `FileCache` to `SeparateBodyCache` with an existing cache.

JPK64:
> What are the steps to reproduce this issue?

Unfortunately, I am no longer able to reproduce the issue myself. Like you described, it seems to work fine. But I do get the behaviour I described when doing the reverse, so creating cache entries with the `SeparateBodyFileCache` and then trying to use those with the `FileCache`. So maybe I got my `poetry` versions mixed up during initial testing?

I'll mark this pull request as ready for review.

radoering:
Thus, the upgrade path seems fine but there is an issue if a user downgrades again or switches back and forth between different versions of Poetry.

I think we should change the cache suffix from `_http` to `_http2` for example. That way both Poetry versions will continue to work and the only "issue" is that there are unused files in the cache.

JPK64:
> I think we should change the cache suffix from `_http` to `_http2` for example. That way both Poetry versions will continue to work and the only "issue" is that there are unused files in the cache.

I have implemented the change you proposed and verified that this fixes the issue.

Secrus:
Sorry, I know I am late to the party, but `_http2` can be understood as HTTP/2 (HTTP v2), which we don't support. How about `_http_cache2`? @radoering WDYT?

radoering:
> Sorry, I know I am late to the party, but `_http2` can be understood as HTTP/2 (HTTP v2), which we don't support. How about `_http_cache2`? @radoering WDYT?

Yeah, I have thought about that, too. But I proposed `_http2` anyway because:

- This directory is inside `cache/repositories/<repository-name>` where normal users do not do anything and even in there it is prefixed with an underscore making it even more internal. (Nevertheless, I have to admit it could confuse contributors.)
- I fear long path issues on Windows if we make it unnecessarily long. It is difficult to guess if we are close to the limit on some systems so I wanted to keep it as short as possible.

I am not even sure if it is worth to have an incrementing number considering this is the first time we change something about it. We could even name it `_http_`.

Doing a quick (rough) calculation, I think we _should_ still be far away from the Windows path limit so that we could also name it `_http_cache2`.

@Secrus Having heard my thoughts, what do you prefer now? `http_cache2`, `_http_`, `_http2` or some new ideas?

Secrus:
> @Secrus Having heard my thoughts, what do you prefer now? `http_cache2`, `_http_`, `_http2` or some new ideas?

I am ok with either `http_cache2` or `_http_`. Anything that is unambiguous.

JPK64:
> I am ok with either http_cache2 or _http_.

I have changed the suffix to `_http_` for now because of what @radoering mentioned:

> I fear long path issues on Windows if we make it unnecessarily long. It is difficult to guess if we are close to the limit on some systems so I wanted to keep it as short as possible.

radoering (approved):

Issue #10028, opened 2025-01-13
`requires-plugins` breaks poetry installation on Windows 


pseusys:
### Description

I have a poetry project, which I install and test on Windows machine in GitHub Actions.
Recently, I tried using the new `requires-plugins` feature for `poethepoet` and `poetry-plugin-export` plugins installation.
The `pyproject.toml` file looks somewhat like this:
```toml
...
[tool.poetry.requires-plugins]
poethepoet = "^0.32.1"
poetry-plugin-export = "^1.9.0"
...
```

However, instead of expected installation, what I receive is an error:
```txt
Ensuring that the Poetry plugins required by the project are available...
The following Poetry plugins are required by the project but are not installed in Poetry's environment:
  - poethepoet (>=0.32.1,<0.33.0)
  - poetry-plugin-export (>=1.9.0,<2.0.0)
Installing Poetry plugins only for the current project...
Updating dependencies
Resolving dependencies...

Package operations: 4 installs, 0 updates, 0 removals

  - Installing pastel (0.2.1)
  - Installing pyyaml (6.0.2)
  - Installing poethepoet (0.32.1)
  - Installing poetry-plugin-export (1.9.0)

  ValueError

  path is on mount 'C:', start on mount 'D:'

  at C:\hostedtoolcache\windows\Python\3.10.11\x64\lib\ntpath.py:747 in relpath
      743|         path_abs = abspath(normpath(path))
      744|         start_drive, start_rest = splitdrive(start_abs)
      745|         path_drive, path_rest = splitdrive(path_abs)
      746|         if normcase(start_drive) != normcase(path_drive):
    > 747|             raise ValueError("path is on mount %r, start on mount %r" % (
      748|                 path_drive, start_drive))
      749| 
      750|         start_list = [x for x in start_rest.split(sep) if x]
      751|         path_list = [x for x in path_rest.split(sep) if x]

Cannot install poethepoet.


Failed to install required Poetry plugins
```

### Workarounds

The plugins still can be installed traditionally, using `poetry self add ...`.
However, I still would like to try the new feature... it works on Linux so nice!

### Poetry Installation Method

pip

### Operating System

`windiws-latest` (that is `Windows Server 2022 OS Version: 10.0.20348 Build 2966`)

### Poetry Version

Poetry (version 2.0.1)

### Poetry Configuration

```bash session
cache-dir = "C:\\Users\\runneradmin\\AppData\\Local\\pypoetry\\Cache"
installer.max-workers = null
[...]

PR #10819, opened 2026-04-03
installer: validate records

dimbleby:
installer 1.0

Almost three years since making the fix, this release includes https://github.com/pypa/installer/pull/186 - which was the reason for disabling record validation.


radoering:
I wonder how much time record validation costs and if we should make it optional. (I assume pip still does no validation?)

dimbleby:
https://github.com/pypa/installer/issues/327 should mostly mitigate extra expense I expect.

Presumably it still cannot be completely free to pass through large files twice, but I guess that the hash calculation dominates.

dimbleby:
some ad hoc benchmarking, installing a torch wheel and many gigabytes of nvidia-cu* dependencies.  Fully populated poetry cache, so we should be measuring only the time taken to copy from the cached wheel to the venv

- skip record validation, do not trust record: 36s
- validate record, do not trust record: 73s
- validate record, trust record: 60s

all on my aging laptop, your mileage may vary.

Not so much of a mitigation as I had hoped to be honest.  Perhaps the installer API still is wrong, perhaps it would be better to have `install()` be capable of optionally reporting mismatch between the record entries that it calculates and those in the wheel.

Then we would skip content validation up-front, but get it during install.

onyx-and-iris:
installer 1.0 also includes a fix for relpath on windows, see #10028.

PR #10871, opened 2026-05-02
simplify FileConfigSource.remove_property()

dimbleby:
Just a little simplification of some ugly code that I bumped into


radoering (approved):
nice

PR #10873, opened 2026-05-02
dont sign commits in tests

dimbleby:
I saw tests failing because I was being asked to provide a password to allow signing.


radoering (approved):
Thanks. I also noticed this issue but was too lazy to investigate which test was the culprit.

PR #10889, opened 2026-05-10
feat(init): support multiple --author values for `poetry init`

### LINKED ISSUE #8864: `poetry init --author` doesn't seem to allow multiple authors on Windows

quick123-666:
## Description

Closes #8864

### Problem

`poetry init --author` only accepts a single author value. Users who want to specify multiple authors (e.g., for projects with multiple maintainers) have no CLI option to do so.

### Solution

Changed the `--author` option to accept multiple values by adding
`multiple=True`, so users can now run:

```
poetry init --author 'Author One <one@example.com>' --author 'Author Two <two@example.com>'
```

**Changes:**

1. **`src/poetry/console/commands/init.py`**: Added `multiple=True` to the `--author` option definition. Updated `_init_pyproject()` to handle a list of authors from CLI, with interactive support for adding multiple authors in a loop.
2. **`src/poetry/layouts/layout.py`**: Changed constructor parameter from `author` (single string) to `authors` (list of strings). Updated `generate_project_content()` to iterate over all authors and append each one to the pyproject.toml authors list.
3. **`tests/console/commands/test_init.py`**: Added test for `--author` with multiple values. Also added test for the interactive multi-author flow.
4. **`tests/conftest.py`**: Updated Layout call to use new `authors` parameter.

### Usage

Non-interactive:
```bash
poetry init --author 'Alice <alice@example.com>' --author 'Bob <bob@example.com>' --name my-package --python '>=3.9'
```

Interactive:
```
Author [n to skip]: Alice <alice@example.com>
Add another author? (leave blank to skip): Bob <bob@example.com>
Add another author? (leave blank to skip):
```

### DIFF
--- a/src/poetry/console/commands/init.py
+++ b/src/poetry/console/commands/init.py
@@ -39,7 +39,7 @@ class InitCommand(Command):
     options: ClassVar[list[Option]] = [
         option("name", None, "Name of the package.", flag=False),
         option("description", None, "Description of the package.", flag=False),
-        option("author", None, "Author name of the package.", flag=False, multiple=True),
+        option("author", None, "Author names of the package.", flag=False, multiple=True),
         option("python", None, "Compatible Python versions.", flag=False),
         option(
             "dependency",
--- a/src/poetry/console/commands/init.py
+++ b/src/poetry/console/commands/init.py
@@ -39,7 +39,7 @@ class InitCommand(Command):
     options: ClassVar[list[Option]] = [
         option("name", None, "Name of the package.", flag=False),
         option("description", None, "Description of the package.", flag=False),
-        option("author", None, "Author name of the package.", flag=False),
+        option("author", None, "Author name of the package.", flag=False, multiple=True),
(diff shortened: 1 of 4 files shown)


dimbleby:
There are now 4 (four) pull requests making this same fix.  Please explain how this one is better than the others.

I guess you all are pointing your bots at the issue list and finding the same candidates?  Maybe include an instruction about "and does not already have an open pull request".

PR #10898, opened 2026-05-16
feat: add support for project-specific min-release-age configuration

finswimmer:
# Pull Request Check List

Poetry 2.4 introduces options to cool down dependency resolutions. These options are only available as configuration options for the Poetry installation itself.

If working on multiple projects especially in a corporate environment it might be necessary to have these options available on a per-project base. This is also useful if renovate or dependabot are used to update the dependencies.

For now i put these option under `[tool.poetry.config]`. But maybe we want to put them directly under `[tool.poetry]` (without `solver.` prefix)?

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -437,6 +437,19 @@ If a source does not provide upload times for a release,
 that release is not filtered out by this setting.
 {{% /note %}}
 
+This setting can also be configured in your `pyproject.toml`:
+
+```toml
+[tool.poetry.config]
+solver.min-release-age = 7
+```
+
+Values set in `[tool.poetry.config]` take precedence over
+[project-local configuration]({{< relref "#project-local-configuration" >}})
+and global configuration.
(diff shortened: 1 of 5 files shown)


radoering:
> For now i put these option under `[tool.poetry.config]`. But maybe we want to put them directly under `[tool.poetry]` (without `solver.` prefix)?

I think the main question is whether we assume that this is just a special case or whether we expect more config settings to be allowed in pyproject.toml in the future. In the first case, it probably does not matter much where we put it. In the latter case, it probably makes sense to have the `tool.poetry.config` entry point and the same structure as in the config files.

finswimmer:
> > For now i put these option under `[tool.poetry.config]`. But maybe we want to put them directly under `[tool.poetry]` (without `solver.` prefix)?
> 
> I think the main question is whether we assume that this is just a special case or whether we expect more config settings to be allowed in pyproject.toml in the future.

That are my thoughts as well. I tend to say this is a special case. But at the end of the day I have no strong opinion here.

PR #10904, opened 2026-05-17
feat: add builtin uninstaller as alternative to calling pip

radoering:
# Pull Request Check List

Introduce a builtin uninstaller (adapted from pip's code) so that we do not have to run a pip subprocess for each uninstall. This makes uninstalls (and updates) faster. The new builtin uninstaller is not used by default. You have to set `installer.builtin-uninstall` to use it. (We can switch it later and remove the setting even later, just as we did with `installer.modern-installation`.)

The second commit makes use of the rollback mechanism in case of an error. Further, it fixes an issue in the legacy pip path where the return code of pip is not checked.

### DIFF
--- a/src/poetry/installation/executor.py
+++ b/src/poetry/installation/executor.py
@@ -21,6 +21,7 @@ from poetry.installation.chooser import Chooser
 from poetry.installation.operations import Install
 from poetry.installation.operations import Uninstall
 from poetry.installation.operations import Update
+from poetry.installation.uninstaller import uninstall_distribution
 from poetry.installation.wheel_installer import WheelInstaller
 from poetry.puzzle.exceptions import SolverProblemError
 from poetry.utils._compat import decode
@@ -78,6 +79,9 @@ class Executor:
         self._verbose = False
         self._wheel_installer = WheelInstaller(self._env)
         self._build_constraints = build_constraints or {}
+        self._use_builtin_uninstall: bool = config.get(
+            "installer.builtin-uninstall", False
+        )
 
         if parallel is None:
             parallel = config.get("installer.parallel", True)
@@ -627,6 +631,12 @@ class Executor:
             src_dir = self._env.path / "src" / package.name
             if src_dir.exists():
                 remove_directory(src_dir, force=True)
+
+        if self._use_builtin_uninstall:
+            pathset = uninstall_distribution(self._env, package.name)
+            if pathset is not None:
+                pathset.commit()
+            return 0
 
         try:
             return self.run_pip("uninstall", package.name, "-y")
--- a/src/poetry/installation/executor.py
+++ b/src/poetry/installation/executor.py
@@ -21,6 +21,7 @@ from poetry.installation.chooser import Chooser
 from poetry.installation.operations import Install
 from poetry.installation.operations import Uninstall
 from poetry.installation.operations import Update
+from poetry.installation.uninstaller import UninstallPathSet
 from poetry.installation.uninstaller import uninstall_distribution
 from poetry.installation.wheel_installer import WheelInstaller
 from poetry.puzzle.exceptions import SolverProblemError
@@ -607,15 +608,27 @@ class Executor:
         )
         self._write(operation, message)
 
+        old_pathset: UninstallPathSet | None = None
         try:
             if operation.job_type == "update":
-                # Uninstall first
-                # TODO: Make an uninstaller and find a way to rollback in case
-                # the new package can't be installed
                 assert isinstance(operation, Update)
-                self._remove(operation.initial_package)
+                result = self._uninstall(operation.initial_package)
+                if isinstance(result, UninstallPathSet):
+                    old_pathset = result
+                elif result != 0:
+                    # pip uninstall was interrupted; surface that code
+                    # instead of attempting the install.
+                    return result
 
-            self._wheel_installer.install(archive)
+            try:
+                self._wheel_installer.install(archive)
+            except BaseException:  # rollback also on KeyboardInterrupt
+                if old_pathset is not None:
+                    old_pathset.rollback()
+                raise
+
+            if old_pathset is not None:
+                old_pathset.commit()
         finally:
             if cleanup_archive:
                 archive.unlink()
@@ -626,7 +639,25 @@ class Executor:
         return self._install(operation)
 
     def _remove(self, package: Package) -> int:
-        # If we have a VCS package, remove its source directory
+        result = self._uninstall(package)
+        if isinstance(result, int):
+            return result
+        result.commit()
+        return 0
+
+    def _uninstall(self, package: Package) -> int | UninstallPathSet:
+        """Stash an installed package's files.
+
+        Returns either an ``UninstallPathSet`` (builtin installer)
+        so the caller can ``.commit()`` to finalize the removal or
+        ``.rollback()`` to restore the files, or an ``int`` return code
+        (legacy pip path, or builtin path when nothing needs uninstalling).
+        ``-2`` means pip was interrupted; ``0`` means success.
+
+        The git VCS source directory (``<env>/src/<pkg>/``) is removed
+        eagerly and is not part of the returned pathset; it is not
+        restored on rollback.
+        """
         if package.source_type == "git":
             src_dir = self._env.path / "src" / package.name
             if src_dir.exists():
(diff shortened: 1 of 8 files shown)


src/poetry/installation/uninstaller.py
```diff
@@ -0,0 +1,473 @@
...
+
+    head, tail = os.path.split(old)
+    if head and tail:
+        try:  # noqa: SIM105 (performance)
+            os.removedirs(head)
+        except OSError:
+            pass
+
+
+class _TempDirectory:
```
dimbleby, line 73: `from tempfile import TemporaryDirectory` and drop this?

src/poetry/installation/uninstaller.py
```diff
@@ -0,0 +1,473 @@
...
+        path = os.path.join(location, str(entry))
+        yield path
+        if path.endswith(".py"):
+            dn, fn = os.path.split(path)
+            base = fn[:-3]
+            yield os.path.join(dn, base + ".pyc")
+            yield os.path.join(dn, base + ".pyo")
+
+
+def compact(paths: Iterable[str]) -> set[str]:
```
dimbleby: my AI friend claims that this is redundant and that compress_for_rename() already returns a minimal, non-overlapping set:

> - wildcards are added only for roots whose every on-disk file is in the
      input, and roots are processed shortest-first, so a wildcard is never
      nested under another wildcard;
>  - a file remains in the 'remaining' set only if its parent (which is
      always in 'unchecked') did not get a wildcard, so it cannot be under
      any wildcard either.

src/poetry/installation/uninstaller.py
```diff
@@ -0,0 +1,473 @@
...
+        for item in fn(*args, **kw):
+            if item not in seen:
+                seen.add(item)
+                yield item
+
+    return unique
+
+
+@_unique
+def _uninstallation_paths(
```
dimbleby, line 143: Could consider dropping `_unique` altogether: this is the only place it is used, and the only consumer of `_uninstallation_paths` is `UninstallPathSet` - which tolerates duplicates anyway.

src/poetry/installation/uninstaller.py
```diff
@@ -0,0 +1,473 @@
...
+    @property
+    def path(self) -> str:
+        return self._path
+
+    def cleanup(self) -> None:
+        if os.path.exists(self._path):
+            shutil.rmtree(self._path, ignore_errors=True)
+
+
+class _AdjacentTempDirectory(_TempDirectory):
```
dimbleby, line 93: pretty sure we could drop this without causing damage.  AIUI the point is to stay on the same filesystem as the original as some sort of perf optimisation?

Looks like pdm has been through the same copy-from-pip-and-remove-legacy-noise approach - [here](https://github.com/pdm-project/pdm/blob/main/src/pdm/installers/uninstallers.py) - and they do not have an `_AdjacentTempDirectory`

src/poetry/installation/executor.py
```diff
@@ -622,19 +639,43 @@ def _update(self, operation: Install | Update) -> int:
...
 
     def _remove(self, package: Package) -> int:
-        # If we have a VCS package, remove its source directory
+        result = self._uninstall(package)
+        if isinstance(result, int):
+            return result
+        result.commit()
+        return 0
+
+    def _uninstall(self, package: Package) -> int | UninstallPathSet:
```
dimbleby: suggestion: return `tuple[int, UninstallPathSet | None]`.  Slightly more convenient for callers, who will not need to check `isinstance()`

radoering:
For future reference: We probably also have to patch `importlib-metadata` because of https://github.com/python/importlib_metadata/issues/535

Otherwise, script files are not uninstalled on Windows with Python 3.10 and 3.11. This is even an issue, although we do not use `importlib-metadata` directly anymore because it is a transient dependency and replaces the original `importlib.metadata.DistributionFinder` in `sys.meta_path` and thereby the `Distribution` class that is returned when searching for a distribution. pip does not seem to suffer from this issue, either because it does not depend on a recent enough version of importlib-metadata or due to the various legacy branches and fallbacks I removed when adopting the code.

src/poetry/installation/uninstaller.py
radoering: My AI friend said:

```
Short answer: No, it's not redundant - though in the common case it's effectively a no-op.

  Here's why. remove() calls them in sequence at uninstaller.py:400-402:

  for_rename = compress_for_rename(self._paths)
  for path in sorted(compact(for_rename)):
      moved.stash(path)

  compress_for_rename() already does most of the compaction work: when every file on disk under a directory is in the
  removal set, it collapses them into a single dir/ wildcard and drops the individual files from remaining
  (uninstaller.py:226-228). And its guard at line 214 prevents nested wildcards (roots are processed shortest-first, so a
  child dir under an already-wildcarded parent is skipped). So for an all-regular-files package, compress_for_rename returns
   a minimal set and compact() changes nothing.

  The gap compact() covers: compress_for_rename only subtracts entries that os.walk reports as files (all_files, line 222).
  Anything os.walk classifies as a subdir - most notably a symlink to a directory, or an empty directory listed in RECORD -
  lands in all_subdirs instead and is never removed from remaining. So you can get output like {/pkg/, /pkg/symlinked_dir}:
  the parent wildcard plus a child entry it failed to subtract.

  That's a real problem without compact(), because of the sort order at line 402: /pkg/ sorts before /pkg/symlinked_dir, so
  remove() would stash the whole /pkg/ directory first, then try to stash /pkg/symlinked_dir \u2014 which no longer exists at its
   original location, causing _renames() to fail. compact() (uninstaller.py:180-196) drops the child because it's prefixed
  by the wildcard parent, preventing the double-stash.

  So: redundant for the typical RECORD of plain files, but a genuine safety net for directory-typed entries (symlinked/empty
   dirs). It also matches pip's original code, which the module comment says it deliberately mirrors.
```

That may be a theoretical issue. I do not think that such wheels are possible (yet). However, it cannot hurt to be robust. Nevertheless, I think only a minimal edit in `compress_for_rename()` is necessary to make compact truly redundant. I added the change (and a test) in a separate commit for now.

src/poetry/installation/uninstaller.py
radoering: Actually, it makes sense to me. In cases where the temporary directory is on another filesystem it should make a significant difference, shoudn't it?

I would give more weight to pip than to pdm in this case because of its broader usage.

dimbleby: > ... it should make a significant difference, shoudn't it?

I don't know.  Maybe?  

Another comparison point: as best I can tell `uv` does not bother with any of this at all.  It [just removes files](https://github.com/astral-sh/uv/blob/81dabd532d7d35520ff62f45df1e562f3fe33e9f/crates/uv-install-wheel/src/uninstall.rs#L15-L97).

That's going to win on simplicity and presumably also on performance, is the pip commit/rollback dance even useful?!

dimbleby: Perhaps the pip behaviour dates from a time before virtual environments were so universal, and disposable.

I speculate that the case uv might make for not taking the extra care is: in the rare case that the uninstall leaves your venv in an halfway state - just delete it and start over.  So long as we reported that something went wrong, we don't have to work so hard to avoid that happening.

dimbleby: I did some archaeology in the pip repo:

- the move/rollback design dates to [2009](https://github.com/pypa/pip/commits/9c2da54c9)
  - of course people had good ideas in 2009 also!  But it's at least plausible that whatever drove pip back then is not so relevant for poetry today
- `AdjacentTemporaryDirectory` came in 2018 and the motivation was not performance at all, rather it was to work around accidentally using too-long paths on Windows - https://github.com/pypa/pip/pull/6029

radoering: This is a good point. On a second thought, I agree that we probably do not need the commit/rollback behavior, which adds some complexity. I have to admit, this feature initially confused me when investigating how pip uninstalls packages. Thus, it is probably a good thing to omit if we do not need it.

In my experience, the most common case where an uninstall fails, is on Windows when another process is blocking a file that should be removed. In that case, it should be fine to error out. If we make sure that the .dist-info directory is the last one to delete, the user can even kill the process and try again.

I will open a PR with an alternative implementation shortly.

PR #10907, opened 2026-05-17
perf/tests: use more appropriate data type for package filter policy packages and add tests

dhimasardinata:
`PackageFilterPolicy` normalizes `installer.no-binary`, `installer.only-binary`, and related package-filter settings once, but `allows()` and `has_exact_package()` previously scanned the normalized sequence each time they were called.

Chooser policy checks run for every candidate link, so this stores the normalized package names as an immutable `frozenset` for repeated membership checks.

A small local benchmark over a 1,001-entry normalized package policy:

```text
old=0.234734s
new=0.006853s
speedup=34.3x
```

# Pull Request Check List

- [x] Updated **documentation** for changed code. N/A, no user-facing behavior change.

Local checks:

```bash
PYTHONPATH=src uv run --with pytest --with pytest-xdist --with responses --with pytest-mock --with deepdiff python -m pytest tests/config/test_config.py tests/installation/test_chooser.py -q
uv run --with ruff python -m ruff check src/poetry/config/config.py tests/config/test_config.py
uv run --with ruff python -m ruff format --check src/poetry/config/config.py tests/config/test_config.py
PYTHONPATH=src uv run --with poetry-core --with packaging --with mypy python -m mypy src/poetry/config/config.py
git diff --check
```

### DIFF
--- a/src/poetry/config/config.py
+++ b/src/poetry/config/config.py
@@ -83,14 +83,14 @@ def build_config_setting_normalizer(val: str) -> Mapping[str, str | Sequence[str
 @dataclasses.dataclass
 class PackageFilterPolicy:
     policy: dataclasses.InitVar[str | list[str] | None]
-    packages: list[str] = dataclasses.field(init=False)
+    packages: frozenset[str] = dataclasses.field(init=False)
 
     def __post_init__(self, policy: str | list[str] | None) -> None:
         if not policy:
             policy = []
         elif isinstance(policy, str):
             policy = self.normalize(policy)
-        self.packages = policy
+        self.packages = frozenset(policy)
 
     def allows(self, package_name: str) -> bool:
         if ":all:" in self.packages:
--- a/tests/config/test_config.py
+++ b/tests/config/test_config.py
@@ -13,6 +13,7 @@ import pytest
 from deepdiff.diff import DeepDiff
 
 from poetry.config.config import Config
+from poetry.config.config import PackageFilterPolicy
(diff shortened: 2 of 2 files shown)


dhimasardinata:
Addressed Sourcery feedback in a0a8f85 by storing normalized packages as an immutable tuple and using a frozenset for repeated membership checks.

src/poetry/config/config.py
```diff
@@ -83,27 +83,29 @@ def build_config_setting_normalizer(val: str) -> Mapping[str, str | Sequence[str
 @dataclasses.dataclass
 class PackageFilterPolicy:
     policy: dataclasses.InitVar[str | list[str] | None]
-    packages: list[str] = dataclasses.field(init=False)
+    packages: tuple[str, ...] = dataclasses.field(init=False)
```
dimbleby: simpler just to store packages as a frozenset in the first place, no need for the duplication?

dhimasardinata: Done in 2b2a547. `PackageFilterPolicy` now stores the normalized packages directly as a `frozenset`, so membership checks use a single immutable field without duplication.

radoering:
> A small local benchmark over a 1,001-entry normalized package policy:

I doubt that real numbers become that high. I assume that 100 is quite much and most of the times it will be less than 10. Nevertheless, frozenset feels like the more appropriate data type.

radoering (approved):

PR #10913, opened 2026-05-19
Fix contributing guide label links

Mirochill:
## Summary

Fix the contributing guide links for the `kind/question` and `area/docs` GitHub labels.

The label names contain `/`, so the label URLs need `%2F` encoding. The `kind/question` link is kept inline inside the note shortcode so the generated website can render it correctly.

Fixes #10765.

## Validation

- `git diff --check HEAD~1..HEAD`
- GitHub API metadata checked for `kind%2Fquestion` and `area%2Fdocs`
- Remote checks green on `af9f0b1`: `Detect changed files`, `Status`, `Tests / FreeBSD Status`, `pre-commit.ci - pr`, and `Sourcery review`
- Full test matrix skipped as docs-only; `Documentation Preview` skipped on the follow-up push

### DIFF
--- a/docs/contributing.md
+++ b/docs/contributing.md
@@ -111,8 +111,7 @@ constantly evolving, and this means that sometimes our documentation has gaps. Y
 editing the existing content to be more accessible, or creating new content such as tutorials, FAQs, etc.
 
 {{% note %}}
-GitHub [Discussions](https://github.com/python-poetry/poetry/discussions) and the
-[kind/question label](https://github.com/python-poetry/poetry/labels/kind/question) are excellent sources for FAQ
+GitHub [Discussions] and the [kind/question label] are excellent sources for FAQ
 candidates.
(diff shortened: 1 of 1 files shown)


radoering (changes requested):
This does not seem to work. See the generated preview. Maybe, the link targets have to be inline in notes. :shrug: By the way, you can also build the docs locally with the [website repo](https://github.com/python-poetry/website).

Mirochill:
Thanks, fixed in af9f0b1 by keeping the links inside the note inline and only encoding the label slash in the GitHub URL.

The latest published checks are green: Detect changed files, Status, pre-commit.ci, Sourcery review, and FreeBSD. The Documentation Preview job was skipped on this follow-up push, so I do not have a refreshed preview URL to cite.

radoering (approved):

Issue #10633, opened 2025-11-28
where to specify the source for poetry self update


berg-a:
### Issue Kind

Missing documentation

### Existing Link

where to specify source for self update

### Description

Hello,
I have troubles with finding in the documentation where i can specify the source used for ´poetry self update´. The poetry update command uses the source specified in in pyproct.toml. It makes sense that the poetry self command ignores that. But since i work in an environment, where i can't access pypi directly i need to specify another source for it to work. But i can't find where i can specify that.

<img width="662" height="493" alt="Image" src="https://github.com/user-attachments/assets/107bdedb-451d-427f-855d-6eb3ca047056" />

radoering:
If I remember correctly, there will be a pyproject.toml in the [config directory](https://python-poetry.org/docs/configuration/#config-directory) after running `self update` (or `self lock`) for the first time. You can probably configure a source there.

berg-a:
Thank you for the reply. Yes this works out. unfortunately there is no `poetry self source add` command, to do it directly from the terminal. I am wondering what is the way to actual solve this issue:
documenting that the user has to run an command that fails to just create the pyprojecttoml in .config sounds like an improper fix for long term.

radoering:
Probably, just document when and where the pyproject.toml is created. Even for the project's pyproject.toml, there is not a corresponding command for each possible setting.

PR #10923, opened 2026-05-25
Document package sources for self update

Mirochill:
# Pull Request Check List

Resolves: #10633

- [ ] Added **tests** for changed code.
- [x] Updated **documentation** for changed code.

Summary:
- Document that `poetry self` commands use a system `pyproject.toml` in Poetry's configuration directory.
- Clarify that `poetry self update` reads package sources from that system project, not from the current project's `pyproject.toml`.
- Link the `[[tool.poetry.source]]` format to the package sources documentation.

Validation:
- `git diff --check`
- `rg -n "system project|configuration#config-directory|repositories/#package-sources|tool\.poetry\.source|current project.*self update|preserves existing" docs\cli.md`

Not run locally: documentation build, tests, or pre-commit.

PR #10930, opened 2026-05-30
docs(cli): document how to configure a custom source for "self update"

sridhar-3009:
## Summary

Addresses #10633.

Users in corporate or air-gapped environments cannot reach PyPI directly, so `poetry self update` fails without a way to specify a custom index. The documentation did not explain where or how to configure one.

## What was missing

`self update` manages Poetry's own runtime environment using a **separate** `pyproject.toml` in Poetry's config directory (e.g. `~/.config/pypoetry/pyproject.toml`), not the project's `pyproject.toml`. This is non-obvious and caused confusion.

## Change

Added a **"Custom sources for self update"** subsection under `### self update` that:

1. Explains that `self update` uses a separate `pyproject.toml` at Poetry's config directory.
2. Shows how to add a private source via `poetry self add --source`.
3. Mentions that the file can be edited directly like any other `pyproject.toml`.

## Test plan

- [ ] Render docs locally and verify the subsection appears correctly under `self update`

### DIFF
--- a/docs/cli.md
+++ b/docs/cli.md
@@ -1053,6 +1053,19 @@ is different in that the packages managed are for Poetry's runtime environment.
 poetry self update
 ```
 
+#### Custom sources for self update
+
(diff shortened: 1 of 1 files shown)


radoering:
already addressed sufficiently in #10923

PR #10944, opened 2026-06-06
fix(solver): prune sibling deps disjoint with override marker context

### LINKED ISSUE #5506: Solver breaks faced with related dependencies that are both conditional

dimbleby:
Another one that looks plausible, has a convincing testcase, is probably right...

---

When the solver retries under an override branch for a duplicate dependency, sibling dependencies whose markers are disjoint with the override's effective marker context were still considered. Their transitive constraints then spuriously conflicted with the overridden constraint, causing resolution to fail even when a valid solution existed (e.g. a transitive dep pinning a different version of the overridden package on a marker branch that does not apply).

Extend the existing overrides_marker_intersection mechanism into complete_package so such siblings are pruned during the solve phase, mirroring the filter already applied to duplicate deps.

Fixes #5506.

### DIFF
--- a/src/poetry/mixology/version_solver.py
+++ b/src/poetry/mixology/version_solver.py
@@ -2,6 +2,7 @@ from __future__ import annotations
 
 import collections
 import functools
+import itertools
 import time
 
 from enum import IntEnum
@@ -12,6 +13,7 @@ from poetry.core.packages.dependency import Dependency
 from poetry.mixology.failure import SolveFailureError
 from poetry.mixology.incompatibility import Incompatibility
 from poetry.mixology.incompatibility_cause import ConflictCauseError
+from poetry.mixology.incompatibility_cause import DependencyCauseError
 from poetry.mixology.incompatibility_cause import NoVersionsCauseError
 from poetry.mixology.incompatibility_cause import RootCauseError
 from poetry.mixology.partial_solution import PartialSolution
@@ -22,13 +24,26 @@ from poetry.packages import PackageCollection
 
 
 if TYPE_CHECKING:
+    from poetry.core.packages.package import Package
     from poetry.core.packages.project_package import ProjectPackage
+    from poetry.core.version.markers import BaseMarker
 
     from poetry.packages import DependencyPackage
     from poetry.puzzle.provider import Provider
 
 
 _conflict = object()
+
+
+def _effective_marker(dependency: Dependency) -> BaseMarker:
+    """
+    The marker describing when a dependency actually applies: its own marker
+    intersected with the marker of the package that required it (a dependency is
+    only reached when its requiring package applies).
+    """
+    return dependency.transitive_marker.without_extras().intersect(
+        dependency.marker.without_extras()
+    )
 
 
 class Preference(IntEnum):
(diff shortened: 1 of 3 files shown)


radoering:
Looks like a clear improvement. I just wonder if we should let it close #5506 because it only fixes the issue if one of both dependencies has multiple constraints. It does not fix the issue if both dependencies have just a single constraint.

For example, it fixes

```
dbt = { version = "0.19.2", markers = "sys_platform != 'darwin'" }
dbt-bigquery = [
    { version = "0.19.2", markers = "sys_platform != 'darwin'" },
    { version = "^1.0.0", markers = "sys_platform == 'darwin' "},
]
```

but not the derived example

```
dbt = { version = "0.19.2", markers = "sys_platform != 'darwin'" }
dbt-bigquery = { version = "^1.0.0", markers = "sys_platform == 'darwin'" },
```

dimbleby:
You are right.

My AI friend has a fix that addresses both but it is a little expensive in locking time.  (Nowhere near #6969 levels, but enough to notice on large solves.)

I will poke at that for a bit.  If you merge this first, or if I do not find a way to recover performance, I will create a separate pull request for consideration.  If everything can be made excellent, I will add a second commit here.

dimbleby:
pushed a supplementary - but more complicated - fix which:
- at solver fail, looks for the case where the conflict only happens in disjoint marker worlds, and could therefore do better by treating them separately
- raises `OverrideNeededError` in that case, using a synthetic package to split the world into two

radoering (approved):

PR #10952, opened 2026-06-13
docs: document `python` and `platform` dependency shorthands

### LINKED ISSUE #9183: Document `platform` and `python` shorthands for long-form dependencies
### Issue Kind

Improving documentation

### Existing Link

https://python-poetry.org/docs/dependency-specification/#using-environment-markers

### Description

The marker documentation in the dependency spec does not currently document the `python` and `platform` short-hands. We should document them, and make clear whether or not they should be used (or are deprecated); the combining semantics with `markers` also needs to be made clear.

We should also explain why they exist, whether or not future short-hand forms are likely, and provide more examples showcasing how to use the `markers` field for end users.

rumitvn:
Closes #9183

### What
Adds a **`python` and `platform` shorthands** subsection to the *Using environment markers* part of the dependency-specification docs.

### Why
The marker docs document the `markers` property but never document the `python` and `platform` shorthand keys available on long-form `[tool.poetry.dependencies]` entries, nor how they combine with an explicit `markers` value (the issue's three asks).

The new subsection covers:
- **`python`** → a version constraint on `python_version` (e.g. `python = "^3.9"` ⇒ `python_version >= "3.9" and python_version < "4.0"`).
- **`platform`** → an exact match on `sys_platform` (e.g. `platform = "darwin"`).
- That they are **not deprecated**, just shorthands, and that combining them with `markers` joins all conditions with `and` (with a worked example).
- That they are only available in the legacy `[tool.poetry.dependencies]` table; the PEP 621 `[project]` table uses inline PEP 508 markers instead.

Scoped to documenting current behavior. Docs-only change.

### DIFF
--- a/docs/dependency-specification.md
+++ b/docs/dependency-specification.md
@@ -586,7 +586,9 @@ shorthands for the most common environment markers:
   For example, `python = "^3.9"` is equivalent to the marker
   `python_version >= "3.9" and python_version < "4.0"`.
 - `platform` is an exact match on `sys_platform`. For example, `platform = "darwin"` is
-  equivalent to the marker `sys_platform == "darwin"`.
+  equivalent to the marker `sys_platform == "darwin"`. The value is the raw
+  [`sys.platform`](https://docs.python.org/3/library/sys.html#sys.platform) string
+  (commonly `darwin`, `win32`, or `linux`), not a friendly name like `macOS`.
 
 ```toml
 [tool.poetry.dependencies]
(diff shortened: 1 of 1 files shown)


rumitvn:
Thanks for the review @sourcery-ai!

- **`platform` values**: clarified that the value is the raw `sys.platform` string (commonly `darwin`, `win32`, `linux`) and not a friendly name like `macOS`, with a link to the CPython `sys.platform` docs.
- **PEP 621 `[project]` table**: the section already notes these shorthands are legacy-table-only and shows the equivalent inline PEP 508 marker for `[project]`, so readers have the preferred alternative inline.

Pushed in 563a5cb.

PR #10970, opened 2026-07-01
Delay parsing requests timeout environment variable

### LINKED ISSUE #10969: Invalid POETRY_REQUESTS_TIMEOUT can fail at import time instead of using config normalization
### Summary

Setting `POETRY_REQUESTS_TIMEOUT` to a non-integer value can make even unrelated commands such as `poetry --version` fail during import with a raw `ValueError` traceback.

### Reproduction

```bash
POETRY_REQUESTS_TIMEOUT=abc python -m poetry --version
```

### Actual behavior

The process fails during import with a traceback ending in:

```text
ValueError: invalid literal for int() with base 10: 'abc'
```

The failing line is the import-time conversion of the environment variable:

```python
REQUESTS_TIMEOUT = int(os.getenv("POETRY_REQUESTS_TIMEOUT", 15))
```

### Expected behavior

Invalid `POETRY_REQUESTS_TIMEOUT` values should ideally be reported as a Poetry configuration error that names the variable and expected type, rather than as an import-time traceback. Also, commands that do not use HTTP requests, such as `poetry --version`, probably should not fail because an HTTP timeout environment variable is malformed.

If the intended product semantics are that malformed environment variables are allowed to fail at import time, it would be helpful to state that explicitly. Otherwise, this looks inconsistent with Poetry's existing typed configuration normalization path.

### Why `int_normalizer` makes this look inconsistent

Poetry already has a configuration normalization mechanism for environment-derived settings. In `Config.get`, environment variables are looked up and passed through the appropriate normalizer:

```python
env = "POETRY_" + "_".join(k.upper().replace("-", "_") for k in keys)
env_value = os.getenv(env)
if env_value is not None:
    return self.process(self._get_normalizer(setting_name)(env_value))
```

For integer settings, Poetry explicitly uses `int_normalizer`:

```python
@staticmethod
def _get_normalizer(name: str) -> Callable[[str], Any]:
    if name in {
        "installer.max-workers",
        "requests.max-retries",
        "solver.lazy-wheel",
    }:
        return int_normalizer
```

That means config-backed integer values already have a single typed conversion point. `POETRY_REQUESTS_TIMEOUT`, however, bypasses this path and is converted directly in `poetry.utils.constants` at import time:

```python
REQUESTS_TIMEOUT = int(os.getenv("POETRY_REQUESTS_TIMEOUT", 15))
```

The parsed value is later consumed as an HTTP timeout, for example by request/authenticator code. The contract drift is that this timeout behaves like a typed numeric configuration, but it does not use the existing typed configuration normalizer and therefore fails with a raw import-time `ValueError`.

### Possible fixes

A few possible directions:

1. Move this setting into Poetry's normal config path and use the existing integer normalization mechanism.
2. Delay parsing until the timeout is actually needed by HTTP code.
3. Catch `ValueError` and raise a Poetry-facing configuration error such as `POETRY_REQUESTS_TIMEOUT must be an integer number of seconds`.

A regression test could set `POETRY_REQUESTS_TIMEOUT=abc` and assert that startup does not produce a raw traceback, especially for commands that do not perform HTTP requests.

---
Submitted with PilotDeck ([OpenBMB/PilotDeck](https://github.com/OpenBMB/PilotDeck)).

nightt5879:
## Summary
- Delay parsing `POETRY_REQUESTS_TIMEOUT` until an HTTP request actually needs the timeout value.
- Keep the default timeout constant import-safe and parse the environment value through Poetry's integer normalizer helper.
- Add regression coverage for `python -m poetry --version` with an invalid timeout environment value.

Fixes #10969

## Validation
- `POETRY_REQUESTS_TIMEOUT=abc .\.venv\Scripts\python.exe -m poetry --version` -> `Poetry (version 2.5.0.dev0)`
- `.\.venv\Scripts\python.exe -m pytest -n 0 --basetemp=.pytest-tmp tests\utils\test_constants.py tests\console\test_application.py::test_application_version_ignores_invalid_requests_timeout` -> 4 passed
- `.\.venv\Scripts\python.exe -m pytest -n 0 --basetemp=.pytest-tmp tests\utils\test_authenticator.py tests\repositories\test_http_repository.py tests\repositories\test_pypi_repository.py tests\publishing` -> 173 passed
- `uvx ruff check src\poetry\utils\constants.py src\poetry\utils\authenticator.py src\poetry\utils\helpers.py src\poetry\repositories\http_repository.py src\poetry\repositories\pypi_repository.py src\poetry\publishing\uploader.py tests\console\test_application.py tests\utils\test_constants.py` -> passed

## Notes
- Full `pytest --basetemp=.pytest-tmp` on this Windows/Python 3.13.13 environment reached 3183 passed / 11 skipped, with 10 existing environment-sensitive failures unrelated to this change. The failures involved Windows path URI expectations, active Python discovery preferring Anaconda Python, and missing embedded pip wheels from the locally resolved virtualenv package.

### DIFF
--- a/src/poetry/publishing/uploader.py
+++ b/src/poetry/publishing/uploader.py
@@ -23,7 +23,7 @@ from requests_toolbelt.multipart import MultipartEncoderMonitor
 
 from poetry.__version__ import __version__
 from poetry.publishing.hash_manager import HashManager
-from poetry.utils.constants import REQUESTS_TIMEOUT
+from poetry.utils.constants import get_requests_timeout
 from poetry.utils.patterns import wheel_file_re
 
 
@@ -239,7 +239,7 @@ class Uploader:
                         data=monitor,
                         allow_redirects=False,
                         headers={"Content-Type": monitor.content_type},
-                        timeout=REQUESTS_TIMEOUT,
+                        timeout=get_requests_timeout(),
                     )
                 if resp is None or 200 <= resp.status_code < 300:
                     bar.set_format(
@@ -308,7 +308,7 @@ class Uploader:
             data=encoder,
             allow_redirects=False,
             headers={"Content-Type": encoder.content_type},
-            timeout=REQUESTS_TIMEOUT,
+            timeout=get_requests_timeout(),
         )
 
         resp.raise_for_status()
--- a/src/poetry/repositories/http_repository.py
+++ b/src/poetry/repositories/http_repository.py
@@ -35,6 +35,7 @@ from poetry.repositories.exceptions import RepositoryError
 from poetry.repositories.link_sources.html import HTMLPage
 from poetry.repositories.link_sources.json import SimpleJsonPage
 from poetry.utils.authenticator import Authenticator
+from poetry.utils.constants import get_requests_timeout
 from poetry.utils.helpers import HTTPRangeRequestSupportedError
 from poetry.utils.helpers import download_file
 from poetry.utils.helpers import get_highest_priority_hash_type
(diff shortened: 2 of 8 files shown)


radoering:
I am not sure whether this is the way we want to go. Maybe, we should check if we can make this a proper config setting instead.

PR #10975, opened 2026-07-09
fix(init): sanitize default package name derived from directory name

### LINKED ISSUE #10974: `init` generates project with spaces in name
### Description

https://packaging.python.org/en/latest/specifications/name-normalization/#name-format

> A valid name consists only of ASCII letters and numbers, period, underscore and hyphen. 

```
c:\temp\my project with spaces>poetry init -n

c:\temp\my project with spaces>type pyproject.toml
[project]
name = "my project with spaces"
...
```

It should be `my-project-with-spaces`.

### Workarounds

no

pipx

### Operating System

Windows

### Poetry Version

2.4.1

### Poetry Configuration

```bash session
cache-dir = "C:\\Users\\usr\\AppData\\Local\\pypoetry\\Cache"
data-dir = "C:\\Users\\usr\\AppData\\Roaming\\pypoetry"
installer.max-workers = null
installer.no-binary = null
installer.only-binary = null
installer.parallel = true
installer.re-resolve = false
keyring.enabled = true
python.installation-dir = "{data-dir}\\python"  # C:\Users\usr\AppData\Roaming\pypoetry\python
requests.max-retries = 0
solver.lazy-wheel = true
solver.min-release-age = 0
solver.min-release-age-exclude = null
solver.min-release-age-exclude-source = null
system-git-client = false
virtualenvs.create = true
virtualenvs.in-project = null
virtualenvs.path = "{cache-dir}\\virtualenvs"  # C:\Users\usr\AppData\Local\pypoetry\Cache\virtualenvs
virtualenvs.use-poetry-python = true
```

### Python Sysconfig

<details>
  

  ```
Platform: "win-amd64"
Python version: "3.13"
Current installation scheme: "nt"

Paths:
        data = "C:\Program Files\WindowsApps\PythonSoftwareFoundation.Python.3.13_3.13.3824.0_x64__qbz5n2kfra8p0"
        include = "C:\Program Files\WindowsApps\PythonSoftwareFoundation.Python.3.13_3.13.3824.0_x64__qbz5n2kfra8p0\Include"
        platinclude = "C:\Program Files\WindowsApps\PythonSoftwareFoundation.Python.3.13_3.13.3824.0_x64__qbz5n2kfra8p0\Include"
        platlib = "C:\Program Files\WindowsApps\PythonSoftwareFoundation.Python.3.13_3.13.3824.0_x64__qbz5n2kfra8p0\Lib\site-packages"
        platstdlib = "C:\Program Files\WindowsApps\PythonSoftwareFoundation.Python.3.13_3.13.3824.0_x64__qbz5n2kfra8p0\Lib"
        purelib = "C:\Program Files\WindowsApps\PythonSoftwareFoundation.Python.3.13_3.13.3824.0_x64__qbz5n2kfra8p0\Lib\site-packages"
        scripts = "C:\Program Files\WindowsApps\PythonSoftwareFoundation.Python.3.13_3.13.3824.0_x64__qbz5n2kfra8p0\Scripts"
        stdlib = "C:\Program Files\WindowsApps\PythonSoftwareFoundation.Python.3.13_3.13.3824.0_x64__qbz5n2kfra8p0\Lib"

Variables:
        BINDIR = "C:\Users\usr\AppData\Local\Microsoft\WindowsApps\PythonSoftwareFoundation.Python.3.13_qbz5n2kfra8p0"
        BINLIBDEST = "C:\Program Files\WindowsApps\PythonSoftwareFoundation.Python.3.13_3.13.3824.0_x64__qbz5n2kfra8p0\Lib"
        EXE = ".exe"
        EXT_SUFFIX = ".cp313-win_amd64.pyd"
        INCLUDEPY = "C:\Program Files\WindowsApps\PythonSoftwareFoundation.Python.3.13_3.13.3824.0_x64__qbz5n2kfra8p0\Include"
        LDLIBRARY = "python313.dll"
        LIBDEST = "C:\Program Files\WindowsApps\PythonSoftwareFoundation.Python.3.13_3.13.3824.0_x64__qbz5n2kfra8p0\Lib"
        LIBDIR = "C:\Program Files\WindowsApps\PythonSoftwareFoundation.Python.3.13_3.13.3824.0_x64__qbz5n2kfra8p0\libs"
        LIBRARY = "python313.dll"
        Py_GIL_DISABLED = "0"
        SOABI = "cp313-win_amd64"
        TZPATH = ""
        VERSION = "313"
        VPATH = "..\.."
        abi_thread = ""
        abiflags = ""
        base = "C:\Program Files\WindowsApps\PythonSoftwareFoundation.Python.3.13_3.13.3824.0_x64__qbz5n2kfra8p0"
        exec_prefix = "C:\Program Files\WindowsApps\PythonSoftwareFoundation.Python.3.13_3.13.3824.0_x64__qbz5n2kfra8p0"
        implementation = "Python"
        implementation_lower = "python"
        installed_base = "C:\Program Files\WindowsApps\PythonSoftwareFoundation.Python.3.13_3.13.3824.0_x64__qbz5n2kfra8p0"
        installed_platbase = "C:\Program Files\WindowsApps\PythonSoftwareFoundation.Python.3.13_3.13.3824.0_x64__qbz5n2kfra8p0"
        platbase = "C:\Program Files\WindowsApps\PythonSoftwareFoundation.Python.3.13_3.13.3824.0_x64__qbz5n2kfra8p0"
        platlibdir = "DLLs"
        prefix = "C:\Program Files\WindowsApps\PythonSoftwareFoundation.Python.3.13_3.13.3824.0_x64__qbz5n2kfra8p0"
        projectbase = "C:\Users\usr\AppData\Local\Microsoft\WindowsApps\PythonSoftwareFoundation.Python.3.13_qbz5n2kfra8p0"
        py_version = "3.13.14"
        py_version_nodot = "313"
        py_version_nodot_plat = "313"
        py_version_short = "3.13"
        srcdir = "C:\Users\usr\AppData\Local\Microsoft\WindowsApps\PythonSoftwareFoundation.Python.3.13_qbz5n2kfra8p0"
        userbase = "C:\Users\usr\AppData\Local\Packages\PythonSoftwareFoundation.Python.3.13_qbz5n2kfra8p0\LocalCache\local-packages"
  ```
</details>

### Example pyproject.toml

```TOML

```

### Poetry Runtime Logs

<details>
  

  ```
c:\temp\my project with spaces>uv run poetry -vvv init -n
Loading configuration file C:\Users\usr\AppData\Roaming\pypoetry\config.toml
  ```
</details>

Socialpranker:
## Summary

Closes #10974.

`poetry init`/`poetry new` (which shares the same code path) used the current directory's name verbatim as the default package name. Per the [PyPA name normalization spec](https://packaging.python.org/en/latest/specifications/name-normalization/#name-format), a valid name may only contain ASCII letters, numbers, period, underscore and hyphen — so a directory like `my project with spaces` produced a `pyproject.toml` with an invalid `name = "my project with spaces"` instead of `name = "my-project-with-spaces"`.

This adds a small `InitCommand._sanitize_package_name` helper that replaces any disallowed character (including whitespace) with a hyphen, then reuses `packaging.utils.canonicalize_name` (already a dependency, already used elsewhere in this file for normalizing dependency names) to collapse/lowercase the result the same way it normalizes any other valid package name.

This only changes the *default* value derived from the directory name; an explicit `--name`/interactive value is left as-is, same as before.

## Test plan

- [x] Added `test_sanitize_package_name` — parametrized unit test covering the reported case plus a few edge cases (repeated separators, mixed `.`/`_`, leading/trailing separators).
- [x] Added `test_noninteractive_sanitizes_default_name_from_directory` — reproduces the exact scenario from the issue (`poetry init` run non-interactively in a directory named `my project with spaces`), asserts the generated `pyproject.toml` has a valid `name`.
- [x] `pytest tests/console/commands/test_init.py` - 73 passed
- [x] `pytest tests/console/commands/test_new.py` - 24 passed (unaffected, `NewCommand` shares this code path)
- [x] `ruff check` / `ruff format --check` - clean
- [x] `mypy src/poetry/console/commands/init.py` - no issues

## AI disclosure

This PR was written with the help of Claude Code (Claude Sonnet 5). All code was reviewed and tested by me before submission.

### DIFF
--- a/src/poetry/console/commands/init.py
+++ b/src/poetry/console/commands/init.py
@@ -475,7 +475,7 @@ The <c1>init</c1> command creates a basic <comment>pyproject.toml</> file in the
             cwd=cwd,
         )
         return [
-            parser.parse(re.sub(r"@\s*latest$", "", requirement, flags=re.I))
+            parser.parse(re.sub(r"@\s*latest$", "", requirement, flags=re.IGNORECASE))
             for requirement in requirements
         ]
 
@@ -501,9 +501,16 @@ The <c1>init</c1> command creates a basic <comment>pyproject.toml</> file in the
         conforms to the PyPA name format: only ASCII letters, numbers,
         period, underscore and hyphen.
         https://packaging.python.org/en/latest/specifications/name-normalization/#name-format
+
+        A directory name that holds no ASCII alphanumerics at all (say a
+        non-Latin script) sanitizes to an empty string, which is not a usable
+        default. Fall back to the lowercased directory name in that case, which
+        is what this used to do for every directory.
         """
         replaced = re.sub(r"[^A-Za-z0-9._-]+", "-", name)
-        return str(canonicalize_name(replaced)).strip("-")
+        sanitized = str(canonicalize_name(replaced)).strip("-")
+
+        return sanitized or name.lower()
 
     @staticmethod
     def _validate_author(author: str, default: str) -> str | None:
--- a/src/poetry/console/commands/init.py
+++ b/src/poetry/console/commands/init.py
@@ -128,7 +128,7 @@ The <c1>init</c1> command creates a basic <comment>pyproject.toml</> file in the
 
         name = self.option("name")
         if not name:
-            name = project_path.name.lower()
+            name = self._sanitize_package_name(project_path.name)
 
             if is_interactive:
                 question = self.create_question(
@@ -494,6 +494,16 @@ The <c1>init</c1> command creates a basic <comment>pyproject.toml</> file in the
             requires[name] = constraint
 
         return requires
+
+    @staticmethod
+    def _sanitize_package_name(name: str) -> str:
+        """Turn an arbitrary string (e.g. a directory name) into a name that
+        conforms to the PyPA name format: only ASCII letters, numbers,
+        period, underscore and hyphen.
+        https://packaging.python.org/en/latest/specifications/name-normalization/#name-format
+        """
+        replaced = re.sub(r"[^A-Za-z0-9._-]+", "-", name)
+        return str(canonicalize_name(replaced)).strip("-")
 
     @staticmethod
     def _validate_author(author: str, default: str) -> str | None:
--- a/src/poetry/console/commands/init.py
+++ b/src/poetry/console/commands/init.py
@@ -498,14 +498,22 @@ The <c1>init</c1> command creates a basic <comment>pyproject.toml</> file in the
     @staticmethod
     def _sanitize_package_name(name: str) -> str:
         """Turn an arbitrary string (e.g. a directory name) into a name that
-        conforms to the PyPA name format: only ASCII letters, numbers,
-        period, underscore and hyphen.
+        conforms to the PyPA name format.
         https://packaging.python.org/en/latest/specifications/name-normalization/#name-format
+
+        Characters outside the format are replaced with a hyphen, then
+        ``canonicalize_name`` normalizes the result, so the return value is
+        narrower than the format allows: lowercase ASCII letters, numbers and
+        hyphens only. Runs of period, underscore and hyphen all collapse to a
+        single hyphen ("My_Project" and "my.project" both give "my-project").
 
         A directory name that holds no ASCII alphanumerics at all (say a
         non-Latin script) sanitizes to an empty string, which is not a usable
         default. Fall back to the lowercased directory name in that case, which
-        is what this used to do for every directory.
+        is what this used to do for every directory. That fallback is the one
+        return value that need not conform to the format -- "日本語" stays
+        "日本語" -- matching the pre-existing behaviour rather than replacing a
+        meaningful directory name with a placeholder.
         """
         replaced = re.sub(r"[^A-Za-z0-9._-]+", "-", name)
         sanitized = str(canonicalize_name(replaced)).strip("-")
--- a/tests/console/commands/test_init.py
+++ b/tests/console/commands/test_init.py
@@ -137,6 +137,11 @@ def test_noninteractive_sanitizes_default_name_from_directory(
         ("My_Package.Name", "my-package-name"),
         ("MyProject", "myproject"),
         ("  --my..name__ ", "my-name"),
+        # nothing survives sanitization: keep the lowercased directory name
+        # rather than returning an empty default
+        ("日本語", "日本語"),
+        ("проект", "проект"),
+        ("---", "---"),
     ],
 )
 def test_sanitize_package_name(name: str, expected: str) -> None:
--- a/tests/console/commands/test_init.py
+++ b/tests/console/commands/test_init.py
@@ -105,6 +105,44 @@ def test_noninteractive(
     assert expected_build_system in toml_content
 
 
+def test_noninteractive_sanitizes_default_name_from_directory(
+    app: PoetryTestApplication,
+    mocker: MockerFixture,
+    poetry: Poetry,
+    repo: DummyRepository,
+    tmp_path: Path,
+) -> None:
+    command = app.find("init")
+    assert isinstance(command, InitCommand)
+    command._pool = poetry.pool
+
+    project_path = tmp_path / "my project with spaces"
+    project_path.mkdir()
+
+    p = mocker.patch("pathlib.Path.cwd")
+    p.return_value = project_path
+
+    tester = CommandTester(command)
+    tester.execute(interactive=False)
+
+    toml_content = (project_path / "pyproject.toml").read_text(encoding="utf-8")
+    assert 'name = "my-project-with-spaces"' in toml_content
+
+
+@pytest.mark.parametrize(
+    ("name", "expected"),
+    [
+        ("foo", "foo"),
+        ("my project with spaces", "my-project-with-spaces"),
+        ("My_Package.Name", "my-package-name"),
+        ("MyProject", "myproject"),
+        ("  --my..name__ ", "my-name"),
+    ],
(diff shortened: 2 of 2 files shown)


Socialpranker:
The `macOS aarch64 (Python 3.11) / mypy` check failed, but it looks like a CI infrastructure issue rather than a real failure — the job was cancelled after 17m26s with "The job was not acquired by Runner of type hosted even after multiple attempts", and the same mypy check passes on every other Python version/platform combination. Could someone with the right access re-run that job?

src/poetry/console/commands/init.py
```diff
@@ -128,7 +128,7 @@ def _init_pyproject(
 
         name = self.option("name")
         if not name:
-            name = project_path.name.lower()
+            name = self._sanitize_package_name(project_path.name)
```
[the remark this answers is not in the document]
Socialpranker: Good catch, and it turned out to be worse than an edge case — thanks.

`_sanitize_package_name` returned an empty string for any directory whose
name holds no ASCII alphanumerics, which includes ordinary non-Latin
directory names, not just punctuation-only ones:

```
directory   before this branch   on 625a9e3
'日本語'      '日本語'               ''
'проект'     'проект'              ''
'ααα'        'ααα'                 ''
'---'        '---'                 ''
```

The first column is what `init` did before this branch (`project_path.name.lower()`),
so this branch was a regression for those users: `poetry init` offered an
empty package name where it used to offer something usable.

Fixed in 5818d36 by falling back to the lowercased directory name when
sanitization leaves nothing:

```python
replaced = re.sub(r"[^A-Za-z0-9._-]+", "-", name)
sanitized = str(canonicalize_name(replaced)).strip("-")

return sanitized or name.lower()
```

I went with the previous behaviour rather than a fixed default like
`my-package`: it keeps the regression at zero, and a hardcoded name would
silently replace a meaningful directory name with a meaningless one. The
fallback value can still be non-conforming, exactly as it was before this
branch — narrowing that is a separate question from the spaces-in-names
problem this PR set out to fix, and in interactive mode the prompt is
pre-filled so it can be edited.

Added the three cases above to `test_sanitize_package_name`. Verified they
fail without the fallback (`assert '' == '日本語'`) and pass with it;
`tests/console/commands/test_init.py` is 76 passed, and `mypy` on
`init.py` is clean.

Socialpranker:
On the second point from the review — the docstring — you were right, and it was wrong rather than merely unclear. Fixed in a93ff3e.

It claimed the return value conforms to the PyPA name format, "only ASCII letters, numbers, period, underscore and hyphen". `canonicalize_name` collapses runs of `.`, `_` and `-` into a single hyphen and lowercases, so periods and underscores never survive:

| directory | returned |
|---|---|
| `My_Project` | `my-project` |
| `my.project` | `my-project` |
| `foo_bar.baz` | `foo-bar-baz` |
| `日本語` | `日本語` |

The last row is the other half of it: the fallback added in 5818d36 deliberately returns a value that does *not* conform to the format, which the old docstring flatly contradicted. The rewrite states both — that the conforming path is narrower than the format allows, and that the fallback is the one exception, by design.

One thing worth flagging for a human reviewer, since it looks like an oversight and isn't: a directory named `---` returns `---`, because sanitizing strips it to empty and the fallback hands back the lowercased directory name. That is exactly what `poetry init` did before this branch, so it is not a regression, and in interactive mode the prompt is pre-filled and editable. Turning non-conforming directory names into a valid default is a broader change than this PR's scope.

`tests/console/commands/test_init.py` is 76 passed, `ruff check` and `ruff format --check` clean on the changed file. Docstring-only change; no behaviour touched.

Socialpranker:
Checking back — did the docstring fix address what you were looking for, or is there another round needed?

src/poetry/console/commands/init.py
```diff
@@ -495,6 +495,31 @@ def _format_requirements(self, requirements: list[dict[str, str]]) -> Requiremen
...
+        A directory name that holds no ASCII alphanumerics at all (say a
+        non-Latin script) sanitizes to an empty string, which is not a usable
+        default. Fall back to the lowercased directory name in that case, which
+        is what this used to do for every directory. That fallback is the one
+        return value that need not conform to the format -- "日本語" stays
+        "日本語" -- matching the pre-existing behaviour rather than replacing a
+        meaningful directory name with a placeholder.
+        """
+        replaced = re.sub(r"[^A-Za-z0-9._-]+", "-", name)
+        sanitized = str(canonicalize_name(replaced)).strip("-")
```
radoering, line 519: Do we really want to canonicalize the name here?

It does not seem to be necessary (see e.g. https://packaging.python.org/en/latest/guides/writing-pyproject-toml/#name). We could still do it because `init` and `new` can be a bit opinionated but I would only tend to do it if there are references that this is considered best practice.

Socialpranker: fair point, dropped it. init just lowercased the dir name before this, so the canonicalization was something i added without much reason. My_Package gives my_package now. shout if you'd rather it kept the dots collapsed too

radoering: Ah, I missed that we had already lowercased it. I think we should choose one of the two extremes: only ensure that the name is valid and even keep uppercase or canonicalize the name completely.

Rethinking about it, canonicalizing the name might be the better choice for the default user because using a canonicalized name is probably less error-prone. Power users who know what they are doing and really do not want a canonicalized name can still edit the pyproject.toml manually.

Sorry, for the back and forth.

Socialpranker:
no worries, went with full canonicalization

radoering (approved):

PR #10976, opened 2026-07-13
fix(solver): preserve extra marker scope across overrides

### LINKED ISSUE #10971: Ambiguous lock file when two simultaneously-installable extras require different versions of a shared transitive dependency
## Description

When two optional-dependency groups (extras) that can legitimately be
installed together each pull in a different version constraint for the
same transitive package, `poetry lock` writes **two separate lock entries**
for that package, distinguished only by `extra == "..."` markers whose
conditions overlap (both are true when the two extras are installed
together). Since only one physical version of a package can ever be
installed, this makes the lock file's outcome for that package
non-deterministic: repeated `poetry install` runs with the *same*
`pyproject.toml`, the *same* `poetry.lock`, and the *same* `--extras` flag
install different versions of the package across runs, purely depending on
the order in which the installer processes the (ambiguous) list of planned
operations.

My case: 

Two extras (`networks`, `sbml`) can be installed together. `networks` pins
`igraph==0.11.9` directly; `sbml` transitively depends on `py4cytoscape`,
which declares an unconstrained `igraph` dependency. `poetry lock` resolves
this into two lock entries for the same package, with overlapping markers:

```toml
[[package]]
name = "igraph"
version = "0.11.9"
markers = "extra == \"networks\" or extra == \"all\" or extra == \"sbml\""

[[package]]
name = "igraph"
version = "1.0.0"
markers = "extra == \"sbml\" or extra == \"all\""
```

Both conditions are true when `sbml` is requested, so the lock file doesn't
actually specify a single install outcome. Which version gets installed
depends on the order the installer processes its plan. Running
`poetry install --extras "networks sbml"` repeatedly, from a fresh
virtualenv each time, with the exact same `pyproject.toml`/`poetry.lock`,
installs different `igraph` versions across runs:

```bash
poetry lock
for i in 1 2 3; do
  rm -rf "$(poetry env info --path 2>/dev/null)"
  poetry install --extras "networks sbml" >/dev/null 2>&1
  poetry run python3 -c "import igraph; print(igraph.__version__)"
done
```

```
1.0.0
0.11.9
1.0.0
```

Confirmed on both Poetry 2.3.3 and the latest release, 2.4.1.

For comparison, `pip install .[networks,sbml]` against the same
`pyproject.toml` resolves deterministically to `0.11.9` every time, since it
resolves the actually-requested extras fresh at install time rather than
precomputing a lock file for every possible extras combination.

## Workarounds

Adding the exact same explicit pin (`igraph==0.11.9`) directly to the
`sbml` extra as well (even though nothing in `sbml` imports `igraph`
directly) makes the pin visible in every marker branch that could otherwise
pull in the unconstrained transitive version, and resolves cleanly.

## Poetry Installation Method

install.python-poetry.org

## Operating System

Ubuntu 22.04.5 LTS

## Poetry Version

Poetry (version 2.3.3), confirmed still present on Poetry (version 2.4.1),
the latest release at time of writing.

## Poetry Configuration

```bash session
cache-dir = "~/.cache/pypoetry"
data-dir = "~/.local/share/pypoetry"
installer.max-workers = null
installer.no-binary = null
installer.only-binary = null
installer.parallel = true
installer.re-resolve = false
keyring.enabled = true
python.installation-dir = "{data-dir}/python"
requests.max-retries = 0
solver.lazy-wheel = true
system-git-client = false
virtualenvs.create = true
virtualenvs.in-project = null
virtualenvs.path = "{cache-dir}/virtualenvs"
```

## Example pyproject.toml

```toml
[build-system]
requires = ["setuptools"]
build-backend = "setuptools.build_meta"

[project]
name = "poetry-extras-igraph-conflict-mwe"
version = "0.0.0"
requires-python = ">=3.12"
dependencies = []

networks = ["networkx", "igraph==0.11.9"]
sbml = ["python-libsbml>=5.19", "tabularqual"]
all = ["python-libsbml>=5.19", "tabularqual", "networkx", "igraph==0.11.9"]

[tool.poetry]
package-mode = false
```

## Poetry Runtime Logs

<details>

  Excerpt from `poetry install --extras "networks sbml" -vvv` (fresh
  virtualenv), showing the resolved lock file's two conflicting `igraph`
  entries producing two operations for the same package within a single
  install plan:

  ```
  - Installing igraph (0.11.9): Skipped for the following reason: Already installed
  ...
  - Updating igraph (0.11.9 -> 1.0.0)
  ```

  A separate run (same lock file, same command, fresh virtualenv) instead
  produced:

  ```
  - Downgrading igraph (1.0.0 -> 0.11.9)
  - Installing igraph (1.0.0): Skipped for the following reason: Already installed
  ```

</details>

snoopuppy582:
## Summary

- preserve root-extra marker worlds while dependencies are completed
- keep extras on transitive dependency edges local to their requiring package
- reuse the existing marker-split conflict recovery instead of reconstructing dependency causes after a failure

## Problem

When incompatible versions are selected through different root extras, Poetry can merge solver results with overlapping lock markers. The same root marker context must survive when those versions are reached through extras of a transitive package.

The distinction is marker scope: a root extra is a global install condition, while an extra belonging to a dependency package is only meaningful while resolving that package.

## Change

Root dependencies initialize their transitive marker from the effective root marker, including implicit membership in a root extra. For every non-root edge, the parent's transitive marker is propagated while extras are removed from the edge's own marker. This preserves the root marker world without treating package-local extras as global conditions.

The transitive marker is assigned explicitly even when it is unconstrained. Leaving it unset would make the accessor fall back to the dependency's own marker and could reintroduce a package-local extra.

Root duplicate requirements containing extra markers use the existing synthetic marker-split override. Compatibility-mode merging preserves extras only for that synthetic override; ordinary package override extras are still removed.

## Tests

- exact #10971 lock regeneration and install dry run
- root-extra conflict regression
- transitive-extra conflict regression
- negative regression where both transitive-extra paths are simultaneously active
- `tests/puzzle`: 361 passed
- `tests/mixology`: 46 passed
- lock/install/update command tests: 113 passed
- mypy and pre-commit passed
- full serial suite: 3185 passed, with 8 local Windows/environment failures unrelated to solver markers

Resolves: #10971

- [x] Added tests for changed code.

## AI assistance

I used OpenAI Codex to help reproduce the issue, analyze marker propagation, draft the implementation and regression tests, and run local validation. I reviewed the simplified marker-scope design and validation results before approving this update and remain responsible for the submitted change.

### DIFF
--- a/src/poetry/mixology/version_solver.py
+++ b/src/poetry/mixology/version_solver.py
@@ -40,10 +40,11 @@ def _effective_marker(dependency: Dependency) -> BaseMarker:
     The marker describing when a dependency actually applies: its own marker
     intersected with the marker of the package that required it (a dependency is
     only reached when its requiring package applies).
+
+    Extras in the transitive marker belong to the root marker world. Extras in
+    the dependency's own marker are local to its requiring package.
     """
-    return dependency.transitive_marker.without_extras().intersect(
-        dependency.marker.without_extras()
-    )
+    return dependency.transitive_marker.intersect(dependency.marker.without_extras())
 
 
 class Preference(IntEnum):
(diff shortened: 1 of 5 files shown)


dimbleby:
My instinct is that this is not correct.  The specific issue reported may happen to involve root extras, but it should generalise to extras produced via transitive dependencies.   

Therefore I doubt that a fix which special-cases root extras can be the right fix.

snoopuppy582:
Thanks, the root-package special case was too narrow. I added a regression where mutually exclusive root-extra paths activate different extras of a transitive package, and a control where both paths are active and must still fail.

The revised patch carries global root markers with the existing synthetic marker-split override. During conflict recovery it propagates those marker worlds through dependency causes, while evaluating non-root extra markers against the depending package's active features. Real package override extras are still removed at the merge boundary.

The puzzle and mixology suites, lock/install/update command tests, mypy, pre-commit, and the original #10971 project pass locally. I have kept the PR in draft for review of this broader approach.

dimbleby:
Well now it looks _too_ complicated!

I had my Claude take a look at this a week or two ago but did not yet take the time to properly think through its proposal.  That is on a computer from which I am currently far away, I can take a look in a few days to see if I am misremembering,  but I think it was both plausible and simpler than this.

dimbleby:
Side note: if all that you, the human, are doing here is acting as an intermediary for the AI - this is low value.  Maintainers can have those conversations directly, no need to spend everyone's time on doing it in this extremely slow way.

PR #10982, opened 2026-07-16
Fix explicit-priority sources not checked by poetry show --outdated

### LINKED ISSUE #10425: poetry show --outdated does not show outdated packages from "explicit" repositories
### Description

Maybe I'm missing something, but try as I might, I can't seem to get `poetry` (version 2.1.3) to show output for outdated packages that come from a private repository that is set as `priority = "explicit"`.

Example line in `[tool.poetry.dependencies]`:
```
my-private-package = {version = "^0.14.0", source = "MyRepo"}
```

in a `pyproject.toml` that also includes:
```
[[tool.poetry.source]]
name = "MyRepo"
url = "https://repo.shared.mycompany.com/repository/pypi-mycompany/simple"
priority = "explicit"
```

If `MyRepo` has another version of the package (say, `0.15.0`) then I would expect `poetry show -o` to list the package as outdated.  But... nope; it's simply not included in the output of `poetry show -o`, even though PyPi packages are correctly shown if they are outdated.

### Workarounds

`poetry show --outdated` works correctly if I specify the private repo as `supplemental` - but not if I keep it as `explicit`.  The latter is preferable since it should avoid poetry checking my private repo for packages that I don't want to source from there.

pipx

### Operating System

MacOS Sequoia 15.5

### Poetry Version

Poetry (version 2.1.3)

### Poetry Configuration

```bash session
cache-dir = "/Users/jclerman/Library/Caches/pypoetry"
data-dir = "/Users/jclerman/Library/Application Support/pypoetry"
installer.max-workers = null
installer.no-binary = null
installer.only-binary = null
installer.parallel = true
installer.re-resolve = true
keyring.enabled = true
python.installation-dir = "{data-dir}/python"  # /Users/jclerman/Library/Application Support/pypoetry/python
repositories.sirona.url = "https://repo.shared.sirona.com/repository/pypi-sirona/simple"
requests.max-retries = 0
solver.lazy-wheel = true
system-git-client = false
virtualenvs.create = true
virtualenvs.in-project = null
virtualenvs.path = "{cache-dir}/virtualenvs"  # /Users/jclerman/Library/Caches/pypoetry/virtualenvs
```

### Python Sysconfig

<details>
  

  ```
  Paste the output of 'python -m sysconfig', over this line.
  ```
</details>

### Example pyproject.toml

```TOML
key lines are included in the bug description - let me know if that isn't enough to reproduce the issue.
```

### Poetry Runtime Logs

[poetry.log](https://github.com/user-attachments/files/20698168/poetry.log)

pctablet505:
## Problem

When a dependency is pinned to a repository with `priority = "explicit"`, `poetry show --outdated` never reports it as outdated, even when a newer version exists in that repository.

## Root cause

`ShowCommand.find_latest_package` looks up each locked package's matching `Dependency` to read `allow_prereleases`, but discards the dependency's `source_name` before calling `VersionSelector.find_best_candidate`. Without a `source`, `RepositoryPool.find_packages` falls back to searching `self.repositories`, which intentionally excludes repositories with `Priority.EXPLICIT` (see the `repositories` property docstring in `repository_pool.py`). So a package that only exists in an explicit repository is silently skipped.

## Fix

Pass the dependency's `source_name` through to `find_best_candidate`, the same way `poetry init`/`poetry add` already do (`ConsoleCommand._find_best_version_for_package`). This lets `RepositoryPool.find_packages` resolve the specific named repository directly, bypassing the explicit-repo exclusion just for that lookup, exactly as intended for a dependency pinned to a source.

## Testing

Added `test_show_outdated_explicit_source` covering a dependency sourced from an explicit-priority repository. Confirmed it fails with the old code (`PackageNotFoundError`, since the explicit repo is never consulted) and passes with the fix.

Fixes #10425

### DIFF
--- a/src/poetry/console/commands/show.py
+++ b/src/poetry/console/commands/show.py
@@ -656,12 +656,18 @@ lists all packages available."""
                     return provider.search_for_direct_origin_dependency(dep)
 
         allow_prereleases: bool | None = None
-        source: str | None = None
         for dep in requires:
             if dep.name == package.name:
                 allow_prereleases = dep.allows_prereleases()
-                source = dep.source_name
                 break
+
+        # Use the locked package's own source rather than guessing from the first
+        # matching dependency. A package may be declared with several constraints
+        # (different markers and/or sources); the one active in the current
+        # environment is the one that was locked, and its source is recorded
+        # unambiguously on the locked package as ``source_reference`` (empty for
+        # the default repository). Direct-origin packages are handled above.
+        source = None if package.is_direct_origin() else package.source_reference
 
         name = package.name
         selector = VersionSelector(self.poetry.pool)
--- a/src/poetry/console/commands/show.py
+++ b/src/poetry/console/commands/show.py
@@ -656,16 +656,18 @@ lists all packages available."""
                     return provider.search_for_direct_origin_dependency(dep)
 
         allow_prereleases: bool | None = None
+        source: str | None = None
         for dep in requires:
             if dep.name == package.name:
                 allow_prereleases = dep.allows_prereleases()
+                source = dep.source_name
                 break
 
         name = package.name
         selector = VersionSelector(self.poetry.pool)
 
         return selector.find_best_candidate(
-            name, f">={package.pretty_version}", allow_prereleases
+            name, f">={package.pretty_version}", allow_prereleases, source=source
         )
 
     def get_update_status(self, latest: Package, package: Package) -> str:
--- a/tests/console/commands/test_show.py
+++ b/tests/console/commands/test_show.py
@@ -2651,6 +2651,111 @@ cachy 0.1.0 0.2.0 Cachy package
 
 
 @output_format_parametrize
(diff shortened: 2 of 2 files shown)


src/poetry/console/commands/show.py
```diff
@@ -656,16 +656,18 @@ def find_latest_package(
                     return provider.search_for_direct_origin_dependency(dep)
 
         allow_prereleases: bool | None = None
+        source: str | None = None
         for dep in requires:
             if dep.name == package.name:
                 allow_prereleases = dep.allows_prereleases()
+                source = dep.source_name
```
Sanjays2402: This takes the first dependency with a matching name, but Poetry supports multiple constraints for one package with different platform markers and sources. On an environment matching the second constraint, `show --outdated` will query the first source instead; can this select the active dependency or use the locked package source?

radoering: fixed in the second commit

radoering (approved):

PR #10983, opened 2026-07-17
tests: deduplicate environment check output helper

loulanyue:
Relates-to: #3155

## Summary

- move the duplicated environment `check_output` mock into `tests.helpers`
- preserve the console tests' `/usr` base prefix and the environment manager tests' runtime `sys.base_prefix`
- import the shared helper directly from each affected test module

This removes 34 net lines while keeping the existing test behavior unchanged.

## Verification

- `pytest -n 0 tests/console/commands/env/test_use.py tests/console/commands/env/test_remove.py tests/utils/env/test_env_manager.py` (`68 passed`)
- `mypy tests/helpers.py tests/console/commands/env/helpers.py tests/console/commands/env/test_remove.py tests/console/commands/env/test_use.py tests/utils/env/test_env_manager.py`
- `pre-commit run --files tests/helpers.py tests/console/commands/env/helpers.py tests/console/commands/env/test_remove.py tests/console/commands/env/test_use.py tests/utils/env/test_env_manager.py`

The local Homebrew environment provides `python3` but no bare `python`, so the pytest run used a temporary PATH entry mapping `python` to `python3`. No repository files were changed for that local environment difference.

## Checklist

- [x] Updated and ran the affected tests.
- [ ] Updated documentation (not applicable; test-only cleanup).

### DIFF
--- a/tests/helpers.py
+++ b/tests/helpers.py
@@ -11,6 +11,7 @@ from typing import TYPE_CHECKING
 
 import keyring
 
+from poetry.core.constraints.version import Version
 from poetry.core.packages.package import Package
 from poetry.core.packages.utils.link import Link
 from poetry.core.vcs.git import ParsedUrl
@@ -26,6 +27,7 @@ from poetry.utils.password_manager import PoetryKeyring
 
 
 if TYPE_CHECKING:
+    from collections.abc import Callable
     from collections.abc import Iterator
     from collections.abc import Mapping
     from typing import Any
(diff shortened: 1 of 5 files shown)


tests/helpers.py
```diff
@@ -52,6 +53,36 @@
 
 # Used as a mock for latest git revision.
 MOCK_DEFAULT_GIT_REVISION = "9cf87a285a2d3fbb0b9fa621997b3acc3631ed24"
+VERSION_3_7_1 = Version.parse("3.7.1")
+
+
+def check_output_wrapper(
+    version: Version = VERSION_3_7_1, base_prefix: str = "/usr"
```
radoering, line 60: Can we get rid of the `base_prefix` parameter and always return `sys.base_prefix` instead of `"/usr"`. Do the tests that do not set this parameter expect the prefix to be `"/usr"`?

PR #10994, opened 2026-07-26
chore: update dependencies

radoering:


src/poetry/utils/env/mock_env.py
```diff
@@ -16,7 +16,7 @@
 class MockEnv(NullEnv):
     def __init__(
         self,
-        version_info: tuple[int, int, int] | PythonVersion = (3, 7, 0),
+        version_info: tuple[int, int, int] | PythonVersion = (3, 9, 0),
```
radoering, line 19: Without this change some tests in `test_executor.py` fail because `virtualenv` refuses to create older venvs since version 2.5.1, see https://virtualenv.pypa.io/en/latest/changelog.html#v21-5-1-2026-06-16

I decided to change the default and adapt tests that depend on the exact python version of the `MockEnv`.

PR #10997, opened 2026-07-29
fix(env): fallback to host Python version when target Python embedded wheel is missing

hatch01:
### Motivation

When running Poetry's test suite with modern versions of `virtualenv` (specifically `v21.5.0` or newer), the test suite fails on tests executing package operations within mock environments (issue faced here: https://github.com/NixOS/nixpkgs/issues/544083). 
This happens because:
1. `MockEnv` defaults to simulating a Python `3.7.0` environment (unless overridden).
2. Modern versions of `virtualenv` have dropped support for older Python versions (Python 3.7 support dropped in `v20.27.1`, and Python 3.8 support dropped in `v21.5.0`), removing their respective embedded seed wheels (`pip`, `setuptools`).
3. When Poetry attempts to retrieve the embedded pip wheel using `get_embed_wheel("pip", "3.7")` inside `get_embedded_wheel`, it returns `None`, raising a fatal `RuntimeError: embedded pip wheel not found`.
Since modern Python packaging wheels are universal (`py3-none-any.whl`), a wheel bundled for a newer version of Python is fully compatible and functional on older/mocked versions.

### Solution

This PR implements a fallback inside `get_embedded_wheel()` to query the host running Python's version (e.g. `sys.version_info`) if the mock/target environment's Python version wheel is not found in the `virtualenv` database.

This keeps Poetry's test suite highly compatible and prevents test regressions when building or testing Poetry on modern linux distributions and package repositories (such as NixOS/Nixpkgs) where Python 3.14 and virtualenv 21.6+ are the defaults.

### DIFF
--- a/src/poetry/utils/env/base_env.py
+++ b/src/poetry/utils/env/base_env.py
@@ -1,6 +1,7 @@
 from __future__ import annotations
 
 import contextlib
+import logging
 import os
 import re
 import subprocess
@@ -23,6 +24,9 @@ from poetry.utils.env.exceptions import EnvCommandError
 from poetry.utils.env.site_packages import SitePackages
 from poetry.utils.helpers import get_real_windows_path
 from poetry.utils.helpers import is_dir_writable
+
+
+logger = logging.getLogger(__name__)
 
 
 if TYPE_CHECKING:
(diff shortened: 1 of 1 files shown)


thunze:
> 1. `MockEnv` defaults to simulating a Python `3.7.0` environment (unless overridden).

https://github.com/python-poetry/poetry/blob/c459f14f1d5012bd0e3b6558c77b0e42c9d452df/src/poetry/utils/env/mock_env.py#L19

It defaults to 3.9 since https://github.com/python-poetry/poetry/pull/10994 from what I can tell.

hatch01:
So probably this PR is not useful if we keep the deps up to date.

PR #10992, opened 2026-07-24
Skip plugin loading for version output

pratyushsinghal7:
# Pull Request Check List

Resolves: #10625

- [x] Added **tests** for changed code.
- [x] Documentation is not required for this internal startup optimization.

## Summary

Skip application plugin discovery when Poetry is invoked with the global `--version` or `-V` flag. Plugins cannot change Poetry's version output, and loading their entry points adds avoidable startup latency.

The check lives in `_load_plugins()` alongside the existing `--no-plugins` fast path. It uses Cleo's parameter boundary handling so version arguments forwarded to subprocesses, such as `poetry run -- python -V`, continue to load Poetry plugins normally.

## Testing

- `.venv/bin/pytest tests/console/test_application.py -q` — 22 passed
- `PATH="/tmp/poetry-test-bin:$PATH" .venv/bin/pytest tests/console -q` — 958 passed, 6 skipped
- `PATH="/tmp/poetry-test-bin:$PATH" .venv/bin/pytest -q` — 3176 passed, 27 skipped
- `.venv/bin/mypy` — success across 374 source files
- `.venv/bin/pre-commit run --files src/poetry/console/application.py tests/console/test_application.py` — passed

PR #11004, opened 2026-08-01
perf: defer the requests imports in poetry.utils.helpers

dexhunter:
## Summary

`src/poetry/console/application.py` imports two filesystem helpers from
`poetry.utils.helpers`:

```python
from poetry.utils.helpers import directory
from poetry.utils.helpers import ensure_path
```

`poetry/utils/helpers.py` imports the HTTP stack at module level:

```python
from requests.exceptions import ChunkedEncodingError
from requests.exceptions import ConnectionError
from requests.utils import atomic_open

from poetry.utils.authenticator import get_default_authenticator
```

All four are used only inside `Downloader`, but every poetry command pays for them at
startup, including ones that never open a socket. This moves them to their point of use
and keeps them importable from the module through a PEP 562 `__getattr__`.

## Numbers

Cost of `import poetry.console.application` in a fresh interpreter, measured as the wall
time of `python -B -c "import poetry.console.application"` minus the wall time of
`python -B -c "pass"` in the same window, median of 9, on an otherwise idle core,
CPython 3.12.13:

| | before | after |
|---|---:|---:|
| import cost | 156.8 ms | 82.3 ms |

`requests` and `urllib3` drop out of `sys.modules` entirely for commands that do not use
them, which is where the time goes.

Two caveats worth stating. The absolute number depends on what else is installed in the
environment: in a second virtualenv on the same machine `requests` alone costs 104 ms
rather than 36 ms, and the same change measured 317 ms to 117 ms there. The ratio is the
stable part, not the milliseconds. And this is process startup only; it does not make
any command's actual work faster.

For context on why startup is worth anything here, #10625 measured `poetry --version` at
about 217 ms and `poetry --version --no-plugins` at about 116 ms. That issue and #10992
are about the plugin-discovery half of that gap. This is the other half, the part that
remains with `--no-plugins`, and the two are independent.

## Behaviour is unchanged

`ChunkedEncodingError`, `ConnectionError`, `atomic_open` and `get_default_authenticator`
remain importable from `poetry.utils.helpers` and resolve to the same objects. `dir()` on
the module is unchanged.

The part that needed care: a module-level `__getattr__` serves attribute access on the
*module object*, but never a bare global name lookup inside the module's own functions.
So writing

```python
except (ChunkedEncodingError, ConnectionError):
```

with `ConnectionError` supplied lazily would silently bind the **builtin**
`ConnectionError`. `requests.exceptions.ConnectionError` is not a subclass of it, so
resumable downloads would stop retrying and turn into hard failures, and nothing in the
existing test suite would notice. The retry site imports both names into function scope
instead, where the local binding shadows the builtin correctly.

## Tests

`tests/utils/test_helpers_lazy_imports.py`, 11 tests:

* each of the four names is still importable, is in `dir()`, and is the same object
* an unknown attribute still raises `AttributeError`
* importing `poetry.console.application` leaves `requests` and `urllib3` out of
  `sys.modules` - this is the regression the change exists to prevent, and it fails on
  current `main`
* `Downloader` still retries `ConnectionError`, `ChunkedEncodingError` and `SSLError`
  (the last subclasses requests' `ConnectionError`, so it is caught by the same clause,
  and only if that name is the requests class), and still propagates `ValueError` and
  `AttributeError` rather than swallowing them in a broad `except`

I checked these fail when they should. On unmodified `main` the import test fails. On a
variant that lets `ConnectionError` fall through to the builtin, and on one that widens
the clause to `except Exception`, the retry test fails.

`tests/utils` gives 358 passed / 23 failed / 21 skipped; the same 23 fail by name on an
unmodified checkout of this base and need Python interpreter managers this environment
does not have. `mypy` reports no issues on the changed file. `ruff check` is clean on
both files, apart from one pre-existing `PLC0206` in `merge_dicts` that is present on
`main` too and is untouched here.

## Notes

I tried a number of shapes for this, each scored against the unmodified module for
identical public surface and identical retry behaviour. The record is at
https://dashboard.weco.ai/share/kStcipB3sa9sf-YTFM_fv0G6_ZyLa61H

The larger variants also defer `logging`, `hashlib`, `tarfile`, `zipfile` and the
module-level `prioritised_hash_types` tables. They measured between 78.8 ms and 81.7 ms,
a 2.9 ms span against a 3.8 ms run-to-run noise band, so they are not distinguishable
from this one. They are three times the diff for no measurable gain, and they move
module-level public attributes behind a hook, so I kept the smaller change. Happy to
push the broader version if you would prefer it.

## Pull Request Check List

### DIFF
--- a/src/poetry/installation/executor.py
+++ b/src/poetry/installation/executor.py
@@ -25,8 +25,8 @@ from poetry.installation.wheel_installer import WheelInstaller
 from poetry.puzzle.exceptions import SolverProblemError
 from poetry.utils._compat import decode
 from poetry.utils.authenticator import Authenticator
+from poetry.utils.download import Downloader
 from poetry.utils.env import EnvCommandError
-from poetry.utils.helpers import Downloader
 from poetry.utils.helpers import get_file_hash
 from poetry.utils.helpers import get_highest_priority_hash_type
 from poetry.utils.helpers import pluralize
--- a/src/poetry/packages/direct_origin.py
+++ b/src/poetry/packages/direct_origin.py
@@ -11,7 +11,7 @@ from poetry.config.config import Config
 from poetry.inspection.info import PackageInfo
 from poetry.inspection.info import PackageInfoError
 from poetry.utils.authenticator import get_default_authenticator
-from poetry.utils.helpers import download_file
+from poetry.utils.download import download_file
 from poetry.utils.helpers import get_file_hash
 from poetry.vcs.git import Git
 
--- a/src/poetry/repositories/http_repository.py
+++ b/src/poetry/repositories/http_repository.py
@@ -36,8 +36,8 @@ from poetry.repositories.link_sources.html import HTMLPage
 from poetry.repositories.link_sources.json import SimpleJsonPage
 from poetry.utils.authenticator import Authenticator
 from poetry.utils.constants import REQUESTS_TIMEOUT
+from poetry.utils.download import download_file
 from poetry.utils.helpers import HTTPRangeRequestSupportedError
-from poetry.utils.helpers import download_file
 from poetry.utils.helpers import get_highest_priority_hash_type
 from poetry.utils.patterns import wheel_file_re
 
--- a/src/poetry/repositories/http_repository.py
+++ b/src/poetry/repositories/http_repository.py
@@ -36,8 +36,8 @@ from poetry.repositories.link_sources.html import HTMLPage
 from poetry.repositories.link_sources.json import SimpleJsonPage
 from poetry.utils.authenticator import Authenticator
 from poetry.utils.constants import REQUESTS_TIMEOUT
+from poetry.utils.download import HTTPRangeRequestSupportedError
 from poetry.utils.download import download_file
-from poetry.utils.helpers import HTTPRangeRequestSupportedError
 from poetry.utils.helpers import get_highest_priority_hash_type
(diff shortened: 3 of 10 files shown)


radoering:
I gave it a quick try and it seems to be worth it. However, I wonder whether we could extract `download_file` and `Downloader` into a separate `download` module and keep the imports in this module global.

We can omit the reexports of requests and authenticator in helpers since this should not be used by anybody. However, when moving `download_file` and `Downloader` into a separate module, we may want to apply this pattern in combination with a `DeprecationWarning` for `download_file` and `Downloader` because they might be used in plugins.

Please let me know if you want to give this proposal a try. Otherwise, I may have a look for myself.

dexhunter:
Thanks, I gave this a try.

I moved `Downloader` and `download_file` into `poetry.utils.download`, kept the requests and authenticator imports global in that module, and changed Poetry's internal callers to import from it directly. `poetry.utils.helpers` now keeps deprecated compatibility reexports for the two public helpers.

I also added coverage for retry exhaustion: with `max_retries=2`, the downloader consumes the initial response plus two resumed responses, then propagates the final connection error.

Validation on this exact commit: 60 focused tests passed; the broader affected set passed 182 tests with 1 skip; mypy and targeted pre-commit checks are clean. The paired import measurement moved from 149.85 ms to 78.08 ms.

radoering (approved):

PR #11011, opened 2026-08-12
Fix: GenericEnv.is_venv() in src/poetry/utils/env/generic_env.py compared...

M001N:
## Summary
Changed is_venv() to resolve both paths before comparing: `return self._path.resolve() != self._base.resolve()`.

## Problem
python-poetry/poetry issue reference: python-poetry/poetry#10991

## Root Cause
GenericEnv.is_venv() in src/poetry/utils/env/generic_env.py compared self._path and self._base with plain inequality (`self._path != self._base`) without resolving symlinks/aliases. When a Python version manager exposes the same interpreter via two different filesystem paths (e.g. an unversioned symlink and the concrete versioned directory it points to), the unresolved paths differ even though they refer to the same interpreter, so is_venv() incorrectly returns True. This causes EnvManager.create_venv() to believe Poetry is already running inside an isolated virtualenv and skip creating a project-specific venv.

## Testing
PASS - 35 passed, 1 skipped (pre-existing nix-only symlink test skipped on Windows); new test also independently verified to fail on unmodified code and pass with the fix.

## Related Issue
python-poetry/poetry#10991

### DIFF
--- a/src/poetry/utils/env/generic_env.py
+++ b/src/poetry/utils/env/generic_env.py
@@ -103,4 +103,4 @@ class GenericEnv(VirtualEnv):
         return super(VirtualEnv, self)._run(cmd, **kwargs)
 
     def is_venv(self) -> bool:
-        return self._path != self._base
+        return self._path.resolve() != self._base.resolve()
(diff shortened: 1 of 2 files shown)


edmorley:
@M001N Btw this PR now has conflicts (I presume due to #11042)

PR #11033, opened 2026-08-29
Cache Poetry JSON schema validation

dimbleby:
Companion to https://github.com/python-poetry/poetry-core/pull/965

As in the other - unlikely to make any difference to regular poetry users, but makes a dent in the unit test execution.  Locally I see it being less impactful than the poetry-core version, but still worth having.

### DIFF
--- a/src/poetry/json/__init__.py
+++ b/src/poetry/json/__init__.py
@@ -2,7 +2,9 @@ from __future__ import annotations
 
 import json
 
+from functools import cache
 from importlib.resources import files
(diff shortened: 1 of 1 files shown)


dimbleby:
[recent pipeline](https://github.com/python-poetry/poetry/actions/runs/33267335436/job/99139641790) 107s [this pipeline](https://github.com/python-poetry/poetry/actions/runs/33273409738/job/99155868237) 96s

10% is roughly in line with what I see locally

radoering (approved):

PR #11037, opened 2026-08-31
Fix sdist extraction on older Python versions

### LINKED ISSUE #11036: Regression in 2.4.2: sdist install fails as root on Python < 3.10.12 / < 3.11.4 with TypeError: getgrnam() argument 'name' must be str, not None
### Description

Since 2.4.2, `poetry install` fails as root while extracting an sdist on Pythons whose `tarfile` has no `data_filter`. We hit it in a Docker build on `python:3.10.4-bullseye` with `crontab 1.0.4`, `facepy 1.0.12` and `PyMeeus 0.5.12`, which have no wheel on PyPI.

On these Pythons `extractall()` in `poetry/utils/helpers.py` uses the ported CPython 3.14 data filter added in #11027. The filter sets `uid`, `gid`, `uname` and `gname` to `None`, and members are then extracted with `archive.extract(member, dest, set_attrs=not member.isdir())`. As root, `tarfile._extract_member()` calls `chown()`, where these Python versions have no `None` guard:

```python
# CPython 3.10.11, tarfile.py
try:
    if grp:
        g = grp.getgrnam(tarinfo.gname)[2]
except KeyError:
    pass
```

`grp.getgrnam(None)` raises `TypeError`, which is not caught, and the install aborts. CPython has `if grp and tarinfo.gname:` there from 3.10.12 and 3.11.4 on.

Checked with the same command, `poetry install --no-root` of `crontab==1.0.4` as root:

| Python | poetry 2.4.2 | poetry 2.4.1 |
|---|---|---|
| 3.10.4 | fails | ok |
| 3.10.11 | fails | not checked |
| 3.10.12 | ok | not checked |
| 3.11.3 | fails | not checked |
| 3.11.4 | ok | not checked |
| 3.10.18 | ok | not checked |

In 2.4.1 the same fallback branch validated member paths only and did not change ownership attributes.

### Reproduction

```console
$ docker run --rm -u root python:3.10.4-bullseye bash -c '
  pip install -q "poetry==2.4.2"
  mkdir /t && cd /t
  printf "[project]\nname=\"repro\"\nversion=\"0.1.0\"\nrequires-python=\">=3.10\"\ndependencies=[\"crontab==1.0.4\"]\n" > pyproject.toml
  POETRY_VIRTUALENVS_CREATE=false poetry install --no-root'
```

The same command with `poetry==2.4.1` succeeds.

### Workarounds

- Run the install as a non-root user, verified on `python:3.10.4-bullseye` with 2.4.2.
- Use Python 3.10.12+ or 3.11.4+, verified above.
- Pin `poetry==2.4.1`.

**Poetry Installation Method:** pip

**Operating System:** Debian 11 (`python:3.10.4-bullseye` container), also seen on GitHub Actions `ubuntu-24.04`

**Poetry Version:** 2.4.2

### Poetry Configuration

```
cache-dir = "/root/.cache/pypoetry"
installer.max-workers = null
installer.no-binary = null
installer.only-binary = null
installer.parallel = true
installer.re-resolve = false
solver.lazy-wheel = true
virtualenvs.create = false
```

### Poetry Runtime Logs

<details><summary>poetry-runtime.log</summary>

```
5  poetry/installation/chef.py:60 in prepare
     →  60│         return self._prepare_sdist(
4  poetry/installation/chef.py:106 in _prepare_sdist
     → 106│             extractall(source=archive, dest=archive_dir, zip=zip)
3  poetry/utils/helpers.py:529 in extractall
     → 529│                     archive.extract(member, dest, set_attrs=not member.isdir())
2  /usr/local/lib/python3.10/tarfile.py:2090 in extract
     → 2090│             self._extract_member(tarinfo, os.path.join(path, tarinfo.name),
1  /usr/local/lib/python3.10/tarfile.py:2178 in _extract_member
     → 2178│             self.chown(tarinfo, targetpath, numeric_owner)

TypeError
getgrnam() argument 'name' must be str, not None
at /usr/local/lib/python3.10/tarfile.py:2284 in chown
    2283│                     if grp:
  → 2284│                         g = grp.getgrnam(tarinfo.gname)[2]
Cannot install crontab.
```
</details>

fredrikblau:
## Summary

The tar security fallback clears ownership names with `None`. Python versions before 3.10.12 and 3.11.4 pass those values to `grp.getgrnam()` / `pwd.getpwnam()`, causing root sdist installs to fail.

Use empty ownership names instead, which preserves the ownership-stripping behavior without triggering the older tarfile bug. Added a regression test that simulates the affected tarfile implementation.

Fixes #11036.

## Tests

- `pytest tests/utils/test_helpers.py -q` (50 passed)
- `pytest` (2,935 passed, 28 skipped; 6 collection errors due missing `deepdiff` before installation, and one unrelated Python 3.14 platform-tag expectation)
- `ruff check tests/utils/test_helpers.py` (passed; existing `PLC0206` remains in `src/poetry/utils/helpers.py`)

### DIFF
--- a/src/poetry/utils/helpers.py
+++ b/src/poetry/utils/helpers.py
@@ -350,14 +350,15 @@ def extractall(source: Path, dest: Path, zip: bool) -> None:
                         if mode != member.mode:
                             new_attrs["mode"] = mode
                     # Ignore ownership for 'data'
-                    if member.uid is not None:
-                        new_attrs["uid"] = None
-                    if member.gid is not None:
-                        new_attrs["gid"] = None
-                    if member.uname is not None:
-                        new_attrs["uname"] = None
-                    if member.gname is not None:
-                        new_attrs["gname"] = None
+                    # Unlike newer versions, TarFile.chown() of older Python
+                    # versions does not tolerate None here: it passes the ids
+                    # straight to os.chown() and looks the names up with
+                    # pwd/grp, catching only KeyError. Use values that are
+                    # accepted and resolve to "leave ownership alone".
+                    new_attrs["uid"] = -1
+                    new_attrs["gid"] = -1
+                    new_attrs["uname"] = ""
+                    new_attrs["gname"] = ""
                     # Check link destination for 'data'
                     if member.islnk() or member.issym():
                         if os.path.isabs(member.linkname):
--- a/tests/utils/test_helpers.py
+++ b/tests/utils/test_helpers.py
@@ -1,6 +1,7 @@
 from __future__ import annotations
 
 import contextlib
+import sys
 import tarfile
 
 from pathlib import Path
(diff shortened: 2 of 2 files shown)


radoering (changes requested):
There are more spots with the pattern "try - except KeyError" and running the repro with Poetry from your branch still fails:

```
TypeError

  uid should be integer, not NoneType
```

(Just replace `poetry==2.4.2` with `git+https://github.com/fredrikblau/poetry.git@fix/older-python-sdist-extraction` in the repro.)

fredrikblau:
Handled the missing-ownership case from the repro. The fallback now always sets uid/gid to -1, and the regression test simulates omitted fields.

`pytest -q tests/utils` passes: 378 passed, 21 skipped. Ruff reports the existing PLC0206 in helpers.py; `git diff --check` passes.

Pandorin:
Tested 90b280f on the reproducer from https://github.com/python-poetry/poetry/issues/11036: it passes. 👍

fredrikblau:
Fixed the mypy failure in the regression setup by using object.__setattr__ to simulate missing ownership fields.

Focused tests: 50 passed. The hosted Windows pytest failure is unrelated PermissionError in test_env.py.

src/poetry/utils/helpers.py
```diff
@@ -350,14 +350,15 @@ def _get_filtered_attrs(
...
+                    # owner name is unavailable.
+                    new_attrs["uid"] = -1
+                    new_attrs["gid"] = -1
                     if member.uname is not None:
-                        new_attrs["uname"] = None
+                        # Older Python versions do not handle None in TarFile.chown.
+                        new_attrs["uname"] = ""
                     if member.gname is not None:
-                        new_attrs["gname"] = None
+                        new_attrs["gname"] = ""
```
radoering: Wouldn't it make more sense to also always set `new_attrs["uname"]` and `new_attrs["gname"]` to `""` because now we handle the case where they are not None but in case they were None we would run into the same exception?

fredrikblau: Yes, you're right — dropped the `is not None` guards so uid/gid/uname/gname are always set to -1/-1/""/"". The names go through `grp.getgrnam()`/`pwd.getpwnam()`, which raise `TypeError` on None but `KeyError` on "", so the empty string falls back to the -1 ids and `os.chown()` leaves ownership alone.

The regression test now builds the members with all four fields set to None and asserts the filter hands `(-1, -1, "", "")` to `chown()`; it fails with `TypeError` if either name is left as None.

Also rebased on main and squashed to one commit. `pytest tests/utils`: 378 passed, 21 skipped. mypy clean; ruff reports only the pre-existing PLC0206 in `helpers.py`.

radoering (approved):

PR #11042, opened 2026-09-05
Reduce virtual environment discovery subprocesses

dimbleby:
Avoid redundant interpreter launches in CLI commands and isolated builds by discovering environment data lazily and reusing it.

Motivated by making the unit tests run faster, but this one should also have a modest impact for regular poetry users.  Saves launching a subprocess python interpreter more often than needed


src/poetry/utils/env/virtual_env.py
```diff
@@ -35,14 +41,22 @@ class VirtualEnv(Env):
...
 
-        # If base is None, it probably means this is
-        # a virtualenv created from VIRTUAL_ENV.
-        # In this case we need to get sys.base_prefix
-        # from inside the virtualenv.
-        if base is None:
+    @property
+    def base(self) -> Path:
+        if self._base is None:
             output = self.run_python_script(GET_BASE_PREFIX)
```
radoering, line 47: Why not use `_environment_data` here?

tests/inspection/test_info.py
```diff
@@ -366,11 +365,9 @@ def test_info_no_setup_pkg_info_no_deps_dynamic(fixture_dir: FixtureDirGetter) -
     assert info.requires_dist is None
 
 
-def test_info_setup_simple(mocker: MockerFixture, demo_setup: Path) -> None:
-    spy = mocker.spy(VirtualEnv, "run")
+def test_info_setup_simple(demo_setup: Path) -> None:
     info = PackageInfo.from_directory(demo_setup)
 
-    assert spy.call_count == 6
```
radoering, line 373: The number changes, but why shouldn't we keep checking it?

tests/utils/env/test_env.py
```diff
@@ -101,19 +101,42 @@ def test_env_get_supported_tags_matches_inside_virtualenv(
...
     expected_tags = list(packaging.tags.sys_tags())
 
     if differing_platform:
         mocker.patch("sysconfig.get_platform", return_value="some_other_platform")
-        expected_call_count = 2
-    else:
-        expected_call_count = 1
 
     assert venv.get_supported_tags() == expected_tags
-    assert run_python_script_spy.call_count == expected_call_count
```
radoering, line 116: Why not keeping this check?

tests/inspection/test_info.py
dimbleby: I do not think it is useful to check the count - as this pull request shows, no-one cares if the number changes; so what is the test pinning?  But we can keep it anyway.

tests/utils/env/test_env.py
dimbleby: again I do not think it is a useful check but we can keep it

tests/inspection/test_info.py
radoering: Maybe. I think lower is better and an increasing number might show a regression if there is no good reason.

radoering (approved):

PR #11043, opened 2026-09-06
docs: update links to relocated PyPI documentation

blackorange-mega:
PyPI's user documentation moved from `warehouse.pypa.io/api-reference/` to `docs.pypi.org`, and pipx moved off `pypa.github.io`. Six links across three docs pages currently 404.

| File | Old link | New link |
| --- | --- | --- |
| `docs/repositories.md` (L335) | `warehouse.pypa.io/api-reference/json.html` | `docs.pypi.org/api/json/` |
| `docs/repositories.md` (L72, L433, L445) | `warehouse.pypa.io/api-reference/legacy.html#upload-api` | `docs.pypi.org/api/upload/` |
| `docs/cli.md` (L952) | `warehouse.pypa.io/api-reference/xml-rpc.html#deprecated-methods` | `docs.pypi.org/api/#api-preference` |
| `docs/contributing.md` (L241) | `pypa.github.io/pipx/` | `pipx.pypa.io/stable/` |

Every replacement URL was checked and returns 200; the old ones return 404.

**One case needed a judgement call.** The XML-RPC page has no direct successor — that API was removed rather than relocated, and `docs.pypi.org/api/` lists only the Index, JSON, BigQuery and Upload APIs. The deprecation notice the old `#deprecated-methods` anchor pointed at now lives in the **API Preference** section of the API index ("No new integrations should use the XML-RPC APIs as they are planned for deprecation"), so I pointed there. The surrounding note about PyPI search still reads correctly. If you would rather drop that link entirely, say the word.

Docs only; no code or behaviour changes.


blackorange-mega:
Good catch — the link text said "warehouse documentation" while pointing at `docs.pypi.org`. Renamed it to "PyPI API documentation".

Checked the other labels I touched and left them alone, since they describe their destinations accurately: "JSON API" → `docs.pypi.org/api/json/`, "Upload API"/"Legacy Upload API" → `docs.pypi.org/api/upload/`. No "warehouse" wording remains in the three files.

radoering (approved):

PR #11051, opened 2026-09-08
Withdrawn: invalid marker simplification for #10989

dajiaohuang:
This PR is withdrawn. The implementation from 9db3711d incorrectly simplified cross-variable markers by comparing marker names only, so it was reverted in f329f78d. The current branch has no net diff against main and proposes no fix. #10989 remains a cosmetic marker discussion; a correct change would require an explicit supported semantics decision for platform_system and sys_platform, which is not established here. This Draft is retained to preserve the review history.


dimbleby:
Without baking in some knowledge of the semantic meaning of `platform_system` and `sys_platform` you can't clean up the ugly markers of #10989.  

The whole point of the issue report is that poetry should somehow know that `platform_system == "Linux"` and `sys_platform == "darwin"` are mutually exclusive.

This pull request doesn't do that: so it can't be addressing #10989.

It looks as though you are "simplifying" `platform_system == "Linux" and sys_platform == "darwin"` to become `platform_system == "Linux"`, which is plainly wrong.

dajiaohuang:
You are right. The previous patch removed `sys_platform == "darwin"` based only on marker names and therefore assumed a semantic relationship that it did not encode. I reverted that change in f329f78; this PR now has no net code change and does not claim to fix #10989. The issue needs an explicit marker-semantics decision before a safe implementation.

PR #11058, opened 2026-09-12
Fix repeated updates caused by stale VCS source directories

### LINKED ISSUE #7089: (🐞) `poetry install` keeps installing things that are already installed, when they were previously a git dependency
- **Poetry version**:  1.2.2
- **Python version**:  3.10
- **OS version and name**:  Windows 10 / wsl Ubuntu 22.04
- **pyproject.toml**: 
```toml
[tool.poetry]
name = "poetry-test"
version = "0.1.0"
description = ""
authors = [""]

[tool.poetry.dependencies]
python = "^3.10"
pendulum = { git = "https://github.com/sdispater/pendulum.git", rev = "9d06ff3c9805869ccfc4163172b658482ec50087" }
#pendulum = "^2"
```

- [🚀] I am on the [latest](https://github.com/python-poetry/poetry/releases/latest) stable Poetry version, installed using a recommended method.
- [🚀] I have searched the [issues](https://github.com/python-poetry/poetry/issues) of this repo and believe that this is not a duplicate.
- [🚀] I have consulted the [FAQ](https://python-poetry.org/docs/faq/) and [blog](https://python-poetry.org/blog/) for any relevant entries or release notes.
- [🚀] If an exception occurs when executing a command, I executed it again in debug mode (`-vvv` option) and have included the output below.

## Steps:
```
👉 poetry update
Updating dependencies
Resolving dependencies... (0.7s)

Package operations: 5 installs, 0 updates, 0 removals

  • Installing six (1.16.0)
  • Installing python-dateutil (2.8.2)
  • Installing time-machine (2.8.2)
  • Installing tzdata (2022.6)
  • Installing pendulum (3.0.0a0 9d06ff3)
```
At this point we update the pendulum dependency to `"^2"`, then continue
```
👉 poetry update
Updating dependencies
Resolving dependencies... (0.7s)

Package operations: 1 install, 1 update, 0 removals

  • Installing pytzdata (2020.1)
  • Updating pendulum (3.0.0a0 9d06ff3 -> 2.1.2)
👉 poetry install

Package operations: 0 installs, 1 update, 0 removals

  • Updating pendulum (2.1.2 9d06ff3 -> 2.1.2)
👉 
```

# What The Glitch❓❓❓❓
`• Updating pendulum (2.1.2 9d06ff3 -> 2.1.2)`

aprylewu:
# Pull Request Check List

Resolves: #7089

- [ ] Updated **documentation** for changed code. N/A: fixes installed-package detection without changing the public interface.

A checkout left in `<venv>/src/<name>` after switching from a Git dependency to an index release is currently enough for Poetry to classify the installed release as a Git package. This makes subsequent installs repeatedly schedule an update even when the installed version is already correct.

Stop treating the presence of that source directory alone as installation metadata. Keep the existing `.pth`, source-local egg-info, and PEP 610 detection paths. The checkout is left intact.

The new regression creates actual temporary dist-info metadata and a source directory, with and without a `.pth` link. On the original code, the unlinked case fails because it is incorrectly classified as Git; the linked case passes.

Validation on macOS / Python 3.13.5:

- `pytest -n 0 tests/repositories/test_installed_repository.py`: 29 passed, including existing `.pth`, legacy egg-info, and PEP 610 cases.
- `pytest -n 0 tests/puzzle/test_transaction.py`: 89 passed.
- An independent probe using actual local Git repositories (no Git-info mock) passed 8 scenarios: the stale checkout now yields an "Already installed" skip instead of an update, while valid `.pth`, legacy egg-info, and PEP 610 sources retain their original metadata.
- `mypy`: no issues in 380 source files.
- `pre-commit run --files src/poetry/repositories/installed_repository.py tests/repositories/test_installed_repository.py`: all applicable checks passed.

The unit tests use temporary metadata and existing Git-info fixtures. I also ran an offline CLI reproduction of #2921 using a local Git dependency at a fixed package version: advancing the branch updates the imported module, lock file, and PEP 610 commit ID, both with normal metadata and after removing the origin metadata. Subsequent `poetry install` runs skip the dependency. Poetry and poetry-core were tested from current main; the entire project test suite was not run.

Compatibility with #3867: that PR introduced the directory fallback removed here. Existing installs without `.pth` or `direct_url.json` lose the inferred Git classification. A Git requirement then schedules a reinstall even at the same version; the executor writes origin metadata so later installs can skip it. This can cause a one-time reinstall for such legacy environments. Four new cases cover unchanged/changed Git revisions with and without origin metadata; `.pth`, source-local egg-info, and PEP 610 discovery remain supported.

Prepared with OpenAI Codex assistance and independently reviewed by another Codex agent.


dimbleby:
This reverses #3867 - presumably it breaks whatever that fixed

aprylewu:
Yes, #3867 introduced exactly this fallback; I should have checked that history before submitting.

I checked #2921 and added four cases in 7e79226 covering same-version Git revision changes, with and without `direct_url.json` and no `.pth`. I also ran the actual CLI against a local Git repository: after advancing the branch, the imported module, lock file and recorded commit all update, including when the previous origin metadata is missing. The next `poetry install` skips the dependency. All 118 related tests, mypy and pre-commit pass.

There is a compatibility change: a metadata-less legacy install is no longer inferred to be Git from the checkout directory. A Git requirement triggers one reinstall, after which the executor writes PEP 610 metadata and subsequent installs skip it. I verified that unchanged-commit migration too and documented this one-time reinstall in the PR description. Existing `.pth`, source-local egg-info and PEP 610 paths remain supported.

Issue #11069, opened 2026-09-18
Update `poetry-core` from 2.4.0 to 2.4.1 in 2.4 release branch


jmertic:
### Description

Since `poetry-core` 2.4.1 has been released, installing poetry from a via the following command...

```
pip compile --generate-hashes --universal -o requirements-ci.txt -
pip install --require-hashes -r requirements-ci.txt
```

...results in this error...

```
ERROR: Cannot install -r requirements-ci.txt (line 688) and poetry-core==2.4.1 because these package versions have conflicting dependencies.

The conflict is caused by:
    The user requested poetry-core==2.4.1
    poetry 2.4.3 depends on poetry-core==2.4.0

Additionally, some packages in these conflicts have no matching distributions available for your environment:
    poetry-core
```

This is due to both `poetry.lock` and `pyproject.toml` pinned to version 2.4.0 vs 2.4.1. So the fix should be just updating the version in both files.

Thanks for all the great work this team does!

### Workarounds

For now, I manually change back to `poetry-core` 2.4.0 after dependabot likes to update to `2.4.1`

### Poetry Installation Method

pip

### Operating System

Any

### Poetry Version

2.4.3

### Poetry Configuration

```bash session
This is an install issue, so no config.
```

### Example pyproject.toml

```TOML

```

### Poetry Runtime Logs

This is an install issue, so no config.

dimbleby:
If `pip compile` produces requirements which it then cannot install: that sounds like a bug with `pip compile` - not something to fix here.

dimbleby:
though since `pip compile` doesn't even seem to be a command that exists, I suppose this could also be closed as nonsense.

```
$ pip compile --generate-hashes --universal -o requirements-ci.txt -
ERROR: unknown command "compile" - maybe you meant "completion"
```

@jmertic please provide a reproducer that actually reproduces something; and then raise a bug report against the appropriate project.

radoering:
poetry-core was not updated intentionally because 2.4.1 introduces a regression for poetry-plugin-export. See https://github.com/python-poetry/poetry/pull/11070#issuecomment-5730002727 for example.

jmertic:
Got it - that's helpful @radoering - I will for not just revert when dependabot wants to update. Thanks!

dimbleby:
@jmertic it sounds likely that you are doing it wrong - unless you maybe are writing a poetry plugin, there should be no reason for you to have a dependency on poetry-core, therefore nothing for dependabot to update.

PR #11070, opened 2026-09-18
BUG: Bump poetry-core pin to 2.4.1 on 2.4 branch

karlhillx:
Closes #11069.

Bumps the exact-version pin on `poetry-core` from `==2.4.0` to `==2.4.1`
in the 2.4 maintenance branch, and regenerates `poetry.lock` so the
content hash and wheel/sdist hashes match the new release.

The pin was frozen to 2.4.0 when the branch was cut; poetry-core 2.4.1
shipped afterward (extra-marker evaluation fix, no API changes). Anyone
resolving Poetry 2.4.3 with `pip-compile --generate-hashes` +
`pip install --require-hashes` against an environment that wants 2.4.1
hits an unresolvable constraint because the published wheel metadata
declares `poetry-core==2.4.0` exactly. Bumping the pin lets hash-verified
installs succeed.

Maintenance-branch scope: this PR targets `2.4`, not `main`. A patch
release (2.4.4) cut from this branch and published to PyPI is what
actually unblocks downstream --require-hashes users; the PR alone just
makes that release possible.

Verification:
- `poetry check` passes
- `poetry update poetry-core` produces a poetry-core-only diff (10 lines,
  0 unrelated bumps)
- `git diff --stat` shows only `pyproject.toml` (1 line) + `poetry.lock`
  (10 lines)
- focused test sweep on `tests/masonry/` and `tests/pyproject/` (the
  marker/extra-handling surface that 2.4.1 changed) passes 12/12


karlhillx:
Closing this — the bump to `poetry-core==2.4.1` introduces a regression
in `poetry-plugin-export`'s `tests/test_exporter_pylock_toml.py::test_export_markers`
(2 parametrized cases fail across Ubuntu/Windows/macOS on Python 3.10).

Failure: the `pylock.toml` exporter no longer emits the expected
`marker = "python_version >= \"3.6\""` line for fixtures using
`python_version >= "3.6" or extra == "extra1"`. The 2.4.1 release notes
flagged an extra-marker evaluation fix; that fix changes how the exporter
serializes these markers, and the plugin tests predate the change.

Notable: `poetry-core==2.4.0` is still pinned on the `2.4.3` release tag
as well as on `2.4` — so this bump hasn't landed upstream either, and
the maintenance branch is intentionally staying on 2.4.0.

I should have caught this before opening the PR — my local verification
ran Poetry's own `tests/masonry/` and `tests/pyproject/` (which exercise
the resolver, not the exporter), and missed `poetry-plugin-export`
entirely because it lives in a separate repo with its own CI. Closing
this PR; the right follow-up is either:

- wait for `poetry-core 2.4.2` (if upstream reverts or refines the
  marker fix) and bump to that, with `poetry-plugin-export` test
  fixtures updated alongside, OR
- stay on `2.4.0` indefinitely, matching what `2.4.3` does today

Apologies for the noise — the failure is real, my PR caused it, and
the correct action is to revert.

